PR #2322, opened 2015-04-13
Make completions scroll to top of list

### LINKED ISSUE #1876: Editor: Move suggested completion item to top of window
_From [w0run.go...@gmail.com](https://code.google.com/u/106279936266297459333/) on 2014-07-01T16:19:46Z_

If I type

import pylab as py

Then, when I type:
  py.s 

then the flying window comes up and shows "py.s_" at the bottom of the window- all other "py.s" offerings are out of sight below the window. It would be much better to have the first matching item at the top of the window, so the others can be seen below. I looked through the editor preferences, and this did not seem to be one of them.

Using Spyder 2.30.rc with Anaconda.

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=1876_

blink1073:
Fixes #1876.

screen shot 2015-04-12 at 9 09 07 pm

### DIFF
--- a/spyderlib/widgets/sourcecode/base.py
+++ b/spyderlib/widgets/sourcecode/base.py
@@ -19,7 +19,7 @@ from spyderlib.qt.QtGui import (QTextCursor, QColor, QFont, QApplication,
                                 QTextEdit, QTextCharFormat, QToolTip,
                                 QListWidget, QPlainTextEdit, QPalette,
                                 QMainWindow, QTextOption, QMouseEvent,
-                                QTextFormat, QClipboard)
+                                QTextFormat, QClipboard, QAbstractItemView)
 from spyderlib.qt.QtCore import Signal, Slot, Qt, QEventLoop, QEvent, QPoint
 from spyderlib.qt.compat import to_qvariant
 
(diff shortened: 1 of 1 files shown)


goanpeca:
Love this :+1:

ccordoba12:
This is really cool!! Thanks @blink1073!!

Could we also narrow down the completion list to only show those items that match the completion text?

goanpeca:
How easy/hard would it be to make the code completion more attractive (like [pycharm](https://www.jetbrains.com/pycharm/))... would it be too much work to have an icon for private methods... constants.. classes... methods etc??

ccordoba12:
That info is provided by Jedi (Unfortunately, I don't know if by Rope too). Maybe @blink1073 could submit a new PR implementing that too? ;-)

goanpeca:
Please @blink1073, it would be something more aesthetic but would make the completion look so much better!!!

blink1073:
@ccordoba12, counter proposal: what if I implemented fuzzy logic and only show the completions from the original list that contain the entered characters?

A quick look shows that we can access this from the rope completion proposals.  Completions could be returned as (completion, type) tuples with associated icons.  I don't mind implementing the logic, but I do not want to get into the icon business.  I'll start another issue for discussion on that.

ccordoba12:
> @ccordoba12, counter proposal: what if I implemented fuzzy logic and only show the completions from the original list that contain the entered characters?

I don't like fuzzy logic that much, but it could be cool to have it.

But let's leave it for another PR. This one is simple enough and solves precisely what it wants to accomplish.

PR #2328, opened 2015-04-15
README.md running from source is a killer feature

techtonik:
The rationale is based on assumption that README.md is written for GitHub users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,8 +53,6 @@ or PySide, and run:
 
 You may want to do this for fixing bugs in Spyder, adding new
 features, learning how Spyder works or just getting a taste of it.
-Make sure you use a stable branch though, because `master` may not
-always be clean.
 
 ## Installation
     
@@ -87,7 +85,7 @@ The easiest way to install Spyder is:
     - Using the [Anaconda Distribution](http://continuum.io/downloads).
     - Installing from source (see below).
 
-* Cross-platoform way from source
+* Cross-platform way from source
 
 You can install Spyder from its zip source package named `spyder-x.y.z.zip`
 [here](https://github.com/spyder-ide/spyder/releases). Then you need to
@@ -95,7 +93,7 @@ use standard Python `setup.py` script:
 
     python setup.py install
 
-Note that `setup.py` script doens't know how uninstall previous versions
+Note that `setup.py` is not able to uninstall previous versions
 of Python packages: it simply copies files on top of an existing
 installation. So the better way is to use `pip` package manager:
 
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 Copyright © 2009-2013 Pierre Raybaut.
 Licensed under the terms of the MIT License
-(see `spyderlib/__init__.py` for details)
+(see `LICENSE` for details)
 
 
 ## Overview
@@ -46,12 +46,15 @@ Spyder may be embedded in your own PyQt4/PySide application.
 
 ## Running from source
 
-It is possible to run Spyder from the source code checkout, given
-that you have PyQt or PySide installed. Just run:
+The fastest way to run Spyder is to get the source code, install PyQt
+or PySide, and run:
 
     python bootstrap.py
 
-This is useful if you want to learn how Spyder works and enhance it.
+You may want to do this for fixing bugs in Spyder, adding new
+features, learning how Spyder works or just getting a taste of it.
+Make sure you use a stable branch though, because `master` may not
+always be clean.
 
 ## Installation
     
@@ -84,7 +87,7 @@ The easiest way to install Spyder is:
     - Using the [Anaconda Distribution](http://continuum.io/downloads).
     - Installing from source (see below).
 
-### Installing from source
+* Cross-platoform way from source
 
 You can also install Spyder from its zip source package. For that you need to
 download and uncompress the file called `spyder-x.y.z.zip`, which can be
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -149,17 +157,6 @@ greater than 2.6.
 
 * Pep8 (style analysis).
 
-## Running from source
-
-It is possible to run Spyder directly (i.e. without installation)
-from the unpacked zip folder (see *Installing from source*) using
-Spyder's bootstrap script like this:
```
Nodd: unpacked zip folder or git repository!
Also, having a reference to "Installing from source" right after saying that you don't need to install it is a bit disturbing, but maybe it's only me.

README.md
```diff
@@ -44,6 +44,14 @@ Spyder may also be used as a PyQt4/PySide extension library (module
 `spyderlib`). For example, the Python interactive shell widget used in
 Spyder may be embedded in your own PyQt4/PySide application.
 
+## Running from source
```
ccordoba12: Why put this section here? I think the place where it was placed was better than this one.

We want to tell people how to install Spyder first, not how to run it from source.

README.md
Nodd: Uh, why did I comment on the removed part ? :confused:

README.md
```diff
@@ -44,6 +44,14 @@ Spyder may also be used as a PyQt4/PySide extension library (module
...
+
+It is possible to run Spyder from the source code checkout, given
+that you have PyQt or PySide installed. Just run:
```
Nodd: Why the reference to PyQt and PySide here ? Spyder also needs numpy, shinx, and others, right ?

ccordoba12: PyQt is the minimum needed. All other modules are optional.

README.md
techtonik: Did you know about the feature before?

README.md
techtonik: Who are we? As I already said, the README.md is for GitHub users, who prefer to work with sources if possible. For everybody else there are other other channels.

Cloning and running is that **fastest** way a person from GitHub can run Spyder. I don't see why we want to put users through the installation rituals.

techtonik:
Added more edits to omit obvious parts and get new/important information to the readers faster.

README.md
ccordoba12: I don't know, maybe using the Readme to help Github users is the best thing to do in 2015 :-)

@Nodd, what do you think?

ccordoba12: If others can comment on this one, that'd be great!

@goanpeca, @blink1073, @SylvainCorlay, what do you think?

README.md
```diff
@@ -76,26 +87,23 @@ The easiest way to install Spyder is:
     - Using the [Anaconda Distribution](http://continuum.io/downloads).
     - Installing from source (see below).
 
-### Installing from source
+* Cross-platoform way from source
```
blink1073: platform*

README.md
blink1073: @ccordoba12, I agree with taking a developer-first stance on a Github readme, as long as we also mention `pip install` somewhere.

README.md
Nodd: Yes.

README.md
```diff
@@ -44,6 +44,17 @@ Spyder may also be used as a PyQt4/PySide extension library (module
...
+
+The fastest way to run Spyder is to get the source code, install PyQt
+or PySide, and run:
+
+    python bootstrap.py
+
+You may want to do this for fixing bugs in Spyder, adding new
+features, learning how Spyder works or just getting a taste of it.
+Make sure you use a stable branch though, because `master` may not
+always be clean.
```
Nodd: master is not always clean, but I can't remember a time where spyder didn't work at all. The precision here seems too much.

README.md
```diff
@@ -76,26 +87,23 @@ The easiest way to install Spyder is:
...
 
     python setup.py install
 
-Note that `distutils` does *not* uninstall previous versions of Python 
-packages: it simply copies files on top of an existing installation. 
-When using this command, it is thus highly recommended to uninstall 
-manually any previous version of Spyder by removing the associated 
-directories ('spyderlib' and 'spyderplugins') from your site-packages 
-directory).
+Note that `setup.py` script doens't know how uninstall previous versions
```
Nodd: how **to** uninstall

Nodd:
BTW python can run modules from zip files. Is in the case for spyder ? I think it requires a __main__.py or something...

README.md
techtonik: If there is user level docs, it is possible to add a redirect to them at the top of README.md The top also needs a link to wiki KB.

README.md
techtonik: Fixed.

README.md
techtonik: Removed stuff about master. Maybe it is just my system, but current master is extremely slow.

README.md
techtonik: Fixed.

techtonik:
I can try to wrap an executable Spyder .zip, but I am not sure that pip will be able to install it - https://github.com/pypa/pip/issues/2406

README.md
Nodd: You should report it then !

Nodd:
That would require to create a separate executable zip file. Not worth the hassle, it was just a random suggestion.

ccordoba12:
I think this is ready. Thanks for the update to our README @techtonik. It was really needed :-)

Merging then.

PR #2331, opened 2015-04-15
Finish Introduction tour

ccordoba12:
This PR adds the missing contents of the *Introduction tour*.

I decided to create it given that this was @goanpeca's work, so I'd like to know his thoughts about my additions.

### DIFF
--- a/spyderlib/tour.py
+++ b/spyderlib/tour.py
@@ -60,17 +60,26 @@ class SpyderWidgets(object):
     menu_edit = ''
 
 
-def get_tours():
-    """ """
-    return get_tour(None)
+def get_tours(index=None):
+    """
+    Get the list of available tours (if index=None), or the your given by
+    index
+    """
+    return get_tour(index)
 
 
 def get_tour(index):
-    """For now this function stores and retrieves the tours.
+    """
+    This function generates a list of tours.
+    
+    The index argument is used to retrieve a particular tour. If None is
+    passed, it will return the full list of tours. If instead -1 is given,
+    this function will return a test tour
 
     To add more tours a new variable needs to be created to hold the list of
-    dics and the tours variable at the bottom of this function needs to be
-    updated accordingly"""
+    dicts and the tours variable at the bottom of this function needs to be
+    updated accordingly
+    """
     sw = SpyderWidgets
 
     # This test should serve as example of keys to use in the tour frame dics
@@ -219,10 +228,10 @@ def get_tour(index):
 
     if index is None:
         return tours
-    elif index == 'test':
-        return test
+    elif index == -1:
+        return [test]
     else:
-        return tours[index]['tour']
+        return [tours[index]]
 
 
 class FadingDialog(QDialog):
--- a/spyderlib/tour.py
+++ b/spyderlib/tour.py
@@ -35,9 +35,9 @@ from spyderlib.utils.qthelpers import (create_action, add_actions)
 class SpyderWidgets(object):
     """List of supported widgets to highlight/decorate"""
     # Panes
-    console_internal = 'console'
-    console_external = 'extconsole'
-    console_ipython = 'ipyconsole'
+    internal_console = 'console'
+    external_console = 'extconsole'
+    ipython_console = 'ipyconsole'
     editor = 'editor'
     editor_line_number_area = 'editor.get_current_editor().linenumberarea'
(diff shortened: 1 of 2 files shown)


goanpeca:
Hey @ccordoba12 thanks for working on this. It looks fine on a quick look, but there is an issue that needs to be solved, maybe the ipython console is not intalled, so we need to include this in the logic of the tours.py

spyderlib/tour.py
```diff
@@ -108,41 +118,72 @@ def get_tour(index):
...
+                           "<a href=\"%s\">here</a>.<br><br>"
+                           "It's also important to note that you can open as "
+                           "many consoles as you want, by going to the "
+                           "menu entry<br><br>"
+                           "<i>Consoles > Open an IPython console</i>.<br><br>"
+                           "Please click on the button below to run some simple "
+                           "code in this console. This will be useful to show "
+                           "you other important Spyder features." % \
+                           qtconsole_link),
+              'widgets': [sw.ipython_console],
```
goanpeca: This will blow if ipython is not installed...

ccordoba12: Yes, we need to verify that `ipython_console` exists before adding this part to the tour.

spyderlib/tour.py
```diff
@@ -108,41 +118,72 @@ def get_tour(index):
...
+             {'title': _("The Python console"),
+              'content': _("You can also run your code on a <b>Python console</b>. "
+                           "These consoles are useful because they are created "
+                           "much faster than an IPython console, and let you "
+                           "run each file in a dedicated console<br><br>"
+                           "By pressing the button below, you will notice that "
+                           "<b>Python consoles</b> are also connected to the "
+                           "Variable Explorer, and that the last one only shows "
+                           "the variables of the currently focused console."),
+              'widgets': [sw.external_console, sw.variable_explorer],
```
goanpeca: I would advice against this based on a suggestion by @Nodd.

Even though the functionality allows for highlighting any number of widgets at the same time, there might be the case where someone has the widgets in the same dockwidget location, in this case this lines will "fail" so it is better to use ONE and only ONE widget per slide. 

If you want to display how the variable explorer stored something... then it should be an additional slide

spyderlib/tour.py
goanpeca: @ccordoba12 can you please use `'Text {0}'.format(variable)` instead of `%`, I think is better and more flexible.

goanpeca:
What should be the default behavior when a widget is not there? 
skip it? make it non-selectable? make it like grayed out in the combobox?

Depending on this I can make a PR to correct this, and also include the suggestion of @Nodd of using a dummy dockwidget instead of making it transparent to events (the temporal fix used right now...)

Nodd:
It would be bad to skip it completely since the tour is there to show all features of spyder to new users.

Maybe add a warning that the corresponding widget it not active ?

ccordoba12:
The only plugin that can be missing right now is the IPython console (which is loaded on demand). All other plugins (even the Pylint one) load fine.

But @Nodd has a good point: if the IPython Console is missing, the tour should show a warning and guide the user on how to install it.

Nodd:
There is also the case where a panel is not enabled. Is this manager ? (if I remember well, @goanpeca told me it was working, but it was a long time ago)

goanpeca:
If the panel is not enabled, it should be a problem... as far as I remember (it will be enabled and brought to the front)

Sure, we can then display a replacement content indicating how to install ipython.

spyderlib/tour.py
ccordoba12: Ok, I'll change it. But my idea would be to use the new custom layouts functionality to always show this tour uisng the *Spyder* layout.

But that's for another PR :-)

ccordoba12:
intro I want to make clear that the purpose of this PR is only to add missing **contents** to the tutorial. 

I'm not going to touch the other issues mentioned by @goanpeca and @Nodd to keep things simple :-)

ccordoba12:
@goanpeca, this is ready for a review. I added reasonable contents to all the Intro tour slides.

goanpeca:
Cool, will test it later on :-)

ccordoba12:
@goanpeca, I think this is ready for merge too ;-)

goanpeca:
Go for it!

ccordoba12:
Oki doki! ;-)

PR #2345, opened 2015-04-21
Add lock/unlock option for panes

### LINKED ISSUE #2096: Feature Request: Add option to lock window/pane layout
_From [doug.t...@gmail.com](https://code.google.com/u/107157864673511712940/) on 2014-12-29T14:22:36Z_

There should be an option to lock the window (panel, pane, widget, whatever you want to call it) layout so that accidental click-and-drag actions do not move windows around.

The most common case, at least for me, is going up to close a tab on the editor and accidentally moving it somewhere else.

If someone shows me where to look, I might be able to patch in this feature. I'd probably just add a checkmark menu item to the View menu.

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=2096_

goanpeca:
Fixes #2096

---

## Description
This PR fixes three things:
- Adds ability to lock panes. The locked state is True by default (i.e. first start of spyder after reset), and is remembered every time is changed.
- Rephrase Dockwidget mention in preferences to panes (the official name used for dockwidgets in Spyder lingo) 
- Reorganize the View menu structure.

## Menu entry

**Before**

image

**Now**

image

## Preferences dialog

New wording (Replaced dockwidgets by panes)

image

## Todo:

- [x] Agree on a shortcut
- [x] Agree on location of menu
- [x] Reorganize entries in View menu.


goanpeca:
@dougthor42 can you give it a try?

goanpeca:
Hmmm I think we should change places where it says 'dockwidget' to pane... otherwise the preferences looks funny....

`Vertical dockwidget title bars` -> dockwidget is a Qt term.. irrelevant to the user

So maybe it should say

`Vertical title bars on panes`

Nodd:
I agree on the location in the menu, but I wouldn't add this option in the preferences at all. The lock state should be remembered, that's all.(and disabled by defalut, new users should be able toe asily change the layout).
I don't have a strong opinion on the shortcut, this one is coherent with *CLose current pane*.

Going back to the menu, what do you think about grouping options dirferentely (it could be another PR). The point is to better group what goes together:
* Toolbars
* -------------------------
* Panes
* Lock panes
* Close current pane
* Maximize cuirrent pane
* -------------------------
* Custom window layouts
* Toggle previous layout (is "toggle" the right word here ? How about "Use", or no verb at all ?)
* Toggle next layout
* -------------------------
* Full screen mode

I don't know where the attached console window comes from so I don't know where to put it.

goanpeca:
:+1: makes sense.

---

Yes i agree on the menu change, what do you think of changing the word dockwidgt by pane? It would be more coherent

goanpeca:
If the others agree with @Nodd I can make the corrections he suggests + if you agree I could also arrange the menu, as it indeed would be better...

I think I would like
```
Toolbars ->
-------------------------
Panes ->
Lock panes
Close current pane
Maximize current pane
-------------------------
Custom window layouts ->
Use previous layout 
Use next layout
-------------------------
Full screen mode
-------------------------
Attached console window
```

Nodd:
:+1: for pane instead of dockwidget.
Also I'd prefer `Previous layout` and `Next layout`.

goanpeca:
:+1: sure ;-)

dougthor42:
Works wonderfully. Thanks a bunch

ccordoba12:
Just a minor comment: I think the lock option should be **on** by default, not off :-)

Will come back later.

goanpeca:
@ccordoba12 by **on** you mean having panes locked by default (no possibility of rearranging them?), I thought you prefer **off**.

I prefer **on** by default as well... but I had the impression the rest of @spyder-ide/collaborators preferred **off**

Nodd:
Personally, one of the first things I do when launching spyder on a new computer is to rearrange the panes. Maybe I'm too much of a "power user".
On the other hand if a new user accidentally closes a pane he may be in trouble. But he may never discover that the panes can be moved around. I don't know what is the most important...

goanpeca:
I personally find super annoying that sometimes the panes get crazy because I missed a mouse click... but when in keyboard only mode (most of the time)... that is hardly a problem. Now *off*  goes against discover-ability maybe, but I also think first impressions are lasting ones, so if a panes goes crazy for a new user the user might think the software is buggy...

dougthor42:
I think that the default, unconfigured value should be Locked Panes = **Off**.

**For new users:**
It's likely that the first thing that someone will look for after accidentally moving a pane will be the "Lock Panes" feature. By having it default to off, they can see that that they can move panes around. Perhaps it's just because I'm used to pane-based programs, but I immediately recognized when I accidentally moved a pane. I like to believe that most users, even brand new ones, would be think the same thing...

**For users coming from PyCharm, Visual Studio, Eclipse, etc:**
The first thing that these users will do is try to move things around to how they like them. After things are set, then I'd imagine that they go look for the lock/unlock option.
If they can't move things right off the bat, they might not think to look for a lock/unlock option.


Or it could just be added to the tutorial.

ccordoba12:
@goanpeca, I changed my mind after altering my layout several times because of misplaced mouse clicks ;-)

@dougthor42, it's true that power users always want to rearrange things to their liking. But I think Spyder is picked up by a lot of newbies, so I'd prefer to adjust things for them.

But power user can always lookup for the option to unlock panels and adjust things again ;-)

goanpeca:
So... change to locked by default? I say :+1: to that.

---

@ccordoba12 any comments on reorganizing the menu?

ccordoba12:
* Yep, :+1: from me too ;-)
* Also :+1: for the shortcut
* About the menu, my proposal is

```
Panes ->
Lock panes
Close current pane
Maximize current pane
-------------------------
Toolbars ->
-------------------------
Custom window layouts ->
Use previous layout 
Use next layout
-------------------------
Full screen mode
-------------------------
Attached console window
```

I think `Panes` are much more important than `Toolbars` and deserve to be first in that menu :-)

goanpeca:
:+1: 

Should I add this change to this PR, or should I open a new one. 

--

Any thought on changing the entries in the preferences dialog where it says: dockwidgets... -> it should read panes... (dockwidget is not a very clear concept for someone not use to Qt)

Nodd:
I originally put the panes near the layouts because the layout change the panes. But I agree that they are more important so at the top is also fine.

`Custom window layouts` should be `Window layouts` because there are predefined ones.

I vote that reordering the menu and changing a few labels can be done in this PR, as long as they are separate commits.

ccordoba12:
Yes, please do those changes in this PR too.

goanpeca:
@ccordoba12 done!

spyderlib/config.py
```diff
@@ -469,11 +470,12 @@ def is_ubuntu():
               # ---- Global ----
               # -- In spyder.py
               '_/close pane': "Shift+Ctrl+F4",
+              '_/lock panes': "Shift+Ctrl+F5",
               '_/preferences': "Ctrl+Alt+Shift+P",
               '_/maximize pane': "Ctrl+Alt+Shift+M",
               '_/fullscreen mode': "F11",
-              '_/toggle next layout': "Shift+Alt+PgDown",
-              '_/toggle previous layout': "Shift+Alt+PgUp",
```
ccordoba12: Since you're removing these options, you need to bump `CONF_VERSION` below

spyderlib/plugins/__init__.py
```diff
@@ -148,9 +148,7 @@ class SpyderPluginMixin(object):
     CONFIGWIDGET_CLASS = None
     ALLOWED_AREAS = Qt.AllDockWidgetAreas
     LOCATION = Qt.LeftDockWidgetArea
-    FEATURES = QDockWidget.DockWidgetClosable | \
-               QDockWidget.DockWidgetFloatable | \
-               QDockWidget.DockWidgetMovable
+    FEATURES = QDockWidget.DockWidgetClosable | QDockWidget.DockWidgetFloatable
```
ccordoba12: Does this line fits in 79 columns?

spyderlib/plugins/configdialog.py
```diff
@@ -684,16 +684,16 @@ def setup_page(self):
...
+        verttabs_box = newcb(_("Vertical tabs in panes"),
                              'vertical_tabs')
-        animated_box = newcb(_("Animated toolbars and dockwidgets"),
+        animated_box = newcb(_("Animated toolbars and panes"),
                              'animated_docks')
         tear_off_box = newcb(_("Tear off menus"), 'tear_off_menus',
                              tip=_("Set this to detach any<br> "
                                    "menu from the main window"))
-        margin_box = newcb(_("Custom dockwidget margin:"),
+        margin_box = newcb(_("Custom margin for panes:"),
```
ccordoba12: :+1: for these changes. Thanks for doing them and making things more consistent :-)

spyderlib/spyder.py
```diff
@@ -542,20 +542,24 @@ def setup(self):
                                         context=Qt.ApplicationShortcut)
             self.register_shortcut(self.close_dockwidget_action, "_",
                                    "Close pane")
-
+            self.lock_dockwidgets_action = create_action(self, _("Lock panes"),
+                                            toggled=self.toggle_lock_dockwidgets,
+                                            context=Qt.ApplicationShortcut)
+            self.register_shortcut(self.lock_dockwidgets_action, "_",
+                                       "Lock panes")
```
ccordoba12: I think the shortcut name is missing here?

spyderlib/spyder.py
```diff
@@ -1242,6 +1250,10 @@ def post_visible_setup(self):
         self.extconsole.setMinimumHeight(0)
 
         if not self.light:
+            # Update lock status of dockidgets (panes)
+            self.lock_dockwidgets_action.setChecked(self.dockwidgets_locked)
```
ccordoba12: I think this line needs to go in the `toggle_lock_dockwidgets` method

spyderlib/spyder.py
```diff
@@ -2517,15 +2535,21 @@ def apply_settings(self):
...
             features = child.FEATURES
             if CONF.get('main', 'vertical_dockwidget_titlebars'):
-                features = features|QDockWidget.DockWidgetVerticalTitleBar
+                features = features | QDockWidget.DockWidgetVerticalTitleBar
+            if not self.dockwidgets_locked:
+                features = features | QDockWidget.DockWidgetMovable
             child.dockwidget.setFeatures(features)
             child.update_margins()
 
-        self.apply_statusbar_settings()
```
ccordoba12: Why did you remove this line?

spyderlib/spyder.py
ccordoba12: Nop, I was wrong. But I think this shortcut should lock **and** unlock panes at the same time ;-)

goanpeca: Agree.. I will rename

spyderlib/spyder.py
goanpeca: This line is the one that will make sure the action is selected according to the status when spyder was closed

spyderlib/spyder.py
goanpeca: I refactor this part because to refresh the lock status of pane I could reuse some code, just scroll down and you will se I just move it down

spyderlib/config.py
goanpeca: I was actually renaming them,

spyderlib/plugins/__init__.py
goanpeca: 79 exactly

ccordoba12: :+1:

spyderlib/config.py
ccordoba12: I know, but that's exactly the same as removing some options and adding new ones. And the removing part needs a bump in `CONF_VERSION` :-)

spyderlib/spyder.py
ccordoba12: Ok

spyderlib/spyder.py
ccordoba12: I see, but it's above this line and not below =p

spyderlib/config.py
goanpeca: Ok, will make the bump then

goanpeca: Right now I have 16.1.0 I should change it to 17.0.0 ?

ccordoba12: Yes

spyderlib/spyder.py
goanpeca: @ccordoba12 would you prefer that instead of having a check in the menu action we changed it? so if panes are locked it should display `Unlock panes` and viceversa, if the panes are currently unlocked, it should display `Lock panes`?

ccordoba12: Nop, not that much. I think a check mark is a bit more intuitive to have than a menu entry whose text is not always the same :-)

Nodd: There is alwaya problem with changing text: is the displayed text the status or the action to perform ? A checkbox is much more obvious.

goanpeca: Yes it is stelled, I just wanted to confirm ;-)

goanpeca:
Should we hide the dockwidget titlebar when locked? 

This was discussed and suggested in #2353 to improve usable space in Spyder.

Nodd:
I vote yes (I would use such a feature) but as an option and disabled by default, because locked is the default, which means that a new user would no see the titlebars at all.

goanpeca:
Sure, I would use it too, is it that problematic to make it on by default?

Nodd:
Imagine a new user, first time launching spyder an looking at the interface: there would be no indication of what all the dockwidgets are (unless there are multiple widgets in the same area, there would be tabs). I see this a a feature for advanced users who already know what each pane is for.

ccordoba12:
I just like to say: *new features = new PRs* :-)

Please keep things simple in every PR to make my life easier :-) Since the purpose of this PR is to add a lock/unlock menu entry, you should leave things up to that point.

goanpeca:
Well in that case please merge :-)

spyderlib/spyder.py
```diff
@@ -485,7 +485,7 @@ def __init__(self, options=None):
         self.is_starting_up = True
         self.is_setting_up = True
 
-        self.dockwidgets_locked = False
+        self.dockwidgets_locked = CONF.get('main', 'panes_locked', True)
```
ccordoba12: Please don't add default values to `CONF.get` (i.e. remove the `True` at the end). This messes up with updates in `CONF_VERSION`

Nodd: Maybe the argument should be removed from the fuction then, to avoid future similar problems ?

goanpeca: Oops.. that is a remnant from the first code ... :+1:  fixed

ccordoba12:
I tested this locally, and it's working as expected.

Thanks for this, great work as usual!

goanpeca:
:+1:

PR #2349, opened 2015-04-22
Add support for language selection in preferences

### LINKED ISSUE #1001: How to change UI language
_From [mentat...@gmail.com](https://code.google.com/u/116256814944482205201/) on 2012-04-21T12:22:36Z_

Hi,

I'd like to change the UI language: default from OS is French in my computer, but I want it in English. I thought it was obvious, but I can't find the settings.

Any idea?
Thanks

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=1001_

goanpeca:
Fixes #1001 

---

## Description

Add support for language selection in preferences. 

This was a bit trickier than expected, namely the configuration of the language needs to be stored in a separate file, because we cannot use CONF at the stage the `_` function is called.

I added a `langconfig` in the .`spyder` folder, which stores the language value ('en', 'fr' etc)

With this change I had to implement some new logic in the config dialog, so now comboboxes (so far) have an extra parameter call `restart`  (by default False).

## Preferences dialog

image

## Pop up message box
When the user changes one or more options flagged with `restart=True` and then clicks on Apply or Ok, the a message box will popup informing the user which options need to restart and asking if he/she wants to restart right away.

image

If the user accepts, any restart specific (defined) settings will be applied and spyder will restart automatically.

### Note:
The message box is now available for any other combobox that is created with the `restart=True` parameter. Adding this functionality to other widgets will be left for another PR.

## Todo
- [x] Display language names In original Language
- [x] Agree on additional file method and name
- [x] Define when to pop the box to the user for restart

## For another PR
- Add the restart functionality to all the other widgets besides comboboxes

### DIFF
--- a/spyderlib/baseconfig.py
+++ b/spyderlib/baseconfig.py
@@ -205,10 +205,11 @@ def get_image_path(name, default="not_found.png"):
 #==============================================================================
 # Translations
 #==============================================================================
-LANG_FILE = osp.join(get_conf_path(), 'langconfig')
+LANG_FILE = get_conf_path('langconfig')
 DEFAULT_LANGUAGE = 'en'
 
-# This needs to be updated every time a new language is added to spyder.
+# This needs to be updated every time a new language is added to spyder, and is
+# used by the Preferences configuration to populate the Language QComboBox
 LANGUAGE_CODES = {'en': u'English',
                   'fr': u'Français',
                   'es': u'Español',
@@ -219,7 +220,9 @@ LANGUAGE_CODES = {'en': u'English',
 def get_available_translations():
     """
     List available translations for spyder based on the folders found in the
-    locale folder.
+    locale folder. This function checks if LANGUAGE_CODES contain the same
+    information that is found in the 'locale' folder to ensure that when a new
+    language is added, LANGUAGE_CODES is updated.
     """
     locale_path = get_module_data_path("spyderlib", relpath="locale",
                                        attr_name='LOCALEPATH')
--- a/spyderlib/baseconfig.py
+++ b/spyderlib/baseconfig.py
@@ -256,13 +256,13 @@ def get_interface_language():
         language = DEFAULT_LANGUAGE
     else:
         spyder_languages = get_available_translations()
-        for spyder_language in spyder_languages:
-            if locale_language == spyder_language:
+        for lang in spyder_languages:
+            if locale_language == lang:
                 language = locale_language
                 break
-            elif locale_language.startswith(spyder_language) or\
-                    spyder_language.startswith(locale_language):
-                language = spyder_language
+            elif locale_language.startswith(lang) or \
+              lang.startswith(locale_language):
+                language = lang
                 break
 
     return language
@@ -298,8 +298,8 @@ def get_translation(modname, dirname=None):
 
     # fixup environment var LANG, LANGUAGE
     language = load_lang_conf()
-    os.environ["LANG"] = language      # Works in windows
-    os.environ["LANGUAGE"] = language  # Works in ubuntu
+    os.environ["LANG"] = language      # Works on Windows
+    os.environ["LANGUAGE"] = language  # Works on Linux
 
     import gettext
     try:
--- a/spyderlib/baseconfig.py
+++ b/spyderlib/baseconfig.py
@@ -14,6 +14,7 @@ sip API incompatibility issue in spyderlib's non-gui modules)
 
 from __future__ import print_function
 
+import locale
 import os.path as osp
 import os
 import sys
@@ -204,18 +205,102 @@ def get_image_path(name, default="not_found.png"):
 #==============================================================================
 # Translations
 #==============================================================================
+LANG_FILE = osp.join(get_conf_path(), '.langconfig')
+DEFAULT_LANGUAGE = 'en'
+
+# This needs to be updated every time a new language is added to spyder.
+LANGUAGE_CODES = {'en': u'English',
+                  'fr': u'Français',
+                  'es': u'Español',
+                  'pt_BR': u'Português (brasileiro)'
+                  }
+
+
+def get_available_translations():
+    """
+    List available translations for spyder based on the folders found in the
+    locale folder.
+    """
+    locale_path = get_module_data_path("spyderlib", relpath="locale",
+                                       attr_name='LOCALEPATH')
+    listdir = os.listdir(locale_path)
+    langs = [d for d in listdir if osp.isdir(osp.join(locale_path, d))]
+    langs = [DEFAULT_LANGUAGE] + langs
+
+    # Check that there is a language code available in case a new translation
+    # is added, to ensure LANGUAGE_CODES is updated.
+    for lang in langs:
+        if lang not in LANGUAGE_CODES:
+            error = _('Update LANGUAGE_CODES if a new translation has been '
+                      'added to Spyder')
+            raise Exception(error)
+    return langs
+
+
+def get_interface_language():
+    """
+    If Spyder has a translation available for the locale language, it will
+    return the version provided by Spyder adjusted for language subdifferences,
+    otherwise it will return DEFAULT_LANGUAGE.
+
+    Example:
+    1.) Spyder provides ('en',  'fr', 'es' and 'pt_BR'), if the locale is
+    either 'en_US' or 'en' or 'en_UK', this function will return 'en'
+
+    2.) Spyder provides ('en',  'fr', 'es' and 'pt_BR'), if the locale is
+    either 'pt' or 'pt_BR', this function will return 'pt_BR'
+    """
+    locale_language = locale.getdefaultlocale()[0]
+
+    if locale_language is None:
+        language = DEFAULT_LANGUAGE
+    else:
+        spyder_languages = get_available_translations()
+        for spyder_language in spyder_languages:
+            if locale_language == spyder_language:
+                language = locale_language
+                break
+            elif locale_language.startswith(spyder_language) or\
+                    spyder_language.startswith(locale_language):
+                language = spyder_language
+                break
+
+    return language
+
+
+def save_lang_conf(value):
+    """Save language setting to language config file"""
+    with open(LANG_FILE, 'w') as f:
+        f.write(value)
+
+
+def load_lang_conf():
+    """
+    Load language setting from language config file if it exists, otherwise
+    try to use the local settings if Spyder provides a translation, or
+    return the default if no translation provided.
+    """
+    if osp.isfile(LANG_FILE):
+        with open(LANG_FILE, 'r') as f:
+            lang = f.read()
+    else:
+        lang = get_interface_language()
+        save_lang_conf(lang)
+    return lang
+
+
 def get_translation(modname, dirname=None):
     """Return translation callback for module *modname*"""
     if dirname is None:
         dirname = modname
     locale_path = get_module_data_path(dirname, relpath="locale",
                                        attr_name='LOCALEPATH')
-    # fixup environment var LANG in case it's unknown
-    if "LANG" not in os.environ:
-        import locale
-        lang = locale.getdefaultlocale()[0]
-        if lang is not None:
-            os.environ["LANG"] = lang
+
+    # fixup environment var LANG, LANGUAGE
+    language = load_lang_conf()
+    os.environ["LANG"] = language      # Works in windows
+    os.environ["LANGUAGE"] = language  # Works in ubuntu
+
     import gettext
     try:
         _trans = gettext.translation(modname, locale_path, codeset="utf-8")
--- a/spyderlib/baseconfig.py
+++ b/spyderlib/baseconfig.py
@@ -205,14 +205,14 @@ def get_image_path(name, default="not_found.png"):
 #==============================================================================
 # Translations
 #==============================================================================
-LANG_FILE = osp.join(get_conf_path(), '.langconfig')
+LANG_FILE = osp.join(get_conf_path(), 'langconfig')
 DEFAULT_LANGUAGE = 'en'
 
 # This needs to be updated every time a new language is added to spyder.
 LANGUAGE_CODES = {'en': u'English',
                   'fr': u'Français',
                   'es': u'Español',
-                  'pt_BR': u'Português (brasileiro)'
+                  'pt_BR': u'Português'
                   }
 
 
@@ -231,8 +231,8 @@ def get_available_translations():
     # is added, to ensure LANGUAGE_CODES is updated.
     for lang in langs:
         if lang not in LANGUAGE_CODES:
-            error = _('Update LANGUAGE_CODES if a new translation has been '
-                      'added to Spyder')
+            error = _('Update LANGUAGE_CODES (inside baseconfig.py) if a new '
+                      'translation has been added to Spyder')
             raise Exception(error)
     return langs
 
--- a/spyderlib/baseconfig.py
+++ b/spyderlib/baseconfig.py
@@ -14,6 +14,7 @@ sip API incompatibility issue in spyderlib's non-gui modules)
 
 from __future__ import print_function
 
+import codecs
 import locale
 import os.path as osp
 import os
@@ -40,7 +41,8 @@ TEST = os.environ.get('SPYDER_TEST')
 #==============================================================================
 # Debug helpers
 #==============================================================================
-STDOUT = sys.stdout
+# This is needed after restarting and using debug_print
+STDOUT = sys.stdout if PY3 else codecs.getwriter('utf-8')(sys.stdout)
 STDERR = sys.stderr
 def _get_debug_env():
     debug_env = os.environ.get('SPYDER_DEBUG', '')
@@ -53,7 +55,13 @@ def debug_print(*message):
     """Output debug messages to stdout"""
     if DEBUG:
         ss = STDOUT
-        print(*message, file=ss)
+        if PY3:
+            # This is needed after restarting and using debug_print
+            for m in message:
+                ss.buffer.write(str(m).encode('utf-8'))
+            print('', file=ss)
+        else:
+            print(*message, file=ss)
 
 #==============================================================================
 # Configuration paths
--- a/spyderlib/config.py
+++ b/spyderlib/config.py
@@ -169,7 +169,6 @@ DEFAULTS = [
               'animated_docks': True,
               'prompt_on_exit': False,
               'panes_locked': True,
-              'interface_language': load_lang_conf(),
               'window/size': (1260, 740),
               'window/position': (10, 10),
               'window/is_maximized': True,
@@ -724,6 +723,9 @@ CONF_VERSION = '17.0.0'
 # Check if it *really* needs to be updated or not
 CONF = UserConfig('spyder', defaults=DEFAULTS, load=True, version=CONF_VERSION,
                   subfolder=SUBFOLDER, backup=True, raw_mode=True)
+
+# Ensures that the config is present on spyder first run
+CONF.set('main', 'interface_language', load_lang_conf())
 
 # Removing old .spyder.ini location:
 old_location = osp.join(get_home_dir(), '.spyder.ini')
--- a/spyderlib/config.py
+++ b/spyderlib/config.py
@@ -18,7 +18,7 @@ import os.path as osp
 # Local import
 from spyderlib.userconfig import UserConfig
 from spyderlib.baseconfig import (CHECK_ALL, EXCLUDED_NAMES, SUBFOLDER,
-                                  get_home_dir, _, load_lang_conf)
+                                  get_home_dir, _)
 from spyderlib.utils import iofuncs, codeanalysis
 
 
@@ -723,9 +723,6 @@ CONF_VERSION = '17.0.0'
 # Check if it *really* needs to be updated or not
 CONF = UserConfig('spyder', defaults=DEFAULTS, load=True, version=CONF_VERSION,
                   subfolder=SUBFOLDER, backup=True, raw_mode=True)
-
-# Ensures that the config is present on spyder first run
-CONF.set('main', 'interface_language', load_lang_conf())
 
 # Removing old .spyder.ini location:
 old_location = osp.join(get_home_dir(), '.spyder.ini')
--- a/spyderlib/config.py
+++ b/spyderlib/config.py
@@ -18,7 +18,7 @@ import os.path as osp
 # Local import
 from spyderlib.userconfig import UserConfig
 from spyderlib.baseconfig import (CHECK_ALL, EXCLUDED_NAMES, SUBFOLDER,
-                                  get_home_dir, _)
+                                  get_home_dir, _, load_lang_conf)
 from spyderlib.utils import iofuncs, codeanalysis
 
 
@@ -169,6 +169,7 @@ DEFAULTS = [
               'animated_docks': True,
               'prompt_on_exit': False,
               'panes_locked': True,
+              'interface_language': load_lang_conf(),
               'window/size': (1260, 740),
               'window/position': (10, 10),
               'window/is_maximized': True,
--- a/spyderlib/plugins/configdialog.py
+++ b/spyderlib/plugins/configdialog.py
@@ -22,7 +22,7 @@ from spyderlib.qt.compat import (to_qvariant, from_qvariant,
                                  getexistingdirectory, getopenfilename)
 
 from spyderlib.baseconfig import (_, running_in_mac_app, LANGUAGE_CODES,
-                                  get_available_translations, save_lang_conf)
+                                  save_lang_conf)
 from spyderlib.config import CONF
 from spyderlib.guiconfig import (CUSTOM_COLOR_SCHEME_NAME,
                                  set_default_color_scheme)
@@ -91,7 +91,6 @@ class ConfigPage(QWidget):
             self.save_to_conf()
             if self.apply_callback is not None:
                 self.apply_callback()
-            self.set_modified(False)
 
             # Since the language cannot be retrieved by CONF and the language
             # is needed before loading CONF, this is an extra method needed to
@@ -104,6 +103,7 @@ class ConfigPage(QWidget):
                 if restart_option in self.changed_options:
                     self.prompt_restart_required()
                     break  # Ensure a single popup is displayed
+            self.set_modified(False)
 
     def load_from_conf(self):
         """Load settings from configuration file"""
@@ -753,8 +753,7 @@ class MainConfigPage(GeneralConfigPage):
         prompt_box = newcb(_("Prompt when exiting"), 'prompt_on_exit')
 
         # Language chooser
-        langs = get_available_translations()
-        choices = list(zip([LANGUAGE_CODES[key] for key in langs], langs))
+        choices = sorted([(val, key) for key, val in LANGUAGE_CODES.items()])
         language_combo = self.create_combobox(_('Language'), choices,
                                               'interface_language',
                                               restart=True)
--- a/spyderlib/plugins/configdialog.py
+++ b/spyderlib/plugins/configdialog.py
@@ -91,6 +91,8 @@ class ConfigPage(QWidget):
             self.save_to_conf()
             if self.apply_callback is not None:
                 self.apply_callback()
+            self.set_modified(False)
+
             # Since the language cannot be retrieved by CONF and the language
             # is needed before loading CONF, this is an extra method needed to
             # ensure that when changes are applied, they are copied to a
@@ -101,10 +103,8 @@ class ConfigPage(QWidget):
             for restart_option in self.restart_options:
                 if restart_option in self.changed_options:
                     self.prompt_restart_required()
-                    break
+                    break  # Ensure a single popup is displayed
 
-            self.set_modified(False)
-    
     def load_from_conf(self):
         """Load settings from configuration file"""
         raise NotImplementedError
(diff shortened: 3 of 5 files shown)


spyderlib/baseconfig.py
```diff
@@ -204,18 +205,102 @@ def get_image_path(name, default="not_found.png"):
 #==============================================================================
 # Translations
 #==============================================================================
+LANG_FILE = osp.join(get_conf_path(), '.langconfig')
```
ccordoba12: I'm ok with adding this new file, but don't call it `.langconfig`, just `langconfig`, so that it's not hidden on Linux :-)

At the end, it's going to be inside of `~/.spyder`, so there is no need to be hidden.

spyderlib/baseconfig.py
```diff
@@ -204,18 +205,102 @@ def get_image_path(name, default="not_found.png"):
...
 # Translations
 #==============================================================================
+LANG_FILE = osp.join(get_conf_path(), '.langconfig')
+DEFAULT_LANGUAGE = 'en'
+
+# This needs to be updated every time a new language is added to spyder.
+LANGUAGE_CODES = {'en': u'English',
+                  'fr': u'Français',
+                  'es': u'Español',
+                  'pt_BR': u'Português (brasileiro)'
```
ccordoba12: Is Brazilian Portuguese too different from Portuguese? If not, I'd vote to just leave Portuguese here

goanpeca: We should ask the guy that did the translation. Can you add him to the thread?

spyderlib/baseconfig.py
goanpeca: :+1:

spyderlib/baseconfig.py
goanpeca: But i do prefer to put just portuguese...

spyderlib/plugins/configdialog.py
```diff
@@ -192,6 +199,7 @@ def current_page_changed(self, index):
         self.apply_btn.setEnabled(widget.is_modified)
         
     def add_page(self, widget):
+        setattr(widget, 'parent_config_dialog', self)
```
goanpeca: Oops i do no longer need this. Will remove it.

spyderlib/baseconfig.py
```diff
@@ -204,18 +205,102 @@ def get_image_path(name, default="not_found.png"):
...
+                  }
+
+
+def get_available_translations():
```
ccordoba12: I don't think we need a function for this ;-)

There are so few translations (and none on the works), that we can just create the list of `langs` by hand here.

spyderlib/baseconfig.py
```diff
@@ -204,18 +205,102 @@ def get_image_path(name, default="not_found.png"):
...
+    # is added, to ensure LANGUAGE_CODES is updated.
+    for lang in langs:
+        if lang not in LANGUAGE_CODES:
+            error = _('Update LANGUAGE_CODES if a new translation has been '
+                      'added to Spyder')
+            raise Exception(error)
+    return langs
+
+
+def get_interface_language():
```
ccordoba12: It also seems like too much to have this function

spyderlib/utils/iofuncs.py
```diff
@@ -362,7 +362,8 @@ def load_dictionary(filename):
                       'spyder.ini', 'temp.py', 'temp.spydata', 'template.py',
                       'history.py', 'history_internal.py', 'workingdir',
                       '.projects', '.spyderproject', '.ropeproject',
-                      'monitor.log', 'monitor_debug.log', 'rope.log')
+                      'monitor.log', 'monitor_debug.log', 'rope.log',
+                      '.langconfig')
```
ccordoba12: `.langconfig` -> `langconfig`

goanpeca: Sure

spyderlib/baseconfig.py
goanpeca: Yes it might be too much, but why doing stuff by hand, if a function can handle it?

I would really like to have many more translations

spyderlib/baseconfig.py
goanpeca: Probably we do not need it now, but it is a harmless function, and it provides a check when adding a new language to make sure things are done properly. I think we should leave it

ccordoba12: Ok, you're right :-)

If you want more translations, you should look at https://www.transifex.com/

I don't know exactly how it works, but I think we need to upload our pot files and volunteers help to translate them to their native languages :-)

spyderlib/baseconfig.py
ccordoba12: Ok, now that it's written, we should leave it  :-)

spyderlib/baseconfig.py
goanpeca: I will take a look a transifex, I had seen it before, but never got the time to go in depth

goanpeca:
@ccordoba12 I updated with the fixes, when you have some time can you take a look at some questions I left in the description?

goanpeca:
Can @spyder-ide/core-developers give this a try and give some comments?

```bash
git clone https://github.com/goanpeca/spyder.git spyder-lang
cd spyder-lang
git checkout -b language
git pull origin language
python bootstrap.py
```

goanpeca:
This PR is waiting for some input from the @spyder-ide/core-developers ...

There are 2 ways of doing this.
1.) Ask the user if he wants to restart for the setting to take place, immediately after changing the setting
2.) Ask the user if he wants to restart for the setting to take place, only after pressing Ok or Apply

I implemented the first one.. but the second one could be also implemented if you consider is better.

@ccordoba12 any ideas, thoughts?

Nodd:
I vote 2 because:
* there could be other settings which ask to reload (changing the icon theme, enabling plugins…)
* It's annoying to have a popup without being able to check if the change was correct.

If 2 is chosen, it should be written that it requires a restart.

goanpeca:
In that case of 2, I would make the popup appear at the end when clicking `ok`, or `apply`

Nodd:
Yep. Bonus points if the dialog contains the reason why the restart is needed (as a list so that other settings can give their information too)

goanpeca:
Yep, you read my mind ;-)

goanpeca:
@Nodd ok, this is now updated to ask only after pressing Ok or Apply and displays a list of all the options that require a restart (and were changed).

Can you give it a try @ccordoba12? 

I would expect this to work in mac without changes but, @blink1073 can you give it a try?

ccordoba12:
The restart popup message is very nice!! It looks very professional! Thanks a lot for working on it @goanpeca!

spyderlib/baseconfig.py
```diff
@@ -204,18 +205,102 @@ def get_image_path(name, default="not_found.png"):
...
+    2.) Spyder provides ('en',  'fr', 'es' and 'pt_BR'), if the locale is
+    either 'pt' or 'pt_BR', this function will return 'pt_BR'
+    """
+    locale_language = locale.getdefaultlocale()[0]
+
+    if locale_language is None:
+        language = DEFAULT_LANGUAGE
+    else:
+        spyder_languages = get_available_translations()
+        for spyder_language in spyder_languages:
```
ccordoba12: Let's simplify the variables here to make it a bit easier to read. What about this?

```python
for lang in spyder_languages
```

goanpeca:
Glad you like it, if this works on Mac, then it should be ready to merge

spyderlib/baseconfig.py
```diff
@@ -204,18 +205,102 @@ def get_image_path(name, default="not_found.png"):
...
+            if locale_language == spyder_language:
+                language = locale_language
+                break
+            elif locale_language.startswith(spyder_language) or\
```
ccordoba12: Missing space: `or\` -> `or \`

spyderlib/baseconfig.py
goanpeca: Ok sure, I will adapt ;-)

spyderlib/baseconfig.py
goanpeca: will fix, but probably will be one line with your suggestion

spyderlib/baseconfig.py
```diff
@@ -204,18 +205,102 @@ def get_image_path(name, default="not_found.png"):
...
+                    spyder_language.startswith(locale_language):
```
ccordoba12: 2 indentation spaces here, not 8, please (I know PEP8, but I don't like it)

spyderlib/baseconfig.py
```diff
@@ -204,18 +205,102 @@ def get_image_path(name, default="not_found.png"):
...
-    # fixup environment var LANG in case it's unknown
-    if "LANG" not in os.environ:
-        import locale
-        lang = locale.getdefaultlocale()[0]
-        if lang is not None:
-            os.environ["LANG"] = lang
+
+    # fixup environment var LANG, LANGUAGE
+    language = load_lang_conf()
+    os.environ["LANG"] = language      # Works in windows
```
ccordoba12: `works in windows` -> `Works on Windows`

The same for Ubuntu below, but changing it to Linux :-)

goanpeca: Is this going to work on Mac for sure?

ccordoba12: Translations on Mac never have worked because Mac doesn't define proper `locale` values.

This work will let users select the translation they want to use :-)

spyderlib/plugins/configdialog.py
```diff
@@ -90,6 +91,18 @@ def apply_changes(self):
...
+            # is needed before loading CONF, this is an extra method needed to
+            # ensure that when changes are applied, they are copied to a
+            # specific file storing the language value.
+            if getattr(self, 'save_lang', False):
+                self.save_lang()
+
+            for restart_option in self.restart_options:
+                if restart_option in self.changed_options:
+                    self.prompt_restart_required()
+                    break
```
ccordoba12: Why this `break` here?

goanpeca: I guess that I am under the impression that if I do not break it here, you will get a popup for every option that the user changed (and requires restart)

ccordoba12: I see, it's fine then :-)

spyderlib/plugins/configdialog.py
```diff
@@ -661,13 +680,42 @@ def get_icon(self):
     def apply_settings(self, options):
         raise NotImplementedError
 
+    def prompt_restart_required(self):
+        """Prompt the user with a request to restart."""
+        restart_opt = self.restart_options
+        changed_opt = self.changed_options
```
ccordoba12: `restart_opt` -> `restart_opts` (because it's plural)
`changed_opt` -> `changed_opts`

goanpeca: fixed

spyderlib/plugins/configdialog.py
```diff
@@ -704,6 +752,13 @@ def setup_page(self):
         margins_layout.addWidget(margin_spin)
         prompt_box = newcb(_("Prompt when exiting"), 'prompt_on_exit')
 
+        # Language chooser
+        langs = get_available_translations()
+        choices = list(zip([LANGUAGE_CODES[key] for key in langs], langs))
```
ccordoba12: What this `choices` variable hold? I mean, is it something like

```python
[('en': 'English', 'es': 'Spanish')]
```

?

goanpeca: I copied this from code above (the choices part and adapted it...) 

It holds what is stored in `LANGUAGE_CODES` in the `baseconfig.py` 

So the combo box display Español, English... etc... and the value is ...es, en... etc..

goanpeca: So actually it is more like 

`choices = [(English, 'en'), ('Español', 'es'), ...]`

spyderlib/baseconfig.py
```diff
@@ -204,18 +205,102 @@ def get_image_path(name, default="not_found.png"):
 #==============================================================================
 # Translations
 #==============================================================================
+LANG_FILE = osp.join(get_conf_path(), 'langconfig')
```
ccordoba12: Please rewrite this line as

```python
LANG_FILE = get_conf_path('langconfig')
```

That's the way `get_conf_path` works :-)

goanpeca: Done

spyderlib/plugins/configdialog.py
ccordoba12: This seems a bit too complex to my taste (I mean, the difference between `LANGUAGE_CODES` and `get_available_translations`).

What will happen when someone adds `pt_PT` or `en_UK`?

ccordoba12:
Everything seems fine to me now, but:

1. I still find the implementation too complex to follow.
2. We still need to test this PR on Mac :-)

spyderlib/plugins/configdialog.py
goanpeca: When someone adds this where?

goanpeca:
Regarding 1.) hmmm

I tried to add all the needed documentation to understand. I will explain and you tell me where you find it confusing.

When someone wants to add a new translation they will create the folder in `locale` and put there the new translation, or variation of a translation.

Once they do this and run the code they will get an error because we need to have an updated language name (in native language) added to `LANGUAGE_CODES`. If they added a new language, no problem they just add i.e. `'de': 'Deutsch'`... if it is a variation they just add i.e. `'pt_PT': 'Portuguese (Portugal)'` and make sure the other variation (if existent) should be then `'pt_BR': 'Portuguese (Brazil)'`

But the truth is, that the second case will most likely never happen...

spyderlib/plugins/configdialog.py
goanpeca: I could just use the `LANGUAGE_CODES` variable and remove the call to `langs = get_available_translations()`

spyderlib/plugins/configdialog.py
```diff
@@ -704,6 +752,12 @@ def setup_page(self):
         margins_layout.addWidget(margin_spin)
         prompt_box = newcb(_("Prompt when exiting"), 'prompt_on_exit')
 
+        # Language chooser
+        choices = sorted([(val, key) for key, val in LANGUAGE_CODES.items()])
+        language_combo = self.create_combobox(_('Language'), choices,
+                                              'interface_language',
+                                              restart=True)
+
```
goanpeca: @ccordoba12 this should be clearer now

ccordoba12: Yep, much better, thanks! Now I don't have objecttions.

goanpeca: :+1:

ccordoba12:
I'll merge this tomorrow, after testing on Mac :-)

goanpeca:
I will have to buy a Mac... ;-)

blink1073:
I can test today on my Mac if you prefer @ccordoba12.

ccordoba12:
That'd be great too @blink1073! I thought you weren't around ;-)

blink1073:
Was at the beach with the little ones.  ;-)

blink1073:
Una problema:

```python
Cargando el inspector de objetos...
  File "/Users/silvester/workspace/spyder/spyderlib/spyder.py", line 3077, in main
  File "/Users/silvester/workspace/spyder/spyderlib/spyder.py", line 2963, in run_spyder
    main.setup()
  File "/Users/silvester/workspace/spyder/spyderlib/spyder.py", line 851, in setup
    self.set_splash(_("Loading outline explorer..."))
  File "/Users/silvester/workspace/spyder/spyderlib/spyder.py", line 2095, in set_splash
    self.debug_print(message)
  File "/Users/silvester/workspace/spyder/spyderlib/spyder.py", line 532, in debug_print
    debug_print(message)
  File "/Users/silvester/workspace/spyder/spyderlib/baseconfig.py", line 56, in debug_print
    print(*message, file=ss)
UnicodeEncodeError: 'ascii' codec can't encode character '\xf3' in position 27: ordinal not in range(128)
```

blink1073:
It looks like `sys.stdout` is encoded as `ascii`, thanks Apple...

goanpeca:
Any suggestions?

blink1073:
Looking into it now.

blink1073:
Hmm, this works fine from an IPython prompt:  `print('€', file=sys.stdout)`

blink1073:
It might have something to do with the restart itself affecting `sys.stdout`.

goanpeca:
If you say No when the popup box asks to restart... and then you close spyder...
and rerun from bootstrap...?

You get the same error?

blink1073:
If I restart after the failure itself it comes up fine.

goanpeca:
If you run without DEBUG ?

blink1073:
Actually if I say NO then it restarts anyway...

goanpeca:
hmm thats a bug then.. it should do that!

blink1073:
Oops, forget that last comment.

goanpeca:
Can you run without Debug? and check again?

blink1073:
Running without debug is fine, and I can print unicode characters from the Spyder Python console.

blink1073:
It appears to just be the restart while in debug mode.

blink1073:
I tried using `sys.stdout` within `debug_print`, instead of the global variable, but it still did not work.

goanpeca:
Can you try this...

Change baseconfig.py  line 52 with this

```python
def debug_print(*message):
    """Output debug messages to stdout"""
    if DEBUG:
        ss = STDOUT
        message = tuple([m.encode('utf-8') for m in message])
        print(*message, file=ss)
```

And try with debug again?

blink1073:
That works but all of the strings print with `b''`

blink1073:
E.g.: `b'Configuration de la fen\xc3\xaatre principale...'`

goanpeca:
and if you do this

```python
def debug_print(*message):
    """Output debug messages to stdout"""
    if DEBUG:
        ss = STDOUT
        message = tuple([to_text_string(m) for m in message])
        print(*message, file=ss)
```

blink1073:
Nope, same UnicodeEncodeError.

goanpeca:
You are on Py3?

blink1073:
Yep, I'll try Py2.

blink1073:
Same problem there.

blink1073:
If I turn DEBUG off, do the restart, and then try and print from with the Spyder Internal Console, I get the same error too.  The other consoles are note affected.

goanpeca:
yes... the internal console seems to be the problem...

But that is something unrelated to the changes I did, seems to be a problem that has been sleeping for some time. I am trying some options

blink1073:
Yes, it appears to be related to the restart itself.

blink1073:
Yup, confirmed.  If I start in English mode and do just a restart, we get this behavior.

goanpeca:
I will try linux.. one sec

goanpeca:
Ok, I can reproduce it, ...fixing!! give me 10 mins

goanpeca:
@blink1073 

Can you add this line to `spyderlib\spyder.py` after line 2737

`env['PYTHONIOENCODING'] = 'utf-8'`

blink1073:
Perfecto!

goanpeca:
both py2 and py3?

blink1073:
Oui.

blink1073:
`debug_print` is super nice, I use it all the time.

Nodd:
This env variable will change the behavior in scripts run from spyder ? This is bad, we had many bug reports because of this. We should change the encoding of the internal console only. I think I have a solution somewhere, I'll try to find it.

goanpeca:
@Nodd I do not think it will be the case... but you can try with an example?

goanpeca:
I can erase the environment variable in the `restart_app.py` just to be sure... so it is only used for the python process that spawns the new Spyder instance.

goanpeca:
Maybe I can pop that variable in `spyderlib\spyder.py` so that it is not there after startup. It is needed when spawning the new Spyder process on restart, but should not be needed afterwards.

blink1073:
But what if the user had explicitly set that variable to something else?  Perhaps only set it if it is not in `os.environ` and not set to something else, then leave it there.

blink1073:
`utf-8` all the things!

goanpeca:
Well, if the user set it to soemthing else, the spyder will break...

blink1073:
Then I say set it and leave it.

goanpeca:
Lets see what @Nodd brings :-)

SylvainCorlay:
I am good with this. Will use that dialog on restarting spyder for the icons as soon as this is merged.

goanpeca:
@Nodd did you have some time to find what you were looking for? [or](https://www.youtube.com/watch?v=EabB55IYyOg)

@ccordoba12  any comments on the fix needed to get this PR working properly, specifically [this last commit](https://github.com/goanpeca/spyder/commit/5bae10be0773ab2a3b3a2a943606ab49e4d69d96)

Nodd:
I used this in some scritps so that `argparse` help doesn't raise a `UnicodeSomethingError` when displaying a help message with utf8 symbols under python2:
```
import codecs
sys.stdout = codecs.getwriter('utf-8')(sys.stdout)
```
Since the internal console is basically `sys.stdout`, it might be used as-is.

goanpeca:
and in python 3 ?

goanpeca:
where should I put this code?

Nodd:
In python3 I didn't have this problem (the unicode support is way better).
It should be put at the same place where sys.stdout is redirected to the internal console, I guess. I have no idea where it is :disappointed:

spyderlib/spyder.py
```diff
@@ -2743,6 +2743,9 @@ def restart(self):
             startupinfo.dwFlags |= subprocess.STARTF_USESHOWWINDOW
             shell = False
         else:
+            # This environment variable is needed so that the new python
+            # process spawned with Popen sets this as the default encoding
+            env['PYTHONIOENCODING'] = 'utf-8'
```
ccordoba12: I think setting this system-wide is not a good idea. For example, It could mess with Pandas trying to import/export files :-)

ccordoba12:
If I'm not wrong, redirection of stdio for the internal console happens in `redirect_internalshell_stdio`, currently present in

https://github.com/spyder-ide/spyder/blob/master/spyderlib/spyder.py#L2435

goanpeca:
I tried putting @Nodd  where @ccordoba12 without success

goanpeca:
@Nodd I think I have it fixed, can you check the latest commit?

@blink1073, or @ccordoba12  can you please test on Mac? remember to run using 

`python bootstrap.py --debug`

blink1073:
It works for me, other than the fact that the restart dialog was entirely in English.

blink1073:
screen shot 2015-05-07 at 8 26 03 pm

goanpeca:
Cause we need To rerun the localización scripts :+1:

goanpeca:
@ccordoba12 can we merge this one so that @SylvainCorlay can add this to his icon PR :smile: ?

SylvainCorlay:
:+1:

SylvainCorlay:
Yeah seems I will need to rebase again anyway!

spyderlib/config.py
```diff
@@ -169,6 +169,7 @@ def is_ubuntu():
               'animated_docks': True,
               'prompt_on_exit': False,
               'panes_locked': True,
+              'interface_language': load_lang_conf(),
```
ccordoba12: I think we don't need to add this here because this is dependent on the user system.

The only options that we add in `config.py` are those that can be changed and/or updated in the future (by bumping `CONF_VERSION`). That's why (for example) we don't add here the window size or state.

ccordoba12: Another example of things we don't add are the run configuration options per file :-)

spyderlib/plugins/configdialog.py
```diff
@@ -90,8 +91,20 @@ def apply_changes(self):
             self.save_to_conf()
             if self.apply_callback is not None:
                 self.apply_callback()
+
+            # Since the language cannot be retrieved by CONF and the language
+            # is needed before loading CONF, this is an extra method needed to
+            # ensure that when changes are applied, they are copied to a
+            # specific file storing the language value.
+            if getattr(self, 'save_lang', False):
```
ccordoba12: Why is this line necessary?

goanpeca: Cause this class is inherited by all config panes. But only the main will for sure have the save_lang method.

spyderlib/config.py
ccordoba12: Please remove this option from here to finally merge this PR :-)

If I understand the code correctly:

1. It'll be saved to `spyder.ini` by the `save_lang` method of `MainConfigPage`
2. It can't be retrieved by `CONF` directly.

So there's no need to have it here.

goanpeca: @ccordoba12 

If I don't add that method to the config, Spyder will not be able to load the preferences after a reset, cause there will be no `interface_language` setting in the .ini file.

Suggestions?

goanpeca: I think we should leave it as it is, otherwise we would have to made unneded changes either to how the preferences are loaded or do an extra check in spyder.y to see if this is the first time it runs.... here it is just one line of code...

spyderlib/plugins/configdialog.py
Nodd: how about changing the name to `__save_lang` ? this way it will be "hidden" to the child class (external plugins included), but the name will be changed at runtime (the class name is added).

goanpeca: sure

SylvainCorlay:
Is there any blocking point for this one?

spyderlib/config.py
```diff
@@ -724,6 +724,9 @@ def is_ubuntu():
 CONF = UserConfig('spyder', defaults=DEFAULTS, load=True, version=CONF_VERSION,
                   subfolder=SUBFOLDER, backup=True, raw_mode=True)
 
+# Ensures that the config is present on spyder first run
+CONF.set('main', 'interface_language', load_lang_conf())
```
goanpeca: @ccordoba12 based on your previous comment I need to add this line here to make sure that on first run of spyder the `interface_language` entry is created. On subsequent starts of spyder that line is harmless so this way is not part of the CONF but ensures the correct behavior. 

You think is ok like this?

goanpeca:
@ccordoba12 this one should be ready to merge as well

SylvainCorlay:
:+1:

spyderlib/config.py
ccordoba12: Yes, I think it's fine :-) But I think it's better to move it `configdialog.py`. Well, if it's possible because there is where this setting is really needed.

spyderlib/plugins/configdialog.py
```diff
@@ -90,8 +91,21 @@ def apply_changes(self):
             self.save_to_conf()
             if self.apply_callback is not None:
                 self.apply_callback()
+
+            # Since the language cannot be retrieved by CONF and the language
+            # is needed before loading CONF, this is an extra method needed to
+            # ensure that when changes are applied, they are copied to a
+            # specific file storing the language value. This only applies to
+            # the main section config.
+            if self.CONF_SECTION == u'main':
```
ccordoba12: I don't understand this change here. Maybe it's better to just move the `CONF.set` line from `config.py` here to avoid this.

ccordoba12: I mean, what I don't understand is why you're looking for `main`.

goanpeca: `main` is the section in charge of setting the interface language, so the `_save_lang` only belongs to that section, before I was using  `if getattr(self, "_save_lang", False):` but I change it so that it was more explicit.

Since this method runs for every child (any other section or plugin) then I need this to avoid an error of calling _save_lang when it does not exist.

spyderlib/config.py
goanpeca: Hmmm ok I will take a look if I find a suitable place there

spyderlib/plugins/configdialog.py
goanpeca: The other way would be to make the save_lang method part of one of the ancestors so that EVERY section would alwyas make that save... but since it is only necessary for main...

ccordoba12: Too much work :-) Then leave this part as it is now.

Please see if you can move the `CONF.set` call from `config.py` to here so that we can (finally!) merge this PR ;-)

goanpeca: Will do! ;-)

spyderlib/plugins/configdialog.py
```diff
@@ -151,6 +165,9 @@ def __init__(self, parent=None):
 
         self.setWindowTitle(_("Preferences"))
         self.setWindowIcon(get_icon("configure.png"))
+
+        # Ensures that the config is present on spyder first run
+        CONF.set('main', 'interface_language', load_lang_conf())
```
goanpeca: Moved here, right at the end of the init :-), should work fine now

goanpeca: @ccordoba12

goanpeca:
Merging :-) @ccordoba12?

ccordoba12:
Yes!! I'm having a problem with Qt5 but it seems to be with my system rather than with this PR (because I'm also having problems in master :-)

ccordoba12:
Thanks for your patience by the way ;-)

goanpeca:
No prob ;-)

Ok, @SylvainCorlay, now is your turn :-)  

Just add `restart=True` to the combobox, i.e.

```python
        language_combo = self.create_combobox(_('Language'), choices,
                                              'interface_language',
                                              restart=True)
```

PR #2357, opened 2015-04-24
Fix some completion issues on the Editor (issue #2352)

### LINKED ISSUE #2352: Some issues with code completion in the Editor
Last night I was testing things after PR #2337 and I found these problems (sorted by importance, from very serious to less so):

1. If you try to get completions of something that can't give none (e.g. `XX<tab>`), the completion machinery **stops** to work, i.e. you can't get completions on the current file, nor on any other (new or already open). This is also true for any file (i.e. be it a Python one or not :-)
2. `import n<Tab>` or `from a<Tab>` are not showing the new module icon on the completion widget
3. For non-Python files, we shouldn't show a transparent icon next to completions, but not icon at all. This is an aesthetic preference, but it seems a bit ugly to have all completions shown with (what appears to be) a wide margin to the right.

blink1073:
Fixes #2352.

### DIFF
--- a/spyderlib/utils/introspection/fallback_plugin.py
+++ b/spyderlib/utils/introspection/fallback_plugin.py
@@ -34,7 +34,7 @@ class FallbackPlugin(IntrospectionPlugin):
         Simple completion based on python-like identifiers and whitespace
         """
         items = []
-        if (info.line.strip().startswith(('import ', 'from ')) and
+        if (info.line.strip().startswith(('import', 'from')) and
                 info.is_python_like):
             items += module_completion(info.line, [info.filename])
             return [(i, 'module') for i in sorted(items)]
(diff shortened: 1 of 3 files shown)


ccordoba12:
Thanks a lot for the ultra fast response ;-) I verified that all errors I found are fixed now!

There is one last thing I forgot to mention in issue #2352. When you write

    import <Tab>

the full list of modules is not returned now, while it was before.

Would you mind to fix this also as part of this PR? Thanks :-)

blink1073:
Fixed!

ccordoba12:
Thanks, it's working now :-)

There is yet another thing I just find out. We should use the new module icon only when doing

    import fo<Tab>

or

    from ba<Tab>

but not when doing

    from foo import ba<Tab>

because that shows every completion as a module, which is obviously wrong :-)

If in this last case our completion libraries can't introspect types (as I think it's the case), we need to show no icons at all :-)

blink1073:
Okay, now we get icons for `import nu<Tab>` and `from nu<Tab>`, but not `from numpy import a<Tab>`.

blink1073:
(unless you have Jedi of course :smile:)

ccordoba12:
Great work man!! Merging :-)

PR #2309, opened 2015-04-08
Replace obsolete setTextColor method by setForeground in qtreewidgetitem

dhoegh:
Fixes #2305

---

@Nodd could you check if this fixes your problem?

PR #2362, opened 2015-04-24
Load spyderplugins from directories

Nodd:
So, this is the first step towards pip-installable plugins. I converted the 3 plugins (pylint, breakpoints and profiler) so that they each have their own directory structure, as would an external plugin.

I tried to stay as close as possible to the original layout.

TODO:
- [x] load images correctly
- [x] check that translations still work
- [x] update gettext_compile.py and gettext_scan.py
- [x] load from `.spyder/userplugins/`
- [x] Try to install an external plugin via pip

Things left for a following PR:
* rework the plugin API
* UI to see loaded plugins, error, to install local plugins...
* command-line option to disable plugins ("fail-safe" mode)


spyderplugins/p_breakpoints/p_breakpoints.py
```diff
@@ -14,10 +14,10 @@
 
 # Local imports
 from spyderlib.baseconfig import get_translation
-_ = get_translation("p_breakpoints", dirname="spyderplugins")
+_ = get_translation("p_breakpoints", dirname="spyderplugins.p_breakpoints")
```
Nodd: There is something strange here, `dirname` should be set to the module name, not to the directory name. (Same problem for all 3 plugins.) Should I change `get_translation` ? This will be quite confusing for future plugin devs.

Nodd:
Note: it will conflict with #2309 since I moved the files around

Nodd:
@ccordoba12 Is `gettext_scan.py` up to date ? I updated the file but it fails before my changes on `do_rescan("spyderlib")`:
```
  File "gettext_scan.py", line 13, in <module>
    do_rescan("spyderlib")
  File "/usr/lib/python2.7/site-packages/guidata/gettext_helpers.py", line 46, in do_rescan
    do_rescan_files(files, modname, dirname)
  File "/usr/lib/python2.7/site-packages/guidata/gettext_helpers.py", line 55, in do_rescan_files
    ]+files)
  File "/usr/lib/python2.7/subprocess.py", line 522, in call
    return Popen(*popenargs, **kwargs).wait()
  File "/usr/lib/python2.7/subprocess.py", line 710, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1335, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
```
I can't test `gettext_compile.py` properly because the previous step fails.

Nodd:
If you want to test this, I created a new branch in spyder_autopep8 (https://github.com/Nodd/spyder_autopep8/tree/new-style) which installs as a namespace package.

goanpeca:
Why keep using the p_ in the package names? we should not use it...

goanpeca:
I liked my version more :p https://github.com/goanpeca/spyderplugin.autopep8

Plus we should decide ASAP the guidelines to follow (enforce) on the name of the packages and the folder structure.

Nodd:
The `p_` prefix is needed to differentiate strandard plugins from io plugins. They are not loaded the same way at all. I don't know yet how to better manage these different kinds of plugins.
Of course it's not needed *inside* the plugin directory, I just wanted to have a minimum working example.

What did you prefer in your version (the `p_` prefix aside) ? The main difference I see is that you put lots of code in `__init__.py`, but that's still doable (but I don't really like it).

In this PR here is what is needed in a plugin ti be loaded:
* it resides in the `spyderplugins` namespace
* its name starts with `p_`
* `PLUGIN_CLASS` is importable from the top level of the plugin and has the interface of a spyder plugin.
This applies to both directory plugins and one-file plugins. Actually the prevuious directory layout would have loaded fine.
Inside the plugin you can do whatever you want, there is a default path for icons (`images`) but it is configurable.

goanpeca:
Basically I do not think we should keep using the p_, and we need to e clear on the naming conventions of the packages. Maybe we need a different way also for io plugins. The point is we should not relly on the old way of loading plugins cause it was kinda hacky...

I liked `spyderplugin.autopep`, even though it was lengthy I think it is a good idea to define a "standard" name for plugins unless the plugins is a hook of another standalone module (like conda-manager, that is why it is not named, spyderplugin.conda or similar...)

goanpeca:
@Nodd thanks for working on this :-), I had started but never made the actual push.

Will check the code tomorrow :p

goanpeca:
Question, would you also address loading plugins from the `.spyder` folder?

bilderbuchi:
knowing nothing much about the plugin code itself, differentiating between two different classes of plugins purely with the name prefix feels strange. maybe it's better to have several classes of plugins (e.g. `io_plugin` and `standard_plugin`), both derived from a common base, or have some flag in the plugin code which makes the "category" clear?

Nodd:
I agree with both of you that the need for the `p_` prefix needs to be removed, but I don't plan to solve it in this PR (otherwise I won't have the time to finish it properly).

@goanpeca Loading plugins from `.spyder` should be easy enough. You can't (and won't be able to)load a plugin from any directory, it will be either in `spyderplugins`, either in `.spyder/userplugins`.
I also prefer the name `spyderplugin.autopep` (maybe I'll change it someday), but I don't think that it makes sense to force plugin devs to use a specific name for their project. As long as the content is compatible, they can use whatever they want.
What we should agree on is a common pattern for "official" plugins (those developed by spyder's team). Mu vote goes to `spyderplugin.<pluginname>`.

Nodd:
Loading from `~/.spyder/userplugins/` now works.

I have problems loading the icons properly, it worked once but not anymore :( Maybe the problem is that I installed spyder in multiple places to test things. I should really learn to user virtualenvs.

spyderlib/otherplugins.py
```diff
@@ -8,52 +8,71 @@
...
-            traceback.print_exc(file=sys.stderr)
+
+    # Load user plugins
+    user_plugins = _get_spyderplugins(USERPLUGIN_PATH, prefix)
+    sys.path.insert(0, USERPLUGIN_PATH)
+    try:
+        for modname in user_plugins:
+            _import_plugin(modname, modlist)
+    finally:
+        del sys.path[0]
```
Nodd: Is in necessary to clean `sys.path` afterwards ? I don't know if this will modify the path when user run scripts in spyder (it would be very bad).

spyderplugins/p_breakpoints/breakpoints.py
```diff
@@ -14,10 +14,10 @@
 
 # Local imports
 from spyderlib.baseconfig import get_translation
-_ = get_translation("p_breakpoints", dirname="spyderplugins")
+_ = get_translation("p_breakpoints", dirname="spyderplugins.p_breakpoints")
```
Nodd: There is something strange here, dirname should be set to the module name, not to the directory name. (Same problem for all 3 plugins.) Should I change get_translation ? This will be quite confusing for future plugin devs.

goanpeca: I gues this function would be clear if we added something like a flag to tell it is a plugin... an user/developer should not need to give the directory info..., so it would be better if it was something like:

```python
_ = get_translation("p_breakpoints", plugin=True)
```

Nodd: THe problem I raised was with `get_translation` itself (the `dirname`argument requires a plugin name), but it's also true that I could simplify this. I'll have a look at it.

Nodd:
I tested everything cleanly using conda envs and from what I can tell it works properly. This covers the internal plugins (breakpoints, pylint et profiler), an a compatible external plugin is https://github.com/Nodd/spyder_autopep8/tree/new-style (not the `master` branch but the `new-style` one).

I would like this PR to be merged before modifying the plugin API itself. There are 2 minor questions left:
* the question to @ccordoba12 about gettext_compile.py and gettext_scan.py (https://github.com/spyder-ide/spyder/pull/2362#issuecomment-96081419)
* I don't like modifying `sys.path` very much but I don't see any other way to do it (https://github.com/spyder-ide/spyder/pull/2362#discussion_r29). Any suggestions ?

spyderlib/otherplugins.py
bilderbuchi: I think if you add something to `sys.path` temporarily, you should definitely remove it. Maybe it's also a good idea to test if you remove the correct element (in case something else got prepended on `sys.path` in the meantime, maybe like so: 

```
if USERPLUGIN_PATH in sys.path:
    sys.path.remove(USERPLUGIN_PATH)
```
(see [here](http://stackoverflow.com/a/2793356/599884))

Nodd:
I used @bilderbuchi for cleaning `sys.path`, I think i's ok like this. I didn't test if it was present because if it's not one plugin did something very nasty !

Nodd:
This one is ready for review, once this is in I'll work on the plugin API.

goanpeca:
:+1: 

Just to know, do you plan to work on a single way of installing plugins, so that IO plugins use the same machinery instead of prepending stuff?, like defining classes of plugins?

Also will you work on the UI for disabling/enabling available plugins detected?

Cheers, looks very cool :)

Nodd:
I don't know how to manage IO plugins vs standard plugins. They are loaded in different part of the code with nothing in common, the do very different things.
Maybe we can have no prefix for standard plugin, and keep the `io_` prefix for IO plugins since they will be less common ?

goanpeca:
Well the io plugins are something specific to which plugin? the variable explorer? if this is the case then the plugin API for the variable explorer should consider this case.

Agree on the UI

Nodd:
io plugins are used in consoles and the variable explorer. They are not UI plugins, I don't see how we could merge the two types of plugins without rewriting a big chunk of spyder's internals. In fact they are so different that we could have a specific `spyderioplugins` namespace, it would be the cleanest way.

goanpeca:
Ok, could be an interesting idea to have that separate namespace, I am not familiar with io plugins or what they do.. I think I have never "use them" 

@ccordoba12 maybe some enlightenment ?

ccordoba12:
@Nodd, I added the `gettext_helpers` module from `guidata`, so that we don't need to install external packages to generate translations. This means you need to rebase this PR to get the change and update our translations :-)

@goanpeca, I really don't know how `io` plugins work because I've never touched its code so far. My only take is agreeing with @Nodd about moving them to `spyderioplugins`.

gettext_compile.py
```diff
@@ -8,6 +8,6 @@
 
 if __name__ == "__main__":
     do_compile("spyderlib")
-    do_compile("p_pylint", "spyderplugins")
-    do_compile("p_profiler", "spyderplugins")
-    do_compile("p_breakpoints", "spyderplugins")
+    do_compile("p_pylint", "spyderplugins/p_pylint")
+    do_compile("p_profiler", "spyderplugins/p_pylint")
+    do_compile("p_breakpoints", "spyderplugins/p_pylint")
```
ccordoba12: Maybe this is why you're getting an error when generating translations. The last two lines here should be

```python
    do_compile("p_profiler", "spyderplugins/p_profiler")
    do_compile("p_breakpoints", "spyderplugins/p_breakpoints")
```

Nodd: Erf, this is a bug indeed, but the problem I get is on the line 10: `do_compile("spyderlib")`. The following lines don't even get a chance to be run.

ccordoba12: Please rebase and re-run again. I also updated a couple of translations that were failing on `spyderlib` :-)

Nodd: `gettext_compile.py ` runs fine, but `gettext_scan.py` gives:
```
  File "gettext_scan.py", line 10, in <module>
    do_rescan("spyderlib")
  File "/home/joseph/prog/spyder.git/gettext_helpers.py", line 46, in do_rescan
    do_rescan_files(files, modname, dirname)
  File "/home/joseph/prog/spyder.git/gettext_helpers.py", line 55, in do_rescan_files
    ]+files)
  File "/usr/lib/python3.4/subprocess.py", line 537, in call
    with Popen(*popenargs, **kwargs) as p:
  File "/usr/lib/python3.4/subprocess.py", line 859, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.4/subprocess.py", line 1457, in _execute_child
    raise child_exception_type(errno_num, err_msg)
FileNotFoundError: [Errno 2] No such file or directory: 'pygettext'
```
Do I have to install `pygettext` from somewhere ?

ccordoba12: In Ubuntu is part of the Python package itself. Maybe in Arch is part of another package ;-)

Nodd:
Gah, it looks like having both a namespace package (in my local git repo) and a normal package of the same name (from my spyder's system) is alywas taking the normal package event if `sys.path` has the namespace package first.
It should not be a problem for standard installs, but it's quite complicated to develop it....

goanpeca:
The thingbis that a namespace package will store the place where it was last called, so it first look in you local repo, and then the spyder systema and leaves that as the location, but i think a namesapce package stores all the places it finds a possible path so we can look for that and make sure in development it uses the one we want.

spyderlib/otherplugins.py
```diff
@@ -8,52 +8,73 @@
...
+from spyderlib.baseconfig import get_conf_path
+from spyderlib.utils.external.path import path as Path
 
 
-# Calculate path to `spyderplugins` package, where Spyder looks for all 3rd
-# party plugin modules
-PLUGIN_PATH = None
-if programs.is_module_installed("spyderplugins"):
-    import spyderplugins
-    PLUGIN_PATH = osp.abspath(spyderplugins.__path__[0])
```
goanpeca: @nodd this is where you could get all paths for a namespace, the __path__[0] is taking only the last imported... But that should contain also the local path were you are devloping.

Nodd:
thanks @goanpeca I see how it works now.

Nodd:
I think I got it, plugins should load fine. I still have to check if.how localization will work.

goanpeca:
@Nodd I ran into the same issues when working on this as well. 

That is why since the beginning I preferred (and insisted) that these plugins were either transformed into standalone plugins (outside the spyder repo), or converted to spyder native plugins (eliminating the spyderplugins folder from spyder repo), so that when developing there was one and only one obvious place to look for plugins, `.spyder\userplugins` which would not conflict or generate this problems when importing...

Nodd:
Importing woks fine now (well, it needs testing on more platforms), I changed the user plugin directory to `.spyder/spyderplugins` so that plugins are always imported as `spyderplugins.myplugin`. This way translations should always be loaded the same way.

Having pip installable and user installable plugins is necessary, I need it myself.

goanpeca:
Ok glad you sorted it out :)

goanpeca:
@Nodd will translations look in the right place all the time? For instance are the translations for `pylint`, `profiler` etc being looked first in the local repo (if using bootstrap) instead of the installed one?

Nodd:
I plan to load them from the same directory the code was loaded, but I don't know how it works now.

goanpeca:
this has conflicts? rebase @Nodd ?

Nodd:
Rebased !
But I have a weird error now:
```
  File "/home/joseph/prog/spyder.git/spyderlib/otherplugins.py", line 53, in _import_plugin
    submodule = imp.load_module(submodule_name, *info)
  File "/usr/lib/python3.4/imp.py", line 235, in load_module
    return load_source(name, filename, file)
  File "/usr/lib/python3.4/imp.py", line 171, in load_source
    module = methods.load()
  File "<frozen importlib._bootstrap>", line 1220, in load
  File "<frozen importlib._bootstrap>", line 1200, in _load_unlocked
  File "<frozen importlib._bootstrap>", line 1129, in _exec
  File "<frozen importlib._bootstrap>", line 1467, in exec_module
  File "<frozen importlib._bootstrap>", line 1577, in get_code
  File "<frozen importlib._bootstrap>", line 672, in _code_to_bytecode
ValueError: unmarshallable object
```

goanpeca:
My dear @Nodd 

Just tested the conda-manager with this PR :-)

image

So it is working :)

goanpeca:
I installed your branch and replaced the anaconda spyder to test this, but if I check the spyderplugins dir, it still looks the old way...

goanpeca:
Ok, so I ran from bootstrap, and works ok as well but I get this error

```python

<module 'spyderplugins.widgets' from 'spyderplugins\widgets\__init__.pyc'>: 'mod
ule' object has no attribute 'PLUGIN_CLASS'
  File "D:\spyder-plugins\spyderlib\spyder.py", line 1141, in setup
    plugin = mod.PLUGIN_CLASS(self)
AttributeError: 'module' object has no attribute 'PLUGIN_CLASS'
<module 'spyderplugins.io_dicom' from 'spyderplugins\io_dicom.pyc'>: 'module' ob
ject has no attribute 'PLUGIN_CLASS'
  File "D:\spyder-plugins\spyderlib\spyder.py", line 1141, in setup
    plugin = mod.PLUGIN_CLASS(self)
AttributeError: 'module' object has no attribute 'PLUGIN_CLASS'
<module 'spyderplugins.io_hdf5' from 'spyderplugins\io_hdf5.pyc'>: 'module' obje
ct has no attribute 'PLUGIN_CLASS'
  File "D:\spyder-plugins\spyderlib\spyder.py", line 1141, in setup
    plugin = mod.PLUGIN_CLASS(self)
AttributeError: 'module' object has no attribute 'PLUGIN_CLASS'

```

Hmm maybe the setup.py needs some fixing cause after deleting the spyderplugins folder from site packages and installing your branch I get this:

image

Otherwise seems to work nicely

goanpeca:
@Nodd ?

Nodd:
Nice to see it works !
I don't know why there is both the old and new files in spyderplugins. I may have failed a rebase.
In any case the warnings are normal, they are due to the leftover files and have nothing to do with the conda manager.

goanpeca:
Not sure what is going on... but when I ran `python setup.py install` that happens... weird

Nodd:
In my branch I don't see the old files in `spyderplugins`. Could you retest and make sure that spyder is correctly uninstalled before installing my branch ?

goanpeca:
I did this already.. I had uninstalled and remove the spyderplugins dir from site-packages,

goanpeca:
I will try again

goanpeca:
I do not see the files in your branch... but when I installed from your branch, then suddenly the spyderplugins folder inside site-packages had both your new structure + the old one... weird

Nodd:
I fixed the install for `spyderioplugins`, but I can't reproduce the problem with `spyderplugins`. My conclusion is that your computer is haunted and you should apply holy water to remove the old ghost version of spyder.

goanpeca:
I just retested and it works fine, I have no clue what happened before :p, so everything is working fine now, we should merge this soon, so we can have a go at the plugin manager.

If you can help me with some generic methods I can implement the UI for it (the plugin manager).

By generic methods I mean something like `get_installed_plugins()`, `get_user_plugins()`, and each plugin should query a method like `is_active_plugin(plugin)` that can return for now always True, and this is where we can inject the check (in CONF or something...)

Cheers

Nodd:
Right now plugins are discovered and loaded on the fly. I have to put an intermediate step in between to be able to load it or not.

goanpeca:
@Nodd we will need a rebase now to update to the new icons code....

Nodd:
True.
I was also wondering if `spyderplugins` should be replaced by `spyderuiplugins`, for 3 reasons:
* Even if it is possible to hack anything in spyder, the plugins are more oriented towards UI addons (be it a dockwidget, a menu entry or a toobar button)
* It is similar to `spyderioplugins`
* And the most important: there would be no weird error messages if both 2.X and a 3.X spyder versions are installed at the same time.

goanpeca:
I agree!, it would make the migration to 3 much easier, go for it.

@ccordoba12, @SylvainCorlay, @blink1073 ?

SylvainCorlay:
Yes, using a new name would help.

ccordoba12:
@Nodd, this needs andes.edu.co/a rebase

ccordoba12:
> I was also wondering if spyderplugins should be replaced by spyderuiplugins

I think (for now) we should rename this dir to spyder_plugins. Inside it we could have subdirs called `ui` (or even better `graphical`), `io`, etc, to not have one dir per type in `site_packages`. This would avoid Spyder polluting it with too many `spyder_foo` subdirs.

But let's leave that decision for another PR. For now, we should just do the rename, specially to avoid problems between 2.x and 3.x.

ccordoba12:
Ok, forget my last comment ;-) I just saw that @Nodd move our `io` plugins to `spyderioplugins`. Then, I'd like to propose my above plan: use a single `spyder_plugins` and inside it subdirs for different types of plugins.

Nodd:
I didn't move anything yet. If I understand correctly you want to first use `spyderioplugins`, the change to `spyder_plugins/io` afterwards ?
I understand what you're proposing but:
* a module name with an underscore is uncommon.
* it will be a bit strange to plugin devs to have to put everything into spyder_plugins/io (but that's a detail)

goanpeca:
I would prefer if we kept it like it is now

`spyderioplugins` -> namespace package

and make `spyderplugins` into `spyderuiplugis` as a namespace package. 

It will avoid the crash between 2.3 and 3.x and will keep things clean and clear. I do not think that we will start making many more namespacepackages besides those 2 so polluting sitepackages should not be an issue

goanpeca:
@Nodd, any chance you get to finish this before Sunday?

Nodd:
Well, my daughter was born this morning, so it will be kinda hard to do anything this we, sorry… It's "just" a matter of moving files and imports around so maybe the easiest way is to merge as-is and change the name afterwards ?
Apart from this I think that everyting else is working fine.

goanpeca:
Congrats :-), I imagined this was happening soon!

This would still need a rebase anyway, dont worry ;-), enjoy your daughter

SylvainCorlay:
Congratulations!
Tous mes voeux de bonheur à toute ta famille.

ccordoba12:
Closing, as this was superseded by PR #2565.

PR #2366, opened 2015-04-26
Update Path.py to version 7.3

### LINKED ISSUE #2361: Use spyderlib/utils/external/path.py internally ?
Should we use path.py given that it's already embedded ? It's really easier to manage paths with it rather than `import os.path as osp`.

This would be for new development only, there is no need to change working code.

Nodd:
Closes #2361.

Feel free to use it for new core in Spyder, it greatly simplifies working with paths.
The documentation is here: https://pythonhosted.org/path.py/api.html

### DIFF
--- a/spyderlib/utils/external/path.py
+++ b/spyderlib/utils/external/path.py
@@ -8,8 +8,8 @@
 # copies of the Software, and to permit persons to whom the Software is
 # furnished to do so, subject to the following conditions:
 #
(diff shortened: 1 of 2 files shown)


Nodd:
Why do I see the commit author as @margorczynski ? That's strange....

ccordoba12:
Seems to be working fine, merging.

Issue #2340, opened 2015-04-20
SublimeText-like file switching widget


d1manson:
In Sublime Text, you press Ctrl-P to get an intelligent, fast switching between open files/files in the current folder.  See [here](http://sublime-text-unofficial-documentation.readthedocs.org/en/latest/file_management/file_management.html).

I find myself trying to use this shortcut in Spyder, and then get annoyed when I have to click through little buttons at the top of the editor.

I imagine that this tool is reasonably straightforward to implement as it's fairly stand alone.

goanpeca:
@d1manson sure, something that seems worth implementing. Do you want to give it a try?

Right now it would conflict with the Print command, so I guess some discussion is needed.

@ccordoba12, @blink1073, @Nodd

Nodd:
There already is `Ctrl+E` for this, but in master I get in the internal console:
```
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/spyderlib/widgets/editor.py", line 643, in open_filelistdialog
    dlg.edit_file.connect(self.set_stack_index)
AttributeError: 'function' object has no attribute 'connect'
```

d1manson:
Ah thanks, I didn't know about that dialogue - it *does* work for me. Although it's not quite as slick as Sublime's it's a massive improvement over clicking on tabs!  Thanks

goanpeca:
Me neither... wow.. we really need to document this!

We could make it more slick... want to tackle that @d1manson ?

Nodd:
THere should be an entry in the file menu to make it more discoverable. I heard about it by a comment froom @ccordoba12 in an issue quite recentely. And I forgot about it until I saw this !

@d1manson what would make it slicker ?

d1manson:
Well, it's already pretty good.  But, for extra slickness (to copy Sublime)...
* I'd like to be able to use up/down keys on the list while the filter box has the focus.
* When filtering, the matched text should be highlighted in the list.
* Whichever item in the list is currently selected should be brought to the front of the editor (but if you Esc out of the dialogue rather than hitting Enter you revert back to where you were).
* The list should be sorted based on recent usage (or at least alphabetically, right now I'm not sure if it's sorted on anything in particular).
* Most importantly - the shortcut should be available in all contexts, not just when the editor has the focus.

I'm not too keen on volunteering for this as I've not done any work on Spyder or with Qt.  But I might have a look at some point this week and see if it's something I could do in a few hours.

[...]

PR #2217, opened 2015-03-01
Keybinding Enhancements

blink1073:
Fixes #729
Fixes #1212
Fixes #2047
Fixes #2347

----

Adds the following keybindings, including kill-ring support:

```python
              'editor/start of line': "Meta+A",
              'editor/end of line': "Meta+E",
              'editor/previous line': "Meta+P",
              'editor/next line': "Meta+N",
              'editor/previous char': "Meta+B",
              'editor/next char': "Meta+F",
              'editor/previous word': "Meta+Left",
              'editor/next word': "Meta+Right",
              'editor/kill to line end': "Meta+K",
              'editor/kill to line start': "Meta+U",
              'editor/yank': 'Meta+Y',
              'editor/rotate kill ring': 'Shift+Meta+Y',
              'editor/kill previous word': 'Meta+Backspace',
              'editor/kill next word': 'Meta+D',
              'editor/start of document': 'Ctrl+Up',
              'editor/end of document': 'Ctrl+Down',
```

Also makes the follow existing key sequences configurable:

```python
              'editor/undo': 'Ctrl+U',
              'editor/redo': 'Ctrl+Y',
              'editor/cut': 'Ctrl+X',
              'editor/copy': 'Ctrl+C',
              'editor/paste': 'Ctrl+V',
              'editor/delete': 'Del',
              'editor/select All': "Ctrl+A",
              'editor/save as': 'Ctrl+Shift+S',
```

The only changed shortcut is

```python
              'editor/save all': 'Ctrl+Alt+S',
```

PR #2367, opened 2015-04-26
New KeySequence Editor for Keyboard Shortcut Preferences

### LINKED ISSUE #478: Slash does not work as keyboard shortcut key
_From [joonp...@gmail.com](https://code.google.com/u/101070586672081573036/) on 2010-11-30T12:51:04Z_

 1. Assign Ctrl as Mod3 and Slash as Key for comment What is the expected output? What do you see instead? The shortcut does not work, and in the Edit menu, the shortcut for Comment is showing as "Ctrl+" What version of the product are you using? On what operating system? 2.0.1, openSUSE X64

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=478_

### LINKED ISSUE #494: Need to hit Enter to change a keyboard shortcut (unintuitive)
_From [christop...@gmail.com](https://code.google.com/u/105753816054647931594/) on 2010-12-10T04:52:50Z_

 1. In keyboard shortcut preferences pane change a modifier or key field.
2. Don't move the focus to a different field
3. Hit OK button What is the expected output? What do you see instead? Expected that change gets applied. Instead the change is lost. It is also observed that the Apply button remains grayed out until the focus moves to a different field in the grid. What version of the product are you using? On what operating system? spyder-2.0.3, OS X 10.6.4 

.

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=494_

### LINKED ISSUE #1940: Add search functionality for keyboard shortcuts
_From [a.molcha...@gmail.com](https://code.google.com/u/105109847022035150285/) on 2014-08-20T15:02:47Z_

 1. try to search for shortcuts in shortcuts configuration menu What version of the product are you using? On what operating system? v 2.2.5

Description:
Please, add shortcuts search (in shortcuts configuration menu) the way it is done in Kate (or maybe better, for example, to be able to search for combinations and keywords separately).

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=1940_

### LINKED ISSUE #2125: Removal of keyboard shortcut causes errors to be thrown
_From [thevp...@gmail.com](https://code.google.com/u/114967603654191624717/) on 2015-01-20T23:56:05Z_

Spyder Version:  2.3.2
Python Version:  3.4.1
Qt Version    :  4.8.6, PyQt4 (API v2) 4.10.4 on Windows
pyflakes >=0.6.0:  0.8.1 (OK)
pep8 >=0.6      :  1.5.7 (OK)
IPython >=0.13  :  2.3.1 (OK)
zmq >=2.1.11    :  14.4.1 (OK)
pygments >=1.6  :  2.0.1 (OK)
pandas >=0.13.1 :  0.14.0 (OK)
sphinx >=0.6.6  :  1.2.3 (OK)
rope >=0.9.2    :  0.9.4-1 (OK)
matplotlib >=1.0:  1.4.2 (OK)
sympy >=0.7.0   :  0.7.5 (OK)
pylint >=0.25   :  1.4.0 (OK) 

1. Go to Keyboard Shortcuts (Tools -> Preferences --> Keyboard Shortcuts)
2. Pick a shortcut of your choice and select the blank value for all shortcut modifiers
3. Double-click on "Key" column for that shortcut, and select a random character.
4. Press "OK", and verify that the shortcut has been "removed" by checking the menus (In my case, I first tested the removal of the shortcut for "Quit", and the character I chose was "0". The option under File-->Quit showed that there was no shortcut to that command.) 

 An error message should pop-up, stating that this is an invalid choice. Or, the command item should be removed completely. However, Spyder accepts the options but starts throwing errors and becomes unstable, throwing errors described in the section below. Also, at this point, I cannot re-define the shortcut - any choices made will be cleared as soon as another box is pressed. Only way to make things stable again is to press "restore keyboard shortcuts". 

Errors produced in console after the problem is created and the Keyboard Shortcuts menu is opened in the Preferences window.

```
AttributeError: 'int' object has no attribute 'modifiers'
'int' object has no attribute 'modifiers'
  File "C:\Anaconda3\lib\site-packages\spyderlib\plugins\shortcuts.py", line 158, in data
    return to_qvariant(Key.MODIFIERNAMES[key.modifiers[0]])
AttributeError: 'int' object has no attribute 'modifiers'
  File "C:\Anaconda3\lib\site-packages\spyderlib\plugins\shortcuts.py", line 160, in data
    return to_qvariant(Key.MODIFIERNAMES[key.modifiers[1]])
AttributeError: 'int' object has no attribute 'modifiers'
  File "C:\Anaconda3\lib\site-packages\spyderlib\plugins\shortcuts.py", line 162, in data
    return to_qvariant(Key.MODIFIERNAMES[key.modifiers[2]])
AttributeError: 'int' object has no attribute 'modifiers'
  File "C:\Anaconda3\lib\site-packages\spyderlib\plugins\shortcuts.py", line 164, in data
    return to_qvariant(Key.KEYS[key.key])
AttributeError: 'int' object has no attribute 'key'
```

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=2125_

blink1073:
Fixes #478
Fixes #494
Fixes #1940
Fixes #2125

----

Allows the user to enter keyboard shortcut preferences by typing them in.

### DIFF
--- a/spyderlib/plugins/shortcuts.py
+++ b/spyderlib/plugins/shortcuts.py
@@ -85,7 +85,7 @@ class HTMLDelegate(QStyledItemDelegate):
         self._margin = margin
 
     def paint(self, painter, option, index):
-        options = QStyleOptionViewItemV4(option)
+        options = QStyleOptionViewItem(option)
         self.initStyleOption(options, index)
 
         style = (QApplication.style() if options.widget is None
@@ -109,7 +109,7 @@ class HTMLDelegate(QStyledItemDelegate):
         painter.restore()
 
     def sizeHint(self, option, index):
-        options = QStyleOptionViewItemV4(option)
+        options = QStyleOptionViewItem(option)
         self.initStyleOption(options, index)
 
         doc = QTextDocument()
--- a/spyderlib/plugins/shortcuts.py
+++ b/spyderlib/plugins/shortcuts.py
@@ -174,6 +174,7 @@ class ShortcutEditor(QDialog):
         self.text_current_sequence = QLabel(sequence)
         self.label_new_sequence = QLabel(_("New shortcut:"))
         self.text_new_sequence = CustomLineEdit(self)
+        self.text_new_sequence.setPlaceholderText(sequence)
         self.helper_button = HelperToolButton()
 
         bbox = QDialogButtonBox(QDialogButtonBox.Ok | QDialogButtonBox.Cancel)
--- a/spyderlib/plugins/shortcuts.py
+++ b/spyderlib/plugins/shortcuts.py
@@ -757,13 +757,13 @@ class ShortcutsConfigPage(GeneralConfigPage):
         vlayout = QVBoxLayout()
         hlayout.addWidget(self.label_finder)
         hlayout.addWidget(self.finder)
-        vlayout.addLayout(hlayout)
         vlayout.addWidget(self.table)
+        vlayout.addLayout(hlayout)
         vlayout.addWidget(self.reset_btn)
         self.setLayout(vlayout)
 
-        self.setTabOrder(self.finder, self.table)
-        self.setTabOrder(self.table, self.reset_btn)
+        self.setTabOrder(self.table, self.finder)
+        self.setTabOrder(self.finder, self.reset_btn)
 
         # Signals and slots
         self.table.proxy_model.dataChanged.connect(
--- a/spyderlib/plugins/shortcuts.py
+++ b/spyderlib/plugins/shortcuts.py
@@ -7,13 +7,6 @@
 """Shortcut management"""
 
 from __future__ import print_function
-<<<<<<< HEAD
-import sys
-
-from spyderlib.qt.QtGui import (QVBoxLayout, QTableView, QMessageBox,
-                                QPushButton, QKeySequence)
-from spyderlib.qt.QtCore import Qt, QAbstractTableModel, QModelIndex
-=======
 import os
 import re
 import sys
@@ -25,10 +18,11 @@ from spyderlib.qt.QtGui import (QVBoxLayout, QTableView, QMessageBox,
                                 QSortFilterProxyModel, QStyledItemDelegate,
                                 QStyleOptionViewItemV4, QApplication,
                                 QTextDocument, QStyle, QSpacerItem,
-                                QAbstractTextDocumentLayout)
+                                QAbstractTextDocumentLayout,
+                                QRegExpValidator, QToolTip,
+                                QHBoxLayout)
 from spyderlib.qt.QtCore import (Qt, QAbstractTableModel, QModelIndex, QSize,
-                                 QPoint)
->>>>>>> 69e106a... Fixing merge
+                                 QPoint, QRegExp)
 from spyderlib.qt.compat import to_qvariant, from_qvariant
 import spyderlib.utils.icon_manager as ima
 
@@ -37,16 +31,15 @@ from spyderlib.baseconfig import _, debug_print
 from spyderlib.guiconfig import (get_shortcut, set_shortcut,
                                  iter_shortcuts, reset_shortcuts)
 from spyderlib.plugins.configdialog import GeneralConfigPage
-<<<<<<< HEAD
-=======
-Qt.Key_F
+from spyderlib.utils.qthelpers import get_std_icon
+from spyderlib.widgets.arraybuilder import HelperToolButton
 
 MODIFIERS = {Qt.Key_Shift: Qt.SHIFT,
              Qt.Key_Control: Qt.CTRL,
              Qt.Key_Alt: Qt.ALT,
              Qt.Key_Meta: Qt.META}
 
-SINGLE_KEYS = ["F{}".format(_i) for _i in range(1, 36)] + [ "Delete", "Escape"]
+SINGLE_KEYS = ["F{}".format(_i) for _i in range(1, 36)] + ["Delete", "Escape"]
 KEYSTRINGS = ["Tab", "Backtab", "Backspace", "Return", "Enter",
               "Pause", "Print", "Clear", "Home", "End", "Left",
               "Up", "Right", "Down", "PageUp", "PageDown"] + \
@@ -61,13 +54,36 @@ KEYSTRINGS = ["Tab", "Backtab", "Backspace", "Return", "Enter",
               "Control", "Alt", "Shift", "Meta"]
 VALID_SINGLE_KEYS = [getattr(Qt, 'Key_{0}'.format(k)) for k in SINGLE_KEYS]
 VALID_KEYS = [getattr(Qt, 'Key_{0}'.format(k)) for k in KEYSTRINGS+SINGLE_KEYS]
+VALID_ACCENT_CHARS = "ÁÉÍOÚáéíúóàèìòùÀÈÌÒÙâêîôûÂÊÎÔÛäëïöüÄËÏÖÜñÑ"
+VALID_FINDER_CHARS = "[A-Za-z\s{0}]".format(VALID_ACCENT_CHARS)
 
 
+class CustomLineEdit(QLineEdit):
+    """QLineEdit that filters its key press and release events."""
+    def __init__(self, parent):
+        super(CustomLineEdit, self).__init__(parent)
+        self.setReadOnly(True)
+        self.setFocusPolicy(Qt.NoFocus)
+
+    def keyPressEvent(self, e):
+        """Qt Override"""
+        self.parent().keyPressEvent(e)
+
+    def keyReleaseEvent(self, e):
+        """Qt Override"""
+        self.parent().keyReleaseEvent(e)
+
+
+# TODO: Move to another place for common use widgets afterwards.
 class HTMLDelegate(QStyledItemDelegate):
     """With this delegate, a QListWidgetItem can render HTML.
 
     Taken from http://stackoverflow.com/a/5443112/2399799
     """
+    def __init__(self, parent, margin=0):
+        super(HTMLDelegate, self).__init__(parent)
+        self._margin = margin
+
     def paint(self, painter, option, index):
         options = QStyleOptionViewItemV4(option)
         self.initStyleOption(options, index)
@@ -76,7 +92,7 @@ class HTMLDelegate(QStyledItemDelegate):
                  else options.widget.style())
 
         doc = QTextDocument()
-        doc.setDocumentMargin(9)  # FIXME: Need to center vertical text
+        doc.setDocumentMargin(self._margin)
         doc.setHtml(options.text)
 
         options.text = ""
@@ -103,57 +119,43 @@ class HTMLDelegate(QStyledItemDelegate):
         return QSize(doc.idealWidth(), doc.size().height())
 
 
-class ShortcutFinder(QDialog):
+class ShortcutFinder(QLineEdit):
+    """Textbox for filtering listed shortcuts in the table."""
     def __init__(self, parent, callback=None):
         super(ShortcutFinder, self).__init__(parent)
-        style = """
-            QDialog {
-              margin:0px;
-              border: 1px solid grey;
-              padding:0px;
-              border-radius: 2px;
-            }"""
-        self.setStyleSheet(style)
 
-        # widget setup
-        self.setWindowFlags(Qt.Window | Qt.Dialog | Qt.FramelessWindowHint)
-        self.setModal(True)
-        self.setWindowOpacity(0.90)
+        self._parent = parent
+        regex = QRegExp(VALID_FINDER_CHARS + "{100}")
+        self.setValidator(QRegExpValidator(regex))
 
-        self.text_edit = QLineEdit()
-        layout = QVBoxLayout()
-        layout.addWidget(self.text_edit)
-        margin = 1
-        layout.setContentsMargins(margin, margin, margin, margin)
-        self.setLayout(layout)
-
+        # Signals
         if callback:
-            self.text_edit.textChanged.connect(callback)
-
-    def text(self):
-        """ """
-        return self.text_edit.text()
+            self.textChanged.connect(callback)
 
     def set_text(self, text):
+        """Set the filter text"""
         text = text.strip()
-        self.text_edit.setText(text)
+        new_text = self.text() + text
+        self.setText(new_text)
 
     def keyPressEvent(self, event):
-        """ """
+        """Qt Override"""
         key = event.key()
         if key in [Qt.Key_Up]:
-            self.parent().previous_row()
+            self._parent.previous_row()
         elif key in [Qt.Key_Down]:
-            self.parent().next_row()
+            self._parent.next_row()
         elif key in [Qt.Key_Enter, Qt.Key_Return]:
-            self.accept()
-        super(ShortcutFinder, self).keyPressEvent(event)
+            self._parent.show_editor()
+        else:
+            super(ShortcutFinder, self).keyPressEvent(event)
 
 
 class ShortcutEditor(QDialog):
-    """ """
-    def __init__(self, parent, context, name, sequence):
+    """A dialog for entering key sequences."""
+    def __init__(self, parent, context, name, sequence, shortcuts):
         super(ShortcutEditor, self).__init__(parent)
+        self.context = context
         self.npressed = 0
         self.keys = set()
         self.key_modifiers = set()
@@ -162,6 +164,7 @@ class ShortcutEditor(QDialog):
         self.sequence = sequence
         self.new_sequence = None
         self.edit_state = True
+        self.shortcuts = shortcuts
 
         # Widgets
         self.label_info = QLabel(_("Press the new shortcut and select 'Ok': \n"
@@ -170,29 +173,39 @@ class ShortcutEditor(QDialog):
         self.label_current_sequence = QLabel(_("Current shortcut:"))
         self.text_current_sequence = QLabel(sequence)
         self.label_new_sequence = QLabel(_("New shortcut:"))
-        self.text_new_sequence = QLineEdit()
+        self.text_new_sequence = CustomLineEdit(self)
+        self.helper_button = HelperToolButton()
 
         bbox = QDialogButtonBox(QDialogButtonBox.Ok | QDialogButtonBox.Cancel)
         self.button_ok = bbox.button(QDialogButtonBox.Ok)
         self.button_cancel = bbox.button(QDialogButtonBox.Cancel)
 
-        self.text_new_sequence.setReadOnly(True)
-        self.text_new_sequence.setFocusPolicy(Qt.NoFocus)
+        # Setup widgets
+        self.setWindowTitle(_('Shortcut: {0}').format(name))
         self.button_ok.setFocusPolicy(Qt.NoFocus)
         self.button_ok.setEnabled(False)
         self.button_cancel.setFocusPolicy(Qt.NoFocus)
+        self.helper_button.setToolTip('')
+        self.helper_button.setFocusPolicy(Qt.NoFocus)
+        style = """
+            QToolButton {
+              margin:1px;
+              border: 0px solid grey;
+              padding:0px;
+              border-radius: 0px;
+            }"""
+        self.helper_button.setStyleSheet(style)
 
-        # Widget setup
-        self.setWindowTitle(_('Shortcut: {0}').format(name))
-
+        # Layout
         spacing = 24
         layout_sequence = QGridLayout()
-        layout_sequence.addWidget(self.label_info, 0, 0, 1, 2)
+        layout_sequence.addWidget(self.label_info, 0, 0, 1, 3)
         layout_sequence.addItem(QSpacerItem(spacing, spacing), 1, 0, 1, 2)
         layout_sequence.addWidget(self.label_current_sequence, 2, 0)
-        layout_sequence.addWidget(self.text_current_sequence, 2, 1)
+        layout_sequence.addWidget(self.text_current_sequence, 2, 2)
         layout_sequence.addWidget(self.label_new_sequence, 3, 0)
-        layout_sequence.addWidget(self.text_new_sequence, 3, 1)
+        layout_sequence.addWidget(self.helper_button, 3, 1)
+        layout_sequence.addWidget(self.text_new_sequence, 3, 2)
 
         layout = QVBoxLayout()
         layout.addLayout(layout_sequence)
@@ -245,6 +258,36 @@ class ShortcutEditor(QDialog):
 
         self.keys.add(key)
 
+    def toggle_state(self):
+        """Switch between shortcut entry and Accept/Cancel shortcut mode."""
+        self.edit_state = not self.edit_state
+
+        if not self.edit_state:
+            self.text_new_sequence.setEnabled(False)
+            if self.button_ok.isEnabled():
+                self.button_ok.setFocus()
+            else:
+                self.button_cancel.setFocus()
+        else:
+            self.text_new_sequence.setEnabled(True)
+            self.text_new_sequence.setFocus()
+
+    def nonedit_keyrelease(self, e):
+        """Key release event for non-edit state."""
+        key = e.key()
+
+        if key == Qt.Key_Escape:
+#            self.close()
+            pass
+
+        if key in [Qt.Key_Left, Qt.Key_Right, Qt.Key_Up,
+                   Qt.Key_Down]:
+
+            if self.button_ok.hasFocus():
+                self.button_cancel.setFocus()
+            else:
+                self.button_ok.setFocus()
+
     def keyReleaseEvent(self, e):
         """Qt override"""
         self.npressed -= 1
@@ -256,7 +299,7 @@ class ShortcutEditor(QDialog):
                 self.nonedit_keyrelease(e)
             else:
                 debug_print('keys: {}'.format(self.keys))
-                if self.keys and len(self.keys) <= 4:
+                if self.keys:
                     self.validate_sequence()
                 self.keys = set()
                 self.key_modifiers = set()
@@ -264,23 +307,54 @@ class ShortcutEditor(QDialog):
                 self.key_text = list()
                 self.npressed = 0
 
-    def nonedit_keyrelease(self, e):
+    def check_conflicts(self):
+        """Check shortcuts for conflicts"""
+        conflicts = []
+        for index, shortcut in enumerate(self.shortcuts):
+            sequence = str(shortcut.key)
+            if sequence == self.new_sequence and \
+                (shortcut.context == self.context or shortcut.context == '_' or
+                 self.context == '_'):
+                conflicts.append(shortcut)
+        return conflicts
+
+    def update_warning(self, reset=False):
         """ """
-        key = e.key()
+        conflicts = self.check_conflicts()
+        if reset:
+            conflicts = []
 
-        if key == Qt.Key_Escape:
-            self.reject()
+        widget = self.helper_button
 
-        if key in [Qt.Key_Left, Qt.Key_Right, Qt.Key_Up,
-                   Qt.Key_Down]:
+        if conflicts:
+            tip_title = _('The new entered shorcut conflicts with:') + '\n'
+            tip_body = ''
+            for s in conflicts:
+                tip_body += ' - {0}: {1}\n'.format(s.context, s.name)
+            tip = '{0}{1}'.format(tip_title, tip_body)
+            widget.setIcon(get_std_icon('MessageBoxWarning'))
+            widget.setToolTip(tip)
+            QToolTip.showText(widget.mapToGlobal(QPoint(0, 5)), tip)
+        else:
+            widget.setToolTip('')
+            QToolTip.hideText()
+            widget.setIcon(get_std_icon('DialogApplyButton'))
 
-            if self.button_ok.hasFocus():
-                self.button_cancel.setFocus()
-            else:
-                self.button_ok.setFocus()
+    def set_sequence(self, sequence):
+        """Set the new shortcut and update buttons."""
+        if self.sequence != sequence:
+            self.button_ok.setEnabled(True)
+            reset = False
+        else:
+            self.button_ok.setEnabled(False)
+            reset = True
+
+        self.text_new_sequence.setText(sequence)
+        self.new_sequence = sequence
+        self.update_warning(reset=reset)
 
     def validate_sequence(self):
-        """ """
+        """Provide additional checks for accepting or rejecting shortcuts."""
         for mod in MODIFIERS:
             non_mod = set(self.key_non_modifiers)
             non_mod.discard(mod)
@@ -308,6 +382,11 @@ class ShortcutEditor(QDialog):
             elif len(self.key_non_modifiers) > 1:
                 return
 
+        # QKeySequence accepts a maximum of 4 different sequences
+        if len(self.keys) > 4:
+            # If the user enters more than 4, take the first 4 only
+            self.keys = set([0, 1, 2, 3])
+
         keys = []
         for i in range(len(self.keys)):
             key_seq = 0
@@ -319,55 +398,15 @@ class ShortcutEditor(QDialog):
         sequence = QKeySequence(*keys)
         self.set_sequence(sequence.toString())
 
-    def toggle_state(self):
-        """ """
-        self.edit_state = not self.edit_state
-
-        if not self.edit_state:
-            self.text_new_sequence.setEnabled(False)
-            if self.button_ok.isEnabled():
-                self.button_ok.setFocus()
-            else:
-                self.button_cancel.setFocus()
-        else:
-            self.text_new_sequence.setEnabled(True)
-            self.text_new_sequence.setFocus()
-
-    def set_sequence(self, sequence):
-        """ """
-        if self.sequence != sequence:
-            self.text_new_sequence.setText(sequence)
-            self.button_ok.setEnabled(True)
-            self.new_sequence = sequence
-        else:
-            self.text_new_sequence.setText(sequence)
-            self.button_ok.setEnabled(False)
-
-    def check_conflicts(self):
-        """Check shortcuts for conflicts"""
-        conflicts = []
-        for index, sh1 in enumerate(self.model.shortcuts):
-            if index == len(self.model.shortcuts)-1:
-                break
-            for sh2 in self.model.shortcuts[index+1:]:
-                if sh2 is sh1:
-                    continue
-                if str(sh2.key) == str(sh1.key) \
-                   and (sh1.context == sh2.context or sh1.context == '_' or
-                        sh2.context == '_'):
-                    conflicts.append((sh1, sh2))
->>>>>>> 69e106a... Fixing merge
-
 
 class Shortcut(object):
-    """ """
     def __init__(self, context, name, key=None):
         self.context = context
         self.name = name
         self.key = key
 
     def __str__(self):
-        return "%s/%s: %s" % (self.context, self.name, self.key)
+        return "{0}/{1}: {2}".format(self.context, self.name, self.key)
 
     def load(self):
         self.key = get_shortcut(self.context, self.name)
@@ -380,22 +419,60 @@ CONTEXT, NAME, SEQUENCE = list(range(3))
 
 
 class ShortcutsModel(QAbstractTableModel):
-    """ """
     def __init__(self, parent):
         QAbstractTableModel.__init__(self)
         self.shortcuts = []
+        self._parent = parent
+        self.letters = ''
+        self.label = QLabel()
+        self.widths = []
+
+        # Needed to compensate for the HTMLDelegate color selection unawarness
+        palette = parent.palette()
+        self.text_color = palette.text().color().name()
+        self.text_color_highlight = palette.highlightedText().color().name()
+
+    def current_index(self):
+        """Get the currently selected index in the parent table view."""
+        i = self._parent.proxy_model.mapToSource(self._parent.currentIndex())
+        return i
+
+    def _enrich_text(self, text, color):
+        """Highlight (bold) letters from the shortcut finder."""
+        template = '<b>{0}</b>'
+        if not self.letters:
+            new_text = text
+        elif self.letters in text:
+            new_text = text.replace(self.letters, template)
+            new_text = new_text.format(self.letters)
+        else:
+            new_text = [l for l in text]
+            temp_text = [l.lower() for l in text]
+            for l in self.letters:
+                if l != u'' and l in temp_text:
+                    index = temp_text.index(l)
+                    new_text[index] = template.format(text[index])
+                    temp_text = [' ']*(index + 1) + temp_text[index+1:]
+            new_text = ''.join(new_text)
+
+        new_text = new_text
+        new_text = '<p style="color:{}">'.format(color) + new_text + '</p>'
+        return new_text
 
     def sortByName(self):
+        """Qt Override"""
         self.shortcuts = sorted(self.shortcuts,
                                 key=lambda x: x.context+'/'+x.name)
         self.reset()
 
     def flags(self, index):
+        """Qt Override"""
         if not index.isValid():
             return Qt.ItemIsEnabled
         return Qt.ItemFlags(QAbstractTableModel.flags(self, index))
 
     def data(self, index, role=Qt.DisplayRole):
+        """Qt Override"""
         if not index.isValid() or \
            not (0 <= index.row() < len(self.shortcuts)):
             return to_qvariant()
@@ -406,13 +483,15 @@ class ShortcutsModel(QAbstractTableModel):
 
         if role == Qt.DisplayRole:
             if column == CONTEXT:
-                return to_qvariant(_(shortcut.context.capitalize()))
+                return to_qvariant(shortcut.context)
             elif column == NAME:
-<<<<<<< HEAD
-                return to_qvariant(_(shortcut.name.capitalize()))
-=======
-                return to_qvariant(self._enrich_text(shortcut.name))
->>>>>>> 69e106a... Fixing merge
+                color = self.text_color
+                if self._parent == QApplication.focusWidget():
+                    if self.current_index().row() == index.row():
+                        color = self.text_color_highlight
+                    else:
+                        color = self.text_color
+                return to_qvariant(self._enrich_text(shortcut.name, color))
             elif column == SEQUENCE:
                 text = QKeySequence(key).toString(QKeySequence.NativeText)
                 return to_qvariant(text)
@@ -421,6 +500,7 @@ class ShortcutsModel(QAbstractTableModel):
         return to_qvariant()
 
     def headerData(self, section, orientation, role=Qt.DisplayRole):
+        """Qt Override"""
         if role == Qt.TextAlignmentRole:
             if orientation == Qt.Horizontal:
                 return to_qvariant(int(Qt.AlignHCenter | Qt.AlignVCenter))
@@ -437,16 +517,15 @@ class ShortcutsModel(QAbstractTableModel):
         return to_qvariant()
 
     def rowCount(self, index=QModelIndex()):
+        """Qt Override"""
         return len(self.shortcuts)
 
     def columnCount(self, index=QModelIndex()):
+        """Qt Override"""
         return 3
-<<<<<<< HEAD
-    
-=======
 
->>>>>>> 69e106a... Fixing merge
     def setData(self, index, value, role=Qt.EditRole):
+        """Qt Override"""
         if index.isValid() and 0 <= index.row() < len(self.shortcuts):
             shortcut = self.shortcuts[index.row()]
             column = index.column()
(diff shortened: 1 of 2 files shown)


blink1073:
So far I just slimmed it down to one column, next I will go and make it editable.

blink1073:
Note to self, use gist: https://gist.github.com/blink1073/946df268c3685a3f443e

blink1073:
Boom:

screen shot 2015-04-26 at 3 03 29 pm

blink1073:
I used the native shortcut display (this is OSX).

blink1073:
@ccordoba12, you can now enter the key sequences, and they are getting set properly in `Shortcut.save`, but the changes are not having any effect on the editor yet.

goanpeca:
Looks cool :-)

Nodd:
Beautiful !

blink1073:
I stand corrected, I had just forgotten to do a `--reset`.  
We can input any valid keyboard sequence, including compound ones.

ccordoba12:
This is a great work @blink1073!! Thanks for this :-)

I just have one suggestion for you, that I don't know if it's implemented in your PR or not. I'd like to have every entry in the new editor localized, i.e. that we can read *Completado del código* (when the locale is *es*) instead of *code completion*.

We are very good at localization, so it's really ugly to not have our shortcuts localized right now.

Nodd:
Also the options displayed should be Capitalized in English too, right ?

ccordoba12:
Yep, that's true too :-)

blink1073:
Fixed!

screen shot 2015-04-27 at 5 42 11 pm

goanpeca:
I have no idea what that means! but looks awesome

goanpeca:
How are we going to localize the name of the shortcuts?

spyderlib/plugins/shortcuts.py
```diff
@@ -159,17 +63,12 @@ def data(self, index, role=Qt.DisplayRole):
         column = index.column()
         if role == Qt.DisplayRole:
             if column == CONTEXT:
-                return to_qvariant(shortcut.context)
+                return to_qvariant(_(shortcut.context.capitalize()))
```
ccordoba12: I don't know if this is enough to get things localized.

Don't we need to pass a real string to `_(...)` instead of something computed from a method/function, so that `gettext` grabs that string for translation?

I say it because I haven't seen any example in our sourcecode of `_(...)` applied to something that's not a string :-)

goanpeca: Yep, I think this wont do, and we will need to add a tuple or something to the shortcuts list and fix where needed in `config.py`?

```python
...
            ('shortcuts',
             {
              # ---- Global ----
              # -- In spyder.py
              '_/close pane'): ("Shift+Ctrl+F4", _('Close pane')),
              '_/preferences'): ("Ctrl+Alt+Shift+P", _('Preferences')),
...
```

blink1073: Is `_` not just a function call, where we pass in the *result*, which is the computed string?

goanpeca: Not really... the gettext scripts when ran, go through the source code and grab any text and create the .po files. This process is done offline every time we need to update the translations. On runtime the _() will fetch the corresponding translated text...

blink1073: So they need to be capitalized in `config.py` then.

blink1073: Seems like that should be a separate PR, non?

goanpeca: Yep, I think is better for another PR, plus not sure what would be the best way @ccordoba12 to implement it...

ccordoba12: Yes, sorry for the noise. Let's leave localization for another PR

Nodd: Only the modifiers are to be localized, right ? In that case the string should be parsed, each modifier translated, then the string would be rebuilt and displayed. I, don't see us providing a translation for every existing possible keybind.

goanpeca: @Nodd I was thinking of providing localization for the `Context` and the `Name` column only

goanpeca:
@blink1073 I am trying this on win7 py2.7 and there is something not working. 

If I do not use any modifier it works great, but if I enter a shortcut that uses a modifier (for instance `Ctrl+A`) I see the shortcut, but as soon as I release the `Ctrl` Key, the shortcut blinks and disappears and an empty table cell... If I close and open again, the table cell remains empty...

blink1073:
My Windows laptop is near dead, @goanpeca, can you try with the new outfitted debug statements and tell me what you find?

goanpeca:
So I opened the preferences and took the first shortcut for testing....

Here is what I get when pressing (I put some comments in between what was printed by the debug_print):

```bash
# Pressed Ctrl (and hold)
key 16777249, npressed: 1 

# Pressed and released F12 (still holding Ctrl)
key 16777275, npressed: 2 
decrementing
keys: set([83886139])

# Release Ctrl
keys: set([])                           
````

**Comment by blink1073** [comment]
Hmm, `self.keys` is getting cleared somewhere.  Can you add `debug_print` in the places where it is cleared?

**Comment by goanpeca** [comment]
Ok I checked now... it seems to be related to the Ctrl key only.

`Shift+F12` works
`Alt+F12` works
`Ctrl+anything`... nope


```
# Press Ctrl
keyPressEvent key 16777249, npressed: 1
# Press and release F12
keyPressEvent key 16777275, npressed: 2
keyPressEvent: decrementing
keyReleaseEvent keys clear second if: set([83886139])
# Release Ctrl
keyReleaseEvent keys clear second if: set([])
```

**Comment by goanpeca** [comment]
@blink1073 

Why decremented only when it is a Ctrl modifier?

``` python
        if modifiers & Qt.ControlModifier:
            key += Qt.CTRL
            self.npressed -= 1
            debug_print('keyPressEvent: decrementing')
```

If I comment `self.npressed -= 1` it works...

goanpeca:
@blink1073 I notice that only shift and alt allow to put sequence of commands, whereas Ctrl does not allow for that... is this desired/normal?

goanpeca:
@blink1073 this is the "fix" I found for the moment, but I am not sure if it results in the desired behavior... 
```python
    def keyReleaseEvent(self, e):
        if self.currentIndex().column() != SEQUENCE:
            debug_print('keyReleaseEvent before clear first if: keys: %s' % self.keys)
            self.keys = set()
            self.npressed = 0
            return

        self.npressed -= 1
        debug_print(self.npressed)

        if self.npressed <= 0:
            if self.keys:
                keySequence = QKeySequence(*self.keys)
                self.model.setData(self.currentIndex(),
                                   keySequence.toString())
            debug_print('keyReleaseEvent before clear second if: %s' % self.keys)
            self.keys = set()
            self.npressed = 0
```

ccordoba12:
@goanpeca, you could make a PR against @blink1073 PR with your changes :-)

blink1073:
@goanpeca, that should fix it.  It looks like the decrement was only needed on OSX (but I'll have to test in Linux to be sure).

blink1073:
I tried it on Win7 and we can now use multi-key Ctrl sequences as well.

goanpeca:
@blink1073 I am not so sure now, that removing that for windows is the solution...

Also... I think we should add the extra if in keyReleaseEvent

```python
            if self.keys:
                keySequence = QKeySequence(*self.keys)
                self.model.setData(self.currentIndex(),
                                   keySequence.toString())
```
So that it only updates the sequence if there is an actual sequence... also, it avoids that pressing a single shift or tab or Ctrl erases the current sequence in the table cell.

blink1073:
Done.

goanpeca:
Now it works like a charm :)

goanpeca:
Working as well on ubuntu :+1:

blink1073:
Thanks @goanpeca, I owe you a beverage of choice.

goanpeca:
:+1: not [this one](https://www.youtube.com/watch?v=mxyGGKWGV70)

Nodd:
When changing a shortcut, I get 
```
  File "/data/prog/blink1073_spyder.git/spyderlib/plugins/shortcuts.py", line 228, in <lambda>
    lambda i1, i2, opt='': self.has_been_modified(opt))
  File "/data/prog/blink1073_spyder.git/spyderlib/plugins/configdialog.py", line 376, in has_been_modified
    self.changed_options.add(option)
TypeError: unhashable type: 'list'
```
But the text is the Sequence column is correctly changed and the new shortcut works.

I find the way to change shortcut as implemented right now too easy. By that I mean that only clicking on a Sequence cell to modify the shortcut can easily lead to accidental changes. A safer way to do it is to have the user double-click anywhere on the line, it opens a dialog which records a unique shortcut, with pushbuttons to validate the modification or to cancel it. This is how it's done Geany or Compiz for example.
The advantage is that you can display other things in the dialog, such as information if the new shortcut was already set somewhere else, and to what action.

Also as in #2217, `Meta + something` is recognized as a shortcut but  it doesn't work in the window afterwards.

goanpeca:
@blink1073, actually I agree with @Nodd suggestiosn, is to easy to "mess up", it would be better that a double click, or an enter in the line of the table and then in this new shortcut editor dialog (something small and modal) you get to enter the actual shortcut plus display info on possible duplicates.

I think we should also allow for finding a shortcut by typing something. If something is typed and it is a letter then a little box with the typed text could appear and if it is a valid entry in a shortcut point to it.... maybe even include some fuzzy search in it. But of course this could go to another PR.

blink1073:
I'm a bit overwhelmed at the moment, do either of you want to take that on as a PR against mine?

ccordoba12:
@goanpeca, would you mind to help @blink1073 with Nodd's suggestions? I also agree with him :-)

goanpeca:
Sure, how do I make a PR against this PR ?

@Nodd can you post a screenshot of how this dialog looks in the editors you mentioned?

ccordoba12:
@goanpeca, you need to:

1. Fork @blink1073's fork of Spyder using Github
2. Clone this fork to your computer
3. Checkout this branch
4. Create a new branch and commit your work to this new branch 
5. Push your work to your fork of @blink1073's fork
6. Create your PR using GH interface

I think that's all. The main idea is to treat @blink1073's fork as if it were the main Spyder repo :-)

goanpeca:
Got it

goanpeca:
So the user can change the shortcut by double clicking or pressing enter on any column of the row 

Ok... basic sketch... give comments please @spyder-ide/core-developers 

image

Nodd:
@goanpeca That's exactly the kind of dialog I was talking about.
A few suggestions:
* Replace "Enter" by "Press" to be more explicit on how to set the shortcut
* The new shortcut should be made more visible. Maybe put the blank line after the current shortcut (and add parenthesis around it).

blink1073:
I concur with @Nodd.

goanpeca:
Thanks for the comments

@blink1073 and @Nodd  is there any key we can use to toggle between entering the shortcut and toggling the Ok and Cancel buttons? it defeats the purpose of working with the keyboard only if you need to use the mouse to accept a given shortcut.

Can we capture Esc maybe to toggle between entring shortcut or selecting buttons? is Esc commonly used as a shortcut ?

blink1073:
What's wrong with a raw "tab" followed by "return"?

goanpeca:
Wrong? what do you mean?

blink1073:
What I mean is that "tab" can't be a defined shortcut, but is the default way to toggle between dialog items.

goanpeca:
Got it, will fix now (I had to disable the focus policy of the buttons so that the down arrows would not give them focus, and hence Tab stops working as well..

blink1073:
Yeah you'll have to detect a standalone "tab" specifically.

goanpeca:
Got it working  :-) 

Should we provide some help text to inform about this behavior? or add the Info button (like the array builder or the sublime filde dialog PR?)

goanpeca:
Ok this is how it looks now... `Tab` is used to switch between modes... (Edit, NonEdit)

### Edit mode: Nothing entered or if using the same shortcut
image

### Edit mode: Shortcut entered
image

### Non Edit mode: Nothing entered or if using the same shortcut
image

### Shortcut entered:
image

goanpeca:
@blink1073 I think there are some problems with using Tab + Shift

For instance if I enter `Shit+Tab` the sequence is regsitered as `Shift+Backtab`

blink1073:
@goanoeca are you wanting to use that in tge dialog or saying it is displayed wrong?  Because I am just telling it to print the actual name of the sequence on your OS.

goanpeca:
I am saying that for instance the current `Go to next line` shortcut is `Ctrl+Shift+Tab` which is an impossible to enter sequence cause if I try to enter that sequence the result is

`Ctrl+Shift+Backtab`, 

image

blink1073:
You need to display the current shortcut the same way as the new shirtcut, using the Native text.

goanpeca:
This PR has the same problem... it has nothing to do with the new changes I am doing

goanpeca:
image

blink1073:
But *that* is the Proper Qt Native representation of the sequence. I'm saying update your "current" sequence to match.

blink1073:
Or does it show up differently before you change it?  If so, there's not much to do about it, as long as the key sequence itself works.

blink1073:
Or if it vexes you do a .replace() for this instance.

goanpeca:
Ok once again...

I open Spyder and this is how it looks after a reset
image

Now I try to change it to use the same sequence and I get this
image

So it is not the same

blink1073:
Does the shortcut still work in the editor?

goanpeca:
The first one (after a reset works)

The second does not

blink1073:
Well poop.

goanpeca:
:poop:?

blink1073:
It looks like I need to intercept this sequence and divert it.

goanpeca:
(I am on win) On mac you do not see this?

blink1073:
Or more generally switch Backtab to Tab key.  I'm on my phone, I haven't actually woken up for the day.  :smile:  Yes, I answer emails at 2am.

goanpeca:
Every sequence will apparently use the modified... so if you press `Ctrl+Shift+[`, it will get stored as `Ctrl+Shift+{`

Go to bed dude...

blink1073:
No this is actually my wake up time.

goanpeca:
why?

blink1073:
No I meant it was 2am when we started this convo, its 7am now.

blink1073:
To be clear, you tried using the altered shortcut in the editor and it did not toggle between files?  Because it worked for me.

goanpeca:
It does not work for me after altered

goanpeca:
The is a windows thing most likely

blink1073:
Try commenting out the `key += Qt.SHIFT` line, maybe we need to do that on Windows.

goanpeca:
If I do that no shift appears in shortcuts

blink1073:
But does the shortcut work at that point?

goanpeca:
from the [Qt website](http://doc.qt.io/qt-4.8/qheaderview.html#details)
> Note that, for letters, the case used in the specification string does not matter. In the above examples, the user does not need to hold down the Shift key to activate a shortcut specified with "Ctrl+P". However, for other keys, the use of Shift as an unspecified extra modifier key can lead to confusion for users of an application whose keyboards have different layouts to those used by the developers. See the Keyboard Layout Issues section below for more details.

blink1073:
Ugh I need to step away from this one for a bit, I'm too frustrated.

goanpeca:
I think I know how to fix it

blink1073:
Buena suerte

goanpeca:
@blink1073  question...

Can we safely assume that a sequence of several shortcuts:

1. Can only be made if one or more modifiers are pressed
1. All the subsequences WILL use the same modifiers?

blink1073:
I think so, from an emacs perspective.

goanpeca:
Any other perspective you can think of?

blink1073:
No, emacs is the only context I hace for complex sequences

goanpeca:
Ok, I will try something and let you know.

goanpeca:
Ok, so for now the only thing to do is to change the sequence to Tab in case we get Backtab... otherwise I do not see any other use case...

blink1073:
What about `Ctrl+Shift+}` and friends?

goanpeca:
Those suffer from the same problem... whenever the shift is pressed the key that arrives t the keyPressEvent is already the modified key (if shift was pressed)

goanpeca:
in my laptop 

`Ctrl+Shift+{` is fine, cause `{` is a key that needs shift (on top of `[`), but if I want to make a shortcut like

`Ctrl+Shift+[` then same problem...

goanpeca:
this issue is present on any key that has a modified version (with shift), so all numbers on top of letters, and all the other `[ ] \ ; ', . / `

goanpeca:
An tab

goanpeca:
Ideas @blink1073?

blink1073:
Might have to go back to the drawing board and use `text()` instead of `key()`. 
http://doc.qt.io/qt-4.8/qkeyevent.html#key

goanpeca:
But what is the expected (right/correct) behavior when using a shortcut that includes shift and uses a letter that has a shifted version?

blink1073:
I'd say it matters more whether it works or not than what it looks like.

blink1073:
If it has to look like `Ctrl+Shift+{` to work, I say so be it.

goanpeca:
My concern is not about looks, is about works as well, but Qt itself seems to be inconsistent...

blink1073:
Yep...

goanpeca:
But text should solve it, I can take care of it, otherwise if you push something I will have to rebase... I made a lot of moves...!

blink1073:
Be my guest my friend.

goanpeca:
More "good news"

If I type `Ctrl+Alt_Shift+O` I get...

image

Cause.. Ctrl+Alt+Key produces a modified character form the extended ascii set

blink1073:
Aren't shortcuts a blast?

blink1073:
This whole debacle is making me want to have the user type out the sequence instead and then do validation on it.

blink1073:
So actually type `S h i f t + T a b`

blink1073:
And then try to run `QKeySequence` on it and see if it barfs.

goanpeca:
I think is a windows only thing....
there is still hope relax ;-)

blink1073:
I still see the visual quirks on Mac, it is just a bit more forgiving on the actual usage.

blink1073:
But, it *is* possible on Mac to do it right, for instance in the System Keyboard preference, if I type `Ctrl+Shift+Tab`, I actually get those three characters printed.

blink1073:
Same for the `. { [` type keys.

goanpeca:
I have fixed the tab issue ad all the keys that have a shift option, so that is ok now

blink1073:
You da man.

blink1073:
So what's left, the unicode sequences?

goanpeca:
Yep, but i have an idea.. Will try it later :smile_cat:

blink1073:
you are a Qt wizard.

blink1073:
A slightly lesser tier perhaps than @ccordoba12.  :smile:

goanpeca:
http://en.wikipedia.org/wiki/AltGr_key

>Therefore, it is recommended that this combination not be used as a modifier in Windows keyboard shortcuts as, depending on the keyboard layout and configuration, someone trying to type a special character with it may accidentally trigger the shortcut,[5] or the keypresses for the shortcut may be inadvertently interpreted as the user trying to input a special character.

ccordoba12:
Great work @goanpeca!! Just a simple suggestion to avoid further problems: why not forbid some shortcut combinations? That's what the old code did to not run into the problems you're seeing right now :-)

goanpeca:
@ccordoba12 we could do that, but still it would cause problems on windows... yes windows is always a source of problems. 

I have a working prototype (still some rough edges) plus now we have fuzzy shortcut search by name :-)

goanpeca:
hopefully tomorrow I can push it so blink can do some mac testing

goanpeca:
Ok, so turns out it is maybe just easier to select a US keyboard with US layout (not the international setting) otherwise we would need to:

1. Detect the current keyboard layout somehow (I havent found yet how to do that)
1. Have a mapping of special keys (Ctrl+Alt) keys in that specific layout, (I tried this for US English (International) and it works.)
1 . With the previous two you could correct any "weird" input and make things work as expected..

But we need to meditate on this a bit I guess.... but I would not like to go back as far as having to type the actual shortcut manually.

blink1073:
I don't think we need to get too fancy.  If we can support the existing shortcuts and the Emacs ones, I'd say we're good.

goanpeca:
Ok @blink1073 as @ccordoba12 suggests we should take a special treatment on windows and that is:

- Have a list of allowed keys to use when having shift pressed... so if shift is a modifier only accept, letters, home, f# etc etc... only. This will apply if the user is trying to input a Ctrl+Shift+Alt+ or Ctrl+Shift +or Shift+ or Alt+Shift+
- Take care of the tab key as it seems to be one special key we need, that is, always make it a Tab (no a Backtab)
- Do not accept Ctrl+Alt combinations with anything outside keys we are certain will not have an AltGr (like page down, page up etc)
- Do not accept any Ctrl+Shift+Alt combination AT ALL!

What do you think?

blink1073:
:+1: to all



—

goanpeca:
Ok this is working much better like this :smile:, now.. which keys should be the only ones allowed to be a shortcut by themselves besides the `F1`, `F2`, etc keys?

blink1073:
Delete and Escape?

Nodd:
Escape will be needed for vim keybinds.

goanpeca:
Ok, added, if you think of anything else let me know :)

goanpeca:
Hi @spyder-ide/core-developers so this is how the new shortcut finder looks, I used the same delegate @d1manson used for his sublime finder, so maybe he should add that to the helper widgets file as well.

If you start typing inside the shortcuts table, a transparent lineedit will appear in on top of the name header, and I included the " fuzzy"  search @blink1073 suggested in #2340. 

Using Up/Down arrows allows to select the different filtered shortcuts and pressing enter opens the shortcut editor, or escape closes the filter.

This is how the filter in action looks
image

Shortcut editor
image


A small [video](http://youtu.be/igm7mktAJG4) which does not display the shortcut finder because of how the software I use records the screen, apparently it does not get the modal dialogs

I still have some fixes/code cleaning to do and maybe display a message or warning icon with an explanation of why certain icons are not recorded (like single letters, or KeySequences of more than 4 sequences) but I guess this is almost done :)

Suggestions?

blink1073:
Excellent, I'll add #1940 to the list of issues closed by this PR.  You rock @goanpeca!

goanpeca:
Now I added conflict check withing the shortcut editor so you get the following if using a shortcut already taken.

## Conflict
image

## No conflict
image

@ccordoba12 can you take a look at the [PR of the PR](https://github.com/blink1073/spyder/pull/1) for comments

Nodd:
The icons are great but they do not show what will spyder do in case of conflict. Will it swap shortcuts? define both (then crash)? erase the first one?
I think it's better to shown something only if there is a conflict, and write down the full sentence instead of displaying a tooltip on hover only. It's not like there is no room left in the dialog.

goanpeca:
@Nodd 

1. This is at the shortcut editor level, if you press ok, then you still get a warning when trying to accept in the preferences dialog
1. The tooltip pops up automatically if you enter a conflicting shortcut.

ccordoba12:
@goanpeca, this is terrific work!!

@blink1073, what do you think about merging @goanpeca's PR as it is and reviewing everything here?

blink1073:
Merged!

blink1073:
@goanpeca, why can't we release focus from the search widget without hitting escape?  Is that easy to fix?

blink1073:
Also, should we show the shortcut as it is being entered?

goanpeca:
@blink1073  sure, i made it modal, but this can be changed. Also what would be the expected behavior? And if you type somethin it should add to what is already in tje filter¿

For entering the shortcut as it is being typed I think will be harder because of how things are structured.

blink1073:
@goanpeca, I was thinking that if you clicked anywhere else it should close the `find` box.
I'll take a look at showing partial shortcuts.

goanpeca:
It should close and erase the filter and display all shortcuts again? That would lead to unexpected behaviors if you use the filter but then want to use the mouse to click one of the filterred shortcuts. I would say displaying partial shortcuts is not worth it.

Nodd:
This "hidden" search bar seems not very discoverable. What if there was only one search bar at the top, for all columns ?

goanpeca:
@Nodd, yes that would be the easiest to implement right :+1:

goanpeca:
@blink1073 I will make a PR against your PR again to address the changes to the search functionality.

The idea is to have a fixed widget on top of the shortcuts table with a label and a textbox so that its is permanently there in case the user want to find something. typing on the table will redirect to the finder.

What do you think?

blink1073:
:+1:

goanpeca:
Ok, so this is the new look of the shortcuts dialog. I moved the finder to a fixed position on top, so now it is easier to find and use

image

goanpeca:
@ccordoba12 this one should be ready to merge now :-)

goanpeca:
I guess this needs a rebase before merging @blink1073

blink1073:
The rebase is a huge mess, something is very messed up with this branch.  I am also not sure what changes should actually be there.

blink1073:
Rebase is done...I always seem to back myself into a corner with Spyder PRs, I never have this problem on other repos...

goanpeca:
Great what's next :smile:

Nodd:
@blink1073 we tend to make quite big PRs, so that's both hard to review and to merge. Now sometimes it is necessary…

goanpeca:
@ccordoba12 can you please take a look at the code on this PR?, Afterwards if @blink1073 is too busy I can try to make a PR against his PR to correct anything.

goanpeca:
@blink1073 , this needs a rebase again due to @SylvainCorlay PR on icons?

blink1073:
This will have to wait until the weekend, I have an eye infection and can barely see.  



—

goanpeca:
@Nodd, you like nitpicking ;-), want to give this PR a code review? (we will change the location of the finder as suggested when some code review starts to take place :-) )

blink1073:
@goanpeca, do you mind making a PR against this to move the search bar to the bottom?

goanpeca:
ok

blink1073:
@goanpeca, you don't need to do a full PR, perhaps just point me in the right direction.

spyderlib/plugins/shortcuts.py
```diff
@@ -187,186 +512,274 @@ def headerData(self, section, orientation, role=Qt.DisplayRole):
...
+        self.reset_btn = QPushButton(_("Reset to default values"))
+
+        # Layout
+        hlayout = QHBoxLayout()
         vlayout = QVBoxLayout()
+        hlayout.addWidget(self.label_finder)
+        hlayout.addWidget(self.finder)
+        vlayout.addLayout(hlayout)
         vlayout.addWidget(self.table)
-        reset_btn = QPushButton(_("Reset to default values"))
```
goanpeca: @blink1073 just change the order of this

from 

```python
vlayout.addLayout(hlayout)
vlayout.addWidget(self.table)
```
To this

```python
vlayout.addWidget(self.table)
vlayout.addLayout(hlayout)
```

spyderlib/plugins/shortcuts.py
```diff
@@ -187,186 +512,274 @@ def headerData(self, section, orientation, role=Qt.DisplayRole):
...
-        reset_btn.clicked.connect(self.reset_to_default)
-        vlayout.addWidget(reset_btn)
+        vlayout.addWidget(self.reset_btn)
         self.setLayout(vlayout)
-        
+
+        self.setTabOrder(self.finder, self.table)
+        self.setTabOrder(self.table, self.reset_btn)
+
```
goanpeca: @blink1073 and also here:

from

```python
self.setTabOrder(self.finder, self.table)
self.setTabOrder(self.table, self.reset_btn)
```

To this

```python
self.setTabOrder(self.table, self.finder)
self.setTabOrder(self.finder, self.reset_btn)
```

blink1073:
Thanks @goanpeca, the new look:

screen shot 2015-06-15 at 6 41 53 am

d1manson:
The fuzzy searching should perhaps be accompanied by a fuzzy-sorting rule such as:

Sort by the number of gaps found between requested letters, i.e. in the example pictured
**c**k un**lo**ck pane**s** would be assigned 2, `= len( ["k un", "ck pane"])`, whereas the other two matches would be assigned 0.

You could obviously use some slightly different variation on this rule, but this would probably suffice.  
The advantage of this is that if you want to search for a part of a word or a short word like "up", the most relevant stuff will be kept at the top of the list.  Maybe this isn't that important here, but if fuzzy searching is going to become a common feature of the IDE then it's worth having a rule such as this I would think.

goanpeca:
Thanks for the input @d1manson.

@ccordoba12 or @Nodd nitpicking is welcome now :)

goanpeca:
The fuzzy for now is just very very basic, this PR is already too big. 

We can eventually use the library @ccordoba12 proposed to make the fuzzy search more fancy.

spyderlib/plugins/shortcuts.py
```diff
@@ -187,186 +512,274 @@ def headerData(self, section, orientation, role=Qt.DisplayRole):
...
+    def keyPressEvent(self, event):
+        """Qt Override"""
+        key = event.key()
+        if key in [Qt.Key_Enter, Qt.Key_Return]:
+            self.show_editor()
+        elif key in [Qt.Key_Tab]:
+            self.parent().reset_btn.setFocus()
+        elif key in [Qt.Key_Backtab]:
+            self.finder.setFocus()
+        elif key in [Qt.Key_Up, Qt.Key_Down, Qt.Key_Left, Qt.Key_Right]:
```
goanpeca: @blink1073 
I think we also need to adjust the code here as well from:

```python
        elif key in [Qt.Key_Tab]:
            self.parent().reset_btn.setFocus()
        elif key in [Qt.Key_Backtab]:
            self.finder.setFocus()
```

to

```python
        elif key in [Qt.Key_Tab]:
            self.finder.setFocus()
        elif key in [Qt.Key_Backtab]:
            self.parent().reset_btn.setFocus()
```

blink1073: Fixed!

goanpeca:
@ccordoba12 you can have fun now reviewing this one ;-)

spyderlib/plugins/shortcuts.py
```diff
@@ -7,179 +7,504 @@
...
                                  iter_shortcuts, reset_shortcuts)
 from spyderlib.plugins.configdialog import GeneralConfigPage
-from spyderlib.py3compat import to_text_string, is_text_string
+from spyderlib.utils.qthelpers import get_std_icon
+from spyderlib.widgets.arraybuilder import HelperToolButton
 
+MODIFIERS = {Qt.Key_Shift: Qt.SHIFT,
+             Qt.Key_Control: Qt.CTRL,
+             Qt.Key_Alt: Qt.ALT,
+             Qt.Key_Meta: Qt.META}
```
ccordoba12: Why is this mapping necessary?

spyderlib/plugins/shortcuts.py
```diff
@@ -7,179 +7,504 @@
...
-        self.modifiers = sorted(modifiers)
-        assert key in self.KEYS
-        self.key = key
+             ["Colon", "Semicolon", "Less", "Equal", "Greater",
+              "Question", "At"] + [chr(_i) for _i in range(65, 91)] + \
+             ["BracketLeft", "Backslash", "BracketRight", "Underscore",
+              "Control", "Alt", "Shift", "Meta"]
+VALID_SINGLE_KEYS = [getattr(Qt, 'Key_{0}'.format(k)) for k in SINGLE_KEYS]
+VALID_KEYS = [getattr(Qt, 'Key_{0}'.format(k)) for k in KEYSTRINGS+SINGLE_KEYS]
+VALID_ACCENT_CHARS = "ÁÉÍOÚáéíúóàèìòùÀÈÌÒÙâêîôûÂÊÎÔÛäëïöüÄËÏÖÜñÑ"
```
ccordoba12: Is it a good idea to accept accent chars as shortcut keys? @SylvainCorlay, what do you think?

goanpeca: This is something I ledt unfinished... but it is to be used in the search box of the... and since at some point we will have localization of the shortcut names, it made sense to include this... but probably there is a better way of doing this

spyderlib/plugins/shortcuts.py
goanpeca: Not sure, I think @blink1073 did this ? ?

spyderlib/plugins/shortcuts.py
```diff
@@ -7,179 +7,504 @@
...
+        self.button_cancel.setFocusPolicy(Qt.NoFocus)
+        self.helper_button.setToolTip('')
+        self.helper_button.setFocusPolicy(Qt.NoFocus)
+        style = """
+            QToolButton {
+              margin:1px;
+              border: 0px solid grey;
+              padding:0px;
+              border-radius: 0px;
+            }"""
```
ccordoba12: Does this style works well on all OSes?

spyderlib/plugins/shortcuts.py
ccordoba12: Ok, nice to know! So this is used only to filter shortcut names.

Then we can leave it, but perhaps adding a comment about its meaning above it

SylvainCorlay: Letters like 'é' are directly accessible on French keyboard. Hence it probably makes sense to allow them.

goanpeca: Of course this will break if someone starts doing weirder scripts...

spyderlib/plugins/shortcuts.py
goanpeca: I have no idea on mac... but Windows and Linux works fine

goanpeca: But I have been using this style in other widgets now... (like the array builder) so maybe we should have a centralized point for styles (that are used in several places?)?

goanpeca:
# Rough edge
On this image from @blink1073, the color of the text should be white... I thought I had fixed that... at least in linux it looks ok

screen shot 2015-06-15 at 6 41 53 am

spyderlib/plugins/shortcuts.py
```diff
@@ -7,179 +7,505 @@
...
 
-from spyderlib.qt.QtGui import (QVBoxLayout, QComboBox, QItemDelegate,
-                                QTableView, QMessageBox, QPushButton)
-from spyderlib.qt.QtCore import Qt, QSize, QAbstractTableModel, QModelIndex
+from spyderlib.qt.QtGui import (QVBoxLayout, QTableView, QMessageBox,
+                                QPushButton, QKeySequence, QDialog,
+                                QDialogButtonBox, QLabel, QGridLayout,
+                                QLineEdit, QAbstractItemView,
+                                QSortFilterProxyModel, QStyledItemDelegate,
+                                QStyleOptionViewItemV4, QApplication,
```
ccordoba12: It seems there's nothing named `QStyleOptionViewItemV4` in Qt5. There is only `QStyleOptionViewItem`, so this PR is failing on Qt5.

spyderlib/plugins/shortcuts.py
ccordoba12: We have a qss stylesheet for Mac already. But let's leave that for another PR :-)

spyderlib/plugins/shortcuts.py
```diff
@@ -7,179 +7,505 @@
...
+        # Ctrl, Shift, Alt, Meta.
+        if (key == Qt.Key_Control or
+                key == Qt.Key_Shift or
+                key == Qt.Key_Alt or
+                key == Qt.Key_Meta):
+            return
+
+        # Check if valid keys
+        if key not in VALID_KEYS:
+            return
```
ccordoba12: This is not working for *dead* keys (as @SylvainCorlay discovered). I mean, dead keys are the ones that need a further key stroke to make them appear, like `<tilde> e` to get `é`

So right now you can enter shortcuts like:

    Ctrl + <tilde> + e

and it'll be accepted (although they do nothing in practice)

spyderlib/plugins/shortcuts.py
```diff
@@ -7,179 +7,505 @@
...
+
+        modifiers = e.modifiers()
+        if modifiers & Qt.ShiftModifier:
+            key += Qt.SHIFT
+        if modifiers & Qt.ControlModifier:
+            key += Qt.CTRL
+            if sys.platform == 'darwin':
+                self.npressed -= 1
```
ccordoba12: Why is this **not** counted as a keystroke?

ccordoba12: Just curious, I just saw that this is used for debugging

spyderlib/plugins/shortcuts.py
```diff
@@ -7,179 +7,505 @@
...
+        """Key release event for non-edit state."""
+        key = e.key()
+
+        if key == Qt.Key_Escape:
+#            self.close()
+            pass
+
+        if key in [Qt.Key_Left, Qt.Key_Right, Qt.Key_Up,
+                   Qt.Key_Down]:
+
```
ccordoba12: Please remove this blank

spyderlib/plugins/shortcuts.py
goanpeca: Hmm, ok I will check this then

spyderlib/plugins/shortcuts.py
ccordoba12: By the way, I don't like this too much. I think just `Up` should take users again to the QLineEdit, not the other keys.

spyderlib/plugins/shortcuts.py
```diff
@@ -7,179 +7,505 @@
...
+            if self.button_ok.hasFocus():
+                self.button_cancel.setFocus()
+            else:
+                self.button_ok.setFocus()
+
+    def keyReleaseEvent(self, e):
+        """Qt override"""
+        self.npressed -= 1
+        if self.npressed <= 0:
+            if len(self.keys) == 1 and list(self.keys)[0] == Qt.Key_Tab:
```
ccordoba12: I think the most natural to move out of the QLineEdit is `Esc`, so I'd prefer to use it instead of `Tab` here. Or maybe we can use both :-)

That'd prevent having `Esc` as a shortcut too, which is nice in my opinion.

spyderlib/plugins/shortcuts.py
```diff
@@ -7,179 +7,505 @@
...
+
+        if conflicts:
+            tip_title = _('The new entered shorcut conflicts with:') + '\n'
+            tip_body = ''
+            for s in conflicts:
+                tip_body += ' - {0}: {1}\n'.format(s.context, s.name)
+            tip = '{0}{1}'.format(tip_title, tip_body)
+            widget.setIcon(get_std_icon('MessageBoxWarning'))
+            widget.setToolTip(tip)
+            QToolTip.showText(widget.mapToGlobal(QPoint(0, 5)), tip)
```
ccordoba12: Unfortunately tooltips disappear immediately if the mouse is above them, and so users can't read their contents.

So I'd suggest to show a QLabel below the new shortcut QLineEdit with the `tip_title` message above. That should be a much better UX than showing a tooltip

spyderlib/plugins/shortcuts.py
ccordoba12: On a second thought, `Esc` should also reset the QLineEdit, so that users can enter a different shortcut, if the one they want to use is already taken (or they change their minds after entering a particular shortcut).

I say it because right now there's no way to introduce a different shortcut other than the initial one. You need to close the dialog and start a new one if you want to do that :-)

spyderlib/plugins/shortcuts.py
goanpeca: @blink1073 can you take care of this?, it is very annoying to make a PR against your PR to fix all the "details"...

spyderlib/plugins/shortcuts.py
```diff
@@ -7,179 +7,505 @@
...
+        else:
+            self.text_new_sequence.setEnabled(True)
+            self.text_new_sequence.setFocus()
+
+    def nonedit_keyrelease(self, e):
+        """Key release event for non-edit state."""
+        key = e.key()
+
+        if key == Qt.Key_Escape:
+#            self.close()
```
ccordoba12: Why is this commented? I think we should use it, i.e. if the QLineEdit is not focused, `Esc` should close the dialog.

spyderlib/plugins/shortcuts.py
goanpeca: You can press `tab` and then click `cancel`  or just `tab` `esc`... I think `Esc` can be used as a shortcut under some contexts?

goanpeca:
Could we perhaps merge it as it is? I can make a new PR to address the pending issues, I feel at this point most of the feedback relates to the changes I introduced so it would make life easier for @blink1073 and myself...

spyderlib/plugins/shortcuts.py
ccordoba12: Yeah, I understand that. But I'd prefer for us to use `Esc` instead of `Tab` here, and I wouldn't like to see `Esc` as a shortcut.

So the behavior for `Esc` should be:

1. If QLineEdit is focused, then call `nonedit_keyrelease` (as it's done here for `Tab`) and reset QLineEdit.
2. If QLineEdit is unfocused, just close the dialog.

ccordoba12:
Ok, @blink1073 says he doesn't have a problem with that. So let me finish reviewing this PR, and then you can create a new one to address my comments.

spyderlib/plugins/shortcuts.py
ccordoba12: I don't agree with @SylvainCorlay, that's too much work to support it. We should only support ascii letters for now, and if people ask more, we could try to improve the situation.

spyderlib/plugins/shortcuts.py
```diff
@@ -7,179 +7,505 @@
...
+            if self.key_non_modifiers[0] not in VALID_SINGLE_KEYS:
+                return
+            # Filter
+            elif len(self.key_non_modifiers) > 1:
+                return
+
+        # QKeySequence accepts a maximum of 4 different sequences
+        if len(self.keys) > 4:
+            # If the user enters more than 4, take the first 4 only
+            self.keys = set([0, 1, 2, 3])
```
ccordoba12: I don't agree with this behavior. I think we shouldn't allow a shortcut in this case.

goanpeca: So if an user enter a sequence of 4 and bigger nothing should happen ???

spyderlib/plugins/shortcuts.py
```diff
@@ -7,179 +7,505 @@
...
-        self.key = keystr2key(get_shortcut(self.context, self.name))
-    
+        self.key = get_shortcut(self.context, self.name)
+
     def save(self):
-        set_shortcut(self.context, self.name, str(self.key))
+        set_shortcut(self.context, self.name, self.key)
+
 
+CONTEXT, NAME, SEQUENCE = list(range(3))
```
ccordoba12: @blink1073, simple change here: `[0, 1, 2]`, which is more readable :-)

spyderlib/plugins/shortcuts.py
```diff
@@ -7,179 +7,505 @@
...
+        self.text_color = palette.text().color().name()
+        self.text_color_highlight = palette.highlightedText().color().name()
+
+    def current_index(self):
+        """Get the currently selected index in the parent table view."""
+        i = self._parent.proxy_model.mapToSource(self._parent.currentIndex())
+        return i
+
+    def _enrich_text(self, text, color):
+        """Highlight (bold) letters from the shortcut finder."""
```
ccordoba12: This is really nice! Pretty good job here ;-)

spyderlib/plugins/shortcuts.py
ccordoba12: I mean, instead of taking part of it

goanpeca: ok

spyderlib/plugins/shortcuts.py
goanpeca: :smile:

spyderlib/plugins/shortcuts.py
```diff
@@ -187,186 +513,274 @@ def headerData(self, section, orientation, role=Qt.DisplayRole):
...
-            elif section == MOD1:
-                return to_qvariant(_("Mod1"))
-            elif section == MOD2:
-                return to_qvariant(_("Mod2"))
-            elif section == MOD3:
-                return to_qvariant(_("Mod3"))
-            elif section == KEY:
-                return to_qvariant(_("Key"))
+            elif section == SEQUENCE:
+                return to_qvariant(_("Sequence"))
```
ccordoba12: I think this column should be called `Shortcut`, not `Sequence`

spyderlib/plugins/shortcuts.py
```diff
@@ -187,186 +513,274 @@ def headerData(self, section, orientation, role=Qt.DisplayRole):
...
-            elif column == KEY:
-                key.key = Key.key_from_str(text)
+            if column == SEQUENCE:
+                shortcut.key = text
             self.dataChanged.emit(index, index)
             return True
         return False
 
+    def update_search_letters(self, text):
+        """ """
```
ccordoba12: Missing docstring

spyderlib/plugins/shortcuts.py
```diff
@@ -187,186 +513,274 @@ def headerData(self, section, orientation, role=Qt.DisplayRole):
...
+        self.letters = text
+        self.reset()
+
+    def update_active_row(self):
+        """ """
```
ccordoba12: Missing docstring

spyderlib/plugins/shortcuts.py
goanpeca: Agreed :)

ccordoba12:
Ok, good to go, on Qt5 too!

PR #2369, opened 2015-04-26
Add drag support for dockwidgets sharing same position

goanpeca:
## Description

This related to dockwidgets that are tabified. Dockwidgets, when tabified, are put inside a QTabBar, but there is no access method to this bar, so to get it and process the event an eventFilter plus some search is needed. I added all the logic inside the event filter class.. but not sure if it is the best approach.

I attach a [video](http://youtu.be/wfvJupX4T5U) so you get the idea.

### DIFF
--- a/spyderlib/plugins/__init__.py
+++ b/spyderlib/plugins/__init__.py
@@ -65,8 +65,6 @@ class TabFilter(QObject):
     This filter also holds the methods needed for the detection of a drag and
     the movement of tabs.
     """
-    sig_tab_moved = Signal(int, int)
-
     def __init__(self, dock_tabbar, main):
         QObject.__init__(self)
         self.dock_tabbar = dock_tabbar
@@ -84,8 +82,8 @@ class TabFilter(QObject):
 
     def _get_plugins(self):
         """
-        Get a list of all the plugins references in the QTabBar to which this
-        event filter was attached.
+        Get a list of all plugin references in the QTabBar to which this
+        event filter is attached.
         """
         plugins = []
         for index in range(self.dock_tabbar.count()):
@@ -95,7 +93,9 @@ class TabFilter(QObject):
 
     def _fix_cursor(self, from_index, to_index):
         """Fix mouse cursor position to adjust for different tab sizes."""
+        # The direction is +1 (moving to the right) or -1 (moving to the left)
         direction = abs(to_index - from_index)/(to_index - from_index)
+
         tab_width = self.dock_tabbar.tabRect(to_index).width()
         tab_x_min = self.dock_tabbar.tabRect(to_index).x()
         tab_x_max = tab_x_min + tab_width
@@ -114,7 +114,11 @@ class TabFilter(QObject):
             cursor.setPos(new_pos)
 
     def eventFilter(self,  obj,  event):
-        """Filter mouse press events."""
+        """Filter mouse press events.
+
+        Events that are captured and not propagated return True. Events that
+        are not captured and are propagated return False.
+        """
         event_type = event.type()
         if event_type == QEvent.MouseButtonPress:
             self.tab_pressed(event)
@@ -155,7 +159,6 @@ class TabFilter(QObject):
 
         from_index, to_index = self.from_index, self.to_index
         if from_index != to_index and from_index != -1 and to_index != -1:
-            self.sig_tab_moved.emit(from_index, to_index)
             self.move_tab(from_index, to_index)
             self._fix_cursor(from_index, to_index)
             self.from_index = to_index
@@ -182,9 +185,11 @@ class TabFilter(QObject):
 
     def show_tab_menu(self):
         """Show the context menu assigned to tabs."""
+        pass
 
     def show_nontab_menu(self):
         """Show the context menu assigned to nontabs section."""
+        pass
 
 
 class SpyderDockWidget(QDockWidget):
--- a/spyderlib/plugins/__init__.py
+++ b/spyderlib/plugins/__init__.py
@@ -23,7 +23,7 @@ from spyderlib.qt import PYQT5
 from spyderlib.qt.QtGui import (QDockWidget, QWidget, QShortcut, QCursor,
                                 QKeySequence, QMainWindow, QApplication,
                                 QTabBar)
-from spyderlib.qt.QtCore import Qt, Signal, QObject, QEvent
+from spyderlib.qt.QtCore import Qt, Signal, QObject, QEvent, QPoint
 
 
 # Stdlib imports
@@ -65,13 +65,15 @@ class TabFilter(QObject):
     This filter also holds the methods needed for the detection of a drag and
     the movement of tabs.
     """
+    sig_tab_moved = Signal(int, int)
+
     def __init__(self, dock_tabbar, main):
         QObject.__init__(self)
         self.dock_tabbar = dock_tabbar
         self.main = main
         self.moving = False
-        self.start_tab = None
-        self.end_tab = None
+        self.from_index = None
+        self.to_index = None
 
     # Helper methods
     def _get_plugin(self, index):
@@ -91,6 +93,26 @@ class TabFilter(QObject):
             plugins.append(plugin)
         return plugins
 
+    def _fix_cursor(self, from_index, to_index):
+        """Fix mouse cursor position to adjust for different tab sizes."""
+        direction = abs(to_index - from_index)/(to_index - from_index)
+        tab_width = self.dock_tabbar.tabRect(to_index).width()
+        tab_x_min = self.dock_tabbar.tabRect(to_index).x()
+        tab_x_max = tab_x_min + tab_width
+        previous_width = self.dock_tabbar.tabRect(to_index - direction).width()
+
+        delta = previous_width - tab_width
+        if delta > 0:
+            delta = delta * direction
+        else:
+            delta = 0
+        cursor = QCursor()
+        pos = self.dock_tabbar.mapFromGlobal(cursor.pos())
+        x, y = pos.x(), pos.y()
+        if x < tab_x_min or x > tab_x_max:
+            new_pos = self.dock_tabbar.mapToGlobal(QPoint(x + delta, y))
+            cursor.setPos(new_pos)
+
     def eventFilter(self,  obj,  event):
         """Filter mouse press events."""
         event_type = event.type()
@@ -107,45 +129,62 @@ class TabFilter(QObject):
 
     def tab_pressed(self, event):
         """Method called when a tab from a QTabBar has been pressed."""
-        if event.button() == Qt.LeftButton:
-            self.start_tab = self.dock_tabbar.tabAt(event.pos())
+        self.from_index = self.dock_tabbar.tabAt(event.pos())
+
+        if event.button() == Qt.RightButton:
+            if self.from_index == -1:
+                self.show_nontab_menu()
+            else:
+                self.show_tab_menu()
 
     def tab_moved(self, event):
         """Method called when a tab from a QTabBar has been moved."""
         # If the left button isn't pressed anymore then return
         if not event.buttons() & Qt.LeftButton:
+            self.to_index = None
             return
-        if not self.moving:
+
+        self.to_index = self.dock_tabbar.tabAt(event.pos())
+
+        if not self.moving and self.from_index != -1 and self.to_index != -1:
             QApplication.setOverrideCursor(Qt.ClosedHandCursor)
             self.moving = True
 
-        pos = event.pos()
-        self.end_tab = self.dock_tabbar.tabAt(pos)
-        if self.end_tab < 0:
-            self.end_tab = self.start_tab
+        if self.to_index == -1:
+            self.to_index = self.from_index
 
-        if self.start_tab != self.end_tab:
-            self.move_tab(self.start_tab, self.end_tab)
-            self.start_tab = self.end_tab
+        from_index, to_index = self.from_index, self.to_index
+        if from_index != to_index and from_index != -1 and to_index != -1:
+            self.sig_tab_moved.emit(from_index, to_index)
+            self.move_tab(from_index, to_index)
+            self._fix_cursor(from_index, to_index)
+            self.from_index = to_index
 
     def tab_released(self, event):
         """Method called when a tab from a QTabBar has been released."""
         QApplication.restoreOverrideCursor()
         self.moving = False
 
-    def move_tab(self, start_tab, end_tab):
-        """Move a tab from a start to and end position."""
+    def move_tab(self, from_index, to_index):
+        """Move a tab from a given index to a given index position."""
         plugins = self._get_plugins()
-        start_plugin = self._get_plugin(start_tab)
-        end_plugin = self._get_plugin(end_tab)
+        from_plugin = self._get_plugin(from_index)
+        to_plugin = self._get_plugin(to_index)
 
-        start = plugins.index(start_plugin)
-        end = plugins.index(end_plugin)
-        plugins[start], plugins[end] = plugins[end], plugins[start]
+        from_idx = plugins.index(from_plugin)
+        to_idx = plugins.index(to_plugin)
+
+        plugins[from_idx], plugins[to_idx] = plugins[to_idx], plugins[from_idx]
 
         for i in range(len(plugins)-1):
             self.main.tabify_plugins(plugins[i], plugins[i+1])
-        start_plugin.dockwidget.raise_()
+        from_plugin.dockwidget.raise_()
+
+    def show_tab_menu(self):
+        """Show the context menu assigned to tabs."""
+
+    def show_nontab_menu(self):
+        """Show the context menu assigned to nontabs section."""
 
 
 class SpyderDockWidget(QDockWidget):
--- a/spyderlib/plugins/__init__.py
+++ b/spyderlib/plugins/__init__.py
@@ -134,7 +134,7 @@ class TabFilter(QObject):
         self.moving = False
 
     def move_tab(self, start_tab, end_tab):
-        """Move a tab from a start to and end position ."""
+        """Move a tab from a start to and end position."""
         plugins = self._get_plugins()
         start_plugin = self._get_plugin(start_tab)
         end_plugin = self._get_plugin(end_tab)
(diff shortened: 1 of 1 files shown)


Nodd:
Nice ! Is it you playing the guitar on the video ? :smile: 
What would you put in the contextual menu ? Entries for closing/setfloating/maximizing ? I think it would be helpful only it the titlebars are hidden as per https://github.com/spyder-ide/spyder/pull/2345#issuecomment-96438578

goanpeca:
Hehe, no, the software I use puts the audio in, but i do play the guitar actually :-)

Now ... well I am considering two things.. if ALL widgets have a corner menu (most do on the upper right corner...) then ybe the options should go there... or also in the tab if right clicking...

I guess is natural to try to right click on a tab...

Nodd:
It's natural to right click but I'd expect to see options about tab management, not about the content of the page. I don't know, maybe people would use it anyway...

goanpeca:
Yes that is what I want, to put tab related stuff... move to right, move to right end..., move to left, move to left end..., close tab, float panel... etc

Nodd:
Honnestly I'm not sure the menu is worth it. In FIrefox I regularely have more than 50 tabs opened and it's a mess to manage, but in spyder I have at most 5 dockpanels at the same place, so actions like move to left or move to end are somewhat useless.

goanpeca:
I see what you mean, then this is almost ready :p

ccordoba12:
First let me say that this is terrific addition!! Right now the way to reorganize dock widgets is to undock and dock them again, so this improves things by an order of magnitude!!

I also agree with @Nodd, I think adding a contextual menu is too much. Let's leave things simple for now :-)

goanpeca:
Ok, I added some docstrings and is ready for a thorough revision ;-) @ccordoba12

goanpeca:
@ccordoba12 this one is good to go, and easier to check (than the editor ine... that is still WIP)

spyderlib/plugins/__init__.py
```diff
@@ -124,6 +126,14 @@ def __init__(self, *args, **kwargs):
         if sys.platform == 'darwin':
             self.setStyleSheet(self.DARWIN_STYLE)
 
+        # Needed for the installation of the event filter
+        self.title = args[0]
+        self.main = args[1]
```
ccordoba12: Where is `args` coming from?

spyderlib/plugins/__init__.py
```diff
@@ -124,6 +126,14 @@ def __init__(self, *args, **kwargs):
         if sys.platform == 'darwin':
             self.setStyleSheet(self.DARWIN_STYLE)
 
+        # Needed for the installation of the event filter
+        self.title = args[0]
+        self.main = args[1]
+        self.bar = None
```
ccordoba12: wouldn't `dock_bar` be better, instead of `bar` here?

spyderlib/plugins/__init__.py
```diff
@@ -131,6 +141,118 @@ def closeEvent(self, event):
...
+    def __init__(self, tabbar, main):
+        QObject.__init__(self)
+        self.tabbar = tabbar
+        self.main = main
+        self.moving = False
+        self.start_tab = None
+        self.end_tab = None
+
+    # Helper methods
+    def _get_widget(self, index):
```
ccordoba12: Why not call this `_get_plugin` instead?

spyderlib/plugins/__init__.py
goanpeca: when create_dockwidget is called the args passed are the title and the parent (the main window in this case)

I agree is not so clear, but I am not so sure how to make it better...

spyderlib/plugins/__init__.py
ccordoba12: I say it because I assume this is returning a `SpyderPlugin` instance, right?

spyderlib/plugins/__init__.py
goanpeca: bar here refers to tabbar... so maybe I should change it

spyderlib/plugins/__init__.py
goanpeca: Well.. we should change the variable in spyder.py that is called widgetlist (other PR) ;-) ... and yes I agree is better with plugins...

spyderlib/plugins/__init__.py
ccordoba12: But it's the `tabbar` that holds dockwidgets, right?

So another possibility would be `dock_tabbar` :-)

spyderlib/plugins/__init__.py
ccordoba12: But we have `get_plugin_title` and `self.main` already. I don't get it ;-)

spyderlib/plugins/__init__.py
goanpeca: ok sure :)

spyderlib/plugins/__init__.py
ccordoba12: Ok

spyderlib/plugins/__init__.py
```diff
@@ -131,6 +141,118 @@ def closeEvent(self, event):
...
+            return True
+        if event_type == QEvent.MouseButtonRelease:
+            self.tab_released(event)
+            return True
+        return False
+
+    def tab_pressed(self, event):
+        """Method called when a tab from a QTabBar has been pressed."""
+        if event.button() == Qt.LeftButton:
+            self.start_tab = self.tabbar.tabAt(event.pos())
```
ccordoba12: Do you want to change the mouse icon here too?

ccordoba12: Ups, sorry, it's below :-)

spyderlib/plugins/__init__.py
goanpeca: But those things belong to the plugin, not to the dockwidget,

spyderlib/plugins/__init__.py
goanpeca: Yep. only when it starts to move ;-)

goanpeca:
@ccordoba12 added corrections.

With regard to the question of the dockwidget... `SpyderPlugin` has a method create_dockwidget that uses `SpyderDockwidget` but the methods are from the plugin (SpyderPlugin) not from the widget, that is why I need to store those values when an instance if `SpyderDockwidget` is created...

spyderlib/plugins/__init__.py
```diff
@@ -55,6 +57,97 @@ def get_icon(self):
...
+
+    def tab_pressed(self, event):
+        """Method called when a tab from a QTabBar has been pressed."""
+        if event.button() == Qt.LeftButton:
+            self.start_tab = self.dock_tabbar.tabAt(event.pos())
+
+    def tab_moved(self, event):
+        """Method called when a tab from a QTabBar has been moved."""
+        # If the left button isn't pressed anymore then return
+        if not event.buttons() & Qt.LeftButton:
```
ccordoba12: Why not a regular `and`, instead of `&` here?

spyderlib/plugins/__init__.py
```diff
@@ -55,6 +57,97 @@ def get_icon(self):
...
+            self.move_tab(self.start_tab, self.end_tab)
+            self.start_tab = self.end_tab
+
+    def tab_released(self, event):
+        """Method called when a tab from a QTabBar has been released."""
+        QApplication.restoreOverrideCursor()
+        self.moving = False
+
+    def move_tab(self, start_tab, end_tab):
+        """Move a tab from a start to and end position ."""
```
ccordoba12: There's an unnecessary blank after `end position`

spyderlib/plugins/__init__.py
```diff
@@ -124,6 +217,14 @@ def __init__(self, *args, **kwargs):
         if sys.platform == 'darwin':
             self.setStyleSheet(self.DARWIN_STYLE)
 
+        # Needed for the installation of the event filter
+        self.title = args[0]
+        self.main = args[1]
+        self.dock_tabbar = None
+
+        # Update the QTabBar everytime dockwidget visibility changes
+        self.visibilityChanged.connect(self.install_tab_event_filter)
```
ccordoba12: Why is this necessary?

spyderlib/plugins/__init__.py
ccordoba12: I see. Then you need to redefine the __init__ method of `SpyderDockWidget` to accept the right args, i.e. something like

```python
def __init__(self, title, main, **kwargs):
    super(SpyderDockWidget, self).__init__(**kwargs)
    self.title = title
    self.main = main
    ...
```

spyderlib/plugins/__init__.py
goanpeca: hmm True, proabably a leftover form the C++ code I took as example

spyderlib/plugins/__init__.py
goanpeca: Got it

spyderlib/plugins/__init__.py
goanpeca: When you move a dockwidget to a new position (location that was not existing before) this event gets triggered so I can add the filter to the QTabBar that gets created.

Was the only way I could think of so far...

spyderlib/plugins/__init__.py
goanpeca: No I checked... buttons is an Or combination so I need the Binary And there...

http://doc.qt.io/qt-4.8/qmouseevent.html#buttons

spyderlib/plugins/__init__.py
goanpeca: Fixed

spyderlib/plugins/__init__.py
goanpeca: Also it even though this operation is triggered often... the filter is only installed once per QTabBar, so it should not impact usability. I hope there was a more intuitive and direct way offered by Qt, but I guess dockwidgets are meant to be used as extra panels in an application with a main  widget.. ;-)

goanpeca: Oops I should have used `dockLocationChanged`

Will adapt the coed

spyderlib/plugins/__init__.py
```diff
@@ -55,6 +57,97 @@ def get_icon(self):
...
+            self.move_tab(self.start_tab, self.end_tab)
+            self.start_tab = self.end_tab
+
+    def tab_released(self, event):
+        """Method called when a tab from a QTabBar has been released."""
+        QApplication.restoreOverrideCursor()
+        self.moving = False
+
+    def move_tab(self, start_tab, end_tab):
+        """Move a tab from a start to and end position."""
```
ccordoba12: `to an end`, not `to and end` :-)

spyderlib/plugins/__init__.py
```diff
@@ -55,6 +57,97 @@ def get_icon(self):
...
+        plugins = self._get_plugins()
+        start_plugin = self._get_plugin(start_tab)
+        end_plugin = self._get_plugin(end_tab)
+
+        start = plugins.index(start_plugin)
+        end = plugins.index(end_plugin)
```
ccordoba12: What about `start_idx` and `end_idx` for these variable names? I think they should be clearer ;-)

spyderlib/plugins/__init__.py
```diff
@@ -119,11 +212,21 @@ class SpyderDockWidget(QDockWidget):
 
     plugin_closed = Signal()
 
-    def __init__(self, *args, **kwargs):
-        super(SpyderDockWidget, self).__init__(*args, **kwargs)
+    def __init__(self, title, parent):
+        super(SpyderDockWidget, self).__init__(title, parent)
```
ccordoba12: Have you tried this in PyQt5? Is it working correctly there?

spyderlib/plugins/__init__.py
goanpeca: Sure I can change...

spyderlib/plugins/__init__.py
goanpeca: Not really, but I do not see why it would not work with PyQt5

goanpeca: I will check later

ccordoba12: Sorry, problems with PyQt5 only happen on mixins. So, there shouldn't be any problem with this one :-)

goanpeca:
I renamed variables to be more consistent with the other PR (editor draggable tabs) and added support to call context menus if needed. For the moment the methods are empty, but I added them in case they are useful in the future.

I also included the cursor fix so that it compensates when the moved tabs have different size.

I made suggested corrections and everything now is working as expected :-), so if @ccordoba12 can check and approve, this one is ready.

spyderlib/plugins/__init__.py
```diff
@@ -55,6 +57,136 @@ def get_icon(self):
...
+    # Helper methods
+    def _get_plugin(self, index):
+        """Get plugin reference based on tab index."""
+        for plugin in self.main.widgetlist:
+            if plugin.get_plugin_title() == self.dock_tabbar.tabText(index):
+                return plugin
+
+    def _get_plugins(self):
+        """
+        Get a list of all the plugins references in the QTabBar to which this
```
ccordoba12: Please change this to `Get a list of all plugin references in the QTabbar ...`

goanpeca: Sure :+1:

spyderlib/plugins/__init__.py
```diff
@@ -55,6 +57,136 @@ def get_icon(self):
...
+        for i in range(len(plugins)-1):
+            self.main.tabify_plugins(plugins[i], plugins[i+1])
+        from_plugin.dockwidget.raise_()
+
+    def show_tab_menu(self):
+        """Show the context menu assigned to tabs."""
+
+    def show_nontab_menu(self):
+        """Show the context menu assigned to nontabs section."""
+
```
ccordoba12: Aren't these two last methods missing a `pass`

goanpeca: I could add them, but the docstring replaces the need for a `pass`... I can add... or leave it like it is

ccordoba12: Please add `pass` to them. It feels weird without it :-)

spyderlib/plugins/__init__.py
```diff
@@ -55,6 +57,136 @@ def get_icon(self):
...
+        if x < tab_x_min or x > tab_x_max:
+            new_pos = self.dock_tabbar.mapToGlobal(QPoint(x + delta, y))
+            cursor.setPos(new_pos)
+
+    def eventFilter(self,  obj,  event):
+        """Filter mouse press events."""
+        event_type = event.type()
+        if event_type == QEvent.MouseButtonPress:
+            self.tab_pressed(event)
+            return False
```
ccordoba12: Why these return values are needed? Could you add a comment about their need and why sometimes are True and others False?

spyderlib/plugins/__init__.py
```diff
@@ -55,6 +57,136 @@ def get_icon(self):
...
+        if not self.moving and self.from_index != -1 and self.to_index != -1:
+            QApplication.setOverrideCursor(Qt.ClosedHandCursor)
+            self.moving = True
+
+        if self.to_index == -1:
+            self.to_index = self.from_index
+
+        from_index, to_index = self.from_index, self.to_index
+        if from_index != to_index and from_index != -1 and to_index != -1:
+            self.sig_tab_moved.emit(from_index, to_index)
```
ccordoba12: Why is this signal needed if nothing connects to it?

goanpeca: At this moment it does nothing, but I guess it could be useful if some behavior wants to be executed whenever a the tabs are reorganized. It is for completeness...

spyderlib/plugins/__init__.py
```diff
@@ -119,11 +251,21 @@ class SpyderDockWidget(QDockWidget):
...
 
+        # Needed for the installation of the event filter
+        self.title = title
+        self.main = parent
+        self.dock_tabbar = None
+
+        # To track dockwidget changes the filter is installed when dockwidget
+        # visibility changes. This installs the filter on startup and also
+        # on dockwidgets that are undocked and then docked to a new location.
+        self.visibilityChanged.connect(self.install_tab_event_filter)
```
ccordoba12: Hadn't you mentioned that you were going to use a different signal here?

spyderlib/plugins/__init__.py
goanpeca: An event filter intercepts the fired events, and it has to return True, if it propagates the events... or False if it captures the event.... 

It is the normal behavior ....

http://doc.qt.io/qt-4.8/eventsandfilters.html#event-filters

spyderlib/plugins/__init__.py
goanpeca: Done

spyderlib/plugins/__init__.py
ccordoba12: Then please remove it :-)

When (if) it's needed, it can easily be added. Right now it's just noise and adds confusion to the code.

spyderlib/plugins/__init__.py
goanpeca: I will add some info to the docsting though

spyderlib/plugins/__init__.py
goanpeca: I did, but then I remembered that it did not work properly, and I revert to this one. The problem of using the other signal is that it does not work when a dockwidget is moved to a new position.

spyderlib/plugins/__init__.py
goanpeca: Done

spyderlib/plugins/__init__.py
ccordoba12: Ok, no problem.

spyderlib/plugins/__init__.py
ccordoba12: Ok, thanks.

spyderlib/plugins/__init__.py
ccordoba12: Ok, thanks!

ccordoba12:
Nothing more from me to say :-) Please make my suggested changes so I can start testing.

goanpeca:
@ccordoba12 done

spyderlib/plugins/__init__.py
```diff
@@ -55,6 +57,141 @@ def get_icon(self):
...
+            self.main.tabify_plugins(plugins[i], plugins[i+1])
+        from_plugin.dockwidget.raise_()
+
+    def show_tab_menu(self):
+        """Show the context menu assigned to tabs."""
+        pass
+
+    def show_nontab_menu(self):
+        """Show the context menu assigned to nontabs section."""
+        pass
```
ccordoba12: Las thing, I think I have the perfect suggestion for this entry

```python
    self.main.createPopupMenu()
```

goanpeca: Ehh ? what does it do?

goanpeca: http://doc.qt.io/qt-4.8/qmainwindow.html#createPopupMenu ?

ccordoba12: It shows the same context menu as when you do a right click over a toolbar.

Essentially, It shows the Help menu but with just a few entries.

goanpeca: it does not work...

goanpeca: Ok fixed it :), now this will only popup when clicking in an area where no tabs are present...

In this image I right clicked to the right of the `File explorer` Tab

image

goanpeca:
@ccordoba12 you want to get the same context menu when right clicking on a tab as well??

I have two methods in this PR, one is `show_tab_menu` (right click on a tab) and `show_nontab_menu` (right click on a QTabBar empty space, generally to the right of the tabs)

ccordoba12:
I'd say yes, for now. But leave the two methods, so if we come up with a better context menu for tabs on the future, we can change it easily :-)

goanpeca:
Ready for testing @ccordoba12

ccordoba12:
Working great, even in PyQt5!

Nothing else than merging ;-)

PR #2371, opened 2015-04-27
Automatically add parens for function completions

### LINKED ISSUE #680: Code completion: insert automatically parentheses for functions/methods
_From [jorisvan...@gmail.com](https://code.google.com/u/101707288318167035253/) on 2011-06-03T08:49:09Z_

1. For example if you type in the editor 'plt.', a list of possibilities will show.
2. When I choose one, for expample 'show', in the editor 'plt.show' shows up.

What is the expected output? What do you see instead? Would it be possible that not 'plt.show' appears, but at once 'plt.show()' (and with the cursor between the brackets), so you don't have to type the brackets yourself?
Because after show, brackets are always needed. What version of the product are you using? On what operating system? Spyder 2.0.11, Ubuntu 11.04

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=680_

blink1073:
Fixes #680.  Automatically appends parens to completions that are known to be a `function`, `method`, or `class`, respecting the same rules as the standard auto parens in the code editor.

### DIFF
--- a/spyderlib/widgets/sourcecode/base.py
+++ b/spyderlib/widgets/sourcecode/base.py
@@ -79,6 +79,7 @@ class CompletionWidget(QListWidget):
                      'class': 'class',
                      'module': 'module'}
 
+        self.type_list = types
         if any(types):
             for (c, t) in zip(completion_list, types):
                 icon = icons_map.get(t, 'no_match')
@@ -155,6 +156,11 @@ class CompletionWidget(QListWidget):
         if (key in (Qt.Key_Return, Qt.Key_Enter) and self.enter_select) \
            or key == Qt.Key_Tab:
             self.item_selected()
+            if self.type_list[self.currentRow()] not in ['class', 'function',
+                                                         'method']:
+                return
+            if self.textedit.close_parentheses_enabled:
+                self.textedit.handle_close_parentheses('')
         elif key in (Qt.Key_Return, Qt.Key_Enter,
                      Qt.Key_Left, Qt.Key_Right) or text in ('.', ':'):
             self.hide()
--- a/spyderlib/widgets/sourcecode/base.py
+++ b/spyderlib/widgets/sourcecode/base.py
@@ -67,6 +67,9 @@ class CompletionWidget(QListWidget):
         completion_list = [c[0] for c in completion_list]
         if len(completion_list) == 1 and not automatic:
             self.textedit.insert_completion(completion_list[0])
+            if self.textedit.close_parentheses_enabled:
+                if types[0] in ['class', 'function', 'method']:
+                    self.textedit.handle_close_parentheses('')
             return
 
         self.completion_list = completion_list
--- a/spyderlib/widgets/sourcecode/codeeditor.py
+++ b/spyderlib/widgets/sourcecode/codeeditor.py
@@ -2572,20 +2572,8 @@ class CodeEditor(TextEditBaseWidget):
             self.stdkey_end(shift, ctrl)
         elif text == '(' and not self.has_selected_text():
             self.hide_completion_widget()
-            position = self.get_position('cursor')
-            s_trailing_text = self.get_text('cursor', 'eol').strip()
-            if self.close_parentheses_enabled and \
-              (len(s_trailing_text) == 0 or \
-              s_trailing_text[0] in (',', ')', ']', '}')):
-                self.insert_text('()')
-                cursor = self.textCursor()
-                cursor.movePosition(QTextCursor.PreviousCharacter)
-                self.setTextCursor(cursor)
-            else:
-                self.insert_text(text)
-            if self.is_python_like() and self.get_text('sol', 'cursor') and \
-              self.calltips:
-                self.sig_show_object_info.emit(position)
+            if self.close_parentheses_enabled:
+                self.handle_close_parentheses(text)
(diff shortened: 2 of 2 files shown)


spyderlib/widgets/sourcecode/base.py
```diff
@@ -155,6 +157,18 @@ def keyPressEvent(self, event):
         if (key in (Qt.Key_Return, Qt.Key_Enter) and self.enter_select) \
            or key == Qt.Key_Tab:
             self.item_selected()
+            if self.type_list[self.currentRow()] not in ['class', 'function',
+                                                          'method']:
+                return
+            edit = self.textedit
+            s_trailing_text = edit.get_text('cursor', 'eol').strip()
```
ccordoba12: Why not `rstrip` instead of `strip` here?

spyderlib/widgets/sourcecode/base.py
```diff
@@ -155,6 +157,18 @@ def keyPressEvent(self, event):
...
            or key == Qt.Key_Tab:
             self.item_selected()
+            if self.type_list[self.currentRow()] not in ['class', 'function',
+                                                          'method']:
+                return
+            edit = self.textedit
+            s_trailing_text = edit.get_text('cursor', 'eol').strip()
+            if edit.close_parentheses_enabled and \
+              (len(s_trailing_text) == 0 or \
+              s_trailing_text[0] in (',', ')', ']', '}')):
```
ccordoba12: One space to left here please, so `s_trailing_text` is placed exactly below `len`.

Also, why this second condition? Could you give some examples of when it applies? I mean, why are there here `]` and `}`?

blink1073: I copied this code straight out of the code editor.  Should we factor this into a method and fix it up?
https://github.com/spyder-ide/spyder/blob/master/spyderlib/widgets/sourcecode/codeeditor.py#L2404

ccordoba12: Yes, it'd be the best I think

blink1073:
Added the new method to handle close parens.

blink1073:
I'm not sure why the check for the other characters is there, but I'm sure someone had a valid reason.  :smile:

blink1073:
As a bonus, we get the calltip now!

blink1073:
Rebased after icon merge.

spyderlib/widgets/sourcecode/codeeditor.py
```diff
@@ -2654,6 +2642,22 @@ def keyPressEvent(self, event):
             if self.is_completion_widget_visible() and text:
                 self.completion_text += text
 
+    def handle_close_parentheses(self, text):
+        if not self.close_parentheses_enabled:
+            return
+        position = self.get_position('cursor')
+        rest = self.get_text('cursor', 'eol').rstrip()
+        if (len(rest) == 0 or rest[0] in (',', ')', ']', '}')):
```
Nodd: `not rest` is more pythonic than `len(rest) == 0`.
Also the surrounding parenthesis are not necessary.

spyderlib/widgets/sourcecode/codeeditor.py
```diff
@@ -2654,6 +2642,22 @@ def keyPressEvent(self, event):
...
+        rest = self.get_text('cursor', 'eol').rstrip()
+        if (len(rest) == 0 or rest[0] in (',', ')', ']', '}')):
+            self.insert_text('()')
+            cursor = self.textCursor()
+            cursor.movePosition(QTextCursor.PreviousCharacter)
+            self.setTextCursor(cursor)
+        else:
+            self.insert_text(text)
+        if self.is_python_like() and self.get_text('sol', 'cursor') and \
+          self.calltips:
```
Nodd: The indentation looks weird here. I'd put 8 more spaces than the `if`.

Nodd:
I added a few style comments.

These trailing whitespaces are really annoying...

blink1073:
Fixed!

ccordoba12:
This is a really nice addition! And it's working as expected :-)

I just have one question: could we make this work when users press `Tab` and there is only one completion possible? I mean, for example, if you write:

    numpy.vec<Tab>

you won't get parenthesis nicely inserted (which, by the way, also shows you the signature tooltip).

ccordoba12:
If that's not possible, we should seriously consider to tackle issue #1914 :-)

blink1073:
Works now!

ccordoba12:
I'm seeing this error in the internal console (with PyQt5 only) when pressing `Tab` in quick succession, and if there are no possible completions to get:

```python-traceback
  File "/home/carlos/Projects/spyder/github-repo/spyderlib/utils/introspection/plugin_manager.py", line 56, in _handle_timeout
    debug_print('got timeout: %s' % self.plugins)
AttributeError: 'NoneType' object has no attribute 'plugins'
```

An example to generate this

    numpy.vectorize<Tab><Tab><Tab>

After this message is shown, each time parens are introduced the same error appears again.

The error seems unrelated to this PR. So it's up to you @blink1073, if you want to fix it here or leave it for later :-)

blink1073:
I'd say leave that one for a different issue.

ccordoba12:
Ok, merging then ;-)

PR #2372, opened 2015-04-27
PR: [WIP] Add support for editor tab reordering

### LINKED ISSUE #564: Editor: drag n' drop (movable) tabs
_From [isamertg...@gmail.com](https://code.google.com/u/112414866069293531019/) on 2011-02-27T06:59:54Z_

We cant move tabs or we cant change their index. If we can, it would be nice.

Best regards.

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=564_

goanpeca:
## Description
Any reordering event will reorder ALL editorstacks opened, so everything remains in sync. Since the editor was sorting by default, I disabled all instances were this was happenning, now we need to discuss what other options should we offer, including sorting and also extra ones (Delete tabs to the right, to the left, delete all tabs except this etc...)

My take on this is that new files should be opened in the order of opening (what happens now with this PR) and if the user wants to order them by some measure, it should be an active action from the user in the context menu.

The [video](http://youtu.be/tbjJ-uM61P0)

## Todo
- [x] Make tabs moving dynamic (move tabs while dragging)
- [x] Erase sorting in preferences
- [x] Disable tab refresh when deleting a tab
- [ ] Add support for tab scrolling when dragging with many tabs open (going to the right or left end should move the tabs)
- [ ] Remove all code reference to fullpath sorting in the calling and creation of Editors?
- [ ] Check nothing is broken!

**For another PR**
- Offer sorting options
- Offer additional options (Delete tabs to the right, to the left, delete all tabs except)?
- Adapt editor preferences panel sorting options

### DIFF
--- a/spyderlib/plugins/editor.py
+++ b/spyderlib/plugins/editor.py
@@ -348,6 +348,7 @@ class Editor(SpyderPluginWidget):
     open_dir = Signal(str)
     breakpoints_saved = Signal()
     run_in_current_extconsole = Signal(str, str, str, bool, bool)
+    sig_tabs_reordered = Signal(int, int)
     
     def __init__(self, parent, ignore_last_opened_files=False):
         if PYQT5:
@@ -439,7 +440,10 @@ class Editor(SpyderPluginWidget):
         self.splitter.setStretchFactor(1, 1)
         layout.addWidget(self.splitter)
         self.setLayout(layout)
-        
+
+        # Tabs reordering
+        self.sig_tabs_reordered.connect(self.tabs_reordered)
+
         # Editor's splitter state
         state = self.get_option('splitter_state', None)
         if state is not None:
@@ -1046,6 +1050,9 @@ class Editor(SpyderPluginWidget):
     def register_editorstack(self, editorstack):
         self.editorstacks.append(editorstack)
         
+        # Needed for tabs reordering calls
+        editorstack.sig_tabs_reordered.connect(self.sig_tabs_reordered)
+
         self.register_widget_shortcuts("Editor", editorstack)
         
         if self.isAncestorOf(editorstack):
@@ -1192,6 +1199,26 @@ class Editor(SpyderPluginWidget):
             if str(id(editorstack)) != editorstack_id_str:
                 editorstack.rename_in_data(index, filename)
 
+    def tabs_reordered(self, start, end):
+        """ """
+        if start < 0 or end < 0:
+            return
+        else:
+            steps = abs(end - start)
+            sign = (end - start) / steps
+            
+            for editorstack in self.editorstacks :
+                data = editorstack.data
+                editorstack.blockSignals(True)
+    
+                for i in range(start, end, sign*1):
+                    editorstack.tabbar.moveTab(i, i+sign)
+                    data[i], data[i+sign] = data[i+sign], data[i]
+    
+                editorstack.blockSignals(False)
+                editorstack.refresh()
+                editorstack.tabbar.setCurrentIndex(start)
+                editorstack.tabbar.setCurrentIndex(end)
 
     #------ Handling editor windows    
     def setup_other_windows(self):
--- a/spyderlib/plugins/editor.py
+++ b/spyderlib/plugins/editor.py
@@ -105,12 +105,12 @@ class EditorConfigPage(PluginConfigPage):
                                     text=_("Text and margin font style"),
                                     fontfilters=QFontComboBox.MonospacedFonts)
         newcb = self.create_checkbox
-        fpsorting_box = newcb(_("Sort files according to full path"),
-                              'fullpath_sorting')
+#        fpsorting_box = newcb(_("Sort files according to full path"),
+#                              'fullpath_sorting')
         showtabbar_box = newcb(_("Show tab bar"), 'show_tab_bar')
 
         interface_layout = QVBoxLayout()
-        interface_layout.addWidget(fpsorting_box)
+#        interface_layout.addWidget(fpsorting_box)
         interface_layout.addWidget(showtabbar_box)
         interface_group.setLayout(interface_layout)
         
--- a/spyderlib/plugins/editor.py
+++ b/spyderlib/plugins/editor.py
@@ -348,7 +348,7 @@ class Editor(SpyderPluginWidget):
     open_dir = Signal(str)
     breakpoints_saved = Signal()
     run_in_current_extconsole = Signal(str, str, str, bool, bool)
-    sig_tabs_reordered = Signal(int, int)
+    sig_tab_moved = Signal(int, int)
     
     def __init__(self, parent, ignore_last_opened_files=False):
         if PYQT5:
@@ -442,7 +442,7 @@ class Editor(SpyderPluginWidget):
         self.setLayout(layout)
 
         # Tabs reordering
-        self.sig_tabs_reordered.connect(self.tabs_reordered)
+        self.sig_tab_moved.connect(self.move_tabs)
 
         # Editor's splitter state
         state = self.get_option('splitter_state', None)
@@ -1051,7 +1051,7 @@ class Editor(SpyderPluginWidget):
         self.editorstacks.append(editorstack)
         
         # Needed for tabs reordering calls
-        editorstack.sig_tabs_reordered.connect(self.sig_tabs_reordered)
+        editorstack.sig_tab_moved.connect(self.sig_tab_moved)
 
         self.register_widget_shortcuts("Editor", editorstack)
         
@@ -1199,24 +1199,32 @@ class Editor(SpyderPluginWidget):
             if str(id(editorstack)) != editorstack_id_str:
                 editorstack.rename_in_data(index, filename)
 
-    def tabs_reordered(self, start, end):
+    def move_tabs(self, start, end):
         """
         When tabs are reordered in a stackeditor this method adjusts tabs
-        in all the stackkeditors
+        in all the stackkeditors.
+
+        If needed can also be used programmatically (directly).
         """
+        # Lower bound check
         if start < 0 or end < 0:
             return
         else:
             steps = abs(end - start)
-            sign = (end - start) / steps
+            step = (end - start) / steps
             
-            for editorstack in self.editorstacks :
+            for editorstack in self.editorstacks:
                 data = editorstack.data
                 editorstack.blockSignals(True)
-    
-                for i in range(start, end, sign):
-                    editorstack.tabbar.moveTab(i, i+sign)
-                    data[i], data[i+sign] = data[i+sign], data[i]
+
+                # If used directly an upper bound check is needed
+                count = editorstack.tabbar.count()
+                if start > count or end > count:
+                    break
+
+                for i in range(start, end, step):
+                    editorstack.tabbar.moveTab(i, i+step)
+                    data[i], data[i+step] = data[i+step], data[i]
     
                 editorstack.blockSignals(False)
                 editorstack.refresh()
--- a/spyderlib/plugins/editor.py
+++ b/spyderlib/plugins/editor.py
@@ -105,12 +105,9 @@ class EditorConfigPage(PluginConfigPage):
                                     text=_("Text and margin font style"),
                                     fontfilters=QFontComboBox.MonospacedFonts)
         newcb = self.create_checkbox
-#        fpsorting_box = newcb(_("Sort files according to full path"),
-#                              'fullpath_sorting')
         showtabbar_box = newcb(_("Show tab bar"), 'show_tab_bar')
 
         interface_layout = QVBoxLayout()
-#        interface_layout.addWidget(fpsorting_box)
         interface_layout.addWidget(showtabbar_box)
         interface_group.setLayout(interface_layout)
         
--- a/spyderlib/plugins/editor.py
+++ b/spyderlib/plugins/editor.py
@@ -1200,7 +1200,10 @@ class Editor(SpyderPluginWidget):
                 editorstack.rename_in_data(index, filename)
 
     def tabs_reordered(self, start, end):
-        """ """
+        """
+        When tabs are reordered in a stackeditor this method adjusts tabs
+        in all the stackkeditors
+        """
         if start < 0 or end < 0:
             return
         else:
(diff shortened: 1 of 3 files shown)


Nodd:
Yes, more tabs ! :smile: 

:+1: for dynamic tabs (realtime moving while dragging), it's more obvious to see what happens.

ccordoba12:
This is fantastic!!! Thanks a lot for working on it! :-)

My opinions about your todos:

1. I agree too that *new files should be opened in the order of opening*. I think right now we're auto-sorting them by their path, but that's not intuitive.
2. This PR should also solve the issue of tab scrolling, not only when dragging but also when closing a file. I mean, there is a very annoying effect right now when closing a file and the tab scrolling buttons are on: the file you're trying to close is (automatically) made the last one from left to the right, regardless of its position. We should change that and not move tabs at all when closing files.
3. Let's leave new features (like new sorting options and additional options) for additional PRs. That will make my life as a reviewer much easier because I won't need to check for correctness of several things in a single PR.

goanpeca:
Should I remove the option in  preferences in this PR `Sort files according to fullpath`?

image

ccordoba12:
Yes, please. Let's leave new things for further PRs ;-)

Nodd:
It's not a new thing, it's an old thing :)

ccordoba12:
I know, but right now I'm not so sure if the best thing is to preserve the previous behavior, and doing so by adding (yet another) option to our crowded Preferences.

But instead of discussing that here, I'd prefer to do it in another PR, so I can review and merge this one quickly :-)

goanpeca:
Well.. the thing is, in this PR I commented the lines were sorting took place, so right now the  option in preferences does nothing, and I would rather comment it (than erase it) so we can focus on things and move on. It would be weird to merge this PR but leave the option there.

goanpeca:
You agree on the grabbing hand icon (ClosedHand) when dragging?

Or any other suggestion from [here](http://doc.qt.io/qt-4.8/qt.html#CursorShape-enum)

ccordoba12:
Yes, I'd like to see the CloseHand icon when dragging.

goanpeca:
@ccordoba12 I think I solved the issue you commented, but there is still another one that is quite annoying. If you have many tabs open and the scrolling buttons are there, if you click on  file closer to the left, the whole tabs will shift and put the current clicked tab to the right... very weird behavior...

Nodd:
Sorry for my previous comment, I was just nitpicking. I agree that the option should be removed for now (and it should be added to the context menu for tabs, this is the case where it makes sense).

goanpeca:
The code in the editor gives me headaches...

goanpeca:
hmmm I think it is a Qt limitation that when you setCurrentIndex for a specific tab, if many tabs (and a scroll buttons present) it will always slide the selected tab to the right of the QTabBar...

This leads to weird behaviors when moving tabs... so the best to have here would be shortcuts to move tabs (Ctrl+Shift+PgUp/PgDn)

spyderlib/plugins/editor.py
```diff
@@ -105,12 +105,12 @@ def setup_page(self):
                                     text=_("Text and margin font style"),
                                     fontfilters=QFontComboBox.MonospacedFonts)
         newcb = self.create_checkbox
-        fpsorting_box = newcb(_("Sort files according to full path"),
-                              'fullpath_sorting')
+#        fpsorting_box = newcb(_("Sort files according to full path"),
+#                              'fullpath_sorting')
```
ccordoba12: So this was a current option? Then we should remove in this PR because our decision is to order files according to their opening order.

spyderlib/widgets/editor.py
```diff
@@ -1917,6 +1925,41 @@ def dropEvent(self, event):
                 editor.insert_text( source.text() )
         event.acceptProposedAction()
 
+    #------ Tab reordering
+    # These methods are called from an event filter installed in 
+    # spyderlib\widets\tabs.py -> BaseTabs
+    def tab_pressed(self, event):
+        """ """
```
ccordoba12: Missing docstring

spyderlib/widgets/editor.py
```diff
@@ -1917,6 +1925,41 @@ def dropEvent(self, event):
...
+            # a method that reorders ALL openned editorstacks
+            self.sig_tabs_reordered.emit(self.moving_start_tab,
+                                         self.moving_end_tab)
+            self.moving_start_tab = self.moving_end_tab
+
+    def tab_released(self, event):
+        """ """
+        QApplication.restoreOverrideCursor()
+        self.moving_tab = False
+
```
ccordoba12: Do these methods really belong here? I mean, they seem to belong to `tabs.py` because they only deal with tabs, right?

spyderlib/widgets/tabs.py
```diff
@@ -13,7 +13,7 @@
 
 from spyderlib.qt.QtGui import (QTabWidget, QMenu, QDrag, QApplication,
                                 QTabBar, QWidget, QHBoxLayout)
-from spyderlib.qt.QtCore import Signal, Qt, QPoint, QMimeData, QByteArray
+from spyderlib.qt.QtCore import Signal, Qt, QPoint, QMimeData, QByteArray, QObject, QEvent
```
ccordoba12: 79 columns, please ;-)

spyderlib/widgets/tabs.py
```diff
@@ -112,7 +135,8 @@ def __init__(self, parent, actions=None, menu=None,
         QTabWidget.__init__(self, parent)
         
         self.setUsesScrollButtons(True)
-        
+        self.filter = TabFilter(parent)  # Need to keep reference
```
ccordoba12: I don't like this too much because the filter is installed for all tabs, but only works for tabs on editor_stack's(?)

Does this have a benefit for our other tabbed widgets or not?

spyderlib/widgets/editor.py
```diff
@@ -1917,6 +1925,41 @@ def dropEvent(self, event):
...
+    def tab_pressed(self, event):
+        """ """
+        if event.button() == Qt.LeftButton:
+            self.moving_start_tab = self.tabbar.tabAt(event.pos())
+            self.moving_end_tab = None
+            self.tabbar.setCurrentIndex(self.moving_start_tab)
+            self.tabs.currentWidget().setFocus()
+
+    def tab_moved(self, event):
+        """ """
```
ccordoba12: Docstring missing

spyderlib/widgets/editor.py
```diff
@@ -1065,7 +1074,7 @@ def rename_in_data(self, index, new_filename):
         set_new_index = index == self.get_stack_index()
         current_fname = self.get_current_filename()
         finfo.filename = new_filename
-        self.data.sort(key=self.__get_sorting_func())
+#        self.data.sort(key=self.__get_sorting_func())
```
ccordoba12: Remove

spyderlib/widgets/editor.py
```diff
@@ -1027,7 +1036,7 @@ def __get_sorting_func(self):
 
     def add_to_data(self, finfo, set_current):
         self.data.append(finfo)
-        self.data.sort(key=self.__get_sorting_func())
+#        self.data.sort(key=self.__get_sorting_func())
```
ccordoba12: Remove

spyderlib/widgets/editor.py
```diff
@@ -945,7 +954,7 @@ def set_fullpath_sorting_enabled(self, state):
         self.fullpath_sorting_enabled = state
         if self.data:
             finfo = self.data[self.get_stack_index()]
-            self.data.sort(key=self.__get_sorting_func())
+#            self.data.sort(key=self.__get_sorting_func())
```
ccordoba12: Remove

spyderlib/widgets/editor.py
goanpeca: I had them originally in tabs.py inside the TabFilter class, but to make the editors work when splitting and having multiple editors (New Editor Windows..) it was too cumbersome to get a hold of the parent editor stack, so I promoted them here, which allows for an easier relation and logic.

But this method is mostly a "hack" to overcome limitations of Qt and the current Editor design

spyderlib/widgets/tabs.py
goanpeca: Not really, I will move the filter to the widget\editor.py file

goanpeca:
@ccordoba12 take a look at this one first that is mostly ready

spyderlib/widgets/tabs.py
ccordoba12: Ok, much better :-)

goanpeca: The thing is that I had to do this incrementally... and as I escalated I had to rearrange and move stuff... it was very tricky!!

goanpeca:
question @ccordoba12 

Thinking a bit about this, I figured that maybe is better to have one and only one way to handle this events in the editor and the other widgets (like the consoles) so I could refactor the code so that everything uses this new method (instead of the current one used by consoles, that is based on [drag and drop](http://www.qtcentre.org/wiki/index.php?title=Movable_Tabs)...)

This would have the benefits of a single codebase, the behavior would be the same for all tab reordering actions (Editors... consoles.... Tabified dockwidgets) and I could add some extra helper functions to this new TabBar, like access methods to scrolling of tabs (programaticaly) which  I could eventually use to fix the tabs unexpected movements when selecting a new one (or deleting one).

goanpeca:
I will test if it works first....

ccordoba12:
I'd really prefer if the new TabBar is used by all our plugins, instead of only the Editor.

I hope you can make it work like that :-)

goanpeca:
I need some input...

@spyder-ide/core-developers what should be the behavior when moving (reordering a tab) when it reaches the right or left end (when it has many tabs in it).

image

ccordoba12:
What do you mean exactly?

goanpeca:
hmm I will make a video.. otherwise it is hard to explain...

goanpeca:
I uploaded this [video](http://youtu.be/ttjmp2S5PYQ), the first part is displaying the behavior when all tabs are visible (no scroll buttons), this tabbar is smart and adjusts the mouse pointer positions according to the next or previous tab widths, which allows for a fluid experience with no strange tab jumps or hiccups.

The second part is with a shrunken tabbar (with scroll buttons), where I tried to also make it smart by autoscrolling if it detects an incomplete next tab, and repositions the cursor in the middle of the tab, this way the user has complete control when moving tabs. My previous question was refering to this second part... I like how it is now, but I would like feedback.

@ccordoba12

ccordoba12:
I think it's fine, I like what I see in the video. It seems very cool!

Nodd:
Your video examples are really cool!

I don't like the second part very much, not having the tab on the right or left end when scrolling is disturbing. The way Firefox does it is more intuitive for me: basically, you have to put the mouse pointer outside the tabbbar, then it scrolls at a given speed (faster if the mouse is further on the side).

goanpeca:
@Nodd I do not understand this "not having the tab on the right or left end when scrolling is disturbing". What you suggest is harder to implement right due to the way the tabbar is designed

goanpeca:
@Nodd I just checked with Firefox... I do not like it so much cause it is too fast, the user has no real control...

Nodd:
es the speed is a bit too fast (I think it's needed with lots of tabs: you scroll a bit too much, the you go back 2 or 3 tabs without scrolling), but I find the principle better.
To be clear: your version is already very good. Actually I don't understand your question, do you already think about another way of doing it ?

goanpeca:
I tried the firefox approach first, but I could not come up with reliable code .. then I asked the question, then I figured doing it how I did it in the video.

But now I have more tools so I can  revisit the firefox version....

goanpeca:
@Nodd I got it to work more like firefox (not as fast!), it is not 100% perfect (somehow I need to look why...), but it is looking good :-)

Nodd:
Nice !

goanpeca:
question @ccordoba12 ... 

So I have a new set of classes that will replace the existing ones so that we have an uniform support for tab handling. 

But the change will include modifications to the consoles and other parts of spyder (where Tabs are used), so... should I add the new classes with the existing ones and make it work only with the editor in the meantime, and in another PR I adapt the consoles to use the new classes, (and hence erase the old classes)

I guess it would make this PR more manageable, and avoid breaking something

ccordoba12:
Yes, I agree with your plan. First use the new classes only to fix the problems we have in the Editor. Then (in another PR) use them to improve things in our consoles.

May I ask what's exactly the difference between your new classes and the old ones?

goanpeca:
The actual scheme has `TabBar`, `BaseTab`, `Tab`. 

`BaseTab` has no drag and drop support and is used by the editors and `Tab` (Which uses `TabBar`) has drag and drop support and is used by the consoles.

I will create a new `TabBar` and then a new (that uses a lot of the existing code) `Tab` (That will use the new `TabBar`), and then `Tab` will be used by both the editor and consoles and will allow for tab reordering (similar to Drag and drop but done in a simpler way) and also will allow for having specific contextual menus in tab or nontabs, or even the scroll buttons (if needed)

ccordoba12:
Thanks for the explanation! Sounds really good, please proceed :-)

ben1983:
This is great! Thank you for working on this.

Is it also possible to move the tab bar to the left? Similar to Matlab, TextWrangler, Firefox, etc.?
I found that a tab bar on the left is absolutely needed if more than ~5 files are open. 

See screenshot.

screen shot 2015-07-20 at 14 31 56

goanpeca:
@ben1983 thanks for the input. Have you tried using the outline explorer for what you suggest?, the outline explorer displays all the opened (if you enable the functionality) files.

The feature requested will not be implemented on this pull request, but I will take note of it.

Cheers

ben1983:
@goanpeca  Thank you for the quick response. Yes, I am using the outline explorer. The problems with the outline explorer are:
1) similar to the tab bar, one cannot change the order of the files in the outline explorer. This make repeated switching between files extremely confusing when 8-10 files are open.
2) when you click on a file in the outline explorer, the outline explorer always open the tree view and shows all the function and loops within that file. If you have >20 functions and loops in one file (which is totally common), all the other files are moved to the bottom off the screen. Switching back and forth becomes a hassle. My suggestions would be:  1) single click to change to the selected file without opening the tree view,  2) double click on a file to open the tree displaying the functions within that file.
Is that hard to implement?

ccordoba12:
This needs a rebase, and I think it's really important to finish it :-)

oldrichsmejkal:
What about this issue? It's pretty old, and it would be nice to have this features...

Thanks.

ccordoba12:
Closing because @goanpeca has a different approach to solve this. It'll be fixed in 3.1 ;-)

PR #2377, opened 2015-04-28
Fix the fact that spyder_win_post_install.py can't be run from pip

### LINKED ISSUE #2188: can't run win_post_install from pip
please add this snippet to spyder_win_post_install.py to mimic bdist_wininst. I cribbed it from pywin32_post_install.py. Thanks!

    try:
        # When this script is run from inside the bdist_wininst installer,
        # file_created() and directory_created() are additional builtin
        # functions which write lines to Python23\pywin32-install.log. This is
        # a list of actions for the uninstaller, the format is inspired by what
        # the Wise installer also creates.
        file_created
        is_bdist_wininst = True
    except NameError:
        is_bdist_wininst = False # we know what it is not - but not what it is :)
        def file_created(file):
            pass
        def directory_created(directory):
            pass
        def get_root_hkey():
            try:
                winreg.OpenKey(winreg.HKEY_LOCAL_MACHINE,
                               root_key_name, 0, winreg.KEY_CREATE_SUB_KEY)
                return winreg.HKEY_LOCAL_MACHINE
            except OSError, details:
                # Either not exist, or no permissions to create subkey means
                # must be HKCU
                return winreg.HKEY_CURRENT_USER

    try:
        create_shortcut
    except NameError:
        # Create a function with the same signature as create_shortcut provided
        # by bdist_wininst
        def create_shortcut(path, description, filename,
                            arguments="", workdir="", iconpath="", iconindex=0):
            import pythoncom
            from win32com.shell import shell, shellcon

            ilink = pythoncom.CoCreateInstance(shell.CLSID_ShellLink, None,
                                               pythoncom.CLSCTX_INPROC_SERVER,
                                               shell.IID_IShellLink)
            ilink.SetPath(path)
            ilink.SetDescription(description)
            if arguments:
                ilink.SetArguments(arguments)
            if workdir:
                ilink.SetWorkingDirectory(workdir)
            if iconpath or iconindex:
                ilink.SetIconLocation(iconpath, iconindex)
            # now save it.
            ipf = ilink.QueryInterface(pythoncom.IID_IPersistFile)
            ipf.Save(filename, 0)

        # Support the same list of "path names" as bdist_wininst.
        def get_special_folder_path(path_name):
            import pythoncom
            from win32com.shell import shell, shellcon

            for maybe in """
                CSIDL_COMMON_STARTMENU CSIDL_STARTMENU CSIDL_COMMON_APPDATA
                CSIDL_LOCAL_APPDATA CSIDL_APPDATA CSIDL_COMMON_DESKTOPDIRECTORY
                CSIDL_DESKTOPDIRECTORY CSIDL_COMMON_STARTUP CSIDL_STARTUP
                CSIDL_COMMON_PROGRAMS CSIDL_PROGRAMS CSIDL_PROGRAM_FILES_COMMON
                CSIDL_PROGRAM_FILES CSIDL_FONTS""".split():
                if maybe == path_name:
                    csidl = getattr(shellcon, maybe)
                    return shell.SHGetSpecialFolderPath(0, csidl, False)
            raise ValueError("%s is an unknown path ID" % (path_name,))

mikofski:
Fixes #2188

----

* add missing `bdist_wininst` functions from `pywin32_postinstall.py`:
 `file_created()`, `directory_created()`, `get_root_key()`, `create_shortcut()`
 and `get_special_folder_path()`
* test for `pythoncom` and return message "pywin32 required" on `ImportError`
* replace `CSIDL_COMMON_PROGRAMS` with `CSIDL_PROGRAMS` to be consistent with
 desktop shortcut which is in user profile, and so that user doesn't have
 to elevate rights in order to install Spyder
* add TODO to remove error message "try running installer as
 administrator" since it's unnecessary if admin rights not required.
* if running `spyder_win_post_install.py -install` manually after installing
 with pip, then `spyder.ico` and `spyder_light.ico` are in the
 `Python27\Scripts\` folder not in the distribution folder, so check if
 `ico_dir` exists and use absolute path to `spyder_win_post_install.__file__`
* clean up desktop and start menu shortcuts on removal
* provide feedback on successful removal of context, desktop and start
 menu shortcuts

Signed-off-by: Mark Mikofski <mark.mikofski@sunpowercorp.com>

### DIFF
--- a/scripts/spyder_win_post_install.py
+++ b/scripts/spyder_win_post_install.py
@@ -222,9 +222,7 @@ if __name__=='__main__':
             try:
                 install()
             except OSError:
-                # TODO: remove this because using user's startmenu not common
-                print("Failed to create Start Menu items, try running "\
-                      "installer as administrator.", file=sys.stderr)
+                print("Failed to create Start Menu items.", file=sys.stderr)
         elif sys.argv[1] == '-remove':
             remove()
         else:
--- a/scripts/spyder_win_post_install.py
+++ b/scripts/spyder_win_post_install.py
@@ -177,43 +177,45 @@ def remove():
         except WindowsError:
             pass
         else:
-            print("Successfully removed Spyder shortcuts from Windows "\
-                  "Explorer context menu.", file=sys.stdout)
-    # clean up desktop
-    desktop_folder = get_special_folder_path("CSIDL_DESKTOPDIRECTORY")
-    fname = osp.join(desktop_folder, 'Spyder.lnk')
-    if osp.isfile(fname):
-        try:
-            os.remove(fname)
-        except OSError:
-            print("Failed to remove %s; you may be able to remove it "\
-                  "manually." % fname, file=sys.stderr)
-        else:
-            print("Successfully removed Spyder shortcuts from your desktop.",
-                  file=sys.stdout)
-    # clean up startmenu
-    start_menu = osp.join(get_special_folder_path('CSIDL_PROGRAMS'),
-                          'Spyder (Py%i.%i %i bit)' % (sys.version_info[0],
-                                                       sys.version_info[1],
-                                                       struct.calcsize('P')*8))
-    if osp.isdir(start_menu):
-        for fname in os.listdir(start_menu):
+            if not is_bdist_wininst:
+                print("Successfully removed Spyder shortcuts from Windows "\
+                      "Explorer context menu.", file=sys.stdout)
+    if not is_bdist_wininst:
+        # clean up desktop
+        desktop_folder = get_special_folder_path("CSIDL_DESKTOPDIRECTORY")
+        fname = osp.join(desktop_folder, 'Spyder.lnk')
+        if osp.isfile(fname):
             try:
-                os.remove(osp.join(start_menu,fname))
+                os.remove(fname)
             except OSError:
                 print("Failed to remove %s; you may be able to remove it "\
                       "manually." % fname, file=sys.stderr)
             else:
-                print("Successfully removed Spyder shortcuts from your "\
+                print("Successfully removed Spyder shortcuts from your desktop.",
+                      file=sys.stdout)
+        # clean up startmenu
+        start_menu = osp.join(get_special_folder_path('CSIDL_PROGRAMS'),
+                              'Spyder (Py%i.%i %i bit)' % (sys.version_info[0],
+                                                           sys.version_info[1],
+                                                           struct.calcsize('P')*8))
+        if osp.isdir(start_menu):
+            for fname in os.listdir(start_menu):
+                try:
+                    os.remove(osp.join(start_menu,fname))
+                except OSError:
+                    print("Failed to remove %s; you may be able to remove it "\
+                          "manually." % fname, file=sys.stderr)
+                else:
+                    print("Successfully removed Spyder shortcuts from your "\
+                          " start menu.", file=sys.stdout)
+            try:
+                os.rmdir(start_menu)
+            except OSError:
+                print("Failed to remove %s; you may be able to remove it "\
+                      "manually." % fname, file=sys.stderr)
+            else:
+                print("Successfully removed Spyder shortcut folder from your "\
                       " start menu.", file=sys.stdout)
-        try:
-            os.rmdir(start_menu)
-        except OSError:
-            print("Failed to remove %s; you may be able to remove it "\
-                  "manually." % fname, file=sys.stderr)
-        else:
-            print("Successfully removed Spyder shortcut folder from your "\
-                  " start menu.", file=sys.stdout)
 
 
 if __name__=='__main__':
--- a/scripts/spyder_win_post_install.py
+++ b/scripts/spyder_win_post_install.py
@@ -21,17 +21,95 @@ KEY_C0 = KEY_C % r"Python.%sFile\shell\%s"
 KEY_C1 = KEY_C0 + r"\command"
 
 
+# ability to run spyder-win-post-install outside of bdist_wininst installer
+# copied from pywin32-win-post-install
+try:
+    # When this script is run from inside the bdist_wininst installer,
+    # file_created() and directory_created() are additional builtin
+    # functions which write lines to Python23\pywin32-install.log. This is
+    # a list of actions for the uninstaller, the format is inspired by what
+    # the Wise installer also creates.
+    file_created
+    is_bdist_wininst = True
+except NameError:
+    is_bdist_wininst = False # we know what it is not - but not what it is :)
+    def file_created(file):
+        pass
+    def directory_created(directory):
+        pass
+    def get_root_hkey():
+        try:
+            winreg.OpenKey(winreg.HKEY_LOCAL_MACHINE,
+                           root_key_name, 0, winreg.KEY_CREATE_SUB_KEY)
+            return winreg.HKEY_LOCAL_MACHINE
+        except OSError, details:
+            # Either not exist, or no permissions to create subkey means
+            # must be HKCU
+            return winreg.HKEY_CURRENT_USER
+try:
+    create_shortcut
+except NameError:
+    # Create a function with the same signature as create_shortcut provided
+    # by bdist_wininst
+    def create_shortcut(path, description, filename,
+                        arguments="", workdir="", iconpath="", iconindex=0):
+        try:
+            import pythoncom
+        except ImportError:
+            print("PyWin32 is required to run windows postinstall manually.",
+                  file=sys.stderr)
+            sys.exit(1)  # pywin32 required
+        from win32com.shell import shell, shellcon
+
+        ilink = pythoncom.CoCreateInstance(shell.CLSID_ShellLink, None,
+                                           pythoncom.CLSCTX_INPROC_SERVER,
+                                           shell.IID_IShellLink)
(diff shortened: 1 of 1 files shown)


scripts/spyder_win_post_install.py
```diff
@@ -21,17 +21,95 @@
 KEY_C1 = KEY_C0 + r"\command"
 
 
+# ability to run spyder-win-post-install outside of bdist_wininst installer
+# copied from pywin32-win-post-install
```
ccordoba12: Could you post here a link to Github (or Bitbucket, etc) where we can read `pywin32-win-post-install` code?

scripts/spyder_win_post_install.py
```diff
@@ -21,17 +21,95 @@
...
+    # a list of actions for the uninstaller, the format is inspired by what
+    # the Wise installer also creates.
+    file_created
+    is_bdist_wininst = True
+except NameError:
+    is_bdist_wininst = False # we know what it is not - but not what it is :)
+    def file_created(file):
+        pass
+    def directory_created(directory):
+        pass
```
ccordoba12: Why these couple of functions are empty?

scripts/spyder_win_post_install.py
mikofski: The [pywin32 mercurial repository is on sourceforge](http://pywin32.hg.sourceforge.net/hgweb/pywin32/pywin32). The most recent build is tagged as [b219](http://pywin32.hg.sourceforge.net/hgweb/pywin32/pywin32/file/b219). Here's a link to  [`pywin32_postinstall.py`](http://pywin32.hg.sourceforge.net/hgweb/pywin32/pywin32/file/b219/pywin32_postinstall.py).

scripts/spyder_win_post_install.py
mikofski: When the user runs the `spyder_win_post_install.py` manually after using pip to install Spyder, there isn't a registered uninstaller for Spyder. The uninstaller is only registered when using the `bdist_wininst` installer. The functions [`file_created()` and `directory_created()`](https://docs.python.org/2/distutils/builtdist.html#the-postinstallation-script) normally register files and directories to be removed with the uninstaller, but for pip users these functions should do nothing. Instead if the shortcut files and startmenu directory exist, they are removed in [L174](https://github.com/spyder-ide/spyder/pull/2377/files#diff-044873a159653f78071aafd4f514e937R174) and [L186](https://github.com/spyder-ide/spyder/pull/2377/files#diff-044873a159653f78071aafd4f514e937R186).

scripts/spyder_win_post_install.py
ccordoba12: Sorry for not being clear enough :-) Please add the link to the file, below the last line of your comment.

ccordoba12:
One thing that I'd like to know is if it's possible to tell pip to run this script after install/uninstall steps.

I mean, can't we add something to `setup.py` to make this happen automatically?

scripts/spyder_win_post_install.py
ccordoba12: ok

Nodd: The explanations should be added as comments in the code, it's far from obvious.

mikofski:
That would be very nice if the post install script could be run automatically, although I was thinking of making that a separate feature after this one gets pulled.

Another feature I thought would be nice is a post install script for other platforms that does the equivalent for linux and mac osx. _IE_ on linux it could create a desktop file in `.local/share/applications/` called `spyder.desktop` that would add spyder to the launcher. And in mac osx it could drop an .app file in the applications drawer that launched spyder instead of having to launch it from the terminal.

Back to your question, though, I don't know how to get pip to run the script. I don't if you can add it to `setup.py` but it seems plausible. It's a good place to start. But I'm okay with just running it manually. When I set up Python for other engineers here at SunPower, I usually use pip, and then manually make a shortcut to spyder on the desktop that runs `C:\Python27\python.exe C:\Python27\scripts\spyder` and then manually change the icon to `C:\Python27\scripts\spyder.ico` - so having this post installer will greatly improve the setup procedure since it will also add the Windows Explorer Context menu support.

Thanks!

scripts/spyder_win_post_install.py
mikofski: check out 330b0aa

scripts/spyder_win_post_install.py
mikofski: There is already a bit of explanation from Mark Hammond on [L27](https://github.com/spyder-ide/spyder/pull/2377#discussion-diff-29477366R27). I added a link to bdist_wininst postinstaller in the standard referencce library and additional comments requested by @Nodd  directly prior to the dummy method definitions in 330b0aa.

scripts/spyder_win_post_install.py
```diff
@@ -21,17 +21,100 @@
...
+    # isn't run from bdist_wininst installer, instead if shortcuts and start menu directory
+    # exist, they are removed when the post install script is called with the -remote option
+    def file_created(file):
+        pass
+    def directory_created(directory):
+        pass
+    def get_root_hkey():
+        try:
+            winreg.OpenKey(winreg.HKEY_LOCAL_MACHINE,
+                           root_key_name, 0, winreg.KEY_CREATE_SUB_KEY)
```
ccordoba12: `root_key_name` is not defined anywhere in your PR, so this call is going to fail, right?

scripts/spyder_win_post_install.py
```diff
@@ -87,6 +173,44 @@ def remove():
             winreg.DeleteKey(root, key)
         except WindowsError:
             pass
+        else:
+            print("Successfully removed Spyder shortcuts from Windows "\
+                  "Explorer context menu.", file=sys.stdout)
+    # clean up desktop
+    desktop_folder = get_special_folder_path("CSIDL_DESKTOPDIRECTORY")
```
ccordoba12: I think this code, i.e. from here to the end of `remove` should be ran only when `is_bdist_wininst == False`, right?

scripts/spyder_win_post_install.py
```diff
@@ -95,6 +219,7 @@ def remove():
             try:
                 install()
             except OSError:
+                # TODO: remove this because using user's startmenu not common
```
ccordoba12: Please finish this `TODO`

ccordoba12:
Ok, my final review is done. Sorry for taking so much time to do it :-)

Please try to address my comments later today or tomorrow morning, if you want to include this PR in **2.3.5**. Else, I'll leave it for **3.0**.

scripts/spyder_win_post_install.py
mikofski: @ccordoba12 good catch, on windows this key is `hkey_local_machine\Software\Python\PythonCore\2.7` for python installed with msi, however, I don't know what it would be on a windows system that has a bootstrapped Python, eg: WinPython, portable python or any python just unzipped in and executed from a standalone folder.

This is only run by `bdist_wininst` and it's sole purpose is to determine if the user has elevated their rights, if so then use `hkey_local_machine` and if not then use `hkey_current_user`. Therefore even if this call fails, it will just fall back to `hkey_current_user` which is what we want anyway, so it doesn't matter. But for those that have installed official python, and have admin rights, then it will use the system registry keys.

ccordoba12: Ok, then we can leave it without problems :-)

Please address my other comments :-)

mikofski:
Thanks @ccordoba12 for your review, I have addressed your current comments:

- [ ] define `root_key_name`
- [ ] complete TODO to remove "run -install with admin rights"
- [ ] only run manual -remove if `is_bdist_wininst == False`

scripts/spyder_win_post_install.py
```diff
@@ -21,17 +21,103 @@
...
+    def get_special_folder_path(path_name):
+        import pythoncom
+        from win32com.shell import shell, shellcon
+
+        for maybe in """
+            CSIDL_COMMON_STARTMENU CSIDL_STARTMENU CSIDL_COMMON_APPDATA
+            CSIDL_LOCAL_APPDATA CSIDL_APPDATA CSIDL_COMMON_DESKTOPDIRECTORY
+            CSIDL_DESKTOPDIRECTORY CSIDL_COMMON_STARTUP CSIDL_STARTUP
+            CSIDL_COMMON_PROGRAMS CSIDL_PROGRAMS CSIDL_PROGRAM_FILES_COMMON
+            CSIDL_PROGRAM_FILES CSIDL_FONTS""".split():
```
Nodd: Why not creating a list directly instead of splitting a string ?
Performance-wise it won't matter but it would be more obvious to understand.  I had to read these lines 3 times to spot the `.split()` here.

ccordoba12: +1 to this last comment from @Nodd.

ccordoba12:
@mikofski, thanks for the quick response! I'll merge it tomorrow and backport it to 2.3.5 (which will be released tomorrow too :-)

scripts/spyder_win_post_install.py
ccordoba12: I'll do this myself after merging.

ccordoba12:
@mikofski, I tested your work before releasing 2.3.5 and it's working great! Thanks a lot for it.

mikofski:
That's great news! Glad I could contribute.

PR #2399, opened 2015-05-04
Add "Find from here..." menu item for directories

techtonik:
This adds implicit dependency to findinfiles plugin/widget
which leads to spaghetti code if there are no instruments
to control them.

There SHOULD be a better way to do this. Maybe a convention
and a documentation. Or maybe command line tools (in case
Spyder itself won't start).


techtonik:
This was designed for stable branch. See #2400.

PR #2400, opened 2015-05-04
PR: Add "Find from here..." menu item for directories

techtonik:
This adds implicit dependency to findinfiles plugin/widget
which leads to spaghetti code if there are no instruments
to control them.

There SHOULD be a better way to do this. Maybe a convention
and a documentation. Or maybe command line tools (in case
Spyder itself won't start).

### DIFF
--- a/spyderlib/widgets/explorer.py
+++ b/spyderlib/widgets/explorer.py
@@ -296,6 +296,11 @@ class DirView(QTreeView):
                                triggered=lambda fnames=fnames:
                                self.open_interpreter(fnames))
         actions.append(action)
+        _title = _("Find from here...")
+        action = create_action(self, _title, icon="find.png",
+                               triggered=lambda:
+                               self.find_from_here(fnames))
+        actions.append(action)
         return actions
         
     def create_context_menu_actions(self):
@@ -437,7 +442,12 @@ class DirView(QTreeView):
         """Open interpreter"""
         for path in sorted(fnames):
             self.parent_widget.emit(SIGNAL("open_interpreter(QString)"), path)
-        
+
+    def find_from_here(self, fnames):
+        """Switch to find dialog"""
+        self.parent_widget.main.findinfiles.find_options.set_directory(fnames[0])
+        self.parent_widget.main.findinfiles.findinfiles_callback()
+
     def run(self, fnames=None):
         """Run Python scripts"""
         if fnames is None:


ccordoba12:
The way to do this and avoid the spaghetti code is:

1. Make the widget (i.e. `widgets/explorer.py`) emit a Qt signal on the `find_from_here` method.
2. Connect to this signal in the plugin (i.e. `plugins/explorer.py`), which has a reference to `main`, and so can connect directly to `findinfiles`.

This is what we do to handle things in the Editor, i.e. pass signals from the widgets to the plugin, and make the plugin connect to other plugins through `main`.

techtonik:
The question is who should provide the find_from_here method? It is not explorer function, but findinfiles plugin. If findinfiles is not available, there should not be menu item.

ccordoba12:
All plugins are available all the time, even if hidden. You can make them appear by calling the method `visibility_changed`.

So the steps I gave should work just fine :-)

techtonik:
This is still spaghetti code. =) I think that proper way is to have findplugin add its menus and actions. This way explorerplugin won't have to know anything about findplugin. We can further limit findplugin to use only known menu interface or API for adding stuff to menus.

ccordoba12:
Well, that's how we do things right now, so it's our current architecture.

In any case, I think the way i suggested you is easier to understand than your current code :-) If only because it fits Spyder model.

Finally, we're not going to re-architecture everything just for this change :-)

techtonik:
The problem, which widget / plugin is loaded first? Is that critical if findinfiles initializes first and reference explorer which is not loaded? Is there a diagram of the process?

Nodd:
If a widget needs another, it will import it first.
There is no diagram, but there are many python tools to create call graphs easily if it helps you.. A big diagram with something like 50 classes and 500 methods is not really helpful.

techtonik:
Like this one? http://techtonik.bitbucket.org/modulegraph/index.html

techtonik:
findinfiles doesn't need explorer and vice versa. It is an optional feature or addon for explorer widget if findinfiles plugin is present.

ccordoba12:
@techtonik, do you plan to address my comments? If not, I'm going to close this PR :-)

techtonik:
@ccordoba12 you're mean that comment? `Finally, we're not going to re-architecture everything just for this change :-)` No. Sorry. I am working on a new architecture of plugins, so that it could be made an external module that is not loaded all the time, and this is my use case. Let me keep it open.

ccordoba12:
These are my comments: https://github.com/spyder-ide/spyder/pull/2400#issuecomment-99581869

ccordoba12:
This has been stalled for along time, so I'm closing it.

@techtonik, please reopen it when/if you have something new to add.

PR #2411, opened 2015-05-06
Improve github issue template

Nodd:
Templates for issues in github had a few problems:
* The list of steps was mangled with the title and was also badly displayed is some cases
* The list of dependencies was hard to read

### DIFF
--- a/spyderlib/dependencies.py
+++ b/spyderlib/dependencies.py
@@ -71,7 +71,6 @@ def add(modname, features, required_version, installed_version=None):
 
 def check(modname):
     """Check if required dependency is installed"""
-    global DEPENDENCIES
     for dependency in DEPENDENCIES:
         if dependency.modname == modname:
             return dependency.check()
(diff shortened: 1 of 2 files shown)


spyderlib/dependencies.py
```diff
@@ -80,7 +79,6 @@ def check(modname):
 
 def status():
     """Return a complete status of Optional Dependencies"""
-    global DEPENDENCIES
```
ccordoba12: Why removing these two references to `DEPENDENCIES` as part of this PR?

Nodd:
Because at first I wanted to use a list for the dependencies so I looked how `dependencies.status()` was coded. This is a simple change to clean up the codebase a bit so I included it anyway. `global` is needed only when defining new variables, not when using them.

spyderlib/dependencies.py
Nodd: Because at first I wanted to use a list for the dependencies so I looked how dependencies.status() was coded. This is a simple change to clean up the codebase a bit so I included it anyway. global is needed only when defining new variables, not when using them.

ccordoba12:
Thanks! Things look much better with your improvements :-)

PR #2412, opened 2015-05-06
Open preferences dialog even if a plugin raises errors

Nodd:
If a (third-party) plugin fails to create its preferences page, it prevents the dialog to be created at all.

This PR allows to skip plugins that raise errors. The errors are still displayed in the intenal console, but the loop can continue (show must go on!).


ccordoba12:
This is working as expected, thanks!

PR #2423, opened 2015-05-07
Reset spyder and restart from within running application

### LINKED ISSUE #2401: Unable to reset settings from the Main Window
Hi,

I have tried reseting the User interface i.e. windows layout from view but I cannot get back to original. 
I have removed the Spyder-Py2.app and reinstalled it but it doesn't restore to default. Also, I have tried using Anaconda but still the same problem. 
screen shot 2015-05-04 at 6 39 11 pm

Any suggestion on proper removal of Spyder or setting it to default. Thanks.

goanpeca:
Fixes #2401

---
## Description

Adds a new entry on the tools menu for reseting spyder from within the application

image

Or as an option in preferences... decision pending
image
After clicking a message box will ask to confirm, probably the message can be more explanatory so the user knows what he\she is about to do.

image

## Todo

- [x] Agree on location of option
- [x] Agree on text of the new option

### DIFF
--- a/spyderlib/plugins/configdialog.py
+++ b/spyderlib/plugins/configdialog.py
@@ -24,9 +24,11 @@ import spyderlib.utils.icon_manager as ima
 
 from spyderlib.baseconfig import (_, running_in_mac_app, LANGUAGE_CODES,
                                   save_lang_conf, load_lang_conf)
+from spyderlib.baseconfig import _, running_in_mac_app
 from spyderlib.config import CONF
 from spyderlib.guiconfig import (CUSTOM_COLOR_SCHEME_NAME,
                                  set_default_color_scheme)
+from spyderlib.utils.qthelpers import get_icon, get_std_icon
 from spyderlib.userconfig import NoDefault
 from spyderlib.widgets.colors import ColorLayout
 from spyderlib.widgets.sourcecode import syntaxhighlighters as sh
@@ -106,6 +108,7 @@ class ConfigPage(QWidget):
                     break  # Ensure a single popup is displayed
             self.set_modified(False)
 
+    
     def load_from_conf(self):
         """Load settings from configuration file"""
         raise NotImplementedError
@@ -124,6 +127,21 @@ class ConfigDialog(QDialog):
     
     def __init__(self, parent=None):
         QDialog.__init__(self, parent)
+
+        self._parent = parent
+        self.main = parent
+
+        # Widgets
+        self.pages_widget = QStackedWidget()
+        self.contents_widget = QListWidget()
+        self.button_reset = QPushButton(_('Reset to defaults'))
+
+        bbox = QDialogButtonBox(QDialogButtonBox.Ok|QDialogButtonBox.Apply
+                                |QDialogButtonBox.Cancel)
+        self.apply_btn = bbox.button(QDialogButtonBox.Apply)
+
+        # Widgets setup
+
         
         # Destroying the C++ object right after closing the dialog box,
         # otherwise it may be garbage-collected in another QThread
@@ -154,6 +172,7 @@ class ConfigDialog(QDialog):
         hsplitter.addWidget(self.pages_widget)
 
         btnlayout = QHBoxLayout()
+        btnlayout.addWidget(self.button_reset)
         btnlayout.addStretch(1)
         btnlayout.addWidget(bbox)
 
@@ -169,6 +188,15 @@ class ConfigDialog(QDialog):
         # Ensures that the config is present on spyder first run
         CONF.set('main', 'interface_language', load_lang_conf())
         
+        # Signals and slots
+        self.button_reset.clicked.connect(self.main.reset_spyder)
+        self.pages_widget.currentChanged.connect(self.current_page_changed)
+        self.contents_widget.currentRowChanged.connect(
+                                             self.pages_widget.setCurrentIndex)
+        bbox.accepted.connect(self.accept)
+        bbox.rejected.connect(self.reject)
+        bbox.clicked.connect(self.button_clicked)
+         
     def get_current_index(self):
         """Return current page index"""
         return self.contents_widget.currentRow()
--- a/spyderlib/plugins/configdialog.py
+++ b/spyderlib/plugins/configdialog.py
@@ -132,38 +132,23 @@ class ConfigDialog(QDialog):
         self.contents_widget = QListWidget()
         self.button_reset = QPushButton(_('Reset to defaults'))
 
-        bbox = QDialogButtonBox(QDialogButtonBox.Ok|QDialogButtonBox.Apply
-                                |QDialogButtonBox.Cancel)
+        bbox = QDialogButtonBox(QDialogButtonBox.Ok | QDialogButtonBox.Apply |
+                                QDialogButtonBox.Cancel)
         self.apply_btn = bbox.button(QDialogButtonBox.Apply)
 
         # Widgets setup
-
-        
         # Destroying the C++ object right after closing the dialog box,
         # otherwise it may be garbage-collected in another QThread
         # (e.g. the editor's analysis thread in Spyder), thus leading to
         # a segmentation fault on UNIX or an application crash on Windows
         self.setAttribute(Qt.WA_DeleteOnClose)
-
-        self.button_reset = QPushButton(_("Reset to defaults"))
-        self.contents_widget = QListWidget()
+        self.setWindowTitle(_('Preferences'))
+        self.setWindowIcon(ima.icon('configure'))
         self.contents_widget.setMovement(QListView.Static)
         self.contents_widget.setSpacing(1)
-
-        bbox = QDialogButtonBox(QDialogButtonBox.Ok|QDialogButtonBox.Apply
-                                |QDialogButtonBox.Cancel)
-        self.apply_btn = bbox.button(QDialogButtonBox.Apply)
-        bbox.accepted.connect(self.accept)
-        bbox.rejected.connect(self.reject)
-        bbox.clicked.connect(self.button_clicked)
-        
-        self.pages_widget = QStackedWidget()
-        self.pages_widget.currentChanged.connect(self.current_page_changed)
-
-        self.contents_widget.currentRowChanged.connect(
-                                             self.pages_widget.setCurrentIndex)
         self.contents_widget.setCurrentRow(0)
 
+        # Layout
         hsplitter = QSplitter()
         hsplitter.addWidget(self.contents_widget)
         hsplitter.addWidget(self.pages_widget)
@@ -179,14 +164,6 @@ class ConfigDialog(QDialog):
 
         self.setLayout(vlayout)
 
-        self.setWindowTitle(_('Preferences'))
-        self.setWindowIcon(ima.icon('configure'))
-
-        self.button_reset.clicked.connect(self.main.reset_spyder)
-
-        # Ensures that the config is present on spyder first run
-        CONF.set('main', 'interface_language', load_lang_conf())
-        
         # Signals and slots
         self.button_reset.clicked.connect(self.main.reset_spyder)
         self.pages_widget.currentChanged.connect(self.current_page_changed)
@@ -195,7 +172,10 @@ class ConfigDialog(QDialog):
         bbox.accepted.connect(self.accept)
         bbox.rejected.connect(self.reject)
         bbox.clicked.connect(self.button_clicked)
-         
+
+        # Ensures that the config is present on spyder first run
+        CONF.set('main', 'interface_language', load_lang_conf())
+
     def get_current_index(self):
         """Return current page index"""
         return self.contents_widget.currentRow()
--- a/spyderlib/plugins/configdialog.py
+++ b/spyderlib/plugins/configdialog.py
@@ -128,7 +128,6 @@ class ConfigDialog(QDialog):
     def __init__(self, parent=None):
         QDialog.__init__(self, parent)
 
-        self._parent = parent
         self.main = parent
 
         # Widgets
--- a/spyderlib/plugins/configdialog.py
+++ b/spyderlib/plugins/configdialog.py
@@ -24,11 +24,9 @@ import spyderlib.utils.icon_manager as ima
 
 from spyderlib.baseconfig import (_, running_in_mac_app, LANGUAGE_CODES,
                                   save_lang_conf, load_lang_conf)
-from spyderlib.baseconfig import _, running_in_mac_app
 from spyderlib.config import CONF
 from spyderlib.guiconfig import (CUSTOM_COLOR_SCHEME_NAME,
                                  set_default_color_scheme)
-from spyderlib.utils.qthelpers import get_icon, get_std_icon
 from spyderlib.userconfig import NoDefault
 from spyderlib.widgets.colors import ColorLayout
 from spyderlib.widgets.sourcecode import syntaxhighlighters as sh
@@ -108,7 +106,6 @@ class ConfigPage(QWidget):
                     break  # Ensure a single popup is displayed
             self.set_modified(False)
 
-    
     def load_from_conf(self):
         """Load settings from configuration file"""
         raise NotImplementedError
--- a/spyderlib/plugins/configdialog.py
+++ b/spyderlib/plugins/configdialog.py
@@ -149,6 +149,7 @@ class ConfigDialog(QDialog):
         # a segmentation fault on UNIX or an application crash on Windows
         self.setAttribute(Qt.WA_DeleteOnClose)
 
+        self.button_reset = QPushButton(_("Reset to defaults"))
         self.contents_widget = QListWidget()
         self.contents_widget.setMovement(QListView.Static)
         self.contents_widget.setSpacing(1)
@@ -184,6 +185,8 @@ class ConfigDialog(QDialog):
 
         self.setWindowTitle(_('Preferences'))
         self.setWindowIcon(ima.icon('configure'))
+
+        self.button_reset.clicked.connect(self.main.reset_spyder)
 
         # Ensures that the config is present on spyder first run
         CONF.set('main', 'interface_language', load_lang_conf())
--- a/spyderlib/restart_app.py
+++ b/spyderlib/restart_app.py
@@ -10,7 +10,7 @@ Restart Spyder
 
 A helper script that allows to restart (and also reset) Spyder from within the
 running application.
-my """
+"""
 
 import ast
 import os
--- a/spyderlib/restart_app.py
+++ b/spyderlib/restart_app.py
@@ -237,7 +237,7 @@ def main():
     # Before launching a new Spyder instance we need to make sure that the
     # previous one has closed. We wait for a fixed and "reasonable" amount of
     # time and check, otherwise an error is launched
-    wait_time = 60*5 if IS_WINDOWS else 60*3 # Seconds
+    wait_time = 90 if IS_WINDOWS else 30  # Seconds
     for counter in range(int(wait_time/SLEEP_TIME)):
         if not is_pid_running(pid):
             break
--- a/spyderlib/restart_app.py
+++ b/spyderlib/restart_app.py
@@ -17,6 +17,7 @@ import os.path as osp
 import subprocess
 import sys
 import time
+import traceback
 
 
 PY2 = sys.version[0] == '2'
@@ -157,7 +158,7 @@ def main():
             pid_reset = p.pid
         except Exception as error:
             print(command)
-            print(error)
+            traceback.print_exc()
             time.sleep(15)
 
         # Wait for reset process to end before restarting
--- a/spyderlib/restart_app.py
+++ b/spyderlib/restart_app.py
@@ -21,7 +21,7 @@ import time
 
 PY2 = sys.version[0] == '2'
 CLOSE_ERROR, RESET_ERROR, RESTART_ERROR = list(range(3))
-
+SLEEP_TIME = 0.2
 
 def _is_pid_running_on_windows(pid):
     """Check if a process is running on windows systems based on the pid."""
@@ -69,7 +69,7 @@ def launch_error_message(type_, error=None):
 
     Parameters
     ----------
-    type : int [CLOSE_ERROR, RESET_ERROR, RESTART_ERROR]
+    type_ : int [CLOSE_ERROR, RESET_ERROR, RESTART_ERROR]
     """
     from spyderlib.baseconfig import _
     from spyderlib.qt.QtGui import QMessageBox, QDialog
@@ -180,10 +180,10 @@ def main():
     shell = os.name != 'nt'
 
     # Wait for original process to end before launching the new instance
-    for counter in range(1500):  # This is equivalent to ~ 5 minute (1500*0.2)
+    for counter in range(range(60*5/SLEEP_TIME)):  # Number in seconds (60*5)
         if not is_pid_running(pid):
             break
-        time.sleep(0.2)  # Throttling control
+        time.sleep(SLEEP_TIME)  # Throttling control
     else:
         # The old spyder instance took too long to close and restart aborts
         launch_error_message(type_=CLOSE_ERROR)
@@ -207,10 +207,10 @@ def main():
             launch_error_message(type_=RESET_ERROR, error=error)
 
         # Wait for reset process to end before restarting
-        for counter in range(300):  # This is ~= 1 minute (300*0.2)
+        for counter in range(60/SLEEP_TIME):  # Number in seconds (60)
             if not is_pid_running(pid_reset):
                 break
-            time.sleep(0.2)  # Throttling control
+            time.sleep(SLEEP_TIME)  # Throttling control
         else:
             # The rest subprocess took too long and it is killed
             p.kill()
--- a/spyderlib/restart_app.py
+++ b/spyderlib/restart_app.py
@@ -21,7 +21,8 @@ import time
 
 PY2 = sys.version[0] == '2'
 CLOSE_ERROR, RESET_ERROR, RESTART_ERROR = list(range(3))
-SLEEP_TIME = 0.2
+SLEEP_TIME = 0.2  # Seconds
+
 
 def _is_pid_running_on_windows(pid):
     """Check if a process is running on windows systems based on the pid."""
@@ -189,7 +190,6 @@ def main():
         launch_error_message(type_=CLOSE_ERROR)
 
     env = os.environ.copy()
-
 
     # Reset spyder if needed
     # -------------------------------------------------------------------------
--- a/spyderlib/restart_app.py
+++ b/spyderlib/restart_app.py
@@ -219,7 +219,7 @@ def main():
     # Restart
     # -------------------------------------------------------------------------
     try:
-        p = subprocess.Popen(command, shell=shell, env=env)
+        subprocess.Popen(command, shell=shell, env=env)
     except Exception as error:
         launch_error_message(type_=RESTART_ERROR, error=error)
 
--- a/spyderlib/restart_app.py
+++ b/spyderlib/restart_app.py
@@ -153,6 +153,7 @@ def main():
         # Try to reset
         try:
             p = subprocess.Popen(command_reset, shell=shell)
+            p.communicate()
             pid_reset = p.pid
         except Exception as error:
             print(command)
--- a/spyderlib/restart_app.py
+++ b/spyderlib/restart_app.py
@@ -193,8 +193,6 @@ def main():
 
     # Reset spyder if needed
     # -------------------------------------------------------------------------
-    reset_ok = True
-
     if reset:
         command_reset = '"{0}" "{1}" {2}'.format(python, spyder, args_reset)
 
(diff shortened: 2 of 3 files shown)


Nodd:
I would put it in the general tab in the preferences window.

goanpeca:
That feels weird, cause the preferences tab defines behaviors not actions.

This is an action, and it would be nice if it has a straightforward place for it.

Nodd:
Yeah, but it is an action applied to the behaviors !
It could be added as another pushbutton along with Apply, OK and Cancel.

goanpeca:
In that case I prefer adding the extra button.

---

Opinions @spyder-ide/core-developers?

goanpeca:
@blink1073, @SylvainCorlay, @ccordoba12 any comments on this one?

blink1073:
I like it as part of the Tools menu, as it is a global action.

goanpeca:
@sylvaincorlay? @ccordoba12 comments?

Nodd:
I really think that resetting the preferences should be done from the preference window. That's the same reason why resetting shortcuts is done from the shortcuts tab.

goanpeca:
I understand my dear, that is why I said, if it was going to be there, then it should be a button next to apply, ok, cancel, etc....

But I would like to hear from @SylvainCorlay and @ccordoba12

goanpeca:
@Nodd there might be a reason to keep the restart option outside. Sometimes if there is an error or conflict the preferences dialog might not open, in which case restarting and reseting could not be done from within the gui.

Nodd:
#2412 should lower the probability of this happening. Also it's still possible to use the commandline switch or remove the directory by hand  (of course it's less user-friendly).

goanpeca:
Ok @spyder-ide/core-developers this is the other option for the button.

This is a very small PR I would like to get off my back so please take a look at:

- Location. Tools menu or Preferences dialog?, or both.
    - image

    - image

- Text of message
- Any comments on code?

Cheers

spyderlib/plugins/configdialog.py
```diff
@@ -943,4 +950,4 @@ def apply_settings(self, options):
         if self.main.historylog is not None:
             self.main.historylog.apply_plugin_settings(['color_scheme_name'])
         if self.main.inspector is not None:
-            self.main.inspector.apply_plugin_settings(['color_scheme_name'])
+            self.main.inspector.apply_plugin_settings(['color_scheme_name'])
```
Nodd: Missing a newline here

goanpeca: oops

goanpeca: Fixed

goanpeca:
@Nodd Fixed

spyderlib/restart_app.py
```diff
@@ -135,6 +144,29 @@ def main():
...
+        command_reset = '"{0}" "{1}" {2}'.format(python, spyder, args_reset)
+
+        # Try to reset
+        try:
+            p = subprocess.Popen(command_reset, shell=shell)
+            p.communicate()
+            pid_reset = p.pid
+        except Exception as error:
+            print(command)
+            print(error)
```
Nodd: Using the `traceback` module would be nicer here, since it prints the full traceback as if the Exception wasn't catched. You can use [`traceback.print_exc()`](https://docs.python.org/3.4/library/traceback.html#traceback.print_exc) with no arguments, it prints the right thing at the right place.

spyderlib/restart_app.py
```diff
@@ -135,6 +144,29 @@ def main():
...
+            time.sleep(15)
```
Nodd: 15s, really ?

spyderlib/restart_app.py
```diff
@@ -135,6 +144,29 @@ def main():
...
+
+        # Wait for reset process to end before restarting
+        while True:
```
Nodd: This is potentially an infinite loop. Should we limit it to 30s or something, and warn the suer that there was an error ?
On the other hand the possibility of infinite loop is really small...

spyderlib/spyder.py
```diff
@@ -713,6 +713,9 @@ def create_edit_action(text, tr_text, icon):
                                                   module_completion.reset(),
                                         tip=_("Refresh list of module names "
                                               "available in PYTHONPATH"))
+            reset_spyder_action = create_action(self,
+                                        _("Reset Spyder to factory defaults"),
```
Nodd: That's not a good pep8 alignment here

spyderlib/restart_app.py
goanpeca: The idea was to be able to see the command line if an error occurs... otherwise it will close too fast... I could reduce to 10... ?

spyderlib/restart_app.py
goanpeca: Great, I will do :)

spyderlib/spyder.py
goanpeca: Ok, will fiix :)

spyderlib/restart_app.py
goanpeca: Well actually it can happen, good catch, so maybe hmmmm 50? that would be like 10 seconds... but it might take longer to restart on windows ... sometimes windows is weird...and takes a long time for a (spyder) process to die

goanpeca:
@Nodd so you suggest something regarding the "infinite" loop?... or should we just let it roll and see how it works?

spyderlib/spyder.py
```diff
@@ -713,6 +713,9 @@ def create_edit_action(text, tr_text, icon):
                                                   module_completion.reset(),
                                         tip=_("Refresh list of module names "
                                               "available in PYTHONPATH"))
+            reset_spyder_action = create_action(
+                self, _("Reset Spyder to factory defaults"),
+                triggered=lambda: self.reset_spyder())
```
Nodd: You can avoid the lambda here I think, using `self.reset_spyder` directly

spyderlib/plugins/configdialog.py
```diff
@@ -124,13 +124,17 @@ class ConfigDialog(QDialog):
     
     def __init__(self, parent=None):
         QDialog.__init__(self, parent)
-        
+
+        self._parent = parent
```
Nodd: I think you don't need this line, since `self.main` is used.

Nodd:
I think the correct behaviour is to wait something like 30s, and if it's not finished, kill the process, stop the restart and display a QMessageBox with an error.
I'm afraid that if we let the infinite loop it'll bite us sometimes in the future for whatever reason...

spyderlib/restart_app.py
```diff
@@ -135,6 +145,29 @@ def main():
...
+
+        # Try to reset
+        try:
+            p = subprocess.Popen(command_reset, shell=shell)
+            p.communicate()
+            pid_reset = p.pid
+        except Exception as error:
+            print(command)
+            traceback.print_exc()
+            time.sleep(15)
```
Nodd: If there is a problem here, the restart should be stopped instead of waiting 15s. This way the error will show up in the internal console.
Actually that means that you don't need to use `traceback.print_exc()`, just reraise the exception!

spyderlib/spyder.py
goanpeca: True!

spyderlib/plugins/configdialog.py
goanpeca: Gotcha

spyderlib/restart_app.py
goanpeca: I am not sure I get you... by internal console you mean? at this point we already left the initial spyder instance and we are inside the restart script. I will add an extra flag so that the restart stops and a QMessageBox displays something.

goanpeca:
Ok @Nodd I added a series of checks for better error handling which include a display of a message box in case errors occur, including close, reset and restart errors. Errors catched in the try,except part will also display the actual error. Errors based on too long running time will not display the error, but a friendly message....

spyderlib/restart_app.py
```diff
@@ -129,18 +180,60 @@ def main():
     shell = os.name != 'nt'
 
     # Wait for original process to end before launching the new instance
-    while True:
+    max_counter = 1500  # Max loops to perform before aborting
+    counter = 0
+    while counter < max_counter:  # This is equivalent to ~ 5 minute (1500*0.2)
```
Nodd: A `for` loop maybe ? ;)

spyderlib/restart_app.py
```diff
@@ -129,18 +180,60 @@ def main():
...
-    while True:
+    max_counter = 1500  # Max loops to perform before aborting
+    counter = 0
+    while counter < max_counter:  # This is equivalent to ~ 5 minute (1500*0.2)
         if not is_pid_running(pid):
             break
         time.sleep(0.2)  # Throttling control
+        counter += 1
+
+    if counter == max_counter:
```
Nodd: With a for loop you can just use `else` here. The `else` part is run if the loop is not broken.

goanpeca: You read my mind... I was just doing this

spyderlib/restart_app.py
```diff
@@ -129,18 +180,60 @@ def main():
...
+            p.communicate()
+            pid_reset = p.pid
+        except Exception as error:
+            p.kill()
+            error_message(type_='reset', error=error)
+
+        counter = 0
+        max_counter = 50
+        # Wait for reset process to end before restarting
+        while counter < max_counter:  # This is ~= 10 seconds (50*0.2)
```
Nodd: Same here for the `for` loop and the `else`

spyderlib/restart_app.py
```diff
@@ -129,18 +180,60 @@ def main():
...
+    # -------------------------------------------------------------------------
+    if reset_ok:
+        # Try to restart
+        try:
+            p = subprocess.Popen(command, shell=shell, env=env)
+        except Exception as error:
+            p.kill()
+            error_message(type_='restart', error=error)
+    else:
+        error_message(type_='reset')
```
Nodd: Why do this here and not when you set `reset_ok = False` ?

spyderlib/restart_app.py
```diff
@@ -129,18 +181,51 @@ def main():
     shell = os.name != 'nt'
 
     # Wait for original process to end before launching the new instance
-    while True:
+    for counter in range(range(60*5/SLEEP_TIME)):  # Number in seconds (60*5)
```
Nodd: Too many ranges here !

spyderlib/restart_app.py
goanpeca: I have fixed this already

spyderlib/restart_app.py
goanpeca: Cause I have two places to define `reset_ok = False`, one for an exception and the other for timeout... so I guess this looked cleaner to me as opposed, to writing `error_message(type_='reset')` in both places (where I define `reset_ok = False`)

spyderlib/restart_app.py
goanpeca: Shit!

spyderlib/restart_app.py
```diff
@@ -129,18 +181,46 @@ def main():
...
+            p = subprocess.Popen(command_reset, shell=shell)
+            p.communicate()
+            pid_reset = p.pid
+        except Exception as error:
+            p.kill()
+            launch_error_message(type_=RESET_ERROR, error=error)
+
+        # Wait for reset process to end before restarting
+        for counter in range(60/SLEEP_TIME):  # Number in seconds (60)
+            if not is_pid_running(pid_reset):
```
Nodd: If there was an exception before, `pid_reset` won't be defined.
I think this block should go in an `else:` statement after the `try:` and `except:`, since you shouldn''t have to wait if there was an error (I guess).

spyderlib/restart_app.py
```diff
@@ -129,18 +181,46 @@ def main():
...
+    # -------------------------------------------------------------------------
+    if reset:
+        command_reset = '"{0}" "{1}" {2}'.format(python, spyder, args_reset)
+
+        try:
+            p = subprocess.Popen(command_reset, shell=shell)
+            p.communicate()
+            pid_reset = p.pid
+        except Exception as error:
+            p.kill()
```
Nodd: `p` could be undefined if the error comes from `Popen`.
Error checking is complicated, even with exceptions !

Nodd:
Sorry for all the nitpicks...

spyderlib/restart_app.py
goanpeca: Indeed, I will make the adjustments, good suggestions

goanpeca:
@Nodd, do not worry about the nitpicks, I prefer comments to no comment at all !

goanpeca:
@Nodd no more nitpicking :-) ?

Nodd:
Nope, It looks fine ! :)
I didn't test it for real, however.

goanpeca:
@ccordoba12 any comments here?

spyderlib/restart_app.py
```diff
@@ -20,6 +20,8 @@
 
 
 PY2 = sys.version[0] == '2'
+CLOSE_ERROR, RESET_ERROR, RESTART_ERROR = list(range(1, 4, 1))
```
ccordoba12: Please add a comment above this line to explan why this number error codes mean

spyderlib/restart_app.py
```diff
@@ -63,6 +65,47 @@ def is_pid_running(pid):
         return _is_pid_running_on_unix(pid)
 
 
+def launch_error_message(type_, error=None):
```
ccordoba12: `type_` seems an ugly error variable name to me. Perhaps `error_type` would be better?

spyderlib/restart_app.py
```diff
@@ -63,6 +65,47 @@ def is_pid_running(pid):
...
+    ----------
+    type_ : int [CLOSE_ERROR, RESET_ERROR, RESTART_ERROR]
+    """
+    from spyderlib.baseconfig import _
+    from spyderlib.qt.QtGui import QMessageBox, QDialog
+    from spyderlib.utils import icon_manager as ima
+    from spyderlib.utils.qthelpers import qapplication
+
+    messages = {CLOSE_ERROR: _("The previous instance of Spyder has not "
+                               "closed.\nRestart aborted."),
```
ccordoba12: Perhaps a better message for this error would be:

    It was not possible to close the previous Spyder instance.
    Restart aborted

spyderlib/restart_app.py
```diff
@@ -63,6 +65,47 @@ def is_pid_running(pid):
...
+                RESET_ERROR: _("Spyder could not reset to factory defaults.\n"
+                               "Restart aborted."),
+                RESTART_ERROR: _("Spyder could not restart.\n"
+                                 "Restart aborted.")}
```
ccordoba12: Another suggestion

    It was not possible to restart Spyder.
    Operation aborted.

spyderlib/restart_app.py
```diff
@@ -129,18 +181,47 @@ def main():
     shell = os.name != 'nt'
 
     # Wait for original process to end before launching the new instance
-    while True:
+    wait_time = 60*5  # Seconds
+    for counter in range(int(wait_time/SLEEP_TIME)):
```
ccordoba12: What does this `for` loop exactly do? Could you add a comment below

    # Wait for original process ...

to say something about it?

ccordoba12:
Ok, first review done.

ccordoba12:
@goanpeca, about your first TODO: I think both locations are fine.

spyderlib/restart_app.py
Nodd: There is a comment one line above :)

ccordoba12: Yeah, but I'd like to see a more detailed explanation.

ccordoba12:
@goanpeca, could you address my comments so we can merge this one?

goanpeca:
Oops.. I did something stupid

goanpeca:
@ccordoba12 fixed!

spyderlib/restart_app.py
```diff
@@ -20,6 +20,11 @@
 
 
 PY2 = sys.version[0] == '2'
+SLEEP_TIME = 0.2  # Seconds for throttling control
+
+# Possible error codes when restarting/reseting spyder
+# Start in 1 cause the error code is used in an if clause
+CLOSE_ERROR, RESET_ERROR, RESTART_ERROR = list(range(1, 4, 1))
```
ccordoba12: This is bit weird to read :-) What about

```python
CLOSE_ERROR, RESET_ERROR, RESTART_ERROR = [1, 2, 3]
```

spyderlib/restart_app.py
```diff
@@ -63,6 +68,50 @@ def is_pid_running(pid):
...
+    from spyderlib.baseconfig import _
+    from spyderlib.qt.QtGui import QMessageBox, QDialog
+    from spyderlib.utils import icon_manager as ima
+    from spyderlib.utils.qthelpers import qapplication
+
+    messages = {CLOSE_ERROR: _("It was not possible to close the previous "
+                               "Spyder instance.\nRestart aborted."),
+                RESET_ERROR: _("Spyder could not reset to factory defaults.\n"
+                               "Restart aborted."),
+                RESTART_ERROR: _("It was not possible restart Spyder.\n"
```
ccordoba12: It was not possible **to** restart

spyderlib/restart_app.py
```diff
@@ -89,29 +138,38 @@ def main():
...
         error = "This script can only be called from within a Spyder instance"
         raise RuntimeError(error)
 
     # Variables were stored as string literals in the environment, so to use
     # them we need to parse them in a safe manner.
     is_bootstrap = ast.literal_eval(is_bootstrap)
-    pid = int(pid)
+    pid = ast.literal_eval(pid)
     args = ast.literal_eval(spyder_args)
+    reset = ast.literal_eval(reset)
```
ccordoba12: What `reset` evals to? `True` or `False`? Just curious :-)

spyderlib/restart_app.py
```diff
@@ -128,19 +186,52 @@ def main():
...
+    # time and check, otherwise an error is launched
+    wait_time = 60*5  # Seconds
+    for counter in range(int(wait_time/SLEEP_TIME)):
         if not is_pid_running(pid):
             break
-        time.sleep(0.2)  # Throttling control
+        time.sleep(SLEEP_TIME)  # Throttling control
+    else:
+        # The old spyder instance took too long to close and restart aborts
+        launch_error_message(type_=CLOSE_ERROR)
```
ccordoba12: Now the arg is called `error_type`, not `type_`

spyderlib/restart_app.py
```diff
@@ -128,19 +186,52 @@ def main():
     # Adjust the command and/or arguments to subprocess depending on the OS
     shell = os.name != 'nt'
 
-    # Wait for original process to end before launching the new instance
-    while True:
+    # Before launching a new Spyder instance we need to make sure that the
+    # previous one has closed. We wait for a fixed and "reasonable" amount of
+    # time and check, otherwise an error is launched
+    wait_time = 60*5  # Seconds
```
ccordoba12: So we're waiting for 5 min. Isn't that way too much time? I mean, specially if we are not giving our users any visual feedback about what's happening :-)

Two questions:
1. Isn't 2 or 3 min at most enough?
2. Can't we show a throbber or something while doing this? I mean, something like we already have when creating kernels for our IPython consoles?

spyderlib/restart_app.py
ccordoba12: No problem, I just saw below that this is the case ;-)

spyderlib/restart_app.py
goanpeca: Oops

spyderlib/restart_app.py
goanpeca: Fair enough

spyderlib/restart_app.py
goanpeca: Oops! 2 am coding!

spyderlib/restart_app.py
goanpeca: :-)

spyderlib/restart_app.py
goanpeca: 1. I guess, but windows sometimes takes sooooooo long!
2. Sure we could, what would you like to see? a "splash screen saying, Spyder is restarting and the throbber?

goanpeca:
@ccordoba12 fixed your comments and updated the code to show the spyder Splash screen with animated ellipsis.

ccordoba12:
This is quite neat!! Thanks a lot for following my suggestion ;-)

One last suggestion: Would it be possible to offer our users the possibility to drop the resetting and restarting process, or continue waiting? This could be done with a `QMessageBox` after 1 min on Linux/Mac and 2 min on Windows.

What do you think?

goanpeca:
Sure we can do that, do you want this as a separate PR? or into this one?

ccordoba12:
I'd like to have it in this one, please :-)

spyderlib/restart_app.py
```diff
@@ -84,34 +93,116 @@ def to_text_string(obj, encoding=None):
...
     spyder_args = os.environ.pop('SPYDER_ARGS', None)
     pid = os.environ.pop('SPYDER_PID', None)
     is_bootstrap = os.environ.pop('SPYDER_IS_BOOTSTRAP', None)
+    reset = os.environ.pop('SPYDER_RESET', None)
 
     # Get the spyder base folder based on this file
     spyder_folder = osp.split(osp.dirname(osp.abspath(__file__)))[0]
 
-    if any([not spyder_args, not pid, not is_bootstrap]):
+    if any([not spyder_args, not pid, not is_bootstrap, not reset]):
```
SylvainCorlay: (not a) or (not b) or (not c) == not (a and b and c)

spyderlib/restart_app.py
```diff
@@ -128,19 +219,72 @@ def main():
...
     env = os.environ.copy()
+
+    # Reset Spyder (if required)
+    # -------------------------------------------------------------------------
+    if reset:
+        restarter.set_splash_message(_('Resetting Spyder to defaults'))
+        command_reset = '"{0}" "{1}" {2}'.format(python, spyder, args_reset)
+
+        try:
+            p = subprocess.Popen(command_reset, shell=shell)
```
SylvainCorlay: env=env

spyderlib/restart_app.py
goanpeca: True :)

spyderlib/restart_app.py
goanpeca: That fixed it ?

SylvainCorlay: neveermind.

goanpeca: ehhh... what ?

spyderlib/spyder.py
```diff
@@ -2735,6 +2754,7 @@ def restart(self):
         env['SPYDER_ARGS'] = spyder_args
         env['SPYDER_PID'] = str(pid)
         env['SPYDER_IS_BOOTSTRAP'] = str(is_bootstrap)
+        env['SPYDER_RESET'] = str(reset)
```
SylvainCorlay: adding `env['PYTHON_PATH'] = sys.path` gets rid of the import errors, but it seems the old Spyder is not shutting down though.

goanpeca: So the problem was only with the reset then,, or even if you just try to restart spyder it was failing?

SylvainCorlay: @blink1073 can you give it a shot in windows?

spyderlib/restart_app.py
```diff
@@ -8,8 +8,9 @@
 """
 Restart Spyder
 
-A helper script that allows Spyder to restart from within the application.
-"""
+A helper script that allows to restart (and also reset) Spyder from within the
+running application.
+my """
```
ccordoba12: It seems to be an error here.

spyderlib/spyder.py
goanpeca: So it was working on windows? @blink1073 ?

blink1073: It works with the latest version on Windows.

goanpeca: Great :-), thanks for checking.

spyderlib/restart_app.py
```diff
@@ -126,21 +232,63 @@ def main():
...
+        except Exception as error:
+            restarter.launch_error_message(error_type=RESET_ERROR, error=error)
+        else:
+            p.communicate()
+            pid_reset = p.pid
+
+        # Before launching a new Spyder instance we need to make sure that the
+        # reset subprocess has closed. We wait for a fixed and "reasonable"
+        # amount of time and check, otherwise an error is launched.
+        wait_time = 60  # Seconds
```
ccordoba12: So we're waiting the same amount of time in all OSes? I thought we decided to wait more on Windows (90 sec) and less on Linux and Mac (30 sec).

spyderlib/spyder.py
```diff
@@ -2735,6 +2754,11 @@ def restart(self):
         env['SPYDER_ARGS'] = spyder_args
         env['SPYDER_PID'] = str(pid)
         env['SPYDER_IS_BOOTSTRAP'] = str(is_bootstrap)
+        env['SPYDER_RESET'] = str(reset)
+        if os.name == 'nt':
+            env['PYTHONPATH'] = ';'.join(sys.path)
+        else:
+            env['PYTHONPATH'] = ':'.join(sys.path)
```
ccordoba12: Do we need to the full `sys.path` here? Or just its first entry?

goanpeca: Hmmm I guess the first entry should be enough, is it a problem that we use all anyway?

spyderlib/restart_app.py
goanpeca: This is the reset process, the reset process normally is fast on windows as well. Is on waiting or the spyder instance to close that takes long, [check this](https://github.com/spyder-ide/spyder/pull/2423/files#diff-6614a100b3eb58a6c02b7b1da9e72fa1R240)

blink1073: I'd say we don't know exactly what order they added things to the path.  Worst case we have duplicated entries with the current method.

ccordoba12: So what you say is that we should leave `sys.path` as it is?

blink1073: I'm saying leave the PR as it is.  :smile:

goanpeca: I concur with blinky

spyderlib/spyder.py
ccordoba12: This could cause issues with our consoles. Specially with our Mac app because it does some funky things to PYTHONPATH.

Question: I think we need this only for `bootstrap` to work, right @goanpeca?

goanpeca: Yep, I would say that is correct. If not running from bootstrap then it must be running from an installed version (site-packages) so the imports should be fine then.

I can make add an if so that the path is updated only if running from bootstrap?

ccordoba12: Even easier: we could `DEV` here like this:

```python
if DEV
    if os.name == 'nt':
        env['PYTHONPATH'] = ';'.join(sys.path)
    else:
        env['PYTHONPATH'] = ':'.join(sys.path)
```

goanpeca: Yes, that is what I meant...

ccordoba12: Ok, nice!

spyderlib/restart_app.py
```diff
@@ -126,21 +232,63 @@ def main():
...
     # Adjust the command and/or arguments to subprocess depending on the OS
-    shell = os.name != 'nt'
+    shell = not IS_WINDOWS
 
-    # Wait for original process to end before launching the new instance
-    while True:
+    # Before launching a new Spyder instance we need to make sure that the
+    # previous one has closed. We wait for a fixed and "reasonable" amount of
+    # time and check, otherwise an error is launched
+    wait_time = 60*5 if IS_WINDOWS else 60*3 # Seconds
```
ccordoba12: So we discussed this with @blink1073 and @SylvainCorlay and all of us agree that these waiting times are way too big. We think nobody is going to wait 3 min, much less 5 min, for Spyder to start :-)

So our proposal for this is: 30 sec for Linux/Mac (perhaps even 20), and 75 sec (90 tops) for Windows.

goanpeca: Ok Done...

30 and 90

goanpeca:
Any my mental health?

goanpeca:
Anything holding this @ccordoba12?

spyderlib/restart_app.py
```diff
@@ -84,34 +96,128 @@ def to_text_string(obj, encoding=None):
...
+                    RESET_ERROR: _("Spyder could not reset to factory "
+                                   "defaults.\nRestart aborted."),
+                    RESTART_ERROR: _("It was not possible to restart Spyder.\n"
+                                     "Operation aborted.")}
+        titles = {CLOSE_ERROR: _("Spyder exit error"),
+                  RESET_ERROR: _("Spyder reset error"),
+                  RESTART_ERROR: _("Spyder restart error")}
+
+        if error:
+            e = error.__repr__()
```
ccordoba12: Why is this necessary?

ccordoba12:
I think waiting 60 secs for `--reset` to work is too much because that would be added to the restart time, right?

Also, I'd like to discuss how the logic for reset + restart works. So `restart_app` calls `spyder --reset` first and then it closes the app, or is the other way around?

goanpeca:
Ok. When you restart first we wait for spyder to close. If that happenned before the max wait time then we launch the reset process, if that went fine and took less than the stated time then the new spyd instance is launched.

ccordoba12:
So this means a user could have to wait up to (30 or 90) + 60 secs for a reset (plus the time it take for a new instance to appear).

Can we reduce the time reset to 20 secs or so? It usually doesn't take that long ;-)

goanpeca:
Yes, you are correct.

Sure, 20 secs is fine (I guess... we will know otherwise!)

goanpeca:
Anything besides that ?

goanpeca:
Anything else @ccordoba12 ?

ccordoba12:
Ok, merging this with @goanpeca (in person) for his own mental health :-P

blink1073:
woot!

PR #2426, opened 2015-05-07
Add option to toggle toolbars visibility

goanpeca:
Related to #2353 

## Description
Adds a toggle toolbars visibility button on view menu. Visibility on (action checked) visibility off (action unchecekd).

The last visible toolbars are updated only when using the toggle action. 

image

## Todo
- [x] Saves last visible toolbars when using the toggle button
- [x] Update config
- [x] Add a shortcut (`Ctrl+Shift+T`)
- [x] Add docstrings

### DIFF
--- a/spyderlib/config.py
+++ b/spyderlib/config.py
@@ -197,7 +197,6 @@ DEFAULTS = [
               'custom_margin': 0,
               'show_internal_console_if_traceback': True,
               'check_updates_on_startup': True,
-              'last_visible_toolbars': [],
               'toolbars_visible': True,
               }),
             ('quick_layouts',
@@ -486,6 +485,7 @@ DEFAULTS = [
               '_/save current layout': "Shift+Alt+S",
               '_/toggle default layout': "Shift+Alt+Home",
               '_/layout preferences': "Shift+Alt+P",
+              '_/show toolbars': "Ctrl+Shift+T",
               '_/restart': "Shift+Alt+R",
               '_/quit': "Ctrl+Q",
               # -- In plugins/editor
--- a/spyderlib/config.py
+++ b/spyderlib/config.py
@@ -196,7 +196,9 @@ DEFAULTS = [
               'use_custom_margin': True,
               'custom_margin': 0,
               'show_internal_console_if_traceback': True,
-              'check_updates_on_startup': True
+              'check_updates_on_startup': True,
+              'last_visible_toolbars': [],
+              'toolbars_visible': True,
               }),
             ('quick_layouts',
              {
--- a/spyderlib/config.py
+++ b/spyderlib/config.py
@@ -485,7 +485,7 @@ DEFAULTS = [
               '_/save current layout': "Shift+Alt+S",
               '_/toggle default layout': "Shift+Alt+Home",
               '_/layout preferences': "Shift+Alt+P",
-              '_/show toolbars': "Ctrl+Shift+T",
+              '_/show toolbars': "Alt+Shift+T",
               '_/restart': "Shift+Alt+R",
               '_/quit': "Ctrl+Q",
               # -- In plugins/editor
--- a/spyderlib/spyder.py
+++ b/spyderlib/spyder.py
@@ -1922,30 +1922,27 @@ class MainWindow(QMainWindow):
             action = plugin.toggle_view_action
             action.setChecked(plugin.dockwidget.isVisible())
 
-    # --- Other
+    # --- Show/Hide toolbars
     def _update_show_toolbars_action(self):
-        """ """
+        """Update the text displayed in the menu entry."""
         if self.toolbars_visible:
             text = _("Hide toolbars")
             tip = _("Hide toolbars")
-#            icon = ima.icon('maximize')
         else:
             text = _("Show toolbars")
             tip = _("Show toolbars")
-#            icon = ima.icon('unmaximize')
         self.show_toolbars_action.setText(text)
         self.show_toolbars_action.setToolTip(tip)
-#        self.show_toolbars_actionsetIcon(icon)
 
     def save_visible_toolbars(self):
-        """ """
+        """Saves the name of the visible toolbars in the .ini file."""
         toolbars = []
         for toolbar in self.visible_toolbars:
             toolbars.append(toolbar.objectName())
         CONF.set('main', 'last_visible_toolbars', toolbars)
 
     def get_visible_toolbars(self):
-        """ """
+        """Collects the visible toolbars."""
         toolbars = []
         for toolbar in self.toolbarslist:
             if toolbar.toggleViewAction().isChecked():
@@ -1953,14 +1950,14 @@ class MainWindow(QMainWindow):
         self.visible_toolbars = toolbars
 
     def load_last_visible_toolbars(self):
-        """ """
+        """Loads the last visible toolbars from the .ini file."""
         toolbars_names = CONF.get('main', 'last_visible_toolbars', default=[])
 
         if toolbars_names:
             dic = {}
             for toolbar in self.toolbarslist:
                 dic[toolbar.objectName()] = toolbar
-    
+
             toolbars = []
             for name in toolbars_names:
                 if name in dic:
@@ -1971,7 +1968,7 @@ class MainWindow(QMainWindow):
         self._update_show_toolbars_action()
 
     def show_toolbars(self):
-        """ """
+        """Show/Hides toolbars."""
         value = not self.toolbars_visible
         CONF.set('main', 'toolbars_visible', value)
         if value:
@@ -1986,6 +1983,7 @@ class MainWindow(QMainWindow):
         self.toolbars_visible = value
         self._update_show_toolbars_action()
 
+    # --- Other
     def plugin_focus_changed(self):
         """Focus has changed from one plugin to another"""
         if self.light:
--- a/spyderlib/spyder.py
+++ b/spyderlib/spyder.py
@@ -350,7 +350,7 @@ class MainWindow(QMainWindow):
 
         # Actions
         self.lock_dockwidgets_action = None
-        self.toggle_toolbars_action = None
+        self.show_toolbars_action = None
         self.close_dockwidget_action = None
         self.find_action = None
         self.find_next_action = None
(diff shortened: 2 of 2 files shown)


goanpeca:
@ccordoba12, @Nodd  what would the expected behavior be after a shutdown?

I think it makes sense to "toggle" as in, if 5 toolbars were on then the toggle will make those (and only those 5) visible/invisible.

My question is... what should be done on startup if all toolbars are invisible? should I keep track of the last set of toolbars that was visible before the disable? or should we just grey out the toggle visibility option if all toolbars are off after a start?

goanpeca:
Comments @spyder-ide/core-developers ? 

>I think it makes sense to "toggle" as in, if 5 toolbars were on then the toggle will make those (and only those 5) visible/invisible.

>My question is... what should be done on startup if all toolbars are invisible? should I keep track of the last set of toolbars that was visible before the disable? or should we just grey out the toggle visibility option if all toolbars are off after a start?

blink1073:
Yeah we could keep track of the current/previously visible toolbars, and toggle would bring back the previous known state, that seems intuitive.

Nodd:
:+1:

spyderlib/config.py
```diff
@@ -192,7 +192,9 @@ def is_ubuntu():
               'use_custom_margin': True,
               'custom_margin': 0,
               'show_internal_console_if_traceback': True,
-              'check_updates_on_startup': True
+              'check_updates_on_startup': True,
+              'last_visible_toolbars': [],
```
ccordoba12: This options is transient, so there's no need to add to it here.

spyderlib/spyder.py
```diff
@@ -1144,7 +1146,11 @@ def add_xydoc(text, pathlist):
                                          self.close_dockwidget_action,
                                          self.maximize_action,
                                          None))
+            self.toggle_toolbars_action = create_action(self,
+                                    _("Toggle toolbar visibility"),
```
ccordoba12: Suggestion here: `Show toolbars`. I think it's easier to understand to what it refers to ;-)

spyderlib/spyder.py
```diff
@@ -1144,7 +1146,11 @@ def add_xydoc(text, pathlist):
                                          self.close_dockwidget_action,
                                          self.maximize_action,
                                          None))
+            self.toggle_toolbars_action = create_action(self,
+                                    _("Toggle toolbar visibility"),
+                                    toggled=self.toggle_toolbars)
```
ccordoba12: This method (`self.toggle_toolbars`) and the corresponding action should be renamed if you decide to follow my suggestion.

spyderlib/spyder.py
```diff
@@ -1928,6 +1940,56 @@ def quick_layout_switch(self, index):
             action = plugin.toggle_view_action
             action.setChecked(plugin.dockwidget.isVisible())
 
+    # --- Other
+    def save_visible_toolbars(self):
+        """ """
+        print('save')
```
ccordoba12: Please remove this `print`

spyderlib/spyder.py
```diff
@@ -1928,6 +1940,56 @@ def quick_layout_switch(self, index):
             action = plugin.toggle_view_action
             action.setChecked(plugin.dockwidget.isVisible())
 
+    # --- Other
+    def save_visible_toolbars(self):
+        """ """
+        print('save')
+        temp = []
```
ccordoba12: This variable should be called `toolbars`

spyderlib/spyder.py
```diff
@@ -1928,6 +1940,56 @@ def quick_layout_switch(self, index):
...
+        """ """
+        print('save')
+        temp = []
+        for toolbar in self.visible_toolbars:
+            temp.append(toolbar.objectName())
+        CONF.set('main', 'last_visible_toolbars', temp)
+
+    def load_last_visible_toolbars(self):
+        """ """
+        print('load')
```
ccordoba12: Remove this `print`

spyderlib/spyder.py
```diff
@@ -1928,6 +1940,56 @@ def quick_layout_switch(self, index):
...
+        toolbars_names = CONF.get('main', 'last_visible_toolbars')
```
ccordoba12: You need to use here the `default` option of `CONF.get`, like this

```python
CONF.get('main', 'last_visible_toolbars', default=[])
```

given that it's not going to be part of `config.py`

spyderlib/spyder.py
```diff
@@ -1928,6 +1940,56 @@ def quick_layout_switch(self, index):
...
+        print(toolbars_names, type(toolbars_names))
```
ccordoba12: Remove `print`

spyderlib/spyder.py
```diff
@@ -1928,6 +1940,56 @@ def quick_layout_switch(self, index):
...
+    
+            self.visible_toolbars = []            
+            toolbars_names = CONF.get('main', 'last_visible_toolbars')
+            for name in toolbars_names:
+                if name in dic:
+                    self.visible_toolbars.append(dic[name])
+        print(self.visible_toolbars)
+
+    def get_visible_toolbars(self):
+        self.visible_toolbars = []
```
ccordoba12: I think it's better to use a temporary variable here, instead of assigning to the attribute directly, and at the end of this method assign `self.visible_toolbars` the value of that temp var.

spyderlib/spyder.py
```diff
@@ -1928,6 +1940,56 @@ def quick_layout_switch(self, index):
...
+        print(toolbars_names, type(toolbars_names))
+        if not toolbars_names:
+            self.get_visible_toolbars()
+        else:
+            dic = {}
+            for toolbar in self.toolbarslist:
+                dic[toolbar.objectName()] = toolbar
+    
+            self.visible_toolbars = []            
+            toolbars_names = CONF.get('main', 'last_visible_toolbars')
```
ccordoba12: You already get the value 'last_visible_toolbars' above, so why the need of this line here?

spyderlib/spyder.py
```diff
@@ -1928,6 +1940,56 @@ def quick_layout_switch(self, index):
...
+            for name in toolbars_names:
+                if name in dic:
+                    self.visible_toolbars.append(dic[name])
+        print(self.visible_toolbars)
```
ccordoba12: Remove `print`

spyderlib/spyder.py
```diff
@@ -1928,6 +1940,56 @@ def quick_layout_switch(self, index):
...
+
+    def get_visible_toolbars(self):
+        self.visible_toolbars = []
+        for toolbar in self.toolbarslist:
+            if toolbar.toggleViewAction().isChecked():
+                self.visible_toolbars.append(toolbar)
+        print(self.visible_toolbars)
```
ccordoba12: Remove `print`

spyderlib/spyder.py
```diff
@@ -1928,6 +1940,56 @@ def quick_layout_switch(self, index):
...
+
+    def toggle_toolbars(self, value):
+        """ """
+        print('toggled', value)
+        print(self.visible_toolbars)
```
ccordoba12: Remove `print`'s

spyderlib/spyder.py
```diff
@@ -1928,6 +1940,56 @@ def quick_layout_switch(self, index):
...
+        CONF.set('main', 'toolbars_visible', value)
+        if not value:
+            pass
+            self.get_visible_toolbars()
```
ccordoba12: Is this line executed by being after the `pass`? I think it's not :-)

spyderlib/spyder.py
```diff
@@ -1928,6 +1940,56 @@ def quick_layout_switch(self, index):
...
+        toolbars_names = CONF.get('main', 'last_visible_toolbars')
+        print(toolbars_names, type(toolbars_names))
+        if not toolbars_names:
+            self.get_visible_toolbars()
+        else:
+            dic = {}
+            for toolbar in self.toolbarslist:
+                dic[toolbar.objectName()] = toolbar
+    
+            self.visible_toolbars = []
```
ccordoba12: I think you should also use a temp var here

spyderlib/spyder.py
```diff
@@ -1268,6 +1274,11 @@ def post_visible_setup(self):
         self.extconsole.setMinimumHeight(0)
 
         if not self.light:
+            # Update toolbar visibility status
+            self.load_last_visible_toolbars()
+            toolbars_visible = CONF.get('main', 'toolbars_visible')
```
ccordoba12: This should be called `show_toolbar_o`. It's a Spyder custom to call option variables that way :-)

ccordoba12: At this point toolbars have been created or not?

ccordoba12:
Ok, first review done.

spyderlib/spyder.py
Nodd: I think it is, `pass` is just a no-op unlike `break` or `return`. Anyway `pass` should be removed.

goanpeca: Yes, sorry this PR was half baked... will try to fix it today

spyderlib/config.py
goanpeca: Yes, I keep forgetting that!

spyderlib/spyder.py
goanpeca: I know it is easier, it was my first option, but then it is weird when having it in a menu with a check box, unless I rename the option when toggling it from 

`Show toolbars` -> `Hide toolbars`

I actually like this option cause it quite explicit always, much like the maximize pane button is updated depending on its status.

spyderlib/spyder.py
goanpeca: Check my previous comment, I would rather have something like `show_hide_toolbars` and change the entry in the menu from `Show toolbars` -> `Hide toolbars` depending on the state.

goanpeca:
I have to add some docstrings still but I fixed all of your comments and now instead of toggling, it uses a trigger (like `maximize pane`) so that the text in the option changes between `Show toolbars` and `Hide toolbars`

goanpeca:
@ccordoba12 updated docstrings!

spyderlib/config.py
```diff
@@ -484,6 +485,7 @@ def is_ubuntu():
               '_/save current layout': "Shift+Alt+S",
               '_/toggle default layout': "Shift+Alt+Home",
               '_/layout preferences': "Shift+Alt+P",
+              '_/show toolbars': "Alt+Shift+T",
```
ccordoba12: @goanpeca, can you confirm that this working for you? I remember you told there was some problem with this change.

goanpeca: I checked, it works

Issue #2394, opened 2015-05-03
Failed to launch when multiple version python installed and not using the default python


cheneave:
Well, it's a very minor issue.
I have installed both python2 and python3, and I'm using python2 as the default python.
The bat file to launch spyder is

    python "%~dpn0" %*

It will try to use python2 to run the spyder.

ccordoba12:
I think there's no `python3` executable on Windows :-)

cheneave:
Is the bat created by setup.py ? I think we can write distutils.sysconfig.PREFIX to the bat when creating it.

ccordoba12:
Please take a look at it and send us a pull request when you find how to improve things :-)

PR #2431, opened 2015-05-09
PR: Use right python.exe to run Spyder when multiple versions of Python are installled on Windows

cheneave:
Use right python.exe to run spyder when multiple versions of python installed on windows

### DIFF
--- a/scripts/spyder.bat
+++ b/scripts/spyder.bat
@@ -1,2 +1,6 @@
 @echo off
-python "%~dpn0" %*
+IF EXIST "%~dpn0\..\..\python.exe" (
+	"%~dpn0\..\..\python.exe" "%~dpn0" %*
+) ELSE (
+	python.exe "%~dpn0" %*
+)


ccordoba12:
I have a couple of things to say:

1. Could you explain a bit further what your change is about?
2. Is not this pull request supposed to fix a bug you already reported? If that's the case, please report the bug number here too.
3. Using a relative path can break projects like WinPython. @stonebig, would you mind to have a look at this pull request?

cheneave:
1.
(1) I tried to use relative path to find the pythonw.exe. If it exists, use it to run, if not , use default pythonw.exe directly. 
(2)Use pythonw.exe instead of python to git rid of the command line window. 
(3) Use "start" command so the bat exits right after the spyder is started.
2. Yes, it's used to fix issue #2394
3. For projects like WinPython, the pythonw.exe won't exist, so it world work just like before.

ccordoba12:
I see, so you want to change `python` for `pythonw`. This will probably invalidate our command line option called `--show-console`.

Could you verify if that's the case or not by running in a terminal

    spyder --show-console

with this branch?

cheneave:
I found no difference weather use --show-console or not. When use python.exe, there's a command line window showing nothing. And when use pythonw.exe, there's no such window.
In all four cases(python or pythonw, with --show-console or not), the internal console  is not shown by default and can be enabled manually.
It seems --show-console option effects nothing.

stonebig:
Hi @ccordoba12 ,

Checking the change:
- it doesn't break current Winpython `Spyder` launcher icon,
- it seems to do the expected improvement when clicking on the spyder.bat (instead of the icon).

I'm ok from the `Winpython` point of view.

ccordoba12:
@cheneave, `--show-console` doesn't have anything to do with the internal console. It has do with **no** closing the console Spyder was started with (i.e. `cmd.exe`).

This is very useful for users to report errors that are only written to `cmd.exe`. Unfortunately, your pull request seems to break that (according to what you reported above :-)

cheneave:
Do you mean something woud be outputed by print() function if the "--show-console" is used. And if "--show-console" is not used, there won't be message printed through the command window is shown?

ccordoba12:
Not by `print` precisely. Usually it's a traceback that can't be printed to the internal console because it's not yet created.

If `--show-console` is not used, then we have code that closes `cmd.exe` automatically, so no tracebacks can be printed there.

cheneave:
Well, I have found the reason. It's because I'm using the conemu as default terminal and the cmd.exe is not closed automatically at this condition.
Using python.exe instead of pythonw.exe works fine.

cheneave:
@ccordoba12 I have modified the bat so that only change 1 change is done. 
(1) It tried to use relative path to find the python.exe. If it exists, use it to run, if not , use default python.exe directly.

PR #2434, opened 2015-05-10
Match for pylint when parsing pylint output

pijyoi:
Pylint program is invoked using PYLINT, which can now be pylint, pylint3, python3-pylint. Fix the regex to just search for PYLINT.

### DIFF
--- a/spyderplugins/widgets/pylintgui.py
+++ b/spyderplugins/widgets/pylintgui.py
@@ -61,7 +61,8 @@ def get_pylint_version():
                                shell=True if os.name == 'nt' else False)
     lines = to_unicode_from_fs(process.stdout.read()).splitlines()
     if lines:
-        match = re.match('(pylint3?|pylint-script.py) ([0-9\.]*)', lines[0])
+        regex = '({0}|pylint-script.py) ([0-9\.]*)'.format(PYLINT)
+        match = re.match(regex, lines[0])
         if match is not None:
             return match.groups()[1]
 


spyderplugins/widgets/pylintgui.py
```diff
@@ -61,7 +61,8 @@ def get_pylint_version():
                                shell=True if os.name == 'nt' else False)
     lines = to_unicode_from_fs(process.stdout.read()).splitlines()
     if lines:
-        match = re.match('(pylint3?|pylint-script.py) ([0-9\.]*)', lines[0])
+        regex = '({0}|pylint-script.py) ([0-9\.]*)'.format(PYLINT)
```
ccordoba12: I think a `?` is missing here after `{0}`, at least according to the changed line above :-)

pijyoi: I understood "pylint3?" to mean to match either "pylint" or "pylint3".
PYLINT equals "pylint" or "pylint3" or "python3-pylint" depending on the
binary that was already previously found.

Was the previous behavior to allow the version string to be different from
the binary name?
- pylint but "pylint3 1.3.1"
- pylint3 but "pylint 1.3.1"
Does this happen in practice?

I originally used "(python3-pylint|pylint3?|pylint-script.py)" but thought
that it was a burden to have to modify the regex each time a new binary
name cropped up.

Or simpler, why not just skip the first word? The code already assumes the
version string is on the first line.

ccordoba12: I think `?` is necesary to parse the output of pylint itself. I'll check that in master and I'll add it if needed.

ccordoba12: Sorry, this was only necessary to detect the Pylint version, so I think your fix is the right one :-)

PR #2473, opened 2015-06-08
PR: Boottime

techtonik:
This adds reporting initialization/bootstrap time for Spyder
Looks into bootstrap.py output and into internal console.

On my machine 'master' takes 15 seconds to start up. =/


spyderlib/start_app.py
```diff
@@ -46,10 +46,13 @@ def send_args_to_spyder(args):
         break
 
 
-def main():
+def main(report_time=None):
```
Nodd: `report_time` is a flag, so `False` as a default value makes more sense.

spyderlib/start_app.py
```diff
@@ -95,6 +98,8 @@ def main():
             if TEST is None:
                 atexit.register(lock.unlock)
             from spyderlib import spyder
+            if report_time:
+                spyder.Stats.start_time = report_time
```
Nodd: I don't understand what this line does ? When does is print the time ?

spyderlib/utils/misc.py
```diff
@@ -11,6 +11,21 @@
...
+import time
+
+def timestamp(start, end):
+    """Get timestamp like 00:00:02.4950 from start and end times that
+       are collected with time.time()
+    """
+    time_lapse = end - start
+    return (time.strftime("%H:%M:%S.", time.gmtime(time_lapse)) +
+        # gmtime() converts float into tuple, but loses milliseconds
+        ("%.4f" % time_lapse).split('.')[1])
```
Nodd: Why don't you pass `time_lapse` directly without the call to `gmtime` ?

spyderlib/start_app.py
techtonik: It just saves the init start time. I couldn't do this from bootstrap, because `spyder` module is not imported there.

spyderlib/utils/misc.py
techtonik: Good point. I need to search the history of bootstrap.py - it is from here.

techtonik: Because [time.gmtime()](https://docs.python.org/2/library/time.html#time.gmtime) doesn't accept `time.time()` output (float), as an argument and requires [struct_time](https://docs.python.org/2/library/time.html#time.struct_time)

Nodd: The function in your last comment should be `time.strftime`, but I see it now, thanks.

spyderlib/start_app.py
Nodd: I missed the `3,147 additions, 3,135 deletions not shown ` in `spyderlib/spyder.py` that are not expanded (that's a lot !). I guess that the value is used here.

Nodd:
I get it now, I had to look at each commit separately. 2 questions:
* Couldn't you keep line feeds as they where ?
* You now there are great profiling tools that give even more details on execution time ? Some in the std lib, some as third-party modules...

techtonik:
CRLF: Yes. I have stashed the patch from my linefeed correction branch and couldn't pop it, because it ended with merge conflicts. I couldn't figure out yet what to do with my existing code to avoid this conflict (mostly due to time constraints). So, [ ] it may become a start for instructions how to correctly rebase your changes after CRLF fixes are merged.

BTW, if you could check that this branch merges with my patch-1 branch, that will show if same CRLF changes in parallel branches conflict or not. I am unlikely to work on this right now. Maybe in few hours. It will save me some time if that check will be done independently.

As for profiling, I want startup info to be in bootstrap.py console output so that everybody always had a clear picture of Spyder startup times and could investigate when it changes. I want this to be backported to 2.3, because it **seems** faster, but I don't have data. There is also a problem that I still couldn't find a way to print to bootstrap console. Probably that was changed after 2.3

Nodd:
If you want data, use a profiler ! :smile:

techtonik:
Not many people know how to do this. It should be automatic.

Nodd:
How many people *need* to do this ? Among them, how many of them don't know how to use a profiler ? And among these, how many are too lazy to google it ?
I'm not against adding a small timer to have an idea of spyder's startup time, but to be able to work on speeding up the program you need a real profiler.
Maybe you can run `bootstrap.py` from spyder itself using the profiler plugin ? :smile:

techtonik:
@Nodd I don't have data on that. I know that me doesn't profile Spyder after every change (and all other devs too), and I want this data to start collecting in Travis logs for quick comparison. For actual speed up work I would of course use something like PyVmMonitor.

ccordoba12:
This needs a rebase or merge with master.

techtonik:
@ccordoba12 have you solved the branch linefeed problem already?

ccordoba12:
Nop, nobody has been bothered by that except you :-)

ccordoba12:
@techtonik, do you want to rebase this PR to fix conflicts?

ccordoba12:
Ah, you said that you needed line endings to be fixed. Since that's not going to happen anytime soon, I'm closing this pull request.

@techtonik, please reopen it if you solve merge conflicts and want this to be merged.

PR #2484, opened 2015-06-13
Remove unnecessary changes to detected kernel json file when connecting to external kernels

### LINKED ISSUE #2468: 'Connect to existing kernel' fails if json file not in PWD
I'm using Spyder 2.3.4 on OS X, trying to connect to an existing (remote) IPython kernel.  I successfully connected to the kernel using

    ipython qtconsole --existing /path/to/json/file/kernel-4957.json

but get an error **Unable to connect to IPython kernel-4957.json** when attempting to connect via Spyder.

I was eventually able to connect by moving the kernel-4957.json file to my home directory (which was also my PWD).  The requirement that the json file is in the PWD is totally unexpected, since the "Connect to an existing terminal" dialog box allows me to navigate to find a json file in any directory on my computer.

SylvainCorlay:
Fixes #2468

----

It is unnecessary to prefix the connection file with `kernel-` and add the `.json` extension. IPython's [find_connection_file](https://github.com/jupyter/jupyter_client/blob/1c77df8fe707171504c4a761093041c7ff130c86/jupyter_client/connect.py#L136) already does that for you.

Besides, by doing it forcibly, we prevent the use of kernel files named otherwise than kernel-*.json, which occurs quite often  when connecting to a remote kernel.

### DIFF
--- a/spyderlib/plugins/ipythonconsole.py
+++ b/spyderlib/plugins/ipythonconsole.py
@@ -1058,8 +1058,6 @@ class IPythonConsole(SpyderPluginWidget):
         # Verifying if the connection file exists
         cf = osp.basename(cf)
         try:
-            if not cf.startswith('kernel') and not cf.endswith('json'):
-                cf = to_text_string('kernel-' + cf + '.json')
             find_connection_file(cf, profile='default')
         except (IOError, UnboundLocalError):
             QMessageBox.critical(self, _('IPython'),


spyderlib/plugins/ipythonconsole.py
```diff
@@ -1058,8 +1058,6 @@ def _create_client_for_kernel(self, cf, hostname, kf, pw):
         # Verifying if the connection file exists
         cf = osp.basename(cf)
         try:
-            if not cf.startswith('kernel') and not cf.endswith('json'):
-                cf = to_text_string('kernel-' + cf + '.json')
```
ccordoba12: The idea is to pass the kernel id (like just `4779`) to this field and be able to connect to that kernel, instead of writing `kernel-4779.json`.

That's why these lines are here and I find this feature really useful :-)

SylvainCorlay: This feature is there without these lines.

SylvainCorlay: find_connection_file already does precisely what you are describing.

ccordoba12: Since what IPython version?

SylvainCorlay: It has been the case for a while AFAIK. Here is the function [find_conneciton_file](https://github.com/ipython/ipython/blob/0.13.x/IPython/lib/kernel.py#L56) in 0.13.2.

*If the argument does not match an existing file, it will be interpreted as a fileglob, and the matching file in the profile's security dir with the latest access time will be used.*

ccordoba12: Thanks Sylvain. I wasn't aware of that ;-)

Issue #2175, opened 2015-02-17
Editor - Show blank spaces suggestions


spyder-issues-migrator:
_From [goanp...@gmail.com](https://code.google.com/u/114799823853069473860/) on 2015-02-14T12:29:43Z_

Spyder Version:  2.4.0dev (4931: r1ff46f376d07 )
Python Version:  2.7.9
Qt Version    :  4.8.5, PySide 1.2.1 on Linux
pyflakes >=0.5.0:  0.5.0 (OK)
pep8 >=0.6      :  1.4.6 (OK)
IPython >=1.0   :  2.3.1 (OK)
zmq >=2.1.11    :  14.4.1 (OK)
pygments >=1.6  :  2.0.1 (OK)
pandas >=0.13.1 :  0.15.2 (OK)
sphinx >=0.6.6  :  1.2.3 (OK)
psutil >=0.3    :  2.1.1 (OK)
jedi >=0.8.1    :  None (NOK)
rope >=0.9.2    :  0.10.2 (OK)
matplotlib >=1.0:  1.4.2 (OK)
sympy >=0.7.0   :  0.7.6 (OK)
pylint >=0.25   :  None (NOK) 

What steps will reproduce the problem?

 1. Activating the 'Show blank spaces' in the Source menu 

What is the expected output? What do you see instead?

 The function works but I have some suggestions/comments...

Please provide any additional information below

.
I would suggest:
1.) The dots are the same color as main text, they should be lighter, to provide guidance without being too intrusive
2.) It would be nice to sync the indentation characters selected in the Editor Advanced preferences to display the same, so if 4 spaces is selected, every 4 spaces, the dot should become a bar `|` or should be a bit different to show this clearly
3.) I think having dots for the indentation before any content in the line is more than enough. The idea of having this is to guide with indentation, but right now every single space is "highlighted", which can be distracting in the middle of definition of variables and statements of code

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=2175_

spyder-issues-migrator:
_From [ccordoba12](https://code.google.com/u/ccordoba12/) on 2015-02-15T15:18:53Z_

Good suggestions. About them:

1. No idea how that could be done. I just activated a Qt option to show blank spaces in QPlainTextEdit widgets. So I don't know how dots color is selected.

2. Vertical bars can be activated too, but that's not related to blank spaces, that's a different issue. So please open a new one for that.

[...]

Issue #1805, opened 2015-02-17
Highlight trailing space


spyder-issues-migrator:
_From [vex...@gmail.com](https://code.google.com/u/115177635129669581971/) on 2014-05-01T05:12:24Z_

It would be neat if spyder could highlight trailing space.

I know that there is an option to remove all trailing space upon save, but that is not always desirable, especially in cases where formatting changes and actual changes should be separate commits. I would rather remove trailing whitespace manually, but I can't see it at the moment.

Cheers

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=1805_

spyder-issues-migrator:
_From [ccordoba12](https://code.google.com/u/ccordoba12/) on 2015-01-01T15:08:34Z_

**Status:** HelpNeeded  
  **Labels:** Cat-Editor

goanpeca:
Well why not activate pep8 then?

ccordoba12:
This already solved. You can view blank spaces now by going to the menu

    Source > Show blank spaces

Nodd:
Showing blank space is not the same as showing trailing blank spaces only. The first one displays little dots everywhere while the second one should display only the one to be removed (and it's harder to do).
Maybe it's easier to highlight them (a simple regexp in the syntax highlighter should do it) with a red background rather than showing a dot ?

ccordoba12:
Ok, you're right. Reopening then ;-)

brhubbar:
I believe this is accomplished by `Completion and Linting > Linting > Underline errors and warnings`

PR #2501, opened 2015-06-23
Make blank space less apparent.

hugobuddel:
Less apparent blank space, improved readability, fixing #2175.
Foreground color of blank space (if shown) is blended with the background color, therefore being less strongly present in both light and dark schemes.

### DIFF
--- a/spyderlib/widgets/sourcecode/syntaxhighlighters.py
+++ b/spyderlib/widgets/sourcecode/syntaxhighlighters.py
@@ -192,7 +192,7 @@ class BaseSH(QSyntaxHighlighter):
             self.color_scheme = color_scheme
         self.setup_formats()
         self.rehighlight()
-    
+
     def highlightBlock(self, text):
         raise NotImplementedError
 
--- a/spyderlib/widgets/sourcecode/syntaxhighlighters.py
+++ b/spyderlib/widgets/sourcecode/syntaxhighlighters.py
@@ -414,6 +414,29 @@ class PythonSH(BaseSH):
                     
             match = self.PROG.search(text, match.end())
  
+        # Make blank space less apparent by setting the foreground
+        # color closer to the background color.
+        # This only has an effect when "Show blank space" is turned on.
+        for space_character in [" ", "\t"]:
+            pos = text.find(space_character, 0)
+            while pos >= 0:
+                posoffset = pos + offset
+                if posoffset >= 0:
+                    format = self.format(posoffset)
+                    color_foreground = format.foreground().color()
+                    color_background = format.background().color()
+                    # Blend the colors.
+                    fr, fg, fb = 0.3, 0.3, 0.3
+                    color_foreground_new = QColor(
+                        fr*color_foreground.red() + (1-fr)*color_background.red(),
+                        fg*color_foreground.green() + (1-fg)*color_background.green(),
+                        fb*color_foreground.blue() + (1-fb)*color_background.blue(),
+                        255,
+                    )
+                    self.setFormat(posoffset, 1, color_foreground_new)
+                
+                pos = text.find(" ", pos+1)
+        
         self.setCurrentBlockState(state)
         
         if oedata is not None:
--- a/spyderlib/widgets/sourcecode/syntaxhighlighters.py
+++ b/spyderlib/widgets/sourcecode/syntaxhighlighters.py
@@ -212,8 +212,10 @@ class BaseSH(QSyntaxHighlighter):
                 start, end = match.span()
                 start = max([0, start+offset])
                 end = max([0, end+offset])
+                # Format trailing spaces at the end of the line. 
                 if end == len(text) and format_trailing is not None:
                     self.setFormat(start, end, format_trailing)
+                # Format leading spaces, e.g. indentation.
                 if start == 0 and format_leading is not None:
                     self.setFormat(start, end, format_leading)
                 format = self.format(start)
@@ -449,10 +451,9 @@ class PythonSH(BaseSH):
         
         self.setCurrentBlockState(state)
         
-        # Use normal format for indentation.
+        # Use normal format for indentation and trailing spaces.
         self.formats['leading'] = self.formats['normal']
-        # Use number format for trailing spaces.
-        self.formats['trailing'] = self.formats['number']
+        self.formats['trailing'] = self.formats['normal']
         self.highlight_spaces(text, offset)
         
         if oedata is not None:
--- a/spyderlib/widgets/sourcecode/syntaxhighlighters.py
+++ b/spyderlib/widgets/sourcecode/syntaxhighlighters.py
@@ -84,8 +84,12 @@ class BaseSH(QSyntaxHighlighter):
     """Base Syntax Highlighter Class"""
     # Syntax highlighting rules:
     PROG = None
+    BLANKPROG = re.compile("\s+")
     # Syntax highlighting states (from one text block to another):
     NORMAL = 0
+    # Syntax highlighting parameters.
+    BLANK_ALPHA_FACTOR = 0.31
+    
     def __init__(self, parent, font=None, color_scheme='Spyder'):
         QSyntaxHighlighter.__init__(self, parent)
         
@@ -187,10 +191,28 @@ class BaseSH(QSyntaxHighlighter):
             self.color_scheme = color_scheme
         self.setup_formats()
         self.rehighlight()
-
+    
     def highlightBlock(self, text):
         raise NotImplementedError
-            
+
+    def highlight_spaces(self, text, offset=0):
+        """
+        Make blank space less apparent by setting the foreground alpha.
+        This only has an effect when 'Show blank space' is turned on.
+        Derived classes could call this function at the end of
+        highlightBlock().
+        """
+        match = self.BLANKPROG.search(text, offset)
+        while match:
+            start, end = match.span()
+            start = max([0, start+offset])
+            end = max([0, end+offset])
+            format = self.format(start)
+            color_foreground = format.foreground().color()
+            color_foreground.setAlphaF(self.BLANK_ALPHA_FACTOR * color_foreground.alphaF())
+            self.setFormat(start, end-start, color_foreground)
+            match = self.BLANKPROG.search(text, match.end())
+    
     def get_outlineexplorer_data(self):
         return self.outlineexplorer_data
 
@@ -202,9 +224,9 @@ class BaseSH(QSyntaxHighlighter):
 
 
 class TextSH(BaseSH):
-    """Simple Text Syntax Highlighter Class (do nothing)"""
+    """Simple Text Syntax Highlighter Class (only highlight spaces)"""
     def highlightBlock(self, text):
-        pass
+        self.highlight_spaces(text)
 
 
 class GenericSH(BaseSH):
@@ -225,6 +247,8 @@ class GenericSH(BaseSH):
                     self.setFormat(start, end-start, self.formats[key])
                     
             match = self.PROG.search(text, match.end())
+        
+        self.highlight_spaces(text)
 
 
 #==============================================================================
@@ -413,20 +437,8 @@ class PythonSH(BaseSH):
                                                    self.formats["keyword"])
                     
             match = self.PROG.search(text, match.end())
- 
-        # Make blank space less apparent by setting the foreground alpha.
-        # This only has an effect when "Show blank space" is turned on.
-        re_blank = re.compile("\s+")
-        match = re_blank.search(text, offset)
-        while match:
-            start, end = match.span()
-            start = max([0, start+offset])
-            end = max([0, end+offset])
-            format = self.format(start)
-            color_foreground = format.foreground().color()
-            color_foreground.setAlpha(80)
-            self.setFormat(start, end-start, color_foreground)
-            match = re_blank.search(text, match.end())
+        
+        self.highlight_spaces(text, offset)
         
         self.setCurrentBlockState(state)
         
@@ -550,7 +562,9 @@ class CppSH(BaseSH):
                         self.setFormat(start, end-start, self.formats[key])
                     
             match = self.PROG.search(text, match.end())
-
+        
+        self.highlight_spaces(text)
+        
         last_state = self.INSIDE_COMMENT if inside_comment else self.NORMAL
         self.setCurrentBlockState(last_state)
 
@@ -625,6 +639,8 @@ class FortranSH(BaseSH):
                                            self.formats["definition"])
                     
             match = self.PROG.search(text, match.end())
+        
+        self.highlight_spaces(text)
 
 class Fortran77SH(FortranSH):
     """Fortran 77 Syntax Highlighter"""
@@ -632,6 +648,7 @@ class Fortran77SH(FortranSH):
         text = to_text_string(text)
         if text.startswith(("c", "C")):
             self.setFormat(0, len(text), self.formats["comment"])
+            self.highlight_spaces(text)
         else:
             FortranSH.highlightBlock(self, text)
             self.setFormat(0, 5, self.formats["comment"])
@@ -687,7 +704,8 @@ class DiffSH(BaseSH):
             self.setFormat(0, len(text), self.formats["number"])
         elif text.startswith("@"):
             self.setFormat(0, len(text), self.formats["builtin"])
-
+        
+        self.highlight_spaces(text)
 
 #==============================================================================
 # NSIS highlighter
@@ -817,6 +835,8 @@ class BaseWebSH(BaseSH):
             
             match = self.PROG.search(text, match.end())
             match_count += 1
+        
+        self.highlight_spaces(text)
 
 def make_html_patterns():
     """Strongly inspired from idlelib.ColorDelegator.make_pat """
@@ -899,7 +919,8 @@ class PygmentsSH(BaseSH):
             start = ct
             ct += len(val)        
             self.setFormat(start, ct-start, self.formats[key])
-
+        
+        self.highlight_spaces(text)
 
 def guess_pygments_highlighter(filename):
     """Factory to generate syntax highlighter for the given filename.
--- a/spyderlib/widgets/sourcecode/syntaxhighlighters.py
+++ b/spyderlib/widgets/sourcecode/syntaxhighlighters.py
@@ -203,7 +203,8 @@ class BaseSH(QSyntaxHighlighter):
         Derived classes could call this function at the end of
         highlightBlock().
         """
-        show_blanks = self.document().defaultTextOption().flags() & QTextOption.ShowTabsAndSpaces
+        flags_text = self.document().defaultTextOption().flags()
+        show_blanks =  flags_text & QTextOption.ShowTabsAndSpaces
         if show_blanks:
             format_leading = self.formats.get("leading", None)
             format_trailing = self.formats.get("trailing", None)
@@ -220,7 +221,8 @@ class BaseSH(QSyntaxHighlighter):
                     self.setFormat(start, end, format_leading)
                 format = self.format(start)
                 color_foreground = format.foreground().color()
-                color_foreground.setAlphaF(self.BLANK_ALPHA_FACTOR * color_foreground.alphaF())
+                alpha_new = self.BLANK_ALPHA_FACTOR * color_foreground.alphaF()
+                color_foreground.setAlphaF(alpha_new)
                 self.setFormat(start, end-start, color_foreground)
                 match = self.BLANKPROG.search(text, match.end())
     
--- a/spyderlib/widgets/sourcecode/syntaxhighlighters.py
+++ b/spyderlib/widgets/sourcecode/syntaxhighlighters.py
@@ -414,28 +414,19 @@ class PythonSH(BaseSH):
                     
             match = self.PROG.search(text, match.end())
  
-        # Make blank space less apparent by setting the foreground
-        # color closer to the background color.
+        # Make blank space less apparent by setting the foreground alpha.
         # This only has an effect when "Show blank space" is turned on.
-        for space_character in [" ", "\t"]:
-            pos = text.find(space_character, 0)
-            while pos >= 0:
-                posoffset = pos + offset
-                if posoffset >= 0:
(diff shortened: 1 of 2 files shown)


hugobuddel:
Left the old syntax highlighting, right the new. Top default Spyder scheme, bottom Spyder/Dark scheme. Especially the spaces in 'real' code improve readability, like the ones around the equal signs.

Apparently something goes wrong with the spaces after the tab character, which I only noticed when making the screenshots.

spaces

hugobuddel:
The issue with the tab character has been fixed. Not going to make new screenshots because it is bedtime here.

Perhaps the code can be improved by matching a white space regexp instead of reformatting one character at a time using a double loop. But it works and isn't that bad IMO.

ccordoba12:
@hugobuddel, thanks a lot for taking the time to work on this :-)

@goanpeca, would you mind to review this PR?

spyderlib/widgets/sourcecode/syntaxhighlighters.py
```diff
@@ -414,6 +414,29 @@ def highlightBlock(self, text):
...
+                    format = self.format(posoffset)
+                    color_foreground = format.foreground().color()
+                    color_background = format.background().color()
+                    # Blend the colors.
+                    fr, fg, fb = 0.3, 0.3, 0.3
+                    color_foreground_new = QColor(
+                        fr*color_foreground.red() + (1-fr)*color_background.red(),
+                        fg*color_foreground.green() + (1-fg)*color_background.green(),
+                        fb*color_foreground.blue() + (1-fb)*color_background.blue(),
+                        255,
```
Nodd: Does it work to only changing the alpha value ? This would require less computations.

hugobuddel:
Indeed @Nodd, now setAlpha and a whitespace regular expression is used. Now it is cleaner, more uniform in code style and probably useful if we ever would like to show new-line characters for some reason.

By the way, funny that `syntaxhighlighters.py` is one of the few files that actually has tabs to highlight.

goanpeca:
Looks good :-) :+1:

spyderlib/widgets/sourcecode/syntaxhighlighters.py
```diff
@@ -414,6 +414,20 @@ def highlightBlock(self, text):
                     
             match = self.PROG.search(text, match.end())
  
+        # Make blank space less apparent by setting the foreground alpha.
+        # This only has an effect when "Show blank space" is turned on.
+        re_blank = re.compile("\s+")
```
goanpeca: This line should perhaps be a class variable (look on top for)

```python
     # Syntax highlighting rules:
     PROG = re.compile(make_python_patterns(), re.S)
     IDPROG = re.compile(r"\s+(\w+)", re.S)
```

This way the compilation is done only once when the class is instantiated, and the code would be more uniform, as you can see no regex is compiled within the highlighting method.

spyderlib/widgets/sourcecode/syntaxhighlighters.py
```diff
@@ -414,6 +414,20 @@ def highlightBlock(self, text):
...
+        # This only has an effect when "Show blank space" is turned on.
+        re_blank = re.compile("\s+")
+        match = re_blank.search(text, offset)
+        while match:
+            start, end = match.span()
+            start = max([0, start+offset])
+            end = max([0, end+offset])
+            format = self.format(start)
+            color_foreground = format.foreground().color()
+            color_foreground.setAlpha(80)
```
goanpeca: Perhaps since you are hard coding the alpha value, it would be better to have it in a more "accessible/set-able" place? @Nodd suggestions on this? 

maybe 

```python
            color_foreground = format.foreground().color()
            color_foreground.setAlpha(self.SPACE_TAB_ALPHA)`

And `SPACE_TAB_ALPHA` can be also a class variable?

**Comment by goanpeca** [comment]
Thinking a bit more about this, this code should work in the other syntax highlighters, so perhaps it could be refactored as a method inside the `BaseSH`  like `def highlight_spaces`, and add the class variables there.

**Comment by hugobuddel** [comment]
Ah, `PROG` is used to prevent compilation of the regular expression. That makes sense. I'll refactor the code a bit further.

About the hard-coded value: it seems to 'expose' too much of the internal working of QColor. Perhaps it would be nicer to multiply with a factor, like:

            color_foreground = format.foreground().color()
            color_foreground.setAlpha(0.3 * color_foreground.alpha())

Or perhaps indeed with a setable class variable. Although personally I don't think it would be worthwhile to provide a user interface for it until someone asks for it.

File: spyderlib/widgets/sourcecode/syntaxhighlighters.py [hunk 4]
```diff
@@ -414,6 +414,20 @@ def highlightBlock(self, text):
...
             match = self.PROG.search(text, match.end())
  
+        # Make blank space less apparent by setting the foreground alpha.
+        # This only has an effect when "Show blank space" is turned on.
+        re_blank = re.compile("\s+")
+        match = re_blank.search(text, offset)
+        while match:
+            start, end = match.span()
+            start = max([0, start+offset])
+            end = max([0, end+offset])
```
Nodd: In which cases `start+offset` or `end+offset` could be negative ?

goanpeca:
Great. Is not about providing an user interface, is to keep the code cleaner I guess... and yes, a factor seems better.

goanpeca:
I like the approach but it would only seem to work on python files so far, isnt it?

spyderlib/widgets/sourcecode/syntaxhighlighters.py
hugobuddel: `start+offset` and `end+offset` are negative when a line in a multiline string starts with spaces.

That code is copied from the loop earlier in the function. Interpreting the code (and some experimentation) taught me that `highlightBlock()` is called for each line of text, which might be part of a multiline string. The relevant quotes are temporarily prefixed to the line. The number of added characters is tracked by the negative valued `offset` variable.

spyderlib/widgets/sourcecode/syntaxhighlighters.py
```diff
@@ -414,6 +414,20 @@ def highlightBlock(self, text):
                     
             match = self.PROG.search(text, match.end())
  
+        # Make blank space less apparent by setting the foreground alpha.
```
goanpeca: All this should have an if statement tied to the option of showing spaces, no?

Also to make it work with other syntax highlighters it should be refactored as a separate method inside BaseSH

hugobuddel:
Now the less pronounced space highlighting should work with all file types. However, I've only checked `.py`, `.f77` and `.cc` because it was easy to guess the required extensions for those. Is there a testing framework to automate this?

The code is moved to `BaseSH.highlight_spaces()`, which is called in the `highlightBlock()` function of each derived class. Although this naming (as suggested by @goanpeca) is inconsistent, I prefer it because now the only deviation from PEP 8 is forced by Qt.

A check for shown spaces isn't necessary, because normal space and tab glyphs have no content. So it doesn't matter if we change the alpha value of those.

goanpeca:
@hugobuddel thanks for all the updates :-) great work. What I meant from, the check is to avoid unnecessary calculations if the option is not selected, cause even if it does not affect the look it will add an overhead that could be easily checked, dont you think?

goanpeca:
About PEP 8 consistency, we follow it for everything except when Qt is involved, so actually `highlight_spaces` falls within our convention :-) 

https://github.com/spyder-ide/spyder/wiki/Dev%3A-Coding-Style

spyderlib/widgets/sourcecode/syntaxhighlighters.py
```diff
@@ -187,10 +191,28 @@ def set_color_scheme(self, color_scheme):
             self.color_scheme = color_scheme
         self.setup_formats()
         self.rehighlight()
-
+
```
Nodd: Please remove those spaces.

hugobuddel: There doesn't seem to be a consensus about spaces on otherwise empty lines, at least not in syntaxhighlighters.py. If anything, it is to keep those spaces.

Personally I like the indentation of empty lines to match the indentation of the next non-empty line. That is the easiest to read with 'show blank space' in my opinion, because it visually groups code that belongs together. Also it prevents the cursor from needlessly jumping around horizontally when using the up/down arrows to move.

Nevertheless, I'll change it. My editors usually add these spaces (which I like) but I remove them before committing if it is a modification of existing code. This one apparently slipped through. (I even kept the line in syntaxhighlighting.py that had just one single space until I couldn't stand it anymore a few commits ago..)

goanpeca:
@hugobuddel indeed most of the spyder code base has a lot of blank lines due to autoindentation in editors etc... but we are trying to fix this on every PR that is released. There is also a pending PR that will most likely erase this blank lines (to be done still!)

hugobuddel:
A personal note about 'fixing' the whitespace created by autoindentation in editors. The whitespace in spyder will not affect me much because I'll be mainly using spyder on my own code. Nevertheless, my motivation for the changes in this pull request seem to conflict with the spyder whitespace policy, so it is good to be aware of this discrepancy even though it probably will not matter much in practice.

For me, empty lines matching the indentation level of the lines of code are
+  beautiful with a good syntax highlighter, which we try to achieve here,
+ useful, because it a)prevents the cursor from jumping horizontally unnecessarily and b) makes it easier to copy/paste code to a Python REPL (although Spyder takes care of that), and
+ harmless.

From that perspective it is not a fix to remove spaces from otherwise empty indentation lines. A fix would be to match the indentation of the next line. The goal of this comment is to give you some more information to make the best decision, which very well might be to remove the lines anyway.  I'll adhere to whatever standard spyder will chose in future commits.

hugobuddel:
Some more changes are made to the pull request:
* Superfluous spaces have been removed :-).
* A check is performed to see whether spaces are shown at all. The code editor instance has an attribute called `blanks_enabled`. Is it possible to reach this attribute from the syntax highlighter instances?
* All file formats are tested. In the end these extensions triggered all syntax highlighter classes: py, jl, diff, json, yaml, cc, html, f, f77.
* The code editor rehighlights when the 'show blanks' option is toggled. However, I noticed that changing the option in the preferences does not always toggle the option in the 'Source' menu. It seems to be related to having multiple files open, but couldn't pinpoint it.

goanpeca:
@hugobuddel thanks for the work on these updates. 

As stated before we try to follow PEP8 as closely as possible, hence empty lines are unwanted and will (and must) be removed from any future PR. This does not mean removing old blank spaces left by previous commits, but making sure that the PR by itself does not introduce new blank spaces.

Hope it is clear :-)

hugobuddel:
Apparently [PEP 8 is ambiguous about indenting empty lines](http://stackoverflow.com/questions/2727988/python-indentation-in-empty-lines). Nevertheless, consensus (through `pep8.py`) seems to be to remove the lines.

Perhaps it was useful to bring it up. The code in this pull request can easily be modified to show trailing spaces in red as suggested in #1805 by modifying the regular expression. However, doing so would destroy the feature's usefulness for me entirely. It would be very aesthetically unpleasing for people who like indented empty lines (e.g., me) to have them show up in red, while the entire point of this PR is to improve readability.

Apparently the opinions on showing whitespace are more diverse than we imagined :-).

hugobuddel:
As far as I'm concerned this PR is as good as it has to be and can be pulled unless there are other concerns.

The goal of this PR was to improve the legibility of the code when spaces and tabs are shown, by making the blanks blend a bit with the background. It does this well by ajusting the alpha channel of the spaces, so it works in light and dark schemes. The performance penalty is low because it checks whether blanks are shown and reuses the regular expression. All syntax highlighters are supported and tested. The code follows the design practices of spyder.

The code can be extended to cover other use cases. E.g. different alpha/color for indentation, intra-line-blanks and trailing blanks. However, those are not my use cases, so I'm not in a hurry to work on them. I suggest to merge this code if it is acceptable and create new issues for those other use cases, if so desired. Perhaps closing #2175 would require some of those other changes.

hugobuddel:
Overloaded `highlight_spaces()` in `PythonSH` to show indentation spaces in multiline docstrings in normal format. This looks better because those spaces are ignored by most (all?) docstring parsers and should therefore not show up as part of the string (i.e. green) when highlighting. However, legitimate spaces at the start of a line are also shown with the normal format. Those legitimate spaces are less common than indentation spaces, so this is an improvement.

goanpeca:
Great, can you post an image? Thanks :-)

hugobuddel:
Left: original. Middle: this PR from yesterday. Right: this PR with latest commit. The color of the space glyphs for the docstring of `temp()` is the best in the right image were the gray matches the indentation. The spaces within the docstring are green as desired. However, it could be argued that the leading spaces inside `my_html` that are now gray ought to be green. This latter case is rare and actually doesn't look so bad, while the docstring looks pretty good.
instring

goanpeca:
@hugobuddel agreed, it does look much better

spyderlib/widgets/sourcecode/syntaxhighlighters.py
```diff
@@ -424,6 +453,36 @@ def highlightBlock(self, text):
...
+                if inside_string and start == 0:
+                    self.setFormat(start, end, self.formats["normal"])
+
+                format = self.format(start)
+                color_foreground = format.foreground().color()
+                color_foreground.setAlphaF(self.BLANK_ALPHA_FACTOR * color_foreground.alphaF())
+                self.setFormat(start, end-start, color_foreground)
+                
+                match = self.BLANKPROG.search(text, match.end())
+
```
goanpeca: @hugobuddel, thanks for the change, it is a very nice work and as soon as the PR is merged I will leave the option ON, but I think we can refactor the code to make this better (the new code is almost an exact copy of the base class, but with a couple of extra lines!)

It would be much better in this case to write an new method in the base syntax highlighter that returns True (or False) by default and gets overloaded in the child syntax highlighters classes as needed

```python

# in the base class
def is_inside_string(self):   # Random name ... if you think of a better one please go ahead
    return True  # or False...

# in the child highlighter class... python in this case
def is_inside_string(self):
    """Overloaded method to account for all the different python strings."""
    inside_string = self.previousBlockState() in (
        self.INSIDE_SQSTRING,
        self.INSIDE_DQSTRING,
        self.INSIDE_SQ3STRING,
        self.INSIDE_DQ3STRING,
         )
    return inside_string
```

This way we avoid duplicate code and make it more future proof.

Cheers ;-)

goanpeca: And then of course you could add your new code inside the original `highlight_spaces` method.

hugobuddel:
Thanks @goanpeca, it is now refactored in an even better way.

I didn't like the duplication of code, and indeed though of generalizing the 'in a string' part, but disliked that as well. You showed my why by taking the time to write it out: are there actually other use cases that resemble the Python multiline docstring like the earlier screenshot? I can't think of many.

Instead, it was simpler to just recolor leading spaces, which is easy to generalize. And, it was trivial to add support for trailing spaces as well! The leading space format is hardcoded to 'normal', which is fine I guess. But the trailing space format is now hardcoded to 'number', because that is dark red in the default scheme. Any suggestions on how to improve on that?

leadingtrailing

spyderlib/widgets/sourcecode/syntaxhighlighters.py
```diff
@@ -190,7 +195,33 @@ def set_color_scheme(self, color_scheme):
...
+        show_blanks = self.document().defaultTextOption().flags() & QTextOption.ShowTabsAndSpaces
+        if show_blanks:
+            format_leading = self.formats.get("leading", None)
+            format_trailing = self.formats.get("trailing", None)
+            match = self.BLANKPROG.search(text, offset)
+            while match:
+                start, end = match.span()
+                start = max([0, start+offset])
+                end = max([0, end+offset])
+                if end == len(text) and format_trailing is not None:
```
goanpeca: Much better, maybe just add a comment on top of this line?

`# Formats points at the start of the line... leading points`

spyderlib/widgets/sourcecode/syntaxhighlighters.py
```diff
@@ -190,7 +195,33 @@ def set_color_scheme(self, color_scheme):
...
+                    self.setFormat(start, end, format_trailing)
+                if start == 0 and format_leading is not None:
```
goanpeca: Same here

`# Formats points at the end of the line... trailing points`

goanpeca:
I think that both trailing and leading points should be normal (color) for now.

hugobuddel:
Added some comments and used normal format for trailing spaces. Perhaps we can support a different format for trailing spaces in the future, the possibility now exists. I'd say the PR is once again as good as it should be!

goanpeca:
Review done!

@Nodd can you take a quick look?

goanpeca:
@ccordoba12 review done!

spyderlib/widgets/sourcecode/syntaxhighlighters.py
```diff
@@ -190,7 +195,35 @@ def set_color_scheme(self, color_scheme):
...
-            
+
+    def highlight_spaces(self, text, offset=0):
+        """
+        Make blank space less apparent by setting the foreground alpha.
+        This only has an effect when 'Show blank space' is turned on.
+        Derived classes could call this function at the end of
+        highlightBlock().
+        """
+        show_blanks = self.document().defaultTextOption().flags() & QTextOption.ShowTabsAndSpaces
```
goanpeca: This line is probably longer than 79?

I am actually flexible on that PEP8 rule but I think @ccordoba12 is more serious about it :-)

spyderlib/widgets/sourcecode/syntaxhighlighters.py
```diff
@@ -190,7 +195,35 @@ def set_color_scheme(self, color_scheme):
...
+                end = max([0, end+offset])
+                # Format trailing spaces at the end of the line. 
+                if end == len(text) and format_trailing is not None:
+                    self.setFormat(start, end, format_trailing)
+                # Format leading spaces, e.g. indentation.
+                if start == 0 and format_leading is not None:
+                    self.setFormat(start, end, format_leading)
+                format = self.format(start)
+                color_foreground = format.foreground().color()
+                color_foreground.setAlphaF(self.BLANK_ALPHA_FACTOR * color_foreground.alphaF())
```
goanpeca: This line is probably longer than 79?

I am actually flexible on that PEP8 rule but I think @ccordoba12 is more serious about it :-)

hugobuddel:
These lines indeed were a bit long and are split up. The code is perhaps easier to understand now by using extra variables.

spyderlib/widgets/sourcecode/syntaxhighlighters.py
```diff
@@ -413,9 +450,14 @@ def highlightBlock(self, text):
                                                    self.formats["keyword"])
                     
             match = self.PROG.search(text, match.end())
- 
+        
         self.setCurrentBlockState(state)
         
+        # Use normal format for indentation and trailing spaces.
+        self.formats['leading'] = self.formats['normal']
+        self.formats['trailing'] = self.formats['normal']
```
goanpeca: @hugobuddel is this only applied to Python then?

hugobuddel: Yes on purpose @goanpeca, because it seems that it is not useful in any of the other syntax highlighter classes. It only recolors the green indentation (and trailing) spaces of multiline strings to the 'normal' format and none of the other classes have something similar: they already color all leading and trailing spaces with the 'normal' format. Or they already have highlighting rules that shouldn't change, like multiline C++ comments or the Fortran rules. I probably missed some cases though and better be save than sorry and default to not changing them.

goanpeca:
Got it, thanks for your work on this

ccordoba12:
@goanpeca, is this ready for merge?

goanpeca:
It is working as expected so I would say yes (unless you find something weird in the code ;-) )

ccordoba12:
Since you reviewed this one, please do the honors of pressing the green button ;-)

goanpeca:
:+1:

hugobuddel:
Thank you @goanpeca @ccordoba12 and @Nodd , it was great working with you. The patch got much better through our conversation!

goanpeca:
@hugobuddel thanks for your work on it :-), it was an issue bothering a lot so of people. It does look much better now.

We are always happy to have new contributors, and hopefully you can continue helping us improve Spyder.

Cheers

hugobuddel:
Perhaps I will contribute more, this was just scratching my own itch :-). So far I like spyder, and I'll try to use it for more projects.

All this space highlighting stuff made me realize that it must be possible to do better than such ASCII-based code editing. It seems archaic. Like the 79 character limit: I agree that it is a good style guide and try to follow it, but it feels bad that it is a necessity at all! Why should we as programmers worry about things like spaces and line breaks? It must be possible to decouple the presentation of the code from how it is stored. We shouldn't need a PR like this at all.

Achieving such a separation of presentation is a whole other discussion. And it  is probably not yet done because it is too hard to get right. Nevertheless, perhaps I'll try to see how far current tools make this possible already now.

goanpeca:
We all start with own itches ;-)

I am not following you, but my take on this is that all theses are about being consistent and at the end code is more read than written and having code written in a consistent manner can take projects farther.

Having a set of rules (even if some might feel archaic) helps in having have a cleaner codebase, helps people reading the code know what to expect, and helps developers in the process.

hugobuddel:
Yes I fully agree that the rules are necessary. And consistency is more important than having perfect rules. Because as you say, it improves collaboration by having easier to understand code.

Perhaps I'll need some more time to figure out what my 'thesis' actually is :-). This thread helped me with my thought process, which is why I mentioned it, but it is not the right place to really discuss it.

It's that our tools (and perhaps languages themselves) could be so much better in assisting us with writing/reading code than they are now. My feeling is that in a perfect tool it wouldn't be necessary or desirable to 'show spaces' at all, even to the extent that it would be conceptually meaningless to do so. But I'll stick with spyder; perfect tools don't exist and spyder is a great leap towards better tools in many ways. 

Hopefully someone will one day revert this merge because it has become useless :-).

goanpeca:
:+1:

PR #2532, opened 2015-07-13
PR: In findreplace, if nothing is highlighted, use word under cursor

sonofeft:
Search/Replace can be simplified by placing the cursor on the word to be
searched and simply hitting the hot key (ctrl-f or ctrl-h)


ccordoba12:
@goanpeca, would you mind to review this one? Thanks ;-)

spyderlib/widgets/findreplace.py
```diff
@@ -198,6 +198,39 @@ def show(self):
         self.visibility_changed.emit(True)
         if self.editor is not None:
             text = self.editor.get_selected_text()
+            
+            # If no text is highlighted for search, use whatever word is under the cursor
+            if len(text)==0:
```
goanpeca: The more pythonic way would be simply to check for

`if text:`

spyderlib/widgets/findreplace.py
```diff
@@ -198,6 +198,39 @@ def show(self):
...
         if self.editor is not None:
             text = self.editor.get_selected_text()
+            
+            # If no text is highlighted for search, use whatever word is under the cursor
+            if len(text)==0:
+                try:
+                    # Get line text and cursor position
+                    line_number, index = self.editor.get_cursor_line_column()
+                    block = self.editor.document().findBlockByNumber(line_number)
+                    data = str( block.text() )
```
goanpeca: Why spaces around parens?

spyderlib/widgets/findreplace.py
```diff
@@ -198,6 +198,39 @@ def show(self):
...
+                    ibeg = index
+                    # Limit the search for the beginning of the word to a "reasonable" amount
+                    while (ibeg>0) and (ibeg>index-80):
```
goanpeca: What is the idea of this "reasonable" amount, can you elaborate on this please?

Also... it should be `while ibeg > 0 and ibeg > (index-80):`

spyderlib/widgets/findreplace.py
```diff
@@ -198,6 +198,39 @@ def show(self):
...
+                        if (data[ibeg].isalnum()) or (data[ibeg]=='_'):
```
goanpeca: `data[ibeg].isalnum() or data[ibeg] == '_':`

spyderlib/widgets/findreplace.py
```diff
@@ -198,6 +198,39 @@ def show(self):
...
+                            ibeg -= 1
+                        else:
+                            ibeg += 1
+                            break
+                    
+                    iend = index
+                    # Limit the search for the end of the word to a "reasonable" amount
+                    while (iend<len(data)) and (iend<index+80):
```
goanpeca: while iend < len(data) and iend < (index + 80):

goanpeca: If you are going to hard code the "reasonable" amount and use it in several places, then it would be better to define a variable...

Something like

`max_search_limit = 80`

spyderlib/widgets/findreplace.py
```diff
@@ -198,6 +198,39 @@ def show(self):
...
+                            iend += 1
+                        else:
+                            break
+                    
+                    text = data[ibeg:iend]
+                    
+                except:
+                    # If this fails for any reason, don't worry about it
+                    #  Just let the user type whatever they want into the search entry box
+                    pass
```
goanpeca: Hmmmm it is a bit funky to leave an empty exception block. 

If there is an error in your code, it will be hard to discover or debug 'cause all gets trapped here. We should define the specific error we are expecting and catch and specific one..

Take a look at [this](https://docs.python.org/2/library/exceptions.html)

goanpeca:
@sonofeft thanks for working on this. 

I made some comments on style and some other minor issues that need to be handled. I will test the functionality later on.

Cheers :+1:

goanpeca:
By the way, is there some issue that is being fixed by this PR, or it was just a random enhancement ;-) ?

spyderlib/widgets/findreplace.py
goanpeca: also the `_` seems to be used in two places, a variable would be be better in this case

spyderlib/widgets/findreplace.py
goanpeca: Is this working on python 3?

Maybe it is better to use the helper function `to_text_string`

sonofeft:
Thanks for all the feedback.  I will make updates soon. 

It was a random enhancement... I like that feature in Scintilla.

sonofeft:
Made suggested changes:

used "to_text_string" helper function
got rid of multiple "_" usages
modeled exception clause after others in spyder codebase
added constant for "reasonable" search limit
made some formatting adjustments to accepted norms
changed "if len(text)" to more pythonic "if text" (including existing code base withing show()"

Let me know if there's anything else you'd like.

spyderlib/widgets/findreplace.py
```diff
@@ -29,32 +30,36 @@
...
     """Return True is pos1 > pos2"""
     return pos1 > pos2
-    
+
 def is_position_inf(pos1, pos2):
     """Return True is pos1 < pos2"""
     return pos1 < pos2
 
+def is_a_var_char(c):
+    """Return True if c is a valid variable character, False otherwise"""
```
goanpeca: `is_valid_var_char` ?

spyderlib/widgets/findreplace.py
```diff
@@ -29,32 +30,36 @@
...
+    return c.isalnum() or (c == '_')
+MAX_SEARCH_LIMIT = 80  # Used in Find/Replace default word logic
```
goanpeca: This could be inside `FindReplace` class as a class variable....

```python
class FindReplace(QWidget):
     """Find widget"""
     MAX_SEARCH_LIMIT = 80  # Used in Find/Replace default word logic
     STYLE = {False: "background-color:rgb(255, 175, 90);",
              True: ""}
     visibility_changed = Signal(bool)
```

and then use it like `self.MAX_SEARCH_LIMIT`

goanpeca:
@sonofeft will try to review this one this weekend!

spyderlib/widgets/findreplace.py
```diff
@@ -191,14 +196,55 @@ def toggle_highlighting(self, state):
...
-            if len(text) > 0:
+
+            # If no text is highlighted for search, use whatever word is under the cursor
+            if not text:
+                try:
+                    # Get line text and cursor position
+                    line_number, index = self.editor.get_cursor_line_column()
+                    block = self.editor.document().findBlockByNumber(line_number)
+                    data = to_text_string(block.text())
+                    if index < len(data):
```
Nodd: The test here boils down to "is the cursor in the line". It seems unecessary, why would it be outside ?

Nodd:
I think the whole functionnality could be replaced by something like `cursor.select(QTextCursor.WordUnderCursor)` (see http://doc.qt.io/qt-4.8/qtextcursor.html#select).
One downside is that you don't have a fine control over what is considered a word, but I don't think that's a problem for such a functionality. It would greatly simplify the code !

ccordoba12:
@sonofeft, what do you think about @Nodd suggestion?

sonofeft:
Simple is good.  
I'm more concerned with providing the functionality to reduce editing keystrokes than a specific implementation.
If a new, simpler implementation is available, I'm all for it.

ccordoba12:
Ok, I have two things to say:

1. Please restore the blanks you removed while working on this PR. That makes reviewing and porting PRs quite complicated, so we are not in favor of that :-)
2. Do you plan to follow @Nodd advice of using `QTextCursor.WordUnderCursor` or not? Sorry, but it was not clear to me :-)

sonofeft:
Sorry about the blanks, I had the Spyder option to remove trailing blanks turned on, so Spyder does it automatically. (I guess the shame of pylint messages was too hard for me to bear)

Also sorry I mistook the @Nodd advice to be more notional than as a concrete proposal.

I can start from a clean fork (with blanks restored) and play around with his idea to see if I can make it work

I assume it would be something like:

cursor = self.editor.textCursor()

I can see where a whole bunch of my code goes away it the above three lines get "text" populated.

If the above three lines look correct to you, I will try to get that done this week.

sonofeft:
Using QTextCursor.WordUnderCursor went surprisingly fast... 
I went ahead and just made the changes.

The testing I've done seems to work fine.

sonofeft:
I need a suggestion from @Nodd to fix the python 3.4 w/o Conda error.

It looks like a Qt version issue.

Nodd:
All tests fail, but not with the same problem. One is:
```
-- Testing standard Python socket interface --
  File "spyderlib/utils/bsdsocket.py", line 160, in <module>
    client.send("data to be catched")
TypeError: 'str' does not support the buffer interface

The command "./continuous_integration/travis/modules_test.sh" exited with 1.
```
And the other with:
```No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.

The build has been terminated
```

All of these are failures, I don't know where they come from. I'll close/reopen the bug to retrigger the checks, because it isn't related to your code.

ccordoba12:
@sonofeft, please rebase or merge with latest `master` and push again.

I changed some things about our integration with Travis and that's not letting our tests to finish in your branch :-)

goanpeca:
@ccordoba12, or @Nodd can you review this one?

goanpeca:
@ccordoba12, or @Nodd can you review this one?

spyderlib/widgets/findreplace.py
```diff
@@ -198,6 +198,14 @@ def show(self):
...
             text = self.editor.get_selected_text()
+
+            # If no text is highlighted for search, use whatever word is under the cursor
+            if not text:
+                cursor = self.editor.textCursor()
+                cursor.select(QTextCursor.WordUnderCursor)
+                text = to_text_string(cursor.selectedText())
+
+            # Now that text value is sorted out, use it for the search
             if len(text) > 0:
```
Nodd: `if text:` would be more pythonic. I know it's not from you but since you (correctly) used `if not text:` just before, it will be more coherent.

Nodd:
Why all the updates to the translation files ? It isn't related to your PR. It looks like a merge with master gone wrong.
Apart from this the code itself looks fine.

sonofeft:
I apologize if I merged badly.  As I recall, I was too out of date to merge cleanly and had to futz around.

On the "if text" comment... are you asking me to change the existing code's "if len(text) > 0" line to "if text"?

Nodd:
I think that you have to revert the changes to spyderlib.mo/spyderlib.po and keep only findreplace.py. Feel free to create another PR from scratch if it's simpler for you, since the changes are very small.

Yes please make the change, it will clean our codebase.

sonofeft:
I went ahead and started a new Pull Request.
Added Find/Replace Enhancement #2957

goanpeca:
Can we close this one then @sonofeft ?

ccordoba12:
That seems to be the case because this has some weird changes (e.g. to our translation files).

@sonofeft, thanks for your interest on this one! :-)

sonofeft:
Glad to help. Thanks for guiding me through the process.

Issue #2527, opened 2015-07-11
More suggestions for the "Spyder 3" icon theme


ccordoba12:
Discussed with @SylvainCorlay in person:

In Preferences:
- [x] Use a keyboard icon for *Keyboard shortcuts* entry
- [x] Use a palette icon for *Syntax coloring*
- [x] Use the Python icon for *Console*
- [x] Use `question-circle` for the *Object Inspector*

Other suggestions:
* Add a menu entry to our main contextual menu so that people can decide the size of icons in toolbars. Suggested options: small: 16x16, medium: 24x24, large: 32x32, big: 48x48.
* Add the possibility to turn off hinting in qtawesome for Linux.

ccordoba12:
I moved the remaining tasks to issue #2490, to have everything in once place.

PR #2533, opened 2015-07-13
Spyder 3 icon theme changes

SylvainCorlay:
As per the suggestions in #2527.
- IP: for the ipython logo
- keyboard icon for the keyboard shortcuts
- a new icon for coloring theme, it is not a palette, but an eyedropper. The advantage is that it is a font-awesome icon.
- question-circle icon for the object inspector.

options


SylvainCorlay:
I have not change the command promt / terminal icon yet because it is used in different places, I need to be sure of what to replace.

ccordoba12:
Great job! Any thoughts @goanpeca?

goanpeca:
Looks cool ! :+1:

SylvainCorlay:
Now using the Python logo for the console config session as asked by Carlos. Ready to get in.

ccordoba12:
I have one suggestion: the `IP:` icon seems a bit too big next to text to its right. Could you make it a bit smaller?

SylvainCorlay:
In that instance only or all the time? The former would require adding a special entry in the icon manager, and I thought the problem with the previous one was that it was too thin...

ccordoba12:
What about this: framing the `IP:` symbol inside a (transparent) rectangle with the top and bottom borders a bit thicker, that represents the console? I mean, something very similar to our current IPython icon :-)

SylvainCorlay:
Well, that is not very "flat design" like font-awesome icons.

ccordoba12:
I don't understand what you mean. The icon would be similar to

http://fortawesome.github.io/Font-Awesome/icon/behance-square

or

http://fortawesome.github.io/Font-Awesome/icon/google-plus-square/

i.e. text inside a square (rectangle in our case) but with the colors inverted.

ccordoba12:
Another example:

http://fortawesome.github.io/Font-Awesome/icon/git-square/

Maybe we can create our IPython icon as these ones (to follow FontAwesome aesthetics :-), i.e. a white `IP:` over a black rectangular background.

SylvainCorlay:
OK, I see what you mean. I will add something like this.

SylvainCorlay:
@ccordoba12 done!
config

SylvainCorlay:
Merge?

ccordoba12:
The IPython icon is a bit blurry when used in our console tabs:

seleccion_001

Can you do something about it? I say it because the other icons look crystal clear ;-)

SylvainCorlay:
It is probably due to an icon size in the tabs that is not pixel perfect...

ccordoba12:
So merge? Or could you take a look at the icon size on tabs to see if we can use 16x16 icons there?

SylvainCorlay:
I can check, but I will be very busy this week.

ccordoba12:
Ok, merging then ;-)

PR #2544, opened 2015-07-15
be more PyQt5 friendly, when PyQt5 is here

### LINKED ISSUE #2545: PyQt5 is less nicely detected per Spyder than by IPython
testing PyQt5, I discovered that Ipython qtconsole detected it without problem, when Spyder was still trying to load PyQt4.

proposal here: https://github.com/spyder-ide/spyder/pull/2544

stonebig:
Fixes #2545

### DIFF
--- a/spyderlib/qt/__init__.py
+++ b/spyderlib/qt/__init__.py
@@ -8,7 +8,6 @@
 """Transitional package (PyQt4 --> PySide)"""
 
 import os
-import imp
 
 try:
     #be friendly with Pyqt5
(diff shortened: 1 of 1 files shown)


stonebig:
https://github.com/spyder-ide/spyder/pull/2544

ccordoba12:
This change would make PyQt5 the default, when both PyQt4 and PyQt5 are present. Unfortunately, I think we are not ready for that. We still want PyQt4 to be the default, at least for another couple of months.

Besides, IPython qtconsole is a much more simpler Qt app than Spyder, so that's why probably they didn't see much problem on making PyQt5 the default :-)

stonebig:
ok, I understand. 

I'll pre-set the 'QT_API' to 'pyqt5', for my special case.

Issue #2395, opened 2015-05-03
Can not show exception information correctly in IPython Console


cheneave:
I'm using the dev version of Spyder, and ipython v3.1.0
When a ipython console is first started, it shows exception information correctly. For example, if I type in a undefined name, it shows a NameError.
But after I run a source file, even an empty source file, the console will only show

Traceback (most recent call last):

  File "D:\Python\Python34\lib\site-packages\IPython\core\interactiveshell.py", line 3044, in run_code
    except self.custom_exceptions:

TypeError: catching classes that do not inherit from BaseException is not allowed

When exception is thrown

cheneave:
I have found a very dirty workaround, just insert 

    self.custom_exceptions = tuple(
            [e for e in self.custom_exceptions if issubclass(type(e),BaseException)])

to line 3031 of Python34\lib\site-packages\IPython\core\interactiveshell.py

ccordoba12:
@cheneave, could you provide a step by step guide to reproduce this error?

cheneave:
1. Create a empty script
2. run it by pressing F5
3. type any function with left parenthesis such as print( then the error will occor

My version of  python and ipython is 

Python 3.4.3 (v3.4.3:9b73f1c3e601, Feb 24 2015, 22:44:40) [MSC v.1600 64 bit (AMD64)]
IPython 3.1.0 -- An enhanced Interactive Python.

Other packages required by spyder are all the latest version from pip3

ccordoba12:
This is an IPython upstream problem, not an Spyder problem. I'll see what I can do about it.

PR #2547, opened 2015-07-16
Fix resetting IPython custom exception hook

takluyver:
Closes gh-2395

Ping @DSLituiev - this should fix the problem you showed me yesterday.


SylvainCorlay:
Looking at the docstring for `set_custom_exc`, I wonder why `None` was there in the first place. Was it required in older versions of IPython?

takluyver:
It's possible, but I don't think that code has changed for some time.

ccordoba12:
@takluyver, thanks a lot for taking a look at this one :-)

@blink1073, please take a look at this one, given that it was you who added this code.

blink1073:
Looks good to me.

PR #2559, opened 2015-07-23
Fix default layout dockwidget on first Spyder start

### LINKED ISSUE #2555: Main window is bigger than screen size, after a first start in Mac and KDE
This is the issue we discussed in person. Just as a reminder for you @goanpeca :-) I'm almost sure this was introduced after the new layouts PR.

This is how it looks for me in KDE, which is really bad.

seleccion_002

Please take a look at this one ASAP because I want this fixed for beta1 ;-)

You can test this with

    python bootstrap.py --test

goanpeca:
Fix #2555

### DIFF
--- a/spyderlib/spyder.py
+++ b/spyderlib/spyder.py
@@ -1496,7 +1496,7 @@ class MainWindow(QMainWindow):
                     'height fraction': [[1.0],          # column 0, row heights
                                         [1.0],          # column 1, row heights
                                         [1.0],          # column 2, row heights
-                                        [0.55, 0.45]],  # column 3, row heights
+                                        [0.50, 0.50]],  # column 3, row heights
                     'hidden widgets': [outline],
                     'hidden toolbars': [],                               
                     }
@@ -1666,7 +1666,7 @@ class MainWindow(QMainWindow):
         self._custom_layout_timer = QTimer(self)
         self._custom_layout_timer.timeout.connect(self.layout_fix_timer)
         self._custom_layout_timer.setSingleShot(True)
-        self._custom_layout_timer.start(100)
+        self._custom_layout_timer.start(5000)
 
     def layout_fix_timer(self):
         """Fixes the height of docks after a new layout is set."""
--- a/spyderlib/spyder.py
+++ b/spyderlib/spyder.py
@@ -1496,7 +1496,7 @@ class MainWindow(QMainWindow):
                     'height fraction': [[1.0],          # column 0, row heights
                                         [1.0],          # column 1, row heights
                                         [1.0],          # column 2, row heights
-                                        [0.50, 0.50]],  # column 3, row heights
+                                        [0.46, 0.54]],  # column 3, row heights
                     'hidden widgets': [outline],
                     'hidden toolbars': [],                               
                     }
--- a/spyderlib/spyder.py
+++ b/spyderlib/spyder.py
@@ -1408,6 +1408,7 @@ class MainWindow(QMainWindow):
         self.first_spyder_run = False
         if hexstate is None and not self.light:
             # First Spyder execution:
+            self.setWindowState(Qt.WindowMaximized)
             self.first_spyder_run = True
             self.setup_default_layouts('default', settings)
             self.extconsole.setMinimumHeight(250)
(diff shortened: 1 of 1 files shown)


goanpeca:
@ccordoba12 can you check if it works now as expected?

spyderlib/spyder.py
```diff
@@ -1701,34 +1649,37 @@ def setup_default_layouts(self, index, settings):
...
+                # estimated area in the window
+                new_height = int(layout['height fraction'][c][r]*height*0.95)
+                dockwidget.setMinimumHeight(new_height)
+                dockwidget.setMaximumHeight(new_height)
 
         self._custom_layout_timer = QTimer(self)
         self._custom_layout_timer.timeout.connect(self.layout_fix_timer)
         self._custom_layout_timer.setSingleShot(True)
-        self._custom_layout_timer.start(100)
+        self._custom_layout_timer.start(5000)
```
ccordoba12: Why this drastic change? I mean, you're changing the timer from 100 to 5000 :-)

goanpeca: Oops, that was for testing only, I will update

spyderlib/spyder.py
```diff
@@ -1718,17 +1669,17 @@ def setup_default_layouts(self, index, settings):
         self._custom_layout_timer.start(100)
```
goanpeca: @ccordoba12  fixed!, is it working?

ccordoba12:
Sorry, still now working. Now the Object Inspector is 1/4 of the vertical screen size and the consoles take the other 3/4. But the window still overflows the screen.

goanpeca:
Well... not sure what is the problem then, cause I cannot reproduce it.

ccordoba12:
Then please create a Kubuntu 14.04 virtual machine, so you can debug and solve this problem by yourself.

Another alternative would be to use the old layout system just the first time (because that one didn't have this problem).

goanpeca:
... ok

ccordoba12:
Sorry, but it's an ugly bug ;-)

goanpeca:
I installed Kubuntu 14.04 in a spare machine I had and checked and I see no overflow in the window when doing a fresh start (I deleted `.spyder2` prior to launching spyder).

I can get it to look right if I set the timer to a much higher value (tested with 5 seconds)...

spyderlib/spyder.py
ccordoba12: It works for me if I comment these lines related to `_custom_layout_timer`. With this change I can prevent the window overflowing issue in my machine :-)

spyderlib/spyder.py
```diff
@@ -1541,14 +1489,14 @@ def setup_default_layouts(self, index, settings):
...
-                                       0.0,
+                                       0.0,             # column 2 width
                                        0.45],           # column 3 width
                     'height fraction': [[1.0],          # column 0, row heights
-                                        [1.0],   # column 1, row heights
-                                        [0.55, 0.45],          # column 2, row heights
-                                        [1.0]],   # column 0, row heights
+                                        [1.0],          # column 1, row heights
+                                        [1.0],          # column 2, row heights
+                                        [0.55, 0.45]],  # column 3, row heights
```
ccordoba12: This looks better when I set it to `[0.5, 0.5]`

spyderlib/spyder.py
ccordoba12: Maybe we could avoid using this timer on a first start? I'd be happy with that temporary solution, at least for beta1 :-)

spyderlib/spyder.py
goanpeca: Ok, is it overflowing still? or is that fixed?

spyderlib/spyder.py
goanpeca: Can you change this timer to `5000` and check again locally?

spyderlib/spyder.py
ccordoba12: Still overflowing, but that's related to the timer you're using above, not to this.

spyderlib/spyder.py
ccordoba12: I did it already, there is no difference. Only when I comment it, I can prevent the overflowing :-)

goanpeca:
Ok, can you check again with this version now?

In my setup (kubuntu 14.04) it oes not overflow, and the sizes look ok.

goanpeca:
I am erasing everything (`.spyder`) and running this version and it looks like this after a clean start.

snapshot1

ccordoba12:
Yes, that's exactly how it's looking for me :-) So it's fixed, great!!

Then there's one thing left: the IPython console is too small compared to the Object Inspector. Please make it appear a bit bigger :-)

spyderlib/spyder.py
```diff
@@ -1541,14 +1489,14 @@ def setup_default_layouts(self, index, settings):
...
-                                       0.0,
+                                       0.0,             # column 2 width
                                        0.45],           # column 3 width
                     'height fraction': [[1.0],          # column 0, row heights
-                                        [1.0],   # column 1, row heights
-                                        [0.55, 0.45],          # column 2, row heights
-                                        [1.0]],   # column 0, row heights
+                                        [1.0],          # column 1, row heights
+                                        [1.0],          # column 2, row heights
+                                        [0.50, 0.50]],  # column 3, row heights
```
ccordoba12: Changing this to `[0.46, 0.54]` makes the IPython Console to have a decent vertical size :-)

goanpeca: I would advice against this. I need to better check what is the available height in the application (Total height minus menu bars and toolbars) and then set the ratios to 0.50, 0.50 which would in theory yield equally sized panes.

Adjusting manually to `[0.46, 0.54]` will make it look good in kubuntu, but it will most likely look bad in windows/osx/ubuntu.

ccordoba12: @goanpeca, could you do this change so I can merge this PR, as we agreed?

goanpeca: @ccordoba12 Done, sorry for the delays

spyderlib/spyder.py
Nodd: Hum, it's still 5000 in master...

Nodd:
I don't remember if code comments trigger a notification, but the timer value is still 5000 in master.

PR #2584, opened 2015-08-02
Homogenize variable explorer UI

goanpeca:
## Description

The "toolbar"  is now horizontal and the options button is aligned to the right, like other plugins.
Also now the variable explorer margins are updated when selection this option for panes inside the preferences

### Before

before-varex

### After
after-varex

## Todo
- [x] Make buttons horizontal
- [x] Align options menu to the right

### DIFF
--- a/spyderlib/plugins/variableexplorer.py
+++ b/spyderlib/plugins/variableexplorer.py
@@ -6,7 +6,7 @@
 
 """Namespace Browser Plugin"""
 
-from spyderlib.qt.QtGui import QStackedWidget, QGroupBox, QVBoxLayout
+from spyderlib.qt.QtGui import QGroupBox, QStackedWidget, QVBoxLayout, QWidget
 from spyderlib.qt.QtCore import Signal
 import spyderlib.utils.icon_manager as ima
 
@@ -75,7 +75,7 @@ class VariableExplorerConfigPage(PluginConfigPage):
         self.setLayout(vlayout)
 
 
-class VariableExplorer(QStackedWidget, SpyderPluginMixin):
+class VariableExplorer(QWidget, SpyderPluginMixin):
     """
     Variable Explorer Plugin
     """
@@ -84,9 +84,17 @@ class VariableExplorer(QStackedWidget, SpyderPluginMixin):
     sig_option_changed = Signal(str, object)
 
     def __init__(self, parent):
-        QStackedWidget.__init__(self, parent)
+        QWidget.__init__(self, parent)
         SpyderPluginMixin.__init__(self, parent)
+
+        # Widgets
+        self.stack = QStackedWidget(self)
         self.shellwidgets = {}
+
+        # Layout
+        layout = QVBoxLayout()
+        layout.addWidget(self.stack)
+        self.setLayout(layout)
 
         # Initialize plugin
         self.initialize_plugin()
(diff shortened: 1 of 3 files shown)


spyderlib/plugins/variableexplorer.py
```diff
@@ -126,20 +137,20 @@ def remove_shellwidget(self, shellwidget_id):
...
+            self.stack.setCurrentWidget(nsb)
             if self.isvisible:
                 nsb.visibility_changed(True)
             
     def import_data(self, fname):
         """Import data in current namespace"""
-        if self.count():
-            nsb = self.currentWidget()
+        if self.stack.count():
+            nsb = self.stack.currentWidget()
```
ccordoba12: Please use here `self.current_widget()` instead. I think it's easier to read :-)

spyderlib/plugins/variableexplorer.py
```diff
@@ -151,7 +162,7 @@ def visibility_changed(self, enable):
         """DockWidget visibility has changed"""
         SpyderPluginMixin.visibility_changed(self, enable)
         for nsb in list(self.shellwidgets.values()):
-            nsb.visibility_changed(enable and nsb is self.currentWidget())
+            nsb.visibility_changed(enable and nsb is self.stack.currentWidget())
```
ccordoba12: `self.current_widget()` here too please

spyderlib/plugins/variableexplorer.py
```diff
@@ -167,7 +178,7 @@ def get_focus_widget(self):
         Return the widget to give focus to when
         this plugin's dockwidget is raised on top-level
         """
-        return self.currentWidget()
+        return self.stack.currentWidget()
```
ccordoba12: `self.current_widget()`

spyderlib/plugins/variableexplorer.py
```diff
@@ -126,20 +137,20 @@ def remove_shellwidget(self, shellwidget_id):
...
             nsb = self.shellwidgets.pop(shellwidget_id)
-            self.removeWidget(nsb)
+            self.stack.removeWidget(nsb)
             nsb.close()
     
     def set_shellwidget_from_id(self, shellwidget_id):
         if shellwidget_id in self.shellwidgets:
             nsb = self.shellwidgets[shellwidget_id]
-            self.setCurrentWidget(nsb)
+            self.stack.setCurrentWidget(nsb)
```
goanpeca: Should I make accesors for the places where `stack` is called... so that all the calls are made to the variable explorer object and inside I make the correct call?

Like I did with `self.current_widget`?

ccordoba12: It'd be clearer, yes.

spyderlib/widgets/externalshell/namespacebrowser.py
```diff
@@ -437,8 +437,7 @@ def collapse(self):
         """Collapse"""
         self.sig_collapse.emit()
 
-    @Slot(list)
```
ccordoba12: Why did you change this?

goanpeca: It was producing an error... when clicking on the import button... it must have been sleeping for a while

ccordoba12: Please revert it, I'll investigate what's happening with that.

goanpeca: I can explain what happened... the way it was before the `import_data` was being call with a boolean as a first argument instead of the filenames like it should, when I removed the slot, the function was called as it should have

goanpeca: I will put a FIXME with a note as well

ccordoba12: Ok

goanpeca:
@ccordoba12 done..

PR #2592, opened 2015-08-03
Require an equals sign for auto insert parens

### LINKED ISSUE #2581: Autoparens appear when autocompleting in the import section
Autoparens is a great addition, but it should not be applied everywhere, such as whem importing something....

For example if I have this and then press `Tab` in this case (* denotes cursor location)

```python
from spyderlib.qt.QtGui import QToolButton, QToolT*
```

We get

```python
from spyderlib.qt.QtGui import QToolButton, QToolTip()
```

Which is not very helpful....

Any thoughts @blink1073?

blink1073:
Fixes #2581.

@goanpeca, there is no reliable way that I know of to detect whether we are on a continuation line for an import statement.  With this change, there must be an `=` on the same line for auto-insert parens to work.

### DIFF
--- a/spyderlib/widgets/sourcecode/base.py
+++ b/spyderlib/widgets/sourcecode/base.py
@@ -69,7 +69,9 @@ class CompletionWidget(QListWidget):
             self.textedit.insert_completion(completion_list[0])
             try:
                 if self.textedit.close_parentheses_enabled:
-                    if types[0] in ['class', 'function', 'method']:
+                    leading_text = self.textedit.get_text('sol', 'cursor')
+                    if ('=' in leading_text and
+                            types[0] in ['class', 'function', 'method']):
                         self.textedit.handle_close_parentheses('')
             except AttributeError:
                 pass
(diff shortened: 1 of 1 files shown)


goanpeca:
Ok, :-), I-ll take it for now then!

ccordoba12:
But what would happen with something like:

```
def my_function(x, y):
    return x + y

my<Tab>
```

?

We wouldn't have parens auto-inserted, right?

blink1073:
That is correct.  You could always just insert the parens yourself.  The more I think about this, the less I like the idea of adding the parens at all.  I can see it being an annoyance more than a help.

ccordoba12:
We could also revert this feature, or leave it as proposed here.

The bad thing is that from our users perspective there'd be some inconsistency about when to expect parens to be auto-inserted. For example, I'd expect them to be inserted in the example I posted above.

blink1073:
Yes, but like I said, we'll never get it 100% right, and it is really annoying when we display it when the user doesn't want the behavior.

ccordoba12:
So revert? I don't have a problem with that :-) I'd prefer a consistent experience above the minimal effort of typing a single paren.

blink1073:
Closing in favor of #2602.

PR #2602, opened 2015-08-09
Revert auto open parens on completion

### LINKED ISSUE #2581: Autoparens appear when autocompleting in the import section

blink1073:
Since we can't properly tell exactly when it is warranted, prefer consistent behavior and do not automatically insert open parentheses.  Fixes #2581.


ccordoba12:
Agreed, so merging :-)

PR #2611, opened 2015-08-12
PR: Add a shortcut to open the last closed file

### LINKED ISSUE #2415: Add shortcut to open last closed tab
Ok, I'm being a bit spoiled with this feature request and Spyder is already insanely awesome.

It would be nice to have a keyboard shortcut to re-open the last tab(s) you just closed.  This would be like Firefox or Chromium where you just hit CTL+SHIFT+T a couple times when you mistakenly close some tabs.

thewhitetulip:
Fixes #2415 
---

Shortcut is Ctrl+Shift+N.
There is an option in the file menu, but the keyboard shortcut seems a better choice.

The code wasn't changed much.

Submitted via a branch


goanpeca:
Thanks for this @thewhitetulip, will review and let you know of any comments to the code.

Cheers

spyderlib/config/main.py
```diff
@@ -553,6 +553,7 @@ def is_ubuntu():
               'editor/show/hide outline': "Ctrl+Alt+O",
               'editor/show/hide project explorer': "Ctrl+Alt+P",
               'editor/new file': "Ctrl+N",
+              'Editor/Open last closed':"Ctrl+Shift+N",
```
goanpeca: No need for the capital 'O', please change to `open_...`

spyderlib/plugins/editor.py
```diff
@@ -944,7 +952,7 @@ def get_plugin_actions(self):
         self.recent_file_menu = QMenu(_("Open &recent"), self)
         self.recent_file_menu.aboutToShow.connect(self.update_recent_file_menu)
```
goanpeca: Not sure if we want to have this option added to the file menu. I would say no it should not

@ccordoba12, @blink1073?

blink1073: It is a pretty common thing to have.

spyderlib/config/main.py
ccordoba12: I don't think we need a shortcut for this action. Besides, `Ctrl+Shift+<Letter>` are reserved to move to panes.

spyderlib/plugins/editor.py
ccordoba12: I prefer to have it on the Editor context menu, not on the File menu

spyderlib/config/main.py
ccordoba12: Ok, so issue #2415 asked for a new shortcut. So let's use the same one as Firefox: `Ctrl+Shift+T`

spyderlib/plugins/editor.py
goanpeca: Yeah, that makes more sense to me (the context menu of the editor), what does Sublime do haha @blink1073?

thewhitetulip: So should I modify the location?

ccordoba12: Yes, please, and also the shortcut, as I mentioned above, to `Ctrl+Shift+T`

blink1073: Ah, Sublime doesn't have it in the File menu, but Chrome does.

thewhitetulip: Actually I have no idea how it went into the file menu, honestly.

thewhitetulip: Also @ccordoba12 why is the New File option in the file menu but the open last closed item in the Editor menu? I don't get the logic + I am lost in the code, have no idea how to move that menu to the Editor context

thewhitetulip: I would really appreciate some help @blink1073 @goanpeca @ccordoba12

ccordoba12: @thewhitetulip, the logic is this: opening a new file and opening the last closed one are different operations, so they don't belong in the same menu.

You can see that by the way this is handled in Firefox. Its File menu has actions to open a new tab and a new window, but not the last closed tab. That option is only present in its Tab context menu (i.e. when you do a right-click over a tab).

Following that logic, that's why I prefer to move this new action to the Editor context menu, which is defined here:

https://github.com/spyder-ide/spyder/blob/master/spyderlib/widgets/sourcecode/codeeditor.py#L2419

If you prefer, we can do that for you after we merge this PR :-)

spyderlib/plugins/editor.py
```diff
@@ -944,7 +952,7 @@ def get_plugin_actions(self):
         self.recent_file_menu = QMenu(_("Open &recent"), self)
         self.recent_file_menu.aboutToShow.connect(self.update_recent_file_menu)
 
-        file_menu_actions = [self.new_action, self.open_action,
+        file_menu_actions = [self.new_action, self.open_last_closed, self.open_action,
```
ccordoba12: Please remove `self.open_last_closed` from here

spyderlib/plugins/editor.py
```diff
@@ -1779,6 +1787,11 @@ def revert(self):
         editorstack = self.get_current_editorstack()
         editorstack.revert()
     
+    @Slot()
+    def open_last_closed(self):
+		""" Reopens the last closed tab """
+		file_to_open = self.recent_files[-1]
+		self.load(file_to_open)
```
ccordoba12: This block is indented to more than 4 files. Please fix it.

spyderlib/plugins/editor.py
```diff
@@ -626,6 +626,14 @@ def get_plugin_actions(self):
         add_shortcut_to_tooltip(self.new_action, context="Editor",
                                 name="New file")
         
+        self.open_last_closed = create_action(self, _("O&pen last closed"),
+                icon=ima.icon('filenew'), tip=_("Open last closed"),
+                triggered=self.open_last_closed)
+        self.register_shortcut(self.open_last_closed, context="Editor",
+                               name="Open last closed")
+        add_shortcut_to_tooltip(self.open_last_closed, context="Editor",
+                                name="Open last closed")
```
ccordoba12: Please remove the this `add_shortcut_to_tooltip` addition. It's not needed in this case.

spyderlib/plugins/editor.py
```diff
@@ -626,6 +626,14 @@ def get_plugin_actions(self):
         add_shortcut_to_tooltip(self.new_action, context="Editor",
                                 name="New file")
         
+        self.open_last_closed = create_action(self, _("O&pen last closed"),
+                icon=ima.icon('filenew'), tip=_("Open last closed"),
```
ccordoba12: Please remove the  `filenew` icon from this action. Opening the last closed file is not the same as opening a new one, so we use the same icon for both, that'd confuse our users.

ccordoba12:
Ok, final review is ready. @thewhitetulip, after you do my suggested changes this is ready for merging.

Sorry for the delay in coming back to you, we're very busy these days with other things.

spyderlib/plugins/editor.py
thewhitetulip: I made the necessary changes, please go ahead and merge! This is my first code PR to spyder :smile:

thewhitetulip:
@ccordoba12 May I ask status about this PR? I had made the necessary changes a long time ago. Also i have no idea why, the Ctrl+Shift+N shortcut works but Ctrl+shift+T doesn't work

goanpeca:
@ccordoba12 ?

goanpeca:
@ccordoba12 is this something we can merge?

ccordoba12:
@goanpeca, nop, according to @thewhitetulip this is not working for him. If you want to take care of his PR and make it work, that'd be great! It should be an easy fix :-)

goanpeca:
ok

thewhitetulip:
Yes, I believe it should be a small enough fix for you guys :+1: 

The behaviour was actually strange, because if I added cmd+shift+T shortcut, then it mysteriously stopped working :D

ccordoba12:
Closing in favor of #3620.

PR #2617, opened 2015-08-13
Show call tips at right position when there are multiple screens

### LINKED ISSUE #2596: Call tips and auto completion tips go out of screen on a second monitor
I have two monitors and it only happens when the window is in the second monitor.

cy18:
Fixes #2596

There is a similar problem of ipython console's auto-completion droplist. That can be fixed by https://github.com/cy18/ipython/commit/cd9551183d8d025c77a3c09a6ec772270624bb80. But the IPython.qt has been deprecated in the master branch of IPython.

### DIFF
--- a/spyderlib/widgets/calltip.py
+++ b/spyderlib/widgets/calltip.py
@@ -198,7 +198,7 @@ class CallTipWidget(QtGui.QLabel):
 
         vertical = 'bottom'
         horizontal = 'Right'
-        if point.y() + tip_height > screen_rect.height():
+        if point.y() + tip_height > screen_rect.height() + screen_rect.y():
             point_ = text_edit.mapToGlobal(cursor_rect.topRight())
             # If tip is still off screen, check if point is in top or bottom
             # half of screen.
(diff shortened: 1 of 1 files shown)


spyderlib/widgets/calltip.py
```diff
@@ -211,7 +211,7 @@ def show_tip(self, point, tip, wrapped_tiplines):
                     vertical = 'top'
             else:
                 vertical = 'top'
-        if point.x() + tip_width > screen_rect.width():
+        if point.x() + tip_width > screen_rect.width() + screen_rect:
```
ccordoba12: I think you are missing here the x coordinate, like this:

    screen_rect.x()

At least that's what you have in the IPython commit you referenced

ccordoba12:
IPython qtconsole is now called Jupyter qtconsole, and its new repo is located here:

https://github.com/jupyter/qtconsole

So you need to do a PR against that repo now :-)

spyderlib/widgets/calltip.py
ccordoba12: I say it because I'm seeing this traceback in your branch:

```python-traceback
  File "/home/carlos/Projects/spyder/github-repo/spyderlib/widgets/calltip.py", line 214, in show_tip
    if point.x() + tip_width > screen_rect.width() + screen_rect:
TypeError: unsupported operand type(s) for +: 'int' and 'QRect'
```

ccordoba12:
I mean, to fix the related completion drop-list problem

cy18:
Sorry for that stupid mistake.

ccordoba12:
Thanks, and don't worry about it ;-)

PR #2638, opened 2015-08-24
object inspector "list index out of range" with jedi

TWAC:
When trying to inspect
```python
from lxml import etree
etree()
```
I got "list index out of range" in the debug console, and no output in Object inspector.

After this change I get no error and some output in Object inspector.


ccordoba12:
@blink1073, could you take a quick look at this one?

ccordoba12:
It seems to be an issue in master because I can't reproduce it in `2.3`

ccordoba12:
I can't reproduce this problem. Could you give us another example to try?

goanpeca:
@TWAC ? any feedback for this?

TWAC:
I think the jedi downgrade fixed this

PR #2655, opened 2015-08-30
PR: Add support for project config based on user config

goanpeca:


spyderlib/config/projects.py
```diff
@@ -0,0 +1,68 @@
+# -*- coding: utf-8 -*-
+"""
+IMPORTANT NOTES:
+1. If you want to *change* the default value of a current option, you need to
+   do a MINOR update in config version, e.g. from 3.0.0 to 3.1.0
+2. If you want to *remove* options that are no longer needed in our codebase,
+   you need to do a MAJOR update in version, e.g. from 3.0.0 to 4.0.0
+3. You don't need to touch this value if you're just adding a new option
+"""
```
ccordoba12: Please add a blank line here

spyderlib/config/projects.py
```diff
@@ -0,0 +1,68 @@
...
+
+    root_path : 
+
+    filename :
+
+    defaults : 
+
+    load : 
+    
+    version :
```
ccordoba12: Please fill the meaning of all these parameters

spyderlib/config/user.py
```diff
@@ -117,8 +118,27 @@ def _write_file(fname):
                 raise(e)
 
     def filename(self):
+        """Defines the name of the configuration file to use."""
+        self._filename = getattr(self, '_filename', None)
+        self._root_path = getattr(self, '_root_path', None)
```
ccordoba12: I don't like this. Please add `_filename` and `_root_path` in `__init__`, and set them to `None` with a comment above them saying that they are need for projects.

spyderlib/config/projects.py
```diff
@@ -0,0 +1,68 @@
...
+
+        self.set_filename(filename)
+        self.set_root_path(root_path)
+
+        UserConfig.__init__(self, name, defaults=defaults, load=load,
+                            version=version, subfolder=None, backup=False,
+                            raw_mode=True, remove_obsolete=True)
+
+    def set_filename(self, filename):
+        """ """
```
ccordoba12: Missing docstring

spyderlib/config/projects.py
```diff
@@ -0,0 +1,68 @@
...
+        self._filename = filename
+
+    def set_root_path(self, root_path):
+        """ """
```
ccordoba12: Missing docstring

spyderlib/config/projects.py
```diff
@@ -0,0 +1,68 @@
+# -*- coding: utf-8 -*-
+"""
```
ccordoba12: Please add copyright assignment

spyderlib/config/user.py
goanpeca: It does not work like that

goanpeca: I correcto, it WILL not work..

ccordoba12:
Ok, first review done! Thanks for the short PR :-)

spyderlib/config/user.py
ccordoba12: Why not? `UserConfig` inherits from `DefaultsConfig`, and `ProjectsConfig` inherits from `UserConfig`.

So if you add these attrs to `DefaultsConfig`'s `__init__`, they should be there for `ProjectsConfig`, right?

goanpeca: Hmmmm I think I did that and it was not working... I can double check (was a looooooooooooooong time ago)

PR #2656, opened 2015-08-30
PR: Load editor files after initial setup, add project mock methods

goanpeca:
## Description 
This is preliminary work for projects. This one specifically moves the loading of files from the editor from the `setup` method to the `post visible setup`. This has some advantages for the handling of projects, plus it reduces the initial loading time of spyder (since launching and getting the main spyder window)

However since files are loaded now after the window is there, it would be nice to add a spinning icon instead of the editor while files are being loaded (and perhaps update the status bar accordingly). This is not currently handled... but could be added, any suggestions @ccordoba12 ?

## Todo
- [ ] Add a waiting screen while files are loaded
- [ ] Complete docstrings


goanpeca:
@ccordoba12 I know I have to add some docstrings still....

spyderlib/plugins/editor.py
```diff
@@ -492,7 +478,49 @@ def __init__(self, parent, ignore_last_opened_files=False):
             self.add_cursor_position_to_history(filename, position)
         self.update_cursorpos_actions()
         self.set_path()
-        
+
+    def set_open_filenames(self):
+        """ """
```
ccordoba12: Just to remind you: Missing docstring

spyderlib/plugins/editor.py
```diff
@@ -492,7 +478,49 @@ def __init__(self, parent, ignore_last_opened_files=False):
...
+        editorstack = self.editorstacks[0]
+        if self.projectexplorer:
+            if not self.projectexplorer.get_active_project():
+                filenames = []
+                filenames += [finfo.filename for finfo in editorstack.data]
+                self.set_option('filenames', filenames)
+
+    def setup_open_files(self):
+        """ """
```
ccordoba12: Missing docstring

spyderlib/plugins/editor.py
```diff
@@ -492,7 +478,49 @@ def __init__(self, parent, ignore_last_opened_files=False):
...
+        active_project_path = None
+        if self.projectexplorer:
+             active_project_path = self.projectexplorer.get_active_project_path()
+
+        spyder_filenames = self.get_option('filenames', [])
+        project_filenames = self.projectexplorer.get_project_filenames()
+        filenames = spyder_filenames[:]
```
ccordoba12: Why this here? I don't understand why you're trying to load files that belong to the project along with those opened when no project was open.

spyderlib/plugins/editor.py
```diff
@@ -492,7 +478,49 @@ def __init__(self, parent, ignore_last_opened_files=False):
...
+            win_layout = self.get_option('windows_layout_settings', None)
+            if win_layout:
+                for layout_settings in win_layout:
+                    self.editorwindows_to_be_created.append(layout_settings)
+            self.set_last_focus_editorstack(self, self.editorstacks[0])
+        else:
+            self.__load_temp_file()
+
+        for filename in close_filenames:
+            self.close_file_from_name(filename)
```
ccordoba12: So we load some files and close others? I don't get it :-s

spyderlib/plugins/editor.py
```diff
@@ -585,14 +613,19 @@ def closing_plugin(self, cancelable=False):
...
         self.set_option('splitter_state', qbytearray_to_str(state))
-        filenames = []
         editorstack = self.editorstacks[0]
-        filenames += [finfo.filename for finfo in editorstack.data]
+
+        active_project_path = None
+        if self.projectexplorer:
+             active_project_path = self.projectexplorer.get_active_project_path()
+        if not active_project_path:
+	  self.set_open_filenames()
```
ccordoba12: Is this line indented only by one space?

spyderlib/plugins/editor.py
goanpeca: Basically the editor is in charge of opening the files, so depending if there was an active project opened, those files should be open, otherwise open the normal files (stored in the spyder.ini)

Do you have a different idea how it should be done?

spyderlib/plugins/editor.py
```diff
@@ -942,13 +975,22 @@ def get_plugin_actions(self):
...
         self.recent_file_menu.aboutToShow.connect(self.update_recent_file_menu)
 
-        file_menu_actions = [self.new_action, self.open_action,
-                             self.recent_file_menu, self.save_action,
+        file_menu_actions = [self.new_action, 
+                             None,
+                             self.open_action,
+                             self.recent_file_menu,
+                             None,
+                             None,
```
ccordoba12: Why two `None`s here?

goanpeca: You will know in the next PR....

goanpeca: For the moment just trust me on this

spyderlib/plugins/editor.py
goanpeca: Will fix

spyderlib/plugins/editor.py
goanpeca: so.. since the new functionality in Spyder never leaves an empty file I cannot close all files and then open cause there will be an infinite untitled file always... so I just load the project files first and then close the other files...

goanpeca: Otherwise we need to have a switch for that functionality (truning on and off keeping or not a file always in the editor) and refactor the code

spyderlib/plugins/editor.py
ccordoba12: But then why are you doing set differences below? Why is not easier to just open `project_filenames`?

spyderlib/plugins/editor.py
ccordoba12: Yes, let's do that but in code, i.e. not adding an option to our Preferences.

That way we can control when we want to open or not an empty file, and avoid all the workarounds you seem the be taken here :-)

spyderlib/plugins/editor.py
ccordoba12: If this is to add Project actions, I thought we've agreed on adding a new menu entry (before Tools, if I remember correctly).

spyderlib/plugins/projectexplorer.py
```diff
@@ -156,3 +156,16 @@ def restore_scrollbar_position(self):
...
+        return None
+
+    def setup_projects(self):
+        pass
+
+    def get_project_filenames(self):
+        return []
+
+    def get_active_project_path(self):
+        return None
```
ccordoba12: Missing blank line at the end here

spyderlib/spyder.py
```diff
@@ -377,6 +377,9 @@ def __init__(self, options=None):
         self.selectall_action = None
         self.maximize_action = None
         self.fullscreen_action = None
+        self.new_project_action_ = None
+        self.open_project_action_ = None
+        self.close_project_action_ = None
```
ccordoba12: These seem to be here for another PR?

spyderlib/plugins/editor.py
goanpeca: I actually like how it is in RStduio so I was planning on doing it that way... comments :-) ?

spyderlib/spyder.py
goanpeca: Yes, I can take it out... :-)

ccordoba12:
I'm not so sure about the `post_visible_setup` move. Let's see first if that's not necessary by controlling when to open an empty file.

goanpeca:
What I really like about the move is that spyder "aparent" startup time is reduced and becomes more deterministic as it will not depend on the number of files an user had open or not when closing.

I really dislike that spyder takes too long to start just because it is loading files in the background. 

Also we need to improve the UI so that while files are loaded (after spyder has started) there is some meaningfull diplay (something rotating, the status bar can be updated....) etc...

goanpeca:
On a different note, the loading of each file is taking too long, I  think we need to fix something... (or maybe qt5 does it faster)

spyderlib/plugins/editor.py
ccordoba12: I'd also to see a new `Projects` menu entry, so that's really visible to our users.

ccordoba12:
> What I really like about the move is that spyder "aparent" startup time is reduced and becomes more deterministic as it will not depend on the number of files an user had open or not when closing.

Ok

> Also we need to improve the UI so that while files are loaded (after spyder has started) there is some meaningfull diplay (something rotating, the status bar can be updated....)

We could use the same trick that we're using for our IPython consoles.

> On a different note, the loading of each file is taking too long, I think we need to fix something... (or maybe qt5 does it faster)

Yep, I noticed it, and yes, PyQt5 is way much faster. But we need to fix that for PyQt4.

spyderlib/plugins/editor.py
goanpeca: It is not very practical... we already have too many entries...

goanpeca:
Got it!, ok a lot of comments and work to be done so ... later!

ccordoba12:
We **need** to add a new menu entry because:

1. You didn't agree on adding a toolbar menu entry, as the one RStudio has.
2. We need to promote Projects more visibly. Right now we almost don't talk about them because they don't work correctly :-)

goanpeca:
Sure I did, but I was planning on having it on the status bar as a status bar widget, cause I do not know how to add a combobox to the menus area

goanpeca:
I mean, if we add a menu entry the we would be duplicating behaviour in 2 menus, so I can add a combobox as a toolbar (like with the global working directory) or in the status bar, as a status bar widget...

ccordoba12:
The idea would not to duplicate things :-) I mean, add a new menu entry for Projects, and don't add the same entries to the File menu. That's all :-)

In fact, I think we should go with that route at the beginning because it's the simplest thing to do. Then we can see if we also need a toolbar entry.

goanpeca:
Ok!

Issue #2268, opened 2015-03-24
Start testing with pytest/pytest-qt and coverage for Spyder


ccordoba12:
I've never thought of creating a test suite because Spyder has a lot of moving parts, and it's hard to sit down and provide a test for each of them.

However, I've heard a lot lately that Spyder is considered unstable by several people. So I think we should start (at least) by testing that Spyder starts with the git-master versions of all its optional dependencies, to notice and fix startup crashes as quickly as we can.

I also just found this project

https://github.com/pytest-dev/pytest-qt

that seems to provide a good foundation to create a more comprehensive test suite.

Pinging @goanpeca and @blink1073 for discussion ;-)

goanpeca:
Having a lot of moving parts is no excuse for not having a full test suite ;-)!

I will take a look at the package you mention. 

I can make a PR to include the yaml file to start at least basic tests of starting spyder and let it run for some time (a couple of minutes)

ccordoba12:
> Having a lot of moving parts is no excuse for not having a full test suite ;-)!

Yes, I know, but given my limited time too, either I fixed bugs or wrote tests. But now that I have more time, I can consider this possibility ;-)

> I can make a PR to include the yaml file to start at least basic tests of starting spyder and let it run for some time (a couple of minutes)

Yes, please do it ASAP. I think pandas 0.16 broke us again :-(

dhoegh:
:100:,  at least some parts have a decent test by `if __name__ == '__main__':` but they do not get executed often and hence do break without anybody noticing.

Nodd:
A test suite should run every test everytime, otherwise it's not a testsuite but just help to develop faster.

That said it's hard to do for GUIs, slow and boring ;)

SylvainCorlay:
Running the parts that have a `__main__` function in batch as part of the TravisCI script could still be a first step as in checking that things "build".

Nodd:
:+1: Didn't think about that.

goanpeca:
I already had started doing this that @SylvainCorlay  mentions (testing the mains inside the widgets). I can get that working for the weekend.

Nodd:
Actually the main testing platform are nose and pytest (please don't use unittest!).
tox is used to run these tests under different environments and can run either pytest or nosetests.
I don't know about QTest but I guess it can be integrated into both (to be confirmed).

*nosetest* has the advantage of being used by numpy, which provides testing functions for arrays, but it's not really important for spyder itself.
[...]

PR #2663, opened 2015-09-04
PR: Colorize docstrings with a separate color

### LINKED ISSUE #295: syntax highlighting of docstring
_From [mproel...@googlemail.com](https://code.google.com/u/105494983286078857315/) on 2010-08-16T01:24:46Z_

Hi,

the docstrings are highlighted as normal strings in spyderlib. Is this meant to be so? Additionally there could be added a new color for docstrings as they don't fit exactly neither to comment nor to string.

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=295_

rthouvenin:
Fixes #295 

## Description

Creates a new element in syntax highlighters (code: `docstring`, English: Documentation) that can be used to give a different color to code that is documentation (python docstrings, javadocs or other special comments, etc).

Also modifies the PythonSH to match docstrings. Docstrings are strings that are the very beginning of a line (module docstrings) or strings right after a definition (def or class). There are a number of tricky cases, and the code proposed here does not handle all of them correctly, as illustrated below (see if you can find other tricky cases!):

docstrings

I think this is good enough for most cases, but of course may require discussions.
Improving it would I think require to complexify the `PythonSH.highlightBlock()` method even more, and it is already quite hard to read in my opinion. By the way, any reason for not using the [tokenize](https://docs.python.org/2/library/tokenize.html) module of Python? I think it would simplify the parsing, especially of strings, and it was made for that.


goanpeca:
Hi @rthouvenin, thanks for working on this, we are in the middle of a beta2 release so bear with us on the code review of this one. 

I agree that the `PythonSH.highlightBlock()` is a bit complex. I think the reason we do not use tokenize is because it is slower, but we will have to revisit this at some point. The simplest would be to use pygments as we do for all the other syntax highlighters that we do not implement ourselves, but that would mean being slower also...

rthouvenin:
No worries, I agree that this is not urgent. And anyway now I have it working on my local Spyder :)

Nodd:
> Docstrings are strings that are the very beginning of a line (module docstrings)

You mean "file" rather that "line" here, right ? Also there could be comments before (like the shebang and the encoding).

rthouvenin:
@Nodd Yes you're right, I was describing the way I implemented it. "at the very beginning of the line" (so no white space before) is an approximation of module docstring that works because other docstrings are necessarily indented (in a function or class). This is why in my example file, the string at the end of the file is colorized as a docstring.
Enforcing beginning of a file would have been trickier because the code is parsed line by line.

Nodd:
The feature would be a nice addition I think, but I'd say that your current implementation is too limited to be included in spyder, too many corner cases fail and we would have lots of complaints from users. So if you could put some more work into this, please do ! (or maybe someone else will take over ?)

rthouvenin:
I'd be OK to try to give it some more time, but as mentioned earlier this would be a major change. Doing it without refactoring the parsing and colorizing code would make it become unreasonably complex and hard to maintain. Also, Gonzalo thinks this could bring performance issues.
Is there any regression test suite and/or profiling set-up that can be used to assert nothing is broken?

Nodd:
About a test suite, see this discussion: https://github.com/spyder-ide/spyder/issues/2268. In short no, there is no test suite. Travis only checks if spyder launches then closes successfully.
Some files have a `if __name__ == "__main__":` block so that manual tests are easier to run.

goanpeca:
What should we do about this PR?

ccordoba12:
@goanpeca, I think pyqode has a good implementation to colorize docstrings different from regular strings. We should borrow that and close this PR instead :-)

PR #2665, opened 2015-09-05
first pull_request and i did it backwards from my anaconda files

skidzo:
I had to change the file spyderlib/start_app.py, because the import: 

    from spyderlib.config import <something>

does not work, and I could not figure out why... maybee because python finds a folder here called config/ that does not have the same purpose... so, config is the wrong name.

instead of renaming i moved everything inside config.py to start_app, because there was the first import...

then i changed all the other imports until spyder started again.

i have anaconda3 and newest PyQt5 AND PyQt4

THANKS!!! this took so long, but worth the wait!


ccordoba12:
I know you're well intended, but we simply can't accept your contribution because it seems you touch every file in our source code by changing their line endings. Besides you modified all our png's (I don't know why :-).

As I explained you in my (personal) email, the `master` branch of Spyder is working correctly with PyQt5.

I guess you installed some third-party PyQt5 packages in Anaconda (because there are not officlal ones). So to have Spyder `master` in that setup, you need to run

1. `conda remove spyder` (to uninstall the official packages).
2. `git clone https://github.com/spyder-ide/spyder.git` (to have our latest version)
3. `cd spyder`
4. `python bootstrasp.py --gui=pyqt5` (to start Spyder from your git clone, without installing it).
5. `python setup.py install` (in case you want to install it).

That's all :-)

skidzo:
Tanks again! :-)

PR #2669, opened 2015-09-05
PR: Use pygments in introspection

blink1073:
Uses the Pygments lexer in introspection instead of the previous regex-based approach.  This will help completions in non-python-like files.  We will do a better job of finding the right object to complete, and finding suitable objects within the same syntax type for completion.

### DIFF
--- a/spyderlib/utils/introspection/fallback_plugin.py
+++ b/spyderlib/utils/introspection/fallback_plugin.py
@@ -15,12 +15,14 @@ import os.path as osp
 import re
 import time
 
+from pygments.token import Token
+
 from spyderlib.utils.debug import log_dt
 from spyderlib.utils import sourcecode, encoding
 from spyderlib.utils.introspection.manager import (
     DEBUG_EDITOR, LOG_FILENAME, IntrospectionPlugin)
 from spyderlib.utils.introspection.utils import (
-    get_parent_until, memoize)
+    get_parent_until, memoize, find_lexer_for_filename, get_keywords)
 
 
 class FallbackPlugin(IntrospectionPlugin):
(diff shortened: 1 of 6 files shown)


ccordoba12:
This needs a rebase.

blink1073:
Done.

ccordoba12:
This one has failing tests and requires a rebase again :-)

blink1073:
All set.

spyderlib/utils/introspection/plugin_manager.py
```diff
@@ -23,6 +23,17 @@
...
 
+from spyderlib.utils.syntaxhighlighters import (
+    custom_extension_lexer_mapping
+)
+
+from pygments.lexers import (
+    get_lexer_for_filename, get_lexer_by_name, TextLexer
+)
+from pygments.util import ClassNotFound
+from pygments. token import Token
```
ccordoba12: Here there is an additional blank in `pygments. token`

spyderlib/utils/introspection/fallback_plugin.py
```diff
@@ -44,17 +46,20 @@ def get_completions(self, info):
...
-            tokens = set(info.split_words(-1))
-            items = [item for item in tokens if
-                     item.startswith(base) and len(item) > len(base)]
-            if '.' in base:
-                start = base.rfind('.') + 1
-            else:
-                start = 0
-
-            items = [i[start:len(base)] + i[len(base):].split('.')[0]
-                     for i in items]
```
ccordoba12: I think we should leave `word` completions because it's useful for prose files like Markdown and Rest.

ccordoba12:
I'm seeing a constant error in this branch after pressing Tab in the internal console:

```python-traceback
  File "/home/carlos/Projects/spyder/github-repo/spyderlib/utils/introspection/plugin_manager.py", line 417, in _introspection_complete
    if info == self.pending:
  File "/home/carlos/Projects/spyder/github-repo/spyderlib/utils/introspection/plugin_manager.py", line 257, in __eq__
    return (self.filename == other.filename
AttributeError: 'NoneType' object has no attribute 'filename'
```

It appears this error is preventing to get completions (at least on my simple test with a C++ file :-)

ccordoba12:
I forgot to mention that error is happening in Python 3

ccordoba12:
It also happens in Python 2

blink1073:
All set!

blink1073:
It looks like something is wrong with the scikit-image rackspace server...

ccordoba12:
@blink1073, please rebase in top of latest `master`. I changed a little bit how our integration with Travis works after your latest rebase, and that's not letting our tests to finish in your branch :-)

blink1073:
Fixed, appveyor failure is unrelated.

ccordoba12:
Yep, I'm investigating the cause of that :-)

ccordoba12:
@blink1073, could you rebase again so that we can retest this PR with the changes I pushed yesterday to Travis and AppVeyor?

Thanks :-)

blink1073:
Rebased, but now there is an unrelated Travis error, good times...

ccordoba12:
Now it passes :-)

ccordoba12:
This seems to not be working correctly on my tests. For example, I'm using

https://github.com/spyder-ide/spyder/blob/master/continuous_integration/travis/build_test.sh

and my findings are:

1. i can get completions for some of the tokens present in the file, like

        con<Tab>

    which gives

        conda
        continuous_integration/conda-recipes

2. I cannot get completions for other tokens, for example

        el<Tab>

    doesn't return anything, but there are several `else`'s present in that file.

3. I can't manage to get completions for simple bash constructs, like

        fo<Tab>

    which should give `for` at least :-)

----

@blink1073, could you confirm my findings and see how to fix them?

spyderlib/utils/introspection/fallback_plugin.py
```diff
@@ -44,17 +46,34 @@ def get_completions(self, info):
...
+                tokens = info.lexer.get_tokens(info.source_code)
+                for (context, token) in tokens:
+                    token = token.strip()
+                    if (context in info.context and token.startswith(info.obj)
+                            and info.obj != token):
+                        items.append(token)
+                # add in keywords if not in a string
+                if context not in Token.Literal.String:
+                    try:
+                        keywords = list(info.lexer.tokens['keywords'][0][0].words)
```
ccordoba12: This is not working for some lexers I tried. For example, `BashLexer`, `JavaLexer` and `JavascriptLexer` don't have a `keywords` key as part of their tokens.

By the way, I'm using Pygments 2.0.2. Maybe the 1.x is different in that respect :-)

blink1073:
Now we fall back to regex-based if the lexer does not have keywords.

ccordoba12:
@blink1073, what lexers have `keywords` as part of their tokens? I reviewed several of them and I couldn't find none :-)

blink1073:
Note to self, add the following to `CodeInfo` to get the keywords:

```python
from pygments import lexers
import re


def get_keywords(lex):
    if 'keywords' in lex.tokens:
        return lex.tokens['keywords'][0][0].words
    keywords = []
    for vals in lex.tokens.values():
        for val in vals:
            try:
                if '|' in val[0] and ')\\b' in val[0]:
                    val = re.sub(r'\\.', '', val[0])
                    val = re.sub('[^0-9a-zA-Z|]+', '', val)
                    keywords.extend(val.split('|'))
            except Exception:
                continue
    return keywords


print(get_keywords(lexers.PythonLexer()))
print(get_keywords(lexers.BashLexer()))
print(get_keywords(lexers.JavaLexer()))
```

ccordoba12:
That's a great start! But it can be improved further :-) Just check (for example) the completions you're missing in `BashLexer` (I mean bash builtins, which are grabbed by your code :-)

blink1073:
Implemented, you can now complete `functi` in a bash script.

blink1073:
I added a test for bash completion, this should be ready.

ccordoba12:
@blink1073, this one needs a rebase now :-)

blink1073:
Updated!

ccordoba12:
Sorry, it's again in conflict (I think because of PR #3032 ;-)

blink1073:
Fixed.

ccordoba12:
Thanks @blink1073! Great addition, as always! :-)

goanpeca:
:+1:

PR #2670, opened 2015-09-06
PR: Add option to toggle Emacs keybindings

### LINKED ISSUE #669: Provide Vim and Emacs key bindings for the editor
_From [r.c.bruno.andre@gmail.com](https://code.google.com/u/r.c.bruno.andre@gmail.com/) on 2011-05-27T12:35:31Z_

Could it be possible to use vim as the text editor in Spyder?

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=669_

blink1073:
The keyboard shortcuts themselves are open for discussion.  I did my best to match this [table](http://wttools.sourceforge.net/emacs-stuff/emacs-keybindings.html).

CAM-Gerlach Edit: Fixes #669 

screen shot 2015-09-07 at 10 27 46 am

screen shot 2015-09-07 at 10 27 54 am


goanpeca:
@Nodd

Nodd:
Are the emacs keybindings read-only ? In this case, checking the button should disable the edition. Also the checkbox should be on the left side, with "Emacs Keybindings as its label.

If it's not, I'd prefer a pushbutton with a label like "Load emacs-like keybondings". If we add more layouts later, the pushbutton could be a QListWidget. I wouldn't expect a checkbox here.

ccordoba12:
I'd like to have a combobox instead of a checkbox, so that we can add also IPython and Vim keybindings later on :-)

blink1073:
Updated.  The old reset behavior is now achieved by explicitly loading the defaults.

goanpeca:
@blink1073 what do you mean by
>  I did my best to match this

Some are not matching?

Also we need some way to handle custom shortcuts? maybe IO load the defaults and change a couple and if for some reason I  load VIM bindings then I lost the changes I made to the spyder defaults... 

Otherwise I think it is pretty cool :-)

blink1073:
@goanpeca, I mean, Emacs uses different terminology and I did my best to translate (I only used Emacs for like 3 months a few years ago).

Yeah, it would be nice to store a set of custom shortcuts, but that is outside the scope of this PR.  :wink:

goanpeca:
Ahhh ok got it :-) 

Sure, not for this PR, but it is good that we agree that it might be useful, cheers

ccordoba12:
Thanks a lot @blink1073!! This is looking really good!

I hope to merge it soon :-)

spyderlib/config/main.py
```diff
@@ -580,6 +580,53 @@ def is_ubuntu():
...
+              'editor/redo': 'Ctrl+X, Ctrl+Y',
+              'editor/cut': 'Ctrl+W',
+              'editor/copy': 'Ctrl+C',
+              'editor/paste': 'Ctrl+Y',
+              'editor/delete': 'Delete',
+              'editor/select all': "Ctrl+X+H",
+              # -- In plugins/editor
+              'editor/new file': "Ctrl+X, Ctrl+N",
+              'editor/open file': "Ctrl+X, Ctrl+F",
+              'editor/save file': "Ctrl+X, Ctrl+W",
```
ccordoba12: This one is wrong, it should be `Ctrl+X, Ctrl+S`

spyderlib/config/main.py
```diff
@@ -580,6 +580,53 @@ def is_ubuntu():
...
+              'editor/save all': "Ctrl+Alt+S",
+              'editor/save as': 'Ctrl+Shift+S',
```
ccordoba12: And this one is `Ctrl+X, Ctrl+W`

spyderlib/config/main.py
```diff
@@ -580,6 +580,53 @@ def is_ubuntu():
...
+              'editor/go to line': 'Ctrl+X, Ctrl+L',
+              'editor/go to previous file': 'Ctrl+Tab',
+              'editor/go to next file': 'Ctrl+Shift+Tab',
+              '_/file switcher': 'Meta+Shift+P',
+              # -- In spyder.py
+              'editor/find text': "Ctrl+S",
+              'editor/find next': "F3",
+              'editor/find previous': "Ctrl+R",
+              'editor/replace text': "Ctrl+X, Ctrl+H",
+              'editor/undo': 'Meta+Backspace',
```
ccordoba12: This is `Ctrl+_`

spyderlib/config/main.py
```diff
@@ -580,6 +580,53 @@ def is_ubuntu():
...
+              'editor/end of document': 'Meta+Right',
+              # -- In widgets/editor
+              'editor/inspect current object': 'Ctrl+X, Ctrl+I',
+              'editor/go to line': 'Ctrl+X, Ctrl+L',
+              'editor/go to previous file': 'Ctrl+Tab',
+              'editor/go to next file': 'Ctrl+Shift+Tab',
+              '_/file switcher': 'Meta+Shift+P',
+              # -- In spyder.py
+              'editor/find text': "Ctrl+S",
+              'editor/find next': "F3",
```
ccordoba12: This one is also `Ctrl+S`. I don't know what to do about it :-)

spyderlib/config/main.py
```diff
@@ -580,6 +580,53 @@ def is_ubuntu():
               # ---- Profiler (in p_profiler) ----
               'profiler/run profiler': "F10"
               }),
+            ('emacs_shortcuts',
+              {
+              # -- In codeeditor
+              'editor/code completion': 'Meta+Tab',
```
ccordoba12: I think we should leave `Ctrl+Space` here because Emacs doesn't have the notion of completion.

spyderlib/config/main.py
```diff
@@ -580,6 +580,53 @@ def is_ubuntu():
...
+              # -- In codeeditor
+              'editor/code completion': 'Meta+Tab',
+              'editor/go to definition': "Ctrl+G",
+              'editor/start of line': "Ctrl+A",
+              'editor/end of line': "Ctrl+E",
+              'editor/previous line': "Ctrl+P",
+              'editor/next line': "Ctrl+N",
+              'editor/previous char': "Ctrl+B",
+              'editor/next char': "Ctrl+F",
+              'editor/previous word': "Meta+B",
```
ccordoba12: This one is `Alt+B`

spyderlib/config/main.py
```diff
@@ -580,6 +580,53 @@ def is_ubuntu():
...
+              'editor/next word': "Meta+F",
```
ccordoba12: And this one is `Alf+F`

spyderlib/config/main.py
```diff
@@ -580,6 +580,53 @@ def is_ubuntu():
...
+              'editor/kill to line end': "Ctrl+K",
+              'editor/kill to line start': "Ctrl+U",
+              'editor/yank': 'Meta+Y',
```
ccordoba12: `yank` is equivalent to `paste`, right?

spyderlib/config/main.py
```diff
@@ -580,6 +580,53 @@ def is_ubuntu():
...
+              'editor/rotate kill ring': 'Shift+Ctrl+Y',
+              'editor/kill previous word': 'Ctrl+Backspace',
+              'editor/kill next word': 'Ctrl+Delete',
+              'editor/start of document': 'Meta+Left',
```
ccordoba12: This is `Alt+<`

spyderlib/config/main.py
```diff
@@ -580,6 +580,53 @@ def is_ubuntu():
...
+              'editor/end of document': 'Meta+Right',
```
ccordoba12: This is `Alt+>`

spyderlib/config/main.py
```diff
@@ -580,6 +580,53 @@ def is_ubuntu():
...
+              # -- In widgets/editor
+              'editor/inspect current object': 'Ctrl+X, Ctrl+I',
+              'editor/go to line': 'Ctrl+X, Ctrl+L',
```
ccordoba12: This is `Alt+G, Alt+G`

spyderlib/config/main.py
```diff
@@ -580,6 +580,53 @@ def is_ubuntu():
...
+              'editor/go to previous file': 'Ctrl+Tab',
+              'editor/go to next file': 'Ctrl+Shift+Tab',
+              '_/file switcher': 'Meta+Shift+P',
```
ccordoba12: This is `Ctrl+X, Ctrl+B`

spyderlib/config/main.py
```diff
@@ -580,6 +580,53 @@ def is_ubuntu():
...
+              # -- In spyder.py
+              'editor/find text': "Ctrl+S",
+              'editor/find next': "F3",
+              'editor/find previous': "Ctrl+R",
+              'editor/replace text': "Ctrl+X, Ctrl+H",
+              'editor/undo': 'Meta+Backspace',
+              'editor/redo': 'Ctrl+X, Ctrl+Y',
```
ccordoba12: Emacs doesn't have redos ;-)

spyderlib/config/main.py
```diff
@@ -580,6 +580,53 @@ def is_ubuntu():
...
+              'editor/cut': 'Ctrl+W',
+              'editor/copy': 'Ctrl+C',
```
ccordoba12: This is `Alt+W`

spyderlib/config/main.py
```diff
@@ -580,6 +580,53 @@ def is_ubuntu():
...
+              'editor/paste': 'Ctrl+Y',
+              'editor/delete': 'Delete',
+              'editor/select all': "Ctrl+X+H",
+              # -- In plugins/editor
+              'editor/new file': "Ctrl+X, Ctrl+N",
```
ccordoba12: Let's leave this one in `Ctrl+N` because Emacs doesn't have a shortcut for new files

spyderlib/config/main.py
```diff
@@ -580,6 +580,53 @@ def is_ubuntu():
...
+              'editor/open file': "Ctrl+X, Ctrl+F",
+              'editor/save file': "Ctrl+X, Ctrl+W",
+              'editor/save all': "Ctrl+Alt+S",
```
ccordoba12: This is `Ctrl+X+S`

spyderlib/plugins/shortcuts.py
```diff
@@ -807,6 +816,22 @@ def apply_settings(self, options):
         self.table.save_shortcuts()
         self.main.apply_shortcuts()
 
+    def load_button_clicked(self, change):
+        selected = self.combo_keybindings.currentText().lower()
+        reset_shortcuts()
```
ccordoba12: Do we really need to do this? What would happen with user defined shortcuts in this case?

goanpeca: They disappear :-p

spyderlib/config/main.py
ccordoba12: Sorry, it is `Alt+G+G`

spyderlib/plugins/shortcuts.py
ccordoba12: Bad thing! We need to find a better solution :-)

goanpeca: Yes that is what I told @blink1073, as soon as the user changes the defaults, a `custom` option should appear in the combobox and those shortcuts should be saved for later reuse...

ccordoba12: Ok, let's leave this for another PR, as @blink1073 mentioned.

However, I don't know if a `custom` option is enough. What if you want the Emacs shortcuts but do some customizations on top of them? And also for the Default ones?

goanpeca: Hmm we would be entering a rabbit hole here...

ccordoba12: I think we should preserve customizations for any scheme, and leave the button called `Reset to default schemes` :-)

goanpeca: Hmmm sure why not :-9, but for another PR :p

spyderlib/config/main.py
blink1073: Yes, but it also affects the Kill Ring.

spyderlib/config/main.py
blink1073: But `Ctrl+N` is go to next line.

blink1073:
Remapped the emacs keybindings.

spyderlib/config/main.py
```diff
@@ -580,6 +580,51 @@ def is_ubuntu():
...
+              'editor/end of line': "Ctrl+E",
+              'editor/previous line': "Ctrl+P",
+              'editor/next line': "Ctrl+N",
+              'editor/previous char': "Ctrl+B",
+              'editor/next char': "Ctrl+F",
+              'editor/previous word': "Alt+B",
+              'editor/next word': "Alt+F",
+              'editor/kill to line end': "Ctrl+K",
+              'editor/kill to line start': "Ctrl+U",
+              'editor/yank': 'Meta+Y',
```
ccordoba12: This one is `Alt+Y`.

spyderlib/config/main.py
```diff
@@ -580,6 +580,51 @@ def is_ubuntu():
...
+              'editor/rotate kill ring': 'Shift+Ctrl+Y',
+              'editor/kill previous word': 'Ctrl+Backspace',
+              'editor/kill next word': 'Ctrl+Delete',
+              'editor/start of document': 'Alt+Left',
```
ccordoba12: This is `Alt+<`, not `Alt+Left` (arrow)

spyderlib/config/main.py
```diff
@@ -580,6 +580,51 @@ def is_ubuntu():
...
+              'editor/end of document': 'Alt+Right',
```
ccordoba12: And this is `Alt+>`

spyderlib/config/main.py
```diff
@@ -580,6 +580,51 @@ def is_ubuntu():
...
+              # -- In widgets/editor
+              'editor/inspect current object': 'Ctrl+X, Ctrl+I',
+              'editor/go to line': 'Alt+G+G',
```
ccordoba12: This is activated by pressing `Alt+G`, so we can remove the second `G` :-)

spyderlib/config/main.py
```diff
@@ -580,6 +580,51 @@ def is_ubuntu():
...
+              'editor/go to previous file': 'Ctrl+Tab',
+              'editor/go to next file': 'Ctrl+Shift+Tab',
+              '_/file switcher': 'Ctrl+X, Ctrl+B',
+              # -- In spyder.py
+              'editor/find text': "Ctrl+S",
+              'editor/find next': "F3",
+              'editor/find previous': "Ctrl+R",
+              'editor/replace text': "Ctrl+X, Ctrl+H",
+              'editor/undo': 'Ctrl+_',
+              'editor/cut': 'Ctrl+W',
```
ccordoba12: This one is not working, it's closing all files (I don't know why :-s)

spyderlib/config/main.py
```diff
@@ -580,6 +580,51 @@ def is_ubuntu():
...
+              'editor/copy': 'Alt+W',
```
ccordoba12: This one is not working either. It just writes a `w` over the selected text, instead of copying it.

spyderlib/config/main.py
```diff
@@ -580,6 +580,51 @@ def is_ubuntu():
...
+              'editor/paste': 'Ctrl+Y',
```
ccordoba12: Not working either. I think the problem is these shortcuts are also defined in the Editor context menu, and so are causing conflicts with the shortcuts present in Preferences.

spyderlib/config/main.py
```diff
@@ -580,6 +580,51 @@ def is_ubuntu():
...
+              'editor/delete': 'Delete',
+              'editor/select all': "Ctrl+X+H",
```
ccordoba12: Not working.

spyderlib/config/main.py
```diff
@@ -580,6 +580,51 @@ def is_ubuntu():
...
+              # -- In plugins/editor
+              'editor/new file': "Ctrl+X, Ctrl+N",
+              'editor/open file': "Ctrl+X, Ctrl+F",
+              'editor/save file': "Ctrl+X, Ctrl+S",
```
ccordoba12: These double Ctrl shortcuts are also not working.

ccordoba12:
A couple of comments:

1. There is a conflict with `Ctrl+Y`: It's assigned to both `redo` and `paste`. That's reported when you open Preferences.
2. None of the `Alt` shortcuts work unless you remove the `&` from all our main menus, for example `&File` (present in `spyder.py`). Although it's not required, I think it's better to remove `&` from submenus too.
3. After doing 2., i found that `Alt+F` and `Alt+B` (moving to next/prev word) are not working as expected. They fail to move among words when pressed repeatedly, i.e. they move only once to the next/prev word.
4. The UI in *Keyboard Shortcuts* should tell users what is the scheme they have currently selected.

----

It seems there's still some work to do to have this working correctly :-)

blink1073:
The more I think about it, the more I think it is a mistake to offer this "mode".  It is too invasive and difficult to get right.  We shouldn't have to change the way the whole application works to support `Alt` keybindings, for instance. 

For example, "The UI in Keyboard Shortcuts should tell users what is the scheme they have currently selected."  This would no longer be strictly true once the user has customized the shortcuts.  

My preference would be to close this PR and leave everything as-is, where we support all of the required keybindings, and leave it up to the user to customize.

Thoughts @spyder-ide/core-developers?

goanpeca:
@blink1073 do not despair (yet!). I think that letting the user know what bindings are on is beyond the point and can be left for another PR.

I will not use these keybidnings but if this will help more people use it I am willing to make the needed changes so that alt can work.

blink1073:
I pushed emacs/redo to `Ctrl+Alt+Y`, since technically there is no redo in emacs.  Also, I fixed the prev/next word behavior.

Nodd:
We could have multiple keybinds set, similar to how syntax highlighting works. There would be one tab for each set, with "Spyder", "Emacs", "Vim", "Custom" entries. Each set would be editable by the user (with a possibility to reset to defaults). The user the decides what set is active.
That may be too complicated to manage, but it's at elast coherent with syntax highlighting.

goanpeca:
@Nodd I think we are moving away from that design in syntax highlighting, cause the defaults themes are already fixed. So the ides there was to have a single custom page where you could start from any of the shipped presets and then modify as you wished. Perhaps we could even add the save as a new custom, so you could choose among several customs in a single page.

But for shortcuts this page ideas.. might make sense.. but not sure :-s

blink1073:
I am :-1: on multiple tabs.  I think it is confusing for syntax highlighting, and would be confusing here.

ccordoba12:
I clearly understand what @Nodd means (always the voice of reason! ;-). However, as I see it, it only requires requires minor modifications to our current UI for Keyboard shortcuts.

I'll do a PR against @blink1073 PR to prove my point :-)

Nodd:
Actually I'm not sure either if tabs are the correct solution.
Having different defaults to load (erasing all customization in the process) seems good enough and simpler for the ui, the user, the configuration file...

Looks like my voice of reason is a bit schizophrenic ! :wink:

goanpeca:
Why not add a button to save keybindings that will pop out a dialog to put the name, and save it somewhere either in the ini or a separate set of files.

Also the user could export keybindings maybe?

blink1073:
Both good points.

blink1073:
Closing this one, as I don't have the bandwidth to work on it.

ccordoba12:
@blink1073, please leave this one open. We know you can't work on this anymore, but we'll rebase your work and finish it for Spyder 3.3 ;-)

jnsebgosselin:
http://doc.qt.io/qtcreator/creator-keyboard-shortcuts.html#emacs-shortcuts

CAM-Gerlach:
@ccordoba12 Is this ever likely to get merged in its current form, or can it be closed for now?

blink1073:
Closing this, since it is too far out of conflict.  New work would have to start from scratch.

CAM-Gerlach:
Vim users rejoice :laughing: 

In all seriousness, its indeed unfortunate. However, hopefully in future versions we can provide better support for keyboard shortcut profiles...its something we've wanted to do for a long time, but we've always had more pressing priorities.

Nodd:
Meanwhile vim users : https://github.com/spyder-ide/spyder-vim/issues/75 😭

PR #2673, opened 2015-09-06
Fix single item completion

### LINKED ISSUE #2672: Autocomplete does not insert correct word when requested with Ctrl+Space
To reproduce: in a python editor, start typing something and when the word is long enough so that there should be only one auto-complete suggestion, invoke it with <kbd>Ctrl</kbd>+<kbd>Space</kbd>. For example, type `enume` and invoke the auto-completion. Because there is only one suggestion, the word is automatically completed. However the auto-complete menu opens anyway, and if I press <kbd>Enter</kbd> to validate and close it, the word is replaced with one where the beginning of the word is repeated. For example, in the case of `enumerate`, it is replaced with `enumenumerate`.

This does NOT happen when there is more than one auto-complete suggestion or when the auto-complete is automatically triggered when typing a dot, even if there is a single suggestion. In both these cases, the word is not automatically completed and pressing <kbd>Enter</kbd> correctly finishes the word.

Also, the problem occurs in latest Spyder 3.0.0dev, but not in the stable release that I have (2.2.5). In 2.2.5, the auto-complete menu does not open and the word is automatically completed to the single suggestion, which is I think the correct behavior (this is also what Eclipse does if I remember correctly). So it seems to be a regression. Will happily help fix it if given pointers.

blink1073:
Fixes #2672

### DIFF
--- a/spyderlib/widgets/sourcecode/base.py
+++ b/spyderlib/widgets/sourcecode/base.py
@@ -31,6 +31,7 @@ from spyderlib.widgets.sourcecode.terminal import ANSIEscapeCodeHandler
 from spyderlib.widgets.mixins import BaseEditMixin
 from spyderlib.widgets.calltip import CallTipWidget
 from spyderlib.py3compat import to_text_string, str_lower, PY3
+from spyderlib.config.base import debug_print
 
 
 def insert_text_to(cursor, text, fmt):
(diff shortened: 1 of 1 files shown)


spyderlib/widgets/sourcecode/base.py
```diff
@@ -31,6 +31,7 @@
 from spyderlib.widgets.mixins import BaseEditMixin
 from spyderlib.widgets.calltip import CallTipWidget
 from spyderlib.py3compat import to_text_string, str_lower, PY3
+from spyderlib.config.base import debug_print
```
ccordoba12: Left over?

blink1073: Yep, fixed.

ccordoba12:
Thanks @blink1073 :-)

PR #2691, opened 2015-09-13
Fixed PyQt5 detection without QT_API env var

PierreRaybaut:
When QT_API environment variable has not been set properly, which is safe to assume when using PyQt5 for the first time, importing the `spyderlib.qt` package will fail. This change allows PyQt5 to be imported and used if QT_API has not been set and if both PyQt4 and PySide packages are not installed.

### DIFF
--- a/spyderlib/qt/__init__.py
+++ b/spyderlib/qt/__init__.py
@@ -15,17 +15,7 @@ assert os.environ['QT_API'] in ('pyqt5', 'pyqt', 'pyside')
 API = os.environ['QT_API']
 API_NAME = {'pyqt5': 'PyQt5', 'pyqt': 'PyQt4', 'pyside': 'PySide'}[API]
 
-PYQT5 = False
-
-if API == 'pyqt5':
-    try:
-        from PyQt5.QtCore import PYQT_VERSION_STR as __version__
-        is_old_pyqt = False
-        is_pyqt46 = False
-        PYQT5 = True
-    except ImportError:
-        pass
-elif API == 'pyqt':
+if API == 'pyqt':
     # Spyder 2.3 is compatible with both #1 and #2 PyQt API,
     # but to avoid issues with IPython and other Qt plugins
     # we choose to support only API #2 for 2.4+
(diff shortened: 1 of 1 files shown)


spyderlib/qt/__init__.py
```diff
@@ -57,6 +54,17 @@
             API_NAME += (" (API v%d)" % sip.getapi('QString'))
         except AttributeError:
             pass
+        from PyQt4 import uic  # analysis:ignore
```
ccordoba12: Why is this necessary now?

spyderlib/qt/__init__.py
```diff
@@ -57,6 +54,17 @@
...
+
+PYQT5 = False
+if API == 'pyqt5':
+    try:
+        from PyQt5.QtCore import PYQT_VERSION_STR as __version__
+        from PyQt5 import uic  # analysis:ignore
```
ccordoba12: Also this `import`? Do you need it for other things?

ccordoba12:
Thanks a lot @PierreRaybaut. Everything seems fine except for the two comments I made :-)

spyderlib/qt/__init__.py
PierreRaybaut: The `spyderlib.qt` package was wrote not only for Spyder but also for other libraries (`guidata`, `winpython` and now `python-qwt`). Adding the line `from PyQt4 import uic` here will simply allow to import the `uic` module with `from spyderlib.qt import uic`, that's all. FYI, the `uic` module is used to run GUI built with the help of QtDesigner (see [this example](https://github.com/PierreRaybaut/guiqwt/blob/master/guiqwt/qtdesigner.py)).

spyderlib/qt/__init__.py
PierreRaybaut: Same explanation as above.

spyderlib/qt/__init__.py
ccordoba12: Ok, no problem. I just wanted to know why you decided to add it given that we don't use QtDesigner in Spyder.

By the way, we're planning to move to use [qtpy](https://github.com/spyder-ide/qtpy) (one of the Spyder org projects) as our Qt shim.

I think that would help your projects too, to avoid copy/pasting the Spyder shim around, and to have everything consolidated in one place.

ccordoba12:
@PierreRaybaut, just a simple recommendation for next time: please use your own fork of Spyder to create pull requests, instead of creating branches in the main repo.

The thing is those branches are propagated to all people that pull regularly from Github (like our other core developers :-)

Of course, you're more than welcome to submit directly to `master` if you prefer to do so :-)

PierreRaybaut:
@ccordoba12 Sorry for the inconvenience, I'm such a newbie with `git` and GitHub. I'll remember it for next time. Keep up the good work.

ccordoba12:
No problem :-) Glad to have you back!

PR #2692, opened 2015-09-14
Add tab completions for PathCombobox

### LINKED ISSUE #2677: Autocomplete for working directory widget
As well as changing colour to show when the input is a valid directory, it would be nice if the working directory input also had autocomplete to speed navigating to directory.

goanpeca:
Fixes #2677

## Description
The PathComboBox now has auto completion which improves the UX. This affects the Global Working Directory Toolbar and the Find in Files pane.

## Screenshot
image

### DIFF
--- a/spyderlib/plugins/inspector.py
+++ b/spyderlib/plugins/inspector.py
@@ -64,7 +64,7 @@ class ObjectComboBox(EditableComboBox):
     QComboBox handling object names
     """
     # Signals
-    valid = Signal(bool)
+    valid = Signal(bool, bool)
 
     def __init__(self, parent):
         EditableComboBox.__init__(self, parent)
@@ -107,15 +107,15 @@ class ObjectComboBox(EditableComboBox):
             if editing:
                 # Combo box text is being modified: invalidate the entry
                 self.show_tip(self.tips[valid])
-                self.valid.emit(False)
+                self.valid.emit(False, False)
             else:
                 # A new item has just been selected
                 if valid:
                     self.selected()
                 else:
-                    self.valid.emit(False)
-        else:
-            self.set_default_style()
+                    self.valid.emit(False, False)
+#        else:
+#            self.set_default_style()
 
 
 class ObjectInspectorConfigPage(PluginConfigPage):
--- a/spyderlib/plugins/inspector.py
+++ b/spyderlib/plugins/inspector.py
@@ -114,8 +114,6 @@ class ObjectComboBox(EditableComboBox):
                     self.selected()
                 else:
                     self.valid.emit(False, False)
-#        else:
-#            self.set_default_style()
 
 
 class ObjectInspectorConfigPage(PluginConfigPage):
--- a/spyderlib/widgets/comboboxes.py
+++ b/spyderlib/widgets/comboboxes.py
@@ -13,7 +13,7 @@
 
 from spyderlib.qt.QtGui import (QComboBox, QFont, QToolTip, QSizePolicy,
                                 QCompleter)
-from spyderlib.qt.QtCore import Signal, Qt, QUrl, QTimer
+from spyderlib.qt.QtCore import Signal, Qt, QUrl, QTimer, QEvent
 
 import glob
 import os
@@ -27,13 +27,25 @@ from spyderlib.py3compat import to_text_string
 class BaseComboBox(QComboBox):
     """Editable combo box base class"""
     valid = Signal(bool)
+    sig_tab_pressed = Signal(bool)
+    sig_double_tab_pressed = Signal(bool)
     
     def __init__(self, parent):
         QComboBox.__init__(self, parent)
         self.setEditable(True)
         self.setCompleter(QCompleter(self))
+        self.numpress = 0
 
     # --- overrides
+    def event(self, event):
+        if (event.type() == QEvent.KeyPress) and (event.key()==Qt.Key_Tab):
+            self.sig_tab_pressed.emit(True)
+            self.numpress += 1
+            if self.numpress == 1:
+                self.presstimer = QTimer.singleShot(400, self.hanlde_keypress)
+            return True
+        return QComboBox.event(self, event)
+
     def keyPressEvent(self, event):
         """Handle key press events"""
         if event.key() == Qt.Key_Return or event.key() == Qt.Key_Enter:
@@ -58,6 +70,12 @@ class BaseComboBox(QComboBox):
         QComboBox.focusOutEvent(self, event)
 
     # --- own methods
+    def hanlde_keypress(self):
+        """ """
+        if self.numpress == 2:
+            self.sig_double_tab_pressed.emit(True)
+        self.numpress = 0
+
     def is_valid(self, qstr):
         """
         Return True if string is valid
@@ -98,6 +116,11 @@ class BaseComboBox(QComboBox):
         if valid or valid is None:
             self.add_current_text()
             return True
+
+    def set_current_text(self, text):
+        """ """
+        lineedit = self.lineEdit()
+        lineedit.setText(to_text_string(text))
 
     def hide_completer(self):
         """Hides the completion widget."""
@@ -186,27 +209,35 @@ class PathComboBox(EditableComboBox):
             self.setSizeAdjustPolicy(QComboBox.AdjustToMinimumContentsLength)
             self.setSizePolicy(QSizePolicy.Expanding, QSizePolicy.Fixed)
         self.tips = {True: _("Press enter to validate this path"),
-                     False: _('This path is incorrect.\n'
-                                    'Enter a correct directory path,\n'
-                                    'then press enter to validate')}
+                     False: ''}
+        self.sig_tab_pressed.connect(self.tab_complete)
+        self.sig_double_tab_pressed.connect(self.double_tab_complete)
 
-    def autocomplete(self, text):
-        """ """
-        self._base = text
+    def _complete_options(self):
+        """Find available completion options."""
         text = to_text_string(self.currentText())
-
-        if osp.isdir(text):
-            if text[-1] != os.sep:
-                return
         opts = glob.glob(text + "*")
         opts = sorted([opt for opt in opts if osp.isdir(opt)])
         self.setCompleter(QCompleter(opts, self))
+        return opts
+
+    def double_tab_complete(self):
+        """If several options available a double tab displays options."""
+        opts = self._complete_options()
+        if len(opts) > 1:
+            self.completer().complete()
+
+    def tab_complete(self):
+        """If there is a single option available one tab completes."""
+        opts = self._complete_options()
+        if len(opts) == 1:
+            self.set_current_text(opts[0] + os.sep)
+            self.hide_completer()
 
     def is_valid(self, qstr=None):
         """Return True if string is valid"""
         if qstr is None:
             qstr = self.currentText()
-        self.autocomplete(qstr)
         return osp.isdir( to_text_string(qstr) )
     
     def selected(self):
--- a/spyderlib/widgets/comboboxes.py
+++ b/spyderlib/widgets/comboboxes.py
@@ -40,6 +40,7 @@ class BaseComboBox(QComboBox):
         self.setEditable(True)
         self.setCompleter(QCompleter(self))
         self.numpress = 0
+        self.selected_text = self.currentText()
 
     # --- overrides
     def event(self, event):
@@ -66,6 +67,7 @@ class BaseComboBox(QComboBox):
                 self.hide_completer()
         elif event.key() == Qt.Key_Escape:
             self.set_current_text(self.selected_text)
+            print([self.selected_text])
             self.hide_completer()
         else:
             QComboBox.keyPressEvent(self, event)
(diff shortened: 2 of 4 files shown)


goanpeca:
Want to give it a try @gb119?

blink1073:
```python
  File "/Users/ssilvester/workspace/spyder-ide/spyderlib/widgets/comboboxes.py", line 200
    return
         ^
TabError: inconsistent use of tabs and spaces in indentation
```

goanpeca:
what the f...

blink1073:
I'd expect to be able to hit Tab to make a completion instead of taking me to the next widget, what do you think?

goanpeca:
hmmm did you use it?

You mean hitting tab once you have selected from the list should be the same as enter (enter is the current behaviour)

blink1073:
I removed the tab locally :smile:
When I hit enter, it just closes the dialog.

blink1073:
I mean *while typing*, just like in the command prompt.

blink1073:
or in our completions widget in the editor.

goanpeca:
Ahhh hmm ok then that means that you want the inline mode, instead of the popup mode, http://doc.qt.io/qt-4.8/qcompleter.html#CompletionMode-enum  ?

goanpeca:
Yes it makes sense totally, will fiddle around a bit more ;-)

blink1073:
Dead link, but yes, that sounds right.

ccordoba12:
Cool @goanpeca! Given this new functionality, I think we should remove the green and red coloring present in that dialog. What do you think?

goanpeca:
I could not agree more with you! I hate that red/green coloring! HATE!

goanpeca:
@blink1073 can you give it a try now? (use tab and double tab... it should feel like bash completion)

@ccordoba12 I have not changed the colors yet.

blink1073:
@goanpeca, beautiful!

goanpeca:
@ccordoba12 can you try this one out? I think we can leave the colors for now... and make those changes in a new PR, lets keep this one short

gb119:
Looks nice. On the colours issue - it's not great UI design to use red/green fonts to signify things as the most common form of colour blindness is red/gree (!). A small tick/cross emblem at the start of the path might have worked better - probably still has a value even with autocomplete as the user can typoe any sort of random string in there.

goanpeca:
Ok so I made a code update to get rid of colors and now display status icons.

Single tab autocompletes if only one option
Double tab shows completions if more that one option
There are three icons, valid, invalid and set. To set the path an enter is required. If the box looses focus without having pressed enter it will returned to the last selected path.

This is how it looks.  
https://youtu.be/zkKWeZf9ul4

goanpeca:
Maybe the icon would look better to the right :-p ?

gb119:
Very nice :-) I guess the icon location is a matter of taste, but I was going left on the analogy with the way valid https sites are shown in most web browsers (at least the ones I use...) which modify the appearance of the left side of the adress entry field. But you guys are the coders not me.....

spyderlib/plugins/inspector.py
```diff
@@ -107,15 +107,15 @@ def validate(self, qstr, editing=True):
...
                 # A new item has just been selected
                 if valid:
                     self.selected()
                 else:
-                    self.valid.emit(False)
-        else:
-            self.set_default_style()
+                    self.valid.emit(False, False)
+#        else:
+#            self.set_default_style()
```
ccordoba12: Please remove these comments

spyderlib/widgets/comboboxes.py
```diff
@@ -1,67 +1,105 @@
...
+    def event(self, event):
+        """Qt Override.
+
+        Filter tab keys and process double tab keys.
+        """
+        if (event.type() == QEvent.KeyPress) and (event.key() == Qt.Key_Tab):
+            self.sig_tab_pressed.emit(True)
+            self.numpress += 1
+            if self.numpress == 1:
+                self.presstimer = QTimer.singleShot(400, self.hanlde_keypress)
```
ccordoba12: 400 milisec seems like too much for double tab. What about 100 or 150?

spyderlib/widgets/comboboxes.py
```diff
@@ -1,67 +1,105 @@
...
         else:
             QComboBox.keyPressEvent(self, event)
 
     def focusOutEvent(self, event):
-        """Handle focus out event"""
+        """Handle focus out event restoring the last valid selected path."""
         # Calling asynchronously the 'add_current_text' to avoid crash
         # https://groups.google.com/group/spyderlib/browse_thread/thread/2257abf530e210bd
-        QTimer.singleShot(50, self.add_current_text_if_valid)
+#        QTimer.singleShot(50, self.add_current_text_if_valid)
```
ccordoba12: Remove comment (if unneeded :-)

spyderlib/plugins/inspector.py
goanpeca: Ok

spyderlib/widgets/comboboxes.py
goanpeca: http://doc.qt.io/qt-4.8/qapplication.html#doubleClickInterval-prop

ccordoba12: Ok

spyderlib/widgets/comboboxes.py
```diff
@@ -1,67 +1,105 @@
...
+        """Qt Override.
+
+        Handle key press events.
+        """
         if event.key() == Qt.Key_Return or event.key() == Qt.Key_Enter:
             if self.add_current_text_if_valid():
                 self.selected()
+                self.hide_completer()
+        elif event.key() == Qt.Key_Escape:
+            self.lineEdit().setText(self.selected_text)
```
ccordoba12: Can't we use here `self.set_current_text` here?

spyderlib/widgets/comboboxes.py
```diff
@@ -1,67 +1,105 @@
...
 
     def focusOutEvent(self, event):
-        """Handle focus out event"""
+        """Handle focus out event restoring the last valid selected path."""
         # Calling asynchronously the 'add_current_text' to avoid crash
         # https://groups.google.com/group/spyderlib/browse_thread/thread/2257abf530e210bd
-        QTimer.singleShot(50, self.add_current_text_if_valid)
+#        QTimer.singleShot(50, self.add_current_text_if_valid)
+        lineedit = self.lineEdit()
+        QTimer.singleShot(50, lambda: lineedit.setText(self.selected_text))
```
ccordoba12: And also `self.set_current_text` here?

spyderlib/widgets/comboboxes.py
goanpeca: yes

spyderlib/widgets/comboboxes.py
```diff
@@ -77,18 +115,32 @@ def add_text(self, text):
...
+
     def add_current_text_if_valid(self):
         """Add current text to combo box history if valid"""
         valid = self.is_valid(self.currentText())
         if valid or valid is None:
             self.add_current_text()
             return True
-        
+        else:
+            self.lineEdit().setText(self.selected_text)
```
ccordoba12: One more for `self.set_current_text`

spyderlib/widgets/comboboxes.py
```diff
@@ -217,14 +288,78 @@ class PythonModulesComboBox(PathComboBox):
...
         return is_module_or_package(to_text_string(qstr))
-    
+
     def selected(self):
         """Action to be executed when a valid item has been selected"""
         EditableComboBox.selected(self)
         self.open_dir.emit(self.currentText())
+
+
+class IconLineEdit(QLineEdit):
```
ccordoba12: This one shouldn't be here but on `helperswidgets`, right?

ccordoba12:
Ok, first review done :-)

spyderlib/widgets/comboboxes.py
goanpeca: I guess ;-)

spyderlib/widgets/comboboxes.py
goanpeca: Ok

goanpeca:
@ccordoba12 thanks, made the suggested changes

ccordoba12:
I like this very much, but I have a couple of suggestions:

1. The *check* emblem should disappear after a while. I see no point in showing it all the time.
2. The double tab functionality is very nice, but I'm afraid it'll be discovered only by Linux guys. We can leave it, but I'd also like to show completions after users write an `osp.sep` symbol at the end of the path. That'd be much nicer, IMHO :-)

----

Besides, I discovered by accident that with this PR Spyder CPU consumption increases a lot (to 80% or so). I don't why that happens, but it needs to be fixed :-)

ccordoba12:
Any progress on this one?

goanpeca:
@ccordoba12 sorry I have been a bit busy, looking into this now, so:

>The check emblem should disappear after a while. I see no point in showing it all the time.

I see no harm in showing it all the time, but I can do something about it on FocusOut to take it away.

>The double tab functionality is very nice, but I'm afraid it'll be discovered only by Linux guys. We can leave it, but I'd also like to show completions after users write an osp.sep symbol at the end of the path. That'd be much nicer, IMHO :-)

Windows guys that use the command line should also be using this feature, but agree most normal windows users are not aware. I will try to make a quick try to see if it can be made without changing too much code

>Besides, I discovered by accident that with this PR Spyder CPU consumption increases a lot (to 80% or so). I don't why that happens, but it needs to be fixed :-)

Yes, I had a bug in a paint event of the Icon for the combo box, it is fixed now.

goanpeca:
- Ok, I tried the separator and it basically breaks all the logic, so I am not going to make that suggestion. I prefer to include that in the video tutorials (eventually) or in a quick start guide, or as the features tour, or as tweets with tips....

- The icon I can do, but I would like some opinions, I think is harmless to just leave it there @spyder-ide/core-developers ?

goanpeca:
There is another issue to discuss here and it is, that there is no way to access the **global directory toolbar** without using the mouse.

We should be able to provide mouse free control of spyder, for advanced devs. In this direction I am considering if it still makes sense to have the global working directory as a toolbar, instead of a small pane. It might be drastic but the global directory toolbar is a weird widget itself...

The other thing to do is to offer a shortcut to focus the toolbar, which is also weird as other toolbars do not follow the behaviour.

ccordoba12:
> Ok, I tried the separator and it basically breaks all the logic

Ok, no problem. This is still an improvement ;-)

> The icon I can do, but I would like some opinions

I'm strongly +1 on hiding it on focus out. It's very distracting to see it there all the time.

ccordoba12:
The AppVeyor failure is related to conda-build being unable to do a git checkout with non-ascii chars on its log.

I think that means we should stop testing on Python 2.7 until that bug is fixed. @goanpeca, what do you think?

goanpeca:
@ccordoba12 ok, I will look into hiding the icon after focus out.

Can you test it on windows?

ccordoba12:
Yes, will do. I'm also going to comment our Appveyor tests for PY 2.7 so your PRs can be merged :-)

goanpeca:
@ccordoba12 done!

Now the icon hides on focus out and shows on focus in. Can you test locally?

ccordoba12:
Will do later today. Thanks for the change :-)

ccordoba12:
You need to rebase this one for my changes to Travis to take effect here.

ccordoba12:
Now I don't like the white space that is left in focus out:

seleccion_001

Can't we just remove the status icon in focus out and add it in focus in?

Nodd:
The text would jump left and right, that would be disturbing. One solution is to put the icon on the right side.

ccordoba12:
@Nodd's solution is nicer, and simpler to implement than mine. @goanpeca could you make it happen?

goanpeca:
Icon is on the right now

image

Any last comments before merging?

goanpeca:
Its failing for one build test

https://travis-ci.org/spyder-ide/spyder/jobs/90208316

@ccordoba12, ideas?

ccordoba12:
It was a random failure. I think you can restart the build when that happens, as all other @spyder-ide/core-developers :-)

If not, please let me know how to change that.

goanpeca:
@ccordoba12 so merge?

spyderlib/widgets/comboboxes.py
```diff
@@ -195,7 +280,7 @@ class UrlComboBox(PathComboBox):
     def __init__(self, parent, adjust_to_contents=False):
         PathComboBox.__init__(self, parent, adjust_to_contents)
         self.editTextChanged.disconnect(self.validate)
-
```
ccordoba12: Please remove blanks *only* if they're around the code you're touching. This is not a problem in this case because we almost never touch `comboboxes`, so it's a blocker for merging :-)

But I've had major headaches trying to merge things between 2.3 and master when that rule is not followed :-)

My idea is to (slowly) move to remove blanks and adhere to pep8, by following Raymond Hettinger advice: https://www.youtube.com/watch?v=wf-BqAjZb8M

ccordoba12:
Very nice addition @goanpeca! Thanks for working on it :-)

spyderlib/widgets/comboboxes.py
goanpeca: Sure, I have seen that video more than once ;-)

goanpeca:
:+1:

PR #2701, opened 2015-09-19
[WIP] Open and close panels with keyboard shortcut

goanpeca:
This benefits users that like to keep their hands on the keyboard. 

Now the shortcuts to open a closed pane (or give the focus to a pane) also close the pane if the specific pane has the focus.

### DIFF
--- a/spyderlib/plugins/__init__.py
+++ b/spyderlib/plugins/__init__.py
@@ -409,13 +409,33 @@ class SpyderPluginMixin(object):
             self.register_shortcut(qshortcut, context, name, default)
     
     def switch_to_plugin(self):
-        """Switch to plugin
-        This method is called when pressing plugin's shortcut key"""
+        """Switch to plugin.
+
+        This method is called when pressing plugin's shortcut key. This method
+        works as a toggle method, if the dockwidget is not visible it will
+        become visible, and if it is "visible" and also focussed it will
+        close and give focus to the previous focussed widget.
+        """
+        lfws = getattr(self.main, "lfws", [])
+        widget = self.get_focus_widget()
+        focus_widget = QApplication.instance().focusWidget()
+
         if not self.ismaximized:
             self.dockwidget.show()
-        if not self.toggle_view_action.isChecked():
+        if self.toggle_view_action.isChecked():
+            if widget.hasFocus():
+                self.visibility_changed(False)
+                self.toggle_view_action.setChecked(False)
+                lfws[-1].setFocus()
+            else:
+                self.visibility_changed(True)
+                self.toggle_view_action.setChecked(True)
+                lfws = [focus_widget]
+        else:
             self.toggle_view_action.setChecked(True)
-        self.visibility_changed(True)
+            self.visibility_changed(True)
+            lfws = [focus_widget]
+        self.main.lfws = lfws
 
     def visibility_changed(self, enable):
         """DockWidget visibility has changed"""


ccordoba12:
What if users want to constantly move between the Editor and an IPython console? Wouldn't this change close and open those panels again and again?

goanpeca:
No I think you misunderstood me, so I will explain with an example


Working on the editor... want to go to the console... hit the shortcut `Ctr+Shigh+Console`. Since I was in the editor, the console did not have the focus, and if it was closed it will be opened and focused.

Now I want to go back to the editor I hit `Ctrl+Shift+Editor`.Since the editor did not have the focus but was already opened, it will simply get the focus.

Only if I hit `Ctrl+Shift+Editor` inside the focused editor, it would actually close it. In this case the focus would go to the last focused widget before the editor.

Nodd:
The problem is that if I come back to spyder after a while and I want to make sure that the focus is in the editor, I'd machinely hit `Ctrl+Shift+Editor` just to be sure, and it would close it.
Opening the pane if it's not is a nice feature. Closing it with the same shortcut doesn't look that good.

ccordoba12:
Totally agree with @Nodd! It'd be very unintuitive to close a pane if it has focus and you press its own shortcut.

But opening a pane if it's closed and you use its shortcut it's a very nice feature :-)

goanpeca:
I understand your points of view, but we need to be consistent. On one side we want to support vim and emacs keybindings. These are all meant to keep the user in the keyboard and not mess around with the mouse... on the other hand, if I open a previously invisible panel, then I am stuck with it... even if I just wanted it for a moment.

I think right now the UI does not work well for keyboard only mode....

goanpeca:
Ok a middle ground would be to never close the editor?

ccordoba12:
> on the other hand, if I open a previously invisible panel, then I am stuck with it... even if I just wanted it for a moment

Then the solution (for me) is to create a different shortcut to close the currently focused pane (whichever it is).

This way, users would need to remember this shortcut when they want to use it, instead of being surprised by Spyder closing a pane when they wanted to focus it.

Nodd:
Yep, a specific shortcut to close the current pane is good. And it already exists, Ctrl+Shift+F4 by default.

goanpeca:
Ok, sure, but still it is very poor UI design.

The options in the panes section are toggle actions.... if you go to the menu and click it will open and show a tick, and will display the shortcut, and if you go and toggle an active pane it will close it, and it will still display the same shortcut.

That would say that the shortcut should also function as a toggle action but it does not, and is what this PR aims to resolve....

ccordoba12:
-1 from me then. There is already a shortcut to close a pane, and that's enough for me. Focus and close are not opposed actions, focus and unfocus (i.e. giving focus to the last focused pane, or something like that) are :-)

@goanpeca, you could propose a simpler shortcut if you find the current one not too good.

goanpeca:
whatever...

ccordoba12:
But the idea of making a pane visible if it was hidden, by pressing its shortcut was a good one. @Nodd agreed with that one too.

So I don't understand why you closed this PR prematurely :-)

Nodd:
Actually, the functionality already exists in spyder. This PR was only for the closing part.

I agree that there is a little inconsistency between the menu entry and the shortcut, (the menu entry toggles and the shortcut focuses), but I think that the functionality is more important here. Accidentally closing the file explorer is not a big deal. Accidentally closing all IPython consoles instead of focusing can potentially wipe out lots of work.

ccordoba12:
Ups, sorry for the misunderstanding :-p

We could also remove the shortcuts, if that's causing the wrong grasp of our UI, but I think not that much people was aware of their existence before I put them in the Panes menu.

Issue #2696, opened 2015-09-16
Enter in the IPython console inserts new line instead of executing current line after kernel restart


derDavidT:
Recently I updated my anaconda package, which installed

    ipython 4.0.0
    ipython-qtconsole 4.0.1
    spyder 2.3.6

After that, whenever I restart the ipython kernel in the qt-console (using `ctrl + .`), the console will not execute any command, e.g.:

    Python 2.7.10 |Anaconda 2.3.0 (64-bit)| (default, May 28 2015, 16:44:52) [MSC v.1500 64 bit (AMD64)]
    Type "copyright", "credits" or "license" for more information.

    IPython 4.0.0 -- An enhanced Interactive Python.
    ?         -> Introduction and overview of IPython's features.
    %quickref -> Quick reference.
    help      -> Python's own help system.
    object?   -> Details about 'object', use 'object??' for extra details.
    %guiref   -> A brief reference about the graphical user interface.

    In [1]: 1
       ...: 

I cannot reproduce this behavior when running ipython-qtconsole outside spyder, i.e. it works as expected. My current workaround ist closing the console, thus killing the kernel, and load a fresh one.

bordingj:
Same behavior here (Python 3.4.3, Anaconda 2.3, Ubuntu 14.04)

petebachant:
I'm getting the same behavior on Windows 10 x64, Python 3.

ccordoba12:
Thanks for reporting. I'll try to fix it for 2.3.7

ccordoba12:
I can reproduce this error both in Python 2 and Python 3. No idea where it's coming from though :-)

dluzenst:
Same problem with Spyder 2.3.6, PyQt4 4.10.4:

             platform : win-64
        conda version : 3.18.1
  conda-build version : 1.17.0
       python version : 2.7.10
     requests version : 2.7.0

ccordoba12:
This is an upstream bug. i.e. meaning a bug that needs to be fixed in Jupyter `qtconsole`: https://github.com/jupyter/qtconsole

I'll try to submit a PR to `qtconsole` after releasing 2.3.7.

Fishyman:
if you hold down shift and press enter it works fine (although not what we want)

davidljung:
Anyone happen to know which qtconsole issue # is the cause of this?

ccordoba12:
I think this issue only happens in Spyder, so that's why it shouldn't be reported upstream.

[...]

Issue #2679, opened 2015-09-08
New IPython consoles or restart of kernel not working because of pkg_resources error


Diti24:
## Description

**What steps will reproduce the problem?**

1. Stat spyder
2. try to create a new ipython console (in Consoles-> Open an IPython console

**What is the expected output? What do you see instead?**
A new console should appear but the new console immediately stops with the following error:

```
Error in sitecustomize; set PYTHONVERBOSE for traceback:
TypeError: 'NoneType' object is not callable
Traceback (most recent call last):
File "/pyzo2014a/lib/python3.4/site‑packages/spyderlib/widgets/externalshell/start_ipython_kernel.py", line 175, in 
from IPython.kernel.zmq.kernelapp import IPKernelApp
File "pyzo2014a/lib/python3.4/site‑packages/IPython/__init__.py", line 48, in 
from .terminal.embed import embed
File "pyzo2014a/lib/python3.4/site‑packages/IPython/terminal/embed.py", line 16, in 
from IPython.core.interactiveshell import DummyMod
File "pyzo2014a/lib/python3.4/site‑packages/IPython/core/interactiveshell.py", line 31, in 
from pickleshare import PickleShareDB
File "pyzo2014a/lib/python3.4/site‑packages/pickleshare.py", line 41, in 
from path import path as Path
File "pyzo2014a/lib/python3.4/site‑packages/path.py", line 122, in 
__version__ = pkg_resources.require('path.py')[0].version
TypeError: 'NoneType' object is not callable
```

**Please provide any additional information below**

## Version and main components

* Spyder Version: 2.3.6 
* Python Version: 3.4.3
* Qt Versions:  4.8.6, PyQt4 (API v2) 4.11.3 on Linux

## Optional dependencies
```
pyflakes >=0.6.0:  0.9.2 (OK)
pep8 >=0.6      :  1.6.2 (OK)
pygments >=1.6  :  2.0.2 (OK)
IPython >=1.0   :  4.0.0 (OK)
zmq >=2.1.11    :  14.7.0 (OK)
qtconsole >=4.0 :  4.0.1 (OK)
pandas >=0.13.1 :  0.16.2 (OK)
sphinx >=0.6.6  :  1.2.3 (OK)
psutil >=0.3    :  2.2.1 (OK)
rope >=0.9.2    :  0.9.4-1 (OK)
jedi >=0.8.1    :  0.9.0 (OK)
matplotlib >=1.0:  1.4.3 (OK)
[...]

PR #2716, opened 2015-09-25
PR: [WIP] Test IPython console

petebachant:
I am trying to create a failing test to detect issue #2679 or #2696. Currently this does not work. Since this is over my head, I'm looking for some feedback on where to turn next. I suppose this is also working towards #2268.

### DIFF
--- a/spyderlib/tests/ipythonconsole.py
+++ b/spyderlib/tests/ipythonconsole.py
@@ -1,0 +1,13 @@
+"""
+Tests for the IPython console.
+"""
+
+def test_new_ipython_console():
+    """Test the ability to open a new IPython console."""
+    from IPython.kernel.zmq.kernelapp import IPKernelApp
+    from spyderlib.plugins.ipythonconsole import IPythonConsole
+    from spyderlib.spyder import MainWindow
+    mw = MainWindow()
+    i = IPythonConsole(mw)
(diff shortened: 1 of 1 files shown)


ccordoba12:
Thanks a lot for this @petebachant. @goanpeca, could you take a look at this one?

goanpeca:
@petebachant thanks for this ! we really need to improve our (non existent  :wink: test codebase).

I think we want to have tests close to the tested feature, so I guess in this case, we would need to move this test inside `spyderlib/plugins/tests` what do you think?

petebachant:
Looking at some other projects, e.g. IPython and pandas, you're right. They do put tests for subpackages inside `{subpackage}/tests`. I will move that at least. Any ideas on how I might simulate opening an IPython console without ever creating the main window? Or is this something that should be done by opening the full app, and using some other testing tools to click the appropriate buttons?

goanpeca:
We could use pytest-qt, are you familiar?

petebachant:
I've used pytest a little but mainly Nose. I will look into pytest-qt when I have time again!

ccordoba12:
@petebachant, the way we integrate with the qtconsole is going to be changed drastically after PR #2878 is merged, so I'm closing this for now.

Please feel free to open a new PR after that that one is merged.

PR #2722, opened 2015-09-28
ENH: Allow develop install mode

Eric89GXL:
Small enhancement that allows doing:
```
python setup.py develop
```
In this mode, any local changes are immediately reflected on next launch without having to modify paths, which is nice.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,6 +23,10 @@ import shutil
 from distutils.core import setup
 from distutils.command.build import build
 from distutils.command.install_data import install_data
+try:
+    import setuptools  # allows python setup.py develop
+except ImportError:
+    pass
 
 # Check for Python 3
 PY3 = sys.version_info[0] == 3


goanpeca:
@Eric89GXL thanks, we are moving over to setup tools soon :-)

Eric89GXL:
This might make a decent stop-gap measure for other devs before full migration. I don't *think* it should cause any problems to have it. And it the meantime, it's still an optional component. But if you prefer to wait for a more complete solution, feel free to close.

goanpeca:
Sure, @ccordoba12 is working on that so it will be his call :-)

ccordoba12:
@Eric89GXL, but we always have used `bootstrap.py` for this kind of purpose. Isn't that enough?

Eric89GXL:
What is the procedure for using `bootstrap`? I'm not familiar with it. It's not a standard thing AFAIK, which is one knock against it.

By contrast, doing `python setup.py develop` is standard across many repos. And once you do it, you can launch Spyder normally as `spyder` from any terminal, and it will use the currently checked out version. But maybe it is missing some magic that `bootstrap` otherwise takes care of...?

Nodd:
I think that the feature itself is nice, develop install is quite handy. I wouldn't personally use it for spyder though because I'd like to keep a working version just in case...
Still I'm afraid that using both distutils and setuptools can lead to conflicts in some corner case, so I'd prefer to wait for the real switch to setuptools.

Eric89GXL:
@Nodd in my experience many packages use `import setuptools`, at least in certain cases / for certain commands (numpy, scikit-learn, etc.) so I think it's pretty safe. I'll go ahead and close, though, since it sounds like people are more comfortable with the full-switch solution.

Eric89GXL:
It has been about 9 months without a `setuptools` transition. Any interest in putting this in in the meantime, given that the full `setuptools` transition might not be a high priority?

ccordoba12:
We did the transition, but only for wheels (please take a look at our setup.py).

I think you could also use setuptools for `develop`, but given our preference for using our `bootstrap.py` script, I think there will be little interest about it.

PR #2740, opened 2015-10-03
Read correctly the run config for profiling

### LINKED ISSUE #2736: Profiler config does not carry command line arguments correctly
Currently their is no run configuration for the profiler. Issue #1110 was opened to request it and indeed it would be nice. In the meanwhile, from what I can see the profiler uses the Run config associated with the file. Indeed if I add command line options in the run config and start the profiler, the options are used. But there seem to be a bug in this mechanism. To reproduce:

 1. Create or open a file that parses the command line but accepts an empty one. For example:

        parser = argparse.ArgumentParser()
        parser.add_argument('--opt', dest='optionalarg')
        args = parser.parse_args()

    Profile it without giving command line options to make sure it's fine.
 2. Create or open another file that also parses the command line. Profile it **with** command line options that the first file would **not** accept.
 3. Come back to the first file, and profile it with <kbd>F10</kbd>

Expected behaviour: The file should be profiled with the options used during the first profiling (i.e. no option) .
Actual behaviour: It is run with the command line options that were given to the second file. If using my example, the argument parser will tell you that it encountered an unknown argument. 

To workaround it, I have to open the Run config, put some command line arguments (even if don't want any), run the file, re-open the config, remove the arguments, and that's only when the file will be run as expected.

rthouvenin:
Fixes #2736

When the profiler is run from the top menu or from a shortcut, and if the run config has disabled command line arguments, the profiler should be called with `args=[]` instead of `args=None`, to ensure that no argument will be given to the script.

### DIFF
--- a/spyderplugins/p_profiler.py
+++ b/spyderplugins/p_profiler.py
@@ -130,7 +130,7 @@ class Profiler(ProfilerWidget, SpyderPluginMixin):
             self.dockwidget.raise_()
         pythonpath = self.main.get_spyder_pythonpath()
         runconf = runconfig.get_run_configuration(filename)
-        wdir, args = None, None
+        wdir, args = None, []
         if runconf is not None:
             if runconf.wdir_enabled:
                 wdir = runconf.wdir


ccordoba12:
Thanks a lot for working on this on getting to the root of the problem!

PR #2758, opened 2015-10-12
Fix failing fallback_plugin tests and add to modules_test

### LINKED ISSUE #2756: fallback_plugin tests are failing
As I was working on PR #2752, i discovered that those tests are failing. I skipped them for now in `modules_test.sh`, but they need to be fixed.

@blink1073, could you take care of that (given that you wrote them)? They should be easy to fix :-)

blink1073:
Fixes #2756.


ccordoba12:
Thanks for the quick response @blink1073!

PR #2761, opened 2015-10-12
Test Spyder with Python 3.5

ccordoba12:
Ok, let's keep going with more testing and now do this! :-)

All Spyder dependencies are available as conda packages in 3.5, so why not?


ccordoba12:
Well, Jedi 0.8.1 is missing, so I'll have to do something about it :-)

PR #2762, opened 2015-10-12
Start testing with PyQt5

ccordoba12:
More testing :-)


ccordoba12:
The only way to keep going with this is to compile qt5 and pyqt5 ourselves. The conda versions I've found are old and with too much deps on system libraries.

ccordoba12:
Success!! We are now testing PyQt5 with PY 2.7 and 3.5, but not 3.4. The idea is too quickly drop support for 3.4 (in a month or two :-)

This is another great step to move forward to 3.0 final ;-)

Nodd:
Why drop 3.4 ? It's quite recent, and not everyone uses Anaconda...

ccordoba12:
I thought we only had a limited number of slots in Travis, but that doesn't seem to be the case :-)

I'll leave it then until 3.5 is adopted more broadly.

SylvainCorlay:
Does anaconda ship PyQt5? If yes, this is great news for retina display users.

ccordoba12:
Nop, not yet. I created packages for Linux (for now). They are available in our official Anaconda channel:

http://anaconda.org/spyder-ide

I hope to create Mac and Windows packages later on :-)

But that's another reason for making PyQt5 the default in Spyder 3.

goanpeca:
@SylvainCorlay can you register at anaconda.org

SylvainCorlay:
Done! I am also @SylvainCorlay on anaconda.org.

goanpeca:
Cool, I will add you to the list of owners, now need @Nodd and @blink1073 to follow your steps :wink:

goanpeca:
@blink1073 is this you https://anaconda.org/blink1073 ?

blink1073:
Yep

PR #2767, opened 2015-10-13
Fix arrayeditor import error in PyQt5

### LINKED ISSUE #2766: fix ArrayEditor under PyQt5
hi,

1136a1e08e74abb65d77dbea452479c02aba454f introduced a bug, as `QItemSelection` and `QItemSelectionRange` are in QtCore now.

flying-sheep:
fixes #2766


ccordoba12:
Thanks a lot for the (really quick!) fix :-) That's why we need to start testing with PyQt5 too!

ccordoba12:
@goanpeca, this is something you should add to `qtpy` too. Just letting you know about it :-)

PR #2776, opened 2015-10-20
Rename spyderlib.widgets.editors to spyderlib.widgets.variableexplorer

ccordoba12:
@PierreRaybaut, I don't agree with your change. I'd had preferred that you would have opened a pull request for it instead of pushing directly to `master`.

These are my reasons:

1. The Variable Explorer is one of the less understood parts of Spyder, and your change really doesn't help other people to grasp how it works internally (which was my rationale to do the move in the first place). On the contrary, it makes things less clearer than before because:
    1. The `varexp` namespace helped to convey the idea that all things under it are related to the variable explorer. We could use `variableexplorer` if you prefer :-)
    2. `editors` sounds to something related to the Editor, which is clearly not the case for Spyder.
    3. `importwizard` and `namespacebrowser` are not editors (in the sense of the Variable Explorer) at all.

2. You're trying to support external projects with this change. I know both Guidata and Spyder are your projects, but there are a lot of people working on Spyder now, so this kind of changes can't be made as lightly as before. Besides, and I hope you don't take this in the wrong way, you've been away of Spyder for two years, so it is now as much the work of the Spyder Team as it is yours :-) 

3. I decided to go with a major bump in Spyder version (i.e. 3.0) precisely to be able to do this kind of breakage and expect other projects to adapt to it.

4. I don't see why Guidata can't import from `spyderlib.widgets.varexp`.


CHANGELOG.md
```diff
@@ -1,21 +1,5 @@
...
-### API Changes
-
-* `spyderlib.dicteditor.DictEditor` has been renamed to `spyderlib.widgets.editors.collectionseditor.CollectionsEditor`
-* Editor widgets were moved from `spyderlib.widgets` to `spyderlib.widgets.editors`:
-    * `spyderlib.widgets.editors.arrayeditor`
-    * `spyderlib.widgets.editors.collectionseditor`
-    * `spyderlib.widgets.editors.objecteditor`
-    * `spyderlib.widgets.editors.texteditor`
-    * `spyderlib.widgets.editors.dataframeeditor`
-
```
ccordoba12: I'm keeping a record of new things for Spyder 3.0 in our wiki, not here

https://github.com/spyder-ide/spyder/wiki/Beta-version-changelog

ccordoba12:
I decided to open a pull request instead of pushing directly to master to discuss this situation.

So I invite the other @spyder-ide/core-developers to join in, and let the rest know what they think about this :-)

PierreRaybaut:
@ccordoba12, I'm really sorry for this change: it was clearly not a good idea to change things so quickly and without asking the other developers about it. But your initial change (moving those widgets to `spyderlib.widgets.varexp`) was so brutal for me... this is my baby!! :-)) I wrote those widgets (`ArrayEditor` and `DictEditor`) in the very beginning, even before Spyder was named Spyder (it was Pydee!).

Let me answer to your first message:
1. I understand the motivation. I think that `spyderlib.widgets.variableexplorer` would be more explicit for users that are external to the project (like me?).
2. You're right, I've been away for a long time and believe me, it was not my intention to impose anything to anyone: I respect your work and the work of others and I realize now that this commit was a little bit impulsive :-). I really thought it was justified at the moment and judging by the date of your commit moving those widgets to `spyderlib.widgets.varexp` (12 days ago), I thought that it would not bother too much people. However, note that it was not solely motivated by the API break in `guidata` but also in other libraries using those very useful widgets: I thought that if I was surprised of it, it would surely surprise any other user.
3. That's a good thing. Change is good. I guess that my only objection would be the name of this package: `varexp`.
4. You're right, it can import anything from anywhere but the thing is to know where to find it. When browsing Spyder source code, I doubt that anyone looking for an array editor widget will look for it in `varexp` package. Again, I was trying to think about other libraries, not only `guidata`.

Sorry again for messing with things... it's like I'm returning home and there are other people living there, I want to do only what *I* want when *I* want... but I can't... those people want a pull-request democracy!? :smile: Just joking (really no hard feelings). You (and the others) have done a great work on Spyder and I don't intend to return and pretend to be the one and only developer.

## Conclusion
Anyway, this being said, `spyderlib.widgets.variableexplorer` (instead of `spyderlib.widgets.varexp`) is fine. No need to spend more time on this. That's perfectly ok for me.
Would you do this for me?

goanpeca:
Hi guys, jumping in, I do like the long name, short names do not save anything, we have autocompletion anyway so really making it `varex`  is making the code harder to understand.

Would it make sense to have the widgets be imported in `widgets/__init__.py` that way we can document them as stand alone widgets?, no api would be broken that way.... even if those widgets are specific to the variable within spyder, they are also general purpose widgets?

PierreRaybaut:
Hi @goanpeca, this would be a good compromise indeed.

Something like this in `spyderlib/widgets/__init__.py` would do the trick:
```python
from spyderlib.widgets.variableexplorer import arrayeditor
from spyderlib.widgets.variableexplorer import collectionseditor
from spyderlib.widgets.variableexplorer import dataframeeditor
from spyderlib.widgets.variableexplorer import objecteditor
from spyderlib.widgets.variableexplorer import texteditor
```

Besides, something like this should be added to `collectionseditor.py`:
```python
import warnings  # move this import to the top of the file, of course

class DictEditor(CollectionsEditor):
        warnings.warn("`DictEditor` has been renamed to `CollectionsEditor` in Spyder 3: "\
                      "please use `CollectionsEditor` instead", RuntimeWarning)
        CollectionsEditor.__init__(self, parent)
```

goanpeca:
@PierreRaybaut exactly what I had in mind :-) !

ccordoba12:
+1 to both of your comments. Thanks a lot for the feedback!

I'll implement them as part of this PR :-)

PierreRaybaut:
See my updated comment above.
I've added the following missing statement in `spyderlib/widgets/__init__.py`:
```python
dicteditor = collectionseditor
```
(otherwise the compatibility would be broken for libraries trying to import `spyderlib.widgets.dicteditor.DictEditor`, and the warning would have been added for nothing)

ccordoba12:
@PierreRaybaut, I really understand your feelings :-) And thanks a lot for not taking these changes in the wrong way.

I have to admit that even I had a bit of hard time understanding how the Variable Explorer works. For example, I decided to rename `DictEditor` because it doesn't only create a view of dictionaries but also of lists and tuples.

Besides, I put a lot of sweat and tears during the last year fixing bugs and shortcomings in the Variable Explorer. So I was hoping that if it's clearer for other people how it works, then they could help me with that :-)

ccordoba12:
Ok, I think is ready :-) I had to move `widgets/sourcecode/syntaxhighlighters.py` to `utils/syntaxhighlighters.py` to avoid circular imports when adding import statements to `widgets/__init__.py`.

Given that that file doesn't contain any widget, I hope you're ok with this change :-) In any case, please let me know that you agree with it before I merge this.

PierreRaybaut:
Looks good to me.

PR #2786, opened 2015-10-30
Move the spyder script to the right place if it isn't present in the site Scripts directory

lpinner:
The post install script assumes the ```spyder``` script is in the site scripts directory, but it isn't always installed there, i.e if spyder was installed using ```pip install --user spyder``` 

If the user then runs spyder_win_post_install.py -install manually, the shortcuts will be created pointing to a non-existent script.

### DIFF
--- a/scripts/spyder_win_post_install.py
+++ b/scripts/spyder_win_post_install.py
@@ -95,7 +95,7 @@ except NameError:
                   file=sys.stderr)
             sys.exit(1)
         from win32com.shell import shell, shellcon
-        
+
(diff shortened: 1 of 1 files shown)


Nodd:
It looks like your line could be always used, even if `spyder` is in `sys.prefix + '/scripts'`. The code would be simpler.

ccordoba12:
Yes, sorry about the failure. I'm seeing how to fix it :-)

ccordoba12:
Thanks a lot for your contribution. It seems the right fix, so I'm merging it ;-)

PR #2810, opened 2015-11-14
Restore the insertion of the "(" character when parameter close_parentheses_enabled is False

### LINKED ISSUE #2795: Option  'Automatic insertion of parentheses, braces and brackets' has issues when un-checked
We can't type `print ("Hello Word") , with Spyder 3 trunk of October 24th morning and 'Automatic insertion of parentheses, braces and brackets' unchecked

issue is on Winpython3.4.3.6 at least 
discovered by @ezbentley  https://github.com/winpython/winpython/issues/193

stonebig:
Fixes #2795

----

Restore the insertion of the "(" character when parameter close_parentheses_enabled is False.
 
I suspect some patch destroyed the functionnality formerly in handle_close_parentheses(text)

This is the simple fix, maybe a more complete fix would remove unused path in def  handle_close_parentheses()

### DIFF
--- a/spyderlib/widgets/sourcecode/codeeditor.py
+++ b/spyderlib/widgets/sourcecode/codeeditor.py
@@ -2603,6 +2603,8 @@ class CodeEditor(TextEditBaseWidget):
             self.hide_completion_widget()
             if self.close_parentheses_enabled:
                 self.handle_close_parentheses(text)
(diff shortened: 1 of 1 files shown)


ccordoba12:
Thanks for the fix, I also think it's the right one :-)

PR #2826, opened 2015-11-23
Remove imports from widgets/__init__ because they are making Spyder crash

ccordoba12:
- Crashes happened when pandas or numpy were not installed because dataframeeditor and arrayeditor import them directly.
- It seems we can't improve further this situation, but at least API changes are documented.

----

I made this PR to let @goanpeca and @PierreRaybaut know about this change. Leaving those imports in `widgets/__init__` to maintain compatibility with Spyder 2 would make Spyder 3 to hard depend on Pandas and NumPy, but we can't afford that for this simple change :-)

### DIFF
--- a/spyderlib/widgets/__init__.py
+++ b/spyderlib/widgets/__init__.py
@@ -13,12 +13,3 @@ Widgets defined in this module may be used in any other Qt-based application
 They are also used in Spyder through the Plugin interface
 (see spyderlib.plugins)
 """
-
-from spyderlib.widgets.variableexplorer import arrayeditor
-from spyderlib.widgets.variableexplorer import collectionseditor
-from spyderlib.widgets.variableexplorer import dataframeeditor
-from spyderlib.widgets.variableexplorer import objecteditor
-from spyderlib.widgets.variableexplorer import texteditor
-
-# For compatibility with Spyder 2
-dicteditor = collectionseditor


goanpeca:
@ccordoba12  Why not just wrap them in try and catch statements?

ccordoba12:
Where? I think it doesn't make sense in `widgets/__init__` because it's public APi, so it should always be available.

And it could make sense in `dataframeeditor` and `arrayeditor`, so that they don't fail without pandas and numpy. But that doesn't make sense to me either because their main purpose is only fulfilled is those modules are present.

goanpeca:
Wherever it makes sense.... but removing stuff is not a solution, 

It feels the same as the "solution" of only allowing Gtk for Gtk users.. in the preferences...

ccordoba12:
In this case we're moving stuff around and it's documented in our Changelog for Spyder 3.0, so upstream projects have to adapt to it.

Leaving things as they are will make Spyder crash when modules we're not demanding in our conda recipe nor in our wheel are not present (i.e. pandas and numpy).

In short, we tried to make life for upstream projects a bit easier by adding those imports to `widgets/__init__`, but that just makes life harder for us, so they'll have to learn the new API, that's all :-)

ccordoba12:
Besides, what would be the code for a try/except in `widgets/__init__`?

```
try:
    from spyderlib.widgets.variableexplorer import arrayeditor
    ...
except:
    arrayeditor = FakeClass
```

ccordoba12:
That doesn't make sense to me, it's much easier for upstream projects to import directly those modules from the new namespace (i.e. `variableexplorer`).

What's the problem with that?

goanpeca:
Ask @PierreRaybaut ;-)

I am not an upstream user so I cannot answer that....

ccordoba12:
@goanpeca, I hope you understand this PR, and also the Gtk stuff, are temporary workarounds that are better than leaving the situation as it currently is (i.e. with a hard crash at startup and no tooltips).

You're welcome to improve things by sending new PRs. But in the meantime I need to move forward, and don't have time to fix things the right way for every little detail that we have to do (and in the case of Gtk, it was not worth it in my opinion because the other themes didn't add a great deal of improvement to the UX, but instead diminished it by messing with tooltips :-)

PierreRaybaut:
Hi there,

I think that it's not worth it: let's remove those import statements and lose backward compatibility as long as it's properly documented (where is it documented by the way?). The upstream projects developers will have to adapt (after all, the import error exception is quite explicit).

-Pierre
________________________________
De : Carlos Cordoba<mailto:notifications@github.com>
Envoyé : ‎23/‎11/‎2015 03:11
À : spyder-ide/spyder<mailto:spyder@noreply.github.com>
Cc : Pierre Raybaut<mailto:pierre.raybaut@gmail.com>
Objet : Re: [spyder] Remove imports from widgets/__init__ because it was making Spyder crash (#2826)

@goanpeca, I hope you understand this PR, and also the Gtk stuff, are temporary workarounds that are better than leaving the situation as it currently is (i.e. with a hard crash at startup and no tooltips).

You're welcome to improve things by sending new PRs. But in the meantime I need to move forward, and don't have time to fix things the right way for every little detail that we have to do (and in the case of Gtk, it was not worth it in my opinion because the other themes didn't add a great deal of improvement to the UX, but instead diminished it by messing with tooltips :-)

---

ccordoba12:
@PierreRaybaut, it is documented in our beta changelog here:

https://github.com/spyder-ide/spyder/wiki/Beta-version-changelog#api-changes

PR #2831, opened 2015-11-26
Remove external dependencies

ccordoba12:
This PR removes libraries that we added to our tree to make Spyder self-contained. But now they are not necessary because we're relying on pip and conda to install dependencies.


ccordoba12:
@picca and @stonebig, I want you to be aware of this change because you're the packagers I know: from now on Spyder depends on `qtawesome`, which depends on `qtpy`, so please be sure to package those libraries for the next Spyder release :-)

Besides, Spyder will also depend on `path.py` and `pickleshare`, but I didn't add them to our `setup.py` because they are pulled by `qtconsole` (which is also a first class dependency for us now :-)

Please review our `setup.py` or our conda recipe (inside continuous_integration) to see what other packages we're declaring as dependencies.

ccordoba12:
@PierreRaybaut, I think you should also be aware of this change. I know you tried for a long time to make Spyder self-contained, but I decided to move away from that model to make things more flexible and easy for external contributors.

picca:
I hope that this is true only for the v3.x series :)

ccordoba12:
@picca, yes, of course :-) This change only applies to Spyder master (i.e. the future 3.x series), no change was done for Spyder 2.3.

ccordoba12:
@Nodd, I think you know the Arch Linux packagers, so please inform them of this change :-)

picca:
Do you know the license of the embeded fonts and icons.

Font Awesome and Elusive Icons.

picca:
The important point for debian is to be able to use the system font instead of your embeded copies.
Do you know if by defautlt qtawesome use the system fonts ?

stonebig:
Great !

Thanks for the head-up @ccordoba12.

Would it be possible to have `qtawesome` and `qtpy` published also as wheels on pypi ?

Otherwise I have to do the transformation by hand or multiply the requirements files for pip.

picca:
I fond this in Debian is it the right fonts ?

Package: fonts-font-awesome
Version: 4.4.0~dfsg-1
Installed-Size: 1063
Maintainer: Debian Fonts Task Force <pkg-fonts-devel@lists.alioth.debian.org>
Architecture: all
Enhances: libjs-twitter-bootstrap, node-less, ruby-sass
Description-en: iconic font designed for use with Twitter Bootstrap
 This font contains about 249 various icon glyphs. Glyphs are designed
 as scalable vector graphics hence display very well at any screen
 size. This font was basically designed to be used with the Twitter
 bootstrap  library but can be used in other places also.
 .
 This package also provides Sass and LESS files to use webfonts from
 this package.

goanpeca:
@stonebig QtPy should be pip instalable...https://pypi.python.org/pypi/QtPy/0.1.2

ccordoba12:
@goanpeca, `qtpy` only has tarballs for installation on PyPi, but I think WinPython (@stonebig's project) is using wheels as its package format.

Could you upload a wheel to help them? It should be pretty easy to generate.

This is the command we use in Spyder:

    python setup.py bdist_wheel --universal

I'm not sure about the command to upload it, but it should be pretty easy to google for it ;-)

ccordoba12:
@SylvainCorlay, could @picca rely on Debian's package to provide FontAwesome instead of the one bundled with QtAwesome?

SylvainCorlay:
@ccordoba12 Debian's font-awesome seems to be pretty old. Besides, apt-get has a different approach to js packaging (one version for everyone) from npm or bower (each project may have different version of its dependencies).

Regarding the bundling of `font-awesome` and `elusive icons` in qtawesome, I was thinking of
 - uploading the script building the charmap in the repo (I have to re-write it first :) )
 - getting font-awesome and elusive icons at build time, via `npm` or `bower` (then working with qtawesome from source would require npm).

ccordoba12:
@SylvainCorlay, I have no problems regarding your plans for QtAwesome, as long we don't need npm or bower to *use* the package. It's fine to have them for development purposes :-)

PierreRaybaut:
@ccordoba12: the choice to keep Spyder self-consistent was fully justified when the development team was limited to... 1 people (or 2 when you joined the project). Now that there are more and more motivated people contributing to the project... and even *leading* the project, separating those modules from Spyder source code seems to be a good choice.

ccordoba12:
@PierreRaybaut, it's really good to know that you also think we're moving in the right direction with this kind of changes :-)

ccordoba12:
Just for the record, the only dependency I'm leaving is `lockfile` for two reasons:

1. This is a Twisted module, and we can't demand Twisted to run Spyder :-)
2. `lockfile` is already taken in PyPi, so we can't our own package out of it.

goanpeca:
@ccordoba12 we should still make a PyPi package with a slightly different name then?

ccordoba12:
We certainly could, if you think it's worth the effort :-)

What name do you propose?

goanpeca:
Can you point me to the file? I have no idea what that module does

ccordoba12:
It's in `spyderlib/utils/external`, and it's used to detect if Spyder is already running or not, by generating a file that is locked by Spyder when it's started.

SylvainCorlay:
"lock" is not taken on pypi!

goanpeca:
It is now  :laughing: 

https://pypi.python.org/pypi?name=lock&version=0.0.1&:action=display

techtonik:
> @PierreRaybaut, I think you should also be aware of this change. I know you tried for a long time to make Spyder self-contained, but I decided to move away from that model to make things more flexible and easier for external contributors.

And how it is easier? I see that this became harder. If previously I can just checkout and run, then now I have to create virtual env, cd, activate it and then run Spyder. That's suxx.

Nodd:
@techtonik I just checkout and run, since I already have all the dependancies installed. And now I won't have a different behavior for my programs inside and outisde spyder because the external dependancies were not up to date...

PR #2847, opened 2015-12-04
Use High dpi pixmaps

SylvainCorlay:
This makes Spyder (even more) beautiful, especially with the new icon theme.

### DIFF
--- a/spyderlib/spyder.py
+++ b/spyderlib/spyder.py
@@ -272,6 +272,8 @@ class MainWindow(QMainWindow):
         QMainWindow.__init__(self)
 
         qapp = QApplication.instance()
+        if PYQT5:
+            qapp.setAttribute(Qt.AA_UseHighDpiPixmaps)
         self.default_style = str(qapp.style().objectName())
 
         self.dialog_manager = DialogManager()


SylvainCorlay:
With PyQt4, this has no effect.

spyderlib/spyder.py
```diff
@@ -272,6 +272,7 @@ def __init__(self, options=None):
         QMainWindow.__init__(self)
 
         qapp = QApplication.instance()
+        qapp.setAttribute(Qt.AA_UseHighDpiPixmaps)
```
ccordoba12: Does PyQt4 have a property called `AA_UseHighDpiPixmaps`?

SylvainCorlay: Actually it does not. I was mistaken.

spyderlib/spyder.py
```diff
@@ -272,6 +272,8 @@ def __init__(self, options=None):
         QMainWindow.__init__(self)
 
         qapp = QApplication.instance()
+        if PYQT5:
+            qapp.setAttribute(Qt.AA_UseHighDpiPixmaps)
         self.default_style = str(qapp.style().objectName())
```
goanpeca: Yes much better!

goanpeca:
:+1:

spyderlib/spyder.py
ccordoba12: Yep, this is nicer!

Issue #820, opened 2015-02-17
Move all color-related options into Preferences -> Color scheme


spyder-issues-migrator:
_From [techtonik@gmail.com](https://code.google.com/u/techtonik@gmail.com/) on 2011-11-01T02:17:32Z_

Spyder Version:  2.1.0
Python Version:  2.7.2 
Qt Version:      4.7.4, PySide 1.0.7 on Linux 

What steps will reproduce the problem?

 To change backround color for Spyder and all its plugins, you need to click through all preference pages to find out which scheme is active.

It will be much more ituitive if all scheme selection and scheme color options were concentrated on the same page. This will require some 'extension point' mechanizm which allows plugins to plug into scheme selection page. Any ideas how to implement this?

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=820_

spyder-issues-migrator:
_From [ccordoba12](https://code.google.com/u/ccordoba12/) on 2014-08-23T13:23:14Z_

**Status:** Accepted  
  **Labels:** Cat-Settings MS-v2.4

goanpeca:
I am doing the redesign of the Syntax Highlighters/Color Scheme preferences of spyder.

This is an initial idea (location of buttons is subject to opinions so please jump in)

![image](https://cloud.githubusercontent.com/assets/3627835/11619029/c2b06c22-9c70-11e5-9945-c66c3cf83cff.png)

Clicking on create new Scheme, or Edit Selected would open something like:
![image](https://cloud.githubusercontent.com/assets/3627835/11619037/e17fd4b2-9c70-11e5-84d6-3853c88509b5.png)

----

I think the combo box should be the global color scheme selector, and every other widget should use that setting.

goanpeca:
The color dialog could also have a "preview editor" to the right, so changes could be rendered in realtime

Also, from here it would be straightforward to export/import color schemes in other formats

ccordoba12:
@goanpeca, awesome! This is really nice :-)

My suggestions:

1. Could it be possible to put all the buttons to the left and the preview to the right? I mean, something like:

        Color scheme               Preview
        Edit selected
        Reset
        Delete
        Create new scheme

2. The Delete button should be disabled when the selected scheme is not a custom one.

3. As you said, later we can add buttons for Import/Export and maybe also change to use a more widely used format for schemes.

goanpeca:
@ccordoba12 

[...]

PR #2861, opened 2015-12-08
PR: Modify color scheme preferences

goanpeca:
Related to #820 

# Description
This PR aims to simplify the user experience for editing the color schemes. 
It also incorporates  the creation/deletion of an unlimited number of custom schemes.

This PR however does not (yet) add the custom schemes to the options in the other plugins (editor etc...) as the idea is to unify the different color schemes into a single global one, located in this updated preferences page. This will be accomplished in a subsequent PR.

The custom name is generated automatically (`custom-N`) but are modifiable. For default schemes colors are modifiable, but the scheme name is not.

# Final Screenshots

image

image


goanpeca:
@nodd, @SylvainCorlay, @blink1073 comments are welcome

spyderlib/config/gui.py
```diff
@@ -162,6 +162,3 @@ def set_default_color_scheme(name, replace=True):
 
 for _name in sh.COLOR_SCHEME_NAMES:
     set_default_color_scheme(_name, replace=False)
-CUSTOM_COLOR_SCHEME_NAME = "Custom"
-set_color_scheme(CUSTOM_COLOR_SCHEME_NAME, sh.get_color_scheme("Spyder"),
-                 replace=False)
```
goanpeca: There is no point in having a custom scheme now

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +916,337 @@ def _save_lang(self):
...
+        TODO:
+        """
+        pass
+
+    def export_scheme(self):
+        """
+        TODO:
+        """
+        pass
+
```
goanpeca: For another PR

spyderlib/utils/syntaxhighlighters.py
```diff
@@ -180,7 +199,8 @@ def setup_formats(self, font=None):
         
     def _check_color_scheme(self, color_scheme):
         if is_text_string(color_scheme):
-            assert color_scheme in COLOR_SCHEME_NAMES
+            pass
+#            assert color_scheme in COLOR_SCHEME_NAMES
         else:
```
goanpeca: This is a very weird check and needs to be improved.

The method should expect one and one type only, but it works if it is a dictionary of color styles or a name of a style, and that is just bad design.

blink1073:
I love the comments on your own PR, very schizophrenic.  I am +100 on this change.

goanpeca:
Hehe, I wanted to save some type, and I am almost sure someone had something to say in that respect,

Would you mind testing it @blink1073 ?

ccordoba12:
A couple of quick suggestions:

1. I don't like the *Create new scheme* button to expand the two columns. I'd like to see it at the bottom but only on the left column (i.e. the *buttons* column :-)
2. In the dropdown menu please change `Scheme` to `Scheme: `

goanpeca:
:+1:

Nodd:
The first thing I thought when viewing the screenshots is that the preview is really a great addition. Just the line numbers are missing.

About the edition of the colors, the entries should be grouped differently, ideally with many group widgets. The minimum would be to group the text values together so that the checkboxes are not all over the place. One group for "text" and one for "background" for example. Or "text", "text highlighting" and "background".

goanpeca:
## Look update

@ccordoba12 

image

@Nodd 
image

---

Plus now we have instant preview, so any changes on an edited theme will reflect in the preview right away (without the need to apply the changes)

goanpeca:
It is looking pretty good already, but I guess the final touch would be to have the preview in the editor dialog, and have it update in every single change inside the actual dialog.

What do you think?

Nodd:
Here is my habitual flow of nitpicks:

* Maybe current cell and current line should be in the Background group, since the are applied to blocks of code and not particular words or characters.
* Background group could be before Highlight group since it's more general (not sure about this one)
* Normal text should be on top
* builtin and keyword should be next of each other

having the preview in the same dialog would be nice too, but you could also put the edition widgets in the main panel next to the current preview. Maybe it's too big to be practical, I'm not sure.

There is too much space between the "Preview" label and the preview widget, I'd completely remove the label since it's already quite obvious.

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +916,408 @@ def _save_lang(self):
...
+                '    def __init__(self):\n'
+                '        bar = 42\n'
+                '        print(bar)\n'
+                )
+        if not scheme_name:
+            scheme_name = self.current_scheme
+        self.preview_editor.setup_editor(linenumbers=True,
+                                         markers=False,
+                                         tab_mode=False,
+                                         font=QFont("Courier New", 10),
```
Nodd: The font should not be hard coded, the configured fond should be sued instead.

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +916,408 @@ def _save_lang(self):
...
+                                         show_blanks=True,
```
Nodd: Same here, the current value should be taken.

Nodd:
After a very quick overview of the code, it seems that all the definitions are stored in the main configuration file.  Ideally each scheme should be in its own file so that it's easier to share, but I'd leave that for another PR (along with import/export)

spyderlib/plugins/configdialog.py
Nodd: Maybe don't include these methods for now since it keeps the code and PRs more readable in the meantime.

spyderlib/plugins/configdialog.py
ccordoba12: +1

spyderlib/plugins/configdialog.py
ccordoba12: +1

ccordoba12:
> I guess the final touch would be to have the preview in the editor dialog

It's fine where it is for me. I think few people will go into the trouble of customizing themes, but most will like to see what a theme looks like instead.

----

> Maybe current cell and current line should be in the Background group

I don't know about this one, but I agree with the other @Nodd suggestions.

spyderlib/plugins/configdialog.py
goanpeca: Yes

spyderlib/plugins/configdialog.py
goanpeca: Yes

goanpeca:
> - Maybe current cell and current line should be in the Background group, since the are applied to blocks of code and not particular words or characters.

I do not agree, cause the idea is to have a highlight group.... independent of what it highlights

> - Background group could be before Highlight group since it's more general (not sure about this one)

I am not sure what general mean? you mean would be changed more often?

> - Normal text should be on top builtin and keyword should be next of each other

Sure, it make sense, I was waiting for you to give me the full order, I just grouped them together...

> - There is too much space between the "Preview" label and the preview widget, I'd completely remove the label since it's already quite obvious.

Sure
> having the preview in the same dialog would be nice too, but you could also put the edition widgets in the main panel next to the current preview. Maybe it's too big to be practical, I'm not sure.

No no no no and no, the idea was to take away so much visual noise from the user when selecting, a theme. The casual user really just selects a default, a power user would either import one, or make one.

The edition dialog is already too large to bring it back to the main menu.

> After a very quick overview of the code, it seems that all the definitions are stored in the main configuration file. Ideally each scheme should be in its own file so that it's easier to share, but I'd leave that for another PR (along with import/export)

:-1: 

To do this I think we need a common use color schemes format, which probably exists and stores stuff as XML or JSON, that is the point of having the Export options, so that you can share them in any format we decide to support, that will come in another PR when those "standards" are pointed out by someone. Any preferences in Spyder is built on the CONF idea, which makes life easy, but restricts us to keep using it, so having separate files would be another layer of possible bugs...

spyderlib/plugins/configdialog.py
goanpeca: It is really harmless code, but also a reminder of something that will/needs to be implemented?

goanpeca:
@nodd, @ccordoba12 

## Look update

image

image

blink1073:
I would leave out the create new button for now, and leave that for another PR, as it is just inviting confusion at the moment.  Otherwise, I am :+1: for merging and iterating.

goanpeca:
@blink1073, I will remove the button but leave the code that does the work ok, as it is perfectly functional?

blink1073:
Yep, makes sense to me.

goanpeca:
@blink1073 by the way did you actually test it :-p ?

blink1073:
Yes, hence my comment about it being confusing :)

goanpeca:
@blink1073 was having some issues with the dialog being too tall, so now is a two columns layout

image

blink1073:
Perfect, thank you, I am no longer confused.  :)

Nodd:
My main concern about the config file full of everything is that when there is a problem in the config we have to reset everything. I'd bet that it's not the color definitions that will cause spyder to fail, but they will be reset anyway along everything else.
Way too many issues are solved by `spyder --reset`, we have something to solve here (on the long term).

It's very nice on 2 columns, but actually I'd change the order again (sorry, I was on my phone before):

* Normal text
* Comment
* String
* Number
* Keyword
* Builtin
* Definition
* Instance (should be renamed to self, or do whe use this color for something else ?)

* Current cell
* Current line
* Occurrence
* Matched parens
* Unmatched parens
* Link

I agree that technically cells and current line are highlighted by the color, but when working on code it's more of a general visual help rather than an important information about what I'm doing in my code. I don't want  to start a big debate, as putting them in highlight works too.

goanpeca:
@Nodd cool, thanks for the input

I understand your config concerns and reset, but that is a whole different discussion, so...

Pas maintenant!

goanpeca:
Gimme a  :+1: and go to bed @Nodd 

image

Nodd:
The screenshots look good to me ! :+1:

spyderlib/config/gui.py
ccordoba12: Ok, no problem with that :-)

spyderlib/config/main.py
```diff
@@ -602,6 +602,8 @@ def is_gtk_desktop():
              {
               'names': ['Emacs', 'IDLE', 'Monokai', 'Pydev', 'Scintilla',
                         'Spyder', 'Spyder/Dark', 'Zenburn'],
+              'custom_names': [],
```
ccordoba12: I think we don't need this option here. Remember that we only add options to `config/main` if we plan to change and/or update them in the future :-)

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +917,419 @@ def _save_lang(self):
...
-                tabs.addTab(self.create_tab(cs_group, def_btn), tabname)
-            else:
-                tabs.addTab(self.create_tab(cs_group), tabname)
-        
+        names = [n for n in self.get_option("names")]
+        custom_names = self.get_option("custom_names")
+
+        # Widgets
+        about_label = QLabel(_("The existing schemes can be edited or you "
+                               "can create new custom color schemes.<br>"))
```
ccordoba12: Please change this text to something like

    Here you can edit the color schemes provided by Spyder or create your own ones by using the options provided below.

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +917,419 @@ def _save_lang(self):
...
+        buttons_preview_layout.addLayout(buttons_layout)
+        buttons_preview_layout.addLayout(preview_layout)
+
+        manage_layout = QVBoxLayout()
+        manage_layout.addWidget(about_label)
+        manage_layout.addLayout(buttons_preview_layout)
+        manage_group = QGroupBox(_("Manage color schemes"))
+        manage_group.setLayout(manage_layout)
+
+        vlayout.addWidget(manage_group)
```
ccordoba12: Could you move these lines (from `manage_layout = QVBoxLayout()` up to here) before `combo_layout = QHBoxLayout()`?

I think this way it'd be easier to read this code because the order it appears here corresponds to how it's shown visually.

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +917,419 @@ def _save_lang(self):
...
-            cs_group.setLayout(cs_layout)
-            if tabname in sh.COLOR_SCHEME_NAMES:
-                def_btn = self.create_button(_("Reset to default values"),
-                                         lambda: self.reset_to_default(tabname))
-                tabs.addTab(self.create_tab(cs_group, def_btn), tabname)
-            else:
-                tabs.addTab(self.create_tab(cs_group), tabname)
-        
+        names = [n for n in self.get_option("names")]
+        custom_names = self.get_option("custom_names")
```
ccordoba12: To avoid using a useless entry in `config/main`, you can use here

```python
self.get_option("custom_names", [])
```

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +917,419 @@ def _save_lang(self):
...
+    @property
+    def current_scheme_index(self):
+        return self.schemes_combo.currentIndex()
+
+    def update_combobox(self):
+        """Recreates the combobox contents."""
+        index = self.current_scheme_index
+        self.schemes_combo.blockSignals(True)
+        names = self.get_option("names")
+        custom_names = self.get_option("custom_names")  # 'custom-n'
```
ccordoba12: Again, `self.get_option("custom_names", [])`

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +917,419 @@ def _save_lang(self):
...
+
+        if custom_names:
+            choices = names + [None] + sorted(custom_names)
+            scheme_choices = list(zip(choices, [a for a in choices]))
```
ccordoba12: Why using the cumbersome `[a for a in choices]` instead of just `choices`? It's the same thing :-) (unless you've forgotten something in the list comprehension).

I think this code (as well as the one below) can be simplified by using a dict comprehension, like this

```python
{c: c.lower() for c in choices}
```

because it seems you want to create a mapping between the scheme names in upper and lower case.

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +917,419 @@ def _save_lang(self):
...
+            combo.insertSeparator(len(names))
+
+        self.schemes_combo.blockSignals(False)
+        self.schemes_combo.setCurrentIndex(index)
+
+    def update_buttons(self):
+        """Updates the enable status of delete and reset buttons."""
+        current_scheme = self.current_scheme.lower()
+        names = [n.lower() for n in self.get_option("names")]
+        delete_visible = current_scheme not in names
```
ccordoba12: This variable should be named `delete_enabled`, not `delete_visible` because we're not hiding the Delete button, but enabling or disabling it.

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +917,419 @@ def _save_lang(self):
...
+                                         color_scheme=scheme_name)
+        self.preview_editor.set_text(text)
+        self.preview_editor.set_language('Python')
+
+    # Actions
+    # -------------------------------------------------------------------------
+    def create_new_scheme(self):
+        """Creates a new color scheme with a custom name."""
+        names = self.get_option('names')
+        custom_names = self.get_option('custom_names')
```
ccordoba12: Again, `self.get_option('custom_names', [])`

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +917,419 @@ def _save_lang(self):
...
+        if not custom_names:
+            custom_names = []
```
ccordoba12: I think these two lines are not necessary

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +917,419 @@ def _save_lang(self):
...
+
+        # Get the available number this new color scheme
+        counter = len(custom_names) - 1
+        custom_index = [int(n.split('-')[-1]) for n in custom_names]
```
ccordoba12: I don't understand what this line does. What kind of name are you expecting in custom names? I mean, why this `n.split('-')[-1]`?

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +917,419 @@ def _save_lang(self):
...
+        for i in range(len(custom_names)):
+            if custom_index[i] != i:
+                counter = i - 1
+                break
+        custom_name = "custom-{0}".format(counter+1)
```
ccordoba12: I see now what format you're giving to custom names. But I don't agree with it. We should leave people to name their schemes as they want, like `Twilight` or `Sublime`, etc.

This way it'd be easier for them to select the scheme they want later :-)

spyderlib/plugins/configdialog.py
ccordoba12: Yep, I agree with @Nodd. Please remove these useless methods. You know where they need to belong in the future, so there's no need to add them for now.

spyderlib/plugins/configdialog.py
goanpeca: :+1:

spyderlib/plugins/configdialog.py
goanpeca: :+1:

spyderlib/plugins/configdialog.py
goanpeca: :+1:

spyderlib/plugins/configdialog.py
goanpeca: agree :+1:

spyderlib/plugins/configdialog.py
goanpeca: agree

spyderlib/plugins/configdialog.py
goanpeca: Custom names for now is 

`custom-N` where N is calculated automatically, so for now there is no possibility of having a custom name....

This will change in the next PR related to this, as it required some extra changes that I considered too much for this pr alone

spyderlib/plugins/configdialog.py
goanpeca: I will implement this, but for now really having it like this make this PR more self contained.

I know how to make it work but want to defer that to a different PR

goanpeca:
Ok, @ccordoba12 fixed all comments, and added editable naming for all the color schemes

image

goanpeca:
Ok, so @ccordoba12 

Only custom schemes can edit the name of the scheme

image

spyderlib/config/main.py
```diff
@@ -586,9 +586,11 @@ def is_gtk_desktop():
               }),
             ('color_schemes',
              {
-              'names': ['Emacs', 'IDLE', 'Monokai', 'Pydev', 'Scintilla',
-                        'Spyder', 'Spyder/Dark', 'Zenburn'],
+              'names': ['emacs', 'idle', 'monokai', 'pydev', 'scintilla',
```
ccordoba12: You need to do a minor update in version in `CONF_VERSION` below (from `23.0.0` to `23.1.0`) for this change to take effect.

goanpeca: Thanks, done

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +916,432 @@ def _save_lang(self):
...
+    @property
+    def current_scheme_index(self):
+        return self.schemes_combo.currentIndex()
+
+    def update_combobox(self):
+        """Recreates the combobox contents."""
+        index = self.current_scheme_index
+        self.schemes_combo.blockSignals(True)
+        names = self.get_option("names", [])
+        custom_names = self.get_option("custom_names")
```
ccordoba12: These two lines are wrong, it should be

```python
names = self.get_option("names")
custom_names = self.get_option("custom_names", [])
```

goanpeca: Nice one!

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +916,432 @@ def _save_lang(self):
...
-                    cs_layout.addWidget(cb_italic, row+1, 3)
-            cs_group.setLayout(cs_layout)
-            if tabname in sh.COLOR_SCHEME_NAMES:
-                def_btn = self.create_button(_("Reset to default values"),
-                                         lambda: self.reset_to_default(tabname))
-                tabs.addTab(self.create_tab(cs_group, def_btn), tabname)
-            else:
-                tabs.addTab(self.create_tab(cs_group), tabname)
-        
+        names = [n for n in self.get_option("names")]
```
ccordoba12: Isn't it easier to just use here

```python
names = self.get_option("names")
```

?

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +916,432 @@ def _save_lang(self):
...
+        self.scheme_editor_dialog = SchemeEditor(parent=self,
+                                                 stack=self.stacked_widget)
+
+        # Widget setup
+        self.scheme_choices_dic = {}
+        about_label.setWordWrap(True)
+        schemes_combo = self.create_combobox(_('Scheme:'),
+                                             [('', '')],
+                                             'selected')
+        self.schemes_combo = schemes_combo.combobox
```
ccordoba12: This is confusing: you have here the `schemes_combo` and `self.schemes_combo` variables, which stand for different things.

Please select a different or more representative for one or the other.

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +916,432 @@ def _save_lang(self):
...
+        create_button = QPushButton(_("Create new scheme"))
+        self.delete_button = QPushButton(_("Delete"))
+        self.preview_editor = CodeEditor(self)
+        self.stacked_widget = QStackedWidget(self)
+        self.reset_button = QPushButton(_("Reset"))
+        self.scheme_editor_dialog = SchemeEditor(parent=self,
+                                                 stack=self.stacked_widget)
+
+        # Widget setup
+        self.scheme_choices_dic = {}
```
ccordoba12: I think it's better to name this `self.scheme_choices_dict` (note the `t` at the end :-)

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +916,432 @@ def _save_lang(self):
...
+        text = ('"""A string"""\n\n'
+                '# A comment\n\n'
+                '# %% A cell\n\n'
+                'class Foo(object):\n'
+                '    def __init__(self):\n'
+                '        bar = 42\n'
+                '        print(bar)\n'
+                )
+        show_blanks = CONF.get('editor', 'blank_spaces')
+        if not scheme_name:
```
ccordoba12: I think `if scheme_name is None` is more readable :-)

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +916,432 @@ def _save_lang(self):
...
+        if dlg.exec_():
+            self.update_combobox()
+            # The +1 is needed because of the separator in the combobox
+            index = (names + custom_names).index(custom_name) + 1
+            self.schemes_combo.setCurrentIndex(index)
+        else:
+            # Delete the config ....
+            custom_names.remove(custom_name)
+            self.set_option('custom_names', custom_names)
+            dlg.delete_color_scheme_stack(custom_name)
```
ccordoba12: When this `else` block is evaluated? When users close `scheme_editor_dialog` without doing any modifications to a newly created custom scheme?

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +916,432 @@ def _save_lang(self):
...
+        scheme_name = self.current_scheme
+
+        # Put the combobox in Spyder by default, when deleting a scheme
+        names = self.get_option('names')
+        self.set_scheme('spyder')
+        self.schemes_combo.setCurrentIndex(names.index('spyder'))
+        self.set_option('selected', 'spyder')
+
+        # Delete from custom_names
+        custom_names = self.get_option('custom_names')
```
ccordoba12: This should be `custom_names = self.get_option('custom_names', [])` to prevent a possible error.

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +916,432 @@ def _save_lang(self):
...
+        dlg = self.scheme_editor_dialog
+        dlg.set_scheme(scheme_name)
+
+    @Slot(str)
+    def reset_to_default(self):
+        """Restore initial values for default color schemes."""
+        # Checks that this is indeed a default scheme
+        scheme = self.current_scheme
+        names = self.get_option('names')
+        if scheme not in names:
```
ccordoba12: This shouldn't be `if scheme in names:` ?

spyderlib/plugins/configdialog.py
goanpeca: Yes I was lowercasing that part and forgot after the logic was changed (which no longer relies on lowercasing)

spyderlib/plugins/configdialog.py
goanpeca: Agreed

spyderlib/plugins/configdialog.py
goanpeca: Sure

spyderlib/plugins/configdialog.py
goanpeca: Sure

spyderlib/plugins/configdialog.py
goanpeca: Exactly, if the user cancels or closes without hitting ok, everything is reverted, that means the new custom scheme is not saved.

This keeps things coherent cause creating a scheme even if the user hits cancel seems very counter intuitive

spyderlib/plugins/configdialog.py
goanpeca: Yes agreed!

spyderlib/plugins/configdialog.py
goanpeca: Yep, I made a mistake :-)

goanpeca:
@ccordoba12 made the fixes, thanks for the review

spyderlib/plugins/configdialog.py
ccordoba12: Nice! Well thought ;-)

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +916,433 @@ def _save_lang(self):
...
+            option = "{0}/{1}".format(scheme_name, key)
+            CONF.remove_option(self.CONF_SECTION, option)
+        CONF.remove_option(self.CONF_SECTION, "{0}/name".format(scheme_name))
+
+        self.update_combobox()
+        self.update_preview()
+
+    def set_scheme(self, scheme_name):
+        """
+        Set the current stack in the dialog to the scheme with 'schema_name'.
```
ccordoba12: Typo here: `schema_name` -> `scheme_name`

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +916,433 @@ def _save_lang(self):
...
+        names = self.get_option('names')
+        if scheme in names:
+            for key in syntaxhighlighters.COLOR_SCHEME_KEYS:
+                option = "{0}/{1}".format(scheme, key)
+                value = CONF.get_default(self.CONF_SECTION, option)
+                self.set_option(option, value)
+
+            option = "{0}/name".format(scheme)
+            value = CONF.get_default(self.CONF_SECTION, option)
+            self.set_option(option, value)
```
ccordoba12: Is really necessary to restore the scheme name too? Since now users are not allowed to change names for default schemes, I'd say these three lines of code are not necessary :-)

spyderlib/utils/syntaxhighlighters.py
```diff
@@ -37,10 +37,29 @@
...
 # Constants
 #==============================================================================
-COLOR_SCHEME_KEYS = ("background", "currentline", "currentcell", "occurence",
-                     "ctrlclick", "sideareas", "matched_p", "unmatched_p",
-                     "normal", "keyword", "builtin", "definition",
-                     "comment", "string", "number", "instance")
+#COLOR_SCHEME_KEYS = ("background", "currentline", "currentcell", "occurence",
+#                     "ctrlclick", "sideareas", "matched_p", "unmatched_p",
+#                     "normal", "keyword", "builtin", "definition",
+#                     "comment", "string", "number", "instance")
```
ccordoba12: Please remove these lines, as they are no longer necessary.

spyderlib/plugins/configdialog.py
goanpeca: Well... I ee no harm in letting people change defaults names if they want to :-p

spyderlib/plugins/configdialog.py
goanpeca: Yep

spyderlib/utils/syntaxhighlighters.py
goanpeca: Sure

ccordoba12:
There's one little error I discovered with live testing: even though you can change a custom scheme name in the scheme editor dialog, that name is not reflected in the schemes combo box of the new preferences pages.

I think it should be changed there too :-)

spyderlib/plugins/configdialog.py
goanpeca: Fixed

spyderlib/plugins/configdialog.py
goanpeca: Fixed

spyderlib/utils/syntaxhighlighters.py
goanpeca: Fixed

goanpeca:
@ccordoba12 fixed the issue and related comments

goanpeca:
@ccordoba12 merged with master....

ccordoba12:
I can't open our Preferences after your update. This is the traceback

```python-traceback
  File "/home/carlos/Projects/spyder/github-repo/spyderlib/spyder.py", line 2614, in edit_preferences
    widget.initialize()
  File "/home/carlos/Projects/spyder/github-repo/spyderlib/plugins/configdialog.py", line 70, in initialize
    self.setup_page()
  File "/home/carlos/Projects/spyder/github-repo/spyderlib/plugins/configdialog.py", line 995, in setup_page
    self.update_combobox()
  File "/home/carlos/Projects/spyder/github-repo/spyderlib/plugins/configdialog.py", line 1030, in update_combobox
    self.scheme_choices_dict[self.get_option('{0}/name'.format(n))] = n
  File "/home/carlos/Projects/spyder/github-repo/spyderlib/plugins/configdialog.py", line 49, in get_option
    return CONF.get(self.CONF_SECTION, option, default)
  File "/home/carlos/Projects/spyder/github-repo/spyderlib/config/user.py", line 380, in get
    raise cp.NoOptionError(option, section)
ConfigParser.NoOptionError: No option 'Custom/name' in section: 'color_schemes'
```

goanpeca:
@ccordoba12 I dont get  that error, can you reset your ini ???

goanpeca:
@ccordoba12 ?

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +916,442 @@ def _save_lang(self):
...
     CONF_SECTION = "color_schemes"
-    
+
     NAME = _("Syntax coloring")
     ICON = ima.icon('eyedropper')
-    
+
     def setup_page(self):
-        tabs = QTabWidget()
         names = self.get_option("names")
```
ccordoba12: Add here `names.pop(names.index(u'Custom'))`

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +916,442 @@ def _save_lang(self):
...
+
+    @property
+    def current_scheme_index(self):
+        return self.schemes_combobox.currentIndex()
+
+    def update_combobox(self):
+        """Recreates the combobox contents."""
+        index = self.current_scheme_index
+        self.schemes_combobox.blockSignals(True)
+        names = self.get_option("names")
```
ccordoba12: Again here `names.pop(names.index(u'Custom'))`

spyderlib/plugins/configdialog.py
```diff
@@ -903,77 +916,442 @@ def _save_lang(self):
...
+        if custom_names:
+            combobox.insertSeparator(len(names))
+
+        self.schemes_combobox.blockSignals(False)
+        self.schemes_combobox.setCurrentIndex(index)
+
+    def update_buttons(self):
+        """Updates the enable status of delete and reset buttons."""
+        current_scheme = self.current_scheme
+        names = self.get_option("names")
```
ccordoba12: And here `names.pop(names.index(u'Custom'))`

ccordoba12:
I added as comments how to fix the bug I'm seeing. The problem is the old `Custom` scheme has not `name` attribute (added by you to the other schemes). And `Custom`  is returned as a valid scheme name when you use `names = self.get_option('names')`, if you are coming from 2.3.

I think the best thing to do (as I did) is to remove `Custom` from `names`. Another option would be to remove all `Custom/*` options from users `spyder.ini`. But that can't be done by a bump in `CONF_VERSION` because there is no `Custom` section in `conf/main.py`. Besides, removing those options must be done once and only once.

Of course, a potential problem will occur if a user decides to name a color scheme `Custom` :-s

goanpeca:
@ccordoba12 thanks for the suggested fix.

There is no potential problem, 'cause all the custom names are stored in a different key.

ccordoba12:
Oh yeah! But now I'm seeing this error when trying to create a new scheme:

```python-traceback
  File "/home/carlos/Projects/spyder/github-repo/spyderlib/plugins/configdialog.py", line 1081, in update_preview
    scheme_name = self.current_scheme
  File "/home/carlos/Projects/spyder/github-repo/spyderlib/plugins/configdialog.py", line 1016, in current_scheme
    return self.scheme_choices_dict[self.current_scheme_name]
KeyError: u''
  File "/home/carlos/Projects/spyder/github-repo/spyderlib/plugins/configdialog.py", line 1055, in update_buttons
    current_scheme = self.current_scheme
  File "/home/carlos/Projects/spyder/github-repo/spyderlib/plugins/configdialog.py", line 1016, in current_scheme
    return self.scheme_choices_dict[self.current_scheme_name]
KeyError: u''
```

ccordoba12:
And this error when trying to delete it (after closing and opening Preferences again after the error mentioned above):

```python-traceback
  File "/home/carlos/Projects/spyder/github-repo/spyderlib/plugins/configdialog.py", line 1158, in delete_scheme
    self.set_scheme('spyder')
  File "/home/carlos/Projects/spyder/github-repo/spyderlib/plugins/configdialog.py", line 1182, in set_scheme
    dlg.set_scheme(scheme_name)
  File "/home/carlos/Projects/spyder/github-repo/spyderlib/plugins/configdialog.py", line 1231, in set_scheme
    self.stack.setCurrentIndex(self.order.index(scheme_name))
ValueError: 'spyder' is not in list
```

goanpeca:
Are this errors because of the changes you suggested?

ccordoba12:
I don't think so. My changes only remove `Custom` as a possible scheme. These errors are in your code.

PR #2879, opened 2015-12-24
Use PyQt5 as default API

flying-sheep:
i think i did this once.

you stated then that it wasn’t ready, but that’s been some time ago and there are now PyQt5 travis tests.

also 3.0 should really ship like that.

### DIFF
--- a/spyderlib/qt/__init__.py
+++ b/spyderlib/qt/__init__.py
@@ -15,6 +15,18 @@ assert os.environ['QT_API'] in ('pyqt5', 'pyqt', 'pyside')
 API = os.environ['QT_API']
 API_NAME = {'pyqt5': 'PyQt5', 'pyqt': 'PyQt4', 'pyside': 'PySide'}[API]
 
+is_old_pyqt = is_pyqt46 = False
+PYQT5 = True
+
+if API == 'pyqt5':
+    try:
+        from PyQt5.QtCore import PYQT_VERSION_STR as __version__
+        from PyQt5 import uic  # analysis:ignore
+    except ImportError:
+        PYQT5 = False
+        API = os.environ['QT_API'] = 'pyqt'
+        API_NAME = 'PyQt4'
+
 if API == 'pyqt':
     # Spyder 2.3 is compatible with both #1 and #2 PyQt API,
     # but to avoid issues with IPython and other Qt plugins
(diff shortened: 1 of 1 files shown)


goanpeca:
@flying-sheep, until there is an official qt5 package in anaconda, I do not think we are doing this.

flying-sheep:
if the API can’t be found, it will fall back to PyQt4, so this change does nothing for people without PyQt5.

the only thing it does is preferring to use PyQt5 if available, which is something that all Qt software should do now.

flying-sheep:
oh it actually doesn’t, i’ll have to reorder things!

ccordoba12:
@goanpeca, I think beta3 is a good point to move to PyQt5 :-)

@flying-sheep, tests for Qt5 are failing in Travis. Please solve the failure.

ccordoba12:
@flying-sheep, I think I fixed the Travis error in master, so please rebase or merge with master to trigger a new build.

However, I think your change breaks Spyder when both PyQt4 and PyQt5 are installed. Please check that in your side.

flying-sheep:
ok, will do

ccordoba12:
Now this breaks **all** our tests in Travis. I really don't know what happens...

flying-sheep:
the non-conda builds break on

`python spyderlib/utils/environ.py`

which hangs indefinitely, because it opens a dialog. why does the travis build call that?

flying-sheep:
ok, the tests work flawlessly here.

unfortunately i neither have any idea what goes wrong with the CI tests.

spyderlib/qt/__init__.py
```diff
@@ -9,12 +9,23 @@
...
+PYQT5 = True
+
+if API == 'pyqt5':
+    try:
+        from PyQt5.QtCore import PYQT_VERSION_STR as __version__
+        from PyQt5 import uic  # analysis:ignore
+    except ImportError:
+        API = os.environ['QT_API'] = 'pyqt'
+        API_NAME = 'PyQt4'
+
```
ccordoba12: I think we should set `PYQT5 = False` here.

ccordoba12: IMHO, it's easier to read and understand.

ccordoba12:
Except for the comment I've just made, I think this ready.

flying-sheep:
fixed

ccordoba12:
Ok, I think this is ready! Merging :-)

flying-sheep:
great :smile:

PR #2886, opened 2015-12-28
Remove official support for PySide

ccordoba12:
This removes references to PySide in our official docs, in our Readme and in `setup.py`.

However, there is (and will always be) support for PySide in `bootstrap.py` and in our Qt shim, in case people want to keep using it.

Reasons to do this:
1. PySide future is uncertain and very little is happening with it currently.
2.  Now that we've moved to support PyQt5 officially (see PR #2879), I want to reduce the number of bindings we support so it be easier to maintain Spyder.
3. PySide support is currently broken in master (Preferences is crashing with it)
4. Nobody has helped to support it (besides Pierre and me).

This was not only my decision, all @spyder-ide/core-developers agreed with it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,9 @@ Spyder is a Python development environment with a lot of features:
 
 * **History log**
 
-Spyder may also be used as a PyQt4/PySide extension library (module 
+Spyder may also be used as a PyQt5/PyQt4 extension library (module 
 `spyderlib`). For example, the Python interactive shell widget used in
-Spyder may be embedded in your own PyQt4/PySide application.
+Spyder may be embedded in your own PyQt5/PyQt4 application.
 
 
 ## Documentation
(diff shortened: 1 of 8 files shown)


ccordoba12:
Letting @picca know about this one.

techtonik:
-1 on that.

techtonik:
> This was not only my decision, all @spyder-ide/core-developers agreed with it.

This link is 404.

techtonik:
Please link to actual discussion.

ccordoba12:
@techtonik, we discussed it in our internal gitter chat, and during a couple of hangouts we had some months ago. So there's no public record of it, sorry.

As I said, Spyder will continue to work with PySide, so there's no problem with that. We only stopped to support it officially, meaning that we (the dev team) are not going to solve more issues related to it.

Support for it will have to come from the community :-)

PR #2890, opened 2015-12-30
Added .idea folder to .gitignore for PyCharm users

dhirschfeld:


Nodd:
That's not very informative. Do you at least have a link explaining why it's useful ?

cjrh:
The PyCharm IDE uses a hidden folder `.idea/` to keep "project information and settings" for the IDE.  If `.idea/` is not added to `.gitignore`, you keep getting asked whether you want to add things like `vcs.xml` to version control.  This is one of the files within the `.idea/` folder.   I'm personally neutral on this PR, even though I've started using PyCharm, but that's the basic background.

dhirschfeld:
As @cjrh mentions it's the folder reserved for PyCharm specific settings so shouldn't be committed to the repo.

This PR simply adds that folder to the `.gitignore` file which improves the development experience for PyCharm users hence lowering the barrier to contribution (in an admittedly small way).

Nodd:
Oh sorry, I thought that it was an issue instead of a pr. I'm fine with it then.

ccordoba12:
I'm fine with this. No problem with our contributors using other IDEs to develop Spyder, though I'd prefer them to use Spyder itself to do it ;-)

cjrh:
@ccordoba12 Think of it as "IDE bootstrapping" 😉

PR #2897, opened 2016-01-03
Fixed typos (thanks to Benjamin Weis)

PierreRaybaut:
reStructured Text --> reStructuredText
occurence --> occurrence
Occurrence --> Occurrence


PierreRaybaut:
@ccordoba12 (and others): it may break an existing "installation" of Spyder based on latest revision (because of the change of name of some settings related to occurrence highlighting, this kind of thing). Anyway, it seems good to fix those shameful typographical errors ;-) (and I'm responsible for most of them...)

PierreRaybaut:
By the way, this PR is required to be merged in order to add German translation to Spyder which is undoubtedly a good addition.

PierreRaybaut:
@ccordoba12:
After some tests, it seems that it's quite harmless.
So unless you have second thoughts on this, I suggest merging this PR.

spyderlib/config/main.py
```diff
@@ -719,7 +719,7 @@ def is_gtk_desktop():
               'zenburn/background':  "#3f3f3f",
               'zenburn/currentline': "#333333",
               'zenburn/currentcell': "#2c2c2c",
-              'zenburn/occurence':   "#7a738f",
+              'zenburn/occurrence':   "#7a738f",
```
ccordoba12: Changing option names requires a *major* bump in `CONF_VERSION` number (present below in this same file)

spyderlib/config/main.py
```diff
@@ -39,7 +39,7 @@
     (_("Patch and diff files"), ('.patch', '.diff', '.rej')),
     (_("Batch files"), ('.bat', '.cmd')),
     (_("Text files"), ('.txt',)),
-    (_("reStructured Text files"), ('.txt', '.rst')),
+    (_("reStructuredText files"), ('.txt', '.rst')),
```
ccordoba12: I don't like this change too much, but I have no problem with it :-)

ccordoba12:
Other than the bump to `CONF_VERSION`, I think this is ready :-)

spyderlib/config/main.py
PierreRaybaut: Honestly, I don't have any opinion on that... the contributor (Benjamin) seems right to point out this minor typo. If it allows Spyder to be translated to a new language, I'm OK with it :)

PR #2913, opened 2016-01-09
Move to use Jupyter imports and remove support for IPython 3

ccordoba12:
We are already requiring Jupyter packages (i.e. `qtconsole` and `nbconvert`) since beta2, so this reflects that in our code too.

Besides, we can express more easily and directly our dependencies by using Jupyter packages instead of the big and monolithic IPython.


ccordoba12:
Letting @picca know about this one.

PR #2919, opened 2016-01-10
PR: Synchronize entry in Working Directory toolbar with console's current working directory

### LINKED ISSUE #2910: Automatically set working directory in console
The current behavior of the current working directory text field in the menu bar might be confusing, especially to beginners and/or people coming from MATLAB. Currently, the working directory is not automatically updated in the console when the working directory is changed in the menu bar (either by directly entering a new path, or navigating up, or by choosing a specific directory). Instead, the user has to click the dedicated "Set as current console's working directory" button.

I suggest to change this behavior such that changes in the working directory set in the menu bar are always automatically propagated to the current (or even all) console.

dzosz:
Changing the directory in the interface tree will now affect consoles' cwd. 

This fixes #2910.

### DIFF
--- a/spyderlib/plugins/workingdirectory.py
+++ b/spyderlib/plugins/workingdirectory.py
@@ -339,6 +339,7 @@ class WorkingDirectory(QToolBar, SpyderPluginMixin):
         self.refresh_plugin()
         if refresh_explorer:
             self.set_explorer_cwd.emit(directory)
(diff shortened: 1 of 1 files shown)


ccordoba12:
Thanks for your contribution!! This correctly fixes the problem reported in #2910 :-)

PR #2952, opened 2016-01-24
PR: Enable 'Save All' if there are files to be saved

jitseniesen:
At the moment 'Save All' is enabled if editor stack 0 has more than one file open and if one of the files opened in editor stack 0 has been changed. However, the 'Save All' action saves all changed files in the current editor stack (not editor stack 0), with no condition on the number of files opened.

This change ensures that 'Save All' is enabled precisely when 'Save All' would do something, that is, when the current editor stack has a file that has been changed.

### DIFF
--- a/spyderlib/plugins/editor.py
+++ b/spyderlib/plugins/editor.py
@@ -1334,11 +1334,10 @@ class Editor(SpyderPluginWidget):
                 action.setEnabled(enable)
                 
     def refresh_save_all_action(self):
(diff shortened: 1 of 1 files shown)


goanpeca:
Thanks @jitseniesen 

LGTM :+1:

ccordoba12:
Yep, looks good to me too. Thanks for your contribution @jitseniesen!

PR #2954, opened 2016-01-25
PR: Support 'alias mode' for generating the .app bundle in OSX

stuarteberg:
On OSX, `py2app` has the ability to create an app in 'alias mode', in which the generated app is just a bare-bones wrapper to launch code from the user's python environment.  This is useful when launching Spyder from an Anaconda environment, for example.  (See ContinuumIO/anaconda-issues#634.)

To generate the alias-mode app, use this command:

```
python create_app.py py2app --alias
```

Then try it out:

```
open dist/Spyder-Py2.app
```

I did not edit the `conda.recipe`, but I think it might be a good idea to do so.  For now, you can find an example recipe that generates the alias-mode app in a separate PR: conda/conda-recipes#560

BTW, To get this to work, I had to make a few bug fixes.  Please review those commits carefully.  In particular, you should scrutinize b81e6d0:

```
    start_ipython_kernel: Add current directory to ipython sys.path as an absolute path, not ''
```

For some reason, the ipython kernel was not starting properly in my build.  Some obscure code in `setuptools.pkg_resources` was raising a `ValueError`, complaining about the current directory.  By changing it to an absolute path instead, the error disappeared.  But using `os.getcwd()` is not quite the same as `''` in this context, because in the resulting ipython kernel, the user won't be able to change the effective `sys.path` by merely typing `cd`.

### DIFF
--- a/create_app.py
+++ b/create_app.py
@@ -86,7 +86,7 @@ INCLUDES = get_stdlib_modules()
 EDIT_EXT = [ext[1:] for ext in EDIT_EXT]
 
 OPTIONS = {
-    'argv_emulation': True,
+    'argv_emulation': False,
     'compressed' : False,
     'optimize': 0,
     'packages': PACKAGES,
--- a/create_app.py
+++ b/create_app.py
@@ -23,6 +23,7 @@ import os
 import os.path as osp
 import subprocess
 import sys
+import re
 
 from IPython.core.completerlib import module_list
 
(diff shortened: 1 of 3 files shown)


ccordoba12:
@stuarteberg, you PR has conflicts. Could you rebase it or merge it with master to update it?

ccordoba12:
We have decided to provide an app structure for Spyder in the `menuinst` package:

https://github.com/ContinuumIO/menuinst/

Thanks for your work Stuart!

PR #2955, opened 2016-01-27
PR: Make backspace move to parent directory in file explorer

goanpeca:
Very simple but needed change to avoid using the mouse :-) when moving within the file explorer

Pressing backspace triggers `go_to_parent_directory` as many OS file managers do.


ccordoba12:
Nice, well thought!! ;-)

PR #2932, opened 2016-01-16
PR: Asynchronous introspection for the Editor

blink1073:
Fixes #2724
Fixes #2407 

----

Perform introspection in separate processes.

Issue #415, opened 2015-02-16
Refactoring support


spyder-issues-migrator:
_From [joan.man...@gmail.com](https://code.google.com/u/105883090728463490757/) on 2010-10-14T06:03:28Z_

Rope can provide a great refactoring support, I note that spyder use it but not for refactoring (masive and safe renaming).

Spyder would provide a great feature if you implement a good support for refactoring.

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=415_

spyder-issues-migrator:
_From [pierre.raybaut](https://code.google.com/u/pierre.raybaut/) on 2010-10-26T04:34:50Z_

**Status:** Accepted  
  **Labels:** -Type-Defect -Priority-Medium Type-Enhancement Priority-Low

spyder-issues-migrator:
_From [pierre.raybaut](https://code.google.com/u/pierre.raybaut/) on 2011-05-16T01:14:51Z_

**Labels:** Cat-Editor

spyder-issues-migrator:
_From [pierre.raybaut](https://code.google.com/u/pierre.raybaut/) on 2011-05-18T00:30:15Z_

**Status:** ContributorNeeded

spyder-issues-migrator:
_From [pierre.raybaut](https://code.google.com/u/pierre.raybaut/) on 2012-03-18T14:11:22Z_

**Status:** HelpNeeded  
  **Labels:** -Type-Enhancement Type-Enh

yw5aj:
Had a couple of friends who switched to PyCharm simply due to the refactoring capability... Felt that this is a really needed feature.

Nodd:
Really needed features really need contributors ! ;)
I agree that it would be great to have, thanks for the reminder.

bmer:
How can I implement this in Spyder? I am really finding that I want this too

CAM-Gerlach:
General note: #1937 is a near duplicate of this one; the nominal scope in that issue's OP is more limited (variable renaming, etc), but while what gets described in the comments (a full refactoring pane, etc.) covers what is being requested here.

cknoll:
This issue has the tag community:Help wanted. I am considering to contribute on this because I would really appreciate that feature but some hints on were to start would be helpful.

ccordoba12:
@cknoll, thanks for offering help. We will address this after PR #4751 is merged, sorry for the confusion.

f4ww4z:
[...]

PR #2964, opened 2016-01-31
PR: Rename variables (by refactoring) with rope

jitseniesen:
I don't think this is quite ready to be merged yet (and given that you are in beta testing, you may prefer to wait till the next release), but I thought it would be good to solicit feedback. I implemented a basic "rename variable" functionality via rope. This is part of issues #415 and #1937.

I'm quite a beginner with Spyder and GitHub and Python GUI so please don't hesitate to comment.
# User experience
1. You put the cursor on the variable that you want to rename:
   rename1
2. You select `Rename` from the `Edit` menu:
   rename2
3. You type the new name in the variable box:
   rename3
4. And Spyder renames the variable: 
   rename4

In this example, both occurrences of the variable `num` in the function `double` are renamed to `arg`. Note that the variable `num` in the main block is not renamed (which I think is the correct behaviour) even though it is highlighted in yellow at step 1.
# Implementation
1. In `Editor.get_plugin_actions()`, create the new menu item and connect it to `Editor.rename_variable()`.
2. In `Editor.rename_variable()`, find the current editor stack and call its `rename_variable()`.
3. In `EditorStack.rename_variable()`, find the current editor and call its `open_rename_variable_dialog()`.
4. In `CodeEditor.open_rename_variable_dialog()`, find the word currently highlighted and ask the user for the new variable name. Return this information, or `None` if the user cancels.
5. Back in `EditorStack.rename_variable()`, assuming that the user did not cancel, find the associated `PluginManager` and call its `rename()`.
6. `PluginManager.rename()` searches through all registered introspection plugins for one that implements a `rename()` member function and calls it.
7. Only `RopePlugin` has a `rename()` member function. It saves the current buffer contents in a temporary file and uses the rope library to get the new code with the renamed variable.
8. The new code is returned to `EditorStack.rename_variable()` which calls `CodeEditor.change_text()` which replaces the old code with the new code.
# Questions and to do
- Everything is done synchronously. This is easier to implement, but it may also be confusing for the user: otherwise it may be that the user tries to rename a variable, nothing seems to happen for a second so the user continues editing, but then the rename function suddenly returns, the contents of the buffer is replaced with the result, and the user loses the edits. However, is there a possibility of interference with other plugin actions executed in a different thread (I don't check `plugin.busy`)?
- If it is better done asynchronously, I guess it should wait until #2932 is merged.
- Should I have used Qt signals?
- Is it okay to save the contents to a temporary file? The rope library seems to be written with files in mind, but it may be possible to work around it.
- The rename functionality in rope comes with many options; I should check them and maybe expose them to the user. In particular, rope can do a rename over multiple files. Could this work with Spyder projects (which, as I understand it, are going to be re-implemented)? 
- After renaming the variable, the cursor is moved to the top of the file which is probably annoying for the user.
- It needs to be made more robust: what if the user clicks on 'Rename' if the cursor is on Python keyword or anything else that is not an identifier? What if the file being edited is not Python? What if the new name the user types in is not a valid identifier? What if rope is not installed? What if in `RopePlugin`, `project` is `None` (can this happen?)? What if rope throws an exception for any reason? All this needs to be handled gracefully.

### DIFF
--- a/spyderlib/plugins/editor.py
+++ b/spyderlib/plugins/editor.py
@@ -1959,7 +1959,7 @@ class Editor(SpyderPluginWidget):
     def rename_variable(self):
         """Open rename variable dialog box."""
         editorstack = self.get_current_editorstack()
-        if editorstack is not None:
+        if editorstack:
             editorstack.rename_variable()
 
     #------ Cursor position history management
--- a/spyderlib/plugins/editor.py
+++ b/spyderlib/plugins/editor.py
@@ -933,6 +933,9 @@ class Editor(SpyderPluginWidget):
         fixindentation_action = create_action(self, _("Fix indentation"),
                       tip=_("Replace tab characters by space characters"),
                       triggered=self.fix_indentation)
+        rename_action = create_action(self, _("Rename"),
+                                      tip=_("Rename a variable"),
+                                      triggered=self.rename_variable)
 
         gotoline_action = create_action(self, _("Go to line..."),
                                         icon=ima.icon('gotoline'),
@@ -1006,7 +1009,8 @@ class Editor(SpyderPluginWidget):
         self.main.debug_toolbar_actions += debug_toolbar_actions
         
         source_menu_actions = [eol_menu, self.showblanks_action,
-                               trailingspaces_action, fixindentation_action]
+                               trailingspaces_action, fixindentation_action,
+                               rename_action]
         self.main.source_menu_actions += source_menu_actions
         
         source_toolbar_actions = [self.todo_list_action,
@@ -1948,6 +1952,13 @@ class Editor(SpyderPluginWidget):
         editorstack = self.get_current_editorstack()
         editorstack.fix_indentation()
                     
+    @Slot()
+    def rename_variable(self):
+        """Open rename variable dialog box."""
+        editorstack = self.get_current_editorstack()
+        if editorstack is not None:
+            editorstack.rename_variable()
+
     #------ Cursor position history management
     def update_cursorpos_actions(self):
         self.previous_edit_cursor_action.setEnabled(
--- a/spyderlib/plugins/editor.py
+++ b/spyderlib/plugins/editor.py
@@ -933,9 +933,11 @@ class Editor(SpyderPluginWidget):
         fixindentation_action = create_action(self, _("Fix indentation"),
                       tip=_("Replace tab characters by space characters"),
                       triggered=self.fix_indentation)
-        rename_action = create_action(self, _("Rename"),
-                                      tip=_("Rename a variable"),
-                                      triggered=self.rename_variable)
+
+        self.rename_action = create_action(self, _("Rename"),
+                                           tip=_("Rename a variable"),
+                                           triggered=self.rename_variable)
+        self.rename_action.setEnabled(False) # cf. register_editorstack()
 
         gotoline_action = create_action(self, _("Go to line..."),
                                         icon=ima.icon('gotoline'),
@@ -1010,7 +1012,7 @@ class Editor(SpyderPluginWidget):
         
         source_menu_actions = [eol_menu, self.showblanks_action,
                                trailingspaces_action, fixindentation_action,
-                               rename_action]
+                               self.rename_action]
         self.main.source_menu_actions += source_menu_actions
         
         source_toolbar_actions = [self.todo_list_action,
@@ -1119,6 +1121,7 @@ class Editor(SpyderPluginWidget):
             editorstack.sig_editor_cursor_position_changed.connect(
                                  self.cursorpos_status.cursor_position_changed)
             editorstack.refresh_eol_chars.connect(self.eol_status.eol_changed)
+            self.rename_action.setEnabled(editorstack.introspector.can_rename())
 
         editorstack.set_help(self.help)
         editorstack.set_io_actions(self.new_action, self.open_action,
--- a/spyderlib/plugins/editor.py
+++ b/spyderlib/plugins/editor.py
@@ -934,9 +934,9 @@ class Editor(SpyderPluginWidget):
                       tip=_("Replace tab characters by space characters"),
                       triggered=self.fix_indentation)
 
-        self.rename_action = create_action(self, _("Rename"),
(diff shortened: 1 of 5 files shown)


ccordoba12:
@jitseniesen, this is very nice!! Thanks for working on it :-)

@Nodd, could you review this one?

spyderlib/plugins/editor.py
```diff
@@ -1922,6 +1926,13 @@ def fix_indentation(self):
         editorstack = self.get_current_editorstack()
         editorstack.fix_indentation()
                     
+    @Slot()
+    def rename_variable(self):
+        """Open rename variable dialog box."""
+        editorstack = self.get_current_editorstack()
+        if editorstack is not None:
```
Nodd: Replace `if editorstack is not None:` by `if editorstack:`

spyderlib/utils/introspection/plugin_manager.py
```diff
@@ -326,6 +326,24 @@ def show_object_info(self, position, auto=True):
...
+        new_name (str): new name of identifier
+        """
+        for plugin in self.plugins.values():
+            try:
+                fun = plugin.rename
+            except AttributeError: 
+                continue
+            return fun(code, position, new_name)
+        else:
+            raise NotImplementedError
```
Nodd: It seems strange to check here if the renaming is possible. If there is no plugin to rename variables, the menu entry should be hidden or disabled. This would also avoid raising `NotImplementedError` in the internal console.

spyderlib/utils/introspection/rope_plugin.py
```diff
@@ -239,6 +239,25 @@ def set_pref(self, key, value):
...
+        temp_filename = 'ropefile-{}.py'.format(os.getpid())
+        temp_module = self.project.root.create_file(temp_filename)
+        temp_module.write(code)
+        from rope.refactor.rename import Rename
+        renamer = Rename(self.project, temp_module, position)
+        changes = renamer.get_changes(new_name)
+        self.project.do(changes)
+        result = temp_module.read()
+        temp_module.remove()
+        return result
```
Nodd: It would be really better to not use a temporary file. I tried to find an alternative but I couldn't find anything quickly so I don't know.
If a temporary file has to be used, it should use a context manager to be sure to remove the file even in case of error, and the name should be more unique that using the pid. See https://docs.python.org/3.5/library/tempfile.html#examples for example.

spyderlib/widgets/sourcecode/codeeditor.py
```diff
@@ -915,6 +915,28 @@ def fix_indentation(self):
             self.setPlainText(text_after)
             self.document().setModified(True)
 
+    def open_rename_variable_dialog(self):
+        """
+        Open rename variable dialog box and return result.
+
+        Returns (source_code, offset, new_name) on success, None otherwise.
+        """
+        from spyderlib.qt.QtGui import QMessageBox, QInputDialog
```
Nodd: Imports are always on top of the file !

spyderlib/widgets/sourcecode/codeeditor.py
```diff
@@ -915,6 +915,28 @@ def fix_indentation(self):
...
+        old_name = self.get_current_word()
+        msg = "Enter new name for '{}':".format(old_name)
+        new_name, ok = QInputDialog.getText(self, 'Rename', msg, text=old_name)
+        if ok and new_name and new_name != old_name:
+            source_code = to_text_string(self.toPlainText())
+            offset = self.get_position('cursor')
+            return (source_code, offset, new_name)
+        else:
+            return None
```
Nodd: `else: return None` feels strange. Removing these lines would work, but just removing `else` and deindenting the return would be clearer.

Nodd:
I had a look at the code but I don't know this part so I don't know if I've missed something.

- I think it's ok to have it synchronous, since all user modifications would be erased
- Where would you use Qt signals ?
- It would be really better to not use temporary files.
- Don't bother with spyder projects, they will be modified and are a bit broken in their current state.

jitseniesen:
@Nodd thanks a lot for having a look so promptly and your comments. I agree with all of them. I'll need some time to act on them and to play with rope in order to avoid the temporary files (which I don't like either).

I saw that go-to-definition uses Qt signals to communicate with the introspector so that's why I wondered whether signals should also be used here. However, go-to-definition is asynchronous while rename-variable is synchronous. Besides, it turns out to be actually quite straightforward to implement rename-variable without signals (once I found out how to do it). So I'm happy not to use Qt signals.

Nodd:
Yes, it's fine with no signals.

jitseniesen:
I addressed @Nodd 's concerns (thanks again), I made some improvements (allow undo, more robust handling of invalid renames), and I rebased (necessary after #2932 which forced me to do some changes). As far as I am concerned, this is ready to be merged.

ccordoba12:
@Nodd, do you have more comments?

Nodd:
After a quick overview the code looks fine but I didn't test it.

jitseniesen:
I rebased the branch so that it merges cleanly after the changes in pyqt.

ccordoba12:
Thanks a lot @jitseniesen! I'll release beta3 and then I'll come back to you to merge this one :-)

ccordoba12:
@jitseniesen, could you rebase or merge with master to solve conflicts?

@Nodd, could you review this one?

jitseniesen:
I will rebase, but I wouldn't mind if this were postponed to 3.1 in the interest of getting 3.0 out of the door as soon as possible.

ccordoba12:
Thanks Jitse!

Nodd:
I can have a look next week. I agree that it's not needed for 3.0. We should set a feature freeze at some point.

ccordoba12:
@jitseniesen, please rebase or merge with master so we can review this one for 3.1 :-)

goanpeca:
Hi @jitseniesen you are probably in unittest mode, just chiming in to see if you plan to continue with this work :-)

ccordoba12:
Yep, it'd be awesome if you could finish this one and also attach it to our projects :-)

ccordoba12:
This needs to be handled through `pyls` now.

PR #3015, opened 2016-02-29
Editor: Fix code completions when working with bootstrap

ccordoba12:
@blink1073, this fixes completions for me when using bootstrap.

I think the solution is the right one because we should only touch `PYTHONPATH` for the `QProcess` that starts `plugin_server` when using bootstrap. In other situations `plugin_server` should be able to find `spyderlib` (the only third-party module it needs) in `site-packages`.

Please review and let me know what you think about this one, given that now you're our Editor completion expert :-)

### DIFF
--- a/spyderlib/utils/introspection/plugin_client.py
+++ b/spyderlib/utils/introspection/plugin_client.py
@@ -12,7 +12,7 @@ import imp
 import sys
 
 # Local imports
-from spyderlib.config.base import debug_print, DEV
+from spyderlib.config.base import debug_print
 from spyderlib.utils.bsdsocket import read_packet, write_packet
 from spyderlib.qt.QtGui import QApplication
 from spyderlib.qt.QtCore import (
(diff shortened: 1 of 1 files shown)


blink1073:
This forces the user to have spyder installed in the virtual/conda environment.  I think the right fix is to use the `dirname` on the results provided by `find_module`, but otherwise leave the code the same.

ccordoba12:
I see, so you want to use the `spyderlib` module from where Spyder is installed to start `plugin_server` with executables of virtual/conda envs.

In that case the solution is to change this

```python
if DEV:
    python_path = imp.find_module('spyderlib')[1]
    env.append("PYTHONPATH=%s" % osp.dirname(python_path))
```

for this

```python
python_path = imp.find_module('spyderlib')[1]
env.append("PYTHONPATH=%s" % osp.dirname(python_path))
```

But we also would need to start `IntrospectionManager` with the executable selected by users in

    Preferences > Console > Advanced Settings

right? Because right now we're not doing that, as can be seen here:

https://github.com/spyder-ide/spyder/pull/2932/files#diff-608bfe6fde6d5ca0e735b2715ebf63c5R404

blink1073:
We also want `rope` or `jedi`, since those might not be environment.  We had punted on using executable setting because @goanpeca said environment handling was in flux.

ccordoba12:
Ok, I understand things better now :-) Your previous code was not working though, but it only requires some minor changes to make it so.

ccordoba12:
@blink1073, I think this is ready now.

blink1073:
Great, thanks!

Issue #2251, opened 2015-03-16
Spyder crashing on very long output


michalgregor:
Spyder Version:  2.3.0rc
Python Version:  3.4.0
Qt Version    :  4.8.6, PyQt4 (API v2) 4.10.4 on Linux
pyflakes >=0.6.0:  0.8.1 (OK)
pep8 >=0.6      :  1.4.6 (OK)
IPython >=0.13  :  None (NOK)
pygments >=1.6  :  1.6 (OK)
sphinx >=0.6.6  :  1.2.3 (OK)
psutil >=0.3    :  1.2.1 (OK)
rope >=0.9.2    :  0.9.3 (OK)
matplotlib >=1.0:  1.4.3 (OK)
pylint >=0.25   :  1.1.0 (OK)

What steps will reproduce the problem?

1. Run a script that produces very long output within a short timespan.
(E.g. theanets do a lot of logging when learning – like hundreds of lines being written into the output all within a few seconds.)
2. Watch spyder crash.

The expected behaviour is not to crash, but display the output.

I am attaching the [trace](https://spyderlib.googlecode.com/issues/attachment?aid=21930000000&name=stacktrace&token=ABZ6GAeRpbBxnaLVvAe4Ss1cfNpmqaFz4Q%3A1426497266268) that gets printed upon the crash. A stack overflow of some sort seems to be going on.

Also – I can fix the problem by commenting out line 276 in spyderlib/widgets/externalshell/baseshell.py:

    def write_output(self):
        self.shell.write(self.get_stdout(), flush=True)
        #QApplication.processEvents()

Is it really absolutely necessary to call processEvents on every write? I mean, surely the updates will get drawn in the next update of the GUI regardless of whether we do this.

I am attaching a [patch](https://spyderlib.googlecode.com/issues/attachment?aid=21930000001&name=spyder.patch&token=ABZ6GAdEJj-u0QrWiBMpGgH1et0FM1ykXw%3A1426497266268) which removes the line in case you may find the solution acceptable.

ccordoba12:
@michalgregor, thanks a lot for taking the time to post your issue report here. Unfortunately, your patch is unreadable because you changed the end-of-lines of `baseshell.py`, and so the patch is equivalent to the entire file! :-)

Could you open a pull request with your changes so that it be easier for us to review and merge?

[...]

PR #3024, opened 2016-03-02
Added QMutex protection to write_output as a workaround for issue #2251

michalgregor:
Pull request regarding a workaround for issue #2251, where spyder crashes on very long output.


ccordoba12:
Could you provide a testing case? I mean, some code that generates a huge output?

michalgregor:
Sure, here goes:
```
for i in range(100000):
		print('{} A rather large number of rather long lines. A rather large number of rather long lines. A rather large number of rather long lines. A rather large number of rather long lines. A rather large number of rather long lines. A rather large number of rather long lines. A rather large number of rather long lines. A rather large number of rather long lines.A rather large number of rather long lines.A rather large number of rather long lines. A rather large number of rather long lines.'.format(i))
```

spyderlib/widgets/externalshell/baseshell.py
```diff
@@ -290,10 +293,29 @@ def get_stderr(self):
...
-        # output. See Issue 2251
-        # QApplication.processEvents()
+
+        while True:
+            self.buffer_lock.lock()
+            messages = self.buffer
+            self.buffer = []
+            self.buffer_lock.unlock()
+
+            if not len(messages): break
```
Nodd: This is clearer without the `len()` and with two lines.

spyderlib/widgets/externalshell/baseshell.py
```diff
@@ -290,10 +293,29 @@ def get_stderr(self):
...
+            else:
```
Nodd: `else` is not needed since the `if` breaks the loop anyway.

spyderlib/widgets/externalshell/baseshell.py
```diff
@@ -290,10 +293,29 @@ def get_stderr(self):
...
+                for msg in messages:
+                    self.shell.write(msg, flush=True)
```
Nodd: It would be more efficient to join the list first, then print. It could be harder to do with Queues however.

spyderlib/widgets/externalshell/baseshell.py
```diff
@@ -290,10 +293,29 @@ def get_stderr(self):
         return self.transcode(qba)
 
     def write_output(self):
+        # if we are already writing something else,
+        # store the present message in a buffer
+        if not self.write_lock.tryLock():
+            self.buffer_lock.lock()
+            self.buffer.append(self.get_stdout())
+            self.buffer_lock.unlock()
+
```
Nodd: Remove this blank line.

Nodd:
Have you considered using a Queue instead of a mutex for the buffer ? It should be simpler.

michalgregor:
Thank you for the feedback. You are right, this should be easier with a Queue – I was not aware that Python had thread-safe queues. I will try to implement your notes as soon as I have some spare time.

michalgregor:
On second thought I have considerable trouble reworking this using Queues. I am not handy with Queues in Python and I can't figure out a way to make sure I get **all** the items from the queue.

I have tried getting items until I get the Empty exception, but that only works if I use the no wait version of get obviously, so essentially if another thread puts in a message, it blocks me, I get and exception, I print what I have and the new message stays in the Queue unprinted).

If you can see a way around this, having more experience with the Queue class, I will happily reimplement this in any way you suggest.

michalgregor:
In the meantime, I have implemented the other suggestions and I have merged the current master.

ccordoba12:
@Nodd, could you please review @michalgregor's work?

Nodd:
The code looks OK, i'll try to test it soon.
I don't know if a test would be useful here, since this bug requires lots of output to be triggered.

michalgregor:
Note that I have attached a [tiny script above](https://github.com/spyder-ide/spyder/pull/3024#issuecomment-191408994) that can be used to test the issue.

Nodd:
I meant automated test with pytest, sorry. I will use your script to test manually.

Nodd:
I tested that it work and the code looks good to me, except one detail:
I just wonder if there can be race conditions: if output arrives between the break L316 and the unlock L320, it will only be printed the next time an ouput is sent. Potentially some text at the end of a script can be hidden.

michalgregor:
Yes, you are right of course, I've overlooked that.

Do you think this new version would be OK? In the new version, the writing thread releases write lock before releasing the buffer. The other thread does another write lock check after adding its stuff to the buffer.

In consequence, either the writing thread intends to continue writing – in which case it does not release the write lock yet – even after releasing the buffer. In this case the second thread can now return.

In the other case the writing thread intends to stop writing at this point, in which case it first releases the write lock, then the buffer and finally it returns. The thread waiting for buffer access can now add its stuff into it and once it is finished with that it can claim the write lock and take over from the original writing thread.

Unfortunately we still cannot completely dispense with the while True loop as far as I can see because if is still possible that another thread will request the buffer while shell.write() is in progress.

Nodd:
I like this version ! My head hurts but I think it works. Thank you for this great job.

ccordoba12:
@michalgregor, thanks a ton for helping us with this!! It was a really critical issue that we didn't know how to solve ;-)

Issue #2884, opened 2015-12-27
template.py editing is cumbersome


nerdoc:
Currently, only the `%(date)s ` and `%(username)s` variables in template.py are replaced.
There should be some generic other variables like `%(projectname)s` and others.
AND - the documentation for that is missing.

What I would do here:
Let the user edit this file (template.py) in the settings (instead of this button "edit template for new modules") and put some explanations there, what the variables are for etc.

ccordoba12:
It's an interesting idea. I'm marking it as doable for a future Spyder version.

nerdoc:
I got the idea from Eclipse. This concept is really good I think. There are more generic templates, for code, and for comments too, depending on which context. PyDev comments of new module look like

    """
    Created on ${date}
    
    @author: ${user}
    """
    ${cursor}

But this is completely configurable within the preferences dialog.

And furthermore, you could add optional default LICENSE headers, that are added to each file at the top.

goanpeca:
if we are going in this direction leta use jinja2.

nerdoc:
Ok - isn't that a bit overkill, using a whole template language with loops etc, for just having a few variables? It could be done by just replacing `${.*}` with the variables?

goanpeca:
We start with few variables and after some months we end up having to maintain and document a "simple home made template language"... not a chance.

I prefer having an "overkill" (well defined proven stable and maintained) template language, that already is included when spyder is installed, than having to put a burden (however minimal) on the developer team

goanpeca:
Eclipse did it like that and that probably makes sense for Eclipse, (which is coded in Java)

I am :-1:  on implementing things like `${user}` for Spyder (Python based), if we can use a well known standard like `{{user}}`

nerdoc:
Ok, FULL ACK. Seems really comprehensible.

Nodd:
Python has a string template, which is proven stable and maintained, and is not overkill. Using jinja to replace a few variables is really overkill, it's better to keep it as an indirect dependency rather than a hard one.
But the main point of the issue is to be able to edit the template directly in the preferences.

goanpeca:
Agree with @nodd, we can use string template ;-)

manmartgarc:
Hi, has there any work done on this? I would love to contribute with this.

ccordoba12:
Nop, you're welcome to work on it. Please read our Contributing guide before starting:

https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md

[...]

PR #3029, opened 2016-03-04
Add shebang line to default template.py

Nodd:
It makes it easier to run scripts outside of spyder on linux.

The template could be further enhanced (add `if __name__ == "__main__"` for example).

### DIFF
--- a/spyderlib/plugins/editor.py
+++ b/spyderlib/plugins/editor.py
@@ -382,10 +382,14 @@ class Editor(SpyderPluginWidget):
         
         # Creating template if it doesn't already exist
         if not osp.isfile(self.TEMPLATE_PATH):
-            header = ['#!/usr/bin/env python' + ('2' if PY2 else '3'),
-                      '# -*- coding: utf-8 -*-',
-                      '"""', 'Created on %(date)s', '',
-                      '@author: %(username)s', '"""', '']
+            if os.name == "nt":
+                shebang = []
+            else:
+                shebang = ['#!/usr/bin/env python' + ('2' if PY2 else '3')]
+            header = shebang + [
+                '# -*- coding: utf-8 -*-',
+                '"""', 'Created on %(date)s', '',
+                '@author: %(username)s', '"""', '']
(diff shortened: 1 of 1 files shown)


spyderlib/plugins/editor.py
```diff
@@ -382,8 +382,10 @@ def __init__(self, parent, ignore_last_opened_files=False):
         
         # Creating template if it doesn't already exist
         if not osp.isfile(self.TEMPLATE_PATH):
-            header = ['# -*- coding: utf-8 -*-', '"""', 'Created on %(date)s',
-                      '', '@author: %(username)s', '"""', '']
+            header = ['#!/usr/bin/env python' + ('2' if PY2 else '3'),
```
ccordoba12: This line should be here only if we are not on Windows, i.e. `if not os.name == 'nt'`

ccordoba12:
`if __name__ == "__main__"` is not for everybody and every situation, so I don't think we should include it.

Nodd:
I agree about `if __name__ == "__main__"`. When #2884 is implemented it will be easily changed by the user.

mdraw:
> The shebang shouldn't appear on windows now.

@Nodd Why would you leave out the shebang on Windows? The shebang line is also interpreted on Windows, see https://docs.python.org/3/using/windows.html#shebang-lines and https://www.python.org/dev/peps/pep-0397/.

Nodd:
@mdraw I left it out because the boss (@ccordoba12) said so ! But since it's interpreted, I'll add it again.

ccordoba12:
I'm -1 on adding a shebang to Windows scripts because:

1. It only works in Python 3.
2. It could confuse Windows users if it comes as `/usr/bin/env python`. I mean, it would (most probably) make them believe that that's an error or something. Experienced users (like @mdraw) can easily add the shebang themselves :-)

Nodd:
I have some counter arguments, but since it's a detail and it's so easy to change I think it's not worth the time. I'll left it as is.

Issue #3026, opened 2016-03-03
File save as: file is unavailable


dacoex:
## Description of your problem

**What steps will reproduce the problem?**
1. Open file in Editor
2. File Menu -> Save as 
3. The dialogue betlow appears. 
4. If I click on Yes, the file is never saved
5. If click on NO, the file is kept in Editor

![spyder_file_unavailable](https://cloud.githubusercontent.com/assets/8660305/13496443/80ae9bda-e150-11e5-9190-e6b277589d18.png)

Why does this come up?

**What is the expected output? What do you see instead?**

Expected / desired:
- The file with the new name ("save as") shall be kept in editor.
- A dialogue may ask if the old file can be closed. The file must not be deleted from the disk.

**Please provide any additional information below**
## Versions and main components
- Spyder Version: 3.0.0
- Python Version: 3.5
- Operating system: Windows 7
## Dependencies

Please go to the menu entry `Help > Optional Dependencies` (or
`Help > Dependencies`), press the button `Copy to clipboard`
and paste the contents below:

```
IPython >=3.0      :  4.0.1 (OK)
jedi >=0.8.1;<0.9.0:  0.8.1 (OK)
matplotlib >=1.0   :  1.5.0 (OK)
numpy >=1.7        :  1.9.3 (OK)
pandas >=0.13.1    :  0.17.1 (OK)
pep8 >=0.6         :  1.6.2 (OK)
pyflakes >=0.6.0   :  0.9.2 (OK)
pygments >=1.6     :  2.0.2 (OK)
pylint >=0.25      :  1.5.1 (OK)
qtconsole >=4.0    :  4.1.1 (OK)
rope >=0.9.4       :  0.9.4-1 (OK)
sphinx >=0.6.6     :  1.3.3 (OK)
sympy >=0.7.3      :  0.7.6.1 (OK)
zmq >=13.0.0       :  15.1.0 (OK)
```

ccordoba12:
@stonebig, can you reproduce this problem?

stonebig:
curious, I would imagine the file is open on internet, on a zip directory or in a read-only directory ?

dacoex:
on the local drive.

stonebig:
Is it a file in a normal directory or in a zipped directory ? which exact version of Spyder 3.0.0 ?

dacoex:
>Is it a file in a normal directory or in a zipped directory ? 

normal, nothing compressed. Windows 7.

[...]

PR #3039, opened 2016-03-06
a better wheel number for coming beta3

stonebig:
would fix this https://github.com/spyder-ide/spyder/issues/3026#issuecomment-192918304


ccordoba12:
@stonebig, our version number is going to be changed when we release beta3, not before that :-)

stonebig:
ok

ccordoba12:
Thanks for understanding :-)

PR #3043, opened 2016-03-08
Copying when nothing is selected no longer affects the clipboard.

### LINKED ISSUE #2835: Control+C should not copy if it has nothing selected on the editor
# Description

Select something on the editor and `Ctrl+C` it... then with nothing selected accidentally typing `Ctrl+C` will replace the content of the clipboard with an empty string... this is annoying.

If nothing is selected, the behavior should be to no put an empty string in the clipboard...

stephenshank:
Fixes #2835 

----

My first attempt at contributing to open source... constructive criticism welcome!

### DIFF
--- a/spyderlib/widgets/sourcecode/base.py
+++ b/spyderlib/widgets/sourcecode/base.py
@@ -492,7 +492,8 @@ class TextEditBaseWidget(QPlainTextEdit, BaseEditMixin):
         Reimplement Qt method
         Copy text to clipboard with correct EOL chars
         """
-        QApplication.clipboard().setText(self.get_selected_text())
+        if len(self.get_selected_text()) > 0:
+            QApplication.clipboard().setText(self.get_selected_text())
     
     def toPlainText(self):
         """
--- a/spyderlib/widgets/sourcecode/base.py
+++ b/spyderlib/widgets/sourcecode/base.py
@@ -492,7 +492,7 @@ class TextEditBaseWidget(QPlainTextEdit, BaseEditMixin):
         Reimplement Qt method
         Copy text to clipboard with correct EOL chars
         """
-        if len(self.get_selected_text()) > 0:
+        if self.get_selected_text():
             QApplication.clipboard().setText(self.get_selected_text())
     
     def toPlainText(self):


spyderlib/widgets/sourcecode/base.py
```diff
@@ -492,7 +492,8 @@ def copy(self):
         Reimplement Qt method
         Copy text to clipboard with correct EOL chars
         """
-        QApplication.clipboard().setText(self.get_selected_text())
+        if len(self.get_selected_text()) > 0:
+            QApplication.clipboard().setText(self.get_selected_text())
```
goanpeca: Thanks for the contribution

In python saying 

`if len(self.get_selected_text()) > 0` is the same as `  if len(self.get_selected_text()):`

So we use the second version (as it is common practice in Python)

ccordoba12:
@stephenshank, thanks a lot for your contribution! What you did is exactly what it needed to be done, so well job :-)

What @goanpeca meant is that removing `> 0` in the code you added is all we require from you to merge this because in Python there's no need of that further check.

I think even `if self.get_selected_text()` is a right solution because if there's no selected text, then `self.get_selected_text()` will return an empty string, and in Python `if ''` (like `if []` for empty lists) returns `False`.

stephenshank:
Thanks to both of you for your feedback! After experimenting with `%timeit`, I ended up going with 
`if self.get_selected_text():`
since this gave the most favorable performance among the three variations (and still resolved the issue with clipboard, of course).

ccordoba12:
Ok, your first contribution to open source is going in ;-)

ccordoba12:
Please notice that this fix is going to be available since Spyder 3.0 beta3, and not in the 2.3.x series (2.3.9 is the next planned release :-)

Nodd:
Welcome to a whole new world ! :smile:

I jusst have a remark about your reasoning. In non critical parts, the correct code isn't usually determined by `%timeit` (even if this result is interesting in itself). In that case performance matters less that maintainability and coherency. Having a better performance is of course a plus, but the user will never notice if Ctrl+C is 0.1ms slower or faster.

`if self.get_selected_text():` is more *pythonic*, which means that it's the less surprising form to someone used to python, so it will be easier in the future to understand, debug or modify.

stephenshank:
Great news! Thanks so much! I can see that I clearly have a lot to learn from this community... you can expect to see more pull requests from me in the future.

PR #3056, opened 2016-03-12
sitecustomize: import matplotlib only when needed

PierreRaybaut:
This PR simply aims at fixing Issue 2257 without importing ``matplotlib`` systematically.

It's really not a good thing to import systematically ``matplotlib`` when executing *any* script from Spyder. Importing ``matplotlib`` is not innocuous at all: it clutters unnecessarily ``sys.path`` and ``sys.modules`` and it affects execution speed. Changes made on ``sitecustomize.py`` should be done very carefully, keeping in mind that executing a script within or outside Spyder should be as close as possible.

So, in this case, I strongly recommend to import ``matplotlib`` only when necessary, that is (according to Issue 2257) when the following criteria are met:

  * IPython console
  * Python 3
  * Linux

In the proposed commit, I just need confirmation that ``IS_IPYTHON`` is the right test for checking if we are "in our IPython consoles" (as stated in the comment above).

### DIFF
--- a/spyderlib/widgets/externalshell/sitecustomize.py
+++ b/spyderlib/widgets/externalshell/sitecustomize.py
@@ -20,6 +20,8 @@ import shlex
 
 
 PY2 = sys.version[0] == '2'
+IS_EXT_INTERPRETER = os.environ.get('EXTERNAL_INTERPRETER', '').lower() == "true"
+IS_IPYTHON = os.environ.get("IPYTHON_KERNEL", "").lower() == "true"
 
 
 #==============================================================================
@@ -168,8 +170,6 @@ if os.name == 'nt':
 #==============================================================================
 # Settings for our MacOs X app
 #==============================================================================
-IS_EXT_INTERPRETER = os.environ.get('EXTERNAL_INTERPRETER', '').lower() == "true"
-
 if sys.platform == 'darwin':
     from spyderlib.config.base import MAC_APP_NAME
     if MAC_APP_NAME in __file__:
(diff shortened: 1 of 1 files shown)


spyderlib/widgets/externalshell/sitecustomize.py
```diff
@@ -324,8 +325,6 @@ def qt_nt_inputhook():
 #==============================================================================
 # Matplotlib settings
 #==============================================================================
-IS_IPYTHON = os.environ.get("IPYTHON_KERNEL", "").lower() == "true"
-
 if matplotlib is not None:
```
ccordoba12: @PierreRaybaut, your change will break this entire block, which is needed to set the Matplotlib backend in our Python consoles (in Windows and Linux).

ccordoba12: And Mac too (silly me :-)

ccordoba12:
> It's really not a good thing to import systematically matplotlib when executing any script from Spyder.

But I thought import things in `sitecustomize` didn't expose them in the Python/IPython interpreter directly. I mean, I thought everything imported in `sitecustomize` lived under the `sitecustomize` namespace (or something like that :-)

Besides, there is the problem I posted above: not importing `matplotlib` globally in `sitecustomize` prevents setting the right backend in our Python consoles. And god knows I worked a lot on making that happen in the right way, so I'd really prefer not to touch that (unless you come with a better solution :-)

ccordoba12:
@PierreRaybaut, I verified that `matplotlib` is not exposed in our consoles main namespace.

It's true that there are a lot of things present in `sys.modules` (and `matplotlib` is not the only one. `pandas`, `numpy` and `PIL` are there too). But:

1. We haven't heard of any complaints from our users that things run differently in Spyder and out of it (except for PyQt things :-).
2. As long as these modules are not exposed in the main namespace, I don't see a problem with that. Furthermore, importing them in sitecustomize let us perform several adjustments (like setting the Matplotlib backend or Pandas encoding) that otherwise I don't know how we could do it.

For those reasons, I'd say we should close this pull request. I hope you agree :-)

ccordoba12:
@PierreRaybaut, I'm going to close this pull request since I haven't heard from you in more than a month.

Please reopen it if you consider you have a new approach to solve this problem.

PR #3060, opened 2016-03-13
Start testing with Qt5 on Windows

ccordoba12:


ccordoba12:
Ok, this is green now, so merging!

PR #3086, opened 2016-03-25
Fix interrupt handling on Windows

blink1073:
The Windows `SetConsoleCtrlHandler` was only sending one argument to the function.

### DIFF
--- a/spyderlib/app/spyder.py
+++ b/spyderlib/app/spyder.py
@@ -285,8 +285,7 @@ class MainWindow(QMainWindow):
                 import win32api
                 win32api.SetConsoleCtrlHandler(signal_handler, True)
             except ImportError:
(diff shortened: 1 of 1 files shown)


ccordoba12:
Now that we are here, I'd like to ask: @blink1073 you added pywin32 to close completion processes after a crash, right? So my question is: should we make it a hard dependency? At least now, Spyder doesn't start on Windows if pywin32 is not present.

blink1073:
Yep, I think requiring `pywin32` is a good idea.

ccordoba12:
Could you elaborate a bit more?

blink1073:
Oh, I see what you mean, updating.

blink1073:
Fixed.

ccordoba12:
Thanks @blink1073 :-)

PR #3095, opened 2016-04-07
Add some indentation tests

Nodd:
I added some tests to check for the indentation function.

It is motivated by a new but that was introduced lately, the cursor is placed at the same column as an opened parenthesis even if no text is entered after the parenthesis (this is my last commented test case).

## TODO:
- [ ] Add more non-regressiàon tests
- [ ] Correct failed tests
- [ ] Add new failed tests and correct them

This way of testing works but isn't practical at all, it would be really better to use libs like pytest for this. Those tests don't need the QApplication to run, only its creation so it doesn't have the same problem other tests have. Some hints were stated in #2268

### DIFF
--- a/spyderlib/widgets/sourcecode/codeeditor.py
+++ b/spyderlib/widgets/sourcecode/codeeditor.py
@@ -2924,44 +2924,6 @@ class TestWidget(QSplitter):
         self.classtree.set_current_editor(self.editor, filename, False, False)
 
 
-class TestIndent(CodeEditor):
-    def __init__(self, parent=None):
-        CodeEditor.__init__(self, parent)
-        self.setup_editor(linenumbers=True, markers=True, tab_mode=False,
-                          font=QFont("Courier New", 10),
-                          show_blanks=True, color_scheme='Zenburn',
-                          language='Python')
-        self.tmp_filename = "test_spyder.tmp"
-
-    def test_fix_indent(self):
-            text = self._get_indent_fix("this_tuple = (1, 2)\n")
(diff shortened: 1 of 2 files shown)


goanpeca:
@nodd, dont we need a QApplication to start a widget?

Nodd:
You need a QApplication but not necessarily .exec_().

Nodd:
I understood the difference I had between my tests and the editor, I forgot to set the language. Now I can focus on fixing the tests.
When #3120 I'll integrate these tests with pytest.

goanpeca:
Sure, I am waiting on @ccordoba12 on doing some cleaning of the travis scripts to get the tests going :-p,

Nodd:
@goanpeca I updated the PR to use pytest (without pytest-qt, it was not necessary in my case). Any comments ?

ccordoba12:
@Nodd, merging with master should fix the failures on Travis and AppVeyor :-)

Is there anything else to do here?

Nodd:
I should add the case from #3214 as KnownFail.
I should also try to fix the failing tests, but that can be left for another PR (by someone else maybe :watch: )

ccordoba12:
Please add the failure of #3214 so I can merge it :-)

You can add other tests later.

Nodd:
Here you go !
The problem is not about adding test case, it's about fixing the failing ones !

Nodd:
I rebased on master, I hope it'l fix the failing tests.

ccordoba12:
@Nodd, is this one ready to be merged?

Nodd:
@ccordoba12 Yes, I merge it.

PR #3102, opened 2016-04-11
Note failure to import `qtpy` in message about missing PyQt libraries

johnyf:
I installed `PyQt5` and `spyder` from source, but `setuptools` did not install `qtpy`. When opening `spyder`, the failure message suggests that `PyQt*` is missing, whereas it was only `qtpy` that was missing. This behavior can be reproduced with `pip uninstall qtpy` in an environment where `spyder` is already working correctly.


ccordoba12:
Thanks a lot for this change @johnyf, it seems sensible.

PR #3105, opened 2016-04-11
Remove our internal Qt shim in favor of QtPy

### LINKED ISSUE #3100: Migrate to qtpy: Remove internal Qt shim used by Spyder.
Remove internal Qt shim used by Spyder and replace by qtpy package.

ccordoba12:
Fixes #3100 

----

Now `qtpy >= 1.0.1` is a hard dependency of Spyder (and not just inherited through QtAwesome).

Pinging @PierreRaybaut, @stonebig and @ghisvail so that they are aware of this change.


ghisvail:
Thanks for the heads-up

ccordoba12:
@ghisvail, no problem :-)

@PierreRaybaut, external projects that use `spyderlib.qt` need to use `qtpy` instead now because this PR also removes that module.

ccordoba12:
I forgot to ping @picca in case this is important for the Spyder-based projects he maintains.

picca:
thanks to ghislain vaillant we now have qtpy inside Debian.

So go ahead :))

ghisvail:
btw @picca I have the latest releases of `python-qtpy` and `python-qtawesome` waiting on mentors ;-)

goanpeca:
I would add as a comment that, the internal qt shim **spyderlib.qt** does not translate directly to **qtpy**. QtPy gives priority to Qt5, hence code is written as if it was using Qt5 and the shim will enforce this, so code like:

```from spyderlib.qt.QtGui import QApplication``` that works with the spyder internal shim

```from qtpy.QtGui import QApplication``` fails, as in Qt5 widgets have been moved to QtWidgets, so the correct import is

```from qtpy.QtWidgets import QApplication```

ccordoba12:
Right, thanks @goanpeca for the clarification ;-)

PierreRaybaut:
Great work guys! Judging by the maturity of `qtpy`, I won't be forced to copy/paste my old `spyderlib.qt` package on every Qt-based project as I did until now (e.g. in `guidata`, `guiqwt` and other internal projects).

@ccordoba12 : I don't think any of my other Python package is relying on `spyderlib.qt`.

ccordoba12:
Glad you approve this move @PierreRaybaut! :-)

goanpeca:
:+1: Awesome :-)

PR #3120, opened 2016-04-19
PR: Add pytests for array builder, code coverage and quantified code

### LINKED ISSUE #2268: Start testing with pytest/pytest-qt and coverage for Spyder

### LINKED ISSUE #3122: Test array builder widget

goanpeca:
Fixes #2268 
Fixes #3122

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,10 +39,9 @@ before_install:
 
 install:
   - ./continuous_integration/travis/install.sh;
-  - if [ "$USE_CONDA" = true ]; then
-      source activate test-environment;
-    fi
   - export PATH="$HOME/miniconda/bin:$PATH";
+  - test $USE_CONDA = true \
+    && source activate test-environment
 
 script:
   - ./continuous_integration/travis/build_test.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,6 +41,8 @@ install:
   - ./continuous_integration/travis/install.sh;
   - if [ "$USE_CONDA" = true ]; then
       source activate test-environment;
+    fi
+  - if [ "$USE_CONDA" = true ]; then
       conda install pytest pytest-cov;
     else
       pip install pytest pytest-cov;
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,12 +42,6 @@ install:
   - if [ "$USE_CONDA" = true ]; then
       source activate test-environment;
     fi
-  - if [ "$USE_CONDA" = true ]; then
-      conda install pytest pytest-cov;
-    else
-      pip install pytest pytest-cov;
-    fi
-  - pip install pytest-qt
   - export PATH="$HOME/miniconda/bin:$PATH";
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,8 +40,9 @@ before_install:
 install:
   - ./continuous_integration/travis/install.sh;
   - export PATH="$HOME/miniconda/bin:$PATH";
-  - test $USE_CONDA = true \
-    && source activate test-environment
+  - if [ “${USE_CONDA}" = “true” ]; then
+        source activate test-environment
+    fi
 
 script:
   - ./continuous_integration/travis/build_test.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,4 +51,4 @@ script:
   - ./continuous_integration/travis/modules_test.sh
   - python runtests.py
   - coveralls
-  - cd doc; make html linkcheck
+#  - cd doc; make html linkcheck
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,8 +40,8 @@ before_install:
 install:
   - ./continuous_integration/travis/install.sh;
   - export PATH="$HOME/miniconda/bin:$PATH";
-  - if [ “${USE_CONDA}" = “true” ]; then
-        source activate test-environment
+  - if [ "${USE_CONDA}" = "true" ]; then
+        source activate test-environment;
     fi
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,10 +43,12 @@ install:
   - if [ "${USE_CONDA}" = "true" ]; then
         source activate test-environment;
     fi
+  - pip install coveralls
 
 script:
   - ./continuous_integration/travis/build_test.sh
   - ./continuous_integration/travis/run_test.sh
   - ./continuous_integration/travis/modules_test.sh
   - python runtests.py
+  - coveralls
   - cd doc; make html linkcheck
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,11 +39,18 @@ before_install:
 
 install:
   - ./continuous_integration/travis/install.sh;
-  - if [ "$USE_CONDA" = true ]; then source activate test-environment; fi
+  - if [ "$USE_CONDA" = true ]; then
+      source activate test-environment;
+      conda install pytest pytest-cov;
+    else
+      pip install pytest pytest-cov;
+    fi
+  - pip install pytest-qt
   - export PATH="$HOME/miniconda/bin:$PATH";
 
 script:
   - ./continuous_integration/travis/build_test.sh
   - ./continuous_integration/travis/run_test.sh
   - ./continuous_integration/travis/modules_test.sh
+  - python runtests.py
   - cd doc; make html linkcheck
--- a/runtests.py
+++ b/runtests.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © 2009- The Spyder Development Team
+# Licensed under the terms of the MIT License
+#
+
+"""
+"""
+
+# Third party imports
+import qtpy
+import pytest
+
+
+def main():
+    pytest.main(['-x', 'spyderlib',  '-v', '-rw', '--durations=10',
+                 '--cov=spyderlib', '--cov-report=term-missing'])
+
+
+if __name__ == '__main__':
+    main()
--- a/runtests.py
+++ b/runtests.py
@@ -5,6 +5,7 @@
 #
 
 """
+File for running tests programmatically.
 """
 
 # Third party imports
@@ -13,6 +14,9 @@ import pytest
 
 
 def main():
+    """
+    Run pytest tests.
+    """
     pytest.main(['-x', 'spyderlib',  '-v', '-rw', '--durations=10',
                  '--cov=spyderlib', '--cov-report=term-missing'])
 
--- a/spyderlib/widgets/arraybuilder.py
+++ b/spyderlib/widgets/arraybuilder.py
@@ -143,7 +143,7 @@ class NumpyArrayTable(QTableWidget):
                 else:
                     value = '0'
 
-                if value.strip() == '':
+                if not value.strip():
                     value = '0'
 
                 text.append(' ')
@@ -206,11 +206,11 @@ class NumpyArrayDialog(QDialog):
         if inline:
             self._button_help.setToolTip(self._help_inline)
             self._text = NumpyArrayInline(self)
-            self.widget = self._text
+            self._widget = self._text
         else:
             self._button_help.setToolTip(self._help_table)
             self._table = NumpyArrayTable(self)
-            self.widget = self._table
+            self._widget = self._table
 
         style = """
             QDialog {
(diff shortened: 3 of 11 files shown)


.travis.yml
```diff
@@ -39,10 +39,19 @@ before_install:
...
-  - if [ "$USE_CONDA" = true ]; then source activate test-environment; fi
+  - if [ "$USE_CONDA" = true ]; then
+      source activate test-environment;
+    fi
+  - if [ "$USE_CONDA" = true ]; then
+      conda install pytest pytest-cov;
+    else
+      pip install pytest pytest-cov;
+    fi
+  - pip install pytest-qt
```
ccordoba12: I think it should be better to move all this instructions to install.sh. Maybe that helps to solve the problem.

goanpeca: Ok, will try now

goanpeca:
@ccordoba12 this is running as soon as we have a package for **pytest-qt** on the **spyder-ide** channel

goanpeca:
@Nodd any comments?

spyderlib/widgets/arraybuilder.py
```diff
@@ -140,8 +143,9 @@ def text(self):
                 else:
                     value = '0'
 
-                if value == '':
+                if value.strip() == '':
```
Nodd: `if not value.strip():` maybe ? (this is just style, but since you're changing it...)

spyderlib/widgets/arraybuilder.py
```diff
@@ -256,7 +260,12 @@ def keyPressEvent(self, event):
             QDialog.keyPressEvent(self, event)
 
     def event(self, event):
-        if event.type() == QEvent.KeyPress and event.key() == Qt.Key_Tab:
+        """
+        Qt Override
+
+        Usefull when in line edit mode.
+        """
+        if event.type() == QEvent.KeyPress and event.key() == Qt.Key_Tab:  # pragma: nocover
```
Nodd: The pragma comment skips the coverage, right ? I don't think it should be there, because we could check the return value when given a specific event.

spyderlib/widgets/arraybuilder.py
```diff
@@ -202,11 +206,11 @@ def __init__(self, parent, inline=True, offset=0):
         if inline:
             self._button_help.setToolTip(self._help_inline)
             self._text = NumpyArrayInline(self)
-            self._widget = self._text
+            self.widget = self._text
```
Nodd: This attribute should stay private even if it's used in the tests. The tests are internal, it's totally fine if they depend on the implementation.
Being able to test correctly is one of the reason that there is no real private attribute in python, so that even private attributes can be tested.

spyderlib/widgets/tests/test_array_builder.py
```diff
@@ -0,0 +1,204 @@
...
+from pytestqt import qtbot  # analysis:ignore
+
+# Local imports
+from spyderlib.widgets.arraybuilder import NumpyArrayDialog
+
+
+class TestNumpyArrayBuilder:
+    def test_array_inline_array(self, qtbot):  # analysis:ignore
+        dlg = NumpyArrayDialog(inline=True)
+        dlg.show()
```
Nodd: Just to be sure, the dialog needs to be shown for the events to be active ?

spyderlib/widgets/tests/test_array_builder.py
```diff
@@ -0,0 +1,204 @@
...
+        qtbot.addWidget(dlg)
+
+        qtbot.keyClicks(dlg.widget, '1 2 3  4 5 6')
+        qtbot.keyPress(dlg.widget, Qt.Key_Return)
+
+        expected_value = 'np.array([[1, 2, 3],\n          [4, 5, 6]])'
+        value = dlg.text()
+        assert expected_value == value
```
Nodd: I would put them in the reverse order, I usually check if what I have is what I want, not the other way around.

spyderlib/widgets/tests/test_array_builder.py
```diff
@@ -0,0 +1,204 @@
...
+    def test_array_table_array(self, qtbot):  # analysis:ignore
+        dlg = NumpyArrayDialog(inline=False)
+        dlg.show()
+        qtbot.addWidget(dlg)
+
+        qtbot.keyClick(dlg.widget, Qt.Key_1)
+        qtbot.keyClick(dlg.widget, Qt.Key_Tab)
+        qtbot.keyClick(dlg.widget, Qt.Key_2)
+        qtbot.keyClick(dlg.widget, Qt.Key_Tab)
+        qtbot.keyClick(dlg.widget, Qt.Key_Backtab)  # Needed hack
```
Nodd: I don't understand the comment. What hack is used ?

spyderlib/widgets/tests/test_array_builder.py
```diff
@@ -0,0 +1,204 @@
...
+        dlg = NumpyArrayDialog(inline=True)
+        dlg.show()
+        qtbot.addWidget(dlg)
+
+        qtbot.keyClicks(dlg.widget, '4 5 6')
+        qtbot.keyPress(dlg.widget, Qt.Key_Return, modifier=Qt.ControlModifier)
+
+        expected_value = 'np.matrix([4, 5, 6])'
+        value = dlg.text()
+        assert expected_value == value
```
Nodd: I think that for short values there is no need to use explicit variables, as it "hides" the case being tested. In this case I prefer:
```python
        assert dlg.text() == 'np.matrix([4, 5, 6])'
```

Test files tent to be very long, if we can shorten the tests it can be a big win.

spyderlib/widgets/tests/test_array_builder.py
```diff
@@ -0,0 +1,204 @@
...
+        qtbot.keyPress(dlg.widget, Qt.Key_Return)
+
+        expected_value = 'np.array([[1, 2, 3],\n          [4, 5, 6]])'
+        value = dlg.text()
+        assert expected_value == value
+
+    def test_array_inline_matrix(self, qtbot):  # analysis:ignore
+        dlg = NumpyArrayDialog(inline=True)
+        dlg.show()
+        qtbot.addWidget(dlg)
```
Nodd: Those 3 lines are used in each test. It's possible to use a fixture to set them automatically for each test. These 3 lines would be replaced by an additional `dlg` argument in the function call.

Nodd:
@goanpeca I added some comments inline, they are actually style comments open to discussion. We have to settle on a few rules to be consistent in the tests.
This is a nice work you did there, pytestqt looks very easy to use.

spyderlib/widgets/arraybuilder.py
goanpeca: Sure ;-)

spyderlib/widgets/tests/test_array_builder.py
goanpeca: Yes you are absolutely right!

spyderlib/widgets/tests/test_array_builder.py
goanpeca: This is related to the fact that during tests the behavior is not  identical to when launching the actual widget.
In theory when pressing Tab a new cell is created and that becomes the current cell.

In the tests the new cell is created but the current cell is the one after the one it should.
I might need to dig deeper but could not find a solution that did not break the widget

spyderlib/widgets/tests/test_array_builder.py
goanpeca: I agree, will do

spyderlib/widgets/tests/test_array_builder.py
goanpeca: Yes I think your are right, will move to a fixture :-)

spyderlib/widgets/tests/test_array_builder.py
goanpeca: Yes, otherwise the widget ignores all the events

spyderlib/widgets/arraybuilder.py
goanpeca: Maybe... still it feels kinda weird :-|

I have mixed feeling about this

spyderlib/widgets/arraybuilder.py
goanpeca: Sure

spyderlib/widgets/arraybuilder.py
Nodd: To me the strangest thing is that the widget is not private while other components are. It looks like it's used somewhere else.
But I don't want to hurt your mixed feelings! :wink: You wrote this widget so you can decide.

spyderlib/widgets/tests/test_array_builder.py
Nodd: I see. The comment should be more precise then, like "Hack: in the tests the selected cell is wrong". It's fine if it's on its own line too, what's important is to tell why the line is needed.

spyderlib/widgets/arraybuilder.py
goanpeca: In general I think of private stuff as things only the same object can access, so if another object (even in the same module) can access something, then it is really not private.

So using a thing from within the tests (another place / object) feels like it should be public?

spyderlib/widgets/tests/test_array_builder.py
```diff
@@ -0,0 +1,204 @@
...
+# Copyright © 2009- The Spyder Development Team
+# Licensed under the terms of the MIT License
+#
+
+"""
+"""
+
+# Third party imports
+from qtpy.QtCore import Qt
+from pytestqt import qtbot  # analysis:ignore
```
Nodd: Why all these `# analysis:ignore`? I don't see what doesn't follow the pep8 or other guidelines.

spyderlib/widgets/arraybuilder.py
Nodd: I agree with you for runtime coed, but I tend to see tests a bit differently. Sometimes you have to do weird things in tests (monkeypatching, mocks, and so on) so accessing a private variable isn't a big deal.

goanpeca: ok you convenced me, will make it private again ;-)

spyderlib/widgets/tests/test_array_builder.py
goanpeca: Will do

spyderlib/widgets/arraybuilder.py
goanpeca: Not sure how to check this part!

goanpeca:
@Nodd thanks for the comments, I addressed adding some fixtures and cleaning the code a bit.

goanpeca:
@PierreRaybaut I am getting this error on travis-ci from your latest commit

```bash
$ cd doc; make html linkcheck

make: *** No rule to make target `html'.  Stop.

The command "cd doc; make html linkcheck" exited with 2.
```

runtests.py
```diff
@@ -0,0 +1,21 @@
...
+"""
+
+# Third party imports
+import qtpy
+import pytest
+
+
+def main():
+    pytest.main(['-x', 'spyderlib',  '-v', '-rw', '--durations=10',
+                 '--cov=spyderlib', '--cov-report=term-missing'])
```
Nodd: These options could be set in setup.cfg so that manually running pytest would use them. Honestly I'm not sure what's best, I'm just thinking aloud here.

spyderlib/widgets/arraybuilder.py
Nodd: Maybe the easiest way is to call directly `event()` with a custom event.
Anyway you shouldn't hide the fact that it's not tested, even if it lowers the coverage stat. Actual measurement is better than a better score.

spyderlib/widgets/tests/test_array_builder.py
```diff
@@ -0,0 +1,177 @@
...
+from spyderlib.widgets.arraybuilder import NumpyArrayDialog
+
+
+# --- Fixtures
+# -----------------------------------------------------------------------------
+@pytest.fixture
+def botinline(qtbot):
+    dialog = NumpyArrayDialog(inline=True)
+    qtbot.addWidget(dialog)
+    qtbot.dialog = dialog
```
Nodd: You don't need this if you return `dialog` directly (same for the other fixtures)

Nodd:
Just a few minor comments and it's all good for me.

runtests.py
goanpeca: I would like to add more stuff here, like flake8 checks, yapf and stuff, so having a separate script is kinda nicer.

That is why I chose to use it like this.

spyderlib/widgets/tests/test_array_builder.py
goanpeca: :+1:

spyderlib/widgets/arraybuilder.py
goanpeca: :+1:

ccordoba12:
Is this coverails GH message going to appear on every commit? Sorry, but I don't like it too much ;-)

goanpeca:
>Is this coverails GH message going to appear on every commit? Sorry, but I don't like it too much ;-)

I dont know and I dont really care, it does not affect anything.

goanpeca:
@Nodd , I think I addressed all the comments.

Anything extra you want to add before merging?

PR #3133, opened 2016-04-22
Add an option to set window title to the command line

### LINKED ISSUE #3129: Is there a way to modify the main window title?
We are working in a multi-user environment with multiple instances of spyder opened. It would be convenient to show the session name (or any string) in the window title. Is there already a method to to this? If not, I can add a command line option to enable this.

eendebakpt:
Fixes #3129

### DIFF
--- a/spyderlib/app/cli_options.py
+++ b/spyderlib/app/cli_options.py
@@ -37,5 +37,7 @@ def get_options():
     parser.add_option('--profile', action='store_true', default=False,
                       help="Profile mode (internal test, "
                            "not related with Python profiling)")
+    parser.add_option('--windowtitle', type=str, default=None,
+                      help="String to show in main window")
     options, args = parser.parse_args()
     return options, args
--- a/spyderlib/app/cli_options.py
+++ b/spyderlib/app/cli_options.py
@@ -37,7 +37,7 @@ def get_options():
     parser.add_option('--profile', action='store_true', default=False,
                       help="Profile mode (internal test, "
                            "not related with Python profiling)")
-    parser.add_option('--windowtitle', type=str, default=None,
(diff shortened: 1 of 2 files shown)


spyderlib/app/cli_options.py
```diff
@@ -37,5 +37,7 @@ def get_options():
     parser.add_option('--profile', action='store_true', default=False,
                       help="Profile mode (internal test, "
                            "not related with Python profiling)")
+    parser.add_option('--windowtitle', type=str, default=None,
```
ccordoba12: Could add an hyphen between `window` and `title`? So this option should be

    --window-title

spyderlib/app/cli_options.py
```diff
@@ -37,5 +37,7 @@ def get_options():
...
+                      help="String to show in main window")
```
ccordoba12: This help text should read

    String to show in the main window title

ccordoba12:
Other than my minor comments, I think this is ready :-)

eendebakpt:
Ok, code is updated with the suggestions.

spyderlib/app/spyder.py
```diff
@@ -439,6 +439,9 @@ def signal_handler(signum, frame=None):
                                                sys.version_info[1])
         if DEBUG:
             title += " [DEBUG MODE %d]" % DEBUG
+        if not options.window_title is None:
```
Nodd: `if options.window_title is not None:` is more commonly used in python, because it looks like an english sentence.

eendebakpt: Ok, changed to code for this

ccordoba12:
@eendebakpt, thanks a lot for your contribution! :-)

PR #3146, opened 2016-05-02
Editor: Move to next line in run_selection() if nothing selected

### LINKED ISSUE #3115: Automatically advance to the next line after pressing F9
There's already a [question on stackoverflow](http://stackoverflow.com/questions/35611033/how-to-automatically-advance-to-the-next-line-after-f9-in-spyder-ide) that perfectly describes my question:

> Coming from RStudio, I'm familiar with the behavior of the cursor automatically moving to the next line after running the current line. I think this makes sense for data analysis. Does Spyder have something similar?

> I'm familiar with the concept of cell and auto-advancing cell in Spyder. However, I want to maintain a pure Python script (that can be run with python file.py) for portability, and don't want to use something that only makes sense on Spyder.

I know some people in my working group and other institutions that would already like to have such a shortcut for procedural data analysis scripts, plotting, ...

Is there a way to realize that behaviour (without using code cells)?

jitseniesen:
Fixes #3115 

-----
@jitseniesen I rewrote the header to the standard we try to use (`Fixes #<issue number>`)

### DIFF
--- a/spyderlib/widgets/editor.py
+++ b/spyderlib/widgets/editor.py
@@ -1773,8 +1773,10 @@ class EditorStack(QWidget):
 
     #------ Run
     def run_selection(self):
-        """Run selected text or current line in console
-        If no text is selected, then advance cursor to next line."""
+        """
+        Run selected text or current line in console. If no text is selected, 
+        then advance cursor to next line.
+        """
         text = self.get_current_editor().get_selection_as_executable_code()
         if text:
             move_next_line = False
--- a/spyderlib/widgets/editor.py
+++ b/spyderlib/widgets/editor.py
@@ -1773,12 +1773,19 @@ class EditorStack(QWidget):
 
     #------ Run
     def run_selection(self):
-        """Run selected text or current line in console"""
+        """Run selected text or current line in console
+        If no text is selected, then advance cursor to next line."""
         text = self.get_current_editor().get_selection_as_executable_code()
-        if not text:
-            line = self.get_current_editor().get_current_line()
+        if text:
+            move_next_line = False
+        else:
+            editor = self.get_current_editor()
+            line = editor.get_current_line()
             text = line.lstrip()
+            move_next_line = True
         self.exec_in_extconsole.emit(text, self.focus_to_editor)
+        if move_next_line:
+            editor.move_cursor_to_next('line', 'down')
 
     def run_cell(self):
         """Run current cell"""


jitseniesen:
It seems to me from the log that the AppVeyor failure occured while the test environment is set up so it is not caused by anything in the pull request.

goanpeca:
@jitseniesen Awesome :-), thanks for working on this.

Indeed, the appveyor tests are broken for some weeks now, @ccordoba12 will fix them soon :-)

spyderlib/widgets/editor.py
```diff
@@ -1773,12 +1773,19 @@ def fix_indentation(self, index=None):
 
     #------ Run
     def run_selection(self):
-        """Run selected text or current line in console"""
+        """Run selected text or current line in console
+        If no text is selected, then advance cursor to next line."""
```
goanpeca: Minor style comment

Can we turn this into

```python
"""
Run selected text or current line in console. If no text is selected, then
advance cursor to next line.
"""
```

goanpeca:
LGTM :+1:

Any comments @spyder-ide/core-developers

blink1073:
Yep, looks sensible to me.

goanpeca:
**looks sensible to me** WTF?

blink1073:
As in I agree with the approach, but did not actually test it ;).

jitseniesen:
Of course I'm happy to follow the house style (regarding formatting of docstrings), so I changed it. I tried to infer how you wanted it from other docstrings, but obviously failed.

goanpeca:
@jitseniesen thanks for the changes... we do not have an specific style written somewhere, but what I am trying to promote is something in the lines of.

- Docstrings are sentences, they need to end with a `.`.
- Docstrings should have a quick one liner summary and a more lengthy description afterwards.
- Doctsrings respect 80 characters length (PEP enforces a smaller value but mehhh...)
- Doctsrings should ALWAYS use double tripple quotes and ALWAYS be written using a separate line as in:

```python
# As in
def test()
    """
    Hello world.
    """

# Instead of 
def test()
    """Hello world."""
```

This makes the rule easy.... ALWAYS use that instead of one for one liners or another for whatever reason.

What do you think?

jitseniesen:
I think it makes sense for a more-than-one-(wo)man project to have style rules and it does not make much sense to argue long about them, so I'm happy to follow those rules.

ccordoba12:
Thanks for working on this @jitseniesen! Great addition :-)

Nodd:
What is preventing merging this PR ? Everything looks fine ?

ccordoba12:
If you feel it's ok, please merge it :-)

PR #3151, opened 2016-05-03
More robust plugin initialization in layout

Nodd:
If a plugin has a problem to initialize, it will crash spyder.

This PR avoids to crash while still providing the full traceback.


goanpeca:
LGTM :+1:

ccordoba12:
Thanks @Nodd :-)

PR #3177, opened 2016-05-23
PR: Fix automatic insertion of colon

### LINKED ISSUE #872: Quotes and colons autocompletion
_From [nunoport...@gmail.com](https://code.google.com/u/108617919618299841919/) on 2011-12-07T06:58:36Z_

Hi,

How can i edit Spyder autocomplete feauture? I can't seem to find in preferences a way to do this...

thanks

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=872_

jitseniesen:
This PR prevents the automatic insertion of a colon at the end of one-line if statements like:
````python
if True: print "Yes!"
````
It also prevents the automatic insertion of a colon at the second line of two-line list comprehensions like:
```python
yes_list = [x for x in [None, None, None, "Yes!"]
            if x is not None]
```
Fixes #872

### DIFF
--- a/spyderlib/widgets/sourcecode/codeeditor.py
+++ b/spyderlib/widgets/sourcecode/codeeditor.py
@@ -2306,6 +2306,18 @@ class CodeEditor(TextEditBaseWidget):
                     return True
         return False
 
+    def __has_colon_not_in_brackets(self, text):
+        """
+        Return whether a string has a colon which is not between brackets.
+        This function returns True if the given string has a colon which is
+        not between a pair of (round, square or curly) brackets. It assumes
+        that the brackets in the string are balanced.
+        """
+        for pos, char in enumerate(text):
+            if char == ':' and not self.__unmatched_braces_in_line(text[:pos]):
+                return True
+        return False
+
     def autoinsert_colons(self):
         """Decide if we want to autoinsert colons"""
         line_text = self.get_text('sol', 'cursor')
@@ -2318,6 +2330,8 @@ class CodeEditor(TextEditBaseWidget):
         elif self.__forbidden_colon_end_char(line_text):
             return False
         elif self.__unmatched_braces_in_line(line_text):
+            return False
+        elif self.__has_colon_not_in_brackets(line_text):
             return False
         else:
             return True
--- a/spyderlib/widgets/sourcecode/codeeditor.py
+++ b/spyderlib/widgets/sourcecode/codeeditor.py
@@ -2300,6 +2300,10 @@ class CodeEditor(TextEditBaseWidget):
                 match = self.find_brace_match(line_pos+pos, char, forward=True)
                 if (match is None) or (match > line_pos+len(text)):
                     return True
+            if char in [')', ']', '}']:
+                match = self.find_brace_match(line_pos+pos, char, forward=False)
+                if (match is None) or (match < line_pos):
+                    return True
         return False
 
     def autoinsert_colons(self):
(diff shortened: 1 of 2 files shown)


jitseniesen:
This does not fix all issues. For instance, even after this is merged, a colon is still erroneously added after the second line of:
```python
yes_list = [x for x in [None, None, None, "Yes!"]
            if x is not None
            or x.startswith("N")]
```
And I'm pretty sure there are other cases in which a colon is either added when it shouldn't be or it is not added where it should. I guess at some point we need to move from ad-hoc implementation to a more complete parsing. Maybe it's better to combine with other parts of the Spyder code base; e.g. automatic indentation?

If #3095 is merged, then we can similarly write tests for automatic insertion of colons.

ccordoba12:
@jitseniesen, thanks so much for looking into this one!! I've been unable to find time to fix it myself :-)

To fix tests in Travis, please merge with master :-)

----

@goanpeca, how coverails works? We can have PRs failing because of it :-)

goanpeca:
@ccordoba12 coverage decreased so we get an "error", but it was marginal, so we can round the value to it does not complain.

ccordoba12:
Could we deactivate that feature of coverails? Or change it to big decrease (say 10% or 20%)?

goanpeca:
We are on 4%, I think a 1% decrease is already very alarming.

goanpeca:
I changed it, so if it decreases 1% it will fail ;-)

ccordoba12:
> We are on 4%, I think a 1% decrease is already very alarming.

Ok, fair enough :-)

jitseniesen:
> To fix tests in Travis, please merge with master :-)

Done. Actually, I rebased instead of merged; I think you guys prefer that.

goanpeca:
@jitseniesen whatever the dev feels better with. Megre is safer, rebase can lead to trouble if done incorrectly. I prefer rebase in general, but sometimes conflicts are very tedious to fix so I go the merge way :-p

Nodd:
Recently a PR was merged to enable testing with pytest. It would be great if you had the motivation to add tests for this, including failing tests (marking them as known fail).

jitseniesen:
@Nodd, I was planning to wait until you finished PR #3095 for testing automatic indentation and use that as a template. :)

I don't know the best way to test the automatic insertion of colons because the code is very tightly coupled to the UI - it's part of the class `CodeEditor` which inherits from a Qt class, and it uses Qt functionality for stuff like finding the start of line. You probably struggled with this yourself and so know all about it.

One possibility would be for me to try and (partially) decouple the automatic insertion of colons from the `CodeEditor` class. This would allow for cleaner testing. Another possibility is to take your work, convert it to pytest, and add testing for the colon stuff. And then of course there is my original plan, to wait for you.

What are your plans and what do you think about this?

goanpeca:
@jitseniesen the idea is to use pytest-qt to test UI widgets, that was the point @Nodd was probably trying to make.

It is very difficult to test this type of code in a decoupled way, cause it is code that works in the actual UI/codeeditor.

You can take a look at https://github.com/spyder-ide/spyder/blob/master/spyderlib/widgets/tests/test_array_builder.py to get the general idea. Basically you can create a fixture that returns a Codeeditor widget and then you test the actual interaction via qtbot (sendclicks and sendkey) and then check that the content you get includes the colon.

If you need some guidance, let me know.

Nodd:
I don't know how exactly is the code layout for colon insertion, but if it's like autoindent you don't need pytest-qt as you can change the text and call the functions directly.
@jitseniesen Well played, you forced me to do it ! :wink: Have a look at #3095

jitseniesen:
Thank you both for your explanations. As you can see, I added some tests for the colon insertion.

ccordoba12:
@jitseniesen, this is really great, not only the fix itself but also the tests!

Merging because everything is green :-)

PR #3210, opened 2016-06-07
PR: Follow the Flask plugin model instead of namespace packages

### LINKED ISSUE #3150: External plugin import
## Description
I had to change the package name for autopep8, because there was a name clash with the autopep8 *package* that's used in autopep8 *plugin*.
In theory spyder plugins should be loaded as `spyplugins.ui.pluginame`, but it seems that somehow `pluginame` is added to `sys.modules` too.

* Spyder Version: 3.0.0dev 
* Python Version: 3.5.1
* Qt Versions: 5.5.1, PyQt5 5.5.1 on Linux

## Dependencies
```
pyflakes >=0.6.0:  1.1.0 (OK)
pep8 >=0.6      :  1.7.0 (OK)
pygments >=1.6  :  2.1.3 (OK)
qtconsole >=4.0 :  4.2.1 (OK)
nbconvert >=4.0 :  4.2.0 (OK)
pandas >=0.13.1 :  0.18.0 (OK)
numpy >=1.7     :  1.11.0 (OK)
sphinx >=0.6.6  :  1.4.1 (OK)
rope >=0.9.4    :  0.9.4-1 (OK)
jedi >=0.8.1    :  0.9.0 (OK)
psutil >=0.3    :  4.1.0 (OK)
matplotlib >=1.0:  1.5.1 (OK)
sympy >=0.7.3   :  None (NOK)
pylint >=0.25   :  1.5.5 (OK)

```

Nodd:
Here is a quick prototype for loading plugins using the flask model.

Fixes #3150.
I went with the two different prefixes because it was simpler, it can be changed in the future.

I still have to look precisely at how to load a module, because from what I saw more functions were deprecated in python 3.5.

TODO:
- [x] Test user plugins (in ~/.spyder)
- [x] Test pip installed plugins
- [x] Remove the removal of whitespace


ccordoba12:
@Nodd, thanks a lot for working on this one!

Quick comments:

1. I'd prefer that we name our plugins like `spyder_ui_breakpoints` instead of `spyderui_breakpoints`. I think it's clearer.
2. Please revert blank removals, unless they are in the area of influence of your change.
3. Why loading code needs to be revisited again?

ccordoba12:
Or maybe `spy_ui_breakpoints` to make things shorter ;-)

goanpeca:
@nodd and @ccordoba12 I dont think adding the ui or io is worth it. Io things are just plugins... so making the difference by name is really just making the names ugly and long.

We should just add some variable in the init of the plugin like

```
from spyderlib.plugin import IO_PLUGIN, UI_PLUGIN

PLUGIN_TYPES = [IO_PLUGIN, UI_PLUGIN]
...
```

Nodd:
@ccordoba12 
1. I hesitated between the two, the additional underscore make it look very long, but I agree it's clearer.
2. Raaah, these freaking whitespaces ! We should burn them all ! OK ok, I'll do it.
3. It's nothing big, right now I commented everything to test things on my machine easily. The code is the same but I'll double check if it's applied to the correct python versions.

@goanpeca From a certain point of view IO plugins are just plugins because they are additions to spyder, but they are actually very different from UI plugins:
* they are used exclusively to load new file format, and have nothing to do with anything else
* they are loaded during the imports (before spyder's initialization) and are considered as constants in multiple parts of the codebase

Also the big problem with differentiating by module variables (apart from being a little magical) is that you have to load the plugin to check which kind it is. Considering how they are loaded and used very differently internally, I sill think it makes sense to differentiate them explicitly in their names. To be honest if we decide to refactor the way IO plugins are loaded, someone else will have to do it because I won't be able to spend time on this.

To keep names shorter, I'd be in favor of having IO plugins starting with `spyder_io_` and generic plugins starting with `spyder_`. I expect that there will be more generic plugins than io plugins in the future. To be exhaustive I'll add that it prevents generic plugin names from starting with `io`, but I don't think it'll be a problem. What do you prefer ?
In any case `io`/`ui` needs to be in the module name only, not necessarily the plugin name itself. `spyder-myplugin` can install the module `spyder_ui_myplugin`.

goanpeca:
@nodd, I understand how it works and how they load differently and how they might load differently. But IO plugins was just a special case of a general plugin that extends the variable explorer... so in a sense is a normal plugin that extends that functionality. It was not coded that way for whatever reason and it starts in a different part of the application for whatever reason, and trying to compensate for that seems like a bad idea.

In general every normal plugin could/should be extendable.

>In any case io/ui needs to be in the module name only, not necessarily the plugin name itself. spyder-myplugin can install the module spyder_ui_myplugin.

I just dislike this very much and find very useful that modules and packages names relate to each other.

Nodd:
It's not "whatever reason". On one side a plugin is a function to load a file with a specific format and extension. On the other side a plugin has a complex API and explicitly needs a dockwidget (a menu-entry plugin needs a hack currently). Even if we name them both "plugins" they have a very different goal (I'm not even speaking about implementation).

It's not just a naming problem, having 2 different interfaces for plugins is weird. I think that the most correct way would be to integrate IO plugins into UI plugins (maybe with a helper class to inherit from ?), feel free to break everything ! :wink: 

> To keep names shorter, I'd be in favor of having IO plugins starting with spyder_io_ and generic plugins starting with spyder_.

What do you think about this ? This way in the future we could change IO plugins only (merging them into ui plugins) while being backward compatible with ui plugins, and keeping the names clean.

goanpeca:
>It's not just a naming problem, having 2 different interfaces for plugins is weird. I think that the most correct way would be to integrate IO plugins into UI plugins (maybe with a helper class to inherit from ?), feel free to break everything ! :wink:

Yes, I would like to have a single plugin interface, that is the whole point :-)

>What do you think about this ? This way in the future we could change IO plugins only (merging them into ui plugins) while being backward compatible with ui plugins, and keeping the names clean.

Yes perfect. I would say the approach to take after this PR is replicate the IO functionality in normal plugins, allow for easy (io) plugin creation without needing a dockwidget (maybe?) and allow for a standard way of adding entries to menus (this last one requires adding some sort of groups to menu entries)

Nodd:
WTF git ? I merged nothing !

goanpeca:
??

Nodd:
All my commits were removed from the PR...

ccordoba12:
@Nodd, it seems you changed all line endings of several files :-)

Nodd:
:sob: I edited some files on windows... I tought it was correctly configured, but it seems I was wrong.

Nodd:
I put the EOL back to the mess they where. I still have to re-mess the whitespaces !

Nodd:
OK, everything should work now ! I also updated spyder.line_profiler and spyder.autopep8 (in my repo, master branch) if you want to test external plugins.

ccordoba12:
Thanks a lot Joseph! We definitely need to move to normalize line endings!!

Could prepare a new PR about that? I promise to merge it and adapt all 
our team PRs accordingly. Thanks :-)

Nodd:
line_profiler (the plugin) is still buggy, I ported mamory_profiler but it's buggy too. You can test the vim plugin too, I just did some work on it.

Nodd:
I did some cleanups in the plugins, they should all work now. You have to use the versions from https://github.com/Nodd?tab=repositories, I'll push to spyder-ide when this PR is merged.

Nodd:
... and I'm slowly starting to hate QuantifiedCode. I like nice-looking code, but this thing is so mechanical that the few useful warnings are lost in the noise.
On this specific PR it should detect when the files are moved around and not modified.

ccordoba12:
@Nodd, is this one ready?

I think the only thing left is to fix `setup.py` to package the new `spyder_*` directories instead of `spyplugins`, but I can do that after merging this one.

Nodd:
I updated the gettext scripts and setup.py (without testing the changes...). Otherwise it should be ready, I don't know if someone tested on another configuration than my machine.

spyderlib/otherplugins.py
```diff
@@ -8,80 +8,69 @@
...
+    import importlib
 
 
-def _get_spyderplugins(plugin_path, base_namespace, plugins_namespace,
-                       modnames, modlist):
-    """Scan the directory `plugin_path` for plugins_namespace package and
-    loads its submodules."""
-    namespace_path = osp.join(plugin_path, base_namespace, plugins_namespace)
+PLUGIN_PREFIX = "spyder_"
+IO_PREFIX = PLUGIN_PREFIX + "io_"
```
ccordoba12: Question here: What's the name for `io` plugins according to this? Is it `spyder_io_foo` or `spyderio_foo`?

I say that because you named our io plugins like `spyderio_foo`, so those names or this line should be changed :-)

Nodd: `spyder_io_foo`. I forgot to rename the files.

setup.py
```diff
@@ -236,7 +236,11 @@ def run(self):
       platforms=['any'],
       packages=get_packages(),
       package_data={LIBNAME: get_package_data(LIBNAME, EXTLIST),
-                    'spyplugins': get_package_data('spyplugins', EXTLIST),
+                    'spyder_breakpoints': get_package_data('spyder_breakpoints', EXTLIST),
+                    'spyder_profiler': get_package_data('spyder_profiler', EXTLIST),
+                    'spyder_pylint': get_package_data('spyder_pylint', EXTLIST),
+                    'spyderio_dcm': get_package_data('spyderio_dcm', EXTLIST),
+                    'spyderio_hdf5': get_package_data('spyderio_hdf5', EXTLIST),
```
ccordoba12: @Nodd, these lines also need to be changed to `spyder_io_foo` :-)

ccordoba12:
Except for a minor detail in `setup.py` this one is ready to go.

Nodd:
Stupid me, I modified `setup.py` before renaming the files.
The travis failure seems like a bug in qtpy with Qt5.

ccordoba12:
The problem in Travis had to do with new conda packages for PyQt5. They had the wrong dependencies, so I removed them for now :-)

I have to recompile Qt5 to fix that problem.

Nodd:
I updated the master branch of all "my" plugins in spyder-ide so that they are compatible. Note that it's not the case for spyder.example ! (and spyder.terminado but this one is empty)

TWAC:
After this merge I am not getting plugins from bdist_wheel.
It looks like get_packages() in setup.py still looks for the spyplugins directory.

Nodd:
@TWAC You're right, thank you for noticing. I sent a PR to correct this.

PR #3218, opened 2016-06-09
PR: Allow empty wdir option to run_python_script_in_terminal

### LINKED ISSUE #3155: Exception raised when running script in external terminal
## Description

If I try to run a script in an external terminal, an exception is raised.

1. Write a Python script (contents do not matter)
2. Click "Configure" in the "Run" menu. 
3. In the window that appears, pick "Execute in external system terminal" and uncheck the other options.
4. Press the "Run" button.

The script does not run. Instead, the traceback below is printed in the external console (I abbreviated the paths). The issue occurs in both 3.0.0b2 and the current git revision.

```
  File "(...)/site-packages/spyderlib/plugins/editor.py", line 2082, in edit_run_configurations
    self.run_file()
  File "(...)/site-packages/spyderlib/plugins/editor.py", line 2129, in run_file
    self.re_run_file()
  File "(...)/site-packages/spyderlib/plugins/editor.py", line 2172, in re_run_file
    systerm, post_mortem)
  File "(...)/site-packages/spyderlib/spyder.py", line 2483, in open_external_console
    interact, debug, python_args)
  File "(...)/site-packages/spyderlib/utils/programs.py", line 183, in run_python_script_in_terminal
    cwd=wdir)
  File "(...)/site-packages/spyderlib/utils/programs.py", line 64, in run_program
    subprocess.Popen([path]+args, cwd=cwd)
  File "(...)/subprocess.py", line 950, in __init__
    restore_signals, start_new_session)
  File "(...)/subprocess.py", line 1544, in _execute_child
    raise child_exception_type(errno_num, err_msg)
FileNotFoundError: [Errno 2] No such file or directory: ''
```

* Spyder Version: 3.0.0b2 
* Python Version: 3.5.1

## Dependencies
```
pyflakes >=0.6.0   :  1.1.0 (OK)
pygments >=1.6     :  2.1.1 (OK)
IPython >=3.0      :  4.1.2 (OK)
zmq >=13.0.0       :  15.2.0 (OK)
qtconsole >=4.0    :  4.1.1 (OK)
pandas >=0.13.1    :  0.17.1 (OK)
numpy >=1.7        :  1.10.4 (OK)
sphinx >=0.6.6     :  1.3.5 (OK)
psutil >=0.3       :  4.1.0 (OK)
rope >=0.9.4       :  0.9.4-1 (OK)
sympy >=0.7.3      :  0.7.6.1 (OK)
pylint >=0.25      :  1.5.4 (OK)
```

jitseniesen:
Fixes #3155: Executing a python script in external terminal without
specifying a working directory leads to wdir = '', which is not handled
correctly.

In the previous version, the working directory was passed as an option
to the script if xterm is used. This did not make sense to me, so I
removed that.

### DIFF
--- a/spyderlib/utils/programs.py
+++ b/spyderlib/utils/programs.py
@@ -252,7 +252,11 @@ def get_python_args(fname, python_args, interact, debug, end_args):
 
 def run_python_script_in_terminal(fname, wdir, args, interact,
                                   debug, python_args):
-    """Run Python script in an external system terminal"""
+    """
+    Run Python script in an external system terminal.
+
+    :str wdir: working directory, may be empty.
+    """
     
     # If fname has spaces on it it can't be ran on Windows, so we have to
     # enclose it in quotes. Also wdir can come with / as os.sep, so we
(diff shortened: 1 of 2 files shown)


goanpeca:
@jitseniesen thanks for working on this. Could we add some tests :-) for this function.

goanpeca:
Also since we are now dealing with wdir possibly being empty, should not we make it a kwargs?, Would this change be too disruptive (many places where this function is called?)

@nodd, @ccordoba12 ?

jitseniesen:
@goanpeca I thought about adding tests, but I found it too hard :-) What this function does, depends on the OS and on whether certain programs are installed. Furthermore, the function actually runs a script. Should we depend on properties of the testing environment? Should we hide all external dependencies behind mocks?

Regarding making wdir a kwarg, I had a quick look and I think that's a good idea in principle. This function seems to be only called once, in MainWindow.open_external_console() in spyder.py. That function just forwards the wdir it is passed, so probably wdir should also be made a kwarg there. The function open_external_console() is used in two places, in Explorer and Editor, so just changing wdir to be a kwarg should be not too hard.

But if you start rewriting, where do you stop, and do we want to do this? I'm not familiar with the code, I don't find it very easy to read, it needs to handle Python2 and Python3 and different OSes so it needs to be careful with string encodings which I don't really know about, it's hard to test, and we are (as I understand it) close to releasing Spyder 3.0. All arguments (for me) to change the code as little as possible.

On having a further look, I wonder whether the windows branch has the same bug. And if we do decide to rewrite, it seems to me that as you suggested, other arguments should be given default values (all except fname, in fact). I also find it odd that WindowsError is caught in this function and displays a dialog box, but NotImplementedError (unsupported OS) is to be caught in the calling function, where it displays a dialog box; to me, this should be done in the same place.

goanpeca:
Hi @jitseniesen yes it might be tricky to make changes right now, so caution at this point would be a good idea :-). Now for testing, in principle the tests folder could include some test script, either as an actual `_script_func_test.py` file, or as a text string inside the `test_***.py`

Basically it should be a simple script that has a verifiable output, so I think such script should just write something to a given location that the user provides via arguments and we should assert that that was the case. Since we run tests on appveyor and travis, we are testing on 2 OSs for now, eventually we will add OSX via travis (or split travis for OSX and circle CI for linux like conda-forge does).


The test script could receive a tmpfile and the content to check it worked, something like

`script_test_external_console.py <tmpfile> <content>`

where tmpfile is a pytest fixture and content is just whatever you want to write to check it worked :-)

What do you think?

jitseniesen:
Thanks. I'll try to write a test.

ccordoba12:
I think this looks good, thanks!

PR #3228, opened 2016-06-15
Fix a bug preventing Spyder to open external file In Mac application 

impact27:
sig_open_external_file is present in spyderlib/app/spyder.py and doesn't match open_external_file (had to add sig_)


ccordoba12:
Thanks for the fix. The test failure is unrelated to your change, so you don't need to worry about it :-)

PR #3234, opened 2016-06-19
Create only one instance of IntrospectionManager for the application

ccordoba12:
Before we were creating one instance of IntrospectionManager per EditorStack, which was causing freezes when adding more EditorStacks (e.g. when a new Editor window was created or when the Editor was splitted).

@blink1073, could you take a look at this one?


blink1073:
Code and behavior LGTM.

ccordoba12:
Thanks @blink1073!

Issue #2648, opened 2015-08-27
Create self-contained Windows Installers


techtonik:
After installing Spyder from `spyder-2.3.6.win32-py2.7.exe` it still fails to run without PySide or PyQt4 installed.

![image](https://cloud.githubusercontent.com/assets/515889/9519598/a93f08ea-4cc8-11e5-889d-e752ffb81e24.png)

It would be nice if instead of this window, Spyder could fetch and install PySide wheel locally.

ccordoba12:
I think this doesn't belong on the Spyder side of things, it's a packaging issue.

techtonik:
@ccordoba12 if Spyder aim is to provide a better UX that IDLE, then bootstrap process is also part of that experience. Take a look at how NLTK resolved that UX issue with `nltk.download()`.

Nodd:
The minimum would be to add a link to PyQt or PySide next to the exe download. It's true that on windows it's very uncommon to install dependencies separately. (It only happened to me once with gtk.)
It could also be an option in the installer.

techtonik:
I downloaded the standard installer for Python packages and I am not sure it can be patched to add such option.

Nodd:
I meant spyder's installer.

ccordoba12:
PySide is just the minimal thing to have Spyder working. But that doesn't give code completion, the IPython/Jupyter console, PyFlakes warnings, etc, etc. So it's not a real solution.

What we need is to create fully self-contained Windows installer. For that we could use this project:

https://github.com/takluyver/pynsist

@techtonik, if you want to contribute a `cfg` file for Spyder, that'd be great :-)

techtonik:
@ccordoba12 I am interested to do this, but I am afraid that I will need some financial help to to get this done.

techtonik:
https://github.com/pyinstaller/pyinstaller looks like a good option.

techtonik:
One of the scenarios that needs to be supported is when the installer includes Python. Because Spyder is needed to teach people. And while installing Python 3.5, there is a problem with Windows updates that needs to be considered.

https://github.com/winpython/winpython/issues/341

takluyver:
[...]

PR #3238, opened 2016-06-20
Update README.md

techtonik:
Fix PythonXY link. Also making Spyder's own installers
the preferable way to install it on Windows

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,10 +68,6 @@ The easiest way to install Spyder is:
 
 ### On Windows:
 
-- Use `pip` to test beta versions of Spyder 3.x:
-
-      pip install --pre -U spyder
-
 - For Spyder 2.x, the easiest way is to use one of scientific
   Python distributions:
   1. [Anaconda](http://continuum.io/downloads)
@@ -80,9 +76,15 @@ The easiest way to install Spyder is:
 
 - There are also [executable installers](https://github.com/spyder-ide/spyder/releases)
   for Spyder 2.x, but they require to install required
-  dependencies, such as PyQt manually, so with Spyder 3.x
+  dependencies (such as PyQt) manually, so with Spyder 3.x
   they are dropped until somebody comes with a solution for more
   friendly installers (see #2648).
+
+- To test beta versions of Spyder 3.x use standard way with
+  `pip`. This doesn't create any start meny or desktop shortcuts,
+  so it is an option for power users.
+
+      pip install --pre -U spyder
 
 
 ### On Mac OSX:
--- a/README.md
+++ b/README.md
@@ -68,19 +68,21 @@ The easiest way to install Spyder is:
 
 ### On Windows:
 
-- You can use `pip`:
+- Use `pip` to test beta versions of Spyder 3.x:
 
-    pip install spyder
+      pip install --pre -U spyder
 
-- Or install one scientific Python distributions that come with Spyder:
+- For Spyder 2.x, the easiest way is to use one of scientific
+  Python distributions:
   1. [Anaconda](http://continuum.io/downloads)
   2. [WinPython](https://winpython.github.io/)
   3. [Python(x,y)](http://python-xy.github.io/)
 
-We used to produce [executable installers](https://github.com/spyder-ide/spyder/releases),
-but they were missing required dependencies and with Spyder 3.x
-we've decided to drop them until somebody comes with a solution
-for more friendly installers (see #2648).
+- There are also [executable installers](https://github.com/spyder-ide/spyder/releases)
+  for Spyder 2.x, but they require to install required
+  dependencies, such as PyQt manually, so with Spyder 3.x
+  they are dropped until somebody comes with a solution for more
+  friendly installers (see #2648).
 
 
 ### On Mac OSX:
--- a/README.md
+++ b/README.md
@@ -73,16 +73,10 @@ The easiest way to install Spyder is:
   1. [Anaconda](http://continuum.io/downloads)
   2. [WinPython](https://winpython.github.io/)
   3. [Python(x,y)](http://python-xy.github.io/)
+- Or use [executable installers](https://github.com/spyder-ide/spyder/releases)
 
-- There are also [executable installers](https://github.com/spyder-ide/spyder/releases)
-  for Spyder 2.x, but they require to install required
-  dependencies (such as PyQt) manually, so with Spyder 3.x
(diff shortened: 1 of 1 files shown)


Nodd:
Thank you for the link update, but why did you change the preferred order on windows ? Python distributions provide a more complete environment with more packages.

techtonik:
@Nodd because primary audience who came here to read README.md need only Spyder. Those who need scientific packet with a lot of stuff are already using those "more complete environment"s or can choose for themselves if they need this stuff.

techtonik:
With this order It also looks like Spyder is not a standalone tool, but an addition to scientific packets, which is not true. I want people to use it instead of Sublime, because it is open source.

README.md
```diff
@@ -68,12 +68,11 @@ The easiest way to install Spyder is:
 
 ### On Windows:
 
-- Or using one of these scientific Python distributions:
+- Using one of our [executable installers](https://github.com/spyder-ide/spyder/releases)
```
ccordoba12: Please remove this line. I'm not going to produce Windows installers for Spyder 3.0.

The main reason is that we're now creating wheels for Spyder that pull all our dependencies correctly. The Windows installers can't install dependencies (much less PyQt) so they are kind of deceiving in that respect.

README.md
```diff
@@ -68,12 +68,11 @@ The easiest way to install Spyder is:
...
+- Or installing one of these scientific Python distributions:
```
ccordoba12: This line should read

    - Using one of these scientific Python distributions:

README.md
```diff
@@ -68,12 +68,11 @@ The easiest way to install Spyder is:
...
   1. [Anaconda](http://continuum.io/downloads)
   2. [WinPython](https://winpython.github.io/)
-  3. [Python(x,y)](http://pythonxy.googlecode.com)
-- Using one of our executable installers, which can be found
-  [here](https://github.com/spyder-ide/spyder/releases).
+  3. [Python(x,y)](http://python-xy.github.io/)
```
ccordoba12: Thanks for updating the PythonXY link!

README.md
ccordoba12: The only reasonable way to readd Windows installers is if we produce self-contained installers with py2exe, cx_freeze or a similar project.

But I don't have time for that, so someone else will have to help me with it ;-)

techtonik: Ok. I can help with #2648, after I bootstrap Spyder on my system, which needs merging #3239 first.

ccordoba12: Ok, thanks a lot for that! But until we have the installers ready, please do these modifications so we can merge this PR.

Then you can restore this text again.

README.md
techtonik: I see that those distributions are missing Spyder 3.x, so should we drop them until they add support for 3rd version?

techtonik:
I reworded the paragraph to attract more people to the problem.

README.md
ccordoba12: No, please leave them.

README.md
```diff
@@ -68,12 +68,20 @@ The easiest way to install Spyder is:
 
 ### On Windows:
 
-- Or using one of these scientific Python distributions:
+- You can use `pip`:
+
+    pip install spyder
```
ccordoba12: Please add this entry after the Scientific distributions one. It's far more easier to install those distros, so I want our users to see it as the first entry :-)

techtonik: > I want our users to see it as the first entry :-)

Is it because your work is sponsored by Continuum or there is another reason? Sorry if I am too blunt. I am ok with leaving it as-is, but I feel like it needs a better explanation, because explanations like "because I want it this way" sound very awkward from here. =) There is nothing wrong with sponsored development as long as it doesn't disrupt the development.

My vision to resist the change is that if people are introduced to Python through Spyder using official tools like `pip`, they would be motivated to help with improving those tools as well as core codebase and infrastructure, rather than helping with alternatives that already got their market share by fixing usability bugs independently.

I still bear the hope to make Spyder a replacement for IDLE.

BTW, you said

> The main reason is that we're now creating wheels for Spyder that pull all our dependencies correctly. 

Where are those wheels? I can't see them in https://github.com/spyder-ide/spyder/releases or in AppVeyor builds.

ccordoba12: > Is it because your work is sponsored by Continuum or there is another reason?

No, it is because it's far more easier to download and run a point-and-click installer, than:

1. Downloading Python
2. Opening a Terminal and running `pip`.
3. Not knowing how to run Spyder (because our wheels don't install menu entries).

So this method is really for power users. Jupyter guys also recommend the same:

http://jupyter.readthedocs.io/en/latest/install.html#id4

> Where are those wheels?

Where they should be, in PyPI:

https://pypi.python.org/pypi/spyder/3.0.0b3

----

After giving it more thought, I'm inclined to remove this block. We already mention how to use `pip` to install Spyder in a cross-platform way, so we don't need to repeat it here:

https://github.com/spyder-ide/spyder#cross-platform-way-from-source

So please remove it :-)

README.md
techtonik: I added clarification about current state of things.

Nodd:
@techtonik The first question after "how do I install spyder" is "how do I install this package". If the package is not pure python they are gonna have a hard time on windows. So it's really better for beginners to install a python distribution rather than standalone packages.

README.md
```diff
@@ -68,12 +68,20 @@ The easiest way to install Spyder is:
...
   2. [WinPython](https://winpython.github.io/)
-  3. [Python(x,y)](http://pythonxy.googlecode.com)
-- Using one of our executable installers, which can be found
-  [here](https://github.com/spyder-ide/spyder/releases).
+  3. [Python(x,y)](http://python-xy.github.io/)
+
+We used to produce [executable installers](https://github.com/spyder-ide/spyder/releases),
+but they were missing required dependencies and with Spyder 3.x
+we've decided to drop them until somebody comes with a solution
+for more friendly installers (see #2648).
```
Nodd: @techtonik I don't think this is needed. This is more informative about your frustration than about installing spyder ! :wink:

techtonik: I am still trying to make it run with bootstrap script, really, and it is failure after failure. =)

Nodd:
I'm fine with this version. We'll have to update it when 3.0 is out, of course.

ccordoba12:
I updated the PythonXY link, but I don't approve the other changes. so closing this one :-)

PR #3239, opened 2016-06-20
PR: bootstrap.py -- execute Spyder as subprocess

techtonik:
This removes Spyder dependency from Python-specific hacks
and bootstrap script by isolating communication interface
to system environment.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -38,8 +38,6 @@ parser.add_option('--hide-console', action='store_true',
                   default=False, help="Hide parent console window (Windows only)")
 parser.add_option('--test', dest="test", action='store_true', default=False,
                   help="Test Spyder with a clean settings dir")
-parser.add_option('--no-apport', action='store_true',
-                  default=False, help="Disable Apport exception hook (Ubuntu)")
 parser.add_option('--debug', action='store_true',
                   default=False, help="Run Spyder in debug mode")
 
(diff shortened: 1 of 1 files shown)


ccordoba12:
So how do you pass extra arguments to Spyder? Could you post a command of how bootstrap work now?

techtonik:
Nothing should have changed for users, except for removed `--no-apport` option, so the syntax for passing options is the same:

    python bootstrap.py -- --spyder

I thought to include `bootstrap.py` run to Travis/Appveyor sanity check, but I don't see any mode that just makes all imports ready, discovers all plugins and exits.

techtonik:
So, if everything is clear here, I am going to add second part..

Nodd:
With this spyder will import Qt twice: once for checking the Qt backend, and one in the subprocess for launching spyder. Did you check the timing to see the difference ?

techtonik:
On my system:
```
Bootstrap completed in 00:00:00.8040
...
Bootstrap completed in 00:00:00.7830
```
I can't install line_profiler because of https://github.com/rkern/line_profiler/issues/63 to say for sure that these are Qt imports that eating 0.8 second. That profiling is better to embed into Spyder itseld.

Anyway, I don't believe that it matters for development run with `bootstrap.py`. It is more important to get all dependencies right, which is going to be added next.

martinfiers:
Just wondering - why exactly is this required? Won't this make debugging of spyder more difficult? (in case you're used to setting breakpoints and interactive debugging).

techtonik:
This is required exactly for that purpose - so that development setup made by bootstrap.py matches external setup where main spyder executable is run.

I also planned to add "requirements" fetching and "virtualenv" creating to bootstrap.py, so that it was still fast `checkout and run` action.

ccordoba12:
Closing because this is really old and I don't see a good use case for it.

PR #3241, opened 2016-06-20
.travis.yml: Upload build wheels to tags

techtonik:
This needs some testing as I've never done this before and don't have access to repository.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,3 +51,14 @@ script:
   - ./continuous_integration/travis/modules_test.sh
   - python runtests.py
   - coveralls
+
+# deploy tagged releases to GitHub
+# https://docs.travis-ci.com/user/deployment/releases
+deploy:
+  provider: releases
+  api_key: "please run `travis setup releases` to make encrypted key"
+  file:
+    - ./dist/spyder-${TRAVIS_TAG}-py2.py3-none-any.whl
+  skip_cleanup: true
+  on:
+    tags: true


ccordoba12:
I don't think we really need this. Building and deploying wheels for Spyder (a pure Python package) is a very simple process and we don't need Travis for that.

Thanks for your efforts, but I'm closing this one :-)

techtonik:
So where are those wheel then? https://github.com/spyder-ide/spyder/releases

ccordoba12:
https://pypi.python.org/pypi/spyder/3.0.0b3

ccordoba12:
We will publish wheels for Spyder 3.0 and higher. We don't have wheels for Spyder 2.x

techtonik:
Nice. Those need to be included into release template https://github.com/spyder-ide/spyder/releases so  that people could discover them for testing.

ccordoba12:
I already mention how to use `pip` to install them. I don't think we need to say more, the process is very transparent for the user :-)

techtonik:
`pip` doesn't install beta wheels, so it is easier to just give it direct link.

techtonik:
Anyway, if you will need to save some hours with automation, Travis is also able to upload wheels to PyPI.

ccordoba12:
> pip doesn't install beta wheels

That's not true. If you run this command

    pip install --pre -U spyder

you'll get our beta wheels (and that command is mentioned in the release notes of beta3 :-)

techtonik:
Still it requires a lot of time (2 hours+) to discover that `pip install --pre -U spyder` is actually installing from binary wheels and doesn't try to build them from source.

Nodd:
Is it really a problem since spyder is pure python ? Anyway a quick look on pypi doesn't take 2 hours.

techtonik:
> Is it really a problem since spyder is pure python ?

And who knows that? Also, installing from source seems to ignore dependencies.

> Anyway a quick look on pypi doesn't take 2 hours.

Show me how quick look on pypi reveals beta wheels.

Nodd:
> Show me how quick look on pypi reveals beta wheels.

My bad, there is no link to other versions.

>> Is it really a problem since spyder is pure python ?

> And who knows that?

pip install works out of the box, that's all you need to know !

>  Also, installing from source seems to ignore dependencies.

Maybe because you have the choice between PyQt4, PyQt5 and PySide, so they ar not set as dependencies ?

PR #3246, opened 2016-06-22
github_issues.py - Fix grammar

jitseniesen:
Rather minor but the dodgy grammar in the release announcements was annoying me :-)

### DIFF
--- a/github_issues.py
+++ b/github_issues.py
@@ -7,7 +7,7 @@
 """
 Build a list of issues per milestone
 
-NOTE: This script works with githupy, not with PyGithub!
+NOTE: This script works with githubpy, not with PyGithub!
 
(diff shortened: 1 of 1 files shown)


goanpeca:
Thanks @jitseniesen  :-)

---

@ccordoba12, did not we take out this little script into a separate location (https://github.com/spyder-ide/loghub) for re-usability?

Should we just completely remove it?

ccordoba12:
@jitseniesen, I fixed the grammatical errors you reported in our Changelog in master, and also in

https://github.com/spyder-ide/loghub/blob/master/loghub.py

which is the replacement of `github_issues.py` and what we're using to generate Changelogs for our other projects (e.g. `qtpy`).

PR #3295, opened 2016-07-15
PR: Fix/Find in files

goanpeca:

### DIFF
--- a/spyderlib/config/utils.py
+++ b/spyderlib/config/utils.py
@@ -96,7 +96,7 @@ def _get_pygments_extensions():
             lexer_exts = [le for le in lexer_exts if not le.endswith('_*')]
             extensions = extensions + list(lexer_exts) + list(other_exts)
 
-    return sorted(extensions)
+    return sorted(list(set(extensions)))
 
 
 #==============================================================================
--- a/spyderlib/config/utils.py
+++ b/spyderlib/config/utils.py
@@ -80,15 +80,23 @@ def _get_pygments_extensions():
     """Return all file type extensions supported by Pygments"""
     # NOTE: Leave this import here to keep startup process fast!
     import pygments.lexers as lexers
+
     extensions = []
-    all_lexers = lexers.get_all_lexers()
-    for lx in all_lexers:
+    for lx in lexers.get_all_lexers():
         lexer_exts = lx[2]
+
         if lexer_exts:
-            lexer_exts = [le[1:] for le in lexer_exts]
+            # Reference: This line was included for leaving untrimmed the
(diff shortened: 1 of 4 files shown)


spyderlib/config/utils.py
```diff
@@ -80,15 +80,23 @@ def _get_pygments_extensions():
...
+            # Reference: This commented line was replaced by the following one
+            # to trim only extensions that start with '*'
+            # lexer_exts = [le[1:] for le in lexer_exts]
+            lexer_exts = [le[1:] for le in lexer_exts if le.startswith('*')]
             lexer_exts = [le for le in lexer_exts if not le.endswith('_*')]
-            extensions = extensions + list(lexer_exts)
-    return extensions
+            extensions = extensions + list(lexer_exts) + list(other_exts)
+
+    return sorted(list(set(extensions)))
```
ccordoba12: What do these changes do exactly?

spyderlib/plugins/findinfiles.py
```diff
@@ -94,14 +97,23 @@ def findinfiles_callback(self):
...
+        patterns = [r'|'.join([ext + r'$' for ext in clean_exts if ext]) +
                     r'|README|INSTALL',
-                    r'\.pyw?$|\.ipy$|\.txt$|\.rst$',
-                    '.']
+                    r'\.ipy$|\.pyw?$|\.rst$|\.txt$',
+                    '.',
+                    ]
         return patterns
-    
+
```
ccordoba12: Please remove this blank :-)

ccordoba12:
Does your changes hide the default (advanced) view of *Find in files*? If not, I'd like to see that as part of this PR ;-)

goanpeca:
Hi @ccordoba12 , I forgot to add the description. Basically the include regex where based on the extension provided by the pygments lexers, but the problem is that the generated regex was invalid. I corrected that and added some tests so we are sure that the Include and Exclude patterns are always valid regular expressions.

I can hide the pane by default. Working on a ack/grin like interface after this PR.

spyderlib/config/utils.py
goanpeca: `lexer_exts = [le[1:] for le in lexer_exts]`

This code is wrong cause not all lexers extensions start with a **`*`**, so I fixed that so the actual generated patterns are valid regex. (this was causing find in files to not work after a clean install)

goanpeca:
@ccordoba12, now the advanced options are hidden by default

ccordoba12:
I tested our edit filters and they are working correctly, so merging ;-)

PR #3296, opened 2016-07-17
PR: Don't hide calltip widget with a timer

### LINKED ISSUE #2644: Editor: Popup showing function arguments only displayed for 3 seconds
## Description

In the editor, when I type the name of a function and a left parenthesis, Spyder displays a helpful popup showing the function's arguments.  Unfortunately, this is only displayed for 3 seconds, and then disappears.  It would be helpful if this popup stayed visible for longer, possibly until I type the closing parenthesis or move the text cursor off of the line.

1. Type the name of a function, followed by the left parenthesis
2. The helpful popup appears
3. Wait 3 seconds
4. The helpful popup disappears, and I am sad
- Spyder Version: 3.0.0b1 (happens in 2.3.x as well)
- Python Version: 2.7.10
- Qt Versions: 4.8.6, PyQt4 (API v2) 4.10.4 on Windows
## Optional dependencies

```
pyflakes >=0.5.0:  0.9.2 (OK)
pep8 >=0.6      :  1.6.2 (OK)
pygments >=1.6  :  2.0.2 (OK)
IPython >=3.0   :  3.2.0 (OK)
zmq >=13.0.0    :  14.7.0 (OK)
pandas >=0.13.1 :  0.16.2 (OK)
sphinx >=0.6.6  :  1.3.1 (OK)
rope >=0.9.4    :  0.9.4 (OK)
jedi >=0.8.1    :  0.8.1 (OK)
matplotlib >=1.0:  1.4.3 (OK)
sympy >=0.7.3   :  0.7.6 (OK)
pylint >=0.25   :  None (NOK)

```

ccordoba12:
Fixes #2644 

----

New behavior in this PR:

1. Calltips won't be hidden unless a matching closing `)` is written by the user, or he/she presses these keys: up, down, or enter/return.
2. If a calltip is already shown, no new calltip is generated (unless there's new info to show).


jitseniesen:
Looks good to me. I tested it and it seems to work. Just a couple of minor points (to prove that I did look at it), none of which are critical IMO:

1. If you type something like `max((b+1)/2)`, then the tooltip for max disappears when you type the first closing parenthesis, while I would have expected it to stay open until you finish the argument list for max.
2. The flag `hide_timer_on` is always False now, so can we remove the timer logic?
3. The function `_find_parenthesis()` in `calltip.py` looks very similar, at least in purpose, as `find_brace_match` in `sourcecode/base.py` .

ccordoba12:
@jitseniesen, thanks for the review :-) My replies to your points:

1. I'm working on fixing this issue.
2. I'd prefer to leave it because I invested a lot of time working on the timer logic and we could also need it in the future for other widgets.
3. Yes. I'll see if I can use `find_brace_match` instead.

AIDAROIGCOMPTON:
Hello
<br>
I am also hoping to disable the 3-second timer, as I am VERY new at Spyder and Python.  Also very knew to gitHub, so I don't know if you've created a fix.  Please let me know the status of your resolution.
<br>
I thank you in advance for your time & effort!

ccordoba12:
@AIDAROIGCOMPTON, this is precisely the fix for the timer. It will be available in Spyder 3.1, to be released in December :-)

goanpeca:
@ccordoba12 is this ready to be merged but on hold to be merged when 3.2 is about to be released, or?

Needs a rebase / merge also

ccordoba12:
This is for 3.1. I'll come back to it after I release 3.0.2 :-)

ccordoba12:
@jitseniesen, I think this one is ready. Would you mind to give one last test? Thanks :-)

jitseniesen:
> If you type something like `max((b+1)/2)`, then the tooltip for max disappears when you type the first closing parenthesis, while I would have expected it to stay open until you finish the argument list for `max`.

This works now. However, if the function is within parentheses, then the tooltip stays visible until you close the outer paranthesis. Example: `2*(max(b)+1)`, the tooltip for `max` stays visible when you complete the argument list and type the first closing parenthesis, and it only goes away after you move beyond the second closing parenthesis. 

Ideally, when Spyder displays the tooltip, it saves the position of the opening parenthesis (which it may well already do), and in the test for whether to hide the calltip, it tests whether the closing parenthesis matches up with the opening parenthesis which caused the tooltip to be displayed.

But maybe this is a separate issue (it looks like it never worked like I described; it just is more apparent now that the calltip doesn't disappear after 3 secs) and should be handled by another PR.

No further comments.

ccordoba12:
@jitseniesen, thanks for the review. I'm aware of the problem you mention and I'll open an issue about it so we can fix it in 3.1.1 ;-)

PR #3313, opened 2016-07-20
Toggle file switcher when pressing its keyboard shortcut

### LINKED ISSUE #3297: Pressing Ctrl+P twice opens file switcher twice

1. Press Ctrl-P
2. Press Ctrl-P again

The first Ctrl-P opens the selection pop up. The second Ctrl-P should (preferably) close it while open, or at least do nothing. Instead, it seems a second pop up is opened over the first one (you then have to press Esc twice to close both of them).

* Spyder Version: Spyder 3.0.0.dev0 (from `spyder-ide` conda channel)
* Python Version: 3.5.2
* Operating system: Linux

## Dependencies

and paste the contents below:

matplotlib >=1.0 :  None (NOK)
numpy >=1.7      :  None (NOK)
psutil >=0.3     :  4.3.0 (OK)
pyflakes >=0.6.0 :  1.2.3 (OK)
sphinx >=0.6.6   :  1.4.1 (OK)

ccordoba12:
Fixes #3297

### DIFF
--- a/spyderlib/widgets/editor.py
+++ b/spyderlib/widgets/editor.py
@@ -581,10 +581,16 @@ class EditorStack(QWidget):
         """Open file list management dialog box"""
         if not self.tabs.count():
             return
+        if self.fileswitcher_dlg is not None and \
+          self.fileswitcher_dlg.is_visible:
+            self.fileswitcher_dlg.hide()
+            self.fileswitcher_dlg.is_visible = False
+            return
         self.fileswitcher_dlg = FileSwitcher(self, self.tabs, self.data)
         self.fileswitcher_dlg.sig_goto_file.connect(self.set_stack_index)
         self.fileswitcher_dlg.sig_close_file.connect(self.close_file)
         self.fileswitcher_dlg.show()
+        self.fileswitcher_dlg.is_visible = True
 
     def update_fileswitcher_dlg(self):
         """Synchronize file list dialog box with editor widget tabs"""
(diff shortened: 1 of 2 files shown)


spyderlib/widgets/editor.py
```diff
@@ -581,10 +581,16 @@ def open_fileswitcher_dlg(self):
         """Open file list management dialog box"""
         if not self.tabs.count():
             return
+        if self.fileswitcher_dlg is not None and \
+          self.fileswitcher_dlg.is_visible:
+            self.fileswitcher_dlg.hide()
+            self.fileswitcher_dlg.is_visible = False
```
Nodd: This is already supported by Qt, with the `visible` property. See http://doc.qt.io/qt-4.8/qwidget.html#visible-prop

ccordoba12: Ops, I didn't know about it. I looked for `isVisible` but it seems QDialogs don't support it.

I'll change it in master.

ccordoba12: It doesn't with Qt5, so it was probably only available with Qt4.

ccordoba12: So I'll leave things as they are ;-)

Nodd: It is still present in Qt5: http://doc.qt.io/qt-5.6/qwidget.html#visible-prop (and QDialog inherits QWidget). So it should work...
Anyway it's not a big deal, this comment is for completeness only.

ccordoba12: I tried to use `visible`, believe me ;-) It just didn't work, I don't know why...

Nodd: I think I found why. The widget is created on demand, and during initialization the widget is not visible yet so a check with `visible()` will be wrong. Your code is as good as any other workaround.

ccordoba12: Ok, good to know ;-)

PR #3324, opened 2016-07-25
Make 'run line' add a blank line if on last line.

### LINKED ISSUE #3293: F9 does not auto advance when it's at the last line

When the cursor is at the last line of code, F9 will execute that line but does not advance to a new (blank) line

F9 should always advance to the new blank line given the interactive data analysis workflow. I expect to be able to write new line of code, hit F9 to execute and advance to a new line at the same time. (Rstudio also has this behavior).

Currently, the analyst has to write, hit Enter to advance to a new line, then after a while go back to execute all the lines that have just been written. The third step is unnecessary and breaks the workflow.

* Spyder Version: 3.0 beta
* Python Version: 2.7
* Operating system: Ubuntu 14.04

jitseniesen:
Fixes #3293: F9 does not auto advance when it's at the last line. Also add tests.

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -8,8 +8,10 @@
 File for running tests programmatically.
 """
 
+# Standard library imports
+import sys
+
 # Third party imports
-import qtpy
 import pytest
 
 
@@ -17,9 +19,9 @@ def main():
     """
     Run pytest tests.
     """
-    pytest.main(['-x', 'spyderlib',  '-v', '-rw', '--durations=10',
-                 '--cov=spyderlib', '--cov-report=term-missing'])
-
+    errno = pytest.main(['-x', 'spyderlib',  '-v', '-rw', '--durations=10',
+                         '--cov=spyderlib', '--cov-report=term-missing'])
+    sys.exit(errno)
 
 if __name__ == '__main__':
     main()
--- a/runtests.py
+++ b/runtests.py
@@ -12,6 +12,7 @@ File for running tests programmatically.
 import sys
 
 # Third party imports
+import qtpy  # to ensure that Qt4 uses API v2
 import pytest
 
 
--- a/spyderlib/widgets/editor.py
+++ b/spyderlib/widgets/editor.py
@@ -1837,7 +1837,7 @@ class EditorStack(QWidget):
         if text:
             self.exec_in_extconsole.emit(text, self.focus_to_editor)
         if editor.is_cursor_on_last_line() and text:
-            editor.append('\n')
+            editor.append(editor.get_line_separator())
         editor.move_cursor_to_next('line', 'down')
 
     def run_cell(self):
--- a/spyderlib/widgets/editor.py
+++ b/spyderlib/widgets/editor.py
@@ -1820,10 +1820,12 @@ class EditorStack(QWidget):
         Run selected text or current line in console. 
 
         If some text is selected, then execute that text in console.
-        If no text is selected, then execute current line and advance cursor
-        to next line. If cursor is on last line and that line is not empty,
-        then add a new blank line and move the cursor there. If cursor is on
-        last line and that line is empty, then do not move cursor.
+
+        If no text is selected, then execute current line, unless current line
+        is empty. Then, advance cursor to next line. If cursor is on last line
+        and that line is not empty, then add a new blank line and move the
+        cursor there. If cursor is on last line and that line is empty, then do
+        not move cursor.
         """
         text = self.get_current_editor().get_selection_as_executable_code()
         if text:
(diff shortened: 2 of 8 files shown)


continuous_integration/appveyor/run_test.bat
```diff
@@ -26,4 +26,8 @@ echo %time%
 echo Success!
 echo %time%
 echo.
-echo ---------------------------------
+
+echo ------- Running py.test ---------
+echo.
+cd %APPVEYOR_BUILD_FOLDER%
+python runtests.py
```
ccordoba12: Please move this to `appveyor.yml` to have the same configuration as in Travis:

https://github.com/spyder-ide/spyder/blob/master/.travis.yml#L52

ccordoba12:
AppVeyor got stuck while running `spyderlib/utils/tests/test_programs.py` so maybe this test shouldn't be ran on Windows.

jitseniesen:
@ccordabo Thanks. Appveyor should skip those tests now. But it would be good to figure out what is going wrong. Out of curiosity, how can I find out where Appveyor got stuck?

ccordoba12:
@jitseniesen, your last commit revealed an error in our tests. You need to move this import:

https://github.com/spyder-ide/spyder/blob/master/spyderlib/utils/tests/test_syntaxhighlighters.py#L8

after the `qtpy` imports so that the PyQt API is set correctly to API `#2` for PyQt4.

----

Please also check that that's the case for the rest of our tests.

jitseniesen:
@ccordoba12 Thanks, I was wondering what was happening ... I'll have a go at it

jitseniesen:
Phew, all tests past finally. The actual fix for the issue was quite easy, but the tests took me ten times longer! Let's hope it will prove to be a good investment.

@ccordoba12 Moving the `qtpy` import up did not do the trick; it looks like pytest-qt sets PyQt API to v1 on Python 2 before I can do anything about it.

ccordoba12:
Thanks a lot for your hard work on making our tests to work correctly!! I know this kind of job takes a lot of time ;-)

goanpeca:
@jitseniesen awesome :-), it does get easier! and we all have battle scars (some more than others :wink:)

runtests.py
```diff
@@ -8,18 +8,20 @@
 File for running tests programmatically.
 """
 
+# Standard library imports
+import sys
+
 # Third party imports
-import qtpy
 import pytest
```
ccordoba12: @jitseniesen, what if we move

    PYTEST_QT_API=pyqt4v2

to here (using `os.environ`, of course)?

That way our tests would run locally too. I mean, right now they only run correctly in Travis and Appveyor :-)

goanpeca: And they could run on circle-ci but someone disable that ...

jitseniesen: That does not work. But it suggested to me that we should intervene in before the `import pytest` in `runtests.py` so I'm trying that now.

Interesting that you are thinking of running `runtests.py` locally; I find it print too much cruft to be useful.

spyderlib/widgets/editor.py
```diff
@@ -1817,20 +1817,28 @@ def fix_indentation(self, index=None):
...
-            editor.move_cursor_to_next('line', 'down')
+            self.exec_in_extconsole.emit(text, self.focus_to_editor)
+            return
+        editor = self.get_current_editor()
+        line = editor.get_current_line()
+        text = line.lstrip()
+        if text:
+            self.exec_in_extconsole.emit(text, self.focus_to_editor)
+        if editor.is_cursor_on_last_line() and text:
+            editor.append('\n')
```
ccordoba12: Please change this to

    editor.append(editor.get_line_separator())

to preserve line endings per operating system.

spyderlib/widgets/tests/test_editor.py
```diff
@@ -0,0 +1,88 @@
...
+
+def test_run_last_line_when_nonempty(qtbot):
+    editorStack, editor = setup_editor(qtbot)
+    editor.stdkey_backspace() # delete empty line at end
+    old_text = editor.toPlainText()
+    with qtbot.waitSignal(editorStack.exec_in_extconsole) as blocker:
+        editorStack.run_selection()
+    assert blocker.signal_triggered
+    assert blocker.args[0] == 'x = 2'
+    assert editor.toPlainText() == old_text + '\n' # check blank line got added
```
ccordoba12: This should ne changed too to

    ... == old_text + editor.get_line_separator()

spyderlib/widgets/editor.py
jitseniesen: Thanks, done

PR #3331, opened 2016-07-28
PR: Add Shortcuts Summary Window

### LINKED ISSUE #3275: Print all keyboard shortcuts to cheatset
Spyder has may good shortcuts. But they are well hidden in the preferences dialogue

**Tools --> Preferences --> Keyboard shortcuts?**

Solution:
add a print (to PDF) function to the keyboard settings so that users can have these open on a separate file while learning the programme.

References:
https://code.google.com/archive/p/spyderlib/issues/578
https://github.com/spyder-ide/spyder/issues/578

- Spyder Version: 3.0.dev
- Python Version: 2.7
- Operating system: linux, ubuntu mint 17.3
## Dependencies

nbconvert >=4.0  :  4.0.0 (OK)
numpy >=1.7      :  1.10.4 (OK)
pandas >=0.13.1  :  0.17.1 (OK)
psutil >=0.3     :  3.3.0 (OK)
pyflakes >=0.5.0 :  1.0.0 (OK)
pygments >=2.0   :  2.0.2 (OK)
sphinx >=0.6.6   :  1.3.1 (OK)
sympy >=0.7.3    :  0.7.6.1 (OK)

rlaverde:
Fixes #3275 

----

Summary shortcuts implementation

I'm still working on it, there is some things missing:
- [x] register shortcut
- [x] calculate font size and amount of elements in a column using screen size

shortcuts_draft1

### DIFF
--- a/spyder/widgets/shortcuts.py
+++ b/spyder/widgets/shortcuts.py
@@ -17,22 +17,27 @@ from itertools import groupby
 from qtpy.QtCore import Qt
 from qtpy.QtGui import QFont
 from qtpy.QtWidgets import (QDialog, QLabel, QGridLayout, QGroupBox,
-                            QVBoxLayout, QHBoxLayout)
+                            QVBoxLayout, QHBoxLayout, QDesktopWidget)
 
 # Local imports
 from spyder.config.base import _
 from spyder.config.gui import iter_shortcuts
 
 # Constants
-SHORTCUT_SHOURTCUTS_SUMMARY = "Ctrl+H"
-MAX_SHOURCUTS_COLUMN = 28
-
+MAX_FONT_SIZE = 16
+MIN_FONT_SIZE = 8
 
 class ShortCutsSummaryDialog(QDialog):
     def __init__(self, parent=None, offset=0, force_float=False):
         QDialog.__init__(self, parent=parent)
 
         self._shortcuts_summary_title = _("Spyder Keyboard ShortCuts")
+
+        # Calculate font and amount of elements in each column according screen size
+        width, height = self.get_screen_resolution()
+        font_size = height / 80
+        font_size = max(min(font_size, MAX_FONT_SIZE), MIN_FONT_SIZE)
+        shortcuts_column = (height - 8 * font_size) / (font_size +16)
 
         # Widgets
         style = """
@@ -43,14 +48,20 @@ class ShortCutsSummaryDialog(QDialog):
             }"""
         self.setStyleSheet(style)
 
-        groupbox_style = "QGroupBox { font-weight: bold;}"
+        font_names = QFont()
+        font_names.setPointSize(font_size)
+        font_names.setBold(True)
 
-        font_titles = QFont()
-        font_titles.setBold(True)
+        font_keystr = QFont()
+        font_keystr.setPointSize(font_size)
+
+        font_title = QFont()
+        font_title.setPointSize(font_size+2)
+        font_title.setBold(True)
 
         title_label = QLabel(self._shortcuts_summary_title)
         title_label.setAlignment(Qt.AlignCenter)
-        title_label.setFont(font_titles)
+        title_label.setFont(font_title)
 
         # iter over shortcuts and create GroupBox for each context
         # with shortcuts in a grid
@@ -75,22 +86,23 @@ class ShortCutsSummaryDialog(QDialog):
                     if context == '_': context = 'Global'
 
                     group = QGroupBox(context.capitalize())
-                    group.setStyleSheet(groupbox_style)
+                    group.setFont(font_names)
 
                     group_layout = QGridLayout()
                     group.setLayout(group_layout)
 
                 # Widgets
                 label_name = QLabel(name.capitalize().replace('_', ' '))
-                label_name.setFont(font_titles)
+                label_name.setFont(font_names)
                 label_keystr = QLabel(keystr)
+                label_keystr.setFont(font_keystr)
 
                 group_layout.addWidget(label_name, i, 0)
                 group_layout.addWidget(label_keystr, i, 1)
 
                 added_shortcuts += 1
 
-                if added_shortcuts >= MAX_SHOURCUTS_COLUMN:
+                if added_shortcuts >= shortcuts_column:
                     column_layout.addWidget(group)
                     columns_layout.addLayout(column_layout)
                     added_shortcuts = 0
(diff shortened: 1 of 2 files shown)


goanpeca:
@spyder-ide/core-developers should we create a 3.x branch so we don't leave opened PRs without supervision for weeks or months until 3.0 is released?

goanpeca:
@Nodd comments?

ccordoba12:
@jitseniesen, it would be nice to hear from you here too :-)

jitseniesen:
Looks quite decent to me. Well done @rlaverde ! But I do have a couple of points ...

1. I might have missed something, but it is not clear from the code how the user is going to get this window.
2. How does this work on a small screen? The window looks rather big and I think it does not always fit.
3. Can we get rid of the underscore in "Switch to variable_explorer" etc?
4. Probably not in the scope of this PR, but I find the inconsistency between "Shift+Ctrl+F4" and "Ctrl+Shift+F12" a bit annoying.
5. What about fixed (not configurable) shortcuts?
6. Looking at the code, is it necessary to define `LABELS`? What if a new context is added, e.g. in a plugin (once we get shortcuts in plugins working properly)? I would special case `'_'` and otherwise just capitalize.

rlaverde:
> I might have missed something, but it is not clear from the code how the user is going to get this window.
I haven't implemented this, I'm not sure where I should register this shortcut (gui? editor?), and which should be the shortcut (Ctrl+H maybe?)

> How does this work on a small screen? The window looks rather big and I think it does not always fit.
hmm yes, font and windows size should be calculated from screen size (I'm working on this)

>Can we get rid of the underscore in "Switch to variable_explorer" etc?

Yes, I added a commit that replace underscores by spaces

> Probably not in the scope of this PR, but I find the inconsistency between "Shift+Ctrl+F4" and "Ctrl+Shift+F12" a bit annoying.

Yes ,I also think this should be in another PR, maybe modify iter_shourtcuts() function in spyderlib/config/gui.py 

>What about fixed (not configurable) shortcuts?

Fixed shortcuts aren't save anywhere,  I think this will be another PR too, that add a function iter_fixed_shortcuts() and modify fixed_shortcut to save it in a global variable maybe.

> Looking at the code, is it necessary to define LABELS? What if a new context is added, e.g. in a plugin (once we get shortcuts in plugins working properly)? I would special case '_' and otherwise just capitalize.

I just add another commit, I deleted LABELS and special case '_' as you advised

jitseniesen:
> I haven't implemented this, I'm not sure where I should register this shortcut (gui? editor?), and which should be the shortcut (Ctrl+H maybe?)

I would make it a global shortcut because the info is about the whole application. Ctrl+H sounds good (I don't think it conflicts with anything), but maybe we want to prefer double modifiers for global shortcuts and keep the single modifier shortcuts as much as possible available for local use. In which case, what about Alt-Shift-H or Ctrl-Shift-K (for keyboard shortcuts)? Ctrl-Shift-H is already taken.

I would also add an item in the help menu, otherwise people will never discover this new feature.

ccordoba12:
@rlaverde, please merge with master to update your work :-)

rlaverde:
> @rlaverde, please merge with master to update your work :-)

I rebase it and also squash some commits

> I would also add an item in the help menu, otherwise people will never discover this new feature.

I added a new element as you suggested, and define Meta+F1 as the shortcut

ccordoba12:
@rlaverde, thanks a lot! The error on Appveyor is cause by errors with conda, don't worry about it :-)

spyder/widgets/shortcuts.py
```diff
@@ -0,0 +1,140 @@
...
+
+# Local imports
+from spyder.config.base import _
+from spyder.config.gui import iter_shortcuts
+
+# Constants
+MAX_FONT_SIZE = 16
+MIN_FONT_SIZE = 8
+
+class ShortCutsSummaryDialog(QDialog):
```
jitseniesen: Please add a docstring with some information about the dialog, what it contains and what it looks like.

spyder/widgets/shortcuts.py
```diff
@@ -0,0 +1,140 @@
...
+    def __init__(self, parent=None, offset=0, force_float=False):
```
jitseniesen: The last two arguments are not used; can they be deleted?

spyder/widgets/shortcuts.py
```diff
@@ -0,0 +1,140 @@
...
+        QDialog.__init__(self, parent=parent)
+
+        self._shortcuts_summary_title = _("Spyder Keyboard ShortCuts")
+
+        # Calculate font and amount of elements in each column according screen size
+        width, height = self.get_screen_resolution()
+        font_size = height / 80
+        font_size = max(min(font_size, MAX_FONT_SIZE), MIN_FONT_SIZE)
+        shortcuts_column = (height - 8 * font_size) / (font_size +16)
```
jitseniesen: This computation results in a window that is too tall on my laptop screen, which has:
```
width = 1366 
height = 737
font_size = 9.2125 
shortcuts_column = 26.30837878036688
```
The resulting window has height 771 (and width 1157).

spyder/widgets/shortcuts.py
```diff
@@ -0,0 +1,140 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © 2009- The Spyder Development Team
+# Licensed under the terms of the MIT License
+# (see spyderlib/__init__.py for details)
+
+"""Shortcut Summary dialog"""
+
+# TODO:
+# calculate windows size from screen size
```
jitseniesen: Is this TODO comment out of date?

spyder/widgets/shortcuts.py
```diff
@@ -0,0 +1,140 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © 2009- The Spyder Development Team
+# Licensed under the terms of the MIT License
+# (see spyderlib/__init__.py for details)
```
jitseniesen: Replace spyderlib by spyder

jitseniesen:
I forgot to say the most important thing: It does work, and it seems almost ready.

By the way, I wonder whether we need to include a close button or something like that. At the moment, the user needs to press `Esc` to close the window, which is not totally obvious.

rlaverde:
@jitseniesen For advise of @ccordoba12 I open a new PR to the 3.x branch, I will fix the things that you commented in the new PR

spyder/widgets/shortcuts.py
rlaverde: I don't know how I could fix this, my laptop have similar dimensions, and the resulting height is 703, there's another way to add elements and avoid oversizing the window?

width = 1366
height = 738
font_size = 9.225
shortcuts_column = 26.3310

jitseniesen: I don't know. I don't understand where you got the formula for `shortcuts_column` from. Maybe I have different fonts installed and the font I end up with has a different size?

Perhaps it's best not to be too smart. Pack all items in four columns of (rougly) equal size, and add a scroll bar if it ends up bigger than the screen.

Alternatively, add the items to the column one-by-one and measure the height of the resulting column; hopefully there is a field `.height` in Qt, but I don't know whether this is possible without displaying it on the screen, and it may be too much effort for the first iteration.

rlaverde: I'll try the second approach if that doesn't work, I'll implement the other one

PR #3333, opened 2016-07-29
Enable high DPI scaling on Qt >= 5.6

### LINKED ISSUE #2119: Spyder doesn't render well on HighDpi screens
_From [mjdep...@gmail.com](https://code.google.com/u/112397940646561315916/) on 2015-01-15T10:51:45Z_

Please note that reporting bugs from Spyder's help menu ("?" > "Report
issue...") will fill version details for Spyder, Python and Qt in this description for you. 

 1. Using a 4K screen with 3840 x 2160 2. 3. 

 expect size of windows and consoles to look normal. The biggest issue is in the variable explorer the headers are cut off height-wise. The preference window comes up really small. What version of the product are you using? On what operating system? spyder 2.3.2 py3.4 windows 8.1 

. I am stuck in a class and super bored right now. In reality these issues are incredibly minor. But would be cool if they were fixed. Thanks!

**Attachment:** [spyder screen shot.PNG](http://code.google.com/p/spyderlib/issues/detail?id=2119)

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=2119_

patstew:
Fixes #2119

### DIFF
--- a/spyderlib/app/spyder.py
+++ b/spyderlib/app/spyder.py
@@ -90,6 +90,12 @@ from qtpy import QtSvg  # analysis:ignore
 # Avoid a bug in Qt: https://bugreports.qt.io/browse/QTBUG-46720
 from qtpy import QtWebEngineWidgets  # analysis:ignore
 
+#==============================================================================
+# Proper high DPI scaling is available in Qt >= 5.6.0. This attibute must
+# be set before creating the application
+#==============================================================================
+if hasattr(Qt, 'AA_EnableHighDpiScaling'):
+    QCoreApplication.setAttribute(Qt.AA_EnableHighDpiScaling, True)
 
 #==============================================================================
 # Create our QApplication instance here because it's needed to render the


ccordoba12:
@patstew, thanks a lot for working on this!! Could we say that this fixes issue #2119? The other minor issues you report could be fixed in different PRs :-)

goanpeca:
Could we use a different version comparison thta using QTVersion  Should we use the qtpy ones?

ccordoba12:
@goanpeca, it requires a bit more code, but I was thinking that too. I'll add a comment to the diff :-)

goanpeca:
@ccordoba12 , or we could do it the way @SylvainCorlay did,

```python
    # For retina displays on qt5
    if hasattr(Qt, 'AA_UseHighDpiPixmaps'):
        app.setAttribute(Qt.AA_UseHighDpiPixmaps)

    if hasattr(Qt, 'AA_EnableHighDpiScaling') and WIN:
        app.setAttribute(Qt.AA_EnableHighDpiScaling)

**Comment by ccordoba12** [comment]
It doesn't work like that because AA_EnableHighDpiScaling needs to be 
set before the application is created.

File: spyderlib/app/spyder.py [hunk 1]
```diff
@@ -90,6 +90,12 @@
 # Avoid a bug in Qt: https://bugreports.qt.io/browse/QTBUG-46720
 from qtpy import QtWebEngineWidgets  # analysis:ignore
 
+#==============================================================================
+# Proper high DPI scaling is available in Qt >= 5.6.0. This attibute must
+# be set before creating the application
+#==============================================================================
+if QT_VERSION >= 0x050600:
+    QCoreApplication.setAttribute(Qt.AA_EnableHighDpiScaling, True)
```
ccordoba12: @patstew, we prefer to avoid a comparison with an hex number, so please write this code like

```python
if hasattr(Qt, 'AA_EnableHighDpiScaling'):
    QCoreApplication.setAttribute(Qt.AA_EnableHighDpiScaling, True)
```

goanpeca:
Sure, on Navigator when I use that the application is already created

patstew:
There you go, I've changed it to use hasattr instead. The attribute has no effect if you run it after a QGuiApplication constructor.

ccordoba12:
Yes, the Qt docs are quite clear about it:

http://doc.qt.io/qt-5/qt.html

(Look for Qt::AA_EnableHighDpiScaling)

PR #3355, opened 2016-08-16
Some improvements to our file switcher

### LINKED ISSUE #3274: Text not visible in new file switcher in KDE
## Description

1. Ctrl+P

I expect to see the text, but instead I see part of the first character and a white rectangle:
spyder

Entering text changes the first letter displayed but it still not possible to see the full text.

* Spyder Version: 3.0.0.dev0
* Python Version: 3.5.1
* Qt Versions: 5.6.0, PyQt5 5.6 on Linux (KDE)

## Dependencies
```
pyflakes >=0.6.0 :  1.2.3 (OK)
numpy >=1.7      :  1.11.0 (OK)
sphinx >=0.6.6   :  1.4.1 (OK)
psutil >=0.3     :  4.1.0 (OK)

```

ccordoba12:
Fixes #3274 

----

* This improves our file swticher to make it more similar to the Sublime and VisualCode ones. In particular, I decided to:
    * Remove line numbers for symbols. This way users can see much more symbols in the dialog.
    * Use the system font for symbols (it was used already for files).
    * Remove tokens for symbols (i.e. `def`, `class`, etc).
    * Make the initial dialog size to be about the same for files and symbols.
    * Use FontAwesome icon for files instead of the one assigned to the Editor tabs.
* I was a bit dissatisfied with the way the switcher looked before, but now I think it looks quite good :-)
* Pinging @goanpeca and @Nodd about it.

----

## Before

seleccion_002

seleccion_003

## After

seleccion_004

seleccion_005

### DIFF
--- a/spyderlib/widgets/fileswitcher.py
+++ b/spyderlib/widgets/fileswitcher.py
@@ -457,9 +457,9 @@ class FileSwitcher(QDialog):
 
     # --- Helper methods: Outline explorer
     def get_symbol_list(self):
-        """Get the object explorer data."""
+        """Get the list of symbols present in the file."""
         try:
-            oedata = self.get_editor().highlighter.get_outlineexplorer_data()
+            oedata = self.get_editor().get_outlineexplorer_data()
         except AttributeError:
             oedata = {}
         return oedata
--- a/spyderlib/widgets/fileswitcher.py
+++ b/spyderlib/widgets/fileswitcher.py
@@ -541,7 +541,7 @@ class FileSwitcher(QDialog):
                 results.append((score_value, line, text, rich_text,
                                 fold_level, icons[index], token))
 
-        template_1 = '<code>{0}<big>{1}</big></code>'
+        template_1 = '<code>{0}</code>{1}'
 
         for (score, line, text, rich_text, fold_level, icon,
              token) in sorted(results):
--- a/spyderlib/widgets/fileswitcher.py
+++ b/spyderlib/widgets/fileswitcher.py
@@ -504,6 +504,9 @@ class FileSwitcher(QDialog):
             self.list.addItem(item)
             self.filtered_path.append(path)
 
+        # To adjust the delegate layout for KDE themes
+        self.list.files_list = True
+
         # Move selected item in list accordingly and update list size
         if current_path in self.filtered_path:
             self.set_current_row(self.filtered_path.index(current_path))
@@ -552,6 +555,9 @@ class FileSwitcher(QDialog):
             item = QListWidgetItem(icon, textline)
             item.setSizeHint(QSize(0, 16))
             self.list.addItem(item)
+
+        # To adjust the delegate layout for KDE themes
+        self.list.files_list = False
 
         # Move selected item in list accordingly
         # NOTE: Doing this is causing two problems:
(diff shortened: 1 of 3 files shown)


goanpeca:
@ccordoba12 

My comments:

> - Remove line numbers for symbols. This way users can see much more symbols in the dialog.

Ok

> - Use the system font for symbols (it was used already for files).

Ok

> Remove tokens for symbols (i.e. def, class, etc).

Ok, but I actually like the def class token and I would have liked to be able to filter by that as well... (the current behavior does not do that anyway...)

> - Make the initial dialog size to be about the same for files and symbols.

Ok

-> Use FontAwesome icon for files instead of the one assigned to the Editor tabs

I really dislike Font Awesome icons for tokens.... they just look horrible.

ccordoba12:
> Ok, but I actually like the def class token and I would have liked to be able to filter by that as well... (the current behavior does not do that anyway...)

We could add a `type:` order to the switcher to filter by token types (and it shouldn't be that hard to do), but that's for another PR. For now all I want is to have the simplicity of other switchers

> I really dislike Font Awesome icons for tokens

Yep, I don't like them either. We need to talk to Sylvain about it :-)

goanpeca:
The idea with projects would be to also display project files (not opened?)

ccordoba12:
And VSCode has a very good syntax for searching for tokens: `@:`. After writing that, its switcher sorts and filters symbols by type.

goanpeca:
>Yes, but using a command. It could something like p: or pf: for project or project files

Can you just get something right away on sublime? (I mean without the p: or pf: ?)

spyderlib/widgets/fileswitcher.py
```diff
@@ -443,7 +443,11 @@ def goto_line(self, line_number):
     # --- Helper methods: Outline explorer
     def get_symbol_list(self):
         """Get the object explorer data."""
-        return self.get_editor().highlighter.get_outlineexplorer_data()
+        try:
+            oedata = self.get_editor().highlighter.get_outlineexplorer_data()
```
goanpeca: Seeing this long call makes me wonder, why we dont have a simpler method... like

```python
self.outlineexplorer_data()
```
not a fan of `get` in methods, this is part of my argument against using PEP8 for naming stuff... instead of Qt conventions. Qt conventions are quit nice and clear.

Nodd:
I love the new look. Could you consider displaying the line numbers in brackets after the names (on the same line) ?

goanpeca:
Agree with @nodd suggestion!

spyderlib/widgets/fileswitcher.py
ccordoba12: Ok, will do ;-)

ccordoba12:
Ok, no problem with that. I'll restore line numbers, but next to the token name :-)

goanpeca:
I think the problem with that (next to the name) is that they dont line up..... which is visually annoying, please post a screenshot when you do the change

ccordoba12:
> Can you just get something right away on sublime? (I mean without the p: or pf: ?)

You're right. I've verified it in VSCode. Something I don't like about VSCode is that it tries to open *any* kind of files, but we could limit that to just plain text files :-)

Nodd:
Maybe right align the line numbers ? I don't know if it's doable.

ccordoba12:
> Maybe right align the line numbers ? I don't know if it's doable.

I couldn't make it work, sorry. Maybe you can give it a try after I merge this one.

PR #3356, opened 2016-08-16
Remove icons from tabs for the Editor and IPython Console

ccordoba12:
This

* ~~Removes the left corner button from tabs, which contained a menu to browse them.~~ 
* Removes icons from our tabs (except for the Python Console, where they are useful because we have several types of tabs).

We talked about this with @goanpeca during SciPy and we think this helps to simplify how Spyder looks by removing visual clutter from the interface. It also improves our UI because the icons for the IPython console and Python files on the Editor are a bit ugly.

----

seleccion_001

seleccion_002

### DIFF
--- a/spyderlib/plugins/history.py
+++ b/spyderlib/plugins/history.py
@@ -13,8 +13,8 @@ import sys
 # Third party imports
 from qtpy import PYQT5
 from qtpy.QtCore import Signal, Slot
-from qtpy.QtWidgets import (QFontDialog, QGroupBox, QHBoxLayout, QInputDialog,
-                            QMenu, QToolButton, QVBoxLayout, QWidget)
+from qtpy.QtWidgets import (QGroupBox, QHBoxLayout, QInputDialog, QMenu,
+                            QToolButton, QVBoxLayout, QWidget)
 
 
 # Local imports
@@ -78,7 +78,6 @@ class HistoryLog(SpyderPluginWidget):
         
         self.editors = []
         self.filenames = []
-        self.icons = []
         if PYQT5:        
             SpyderPluginWidget.__init__(self, parent, main = parent)
         else:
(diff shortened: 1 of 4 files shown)


goanpeca:
:+1: !

Nodd:
The menu is useful for file switching in the editor, is it removed too ? I often have lots of opened files and switching between them is not that easy.

Not sure about removing the icon for files, it's important to know at a glance if a tab contains a python file or something else. But it's not that important.

goanpeca:
>The menu is useful for file switching in the editor,

I would say that the prefered method is to use now Ctrl+P (File switcher ... it will also keep your hands in the keyboard). Using that menu requires using the mouse. We talked with @ccordoba12 on ways of making the UI less crowded and this is definitely a way of achieving this. 

>Not sure about removing the icon for files, it's important to know at a glance if a tab contains a python file or something else. But it's not that important.

Not entirely sure about this, but it does help in streamlining the UI, so I am in favor of that. 

---

I would like spyder to keep moving in a direction where we remove complexity instead of adding complexity to the UI. This might mean rethinking stuff or removing things.

jitseniesen:
I agree that currently the UI looks cluttered and needs to be simplified. For that reason, removing the icons looks like a good idea to me. In principle, I also like the idea of removing the menu. However, I find that I still use the menu to switch tabs instead of Ctrl-P. I think the main reason for this is that the file switcher shows only four file names.

Nodd:
Some people don't like to use the keyboard only when programming. Ctrl+p is nice for sublime users, bot I personally can't use sublime because of that.

A related problem is that I have a visual memory, so it helps me a lot to see the full list of opened files. I never think about the characters in the filename but I tend to consider the name as a whole, so it's not practical to type a few letters.

All this to say that people work differently and having multiple options to do the same thing is not necessarily a bad thing. I'm nll for removing clutter but whatch out to not go the gnome way, it didn't go very well.

ccordoba12:
> I agree that currently the UI looks cluttered and needs to be simplified.

What other things would you like to see remove @jitseniesen?

> I think the main reason for this is that the file switcher shows only four file names.

I'll add more entries to it in PR #3355.

----

> All this to say that people work differently and having multiple options to do the same thing is not necessarily a bad thing.

So we're tied: 2-2 to remove/leave that button.I think we should restore it then, but only for the Editor or all our other widgets?

goanpeca:
Also, another change that will make it easier for users is when the tabbar is reworked so it can stay horizontal and in the left, like several users have already requested.

ccordoba12:
Ok, I'm going to restore "Browse tabs" menus for all tabbed plugins (I think it's useful to have them in our consoles too).

Everyone agrees?

goanpeca:
ok

PR #3386, opened 2016-08-31
Allow plugins with no dockwidgets

Nodd:
- ~~Restore plugin compatibility (`PluginConfigPage` was removed from `__init__.py`)~~
- Allow plugins with no dockwidgets (set to None for example, the attribute must exist for now)

### DIFF
--- a/spyder/plugins/__init__.py
+++ b/spyder/plugins/__init__.py
@@ -518,6 +518,8 @@ class SpyderPluginMixin(object):
     
     def toggle_view(self, checked):
         """Toggle view"""
+        if not self.dockwidget:
+            return
         if checked:
             self.dockwidget.show()
             self.dockwidget.raise_()
--- a/spyder/plugins/__init__.py
+++ b/spyder/plugins/__init__.py
@@ -37,6 +37,7 @@ from spyder.config.user import NoDefault
 from spyder.py3compat import configparser, is_text_string
 from spyder.utils import icon_manager as ima
 from spyder.utils.qthelpers import create_action, toggle_actions
+from spyder.plugins.configdialog import PluginConfigPage  # Plugin compatibility
 
 
 class TabFilter(QObject):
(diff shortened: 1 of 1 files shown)


Nodd:
```
@pytest.mark.skipif(os.name == 'nt', reason='gets stuck on Windows') # FIXME
    def test_run_python_script_in_terminal(tmpdir):
```
Well now it fails on Linux too... And it's not my fault !

spyder/plugins/__init__.py
```diff
@@ -37,6 +37,7 @@
 from spyder.py3compat import configparser, is_text_string
 from spyder.utils import icon_manager as ima
 from spyder.utils.qthelpers import create_action, toggle_actions
+from spyder.plugins.configdialog import PluginConfigPage  # Plugin compatibility
```
ccordoba12: This causes a circular import, that's why I removed it :-)

ccordoba12: I mean, the thing is when you do

    from spyder.plugins import configdialog

that causes `configdialog` to import itself again from here.

jitseniesen:
> Well now it fails on Linux too... And it's not my fault !

I had some problems with that test using Python2 on my Linux box a few days ago. The change in e9e7cff fixed it for me. Might be worth a try.

spyder/plugins/__init__.py
Nodd: I didn't spot this, but it doesn't look like it's a problem. It's pretty common to import submodules in `__init__.py`.  See for example this discussion: https://groups.google.com/forum/#!topic/comp.lang.python/HYChxtsrhnw. In that case it's fine.

Now if you want me to remove it because it keeps the code clean or whatever, I can do it !

ccordoba12: Well, yes, this also makes the code cleaner because now all Config classes are placed in `spyder.plugins.configdialog`. So please remove it because of that :-)

ccordoba12:
@Nodd, could you finish this one to merge it?

jitseniesen:
> I had some problems with that test using Python2 on my Linux box a few days ago. The change in e9e7cff fixed it for me. Might be worth a try.

This fix is now in master as 48a0ad2c2c3440fe6c3f1801de5bea1b38f272a9, so if you rebase you should get it.

Nodd:
@ccordoba12 It should be good once the tests have finished running

ccordoba12:
Sweet! Thanks a lot :-)

PR #3423, opened 2016-09-15
Try to fix plotting on Windows for the Python Console

ccordoba12:
Plotting was broken for the Python console and this fixes it in my local tests. It could also be a fix for issue #1831 (at least I don't get more hangups after pressing twice F5 with plots in a file).

@divenex, could you test this pull request and verify if that's true or not for you? Thanks ;-)


ccordoba12:
I need to release beta7 today. @divenex, please let me know if this fixes #1831 after I release it.

PR #3431, opened 2016-09-17
PR: Use correct version for qtpy

goanpeca:
Use `__version__`


ccordoba12:
Please change `PYQT_VERSION` here too (else Spyder fails to start!!)

https://github.com/spyder-ide/spyder/blob/master/spyder/requirements.py#L41

goanpeca:
Yes will do, :-p

ccordoba12:
@goanpeca, on a second check, we're correctly using `qtpy.PYQT_VERSION` in Spyder, so we don't need these changes :-)

In the places that `PYQT_VERSION` appears, we need to check for the PyQt version, not for the qtpy version, which is what `qtpy.__version__` now gives :-)

PR #3439, opened 2016-09-20
Fix profiler

Qlogin:
Fix using `add_pathlist_to_PYTHONPATH` in _profilergui.py_
This function was moved to utils/misc in commit 252c9adc49


ccordoba12:
Thanks a lot @Qlogin!!

PR #3468, opened 2016-09-27
Move plugin creation to spyder/api/plugins.py for decoupling 

### LINKED ISSUE #3448: Break plugin creation into a spyder/api module for better decoupling and organization
Move the content from

https://github.com/spyder-ide/spyder/blob/master/spyder/plugins/__init__.py

To
https://github.com/spyder-ide/spyder/blob/master/spyder/api/plugins.py

This will more easily decouple what can be extended in spyder and what cannot.

For backwards compatibility plugins using the old way will still continue to work while we make the migration

----

And yes this is just a first step of other decoupling work in progress @Nodd  ;-)

rlaverde:
Fixes #3448


goanpeca:
We should also move from 

`spyder.plugins.configdialog import PluginConfigPage` 

into 

`spyder.api.preferences`  (@ccordoba12 do you have a better name suggestion?)

Since the Preferences is something that third party users will also need to import in order to extend.

goanpeca:
@Nodd the idea here (and the api decoupling is a next step) is that third party developers should find all they need in 

`spyder.api`

This will maybe also imply moving some other widgets and some other plugins, but we are doing this stepwise.

rlaverde:
> We should also move from
`spyder.plugins.configdialog import PluginConfigPage`

Some plugin also use `spyder.plugins.configdialog GeneralConfigPage`, and `spyder.plugins.runconfig`, should them also be moved? or plugins will be encouraged to only use  `PluginConfigPage`?

ccordoba12:
> Some plugin also use spyder.plugins.configdialog GeneralConfigPage, and spyder.plugins.runconfig, should them also be moved?

I don't think so. `GeneralConfigPage` is used to create config pages that not associated to a plugin (like *Keyboard shortcuts*) and *runconfig* is only needed by our consoles.

Nodd:
`runconfig` can be used to get the python interpreter to use for code analysis or running tests for example.

ccordoba12:
> runconfig can be used to get the python interpreter to use for code analysis or running tests for example.

Mmmh, no. The Python interpreter is not tied to `runconfig`. I added a new config page called precisely *Python interpreter* in 3.0 to select the interpreter used for all consoles in Spyder (Python and IPython).

And when we add the possibility to select your conda or virtual envs to projects, that will define the interpreter used by consoles started when a project is active.

ccordoba12:
@rlaverde, failures in Travis and AppVeyor are generated because you're doing this incorrect `import`:

```python-traceback
  File "spyder/widgets/projects/configdialog.py", line 13, in <module>
    from spyder.api.preferences import ConfigDialog
ImportError: cannot import name ConfigDialog
```

rlaverde:
@ccordoba12 I changed the last commit, (In the previous commit I had changed two imports of  ConfigDialog by mistake.)

spyder/api/plugins.py
```diff
@@ -0,0 +1,593 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+
+"""
+spyder.api.plugins
+==============
```
ccordoba12: Please complete this markers to be

```
spyder.api.plugins
==================
```

spyder/api/plugins.py
```diff
@@ -0,0 +1,593 @@
...
+
+Here, 'plugins' are widgets designed specifically for Spyder
```
ccordoba12: Please change this to

    Here, 'plugins' are dock widgets ...

spyder/api/preferences.py
```diff
@@ -0,0 +1,41 @@
...
+    """Plugin configuration dialog box page widget"""
+    def __init__(self, plugin, parent):
+        self.plugin = plugin
+        self.get_option = plugin.get_option
+        self.set_option = plugin.set_option
+        self.get_font = plugin.get_plugin_font
+        self.apply_settings = plugin.apply_plugin_settings
+        SpyderConfigPage.__init__(self, parent)
+
+    def get_name(self):
```
ccordoba12: Please add a docstring here

spyder/api/preferences.py
```diff
@@ -0,0 +1,41 @@
...
+        return self.plugin.get_plugin_title()
+
+    def get_icon(self):
```
ccordoba12: Add a docstring here too

ccordoba12:
Please also add our header and a docstring to

    spyder/api/__init__.py 

describing the purpose of the `api` module :-)

ccordoba12:
This looks good to me. Just some minor comments for your @rlaverde :-)

spyder/api/plugins.py
ccordoba12: Right, I forgot about it! Then this change is not necessary @rlaverde :-)

rlaverde:
@ccordoba12 I just addressed your comments and added some docstrings, are they ok?

spyder/api/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+
+"""
+spyder.api
+==========
+
+This package contains bases clases that could be use it to build
+plugins to extend Spyder IDE.
+
+This API should be considered pre-release and is subject to change
+(final API version will be released with Spyder 4.0)
```
ccordoba12: Please change this to

> The final API version will be released in Spyder 4.0

ccordoba12:
Thanks @rlaverde, great work!

rlaverde:
You're welcome :smile:

PR #3491, opened 2016-10-02
PR: Make Pygments to work correctly with QSyntaxHighlighter

### LINKED ISSUE #2122: QSyntaxHighlighter and Pygments do not properly interoperate, leading to occasional mis-highlighting
_From [randy.he...@clockworklab.net](https://code.google.com/u/109438186339045640625/) on 2015-01-18T11:36:30Z_

As noted in my recent pull request [1], certain filetypes will not highlight well when using Pygments.  This is because the base class, QSyntaxHighlighter [2] breaks text into blocks (individual lines) and tries to highlight them one at a time [3], whereas Pygments is set up to highlight whole files at once; when Pygments is asked to highlight individual lines, it can't always give the right results (see, for example, multi-line comments in a css file like this one [4]).

This disconnect will be tough to work around.  IPython has a sophisticated attempt [5], but I don't know how well it can be adapted to Spyder.  I'll investigate and hopefully have some sort of fix I can offer, but I'm not sure I can solve it.

[1] bitbucket.org/spyder-ide/spyderlib/pull-request/97/
[2] http://qt-project.org/doc/qt-4.8/qsyntaxhighlighter.html [3] https://bitbucket.org/spyder-ide/spyderlib/src/b5a5ef2bb4d59a437a111fff7cb03faf09e86d3f/spyderlib/widgets/sourcecode/syntaxhighlighters.py?at=default#cl-856 [4] https://raw.githubusercontent.com/daneden/animate.css/master/animate.css [5] https://github.com/ipython/ipython/blob/master/IPython/qt/console/pygments_highlighter.py

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=2122_

ccordoba12:
Fixes #2122 

---

**Context**:
- The purpose of this PR is to finish @randyheydon's work started here: https://bitbucket.org/spyder-ide/spyderlib/pull-requests/103/
- Right now we use Pygments to get tokens (and then highlight them) in a line by line basis. However, that's not how Pygments really works: Pygments needs to tokenize the whole file to correctly identify comments, strings, classes, etc.
- This is why, for instance, multi-line comments in CSS files are not highlighted correctly in Spyder right now.
- So we need to pass Pygments the entire file (instead of the current line) for tokenization, _but_ in a way that doesn't introduce visible lags when working with big files (i.e. files with more than 1000 lines of code). @randyheydon already did the first part, and @goanpeca have done the second one.

TODO:
- [x] Move Pygments tokenization to a thread
- [x] Only apply it when the user stops typing


ccordoba12:
*Note to self*: This could work for task **2**:

https://wiki.qt.io/Delay_action_to_wait_for_user_interaction

ccordoba12:
Also see this for how much to wait to decide if a user has stopped typing:

http://ux.stackexchange.com/a/38545

goanpeca:
@ccordoba12 could you provide some context on the description?

ccordoba12:
@goanpeca, sorry for being so brief :-) I updated the description to be more informative.

randyheydon:
Thanks for carrying this on @ccordoba12.  I don't think I'll have a chance to help with this, but I would like to make a quick note.  The best way to fix this, I think, would be to change Pygments to support incremental lexing.  In the past, I believed they would not accept such a change, since it's outside of their scope.  However, I've since noticed that they have the same problem with the `-s` (streaming) option to the `pygmentize` script.  It currently does the same thing as Spyder, by highlighting each line individually.  Therefore, it should be possible to change Pygments to create an incremental highlighting mode that could be used by both `pygments -s` and Spyder.

I haven't looked into it too much, but I imagine this would be done with a new or modified method that takes the current lexer state as an argument, and includes the final lexer state as part of the returned data.  That way, you could parse a line based on previous state, highlight that line, then parse the next line with the new state.  Some of the individual language parsers would also need to be modified so that they split multi-line constructs into shorter constructs with intermediate states.

goanpeca:
@ccordoba12 what is the status of this? I could add the generic threaded worker from the work I will add for the `spyder-anaconda` plugin. We could then use that worker to run things like what we want to accomplish here?

jitseniesen:
@goanpeca Regarding this "generic threaded worker" you are working on, does the worker run on a different thread in the same process, or a different process? 

I'm asking because for the unittest plugin, I need a worker in a different process which can continuously communicate the results back to spyder, so that the display can be updated in real time instead of having to wait until all tests are run.

goanpeca:
@jitseniesen yes, I will have a generic worker for running python stuff (a method/func? that may be a long process) and a specific set of workers for a couple of tasks. A QProcess worker (that effectively runs a separate process and periodically signals (worker, stdout, stderr) so you can use it as well. I will add this code as generic stuff on spyder so any plugin can use this workers for long running processes.

jitseniesen:
Excellent, I'll wait for it to appear so that I can use it instead of making my own. Perhaps we can also make the introspection stuff use your generic worker.

goanpeca:
>Perhaps we can also make the introspection stuff use your generic worker.

We are planning on using https://github.com/Microsoft/language-server-protocol, which already provides a generic framework for what we need, plus there is already a python implementation plus many other languages, so we would get completion, linting for free for all those languages

spyder/utils/workers.py
```diff
@@ -0,0 +1,358 @@
...
+            output = self.func(*self.args, **self.kwargs)
+        except Exception as err:
+            error = err
+
+        if not self._is_finished:
+            self.sig_finished.emit(self, output, error)
+        self._is_finished = True
+
+        
+class ProcessWorker(QObject):
```
goanpeca: @jitseniesen this is the idea for the process worker, maybe you could take a look and see if it fits your needs.

jitseniesen: @goanpeca Yes, I think that will work for me.

ccordoba12:
@goanpeca, how is this going? We need to merge this one as soon as we can, please.

spyder/utils/syntaxhighlighters.py
```diff
@@ -1074,33 +1075,81 @@ def __init__(self, parent, font=None, color_scheme=None):
         # Load Pygments' Lexer
         if self._lang_name is not None:
             self._lexer = get_lexer_by_name(self._lang_name)
+
+            
+
```
ccordoba12: Remove two blanks here and just leave one.

spyder/utils/syntaxhighlighters.py
```diff
@@ -1074,33 +1075,81 @@ def __init__(self, parent, font=None, color_scheme=None):
...
+        # Connect document updates to a re-lexing of the document.
+        # parent.contentsChange.connect(self._make_charlist)
         BaseSH.__init__(self, parent, font, color_scheme)
+        self._worker_manager = WorkerManager()
+        self._charlist = []
+        self._allow_highlight = True
+
+    def make_charlist(self):
+        """"""
```
ccordoba12: Missing docstring.

spyder/utils/syntaxhighlighters.py
```diff
@@ -1074,33 +1075,81 @@ def __init__(self, parent, font=None, color_scheme=None):
...
+        def test_output(worker, output, error):
```
ccordoba12: Why is this function called `test_output`? It would seem to imply that it's only used for tests.

goanpeca: Fixing!

spyder/utils/syntaxhighlighters.py
goanpeca: Fixing!

spyder/widgets/sourcecode/codeeditor.py
```diff
@@ -470,6 +470,13 @@ def __init__(self, parent=None):
         # Update breakpoints if the number of lines in the file changes
         self.blockCountChanged.connect(self.update_breakpoints)
 
+        # Highlight using Pygments highlighter timer
+        self.timer_syntax_highlight = QTimer(self)
+        self.timer_syntax_highlight.setSingleShot(True)
+        self.timer_syntax_highlight.setInterval(300)
```
ccordoba12: Please add a comment to explain this `300` here.

spyder/utils/syntaxhighlighters.py
```diff
@@ -1052,6 +1052,7 @@ class PygmentsSH(BaseSH):
     # Store the language name and a ref to the lexer
     _lang_name = None
     _lexer = None
+    _charlist = []
```
ccordoba12: This doesn't seem to be needed because `_charlist` is also initialized in `__init__` below.

spyder/utils/syntaxhighlighters.py
```diff
@@ -1074,33 +1075,91 @@ def __init__(self, parent, font=None, color_scheme=None):
...
-        # Partial (parent-> child) matches
-        for key in self._tokmap:
-            if typ in key.subtypes:
-                return self._tokmap[key]
-        return 'normal'
+        # This worker runs in a thread to avoid blocking when doing full file
+        # parsing
+        self._worker_manager = WorkerManager()
+
+        # Store the format for all the tokens after pygments parsing
```
ccordoba12: `pygments` -> `Pygments`

spyder/utils/syntaxhighlighters.py
```diff
@@ -1074,33 +1075,91 @@ def __init__(self, parent, font=None, color_scheme=None):
...
+        self._charlist = []
+
+        # Flag variable to avoid unnecessary highlights if the worker has not
+        # yet finish processing
```
ccordoba12: `finish` -> `finished`

ccordoba12:
@goanpeca, do you want me to finish this one?

goanpeca:
Oops I did not see the comments

ccordoba12:
Thanks a lot for helping me to finish this one @goanpeca!! Great work!

PR #3512, opened 2016-10-07
Move edge line code out of editor to EdgeLine class

### LINKED ISSUE #1778: Add a 72-char vertical line for docstrings
_From [robro...@gmail.com](https://code.google.com/u/114002459288016453009/) on 2014-04-03T09:56:39Z_

Just an idea - it would be awesome if Spyder had an additional 72-char vertical line, to mark the maximum number of chars allowed for docstrings.

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=1778_

rlaverde:
Part of the decoupling of panels from code editor

Partially fixes #1778 missing add behavior in settings window

### DIFF
--- a/spyder/widgets/panels/edgeline.py
+++ b/spyder/widgets/panels/edgeline.py
@@ -18,11 +18,11 @@ class EdgeLine(QWidget):
 
     # --- Qt Overrides
     # -----------------------------------------------------------------
-    def __init__(self, editor, color=Qt.darkGray, columns=[79]):
+    def __init__(self, editor, colors=[Qt.darkGray], columns=[79]):
         QWidget.__init__(self, editor)
         self.editor = editor
         self.columns = columns
-        self.color = color
+        self.colors = colors
         self.setAttribute(Qt.WA_TransparentForMouseEvents)
 
         self._enabled = True
@@ -30,13 +30,14 @@ class EdgeLine(QWidget):
     def paintEvent(self, event):
         """Override Qt method"""
         painter = QPainter(self)
-        color = QColor(self.color)
-        color.setAlphaF(.5)
         rect = event.rect()
-        painter.setPen(color)
 
         offsets = [col - min(self.columns) for col in self.columns]
-        for offset in offsets:
+        for offset, qcolor in zip(offsets, self.colors):
+            color = QColor(qcolor)
+            color.setAlphaF(.5)
+            painter.setPen(color)
+
             x = rect.left() + self.editor.fontMetrics().width(offset * '9')
             painter.drawLine(x, rect.top(), x, rect.bottom())
 
--- a/spyder/widgets/panels/edgeline.py
+++ b/spyder/widgets/panels/edgeline.py
@@ -32,7 +32,13 @@ class EdgeLine(QWidget):
         painter = QPainter(self)
         color = QColor(self.color)
         color.setAlphaF(.5)
-        painter.fillRect(event.rect(), color)
+        rect = event.rect()
+        painter.setPen(color)
+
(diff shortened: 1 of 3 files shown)


spyder/widgets/panels/edgeline.py
```diff
@@ -0,0 +1,47 @@
...
+        self.code_editor = editor
+        self.column = 79
+        self.setAttribute(Qt.WA_TransparentForMouseEvents)
+
+        self._enabled = True
+
+    def paintEvent(self, event):
+        """Override Qt method"""
+        painter = QPainter(self)
+        color = QColor(Qt.darkGray)
```
rlaverde: yes, It could be an optional parameter

spyder/widgets/panels/edgeline.py
```diff
@@ -0,0 +1,47 @@
...
+
+class EdgeLine(QWidget):
+    """Source code editor's edge line (default: 79 columns, PEP8)"""
+
+    # --- Qt Overrides
+    # -----------------------------------------------------------------
+    def __init__(self, editor):
+        QWidget.__init__(self, editor)
+        self.code_editor = editor
+        self.column = 79
```
rlaverde: hmm yes, PaintEvent could draw the several lines. I will try to implement this

goanpeca: Yes please, lets do this, even if we dont we expose the functionality yet.

Basically what you can do is turn the column into a list, but for the moment we only take the first item, so that spyder continues to work just fine ;-)

spyder/widgets/panels/edgeline.py
rlaverde: It should be a function `set_color()` or a property?:
```
@property.setter
def colors(self):
    [...]
```

spyder/widgets/panels/edgeline.py
rlaverde: should I add the behavior in settings window, or It will be other PR?

goanpeca: other pr just nake sure it work in code first

spyder/widgets/panels/edgeline.py
goanpeca: Well here is where qt and python differ.... I guess a fubction to be consistent with the codebase

rlaverde: But thinking about the API, I think will be more consistent to use properties (because is the pythonic way)

rlaverde:
@goanpeca I added support to multiples edgelines (it even support different colors) but I have an error with the cursor in the area covered by the edgeline widget, the cursor draw a gray line behind it. (I need some help I don't know what is causing this little error)

edge_line

goanpeca:
@rlaverde is this happenning when more that one edgeline is drawn? or ?

Some context :-p please

rlaverde:
yes, when only one edgeline is drawn it wont happen (widget width will be 1)

and when multiples lines are drawn It wont happen after the last line, or before the first line (only between them)

rlaverde:
@goanpeca I just fixed the issue of the little line drawn behind the cursor, It was caused because I was using `event.rect()` for painting calculations, then when the cursor was moved It triggered a little repainted of the region around it, and painted a little part of the first edgeline.

goanpeca:
@rlaverde nice :-), LGTM

goanpeca:
Ready to merge?

rlaverde:
Yes, It's ready, add options for multiple lines in settings window will be other PR.

goanpeca:
Awesome, lets wait for tests to run, could you create a new issue to deal with that (adding multiple lines on configuration?)

And refer to this PR and the issue?

PR #3543, opened 2016-10-15
PR: Change of the label in the dependencies dialog

### LINKED ISSUE #3522: Dependencies-Dialog not updating correctly when installing dependencies while spyder is running
## Description
This is a minor bug I noticed while installing dependencies with spyder running.
Spyder recognises newly installed dependencies, but will not insert the installed version number in the dependency info (or this bug report as can be seen by the additional information/dependencies below: pandas version number is displayed as None, but it is labeled as (OK). ) My guess would be that the version number is only requested at startup while the label (OK/NOK) as well as the color for each dependency is checked when opening the dialog?

1. Open spyder with missing optional dependencies 
2. Check dependencies-info using Help->Dependencies... 
3. Install any missing dependency using pip (in the same environment)
4. Check dependencies-info again (close and reopen if still open)

The expected output would be the version of the installed dependency. Instead we see that the dependency is not red anymore and is labeled as (OK) but not version number is displayed.

Restarting spyder fixes this problem, as the version number is correctly displayed then.

Uninstalling dependencies while spyder is running causes a similar (maybe even more serious?) bug, as in the fact that spyder does not seem to realize that the dependency is missing at all (still shows the old version number and labels it as (OK)). I just do not think this is so serious as people usually know when they uninstalled something.

* Spyder Version: 3.0.0 
* Python Version: 2.7.12

## Dependencies
```
pandas >=0.13.1  :  None (OK)
rope >=0.9.4     :  0.10.3 (OK)

```

dalthviz:
Fixes #3522

### DIFF
--- a/spyder/widgets/dependencies.py
+++ b/spyder/widgets/dependencies.py
@@ -145,7 +145,9 @@ class DependenciesDialog(QDialog):
                               "them.<br><br>"
                               "<b>Note</b>: You can safely use Spyder "
                               "without the following modules installed: "
-                              "<b>%s</b> and <b>%s</b>")
+                              "<b>%s</b> and <b>%s</b>. Also note that new "
+                              "dependencies or changed ones will be correctly "
+                              "detected only after Spyder is restarted.")
                               % (', '.join(opt_mods[:-1]), opt_mods[-1]))
         self.label.setWordWrap(True)
         self.label.setAlignment(Qt.AlignJustify)
--- a/spyder/widgets/dependencies.py
+++ b/spyder/widgets/dependencies.py
@@ -145,7 +145,8 @@ class DependenciesDialog(QDialog):
                               "them.<br><br>"
                               "<b>Note</b>: You can safely use Spyder "
                               "without the following modules installed: "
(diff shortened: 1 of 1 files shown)


dalthviz:
@ccordoba12  this is ready for review :)

spyder/widgets/dependencies.py
```diff
@@ -145,7 +145,9 @@ def __init__(self, parent):
                               "them.<br><br>"
                               "<b>Note</b>: You can safely use Spyder "
                               "without the following modules installed: "
-                              "<b>%s</b> and <b>%s</b>")
+                              "<b>%s</b> and <b>%s</b>. Also note that new "
```
ccordoba12: Great! I like a lot the wording you decided to use. Just a couple of comments:

1. Add a `Please` at the beginning of the note:

        Please also note that new ....

2. Add this note in a new line by adding a `<br><br>` like this:

        "<b>%s</b> and <b>%s</b>.<br><br>"
        "Please also note that new ...."

That's all :-)

ccordoba12:
Thanks @dalthviz!

dalthviz:
You’re welcome @ccordoba12 👍

PR #3570, opened 2016-10-20
Default file extension for Open file dialog should be .py

### LINKED ISSUE #3436: Default file extension should be .py instead of empty
## The default file type is empty in the save dialogs

1. Open any document and use save as dialog
2. if there exists any remove the extension
3. Click save

The expected outcome (I think) the file should be save with a defauly .py extension instead of no-extension state. 

You can see from the screenshot possible extension recommendation windows is also empty

zzzz

* Spyder Version: 3.0.0b7
* Python Version: 3.6
* Operating system: Win10

## Dependencies

malliwi88:
Fixes #3436

### DIFF
--- a/spyder/plugins/editor.py
+++ b/spyder/plugins/editor.py
@@ -19,7 +19,8 @@ import time
 
 # Third party imports
 from qtpy import API, PYQT5
-from qtpy.compat import from_qvariant, getopenfilenames, to_qvariant
+from qtpy.compat import (from_qvariant, getsavefilename, getopenfilenames,
+                         to_qvariant)
 from qtpy.QtCore import QByteArray, Qt, Signal, Slot
 from qtpy.QtGui import QKeySequence
 from qtpy.QtPrintSupport import QAbstractPrintDialog, QPrintDialog, QPrinter
@@ -1887,14 +1888,9 @@ class Editor(SpyderPluginWidget):
         return editorstack.save(index=index, force=force)
     
     @Slot()
-    def save_as(self):
-        """Save *as* the currently edited file"""
-        editorstack = self.get_current_editorstack()
-        if editorstack.save_as():
-            fname = editorstack.get_current_filename()
-            if CONF.get('workingdir', 'editor/save/auto_set_to_basedir'):
-                self.open_dir.emit(osp.dirname(fname))
-            self.__add_recent_file(fname)
+    def save_as(self, index=None):
+        """Save file as..."""
+        self.get_current_editorstack().save_as(index=index)
     
     @Slot()
     def save_all(self):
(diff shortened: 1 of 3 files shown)


spyder/widgets/editor.py
```diff
@@ -25,13 +25,15 @@
 from qtpy.QtGui import QFont, QKeySequence
 from qtpy.QtWidgets import (QAction, QApplication, QHBoxLayout, QMainWindow,
                             QMessageBox, QMenu, QSplitter, QVBoxLayout,
-                            QWidget)
+                            QWidget, QFileDialog)
```
goanpeca: Why did you add QFileDialog? it seems to not be used

goanpeca: Oops my bad, I see where you used it

spyder/widgets/editor.py
```diff
@@ -1309,11 +1311,24 @@ def select_savename(self, original_filename):
...
             # Save the currently edited file
+            if not self.get_stack_count():
+                return
             index = self.get_stack_index()
+
         finfo = self.data[index]
-        filename = self.select_savename(finfo.filename)
+
+        self.redirect_stdio.emit(False)
+        selectedfilter = ''
```
goanpeca: Why is selected filter an empty string?

spyder/plugins/editor.py
```diff
@@ -1887,14 +1888,9 @@ def save(self, index=None, force=False):
...
     
     @Slot()
-    def save_as(self):
-        """Save *as* the currently edited file"""
-        editorstack = self.get_current_editorstack()
-        if editorstack.save_as():
-            fname = editorstack.get_current_filename()
-            if CONF.get('workingdir', 'editor/save/auto_set_to_basedir'):
-                self.open_dir.emit(osp.dirname(fname))
-            self.__add_recent_file(fname)
```
goanpeca: Why did you remove all this logic?

spyder/widgets/editor.py
```diff
@@ -1325,10 +1340,6 @@ def save_as(self, index=None):
 
             new_index = self.rename_in_data(index, new_filename=filename)
 
-            # We pass self object ID as a QString, because otherwise it would
-            # depend on the platform: long for 64bit, int for 32bit. Replacing
-            # by long all the time is not working on some 32bit platforms
-            # (see Issue 1094, Issue 1098)
```
goanpeca: Why did you remove that comment explaining why the changes are needed ?

malliwi88: Hi @goanpeca these comments do not belong to my amendment, so I preferred to remove trash.

goanpeca: >Hi @goanpeca these comments do not belong to my amendment, so I preferred to remove trash.

If you did not made those comments, and the comments do not refer to code you are modifying, how is that trash??? please undo those changes. You should never remove comments of work that is not related to your work

spyder/plugins/editor.py
malliwi88: @goanpeca I remove it because integrate all logic on Widget module, And only call save_as functionality from Widget module with this sentence... on line 1893 (plugins/editor.py)
> `self.get_current_editorstack().save_as(index=index)`

PR #3574, opened 2016-10-21
PR: Improve style of scientific startup script

adler-j:
- Moved the `__has_...` so it is not defined in two different locations
- Import all packages uniformly (no special treatment of guiqwt)
- Use new style formatting.
- Give correct "Imported ..." message even if numpy is not installed.


ccordoba12:
Thanks a lot for your contribution. We'll include it in Spyder 3.1 :-)

ccordoba12:
This looks good to me. Thanks for your contribution!

PR #3577, opened 2016-10-21
PR: Use inspect.getfullargspec() in getdoc for PY3 to support annotated functions.

### LINKED ISSUE #3571: Help fails to produce Rich Text on annotated functions
# Description of your problem
## What steps will reproduce the problem?
1. Create a function and use function annotations, à la [PEP-3107](https://www.python.org/dev/peps/pep-3107/) and save the module (in this case to `doc_check.py`). For example:

``` python
def test_function(value: int) -> int:
    """This is a very boring function

    This function just returns whatever you put into it.

    Parameters
    ----------
    value
        This is an integer

    Returns
    -------
    test_function
        This is the same as whatever we put in the function
    """
    return value
```
1. Import the module in the command window
2. Type the module.function name and press Ctrl+I. 
### Expected Output

I expect to see Rich Text help in the Help pane, similar to that generated from a sphinx-quickstart html output:
test_function_html_output

(It should be noted that in general section headers are handled well in the Help pane--I'm not sure why they're not showing up in my basic Sphinx setup. Napoleon handles them well:
test_function_html_output_napoleon
### What I see

In the Spyder Help pane I just get **No documentation available**
- Spyder Version: 3.0.1
- Python Version: 3.4.5
- Operating system: Windows 7 Enterprise 64-bit
## Dependencies

and paste the contents below:
sympy >=0.7.3    :  1.0 (OK)

silentquasar:
inspect.getargspec() doesn't get the function annotations, which causes Help to fail on annotated functions in Python 3. getfullargspec does have the capability, but it doesn't exist in Python 2. Suggested fix provided here.

Fixes #3571.

### DIFF
--- a/spyder/utils/dochelpers.py
+++ b/spyder/utils/dochelpers.py
@@ -15,7 +15,7 @@ import re
 from spyder.utils import encoding
 from spyder.py3compat import (is_text_string, builtins, get_meth_func,
                               get_meth_class_inst, get_meth_class,
-                              get_func_defaults, to_text_string)
+                              get_func_defaults, to_text_string, PY2)
 
 
 SYMBOLS = r"[^\'\"a-zA-Z0-9_.]"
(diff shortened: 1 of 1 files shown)


ccordoba12:
@silentquasar, nice!! Thanks a lot for your contribution :-)

We'll include it in Spyder 3.1.

ccordoba12:
It looks good to me. Thanks!

PR #3578, opened 2016-10-21
PR: Move language funcs into a subfolder

goanpeca:
All the files on the repo root create a bit of a mess...

### DIFF
--- a/language_scripts/gettext_compile.py
+++ b/language_scripts/gettext_compile.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+
+# Standard library imports
+import os
+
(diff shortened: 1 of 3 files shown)


malliwi88:
Nice!! :) more order on the repo.

ccordoba12:
@goanpeca, +1 for this, but I have some comments:

1. Did you move these language scripts or just added them as new files? It would seem you added them because the GH doesn't report the move :-)
2. Did you test that these scripts are still working after the move? I hadn't moved them before because they assume the Spyder modules are in the same directory as the one they are located in.
3. You did this PR against master, so it can't be marked for 3.1.

goanpeca:
@ccordoba12, hmmmm true, I will move them and then make the changes.

Actually on OSX these scripts dont work :-|, (but because some files are not found) I will check on linux or win ;-)

ccordoba12:
Closing because these files are going to be moved to a new repo.

PR #3580, opened 2016-10-21
PR: Fix issue #3436 

### LINKED ISSUE #3436: Default file extension should be .py instead of empty

malliwi88:
Fixes #3436

Default file extension should be .py (Save_as Widget).

### DIFF
--- a/spyder/widgets/editor.py
+++ b/spyder/widgets/editor.py
@@ -32,6 +32,8 @@ from spyder.config.base import _, DEBUG, STDERR, STDOUT
 from spyder.config.gui import (config_shortcut, fixed_shortcut,
                                RUN_CELL_SHORTCUT,
                                RUN_CELL_AND_ADVANCE_SHORTCUT)
+from spyder.config.utils import (get_edit_filetypes, get_edit_filters,
+                                 get_filter)
 from spyder.py3compat import qbytearray_to_str, to_text_string, u
 from spyder.utils import icon_manager as ima
 from spyder.utils import (codeanalysis, encoding, sourcecode,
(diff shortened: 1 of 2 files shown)


malliwi88:
@goanpeca I have a Read-only Zenhub Board, please. You can change the Milestone for this PR

goanpeca:
@malliwi88 great :-), this is exactly the fix I had in mind ;-)

spyder/widgets/editor.py
```diff
@@ -1302,7 +1304,11 @@ def file_saved_in_other_editorstack(self, index, filename):
     def select_savename(self, original_filename):
         self.redirect_stdio.emit(False)
         filename, _selfilter = getsavefilename(self, _("Save file"),
-                                               original_filename)
+                                               original_filename,
+                                               get_edit_filters(),
+                                               get_filter(get_edit_filetypes(),
+                                                          osp.splitext(original_filename)[1])
+                                               )
```
goanpeca: what happens if the file does not have an extension?

Could you test if you open a file without extension on the editor?

for example this one https://github.com/spyder-ide/spyder/blob/master/LICENSE

malliwi88: Hi @goanpeca these are my results just look like that on:

* Win 7 
fix-save_as win 7 seleccion_006

* Linux
fix-save_as seleccion_006

goanpeca:
Ok LGTM :-)

spyder/widgets/editor.py
```diff
@@ -1302,7 +1304,11 @@ def file_saved_in_other_editorstack(self, index, filename):
     def select_savename(self, original_filename):
         self.redirect_stdio.emit(False)
         filename, _selfilter = getsavefilename(self, _("Save file"),
-                                               original_filename)
+                                               original_filename,
+                                               get_edit_filters(),
+                                               get_filter(get_edit_filetypes(),
```
ccordoba12: Perhaps if we change here `get_edit_filetypes` to `EDIT_FILETYPES` we could get the expected result :-)

PR #3581, opened 2016-10-22
PR: remove spyder/widgets/tests/__init__.py because of error in execution of pytest -rxs

malliwi88:
@ccordoba12 @goanpeca 

According to [Good Practices for Pytest](http://docs.pytest.org/en/latest/goodpractices.html#choosing-a-test-layout-import-rules). The structure for **test layout / import rules** on `spyder/spyder/widgets/tests/` doesn't need `__init__.py` file.

If we don't need to keep the test names. It solves errors messages over execution of `pytest -rxs`

* with __init__.py
seleccion_001

* without __init__.py
seleccion_002

Additionally, `pytest -rxs` let us see specific location of *XFAIL error* over test.  It facilitates the detection of faults in problems such as *auto_indent* those reported @rlaverde.


malliwi88:
As I can see in [Spyder Dev Testing](https://github.com/spyder-ide/spyder/wiki/Dev:-Testing)

> When using pytest directly, the command is basically py.test (there are some gotchas, explained below). This runs all tests in .....
> If you use pytest directly, you need to make sure that the root of the Spyder source tree is included in the Python path, because the test files import spyder and its modules. You can use the PYTHONPATH environment variable for this. **If you don't do this, then the tests will fail, or (if you have Spyder installed) they will test the installed version of Spyder.**

Instead [Pytest Doc] () say this about of layout/test functionality
> avoid “__init__.py” files in your test directories. This way your tests can run easily against an installed version of mypkg, independently from the installed package if it contains the tests or not.

So avoiding conflict over pytest execution on development such as picture above or this reported problem on [StackOverflow Question](
http://stackoverflow.com/questions/12582503/py-test-test-discovery-failure-when-tests-in-different-directories-are-called)

jitseniesen:
This is a good idea (actually, I thought I had already done this myself!). The pytest docs are a bit ambivalent, because in the next paragraph they say:

> With inlined tests [which we have] you might put `__init__.py` into test directories and make them installable as part of your application.

But I think it's indeed better to remove `__init__.py`. In any case, we should be consistent, and there is only one test directory with `__init__.py`.

@malliwi88 Feel free to edit  [Spyder Dev Testing](https://github.com/spyder-ide/spyder/wiki/Dev:-Testing). It's just something I wrote without too much thought.

ccordoba12:
Thanks @malliwi88!! Merging now :-)

Issue #3592, opened 2016-10-24
Add multiple edgelines options in settings window


rlaverde:
After #3512 edgeline widget support several lines with different colors. But this functionality isn't exposed through settings window.

How this should be done?
- Settings `'edge_line': True, 'edge_line_column': 79` changed by a list `'edge_line': [72,79]`?
- another option for colors `'edge_line_colors': []`? or this colors will be part of the syntax color scheme?
- How It'll look like in the settings page?

goanpeca:
@ccordoba12 what do you suggest?

We could do 'edge_line': True, 'edge_line_columns': [79, 89]

So we change edge_line_column to edge_line_columns and avoid some problems moving forward?

ccordoba12:
> Settings 'edge_line': True, 'edge_line_column': 79 changed by a list 'edge_line': [72,79]?

I agree with 'edge_line_columns: [79]` in `config/main.py`. Having `72` for docstrings is nice, but it shouldn't be the default because I think most people (especially newbies) don't know about that convention.

Besides, we should adjust the pep8 config to follow this setting too.

> another option for colors

Let's not add different colors, that can be confusing.

> How It'll look like in the settings page?

It could be a combo box with a list of columns separated by commas, like this option

    IPython Console > Startup > Run Code

goanpeca:
@ccordoba12 
>It could be a combo box with a list of columns separated by commas, like this option

You meant textbox?

ccordoba12:
Yep, sorry :-)

goanpeca:
@rlaverde you can work on this then :-p

PR #3607, opened 2016-10-27
PR: Multiple edge lines preferences

### LINKED ISSUE #1778: Add a 72-char vertical line for docstrings

rlaverde:
Fixes #1778 and #3592

### DIFF
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -211,7 +211,7 @@ DEFAULTS = [
               'line_numbers': True,
               'blank_spaces': False,
               'edge_line': True,
-              'edge_line_column': 79,
+              'edge_line_columns': '79',
               'toolbox_panel': True,
               'calltips': True,
               'go_to_definition': True,
@@ -652,7 +652,7 @@ DEFAULTS = [
 #    or if you want to *rename* options, then you need to do a MAJOR update in
 #    version, e.g. from 3.0.0 to 4.0.0
 # 3. You don't need to touch this value if you're just adding a new option
-CONF_VERSION = '31.0.0'
+CONF_VERSION = '32.0.0'
 
 # Main configuration instance
 try:
--- a/spyder/plugins/configdialog.py
+++ b/spyder/plugins/configdialog.py
@@ -15,8 +15,8 @@ import os.path as osp
 from qtpy import API
 from qtpy.compat import (getexistingdirectory, getopenfilename, from_qvariant,
                          to_qvariant)
-from qtpy.QtCore import QSize, Qt, Signal, Slot
-from qtpy.QtGui import QColor
+from qtpy.QtCore import QSize, Qt, Signal, Slot, QRegExp
+from qtpy.QtGui import QColor, QRegExpValidator
 from qtpy.QtWidgets import (QButtonGroup, QCheckBox, QComboBox, QDialog,
                             QDialogButtonBox, QDoubleSpinBox, QFontComboBox,
                             QGridLayout, QGroupBox, QHBoxLayout, QLabel,
@@ -465,7 +465,7 @@ class SpyderConfigPage(ConfigPage, ConfigAccessMixin):
         return radiobutton
     
     def create_lineedit(self, text, option, default=NoDefault,
-                        tip=None, alignment=Qt.Vertical):
+                        tip=None, alignment=Qt.Vertical, regex=None):
         label = QLabel(text)
         label.setWordWrap(True)
         edit = QLineEdit()
@@ -475,6 +475,8 @@ class SpyderConfigPage(ConfigPage, ConfigAccessMixin):
         layout.setContentsMargins(0, 0, 0, 0)
         if tip:
             edit.setToolTip(tip)
+        if regex:
+            edit.setValidator(QRegExpValidator(QRegExp(regex)))
         self.lineedits[edit] = (option, default)
         widget = QWidget(self)
         widget.label = label
--- a/spyder/widgets/editor.py
+++ b/spyder/widgets/editor.py
@@ -375,7 +375,7 @@ class EditorStack(QWidget):
         self.linenumbers_enabled = True
         self.blanks_enabled = False
         self.edgeline_enabled = True
-        self.edgeline_columns = [79]
+        self.edgeline_columns = (79,)
         self.codecompletion_auto_enabled = True
         self.codecompletion_case_enabled = False
         self.codecompletion_enter_enabled = False
--- a/spyder/widgets/editor.py
+++ b/spyder/widgets/editor.py
@@ -375,7 +375,7 @@ class EditorStack(QWidget):
         self.linenumbers_enabled = True
         self.blanks_enabled = False
         self.edgeline_enabled = True
-        self.edgeline_column = 79
+        self.edgeline_columns = [79]
         self.codecompletion_auto_enabled = True
         self.codecompletion_case_enabled = False
         self.codecompletion_enter_enabled = False
@@ -755,12 +755,12 @@ class EditorStack(QWidget):
             for finfo in self.data:
                 finfo.editor.edge_line.set_enabled(state)
 
-    def set_edgeline_column(self, column):
+    def set_edgeline_columns(self, columns):
         # CONF.get(self.CONF_SECTION, 'edge_line_column')
-        self.edgeline_column = column
+        self.edgeline_columns = columns
         if self.data:
             for finfo in self.data:
-                finfo.editor.edge_line.set_column(column)
+                finfo.editor.edge_line.set_columns(columns)
 
     def set_codecompletion_auto_enabled(self, state):
         # CONF.get(self.CONF_SECTION, 'codecompletion_auto')
@@ -1706,7 +1706,7 @@ class EditorStack(QWidget):
                 linenumbers=self.linenumbers_enabled,
                 show_blanks=self.blanks_enabled,
                 edge_line=self.edgeline_enabled,
-                edge_line_column=self.edgeline_column, language=language,
+                edge_line_columns=self.edgeline_columns, language=language,
                 markers=self.has_markers(), font=self.default_font,
                 color_scheme=self.color_scheme,
                 wrap=self.wrap_enabled, tab_mode=self.tabmode_enabled,
--- a/spyder/widgets/panels/edgeline.py
+++ b/spyder/widgets/panels/edgeline.py
@@ -33,7 +33,7 @@ class EdgeLine(QWidget):
         size = self.size()
 
         offsets = [col - min(self.columns) for col in self.columns]
-        for offset, qcolor in zip(offsets, self.colors):
+        for offset, qcolor in zip(offsets, cycle(self.colors)):
             color = QColor(qcolor)
             color.setAlphaF(.5)
             painter.setPen(color)
@@ -49,9 +49,13 @@ class EdgeLine(QWidget):
         self._enabled = state
         self.setVisible(state)
 
-    def set_column(self, column, index=0):
-        """Set edge line column value"""
-        self.columns[index] = column
+    def set_columns(self, columns):
+        """Set edge line columns values."""
+        if isinstance(columns, list):
+            self.columns = columns
+        elif isinstance(columns, str):
+            self.columns = [int(e) for e in columns.split(',')]
+
         self.update()
 
     def set_geometry(self, cr):
--- a/spyder/widgets/panels/edgeline.py
+++ b/spyder/widgets/panels/edgeline.py
@@ -5,7 +5,7 @@
 # (see spyder/__init__.py for details)
 
 """
-This module contains the edge line numebr panel
+This module contains the edge line panel
 """
 
 from qtpy.QtWidgets import QWidget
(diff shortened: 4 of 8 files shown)


spyder/config/main.py
```diff
@@ -206,6 +206,7 @@
               'blank_spaces': False,
               'edge_line': True,
               'edge_line_column': 79,
```
goanpeca: Maybe we need to add a comment here to say we are no longer using this, or remove the config and bump the version number of the config as it is described on the bottom of this file.

@ccordoba12 ?

goanpeca:
@rlaverde could you paste a screenshot of how the preferences look and how the editor looks?

rlaverde:
@goanpeca 
edge_line2

At the moment all the lines are gray, I'm planning to use syntax colors.

goanpeca:
@rlaverde we should also add this setting to the project configuration, so that projects can define that setting per project

ccordoba12:
@rlaverde, what do you mean by

> At the moment all the lines are gray, I'm planning to use syntax colors.

?

I'd prefer all lines to be gray :-)

goanpeca:
@ccordoba12 he means that instead of hardcoding a color, he is using a color from the syntax color scheme

ccordoba12:
Ok, that's nice :-) But what color?

goanpeca:
https://github.com/spyder-ide/spyder/pull/3607/files#diff-7c6fb7800dd1ba0dce14f4a15c8d05a1R58

But actually... @rlaverde I think that only makes sense for python.... so we should use one and the same color for all the lines.

ccordoba12:
> so we should use one and the same color for all the lines.

Yep, I agree.

rlaverde:
> so we should use one and the same color for all the lines.
@goanpeca comments color?

goanpeca:
@rlaverde I guess that makes sense

rlaverde:
I'm also going to remove the logic to use different colors for each line, or should I leave it?

goanpeca:
>I'm also going to remove the logic to use different colors for each line, or should I leave it?

Please remove

spyder/plugins/configdialog.py
```diff
@@ -398,7 +401,13 @@ def save_to_conf(self):
         for radiobutton, (option, _default) in list(self.radiobuttons.items()):
             self.set_option(option, radiobutton.isChecked())
         for lineedit, (option, _default) in list(self.lineedits.items()):
-            self.set_option(option, to_text_string(lineedit.text()))
+            option_value = lineedit.text()
+            old_value = self.get_option(option, _default)
+            if isinstance(old_value, list):
+                option_value = [int(e) for e in option_value.split(',')]
```
ccordoba12: I don't know if this is a good decision. What would happen here if another option is a list of strings?

ccordoba12: To address this I think we we should change it to

```python
try:
    option_value = [int(e) for e in option_value.split(',')]
except ValueError:
    option_value = [e for e in option_value.split(',')]
```

@goanpeca, what do you think?

goanpeca: Why do we need to turn it into a number?

ccordoba12: `set_option` saves options as Python objects.

goanpeca: Yes but why do we need to make a list of ints, and not use a list of strings?

ccordoba12:
Right, good question :-)

spyder/plugins/configdialog.py
goanpeca: @rlaverde, since this is a textbox, and we cannot safely assume the type of content we need to either NOT make it a list of ints, and handle that on the widget (check if int if not turn into an int), otherwise we will need to define a subtype of textbox that is explicit about the type... or add some extra options so we can safely define if it is a list and what type of items

https://github.com/spyder-ide/spyder/blob/master/spyder/plugins/configdialog.py#L467

@ccordoba12 ?

goanpeca: I like the sound of 2 in the medium term, but for this PR probably its easier and faster to make 1?

goanpeca: I would say its not urgent work, so I would take the simplest solution possible

rlaverde:
@goanpeca I added a new commit moving the logic to the `set_colums` method, I really struggle making this work, because this option will accept different values that could be interpreted as different python types (list, int, str).

what do you think? I'm not sure if this is the way to solve this.

spyder/widgets/panels/edgeline.py
```diff
@@ -49,11 +48,26 @@ def set_enabled(self, state):
...
+        if isinstance(columns, list):
+            self.columns = columns
+        elif isinstance(columns, int):
+            self.columns = [columns]
+        elif isinstance(columns, str):
+            try:
+                self.columns = [int(e) for e in columns.split(',')]
+            except:
+                #TODO Log error, invalid columns argument
+                pass
```
goanpeca: @rlaverde its fine, maybe just add a comment to say that Configuration of a text box for now does not assume anything about the content so that it is clear why this is needed, ok :-) ?

goanpeca:
@rlaverde ready to go?

rlaverde:
@goanpeca yes, I think It's ready

ccordoba12:
I'll take a last look at this one this afternoon and merge it if it's ready.

Thanks for your patience @rlaverde :-)

spyder/config/main.py
```diff
@@ -627,7 +627,7 @@
 #    or if you want to *rename* options, then you need to do a MAJOR update in
 #    version, e.g. from 3.0.0 to 4.0.0
 # 3. You don't need to touch this value if you're just adding a new option
-CONF_VERSION = '30.1.0'
+CONF_VERSION = '31.1.0'
```
ccordoba12: This needs to be `31.0.0`

spyder/plugins/configdialog.py
```diff
@@ -475,6 +480,8 @@ def create_lineedit(self, text, option, default=NoDefault,
         layout.setContentsMargins(0, 0, 0, 0)
         if tip:
             edit.setToolTip(tip)
+        if regex:
+            edit.setValidator(QRegExpValidator(QRegExp(regex)))
```
ccordoba12: Nevermind, I saw it below

spyder/plugins/configdialog.py
```diff
@@ -317,7 +317,12 @@ def load_from_conf(self):
             radiobutton.toggled.connect(lambda _foo, opt=option:
                                         self.has_been_modified(opt))
         for lineedit, (option, default) in list(self.lineedits.items()):
-            lineedit.setText(self.get_option(option, default))
+            option_value = self.get_option(option, default)
+            if isinstance(option_value, list):
+                option_value = ",".join(str(e) for e in option_value)
+            if isinstance(option_value, int):
+                option_value = str(option_value)
+            lineedit.setText(option_value)
```
ccordoba12: I sorry, I see it. It's to be able to set the option's text in our lineedit's

spyder/widgets/panels/edgeline.py
```diff
@@ -49,11 +48,23 @@ def set_enabled(self, state):
...
-    def set_column(self, column, index=0):
-        """Set edge line column value"""
-        self.columns[index] = column
+    def set_columns(self, columns):
+        """Set edge line columns values."""
+        if isinstance(columns, list):
+            self.columns = columns
+        elif isinstance(columns, int):
+            self.columns = [columns]
+        elif isinstance(columns, str):
```
ccordoba12: Do we really need this last validation?

ccordoba12:
I'm sorry but one thing that's not clear to me is what value is passed as columns to `edge_line`. Is it a list of numbers, a list of strings or a single string?

goanpeca:
>I'm sorry but one thing that's not clear to me is what value is passed as columns to edge_line. Is it a list of numbers, a list of strings or a single string?

The line edit stores either `79,99` or `77` or `79, 99, 105`

Always a string... but that string can be a single number or several numbers (that get transformed to a list)

spyder/config/main.py
```diff
@@ -203,7 +203,7 @@
               'line_numbers': True,
               'blank_spaces': False,
               'edge_line': True,
-              'edge_line_column': 79,
+              'edge_line_columns': [79],
```
goanpeca: @rlaverde i think we should not store the settings as a list but only as a string.

`'79'` or `'79,99'`

This way, the regex is used when defining the line edit, but we dont need to make check on the configuration side. We should only make these transformatios (check if it has commas... split by commas etc...) on the set_columns method

I think this way is clearer @ccordoba12 ?

ccordoba12: Yes, it is. Thanks!

spyder/widgets/panels/edgeline.py
rlaverde: hmm no, It was necessary when there wasn't the regex validation, I think this validation could be deleted and use just `else`

rlaverde:
> Always a string... but that string can be a single number or several numbers (that get transformed into a list)

Now they get transformed into a int, or a tuple
If the default value is a `list`they get trnasformed into a list (that's the reason I had used a list)

rlaverde:
@ccordoba12 I addressed your comments, Should I rebase it and squash some commits?

spyder/plugins/configdialog.py
```diff
@@ -318,7 +318,7 @@ def load_from_conf(self):
                                         self.has_been_modified(opt))
         for lineedit, (option, default) in list(self.lineedits.items()):
             option_value = self.get_option(option, default)
-            if isinstance(option_value, list):
+            if isinstance(option_value, tuple):
```
goanpeca: This should not be here. We dont want to change the logic of the line edit

goanpeca: Ahh ok :-)

spyder/plugins/configdialog.py
```diff
@@ -317,7 +317,12 @@ def load_from_conf(self):
             radiobutton.toggled.connect(lambda _foo, opt=option:
                                         self.has_been_modified(opt))
         for lineedit, (option, default) in list(self.lineedits.items()):
-            lineedit.setText(self.get_option(option, default))
+            option_value = self.get_option(option, default)
+            if isinstance(option_value, tuple):
+                option_value = ",".join(str(e) for e in option_value)
```
ccordoba12: `str` returns bytes for Python 2 and Unicode for Python 3. We should use only format here.

So I think we should change this call to `str` (and the one below) to `to_text_string`

spyder/plugins/configdialog.py
```diff
@@ -317,7 +317,12 @@ def load_from_conf(self):
             radiobutton.toggled.connect(lambda _foo, opt=option:
                                         self.has_been_modified(opt))
         for lineedit, (option, default) in list(self.lineedits.items()):
-            lineedit.setText(self.get_option(option, default))
+            option_value = self.get_option(option, default)
+            if isinstance(option_value, tuple):
```
ccordoba12: In the `set_columns` method of `EdgeLine` you're transforming all values saved to `edge_line_columns` to lists, but here you're checking only for tuples.

So which one is right? Tuples or lists? :-)

rlaverde: `list` for `EdgeLine` class and `tuple` for configuration

I think Its better to use list for `columns` attribute of `EdgeLine` class; and when loading data from configuration, `get_option` will return a tuple when reading comma separated string (e.g. `72,80`)

ccordoba12: Why not use lists or tuples for both? I find this very confusing, so I really want you to decide for one or the other :-)

ccordoba12: +1 from me too for tuples :-)

rlaverde:
I modified UserConfig.get method and now it's not necessary to add extra logic to `loadconfig`, `edgeline_columns` option will always load and saved as `str`, and the conversion to `list` will be managed by `EdgeLine` class.

rlaverde:
I rebase, reorder and squash some commits in order to make the history clearer, I think this is finally ready :see_no_evil:

spyder/widgets/editor.py
```diff
@@ -375,7 +375,7 @@ def __init__(self, parent, actions):
         self.linenumbers_enabled = True
         self.blanks_enabled = False
         self.edgeline_enabled = True
-        self.edgeline_column = 79
+        self.edgeline_columns = [79]
```
ccordoba12: Please change this to a tuple

ccordoba12:
> I modified UserConfig.get method and now it's not necessary to add extra logic to loadconfig

Great work at finding such a great and simple solution!! Kudos to you! :-)

> I think this is finally ready

I left a couple of last comments, for consistency. Then I'll merge (promised!)

ccordoba12:
Thanks @rlaverde!

goanpeca:
😄

PR #3539, opened 2016-10-14
PR: Remember format for floats in dataframe editor

jitseniesen:
This pull request is to make Spyder remember the format that the user sets for floats in the DataFrameEditor by storing it in the user configuration settings. Fixes #2728.

- A new config setting `dataframe_format` is introduced in the variable explorer section in the user config. This setting contains the format, without the leading `%` character (e.g., if the format is `'%.3f'` then the setting is stored as `'.3f'`). The reason for this is that it is apparently impossible to have a config setting with a `%` character, because this is used in `ConfigParser` for interpolation.
- This setting is passed (after adding a `%` character in the front) from the `VariableExplorer` plugin via the `NamespaceBrowser` and `CollectionEditor` widgets to the `DataFrameEditor` widget, where it is used to set the format for floats.
- If the user changes the format in the `DataFrameEditor` widget, then this information is passed back using a `sig_option_changed` signal via `CollectionEditor` and `NamespaceBrowser` to `VariableExplorer`. There the leading `%` character is stripped and the format is stored in the user config.

PR #3611, opened 2016-10-28
 PR: Remember format for floats in DataFrame editor

### LINKED ISSUE #2728: Add the default format in dataframe editor to user preferences
The default format should be defined at out of the widget as [asked here](http://stackoverflow.com/questions/32438504/spyder-dataframe-editor-format?rq=1).

jitseniesen:
Fixes #2728.

----

This is #3539 <s>raised from the dead</s> but based now against 3.x branch. 

This pull request is to make Spyder remember the format that the user sets for floats in the DataFrameEditor by storing it in the user configuration settings.
- A new config setting `dataframe_format` is introduced in the variable explorer section in the user config. This setting contains the format, without the leading `%` character (e.g., if the format is `'%.3f'` then the setting is stored as `'.3f'`). The reason for this is that it is apparently impossible to have a config setting with a `%` character, because this is used in `ConfigParser` for interpolation.
- This setting is passed (after adding a `%` character in the front) from the `VariableExplorer` plugin via the `NamespaceBrowser` and `CollectionEditor` widgets to the `DataFrameEditor` widget, where it is used to set the format for floats.
- If the user changes the format in the `DataFrameEditor` widget, then this information is passed back using a `sig_option_changed` signal via `CollectionEditor` and `NamespaceBrowser` to `VariableExplorer`. There the leading `%` character is stripped and the format is stored in the user config.

### DIFF
--- a/spyder/plugins/variableexplorer.py
+++ b/spyder/plugins/variableexplorer.py
@@ -104,13 +104,23 @@ class VariableExplorer(QWidget, SpyderPluginMixin):
     @staticmethod
     def get_settings():
         """
-        Return Variable Explorer settings dictionary
-        (i.e. namespace browser settings according to Spyder's configuration file)
+        Retrieve all Variable Explorer configuration settings
+        
+        Specifically, return the settings in CONF_SECTION with keys in 
+        REMOTE_SETTINGS, and the setting 'dataframe_format'.
+        
+        Returns:
+            dict: settings
         """
         settings = {}
 #        CONF.load_from_ini() # necessary only when called from another process
         for name in REMOTE_SETTINGS:
             settings[name] = CONF.get(VariableExplorer.CONF_SECTION, name)
+
+        # dataframe_format is stored without percent sign in config
+        # to avoid interference with ConfigParser's interpolation
+        name = 'dataframe_format'
+        settings[name] = '%' + CONF.get(VariableExplorer.CONF_SECTION, name)
         return settings
 
     # ----- Stack accesors ----------------------------------------------------
@@ -131,6 +141,12 @@ class VariableExplorer(QWidget, SpyderPluginMixin):
 
     # ----- Public API --------------------------------------------------------
     def add_shellwidget(self, shellwidget):
+        """
+        Register shell with variable explorer.
+
+        This function opens a new NamespaceBrowser for browsing the variables
+        in the shell.
+        """
         shellwidget_id = id(shellwidget)
         # Add shell only once: this method may be called two times in a row
         # by the External console plugin (dev. convenience)
--- a/spyder/plugins/variableexplorer.py
+++ b/spyder/plugins/variableexplorer.py
@@ -7,7 +7,7 @@
 """Variable Explorer Plugin"""
 
 # Third party imports
-from qtpy.QtCore import Signal
+from qtpy.QtCore import Signal, Slot
 from qtpy.QtWidgets import QGroupBox, QStackedWidget, QVBoxLayout, QWidget
 
 # Local imports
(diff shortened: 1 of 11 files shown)


jitseniesen:
I addresses @goanpeca's comments, which were mostly about formatting but he did point out one important issue: if the user gives a format that starts with a `%`, an assertion is triggered. This is handled more gracefully by the last commit (d765a74).

goanpeca:
LGTM :-) I would say marge, but we are waiting on @ccordoba12 3.02 release :-)

spyder/widgets/variableexplorer/collectionseditor.py
```diff
@@ -511,6 +514,17 @@ def create_dialog(self, editor, data):
...
+    @Slot(str, object)
+    def change_option(self, option_name, new_value):
+        """
+        Change configuration option.
+
+        This function is called when a `sig_option_changed` signal is received.
+        At the moment, this signal can only come from a DataFrameEditor.
+        """
+        assert option_name == 'dataframe_format'
+        self.parent().set_dataframe_format(new_value)
```
ccordoba12: Could we move this method to `dataframeeditor`? It has nothing to do with `collectionseditor` :-)

spyder/widgets/variableexplorer/collectionseditor.py
```diff
@@ -868,6 +882,17 @@ def toggle_minmax(self, state):
...
+    @Slot(str)
+    def set_dataframe_format(self, new_format):
+        """
+        Set format to use in DataframeEditor.
+
+        Args:
+            new_format (string): e.g. "%.3f"
+        """
+        self.sig_option_changed.emit('dataframe_format', new_format)
+        self.model.dataframe_format = new_format
```
ccordoba12: Also this method

ccordoba12:
This looks very good to me. I just added some comments to simplify the code a little bit ;-)

ccordoba12:
This also needs a rebase/merge with 3.x now :-)

jitseniesen:
Rebased but I still have to look at the comments re `collectioneditor.py`

spyder/plugins/variableexplorer.py
```diff
@@ -104,15 +104,41 @@ def __init__(self, parent):
...
+        
+        Returns:
+            dict: settings
         """
+        section = VariableExplorer.CONF_SECTION
         settings = {}
-#        CONF.load_from_ini() # necessary only when called from another process
         for name in REMOTE_SETTINGS:
-            settings[name] = CONF.get(VariableExplorer.CONF_SECTION, name)
+            settings[name] = CONF.get(section, name)
```
jitseniesen: This is a static method so there is no `self`. I will convert it to a normal method but I discovered that it is called in a number of places so I need to change them ...

ccordoba12: Ok, thanks for looking at this. I think it'd be nicer to have it as a method.

spyder/widgets/variableexplorer/collectionseditor.py
jitseniesen: The `collectionseditor` creates the `dataframeeditor` so it has to inject the relevant configuration options (here only the dataframe format) in the `dataframeeditor`.

As I understand the widget/plugin separation, widgets should get all relevant configuration options at creation time. Widgets are not supposed to query the corresponding plugin because it should be possible to run them stand alone. In this instance, the `collectioneditor` needs to keep track of the dataframe format, so I don't see how to move this method to `dataframeeditor`.

jitseniesen:
I replied to the comment re `collectioneditor` and took care of the rest.

ccordoba12:
I suggested a last couple of changes, the I'd say it's ready to merge :-)

jitseniesen:
> I suggested a last couple of changes, then I'd say it's ready to merge :-)

Done

ccordoba12:
Ok, thanks a lot @jitseniesen!

PR #3626, opened 2016-10-31
PR: Added management for delete operation of a project

### LINKED ISSUE #3618: Removing a project deletes also the files on the disk
When removing a Spyder project within the project explorer, not only is the project deleted but also all files contained in the HDD without any warning about that. This is surely not the behavior one expects, as in Spyder 2 the deletion of files had to be explicitly selected. Other project/versioning software usually also has the Spyder 2 behavior regarding project deletion.

1. Create a project, either new or from an existing directory
2. Delete the project

The **Project** created by Spyder should be removed. But not all the other files contained in the project-directory.

I understand that this case may be wanted sometimes. But it surely should not be the default behavior. What would otherwise be the proper way to just delete a project (NOT the files)?
My proposal would be to either:
- change the behavior of the deletion function to only delete the Spyder-project or
- **explicitly** state that this will also delete ALL the files **from the disk**

- Spyder Version: 3.0.1
- Python Version: 2.7
- Operating system: Ubuntu 16.04 64x
## Dependencies

I think deps don't matter here

Best,
Mayou

dalthviz:
Fixes #3618

### DIFF
--- a/spyder/plugins/explorer.py
+++ b/spyder/plugins/explorer.py
@@ -98,7 +98,7 @@ class Explorer(ExplorerWidget, SpyderPluginMixin):
     def closing_plugin(self, cancelable=False):
         """Perform actions before parent main window is closed"""
         return True
-   
+
     #------ Public API ---------------------------------------------------------
     def chdir(self, directory):
         """Set working directory"""
--- a/spyder/plugins/explorer.py
+++ b/spyder/plugins/explorer.py
@@ -98,7 +98,7 @@ class Explorer(ExplorerWidget, SpyderPluginMixin):
     def closing_plugin(self, cancelable=False):
         """Perform actions before parent main window is closed"""
         return True
-
+   
     #------ Public API ---------------------------------------------------------
     def chdir(self, directory):
         """Set working directory"""
--- a/spyder/plugins/projects.py
+++ b/spyder/plugins/projects.py
@@ -13,7 +13,7 @@ updating the file tree explorer associated with a project
 
 # Standard library imports
 import os.path as osp
-
+import shutil
 # Third party imports
 from qtpy.compat import getexistingdirectory
 from qtpy.QtCore import Signal, Slot
@@ -22,7 +22,7 @@ from qtpy.QtWidgets import QMenu, QMessageBox
 # Local imports
 from spyder.config.base import _, get_home_dir
 from spyder.plugins import SpyderPluginMixin
-from spyder.py3compat import is_text_string, getcwd
+from spyder.py3compat import is_text_string, getcwd, to_text_string
 from spyder.utils import icon_manager as ima
 from spyder.utils.qthelpers import add_actions, create_action
 from spyder.widgets.projects.explorer import ProjectExplorerWidget
@@ -93,6 +93,9 @@ class Projects(ProjectExplorerWidget, SpyderPluginMixin):
         self.close_project_action = create_action(self,
                                     _("Close Project"),
                                     triggered=self.close_project)
+        self.delete_project_action = create_action(self,
+                                    _("Delete Project"),
+                                    triggered=self.delete_project)
         self.clear_recent_projects_action =\
             create_action(self, _("Clear this list"),
                           triggered=self.clear_recent_projects)
@@ -106,6 +109,7 @@ class Projects(ProjectExplorerWidget, SpyderPluginMixin):
                                             None,
                                             self.open_project_action,
                                             self.close_project_action,
+                                            self.delete_project_action,
                                             None,
                                             self.recent_project_menu,
                                             explorer_action]
@@ -190,6 +194,7 @@ class Projects(ProjectExplorerWidget, SpyderPluginMixin):
 
         active = bool(self.get_active_project_path())
         self.close_project_action.setEnabled(active)
+        self.delete_project_action.setEnabled(active)
         self.edit_project_preferences_action.setEnabled(active)
 
     def edit_project_preferences(self):
@@ -277,6 +282,31 @@ class Projects(ProjectExplorerWidget, SpyderPluginMixin):
             self.clear()
             self.restart_consoles()
 
+    def delete_project(self):
+        """
+        Delete current project without deleting the files in the directory
+        """
+        if self.current_active_project:
+            path = self.current_active_project.root_path
+            buttons = QMessageBox.Yes|QMessageBox.No
+            answer = QMessageBox.warning(self, _("Delete"),
+                                 _("Do you really want "
+                                   "to delete <b>%s</b>?<br><br>"
+                                   "<b>Note:</b> This action will only delete "
+                                   "the project. The files are going to be "
+                                   "preserved on disk."
+                                   ) % osp.basename(path), buttons)
+            if answer == QMessageBox.Yes:
+                try:
+                    self.close_project()
+                    shutil.rmtree(osp.join(path,'.spyproject'))
+                except EnvironmentError as error:
+                    action_str = _('delete_project')
+                    QMessageBox.critical(self, _("Project Explorer"),
+                                    _("<b>Unable to %s <i>%s</i></b>"
+                                      "<br><br>Error message:<br>%s" )
+                                    % (action_str, path, to_text_string(error)))
+            
     def clear_recent_projects(self):
         """Clear the list of recent projects"""
         self.recent_projects = []
--- a/spyder/plugins/projects.py
+++ b/spyder/plugins/projects.py
@@ -71,8 +71,6 @@ class Projects(ProjectExplorerWidget, SpyderPluginMixin):
         self.initialize_plugin()
         self.setup_project(self.get_active_project_path())
         
-        #self.treewidget.delete_project.connect(self.delete_project)
-
     #------ SpyderPluginWidget API ---------------------------------------------
     def get_plugin_title(self):
         """Return widget title"""
@@ -285,30 +283,6 @@ class Projects(ProjectExplorerWidget, SpyderPluginMixin):
             self.clear()
             self.restart_consoles()
     
-    """
-    def delete_project(self):
-        Delete current project without deleting the files in the directory.
-        if self.current_active_project:
-            path = self.current_active_project.root_path
-            buttons = QMessageBox.Yes|QMessageBox.No
-            answer = QMessageBox.warning(self, _("Delete"),
-                                 _("Do you really want "
-                                   "to delete <b>{filename}</b>?<br><br>"
-                                   "<b>Note:</b> This action will only delete "
-                                   "the project. Its files are going to be "
-                                   "preserved on disk."
-                                   ).format(filename=osp.basename(path)),
-                                   buttons)
-            if answer == QMessageBox.Yes:
-                try:
-                    self.close_project()
-                    shutil.rmtree(osp.join(path,'.spyproject'))
-                except EnvironmentError as error:
-                    QMessageBox.critical(self, _("Project Explorer"),
-                                    _("<b>Unable to delete <i>{varpath}</i></b>"
-                                      "<br><br>The error message was:<br>{error}" )
-                                    .format(varpath=path,error=to_text_string(error)))"""
-                                    
     def clear_recent_projects(self):
         """Clear the list of recent projects"""
         self.recent_projects = []
--- a/spyder/plugins/projects.py
+++ b/spyder/plugins/projects.py
@@ -70,6 +70,8 @@ class Projects(ProjectExplorerWidget, SpyderPluginMixin):
         # Initialize plugin
         self.initialize_plugin()
         self.setup_project(self.get_active_project_path())
+        
+        self.treewidget.delete_project.connect(self.delete_project)
 
     #------ SpyderPluginWidget API ---------------------------------------------
     def get_plugin_title(self):
@@ -292,7 +294,7 @@ class Projects(ProjectExplorerWidget, SpyderPluginMixin):
                                  _("Do you really want "
                                    "to delete <b>{filename}</b>?<br><br>"
                                    "<b>Note:</b> This action will only delete "
-                                   "the project. The files are going to be "
+                                   "the project. Its files are going to be "
                                    "preserved on disk."
                                    ).format(filename=osp.basename(path)),
                                    buttons)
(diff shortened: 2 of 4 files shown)


spyder/plugins/projects.py
```diff
@@ -13,7 +13,7 @@
 
 # Standard library imports
 import os.path as osp
-
+import shutil
```
goanpeca: Add a space here @dalthviz :-)

spyder/plugins/projects.py
```diff
@@ -277,6 +282,31 @@ def close_project(self):
             self.clear()
             self.restart_consoles()
 
+    def delete_project(self):
+        """
+        Delete current project without deleting the files in the directory
```
goanpeca: A single line and a dot at the end please :-)

`"""Delete current project without deleting the files in the directory."""`

spyder/plugins/projects.py
```diff
@@ -277,6 +282,31 @@ def close_project(self):
...
+        if self.current_active_project:
+            path = self.current_active_project.root_path
+            buttons = QMessageBox.Yes|QMessageBox.No
+            answer = QMessageBox.warning(self, _("Delete"),
+                                 _("Do you really want "
+                                   "to delete <b>%s</b>?<br><br>"
+                                   "<b>Note:</b> This action will only delete "
+                                   "the project. The files are going to be "
+                                   "preserved on disk."
+                                   ) % osp.basename(path), buttons)
```
goanpeca: In general is nicer to use [format](https://docs.python.org/3.5/library/string.html#format-examples)

spyder/plugins/projects.py
```diff
@@ -277,6 +282,31 @@ def close_project(self):
...
+                try:
+                    self.close_project()
+                    shutil.rmtree(osp.join(path,'.spyproject'))
+                except EnvironmentError as error:
+                    action_str = _('delete_project')
+                    QMessageBox.critical(self, _("Project Explorer"),
+                                    _("<b>Unable to %s <i>%s</i></b>"
+                                      "<br><br>Error message:<br>%s" )
+                                    % (action_str, path, to_text_string(error)))
+
```
goanpeca: Same here

spyder/widgets/explorer.py
```diff
@@ -532,7 +532,22 @@ def delete_file(self, fname, multiple, yes_to_all):
                               "<br><br>Error message:<br>%s"
                               ) % (action_str, fname, to_text_string(error)))
         return False
-
+    
+    def delete_project(self, pname):
+        """Deletes the .spyproject diretory"""
```
goanpeca: typo, diretory -> directory

dot at the end. and imperative mood ;-)

`"""Delete the .spyproject directory."""`

spyder/widgets/explorer.py
```diff
@@ -532,7 +532,22 @@ def delete_file(self, fname, multiple, yes_to_all):
...
+    def delete_project(self, pname):
+        """Deletes the .spyproject diretory"""
+        buttons = QMessageBox.Yes|QMessageBox.No
+        answer = QMessageBox.warning(self, _("Delete"),
+                                 _("Do you really want "
+                                   "to delete <b>%s</b>?<br><br>"
+                                   "<b>Note:</b> This action will only delete "
+                                   "the project. The files are going to be "
+                                   "preserved on disk."
+                                   ) % osp.basename(pname), buttons)
```
goanpeca: format is nicer :-p

spyder/widgets/explorer.py
```diff
@@ -532,7 +532,22 @@ def delete_file(self, fname, multiple, yes_to_all):
...
+        if answer == QMessageBox.Yes:
+            self.parent_widget.delete_project()
+            return True
+        return False
```
goanpeca: In general I (probably is a personal opinion?) do not like the idea of having multiple exit points on a method (several returns...)

In this case they are so close to each other that it does not really matter but I believe having something like

is more clear
```python
result = False
if answer == QMessageBox.Yes:
    self.parent_widget.delete_project()
    result = True

return result
```

Or even

```python
result = answer == QMessageBox.Yes
if result:
    self.parent_widget.delete_project()
return result
```

spyder/widgets/explorer.py
```diff
@@ -541,11 +556,14 @@ def delete(self, fnames=None):
...
-                # Canceled
-                return
-
+            if osp.isdir(fname) and osp.exists(osp.join(fname,'.spyproject')):
+                self.delete_project(fname)
+            else:    
+                yes_to_all = self.delete_file(fname, multiple, yes_to_all)
+                if yes_to_all is not None and not yes_to_all:
+                    # Canceled
+                    return
```
goanpeca: Better to `break` the loop?

spyder/plugins/projects.py
```diff
@@ -277,6 +283,31 @@ def close_project(self):
...
+                try:
+                    self.close_project()
+                    shutil.rmtree(osp.join(path,'.spyproject'))
+                except EnvironmentError as error:
+                    action_str = _('delete_project')
+                    QMessageBox.critical(self, _("Project Explorer"),
+                                    _("<b>Unable to {action} <i>{varpath}</i></b>"
+                                      "<br><br>Error message:<br>{text}" )
+                                    .format(action = action_str, varpath=path, 
+                                    text=to_text_string(error)))
```
ccordoba12: Please change these lines to

```python
_("<b>Unable to delete the project <i>{path}</i></b>"
  "<br><br>The error message was:<br>{error}" )
  .format(path=path, error=to_text_string(error)))
```

spyder/widgets/explorer.py
```diff
@@ -541,11 +545,14 @@ def delete(self, fnames=None):
...
-                return
-
+            if osp.isdir(fname) and osp.exists(osp.join(fname,'.spyproject')):
+                self.delete_project(fname)
+            else:    
+                yes_to_all = self.delete_file(fname, multiple, yes_to_all)
+                if yes_to_all is not None and not yes_to_all:
+                    # Canceled
+                    break
+
```
ccordoba12: These changes must not be here. They need to be placed in `widgets/projects/explorer.py` because this is a functionality that belongs only to projects and not to the file explorer.

In that file you need to define a new `delete` method in `ExplorerTreeWidget`, which will override the `delete` method of `FilteredDirView` (of which `ExplorerTreeWidget` is a subclass of).

Besides, the logic you defined can be improved a bit to make it easier to follow. You could write something like:

```python
if fname == self.root_path:
    self.delete_project(fname)
elif '.spyproject' in fname:
    self.delete_project(fname)
else:
   ...
```

The first condition applies when users try to delete the first entry in the explorer (which corresponds to the project's root directory), and the second one to the `.spyproject` dir (which is only visible on Windows, as we discussed yesterday :-).

ccordoba12: Sorry, since `delete_project` in `plugins/projects.py` doesn't receive any arguments, the logic is

```python
if fname == self.root_path:
    self.delete_project.emit()
elif '.spyproject' in fname:
    self.delete_project.emit()
else:
   ...
```

```python
delete_project = Signal()
```

spyder/widgets/explorer.py
```diff
@@ -12,6 +12,7 @@
 # pylint: disable=R0201
 
 # Standard library imports
+from __future__ import print_function
```
ccordoba12: I'm sure this is not needed :-)

spyder/widgets/explorer.py
```diff
@@ -532,7 +535,7 @@ def delete_file(self, fname, multiple, yes_to_all):
                               "<br><br>Error message:<br>%s"
                               ) % (action_str, fname, to_text_string(error)))
         return False
-
+
```
ccordoba12: Remove blank

spyder/widgets/explorer.py
```diff
@@ -541,11 +544,20 @@ def delete(self, fnames=None):
...
         yes_to_all = None
         for fname in fnames:
-            yes_to_all = self.delete_file(fname, multiple, yes_to_all)
-            if yes_to_all is not None and not yes_to_all:
-                # Canceled
-                return
-
+            spyproject_path = osp.join(fname,'.spyproject')
+            if osp.isdir(fname) and osp.exists(spyproject_path):
+                QMessageBox.information(self, _('Archive Explorer'),
```
ccordoba12: Please change `Archive Explorer` to `File Explorer`

spyder/widgets/projects/explorer.py
```diff
@@ -30,6 +30,8 @@
 
 class ExplorerTreeWidget(FilteredDirView):
     """Explorer tree widget"""
+    
+    delete_project = Signal()
```
ccordoba12: I forgot to tell you that we decided to add the `sig_` prefix for new signals, so this one should be `sig_delete_project` :-)

spyder/widgets/projects/explorer.py
```diff
@@ -149,12 +151,29 @@ def dropEvent(self, event):
                                        "<br><br>Error message:<br>%s"
                                        ) % (action_str, src,
                                             to_text_string(error)))
+   
+    @Slot()
```
ccordoba12: This doesn't need to be here because this method is not a slot :-)

spyder/widgets/projects/explorer.py
```diff
@@ -149,12 +151,29 @@ def dropEvent(self, event):
...
+                if yes_to_all is not None and not yes_to_all:
+                    # Canceled
+                    break
 
 
 class ProjectExplorerWidget(QWidget):
     """Project Explorer"""
     sig_option_changed = Signal(str, object)
     sig_open_file = Signal(str)
+    sig_delete_project = Signal()
```
ccordoba12: This is not needed because it's defined in `plugins/projects.py`

spyder/widgets/projects/explorer.py
```diff
@@ -211,11 +230,16 @@ def setup_project(self, directory):
         self.emptywidget.hide()
         self.treewidget.show()
         self.layout().addWidget(self.treewidget)
-
+
```
ccordoba12: Remove blank

spyder/widgets/projects/explorer.py
goanpeca: But should not this be a standalone widget? shoul the treeview connect to this signal?

ccordoba12: Mmmh, you're right. @dalthviz, please move the `delete_project` method from `plugins/projects.py` to this widget :-)

ccordoba12: But even after doing that, this signal is not necessary.

spyder/widgets/projects/explorer.py
```diff
@@ -211,11 +230,16 @@ def setup_project(self, directory):
...
         self.set_project_dir(directory)
-
-
+     
+        # Signal to delete the project
+        self.treewidget.delete_project.connect(self.delete_project)
+    
+    def delete_project(self):
+        self.sig_delete_project.emit()
+
```
ccordoba12: And this needs to be replaced by `delete_project` of `plugins/projects.py`

ccordoba12:
@dalthviz, great job so far!! We left some final comments, but almost all are about form and not substance (which you handled very well :-).

Keep up the good work!

spyder/widgets/explorer.py
```diff
@@ -541,11 +544,20 @@ def delete(self, fnames=None):
...
+                                        _("The current directory contains a "
+                                        "project.<br><br>If you want to delete"
+                                        " the project, please go to "
+                                        "<b>Projects</b> &raquo; <b>Delete "
+                                        "Project</b>"))
+            else:    
+                yes_to_all = self.delete_file(fname, multiple, yes_to_all)
+                if yes_to_all is not None and not yes_to_all:
+                    # Canceled
+                    break
```
dalthviz: This change was part of one of the comments of the first review that Gonzalo made.

ccordoba12: Ok, no prob then :-)

spyder/plugins/explorer.py
```diff
@@ -98,7 +98,7 @@ def refresh_plugin(self, new_path=None, force_current=True):
     def closing_plugin(self, cancelable=False):
         """Perform actions before parent main window is closed"""
         return True
-
+
```
ccordoba12: Don't forget this one :-)

spyder/plugins/projects.py
```diff
@@ -69,6 +70,8 @@ def __init__(self, parent=None):
         # Initialize plugin
         self.initialize_plugin()
         self.setup_project(self.get_active_project_path())
+        
+        #self.treewidget.delete_project.connect(self.delete_project)
```
ccordoba12: Remove this comment

spyder/widgets/explorer.py
ccordoba12: Don't forget

spyder/widgets/projects/explorer.py
ccordoba12: This is not needed, please remove it

spyder/widgets/projects/explorer.py
ccordoba12: Don't forget :-)

spyder/widgets/explorer.py
```diff
@@ -1046,7 +1066,7 @@ def chdir(self, directory=None, browsing_history=False):
                                  _("You don't have the right permissions to "
                                    "open this directory"))
 
-
+
```
dalthviz: But this was to separete the class definition with two blank lines, no?

spyder/plugins/projects.py
```diff
@@ -276,7 +282,7 @@ def close_project(self):
             self.dockwidget.close()
             self.clear()
             self.restart_consoles()
-
+
```
ccordoba12: Blank

spyder/plugins/projects.py
```diff
@@ -69,7 +70,7 @@ def __init__(self, parent=None):
         # Initialize plugin
         self.initialize_plugin()
         self.setup_project(self.get_active_project_path())
-
+
```
ccordoba12: Blank

spyder/plugins/projects.py
```diff
@@ -22,7 +23,7 @@
 # Local imports
 from spyder.config.base import _, get_home_dir
 from spyder.plugins import SpyderPluginMixin
-from spyder.py3compat import is_text_string, getcwd
+from spyder.py3compat import is_text_string, getcwd, to_text_string
```
ccordoba12: This import (to_text_string) is not needed in this file anymore :-)

ccordoba12:
Great stuff!! I like the result a lot :-)

PR #3629, opened 2016-10-31
PR: Add shortcut for UpperLower functionality

malliwi88:
Now I will work on the logic over codeeditor.py for add the logic

PR #3639, opened 2016-11-02
Style suggestions made

malliwi88:


ccordoba12:
@malliwi88, why are you opening a new PR every time you push new changes??

You don't need to do that!! This is all wrong, and please don't do it again (you've done it a lot of times already!). So I'm going to close this PR and ask you to continue your work in PR #3629.

It seems you're forgetting to checkout your branch before committing, so before doing any commit, please be sure to be in the right branch :-)

PR #3640, opened 2016-11-02
PR: Check if files are text files using the binaryornot package

### LINKED ISSUE #3631: Edit context menu appears only for utf-8 files in project explorer

I created a new project from an existing folder. Some .py-files in the Project explorer had an "Edit" context menu, others only an "Open" context menu. Drag and drop into the editor would also not work. 

A short look in the sources revealed that in explorer.py line 261:
def create_file_manage_actions
only utf encoded files are accepted as editable. 

I have no problem with this behavior but it should be easy to find in the documentation. 

1. Create a .py file with ANSI encoding and try the context menu

* Spyder Version: 3.0.1
* Python Version: 3.5.2
* Operating system: Win 10

## Dependencies

and paste the contents below:

pandas >=0.13.1  :  0.19.0 (OK)
sphinx >=0.6.6   :  1.3.1 (OK)
sympy >=0.7.3    :  1.0 (OK)

### LINKED ISSUE #3670: Dragging and dropping files in the Editor is failing
## Description of problem

I've just upgraded from Spyder 2.x to Spyder 3.x, and noticed it is no longer possible to drag-and-drop files to open them. Not sure if this is intentional functionality or not, but opening files on windows now has become a bit clunky.

**Try the following steps to reproduce the problem:**

1. Have Spyder 3.0.x version installed
2. Open Spyder & drag drop .py file into editor

**Expected and given output:**

Based on the previous version of Spyder, I expect dragging and dropping in a file to open that file in the editor, instead it's copying the file location.

* Spyder Version: 3.0.1
* Python Version: 3.5

rlaverde:
~~Add misc.is_supported_file function, and use it instead of encoding.is_text_file~~

**EDIT**
Add [binaryornot](https://github.com/audreyr/binaryornot) package  and use it to determinate if a file is text

Fixes #3631 
Fixes #3670

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,7 @@ a Python version greater than 2.7 (Python 3.2 is not supported anymore).
 * **PyZMQ**: Run introspection services asynchronously.
 * **QtPy** 1.1.0+: Abstracion layer for Python Qt bindings so that Spyder can run on PyQt4
   and PyQt5.
+* **Chardet**: Character encoding auto-detection in Python.
 
 ### Optional dependencies
 
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -32,6 +32,7 @@ requirements:
     - qtawesome
     - qtpy >=1.1.0
     - pyzmq
+    - chardet >=2.0.0
 
 about:
   home: https://github.com/spyder-ide/spyder
--- a/continuous_integration/conda-recipes/spyder/meta.yaml
+++ b/continuous_integration/conda-recipes/spyder/meta.yaml
@@ -33,6 +33,7 @@ requirements:
     - pylint
     - qtawesome
     - qtpy
+    - chardet >=2.0.0
 
 about:
   home: https://github.com/spyder-ide/spyder
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -194,6 +194,8 @@ The requirements to run Spyder are:
 * `QtPy <https://github.com/spyder-ide/qtpy>`_ >=1.1.0 -- To run Spyder with PyQt4 or
   PyQt5 seamlessly.
 
+* `Chardet <https://github.com/chardet/chardet>`_ >=2.0.0-- Character encoding auto-detection
+  in Python.
 
 Optional modules
 ~~~~~~~~~~~~~~~~
--- a/setup.py
+++ b/setup.py
@@ -283,7 +283,8 @@ install_requires = [
     'qtawesome',
     'qtpy>=1.1.0',
     'pickleshare',
-    'pyzmq'
+    'pyzmq',
+    'chardet>=2.0.0',
 ]
 
 if 'setuptools' in sys.modules:
--- a/spyder/utils/encoding.py
+++ b/spyder/utils/encoding.py
@@ -20,6 +20,7 @@ from codecs import BOM_UTF8, BOM_UTF16, BOM_UTF32, getincrementaldecoder
 # Local imports
 from spyder.py3compat import (is_string, to_text_string, is_binary_string,
                               is_unicode)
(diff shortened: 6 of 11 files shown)


ccordoba12:
@rlaverde, what you did is an easy workaround because the main problem is that `is_text_file` is failing for files with an encoding different from `utf-8` and with non-ascii characters on them.

The right fix (after reading the Stackoverflow post where `is_text_file` comes from) is to define a new function called `is_binary_file` (in `utils/encoding.py`):

```python
def is_binary(filename):
    """Return true if the given filename is binary.

    Taken from: http://stackoverflow.com/a/3002505

    author: Trent Mick <TrentM@ActiveState.com>
    author: Jorge Orpinel <jorge@orpinel.com>
    """
    fin = open(filename, 'rb')
    try:
        CHUNKSIZE = 1024
        while 1:
            chunk = fin.read(CHUNKSIZE)
            if b'\0' in chunk: # found null byte
                return True
            if len(chunk) < CHUNKSIZE:
                break # done
    finally:
        fin.close()

    return False
```

and redefine `is_text_file` to be

```python
def is_text_file(filename):
    """
    Test if the given path is a text-like file.
    """
    not is_binary_file(filename)
```

ccordoba12:
On a second thought. it seems more reliable to use an external package for this. Please add this package: https://github.com/audreyr/binaryornot to `utils/external`, import its function called `is_binary` in `utils/encoding.py` and define `is_text_file` in the same way as above.

----

Note: Since this is needed for `3.x` we can't add a new dependency on that package. Besides, it's very small and it seems to move very slowly :-)

rlaverde:
@ccordoba12 I added binaryornot and change `is_text_file` function as you suggested, should I create a PR against master adding the package as a dependency?

ccordoba12:
> should I create a PR against master adding the package as a dependency?

Not right now because that will force Continuum and Linux distros to create `binaryornot` packages in their respective repos.

Since it's a very small package, I think we can live with it in our repo for now :-)

rlaverde:
[chardet](https://github.com/chardet/chardet) module is missing, should I also include it? (It isn't a tiny one)

ccordoba12:
> chardet module is missing

I didn't notice that! Thanks for noticing it ;-)

> should I also include it? (It isn't a tiny one)

Nop, please don't (especially if it's not tiny :-). What that heck, please add `chardet` to our setup.py,  conda.recipe, README.md and doc/installation.rst.

I don't think Ubuntu is going to pick up **3.1** for 17.04, so I don't see problems in adding a new dependency.

ccordoba12:
And please also add a simple test in `utils` to verify that `binaryornot` is working correctly (i.e. that `chardet` is picked up as a dependency) by calling `is_text_file`.

rlaverde:
> I don't think Ubuntu is going to pick up 3.1 for 17.04, so I don't see problems in adding a new dependency.

I don't think this will be a problem chardet is already available in [ubuntu](http://packages.ubuntu.com/zesty/python-chardet) (also [debian](https://packages.debian.org/jessie/python-chardet)) repositories

ccordoba12:
Annnd by using `chardet` we could correctly report the encoding of files in the Editor!

I'll open a new issue for you to work at @rlaverde!

rlaverde:
This is failing because a circular importing, that only happen on windows
encoding.py --> external (__init__) --> base.py --> encoding.py

https://github.com/spyder-ide/spyder/blob/master/spyder/utils/external/__init__.py#L22

ccordoba12:
> how this could be fixed?

Just remove all code in `utils/external/__init__.py`. It's not needed anymore, I just forget to remove it in Spyder 3.0 ;-)

ccordoba12:
Nice, good job!

Now that `chardet` is an official dep, please go on with issue #3731.

PR #3653, opened 2016-11-05
PR: Added support to the principal numeric Numpy types in the Variable Explorer

### LINKED ISSUE #3638: Variable explorer should display the value of numpy.int64 variables
Execute
```
import numpy
j = numpy.int64(17)
```
.

Now `j` shows up in the variable explorer, but the Value column displays `class 'numpy.int64`. It should display the value `17` instead.

(My current environment is described in #3562.)

dalthviz:
Fixes #3638

### DIFF
--- a/spyder/widgets/variableexplorer/utils.py
+++ b/spyder/widgets/variableexplorer/utils.py
@@ -44,10 +44,12 @@ class FakeObject(object):
 
 
 #==============================================================================
-# Numpy arrays support
+# Numpy arrays and numeric support
 #==============================================================================
 try:
-    from numpy import ndarray, array, matrix, recarray
+    from numpy import ndarray, array, matrix, recarray, \
+                      int64, int32, float64, float32, complex_, \
+                      complex64, complex128
     from numpy.ma import MaskedArray
     from numpy import savetxt as np_savetxt
     from numpy import set_printoptions as np_set_printoptions
@@ -289,6 +291,18 @@ def value_to_display(value, minmax=False):
             display = value
         elif isinstance(value, NUMERIC_TYPES) or isinstance(value, bool) or \
           isinstance(value, datetime.date):
+            display = repr(value)
+        elif isinstance(value, int64):
+            display = repr(value)
+        elif isinstance(value, int32):
+            display = repr(value)
+        elif isinstance(value, float64):
+            display = repr(value)
+        elif isinstance(value, float32):
+            display = repr(value)
+        elif isinstance(value, complex_) or isinstance(value, complex128) :
+            display = repr(value)
+        elif isinstance(value, complex64):
             display = repr(value)
         else:
             # Note: Don't trust on repr's. They can be inefficient and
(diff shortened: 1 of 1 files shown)


dalthviz:
Added support in the variable explorer to the following numpy data types: `int32`, `int64`, `float32`, `float64`, `complex64`, `complex_` (shorthand for `complex128`) and `complex128`

numpy

wtheis:
If possible, please include this in 3.0.2.
(#3638 is currently 3.1.)

spyder/widgets/variableexplorer/utils.py
```diff
@@ -44,10 +44,12 @@ class FakeObject(object):
...
 
 #==============================================================================
-# Numpy arrays support
+# Numpy arrays and numeric support
 #==============================================================================
 try:
-    from numpy import ndarray, array, matrix, recarray
+    from numpy import ndarray, array, matrix, recarray, \
+                      int64, int32, float64, float32, complex_, \
+                      complex64, complex128
```
ccordoba12: Please enclose imports in parenthesis

spyder/widgets/variableexplorer/utils.py
```diff
@@ -292,17 +295,7 @@ def value_to_display(value, minmax=False):
...
-        elif isinstance(value, int32):
-            display = repr(value)
-        elif isinstance(value, float64):
-            display = repr(value)
-        elif isinstance(value, float32):
-            display = repr(value)
-        elif isinstance(value, complex_) or isinstance(value, complex128) :
-            display = repr(value)
-        elif isinstance(value, complex64):
+        elif True in [isinstance(value, x) for x in numeric_type_list]:
```
jitseniesen: I'd write this as `any(isinstance(value, x) for x in numeric_type_list)` or `isinstance(value, numeric_type_tuple)`; in the latter case, I think it needs to be a tuple of types, not a list of types.

ccordoba12: I also like the second option. @dalthviz, please do the corresponding change :-)

ccordoba12:
> If possible, please include this in 3.0.2.

@wtheis, unfortunately 3.0.2 is almost done, and we don't to delay it longer.

spyder/widgets/variableexplorer/utils.py
```diff
@@ -244,6 +247,8 @@ def unsorted_unique(lista):
 def value_to_display(value, minmax=False):
     """Convert value for display purpose"""
     try:
+        numeric_type_tuple = (int64, int32, float64, float32, \
```
ccordoba12: @dalthviz, I think you don't need a continuation line (`\`) here

ccordoba12: Also, please rename this variable to `numeric_numpy_types`

spyder/widgets/variableexplorer/utils.py
```diff
@@ -44,16 +44,19 @@ class FakeObject(object):
 
 
 #==============================================================================
-# Numpy arrays support
+# Numpy arrays and numeric support
```
ccordoba12: Please change this to `Numpy arrays and numeric types support`

PR #3656, opened 2016-11-06
PR: Added shortcut for fallback completion

### LINKED ISSUE #3469: simple word completetion
this is a feature request.
In many IDEs (pycharm, sublime, vim, and in idle) there is a simple word completion (usually mapped to Alt+/)
this will autocomplete any word in the current file without understanding python at all
It is very useful when writing function and autocompleting the name of new variables from similar functions or documentation or similar.

example of usage:

    def foobar(var_list_a):
        """ returns calculated prices_of_list """
        price_of<Alt+/>

will autocomplete to `prices_of_list`. regular autocomplete can't do it because it is in a documentation.
This shouldn't be a very hard feature I guess.

Thanks!

dalthviz:
Fixes #3469


dalthviz:
Now you can use `Alt+/` to get fallback completion suggestions:

fallback

Issue #3651, opened 2016-11-04
Improve indentation tests


rlaverde:
Indentation tests only test for indentation using spaces, and It's necessary to also test for indentation using tabs (that is partially broken #3508)

I'm thinking about using parametrize fixture to make test support several options

and I have consider different ways to organize test:

1: Each test will have parametrize options for tabs and spaces:
```
@pytest.mark.parametrize("indent_chars, text_input, expected", [
    ("    ", "\ndef function():\n", "\ndef function():\n    "),
    ("\t", "\ndef function():\n", "\ndef function():\n\t"),
])
def test_def_with_newline(indent_chars, text_input, expected):
    text = get_indent_fix(text_input, indent_chars)
    assert text == expected, repr(text)
```

2. Each test will have parametrize options for tabs and spaces, but strings will be generated into the test:

```
@pytest.mark.parametrize("indent_chars", [
    ("    "), ("\t"),
])
def test_def_with_newline(indent_chars):
    text = get_indent_fix("\ndef function():\n", indent_chars)
    assert text == "\ndef function():\n" + indent_chars, repr(text)
```

3. will be one test or maybe two (tabs and spaces) with several parametrized options:

```
@pytest.mark.parametrize("indent_chars, text_input, expected", [
    ("    ", "\ndef function():\n", "\ndef function():\n    "),
    ("\t", "\ndef function():\n", "\ndef function():\n\t"),

    ("    ","def function():\n    print []\n" , "def function():\n\tprint []\n    "),
    ("\t","def function():\n\tprint []\n" , "def function():\n\tprint []\n\t"),
])
def test_indentation(indent_chars, text_input, expected):
    text = get_indent_fix(text_input, indent_chars)
    assert text == expected, repr(text)
```

which option could be the best?

cc: @Nodd @goanpeca

ccordoba12:
Was this already fixed in #3680 @rlaverde?

ccordoba12:
Sorry, #3663?

rlaverde:
Yes, I implemented the this way

PR #3487, opened 2016-09-30
PR: Add hanging Indentation to the Editor

yulkang:
Fixes #3480 

----

This commit has an option for hanging indent at menu - python - Preferences - Editor - Advanced settings - Hanging indentation. It is unchecked by default, but when checked, allows hanging indentation, as follows:

```
def long_function_name(
        var1, var2):
    print("This works. Yeah!")

if long_function_name(
        var1, var2):
    print("also works")
```

The case I found not working is the following, where the subsequent line is indented one step too much:

```
if long_array_name[
        var1]:
            print("this does not work yet")
```

But such cases are rarer for me, so it would be useful even as it stands now. I haven't found why this happens, and I would appreciate help.

I checked the following and they work as expected, too:
- the new option is saved after quitting, and
- the default behavior, vertical alignment, is recovered when the new option is unchecked.

Please let me know if there are other issues.

PR #3663, opened 2016-11-08
PR: Fix some errors of Indentation with tabs

### LINKED ISSUE #3508: Indentation issue using Tabs

Using tabs for indentation, when indenting a new line in the middle of a statement (e.g. between parentheses/braces/brackets), it inserts more tabs than expected (looks like current indentation + 5), and there is no way to configure it.  
1. Set up the editor to use Tabulations for indentation
2. in the editor, type a = {}
3. with the cursor between the braces hit enter

expected: the new line is indented with 1 (or 2; configurable) tabs
actual: the new line is indented with 5 tabs

Hitting tab after removing any/all indentation immediately adds back to the original number of tabs.
Even if the first line in the continuation block (additional lines associated with a single statement) is indented the way I expect, subsequent lines still start with the original number of tabs rather than aligning with the previous line.  This is the most obnoxious part of this issue as that means I have to manually correct each line of a continuation block.

Suggested behavior:
Continuation lines should indent with one more unit of indentation than the parent of that continuation block.  Subsequent lines should indent with the same indentation as the previous line.  Continuation blocks may be nested (e.g. if calling a function with many parameters, where one parameter is another function call with many parameters and I want to have each parameter on its own line, or if defining a nested json-like structure of dicts and lists), in which case each new nesting would increase the indentation by one unit of indentation.

It would be nice if the continuation indentation were configurable as well, as some users may prefer 2 indents instead of 1.  
- Spyder Version: 3.0.0
- Python Version: 3.4.3
- Operating system: Ubuntu 14.04
- Qt and PyQt Version: 5.2.1
## Dependencies

and paste the contents below:

I don't see a menu.  
I installed it using pip.

rlaverde:
Fixes #3508 

----------------

- Added some test for indenting with tabs, I used the 3dr way that I mencioned in #3651 (but didn't change spaces indentation tests)
- Added test for #3508 
- Change how `correct_indent` is calculated, when using tabs It will be calculated with spaces (each tab 4 spaces) and then convert it to tabs and reminder to spaces

I think this could be improved if the option `tab_stop_width` is defined in terms of spaces instead of pixels.

### DIFF
--- a/spyder/widgets/mixins.py
+++ b/spyder/widgets/mixins.py
@@ -382,6 +382,7 @@ class BaseEditMixin(object):
     def get_block_indentation(self, block_nb):
         """Return line indentation (character number)"""
         text = to_text_string(self.document().findBlockByNumber(block_nb).text())
+        text = text.replace("\t", " "*4)
         return len(text)-len(text.lstrip())
     
     def get_selection_bounds(self):
--- a/spyder/widgets/sourcecode/codeeditor.py
+++ b/spyder/widgets/sourcecode/codeeditor.py
@@ -1926,11 +1926,12 @@ class CodeEditor(TextEditBaseWidget):
                 else:
                     break
 
-        if not prevline:
-            return False
+        if prevline:
+            correct_indent = self.get_block_indentation(prevline)
+        else:
+            correct_indent = 0
 
         indent = self.get_block_indentation(block_nb)
-        correct_indent = self.get_block_indentation(prevline)
 
         if add_indent:
             if self.indent_chars == '\t':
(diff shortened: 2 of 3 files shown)


jitseniesen:
Does this interfere with #3487? Just checking, because that's also about indentation, but it changes a different part of the code.

rlaverde:
Yes, I think it interferes with #3487, I'll rebase my code an resolve conflicts after #3487 will be merged

ccordoba12:
This looks quite good to me, and tests are passing too, which is great!

PR #3664, opened 2016-11-08
Free memory when closing a Variable Explorer editor

### LINKED ISSUE #3513: Spyder does not free up memory after closing windows with datasets in the Variable explorer
## Description

0.~Optionally, display memory usage (Preferences --> General --> Advanced Settings --> Check the box named 'Show memory usage')
1. Read a large data set. For ex.: [.csv 300 MB](https://contest.sdsj.ru/upload/transactions.csv) (the link may not be assessible after November, 6th, 2016)
using the following code

`import pandas as pd`
`raw = pd.read_csv('transactions.csv')`

2.~Once you have read it into your RAM, try to glimpse at it by double clicking in the Variable Explorer. You will see a window that displays your data set.
3. Close this window
4. Repeat Step 2 until your RAM is full.

1. It is reasonable to show only a couple of hundreds of rows and when a user wants more -- load more.
2. Free up RAM after Step 3.
- Spyder Version: 3.0.0 
- Python Version: 3.5.2
- Qt Versions: 4.8.7, PyQt4 (API v2) 4.11.4 on Linux
## Dependencies

```
pandas >=0.13.1  :  0.17.1 (OK)
numpy >=1.7      :  1.11.0 (OK)

```

ccordoba12:
Fixes #3513.

----

This doesn't completely solve #3513, but it's the best I could come up with. Unfortunately, I don't know how things could be improved :-/

* Pros:
  - It prevents Spyder to fully eat your RAM when inspecting several large variables.
* Cons:
  - It doesn't free RAM for the *first* variable you inspect. A trick to make Spyder freeing that memory is to create a small variable (e.g. an empty dataframe), and then open and close it.

### DIFF
--- a/spyder/widgets/variableexplorer/collectionseditor.py
+++ b/spyder/widgets/variableexplorer/collectionseditor.py
@@ -19,6 +19,7 @@ Collections (i.e. dictionary, list and tuple) editor widget and dialog
 # Standard library imports
 from __future__ import print_function
 import datetime
+import gc
 import sys
 import warnings
 
@@ -519,9 +520,15 @@ class CollectionsDelegate(QItemDelegate):
             conv_func = data.get('conv', lambda v: v)
             self.set_value(index, conv_func(value))
         self._editors.pop(editor_id)
+        self.free_memory()
         
     def editor_rejected(self, editor_id):
         self._editors.pop(editor_id)
+        self.free_memory()
+
+    def free_memory(self):
+        """Free memory after closing an editor."""
+        gc.collect()
 
     def commitAndCloseEditor(self):
         """Overriding method commitAndCloseEditor"""


vdyashin:
Thanks. Looking forward to see the next update :)

ccordoba12:
Please be aware that this will only work when you press the *Ok* or *Cancel* buttons of our viewers.

Besides, I'm planning to remove the *Ok* button in Spyder 3.1 because if you press it for large variables, it'll eat your memory too.

vdyashin:
@ccordoba12 okay thank you.

By the way, what is the difference between pressing X on top of the window, 'Cancel', and 'Ok'? As far as I am concerned when I change a value by double-clicking on it in variable explorer I can either save (Ok) or do not (Cancel or X button) or is there a more subtle difference?

ccordoba12:
* *Cancel* and *X* just dismiss the dialog (and now they'll free the memory used to show a variable).
* *Ok* takes the new value in the viewer, serializes it and sends it to the kernel, so it can be updated there. The problem with *Ok* is that for large variables the serialization/deserialization necessary to move the value from Spyder to the IPython kernel consumes quite a lot of memory and can block your computer completely.

I still have to investigate if we're doing something wrong with *Ok* or not :-)

PR #3676, opened 2016-11-11
PR: Refactor and simplify NamespaceBrowser widget

ccordoba12:
This PR also removes the connection between the Variable Explorer and the Python Console. This starts the process of moving that console out of Spyder because it's unmaintained.


ccordoba12:
I think this is ready. Let's merge it and keep iterating if need be :-)

PR #3682, opened 2016-11-12
Add new fuzzy codecompletion feature

malliwi88:
At the moment it's a demo version for this feature.

* I'm planning to add automatic recognition for association of icons with
completion item (Actually I have problems with this)

* I would like to add some description on right of suggested word, for
indicate your nature (module, class, instances, functions, etc.)

* And finally, if suggested word corresponds with a function, display
details about arguments that it will need (It's  would be util for a
future IntelliSense code completion feature)

What do you think guys?
@ccordoba12  @goanpeca

### DIFF
--- a/spyder/widgets/sourcecode/base.py
+++ b/spyder/widgets/sourcecode/base.py
@@ -18,7 +18,7 @@ import sys
 
 # Third party imports
 from qtpy.compat import to_qvariant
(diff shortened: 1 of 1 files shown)


malliwi88:
Hi 

I will close this PR for replacing it. A had some troubles with my fork, so I think, prevention is better than cure =P

PR #3692, opened 2016-11-15
Add new fuzzy codecompletion feature

malliwi88:
At the moment it's a demo version for this feature.

* It hasn't code completion without case sensitive because I thinking is needed enrich with more suggestions the list of code completion for the widget will have several options during case sensitive search.

* I'm planning to add automatic recognition for association of icons with
completion item (Actually I have problems with this)

* I would like to add some description on right of suggested word, for
indicate your nature (module, class, instances, functions, etc.)

* And finally, if suggested word corresponds with a function, display
details about arguments that it will need (It's would be util for a
future IntelliSense code completion feature)

What do you think guys?
@ccordoba12 @goanpeca

### DIFF
--- a/spyder/widgets/sourcecode/base.py
+++ b/spyder/widgets/sourcecode/base.py
@@ -18,7 +18,7 @@ import sys
 
 # Third party imports
 from qtpy.compat import to_qvariant
-from qtpy.QtCore import QEvent, QEventLoop, QPoint, Qt, Signal, Slot
+from qtpy.QtCore import QEvent, QEventLoop, QPoint, Qt, Signal,QSize, Slot
 from qtpy.QtGui import (QClipboard, QColor, QFont, QMouseEvent, QPalette,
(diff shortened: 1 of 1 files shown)


goanpeca:
@malliwi88 can you paste some screenshots or a small animation using something like http://www.cockos.com/licecap/

ccordoba12:
This is a good start, thanks @malliwi88! You have a problem with your PR though, so I think you need to recreate it yet again :-)

The thing is you didn't create a branch for this PR, and instead pushed directly to the `3.x` branch of your fork. This will prevent you to create new PRs until this PR is merged :-)

To avoid this in the future, you need to write these commands:

    git checkout 3.x
    git checkout -b fuzzy-completion

The first command selects the `3.x` branch in your fork, and the second one will create a new branch for your work (in this case called `fuzzy-completion`)

ccordoba12:
I'm going to close this one so you can remake @malliwi88 :-)

PR #3696, opened 2016-11-17
PR: Use QRegularExpression for find and replace (only for PyQt5)

### LINKED ISSUE #3408: Find and replace: Greedy regular expressions
Spyder version: 2.3.9
OS: Ubuntu 14
Python 2.7

Find and replace doesn't work for non greedy regular expressions. This regular expression:

```
        \{\$documento\['(.+?)'\]\}
```

doesnt recognize this text:

```
{$documento['DOC_CODIGO']}','{$documento['DOC_DOCID_EDP']}','{$documento['DOC_ESTADO']}','{$documento['DOC_BORRADO']}','{$documento['DOC_SAP_CODMATERIAL']}'
```

But this other regular expressiones does (without the question mark):

```
        \{\$documento\['(\w+)'\]\}
```

rlaverde:
Partially fixes #3408 (only PyQt5)

`QRegExp` didn't support lazy quantifiers, like *?, +?, etc. It's better to use `QRegularExpression`.

Although `QRegularExpression` is only avalaible since Qt5, so It'll continue using `QRegExp` for PyQT4

### DIFF
--- a/spyder/widgets/mixins.py
+++ b/spyder/widgets/mixins.py
@@ -22,11 +22,11 @@ import textwrap
 from qtpy.QtCore import QPoint, Qt
 from qtpy.QtGui import QCursor, QTextCursor, QTextDocument
 from qtpy.QtWidgets import QApplication, QToolTip
-from qtpy import PYQT5, PYQT4
+from qtpy import PYQT5
 
 if PYQT5:
     from qtpy.QtCore import QRegularExpression
-elif PYQT4:
+else:
     from qtpy.QtCore import QRegExp
 
 # Local imports
@@ -492,7 +492,7 @@ class BaseEditMixin(object):
             if case:
                 pattern.setPatternOptions(
                     QRegularExpression.CaseInsensitiveOption)
-        elif PYQT4:
+        else:
             pattern = QRegExp(r"\b{}\b".format(text)
                               if words else text, Qt.CaseSensitive if case else
                               Qt.CaseInsensitive, QRegExp.RegExp2)
--- a/spyder/widgets/mixins.py
+++ b/spyder/widgets/mixins.py
@@ -19,9 +19,15 @@ import sre_constants
 import textwrap
 
 # Third party imports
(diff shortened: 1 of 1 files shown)


ccordoba12:
Nice, thanks for the quick fix @rlaverde!

Could you add a test for this?

spyder/widgets/mixins.py
```diff
@@ -19,9 +19,15 @@
...
 # Third party imports
-from qtpy.QtCore import QPoint, QRegExp, Qt
+from qtpy.QtCore import QPoint, Qt
 from qtpy.QtGui import QCursor, QTextCursor, QTextDocument
 from qtpy.QtWidgets import QApplication, QToolTip
+from qtpy import PYQT5, PYQT4
+
+if PYQT5:
+    from qtpy.QtCore import QRegularExpression
+elif PYQT4:
```
ccordoba12: Please use `else` here

spyder/widgets/mixins.py
```diff
@@ -480,9 +486,17 @@ def find_text(self, text, changed=True, forward=True, case=False,
...
-        pattern = QRegExp(r"\b%s\b" % text if words else text,
-                          Qt.CaseSensitive if case else Qt.CaseInsensitive,
-                          QRegExp.RegExp2)
+        if PYQT5:
+            pattern = QRegularExpression(r"\b{}\b".format(text) if words else
+                                         text)
+            if case:
+                pattern.setPatternOptions(
+                    QRegularExpression.CaseInsensitiveOption)
+        elif PYQT4:
```
ccordoba12: Also use `else` here :-)

ccordoba12:
Supporting this for PyQt5 we will have to eventually deprecate PyQt4 because it's not maintained anymore.

ccordoba12:
@rlaverde, are you working on test for this one?

ccordoba12:
LGTM, merging. Thanks @rlaverde :-)

PR #3717, opened 2016-11-21
PR: Load kernel json files anywhere in the filesystem

### LINKED ISSUE #3630: Load external kernel json files from anywhere in the file system
Right now is only possible to connect to a external kernel if the .json connection file is placed in jupyter runtime-dir path.

mariacamilaremolinagutierrez:
Fixes #3630

### DIFF
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -1377,9 +1377,8 @@ class IPythonConsole(SpyderPluginWidget):
                                   password):
         # Verifying if the connection file exists
         try:
-            cf = osp.split(connection_file)
-            cf_path = cf[0]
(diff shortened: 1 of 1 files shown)


spyder/plugins/ipythonconsole.py
```diff
@@ -1377,7 +1377,11 @@ def _create_client_for_kernel(self, connection_file, hostname, sshkey,
                                   password):
         # Verifying if the connection file exists
         try:
-            connection_file = find_connection_file(osp.basename(connection_file))
+            cf = osp.split(connection_file)
+            cf_path = cf[0]
+            cf_filename = cf[1]
```
ccordoba12: Please use `osp.basename` for the filename and `osp.dirname` for its directory.

PR #3719, opened 2016-11-21
PR: Spyder wont crash if it can not acces jupyter runtime dir

### LINKED ISSUE #3706: Spyder fails to create kernel.json files if jupyter_runtime_dir is not writable
This error was reported here:

http://stackoverflow.com/q/40643760/438386

----

~~Solution: If `jupyter_runtime_dir` is not writable we need to switch to `TEMPDIR` (defined in `utils/programs`) to save and load kernel.json files.~~

@mariacamilaremolinagutierrez, I'm assigning this one to you because you also have issue #3630.

mariacamilaremolinagutierrez:
Fixes #3706


spyder/plugins/ipythonconsole.py
```diff
@@ -861,12 +861,21 @@ def create_new_client(self, give_focus=True):
         """Create a new client"""
         self.master_clients += 1
         name = "%d/A" % self.master_clients
+        try:
+            cf = self._new_connection_file()
+        except PermissionError:
+            QMessageBox.warning(self, _('Warning'),
+                _("The jupyter_runtime_dir directory is not writable and it <br>"
+                "is required to create IPython consoles. Please create the <br>"
+                "directory: %s or make it writable.") % jupyter_runtime_dir(),
```
goanpeca: Could we use format formatting :-) ?

`'some text {}'.format(value)`

mariacamilaremolinagutierrez: Yeah absolutely, let me change in a moment :)

ccordoba12:
@mariacamilaremolinagutierrez, glad to see you back on track again!

ccordoba12:
@mariacamilaremolinagutierrez, could you a rebase to remove the merge you introduced in your last commit? Thanks!

spyder/plugins/ipythonconsole.py
```diff
@@ -861,12 +861,19 @@ def create_new_client(self, give_focus=True):
         """Create a new client"""
         self.master_clients += 1
         name = "%d/A" % self.master_clients
+        cf = self._new_connection_file()
+        if cf is None:
+            QMessageBox.warning(self, _('Warning'),
```
ccordoba12: Please don't use `QMessageBox.warning` for this message, but instead `client.show_kernel_error`, as it's done some lines below. This has the advantage of showing the message inside IPython consoles, and so it won't be easy to dismiss by our users (something that usually happens with modal dialogs as `QMessageBox`).

Also, when you do this change, please remove the `<br>`'s you added at the end of every line. We don't need them in this case :-)

mariacamilaremolinagutierrez: @ccordoba12 I already fixed the <br>, however I can't use the ```client.show_kernel_error``` because in order to create the client the conecction file has to be not None. Is there another option to not use ```QMessageBox.warning```?

ccordoba12: > however I can't use the client.show_kernel_error because in order to create the client the conecction file has to be not None.

That's not right. `ClientWidget` can be instantiated without a connection file and also without a kernel.

So please move the validation of `cf is None` just after `add_tab` is called and use `client.show_kernel_error` to show our new error message.

spyder/plugins/ipythonconsole.py
```diff
@@ -877,6 +871,13 @@ def create_new_client(self, give_focus=True):
                               menu_actions=self.menu_actions)
         self.add_tab(client, name=client.get_name())
 
+        if cf is None:
+            error_msg = ("The directory {} is not writable and it is "
```
ccordoba12: Please add the underscore function (for translations) at the beginning of this string, like this

    error_msg = _("The directory {} ...

Else this string won't be picked up by gettext to be translated :-)

spyder/plugins/ipythonconsole.py
```diff
@@ -877,6 +871,13 @@ def create_new_client(self, give_focus=True):
                               menu_actions=self.menu_actions)
         self.add_tab(client, name=client.get_name())
 
+        if cf is None:
+            error_msg = ("The directory {} is not writable and it is "
+                         "required to create IPython consoles. Please make "
+                         "it writable.").format(jupyter_runtime_dir())
+            client.show_kernel_error(_(error_msg))
```
ccordoba12: After the previous change, this only needs to be

    client.show_kernel_error(error_msg)

spyder/plugins/ipythonconsole.py
```diff
@@ -1359,7 +1367,10 @@ def _new_connection_file(self):
         """
         # Check if jupyter_runtime_dir exists (Spyder addition)
         if not osp.isdir(jupyter_runtime_dir()):
-            os.makedirs(jupyter_runtime_dir())
+            try:
+                os.makedirs(jupyter_runtime_dir())
+            except PermissionError as pe:
```
ccordoba12: Last thing: there's no need of capturing the error (`as pe`) because we're not doing anything with it :-)

ccordoba12:
Ok, this is ready. Thanks a lot @mariacamilaremolinagutierrez!

By the way, doing a squash with merges is not an easy operation (cf. this [SO post](http://stackoverflow.com/questions/1725708/git-rebase-interactive-squash-merge-commits-together)) so let's leave things like they are :-)

PR #3726, opened 2016-11-22
Feature/get words

malliwi88:
The method for extract all words from source code for improving fallback_plugin.

@goanpeca


spyder/utils/introspection/utils.py
```diff
@@ -210,6 +210,15 @@ def get_keywords(lexer):
                 continue
     return keywords
 
+def get_words(text_from_file):
+    """ Completion inline
```
goanpeca: Please dont leave spaces before the docstrings

Use 

```python
"""
One liner description.

More info as needed.
"""
```

spyder/utils/introspection/utils.py
```diff
@@ -210,6 +210,15 @@ def get_keywords(lexer):
...
+    Extract the list of words that contains in text_from_file,
+    to carry out the inline completion similar to VSCode.
+    """
+    lines = [re.sub(r'([^a-zA-Z-_])', ' ', line, flags=re.UNICODE).split() \
+             for line in text_from_file.split("\n")]
```
goanpeca: As a general rule If you cannot fit a list comprehension in a single line, then you should not use one.

Just use a normal for that is more readable.

Also using `   \   ` to join code lines is discouraged as a style guide.

spyder/utils/introspection/utils.py
```diff
@@ -210,6 +210,15 @@ def get_keywords(lexer):
...
+    words = list(set([x for words in lines for x in words if x != []]))
```
goanpeca: Why not use a generator? It will make it more performant as the list does not need to be generated as an intermediate step

`words = list(set(x for words in lines for x in words if x != []))`

spyder/utils/introspection/utils.py
malliwi88: Ok

malliwi88:
Ok @goanpeca all test are designed.

Only, I wait for finish test or additional suggestions.

spyder/utils/introspection/utils.py
```diff
@@ -211,6 +211,30 @@ def get_keywords(lexer):
     return keywords
 
 
+def get_words_file(file_in):
+    """
+    Completion inline
```
goanpeca: Docstrings are sentences and need to end in a period. Also we can improve the one line description from `Completion inline` to something like `Extract all words from a source code file to be used in code completion.`

spyder/utils/introspection/utils.py
```diff
@@ -211,6 +211,30 @@ def get_keywords(lexer):
     return keywords
 
 
+def get_words_file(file_in):
```
goanpeca: This methods will be called most likely using the content (that is already available for an editor) and not the file name, so the signature should be more like

```python
def get_words_file(file_path=None, content=None):
    if file_path is None and content is None or file_path and content:
        error_msg = _('Must provide one of `file_path` or `content`')
```

spyder/utils/introspection/utils.py
```diff
@@ -211,6 +211,30 @@ def get_keywords(lexer):
...
+    ext = os.path.splitext(file_in)[1]
+    if ext in ['.py', '.css', 'md']:
+        print("ext", ext[1])
+        regex = re.compile(r'([^a-zA-Z-_])')
+    elif ext in ['.R', '.c', '.cpp, java']:
+        print("ext", ext[1])
+        regex = re.compile(r'([^a-zA-Z_])')
+    else:
+        regex = re.compile(r'([^a-zA-Z])')
+        print("ext", ext[1])
```
goanpeca: All these prints should not be here

spyder/utils/introspection/utils.py
goanpeca: This means the logic you are using for getting the extension, needs to be an extra parameter as well.

spyder/utils/introspection/utils.py
```diff
@@ -211,6 +211,30 @@ def get_keywords(lexer):
...
+
+    with open(file_in, 'r') as infile:
```
goanpeca: Fix to use content directly or file_path

spyder/utils/introspection/utils.py
```diff
@@ -211,26 +211,25 @@ def get_keywords(lexer):
     return keywords
 
 
-def get_words_file(file_in):
+def get_words_file(file_path=None, content=None):
```
goanpeca: I think we will need an extra keyword param called extension, otherwise how can we tell the extension if only content was provided?

spyder/utils/tests/test_get_words.py
```diff
@@ -0,0 +1,117 @@
...
+            'World', 'lang', 'html', 'DOCTYPE', 'utf']
+    assert sorted(words) == sorted(real)
+
+def test_get_words_python():
+    """Test for get word from html file syntax"""
+
+    path = dirname(dirname(abspath(__file__)))
+    f_in = path+"/tests/data/example.py"
+    words = get_words_file(f_in)
+    real = ['a', 'sure', 'np', 'Z_', 'Simple', 'R', 'utn', 'values',
```
goanpeca: expected_words maybe is a better name than real

spyder/utils/introspection/utils.py
goanpeca: :+1: cool keep up the good work :-)

ccordoba12:
@malliwi88, please also merge this one with 3.x to solve errors in Travis :-)

ccordoba12:
@malliwi88, please address @goanpeca's comments and merge with 3.x so we can merge this one.

PR #3727, opened 2016-11-22
PR: Remove SpyderPluginMixin and improve external API

ccordoba12:
~~**Important note**: This change breaks compatibility between external plugins developed for Spyder 3.0 and 4.0 (but I think it's for the better).~~

**Important note**: Compatibility with Spyder 3.0 is maintained as long as plugins subclass from `SpyderPluginWidget`.

----

Due to differences in initialization of mixins between PyQt4 and PyQt5:

http://pyqt.sourceforge.net/Docs/PyQt5/pyqt4_differences.html#cooperative-multi-inheritance

I preferred to remove `SpyderPluginMixin` and convert it into `BasePluginWidget` and `PluginWidget`.

I think `SpyderPluginMixin` was a trick introduced by Pierre to more easily develop plugins with a single widget. However, by removing that mixing and creating all plugin widgets as attributes, we can have a way cleaner and simpler API for all plugins (the previous API was a mess!).

Now people that read Spyder source code will understand our design model: widgets are isolated, they are embedded in plugins, and if they need to communicate info to other plugins, they must emit Qt signals which are captured and routed by its corresponding plugin.

----

TODO:

- [x] Define a more basic class from which `PluginWidget` will inherit, This class will contain the methods we don't want users to know about in `spyder.api` (like `create_toggle_view_action`).


jitseniesen:
Is it possible to have a plugin without a widget? I think this could be useful for things like adding a menu item, or support for a programming language, or support for some types in the variable explorer.

ccordoba12:
> Is it possible to have a plugin without a widget?

Not right now, but that's the next step. This just cleans our internal API for plugins with widgets and put it in decent shape :-)

After merging this one (which I think is ready), I'll open PRs in all our third-party plugin repos to make them subclass from `SpyderPluginWidget` instead of `SpyderPluginMixin`.

spyder/api/plugins.py
```diff
@@ -8,252 +8,51 @@
...
-        self.show_nontab_menu(event)
+    Public interface for Spyder plugins.
 
-    def show_nontab_menu(self, event):
-        """Show the context menu assigned to nontabs section."""
-        menu = self.main.createPopupMenu()
-        menu.exec_(self.dock_tabbar.mapToGlobal(event.pos()))
-
-
-class SpyderDockWidget(QDockWidget):
```
goanpeca: @ccordoba12 so now tabs of docked widgets cannot be moved? or where did this logic go?

ccordoba12: Don't worry, I moved this code to `widgets/dock.py` because I don't think we need to expose it as part of our API :-)

ccordoba12:
Ok, this is going in. Letting know @Nodd and @jitseniesen about it.

PR #3735, opened 2016-11-24
Showing time units on the profiler

malliwi88:
Maybe it is necessary to do a little cleaning to the profiler, I found a lot of code not functional inside profiler.

The final result looks like this
spyder 3 1 0 dev0 python 3 5 _001

### DIFF
--- a/spyder_profiler/widgets/profilergui.py
+++ b/spyder_profiler/widgets/profilergui.py
@@ -510,12 +510,53 @@ class ProfilerDataTree(QTreeWidget):
                 color = "red"
         return [format[0] % x[0], [diff_str, color]]
 
+    def format_measure(self, data):
 
-    def format_output(self,child_key):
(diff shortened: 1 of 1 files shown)


ccordoba12:
What's the difference between *Total time* and *Local time*?

ccordoba12:
@malliwi88, please merge with master to fix the errors in Travis. For that, you need to first checkout your branch, then run

    git merge master

and commit your changes.

ccordoba12:
Sorry, the merge is against the 3.x branch, so the command is

    git merge 3.x

Nodd:
@ccordobat Local time is roughly equivalent to the total time of the function minus the total time of all the subfunctions. It represents the time actually spent in the function and the builtins.

ccordoba12:
Thanks for the explanation @Nodd :-)

PR #3749, opened 2016-11-26
PR: Fix View/Toolbars menu

### LINKED ISSUE #3746: Menu view/toolbar has empty contents
on my installs with spyder 3.0.1 and 3.0.2 the menuoption for toolbars is not showing.

the installs are fresh anaconda installs on windows 64-bit, and then an upgrade of spyder.
I think it worked with the 3.0.0, since I did change the toolbars on one computer,
but that is no longer possible.

did not directly find this mentioned anywhere - if it is a duplicate just close.

dalthviz:
Fixes #3746

### DIFF
--- a/spyder/utils/qthelpers.py
+++ b/spyder/utils/qthelpers.py
@@ -303,6 +303,11 @@ def add_actions(target, actions, insert_before=None):
                 target.addMenu(action)
             else:
                 target.insertMenu(insert_before, action)
+        elif isinstance(action, QAction):
+            if insert_before is None:
+                target.addAction(action)
+            else:
+                target.insertAction(insert_before, action)
         elif isinstance(action, SpyderAction):
             if isinstance(target, QMenu) or not isinstance(target, QToolBar):
                 action = action.no_icon_action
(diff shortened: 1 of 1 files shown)


dalthviz:
The problem was a missing validation to add QAction elements in the method add_actions in `spyder/utils/qthelpers.py`.

spyder/utils/qthelpers.py
```diff
@@ -303,6 +303,11 @@ def add_actions(target, actions, insert_before=None):
...
             else:
                 target.insertMenu(insert_before, action)
+        elif isinstance(action, QAction):
+            if insert_before is None:
+                target.addAction(action)
+            else:
+                target.insertAction(insert_before, action)
         elif isinstance(action, SpyderAction):
             if isinstance(target, QMenu) or not isinstance(target, QToolBar):
                 action = action.no_icon_action
```
ccordoba12: This can be made simpler by changing this block and removing the code you introduced above, like this:

```python
elif isinstance(action, QAction):
    if isinstance(action, SpyderAction):
        if isinstance(target, QMenu) or not isinstance(target, QToolBar):
            action = action.no_icon_action
```

ccordoba12:
Thanks @dalthviz for looking into this one. I added a comment simplify our code, but other than that, this is ready :-)

ccordoba12:
Thanks @dalthviz!

PR #3754, opened 2016-11-28
PR: Added validation to window/position config value saved in spyder.ini

### LINKED ISSUE #3748: Bad window size in spyder.ini prevents Spyder from showing

1. enable dual monitors. move spyder to the right monitor. exit spyder.
2. disable the right monitor. (or connect over remote desktop)
3. load spyder.
- _alternatively, just edit your .spyder-py3/spyder.ini and `set window/position = (4134, 197)`_

I expect to see spyder, but the window cannot be seen anywhere on the screen.

## Poor fix:
`I could spyder --reset`, but that loses lots of settings I spent a long time to set up.

## Best FIX (a developer should do this):
The last used window position is stored in my user folder /.spyder-py3/spyder.ini
Set the window position to a sane x,y pair like (100,100):
`window/position = (100, 100)`
It would be great if spyder could detect the width of the desktop, and if the window position X is greater than this width it decreases the number to compensate.

* Spyder Version: 3.0.2
* Python Version: 3.5.2

dalthviz:
Fixes #3748

### DIFF
--- a/spyder/config/user.py
+++ b/spyder/config/user.py
@@ -19,6 +19,7 @@ import re
 import os.path as osp
 import shutil
 import time
+import sys
 
 # Local imports
 from spyder.config.base import (get_conf_path, get_home_dir,
@@ -26,6 +27,7 @@ from spyder.config.base import (get_conf_path, get_home_dir,
 from spyder.utils.programs import check_version
 from spyder.py3compat import configparser as cp
 from spyder.py3compat import PY2, is_text_string, to_text_string
+from qtpy.QtWidgets import QApplication
 
 # Std imports for Python 2
 if PY2:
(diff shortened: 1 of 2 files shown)


spyder/config/user.py
```diff
@@ -397,6 +398,17 @@ def get(self, section, option, default=NoDefault):
...
+        # It's necessary to verify if the window/position value is valid 
+        # with the current screen. See issue 3748
+        if option == 'window/position':
+            width = value[0]
+            height = value[1]
+            screenShape = QApplication.desktop().geometry()
+            current_width = screenShape.width()
+            current_height = screenShape.height()
+            if current_width < width or current_height < height:
+                value = self.get_default(section, option)
```
ccordoba12: Although this is the right thing to do, this is not the right place. You need where this setting is applied (most probably in `app/mainwindow.py`) and use this validation there.

ccordoba12:
Other than a minor comment, this looks good to me. Thanks @dalthviz!

ccordoba12:
Good job, thanks @dalthviz!

PR #3760, opened 2016-11-28
PR: Define tabulation width in term of spaces instead of pixels

### LINKED ISSUE #3729: Define tabulation width in term of spaces instead of pixels
In PR #3663 I assume that tabulation width will be always 4 spaces, and indentation could look wrong depending of `tab_stop_width` value,  I think this could be improved defining it in spaces instead of pixels

I suggest:

- change:   
/config/main.py
`
'tab_stop_width': 40,
`
to
`
'tab_stop_width_spaces': 4,`

- use tab_stop_width_spaces in `fix_indent` calculations

- and set tab stop using space width
`
self.setTabStopWidth('tab_stop_width_spaces * self.editor.fontMetrics().width(' '))
`

rlaverde:
Fixes #3729

### DIFF
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -629,7 +629,7 @@ DEFAULTS = [
 #    or if you want to *rename* options, then you need to do a MAJOR update in
 #    version, e.g. from 3.0.0 to 4.0.0
 # 3. You don't need to touch this value if you're just adding a new option
-CONF_VERSION = '30.2.0'
+CONF_VERSION = '30.0.0'
 
 # Main configuration instance
 try:
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -214,7 +214,7 @@ DEFAULTS = [
               'add_colons': True,
               'auto_unindent': True,
               'indent_chars': '*    *',
-              'tab_stop_width': 40,
+              'tab_stop_width_spaces': 4,
               'codecompletion/auto': True,
               'codecompletion/enter_key': True,
               'codecompletion/case_sensitive': True,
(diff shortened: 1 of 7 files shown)


spyder/config/main.py
```diff
@@ -629,7 +629,7 @@
 #    or if you want to *rename* options, then you need to do a MAJOR update in
 #    version, e.g. from 3.0.0 to 4.0.0
 # 3. You don't need to touch this value if you're just adding a new option
-CONF_VERSION = '29.2.0'
+CONF_VERSION = '30.2.0'
```
ccordoba12: This should be `30.0.0`

ccordoba12:
Please add a test for this (even if very simple :-)

spyder/widgets/sourcecode/base.py
```diff
@@ -277,6 +278,11 @@ def setup_completion(self):
     def set_indent_chars(self, indent_chars):
         self.indent_chars = indent_chars
 
+    def set_tab_stop_width_spaces(self, tab_stop_width_spaces):
+        self.tab_stop_width_spaces = tab_stop_width_spaces
+        self.setTabStopWidth(tab_stop_width_spaces
+                             * self.fontMetrics().width('9'))
```
ccordoba12: What does this do? This `9` here seems pretty specific ;-)

ccordoba12:
This looks very good to me. I just have a minor doubt about the implementation.

spyder/widgets/sourcecode/base.py
rlaverde: I think other character could work, `self.fontMetrics().width('9')` returns the width of a character in the editor

I use '9' because It is used in other places with the same purpose:
https://github.com/spyder-ide/spyder/blob/master/spyder/widgets/panels/edgeline.py#L41
https://github.com/spyder-ide/spyder/blob/master/spyder/widgets/panels/line_number.py#L171

Actually this magic it's pretty old https://github.com/spyder-ide/spyder/commit/f7c68f33574f2099f6cdab93762befda4d20bf7c#diff-a032958c9839f940f8038a099a3bf970R1072

Issue #3772, opened 2016-12-02
Configparser error leads to crash of Spyder 3 under Python(x,y)


ccordoba12:
See this post of Stackoverflow about it:

http://stackoverflow.com/questions/40913775/file-parser-leads-to-crash-of-spyder-3-under-pythonx-y

----

@dalthviz, this is an easy one for you. To solve it, please remove all calls and references to the `xy` module in `app/mainwindow.py`.

PR #3774, opened 2016-12-02
PR: Delete references to the 'xy' module.

dalthviz:
Fixes: #3772


ccordoba12:
There's also a reference to the `xy` module in `create_module_bookmark_actions`, from `utils/qthelpers.py`. Please remove it from there too.

ccordoba12:
I'm sorry to do this but PythonXY development seems stalled now and (unfortunately) people still wants to use it along with Spyder.

So we can't rely on its configuration module (`xy`) in our side.

PR #3654, opened 2016-11-05
PR: Add regex to highlight decorators

rlaverde:
Fixes #3608

---

I fixed it in a really simple way, I reused instance decorator, is that ok? or should I add a new color theme color for decorators?

PR #3776, opened 2016-12-02
Improve highlighter decorator regexp

rlaverde:
In  #3654 I didn't take in account that decorators are variables separated by points, and variables contain letters, numbers and underscores, and should start by a letter or underscore

Some decorators weren't detected:
```
@decorator_with_underscore
@decorator12
```
and others were detected and shouldn't be detected:
```
@.decorator
@decorator..points
@decorator.
```


ccordoba12:
Thanks @rlaverde!

PR #3780, opened 2016-12-03
PR: Only show base filename in Python console titles

### LINKED ISSUE #1165: Console tabs are too big because they display the full filename
_From [andreasn...@googlemail.com](https://code.google.com/u/106330885117898830116/) on 2012-10-04T15:20:39Z_

The console tabs of running or run programs are quite big, because they display the complete filename including path, e.g.:

"Console - /home/user/python/myproject/util/module.py"

On some Spyder window setups, this big tab uses much space and thus leaves very little space for the other views tabs, I have put in the same pane. These other tabs then often show only one letter and it is somehow unpleasing to the eye.

Not so long ago, only the filename ("Console - module.py") was displayed in these tabs. Can we have an option for choosing between long and short filenames in the console tabs?

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=1165_

malliwi88:
Fixes #1165 

----

Okay, we're talking about this, right?

seleccion_012

This is the result of PR
seleccion_013

Thanks @ccordoba12

### DIFF
--- a/spyder/plugins/externalconsole.py
+++ b/spyder/plugins/externalconsole.py
@@ -712,8 +712,8 @@ class ExternalConsole(SpyderPluginWidget):
             fname = self.filenames[index]
             if fname:
                 # solve issue 1165
-                filename = osp.basename(fname)
-                title += ' - ' + to_text_string(filename)
+                fname = osp.basename(fname)
+                title += ' - ' + to_text_string(fname)
         return title
     
     def get_plugin_icon(self):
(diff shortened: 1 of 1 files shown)


spyder/plugins/externalconsole.py
```diff
@@ -711,7 +711,9 @@ def get_plugin_title(self):
             index = self.tabwidget.currentIndex()
             fname = self.filenames[index]
             if fname:
-                title += ' - ' + to_text_string(fname)
+                # solve issue 1165
+                filename = osp.basename(fname)
```
ccordoba12: Please change this to

    fname = osp.basename(fname)

malliwi88:
Ready for merge @ccordoba12?

ccordoba12:
This is not the right thing to do, sorry @malliwi88. We need to disambiguate tabs if users evaluate different files with the same name.

PR #3791, opened 2016-12-06
PR: Catch error raised in isbinaryornot when trying to open a directory

rlaverde:
After #3640 an error was caused when right clicking over a directory in project explorer (I didn't notice while doing #3640, sorry :disappointed:)

```
>>> Traceback (most recent call last):
  File "/home/rlaverde/code/spyder/spyder/widgets/explorer.py", line 404, in contextMenuEvent
    self.update_menu()
  File "/home/rlaverde/code/spyder/spyder/widgets/explorer.py", line 383, in update_menu
    add_actions(self.menu, self.create_context_menu_actions())
  File "/home/rlaverde/code/spyder/spyder/widgets/explorer.py", line 370, in create_context_menu_actions
    actions += self.create_file_manage_actions(fnames)
  File "/home/rlaverde/code/spyder/spyder/widgets/explorer.py", line 270, in create_file_manage_actions
    only_valid = all([encoding.is_text_file(_fn) for _fn in fnames])
  File "/home/rlaverde/code/spyder/spyder/widgets/explorer.py", line 270, in <listcomp>
    only_valid = all([encoding.is_text_file(_fn) for _fn in fnames])
  File "/home/rlaverde/code/spyder/spyder/utils/encoding.py", line 252, in is_text_file
    return not is_binary(filename)
  File "/home/rlaverde/code/spyder/spyder/utils/external/binaryornot/check.py", line 32, in is_binary
    chunk = get_starting_chunk(filename)
  File "/home/rlaverde/code/spyder/spyder/utils/external/binaryornot/helpers.py", line 32, in get_starting_chunk
    with open(filename, 'rb') as f:
IsADirectoryError: [Errno 21] Is a directory: '/home/rlaverde/code/spyder/spyder/defaults'
```

### DIFF
--- a/spyder/utils/encoding.py
+++ b/spyder/utils/encoding.py
@@ -249,4 +249,7 @@ def is_text_file(filename):
     """
     Test if the given path is a text-like file.
     """
-    return not is_binary(filename)
+    try:
+        return not is_binary(filename)
+    except IsADirectoryError:
+        return False
--- a/spyder/utils/encoding.py
+++ b/spyder/utils/encoding.py
@@ -251,5 +251,5 @@ def is_text_file(filename):
     """
     try:
         return not is_binary(filename)
-    except IsADirectoryError:
+    except (OSError, IOError):
         return False


spyder/utils/encoding.py
```diff
@@ -249,4 +249,7 @@ def is_text_file(filename):
     """
     Test if the given path is a text-like file.
     """
-    return not is_binary(filename)
+    try:
+        return not is_binary(filename)
+    except IsADirectoryError:
```
ccordoba12: This is not a builtin Python error (at least not on Linux). Maybe it's an error defined in `binaryornot`

ccordoba12: It doesn't exist in Python 2 :-/

rlaverde: In Python2 `IOError` will be raised https://docs.python.org/2.7/library/exceptions.html#exceptions.IOError

I don't know how is the best way to deal with different exceptions and python2 and python3

should I use `PY2`?:
```
if PY2:
    try:
        return not is_binary(filename)
    except IOError:
        return False
else
    try:
        return not is_binary(filename)
    except IsADirectoryError:
        return False
```

or just catch any exception:
```
    try:
        return not is_binary(filename)
    except:
        return False
```

goanpeca:
@rlaverde is Six dealing with this in some way already?

spyder/utils/encoding.py
ccordoba12: Let's go for your first version for now :-)

rlaverde:
@goanpeca I think six doesn't deal with different exceptions:
I have found this https://bitbucket.org/gutworth/six/issues/78/new-built-in-exceptions

Maybe using `six.moves,` could be and workaround for this

spyder/utils/encoding.py
rlaverde: Another way could be to catch `IsADirectoryError` parent Exception:

```
try:
    return not is_binary(filename)
except (IOError, OSError):
    return False
```

ccordoba12: Ok, this looks better because it makes the code more compact :-)

ccordoba12:
@rlaverde, please merge with 3.x to pick up our configuration settings for CircleCI

ccordoba12:
Thanks @rlaverde!

PR #3795, opened 2016-12-07
PR: Use a menu separator constant instead of None

### LINKED ISSUE #3794: Create menu separator instead of None
Use a menu separator so that it is clear that None is used to separate the menu options. 

image

mariacamilaremolinagutierrez:
Fixes #3794


spyder/utils/qthelpers.py
```diff
@@ -499,6 +499,10 @@ def show_std_icons():
     dialog.show()
     sys.exit(app.exec_())
 
+    
+def get_menu_separator():
```
ccordoba12: Let's call this `menu_separator`.

spyder/plugins/editor.py
```diff
@@ -973,25 +973,27 @@ def get_plugin_actions(self):
         # ---- File menu/toolbar construction ----
         self.recent_file_menu = QMenu(_("Open &recent"), self)
         self.recent_file_menu.aboutToShow.connect(self.update_recent_file_menu)
-
+        
+        menu_separator = get_menu_separator()
```
ccordoba12: Please remove this assigment

ccordoba12:
Thanks for working on this @mariacamilaremolinagutierrez :-)

However, there are many other places where None is used as menu separator (i.e. not just in the Editor).

Please find them all and do the same replacement you already did for the Editor.

ccordoba12:
There are a lot of style failures. Please review CircleCI to see how to fix them :-)

mariacamilaremolinagutierrez:
@ccordoba12 Wow there are, nice work on CI, it is super useful. I am making several commits but I'll squash the ones regarding style and some merges :)

Also, I'm not sure about this one:

```E402: module level import not at top of file```

Because that would mean changing the header structure:

image

Should I import it up?

**Comment by ccordoba12** [comment]
Please don't change the import order in `mainwindow`. I carefully optimized it to start Spyder as fast as it's possible.

However, ciocheck is still reporting errors in `plugins/editor.py` and `plugins/findinfiles.py`

**Comment by mariacamilaremolinagutierrez** [comment]
@ccordoba12 do you know how to re run tests without making another commit? I'm getting a _HTTPError: 500 Server Error_ in the CI failure.

**Comment by ccordoba12** [comment]
I restarted the build. There are still some failures (besides the import order :-)

**Comment by mariacamilaremolinagutierrez** [comment]
@ccordoba12 I don't quite understand the last error I have left. How should I indent it?

image

image

**Comment by goanpeca** [comment]
@mariacamilaremolinagutierrez using `\` to concatenate lines in Python is super ugly and I would strongly advice against using them. In this particular casse you could concatenate stuff like this:

```python
a = []
b = []
c = []
new = (a +
       b +
       ['something'] +
       c +
       ['something'])
```

Parenthesis used like this serve the same purpose and make the code much cleaner.

Also PEP8 requires that the operators remain at the end of the line instead of at the beginning, so

```python
# PEP8 ok
a = (5123123123123123123132134123123123132 +
     6)

# PEP8 fails
a = (5123123123123123123132134123123123132
     + 6)
```

Although that rule is not perfect to be honest...

mariacamilaremolinagutierrez:
@goanpeca Awesome! Thank you very much :)

@ccordoba12 The CI test that's failing is only the import one I mentioned earlier. And the style commits were squashed in one.

spyder/utils/qthelpers.py
```diff
@@ -499,6 +499,10 @@ def show_std_icons():
     dialog.show()
     sys.exit(app.exec_())
 
+    
+def menu_separator():
```
ccordoba12: Let's make this one a constant instead of a function to use it more easily, like this

menu_separator = None

ccordoba12:
Great job at fixing the style issues!

I left a last comment, then this would be ready :-)

spyder/utils/qthelpers.py
goanpeca: then make it upper case!

ccordoba12: Ok, sure, no problem with that :-)

rlaverde:
> Also PEP8 requires that the operators remain at the end of the line instead of at the beginning

@goanpeca Both ways are accepted, and operators at the beginning Knuth's style are preferred
https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

ccordoba12:
Thanks @mariacamilaremolinagutierrez, this looks good to me now :-)

Issue #2512, opened 2015-07-02
Index column of Pandas DataFrame should be visible all the time


scls19fr:
Hello,

Index of Pandas DataFrame should be in a frozen pane in variable explorer

![capture d ecran 2015-07-02 a 22 36 20](https://cloud.githubusercontent.com/assets/109167/8487256/20e02cec-210b-11e5-853e-b1cc924f66de.png)
![capture d ecran 2015-07-02 a 22 36 30](https://cloud.githubusercontent.com/assets/109167/8487259/226be470-210b-11e5-8ad5-8cedc575288a.png)

Kind regards

ccordoba12:
I can take a look at this one. Don't know if it's possible in Qt though :-)

bilderbuchi:
also, why are the columns so unnecessarily wide? seems to waste an awful lot of space :-/

ccordoba12:
@andfoy, please work on this one. The idea is to make the first column of our DataFrame viewer fixed when people scroll to left and right.

The file you need to work on is `spyder/widgets/variableexplorer/dataframeeditor.py`.

And here you can find how to do it in Qt:

http://doc.qt.io/qt-5/qtwidgets-itemviews-frozencolumn-example.html

scls19fr:
Thanks

ccordoba12:
You're welcome! Thanks a lot for this great suggestion :-)

PR #3810, opened 2016-12-09
PR: Index column of DataFrame view should be always visible (Issue #2512) 

andfoy:
This fixes Issue #2512 : Index column is now freezed


ccordoba12:
@andfoy, why did you close this PR?

andfoy:
@ccordoba12 I forgot to replace an ``xrange`` with a ``range``, which caused errors when Spyder was executed on Python 3

PR #3812, opened 2016-12-09
PR: DataFrame viewer now displays tooltip for column headers

### LINKED ISSUE #3721: DataFrame viewer should display tooltips for truncated column headers
When a DataFrame with long column names is viewed, the column names are truncated. It would be useful to have the full column names displayed as tooltips when pointing at them.

I my case, I have got many columns with long names, so I want to avoid pressing the Resize button to keep the data compact.

(Seen in WinPython-64bit-3.5.2.3 with "Spyder 3.1.0.dev0".)

camendoza94:
Fixes #3721

----

second issue

### DIFF
--- a/spyder/widgets/variableexplorer/dataframeeditor.py
+++ b/spyder/widgets/variableexplorer/dataframeeditor.py
@@ -195,7 +195,8 @@ class DataFrameModel(QAbstractTableModel):
                         header = to_text_string(self.df_header[0])
                     return to_qvariant(header)
                 else:
-                    return to_qvariant(to_text_string(self.df_header[section-1]))
+                    header = to_text_string(self.df_header[section - 1])
+                    return to_qvariant(header)
             else:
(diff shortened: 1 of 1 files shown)


ccordoba12:
Great work @camendoza94! Don't worry about the tests failure. That's something we need to fix in our side :-)

ccordoba12:
@jitseniesen, could you review this one?

jitseniesen:
Is it possible to display the tooltip only if the column name is so long that it does not fit in the column header? It seems unnecessary to display the tooltip if the user can already see the whole name.

In the previous version, `headerData()` always returned something, but in your version, it does not return anything (technically, it returns `None`) if `role` is neither `DisplayRole` nor `TooltipRole`. Why did you change this?

Prikers:
I think this can be closed  as it is already covered as part of @dalthviz work in #3873
image

PR #3813, opened 2016-12-09
PR: Fixed encoding issue when changing cwd in the IPython console

### LINKED ISSUE #3807: Directory with non-ascii characters throws an error when it's made the working directory on the IPython console
Changing the working directory to one named with non-ascii characters throws an error in the internal console.

1. Change the working directory to one with non-ascii charaters in his name. 

Change of the directory without error. Instead:

`>>> Traceback (most recent call last):
  File "c:\Users\Daniel\Desktop\Spyder\spyder\spyder\plugins\ipythonconsole.py", line 845, in set_current_client_working_directory
    shellwidget.set_cwd(directory)
  File "c:\Users\Daniel\Desktop\Spyder\spyder\spyder\widgets\ipythonconsole\shell.py", line 86, in set_cwd
    "get_ipython().kernel.set_cwd(r'{}')".format(dirname))
UnicodeEncodeError: 'ascii' codec can't encode characters in position 24-27: ordinal not in range(128)`

* Spyder Version: 3.0.2
* Python Version: 2.7.12
* Operating system: Windows/Linux

dalthviz:
Fixes #3807


spyder/utils/debug.py
```diff
@@ -33,7 +33,7 @@ def log_last_error(fname, context=None):
         print("Context", file=fd)
         print("-------", file=fd)
         print("", file=fd)
-        print(context, file=fd)
+        print(u' '.join(context).encode('utf-8').strip(), file=fd)
```
ccordoba12: Why did you have to do this change?

dalthviz: This was needed because an error was thrown in the python console when changing the directory to one with non - ascii name. Actually was really strange when I saw it. I also saw that the error causes that a process in the python console crashes and then, if you try to change again the working directory, Spyder crashes too.

ccordoba12: Does the crash only occurs on Python 2? Or on Python 3 too?

dalthviz: It only occurs with Python 2.

ccordoba12: Then please apply this change only for Python 2 and leave the previous line for Python 3

ccordoba12:
Thanks @dalthviz, this looks good now :-)

PR #3824, opened 2016-12-10
PR: Create a separate window when undocking plugins

### LINKED ISSUE #3563: Docks confusion when pressing undocking button
Some feedback from a lecture @jitseniesen gave recently (October 2016)

> Somebody complained that the dock window interface is confusing; the undock button looks like it would maximize, and he got confused with the undocked window because it does act like a normal window does (this is on Linux)

### LINKED ISSUE #3790: Create new windows when undocking all plugins
There are two improvements I'd like to see for new Editor windows

1. Make their toolbar to be exactly the same as the one present in the main window (except for the working directory toolbar, which we don't need to show in new windows).
2. When the Editor plugin is undocked, we must close it and show a new window instead. For that we need to connect a new method to the topLevelChanged signal of its QDockWidget:

    http://doc.qt.io/qt-4.8/qdockwidget.html#topLevelChanged

@dalthviz, please work on this in two different PRs. Please start with point 1. which is easier, and then move to 2.

dalthviz:
Fixes #3790 
Fixes #3563


ccordoba12:
Please update with the latest changes in the 3.x branch to fix the errors in CircleCI

spyder/app/mainwindow.py
```diff
@@ -2209,6 +2210,15 @@ def __update_fullscreen_action(self):
         self.fullscreen_action.setIcon(icon)
 
     @Slot()
+    def undock_editor(self):
+        """Open a new window instance of the Editor instead of undoking it."""
```
ccordoba12: `undoking` is misspelled here. It's `undocking`

ccordoba12:
There's one thing that worries me about this change. What happens when users press `Ctrl+Shift+E` to move to the Editor?

I think that makes the Editor pane to appear in the main window, right?

dalthviz:
If you are with the focus in the main window, yes it makes appear the Editor pane. If you, in the other hand, have the focus in a new window it doesn't appear.

ccordoba12:
> If you are with the focus in the main window, yes it makes appear the Editor pane

Then we need to fix this, and make `Ctrl+Shift+E` to move to the new window, if the pane is hidden and an Editor window is visible.

We also need to prevent to more than window for this to make sense :-)

ccordoba12:
Sorry @dalthviz for my big delay on reviewing this one. I hope to do it after we release 3.1 :-)

goanpeca:
@dalthviz please rebase or merge with the latest 3,x to fix conflicts

spyder/app/mainwindow.py
```diff
@@ -2240,6 +2241,17 @@ def __update_fullscreen_action(self):
...
     @Slot()
+    def undock_editor(self):
+        """Open a new window instance of the Editor instead of undocking it."""
+        if self.editor.dockwidget.isFloating() and not self.editor.undocked:
+            self.editor.dockwidget.setVisible(False)
+            self.editor.create_new_window()
+            self.editor.toggle_view_action.setChecked(False)
+            self.editor.dockwidget.setFloating(False)
+        self.editor.undocked = False
+        self.editor.get_current_editorstack().new_window = False
```
ccordoba12: @dalthviz, does this method need to be here in the main window? As far as I can see, it doesn't uses any attribute/method of the window, so I think we can safely move it to the the Editor plugin file.

spyder/app/mainwindow.py
```diff
@@ -799,6 +799,7 @@ def create_edit_action(text, tr_text, icon):
         from spyder.plugins.editor import Editor
         self.editor = Editor(self)
         self.editor.register_plugin()
+        self.editor.dockwidget.topLevelChanged.connect(self.undock_editor)
```
ccordoba12: This line can also be moved to the Editor plugin. We can make it the last one of its `register_plugin` method.

ccordoba12:
@dalthviz, please merge with 3.x and finish this one for Spyder 3.2. I think it's almost ready ;-)

ccordoba12:
@dalthviz, please create an animated gif so that @goanpeca and me could see how this is working.

dalthviz:
Preview:

undock

ccordoba12:
@goanpeca, what do you think? There is an action in the Editor options menu to undock it so users can move next to other plugins.

ccordoba12:
Problem is this is not the same behavior for the rest of plugins.

ccordoba12:
@dalthviz, this needs another merge with `3.x`.

goanpeca:
@ccordoba12 yeah, this change would be a bit inconsistent with the rest and is better suited for 4.x

ccordoba12:
Ok, you're right. @dalthviz, please rebase this PR in top of master and keep working on it, so that when users press the undock button they see a new window for *every* plugin, and not just the Editor.

This functionality could go to `plugins/base.py` in master.

dalthviz:
@ccordoba12 @goanpeca When working in this PR I found that some plugins (atleast 8) doesn't have a toolbutton to add the undock action. What should we do?

ccordoba12:
Our decision is that (for consistency) all plugins must have an Options menu, that should contain at least the undocking entry.

spyder/api/plugins.py
```diff
@@ -311,3 +312,17 @@ def check_compatibility(self):
         message = ''
         valid = True
         return valid, message
+
+class PluginMainWindow(QMainWindow):
```
ccordoba12: Please move this to `plugins/base.py`

ccordoba12:

dalthviz:
@ccordoba12 @rlaverde a preview:

undocking

Also, while doing the preview I realized that we should need to verify if the third party plugins need some ajustment to fit with the changes in here. As an example the `unit-test` plugin:

imagen

Should I create and resolve in each repository an issue related with this?

ccordoba12:
> Should I create and resolve in each repository an issue related with this?

Yep, please open an issue in all our plugins that require a fix.

ccordoba12:
By the way, great @dalthviz! This is looking really, really nice!

rlaverde:
I just tested locally and It looks great, maybe the only think that I don't like is when undocking plugins are somehow small and square; maybe they should be rectangular and a little bigger, or use the dock dimensions.

ccordoba12:
> the only think that I don't like is that when undocking plugins are somehow small and square

Could you post a screenshot to see what you mean?

rlaverde:
>Could you post a screenshot to see what you mean?

If you see the video that @dalthviz posted, after undocking the plugin is necessary to resize the window because it is somehow small

spyder/widgets/dock.py
```diff
@@ -153,10 +153,10 @@ def show_nontab_menu(self, event):
 
 class SpyderDockWidget(QDockWidget):
     """Subclass to override needed methods"""
-    plugin_closed = Signal()
+    plugin_closed = Signal(object)
 
     def __init__(self, title, parent):
-        super(SpyderDockWidget, self).__init__(title, parent)
+        super(SpyderDockWidget, self).__init__(title)
```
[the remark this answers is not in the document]
dalthviz: I think this change is not necessary.  I will revert this :+1:

spyder/plugins/base.py
```diff
@@ -169,3 +172,53 @@ def toggle_view(self, checked):
...
+    def create_mainwindow(self):
+        """
+        Create a QMainWindow instance containing this plugin.
+        """
+        raise NotImplementedError
+
+    def create_undock_action(self):
+        """Create the undock action for the plugin."""
+        self.undock_action = create_action(self,
+                                           _("Undock the plugin"),
```
ccordoba12: Let's change this to simply `Undock`, instead of `Undock the plugin`.

spyder/plugins/base.py
```diff
@@ -169,3 +172,53 @@ def toggle_view(self, checked):
...
+                                           triggered=self.undock_plugin)
+
+    def undock_plugin(self):
+        """Undocks the plugin from the MainWindow."""
+        self.undocked = True
+        self.dockwidget.setFloating(True)
+        self.undock_action.setDisabled(True)
+
+
+class PluginMainWindow(QMainWindow):
```
ccordoba12: Please move this class to be before `BasePluginWidget`

spyder/api/plugins.py
```diff
@@ -66,6 +68,10 @@ def __init__(self, main=None):
         self.mainwindow = None
         self.ismaximized = False
         self.isvisible = False
+        self.options_button = create_toolbutton(self, text=_('Options'),
+                                                icon=ima.icon('tooloptions'))
+        self.options_button.setPopupMode(QToolButton.InstantPopup)
+        self.menu = QMenu(self)
```
ccordoba12: `menu` is too generic. Please change it to `options_menu`.

ccordoba12:
@dalthviz, I left some simple comments, then I'll merge.

ccordoba12:
@dalthviz, please squash the last two commits into one with the message "Fix tests".

PR #3826, opened 2016-12-10
PR: Improve auto indentation of line following indented block

### LINKED ISSUE #3800: Auto indentation of line following indented block
The Problem:
If I have a block of indented code, followed by a block of un-indented code, and I attempt to insert additional (blank) lines between the two, pressing enter causes the un-indented line of code to automatically indent. 

If this is a bug, it needs to be fixed, if it is a "feature" then it needs to be made optional. 

1. A line of indented code, followed by a line of un-indented code.
2. Press enter on the un-indented line of code.
image

Expected Output: the "un-indented" line of code moves down a line, and remains un-indented.
image

Actual Output: the "un-indented" line of code moves down a line and becomes indented.
image

Note, this occurs if the two lines directly follow each other, or if there are additional blank (un-indented) lines between them. 

I tried toggling the auto indention in the options, but it didn't make a difference.
image
 
Thanks.

* Spyder Version: 3.0.2
* Python Version: 3.5.2 64bits, Qt 5.6.0, PyQt5 5.6 on Windows

## Dependencies
sphinx >=0.6.6   :  1.5 (OK)
sympy >=0.7.3    :  1.0 (OK)

pwoosam:
Fixes #3800 

----

Added checkbox to editor's preferences to enable and disable "strict indent".

With strict indenting enabled, automatic indentation has the correct behavior expected in issue #3800.
Disabling strict indenting will have the same automatic indentation behavior as before the update.


ccordoba12:
Very quick feedback: I don't agree with adding a configuration option for this. We just have to come with a more intelligent indentation algorithm.

By the way, this is the same indentation algorithm used in the Jupyter notebook (through Codemirror).

ccordoba12:
Let's follow what VSCode and Sublime Text do: if the previous lines is unindented, leave the next line unindented too.

That way we won't need to add an option to control this indentation behavior and will fix #3800 too :-)

ccordoba12:
Also, please merge your branch with the latest 3.x one to fix the errors in CircleCI

pwoosam:
Yeah, I wasn't too sure whether or not to make this a configuration option or not. I wasn't sure if there was anyone out there that saw it as a feature. :p

And that sounds like a good plan, I'll get on it!

The algorithm has some other issues with multi-line statements too where it does not indent enough. I'll open an issue if there isn't already one.

ccordoba12:
Thanks @pwoosam!

ccordoba12:
Your work is breaking a test for indentation. Please review the failure in CircleCI to fix the problem :-)

pwoosam:
@ccordoba12 It took me a while, but I finally got around to fixing the failure. However, I'm not completely confident that this code doesn't break any other auto indent features not covered in the tests.

When I was playing around with it, I noticed small mistakes could break other seemingly unrelated such as intelligent backspace.

ccordoba12:
It'd be great if you could add a test for your work in `widgets/sourcecode/tests/test_autoindent.py`. Then we'll be sure of not breaking this feature in the future :-)

pwoosam:
I actually did not completely fix tabs in 9c99ba4. It's having some issues indenting with tab. If the above line has no whitespace, it won't indent.

I think I went wrong with getting the indentation from the line above the cursor and I used the indentation from prevline and just added an extra indent if that line ended in a colon.

pwoosam:
All I had to do was move a couple of lines down! Such a simple mistake!

I'm much more confident in this fix now :)

ccordoba12:
@pwoosam, there's one use case I'd like to see implemented here too:

```
def f(x):
    return x
<Enter here (in empty/unindented line)>
<Leave next line unindented too, instead of moving cursor at the return indentation level>
```

pwoosam:
@ccordoba12 Would you also like to unindent if you press return after a return statement just like it currently does with continue, break and pass?

ccordoba12:
> Would you also like to unindent if you press return after a return statement?

Yeah, that'd be great too! Thanks for thinking on it :-)

pwoosam:
So currently the way I have this implemented for return statements is that it will automatically unindent the next line just like with break, continue and pass. 
```
def f(x):\n
    return x\n
<new line starts here>
```
It also does as you asked:
```
def f(x):\n
    return x\n
\n
<new line starts here>
```
And it will also allow this:
```
def f(x):\n
    if x > 0:\n
        return x\n
\n
<new line starts here>
```

This would also be useful for break, continue and pass. Here is an example:
```
def f(x):\n
    for i in range(x):\n
        if i % 90:\n
            print(i)\n
        else:\n
            break\n
\n
<new line starts here>
```
```
def f(x):\n
    for i in range(x):\n
        if i % 90:\n
            print(i)\n
        else:\n
            continue\n
\n
<new line starts here>
```
```
def f(x):\n
    if x > 0:\n
        pass\n
\n
<new line starts here>
```
Currently they do:
```
def f(x):\n
    for i in range(x):\n
        if i % 90:\n
            print(i)\n
        else:\n
            break\n
\n
        <new line starts here>
```
```
def f(x):\n
    for i in range(x):\n
        if i % 90:\n
            print(i)\n
        else:\n
            continue\n
\n
        <new line starts here>
```
```
def f(x):\n
    if x > 0:\n
        pass\n
\n
    <new line starts here>
```

pwoosam:
@ccordoba12 Let me know if you would like to see the above implemented for break, continue and pass as well. Otherwise, there is just a print statement that I forgot to remove at line 1980, then this is ready.

ccordoba12:
@pwoosam, I have one final suggestion (and sorry if my previous example with `return` was not clear enough).

I just want to see that a return in *any* line that is unindented and blank, leaves the next line unindented too. That's how VSCode works and I find it a good approach :-)

pwoosam:
@ccordoba12 Sorry, I'm not too familiar with VSCode and I was confused by the return statement. :p 

I believe I've gotten the behavior you wanted now though!

It should keep the same indentation as the line you pressed enter on as long as that line was unindented(relative to the indentation of `prevline`) and blank.

ccordoba12:
@pwoosam, thanks a lot for the extra effort, it's working great now!

Hope to see more contributions from you in the future, you're doing great so far! ;-)

PR #3827, opened 2016-12-10
PR: Don't run Pylint with ciocheck

ccordoba12:
It seems this is the cause of repeated hanging in CircleCI


goanpeca:
:+1: its o pylint complains too much

ccordoba12:
Yep, and it's the one that was making CircleCI hang, so I had no choice :-)

PR #3834, opened 2016-12-13
PR: Add a command line option to load projects 

### LINKED ISSUE #3830: Add a command line option to load projects
Is there an option to load a specific project with command line options? Or, alternatively, specify a  different set of configuration files? The reason is that we are working with multiple users on the same account, and we do not want to have our projects and settings mixed.

We already have different spyder versions installed using environments (with conda, windows), but these different spyder versions all use the same configuration files.

* Spyder Version: 3.0.2
* Python Version: 3.5
* Operating system: windows

andfoy:
Fixes #3830 
The flag option to use this function is ```--project <path>```


spyder/app/cli_options.py
```diff
@@ -37,5 +37,8 @@ def get_options():
                            "not related with Python profiling)")
     parser.add_option('--window-title', type=str, default=None,
                       help="String to show in the main window title")
+    parser.add_option('--project', default=None, type=str,
```
goanpeca: Could we add a short flag as well?

`-p` what do you think @ccordoba12?

ccordoba12: Yeah, why not?

goanpeca: Do we have a way of opening a file? that would be useful as well

spyder somefile.py

ccordoba12: Yes, it has been there since Spyder 2.2 :-)

goanpeca: Ok good :-)

ccordoba12:
This looks to me, thanks @andfoy!

PR #3842, opened 2016-12-15
PR: Validate that Main Interpreter is really a Python interpreter

### LINKED ISSUE #3756: Verify that the interpreter selected in Main Interpreter is really a Python interpreter
It seems people sometimes selects a wrong Python executable in

    Preferences > Main Interpreter

and that leads to not being able to start Python or IPython consoles (see for example issue #3750).

Because of that we need to verify that they are correctly selecting a Python interpreter and not other executables (e.g. the IPython one).

----

@mariacamilaremolinagutierrez, please work on this one.

mariacamilaremolinagutierrez:
Fixes #3756


spyder/plugins/maininterpreter.py
```diff
@@ -153,6 +158,28 @@ def python_executable_switched(self, custom):
             if custom:
                 self.warn_python_compatibility(cust_pyexec)
 
+    def is_python_interpreter(self, pyexec):
+        if not osp.isfile(pyexec):
+            return False
+        args = ["-c", "print('valid_python_interpreter')"]
```
ccordoba12: You can also run in a terminal:

    ipython -c 'print("foo")'

so this is not enough for a test.

ccordoba12: I think a better test is

1. Verify that the file is called `python`, `python2` or `python3`
2. Verify that it's a binary (using `is_text_file` from utils/encoding.py)
3. Run the command you defined here.

If those three criteria are met, then we can be more certain (although no completely sure :-) that the selected file is a Python interpreter.

ccordoba12: The problem with point 1. on Linux is that `python{2,3}` are symlinks, so you should follow them to the file they point to and check that it's a binary.

Also, about 1., users could introduce `python2.7` or `python3.5`, so that should be covered too here.

andfoy: Isn't it possible to execute this command?
```sh
python --version
```

ccordoba12: Problem is this also works for `ipython` (although we could also filter for 2.7.x and 3.{4,5,6})

Besides, most Linux programs have a `--version` option and some could coincide with the Python versions I mentioned above :-)

goanpeca: python -h 
and
ipython -h  ?

We could regex for stuff expected to be found

ccordoba12: That's a better idea. We could regex for the first line contents, which are very different.

spyder/plugins/maininterpreter.py
```diff
@@ -161,18 +161,17 @@ def python_executable_switched(self, custom):
     def is_python_interpreter(self, pyexec):
         if not osp.isfile(pyexec):
             return False
-        args = ["-c", "print('valid_python_interpreter')"]
         try:
-            proc = programs.run_program(pyexec, args)
+            proc = programs.run_program(pyexec, ["-h"])
             valid = str(proc.communicate()[0])
-            if not valid.startswith("b'valid_python_interpreter"):
+            if len(valid) > 2 and valid[2:].startswith("usage: {}".format(pyexec)):
```
ccordoba12: I think this match is still too weak. Please create a more complete function with some of my previous suggestions, add it to utils/programs.py and also add a test for it in utils/tests/test_programs.py

ccordoba12: Especially, I'd like to see a validation for the selected file being a binary

ccordoba12: Py34, not 24 ;-)

goanpeca: oops

mariacamilaremolinagutierrez:
@ccordoba12 I already make all the changes you suggested except for the tests in utils/tests/test_programs.py. I did test it on my computer because I knew the paths of real and fake python interpreters, but in this one I'd have the issue of knowing which file to use. I thought about using the python interpreter that runs spyder, if you think this is correct, is there a function already in spyder that can tell me which python is running spyder? Thank you!

spyder/utils/programs.py
```diff
@@ -462,6 +462,24 @@ def is_module_installed(module_name, version=None, installed_version=None,
             
             return check_version(actver, version, symb)
 
+def is_python_interpreter(filename):
+    """Evaluate wether a file is a python interpreter or not."""
+    real_filename = os.path.realpath(filename) # To follow symlink if existent
+    if not osp.isfile(real_filename) or encoding.is_text_file(real_filename):
+        return False
+    possible_names = ['python', 'python2', 'python2.7', 'python3', 'python3.4',
+                      'python3.5']
```
goanpeca: Python 3.6 is our, should we have a more robust thing moving forward?, otherwise we will need to update this every now and then (which seems something we can easily forget)

spyder/utils/programs.py
```diff
@@ -462,6 +462,24 @@ def is_module_installed(module_name, version=None, installed_version=None,
             
             return check_version(actver, version, symb)
 
+def is_python_interpreter(filename):
+    """Evaluate wether a file is a python interpreter or not."""
+    real_filename = os.path.realpath(filename) # To follow symlink if existent
```
goanpeca: There should be 2 spaces after code finishes and a comment starts.

`real_filename = os.path.realpath(filename)  # To follow symlink if existent`

spyder/utils/programs.py
```diff
@@ -462,6 +462,24 @@ def is_module_installed(module_name, version=None, installed_version=None,
...
+    if not osp.isfile(real_filename) or encoding.is_text_file(real_filename):
+        return False
+    possible_names = ['python', 'python2', 'python2.7', 'python3', 'python3.4',
+                      'python3.5']
+    if not any([real_filename.endswith(pn) for pn in possible_names]):
+        return False
+    try:
+        proc = run_program(filename, ["-h"])
+        valid = str(proc.communicate()[0])
```
goanpeca: Is this a safe Py27 and Py3 way of doing this?, to_text_string? @ccordoba12 ?

ccordoba12: Yeah, please use `to_text_string` instead of `str` here :-)

spyder/utils/programs.py
goanpeca: Maybe using a regex is better

`python(?P<version>\d\.?\d*)?$`

This will match any of the options plus any version.. then we can check that version part (if not empty) is a float

ccordoba12:
@mariacamilaremolinagutierrez, the result is really nice! Thanks a lot for the extra effort ;-)

About the tests, please create several of them:

1. On Linux test with `$HOME/miniconda/bin/python` and `$HOME/miniconda/bin/ipython` (as false interpreter).
2. On Windows test with `%PYTHON%\bin\python.exe` and `%PYTHON%\Scripts\ipython.exe` (`%PYTHON%` is an env var we set in Appveyor).

You must skip those four tests if the `CI` environment variable is not set, to only run them in our CI services.

spyder/utils/programs.py
ccordoba12: True, we need to be smarter about this list. What if create a simple regexp that checks for `python3.*`?

goanpeca: Well, this is a generic one, so we dont have to have both a list with options and a regex

`r'python(?P<version>\d\.?\d*)?$'`

There is someone doing a python2.8 out there also ;-)

ccordoba12: Ok, better then.

goanpeca: @mariacamilaremolinagutierrez in order to test this, we need to break this small helper function

```python
def is_python_interpreter_valid_name(fname):
    """Check that the python interpreter file has a valid name."""
    pattern = r'python(?P<version>\d\.?\d*)?$'
    # etc......
```

That way we can test this method by itself to make sure it is grabbing the right stuff.

spyder/utils/tests/test_programs.py
```diff
@@ -36,7 +37,39 @@ def test_run_python_script_in_terminal_with_wdir_empty(tmpdir, qtbot):
...
     assert res == 'done'
 
-    
+@pytest.mark.skipif(os.environ.get('TEST_CI_WIDGETS', None) == True,
+                    reason='To only run the test in spyders CI services.')
+def test_is_python_interpreter_linux_true():
+    filename = "$HOME/miniconda/bin/python"
+    valid = is_python_interpreter(filename)
+
+    assert valid == True
```
goanpeca: `assert valid` works

spyder/utils/tests/test_programs.py
```diff
@@ -36,7 +37,39 @@ def test_run_python_script_in_terminal_with_wdir_empty(tmpdir, qtbot):
...
+
+@pytest.mark.skipif(os.environ.get('TEST_CI_WIDGETS', None) == True,
+                    reason='To only run the test in spyders CI services.')
+def test_is_python_interpreter_linux_false():
+    filename = "$HOME/miniconda/bin/ipython"
+    valid = is_python_interpreter(filename)
+
+    assert valid == False
```
goanpeca: `assert not valid` also works

spyder/utils/tests/test_programs.py
```diff
@@ -36,7 +37,39 @@ def test_run_python_script_in_terminal_with_wdir_empty(tmpdir, qtbot):
...
+
+@pytest.mark.skipif(os.environ.get('TEST_CI_WIDGETS', None) == True,
+                    reason='To only run the test in spyders CI services.')
+def test_is_python_interpreter_windows_true():
+    filename = "%PYTHON%\bin\python.exe"
+    valid = is_python_interpreter(filename)
+
+    assert valid == True
```
goanpeca: Same here

spyder/utils/tests/test_programs.py
```diff
@@ -36,7 +37,39 @@ def test_run_python_script_in_terminal_with_wdir_empty(tmpdir, qtbot):
...
+
+@pytest.mark.skipif(os.environ.get('TEST_CI_WIDGETS', None) == True,
+                    reason='To only run the test in spyders CI services.')
+def test_is_python_interpreter_windows_false():
+    filename = "%PYTHON%\Scripts\ipython.exe"
+    valid = is_python_interpreter(filename)
+
+    assert valid == False
```
goanpeca: Same here

spyder/utils/tests/test_programs.py
jitseniesen: I'd write `assert is_python_interpreter(filename)` or `assert not is_python_interpreter(filename)`, depending on whether it should be True or False.

goanpeca: Yep, even more concise!

Happy new year @jitseniesen  :-) 🎉

jitseniesen: Also the best wishes for 2017 to you, @goanpeca! and to whoever else reads this :beer: :balloon:

spyder/utils/tests/test_programs.py
```diff
@@ -36,7 +37,39 @@ def test_run_python_script_in_terminal_with_wdir_empty(tmpdir, qtbot):
     res = outfilepath.read()
     assert res == 'done'
 
-    
+@pytest.mark.skipif(os.environ.get('TEST_CI_WIDGETS', None) == True,
```
ccordoba12: We don't set `TEST_CI_WIDGETS` in CircleCI because we don't test our widgets there. That's why I told you to use the `CI` variable instead, which is set automatically by CircleCI, Travis and AppVeyor regardless of what we do in those services.

spyder/utils/tests/test_programs.py
ccordoba12: Happy new year to you too @jitseniesen! Best wishes to you and your family for 2017 ;-)

spyder/utils/tests/test_programs.py
```diff
@@ -36,7 +37,39 @@ def test_run_python_script_in_terminal_with_wdir_empty(tmpdir, qtbot):
...
+                    reason='To only run the test in spyders CI services.')
```
ccordoba12: Change this to

    It only runs in CI services.

spyder/utils/tests/test_programs.py
```diff
@@ -36,7 +37,39 @@ def test_run_python_script_in_terminal_with_wdir_empty(tmpdir, qtbot):
...
+def test_is_python_interpreter_linux_true():
```
ccordoba12: Simplify this name like this

    test_is_valid_interpreter

spyder/utils/tests/test_programs.py
```diff
@@ -36,7 +37,39 @@ def test_run_python_script_in_terminal_with_wdir_empty(tmpdir, qtbot):
...
+    filename = "$HOME/miniconda/bin/python"
+    valid = is_python_interpreter(filename)
+
+    assert valid == True
+
+@pytest.mark.skipif(os.environ.get('TEST_CI_WIDGETS', None) == True,
+                    reason='To only run the test in spyders CI services.')
+def test_is_python_interpreter_linux_false():
```
ccordoba12: Please change this name to

    is_invalid_interpreter

ccordoba12:
@mariacamilaremolinagutierrez, there are some details to polish but this on the right track!

I'd suggest that instead of having four tests (which, I know, I suggested at the beginning) we could only have two, like this:

```python
test_is_valid_interpreter():
    if os.name == 'nt':
        # Windows code here
    else:
        # Linux code here
```

and

```python
test_is_invalid_interpreter():
    if os.name == 'nt':
        # Windows code here
    else:
        # Linux code here
```

ccordoba12:
Or tests could be even simpler. At the beginning of `test_programs.py` you could define variables with the valid and invalid interpreters per operating system, and then the `test_is_{in}valid` tests would have the exact same form, like this:

```python
if os.name == 'nt':
    VALID_INTERPRETER = '%PYTHON%\...'
    INVALID_INTERPRETER = '%PYTHON%\Scripts\...'
else:
    VALID_INTERPRETER = '$HOME/...'
    INVALID_INTERPRETER = '$HOME/...'

def test_is_valid_interpreter()
    assert is_python_interpreter(VALID_INTERPRETER)
```

spyder/utils/tests/test_programs.py
mariacamilaremolinagutierrez: Awesome, thanks a lot to the 3 of you and have a wonderful 2017 :heart:

spyder/utils/tests/test_programs.py
```diff
@@ -6,10 +6,17 @@
...
 import os
-
 import pytest
 
-from spyder.utils.programs import run_python_script_in_terminal
+from spyder.utils.programs import (run_python_script_in_terminal,
+                                   is_python_interpreter)
+
+if os.name == 'nt':
+    VALID_INTERPRETER = r'%PYTHON%\bin\python.exe'
```
ccordoba12: Sorry, my code was a simple draft of what you should do, not the right solution (that's why tests are still failing :-). So this should be

```python
python_dir = os.environ['PYTHON']
VALID_INTERPRETER = os.path.join(python_dir, 'bin', 'python.exe')
```

spyder/utils/tests/test_programs.py
```diff
@@ -6,10 +6,17 @@
...
+    INVALID_INTERPRETER = r'%PYTHON%\Scripts\ipython.exe'
+else:
+    VALID_INTERPRETER = r'$HOME/miniconda/bin/python'
```
ccordoba12: And this needs to be (please check it until tests pass)

```python
home = os.environ['HOME']
VALID_INTERPRETER = os.path.join(home, 'miniconda', 'bin', 'python')
```

spyder/utils/tests/test_programs.py
ccordoba12: Sorry, small correction (also please fix all what's needed until tests pass ;-)

```python
python_dir = os.environ['PYTHON']
VALID_INTERPRETER = os.path.join(python_dir, 'python.exe')
```

spyder/utils/programs.py
```diff
@@ -462,6 +462,29 @@ def is_module_installed(module_name, version=None, installed_version=None,
             
             return check_version(actver, version, symb)
 
+def is_python_interpreter_valid_name(filename):
```
ccordoba12: Please add a test to verify this function is working correctly (as @goanpeca asked you)

ccordoba12:
@mariacamilaremolinagutierrez, please don't touch this branch. I'm working on it to finish this PR.

spyder/utils/tests/test_programs.py
```diff
@@ -37,37 +43,15 @@ def test_run_python_script_in_terminal_with_wdir_empty(tmpdir, qtbot):
...
-    valid = is_python_interpreter(filename)
-
-    assert valid == True
-
-@pytest.mark.skipif(os.environ.get('TEST_CI_WIDGETS', None) == True,
-                    reason='To only run the test in spyders CI services.')
-def test_is_python_interpreter_windows_false():
-    filename = "%PYTHON%\Scripts\ipython.exe"
-    valid = is_python_interpreter(filename)
+@pytest.mark.skipif(os.environ.get('CI', None) == True,
```
jitseniesen: @mariacamilaremolinagutierrez Is this the right condition? It appears that the test will be skipped if CI is set to True, but I think you want the test to run when CI is set to True.

ccordoba12: The problem is these tests are running locally (which is what I wanted to prevent)

@mariacamilaremolinagutierrez, please fix that.

PR #3847, opened 2016-12-16
PR: Make regex fully functional when using find & replace 

### LINKED ISSUE #3249: Regex is not fully functional when using to find&replace
I want to use the following regex
`\["([a-z_]*)"\]`
and want to replace it with 
`.\1`
to change pars["delta_s"] to pars.delta_s or similar. This does work on some lines but not on all. The regular expression is met, but somehow Spyder fails to actual replace the string. More importantly, when clicking Replace/find the cursor does not (always) jump to the next hit but skips one hit. If then replace/find is clicked again it jumps bakc to the one it skipped before

andfoy:
Fixes #3249

### DIFF
--- a/spyder/widgets/findreplace.py
+++ b/spyder/widgets/findreplace.py
@@ -360,7 +360,8 @@ class FindReplace(QWidget):
                     cmptxt1 = search_text if case else search_text.lower()
                     cmptxt2 = seltxt if case else seltxt.lower()
                     if not pattern:
-                        if self.editor.has_selected_text() and cmptxt1 == cmptxt2:
+                        has_selected = self.editor.has_selected_text()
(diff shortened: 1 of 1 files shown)


ccordoba12:
@andfoy, there are a couple of style failures in CircleCI. Please fix them :-)

andfoy:
@ccordoba12 It was a classic PEP8 error :)

goanpeca:
Any chance we take this opportunity to make a test that checks this problem?

ccordoba12:
@goanpeca, it's not easy to add a test for this because it lies in the interaction between the find/replace widget and the Editor. However, I manually tested that things are working as expected now :-)

PR #3853, opened 2016-12-17
PR: Move rope_patch.py to utils/introspection

ccordoba12:
Also add pytest's for rope_plugin.


rlaverde:
`test_get_info` is failing for me, I tested with `defaults` and `conda-forge` channels:
`numpy                     1.11.2                   py35_0` 

`numpy                     1.11.2          py35_blas_openblas_200  [blas_openblas]  conda-forge`

```
    @pytest.mark.skipif(not numpy, reason="Numpy required")
    def test_get_info():
        source_code = "import numpy; numpy.ones"
        docs = p.get_info(CodeInfo('info', source_code, len(source_code), __file__))
>       assert docs['calltip'].startswith('ones(') and docs['name'] == 'ones'
E       assert (True and 'numpy.core.numeric.ones' == 'ones'
E        +  where True = <built-in method startswith of str object at 0x7f0e2bd21500>('ones(')
E        +    where <built-in method startswith of str object at 0x7f0e2bd21500> = "ones(shape, dtype=None, order='C')".startswith
E         - numpy.core.numeric.ones
E         + ones)
```

Have I something wrong with my local configuration? I recreate the enviroment with `conda create -n spyder spyder python=3.5`, `conda install pytest pytest-cov pytest-qt numpy pandas`, and It's still happening

ccordoba12:
I tested it locally and it works fine for me, but I'll try to reproduce it with your commands

PR #3854, opened 2016-12-17
PR: Add disambiguation functionality to the Python console

### LINKED ISSUE #1165: Console tabs are too big because they display the full filename

dalthviz:
Fixes #1165

### DIFF
--- a/spyder/plugins/externalconsole.py
+++ b/spyder/plugins/externalconsole.py
@@ -27,7 +27,7 @@ from qtpy.QtWidgets import (QButtonGroup, QGroupBox, QHBoxLayout, QLabel,
 from spyder import dependencies
 from spyder.config.base import _, SCIENTIFIC_STARTUP
 from spyder.config.main import CONF
-from spyder.utils import encoding, programs
+from spyder.utils import encoding, programs, sourcecode
 from spyder.utils import icon_manager as ima
 from spyder.utils.misc import (get_error_match, get_python_executable,
                                is_python_script, remove_backslashes)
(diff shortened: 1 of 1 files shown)


dalthviz:
A preview of the change:

pythonconsole

spyder/plugins/externalconsole.py
```diff
@@ -707,22 +723,15 @@ def process_finished(self, shell_id):
...
+        """Return widget icon."""
         return ima.icon('console')
     
     def get_focus_widget(self):
-        """
-        Return the widget to give focus to when
-        this plugin's dockwidget is raised on top-level
-        """
+        """Return the widget to give focus to when
+        this plugin's dockwidget is raised on top-level"""
```
ccordoba12: Let's change it to ```"""Return the widget to give focus to."""``` to fix the style error.

ccordoba12:
Since style tests are now passing, please squash the last four commits into one and push again :-)

ccordoba12:

PR #3861, opened 2016-12-20
PR: Show kernel initialization errors in the IPython console

### LINKED ISSUE #3458: Feature Request: Clear Console After Kernel Restart
The previous behaviour of spyder was to clear the console after a kernel restart whereas now it simply prints a `Restarting kernel...` and a horizontal line but leaves the full history in the console:
image

Whilst this may sometimes be desirable, at other times the history of the previous session is simply noise and makes it difficult to determine where the output from the new session started. This would be less of an issue if there were a way to clear the console manually but the `%clear` magic just scrolls down and doesn't actually clear the contents of the console. 

It would be nice if there were a configuration setting so you could choose whether or not to clear the console after a kernel restart.

ccordoba12:
Fixes #3458.

----

This PR does several things:

1. Redirects the kernel process stderr output to a file and shows it in case of failure in our consoles. With this I'm trying to avoid the infinite stream of *Kernel died, restarting* messages and give feedback to users of what's the cause of the error.
2. Introduces a shortcut to restart kernels, so that we can override the one defined in qtconsole and control this process ourselves.
3. Clears the console on restarts.

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -301,6 +301,10 @@ class MainWindow(QMainWindow):
             mac_style = mac_style.replace('$IMAGE_PATH', img_path)
             self.setStyleSheet(mac_style)
 
+        # Create our TEMPDIR
+        if not osp.isdir(programs.TEMPDIR):
+            os.mkdir(programs.TEMPDIR)
+
         # Shortcut management data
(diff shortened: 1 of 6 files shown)


ccordoba12:
Ok, let's hope this one improves things for people with kernel problems :-)

PR #3864, opened 2016-12-20
PR: Extra selections is now an ordered dictionary

### LINKED ISSUE #3859: Extra selections should be ordered
##### This is more of a suggestion:
Extra selections dict in base.py is a dictionary, but an OrderedDict may be a better option.
Using an ordered dictionary would allow some selections to take precedence over others, ensuring that the last selection added to extra selections is the one which is highlighted.

This would be a solution to [issue 7](https://github.com/spyder-ide/spyder-vim/issues/7) in spyder-vim.

* Spyder Version: latest
* Python Version: 3.5.2

pwoosam:
Extra selections is ordered and sorted.
The smaller selection is given higher priority.

Fixes #3859 
Fixes spyder-ide/spyder-vim#7


ccordoba12:
Thanks @pwoosam, looks good to me :-)

PR #3871, opened 2016-12-21
PR: Add option to "clear all" variables before running file

### LINKED ISSUE #2563: Enhancement: add option to "clear all" variables before running script
I can define a variable in the console and have that same variable be available in script; the equivalence of Matlab's `clear all` isn't an option in preferences.

That is, at the IPython prompt I can define `a = 10`. Then if my script contains one line,

``` python
# script.py
print(a) # prints 10 even though `a` is not defined *in script*
```

It'd certainly be nice to disable that behavior. Other interpreters (e.g., default Python, IPython's `%run`) don't have this behavior enabled -- it throws `NameError: name 'a' is not defined`. Matlab has this feature where it reads variables defined from the variable explorer (and not the script) but it can be easily disabled with `clear all`. In Spyder, it requires a clever hack to "clear all".
### Proposed solution

Under "Run preferences" or in Spyder preferences > Run, I propose an option to enable/disable the equivalent of Matlab's "clear all". I propose the wording/default option be something like "[x] Variables in script must be defined in script. Otherwise, variables must be defined in Variable Explorer."
### Current hack

In the settings under IPython console > Startup, I have enabled a file to be run whenever I run a file. That file consists of a script to clear all the variables (which I found [here](http://scientificcomputingco.blogspot.com/2013/10/python-spyder-clear-workspaceshell.html)):

``` python
def clear_all():
    """Clears all the variables from the workspace of the spyder application."""
    gl = globals().copy()
    for var in gl:
        if var[0] == '_': continue
        if 'func' in str(globals()[var]): continue
        if 'module' in str(globals()[var]): continue

        del globals()[var]
if __name__ == "__main__":
    clear_all()
```

**Details:** I'm running Spyder 2.3.5.2

andfoy:
Fixes #2563

### DIFF
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -809,7 +809,7 @@ class IPythonConsole(SpyderPluginWidget):
             return client.shellwidget
 
     def run_script_in_current_client(self, filename, wdir, args, debug,
-            post_mortem):
+            post_mortem, clear_variables):
         """Run script in current client, if any"""
         norm = lambda text: remove_backslashes(to_text_string(text))
         client = self.get_current_client()
@@ -832,7 +832,7 @@ class IPythonConsole(SpyderPluginWidget):
                 line += "\"%s\"" % to_text_string(filename)
                 if args:
                     line += " %s" % norm(args)
-            self.execute_code(line)
+            self.execute_code(line, clear_variables)
             self.visibility_changed(True)
             self.raise_()
         else:
@@ -849,9 +849,11 @@ class IPythonConsole(SpyderPluginWidget):
             directory = encoding.to_unicode_from_fs(directory)
             shellwidget.set_cwd(directory)
 
-    def execute_code(self, lines):
+    def execute_code(self, lines, clear_variables):
         sw = self.get_current_shellwidget()
         if sw is not None:
+            if clear_variables:
+                sw.reset_namespace(True)
             sw.execute(to_text_string(lines))
             self.activateWindow()
             self.get_current_client().get_control().setFocus()
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -854,7 +854,7 @@ class IPythonConsole(SpyderPluginWidget):
         sw = self.get_current_shellwidget()
         if sw is not None:
             if clear_variables:
-                sw.reset_namespace(True)
+                sw.reset_namespace(force=True)
             sw.execute(to_text_string(lines))
             self.activateWindow()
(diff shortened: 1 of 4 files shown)


ccordoba12:
Please amend the last commit message with something like

    Fix style issues

instead of

    squashed everything after 7d8edd1

I think it's clearer to have such a message for Spyder git history :-)

spyder/widgets/ipythonconsole/shell.py
```diff
@@ -139,18 +139,20 @@ def short_banner(self):
...
-    def reset_namespace(self):
-        """Resets the namespace by removing all names defined by the user"""
-
-        reply = QMessageBox.question(
-            self,
-            _("Reset IPython namespace"),
-            _("All user-defined variables will be removed."
-            "<br>Are you sure you want to reset the namespace?"),
-            QMessageBox.Yes | QMessageBox.No,
+    def reset_namespace(self, predefined=False):
```
ccordoba12: Please change this to

    def reset_namespace(self, force=False):

spyder/widgets/ipythonconsole/shell.py
```diff
@@ -139,18 +139,20 @@ def short_banner(self):
...
+            reply = QMessageBox.question(self, reset_str,
+                                         warn_str,
+                                         QMessageBox.Yes | QMessageBox.No
             )
 
-        if reply == QMessageBox.Yes:
+            if reply == QMessageBox.Yes:
+                self.execute("%reset -f")
+        else:
             self.execute("%reset -f")
```
ccordoba12: To avoid saving this line on every execution of `runfile`, please change this to

```python
shellwidget = self.get_current_shellwidget()
if shellwidget is not None:
    shellwidget.silent_execute("%reset -f")
```

ccordoba12: Sorry, this just needs to be

```python
self.silent_execute("%reset -f")
```

(I thought this method was part of the IPyhon console plugin).

spyder/plugins/ipythonconsole.py
```diff
@@ -849,9 +849,12 @@ def set_current_client_working_directory(self, directory):
             directory = encoding.to_unicode_from_fs(directory)
             shellwidget.set_cwd(directory)
 
-    def execute_code(self, lines):
+    def execute_code(self, lines, clear_variables):
+        """Execute code instructions."""
         sw = self.get_current_shellwidget()
         if sw is not None:
+            if clear_variables:
+                sw.reset_namespace(True)
```
ccordoba12: Please change this to

    sw.reset_namespace(force=True)

PR #3878, opened 2016-12-22
PR: Add function/class search shortcut, menu and toolbar entries

### LINKED ISSUE #3855: Add fuzzy function/class search shortcut
Similar to the ctrl+p shortcut that fuzzy search filenames, it will be nice to steal two more shortcuts from sublime:
ctrl+r: same as ctrl+p and writing @ (which will the fuzzy searches function/classes in the current file)
ctrl+shift+r: same as ctrl+r but project and open-files wise (searches for functions in the whole project and the currently open files)

(at least the ctrl+r one is very simple, sublime implements it exactly as ctrl+p and writing @ - so that the search bar has @ in the beginning)
what do you think?

andfoy:
Fixes #3855


spyder/defaults/defaults-3.0.0.ini
```diff
@@ -318,7 +318,8 @@ _/debug step into = Ctrl+F11
 _/fullscreen mode = F11
 editor/toggle comment = Ctrl+1
 editor/go to definition = Ctrl+G
-editor/show/hide project explorer = Ctrl+Alt+P
+editor/show/hide project explorer = Ctrl+Alt+J
+_/symbol finder = Ctrl+Alt+P
 _/debug step return = Ctrl+Shift+F11
```
goanpeca: @ccordoba12  I thought we were supposed to not touch these files?

andfoy: Sorry! I thought that file should be modified because a default shortcut was altered and a new one was introduced in this PR

goanpeca: https://github.com/spyder-ide/spyder/tree/master/spyder/defaults

:-)

andfoy: So, should I remove those edits?

goanpeca: Yep

spyder/plugins/editor.py
```diff
@@ -651,6 +651,14 @@ def get_plugin_actions(self):
         self.register_shortcut(self.file_switcher_action, context="_",
                                name="File switcher", add_sc_to_tip=True)
 
+        self.symbol_finder_action = create_action(self, _('Symbol finder...'),
+                                            icon=ima.icon('symbol_find'),
+                                            tip=_('Fast symbol search in file'),
+                                            triggered=self.call_symbol_finder,
+                                            context=Qt.ApplicationShortcut)
```
goanpeca: Not saying you should change this, just a comment for reference.

```python

# I tend to favor this one for several reasons. 
# It can be automatically performed and enforced by tools like yapf
# it results in more lines of code but ir more readable as each line corresponds 
# to a parameter
self.symbol_finder_action = create_action(
    self,
    _('Symbol finder...'),
    icon=ima.icon('symbol_find'),
    tip=_('Fast symbol search in file'),
    triggered=self.call_symbol_finder,
    context=Qt.ApplicationShortcut)

# or 

self.symbol_finder_action = create_action(self,
                                          _('Symbol finder...'),
                                          icon=ima.icon('symbol_find'),
                                          tip=_('Fast symbol search in file'),
                                          triggered=self.call_symbol_finder,
                                          context=Qt.ApplicationShortcut)
```

spyder/config/main.py
```diff
@@ -426,7 +427,7 @@
               'editor/run cell and advance': RUN_CELL_AND_ADVANCE_SHORTCUT,
               # -- In plugins/editor.py
               'editor/show/hide outline': "Ctrl+Alt+O",
-              'editor/show/hide project explorer': "Ctrl+Alt+P",
```
ccordoba12: I think this shortcut is no more needed because

1. If Spyder has no project active, the Project explorer is empty.
2. If the project is active, I think users would expect the Project explorer to be visible all the time.

@goanpeca, what do you think?

goanpeca: Agreed!

ccordoba12: Ok, then @andfoy please remove this shortcut as part of this PR.

jitseniesen: > If the project is active, I think users would expect the Project explorer to be visible all the time.

I don't mind that much about the shortcut, but I don't expect the Project explorer to be visible all the time. I actually close the Project explorer when working on my laptop because the screen is not big enough, and I find it pretty annoying that it always comes back when I launch Spyder. Even on my desktop screen, which is big enough, I don't find the Project explorer particularly useful - I only use it when opening files.

ccordoba12: Well, the Project Explorer can be closed at any time, so it's not enforced on users. It's just a visual clue that you're working on a project.

>  I find it pretty annoying that it always comes back when I launch Spyder

Please submit a PR to fix that. I don't have a problem with it :-)

ccordoba12:
I really like this!! Nice job at adding a menu and toolbar entries for this action :+1: :-)

Issue #2514, opened 2015-07-02
DataFrames with headers that contain BOM utf-8 data are freezing/crashing Spyder (in Python 2)


kapoost:
I've found that reason of Variable explorer crashes might be BOM gibberish (for example: '\xef\xbb\xbf') that appears at beginning of first column name. It is invisible for df.columns but appears when I call df.columns[0]. For some reason pd.read_csv() doesn't strip this.

**Workaround is always check name using df.columns[0] & changing it before using Variable explorer**

ccordoba12:
@kapoost, thanks for creating this issue. On question: what do you mean exactly by

> changing it before using Variable explorer

How do we need to change it?

kapoost:
As I've mentioned - it is only workaround so basically I'm doing it like this:
```python
in: a = pd.read_csv('/Users/kapoost/Downloads/Datausage-9.csv') #this csv starts with BOM
in: a.columns #would not show any problem but...
out: Index([u'﻿Agency', u'Advertiser', u'Country', u'SegmentsGroup', u'Impressions', u'Revenue', u'Leads', u'Clicks'], dtype='object')
in: a.columns[0] #this will show the source of our crashes
out: '\xef\xbb\xbfAgency'
in: a.columns = ['Agency', 'Advertiser', 'Country', 'SegmentsGroup', 'Impressions', 'Revenue', 'Leads', 'Clicks'] #manual overwrite names of columns will erase BOM element. This DataFrame would not cause problem with Variable explorer :)
```
this gives me DataFrame without faulty name of first column so  Spyder don't crashes when I'm opening Variable explorer and browsing data in visual mode :)

I think that simplest solution would be correcting read_csv() But I've found that it is old unresolved issue: https://github.com/pydata/pandas/issues/4793

ccordoba12:
@kapoost, I'm using the file mentioned in this comment: https://github.com/pydata/pandas/issues/4793#issuecomment-25310750 to test this problem. I can reproduce it in Python 2 but not in Python 3 (in case you're interested).

I think I know how to fix it :-)

PR #3898, opened 2016-12-30
PR: Get the proper encoding for the header text of Dataframe editor

### LINKED ISSUE #3896: Spyder Variable explorer crashes when double clicking a 2 line DataFrame
## Description

1. place a csv version of the attached fuct.csv on c drive
2. Run the below:

        import pandas as pd
        dfraw = pd.read_csv(r"C:\fuct.csv")

3. go to Variable explorer and double click to view dfraw

I attach the crash message in Spyder

spyder_crash

The file can be saved as csv (the input to dataframe)
[fuct.txt](https://github.com/spyder-ide/spyder/files/678862/fuct.txt)

* Spyder Version: 3.0.2 
* Python Version: 2.7.12
* Qt Versions: 5.6.0, PyQt5 5
[fuct.txt](https://github.com/spyder-ide/spyder/files/678855/fuct.txt)

.6 on Windows

## Dependencies
```
sympy >=0.7.3    :  1.0 (OK)

```

dalthviz:
Fixes #3896


spyder/widgets/variableexplorer/dataframeeditor.py
```diff
@@ -197,6 +197,10 @@ def headerData(self, section, orientation, role=Qt.DisplayRole):
                 except:
                     header = to_text_string(self.df_header[0])
                 return to_qvariant(header)
+            elif PY2:
+                header = self.df_header[section-1]
```
goanpeca: `section - 1` :-p

spyder/widgets/variableexplorer/dataframeeditor.py
```diff
@@ -197,6 +198,10 @@ def headerData(self, section, orientation, role=Qt.DisplayRole):
                 except:
                     header = to_text_string(self.df_header[0])
                 return to_qvariant(header)
+            elif PY2 and isinstance(self.df_header[section-1], TEXT_TYPES):
+                header = self.df_header[section-1]
```
ccordoba12: Please add a comment above this line explaining the need of this change, like in the previous `elif`.

ccordoba12:
Great job @dalthviz!

I also want a test for this and (as a bonus) for the other `elif` in `headerData` (the one that fixes issue #2514). You could use the data provided by the OPs in both issues, and add those tests to `widgets/variableexplorer/tests/test_dataframeeditor.py`

spyder/widgets/variableexplorer/tests/test_dataframeeditor.py
```diff
@@ -174,6 +174,15 @@ def test_change_format_with_format_not_starting_with_percent(qtbot, monkeypatch)
...
         editor.change_format()
 
+def test_header_encoding():
+    df = DataFrame({"ä": [1],
+                    "ï»¿Unieke_Idcode": ['a'],
+                    1: ['abcd'],
+                    "f": ['31/12/16'],
+                    "": ['']})
+    editor = DataFrameEditor(None)
+    editor.setup_and_check(df)
```
ccordoba12: Tests always assert that the result of an operation is equal to some expected value. However, here you're not doing that. You're just testing that the DataFrame editor doesn't crash, but not that the header information is correct.

Also, please add a test for issue #2514, as I told you so ;-)

spyder/widgets/variableexplorer/tests/test.csv
```diff
@@ -0,0 +1,2 @@
+﻿,﻿Unieke_Idcode,a,b,c,d,f,g,h,i,j,k,l,m,n,o,p,q
+0,flooooo,0.55,10100103,595,sdsdsdf,sadsf,70,..,X1,30,01/02/2014,8.72E+12,sdsf,dsffds,dsdfbvdfg,ff,
```
ccordoba12: Please call this file `issue_3896.csv`

spyder/widgets/variableexplorer/tests/test_dataframeeditor.py
```diff
@@ -174,6 +179,36 @@ def test_change_format_with_format_not_starting_with_percent(qtbot, monkeypatch)
     with qtbot.assertNotEmitted(editor.sig_option_changed):
         editor.change_format()
 
+def test_header_encoding():
+    data =  u"""a,Unieke_Idcode,ü,b,1
+            0,flooooo,0.55,10100103,31/12/2016
+            """
+    test_data = StringIO(data)
```
ccordoba12: Let's try not to use `StringIO` but instead read the file directly, like

    read_csv('issue_3896.csv')

spyder/widgets/variableexplorer/tests/test_dataframeeditor.py
```diff
@@ -174,6 +179,36 @@ def test_change_format_with_format_not_starting_with_percent(qtbot, monkeypatch)
...
+    df = read_csv(test_data, encoding='utf-8')
```
ccordoba12: Don't pass an `encoding` parameter here. The idea of the code you added is for us to detect the encoding of header.

spyder/widgets/variableexplorer/tests/test_dataframeeditor.py
```diff
@@ -174,6 +179,36 @@ def test_change_format_with_format_not_starting_with_percent(qtbot, monkeypatch)
...
+    assert model.headerData(0, orientation=Qt.Horizontal) == u"Index"
+    assert model.headerData(1, orientation=Qt.Horizontal) == u"a"
+    assert model.headerData(2, orientation=Qt.Horizontal) == u"Unieke_Idcode"
+    assert model.headerData(3, orientation=Qt.Horizontal) == u"ü"
+    assert model.headerData(4, orientation=Qt.Horizontal) == u"b"
+    assert model.headerData(5, orientation=Qt.Horizontal) == u"1"
+
+def test_header_bom():
+    df = read_csv('spyder/widgets/variableexplorer/tests/test.csv',
+                  encoding='utf-8-sig')
```
ccordoba12: Please use for this file the one I'm attaching in this comment and call it `issue_2514.csv`. Also

1. There's no need to add the entire path of the file because `read_csv` tries to load it first from the current working directory.
2. Again, drop the encoding parameter.

[2514.txt](https://github.com/spyder-ide/spyder/files/680294/2514.txt)

spyder/widgets/variableexplorer/tests/test_dataframeeditor.py
```diff
@@ -174,6 +179,36 @@ def test_change_format_with_format_not_starting_with_percent(qtbot, monkeypatch)
...
+    editor = DataFrameEditor(None)
+    editor.setup_and_check(df)
+    model = editor.dataModel
+    assert model.headerData(0, orientation=Qt.Horizontal) == "Index"
```
ccordoba12: In this case we only need to test for the first header, not the rest of them :-)

spyder/widgets/variableexplorer/tests/test_dataframeeditor.py
dalthviz: @ccordoba12 I will change the `StringIO` and read a file directly 👍 , however the file that you point has as encoding `utf-8 BOM` (`'utf-8-sig'`) , taking this into account, is this the correct file to use in the encoding test or maybe I need to create another one with an unicode character different from the BOM one?

spyder/widgets/variableexplorer/tests/test_dataframeeditor.py
dalthviz: @ccordoba12 here I had to add that parameter to make the test work in Python 3. However I can add a validation to only put the `encoding` parameter if the version is 3, and let without that parameter the `read_csv` statement for Python 2.

ccordoba12: Yes, please do that :-)

spyder/widgets/variableexplorer/tests/test_dataframeeditor.py
ccordoba12: Ok, I see. Then you can get the cwd of a file being run in Python with

    os.path.dirname(os.path.realpath(__file__))

so you can add that as a constant at the top of this file.

spyder/widgets/variableexplorer/tests/test_dataframeeditor.py
```diff
@@ -174,6 +179,34 @@ def test_change_format_with_format_not_starting_with_percent(qtbot, monkeypatch)
     with qtbot.assertNotEmitted(editor.sig_option_changed):
         editor.change_format()
 
+def test_header_bom():
+    if PY2:
+        df = read_csv(os.path.join(FILES_PATH, 'issue_2514.csv'))
+    else:
+        df = read_csv(os.path.join(FILES_PATH, 'issue_2514.csv'),
+                      encoding='utf-8-sig')
```
ccordoba12: Given that tests are not passing without setting an encoding in PY3, I think we should remove the conditions we put in `headerData` to make things PY2 only.

@dalthviz, what do you think?

dalthviz: @ccordoba12 I think that maybe more than only remove the condition for PY2 will be needed since the `to_text_string` method does nothing if the object is an instance of `str` in PY3. So, should we add a way to handle the BOM character for PY3?

ccordoba12: Yes, I think we should. Please do it as part of this PR.

ccordoba12:
This looks really great, and I think adding tests helped us to improve our understanding of the situation considerably.

So there's nothing else for me than pressing the Green button :-)

PR #3906, opened 2017-01-01
PR: Don't show TextEditor if it fails to be initialized

### LINKED ISSUE #3673: Error when trying to view the contents of a binary string in the Variable Explorer for Python 2
If you evaluate this code in Python 2

```python
import string
dig_its = string.digits;
translate_digits = string.maketrans(dig_its,len(dig_its)*' ')
```

and then do a double-click on the variable called `translate_digits` in the Variable Explorer, you get this error in the internal console

```python-traceback
  File "/home/carlos/Projects/spyder/github-repo/spyder/widgets/variableexplorer/collectionseditor.py", line 483, in createEditor
    editor = TextEditor(value, key)
  File "/home/carlos/Projects/spyder/github-repo/spyder/widgets/variableexplorer/texteditor.py", line 44, in __init__
    text = to_text_string(text, 'utf8')
  File "/home/carlos/Projects/spyder/github-repo/spyder/py3compat.py", line 137, in to_text_string
    return unicode(obj, encoding)
  File "/home/carlos/.virtualenvs/test-ipy-stable/lib/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeDecodeError: 'utf8' codec can't decode byte 0x80 in position 128: invalid start byte
```

i think the solution is if we fail to initialize `TextEditor`, then we shouldn't show it.

@mariacamilaremolinagutierrez, `DataFrameEditor` has a method called `setup_and_check`, for you to look at. `TextEditor` should have a similar method to use it as `DataFrameEditor.setup_and_check` is also used.

mariacamilaremolinagutierrez:
Fixes #3673


ccordoba12:
@mariacamilaremolinagutierrez, this is not the solution I asked you to implement. What I want is for you to add a method called `setup_and_check` to the `TextEditor` class, and use it to verify if it's able to display strings passed to it. `setup_and_check` must return `True` is there are no problems and `False` otherwise.

Besides, please add a test for that method using the code I put on issue #3673 in

    spyder/widgets/variableexplorer/tests/test_texteditor.py

mariacamilaremolinagutierrez:
@ccordoba12 This is ready :)

PR #3908, opened 2017-01-02
PR: IPython Console adapts to Spyder color scheme

### LINKED ISSUE #916: Add color schemes to IPython console
_From [Pengkui....@gmail.com](https://code.google.com/u/109636005220771676064/) on 2012-01-22T23:45:51Z_

I wonder if Spyder can provide color scheme options for the IPython windows (as well as the Console windows)... I use black background on the main editor and the object inspector. And the white background of IPython and Console looks a bit annoying... Attached is how it looks on my computer.

**Attachment:** [spyder-color.jpg](http://code.google.com/p/spyderlib/issues/detail?id=916)

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=916_

mariacamilaremolinagutierrez:
Fixes #916 

ipython_color

### DIFF
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -988,6 +988,12 @@ class IPythonConsole(SpyderPluginWidget):
         if out_prompt_o:
             spy_cfg.JupyterWidget.out_prompt = out_prompt_o
 
+        # Style
+        syntax_style = self.get_syntax_style()
+        style_sheet = self.get_style_sheet()
+        spy_cfg.JupyterWidget.syntax_style = syntax_style
+        spy_cfg.JupyterWidget.style_sheet = style_sheet
+
         # Editor for %edit
         if CONF.get('main', 'single_instance'):
             spy_cfg.JupyterWidget.editor = self.set_editor()
@@ -1036,8 +1042,6 @@ class IPythonConsole(SpyderPluginWidget):
             pylab = self.get_option('pylab'),
             autoload_pylab = self.get_option('pylab/autoload'),
             sympy = self.get_option('symbolic_math'),
-            style_sheet = self.get_style_sheet(),
-            syntax_style = self.get_syntax_style(),
             show_banner = self.get_option('show_banner')
         )
 
@@ -1132,7 +1136,6 @@ class IPythonConsole(SpyderPluginWidget):
                 return ''
 
         selected_color_scheme = CONF.get('color_schemes', 'selected')
-        print(selected_color_scheme)
         color_scheme = get_color_scheme(selected_color_scheme)
 
         fon_c, fon_fw, fon_fs = color_scheme['normal']  # 'normal': ('#ffffff', False, False)
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -1046,11 +1046,28 @@ class IPythonConsole(SpyderPluginWidget):
             pylab = self.get_option('pylab'),
             autoload_pylab = self.get_option('pylab/autoload'),
             sympy = self.get_option('symbolic_math'),
-            light_color = self.get_option('light_color'),
+            style_sheet = self.get_style_sheet(),
             show_banner = self.get_option('show_banner')
         )
 
         return options
+
+    def get_style_sheet(self):
+        
+        sheet = """QPlainTextEdit, QTextEdit {
+                                              background-color: white;
+                                              color: navy ;
+                                              selection-background-color: #ccc
+                                              }
+                  .error { color: red; }
+                  .in-prompt { color: navy; }
+                  .in-prompt-number { font-weight: bold; }
+                  .out-prompt { color: darkred; }
+                  .out-prompt-number { font-weight: bold; }
+                  /* .inverted is used to highlight selected completion */
+                  .inverted { background-color: black ; color: white; }"""
+
+        return sheet
 
     def register_client(self, client, give_focus=True):
         """Register new client"""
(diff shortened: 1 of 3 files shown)


mariacamilaremolinagutierrez:
@ccordoba12 This PR still have lots of style corrections to be made so please ignore them for now, I will correct them when the content is approved. 

I have 2 main problems here (a big and a small one):

- big: when changing the color scheme, the new one is defined and set but for some reason qtconsole only updates the style_sheet and not the syntax_style (even though I call jyputer_widget's ´´´self._style_sheet_changed()´´´ and ´´´self._syntax_style_changed()´´´). So I think this is a qtconsole problem because is only refreshing half of it. However when restarting spyder it works good. But I know this is not the objective and what I find weird is that it does update a part. I also checked if maybe the syntax was not being updated correctly (but it is!).

- small: I want to sent the selection_background_color and the error_color in the style_sheet, however I can't find them in spyder. They are not in the color_scheme values. Do you know where they are? 

Apart from this I think everything is working. It took me a lot to figure out a way to set the syntax but I hope you like it :)

ccordoba12:
This is so nice!! Thanks a lot for working on this!

For now, one important comment: you did this PR against master and not 3.x, so you prefer this to go for Spyder 4?

ccordoba12:
About your problems:

1. What happens if you just restart an IPython console after changing schemes? Does it fix this problem?
2. What are those colors used for?

ccordoba12:
Another important thing to consider is what happens with colors for tracebacks. Did you try them with different schemes @mariacamilaremolinagutierrez?

mariacamilaremolinagutierrez:
@ccordoba12 Oh I'm sorry, I have no preference, I just did the PR to master because I thought new features went there. But if it is sooner for me is better :) About the comments:

1. If I just hit apply and ok in the settings, the editor gains the new scheme but the console doesn't. You have to close it (as it was before with the light and dark settings). After closing the console, when the next Console 2/A is opened, is that the consoles gets updated (but only stylesheet, not syntax). I tried restarting the kernel but that doesn't help. Is only after restarting spyder. I wanted to try this in qtconsole directly but once the console is running I can't seem to find a way of changing its display, only when calling ```python -m qtconsole --style 'name_of_syntax_style_file' ```. Do you know if there's a way? I could try adding a quick pop up in qtconsole to run these same updates with a different scheme to check if it is spyder or qt. How does this sound? 

2. ```selection_background_color``` is the color used when you select some text, I check and in all the spyder themes is always this blue: 

image

Do you know where in spyder is this blue color set? So I can put it in the style_sheet of the console.

And ```error_color``` I think is now part of spyder now that I realize, because the editor would never need to show this (this I think is related to the traceback you just mentioned). So what I'm thinking for it (as we don't set it for the themes) is taking the same color of another thing in the theme (to guarantee it is going to be visible). What do you think?

ccordoba12:
> I just did the PR to master because I thought new features went there. But if it is sooner for me is better :)

There are two types of changes:

1. Additions that break our current API or move/delete a lot of code (e.g. the code folding work of @rlaverde). They must done in master.
2. Additions that don't break our API (e.g. this case) or bugfixes. They can be done in 3.x.

So if you want to see this released sooner, you need to redo your PR against 3.x :-)

----

About the comments:

1. I really don't know how the style can be changed while qtconsole is running. This looks like new functionality we need to add to qtconsole :-)
2. I don't think we define `selection_background_color` but we could add it to our color schemes. About `error_color`, we have to take a look at it carefully because tracebacks are colored using other mechanisms in qtconsole.

----

I think this work is going to be delayed to 3.2 because there are still several things we need to look at and the feature freeze for 3.1 is almost here :-) But great job at making this work!

mariacamilaremolinagutierrez:
@ccordoba12:

About the branch, this latest commit will be the last in this PR. I will re do it but checking out from 3.x now. 

About the comments:

1. I made this PR to try to do it (https://github.com/jupyter/qtconsole/pull/184/) but as I comment there, I can only manage to change the color when creating a new tab (i.e. calling ```new_frontend_master()``` inside qtconsole/qtconsoleapp.py). In this new commit I am attempting to do the same. However calling ```create_new_client()``` inside spyder/plugins/ipythonconsole.py (with the new config options updated) does not have the same effect. Do you know what is the homologous for ```new_frontend_master()``` in spyder/plugins/ipythonconsole.py?

2. If we don't define never ```selection_background_color```, where is that color of blue coming from? About ```error_color```, I'll check it after I figure out these problems first.

ccordoba12:
1. Please check how we change color schemes in the Editor. I'm sure this can be done at the widget level, without restarting or opening new files. That code probably needs to be added to `qtconsole` too.
2. That blue must come from Qt, but I agree we should define it in our color schemes. However, that's work for another PR, so please don't add those changes here :-)

spyder/plugins/ipythonconsole.py
```diff
@@ -1046,12 +1042,163 @@ def additional_options(self):
...
+        ins_c, ins_fw, ins_fs = color_scheme['instance']  # 'instance': ('#be5f00', False, True)
+        instance_color =  ins_c
+        instance_font_weight = give_font_weight(ins_fw)
+        instance_font_style = give_font_style(ins_fs)
+        
+        import pygments
+        syntax_path = pygments.__file__.rpartition('/')[0] + '/styles/'
+        syntax_name = 'spyder'
+        syntax_filename = syntax_path + syntax_name + '.py'
+
```
goanpeca:
```
styles = 
{
  'Name.Class': [definition_font_style, definition_font_weight, definition_color,
}
```

create_style_strng(dic):
return str

with open(): {
   f.write(str)
}

PR #3913, opened 2017-01-04
change to setuptools so that we can build and distribute a wheel

James-Dengel:
a simple change intended to make packaging spyder as a wheel possible.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ import subprocess
 import sys
 import shutil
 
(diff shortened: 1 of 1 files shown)


ccordoba12:
Thanks for your contribution, but this is not really needed. You can just run

    python2 setup.py bdist_wheel

to create wheels for Spyder. That's how we create the wheels we upload to PyPi :-)

PR #3918, opened 2017-01-05
PR: Argparse execution in IPython Kernel is now successful

### LINKED ISSUE #3883: Error when executing argparse in IPython console
The following code snippet fails when it is executed with `F9` or in cell mode.

```
import argparse
parser = argparse.ArgumentParser()
args = parser.parse_args()
```

The reason is the arguments passed to the ipython console, in particular the `-f`:

```
>>> import sys
>>> print(sys.argv)
['/usr/local/lib/python3.5/dist-packages/spyder/utils/ipython/start_kernel.py', '-f', '/run/user/1000/jupyter/kernel-690215617a08.json']
```

1. Run code snippet above in ipython console
2. An error message is generated

The code snippet does work when executed in the python console. It also worked with previous versions of spyder. The issue is probably the same as #1557.

* Spyder Version: Spyder 4.0.0.dev0 
* Python Version: 3.5.2
* Operating system: Ubuntu

andfoy:
Fixes #3883

### DIFF
--- a/spyder/utils/ipython/start_kernel.py
+++ b/spyder/utils/ipython/start_kernel.py
@@ -34,7 +34,6 @@ init_session()
 
     return lines
 
-
 def kernel_config():
     """Create a config object with IPython kernel options"""
     from IPython.core.application import get_ipython_dir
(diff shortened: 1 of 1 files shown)


ccordoba12:
Works great, thanks @andfoy!

PR #3929, opened 2017-01-07
PR:  Save file with original encoding if possible

### LINKED ISSUE #3753: Saving a file is not respecting the encoding detected with chardet
After PR #3742, we're now detecting (and displaying) correctly files with encodings like Big5 (which is more than what Sublime and VSCode can do by default :-)

However, after doing changes to these files, Spyder is saving them as Unicode instead of respecting their encodings.

@rlaverde, please solve this one.

rlaverde:
Fixes #3753

---------------

`encodings.write()` will try to save the file with the original encoding as first option.

### DIFF
--- a/spyder/utils/encoding.py
+++ b/spyder/utils/encoding.py
@@ -171,6 +171,13 @@ def encode(text, orig_coding):
     if orig_coding == 'utf-8-bom':
         return BOM_UTF8 + text.encode("utf-8"), 'utf-8-bom'
     
+    # Try saving with original encoding
+    if orig_coding:
+        try:
+            return text.encode(orig_coding), orig_coding
+        except UnicodeError:
+            pass
+
     # Try declared coding spec
     coding = get_coding(text)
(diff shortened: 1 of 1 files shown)


ccordoba12:
Great job, now this works as expected :-)

spyder/utils/encoding.py
```diff
@@ -171,6 +171,13 @@ def encode(text, orig_coding):
     if orig_coding == 'utf-8-bom':
         return BOM_UTF8 + text.encode("utf-8"), 'utf-8-bom'
     
+    # Try saving with original encoding
+    if orig_coding:
+        try:
+            return text.encode(orig_coding), orig_coding
+        except UnicodeError:
```
ccordoba12: @rlaverde, I failed to see that in all other places we except in two errors: `UnicodeError` and `LookupError`. This shouldn't be an exception.

Please fix that in another PR.

PR #3933, opened 2017-01-07
PR: Catch LookupError when encoding in encode() function

rlaverde:
In #3929 I forget to catch LookUpError, when encoping text


ccordoba12:
Thanks @rlaverde for the quick response :-)

PR #3935, opened 2017-01-08
PR: Add validation for 'text/plain' key in all if clauses of handle_exec_method

### LINKED ISSUE #3932: Add verification for 'text/plain' key in all evaluations of the IPython console
We need to verify that `data` and `text/plain` exist in all `if` clauses of `handle_exec_method` (present in `widgets/ipythonconsole/shell.py`).

@dalthviz, please work on this one.

----

Please see this message for details:

https://groups.google.com/d/msg/spyderlib/UzrM0VP4sOU/Eh-JSjRPFQAJ

dalthviz:
FIxes #3932

### DIFF
--- a/spyder/widgets/ipythonconsole/shell.py
+++ b/spyder/widgets/ipythonconsole/shell.py
@@ -240,11 +240,14 @@ the sympy module (e.g. plot)
                 if 'get_namespace_view' in method:
                     if data is not None and 'text/plain' in data:
                         view = ast.literal_eval(data['text/plain'])
-                        self.sig_namespace_view.emit(view)
                     else:
                         view = None
-                elif 'get_var_properties' in method and 'text/plain' in data:
-                    properties = ast.literal_eval(data['text/plain'])
+                    self.sig_namespace_view.emit(view)
+                elif 'get_var_properties' in method:
+                    if data is not None and 'text/plain' in data:
+                        properties = ast.literal_eval(data['text/plain'])
+                    else:
+                        properties = None
                     self.sig_var_properties.emit(properties)
                 else:
                     if data is not None and 'text/plain' in data:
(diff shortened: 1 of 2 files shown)


spyder/widgets/ipythonconsole/shell.py
```diff
@@ -243,11 +243,11 @@ def handle_exec_method(self, msg):
                         self.sig_namespace_view.emit(view)
                     else:
                         view = None
-                elif 'get_var_properties' in method:
+                elif 'get_var_properties' in method and 'text/plain' in data:
```
ccordoba12: I think this is not right. It's clearer if you write it as:

```python
elif 'get_var_properties' in method:
    if data is not None and 'text/plain' in data:
        properties = ast.literal_eval(data['text/plain'])
        self.sig_var_properties.emit(properties)
```

At least it has the same form as the other `if` clauses here, which makes things easier to read.

dalthviz: We need to add the `else` statement making the `view = None` for this clause too, right?

ccordoba12: Right, but in this case we need to set `properties` to `None`. So this needs to be

```python
elif 'get_var_properties' in method:
    if data is not None and 'text/plain' in data:
        properties = ast.literal_eval(data['text/plain'])
    else:
        properties = None
    self.sig_var_properties.emit(properties)
```

However, this requires and additional change in the `set_var_properties` method of `widgtes/variableexplorer/namespacebrowser.py` so that it doesn't set `var_properties` if `properties` is `None`

ccordoba12: Also, please change the above `if` clause to be

```python
if 'get_namespace_view' in method:
    if data is not None and 'text/plain' in data:
        view = ast.literal_eval(data['text/plain'])
    else:
        view = None
    self.sig_namespace_view.emit(view)
```

That way we always emit a signal there.

ccordoba12:
Thanks @dalthviz!

PR #3946, opened 2017-01-10
PR: Make Editor tabs movable

rlaverde:
Fixes: #564
Fixes: #1170

------------
I made it really simple, just reused some code from #2372

### DIFF
--- a/spyder/plugins/editor.py
+++ b/spyder/plugins/editor.py
@@ -105,12 +105,9 @@ class EditorConfigPage(PluginConfigPage):
         
         interface_group = QGroupBox(_("Interface"))
         newcb = self.create_checkbox
-        fpsorting_box = newcb(_("Sort files according to full path"),
-                              'fullpath_sorting')
         showtabbar_box = newcb(_("Show tab bar"), 'show_tab_bar')
 
         interface_layout = QVBoxLayout()
-        interface_layout.addWidget(fpsorting_box)
         interface_layout.addWidget(showtabbar_box)
         interface_group.setLayout(interface_layout)
         
--- a/spyder/plugins/editor.py
+++ b/spyder/plugins/editor.py
@@ -1300,6 +1300,7 @@ class Editor(SpyderPluginWidget):
         editorstack.sig_prev_edit_pos.connect(self.go_to_last_edit_location)
         editorstack.sig_prev_cursor.connect(self.go_to_previous_cursor_position)
         editorstack.sig_next_cursor.connect(self.go_to_next_cursor_position)
+        editorstack.tabs.tabBar().tabMoved.connect(self.move_editorstack_data)
 
     def unregister_editorstack(self, editorstack):
         """Removing editorstack only if it's not the last remaining"""
@@ -2638,3 +2639,23 @@ class Editor(SpyderPluginWidget):
         """Change the value of create_new_file_if_empty"""
         for editorstack in self.editorstacks:
             editorstack.create_new_file_if_empty = value
+
+    def move_editorstack_data(self, start, end):
+        """
+        Move editorstack.data to be synchronized when tabs are moved.
+        """
+        if start < 0 or end < 0:
+            return
+        else:
+            steps = abs(end - start)
+            sign = (end - start) // steps
+
+            for editorstack in self.editorstacks :
+                data = editorstack.data
+                editorstack.blockSignals(True)
+
+                for i in range(start, end, sign*1):
+                    data[i], data[i+sign] = data[i+sign], data[i]
+
+                editorstack.blockSignals(False)
+                editorstack.refresh()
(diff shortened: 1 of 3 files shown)


ccordoba12:
This is great! Thanks a lot for it :-)

@goanpeca, please review it.

spyder/plugins/editor.py
```diff
@@ -2643,3 +2641,23 @@ def set_create_new_file_if_empty(self, value):
         """Change the value of create_new_file_if_empty"""
         for editorstack in self.editorstacks:
             editorstack.create_new_file_if_empty = value
+
+    def move_editorstack_data(self, start, end):
+        """
+        Move editorstack.data to be synchronized when tabs are moved.
```
goanpeca: Single line is shorter than 79?

spyder/plugins/editor.py
```diff
@@ -2643,3 +2641,23 @@ def set_create_new_file_if_empty(self, value):
...
+        else:
+            steps = abs(end - start)
+            sign = (end - start) // steps
+
+            for editorstack in self.editorstacks :
+                data = editorstack.data
+                editorstack.blockSignals(True)
+
+                for i in range(start, end, sign*1):
+                    data[i], data[i+sign] = data[i+sign], data[i]
```
goanpeca: pace around `+` sign ?

spyder/plugins/editor.py
```diff
@@ -2643,3 +2641,23 @@ def set_create_new_file_if_empty(self, value):
...
+
+    def move_editorstack_data(self, start, end):
+        """
+        Move editorstack.data to be synchronized when tabs are moved.
+        """
+        if start < 0 or end < 0:
+            return
+        else:
+            steps = abs(end - start)
+            sign = (end - start) // steps
```
goanpeca: maybe sign is not the best name hmmm

`direction` or `step` maybe? and adding a comment saying +1 for right, -1 for left?

spyder/plugins/editor.py
goanpeca: It would be probably better to invert the position of the for so that we move all editors at the same time 

```python

for i in range(start, end, sign * 1):
    for editorstack in self.editorstacks :
        data = editorstack.data
        editorstack.blockSignals(True)

        data[i], data[i + sign] = data[i + sign], data[i]

        editorstack.blockSignals(False)
        editorstack.refresh()
```

It feels that in case there is an error it would leave the UI in a better state.

Thoughts @rlaverde, @ccordoba12 ?

spyder/widgets/editor.py
```diff
@@ -1005,16 +1005,8 @@ def get_tab_tip(self, filename, is_modified=None, is_readonly=None):
             else:
                 return text % (osp.basename(filename), osp.dirname(filename))
 
-    def __get_sorting_func(self):
```
goanpeca: Is this a method we would like to preserve in case the user want to sort (by adding an option in the editor menu?)

@ccordoba12 ?

spyder/plugins/editor.py
rlaverde: Yes, I will change it

spyder/plugins/editor.py
rlaverde: I think `direction` is better.

spyder/plugins/editor.py
rlaverde: I don't think that space is really needed

and I think that inverting the order maybe could cause some undesired effect, because signals won't be blocked all the time while modifying an editorstack

also It will be a little overhead to call `editorstack.blockSignals()`and `editorstack.refresh()` so many times.

rlaverde:
This have a glitch related with #1170, It have to be solved first.

spyder/widgets/editor.py
ccordoba12: I don't think so, the less options we have the better.

goanpeca:
You found the culprit @rlaverde for the moving tab glitch?

rlaverde:
> You found the culprit @rlaverde for the moving tab glitch?

Yes I fixed it, this line is causing the glitch https://github.com/spyder-ide/spyder/blob/master/spyder/widgets/editor.py#L1460

I implement an workaround in the last commit, but I'm not sure if it's the best solution.

rlaverde:
TODO: This is causing strange behaviors with changing files shortcuts (ctrl+tab) and closing opening files, because these implementation assume an strict order in `stackeditors`

rlaverde:
> TODO: This is causing strange behaviors with changing files shortcuts (ctrl+tab) and closing opening files, because these implementation assume an strict order in stackeditors

I was wrong, that was the behavior previous my changes, Ctrl+Tab, Ctrl+Shit+Tab have a really strange behavior, I think this is ready, and I'll implement MRU(#43) in another PR

goanpeca:
@rlaverde thanks!

ccordoba12:
@rlaverde, please do a rebase to see if that fixes the errors in Travis.

rlaverde:
> please do a rebase to see if that fixes the errors in Travis.

Done :smile:

Issue #3949, opened 2017-01-12
F9 broken in latest git master


dhirschfeld:
Pressing F9 in the editor gives the below traceback in the internal console:
```
>>> Traceback (most recent call last):
  File "C:\dev\packages\spyder\spyder\app\mainwindow.py", line 2415, in execute_in_external_console
    console.execute_code(lines)
TypeError: execute_code() missing 1 required positional argument: 'clear_variables'
```
![image](https://cloud.githubusercontent.com/assets/881019/21878445/18c763e0-d8de-11e6-8151-216783d600db.png)

dhirschfeld:
https://github.com/spyder-ide/spyder/blob/master/spyder/app/mainwindow.py#L2415
https://github.com/spyder-ide/spyder/blob/master/spyder/plugins/ipythonconsole.py#L848

goanpeca:
@rlaverde please take this one

ccordoba12:
This has to be done against 3.x not master. @rlaverde, please make `clear_variables=False` at the line pointed by @dhirschfeld in `ipythonconsole.py`.

PR #3951, opened 2017-01-12
PR: Add default value to clear_variables=False in execute code method

rlaverde:
Fixes: #3949


ccordoba12:
Thanks a lot @rlaverde!

PR #3952, opened 2017-01-12
PR: Make File Switcher leave focus on its last selected file when it loses focus

### LINKED ISSUE #3926: File switcher doesn't leave focus in its last selected file when it looses focus
When you select a file in the File switcher and then use the mouse to give focus to the Editor or the IPython console, then the switcher doesn't leave focus in that file but instead return focus to the file that had focus when the user *first* pressed `Ctrl+P`.

This is counterintuitive and must be fixed. @andfoy, please work on it.

andfoy:
Fixes #3926


ccordoba12:
@andfoy, could you try to add a test for this one?

PR #3958, opened 2017-01-14
PR: Comment lines taking into account code indentation and pep8

### LINKED ISSUE #1785: Spyder comments (Ctrl+1 and Ctrl+4) generate pep8 warnings

Spyder Version:  2.2.5
Python Version:  2.7.5+
matplotlib >=1.0:  1.2.1 (OK)
pep8 >=0.6      :  1.5.4 (OK)
pyflakes >=0.5.0:  0.8.1 (OK)
pylint >=0.25   :  1.1.0 (OK)
rope >=0.9.2    :  0.9.4 (OK)
sphinx >=0.6.6  :  1.2.2 (OK)
sympy >=0.7.0   :  0.7.3 (OK) 

         #somefunc(arg1, arg2)

         # somefunc(arg1, arg2)

### LINKED ISSUE #2845: Comment out lines according to indentation
# Description

Commented lines are commented always on column 0, no matter what the indent. This results in the fold detector "incorrectly" detecting a fold where lines are commented out. Although not really an issue I would suggest to adapt the code to comment lines (Comment/Uncomment Ctl+1) to put the '#' on the indented level. This would allow for much cleaner code (this is my opinion) and would eliminate this 'bug'.

dalthviz:
Fixes #2845 
Fixes #1785

Edit:
Takes into account some ideas from @bcolsen and @Ericvulpi, thanks for your help in this one :)


dalthviz:
A preview:

commented_lines

goanpeca:
Lets add a space between the `#` and the code so it is valid pep8 and ensure that we check if there is a space or not also when de-denting.

So that

```python
# def hello():
    # pass

#def hello():
    #pass

```

Both work when removing the comment

ccordoba12:
@bcolsen, what do you think about this one?

@rlaverde, would this help code folding to work correctly when big chunks are commented?

rlaverde:
> would this help code folding to work correctly when big chunks are commented?

Hmm, this was the original solution, but we can't expect that lines will be always comment out with spyder shortcut, because of that I make code folding ignore commented lines in #4728, they could have an arbitrary indentation and that will break codefolding

Although I really like this change, IMO commented out lines at the first of the line is visually annoying. also other editors comment out lines according identation

bcolsen:
I really like this style because it keeps the flow of the code even with comments in the middle.

The code is good at undoing the comments even crazy stuff like this:
```
# if __name__=="__main__":
#     app = QApplication(sys.argv)
    #btn = Test()
#     btn.show()
#     app.exec_()
```
However, it is not backwards compatible with the previous block comments, and can break code that has been manually commented out at the beginning of the line.  These can be hard to spot and make whitespace management a difficult chore.

But I think this problem can be fixed.

The code fails on this:
```
#if __name__=="__main__":
#    app = QApplication(sys.argv)
#    btn = Test()
#    btn.show()
#    app.exec_()
``` 
It deletes the spaces in front of the indented section leaving 3 spaces instead of for.

Should be possible to tell whether a space needs to be deleted or not by:
1. Count the number of spaces
2. Mod the count by number of spaces set for indents in the file
3. If mod equals 1 remove the extra space
4. Else don't remove any spaces

This wouldn't work perfectly for long line continuations, but those won't break your programs like other indents do.

ccordoba12:
@dalthviz, please try to fix the use case shown by @bcolsen.

ccordoba12:
@dalthviz, great work so far. Please squash all commits referring to `Testing` and also make sure that this PR also fixes issue #1785.

ccordoba12:
You can take a look at PR #4720 to see how to do it.

dalthviz:
Preview:

comments

ccordoba12:
@bcolsen, could you help us to manually test this PR and see if it passes all your use cases? Thanks!

bcolsen:
> @bcolsen, could you help us to manually test this PR and see if it passes all your use cases? Thanks!

Ok I think this might be more than you asked for but I spent a few hours looking at this.  Here are some issues I found.

## Issue 1

If I highlight a section right to the beginning of a line, when I press ctrl-1 it adds the last character of the previous line to the active selection.

lineabove

## Issue 2

Comment that added the `#` and a space at the beginning of a line are not uncommented properly. This would be unlikely to happen except that block comments work this way in this PR.

I know this is pain fully bad, but it could happen.

block comments

This is also the default block comments in pycharm:

block comments

## Issue 3

I will preface this with the fact that I don't really care about compatibility of commenting out large blocks of code with pep8. I don't think that the code style guide would think commenting out chunks of code is a good code style. I mostly use code commenting as a quick way of reusing and testing code to make quick data analysis scripts. I think this my be the most common use of spyder.

tldr; I don't think this code change should make more pep8 problems

pep8_errors

# Solutions

Solution number here for ease of reference. They do not correspond to the issue numbers.

## Solution 1
Drop ctrl-4/5 style block comments from spyder. They are not standard and do the same as ctrl-1 comments. This will make things easier and you can use ctrl-1 to undo legacy block comments except deleting the top and bottom lines(not too bad..it doesn't make broken code)

## Solution 2
I think that the block comment style in pycharm makes it easy to see the block of code as opposed to the current indented comments(which pep8 also doesn't like)

pychar
## Solution 3
I don't know how the current code works for removing spaces but you can check that the line is even or odd after the `#` is removed. If it's odd remove a space(I haven't seen anyone on the 3-space indent band wagon but they should stop anyways) 

## Bonus: 

Watch pycharm eat my code(it always removes a space. This is incompatible with manual commenting..I think we can do better):

pycharm_eat

dalthviz:
@ccordoba12 @bcolsen this is a preview:

comments


@bcolsen could you test it please? thanks :)

bcolsen:
Wow that's great it's much more robust pycharm and I like the look of it. I gave it quite a bit of abuse and it held up nicely.  I like the unblocking of block comments that's slick.

Now onto the bugs ;)

## Issue 1

If I highlight the beginning of a line and comment out the section I start loosing selection off the end of the selection.  It looks like the number of characters decreased by twice the number of lines that are highlighted in the selection.

Highlight:
sc_highlight
Comment:
sc_comment
Uncomment and Comment again:
sc_comment_again

## Issue 2

You may have noticed in the screen shots above that the comments are not in line with the text this seems to happen when a line is blank. If you have a space in the line it skips it but not if its blank.

## Issue 3

if you uncomment a block comment with at commented line light highlighted it completely uncomments the line instead of only removing the block comments.

sc_block_comments

sc_unblock

## Issue 4

The new unblocking of the block comments seems incompatible with the old blocks.  I'm not sure how the blocks are detected so I don't know if it's possible.  It would be nice, but not the end of the world either way.

dalthviz:
@bcolsen thanks again for the good feedback :+1:, could you test the latest changes? I think this is almost finished :)

bcolsen:
Issue 1 and 4 are solved now. :+1:

I'm back on my laptop so I made some gifs of the other two.

## Issue 2

The blank lines getting skipped now but they are still counting as smallest indent so the comments are still on the far left.

blank lines

## Issue 3

The comment in block comments are still getting uncommented when I use `ctrl-1`. If you press undo after you remove the block comments the original comment is restored and undo again to bring back the block comment.

unblock_comments

rlaverde:
Maybe to resolve these issues we could consider two shortcuts for comment/uncomment code, instead of only one. (i.e `Ctrl+1` will always comment code, and `Ctrl+Shift+1` will uncomment code )

bcolsen:
I was actually thinking the same thing

Crtl-shift-1 is a little awkward but it makes sense

dalthviz:
@bcolsen I made some more changes can you test again? :). About the reasons of the issues still there: in the case of the unblockcomment I was not returning in the method and in the case of the indentation level of the comment it was a missing validation. 

About the suggestion of @rlaverde (by the way, thanks for the feedback :+1:), thinking beyond the issues that @bcolsen helped to find, you guys think that could be helpful to have a different shortcut to manage the comment/uncomment? What do you think @ccordoba12?

bcolsen:
@dalthviz Everything looks great to me here! I couldn't find anything that would I would normally do or end up with that can break. :+1: 

As for the `ctrl-shift-1` for removing comments:  

## Keep the current commenting tool the same

I think `ctrl-1` is very easy as it is now and should remain the same, removing comments if only all the selected lines are commented. The `ctrl-1` code already comments mixed commented and uncommented code. I can't think of a time I would want to double or triple comment something that is already commented out.

## Add an option for just uncommenting

I have had times when I wanted to uncomment a whole section that had mixed commented and uncommented code.  A `ctrl-shift-1` uncomment could be very handy in these situations.

For example, pressing `ctrl-shift-1` twice would completely uncomment this code with only one selection. With the current commenting code this would require three selections and three key presses

```python
# if language.lower() in value:
#     self.supported_language = True
#     # sh_class, comment_string, CFMatch = self.LANGUAGES[key]
#     # self.comment_string = comment_string
    if key in CELL_LANGUAGES:
        self.supported_cell_language = True
        self.cell_separators = CELL_LANGUAGES[key]
#     if CFMatch is None:
#         self.classfunc_match = None
#     else:
#         self.classfunc_match = CFMatch()
```

This way we can keep the commenting tool "smart"(or contextual) like it is now about when to comment or uncomment, as well as give the option to take an arbitrary block of code and uncomment all of it.

dalthviz:
Thanks a lot @bcolsen for all your help with this :+1:

About the new shortcut for only uncommenting, from that point of view could be interesting, however I think that it could be left for another issue @ccordoba12 @rlaverde what do you guys think?

ccordoba12:
@bcolsen, thanks a lot for your careful and detailed review! It helped us a lot to improve this!

About adding an option to just uncommenting, it looks like a reasonable idea. Please open a new issue about it, so we can continue our discussion there.

spyder/widgets/sourcecode/codeeditor.py
```diff
@@ -1590,28 +1629,68 @@ def remove_prefix(self, prefix):
...
         else:
             # Remove prefix from current line
             cursor.movePosition(QTextCursor.StartOfBlock)
             line_text = to_text_string(cursor.block().text())
-            if (prefix.strip() and line_text.lstrip().startswith(prefix)
-                or line_text.startswith(prefix)):
+            self.__remove_prefix(prefix, cursor, line_text)
+
+    def __remove_prefix(self, prefix, cursor, line_text):
+        """Handle the remove of the prefix for a single line."""
```
ccordoba12: Please change this to

    Handle the removal of prefix comment for a single line

spyder/widgets/sourcecode/codeeditor.py
```diff
@@ -1960,14 +2039,21 @@ def is_comment(self, block):
...
         """Uncomment current line or selection."""
-        self.remove_prefix(self.comment_string)
-
-    def __blockcomment_bar(self):
-        return self.comment_string + ' ' + '=' * (78 - len(self.comment_string))
+        blockcomment = self.unblockcomment()
+        if not blockcomment:
+            self.remove_prefix(self.comment_string)
+
+    def __blockcomment_bar(self, compatibility=False):
```
ccordoba12: Please add a docstring to this method.

ccordoba12:
@dalthviz, I left a couple of minor comments, then I'll merge ;-)

PR #3994, opened 2017-01-18
PR: Use QRegularExpression only for Qt 5.5+

### LINKED ISSUE #3989: Spyder won't start after upgrade to 3.1.0 due to an error with QRegularExpression

1. Start spyder from terminal

I expect spyder to start, instead I get an error:

> spyder
QML debugging is enabled. Only use this in a safe environment.
  File "/usr/lib/python2.7/site-packages/spyder/app/mainwindow.py", line 2979, in main
  File "/usr/lib/python2.7/site-packages/spyder/app/mainwindow.py", line 2883, in run_spyder
    main.setup()
  File "/usr/lib/python2.7/site-packages/spyder/app/mainwindow.py", line 780, in setup
    message=_("Spyder Internal Console\n\n"
  File "/usr/lib/python2.7/site-packages/spyder/plugins/console.py", line 79, in __init__
    self.find_widget.set_editor(self.shell)
  File "/usr/lib/python2.7/site-packages/spyder/widgets/findreplace.py", line 280, in set_editor
    self.refresh()
  File "/usr/lib/python2.7/site-packages/spyder/widgets/findreplace.py", line 259, in refresh
    self.find()
  File "/usr/lib/python2.7/site-packages/spyder/widgets/findreplace.py", line 326, in find
    self.editor.find_text('')
  File "/usr/lib/python2.7/site-packages/spyder/widgets/mixins.py", line 513, in find_text
    found_cursor = self.document().find(pattern, cursor, findflag)
TypeError: arguments did not match any overloaded call:
  QTextDocument.find(str, int position=0, QTextDocument.FindFlags options=0): argument 1 has unexpected type 'QRegularExpression'
  QTextDocument.find(QRegExp, int position=0, QTextDocument.FindFlags options=0): argument 1 has unexpected type 'QRegularExpression'
  QTextDocument.find(str, QTextCursor, QTextDocument.FindFlags options=0): argument 1 has unexpected type 'QRegularExpression'
  QTextDocument.find(QRegExp, QTextCursor, QTextDocument.FindFlags options=0): argument 1 has unexpected type 'QRegularExpression'

* Spyder Version: 3.1.0
* Python Version: 2.7.12
* Operating system: OpenSuse 13.2

## Dependencies

and paste the contents below:
I can't get spyder to start, so ...

rlaverde:
Fixes #3989

### DIFF
--- a/spyder/widgets/mixins.py
+++ b/spyder/widgets/mixins.py
@@ -24,22 +24,21 @@ from qtpy.QtGui import QCursor, QTextCursor, QTextDocument
 from qtpy.QtWidgets import QApplication, QToolTip
 from qtpy import QT_VERSION
 
-QT55_VERSION = tuple(int(x) for x in QT_VERSION.split('.')) >= (5, 5)
-
-if QT55_VERSION:
-    from qtpy.QtCore import QRegularExpression
-else:
-    from qtpy.QtCore import QRegExp
-
 # Local imports
 from spyder.config.base import _
 from spyder.py3compat import is_text_string, to_text_string, u
-from spyder.utils import encoding, sourcecode
+from spyder.utils import encoding, sourcecode, programs
 from spyder.utils.dochelpers import (getargspecfromtext, getobj,
                                      getsignaturefromtext)
 from spyder.utils.misc import get_error_match
 from spyder.widgets.arraybuilder import NumpyArrayDialog
 
+QT55_VERSION = programs.check_version(QT_VERSION, "5.5", ">=")
+
+if QT55_VERSION:
+    from qtpy.QtCore import QRegularExpression
+else:
+    from qtpy.QtCore import QRegExp
 
 HISTORY_FILENAMES = []
 
--- a/spyder/widgets/mixins.py
+++ b/spyder/widgets/mixins.py
@@ -22,9 +22,11 @@ import textwrap
 from qtpy.QtCore import QPoint, Qt
 from qtpy.QtGui import QCursor, QTextCursor, QTextDocument
 from qtpy.QtWidgets import QApplication, QToolTip
-from qtpy import PYQT5
+from qtpy import PYQT_VERSION
 
-if PYQT5:
+PYQT55_VERSION = tuple(int(x) for x in PYQT_VERSION.split('.')) >= (5, 5)
+
+if PYQT55_VERSION:
     from qtpy.QtCore import QRegularExpression
 else:
     from qtpy.QtCore import QRegExp
@@ -490,7 +492,7 @@ class BaseEditMixin(object):
             moves += [QTextCursor.End]
         if not regexp:
             text = re.escape(to_text_string(text))
-        if PYQT5:
+        if PYQT55_VERSION:
             pattern = QRegularExpression(r"\b{}\b".format(text) if words else
                                          text)
             if case:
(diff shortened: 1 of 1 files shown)


spyder/widgets/mixins.py
```diff
@@ -22,9 +22,11 @@
 from qtpy.QtCore import QPoint, Qt
 from qtpy.QtGui import QCursor, QTextCursor, QTextDocument
 from qtpy.QtWidgets import QApplication, QToolTip
-from qtpy import PYQT5
+from qtpy import PYQT_VERSION
 
-if PYQT5:
+PYQT_VERSION = float(PYQT_VERSION)
+
```
goanpeca: Maybe its cleaner to define

`PYQT55_VERSION = float(PYQT_VERSION) > 5.5`

And use that in the conditions below?

rlaverde: hmm no, It's given an error due versions like 4.11.4

I'll implement:
`MINOR_VERSION = int(PYQT_VERSION.split('.')[1])`

goanpeca: Well.. sure, I meant using a single global thing, not how you were comparing versions (which, yes is probably wrong)

No need

You can compare tuples so use that
```python
a = (5, 5)
b = (5, 5, 1)

a > b
b < a
```

goanpeca:
Well.. sure, I meant using a single global thing, not how you were comparing versions (which, yes is probably wrong)

No need

You can compare tuples also
```python
a = (5, 5)
b = (5, 5, 1)

a > b
b < a
```


But I guess what you did now works also

rlaverde:
@goanpeca I followed your both advices, I think this is ready :)

spyder/widgets/mixins.py
```diff
@@ -22,9 +22,11 @@
 from qtpy.QtCore import QPoint, Qt
 from qtpy.QtGui import QCursor, QTextCursor, QTextDocument
 from qtpy.QtWidgets import QApplication, QToolTip
-from qtpy import PYQT5
+from qtpy import PYQT_VERSION
 
-if PYQT5:
+PYQT55_VERSION = tuple(int(x) for x in PYQT_VERSION.split('.')) >= (5, 5)
```
ccordoba12: @rlaverde, although Qt and PyQt versions match (e.g. Qt 5.5 and PyQt 5.5 are installed at the same time), that's not always the case. For example, Ubuntu 16.10 comes with Qt 5.6 and PyQt 5.7.

So please use QT_VERSION instead of PYQT_VERSION here. We define QT_VERSION here:

https://github.com/spyder-ide/qtpy/blob/master/qtpy/__init__.py#L87

spyder/widgets/mixins.py
```diff
@@ -22,9 +22,11 @@
 from qtpy.QtCore import QPoint, Qt
 from qtpy.QtGui import QCursor, QTextCursor, QTextDocument
 from qtpy.QtWidgets import QApplication, QToolTip
-from qtpy import PYQT5
+from qtpy import QT_VERSION
 
-if PYQT5:
+QT55_VERSION = tuple(int(x) for x in QT_VERSION.split('.')) >= (5, 5)
```
ccordoba12: Please use `check_version` from `utils/programs.py` to do the comparison here

ccordoba12:
Thanks @rlaverde!

PR #3998, opened 2017-01-18
PR: Normalize all line endings to LF

goanpeca:


ccordoba12:
Closing, again. Perhaps we can do this after we merge split-plugins.

Issue #3992, opened 2017-01-18
Remove unnecessary newlines in run cell


talkaminker:
When running a cell, the cell that is actually ran in the ipython contains 2 redundant newlines in the end.
this is inconvenient  because it reduces the amount of text you see in ipython console.
 here is a fix:
in: `spyder/widgets/sourcecode/base.py `in the function `get_cell_as_executable_code`
instead of 
`return text`
write:
`return text.rstrip()`

This will also remove endlines added by the user which I think is actually a desired behavior anyway (for example, I tend to add newlines just to visually separate between one cell and the next one)

If you think this is good, I can create a pull request with the fix.

Note: right now, the newlines which are in the beginning of the cell are stripped anyway, so removing the trailing newlines just seems natural.

ccordoba12:
I think at least one new line is needed to get the console to evaluate the cell.

But if you make that work, please send us a pull request :-)

talkaminker:
I tested it before opening the issue, it works. I will add a pull request

goanpeca:
@ccordoba12 this should beagainst what branch ?

ccordoba12:
It's a very simple change, so I'd say `3.1.x`.

----

We need to write a guide about how we manage our branches so that it's easier for external contributors to work with us.

PR #3993, opened 2017-01-18
remove unnecessary newlines in run cell

talkaminker:
Fixes https://github.com/spyder-ide/spyder/issues/3992

PR #4006, opened 2017-01-19
PR: Remove unnecessary new lines when running cells

talkaminker:
Fixes https://github.com/spyder-ide/spyder/issues/3992

This is a copy of another PR (https://github.com/spyder-ide/spyder/pull/3993) which I closed because it was against master. this one is against 3.1.X


ccordoba12:
Thanks for your patience @talkaminker!

PR #4019, opened 2017-01-21
PR: Handle errors while debugging code when importing tqdm

### LINKED ISSUE #4003: Debugging seems to break with recent versions of Spyder and tqdm
Debugging seems to break with recent versions of Spyder and tqdm. The internal console gives the following output:

```
  File "D:\code\thirdparty\spyder\spyder\widgets\ipythonconsole\namespacebrowser.py", line 49, in <lambda>
    self.namespacebrowser.process_remote_view(data))
  File "D:\code\thirdparty\spyder\spyder\widgets\variableexplorer\namespacebrowser.py", line 232, in process_remote_view
    self.set_data(remote_view)
  File "D:\code\thirdparty\spyder\spyder\widgets\variableexplorer\namespacebrowser.py", line 243, in set_data
    self.editor.set_data(data)
  File "D:\code\thirdparty\spyder\spyder\widgets\variableexplorer\collectionseditor.py", line 830, in set_data
    self.model.set_data(data, self.dictfilter)
  File "D:\code\thirdparty\spyder\spyder\widgets\variableexplorer\collectionseditor.py", line 170, in set_data
    self.set_size_and_type()
  File "D:\code\thirdparty\spyder\spyder\widgets\variableexplorer\collectionseditor.py", line 185, in set_size_and_type
    for index in range(start, stop) ]
TypeError: 'int' object has no attribute '__getitem__'
```

1. Make a simple script (e.g.):

```
import tqdm

1 + 1

b = 3 

print "Awesome"

c = 4

print "Awesome as well"
```

2. Put a breakpoint anywhere (e.g. print "Awesome")
3. Try to debug and press continue.

Expect output would be to stop at the line and show a short print of the code:

```
     12 b = 3
     13 
2--> 14 print "Awesome"
     15 
     16 c = 4
```

instead you get:

```
{'e': {'color': '#0000ff', 'view': '2.718281828459045', 'type': u'float', 'size': 1}, 'euler_gamma': {'color': '#0000ff', 'view': '0.5772156649015329', 'type': u'float', 'size': 1}, 'pi': {'color': '#0000ff', 'view': '3.141592653589793', 'type': u'float', 'size': 1}, 'b': {'color': '#0000ff', 'view': '3', 'type': u'int', 'size': 1}}
```

* Spyder Version: 3.1.0
* Python Version: 2.7.13
* Operating system: Windows 10 x64

## Dependencies

and paste the contents below:

nbconvert >=4.0  :  5.0.0 (OK)
pyflakes >=0.5.0 :  1.5.0 (OK)
rope >=0.9.4     :  0.10.3 (OK)
sympy >=0.7.3    :  1.0 (OK)

dalthviz:
Fixes #4003 

-------
~~For now this only helps to prevent showing errors. However the debug function is not working properly with the `tqdm` import.~~

~~Now the debug function works but without showing colors.~~

This now only handles not showing errors when importing tqdm in debug mode. Also restarts the kernel to restore the debug functionality in the current kernel.


ccordoba12:
I like this, good job @dalthviz!

What's still missing from this PR?

dalthviz:
@ccordoba12 the thing with this is that the changes in this PR only ensure that no error is thrown. However the debug output is not correct after you run the import line, this a preview of what you see:

imagen

Also, something strange is that if you press `Enter` in the console you can see the correct output but without colors:

imagen

spyder/widgets/ipythonconsole/debugging.py
```diff
@@ -12,10 +12,12 @@
 import ast
 
 from qtpy.QtCore import QEventLoop
+from qtpy.QtWidgets import QMessageBox
```
ccordoba12: This line is not needed anymore

spyder/widgets/ipythonconsole/debugging.py
```diff
@@ -12,10 +12,12 @@
...
 
 from qtconsole.rich_jupyter_widget import RichJupyterWidget
 
 from spyder.py3compat import to_text_string
+from spyder.config.base import _
```
ccordoba12: This line is also not needed now :-)

ccordoba12:
This is looking really good. I think the only missing thing is a test for this ;-)

spyder/app/tests/debug.py
```diff
@@ -0,0 +1,3 @@
+# -*- coding: utf-8 -*-
+import tqdm
+a = 1
```
ccordoba12: Please don't add this file and instead evaluate this two lines in the IPython console during the test. That would avoid adding `tqdm` as a dependency in Travis (which is not really needed).

spyder/app/tests/test_mainwindow.py
```diff
@@ -80,6 +80,41 @@ def close_widget():
 #==============================================================================
 @flaky(max_runs=10)
 @pytest.mark.skipif(os.name == 'nt', reason="It times out sometimes on Windows")
+def test_debug_error_code(main_window, qtbot):
+    """Test that the debug in IPython console handles errors in stdout."""
+    # Wait until the window is fully up
+    shell = main_window.ipyconsole.get_current_shellwidget()
+    control = shell._control
+    qtbot.waitUntil(lambda: shell._prompt_html is not None, timeout=SHELL_TIMEOUT)
+
```
ccordoba12: You can evaluate the lines of `debug.py` here, like this

```python
lines = ['import tqdm', 'a = 1']
for line in lines:
    shell.execute(line)
    qtbot.wait(200)
```

spyder/app/tests/test_mainwindow.py
```diff
@@ -80,6 +80,41 @@ def close_widget():
...
+    debug_next_action = main_window.debug_toolbar_actions[1]
+    debug_next_button = main_window.debug_toolbar.widgetForAction(debug_next_action)
+    qtbot.mouseClick(debug_next_button, Qt.LeftButton)
+    qtbot.wait(3000)
+    
+    # Run the current line
+    qtbot.mouseClick(debug_next_button, Qt.LeftButton)
+    qtbot.wait(3000)
+
+    assert "Kernel restarting because an error raised while debugging (stdout change)" in control.toPlainText()
```
ccordoba12: Please make this `assert` simpler, i.e. instead of verifying that this text is present in the console, just verify that there's no `a` value in the console (because the kernel was restarted :-).

For that, please modify the `get_value` method of `SpyderKernel` (in `utils/ipython/spyder_kernel.py`), like this

```python
def get_value(self, name):
   """Get the value of a variable"""
    ns = self._get_current_namespace()
    try:
        value = ns[name]
    except KeyError:
        value = (name, '__spy_missing_value__')
    publish_data({'__spy_data__': value})
```

and then change the assert line above like this

```python
assert shell.get_value('a') == ('a', '__spy_missing_value__')
```

spyder/widgets/ipythonconsole/client.py
```diff
@@ -323,33 +326,62 @@ def restart_kernel(self):
...
+                        sw._append_plain_text(
+                            _('Error restarting kernel: %s\n') % e,
+                            before_prompt=True
+                        )
+                    else:
+                        sw.reset(clear=False)
+                        sw._append_html(_("<br>Kernel restarting because an "
+                                          "error raised while debugging "
+                                          "(stdout change)\n<hr><br>"),
+                                        before_prompt=False)
```
ccordoba12: Please change this message to

> Restarting kernel because an error occurred while debugging

spyder/widgets/ipythonconsole/client.py
```diff
@@ -323,33 +326,62 @@ def restart_kernel(self):
...
+                        sw._append_plain_text(
+                            _('Error restarting kernel: %s\n') % e,
+                            before_prompt=True
+                        )
+                    else:
+                        sw.reset(clear=True)
+                        if sw._input_none:
+                            sw._append_html(_("<br>Kernel restarting because an "
+                                              "error raised while debugging "
+                                              "(stdout change)\n<hr><br>"),
```
ccordoba12: Also change this message with the same one above.

ccordoba12:
@dalthviz, I removed all commit related to tests in this PR. The thing this error doesn't occur on Linux, only on Windows, and we can't run tests for the main window there :-/

spyder/widgets/ipythonconsole/client.py
ccordoba12: Please make this change again because it went away after I removed the test commits. Sorry.

spyder/widgets/ipythonconsole/client.py
```diff
@@ -323,33 +326,62 @@ def restart_kernel(self):
...
+        # See issue 4003
+        if sw._input_none:
             if sw.kernel_manager:
-                if self.infowidget.isVisible():
-                    self.infowidget.hide()
-                    sw.show()
-                try:
-                    sw.kernel_manager.restart_kernel()
-                except RuntimeError as e:
+                    if self.infowidget.isVisible():
```
ccordoba12: This whole block is indented 4 more line than necessary. Please fix that.

spyder/widgets/ipythonconsole/debugging.py
```diff
@@ -25,7 +25,8 @@ class DebuggingWidget(RichJupyterWidget):
     Spyder
     """
 
-    _input_reply = None
+    _input_reply = {}
+    _input_none = False
```
ccordoba12: Please change this name to `_input_reply_failed`

spyder/widgets/ipythonconsole/shell.py
```diff
@@ -40,6 +40,7 @@ class ShellWidget(NamepaceBrowserWidget, HelpWidget, DebuggingWidget):
     sig_input_reply = Signal()
     sig_pdb_step = Signal(str, int)
     sig_prompt_ready = Signal()
+    sig_debug_restart = Signal()
```
ccordoba12: Please change this to `sig_dbg_kernel_restart`

ccordoba12:
@dalthviz, I left some other comments, mainly about naming conventions. Then this should be ready to be merged.

I really appreciate all the effort you have put to fix this problem. What a nasty bug!!

spyder/widgets/ipythonconsole/client.py
ccordoba12: You also need to redo this change.

spyder/widgets/ipythonconsole/shell.py
```diff
@@ -283,7 +284,7 @@ def _banner_default(self):
 
     def _kernel_restarted_message(self, died=True):
         msg = _("Kernel died, restarting") if died else _("Kernel restarting")
-        self.sig_kernel_restarted.emit(msg)
```
ccordoba12: Please revert this change. This is not related to the things you're trying to solve in this PR.

spyder/widgets/ipythonconsole/client.py
```diff
@@ -176,6 +176,9 @@ def configure_shellwidget(self, give_focus=True):
         # To show kernel restarted/died messages
         self.shellwidget.sig_kernel_restarted.connect(
             self.kernel_restarted_message)
+
```
ccordoba12: Add a comment above this line explaining why this is needed.

ccordoba12:
@dalthviz, I finally managed to add tests for this, but in a different way than we had thought at the beginning :-)

Issue #4021, opened 2017-01-21
Spyder fails to import numpy


seawolf42:
## Description of your problem

**What steps will reproduce the problem?**

1. Install spyder, numpy, and other packages in a virtualenv
2. start spyder
3. create a python script that just imports numpy
4. execute the first line of the script (run -> run cell)

**What is the expected output? What do you see instead?**

expected: success loading the module

actual:
Spyder Internal Console

This console is used to report application
internal errors and to inspect Spyder
internals with the following commands:
  spy.app, spy.window, dir(spy)

Please don't use it to run your code

>>> Traceback (most recent call last):
  File "/Users/seawolf/dev/py_virtualenv/machinelearning-udemy/lib/python3.6/site-packages/spyder/widgets/sourcecode/codeeditor.py", line 2698, in keyPressEvent
    self.fix_indent(comment_or_string=cmt_or_str)
  File "/Users/seawolf/dev/py_virtualenv/machinelearning-udemy/lib/python3.6/site-packages/spyder/widgets/sourcecode/codeeditor.py", line 1968, in fix_indent
    if prevtext.endswith(':') and self.is_python_like():
UnboundLocalError: local variable 'prevtext' referenced before assignment

**Please provide any additional information below**

all packages in the virtualenv:
alabaster==0.7.9
appnope==0.1.0
astroid==1.4.9
Babel==2.3.4
bleach==1.5.0
chardet==2.3.0
cycler==0.10.0
decorator==4.0.11
docutils==0.13.1
entrypoints==0.2.2
html5lib==0.9999999
imagesize==0.7.1
ipykernel==4.5.2
ipython==5.1.0
ipython-genutils==0.1.0
ipywidgets==5.2.2
isort==4.2.5
jedi==0.9.0
Jinja2==2.9.4
jsonschema==2.5.1
jupyter==1.0.0
jupyter-client==4.4.0
jupyter-console==5.0.0
jupyter-core==4.2.1
lazy-object-proxy==1.2.2
MarkupSafe==0.23
matplotlib==1.5.3
mccabe==0.5.3
mistune==0.7.3
nbconvert==5.1.1
nbformat==4.2.0
notebook==4.3.1
numpy==1.11.3
numpydoc==0.6.0
pandas==0.19.2
pandocfilters==1.4.1
pep8==1.7.0
pexpect==4.2.1
pickleshare==0.7.4
prompt-toolkit==1.0.9
psutil==5.0.1
ptyprocess==0.5.1
[...]

PR #4029, opened 2017-01-23
PR: Fix fix_indent() when editor cursor is on the first line

pwoosam:
Fixes issue #4021


ccordoba12:
Thanks for working on this Patrick! Could you add test for it?

pwoosam:
Not a problem :)

It's the least I could do after I introduced this bug! 😅

PR #4040, opened 2017-01-24
PR: Show internal errors in a QMessageBox

rlaverde:
Fixes: #1525 

~~This isn't ready, I implemented but the behavior isn't the expected, the QMessageBox is shown but for every single line of the traceback (because `traceback_available` signal is emitted for every line).~~

~~any ideas of how I could fix this?~~

I implemented using a QMessageBox with non modal behavior

traceback_box

### DIFF
--- a/spyder/plugins/console.py
+++ b/spyder/plugins/console.py
@@ -19,8 +19,9 @@ import sys
 # Third party imports
 from qtpy import PYQT5
 from qtpy.compat import getopenfilename
-from qtpy.QtCore import Signal, Slot
-from qtpy.QtWidgets import QInputDialog, QLineEdit, QMenu, QVBoxLayout
+from qtpy.QtCore import Signal, Slot, Qt
+from qtpy.QtWidgets import (QInputDialog, QLineEdit, QMenu, QVBoxLayout,
+                            QMessageBox)
 
 # Local imports
 from spyder.config.base import _, debug_print
@@ -91,7 +92,11 @@ class Console(SpyderPluginWidget):
             
         # Accepting drops
         self.setAcceptDrops(True)
-        
+
+        # Traceback MessageBox
+        self.msgbox_traceback= None
+        self.error_traceback = ""
+
     #------ Private API --------------------------------------------------------
     def set_historylog(self, historylog):
         """Bind historylog instance to this console
@@ -197,13 +202,33 @@ class Console(SpyderPluginWidget):
         # Connecting the following signal once the dockwidget has been created:
         self.shell.traceback_available.connect(self.traceback_available)
     
-    def traceback_available(self):
+    def traceback_available(self, text):
         """Traceback is available in the internal console: showing the 
         internal console automatically to warn the user"""
         if CONF.get('main', 'show_internal_console_if_traceback', False):
             self.dockwidget.show()
             self.dockwidget.raise_()
-        
+        else:
+            if self.msgbox_traceback is None:
+                self.msgbox_traceback = QMessageBox(
+                    QMessageBox.Critical,
+                    _('Error'),
+                    _("<b>Spyder-IDE has encountered a problem.</b>"
+                      "Sorry for the inconvenience."
+                      "<b>Please tell us about this problem.</b>"
+                      "<br><br>"
+                      "You can submit this error to the github issue tracker"),
+                    QMessageBox.Cancel | QMessageBox.Ok,
+                    parent=self)
+
+                self.msgbox_traceback.setWindowModality(Qt.NonModal)
+                self.error_traceback = ""
+                self.msgbox_traceback.show()
+
+            self.error_traceback += text
+            self.msgbox_traceback.setDetailedText(self.error_traceback)
+
+
     #------ Public API ---------------------------------------------------------
     @Slot()
     def quit(self):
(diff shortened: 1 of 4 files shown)


goanpeca:
I dont think a QMessageBox will do then, you could create a custom dialog with a QTableView or something that would keep growing as more messages keep coming.

rlaverde:
@goanpeca Following your idea I found a way QMessageBox could be updated continuously :smile:

goanpeca:
:+1:

goanpeca:
@ccordoba12 do we want rich text on the details or not? and should they be links to the actual files?

Do we want to add a submit button so users can submit that error directly?

With that is that the issue tracker can start to grow a lot, which is good and bad at the same time!

ccordoba12:
> do we want rich text on the details or not? and should they be links to the actual files?

Perhaps not because that could confuse users. Besides, clickable links are already available in the internal console and that's mostly useful for us ;-)

What we could have is highlighted tracebacks, but we could leave that for another PR.

> Do we want to add a submit button so users can submit that error directly?

Yep, that's really important to have ;-)

rlaverde:
I added the submit button, and rebase it

spectacle m19702

spyder/widgets/sourcecode/codeeditor.py
```diff
@@ -2425,6 +2425,7 @@ def __get_current_color(self, cursor=None):
 
     def in_comment_or_string(self, cursor=None):
         """Is the cursor inside or next to a comment or string?"""
+        raise Exception
```
ccordoba12: This is making our tests to fail. It seems an artifact of the rebase, so please remove it.

ccordoba12:
@rlaverde, could use an instance of CodeEditor to show the tracebacks? I think that would make the dialog more useful :-)

spyder/widgets/sourcecode/codeeditor.py
rlaverde: Ohhh! sorry, that was for testing, I didn't meant to commit that 🙈

rlaverde:
> Could we use an instance of CodeEditor to show the tracebacks? I think that would make the dialog more useful :-)

I used a customized QMessageBox, for embedding the CodeEditor I will have to rewrite it using a Dialog and implement the layout and functionality

goanpeca:
>@rlaverde, could we use an instance of CodeEditor to show the tracebacks? I think that would make the dialog more useful :-)

@ccordoba12 I would say for now this is good enough and as @rlaverde he would have to basically redo all his PR for what you are asking

ccordoba12:
> I would say for now this is good enough and as @rlaverde he would have to basically redo all his PR for what you are asking

Ok, no problem, let's leave things as they are.

One more thing: we need to preserve the old behavior when DEV is active. That's way more practical for us developers because traceback links are clickable in the Internal console.

rlaverde:
> One more thing: we need to preserve the old behavior when DEV is active. That's way more practical for us developers because traceback links are clickable in the Internal console.

Done, also when `DEBUG` or `PYTEST` is activated

spyder/plugins/console.py
```diff
@@ -197,13 +202,44 @@ def register_plugin(self):
...
         if CONF.get('main', 'show_internal_console_if_traceback', False):
             self.dockwidget.show()
             self.dockwidget.raise_()
-        
+        else:
+            if self.msgbox_traceback is None:
+                self.msgbox_traceback = QMessageBox(
+                    QMessageBox.Critical,
+                    _('Error'),
+                    _("<b>Spyder-IDE has encountered a problem.</b>"
```
ccordoba12: Remove the `-IDE` here.

spyder/plugins/console.py
```diff
@@ -197,13 +202,44 @@ def register_plugin(self):
...
+                      "Sorry for the inconvenience."
+                      "<b>Please tell us about this problem.</b>"
+                      "<br><br>"
+                      "You can submit this error to the github issue tracker"),
```
ccordoba12: Let's change this to

> You can automatically submit this error to our Github issues tracker.<br>
*Note*: You need a Github account for that.

spyder/plugins/console.py
```diff
@@ -197,13 +202,44 @@ def register_plugin(self):
...
+                      QMessageBox.Ok,
+                    parent=self)
+
+                self.submit_btn = self.msgbox_traceback.addButton(
+                        _('Submit to github'), QMessageBox.YesRole)
```
ccordoba12: `github` -> `Github`

spyder/plugins/console.py
ccordoba12: Also add a `<br>` at the end.

spyder/plugins/console.py
```diff
@@ -197,13 +202,44 @@ def register_plugin(self):
...
             self.dockwidget.raise_()
-        
+        else:
+            if self.msgbox_traceback is None:
+                self.msgbox_traceback = QMessageBox(
+                    QMessageBox.Critical,
+                    _('Error'),
+                    _("<b>Spyder-IDE has encountered a problem.</b>"
+                      "Sorry for the inconvenience."
+                      "<b>Please tell us about this problem.</b>"
```
ccordoba12: Let's remove this line.

ccordoba12:
I added some minor formatting comments. After that, I think this is ready!

rlaverde:
I just addressed your comments, this is how it looks like:

spectacle jg8171

ccordoba12:
I like it a lot, thanks!

@goanpeca, what do you say?

PR #4044, opened 2017-01-24
PR: Add "run selection" icon to run toolbar

### LINKED ISSUE #4020: Reopen  -- Add "run selection" icon
I think this needs to be considered.  Just reoopned this.

ahed87 commented on Jan 10, 2016
Hi,
an enhancement idea triggered by issue #2490
a small (or large) suggestion related to this area would be to add one run icon.
I quite often select a piece of code and use the right-click -> "run selection or current line".
When doing that I typically have made the selection with keys, scrolled a bit with the mouse and that's when I miss an icon to press to run the selected area. I quite often end up pressing the regular run-icon when I really want to run a selection, since it's soo much easier than doing the right-click, especially if that area is out of focus in the editor.

I have no strong opinions of how such an icon would look like (like today's run with an 's' in it?), and were it should be placed (inbetween current run and runcell icons?).

mariacamilaremolinagutierrez:
Fixes #4020 

This also removes the "Run settings" icon. The run toolbar looks like this now:

image

### DIFF
--- a/spyder/plugins/editor.py
+++ b/spyder/plugins/editor.py
@@ -1051,8 +1051,8 @@ class Editor(SpyderPluginWidget):
                             configure_action, MENU_SEPARATOR]
         self.main.run_menu_actions += run_menu_actions
         run_toolbar_actions = [run_action, run_cell_action,
-                               run_cell_advance_action, re_run_action,
-                               configure_action]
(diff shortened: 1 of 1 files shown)


ccordoba12:
Nice, thanks!

PR #4045, opened 2017-01-25
PR: Make status bar widgets to have a fixed width

### LINKED ISSUE #2238: Preserve size in status bar widgets
this is only a very minor visual bug, that is most probably not important for other people, but it really distracts me..

could you make the status bar (bottom bar) have fixed field sizes? my cpu usage usually hovers around 8-12%, always jumping between one and two digit, making the whole status bar move a little bit to the left.. and catching my attention.. changing numbers are ok, shifting layout is to much..

Since it probably just me, I'll try to have a look at it if I got some more time..

mariacamilaremolinagutierrez:
Fixes #2238


ccordoba12:
Could you post screenshots of before and after this change?

mariacamilaremolinagutierrez:
Sure:

image

image

I don't know if it is easy to see, but the widths of the labels are fixed no matter the number.

ccordoba12:
Looks good to me. @goanpeca, do you agree?

goanpeca:
Yes and no, I think what we should aim for is to have widgets aligned to the right as well, so not only setting a fixed width (a bit arbitrarily at the moment), but also using strecthes in the Layout definition.

So for example between the CPU text and the 29% there should be a stretch so that the widget move to the right, so they are right aligned.

goanpeca:
So that the widget look like

```

CPU     15%
CPU      5%

and not

CPU     15%
CPU     5%

PR #4104, opened 2017-02-04
PR: Pin Jedi to 0.9.0

ccordoba12:
I ran our tests with the new 0.10.0 version and several of them are failing.

Besides, our new completions for Numpy and Matplotlib are failing with that version too.


ghisvail:
Debian Stretch will ship with jedi `v0.10.0`, how should we proceed?

ccordoba12:
We are aiming at being compatible with Jedi 0.10 in version 3.1.4, to be released in a month or so.

If the addition of Jedi 0.10 is imminent in Debian Stretch, then users will have to depend on Rope until 3.1.4 is out.

ghisvail:
> If the addition of Jedi 0.10 is imminent in Debian Stretch

It is there already.

> users will have to depend on Rope until 3.1.4 is out

Rope is Python 2 only on Debian (version 0.10.3). Upstream got *reliable* support for Python 3 only recently (version 0.10.4), unfortunately not in time for the freeze. How important are `rope` / `jedi` for spyder ?

ccordoba12:
> How important are rope / jedi for spyder ?

They provide code completion for the Editor, so I'd say pretty important :-)

ghisvail:
I meant `spyder` actually works fine without any of rope / jedi installed right? Sure, it would be a missing feature (whose importance is relative), but the rest of the editor will remain perfectly usable am I correct?

Otherwise, how far are you from having compatibility with 0.10.0? Is that something which could be cherry-picked on top of 3.1.3?

ccordoba12:
> Otherwise, how far are you from having compatibility with 0.10.0?

I found some problems with 0.10.0 that we either have report upstream as errors or fix in our side. So I don't know when we'll be compatible with 0.10.0.

And there's also PR #4121.

> Is that something which could be cherry-picked on top of 3.1.3?

I'd think you can cherry-pick our changes, yep.

ghisvail:
So, the corresponding patch to PR #4121 would make spyder 3.1.3 work with jedi 0.10.0 or is there more to be done?

ccordoba12:
> is there more to be done?

Yes, there is. Our test suite is not passing with 0.10.0 even after applying #4121.

Issue #4085, opened 2017-02-01
Crash in "New window" when opening new file.


pnavarre:
## Description of your problem

**What steps will reproduce the problem?**

1. Open Spyder. On the tab of the current file open in Editor, I right-click and select "New window". A new window with Editor and the current file opens correctly.
2. In the new window I click "Open file" icon. In the "Open file" window I select a second file to open, click on it (turns blue) and then click "Open" button. 
3. After pressing "Open" Spyder crashes with message:
27558 Segmentation fault      (core dumped) spyder3

**What is the expected output? What do you see instead?**
I expect Spyder to open the new file in the Editor (in both windows open). Instead, Spyder crashes.

**Please provide any additional information below**

When writing this issue it took me long time to execute each step and then Spyder didn't crash. Then I repeated following the steps without waiting and crashes again. Also, the first time I noticed the issue Spyder hanged for long time before crashing.

I use a virtual environment with Python 3.5.3 on Ubuntu 16.04. I install PyQt5 v5.7.1 from pip. This comes with a local copy of Qt (v5.7.1) so it doesn't mess with the Ubuntu Qt. I compile Spyder 3.1.2 from source in the virtual environment with all necessary dependencies (see below).
Probably unrelated, but I also use nvidia driver which has an issue reported in #3226. I fix it manually according to the post, using PyOpenGL.

## Versions and main components

* Spyder Version: 3.1.2
* Python Version: 3.5.3
* Operating system: Ubuntu LTS 16.04

## Dependencies

Please go to the menu entry `Help > Optional Dependencies` (or
`Help > Dependencies`), press the button `Copy to clipboard`
and paste the contents below:

jedi >=0.8.1     :  0.9.0 (OK)
matplotlib >=1.0 :  2.0.0 (OK)
nbconvert >=4.0  :  5.1.1 (OK)
numpy >=1.7      :  1.12.0 (OK)
pandas >=0.13.1  :  0.19.2 (OK)
pep8 >=0.6       :  1.7.0 (OK)
psutil >=0.3     :  5.0.1 (OK)
pyflakes >=0.6.0 :  1.5.0 (OK)
pygments >=2.0   :  2.2.0 (OK)
pylint >=0.25    :  1.6.5 (OK)
qtconsole >=4.2.0:  4.2.1 (OK)
rope >=0.9.4     :  0.10.3 (OK)
sphinx >=0.6.6   :  1.5.2 (OK)
sympy >=0.7.3    :  1.0 (OK)

PR #4110, opened 2017-02-06
PR: Add more tests for the main window

ccordoba12:
This adds a test for issue #4085 and also for running cells and files.

### DIFF
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -146,6 +146,7 @@ def test_run_code(main_window, qtbot):
     main_window.close()
 
 
+@pytest.mark.skipif(os.name == 'nt', reason="It's timing out sometimes on Windows")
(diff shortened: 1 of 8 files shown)


spyder/app/tests/test_mainwindow.py
```diff
@@ -23,11 +62,125 @@ def main_window():
...
+
+    # Create a new editor window
+    # Note: editor.load() uses the current editorstack by default
+    main_window.editor.create_new_window()
+    main_window.editor.load()
+
+    # Perform the test
+    # Note: There's always one file open in the Editor
+    editorstack = main_window.editor.get_current_editorstack()
+    assert editorstack.get_stack_count() == 2
```
rlaverde: This test fails when there are previous files open in Spyder (in a previous session)

ccordoba12: It's meant to be run in CIs. We could skip it if we're not running it not in a CI.

PR #4112, opened 2017-02-06
Add AppStream metainfo file

ximion:
This patch adds an AppStream metainfo file to Spyder, providing metadata to make Spyder visible in the software centers of Linux distributions, like KDE Discover and GNOME Software (the default software center in Ubuntu and Fedora).

At time, we process the .desktop file in Debian/Ubuntu to make Spyder show up, but that's not ideal - with this patch, you gain control over the information that is shown and a screenshot gets displayed as well.

The metainfo XML file can be translated, so you can hook it up into your l10n system if you have one and want that.

More information on AppStream can be found [here](https://www.freedesktop.org/software/appstream/docs/), and there is a [quickstart guide](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html#sect-Quickstart-DesktopApps) for application maintainers too.

This patch makes the Python3 version of Spyder the only version to be shown in software centers - the other versions are not shown. I think that is a good and future-proof decision, but if you think otherwise, let me know and I will adjust the patch.

You can validate the metainfo file using the command `appstreamcli validate spyder3.appdata.xml`.

Additionally to adding this metadata, you might also consider the following:
  * Rename the .desktop files to follow the reverse-DNS scheme as recommended by the [specification](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s02.html). - This will kill warnings when AppStream validates the metainfo file.
  * Ship more icon sizes (or an SVG icon) in the canonical locations in `/usr/share/icons/hicolor/<size>/apps` to make icon handling easier (and save desktops from resizing stuff a lot).

Cheers,
    Matthias

### DIFF
--- a/scripts/spyder.desktop
+++ b/scripts/spyder.desktop
@@ -11,4 +11,4 @@ Icon=spyder
 Terminal=false
 StartupNotify=true
 MimeType=text/x-python;
-
+X-AppStream-Ignore=True
--- a/scripts/spyder3.appdata.xml
+++ b/scripts/spyder3.appdata.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="utf-8"?>
+<component type="desktop-application">
+  <id>spyder3.desktop</id>
+  <name>Spyder</name>
+  <summary>Scientific Python Development Environment</summary>
+
+  <metadata_license>MIT</metadata_license>
+  <project_license>MIT</project_license>
+  <developer_name>Spyder Project Contributors</developer_name>
+
+  <description>
+    <p>
+      Spyder is a free open-source Python development environment providing MATLAB-like features in a simple and light-weighted software.
+    </p>
+    <p>
+      It contains a powerful interactive development environment for the Python language with advanced editing, interactive testing, debugging and introspection features
+      and a numerical computing environment thanks to the support of IPython (enhanced interactive Python interpreter) and popular Python libraries such as NumPy
+      (linear algebra), SciPy (signal and image processing) or matplotlib (interactive 2D/3D plotting).
+    </p>
+    <p>
+      Spyder may also be used as a library providing powerful console-related widgets for your PyQt-based applications – for example, it may be used to integrate
+      a debugging console directly in the layout of your graphical user interface.
+    </p>
+  </description>
+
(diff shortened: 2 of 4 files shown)


ccordoba12:
Thanks a lot for your contribution! I see no problem with this.

Would you mind to redo this PR against the `3.1.x` branch, so that we can add it to one of our bugfix releases? The master branch is for Spyder 4, which will come in a year or so.

> Rename the .desktop files to follow the reverse-DNS scheme

I don't understand this. What should be the name of our desktop file?

> Ship more icon sizes

We are in the process of remaking our icon, but our current icon is an SVG one ;-)

ximion:
>> Rename the .desktop files to follow the reverse-DNS scheme
> I don't understand this. What should be the name of our desktop file?

Something like `org.pythonhosted.spyder3.desktop` or `com.github.spyder3.desktop` - those IDs are used because they are unique on the system, which allows apps to be D-Bus activated on Wayland and also makes it easy to map icons to applications on Wayland.

>> Ship more icon sizes
> We are in the process of remaking our icon, but our current icon is an SVG one ;-)

This of course makes this issue way less relevant - since using `/usr/share/pixmaps` is discouraged, one could place the svg icon in `/usr/share/icons/hicolor/scalable/apps` directory though.

> Would you mind to redo this PR against the 3.1.x branch

Should I do a separate PR for that?

ccordoba12:
> Something like org.pythonhosted.spyder3.desktop or com.github.spyder3.desktop

Ok, we'll think about it.

> This of course makes this issue way less relevant - since using /usr/share/pixmaps is discouraged, one could place the svg icon in /usr/share/icons/hicolor/scalable/apps directory though

I think leaving it in `pixmaps` is fine for now (for old distros).

> Should I do a separate PR for that?

Yep, you need to do it against the 3.1.x branch.

ximion:
PR for the 3.1.x branch is filed as https://github.com/spyder-ide/spyder/pull/4123

ximion:
grafik
Btw, this is how the data looks like in KDE Discover.

ccordoba12:
Really nice!! Thanks!

I'm closing this one because it's not necessary anymore ;-)

PR #4121, opened 2017-02-08
PR: Monkey patch Jedi 0.10.0 for numpydoc

bcolsen:
This works now...but not ready for review yet.  I think I need some PEP8
It's a bit more of a mess than I would like but I have a plan to get this upstream faster.


ccordoba12:
Thanks a lot @bcolsen fir giving us a hand with this!

bcolsen:
@ccordoba12 If it passes testing, It's ready for review.

ccordoba12:
You need to remove the pins I introduced on PR #4104 for setup.py and meta.yaml

ghisvail:
Subscribing myself to this bug, since this would enable code completion to work on Debian with the packaged `jedi` version 0.10.0.

ghisvail:
Just as a test run, I cherry-picked a squashed version of this PR on top of the current `spyder-3.1.3` package on Debian, and code completion is now working with the packaged version of Jedi.

Many thanks @bcolsen.

ccordoba12:
It's true!! @bcolsen's work makes all our test suite to pass with Jedi 0.10.0. Thanks a lot @bcolsen!!

Please merge with `3.1.x` to get some fixes to our tests, and then I'll merge ;-)

spyder/utils/introspection/jedi_patch.py
```diff
@@ -22,10 +22,11 @@ def apply():
 
     See [1] and [2] module docstring."""
     from spyder.utils.programs import is_module_installed
-    if is_module_installed('jedi', '=0.9.0'):
+    if (is_module_installed('jedi', '=0.9.0') or
+            is_module_installed('jedi', '=0.10.0')):
```
ccordoba12: Please change this to

```python
if (is_module_installed('jedi', '>=0.9.0'):
```

spyder/utils/introspection/jedi_patch.py
```diff
@@ -22,10 +22,11 @@ def apply():
...
         import jedi
     else:
-        raise ImportError("jedi %s can't be patched" % jedi.__version__)
+        raise ImportError("jedi not =0.9.0 or 0.10.0, can't be patched")
```
ccordoba12: raise ImportError("jedi not >=0.9.0, can't be patched")

spyder/utils/introspection/numpy_docstr.py
```diff
@@ -106,8 +108,8 @@ def _search_return_in_numpydocstr(docstr):
     return found
 
 
-@memoize_default(None, evaluator_is_first_arg=True)
-def find_return_types(evaluator, func):
+#@memoize_default()
```
ccordoba12: So this should be removed or not?

ccordoba12:
I also left another couple of (very simple) comments ;-)

spyder/utils/introspection/numpy_docstr.py
bcolsen: I left it there to show that I disabled caching for that function to make monkey patch work.

I had to change the way I'm monkey patching and I couldn't get the caching to work.  Luckily, I didn't notice a difference in the run speed.

I am able to cache with the upstream patch.

I can give you more details on the problem later today.

ccordoba12: Please add a comment above this line mentioning this reason.

spyder/utils/introspection/jedi_patch.py
bcolsen: I remember that I wrote it that way so that when Jedi 0.11.0 comes out my patching won't thrash it(like it did when 0.10.0 came out).  I am also hoping not to have to patch 0.11.0  :-)

I can still make the change if you think that's best.

ccordoba12: Ok, fair enough ;-) Please leave it as it is then.

bcolsen:
I fixed the caching....I just changed it back to the way it was in 0.9.0 and that still works in 0.10.0

If it passes the tests it's ready to go.

ccordoba12:
@bcolsen, I added Jedi 0.10.0 to CircleCI and, as you can see, our tests are failing there (the other CIs are still using Jedi 0.9.0 because they can't be modified so easily, so please only look at the results in CircleCI).

Besides, I'd like to propose you to drop support for 0.9.0 and only focus on 0.10.0. The patches for 0.9 and 0.10 are very different, and since we won't be able to test both versions, I'd prefer to drop support for the oldest one. That'd also simplify your work ;-)

bcolsen:
Wouldn't dropping jedi 0.10 simplify my work even more, 0.9 already works ;-)

Joking aside, I everything seems to work now. If you want, I can take out the 0.9 code but I think some users might not like that in a point release.  Maybe wait for 3.2?

The current solution:
If at first you don't succeed...Change the test.  ;-)

The problem:
jedi_plugin.py (line 61)
```Python
    def get_info(self, info):
        call_def = self.get_jedi_object('goto_definitions', info)
        for cd in call_def:
            if cd.doc and not cd.doc.rstrip().endswith(')'):
                call_def = cd
                break
        else:
            call_def = call_def[0]
```
This code gives an `IndexError` when `call_def` is an empty list.

Jedi 0.10 without any patching fails the two spyder tests with `get_info()`. As you can see in the code diff I fixed this by removing the extra characters after the function names. I'm not sure why the `get_info()` tests had extra characters after the function names, but It's strange that it works in 0.9 either.

Jedi 0.10.0
```python
In [1]: import jedi
In [2]: jedi.__version__
Out[2]: '0.10.0'
In [3]: jedi.Script('import os: os.walk(').goto_definitions()
Out[3]: []
In [4]: jedi.Script('import os: os.walk').goto_definitions()
Out[4]: [<Definition def walk>]
```
Good, not an empty list. I changed the test and now 0.10.0 passes

Jedi 0.9.0
```python
In [1]: import jedi
In [2]: jedi.__version__
Out[2]: '0.9.0'
In [3]: jedi.Script('import os: os.walk').goto_definitions()
Out[3]: []
In [4]: jedi.Script('import os: os.walk(').goto_definitions()
Out[4]: []
```
How does that work? I must be missing something here, 0.9.0 also passes.

For interest, considering that the pluging seem to be written before 0.9

Jedi 0.8.1
```python
In [1]: import jedi
In [2]: jedi.__version__
Out[2]: '0.8.1-final0'
In [3]: jedi.Script('import os: os.walk(').goto_definitions()
Out[3]: []
In [4]: jedi.Script('import os: os.walk').goto_definitions()
Out[4]: [<Definition module os>]
```
Different again?

ghisvail:
@ccordoba12 Are you satisfied with the latest changes from @bcolsen ? If so, could you ack it (even if pushed later), so I can start cherry-picking the final fix to the Debian package. Thanks.

ccordoba12:
I'll give it a final review on Wednesday.

ghisvail:
Any news?

ghisvail:
At least, can I have confirmation that all commits up to f9fe24d can be squashed and cherry-picked to v3.1.2 or v3.1.3 to get code completion working with Jedi 0.10? Thanks.

It is getting very unlikely `v3.1.4` will be granted a freeze exception for Debian Stretch (due to the large change set since `v3.1.2`), but I can at least backport fixes for regressions in usability, such as [broken code completion](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=855873).

ccordoba12:
@ghisvail, tests are passing, so I'd say yes. However, I'd like to give this one a little bit of manual testing before merging. I'll promise you I'll do that this weekend ;-)

ghisvail:
As far as testing is concerned, I have used a package I built locally with `3.1.3` + this patch cherry-picked on top of it and enjoyed using `spyder` with working code completion for the week.

ccordoba12:
Ok, merging this one then :-)

Thanks a lot @bcolsen!

ghisvail:
@ccordoba12 Unfortunately, this PR does not apply cleanly on top of `3.1.2`. It does on `3.1.3`, so I will be rolling out an update of the packaging in Debian `experimental` with this fix, and try to make my case for an unblock request for Debian Stretch.

ccordoba12:
3.1.3 fixes a lot of segfaults introduced in 3.1.0 because of some odd focus bugs, so I'd really prefer if you use it instead of 3.1.2.

ghisvail:
It would help me build my case if you could provide a separate listing containing only these fixed segfaults on Linux. The more comprehensive my request is, the more likely it will be accepted. Thanks.

ccordoba12:
Please take a look at our Changelog. Those segfaults affected all our supported OSes.

ghisvail:
> Please take a look at our Changelog

Since you personally reviewed most, if not all, the PR listed in this change log, you are best placed to identify the most critical issues. Even if it is just the top 3 or 5.

ccordoba12:
So you want to use only certain patches?

ccordoba12:
I mean, it's a bugfix release, so why not package the full release instead?

ghisvail:
> So you want to use only certain patches?

No.

> I mean, it's a bugfix release, so why not package the full release instead?

That's the plan.

I'll provide some more context. Debian is in freeze, meaning new versions of software no longer migrate automatically to what is going to become the next stable release. Instead, updates of packages should be focusing on addressing release critical bugs affecting the frozen packages (such as failures to build, security issues, crashes...).

Updates to packages *can* migrate if an unblock request is granted from the release team, who will manually review the proposed changes and assess the necessity of such update (here to `3.1.3`) from the current version in Stretch (here `3.1.2`). Just providing a link to a big change log and writing "it is a bug fix release, do it." is not enough and is likely to upset the reviewers more than anything.

Hence myself asking whether you could make a selection of the most critical bugs fixed by `3.1.3`, so I can provide a better case to the release team by pinpointing a few reproducible crashes / segfaults.

ccordoba12:
Ok, thanks for the thorough explanation. I'll provide you with the highlights of 3.1.3 this weekend.

ghisvail:
Did you have a chance to look at it?

We now have version 3.1.3 + this fix in Debian experimental and would like to draft the unblock request quite soon.

Thanks.

ccordoba12:
The general picture: I consider **3.1.3** the first version in the 3.1 series that is stable enough to be used by the general public. All previous versions were semi-broken because of important bugs and I don't recommend to use them.

Critical bugs fixed in 3.1.3 (I don't link them with `#` because it makes no sense to do it in this issue):

* 4127 - Crash at startup, reported by Ubuntu users.
* 4122 - Segfault after closing the Run settings dialog
* 4088 - Segfault when trying to save figures in the IPython console.
* 4085 - Segfault when opening a file in a new Editor window.
* 4066 - Segfault when removing a variable in the console while it was being viewed in the Variable Explorer.

All these issues were caused by a couple of focus bugs introduced by our junior developers. It was not their fault really, we simply didn't know that changing focus policies for the main window and the application will have those terrible effects.

To avoid this in the future, we have expanded our test suite as much as we can in 3.1.3 and 3.1.4.

ghisvail:
Thanks for providing the list of critical bugs, here is what I found:

> 4127 - Crash at startup, reported by Ubuntu users.

Not reproducible on `Debian`

> 4122 - Segfault after closing the Run settings dialog

Not reproducible on `Debian` (makes sense, we use the Qt5 backend and it only happens on Qt4)

> 4088 - Segfault when trying to save figures in the IPython console

Reproducible on `Debian`

> 4085 - Segfault when opening a file in a new Editor window.

Not reproducible on `Debian` (not sure why)

> 4066 - Segfault when removing a variable in the console while it was being viewed in the Variable Explorer.

Reproducible on Debian, but not your usual use case of the var explorer tbh.

To sum up, that gives me 2 bugs out of 5 which are reproducible with the current package, including one for an atypical use case. That does not give me much of a case but I'll try.

ccordoba12:
I hope my opinion counts too :-) I mean, there are other annoying bugs (like wrong auto-indentation, no new breakpoints in the IPython console while debugging and broken maximizing-the-current-pane button) and I don't want to keep telling people: please update because those things were fixed already.

I have to say that that really bothers me as a maintainer: we're doing our duty of releasing bugfix versions, but they are not being picked up by Debian. At least those versions should be easily backported to the stable Debian branch.

Sigh, unfortunately I know that's a lost battle with Debian....

ghisvail:
> I hope my opinion counts too :-)

I voiced your opinion, and referenced all your supporting material, in the following request for [unblock](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=858841). The rest is up to the release team at this point.

> I have to say that that really bothers me as a maintainer: we're doing our duty of releasing bugfix versions, but they are not being picked up by Debian.

They are. The software is currently up-to-date in the `experimental` channel.

> At least those versions should be easily backported to the stable Debian branch.

We will provide backports of bugfix releases in a `stretch-backports` channel.

ccordoba12:
Ok, thanks for the good news!

ghisvail:
The unblock request has been accepted. Debian Stretch will have 3.1.3 + this fix + #4311 for now. If 3.1.4 gets released soon enough, I might try to request another unblock again.

PR #4125, opened 2017-02-09
PR: Make loading IPython kernels by using their ids to work again

### LINKED ISSUE #4053: Spyder 3.1.2 cannot connect to kernel as 3.0.2
I can choose the file directory to connect to a kernel like "kernel-100.json",
but when I just type 100, It shows that cannot connect to kernel, however this can be done in spyder 3.0.2

andfoy:
Fixes #4053

### DIFF
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -206,12 +206,12 @@ class KernelConnectionDialog(QDialog):
         ssh_form.addRow(_('Password'), self.pw)
         
         # Ok and Cancel buttons
-        accept_btns = QDialogButtonBox(
+        self.accept_btns = QDialogButtonBox(
             QDialogButtonBox.Ok | QDialogButtonBox.Cancel,
             Qt.Horizontal, self)
 
-        accept_btns.accepted.connect(self.accept)
-        accept_btns.rejected.connect(self.reject)
+        self.accept_btns.accepted.connect(self.accept)
+        self.accept_btns.rejected.connect(self.reject)
 
         # Dialog layout
         layout = QVBoxLayout(self)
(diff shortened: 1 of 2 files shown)


andfoy:
@ccordoba12 I don't exactly how to manage the QDialog interactions while testing this functionality. After creating the fixture and trying to press Enter button using qtbot, the action doesn't take place, and the dialog must be closed manually

goanpeca:
@andfoy you need to use a Timer to trigger an event on the dialog after you have interacted with it. Also you need to have a way of keeping a reference to the dialog so you can interact with it.

Look at https://github.com/pytest-dev/pytest-qt/issues/18

jitseniesen:
The approach I took in [spyder_unittest/widgets/tests/test_configdialog.py](https://github.com/spyder-ide/spyder-unittest/blob/master/spyder_unittest/widgets/tests/test_configdialog.py), which tests [spyder_unittest/widgets/configdialog.py](https://github.com/spyder-ide/spyder-unittest/blob/master/spyder_unittest/widgets/configdialog.py), is never to test a model dialog box.

You can test the dialog box by itself by not calling `.exec_()` on it. Then there is a small bit of code which is untested (the static method `get_connection_parameters()` in your case), which you should make as small as possible; see `ask_for_config()` in my `configdialog.py` for inspiration. Finally, you can test the calling code by mocking `get_connection_parameters()`.

I guess you can use a Timer as goanpeca suggested to test the remaining glue code, but I never tried that.

ccordoba12:
@andfoy, as @goanpeca said, one way to test a modal dialog is to use a timer to interact with it after it shows up. I implemented that for some main window tests here:

https://github.com/spyder-ide/spyder/blob/master/spyder/app/tests/test_mainwindow.py#L171

which fires up this function

https://github.com/spyder-ide/spyder/blob/master/spyder/app/tests/test_mainwindow.py#L43

In your case, you don't need to look for a widget that is a QFileDialog but a QDialog instead.

----

To select the `Ok` button of the KernelConnectionDialog and press it, please take a look at this

https://github.com/spyder-ide/spyder/blob/master/spyder/app/tests/test_mainwindow.py#L241

PR #4151, opened 2017-02-16
PR: Make tests more reliable

ccordoba12:


goanpeca:
@ccordoba12 it would be much better to create signals that send messages once something has happened. Adding timeouts is prone to fail randomly or make tests longer that they should.

ccordoba12:
I think the problem is the main window sometimes doesn't receive focus and that makes our tests to timeout. So signals wouldn't be of help here ;-)

I removed the previous commit and instead I'm hiding now the splash screen on tests to see if that helps.

ccordoba12:
This is my best effort to fix our issues in CircleCI and Appveyor. I'm still not sure if it works for sure because those failures are not deterministic :-)

Issue #4153, opened 2017-02-16
vertical end-of-line in wrong place


drammock:
## Description

The setting "show vertical line after XX characters" draws the line in the wrong place, and changing its value does not change line position after clicking "apply" or "OK".

**What steps will reproduce the problem?**

1. open spyder
2. enable vertical line after xx characters
3. set xx to any number other than 77

**What is the expected output? What do you see instead?**
Line should appear after the specified number of characters.
Instead, line appears after 77 characters no matter what setting is used (see screenshot)

![screenshot](https://cloud.githubusercontent.com/assets/1810515/23036128/75fd3e18-f436-11e6-9415-dbb5ec2d5260.png)

**Please provide any additional information below**

My font is set as [Source Code Pro](https://github.com/adobe-fonts/source-code-pro) at size 9.  Changing typeface or pointsize does not appear to fix this, though I only tried a couple fonts and a couple pointsizes.

## Version and main components

* Spyder Version: 4.0.0.dev0 7e9a9bf
* Python Version: 2.7.12
* Qt Versions: 4.8.7, PyQt4 (API v2) 4.11.4 on Linux

## Dependencies
```
pyflakes >=0.5.0 :  1.1.0 (OK)
pep8 >=0.6       :  1.7.0 (OK)
pygments >=2.0   :  2.1.3 (OK)
qtconsole >=4.2.0:  4.2.1 (OK)
nbconvert >=4.0  :  4.2.0 (OK)
pandas >=0.13.1  :  0.18.0+104.g1c8816f (OK)
numpy >=1.7      :  1.12.0.dev0+2af06c8 (OK)
sphinx >=0.6.6   :  1.5a0 (OK)
rope >=0.9.4     :  0.10.3 (OK)
jedi =0.9.0      :  0.9.0 (OK)
psutil >=0.3     :  3.4.2 (OK)
matplotlib >=1.0 :  1.5.1+1643.g0423430 (OK)
sympy >=0.7.3    :  None (NOK)
pylint >=0.25    :  1.5.5 (OK)

```

rlaverde:
I'm able to reproduce this error, It happens when using python2.7, I will start tracing it

@ccordoba12 Are there any plans for stop supporting python2.7? (maybe after spyder4)

ccordoba12:
In Spyder 5, I think.

PR #4155, opened 2017-02-16
PR: Fix error edge line in python2.7

rlaverde:
Fixes: #4153

### DIFF
--- a/spyder/widgets/panels/manager.py
+++ b/spyder/widgets/panels/manager.py
@@ -14,7 +14,7 @@ Adapted from https://github.com/pyQode/pyqode.core/blob/master/pyqode/core/manag
 from spyder.api.manager import Manager
 from spyder.api.panel import Panel
 from spyder.config.base import debug_print
-
+from spyder.py3compat import is_text_string
 
 class PanelsManager(Manager):
     """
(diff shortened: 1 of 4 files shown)


goanpeca:
@rlaverde now that you did this.

Could we add a test to find with a regex or something

`isinstance(.*,.*str)`  ?

rlaverde:
> Could we add a test to find with a regex or something

I added it in the last commit, what do you think?

spyder/widgets/panels/manager.py
```diff
@@ -94,7 +94,7 @@ def get(self, name_or_klass):
         :param name_or_klass: Name or class of the panel to retrieve.
         :return: The specified panel instance.
         """
-        if not isinstance(name_or_klass, str):
+        if not is_text_string(name_or_klass, str):
```
ccordoba12: @rlaverde, this call is wrong because `is_text_string` only receives one argument.

spyder/workers/updates.py
```diff
@@ -76,7 +76,7 @@ def start(self):
                 data = page.read()
 
                 # Needed step for python3 compatibility
-                if not isinstance(data, str):
+                if not is_text_string(data, str):
```
ccordoba12: This is also wrong, and it's what's making our tests to time out in master because checking if there's a new Spyder version is failing now.

PR #4165, opened 2017-02-18
PR: Loading spydata files now uncompress them inside a temp folder

### LINKED ISSUE #3917: Failed to import data in Variable Explorer on Windows

Import data is not working.

1. In variable explorer, export data as `a.spydata`
2. Import `a.spydata`

In source code `spyder/widgets/ipythonconsole/namespacebrowser.py`
```python
    def load_data(self, filename, ext):
        # Wait until the kernel tries to load the file
        wait_loop = QEventLoop()
        self.sig_got_reply.connect(wait_loop.quit)
        self.silent_exec_method(
HERE--->        "get_ipython().kernel.load_data('%s', '%s')" % (filename, ext))
        wait_loop.exec_()
```
I'm using Windows, so `filename` contains backslash, which leads to this error. Escaping should be done here.

It's so hard to find out the solution to this problem... And I think all `silent_execute` is not safe because the same problem could happen.

But weirdly, spyder with python3 works fine.

* Spyder Version: 3.0.2
* Python Version: Python 2.7.10 (default, May 23 2015, 09:44:00) [MSC v.1500 64 bit (AMD64)]

## Dependencies

and paste the contents below:

andfoy:
Fixes #3917

This error was caused due to write permission restrictions inside a folder. Now the files are uncompressed in a temporal folder, which allows to bypass any write restriction present in the directory where the spydata file is located.


andfoy:
@ccordoba12 This PR is ready to go!

spyder/utils/iofuncs.py
```diff
@@ -338,13 +339,16 @@ def load_dictionary(filename):
...
+    os.chdir(tmp_folder)
     data = None
     error_message = None
     try:
         tar = tarfile.open(filename, "r")
         tar.extractall()
-        pickle_filename = osp.splitext(filename)[0]+'.pickle'
+        data_file = osp.basename(filename)
+        pickle_filename = osp.splitext(data_file)[0]+'.pickle'
+        print(pickle_filename)
```
goanpeca: delete print?

spyder/utils/iofuncs.py
```diff
@@ -368,11 +372,10 @@ def load_dictionary(filename):
                         data[name].insert(index, arr)
             except KeyError:
                 pass
-        for fname in [pickle_filename]+[fn for fn in list(saved_arrays.values())]:
-            os.remove(fname)
     except (EOFError, ValueError) as error:
         error_message = to_text_string(error)
     os.chdir(old_cwd)
+    shutil.rmtree(tmp_folder)
```
goanpeca: Any chance this operation can fail? try/except?

ccordoba12:
Very good job @andfoy!! Specially for noticing that we needed to uncompress spydata files in a temp dir in order to load them :-)

Just a recommendation for next time: please try to avoid so many merges, or in case you do them, please squash them at the end ;-)

Issue #4174, opened 2017-02-19
Add our required packages for testing to setup.py


ccordoba12:
This is needed to allow external packagers (e.g. Debian) to run our test suite in a simple and reproducible way.

For an example of how to do it, please see:

https://github.com/jupyter/notebook/blob/master/setup.py#L156

rlaverde:
I'm not sure if It is the best way to add test dependencies, some people discourage the abuse of `extras_requirements` an prefer [tox](https://tox.readthedocs.io/en/latest/) instead.

What do you think? maybe It'll be better to add tox.ini? although tox [doesn't support ](https://github.com/tox-dev/tox/issues/273) conda enviroments :disappointed:

ccordoba12:
Let's keep it simple for now. We can later consider tox, for now all I want is to have a place (beside our CI config files) to register our test deps.

bilderbuchi:
Wouldn't you use setuptools' `tests_require` for packaging, well, *test* dependencies, instead of `extras_require`?

ccordoba12:
`setuptools` is the devil ;-) I really don't like it, sorry. I still prefer to use `distutils`.

bilderbuchi:
ah, I see.

PR #4185, opened 2017-02-22
PR: Add testing requirements to setup.py

rlaverde:
Fixes: #4174

-------
I take the packages that I found in the continuum integration recipes.

I'm not sure if `numpy` should also be added (pandas will install it as a dependency), and if `pillow` if needed (circle-ci recipe doesn't install it)

### DIFF
--- a/continuous_integration/travis/install.sh
+++ b/continuous_integration/travis/install.sh
@@ -61,14 +61,10 @@ install_conda()
 install_pip()
 {
     # Install PyQt5
-    conda install 'pyqt>=5.6.0';
+    pip install pyqt5;
 
     # Install testing packages
-    pip install pytest pytest-cov pytest-qt mock
-
-    # Install extra packages
-    EXTRA_PACKAGES="matplotlib pandas sympy pyzmq pillow"
-    pip install $EXTRA_PACKAGES
+    pip install -e .[test]
 }
 
 
--- a/continuous_integration/travis/install.sh
+++ b/continuous_integration/travis/install.sh
@@ -61,7 +61,7 @@ install_conda()
 install_pip()
 {
     # Install PyQt5
-    pip install pyqt5;
+    conda install 'pyqt>=5.6.0';
 }
 
 
--- a/continuous_integration/travis/install.sh
+++ b/continuous_integration/travis/install.sh
@@ -62,9 +62,6 @@ install_pip()
 {
     # Install PyQt5
     pip install pyqt5;
-
-    # Install testing packages
-    pip install -e .[test]
 }
 
 
--- a/continuous_integration/travis/run_test.sh
+++ b/continuous_integration/travis/run_test.sh
@@ -26,6 +26,9 @@ if [ "$USE_CONDA" = true ] ; then
 else
     cd $FULL_SPYDER_CLONE
     pip install dist/spyder-*.whl
+
+    # Install testing packages
+    pip install -e .[test]
 fi
 
 
--- a/continuous_integration/travis/run_test.sh
+++ b/continuous_integration/travis/run_test.sh
@@ -25,10 +25,8 @@ if [ "$USE_CONDA" = true ] ; then
     conda install -q $EXTRA_PACKAGES
 else
     cd $FULL_SPYDER_CLONE
-    pip install dist/spyder-*.whl
-
-    # Install testing packages
-    pip install -e .[test]
+    export WHEEL=`ls dist/spyder-*.whl`
+    pip install $WHEEL[test]
 fi
 
 
--- a/continuous_integration/travis/run_test.sh
+++ b/continuous_integration/travis/run_test.sh
@@ -6,14 +6,14 @@ set -ex
 export TEST_CI_APP=True
 
 
-# Extra packages to install besides Spyder regular dependencies
-# We install them here and not in travis_install.sh to see if
-# Spyder is correctly pulling its deps (some of them are shared
-# with mpl)
-export EXTRA_PACKAGES="nomkl pandas sympy pillow scipy"
-
 # Install our builds of Spyder
 if [ "$USE_CONDA" = true ] ; then
+    # Extra packages to install besides Spyder regular dependencies
+    # We install them here and not in travis_install.sh to see if
+    # Spyder is correctly pulling its deps (some of them are shared
+    # with mpl)
+    export EXTRA_PACKAGES="nomkl pandas sympy pillow scipy"
+
     # Move to a tmp dir
     mkdir ~/tmp
     cd ~/tmp
--- a/setup.py
+++ b/setup.py
@@ -290,8 +290,16 @@ install_requires = [
 
 extras_require = {
     'test:python_version == "2.7"': ['mock'],
-    'test': ['pytest', 'pytest-qt', 'pytest-cov', 'mock', 'coveralls',
-             'flaky', 'pandas', 'scipy', 'sympy', 'pillow'],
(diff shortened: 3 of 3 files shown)


setup.py
```diff
@@ -288,8 +288,15 @@ def run(self):
     'numpydoc',
 ]
 
+extra_require = {
```
ccordoba12: `extra_require` -> `extras_require`

setup.py
```diff
@@ -288,8 +288,15 @@ def run(self):
...
+    'test:python_version == "2.7"': ['mock'],
```
ccordoba12: How does this syntax work? I'm just curious and I'd like to understand how it works :-)

setup.py
ccordoba12: The failure in Travis is caused because of this.

ccordoba12:
> I'm not sure if numpy should also be added (pandas will install it as a dependency), and if pillow if needed (circle-ci recipe doesn't install it)

Ok, no problem about Numpy., i.e. you can remove it.

Pillow is used to generate image representations of arrays in the Variable Explorer. So please add a test for that and leave it as a test dependency ;-)

setup.py
rlaverde: I only found that documented in the[ wheels docs](https://wheel.readthedocs.io/en/latest/#defining-conditional-dependencies)

The idea is that you could run:
`pip install spyder[test]`
and It will install this extra dependencies

ccordoba12:
Ok, good. Now please modify our Travis scripts to install our wheel with `[test]`, then install `pyqt5` (i.e. remove our conda install of pyqt5 for pip) and finally run our pytest's there.

The idea is to use our Travis pip slot to test the latest PyQt 5 (which is 5.8 right now) because conda is going to stay at 5.6 for a loong time ;-)

setup.py
```diff
@@ -288,8 +288,15 @@ def run(self):
     'numpydoc',
 ]
 
+extras_require = {
+    'test:python_version == "2.7"': ['mock'],
+    'test': ['pytest', 'pytest-qt', 'pytest-cov', 'mock', 'coveralls',
+             'ciocheck', 'flaky', 'pandas', 'scipy', 'sympy', 'pillow'],
```
ccordoba12: `ciocheck` is not in PiPy. Please remove it.

continuous_integration/travis/install.sh
```diff
@@ -18,6 +18,7 @@ download_code()
     if [ "$PR" != "false" ] ; then
         cd $FULL_SPYDER_CLONE
         git fetch origin pull/$PR/head:travis_pr_$PR
+        git checkout travis_pr_$TRAVIS_PULL_REQUEST
```
ccordoba12: This is done in `build_test.sh`, so you don't need to do it here:

https://github.com/spyder-ide/spyder/blob/master/continuous_integration/travis/build_test.sh#L10

setup.py
rlaverde: but it is listed in pipy https://pypi.python.org/pypi/ciocheck/0.1.1

continuous_integration/travis/install.sh
```diff
@@ -61,14 +62,10 @@ install_conda()
...
-    conda install 'pyqt>=5.6.0';
+    pip install pyqt5;
 
     # Install testing packages
-    pip install pytest pytest-cov pytest-qt mock
-
-    # Install extra packages
-    EXTRA_PACKAGES="matplotlib pandas sympy pyzmq pillow"
-    pip install $EXTRA_PACKAGES
+    pip install -e .[test]
```
ccordoba12: This is not needed here. Please do it after the wheel has been built, i.e. here:

https://github.com/spyder-ide/spyder/blob/master/continuous_integration/travis/build_test.sh#L10

rlaverde: ok, this will allow to delete the duplicated checkout

continuous_integration/travis/install.sh
rlaverde: yes, but if I try to run `pip install -e .[test]` It'll fail because It was in the master branch

setup.py
ccordoba12: But Travis is complaining about the lack of a package for ciocheck.

In any case, pytest doesn't run ciocheck, so it's not really a test dependency.

continuous_integration/travis/run_test.sh
```diff
@@ -26,6 +26,9 @@ if [ "$USE_CONDA" = true ] ; then
 else
     cd $FULL_SPYDER_CLONE
     pip install dist/spyder-*.whl
+
+    # Install testing packages
+    pip install -e .[test]
```
ccordoba12: Is this line necessary? I mean, what I want is to install our wheel and see if it pulls our test dependencies. So something like

     pip install dist/spyder-*.whl[test]

should work (but I'm not sure :-)

rlaverde: I'm not sure neither :see_no_evil:

I added a commit to see if this work

rlaverde: It didn't work because `dist/spyder-*.whl` is a filename

rlaverde:
@ccordoba12 I think the installation using pip is partially working, but I'm blocked because the issue with QtWebEngineWidgets not avalaible in debian/ubuntu

continuous_integration/travis/install.sh
```diff
@@ -61,14 +61,7 @@ install_conda()
...
 {
     # Install PyQt5
-    conda install 'pyqt>=5.6.0';
-
-    # Install testing packages
-    pip install pytest pytest-cov pytest-qt mock
-
-    # Install extra packages
-    EXTRA_PACKAGES="matplotlib pandas sympy pyzmq pillow"
-    pip install $EXTRA_PACKAGES
```
ccordoba12: Please revert this change for now. I have to plan on how to start using the pyqt5 wheels on CircleCI.

continuous_integration/travis/run_test.sh
ccordoba12: I'll fix this one before merging this PR ;-)

continuous_integration/travis/install.sh
rlaverde: I only revert the installation of pyqt (to install it with conda), the other dependencies will be installed by `pip install -e .[test]`

ccordoba12: You're totally right :-)

ccordoba12:
@ghisvail, I think this could be useful for you guys if you want to run our tests in the Debian infrastructure.

ghisvail:
Could be, assuming you guys distribute your tests following the pytest [guidelines](http://doc.pytest.org/en/latest/goodpractices.html#choosing-a-test-layout-import-rules). The Debian packaging is not currently CI tested on the Debian infrastructure.

Another thing which could help would be to verify whether `spyder` can be tested with `pytest-xvfb` instead of manual calls to `xvfb` / `xvfb-run`. The latter are known to be occasionally unreliable on parallel execution setups in our experience. Same comment applies to `qtpy` and `qtawesome`. It worked for `spyder-unittest`.

ccordoba12:
> assuming you guys distribute your tests following the pytest guidelines

I think we are, by "inlining test directories into [our] application package"

> Another thing which could help would be to verify whether spyder can be tested with pytest-xvfb

CircleCi always starts a xvfb instance, so we haven't seen the need of using pytest-xvfb. However, I'm not opposed to idea :-) Could you open a PR to see how that could work?

PR #4198, opened 2017-02-24
PR : Add Cython files import and run support

JGoutin:
This PR improve Cython support by adding ".pyx" importing and running abilities to consoles.
- Automatically compile ".pyx" files when importing them with `import` (Enable `pyximport` with file reload option and NumPy support).
- Allow to run ".pyx" file with the "Run" button and the `runfile ` function (Using `cython_inline` function).
- Automatically enable IPython Cython magic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,7 @@ a Python version greater than 2.7 (Python 3.2 is not supported anymore).
 * **Numpy**: View and edit two or three dimensional arrays in the Variable Explorer.
 * **SymPy**: Symbolic mathematics in the IPython console.
 * **SciPy**: Import Matlab workspace files in the Variable Explorer.
+* **Cython**: Run Cython files in the IPython console.
 
 
 ## More information
--- a/circle.yml
+++ b/circle.yml
@@ -11,7 +11,7 @@ machine:
     CONDA_DEPENDENCIES_FLAGS: "--quiet"
     CONDA_DEPENDENCIES: >
       rope jedi pyflakes sphinx pygments pylint pep8 psutil nbconvert qtawesome pickleshare
-      qtpy pyzmq chardet mock nomkl pandas pytest pytest-cov numpydoc scipy
+      qtpy pyzmq chardet mock nomkl pandas pytest pytest-cov numpydoc scipy cython
     PIP_DEPENDENCIES: "coveralls pytest-qt flaky"
 
 dependencies:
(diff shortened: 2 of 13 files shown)


ccordoba12:
This is a really great addition, thanks a lot for it!! I think the only thing missing is a test for this.

For that, please take a look at `test_run_code` in `spyder/app/tests/test_mainwindow.py` to see how to create one for Cython. You also need to add `cython` as a conda dependency in `circle.yml`.

ccordoba12:
And your test should go to `spyder/app/tests/test_mainwindow.py` too.

JGoutin:
`ImportError: No module named 'pyx_lib'`
Is there a way to add this path to `sys.path` on active IPython console ?

ccordoba12:
I'll help you to fix your test in a couple of days.

JGoutin:
Thanks !

ccordoba12:
@JGoutin, I changed your code to run Cython code using the Cython magic instead of `cython_inline` (in `sitecustomize.py`).

This allows to run any Cython file (with `cython_inline` you can't run a file that starts with `cdef` or `cpdef`) and also get its results injected in the console namespace (which is why your tests were failing before, because `cython_inline` doesn't do that).

However, this approach doesn't take into account the variables defined in the current namespace. But I think that's a minor price to pay for having things to work in a more intuitive way.

Please let me know what you think :-)

JGoutin:
Yes, `cython_inline` seem to have some limitations. Using the magic is better, I agree.

A solution may be to run the compiled Cyhon file in the console (generated with `pyximport.pyxbuild.pyx_to_dll`). Actually, I don't find a way to run the file, but I am still searching.

ccordoba12:
Tests are passing (ignore Travis, it's down now), so this is ready from my point of view.

@JGoutin?

goanpeca:
@ccordoba12 maybe we can create a `spyder-extras`? or somethin metapackage to include all the optional things?

ccordoba12:
> or something conda metapackage to include all the optional things?

Yeah, it could be nice to have it for `conda`, but I wouldn't want to change things for `pip` ;-)

JGoutin:
It's OK for me. If I found an improvement, I'll create a new PR.

And thanks for fixing tests !

ccordoba12:
> It's OK for me. If I found an improvement, I'll create a new PR.

Please make new PRs against the `3.x` branch. That's because we're going to release your work in Spyder 3.2 (`master` is for Spyder 4.0)

goanpeca:
@JGoutin thanks for this contribution :-). Would you be interested in writing/cowriting a small blog post to showcase this feature?

JGoutin:
@goanpeca Sorry, I am not interested by writing something. 😄

Issue #2627, opened 2015-08-19
Add optional display of current class/function that you're in.


dougthor42:
# Summary:

Sometimes you end up having classes with a lot of different methods, or functions that are too long.

When editing these situations, it can be pretty easy to forget which class you're in (especially if the class has a lot of override methods...).
# Proposal:

Add a small display (2 lines) at the top of the editor which shows the class and function/method definition line of the block that the cursor is in.
# Example:

![spyder_current_class function_display_example](https://cloud.githubusercontent.com/assets/5386897/9363584/42adb7ce-465c-11e5-9e6f-327becb09dcf.png)

Note the two different-background-colored lines at the top of the editor. The line number backgrounds are also slightly different colors, both from the line number area and the code editor area. Also note where the cursor is (line 3334) - this defines what is displayed. Sorry for the poor Paint skills...
# Details:

The topmost line of the editor display shows the class that you're currently working on.

The 2nd line displays the function/method that your cursor is currently in. In the example, the function is short and fits on the screen, so this is not really providing any useful information. If the method didn't fit on the display, then this would be useful.
## Visuals

Both lines should have some visual indicator which makes it obvious that they are code in another part of the program. Both lines should also include the line number.

In the example, I've colored the background differently. This has an advantage of not using any additional screen space. If using a different background color, it should be settable in the color schemes of course.

Another option would be to have a horizontal bar or two (but that's up to 10 extra pixels!) or to even combine the two.
### Show and Hide

When showing or hiding these indicators, the remaining code should not "jump down" or "jump up" - these two lines should essentially overwrite the top two lines of the editor.

Compare the **off**:
![spyder_current_class function_display_example_off](https://cloud.githubusercontent.com/assets/5386897/9363897/09cdfcbe-465e-11e5-8fcb-794ee661faef.png)

to **On**:
![spyder_current_class function_display_example_on](https://cloud.githubusercontent.com/assets/5386897/9363898/0d25cacc-465e-11e5-8830-2ec694d92c8d.png)
### Clicking

When clicking on a line, the cursor should be moved to the class or function/method definition (just like selecting it in the outline widget).
### Cases

This part requires some feedback. How should this act in the following situations? My choices are in bold.
1. Cursor not inside a class or function/method (so, at the module level)
   - both lines not shown (so, same as current editor)
   - both lines empty (but the background/separator is still there)
   - **1st line displays `<module>` (or something to that effect) with line number 0, 2nd line empty**
[...]

PR #4225, opened 2017-03-03
PR: Add a new panel to show/explore class and methods/functions present in the current file

dougthor42:
### Summary
This PR adds a class and function dropdown to the top of the editor and below the tabs. This implements #2627.
image

The idea is to mimic Visual Studio's feature:
image

**This PR is still a WIP**. I have the very basics of the widget done, but I'm having trouble with:
1. how to integrate it with the editor
2. how to connect the signal to the cursor_move event.

> Note: This PR is currently based against the `v3.1.3` tag. Let me know if this should be changed.

### (Planned) Features
+ Displayed class and method names update as the cursor moves. It will always display the current class and method that the cursor is in.
  + If not in a class or method, "\<None\>" will be displayed.
  + Optionally show line number and part of the call signature
+ Class dropdown is populated with all classes in the file
  + **To discuss:** sorted by line number or by class name?
+ Method dropdown is populated with all methods belonging to the class that the cursor resides in
  + **To discuss:** also show module-level functions?
  + **To discuss:** show all functions and methods, not just those belonging to the class? (I vote no on this)
+ User can click on a dropdown and select a class or method/function. Cursor jumps to definition

#### 1: How to integrate it with the editor?
I'm having a hard time figuring out how to insert the dropdown widget between the Tabs and the Editor itself. I'm currently just putting it below the editor for testing purposes.

It looks like I should be adding it to the `spyder/widgets/sourcecode/CodeEditor` class, but that inherits from `TextEditBaseWidget` and in order to set a layout, it needs to be a `QWidget`.

I could potentially make a wrapper around CodeEditor... would that be the best way to go?

The discussion long ago on #2627 was to overlay the class/method display and a borderless Widget was discussed - should I look into that more?

#### 2: How to connect the signals?
I think this will depend pretty significantly on how it's integrated into the editor. Would the best thing to do be to piggyback on the `spyder/widgets/sourcecode/CodeEditor.sig_cursor_position_changed` signal?

### Other Comments
+ I'll be squashing things into a reasonable number of commits before removing WIP.
+ I'll revert the changes to `.gitignore` before removing WIP.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,5 +20,3 @@ spyder_crash.log
 
 # git .orig files
 *.orig
-
-.venv*
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ spyder_crash.log
 
 # git .orig files
 *.orig
+
+.venv*
--- a/spyder/widgets/editor.py
+++ b/spyder/widgets/editor.py
@@ -515,6 +515,7 @@ class EditorStack(QWidget):
         self.auto_unindent_enabled = True
         self.indent_chars = " "*4
         self.tab_stop_width_spaces = 4
+        self.show_class_func_dropdown = True
         self.help_enabled = False
         self.default_font = None
         self.wrap_enabled = False
@@ -841,6 +842,7 @@ class EditorStack(QWidget):
         self.title = text
 
     def set_classfunc_dropdown_visible(self, state):
+        self.show_class_func_dropdown = state
         if self.data:
             for finfo in self.data:
                 if finfo.editor.is_python_like():
@@ -1899,7 +1901,8 @@ class EditorStack(QWidget):
                 indent_chars=self.indent_chars,
                 tab_stop_width_spaces=self.tab_stop_width_spaces,
                 cloned_from=cloned_from,
-                filename=fname)
+                filename=fname,
+                show_class_func_dropdown=self.show_class_func_dropdown)
         if cloned_from is None:
             editor.set_text(txt)
             editor.document().setModified(False)
--- a/spyder/widgets/editor.py
+++ b/spyder/widgets/editor.py
@@ -840,6 +840,12 @@ class EditorStack(QWidget):
     def set_title(self, text):
         self.title = text
 
+    def set_classfunc_dropdown_visible(self, state):
+        if self.data:
+            for finfo in self.data:
+                if finfo.editor.is_python_like():
+                    finfo.editor.classfuncdropdown.setVisible(state)
+
     def __update_editor_margins(self, editor):
         editor.linenumberarea.setup_margins(linenumbers=self.linenumbers_enabled,
                              markers=self.has_markers())
--- a/spyder/widgets/panels/classfunctiondropdown.py
+++ b/spyder/widgets/panels/classfunctiondropdown.py
@@ -236,11 +236,26 @@ def update_selected_cb(parents, combobox):
 
 class FoldScopeHelper(object):
     """
+    This is a helper class to make using FoldScope easier.
+
+    It is a wrapper around a FoldScope object and an OutlineExplorerData
+    object, raises certain attributes to this level, and gives them more
+    descriptive names.
+
+    It defines some nicely-formatted ``str`` and ``repr`` and allows
+    for easy tracking of the parents of a given FoldScope.
 
     Parameters
     ----------
     fold_scope : :class:`spyder.widgets.sourcecode.folding.FoldScope`
     oed : :class:`spyder.utils.syntaxhighlighters.OutlineExplorerData`
+
+    Properties
+    ----------
+    parents : list of :class:`FoldScopeHelper`
+        The parents of this ``FoldScopeHelper`` object. The 1st index will
+        be the top-level parent defined at the module level while the
+        last index will be the class or funtion that contains this object.
     """
 
     def __init__(self, fold_scope, oed):
--- a/spyder/widgets/panels/classfunctiondropdown.py
+++ b/spyder/widgets/panels/classfunctiondropdown.py
@@ -141,6 +141,33 @@ def _adjust_parent_stack(fsh, prev, parents):
         pass
 
 
+def _split_classes_and_methods(folds):
+    """
+    Split out classes and methods into two separate lists.
+
+    Parameters
+    ----------
+    folds : list of :class:`FoldScopeHelper`
+        The result of :func:`_get_fold_levels`.
+
+    Returns
+    -------
+    classes, functions: list of :class:`FoldScopeHelper`
+        Two separate lists of :class:`FoldScopeHelper` objects. The former
+        contains only class definitions while the latter contains only
+        function/method definitions.
+    """
+    classes = []
+    functions = []
+    for fold in folds:
+        if fold.def_type == OED.FUNCTION_TOKEN:
+            functions.append(fold)
+        elif fold.def_type == OED.CLASS_TOKEN:
+            classes.append(fold)
+
+    return classes, functions
+
+
 class FoldScopeHelper(object):
     """
 
--- a/spyder/widgets/panels/classfunctiondropdown.py
+++ b/spyder/widgets/panels/classfunctiondropdown.py
@@ -8,7 +8,10 @@
 Editor widget based on QtGui.QPlainTextEdit
 """
 import operator
+import copy
 
+from spyder.widgets.sourcecode.folding import FoldScope
+from spyder.utils.editor import TextBlockHelper
 from spyder.utils.syntaxhighlighters import OutlineExplorerData as OED
 from spyder.utils import icon_manager as ima
 
@@ -64,6 +67,49 @@ def populate(combobox, data):
             combobox.addItem(icon, fqn, item)
         else:
             combobox.addItem(fqn, item)
+
+
+def _get_fold_levels(editor):
+    """
+    Return a list of all the class/function definition ranges.
+
+    Parameters
+    ----------
+    editor :
+
+    Returns
+    -------
+    folds : list of :class:`FoldScopeHelper`
+        A list of all the class or function defintion fold points.
+    """
+    block = editor.document().firstBlock()
+    oed = editor.get_outlineexplorer_data()
+
+    folds = []
+    parents = []
+    prev = None
+
+    while block.isValid():
+        if TextBlockHelper.is_fold_trigger(block):
+            try:
+                data = oed[block.firstLineNumber()]
+
+                if data.def_type in (OED.CLASS, OED.FUNCTION):
+                    fsh = FoldScopeHelper(FoldScope(block), data)
+
+                    # Determine the parents of the item using a stack.
+                    _adjust_parent_stack(fsh, prev, parents)
+
+                    # Update the parents of this FoldScopeHelper item
+                    fsh.parents = copy.copy(parents)
+                    folds.append(fsh)
+                    prev = fsh
+            except KeyError:
+                pass
+
+        block = block.next()
+
+    return folds
 
 
 def _adjust_parent_stack(fsh, prev, parents):
--- a/spyder/widgets/panels/classfunctiondropdown.py
+++ b/spyder/widgets/panels/classfunctiondropdown.py
@@ -397,8 +397,8 @@ class ClassFunctionDropdown(Panel):
         sender = self.sender()
         data = sender.itemData(sender.currentIndex())
 
-        # TODO: don't trigger a _handle_cursor_position_change_event
-        self.editor.go_to_line(data.line + 1)
+        if isinstance(data, FoldScopeHelper):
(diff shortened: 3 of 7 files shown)


goanpeca:
This should probably be a new panel @rlaverde ?

rlaverde:
> This should probably be a new panel

@goanpeca  Yes, I think that will the best way to implement it

@dougthor42

To make a panel you make your widget inhered from [Panels class](https://github.com/spyder-ide/spyder/blob/master/spyder/api/panel.py), and override the methods that you need (I think the only mandatory is  `sizeHint()` so PanelsManager will know how to draw your panel)

```
from spyder.api.panel import Panel

class ClassFunctionDropdown(Panel):
    ...
    def sizeHint(self):
        ....
```

and in CodeEditor add it to [panels manager](https://github.com/spyder-ide/spyder/blob/master/spyder/widgets/sourcecode/codeeditor.py#L270)

```
from spyder.widgets.panels.classfunctiondropdown import  ClassFunctionDropdown

self.panels.register(ClassFunctionDropdown(), Panel.Position.TOP)
```

You can take a look at [panels directory](https://github.com/spyder-ide/spyder/tree/master/spyder/widgets/panels) there are several panels: codefolding, scrollflagarea, and linenumbearea (also edgeline but it doesn't use the api yet) and take them as an example of how to create a new Panel

spyder/widgets/classfunctiondropdown.py
```diff
@@ -0,0 +1,210 @@
+# -*- coding: utf-8 -*-
```
rlaverde: Move this file to panels folder https://github.com/spyder-ide/spyder/tree/master/spyder/widgets/panels

spyder/widgets/classfunctiondropdown.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+def create_widget_list(data):
+    """Create the list of items to populate the dropdown."""
+    # Convert our data to a list and sort it by line num.
+    sorted_data = sorted(list(data.items()))
+    return list(format_for_widget_text(k, v) for k, v in sorted_data)
+
+
+class ClassFunctionDropdown(QWidget):
```
rlaverde: Make this class inherit from `Panel` class

spyder/widgets/editor.py
```diff
@@ -578,6 +579,11 @@ def setup_editorstack(self, parent, layout):
             layout.addWidget(tab_container)
         else:
             layout.addWidget(self.tabs)
+
```
rlaverde: I think that using panel class won't be necessary to modify EditorStack

spyder/widgets/sourcecode/codeeditor.py
```diff
@@ -318,6 +319,40 @@ def get_file_language(filename, text=None):
     return language
 
 
+class CodeEditorWrapper(QWidget):
```
rlaverde: Also this class won't be need

dougthor42:
Rebased to master and switched to using the panel API - made things much easier, thanks guys.

Still a fair amount of work to do:
+ A custom model/view for the QComboBox to make things easier
+ Move `populate_classes_and_methods()` out of `paintEvent` because that happens *way* too often.
+ Get the line jump working, though that shouldn't be too difficult.
+ other stuff.

ccordoba12:
Great work here @dougthor42, thanks a lot for your contribution!

dougthor42:
Well it's now at a usable state. You can jump to classes/methods with ease and the class/method that the cursor is currently in is updated as you move about.

Still to do:
+ subset the method list - I only want to see methods that belong to the class I'm currently in.
+ ~~Add icons? Sure why not - the same ones as the object browser.~~ JK just added. But they still need a bit of work: they're currently not differentiating between methods and functions
+ Documentation
+ Unit test as best I can (where to put these? a new `panels/tests` folder?)
+ cleanup.
+ Out of scope: Optionally sort by line number rather than by class/method name.

ccordoba12:
> Unit test as best I can (where to put these? a new panels/tests folder?

`widgets/tests/test_codeeditor.py` is probably a better place.

You could define a file with some dummy classes and methods, open it in the Editor and query the state of this new panel at different lines to verify that it is showing the right class and method at that point.

> Out of scope: Optionally sort by line number rather than by class/method name.

Please leave that for another PR.

dougthor42:
I'm having some trouble getting the "which definition is the cursor in" code to work - there are so many cases to check against that this thing just keeps growing. I have over 200 asserts across 49 unit tests right now and I keep finding more to check for.

@ccordoba12 Do you already have some code in Spyder that returns which function/class definition a given line number belongs to? Or do you know of a library that does so?

If not, I'll look into using the abstract syntax tree instead of essentially rolling my own. It feels like the parsing has effectively already been done with the `OutlineExplorerData` (which is what I'm building off of). The problem with using the AST is that it will only work with Python.

rlaverde:
> Do you already have some code in Spyder that returns which function/class definition a given line number belongs to? Or do you know of a library that does so?

@dougthor42  I'm not sure, but the outline explorer have a function similar at the one you are implemented, maybe you could reuse some code from the [widget](https://github.com/spyder-ide/spyder/blob/master/spyder/widgets/editortools.py#L183) or [plugin](https://github.com/spyder-ide/spyder/blob/master/spyder/plugins/outlineexplorer.py) (you can activate it using Ctrl+Shit+O)

it gets data form the [syntax highlighter ](https://github.com/spyder-ide/spyder/blob/master/spyder/utils/syntaxhighlighters.py#L275)

dougthor42:
Another proof-of-concept working, this time working off of the stuff you guys already have for folding (`FoldScope`). It took me quite a while to figure out how to use it, but I got there.

Still to do:
+ [ ] sorting of the dropdowns
+ [ ] display fully qualified names (`Class.Method`) or just the function name?
+ [ ] subset the function/method dropdown to only display those that are of the same class.

dougthor42:
@rlaverde Ready for review.

.gitignore
```diff
@@ -20,3 +20,5 @@ spyder_crash.log
 
 # git .orig files
 *.orig
+
+.venv*
```
rlaverde: Please undo this change

spyder/widgets/panels/classfunctiondropdown.py
```diff
@@ -0,0 +1,406 @@
...
+        item = parents[-1]
+        for i in range(combobox.count()):
+            if combobox.itemData(i) == item:
+                combobox.setCurrentIndex(i)
+                break
+
+
+class FoldScopeHelper(object):
+    """
+
```
rlaverde: Add some explanation to the docstring

spyder/widgets/panels/classfunctiondropdown.py
```diff
@@ -0,0 +1,406 @@
...
+    def _handle_cursor_position_change_event(self, linenum, column):
+        self._update_data()
+        self.update_selected(linenum)
+
+
+# ##########################################################################
+# Classes and Functions for Testing.
+# ##########################################################################
+if __name__ == "__main__":
+    print("To demo this panel, run spyder/widgets/sourcecode/codeeditor.py")
```
rlaverde: prints are discouraged in a spyder test, remove it, or add `# spyder: test-skip` to the end of the line

`print("To demo this panel, run spyder/widgets/sourcecode/codeeditor.py") # spyder: test-skip`

spyder/widgets/panels/classfunctiondropdown.py
```diff
@@ -0,0 +1,406 @@
...
+        populate(self.class_cb, self.classes)
+        populate(self.method_cb, self.funcs)
+
+    def combobox_activated(self):
+        """Move the cursor to the selected definition."""
+        sender = self.sender()
+        data = sender.itemData(sender.currentIndex())
+
+        # TODO: don't trigger a _handle_cursor_position_change_event
+        self.editor.go_to_line(data.line + 1)
```
rlaverde: This line is throwing an error when the `<none>` item is selected, I think and `if` should be added:

```
if data:
    self.editor.go_to_line(data.line + 1)
```

spyder/widgets/sourcecode/codeeditor.py
```diff
@@ -284,7 +285,13 @@ def __init__(self, parent=None):
...
         self.linenumberarea = self.panels.register(LineNumberArea(self))
         self.updateRequest.connect(self.linenumberarea.update_)
-
+        
+        # Class and Method/Function Dropdowns
+        self.classfuncdropdown = self.panels.register(
+            ClassFunctionDropdown(self),
+            Panel.Position.TOP,
+        )
+
```
rlaverde: Because this panel only work with python files, I think should be hidden if there is no OE data for the file

Maybe in the `CodeEditor.setup_editor` method

```
setup_editor(...)
    ...
    if not self.is_python_like():
        self.classfuncdropdown.setVisible(False)
```

or maybe in the `ClassFunctionDropdown` class if `self.folds` is empty

rlaverde:
I forget to mention that a rebase or merge with master is necessary to fix CI errors (also remove or skip the print in `classfunctiondropdown.py`)

@goanpeca @ccordoba12 could you also review this PR?

.gitignore
dougthor42: Ah, I said that I would and then I forgot to do so! Thanks for the reminder.

spyder/widgets/panels/classfunctiondropdown.py
dougthor42: I went with something a little more robust:

```
if isinstance(data, FoldScopeHelper):
    blah
```

spyder/widgets/panels/classfunctiondropdown.py
dougthor42: Removed. The instructions were moved up to the module docstring.

spyder/widgets/sourcecode/codeeditor.py
dougthor42: Added. I went with your 1st suggestion of putting it in `setup_editor()`.

dougthor42:
@rlaverde I've made the requested changes and rebased to master.

I am a little unsure about what I did to add the preferences checkbox, so you might want to look at those changes closely. It works, but I'm not sure if there is a better way to do it.

goanpeca:
@dougthor42 could you paste a screenshot of how the preferences dialog looks after the change?

dougthor42:
@goanpeca Here ya go:

image

goanpeca:
Looks good (the code regarding the preferences does too)

rlaverde:
> @rlaverde I've made the requested changes and rebased to master.

Thanks, It looks pretty well, I think this PR is ready 😄 

> I am a little unsure about what I did to add the preferences checkbox, so you might want to look at those changes closely. It works, but I'm not sure if there is a better way to do it.

It was ok, but the preference wasn't loaded when opening spyder, I added a commit to fix it

spyder/widgets/panels/classfunctiondropdown.py
```diff
@@ -0,0 +1,416 @@
...
+A Class and Function Dropdown Panel for Spyder.
+
+To demo this panel, run spyder/widgets/sourcecode/codeeditor.py
+"""
+import operator
+import copy
+
+from qtpy.QtWidgets import (QComboBox,
+                            QHBoxLayout,
+                            )
```
ccordoba12: Please remove parentheses in this import statement and make it fit in a single line.

spyder/widgets/panels/classfunctiondropdown.py
```diff
@@ -0,0 +1,416 @@
...
+            combobox.addItem(fqn, item)
+
+
+def _get_fold_levels(editor):
+    """
+    Return a list of all the class/function definition ranges.
+
+    Parameters
+    ----------
+    editor :
```
ccordoba12: Please complete this line. What type of object is `editor`?

spyder/widgets/panels/classfunctiondropdown.py
```diff
@@ -0,0 +1,416 @@
...
+
+
+class ClassFunctionDropdown(Panel):
+    """
+    Class and Function/Method Dropdowns Widget.
+
+    Parameters
+    ----------
+    editor :
+        The editor to act on.
```
ccordoba12: What type of object is `editor`? Please mention it here.

ccordoba12:
I must say that this is so slick!! Thanks a lot for this addition @dougthor42!!

I left some very simple comments, then I'll merge. But before solving them, please rebase/merge with master to pick up some changes to our testing machinery.

dougthor42:
@ccordoba12 Done. It would have been sooner but you caught me while I was at lunch 😉

ccordoba12:
Great! Thanks a lot for the quick response!

ba05:
Is the line number also supposed to appear? I am not seeing it. Running Spyder 4 beta 1.

ccordoba12:
No, it's not.

PR #4237, opened 2017-03-08
PR: Don't show internet conection errors on startup when trying to check for new releases

### LINKED ISSUE #4217: "Unable to connect to the internet" problem
When spyder starts up, or when I check for update, I get this warning. I am behind a company firewall. I do not know how to configure Spyder to know about our proxy.

I installed Spyder via pip install, and I do not have anaconda. From a browser I can get to "https://api.github.com/repos/spyder-ide/spyder/releases" without any problem.

Thank you.

* Spyder Version: 3.1.3
* Python Version: 2.7.13
* Qt Version: 4
* PyQt Version: 4.11.4
* Operating system: Windows 7

## Dependencies

jedi =0.9.0      :  0.9.0 (OK)
matplotlib >=1.0 :  1.5.2 (OK)
pyflakes >=0.5.0 :  1.5.0 (OK)
pylint >=0.25    :  1.6.5 (OK)
sphinx >=0.6.6   :  1.5.3 (OK)

rlaverde:
Fixes #4217

### DIFF
--- a/spyder/workers/updates.py
+++ b/spyder/workers/updates.py
@@ -32,11 +32,12 @@ class WorkerUpdates(QObject):
     """
     sig_ready = Signal()
 
-    def __init__(self, parent):
+    def __init__(self, parent, startup):
         QObject.__init__(self)
         self._parent = parent
         self.error = None
         self.latest_release = None
+        self.startup = startup
 
     def check_update_available(self, version, releases):
         """Checks if there is an update available.
@@ -95,5 +96,7 @@ class WorkerUpdates(QObject):
         except Exception:
             error_msg = _('Unable to check for updates.')
 
-        self.error = error_msg
-        self.sig_ready.emit()
+        # Don't show dialog when starting up spyder and an error occur
+        if not self.startup or error_msg is None:
+            self.error = error_msg
+            self.sig_ready.emit()
--- a/spyder/workers/updates.py
+++ b/spyder/workers/updates.py
@@ -97,6 +97,6 @@ class WorkerUpdates(QObject):
             error_msg = _('Unable to check for updates.')
 
         # Don't show dialog when starting up spyder and an error occur
-        if not self.startup or error_msg is None:
+        if not (self.startup and error_msg is not None):
             self.error = error_msg
             self.sig_ready.emit()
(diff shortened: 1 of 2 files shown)


spyder/workers/updates.py
```diff
@@ -95,5 +96,7 @@ def start(self):
         except Exception:
             error_msg = _('Unable to check for updates.')
 
-        self.error = error_msg
-        self.sig_ready.emit()
+        # Don't show dialog when starting up spyder and an error occur
+        if not self.startup or error_msg is None:
```
ccordoba12: Shouldn't this be `... or error_msg is not None`?

rlaverde: Hmm no, the idea is not to show the dialog when there is an error:

Maybe it will be easy to understand:
`if not (self.startup and error_message is not None):`

Or 
`if self.startup and error_message is not None:
    return`

ccordoba12: > Hmm no, the idea is not to show the dialog when there is an error

I don't agree. The idea (for me) is to not show the dialog if there are errors but only at startup (because it's very annoying). However, if you want to manually check for updates, Spyder should an error message in case it can't do it.

rlaverde: Yes, that's the idea, the expression will evaluate `True` when startup is `False` (regardless if `error_message` is `None` or not)

ccordoba12: Ok, but if `error_msg` is *different* from `None`, we also want to show the dialog. And that's not the case here because we enter this block if `error_msg` is `None` (which is a contradiction for me and what motivated me to write my first comment).

spyder/workers/updates.py
```diff
@@ -95,5 +96,7 @@ def start(self):
         except Exception:
             error_msg = _('Unable to check for updates.')
 
-        self.error = error_msg
-        self.sig_ready.emit()
+        # Don't show dialog when starting up spyder and an error occur
+        if not (self.startup and error_message is not None):
```
ccordoba12: `error_message` -> `error_msg`

rlaverde: Fixed 😄

PR #4251, opened 2017-03-12
PR: Add pytest-xvfb to our test dependencies

ccordoba12:
@ghisvail, this is what you wanted :-)


ghisvail:
Absolutely. That would be a great step towards enabling CI within Debian. Thanks for implementing this.

Issue #4141, opened 2017-02-13
Code completion breaks with  Spyder 3 for modules added through the Python path manager


arita37:
## Description

**What steps will reproduce the problem?**
1.  Jedi cannot starts auto-complete on custom module, some modules (numpy, pandas can auto-complete).
2. 
3. 

**What is the expected output? What do you see instead?**
Spyder 2.3  and Python 2.7 work well and Jedi auto-complete does not break.

**Please provide any additional information below**

## Version and main components
* Spyder Version: 3.0.0 
* Python Version: 2.7.12
* Qt Versions: 5.6.0, PyQt5 5.6 on Windows

## Dependencies
```
jedi >=0.8.1     :  0.8.1 (OK)
matplotlib >=1.0 :  1.5.3 (OK)
nbconvert >=4.0  :  4.2.0 (OK)
numpy >=1.7      :  1.11.3 (OK)
pandas >=0.13.1  :  0.19.2 (OK)
pep8 >=0.6       :  1.7.0 (OK)
pyflakes >=0.5.0 :  1.3.0 (OK)
pygments >=2.0   :  2.1.3 (OK)
pylint >=0.25    :  1.5.4 (OK)
qtconsole >=4.2.0:  4.2.1 (OK)
rope >=0.9.4     :  0.9.4 (OK)
sphinx >=0.6.6   :  1.4.6 (OK)
sympy >=0.7.3    :  1.0 (OK)


```

ccordoba12:
Please update Jedi to 0.9.0 and try again (Spyder doesn't work with Jedi 0.10.0 yet).

arita37:
Jedi 0.9.0 seems not working at all.

Jedi 0.8.1 works in this case:
    module is inside sites-packages  (like numpy)

If I add my own custom directory in the Python Path manager
  and do import myModule as mm
it does not work....

It used to work with Python 2.7 and Spyder 2.3.1.....

How to investigate the break ?

ccordoba12:
> Jedi 0.9.0 seems not working at all.

That's very strange. All our tests are passing with Jedi 0.9.

> If I add my own custom directory in the Python Path manager
and do import myModule as mm

Thanks for the example. We'll take a look at it.

> It used to work with Python 2.7 and Spyder 2.3.1

Jedi support was introduced in Spyder 2.3.3, so completions were provided by Rope, not Jedi.

arita37:
In this configuration, Jedi works well with custom packages from PythonPathManager:


[...]

PR #4263, opened 2017-03-15
PR: Add Spyder PYTHONPATH to PluginClient

rlaverde:
Fixes: #4141

-------------
I think this is ready, but this is causing plugins to be restarted at startup, (normally 1 time, and 2 times when there is a project open)

It's that ok? or how It could be avoided?

### DIFF
--- a/spyder/plugins/editor.py
+++ b/spyder/plugins/editor.py
@@ -426,7 +426,8 @@ class Editor(SpyderPluginWidget):
         self.toolbar_list = None
         self.menu_list = None
 
-        self.introspector = IntrospectionManager()
+        self.introspector = IntrospectionManager(
+                extra_path=self.main.get_spyder_pythonpath())
 
         # Setup new windows:
         self.main.all_actions_defined.connect(self.setup_other_windows)
@@ -1463,7 +1464,7 @@ class Editor(SpyderPluginWidget):
         for finfo in self.editorstacks[0].data:
             finfo.path = self.main.get_spyder_pythonpath()
         if self.introspector:
-            self.introspector.change_spyder_path(
+            self.introspector.change_extra_path(
                     self.main.get_spyder_pythonpath())
     
     #------ Refresh methods
--- a/spyder/plugins/editor.py
+++ b/spyder/plugins/editor.py
@@ -1462,6 +1462,9 @@ class Editor(SpyderPluginWidget):
     def set_path(self):
         for finfo in self.editorstacks[0].data:
             finfo.path = self.main.get_spyder_pythonpath()
+        if self.introspector:
+            self.introspector.change_spyder_path(
+                    self.main.get_spyder_pythonpath())
     
     #------ Refresh methods
     def refresh_file_dependent_actions(self):
--- a/spyder/utils/introspection/plugin_client.py
+++ b/spyder/utils/introspection/plugin_client.py
@@ -41,7 +41,8 @@ class AsyncClient(QObject):
     received = Signal(object)
 
     def __init__(self, target, executable=None, name=None,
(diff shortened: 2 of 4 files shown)


ccordoba12:
@rlaverde, please add a test for this. From now on I want to see tests for every new features we add.

spyder/plugins/editor.py
```diff
@@ -1462,6 +1462,8 @@ def set_current_filename(self, filename, editorwindow=None):
     def set_path(self):
         for finfo in self.editorstacks[0].data:
             finfo.path = self.main.get_spyder_pythonpath()
+        if self.introspector:
+            self.introspector.change_spyder_path(self.main.get_spyder_pythonpath())
```
ccordoba12: It looks wider then 79 columns.

spyder/utils/introspection/plugin_client.py
```diff
@@ -41,7 +41,7 @@ class AsyncClient(QObject):
     received = Signal(object)
 
     def __init__(self, target, executable=None, name=None,
-                 extra_args=None, libs=None, cwd=None, env=None):
+                 extra_args=None, libs=None, cwd=None, env=None, extra_path=None):
```
ccordoba12: Move `extra_paths=None` to a new line below this one

spyder/utils/introspection/plugin_client.py
```diff
@@ -86,6 +89,11 @@ def run(self):
                     python_path = osp.pathsep.join([python_path, path])
                 except ImportError:
                     pass
+            if self.extra_path:
+                try:
+                    python_path = osp.pathsep.join([python_path] + self.extra_path)
```
ccordoba12: Looks wider than 79

ccordoba12:
> It's that ok? or how It could be avoided?

This happens because Spyder's PYTHONPATH is updated at startup and when changing projects, so i think you should try to restart plugins exactly when it's needed.

ccordoba12:
@rlaverde, your test is failing on Windows. Please skip it there.

rlaverde:
@ccordoba12 I forget that windows use a different path separator, I added a commit to fix that test

spyder/utils/introspection/plugin_client.py
```diff
@@ -77,7 +81,7 @@ def run(self):
         # Set up environment variables.
         processEnvironment = QProcessEnvironment()
         env = self.process.systemEnvironment()
-        if (self.env and 'PYTHONPATH' not in self.env) or DEV:
+        if (self.env and 'PYTHONPATH' not in self.env) or self.env is None:
```
ccordoba12: Please restore `DEV` here. I think it's needed to make completions work when Spyder is started with `bootstrap.py`

ccordoba12:
@rlaverde, how many times our completion plugins are restarted when switching projects?

rlaverde:
> how many times are our completion plugins are restarted when switching projects? Two as before, or just one?

- One restart when switching projects
- No restart at startup
- One restart at startup with a project open (load and then restart)

spyder/utils/introspection/plugin_client.py
rlaverde: hmm no, I tested and this condition evaluate to `True` when using `./bootstrap.py
`

I changed because previously it only evaluates to `True` when `DEV` was `True` (i. e. libraries, and spyder path only will be added in dev mode)

ccordoba12: Ok, thanks for the clarification!

ccordoba12:
The number of restarts are exactly the ones needed. Thanks for improving that!

ccordoba12:
I restarted Circle again to make coveralls to pass before merging.

PR #4290, opened 2017-03-22
PR: Add support for other plugins in File Switcher

dalthviz:
Changes to make the fileswitcher work with the notebookplugin https://github.com/spyder-ide/spyder-notebook/issues/45

### DIFF
--- a/spyder/widgets/fileswitcher.py
+++ b/spyder/widgets/fileswitcher.py
@@ -329,7 +329,7 @@ class FileSwitcher(QDialog):
             if editor is self.initial_editor:
                 self.initial_path = paths[i]
             # This try is needed to make the fileswitcher work with 
-            # plugins that does not have a textCursor like the notebookpuglin.
+            # plugins that does not have a textCursor.
             try:
                 self.initial_cursors[paths[i]] = editor.textCursor()
             except AttributeError:
@@ -442,12 +442,7 @@ class FileSwitcher(QDialog):
         elif path:
             return self.tabs.widget(index)
         else:
-            # this try is needed in order to get
-            # the current client in the notebookplugin
-            try:
-                return self.parent().get_current_editor()
-            except AttributeError:
-                return self.parent().get_current_client()
+            return self.tabs.currentWidget()
 
     def set_editor_cursor(self, editor, cursor):
         """Set the cursor of an editor."""
--- a/spyder/widgets/fileswitcher.py
+++ b/spyder/widgets/fileswitcher.py
@@ -328,7 +328,12 @@ class FileSwitcher(QDialog):
         for i, editor in enumerate(self.editors):
             if editor is self.initial_editor:
                 self.initial_path = paths[i]
-            self.initial_cursors[paths[i]] = editor.textCursor()
+            # This try is needed to make the fileswitcher work with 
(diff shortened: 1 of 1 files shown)


spyder/widgets/fileswitcher.py
```diff
@@ -328,7 +328,12 @@ def save_initial_state(self):
         for i, editor in enumerate(self.editors):
             if editor is self.initial_editor:
                 self.initial_path = paths[i]
-            self.initial_cursors[paths[i]] = editor.textCursor()
+            # This try is needed to make the fileswitcher work with 
+            # plugins that does not have a textCursor like the notebookpuglin.
```
ccordoba12: Please remove from this line `like the notebookpuglin`

spyder/widgets/fileswitcher.py
```diff
@@ -437,7 +442,12 @@ def get_editor(self, index=None, path=None):
         elif path:
             return self.tabs.widget(index)
         else:
-            return self.parent().get_current_editor()
+            # this try is needed in order to get
+            # the current client in the notebookplugin
+            try:
+                return self.parent().get_current_editor()
+            except AttributeError:
+                return self.parent().get_current_client()
```
ccordoba12: I don't like this because it would only work for the Notebook plugin.

Please change this whole block to

```python
self.parent().tabs.currentWidget()
```

because that's what `get_current_editor` returns, and in `NotebookPlugin` rename `self.tabwidget` to `self.tabs`.

That should work for both the Editor and the Notebook without specializing for any particular plugin.

ccordoba12: Sorry, there's no need to change anything in `NotebookPlugin`. Just the first change, i.e.

```python
self.tabs.currentWidget()
```

is enough ;-)

ccordoba12:
What happens when you write `@` in the file switcher associated with the Notebook? That probably gives an error so we need to deactivate it in this PR too.

dalthviz:
@ccordoba12 About the `@` nothing happens, it just shows nothing as an option to select:

fss

goanpeca:
@ccordoba12 @dalthviz we should change the name of the plugin to just Notebook, having jupyter just takes up space.

dalthviz:
@goanpeca :+1:

ccordoba12:
> About the @ nothing happens

Great!! Less work to do ;-)

> we should change the name of the plugin to just Notebook or Notebooks

Since we're going to remove the notebook header, I think it's better to leave *Jupyter notebook* so that people can clearly identify see what's the notebook provenance.

goanpeca:
@ccordoba12 
>Since we're going to remove the notebook header, I think it's better to leave Jupyter notebook so that people can clearly identify what's the notebook provenance.

I don't see us supporting any other type of notebook so why bother, it also goe sin line with the request I made about making shorter names for all plugins,

- `static analysis` -> `linter`
- `Jupyter notebook` -> `Notebooks`
- `Ipython console` -> `Console` etc

ccordoba12:
I agree it's a good idea to rename `Jupyter notebook` to `Notebook` before releasing 0.1. @dalthviz, please do that.

We can do the reanames for Spyder 4.0

PR #4302, opened 2017-03-28
PR: Add a tab switcher dialog to the Editor

### LINKED ISSUE #43: Make Ctrl+Tab to navigate between tabs in Most Recently Used (MRU) order
_From [y.bonne...@gmail.com](https://code.google.com/u/111370336049651795761/) on 2009-09-13T12:00:46Z_

Ideally, Ctrl + Tab could be used to navigate between tabs in Most 
Recently Used order.

You can see MRU in browsers or Notepad++ - very convenient to repeatedly 
switch between two tabs.

The Matlabesque Ctrl+ PgUP/PgDwn becomes painful if the tabs are not 
close, and the Ctrl + Tab is easy to remember, by analogy with the Alt + 
Tab.

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=43_

rlaverde:
Fixes #43

-----------
TODO:
- [x] Fix error when closing a file (stack history isn't updated)

tabs_mru


spyder/widgets/editor.py
```diff
@@ -274,6 +274,70 @@ def breakpoints_changed(self):
...
+        self.editor = parent
+        self.stack_history = stack_history
+        self.tabs = tabs
+
+        self.setSelectionMode(QListWidget.SingleSelection)
+        self.itemActivated.connect(self.item_selected)
+
+        self.id_list = []
+        self.load_data()
+        size = CONF.get('main', 'completion/size')
```
rlaverde: I reused completion widget size widget, is that ok? should I add a setting? or leave a fixed size

spyder/widgets/editor.py
```diff
@@ -1979,6 +2041,18 @@ def dropEvent(self, event):
                 editor.insert_text( source.text() )
         event.acceptProposedAction()
 
+    def keyReleaseEvent(self, event):
+        """Reimplement Qt method.
+
+        Handle "most recent used" tab behavior,
+        When ctrl is released and tab_switcher is visible, tab will be changed.
+        """
+        if (event.key() == Qt.Key_Control and self.tabs_switcher is not None
```
rlaverde: This will break if It's set a shortcut without "Ctrl" (Although I think that will be a really rare use case)

should I add a validation when configuring the shortcut?

spyder/widgets/editor.py
ccordoba12: What if you use the same and positioning of the file switcher?

spyder/widgets/editor.py
ccordoba12: Yes, please add the validation to prevent this possibility.

spyder/widgets/editor.py
rlaverde: I think It's better in the center of the editor (I take inspiration from Kate), at the center you will have to move less the eyes than if it's at the top.

Or maybe something like 1/4(editor height) from top?

ccordoba12: I think it's better to have both widgets (the file switcher and this one) placed at the same location to avoid breaking expectations from users (i.e. where to expect each dialog in the screen). Since the file switcher came first, this one has to follow it.

ccordoba12: Besides, that's what VSCode also does, i.e. place both widgets at the same location.

spyder/widgets/editor.py
```diff
@@ -274,6 +274,70 @@ def breakpoints_changed(self):
             self.save_breakpoints.emit(self.filename, repr(breakpoints))
 
 
+class TabSwitcherWidget(QListWidget):
+    """Show tabs in mru order and change between them."""
```
goanpeca: one space here

spyder/widgets/editor.py
goanpeca: Yes, on the top is better for consistency. But we can evaluate the UX of all those dialogs in a separate issue

spyder/widgets/editor.py
```diff
@@ -274,6 +274,70 @@ def breakpoints_changed(self):
...
+        Add elements in inverse order of stack_history.
+        """
+        # Fill a list to track ids and corresponding tab index
+        self.id_list = [id(self.tabs.widget(i))
+                        for i in range(self.tabs.count())]
+
+        for index in self.stack_history[::-1]:
+            text = self.tabs.tabText(self.id_list.index(index))
+            item = QListWidgetItem(ima.icon('FileIcon'), text)
+            self.addItem(item)
```
goanpeca: As discussed @rlaverde lets move all the stack history logic into a separate class (StackHistory)

spyder/widgets/editor.py
```diff
@@ -274,6 +274,70 @@ def breakpoints_changed(self):
...
+
+    def item_selected(self, item=None):
+        """Change to the selected document and hide this widget."""
+        if item is None:
+            item = self.currentItem()
+
+        # stack history is in inverse order
+        id_ = self.stack_history[-(self.currentRow()+1)]
+
+        index = self.id_list.index(id_)
```
goanpeca: Lets remove this logic and put inside the `StackHistory` widget

spyder/widgets/editor.py
rlaverde: I'll add an option for also considering Alt, I think the shortcuts validation could be addressed in other Issue in a more general and reusable way

goanpeca: Agreed!

ccordoba12:
@rlaverde, please make my suggested changes so I can merge this one.

rlaverde:
> @rlaverde, please make my suggested changes so I can merge this one.

I already moved it to the top, and the shortcuts validation should be address in a more general way in another PR

spectacle z15337

spyder/widgets/editor.py
```diff
@@ -274,6 +275,111 @@ def breakpoints_changed(self):
             self.save_breakpoints.emit(self.filename, repr(breakpoints))
 
 
+class StackHistory(MutableSequence):
+    """Handles editor stack history"""
```
goanpeca: period at the end :-), remeber docstrings are sentences

spyder/widgets/editor.py
```diff
@@ -274,6 +275,111 @@ def breakpoints_changed(self):
...
+                        for _i in range(self.editor.tabs.count())]
+
+    def refresh(self):
+        """Remove editors that are not longer open."""
+        self._update_id_list()
+        for _id in self.history[:]:
+            if _id not in self.id_list:
+                self.history.remove(_id)
+
+    def __len__(self): return len(self.history)
```
goanpeca: split in 2 lines

spyder/widgets/editor.py
```diff
@@ -274,6 +275,111 @@ def breakpoints_changed(self):
...
+
+    def __getitem__(self, i):
+        return self.id_list.index(self.history[i])
+
+    def __delitem__(self, i): del self.history[i]
```
goanpeca: same here

spyder/widgets/editor.py
```diff
@@ -274,6 +275,111 @@ def breakpoints_changed(self):
...
+
+    def __setitem__(self, i, v):
+        _id = id(self.editor.tabs.widget(v))
+        self.history[i] = _id
+
+    def __str__(self):
+        return str(list(self))
+
+    def insert(self, i, v):
+        _id = id(self.editor.tabs.widget(v))
```
goanpeca: docstring :-p ?

spyder/widgets/editor.py
```diff
@@ -274,6 +275,111 @@ def breakpoints_changed(self):
...
+        self.history.insert(i, _id)
+
+    def remove(self, index):
```
goanpeca: docstring?

spyder/widgets/editor.py
```diff
@@ -274,6 +275,111 @@ def breakpoints_changed(self):
...
+
+    def select_row(self, steps):
+        """Move selected row a number of steps.
+
+        Iterates in a cyclic behaviour.
+        """
+        row = (self.currentRow() + steps) % self.count()
+        self.setCurrentRow(row)
+
+    def set_dialog_position(self):
```
goanpeca: Is there a way we can generalize this method, so that it can be used by the fileswitcher and this one, and any other pop up dialog we use?

Maybe have a base PopUpDialog class that handles this based on the parent?

rlaverde: hmm yes, ~maybe reusing some of the Modes logic introduced with editor panels~

```
class PopUpDialog(QWidget):
    ...
    def set_dialog_position(self, top, left):
```


top and left will be numbers from 0 to 1 than indicate the position i.e-:

(0.5, 0.5) will be the center
(0, 0,5) will be the top-center

goanpeca: Yeah something like that would be vey useful... probably out of the scope of this PR, but you could do it afterwards? please create an issue for it :-)

ccordoba12:
@rlaverde, please merge with `3.x` to fix the error in CircleCI.

spyder/widgets/editor.py
```diff
@@ -274,6 +275,119 @@ def breakpoints_changed(self):
             self.save_breakpoints.emit(self.filename, repr(breakpoints))
 
 
+class StackHistory(MutableSequence):
```
ccordoba12: Please move this to `spyder/utils/editor.py`

ccordoba12: Sorry, don't worry about it. We simply need to split this file in multiple files because it's huge!!

But let's do that in `master` (I'll do it).

ccordoba12: I mean, I'll do it after we release 3.2

ccordoba12:
I found an annoying behavior: some files get an ampersand in their name:

seleccion_002

@rlaverde, please fix that and then I'll merge ;-)

rlaverde:
@ccordoba12 I'm not able to reproduce that behavior, any hint of what is causing it?

ccordoba12:
It happens in some versions of Qt5, I think. Don't worry, it was a simple fix and I made it myself ;-)

PR #4326, opened 2017-04-05
Hungarian language support

blackPantherOS:
Hi! We added initial Hungarian support and pre-translated files


spyder/config/base.py
```diff
@@ -229,6 +229,7 @@ def get_image_path(name, default="not_found.png"):
 LANGUAGE_CODES = {'en': u'English',
                   'fr': u'Français',
                   'es': u'Español',
+                  'hu': u'Hungarian',
```
goanpeca: Thanks a lot for this :-)

Now, the idea here is to use the Native language name, I am guessing `Magyar` ?

ccordoba12: Yep, please use here the name for Hungarian in Hungarian itself.

ccordoba12:
@blackPantherOS, thanks a lot for this!! It's a great contribution.

Unfortunately, since you did your translation against our `master` branch, it will be available when we release Spyder **4**, and not in any of our **3.x** releases.

I hope that's ok for you.

spyder/config/base.py
blackPantherOS: Hungarian in English and Magyar in Hungarian :) I think fine both if either available

ccordoba12: Our convention is to use the native language name, as you can see with the other languages.

So please make our requested change.

ccordoba12:
@blackPantherOS, one last question: would you be available in the future to keep updating this translation?

goanpeca:
@martindurant want to help ?

blackPantherOS:
@ccordoba12 Sorry because 
I'm too late that I write any response. We have more thousand other project in blackPanther OS. Not have too many time for one project and translation, time to time we would like to help you

ccordoba12:
Closing in favor of #4473. @blackPantherOS, don't worry about it, your work will be available in Spyder 3.2!

PR #4341, opened 2017-04-08
PR: Make jedi_patch.py to work with Jedi 0.10.x versions

ccordoba12:
Pinging @ghisvail about this one.


ghisvail:
Could you detail a bit more what this PR solves please?

I thought #4121 already fixed Jedi autocompletion...?

Cheers,

ccordoba12:
Jedi `0.10.2` was released some days ago and Spyder was not able to work it because #4121 was *only* using Jedi `0.10.0`.

I thought that was fine because Jedi take some much time to do new releases. But that was not the case with `0.10`.

ghisvail:
Oh I see now. Thanks for heads-up.

PR #4352, opened 2017-04-11
PR: Add Markdown syntax highlighter.

rlaverde:
As a part of the development of https://github.com/spyder-ide/spyder-reports is necessary to add markdown support to the editor


ccordoba12:
Thanks a lot for this @rlaverde!! Could you post a screenshot of a Markdown file with code blocks? Thanks!

rlaverde:
@ccordoba12 
markdown sh

I think this is kinda ready,

Although I've run into an issue: there isn't and standard way (maybe an api) to extend BaseSH, and different languages need different keys, styles for that keys

i.e.:
- python have: keyword, string, number, builtin, definition...
- md have: titles, links, strike-through, italic, bold, code, lists... 

I'd to do an workaround to add new keys-styles, but that's styles aren't ~~fully~~ configurable

cc @goanpeca

spyder/utils/syntaxhighlighters.py
```diff
@@ -910,6 +910,130 @@ class HtmlSH(BaseWebSH):
...
+                            self.setFormat(0, len(text),
+                                           self.formats["normal"])
+                            self.setCurrentBlockState(self.NORMAL)
+                        else:
+                            continue
+                    else:
+                        if key == "code":
+                            # Change to code
+                            self.setFormat(0, len(text), self.formats["code"])
+                            self.setCurrentBlockState(self.CODE)
```
ccordoba12: Could we use here the highlighter corresponding to the code language?

ccordoba12:
> there isn't and standard way (maybe an api) to extend BaseSH

I think Markdown is a special case, so let's leave things as they are for now.

----

In your screenshot, the last line is not highlighted correctly: `but` is in green when it should be in black.

spyder/utils/syntaxhighlighters.py
rlaverde: That's a nice idea, I'm not sure, I'll try to do it

rlaverde:
> I think Markdown is a special case

I don't think so, It'll happen with other markup languages (html, rest), or config files (ini, yaml), also with some frameworks (jinja), also with other programming languages that aren't python-like.

>so let's leave things as they are for now.

yes, but maybe we could improve this for spyder4

In your screenshot, the last line is not highlighted correctly: but is in green when it should be in black.

Thanks, I didn't notice, I fixed the regex 😄

spyder/utils/syntaxhighlighters.py
goanpeca: @ccordoba12 its a nice idea indeed, but I think we should postpone that to 4.0 as it would probably require a better re-estructuring of syntax highlighters.

I think for what we need this is already a good step

ccordoba12: No problem.

ccordoba12:
> It'll happen with other markup languages (html, rest), or config files (ini, yaml), also with some frameworks (jinja)

We already have support for all those formats (except Jinja). But the main idea is that Markdown is a special case because there's no Pygments support for it.

> yes, but maybe we could improve this for spyder4

Maybe for 4.1, for now let's focus on getting this merged ;-)

----

Please merge with 3.x to fix CircleCI, and then I'll merge.

ccordoba12:
This is ready (and I haven't noticed, sorry :-). Merging.

PR #4370, opened 2017-04-19
PR: Remove misplaced flags from re.sub() call

jwilk:
The 4th argument of `re.sub()` is maximum number of substitutions, not flags.
`re.UNICODE` is the default in Python 3, so it doesn't have to be passed explicitly to re.sub().

Found using [pydiatra](https://github.com/jwilk/pydiatra).


ccordoba12:
Thanks a lot for your contribution and sorry for the delay in coming back to you.

I rebased your work on top of our `3.x` branch, so we can include it in Spyder 3.2 (our next release).

PR #4425, opened 2017-05-03
PR: Expose some widgets and utility functions through the api.

### LINKED ISSUE #4397: Define a section in the API for extras and utilities
I'm opening this issue to make a list of utilities that could be used by plugins and could be exposed in the api

As @goanpeca says [here](https://github.com/spyder-ide/spyder-reports/issues/11#issuecomment-297076695)  "I think we should only allow importing from api when doing a plugin"

At the moment I found the followings functions and modules being used by plugins (and aren't in the API yet):
```
from spyder.config.base import _, get_translation, get_conf_path, DEV
from spyder.config.gui import set_shortcut, config_shortcut

from spyder.utils.qthelpers import create_action, create_toolbutton, add_actions, get_icon

from spyder.widgets.comboboxes import PythonModulesComboBox
from spyder.widgets.variableexplorer.texteditor import TextEditor
from spyder.widgets.browser import WebView, FrameWebView

from spyder.plugins.runconfig import get_run_configuration

from spyder.utils import programs
from spyder.utils.misc import add_pathlist_to_PYTHONPATH, get_python_executable, select_port

from spyder.py3compat import getcwd, to_text_string, is_text_string
```

which ones we will exposes through the api? and how we will organize them?

rlaverde:
Fixes #4397

I just imported functions and classes into the api, should I create dummy functions to expose stuff https://github.com/spyder-ide/spyder/issues/4397#issuecomment-297181157 ?

### DIFF
--- a/spyder/api/widgets.py
+++ b/spyder/api/widgets.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+
+"""
+spyder.api.widgets
+==================
+
+This package exposes shareable widgets that can be used in
+third-party plugins.
+"""
+
+from spyder.widgets.tabs import Tabs
+from spyder.widgets.comboboxes import PythonModulesComboBox
+from spyder.widgets.variableexplorer.texteditor import TextEditor
+from spyder.widgets.browser import WebView, FrameWebView
--- a/spyder/api/__init__.py
+++ b/spyder/api/__init__.py
@@ -14,3 +14,6 @@ third-party plugins to extend Spyder.
 This API should be considered pre-release and is subject to change
 The final API version will be released in Spyder 4.0
 """
+
+# Expose python3 compatibility module through the api
+import spyder.py3compat as py3compat
--- a/spyder/api/utils.py
+++ b/spyder/api/utils.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+
+"""
+spyder.api.utils
+==================
+
+This package exposes utility functions that can be used in
+third-party plugins.
+"""
+
+from spyder.config.base import _, get_translation, get_conf_path, DEV
+from spyder.config.gui import set_shortcut, config_shortcut
+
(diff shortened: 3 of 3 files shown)


spyder/api/widgets.py
```diff
@@ -0,0 +1,18 @@
...
+spyder.api.widgets
+==================
+
+This package exposes shareable widgets that can be used in
+third-party plugins.
+"""
+
+from spyder.widgets.tabs import Tabs
+from spyder.widgets.comboboxes import PythonModulesComboBox
+from spyder.widgets.variableexplorer.texteditor import TextEditor
```
goanpeca: `from spyder.widgets.variableexplorer.texteditor import TextEditor` this is the only one that feels weird here because it imports from an actual "complex widget"

rlaverde: hmm yes, and the name could be confusing with the Editor.

maybe something straightforward  like this:
`from spyder.widgets.variableexplorer.texteditor import TextEditor as VariableExplorerTextEditor`

or create an `api/variableexporer.py` file that will expose an api of the VariableExplorer plugin like the Editor plugin?

goanpeca: >or create an api/variableexporer.py file that will expose an api of the VariableExplorer plugin like the Editor plugin?

This would be the best on the long run.

jitseniesen: Just for context, in the plugins that I am familiar with (line_profiler, memory_profiler, unittest) `TextEditor` is used to show a block of text to the user. The edit functionality is not used at all. In the long run, it may be possible to use a function which is not so deeply coupled.

rlaverde: TextEditor class doesn't depended of VariableExplorer, maybe we could move it outside of the variableexplorer module to the api.

https://github.com/spyder-ide/spyder/blob/master/spyder/widgets/variableexplorer/texteditor.py

goanpeca: >maybe we could move it outside of the variableexplorer module to the api.

Well we need to move it somewhere, not exactly the api... but `helperwidgets` ? or ?

rlaverde: `helperwidgets` seems fine, or maybe its own file `dialogeditor.py`, `simplededitor.py`?

goanpeca: Sure, any of those work, thoughts @ccordoba12 ?

ccordoba12:
Closing for now. We should revisit this once the `split-plugins` branch is merged.

PR #4428, opened 2017-05-05
PR: Use setup.cfg and conftest.py instead of runtest.py

### LINKED ISSUE #4424: Use setup.cfg or pytest.ini instead of runtest.py
The way how spyder tests are executed is a little non-standard

`runtest.py` file could be replaced for a configuration in `setup.cfg` or `pytest.ini`:

`addopts =-x -v -rw --durations=10 --cov=spyder_reports --cov-report=term-missing`

and tests will be executed using:
`$ pytest`
or ` $pytest spyder`

I didn't found a reason to use `runtest.py` file, and using `setup.cfg` or `pytest.ini` is more standard

rlaverde:
Fixes #4424

~This is just a proof of concept, I hope this works 🙈~ It works!! 😄 
~I know I shouldn't be working on this~


setup.cfg
```diff
@@ -0,0 +1,2 @@
+[tool:pytest]
+addopts =-x -v -rw --durations=10 --cov=spyder --cov-report=term-missing
```
jitseniesen: These options are useful (or even necessary) when running the tests in the CI server, but in my opinion they are less useful when a user or developer runs the tests themselves. I think it would be better to move these options to the CI scripts (if possible).

bilderbuchi: >  (if possible)

you could either define an environment variable [`PYTEST_ADDOPTS`](https://docs.pytest.org/en/latest/customize.html#adding-default-options) in the CI scripts, or rename the config file to, e.g., `pytest-ci.cfg` and use the `-c file` command line flag to discover *that* config file.

rlaverde: These options are for:
```
-x exit at the first error
-v verbose
-rw display internal (poytest) warnings
--duration=10 show the 10 most slowest test
--cov --cov-report generate coverage report
```
I don't see any harm in leaving these options as default (when a developer or user run the test) that was the `runtest.py` behavior, also you could expect the same output locally and in CI

ccordoba12: I'm +1 on leaving things as they are.

ccordoba12:
Great job @rlaverde! Before merging, I'd like you to make one of the tests of main window to fail, to see if pytest fails as before.

ccordoba12:
Please ping me when you do it.

goanpeca:
@ccordoba12

ccordoba12:
Unfortunately, tests are hanging up on Windows (instead of finishing with a failure), as you can see in Appveyor (I'm going to kill this build to avoid an unnecessary and time consuming process).

I avoided this in `run_tests.py` by forcing an `exit 1` call on failure but I don't know how to do the same here.

setup.cfg
jitseniesen: I find the options annoying because they produce too much output, making it hard for me to see the information I am usually interested in (do all tests pass, and if not, which tests fail). But if this annoyance isn't shared, then by all means leave things as they are and I'll find a way to override the options if I want.

ccordoba12: @jitseniesen, what's the most annoying option for you?

bilderbuchi:
Maybe with a session-scoped autouse fixture?

ccordoba12:
@bilderbuchi, how that would work?

bilderbuchi:
@ccordoba12 like this:

```python
$ cat conftest.py 
import pytest


@pytest.fixture(scope="session", autouse=True)
def ensure_exit(request):
    """Make sure that a pytest run can exit even if Qt hangs."""
    print('In session fixture: Start')
    yield
    print('In session fixture: After tests have run')
    if request.node.testsfailed:
        print('Some tests have failed. Exiting hard...')
        raise SystemExit(1)

$ cat test_something.py 
def test_passes():
    print('A passing test')
    assert True


def test_fails():
    print('A failing test')
    assert False

$ pytest -s
======================================================== test session starts ========================================================
platform linux -- Python 3.5.3, pytest-3.0.7, py-1.4.33, pluggy-0.4.0
rootdir: <snip>, inifile:
collected 2 items 

test_something.py In session fixture: Start
A passing test
.A failing test
FIn session fixture: After tests have run
Some tests have failed. Exiting hard...
E

============================================================== ERRORS ===============================================================
__________________________________________________ ERROR at teardown of test_fails __________________________________________________

request = <SubRequest 'ensure_exit' for <Function 'test_passes'>>

    @pytest.fixture(scope="session", autouse=True)
    def ensure_exit(request):
        """Make sure that a pytest run can exit even if Qt hangs."""
        print('In session fixture: Start')
        yield
        print('In session fixture: After tests have run')
        if request.node.testsfailed:
            print('Some tests have failed. Exiting hard...')
>           raise SystemExit(1)
E           SystemExit: 1

conftest.py:12: SystemExit
============================================================= FAILURES ==============================================================
____________________________________________________________ test_fails _____________________________________________________________

    def test_fails():
        print('A failing test')
>       assert False
E       assert False

test_something.py:8: AssertionError
============================================ 1 failed, 1 passed, 1 error in 0.03 seconds ============================================
```

I added a couple of prints and ` -s` to make the program flow visible, normally it would be less noisy.

Note that I don't fully understand why you need that in the first place. There might be a more elegant way from within pytest to exit cleanly, without kicking the legs out from under pytest's feet like you seem to do currently, and getting an Error.

ccordoba12:
@rlaverde, please try to use @bilderbuchi's suggestion.

@bilderbuchi, the problem is (I think) pytest could be running things in the background (e.g. closing the main window) when an error is generated, and that doesn't allow it the pytest process to finish on errors.

bilderbuchi:
Hm, could you put together/suggest a minimal example? Then I could take this up with the purest devs, but ud like to have a good failure description before that.

bilderbuchi:
Is this something that only happens on appveyor, or on local windows machines, too? Is there a previous spyder issue about that?

ccordoba12:
Closing for now because our tests are working relatively well.

PR #4443, opened 2017-05-08
PR: Add icon earlier when running application

### LINKED ISSUE #4444: Add icon to mac dock when running with bootstrap

goanpeca:
Fixes #4444

This adds the spyder icon to the mac dock when running from bootstrap.

screen shot 2017-05-08 at 1 53 35 pm

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -78,7 +78,7 @@ from qtpy import API, PYQT5
 from qtpy.compat import from_qvariant
 from qtpy.QtCore import (QByteArray, QCoreApplication, QPoint, QSize, Qt,
                          QThread, QTimer, QUrl, Signal, Slot)
-from qtpy.QtGui import QColor, QDesktopServices, QKeySequence, QPixmap
+from qtpy.QtGui import QColor, QDesktopServices, QIcon, QKeySequence, QPixmap
 from qtpy.QtWidgets import (QAction, QApplication, QDockWidget, QMainWindow,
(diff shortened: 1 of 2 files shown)


ccordoba12:
Great work!! This was something we needed since a long time ago!

A couple of questions:

1. Should we restrict this to macOS only? I mean, this could interfere with how we set our icon on Windows.
2. This doesn't apply only to bootstrap (there's no  `DEV` usage here). So my question is: would this apply to a properly installed Spyder too?

goanpeca:
>1. Should we restrict this to macOS only? I mean, this could interfere with how we set our icon on Windows.

I have no idea :-|, how should I check this?

>2. This doesn't apply only to bootstrap (there's no DEV usage here). So my question is: would this apply to a properly installed Spyder too?

Yes you are right. Well on mac in we need to use the app bundles to get the right behavior, so that is fixed on that side already. Regarding Linux and windows I am not sure.

I could add the DEV flag? I could install 3.1.4 and patch it there on win and linux to see what is the resulting behavior, at least for conda installs.

goanpeca:
Actually @ccordoba12 I think my method is better because it does it earlier in the launch process. Now we do it inside the MainWindow, whereas the way I do it I use [app.setWindowIcon](http://doc.qt.io/qt-5/qguiapplication.html#windowIcon-prop). Maybe I should just remove the way we currently do it and replace with this method.

goanpeca:
@ccordoba12 I think this is working just fine on mac, linux and windows

ccordoba12:
Failures in CircleCI are unrelated to this.

Issue #4316, opened 2017-04-01
Spyder looks terrible on a very high resolution screen


wyntervivaldi:
## Description of your problem
![image](https://cloud.githubusercontent.com/assets/26839804/24579792/7ac37cf8-172e-11e7-8643-f46af1dac2d5.png)
Display scaling is on, but my text on the tabs and the like are really tiny? I also cannot drop the reso as it causes issues on this laptop, my laptop is  new but is super buggy and I really don't have time to exchange it and what not.
**What steps will reproduce the problem?**

1. Leaving the com at original reso
2. 
3. 

**What is the expected output? What do you see instead?**

**Please provide any additional information below**

## Versions and main components

* Spyder Version: 3.1.2
* Python Version: 2.7
* Qt Version: 5.6.2
* PyQt Version: 5.6
* Operating system: Windows 10

## Dependencies

Please go to the menu entry `Help > Optional Dependencies` (or
`Help > Dependencies`), press the button `Copy to clipboard`
and paste the contents below:
jedi >=0.8.1     :  0.9.0 (OK)
matplotlib >=1.0 :  2.0.0 (OK)
nbconvert >=4.0  :  4.2.0 (OK)
numpy >=1.7      :  1.11.3 (OK)
pandas >=0.13.1  :  0.19.2 (OK)
pep8 >=0.6       :  1.7.0 (OK)
pyflakes >=0.5.0 :  1.5.0 (OK)
pygments >=2.0   :  2.1.3 (OK)
pylint >=0.25    :  1.6.4 (OK)
qtconsole >=4.2.0:  4.2.1 (OK)
rope >=0.9.4     :  0.9.4 (OK)
sphinx >=0.6.6   :  1.5.1 (OK)
sympy >=0.7.3    :  1.0 (OK)

I really don't know my way around spyder and I'm a student, so a super idiot proof guide would really help, thanks! I also don't know how to mess around with it much; nor can I ><

ccordoba12:
I have no idea why that's happening, sorry. What's your screen resolution?

juejung:
Same problem on a Dell XPS 13, high dpi laptop under Ubuntu 16.10 with high-dpi scaling on. Ubuntu Unity has a central scaling option for high dpi screens.
The button size in spyder is okay, editor and command window fonts are okay, but the fonts in the window frames and the menu entries are tiny - just like in the screenshot above.
[...]

PR #4454, opened 2017-05-10
PR: Add an option to set a custom HiDpi scale factor

rlaverde:
Fixes: #4316

This is working, but It'll fail when setting variables that could be interpreted as int or floats (`2`, `2.5` but not `2;`, `2.5;`)

It'll be necessary to backport some changes from master (introduced by the multiple edgeline configuration)

- The support for regex in config lineedit options. dc5d4a8e8b58a0e92da48cb9c66ff98ed928382e
- The change for preserving the type of preferences 0c9c97c25ebd0e079cd94e917aed62df41714d46 and 5b26fe81751bdac0bbd58c9302d9f7d24b99c6a2

hidpi resolutpion


goanpeca:
@ccordoba12 @rlaverde why dont use a single option + a QDoubleSpinBox? The default entry would be 1.0, why do we need an extra entry????

screen shot 2017-05-10 at 12 14 33 pm

rlaverde:
Using a QDoubleSpinBox won't be necessary to backport the changes, but it won't allow configuration for multiple screens.

And I think that the differentiation between auto-scaling and custom-scalling is necessary (3 options)

ccordoba12:
> It'll be necessary to backport some changes from master (introduced by the multiple edgeline configuration)

Ok. please backport the corresponding changes, They don't seem too disruptive, so we can have them in 3.2.

rlaverde:
> Ok, please backport the corresponding changes

Done 😄

spyder/config/main.py
```diff
@@ -73,7 +73,11 @@
               'single_instance': True,
               'open_files_port': OPEN_FILES_PORT,
               'tear_off_menus': False,
+              'normal_screen_resolution': True,
               'high_dpi_scaling': False,
+              'high_dpi_custom_scale_factor': False,
+              'high_dpi_custom_scale_factors': '2;',
```
ccordoba12: Please here a float number, e.g. 1.5 or 2.5, so that people understands that they can adjust the scale factor to anything they want.

Also remove the `;` at the end (which doesn't make sense for a single screen).

spyder/plugins/configdialog.py
```diff
@@ -938,6 +940,49 @@ def setup_page(self):
...
+                                'normal_screen_resolution',
+                                button_group=screen_resolution_bg)
+        auto_scale_radio = self.create_radiobutton(
+                                _("Enable auto high DPI scaling"),
+                                'high_dpi_scaling',
+                                button_group=screen_resolution_bg,
+                                tip=_("Set this for high DPI displays"))
+
+        custom_scaling_radio = self.create_radiobutton(
+                                _("Set a custom high DPI scaling"),
```
ccordoba12: This tooltip needs to be more descriptive, i.e. we need to explain here that users can introduce non-integer scale factors and that they can also set factors for different screens, separated by colons.

ccordoba12:
I don't see here any adjustment for the contents of the Help plugin. @rlaverde, is that necessary or not?

spyder/plugins/configdialog.py
goanpeca: We should also add a link to Qt documentation that explains this, like http://doc.qt.io/qt-5/highdpi.html

spyder/config/main.py
rlaverde: A float number could cause glitches, as said in the second question of this [blog](https://blog.qt.io/blog/2016/01/26/high-dpi-support-in-qt-5-6/)

rlaverde:
I added more information of how to set this settings.

spectacle d26270

> I don't see here any adjustment for the contents of the Help plugin. @rlaverde, is that necessary or not?

I forget about this, I'll try to do it, but I don't know how to test it

spyder/config/main.py
ccordoba12: Yeah, but that's something for users to try. The point is showing them that setting float factors is a possibility (and users on issue #4316 already made use of that possibility).

spyder/plugins/configdialog.py
ccordoba12: I don't know how to do that in a tooltip ;-)

spyder/config/main.py
ccordoba12: Also, is the colon really necessary? I'd prefer to remove it.

ccordoba12:
> I forget about this, I'll try to do it, but I don't know how to test it

Just set a custom scale factor and verify that the text in the Help plugin also scales. If not, you'll have to search for a solution on the web.

spyder/config/main.py
rlaverde: I changed it for "2.5", the ";" was necessary before backporting the changes

spyder/plugins/configdialog.py
rlaverde: I put the link in a label

spyder/config/main.py
ccordoba12: Ok, great!! Then `2.5` (without colon) it is :-)

spyder/plugins/configdialog.py
ccordoba12: Nice, thanks!!

ccordoba12:
@rlaverde, please ping me if/when you solve the problem with Help.

rlaverde:
@ccordoba12 I'm testing and that seems to be an issue with `QWebEngineView` (also happens with spyder-reports), the QT [wiki](https://doc.qt.io/qt-5/qtwebengine-webenginewidgets-minimal-example.html) advice to set `Qt.AA_EnableHighDpiScaling` to `True`, but doesn't mention `QT_SCREEN_SCALE_FACTORS`

I test with a minimal QWebEngineView and It doesn't scale

ccordoba12:
What if we try to scale things using CSS or something like that?

rlaverde:
> What if we try to scale things using CSS or something like that?

Maybe that will work, generate a CSS based in `high_dpi_custom_scale_factors`

ccordoba12:
Yes, exactly!

rlaverde:
@ccordoba12 I think I found the error with Help plugin, font was being assigned sending a `point size` to method `setFontSize()` that receive a `pixel size`.

ccordoba12:
Nice!! Could you post a new screenshot with a custom scale factor?

Also, changing screen resolution requires a restart, so please be sure to demand a restart for the newly introduced options here.

rlaverde:
>Nice!! Could you post a new screenshot with a custom scale factor?

spectacle cs6068

goanpeca:
I will be test this in the afternoon, going out now.

ccordoba12:
Ok, things look really good so far. The only thing missing is requiring a restart, and that would be it :-)

rlaverde:
>The only thing missing is requiring a restart, and that would be it :-)

Done, although I don't find a easy way to manage the preferences as a group, and set the restart for each preference.

ccordoba12:
I'll leave @goanpeca the final word for this one.

PR #4456, opened 2017-05-10
PR: Add more regular code annotation types

### LINKED ISSUE #4452: Add more annotation keywords to the Editor (e.g. BUG, OPTIMIZE, etc)
Code annotation analysis makes a blue check sign in the editor, which is a useful feature to find TODOs or FIXMEs.

But could it add more words? On the net, and more importantly in [pep8 index](https://www.python.org/dev/peps/pep-0350/#mnemonics), we can find many other tags, eg BUG, HACK, OPTIMIZE, etc.

cvanwynsberghe:
Fixes #4452 

----

The PR corresponds to the #4452 issue. It adds code tags HACK, BUG, OPTIMIZE, !!!, ??? to the common TODO and FIXME.

To keep the most common ones, I did a compromise according to the pep8-350 and many forums/websites relating to this topic.

hopefully I PR-ed pushed pulled all correctly :)

### DIFF
--- a/spyder/utils/codeanalysis.py
+++ b/spyder/utils/codeanalysis.py
@@ -24,7 +24,8 @@ DEBUG_EDITOR = DEBUG >= 3
 #==============================================================================
 # Pyflakes/pep8 code analysis
 #==============================================================================
-TASKS_PATTERN = r"(^|#)[ ]*(TODO|FIXME|XXX|HINT|TIP|@todo)([^#]*)"
+TASKS_PATTERN = r"(^|#)[ ]*(TODO|FIXME|XXX|HINT|TIP|@todo|" \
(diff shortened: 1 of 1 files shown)


ccordoba12:
Cool!! Things look right, so this is going in.

Thanks a lot for your contribution. It'll be available in Spyder 3.2.

afvincent:
@cvanwynsberghe I wonder if this [line](https://github.com/spyder-ide/spyder/blob/3.x/spyder/plugins/editor.py#L247) shouldn't be updated to reflect the new annotations introduced by your PR? (I stumble upon it when looking at a translation file ^^)

cvanwynsberghe:
@afvincent Thank you for your remark, there is a new PR #4465 to fix this.

PR #4466, opened 2017-05-13
Fixup rope dependency

ghisvail:
Starting from version 0.10.5, rope is compatible with Python 3.

It is still worth keeping the lower versioned dependency for Python 2 to ease potential backports.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -272,7 +272,7 @@ if any(arg == 'bdist_wheel' for arg in sys.argv):
     import setuptools     # analysis:ignore
 
 install_requires = [
-    'rope_py3k' if PY3 else 'rope>=0.9.4',
+    'rope>=0.10.5' if PY3 else 'rope>=0.9.4',
     'jedi>=0.9.0',
     'pyflakes',
     'pygments>=2.0',


ccordoba12:
This needs other additions and we plan to make it for Spyder 3.2, but thanks for your contribution.

ghisvail:
Could you be a bit more explicit? Does it mean `rope` integration still does not work with `0.10.5`?

ccordoba12:
I meant we need to update our docs and Readme to reflect this change, and probably also update our CI integrations.

Besides, you did this PR against master, so you'd need to redo it against 3.x.

ghisvail:
> Could you be a bit more explicit?
> I meant we need to update our docs and Readme to reflect this change

By *explicit*, I meant: can you confirm whether or not spyder (with Python 3) works correctly with `rope` version 0.10.5? If so, then I should be able to apply this patch to enable rope support on Python 3 for the next update of the Debian package, right?

> so you'd need to redo it against 3.x.

Done (using the PR edit menu)

ccordoba12:
Yes, things work just fine with 0.10.5, but please let us do this PR because we know all the necessary things that need to be done (so it'll be faster than explaining them to you).

Sorry, I know you're well intended but I have to prioritize my time.

ghisvail:
> Yes, things work just fine with 0.10.5

That's all I needed to know.

> please let us do this PR [...] I know you're well intended but I have to prioritize my time.

Which makes it sound mine is less valuable than yours :(

Anyway you asked me to back off, so I won't insist. Good luck with the rest.

ccordoba12:
> Which makes it sound mine is less valuable than yours

Of course not, it's just that this is a maintainer thingy that's easier to handle by myself.

Issue #4463, opened 2017-05-12
Code folding does not work on indentation level 0


sylvain-bougnoux:
AFAIU code folding works on indentation.
Recently I cannot fold anymore "def" and "class", ie all stuff without indentation (ie level 0).
Then in function, "for"  or in class "def method", it works fine.
AFAIR it used to work a few weeks ago.
By the way thanks a lot for this great addition.

rlaverde:
> AFAIU code folding works on indentation.

Yes, This have to be improve #4432 

> Recently I cannot fold anymore "def" and "class", ie all stuff without indentation (ie level 0).

This is working for me with the latest `master`, could you give me an example to reproduce this bug?

sylvain-bougnoux:
Hum, as I had launch issues I did several spyder reset. It down on me, that indentation default (4 spaces) was previously set to my preference (2 spaces & tab width=2 spaces).
Therefore I change them back, within advanced option of preference/editor, and restart IDE. *Now code folding works*. If I put back to 4 it does not. Hence you should be able to reproduce the bug tweaking these options.
 
I don’t if it is truly the reason, but definitely it should work on any code, not only the one compatible with your preference (obviously J).
 
Once again thanks a lot.

rlaverde:
I see, the issue is that fold levels were calculated with the previous indentation, and the blocks edited after the change will have a fold level calculated with a different indentation value

I think we could solve this is making indentation levels independent of config preference (indent_chars), using a "heuristic" (the first indentation in the file) to calculate fold levels instead.

@ccordoba12 @goanpeca what do you think?

ccordoba12:
@rlaverde, so the problem lies on using different indentation levels on the same file?

rlaverde:
> so the problem lies on using different indentation levels in the same file?

When start editing a file, then change the configuration to a different indent_chars, and continue editing, the block information about fold level is updated when editing the block(line)

Also different indentation_chars in a file will broke the folding (mixing tabs and spaces, or some lines 4-spaces and others 3-spaces), the FoldDetector expects all indents to be multiply of indent_chars

ccordoba12:
Ok, then your proposed solution sounds good. Please start implementing it.

sylvain-bougnoux:
> I think we could solve this is making indentation levels independent of config preference (indent_chars), using a "heuristic" (the first indentation in the file) to calculate fold levels instead. 
[...]

Issue #4432, opened 2017-05-05
Implement fold detectors for other languages


rlaverde:
At the moment there is only a generic fold detector based in indentation.

One of the things missing after #3833

goanpeca:
I think we should close this, as now this functionality is deferred to language servers  @ccordoba12

ccordoba12:
Agreed.

PR #4503, opened 2017-05-22
PR: Fold improvements

rlaverde:
Fixes: #4463

Now `IndentFoldDetector` will take in account previous fold indentation before using `CodeEditor.indent_chars`

I tried the solution that I propose in https://github.com/spyder-ide/spyder/issues/4463#issuecomment-301543920 but didn't work very well, the solution that I implemented is partially based in https://github.com/spyder-ide/spyder/issues/4463#issuecomment-301705657

### DIFF
--- a/spyder/widgets/sourcecode/api/folding.py
+++ b/spyder/widgets/sourcecode/api/folding.py
@@ -47,7 +47,8 @@ def print_tree(editor, file=sys.stdout, print_blocks=False, return_list=False):
                       (block.blockNumber() + 1, lvl, visible), file=file)
         block = block.next()
 
-    if return_list: return output_list
+    if return_list:
+        return output_list
 
 class FoldDetector(object):
     """
--- a/spyder/widgets/sourcecode/api/folding.py
+++ b/spyder/widgets/sourcecode/api/folding.py
@@ -14,7 +14,7 @@ import sys
 from spyder.utils.editor import TextBlockHelper
 
 
-def print_tree(editor, file=sys.stdout, print_blocks=False):
+def print_tree(editor, file=sys.stdout, print_blocks=False, return_list=False):
     """
     Prints the editor fold tree to stdout, for debugging purpose.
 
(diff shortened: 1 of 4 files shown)


ccordoba12:
Is there something else you plan to add here? I ask because you marked this as WIP :-)

rlaverde:
Is there something else you plan to add here? I ask because you marked this as WIP :-)

I was planning to also address #4432, but after thinking a little about this, I think is better to address it after restructuring how languages are managed into the Editor (so I'll leave it for another PR)

spyder/widgets/sourcecode/api/folding.py
```diff
@@ -31,14 +33,21 @@ def print_tree(editor, file=sys.stdout, print_blocks=False):
...
-            print('l%d:%s%s' %
-                  (block.blockNumber() + 1, lvl, visible), file=file)
+            if return_list:
+                output_list.append([block.blockNumber() + 1, lvl, visible])
+            else:
+                print('l%d:%s%s' %
+                      (block.blockNumber() + 1, lvl, visible), file=file)
         block = block.next()
 
+    if return_list: return output_list
```
ccordoba12: Please move this `return output_list` to a new line.

spyder/widgets/sourcecode/tests/test_folding.py
```diff
@@ -0,0 +1,60 @@
...
+from spyder.widgets.sourcecode.api.folding import print_tree
+
+
+# --- Fixtures
+# -----------------------------------------------------------------------------
+@pytest.fixture()
+def get_fold_levels():
+    """setup editor and return fold levels."""
+
+    def get_fold_levels():
```
ccordoba12: Why this nested function with the same name as the outer? Can't we get rid of it?

ccordoba12:
This looks good to me. I left a couple of minor comments, then I'll merge.

spyder/widgets/sourcecode/tests/test_folding.py
rlaverde: The fixture will return a function, It will allow the fixture to receive parameters, maybe the text for the editor, or another FoldDetector that will be useful for other tests.

I found this trick in stackoverflow, although reading a little more is better to use [indirect parametrization](https://docs.pytest.org/en/latest/example/parametrize.html#apply-indirect-on-particular-arguments) I'll change it

PR #4535, opened 2017-06-01
PR: Change Editor title to display file path again

dalthviz:
Fixes #4259

PR #4546, opened 2017-06-02
PR:Add missing sig prefix to signal call.

rlaverde:
After merging #4535 with master, this line cause an error

```
  File "/home/rlaverde/code/spyder/spyder/plugins/editor.py", line 590, in visibility_changed
AttributeError: 'function' object has no attribute 'emit'
```


ccordoba12:
This is already fixed.

rlaverde:
Yes, I didn't see #4540 😞

Issue #4524, opened 2017-05-31
Remove the Python console


ccordoba12:
This issue describes our reasons to remove the Python console from Spyder:

1. **Broken on Windows**: Executing code on it usually doesn't return the prompt. The situation is worst when Matplotlib integration is turned on because you have to press `Enter` several times to get a prompt back.
2. **Matplotlib integration**: Nowadays Matplotlib is tightly integrated with IPython and it's very difficult to have a different console with good Matplotlib support. Right now the only backend we are able to support in our Python consoles is the Qt one (not even Tk works!), and (as said above) it's broken on Windows.
3. **Different implementations**: We are a small project and don't have the resources nor the time to maintain two consoles with very different implementations (we're also in charge of `qtconsole`).
4. **Debugging**: It's not possible to detect when a Python console is running a debugging session, so we can't adapt certain functionality for debugging only (e.g. disabling buttons on the Debug toolbar).
  Besides, we plan to improve debugging in Spyder 4 by creating an IPbd kernel, so that users can run multiline code from the Editor, have code completion and history browsing while debugging. Unfortunately, this is something we can't support in the Python console.
5. **History log**: The history log is semi-broken because we can't make it work reliably for both consoles.
6. **Not testable**: It's not easy to add tests for the Python console because you can't tell for sure when the console is up and ready nor when it has finished a computation. That's why there are no tests for it right now.

----

TODO:

- [x] Remove connection with the Variable Explorer - PR #4525.
- [x] Remove connection with Help - PR #4533.
- [x] Remove connection with the Editor - PR #4541.
- [x] Don't check for the presence of the IPython console (i.e. make it mandatory instead of optional) - PR #4541.
- [x] Remove connections with the rest of plugins - PR #4549.
- [x] Add *Environment Variables* and *sys.path contents* entries to the IPython console - PR #4562.
- [x] Remove text about the Python console in the Run options dialog - PR #4691.
- [x] Add *Execute in a dedicated console* option to the IPython console - PR #4691.
- [x] Remove all code related to the Python console - PR #4713.
- [x] Clean `sitecustomize.py` - PR #4713.

koradir:
Will "Add Execute in a dedicated console option to the IPython console." once implemented still resolve issues like 
https://stackoverflow.com/questions/38042329/python-kivy-app-running-only-once
?

ccordoba12:
Yes, I think so because the kernel will be restarted every time on a dedicated console.

PR #4549, opened 2017-06-02
PR: Stop loading the Python console

ccordoba12:
This is part of issue #4524. It also removes the last connections between that console and all other plugins.


ccordoba12:
The failure is due to a decrease in coverage. I don't know why is caused, but coverage should increase again once we remove the code of the Python console.

PR #4548, opened 2017-06-02
PR: Move Help plugin to a module

rlaverde:
First PR of the spyder modularization, the idea is to move each plugin, its widgets and utilities to a separate folder and follow the same structure for internal and external plugins

Example:

```
spyder/
    plugins/
        help/
            plugin.py
            confpage.py
            __init__.py
            utils/
                 exampleutils.py
            widgets/
                examplewigdet.py
```

`utils` and `wigdets` could be directories or just one file depending of the complexity

PR #4557, opened 2017-06-03
PR: Split Variable Explorer module

rlaverde:
Move variable explorer to a module

```
variableexplorer/
├── confpage.py
├── __init__.py
├── plugin.py
├── tests
│   ├── __init__.py
│   ├── issue_2514.csv
│   ├── issue_3896.csv
│   ├── test_arrayeditor.py
│   ├── test_collectioneditor.py
│   ├── test_dataframeeditor.py
│   ├── test_importwizard.py
│   ├── test_namespacebrowser.py
│   ├── test_texteditor.py
│   ├── test_utils.py
│   └── test_variableexplorer.py
├── utils.py
└── widgets
    ├── importwizard.py
    ├── __init__.py
    ├── namespacebrowser.py
    └── viewers
        ├── array.py
        ├── collections.py
        ├── dataframe.py
        ├── __init__.py
        ├── object.py
        └── text.py
```

~This have to be rebased after #4548~ done


ccordoba12:
I'd like to move our viewers to `spyder/plugins/variableexplorer/viewers` to have a smaller import path. 

Besides, our viewers are not really plugin widgets, i.e. they are not embedded in the plugin but can work independently of it. So that's another reason to have them out of widgets.

@goanpeca, what do you say?

goanpeca:
Are these widgets core so that they go inside `spyder/widgets/` or they really only make sense for the variable explorer?

@ccordoba12 I dont think having a shorter import path should be a reason to break a nicely structured plugin. Viewers are widgets so they should live inside the widgets module, we just happen to group them inside a submodule called viewers cause is nicer. So I am :-1: on that.

We should be using relative imports anyway

spyder/plugins/variableexplorer/confpage.py
```diff
@@ -0,0 +1,50 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+
+"""Help Plugin"""
```
goanpeca: help?

rlaverde: 🙈

spyder/config/base.py
```diff
@@ -375,7 +375,7 @@ def get_supported_types():
     dict(picklable=picklable_types, editable=editables_types)
          
     See:
-    get_remote_data function in spyder/widgets/variableexplorer/utils/monitor.py
+    get_remote_data function in spyder/plugins/variableexplorer/widgets/utils/monitor.py
```
ccordoba12: This needs to be `spyder/plugins/variableexplorer/utils.py`

ccordoba12:
Ok, no problem. Then @rlaverde, please make all import statements here to fit in 80 columns, e.g.

from spyder.plugins.variableexplorer.widgets.viewers.collections import (
    CollectionsEditor)

etc.

I really don't like to have longer lines :-)

rlaverde:
>Ok, no problem. Then @rlaverde, please make all import statements here to fit in 80 columns

It's that really necessary? unnecessary wrapped lines doesn't look good, sometimes is better a line longer than 80 characters. (yapf philosophy, i.e. if I yapfy this files the long imports didn't change.)

We could also consider changing the folder name "variableexplorer" is 16 characters long, "namespace" or "workspace" will be only 9.

goanpeca:
>We could also consider changing the folder name "variableexplorer" is 16 characters long, "namespace" or "workspace" will be only 9.

I would support that :+1:

>It's that really necessary? unnecessary wrapped lines doesn't look good, sometimes is better a line longer than 80 characters. 

I think modern software development no longer requires this outdated rule.  I think using 100-ish is quite normal these days, so keeping on 79 for the sake of an old standard is not relevant nowadays.

ccordoba12:
> We could also consider changing the folder name "variableexplorer" is 16 characters long, "namespace" or "workspace" will be only 9.

Not for this PR :-) If/when we rename this plugin's name, then we'll change its namespace name too.

> I think modern software development no longer requires this outdated rule. I think using 100-ish is quite normal these days, so keeping on 79 for the sake of an old standard is not relevant nowadays.

All our code adheres to the 79 limit and this is just a refactoring PR, so this is not the place to start changing it.

So @rlaverde, please fix those imports and also all lines that are wider than 79 columns (except for test files).

goanpeca:
@rlaverde is this ready?

rlaverde:
@rlaverde is this ready?

It is missing the fixes in lines wider than 79, I think that I should manually change them (no magic regex and sed 😢 )

ccordoba12:
> I think that I should manually change them

Sorry for that, but I think there are no more than 10 of those ;-)

goanpeca:
@rlaverde please fix that so we can keep moving with the split !

ccordoba12:
To solve the errors in our tests, please apply here the changes I did in PR #4613.

rlaverde:
> To solve the errors in our tests, please apply here the changes I did in PR #4613.

Done, I cherry-picked the commits

ccordoba12:
Tests for PyQt4 are faliing, I don't know why :-/

rlaverde:
I think it's related to the error solved in 70c2b95390648deef831ccb27a9c4eabaf4e87e7 (cherry-picking 🙈 )

ccordoba12:
You were right, now these tests are passing ;-) Merging, since this is ready.

I just have an unrelated question: what will happen with those cherry-picked commits when we merge with master?

rlaverde:
> I just have an unrelated question: what will happen with those cherry-picked commits when we merge with master?

I bet (and hope) git is smart enough to notice that both commits introduced the same changes and won't be merge conflicts.

goanpeca:
@rlaverde bet or hope :-p you can't have both ;-)

PR #4561, opened 2017-06-04
PR: Use Python module to launch pylint

avylove:
I ran into a problem getting the right version of pylint to launch within the plugin since it's not configurable and expects 'pylint' to used for Python 2 and 'pylint3' or 'python3-pylint' to be used for Python 3.

In my configuration (Fedora 25) the system-installed pylint is 'pylint' and 'python3-pylint', however in order to use a newer version, I've installed pylint using pip3 and the --user option to install it under ~/.local. This creates a 'pylint' executable for Python3 in ~/.local/bin.

In order to use the newer version within Spyder, I have to modify my system PATH as well as rename or symlink ~/.local/bin/pylint to 'pylint3' or 'python3-pylint'.

My proposal is to avoid all this by using the pylint module which is already a requirement for Spyder. The module can be imported to determine the version and can be executed using '-m' for both cpython and pypy.

Here's my attempt to do this. It works well, but I'm not sure of there is a better way to do parts of it.

### DIFF
--- a/spyder_pylint/pylint.py
+++ b/spyder_pylint/pylint.py
@@ -23,8 +23,9 @@ from spyder.config.base import get_translation
 from spyder.plugins import SpyderPluginMixin
 from spyder.plugins.configdialog import PluginConfigPage
 from spyder.utils import icon_manager as ima
+from spyder.utils.programs import is_module_installed
 from spyder.utils.qthelpers import create_action, MENU_SEPARATOR
-from .widgets.pylintgui import (PYLINT_PATH, PylintWidget)
+from .widgets.pylintgui import PylintWidget
 
 
 _ = get_translation("pylint", "spyder_pylint")
@@ -132,7 +133,7 @@ class Pylint(PylintWidget, SpyderPluginMixin):
         
         pylint_act = create_action(self, _("Run static code analysis"),
                                    triggered=self.run_pylint)
-        pylint_act.setEnabled(PYLINT_PATH is not None)
+        pylint_act.setEnabled(is_module_installed('pylint'))
         self.register_shortcut(pylint_act, context="Pylint",
                                name="Run analysis")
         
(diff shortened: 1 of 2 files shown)


jitseniesen:
Thanks for your contribution. I don't know this bit of code, but in my experience with the unittest plugin calling modules with `python -m` is indeed more robust than using the executables.

ccordoba12:
@avylove, thanks a lot for your contribution! We'd like to include it in our upcoming release, so I'd like to ask you to rebase your branch against our `3.x` branch.

As it's clearly stated in our [contributing guidelines](https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md), changes to master are meant for the next major Spyder release, which could take 8 to 10 months to go out.

ccordoba12:
@rlaverde, what's the git command to do this?

rlaverde:
>@rlaverde, what's the git command to do this?

`git rebase --onto <new_base> <old_base> <branch>`

`git rebase --onto 3.x master pylint_from_module`

avylove:
Rebased to 3.x. Not sure why that test is failing. Ideas?

ccordoba12:
@avylove, that failure was unrelated to your work. After a restart, all our test are passing.

@rlaverde, please open a PR and add those git commands to our Contributing guide. Thanks!

avylove:
Found a bug in the original code I'd like to put a PR in for, but it touches the same lines. Should I wait until this gets commited to 3.x?

Basically need to add '--output-format=text' to the pylint options in order to account for things like output-format=colorized in pylintrc

ccordoba12:
Please do it here. Fortunately this PR hasn't been merged, so there's no problem with that.

ccordoba12:
@avylove, I did a quick review and everything looks good to me.

I think we should have done this after requiring `pylint` as a required dependency. Thanks a lot for helping us with this!

Issue #2987, opened 2016-02-15
Add feature that shows vertical lines to help identifying indentation structure


webcsm:

ccordoba12:
There is already an option under the menu entry

    Source > Show blank spaces

that helps with indentation.

Unfortunately, adding vertical lines is not an easy task, so we're not planning to do it right now.

webcsm:
I totally understand. Thanks for your quick reply.

webcsm:
I saw that you have a number of issues there. Do you accept help with PRs? I might help with something.

ccordoba12:
Sure, help is always welcome!

Issues in `wishlist` are things we think are interesting but that we don't plan to implement ourselves.

chandu8542:
@ccordoba12 . IS there any update on adding vertical line feature to demarcate the code indentation? After  3-4 level of indentation, we are lost in adjusting indentation.

ccordoba12:
We'll take a look at this for Spyder 4. Sorry, but it's not possible to add it to Spyder 3 because it requires several internal changes.

zizzipupp:
Slightly off-topic, sorry, but when is Spyder 4 scheduled for release?

ccordoba12:
During the last quarter of this year.

Wurtzinator:
@ccordoba12 

Will the indentation line be implemented anytime soon or are there other extensions?
I love Spdyer, but I really need  the indentation lines as a newbie :D

ccordoba12:
It's already in beta1.

LovLearn:
Hello @ccordoba12 ,
Could you please give me the direction on how to have find beta1 and have it set up? I am new to spyder. Thanks.

Dodelo:
> Hello @ccordoba12 ,
> Could you please give me the direction on how to have find beta1 and have it set up? I am new to spyder. Thanks.

In case you're still interested: look over here https://github.com/spyder-ide/spyder/releases/tag/v4.0.0b1

abhijit1990:
still not added

vivalaelectro:
> still not added

I see the feature in Spyder IDE 5.4.3. The option
```
Source > Show indent guides
```
must be activated to show the vertical lines.

ccordoba12:
That's right @vivalaelectro. We haven't enabled this feature by default because it still has some issues that we hope to solve in Spyder 6.

PR #4572, opened 2017-06-07
PR: Add indentation guidelines

rlaverde:
Fixes: #2987 

I added a floating type panel, and move `edgeline` and `indentguides` to use it.

I didn't find a way to abstract all the logic in the `PanelManager`, and floatings Panels need to have a `set_geometry` method

This is how it looks like, It uses the comments color (same as edgeline)
spectacle j13255

### DIFF
--- a/spyder/widgets/panels/indentationguides.py
+++ b/spyder/widgets/panels/indentationguides.py
@@ -27,7 +27,6 @@ class IndentationGuide(Panel):
         self.color = Qt.darkGray
         self.i_width = 4
 
-        self.setAttribute(Qt.WA_TransparentForMouseEvents)
         self._enabled = True
 
     def paintEvent(self, event):
--- a/spyder/widgets/panels/indentationguides.py
+++ b/spyder/widgets/panels/indentationguides.py
@@ -28,8 +28,6 @@ class IndentationGuide(Panel):
         self.color = Qt.darkGray
         self.i_width = 4
 
-        self._enabled = True
-
     def paintEvent(self, event):
         """Override Qt method."""
         painter = QPainter(self)
--- a/spyder/widgets/panels/indentationguides.py
+++ b/spyder/widgets/panels/indentationguides.py
@@ -1,0 +1,80 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+
+"""
+This module contains the indentation guide panel
+"""
+
+from qtpy.QtWidgets import QWidget
+from qtpy.QtCore import Qt, QRect, QPoint
+from qtpy.QtGui import QPainter, QColor
+
+
+class IndentationGuide(QWidget):
+    """Source code editor's edge line (default: 79 columns, PEP8)"""
+
+    # --- Qt Overrides
+    # -----------------------------------------------------------------
+    def __init__(self, editor, color=Qt.darkGray, indentation_width=4):
+        QWidget.__init__(self, editor)
+        self.editor = editor
+        self.color = color
+        self.i_width = indentation_width
+
+        self.setAttribute(Qt.WA_TransparentForMouseEvents)
+        self._enabled = True
+
+    def paintEvent(self, event):
+        """Override Qt method"""
+        painter = QPainter(self)
+
+        color = QColor(self.color)
+        color.setAlphaF(.5)
+        painter.setPen(color)
+
+        for top, line_number, block in self.editor.visible_blocks:
+            bottom = top + int(self.editor.blockBoundingRect(block).height())
+
+            # replace tabs for spaces
+            text = block.text().replace('\t', ' ' * self.i_width)
+
+            indentation = len(text) -len(text.lstrip())
+            for i in range(self.i_width, indentation, self.i_width):
+                x = self.editor.fontMetrics().width(i * '9')
+                painter.drawLine(x, top, x, bottom)
+
+    # --- Other methods
+    # -----------------------------------------------------------------
+
+    def set_enabled(self, state):
+        """Toggle edge line visibility"""
+        self._enabled = state
+        self.setVisible(state)
+
+    def update_color(self):
+        """
+        Set edgeline color using syntax highlighter color for comments
+        """
+        self.color = self.editor.highlighter.get_color_name('comment')
+
+    def set_indentation_width(self, indentation_width):
+        """Set indentation width to be used to draw indent guides."""
+        self.i_width = indentation_width
+
+    def set_geometry(self, cr):
+        """Calculate and set geometry of edge line panel.
+            start --> fist line position
+            width --> max position - min position
+        """
+        offset = self.editor.contentOffset()
+        x = self.editor.blockBoundingGeometry(self.editor.firstVisibleBlock()) \
+            .translated(offset.x(), offset.y()).left() + 5
+
+        top_left = QPoint(x, cr.top())
+        top_left = self.editor.calculate_real_position(top_left)
+        bottom_right = QPoint(top_left.x() + cr.width(), cr.bottom())
+
+        self.setGeometry(QRect(top_left, bottom_right))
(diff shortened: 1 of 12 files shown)


goanpeca:
Nice :-), like this a lot!

spyder/widgets/panels/indentationguides.py
```diff
@@ -0,0 +1,79 @@
...
+    """Source code editor's edge line (default: 79 columns, PEP8)"""
+
+    # --- Qt Overrides
+    # -----------------------------------------------------------------
+    def __init__(self, editor):
+        Panel.__init__(self, editor)
+        self.color = Qt.darkGray
+        self.i_width = 4
+
+        self.setAttribute(Qt.WA_TransparentForMouseEvents)
```
goanpeca: Why duplicate here?

spyder/widgets/panels/indentationguides.py
```diff
@@ -0,0 +1,79 @@
...
+        self._enabled = True
+
+    def paintEvent(self, event):
+        """Override Qt method"""
```
goanpeca: Period on docstrings

spyder/widgets/panels/indentationguides.py
```diff
@@ -0,0 +1,79 @@
...
+            indentation = len(text) -len(text.lstrip())
+            for i in range(self.i_width, indentation, self.i_width):
+                x = self.editor.fontMetrics().width(i * '9')
+                painter.drawLine(x, top, x, bottom)
+
+    # --- Other methods
+    # -----------------------------------------------------------------
+
+    def set_enabled(self, state):
+        """Toggle edge line visibility"""
```
goanpeca: same

spyder/widgets/panels/indentationguides.py
```diff
@@ -0,0 +1,79 @@
...
+        self._enabled = state
+        self.setVisible(state)
+
+    def update_color(self):
+        """
+        Set edgeline color using syntax highlighter color for comments
```
goanpeca: same

spyder/widgets/panels/indentationguides.py
rlaverde: I forget to deleted. when moving it to panel class

spyder/widgets/panels/indentationguides.py
```diff
@@ -0,0 +1,79 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+
+"""
+This module contains the indentation guide panel
```
goanpeca: period

ccordoba12:
Screenshot, please? :-)

rlaverde:
>Screenshot, please? :-)

Added 😄

ccordoba12:
Neat, but the guidelines are discontinuous? Can we do something about it?

goanpeca:
>Neat, but the guidelines are discontinuous?

@ccordoba12, please post a screenshot of what you expect? on vscode, or sublime... or ?

goanpeca:
@rlaverde screenshot?

rlaverde:
> Neat, but the guidelines are discontinuous? Can we do something about it?

It's drawn using only the current line, I added a little hack to avoid discontinuity (preserve previous indentation), I think this is the best possible without making it "smart" and adding a lot of complexity.


spectacle y14297

goanpeca:
I think it looks ok, we will probably need information from the fold detector eventually to make it better?

rlaverde:
> we will probably need information from the fold detector eventually to make it better?

I hadn't thought about that 🙈 It was way easy to use folding levels and less error prone. I looks the same as the last screenshot.

spyder/widgets/panels/indentationguides.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def set_indentation_width(self, indentation_width):
+        """Set indentation width to be used to draw indent guides."""
+        self.i_width = indentation_width
+
+    def set_geometry(self, cr):
+        """Calculate and set geometry of indentation guides panel."""
+        offset = self.editor.contentOffset()
+        x = self.editor.blockBoundingGeometry(self.editor.firstVisibleBlock()) \
+            .translated(offset.x(), offset.y()).left() + 5
```
goanpeca: What is this `5`?

rlaverde: A hardcoded offset used in the edgeline panel (I don't have any idea why, but I delete it and the panel is misplaced) 😕

goanpeca: hmmm ok, we need to dig where this 5 comes from :-p

rlaverde: https://github.com/spyder-ide/spyder/commit/f7c68f33574f2099f6cdab93762befda4d20bf7c#diff-a032958c9839f940f8038a099a3bf970R1072

ccordoba12: So Pïerre is the only one who knows why that 5 is here :-/

rlaverde: I think that is something like the margin of the editor (but I'm not sure)

goanpeca: Ok whatever it is, we need to put in a constant or use it within the methods that return the offsets, so that it is fixed in one place!

ccordoba12: Right, but that's not work for this PR, I'd say ;-)

spyder/widgets/panels/linenumber.py
```diff
@@ -178,14 +178,6 @@ def get_markers_margin(self):
...
             return 0
 
-    def update_(self, qrect, dy):
-        """Update line number area"""
-        if dy:
-            self.scroll(0, dy)
-        else:
-            self.update(0, qrect.y(),
-                        self.width(),
-                        qrect.height())
```
ccordoba12: Why is this not needed anymore?

ccordoba12:
This looks terrific!! Great work @rlaverde!

A couple of comments:

1. Please add a menu entry in the *Source* menu to easily activate/deactivate these guidelines. This is similar to the *Show blank spaces* entry.
2. A test for this would be desirable (even if very simple).

goanpeca:
>A test for this would be desirable (even if very simple).

This is a UI exclusive thing, it would be kinda hard to test I think :-|

spyder/widgets/panels/indentationguides.py
goanpeca: Yes, @rlaverde lets open a techdebt issue to track this problem so we can solve it later on

goanpeca:
>Please add a menu entry in the Source menu to easily activate/deactivate these guidelines. This is similar to the Show blank spaces entry.

@ccordoba12 I would say spaces and guidelines fall in the same thing, so maybe we can just merge these options together? I say that in the spirit of keeping the UI and options simple. Thoughts?

ccordoba12:
> This is a UI exclusive thing, it would be kinda hard to test I think :-|

A test could consist of creating a Python file with indentation and test that guidelines are visible. That would be enough for me.

> I would say spaces and guidelines fall in the same thing, so maybe we can just merge these options together? I say that in the spirit of keeping the UI and options simple. Thoughts?

I would like to use guidelines only, so I'd prefer to maintain the two options separate.

spyder/widgets/panels/linenumber.py
rlaverde: It's implemented in the `PanelManager` class, It's enough to set `scrollable=True` in a `Panel` to make it scroll with the editor

goanpeca:
>A test could consist of creating a Python file with indentation and test that guidelines are visible. 

Guidelines are painted, not a widget, how do you check that?

rlaverde:
I added a simple test that check that panels are added to the editor and are enabled, I don't know how to test the internal behavior of the Paint method.

goanpeca:
>I don't know how to test the internal behavior of the Paint method.

Neither do I.... aside from doing more complex image processing...

ccordoba12:
> I don't know how to test the internal behavior of the Paint method.

Fair enough. The test you added is sufficient.

So the only thing missing here is adding an option to `Source`. Then I'll merge :-)

ccordoba12:
Is this ready?

rlaverde:
> Is this ready?

Yes, the refactoring related to the hardcoded offset will be addressed in other PR

ccordoba12:
Great, thanks!

maxwell8888:
@ccordoba12 What version of Spyder is this available in? I am running 3.2.6 and I can't find the option in the source menu. Thanks!

PR #4056, opened 2017-01-27
PR: Several Find in Files improvements

andfoy:
Fixes #4197
Fixes #4134
Fixes #2963
Fixes #2850
Fixes #2730
Fixes #1555   

----

After checking issue #4005, I've identified that the widget in it's current incarnation doesn't allow to track the status of the search, that implies that the User doesn't know if the search is taking place or not, especially when the search is launched in a big directory. Taking in account this requirement, a progress bar and a text field that indicates the current file scanned is added. Also, some efficiency issues and codification mismatches are being tackled in the present PR.

- [x] Add spinner and indicator of search progress 
- [x] Inform user if the search/exclude/include regexp patterns are incorrect
- [x] Improve search and scanning of files (Do not recover all files before searching a term)
- [x] Improve file and string encoding to bypass and correct errors associated to codification
- [x] Fix Spyder crash after a long line is matched and tried to display on the results window
- [x] Simplify visualization of results (The current tree directory visualization is useless if a file resides inside a deep folder)
- [x] Break results if the pattern was found in a single line 
- [x] Display results as search takes place
- [x] Replace default regex suggestions
- [x] Remove previous search results when a new search takes place
- [x] Remove unused search options
- [x] Search path now refers to global explorer path
- [x] Omit binary files
- [x] Add options to search on the current file, project or current path
- [x] Add test
- [x] Allow to order results alphabetically from top header

Now the Widget should look like this while a search task is taking place
peek 25-02-2017 16-25

PR #4599, opened 2017-06-13
PR: More find in files improvements

### LINKED ISSUE #4596: Even more improvements to Find in Files
I want to see two final improvements to Find in Files:

1. Show the full line in the seach results if a line is less than 80 chars, and four words to left and right if the line is longer.
2. Expand the results found in a file as soon as the file is finished to be scanned, and not to wait until we finish to scan the whole path or project. This will imply to not sort files by name, I think, just results by line number.

andfoy:
Fixes #4596 

- [x] File line matches are automatically expanded
- [x] Lines are cropped only if they have more than 80 characters


ccordoba12:
Great! Thanks a lot for working on this. I'll look at it in couple of hours.

ccordoba12:
@andfoy, unfortunately this is not working for minified Javascript files:

seleccion_001

Perhaps we could also take 10 chars to left and right, and if their length is smaller than taking 4 words to left and right, then use that instead. What do you think?

andfoy:
@ccordoba12 It sounds great! I'll implement it and test if it works

ccordoba12:
Let's make it 20 chars to left and right. 10 seems too little.

ccordoba12:
@andfoy, this is looking great! I fixed a minor bug I found, and still have some more comments:

1. Please truncate results with 30 chars to left and right, instead of 20. 20 is still too little.
2. Is it possible to append new files and their results at the end of the tree widget? The thing is it's quite distracting to be reading results of one file and then see those results *disappear* from view because the results of a new file are shown instead. The other solution would be to not sort files by name. In any case, we need to find a solution for this.
3. There are some terms that are not highlighted on the result line. For example, I looked for the term **append** in the Spyder source code, and, as you can see below, it's not highlighted in some cases. Please take a look at that too.

    seleccion_001

andfoy:
@ccordoba12  Isn't it HTML module builtin on Python? https://docs.python.org/3/library/html.html

andfoy:
@ccordoba12 Should I disable sorting feature?

ccordoba12:
No, the idea would be that after a search has finished, users would be able to sort files, but not while it's running.

andfoy:
I will implement that then!

ccordoba12:
You left a print statement around, that's why test are failing.

spyder/widgets/findinfiles.py
```diff
@@ -786,6 +797,8 @@ def closing_widget(self):
 
     def search_complete(self, completed):
         """Current search thread has finished"""
+        self.result_browser.enable_sorting(True)
+        self.result_browser.enable_sorting(True)
```
ccordoba12: This line is repeated.

spyder/widgets/findinfiles.py
```diff
@@ -576,6 +612,10 @@ def activated(self, item):
             filename, lineno, colno = itemdata
             self.parent().edit_goto.emit(filename, lineno, self.search_text)
 
+    def enable_sorting(self, flag):
+        """Enable result sorting after search is complete."""
+        self.sorting_status['status'] = flag
```
ccordoba12: Why this has to be a dictionary instead of a property? I mean, you could simply have

    self.sorting = True/False

instead of having a method called `enable_sorting` that writes to a dict.

andfoy: But if I modify a property, will it propagate along all elements?

spyder/widgets/findinfiles.py
```diff
@@ -596,7 +637,8 @@ def append_result(self, results, num_matches):
         filename, lineno, colno, line = results
 
         if filename not in self.files:
-            item = FileMatchItem(self, filename)
+            item = FileMatchItem(self, filename, self.sorting_status)
```
andfoy: I disable sorting only on file entries and not line entries, to accomplish that I use the dict

spyder/widgets/findinfiles.py
ccordoba12: Ok, then let's refactor this a bit so it's easier to understand. What about

    self.sorting['status'] = 'on'/'off'

?

ccordoba12:
I tested your latest changes and there are some more bugs:

1. Results are sorted automatically after a search finishes. That needs to be avoided, i.e. sorting needs to be applied *only* if the user asks for it.
2. Results appear twice sometimes

    seleccion_001

3. The interface is quite unresponsive while a search is taking place, e.g. scrolling in the Editor is very slow. Can we do something about it?

ccordoba12:
Forget about point 2. I had the same files in two different directories because I was creating wheels for Spyder before to fix another bug ;-)

andfoy:
It seems that disabling sorting features while a search is taking place helps to unblock the result browser. However, most of the overhead is caused by adding multiple results simultaneously

ccordoba12:
> most of the overhead is caused by adding multiple results simultaneously

Ok, what if we show new results after a file has been scanned completely, instead of showing each new result at a time?

andfoy:
So, should we fallback to our previous search behaviour and compare it with the actual?

ccordoba12:
> So, should we fallback to our previous search behaviour and compare it with the actual?

What previous one exactly? The one in 3.1 or before you did this PR?

andfoy:
The one present on #4056

andfoy:
I think performance dropped by using whole file matches. Should I try to disable automatic expansion of results?

ccordoba12:
> I think performance dropped by using whole file matches

I don't understand what you mean

> Should I try to disable automatic expansion of results?

I think rendering results is not the problem. I disabled all rendering to try to find where performance is being affected, and the Editor was still very unresponsive.

So I don't think that's the cause, so it should be in SearchThread instead.

andfoy:
Maybe line cropping should be done on the results viewer instead of being done on the Thread?

ccordoba12:
Probably, I was also thinking that.

andfoy:
I'll give it a try

andfoy:
It seems that truncating lines outside the Thread improves search performance

andfoy:
@ccordoba12 Does the last commit addresses the two issues?

ccordoba12:
Absolutely!! There's still a bit of lag but the situation improved considerably! Please do the required cleanup (you left a lot of comments in your last commit), then I'll merge.

I'm really happy with the results! Thanks a lot for putting so much love in our *Find in Files*.

spyder/widgets/findinfiles.py
```diff
@@ -126,50 +130,10 @@ def find_files_in_path(self, path):
...
-        right = ' '.join(right)
-
-        trunc_line = '{0}<b>{1}</b>{2}'.format(left, match, right)
-        return trunc_line
-
     def find_string_in_file(self, fname):
         self.error_flag = False
         self.sig_current_file.emit(fname)
-        # results = {}
+        # results = []
```
ccordoba12: Please also remove this comment.

Issue #3332, opened 2016-07-28
Spyder can't start because of problems with OpenGL and PyQt5


abdelouahabb:
I get this code:

```
C:\Users\Abdelouahab\Desktop\spyder-master>bootstrap.py > c:/aa.txt
Failed to load libEGL (Le module spÚcifiÚ est introuvable.)
QWindowsEGLStaticContext::create: Failed to load and resolve libEGL functions
Failed to load opengl32sw.dll (Le module spÚcifiÚ est introuvable.)
Failed to load and resolve WGL/OpenGL functions
Failed to load libEGL (Le module spÚcifiÚ est introuvable.)
QWindowsEGLStaticContext::create: Failed to load and resolve libEGL functions
Failed to load opengl32sw.dll (Le module spÚcifiÚ est introuvable.)
Failed to load and resolve WGL/OpenGL functions
QWindowsNativeInterface::nativeResourceForContext: 'renderingcontext' requested
for null context or context without handle.
Failed to load libEGL (Le module spÚcifiÚ est introuvable.)
QWindowsEGLStaticContext::create: Failed to load and resolve libEGL functions
Failed to load opengl32sw.dll (Le module spÚcifiÚ est introuvable.)
Failed to load and resolve WGL/OpenGL functions
Failed to load libEGL (Le module spÚcifiÚ est introuvable.)
QWindowsEGLStaticContext::create: Failed to load and resolve libEGL functions
Failed to load opengl32sw.dll (Le module spÚcifiÚ est introuvable.)
Failed to load and resolve WGL/OpenGL functions
QOpenGLWidget: Failed to create context
Failed to load libEGL (Le module spÚcifiÚ est introuvable.)
QWindowsEGLStaticContext::create: Failed to load and resolve libEGL functions
Failed to load opengl32sw.dll (Le module spÚcifiÚ est introuvable.)
Failed to load and resolve WGL/OpenGL functions
QOpenGLWidget: Failed to create context
Failed to load libEGL (Le module spÚcifiÚ est introuvable.)
QWindowsEGLStaticContext::create: Failed to load and resolve libEGL functions
Failed to load opengl32sw.dll (Le module spÚcifiÚ est introuvable.)
Failed to load and resolve WGL/OpenGL functions
QOpenGLWidget: Failed to create context
Failed to load libEGL (Le module spÚcifiÚ est introuvable.)
QWindowsEGLStaticContext::create: Failed to load and resolve libEGL functions
[...]

PR #4601, opened 2017-06-13
PR: Add pyopengl to setup.py to fix errors with some Nvidia/Intel drivers

rlaverde:
Fixes: #3332

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -288,6 +288,7 @@ install_requires = [
     'pyzmq',
     'chardet>=2.0.0',
     'numpydoc',
+    'pyopengl',  # This is needed only in pip installations see issue #3332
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -288,6 +288,7 @@ def run(self):
     'pyzmq',
     'chardet>=2.0.0',
     'numpydoc',
+    'pyopengl',  # This is needed only in pip installations see issue #3332
```
goanpeca: Is Linux only ?

rlaverde: No, according to the issue thread this also happen with Windows.

goanpeca: Ok :+1:

PR #4612, opened 2017-06-15
PR: Add same behaviour for go to previous cell as the one in Matlab

### LINKED ISSUE #4611: Goto previous cell behaviour is different than other IDEs
Goto previous cell should go to previous cell only when the cursor is at the first line of the cell (the cell marker), otherwise it should go to the beginning of the current cell (same as in MATLAB, Atom etc.)

While indeed "goto previous cell" is the definition of the action, its actual action should be dependant on the cursor location.

1. Create 2 cells with several lines each
2. Place the cursor at the second line (or any other line except the cell marker) of the second cell
3. Select Goto previous cell 

The cursor should be placed at the current cell marker rather than the cell marker of the previous cell

* Spyder Version: spyder from GIT obtained on 1/Jun/2017
* Python Version: 3.5.2
* Qt Version:5.6.2
* PyQt Version:5.6
* Operating system:Linux

dalthviz:
Fixes #4611 

Preview:
cell

### DIFF
--- a/spyder/widgets/sourcecode/base.py
+++ b/spyder/widgets/sourcecode/base.py
@@ -799,7 +799,7 @@ class TextEditBaseWidget(QPlainTextEdit, BaseEditMixin):
             cur_pos = prev_pos = cursor.position()
 
         while not self.is_cell_separator(cursor):
-            # Moving to the previous code cell
+            # Move to the previous cell or the beginning of the current cell
             cursor.movePosition(QTextCursor.PreviousBlock)
             prev_pos = cur_pos
             cur_pos = cursor.position()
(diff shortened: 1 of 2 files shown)


ccordoba12:
Nice! @rlaverde, please take a look at this one.

@liavassif, is this what you want?

liavassif:
@ccordoba12 Yes!
@dalthviz Thanks!

spyder/widgets/sourcecode/base.py
```diff
@@ -793,25 +793,19 @@ def go_to_previous_cell(self):
         cursor = self.textCursor()
         cur_pos = prev_pos = cursor.position()
 
-        # Move to the begining of the cell
-        while not self.is_cell_separator(cursor):
-            # Moving to the previous code cell
+        if self.is_cell_separator(cursor):
+            # Move to the previous cell
```
rlaverde: Maybe explain the behavior in the comment, not always will move to the previous cell: "Move to the previous cell or the beginning of the current cell"

dalthviz: @rlaverde I made the change in the comment below this one, it's ok there or should I make the change here?

rlaverde: Yes, It's ok, It's even better where you did the change

PR #4626, opened 2017-06-20
PR: Create a unique instance of the file switcher attached to the main window

### LINKED ISSUE #4575: Make a unique instance of the file switcher
The idea here is to make a generic file switcher attached to the main window that could display the files of the plugins, for example the Editor and the Notebook (and others if needed).

The file switcher should display a list with the possible files grouped by plugin like:

**Editor**
_file1_
_file2_

**Notebook**
_notebook1_
_notebook2_

The plugins should give the file switcher a list with the current files that are managing.

@ccordoba12 
@goanpeca , what do you think?

dalthviz:
Fixes #4575 

A plugin could be attached to the file switcher by using the `add_to_fileswitcher` method of the Main Window. Also, the plugin needs to implement two methods: `get_current_tab_manager` and `set_stack_index` method. The `set_stack_index` method should be placed in the widget that manages the tabs.


spyder/widgets/editor.py
```diff
@@ -766,9 +766,9 @@ def open_fileswitcher_dlg(self):
             self.fileswitcher_dlg.hide()
             self.fileswitcher_dlg.is_visible = False
             return
-        self.fileswitcher_dlg = FileSwitcher(self, self.tabs, self.data)
+        self.fileswitcher_dlg = FileSwitcher(self, self.tabs, self.data, self)
         self.fileswitcher_dlg.sig_goto_file.connect(self.set_stack_index)
-        self.fileswitcher_dlg.sig_close_file.connect(self.close_file)
+        self.fileswitcher_dlg.setup()
         self.fileswitcher_dlg.show()
         self.fileswitcher_dlg.is_visible = True
```
ccordoba12: Do we still need this instance of the file switcher here? I mean, can we remove it given that we have now a single and global instance?

dalthviz: I left this for the windows that the editor could create, but a think it is possible to attach that windows to the file switcher too.

ccordoba12:
Please merge with 3.x to fix errors in our tests.

spyder/widgets/editor.py
ccordoba12: Ok, but how is the file switcher called for those windows?

dalthviz: As it is in this moment for those windows only by clicking in the action of the menu.

ccordoba12: @goanpeca, what do you think? should we leave it or remove it?

goanpeca: Hmmm not a strong opinion either way :-|

ccordoba12: Let's leave it for now then ;-)

spyder/app/mainwindow.py
```diff
@@ -2720,6 +2745,36 @@ def show_tour(self, index):
...
+        self.fileswitcher_dlg.setup()
+        self.fileswitcher_dlg.show()
+        self.fileswitcher_dlg.is_visible = True
+
+    def open_symbolfinder_dlg(self):
+        """Open symbol list management dialog box."""
+        self.open_fileswitcher_dlg()
+        self.fileswitcher_dlg.set_search_text('@')
+
+    def add_to_fileswitcher(self, tabs, data, plugin):
```
ccordoba12: What would be `data` for the Notebook or the IPython console?

ccordoba12: I'd say it's `None`, so in this case we should change the signature of this function to be

    def add_to_fileswitcher(self, plugin, tabs, data=None):

spyder/widgets/fileswitcher.py
```diff
@@ -221,24 +222,26 @@ def focusOutEvent(self, event):
...
-    sig_close_file = Signal(int)
+    sig_goto_file = Signal(int, object)
 
     # Constants that define the mode in which the list widget is working
     # FILE_MODE is for a list of files, SYMBOL_MODE if for a list of symbols
     # in a given file when using the '@' symbol.
     FILE_MODE, SYMBOL_MODE = [1, 2]
 
-    def __init__(self, parent, tabs, data):
+    def __init__(self, parent, tabs, data, plugin):
```
ccordoba12: Let's change this to be

    def __init__(self, parent, plugin, tabs, data):

spyder/app/mainwindow.py
dalthviz: For the Notebook `data` is `self.clients`

ccordoba12: Ok, then let's change the signature anyway to be

    def add_to_fileswitcher(self, plugin, tabs, data)

spyder/app/mainwindow.py
```diff
@@ -2720,6 +2745,36 @@ def show_tour(self, index):
...
+        self.fileswitcher_dlg.is_visible = True
+
+    def open_symbolfinder_dlg(self):
+        """Open symbol list management dialog box."""
+        self.open_fileswitcher_dlg()
+        self.fileswitcher_dlg.set_search_text('@')
+
+    def add_to_fileswitcher(self, plugin, tabs, data):
+        """Add a plugin to the File Switcher."""
+        if self.fileswitcher_dlg == None:
```
ccordoba12: The convention when comparing with `None` is:

    if self.fileswitcher_dlg is None:

Please change it to be this way.

spyder/widgets/fileswitcher.py
```diff
@@ -287,44 +290,46 @@ def __init__(self, parent, tabs, data):
...
     def line_count(self):
-        return [editor.get_line_count() for editor in self.editors]
+        return [widget.get_line_count() for widget in self.widgets]
 
     @property
     def save_status(self):
-        return [getattr(td, 'newly_created', False) for td in self.data]
+        return [getattr(td,
+                        'newly_created',
+                        False) for da in self.plugins_data for td in da]
```
ccordoba12: This is too complex to have it as list comprehension. Please simplify it using two list comprehensions instead.

spyder/widgets/fileswitcher.py
```diff
@@ -287,44 +290,46 @@ def __init__(self, parent, tabs, data):
...
 
     @property
     def paths(self):
-        return [getattr(td, 'filename', None) for td in self.data]
+        return [getattr(td,
+                        'filename',
+                        None) for da in self.plugins_data for td in da]
```
ccordoba12: The same comment applies here: please use two list comprehensions.

spyder/widgets/fileswitcher.py
```diff
@@ -335,13 +340,13 @@ def set_search_text(self, _str):
         self.edit.setText(_str)
 
     def save_initial_state(self):
-        """Saves initial cursors and initial active editor."""
+        """Saves initial cursors and initial active widget."""
```
ccordoba12: Saves -> Save

spyder/app/mainwindow.py
```diff
@@ -322,6 +323,9 @@ def signal_handler(signum, frame=None):
         # Tour  # TODO: Should I consider it a plugin?? or?
         self.tour = None
         self.tours_available = None
+        
+        # File switcher
+        self.fileswitcher_dlg = None
```
ccordoba12: fileswitcher_dlg -> fileswitcher

spyder/app/mainwindow.py
```diff
@@ -2720,6 +2745,36 @@ def show_tour(self, index):
         self.tour.set_tour(index, frames, self)
         self.tour.start_tour()
 
+    # ---- Global File Switcher
+    def open_fileswitcher_dlg(self):
```
ccordoba12: open_fileswitcher_dlg -> open_fileswitcher

spyder/app/mainwindow.py
```diff
@@ -2720,6 +2745,36 @@ def show_tour(self, index):
...
+        if self.fileswitcher_dlg is not None and \
+          self.fileswitcher_dlg.is_visible:
+            self.fileswitcher_dlg.hide()
+            self.fileswitcher_dlg.is_visible = False
+            return
+        self.fileswitcher_dlg.setup()
+        self.fileswitcher_dlg.show()
+        self.fileswitcher_dlg.is_visible = True
+
+    def open_symbolfinder_dlg(self):
```
ccordoba12: open_symbolfinder_dlg -> open_symbolfinder

spyder/widgets/editor.py
```diff
@@ -766,9 +766,9 @@ def open_fileswitcher_dlg(self):
             self.fileswitcher_dlg.hide()
             self.fileswitcher_dlg.is_visible = False
             return
-        self.fileswitcher_dlg = FileSwitcher(self, self.tabs, self.data)
+        self.fileswitcher_dlg = FileSwitcher(self, self, self.tabs, self.data)
```
ccordoba12: fileswitcher_dlg -> fileswitcher

ccordoba12:
@dalthviz, I also think we need to define here a different icon for the files/tabs of each plugin. So when a plugin registers to the file switcher, it needs to tell it which icon is going to use.

For now let's use these icons:

1. `TextFileIcon` for the Editor.
2. `FileIcon` for the Notebook.
3. `ipython_console` for our consoles (when we implement this, so not for this PR).

spyder/widgets/fileswitcher.py
dalthviz: Something like would work?:
```
save_status = []
    for da in self.plugins_data:
        save_status += [getattr(td, 'newly_created', False) for td in da]
return save_status
```

Also, I should do this for the `widgets`, `paths` and `filenames` properties, right?

ccordoba12: Yes, that works too.

> Also, I should do this for the widgets, paths and filenames properties, right?

Yep.

spyder/plugins/editor.py
```diff
@@ -1183,6 +1164,8 @@ def register_plugin(self):
         if not editorstack.data:
             self.__load_temp_file()
         self.main.add_dockwidget(self)
+        self.main.add_to_fileswitcher(self, editorstack.tabs, editorstack.data,
+                                      icon='TextFileIcon')
```
ccordoba12: This should not be the name of the icon, but the icon returned by our icon manager, i.e.

    icon=ima.icon('TextFileIcon'))

Please change this to reflect that.

ccordoba12:
The last thing to address here is that now we can increase the number of characters displayed in each file path. That number is determined in the `shorten_paths` function of `widgets/fileswitcher.py`.

@dalthviz, please play with that function to see if you can improve it to display more characters.

spyder/widgets/fileswitcher.py
```diff
@@ -536,11 +589,22 @@ def setup_file_list(self, filter_text, current_path):
...
-            text = result[-1]
             path = paths[index]
-            item = QListWidgetItem(ima.icon('FileIcon'), text)
+            icon = icons[index]
+            text = ''
+            try:
+                title = self.widgets[index][1].get_plugin_title().split(' - ')
+                if plugin != title[0]:
+                    plugin = title[0]
+                    text += '<big><b>' + plugin + '</b></big><br>'
```
ccordoba12: Let's add some vertical separation between the plugin title and its list of files/tabs.

ccordoba12:
@dalthviz, please post a screenshot to see how is this looking :-)

spyder/app/mainwindow.py
```diff
@@ -2720,6 +2745,36 @@ def show_tour(self, index):
...
+            self.fileswitcher.is_visible = False
+            return
+        self.fileswitcher.setup()
+        self.fileswitcher.show()
+        self.fileswitcher.is_visible = True
+
+    def open_symbolfinder(self):
+        """Open symbol list management dialog box."""
+        self.open_fileswitcher()
+        self.fileswitcher.set_search_text('@')
```
ccordoba12: When opening the symbol switcher we should give focus to the Editor because it has no meaning for the other plugins, right?

dalthviz: 👍

dalthviz:
Preview:

fileswitch

ccordoba12:
Why don't the icons for the file and symbol switchers appear in the File toolbar in your screenshot?

dalthviz:
I didn't add the actions in the `file_toolbar_actions` list. The last commit do that 👍

ccordoba12:
I think this is ready, great work @dalthviz!

PR #4627, opened 2017-06-20
PR: Add pyopengl to Linux wheels only

ccordoba12:
The problem is there are no wheels for `pyopengl` on Windows or macOS, so people would need a compiler to pip-install Spyder in those platforms.


ccordoba12:
There aren't wheels for Linux either, but it's easier to install a compiler there anyway.

PR #4638, opened 2017-06-22
PR: Implement "Replace in selection" in the Editor

### LINKED ISSUE #358: Find/Replace widget: add option to find/replace in selection only
_From [stream.n...@gmx.de](https://code.google.com/u/104425159064108662638/) on 2010-09-13T03:06:25Z_

 At the moment the Find-Replace dialog allows only to replace single tokens or all of them in the current document. In addition I would find it very useful to have a checkbox like "Replace in selection only" and "Replace All" means then that all of the occurrences in the selection only are replaced. What version of the product are you using? On what operating system? 2.0.0 Beta 3

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=358_

spillz:
Fixes #358 

----

Supports replacing all instances of the search text with the replace text in a selection of text. Changes find/replace dialog to add buttons for "replace in selection" and "replace all" and removes the "replace all" checkbox. Tweaks behavior of automatically filling the searchbox depending on whether multiple lines were selected.

A refinement would be to either add shortcuts for the three buttons or combine the three buttons into one and use radio/combo to control scope.

### DIFF
--- a/spyder/widgets/findreplace.py
+++ b/spyder/widgets/findreplace.py
@@ -197,7 +197,7 @@ class FindReplace(QWidget):
                                    name='Find previous', parent=parent)
         togglefind = config_shortcut(self.show, context='_',
                                      name='Find text', parent=parent)
-        togglereplace = config_shortcut(self.toggle_replace_widgets,
+        togglereplace = config_shortcut(self.show_replace,
                                         context='_', name='Replace text',
                                         parent=parent)
         hide = config_shortcut(self.hide, context='_', name='hide find and replace',
@@ -488,18 +488,6 @@ class FindReplace(QWidget):
     @Slot()
     def replace_find_selection(self, focus_replace_text=False):
         """Replace and find in the current selection"""
-        def _word_replacer(match):
-            """Used to replace only match objects that are whole
-            words when the user has selected whole word search"""
-            unchanged = match.group(0)
-            if match.start(0)>0 and re.search(r'\w', seltxt[match.start(0)-1], flags=word_flags):
-                return unchanged
-            if match.end(0)<len(seltxt) and re.search(r'\w', seltxt[match.end(0)+1], flags=word_flags):
-                return unchanged
-            if re.search(r'\W', unchanged, flags=word_flags):
-                return unchanged
-            return match.expand(replace_text) 
-
         if (self.editor is not None):
             replace_text = to_text_string(self.replace_text.currentText())
             search_text = to_text_string(self.search_text.currentText())
@@ -517,8 +505,12 @@ class FindReplace(QWidget):
             if words:
                 #If whole words is checked we need to check that each match
                 #is actually a whole word before replacing
-                word_flags = 0 if case else re.IGNORECASE
-                replacement = re.sub(pattern, _word_replacer, seltxt, flags=re_flags)
+                try:
+                    re.compile(pattern)
+                except re.error:
+                    return #if the pattern won't compile cancel the find/replace
+                word_pattern = r'\b{pattern}\b'.format(pattern = pattern)
+                replacement = re.sub(word_pattern, replace_text, seltxt, flags=re_flags)
             else:
                 replacement = re.sub(pattern, replace_text, seltxt, flags=re_flags)
             if replacement != seltxt:
--- a/spyder/widgets/findreplace.py
+++ b/spyder/widgets/findreplace.py
@@ -238,7 +238,7 @@ class FindReplace(QWidget):
             else:
                 self.clear_matches()
         
-    def show(self, hide_replace = True):
+    def show(self, hide_replace=True):
         """Overrides Qt Method"""
         QWidget.show(self)
         self.visibility_changed.emit(True)
@@ -247,10 +247,12 @@ class FindReplace(QWidget):
                 if self.replace_widgets[0].isVisible():
                     self.hide_replace()
             text = self.editor.get_selected_text()
+            # When selecting several lines, and replace box is activated the
+            # text won't be replaced for the selection
             if hide_replace or len(text.splitlines())<=1:
                 highlighted = True
-                # If no text is highlighted for search, use whatever word is under
-                # the cursor
+                # If no text is highlighted for search, use whatever word is
+                # under the cursor
                 if not text:
                     highlighted = False
                     try:
@@ -283,7 +285,7 @@ class FindReplace(QWidget):
         
     def show_replace(self):
         """Show replace widgets"""
-        self.show(hide_replace = False)
+        self.show(hide_replace=False)
         for widget in self.replace_widgets:
             widget.show()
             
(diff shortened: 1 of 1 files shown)


ccordoba12:
@spillz, thanks for the extra effort. I'll let @rlaverde do the review.

spyder/widgets/findreplace.py
```diff
@@ -444,10 +461,41 @@ def replace_find(self, focus_replace_text=False):
...
                 self.replace_text.setFocus()
+
+    @Slot()
+    def replace_find_all(self, focus_replace_text=False):
+        """Replace and find all matching occurrences"""
+        self.replace_find(focus_replace_text, replace_all=True)
+
+                
+    @Slot()
+    def replace_find_selection(self, focus_replace_text=False):
```
rlaverde: I'm not sure if using a custom search is the best way, because the other searching methods use the [findText](http://doc.qt.io/qt-5/qwebengineview.html#findText) function of `QWebEngineView`.

Although I'm not sure if this could be implemented using `QWebEngineView.findText` because it doesn't support searching in a selection, also multiple selection should be managed adding more complexity.

spyder/widgets/findreplace.py
```diff
@@ -329,7 +345,8 @@ def find_previous(self):
     def text_has_been_edited(self, text):
         """Find text has been edited (this slot won't be triggered when 
         setting the search pattern combo box text programmatically"""
-        self.find(changed=True, forward=True, start_highlight_timer=True)
+        if not self.replace_widgets[0].isVisible() or len(to_text_string(self.editor.get_selected_text()).splitlines())<=1:
```
rlaverde: Add a comment for explaining this behavior too: "When selecting several lines, and replace box is activated, dynamically search is deactivated, for avoid changing the selection."

rlaverde:
@ccordoba12 

I think this plugin need a refactoring (not as part of this PR) the different searching functions `find` `replace_find` `replace_find_selection` should use the same "backend", there is duplicated logic and functions use different ways to do something very similar.

ccordoba12:
Ok, but that's for Spyder 4.

spyder/widgets/findreplace.py
ccordoba12: Also, this line is way to long. Please break it using a `\`

spyder/widgets/findreplace.py
```diff
@@ -444,10 +461,41 @@ def replace_find(self, focus_replace_text=False):
...
+            search_text = to_text_string(self.search_text.currentText())
+            pattern = search_text if self.re_button.isChecked() else None
+            case = self.case_button.isChecked()
+            re_flags = 0 if case else re.IGNORECASE
+
+            cursor = self.editor.textCursor()
+            cursor.beginEditBlock()
+            seltxt = to_text_string(self.editor.get_selected_text())
+            if not pattern:
+                replacement = re.sub(re.escape(search_text), re.escape(replace_text), seltxt, flags=re_flags)
```
ccordoba12: This line is also too long. Please make it in 79 columns

ccordoba12:
@rlaverde, did you test this locally to see if it's working as expected?

rlaverde:
> did you test this locally to see if it's working as expected?

yes

spyder/widgets/findreplace.py
spillz: True that there is a small risk that there are different results for the two regex implementations. I think the risk is relatively small because we don't highlight find results when the user chooses a selection. And there's always UNDO...

spillz: Actually a bigger inconsistency is that I didn't implement the whole word restriction. DOH! And it might ultimately be simpler to use the QT implementation because python's re doesn't handle whole words. Let me give this some thought over the weekend.

spillz:
The latest changes handle the whole word option and take care of deactivating the dynamic search multiple line selections. In my testing I noticed a few little bugs in the way that the original implementation of "replace all" works. I haven't fixed that here, but that could be the subject of a different PR.

ccordoba12:
Our tests are failing because you left a `print` statement behind.

> In my testing I noticed a few little bugs in the way that the original implementation of "replace all" works

Could you fix them here too? I mean, if they are simple enough, we could use the opportunity to fix them.

spillz:
>Could you fix them here too? I mean, if they are simple enough, we could use the opportunity to fix them.

They aren't trivial to fix. They stem from the mixed use of QT and python re in the original implementation.

Sorry about the print statement.  Removed it.

ccordoba12:
> They aren't trivial to fix. They stem from the mixed use of QT and python re in the original implementation.

Ok, no problem. Please open an issue about them so we don't forget to fix them.

> Sorry about the print statement. Removed it.

No problem, that's why we have a test to check for no print's.

ccordoba12:
@rlaverde, please give a last review to @spillz changes to see if we can merge them after his last changes.

spyder/widgets/findreplace.py
```diff
@@ -444,10 +472,60 @@ def replace_find(self, focus_replace_text=False):
...
+            cursor.beginEditBlock()
+            seltxt = to_text_string(self.editor.get_selected_text())
+            if not pattern:
+                pattern = re.escape(search_text)
+                replace_text = re.escape(replace_text)
+            if words:
+                #If whole words is checked we need to check that each match
+                #is actually a whole word before replacing
+                word_flags = 0 if case else re.IGNORECASE
+                replacement = re.sub(pattern, _word_replacer, seltxt, flags=re_flags)
```
rlaverde: Why _word_replacer is needed? I think that this could be achieved changing the pattern, like other functions do:
```
if words:
    pattern = r"\b{}\b".format(patern)
```

spillz:
Good point. I think I mistakenly thought the original find implementation
was rejecting any matches if there were spaces in the search term but it
doesn't look like that's the case. It does seem to match the "\b...\b"

But if someone puts in "\" as their regex then they will be replacing
instances of "\b" that start with a whole word. But maybe it is enough to
make sure the pattern compiles before enclosing it in \b?


On Jun 27, 2017 5:47 PM, "Rafael Laverde" <notifications@github.com> wrote:

*@rlaverde* commented on this pull request.
------------------------------

In spyder/widgets/findreplace.py
<https://github.com/spyder-ide/spyder/pull/4638#discussion_r124405002>:

> +            pattern = search_text if self.re_button.isChecked() else None
+            case = self.case_button.isChecked()
+            words = self.words_button.isChecked()
+            re_flags = re.MULTILINE if case else re.IGNORECASE|re.MULTILINE
+
+            cursor = self.editor.textCursor()
+            cursor.beginEditBlock()
+            seltxt = to_text_string(self.editor.get_selected_text())
+            if not pattern:
+                pattern = re.escape(search_text)
+                replace_text = re.escape(replace_text)
+            if words:
+                #If whole words is checked we need to check that each match
+                #is actually a whole word before replacing
+                word_flags = 0 if case else re.IGNORECASE
+                replacement = re.sub(pattern, _word_replacer, seltxt,
flags=re_flags)

Why _word_replacer is needed? I think that this could be achieved changing
the pattern, like other functions do:

if words:
    pattern = r"\b{}\b".format(patern)

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/spyder-ide/spyder/pull/4638#pullrequestreview-46690520>,
or mute the thread

ccordoba12:
@rlaverde, please take a decision about this one. We could leave it for 3.2.1 if you prefer.

rlaverde:
>But if someone puts in "\" as their regex then they will be replacing instances of "\b" that start with a whole word. But maybe it is enough to make sure the pattern compiles before enclosing it in \b?

Yes, I also think that will be enough, please implement it using "\b...\b"

>@rlaverde, please take a decision about this one. We could leave it for 3.2.1 if you prefer.

I think this could be in 3.2, Only the change for whole words is missing.

ccordoba12:
@spillz, please address @rlaverde comment ASAP if you want to see this in 3.2. You have until tomorrow (Wednesday) to do that.

Else I'll move this for 3.2.1.

spillz:
I should be able to do it this evening (US Eastern time).

ccordoba12:
That's fine. Thanks a lot for taking the time to do it.

ccordoba12:
@rlaverde, please take a look at this one and approve it, if it's ready.

ccordoba12:
Thanks for your work @spillz!

PR #4644, opened 2017-06-24
PR: Add a check to verify compatibility to plugins

### LINKED ISSUE #4641: Add a check_compatibility method to PluginMixin
On 3.x

https://github.com/spyder-ide/spyder/blob/master/spyder/api/plugins.py

Create a new method for `class SpyderPluginWidget(PluginWidget):`

```python
# For instance Qt4 can be checked here or other things...

    def check_compatibility(self):
        """
        This method can be implemented to check compatibility of plugin on a given os.

        `message` should give information in case of non compatibility:
        For example: 'This plugin does not work with Qt4'
        """
        message = ''
        valid = True
        return valid, message

```

On top of this we need to change the loading of plugins to load it only if compatible

@ccordoba12, @andfoy, @rlaverde  comments, ideas?

dalthviz:
Fixes #4641

### DIFF
--- a/spyder/plugins/__init__.py
+++ b/spyder/plugins/__init__.py
@@ -27,7 +27,7 @@ from qtpy import PYQT5
 from qtpy.QtCore import QEvent, QObject, QPoint, Qt, Signal
 from qtpy.QtGui import QCursor, QKeySequence
 from qtpy.QtWidgets import (QApplication, QDockWidget, QMainWindow,
-                            QShortcut, QTabBar, QWidget)
+                            QShortcut, QTabBar, QWidget, QMessageBox)
 
 # Local imports
 from spyder.config.base import _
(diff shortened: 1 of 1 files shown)


spyder/plugins/__init__.py
```diff
@@ -538,12 +538,30 @@ class SpyderPluginWidget(QWidget, SpyderPluginMixin):
...
             SpyderPluginMixin.__init__(self, parent)
-        
+            if not check_compatibility:
+                messageBox = QMessageBox(self)
+                messageBox.setWindowModality(Qt.NonModal)
+                messageBox.setAttribute(Qt.WA_DeleteOnClose)
+                messageBox.setWindowTitle('Compatibility Check')
+                messageBox.setText(message)
+                messageBox.setStandardButtons(QMessageBox.Ok)
+                messageBox.show()
```
ccordoba12: Let's put all this  in another method called show_compatibility_message, to not repeat twice in different places.

spyder/plugins/__init__.py
```diff
@@ -591,3 +597,25 @@ def get_plugin_actions(self):
     def register_plugin(self):
         """Register plugin in Spyder's main window"""
         raise NotImplementedError
+
+    def check_compatibility(self):
+        """
+        This method can be implemented to check compatibility of plugin on
+        a given os.
```
ccordoba12: Please change this to

    ... to check compatibility of a plugin for a given condition.

ccordoba12: Better instead

    ... to check compatibility of the plugin for a given condition.

PR #4651, opened 2017-06-25
PR: Avoid comments to generate pep8 warnings

### LINKED ISSUE #1785: Spyder comments (Ctrl+1 and Ctrl+4) generate pep8 warnings

Ericvulpi:
Fixes #1785.

----

Comment function : adds a space after the comment string (# for Python)
Uncomment function : removes the comment string and removes one space
if there is one


Ericvulpi:
3rd time is the charm ... hopefully

ccordoba12:
@Ericvulpi, thanks a lot for your contribution.

spyder/widgets/sourcecode/codeeditor.py
```diff
@@ -2206,11 +2206,12 @@ def toggle_comment(self):
...
-        self.add_prefix(self.comment_string)
+        self.add_prefix(self.comment_string + " ")
 
     def uncomment(self):
         """Uncomment current line or selection."""
         self.remove_prefix(self.comment_string)
+        self.remove_prefix(" ")
 
     def __blockcomment_bar(self):
         return self.comment_string + '='*(79-len(self.comment_string))
```
dalthviz: @Ericvulpi I think that we need to change the line above like you proposed adding a space  (`' '`) for the block comments. It should be something like:

```python
        return self.comment_string + ' ' + '=' * (78 - len(self.comment_string))
```

You can test this with `Ctrl+4` (blockcomment) and `Ctrl+5` (unblockcomment). The change should make a block comment that starts with `# ===...` instead of `#===...` (note the space between the `#` and the `=` symbols).

Ericvulpi: It was right under my nose! but i dont use block comments so didn't notice it

ccordoba12:
@dalthviz, do you approve this one?

dalthviz:
@ccordoba12 I think the changes are ok, however I don't know the reason for the tests to fail.

ccordoba12:
@Ericvulpi, please rebase this one on top of our latest 3.x branch to fix the test errors. Then I'll merge :-)

spyder/widgets/sourcecode/codeeditor.py
```diff
@@ -2206,14 +2206,15 @@ def toggle_comment(self):
...
+        self.add_prefix(self.comment_string + " ")
 
     def uncomment(self):
         """Uncomment current line or selection."""
         self.remove_prefix(self.comment_string)
+        self.remove_prefix(" ")
 
     def __blockcomment_bar(self):
-        return self.comment_string + '='*(79-len(self.comment_string))
+        return self.comment_string + ' ' + '=' * (78 - len(self.comment_string))
```
goanpeca: Why is that hardcoded 78?

dalthviz: It is because of the space added between the comment string and the '='. Maybe this could be rewritten as:

```python
return self.comment_string + ' ' + '=' * (79 - len(self.comment_string + ' '))
```

goanpeca: No I mean why is 79 hardcoded!!, that is an adjustable setting

dalthviz: Oh, then we should depend on that config here?, I think the 79 is from the maximum line length in PEP8

goanpeca: @ccordoba12 ?

ccordoba12: @dalthviz is right, 79 is the default column width for pep8. This width can be configured for pycodestyle, but currently is not connected to our config system.

So let's leave 79 for now.

goanpeca: @dalthviz please open an issue to keep track of this missing connection

ccordoba12:
Thanks @Ericvulpi for your contribution, this is going in!

bcolsen:
@ccordoba12 This patch breaks backwards compatibility when uncommenting scripts with comments in code blocks.  eg.

Script with old comment: 
```
for i in range(10):
#    print(i) # Made with old comment system
    b = i + i
```

Now uncomment with latest 3.2: 
```
for i in range(10):
   print(i) # Error
    b = i + i
```
Imagine a whole block of those with different block depths.(Maybe I this is a bad practice but I don't think I'm alone in this)

I can't even shift-tab and tab back to replace the 3 spaces with 4 (maybe another bug here)

I think pep8 errors are small compared to code errors introduced from uncommenting.

I would suggest holding off on this until we get comments like this:
```
for i in range(10):
    # print(i) # Made with old comment system
    b = i + i
```
This might break backward compatibility but at least I think this is waiting until 4.0.

ccordoba12:
@bcolsen, thanks for your feedback. I think you're right, we need to revert this change.

@dalthviz, please create a PR to revert this. @Ericvulpi, sorry for this, we'll fix this for good in Spyder 4.

PR #4661, opened 2017-06-27
PR: Fix fileswitcher size during its setup

### LINKED ISSUE #4648: File switcher is changing size when changing from symbols to files
Easy way to reproduce this

1. Press the file switcher button in the file toolbar.
2. Press `Esc` to hide the file switcher.
3. Press the symbol finder button in the file toolbar.
4. Press again the file switcher button.

@dalthviz, besides fixing this error, we have to make sure to remove all text present in the switcher every time we're going to show it again.

dalthviz:
Fixes #4648

### DIFF
--- a/spyder/widgets/fileswitcher.py
+++ b/spyder/widgets/fileswitcher.py
@@ -673,6 +673,9 @@ class FileSwitcher(QDialog):
         oedata = self.get_symbol_list()
         icons = get_python_symbol_icons(oedata)
 
+        # The list of shorten paths here is needed in order to have the same
+        # point of measurement for the list widget width as in the file list
+        # See issue 4648
         short_paths = shorten_paths(self.paths, self.save_status)
         symbol_list = process_python_symbol_data(oedata)
         line_fold_token = [(item[0], item[2], item[3]) for item in symbol_list]
--- a/spyder/widgets/fileswitcher.py
+++ b/spyder/widgets/fileswitcher.py
@@ -477,6 +477,9 @@ class FileSwitcher(QDialog):
 
     def previous_row(self):
         """Select previous row in list widget."""
+        if self.mode == self.SYMBOL_MODE:
+            self.select_row(-1)
+            return
         prev_row = self.current_row() - 1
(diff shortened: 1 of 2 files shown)


spyder/widgets/fileswitcher.py
```diff
@@ -667,6 +673,7 @@ def setup_symbol_list(self, filter_text, current_path):
         oedata = self.get_symbol_list()
         icons = get_python_symbol_icons(oedata)
 
+        short_paths = shorten_paths(self.paths, self.save_status)
```
ccordoba12: Please add a comment above this line to explain why this is needed.

spyder/widgets/fileswitcher.py
```diff
@@ -652,7 +658,7 @@ def setup_file_list(self, filter_text, current_path):
             self.set_current_row(self.filtered_path.index(current_path))
         elif self.filtered_path:
             self.set_current_row(0)
-        self.fix_size(short_paths, extra=200)
+        self.fix_size(short_paths, extra=100)
```
ccordoba12: Why the change from 200 to 100?

spyder/widgets/fileswitcher.py
```diff
@@ -488,6 +491,9 @@ def previous_row(self):
 
     def next_row(self):
         """Select next row in list widget."""
+        if self.mode == self.SYMBOL_MODE:
+            self.select_row(+1)
+            return
```
ccordoba12: Why is this needed?

spyder/widgets/fileswitcher.py
dalthviz: Since the list for the symbols normally has less width, I was thinking that would be better to reduce the width given by the `extra` param.

spyder/widgets/fileswitcher.py
dalthviz: This is needed since I found a new bug that doesn't allow you to move with the symbol finder. Since is the same element, I thought this could be included in this PR, should I open an issue with a different PR?

Issue #4663, opened 2017-06-27
Inspecting Numpy bool arrays in Var. Explorer uses deprecated '-' operator in Numpy


GKosiba:
## Description of your problem
(..)/site-packages/spyderlib/widgets/arrayeditor.py", line 249, in data
/(self.vmax-self.vmin)
TypeError: numpy boolean subtract, the operator '-', is deprecated, use the bitwise_xor, the '^' operator, or the logical_xor function instead. 
**What steps will reproduce the problem?**

1. import numpy as np
2. test = np.zeros((3,3),dtype='bool')
3. Double click test in variable explorer window

**What is the expected output? What do you see instead?**
Popup window of the array w/ values. Crash (Seg. Fault)

**Please provide any additional information below**

## Versions and main components

* Spyder Version: 3.1.4
* Python Version: 2.7.5
* Qt Version: 4.8.5
* PyQt Version: 4.11.4
* Operating system: RHEL7

## Dependencies
jedi >=0.9.0     :  0.10.2 (OK)
matplotlib >=1.0 :  2.0.2 (OK)
nbconvert >=4.0  :  5.2.1 (OK)
numpy >=1.7      :  1.14.0.dev0+03283c2 (OK)
pandas >=0.13.1  :  0.18.1 (OK)
pep8 >=0.6       :  1.7.0 (OK)
psutil >=0.3     :  5.2.2 (OK)
pyflakes >=0.5.0 :  1.5.0 (OK)
pygments >=2.0   :  2.2.0 (OK)
pylint >=0.25    :  1.7.2 (OK)
qtconsole >=4.2.0:  4.3.0 (OK)
rope >=0.9.4     :  0.10.5 (OK)
sphinx >=0.6.6   :  1.6.2 (OK)
sympy >=0.7.3    :  None (NOK)

Please go to the menu entry `Help > Optional Dependencies` (or
`Help > Dependencies`), press the button `Copy to clipboard`
and paste the contents below:

ccordoba12:
Thanks for reporting. The fix seems simple enough, would you like to submit a pull request for it? If you, please make your PR against our 3.x branch.

PR #4667, opened 2017-06-27
PR: Update ArrayEditor to correct deprecated Numpy operator

GKosiba:
Changed numpy boolean subract operator '-' to np.logical_xor. As suggested in the numpy deprec warning.
Tested locally with no error originally described in #4663

### DIFF
--- a/spyder/widgets/variableexplorer/arrayeditor.py
+++ b/spyder/widgets/variableexplorer/arrayeditor.py
@@ -267,7 +267,7 @@ class ArrayModel(QAbstractTableModel):
           and value is not np.ma.masked:
             hue = self.hue0+\
                   self.dhue*(float(self.vmax)-self.color_func(value)) \
-                  /(float(self.vmax)-self.vmin)
+                  /(float(self.vmax)-self.vmin) #float convert to handle bool arrays
             hue = float(np.abs(hue))
             color = QColor.fromHsvF(hue, self.sat, self.val, self.alp)
             return to_qvariant(color)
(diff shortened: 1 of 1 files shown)


ccordoba12:
Is this backwards compatible? I mean, does it work with previous Numpy versions?

ccordoba12:
And of course. thanks a lot for your contribution!!

jitseniesen:
What about when the array loaded in the variable explorer does not have boolean entries?

GKosiba:
Sorry, didn't have a lot of time yesterday, but that wasn't a fix.
Essentially this line is scaling the hue of each index in the pop-up 'array' window relative to the maximum value in the array. 

This new edit calculates this scaling factor (max-value)/(max-min) with max as float each time. The issue only occurs for boolean with the deprecated minus operator but float of 0 or 1 minus value of 0 or 1 gives the same value for the scaling factor and was converted to float anyways in the next line.

Tested with bool and non-bool array types and older numpy versions.

ccordoba12:
@jitseniesen, do you approve this one?

jitseniesen:
@GKosiba Looks good; but one detail: could you please add a short comment explaining why the conversion to `float` is there? e.g. "convert to float to handle bool arrays". I am afraid the conversion will puzzle us later.

GKosiba:
@ccordoba12 @jitseniesen Comment added

ccordoba12:
Thanks @GKosiba for your contribution and @jitseniesen for reviewing it.

This is going in!

Issue #4679, opened 2017-07-01
Add IPython to Help > Dependencies


bcolsen:
## Description of your problem
This might be planned as part of #4524 already, but I noticed that IPython wasn't listed as a dependency in Help->Dependencies. I noticed this when had a problem with #4214 and an upgrade IPython from 5.1 to 6.1 to fix it.

## Versions and main components

* Spyder Version: 3.x branch
* Python Version: 3.6
* Qt Version: 5.6
* PyQt Version: 5.6
* Operating system:

## Dependencies

cython >=0.21    :  0.25.2 (OK)
jedi >=0.9.0     :  0.10.2 (OK)
nbconvert >=4.0  :  4.2.0 (OK)
numpy >=1.7      :  1.11.3 (OK)
pandas >=0.13.1  :  0.19.2 (OK)
psutil >=0.3     :  5.0.1 (OK)
pycodestyle >=2.3:  2.3.1 (OK)
pyflakes >=0.6.0 :  1.5.0 (OK)
pygments >=2.0   :  2.1.3 (OK)
pylint >=0.25    :  1.6.4 (OK)
qtconsole >=4.2.0:  4.3.0 (OK)
rope >=0.9.4     :  0.9.4-1 (OK)
sphinx >=0.6.6   :  1.5.1 (OK)
sympy >=0.7.3    :  1.0 (OK)

ccordoba12:
I didn't thought it was important to list it. I mean, IPython is an indirect dependency of `qtconsole`, which is really our direct dependency.

But your problem proved that we should definitely list IPython. Could you create a PR for this? The fix is really simple, you only needs to add it here:

https://github.com/spyder-ide/spyder/blob/master/spyder/plugins/ipythonconsole.py#L70

The only quirk is that there won't be IPython `>=6` for Python 2. And you can set `4.0` as the minimal version, I think that's fine.

bcolsen:
Looks simple enough.  I'll give it a shot.

PR #4682, opened 2017-07-01
PR: Added IPython >= 4.0 as a dependency

bcolsen:
This implements #4679 to make IPython >= 4.0 a dependency


spyder/plugins/ipythonconsole.py
```diff
@@ -72,6 +72,9 @@
 dependencies.add("qtconsole", _("Integrate the IPython console"),
                  required_version=QTCONSOLE_REQVER)
 
+IPYTHON_REQVER = ">=4.0"
```
ccordoba12: This needs to be more fine grained, like this

     IPYTHON_REQVER = ">=4.0;<6.0" if PY2 else ">=4.0"

PR #4685, opened 2017-07-01
PR: Format change in array editor

### LINKED ISSUE #4684: Array editor: Change default float formatter from %.3f to %.6g
Currently this array:
```
a = np.array([3.2e-9,45.6e56,22.3,57666.6,0.00000004])
```
looks like this in the array editor with the default %.3f formatting:

%.3f

I can't tell what most of those numbers are.

But it could look like this with %.6g formatting:

screenshot from 2017-07-01 02-04-30

If you think this is a good idea I should be able to do a PR.

pecox:
Fixes #4684.

Long numbers are better viewed in shorter format
according to issue [#4684](https://github.com/spyder-ide/spyder/issues/4684)


jitseniesen:
What about making the same change to the dataframe editor?

ccordoba12:
Sure, I agree with that too. @jitseniesen, please explain @pecox how to do it.

@pecox, please rebase your PR in top of 3.x so we can merge it in Spyder 3.2. You can find how to do that here:

https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md#changing-base-branch

bcolsen:
Wow speedy PR. Thanks @pecox !

I think you will also need to change %.3f in the SUPPORTED_FORMATS dict in line 44 
https://github.com/spyder-ide/spyder/blob/master/spyder/widgets/variableexplorer/arrayeditor.py#L44
```
SUPPORTED_FORMATS = {
                     'single': '%.3f',
                     'double': '%.3f',
                     'float_': '%.3f',
                     'longfloat': '%.3f',
                     'float16': '%.3f',
                     'float32': '%.3f',
                     'float64': '%.3f',
                     'float96': '%.3f',
                     'float128': '%.3f',
```

About the rebase:
First thing `branch` is a confusing name for a branch but I think it should still work. For this one I would pick something more meaningful like `array_editor`.

I'm going to preface this with saying I'm not that great at git, but I'll try to help.

"First, rewinding head to replay your work on top of it..."
This is what a rebase does it takes all your commit you made and applies it to a new base. 
https://git-scm.com/book/en/v2/Git-Branching-Rebasing
https://stackoverflow.com/questions/29914052/i-cant-understand-the-behaviour-of-git-rebase-onto

So that's fine, but then you should see something like:
```
Applying: Format change in array editor
Applying: Default float format change
```
If you see this things are going good.

If not then something's wrong. What's the output of:
`git log`

ccordoba12:
@pecox, I'm closing this PR in favor of PR #4955, which is against `3.x`.

I think you couldn't move this PR to `3.x` because you made it *directly* against `master`. So for next time, please always use a branch to submit your work.

PR #4691, opened 2017-07-03
PR: Implement dedicated IPython consoles

### LINKED ISSUE #4418: No option for dedicated IPython console
Upon hearing about the pending removal of the Python console option, I looked into migrating my workflow to IPython. However, my workflow currently involves using the "Execute in a new dedicated Python console" option. There is no equivalent for IPython.

There is the "clear all variables" IPython option, but this is not equivalent. In my current workflow, I could have, say, 5 scripts open, select each one and hit F5, and it would create a new console for each. I don't see a way to do this with IPython, other than opening 5 new windows, selecting each one in sequence while selecting my script, and hitting F5. This is inferior for rapidly iterating over code, because you need to e.g. keep track of which window belongs to which script, etc., which Spyder does automatically for the dedicated Python case.

I guess this could be solved simply by adding a "Execute in a new dedicated IPython console" option, alongside a "Dedicated IPython console" sub-list. Should this be done?

ccordoba12:
Fixes #4418.

This is part of issue #4524.

----

TODO

- [x] Close all Matplotlib plots before a new run.

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -2107,7 +2107,8 @@ class MainWindow(QMainWindow):
 
     def remove_tmpdir(self):
         """Remove Spyder temporary directory"""
-        shutil.rmtree(programs.TEMPDIR, ignore_errors=True)
+        if CONF.get('main', 'single_instance') and not self.new_instance:
+            shutil.rmtree(programs.TEMPDIR, ignore_errors=True)
 
     def closeEvent(self, event):
         """closeEvent reimplementation"""
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -2433,8 +2433,8 @@ class MainWindow(QMainWindow):
 
     def execute_in_external_console(self, lines, focus_to_editor):
         """
-        Execute lines in external or IPython console and eventually set focus
-        to the editor
+        Execute lines in IPython console and eventually set focus
+        to the Editor.
         """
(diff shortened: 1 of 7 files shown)


ccordoba12:
Let's merge this for now to keep things moving. I'll probably need to revisit this in 3.2.1, but it's working reasonably well for now.

Eric89GXL:
Problems during migration (so far):

1. If there is no console currently open, running a script crashes the IPython kernel, it restarts it but doesn't try to re-run the script. I can't see why the kernel dies because it's restarted immediately (I can only see a flash that it failed). I'll try updating my IPython / Jupyter to try to fix this, but in the meantime it's difficult to debug.

2. Whenever a script would crash before, I could do `import pdb; pdb.pm()` to immediately go to the failing code. This now puts me into o stack frame within `IPython/core/compilerop.py` down to `IPython/core/interactiveshell.py` (two frames) so I can't get to any variables.

Should I open a new issue to track such migration problems?

ccordoba12:
@Eric89GXL, yes, please open a new issue for these problems. I don't understand too well you first problem though, so could you post an animated GIF to better describe it? You can do it with a very nice program called Licecap (if you're on Windows or Mac).

Eric89GXL:
The `pdb` problem seems to have gone away (maybe with a restart of Spyder?), I'll open new issues as things come up. Thanks @ccordoba12

ccordoba12:
Great to know!

PR #4700, opened 2017-07-05
PR: Add disambiguation for dedicated IPython consoles

### LINKED ISSUE #4694: Disambiguate dedicated IPython consoles
@dalthviz, please use a similar approach to the one you used in PR #3854 in the IPython console.

*Notes*:
* You only need to apply disambiguation to dedicated consoles.
* Dedicated consoles are created with the method called `create_client_for_file`.

----

For this we need to merge first PR #4691.

dalthviz:
Fixes #4694

### DIFF
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -909,6 +909,8 @@ class IPythonConsole(SpyderPluginWidget):
 
     @Slot()
     @Slot(bool)
+    @Slot(object)
+    @Slot(bool, object)
     def create_new_client(self, give_focus=True, filename=None):
         """Create a new client"""
         self.master_clients += 1
@@ -1386,13 +1388,13 @@ class IPythonConsole(SpyderPluginWidget):
     def disambiguate_fname(self, fname):
         """Generate a file name without ambiguation."""
         files_path_list = [filename for filename in self.filenames
-                           if filename is not None]
+                           if filename]
         return sourcecode.disambiguate_fname(files_path_list, fname)
 
     def update_tabs_text(self):
         """Update the text from the tabs."""
         for index, fname in enumerate(self.filenames):
-            if fname is not None:
+            if fname:
                 client = self.clients[index]
                 self.rename_client_tab(client, self.disambiguate_fname(fname))
 
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -909,9 +909,9 @@ class IPythonConsole(SpyderPluginWidget):
 
     @Slot()
     @Slot(bool)
-    @Slot(object)
-    @Slot(bool, object)
-    def create_new_client(self, give_focus=True, filename=None):
+    @Slot(str)
+    @Slot(bool, str)
+    def create_new_client(self, give_focus=True, filename=''):
         """Create a new client"""
         self.master_clients += 1
         client_id = dict(int_id=to_text_string(self.master_clients),
@@ -1362,7 +1362,7 @@ class IPythonConsole(SpyderPluginWidget):
             client.restart_kernel()
 
     #------ Public API (for tabs) ---------------------------------------------
-    def add_tab(self, widget, name, filename=None):
+    def add_tab(self, widget, name, filename=''):
         """Add tab"""
         self.clients.append(widget)
         index = self.tabwidget.addTab(widget, name)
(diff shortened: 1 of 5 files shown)


spyder/plugins/ipythonconsole.py
```diff
@@ -1355,25 +1360,42 @@ def restart_kernel(self):
...
         """
         Move tab (tabs themselves have already been moved by the tabwidget)
         """
+        filename = self.filenames.pop(index_from)
         client = self.clients.pop(index_from)
+        self.filenames.insert(index_to, filename)
         self.clients.insert(index_to, client)
         self.update_plugin_title.emit()
 
+    def get_tab_text(self, fname):
```
ccordoba12: get_tab_text -> disambiguate_fname

spyder/plugins/ipythonconsole.py
```diff
@@ -1355,25 +1360,42 @@ def restart_kernel(self):
...
+        """Get tab text without ambiguation."""
```
ccordoba12: Let's change this to

    """Generate a filename without ambiguation."""

spyder/plugins/ipythonconsole.py
```diff
@@ -1265,15 +1269,16 @@ def create_client_from_path(self, path):
...
+        self.create_new_client(filename=filename)
 
         # Don't increase the count of master clients
         self.master_clients -= 1
 
         # Rename client tab with filename
         client = self.get_current_client()
         client.allow_rename = False
-        self.rename_client_tab(client, filename)
+        fname = self.get_tab_text(filename)
```
ccordoba12: fname -> tab_text

spyder/plugins/ipythonconsole.py
```diff
@@ -1355,25 +1360,42 @@ def restart_kernel(self):
...
         client = self.clients.pop(index_from)
+        self.filenames.insert(index_to, filename)
         self.clients.insert(index_to, client)
         self.update_plugin_title.emit()
 
+    def get_tab_text(self, fname):
+        """Get tab text without ambiguation."""
+        files_path_list = [filename for filename in self.filenames
+                           if filename is not None]
+        return sourcecode.get_file_title(files_path_list, fname)
```
ccordoba12: Please rename `get_file_title` in `sourcecode` to `disambiguate_fname` and make the corresponding change in the Editor.

ccordoba12:
@dalthviz, I left several comments related to renaming variables and functions to better understand your code.

Also, this is missing a test. It could be like this:

1. Add it to test_ipythonconsole.py
2. You can create two clients with these filenames: `/tmp/a/b/c.py` and `/tmp/a/d/e.py`.
3. You must assert that the tab names are `b/c.py` and `d/e.py` (if I understand our disambiguation algorithm correctly :-).

ccordoba12:
I think this is ready. @dalthviz, could you post a screenshot to see how this looks, please?

spyder/plugins/ipythonconsole.py
```diff
@@ -907,7 +909,7 @@ def write_to_stdin(self, line):
 
     @Slot()
     @Slot(bool)
```
ccordoba12: Please add two more signatures below this line

    @Slot(object)
    @Slot(bool, object)

to cover all possible entries of `create_new_client`.

dalthviz:
Preview:

Running two files with paths like:
`.../a/b/c.py`
`.../a/d/c.py`

imagen

spyder/plugins/ipythonconsole.py
ccordoba12: Sorry, give me a moment to think this.

ccordoba12: This needs to be

    @Slot(str)
    @Slot(bool, str)

and you need to change below

    filename=''

instead of

    filename=None

spyder/plugins/ipythonconsole.py
```diff
@@ -1355,25 +1360,42 @@ def restart_kernel(self):
...
+        filename = self.filenames.pop(index_from)
         client = self.clients.pop(index_from)
+        self.filenames.insert(index_to, filename)
         self.clients.insert(index_to, client)
         self.update_plugin_title.emit()
 
+    def disambiguate_fname(self, fname):
+        """Generate a file name without ambiguation."""
+        files_path_list = [filename for filename in self.filenames
+                           if filename is not None]
```
ccordoba12: With my last suggested change, you need to change this line to

    if filename

instead of

    if filename is not None

spyder/plugins/ipythonconsole.py
```diff
@@ -1355,25 +1360,42 @@ def restart_kernel(self):
...
+        return sourcecode.disambiguate_fname(files_path_list, fname)
+
+    def update_tabs_text(self):
+        """Update the text from the tabs."""
+        for index, fname in enumerate(self.filenames):
+            if fname is not None:
```
ccordoba12: I think this also needs to change to

    if fname:

PR #4711, opened 2017-07-06
PR: Show complete exception (not only the traceback) in error QMessageBox.

### LINKED ISSUE #4698: Fancy exception handler does not show exception
At some point in the 3.x branch, a fancy exception handler was implemented which shows a window like ...

screenshot_20170704_221613

when Spyder encounters the following exception (which I forced):

```
  File "/home/jitse/misc/spyder/spyder/plugins/ipythonconsole.py", line 955, in <lambda>
    kc.started_channels.connect(lambda c=client: self.process_started(c))
  File "/home/jitse/misc/spyder/spyder/plugins/ipythonconsole.py", line 1387, in process_started
    self.variableexplorer.add_shellwidget(client.shellwidget)
  File "/home/jitse/misc/spyder/spyder/plugins/variableexplorer.py", line 143, in add_shellwidget
    nsb = NamespaceBrowser(self)
  File "/home/jitse/misc/spyder/spyder/widgets/variableexplorer/namespacebrowser.py", line 65, in __init__
    QWdget.__init__(self, parent)
NameError: name 'QWdget' is not defined
```

The window only shows the traceback, not the exception itself (`NameError: name 'QWdget' is not defined`). As a consequence (I presume), when using the "Submit to GitHub" button, this information is also not included. I think this will rather complicate the debugging for us developers.

* Spyder Version: 3.2.0.dev0 8dc91638
* Python Version: 3.6.1

## Dependencies
```
jedi >=0.9.0     :  0.9.0 (OK)
sympy >=0.7.3    :  1.0 (OK)

```

rlaverde:
Fixes #4698

spectacle m18475

### DIFF
--- a/spyder/widgets/sourcecode/base.py
+++ b/spyder/widgets/sourcecode/base.py
@@ -1270,7 +1270,7 @@ class ConsoleBaseWidget(TextEditBaseWidget):
     """Console base widget"""
     BRACE_MATCHING_SCOPE = ('sol', 'eol')
     COLOR_PATTERN = re.compile('\x01?\x1b\[(.*?)m\x02?')
-    traceback_available = Signal(str)
+    exception_occurred = Signal(str)
     userListActivated = Signal(int, str)
     completion_widget_activated = Signal(str)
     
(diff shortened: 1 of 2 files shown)


spyder/widgets/sourcecode/base.py
```diff
@@ -1379,20 +1379,17 @@ def append_text_to_shell(self, text, error, prompt):
             text = text[index+1:]
             self.clear()
         if error:
-            is_traceback = False
             for text in text.splitlines(True):
                 if text.startswith('  File') \
                 and not text.startswith('  File "<'):
-                    is_traceback = True
```
ccordoba12: We need to keep this to activate the dialog only when text printed in the internal console is *really* a traceback. Else the error dialog is shown for things like this

seleccion_001

which is no real error, just a warning printed through stderr :-)

ccordoba12:
Please also make that the dialog always *Show details*, so that people can decide if they are in front of a real or not.

Else, we could be flooded by tons of false positives.

PR #4720, opened 2017-07-08
PR: Avoid comments to generate pep8 warnings - bis

### LINKED ISSUE #1785: Spyder comments (Ctrl+1 and Ctrl+4) generate pep8 warnings

Ericvulpi:
Still attempts to fix #1785. 

Comment function :
Adds the comment string (# for Python) plus a space, after the indent, just before the text

Uncomment function :
Removes the comment string and removes one space if there is one after the comment string and if the indent length is not an even number std indent length (to avoid breaking code)


Ericvulpi:
This follows previous PR : https://github.com/spyder-ide/spyder/pull/4651

I'm giving it another try, more to get some feedback than to provide final solution

I see 2 main choices that should be discussed (maybe depending on which version of Spyder is the target) :
- When "commenting" several lines, should the comment string be added at a single position (lost to the left among the selected lines) or at the indent of each line => second solution is the one implemented in this PR
- When "uncommenting", I've proposed a check on the indent length : if it's an even number of std indent length then no space is removed to avoid breaking code. This may have the following side effect, as far as I can see :
    - break asci logo
    - mess with some alignment of bracket when a function is too long to fit on one line and has been split on multiple lines

Also, any advice on how to write proper code or do pull request is welcome (i'm new at this :)

ccordoba12:
@Ericvulpi, thanks for the extra effort. I'm leaving this for 3.2.1 because we're about to release 3.2 and (unfortunately) we don't have time to review your work.

Ericvulpi:
No rush.
And as @bcolsen mentioned, this may have to wait for 4.0 if the solution still breaks some code

ccordoba12:
@Ericvulpi, sorry, but @dalthviz is working on a new way to add comments to the Editor for Spyder 4 (in PR #3958) and he's going to solve #1785 as part of that effort.

Ericvulpi:
No problem, PR #3958 looks like a more thorough solution

PR #4731, opened 2017-07-11
PR: Make searching text to expand folded blocks

### LINKED ISSUE #4709: Find does not open code-folded blocks
This one is probably trickier.
if the target is inside a folded bloc, then searching for it does not make anything. Same with successive search.
The difficulty would be to open a folded bloc, and close it back automatically on successive F3 (expected behavior).

rlaverde:
Fixes #4709

### DIFF
--- a/spyder/utils/editor.py
+++ b/spyder/utils/editor.py
@@ -134,19 +134,25 @@ class TextHelper(object):
                                      column)
         if move:
             block = text_cursor.block()
-            # unfold parent fold trigger if the block is collapsed
-            try:
-                folding_panel = self._editor.panels.get('FoldingPanel')
-            except KeyError:
-                pass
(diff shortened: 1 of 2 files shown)


ccordoba12:
Can we inspect the text inside a folded block to see if we should expand it before expanding it?

rlaverde:
> Can we inspect the text inside a folded block to see if we should expand it before expanding it?

The search already do that, as I said in the issue, the search works and select the text inside the folded block (but It can be saw because the block is folded)

The solution that I implement only expands the folded block when the text is found inside it.

ccordoba12:
> The solution that I implement only expands the folded block when the text is found inside it.

Great! That's what I wanted to see. Merging then :-)

PR #4738, opened 2017-07-13
PR: Update pixmap height calculation

### LINKED ISSUE #4734: High DPI scaling messes up placement of breakpoint red circles
## The breakpoint red circle is drawn between two lines instead of on the relevant line when high DPI scaling is on.

1. Use 4K or other high DPI monitor with scaling turned on (may be unnecessary, I don't have an old one to test right now?)
2. In general preferences, check "Enable high DPI scaling"
3. Restart Spyder to enable high DPI scaling
3. Place a breakpoint in a file

The red circle should be nicely centered on the line with the intended breakpoint.  Instead the circle is drawn about half a line high and it is unclear which line it corresponds to.  Further, trying to turn it off often accidentally leads to turning another one on, but offset from the line you intended.

The breakpoint shown below is intended to be on line 8 with the command "pass", but shows up between lines 7 and 8.
breakpoint

Note, that turning off the scaling fixes the alignment, but then you are stuck with really small icons and circles.
breakpoint2

* Spyder Version: 3.1.4.dev0
* Python Version: 3.6.1
* Qt Version: 5.8.0
* PyQt Version: 5.8.1

## Dependencies

jedi >=0.9.0     :  0.10.0 (OK)
numpy >=1.7      :  1.11.3 (OK)
pylint >=0.25    :  1.6.5 (OK)
sphinx >=0.6.6   :  1.5.2 (OK)
sympy >=0.7.3    :  1.0 (OK)

DStauffman:
Fixes #4734 

----

The font height and pixmap height were in incompatible units when DPI scaling is on.  Dividing the qpixmap height by self.devicePixelRatio() converts it into device independent units.  See http://doc.qt.io/qt-5/qpixmap.html#devicePixelRatio

### DIFF
--- a/spyder/widgets/sourcecode/codeeditor.py
+++ b/spyder/widgets/sourcecode/codeeditor.py
@@ -1179,7 +1179,8 @@ class CodeEditor(TextEditBaseWidget):
         active_line_number = active_block.blockNumber() + 1
 
         def draw_pixmap(ytop, pixmap):
-            painter.drawPixmap(0, ytop + (font_height-pixmap.height()) / 2,
+            pixmap_height = pixmap.height() / pixmap.devicePixelRatio()
+            painter.drawPixmap(0, ytop + (font_height-pixmap_height) / 2,
                                pixmap)
 
         for top, line_number, block in self.visible_blocks:
--- a/spyder/widgets/sourcecode/codeeditor.py
+++ b/spyder/widgets/sourcecode/codeeditor.py
@@ -27,7 +27,7 @@ import sys
 import time
 
 # Third party imports
-from qtpy import is_pyqt46
+from qtpy import is_pyqt46, QT_VERSION
 from qtpy.compat import to_qvariant
 from qtpy.QtCore import QRect, QRegExp, QSize, Qt, QTimer, Signal, Slot
 from qtpy.QtGui import (QBrush, QColor, QCursor, QFont, QIntValidator,
@@ -52,6 +52,7 @@ from spyder.utils import icon_manager as ima
 from spyder.utils import syntaxhighlighters as sh
 from spyder.utils import encoding, sourcecode
 from spyder.utils.dochelpers import getobj
+from spyder.utils.programs import check_version
 from spyder.utils.qthelpers import add_actions, create_action, mimedata2url
 from spyder.utils.sourcecode import ALL_LANGUAGES, CELL_LANGUAGES
 from spyder.widgets.editortools import PythonCFM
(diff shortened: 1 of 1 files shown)


spyder/widgets/sourcecode/codeeditor.py
```diff
@@ -1179,7 +1179,8 @@ def linenumberarea_paint_event(self, event):
         active_line_number = active_block.blockNumber() + 1
 
         def draw_pixmap(ytop, pixmap):
-            painter.drawPixmap(0, ytop + (font_height-pixmap.height()) / 2,
+            pixmap_height = pixmap.height() / pixmap.devicePixelRatio()
```
rlaverde: `devicePixelRatio` was introduced in Qt5.5, that is making tests in Qt4 fail, maybe you could add a try.. except and leave the old behavior for Qt4

goanpeca: We have variables for checking Qt4 or Qt5 let's use those!

DStauffman: I updated the logic to use the existing is_pyqt46 flag and restored the original behavior for PyQt4.  It's as of yet untested on PyQt4, as I don't have that installed right now.  Let me know if it works, or I can test it later.

spyder/widgets/sourcecode/codeeditor.py
```diff
@@ -1179,7 +1179,12 @@ def linenumberarea_paint_event(self, event):
         active_line_number = active_block.blockNumber() + 1
 
         def draw_pixmap(ytop, pixmap):
-            painter.drawPixmap(0, ytop + (font_height-pixmap.height()) / 2,
+            if is_pyqt46:
```
goanpeca: I think we need 

```python
from qtpy import PYQT4, PYQT5

 if PYQT4:
```


https://github.com/spyder-ide/qtpy/blob/master/qtpy/__init__.py#L94

ccordoba12: Nop, it needs to be more fine-grained because we support PyQt >=5.2. So the solution is something along the lines

```python
from qtpy import QT_VERSION
QT55_VERSION = programs.check_version(QT_VERSION, "5.5", ">=")

if not QT55_VERSION:
    pixmap_height = pixmap.height()
else:
    ....
```

Of course, you need to add the imports and constants in the right places.

DStauffman: Ok, the more fine grained option is now in there.  Thanks for the help in knowing what to import.

PR #4289, opened 2017-03-22
[WIP] PR: Add basic snnnipet support to text editor.

rlaverde:
Fixes #588

-----------------------------
I added just a basic support, at the moment:
- load and save snippets to `snippets` directory in the config folder
- insert snippets in the text editor

TODO:
- Add variables to snippets `${1:'default_value'}`
- Expose saving functionality in editor config preferences
- Decouple some functionality in another class `Snippet` that will be make it more maintainable and extensible (It will be useful to add things like scope or language in a future)
- Add some basics snippets that will be distributed with spyder ?
- Add some tests and docstrings

I'm not sure what language should be use it, at the moment is using `json`, but I'm planning to use `yaml`, because It is easy to read and write (for humans), and allows comments.

PR #4751, opened 2017-07-15
PR: Migrate introspection services to use the Language Server Protocol (LSP)

### LINKED ISSUE #4742: Adapt Code Introspection, Autocompletion and Linting editor functionalities to comply with MS/VSCode Language Server specification
We should write a full-compliant [Microsoft/language-server-protocol](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md)  client. This should enable us to support editing capabilities such as Linting and Autocompletion on Python and other languages only by installing the correct server. 

This reform also aims to simplify the current editor client-server architecture by only using a single client/server that performs all editing and code introspection tasks.

andfoy:
Fixes #4742 

On this PR we will adapt editor code introspection and linting functionality to comply with [Microsoft/language-server-protocol](https://microsoft.github.io/language-server-protocol/specification) specifications. During this PR, the following tasks must be accomplished:

- [x] Write a simple TCP/ZMQ transport proxy that communicates between an LSP server and our client.
- [x] Write a generic TCP v3.0 Language Server Protocol Client (lsp-client) that communicates with any given language server
- [x] Improve ZMQ interprocess communication
- [x] ~Start specific language LSP server when a file of that language is first opened~
- [x] ~Close specific language LSP server when there aren't any files of that language opened~ 
- [x] Integrate lsp-client to our existing introspection calls
- [x] Integrate lsp-client to handle linting functionality
- [x] ~Handle hover calls~
- [x] ~Integrate lsp-client to perform symbol searching~
- [x] Integrate lsp-client to retrieve function definitions and documentation
- [x] Integrate lsp-client to handle go-to-definition requests
- [x] ~Integrate lsp-client to handle code formatting~
- [x] Add file save and file close calls
- [x] Improve teardown and application exit closing behaviour
- [x] Add a settings panel to configure and add different Language-servers
- [x] ~Add client/transport tests~

**Note:** Crossed out features of this list are to be implemented/discussed on subsequent PRs, please see: https://github.com/spyder-ide/spyder/pull/4751#issuecomment-333618640


spyder/utils/code_analysis/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+
+# Language server communication verbosity at server logs.
+TRACE = 'messages'
+if DEV:
+    TRACE = 'verbose'
+
+
+# Spyder editor capabilities
+EDITOR_CAPABILITES = {
+    "workspace": {
```
andfoy: Do the editor supports all the following functionalities?

ccordoba12: I don't know what these terms refer to. Could you add a comment above each one to understand what they mean?

spyder/utils/code_analysis/client.py
```diff
@@ -0,0 +1,225 @@
...
+else:
+    import pathlib
+
+
+TIMEOUT = 5000
+PID = os.getpid()
+
+
+parser = argparse.ArgumentParser(
+    description='ZMQ Python-based MS Language-Server v3.0 client for Spyder')
```
ccordoba12: So do you plan to use zmq sockets? Why not regular sockets?

andfoy: Well I saw that the current implementation is based on ZMQ, but we could use also dedicated sockets per each editor connection, what do you prefer @ccordoba12 @goanpeca?

andfoy: Shall we continue using zmq?

goanpeca: I have no opinion, if using zmq makes the core easier to read and write, then why not?

ccordoba12: Yes, please do. I think Zmq sockets are more robust than plain sockets.

andfoy: ZMQ will be then!

spyder/utils/code_analysis/__init__.py
```diff
@@ -0,0 +1,224 @@
...
+
+TRACE = 'messages'
+if DEV:
+    TRACE = 'verbose'
+
+
+# WorkspaceClientCapabilities define capabilities the
+# editor / tool provides on the workspace
+
+WORKSPACE_CAPABILITIES = {
```
andfoy: @ccordoba12 Are the capabilities descriptions more clear than the last time?

spyder/utils/code_analysis/client.py
```diff
@@ -218,16 +226,45 @@ def __send_request(self, request):
...
-    except KeyboardInterrupt:
-        client.stop()
-        sys.exit(0)
+        while True:
+            client.listen()
+    except TerminateSignal:
+        pass
+    client.stop()
+    sig_manager.restore()
+    os.kill(os.getpid(), signal.SIGTERM)
```
ccordoba12: This doesn't work on Windows, i.e. processes can't be killed this way there. If you want a cross-platform way of doing this, please take a look at `psutil.kill`.

spyder/utils/code_analysis/__init__.py
```diff
@@ -0,0 +1,224 @@
...
+        # Code hints can be turned on/off dynamically.
+        "dynamicRegistration": True
+    },
+
+    # Editor can display additional commands/statistics per each line.
+    # The code lens request is sent from the client to the server to compute
+    # code lenses for a given text document.
+    # A code lens represents a command that should be shown along with
+    # source text, like the number of references, a way to run tests, etc.
+    "codeLens": {
```
andfoy: I think Spyder doesn't support code lens, does it?

goanpeca: ehhh not sure I understand what it does?

andfoy: For example, showing who was the last one that modified a line

andfoy:
### Plugin configuration page preview:

peek 27-08-2017 17-05

goanpeca:
@andfoy why do we need to expose such fined grained details yo the user via the UI?

andfoy:
Because LSP aims to be a generic framework to handle any language, an use may replace or add new servers to interact with Spyder, that includes the possibility of communicating with remote instances.

andfoy:
@ccordoba12 @goanpeca I'm proud to present the first LSP migration milestone: Linting notifications!

linter

andfoy:
It seems that highlighting regions are working only with some fonts:

captura de pantalla de 2017-09-20 18-26-11

captura de pantalla de 2017-09-20 18-26-46

Any ideas @goanpeca @ccordoba12 ?

ccordoba12:
No idea, but this is a terrific achievement! Nice job @andfoy!

andfoy:
I've decided to highlight the warnings by changing the background color instead of underlining the lines.

captura de pantalla de 2017-09-20 19-10-19

ccordoba12:
That's really nice! But I think it should only be shown only when the mouse is over the error mark.

Also, you need to use the new decorations stuff introduced by @rlaverde in PR #5203.

goanpeca:
This is looking pretty neat @andfoy whats next on the plate? Completions?

andfoy:
> But I think it should only be shown only when the mouse is over the error mark.

I think we should display the warning regions at all times, this guarantees that all errors and warnings are taken into account and corrected eventually. I don't know if this is the default behaviour for most editors, but at least on Sublime this is true:

captura de pantalla de 2017-09-20 19-52-21

spyder/utils/code_analysis/__init__.py
rlaverde: No, although It'll be a great addition, (and integration with git), I'll look forward to implement this after you finish this 😄

ccordoba12:
@andfoy, Sublime doesn't have as many decorations as we do (line, cell and folding). So please use a transient decoration instead, to not visually bother the user.

rlaverde:
>This is actually due to the difference between the font employed by TextDecoration and the font defined by the user.

Yes, It's also failing in master when some fonts, thanks for noticing it, I think that setting the user defined font to the `TextDecoration` will be enough, I'll make a PR with that fix.

>That and also snippet suggestions

Language server can provide snippets? or only the completions? how that will interact with #4289 ?

PD: I just take a look and it looks pretty great. 🎉

andfoy:
> Language server can provide snippets? or only the completions? how that will interact with #4289 ?

@rlaverde It seems that LSP servers can provide snippets and completion hints. I would take a more careful look into this to see if we can use the solution provided by PR #4289

```typescript
/**
 * Defines whether the insert text in a completion item should be interpreted as
 * plain text or a snippet.
 */
namespace InsertTextFormat {
	/**
	 * The primary text to be inserted is treated as a plain string.
	 */
	export const PlainText = 1;

	/**
	 * The primary text to be inserted is treated as a snippet.
	 *
	 * A snippet can define tab stops and placeholders with `$1`, `$2`
	 * and `${3:foo}`. `$0` defines the final tab stop, it defaults to
	 * the end of the snippet. Placeholders with equal identifiers are linked,
	 * that is typing in one will update others too.
	 *
	 * See also: https://github.com/Microsoft/vscode/blob/master/src/vs/editor/contrib/snippet/common/snippet.md
	 */
	export const Snippet = 2;
}
```

andfoy:
@ccordoba12 Is this the expected warning selection highlighting behaviour?

highlight

goanpeca:
Looks nice but it seems the selection stays forever ? or there is a timeout ?

rlaverde:
>Looks nice but it seems the selection stays forever ? or there is a timeout ?

Maybe using the `leaveFunction` in the `LineNumeberArea` to remove the decoration.

andfoy:
@goanpeca It disappears until another warning is selected, or if the document is updated.

goanpeca:
>t disappears until another warning is selected, or if the document is updated.

Maybe there should be timeout?

goanpeca:
:-|, don't know 2 seconds?

andfoy:
@goanpeca @ccordoba12 I present the second milestone of the LSP migration: Code autocompletion

completion

andfoy:
I've just added a tooltip to display the related documentation each time a completion object is selected: 
tooltip

goanpeca:
@andfoy that is pretty cool. I like the idea, but I think we can improve the UI to display this in a more compact nicer way, but do not worry about that too much :-)

andfoy:
@goanpeca I will leave it as it was presented on the previous GIF, but I agree that a better UI solution can be implemented, however let's leave that for another PR

andfoy:
@goanpeca @ccordoba12 Yet another milestone: Signature hints

signature


**Note:** I don't know exactly why the title is being splitted

andfoy:
@ccordoba12 @goanpeca @rlaverde What do you think about our new hovering feature:

hover

rlaverde:
>What do you think about our new hovering feature

I like it, my only concern is that it could be implemented with an `EditorExtension` instead of adding more code to the `CodeEditor`

andfoy:
> it could be implemented with an EditorExtension

How can I implement it as an EditorExtension?

rlaverde:
>How can I implement it as an EditorExtension?
- Create a new Editor Extension in `sourcecode/extensions` 
- Move the `request_signature `and `process_signatures` to it (changing references to `self` for `self.editor`) 
- Create a method `_on_key_pressed` and connect it to `CodeEditor.sig_key_pressed`, and move the logic that you put into `KeyPressEvent`, and `handle_parentheses` to it
- Add the extension to the `CodeEditor.editor_extensions`

You could take a look to https://github.com/spyder-ide/spyder/blob/master/spyder/widgets/sourcecode/extensions/closequotes.py

You could use `on_state_changed()` or `on_install()` to connect `sig_key_pressed` (I don't know if this option could be activated/desactivated)

andfoy:
After an initial discussion with @ccordoba12, we've decided to delay/disable some features (to be) introduced by this PR:
* Hover features
* Documentation tooltips during auto-completion
* Find all symbol references

All the previous UI features and others that are available as part of the LSP protocol shall be discussed before being implemented.

ccordoba12:
@rlaverde is going to start to review this one. Good luck!

andfoy:
Here it is: Go To Definition
gotod

goanpeca:
@andfoy yeiii! whats missing now :-p ?

ccordoba12:
It's only missing clean up and adapting our old tests to run with LSP.

rlaverde:
@andfoy I'm trying to test it locally and It isn't working for me, could  you review that there isn't some locals configurations that you haven't committed?

This error happens at startup, and spyder freezes when writing anything that fire a completion call:
```
  File "/home/rlaverde/code/spyder/spyder/plugins/editor.py", line 615, in report_open_file
    language.lower(), self.lsp_editor_settings[language.lower()])
KeyError: 'python'
```

andfoy:
@rlaverde Right now, the settings are defined to use an external instance of the server, please check out your configuration at the LSP configuration tab.

goanpeca:
@andfoy any chance of trying to use lua as another POC ?

andfoy:
@goanpeca Yes, it should be supported by the Lua LSP server https://github.com/Alloyed/lua-lsp

goanpeca:
@andfoy ok, we need to get this plugins going as conda packages!

andfoy:
@ccordoba12 @goanpeca I think we're only missing general code cleanup and fixing the tests

ccordoba12:
Great! I'll add `pyls` to Travis and Appveyor this weekend so you can start fixing our tests.

andfoy:
@ccordoba12 I'm ready to start with test adaptation

andfoy:
@ccordoba12 I think I'll need some help to adapt the tests

andfoy:
Note to self: Test and fix incompabilities on Windows

ccordoba12:
@andfoy, it's great to see that our tests are finally passing in Travis and also in Appveyor for Python 2, Cheers for that!!

However, they are taking now 2 minutes more to finish than before. That's probably because `pyls` is started on every test but only used on very few.

That's why I added the `SPY_TEST_USE_INTROSPECTION` env var, i.e. to make our tests use introspection only when that var is `True`. So please use it to decide when to start or `pyls` in our tests.

ccordoba12:
Also, that could be causing the hangs in Appveyor and Python 3.6. I mean, what I noticed with our previous architecture is that our introspection servers consume a lot of memory on their own and they are not killed until our entire test suite finishes.

With 3 servers started per test in `test_mainwindow.py` and `test_editor`, that was causing a lot of problems in our CIs.

ccordoba12:
Also, please revert the blank removals you introduced in several files (e.g. `app/mainwindow.py` and `widgets/mixins.py`). Those would make my life really hard when merging with `split-plugins`.

andfoy:
@ccordoba12 I'm having random stalled tests on Python 3.6 on Appveyor, also, the ``NO_CONDA`` setup is failling on Travis, although if you debug manually via SSH, it will work without any problem. Any ideas?

andfoy:
> That's why I added the SPY_TEST_USE_INTROSPECTION env var, i.e. to make our tests use introspection only when that var is True. So please use it to decide when to start or pyls in our tests.

Right now, an instance of pyls is launched each time a mainwindow is created, shall I override that behaviour?

ccordoba12:
> Right now, an instance of pyls is launched each time a mainwindow is created, shall I override that behaviour?

Yes, you need to launch `pyls` *only* when `os.environ.get('SPY_TEST_USE_INTROSPECTION') = True`. That env var is set when a test is marked with `@pytest.mark.use_introspection`.

andfoy:
> Yes, you need to launch pyls only when os.environ.get('SPY_TEST_USE_INTROSPECTION') = True. That env var is set when a test is marked with @pytest.mark.use_introspection.

But the env var should be checked directly on ``mainwindow.py``?

andfoy:
@ccordoba12 Is ``CI`` set when running on a CI?

ccordoba12:
> Is CI set when running on a CI?

Yes, on Travis and Appveyor. But not by us, by them.

ccordoba12:
@andfoy, please skip all `pyls` tests on Windows and Python 3.6 to see if that stops Appveyor from hanging.

ccordoba12:
@andfoy, this is terrific! Finally all is green, yay!!

However, this PR will add almost 3400 lines! What modules or files do you want us to review in depth?

andfoy:
> However, this PR will add almost 3400 lines! What modules or files do you want us to review in depth?

It would be good to review the changes done on codeeditor, as they represent the final achievement of this PR. Also, it would be nice to do some kind of beta testing for the feature replacement on a real scenario, such that we can test the actual performance of the client.

andfoy:
@ccordoba12 Could you please give this PR a try?

andfoy:
@ccordoba12 I've added an option to set LSP options as a JSON object on the LSP configuration panel. With this addition, the PR should be complete.

captura de pantalla de 2018-06-29 19-47-23

ccordoba12:
Tasks before merging this PR:

1. Revert blank removed spaces.
2. Remove unwanted dependencies (`coloredlogs`).
3. Put lsp logs in Spyder config dir.
4. Ask to pyls devs why Rope is not taking precedence over Jedi when Rope is faster.
5. Fix Rope completions on untitled files in pyls (deactivate Rope for now in Spyder).

ccordoba12:
Also:

6. Deactivate every plugin except Jedi and pyflakes.

ccordoba12:
@andfoy, the error in Circle seems valid, unless you prefer to skip testing `spyder/utils/code_analysis/lsp_client.py`.

PR #4758, opened 2017-07-18
PR: Add an option to set cursor blinking speed in miliseconds

### LINKED ISSUE #4277: Option to Disable Blinking Cursor
Spyder does not respect system keyboard settings, and there seems to be no way to disable the blinking cursor in the editor, integrated consoles, etc. All modern IDEs provide an option to disable cursor blinking; the flashing cursor is known to irritate a lot of developers. (www.jurta.org/en/prog/noblink).

How can one tweak the Spyder settings, configuration files to make this happen?

I'm using: Linux (Ubuntu 16.04), Anaconda 4.3. This issue was also noted by a few developer friends on Windows.

jnsebgosselin:
Fixes #4277

caret_blinking


goanpeca:
@jnsebgosselin thanks for the contribution, you need to align the two spinboxes (the one from custom margins from panes) Look at the code that does this for the fonts on upper part of your screenshot

goanpeca:
@jnsebgosselin also you should make the PR against 3.x I think @ccordoba12 ?

jnsebgosselin:
Preview with aligned custom margins and caret blinking spinboxes and slabels:

caret_blinking2

goanpeca:
Looks good, except for the branch @ccordoba12 ?

ccordoba12:
@jnsebgosselin, thanks a lot for your contributions! However, we need to ask you to move all your branches from the 3.1.x branch (which is not going to receive more updates) to the 3.x branch (which is our maintenance branch right now).

Sorry for not being clear enough since your first contribution, but you can find the instructions to do that here

https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md#spyder-branches

jnsebgosselin:
@ccordoba12 kk good. I finally managed to rebase this PR. I hope it is not too messy. I can create a new clean PR if required.

Thanks for the link, I'll read it carefully and I will try to improve my setup.

spyder/app/mainwindow.py
```diff
@@ -2539,6 +2539,11 @@ def apply_settings(self):
 
         self.apply_panes_settings()
         self.apply_statusbar_settings()
+        
+        if CONF.get('main', 'use_custom_caret_blinking') is True:
```
ccordoba12: There's no need for `is True` at the end of this line, so please remove it.

spyder/config/main.py
```diff
@@ -93,6 +93,8 @@
               'cpu_usage/timeout': 2000,
               'use_custom_margin': True,
               'custom_margin': 0,
+              'use_custom_caret_blinking': False,
+              'custom_caret_blinking': 1000,
```
ccordoba12: Instead of using `caret` as part of these options, please use `cursor` instead, i.e. `use_custom_cursor_blinking` and `custom_cursor_blinking`.

ccordoba12:
Changing this option requires a Spyder restart because it's applied at the window level.

@rlaverde, please instruct @jnsebgosselin on how to do that.

rlaverde:
@jnsebgosselin Thanks, for you contribution, in order to make this an resting option, you need to add restart support to `create_spinbox()` and `create_checkbox`, and in def `load_from_conf()` add the labels to the `restart_options` array. And use it when calling the creating of the checkbox and he spinbox

You can take a look at this commit https://github.com/spyder-ide/spyder/pull/4454/commits/8fafe45279d0b679f2986fc9ec49961cb1c8443c

jnsebgosselin:
@rlaverde @ccordoba12 Thank you. But are you sure this needs a restart? Seems to me like this can be applied at run-time. At least, it is on Windows.

That being said, something was not working as expected with the code. When the option to use a custom cursor blinking time was unchecked, after a custom value had been set, the value was not reset to the OS default. `QApplication.cursorFlashTime()` actually returns the value for the current application instance, not the default value set for the OS. So basically, nothing was being changed.

I haven't found an easy way to get the OS default value for the cursor flash time. This seems to be very platform dependent. Unless you guys know of a better way, maybe we could cache the value of `QApplication.cursorFlashTime()` at startup?This way, no restart would be needed for this setting, which would be better I think.  I've made the changes to the PR so you can look it up already.

spyder/config/main.py
```diff
@@ -93,6 +93,8 @@
               'cpu_usage/timeout': 2000,
               'use_custom_margin': True,
               'custom_margin': 0,
+              'use_custom_cursor_blinking': False,
+              'custom_cursor_blinking': 1000,
```
goanpeca: But this value should be used then `'custom_cursor_blinking': CURSORBLINK_OSDEFAULT`, no?

jnsebgosselin: You are right. This makes sense to set the default value here as the default value for the OS. I will change it.

jnsebgosselin: @goanpeca in the header of this file (`main.py`), it is written:
> Note: Leave this file free of Qt related imports [...]

So I don't know how I can link `CURSORBLINK_OSDEFAULT`, whose value must be obtained with QApplication... Do you have any idea? Maybe another solution could be to set the default value when creating the spinbox within the `configdialog.py` module instead of here? Something like:
```
cursor_spin = self.create_spinbox( 
        "", _("ms"), 'custom_cursor_blinking',
        default = QApplication.cursorFlashTime(),
        min_ = 0, max_ = 5000)
```

goanpeca:
@ccordoba12 ?

spyder/config/main.py
ccordoba12: @jnsebgosselin, your solution is the only one that occurs to me too.

Since we can't adjust this value in our config system, please remove it from here. It'll be written to `spyder.ini` when the spinbox is created.

spyder/plugins/configdialog.py
```diff
@@ -887,7 +889,8 @@ def setup_page(self):
         cursor_box = newcb(_("Cursor blinking:"),
                            'use_custom_cursor_blinking')
         cursor_spin = self.create_spinbox("", _("ms"), 'custom_cursor_blinking',
-                                          0, 0, 5000)
+                                          default = QApplication.cursorFlashTime(),
+                                          min_ = 0, max_ = 5000, step = 10)
```
ccordoba12: Please remove spaces around `=`. They are not needed for keyword args.

spyder/plugins/configdialog.py
```diff
@@ -328,6 +329,7 @@ def load_from_conf(self):
             if lineedit.restart_required:
                 self.restart_options[option] = lineedit.label_text
         for spinbox, (option, default) in list(self.spinboxes.items()):
+            print(option, default, self.get_option(option, default))
```
ccordoba12: Please remove this `print` here.

spyder/config/main.py
```diff
@@ -94,7 +94,6 @@
               'use_custom_margin': True,
               'custom_margin': 0,
               'use_custom_cursor_blinking': False,
-              'custom_cursor_blinking': 1000,
```
goanpeca: @ccordoba12 and @jnsebgosselin I think we still need to have the option here, the only difference is that it would be `None` by default, and when building the spinbox, we would check if it is None and update with the default by QApplication, or just use the stored value.

jnsebgosselin: If I do that, then delete my .init files and restart Spyder, I get the following traceback:
```
  File "C:\Users\jsgosselin\spyder\spyder\app\mainwindow.py", line 2585, in edit_preferences
    widget.initialize()
  File "C:\Users\jsgosselin\spyder\spyder\plugins\configdialog.py", line 76, in initialize
    self.load_from_conf()
  File "C:\Users\jsgosselin\spyder\spyder\plugins\configdialog.py", line 332, in load_from_conf
    spinbox.setValue(self.get_option(option, default))
TypeError: setValue(self, int): argument 1 has unexpected type 'NoneType'
```

ccordoba12: @goanpeca, there is no need to do that because values here are placed here only if we plan to change/update them in the future.

ccordoba12: But this value is OS dependent, so we can't put it here.

goanpeca: Ok

andfoy:
@ccordoba12 This is working as expected, both on Windows and Linux, without any Spyder reboot required

ccordoba12:
Thanks a lot for testing @andfoy! Merging then ;-)

PR #4764, opened 2017-07-18
PR: Fix tab does not switch when ctrl is released issue

### LINKED ISSUE #4763: In the Editor, tab does not switch when ctrl is released and focus remains on the tab_switcher

1. While focus is in the Editor, hold ctrl key and click Tab.
2. Release ctrl key

_Expected output:_ When ctrl key is released, Editor tab should change to the one selected and tab_switcher should close.

_Actual output:_ When ctrl key is released, Editor tab does not change and focus remains on the tab_switcher. A mouse double-click on the tab_switcher is required for it to switch tab and to close.

* Spyder Version: 3.2.0.dev0
* Python Version: 3.6.1
* Qt Version: 4.8.7
* PyQt Version: 4.11.4
* Operating system: Win7 64bits

## Dependencies

tab_switcher

jnsebgosselin:
Fixes #4763

----

It seems like the ctrl key release event never reach the [EditorStack ](https://github.com/spyder-ide/spyder/blob/master/spyder/widgets/editor.py). It seems to be swallowed by the [CodeEditor](https://github.com/spyder-ide/spyder/blob/master/spyder/widgets/sourcecode/codeeditor.py).

The proposed solution to solve this issue it to install an eventFilter on each `editor` added to
the `EditorStack` and to move the code that is in the `keyReleaseEvent` in the `eventFilter` method.

There may be a more elegant way to solve this directly in the CodeEditor class. I can look into it if this approach is preferred.

### DIFF
--- a/spyder/widgets/editor.py
+++ b/spyder/widgets/editor.py
@@ -22,7 +22,7 @@ from collections import MutableSequence
 from qtpy import is_pyqt46
 from qtpy.compat import getsavefilename
 from qtpy.QtCore import (QByteArray, QFileInfo, QObject, QPoint, QSize, Qt,
-                         QThread, QTimer, Signal, Slot, QEvent)
+                         QThread, QTimer, Signal, Slot)
 from qtpy.QtGui import QFont
 from qtpy.QtWidgets import (QAction, QApplication, QHBoxLayout, QMainWindow,
                             QMessageBox, QMenu, QSplitter, QVBoxLayout,
@@ -1896,7 +1896,6 @@ class EditorStack(QWidget):
         editor.get_completions.connect(introspector.get_completions)
         editor.sig_show_object_info.connect(introspector.show_object_info)
         editor.go_to_definition.connect(introspector.go_to_definition)
-        editor.installEventFilter(self)
 
         finfo = FileInfo(fname, enc, editor, new, self.threadmanager,
                          self.introspector)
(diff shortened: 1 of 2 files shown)


jnsebgosselin:
Ok, I've found another way to solve this issue that is better I think. We can add event.ignore() in keyReleaseEvent method of CodeEditor class so that the event is propagated to the parent. I think it is a better approach than using an eventFilter in the parent. 

The PR was modified accordingly.

ccordoba12:

spyder/widgets/editor.py
```diff
@@ -2185,7 +2185,7 @@ def keyReleaseEvent(self, event):
                         self.tabs_switcher = None
                         return
 
-        super(EditorStack, self).keyPressEvent(event)
+        super(EditorStack, self).keyReleaseEvent(event)
```
rlaverde: My fault, I introduced this typo, sorry 😞

spyder/widgets/sourcecode/codeeditor.py
```diff
@@ -2746,6 +2746,7 @@ def keyReleaseEvent(self, event):
         """Override Qt method."""
         self.timer_syntax_highlight.start()
         super(CodeEditor, self).keyReleaseEvent(event)
+        event.ignore()
```
rlaverde: This was the error, In qt4 the event wasn't propagated to the parent (I had tested it with pyqt5)

PR #4799, opened 2017-07-24
PR: Delete scientific_startup.py

### LINKED ISSUE #4797: Remove scientific_startup.py
I forgot to do this when I removed the Python console.

@dalthviz, this is an easy one for you. There are also a couple of references about scientific_startup.py that you need to remove.

dalthviz:
Fixes #4797


create_exe.py
```diff
@@ -37,8 +37,7 @@ def create_executable():
         dist.add_modules('guiqwt')
     except ImportError:
         pass
-    dist.includes += ['spyder.scientific_startup',
-                      'spyder.widgets.externalshell.sitecustomize']
+    dist.includes += ['spyder.widgets.externalshell.sitecustomize']
```
ccordoba12: This file is now `spyder.utils.site.sitecustomize`, so please change it too.

PR #4813, opened 2017-07-25
PR: Make highlighting colors in the IPython Console match those used in the Spyder Editor

### LINKED ISSUE #4806: Change color of matched parens in the IPython console
Hi, is there a way to change the color of matched parenthesis automatic selection in the IPython console? 
I'm using a dark theme, and the white selection that the console is using along with the almost white text that my theme has is making me very difficult to recognize where my cursor is when I'm calling a procedure. Here you can see some screenshots: 
screenshot from 2017-07-24 17-32-50
screenshot from 2017-07-24 17-33-15
It would be nice if the new feature of color scheme matching in the ipython console could cover matched parenthesis and selection text as well.
If there's a way by editing some config file, please let me know. It's kinda frustrating to guess if my cursor is already inside parenthesis or outside.
Or if I could disable matched parens in the ipython console would be nice too.
I'm using:
Spyder 3.2.0
IPython 6.1.0
Python 3.5.2
Ubuntu 16.04

jnsebgosselin:
Fixes #4806

----

- Changes in `style.py` is to let the text highlighting color in the IPython console to be the same as for the Spyder main application instead of using the light gray color that was hard-coded.
- Changes in `shell.py` is to set the _matched-bracket_ highlighting color to the color defined in Spyder
preferences instead of using the hard coded light gray color.

### DIFF
--- a/spyder/utils/ipython/style.py
+++ b/spyder/utils/ipython/style.py
@@ -49,7 +49,6 @@ def create_qss_style(color_scheme):
         in_prompt_color = 'lime'
         out_prompt_color = 'red'
     background_color = color_scheme['background']
-    selection_background_color = '#ccc'
     error_color = 'red'
     in_prompt_number_font_weight = 'bold'
     out_prompt_number_font_weight = 'bold'
--- a/spyder/utils/ipython/style.py
+++ b/spyder/utils/ipython/style.py
@@ -59,7 +59,6 @@ def create_qss_style(color_scheme):
     sheet = """QPlainTextEdit, QTextEdit, ControlWidget {{
                                           color: {} ;
                                           background-color: {};
-                                          selection-background-color: {}
                                          }}
               .error {{ color: {}; }}
               .in-prompt {{ color: {}; }}
@@ -70,7 +69,6 @@ def create_qss_style(color_scheme):
               """
 
     sheet_formatted = sheet.format(font_color, background_color,
-                                   selection_background_color,
                                    error_color,
                                    in_prompt_color, in_prompt_color,
                                    in_prompt_number_font_weight,
--- a/spyder/widgets/ipythonconsole/shell.py
+++ b/spyder/widgets/ipythonconsole/shell.py
@@ -12,11 +12,13 @@ import ast
 import uuid
 
 from qtpy.QtCore import Signal
+from qtpy.QtGui import QColor
 from qtpy.QtWidgets import QMessageBox
 from spyder.config.base import _
 from spyder.config.gui import config_shortcut
 from spyder.py3compat import to_text_string
 from spyder.utils import programs
+from spyder.utils import syntaxhighlighters as sh
 from spyder.utils.ipython.style import create_qss_style, create_style_class
 from spyder.widgets.ipythonconsole import (ControlWidget, DebuggingWidget,
                                            HelpWidget, NamepaceBrowserWidget,
@@ -70,6 +72,9 @@ class ShellWidget(NamepaceBrowserWidget, HelpWidget, DebuggingWidget):
         # To save kernel replies in silent execution
         self._kernel_reply = None
 
+        color_scheme = kw['config']['JupyterWidget']['syntax_style']
+        self.set_bracket_matcher_color_scheme(color_scheme)
+                
     #---- Public API ----------------------------------------------------------
     def set_exit_callback(self):
         """Set exit callback for this shell."""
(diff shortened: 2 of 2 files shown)


ccordoba12:
@jnsebgosselin, thanks a lot for your quick PR!

However, this PR needs to be against `3.x` since it's a bugfix and doesn't add new features. So, please move it to that branch.

For next time, I'd like to let you know that we decide which branch to use for each issue by using the *Milestone* tag (located to the right of every issue/PR). So I tagged issue #4806 with the *3.2.1* milestone.

jnsebgosselin:
@ccordoba12  ahhh I see, the Milestone tag, got it. Will check it up next time.

ccordoba12:
Thanks for understanding!

rlaverde:
You need to rebase your branch onto `3.x`

https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md#changing-base-branch

jnsebgosselin:
@rlaverde  hehe this is what I'm trying to do but I am not able to push the rebase to the remote. It is rejected with the following error and I'm stuck, trying to find a solution on the world wide web :) Any tips?

untitled

rlaverde:
@jnsebgosselin add `-f` to the push command

`git push origin ipython_console_highlight_settings -f`

jnsebgosselin:
@rlaverde  Oh yeah! Many thanks

goanpeca:
@jnsebgosselin just bear in mind, that -f, will force push stuff and overwrite the repo, so use carefully :-) (when rebasing)

ccordoba12:
@dalthviz, please review this one since you were in charge of adding colors to the IPython console.

spyder/widgets/ipythonconsole/shell.py
```diff
@@ -110,9 +114,15 @@ def get_cwd(self):
             return
         else:
             self.silent_exec_method(code)
+            
+    def set_bracket_matcher_color_scheme(self, color_scheme):
```
dalthviz: Please add a docstring here. Something like `"""Set color scheme for matched parentheses."""`

spyder/widgets/ipythonconsole/shell.py
```diff
@@ -70,6 +71,9 @@ def __init__(self, ipyclient, additional_options, interpreter_versions,
         # To save kernel replies in silent execution
         self._kernel_reply = None
 
+        color_scheme = kw['config']['JupyterWidget']['syntax_style']
+        self.set_bracket_matcher_color_scheme(color_scheme)
```
dalthviz: Maybe this two lines could be replaced by something like
```python
self.set_bracket_matcher_color_scheme(self.syntax_style)
```
What do you guys think @jnsebgosselin @ccordoba12 ? 

Also, please add a comment above, something like `# To set the color of the matched parentheses. `

jnsebgosselin: Yes you are right and this works. I didn't see it because it is not obvious where this is inherited.

spyder/utils/ipython/style.py
```diff
@@ -49,7 +49,6 @@ def give_font_style(is_italic):
         in_prompt_color = 'lime'
         out_prompt_color = 'red'
     background_color = color_scheme['background']
-    selection_background_color = '#ccc'
```
ccordoba12: Why do you have to remove this as part of this PR?

ccordoba12: Sorry, I just saw you mentioned the reason in your description :-)

spyder/widgets/ipythonconsole/shell.py
```diff
@@ -70,6 +71,9 @@ def __init__(self, ipyclient, additional_options, interpreter_versions,
         # To save kernel replies in silent execution
         self._kernel_reply = None
 
+        # To set the color of the matched parentheses
+        self.set_bracket_matcher_color_scheme(self.syntax_style)
```
ccordoba12: Why is it necessary to add this here? I mean, the call to `set_bracket_matcher_color_scheme` in `set_color_scheme` should be enough, shouldn't it?

jnsebgosselin: `set_color_scheme` does not seem to be called anywhere on startup. The initial settings are actually passed to the `JupytherWidget` as argument in the `super` call and the settings are applied directly within the `JupytherWidget`. However, these settings do not include the _matched parentheses color_, since the default hard-coded value is used. That is why we have to set it separately here in the `__init__` on our side.

So if we do not add this there, the setting to the  _matched parentheses color_ is applied only when there is actually a change to the color scheme after the Spyder is started and a call is made to `set_bracket_matcher_color_scheme`.

ccordoba12: Ok, got it.

Could you improve your comment above this line with a short summary of what you just said? Else, I'm sure me or someone else will be tempted to remove this line in the future ;-)

jnsebgosselin: Of course, good idea.

jnsebgosselin: I'm not sure it is ok. I've tried to keep it as short as possible. What do you guys think?

```
# Set the color of the matched parentheses here since the qtconsole
# uses a hard-coded value that is not modified when the color scheme is
# set in the qtconsole constructor.
```

dalthviz: I think its ok 👍 maybe you could add a reference to the issue, something like `See issue 4806`. Also, thank you for your help 😄, nice job!

ccordoba12: Yep, that text is descriptive enough, thanks!

ccordoba12:
Thanks @jnsebgosselin, great work!

PR #4820, opened 2017-07-26
PR: Add a button to Find in Files to make searches case sensitive

### LINKED ISSUE #4819: Find in Files should have ignore case flag
Similar to the editor search dialog, Find in Files should have an option to ignore case when searching. Like the editor search dialog, the button should be located to the right of the regex toggle.

This would be more very useful and make the UI more consistent.

andfoy:
Fixes #4819 

## Preview
captura de pantalla de 2017-07-26 16-23-39


spyder/widgets/tests/data/ham.erl
```diff
@@ -0,0 +1,10 @@
+ham(N) -> ham(N, 1, 1).
```
ccordoba12: Adding a new file type requires changing `setup.py` and `MANIFEST.in` because we package our tests and all their supporting files as part of Spyder.

So please use a file extension that we already support, such as `.txt`.

spyder/plugins/findinfiles.py
```diff
@@ -52,11 +52,13 @@ def __init__(self, parent=None):
         exclude_regexp = self.get_option('exclude_regexp')
         in_python_path = self.get_option('in_python_path')
         more_options = self.get_option('more_options')
+        case_sensitive = self.get_option('case_sensitive', True)
```
ccordoba12: Please don't use a default value here. Instead, add `case_sensitive` as an option of the `find_in_files` section in `config/main.py`.

ccordoba12:
This looks really good to me. I only left a couple of minor comments :-)

PR #4822, opened 2017-07-27
PR: Extend Find in Files to select other search directories

### LINKED ISSUE #4814: Extend Find in Files to select other directories
As suggested by @jnsebgosselin and @avylove on #4809

andfoy:
Fixes #4814 

## Preview
peek 27-07-2017 17-05


andfoy:
@ccordoba12 I have some problems choosing a directory using ``qtpy.compat.getexistingdirectory``. When the folder selection window is closed, it keeps reappearing instead of closing, this happens for both Cancel and Accept options. After that, Spyder cannot be closed unless is killed with a ``SIGKILL`` signal.

andfoy:
@ccordoba12 Does it look right?

ccordoba12:
I like it, a lot!

@jnsebgosselin, @avylove, what do you think?

ccordoba12:
Questions:

1. How many paths are we saving in the drop-down menu?
2. Are those paths saved among sessions?

ccordoba12:
In point 2. I meant, are those paths saved when Spyder is closed and open again?

andfoy:
On a single session we're storing all paths as the user would like, however, between sessions, we're just storing the 15 most recent entries.

goanpeca:
Looks pretty good @andfoy :+1:

ccordoba12:
> On a single session we're storing all paths as the user would like

That could cause problems. Please limit that to the 15 most recent paths too.

goanpeca:
@andfoy do we have a 'search in current project' option ?

andfoy:
@goanpeca Yes, I forgot to show it on the preview GIF. The option is only enabled when a project is opened

jnsebgosselin:
@andfoy This looks great and very intuitive. I like it. Good job.

avylove:
@andfoy Looks great! Good solution!

ccordoba12:
@andfoy, this requires now a merge with 3.x to fix conflicts.

andfoy:
@ccordoba12 I think this one is ready

PR #4828, opened 2017-07-27
PR: Fix error non ascii current directory directory

### LINKED ISSUE #4816: Error when trying to change to non-ascii directories in the Working Directory toolbar
## Description

1. 
2. 
3. 

  File "C:\Users\len\Anaconda2\lib\site-packages\spyder\plugins\ipythonconsole.py", line 895, in set_working_directory
    refresh_console=False)
  File "C:\Users\len\Anaconda2\lib\site-packages\spyder\plugins\workingdirectory.py", line 286, in chdir
    os.chdir( to_text_string(directory) )
WindowsError: [Error 3] : u'D:\\\u0376\ufffd\ufffd\\\ufffd\ufffd\ufffd\ufffd\ufffd\ufffd'

* Spyder Version: 3.2.0 
* Python Version: 2.7.13

## Dependencies
```
pyflakes >=0.5.0  :  1.5.0 (OK)
pycodestyle >=2.3 :  2.3.1 (OK)
pandas >=0.13.1   :  0.20.1 (OK)
numpy >=1.7       :  1.12.1 (OK)
sphinx >=0.6.6    :  1.6.3 (OK)
rope >=0.9.4      :  0.10.5 (OK)
jedi >=0.9.0      :  0.10.2 (OK)
nbconvert >=4.0   :  5.2.1 (OK)
sympy >=0.7.3     :  1.0 (OK)
cython >=0.21     :  0.25.2 (OK)
qtconsole >=4.2.0 :  4.3.0 (OK)
IPython >=4.0;<6.0:  5.3.0 (OK)
pylint >=0.25     :  1.7.2 (OK)

```

rlaverde:
Fixes #4816

### DIFF
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -345,8 +345,9 @@ def test_change_types_in_varexp(main_window, qtbot):
     assert shell.get_value('a') == 10
 
 
+@pytest.mark.parametrize("test_directory", ["test_dir", "non_ascii_ñ_í_ç"])
 @flaky(max_runs=3)
-def test_change_cwd_ipython_console(main_window, qtbot, tmpdir):
+def test_change_cwd_ipython_console(main_window, qtbot, tmpdir, test_directory):
     """
     Test synchronization with working directory and File Explorer when
     changing cwd in the IPython console.
@@ -356,7 +357,7 @@ def test_change_cwd_ipython_console(main_window, qtbot, tmpdir):
     qtbot.waitUntil(lambda: shell._prompt_html is not None, timeout=SHELL_TIMEOUT)
 
     # Change directory in ipython console using %cd
-    temp_dir = str(tmpdir.mkdir("test_dir"))
+    temp_dir = str(tmpdir.mkdir(test_directory))
     with qtbot.waitSignal(shell.executed):
         shell.execute("%cd {}".format(temp_dir))
     qtbot.wait(1000)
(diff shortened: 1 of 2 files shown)


ccordoba12:
@rlaverde, the solution to this is more complex than I originally thought. I already have a fix in my local fork, so please don't add more commits here.

spyder/widgets/ipythonconsole/shell.py
```diff
@@ -92,7 +94,11 @@ def is_spyder_kernel(self):
 
     def set_cwd(self, dirname):
         """Set shell current working directory."""
-        code = u"get_ipython().kernel.set_cwd(r'{}')".format(dirname)
+        # Replace single for double backslashes on Windows
+        if os.name == 'nt':
+            dirname = dirname.replace(u"\\", u"\\\\")
```
rlaverde: Maybe we could use [os.path.normpath](https://docs.python.org/3.6/library/os.path.html#os.path.normpath) instead

goanpeca: Agreed, this solution is very brittle

ccordoba12: Ok, sure. Its help doesn't mention that it converts `\` to `\\`, but please open a PR for this. If tests pass then I'm for it.

rlaverde: I thought you was removing extra slashes, instead of escaping them, maybe using raw string will do the trick

ccordoba12: I don't know how to combine raw strings with unicode. I did some quick experiments yesterday but I couldn't find how to do it.

PR #4850, opened 2017-07-28
PR: Update only current editor stack when moving tabs.

### LINKED ISSUE #4831: Error whe rearranging tabs and multiple Editors
https://github.com/spyder-ide/spyder/issues/564#issuecomment-318436787

When I have two editors open with Spyder 3.2 (MacOS X version), I can rearrange the tabs in either window. When I do this with say editor 1, the filename tab order will update on both editors, but editor 2 will not update the actual file in that location. So in editor 1 I will change the order to File2, File1, File3, then editor 2 will show the tab order as File2, File1, File3--but the contents of editor 2 are still File1, File2, File3. Also, if I am editing the contents of File2 in editor 2 (labeled File1), File2 will not save until I switch to the tab labeled File2 in either editor--regardless of the file contents shown. However, the file does save correctly. A restart of Spyder fixes the labeling problem, but any rearrangement of the tabs has the same effect.

rlaverde:
Fixes #4831

### DIFF
--- a/spyder/plugins/editor.py
+++ b/spyder/plugins/editor.py
@@ -2681,6 +2681,7 @@ class Editor(SpyderPluginWidget):
         for editorstack in self.editorstacks:
             editorstack.create_new_file_if_empty = value
 
+    @Slot(int, int)
     def move_editorstack_data(self, start, end):
         """Move editorstack.data to be synchronized when tabs are moved."""
         if start < 0 or end < 0:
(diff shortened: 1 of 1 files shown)


ccordoba12:
Shouldn't we try to update all editor stacks? Or perhaps that's confusing...

@goanpeca, what do you think?

goanpeca:
@ccordoba12 I dont really use this feature so I am not sure what users expect :-|

ccordoba12:
Let's merge this for now because the bug is quite critical. I don't think users would expect that both windows should be synced when moving tabs.

rlaverde:
> I don't think users would expect that both windows should be synced when moving tabs.

Personally I'll prefer the both windows to be totally independent, not only the position of the tabs not to be synced, but also different files opened, and when closing a tab to only be closed in that window

goanpeca:
>Personally I'll prefer the both windows to be totally independent, not only the position of the tabs not to be synced, but also different files opened, and when closing a tab to only be closed in that window

Well.. that is not really the purpose of opening two editors... for that just open 2 instances of spyder

ccordoba12:
I agree with @goanpeca. @dalthviz is working on making all plugins to open a new window when detached from the main window, so we'll have to rethink this solution for Spyder 4.

@dalthviz, could you a open an issue about it, so we don't forget to address it? I think what we need to do is syncing tabs in both instances, so that when external windows are closed, tabs have the same order in the main window.

rlaverde:
> Well.. that is not really the purpose of opening two editors... for that just open 2 instances of spyder

Normally I don't want all the files to be opened in two windows, and opening two different instances isn't the same It implies different console, explorer, find infiles.. etc

PR #4861, opened 2017-07-30
PR: Install qtpy and qtconsole from Github when testing

ccordoba12:
This way we will always be sure that Spyder is not broken when we make new releases of those packages.


goanpeca:
But we should test at least one worker under normal circumstances and not with master all the time?

ccordoba12:
You're right. Please see PR #4875.

PR #4864, opened 2017-07-31
PR: Fix Run option in File and Project explorers

### LINKED ISSUE #4769: Exception raised when running file from project explorer
This looks like a simple one but I don't have time now to dig into it.

## Description

1. Right-click on a Python file in the Project explorer
2. Select 'Run' in the context menu

The exception below is raised.

  File "/home/jitse/misc/spyder/spyder/widgets/explorer.py", line 557, in run
    self.parent_widget.run.emit(fname)
AttributeError: 'Projects' object has no attribute 'run'

* Spyder Version: 3.2.0.dev0 76cb1332
* Python Version: 3.6.1

## Dependencies
```
jedi >=0.9.0     :  0.9.0 (OK)
sympy >=0.7.3    :  1.0 (OK)

```

andfoy:
Fixes #4769

### DIFF
--- a/spyder/plugins/explorer.py
+++ b/spyder/plugins/explorer.py
@@ -73,11 +73,6 @@ class Explorer(ExplorerWidget, SpyderPluginMixin):
         self.renamed.connect(self.main.editor.renamed)
         self.main.editor.open_dir.connect(self.chdir)
         self.create_module.connect(self.main.editor.new)
-        self.run.connect(
-                     lambda fname:
-                     self.main.open_external_console(to_text_string(fname),
-                                         osp.dirname(to_text_string(fname)),
-                                         '', False, False, True, '', True))
         # Signal "set_explorer_cwd(QString)" will refresh only the
         # contents of path passed by the signal in explorer:
         self.main.workingdirectory.set_explorer_cwd.connect(
--- a/spyder/plugins/explorer.py
+++ b/spyder/plugins/explorer.py
@@ -77,7 +77,7 @@ class Explorer(ExplorerWidget, SpyderPluginMixin):
                      lambda fname:
                      self.main.open_external_console(to_text_string(fname),
                                          osp.dirname(to_text_string(fname)),
-                                         '', False, False, True, '', False))
+                                         '', False, False, True, '', True))
         # Signal "set_explorer_cwd(QString)" will refresh only the
         # contents of path passed by the signal in explorer:
         self.main.workingdirectory.set_explorer_cwd.connect(
--- a/spyder/plugins/projects.py
+++ b/spyder/plugins/projects.py
@@ -154,9 +154,6 @@ class Projects(ProjectExplorerWidget, SpyderPluginMixin):
         self.sig_project_closed.connect(
             lambda v: self.editor.setup_open_files())
         self.recent_project_menu.aboutToShow.connect(self.setup_menu_actions)
-        self.run.connect(lambda fname: self.main.open_external_console(
-            to_text_string(fname), osp.dirname(to_text_string(fname)),
-            '', False, False, True, '', True))
 
     def refresh_plugin(self):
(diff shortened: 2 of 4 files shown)


spyder/utils/programs.py
```diff
@@ -288,7 +287,10 @@ def run_python_script_in_terminal(fname, wdir, args, interact,
                                    "an external terminal"),
                                  QMessageBox.Ok)
     elif os.name == 'posix':
-        programs = [{'cmd': 'gnome-terminal',
+        programs = [{'cmd': 'x-terminal-emulator',
+                     'wdir-option': '--working-directory',
+                     'execute-option': '-x'},
```
ccordoba12: It seems this change broke one of our tests. Is it really necessary to solve issue #4769?

andfoy: Actually no, I observed that external scripts are opened on a different terminal instead on the system default one. On my case I use Tilix, but it opens gnome-terminal, so I thought it would be better to open the custom one. However I can remove this one

ccordoba12: Yes, please open a new PR for this change.

ccordoba12: I'm not against it, it's just that this PR is to solve issue #4769 only.

jitseniesen: I put the test in if I remember correctly, but it just reflects the code as it was at the time, so feel free to change this (in another PR).

spyder/plugins/projects.py
```diff
@@ -153,6 +154,9 @@ def register_plugin(self):
         self.sig_project_closed.connect(
             lambda v: self.editor.setup_open_files())
         self.recent_project_menu.aboutToShow.connect(self.setup_menu_actions)
+        self.run.connect(lambda fname: self.main.open_external_console(
```
ccordoba12: @andfoy, let's make the file to be run in a dedicated console (instead of in an external one). I think it makes more sense (at least to me).

The problem is external consoles are immediately closed, unless you change the run options for your file.

spyder/plugins/explorer.py
```diff
@@ -77,7 +77,7 @@ def register_plugin(self):
                      lambda fname:
                      self.main.open_external_console(to_text_string(fname),
                                          osp.dirname(to_text_string(fname)),
-                                         '', False, False, True, '', False))
```
ccordoba12: Why did you change this?

spyder/plugins/projects.py
andfoy: I think we don't have an option to run a file on a dedicated console: [app/mainwindow.py#L2423](https://github.com/spyder-ide/spyder/blob/master/spyder/app/mainwindow.py#L2423)

spyder/plugins/explorer.py
andfoy: See above comment :-)

spyder/plugins/projects.py
ccordoba12: Please use `self.main.ipyconsole.create_cleint_for_file` instead. That's the way to run a file in a dedicated console.

spyder/plugins/explorer.py
andfoy: Do I use ``self.main.ipyconsole.create_cleint_for_file`` also for this one?

ccordoba12: Yeah, I think so. I think it's better than running in an external console.

spyder/plugins/projects.py
andfoy: Does ``create_client_for_file`` executes the file? It seems that it opens a new console without executing the file

andfoy: We should use ``run_script`` instead

andfoy:
@ccordoba12 I think this one is ready

PR #4876, opened 2017-08-01
PR: Avoid an error when tests are re-run multiple times in Travis

ccordoba12:


ccordoba12:
This doesn't seem to be necessary after my last changes to `test_runconfig_workdir` but I'll leave it open just in case.

Issue #4839, opened 2017-07-28
Make Spyder remember tab order


gabrielclow:
With Spyder 3.2 we gained the ability to sort tabs on the Editor by dragging and dropping. I love this feature. The problem is that it doesn't remember the order I chose when I close and reopen. This makes this feature less useful because I have to sort it every time I open Spyder.

Thanks in advance.

ccordoba12:
@rlaverde, is this possible?

rlaverde:
I'm not sure, I don't know how this data is saved/loaded when closing/opening spyder, I will try to address it with #4831

ccordoba12:
Please don't. Leave this for another PR.

goanpeca:
>@rlaverde, is this possible?

It should be, but we will need a separate entry on the config for these

rlaverde:
I'm testing and spyder remembers the tab order, files are saved and loaded in the tab order.

@gabrielclow could you give a reproducible example of when the file order changes?

gabrielclow:
1. Open Spyder and have this tab order:
![21](https://user-images.githubusercontent.com/9113853/28848688-4cce5a72-76ea-11e7-88e6-6ae2ce6b9904.PNG)
2. Do nothing, close Spyder, open again and now this tab order:
![22](https://user-images.githubusercontent.com/9113853/28848700-555c9f3c-76ea-11e7-862c-a4a7c7d17787.PNG)
3. Do nothing, close Spyder, open again and once more, another tab order:
![23](https://user-images.githubusercontent.com/9113853/28848715-5ea68cc4-76ea-11e7-925f-17cf074f736e.PNG)

But there is more information: I tested some more and apparently this problem only happens when I have a project open. It looks like the problem does not occur when all projects are closed.

gabrielclow:
1. Open Spyder and have this tab order:
![21](https://user-images.githubusercontent.com/9113853/28848953-39900ee6-76eb-11e7-8e79-bbbd952b9e3e.PNG)
2. Do nothing, close Spyder, open again and now this tab order:
![22](https://user-images.githubusercontent.com/9113853/28848958-3e689212-76eb-11e7-9e3f-b0546ec03b7f.PNG)
3. Do nothing, close Spyder, open again and once more, another tab order:
![23](https://user-images.githubusercontent.com/9113853/28848972-485f1f3e-76eb-11e7-9e97-060dab8138c4.PNG)

[...]

PR #4880, opened 2017-08-02
PR: Preserve order when saving/loading files from a project.

rlaverde:
Fixes: #4839 

When saving loading files from project configuration the order wasn't preserved, `set` doesn't preserve the order (except in python3.6).

Searching I found that using `collections.OrderedDict` is maybe the best way to delete duplicates and preserve order regardless the python interpreter implementation.


andfoy:
I think you can use frozenset also: https://docs.python.org/2/library/stdtypes.html#frozenset

rlaverde:
> I think you can use frozenset

But frozenset doesn't guaranties order neither

PR #4885, opened 2017-08-03
PR: Add a button in the Variable Explorer toolbar to reset namespace

### LINKED ISSUE #313: Variable Explorer: Add a button to remove all variables at once
_From [tim.mich...@gmail.com](https://code.google.com/u/117399810383525443135/) on 2010-08-25T09:05:56Z_

Please add a possibility to remove all variables from the namespace at once (without having the need to select them before).

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=313_

jnsebgosselin:
Fixes #313

----

When clicking on the button, %reset is called in the corresponding
IPython console with `force=True`, i.e. no confirmation is asked from
the user before doing it.

reset_the_namespace

### DIFF
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -144,6 +144,7 @@ DEFAULTS = [
               'use_pager': False,
               'show_calltips': True,
               'ask_before_closing': False,
+              'show_reset_namespace_warning': True,
               'buffer_size': 500,
               'pylab': True,
               'pylab/autoload': False,
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -301,11 +301,10 @@ class IPythonConsoleConfigPage(PluginConfigPage):
         ask_box = newcb(_("Ask for confirmation before closing"),
                         'ask_before_closing')
         reset_namespace_box = newcb(
-                _("Ask for confirmation before reseting the IPython "
-                  "console namespace"), 'show_reset_namespace_warning',
+                _("Ask for confirmation before resetting the namespace."),
+                'show_reset_namespace_warning',
                 tip=_("This option lets you hide the warning message shown\n"
-                      "when reseting the IPython console namespace from\n"
-                      "the Spyder GUI"))
+                      "when resetting the namespace from Spyder."))
 
         interface_layout = QVBoxLayout()
         interface_layout.addWidget(banner_box)
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -303,8 +303,8 @@ class IPythonConsoleConfigPage(PluginConfigPage):
                 _("Ask for confirmation before reseting the IPython "
                   "console namespace"), 'show_reset_namespace_warning',
                 tip=_("This option lets you hide the warning message shown\n"
-                      "when reseting the IPython console namespace from the\n"
-                      "variable explorer 'reset namespace' button."))
+                      "when reseting the IPython console namespace from\n"
+                      "the Spyder GUI"))
 
         interface_layout = QVBoxLayout()
         interface_layout.addWidget(banner_box)
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -299,12 +299,19 @@ class IPythonConsoleConfigPage(PluginConfigPage):
         calltips_box = newcb(_("Display balloon tips"), 'show_calltips')
         ask_box = newcb(_("Ask for confirmation before closing"),
                         'ask_before_closing')
+        reset_namespace_box = newcb(
+                _("Ask for confirmation before reseting the IPython "
+                  "console namespace"), 'show_reset_namespace_warning',
+                tip=_("This option lets you hide the warning message shown\n"
+                      "when reseting the IPython console namespace from the\n"
+                      "variable explorer 'reset namespace' button."))
 
         interface_layout = QVBoxLayout()
         interface_layout.addWidget(banner_box)
         interface_layout.addWidget(pager_box)
         interface_layout.addWidget(calltips_box)
         interface_layout.addWidget(ask_box)
+        interface_layout.addWidget(reset_namespace_box)
         interface_group.setLayout(interface_layout)
 
         comp_group = QGroupBox(_("Completion Type"))
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -919,9 +919,9 @@ class IPythonConsole(SpyderPluginWidget):
                     sw.silent_execute('%clear')
                     sw.silent_execute(
                         'get_ipython().kernel.close_all_mpl_figures()')
-                    sw.reset_namespace(force=True)
+                    sw.reset_namespace(warning=False, silent=True)
                 elif current_client and clear_variables:
-                    sw.reset_namespace(force=True)
+                    sw.reset_namespace(warning=False, silent=True)
                 sw.execute(to_text_string(to_text_string(lines)))
             self.activateWindow()
             self.get_current_client().get_control().setFocus()
--- a/spyder/widgets/ipythonconsole/shell.py
+++ b/spyder/widgets/ipythonconsole/shell.py
@@ -14,6 +14,7 @@ import uuid
 
 from qtpy.QtCore import Signal
 from qtpy.QtWidgets import QMessageBox
+from spyder.config.main import CONF
 from spyder.config.base import _
 from spyder.config.gui import config_shortcut
 from spyder.py3compat import PY2, to_text_string
@@ -21,6 +22,7 @@ from spyder.utils import encoding
 from spyder.utils import programs
 from spyder.utils import syntaxhighlighters as sh
 from spyder.utils.ipython.style import create_qss_style, create_style_class
+from spyder.widgets.helperwidgets import MessageCheckBox
 from spyder.widgets.ipythonconsole import (ControlWidget, DebuggingWidget,
                                            HelpWidget, NamepaceBrowserWidget,
                                            PageControlWidget)
@@ -40,6 +42,7 @@ class ShellWidget(NamepaceBrowserWidget, HelpWidget, DebuggingWidget):
     sig_var_properties = Signal(object)
     sig_show_syspath = Signal(object)
     sig_show_env = Signal(object)
+    sig_namespace_reseted = Signal()
 
     # For DebuggingWidget
     sig_pdb_step = Signal(str, int)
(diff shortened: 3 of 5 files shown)


ccordoba12:
Nice! Thanks a lot for this @jnsebgosselin!

goanpeca:
@jnsebgosselin could we make the button look flat, like the others?

jnsebgosselin:
@goanpeca it is like the others. l think it's because my mouse cursor was on it, but the cursor is not shown in the screenshot lol.

goanpeca:
@jnsebgosselin ahh ok :-), then LGTM :-)

goanpeca:
Merging @ccordoba12 ?

jnsebgosselin:
@goanpeca Where should it be available in the menu? Under console I guess?

goanpeca:
@ccordoba12 under the cog menu in the variable explorer? and under the consoles as well?

jnsebgosselin:
@goanpeca  Ahh I see. Ok, I can add it.

jnsebgosselin:
Here is another screenshot showing the warning message and the option to mute the warning in the preferences.

reset_with_namespace

goanpeca:
@jnsebgosselin pretty cool :-)

jnsebgosselin:
I've just seen that the method to clear the IPython console in `shell.py` can be executed with the ctr-alt-R shortcut in the UI. There is also a warning message, but it cannot be muted by the user. I was already connecting to this method to clear the console but I didn't realise that this method was doing stuff at the UI level. I assumed this was more low level stuff.

So, I will try to blend what I've done with what already exists so that everything is handled by the same code base. Therefore, please wait a  little before reviewing, I'm still working on it.

jnsebgosselin:
I still need to add the functionality to the cog menu. I will do it tonight.

jnsebgosselin:
@ccordoba12 @goanpeca I'm working to add a "reset namespace" action to the cog menu of the variable explorer and of the IPython console as suggested by @goanpeca . However, I'm not sure this belong there. None of the actions in the variable explorer toolbar appear in the cog menu. Also, none of the action in the context menu appears in the the cog menu of the console. So this does not feel consistent in both cases to only add the reset namespace action there... What do you guys think?

spyder/config/main.py
```diff
@@ -144,6 +144,7 @@
               'use_pager': False,
               'show_calltips': True,
               'ask_before_closing': False,
+              'show_reset_namespace_warning': True,
```
ccordoba12: Please move this option to the `variable_explorer` section

ccordoba12: Sorry, it's ok here.

jnsebgosselin:
Here an animation to help you further in reviewing this. Please see also the updates in my last comment.

reset_namespace

spyder/widgets/ipythonconsole/shell.py
```diff
@@ -219,27 +222,44 @@ def clear_console(self):
...
+
+        if self._reading:
+            self.dbg_exec_magic('reset', '-f')
         else:
-            if self._reading:
-                self.dbg_exec_magic('reset', '-f')
-            else:
+            if silent:
                 self.silent_execute("%reset -f")
+                self.sig_namespace_reseted.emit()
```
rlaverde: Why the signal is only emitted when silent is True?

jnsebgosselin: When silent is False, something is written in the console, which trigger a refresh of the variable table automatically, so the signal is not needed. I was afraid that by emitting `sig_namespace_reseted` also when silent is False, we would trigger two calls to `refresh_table`.

rlaverde: >I was afraid that by emitting sig_namespace_reseted also when silent is False, we would trigger two calls to refresh_table.

Hmm yes, that will happen; but I'll expect a signal named `sig_namespace_reseted` to be emited every time that the namespace is reseted

I don't know how to improve this, maybe a signal `namespace_changed` that'll be emitted by all things that change the namespace, and will be connected to `refresh_table`, although that is out of the scope of this PR

In the meanwhile we could add a comment explaining why the signal is only emitted  when silent=True

jnsebgosselin: Ok good idea, I will add a comment.

goanpeca:
Ready to merge?

jnsebgosselin:
@goanpeca in my opinion, it is ready. 

But I've never received any feedback about the [add action to cog menu](https://github.com/spyder-ide/spyder/pull/4885#issuecomment-321945294) suggestion you made previously. Should I add it or not?

goanpeca:
>But I've never received any feedback about the add action to cog menu suggestion you made previously. Should I add it or not?

I like what you did, it makes sense

spyder/plugins/ipythonconsole.py
```diff
@@ -300,12 +300,19 @@ def setup_page(self):
         calltips_box = newcb(_("Display balloon tips"), 'show_calltips')
         ask_box = newcb(_("Ask for confirmation before closing"),
                         'ask_before_closing')
+        reset_namespace_box = newcb(
+                _("Ask for confirmation before reseting the IPython "
+                  "console namespace"), 'show_reset_namespace_warning',
```
ccordoba12: Please change this text to

> Ask for confirmation before resetting the namespace

spyder/plugins/ipythonconsole.py
```diff
@@ -300,12 +300,19 @@ def setup_page(self):
         calltips_box = newcb(_("Display balloon tips"), 'show_calltips')
         ask_box = newcb(_("Ask for confirmation before closing"),
                         'ask_before_closing')
+        reset_namespace_box = newcb(
+                _("Ask for confirmation before reseting the IPython "
+                  "console namespace"), 'show_reset_namespace_warning',
+                tip=_("This option lets you hide the warning message shown\n"
+                      "when reseting the IPython console namespace from\n"
+                      "the Spyder GUI"))
```
ccordoba12: Please change this to

> This option lets you hide the warning message shown when resetting the namespace from Spyder

spyder/widgets/ipythonconsole/shell.py
```diff
@@ -220,27 +223,49 @@ def clear_console(self):
         else:
             self.execute("%clear")
 
-    def reset_namespace(self, force=False):
+    def reset_namespace(self, warning=None, silent=True):
```
ccordoba12: I think this is easier to understand if we make `warning=False` by default.

spyder/widgets/ipythonconsole/shell.py
```diff
@@ -220,27 +223,49 @@ def clear_console(self):
...
-            if reply == QMessageBox.Yes:
-                if self._reading:
-                    self.dbg_exec_magic('reset', '-f')
-                else:
-                    self.execute("%reset -f")
+
+        if warning is None:
+            section = 'ipython_console'
+            option = 'show_reset_namespace_warning'
+            warning = CONF.get(section, option)
```
ccordoba12: By making `warning=False`, we can remove this entire block.

spyder/widgets/ipythonconsole/shell.py
```diff
@@ -220,27 +223,49 @@ def clear_console(self):
...
-                self.dbg_exec_magic('reset', '-f')
-            else:
+            if silent:
                 self.silent_execute("%reset -f")
+                self.sig_namespace_reseted_silently.emit()
+                # This signal is needed to force a refresh of the
+                # namespacebrowser table when the reset is done silently. When
+                # silent is False, something is written in the console,
+                # which trigger a refresh of the variable table automatically.
+                # See PR #4885
```
ccordoba12: We don't need this additional signal because you can refresh the Variable Explorer from every `shellwidget`, like this

```python
self.refresh_namespacebrowser()
```

spyder/widgets/variableexplorer/namespacebrowser.py
```diff
@@ -180,6 +180,7 @@ def set_shellwidget(self, shellwidget):
         """Bind shellwidget instance to namespace browser"""
         self.shellwidget = shellwidget
         shellwidget.set_namespacebrowser(self)
+        shellwidget.sig_namespace_reseted_silently.connect(self.refresh_table)
```
ccordoba12: Please remove this.

spyder/widgets/ipythonconsole/shell.py
```diff
@@ -41,6 +43,7 @@ class ShellWidget(NamepaceBrowserWidget, HelpWidget, DebuggingWidget):
     sig_var_properties = Signal(object)
     sig_show_syspath = Signal(object)
     sig_show_env = Signal(object)
+    sig_namespace_reseted_silently = Signal()
```
ccordoba12: This is not necessary anymore.

spyder/widgets/variableexplorer/namespacebrowser.py
```diff
@@ -448,7 +453,11 @@ def import_data(self, filenames=None):
                                        "<br><br>Error message:<br>%s"
                                        ) % (self.filename, error_message))
             self.refresh_table()
-            
+
+    @Slot()
+    def reset_namespace(self):
+        self.shellwidget.reset_namespace(silent=True)
```
ccordoba12: To use the option saved in our config system, you need to change this to

```python
warning = CONF.get('ipython_console',
                   'show_reset_namespace_warning')
self.shellwidget.reset_namespace(warning=warning,
                                 silent=True)
```

ccordoba12:
Sorry for the delay, I was busy with 3.2.2. I did a thorough review and I think my suggested changes would help to simply the code a lot.

----

> But I've never received any feedback about the add action to cog menu suggestion you made previously. Should I add it or not?

No, please don't add it. The cog menu only contains config options right now, so this entry will look strange there.

----

@jnsebgosselin, please address my comments today or tomorrow so we can include this in 3.2.3. We need to release it this weekend because of an error introduced in the IPython console on Windows.

Else, this will have to wait until 3.2.4. Hopefully I think this wouldn't take you more than half an hour of work ;-)

spyder/widgets/ipythonconsole/shell.py
jnsebgosselin: Nice, thank you!

spyder/widgets/ipythonconsole/shell.py
jnsebgosselin: I agree it makes the code clearer. However, `warning` will then always be `False` when `reset_namespace` is called from the console context menu. Therefore, I created a new method called `_reset_namespace` that is now used as a slot for the "reset namespace" action from the context menu instead of connecting it directly to the `reset_namespace` method.

jnsebgosselin:
@Thanks for the very good revision @ccordoba12. I applied everything as is, except the part with the `warning=False` by default because it would have broken some functionality of this new feature. You can check if my solution is ok and I can make some final revisions tomorrow before the release.

Issue #4504, opened 2017-05-23
Exceptions swallow the prompt!


dhirschfeld:
*Sometimes* when an exception occurs the prompt doesn't come back until you specifically press the `ESC` button. 

**What steps will reproduce the problem?**

1. Try to execute some code which throws an excpetion.
2. *somethimes* you might not get the prompt back :disappointed: 

**What is the expected output? What do you see instead?**
I expect to immediately get the next input prompt but no prompt is shown until you press `ESC`.

**Please provide any additional information below**
In the GIF below the "hanging" behaviour is seen the last time I execute the incorrect code when the prompt isn't returned until I press `ESC`

![exception](https://cloud.githubusercontent.com/assets/881019/26334785/e3d514d2-3faa-11e7-9752-91c220f43bb9.gif)

## Versions and main components
This is on `master`
![image](https://cloud.githubusercontent.com/assets/881019/26334828/250fddec-3fab-11e7-9546-48fc468880a1.png)

## Dependencies
```
cython >=0.21    :  0.25.2 (OK)
jedi >=0.9.0     :  0.10.2 (OK)
matplotlib >=1.0 :  2.0.2 (OK)
nbconvert >=4.0  :  5.1.1 (OK)
numpy >=1.7      :  1.12.1 (OK)
pandas >=0.13.1  :  0.20.1 (OK)
pycodestyle >=2.3:  None (NOK)
pyflakes >=0.6.0 :  1.5.0 (OK)
pygments >=2.0   :  2.2.0 (OK)
pylint >=0.25    :  1.6.5 (OK)
qtconsole >=4.2.0:  4.3.0 (OK)
rope >=0.9.4     :  0.10.5 (OK)
sphinx >=0.6.6   :  1.6.1 (OK)
sympy >=0.7.3    :  1.0 (OK)
```

impact27:
I see the same thing but I would not say that I don't get the prompt back. It is just that sometimes exceptions are written in the next input. In dhirschfeld's gif, the exception thrown in `In[32]` is displayed as text after `In[33]`. It should therefore be removed by either `esc` or removing the characters one by one. If I had to guess I would say that the prompt is coming back before the exception  is displayed.

impact27:
I tried to implement `_handle_error` in `NamepaceBrowserWidget` and it sometimes gets called after `_handle_execute_reply`, which causes this issue.

ccordoba12:
[...]

PR #4914, opened 2017-08-08
PR: Display prompt after errors

impact27:
This solves issue #4504  but will sometimes display two `in[]`

### DIFF
--- a/spyder/widgets/ipythonconsole/namespacebrowser.py
+++ b/spyder/widgets/ipythonconsole/namespacebrowser.py
@@ -189,14 +189,7 @@ class NamepaceBrowserWidget(RichJupyterWidget):
         if pdb_state is not None and isinstance(pdb_state, dict):
             self.refresh_from_pdb(pdb_state)
 
-    # ---- Private API (overrode by us) ----------------------------
-    def _handle_error(self, msg):
-        """
-        Reimplemented to reset the prompt if the error comes after the reply
-        """
-        super()._handle_error(msg)
-        self._show_interpreter_prompt()
-    
+    # ---- Private API (overrode by us) ----------------------------    
     def _handle_execute_reply(self, msg):
         """
         Reimplemented to handle communications between Spyder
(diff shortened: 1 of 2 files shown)


impact27:
screen shot 2017-08-08 at 15 47 30
To clarify:
There is always an extra `In[]` when an exception occur with this workaround. But at least the last input line is clean.

ccordoba12:
@dhirschfeld, what do you think about @impact27's solution?

spyder/widgets/ipythonconsole/namespacebrowser.py
```diff
@@ -190,6 +190,13 @@ def _handle_data_message(self, msg):
             self.refresh_from_pdb(pdb_state)
 
     # ---- Private API (overrode by us) ----------------------------
+    def _handle_error(self, msg):
+        """
+        Reimplemented to reset the prompt if the error comes after the reply
+        """
+        super()._handle_error(msg)
```
ccordoba12: I think this is wrong. It has to be

    super(NamepaceBrowserWidget, self)._handle_error(msg)

Also, please move this addition to `ShellWidget` because it's not related to the Variable Explorer.

dhirschfeld:
It seems a bit of a quirky solution - always insert an extra prompt after an exception and sometimes the exception will be printed in that prompt instead of associated with the prompt which caused it.

That said, as a work-around I think it does lead to a better end-user experience as you don't have to escape out of a traceback anymore.

If the underlying issue is with `ipykernel` perhaps it should be fixed there?

impact27:
It is definitely a work-around. If this is an issue with `ipykernel` wouldn't it show in every iPython consoles? Do you think we should open a bug report with `ipykernel` even though this might be related to spyder only?

ccordoba12:
Our tests are failing in PyQt4, but I'm not exactly sure how to fix them...

impact27:
If the error is "AttributeError: 'super' object has no attribute '_handle_error'" I could add a test to check if super has _handle_error? Worst case the msg is not handled, which means the error is not printed. I am not sure how this is happening?

ccordoba12:
> I could add a test to check if super has _handle_error? Worst case the msg is not handled, which means the error is not printed

I agree with this option, but please do it only for PyQt4. There's a constant in `qtpy` called `PYQT4` that let's you check if Spyder is running with those bindings.

> I am not sure how this is happening?

No idea either, but it's not so serious because very few people is using PyQt4 now.

PR #4939, opened 2017-08-11
PR: Some File Explorer Enhancements

### LINKED ISSUE #2406: File explorer enhancements
# Description

I have some suggestions on how to enhance the spyder file explorer panel:

At the moment if a file-type/content is supported by the spyder editor, it has an context action "Edit", otherwise it has the somewhat ubiquitous entry "Open" which delegates to the OS. This also results in folders having an "Open" command that opens the OS file explorer at that position.

I came up with these ideas, after trying to delete multiple files at once from within spyder. Which apparently is not possible because a) there is no multiselection+delete b) I could not easily switch to the OS file explorer and delete them there.
# Todo

I would propose
- [ ] an dedicated context menu entry for "open in OS file explorer" common to both files and folders.
- [ ] a "open via OS" action, i.e. allow the OS to determine what program to use to open the file.
- [ ] remove "Open" on folders ~~(or change its function to actually opening it in the spyder file 
  explorer, i.e. same as double-click action)~~
- [ ] Add multiselection (i.e. Ctrl/Shift+Click) and enable the (appropriate?) context menu actions to work on that selection. 
- [ ] remove "global" options from context menu. Move them to panel options menu instead.

jonnyhsu:
Fixes #2406 

This pull request fixes issues in #2406, plus a few minor improvements

Changes:
* Add "Open With OS" to all files, but not folders. This action opens the selected files with the default program.
* Remove "Open" from folders, because it was a duplicate functionality of "Open in external file browser"
* Add multi-select (ctrl / shift + click)
* Remove global options from context menu. These are not file related actions and are all present in the gear menu on the top right of the file explorer.
* Rename "Open in external file browser" to the more commonly used "Show in Folder" (MacOS remains "Show in Finder").
* Make "Show in Folder" action highlight the file in Windows Explorer. Previously it would open the folder but not highlight the file. MacOS already has this implemented, and in linux it is not implemented because xdg-open doesn't support highlighting files.
* Move "Open With OS" and "Show in Folder" out of the file editing section of the context menu into their own section.

The context menu will show different actions depending on which files are selected. Other than the changes mentioned above, the rules for what actions are shown remain the same. When multiple files are selected, the set of actions shown is the intersection of the actions for the individual files. Commit/Browse Repository are not shown for multiple selections because they don't operate on multiple files.

*.py file:
py_only

folder:
folder_only

Multi select: folder, *.py, binary file
folder_py_bin

### DIFF
--- a/spyder/widgets/explorer.py
+++ b/spyder/widgets/explorer.py
@@ -346,7 +346,8 @@ class DirView(QTreeView):
         rename_action = create_action(self, _("Rename..."),
                                       icon=ima.icon('rename'),
                                       triggered=self.rename)
-        open_action = create_action(self, _("Open"), triggered=self.open)
+        open_external_action = create_action(self, _("Open With OS"), 
+                                             triggered=self.open_external)
         ipynb_convert_action = create_action(self, _("Convert to Python script"),
                                              icon=ima.icon('python'),
                                              triggered=self.convert_notebooks)
@@ -356,8 +357,8 @@ class DirView(QTreeView):
             actions.append(run_action)
         if only_valid and only_files:
             actions.append(edit_action)
-        else:
-            actions.append(open_action)
+        if only_files:
+            actions.append(open_external_action)
         if sys.platform == 'darwin':
             text=_("Show in Finder")
         else:
@@ -533,6 +534,14 @@ class DirView(QTreeView):
                 self.parent_widget.sig_open_file.emit(fname)
             else:
                 self.open_outside_spyder([fname])
+                
+    @Slot()
+    def open_external(self, fnames=None):
+        """Open files with default application"""
+        if fnames is None:
+            fnames = self.get_selected_filenames()
+        for fname in fnames:
+            self.open_outside_spyder([fname])
         
     def open_outside_spyder(self, fnames):
         """Open file outside Spyder with the appropriate application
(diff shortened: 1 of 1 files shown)


ccordoba12:
Thanks a lot for your contribution! It looks really nice!

We'll try to review it soon :-)

spyder/widgets/explorer.py
```diff
@@ -45,15 +45,15 @@
 except:
     nbexporter = None    # analysis:ignore
 
+
 def open_file_in_external_explorer(filename):
     if sys.platform == "darwin":
         subprocess.call(["open", "-R", filename])
+    if os.name == 'nt':
```
rlaverde: I think this should be change it for `elif os.name == 'nt':`

```
if sys.platform == "darwin":
    ...
elif os.name == 'nt':
   ...
else:
    ...
```

if not in macOS it will execute both mac and linux command

jonnyhsu: Ah yes, I'm glad you caught that. I'll fix it now.

ccordoba12:
@jonnyhsu, sorry for the delay in merging your work.

@dalthviz, please test locally that all @jonnyhsu additions work for you, then I'll merge.

dalthviz:
@ccordoba12 I have tested locally and it works for me :+1:

goanpeca:
Let's merge then :-p @ccordoba12.

Awesome work @jonnyhsu! Thanks for the review @dalthviz!

PR #4949, opened 2017-08-12
PR: Add basic pep8speaks config file.

rlaverde:
Looking at the configurations https://github.com/OrkoHunter/pep8speaks#configuration I think `diff_only` is the only option we need to change.

https://pep8speaks.com/

### DIFF
--- a/.pep8speaks.yml
+++ b/.pep8speaks.yml
@@ -1,0 +1,4 @@
+# File : .pep8speaks.yml
+
+scanner:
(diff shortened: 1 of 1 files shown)


ccordoba12:
@rlaverde, please move this one to `3.x` so we have pep8 working in that branch and master (when I merge this PR).

ccordoba12:
I activated the service just now. Let's see how this goes ;-)

Issue #4509, opened 2017-05-23
UX/UI Design review


goanpeca:
This epic will group all the issues related to making the UI for Spyder more simple and usable

Spyder has grown organically and we just keep adding and adding toolbars, buttons menu and there is no roadmap or coherent direction for this.

PR #4954, opened 2017-08-13
PR: Unify layout for panes and widgets

### LINKED ISSUE #4953: Consistent layout in different panes

The editor and other panes should use the same vertical position for the main widget. This would result in a more consistent look.

**Current behavior**: Panes with toolbars have an additional offset.

Example: [ Editor | Explorer | Help ]
2017_08_13_18_24_49_spyder_3 2 1 dev0_python_3 6_

**Expected**: For Panes with a 'single' line of controls (editor tabs / toolbar), the main widget should start at the same vertical offset for all panes.

Example: [ Editor | Explorer | Help ]
2017_08_13_18_21_49_spyder_3 2 1 dev0_python_3 6_

The differences arise because the mechanisms are different:
1. The editor uses a tab widget.
2. (Some) other panes use a QVBoxLayout containing one widget for the controls and a separate main widget.

The height of the tab bar (1.) is not equal to the height of the control widgets plus layout spacing (2.).
One solution is to reduce the layout spacing (`layout.setSpacing(2)` in the above example for expected).

Caveats:
- This works on Windows 10. Other window systems may have other dimensions.
- It's a bit difficult to decide on the really optimal spacing: With spacing 2 the editor (blue outline) and the main widgets of the other panes are exactly aligned. However the tab bar itself has a grey bottom line. thus the editor appears to be a pixel above. But given that spacing 2 results in equal top and bottom margin for the combo box and line edit, I still consider it the best value.

2017_08_13_19_26_01_2017_08_13_18_21_49_spyder_3 2 1 dev0_python_3 6_ png_paint net_v4 0 17

If desired, i can provide a pull request.

* Spyder Version: 3.2.1.dev0 af176c3
* Python Version: 3.6.2
* Qt Versions: 5.6.2, PyQt5 5.6

## Dependencies

numpy >=1.7      :  1.13.1 (OK)

timhoffm:
fixes #4953 

`create_control_layout()` provides a unified way of creating the required layouts. Here's an example how it will look like:

2017-08-13 22_19_53-spyder 3 2 1 dev0 python 3 6 debug mode 1

The panes with multiple lines of control widgets (bottom panes in the screenshot) can be addressed at a later time.

### DIFF
--- a/spyder/utils/qthelpers.py
+++ b/spyder/utils/qthelpers.py
@@ -17,9 +17,9 @@ from qtpy.compat import to_qvariant, from_qvariant
 from qtpy.QtCore import (QEvent, QLibraryInfo, QLocale, QObject, Qt, QTimer,
                          QTranslator, Signal, Slot)
 from qtpy.QtGui import QIcon, QKeyEvent, QKeySequence, QPixmap
-from qtpy.QtWidgets import (QAction, QApplication, QHBoxLayout,
-                            QLabel, QLayout, QLineEdit, QMenu, QStyle,
-                            QToolBar, QToolButton, QVBoxLayout, QWidget)
+from qtpy.QtWidgets import (QAction, QApplication, QHBoxLayout, QLabel,
+                            QLineEdit, QMenu, QStyle, QToolBar, QToolButton,
+                            QVBoxLayout, QWidget)
 
 # Local imports
 from spyder.config.base import get_image_path, running_in_mac_app
@@ -501,41 +501,6 @@ def show_std_icons():
     dialog = ShowStdIcons(None)
     dialog.show()
     sys.exit(app.exec_())
-
-
-def create_control_layout(controls, main_widget):
-    """
-    Returns a layout for a set of controls above a main widget. This is a
-    standard layout for many panes and some other widgets.
-
-    The use-case is currently limited to a single line of controls.
-
-    controls: a list of QWidgets to be placed in a HBoxLayout at the top.
-        You may use 'stretch' to add a stretch to the layout.
-        If a more complex control of the layout is needed, you can pass a
-        ready made layout instead.
-    main_widget: the main widget. Can be None, if you want to add this
-        manually later on.
-    """
-    layout = QVBoxLayout()
-    layout.setContentsMargins(0, 0, 0, 0)
-    layout.setSpacing(2)
-
-    if isinstance(controls, QLayout):
-        layout.addLayout(controls)
-    else:
-        clayout = QHBoxLayout()
-        clayout.setAlignment(Qt.AlignLeft)
-        for control in controls:
-            if control == 'stretch':
-                clayout.addStretch()
-            else:
-                clayout.addWidget(control)
-        layout.addLayout(clayout)
-
-    if main_widget is not None:
-        layout.addWidget(main_widget)
-    return layout
 
 
 MENU_SEPARATOR = None
--- a/spyder/utils/qthelpers.py
+++ b/spyder/utils/qthelpers.py
@@ -523,7 +523,7 @@ def calc_tools_spacing(tools_layout):
     """
     metrics = {  # (tabbar_height, offset)
         'nt.fusion': (32, 0),
-        'nt.windowsvista': (21, 2),
+        'nt.windowsvista': (21, 3),
         'nt.windowsxp': (24, 0),
         'nt.windows': (21, 3),
         'posix.breeze': (28, -1),
--- a/spyder/utils/qthelpers.py
+++ b/spyder/utils/qthelpers.py
@@ -13,6 +13,7 @@ import re
 import sys
 
 # Third party imports
+from PyQt5.QtWidgets import QStyleFactory
 from qtpy.compat import to_qvariant, from_qvariant
 from qtpy.QtCore import (QEvent, QLibraryInfo, QLocale, QObject, Qt, QTimer,
                          QTranslator, Signal, Slot)
@@ -516,6 +517,18 @@ def create_plugin_layout(tools_layout, main_widget):
     """
     layout = QVBoxLayout()
     layout.setContentsMargins(0, 0, 0, 0)
+
+    if os.name == 'nt':
+        # adapt spacing so that main_widget has the same vertical position as
+        # the editor widgets (which have tabs above)
+        # TODO: This is a hard coded workaround until we find a way to
+        #       calculate is from the metrics.
+        spacings = {'fusion': 9, 'windowsvista': 2, 'windowsxp': 2,
+                    'windows': 2}
+        style_name = qapplication().style().property('name')
+        if style_name in spacings:
+            layout.setSpacing(spacings[style_name])
+
     layout.addLayout(tools_layout)
     if main_widget is not None:
         layout.addWidget(main_widget)
--- a/spyder/utils/qthelpers.py
+++ b/spyder/utils/qthelpers.py
@@ -504,6 +504,46 @@ def show_std_icons():
     sys.exit(app.exec_())
 
 
+def calc_tools_spacing(tools_layout):
+    """
+    Return a spacing (int) or None if we don't have the appropriate metrics
+    to calculate the spacing.
+
+    We're trying to adapt the spacing below the tools_layout spacing so that
+    the main_widget has the same vertical position as the editor widgets
+    (which have tabs above).
+
+    The required spacing is
+
+        spacing = tabbar_height - tools_height + offset
+
+    where the tabbar_heights were empirically determined for a combination of
+    operating systems and styles. Offsets were manually adjusted, so that the
+    heights of main_widgets and editor widgets match. This is probably
+    caused by a still not understood element of the layout and style metrics.
+    """
+    metrics = {  # (tabbar_height, offset)
+        'nt.fusion': (32, 0),
+        'nt.windowsvista': (21, 2),
+        'nt.windowsxp': (24, 0),
+        'nt.windows': (21, 3),
+        'posix.breeze': (28, -1),
+        'posix.oxygen': (38, -2),
+        'posix.qtcurve': (27, 0),
+        'posix.windows': (26, 0),
+        'posix.fusion': (32, 0),
+    }
+
+    style_name = qapplication().style().property('name')
+    key = '%s.%s' % (os.name, style_name)
+
+    if key in metrics:
+        tabbar_height, offset = metrics[key]
+        tools_height = tools_layout.sizeHint().height()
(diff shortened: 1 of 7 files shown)


ccordoba12:
@goanpeca, would you like to take a look at this one?

spyder/utils/qthelpers.py
```diff
@@ -502,7 +502,42 @@ def show_std_icons():
...
+    controls: a list of QWidgets to be placed in a HBoxLayout at the top.
+        You may use 'stretch' to add a stretch to the layout.
+        If a more complex control of the layout is needed, you can pass a
+        ready made layout instead.
+    main_widget: the main widget. Can be None, if you want to add this
+        manually later on.
+    """
+    layout = QVBoxLayout()
+    layout.setContentsMargins(0, 0, 0, 0)
+    layout.setSpacing(2)
```
goanpeca: @timhoffm where is this value coming from? Could we make it OS dependent on some Qt metric?

spyder/utils/qthelpers.py
```diff
@@ -502,7 +502,42 @@ def show_std_icons():
     dialog.show()
     sys.exit(app.exec_())
 
-    
+
+def create_control_layout(controls, main_widget):
```
goanpeca: Maybe it is better if we have 

`create_control_layout(widgets=None, layout=None, main_widget)` and we can either provide one or the other but not both?

goanpeca:
>@goanpeca, would you like to take a look at this one?

Yep, left some comments

spyder/widgets/explorer.py
```diff
@@ -1201,16 +1202,9 @@ def __init__(self, parent=None, name_filters=['*.py', '*.pyw'],
...
-        blayout.addWidget(button_previous)
-        blayout.addWidget(button_next)
-        blayout.addWidget(button_parent)
-        blayout.addStretch()
-        blayout.addWidget(self.button_menu)
-
-        layout = QVBoxLayout()
-        layout.addLayout(blayout)
-        layout.addWidget(self.treewidget)
+        controls = [button_previous, button_next, button_parent, 'stretch',
```
goanpeca: I don't like the `'stretch'` idea. I think it is better to just leave this outside of the control_layout func.

spyder/utils/qthelpers.py
timhoffm: As described in #4953, this is the value to align the main widgets to the editors, i.e. `"height(controls) + spacing = height(editor tabbar)"`. I've determined this experimentally on windows. It may be os-dependent and even style dependent (should be checked). There's definitively room for improvement here. If you want, we could limit this to windows as a first step:

    if os.name == 'nt':
        layout.setSpacing(2)

Then, the patch would not modify the layout on other OS, but we already have a central place to do this later on.

I'm not sure, if the value is easily computable from Qt metrics. We might have to test and provide a hard-coded list of values.

spyder/utils/qthelpers.py
timhoffm: I've intentionally chosen the first approach. First, you would even need `create_control_layout(widgets=None, layout=None, main_widget=None)` and say: "provide `main_widget` and either `widgets` or `layout`. Alternatively, putting `main_widgets` first as positional argument is also not very nice, because it does not reflect the natural order in which the widgets will be drawn.

Secondly, I think it's perfectly right to use just one variable for "the things to put at the top", no matter what this actually is. Having the ability to put either a list of widgets or a layout is a suitable generalization.

spyder/widgets/explorer.py
timhoffm: I see the list `controls` as a recipe telling the function how the layout should look like.

I'm not dogmatic about `'stetch'`. It's basically a convenience, because you'll often need it in the layout and would have to manually create the horizontal layouts if you don't have something like this. So it saves work.

If you're using `None` as a separator elsewhere we might go with that. Even though I would strongly go for `'stetch'`, when building from scratch. It's more explicit than `None` and more compact than adding an additional comment. Maybe even better would be `'add_stretch'`. Also, it's extendable in the future, if you want a fixed stretch (`'add_stretch(10)'`).

spyder/utils/qthelpers.py
goanpeca: Yes it should be OS centric, and you could use http://doc.qt.io/qt-5/qstyle.html#PixelMetric-enum

spyder/utils/qthelpers.py
goanpeca: @timhoffm I disagree, it makes the code less readable, and we need to add extra logic to see if one thing is a layout or a completely different thing.

spyder/widgets/explorer.py
goanpeca: @timhoffm if you really want to go in that direction then we should have a `StretchWidget` or something but mixing text with widgets is just wrong and error prone cause you can have a typo and we are not even checking if it text or not, just checking if it is a specific text (not suggesting we even do that).

spyder/utils/qthelpers.py
timhoffm: I've played around with this, but didn't get it to work. Querying the `pixelMetric()` crashes the progam. The docs are quite thin on this. I've posted a [question on stackoverflow](
https://stackoverflow.com/questions/45720566/how-to-determine-the-height-of-a-qtabbar).

spyder/utils/qthelpers.py
timhoffm: Well, we agree to disagree, then. :)

In a statically typed language, you would overload the function. Python does not do this because of dynamic typing. IMO `numpy.sin(2)`, `numpy.sin(numpy.array([2, 3]))`, `numpy.sin([2, 3])` would be a comparable case of allowing different types. Anyway, if you don't want this, we'll try something else.

The version with different keyword args (some of which must or must not be used) is worse than the proposed extra logic within the function. You would need the logic to check that a consistent set of arguments is provided. And then the extra logic to decide what to do.

To get things done, I propose to go one step back. I'll leave the creation of the tools layout to the individual dock widgets (as was till now) and only provide `create_control_layout(tools_layout, main_widget)`.

spyder/widgets/explorer.py
timhoffm: This is not necessary anymore if I don't handle the creation of the HBoxLayout.

spyder/utils/qthelpers.py
goanpeca: @timhoffm I understand the concept that you try to illustrate, but in this case it is not making things better, it is mixing a lot of concepts, widgets, layouts, strings that have a specific meaning only if you used the right string, and the idea at the end is to make the code understandable to everyone and not only you in this case.

`create_control_layout(tools_layout, main_widget)` sounds better, but to be honest the general solution to this issue should be to handle the top controls (if any) at the plugin level. @rlaverde do you have some suggestions here?

rlaverde: I'm not sure what if this is the best way to do this:
- It take in account plugins with one controls line (for example findinfiles have two lines), if we are going to standardize how controls are created I think that should be for all plugins
- I think the name isn't the best, sometimes it creates the main layout of the plugin, and sometimes it's the controls layout

A more general solution (out of the scope of this PR) thinking also about third party plugins, could be abstract the creation of the Main layout in the SpyderPlugin class, a plugin will define: a main widget, a controls_bar, and corner_actions; the plugins will lose flexibility but we could guaranty that the all look the same.

goanpeca: >SpyderPlugin class, a plugin will define: a main widget, a controls_bar, and corner_actions; the plugins will lose flexibility but we could guaranty that the all look the same.

Precisely my point and what I really prefer.

spyder/utils/qthelpers.py
rlaverde: I agree with @goanpeca this value should be OS dependent, I tested in linux and the plugins don't look aligned

spyder/utils/qthelpers.py
timhoffm: There would be a lot to say and discuss on the general UI design, but that's more in the scope of #4509.

I'm only trying to do a first small step here by:

1) extracting the currently common pattern of a single line of controls to a dedicated method. This is currently used sometimes in the widgets, sometimes in the plugin. I've intentionally not touched this to keep it simple.
2) Trying to find a way to to align this pattern with the tabbed editor.

I'll rewrite the patch to the simpler `create_control_layout(tools_layout, main_widget)` (better names are welcome, if you have one). Additionally, I'll limit the spacing modification to windows for the time being until I find a solution that's also working more generally.

That's not the solution to it alls, but one step better in both aspects. We can see further then.

timhoffm:
A new, try:

- only the simpler function to create the layout
- a semi-empirical approach to adapt the spacing. Works on windows and a number of Linux systems. Doesn't modify the spacing on systems that have not been tested and verified to work.

spyder/utils/qthelpers.py
```diff
@@ -502,7 +502,70 @@ def show_std_icons():
     dialog.show()
     sys.exit(app.exec_())
 
-    
+
+def calc_tools_spacing(tools_layout):
```
rlaverde: I didn't knew that os could differ that much 😲 

I tested in linux with kde and works as expected 😄

spyder/utils/qthelpers.py
```diff
@@ -502,7 +502,70 @@ def show_std_icons():
...
+    key = '%s.%s' % (os.name, style_name)
+
+    if key in metrics:
+        tabbar_height, offset = metrics[key]
+        tools_height = tools_layout.sizeHint().height()
+        spacing = tabbar_height - tools_height + offset
+        return max(spacing, 0)
+
+
+def create_plugin_layout(tools_layout, main_widget):
```
rlaverde: I like how you change this, It has a better name, and make one thing and do it right.

Maybe add a default value for main_widget
`def create_plugin_layout(tools_layout, main_widget=None):`

goanpeca:
@timhoffm yes much better thanks for the work :-)

spyder/utils/qthelpers.py
ccordoba12: Do you prefer @timhoffm to do you this change or not?

rlaverde: Yes, It think It'll  be easy to notice that the main_widget is optional.

Although It's not a critical change 🙈 without the change is ok too

ccordoba12: I agree with @rlaverde. @timhoffm, please do this minor change, then I'll merge.

timhoffm: @ccordoba12 Change done. Thanks all for the discussions. Sometimes a design has to ripen and it helps if people with different perspectives have a look at it. 😃

PR #4915, opened 2017-08-08
PR: Do not full-restart plugins when changing pythonpath

rlaverde:
Tries to fix #4410
see https://github.com/spyder-ide/spyder/pull/4783#issuecomment-320978826

Only restart `PluginClient` internals (process, socket and notifier) when setting a new extra path. Previously changing extra_path delete and create again `PluginManager` and `PluginClient` objects.

Although this PR avoid the full restart of `PluginManager` and `PluginClientt`, the `QProcess` has to be restarted in order to set a new environment variable.

I also cherrypick a test from #4783

PR #4962, opened 2017-08-14
PR: Send Spyder python path to introspection plugins

rlaverde:
Fixes: #4410

Supersedes #4915
Undo some changes of  #4263

At the moment the spyder path is only used by jedi

### DIFF
--- a/spyder/plugins/tests/test_editor_introspection.py
+++ b/spyder/plugins/tests/test_editor_introspection.py
@@ -58,7 +58,7 @@ def setup_editor(qtbot, monkeypatch):
     editor.introspector.plugin_manager.close()
 
 
-@flaky(max_runs=3)
+@pytest.mark.skipif(True, reason="This test fails too much in the CI :(")
 @pytest.mark.skipif(not JEDI_010,
                     reason="This feature is only supported in jedy >= 0.10")
 def test_introspection(setup_editor):
--- a/spyder/plugins/tests/test_editor_introspection.py
+++ b/spyder/plugins/tests/test_editor_introspection.py
@@ -58,7 +58,8 @@ def setup_editor(qtbot, monkeypatch):
     editor.introspector.plugin_manager.close()
 
 
-@pytest.mark.skipif(True, reason="This test fails too much in the CI :(")
+@pytest.mark.skipif(os.environ.get('CI', None) is not None,
+                    reason="This test fails too much in the CI :(")
 @pytest.mark.skipif(not JEDI_010,
                     reason="This feature is only supported in jedy >= 0.10")
 def test_introspection(setup_editor):
--- a/spyder/plugins/tests/test_editor_introspection.py
+++ b/spyder/plugins/tests/test_editor_introspection.py
@@ -1,0 +1,96 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+#
+"""Tests for the Editor plugin."""
+
+# Third party imports
+import pytest
+import os
+import os.path as osp
+from flaky import flaky
+
+try:
+    from unittest.mock import Mock
+except ImportError:
+    from mock import Mock  # Python 2
+
+from qtpy.QtWidgets import QWidget, QApplication
+from qtpy.QtCore import Qt
+
+from spyder.utils.introspection.jedi_plugin import JEDI_010
+from spyder.utils.qthelpers import qapplication
+
+# Location of this file
+LOCATION = osp.realpath(osp.join(os.getcwd(), osp.dirname(__file__)))
+
+
+@pytest.fixture
+def setup_editor(qtbot, monkeypatch):
+    """Set up the Editor plugin."""
+    app = qapplication()
+    os.environ['SPY_TEST_USE_INTROSPECTION'] = 'True'
+    monkeypatch.setattr('spyder.dependencies', Mock())
+    from spyder.plugins.editor import Editor
+
+    monkeypatch.setattr('spyder.plugins.editor.add_actions', Mock())
+
+    class MainMock(QWidget):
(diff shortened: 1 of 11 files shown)


spyder/utils/introspection/jedi_plugin.py
```diff
@@ -30,6 +30,8 @@
 except ImportError:
     jedi = None
 
+JEDI_010 = [int(i) for i in jedi.__version__.split('.')] >= [0, 10, 0]
```
ccordoba12: We have `programs.is_module_installed` exactly for this :-)

spyder/app/tests/test_mainwindow.py
```diff
@@ -110,7 +111,7 @@ def main_window(request):
     # (it's faster and less memory consuming not to use it!)
     marker = request.node.get_marker('use_introspection')
     if marker:
-        os.environ['SPY_TEST_USE_INTROSPECTION'] = 'True'
+        os.environ['SPY_TEST_USE_INTROSPECTION'] = 'USE_INTROSPECTION'
```
ccordoba12: Please change this to `True`. That could fix the failing test below.

ccordoba12:
The way to implement this for Rope is the following:

1. The `project` attribute of `RopePlugin` has a `prefs` sub-attribute, as can be seen here:

    https://github.com/spyder-ide/spyder/blob/3.x/spyder/utils/introspection/rope_plugin.py#L74

2. You can set a property in `prefs` called `python_path`, which is a list of additional paths to perform code completion. I think the way to do this is

        self.project.prefs.set('python_path', ['path1', 'path2'])

    but you need to test it to see if it's right.

3. The previous step needs to be done every time we call the `get_completions` method of `RopePlugin`.

spyder/app/tests/test_mainwindow.py
```diff
@@ -110,7 +111,7 @@ def main_window(request):
     # (it's faster and less memory consuming not to use it!)
     marker = request.node.get_marker('use_introspection')
     if marker:
-        os.environ['SPY_TEST_USE_INTROSPECTION'] = 'True'
+        os.environ['SPY_TEST_USE_INTROSPECTION'] = True
```
ccordoba12: This needs to be `'True'`, not `True`.

ccordoba12:
@rlaverde, please remove @dalthviz test because is timing out. Instead, please add tests to `test_rope_plugin.py` and `test_jedi_plugin.py` that get completions of some test module placed in a temp directory.

rlaverde:
>please remove @dalthviz test because is timing out. Instead, please add tests to test_rope_plugin.py and test_jedi_plugin.py that get completions of some test module placed in a temp directory.

I was trying to fix it without success 😞 I'll also try to move it to an editor test, maybe loading the main window is what's causing it to timeout.

ccordoba12:
@rlaverde, I think that adding the test to test_rope_plugin.py and test_jedi_plugin.py is easier and simpler than testing completion in our widgets.

rlaverde:
I think that adding the test to test_rope_plugin.py and test_jedi_plugin.py is easier and simpler than testing completion in our widgets.

Yes, I'll give a last try to the integration test and I'll do that (the test is running pretty well in my local configuration, I don't know way CI is failing 😞 )

andfoy:
@rlaverde Have you tried to debug those tests via SSH on Travis (https://docs.travis-ci.com/user/running-build-in-debug-mode/) and RDP on AppVeyor (https://www.appveyor.com/docs/how-to/rdp-to-build-worker/)?

spyder/plugins/tests/test_editor_introspection.py
```diff
@@ -0,0 +1,96 @@
...
+    editor.new(fname="test.py", text="")
+    editor.introspector.set_editor_widget(editor.editorstacks[0])
+
+    yield editor, qtbot
+    # teardown
+    os.environ['SPY_TEST_USE_INTROSPECTION'] = 'False'
+    editor.introspector.plugin_manager.close()
+
+
+@pytest.mark.skipif(True, reason="This test fails too much in the CI :(")
```
ccordoba12: Let's remove this condition and change the one below to

    not JEDI_010 or os.environ.get('CI', False)

so we can run this test locally.

ccordoba12:
@dalthviz, @arman00 and @laikh, please help us to test this PR to see if it fixes the problem with code completion in the Editor when using projects.

Thanks for your help!

dalthviz:
@ccordoba12 It works for me :+1: 🎉 , I tested opening and closing projects and changing the PYTHONPATH.

arman00:
Unfortunately it doesn't work for me. After I cloned from github , I switched to the 3.x branch. Version is 3.2.2.dev0.

It works if I create a new project, but with my existing project which has different modules in different folders, it doesn't work. I add those folders into the PYTHONPATH.

dalthviz:
Hi @arman00 to test this you need to checkout the @rlaverde branch `introspection-python-path`, did you do that?

arman00:
Oh I see... Didn't know that :) 
I will do it as soon as I have the chance again
Thanks

dalthviz:
Just in case, to test this you can clone the [fork](https://github.com/rlaverde/spyder) of @rlaverde and then checkout the corresponding branch.

goanpeca:
You can do

```
$ git clone https://github.com/rlaverde/spyder.git
$ cd spyder
$ git checkout introspection-python-path
```

laikh:
Hi guys,

Finally managed to test your solution and am happy to report that this fixed the issue for me!! :+1: 
I tested it with version 3.2.0 as 3.2.1 seems to be still not available for anaconda.

Thanx a lot, looking forward to the next version!

sincerely,
Michael

ccordoba12:
Great news!! Thanks a lot to you for helping us to test this :-)

Issue #4591, opened 2017-06-12
Split all plugins to be in their own modules


rlaverde:
An issue to keep track of spyder modularization:

- [x]  Help #4548
- [x]  Online Help #4565
- [x]  Variable Explorer #4557 
- [x]  Find in Files #4569
- [x]  Editor #4975
- [x]  Outline explorer #4812
- [x] IPython Console #5206
- [x] Internal Console #5207
- [x] History #4593
- [x]  Projects (project explorer) #5214
- [x]  Explorer #4974
- [x] Working Directory ~(will be merged with explorer?)~ #5219 
- [x] Preferences module #5216
- [x] Fix error in variableexplorer (https://github.com/spyder-ide/spyder/pull/5206#issuecomment-329018836) #5263

External plugins (spyder_breakpoints, spyder_breakpoints, spyder_io_hdf5, spyder_profiler, spyder_pylint)

- [x] Move back to the plugins folder (also change them to follow the same folder structure): #5276
- [x] merge translations #5438

Other things:
- [x] Merge with master and move new files (after 4.0beta2 release) #7725 

Other ideas (to further discuss):
- Add `PLUGIN_CLASS` to  internal plugins, and load them as external plugins are loaded

goanpeca:
Profiler, lint etc... should be brought back to the plugins folder

rlaverde:
> Profiler, lint etc... should be brought back to the plugins folder

yes, I forgot that

goanpeca:
We need a separate issue for after this, cause the idea is that these plugins should really be extending the variable explorer API. Currently the way they load is too weird...

spyder_io_dcm
 spyder_io_hdf5

ccordoba12:
Thanks for opening this one to keep track of this effort.

goanpeca:
@rlaverde @ccordoba12 how should we handle the preferences "plugin", its not really a plugin at the moment, more of a core feature of spyder, so we need to move it out of plugins folder?

Thoughts?

ccordoba12:
Right, we should move all Python modules related to Preferences (e.g. 
configdialog.py, rundialog.py, etc) to a different module called 
`preferences`, out of plugins and at the same level of it.

goanpeca:
Also @rlaverde what about https://github.com/spyder-ide/spyder/tree/split-plugins/spyder/plugins/tests

That belongs to the findinfiles plugin?

ccordoba12:
@goanpeca is right, that one is missing.

[...]

PR #4974, opened 2017-08-16
PR: Split Explorer module

rlaverde:
Part of #4591 

Should this module be merged with project explorer? If yes, how?

I think It's better to have two modules, and project module depending of explorer module


ccordoba12:
> I think It's better to have two modules, and project module depending of explorer module

Yes, I also it's better to have two separate plugins.

spyder/plugins/explorer/__init__.py
```diff
@@ -0,0 +1,12 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+
+"""
+spyder.plugins.explorer
+===========
```
ccordoba12: Please extend these `==` to cover the entire line above.

rlaverde:
I think this is ready :smile:

ccordoba12:
> or should I merge working directory into it?

Nop. Let's leave things like that after we merge `split-plugins` with master.

PR #4993, opened 2017-08-18
PR: Fix handling of line count in the fileswitcher

### LINKED ISSUE #4989: Spyder Issue when using ctrl+P : goto line
## Description

1. 
2. 
3. 

  File "C:\Program Files\Anaconda3\lib\site-packages\spyder\widgets\fileswitcher.py", line 755, in setup
    self.setup_file_list(filter_text, current_path)
  File "C:\Program Files\Anaconda3\lib\site-packages\spyder\widgets\fileswitcher.py", line 629, in setup_file_list
    text_item += " [{0:} {1:}]".format(self.line_count[index],
AttributeError: 'FileSwitcher' object has no attribute 'line_count'

* Spyder Version: 3.2.1 
* Python Version: 3.5.2
* Qt Versions: 5.6.0, PyQt5 5.6 on Windows

## Dependencies
```
sympy >=0.7.3    :  1.0 (OK)
cython >=0.21    :  0.24.1 (OK)

```

dalthviz:
Fixes #4989

### DIFF
--- a/spyder/widgets/fileswitcher.py
+++ b/spyder/widgets/fileswitcher.py
@@ -306,7 +306,7 @@ class FileSwitcher(QDialog):
         for widget in self.widgets:
             try:
                 line_count.append(widget[0].get_line_count())
-            except:
+            except AttributeError:
                 line_count.append(0)
         return line_count
 
--- a/spyder/widgets/fileswitcher.py
+++ b/spyder/widgets/fileswitcher.py
@@ -302,7 +302,13 @@ class FileSwitcher(QDialog):
 
     @property
     def line_count(self):
-        return [widget.get_line_count() for widget in self.widgets]
+        line_count = []
+        for widget in self.widgets:
+            try:
+                line_count.append(widget[0].get_line_count())
+            except:
+                line_count.append(0)
+        return line_count
 
     @property
     def save_status(self):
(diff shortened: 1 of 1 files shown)


spyder/widgets/fileswitcher.py
```diff
@@ -302,7 +302,13 @@ def widgets(self):
 
     @property
     def line_count(self):
-        return [widget.get_line_count() for widget in self.widgets]
+        line_count = []
+        for widget in self.widgets:
+            try:
+                line_count.append(widget[0].get_line_count())
```
goanpeca: @dalthviz if the possible exception is only triggered by

`widget[0].get_line_count()` then I suggest this is written as:

```python
line_count = []
for widget in self.widgets:
    try:
        current_line_count = widget[0].get_line_count()
    except AttributeError:
        current_line_count = 0
    line_count.append(current_line_count)
```

It is subtle difference, but makes clear what the possible error might be

ccordoba12: Why do we need to add 0 to the line count?

spyder/widgets/fileswitcher.py
```diff
@@ -609,12 +615,16 @@ def setup_file_list(self, filter_text, current_path):
         # Get optional line number
         if trying_for_line_number:
             filter_text, line_number = filter_text.split(':')
+            # Get all the available filenames
+            scores = get_search_scores('', self.filenames,
+                                       template="<b>{0}</b>")
```
ccordoba12: Why do we need to this? It seems wrong to me because in this place we're trying to get the line number introduced by the user.

dalthviz:
Preview:

Before:
before

After:
after

spyder/widgets/fileswitcher.py
dalthviz: I put it there as a default value. In the preview could be more clear.

spyder/widgets/fileswitcher.py
dalthviz: I did this in order to show all the options with their respective line count. It could be more clear with the preview below.

ccordoba12:
@dalthviz, thanks for the screenshot, but I disagree with this. Since we also want to add consoles and terminals to the File switcher, I think we should only show widgets with the `get_line_count` method when filtering by `line:`

PR #5002, opened 2017-08-18
PR: Add Editor extensions

rlaverde:
Fixes: #877 

--------------
See https://github.com/spyder-ide/spyder/pull/4987#issuecomment-323097210
I use #877 as example of an editor extension

They idea is to create an extension in `spyder/widgets/sourcecode/extensions/` with a class that Inherits o `EditorExtension` (api.editor.extension), create custom methods and connect them in the `on_uninstall` method, and the  add  them to the `Codeeditor.editor_extensions`

```
class MultiCursorsEditorFeature(EditorFeature):
     def _on_key_pressed():
          ....
     def _on_mouse_released():

     def `on_uninstall():
           self._editor.key_pressed.connect(self._on_key_pressed)
           self._editor.sig_mouse_pressed.connect(self._on_mouse_pressed)

class CodeEditor():

    def __init__(self):
         editor_extensions = EditorExtensionsManager
         editor_extensions.register(MultiCursorEditorsFeature)
```

There is already some signals for key press, and paint event,  ~I also added a signal for mouse pressed.~ Maybe there is missing a signal for timerEvent.

ping @jnsebgosselin


spyder/widgets/sourcecode/codeeditor.py
```diff
@@ -247,6 +249,8 @@ class CodeEditor(TextEditBaseWidget):
     #: Signal emitted when a key is pressed
     key_pressed = Signal(QKeyEvent)
 
+    sig_mouse_pressed = Signal(QMouseEvent)
```
ccordoba12: This doesn't seem necessary as part of this PR, right?

ccordoba12:
I really like this, it's very neat! I just have one suggestion: could we add as part of this PR an `EditorExtension` that involves some kind of text manipulation in `CodeEditor`? It can be something very simple, nothing fancy.

That way it'd be easier to understand how to use extensions for the rest of us ;-)

rlaverde:
@ccordoba12 This is ready, I moved close quote logic to an editor extension, and write some tests, although I didn't found any errors with that code.

Should I enable close quote by default?

spyder/widgets/sourcecode/editorextensions/closequotes.py
```diff
@@ -0,0 +1,106 @@
...
+    text = text.replace('\\"', '')
+    if text.count('"') % 2:
+        return '"'
+    elif text.count("'") % 2:
+        return "'"
+    else:
+        return ''
+
+
+class QuoteEditorExtension(EditorExtension):
```
ccordoba12: Let's call this `CloseQuoteExtension`

ccordoba12:
> although I didn't found any errors with that code.

I think issue #877 has several examples. Did you look at it?

> Should I enable close quote by default?

If tests pass, why not? ;-)

spyder/widgets/sourcecode/editorextensions/manager.py
```diff
@@ -16,17 +17,16 @@
...
-    Manages the list of editor extensions of the CodeEdit widget.
-    """
+    """Manages the list of editor extensions of the CodeEdit widget."""
 
     def __init__(self, editor):
+        """Initialize and add a reference to the editor."""
         super(EditorExtensionsManager, self).__init__(editor)
         self._extensions = {}
 
     def append(self, extension):
```
goanpeca: Add/remove are the antonyms, not append/remove

spyder/widgets/sourcecode/editorextensions/closequotes.py
```diff
@@ -3,9 +3,7 @@
 # Copyright © Spyder Project Contributors
 # Licensed under the terms of the MIT License
 # (see spyder/__init__.py for details)
-"""
-This module contains the close quotes editor extension
-"""
+"""This module contains the close quotes editor extension."""
```
goanpeca: We could just use `extensions` folder? like `spyder/widgets/sourcecode/extensions/closequotes.py`

ccordoba12:
I agree with @goanpeca to use only the `sourcecode/extensions` for this directory instead of `sourcecode/editorextensions`.

rlaverde:
I agree with @goanpeca to use only the sourcecode/extensions for this directory instead of sourcecode/editorextensions.

Yes, me too (at first I thought that editorextensions was more specific, anyway they are into the `sourcecode` folder, or the `plugins/editor`, after split-plugins, so that long name is redundant)

spyder/widgets/sourcecode/editorextensions/manager.py
rlaverde: I think append is more list-pythonic, should I change it for add?

goanpeca: Yes it is, but append even your docsting uses ADD and not APPEND, so you can see why append is not really a good choice, independent of the structure used to store extensions (dict, list, set or whatever). So either we append/prepend or add/remove, insert/pop, but not a mix of those. So yes, please change to `add`

rlaverde:
> I think issue #877 has several examples. Did you look at it?

I implement https://github.com/spyder-ide/spyder/issues/877#issuecomment-74600265 although I change the behavior, It'll just keep the text selected allowing to add as many quotes as you like.

The parenthesis are managed by other logic, so integrate it with this extension could break some of their functionality

ccordoba12:
> I implement #877 (comment) although I change the behavior, It'll just keep the text selected allowing to add as many quotes as you like.

Ok, no problem. If more issues appear in the future, now we can fix and add tests for this quite easily.

spyder/widgets/sourcecode/editorextensions/manager.py
goanpeca: @rlaverde ?

rlaverde: I didn't see your response because It was collapsed, I'm going to change it for `add`

PR #5011, opened 2017-08-19
PR: Add all possible string prefixes for syntax highlighting

### LINKED ISSUE #4884: prefix 'b' of bytes is not highlighting in the editor window
## Description

when inputting the bytes-type string in the editor window, the prefix 'b' is not highlighting with the same color of string.

* Spyder Version: 3.2.0 
* Python Version: 3.6.1

## Dependencies
```
sympy >=0.7.3    :  1.0 (OK)

```

Prikers:
Fixes #4884 

----

As suggested in #4884 the string prefixes should be highlighted with the same color as the string itself. The list of all possible prefixes is given [here](https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals) for Python 3 and [here](https://docs.python.org/2/reference/lexical_analysis.html#string-literals) for Python 2. 
image

I have added a small test as well but not sure it is actually necessary. Is it?

### DIFF
--- a/spyder/utils/syntaxhighlighters.py
+++ b/spyder/utils/syntaxhighlighters.py
@@ -14,7 +14,6 @@ from __future__ import print_function
 import keyword
 import os
 import re
-import sys
 import weakref
 
 # Third party imports
@@ -27,7 +26,7 @@ from qtpy.QtWidgets import QApplication
 from spyder import dependencies
 from spyder.config.base import _
 from spyder.config.main import CONF
-from spyder.py3compat import builtins, is_text_string, to_text_string
+from spyder.py3compat import builtins, is_text_string, to_text_string, PY3
 from spyder.utils.sourcecode import CELL_LANGUAGES
 from spyder.utils.editor import TextBlockHelper as tbh
 from spyder.utils.workers import WorkerManager
@@ -341,7 +340,7 @@ def make_python_patterns(additional_keywords=[], additional_builtins=[]):
                   r"\b[+-]?0[oO][0-7]+[lL]?\b",
                   r"\b[+-]?0[bB][01]+[lL]?\b",
                   r"\b[+-]?[0-9]+(?:\.[0-9]+)?(?:[eE][+-]?[0-9]+)?[jJ]?\b"])
-    if sys.version_info[0] == 3:
+    if PY3:
         prefix = "r|u|R|U|f|F|fr|Fr|fR|FR|rf|rF|Rf|RF|b|B|br|Br|bR|BR|rb|rB|Rb|RB"
     else:
         prefix = "r|u|ur|R|U|UR|Ur|uR|b|B|br|Br|bR|BR"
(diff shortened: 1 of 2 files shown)


spyder/utils/syntaxhighlighters.py
```diff
@@ -340,14 +341,20 @@ def make_python_patterns(additional_keywords=[], additional_builtins=[]):
...
                   r"\b[+-]?[0-9]+(?:\.[0-9]+)?(?:[eE][+-]?[0-9]+)?[jJ]?\b"])
-    sqstring =     r"(\b[rRuU])?'[^'\\\n]*(\\.[^'\\\n]*)*'?"
-    dqstring =     r'(\b[rRuU])?"[^"\\\n]*(\\.[^"\\\n]*)*"?'
-    uf_sqstring =  r"(\b[rRuU])?'[^'\\\n]*(\\.[^'\\\n]*)*(\\)$(?!')$"
-    uf_dqstring =  r'(\b[rRuU])?"[^"\\\n]*(\\.[^"\\\n]*)*(\\)$(?!")$'
-    sq3string =    r"(\b[rRuU])?'''[^'\\]*((\\.|'(?!''))[^'\\]*)*(''')?"
-    dq3string =    r'(\b[rRuU])?"""[^"\\]*((\\.|"(?!""))[^"\\]*)*(""")?'
-    uf_sq3string = r"(\b[rRuU])?'''[^'\\]*((\\.|'(?!''))[^'\\]*)*(\\)?(?!''')$"
-    uf_dq3string = r'(\b[rRuU])?"""[^"\\]*((\\.|"(?!""))[^"\\]*)*(\\)?(?!""")$'
+    if sys.version_info[0] == 3:
```
rlaverde: Please use `PY3`

`from spyder.py3compat import PY3`

spyder/utils/tests/test_syntaxhighlighters.py
```diff
@@ -47,6 +47,26 @@ def test_HtmlSH_unclosed_commend():
     compare_formats(doc.firstBlock().layout().additionalFormats(), res, sh)
 
 
+def test_python_string_prefix():
```
rlaverde: This test is failing in python2, I think is because the f-string.

rlaverde:
> I have added a small test as well but not sure it is actually necessary. Is it?

Yes, tests are never unnecessary 😉

Prikers:
There are a some pep8 issues remaining but I am not sure I should correct them:

- Multiple spaces after `=` operator were already there and they actually make this part more readable
- Long line (82 > 79) cannot be easily broken down without altering the readability

As for the tests, I am not sure why it is failing on Python 3.5, PyQT4...

ccordoba12:
> There are a some pep8 issues remaining but I am not sure I should correct them

Don't worry about them.

> As for the tests, I am not sure why it is failing on Python 3.5, PyQT4

The failure is unrelated to your work, don't worry about it.

----

@Prikers, thanks a lot for your contribution. This is going to be part of Spyder 4, to be released next year.

PR #5013, opened 2017-08-20
PR: Add test with non-ascii directory for consoles stderr file

### LINKED ISSUE #4836: Kernels fail to start because they can't create stderr files in non-ascii accounts on Windows
## Description

1. 
2. 
3. 

  File "C:\Anaconda2\lib\site-packages\spyder\plugins\__init__.py", line 511, in <lambda>
    toggled=lambda checked: self.toggle_view(checked),
  File "C:\Anaconda2\lib\site-packages\spyder\plugins\ipythonconsole.py", line 695, in toggle_view
    self.create_new_client(give_focus=False)
  File "C:\Anaconda2\lib\site-packages\spyder\plugins\ipythonconsole.py", line 974, in create_new_client
    self.connect_client_to_kernel(client)
  File "C:\Anaconda2\lib\site-packages\spyder\plugins\ipythonconsole.py", line 995, in connect_client_to_kernel
    stderr_file)
  File "C:\Anaconda2\lib\site-packages\spyder\plugins\ipythonconsole.py", line 1384, in create_kernel_manager_and_kernel_client
    kernel_manager.start_kernel(stderr=stderr)
  File "C:\Anaconda2\lib\site-packages\jupyter_client\manager.py", line 243, in start_kernel
    **kw)
  File "C:\Anaconda2\lib\site-packages\jupyter_client\manager.py", line 189, in _launch_kernel
    return launch_kernel(kernel_cmd, **kw)
  File "C:\Anaconda2\lib\site-packages\jupyter_client\launcher.py", line 123, in launch_kernel
    proc = Popen(cmd, **kwargs)
  File "C:\Anaconda2\lib\subprocess.py", line 390, in __init__
    errread, errwrite)
  File "C:\Anaconda2\lib\sub

dalthviz:
Fixes #4836

### DIFF
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -619,8 +619,9 @@ class IPythonConsole(SpyderPluginWidget):
 
         # Create temp dir on testing to save kernel errors
         if self.testing:
-            if not osp.isdir(programs.TEMPDIR):
-                os.mkdir(programs.TEMPDIR)
+            if not osp.isdir(osp.join(programs.TEMPDIR, '測試', 'اختبار')):
+                os.mkdir(osp.join(programs.TEMPDIR, '測試', 'اختبار'))
+
 
         layout = QVBoxLayout()
         self.tabwidget = Tabs(self, self.menu_actions, rename_tabs=True,
@@ -932,7 +933,7 @@ class IPythonConsole(SpyderPluginWidget):
     @Slot(bool)
     @Slot(str)
     @Slot(bool, str)
-    def create_new_client(self, give_focus=True, filename=''):
+    def create_new_client(self, give_focus=True, filename='', testing=False):
         """Create a new client"""
         self.master_clients += 1
         client_id = dict(int_id=to_text_string(self.master_clients),
@@ -944,7 +945,7 @@ class IPythonConsole(SpyderPluginWidget):
                               additional_options=self.additional_options(),
                               interpreter_versions=self.interpreter_versions(),
                               connection_file=cf,
-                              menu_actions=self.menu_actions)
+                              menu_actions=self.menu_actions, testing=testing)
         self.add_tab(client, name=client.get_name(), filename=filename)
 
         if cf is None:
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -619,7 +619,7 @@ class IPythonConsole(SpyderPluginWidget):
 
         # Create temp dir on testing to save kernel errors
         if self.testing:
-            if not osp.isdir(osp.join(programs.TEMPDIR, '測試', 'اختبار')):
+            if not osp.isdir(osp.join(programs.TEMPDIR, u'測試', u'اختبار')):
                 os.makedirs(osp.join(programs.TEMPDIR, '測試', 'اختبار'))
 
 
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -620,7 +620,7 @@ class IPythonConsole(SpyderPluginWidget):
         # Create temp dir on testing to save kernel errors
         if self.testing:
             if not osp.isdir(osp.join(programs.TEMPDIR, u'測試', u'اختبار')):
-                os.makedirs(osp.join(programs.TEMPDIR, '測試', 'اختبار'))
+                os.makedirs(osp.join(programs.TEMPDIR, u'測試', u'اختبار'))
 
 
         layout = QVBoxLayout()
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -620,7 +620,7 @@ class IPythonConsole(SpyderPluginWidget):
         # Create temp dir on testing to save kernel errors
         if self.testing:
             if not osp.isdir(osp.join(programs.TEMPDIR, '測試', 'اختبار')):
-                os.mkdir(osp.join(programs.TEMPDIR, '測試', 'اختبار'))
+                os.makedirs(osp.join(programs.TEMPDIR, '測試', 'اختبار'))
 
 
         layout = QVBoxLayout()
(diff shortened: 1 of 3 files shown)


dalthviz:
@ccordoba12 should I do something else with this one?

spyder/plugins/ipythonconsole.py
```diff
@@ -619,8 +619,9 @@ def __init__(self, parent, testing=False):
 
         # Create temp dir on testing to save kernel errors
         if self.testing:
-            if not osp.isdir(programs.TEMPDIR):
-                os.mkdir(programs.TEMPDIR)
+            if not osp.isdir(osp.join(programs.TEMPDIR, u'測試', u'اختبار')):
+                os.makedirs(osp.join(programs.TEMPDIR, u'測試', u'اختبار'))
```
ccordoba12: @dalthviz, I don't like this test directory to be part of our sourcecode.

Please add a new parameter to `__init__` called `test_dir=TEMPDIR`. With that this should look like:

```python
if self.testing:
    if not osp.isdir(test_dir):
        os.makedirs(test_dir)
```

Then in our tests you can change that parameter by this directory containing Chinese and Arabic characters.

spyder/widgets/ipythonconsole/client.py
```diff
@@ -163,7 +165,10 @@ def stderr_file(self):
         """Filename to save kernel stderr output."""
         if self.connection_file is not None:
             stderr_file = self.kernel_id + '.stderr'
-            stderr_file = osp.join(TEMPDIR, stderr_file)
+            if not self.testing:
+                stderr_file = osp.join(TEMPDIR, stderr_file)
+            else:
+                stderr_file = osp.join(TEMPDIR, u'測試', u'اختبار', stderr_file)
```
ccordoba12: Please also find a solution similar to the one I posted above to avoid putting this temp directory as part of our source code.

spyder/plugins/tests/test_ipythonconsole.py
```diff
@@ -74,9 +74,45 @@ def close_console():
     return console
 
 
+@pytest.fixture
+def ipyconsole_stderr(request):
```
ccordoba12: @dalthviz, are you sure we need an extra fixture for this? Is it not possible to add an extra parameter to the previous fixture?

spyder/plugins/ipythonconsole.py
```diff
@@ -590,7 +590,7 @@ class IPythonConsole(SpyderPluginWidget):
     focus_changed = Signal()
     edit_goto = Signal((str, int, str), (str, int, str, bool))
 
-    def __init__(self, parent, testing=False):
+    def __init__(self, parent, testing=False, test_dir=programs.TEMPDIR):
```
ccordoba12: Please use here only `TEMPDIR`, so import it from `programs` with

    from programs import TEMPDIR

spyder/plugins/ipythonconsole.py
```diff
@@ -932,7 +933,8 @@ def write_to_stdin(self, line):
     @Slot(bool)
     @Slot(str)
     @Slot(bool, str)
-    def create_new_client(self, give_focus=True, filename=''):
+    def create_new_client(self, give_focus=True, filename='',
+                          stderr_dir=programs.TEMPDIR):
```
ccordoba12: Please don't add a new parameter here because we'd need to also add new `Slot`'s

spyder/plugins/ipythonconsole.py
```diff
@@ -944,7 +946,8 @@ def create_new_client(self, give_focus=True, filename=''):
                               additional_options=self.additional_options(),
                               interpreter_versions=self.interpreter_versions(),
                               connection_file=cf,
-                              menu_actions=self.menu_actions)
+                              menu_actions=self.menu_actions,
+                              stderr_dir=stderr_dir)
```
ccordoba12: Instead of doing this, please change it to

    if self.testing:
        client.stderr_dir = self.test_dir

spyder/plugins/ipythonconsole.py
ccordoba12: We also need to add

    self.test_dir = test_dir

spyder/widgets/ipythonconsole/client.py
```diff
@@ -95,7 +95,8 @@ def __init__(self, plugin, id_,
                  additional_options, interpreter_versions,
                  connection_file=None, hostname=None,
                  menu_actions=None, slave=False,
-                 external_kernel=False, given_name=None):
+                 external_kernel=False, given_name=None,
+                 stderr_dir=TEMPDIR):
```
ccordoba12: Please remove this

spyder/widgets/ipythonconsole/client.py
```diff
@@ -113,6 +114,7 @@ def __init__(self, plugin, id_,
         self.stop_icon = ima.icon('stop')
         self.history = []
         self.allow_rename = True
+        self.stderr_dir = stderr_dir
```
ccordoba12: This needs to be

    # Used only for testing
    self.stderr_dir = None

spyder/widgets/ipythonconsole/client.py
```diff
@@ -163,7 +165,7 @@ def stderr_file(self):
         """Filename to save kernel stderr output."""
         if self.connection_file is not None:
             stderr_file = self.kernel_id + '.stderr'
-            stderr_file = osp.join(TEMPDIR, stderr_file)
+            stderr_file = osp.join(self.stderr_dir, stderr_file)
```
ccordoba12: Finally, this needs to be

    if self.stderr_dir is not None:
        stderr_file = osp.join(self.stderr_dir, stderr_file)
    else:
        stderr_file = osp.join(TEMPDIR, stderr_file)

spyder/plugins/tests/test_ipythonconsole.py
dalthviz: Yep, you are right :+1: no need of a extra fixture

PR #5028, opened 2017-08-22
PR: Make value_to_display more robust

### LINKED ISSUE #4221: Spyder stops executing commands if Variable explorer shows complicated objects
## Description
I often found that spyder sometimes becomes unusable because some complicated object is present in the variable explorer
even running 1+1 just hangs and takes ages to be executed
Here is a working example

run the following code:

```python
from collections import defaultdict
import numpy as np
bug = defaultdict(list)
for i in range(50000):
    a = {j:np.random.rand(10) for j in range(10)}
    bug[i] = a
```
    
Now running even the simplest stuff in the ipython console is unbearably slow.
changing `bug` to be of type dict solves it

I dont have autorefresh enabled (I didn't change any options in th Varibale explorer)

My guess is that the code that creates the string representation for variable explorer doesn't have an optimization for very large defaultdict.
This makes defaultdict unusable for me.

* Spyder Version: 3.1.3 
* Python Version: 2.7.12

## Dependencies
```
jedi =0.9.0      :  0.9.0 (OK)
sympy >=0.7.3    :  1.0 (OK)

```

ccordoba12:
Fixes #4221 

----

This PR

1. Improves how any object is displayed in the Variable Explorer. Now objects will be shown as `foo object of bar module` (except for know objects, like Dataframes).
2. Doesn't try to display subclasses of list, dict and tuple, to avoid things like issue #4221 (which was using a generalization of a dict that doesn't have an optimized `repr`).
3. Applies `value_to_display` recursively to the elements of lists, dicts and tuples. This prevents freezes in the Variable Explorer when people add objects with huge `repr`'s as elements of those collection types.

Screenshots:

*Before*

seleccion_001

*After*

seleccion_002

----

TODO:

- [x] Tests

[skip pep8]

### DIFF
--- a/spyder/widgets/variableexplorer/tests/test_utils.py
+++ b/spyder/widgets/variableexplorer/tests/test_utils.py
@@ -7,12 +7,30 @@
 Tests for utils.py
 """
 
+from collections import defaultdict
+
 # Third party imports
+import numpy as np
+import pandas as pd
 import pytest
 
 # Local imports
 from spyder.config.base import get_supported_types
-from spyder.widgets.variableexplorer.utils import sort_against, is_supported
+from spyder.widgets.variableexplorer.utils import (sort_against,
+    is_supported, value_to_display)
+
+
+def generate_complex_object():
+    """Taken from issue #4221."""
+    bug = defaultdict(list)
+    for i in range(50000):
+        a = {j:np.random.rand(10) for j in range(10)}
+        bug[i] = a
+    return bug
+
+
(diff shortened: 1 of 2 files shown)


goanpeca:
@ccordoba12 could you add a scrrenshot of how it looks?

ccordoba12:
@goanpeca, I updated the description of this PR and add before/after screenshots.

@jitseniesen, what do you think about this one?

PR #5045, opened 2017-08-24
PR: Allow values with '%' in config system

### LINKED ISSUE #4921: IPython: Run Code at Startup - Magics
## Description

1. Start Spyder
2. Open Preferences
3. Select IPython console preferences
4. Select Startup Tab
5. Enter `%load_ext autoreload, %autoreload 2`
6. Click Apply

* Expected: Allow these valid IPython magics to be stored in the config and loaded on Console startup
* Actual: The commands cause the error below

  File "C:\Anaconda3\lib\site-packages\spyder\plugins\configdialog.py", line 221, in button_clicked
    configpage.apply_changes()
  File "C:\Anaconda3\lib\site-packages\spyder\plugins\configdialog.py", line 100, in apply_changes
    self.save_to_conf()
  File "C:\Anaconda3\lib\site-packages\spyder\plugins\configdialog.py", line 408, in save_to_conf
    self.set_option(option, to_text_string(lineedit.text()))
  File "C:\Anaconda3\lib\site-packages\spyder\plugins\__init__.py", line 437, in set_option
    CONF.set(self.CONF_SECTION, str(option), value)
  File "C:\Anaconda3\lib\site-packages\spyder\config\user.py", line 443, in set
    self._set(section, option, value, verbose)
  File "C:\Anaconda3\lib\site-packages\spyder\config\user.py", line 86, in _set
    cp.ConfigParser.set(self, section, option, value)
  File "C:\Anaconda3\lib\configparser.py", line 1193, in set
    super().set(section, option, value)
  File "C:\Anaconda3\lib\configparser.py", line 894, in set
    value)
  File "C:\Anaconda3\lib\configparser.py", line 402, in before_set
    "position %d" % (value, tmp_value.find('%')))
ValueError: invalid interpolation syntax in '%load_ext autoreload, %a

**Work Around**
`from IPython import get_ipython, ipython = get_ipython(), ipython.magic("load_ext autoreload"), ipython.magic("autoreload 2")`

dalthviz:
Fixes #4921

### DIFF
--- a/spyder/config/tests/test_user.py
+++ b/spyder/config/tests/test_user.py
@@ -28,6 +28,13 @@ def userconfig(tmpdir, monkeypatch):
     return UserConfig('foo', defaults={}, subfolder=True,
                       version='1.0.0', raw_mode=True)
 
+
+def test_userconfig_set_percentage_string(userconfig):
+    """Test to set an option with a '%'."""
+    userconfig.set('section', 'option', '%value')
+    assert userconfig.get('section', 'option') == '%value'
+
+
 def test_userconfig_get_string_from_inifile(userconfig):
     assert userconfig.get('section', 'option') == 'value'
 
--- a/spyder/config/user.py
+++ b/spyder/config/user.py
@@ -48,7 +48,10 @@ class DefaultsConfig(cp.ConfigParser):
     UserConfig
     """
     def __init__(self, name, subfolder):
-        cp.ConfigParser.__init__(self)
+        if PY2:
+            cp.ConfigParser.__init__(self)
+        else:
+            cp.ConfigParser.__init__(self, interpolation=None)
         self.name = name
         self.subfolder = subfolder
 


jitseniesen:
I am not sure this actually fixes #4921. It does prevent the error when the user enters `%load_ext autoreload, %autoreload 2` by removing the `%` characters when saving the config, but when the config value is loaded the `%` characters are not added back in so the value is now `load_ext autoreload, autoreload 2`.

dalthviz:
Yep, you are right @jitseniesen, it doesn't restore the `%`, maybe should I try to replace the `%` with something else and then restore it by replacing again? or maybe created a new line edit only for magics and add the `%` to each entry? What do you think @ccordoba12 ?

Although with or without the `%` the magic is executed.

ccordoba12:
>  it doesn't restore the %, maybe should I try to replace the % with something else and then restore it by replacing again?

I agree with this idea. I think it's much simpler than adding a lineedit only for magics.

@dalthviz, please implement this and add a test for it.

jitseniesen:
An alternative would be to find out why our config system does not allow for `%` to be saved. I remember I had a quick look once and the fact that we use `RawConfigParser` should allow us to use `%` characters. However, at one point in the code this does not work; I forget where and I could not immediately find out why.

This may be more work, but replacing `%` with another character like `¿` caused problems if the user uses the second character somewhere in the configuration (admittedly pretty unlikely). It might also cause issues with how to encode that character, especially in Python 2, though the tests pass so maybe it's okay (or maybe you did not add a test).

dalthviz:
Totally right @jitseniesen. Testing a little bit and reading the docs the problem was in Python 3 since, for `ConfigParser`, there is a default value for a keyword argument `interpolation` that is [BasicInterpolation](https://docs.python.org/3/library/configparser.html#interpolation-of-values)

ccordoba12:
The solution is sweet and simple. Thanks @jitseniesen for your suggestion!

jitseniesen:
It's @dalthviz who found the solution; I could not find it when I looked for it.

dalthviz:
Thanks for the comments @jitseniesen, they really helped us to gain more perspective in the problem and helped me to find this nice solution!

PR #4756, opened 2017-07-17
PR: Add checkboxes to PYTHONPATH manager to enable/disable paths

jnsebgosselin:
Fixes #1105

----

This is an initial implementation for revision. This was developed in Windows 7 64bit with PyQt4.

path_mngr_checkbox

PR #5051, opened 2017-08-24
PR: Make inactive paths to be ignored when Spyder's path list is synchronized with PYTHONPATH on Windows

### LINKED ISSUE #1105: Add the possibility to disable entries in the PYTHONPATH manager
_From [eng...@gmail.com](https://code.google.com/u/113881073069619832939/) on 2012-06-29T01:58:32Z_

It would be really great if it was possible to disable entries on the pythonpath instead of removing them.

I was thinking about something with a check box next to the different entries and then when unchecked the entry becomes grey and does not go on the pythonpath.

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=1105_

jnsebgosselin:
This PR is to expand the new feature (PR: https://github.com/spyder-ide/spyder/pull/4756) that allowed to check and uncheck paths in the pathmanager to quickly add or remove them from the Spyder's path list.

With this PR, paths that are unchecked won't be added to the PYTHONPATH environment variable when synchronizing with  Spyder's path list. Note that the "synchonize Spyder's PYTHONPATH with PYTHONPATH environment variable" feature is valid only for Windows environment.

Also, added a test for the synchonize feature.

image

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,7 +13,7 @@ environment:
       rope pyflakes sphinx pygments pylint pycodestyle psutil nbconvert
       qtawesome pickleshare pyzmq chardet mock pandas pytest
       pytest-cov numpydoc scipy pillow qtconsole matplotlib jedi
-    PIP_DEPENDENCIES: "pytest-qt pytest-timeout flaky"
+    PIP_DEPENDENCIES: "pytest-qt pytest-mock pytest-timeout flaky"
 
   matrix:
     - PYTHON_VERSION: "2.7"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,7 +12,7 @@ environment:
     CONDA_DEPENDENCIES: >
       rope pyflakes sphinx pygments pylint pycodestyle psutil nbconvert
       qtawesome pickleshare pyzmq chardet mock pandas pytest
-      pytest-cov numpydoc scipy pillow qtconsole matplotlib jedi
+      pytest-cov numpydoc scipy pillow qtconsole matplotlib jedi pywin32
     PIP_DEPENDENCIES: "pytest-qt pytest-mock pytest-timeout flaky"
 
   matrix:
--- a/continuous_integration/circle/install.sh
+++ b/continuous_integration/circle/install.sh
@@ -4,7 +4,7 @@ export CONDA_DEPENDENCIES_FLAGS="--quiet"
 export CONDA_DEPENDENCIES="rope pyflakes sphinx pygments pylint psutil nbconvert \
                            qtawesome pickleshare qtpy pyzmq chardet mock nomkl pandas \
                            pytest pytest-cov numpydoc scipy cython pillow"
-export PIP_DEPENDENCIES="coveralls pytest-qt 'pytest-mock' pytest-xvfb flaky jedi pycodestyle"
+export PIP_DEPENDENCIES="coveralls pytest-qt pytest-mock pytest-xvfb flaky jedi pycodestyle"
 
 # Download and install miniconda and conda/pip dependencies
 # with astropy helpers
--- a/continuous_integration/circle/install.sh
+++ b/continuous_integration/circle/install.sh
@@ -4,7 +4,7 @@ export CONDA_DEPENDENCIES_FLAGS="--quiet"
 export CONDA_DEPENDENCIES="rope pyflakes sphinx pygments pylint psutil nbconvert \
                            qtawesome pickleshare qtpy pyzmq chardet mock nomkl pandas \
                            pytest pytest-cov numpydoc scipy cython pillow"
-export PIP_DEPENDENCIES="coveralls pytest-qt pytest-xvfb flaky jedi pycodestyle"
+export PIP_DEPENDENCIES="coveralls pytest-qt 'pytest-mock' pytest-xvfb flaky jedi pycodestyle"
 
 # Download and install miniconda and conda/pip dependencies
 # with astropy helpers
--- a/continuous_integration/travis/install.sh
+++ b/continuous_integration/travis/install.sh
@@ -10,7 +10,7 @@ else
     export CONDA_DEPENDENCIES="rope pyflakes sphinx pygments pylint psutil nbconvert \
                                qtawesome pickleshare qtpy pyzmq chardet mock nomkl pandas \
                                pytest pytest-cov numpydoc scipy cython pillow jedi pycodestyle"
-    export PIP_DEPENDENCIES="coveralls pytest-qt pytest-timeout flaky"
+    export PIP_DEPENDENCIES="coveralls pytest-qt 'pytest-mock' pytest-timeout flaky"
 fi
 
 
--- a/continuous_integration/travis/install.sh
+++ b/continuous_integration/travis/install.sh
@@ -10,7 +10,7 @@ else
     export CONDA_DEPENDENCIES="rope pyflakes sphinx pygments pylint psutil nbconvert \
                                qtawesome pickleshare qtpy pyzmq chardet mock nomkl pandas \
                                pytest pytest-cov numpydoc scipy cython pillow jedi pycodestyle"
-    export PIP_DEPENDENCIES="coveralls pytest-qt 'pytest-mock' pytest-timeout flaky"
+    export PIP_DEPENDENCIES="coveralls pytest-qt pytest-mock pytest-timeout flaky"
 fi
 
 
--- a/spyder/widgets/tests/test_pathmanager.py
+++ b/spyder/widgets/tests/test_pathmanager.py
@@ -20,10 +20,10 @@ except ImportError:
 import pytest
 from qtpy import PYQT4
 from qtpy.QtCore import Qt
-from qtpy.QtWidgets import QMessageBox
 
 # Local imports
 from spyder.py3compat import PY3
+from spyder.widgets import pathmanager as pathmanager_mod
 from spyder.widgets.pathmanager import PathManager
 
 
@@ -67,7 +67,8 @@ def test_check_uncheck_path(qtbot):
     assert pathmanager.not_active_pathlist == []
 
 
-@patch.object(QMessageBox, 'question', return_value=QMessageBox.Yes)
+@patch.object(pathmanager_mod.QMessageBox, 'question',
+              return_value=pathmanager_mod.QMessageBox.Yes)
 def test_synchronize_with_PYTHONPATH(qtbot):
     if os.name != 'nt':
         return
@@ -104,3 +105,4 @@ def test_synchronize_with_PYTHONPATH(qtbot):
 
 if __name__ == "__main__":
     pytest.main([os.path.basename(__file__)])
+    # pytest.main()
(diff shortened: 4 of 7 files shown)


jnsebgosselin:
@goanpeca Hey, what can I do to solve the "import" error I get in Travis? Thank you!

goanpeca:
So that method seems to be... windows only or ?

https://github.com/spyder-ide/spyder/blob/3.x/spyder/utils/environ.py#L58

jnsebgosselin:
Ah, you are right I think `get_user_env` and `set_user_env` are only for Windows. What happens in Linux then when the [synchronize ](https://github.com/spyder-ide/spyder/blob/master/spyder/widgets/pathmanager.py#L145)slot is called? No error is raised then when `get_user_env` and `set_user_env` are called?

jnsebgosselin:
Bah, got it, the button doesn't even exist in Linux (https://github.com/spyder-ide/spyder/blob/master/spyder/widgets/pathmanager.py#L134)! So my test is useless... Is there anything I can do about it or I just delete?

goanpeca:
@jnsebgosselin well... skip tests on mac and linux? and leave on wind?

jnsebgosselin:
Ok I will do this, thank you very much @goanpeca

ccordoba12:
> Note that this feature is valid only for Windows environment.

What happens with Linux and macOS?

jnsebgosselin:
@ccordoba12 Nothing. The button `synchronize` is not generated in Linux in MacOS, so these function are never supposed to be called in these environments.

If they are called programmatically on those system however, an `ImportError` is raised.

continuous_integration/circle/install.sh
```diff
@@ -4,7 +4,7 @@ export CONDA_DEPENDENCIES_FLAGS="--quiet"
 export CONDA_DEPENDENCIES="rope pyflakes sphinx pygments pylint psutil nbconvert \
                            qtawesome pickleshare qtpy pyzmq chardet mock nomkl pandas \
                            pytest pytest-cov numpydoc scipy cython pillow"
-export PIP_DEPENDENCIES="coveralls pytest-qt pytest-xvfb flaky jedi pycodestyle"
+export PIP_DEPENDENCIES="coveralls pytest-qt 'pytest-mock' pytest-xvfb flaky jedi pycodestyle"
```
ccordoba12: Please remove the quotes around pytest-mock here.

continuous_integration/travis/install.sh
```diff
@@ -10,7 +10,7 @@ else
     export CONDA_DEPENDENCIES="rope pyflakes sphinx pygments pylint psutil nbconvert \
                                qtawesome pickleshare qtpy pyzmq chardet mock nomkl pandas \
                                pytest pytest-cov numpydoc scipy cython pillow jedi pycodestyle"
-    export PIP_DEPENDENCIES="coveralls pytest-qt pytest-timeout flaky"
+    export PIP_DEPENDENCIES="coveralls pytest-qt 'pytest-mock' pytest-timeout flaky"
```
ccordoba12: Also here.

ccordoba12:
> The button synchronize is not generated in Linux in MacOS, so this function are never supposed to be called in these environments.

Ok, but I meant if on Linux and macOS inactive are removed from Spyder's PYTHONPATH or not :-)

jnsebgosselin:
> Ok, but I meant if on Linux and macOS inactive are removed from Spyder's PYTHONPATH or not :-)

Yes, the inactive paths on Linux and macOS should be removed from Spyder's PYTHONPATH . This was covered in the previous PR and is covered by this [condition ](https://github.com/spyder-ide/spyder/blob/master/spyder/app/mainwindow.py#L2492) in the [mainwindow](https://github.com/spyder-ide/spyder/blob/master/spyder/app/mainwindow.py).

This PR only addresses the  "synchonize Spyder's PYTHONPATH with environment variable PYTHONPATH" feature.

jnsebgosselin:
> Ok, but I meant if on Linux and macOS inactive are removed from Spyder's PYTHONPATH or not :-)

@ccordoba12  Maybe a test should be added to `mainwindow` to test it is indeed the case?

jnsebgosselin:
The test on AppVeyor is failing because it timed out due to the fact that the mocking doesn't seem to work on the Dialog. I'm trying to understand why. The test is running alright on my setup. If anyone could give me any advice, I'll be grateful.

ccordoba12:
To test a dialog you need to define a function to detect it among all Spyder widgets and interact with it inside that function.

For example, please see `open_client_from_connection_info` and its uses inside `plugins/tests/test_ipythonconsole.py`.

goanpeca:
If the test is ysing the main window.. then it is better to store the dialog in an instance variable so that it xan be used without searching for it

jnsebgosselin:
@ccordoba12 @goanpeca Thank you very much guys! The thing is though, I don't want to test the dialog per se, that is why I'm mocking it and patching its `question` method [here ](https://github.com/spyder-ide/spyder/pull/5051/files#diff-cf06a23e9b1a7b7f4bc6daae08430bddR82) and [here](https://github.com/spyder-ide/spyder/pull/5051/files#diff-cf06a23e9b1a7b7f4bc6daae08430bddR101). It is working fine locally on my machine. I think that when it runs in AppVeyor, the mocking fails because QMessageBox is defined somewhere else... or maybe (and it is quite probable lol) I'm missing something else.

Anyway, unless you guys tell me my approach is impossible for testing Spyder with AppVeyor, I'll continue to investigate a little bit and if it doesn't work, I'll go your route and interact myself with this modal dialog with some timer as I have seen it done in other Spyder tests.

jitseniesen:
@jnsebgosselin I know from experience that this kind of issue can be really frustrating and time consuming. I mocked a `QMessageBox` in [here](https://github.com/spyder-ide/spyder-unittest/blob/master/spyder_unittest/widgets/tests/test_unittestgui.py#L90) and I don't see a reason why your code should not work. I share your reluctance for the timer approach used in other Spyder tests.

My guess would be that your mocking works fine, but for some reason, a modal dialog box like an error message is shown on AppVeyor.

More generally, it's very inconvenient that if a modal dialog box is shown unexpectedly in a test, the test suite hangs and terminates after an hour while giving you no idea what's going on. I have wondered whether it would be possible to use [pytest-timeout](https://pypi.python.org/pypi/pytest-timeout) to get tests terminated after say a minute and ideally get some more information but I never found the time to investigate this.

jnsebgosselin:
@jitseniesen This is very good insights! I've never thought that the test may be failing because of something else than my mocking. I'll investigate that. Thank you very much.

andfoy:
@jnsebgosselin Otherwise, try to login to the Appveyor machine using RDP, that's how I debug the Windows tests at spyder-terminal https://www.google.com/url?q=https://www.appveyor.com/docs/how-to/rdp-to-build-worker/&sa=U&ved=0ahUKEwj8-4CH3_fVAhWB4CYKHSQxDmQQFggLMAA&usg=AFQjCNG3Vpyt05XqbdHDENd6MS_MhWl_fA

spyder/widgets/tests/test_pathmanager.py
```diff
@@ -60,6 +62,58 @@ def test_check_uncheck_path(qtbot):
...
+
+    # Restore PYTHONPATH to its original state
+    env['PYTHONPATH'] = original_pathlist
+    set_user_env(listdict2envdict(env))
+
+
 if __name__ == "__main__":
-    import os
     pytest.main([os.path.basename(__file__)])
+    # pytest.main()
```
ccordoba12: Please remove this comment if it's not necessary.

spyder/widgets/tests/test_pathmanager.py
```diff
@@ -60,6 +62,58 @@ def test_check_uncheck_path(qtbot):
     assert pathmanager.not_active_pathlist == []
 
 
+def test_synchronize_with_PYTHONPATH(qtbot, mocker):
+    if os.name != 'nt':
+        return
```
ccordoba12: The way to skip tests is by using `pytest.mark.skipif`, e.g.

https://github.com/spyder-ide/spyder/blob/3.x/spyder/app/tests/test_mainwindow.py#L136

ccordoba12:
Congrats @jnsebgosselin, you made your test to pass!

I left a couple of minor comments, then I'll merge.

jnsebgosselin:
@ccordoba12 well hehe, it is passing because I did not ran the entire test. I'm trying to find where in the test AppVeyor is blocking. Sorry for all this noise. I think I will try to install AppVeyor on my fork and test it there instead.

jnsebgosselin:
@andfoy Thank you. I'll test it tonight.

goanpeca:
@jnsebgosselin and devs I really encourage you to never use Static Helper methods that create and exec_ right away, like `QMessageBox.question`. They make testing super difficult and are a pain the `(_|_)`. @jnsebgosselin you should probably change the QMessageBox.question to a normal `self._dlg_question = QMessageBox()..... self._dlg_question.exec_()` and save it in an instance variable, that way you can use a timer and reliably target the right widget!

appveyor.yml
```diff
@@ -12,7 +12,7 @@ environment:
     CONDA_DEPENDENCIES: >
       rope pyflakes sphinx pygments pylint pycodestyle psutil nbconvert
       qtawesome pickleshare pyzmq chardet mock pandas pytest
-      pytest-cov numpydoc scipy pillow qtconsole matplotlib jedi
+      pytest-cov numpydoc scipy pillow qtconsole matplotlib jedi pywin32
```
andfoy: @jnsebgosselin Why are you adding pywin32 as a test dependency? According to its author, packages shouldn't depend on it: https://github.com/spyder-ide/pywinpty/issues/52#issuecomment-324114565

jnsebgosselin: Because it is needed for my test to pass. Otherwise, it hangs due to this [warning message dialog](https://github.com/spyder-ide/spyder/blob/master/spyder/utils/environ.py#L95). @jitseniesen was right, my mocking was working as expected, it was this other message that was blocking the test.

jnsebgosselin: So I don't know how I should handle that. If I can't use pywin32, then I think I can't really test this feature, unless I'm able to logout/login of the VM during the test.

ccordoba12: I don't have a problem with this, given that it's currently required by the @jnsebgosselin is trying to test.

Of course, we should look at removing this in the future.

andfoy: Let's see what @ccordoba12 thinks!

ccordoba12:
Ok, I think this is ready. @jnsebgosselin?

spyder/widgets/tests/test_pathmanager.py
```diff
@@ -60,6 +62,57 @@ def test_check_uncheck_path(qtbot):
     assert pathmanager.not_active_pathlist == []
 
 
+@pytest.mark.skipif(os.name != 'nt',
```
ccordoba12: Please also skip this test if `pywin32` is not present because it's not part of our listed dependencies.

jnsebgosselin: For this I just have to write: `os.name != 'nt' or not pywin32`?

ccordoba12: You can use `programs.is_module_installed` for this.

PR #5089, opened 2017-08-30
PR: Exit the debugger if we are about to enter to sitecustomize.py

### LINKED ISSUE #4680: Debug: "Run Current Line" past the end of a script doesn't exit the debugger
Using "Run Current Line (ctrl-f10)" past the end of a program doesn't exit, but starts debugging in sitecustomize.py

1. start debugging with this script:
```
#!/usr/bin/env python3
# -*- coding: utf-8 -*-
x = 1
```
2. press `ctrl-f10` 2 or 3 times (just past the end of the script) 
3. sitecostomize.py will popup

It should end debugging

* Spyder Version: 3.x branch
* Python Version: 3.6
* Qt Version: 5.6
* PyQt Version: 5.6
* Operating system:

## Dependencies

numpy >=1.7      :  1.11.3 (OK)
sympy >=0.7.3    :  1.0 (OK)

dalthviz:
Fixes #4680

### DIFF
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -846,11 +846,9 @@ def test_c_and_n_pdb_commands(main_window, qtbot):
     qtbot.keyClick(control, Qt.Key_Enter)
     qtbot.wait(500)
 
-    # Wait until the prompt appear
-    shell = main_window.ipyconsole.get_current_shellwidget()
-    control = shell._control
-    qtbot.waitUntil(lambda: shell._prompt_html is not None,
-                    timeout=SHELL_TIMEOUT)
+    # Assert that the prompt appear
+    shell.clear_console()
+    assert 'In [3]:' in control.toPlainText()
 
     # Remove breakpoint and close test file
     main_window.editor.clear_all_breakpoints()
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -830,6 +830,28 @@ def test_c_and_n_pdb_commands(main_window, qtbot):
     qtbot.wait(500)
     assert nsb.editor.model.rowCount() == 2
 
+    # Verify that doesn't go to sitecustomize.py with next and stops
+    # the debugging session.
+    qtbot.keyClicks(control, 'n')
+    qtbot.keyClick(control, Qt.Key_Enter)
+    qtbot.wait(500)
+
+    qtbot.keyClicks(control, 'n')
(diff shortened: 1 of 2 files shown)


spyder/app/tests/test_mainwindow.py
```diff
@@ -830,6 +830,28 @@ def test_c_and_n_pdb_commands(main_window, qtbot):
...
+
+    qtbot.keyClicks(control, 'n')
+    qtbot.keyClick(control, Qt.Key_Enter)
+    qtbot.wait(500)
+
+    # Wait until the prompt appear
+    shell = main_window.ipyconsole.get_current_shellwidget()
+    control = shell._control
+    qtbot.waitUntil(lambda: shell._prompt_html is not None,
+                    timeout=SHELL_TIMEOUT)
```
ccordoba12: This validation only works the first time the console is created. After that `shell._prompt_html`always has some value.

Instead I'd suggest that at this point we should clear the console and assert that the text present in the console is `In [x]:` where the `x` is the expected value of the input prompt, e.g. `In [2]:`

spyder/app/tests/test_mainwindow.py
```diff
@@ -830,6 +830,28 @@ def test_c_and_n_pdb_commands(main_window, qtbot):
...
+
+    assert nsb.editor.model.rowCount() == 3
+
+    qtbot.keyClicks(control, 'n')
+    qtbot.keyClick(control, Qt.Key_Enter)
+    qtbot.wait(500)
+
+    # Wait until the prompt appear
+    shell = main_window.ipyconsole.get_current_shellwidget()
+    control = shell._control
```
ccordoba12: These two lines are not needed because `shell` and `control` are defined above.

spyder/utils/site/sitecustomize.py
```diff
@@ -423,12 +423,15 @@ def user_return(self, frame, return_value):
 
 @monkeypatch_method(pdb.Pdb, 'Pdb')
 def interaction(self, frame, traceback):
-    self.setup(frame, traceback)
-    if self.send_initial_notification:
-        self.notify_spyder(frame)
-    self.print_stack_entry(self.stack[self.curindex])
-    self._cmdloop()
-    self.forget()
+    if "sitecustomize.py" not in frame.f_code.co_filename:
```
ccordoba12: Please verify that this works in Python 2

ccordoba12: Maybe this needs `u'sitecustomize.py'`

dalthviz: Yep it works :+1:

PR #5122, opened 2017-09-04
PR: Enable scrolling past the end of the document

### LINKED ISSUE #3345: Feature suggestion:  scroll over the end of file in the editor
Currently, if you have 100 lines of code, the scroller won't go over line 100, which means if you want to add more code, you are always typing at the bottom of the editor window (or you fanatically press enters until enough carriage returns are added to the end of file...). 

Please allow scrolling over to the eof, so that user can always work in the middle of screen.

P.S. the visual studio and VS Code has this feature.

Prikers:
Fixes #3345

Hi team, 
This is an attempt to solve #3345 and #2063: I added an option (default to `False`) to enable the user to scroll past the end of the document in order to have the last line of code on top of the screen (without having to added unnecessary blank lines). 
I don't know if this is an option you are willing to add to Spyder. I find it very useful but that is not my call :-)
The option is already available in QtPlainTextEdit (`setCenterOnScroll(True)`) so it did not add a lot of complexity to the editor code. The `scrollflagarea` does not need to be modified.

scrollpastend

For some reason however I can't find a way to update the editor immediatly after adding the option. The tab that has focus is not updated while the others are updated. As soon as I add a new line (or remove one) it gets updated. Could you point me to what I am missing? 
- [x] Update editor after adding the option

### DIFF
--- a/spyder/widgets/sourcecode/codeeditor.py
+++ b/spyder/widgets/sourcecode/codeeditor.py
@@ -1214,7 +1214,7 @@ class CodeEditor(TextEditBaseWidget):
         """
         self.scrollpastend_enabled = state
         self.setCenterOnScroll(state)
-        self.update()
+        self.setDocument(self.document())
 
     def resizeEvent(self, event):
         """Reimplemented Qt method to handle p resizing"""
--- a/spyder/widgets/sourcecode/codeeditor.py
+++ b/spyder/widgets/sourcecode/codeeditor.py
@@ -288,6 +288,9 @@ class CodeEditor(TextEditBaseWidget):
         # Blanks enabled
         self.blanks_enabled = False
 
+        # Scrolling past the end of the document
+        self.scrollpastend_enabled = False
+
         self.background = QColor('white')
         self.decorations = TextDecorationsManager(self)
 
@@ -597,8 +600,8 @@ class CodeEditor(TextEditBaseWidget):
                      add_colons=True, auto_unindent=True, indent_chars=" "*4,
                      tab_stop_width_spaces=4, cloned_from=None, filename=None,
                      occurrence_timeout=1500, show_class_func_dropdown=True,
-                     indent_guides=False):
-        
+                     indent_guides=False, scroll_past_end=False):
+
         # Code completion and calltips
         self.set_codecompletion_auto(codecompletion_auto)
         self.set_codecompletion_case(codecompletion_case)
@@ -628,6 +631,9 @@ class CodeEditor(TextEditBaseWidget):
 
         # Blanks
         self.set_blanks_enabled(show_blanks)
+
+        # Scrolling past the end
+        self.set_scrollpastend_enabled(scroll_past_end)
 
         # Line number area
         if cloned_from:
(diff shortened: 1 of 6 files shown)


Prikers:
I accidentally commited the removal of all trailling white spaces (automatically done with spyder). Should I revert that?

ccordoba12:
Yep, please revert.

spyder/widgets/sourcecode/codeeditor.py
```diff
@@ -1201,6 +1207,15 @@ def set_blanks_enabled(self, state):
...
         self.rehighlight()
 
+    def set_scrollpastend_enabled(self, state):
+        """
+        Allow user to scroll past the end of the document to have the last
+        line on top of the screen
+        """
+        self.scrollpastend_enabled = state
+        self.setCenterOnScroll(state)
+        self.update()
```
jnsebgosselin: replace `self.update()` by `self.setDocument(self.document())` and it should solve your issue about the option not being applied at runtime.

Prikers: Works like a charm! Thanks @jnsebgosselin!

ccordoba12:
@Prikers, please create a new branch and PR for your last work (the Pandas Index one) to not mix things with this PR. Thanks!

Prikers:
So sorry! I forgot that I had already pushed this first PR...
Could you help me on that one? I was thinking doing a git reset to go back to 7cefc10 and then `git push origin --force` (as I cannot simply push to origin). Is that a bad practice? 
Again all appologies :confused:

rlaverde:
>Could you help me on that one? I was thinking doing a git reset to go back to 7cefc10 and then git push origin --force (as I cannot simply push to origin). Is that a bad practice?

Yes, It's somehow a bad practice, but It's the only way to remove those commits, remember to be very careful when using `--force`

And remember to create new branches for each PR.

Prikers:
ok. Sorry again about that. 
It is now clean and (I think) ready.

ccordoba12:
@Prikers, there are conflicts now. Please merge with master to solve it.

Prikers:
@ccordoba12, @dalthviz, @jnsebgosselin, @rlaverde thanks for considering, reviewing and helphing with this PR! I have merge upstream/master, there are no conflicts anymore. 
As for the pep8 issue, I am not sure how I should handle it... If I add some indentation to [Line999](https://github.com/spyder-ide/spyder/blob/7cefc101f61197a9c39887458181ce7c51d8d088/spyder/plugins/editor.py#L999) for visual indent, line 1000 will not be aligned anymore. And because of the 79 characters limit I can not align line 1000. I tried to stick to the style of the previous lines ([Line991](https://github.com/spyder-ide/spyder/blob/7cefc101f61197a9c39887458181ce7c51d8d088/spyder/plugins/editor.py#L991) for example). @dalthviz what would you suggest?

dalthviz:
@Prikers in that case maybe try writing the `create_action` like this:

```python
        self.scrollpastend_action = create_action(
                                        self,
                                        _("Scroll past the end"),
                                        toggled=self.toggle_scroll_past_end)
```
Note that in the first line (when openning the create_action) you don't have any arguments. I think with this the code style issue will be fixed :). Here a [link to pep8 for reference](http://legacy.python.org/dev/peps/pep-0008/#indentation)

ccordoba12:
Thanks @Prikers for your contribution!

goanpeca:
@Prikers I invited you to the [spyder/plubic](https://gitter.im/spyder-ide/public) chat on gitter. Could you log in when you have some time :-) ?

Prikers:
I have just joined 😄 . 
Thanks @goanpeca !

Issue #4856, opened 2017-07-29
Some issues with dedicated consoles


goanpeca:
The dedicated run in ipython console has some issues.

1.) When a script runs, there is a flash for a moment of
<img width="698" alt="screen shot 2017-07-29 at 11 21 25" src="https://user-images.githubusercontent.com/3627835/28746344-17753270-7450-11e7-8d6b-cc1dfbe17b65.png">
before it goes blank again, but then

2.) any output (prints, exceptions) that may or may not occur in the script never show up

3.) Changing the working directory (file explorer, global working directory) every time a new scripts runs turns out to be a very annoying behavior, and I think we should rethink that

ccordoba12:
> When a script runs, there is a flash for a moment of

That's an easy fix.

> any output (prints, exceptions) that may or may not occur in the script never show up

This probably happens the first time you run in a dedicated console. Errors always appear for me afterwards.

> Changing the working directory (file explorer, global working directory) every time a new scripts runs turns out to be a very annoying behavior, and I think we should rethink that

This has been the default behavior for all Spyder history, i.e. changing the cwd to the current file directory on execution, so I'm reluctant to change it.

We are now simply reflecting that in our UI, which is what most people wanted.

goanpeca:
>This probably happens the first time you run in a dedicated console. Errors always appear for me afterwards.

Yes, indeed, but this is still a bug, users should not need to run a script twice to see what failed.

>This has been the default behavior for all Spyder history, i.e. changing the cwd to the current file directory on execution, so I'm reluctant to change it.
We are now simply reflecting that in our UI, which is what most people wanted.

It may be, but it provides a very annoying experience

ccordoba12:
> Yes, indeed, but this is still a bug, users should not need to run a script twice to see what failed.

Sure, I just wanted to verify that it only happens once.

> It may be, but it provides a very annoying experience

Please propose an alternative. Just complaining is not constructive.

goanpeca:
I guess I have to adapt my workflow. If I think of something, will let you know

ccordoba12:
We could change the default option to *not* change the cwd on evaluation, which I also think is a saner default.

@rlaverde, @dalthviz, @andfoy, what do you think?

rlaverde:
>We could change the default option to not change the cwd on evaluation, which I also think is a saner default.

Yes, I also find annoying that the cwd changes too frequently

andfoy:
[...]

PR #5127, opened 2017-09-04
PR: Wait until console It's ready before executing code.

rlaverde:
Fix 1 and 2 of #4856

Fix error of banner and prompt being deleted when executing in a dedicated console.

Now It looks as expected when running in a dedicated console:
spectacle lh4836

### DIFF
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -935,13 +935,13 @@ class IPythonConsole(SpyderPluginWidget):
                 pass
             else:
                 if not current_client:
+                    # wait until console it's ready before executing code
+                    self.wait_shell_is_ready(sw)
                     # Clear console and reset namespace for
                     # dedicated clients
                     sw.silent_execute(
                         'get_ipython().kernel.close_all_mpl_figures()')
                     sw.reset_namespace(force=True)
-                    # wait until console it's ready before executing code
-                    self.wait_shell_is_ready(sw)
                 elif current_client and clear_variables:
                     sw.reset_namespace(force=True)
                 sw.execute(to_text_string(to_text_string(lines)))
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -28,7 +28,7 @@ from qtconsole.client import QtKernelClient
 from qtconsole.manager import QtKernelManager
 from qtpy import PYQT5
 from qtpy.compat import getopenfilename
-from qtpy.QtCore import Qt, Signal, Slot
+from qtpy.QtCore import Qt, Signal, Slot, QEventLoop, QTimer
 from qtpy.QtWidgets import (QApplication, QCheckBox, QDialog, QDialogButtonBox,
                             QFormLayout, QGridLayout, QGroupBox, QHBoxLayout,
                             QLabel, QLineEdit, QMessageBox, QPushButton,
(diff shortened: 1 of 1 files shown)


spyder/plugins/ipythonconsole.py
```diff
@@ -921,12 +937,13 @@ def execute_code(self, lines, current_client=True, clear_variables=False):
...
                     # Clear console and reset namespace for
                     # dedicated clients
-                    sw.silent_execute('%clear')
                     sw.silent_execute(
                         'get_ipython().kernel.close_all_mpl_figures()')
                     sw.reset_namespace(force=True)
                 elif current_client and clear_variables:
                     sw.reset_namespace(force=True)
+                # wait until console it's ready before executing code
+                self.wait_shell_is_ready(sw)
```
ccordoba12: Please move this line to the `if not current_client` block.

goanpeca:
@rlaverde does this mean that the code execution is now slower somehow, because of this timeout? or we take the same time, but simply display info at a later step?

spyder/plugins/ipythonconsole.py
```diff
@@ -911,6 +911,22 @@ def update_working_directory(self):
...
+        """Wait until console It's ready and banner is correctly displayed.
+
+        shell (ShellWidget)
+        timeout (int) milliseconds
+        """
+        i = 0
+        loop = QEventLoop()
+        timer = QTimer()
+        timer.timeout.connect(loop.quit)
+        timer.start(300)
```
ccordoba12: Are we waiting only 300 ms for the prompt to be ready?

ccordoba12: Ok, I saw above that `timeout=3000`, but this should be higher. I think we should set it to `15000` at least.

rlaverde: We are waiting a maximum of 3seconds, and checking every 300ms, testing in my machine It takes ~1.5 seconds to be ready

rlaverde:
> does this mean that the code execution is now slower somehow, because of this timeout?

Only when opening new consoles, It'll wait until console is ready before executing the script, the timeout It's to avoid waiting too much time.

ccordoba12:
> the timeout It's to avoid waiting too much time.

What do you mean by this? What happens if the timeout if finished?

rlaverde:
>What do you mean by this?

It the console is never ready (maybe for an error in the initialization), continuing this infinite loop could hang spyder, the timeout is the maximum time to wait for the console to be ready

>What happens if the timeout is finished?

It'll return and the code will executed, Maybe deleting the prompt, or mixing the output with the prompt

spyder/plugins/ipythonconsole.py
```diff
@@ -921,10 +937,11 @@ def execute_code(self, lines, current_client=True, clear_variables=False):
                 if not current_client:
                     # Clear console and reset namespace for
                     # dedicated clients
-                    sw.silent_execute('%clear')
                     sw.silent_execute(
                         'get_ipython().kernel.close_all_mpl_figures()')
                     sw.reset_namespace(force=True)
+                    # wait until console it's ready before executing code
+                    self.wait_shell_is_ready(sw)
```
ccordoba12: This has to be before the other two previous calls (i.e. `sw.silent_execute` and `sw.reset_namespace`) because those require a running kernel to work.

rlaverde:
@ccordoba12 I moved the code, although I don't think it will be a difference, because the wait (for the banner) is necessary for commands that print something.

ccordoba12:
I was not too happy with this approach because I found it a bit too complicated. So I developed #5301, which I find simpler to understand.

Issue #5114, opened 2017-09-02
KeyboardInterrupt error in waitingspinner


awsiitchs:
## Description

**What steps will reproduce the problem?**

1. 
2. 
3. 

**What is the expected output? What do you see instead?**

**Please provide any additional information below**

  File "/Users/Leo/anaconda/lib/python3.6/site-packages/spyder/widgets/waitingspinner.py", line 177, in rotate
    def rotate(self):
KeyboardInterrupt

## Version and main components

* Spyder Version: 3.2.0 
* Python Version: 3.6.2
* Qt Versions: 5.6.2, PyQt5 5.6 on Darwin

## Dependencies
```
pyflakes >=0.6.0 :  1.6.0 (OK)
pycodestyle >=2.3:  2.3.1 (OK)
pygments >=2.0   :  2.2.0 (OK)
pandas >=0.13.1  :  0.20.3 (OK)
numpy >=1.7      :  1.13.1 (OK)
sphinx >=0.6.6   :  1.6.3 (OK)
rope >=0.9.4     :  0.10.5 (OK)
jedi >=0.9.0     :  0.10.2 (OK)
psutil >=0.3     :  5.2.2 (OK)
nbconvert >=4.0  :  5.2.1 (OK)
sympy >=0.7.3    :  1.1.1 (OK)
cython >=0.21    :  0.26 (OK)
qtconsole >=4.2.0:  4.3.0 (OK)
IPython >=4.0    :  6.1.0 (OK)
pylint >=0.25    :  1.7.2 (OK)

```

ccordoba12:
Thanks for reporting. We'll take a look at this in a future release.

rlaverde:
I think this is caused because the waiting spinner is started in the creation of find in files, and never stopped (only the progress bar is hidden/shown) https://github.com/spyder-ide/spyder/blob/3.x/spyder/widgets/findinfiles.py#L784

ccordoba12:
Thanks for noticing @rlaverde! Should we stop it then after it's hidden?

rlaverde:
Yes, start it when progress bar is shown and stop it when progress bar is hidden.

Although if you press Ctrl+C while it's shown and spinning (i.e. searching in files) the error will occur

goanpeca:
>Although if you press Ctrl+C while it's shown and spinning (i.e. searching in files) the error will occur

We should be able to prevent this

rlaverde:
[...]

PR #5129, opened 2017-09-04
PR: Avoid KeyboardInterrupt to raise an error in waitingspinner

rlaverde:
Fixes: #5114

### DIFF
--- a/spyder/widgets/findinfiles.py
+++ b/spyder/widgets/findinfiles.py
@@ -781,7 +781,6 @@ class FileProgressBar(QWidget):
         self.spinner = QWaitingSpinner(self, centerOnParent=False)
         self.spinner.setNumberOfLines(12)
         self.spinner.setInnerRadius(2)
-        self.spinner.start()
         layout = QHBoxLayout()
         layout.addWidget(self.spinner)
         layout.addWidget(self.status_text)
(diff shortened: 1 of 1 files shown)


spyder/app/mainwindow.py
```diff
@@ -288,6 +288,7 @@ def signal_handler(signum, frame=None):
                 pass
         else:
             signal.signal(signal.SIGTERM, signal_handler)
+            signal.signal(signal.SIGINT, signal_handler)
```
ccordoba12: Why is this here?

rlaverde: https://github.com/spyder-ide/spyder/issues/5114#issuecomment-327024883

It make possible to kill spyder using `Ctrl+C`, without this change It will still raise the report error message box when pressing `Ctrl+C`

ccordoba12: Please remove this. We need to discuss it in another PR.

goanpeca: @ccordoba12 why? I think it is important for people to be able to quit spyder on the terminal using `Ctrl+C`

ccordoba12: When Spyder freezes for whatever reason, it's very useful to press `Ctrl+C` in the terminal and get the traceback of where it was freezed.

rlaverde: But, the users will continue reporting false positives (now random), of when they tried to close spyder with `Ctrl+C`

ccordoba12: But that only happens when users press `Ctrl+C` in a terminal, and I doubt there are many of them.

In any case, I think this needs more discussion and is not related to this PR. So please remove it and open a new PR with change only.

rlaverde:
@ccordoba12 I move that changes to a new PR #5159 and let only waiting spinner changes in this PR

ccordoba12:
Great! Thanks @rlaverde!

Issue #4948, opened 2017-08-11
Class function dropdown should be deactivated by default


rlaverde:
Instead we could add an option in the Source Menu, to activate/deactivate it more easily

PR #5130, opened 2017-09-04
PR: Deactivate class/function dropdown by default

rlaverde:
Fixes: #4948

### DIFF
--- a/spyder/plugins/editor.py
+++ b/spyder/plugins/editor.py
@@ -113,8 +113,9 @@ class EditorConfigPage(PluginConfigPage):
         interface_group = QGroupBox(_("Interface"))
         newcb = self.create_checkbox
         showtabbar_box = newcb(_("Show tab bar"), 'show_tab_bar')
-        showclassfuncdropdown_box = newcb(_("Show Class/Function Dropdown"),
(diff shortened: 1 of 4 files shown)


spyder/plugins/editor.py
```diff
@@ -995,6 +995,13 @@ def get_plugin_actions(self):
         showindentguides_action = create_action(self, _("Show indent guides."),
                                                toggled=self.toggle_show_indent_guides)
         showindentguides_action.setChecked(CONF.get('editor', 'indent_guides'))
+
+        show_classfunc_dropdown_action = create_action(
+                self, _("Show class function dropdown."),
```
ccordoba12: Please change this to

    Show class and methods dropdown

ccordoba12:
I left a minor comment, then I'll merge.

spyder/plugins/editor.py
rlaverde: In the preferences says: `Show Class/Function Dropdown`, should I change it too?

ccordoba12: I wasn't aware of that. Then let's leave it (in both places) as

    Show selector for classes and functions

PR #5132, opened 2017-09-04
PR: Improve how floating panels are painted

### LINKED ISSUE #4590: Improve floating panels in the editor
The magic for calculate the position of edgeline and indentguides, could be improve it and if hardcoded should be in only one place.
https://github.com/spyder-ide/spyder/pull/4572#discussion_r120737977

rlaverde:
Fixes #4590 

After some debugging I realize that the hardcoded `5` correspond to `self.editor.document().documentMargin()`

### DIFF
--- a/spyder/widgets/panels/edgeline.py
+++ b/spyder/widgets/panels/edgeline.py
@@ -70,7 +70,7 @@ class EdgeLine(Panel):
         offset = self.editor.contentOffset()
         x = self.editor.blockBoundingGeometry(self.editor.firstVisibleBlock()) \
             .translated(offset.x(), offset.y()).left() \
-            +self.editor.fontMetrics().width('9'*min(self.columns))+5
+            +self.editor.fontMetrics().width('9'*min(self.columns)) + self.editor.document().documentMargin()
 
         top_left = QPoint(x, cr.top())
         top_left = self.editor.calculate_real_position(top_left)
--- a/spyder/widgets/panels/edgeline.py
+++ b/spyder/widgets/panels/edgeline.py
@@ -8,7 +8,7 @@
 This module contains the edge line panel
 """
 
-from qtpy.QtCore import Qt, QRect, QPoint
+from qtpy.QtCore import Qt
 from qtpy.QtGui import QPainter, QColor
 
 from spyder.py3compat import is_text_string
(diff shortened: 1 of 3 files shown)


spyder/widgets/panels/edgeline.py
```diff
@@ -61,19 +61,13 @@ def update_color(self):
...
             width --> max position - min position
         """
-        width = self.editor.fontMetrics().width('9'*(max(self.columns) - min(self.columns))) + 1
-        offset = self.editor.contentOffset()
-        x = self.editor.blockBoundingGeometry(self.editor.firstVisibleBlock()) \
-            .translated(offset.x(), offset.y()).left() \
-            +self.editor.fontMetrics().width('9'*min(self.columns))+5
+        x = self.editor.fontMetrics().width('9'*min(self.columns))
+        width = self.editor.fontMetrics().width('9'*(max(self.columns)
+                                                - min(self.columns))) + 1
```
goanpeca: So what is the 1 for :-p

rlaverde: It's the width of an edge line

goanpeca: If this is the case, then this might fail on other Operating systems. We should use Qt metrics then. otherwise you just changed a 5 for a 1

rlaverde: That one was always there, I'll search the best way to make it platform independent

rlaverde: If we don't set the QPen width it will be 1 pixel width. http://doc.qt.io/qt-4.8/qpen.html#details

What I should do? define a constant `WIDTH = 1`? add `width` as a property of the class?
 add a comment?

goanpeca: Maybe?

rlaverde:
@goanpeca I decided to remove that logic instead, there is not any good reason to have a custom width and position for EdgeLine panel, instead I'll use the same dimensions as the editor.

Also using custom dimensions is somehow error-prone, because the geometry have to be re-set when the preferences change.

ccordoba12:
Is this ready?

rlaverde:
>Is this ready?

Yes 😄

PR #5159, opened 2017-09-07
PR: Make Ctrl+C to quit Spyder on Posix systems

### LINKED ISSUE #5305: Error when pressing Ctrl+C in the terminal where Spyder was started
## Description

1. 
2. 
3. 

  File "/home/salalaika/application/Anaconda3/lib/python3.6/site-packages/spyder/widgets/status.py", line 108, in update_label
    def update_label(self):
KeyboardInterrupt

* Spyder Version: 3.2.3 
* Python Version: 3.6.1

## Dependencies
```
sympy >=0.7.3    :  1.0 (OK)

```

rlaverde:
Fixes #5305

Related to #5114

See: https://github.com/spyder-ide/spyder/pull/5129#discussion_r136875502

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -288,7 +288,11 @@ class MainWindow(QMainWindow):
                 pass
         else:
             signal.signal(signal.SIGTERM, signal_handler)
-            signal.signal(signal.SIGINT, signal_handler)
+            if not DEV:
+                # Make spyder quit when presing ctrl+C in the console
+                # In DEV Ctrl+C doesn't quit, because it helps to
+                # capture the traceback when spyder freezes
+                signal.signal(signal.SIGINT, signal_handler)
 
         # Use a custom Qt stylesheet
         if sys.platform == 'darwin':
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -288,6 +288,7 @@ class MainWindow(QMainWindow):
                 pass
         else:
             signal.signal(signal.SIGTERM, signal_handler)
+            signal.signal(signal.SIGINT, signal_handler)
 
         # Use a custom Qt stylesheet
         if sys.platform == 'darwin':


spyder/app/mainwindow.py
```diff
@@ -288,6 +288,7 @@ def signal_handler(signum, frame=None):
                 pass
         else:
             signal.signal(signal.SIGTERM, signal_handler)
+            signal.signal(signal.SIGINT, signal_handler)
```
ccordoba12: Please change this to

```
if not DEV:
    ...
```

goanpeca: Why?

ccordoba12: Because it's useful to have it when working in DEV mode. That way you can get a traceback when Spyder freezes for whatever reason.

goanpeca: >Because it's useful to have it when working in DEV mode. That way you can get a traceback when Spyder freezes for whatever reason.

Still this is very unconventional behavior. I am sure there must be a way where we can get a traceback AND quit...

ccordoba12: I think very few people uses DEV, so I don't think we should look for more.

rlaverde: >Still this is very unconventional behavior. I am sure there must be a way where we can get a traceback AND quit...

Raising KeyInterruptError in the internal console when the traceback correspond to a KeyInterruptError

goanpeca: could we do that? raise something, wait a bit and then close spyder?

rlaverde: Hmm yes, raising a exception in the  internal console will make spyder quit, although the traceback won't correspond 😕

goanpeca: >Hmm yes, raising a exception in the internal console will make spyder quit, although the traceback won't correspond 😕

Then it is not the solution I requested!

rlaverde: We could also print the other traceback, although I don't like that option.

Maybe skipping the keyboard error where internal console catch errors, although I don't understand the internals of the console and idk if it's possible

ccordoba12: @rlaverde, please finish this one with my initial suggestion. I know that's not ideal, but it will affect mainly to us.

spyder/app/mainwindow.py
```diff
@@ -288,7 +288,11 @@ def signal_handler(signum, frame=None):
                 pass
         else:
             signal.signal(signal.SIGTERM, signal_handler)
-            signal.signal(signal.SIGINT, signal_handler)
+            if not DEV:
+                # Make spyder quit when presing ctrl+C in the console
+                # In DEV because Ctrl+C doens't quit, because it helps to
```
ccordoba12: Please remove the first `because` in this line.

ccordoba12: This also needs a rebase against 3.x to fix the error in Circle.

ccordoba12: `doens't` -> `doesn't`

Issue #5176, opened 2017-09-08
Unify how extra selection are added to the editor


rlaverde:
The idea is to move how extra selections for cell, lines, search and others that use these functions:
https://github.com/spyder-ide/spyder/blob/master/spyder/widgets/sourcecode/base.py#L319

Move them to use TextDecoration API:
https://github.com/spyder-ide/spyder/blob/master/spyder/widgets/sourcecode/api/decoration.py

This is needed for #4289

rlaverde:
This is also causing a little error: Selecting a codefolding arrow will add codefolding extraselection, but editor extraselections will be removed (current line, current cell, highlight words)

PR #5203, opened 2017-09-11
PR: Unify how extra selections are added to the Editor

rlaverde:
Fixes: #5176 

I'm not sure if adding `DecorationManager` to the `TextEditBaseWidget` is ok (commit f408110), or should `DecorationManager` belong to `CodeEditor` and use `try..  except` in `TextEditBaseWidget` methods

### DIFF
--- a/spyder/widgets/panels/codefolding.py
+++ b/spyder/widgets/panels/codefolding.py
@@ -348,6 +348,15 @@ class FoldingPanel(Panel):
         return color
 
     def _decorate_block(self, start, end):
+        """
+        Create a decoration and addedit to the editor.
+
+        NOTE: codefolding decorations will have draw_order=1
+
+        Args:
+            start (int) start line of the decoration
+            end (int) end line of the decoration
+        """
         color = self._get_scope_highlight_color()
         d = TextDecoration(self.editor.document(),
                            start_line=start, end_line=end+1, draw_order=1)
--- a/spyder/widgets/panels/codefolding.py
+++ b/spyder/widgets/panels/codefolding.py
@@ -16,7 +16,8 @@ from qtpy.QtGui import (QTextBlock, QColor, QFontMetricsF, QPainter,
                         QLinearGradient, QPen, QPalette, QResizeEvent,
                         QCursor)
 
-from spyder.widgets.sourcecode.api.decoration import TextDecoration
+from spyder.widgets.sourcecode.api.decoration import (TextDecoration,
+                                                      DRAW_ORDERS)
 from spyder.widgets.sourcecode.folding import FoldScope
 from spyder.api.panel import Panel
 from spyder.utils.editor import (TextBlockHelper, TextHelper, DelayJobRunner,
@@ -351,15 +352,14 @@ class FoldingPanel(Panel):
         """
         Create a decoration and add it to the editor.
 
-        NOTE: codefolding decorations will have draw_order=1
-
         Args:
             start (int) start line of the decoration
             end (int) end line of the decoration
         """
         color = self._get_scope_highlight_color()
-        d = TextDecoration(self.editor.document(),
-                           start_line=start, end_line=end+1, draw_order=1)
+        draw_order = DRAW_ORDERS.get('codefolding')
+        d = TextDecoration(self.editor.document(), start_line=start,
+                           end_line=end+1, draw_order=draw_order)
         d.set_background(color)
         d.set_full_width(True, clear=False)
         self.editor.decorations.add(d)
@@ -442,10 +442,10 @@ class FoldingPanel(Panel):
         Add fold decorations (boxes arround a folded block in the editor
         widget).
         """
-        deco = TextDecoration(block, draw_order=1)
+        draw_order = DRAW_ORDERS.get('codefolding')
+        deco = TextDecoration(block, draw_order=draw_order)
         deco.signals.clicked.connect(self._on_fold_deco_clicked)
         deco.tooltip = region.text(max_lines=25)
-        deco.draw_order = 1
         deco.block = block
         deco.select_line()
         deco.set_outline(drift_color(
--- a/spyder/widgets/panels/codefolding.py
+++ b/spyder/widgets/panels/codefolding.py
@@ -353,7 +353,7 @@ class FoldingPanel(Panel):
                            start_line=start, end_line=end+1, draw_order=1)
         d.set_background(color)
         d.set_full_width(True, clear=False)
-        self.editor.decorations.append(d)
+        self.editor.decorations.add(d)
         self._scope_decos.append(d)
 
     def _highlight_block(self, block):
@@ -444,7 +444,7 @@ class FoldingPanel(Panel):
         deco.set_background(self._get_scope_highlight_color())
         deco.set_foreground(QColor('#808080'))
         self._block_decos.append(deco)
-        self.editor.decorations.append(deco)
+        self.editor.decorations.add(deco)
 
     def toggle_fold_trigger(self, block):
         """
@@ -556,7 +556,7 @@ class FoldingPanel(Panel):
                 deco.set_outline(drift_color(
                     self._get_scope_highlight_color(), 110))
                 deco.set_background(self._get_scope_highlight_color())
-                self.editor.decorations.append(deco)
+                self.editor.decorations.add(deco)
         self._prev_cursor = cursor
 
     def _refresh_editor_and_scrollbars(self):
--- a/spyder/widgets/panels/codefolding.py
+++ b/spyder/widgets/panels/codefolding.py
@@ -349,7 +349,7 @@ class FoldingPanel(Panel):
 
     def _decorate_block(self, start, end):
         """
-        Create a decoration and addedit to the editor.
+        Create a decoration and add it to the editor.
 
         NOTE: codefolding decorations will have draw_order=1
 
--- a/spyder/widgets/panels/codefolding.py
+++ b/spyder/widgets/panels/codefolding.py
@@ -350,7 +350,7 @@ class FoldingPanel(Panel):
     def _decorate_block(self, start, end):
         color = self._get_scope_highlight_color()
         d = TextDecoration(self.editor.document(),
-                           start_line=start, end_line=end+1)
+                           start_line=start, end_line=end+1, draw_order=1)
         d.set_background(color)
         d.set_full_width(True, clear=False)
         self.editor.decorations.append(d)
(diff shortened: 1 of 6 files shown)


goanpeca:
@rlaverde it depends on where we are using `TextEditBaseWidget` and if in those places we require the functionality. Internal console, history log, I guess? @ccordoba12 ?

rlaverde:
Searching all the codebase, `ExtraSelection` objects were only created in the base `TextEditBaseWidget` class. 

Console and history add extra selections, but they do it through the `TextEditBaseWidget`methods.

rlaverde:
I'm getting tired of restarting the travis python2.7 building, and every time a different test fails (unrelated to this PR)

spyder/widgets/panels/codefolding.py
```diff
@@ -350,7 +350,7 @@ def _get_scope_highlight_color(self):
     def _decorate_block(self, start, end):
         color = self._get_scope_highlight_color()
```
goanpeca: A docstring would be nice :-) with some explanation on the args.

spyder/widgets/sourcecode/base.py
```diff
@@ -34,7 +33,8 @@
 from spyder.widgets.calltip import CallTipWidget
 from spyder.widgets.mixins import BaseEditMixin
 from spyder.widgets.sourcecode.terminal import ANSIEscapeCodeHandler
-
+from spyder.widgets.sourcecode.api.decoration import TextDecoration
+from spyder.widgets.sourcecode.utils.decoration import TextDecorationsManager
```
goanpeca: Why we have the manager outside of the api, is because we don't need/want to expose that?

spyder/widgets/sourcecode/base.py
```diff
@@ -34,7 +33,8 @@
 from spyder.widgets.calltip import CallTipWidget
 from spyder.widgets.mixins import BaseEditMixin
 from spyder.widgets.sourcecode.terminal import ANSIEscapeCodeHandler
-
+from spyder.widgets.sourcecode.api.decoration import TextDecoration
+from spyder.widgets.sourcecode.utils.decoration import TextDecorationsManager
 
 def insert_text_to(cursor, text, fmt):
     """Helper to print text, taking into account backspaces"""
```
goanpeca: period at the end of the docstring ;-)

spyder/widgets/sourcecode/base.py
```diff
@@ -236,7 +236,7 @@ def __init__(self, parent=None):
         BaseEditMixin.__init__(self)
         self.setAttribute(Qt.WA_DeleteOnClose)
         
-        self.extra_selections_dict = OrderedDict()
+        self.extra_selections_dict = {}
```
goanpeca: Why the change is not needed? because of the `draworder` parameter?

spyder/widgets/sourcecode/base.py
```diff
@@ -330,30 +332,28 @@ def get_extra_selections(self, key):
         return self.extra_selections_dict.get(key, [])
 
     def set_extra_selections(self, key, extra_selections):
+        # current line has to be highlighted first
```
goanpeca: Maybe a docstring would be nice, also stating what key and extra_selections is supposed to be. Like a poors-man typing :-p

spyder/widgets/sourcecode/utils/decoration.py
```diff
@@ -50,6 +49,21 @@ def remove(self, decoration):
         except ValueError:
             return False
 
+    def extend(self, decorations):
```
goanpeca: I still prefer add :-p

spyder/widgets/sourcecode/base.py
rlaverde: Hmm yes, I don't think there is a reason to expose it

Normally you will create new TextDecoration and append them to the editor.decorations.

spyder/widgets/sourcecode/base.py
rlaverde: I didn't add this docstring 🙈  (I'll improve it with some clarifications of the  new behavior and fix that style error. )

spyder/widgets/sourcecode/base.py
rlaverde: Partially yes, extr selections are now ordered  by the `DecorationManager`,

I moved this logic to `DecorationManager._order_decorations` this will allow that all decorations added to the editor will be order (according draw_order and selection length)

spyder/widgets/sourcecode/base.py
goanpeca: I know, but its a ood opportunity to clean it up a bit :-)

spyder/widgets/sourcecode/base.py
rlaverde: yep, them I'll improve all docstring of  modified functions (no only the ones that I added)

spyder/widgets/sourcecode/utils/decoration.py
rlaverde: `extend` is different to `add`. This function extends a list with another list
Maybe `append` could be changed to `add`

goanpeca: Well sounds like extend and append for this hight level case should be handled as a single thing/method/function. I know this was supposed to look like a list but it makes no sense that it has to follow the list primitives, this is a higher level object, not a list.

rlaverde: Ok, then I'll fusion append and extend into a function named add

spyder/widgets/sourcecode/base.py
rlaverde: I improved other docstrings, but I didn't modify this function at all, fixing that docstring will add noise to the PR

spyder/widgets/panels/codefolding.py
```diff
@@ -348,12 +348,21 @@ def _get_scope_highlight_color(self):
         return color
 
     def _decorate_block(self, start, end):
+        """
+        Create a decoration and addedit to the editor.
```
ccordoba12: `addedit` -> `add it`?

spyder/widgets/sourcecode/base.py
```diff
@@ -317,43 +319,61 @@ def set_palette(self, background, foreground):
...
+
+        Args:
+            key (str) name of the extra selections group.
+            extra_selections (list of sourcecode.api.TextDecoration).
+        """
+        # current line has to be highlighted first
+        draw_orders = {'current_cell': 0,
+                       # draw order = 1 is used in codefolding
+                       'current_line': 2}
+        draw_order = draw_orders.get(key,  5)
```
ccordoba12: Why `5` here?

spyder/widgets/sourcecode/base.py
```diff
@@ -317,43 +319,61 @@ def set_palette(self, background, foreground):
...
+
+        NOTE: This will remove previous decorations added to  the same key.
+
+        Args:
+            key (str) name of the extra selections group.
+            extra_selections (list of sourcecode.api.TextDecoration).
+        """
+        # current line has to be highlighted first
+        draw_orders = {'current_cell': 0,
+                       # draw order = 1 is used in codefolding
```
ccordoba12: Why `0` can't be used for code folding?

spyder/widgets/sourcecode/tests/test_extra_selections.py
```diff
@@ -0,0 +1,62 @@
...
+    editor = construct_editor()
+
+    # Move cursor over 'some_variable'
+    editor.go_to_line(2)
+    cursor = editor.textCursor()
+    cursor.movePosition(QTextCursor.Right, n=5)
+    editor.setTextCursor(cursor)
+
+    qtbot.waitUntil(lambda: len(editor.extraSelections()) >= 6, timeout=2000)
+    print(len(editor.extraSelections()))
```
ccordoba12: Is this print really needed?

goanpeca: I think it ts a good practice to have prints inside tests cause when the test fails you get more debugging info. Dont know of it applies to this though.

spyder/widgets/panels/codefolding.py
rlaverde: Yes :-)

spyder/widgets/sourcecode/base.py
rlaverde: It just have to be a number greater that current_line and curreent_cell.

I choose 5 to leave 3 and 4 free, maybe somebody will need to add something in front of the current line, but behind if the other selections.

spyder/widgets/sourcecode/base.py
rlaverde: I put 1 to make it appear in front of current cell, but behind other selections.

I'm not sure, what is the best order, codefolding in the back? Or maybe in the front and be a little transparent?

spyder/widgets/sourcecode/tests/test_extra_selections.py
rlaverde: Hmm no, I was using it locally for know the variables when the test failed. I'll remove it

goanpeca: >I was using it locally for know the variables when the test failed

Well.. if the test fails, it would be nice to have that information, no? otherwise we just get an assertion error that might not provide enough info

rlaverde: This print wasn't giving too much information.

But adding print to test could be a good practice, for easy now what is failing in the CI when a test fails.

spyder/widgets/sourcecode/base.py
ccordoba12: Please define and use constants for this. Something like:

    CELL_ORDER = 0
    FOLDING_ODER = 1
    LINE_ORDER = 2

and use those constants here.

spyder/widgets/sourcecode/base.py
ccordoba12: This needs to be explained when you define the constants I mentioned above.

spyder/widgets/sourcecode/base.py
rlaverde: I'll use this same dict an expose it as a part of the decoration api.

ccordoba12: Great!

rlaverde: I just add a commit with the DRAW_ORDERS dict, I didn't leave 3 and 4 unused because now its easier to modify the draworders values.

ccordoba12:
This looks really good to me. @goanpeca?

PR #5206, opened 2017-09-12
PR: Split IPython Console module

rlaverde:
```
ipythonconsole/
├── __init__.py
├── plugin.py
├── tests
│   ├── __init__.py
│   └── test_ipythonconsole.py
├── utils
│   ├── __init__.py
│   ├── kernelspec.py
│   ├── spyder_kernel.py
│   ├── start_kernel.py
│   ├── style.py
│   ├── site
│   │   ├── __init__.py
│   │   ├── osx_app_site.py
│   │   └── sitecustomize.py
│   ├── templates
│   │   ├── blank.html
│   │   ├── kernel_error.html
│   │   └── loading.html
│   └── tests
│       ├── __init__.py
│       └── test_spyder_kernel.py
└── widgets
    ├── client.py
    ├── control.py
    ├── debugging.py
    ├── help.py
    ├── __init__.py
    ├── namespacebrowser.py
    └── shell.py
```


ccordoba12:
`utils/ipython` and `utils/site` need to be moved to `plugins/ipythonconsole` too.

rlaverde:
I haven't notice that adding deep 5 in `modules_test.sh` is needed because plugins make spyder tree a level deeper.

Because of that files in` variable/widgets/viewers` weren't executed, and kwon they are falling because aa viewer is named `collections` and interfere with the `collections` python builtin module ( that file was renamed in #4557)

Shuold I rename that file? maybe `collection` without the `s`

goanpeca:
Hi I think tempkates should go elsewhere maybe? Not sure it belongs inside utils. Maybe we should have a static folder?

rlaverde:
>Hi I think tempkates should go elsewhere maybe? Not sure it belongs inside utils. Maybe we should have a static folder?

Yes,I like this idea, also Help plugin have  some static files. What will be the name of this folder? `static`, `assets`?

goanpeca:
>static, assets

Hmm no real preference. static is more "traditional", assets is more general. Any preference @ccordoba12 ?

ccordoba12:
Let's leave things as they are, for now. All Help assets are going to 
disappear once we replace them with docrepr (something we can do for 
beta2 or beta3).

goanpeca:
>Let's leave things as they are, for now. All Help assets are going to
disappear once we replace them with docrepr (something we can do for
beta2 or beta3).

This is for the ipython plugin (not help :-p, that was just a side note), the idea, is to clean up the structure a bit so still we have templates inside a python module, which is a bit odd :-p

rlaverde:
Then I won't move Help plugin static files.

But other plugins could need a folder for static  stuff (for example `spyder_reports` have a welcome.md, also `spyder_notebook` have some static files).

goanpeca:
>But other plugins could need a folder for static stuff (for example spyder_reports have a welcome.md, also spyder_notebook have some static files).

Exactly, we still need that folder in the structure of some plugins

ccordoba12:
Then let's put `templates` in

    ipythonconsole/static/templates

Is that ok?

ccordoba12:
Or maybe even better

    ipythonconsole/assets/templates

`static` is a bit odd because we're not developing web pages ;-)

goanpeca:
`assets` it is then @rlaverde ;-)

rlaverde:
And what about the error I mention in https://github.com/spyder-ide/spyder/pull/5206#issuecomment-328848619 (collections name interfering with bulitin)

goanpeca:
>(collections name interfering with bulitin)

That works for me, but I guess there is a more generic term? Collections refers to dict like objects, or ?

rlaverde:
>Collections refers to dict like objects, or ?

Dictionaries, lists, tuples...

ccordoba12:
> and now they are falling because aa viewer is named collections and interfere with the collections python builtin module

Please rename that module to `iterables`, and all places where `collection` or `collections` is used as part of the name of class or a method.

I'd rather prefer to do that in another PR though, and leave commented level 5 for now.

rlaverde:
>I'd rather prefer to do that in another PR though, and leave commented level 5 for now.

Ok, I'll comment out and add a TODO

PR #5214, opened 2017-09-12
PR: Split Projects module

rlaverde:
```
projects/
├── api.py
├── confpage.py
├── __init__.py
├── plugin.py
├── projectstypes
│   ├── __init__.py
│   └── python.py
├── utils
│   ├── config.py
│   └── __init__.py
└── widgets
    ├── explorer.py
    ├── __init__.py
    ├── projectdialog.py
    └── tests
        ├── __init__.py
        ├── test_configdialog.py
        ├── test_projectdialog.py
        └── test_project_explorer.py

```

~I'm not sure where is the best place  to put `type` folder~, I put it in the root and move its `__init__.py` to `apy.py`


goanpeca:
>I'm not sure where is the best place to put type folder

Well they are not really widgets, so maybe at the same level as the widgets folder? or with the api?

rlaverde:
>Well they are not really widgets, so maybe at the same level as the widgets folder? or with the api?

I think that at the same level of the widget folder. (as in outlineexplorer)

Also the `types/__init__` contains the `EmptyProject`  I think that should be part of the api

goanpeca:
>Also the types/__init__ contains the EmptyProject I think that should be part of the api

Yeah, not sure how this goes, but the base project needs to be extendable by other type of projects

ccordoba12:
`type` is a Python keyword, so we shouldn't use it as a module name.

What about `projectstype` or `projtypes`?

rlaverde:
Maybe languages? although we will need a more general word

@goanpeca do you remember the word that we think that could work for the api?

ccordoba12:
@rlaverde, please rebase and finish this one.

rlaverde:
I'll use `projectstypes`. Although we need yo unify this names, In `OutlineExplorer` a file with a similar function is called `languages.py`

ccordoba12:
This looks good to me. I just a minor suggestion: renaming `projectstypes` to `projecttypes` (two plural words in English is rarely used).

Then I'll merge.

rlaverde:
>I just have a minor suggestion: renaming projectstypes to projecttypes

Done

Issue #4865, opened 2017-07-31
Synchronization between multiple editors


dalthviz:
As discussed in #4850

ccordoba12:
@rlaverde, please work on this one.

rlaverde:
I don't understand clearly what's the idea of this change, Sync all editorstacks, and move tabs in all editors is not the expected behavior, even you says that https://github.com/spyder-ide/spyder/pull/4850#issuecomment-318915842

I'm 👎 to this issue, It'll be confusing and somehow error prone, other editors doesn't sync tabs between multiple editors, neither in split view.

jnsebgosselin:
From a [comment](https://github.com/spyder-ide/spyder/pull/4850#issuecomment-319083206) of @rlaverde :
>Personally I'll prefer the both windows to be totally independent, not only the position of the tabs not to be synced, but also different files opened, and when closing a tab to only be closed in that window

Lets say I'm working on PR X with file 1, file 2, and file 3 opened in an Editor plugin. Now I need to work on PR Y that requires files 4, file 5, and file 6, but I don't want to close file 1, file 2 and file 3 because I will need them later. Then I need to work on another PR and so on. I usually end up with a very crowded tab bar in my editor.

One of the thing I would like is the ability to add more than one dockable editor plugin to Spyder. Each editor plugin would be completely independent from each other. The editorstacks within each editor plugin would be synced however.

ccordoba12:
@dalthviz, what was *exactly* this issue about? Sorry, I don't remember ;-)

dalthviz:
@ccordoba12 I think this is related with the synchronization of the tabs to preserve the same order when opening a new window. Although I'm not quite sure either.

PR #5221, opened 2017-09-13
PR: Make the fileswitcher aware of split editorstacks

### LINKED ISSUE #5057: Fast file switch and fast symbol search do not work properly if the editor is splitted
## Description
Fast file switch and fast symbol search do not work on the split panels. They only work in the first (main) panel, whether there is a split or not.

Fast file switch:
1. Open two or more files in the editor
2. Split the editor panel
3. Select the split panel (not the first main panel)
4. ctrl+p and select a different file than the current file in the panel

Fast symbol search:
1. Split the editor panel
2. Select the split panel (not the first panel)
3. ctrl+alt+p and select a symbol in the file to go to that line

Fast file switch:
expected: the selected file is shown the split panel
actual: the current file in the panel is not replaced

Fast symbol search:
expected: the line of the selected symbol should be highlighted in the split panel
actual: 
- always highlights the line in the first main panel, not in the split panel. 
- if the file in the split panel is different from the file in the first main panel, the symbol list is also of the main panel instead of the split panel

* Spyder Version: 3.2.1 
* Python Version: 3.5.3
* Qt Versions: 5.9.1, PyQt5 5.9 on Windows

## Dependencies
```
numpy >=1.7      :  1.13.1 (OK)
cython >=0.21    :  0.26 (OK)

```

jnsebgosselin:
Fixes #5057

- [x] Connect split editor to fileswitcher for file switching
- [x] Connect split editor to fileswitcher for symbol search
- [x] Make things work when the file order is not synchronised across split editorstacks

ctrl_p_fileswitcher

### DIFF
--- a/spyder/plugins/editor.py
+++ b/spyder/plugins/editor.py
@@ -1214,10 +1214,17 @@ class Editor(SpyderPluginWidget):
                 if win.isAncestorOf(editorstack):
                     self.set_last_focus_editorstack(win, editorstack)
 
-    #------ Handling editorstacks
+    # ------ Handling editorstacks
     def register_editorstack(self, editorstack):
         self.editorstacks.append(editorstack)
         self.register_widget_shortcuts(editorstack)
+        if len(self.editorstacks) > 1 and self.main is not None:
+            # The first editostack is registered automatically with Spyder's
+            # main window through the `register_plugin` method. Only additional
+            # editors added by splitting need to be registered.
+            # See Issue #5057.
+            self.main.fileswitcher.sig_goto_file.connect(
+                      editorstack.set_stack_index)
 
         if self.isAncestorOf(editorstack):
             # editorstack is a child of the Editor plugin
--- a/spyder/plugins/editor.py
+++ b/spyder/plugins/editor.py
@@ -2277,11 +2277,12 @@ class Editor(SpyderPluginWidget):
         self.__move_cursor_position(1)
 
     @Slot()
-    def go_to_line(self):
+    def go_to_line(self, line=None):
         """Open 'go to line' dialog"""
         editorstack = self.get_current_editorstack()
+        print(editorstack)
         if editorstack is not None:
-            editorstack.go_to_line()
+            editorstack.go_to_line(line)
 
     @Slot()
     def set_or_clear_breakpoint(self):
--- a/spyder/widgets/fileswitcher.py
+++ b/spyder/widgets/fileswitcher.py
@@ -541,7 +541,7 @@ class FileSwitcher(QDialog):
         return data
 
     def get_plugin_tabwidget(self, plugin):
-        """Get the data object of the plugin's current tab manager."""
+        """Get the tabwidget of the plugin's current tab manager."""
         # The tab widget is named "tabs" in the editor plugin while it is
         # named "tabwidget" in the notebook plugin.
         try:
--- a/spyder/widgets/fileswitcher.py
+++ b/spyder/widgets/fileswitcher.py
@@ -558,9 +558,8 @@ class FileSwitcher(QDialog):
         """Go to specified line number in current active editor."""
         if line_number:
             line_number = int(line_number)
-            editor = self.get_widget()
             try:
-                editor.go_to_line(min(line_number, editor.get_line_count()))
+                self.plugin.go_to_line(line_number)
             except AttributeError:
                 pass
 
@@ -616,6 +615,8 @@ class FileSwitcher(QDialog):
         # Get optional line number
         if trying_for_line_number:
             filter_text, line_number = filter_text.split(':')
+            if line_number == '':
+                line_number = None
             # Get all the available filenames
             scores = get_search_scores('', self.filenames,
                                        template="<b>{0}</b>")
@@ -625,7 +626,6 @@ class FileSwitcher(QDialog):
             # "fuzzy" matching
             scores = get_search_scores(filter_text, self.filenames,
                                        template="<b>{0}</b>")
-
 
         # Get max width to determine if shortpaths should be used
         max_width = self.get_item_size(paths)[0]
--- a/spyder/widgets/fileswitcher.py
+++ b/spyder/widgets/fileswitcher.py
@@ -295,7 +295,8 @@ class FileSwitcher(QDialog):
     @property
     def widgets(self):
         widgets = []
-        for tabs, plugin in self.plugins_tabs:
+        for plugin in self.plugins_instances:
+            tabs = self.get_plugin_tabwidget(plugin)
             widgets += [(tabs.widget(index), plugin) for
                         index in range(tabs.count())]
         return widgets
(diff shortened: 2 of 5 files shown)


ccordoba12:

spyder/plugins/editor.py
```diff
@@ -1214,10 +1214,16 @@ def save_focus_editorstack(self):
...
 
-    #------ Handling editorstacks
+    # ------ Handling editorstacks
     def register_editorstack(self, editorstack):
         self.editorstacks.append(editorstack)
         self.register_widget_shortcuts(editorstack)
+        if len(self.editorstacks) > 1 and self.main is not None:
+            # The first editostack is registered automatically with Spyder's
+            # main window through the `register_plugin` method. Only additional
+            # editors added by splitting need to be registered.
```
dalthviz: Please add the number of the issue here for reference, and I think this approach is ok :+1: Also, when testing this I realized there is similar issue, but with the symbol search (`Ctrl +Alt + p`), it works with the first panel only too. 

For the symbol search, one option to make it work is to try to get the current editor of the Editor plugin when using `goto_line` in the fileswitcher, I think something like this should work:

Change [the `goto_line` method](https://github.com/spyder-ide/spyder/blob/3.x/spyder/widgets/fileswitcher.py#L561) for:
```python
    def goto_line(self, line_number):
        """Go to specified line number in current active editor."""
        if line_number:
            line_number = int(line_number)
            try:
                editor = self.plugin.get_current_editor()
                editor.go_to_line(min(line_number, editor.get_line_count()))
            except AttributeError:
                pass
```
If you have any question let me know, and thanks for working in this one @jnsebgosselin! :)

jnsebgosselin: Thank you very much @dalthviz for reviewing this! I'll do the modifications and fix things for symbol search soon.

jnsebgosselin:
Ok, this has not been as easy as I had anticipated. The `FileSwitcher` structure was not well adapted to handle the editor plugin with multiple editorstacks, so I had to reworked a little bit the code. Here are the two most important changes I made:

1. For the symbol search functionality, instead of acting directly on the plugins in the [goto ](https://github.com/spyder-ide/spyder/pull/5221/files#diff-c1ea14606c18e2c1e228e0b1071dad65L557) method, I created the [sig_goto_line](https://github.com/spyder-ide/spyder/pull/5221/files#diff-c1ea14606c18e2c1e228e0b1071dad65R226) and used the same approach as for the [sig_goto_file ](https://github.com/spyder-ide/spyder/pull/5221/files#diff-c1ea14606c18e2c1e228e0b1071dad65L225) signal to connect it to each plugin.

2. The problem with the fileswitcher (in regard with split editor) is that it cannot acknowledge that the editor plugin can have multiple editorstack active at a time. To solve this, instead of using the info that are saved in `self.plugins_tabs` and `self.plugins_data` , we go get directly the `data` and `tab widget` from the plugin themselves using `plugin.get_current_tab_manager()`. Therefore, with the editor, we get the info regarding the current Editorstack and the fileswitcher then work as expected. 

On of the problem with that approach though, is that the plugin structure are not uniform, so the `data` and `tabWidget` are not accessed through the same name. See [here](https://github.com/spyder-ide/spyder/pull/5221/files#diff-c1ea14606c18e2c1e228e0b1071dad65R541).

I still have some refining and revision to do on this PR. I would just like to know if you guys (@ccordoba12  and @dalthviz ) are ok with the general approach I've used to solve this problem before I continue to work further on this.

jnsebgosselin:
Also, Issue #4865 will have an impact on this.

spyder/plugins/editor.py
jnsebgosselin: @dalthviz This was not working when tabs were moved in one of the editor and doing a fileswitch or a symbol search in the other. I had to work a little bit harder to make this work. See my comment below.

ccordoba12:
@jnsebgosselin, I don't have time to give you feedback right now, but you need to be aware that we plan to connect the file switcher not only to the Editor and notebook, but also to the IPython console and `spyder-terminal`. So you need to be sure that your changes are generalizable to *any* plugin that want to hook to the switcher.

jnsebgosselin:
>So you need to be sure that your changes are generalizable to any plugin that want to hook to the switcher.

I understand. I'll make some changes to be sure of that, I have an idea.

dalthviz:
Thanks a lot for your help in this one @jnsebgosselin !, about the name of some attributes needed for the fileswitcher probably we should standarize this kind of stuff in all the plugins as in the example of the name of the tabWidget. Maybe adding a function that a plugin should implement to return this kind of attributes?.

jnsebgosselin:
@dalthviz Thank you very much for your feedback. I was a little bit stalling on this one.

> About the name of some attributes needed for the fileswitcher probably we should standarize this kind of stuff in all the plugins as in the example of the name of the tabWidget. Maybe adding a function that a plugin should implement to return this kind of attributes?.

I was hoping you would say that! Something like:

```
@property
def tabWidget(self):
        """
        This must be reimplemented by plugins that need to connect to the FileSwitcher.
        """
        pass
```

jnsebgosselin:
@ccordoba12 @dalthviz I'm not happy with this PR, I'm working on it to propose something better. Please wait before reviewing again.

ccordoba12:
@jnsebgosselin, any progress on this one?

jnsebgosselin:
@ccordoba12 
I've reworked this PR to fix the issues that were present in the 3.x branch. I've tried to do it with a minimum amount of changes. I've tried to address one issue per commit so it should be easier for you to review.

I think that for spyder 4, some refactoring would be welcomed.

If you have time to review this today, I can make changes tonight so this can be part of the next release.

ccordoba12:
@dalthviz, please take a look at this one again.

spyder/widgets/fileswitcher.py
```diff
@@ -526,18 +529,35 @@ def get_stack_index(self, stack_index, plugin_index):
         return real_index
 
     # --- Helper methods: Widget
+    def get_plugin_data(self, plugin):
+        # The data object is named "data" in the editor plugin while it is
```
dalthviz: Please @jnsebgosselin add a doctring for this method.

spyder/widgets/fileswitcher.py
```diff
@@ -526,18 +529,35 @@ def get_stack_index(self, stack_index, plugin_index):
...
+        # named "clients" in the notebook plugin.
+        try:
+            data = plugin.get_current_tab_manager().data
+        except AttributeError:
+            data = plugin.get_current_tab_manager().clients
+
+        return data
+
+    def get_plugin_tabwidget(self, plugin):
```
dalthviz: Same here, please add a doctring.

ccordoba12:
@jnsebgosselin, please make the changes suggested by @dalthviz so we can merge this one.

ccordoba12:
@dalthviz, please add the missing docstrings to Jean's branch so we can merge this one. It seems Jean is busy and I want this one to be merged soon.

Jean, I hope it's ok with you :-)

jnsebgosselin:
@ccordoba12 sorry for that. I'm doing this right now, not that I have any problem with @dalthviz doing it, but I guess this is simpler if I do it.

ccordoba12:
Great! Thanks for the quick response, and hopefully I didn't bother you too much with this.

jnsebgosselin:
@ccordoba12 No not at all. I forgot to do it, so thanks for the reminder. 

September through March is a busier time for me, so I'll contribute less, but i'm still there and following what is going on.

ccordoba12:
Ok, thanks for letting us know about it.

Issue #5226, opened 2017-09-13
error when closing a file


impact27:
## Description

**What steps will reproduce the problem?**
1. ????
2. Press the cross in the editor to close the file

**Please provide any additional information below**

  File "/Users/quentinpeter/pyscripts/spyder/spyder/widgets/editor.py", line 1394, in close_file
    new_index = self._get_previous_file_index()
  File "/Users/quentinpeter/pyscripts/spyder/spyder/widgets/editor.py", line 1757, in _get_previous_file_index
    w_id = self.stack_history.pop(last)
  File "/usr/local/Cellar/python3/3.6.2/Frameworks/Python.framework/Versions/3.6/lib/python3.6/_collections_abc.py", line 993, in pop
    v = self[index]
  File "/Users/quentinpeter/pyscripts/spyder/spyder/widgets/editor.py", line 307, in __getitem__
    return self.id_list.index(self.history[i])
ValueError: 4342516136 is not in list

## Version and main components

* Spyder Version: 4.0.0.dev0 474e59379
* Python Version: 3.6.2
* Qt Versions: 5.9.1, PyQt5 5.9 on Darwin

## Dependencies
```
pyflakes >=0.6.0 :  1.6.0 (OK)
pycodestyle >=2.3:  2.3.1 (OK)
pygments >=2.0   :  2.2.0 (OK)
pandas >=0.13.1  :  0.20.3 (OK)
numpy >=1.7      :  1.13.1 (OK)
sphinx >=0.6.6   :  1.6.3 (OK)
rope >=0.9.4     :  0.10.5 (OK)
jedi >=0.9.0     :  0.10.2 (OK)
psutil >=0.3     :  5.2.2 (OK)
nbconvert >=4.0  :  5.2.1 (OK)
sympy >=0.7.3    :  1.1.1 (OK)
cython >=0.21    :  0.26 (OK)
qtconsole >=4.2.0:  4.3.0 (OK)
IPython >=4.0    :  6.1.0 (OK)
pylint >=0.25    :  1.7.2 (OK)

```

ccordoba12:
@rlaverde, please take a look at this one.

rlaverde:
I move it to 3.2.4 because this code is also present in 3.x

PR #5229, opened 2017-09-13
PR: Editor historystack improvements

rlaverde:
Fixes: #5226 

I think the error was the unnecessary side effect of `_get_previous_file_index`, although I wasn't able to reproduce #5226

### DIFF
--- a/spyder/widgets/editor.py
+++ b/spyder/widgets/editor.py
@@ -1723,12 +1723,11 @@ class EditorStack(QWidget):
                 pass
 
     def _get_previous_file_index(self):
-        if len(self.stack_history) > 1:
-            last = len(self.stack_history)-1
-            w_id = self.stack_history.pop(last)
-            self.stack_history.insert(0, w_id)
-
-            return self.stack_history[last]
+        """Return the penultimate element of the stack history."""
+        try:
+            return self.stack_history[-2]
+        except IndexError:
+            return None
 
     def tab_navigation_mru(self, forward=True):
         """
--- a/spyder/widgets/editor.py
+++ b/spyder/widgets/editor.py
@@ -304,7 +304,12 @@ class StackHistory(MutableSequence):
         return len(self.history)
 
     def __getitem__(self, i):
-        return self.id_list.index(self.history[i])
+        self._update_id_list()
+        try:
+            return self.id_list.index(self.history[i])
+        except ValueError:
+            self.refresh
+            raise IndexError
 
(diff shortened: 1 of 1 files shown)


spyder/widgets/editor.py
```diff
@@ -304,7 +304,12 @@ def __len__(self):
         return len(self.history)
 
     def __getitem__(self, i):
-        return self.id_list.index(self.history[i])
+        self._update_id_list()
+        try:
+            return self.id_list.index(self.history[i])
+        except ValueError:
+            self.refresh
```
ccordoba12: This should be `self.refresh()`, right?

spyder/widgets/editor.py
```diff
@@ -304,7 +304,12 @@ def __len__(self):
         return len(self.history)
 
     def __getitem__(self, i):
-        return self.id_list.index(self.history[i])
+        self._update_id_list()
+        try:
+            return self.id_list.index(self.history[i])
+        except ValueError:
+            self.refresh
+            raise IndexError
```
ccordoba12: Do we really need to raise an `IndexError` here?

spyder/widgets/editor.py
```diff
@@ -1718,12 +1723,11 @@ def current_changed(self, index):
                 pass
 
     def _get_previous_file_index(self):
-        if len(self.stack_history) > 1:
-            last = len(self.stack_history)-1
-            w_id = self.stack_history.pop(last)
-            self.stack_history.insert(0, w_id)
```
ccordoba12: Why did you remove this?

spyder/widgets/editor.py
rlaverde: Yes, thanks for noticing it

spyder/widgets/editor.py
rlaverde: I think it's the expected bevahior, when an elementary isn't found in a list an IndexError is raised.
The other option will be to return None.

spyder/widgets/editor.py
rlaverde: This was an unnecessary side effect, that could cause the corruption of the historystack order.

This reorder the editor stack before closing the file, the closing could be cancelled, an the editor stack will have a wrong order.

Also this was unnecessary, because when a file is closed the CurrentChanged  signal is be emited, and that cause the history_stack to be updated.

ccordoba12: Ok, thanks for the explanation :-)

PR #5230, opened 2017-09-13
PR: Support for sets in the Variable Explorer

### LINKED ISSUE #2355: Variable Explorer does not show sets
IPython >=0.13  :  3.0.0 (OK)
matplotlib >=1.0:  1.3.1 (OK)
pandas >=0.13.1 :  0.13.1 (OK)
pep8 >=0.6      :  1.6.2 (OK)
psutil >=0.3    :  2.2.1 (OK)
pyflakes >=0.6.0:  0.8.1 (OK)
pygments >=1.6  :  2.0.2 (OK)
pylint >=0.25   :  1.4.1 (OK)
rope >=0.9.2    :  0.9.4-1 (OK)
sphinx >=0.6.6  :  1.2.2 (OK)
sympy >=0.7.0   :  0.7.6 (OK)
zmq >=2.1.11    :  14.5.0 (OK)
Spyder 2.3.2
Python 3.4.0 64bits, Qt 4.8.6, PyQt4 (API v2) 4.10.4 on Linux

Steps to reproduce:

> x = {'a','b','c'}
> Look in variable explorer. No variable named x appears

Expected behavior:
show 'x' in variable explorer

Also:

xlist1 = [{'a','b','c'}] 
does not work, while,
xlist2 = ['foo',{'a','b','c'}]
does. 
screenshot-xlist2 - list 2 elements
The str representation looks good, but it it getting thr wrong length

Prikers:
Fixes #2355

----

`set` is now a supported type in the variable explorer. Now I chose the `Qt.darkGreen` color as it was not already used for other objects: if that is not ok there are some other colors left :smile: 
capture3
capture4

### DIFF
--- a/doc/variableexplorer.rst
+++ b/doc/variableexplorer.rst
@@ -62,6 +62,7 @@ supported are:
 #. Floats
 #. Complex numbers
 #. Lists
+#. Sets
 #. Dictionaries
 #. Tuples
 #. Strings
--- a/spyder/config/base.py
+++ b/spyder/config/base.py
@@ -380,7 +380,7 @@ def get_supported_types():
     If you update this list, don't forget to update doc/variablexplorer.rst
     """
     from datetime import date
-    editable_types = [int, float, complex, list, dict, tuple, date
+    editable_types = [int, float, complex, list, set, dict, tuple, date
                       ] + list(TEXT_TYPES) + list(INT_TYPES)
     try:
         from numpy import ndarray, matrix, generic
--- a/spyder/utils/ipython/spyder_kernel.py
+++ b/spyder/utils/ipython/spyder_kernel.py
@@ -131,6 +131,7 @@ class SpyderKernel(IPythonKernel):
                 properties[name] = {
                     'is_list':  isinstance(value, (tuple, list)),
                     'is_dict':  isinstance(value, dict),
+                    'is_set': isinstance(value, set),
                     'len': self._get_len(value),
                     'is_array': self._is_array(value),
                     'is_image': self._is_image(value),
--- a/spyder/widgets/variableexplorer/collectionseditor.py
+++ b/spyder/widgets/variableexplorer/collectionseditor.py
@@ -1525,6 +1525,7 @@ def get_test_data():
             'str': 'kjkj kj k j j kj k jkj',
             'unicode': to_text_string('éù', 'utf-8'),
             'list': [1, 3, [sorted, 5, 6], 'kjkj', None],
+            'set': {1, 2, 1, 3, None, 'A', 'B', 'C', True, False},
             'tuple': ([1, testdate, testdict], 'kjkj', None),
             'dict': testdict,
             'float': 1.2233,
--- a/spyder/widgets/variableexplorer/collectionseditor.py
+++ b/spyder/widgets/variableexplorer/collectionseditor.py
@@ -5,7 +5,7 @@
 # (see spyder/__init__.py for details)
 
 """
-Collections (i.e. dictionary, list and tuple) editor widget and dialog
+Collections (i.e. dictionary, list, set and tuple) editor widget and dialog
 """
 
 #TODO: Multiple selection: open as many editors (array/dict/...) as necessary,
@@ -144,6 +144,10 @@ class ReadOnlyCollectionsModel(QAbstractTableModel):
         elif isinstance(data, list):
             self.keys = list(range(len(data)))
             self.title += _("List")
+        elif isinstance(data, set):
+            self.keys = list(range(len(data)))
+            self.title += _("Set")
+            self._data = list(data)
         elif isinstance(data, dict):
             self.keys = list(data.keys())
             self.title += _("Dictionary")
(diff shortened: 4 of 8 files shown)


rlaverde:
I think that sets should be editable, How difficult will be to edit `sets`? (I tried changing readonly to True, but I convert sets to lists.)

goanpeca:
>I think that sets should be editable 

Sets cannot be edited by index:
```
python
a = {1, 2}
a[0]

  File "<ipython-input-10-5ccf417d7af1>", line 1, in <module>
    a[0]

TypeError: 'set' object does not support indexing
```

goanpeca:
@Prikers thanks for the work. Now, why does simple_set says `List` on the header, also, sets by definition are not ordered containers, so displaying an index is not correct cause sets are not even indexable. However I think users would expect to be able o add and remove items, any chance we could include that?

rlaverde:
>sets cannot be edited by index:

But you could replace an element

```
In[8]: a = {10, 12, 15}
In[9]: a.remove(15)
In[10]: a.add(23)
In[11]: a
Out[12]: {10, 12, 23}
```

I don't know how the internals of variable explorer works, and how easy something like this could be do it

goanpeca:
>I don't know how the internals of variable explorer works, and how easy something like this could be do it

That is what I am suggesting, removing and adding is not the same as editing. And emulating that behavior is conceptually wrong and leads to believing they can be accessed by index.

Prikers:
1. Hum... Good catch @goanpeca regarding the title! It is actually more subtle than I anticipated. The problem is that, as you said, `set` does not support indexing. The code heavily depends on indexes for displaying collections so I had converted `set` into a list under the hood in order not to reimplement everything. Turns out that is why the title is `List`. I now do not think turning the set into a list is the best option. What do you think? Apart from the title (which could be manually fixed) is it a bad idea?

2. @rlaverde: It was on purpose that I made it not editable, but you are right, unlike `tuple` we can add or remove elements. This should definitely be implemented. Thanks for catching that!

goanpeca:
>I now do not think turning the set into a list is the best option. What do you think? Apart from the title (which could be manually fixed) is it a bad idea?

I understand the temptation to transform into a list, but I think we should not do this but generalize the concept for a set or even create a new editor just for handling sets, which is a non indexable non sorted container, mutable container

Prikers:
That's wise :smile: 
Then should I create a new file `seteditor.py` to create a new editor or simply adjust the `collectioneditor` to handle the non indexable objects by checking `if isinstance(data, set)` whenever it is necessary?

goanpeca:
I think we should be able to use the collections editor. But instead of checking every time if it is a given object, I think it would be better to define characteristics for an object and act accordingly.

`is_mutable`
`is_ordered`

But what do you think @ccordoba12?

ccordoba12:
I think what @Prikers has done is the best we can do for now. Before merging, I'd like to see these improvements :

1. Use `Set` instead of `List` in the title of the viewer (as already mentioned).
2. Remove the `Index` column when displaying sets.

ccordoba12:
> That is what I am suggesting, removing and adding is not the same as editing.

None of our editors allow to add/remove new content to an object, so what @Prikers did (making the contents read-only) is the right thing to do for sets.

> But instead of checking every time if it is a given object, I think it would be better to define characteristics for an object and act accordingly.

My plan for beta2 or beta3 is to use the [multipledispatch](https://github.com/mrocklin/multipledispatch) library to define the behavior of several functions in the Variable Explorer according to the object's type.

Prikers:
> Use Set instead of List in the title of the viewer (as already mentioned).
Remove the Index column when displaying sets.

Will do!

> None of our editors allow to add/remove new content to an object

That's what I thought but it turns out it is possible :open_mouth: 
Anyway if the whole thing will be refactored it may not be necessary to implement too many features
dict

Prikers:
Exploring a bit more I noted a bug:

- Create a `tuple` objet
- Open it in the variable explorer
- Right click on a line to get the same menu as in the above gif
- Click on `rename` or `duplicate` gerenates an error in the internal console
```
  File "C:\...\spyder\widgets\variableexplorer\collectionseditor.py", line 978, in rename_item
    self.copy_item(True)
  File "C:\...\spyder\widgets\variableexplorer\collectionseditor.py", line 961, in copy_item
    QLineEdit.Normal, orig_key)
TypeError: getText(QWidget, str, str, echo: QLineEdit.EchoMode = QLineEdit.Normal, text: str = '', flags: Union[Qt.WindowFlags, Qt.WindowType] = Qt.WindowFlags(), inputMethodHints: Union[Qt.InputMethodHints, Qt.InputMethodHint] = Qt.ImhNone): argument 5 has unexpected type 'int'
```

ccordoba12:
@Prikers, for now please finish your support for sets with my recommendations. Then you can open a new PR to fix the error you found and add support for add/remove elements to sets.

Prikers:
[*sorry for the delay*] I have hidden the first column (`Index`) for sets and corrected the issue with the title. I will create another issue/PR for the add/remove support. 
capture5

Now I noticed that `Set` is not part of the .pot files for translations. I have locally another commit with the spyder/locale/spyder.pot file modified (and the one in French as well) but I am not sure I should do it. Are these .pot files automatically generated at some specific point in time?

ccordoba12:
> Now I noticed that Set is not part of the .pot files for translations.

Translations will be updated when we are about to release Spyder 4, to not ask our translators to constantly update a moving target (because we can add/remove a lot of translation strings during our betas).

So, don't worry about this ;-)

PR #5260, opened 2017-09-17
PR: Add support for object arrays in the Variable Explorer

### LINKED ISSUE #8834: Hard crash to desktop when opening Numpy array in VarExp  with one or more np.object fields

@ccordoba12 I get a consistently reproducible (including under ``--safe-mode``) Spyder hard crash to desktop on the latest ``master``, trying to open a Numpy record array (``np.void`` type) on which at least one of the sub-arrays is of the ``np.object`` type. Can be easily repro'd by attempting to open e.g. `` x = np.array([("spam", 2), ("eggs", 4)], dtype=[('x', np.object), ('y', int)])``in the Variable Explorer. The exception is:

```python-traceback
  File "C:\Users\C. A. M. Gerlach\Documents\dev\SpyderDev\spyder\spyder\plugins\variableexplorer\widgets\arrayeditor.py", line 279, in data
    / (float(self.vmax) - self.vmin))
ValueError: could not convert string to float: 'fsd'
```

Curiously, what actually triggers the hard crash, though, appears to be something to do with the resulting error dialog as if I enable ``--debug`` mode I don't get the Qt crash, just the exception printed to the terminal, and the crash happens a few hundred ms after the error dialog appears. The console messages (debug off) that appear simultanious to the hard crash are:

```
QWidget::repaint: Recursive repaint detected
QBackingStore::endPaint() called with active painter on backingstore paint device
```
   
 
## Proposed solution

 The root cause of the exception that initially triggers the error dialog (which actually causes the crash itself) is Spyder supporting record arrays just fine, but not any object arrays they contain.

Therefore, the simplest way to handle this would just be to check the types of the contained fields for np.void arrays to ensure none of them are ``np.object``  and either exclude them from the popup list at bottom left of the ArrayEditor (possibly with a warning on opening the array listing the object fields that could not be viewed), or show an error dialog when trying to switch to them (just like we do for object arrays opened from the other viewer/editors). If all the array's fields are np.object, then it we make it behave exactly as if the array itself were np.object (error dialog).

## Versions

* Spyder version: 4.0.0.dev0 6a918f8b5
* Python version: 3.7.2
* Qt version: 5.9.6
* PyQt5 version: 5.9.2

### Dependencies

```
sympy >=0.7.3     :  1.3 (OK)
```

ccordoba12:
After PR #5259, I realized it's not that hard to support object arrays.

TODO:

- [x] Use `value_to display` to show the contents of the array (else big repr's would make the viewer to take a long time to appear).
- [x] Tests.

----

Note: When developing this PR I also noticed how to fix issue #8834.

Fixes #8834.


steff456:
@ccordoba12 The main idea for these PR is to visualice in the object explorer np arrays when they are made up of objects?

As well, I'm going to create a new PR for solving this one because the merge with our current master is huge and has a lot of conflicts.

PR #5262, opened 2017-09-17
PR: Update version in splash screen

### LINKED ISSUE #5256: Update spyder splash for spyder4
Th splash still says spyder 3

Are we going to update the logo two? Maybe for the logo we are using in https://spyder-ide.github.io/

@ccordoba12 @goanpeca

rlaverde:
Fixes #5256 

This is how this looks like:

spectacle xm1713


ccordoba12:
Nice!! beta1 is closer and closer to be a reality :-)

Issue #5131, opened 2017-09-04
Sources checkable preferences aren't sync with preferences


rlaverde:
## Description of your problem

- Changing blank spaces, or indent guides from `Source` menu doesn't update preference
- Changing blank spaces, or indent guides in preferences doesn't update preference `Source` menu

**What steps will reproduce the problem?**

1. Activate/deactivate show blank spaces, or show indent guides form source menu
2. Go to Tool>Preferences>Editor
3. The changes aren't reflected
4. Restart spyder, the changes wasn't saved

--
1. Go to Tool>Preferences>Editor
2. Activate/deactivate show blank spaces, or show indent guides 
3. Go to source menu, the changes aren't reflected

* Spyder Version: It happens in spyder3 and 4, with the difference that indent guides isn't present in spyder3

I'm not sure what is the best way to fix this, after #5130 and #5122 it will be four different checkable options, implementing each one independently will cause some duplicated code.

ccordoba12:
@rlaverde, we need to solve this one for beta1.

PR #5274, opened 2017-09-18
PR: Remove duplicated logic in Editor checkable Actions

rlaverde:
Fixes: #5131

I think this could be improved but that will need a refactoring of how actions are created, `create_action` logic could be moved to `__init__` method of `SpyderAction` and that will allow to inherit from that class, and create a  `CheckableEditorAction` and abstract the logic in that class. (Although that is out of the scope of this PR)


ccordoba12:
It looks really good!!

@jnsebgosselin, @goanpeca, what do you think about this one?

PR #5276, opened 2017-09-18
PR: Move external plugins inside the main package again

rlaverde:
Missing:

- Merge and delete translations files

~I manage to update the paths using `sed`, and merge `pot` files with `msgcat`, but trying to merge po files with `msgcat` results in a lot of lines modified because line wrap change~ I dropped that commits

@ccordoba12 I think I need help merging the translations, I'm not familiar with translation utilities and I don't know what is the best way to merge them

----

I leave the load of the plugins dynamically using `importlib` see e2e2a6aad96f4e4cce9a532ae44e8a2358919fc6
I think dynamically importing is better because importing each one explicitly (as other plugins do) will add a lot of duplicated code


ccordoba12:
Please revert all changes made to the translation files. I think we will have to merge them differently once we regenerate our pot files.

rlaverde:
>Please revert all changes made to the translation files. I think we will have to merge them differently once we regenerate our pot files.

Ok, I will drop those commits

goanpeca:
Is the plan to also modularize the `locale` for each plugin? or only for plugins that do not live in the spyder repo

rlaverde:
Is the plan to also modularize the locale for each plugin? or only for plugins that do not live in the spyder repo

It'll be nice to modularize it too, as our idea is to make plugins independent; although It'll add duplicated translations,  and I don't know how easy will to manage several translation files instead of one

ccordoba12:
> Is the plan to also modularize the locale for each plugin?

Nop, that's not the plan. It's simpler for translators to have a single file instead of 10 or more to translate.

spyder/app/mainwindow.py
```diff
@@ -905,6 +906,17 @@ def create_edit_action(text, tr_text, icon):
...
+        # Load other plugins (former external plugins)
+        other_plugins = ['breakpoints', 'profiler', 'pylint']
+        for plugin_name in other_plugins:
+            if CONF.get(plugin_name, 'enable'):
+                module = importlib.import_module(
+                        'spyder.plugins.{}'.format(plugin_name))
+                plugin = module.PLUGIN_CLASS(self)
+                if plugin.check_compatibility()[0]:
+                    self.thirdparty_plugins.append(plugin)
+                    plugin.register_plugin()
```
ccordoba12: This shouldn't be necessary now.

spyder/plugins/breakpoints/__init__.py
```diff
@@ -3,4 +3,4 @@
 #==============================================================================
 # The following statement is required to register this 3rd party plugin:
 #==============================================================================
-from .breakpoints import Breakpoints as PLUGIN_CLASS
```
ccordoba12: Please remove this line from all plugins. This is not necessary anymore because they are now core plugins.

ccordoba12:
Please rebase and finish this one.

spyder/plugins/breakpoints/__init__.py
rlaverde: I think instead of removing it from these plugins, we could add it to other internal plugins, also with the variables for config files that @goanpeca suggested in https://github.com/spyder-ide/spyder/issues/4591#issuecomment-330835315

I prefer that internal and external plugins have the same structure, so in a future we could refactor and load all plugins in the same way.

spyder/app/mainwindow.py
rlaverde: I think this is necessary, It I remove it I should move the logic of register_plugin to the mainwindow (and personally I don't like that approach)

spyder/plugins/breakpoints/__init__.py
goanpeca: I agree with @rlaverde

rlaverde:
This is ready, I added a TODO, for some refactoring of other plugins, also moving the translations is missing, I 'll address them in other PRs

Note: Moving translations x should be do it before merging `split-plugins` with master, because it raises some errors when trying to import the old directories as modules.

spyder/plugins/breakpoints/__init__.py
ccordoba12: We discussed this with @rlaverde yesterday and agreed that it'd be nice to have for all plugins indeed.

PR #5290, opened 2017-09-20
PR: Prevent unchecked paths in the pathmanager to be added to the IPython console sys.path

### LINKED ISSUE #5289: Checking/Unchecking path in the pathmanager is not working as expected in the IPython console

1. In the pathmanager, add a path to spyder pythonpath
2. Open a new ipython console and check that the path was added to `sys.path`
3. Go back in the pathmanager and uncheck the path that was just added.
4. Open a new iptyhon console and check that the unchecked paths were removed with `sys.path`.

pythonpath

Expected:
The unchecked paths should not appear in `sys.path` when unchecked in the pathmanager.

What happens:
The unchecked paths are not removed from `sys.path` in the IPython console, nor in the internal console.

This problem is due to two causes. The first cause is that the [remove_path_from_sys_path ](https://github.com/spyder-ide/spyder/blob/master/spyder/app/mainwindow.py#L2496) in `mainwindow` is not robust enough. In my environment, a path is added at index 0 in `sys.path`, resulting in `remove_path_from_sys_path` to actually removing nothing from it. This explains why this was not working for the internal console.

spyder_syspath

The second cause is that unchecking path in the pathmanager only affect `sys.path` in `mainwindow`. The [get_spyder_pythonpath ](https://github.com/spyder-ide/spyder/blob/master/spyder/app/mainwindow.py#L2486) still returns all checked and unchecked path from the pathmanager. This method needs to return instead only the active spyderpaths so that the IPython console plugin can set its own pythonpath correctly when calling [self.main.get_spyder_pythonpath](https://github.com/spyder-ide/spyder/blob/master/spyder/plugins/ipythonconsole.py#L1388) (I think this is what is going on, @ccordoba12  can you confirm this please?). 

I naively assumed when implementing this in PR #4756 that updating `sys.path` at the mainwindow level was enough. It was working with the python console when I tested it. I will submit a PR that will address both of the aforementioned issues. 

However, a third issue still remains, which is not critical. When unchecking paths in the pythonpath manager, the changes will only take effect in **newly created IPython console.** The pythonpath of the already opened IPython console won't be updated. Is this acceptable? @ccordoba12 Is there a way to refresh the IPython console pythonpath, like it seems to be done when opening a spyder project?

* Spyder Version: branch 3.x and master
* Python Version: 3.6.1
* Qt Version: 5.9.1
* PyQt Version: 5.9
* Operating system: Windows 7 64 bits

## Dependencies

jnsebgosselin:
Fixes #5289
Related to Issue #1105 and PR #4756

----
### Changes summary

- Made the [remove_path_from_sys_path ](https://github.com/spyder-ide/spyder/blob/master/spyder/app/mainwindow.py#L2496) method more robust

- Made the [get_spyder_pythonpath ](https://github.com/spyder-ide/spyder/blob/master/spyder/app/mainwindow.py#L2486)  method return  only the active spyderpaths.

----

make_uncheck_path_work

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -2486,22 +2486,23 @@ class MainWindow(QMainWindow):
         elif osp.isfile(osp.join(CWD, fname)):
             self.open_file(osp.join(CWD, fname), external=True)
 
-    #---- PYTHONPATH management, etc.
+    # ---- PYTHONPATH management, etc.
     def get_spyder_pythonpath(self):
         """Return Spyder PYTHONPATH"""
-        return self.path+self.project_path
+        active_path = [p for p in self.path if p not in self.not_active_path]
+        return active_path+self.project_path
 
     def add_path_to_sys_path(self):
         """Add Spyder path to sys.path"""
         for path in reversed(self.get_spyder_pythonpath()):
-            if path not in self.not_active_path:
-                sys.path.insert(1, path)
+            sys.path.insert(1, path)
 
     def remove_path_from_sys_path(self):
         """Remove Spyder path from sys.path"""
         sys_path = sys.path
-        while sys_path[1] in self.get_spyder_pythonpath():
-            sys_path.pop(1)
+        for path in self.path:
+            while path in sys_path:
+                sys_path.remove(path)
 
     @Slot()
     def path_manager_callback(self):
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -2490,7 +2490,7 @@ class MainWindow(QMainWindow):
     def get_spyder_pythonpath(self):
         """Return Spyder PYTHONPATH"""
         active_path = [p for p in self.path if p not in self.not_active_path]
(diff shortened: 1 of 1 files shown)


spyder/app/mainwindow.py
```diff
@@ -2486,22 +2486,23 @@ def open_external_file(self, fname):
         elif osp.isfile(osp.join(CWD, fname)):
             self.open_file(osp.join(CWD, fname), external=True)
 
-    #---- PYTHONPATH management, etc.
+    # ---- PYTHONPATH management, etc.
     def get_spyder_pythonpath(self):
         """Return Spyder PYTHONPATH"""
-        return self.path+self.project_path
+        active_path = [p for p in self.path if p not in self.not_active_path]
+        return active_path+self.project_path
```
goanpeca: `return active_path + self.project_path`

ccordoba12:
@jnsebgosselin, please merge this PR with the latest 3.x to fix the errors on Circle and Appveyor.

PR #5309, opened 2017-09-21
PR: Update font decorations when adding them to the editor.

rlaverde:
See https://github.com/spyder-ide/spyder/pull/4751#issuecomment-331008782

@andfoy could you test  if this fix the error with underlined decorations?

### DIFF
--- a/spyder/widgets/sourcecode/codeeditor.py
+++ b/spyder/widgets/sourcecode/codeeditor.py
@@ -942,7 +942,7 @@ class CodeEditor(TextEditBaseWidget):
 
     def __highlight_selection(self, key, cursor, foreground_color=None,
                         background_color=None, underline_color=None,
-                        underline_style=QTextCharFormat.SpellCheckUnderline,
+                        underline_style=QTextCharFormat.WaveUnderline,
                         update=False):
         extra_selections = self.get_extra_selections(key)
         selection = TextDecoration(cursor)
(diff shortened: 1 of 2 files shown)


ccordoba12:
Please rebase on top of the latest master to fix the errors in Circle and Travis.

andfoy:
@rlaverde I'm afraid to say that the update is not working with most of the monospaced fonts available on my system.

rlaverde:
@andfoy the  change that you did in https://github.com/spyder-ide/spyder/pull/4751/commits/f177bda7ba615cfb4aa7573c4189ecf18cb51352 sending the font to TextDecoration works? I based this change PR in that change.

>I'm afraid to say that the update is not working with most of the monospaced fonts available on my system.

Yes, It only fixed it for some fonts 😞 also this error happens with 3.x I'll debug even more and try to find a better solution

andfoy:
> sending the font to TextDecoration works?

I think that changing the font may work for some fonts. However, I haven't found which parameter should be adjusted to achieve full compatibility with all monospaced fonts. In summary, is not working for all fonts.

rlaverde:
I think I found the bug: some fonts doesn't support `QTextCharFormat.SpellCheckUnderline` (I don't know way), using `QTextCharFormat.WaveUnderline` (or another style) works fine
http://doc.qt.io/qt-5/qtextcharformat.html#UnderlineStyle-enum

@andfoy Could you test it again? 🙈

andfoy:
@rlaverde It is now working at last! I think that outlined selections are not working well, but I guess this ii font-dependent also

PR #5320, opened 2017-09-22
PR: Fix editing floats in array editor

### LINKED ISSUE #5003: Error when trying to edit elements of a float numpy array in variable explorer
## Error when trying to edit elements of a float numpy array in variable explorer

1. Doubleclick a float numpy array in variable explorer 
2. Try to edit the fractional part of any element

The value of the chosen element is expected to be changed, but it is not.

* Spyder Version: 3.2.1 
* Python Version: 3.6.1

## Dependencies
```
sympy >=0.7.3    :  1.0 (OK)

```

timhoffm:
Fixes #5003 

By default, the validator uses the system locale, which may use a comma as a decimal point. I don't think, we'd want localization here. Python itself does always use a decimal point and that's what we should do as well.

### DIFF
--- a/spyder/widgets/variableexplorer/arrayeditor.py
+++ b/spyder/widgets/variableexplorer/arrayeditor.py
@@ -18,7 +18,7 @@ from __future__ import print_function
 
 # Third party imports
 from qtpy.compat import from_qvariant, to_qvariant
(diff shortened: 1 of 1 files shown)


ccordoba12:
@timhoffm, thanks a lot for helping us to solve this problem!! It'll be part for 3.2.4 :-)

Issue #5099, opened 2017-08-31
Closing last open file while being in non-existing working dir raises exception


kohr-h:
## Description

When closing the last file in the editor while the working directory is non-existing (e.g. after changing git branch) I get this exception:

```
File "[...]/spyder/plugins/editor.py", line 1746, in new
    basedir = getcwd()
FileNotFoundError: [Errno 2] No such file or directory
```

The same happens when I immediately try to open a file without changing the working dir.

**Steps to reproduce the problem**

1. Make new directory and make it the working directory. Have a file open.
2. Remove the directory on disk.
3. Close all files in the editor.

## Version and main components

* Spyder Version: 3.2.1 
* Python Version: 3.6.2
* Qt Versions: 5.6.2, PyQt5 5.6 on Linux

## Dependencies
```
pyflakes >=0.6.0 :  1.6.0 (OK)
pycodestyle >=2.3:  2.3.1 (OK)
pygments >=2.0   :  2.2.0 (OK)
pandas >=0.13.1  :  None (NOK)
numpy >=1.7      :  1.12.1 (OK)
sphinx >=0.6.6   :  1.6.3 (OK)
rope >=0.9.4     :  0.9.4-1 (OK)
jedi >=0.9.0     :  0.10.2 (OK)
psutil >=0.3     :  5.2.2 (OK)
nbconvert >=4.0  :  5.2.1 (OK)
sympy >=0.7.3    :  None (NOK)
cython >=0.21    :  None (NOK)
qtconsole >=4.2.0:  4.3.1 (OK)
IPython >=4.0    :  6.1.0 (OK)
pylint >=0.25    :  1.7.2 (OK)

```

ccordoba12:
Thanks for reporting. We'll take a look at this after we release 3.2.2

ccordoba12:
@rlaverde, please work on this one.

PR #5336, opened 2017-09-26
PR: Add a safe getcwd function to handle errors when the current working was deleted

rlaverde:
Fixes: #5099

I think that returning home directory is better than returning None, because the calls of getcwd expect a directory, and returning None will cause other errors

Maybe a warning should be printed when falling back to home directory

I change it everywhere, except the tests and ipykernel file


ccordoba12:
This needs a rebase now and please also fix the pep8 issues reported above.

PR #5337, opened 2017-09-26
PR: Display all lists in the variable explorer

### LINKED ISSUE #5313: Not all lists are supported by the variable explorer

1. Create two lists `a = [len, 1]` and `b = [1, len]` 
2. Check the variable explorer
3. `a` is not supported while `b` is

Both `a` and `b` are lists. They should be both supported.

Prikers:
Fixes #5313

---------
To fix this issue we can remove the chunk of code that is under `elif iterate:` (in [`is_supported`](https://github.com/spyder-ide/spyder/blob/master/spyder/widgets/variableexplorer/utils.py#L511) function) but I did not feel confortable about it. There may be a reason for this piece of code that I am not able to see right now. 
As a compromise I just turned it off by default but it is still there in case we figure out why it was here in the first place.
Does it seem acceptable?

### DIFF
--- a/spyder/widgets/variableexplorer/tests/test_utils.py
+++ b/spyder/widgets/variableexplorer/tests/test_utils.py
@@ -114,9 +114,9 @@ def test_list_display():
 
     # List starting with a non-supported object (#5313)
     supported_types = tuple(get_supported_types()['editable'])
-    l = [len, 1]
-    assert value_to_display(l) == '[builtin_function_or_method, 1]'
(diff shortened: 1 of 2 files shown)


goanpeca:
Awesome @Prikers is this ready :-) ?

Prikers:
@goanpeca yes, I think it does what was expected. 
```python
l1 = [len, max, min]
l2 = [len, 1, 2, 3]
l2 = [l2, 4, 5]
```
image

PR #5341, opened 2017-09-27
PR: Trade ipykernel.serialize for cloudpickle and remove use of publish_data

### LINKED ISSUE #5330: Error in variable explorer for dicts with mixed-type keys

1. Create a dict with `str` and `int` keys: `d = {'1': 1, 2: 2}`
2. Open it in the variable explorer
3. Click ok to close the window: error pops-out

```python
  File "...\spyder\widgets\variableexplorer\collectionseditor.py", line 527, in <lambda>
    lambda eid=id(editor): self.editor_accepted(eid))
  File "...\spyder\widgets\variableexplorer\collectionseditor.py", line 549, in editor_accepted
    self.set_value(index, conv_func(value))
  File "...\spyder\widgets\variableexplorer\collectionseditor.py", line 1370, in set_value
    self.set_value_func(name, value)
  File "...\spyder\widgets\variableexplorer\namespacebrowser.py", line 304, in set_value
    value = serialize_object(value)
  File "...\ipykernel\serialize.py", line 85, in serialize_object
    for k in sorted(obj):
TypeError: '<' not supported between instances of 'int' and 'str'
```

* Spyder Version: 3.2.3
* Python Version: 3.6

## Dependencies
sympy >=0.7.3    :  1.0 (OK)

### LINKED ISSUE #5830: Variable Explorer is leaking memory when editing variables

I am loading some .txt files into Spyder as arrays with numpy. The files load fine and I can enter editing mode in the variable explorer once the files have loaded, and there are no issues. However, upon exiting editing mode, it only takes a couple of seconds before one of the python processes already running (visible in task manager) starts to take up an increasing amount of memory (until all available memory is used).  This does only happen for files of a large size. I am not sure where the threshold is, but the stylized reproduction example below should provide some indication (at least in my case). 

I have tried to search for a solution both on GitHub and stackoverflow, but without any luck. However, should I have missed anything please point me in the direction of the solution. Also, I am very new to both Python and Spyder (3 days), so please forgive me if this is a stupid question. 

1. Create two .txt files that are each are a vector of zeroes separated by spaces, but with different number of entries (78396 and 1157770). 
2. Load two .txt files with as arrays with numpy using: txtfile = np.loadtxt('file.txt',delimiter = ' ')
3. Enter editing mode for the smaller array, then exit - nothing interesting should happen.
4. Enter editing mode for the larger array, then exit - python process should start to eat up memory

* Spyder Version: 3.2.4
* Python Version: 3.6.3
* Qt Version: 5.6.2
* PyQt Version: 5.6.0
* Operating system: Microsoft Windows 10 Pro

## Dependencies

minrk:
- use custom `spyder_msg` message type rather than IPython Parallel-specific `data_pub`.
  This removes the need for special `__spy__` names because it is no longer a shared namespace.
- use `cloudpickle` for serializing objects rather than deprecated `ipykernel.serialize`
  that is reserved for IPython Parallel.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,7 +11,7 @@ environment:
     CONDA_DEPENDENCIES_FLAGS: "--quiet"
     CONDA_DEPENDENCIES: >
       rope pyflakes sphinx pygments pylint pycodestyle psutil nbconvert
-      qtawesome pickleshare pyzmq chardet mock pandas pytest
+      qtawesome cloudpickle pickleshare pyzmq chardet mock pandas pytest
       pytest-cov numpydoc scipy pillow qtconsole matplotlib jedi pywin32
     PIP_DEPENDENCIES: "pytest-qt pytest-mock pytest-timeout flaky"
 
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -33,6 +33,7 @@ requirements:
     - pyzmq
     - chardet >=2.0.0
     - numpydoc
+    - cloudpickle
 
 about:
   home: https://github.com/spyder-ide/spyder
--- a/continuous_integration/circle/install.sh
+++ b/continuous_integration/circle/install.sh
@@ -3,7 +3,7 @@
 export CONDA_DEPENDENCIES_FLAGS="--quiet"
 export CONDA_DEPENDENCIES="rope pyflakes sphinx pygments pylint psutil nbconvert \
                            qtawesome pickleshare qtpy pyzmq chardet mock nomkl pandas \
-                           pytest pytest-cov numpydoc scipy cython pillow"
+                           pytest pytest-cov numpydoc scipy cython pillow cloudpickle"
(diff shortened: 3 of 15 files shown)


ccordoba12:
@minrk, thanks a lot for taking the time to help us improve our architecture! You know better than anyone else the Jupyter/IPython intricacies, so we really appreciate it.

We were thinking to use `dill` instead of `cloudpickle` to improve serialization, because `dill` has also the ability to restore a full IPython interpreter session, something we'd like to add in the future.

Any particular advice for using `cloudpickle`?

minrk:
No, I don't think there's a big separator between dill and cloudpickle. cloudpickle's just my default choice these days. No objection to switching to dill.

ccordoba12:
Thanks a lot @minrk for this! It's a really great contribution!!

I decided it to move it to our stable branch to avoid Spyder to fail when `ipykernel` 5.0 is released. I also fixed several issues with Python 2.

ccordoba12:
Pinging @ghisvail about this one because it adds `cloudpickle` as a new dep. We hope to not add more new deps in 3.x, but this one was really necessary.

Issue #5326, opened 2017-09-25
Improve debug logging


rlaverde:
Improve logging could help with better error reporting #5247, and won't be too hard to implement. (also will help a lot when debugging spyder)

At the moment the only way to debug/log information in spyder is `debug_print`, and using DEBUG variable

I think we could move to use `logging` library, It'll allow us to configure levels, handlers, filters, and support scenarios like:

- Debug specific plugins: something like: spyder --debug=editor could be achieved using filters

Related to #5247
- Append last 5 debug loggings to an error report
- Log all errors to a log file that could be requested when reporting an error
- Even we could attach a socket handler to a webservice and report errors in the background.

cc @goanpeca @ccordoba12

ccordoba12:
I like it, but let's leave it for beta2.

PR #5346, opened 2017-09-27
PR: Change error report dialog to enter issue description before submitting it

### LINKED ISSUE #5247: Improve interface to report errors
It seems browsers only accept a limited number of characters in the status bar, and so there are several issues reported by our users that end up being incomplete.

So, instead of that, we should use the Github API to create new issues:

https://developer.github.com/v3/issues/#create-an-issue

@rlaverde, please take care of this one.

rlaverde:
Fixes #5247.

----

It looks like this, the ok button is activated when each text has more than 10 words, this has to be improved because it isn't obvious how the ok button will be activated

Also a button to dismiss and don't show the dialog again should be added 

spectacle p15381

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -2380,7 +2380,7 @@ class MainWindow(QMainWindow):
         dlg.set_data(dependencies.DEPENDENCIES)
         dlg.exec_()
 
-    def render_issue(self, traceback=""):
+    def render_issue(self, description, traceback):
         versions = get_versions()
         # Get git revision for development version
         revision = ''
@@ -2389,38 +2389,35 @@ class MainWindow(QMainWindow):
         issue_template = """\
 ## Description
 
-**What steps will reproduce the problem?**
+{description}
 
-1. 
-2. 
-3. 
+## Traceback
 
-**What is the expected output? What do you see instead?**
-
-
-**Please provide any additional information below**
-
-%s
+```python-traceback
+{traceback}
+```
 
 ## Version and main components
 
-* Spyder Version: %s %s
-* Python Version: %s
-* Qt Versions: %s, %s %s on %s
+* Spyder Version: {spyder_version} {commit}
+* Python Version: {python_version}
+* Qt Versions: {qt_version}, {qt_api} {qt_api_ver} on {system_version}
 
 ## Dependencies
+
 ```
-%s
+{dependencies}
 ```
-""" % (traceback,
-       versions['spyder'],
-       revision,
-       versions['python'],
-       versions['qt'],
-       versions['qt_api'],
-       versions['qt_api_ver'],
-       versions['system'],
-       dependencies.status())
+""".format(description=description,
+           traceback=traceback,
+           spyder_version=versions['spyder'],
+           commit=revision,
+           python_version=versions['python'],
+           qt_version=versions['qt'],
+           qt_api=versions['qt_api'],
+           qt_api_ver=versions['qt_api_ver'],
+           system_version=versions['system'],
+           dependencies=dependencies.status())
 
         return issue_template
 
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -2381,7 +2381,7 @@ class MainWindow(QMainWindow):
         dlg.exec_()
 
     @Slot()
-    def report_issue(self, traceback="", description="1. \n2. \n3. ", expected=""):
+    def report_issue(self, traceback=""):
         if PY3:
             from urllib.parse import quote
         else:
@@ -2396,11 +2396,12 @@ class MainWindow(QMainWindow):
 
 **What steps will reproduce the problem?**
 
-%s
+1. 
+2. 
+3. 
 
 **What is the expected output? What do you see instead?**
 
-%s
 
 **Please provide any additional information below**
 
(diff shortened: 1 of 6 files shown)


goanpeca:
@rlaverde how about using a `placeholderText` ?

jitseniesen:
Is the second field useful? The answer to the question "What do you see instead" is always: an unhandled exception error box.

rlaverde:
> how about using a placeholderText ?

Instead of labels or for indicate that the text should be minimum 10 words length?

>Is the second field useful? The answer to the question "What do you see instead" is always: an unhandled exception error box.

Yes, should I remove the second filed? or only the second question?

goanpeca:
>how about using a placeholderText ?
>Instead of labels or for indicate that the text should be minimum 10 words length?

I don't think we should say it hs to be a minimum length, but we can say. `This field needs to include a description before you can send the report`

goanpeca:
But I insist again this (issue reporting automatically) is a terrible idea and does not scale well.

rlaverde:
> I don't think we should say it hs to be a minimum length, but we can say. This field needs to include a description before you can send the report

Yes I like that, and maybe a label that indicate that the text should be longer

>But I insist again this (issue reporting automatically) is a terrible idea and does not scale well.

I agree with you, but it's better to be flooded with issues with a description, that with empty issues 🙈 

Maybe the solution, is to improve spyder logging system #5326 and use a different service for automate issues reports

goanpeca:
>Maybe the solution, is to improve spyder logging system #5326 and use a different service for automate issues reports

Yes indeed that is the right solution

ccordoba12:
> should I remove the second filed? or only the second question?

I agree with @jitseniesen: we should remove that field.

----

@rlaverde, please add

    1.
    2.
    3.

to the first field. And a placeholder text suggested by @goanpeca (or maybe a label over the field would be better).

ccordoba12:
My suggestions to finish this one:

1. Remove the two panels and only leave one.
2. Use an instance of CodeEditor with Markdown and the same color scheme of the Editor to enter text.
3. Add a button to dismiss the dialog during an entire session.

rlaverde:
This is how it looks like:

spectacle e18905

spectacle q18925

goanpeca:
@rlaverde I suggest the following:

1. Get rid of the additional text editor with the 1 2 3
2. Change the Submit to github to `Create New Issue`
3. Create a copy to clipboard button so users can copy `the template` + `the system details` + `the error content`
4. The link that opens the create new github should erase the body message and add a generic title "Spyder Error Report". Its important to erase the body otherwise the template would allow them to report an issue with some prefilled empty content.!
5. If we got rid of the `Hide details` button the UI would be less crowded

`Create New Issue | Ok | Dismiss`

Prikers:
What is the difference between `OK` and `Dismiss` in this dialog widget?

ccordoba12:
Great point @Prikers! @rlaverde, I think we should remove `Ok` and only leave `Dismiss`.

`Dismiss` means *don't show this dialog during this session*. @Prikers, is that something that can be grasped from `Dismiss`?

Prikers:
@ccordoba12 yes that what I'd expect from a `Dismiss` button. I was indeed thinking that `Ok` could be removed as it does not bring a lot of value compared to `Dismiss`.

rlaverde:
Now it looks like this:

spectacle a15172

goanpeca:
@Prikers suggested we used Dismiss instead of Ok, (they are probably doing the same thing now?)

Thoughts @ccordoba12, @rlaverde ?

rlaverde:
I thought we were adding something to copy the content, a context menu, a left click on the message, a Copy to clipboard button ?

I change the message of the dialog, adding another button will be an extra click

- @Prikers suggested we used Dismiss instead of Ok, (they are probably doing the same thing now?)

I removed the `Ok` button, leaving the `Dimiss` button

goanpeca:
- @rlaverde all the text in Note should be italic, not just the Note.
- If the user accidentally replaces the clipboard with something, it should be possible to copy with one mouse click, so if there is no button, now the user has to select a text, and press a key or make 2 mouse clicks, which in case of user errors (very common) is still better to have an easy way of copying. The copy button does not need to be in line with the other buttons, it can be located within the editor frame.

rlaverde:
@goanpeca Doing what you suggest is not possible in a normal `QDialog`, I'll have to rewrite it to allow a button with a different position

goanpeca:
@rlaverde the embeded editor cannot be a widget that contains a button annd the editor?

rlaverde:
>the embeded editor cannot be a widget that contains a button and the editor?

That isn't an embebed editor, is the Show details functionality of the `QDialog`

goanpeca:
Hmm ok

ccordoba12:
@CAM-Gerlach, this is how the dialog looks now

seleccion_001

and how the report looks on Github

seleccion_002

After pasting the contents, you'll get this:

seleccion_003

----

Of course, this is to have an initial version on top of which you can add the rest of your improvements. What do you think?

ccordoba12:
New screenshot with the initial label text made more emphatic:

seleccion_001

ccordoba12:
Last version, with a monospace font and a better color scheme:

seleccion_001

ccordoba12:
I think this is ready! In my last commits I made the header on the description field read-only, so users are unable to remove it.

I also remove the listing (`1. 2. 3.`) because it doesn't work with a read-only header.

jitseniesen:
The button should read "Dismiss" :smile:

ccordoba12:
> The button should read "Dismiss"

Fixed. Thanks for noticing it!

ccordoba12:
This turned out to be really big, but I think it's finally ready!

PR #5367, opened 2017-09-30
PR: Add an option to the Editor to convert EOL characters on save

### LINKED ISSUE #5365: Add "Save all files with <eol> EOL characters" option
There should be an option on the editor to always save files with LF / CRLF / CR line endings.

Basically this would run `Source -> Convert end-of-line characters` on every save. There might be a more sophisticated or faster way to do it, but I'm not aware of one.

Option screen could look like:

image

I'll open a PR for this to see how you guys like it. Defaults will be the status-quo: EOLs are not changed on save.

(Edit1: replaced text representation of option screen with image)

dougthor42:
This PR adds an option to the editor to convert EoL characters on save.

It simply adds another check to the `widgets.editor.EditorStack.save()` method, right after the remove_trailing_whitespace action.

The configuration was added to the `Editor -> Advanced settings` tab in the preferences and looks like so:

image
image

**Note:** At the time of submitting this PR, no unit tests are included - I might need some help with that if it's something I need to add. (e.g.: where are tests located? What's the procedure for testing something like this?)

This fixes #5365.

### DIFF
--- a/spyder/utils/sourcecode.py
+++ b/spyder/utils/sourcecode.py
@@ -56,8 +56,11 @@ def get_os_name_from_eol_chars(eol_chars):
 
 
 def get_eol_chars_from_os_name(os_name):
-    """Return EOL characters from OS name"""
-    for eol_chars, name in EOL_CHARS:
+    """Return EOL characters from OS name.
+
+    `os_name` can also be 'CR', 'CRLF', or 'LF'"""
+    extra_eol_chars = ("\r\n", 'CRLF'), ("\n", 'LF'), ("\r", 'CR')
+    for eol_chars, name in EOL_CHARS + extra_eol_chars:
         if name == os_name:
             return eol_chars
 
--- a/spyder/utils/sourcecode.py
+++ b/spyder/utils/sourcecode.py
@@ -56,11 +56,8 @@ def get_os_name_from_eol_chars(eol_chars):
 
 
 def get_eol_chars_from_os_name(os_name):
-    """Return EOL characters from OS name.
-
-    `os_name` can also be 'CR', 'CRLF', or 'LF'"""
-    extra_eol_chars = ("\r\n", 'CRLF'), ("\n", 'LF'), ("\r", 'CR')
-    for eol_chars, name in EOL_CHARS + extra_eol_chars:
+    """Return EOL characters from OS name"""
+    for eol_chars, name in EOL_CHARS:
         if name == os_name:
             return eol_chars
 
--- a/spyder/widgets/editor.py
+++ b/spyder/widgets/editor.py
@@ -1596,10 +1596,7 @@ class EditorStack(QWidget):
         if self.always_remove_trailing_spaces:
             self.remove_trailing_spaces(index)
         if self.convert_eol_on_save:
-            # hack to account for the fact that the config file saves
-            # CR/LF/CRLF while set_os_eol_chars wants the os.name value.
-            osname_lookup = {'LF': 'posix', 'CRLF': 'nt', 'CR': 'mac'}
-            osname = osname_lookup[self.convert_eol_on_save_to]
+            osname = self.convert_eol_on_save_to
             self.set_os_eol_chars(osname=osname)
         txt = to_text_string(finfo.editor.get_text_with_eol())
         try:
@@ -2225,12 +2222,12 @@ class EditorStack(QWidget):
 
     def set_os_eol_chars(self, index=None, osname=None):
         """Sets the EOL character(s) based on the operating system.
-        
+
         If `osname` is None, then the default line endings for the current
         operating system (`os.name` value) will be used.
-        
+
         `osname` can be one of:
-            ('posix', 'nt', 'java')
+            ('posix', 'nt', 'java', 'CR', 'LF', 'CRFL')
         """
         if osname is None:
             osname = os.name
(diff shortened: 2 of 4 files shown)


ccordoba12:
@dougthor42, thanks a lot for your contribution! It's something we've been asked to implement this by several users, but we haven't had the time to do it.

spyder/widgets/editor.py
```diff
@@ -1583,6 +1591,12 @@ def save(self, index=None, force=False):
             return self.save_as(index=index)
         if self.always_remove_trailing_spaces:
             self.remove_trailing_spaces(index)
+        if self.convert_eol_on_save:
+            # hack to account for the fact that the config file saves
```
rlaverde: Maybe this logic could be moved to a function in `spyder/utils/sorcecode.py`, `EOL_CHARS` could be changed, to also include `CR/LF/CRLF`:

`EOL_CHARS = (("\r\n", 'nt', 'CRLF'), ("\n", 'posix', 'LR'), ("\r", 'mac', 'CR'))`

spyder/widgets/editor.py
```diff
@@ -1175,6 +1177,12 @@ def set_always_remove_trailing_spaces(self, state):
         # CONF.get(self.CONF_SECTION, 'always_remove_trailing_spaces')
         self.always_remove_trailing_spaces = state
 
+    def set_convert_eol_on_save(self, state):
```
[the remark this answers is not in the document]
dougthor42: Done

spyder/widgets/editor.py
```diff
@@ -2205,11 +2219,13 @@ def load(self, filename, set_current=True):
         self.is_analysis_done = False
         return finfo
 
-    def set_os_eol_chars(self, index=None):
+    def set_os_eol_chars(self, index=None, osname=None):
```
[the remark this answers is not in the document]
dougthor42: Done

dougthor42:
I've made the requested updates. Please let me know if there's anything else.

Glad to help 😄

ccordoba12:
@rlaverde, do you approve this one?

spyder/utils/sourcecode.py
```diff
@@ -56,8 +56,11 @@ def get_os_name_from_eol_chars(eol_chars):
 
 
 def get_eol_chars_from_os_name(os_name):
-    """Return EOL characters from OS name"""
-    for eol_chars, name in EOL_CHARS:
+    """Return EOL characters from OS name.
```
rlaverde: I think is a little confusing that It could also accept `CRLF`, `LF`, and `CR`,when they aren't os names

spyder/widgets/editor.py
```diff
@@ -2205,11 +2220,21 @@ def load(self, filename, set_current=True):
...
 
-    def set_os_eol_chars(self, index=None):
+    def set_os_eol_chars(self, index=None, osname=None):
+        """Sets the EOL character(s) based on the operating system.
+
+        If `osname` is None, then the default line endings for the current
+        operating system (`os.name` value) will be used.
+
+        `osname` can be one of:
+            ('posix', 'nt', 'java', 'CR', 'LF', 'CRFL')
```
rlaverde: Same here, 'CR', 'LF', 'CRFL' aren't os names

spyder/utils/sourcecode.py
dougthor42: So how would you like me to proceed? I'm assuming we'd prefer to not rename the function or the args - I'm sure it's used elsewhere and it might also be called with kwargs sometimes.

I could create a new function, but that would be almost a carbon copy of this one.

rlaverde: I'm not sure, maybe the best approach will be to rename it and add kwargs: `get_eol_chars(os_name=None, eol=None)` (but that will break compatibility as you said)

We could also consider adding a function `eol_to_os_name(eol):` that will convert `CRLF`, `LF`, and `CR` in windows linux or mac, but It'll be somehow silly.

@ccordoba12 What do you think?

ccordoba12: I don't think @dougthor42's work should be to fix this because it's already present. We should leave it for another PR.

ccordoba12: @rlaverde, other than this, are you good with this PR?

rlaverde: >I don't think @dougthor42's work should be to fix this because it's already present. We should leave it for another PR.

But  @dougthor42 changed this  function to also receive CRLF, LF, CR

> other than this, are you good with this PR?

Yes, this is my only complain with this PR, If It's isn't an issue this PR is ready to be merged 😄

dougthor42: If we don't like the change to `get_eol_chars_from_os_name`, we can simply go back to c9a002da and things should still work just fine. It'll mean that little hack at `spyder/widgets/editor.py` line 1601 would still exist, but I don't think that's a big deal.

ccordoba12: @dougthor42, ok, please revert to that commit so we can merge this one. This is the last PR before releasing 4.0 beta1!

dougthor42:
@rlaverde @ccordoba12 I've reverted 1a0c69a97c030d95c274b0ae87ac08134d22c7ef (which added the confusing changes to `get_eol_chars_from_os_name`).

ccordoba12:
Thanks @dougthor42!

@rlaverde, do you approve this one?

rlaverde:
>@rlaverde, do you approve this one?

Yes

PR #5417, opened 2017-10-07
PR: Fix PyQt4 compatibility issue with the QHeaderView class in findinfiles module

### LINKED ISSUE #5416: Cannot get past splash screen

1. When starting Spyder the splash screen will pop up and make variable amounts of progress then close but the application itself doesn't open.

2. If I try to open Spyder again I get the "Spyder crashed during last session." popup warning.

3. I then reset to defaults by running Spyder with the command line option 'spyder --reset

4. Thee splash screen image pops up for a sec disappears and then nothing happens

5. I try to open Spyder again and end up at step 1 again.

The expected outcome is for the main window to load instead I just get the splash screen to pop up for a minute. 

If i try running spyder again using the command prompt 'spyder --show-console'
I get the following error 
AttributeError: 'QHeaderView' object has no attribute 'setSectionsClickable'

I have tried restarting my desktop and I have tried uninstalling and reinstalling spydering using the anaconda prompt 'conda uinstall spyder' and 'conda install spyder' to no avail

The last time the main window was open it crashed with a kernel error in the console as soon as it opened

* Spyder Version: 3.2.3
* Python Version: 2.7.12
* Qt Version: 4.8.7
* PyQt Version: 4.11.4

## Dependencies

and paste the contents below:

jnsebgosselin:
I've realised while looking at the tests that this PR is not the good way to solve Issue #5416. The problem should be already patch in `qtpy` and it should not be addressed in spyder directly.

----

The [setClickable](http://doc.qt.io/qt-4.8/qheaderview.html#setClickable) method in `QHeaderView` has been replaced in Qt4 by [setSectionsClickable ](http://doc.qt.io/qt-/qheaderview.html#setSectionsClickable) in Qt5.

I added a `if PYQT5` condition in the `findinfiles` module that will uses `setSectionsClickable` if PyQt5 is used and `setClickable` otherwise.

```python
if PYQT5:
    self.header().setSectionsClickable(flag == ON)
else:
    self.header().setClickable(flag == ON)
```

### DIFF
--- a/spyder/widgets/findinfiles.py
+++ b/spyder/widgets/findinfiles.py
@@ -22,6 +22,7 @@ import math
 import traceback
 
 # Third party imports
+from qtpy import PYQT5
 from qtpy.compat import getexistingdirectory
 from qtpy.QtGui import QAbstractTextDocumentLayout, QTextDocument
 from qtpy.QtCore import QMutex, QMutexLocker, Qt, QThread, Signal, Slot, QSize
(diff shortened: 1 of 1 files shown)


jnsebgosselin:
Closing this PR since this should not be handled in spyder directly but in qtpy, see updated OP.

PR #5438, opened 2017-10-10
PR: Merge translations from former external plugins

rlaverde:
I used `msgcat` for merging translations, and manually resolve some conflicts with strings in different `po` files with different translations (I leave the` spyder.po` translation in all the cases)

I didn't regenerate the compile `.mo` files, neither run `gettext_scan`, that will add unnecessary noise to the PR (also those commands will be executed before the release)


rlaverde:
I think this was the last `split-plugins` PR, so only the  merge with master  is missing 😄

PR #5472, opened 2017-10-15
PR: Don't rely on process name when detecting a previous instance

### LINKED ISSUE #4186: Spyder always opens a new instance

Spyder opens new instance without requesting one. This is a reminder. Details have been previously reported.

1. From the command line, typing 'spyder filename.py' always opens a new instance of spyder, even when the opposite behaviour (i.e., 'use a single instance) is selected in the preferences. 

2. 
3. 

* Spyder Version: 3.1.3
* Python Version:  Python 3.5.2 :: Anaconda custom (x86_64)
* Qt Version:   5.6.2  
* PyQt Version: PyQT5 5.6 on Darwin
* Operating system: Mac OS 10.12.3

avylove:
Fixes #4186 

----

Instead of relying on the process name, up to the first three command line arguments are reduced to their basenames and compared against 'spyder', 'spyder3', and 'bootstrap.py'.

This should mimic the existing behavior and address cases where the process name does not show spyder.

When submitting this pull request, I found another option, using setproctitle. Will mention in #4186.

### DIFF
--- a/spyder/utils/external/lockfile.py
+++ b/spyder/utils/external/lockfile.py
@@ -182,10 +182,12 @@ class FilesystemLock:
                             conditions = ['spyder' in c.lower()
                                           for c in p.cmdline()]
                         else:
-                            conditions = [p.name() == 'spyder',
-                                          p.name() == 'spyder3']
-                        # For DEV
-                        conditions += ['bootstrap.py' in p.cmdline()]
+                            # Valid names for main script
+                            names = set(['spyder', 'spyder3', 'bootstrap.py'])
+                            # Check the first three command line arguments
+                            arguments = set(os.path.basename(arg)
+                                            for arg in p.cmdline()[:3])
+                            conditions = [names & arguments]
                         if not any(conditions):
                             raise(OSError(errno.ESRCH, 'No such process'))
                     except OSError as e:


spyder/utils/external/lockfile.py
```diff
@@ -182,10 +182,12 @@ def lock(self):
...
-                            conditions = [p.name() == 'spyder',
-                                          p.name() == 'spyder3']
-                        # For DEV
-                        conditions += ['bootstrap.py' in p.cmdline()]
+                            # Valid names for main script
+                            names = set(['spyder', 'spyder3', 'bootstrap.py'])
+                            # Check the first three command line arguments
+                            arguments = set(os.path.basename(arg)
+                                            for arg in p.cmdline()[:3])
+                            conditions = [names & arguments]
```
ccordoba12: What does this line do?

ccordoba12:
@avylove, thanks a lot for helping us to solve this problem!

spyder/utils/external/lockfile.py
avylove: First gets the basename for each of the command line arguments, up to the first 3. For example, ```['/Users/gkv/anaconda/python.app/Contents/MacOS/python', '/Users/gkv/anaconda/bin/spyder']``` would become ```set(['python', 'spyder'])```, then it does an intersection of the sets so only the values in both sets remain.

jitseniesen: So why do you put the intersection in a list and then use `if not any`? 

How about `if not (names & arguments)` ?

avylove: I kept the existing convention. Otherwise It would require changing the whole if block. From an efficiency standpoint, there is little to no gain.

Issue #5475, opened 2017-10-15
'Show in external file explorer' on editor menu more than once for split panels


csabella:
When in the editor, every time the panel is split, the gears menu adds an additional option for 'Show in external file explorer'.

**What steps will reproduce the problem?**

1.  Open the editor.
2.  Click on the gears and see 'Show in external file explorer' is listed once.
3.  Split the panel.
4.  Check the gears on the new panel.  'Show in external file explorer' is now listed twice.
5.  Rinse and repeat.  It's added one more time for each split.

**What is the expected output? What do you see instead?**
I expect the menu option to only be listed once.

**Please provide any additional information below**
I believe this was introduced with PR #4337.  In the EditorStack class in widgets editor.py, the incoming argument of actions is appened to(`actions.append(external_fileexp_action)`).  Since this is a list, the append causes it to be changed in the calling routine (EditorSplitter class).  When a new EditorStack is created, the new list with the appended action is sent instead of the original list being sent.

## Versions and main components

* Spyder Version:  3.x and 4.0
* Python Version:  3.6.3
* Qt Version:
* PyQt Version:
* Operating system:  Ubuntu 16.04

## Dependencies

Please go to the menu entry `Help > Optional Dependencies` (or
`Help > Dependencies`), press the button `Copy to clipboard`
and paste the contents below:

PR #5512, opened 2017-10-19
PR: Add shortcuts for splitting and closing panels

### LINKED ISSUE #2419: Shortcuts for splitting and closing panels
Another kind of spoiled feature request from a huge Spyder fan:  it would be nice to have shortcuts to split the panel in the editor instead of having to hit the `Options` button, then `Split horizontally` each time.

And also a shortcut for closing the panel.

csabella:
Fixes #2419.

1.  Adds shortcut `Ctrl-{` for vertical split and `Ctrl-_` for horizontal split based on similar shortcuts in the Eclipse editor.  Adds shortcut `Alt-Shift-W` to close the current split panel.  This was an arbitrary choice based on the other `Close` shortcuts.
2.  Added some docstrings to the EditorSplitter class.
3.  Added more tests to test_editor.py for the EditorSplitter class.


ccordoba12:
@csabella, thanks a lot for you continued contributions!

spyder/widgets/editor.py
```diff
@@ -2454,18 +2510,44 @@ def split(self, orientation=Qt.Vertical):
...
+                           cfname: EditorStack current file name.
+                           clines: Current line number for each file in the
+                               EditorStack.
+        """
         splitsettings = []
         for editorstack, orientation in self.iter_editorstacks():
             clines = []
             cfname = ''
+            # XXX - this overrides value from the loop to always be False?
             orientation = False
```
dalthviz: I think you are right with the value for `orientation`, but I'm not quite sure. @ccordoba12 do you know the reason to do this? After clarifing this, please @csabella, write the explanation instead of the current comment that you left.

spyder/widgets/tests/test_editor.py
```diff
@@ -106,11 +107,35 @@ def editor_folding_bot(base_editor_bot):
...
+    es = editor_splitter_bot
+
+    # Allow the split() to duplicate editor stacks.
+    def clone(editorstack):
+        editorstack.close_action.setEnabled(False)
+        editorstack.set_introspector(Mock())
+        editorstack.set_find_widget(Mock())
+        editorstack.set_io_actions(Mock(), Mock(), Mock(), Mock())
+        editorstack.new('foo.py', 'utf-8', 'a = 1\nprint(a)\n\nx = 2')
+        editorstack.new('spam.py', 'utf-8', 'print(spam)')
```
dalthviz: Please change `spam` for another word, I'm not quite sure but maybe this is related with the find in files test failing.

spyder/widgets/editor.py
csabella: @dalthviz   This was changed recently by andfoy and I put this comment so that it could be researched further.  What I found so far is that the `orientation` boolean from here is not used in any of the loops that read `splitsettings` (in plugins/editor or widgets/editor) except for in `set_layout_settings` in `EditorSplitter` which has the following line (it is mapped into `is_vertical` in the loop):

    splitter.split(Qt.Vertical if is_vertical else Qt.Horizontal)

but somehow the orientation of the splits end up correct on the restore.  I believe that `hexstate` within `set_layout_settings` is applying the correct split orientation after the fact, so `orientation` may not even need to be saved at all in `splitsettings`.  Although, after Travis issues, I'm not sure what PyQt4 is doing, but it's definitely a different hexstate than PyQt5. 

Interestingly, in `set_layout_settings`, even `cfname` isn't used.  There is an inline comment about `go_to_line` and I had quite a bit of difficulty writing a test for the line search within `set_layout_settings` because it just seemed to not fit in with the rest of the code.

spyder/widgets/tests/test_editor.py
```diff
@@ -458,29 +483,219 @@ def test_get_current_word(base_editor_bot):
...
-
-    assert isinstance(es.editorstack, EditorStack)
-    es.plugin.register_editorstack.assert_called_with(es.editorstack)
-    es.plugin.unregister_editorstack.assert_not_called()
-    es.plugin.clone_editorstack.assert_not_called()
-
-    assert es.count() == 1
-    assert es.widget(0) == es.editorstack
+class TestEditorSplitter:
+    """Tests for EditorSplitter methods."""
```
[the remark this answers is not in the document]
csabella: @dalthviz   Since editor.py contains many classes, I was trying to write unit tests that tested just the function of one class and not so much the interaction between the classes or the GUI parts overall.  I wanted to specify that these tests were over methods of the EditorSplitter class.  Originally, I didn't have the class and I was writing test names such as `test_editorsplitter_split()`, but I think putting them into a Test Class made it a little less verbose and a little clearer.  For example, `test_init` and `test_close` could apply to any of the classes in the module, so this organizes them as tests just for EditorSplitter.

The other thought I had about using the Test Class was that, if editor.py is refactored where each class becomes its own module, then it would be easy to find the corresponding tests and refactor them into their own source as well.

If you prefer that there aren't classes within the test modules, I can change it.  

FYI, the test_init that I was writing for this PR was how I discovered issue #5475.  These methods alter the state of many properties, so that's what I was trying to add tests for.  If a certain state is expected now, then that's what should be expected after any code change is made.

spyder/widgets/tests/test_editor.py
csabella: Done.

spyder/widgets/tests/test_editor.py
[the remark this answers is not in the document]
rlaverde: >The other thought I had about using the Test Class was that, if editor.py is refactored where each class becomes its own module, then it would be easy to find the corresponding tests and refactor them into their own source as well.

Yes, moving each class to a module is the idea after split-plugins merge, and test will be in different files too

Maybe you could move this test to a new file instead spyder/widgets/tests/test_editorsplitter.py

spyder/widgets/editor.py
ccordoba12: @andfoy, please chime in this discussion. I haven't touched the splitting code, so I can't say anything constructive here, sorry.

spyder/widgets/tests/test_editor.py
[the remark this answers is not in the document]
csabella: @rlaverde   I've split the tests into a separate file.

spyder/widgets/editor.py
andfoy: @ccordoba12 @csabella To be honest, last time that I modified the splitting behaviour, I was not sure about some of the settings and definitions that were disposed there.

spyder/widgets/editor.py
```diff
@@ -2441,6 +2487,16 @@ def editorsplitter_closed(self):
         self.__give_focus_to_remaining_editor()
 
     def split(self, orientation=Qt.Vertical):
+        """Create and attach a new EditorSplitter to the current EditorSplitter.
+
+        The new EditorSplitter widget will contain an EditorStack that
+        is a clone of the current EditorStack.
+
+        A single EditorSplitter instance can be split multiple times, but the
+        orientatation will be the same for all the direct splits.  If one of
```
rlaverde: *orientation

spyder/widgets/editor.py
rlaverde: It was changed in https://github.com/spyder-ide/spyder/pull/4869, but personally I didn't find a good reason for setting it to `False`, although if it's working is better not to change it (and debbuging it is out of the scope of this PR)

ccordoba12: Ok, thanks @rlaverde. Let's merge this one as it is.

We'll see later how to improve this logic.

PR #5537, opened 2017-10-22
PR: Add clear button to find comboboxes

### LINKED ISSUE #5373: Add Clear button to (find) entry boxes

Recently I searched for a lot of strings – one after the other. They all came from an electronic document, so I could use copy'n'paste. Since I work on Linux, I just selected the search strings by double clicking it and wanted to paste them with a middle click into the find box. But the box was still populated with the search string from the previous search. This string had to be deleted first. Selecting it and middle click obviously does not work, so had to click into the entry box, switch to the keyboard (CTRL-A, DEL) and paste again with the mouse.

It would have been so much quicker if there were a "Clear search box" button as you can now see nearly everywhere.

Could you please provide such a button in every search box (I can come up with four: The _Editor_ has Find, Find + Replace, the _Find in files_ has one, _Help_ another, but there might be more.) Maybe other entry boxes can also gain from a Clear button?

* Spyder Version: 3.2.3 
* Python Version: 3.5.2 64bits
* Qt Version: 5.5.1
* PyQt Version: 5.5.1
* Operating system: Linux Mint 18.1

## Dependencies

and paste the contents below:
psutil >=0.3     :  5.3.1 (OK)
sympy >=0.7.3    :  1.0 (OK)

timhoffm:
Fixes #5373

### DIFF
--- a/spyder/widgets/comboboxes.py
+++ b/spyder/widgets/comboboxes.py
@@ -135,6 +135,8 @@ class PatternComboBox(BaseComboBox):
     def __init__(self, parent, items=None, tip=None,
                  adjust_to_minimum=True):
         BaseComboBox.__init__(self, parent)
+        if hasattr(self.lineEdit(), 'setClearButtonEnabled'):  # only Qt >= 5.2
+            self.lineEdit().setClearButtonEnabled(True)
         if adjust_to_minimum:
             self.setSizeAdjustPolicy(QComboBox.AdjustToMinimumContentsLength)
         self.setSizePolicy(QSizePolicy.Expanding, QSizePolicy.Fixed)


spyder/widgets/comboboxes.py
```diff
@@ -135,6 +135,8 @@ class PatternComboBox(BaseComboBox):
     def __init__(self, parent, items=None, tip=None,
                  adjust_to_minimum=True):
         BaseComboBox.__init__(self, parent)
+        if hasattr(self.lineEdit(), 'setClearButtonEnabled'):  # only Qt >= 5.2
```
ccordoba12: Given that this only works for `Qt > = 5.2`, you need to enclose the lines you added in

```python
if PYQT5:
    ...
```

where `PYQT5` can be imported like this


```python
from qtpy import PYQT5
```

timhoffm: The `if hasattr()` serves exactly that purpose: Only use this feature if available. This should work without explicity checking a Qt version.

Also, I would assume that a simple `if PYQT5` would be `True` for  5.0 <= Qt < 5.2, but they still don't have the clear button.

ccordoba12: Sorry, you're right, `hasattr` should be enough.

Besides, we only support `Qt 5.2+`

ccordoba12:
@dalthviz, please review that this is working locally for you.

spyder/widgets/comboboxes.py
timhoffm: > Besides, we only support Qt 5.2+

In that case, just `self.lineEdit().setClearButtonEnabled(True)` would be enough.

The `if` is unneccesary clutter then. Should I remove it for brevity and clarity of the code?

ccordoba12: Nop, because we also support PyQt4.

dalthviz:
@ccordoba12  I tested locally and worked for me :+1: (tested in PyQt4 and PyQt5)

ccordoba12:
Thanks @dalthviz!

PR #5538, opened 2017-10-22
PR: Reduce row height in dependencies dialog

timhoffm:
By default, Qt uses an unnecessarily large row height for tables. This results in not many rows being visible in the dependency dialog.

This patch reduces the row height of the dependency table.

Before:
grafik

After:
grafik


ccordoba12:
Thanks for your contribution!

Issue #5558, opened 2017-10-24
Setting "Clear Line" Keyboard Shortcut Does Not Work Properly


CAM-Gerlach:
## Description

The default KB shortcut for clear line in the console requires some gymnastics (Shift-Esc) and takes far longer than my typical Ctrl-A, Backspace, which does not work for whatever reason (as Ctrl-A/Select-All does not select the current line in the console, as it presumably should). Thus, I went to set it, but seemingly regardless of what I set it to, there was some problem or another. 

Setting it to Ctrl-D made it behave like the delete key, deleting one character at a time in front of the cursor. Setting it to Shift-Ctrl-Backspace made it delete one word at a time, like Ctrl-Backspace, and Shift-Ctrl-Delete did the same thing, except forward-delete like Ctrl-Del. Setting Alt-D selected debug in the Debug menu, like Alt-keys otherwise do. Setting Ctrl-Alt-Shift F made the cursor advance one character forward at a time without deleting, like the right arrow. Setting Ctrl-Shift-A made it select all the text in the console window (pity Ctrl-A didn't do anything, i.e. select the current line like in most other consoles, or I'd be in business without having to worry about this shortcut). Finally, setting an arbitrary shortcut (Ctrl-Shift-Alt-J) made it just type the letter J. However, Shift-Esc, as well as just Esc, still cleared the line even though it wasn't set. 

To make an overly long story short, these are all different shortcuts, which function regardless of how this shortcut it set, which leads to the obvious conclusion that changing the setting isn't actually doing anything. Perhaps it it might have something to do with it being specifically for the console, as it does not conflict (in the KB shortcuts dialog) with other non-console shortcuts, as it should, but neither does it actually do anything to change it.

**What steps will reproduce the problem?**

1. Set "Clear Line" shortcut to something other than the default
2. Attempt to use new shortcut
3. Fail

**What is the expected output? What do you see instead?**

Expected: Set shortcut (ideally, personally I'd prefer Ctrl-D to mirror the editor) clears current line in the console. Also, see no reason why Ctrl-A shouldn't select the current line, as it does in most other (decent) consoles, which then enables easy copying, pasting, deletion, etc.

Actual: Selected shortcut does nothing, or performs some other function, while previous shortcut does not work. Ctrl-A does nothing, while Ctrl-Shift-A selects all console text (not just current line).

**Additional Information**

[...]

PR #5560, opened 2017-10-24
PR: Make clear line shortcut configurable in the IPython console

rlaverde:
Fixes: #5558

I'm not sure if it's the best solution because the `Shift+Esc` shortcut isn't replaced, instead a new shortcut is added for the same action


ccordoba12:
This is a temporary solution, but it's fine. What we need to do for the future is to make all `qtconsole` shortcuts configurable.

Do you want to work on that one?

rlaverde:
>What we need to do for the future is to make all qtconsole shortcuts configurable.
Do you want to work on that one?

Hmm yes, but those changes should be done in qtconsole? something like exposing a function for reconfigurate shortcuts?

ccordoba12:
> those changes should be done in qtconsole?

Yep, they need to be done in `qtconsole` itself, perhaps by exposing a lot of class attributes that can be redefined by us when sub-classing `RichJupyterWidget`.

Issue #5578, opened 2017-10-26
Can't open preferences anymore as I get an error message


JakobJakobson:
## Description
After I resetted Spyder3 and set it to default setting, I can't open the preferences in
Tools -> Preferences
anymore as I shows me an error message with the option to file a bug report on GitHub.

**What steps will reproduce the problem?**

1. Take a Fedora 26 Workstation installation.
2. Customize it a bit.
3. Run spyder3 --reset in terminal.
4. Run spyder3 --default in terminal.
5. Start spyder3 and try to open the preferences.

**What is the expected output? What do you see instead?**
I get an error window with the option to file the error on GitHub. The automatically collected data is shown below.
Instead I would expect to get into the preferences menue.

**Please provide any additional information below**
As a side comment: Apparently it does only happen if you apply the `spyder3 --defaults` command IMMEDIATELY after the `spyder3 --reset` command. Using just one of them gives no error message when you try to open the preferences.

```python
  File "/usr/lib/python3.6/site-packages/spyder/app/mainwindow.py", line 2602, in edit_preferences
    widget.initialize()
  File "/usr/lib/python3.6/site-packages/spyder/plugins/configdialog.py", line 76, in initialize
    self.load_from_conf()
  File "/usr/lib/python3.6/site-packages/spyder/plugins/configdialog.py", line 380, in load_from_conf
    color, bold, italic = self.get_option(option, default)
TypeError: 'NoneType' object is not iterable
```

## Version and main components

* Spyder Version: 3.2.2 
* Python Version: 3.6.2
* Qt Versions: 4.8.7, PyQt4 (API v2) 4.12 on Linux

## Dependencies
```
pyflakes >=0.6.0 :  1.5.0 (OK)
pycodestyle >=2.3:  2.3.1 (OK)
pygments >=2.0   :  2.2.0 (OK)
pandas >=0.13.1  :  0.20.1 (OK)
numpy >=1.7      :  1.12.1 (OK)
sphinx >=0.6.6   :  1.5.2 (OK)
rope >=0.9.4     :  None (NOK)
jedi >=0.9.0     :  0.10.2 (OK)
psutil >=0.3     :  5.0.1 (OK)
nbconvert >=4.0  :  4.2.0 (OK)
sympy >=0.7.3    :  1.1.1 (OK)
cython >=0.21    :  None (NOK)
qtconsole >=4.2.0:  4.2.1 (OK)
IPython >=4.0    :  5.3.0 (OK)
[...]

PR #5584, opened 2017-10-26
PR: Fix typo in Solarized color schemes.

rlaverde:
Fixes: #5578


rlaverde:
>However, I think this needs a major bump in CONF_VERSION for changes to take effect in older installations.

Yes, I agree, renaming is like deleting old configs, and adding new ones

PR #5585, opened 2017-10-27
PR: Fix ValueError when switching Python interpreters

### LINKED ISSUE #5402: ValueError: invalid literal for int() with base 10: b''
## Description

1. 
2. 
3. 

  File "/anaconda3/lib/python3.6/site-packages/spyder/plugins/maininterpreter.py", line 162, in python_executable_switched
    self.warn_python_compatibility(cust_pyexec)
  File "/anaconda3/lib/python3.6/site-packages/spyder/plugins/maininterpreter.py", line 171, in warn_python_compatibility
    console_version = int(proc.communicate()[0])
ValueError: invalid literal for int() with base 10: b''

* Spyder Version: 3.2.3 
* Python Version: 3.6.2
* Qt Versions: 5.6.2, PyQt5 5.6 on Darwin

## Dependencies
```

```

csabella:
Fixes #5402.

In issue #5402, the call to sys.version_info was returning `b''` instead of `b'3'`.  The ValueError was a result of calling int() on an empty bytes value.  The change checks for this exception and treats it the same as the call to `programs.is_python_interpreter()`.

I don't know why the subprocess call with the program name and -h option would work in `is_python_interpreter()`, but the -c option call from `warn_python_compatibility` wouldn't returning a valid value.  The assumption made was that if the -c option doesn't work, then something else might be wrong with the interpreter, so it switches back to the default.

### DIFF
--- a/spyder/plugins/maininterpreter.py
+++ b/spyder/plugins/maininterpreter.py
@@ -139,26 +139,24 @@ class MainInterpreterConfigPage(GeneralConfigPage):
         """Custom Python executable value has been changed"""
         if not self.cus_exec_radio.isChecked():
             return
+        def_pyexec = get_python_executable()
         if not is_text_string(pyexec):
             pyexec = to_text_string(pyexec.toUtf8(), 'utf-8')
+        if pyexec == def_pyexec:
+            return
         if (not programs.is_python_interpreter(pyexec) or
             not self.warn_python_compatibility(pyexec)):
             QMessageBox.warning(self, _('Warning'),
                     _("You selected an invalid Python interpreter for the "
                       "console so the previous interpreter will stay. Please "
                       "make sure to select a valid one."), QMessageBox.Ok)
-            self.pyexec_edit.setText(get_python_executable())
+            self.pyexec_edit.setText(def_pyexec)
             return
 
     def python_executable_switched(self, custom):
         """Python executable default/custom radio button has been toggled"""
-        def_pyexec = get_python_executable()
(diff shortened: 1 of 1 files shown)


spyder/plugins/maininterpreter.py
```diff
@@ -158,8 +157,8 @@ def python_executable_switched(self, custom):
         if not is_text_string(cust_pyexec):
             cust_pyexec = to_text_string(cust_pyexec.toUtf8(), 'utf-8')
         if def_pyexec != cust_pyexec:
-            if custom:
-                self.warn_python_compatibility(cust_pyexec)
+            if custom and not self.warn_python_compatibility(cust_pyexec):
+                self.pyexec_edit.setText(def_pyexec)
```
ccordoba12: I think we should show the same warning as above before doing this, i.e.

    You selected an invalid Python interpreter for the console so the previous ...

ccordoba12:
Thanks @csabella! This looks good to me, except for my minor comment.

spyder/plugins/maininterpreter.py
csabella: @ccordoba12   

`python_executable_changed` and `python_executable_switched` had very similar code, especially with the addition of this message.  Instead of duplicating the message, I instead changed it to call `python_executable_changed` from `python_executable_switched`.  I hope that's OK.

ccordoba12:
@csabella, please fix the pep8 issue here.

Issue #5386, opened 2017-10-03
Replace "\" by "\\"


badiay:
## Description

**What steps will reproduce the problem?**

1.  Tried to replace all "\" by "\\"
2.  Crash 
3. my code : 
" 
import pandas as pd 

a = pd.read_csv("C:\\Users\badia\Desktop\PSB\s1\Python\Exercices\data.csv")
"

**What is the expected output? What do you see instead?**

**Please provide any additional information below**

  File "C:\Users\badia\Anaconda3\lib\site-packages\spyder\widgets\findreplace.py", line 504, in replace_find_all
    self.replace_find(focus_replace_text, replace_all=True)
  File "C:\Users\badia\Anaconda3\lib\site-packages\spyder\widgets\findreplace.py", line 449, in replace_find
    if len(re.findall(pattern, cmptxt2)) > 0:
  File "C:\Users\badia\Anaconda3\lib\re.py", line 222, in findall
    return _compile(pattern, flags).findall(string)
  File "C:\Users\badia\Anaconda3\lib\re.py", line 301, in _compile
    p = sre_compile.compile(pattern, flags)
  File "C:\Users\badia\Anaconda3\lib\sre_compile.py", line 562, in compile
    p = sre_parse.parse(p, flags)
  File "C:\Users\badia\Anaconda3\lib\sre_parse.py", line 847, in parse
    source = Tokenizer(str)
  File "C:\Users\badia\Anaconda3\lib\sre_parse.py", line 231, in __init__
    self.__next()
  File "C:\Users\badia\Anaconda3\lib\sre_parse.py", line 245, in __next
    self.string, len(self.string) - 1) from None
sre_constants.error: bad escape (end of pattern) at position 0

## Version and main components

* Spyder Version: 3.2.3 
* Python Version: 3.6.2
* Qt Versions: 5.6.2, PyQt5 5.6 on Win

ccordoba12:
Thanks for reporting. We'll fix this problem is a future release.

timhoffm:
Note: This happens if the user has activated regular expressions. `\` is of course not a valid expression to search for.

One part of the problem is that the user was probably not aware of this fact. This is addressed by the above PR [#5614].

[...]

PR #5614, opened 2017-10-30
PR: Better feedback for regexp errors in find/replace

timhoffm:
Based on [#5386] in which the user did not realize that he was searching with an invalid regular expression.

This patch gives feedback on invalid regular expressions by coloring the search field and adding a tooltip.

### DIFF
--- a/spyder/widgets/findreplace.py
+++ b/spyder/widgets/findreplace.py
@@ -390,7 +390,7 @@ class FindReplace(QWidget):
         # When several lines are selected in the editor and replace box is activated, 
         # dynamic search is deactivated to prevent changing the selection. Otherwise
         # we show matching items.
-        def rexexp_error_msg(pattern):
+        def regexp_error_msg(pattern):
             """Returns None if the pattern is a valid regular expression or
             a string describing why the pattern is invalid.
             """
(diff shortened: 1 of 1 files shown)


goanpeca:
@timhoffm thanks a lot for this. could you paste a couple of screenshots to better visualize the changes made by this PR?

ccordoba12:

timhoffm:
Here's a screenshot illustrating the changes for invalid regular expressions.

Top: "no matches" in a normal search
Bottom: invalid regular expression with tooltip.

grafik

spyder/widgets/findreplace.py
```diff
@@ -400,7 +417,18 @@ def find(self, changed=True, forward=True,
...
+                stylesheet = self.STYLE['regexp_error']
+                tooltip = self.TOOLTIP['regexp_error']
+                error_msg = rexexp_error_msg(text)
+                if error_msg:
+                    tooltip += ': ' + error_msg
+            else:
+                stylesheet = self.STYLE[found]
+                tooltip = self.TOOLTIP[found]
+            self.search_text.lineEdit().setStyleSheet(stylesheet)
+            self.search_text.setToolTip(tooltip)
```
rlaverde: This code could be refactored, It has repeated calls to `rexexp_error_msg`, also both "if" are checking for the error message:
```
            stylesheet = self.STYLE[found]
            tooltip = self.TOOLTIP[found]
            if not found and regexp:
                error_msg = rexexp_error_msg(text)    
                if error_msg:
                    stylesheet = self.STYLE['regexp_error']
                    tooltip = "{}: {}".format(self.TOOLTIP['regexp_error'], error_msg)

            self.search_text.lineEdit().setStyleSheet(stylesheet)
            self.search_text.setToolTip(tooltip)
```

timhoffm: Thanks that's a more readable version (Note that calling `rexexp_error_msg()` twice is not a big deal in itself since `re.compile()` is cached).

I've applied your suggestion with the exception of not switching the string formatting. IMHO `"{}: {}".format(` is not making it more readable in this case (And also I don't want pep8 to moan again about the line length).

spyder/widgets/findreplace.py
```diff
@@ -50,7 +50,14 @@ class FindReplace(QWidget):
     """Find widget"""
     STYLE = {False: "background-color:rgb(255, 175, 90);",
              True: "",
-             None: ""}
+             None: "",
+             'regexp_error': "background-color:rgb(255, 20, 20);",
```
ccordoba12: Could we add some alpha here to tone down a bit the this red? I mean, as it is, it's a bit too strong for the eye (in my opinion).

ccordoba12:
@timhoffm, I left a minor comment. The rest looks good to me.

spyder/widgets/findreplace.py
timhoffm: Feel free do choose any shade of red you want. I didn't put any effort in chosing the color. IMO it should be a little stronger than the color for "not found".

There's no need for alpha though because it's the background anyway. Just adapt the rgb values.

PR #5618, opened 2017-10-31
PR: Catch search regexp error while doing replace/find

### LINKED ISSUE #5386: Replace "\" by "\\"

timhoffm:
fixes #5386

Perfroming `replace_find()` and `replace_find_selection()` with an invalid regexp used to raise an exception. After this patch, they will do nothing and silently return.

### DIFF
--- a/spyder/widgets/findreplace.py
+++ b/spyder/widgets/findreplace.py
@@ -426,7 +426,12 @@ class FindReplace(QWidget):
         if (self.editor is not None):
             replace_text = to_text_string(self.replace_text.currentText())
             search_text = to_text_string(self.search_text.currentText())
-            pattern = search_text if self.re_button.isChecked() else None
+            re_pattern = None
+            if self.re_button.isChecked():
+                try:
+                    re_pattern = re.compile(search_text)
+                except re.error:
+                    return  # do nothing with an invalid regexp
             case = self.case_button.isChecked()
             first = True
             cursor = None
@@ -436,7 +441,7 @@ class FindReplace(QWidget):
                     seltxt = to_text_string(self.editor.get_selected_text())
                     cmptxt1 = search_text if case else search_text.lower()
                     cmptxt2 = seltxt if case else seltxt.lower()
-                    if not pattern:
+                    if re_pattern is None:
                         has_selected = self.editor.has_selected_text()
                         if has_selected and cmptxt1 == cmptxt2:
                             # Text was already found, do nothing
@@ -446,7 +451,7 @@ class FindReplace(QWidget):
                                              rehighlight=False):
                                 break
                     else:
-                        if len(re.findall(pattern, cmptxt2)) > 0:
+                        if len(re_pattern.findall(cmptxt2)) > 0:
                             pass
                         else:
                             if not self.find(changed=False, forward=True,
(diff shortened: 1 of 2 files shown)


ccordoba12:
Nice! Thanks a lot for this!

@dalthviz, please review @timhoffm's work.

rlaverde:
Maybe you could generalize the function `rexexp_error_msg`  of #5614 and use in this PR too, something like:
```
def safe_regex(pattern):
    ....
    return compile_pattern, error 
```

ccordoba12:
@timhoffm, don't worry about PEP8 issues in our test files. You can safely ignore `pep8speaks` in that case.

timhoffm:
@ccordoba12 If you don't want `pep8speaks` on test files, you could add 

    pycodestyle:
        exclude: test_*.py

to `.pep8speaks.yml`

dalthviz:
About what @rlaverde says (in terms of the generalization of the `rexexp_error_msg` of #5614) it could be a possibility. Maybe we could merge this one first an decide what to do regarding the `rexexp_error_msg` in #5614 

What do you guys think @rlaverde, @ccordoba12, @timhoffm?

ccordoba12:
I agree, let's merge this one first. I think there's an error with Pandas that's making our tests to fail, so I'll take a look at that first before merging more work.

timhoffm:
@rlaverde, @ccordoba12, @dalthviz 

I would do one step at a time. Both this PR and #5614 are reasonable solutions in themselves and can be merged as they are. In a second step we can then see if a generalization of `regexp_error_msg` (note: typo is fixed) makes sense. I'm not yet sure about this. In this request we don't need the message and in #5614 we don't need the pattern.

ccordoba12:
> In this request we don't need the message and in #5614 we don't need the pattern.

@rlaverde?

rlaverde:
>In this request we don't need the message and in #5614 we don't need the pattern

yes, that's true, but  in both the try..except  is repeated
 
```
try:
    re_pattern = re.compile(search_text)
except re.error:
    return
```

ccordoba12:
Let's leave a further refactoring for another PR.

Issue #3612, opened 2016-10-28
Variable explorer should list datetime.timedelta variables


jondo:
Define a variable via

```
import datetime
x = datetime.timedelta()
```

Now  the variable should be listed in the variable explorer,
similar to `x = datetime.datetime(2016, 10, 28)`.

(My current environment is described in #3562.)

PR #5619, opened 2017-10-31
PR: Support datetime.timedelta in variable explorer

timhoffm:
implements #3612

### DIFF
--- a/spyder/widgets/variableexplorer/utils.py
+++ b/spyder/widgets/variableexplorer/utils.py
@@ -175,7 +175,27 @@ def datestr_to_datetime(value):
 
 
 def str_to_timedelta(value):
-    m = re.match('^(?:datetime\.timedelta)?\(?([^\)]*)\)?$', value)
+    """Convert a string to a datetime.timedelta value.
+
+    The following strings are accepted:
+
+        - 'datetime.timedelta(1, 5, 12345)'
+        - 'timedelta(1, 5, 12345)'
+        - '(1, 5, 12345)'
+        - '1, 5, 12345'
+        - '1'
+
+    if there are less then three parameters, the missing parameters are
+    assumed to be 0. Variations in the spacing of the parameters are allowed.
+
+    Raises:
+        ValueError for strings not matching the above criterion.
+
+    """
+    m = re.match('^(?:(?:datetime\.)?timedelta)?'
+                 '\(?'
+                 '([^)]*)'
+                 '\)?$', value)
     if not m:
         raise ValueError('Invalid string for datetime.timedelta')
     args = [int(a.strip()) for a in m.group(1).split(',')]
--- a/spyder/widgets/variableexplorer/utils.py
+++ b/spyder/widgets/variableexplorer/utils.py
@@ -173,6 +173,7 @@ def datestr_to_datetime(value):
     print(value, "-->", v)  # spyder: test-skip
     return v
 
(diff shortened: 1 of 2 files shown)


spyder/widgets/variableexplorer/utils.py
```diff
@@ -174,6 +174,14 @@ def datestr_to_datetime(value):
     return v
 
 
+def str_to_timedelta(value):
+    m = re.match('^(?:datetime\.timedelta)?\(?([^\)]*)\)?$', value)
+    if not m:
+        raise ValueError('Invalid string for datetime.timedelta')
```
ccordoba12: I don't think we should return an Error because Spyder can freeze because of it.

I think it's better to return the original value. That way people would understand they did something wrong when writing a new timedelta.

spyder/widgets/variableexplorer/utils.py
```diff
@@ -174,6 +174,14 @@ def datestr_to_datetime(value):
     return v
 
 
+def str_to_timedelta(value):
+    m = re.match('^(?:datetime\.timedelta)?\(?([^\)]*)\)?$', value)
```
ccordoba12: What's this regexp trying to match? Could you post an example, please?

timhoffm: The regexp matches `datetime.timedelta(ANYTHING)` where
`datetime.timedelta` and the brackets are optional. ANYTHING is then further processed.

I assume that ANYTHING is either a single `int` or a list of `ints` separated by colons and possibly whitespace. Examples of matches:

- `datetime.timedelta(1)`
- `datetime.timedelta(1, 5, 12345)`
- `(1, 5, 12345)`
- `1`
- `1, 5, 12345`

all are converted to `datetime.timedelta(1)` or `datetime.timedelta(1, 5, 12345)` obejcts.

These represent the valid string inputs if the user wants to set a new value to a `datetime.timedelta` object via the variable explorer.

ccordoba12: Ok, thanks for the explanation. It's clear to me now :+1:

spyder/widgets/variableexplorer/utils.py
ccordoba12: Please solve this one, and I'll merge.

spyder/widgets/variableexplorer/utils.py
goanpeca: Should then `timedelta(1)` be valid as well?

timhoffm: > Should then timedelta(1) be valid as well?

Fine by me. Any reasonable string could be included. We can simply adapt the regexp.

> Feels like we should probably have a date/delta editor that has a box for each part?

Wouldn't be wrong. But for me, that's too much effort. I won't implement it. After all, editing timedeltas is probably a niche case and a string interface is good enough.

ccordoba12: Yeah, I agree.

timhoffm: I've extended the matching to `timedelta(1)` and added some more documentation

goanpeca: Thanks :-)

ccordoba12:
Note: The failure in Travis is unrelated to this PR.

PR #5634, opened 2017-11-01
PR: Catch not found file error when loading profiler results.

### LINKED ISSUE #5587: Crash on profiler because of lack of profiler.results
## Description

1. 
2. 
3. 

  File "/home/amitruf/anaconda3/lib/python3.5/site-packages/spyder_profiler/widgets/profilergui.py", line 263, in <lambda>
    self.finished(ec, es))
  File "/home/amitruf/anaconda3/lib/python3.5/site-packages/spyder_profiler/widgets/profilergui.py", line 328, in finished
    self.show_data(justanalyzed=True)
  File "/home/amitruf/anaconda3/lib/python3.5/site-packages/spyder_profiler/widgets/profilergui.py", line 349, in show_data
    self.datatree.load_data(self.DATAPATH)
  File "/home/amitruf/anaconda3/lib/python3.5/site-packages/spyder_profiler/widgets/profilergui.py", line 461, in load_data
    stats_indi = [pstats.Stats(profdatafile),]
  File "/home/amitruf/anaconda3/lib/python3.5/pstats.py", line 72, in __init__
    self.init(arg)
  File "/home/amitruf/anaconda3/lib/python3.5/pstats.py", line 86, in init
    self.load_stats(arg)
  File "/home/amitruf/anaconda3/lib/python3.5/pstats.py", line 99, in load_stats
    with open(arg, 'rb') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/amitruf/.config/spyder-py3/profiler.results'

* Spyder Version: 3.2.4 
* Python Version: 3.5.4

## Dependencies
```
psutil >=0.3     :  5.4.0 (OK)

```

rlaverde:
Fixes #5587

### DIFF
--- a/spyder_profiler/widgets/profilergui.py
+++ b/spyder_profiler/widgets/profilergui.py
@@ -469,6 +469,7 @@ class ProfilerDataTree(QTreeWidget):
                     self, _("Error"),
                     _("Error when trying to load profiler results"))
                 debug_print("Error when calling pstats, {}".format(e))
+                self.compare_file = None
         map(lambda x: x.calc_callees(), stats_indi)
         self.profdata.calc_callees()
         self.stats1 = stats_indi
--- a/spyder_profiler/widgets/profilergui.py
+++ b/spyder_profiler/widgets/profilergui.py
@@ -30,7 +30,7 @@ from qtpy.QtWidgets import (QApplication, QHBoxLayout, QLabel, QMessageBox,
                             QTreeWidget, QTreeWidgetItem, QVBoxLayout, QWidget)
 
 # Local imports
-from spyder.config.base import get_conf_path, get_translation
+from spyder.config.base import get_conf_path, get_translation, debug_print
 from spyder.py3compat import to_text_string
 from spyder.utils import icon_manager as ima
 from spyder.utils.qthelpers import (create_toolbutton, get_item_user_text,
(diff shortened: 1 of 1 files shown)


spyder_profiler/widgets/profilergui.py
```diff
@@ -462,7 +462,13 @@ def load_data(self, profdatafile):
         self.profdata = stats_indi[0]
         
         if self.compare_file is not None:
-            stats_indi.append(pstats.Stats(self.compare_file))
+            try:
+                stats_indi.append(pstats.Stats(self.compare_file))
+            except IOError as e:
+                QMessageBox.critical(
+                    self, _("Error"),
+                    _("Error when trying to load profiler results"))
```
ccordoba12: I think there's no need to show this error message. We can simply swallow the error...

@rlaverde?

ccordoba12: I mean, what's the advantage for users to receive this message? Is there something they can do about it?

rlaverde: Users will know that the profiler failed, if not It'll be strange that after pressing run nothing happened

ccordoba12: I see, but then we should also make `self.compare_file = None` before showing the message and make this method return at this point, right?

rlaverde: Only making the self.compare_file = None, the rest of the method should be executed because there could be previous results that need to be loaded

ccordoba12: Ok, good :+1:

Issue #5495, opened 2017-10-18
Change --show-console for --hide-console


ccordoba12:
It's a bit strange for users and developers alike that when Spyder is started from a terminal on Windows, that terminal is hidden automatically.

That's why I propose to deprecate the `--show-console` option in Spyder 3.2.5 and remove it in 4 (by deprecating it I mean making it do nothing). Instead, we should replace it with an option called `--hide-console` to hide the terminal if users prefer that.

@rlaverde, please work on this one.

PR #5635, opened 2017-11-01
PR: Change default behavior of --show-console option

rlaverde:
Fixes: #5495
- Deprecate --show-console
- Add hide console cli parameter

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -28,12 +28,15 @@ parser = optparse.OptionParser(
     usage="python bootstrap.py [options] [-- spyder_options]",
     epilog="""\
 Arguments for Spyder's main script are specified after the --
-symbol (example: `python bootstrap.py -- --show-console`).
+symbol (example: `python bootstrap.py -- --hide-console`).
 Type `python bootstrap.py -- --help` to read about Spyder
 options.""")
 parser.add_option('--gui', default=None,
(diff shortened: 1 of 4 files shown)


bootstrap.py
```diff
@@ -154,9 +157,13 @@
 
 # --- Executing Spyder
 
-if not options.hide_console and os.name == 'nt':
-    print("0x. Enforcing parent console (Windows only)")
-    sys.argv.append("--show-console")  # Windows only: show parent console
+if options.show_console:
+    print("(Deprecated) --show console do nothing, now the default behavior "
```
ccordoba12: do -> does

ccordoba12:
Nice! I left a minor comment, then I'll merge.

spyder/app/mainwindow.py
```diff
@@ -3047,8 +3047,13 @@ def main():
     # otherwise, optparse won't be able to exit if --help option is passed
     options, args = get_options()
 
+    if options.show_console:
+        print("(Deprecated) --show console does nothing, now the default "
+              " behavior is to show the console, use --hide-console if you "
+              "want to hide it")
```
ccordoba12: Where this `print` is printed? I'd say is not necessary...

rlaverde: It's printed in the console where spyder --show-console is executed

ccordoba12: Ok, thanks for the clarification!

PR #5657, opened 2017-11-05
PR: Allow undo/redo on Source > Fix indentation

csabella:
Fixes issue #1754.

`Fix indentation` converts all the tabs in a document to 4 spaces.  Allow undo/redo on this operation.

### DIFF
--- a/spyder/widgets/sourcecode/codeeditor.py
+++ b/spyder/widgets/sourcecode/codeeditor.py
@@ -907,12 +907,14 @@ class CodeEditor(TextEditBaseWidget):
         cursor.endEditBlock()
 
     def fix_indentation(self):
-        """Replace tabs by spaces"""
+        """Replace tabs by 4 spaces."""
         text_before = to_text_string(self.toPlainText())
         text_after = sourcecode.fix_indentation(text_before)
         if text_before != text_after:
-            self.setPlainText(text_after)
-            self.document().setModified(True)
+            # We do the following rather than using self.setPlainText
+            # to benefit from QTextEdit's undo/redo feature.
+            self.selectAll()
+            self.insertPlainText(text_after)
 
     def get_current_object(self):
         """Return current object (string) """
--- a/spyder/widgets/sourcecode/tests/test_indentation.py
+++ b/spyder/widgets/sourcecode/tests/test_indentation.py
@@ -67,7 +67,7 @@ def test_single_line_indent(code_editor_indent_bot):
 
 
 def test_selection_indent(code_editor_indent_bot):
-    """Test indentation with selection of more tha one line."""
+    """Test indentation with selection of more than one line."""
     editor, qtbot = code_editor_indent_bot
     text = ("class a():\n"
             "self.b = 1\n"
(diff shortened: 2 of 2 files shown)


spyder/widgets/sourcecode/codeeditor.py
```diff
@@ -907,12 +907,14 @@ def remove_trailing_spaces(self):
...
+        """Replace tabs by 4 spaces."""
         text_before = to_text_string(self.toPlainText())
         text_after = sourcecode.fix_indentation(text_before)
         if text_before != text_after:
-            self.setPlainText(text_after)
-            self.document().setModified(True)
+            # We do the following rather than using self.setPlainText
+            # to benefit from QTextEdit's undo/redo feature.
+            self.selectAll()
+            self.insertPlainText(text_after)
```
goanpeca: Nice catch!

goanpeca:
@rlaverde want to give it another pass?

spyder/widgets/sourcecode/codeeditor.py
```diff
@@ -907,12 +907,14 @@ def remove_trailing_spaces(self):
         cursor.endEditBlock()
 
     def fix_indentation(self):
-        """Replace tabs by spaces"""
+        """Replace tabs by 4 spaces."""
```
goanpeca: @csabella isn't this fix dependent on the preference setting for what a tab should represent? Python is 4 but a user could change this to 3 or 2 if they follow google's ugly style guide.

spyder/widgets/sourcecode/tests/test_indentation.py
```diff
@@ -84,5 +84,35 @@ def test_selection_indent(code_editor_indent_bot):
                     )
 
 
+def test_fix_indentation(code_editor_indent_bot):
+    """Test fix_indentation() method."""
+    ed, qtbot = code_editor_indent_bot
```
goanpeca: Minor nitpick, but maybe editor is a better word :-), we are not saving that much space bellow anyway.

spyder/widgets/sourcecode/codeeditor.py
ccordoba12: Yes, @goanpeca is right about this one.

csabella: It doesn't use the settings now.  It's hard-coded as 4, which is why I made the comment.  It's in `sourcecode/utils.py`.
```
def fix_indentation(text):
    """Replace tabs by spaces"""
    return text.replace('\t', ' '*4)
```

goanpeca: @csabella so then we need to fix fix_indentation perhpas?

csabella: @goanpeca   Sure.  Should it be in this PR or a separate issue?

ccordoba12: I think @dalthviz fixed this one. @dalthviz?

dalthviz: I fixed something regarding the use of preferences but in the comment/uncomment functionality in #5470, is that related with this?

csabella: It's similar, but not the same.

ccordoba12: Ok, let's leave this for another PR then.

csabella: Opened issue #5678 for the number of spaces.

PR #5689, opened 2017-11-07
PR: Convert command line options from optparse to argparse

### LINKED ISSUE #5639: Use argparse instead of optparse
optparse is deprecated see: https://docs.python.org/2.7/library/optparse.html

https://github.com/spyder-ide/spyder/blob/master/spyder/app/cli_options.py

csabella:
Fixes #5639.

I tried to minimize the changes by keeping the return value from `cli_options.py` the same as it was before, even though `parse_args` returns one value instead of two.

### DIFF
--- a/spyder/app/tests/test_cli_options.py
+++ b/spyder/app/tests/test_cli_options.py
@@ -1,0 +1,64 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+#
+
+"""
+Tests for cli_options.py
+"""
+import pytest
+from spyder.app import cli_options
+
+
+def test_get_options():
+    getopt = cli_options.get_options
+
+    options, args = getopt([])
+    assert not options.new_instance
+    assert not options.reset_to_defaults
+    assert not options.reset_config_files
+    assert not options.optimize
+    assert options.working_directory is None
+    assert not options.hide_console
+    assert not options.show_console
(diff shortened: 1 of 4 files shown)


ccordoba12:

ccordoba12:
Thanks @csabella for your contribution!

Could add a simple test for this? I mean, like testing that running in a terminal that `spyder --help` returns part of a given string?

csabella:
@ccordoba12   I added some tests for `get_option`.  I'm not sure if this is what you were looking for.  There is already a `test_mainwindow` which calls `get_option`.

spyder/app/tests/test_cli_options.py
```diff
@@ -0,0 +1,64 @@
...
+    options, args = getopt('--window-title MyWindow'.split())
+    assert options.window_title == 'MyWindow'
+
+    options, args = getopt('-p myproject test_file.py another_file.py'.split())
+    assert options.open_project == 'myproject'
+    assert options.files == ['test_file.py', 'another_file.py']
+    assert args == ['test_file.py', 'another_file.py']
+
+    with pytest.raises(SystemExit):
+        options, args = getopt(['--version'])
```
rlaverde: While reviewing this PR, I noticed that spyder doesn't have a `--version` option 

Maybe we could added it (in other PR), @ccordoba12 ?

ccordoba12: Yep, sure. Please go for it, when you have some free time ;-)

PR #5691, opened 2017-11-07
PR: Improve dependencies dialog

### LINKED ISSUE #5652: Improved dependency dialog

The current dialog is not quite clear:

grafik

- it has a lengthy textual disclaimer.
- at first I got confused, because "required" does refer to the version number, whereas I thought it was used in the sense of mandatory.
- no obvious distinction bewteen mandatory and optional dependencies: I cannot see if I'm missing some optional packages that might add additional features.

**Proposal:**

I propose to exchange the table for a tree view:

~~~
+ Mandatory Packages
  - IPython
  - ...
+ Optional Packages
  - numpy
  - ...
~~~

The columns Required, Installed and Provided features will remain.

The "(OK)" text from Installed is removed. Instead, there will be an icon at the front indicating if there is an issue.

If there is interest, I can work on a pull request. Is this something for 3.x or for 4.x?

timhoffm:
Fixes #5652 

Following are screenshots of the new dialog.

Regular Spyder session:
grafik

Test case with some errors:
grafik

### DIFF
--- a/spyder/utils/icon_manager.py
+++ b/spyder/utils/icon_manager.py
@@ -217,6 +217,9 @@ _qtaargs = {
     'folding.arrow_right_on':  [('fa.caret-right',), {}],
     'folding.arrow_down_off':  [('fa.caret-down',), {'color': 'gray'}],
     'folding.arrow_down_on':   [('fa.caret-down',), {}],
+    'dependency_ok':           [('fa.check',), {}],
+    'dependency_warning':      [('fa.warning',), {'color': 'orange'}],
+    'dependency_error':        [('fa.warning',), {'color': 'darkred'}],
 }
 
 
--- a/spyder/widgets/dependencies.py
+++ b/spyder/widgets/dependencies.py
@@ -73,9 +73,9 @@ class DependenciesDialog(QDialog):
 
         self.label = QLabel(_("Optional modules are not required to run "
                               "Spyder but enhance its functions."))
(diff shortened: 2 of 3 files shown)


ccordoba12:
I really like it too!! Thanks for this @timhoffm!

I only have one suggestion: the text about optional dependencies should not be in a smaller font size, but in the same size as the text in the dialog.

timhoffm:
@ccordoba12 The font size issue is already present in the current solution. It happens with a 125% scaling.

grafik

goanpeca:
It looks great @timhoffm, I don't mind that much the current font size.

timhoffm:
@rlaverde Basically this pull request is just another way of visualizing the same data. The current `test_dependencies()` does only test that a dialog exists.

I don't have much experience in UI testing. Could you give me a hint what to test and how to do that?

rlaverde:
I think something simple, adding an optional dependency to the existing `test_dependencies()` is enough:

`dependencies.add("sympy", "Symbolic Mathematics", ">=10.0", optional=True)`

ccordoba12:
> The font size issue is already present in the current solution

Ok, didn't remember that! Let's leave things as they are then.

rlaverde:
I did the change that I suggested, so this is ready to be merged

rlaverde:
@ccordoba12 merge?

ccordoba12:
Yep. Thanks a lot @timhoffm! This is a great improvement!!

PR #5706, opened 2017-11-09
PR: Splash SVG font change and convert font to path

### LINKED ISSUE #5667: Strange font rendering in the splash

The spyder splash looks like this on Ubuntu:

current_splash

This is the same for spyder 3.

In the splash SVG file `splash.svg`, I can see that the font is set to Candara, one of the Microsoft "*C*" fonts, which is not available on my system. This font will be unavailable to anyone that is either not on windows or doesn't have Microsoft Office installed.

Unfortunately embedding Candara in the SVG or distributing it with Spyder is, most likely, against the font's license.  Embedding fonts in SVG is also not well supported.

## Potential Solution

Inkscape can be used to convert SVG text to a path.  The font will no longer be editable but the font forms will be the same on all systems regardless of installed fonts.

## Suggestion

As a suggestion, an open licensed font such as [Quattrocento Sans](https://fonts.google.com/specimen/Quattrocento+Sans) could be used instead,  but the text should still be converted to a path.

proposed_splash

bcolsen:
Fixes #5667 

Previous Splash:

current_splash

New Splash:

splash_new

I decided that I would be easier to include editable text and instructions for use in the svg file.
splash.svg looks like this in Inkscape:

splash editable text

> You haven't added new icons to Spyder. Please leave decisions about what
   icons to use to us :)

I'm glad I only changed the splash screen :)

### DIFF
--- a/spyder/images/splash.svg
+++ b/spyder/images/splash.svg
@@ -15,7 +15,7 @@
    height="250"
    id="svg2"
    sodipodi:version="0.32"
-   inkscape:version="0.92.2 (5c3e80d, 2017-08-06)"
+   inkscape:version="0.92.2 (unknown)"
    sodipodi:docname="splash.svg"
    inkscape:output_extension="org.inkscape.output.svg.inkscape"
    inkscape:export-filename="D:\Python\spyder\img_src\ico32.png"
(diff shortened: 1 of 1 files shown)


ccordoba12:
Thanks @bcolsen!

Issue #5721, opened 2017-11-10
Update obsolete QMessageBox Standard Button values


csabella:
QMessageBox.YesAll and QMessageBox.NoAll are obsolete.

https://doc.qt.io/qt-4.8/qmessagebox.html#StandardButton-enum

PR #5722, opened 2017-11-10
PR: Update obsolete enums in QMessageBox

csabella:
Fixes issue #5721.


rlaverde:
Thanks for noticing and fixing this 😄 

It looks like this enums where deprecated a long time ago (since Qt4.2), so there's not need to check the Qt version

PR #5742, opened 2017-11-14
PR: Prevent Editor dockwidget to become visible when clicking on the outline explorer if the Editor is in a separate window

### LINKED ISSUE #2883: Second Editor appears in wrong window
I have the Spyder Editor in a separate window from the Consoles and Inspector etc. When I launch Spyder, both windows appear correctly. In the Editor, I then click somewhere in the Outline. Then a second Editor opens in the Consoles window (left side, horizontal split).
Even if I close this second Editor, choose View -> Set layout, the same will happen next time Spyder is launched.

Spyder 2.3.8, Windows 64-bit.

jnsebgosselin:
Fixes #2883 

Before PR #3824, Issue #2883 was hard to reproduce. But now that separate windows are created by default when undocking widgets, this issue has become easily reproducible and is much more annoying. One simply has to undock the editor and click on the outline explorer to trigger it (see animation below).

This is due to the fact that a `setVisible(True)`, `setFocus()` and `raise_()` is systematically called on the editor dockwidget in the [load](https://github.com/spyder-ide/spyder/blob/master/spyder/plugins/editor.py#L2033) method, which is the method that is connected to the action of clicking on the outline explorer. To fix Issue #2883, we simply have to prevent these calls when the editor is in a separate window and call `setFocus` and `raise_()` on this separate window instead.

outline_focus_dockwidget

### DIFF
--- a/spyder/plugins/editor.py
+++ b/spyder/plugins/editor.py
@@ -1877,15 +1877,23 @@ class Editor(SpyderPluginWidget):
                 filenames = [osp.normpath(fname) for fname in filenames]
             else:
                 return
-            
+
         focus_widget = QApplication.focusWidget()
-        if self.dockwidget and not self.ismaximized and\
-           (not self.dockwidget.isAncestorOf(focus_widget)\
-            and not isinstance(focus_widget, CodeEditor)):
+        if self.editorwindows and not self.dockwidget.isVisible():
+            # We override the editorwindow variable to force a focus on
+            # the editor window instead of the hidden editor dockwidget.
+            # See PR #5742.
+            if editorwindow not in self.editorwindows:
+                editorwindow = self.editorwindows[0]
+            editorwindow.setFocus()
+            editorwindow.raise_()
+        elif (self.dockwidget and not self.ismaximized
+              and not self.dockwidget.isAncestorOf(focus_widget)
+              and not isinstance(focus_widget, CodeEditor)):
             self.dockwidget.setVisible(True)
             self.dockwidget.setFocus()
             self.dockwidget.raise_()
-        
+
         def _convert(fname):
             fname = osp.abspath(encoding.to_unicode_from_fs(fname))
             if os.name == 'nt' and len(fname) >= 2 and fname[1] == ':':


jnsebgosselin:
Moreover, in this PR, we are also forcing the  docked outline explorer to connect to the editor in the separate window instead of connecting to the invisible docked editor. This is done by overriding the value of the `editorwindow` variable that was passed as argument by the `edit_goto` signal of the outline explorer with `editorwindow = self.editorwindows[0]`.

fixed_outline_focus_dockwidget

jnsebgosselin:
Also, fixing this issue for the 3.x branch is a little bit more complicated because in this branch one or more editors in separate windows are allowed to live at the same time alongside a docked editor. I'll try to see what I can do and I will submit another PR for the 3.x branch if I find an acceptable solution.

jnsebgosselin:
I will continue to work on this one to see if I could find a way to fix this for Spyder 3 and 4 with a single PR. Below I try to summarize a little bit more the situation and the main difference between Spyder 3 and 4 and what is the behaviour I think should be expected when clicking on the outline explorer.

## Spyder 3

In Spyder 3, undocking the editor will simply undock the dockwidget in which the editor lives. It is however possible to open an unlimited amount of new editorstack instances in separate window through the action "New Window" in the cog menu. Opening an editorstack in a new window will not hide the dockwidget. Therefore in Spyder 3, a docked editor can live alongside one or more editors in separate windows. This is why Issue #2883 is harder (but not impossible) to fix for Spyder 3 than 4.

image

## Spyder 4

PR #3824 changed this behaviour in Spyder4. Undocking the editor while `Lock panes` is checked will now open a new editorstack instance in a separate window. The editorstacks that lived in the dockwidget are automatically hidden when doing this, but are not destroyed. They are restored when the separate editor window is closed. In Spyder4, it is not possible anymore to generate more than one separate editor window. So if I understand correctly PR #3824, the intended behaviour in Spyder 4 is to have the editor either in a dockwidget or in a unique separate window, but not both at the same time.

So following this line of thoughts, here is what I think should happen in Spyder 4 when clicking on the outline explorer:

- Clicking on either the sidebar or docked outline explorer while the editor is in a separate window should focus and return to the editor in the separate window. This operation should not focus and make the  dockwidget visible like it currently does.

- Clicking on the docked outline explorer while there is no editor visible at all should focus the main editorstack and make the editor dockwidget visible again.

- Clicking on the outline explorer while the editor is in a dockwidget should focus and return to the last focused editor in the current editorstack of the dockwidget.

ccordoba12:
My thoughts:

1. Please don't waste your time with Spyder 3. The solution there seems quite complex and probably we won't be able to merge it with master. Besides, it's not a terribly bad bug to spend a lot of time with it.
2. You're right about our proposed solution for Spyder 4, i.e. Spyder will only have one Editor instance visible all the time (either docked or in a new window). The same goes for all other plugins.
3. I agree with all the thoughts you proposed about what *should happen in Spyder 4 when clicking on the outline explorer*.

jnsebgosselin:
This PR also fixes the undesired behaviour when opening a file from the the Spyder mainwindow when the editor is in a separate window.

fixed_outline_focus_dockwidget2

jnsebgosselin:
Thank you @ccordoba12 for your inputs!

I've made some minor changes and tested it for Spyder 3. The PR solve the undesired behaviour reported in Issue #2883 for Spyder 3. The only thing that doesn't work perfectly in Spyder3 is when multiple editorwindows were created and a `edit_goto` is emitted from the mainwindow while the editor dockwidget is not visible. In this case, because there is no way to easily get the last focused editorwindow, the first editorwindow in the list will be focused.

Everything else seems to be working as expected though.

ccordoba12:
> Everything else seems to be working as expected though.

Great! Thanks a lot for taking a look at this! Please move this PR to `3.x` so we can start to review it.

@dalthviz, please review Jean's work.

ccordoba12:
Pinging @dalthviz again about this one ;-)

PR #5746, opened 2017-11-14
PR: Manage types that extend str in value_to_display

### LINKED ISSUE #5636: "ValueError: malformed node or string" when evaluating code that contains Path elements
## Description

1. 
2. 
3. 

  File "/usr/local/lib/python2.7/dist-packages/qtconsole/base_frontend_mixin.py", line 163, in _dispatch
    handler(msg)
  File "/usr/local/lib/python2.7/dist-packages/spyder/widgets/ipythonconsole/namespacebrowser.py", line 220, in _handle_execute_reply
    self.handle_exec_method(msg)
  File "/usr/local/lib/python2.7/dist-packages/spyder/widgets/ipythonconsole/shell.py", line 351, in handle_exec_method
    view = ast.literal_eval(literal)
  File "/usr/lib/python2.7/ast.py", line 80, in literal_eval
    return _convert(node_or_string)
  File "/usr/lib/python2.7/ast.py", line 63, in _convert
    in zip(node.keys, node.values))
  File "/usr/lib/python2.7/ast.py", line 62, in <genexpr>
    return dict((_convert(k), _convert(v)) for k, v
  File "/usr/lib/python2.7/ast.py", line 63, in _convert
    in zip(node.keys, node.values))
  File "/usr/lib/python2.7/ast.py", line 62, in <genexpr>
    return dict((_convert(k), _convert(v)) for k, v
  File "/usr/lib/python2.7/ast.py", line 79, in _convert
    raise ValueError('malformed string')
ValueError: malformed string

* Spyder Version: 3.2.3 
* Python Version: 2.7.12

## Dependencies
```
pyflakes >=0.5.0  :  1.6.0 (OK)
pycodestyle >=2.3 :  2.3.1 (OK)
pandas >=0.13.1   :  0.16.2 (OK)
numpy >=1.7       :  1.12.1 (OK)
sphinx >=0.6.6    :  1.6.3 (OK)
rope >=0.9.4      :  0.10.7 (OK)
jedi >=0.9.0      :  0.10.2 (OK)
psutil >=0.3      :  5.3.1 (OK)
sympy >=0.7.3     :  0.7.6 (OK)
cython >=0.21     :  0.22.1 (OK)
qtconsole >=4.2.0 :  4.3.1 (OK)
IPython >=4.0;<6.0:  5.4.1 (OK)
pylint >=0.25     :  1.7.2 (OK)

```

dalthviz:
Fixes #5636


dalthviz:
@ccordoba12 I think this is ready

spyder/widgets/variableexplorer/utils.py
```diff
@@ -11,12 +11,14 @@
 from __future__ import print_function
 
 from itertools import islice
+import ast
```
ccordoba12: This import is not necessary anymore.

spyder/py3compat.py
```diff
@@ -82,6 +82,16 @@
 #==============================================================================
 # Strings
 #==============================================================================
+def is_type_text_string(obj):
+    """Return True if `obj` is type text string, False if it is anything else,
+    like binary data (Python 3) or QString (Python 2, PyQt API #1)"""
+    if PY2:
+        # Python 2
+        return type(obj) in [basestring]
```
ccordoba12: Why do you need to enclose `basetring` in brackets?

spyder/py3compat.py
```diff
@@ -82,6 +82,16 @@
 #==============================================================================
 # Strings
 #==============================================================================
+def is_type_text_string(obj):
```
ccordoba12: What's the purpose of this function? I mean, why does it cover that it's not covered by `is_text_string`? (I guess the test will reveal it, and that's why I asked you for it :-).

spyder/py3compat.py
dalthviz: I will change this :+1:

spyder/py3compat.py
dalthviz: The propouse of this is to avoid letting in values in the `value_to_display` handling of strings. So for example, in the case of `Path` from the `path` package, it extends `str`, so a value passed to the `value_to_display` function currently would fall in managing a object `Path` as a string instead of making a default display of the element. For example, in Python 3, something that is instance of `str` normally should be type `str`, but if is something like `Path` it would be a instance of `str` and would have type `path.Path` as his type, then it shouldn't have to be treated as a string.

tyralla:
You are right, @dalthviz  and @ccordoba12.   My issue #5864 was actually a duplicate.

I merged the proposed changes to modules `pycompat` and `widgets\variableexplorer\utils` manually.  Now everything works fine.

Many thanks for the fast help!

ccordoba12:
@dalthviz, can't we take the repr of objects that extent `str` to show it in the Variable Explorer?

dalthviz:
@ccordoba12 thats the problem, we can be sure if the `__repr__` method is implemented in a way that a call with `ast.literal_eval` will not cause a `ValueError`

ccordoba12:
@dalthviz, I simplified the test you added and squashed a lot of commits you made about it. Please don't add more commits here, this is ready.

spyder/py3compat.py
ccordoba12: This turned to be a good idea, thanks @dalthviz!

PR #5764, opened 2017-11-15
PR: Handle namespace browser column width

### LINKED ISSUE #5543: Variable explorer column widths are re-generated in every console evaluation
## Description

The variable explorer window re-generates the column widths on creation of every new variable, over-riding any user modification of the column widths. This is annoying as it also tends to make th type column wide enough to display full type information even though other columns (e.g. the Value) would be the ones the user is more interested in.

1.  Set numpy and pylab to be automatically loaded. Observe that the variable explorer window has  entries for a number of items with long type names (e.g. cast and nbytes).
2.  Shrink the size the of the type column to allow the Value column to be larger
3.  Create a new variable in the console
4. Goto 2

Either a smarter layout of columns  when the variable explorer is horizontally constrained, or some mechanism to 'freeze' the column widths or hide/collapse columns.

* Spyder Version: 3.2.4 
* Python Version: 3.6.1

## Dependencies
```
sympy >=0.7.3    :  1.0 (OK)

```

Prikers:
Fixes #5543

### DIFF
--- a/spyder/plugins/variableexplorer/widgets/collectionseditor.py
+++ b/spyder/plugins/variableexplorer/widgets/collectionseditor.py
@@ -32,7 +32,7 @@ from qtpy.QtCore import (QAbstractTableModel, QDateTime, QModelIndex, Qt,
                          Signal, Slot)
 from qtpy.QtGui import QColor, QKeySequence
 from qtpy.QtWidgets import (QAbstractItemDelegate, QApplication, QDateEdit,
-                            QDateTimeEdit, QDialog, QHBoxLayout,
+                            QDateTimeEdit, QDialog, QHBoxLayout, QHeaderView,
                             QInputDialog, QItemDelegate, QLineEdit, QMenu,
                             QMessageBox, QPushButton, QTableView,
                             QVBoxLayout, QWidget)
@@ -685,6 +685,32 @@ class CollectionsDelegate(QItemDelegate):
         self.set_value(index, value)
 
 
+class BaseHeaderView(QHeaderView):
+    """
+    A header view for the BaseTableView that emits a signal when the width of
+    one of its section is resized by the user.
+    """
+    sig_user_resized_section = Signal(int, int, int)
+
+    def __init__(self, parent=None):
+        super(BaseHeaderView, self).__init__(Qt.Horizontal, parent)
+        self._section_handle_ispressed = False
+        self.sectionResized.connect(self.sectionResizeEvent)
+
+    def mousePressEvent(self, e):
+        self._section_handle_ispressed = (self.cursor().shape() ==
+                                          Qt.SplitHCursor)
+        super(BaseHeaderView, self).mousePressEvent(e)
+
+    def mouseReleaseEvent(self, e):
+        self._section_handle_ispressed = False
+        super(BaseHeaderView, self).mouseReleaseEvent(e)
+
+    def sectionResizeEvent(self, logicalIndex, oldSize, newSize):
+        if self._section_handle_ispressed:
+            self.sig_user_resized_section.emit(logicalIndex, oldSize, newSize)
+
+
 class BaseTableView(QTableView):
     """Base collection editor table view"""
     sig_option_changed = Signal(str, object)
@@ -711,8 +737,10 @@ class BaseTableView(QTableView):
         self.duplicate_action = None
         self.delegate = None
         self.setAcceptDrops(True)
-        self._updating_column_width = False
         self.automatic_column_width = True
+        self.setHorizontalHeader(BaseHeaderView(parent=self))
+        self.horizontalHeader().sig_user_resized_section.connect(
+            self.user_resize_columns)
 
     def setup_table(self):
         """Setup table"""
@@ -869,14 +897,16 @@ class BaseTableView(QTableView):
         self.hist_action.setVisible(condition_hist or is_list)
         self.imshow_action.setVisible(condition_imshow)
         self.save_array_action.setVisible(is_array)
-        
+
+    def user_resize_columns(self, logical_index, old_size, new_size):
+        """Handle the user resize action."""
+        self.automatic_column_width = False
+
     def adjust_columns(self):
         """Resize two first columns to contents"""
         if self.automatic_column_width:
-            self._updating_column_width = True
             for col in range(3):
                 self.resizeColumnToContents(col)
-            self._updating_column_width = False
 
     def set_data(self, data):
         """Set table data"""
--- a/spyder/plugins/variableexplorer/widgets/collectionseditor.py
+++ b/spyder/plugins/variableexplorer/widgets/collectionseditor.py
@@ -711,7 +711,9 @@ class BaseTableView(QTableView):
         self.duplicate_action = None
         self.delegate = None
         self.setAcceptDrops(True)
-        
+        self._updating_column_width = False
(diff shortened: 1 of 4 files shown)


Prikers:
As suggested by @gb119 the behavior is as follows:

- The column width resizing is automatic.
- If the user manually changes any column width (by dragging the header) if sets `automatic_column_width` to `False` in the NamespaceBrowser. Column width then does not change anymore unless user adjusts it.
- Hitting the "Reset Namespace" button resets the original (automatic) behavior

colwidth

Prikers:
There are however a couple of caveats that I was not able to get rid of...

- [ ] If the very first action (before running a script) is adding a variable in the iPython console, it sets the `automatic_column_width` to False. I was not able to identify where in the code a column resizing is called... Printing the traceback did not help. Would you have any idea?
- [ ] For a misterious reason the whole thing stops working as soon as I remove the print statement I had in `resize_columns` in `namespacebrowser.py`. The screenshot was taken with a print statement in there. I removed the print statement and it stops working: could you guide me for this one?

ccordoba12:
@Prikers, please merge this one with 3.x to fix the merge conflict. I'll help you during this weekend to identify the source of your first problem. For the other one, if a `print` is necessary, then we should leave it.

Prikers:
Thanks @ccordoba12! I have merged with the latest `3.x`.

ccordoba12:
@Prikers, could you add the `print` you say it's necessary as part of this PR? Also, please add `# spyder: test-skip` next to it, so that our tests don't fail because of that.

jnsebgosselin:
Great work @Prikers, I like that. Would it be a good addition to add a button in the plugin toolbar that would `Resize column to content` and reset the `automatic_column_width`.

I'm proposing this, because first, someone might want to restore the automatic resizing behaviour without wiping the namespace. Second, that would make the implementation of a similar feature for https://github.com/spyder-ide/spyder/issues/6179 easier and more consistent with this PR, because we do not have a `Remove all variables` tool button in the `File Explorer` plugin.

Prikers:
@ccordoba12 sorry for being late. I added the print statement. 
@jnsebgosselin good idea! Sounds good to me. @ccordoba12 should it be part of this PR (which is `3.x`) or should I do that in another one?

gb119:
Is there a reason why to add another toolbar button rather than add a context menu entry? That avoids the question of finding a suitable icon (at the expense of being another string to internationalise I guess), and it feels like quite a small bit of functionality to be given an-always-visible bit of UI....

But you guys are the coders, I'm just the fussy end-user :-) !

jnsebgosselin:
@gb119 that is also a good idea. I didn't know that there was an action `Resize rows to content` in the context menu. We might as well add a `Resize columns to content` action too.

Anyway, either we add a toolbutton, a new action in the context menu, or both, but I think we need to add something if we want to be able to fix #6179 in a Ux consistent way.

If we add this functionality, maybe that clicking on the `Reset all variables` toolbutton should not reset the internal flag?

image

jnsebgosselin:
@Prikers I'll help you debug the `Print('.')` mysterious behaviour. It should not be needed. I will try to take a look at it tonight.

Prikers:
@jnsebgosselin you're right. A `time.sleep(xxx)` may be more appropriate?

jnsebgosselin:
@Prikers I've looked into this for some time and couldn't find a viable solution... I'm sorry. My guess is that some events/signals are done asynchronously and are fighting for the custom `_updating_column_width` flag.

I understand that at first it seemed appealing to use a custom flag to block when sections are resized programmatically. That would have been my first approach too. This works well for a simple application, but for a multilayered application like Spyder, I'm not sure it is the best route to take. It is not easy to understand and to debug and there is a high risk that it is going to be broken accidentally in the future by another PR.

I think that a better approach here would be to use in `BaseTableView` a custom class derived from `QHeaderView` that sends a custom signal when the section is resized by the user. I think this will make your code easier to understand, more modular, and more robust. If you think that could be a good idea, maybe you can start with the code I've put toghether below. Use `setHorizontalHeader` to install the custom header in the `BaseTableView`. Note that I haven't tested it extensively.

```python
class BaseHeaderView(QHeaderView):
    """
    A header view for the BaseTableView that emits a signal when the width of
    one of its section is resized by the user.
    """
    sig_user_resized_section = Signal(int, int, int)

        super(BaseHeaderView, self).__init__(Qt.Horizontal, parent)
        self._section_handle_ispressed = False
        self.sectionResized.connect(self.sectionResizeEvent)

    def mousePressEvent(self, e):
        self._section_handle_ispressed = (self.cursor().shape() ==
                                          Qt.SplitHCursor)
        super(BaseHeaderView, self).mousePressEvent(e)

    def mouseReleaseEvent(self, e):
        self._section_handle_ispressed = False
        super(BaseHeaderView, self).mouseReleaseEvent(e)

    def sectionResizeEvent(self, logicalIndex, oldSize, newSize):
        if self._section_handle_ispressed:
            self.sig_user_resized_section.emit(logicalIndex, oldSize, newSize)
```

ccordoba12:
Since this still needs work, I'm moving it for 3.2.8

Prikers:
@jnsebgosselin indeed I was not super happy with the result. I also think this is due to asynchronous processes. But I could not think of your solution at that time: it seems like a great one, thanks a lot! :+1: 
As soon as I find some time I will update this PR (or start another one from scratch if simpler). Thanks again!

CAM-Gerlach:
Hey @Prikers , any plans to finish this? It looks like a nice addition; thanks!

ccordoba12:
@dalthviz, please move this one to master and try to finish it.

dalthviz:
A new preview:

nsb

ccordoba12:
It looks really good @dalthviz!

Did you solve the problems reported by @Prikers in his comment above?

https://github.com/spyder-ide/spyder/pull/5764#issuecomment-344746897

dalthviz:
Yep, following @jnsebgosselin suggestion of using a custom header, the problems where solved (no need of a `print` statement and, even if you add a variable before running a script, the autoresize flag keeps the `True` value as you can see in the preview)

spyder/plugins/variableexplorer/widgets/collectionseditor.py
```diff
@@ -685,6 +685,32 @@ def setModelData(self, editor, model, index):
         self.set_value(index, value)
 
 
+class BaseHeaderView(QHeaderView):
+    """
+    A header view for the BaseTableView that emits a signal when the width of
+    one of its section is resized by the user.
```
ccordoba12: Please change `section` for `sections` here (it's the grammatically correct expression in the context of this sentence).

spyder/plugins/variableexplorer/widgets/collectionseditor.py
```diff
@@ -685,6 +685,32 @@ def setModelData(self, editor, model, index):
...
+class BaseHeaderView(QHeaderView):
+    """
+    A header view for the BaseTableView that emits a signal when the width of
+    one of its section is resized by the user.
+    """
+    sig_user_resized_section = Signal(int, int, int)
+
+    def __init__(self, parent=None):
+        super(BaseHeaderView, self).__init__(Qt.Horizontal, parent)
+        self._section_handle_ispressed = False
```
ccordoba12: Please change the name of this attribute to `self._handle_section_is_pressed`.

PR #5771, opened 2017-11-16
PR: Make arrow-key selection in completion widget "roll over"

### LINKED ISSUE #5763: Make arrow-key selection in multiple-option tab-completion dialogs "roll over"
A bit of a long winded name, but essentially a simple fix (hopefully). In tab completion dialogs, when multiple completion options are possible, it would be a huge efficiency boost to have scrolling the list with up or downarrow "roll over" when reaching the end; i.e., when the first item is selected (as when first getting the dialog), pressing "uparrow" will select the last item, and continue up from there with subsequent presses; similarly, "downarrow" on the last item will go back to the beginning. 

With a list of say n=3 items, it means that any item can be reached with at most one keypress vs. two, and with n=5 with at most two keypresses vs. four. In particular, it can save a lot of time being able to hit "up" and go directly to the last item, if you know that's the one you want. 

(Platform: Windows 8.1/10 x64 Pro/Ent, in case it matters...) Thanks!

jnsebgosselin:
Fixes #5763

----

### What this does:

- When focus is on the `CompletionWidget` and the <kbd>Up</kbd> arrow key is clicked while already at the top of the list, go to the bottom of the list.
- When focus is on the `CompletionWidget` and the <kbd>Down</kbd> arrow key is clicked while already at the bottom of the list, return at the top of the list.

rollover_completionlist


spyder/widgets/sourcecode/base.py
```diff
@@ -174,7 +174,12 @@ def keyPressEvent(self, event):
         elif key in (Qt.Key_Up, Qt.Key_Down, Qt.Key_PageUp, Qt.Key_PageDown,
                      Qt.Key_Home, Qt.Key_End,
                      Qt.Key_CapsLock) and not modifier:
-            QListWidget.keyPressEvent(self, event)
+            if key == Qt.Key_Up and self.currentRow() == 0:
+                self.setCurrentRow(self.count()-1)
```
goanpeca: self.setCurrentRow(self.count() - 1)

PR #5784, opened 2017-11-17
PR: Change for file save on split editor when tabs are moved

### LINKED ISSUE #5703: Save after moving editor tabs with split panes changes wrong files
I couldn't find if this was an open issue, so I apologize if it's a duplicate.

If the editor is open with split panes and the tabs are manually moved in the panes, after editing and saving a tab (file) that is in a different order than the split, the other pane then is updated with the information in the same order.  Following the steps to reproduce will make this more clear than a description.

0. Note: do this on a test file, not a real one!
1. Open three files in the editor.
2. Split vertically or horizontally.
3. Move the order of the tabs on the top pane, so instead of 1-2-3, it's now 3-2-1 (or whatever).  The split pane stays in order 1-2-3.
4.  Edit the file on tab 3 and then click save.
5.  Click on the split pane.
6.  The name of the file on tab 3 changes to match pane 1, but the file doesn't change.  So, the split panel now has 2 tabs open that point to the same file name.
7.  Edit the file in the split panel where the tab name changed and then save.  Notice that all the open files now match that data -- even the ones that originally had different names.

I expect the file names on the tabs to not change and the text on each file to sync between tabs based on filename, not based on tab position.

* Spyder Version: 3.x and 4.x
* Python Version: 3.6.3
* Qt Version: 5.6.2
* PyQt Version:  PyQt5 5.6
* Operating system: Ubuntu

## Dependencies

and paste the contents below:

csabella:
Fixes #5703.

This affected `save()`, `save_as()`, and `save_copy_as()` directly.  Also added changes for `set_todo_results()` and `set_analysis_results()` as they were affected by the issue of using the same index across editor stacks.

### DIFF
--- a/spyder/widgets/editor.py
+++ b/spyder/widgets/editor.py
@@ -440,8 +440,8 @@ class EditorStack(QWidget):
     zoom_out = Signal()
     zoom_reset = Signal()
     sig_close_file = Signal(str, str)
-    file_saved = Signal(str, int, str)
-    file_renamed_in_data = Signal(str, int, str)
+    file_saved = Signal(str, str, str)
+    file_renamed_in_data = Signal(str, str, str)
     create_new_window = Signal()
     opened_files_list_changed = Signal()
     analysis_results_changed = Signal()
@@ -1231,7 +1231,10 @@ class EditorStack(QWidget):
             self.tabs.setTabToolTip(index, tab_tip)
         self.tabs.blockSignals(False)
 
-    def rename_in_data(self, index, new_filename):
+    def rename_in_data(self, original_filename, new_filename):
+        index = self.has_filename(original_filename)
+        if index is None:
+            return
         finfo = self.data[index]
         if osp.splitext(finfo.filename)[1] != osp.splitext(new_filename)[1]:
             # File type has changed!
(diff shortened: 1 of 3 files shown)


spyder/widgets/editor.py
```diff
@@ -1690,8 +1780,6 @@ def save_copy_as(self, index=None):
             txt = to_text_string(finfo.editor.get_text_with_eol())
             try:
                 finfo.encoding = encoding.write(txt, filename, finfo.encoding)
-                self.file_saved.emit(str(id(self)), index, filename)
```
ccordoba12: Why is this not needed anymore?

ccordoba12:
Thanks for your contribution @csabella! It looks good to me, except for a minor comment I left. Please also fix the style issues reported by `pep8speaks`.

spyder/widgets/editor.py
csabella: @ccordoba12   The `copy as` function creates a new file with the same information as the current index, but it doesn't make any changes to the current index or current filename (in the tab or the self.data). The `file_saved.emit` was updating the other split panels for the current index/filename, but this isn't necessary since it's not changed at all.

It's simple to recreate to see what's happening, even without moving tabs.
1.  Open one file in one panel.
2.  Split horizontally or vertically.
3.  Click `save_copy_as` and save the file under a new name.  You'll see the new file open in both panels, so there are now 2 tabs.
4.  Click focus onto the other panel.
5.  Notice that both tabs get updated with the same file name and no longer match the first panel.

This one really isn't related to moving tabs, but I found it while testing my changes.  If you want, I can move it to a separate PR.

ccordoba12: Don't worry, it's really good that you have detected and solved this problem as part of this PR.

ccordoba12:
I think the reason why Pierre (Spyder's original author) didn't implement the functionality to move tabs in the Editor before was because of all the issues that would arise with split panes.

But thanks to the great work of Cheryl and @jnsebgosselin, now this is a thing from the past!!

PR #5836, opened 2017-11-26
PR: Give a more modern appearence to the tour

timhoffm:
# Preface

This is a minor change and I don't know if you agree with it. If you don't, simly reject it. I didn't bother to first create an issue to discuss it, because it would be more work than just doing the pull request.

# What this is about

The current tour looks like this

grafik

IMHO OS design trends have changed to more flat and square style. So the rounded corners look a bit old-fashioned. I propose to just remove them.

grafik

### DIFF
--- a/spyder/app/tour.py
+++ b/spyder/app/tour.py
@@ -582,7 +582,7 @@ class FadingTipBox(FadingDialog):
                              color: rgbs(0,0,0,100%);
                              border-style: outset;
                              border-width: 1px;
-                             border-radius: 3px;
+                             border-radius: 0px;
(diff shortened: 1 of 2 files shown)


ccordoba12:
I like it! Thanks a lot for your contribution @timhoffm!

timhoffm:
@ccordoba12 Great to hear!

In that case, I'd proprose to also change the PushButtons to ToolButtons with icons:

grafik

If you approve, I'll add this to the pull request.

ccordoba12:
Great work! It looks really nice now!!

PR #5842, opened 2017-11-27
PR: Remove `if/else/try/for` statements from the outline explorer tree

### LINKED ISSUE #5821: Outline Explorer: disable `if/else/try/for` statements
The Outline Explorer is really useful to jump to certain parts of the code quickly -- especially to functions, classes, code cells, and even comments starting with `# ---`. Especially the latter two help to structure a longer script better. However, the Outline Explorer also shows _every_ `if/else/try/for` which makes the outline basically useless: for a rather long file, the actual main structures (functions, cells, etc.) are somewhere in between several `if/else/try/for` code statements that are rather meaningless for an actual "outline" (especially the single `else:` lines).

For this reason I ask whether you can implement an additional option in the right-click menu like "Show if/else/try/for... statements" ... just as there exists "Show special comments" to toggle code cell and `# ---` comments. Thanks for considering!

jnsebgosselin:
Fixes #5821

- [x] Fixes issue in the code
- [x] Write tests

The goal of this PR is to prevent `if/else/try/for` statements from showing in the outline explorer tree.

image

### DIFF
--- a/spyder/widgets/tests/test_outline_explorer.py
+++ b/spyder/widgets/tests/test_outline_explorer.py
@@ -1,0 +1,86 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+#
+
+"""
+Tests for editor.py
+"""
+
+# Third party imports
+import pytest
+
+# Local imports
+from spyder.widgets.editortools import OutlineExplorerWidget
+from spyder.widgets.sourcecode.codeeditor import CodeEditor
+
+
+text = ("# -*- coding: utf-8 -*-\n"
+        "\n"
+        "# %% functions\n"
+        "\n"
+        "\n"
+        "# ---- func 1 and 2\n"
+        "\n"
+        "def func1():\n"
+        "    for i in range(3):\n"
+        "        print(i)\n"
+        "\n"
+        "\n"
+        "def func3():\n"
+        "    if True:\n"
+        "        pass\n"
+        "\n"
+        "\n"
+        "# ---- func 3\n"
+        "\n"
+        "def func3():\n"
+        "    pass\n"
+        "\n"
+        "\n"
+        "# %% classes\n"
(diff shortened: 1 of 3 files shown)


jnsebgosselin:
For comparison, here is a screenshot for the same file before the PR:

image

jnsebgosselin:
@ccordoba12 Ok, I added the blank spaces back, sorry for that. 

I just need to write a test now. I would like to add a docstring to the method `populate_branch` also. Do you have a suggestion?

spyder/widgets/tests/test_outline_explorer.py
```diff
@@ -0,0 +1,86 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+#
+
+"""
+Tests for editor.py
```
ccordoba12: This is wrong.

ccordoba12:
> Do you have a suggestion?

Nop. Whatever you decided is fine.

jnsebgosselin:
@ccordoba12 I've tested this change for a few days and I like it. The only thing though is I'm missing when `if __name__ == '__main__': ` was displayed in the outline explorer tree. I'm tempted to let this one pass through. What do you think?

Also, I'll try to work on the tests soon.

jnsebgosselin:
> @ccordoba12 I've tested this change for a few days and I like it. The only thing though is I'm missing when if __name__ == '__main__': was displayed in the outline explorer tree. I'm tempted to let this one pass through. What do you think?

Ok forget it, I've changed my mind. If one wants a link to `if __name__ == '__main__': `, just add a `# ----` comment before it. I'll start working on the tests now.

ccordoba12:
Ok, great, so just the test is missing :-)

ccordoba12:
@jnsebgosselin, is this ready?

jnsebgosselin:
Yes, it is ready.

PR #5947, opened 2017-12-10
PR: Collect virtual environment name (also should work for conda envs)

### LINKED ISSUE #5803: Spyder should be conda environment-sensitive
I use spyder in different anaoconda environments. I typically create them by cloning root of a standard base anaconda environment. 

This creates start menu entries "Spyder (env_foo)", "Spyder (env_bar)", and so, which work fine.

The possible improvements I would see to the current behavior:
1. Indicate in the title bar the environment relevant to the running spyder. If I have open one spyder from yesterday I may not remember which environment is behind it.
2. Have environment-related preferences and session-persistence. For example the open scripts in the editor, recent files...
3. Allow multiple spyder instances from different environments. This right now is not working for me (the 2nd instance of spyder is never launched?)

albertogomcas:
Fixes #5803

I had a go at this, I tried a few approaches for conda and virtual environments, for some reason os.environ['CONDA_PREFIX'] did not work reliably for me.

However [sys.prefix](https://stackoverflow.com/questions/1871549/python-determine-if-running-inside-virtualenv) provides a reasonable result both for venv and conda environments. Mind that the root conda environment is also detected, which for me is not a problem.

I only tested this with windows. In any case, the proposed changes should fail cleanly and leave the title as-is if something does not work. 

image
(Top: spyder running in conda root, bottom, running in a conda environment named "condaclone")

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -222,6 +222,19 @@ def get_focus_python_shell():
     elif isinstance(widget, ExternalPythonShell):
         return widget.shell
 
+def get_environment_infostr():
+    """Find if running inside a conda and/or venv
+    Return None if not the case or anything fails"""
+    debug_print("Collecting environment information")
+    try:
+        try:
+            return "env: {}".format(os.path.basename(sys.prefix))
+        except AttributeError:
+            return ""
+    except Exception as e:
+        debug_print("Exception {} while trying to obtain environment information".format(e))
+
+    return ""
 
 #==============================================================================
 # Main Window
@@ -446,6 +459,10 @@ class MainWindow(QMainWindow):
         else:
             title = "Spyder (Python %s.%s)" % (sys.version_info[0],
                                                sys.version_info[1])
+        env_info = get_environment_infostr()
+        if env_info:
+            title += " [%s]" % env_info
+
         if DEBUG:
             title += " [DEBUG MODE %d]" % DEBUG
         if options.window_title is not None:
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -222,6 +222,7 @@ def get_focus_python_shell():
     elif isinstance(widget, ExternalPythonShell):
         return widget.shell
 
+
 def get_environment_infostr():
     """Find if running inside a conda and/or venv
     Return None if not the case or anything fails"""
@@ -232,7 +233,9 @@ def get_environment_infostr():
         except AttributeError:
             return ""
     except Exception as e:
-        debug_print("Exception {} while trying to obtain environment information".format(e))
+        debug_print(
+        "Exception {} while trying to obtain environment information".format(e)
+        )
 
     return ""
 


CAM-Gerlach:
@albertogomcas Pro tip: Link issues and PRs to one another properly by simply typing # and the issue number, like this: Partially addresses #5803  Also, you have a few PEP8 issues to fix, FYI.

CAM-Gerlach:
@albertogomcas Any updates? Also, can you add a quick test for this?

albertogomcas:
Sorry, I have very little time lately (little new member at home). I fixed the PEP8 issues, but I have trouble with the tests. If I lauch the test run with pytest it passes test_cli_options but when it starts with test_mainwindow a Spyder instance is launched and nothing else happens. I have to close those by hand, generating errors/skips.

This is before I add any test myself, but of course prevents me from trying anything.

spyder/app/mainwindow.py
```diff
@@ -223,6 +223,22 @@ def get_focus_python_shell():
         return widget.shell
 
 
+def get_environment_infostr():
+    """Find if running inside a conda and/or venv
```
CAM-Gerlach: Technically, shouldn't be this something like "Get the name of the conda or venv environment if running inside one" as the current wording implies it returns a bool of whether or not it is running in a specific env?

spyder/app/mainwindow.py
```diff
@@ -223,6 +223,22 @@ def get_focus_python_shell():
...
+            return "env: {}".format(os.path.basename(sys.prefix))
+        except AttributeError:
+            return ""
+    except Exception as e:
+        debug_print(
+        "Exception {} while trying to obtain environment information".format(e)
+        )
+
+    return ""
+
```
CAM-Gerlach: Extra linebreak here in case you didn't catch this already.

jitseniesen:
@albertogomcas You need to run the tests using the `runtests.py` script, if I remember correctly.

Congratulations with the family addition! Don't worry, everybody understands that time is hard to find in these circumstances.

spyder/app/mainwindow.py
```diff
@@ -223,6 +223,22 @@ def get_focus_python_shell():
...
+        try:
+            return "env: {}".format(os.path.basename(sys.prefix))
+        except AttributeError:
+            return ""
+    except Exception as e:
+        debug_print(
+        "Exception {} while trying to obtain environment information".format(e)
+        )
+
+    return ""
```
CAM-Gerlach: You say "Return None if not the case or anything fails" but it actually returns an empty string.

spyder/app/mainwindow.py
```diff
@@ -223,6 +223,22 @@ def get_focus_python_shell():
...
 
+def get_environment_infostr():
+    """Find if running inside a conda and/or venv
+    Return None if not the case or anything fails"""
+    debug_print("Collecting environment information")
+    try:
+        try:
+            return "env: {}".format(os.path.basename(sys.prefix))
+        except AttributeError:
+            return ""
```
CAM-Gerlach: Same here.

CAM-Gerlach:
> Sorry, I have very little time lately (little new member at home).

No worries; no rush.

> If I lauch the test run with pytest it passes test_cli_options but when it starts with test_mainwindow a Spyder instance is launched and nothing else happens.

I believe its mentioned a few places in our documentation, but you need to run the tests (at least those involving the main window) with ``python runtests.py`` from the root of your local copy of the repo, and in an environment with the test dependencies installed of course (which I assume you do since it at least launched the main window). Also, if you have further test problems, make sure you rebase your branch on the latest ``master`` since we've made a number of fixes/improvements in the test suite since you've started.

> This is before I add any test myself, but of course prevents me from trying anything.

It shouldn't be too difficult, particularly if you just test the helper function and that it matches the relevant env variable (hardcoding it would be much too flimsy).

CAM-Gerlach:
Hey @albertogomcas , any updates?

CAM-Gerlach:
@albertogomcas ? Let us know the status on this, if you can; the changes requested are relatively minor. If we don't hear from you soon, this will probably be closed as stale. Thanks!

CAM-Gerlach:
@albertogomcas Any interest in finishing this? Thanks!

CAM-Gerlach:
@albertogomcas Hey, any chance this will get finished anytime soon? It shouldn't be too much more work than you've already done and would be a nice addition to Spyder. Thanks!

goanpeca:
Thanks for the initiative @albertogomcas. With recent work for Spyder 4, we will include a new StatusWidget for displaying this information on the status bar, plus a selector of the different environments found.

Thanks

PR #5991, opened 2017-12-15
PR: Properly disable editing in Variable Explorer for values in immutable collections (e.g. tuples)

### LINKED ISSUE #5953: Tuples can be "edited" in Variable Explorer but changes are discarded
## Description

In the Variable Explorer, if one opens a tuple and double clicks on one of the elements to edit, the field becomes editable and the user can enter, modify or delete text. However, upon hitting return/okay, the changes are (of course, as they should be) discarded. It would seem for consistency with tuples' immutability and to minimize new user confusion about their changes not being saved, that it would be best to make the field selectable (e.g. can select text like a text box as you can currently) but not editable. I presume there's a way to do that by selecting a different type or parameter for the Qt text field object?

1. Create any tuple, e.g. 
```
tup = ("foo", "bar", 42)
```
2. Open in variable explorer and double click an element value
3. Edit field contents then click off/enter

Expected: When double clicked, text becomes selectable as if it were a text field, but cannot be changed/edited. 
Actual: Field can be edited like any other text box, but changes of course discarded when clicked off of/"accepted"

* Spyder Version: 3.2.4 
* Python Version: 3.6.3

## Dependencies
```

```

CAM-Gerlach:
Fixes #5953

@ccordoba12 @Prikers 

First "real" (code) PR. Before, values in immutable (readonly=True) collections were able to be "edited" in a textbox on doubleclick in a Variable Explorer subwindow, but the changes wouldn't stick (as they shouldn't) once enter/etc was pressed, leading to inconsistency and potential user confusion. 

To remedy this, a naive fix as presented here would be to simply check if the readonly property is set when attempting to edit a directly editable value (i.e. not an object that can itself be inspected in a variable explorer subwindow, which might be desired even inside a tuple), and if so, do nothing. Took about 60 seconds to actually write and test the relevant line, and 60 minutes to manually track down where to stick said line...no surprise, lol.

Is this the best approach? Does it break anything else? I ran some simple manual tests with mutable and immutable objects, and viewable objects inside tuples, and everything check out okay, but I'm sure there's edge/corner cases this isn't covering. I rarely do much OO programing, and virtually never anything with a GUI/Qt, so not really sure...

As an aside, I couldn't run the test suite myself since (whether I ran it on the various base branches or on my test branch, directly through `py.test`, or `runtests.py`), the first test would run, a spyder window would open with various actions performed, and then it would stop and get stuck either doing nothing (on, py.test), or on a "Save/Don't Save/Cancel" dialog, and it would either timeout (in the first case), or fail (no matter what I pressed on the dialog, or if I waited several minutes)...not sure what's going on there.

Thanks!

### DIFF
--- a/spyder/widgets/variableexplorer/collectionseditor.py
+++ b/spyder/widgets/variableexplorer/collectionseditor.py
@@ -673,7 +673,7 @@ class BaseTableView(QTableView):
                                           triggered=self.paste)
         self.copy_action = create_action(self, _("Copy"),
                                          icon=ima.icon('editcopy'),
-                                         triggered=self.copy)                                      
+                                         triggered=self.copy)
         self.edit_action = create_action(self, _("Edit"),
                                          icon=ima.icon('edit'),
                                          triggered=self.edit_item)
--- a/spyder/widgets/variableexplorer/collectionseditor.py
+++ b/spyder/widgets/variableexplorer/collectionseditor.py
@@ -476,28 +476,36 @@ class CollectionsDelegate(QItemDelegate):
             return None
         #---editor = QDateTimeEdit
         elif isinstance(value, datetime.datetime):
-            editor = QDateTimeEdit(value, parent)
-            editor.setCalendarPopup(True)
-            editor.setFont(get_font(font_size_delta=DEFAULT_SMALL_DELTA))
-            return editor
+            if readonly:
+                return None
+            else:
+                editor = QDateTimeEdit(value, parent)
+                editor.setCalendarPopup(True)
+                editor.setFont(get_font(font_size_delta=DEFAULT_SMALL_DELTA))
+                return editor
         #---editor = QDateEdit
         elif isinstance(value, datetime.date):
-            editor = QDateEdit(value, parent)
-            editor.setCalendarPopup(True)
-            editor.setFont(get_font(font_size_delta=DEFAULT_SMALL_DELTA))
-            return editor
+            if readonly:
+                return None
+            else:
+                editor = QDateEdit(value, parent)
+                editor.setCalendarPopup(True)
+                editor.setFont(get_font(font_size_delta=DEFAULT_SMALL_DELTA))
+                return editor
         #---editor = TextEditor
         elif is_text_string(value) and len(value) > 40:
             te = TextEditor(None)
             if te.setup_and_check(value):
-                editor = TextEditor(value, key)
+                editor = TextEditor(value, key, readonly=readonly)
                 self.create_dialog(editor, dict(model=index.model(),
                                                 editor=editor, key=key,
                                                 readonly=readonly))
             return None
         #---editor = QLineEdit
         elif is_editable_type(value):
-            if not readonly:
+            if readonly:
+                return None
+            else:
                 editor = QLineEdit(parent)
                 editor.setFont(get_font(font_size_delta=DEFAULT_SMALL_DELTA))
                 editor.setAlignment(Qt.AlignLeft)
--- a/spyder/widgets/variableexplorer/collectionseditor.py
+++ b/spyder/widgets/variableexplorer/collectionseditor.py
@@ -474,21 +474,15 @@ class CollectionsDelegate(QItemDelegate):
             self.create_dialog(editor, dict(model=index.model(), editor=editor,
                                             key=key, readonly=readonly))
             return None
-        #---editor = QDateTimeEdit
-        elif isinstance(value, datetime.datetime):
-            if readonly:
-                return None
-            else:
-                editor = QDateTimeEdit(value, parent)
-                editor.setCalendarPopup(True)
-                editor.setFont(get_font(font_size_delta=DEFAULT_SMALL_DELTA))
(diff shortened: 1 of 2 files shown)


CAM-Gerlach:
Looking more into it, this naive approach doesn't account for dates/datetimes and long text strings that are handled separately. Accordingly, I'll amend this commit to deal with those cases to. I propose having dates and times not open like normal text, and I'll see if I can get long text to open to a textbox for easier/more complete viewing, but not actually be editable which can presumably be done through Qt methods or the like.

With regard to the former, however, is that currently Spyder does not display datetime objects in a "pretty" format when viewing inline in the variable explorer, only in "edit" mode, so not sure if the best approach is to do the above and then later go back and improve the display of dates when not selected (probably not a bad idea anyway at some point), or to enable edit mode as now but just not allow the user to change values, if that is easily implementable.

CAM-Gerlach:
For long text strings, the fix was quite simple—just pass the readonly argument through to the TextEditor constructor, and bam, read-only text editor to allow full viewing, just not editing.

For dates/datetimes, currently I just have them do nothing when edited as I couldn't find an easy way to make the resultant QDate(Time)Edit object, well, uneditable, and aside from looking prettier there appears to be little point. However, I can look into making the static representation of dates and times more pretty/human readable like in the QDate(Time)Editor rather than just the raw format, either in this or a separate PR.

CAM-Gerlach:
Seeing that the date and datetime editor branches are (now) entirely duplicate code except for one line, I refactored it to unify them. Good idea? Bad idea? I can rollback if you disagree, but I figured I'd put it out there for your feedback.

Also, can the pep8speaks warning be ignored, given that I didn't change the format of the comment line and all the others use the same format?

spyder/widgets/variableexplorer/collectionseditor.py
```diff
@@ -474,38 +474,41 @@ def createEditor(self, parent, option, index):
...
+        #---editor = QDateEdit or QDateTimeEdit
         elif isinstance(value, datetime.date):
-            editor = QDateEdit(value, parent)
-            editor.setCalendarPopup(True)
-            editor.setFont(get_font(font_size_delta=DEFAULT_SMALL_DELTA))
-            return editor
+            if readonly:
+                return None
+            else:
+                if isinstance(value, datetime.datetime):
```
ccordoba12: Is `datetime.datetime` an instance of `datetime.date`?

ccordoba12:
> For dates/datetimes, currently I just have them do nothing when edited as I couldn't find an easy way to make the resultant QDate(Time)Edit object, well, uneditable

I think that's fine for now.

> However, I can look into making the static representation of dates and times more pretty/human readable like in the QDate(Time)Editor rather than just the raw format, either in this or a separate PR.

Please try to do that in another PR. PRs should stay as simple and focused and possible so we can merge them quickly.

spyder/widgets/variableexplorer/collectionseditor.py
ccordoba12: If this is the case, then your refactoring is correct. Else, you have to revert it.

CAM-Gerlach: Yes indeed, and I tested to confirm.

CAM-Gerlach:
Just realized while checking the status of this PR that I must have accidentally edited your comment thinking it was mine, not being used to my new Member powers, x(

> Please try to do that in another PR. PRs should stay as simple and focused and possible so we can merge them quickly.

Yes indeed; already working on it. Should be simple enough, but well out of scope of this one.

CAM-Gerlach:
I assume I'll want to add a test to ensure directly editable values (and perhaps such values inside of sub-objects, lists, etc) really are treated as uneditable when part of an immutable object (e.g. a tuple)?

ccordoba12:
@CAM-Gerlach, it seems a really hard test to create. I think what you've done so far is really good for now.

CAM-Gerlach:
Okay. A simple way to unit test my proximate change would be just `assert`ing that, when `readonly=True`, the `createEditor` method of `CollectionsDelegate` returns `None` for directly editable values (and properly returns an editor when `readonly=False`, and for collections types). Of course, that wouldn't catch everything that could go wrong, but it would ensure there isn't a regression on the current change....but if you think it isn't worth it, then its up to you of course and feel free to merge as is.

ccordoba12:
Nice! I think it's a good compromise, so please go ahead with it.

CAM-Gerlach:
Sure, will do...

ccordoba12:
Is this one ready?

CAM-Gerlach:
Yup, if you think so.

goanpeca:
Awesome work @CAM-Gerlach  :-)

CAM-Gerlach:
^_^ Sorry I got a tad frustrated; really, the only tough part where I got stuck was knowing (how) to monkeypatch the relative rather than absolute paths to the modules. After I took a break to get some sleep I figured that had to be it...I just should have taken a step back earlier instead of getting too sucked in, heh.

PR #6042, opened 2017-12-23
PR: Add test coverage reporting with AppVeyor to Codecov

jnsebgosselin:
- [x] setup codecov https://github.com/codecov/example-python
- [x] setup the `.codecov.yml`

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,26 @@
+codecov:
+  notify:
+    require_ci_to_pass: yes
+
+codecov:
+  branch: master
+
+coverage:
+  precision: 2
+  round: down
+  range: "70...100"
+  
+  status:
+    project: yes
+    patch: no
+    changes: no
+    
+parsers:
+  gcov:
+    branch_detection:
+      conditional: yes
+      loop: yes
+      method: no
+      macro: no
+   
+comment: false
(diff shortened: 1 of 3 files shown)


jnsebgosselin:
@ccordoba12 Do you want to add a badge for codecov on the README?

.codecov.yml
```diff
@@ -0,0 +1,26 @@
+codecov:
+  notify:
+    require_ci_to_pass: yes
+
+codecov:
+  branch: master
```
ccordoba12: We should run this in 3.x too.

ccordoba12:
> Do you want to add a badge for codecov on the README?

Ok, no problem.

.codecov.yml
ccordoba12: So please move this PR to the 3.x branch.

jnsebgosselin: > We should run this in 3.x too

From my understanding, the `branch` option in the .yaml is used to specify which branch is displayed by default on the codecov project homepage. From [codecov documentation](https://docs.codecov.io/v4.3.6/docs/codecov-yaml#section-default-branch):

"This will change the branch used on the default project dashboard"

I think that master is used by default, so I've only kept this option in the .yaml file in case we want to change it later. Do you prefer to have the 3.x instead?

> So please move this PR to the 3.x branch.

You are right, I will rebase this PR to the 3.x branch.

jnsebgosselin:
Thank you very much for the detailed description @CAM-Gerlach . Just to be sure I understood correctly, you are skipping some of the tests because they fail locally on your machine, even if they pass on our CI?

CAM-Gerlach:
@jnsebgosselin I think you mean to reply on #6044 :P

jnsebgosselin:
@CAM-Gerlach I was wondering where my answer went lol !

PR #6057, opened 2017-12-26
PR: Add coverage files and Notepad++ backups to gitignore

CAM-Gerlach:
Just as it says on the tin, a rather trivial pull request to add all code coverage files (`coverage.MACHINENAME.####.####`) to the `.gitignore` list, along with Notepad++ auto-backup directories, the former per consultation with @ccordoba12  and the latter mostly for convenience (if not desired, I'm fine with omitting it).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@
 # gedit files
 *~
 
+# Notepad++ files
+nppBackup/
+
 # Special dirs and files
 build/
 dist/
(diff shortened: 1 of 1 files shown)


CAM-Gerlach:
For some reason, the cancelled AppVeyor build for the closed accidental duplicate PR (against a different branch, with a different name, tags and the like) got linked to this PR instead of the actual AppVeyor build for this one. However, once the real one passed, looks like it fixed itself.

PR #6081, opened 2017-12-29
PR: Fix error when attempting to view module contents in CollectionsEditor

### LINKED ISSUE #6080: Variable Explorer throws an error instead of displaying contents when attempting to inspect any loaded modules
## Description

I double-clicked a module which was the value of a `dict` in variable explorer and got an error.

1. Import a module (example: `import os`)
2. Initialize a `dict` in a variable (example: `d=dict()`)
3. Assign the module to the dict (example: `d['os']=os`)
4. Double-click the cyan entry of your variable on variable explorer
5. Double-click the white entry which corresponds to your module
6. You see the error reporting window.

I expected seeing the global variables defined in the module, but I got an error reporting window.

#### Stack trace
```
  File "/usr/local/lib/python3.6/dist-packages/spyder/widgets/variableexplorer/collectionseditor.py", line 513, in createEditor
    readonly=readonly)
  File "/usr/local/lib/python3.6/dist-packages/spyder/widgets/variableexplorer/collectionseditor.py", line 1315, in setup
    self.data_copy = copy.deepcopy(data)
  File "/usr/lib/python3.6/copy.py", line 169, in deepcopy
    rv = reductor(4)
TypeError: can't pickle module objects
```

#### Screenshot
screenshot_2017-12-28_16-23-15

* Ubuntu Studio 17.10
* Spyder Version: 3.2.5 
* Python Version: 3.6.3
* Qt Versions: 5.9.3, PyQt5 5.9.2 on Linux

## Dependencies
```
sphinx >=0.6.6   :  1.6.5 (OK)
psutil >=0.3     :  5.4.2 (OK)
pylint >=0.25    :  1.8.1 (OK)

```

CAM-Gerlach:
Fixes #6080 , where Spyder would produce an error instead of viewing module contents, likely related to the recent introduction of `cloudpickle`.

Also includes a regression test making sure the editor is created without error and is properly readonly.

### DIFF
--- a/spyder/widgets/variableexplorer/tests/test_collectioneditor.py
+++ b/spyder/widgets/variableexplorer/tests/test_collectioneditor.py
@@ -8,6 +8,7 @@ Tests for collectionseditor.py
 """
 
 # Standard library imports
+import os  # Example module for testing display inside collecitoneditor
 import copy
 import datetime
 try:
@@ -21,7 +22,9 @@ import pytest
 
 # Local imports
 from spyder.widgets.variableexplorer.collectionseditor import (
-    CollectionsEditorTableView, CollectionsModel, LARGE_NROWS, ROWS_TO_LOAD)
+    CollectionsEditorTableView, CollectionsModel, CollectionsEditor,
+    LARGE_NROWS, ROWS_TO_LOAD)
+
 
 # Helper functions
 def data(cm, i, j):
@@ -163,8 +166,9 @@ def test_rename_and_duplicate_item_in_collection_editor():
 
 
 def test_edit_mutable_and_immutable_types(qtbot, monkeypatch):
-    # To ensure mutable types (lists, dicts) and individual values are editable
-    # But not immutable ones (tuples) or anything inside of them, per #5991
+    """Check to ensure mutable types (lists, dicts) and individual values are
+    editable, but not immutable ones (tuples) or anything inside of them,
+    per #5991"""
     MockQLineEdit = Mock()
     attr_to_patch_qlineedit = ('spyder.widgets.variableexplorer.' +
                                'collectionseditor.QLineEdit')
@@ -263,6 +267,15 @@ def test_edit_mutable_and_immutable_types(qtbot, monkeypatch):
     mockCollectionsEditor_instance.setup.assert_called_with(ANY, ANY,
                                                             icon=ANY,
                                                             readonly=True)
+
+
+def test_view_module_in_coledit(qtbot, monkeypatch):
+    """Checks that modules don't produce an error when trying to open them in
+    Variable Explorer, and are set as readonly. Regression test for #6080"""
+
+    editor = CollectionsEditor()
+    editor.setup(os, "module_test", readonly=False)
+    assert editor.widget.editor.readonly
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


spyder/widgets/variableexplorer/tests/test_collectioneditor.py
```diff
@@ -265,5 +269,14 @@ def test_edit_mutable_and_immutable_types(qtbot, monkeypatch):
                                                             readonly=True)
 
 
+def test_view_module_in_coledit(qtbot, monkeypatch):
+    """Checks that modules don't produce an error when trying to open them in
```
ccordoba12: Please change `Checks` to `Check`

ccordoba12:
@CAM-Gerlach, could you post a screenshot to see how it looks to view a module in the Variable Explorer now?

spyder/widgets/variableexplorer/tests/test_collectioneditor.py
CAM-Gerlach: Oops; fixed the one above but missed that one. Thanks for the catch; corrected now.

CAM-Gerlach:
Sure, here ya go, though it wasn't really at all my doing. Is it SOP to post screenshots of any GUI visible changes from a PR?

image

CAM-Gerlach:
Not sure what was up with the one failed Travis CI check. My test passed locally in several trials, and did the same for every commit (all three should be functionally identical), every build on every platform, except for the one Py3.5, PyQt4 build on Travis that failed, due to my test causing an exception to occur. However, I told it to rerun that build and, of course, it passed again this time. Any idea how such a simple test could possibly be randomly unreliable in rare cases like that? 

It seems the one time it failed, it was because the exception that it now catches wasn't thrown as it was supposed to be inside the deepcopy function, and so it proceeded until it hit a different exception and then errored out. [Here](https://travis-ci.org/spyder-ide/spyder/jobs/322780469)'s the build output if that's helpful. Seems the problem might be with deepcopy not always throwing the error consistent? If so, how could something relatively simple like that be inconsistent—where's the entropy coming from?

ccordoba12:
> Any idea how such a simple test could possibly be randomly unreliable in rare cases like that?

Nop, sorry. Sometimes tests fail for no good reason. So please add the `flaky` decorator to your test to avoid rerunning our entire suite if it fails.

CAM-Gerlach:
Did so, but now seeing the same failure on a different build, and it seems it either passes all three flaky tests or fails all three, so it seems something further needs to be done. As it _seems_ the unreliability is in `deepcopy` itself detecting the module's unpickleability and raising the correct exception, I'll just have it catch the AttributeError it is throwing as well and handle it the same way as the TypeError, which should fix the root problem. That should be safe, right? (Hopefully not famous last words, heh...)

CAM-Gerlach:
Also, out of curiosity, is there a reason `import copy` is done inside that function and not at the top of the file like usual?

ccordoba12:
> Also, out of curiosity, is there a reason import copy is done inside that function and not at the top of the file like usual?

That's Pierre's work, so no idea.

CAM-Gerlach:
It couldn't have anything to do with the issue, I wouldn't think?

Other than that, is this good to go?

Issue #5426, opened 2017-10-09
Segmentation fault when trying to change keyboard shortcuts in Preferences


Griffintaur:
## Description of your problem
There is a segmentation dump issue and spyder is getting crashed whenever I tried to change the default key combination for comment in Spyder. [Ubuntu]
**What steps will reproduce the problem?**
1. Launch and go to preferences
2. Click on the keyboard shortcuts and try to change it.
3. Segmentation dump issue in Ubuntu

**What is the expected output? What do you see instead?**
No crash. changes should be successful.

**Please provide any additional information below**

## Versions and main components

* Spyder Version: 3.2.3
* Python Version: 3.6.2
* Qt Version:
* PyQt Version:
* Operating system: Ububtu

## Dependencies

Please go to the menu entry `Help > Optional Dependencies` (or
`Help > Dependencies`), press the button `Copy to clipboard`
and paste the contents below:

ccordoba12:
What are your Qt and PyQt versions?

ccordoba12:
Also, please run in a terminal

    spyder --reset

and try again

csabella:
I'm using Ubuntu 16.04 as well and I get the segmentation fault if I open any keyboard shortcut window, even if I click the Cancel button without making any changes.

ccordoba12:
As I said before, what are your Qt and PyQt versions?

csabella:
Sorry, I wasn't trying to report a bug, just trying to help provide debugging information.  But, here are my versions:
[Python 3.6.3 64bits, Qt 5.6.2, PyQt5 5.6 on Linux]

ccordoba12:
Thanks @csabella!

ccordoba12:
@csabella, is this still an issue for you?

csabella:
@ccordoba12   It still happens on 3.x, but does not happen on master.

ccordoba12:
I can't reproduce this in Kubuntu 17.04.

csabella:
On the last update, I said it wasn't happening on master, but now it's happening on master for me, as well as 3.x and the release version of 3.2.5.

I open Preferences, then Keyboard Shortcuts, double click on 'Find Next/F3', press F4 for the new shortcut, and click OK.  It closes Spyder and gives a seg fault.  I looked at this a little bit when it was first reported, but I'm going to spend some more time on it now to try to track it down.

ccordoba12:
Thanks Cheryl! I don't think this is going to be easy to fix...

[...]

PR #6092, opened 2017-12-29
PR: Fix segfault on shortcut preferences when using mouse

csabella:
For issue #5426.

To recreate the issue (at least on Ubuntu):
- Open Preferences and click on Keyboard Shortcuts.  
- Select a shortcut to edit, either by using Enter on the keyboard or by double clicking the line (meaning that mouse vs. keyboard isn't important at this step). 
- Without this change, if you immediately click Cancel, the seg fault occurs.  
- However, if instead of clicking, you use the keyboard 'Tab' to highlight cancel (as the text in the dialog suggests to do to switch focus), then you can press Enter on the Cancel button and not get the seg fault. 
- The same thing happens with the OK button.

This seemed to be a difference between keyboard entry and mouse clicking on the buttons.   In the code, the only difference was setting widget focus, so by adding focus to the mouse click, it prevents the seg fault.

I don't know the underlying cause, so this may be a hack in order to fix it, but it does work.

### DIFF
--- a/spyder/plugins/shortcuts.py
+++ b/spyder/plugins/shortcuts.py
@@ -16,7 +16,7 @@ import sys
 from qtpy import PYQT5
 from qtpy.compat import from_qvariant, to_qvariant
 from qtpy.QtCore import (QAbstractTableModel, QModelIndex, QRegExp,
-                         QSortFilterProxyModel, Qt)
+                         QSortFilterProxyModel, Qt, Slot)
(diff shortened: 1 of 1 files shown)


ccordoba12:
Thanks a lot for this @csabella! Could you move this PR to `3.x` since you said this bug is also present there? Thanks!

csabella:
I had done it under 3.x, but forgot to change it when creating the PR.  Thanks for catching it and sorry about that!

PR #6094, opened 2017-12-30
PR: Make Tab key properly switch between find and replace fields

### LINKED ISSUE #3674: Tab in-between find and replace text boxes
The tab key doesn't move the cursor from the find to replace text boxes as expected.

1. Hit control-F to open find text box.  Type in some text.
2. Hit control-R to open replace text box.  Type in some other text.
3. Press shift-tab to move from replace box back into text box.
4. Press tab to move from text box back forward into replace box. (This step doesn't do anything)

The tab to move from the text to replace box doesn't work.  You stay within the text box.  Pressing shift-tab to move backwards works as expected.

* Spyder Version: 3.1.0
* Python Version: 3.5.2
* Operating system: Win 7

I remember this working fine a while ago, but I've since noticed this on some earlier versions of Spyder.  It's finally started to bother me enough to open the issue.  Thanks for any help!

## Dependencies

pandas >=0.13.1  :  0.19.1 (OK)
sphinx >=0.6.6   :  1.4.8 (OK)
sympy >=0.7.3    :  1.0 (OK)

CAM-Gerlach:
Fixes #3674 .

After way too many hours trying to trace exactly _why_ tab (and not shift-tab) was suppressed on those fields, and only those fields, I eventually gave up and just stuck in a crude little hack that _seems_ to do the job. However, I'm not sure I'm at all happy with it. Thoughts? Better approaches? Side affects I'm missing? My lack of GUI/Qt experience is really showing here, but I certainly learned quite a bit.

Also, there is a probably not too meaningful in everyday use, but not entirely trivial difference in behavior between when just the find vs. when the find and replace boxes are visible. For the latter, tab only goes from the find box to the replace box, and no further; for the former, successive tab depresses takes one from the find box to the buttons to its right (versus before, all of those buttons were inaccessible via tab). On one hand, this behavior adds (keyboard) functionality over e.g. disabling tab when only the find box is active, as before. However, at least from one viewpoint it isn't quite as strictly consistent, though personally I don't really think it matters appreciably. 

I fiddled with additional `setTabOrder` calls to try to get it to either cycle back and forth between the two boxes, or tab further to the replace/replace sel/replace all buttons for easier access, but neither of those possibilities worked besides the first `self.search_text.setTabOrder(self.search_text, self.replace_text)`, and in some cases prevented that from working entirely, so I just dropped it for now to get this out the door.

On the other hand, I (eventually) did get my test to work out; it actually checks both that the right function gets called behind the scenes when Tab is pressed in the search box with the replace box up using Mock objects, in classic unit test fashion, and also tests it from a more "real world" angle by actually testing that the focus was moved properly in the GUI.

### DIFF
--- a/spyder/widgets/findreplace.py
+++ b/spyder/widgets/findreplace.py
@@ -176,14 +176,14 @@ class FindReplace(QWidget):
         self.highlight_timer.timeout.connect(self.highlight_matches)
         self.search_text.installEventFilter(self)
 
-
     def eventFilter(self, widget, event):
         """Event filter for search_text widget.
 
         Emits signals when presing Enter and Shift+Enter.
         This signals are used for search forward and backward.
+        Also, a crude hack to get tab working in the Find/Replace boxes.
         """
-        if (event.type() == QEvent.KeyPress):
+        if event.type() == QEvent.KeyPress:
             key = event.key()
             shift = event.modifiers() & Qt.ShiftModifier
 
@@ -193,8 +193,10 @@ class FindReplace(QWidget):
                 else:
                     self.return_pressed.emit()
 
-        return super(FindReplace, self).eventFilter(widget, event)
+            if key == Qt.Key_Tab:
+                self.focusNextChild()
 
+        return super(FindReplace, self).eventFilter(widget, event)
 
     def create_shortcuts(self, parent):
         """Create shortcuts for this widget"""
--- a/spyder/widgets/tests/test_editor.py
+++ b/spyder/widgets/tests/test_editor.py
@@ -17,6 +17,7 @@ except ImportError:
 
 # Third party imports
 import pytest
+from flaky import flaky
 from qtpy.QtCore import Qt
 from qtpy.QtGui import QTextCursor
 
@@ -483,6 +484,7 @@ def test_tab_keypress_properly_caught_find_replace(editor_find_replace_bot):
     finder.focusNextChild.assert_called_once_with()
 
 
+@flaky(max_runs=3)
 @pytest.mark.skipif(platform.startswith('linux'),
                     reason="This test fails on Linux, for unknown reasons.")
 def test_tab_moves_focus_from_search_to_replace(editor_find_replace_bot):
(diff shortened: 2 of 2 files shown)


CAM-Gerlach:
Looks like it failed again, same reason. `finder.search_text.hasFocus()` is returning false on Linux even though I explicitly set the focus there the line before. Can anyone on Linux right now give me an idea what's going on? My old box is offline right now. 

If necessary, since I test it by two different methods, I can just split them into two tests and have the "real world" one that's failing skip on Linux, assuming the other one passes.

ccordoba12:
> Can anyone on Linux right now give me an idea what's going on?

No idea, sorry.

> I can just split them into two tests and have the "real world" one that's failing skip on Linux

It sounds like a good plan. Please go ahead with it.

CAM-Gerlach:
@ccordoba12 'Tis done.

jitseniesen:
> `finder.search_text.hasFocus()` is returning false on Linux even though I explicitly set the focus there with `finder.search_text.setFocus()` the line before, and it should start there by default anyway.

@CAM-Gerlach I don't really know, but this page might be relevant: http://pytest-qt.readthedocs.io/en/latest/wait_until.html See in particular the sentence:

> The window.edit.setFocus() may not be processed immediately, only in a future event loop, which might lead to this test to work sometimes and fail in others (a flaky test).

CAM-Gerlach:
Thanks @jitseniesen , not sure how I missed that. That would probably explain a lot of the variability I've been seeing with tests in general...so its sorta like a race condition, except between one statement and the next, due to the asynchronous nature of certain calls? Most of that (I did see one random failure in testing likely due to that) should be addressed already with my most recent previous commit, adding `flaky` and `100 ms` waits between the key actions and their asserts...as long as you think that's enough? 

However, I'm not sure this explains the problem with Linux; every single build failed on every attempt, at the very first ("control") assert statement (which should have already been true when the test fixtures were setup, as focus is supposed to start on the search box with no other intervention). Nevertheless, its worth a shot I'm sure, so I'll shoot that up now. Thanks!

CAM-Gerlach:
Nope, didn't work sorry. Oh well...anything else I should try, or that you need before you merge this? Thanks.

CAM-Gerlach:
Nope, looks like it indeed didn't work :( . Anything else I should try, or have any feedback on my rather hacky approach of enabling tab in this dialog?

jitseniesen:
No, I don't know then. ☹️

PR #6095, opened 2017-12-30
PR: Fixes bugs (3) and adds improvements (4) to the find in files functionality.

### LINKED ISSUE #5306: Paths in the "Search in" combobox of the "Find in File" plugin becomes corrupted

Related to PR #4822 that fixed Issues #4814 and #4809.

1. I don't know... I'm sorry. I'll add more info in the future if I can. I'm using this feature a lot, so I'll keep my eyes opened!

Sometimes, _Other directories_ in the  _Search in_ combobox of the _Find in File_ plugin becomes corrupted with no way of removing them. For instance, the `<PyQt5.QtWidgets.QListWidget...` shown in the screenshot below was actually `C:\Users\jsgosselin\spyder` before it became corrupted at some point.  This is the third time that this problem occurs for me. 

Besides, I think an option should be added below _Select other directory_ to clear the list.

image

* Spyder Version: branch 3.x and master
* Python Version: 3.6.2
* Qt Version: 5.9.1
* PyQt Version: 5.9
* Operating system: Windows 7 64 bits

## Dependencies

jnsebgosselin:
Fixes #5306 

- [x] Fix **issue** reported in https://github.com/spyder-ide/spyder/issues/5306#issuecomment-354492631
- [x] Fix **issue** reported in https://github.com/spyder-ide/spyder/issues/5306#issuecomment-354493874 
- [x] Fix **issue** reported in https://github.com/spyder-ide/spyder/issues/5306#issuecomment-354506402

- [x] When populating the `Search in` QComboBox, do not add the paths that do not exists anymore.
- [ ] When hovering over an item of the QComboBox and pressing <kbd>Del</kbd>, remove the item from the list.
- [ ] Add a  `Clear this list` option below `Select other directory`.
- [x] Show the full path when hovering over an item in the `Search in` QComboBox.

- [ ] Expand the **test** for the FindInFiles Widget and Plugin.

### DIFF
--- a/spyder/widgets/findinfiles.py
+++ b/spyder/widgets/findinfiles.py
@@ -25,10 +25,10 @@ import traceback
 from qtpy.compat import getexistingdirectory
 from qtpy.QtGui import QAbstractTextDocumentLayout, QTextDocument
 from qtpy.QtCore import QMutex, QMutexLocker, Qt, QThread, Signal, Slot, QSize
-from qtpy.QtWidgets import (QHBoxLayout, QLabel, QListWidget, QSizePolicy,
+from qtpy.QtWidgets import (QHBoxLayout, QLabel, QComboBox, QSizePolicy,
                             QTreeWidgetItem, QVBoxLayout, QWidget,
                             QStyledItemDelegate, QStyleOptionViewItem,
-                            QApplication, QStyle, QListWidgetItem)
+                            QApplication, QStyle)
 
 # Local imports
 from spyder.config.base import _
@@ -203,21 +203,146 @@ class SearchThread(QThread):
         return self.results, self.pathlist, self.total_matches, self.error_flag
 
 
-class ExternalPathItem(QListWidgetItem):
-    def __init__(self, parent, path):
-        self.path = path
-        QListWidgetItem.__init__(self, self.__repr__(), parent)
+class SearchInComboBox(QComboBox):
+    """
+    Non editable combo box handling the path locations of the FindOptions
+    widget.
+    """
+    def __init__(self, external_path_history=[], parent=None):
+        super(SearchInComboBox, self).__init__(parent)
+        self.setSizePolicy(QSizePolicy.Expanding, QSizePolicy.Fixed)
+        self.setToolTip(_('Search directory'))
+        self.setEditable(False)
 
-    def __repr__(self):
-        if len(self.path) > MAX_PATH_LENGTH:
-            return truncate_path(self.path)
-        return self.path
+        self.path = ''
+        self.project_path = None
+        self.file_path = None
+        self.external_path = None
 
-    def __str__(self):
-        return self.__repr__()
+        self.addItem(_("Current working directory"))
+        ttip = ("Search in all files and directories present on the current"
+                " Spyder path")
+        self.setItemData(0, ttip, Qt.ToolTipRole)
 
-    def __unicode__(self):
-        return self.__repr__()
+        self.addItem(_("Project"))
+        ttip = _("Search in all files and directories present on the"
+                 " current project path (if opened)")
+        self.setItemData(1, ttip, Qt.ToolTipRole)
+        self.model().item(1, 0).setEnabled(False)
+
+        self.addItem(_("File").replace('&', ''))
+        ttip = _("Search in current opened file")
+        self.setItemData(2, ttip, Qt.ToolTipRole)
+
+        self.addItem(_("Select other directory"))
+        ttip = _("Search in other folder present on the file system")
+        self.setItemData(3, ttip, Qt.ToolTipRole)
+
+        self.insertSeparator(3)
+        self.insertSeparator(5)
+
+        for path in external_path_history:
+            self.add_external_path(path)
+
+        self.currentIndexChanged.connect(self.path_selection_changed)
+
+    def add_external_path(self, path):
+        """
+        Adds an external path to the combobox if it exists on the file system.
+        If the path is already listed in the combobox, it is removed from its
+        current position and added back at the end. If the maximum number of
+        paths is reached, the oldest external path is removed from the list.
+        """
+        if not os.path.exists(path):
+            return
+        self.removeItem(self.findText(path))
+        self.addItem(path)
+        self.setItemData(self.count()-1, path, Qt.ToolTipRole)
+        while self.count() > MAX_PATH_HISTORY+6:
+            self.removeItem(6)
+
+    def get_external_paths(self):
+        """Returns a list of the external paths listed in the combobox."""
+        return [to_text_string(self.itemText(i))
+                for i in range(6, self.count())]
+
+    def get_current_searchpath(self):
+        """
+        Returns the path corresponding to the currently selected item
+        in the combobox.
+        """
+        idx = self.currentIndex()
+        if idx == CWD:
+            return self.path
+        elif idx == PROJECT:
+            return self.project_path
+        elif idx == FILE_PATH:
+            return self.file_path
+        else:
+            return self.external_path
+
+    def is_file_search(self):
+        """Returns whether the current search path is a file."""
+        if self.currentIndex() == FILE_PATH:
+            return True
+        else:
+            return False
+
+    @Slot()
+    def path_selection_changed(self):
+        """Handles when the current index of the combobox changes."""
+        idx = self.currentIndex()
+        if idx == EXTERNAL_PATH:
+            external_path = self.select_directory()
+            if len(external_path) > 0:
+                self.add_external_path(external_path)
+                self.setCurrentIndex(self.count()-1)
+            else:
+                self.setCurrentIndex(CWD)
+        elif idx > EXTERNAL_PATH:
+            self.external_path = to_text_string(self.itemText(idx))
+
+    @Slot()
+    def select_directory(self):
+        """Select directory"""
+        self.__redirect_stdio_emit(False)
+        directory = getexistingdirectory(
+                self, _("Select directory"), self.path)
+        if directory:
+            directory = to_text_string(osp.abspath(to_text_string(directory)))
+        self.__redirect_stdio_emit(True)
+        return directory
+
+    def set_project_path(self, path):
+        """
+        Sets the project path and disables the project search in the combobox
+        if the value of path is None.
+        """
+        if path is None:
+            self.project_path = None
+            self.model().item(PROJECT, 0).setEnabled(False)
+            if self.currentIndex() == PROJECT:
+                self.setCurrentIndex(CWD)
+        else:
+            path = to_text_string(osp.abspath(to_text_string(path)))
+            self.project_path = path
+            self.model().item(PROJECT, 0).setEnabled(True)
+
+    def __redirect_stdio_emit(self, value):
+        """
+        Searches through the parent tree to see if it is possible to emit the
+        redirect_stdio signal.
+        This logic allows to test the SearchInComboBox select_directory method
+        outside of the FindInFiles plugin.
+        """
+        parent = self.parent()
+        while parent is not None:
+            try:
+                parent.redirect_stdio.emit(value)
+            except AttributeError:
+                parent = parent.parent()
+            else:
+                break
 
 
 class FindOptions(QWidget):
@@ -234,12 +359,6 @@ class FindOptions(QWidget):
 
         if search_path is None:
             search_path = getcwd_or_home()
-
-        self.path = ''
-        self.project_path = None
-        self.file_path = None
-        self.external_path = None
-        self.external_path_history = external_path_history
 
         if not isinstance(search_text, (list, tuple)):
             search_text = [search_text]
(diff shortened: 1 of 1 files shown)


jnsebgosselin:
### Mouse Over and Delete key press demo:
fif_demo1

### Select other directory and Clear this list demo:
fif_demo2

ccordoba12:
Jean, could you also fix codecov to show failures only when coverage decreases in 1%? Thanks!

jnsebgosselin:
@ccordoba12 While working on the tests for the [FindInFiles](https://github.com/spyder-ide/spyder/blob/v3.2.5/spyder/plugins/findinfiles.py#L31) plugin, I found out an additional bug. 

The [get_options](https://github.com/spyder-ide/spyder/blob/v3.2.5/spyder/widgets/findinfiles.py#L417) method of the [FindOptions](https://github.com/spyder-ide/spyder/blob/v3.2.5/spyder/widgets/findinfiles.py#L223) widget is used for two things:
- It is used to get the options when [find](https://github.com/spyder-ide/spyder/blob/v3.2.5/spyder/widgets/findinfiles.py#L863) is run. In this case, `all=False` and the options are defined [here](https://github.com/spyder-ide/spyder/blob/v3.2.5/spyder/widgets/findinfiles.py#L487).
- It is also used in the [closing_plugin](https://github.com/spyder-ide/spyder/blob/v3.2.5/spyder/plugins/findinfiles.py#L157) method of the `FindInFiles` plugin for saving the options to the config file. In this case, `all=True` and the options are defined [here](https://github.com/spyder-ide/spyder/blob/v3.2.5/spyder/widgets/findinfiles.py#L473).

The problem is that there is safeguards within the `get_options` method to prevent the `find` method to be run if one of the parameters is not correct. However, these safeguards also prevent the find in files options to be saved to the config file when Spyder is closed.

For example, the options won't be saved to the config file if the `Search pattern` line edit is empty:
1. Go in the `FindInFiles` plugin and add an external path to the `Search in` combobox.
2. Leave the `Search pattern` line edit empty and close Spyder.
3. Restart Spyder: the external path that was added in 1 does not appear in the list of the  `Search in` combobox.  It was not saved to the config file due to the safeguard [here](https://github.com/spyder-ide/spyder/blob/v3.2.5/spyder/widgets/findinfiles.py#L423).

I think this should be fixed in another PR. I will built my tests around this issue for the moment.

jnsebgosselin:
@ccordoba12 I think this is ready to be reviewed. I'll go over it again tomorrow just to be sure. There is a lot of stuff to review in this, so feel free to push it back to 3.2.7 if you do not have the time for 3.2.6. I'll try to document a little better the changes I've made later in the description of this PR.

Also, I'm not sure I understand well the use of `to_text_string`. It was used everywhere in `findinfiles.py`, so I did the same in the changes I've made.

More than 350 lines of tests added for a meager 0.3% increase in coverage... damn.

spyder/widgets/findinfiles.py
```diff
@@ -24,11 +24,12 @@
 # Third party imports
 from qtpy.compat import getexistingdirectory
 from qtpy.QtGui import QAbstractTextDocumentLayout, QTextDocument
-from qtpy.QtCore import QMutex, QMutexLocker, Qt, QThread, Signal, Slot, QSize
-from qtpy.QtWidgets import (QHBoxLayout, QLabel, QListWidget, QSizePolicy,
+from qtpy.QtCore import (QMutex, QMutexLocker, Qt, QThread, Signal, Slot,
+                         QSize, QEvent)
```
ccordoba12: Please organize these imports in alphabetical order

spyder/widgets/findinfiles.py
```diff
@@ -203,21 +206,181 @@ def get_results(self):
...
+        self.view().installEventFilter(self)
+
+    def add_external_path(self, path):
+        """
+        Adds an external path to the combobox if it exists on the file system.
+        If the path is already listed in the combobox, it is removed from its
+        current position and added back at the end. If the maximum number of
+        paths is reached, the oldest external path is removed from the list.
+        """
+        if not os.path.exists(path):
```
ccordoba12: Change `os.path` to `osp` (that's our custom in Spyder).

spyder/widgets/findinfiles.py
```diff
@@ -203,21 +206,181 @@ def get_results(self):
...
+            return
+        self.removeItem(self.findText(path))
+        self.addItem(path)
+        self.setItemData(self.count()-1, path, Qt.ToolTipRole)
```
ccordoba12: Spaces around `-`

spyder/widgets/findinfiles.py
```diff
@@ -203,21 +206,181 @@ def get_results(self):
...
+        while self.count() > MAX_PATH_HISTORY+EXTERNAL_PATHS:
```
ccordoba12: Spaces around `+`

spyder/widgets/findinfiles.py
```diff
@@ -203,21 +206,181 @@ def get_results(self):
...
+            self.removeItem(EXTERNAL_PATHS)
+
+    def get_external_paths(self):
+        """Returns a list of the external paths listed in the combobox."""
+        return [to_text_string(self.itemText(i))
```
ccordoba12: We probably need to use here `to_unicode_from_fs` (from `utils/encoding.py`) to deal with non-ascii dirs on Windows that uses encodings like cp-1251 (Russian).

spyder/widgets/findinfiles.py
```diff
@@ -203,21 +206,181 @@ def get_results(self):
...
+
+    @Slot()
+    def path_selection_changed(self):
+        """Handles when the current index of the combobox changes."""
+        idx = self.currentIndex()
+        if idx == SELECT_OTHER:
+            external_path = self.select_directory()
+            if len(external_path) > 0:
+                self.add_external_path(external_path)
+                self.setCurrentIndex(self.count()-1)
```
ccordoba12: Spaces around `-`

spyder/widgets/findinfiles.py
ccordoba12: Sorry, this is not the proper place to do it. I'll point to the right place below.

spyder/widgets/findinfiles.py
```diff
@@ -203,21 +206,181 @@ def get_results(self):
...
+            self.external_path = to_text_string(self.itemText(idx))
+
+    @Slot()
+    def select_directory(self):
+        """Select directory"""
+        self.__redirect_stdio_emit(False)
+        directory = getexistingdirectory(
+                self, _("Select directory"), self.path)
+        if directory:
+            directory = to_text_string(osp.abspath(to_text_string(directory)))
```
ccordoba12: This is the right place to use `to_unicode_from_fs`. So this should be

    directory = encoding.to_unicode_from_fs(osp.abspath(directory))

spyder/widgets/findinfiles.py
```diff
@@ -203,21 +206,181 @@ def get_results(self):
...
+        Sets the project path and disables the project search in the combobox
+        if the value of path is None.
+        """
+        if path is None:
+            self.project_path = None
+            self.model().item(PROJECT, 0).setEnabled(False)
+            if self.currentIndex() == PROJECT:
+                self.setCurrentIndex(CWD)
+        else:
+            path = to_text_string(osp.abspath(to_text_string(path)))
```
ccordoba12: Change this to

    path = encoding.to_unicode_from_fs(osp.abspath(path))

Could you add some tests to verity that this is correct in Python 2?

spyder/widgets/findinfiles.py
```diff
@@ -203,21 +206,181 @@ def get_results(self):
...
+    def eventFilter(self, widget, event):
+        """Used to handle key events on the QListView of the combobox."""
+        if event.type() == QEvent.KeyPress and event.key() == Qt.Key_Delete:
+            index = self.view().currentIndex().row()
+            if index >= EXTERNAL_PATHS:
+                # Remove item and update the view.
+                self.removeItem(index)
+                self.showPopup()
+                # Set the view selection so that it doesn't bounce around.
+                new_index = min(self.count()-1, index)
```
ccordoba12: Spaces around `-`

spyder/widgets/findinfiles.py
ccordoba12: Mmmh, it seems we're not transforming to Unicode our project paths, so maybe this is for another PR.

ccordoba12:
I like a lot the new `SearchInComboBox` class you added. It makes the code more modular and easier to understand. And, of course, I also like the new tests you added!

I left some minor comments, except for the use of `to_unicode_from_fs` in one place. Then I'd say this is ready.

spyder/widgets/findinfiles.py
ccordoba12: I opened issue #6103 about this (which I'll fix for 3.2.6), so please change this line to

    path = osp.abspath(path)

jnsebgosselin:
Thank you very much Carlos for this quick review, this is very nice!

### To do:
- [ ] Assert non-ASCII path in the tests.
- [ ] Do the requested style changes.
- [ ] Fix the `to_unicode_from_fs` and `to_text_string`.

spyder/widgets/findinfiles.py
jnsebgosselin: Should I replace `to_text_string(self.itemText(i)` by `self.itemText(i)`?

I do not understand what `to_text_string` is doing here since `self.itemText(i)` is already returning a unicode str by default as stated, for instance, in the [PySide documentation](https://srinikom.github.io/pyside-docs/PySide/QtGui/QComboBox.html#PySide.QtGui.PySide.QtGui.QComboBox.itemText)). Or is it returning a `QString` in some version of PyQt on python 2?

All this encoding stuff is mumbo jumbo to me, no matter how many times I've read on this subject XD.

spyder/widgets/findinfiles.py
```diff
@@ -203,21 +206,181 @@ def get_results(self):
...
+        elif idx == CLEAR_LIST:
+            reply = QMessageBox.question(
+                    self, _("Clear other directories"),
+                    _("Do you want to clear the list of other directories?"),
+                    QMessageBox.Yes | QMessageBox.No)
+            if reply == QMessageBox.Yes:
+                self.clear_external_paths()
+            self.setCurrentIndex(0)
+        elif idx >= EXTERNAL_PATHS:
+            self.external_path = to_text_string(self.itemText(idx))
```
jnsebgosselin: Same comment here, should I replace `to_text_string(self.itemText(idx))` by `self.itemText(idx)`.

spyder/widgets/findinfiles.py
```diff
@@ -487,32 +574,30 @@ def get_options(self, all=False):
...
-        self.parent().redirect_stdio.emit(True)
-        return directory
+    @property
+    def path(self):
+        return self.path_selection_combo.path
 
     def set_directory(self, directory):
-        self.path = to_text_string(osp.abspath(to_text_string(directory)))
+        self.path_selection_combo.path = to_text_string(osp.abspath(
+                to_text_string(directory)))
```
jnsebgosselin: Should `to_text_string(osp.abspath(to_text_string(directory)))` be changed to `osp.abspath(directory)` or `to_unicode_from_fs(osp.abspath(path))`?

spyder/widgets/findinfiles.py
```diff
@@ -487,32 +574,30 @@ def get_options(self, all=False):
...
-        self.project_path = None
+        self.path_selection_combo.set_project_path(None)
+
+    @property
+    def file_path(self):
+        return self.path_selection_combo.file_path
 
     def set_file_path(self, path):
-        self.file_path = path
+        self.path_selection_combo.file_path = path
```
jnsebgosselin: Should `path` be changed for `osp.abspath(path)` or `to_unicode_from_fs(osp.abspath(path))`?

spyder/widgets/findinfiles.py
jnsebgosselin: Ok, I've seen that `to_text_string` was always used with `itemText` everywhere else. So maybe it is safer to keep it that way.

jnsebgosselin: OK, I've read that Python 2, PyQt API #1 is returning a `QString`, so I'll keep the `to_text_string`.

jnsebgosselin:
I added a non-ASCII path to the tests... and all hell breaks loose lol

The test is passing for Python3 in AppVeyor, but not for Python 2. I don't understand why the tests are failing on Travis though, I'm not sure it is related to this PR...

jnsebgosselin:
The "good news" is that I'm able to make the test crash locally with Python2, so I should be able to debug this... but probably not until next year XD

jnsebgosselin:
@ccordoba12 Just so you know, this is ready, but I would like you to check again carefully some of the places where we return a `path`. 

So good thinking in pushing it back to 3.2.7. It wouldn't be a bad idea if you could play a little bit with the changes I've made locally also.

spyder/widgets/findinfiles.py
ccordoba12: Ok, good idea.

spyder/widgets/findinfiles.py
ccordoba12: I see this is only used in one place:

https://github.com/jnsebgosselin/spyder/blob/8b9483d72967c4580143fde6adaeec38704f0dbd/spyder/plugins/findinfiles.py#L78

So this should be `osp.abspath(directory)` because `getcwd_or_home` already returns unicode.

spyder/widgets/findinfiles.py
ccordoba12: Again, this is only used in one place

https://github.com/jnsebgosselin/spyder/blob/8b9483d72967c4580143fde6adaeec38704f0dbd/spyder/plugins/findinfiles.py#L86

and, since it sets the path of current file in the Editor, I think it's fine as it is, or you can change it to `osp.abspath(path)` if you want to be certain to have a full path.

ccordoba12:
> but I would like you to check again carefully some of the places where we return a path.

I left my comments above.

> So good thinking in pushing it back to 3.2.7.

Yep, I agree too. If things go as planned, we'll release 3.2.7 by the end of January.

spyder/widgets/findinfiles.py
jnsebgosselin: Ok, I'll leave it like that. If it ain't broke, don't fix it.

spyder/widgets/findinfiles.py
jnsebgosselin: Ok thank you. I'll remove the `to_text_string` here.

PR #6118, opened 2018-01-03
PR: Fix crash and error dialog when triggering OverflowError (too large int) when editing arrays and dataframes

### LINKED ISSUE #6114: ArrayEditor crashes Spyder, DataFrame Editor generates OverflowError when entering a value for a cell larger than the max int for the current int size
## Description

Upon entering/pasting a value into a cell in a dataframe or array editor opened from Variable Explorer that is large enough to cause an overflow of that type, Spyder doesn't catch the resulting `OverflowError` and handle it appropriately, instead displaying an error dialog when the cell is clicked off to commit the value (which is, of course, left unchanged). Additionally, for arrays, something in the exception handling isn't right, such that this situation crashes the entire program with the standard "Python has stopped working error"...*except* if ``SPYDER_DEBUG=3``, in which case the error message is just printed to the shell spyder is started from and everything else proceeds normally.

1. Create a pandas DataFrame of type ``int32`` or ``int64`` (other integer types aren't supported at all), e.g. with
```
import pandas as pd
df = pd.DataFrame([0, 0, 0], dtype=np.int32)
```
or an array, with e.g.
```
import numpy as np
test_array = np.array([0, 0, 0], dtype=np.int32)
``` 
2. Open the dataframe or array from Variable Explorer in a DataFrame/Array Editor 
3. Paste in a value larger than the data type can hold, e.g. ``17179869184`` for ``int32`` or ``73786976294838206464`` for ``int64``, and hit enter or click off.

Expected: No hard crash with arrays, and a popup warning dialog similar to the behavior entering a text string in a float column.

When entering a value too large for the current type (int32) but not larger than 64 bit (int64) in an array:
 
```
  File "C:\Users\christopher.gerlach\Anaconda3\lib\site-packages\spyder\widgets\variableexplorer\arrayeditor.py", line 318, in setData
    self.test_array[0] = val # will raise an Exception eventually
OverflowError: Python int too large to convert to C long
```
 
When entering a value greater than an int64 into an int64 cell in an array:
 
```
  File "C:\Users\christopher.gerlach\Anaconda3\lib\site-packages\spyder\widgets\variableexplorer\arrayeditor.py", line 318, in setData
    self.test_array[0] = val # will raise an Exception eventually
OverflowError: int too big to convert
```
 
Further, both cases then print the following (followed by a hard crash, “Python has stopped working” etc., except in debug mode (SPYDER_DEBUG=3), where everything seems to work fine other than printing the error to the debug console, and of course not actually updating the array with the new value):
 
```
During handling of the above exception, another exception occurred:
 
  File "C:\Users\christopher.gerlach\Anaconda3\lib\site-packages\spyder\widgets\variableexplorer\arrayeditor.py", line 320, in setData
    print(type(e.message))  # spyder: test-skip
AttributeError: 'OverflowError' object has no attribute 'message'
```
 
 
 
For Dataframes, similar occurs but the error is caught by Spyder before it crashes.
 
When entering a value too large for the current type (int32) but not larger than 64 bit (int64) in a df:
 
```
  File "C:\Users\christopher.gerlach\Anaconda3\lib\site-packages\spyder\widgets\variableexplorer\dataframeeditor.py", line 368, in setData
    self.df.iloc[row, column-1] = current_value.__class__(val)
OverflowError: Python int too large to convert to C long
```
 
When entering a value greater than an int64 into an int64 cell in a df:
 
```
File "C:\Users\christopher.gerlach\Anaconda3\lib\site-packages\spyder\widgets\variableexplorer\dataframeeditor.py", line 368, in setData
self.df.iloc[row, column-1] = current_value.__class__(val)
OverflowError: int too big to convert
```

**Proposed Solution:**

First of all, fix the exception handling that allows the application hard crash to occur. Then, catch the error in each case, and generate a popup warning dialog with the error similar to the behavior when entering e.g. a value that cannot be converted to the specified data type, like a text string in a numeric column. Ideally, both this case and the “cannot convert” one mentioned should abort while leaving the user with the edit-mode text field with the value they entered so they can more easily correct it, rather than just kicking them out to non-edit mode, showing the previous value, but this would probably take more work than the current behavior.
 

* Spyder Version: 3.2.5 
* Python Version: 3.6.3

## Dependencies
```

```

CAM-Gerlach:
Fixes #6114 .

Spyder array and dataframe editor now fail gracefully when a value too large for the given (int) datatype is entered in a cell, showing a warning message rather than crashing or displaying an error dialog, respectively (the former due to a Python3-incompatible ``e.message`` syntax of retrieving the error string in that case). Also, a bit of minor associated cleanup.

TODO: Add a simple test for each, ETA Tomorrow...err, later today (Wednesday). Assuming that happens in time, it could be included in 3.2.6, though no pressure on my end.

Also, one really weird bug I wasn't yet able to track down: Before the Array Editor change, Spyder would crash completely ("Python has stopped working...") due to the exception handling bug, *except* when SPYDER_DEBUG=3. In that case, it would just print the error traceback to the shell it was launched in and continue on. However, conversely, once this fix was implemented it now works fine in all scenarios, *except* when SPYDER_DEBUG=3 and one hits enter/return to confirm the whole Array editor dialog, rather than just clicking off of the current cell to save just it. In the former case, the warning dialog is triggered as it should, but after hitting okay, the Variable Explorer window c;loses and then a "Python has stopped working..." crash is triggered. I was able to get the following traceback printed to console:

```
QObject::installEventFilter(): Cannot filter events for objects in a different thread.
  File "C:\Users\C. A. M. Gerlach\Documents\dev\SpyderDev\spyder\spyder\widgets\variableexplorer\arrayeditor.py", line 393, in commitAndCloseEditor
    self.closeEditor.emit(editor, QAbstractItemDelegate.NoHint)
RuntimeError: wrapped C/C++ object of type QLineEdit has been deleted
```

When I tried catching and passing that error, it would still crash, presumably somewhere else maybe in C++ that I couldn't see. Again, this only happened in SPYDER_DEBUG=3, when hitting enter right from the edit box, and with an OverflowError of course. If you have any insight, please let me know; otherwise just ignore this and I'll look into it more tomorrow...er, today.

### DIFF
--- a/spyder/widgets/variableexplorer/tests/test_dataframeeditor.py
+++ b/spyder/widgets/variableexplorer/tests/test_dataframeeditor.py
@@ -290,6 +290,9 @@ def test_dataframemodel_set_data_overflow(monkeypatch):
 
 
 @flaky(max_runs=3)
+@pytest.mark.skipif(os.environ.get('CI', None) is not None and
+                    system() == "Linux",
+                    reason="Fails on Travis CI for some reason.")
 def test_dataframeeditor_edit_overflow(qtbot, monkeypatch):
     """Test #6114: entry of an overflow int is caught and handled properly"""
     MockQMessageBox = Mock()
--- a/spyder/widgets/variableexplorer/tests/test_dataframeeditor.py
+++ b/spyder/widgets/variableexplorer/tests/test_dataframeeditor.py
@@ -23,6 +23,7 @@ from qtpy.QtGui import QColor
 from qtpy.QtCore import Qt, QTimer
 import numpy
 import pytest
+from flaky import flaky
 
 # Local imports
 from spyder.utils.programs import is_module_installed
@@ -263,6 +264,25 @@ def test_sort_dataframe_with_category_dtypes(qtbot):  # cf. issue 5361
     assert data(dfm, 1, 1) == 'category'
 
 
+def test_dataframemodel_set_data_overflow(monkeypatch):
+    """Unit test #6114: entry of an overflow int caught and handled properly"""
+    MockQMessageBox = Mock()
+    attr_to_patch = ('spyder.widgets.variableexplorer' +
+                     '.dataframeeditor.QMessageBox')
+    monkeypatch.setattr(attr_to_patch, MockQMessageBox)
+
+    for int_type, bit_exponent in [(numpy.int32, 34), (numpy.int64, 66)]:
+        test_df = DataFrame(numpy.arange(2, 7), dtype=int_type)
+        model = DataFrameModel(test_df.copy())
+        index = model.createIndex(2, 1)
+        assert not model.setData(index, str(int(2 ** bit_exponent)))
+        MockQMessageBox.critical.assert_called_with(ANY, "Error", ANY)
+        assert MockQMessageBox.critical.call_count == (bit_exponent - 2) // 32
+        assert numpy.sum(test_df[0].as_matrix() ==
+                         model.df.as_matrix()) == len(test_df)
+
+
+@flaky(max_runs=3)
 def test_dataframeeditor_edit_overflow(qtbot, monkeypatch):
     """Test #6114: entry of an overflow int is caught and handled properly"""
     MockQMessageBox = Mock()
@@ -271,11 +291,8 @@ def test_dataframeeditor_edit_overflow(qtbot, monkeypatch):
     monkeypatch.setattr(attr_to_patch, MockQMessageBox)
 
     expected_df = DataFrame([5, 6, 7, 3, 4])
-    error_text_32 = "OverflowError: Python int too large to convert to C long"
-    error_text_64 = "OverflowError: int too big to convert"
-    test_parameters = [(numpy.int32, error_text_32, 34),
-                       (numpy.int64, error_text_64, 66)]
-    for int_type, error_text, bit_exponet in test_parameters:
+    test_parameters = [(numpy.int32, 34), (numpy.int64, 66)]
+    for int_type, bit_exponet in test_parameters:
         test_df = DataFrame(numpy.arange(0, 5), dtype=int_type)
         dialog = DataFrameEditor()
         assert dialog.setup_and_check(test_df, 'Test Dataframe')
@@ -289,7 +306,7 @@ def test_dataframeeditor_edit_overflow(qtbot, monkeypatch):
         qtbot.keyPress(view, Qt.Key_Space)
         qtbot.keyClicks(view.focusWidget(), str(int(2 ** bit_exponet)))
         qtbot.keyPress(view.focusWidget(), Qt.Key_Down)
-        MockQMessageBox.critical.assert_called_with(ANY, "Error", error_text)
+        MockQMessageBox.critical.assert_called_with(ANY, "Error", ANY)
         assert MockQMessageBox.critical.call_count == (bit_exponet - 2) // 32
         qtbot.keyClicks(view, '7')
         qtbot.keyPress(view, Qt.Key_Up)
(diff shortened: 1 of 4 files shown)


CAM-Gerlach:
All builds on Travis, but not AppVeyor or CircleCI, are failing all four tests, apparently due to self.test_array[0] = val not raising an error (OverflowError) when it should, despite confirming that ``val`` is the same as I'm seeing on my machine, i.e. ``1 ** 34**. I'm baffled as to why that might be; it appears both are running identical or nearly so numpy and python versions, and the difference is the same across Python 3.6, 3.5 and 2.7; only different CIs running the test suite. Any ideas?

jitseniesen:
There is also no OverflowError being raised on my Linux box when assigning `2 ** 34` to an `int32` array. Here is a plain (i.e. outside Spyder) IPython session:

```
Python 3.6.3 |Anaconda custom (64-bit)| (default, Oct 13 2017, 12:02:49)
Type 'copyright', 'credits' or 'license' for more information
IPython 6.0.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import numpy as np

In [2]: arr = np.array([1,2,3], dtype=np.int32)

In [3]: arr[0] = 2 ** 34

In [4]: arr
Out[4]: array([0, 2, 3], dtype=int32)

In [5]: arr[0] = 2 ** 66
---------------------------------------------------------------------------
OverflowError                             Traceback (most recent call last)
<ipython-input-5-3cf8afbda221> in <module>()
----> 1 arr[0] = 2 ** 66

OverflowError: Python int too large to convert to C long
```

jitseniesen:
numpy/numpy#6289 seems to confirm this

CAM-Gerlach:
@jitseniesen Sorry to bother you again @jitseniesen , but any idea on this one? Its only `test_dataframeeditor_edit_overflow` that's failing, and it does so on every Travis build consistently; the OverflowError is triggered as it should be as the unit test of that passes, and the overall qtbot navigation seems to work also since ``test_arrayeditor_edit_overflow`` passes too. If it isn't immediately obvious, maybe test it with `qtbot.wait(1000)`'s and see what it does, if you can? I really need to get my old Linux machine up and running, but I'm not even sure if its bootable at this point.

spyder/widgets/variableexplorer/tests/test_dataframeeditor.py
```diff
@@ -263,5 +265,72 @@ def test_sort_dataframe_with_category_dtypes(qtbot):  # cf. issue 5361
...
+        assert not model.setData(index, str(int(2 ** bit_exponent)))
+        MockQMessageBox.critical.assert_called_with(ANY, "Error", ANY)
+        assert MockQMessageBox.critical.call_count == idx
+        assert numpy.sum(test_df[0].as_matrix() ==
+                         model.df.as_matrix()) == len(test_df)
+
+
+@flaky(max_runs=3)
+@pytest.mark.skipif(os.environ.get('CI', None) is not None and
+                    system() == "Linux",
```
ccordoba12: Please use `sys.platform.startswith('linux')` here

ccordoba12:
@CAM-Gerlach, I left a minor comment, then I'd say this is ready.

spyder/widgets/variableexplorer/tests/test_dataframeeditor.py
CAM-Gerlach: Sure, that's what I used on the other one previously, but couldn't remember the exact syntax. Just curious, what's the advantage over platform.system()?

ccordoba12: It's just our custom throughout our entire codebase.

CAM-Gerlach: Makes sense, thanks.

CAM-Gerlach:
Updated as requested, plus more minor cleanup. CIs are still running, but I'm going to try to get a little sleep before work so wanted to let you know before. Thanks!

CAM-Gerlach:
@ccordoba12 Checks passed, so should be ready to go.

PR #6123, opened 2018-01-04
PR: Allow tab cycling shortcuts to be configurable in Editor

### LINKED ISSUE #4842: Add ability to change Ctrl+PgUp and PgDn shortcuts
I really dislike editors with MRU behaviour but at the same time I really like Ctrl+Tab and Ctrl+Shift+Tab because I don't have to remove my hand from the mouse like with Ctrl+PgUp or PgDn. I wish I had the ability to customize Ctrl+PgUp and PgDn just like Tab and Shift+Tab so I could replace them and use my preferred behaviour with my preferred shortcuts.

Thanks in advance

csabella:
Fixes #4842.  

- Added shortcuts to cycle through files.
- If statement to convert `Backtab` to `Shift+Tab` was only for 'nt', but problem existed on linux too, so removed the conditional.
- Tab navigation routine already existed in `Tabs` class, so moved it to `BaseTabs` class for it to be used in the editor (which creates a `BaseTabs` widget) without affecting existing code (like in iPython console).

### DIFF
--- a/spyder/widgets/tabs.py
+++ b/spyder/widgets/tabs.py
@@ -385,6 +385,16 @@ class BaseTabs(QTabWidget):
                 handled = True
         if not handled:
             QTabWidget.keyPressEvent(self, event)
+
+    def tab_navigate(self, delta=1):
+        """Ctrl+Tab"""
+        if delta > 0 and self.currentIndex() == self.count()-1:
+            index = delta-1
+        elif delta < 0 and self.currentIndex() == 0:
+            index = self.count()+delta
+        else:
+            index = self.currentIndex()+delta
+        self.setCurrentIndex(index)
         
     def set_close_function(self, func):
         """Setting Tabs close function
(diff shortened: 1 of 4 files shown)


spyder/widgets/tabs.py
```diff
@@ -385,6 +385,16 @@ def keyPressEvent(self, event):
                 handled = True
         if not handled:
             QTabWidget.keyPressEvent(self, event)
+
+    def tab_navigate(self, delta=1):
```
ccordoba12: @csabella, `Ctrl+Tab` is handled differently in the Editor, because there we show a dialog to select among the current opened files. So I think you need to revert this change.

csabella: @ccordoba12   It still works correctly in the Editor even with moving this.  Without changing preferences, `Ctrl+Tab` brings up the dialog to select the file in MRU  order and `Ctrl+PgDown` cycles through the files.  But, this code allows those keys to be configurable.

ccordoba12: O, great! Then this is ready.

PR #6129, opened 2018-01-04
PR: Enable the display of many more numpy array types in the Variable Explorer

CAM-Gerlach:
Currently, numpy arrays of types int64, int32, float64, float32, complex64, and complex128 are displayed with `repr` instead of `str`, leading to unnecessarily/redundant clutter being displayed (that its an array, and its dtype usually, both already displayed natively by VarExp. Further, the types  int16, int8, uint64, uint32, uint16, uint8, float16, and bool aren't displayed at all; just as ``ndarray object of numpy module``. Therefore, this PR fixes both of those issues.

Before:
image

After:
image

TODO: Add simple tests for proper display

### DIFF
--- a/spyder/widgets/variableexplorer/utils.py
+++ b/spyder/widgets/variableexplorer/utils.py
@@ -50,14 +50,16 @@ class FakeObject(object):
 #==============================================================================
 try:
     from numpy import (ndarray, array, matrix, recarray,
-                       int64, int32, float64, float32,
-                       complex64, complex128)
+                       int64, int32, int16, int8, uint64, uint32, uint16, uint8,
+                       float64, float32, float16, complex64, complex128, bool_)
     from numpy.ma import MaskedArray
     from numpy import savetxt as np_savetxt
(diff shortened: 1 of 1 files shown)


ccordoba12:
Nice!! Thanks a lot for this one :+1:

CAM-Gerlach:
@ccordoba12 Thanks! It was really quite a trivial thing to implement; I noticed the issue while testing #6118 and (a rarity these days) didn't take even me more than a few minutes to figure out. 

I left the tests for later, since going by past measures they are bound to take way longer than the actual code (at least for Spyder, a 5 or 10:1 ratio isn't uncommon with all the many problems so far), and before 3.2.6 drops I need to finish the FAQ stuff plus  #6132 's tests and a tweak to consistently order (alphabetically, ideally) the dependency list in the  dialog/manual copy vs. the auto report, as it has been very difficult to compare the two different formats, both ordered seemingly arbitrarily, to check for commonalities in issue reports.

PR #6132, opened 2018-01-04
PR: Fix TypeError editing any complex val in DataFrameEditor, and bug resulting in bool_s being treated as uneditable

### LINKED ISSUE #6115: DataFrame Editor raises TypeError when editing any complex type
## Description

When editing a column of a complex type (``complex64``, ``complex128`` and the built in ``complex``), the DataFrame Editor raises a ``TypeError`` when attempting to confirm/commit any edits to any of the values in the column. This seems to occur regardless of what value is entered, whether the original (displayed as “###j), a tuple, a single integer, etc. However, it works just fine for arrays, so perhaps that logic can just be ported over?

1. Create a DataFrame with at least one column being of a complex type. E.g.
```
import pandas as pd
import numpy as np
df = pd.DataFrame([0, 0, 0], dtype=np.complex128)
```
2. Open the DataFrame in the Data Frame editor
3. Doubleclick one of the complex values to edit and click away again to commit

Expected: Either a functioning ability to input a complex value as works perfectly well for arrays, or at the very least a "The type of the cell is not a supported type" warning or just setting the cell to read only.

Instead, an error dialog with :

```
File "C:\Users\christopher.gerlach\Anaconda3\lib\site-packages\spyder\widgets\variableexplorer\dataframeeditor.py", line 368, in setData
self.df.iloc[row, column-1] = current_value.__class__(val)
TypeError: must be real number, not str
```

**Proposed solution**

Port whatever logic is currently used for ArrayEditor to handle working with complex types over to the DataFrame Editor, if that's practicable. If not, and there's no other straightforward way to fix this, just catch the exception and pop up a "The type of the cell is not a supported type" warning dialog on attempting to edit instead for the time being, or disable editing completely if possible (so a double click action is just ignored).

* Spyder Version: 3.2.5 
* Python Version: 3.6.3

## Dependencies
```

```

CAM-Gerlach:
Fixes #6115 

Currently, complex datatypes are treated as being editable in dataframe editor, but they actually aren't currently (though they are in arrays, and that functionality will be ported over Soon™). This leads to DataFrameEditor raising an uncaught ``TypeError`` if one is attempted to be edited, even if no changes are made. 

Also, while investigating that bug, it was also noticed that bools were not allowed to be edited, even though they were intended to be, due to only including the python `bool` type which the numpy `bool_` type, i.e. the actual standard boolean type for numpy arrays, not inheriting from the former, which was also easily fixed. 

In addition, minor enhancements were made to "false detection" for booleans to cover the standard values treated as "False" by Python. Eventually, the simpler/easier system for arrays (simply clicking to toggle true/false) will likely be ported over, but this should do for now.

TODO: Simple tests for both

### DIFF
--- a/spyder/widgets/variableexplorer/dataframeeditor.py
+++ b/spyder/widgets/variableexplorer/dataframeeditor.py
@@ -42,7 +42,7 @@ from spyder.widgets.variableexplorer.arrayeditor import get_idx_rect
 REAL_NUMBER_TYPES = (float, int, np.int64, np.int32)
 COMPLEX_NUMBER_TYPES = (complex, np.complex64, np.complex128)
 # Used to convert bool intrance to false since bool('False') will return True
-_bool_false = ['false', 'f', '0', '0.', '0.0']
+_bool_false = ['false', 'f', '0', '0.', '0.0', ' ']
 
 # Default format for data frames with floats
 DEFAULT_FORMAT = '%.6g'
--- a/spyder/widgets/variableexplorer/dataframeeditor.py
+++ b/spyder/widgets/variableexplorer/dataframeeditor.py
@@ -361,7 +361,7 @@ class DataFrameModel(QAbstractTableModel):
             current_value = self.get_value(row, column-1)
             if isinstance(current_value, bool):
                 val = bool_false_check(val)
-            supported_types = (bool,) + REAL_NUMBER_TYPES + COMPLEX_NUMBER_TYPES
+            supported_types = (bool,) + REAL_NUMBER_TYPES
             if (isinstance(current_value, supported_types) or
                     is_text_string(current_value)):
                 try:
--- a/spyder/widgets/variableexplorer/dataframeeditor.py
+++ b/spyder/widgets/variableexplorer/dataframeeditor.py
@@ -372,8 +372,8 @@ class DataFrameModel(QAbstractTableModel):
                     return False
             else:
                 QMessageBox.critical(self.dialog, "Error",
-                                     "The type of the cell is not a supported "
-                                     "type")
+                                     "Editing dtype {0!s} not yet supported."
+                                     .format(type(current_value).__name__))
                 return False
         self.max_min_col_update()
         return True
--- a/spyder/widgets/variableexplorer/dataframeeditor.py
+++ b/spyder/widgets/variableexplorer/dataframeeditor.py
@@ -42,7 +42,7 @@ from spyder.widgets.variableexplorer.arrayeditor import get_idx_rect
 REAL_NUMBER_TYPES = (float, int, np.int64, np.int32)
 COMPLEX_NUMBER_TYPES = (complex, np.complex64, np.complex128)
 # Used to convert bool intrance to false since bool('False') will return True
-_bool_false = ['false', '0']
+_bool_false = ['false', 'f', '0', '0.', '0.0']
 
 # Default format for data frames with floats
 DEFAULT_FORMAT = '%.6g'
@@ -359,9 +359,9 @@ class DataFrameModel(QAbstractTableModel):
         else:
             val = from_qvariant(value, str)
             current_value = self.get_value(row, column-1)
-            if isinstance(current_value, bool):
+            if isinstance(current_value, (bool, np.bool_)):
                 val = bool_false_check(val)
-            supported_types = (bool,) + REAL_NUMBER_TYPES
+            supported_types = (bool, np.bool_) + REAL_NUMBER_TYPES
             if (isinstance(current_value, supported_types) or
                     is_text_string(current_value)):
                 try:
--- a/spyder/widgets/variableexplorer/tests/test_dataframeeditor.py
+++ b/spyder/widgets/variableexplorer/tests/test_dataframeeditor.py
@@ -357,6 +357,9 @@ def test_dataframemodel_set_data_complex(monkeypatch):
 
 
 @flaky(max_runs=3)
+@pytest.mark.skipif(os.environ.get('CI', None) is not None and
+                    platform.startswith('linux'),
+                    reason="Fails on Travis for no good reason.")
 def test_dataframeeditor_edit_complex(qtbot, monkeypatch):
     """Test for #6115: editing complex dtypes raises error in df editor"""
     MockQMessageBox = Mock()
@@ -402,7 +405,7 @@ def test_dataframemodel_set_data_bool(monkeypatch):
                      '.dataframeeditor.QMessageBox')
     monkeypatch.setattr(attr_to_patch, MockQMessageBox)
 
-    test_params = [(1, numpy.bool_), (2, numpy.bool), (3, bool)]
+    test_params = [numpy.bool_, numpy.bool, bool]
     test_strs = ['foo', 'false', 'f', '0', '0.', '0.0', '', ' ']
     expected_df = DataFrame([1, 0, 0, 0, 0, 0, 0, 0, 0], dtype=bool)
 
@@ -424,7 +427,7 @@ def test_dataframeeditor_edit_bool(qtbot, monkeypatch):
                      '.dataframeeditor.QMessageBox')
     monkeypatch.setattr(attr_to_patch, MockQMessageBox)
 
-    test_params = [(1, numpy.bool_), (2, numpy.bool), (3, bool)]
+    test_params = [numpy.bool_, numpy.bool, bool]
     test_strs = ['foo', 'false', 'f', '0', '0.', '0.0', '', ' ']
     expected_df = DataFrame([1, 0, 0, 0, 0, 0, 0, 0, 0], dtype=bool)
 
--- a/spyder/widgets/variableexplorer/tests/test_dataframeeditor.py
+++ b/spyder/widgets/variableexplorer/tests/test_dataframeeditor.py
@@ -396,5 +396,59 @@ def test_dataframeeditor_edit_complex(qtbot, monkeypatch):
                          dialog.get_value().as_matrix()) == len(test_df)
 
 
+def test_dataframemodel_set_data_bool(monkeypatch):
+    """Unit test that bools are editible in df and false-y strs are detected"""
+    MockQMessageBox = Mock()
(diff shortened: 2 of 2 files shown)


ccordoba12:
@CAM-Gerlach, this needs a merge with 3.x and tests. It's the last PR before releasing 3.2.6 :-)

CAM-Gerlach:
@ccordoba12 Thanks for the heads up and sorry about that, I had to do an all nighter thanks mostly to the #6114 test issues, and then had work; already rebased it and was just working on the tests now—I've based them off those for that former issue.

But wait, if this is the last PR then what about adding the troubleshooting link that I was going to do next, at least to a few low hanging fruit places (issue template, error dialog, "crashed last time" startup message, and in the readme/doc? Those at least shouldn't be more than a copy and paste addition to a text string, with no functional changes, but if we don't then our efforts assembling the guide (both helping users and making you and my life easier dealing with fewer issues) won't pay off for another 2 months since the people who need it are unlikely to find it on their own.

I'm sorry I didn't do it sooner, but the #6114 fiasco (plus a medical emergency in the family) derailed those plans, and I thought I had at least another day (or I would have taken off from work today and gone in tomorrow). I can have the PR ready by midnight CST, unless I run into major issues with these tests.  Can it still make it? Pretty please :)

ccordoba12:
> But wait, if this is the last PR then what about adding the troubleshooting link that I was going to do next, at least to a few low hanging fruit places (issue template, error dialog, "crashed last time" startup message, and in the readme/doc?

Ok, no prob. I can delay the release until Monday. Is that ok for you?

CAM-Gerlach:
> Ok, no prob. I can delay the release until Monday. Is that ok for you?

Thanks! As mentioned, I can have it done by later tonight, if that works for you, assuming nothing major goes wrong with the #6115 tests, so hopefully that would still allow for a Friday release...I leave for a conference after tomorrow and won't have much free time or internet access, so I'm not sure how much more additional work that later deadline would allow me to get in anyway.

CAM-Gerlach:
@ccordoba12 Since whatever issue that is causing the "real world" dataframe edit tests on Linux to fail appears it may be systemic, likely the same as that affecting the previous set, and likely others I've had to skip in the past, are you able to do me a big favor and manually test if the expected behavior (Warning dialog when editing a complex df, e.g. ``test_df = pd.DataFrame([0, 1, 2, 3, 4], dtype=np.complex128)`` actually occurs as it should, and see what's actually happening during the test, e.g. by injecting ``qtbot.wait(1000)`` statements and watching closely?

I would ask @jitseniesen as well, but he actually said the previous test that failed on your machine passed on his, so not sure what's going on. Since ``dataframeeditor_edit_overflow`` fails yet the essentially identical (except action an an arrayeditor) ``arrayeditor_edit_overflow`` succeeds, it seems like it has to be something specific to dfs.

jitseniesen:
Just to confirm that the "Warning dialog when editing a complex df" does indeed appear on my linux box.

The new dataframe tests fail on my box, but with a different cause than on the Travis server.
For `test_dataframeeditor_edit_complex`, I get:
```
E           AssertionError: Expected call: critical(<ANY>, 'Error', 'Editing dtype complex64 not yet supported.')
E           Actual call: critical(<spyder.widgets.variableexplorer.dataframeeditor.DataFrameEditor object at 0x7fb164080798>, 'Error', 'Editing dtype complex128 not yet supported.')
```
For `test_dataframeeditor_edit_bool`, I get:
```
>           assert (numpy.sum(expected_df[0].as_matrix() ==
                              dialog.get_value().as_matrix()[:, 0]) ==
                    len(expected_df))
E           assert 1 == 9
E            +  where 1 = <function sum at 0x7fb188b396a8>(array([ True,...], dtype=bool) == array([False, ...], dtype=bool)
E            +    where <function sum at 0x7fb188b396a8> = numpy.sum
E               Use -v to get the full diff)
E            +  and   9 = len(       0\n0   True\n1  False\n2  False\n3  False\n4  False\n5  False\n6  False\n7  False\n8  False)

spyder/widgets/variableexplorer/tests/test_dataframeeditor.py:450: AssertionError
```

CAM-Gerlach:
@jitseniesen Thanks! The first one seems to be due to inconstant/platform specific data type handling, which is easily fixed by just not checking the exact error message text, which is probably a better idea anyway.

As for the second, can you run with the ``-vv`` flag passed to ``pytest``, so I can actually see what' s going on with the array? E.g. 
```
pytest spyder/widgets/variableexplorer/tests/test_dataframeeditor.py -vv -rw
```

Thanks!

jitseniesen:
Now,  `test_dataframeeditor_edit_complex` fails with:
```
           assert MockQMessageBox.critical.call_count == count * 2 - 1
E           AssertionError: assert 2 == ((2 * 2) - 1)
E            +  where 2 = <Mock name='mock.critical' id='140692473287904'>.call_count
E            +    where <Mock name='mock.critical' id='140692473287904'> = <Mock id='140692473289192'>.critical

spyder/widgets/variableexplorer/tests/test_dataframeeditor.py:384: AssertionError
```
I think that the failure is not "due to inconstant/platform specific data type handling" (which was also what I originally thought). Instead, the message box does not trigger in the second iteration of the for loop, so the message that `assert_called_with` checks still has `complex128` in it from the first iteration.

I checked by hand that the message box is triggered as expected with a `complex64` data frame.

After some experimentation, I found a solution on my box: add `qrbot.wait(1000)` immediately after you simulate pressing the `Return` key. The test now passes reliably. Even better, the same change also fixes the `test_dataframeeditor_edit_bool` test.

For the record, the full error message for the edit bool test is:
```
E           assert 1 == 9
E            +  where 1 = <function sum at 0x7f9d43e7d488>(array([ True,...], dtype=bool) == array([False, ...], dtype=bool)
E            +    where <function sum at 0x7f9d43e7d488> = numpy.sum
E               Full diff:
E               - array([ True, False, False, False, False, False, False, False, False], dtype=bool)
E               + array([False,  True,  True,  True,  True,  True,  True,  True, False], dtype=bool))
E            +  and   9 = len(       0\n0   True\n1  False\n2  False\n3  False\n4  False\n5  False\n6  False\n7  False\n8  False)
```

CAM-Gerlach:
Thanks so much @jitseniesen for your diligent help. I should have those build failures fixed on your platform, but unfortunately Travis still has that other error, same as the other PR we tried to work through, and probably the same as the one before that—and, more promisingly, possibly the same that @ccordoba12 had that. Obviously I'd like to get to the bottom of this persistent issue and fix it for good, but I can't do much without a Linux box, and in any case it appears it is only failing on certain machines, not it might not even do me any good. 

As I don't want to delay the release any further than i already have, so I've told those two integration tests to just skip for now on Travis, since they have the corresponding unit tests and the other CIs to back them up. However, if @ccordoba12 can investigate this a bit further once we have the time, or at least if those tests fail again on his machine, then that would likely be very helpful in figuring out what keeps happening rather than just skipping tests. Thanks!

In any case, unless @ccordoba12 has a fix, I need to try to get at least a little sleep before another workday today, an after work meeting to teach some people Spyder, getting ready for my trip and then leaving for my flight at 4AM tomorrow.

jitseniesen:
In fact, even with `wait(1000)` as I suggested the tests do not pass as reliably on my box as I thought. I did have another look, but without much success. Here are some loose observations:
* Sometimes the DataFrameEditor window does not close and the tests fails
* The docs for `keyPress()` at https://doc.qt.io/qt-5/qtest.html#keyPress say: "At some point you should release the key using `keyRelease()`." Perhaps we should use `keyClick()` instead. See https://stackoverflow.com/questions/9451559/what-is-the-difference-between-clicked-and-pressed-in-qt for the difference.
* I tried giving a `delay` argument but that leads to a Segmentation fault.

ccordoba12:
> However, if @ccordoba12 can investigate this a bit further once we have the time

Given that there's no platform specific code related to the tests you introduced here (and also not in `dataframeeditor.py`), I'm fine with these tests being run on Windows for now.

----

I also forgot that I tackled this focusing problem when our Pytest's were running in CircleCI! According to the last comment of this issue:

https://github.com/TestFX/TestFX/issues/158

you need to install a windows manager (e.g. `matchbox-window-manager`) to fix this. The code for that is still there in our `circle.yml`:

https://github.com/spyder-ide/spyder/blob/master/circle.yml#L12-L21

But you can work on that in another PR :-)

CAM-Gerlach:
> I also forgot that I tackled this focusing problem when our Pytest's were running in CircleCI! 

Nice! That would really explain it, since the common failure mode occurred during manipulating a window and grabbing focus. I actually happened to noticed that mention in the CircleCI test config/log, but assumed it wasn't a problem on Travis. However, on the other hand I'm curious why the ``edit_array_overflow`` test passes, when it essentially does the same stuff, just on a ``DataFrameEditor`` instead of an ``ArrayEditor``; also, whether the failure on your machine was at the same assert/for the same reason.

> You need to install a window manager (e.g. matchbox-window-manager) to fix this.

How tough would this be for a total newbie to all of this stuff? Just sticking the ``sudo install matchbox,,,`` line into the right spot in the ``Travis.yml`` file? I'm sure it must be more complicated than that, but I have no idea. Thanks.

CAM-Gerlach:
> Sometimes the DataFrameEditor window does not close and the tests fails

Hmm, thanks @jitseniesen . I wonder if the return key press isn't getting received, or if its something going wrong in the close process itself. Either way, I'm sure there's a method I can try using to force it to close...

>  At some point you should release the key using keyRelease()

Right, I read that several places and was wondering about that, but I never noticed a difference when I swapped them out in numerous tests, at least on my machine, and keyPress is used in a number of other tests just fine without keyRelease, so I'm not sure what's going on.

ccordoba12:
> How tough would this be for a total newbie to all of this stuff?

It shouldn't be that hard to achieve. First you need to add `sudo apt-get install ...` to our `.travis.yml`:

https://docs.travis-ci.com/user/installing-dependencies/

Then you have to start the window manager in the background:

https://github.com/travis-ci/travis-ci/issues/1321

CAM-Gerlach:
Thanks @ccordoba12 ; I'll give it a shot soon.

PR #6084, opened 2017-12-29
PR: Major overhaul of Readme, contributing and install docs

CAM-Gerlach:
To partially address #6075 as well as a related problem that @csabella and I reproduced that prevented installing from source, as well as more generally make clear our recommendation to use `conda`/Anaconda over `pip` and other distributions and our inability to intervene to solve installation issues with those who choose the latter (as we get many every day on our repo), I've revised the readme, CONTRIBUTING.md and installation documentation accordingly. As part of that, as I'd been hoping to for a while, I took the opportunity to reorganize those sections to eliminate redundancy and outline the alternatives clearly, as well as associated minor cleanup and copyediting. I left the dependancy, etc. and other sections as alone as possible to avoid possible conflicts when forward-porting to `master`.

PR #6137, opened 2018-01-05
PR: Add Troubleshooting links and blurb to various appropriate places in Spyder and doc

CAM-Gerlach:
Adds a link and a short blurb regarding the new troubleshooting guide/FAQ in various appropriate places around the Spyder UI, including:

- About dialog
- Blurb/link as part of any crash/error dialog
- Blurb/link as part of the "crashed last time" startup message recommending spyder --reset
- Blurb/link in default Spyder issue template under .github/ISSUE_TEMPLATE.md
- And the auto-created/submitted one
- In the Readme, contributing.md, documentation index, and install doc
- Menu item in the help menu

Also some associated cleanup and polishing of related error and help guidance text, to help better guide users.

Finally, also includes a ``TROUBLESHOOTING.md`` document, along the lines of ``CONTRIBUTING.md`` which features a link and blurb to the full online guide/FAQ, as well as a few of the static sections that are likely to stay relatively unchanged, i.e, general troubleshooting tips and strategy, how to submit an issue, etc. Therefore, there is no need to hassle with maintaining it regularly, as it won't really become stale; and as its Markdown just like the wiki, it doesn't restrict compatibility whatsoever, and the relevant sections, if updated, can be easily copied and pasted in every once in a while if desired.

Thanks!

### DIFF
--- a/spyder/__init__.py
+++ b/spyder/__init__.py
@@ -33,6 +33,7 @@ __version__ = '.'.join(map(str, version_info))
 __license__ = __doc__
 __project_url__ = 'https://github.com/spyder-ide/spyder'
 __forum_url__   = 'http://groups.google.com/group/spyderlib'
+__trouble_url__ = __project_url__ + '/wiki/Troubleshooting-Guide-and-FAQ'
 
 # Dear (Debian, RPM, ...) package makers, please feel free to customize the
 # following path to module's data (images) and translations:
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -140,7 +140,8 @@ else:
 #==============================================================================
 # Local utility imports
 #==============================================================================
-from spyder import __version__, __project_url__, __forum_url__, get_versions
+from spyder import (__version__, __project_url__, __forum_url__,
+                    __trouble_url__, get_versions)
 from spyder.config.base import (get_conf_path, get_module_data_path,
                                 get_module_source_path, STDERR, DEBUG,
                                 debug_print, MAC_APP_NAME, get_home_dir,
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -937,6 +937,9 @@ class MainWindow(QMainWindow):
         self.set_splash(_("Setting up main window..."))
 
         # Help menu
+        trouble_action = create_action(self,
+                                        _("Troubleshooting..."),
+                                        triggered=self.trouble_guide)
         dep_action = create_action(self, _("Dependencies..."),
                                     triggered=self.show_dependencies,
                                     icon=ima.icon('advanced'))
@@ -1011,7 +1014,8 @@ class MainWindow(QMainWindow):
 
         self.help_menu_actions = [doc_action, tut_action, shortcuts_action,
                                   self.tours_menu,
-                                  MENU_SEPARATOR, report_action, dep_action,
+                                  MENU_SEPARATOR, trouble_action,
+                                  report_action, dep_action,
                                   self.check_updates_action, support_action,
                                   MENU_SEPARATOR]
         # Python documentation
@@ -2459,6 +2463,9 @@ class MainWindow(QMainWindow):
             if title:
                 url.addEncodedQueryItem("title", quote(title))
 
+    @Slot()
+    def trouble_guide(self):
+        url = QUrl(__trouble_url__)
         QDesktopServices.openUrl(url)
 
     @Slot()
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -3125,7 +3125,8 @@ def main():
         CONF.set('main', 'crash', False)
         if SPLASH is not None:
             SPLASH.hide()
-        QMessageBox.information(None, "Spyder",
+        QMessageBox.information(
+            None, "Spyder",
             "Spyder crashed during last session.<br><br>"
             "If Spyder does not start at all and <u>before submitting a "
             "bug report</u>, please try to reset settings to defaults by "
@@ -3135,12 +3136,15 @@ def main():
             "<span style=\'color: #ff5555\'><b>Warning:</b></span> "
             "this command will remove all your Spyder configuration files "
             "located in '%s').<br><br>"
-            "If restoring the default settings does not help, please take "
+            "If Spyder still fails to launch, you should consult our "
+            "comprehensive <b><a href=\"%s\">Troubleshooting Guide</a></b>, "
+            "which when followed carefully solves the vast majority of "
+            "crashes; also, take "
             "the time to search for <a href=\"%s\">known bugs</a> or "
             "<a href=\"%s\">discussions</a> matching your situation before "
-            "eventually creating a new issue <a href=\"%s\">here</a>. "
+            "submitting a report to our <a href=\"%s\">issue tracker</a>. "
             "Your feedback will always be greatly appreciated."
-            "" % (get_conf_path(), __project_url__,
+            "" % (get_conf_path(), __trouble_url__, __project_url__,
                   __forum_url__, __project_url__))
 
     # Create main window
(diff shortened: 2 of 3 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -1,3 +1,9 @@
+**PLEASE READ:** Before posting this report, *please* carefully read our **Troubleshooting Guide** at https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ and search the issues page here for your error message and problem description, as the great majority of bugs are either duplicates, or can be fixed on the user side with a few easy steps.
```
ccordoba12: Please break this at 80 characters or so, to not have it in a single line.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,3 +1,9 @@
...
+
+Also, if you do post an issue here, please describe step by step what you were doing when the error occured, in as much detail as possible. Otherwise, we will have to close your issue after one week (7 days) if we can't reproduce it. Thanks.
```
ccordoba12: Also break this single line in multiple ones.

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,492 @@
+Spyder Troubleshooting Guide
```
ccordoba12: Is this file referenced from somewhere else? If not, my vote is to remove it and only leave the one in the wiki.

The problem with having two places with the same content is that (inevitably) they will diverge at some point in time.

ccordoba12:
@CAM-Gerlach, please post screenshots to see how this looks in all Spyder places you added this. Thanks!

.github/ISSUE_TEMPLATE.md
CAM-Gerlach: I originally had it that way, but when it I tested how it would actually display on Github, it took up a a lot of excessive vertical space, since Github normally soft-wraps at 105 - 115 characters (at least in my browser) but respects the breaks in the file instead of rewrapping it properly and thus takes up ~50% more vertical room. 

I don't like leaving really long lines like this either, but pragmatically, as the intended use is for display on Github, and this file will normally never be seen/directly used (it only has one change in its entire history since the day it was first committed a few years ago, compared to several dozen issues created each day—around a ~5,000:1 difference), it seemed to make more sense to optimize for where it will actually matter 99.96% of the time. 

On the other hand, the narrower columns break nicely (except for the link, which I don't embed since its primarily for use in the raw/text view anyway) and help readability, It's not a big deal but just let me know which you prefer; you can compare them in the following screenshots illustrating the difference:

Editing View:

image
vs.
image


Rendered View:

image
vs.
image

.github/ISSUE_TEMPLATE.md
CAM-Gerlach: See above comment.

TROUBLESHOOTING.md
CAM-Gerlach: > Is this file referenced from somewhere else? If not, my vote is to remove it and only leave the one in the wiki.

Fair enough; I figured this could be an objection, which is why I made it the last commit and not reference it elsewhere to make it easy to revert/remove if needed :) 

> The problem with having two places with the same content is that (inevitably) they will diverge at some point in time.

I wasn't going to include it for the same reason, but I ended up reconsidering it in the end after I specifically tailored it to only include the essentially static general troubleshooting content, which shouldn't change substantially at least over a reasonable timescale so it won't ever be that out of date, and it also includes a disclaimer and link at the top to the full/live version of the guide. Conveniently, the sections I included conveniently form the first and last chunk of the full version, and it uses the exact same Github markdown syntax, so all that needs to be done to update it every few months, if need be, is a simple copy and paste (which I am more than happy to do before each release, assuming there are actually any changes to the relevant sections). 

On the other hand, the benefits of having it there aren't enormous, but the main reason is that troubleshooting help is included with the program itself, same as the other documentation, readme, contributing, etc (which in some cases could also be argued to be duplicative, and much harder to maintain due to different syntax and structure), and so if the user doesn't have access to the Internet (or Github, in some countries), they are still able to get help. As a secondary benefit, it also makes it (and the linked guide) more visible/easily accessible to those browsing the source locally or on Github, and serves as a tracked, publicly accessible backup of the key sections in case Github wikis ever go down or go away.

However, I'm certainly far from wed to it, and already mentally prepared myself for the possible breakup :). With the above in mind, let me know what you'd like me to do with it. Thanks!

CAM-Gerlach:
Sorry for the delay; just helped a friend get started using Spyder. I should have included them from the beginning, since already had to bring each one up anyway when checking that they displayed correctly. Here are the screenshots:

**About Spyder dialog:**
image

**"Crashed last time" dialog:**
image

**Error dialog:**
image

**Help Menu item:**
image

**Autogenerated error report**
image

Thanks!

ccordoba12:
@CAM-Gerlach, this all looks very well! But since this involves adding new strings to the interface, I think I'll have to leave it for 3.2.7, sorry.

You could also help us to finish PR #5346, which would go well with your work here.

ccordoba12:
I know you worked very hard to get this done in very short time, but the strings this PR changes are quite visible and so I'd prefer not to push them in a rush and leave them for 3.2.7 instead.

If you like, we could have that release by the end of the month with these changes, your PR about our README and the improvements to the report issues dialog I mentioned above. That would be a very good release too!

What do you say?

CAM-Gerlach:
> But since this PR involves adding new strings to the interface, I think I'll have to leave it for 3.2.7, sorry.

Okay...I assume for localization/internationalization? To be completely honest, I'm not quite sure how else one would expect to add a link and blurb to the UI...hieroglyphics? :P

CAM-Gerlach:
> You could also help us to finish PR #5346, which would go well with your work here.

Oh, nice! I was actually planning to add something like that anyway, but it didn't seem time-realistic nor really in scope for this PR.

ccordoba12:
> Okay...I assume for localization/internationalization?

Yep, that's right.

> To be completely honest, I'm not quite sure how else one would expect to add a link and blurb to the UI...hieroglyphics? :P

If a string changes, then it falls back to English automatically. So our international users would see (for example) our Issues dialog in English. And that's not really useful :-)

CAM-Gerlach:
> If a string changes, then it falls back to English automatically. So our international users would see (for example) our Issues dialog in English. And that's not really useful :-)

Right, I understand that part, and we certainly don't want to leave Spyder's (many) international users out in the cold, especially on something important like this. Of course, the other issue is that the guide itself isn't translated...

What I was referring to was why it wasn't clear that this would be the case at some point before now. :-)

CAM-Gerlach:
> What do you say?

Sure, I guess. To be honest, while I was a bit miffed at first after pulling an all nighter (well, technically two) and calling in sick to work to make sure this got done, that was all my own (questionable) decisions, not anyone else's.  If there's one lesson I've learned in life as I've grown and matured, there is everything to be gained from accepting life as it comes and focusing on making the best of it. Though, in all fairness it would have been nice to know a *little* beforehand :-) 

One such possibility: Since the github issue template (both the manual and auto versions) is not localized, and that would by itself be able to (one would hope) be seen by at least all the users submitting issues, would you accept for 3.2.6 another PR with just that change?

Prikers:
I can do the French translation whenever you need me to, it should not take long. I mean for the new strings. For the troubleshooting guide itself it may take a bit more time, it is quite detailed :+1:

CAM-Gerlach:
> For the troubleshooting guide itself it may take a bit more time, it is quite detailed 

If you mean the FAQ version, then it will get much more detailed still...Soon™

So the readme, contributing.md, etc. are indeed translated too?

ccordoba12:
> So the readme, contributing.md, etc. are indeed translated too?

Nop, they aren't. I don't know if that's a good idea though.

CAM-Gerlach:
Rebased and pared down to just the core UI commits, minus the report error dialog, which will be included in #6154 along with the github issue template and auto-report enhancements, and moved the doc additions and improvements to #6084 . Further, deconflicted with #5346 , implemented basic tests for the menu item and troubleshooting URL resolution, and did some additional minor cleanup. Should be ready to go on my end, unless you'd like very basic tests (essentially that they display and close properly, since they have no other specific functionality to test) for the "about" and "crashed last session" dialogs.

PR #6138, opened 2018-01-05
PR: Mark introspection test slow as its the slowest in the entire suite

CAM-Gerlach:
A one line PR: Just mark the ``editor_introspection`` test as ``slow``, as due to the timeouts required to make it pass reliably on all systems, it is the slowest in the entire suite, taking over 30s just by itself. Therefore, mark it as such.

### DIFF
--- a/spyder/plugins/tests/test_editor_introspection.py
+++ b/spyder/plugins/tests/test_editor_introspection.py
@@ -15,12 +15,13 @@ try:
     from unittest.mock import Mock
 except ImportError:
     from mock import Mock  # Python 2
-
 from qtpy.QtWidgets import QWidget, QApplication
 from qtpy.QtCore import Qt
 
+# Local imports
 from spyder.utils.introspection.jedi_plugin import JEDI_010
 from spyder.utils.qthelpers import qapplication
+from spyder.py3compat import PY2
 
 # Location of this file
 LOCATION = osp.realpath(osp.join(os.getcwd(), osp.dirname(__file__)))
@@ -59,6 +60,7 @@ def setup_editor(qtbot, monkeypatch):
 
 
 @pytest.mark.slow
+@pytest.mark.skipif(PY2, reason="Strange segfaults with other tests on Py2.")
 @pytest.mark.skipif(not JEDI_010,
                     reason="This feature is only supported in jedy >= 0.10")
 def test_introspection(setup_editor):
--- a/spyder/plugins/tests/test_editor_introspection.py
+++ b/spyder/plugins/tests/test_editor_introspection.py
@@ -15,6 +15,7 @@ try:
     from unittest.mock import Mock
 except ImportError:
     from mock import Mock  # Python 2
+from qtpy import PYQT4
 from qtpy.QtWidgets import QWidget, QApplication
 from qtpy.QtCore import Qt
 
@@ -60,7 +61,8 @@ def setup_editor(qtbot, monkeypatch):
 
 
 @pytest.mark.slow
-@pytest.mark.skipif(PY2, reason="Strange segfaults with other tests on Py2.")
+@pytest.mark.skipif(PY2 or PYQT4,
+                    reason="Segfaults with other tests on Py2 and PyQt4.")
 @pytest.mark.skipif(not JEDI_010,
                     reason="This feature is only supported in jedy >= 0.10")
 def test_introspection(setup_editor):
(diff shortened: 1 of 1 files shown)


spyder/plugins/tests/test_editor_introspection.py
```diff
@@ -58,6 +58,7 @@ def get_spyder_pythonpath(*args):
     editor.introspector.plugin_manager.close()
 
 
+@pytest.mark.slow
 @pytest.mark.skipif(os.environ.get('CI', None) is not None,
                     reason="This test fails too much in the CI :(")
```
ccordoba12: Please remove this condition to see if it passes in our CIs because else this test wouldn't be run anywhere.

CAM-Gerlach: Ah yes, silly me. I tested it previously with #6044 since it was failing locally too, but after adding huge ``qtbot.wait()``s I got it to pass, though still not perfectly consistently. I figured that had a good chance at fixing it on the CIs, but at the cost of a minimum of 30+ seconds added to every build in the best case; if it failed (which just happened sometimes maybe 1 out of 10 on my machine) and reran 3x, that would multiply to 90, and with 3 flaky runs, (which never seemed to change the outcome and just triple the runtime when it failed, so I dropped them in #6099 ) it would be around 5 extra minutes of build time worst case, before failing anyway. That said, doesn't hurt giving it a shot...

ccordoba12:
Please rebase this PR in top of latest 3.x and skip `test_introspection` in Python 2 to avoid some strange segfaults with other tests.

jnsebgosselin:
> Is there a way to disable codecov for a specific PR? Couldn't find it...both this one and #6137 its failing, even though clearly neither add any real new executable code to the project.

I've raised the treshold of codecov from 0% to 1% before failing. But it is in https://github.com/spyder-ide/spyder/pull/6095, which has not been merged yet. Sorry for the inconvenience.

CAM-Gerlach:
> Please rebase this PR in top of latest 3.x and skip test_introspection in Python 2 to avoid some strange segfaults with other tests.

Done and done. However, this time the Py3.5/PyQt4 test failed in the same manner; a segfault a few tests after the introspection one (though not on the same test I believe). Maybe some kind of persistent object is sticking around causing trouble? I restarted the previous tests that failed just to see if it was reproducible, and sure enough they failed again. 

No idea what's really going on, but it seems like it fails on both Py2 and Py3, and PyQt4 and PyQt5, so not sure any build is safe. It seems this one test is just nothing but trouble...perhaps it *is* almost better to both mark it as slow, and skip it at least on Linux + CI (i.e. Travis), just let AppVeyor run it along with users if they want to (with the right flag to run slow tests, presumably).

> I've raised the treshold of codecov from 0% to 1% before failing. But it is in #6095, which has not been merged yet. Sorry for the inconvenience.

No worries. I just didn't want @ccordoba12 to think the actual tests were failing or something.

ccordoba12:
> No idea what's really going on, but it seems like it fails on both Py2 and Py3, and PyQt4 and PyQt5

Let's try one more time, but now please skip it in PyQt4. I think Travis should pass this time.

ccordoba12:
Now things pass in Travis and AppVeyor, and with a nice increase in coverage! Sweet!

spyder/plugins/tests/test_editor_introspection.py
```diff
@@ -58,8 +60,9 @@ def get_spyder_pythonpath(*args):
     editor.introspector.plugin_manager.close()
 
 
-@pytest.mark.skipif(os.environ.get('CI', None) is not None,
-                    reason="This test fails too much in the CI :(")
+@pytest.mark.slow
+@pytest.mark.skipif(PY2 or PYQT4,
```
ccordoba12: Last time: please remove `PY2` from here.

ccordoba12: I think the failures only happen in PyQt4.

CAM-Gerlach: But it segfaulted on Py2 + PyQt5 too... ?

CAM-Gerlach: I can't see it now because I rebased it, but I'm 99% sure both Python tests failed, the QtPy4 one with segfault and the QtPy5 just hung and stopped, and then I restarted the QtPy5 one and it segfaulted too. But if you're sure, I can resend it again...

CAM-Gerlach: Yup, [here](https://travis-ci.org/spyder-ide/spyder/jobs/325618685) you go. I cancelled midway but it did indeed segfault.

ccordoba12: Ok, not prob. Thanks for trying :+1:

PR #6153, opened 2018-01-06
PR: Cannot change windows layout on first start after reset

### LINKED ISSUE #6152: Not able to change layouts after resetting

After a `spyder --reset`, when opening spyder for the first time, not able to change the layout  under `View -> Windows Layout`.

1. spyder --reset
2. open spyder 
3. go into View -> Windows Layout and select another layout -OR- use the toggle keys to try to change to another layout.

I expect the layout to change, but instead nothing happens.

* Spyder Version:  3.2.6.dev0
* Python Version: 3.6.4
* Qt Version: 5.6.2
* PyQt Version: 5.6
* Operating system:  Ubuntu 16.04

## Dependencies

and paste the contents below:

csabella:
Fixes #6152.

The setup code for creating the layout configuration was saving the current layout and then overwriting the state to be None.

When this config was initially saved, it was also creating `self.defaults` in UserConfig.  However, later changing the `state` to None was not changing the value in `self.defaults`.  This was causing an issue when trying to change the layout because the defaults didn't have the same `None` value that the non-defaults had.

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -1410,7 +1410,7 @@ class MainWindow(QMainWindow):
             self.setWindowState(Qt.WindowMaximized)
         self.setUpdatesEnabled(True)
 
-    def save_current_window_settings(self, prefix, section='main'):
+    def save_current_window_settings(self, prefix, section='main', none_state=False):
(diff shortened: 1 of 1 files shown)


csabella:
Thanks!

PR #6154, opened 2018-01-07
PR: Add troubleshooting blurb and OS version to issue template and enhance UI

CAM-Gerlach:
Essentially, just the Github issue template and auto-report format from #6137 to simplify that one, plus a few extra changes to harmonize the two somewhat divergenct templates, decrease wordiness, and minor cleanup, plus print OS version.

However, we have a significant problem: Likely similar to that described in #5247 , the length gets cut off after a certain point already with just the template and only a few lines of traceback, as despite the shortening elsewhere the overall length is still somewhat longer with the extra ~400 characters from the blurb at the top, allowing around 1300 chars total before truncating. (This was tested in Firefox 52 ESR), and in some rare cases depending on the actual length of the text, etc, the GET request fails completely, presumably because the split happened on some sort of critical char/boundary. And to be honest, this is nothing new; I see dep reports cut off all the time due to this, even if the user did everything else right.

However, I'm not sure how to fix this unless we switch to the Github API, or send the requests on to our own servers and *then* pass them on to Github...seems like we probably shouldn't merge this for the time being, though the actual EditTemplate.md is unaffected obviously. 

- [x] Rebase and refactor after #5346 and #6207 are merged
    - [x] Re-harmonize and further refine error report text
- [x] Apply error dialog text enhancements and add troubleshooting guide link, pulled from #6137 
- [x] Implement a few further basic tests
    - [x] Test that URL is actually getting called
    - [x] Test that URL parameters are set properly
    - [x] Test basic render_issue functionality
- [x] Implement further minor enhancements and text tweaks on top of #5346 as previously planned
    - [x] Fix broken reporting functionality and general code cleanup
    - [x] Check for DEBUG in addition to DEV
    - [x] Further refinements to existing placeholder, etc. text
    - [x] Other minor things mentioned in that and #6207 threads


CAM-Gerlach:
And, yet another segfault on the ``introspection`` test, this time on Py3.5 + PyQt5, as I suspected might happen. Therefore, we presumably need to disable that pesky test...*again*.

See update above; I accidentally hit enter and it submitted it too soon, with no way for me to actually just delete it.

ccordoba12:
> And, yet another segfault on the introspection test, this time on Py3.5 + PyQt5

The test I see failing is `test_auto_backend`. Did you create your branch from the latest 3.x? Because `test_auto_backend` only passes if you did that.

CAM-Gerlach:
> The test I see failing is `test_auto_backend`. Did you create your branch from the latest 3.x? Because `test_auto_backend` only passes if you did that.

Yes, I just double checked it indeed has the latest ``3.x`` commit ``ca27d3df``. which is the merge commit of my ``mark-introspection-slow`` branch—the one that I'm suggesting likely introduced the problem because it is the same flaky behavior seen there.

spyder/app/mainwindow.py
```diff
@@ -2375,40 +2376,48 @@ def report_issue(self, traceback=""):
...
+        reminder_message = (
+            "**PLEASE READ:** Before reporting here, *please* carefully "
+            "read our **Troubleshooting Guide**: <{0!s}> and search the issue "
+            "page for your error/problem, as most "
+            "posted bugs are duplicates or easy fixes.\n\n"
+            "Also, if you do post here, please describe step by step in "
+            "detail what you were doing when the error occurred. "
+            "Otherwise, your issue will likely be closed after 7 days. "
+            "Thanks.\n\n"
+            "**Note:** Delete this message before submitting."
```
ccordoba12: Does this part go to the text published on Github when users press the *Submit to Github* button in our report errors dialog?

CAM-Gerlach: Yes, so the apparent majority of users who are sending auto-generated error reports, or error reports from the Help menu, which appears to be the dominant source of "drive by" traceback only reports (rather than navigating to Github and hitting the "New Issue" button, where that text is filled from the .github issue template) will see it _before_ they actually hit the submit their user-side/duplicate/incomplete issue and can get help faster. Or, if they don't remove it, we know they didn't even bother to read at it (as if they did, they would have seen the note saying it should be removed), giving stronger cause to summarily close it if desired.

Therein lies the problem, as mentioned above—due to the issue discussed in #5247 , at least in Firefox 52 ESR error reports are getting cut off around or below (due to % encoding) ~1300 real characters, and after editing it and the error template itself way down, perhaps more than you're comfortable with, that leaves around 100 characters of buffer after the dependencies are included. This should be okay for now for reports from the help menu, as they don't ever include more characters than this, and there's room for 2 or 3 more deps if need be, and to refine the format a bit more (maybe adding back a few words/headings here and there). Also, the change to the actual issue template is a non-issue since its done from Github, not Spyder.

However, this is a yuge problem for the autogenerated reports in many browsers, because it only leaves room for the shortest of tracebacks, and anything beyond that starts cutting off the dependencies, or (if long enough) the traceback itself. While this is already a problem (see #5247 ), this would exacerbate it.

Therefore, I suggest for now:

* Updating the Github issue template with the blurb, (currently implemented with this PR) as it is not affected
* Slimming down the existing basic structure of the Spyder-sent report where possible (currently implemented, and I can apply a few more tweaks to save a few more chars), to help a little more with traceback cutoff for now
* Not including the disclaimer (or maybe a one-line one for now) if a traceback is included, to leave room for the latter (will be implemented in this PR momentarily), otherwise, including it for now (implemented here)
* Including a interstitial dialog displaying the blurb if "Send to Github" is pressed from the error dialog, as originally proposed (future PR), or even better just integrating the text with the dialog from #5346 (that other PR, or direct followup)
* Some better alternative to the current "send the full text of the template in a GET request through the browser" method, that either just sends the traceback and the dep versions, maybe encoded in a char saving format, and then adds the static parts later...somehow...or use the Github API, or send the former to a Snyder-team-controlled webpage that adds the latter, prompts the user for the description, and sends that to Github by a non-truncated means...but maybe that's complicating things. (In the future, and probably not by me, but hopefully soon-ish)

In any case, I will submit a new commit soon with the immediate changes mentioned above.

CAM-Gerlach:
Also, is there a way we can just include it in this header?

image

ccordoba12:
> Also, is there a way we can just include it in this header?

Don't know, I guess you need to ask Github staff for that.

spyder/app/mainwindow.py
ccordoba12: > Yes

Ok, so this needs PR #5346 to be merged first because @rlaverde discovered that many tracebacks are incomplete because it seems people are sending reports with Internet Explorer, which has a limit of 2000 chars in its URL bar.

> Including a interstitial dialog displaying the blurb if "Send to Github"

Good idea. Please get the branch in PR #5346 and open a new PR with it and your new changes.

> Some better alternative to the current "send the full text of the template in a GET request through the browser" method

Our idea is to force users to paste the contents of the error contained in the dialog into Github, instead of sending the whole error to the browser.

CAM-Gerlach: > Ok, so this needs PR #5346 to be merged first because @rlaverde discovered that many tracebacks are incomplete because it seems people are sending reports with Internet Explorer, which has a limit of 2000 chars in its URL bar.

Well, the original PR yes, eventually to accomplish all the above bullets, but not as it now stands. As discussed at (perhaps too much) length in my previous comment, I've revised what I have here to take that into account, so it is a flat improvement on the current situation for the time being:

* Adds the full error blurb to the semi-automated and manual Github reports with enough chars   to spare for 3 additional dependencies even on the dev version (with a longer Spyder version code + SHA1), 
* Has 46 more chars for tracebacks (vs. the current ``3.x``) with a one-liner version of the blurb
* Includes an additional data point (OS Version) and refined formatting)
* Also helps mitigate the cut-off problem by a simple implementation of

> Our idea is to force users to paste the contents of the error contained in the dialog into Github, instead of sending the whole error to the browser

And, crucially for getting merged promptly (ideally, for ``3.2.6``), it accomplishes this this without actually changing any existing UI text or program functionality, just the text/formatting of the error reports.

As mentioned, from my testing it appears this is also a problem on (at least) Firefox 52 ESR and Chrome on Windows 8.1 (which should natively support it); so it likely happens with **all browsers on Windows**. It seems _many_ reports are cut off right now, especially the dependencies. Therefore, a `3.2.6` merge on this would help mitigate this until `3.2.7` or `3.2.8`, if that's still possible.

CAM-Gerlach:
@ccordoba12 Pushed a new version, as referenced in my latest reply in the review comment chain above, that implements what I discussed in my comment before your latest reply (and should also help address your concerns in the followup). Here are some screenshots:

### Report Issue from Help Menu
**Current 3.x**
image
**This PR**
image

**Send to Github Error Report**
**3.x**
image
**This PR**
image

spyder/app/mainwindow.py
```diff
@@ -2375,40 +2376,75 @@ def report_issue(self, traceback=""):
...
+12345678001234567800123456780012345678001234567800123456780012345678001234567800
+12345678001234567800123456780012345678001234567800123456780012345678001234567800
+12345678001234567800123456780012345678001234567800123456780012345678001234567800
+12345678001234567800123456780012345678001234567800123456780012345678001234567800
+12345678001234567800123456780012345678001234567800123456780012345678001234567800
+12345678001234567800123456780012345678001234567800123456780012345678001234567800
+12345678001234567800123456780012345678001234567800123456780012345678001234567800
+12345678001234567800123456780012345678001234567800123456780012345678001234567800
+12345678001234567800123456780012345678001234567800123456780012345678001234567800
+12345678001234567800123456780012345678001234567800123456780012345678001234567800
```
ccordoba12: What is this big block doing here? Is it an image or something?

ccordoba12: More questions:

* Is this the SHA1 string you were talking about above?
* Could you better explain why this is needed?
* Could you move it to the beginning of this file as a constant? It could be called `SHA1_REPORT_ERROR` or something like that.
* Since this constant seems to be the same string repeated several times, please generate it using the multiplication properties of strings in Python, i.e. `SHA1_REPORT_ERROR = initial_string * 20`.

spyder/app/mainwindow.py
ccordoba12: Please don't write so long comments on the revision thread but on the comments of the PR itself.

spyder/app/mainwindow.py
```diff
@@ -2375,40 +2376,75 @@ def report_issue(self, traceback=""):
...
+            ).format(__trouble_url__)
+        if traceback:
+            reminder_message = ("IMPORTANT: Read troubleshooting guide at "
+                                "{0!s} before posting!"
+                                ).format(__trouble_url_short__)
+        else:
+            traceback = "PASTE HERE"
         issue_template = """\
-## Description
+%s
```
ccordoba12: Let's use `format` here, instead of simple string replacement because now it's really hard to understand what each replacement corresponds to. So this should be (if I'm correct) `{reminder}`, instead of `%s`.

spyder/app/mainwindow.py
```diff
@@ -2375,40 +2376,75 @@ def report_issue(self, traceback=""):
...
 2. 
 3. 
 
-**What is the expected output? What do you see instead?**
-
-
-**Please provide any additional information below**
-
+## Paste Traceback/Error Below
+(from `View > Panels > Internal Console`)
```
ccordoba12: I don't like this. The traceback should be copied automatically to the clipboard when users press *Send to Github* and they'd only need to press `Ctrl+V` or do a browser `Paste` here to paste it.

Besides, when the error dialog is active, tracebacks are not pasted to the internal console.

spyder/app/mainwindow.py
ccordoba12: Reasons for me to ask you that:

1. Your comments have very little to do with my simple question.
2. Now I have to skip this long thread to do other revisions.

ccordoba12:
> Therefore, a 3.2.6 merge on this would help mitigate this until 3.2.7

This PR still needs work, so I don't think is ready for 3.2.6. As I said before, I don't want to rush this and it'd be better to merge PR #5346 first.

@CAM-Gerlach, I really appreciate your work but we can make a new release whenever we want. I think 3.2.7 could only include these fixes and be released in three weeks at most, depending on how much work and time is put on it (I can move bugs not related to report errors improvements for 3.2.8).

spyder/app/mainwindow.py
CAM-Gerlach: I'm sorry; I was unaware of that convention. Would you like me to delete them and repost them on the main thread?

spyder/app/mainwindow.py
CAM-Gerlach: Its a total, rank, n00b mistake on my part; I must have somehow missed it despite usually being very careful to check ``git status`` and ``git diff`` every time before I commit. Its just a simple kludge for testing the maximum report length before cutoff that was never supposed to be committed, which happened despite several precautions I took to make sure I didn't screw up, as I often do. I'm sorry. Needless to say, it'll be gone in the next one.

spyder/app/mainwindow.py
CAM-Gerlach: Will do, thanks. I typically use ``format`` myself but didn't want to mess with the existing code without being instructed to or a strong reason.

spyder/app/mainwindow.py
CAM-Gerlach: Okay, I will implement that as requested. My concern is that users, following previous behavior trends, will as often as not fail to copy the traceback and thus we'll get nothing, but I suppose there is only one way to find out...

CAM-Gerlach: Or will #5346 implement it instead? If so, should to be done here (other than removing the ``(from View > ...) line)?

CAM-Gerlach:
> This PR still needs work, so I don't think it's ready for 3.2.6. As I said before, I don't want to rush this and it'd be better to merge PR #5346 first.

Absolutely, the right call for sure of course. I feel rather silly trying to push it, and I'm sorry I did so. Thanks for your understanding.

CAM-Gerlach:
Sorry for the delay; I was at the conference from 7AM through 10PM here without my laptop, though I did ask around about Spyder.

In any case, implemented all the suggestions above, plus a test and some minor refactoring/cleanup. Let me know where you'd like me to go from here, to best integrate with #5346 . However, it now does not depend on anything in that PR to function or deliver an improved user experience from the get-go, but would certainly introduce merge conflicts with that PR, which would have to be fixed by rebasing one after the other is submitted.

ccordoba12:
I think the best course of action is to finish #5346 first and then decide about any other improvements on top of that.

The idea to finish that one is that after *Submit to Github* is pressed, users are sent to a new template page, where they should be asked to paste the contents of their clipboard. Those contents will be saved by us after pressing *Submit to Github* and they will include the description of the problem, all the necessary information (Operating system, release version, etc) and the full traceback.

I am going to finish that PR myself because it's a bit tricky :-)

CAM-Gerlach:
Okay, sure thing—that makes a lot of sense.

> The idea to finish that one is that after Submit to Github is pressed, users are sent to a new template page, where they should be asked to paste the contents of their clipboard.

Oh okay, so sort of like this idea?

> send the former to a Snyder-team-controlled webpage that adds the latter, prompts the user for the description, and sends that to Github by a non-truncated means

I figured that could make sense for the longer term, but wasn't sure how realistic it was. Or do you mean something else?

CAM-Gerlach:
@ccordoba12 If you'd like me to, I can do the two tasks listed on #6207 here since they (rewording the relevant prefs option and increasing the char count) are fairly in scope, if they would help that PR (upon which the rebase of this one depends) get merged faster and take a little more work off your shoulders. There's still the unresolved question I posed on that PR (wrt providing useful functionality beyond that of the existing prefs option and irreversibly suppressing all errors for the session, and revising the text accordingly) so I certainly don't want to rush it though.

ccordoba12:
The latest release of VSCode includes this dialog to report errors:

imagen

It looks quite close to what we're trying to achieve in our release. Maybe it can serve you as inspiration @CAM-Gerlach.

CAM-Gerlach:
Thanks, that indeed looks like something good to use as inspiration. In particular, the dropdown menu would be useful, but not sure how to use that to automatically e.g. tag the post as I assume VSCode uses it for. I can also add a title field if time permits, and use it as a general guide to refine the text as need be. Just got back from my trip a few minutes ago, so I'll take a look at it (finally!) tomorrow after I work through my backlog.

CAM-Gerlach:
@ccordoba12 Just finished a marathon of a rebase; with that out of the way, aside from what will probably end up being a mostly manual cherry-pick of the error report dialog text, etc. enhancements the rest is chump change, more or less.

Just FYI, since #5346 issue reporting, both on error and from the menu, was broken since the actual statement that opened the URL was removed; I fixed that, of course. Funny enough, I had previously written a test with this PR that caught it, but initially assumed it was a false positive due to the changes. Both are now fixed and implemented.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,28 +1,42 @@
+**PLEASE READ:** Before reporting here, *please* carefully read our **Troubleshooting Guide** at <https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ> and search the issues page here for your error/problem, as most postd bugs are duplicates or easy fixes.
```
ccordoba12: I think we should enclose this paragraph in a markdown comment. I'd expect a lot of people forget to remove it.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,28 +1,42 @@
...
+
+If you do post an issue here, please describe step by step in detail what you were doing when the error occurred,. Otherwise, your issue will likely be closed after 7 days. Thanks.
```
ccordoba12: This paragraph should be in markdown comment too.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,28 +1,42 @@
...
+
+**Note:** Delete this message before submitting.
```
ccordoba12: Then we could remove this one, which is a bit misleading.

.github/ISSUE_TEMPLATE.md
```diff
@@ -1,28 +1,42 @@
...
 1. 
 2. 
 3. 
 
-**What is the expected output? What do you see instead?**
+## What is the expected output? What do you see instead?
+
+
+
+### Paste Traceback/Error Below (if applicable)
```
ccordoba12: I don't think this section should be part of the issue template because that's why we have our new error dialog.

spyder/app/mainwindow.py
```diff
@@ -2405,8 +2405,19 @@ def render_issue(self, description='', traceback=''):
...
+        # Store and format the reminder message for the troubleshooting guide
+        reminder_message = (
+            "**PLEASE READ:** Before submitting here, please carefully "
+            "consult our **Troubleshooting Guide**: <{0}> and search the "
+            "issue page for your error/problem, as most "
+            "posted bugs are duplicates or easy fixes.\n\n"
+            "If you don't find anything, enter a step-by-step description "
+            "(in English) of what led up to the problem; issue reports "
+            "without a clear way to reproduce them will be closed. Thanks.\n\n"
+            "**Note:** Please delete this message before submitting."
```
ccordoba12: As I mentioned above, this should be in a markdown comment.

ccordoba12:
@CAM-Gerlach, everything looks pretty good! I just left some minor comments :-)

CAM-Gerlach:
@ccordoba12 Thanks for all the feedback; I'll take care of it as soon as I get home from work today.

.github/ISSUE_TEMPLATE.md
CAM-Gerlach: I would otherwise do so, but my reason behind making it visible and asking the user to delete it themselves is that it serves as a canary; if it show up not-deleted on a post, its an obvious sign to both us and the user that they very likely haven't bothered to read the "Please Read" section, which states quite clearly that they should delete it, and therefore further implies that they haven't done the basic due diligence of reading the troubleshooting guide or checking for other issues first. Furthermore, making it show up on the actual post gives the user more chances to see it, both in the preview and once they've posted it. However, if you still think I should comment it, then I will of course do so; let me know.

.github/ISSUE_TEMPLATE.md
CAM-Gerlach: See above.

.github/ISSUE_TEMPLATE.md
CAM-Gerlach: See above.

spyder/app/mainwindow.py
CAM-Gerlach: See above.

.github/ISSUE_TEMPLATE.md
CAM-Gerlach: This is just for the template, for users who report manually via Github, not our automatic system (e.g. if the error occurs on Spyder startup, they can't get it to launch at all, or they temporarily/permanently disabled the error dialog). Many of those cases would likely have a traceback, or at least an error message, and this would make it clear we need it if so. Let me know if you still want me to remove it.

CAM-Gerlach:
@ccordoba12 Hey, had some feedback for your feedback. Let me know what you'd like me to do, and if needed I'll push it ASAP.

.github/ISSUE_TEMPLATE.md
ccordoba12: It seems we're split in this decision. @spyder-ide/core-developers, @spyder-ide/junior-developers, what do you think about this?

andfoy: I agree with @ccordoba12, its most likely that people will forget to remove it from their troubleshooting reports

jnsebgosselin: I would like to see a screenshot first :P

CAM-Gerlach: > I agree, its most likely that people will forget to remove it from their troubleshooting reports

As mentioned above, I don't disagree that some people will.  However, my argument is that _that's the point_. If they haven't read it carefully enough to see that they should delete it once they've read it, then what I contend (which is certainly up for debate) is that this they would also very likely have either not bothered to read the lines with the troubleshooting guide, or at least not had the similar diligence to read and follow the guide itself.

ccordoba12: @jnsebgosselin, this text would be the first of our issue template to report new bugs here on Github. It's not part of the interface itself.

CAM-Gerlach: @jnsebgosselin Just scroll up :P

csabella: If it's a Markdown comment, then that means the user will see it, but it won't show up in the bug tracker?

jnsebgosselin: Alright thanks. Because of the VSCode screenshot, I thought it was UI related. I'm going to take a look at this and think about it later tonight.

ccordoba12: > If it's a Markdown comment, then that means the user will see it, but it won't show up in the bug tracker?

Yep @csabella, that's my proposal. However, @CAM-Gerlach would prefer it to leave it as a text part of the issue report.

csabella: @ccordoba12   Thanks.  I vote for Markdown comment.    There are some people who are careful about the report they file, read everything, and would delete the comment.  But, there are others that aren't as careful (for a variety of reasons) and I don't know if blowing up the reports would really make a difference.  Yes, it may be a canary, but it's also possible they would delete it and still not have read the troubleshooting page.

However, this is also bikeshedding since we can really try it both ways and see which works better.  :-)

csabella: @CAM-Gerlach   Can you look over your proposed changes?  There are some typos.  I wasn't sure if this was still a WIP or not, so maybe you're already taking care of it.  Thanks!

ccordoba12: @csabella, this is supposed to be ready to be merged, except for this part. So please point the typos so CAM can fix them.

.github/ISSUE_TEMPLATE.md
csabella: End of sentence - occurred,.

spyder/app/mainwindow.py
```diff
@@ -2897,7 +2918,7 @@ def _check_updates_ready(self):
         latest_release = self.worker_updates.latest_release
         error_msg = self.worker_updates.error
 
-        url_r = 'https://github.com/spyder-ide/spyder/releases'
+        url_r = __project_url__ + '/releases'
```
csabella: Format string?

spyder/app/mainwindow.py
```diff
@@ -2459,7 +2477,7 @@ def report_issue(self, body=None, title=None):
         if body is None:
             body = self.render_issue()
 
-        url = QUrl("https://github.com/spyder-ide/spyder/issues/new")
+        url = QUrl(__project_url__ + '/issues/new')
```
csabella: This may be a little nitpicky, but since string concatenation isn't encouraged, maybe use format strings?

I'm getting spoiled by 3.6, `url = QUrl(f'{__project_url__}/issues/new')`, even though Spyder can't use that yet.

.github/ISSUE_TEMPLATE.md
CAM-Gerlach: @csabella Thanks for going through it...no spellchecker in Spyder and while I'm generally good at finding errors in others' prose, I make plenty of mistakes myself and don't always have the detachment from my own work to notice them right away.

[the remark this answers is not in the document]
CAM-Gerlach: Fixed, thanks,

.github/ISSUE_TEMPLATE.md
CAM-Gerlach: Fixed, thanks,

spyder/app/mainwindow.py
CAM-Gerlach: Sure CAM do, but any particular advantage in this context (one variable string, one fixed string) to offset the lesser code clarity stemming from loss of sequentiality and greater complexity?

spyder/app/mainwindow.py
CAM-Gerlach: > This may be a little nitpicky, but since string concatenation isn't encouraged, maybe use format strings?

Right, but isn't that just for concatenating more than two strings, and even then as a performance benefit on the order of microseconds unless you're concatenating a ton of strings or doing it in a loop? For two strings, both my own experience and what I've read has indicated a simple + is both clearest and just as fast.

> I'm getting spoiled by 3.6, `url = QUrl(f'{__project_url__}/issues/new')`, even though Spyder can't use that yet.

Same...

.github/ISSUE_TEMPLATE.md
jnsebgosselin: I vote for enclosing in markdown. Less work for everyone.

jnsebgosselin: I would also close incomplete reports right off the bat. Such reports never get additional info anyway.

ccordoba12: @CAM-Gerlach, I think the issue is settled now. Please enclose the necessary text in a markdown comment, so we can merge this one ASAP.

.github/ISSUE_TEMPLATE.md
ccordoba12: Ok, let's leave this section for now.

.github/ISSUE_TEMPLATE.md
CAM-Gerlach: @ccordoba12 Yes indeed, I've done so (and a few other fixes), and final CI tests are running now.

CAM-Gerlach:
@ccordoba12 The Travis PyQt5 builds kept segfaulting at ``test_get_syspath`` but eventually succeeded on the third try, and the Py3.5 build had one of the usual random ``test_calltip`` failures but I restarted it and it worked fine. 

In any case, thanks all for your feedback. I Markdown-commented the message and removed the delete line, as agreed. I also made a few other similar instructions in the template into markdown comments, and uncommented a few lines that were inside backticks (and thus wouldn't work anyway), along with fixing a few inconsistencies/issues in the text and formatting between the various versions and the typos @csabella pointed out.

Screenshots:

Slightly revised error reporting dialog:

image

Github web error report (issue template):

image

Help menu "Report Issue" report:

image

Automatic issue report from new dialog:

image

.github/ISSUE_TEMPLATE.md
CAM-Gerlach: @jnsebgosselin I certainly wouldn't say "never", but it does seem pretty rare for those that are just a traceback and not even a descriptive title. We could just close such "driveby" issues with our usual message, and ask the OP to reopen and comment if so. However, it is only a few minutes per day to close the <kbd>status:Awaiting Followup</kbd> issues as it is, and they're easy to keep track of, so it would only save a few minutes of my time per day at most, and we should see even fewer of them once this PR and its brethren go into effect, so its really a wash IMO.

ccordoba12:
@CAM-Gerlach, is this one ready?

CAM-Gerlach:
@ccordoba12 Yes.

spyder/app/mainwindow.py
csabella: No, you're right.

csabella:
Congrats!  Great job.

CAM-Gerlach:
@csabella Thanks to you for spotting my silly typos, to @ccordoba12 for all his help, and @ccordoba12 and @rlaverde for doing most of the real work in their PRs.

PR #6158, opened 2018-01-07
PR: Skip some tests because they are failing locally or segfaulting

ccordoba12:


ccordoba12:
@CAM-Gerlach, I skipped `test_introspection` again because it was making our tests too flaky. However, I'll leave it on master because it seems to be working fine there.

@jnsebgosselin, I applied to the codecov fix you made in your last PR here. Sorry but it was badly needed ;-)

CAM-Gerlach:
@ccordoba12 Thanks; unfortunately my suspicions in #6138 turned out to be correct, at least about ``3.x``, though glad it at least works on ``master``. I was already prepping a quick PR but glad you could get it done even faster.

PR #6174, opened 2018-01-08
PR: Sort files alphabetically in the File Switcher

### LINKED ISSUE #5911: Feature Request: Sort tabs alphabetically

I see that it's been a long way to get the tabs in Spyder to become moveable: https://github.com/spyder-ide/spyder/issues/564 

However, it would be really useful if Spyder would either (a) provide a standard setting to sort them alphabetically or (b) at least save the position according to the users preferences.
When a project consists of 10 or more .py-files, at some point searching for the correct tabs becomes really unhandy...

* Spyder Version: 3.2.4
* Python Version: 2.7.14 64bits
* Qt Version: 5.6.2
* PyQt Version: 5.6
* Operating system: Windows 7 x64 Professional

JakobJakobson:
Possible way to fix #5911

### DIFF
--- a/spyder/widgets/fileswitcher.py
+++ b/spyder/widgets/fileswitcher.py
@@ -672,7 +672,7 @@ class FileSwitcher(QDialog):
         # Sort the obtained scores and populate the list widget
         self.filtered_path = []
         plugin = None
-        for result in sorted(results):
+        for result in sorted(results, key=lambda tes: tes[2]):
             index = result[1]
             path = paths[index]
             icon = icons[index]


ccordoba12:
Thanks for your contribution! @jnsebgosselin, could you review this one?

jnsebgosselin:
Hey @JakobJakobson, thank you very much for taking the time to take a look at this! I've tested your PR locally and here are some observations:

When there is more than one plugin registered to the fileswitcher (like the Editor + Notebook), it is not working as expected:

fileswitcher_sorting

So I think you'll need to find a way to sort the list returned by `self.widgets` appropriately [here](https://github.com/spyder-ide/spyder/pull/6174/files#diff-c1ea14606c18e2c1e228e0b1071dad65R681) to avoid that.

Also, you'll need to check that the correct row is selected in the `fileswitcher` when it is shown. It should be the last focused file. This is handled [here](https://github.com/spyder-ide/spyder/pull/6174/files#diff-c1ea14606c18e2c1e228e0b1071dad65R706).

spyder/widgets/fileswitcher.py
```diff
@@ -672,7 +672,7 @@ def setup_file_list(self, filter_text, current_path):
         # Sort the obtained scores and populate the list widget
         self.filtered_path = []
         plugin = None
-        for result in sorted(results):
+        for result in sorted(results, key=lambda tes: tes[2]):
```
jnsebgosselin: Please, use a parameter name that is more meaningful instead of `tes`

ccordoba12:
Closing because review was not addressed.

Issue #5813, opened 2017-11-21
TypeError: tabText(self, int) when moving plugin tabs


JulienSt:
## Description

**What steps will reproduce the problem?**

1. 
2. 
3. 

**What is the expected output? What do you see instead?**

**Please provide any additional information below**

```
  File "C:\ProgramData\Anaconda3\lib\site-packages\spyder\plugins\__init__.py", line 110, in eventFilter
    self.tab_moved(event)
  File "C:\ProgramData\Anaconda3\lib\site-packages\spyder\plugins\__init__.py", line 146, in tab_moved
    self.move_tab(from_index, to_index)
  File "C:\ProgramData\Anaconda3\lib\site-packages\spyder\plugins\__init__.py", line 158, in move_tab
    from_plugin = self._get_plugin(from_index)
  File "C:\ProgramData\Anaconda3\lib\site-packages\spyder\plugins\__init__.py", line 62, in _get_plugin
    tab_text = self.dock_tabbar.tabText(index).replace('&', '')
TypeError: tabText(self, int): argument 1 has unexpected type 'NoneType'
```

## Version and main components

* Spyder Version: 3.2.4 
* Python Version: 3.6.3
* Qt Versions: 5.6.2, PyQt5 5.6 on Windows

## Dependencies
```
pyflakes >=0.6.0 :  1.6.0 (OK)
pycodestyle >=2.3:  2.3.1 (OK)
pygments >=2.0   :  2.2.0 (OK)
pandas >=0.13.1  :  0.20.3 (OK)
numpy >=1.7      :  1.13.3 (OK)
sphinx >=0.6.6   :  1.6.3 (OK)
rope >=0.9.4     :  0.10.5 (OK)
jedi >=0.9.0     :  0.10.2 (OK)
nbconvert >=4.0  :  5.3.1 (OK)
sympy >=0

CAM-Gerlach:
Something related to naming and moving around tabs somewhere in the UI, it seems. Others may know more specifically, but you describe what you were doing around the time the problem occurred?

ccordoba12:
This happens when moving plugin tabs to reorganize them (e.g. moving the IPython console to be next to the History Log).

I really don't understand why this error happens, but we can catch it and simply pass if it shows up.

CAM-Gerlach:
@ccordoba12  Is the help wanted tag for just catching the error, or tracing it more thoroughly?

ccordoba12:
Just catching the error.

[...]

PR #6207, opened 2018-01-14
PR: Use a checkbox in error dialog to dismiss it instead of a button

ccordoba12:
Screenshot:

seleccion_001

----

TODO:

- [x] Increase character count
- [x] Reword Preferences option about showing the internal console.

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -464,10 +464,9 @@ class MainWindow(QMainWindow):
         if options.window_title is not None:
             title += ' -- ' + options.window_title
 
-        if DEBUG or PYTEST:
-            # Show errors in internal console when developing or testing.
-            CONF.set('main', 'show_internal_console_if_traceback', True)
-
+        if PYTEST:
+            # Show errors in internal console when testing.
+            CONF.set('main', 'show_internal_errors', False)
 
         self.base_title = title
         self.update_window_title()
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -94,7 +94,7 @@ DEFAULTS = [
               'use_custom_margin': True,
               'custom_margin': 0,
               'use_custom_cursor_blinking': False,
-              'show_internal_console_if_traceback': False,
+              'show_internal_errors': True,
               'check_updates_on_startup': True,
               'toolbars_visible': True,
               # Global Spyder fonts
@@ -636,7 +636,7 @@ DEFAULTS = [
 #    or if you want to *rename* options, then you need to do a MAJOR update in
 #    version, e.g. from 3.0.0 to 4.0.0
 # 3. You don't need to touch this value if you're just adding a new option
-CONF_VERSION = '42.0.0'
+CONF_VERSION = '43.0.0'
 
 # Main configuration instance
 try:
--- a/spyder/plugins/configdialog.py
+++ b/spyder/plugins/configdialog.py
@@ -830,9 +830,8 @@ class MainConfigPage(GeneralConfigPage):
                                           "Python files in an already running "
                                           "instance (Requires a restart)"))
         prompt_box = newcb(_("Prompt when exiting"), 'prompt_on_exit')
-        popup_console_box = newcb(_("Pop up internal console when internal "
-                                    "errors appear"),
-                                  'show_internal_console_if_traceback')
+        popup_console_box = newcb(_("Show internal Spyder errors to report "
+                                    "them to Github"), 'show_internal_errors')
         check_updates = newcb(_("Check for updates on startup"),
                               'check_updates_on_startup')
 
--- a/spyder/plugins/console.py
+++ b/spyder/plugins/console.py
@@ -97,7 +97,7 @@ class Console(SpyderPluginWidget):
         # Traceback MessageBox
         self.error_dlg = None
         self.error_traceback = ""
-        self.dimiss_error = False
+        self.dismiss_error = False
 
     #------ Private API --------------------------------------------------------
     def set_historylog(self, historylog):
@@ -211,7 +211,7 @@ class Console(SpyderPluginWidget):
         Show a QDialog or the internal console to warn the user.
         """
         # Skip errors without traceback or dismiss
-        if (not is_traceback and self.error_dlg is None) or self.dimiss_error:
+        if (not is_traceback and self.error_dlg is None) or self.dismiss_error:
             return
 
         if CONF.get('main', 'show_internal_console_if_traceback'):
@@ -220,19 +220,16 @@ class Console(SpyderPluginWidget):
         else:
             if self.error_dlg is None:
                 self.error_dlg = SpyderErrorDialog(self)
-
-                self.error_dlg.dimiss_btn.clicked.connect(
-                    self.dismiss_error_dlg)
+                self.error_dlg.close_btn.clicked.connect(self.close_error_dlg)
                 self.error_dlg.rejected.connect(self.remove_error_dlg)
                 self.error_dlg.details.go_to_error.connect(self.go_to_error)
-
                 self.error_dlg.show()
-
             self.error_dlg.append_traceback(text)
 
-    def dismiss_error_dlg(self):
-        """Dismiss error dialog."""
-        self.dimiss_error = True
+    def close_error_dlg(self):
+        """Close error dialog."""
+        if self.error_dlg.dismiss_box.isChecked():
+            self.dismiss_error = True
         self.error_dlg.reject()
 
     def remove_error_dlg(self):
--- a/spyder/plugins/console.py
+++ b/spyder/plugins/console.py
@@ -23,7 +23,7 @@ from qtpy.QtCore import Signal, Slot
 from qtpy.QtWidgets import QInputDialog, QLineEdit, QMenu, QVBoxLayout
 
 # Local imports
-from spyder.config.base import _, debug_print
+from spyder.config.base import _, DEV, debug_print
 from spyder.config.main import CONF
 from spyder.utils import icon_manager as ima
 from spyder.utils.environ import EnvDialog
@@ -214,10 +214,7 @@ class Console(SpyderPluginWidget):
         if (not is_traceback and self.error_dlg is None) or self.dismiss_error:
             return
 
-        if CONF.get('main', 'show_internal_console_if_traceback'):
-            self.dockwidget.show()
-            self.dockwidget.raise_()
-        else:
+        if CONF.get('main', 'show_internal_errors'):
             if self.error_dlg is None:
                 self.error_dlg = SpyderErrorDialog(self)
                 self.error_dlg.close_btn.clicked.connect(self.close_error_dlg)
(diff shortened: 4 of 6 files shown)


ccordoba12:
The checkbox only dismiss the dialog during the session. I don't want to put a more explicative text next to the checkbox because in other languages that text and the *Enter at least 15 characters* one could overlap.

CAM-Gerlach:
Thanks, basically just like I was thinking. However, there are two changes I would consider, the former making the latter less necessary if its too much trouble:

* **Hides any following error, even ones completely unrelated to the original, with not even an indication they occurred** *Proposed solution:* Store the previous traceback in an attribute of ``Console`` at the end of the ``exception_occurred`` routine, and only hide the dialog it if previous_traceback == current_traceback and dismiss == True , since this fulfills the intended purpose (hiding repeated identical errors, while showing new ones). It might miss the small proportion of errors that are similar but not identical, or multiple different errors occurring repeatedly in a session (which could in turn be solved by fuzzy matching and keeping a list of previous errors, respectively), but if it is that serious, it the dialog could simply be disabled with the aforementioned preferences option, and again, this is a corner case of an edge case (repeated error dialog becoming annoying) that I'm not aware of actually being a significant problem in the first place, vs. hiding all error reports till the end of the session which could create one (being unable to report or even be aware of a different error).
* **No way to re-enable the dialog once dismissed, if the user encounters a different error, if they didn't report it originally and now want to, if they did it by accident, etc. leading to much greater user inconvenience in reporting error** Proposed solution: The above fixes the former case, but a preferences checkbox or button ("Show all hidden errors") would fix the rest.

Thanks for considering!

jitseniesen:
I wonder whether it would cut down on low-quality reports if we make it clearer that reporting issues is optional by changing the text at the top to something like:

"Reporting the error to the Spyder developers may help them to solve the issue. If you wish to do so, please enter below a step-by-step description ..."

CAM-Gerlach:
@jitseniesen Once this goes in I'll rebase #6154 and add the error dialog text change pulled from #6137 , which is (more or less) to that effect; more specifically, it aims to encourage useful, high quality reports whenever possible, while first guiding users to the other self-help methods to resolve any user-side, duplicate, or non-Spyder related problems. Also, IMO, I'll propse increasing the character count to 30 or more unless there is a pressing reason not to do so, as virtually any issue I can think of would take that many or more characters to describe in even a minimal level of detail, to raise the bar just a tiny bit higher against low-effort/drive by submissions without substantially discouraging anyone else. Then again, I'm not known for my brevity...

I'll certainly consider the intent behind the above comment in revising it, though IMO we should cut right to the heart of the matter and plainly (though diplomatically) emphasize both the importance of reporting detailed, confirmed issues, and in making sure issues are valid, unique and germane to the Spyder codebase, in order to best maximize our ability to help them and the Spyder community as a whole. With the character requirement, initially disabled "send to Github" button, clearer descriptive text, and more visible "Dismiss"/"Close" button, it should be more than obvious that doing so is optional, and contingent to making the description high enough quality.

jitseniesen:
@CAM-Gerlach Sorry to make you write that much text; I got a bit lost with all the Pull Requests.

CAM-Gerlach:
> I got a bit lost with all the Pull Requests.

Join the club :D 

> Sorry to make you write that much text

> Then again, I'm not known for my brevity...

ccordoba12:
> Hides any following error, even ones completely unrelated to the original, with not even an indication they occurred

If that's what users want, we can't force them to report the error that occurred first or any other errors.

> Proposed solution: Store the previous traceback in an attribute of Console at the end of the exception_occurred routine, and only hide the dialog it if previous_traceback == current_traceback and dismiss == True , since this fulfills the intended purpose (hiding repeated identical errors, while showing new ones)

I don't agree with this. It would be bad UX (User Experience) if we'd do that because the checkbox is quite clear and we need to respect users wishes. What I propose is to make the text associated with that box even clearer: *Don't show this until next restart*. @CAM-Gerlach, what do you think?

> I'll propse increasing the character count to 30 or more

Agreed. I'll increase the count.

CAM-Gerlach:
> If that's what users want, we can't force them to report the error that occurred first or any other errors.

Right. My concern is that may not be what users want (if they misinterpret the checkbox to mean for just that specific error/type of error, as it isn't clear on that point IMO), and also that they may change their mind about what they want upon getting a completely different error, and be unable to do anything about it.

> It would be bad UX (User Experience) if we'd do that because the checkbox is quite clear and we need to respect users wishes

Yes it would, under those premises, but A. I'm not sure it _is_ clear with respect to the above (as you address below), and B. I'm not sure it fits with the potential user need in the scenario you described as being the justification for it being there in the first place, namely getting the same error repeatedly during a session, since it would still handle that (already edge, and I understand to be theoretical) case, at least aside from a few small corner cases of that, Again, in cases where the user wanted to not see the dialog again for any error messages, that case is _already_ handled with the aforementioned preferences option. 

Furthermore, it could be a cause for even more user frustration if they are unable to get the dialog to reappear for any error without a full application and session restart; essentially every application I've encountered that has this UI pattern for dialogs has a way to show them again, and this is minimized if this option is limited to acting on a specific type of error, while the reversible preferences option that already exists handles the more troublesome general case of all error dialogs. Of course, as suggested previously, we'd want to make sure the text labels of both of these options to make the distinction clear.

> What I propose is to make the text associated with that box even clearer: ``Don't show this until next restart.``

Yes, certainly this would help; whatever we end up going with for the above, above all we want to make it clear to the user what they're agreeing to especially since for now its irreversible. Even higher priority than including the restart bit, though, is making clear whether it acts for only the specific error type, or all errors. Therefore, I suggest one of the two following, as appropriate to what we decide above:

``Don't show again for any error this session``
``Don't show this error again this session``

Which are 44 and 41 chars vs 36 for the above suggestion; if length is a problem for the translations, you can shorten the char counter text to

``Enter at least 30 chars"

Assuming the longest languages have appropriate abbreviations

ccordoba12:
@CAM-Gerlach, I think this is ready:

* I increased the character count to 20 (30 seems too much, StackOverflow uses 15 for comments).
* I changed the checkbox text to be clearer (see screenshot above).
* I renamed the previous option about the internal console to *Show internal Spyder errors to report them to Github*.

After everything is green, I'll merge so you can finish #6154.

CAM-Gerlach:
Thanks; I'll work on it Saturday (I'm booked solid through the rest of the day with work, meetings and other obligations, though hopefully I'll have an hour or two somewhere to finally push the #5813 fix). My current aim date to finish it and the other minor bits and pieces for 3.2.7 on my end (the aforementioned TypeError, rationalizing and harmonizing the order of the dep lists, making the about, dep, etc. dialogs non-modal and copyable as they're been complaints, and tests) is currently end of day Sunday.

> I increased the character count to 20 (30 seems too much, StackOverflow uses 15 for comments).

Right. Considering this is an OP, not a comment, and that even the above line has 95 characters, almost 5x the requirement, not so sure its really that much, but we can certainly start with it, see what we get and adjust from there...better to start too low than too high, I'd think.

> I changed the checkbox text to be clearer (see screenshot above).

Much better, IMO. I'd favor "for any error" in place of "during", as it makes it even clearer at the cost of 5 extra chars (in English), which can be recouped by removing the redundant "more" in the "characters to go" string. However, if you think this will still be a problem for translators, then I'd just cut those two anyway since neither adds to the meaning. Let me know which of the two you'd like me to do in my PR (or if there is some reason they shouldn't be removed).

> I renamed the previous option about the internal console to Show internal Spyder errors to report them to Github.

For reference, the text is

``"Show internal Spyder errors to report them to Github"``

It makes more sense than the previous, but IMO it could be clearer, and not sure its totally accurate since errors are still shown when it is unchecked, just in the Internal Console rather than the dialog. I'll propose modifying it to either:

``Show dialog when internal Spyder errors occur and offer to report them`

or

``Show report dialog when internal Spyder errors occur``

ccordoba12:
> we can certainly start with it, see what we get and adjust from there...better to start too low than too high, I'd think.

Agreed. Let's see how things go after we release 3.2.7.

> I'd favor "for any error" in place of "during"

I think that's too verbose. The current text is clear enough to me.

> not sure its totally accurate since errors are still shown when it is unchecked, just in the Internal Console rather than the dialog

I changed the logic in this PR and now the internal console is not raised and given focus if that option is off. So (i) the text of that option makes sense now and there's no need to change it as you suggest; and (ii) we avoid showing the internal console *at all*, which is good because that has been a source of confusion for a long time (we always hide the internal console at startup and now we don't need to show it to display errors).

The only situation I decided is important enough to show the internal console is when you're developing Spyder (i.e. you run it with `python bootstrap.py`) and that option is off.

CAM-Gerlach:
>> I'd favor "for any error" in place of "during"

> I think that's too verbose. The current text is clear enough to me.

Okay, fair enough. Then how about ``Hide all errors this session``? That's 8 chars shorter than the current string, but if you're concerned it won't be clear enough that it refers to internal Spder errors reported with the dialog then we can just cut ``during`` and be done with it.

> I changed the logic in this PR and now the internal console is not raised and given focus if that option is off. [...] The only situation I decided is important enough to show the internal console is when you're developing Spyder (i.e. you run it with python bootstrap.py) and that option is off. 

Ah, that would explain my confusion. I tested the change with regard to this point, but it was from a source checkout, and somehow misread ``elif DEV`` as just ``else`` in checking the code for this. I still think the text could be tweaked a little to make it less syntactically awkward/grammatically questionable, but its no longer a question of meaning.

However, what about when debug mode, i.e. ``debug_print``? Wouldn't the user be explicitly requesting any such messages then, as that is the purpose of such a mode—to get the maximum amount of information about errors? The file imports it and it was previously use in such a fashion, but isn't at least in this revision. If you agree, I can change it to ``DEV or debug_print`` in my PR, but maybe there is some reason why not.

ccordoba12:
> Hide all errors this session

I think this could even be `Hide all other errors in this session`.

CAM-Gerlach:
> Much better! I agree with this one :+1:

^_^

> I did my best, but I also felt it wasn't too well worded. Since I'm not a native speaker, I defer to your better judgment on this one.

In all honesty, your command of the written language is likely better than the average American. Its just that I happen to be a few standard deviations above the mean, have worked as a writer, writing coach and copyeditor on the side, and most of all have a pretty heavy style, such that rarely will a significant length of text written by a non-professional receive at least some tweaking under my pen...and be aware that I'm not nearly as good when doing the same to my own copy, so I'm far from immune to errors there. Also, to be fair I'm sure your Spanish is at least 10x better than mine ever was, haha.

> I haven't thought about that one, but sure, the condition can be `elif DEV or DEBUG`.

Okay, I can implement this in #6154 . Thanks!

> I think this could even be `Hide all other errors in this session`.

In that case, if you want to qualify "errors" I would suggest ``Hide all future errors this session``, as it makes the object more clear, flows a little better and is two chars shorter (the above being one longer than your original). I went back and forth between that and what I suggested above, since the extra qualifier provides marginal extra clarity at the cost of a greater degree of redundancy.

Issue #6008, opened 2017-12-18
Dataframe explorer: add column filters


Prikers:
This is an enhancement proposal, not a bug. It would be really nice if we could add filters on top of dataframe columns to enable quick filtering and exploration. Something that would like this (this is only a draft trial):

![image](https://user-images.githubusercontent.com/20270441/34121003-df903a9c-e427-11e7-846b-c86deaa0b631.png)

We could set text in the filter boxes that would filter the dataframe the very same way it does when doing `df.loc[condition]`.  If you like the idea I can give it more thoughts.

CAM-Gerlach:
Sounds like a terrific improvement! I had been pondering something along those lines as well; I find it very useful in Rstudio so bringing it to Spyder would be quite the treat.

ccordoba12:
Great idea @Prikers! Let's hope you can implement it for Spyder 4 :+1: 

One thing though: I think RStudio uses a single filter box that searches across columns. How does that sound for you?

CAM-Gerlach:
Personally, I can see advantages to both (and both could be implemented, as in Rstudio—it has both a quick search field and per-column filtering as suggested here), but, if only one was possible, would lean toward favoring the multi column approach, as it enables significantly more powerful/discriminatory search capabilities, and possibly much better search speed, at the cost of a few cases in e.g. wide dfs where one would want to search for the same string in multiple columns. 

This could be enabled and made even more powerful by a dropdown menu or radio buttons (etc.) toggling the columns between and, and or search modes. Even better, it would enable present or future implementation of quick logical/conditional search going well beyond just string matching; you could, for example, type `> 5.5` above a numeric column to filter for just results above that value, or match multiple words in a string column with wildcards or AND/OR logic; a factor could allow easy selection of one or more levels to display or exclude.

On the other hand, a quick search field is also nice, as it is easier and more predictable to reach (possibly even more so when bound to a hotkey), allows for matching in multiple columns at once (occasionally handy), and doesn't force the user to immediately remember, find and scroll to the column of the data their looking for. However, if I had to pick one, it would have to be the former...but that's just me.

zenithbb:
now is 2022, can we get the feather in version 5?

sdas1983:
When will this feature gets finally added to Spyder? Any timeline? 
Absolutely needed this feature.

sdas1983:
I just find that Qgrid is a Jupyter notebook widget that gives the functionality of filtering dataframes in GUI.
[...]

PR #6237, opened 2018-01-17
PR: Catch OutOfBoundsDatetime errors from pandas

### LINKED ISSUE #6177: Uncaught error when opening a pandas DataFrame with out of bounds timestamps
## Description

1. Create a Dataframe with list of dictionaries
2. two elements of the dictionary are type of datetime
3. Try open it in spyder via click in variable explorer
4. Throws a lot of exceptions until it crashes

  File "C:\Users\pfr\AppData\Local\Continuum\anaconda3\lib\site-packages\spyder\widgets\variableexplorer\dataframeeditor.py", line 260, in get_value
    value = self.df.iat[row, column]
  File "C:\Users\pfr\AppData\Local\Continuum\anaconda3\lib\site-packages\pandas\core\indexing.py", line 1786, in __getitem__
    return self.obj.get_value(*key, takeable=self._takeable)
  File "C:\Users\pfr\AppData\Local\Continuum\anaconda3\lib\site-packages\pandas\core\frame.py", line 1817, in get_value
    return _maybe_box_datetimelike(series._values[index])
  File "C:\Users\pfr\AppData\Local\Continuum\anaconda3\lib\site-packages\pandas\core\common.py", line 172, in _maybe_box_datetimelike
    value = tslib.Timestamp(value)
  File "pandas/_libs/tslib.pyx", line 402, in pandas._libs.tslib.Timestamp.__new__ (pandas\_libs\tslib.c:10051)
  File "pandas/_libs/tslib.pyx", line 1517, in pandas._libs.tslib.convert_to_tsobject (pandas\_libs\tslib.c:28598)
  File "pandas/_libs/tslib.pyx", line 1774, in pandas._libs.tslib._check_dts_bounds (pandas\_libs\tslib.c:32752)
pandas._libs.tslib.OutOfBoundsDatetime: Out of bounds nanosecond timestamp: 9999-01-01 00:00:00

  File "C:\Users\pfr\AppData\Local\Continuum\anaconda3\lib\site-packages\spyder\widge

Prikers:
Fixes #6177

### DIFF
--- a/spyder/widgets/variableexplorer/dataframeeditor.py
+++ b/spyder/widgets/variableexplorer/dataframeeditor.py
@@ -24,7 +24,10 @@ from qtpy.QtWidgets import (QApplication, QCheckBox, QDialogButtonBox, QDialog,
                             QHeaderView)
 
 from pandas import DataFrame, DatetimeIndex, Series
-from pandas._libs.tslib import OutOfBoundsDatetime
+try:
+    from pandas._libs.tslib import OutOfBoundsDatetime
+except ImportError:  # For pandas version < 0.20
+    from pandas.tslib import OutOfBoundsDatetime
 import numpy as np
 
 # Local imports
--- a/spyder/widgets/variableexplorer/dataframeeditor.py
+++ b/spyder/widgets/variableexplorer/dataframeeditor.py
@@ -24,6 +24,7 @@ from qtpy.QtWidgets import (QApplication, QCheckBox, QDialogButtonBox, QDialog,
                             QHeaderView)
 
 from pandas import DataFrame, DatetimeIndex, Series
+from pandas._libs.tslib import OutOfBoundsDatetime
 import numpy as np
 
 # Local imports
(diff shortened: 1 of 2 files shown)


spyder/widgets/variableexplorer/dataframeeditor.py
```diff
@@ -24,6 +24,7 @@
                             QHeaderView)
 
 from pandas import DataFrame, DatetimeIndex, Series
+from pandas._libs.tslib import OutOfBoundsDatetime
```
ccordoba12: The `pandas._libs` namespace is only available for in Pandas 0.20, but we support much older versions of Pandas.

How could we address that?

Prikers: Oh! I was not aware of that, sorry. What is the oldest version we should support? We can access it from `from pandas.tslib import OutOfBoundsDatetime` in version 0.16. I can check for even older versions. The import will then depend on the pandas version.

ccordoba12: > The import will then depend on the pandas version.
 
Yep, you can change this to

```python
try:
    from pandas._libs.tslib import OutOfBoundsDatetime
except:
    from pandas.tslib import OutOfBoundsDatetime
```

That will work for Pandas 0.13.1+, which is the oldest version we *supposedly* support.

If you think we should update that to a more recent version, feel free to do it in another PR for 3.x :-)

Prikers: Done! Thanks.
For the moment I do not see any strong reason to update the requirement for pandas version but maybe when working on #6008. Is anyone still using 0.13 somewhere? 😄

PR #6257, opened 2018-01-20
Update Python(x,y) URL

Uzume:
I noticed the link to Python(x,y) was outdated in your documentation.

### DIFF
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -14,7 +14,7 @@ Spyder is already included in these *Python Scientific Distributions*:
 
 * `Anaconda <https://www.anaconda.com/download/>`_
 * `WinPython <https://winpython.github.io/>`_
-* `Python(x,y) <https://code.google.com/p/pythonxy>`_
+* `Python(x,y) <https://python-xy.github.io/>`_
 
 You can start using it immediately after installing one of them (you only need
 to install one!).


CAM-Gerlach:
Hey, thanks for pointing that out. We actually were actually going to drop listing Python(x,y) in #6084 because I was informed it was no longer kept up to date, but perhaps that was because the URL was old? @ccordoba12 Do we want to add it back then? 

If so, its actually a couple other places as well that will need to be changed, and #6084 to be merged soon will break this one, so if we do add it back, would you mind if I just implement this and fix the other places as a small tweak to that PR to keep things simple?

Also, in turn their link to our site on the Python(x,y) page is out of date/broken as well; our new site is/will be https://spyder-ide.github.io/ but it is still very much under construction, so probably better to point people to our github here for now, I would think.

Uzume:
Yes, their website (even the "new" one) does seem rather dated and the last update was a few years back.

I am good with this being dropped and/or otherwise changes/tweaks, however, I wonder how long until #6084 is merged. I wonder if it would be interesting/prudent to update references to Python(x,y) before it is merged if it will be outstanding long.

CAM-Gerlach:
As far as I'm concerned it should be ready to go, pending final review, unless it is decided that Python(x, y) should be added back. It'll certainly make the next release (3.2.7), due out by the end of the month, as along with the troubleshooting/help/report text/UI/doc changes it is one of the central/motivating changes for releasing 3.2.7 on a compressed schedule.

PR #6286, opened 2018-01-24
PR: Install and run window manager in Travis to fix focus issues and unskip now-passing tests

CAM-Gerlach:
I've been noticing consistent problems with tests failing on Travis when focus and other GUI elements are involved. Accordingly, in #6132 @ccordoba12 suggested installing a window manager, as that fixed similar issues with CircleCI in #4151 . Therefore, this is a shot at doing the same for Travis, which would hopefully resolve a ton of frustration I'm having with a the tests in a number of recent PRs. First crack at it will just ensure the code runs without error, and then I'll try enabling tests that failed due to focus/etc. issues to see if this fixes it. However, it can be limited to ensure this PR can get merged quickly; the primary objective is to verify this is actually working and fixes at least some of the problems.

### DIFF
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -161,11 +161,11 @@ def main_window(request):
 # avoid possible timeouts in Appveyor
 @pytest.mark.use_introspection
 @flaky(max_runs=3)
-@pytest.mark.skipif(os.environ.get('CI', None) is not None or os.name == 'nt' or PY2,
-                    reason="It times out in our CIs, and apparently Windows.")
-@pytest.mark.timeout(timeout=60, method='thread')
+@pytest.mark.skipif(os.name == 'nt',
+                    reason="It times out on Windows locally and on AppVeyor.")
+@pytest.mark.timeout(timeout=45, method='thread')
 def test_calltip(main_window, qtbot):
-    """Hide the calltip in the editor when a matching ')' is found."""
+    """Test that the calltip in editor is hidden when matching ')' is found."""
     # Load test file
     text = 'a = [1,2,3]\n(max'
     main_window.editor.new(fname="test.py", text=text)
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -159,6 +159,7 @@ def main_window(request):
 #==============================================================================
 # IMPORTANT NOTE: Please leave this test to be the first one here to
 # avoid possible timeouts in Appveyor
+@pytest.mark.slow
 @pytest.mark.use_introspection
 @flaky(max_runs=3)
 @pytest.mark.skipif(os.name == 'nt',
--- a/spyder/plugins/tests/test_ipythonconsole.py
+++ b/spyder/plugins/tests/test_ipythonconsole.py
@@ -456,6 +456,7 @@ def test_read_stderr(ipyconsole, qtbot):
     assert content == client._read_stderr()
 
 
+@pytest.mark.slow
 @flaky(max_runs=10)
 @pytest.mark.no_xvfb
 @pytest.mark.skipif(os.environ.get('CI', None) is not None and os.name == 'nt',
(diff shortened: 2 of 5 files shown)


CAM-Gerlach:
@ccordoba12 The great news is that after implementing this, testing, and fiddling around a bit, I not only got all three of my previously skipped focus/"realistic" UI tests passing consistently on all builds (but one), but also got ``test_values_dbg`` and the infamous ``test_calltip`` working just fine without any timeout problems on both Travis and Python 2, with zero failures (either ``flaky`` or full suite reruns) across the full suite of builds (again, with one exception) during 2-3 full run-throughs. In total, this increases coverage by almost a full 1%. Also, I did a bit of associated minor cleanup and timeout refinements.

However, the one remaining issue is that the Python 3.5 PyQt5 build consistently segfaults every time as soon as the test collection phase begins before any tests are collected, which started as soon as I added the few lines to load the window manager (before unskipping any tests or making any other changes); I can only assume it is no accident that this build is the one with different setup procedures, using ``pip`` instead of ``conda`` to install packages, though I'm not sure exactly why this would make it segfault immediately at the start of test collection, before even really doing anything. Any insight? It would be a rather kludgy and awkward to maintain "fix" to disable the window manager and the now-failing-again tests on just this build, so hopefully there's a better solution. Thanks!

spyder/widgets/tests/test_editor.py
```diff
@@ -120,7 +121,7 @@ def test_find_number_matches(qtbot):
 
 def test_move_current_line_up(editor_bot):
     editor_stack, editor, qtbot = editor_bot
-
```
ccordoba12: Please revert all blank removals you introduced in this file.

spyder/app/tests/test_mainwindow.py
```diff
@@ -161,11 +161,11 @@ def close_window():
 # avoid possible timeouts in Appveyor
 @pytest.mark.use_introspection
 @flaky(max_runs=3)
-@pytest.mark.skipif(os.environ.get('CI', None) is not None or os.name == 'nt' or PY2,
-                    reason="It times out in our CIs, and apparently Windows.")
-@pytest.mark.timeout(timeout=60, method='thread')
+@pytest.mark.skipif(os.name == 'nt',
+                    reason="It times out on Windows locally and on AppVeyor.")
+@pytest.mark.timeout(timeout=45, method='thread')
```
ccordoba12: Please add `pytest.mark.slow`

spyder/plugins/tests/test_ipythonconsole.py
```diff
@@ -458,8 +458,9 @@ def test_read_stderr(ipyconsole, qtbot):
 
 @flaky(max_runs=10)
 @pytest.mark.no_xvfb
-@pytest.mark.skipif(os.environ.get('CI', None) is not None,
-                    reason="It times out in our CIs")
+@pytest.mark.skipif(os.environ.get('CI', None) is not None and os.name == 'nt',
+                    reason="It times out on AppVeyor.")
+@pytest.mark.timeout(timeout=20, method='thread')
```
ccordoba12: Please add `pytest.mark.slow`

ccordoba12:
> In total, this increases coverage by almost a full 1%. Also, I did a bit of associated minor cleanup and timeout refinements.

Wow, this is great @CAM-Gerlach! Thanks a lot!

> However, the one remaining issue is that the Python 3.5 PyQt5 build consistently segfaults every time as soon as the test collection phase begins before any tests are collected

Yep, I fixed that already in PR #6289. So please address my comment, do a rebase to pick up #6289 and then this one will be ready ;-)

spyder/widgets/tests/test_editor.py
CAM-Gerlach: Oops, accidentally left "remove trailing whitespace when saving" on...would be nice to have a toggle somewhere in the UI to make it less of a pain to be going in an out of preferences to keep changing/verifying it. Fortunately, it happened to be the last thing in my undo history so it has easy to revert.

spyder/app/tests/test_mainwindow.py
CAM-Gerlach: Yup, was thinking the same.

spyder/plugins/tests/test_ipythonconsole.py
CAM-Gerlach: Yup.

CAM-Gerlach:
@ccordoba12 Thanks! Should be good to go now.

PR #6291, opened 2018-01-24
Better browse_history

impact27:
Prevent browse_history from returning the same text

### DIFF
--- a/spyder/widgets/mixins.py
+++ b/spyder/widgets/mixins.py
@@ -787,12 +787,12 @@ class BrowseHistoryMixin(object):
         if self.is_cursor_before('eol') and self.hist_wholeline:
             self.hist_wholeline = False
         tocursor = self.get_current_line_to_cursor()
-        
+
         if self.hist_wholeline:
             start_text = self.get_text(self.current_prompt_pos, 'eol')
         else:
             start_text = self.get_text('cursor', 'eol')
-        
+
         text = start_text
         while text == start_text and self.histidx != 0:
             text, self.histidx = self.find_in_history(tocursor, self.histidx,
--- a/spyder/widgets/mixins.py
+++ b/spyder/widgets/mixins.py
@@ -787,8 +787,16 @@ class BrowseHistoryMixin(object):
         if self.is_cursor_before('eol') and self.hist_wholeline:
             self.hist_wholeline = False
         tocursor = self.get_current_line_to_cursor()
-        text, self.histidx = self.find_in_history(tocursor, self.histidx,
-                                                  backward)
+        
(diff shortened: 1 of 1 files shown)


impact27:
Why do I get an email from PEP8 Speaks complaining about W293 blank line contains whitespace if the comment above doesn't say anything?

csabella:
@impact27   Is there an open issue for this? If not, could you explain what your fix does?  I apologize if it's a dumb question, but I don't understand how to recreate it based on the description 'Prevent browse_history from returning the same text'.  Thanks!

impact27:
Sorry, I opened  #6295

CAM-Gerlach:
Since this is a such a small change, is a pretty big bang for your buck improvement in quality of life (at least in your case, and mine as well), and arguably fixes an annoying "bug" at least in the UX rather than adding anything truly new, is there a particular reason why it should be based against ``master`` vs. ``3.x``, given that if the former neither you nor anyone will see it for at least a year?

CAM-Gerlach:
But seriously, @impact27 thanks for doing this! I had it deep in the queue to be completed someday as it frustrated me too, but certainly wouldn't have gotten to it for a while if it wasn't for your help.

csabella:
@impact27   Thanks for opening the issue with the explanation.  It makes sense now.  :-)

csabella:
@impact27   I tested this out locally and it's crashing my system on the down key, so with the current patch, that needs to be looked at.

However, I'm wondering if it would be better to just not add duplicates to the history at all instead of skipping ones that are the same.

impact27:
That makes sense but I am unsure if the history is useful somewhere else. It is not really a history if you don't record everything?
Also should the pdb commands be saved at all? (u, d, s, n, ...) 
I am not sure if it is not something different conceptually (pdb command vs python code).
I can see it crashing too, I forgot to check if the end of history was reached (I only have the beginning). I will do that soon.

csabella:
@impact27 In the non-debugging history of the console itself, consecutive duplicates aren't stored, so changing the debugging history would make it consistent with that.

For example, if you type the following into the console:
```
a = 5
b = 10
a = 5
a = 5
a = 5
```
and then look at the History Log tab, you'll see:
```
## ---(Thu Jan 25 08:18:15 2018)---
a = 5
b = 10
a = 5
```

So, it stored `a = 5` twice because they were non-consecutive, but the other `a = 5` after that were not recorded.  I think it would be reasonable for the debugging history to do the same.

impact27:
I opened #6301 which prevents a same line to be added to history

ccordoba12:
Thanks @impact27! Should we close this one then?

PR #6301, opened 2018-01-26
PR: Don't add repeated entries to history while on pdb

### LINKED ISSUE #6295: pdb history doesn't collapse similar text

1. debug
2. enter something interesting
3. press 'n' 100 times
4. press 'up' 101 times to get back to the interesting stuff

I don't see the point of 'up' returning the same text 100 times.

impact27:
Fixes #6295

### DIFF
--- a/spyder/widgets/ipythonconsole/debugging.py
+++ b/spyder/widgets/ipythonconsole/debugging.py
@@ -73,7 +73,7 @@ class DebuggingWidget(RichJupyterWidget):
 
         def callback(line):
             # Save history to browse it later
-            if not (len(self._control.history) > 0 
+            if not (len(self._control.history) > 0
                     and self._control.history[-1] == line):
                 self._control.history.append(line)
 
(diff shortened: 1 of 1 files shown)


csabella:
Sorry, this got merged before I could ask a question.  I noticed when using ipdb that pressing Enter repeats the last command (which I think is actually kind of cool).  I wonder if these blank `lines` should be skipped from being added to history.

impact27:
Well my point of view is that even the pub commands should not go in the history. But this might be another issue. I actually added a ‘if len(line) > 1’ to my personal spyder to get rid of the single character commands. But it is a ugly hack so I haven’t pushed it.

csabella:
Actually, you wouldn't need all that.
```
if (line and
   (not self._control.history or not self._control.history[-1] == line)):
```

I don't think they go into the real history, do they?

impact27:
Sure but my point was that I also wanted to get rid of the ‘n, s, I, d’ that are specific to PDB, and only keep the actual python code.

csabella:
@impact27   Ah, I see.  Thanks for clarifying.  Since the change narrows it down to just one copy of those commands, I don't find that they are in the way at all. But, I see your point.  I guess you would need to have a set of them or look to see how the program handles them now to know the difference between a single character variable name and a pdb command.

PR #6333, opened 2018-01-30
PR: Fix and catch typeerror thrown when from_index is None while moving plugin tab

### LINKED ISSUE #5813: TypeError: tabText(self, int) when moving plugin tabs

CAM-Gerlach:
Fixes #5813 .

We're not sure why this happens, but this PR adds several protections aimed at both avoiding its occurrence and catching it just in case it occurs anyway. Also added a simple unit and an integration test to check this, though they may not be perfect as they can't really get to the root cause, and fixed a few issues with a recently added test that was causing random test failures (and re-enabled it on the CIs, since it should run now that I added the window manager).

Wasn't exactly sure where to best catch the errors, so please comment on that decision. Thanks!

### DIFF
--- a/spyder/plugins/tests/test_init.py
+++ b/spyder/plugins/tests/test_init.py
@@ -56,6 +56,26 @@ def main_window(request):
 # =============================================================================
 # Tests
 # =============================================================================
+def test_tabfilter_typeerror_simple():
+    """Test for #5813 ; event filter handles None indicies when moving tabs."""
+    MockEvent = MagicMock()
+    MockEvent.return_value.type.return_value = QEvent.MouseMove
+    MockEvent.return_value.pos.return_value = 0
+    mockEvent_instance = MockEvent()
+
+    MockTabBar = MagicMock()
+    MockTabBar.return_value.tabAt.return_value = 0
+    mockTabBar_instance = MockTabBar()
+
+    test_tabfilter = TabFilter(mockTabBar_instance, main_window)
+    test_tabfilter.from_index = None
+    test_tabfilter.moving = True
+
+    assert test_tabfilter.eventFilter(None, mockEvent_instance)
+    mockEvent_instance.pos.assert_called_once()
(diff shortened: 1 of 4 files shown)


CAM-Gerlach:
@ccordoba12 Good to go on my end; awaiting your review and feedback. Thanks!

spyder/plugins/tests/test_init.py
```diff
@@ -0,0 +1,98 @@
...
+
+    # Start the window
+    window = start.main()
+
+    # Teardown
+    def close_window():
+        window.close()
+    request.addfinalizer(close_window)
+
+    return window
```
ccordoba12: Please don't copy this fixture here. Instead remove it and move `test_tabfilter_typeerror_full` to `test_mainwindow.py`.

CAM-Gerlach: Thanks; makes sense. 'Tis done.

CAM-Gerlach:
@ccordoba12 Changed as requested.

Issue #6392, opened 2018-02-06
Sympy not embedding inline


martmists:
## Description of your problem
When trying to display sympy modules in the IPython console, it simply displays the repr, rather than embedding it:

**What steps will reproduce the problem?**

1. `from sympy import *` (which spyder claims to already have done but doesn't work)
2. `Integral(symbols("x"))`

**What is the expected output? What do you see instead?**

![image](https://user-images.githubusercontent.com/16361449/35856280-d6f703cc-0b35-11e8-9f77-b445b7426d33.png)

The regular IPython console (Terminal: Termite) does display it in unicode, but still not in the rendered LaTeX:

![image](https://user-images.githubusercontent.com/16361449/35856544-a174d26e-0b36-11e8-811c-31a316276b6a.png)

I'm guessing this is partially related to the IPython console claiming it instantiated variables (matplotlib, numpy, sympy) but I can't access any variables they claimed to have displayed. I cannot re-run init_session as it has already been called by Spyder.
**Please provide any additional information below**

## Versions and main components

* Spyder Version: 3.2.6
* Python Version: 3.6.4
* Qt Version: 5.10.0
* PyQt Version: 5.10
* Operating system: Arch Linux x64

## Dependencies

Please go to the menu entry `Help > Optional Dependencies` (or
`Help > Dependencies`), press the button `Copy to clipboard`
and paste the contents below:

IPython >=4.0    :  6.2.1 (OK)
cython >=0.21    :  None (NOK)
jedi >=0.9.0     :  0.11.1 (OK)
nbconvert >=4.0  :  5.3.1 (OK)
numpy >=1.7      :  1.14.0 (OK)
pandas >=0.13.1  :  None (NOK)
psutil >=0.3     :  5.4.3 (OK)
pycodestyle >=2.3:  2.3.1 (OK)
pyflakes >=0.6.0 :  1.6.0 (OK)
pygments >=2.0   :  2.2.0 (OK)
pylint >=0.25    :  1.8.2 (OK)
qtconsole >=4.2.0:  4.3.1 (OK)
rope >=0.9.4     :  0.10.7 (OK)
sphinx >=0.6.6   :  1.6.6 (OK)
sympy >=0.7.3    :  1.1.1 (OK)

ccordoba12:
Before trying to get the output of `Integral(symbols("x"))`, you need to run

```python
from sympy import init_printing()
init_printing()
```

ccordoba12:
[...]

PR #6404, opened 2018-02-07
Sympy above 1.0 may not work properly (see #6392)

martmists:


ccordoba12:
Thanks for your contribution. This needs to go against our `3.x` branch, not `master`, because `master` is for Spyder 4, which is still a year away.

So please open a new PR against `3.x`.

PR #6408, opened 2018-02-07
PR: Change share/pixmaps to share/icons in setup.py

### LINKED ISSUE #6188: Icon should be moved from share/pixmaps to share/icons
While /usr/share/pixmaps is in the icon path, ~/.local/share/pixmaps is not, so local installs (e.g. by pip install spyder3 --user) have no icon for the menu item. Moreover the spyder icon is a .png, not a .xpm, and the pixmap format (and location) is supported but (afaiu) deprecated.

I'm using spyder 3.2.6 with python 3.5.2 on Linux Mint 18.3

JakobJakobson:
I don't know if that's enough to fix #6188 ?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -79,11 +79,11 @@ def get_data_files():
     if sys.platform.startswith('linux'):
         if PY3:
             data_files = [('share/applications', ['scripts/spyder3.desktop']),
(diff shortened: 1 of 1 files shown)


ccordoba12:
I'll take a look at this and let you know.

PR #6423, opened 2018-02-09
PR: Add Ctrl+Shift+Enter to enter a new line in the Editor

### LINKED ISSUE #5466: How do I move to a new line while I'm in the middle of the current line?
It was line:

```
C.append(35|)
...........|
.......----------
.......The cursor
```

At this point, I don't want to press the ")" and then press  the enter key to start a new line. 
I just want to start a new line and move the cursor there (like Ctrl + Enter in VS Code), how should I do (I mean a shortcut key is helpful)?

The best I can do was to keep an empty line at the end of the file, and "move" to the next line, but it doesn't solve the problem

sarbot:
Fixes #5466

A keybind to start a new line without breaking the current (bevore it ends) is included in many editors. This is very handy for people using autocompletition for brackets

### DIFF
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -348,6 +348,7 @@ DEFAULTS = [
               'editor/unindent': 'Ctrl+[',
               'editor/move line up': "Alt+Up",
               'editor/move line down': "Alt+Down",
+              'editor/go to new line': "Ctrl+Shift+Return",
               'editor/go to definition': "Ctrl+G",
               'editor/toggle comment': "Ctrl+1",
               'editor/blockcomment': "Ctrl+4",
--- a/spyder/widgets/sourcecode/base.py
+++ b/spyder/widgets/sourcecode/base.py
@@ -974,6 +974,11 @@ class TextEditBaseWidget(QPlainTextEdit, BaseEditMixin):
         """Move down current line or selected text"""
         self.__move_line_or_selection(after_current_line=True)
         
+    def go_to_new_line(self):
+        """Go to the end of the current line and create a new line"""
+        self.stdkey_end(False, False)
+        self.insert_text(self.get_line_separator())
+        
     def extend_selection_to_complete_lines(self):
         """Extend current selection to complete lines"""
         cursor = self.textCursor()
--- a/spyder/widgets/sourcecode/codeeditor.py
+++ b/spyder/widgets/sourcecode/codeeditor.py
@@ -546,6 +546,8 @@ class CodeEditor(TextEditBaseWidget):
                                      name='Move line up', parent=self)
         movelinedown = config_shortcut(self.move_line_down, context='Editor',
                                        name='Move line down', parent=self)
+        gotonewline = config_shortcut(self.go_to_new_line, context='Editor',
+                                      name='Go to new line', parent=self)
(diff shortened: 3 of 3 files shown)


jnsebgosselin:
@sarbot Thanks for your contribution! Is this PR fixing https://github.com/spyder-ide/spyder/issues/5466?

jnsebgosselin:
Alright, tyvm, this is a great addition!

I've edited you OP by adding "Fixes #5466", so that this PR closes Issue #5466 when it is merged.

ccordoba12:
Great job so far @sarbot! The only thing missing is a test, as @CAM-Gerlach pointed out.

Don't worry about the failing test, it's unrelated to your work.

CAM-Gerlach:
@ccordoba12 If this gets merged without the shortcut being configurable, if I currently have my "Run Cell" shortcut set to Ctrl-Shift-Enter (just like Rstudio and some other applications use), am I right in assuming that *something* will break in my setup, either this functionality or the ability to use the proper shortcut for "Run Cell"?

@sarbot I'm not 100% sure but you seem to be on the right track; I know there's a few different places you need to add something. Why don't you give it a shot and see? You could also look at other PRs adding configurable shortcuts, like e.g. #6123 (or just search the repo for "Add shortcut for") to check.

ccordoba12:
> am I right in assuming that something will break in my setup, either this functionality or the ability to use the proper shortcut for "Run Cell"?

I don't know, probably. Could you test it locally and report back?

CAM-Gerlach:
Nvm @ccordoba12 , he added the configurable shortcut. Thanks @sarbot ! I'll test it anyway just to make sure.

CAM-Gerlach:
Okay, tested it out. Starting out, neither function works; Ctrl-Shift-Enter does nothing. However, after I go to prefs, where I'm warned about the conflicting shortcut, and change it to something unique, this function works just fine as expected (and so does run cell). Thanks!

CAM-Gerlach:
BTW, if you ``rebase`` on the latest ``3.x`` and then ``push -f``, it should fix the test failure.

> Is `ctrl+shift+Return` a spyder default, or common keybinding to run the cell or just you personal configuration? 

@sarbot Nope, its not a default, though it is for Rstudio (and possibly others) which we plan to implement as a built-in hotkey preset in the upcoming Spyder 4. Once it is reset to something unique (which is possible thanks to you setting up the configurable shortcut), everything works perfectly, though if other users employ it, it could cause some confusion at first However, this is really more of a broader UX question of whether and how to better handle shortcuts that are added in a new release that conflict with existing added ones, and isn't really the fault of this PR at all. Thanks for checking!

> I've never written tests in python before but I'm going to check the tests for other functions and try to understand it.

It shouldn't be too tough especially since we're using ``pytest``; I never had either not to long ago and it wasn't too hard to learn at all. There should be plenty of good examples to base it off, but my first guess at the overall approach would be just creating a new editor with a few lines of example text, and then testing that the shortcut has the desired effect (e.g. triggering it via qtbot.keyClick, and/or by directly calling the function).

ccordoba12:
This is nice and a test for it seems too much of a burden. Let's merge it and see how it goes.

PR #6427, opened 2018-02-10
PR: Rearrange find and replace layout and make it more adaptive to various screen widths

### LINKED ISSUE #6312: Find and replace panel prevents window snapping in Gnome 3
## Description

1. Open Spyder
1a. Project Manage opens by itselt #6292
2. Drag window to either side to snap it (or press, e.g. Win+Right)
3. Open Replace (Ctrl+R)

**What is the expected output?**
Just showing the find-replace fields

**What do you see instead?**
The windows desnaps and cannot be snapped again while the search field **and** Project Manager are there.

I think the issue is due to something like preset minimal length of the fields with buttons. I suggest changing the layout to a more intuitive layoute like this:
```
Find:    |______|
Replace: |______|
|_|buttons below|_|

```

* GNOME Shell 3.26.2
* XWAYLAND0 connected 1920x1080+0+0 (normal left inverted right x axis y axis) 480mm x 270mm
* Spyder Version: 3.2.6 
* Python Version: 3.6.4
* Qt Versions: 5.10.0, PyQt5 5.10 on Linux

## Dependencies
```
sphinx >=0.6.6   :  1.6.6 (OK)
rope >=0.9.4     :  0.10.6 (OK)
psutil >=0.3     :  5.4.3 (OK)

```

homocomputeris:
I've tried to fix #6312 by rearranging the layout and making it more adaptive to the width of the editor tab.
I expect the CI tests to fail :)

### DIFF
--- a/spyder/widgets/findreplace.py
+++ b/spyder/widgets/findreplace.py
@@ -32,7 +32,8 @@ from spyder.widgets.comboboxes import PatternComboBox
 def is_position_sup(pos1, pos2):
     """Return True is pos1 > pos2"""
     return pos1 > pos2
-    
+
+
 def is_position_inf(pos1, pos2):
     """Return True is pos1 < pos2"""
     return pos1 < pos2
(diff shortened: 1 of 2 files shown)


ccordoba12:
Thanks for your contribution @homocomputeris! Could you post screenshots of before and after your work? Thanks!

homocomputeris:
spyder_find_replace_fix

homocomputeris:
@CAM-Gerlach I can revert whitespaces and blocks but I can do nothing with PEP's line lengths and indents because in this case they are due to very long var names, which I tried to refactor but it turned out ever worse.

CAM-Gerlach:
@homocomputeris 

Why not do

```
        self.replace_button = create_toolbutton(
            self, text=_('Replace and find next'),
            icon=ima.icon('DialogApplyButton'),
            triggered=self.replace_find,
            text_beside_icon=True)
```

or even

```
        self.replace_button = (
            create_toolbutton(self, text=_('Replace and find next'),
                              icon=ima.icon('DialogApplyButton'),
                              triggered=self.replace_find,
                              text_beside_icon=True))
```

Both are PEP8 compliant (IIRC) and more correct than the current arrangement; I believe the former is probably more conventional and makes more sense.

homocomputeris:
@CAM-Gerlach Thanks for advice.

homocomputeris:
I don't understand why 3.5 build fails

CAM-Gerlach:
It was just a random failiure, apparently. I've restarted that build, and it looks like its on its way to success.

ccordoba12:
@CAM-Gerlach, I'm working on a fix for the failure in Python 3.5. It has to do with a new release of Pyzmq.

@homocomputeris, does the new rearrangement only applies when the Editor size is small enough to display all buttons next to the search and replace fields?

homocomputeris:
@ccordoba12 It always has 4 lines. The problems it solves:
- always enough space for reasonable input
- layout doesn't jitter because of search results
- allows narrow Editor tab
- thus, fits half of a lowres screen without unsnapping

It weren't pinned to Editor and its size didn't depend on tabs arrangement, it would be more reasonable to make 2 lines like in PyCharm (short input field) or SublimeText (long input field).

CAM-Gerlach:
@homocomputeris Looks like its fixed, so if you ``rebase`` against the latest ``3.x`` the failure should clear itself up.

homocomputeris:
@CAM-Gerlach The fix worked.

homocomputeris:
No worries.
Actually, I based this layout on Sublime's and PyCharms. (And the current one has 2 lines.) It seems intuitive, simple, and allows quite long search lines.

From my point of view as a user: I don't see any problem in 2-3 lines fewer in the editor because search scrolls to the next match. Neither I leave this tool open after replacing.
However, I would definitely not want the interface to change without my permission: if Search is the first row, and Replace is the second, I do not expect them to move or change size (as Search does now when `no matches` changes).

To be honest, I have no idea how to take into account all possible screen, window and tabs widths. Will these fields jump every time a user opens or resizes a tab? If Spyder desnaps on full HD screens, I don't know what should be considered _not_ narrow.
So, I'm not sure if it's worth implementing.

spyder/widgets/findreplace.py
```diff
@@ -397,8 +424,9 @@ def regexp_error_msg(pattern):
                 return str(e)
             return None
 
-        if multiline_replace_check and self.replace_widgets[0].isVisible() and \
-           len(to_text_string(self.editor.get_selected_text()).splitlines())>1:
+        if (multiline_replace_check and self.replace_widgets[0].isVisible()
+                and len(to_text_string(
+                    self.editor.get_selected_text()).splitlines())>1):
```
CAM-Gerlach: Missing whitespace around ``>`` operator here, which is one of the explicitly required ones per PEP8.

ccordoba12:
I'll take a look at how this looks when I have some time. Sorry for the delay in coming back to you @homocomputeris.

CAM-Gerlach:
@goanpeca Notepad++ also has a floating window (which offers a lot more features and functionality), but on the UI side I personally greatly prefer the built-in approach, as the floating windows tend to get in my way and become an annoyance.

Also, not sure why CircleCI is core dumping, and it says I don't have permissions to restart it even though it shows me as logged in (???).

ccordoba12:
@homocomputeris, I don't like your proposed solution. We shouldn't use two rows *by default* for find and two more for replace because there's a real waste of vertical and horizontal space on wide screens.

Instead, the design should be responsive and use two rows for each only when the width of the application is smaller than certain value (e.g. 800 px).

homocomputeris:
> a real waste of vertical and horizontal space on wide screens

It seems to me that a natural workflow supposes closing the search bar after usage.
During search in doesn't matter a lot because it scrolls to the next match.

> application is smaller than certain value (e.g. 800 px).

Let's [optimistically assume](http://gs.statcounter.com/screen-resolution-stats/desktop/worldwide) that average screen size is 1920-by-1080.
If I use quite intuitive (and default?) layout with editor on the left side and console-whatever on the right, considering scrollbars and separator, the editor's width is approx. 900 px.
In current layout, besides uneven field sizes and the Search field that stretches when results are found, the replace field is way too shorter. Well, of course, it is possible to make them even by using some cryptic icons instead of text or abbreviations like 'RafN', 'RS' etc., sacrificing usability.

Finally, adaptive layout will impose minimal width of the fields that will desnap the window on smaller screens as it does now (which was the initial reason).

ccordoba12:
> So we can potentially move forward on this, how would you suggest @homocomputeris implement such a responsive approach in our GUI?

I don't know, but this is his itch the one he's trying to scratch, not ours, so he have to find a better solution.

homocomputeris:
> I would be very surprised if the average scientist, engineer or data analyst would have a significantly larger (and quite possibly multi-monitor) screen than the average web browsing consumer user, both considering their need for it and higher-class machines in general, their higher status, and the overwhelming dominance of laptops with much smaller screens among consumers vs. workstations or at least laptops connected to additional screens for data scientists.

It is a reasonable notion.
In my experience, though, institutional hardware is usually quite old and is 1600 by 1200 in the best case. And even doesn't have DVI.
As you can see at [a marketplace catalogue](https://geizhals.eu/?m=1), the most manufactured models (and, obviously, demanded) are still, for general public, FullHD.
BTW, you can add optional telemetry to know screen resolutions exactly. This might be useful data for UX designers.

> I don't know, but this is his itch the one he's trying to scratch, not ours, so he have to find a better solution.
Might be.

homocomputeris:
> My recommendation for @homocomputeris would be reverting the layout change, which is the subject of contention

This is a nice solution.
Unfortunately, ATM I don't have much time for this, and the problem is small and doesn't bother me because I switched to PyCharm.
So I stall this.

CAM-Gerlach:
> In my experience, though, institutional hardware is usually quite old and is 1600 by 1200 in the best case.
> And even doesn't have DVI.

Yeah, I guess it depends on the institution...I must be even luckier than I thought.

> As you can see at a marketplace catalogue, the most manufactured models (and, obviously, demanded) are still, for general public, FullHD.

Oh, absolutely, and will be for a long time. 

> BTW, you can add optional telemetry to know screen resolutions exactly. This might be useful data for UX designers.

Yeah, we're indeed aware, but I'm pretty sure that won't ever happen any time soon due to privacy and trust concerns. Even a simple proposal of recording how many hits we get at our update server that Spyder already connects to, to get a rough idea of how many regular users we have (AFAIK, any estimates we have are simply guesses) was determined to be off limits.

jnsebgosselin:
I can provide some kind of flowlayout solution if needed. Below is an example. I could also limit the flowlayout to move the buttons all at once or not at all, so that we get a 2 rows or 4 rows setup only.

flowlayout_example

This doesn't fix the jittering due to the change of width of the `no match` label. I can't say that this is really bothering me to be honest...

flowlayout_example_jittering

CAM-Gerlach:
@homocomputeris @jnsebgosselin Any chance of finishing this? Thanks!

CAM-Gerlach:
This seems unlikely to get merged and accepted in its current form, and its picked up a number of merge conflicts, so closing this for now.

goanpeca:
Thoughts on this PR @ccordoba12 ?

ccordoba12:
I mostly agree with the changes here. My main objection is that the layout needs to be flexible, i.e. we should not put the buttons below the search/replace QLineEdit's when there's enough horizontal space available. Instead, we should make things work as @jnsebgosselin displayed in this https://github.com/spyder-ide/spyder/pull/6427#issuecomment-397788773 above.

@goanpeca, could you take care of this?

ccordoba12:
And this also needs to be moved to master.

jnsebgosselin:
I think I still have the code I used to make the POC with the flexible layout. I can share it if needed.

ccordoba12:
> I can share it if needed.

That'd be really helpful @jnsebgosselin!

jnsebgosselin:
You can check this out here:
https://github.com/jnsebgosselin/spyder/commit/c9b84570ed9f0403e0c2a4793528aa33ccd19e3e

CAM-Gerlach:
Fantastic! Great to see this moving again.

ccordoba12:
@steff456, please take care of this one. For this you need to

1. Add the OP's git clone as one of your git remote's, with the command:

        git remote add homocomputeris https://github.com/homocomputeris/spyder.git

2. Fetch the contents of that remote and create an associated branch in your clone, with the commands:

        git fetch homocomputeris
        git checkout fix-find-replace

3. Move that branch to master as described [here](https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md#changing-the-base-branch) and fix all the possible merge conflicts.
2. Apply @jnsebgosselin changes mentioned in his https://github.com/spyder-ide/spyder/pull/6427#issuecomment-490894381 above.
3. Verify that things are working as displayed here: https://github.com/spyder-ide/spyder/pull/6427#issuecomment-397788773

Thanks @steff456!

jnsebgosselin:
@ccordoba12 I think it would be better to wait for my work in PR#9452 to be completed and merged before working on this because this is closely related.

ccordoba12:
Would it be too much trouble for you to merge that PR with master after this is merged?

jnsebgosselin:
Yes, I think a lot of work done here would have to be undone for this to work with what I'm proposing in PR #9452

ccordoba12:
Should we close this one then?

jnsebgosselin:
It depends if what I'm proposing in PR #9452 is accepted or not. If it is, then we can close this PR and I can take care of implementing the fix to solve the related layout issue.

PR #6473, opened 2018-02-17
PR: Skip test_completions_custom_path if jedi version before 0.10.0

### LINKED ISSUE #6447: test_completions_custom_path fails with jedi 0.9.0 (Spyder 3)
The `test_completions_custom_path` fails if jedi 0.9.0 is installed and succeeds with jedi 0.10.2. This test was introduced in PR #4962 by @rlaverde to fix #4410, a bug with code completion. As I understand it, the feature that the test is written for requires jedi 0.10.0 or higher. Spyder only requires jedi 0.9.0. This affects both the `master` and the `3.x` branch.

One possible solution is to the test if jedi 0.9.x is installed. This is done with another test in the same PR.

Alternatively, we could require jedi 0.10.0. This would simplify our code base and ensure that everybody profits from the bug fix (the original bug seems to have taken quite some effort to be tracked down). For the record, jedi 0.10.0 was released February 2017 on PIP and May 2017 on conda.

We could also take the first approach for Spyder 3.x and the second approach for Spyder 4.x. Bumping the requirement in a minor release is perhaps too big a change in this case.

Suggestions?

jitseniesen:
Fixes #6447.

### DIFF
--- a/spyder/utils/introspection/tests/test_jedi_plugin.py
+++ b/spyder/utils/introspection/tests/test_jedi_plugin.py
@@ -11,8 +11,9 @@ import pytest
 import os
 import os.path as osp
 
-from spyder.utils.introspection.manager import CodeInfo
 from spyder.utils.introspection import jedi_plugin
+from spyder.utils.introspection.jedi_plugin import JEDI_010
+from spyder.utils.introspection.manager import CodeInfo
 
 try:
     import numpydoc
@@ -118,6 +119,8 @@ def test_matplotlib_fig_returns():
     assert ('add_axes', 'function') in completions
 
 
+@pytest.mark.skipif(not JEDI_010,
+                    reason="This feature is only supported in jedi >= 0.10")
 def test_completions_custom_path():
     source_code = dedent('import test_')
     completions = p.get_completions(CodeInfo('completions', source_code,


CAM-Gerlach:
As a bit of an aside, won't this silently skip with Jedi 0.11 and later? ``test_introspection`` in ``test_editor_introspection`` has the same issue. At least from a naive point of view, I'd think it would be better to fail if for some reason Jedi 0.11 broke it then skip silently in all cases, as it might go unnoticed for a while. Ergo, isn't there a better way, something like ``jedi.__version__ >= 0.10``, or at the very least making the test skip only if the Jedi version of 0.9, not if it was not 0.10?

Also @ccordoba12 not sure what's going on with the Travis PyQt5 builds; I see the same segfault the first two runs out of three on `test_get_syspath`` as I pointed out on the final CI build of the Issue Report PR, and then the third one appears to complete all the tests, codecov etc. successfully and then somehow segfault at the very end of cleaning up.

jitseniesen:
@CAM-Gerlach Thanks for your comments. I plan to remove the `JEDI_010` variable for Spyder 4 so the issue with the name will go away.

This is ready to be merged as far as I am concerned.

PR #6528, opened 2018-02-26
PR: Restore the previous wording for package versions in error report

ccordoba12:
This way it's easier to read, at least for me.

@CAM-Gerlach?

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -2442,11 +2442,11 @@ class MainWindow(QMainWindow):
 
 ## Package Versions
 
-* Spyder {spyder_version} `{commit}`
-* Python {python_version}
-* Qt {qt_version}
-* {qt_api_name} {qt_api_version}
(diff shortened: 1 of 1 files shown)


spyder/app/mainwindow.py
```diff
@@ -2442,11 +2442,11 @@ def render_issue(self, description='', traceback=''):
 
 ## Package Versions
 
-* Spyder {spyder_version} `{commit}`
```
CAM-Gerlach: Is there a reason you removed the backticks here around the commit?

ccordoba12: Because if you're not running Spyder from a git repo then `commit` is empty.

CAM-Gerlach: Oh, my fault. At some point, those were only inserted if Spyder was a dev build, but sometime between the initial version, the rebase, and the final one the logic changed. Thanks for the catch.

PR #6540, opened 2018-02-27
PR: Delay check whether we are running under pytest until run time

jitseniesen:
Currently, the test whether the environment variable `SPYDER_PYTEST` is set (which indicates whether we are running under pytest) is done at import time. However, if the tests are run using the unittest plugin, this import happens before the environment variable is actually set. Thus, this PR delays the test until run time.

With this PR and spyder-ide/spyder-unittest#114, I can run the Spyder test suite successfully within Spyder. :tada:

### DIFF
--- a/spyder/app/start.py
+++ b/spyder/app/start.py
@@ -21,8 +21,8 @@ except:
 
 # Local imports
 from spyder.app.cli_options import get_options
-from spyder.config.base import (
-        get_conf_path, running_in_mac_app, running_under_pytest)
+from spyder.config.base import (get_conf_path, running_in_mac_app,
+                                running_under_pytest)
 from spyder.config.main import CONF
 from spyder.utils.external import lockfile
 from spyder.py3compat import is_unicode
--- a/spyder/app/start.py
+++ b/spyder/app/start.py
@@ -21,7 +21,8 @@ except:
 
 # Local imports
 from spyder.app.cli_options import get_options
-from spyder.config.base import PYTEST, get_conf_path, running_in_mac_app
+from spyder.config.base import (
+        get_conf_path, running_in_mac_app, running_under_pytest)
 from spyder.config.main import CONF
 from spyder.utils.external import lockfile
 from spyder.py3compat import is_unicode
@@ -64,7 +65,7 @@ def main():
     options to the application.
     """
     # Parse command line options
-    if PYTEST:
+    if running_under_pytest():
         try:
             from unittest.mock import Mock
         except ImportError:
@@ -146,7 +147,7 @@ def main():
             # executing this script because it doesn't make
             # sense
             from spyder.app import mainwindow
-            if PYTEST:
+            if running_under_pytest():
                 return mainwindow.main()
             else:
                 mainwindow.main()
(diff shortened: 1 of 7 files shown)


jitseniesen:
The important change is the one at the top of `spyder/config/base.py`. The other changes are just renames.

I checked the time taken to evaluate `running_under_pytest()`, because I was afraid that it might be a bottleneck, but it's only one or two microsecond on my laptop.

The PEP8 complaint is not my fault.

spyder/app/start.py
```diff
@@ -21,7 +21,8 @@
 
 # Local imports
 from spyder.app.cli_options import get_options
-from spyder.config.base import PYTEST, get_conf_path, running_in_mac_app
+from spyder.config.base import (
+        get_conf_path, running_in_mac_app, running_under_pytest)
```
ccordoba12: Please change this to

```python
from spyder.config.base import (get_conf_path, running_in_mac_app,
                                running_under_pytest)
```

ccordoba12:
> With this PR and spyder-ide/spyder-unittest#114, I can run the Spyder test suite successfully within Spyder

Really cool! I didn't think this was possible, given the complexity of Spyder's test suite. Thanks a lot for working on this!

I only have one question: any reason to not make this against 3.x? Given that we still have to support that branch, I'd prefer not to deal with two different ways of checking for Pytest in our codebase.

spyder/app/start.py
ccordoba12: This is just the way we write multi-line imports in most of our codebase :-)

spyder/config/base.py
```diff
@@ -39,9 +39,14 @@
 TEST = os.environ.get('SPYDER_TEST')
 
 
-# To do some adjustments for pytest
-# This env var is defined in runtests.py
-PYTEST = os.environ.get('SPYDER_PYTEST')
+def running_under_pytest():
+    """
+    Return True if currently running under py.test.
```
CAM-Gerlach: This is trivially minor, but though while PEP 257 allows it, it seems to be convention that we put the summary one-liner on the first line, next to the ``'''``.

jitseniesen: Really? I always thought it was the other way around ... I had a quick and random look in our code base. To me, it looks like we have far too few multi-line docstrings and that there is no clear preference for either style. We may need an executive decision here.

jitseniesen:
>  any reason to not make this against 3.x? 

Only that it did not seem necessary to me and that I hadn't tested it. But I just tested it now and it works, so I will change the base.

spyder/config/base.py
goanpeca: Hmmm actually I like either


```
def func():
    """Summary."""
```

or

```
def func():
    """
    Summary.

    Something extra.
    """
```

I don't like this at all :-p

```
def func():
    """Summary.

    Something extra.
    """
```

ccordoba12: +1 for

```python
def func():
    """
    Summary.

    Some
    """
```

I like to write docstrings this way too.

ccordoba12:
> But I just tested it now and it works, so I will change the base.

Great, thanks!

Issue #5514, opened 2017-10-20
Spyder crashes after typing plt.plot( at the ipython console


evangelos1:
## Description of your problem
Typing plt.plot( at the ipython console results in Spyder crashing. If I open Spyder via a terminal, then the following message appears: Segmentation fault (core dumped). 

**What steps will reproduce the problem?**

1. Run Spyder
2. Type plt.plot( at the ipython console

**What is the expected output? What do you see instead?**
I expect to see documentation relevant to plt.plot at the help window. Instead, Spyder crashes to the desktop.

**Please provide any additional information below**

The problem does not occur when:

1. I type plt.plot( at the editor. However, the help window shows only the basic arguments of the command and displays the following message: No further help available

2. I type matplotlib.pyplot.plot( at the ipython console

3. I turn off help for the ipython console

3. I type other pyplot functions (e.g. scatter, loglog etc)

4. I run Anaconda/Spyder on macOS High Sierra

## Versions and main components

* Spyder Version: 3.2.4
* Python Version: 3.6.3 (also 2.7.14)
* Qt Version: 5.6.2
* PyQt Version: 5.6.0
* Operating system: Ubuntu 16.04 LTS

* Matplotlib: 2.1.0 and 2.0.2

## Dependencies

IPython >=4.0    :  6.1.0 (OK)
cython >=0.21    :  0.26.1 (OK)
jedi >=0.9.0     :  0.10.2 (OK)
nbconvert >=4.0  :  5.3.1 (OK)
numpy >=1.7      :  1.13.3 (OK)
pandas >=0.13.1  :  0.20.3 (OK)
psutil >=0.3     :  5.4.0 (OK)
pycodestyle >=2.3:  2.3.1 (OK)
pyflakes >=0.6.0 :  1.6.0 (OK)
pygments >=2.0   :  2.2.0 (OK)
pylint >=0.25    :  1.7.4 (OK)
qtconsole >=4.2.0:  4.3.1 (OK)
rope >=0.9.4     :  0.10.5 (OK)
sphinx >=0.6.6   :  1.6.3 (OK)
sympy >=0.7.3    :  1.1.1 (OK)

ccordoba12:
> I run Anaconda/Spyder on macOS High Sierra

This is the problem, and it's not related to Spyder. Anaconda has not updated Qt to work with High Sierra, so you'll have to wait a couple of weeks until that happens.

ccordoba12:
[...]

PR #6575, opened 2018-03-02
PR: Update deprecated Sphinx directive html_use_smartypants

### LINKED ISSUE #6570: Sphinx html_use_smartypants is deprecated as of 1.6 and removed in 1.7
http://www.sphinx-doc.org/en/stable/config.html#confval-smartquotes

This is used in spyder/utils/help/conf.py

csabella:
Fixes #6570.

http://www.sphinx-doc.org/en/stable/config.html#confval-html_use_smartypants

### DIFF
--- a/spyder/utils/help/conf.py
+++ b/spyder/utils/help/conf.py
@@ -97,10 +97,10 @@ html_static_path = ['static']
 # pages
 html_context = {}
 
-# If true, SmartyPants will be used to convert quotes and dashes to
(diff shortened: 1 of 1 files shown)


csabella:
@ccordoba12   I wonder if I should put a versioning check around this change so that #5514 doesn't come back?

ccordoba12:
> I wonder if I should put a versioning check around this change

Yes, we need to do it. I was about to mentioning it.

PR #6715, opened 2018-03-11
PR: Replace deprecated function in inline CollectionsEditor tests and add recently supported data types

CAM-Gerlach:
In the inline tests for CollectionsEditor, ``np.random.random_integers`` is called, which has been deprecated for quite some time. To avoid problems when/if it is finally removed, I replaced it with ``np.random.randint``. Also, added a number of additional datatypes which we now support in the years since the test was written, for full coverage, and did some minor cleanup/refinements to the docstrings.

### DIFF
--- a/spyder/widgets/variableexplorer/collectionseditor.py
+++ b/spyder/widgets/variableexplorer/collectionseditor.py
@@ -1456,7 +1456,7 @@ def get_test_data():
     """Create test data."""
     import numpy as np
     from spyder.pil_patch import Image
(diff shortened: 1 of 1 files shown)


CAM-Gerlach:
@ccordoba12 Looks like increasing the #6714 didn't help much, since the AppVeyor Py3.6 build failed due to a timeout on the same test, ``test_set_elapsed_time``, requiring a restart to get it to pass...

PR #6736, opened 2018-03-12
PR: Error report stopgap to make instructions harder to miss

CAM-Gerlach:
To help raise the bar slightly as to the quality of reports and avoid user mistakes, this PR bumps up the font size on the issue dialog (as its currently quite small), makes the "Paste clipboard here" text on the automatic github report even more blaring, and increases the minimum character count to 50 so users put a little more thought into their reports.

Screenshot of issue dialog with larger font:

image

### DIFF
--- a/spyder/widgets/reporterror.py
+++ b/spyder/widgets/reporterror.py
@@ -28,7 +28,7 @@ from spyder.widgets.sourcecode.base import ConsoleBaseWidget
 
 # Minimum number of characters to introduce in the description field
 # before being able to send the report to Github.
-MIN_CHARS = 20
+MIN_CHARS = 50
 
 
 class DescriptionWidget(CodeEditor):
--- a/spyder/widgets/reporterror.py
+++ b/spyder/widgets/reporterror.py
@@ -142,7 +142,7 @@ class SpyderErrorDialog(QDialog):
         self.main_label.setOpenExternalLinks(True)
         self.main_label.setWordWrap(True)
         self.main_label.setAlignment(Qt.AlignJustify)
-        self.main_label.setStyleSheet('font-size: 11pt;')
+        self.main_label.setStyleSheet('font-size: 12px;')
 
         # Field to input the description of the problem
         self.input_description = DescriptionWidget(self)
--- a/spyder/widgets/reporterror.py
+++ b/spyder/widgets/reporterror.py
@@ -142,6 +142,7 @@ class SpyderErrorDialog(QDialog):
         self.main_label.setOpenExternalLinks(True)
         self.main_label.setWordWrap(True)
         self.main_label.setAlignment(Qt.AlignJustify)
+        self.main_label.setStyleSheet('font-size: 11pt;')
 
         # Field to input the description of the problem
         self.input_description = DescriptionWidget(self)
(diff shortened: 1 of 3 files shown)


CAM-Gerlach:
Python 3.5 test_connection_to_external_kernel is failing with ``AttributeError: 'QtKernelClient' object has no attribute 'iopub_channel'``, that is almost certainly not connected to anything in this PR (since it makes no functional changes). Will try a rerun...

spyder/widgets/reporterror.py
```diff
@@ -142,6 +142,7 @@ def __init__(self, parent=None):
         self.main_label.setOpenExternalLinks(True)
         self.main_label.setWordWrap(True)
         self.main_label.setAlignment(Qt.AlignJustify)
+        self.main_label.setStyleSheet('font-size: 11pt;')
```
ccordoba12: I don't like this. Text looks too big compared to the Description field. I think you should revert it.

CAM-Gerlach: Okay, I see what you mean. I backed the size down to the size in the description field. However, its still larger than the original, which was actually way smaller than the text in the field, and so I could actually see how people could miss or gloss over it (I actually have a little trouble reading it myself even). I'll drop a screenshot; let me know what you think.

CAM-Gerlach:
Original text size:

image


Revised text size:

image

CAM-Gerlach:
Also, I noticed a silly typo in the text: "Thanks for helping us making Spyder better for everyone!" should read either "Thanks for helping us **make** Spyder better for everyone!" or "Thanks for helping us **in** making Spyder better for everyone!" (I think I'd go with the former), but I figure if I change it, even though it only affects the English translation (human translators being smart enough to not carry over the typo), I figure it'll break the link with the others if I understand how the system works correctly. Assuming that's the case, do you mind going ahead and fixing it in #6707 if it isn't already?

ccordoba12:
Now things look much better! Thanks @CAM-Gerlach!

> Python 3.5 test_connection_to_external_kernel is failing with AttributeError: 'QtKernelClient' object has no attribute 'iopub_channel'

I verified this bug was introduced by `jupyter_client` 5.2.3 (the latest release). I'll pin it to 5.2.2 for now in this PR, so we can release 3.2.8. I have no idea how to fix it.

> Assuming that's the case, do you mind going ahead and fixing it in #6707 if it isn't already?

Ok, will do.

CAM-Gerlach:
> Now things look much better! Thanks @CAM-Gerlach!

Great, glad you like it @ccordoba12 !

> I verified this bug was introduced by jupyter_client 5.2.3 (the latest release). I'll pin it to 5.2.2 for now in this PR, so we can release 3.2.8. I have no idea how to fix it.

Hopefully our users don't run into it by themselves...but if so, we'll at least know what to tell them. Will just push the pin to 5.2.2 commit to my branch or do you want me to do it?

> Ok, will do.

Thanks.

ccordoba12:
> Will just push the pin to 5.2.2 commit to my branch or would you like me to do it?

I'll do it because it has to be done in several places.

CAM-Gerlach:
Thanks!

Issue #6451, opened 2018-02-14
Setting a path containing a single quote


pcw00:
## Description
(I'm new to both Python and Spyder, so I apologise if my terminology isn't correct)
Spyder 3.2.6 is unable to pass paths for working directories that contain a single quote to the IPython interpreter. I use my uni OneDrive account which has a single quote in the path that I cannot change.

**What steps will reproduce the problem?**

1. Click on the folder icon at the top right of the title bar (Windows 10).

2. Select a path with a folder containing a single quote, e.g. 
C:\Users\xxxx\OneDrive - Queen's University Belfast\test

3. IPython console throws an error. If I copy and paste the command shown in the error message then either (i) escape the single quote or (ii) change the path string to being enclosed in double quotes, it seems to work--e.g.:
get_ipython().kernel.set_cwd(u'C:\\Users\\xxxx\\OneDrive - Queen\'s University Belfast\\test')
(N.b. there is a backslash before the single quote in "Queen's" in the line above--it seems to get stripped by the webpage "Queen\\'s")
OR
get_ipython().kernel.set_cwd(u"C:\\Users\\xxxx\\OneDrive - Queen's University Belfast\\test")

**What is the expected output? What do you see instead?**
  File "<ipython-input-4-523d9ca4caac>", line 1
    get_ipython().kernel.set_cwd(u'C:\\Users\\xxxx\\OneDrive - Queen's University Belfast\\test')
                                                                     ^
SyntaxError: invalid syntax

**Please provide any additional information below**

## Version and main components

* Spyder Version: 3.2.6 
* Python Version: 3.6.4
* Qt Versions: 5.6.2, PyQt5 5.6 on Windows

## Dependencies
```
pyflakes >=0.6.0 :  1.6.0 (OK)
pycodestyle >=2.3:  2.3.1 (OK)
pygments >=2.0   :  2.2.0 (OK)
pandas >=0.13.1  :  0.20.3 (OK)
numpy >=1.7      :  1.13.3 (OK)
sphinx >=0.6.6   :  1.6.3 (OK)
rope >=0.9.4     :  0.10.5 (OK)
jedi >=0.9.0     :  0.10.2 (OK)
nbconvert >=4.0  :  5.3.1 (OK)
sympy >=0.7.3    :  1.1.1 (OK)
cython >=0.21    :  0.26.1 (OK)
qtconsole >=4.2.0:  4.3.1 (OK)
IPython >=4.0    :  6.1.0 (OK)
[...]

PR #6750, opened 2018-03-13
PR: Enclose python exe path in double quotes

### LINKED ISSUE #6746: 'Execute in an external system terminal' doesn't work if there is spaces in the path where the python interpreter is installed in Windows
## Problem Description

I'm trying to execute a script in an external system terminal on Windows 10 64bit and it is not working.

Let's say I have this very simple script :
```python
# -*- coding: utf-8 -*-
print('test')
```
1.  Run the script with <kbd>F5</kbd> using the configuration per file shown below so that the script is executed in an external system terminal and so that I can interact with the Python console after execution.

image

I expect the script to run in an external system terminal and to be able to see the result in the Python console. Instead, the cmd windows is shown for a fraction of a second, then nothing happens.

There is no traceback/error.

## Package Versions

* Spyder: Spyder 4.0.0.dev0 
* Python: Python 3.6.4 64bits
* Qt: Qt 5.9.1
* PyQt: PyQt5 5.9
* Operating System: Windows 10 amd64

### Dependencies

```
numpy >=1.7      :  1.14.2 (OK)
sphinx >=0.6.6   :  1.7.1 (OK)
```

jnsebgosselin:
Fixes #6746

Enclose the path of the python executable in double quotes so that executing a script in an external system terminal works correctly when the path to the Python interpreter contains spaces.

### DIFF
--- a/spyder/utils/programs.py
+++ b/spyder/utils/programs.py
@@ -253,13 +253,14 @@ def get_python_args(fname, python_args, interact, debug, end_args):
 
 
 def run_python_script_in_terminal(fname, wdir, args, interact,
-                                  debug, python_args):
+                                  debug, python_args, executable=None):
     """
     Run Python script in an external system terminal.
 
     :str wdir: working directory, may be empty.
     """
-    python_exe = get_python_executable()
+    if executable is None:
+        executable = get_python_executable()
 
     # If fname or python_exe contains spaces, it can't be ran on Windows, so we
     # have to enclose them in quotes. Also wdir can come with / as os.sep, so
@@ -267,10 +268,12 @@ def run_python_script_in_terminal(fname, wdir, args, interact,
     if os.name == 'nt':
         fname = '"' + fname + '"'
         wdir = wdir.replace('/', '\\')
-        python_exe = '"' + python_exe + '"'
+        executable = '"' + executable + '"'
 
-    p_args = [python_exe]
+    p_args = [executable]
     p_args += get_python_args(fname, python_args, interact, debug, args)
+    
+    print(executable)
 
     if os.name == 'nt':
         cmd = 'start cmd.exe /c "cd %s && ' % wdir + ' '.join(p_args) + '"'
--- a/spyder/utils/programs.py
+++ b/spyder/utils/programs.py
@@ -272,8 +272,6 @@ def run_python_script_in_terminal(fname, wdir, args, interact,
 
     p_args = [executable]
     p_args += get_python_args(fname, python_args, interact, debug, args)
-    
-    print(executable)
 
     if os.name == 'nt':
         cmd = 'start cmd.exe /c "cd %s && ' % wdir + ' '.join(p_args) + '"'
--- a/spyder/utils/programs.py
+++ b/spyder/utils/programs.py
@@ -259,16 +259,19 @@ def run_python_script_in_terminal(fname, wdir, args, interact,
 
     :str wdir: working directory, may be empty.
     """
-    # If fname has spaces on it it can't be ran on Windows, so we have to
-    # enclose it in quotes. Also wdir can come with / as os.sep, so we
-    # need to take care of it
+    python_exe = get_python_executable()
+
+    # If fname or python_exe contains spaces, it can't be ran on Windows, so we
+    # have to enclose them in quotes. Also wdir can come with / as os.sep, so
+    # we need to take care of it.
     if os.name == 'nt':
         fname = '"' + fname + '"'
         wdir = wdir.replace('/', '\\')
-    
-    p_args = [get_python_executable()]
+        python_exe = '"' + python_exe + '"'
+
+    p_args = [python_exe]
     p_args += get_python_args(fname, python_args, interact, debug, args)
-    
+
     if os.name == 'nt':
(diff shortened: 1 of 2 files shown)


spyder/utils/programs.py
```diff
@@ -259,16 +259,19 @@ def run_python_script_in_terminal(fname, wdir, args, interact,
...
+
+    # If fname or python_exe contains spaces, it can't be ran on Windows, so we
+    # have to enclose them in quotes. Also wdir can come with / as os.sep, so
+    # we need to take care of it.
     if os.name == 'nt':
         fname = '"' + fname + '"'
         wdir = wdir.replace('/', '\\')
-    
-    p_args = [get_python_executable()]
+        python_exe = '"' + python_exe + '"'
```
csabella: Maybe enclose in triple quotes to not run into #6451?  I'm not sure if it's necessary.

ccordoba12: Good idea @csabella.

ccordoba12:
@jnsebgosselin, the test failures are unrelated to this. I'll fix them in #6736.

spyder/utils/programs.py
jnsebgosselin: Good point. I'll take that into consideration and push a commit with the changes.

jnsebgosselin:
Ok, after this PR, I'm aiming to be able to execute this script
```
# -*- coding: utf-8 -*-
print('test')
```
in an external system console with a Python executable located in `C:\Python36's Path Test` in a file saved in `C:\Queen's Space\let's test this.py`.

Anything else I should include?

ccordoba12:
Using utf-8 chars in the path is always a good idea (e.g. Chinese or Arabic chars).

jnsebgosselin:
@ccordoba12 Have we defined a typical string to use to test this? If not, maybe we should so that our test for this are consistent across?

ccordoba12:
There are some Chinese and Arabic chars floating around in several test files. Perhaps you could try to define them as a constant in a single place in this PR.

jnsebgosselin:
> Perhaps you could try to define them as a constant in a single place in this PR.

Ok good idea, I'll try that.

jnsebgosselin:
@ccordoba12 I'm setting the path to the interpreter in the `Python  interpreter` option as in the screenshot below.

image

It seems the IPython console is picking the right Python interpreter, since it throws an error about ipykernel and cloudpickle not being installed, which is expected since the Python env is empty.

image

But, when I'm running the script in a external console, it doesn't use the interpreter I've set in the option, it is using the Python interpreter used to run Spyder (I think).

Is it by design it is done that way? That would be more useful if the script was executed with the Python interpreter I've set in the option instead.

jnsebgosselin:
This is the function used to fetch the path to the python executable located in `spyder\utils\misc.py` :

```python
def get_python_executable():
    """Return path to Python executable"""
    executable = sys.executable.replace("pythonw.exe", "python.exe")
    if executable.endswith("spyder.exe"):
        # py2exe distribution
        executable = "python.exe"
return executable
```

So I guess it indeed return the path to the Python executable in which Spyder is running?

ccordoba12:
> So I guess it indeed return the path to the Python executable in which Spyder is running?

Yes. `sys.executable` returns the interpreter on which Spyder is running.

jnsebgosselin:
> Yes. sys.executable returns the interpreter on which Spyder is running.

But why not using the Python interpreter that is set in the preferences?

ccordoba12:
> But why not using the Python interpreter that is set in the preferences?

I have no problem with that. I think that's what most people would expect, but no one has complained about it before.

jnsebgosselin:
> I have no problem with that. I think that's what most people would expect, but no one has complained about it before.

Can I implement this in this PR? If so, should this be done against master?

ccordoba12:
> Can I implement this in this PR?

Yes, please do.

> If so, should this be done against master?

3.x is fine. I think this change is no disruptive enough to justify making it against master.

spyder/utils/programs.py
```diff
@@ -253,22 +253,28 @@ def get_python_args(fname, python_args, interact, debug, end_args):
...
         fname = '"' + fname + '"'
         wdir = wdir.replace('/', '\\')
-    
-    p_args = [get_python_executable()]
+        executable = '"' + executable + '"'
+
+    p_args = [executable]
     p_args += get_python_args(fname, python_args, interact, debug, args)
     
+    print(executable)
```
CAM-Gerlach: This is the line causing problems in the test that no print functions were used in the code. If you're printing this for debugging, ``debug_print`` is what it recommends as an alternative.

jnsebgosselin: hehe yes I know, but thanks. I always forget to remove them...

CAM-Gerlach: I figured that was the other alternative, haha—happens to me too.

ccordoba12:
@jnsebgosselin, adding a test for this is not easy, so I think this is ready.

ccordoba12:
The failure in CircleCI is unrelated to this.

jnsebgosselin:
Thanks a lot @ccordoba12 for reviewing and rounding up this PR!

ccordoba12:
No problem :+1:

Issue #2158, opened 2015-02-17
runfile with path containing apostrophes (quotes) will not work


spyder-issues-migrator:
_From [mt.bo...@hotmail.com](https://code.google.com/u/112633889236229880690/) on 2015-02-04T15:53:30Z_

Please note that reporting bugs from Spyder's help menu ("?" > "Report
issue...") will fill version details for Spyder, Python and Qt in this description for you. 

What steps will reproduce the problem?

 1. When running a file in the spyder python console, it calls runfile('file location') (if I run from the editor-ie, press F5). The problem with this is if the file address contains a folder with an apostrophe in it, it thinks the string has ended there when there is in fact more of the address.

For example:

runfile('C:/Users/username/John's Files/Python/Scripts') 

This will attempt to find a file in a folder C:/Users/username/John, which may or may necessarily exist, and regardless the file I wish to execute obviously isn't located there.

I can easily correct this myself by calling the file myself and using double quotes as opposed to single quotes. However, I prefer to run from the editor by pressing F5, and its a bit concerning to me that it defaults to using single quotes when F5 is pressed. I'd rather not have to change all my folder names just so that spyder can run files contained in them. 

What is the expected output? What do you see instead?

 I expect the file to run, and instead I get an invalid syntax error on line 1 of the console What version of the product are you using? On what operating system? spyder 2.3.1 on python 3.4 on windows 8.1 

Please provide any additional information below

.

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=2158_

spyder-issues-migrator:
_From [ccordoba12](https://code.google.com/u/ccordoba12/) on 2015-02-04T14:12:07Z_

Thanks for reporting. I think we need to either:

1. Escape apostrophes

2. Change to use double apostrophes (") in runfile

**Labels:** Cat-Editor MS-v2.3.4

spyder-issues-migrator:
_From [mt.bo...@hotmail.com](https://code.google.com/u/112633889236229880690/) on 2015-02-04T20:56:30Z_

I think what bothered me most was that the error message and debugging didn't help me pinpoint the issue whatsoever. If it had given me a message like "error: file location could not be found" instead pf just "invalid syntax" I wouldn't have spent so much time trying to figure out what was wrong. I appreciate your timely response and concern.

spyder-issues-migrator:
_From [ccordoba12](https://code.google.com/u/ccordoba12/) on 2015-02-05T07:40:06Z_

[...]

PR #6772, opened 2018-03-16
PR: Fix a bug when running a file in the IPython console with a single quote character in its name

### LINKED ISSUE #6771: Cannot runfile if the name of the file contains a `'` of `"` character in Windows
## Problem Description

Trying to run a file named `test's.py ` in the IPython console with <kbd>F5</kbd> doesn't work in Windows.

This issue needs to be fix before I can progress further on https://github.com/spyder-ide/spyder/pull/6750.

1. Create a simple python script and save it in file named it `test's.py ` or `test"s.py `
2. Run the file in the current IPython console with <kbd>F5</kbd>

I expect the file to run, but instead, there is an error message.

```python-traceback

runfile('C:/test\'s.py', wdir='C:/test')

  File "<ipython-input-3-cfd6b8e87c59>", line 1, in <module>
    runfile('C:/test\'s.py', wdir='C:/test')

  File "E:\spyder\spyder\utils\site\sitecustomize.py", line 695, in runfile
    os.chdir(wdir)

FileNotFoundError: [WinError 2] The system cannot find the file specified: 'C:/test'

```

## Package Versions

* Spyder: Spyder Spyder 3.2.8.dev0 
* Python: Python 3.6.4 64bits
* Qt: Qt 5.9.1
* PyQt: PyQt5 5.9
* Operating System: Windows 10 amd64

### Dependencies

```
sphinx >=0.6.6   :  1.7.1 (OK)

```

jnsebgosselin:
Fixes  #6771

In [spyder/plugins/editor.py](https://github.com/spyder-ide/spyder/blob/v3.2.8/spyder/plugins/editor.py), single and double quotes are escaped before running a file from the editor with:

https://github.com/spyder-ide/spyder/blob/v3.2.8/spyder/plugins/editor.py#L2380
```
# Escape single and double quotes in fname (Fixes Issue 2158)
fname = fname.replace("'", r"\'")
fname = fname.replace('"', r'\"')
```

In windows, this cause `osp.dirname(fname)` to return the wrong path if there was a `'` or `"` character in the name of the file. For example : 

```
import os.path as osp
fname="C:\test'path.py"
fname = fname.replace("'", r"\'")
osp.dirname(fname)
```

returns in Windows `'C:\test'` instead of `'C:`. Therefore, to fix this issue, fname's dirname must be saved in a temporary variable before the single or double quote are escaped.

Note:
In Windows, the following characters are not allowed for folder or file names : 
image

### DIFF
--- a/spyder/plugins/editor.py
+++ b/spyder/plugins/editor.py
@@ -2376,11 +2376,15 @@ class Editor(SpyderPluginWidget):
         if editorstack.save():
             editor = self.get_current_editor()
             fname = osp.abspath(self.get_current_filename())
-            
+
+            # Get fname's dirname before we escape the single and double
+            # quotes (Fixes Issue #6771)
+            dirname = osp.dirname(fname)
+
             # Escape single and double quotes in fname (Fixes Issue 2158)
             fname = fname.replace("'", r"\'")
             fname = fname.replace('"', r'\"')
-            
+
             runconf = get_run_configuration(fname)
             if runconf is None:
                 dialog = RunConfigOneDialog(self)
@@ -2412,7 +2416,7 @@ class Editor(SpyderPluginWidget):
             clear_namespace = runconf.clear_namespace
 
             if runconf.file_dir:
-                wdir = osp.dirname(fname)
+                wdir = dirname
             elif runconf.cw_dir:
                 wdir = ''
             elif osp.isdir(runconf.dir):


jnsebgosselin:
The test that covers this (`test_run_code`) is skipped for Windows... so that won't be very useful. But I've tested it locally and the test fails without the changes made in this PR, but pass with the changes.

spyder/app/tests/test_mainwindow.py
```diff
@@ -763,19 +773,20 @@ def test_set_new_breakpoints(main_window, qtbot):
     main_window.editor.clear_all_breakpoints()
     main_window.editor.close_file()
 
-
```
CAM-Gerlach: Why was this line removed? Two spaces between top-level functions, I thought...accident?

spyder/app/tests/test_mainwindow.py
```diff
@@ -1465,4 +1476,4 @@ def test_render_issue():
 
 
 if __name__ == "__main__":
-    pytest.main()
+    pytest.main(['-x', os.path.basename(__file__), '-v', '-rw'])
```
CAM-Gerlach: Is there is a reason this was changed, particularly adding the ``-x`` option? In general, when running locally I usually use ``-v`` (or ``-vv``) and always use ``-rw``, but almost never want ``-x`` and would have to routinely modify the ``runtests.py`` file to change this if I wanted to run the whole test suite with this.

spyder/plugins/editor.py
```diff
@@ -2376,11 +2376,16 @@ def run_file(self, debug=False):
...
-            fname = fname.replace("'", r"\'")
-            fname = fname.replace('"', r'\"')
-            
+
+            # Get fname's dirname before we escape the single and double
+            # quotes (Fixes Issue #6771)
+            dirname = osp.dirname(fname)
+
+            # Escape single and double quotes in fname and dirname
+            # (Fixes Issue 2158)
```
CAM-Gerlach: Add # sign for consistency with previous comment ( #2158 )

spyder/app/tests/test_mainwindow.py
jnsebgosselin: Running the whole suite when trying to make a single test work is not very efficient. I forgot to exclude this change when committing, I'll simply revert in another commit.

spyder/app/tests/test_mainwindow.py
jnsebgosselin: yep, ty

spyder/app/tests/test_mainwindow.py
```diff
@@ -135,6 +135,16 @@ def find_desired_tab_in_window(tab_name, window):
     return None, None
 
 
+def setup_file_test():
```
ccordoba12: Please use the `tmpdir` fixture for the location of this file.

spyder/app/tests/test_mainwindow.py
```diff
@@ -770,12 +780,14 @@ def test_set_new_breakpoints(main_window, qtbot):
 def test_run_code(main_window, qtbot):
```
ccordoba12: Please remove the `skipif` condition for Windows here.

ccordoba12:
The failure in CircleCI is unrelated to this.

jnsebgosselin:
Thanks @ccordoba12 !

PR #6790, opened 2018-03-18
PR: Add an option to control bbox_inches option when plotting inline

### LINKED ISSUE #6619: The figure layout is not respected when plotting inline in the IPython console
As explained in this [StackOverflow](https://stackoverflow.com/a/37879281/4481445) answer,  the ipython magics are enforcing `bbox_inches='tight'` by default. This means that the following code :

```python
import numpy as np

plt.close('all')

fig, ax = plt.subplots()
fig.set_size_inches(8,  5)
ax.set_position([0, 0, 1, 1])
ax.plot(np.random.rand(50),  np.random.rand(50))

plt.show()
```

results in the IPython console when plotting inline as :

<p  align="center">>

But, as explicitly specified in my code, what I want is that :

<p  align="center">>

This behavior is unacceptable for me when I'm trying to format my figures in a very specific layout for a paper and is the reason why I've never used inline plotting in the past.

 To avoid this, one has to override the default magic with :

`%config InlineBackend.print_figure_kwargs = {'bbox_inches':None}`

As discussed with @ccordoba12, this should be added in the Preferences of the IPython console, under the tab `Graphics`, in the section `Inline backend`.

References :
- https://stackoverflow.com/questions/37864735/matplotlib-and-ipython-notebook-displaying-exactly-the-figure-that-will-be-save
- http://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-matplotlib
- https://matplotlib.org/api/_as_gen/matplotlib.pyplot.savefig.html

jnsebgosselin:
Fixes #6619

image

### DIFF
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -396,7 +396,7 @@ class IPythonConsoleConfigPage(PluginConfigPage):
         backend_group.setLayout(backend_layout)
         backend_group.setEnabled(self.get_option('pylab'))
         pylab_box.toggled.connect(backend_group.setEnabled)
-        
+
         # Inline backend Group
         inline_group = QGroupBox(_("Inline backend"))
         inline_label = QLabel(_("Decide how to render the figures created by "
@@ -404,7 +404,8 @@ class IPythonConsoleConfigPage(PluginConfigPage):
         inline_label.setWordWrap(True)
         formats = (("PNG", 0), ("SVG", 1))
         format_box = self.create_combobox(_("Format:")+"   ", formats,
-                                       'pylab/inline/figure_format', default=0)
+                                          'pylab/inline/figure_format',
+                                          default=0)
         resolution_spin = self.create_spinbox(
                         _("Resolution:")+"  ", " "+_("dpi"),
                         'pylab/inline/resolution', min_=50, max_=999, step=0.1,
@@ -418,7 +419,14 @@ class IPythonConsoleConfigPage(PluginConfigPage):
                           _("Height:")+"  ", " "+_("inches"),
                           'pylab/inline/height', min_=1, max_=20, step=1,
                           tip=_("Default is 4"))
-        
+        bbox_inches_box = newcb(
+                _("Set bbox_inches to 'tight' when plotting inline"),
+                'pylab/inline/bbox_inches',
+                tip=_("Sets bbox_inches to 'tight' when plotting inline."
+                      " This can cause discrepancies between the displayed"
+                      " image and the identical image created using savefig.")
+                )
+
         inline_v_layout = QVBoxLayout()
         inline_v_layout.addWidget(inline_label)
         inline_layout = QGridLayout()
@@ -433,6 +441,8 @@ class IPythonConsoleConfigPage(PluginConfigPage):
         inline_layout.addWidget(height_spin.plabel, 4, 0)
         inline_layout.addWidget(height_spin.spinbox, 4, 1)
         inline_layout.addWidget(height_spin.slabel, 4, 2)
+        inline_layout.addWidget(bbox_inches_box, 5, 0, 1, 4)
+
         inline_h_layout = QHBoxLayout()
         inline_h_layout.addLayout(inline_layout)
         inline_h_layout.addStretch(1)
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -154,6 +154,7 @@ DEFAULTS = [
               'pylab/inline/resolution': 72,
               'pylab/inline/width': 6,
               'pylab/inline/height': 4,
+              'pylab/inline/bbox_inches': False,
               'startup/run_lines': '',
               'startup/use_run_file': False,
               'startup/run_file': '',


jnsebgosselin:
inline_bbox_inches_args

jnsebgosselin:
@ccordoba12 Since the result is a png, that's going to be a though one to test. Do you have any idea?

Also, where the test should go? In `spyder\app\tests\test_mainwindow.py` or `spyder\plugins\tests\test_ipythonconsole.py`?

ccordoba12:
> Since the result is a png, that's going to be a though one to test. Do you have any idea?

Yep, I think you need to produce a sample figure with `bbox_inches=tight` (e.g. `plt.plot(range(10))`). Then you need to compare the output generated by our console after activating the option with the previously saved sample image. For that please use this Matplotlib function:

https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/testing/compare.py#L341

Then if `assert compare_images(...)` then things are working as expected.

spyder/plugins/ipythonconsole.py
```diff
@@ -418,7 +419,14 @@ def setup_page(self):
                           _("Height:")+"  ", " "+_("inches"),
                           'pylab/inline/height', min_=1, max_=20, step=1,
                           tip=_("Default is 4"))
-        
+        bbox_inches_box = newcb(
+                _("Set bbox_inches to 'tight' when plotting inline"),
```
ccordoba12: I think this can be worded better. Something like

> Use a tight layout for inline plots

CAM-Gerlach: Yes, I like that.

spyder/plugins/ipythonconsole.py
```diff
@@ -418,7 +419,14 @@ def setup_page(self):
                           _("Height:")+"  ", " "+_("inches"),
                           'pylab/inline/height', min_=1, max_=20, step=1,
                           tip=_("Default is 4"))
-        
+        bbox_inches_box = newcb(
+                _("Set bbox_inches to 'tight' when plotting inline"),
+                'pylab/inline/bbox_inches',
+                tip=_("Set bbox_inches to 'tight' when plotting inline."
+                      " This can cause discrepancies between the displayed"
```
CAM-Gerlach: For the calltip, I suggest something like, "Sets bbox_inches to "tight" when plotting inline with matplotlib. [When enabled/disabled], can cause discrepancies between the displayed version and the image when written out with savefig."

CAM-Gerlach: Like it. Would suggest "and that created using savefig." as a cleaner ending; I wasn't happy with mine either.

jnsebgosselin: Thanks for the feedback @CAM-Gerlach , I appreciate it.

jnsebgosselin:
@ccordoba12 I think this is ready for a final review.

spyder/utils/ipython/start_kernel.py
```diff
@@ -164,6 +164,12 @@ def kernel_config():
                     spy_cfg.InlineBackend.rc['figure.figsize'] = (width_o,
                                                                   height_o)
 
+                # Print figure kwargs
+                bbox_inches_o = os.environ.get('SPY_BBOX_INCHES_O')
+                bbox_inches = 'tight' if bbox_inches_o == 'True' else None
+                spy_cfg.InlineBackend.print_figure_kwargs.update(
+                        {'bbox_inches': bbox_inches})
```
CAM-Gerlach: Same here; see above.

spyder/plugins/ipythonconsole.py
```diff
@@ -418,7 +419,15 @@ def setup_page(self):
                           _("Height:")+"  ", " "+_("inches"),
                           'pylab/inline/height', min_=1, max_=20, step=1,
                           tip=_("Default is 4"))
-        
+        bbox_inches_box = newcb(
+                _("Use a tight layout for inline plots"),
```
CAM-Gerlach: Hanging indent should be one indent level (four spaces) per PEP8, not two (eight), except in the case of a function definition. Spyder's automatic handling of it is in error, as I've pointed out in another issue a while ago.

jnsebgosselin: Thanks for the review. In the PEP8, it is said though that:
> The 4-space rule is optional for continuation lines.
>
> Optional:
>
> # Hanging indents *may* be indented to other than 4 spaces.
> foo = long_function_name(
>   var_one, var_two,
>   var_three, var_four)

source: https://www.python.org/dev/peps/pep-0008/#indentation 

Is this information somewhat outdated?

jnsebgosselin: Though, I also agree it is better to keep it to a consistent 4-space intent that eats less horizontal space than a 8-spaces indent, so I'm going to correct this anyway. I was just curious.

CAM-Gerlach: Thanks for pointing that out. Of course, you would be much more aware than I if that was the case, so I presume that it is not. However, it seems prudent to adopt the standard four space convention for consistency with existing code and that from our many contributors—absent a strong reason to choose differently, which you might have.

jnsebgosselin: I'm not a professional programmer either... so I don't know better than you about this :)

I also agree with you. A 8 spaces hanging-intent doesn't make much sense if you think about it. In the first place, it is done for lack of horizontal space, while it eats 4 optional spaces..

So yeah, I think that Spyder automatic indentation, while not being wrong strictly speaking, is not opting for the smarter choice here.

CAM-Gerlach: You're a heck of a lot closer to one than me, haha.

Particularly with the automatic indentation, , it seems only sensible that we go with the standard convention if we had to pick one, even if it isn't strictly binding. If I recall, the current 8 spaces was only supposed to happen for function definitions (a relatively small case), but either that was never implemented to differentiate between the two, or there is a bug in the implementation.

spyder/app/tests/test_mainwindow.py
```diff
@@ -1196,6 +1198,68 @@ def test_varexp_magic_dbg(main_window, qtbot):
...
+
+    # Get the image name from the html, fetch the image from the shell, and
+    # then save it to a file.
+    html = shell._control.toHtml()
+    img_name = re.search('''<img src="(.+?)" /></p>''', html).group(1)
+    qimage2 = shell._get_image(img_name)
+    assert isinstance(qimage2, QImage)
+
+    # Compare both images and assert that they are not identical.
+    assert qimage1 != qimage2
```
ccordoba12: Could you use the Matplotlib function I mentioned before to do a more meaningful comparison?

jnsebgosselin: OK I think I understand what were suggesting now. 

I need to produce a sample image locally first and then upload it along with my PR. Than, I need to compare the results of the test with that sample image. I didn't understood what you meant at first, sorry for that. That is why I failed to see the point to use the matplotlib function to compare both images, while I could simply compare the QImages.

jnsebgosselin: It is done. It is failing for PyQt4. Probably because the inline output is slightly different than for PyQt5. I'm tempted to skip it.

jnsebgosselin: I've just created a new Python 3.5 env. with PyQt4. I've noticed that the version of matplotlib that is installed in this environment is very old: version 1.5.1. So there is noticeable differences between the graphs produces with the default parameters.

I think I can make this work by setting stricter constraints on the plot.

jnsebgosselin: Ok no, scrap my last comment. It is set to 'tight' by default and I don't know why it is still failing... !

jnsebgosselin: Carlos, the resulting figure when `bbox_inches='tight'` depends on the matplotlib version. Therefore, I cannot produce a unique sample figure that will allow the test to pass on every environment we are supporting for the case where `bbox_inches='tight'`.

I can create a 2 different versions of the sample image that will allow the test to pass with recent and older versions of Matplotlib. However, that will make the test highly dependent on future changes of the matplotlib `tight_layout` mechanics.

A better option imo would be to assert that the resulting figure on the CI fit the sample image only for the `bbox_inches=None` case and assert that the comparison fails when `bbox_inches='tight'` with the same sample figure.

ccordoba12: > A better option imo would be to assert that the resulting figure on the CI fit the sample image only for the bbox_inches=None case and assert that the comparison fails when bbox_inches='tight' with the same sample figure.

Ok, no problem.

ccordoba12: Also, if you want to skip this test on PyQt4, please do.

jitseniesen: Is it possible to generate the sample image in matplotlib but without using Spyder as part of the test, and then compare with the image generated in Spyder's IPython console? That should make sure that the text works with whatever matplotlib version is installed on the test computer (at the cost of requiring more test time).

ccordoba12: That's a good idea too.

jnsebgosselin: @jitseniesen thanks for the idea. This is interesting.

I think what could be done is to save the figure with `savefig` by specifying the `bbox_inches` argument and then compare the result with the figure produced inline. I think that should do it!

jnsebgosselin: Oh my gosh... It worked ! 😱 
Thank you @jitseniesen

spyder/app/tests/test_mainwindow.py
```diff
@@ -29,8 +30,11 @@
 from qtpy import PYQT4, PYQT5, PYQT_VERSION
 from qtpy.QtCore import Qt, QTimer, QEvent, QUrl
 from qtpy.QtTest import QTest
+from qtpy.QtGui import QImage
 from qtpy.QtWidgets import QApplication, QFileDialog, QLineEdit, QTabBar
 from qtpy.QtWebEngineWidgets import WEBENGINE
+from matplotlib.testing.compare import compare_images
+from matplotlib.testing.exceptions import ImageComparisonFailure
```
ccordoba12: This import is not used. Please remove it.

PR #6870, opened 2018-03-31
PR: Add checklists for Github issue report and pull requests

CAM-Gerlach:
### Pull Request Checklist

* [x] ~~Wrote at least one-line docstrings for any new functions~~
* [x] ~~Added at least one unit test covering the changes, if at all possible~~
* [x] ~~Noted what issue(s) this pull request resolves, creating one if needed~~

## Description of Changes

This PR adds a checklist for the user to complete before submitting an issue report (both in the issue template and submitted from the Spyder Help menu; automatic/semi-auto error reports are unaffected), to hopefully help increase the quality of issues received and enable speedy closure of those that aren't helpful. To avoid any chance of an incomplete report due to length limitations, the text verbosity for the Help menu report (particularly the mostly redundant reminder) was cut down where possible.

image

Also, it adds a heavily revised PR template (as used here) with a checklist (top), along with formatting enhancements, some additional items and helpful tips in the comments.

### Issue(s) Resolved

As this is an internal change to the Github templates, none believed needed.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -2,21 +2,16 @@
 
 ### Issue Report Checklist
 
-* [ ] Ensured the issue is with the core Spyder IDE, not Python, a package, your code, a Spyder plugin, etc.
-* [ ] Searched the [issue page](https://github.com/spyder-ide/spyder/issues?q=is%3Aissue) for similar reports by description and/or the last line of the error/traceback
+* [ ] Searched the [issues page](https://github.com/spyder-ide/spyder/issues?q=is%3Aissue) for similar reports
 * [ ] Read the relevant sections of the [Spyder Troubleshooting Guide](https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ) and followed its advice
-* [ ] Updated with ``conda update spyder`` (or ``pip update spyder``) and reproduced the issue
-* [ ] Could not reproduce with a bare ``qtconsole`` (from Anaconda Navigator or ``jupyter qtconsole``)
-* [ ] Completed the following troubleshooting steps (if a Spyder component is not working/launching)
+* [ ] Reproduced the issue after updating with ``conda update spyder`` (or ``pip``, if not using Anaconda)
+* [ ] Could not reproduce inside ``jupyter qtconsole`` (if console-related)
+* [ ] Tried basic troubleshooting (if a bug/error)
     * [ ] Restarted Spyder
     * [ ] Reset preferences with ``spyder --reset``
-    * [ ] Cleanly reinstalled the latest version of the Anaconda distribution
+    * [ ] Reinstalled the latest version of [Anaconda](https://www.anaconda.com/download/)
     * [ ] Tried the other applicable steps from the Troubleshooting Guide
-* [ ] Completed *all* relevant sections below:
-    * [ ] Problem description
-    * [ ] Steps to reproduce
-    * [ ] Version and dependency information
-    * [ ] Traceback/error/output if applicable
+* [ ] Completed the **Problem Description**, **Steps to Reproduce** and **Version** sections below
 
 
 ## Problem Description
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -20,9 +20,9 @@
 
 ### What steps reproduce the problem?
 
-1.
-2.
-3.
+1. 
+2. 
+3. 
 
 ### What is the expected output? What do you see instead?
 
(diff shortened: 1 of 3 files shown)


CAM-Gerlach:
@ccordoba12 Need anything from me to get this merged?

CAM-Gerlach:
@ccordoba12 ?

ccordoba12:
I think there are way too many boxes to check. Could you reduce the list a bit?

CAM-Gerlach:
> I think there are way too many boxes to check. Could you reduce the list a bit?

Okay, sure. I reduced the number of boxes in the issue report template by 1/3, and considerably cut down most of the text. The Help menu semi automated report version is even shorter still. Also, I cut the pull request one down by 1/4 and similarly eliminated some extra text. Let me know what you think,

.github/ISSUE_TEMPLATE.md
```diff
@@ -25,13 +37,16 @@ PASTE TRACEBACK HERE
...
-* Spyder:
-* Python:
-* Qt:
-* PyQt:
-* Operating System:
+* Spyder version:
+* Python version:
+* Qt version:
+* PyQt version:
+* OS name and version:
```
ccordoba12: `OS` is a term some people could not understand. So please revert this to `Operating System:`

spyder/app/mainwindow.py
```diff
@@ -2420,34 +2420,53 @@ def render_issue(self, description='', traceback=''):
...
 
-## Package Versions
+## Versions
 
-* Spyder version: {spyder_version} {commit}
-* Python version: {python_version}
-* Qt version: {qt_version}
-* {qt_api_name} version: {qt_api_version}
-* Operating system: {os_name} {os_version}
+* Spyder ver: {spyder_version} {commit}
```
ccordoba12: I think we have talked about this before. `ver` can be confusing and is not explicit enough for non-English speakers. So please revert this to `version`.

spyder/app/mainwindow.py
```diff
@@ -2420,34 +2420,53 @@ def render_issue(self, description='', traceback=''):
...
+* Python ver: {python_version}
+* Qt ver: {qt_version}
+* {qt_api_name} ver: {qt_api_version}
+* OS: {os_name} {os_version}
```
ccordoba12: Please revert this to `Operating System`.

.github/ISSUE_TEMPLATE.md
CAM-Gerlach: The problem (and I think I mentioned this before as well) is that we're out of characters to play around with, as aside from a small buffer I built in for one or two lines more of deps, the report starts getting cut off or outright producing errors when trying to send it to github, and I already cut and cut and cut the issue checklist down to under 1/3 of its original length as well as the disclaimer, headers etc . The real issue is that the dependency list takes up ~3-4x the characters it should due to all the fancy symbols used needed percent encoding; even a small change here (eliminating one or two deps that rarely matter; using one or two unnecessary symbols in the plain text version, etc. will cut dozens or even hundreds of characters from the final URL.

Please advise what you'd like me to do—fix the plain text display of the deps? Convert this to use the new error dialog? Just risk it and add back the chars? If I do the latter, adding even one more dep (as is discussed elsewhere) will likely break the whole URL based error reporting mechanism, or at least come dangerously close.

spyder/app/mainwindow.py
CAM-Gerlach: I agree, explicit is better than implicit, and I don't often use abbreviations myself. See above for the reasons and alternatives.

spyder/app/mainwindow.py
CAM-Gerlach: See abpve/

.github/ISSUE_TEMPLATE.md
CAM-Gerlach: Let me know what you'd like me to do about this, @ccordoba12 . Thanks!

ccordoba12: Sorry for the delay in coming back to you. Please remove this entry in the new list:

> Completed the problem description, steps to reproduce, ...

That's kind of obvious and it should be enough to add the characters that we need.

CAM-Gerlach: Sorry for the confusion; I meant to make the above initial comment on the mainwindow review item instead. That entry only exists (in that lengthy form at least) in the issue report template; the relevant text for the Help menu Report Issue item is in ``mainwindow``. However, 

> Completed all applicable sections below

while much shorter, does exist and, once percent encoded, is indeed around the same length as a dependency line, much longer than "versions". Unfortunately it isn't as obvious as one would think as many reporters don't do it, but at least it is something we can easily see for ourselves. I'll go ahead and eliminate it and spell out versions and OS, and push the changes momentarily; that should at least give us a little breathing room for the time being.

PR #6937, opened 2018-04-10
PR: Check "use a tight layout for inline plots" by default in the preferences

jnsebgosselin:
This is a follow up of PR #6790.

As discussed with @ccordoba12, the `use a tight layout for inline plots` option should be checked by default since this is the default in IPython and users are expecting this.

image


spyder/config/main.py
```diff
@@ -154,7 +154,7 @@
               'pylab/inline/resolution': 72,
               'pylab/inline/width': 6,
               'pylab/inline/height': 4,
-              'pylab/inline/bbox_inches': False,
```
ccordoba12: Please bump `CONF_VERSION` for this change to take effect. Just follow the instructions above it to see what number you need to set for it.

ccordoba12:
@jnsebgosselin, please rebase this on top of latest 3.x to fix the error in Circle. Thanks!

PR #6961, opened 2018-04-14
PR: Remove support for PyQt4

### LINKED ISSUE #6889: Remove support for PyQt4
Our reasons to do it are:

* Qt4 was discontinued in December 2015, more than two years ago.
* Spyder is quite buggy while running in PyQt4, as reported by our own developers.
* None of the core devs is running Spyder in PyQt4.
* Several of our tests segfault in PyQt4.
* By removing support for PyQt4 we could employ our limited resources in improving other aspects of the project.

TODO

- [x] Remove PyQt4 from our requirements and raise PyQt5 requirement to 5.5
- [x] Remove PyQt4 from our tests.

ccordoba12:
Fixes #6889.

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -124,7 +124,7 @@ if options.gui is None:
         print("02. PyQt5 is detected, selecting")
         os.environ['QT_API'] = 'pyqt5'
     except ImportError:
-        print("02. No PyQt5 detected (deprecated)")
+        sys.exit("ERROR: No PyQt5 detected!")
 else:
     print ("02. Skipping GUI toolkit detection")
     os.environ['QT_API'] = options.gui
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -117,21 +117,14 @@ sys.path.insert(0, DEVPATH)
 print("01. Patched sys.path with %s" % DEVPATH)
 
 
-# Selecting the GUI toolkit: PyQt5 if installed, otherwise PySide or PyQt4
-# (Note: PyQt4 is still the officially supported GUI toolkit for Spyder)
+# Selecting the GUI toolkit: PyQt5 if installed
 if options.gui is None:
     try:
         import PyQt5  # analysis:ignore
         print("02. PyQt5 is detected, selecting")
         os.environ['QT_API'] = 'pyqt5'
     except ImportError:
-        try:
-            import PyQt4  # analysis:ignore
-            print("02. PyQt4 is detected, selecting")
-            os.environ['QT_API'] = 'pyqt'
-        except ImportError:
-            print("02. No PyQt5 or PyQt4 detected, using PySide if available "
-                  "(deprecated)")
+        print("02. No PyQt5 detected (deprecated)")
 else:
     print ("02. Skipping GUI toolkit detection")
     os.environ['QT_API'] = options.gui
--- a/doc/installation.rst
+++ b/doc/installation.rst
@@ -133,9 +133,7 @@ The requirements to run Spyder are:
 
 * `Python <http://www.python.org/>`_ 2.7 or >=3.3
 
-* `PyQt5 <https://www.riverbankcomputing.com/software/pyqt/download5>`_ >=5.2 or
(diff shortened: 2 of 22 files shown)


bootstrap.py
```diff
@@ -117,21 +117,14 @@
...
         os.environ['QT_API'] = 'pyqt5'
     except ImportError:
-        try:
-            import PyQt4  # analysis:ignore
-            print("02. PyQt4 is detected, selecting")
-            os.environ['QT_API'] = 'pyqt'
-        except ImportError:
-            print("02. No PyQt5 or PyQt4 detected, using PySide if available "
-                  "(deprecated)")
+        print("02. No PyQt5 detected (deprecated)")
```
CAM-Gerlach: What does ``(deprecated)`` refer to here? Before it referred to the deprecated ``pyside`` support but that was deleted, so since PyQt5 is (hopefully) not deprecated also we can presumably delete that parenthetical?

doc/installation.rst
```diff
@@ -133,9 +133,7 @@ The requirements to run Spyder are:
 
 * `Python <http://www.python.org/>`_ 2.7 or >=3.3
 
-* `PyQt5 <https://www.riverbankcomputing.com/software/pyqt/download5>`_ >=5.2 or
-  `PyQt4 <https://www.riverbankcomputing.com/software/pyqt/download>`_ >=4.6.0
-  (PyQt5 is recommended).
+* `PyQt5 <https://www.riverbankcomputing.com/software/pyqt/download5>`_ >=5.5
```
CAM-Gerlach: Doc changes; see top level comment.

doc/overview.rst
```diff
@@ -75,6 +75,6 @@ Key features:
 * :doc:`projects`
 
 
-Spyder may also be used as a PyQt5 or PyQt4 extension library 
-(module 'spyder'). For example, the Python interactive shell widget
-used in Spyder may be embedded in your own PyQt5 or PyQt4 application.
+Spyder may also be used as a PyQt5 extension library (module
```
CAM-Gerlach: More doc changes.

setup.py
```diff
@@ -233,15 +233,15 @@ def run(self):
       long_description=
 """Spyder is an interactive Python development environment providing
 MATLAB-like features in a simple and light-weighted software.
-It also provides ready-to-use pure-Python widgets to your PyQt5 or
-PyQt4 application: source code editor with syntax highlighting and
+It also provides ready-to-use pure-Python widgets to your PyQt5
+application: source code editor with syntax highlighting and
```
CAM-Gerlach: Should be "**A** source code editor"

CAM-Gerlach:
@ccordoba12 Left a comment with a few minor issues, mainly about the doc changes potentially going to the wrong repo.

bootstrap.py
ccordoba12: Sorry, I have to remove that deprecated.

CAM-Gerlach:
@ccordoba12 Okay, sure—whatever you think is best. Would you like me to open a PR to do that, or will you handle it? Presumably, along with deleting the files we'll also need to remove the relevant section in ``setup.py``—anything else? Our CIs here aren't set up to build the docs or anything, right? Thanks.

ccordoba12:
> Would you like me to open a PR to do that, or will you handle it?

Let me do it. I know all the things that have to be removed.

CAM-Gerlach:
@ccordoba12 As part of this PR, would you also want to consider updating the minimum Qt/PyQt version to 5.6? In the thousands of bug reports from the past year, I don't know if I've seen a single user using 5.5, it is long unsupported not being a LTS like 5.6, it isn't compatible with the high dpi scaling Spyder uses, we don't test with it, and in some of these places the min Qt/PyQt version listed is even older, e.g. 5.2.

doc/installation.rst
CAM-Gerlach: @ccordoba12 Just FYI, if you don't get to it before I do my pass for baseline correctness updates (which will happen as soon as we resolve spyder-ide/spyder-docs#14 ) I can just do it as one part of that, as it is a close fit for the purpose and scope of the issue its resolving.

PR #6977, opened 2018-04-17
PR: Skip tests numpy_returns and matplotlib_figure_returns for now on jedi >=0.12 and pin CIs to 0.11.1

CAM-Gerlach:
On the newest CI builds, the jedi tests ``test_numpy_returns`` and ``test_matplotlib_figure_returns`` are failing to get completions, causing the test to fail. @ccordoba12 determined this was due to ``jedi`` 0.12, and thus we're skipping the test for now on this version and later until a more permanent fix can be devised.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -47,11 +47,13 @@ install:
   - "conda install jupyter_client=5.2.2"
   # Fix problems with latest pyqt
   - "conda install pyqt=5.6*"
+  # Fix test issues with the latest jedi 0.12 for now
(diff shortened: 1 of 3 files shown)


CAM-Gerlach:
Looks like everything's failing now that the ``jedi`` update got pushed to ``conda``, and ``matplotlib_figure_returns`` is also a problem which I skipped as well. Can we just pin to ``jedi`` 0.11 on our CIs so we don't have to skip our tests on all platforms?

PR #6986, opened 2018-04-18
PR: Adapt sphinxify for changes in Sphinx 1.8

larsoner:
In sphinx 1.8 (dev), the source and destination directory cannot be the same for a build, so `sphinxify` fails. This sets the destination directory to be a subdirectory.

### DIFF
--- a/spyder/utils/help/sphinxify.py
+++ b/spyder/utils/help/sphinxify.py
@@ -166,15 +166,14 @@ def sphinxify(docstring, context, buildername='html'):
 
     srcdir = mkdtemp()
     srcdir = encoding.to_unicode_from_fs(srcdir)
+    destdir = osp.join(srcdir, '_build')
 
-    base_name = osp.join(srcdir, 'docstring')
-    rst_name = base_name + '.rst'
-
+    rst_name = osp.join(srcdir, 'docstring.rst')
     if buildername == 'html':
         suffix = '.html'
     else:
         suffix = '.txt'
-    output_name = base_name + suffix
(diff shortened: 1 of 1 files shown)


larsoner:
How do you envision the test? The source and destination directories are created temporarily (and removed) in the call to the function. Isn't the functionality is implicitly tested by the existing tests in `spyder/plugins/tests/test_help.py` which check rendering content? I don't see any separate `sphinxify` tests via `git grep` (e.g., no `spyder/utils/tests/test_sphinxify.py`), though I could add one if the `test_help.py` ones don't seem sufficient already.

The real TDD test would be to add a `sphinx-dev` build to the CIs. I can do that if you want.

ccordoba12:
@larsoner, could you move this PR to the 3.x branch? You can do that by following our [Contributing](https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md#changing-the-base-branch)  guide.

After that, I'll add `sphinx-dev` to one of our CIs so we can test this.

larsoner:
I rebased to `3.x` from `master` according to the instructions. Should I should open a separate PR into `3.x` (and possibly close this one, which is still into `master`)?

CAM-Gerlach:
Nope, you just need to change the PR's base branch (with the Edit button up top), which I just did for you now, and then you should be good to go.

ccordoba12:
Thanks @larsoner! I'll try to add `sphinx-dev` to our CIs this weekend, so we can test this.

ccordoba12:
Let's merge this one, I don't have time right now to add sphinx-dev to our CIs.

PR #7099, opened 2018-05-10
Shortcuts to open pylab and sympy consoles

### LINKED ISSUE #7078: [ENH] Shortcuts to open pylab and sympy consoles
Here is the idea in a picture:

new console

I find that a pylab(matplotlib and numpy) and sympy consoles are not really compatible with a lot of overlapping commands such as "plot".  

I do like using sympy as a super powered graphing calculator but I can't activate it by default in the config because it's too crazy to have as a default console, and going through the config to enable it when I need it is a pain.

Unfortunately Sympy is broken in spyder due to https://github.com/sympy/sympy/issues/13319 so I won't be able to test.

My work mate who is trying to ditch Matlab and a TI-83(graphing calc) thought this would be great. I think I can do it without too many changes. What do you think?

bcolsen:
### Pull Request Checklist

## Description of Changes

Make options to open Sympy or Pylab(numpy, matplotlib) console regardless of current preferences:

newconsole

I implemented this with the least amount of structural change I could, but there ended up needing to add function keyword to several function in order to do so.

## Motivation

I find that a pylab(matplotlib and numpy) and sympy consoles are not really compatible with a lot of overlapping commands such as "plot".

I do like using sympy as a super powered graphing calculator but I can't activate it by default in the config because it's too crazy to have as a default console, and going through the config to enable it when I need it is a pain.

Unfortunately Sympy is broken in spyder due to sympy/sympy#13319 so I won't be able to test.

### Issue(s) Resolved

Fixes #7078


jitseniesen:
How does this interact with the current option in the preferences to use pylab or sympy by default?

bcolsen:
@jitseniesen Good question!

Presently it only overrides your default options if they are set to False.  For example if both of the pylab and sympy options are checked in the preference then all three new consoles will load all the packages and be identical.  Both of these options are False out of the box so I don't think that's too bad behavior.

For me I leave Pylab True all the time and write my scripts in name-spaced python and do quick calculations on the results in the console using the shorter pylab fuctions. Sympy loads after Pylab and over writes some of Pylab(plot for example) so loading Sympy and Pylab gives you a working Sympy still.

## Option 1
I could easily change to override the defaults and make: Default, Pylab(only pylab) and Sympy(only sympy). After thinking about it, I think this would be the best and least confusing thing to do.

## Option 2
It should be simple to make 4 options in the menus: New console(with your setting), New "plain" console(with no extra modules), Pylab(only pylab) and Sympy(only sympy).  The "plain" console would need a better name, and it's probably not that useful, but might be handy

## Option 3
This would necessitate the most changes, and be most work :-) 

Have three options in the user preferences as a drop down selector: IPython(no extra modules), Pylab or Sympy. Then allow the user to pick which they would like open by default in the options, and have the default, "plain", Pylab and Sympy options in the console drop down menu. This wouldn't allow stacking of Pylab and Sympy, but that isn't a great option anyways with the name-space collisions. I could make another option with both, but I think both is a bad idea anyway. This could even allow scripts to open in a specific type of dedicated console from the run options.

I will do Option 1, because I think that's best (and easy at this point).  I can also see how feasible Option 3, because I think that's the simplest user experience.

spyder/config/main.py
```diff
@@ -432,6 +432,8 @@
               'profiler/run profiler': "F10",
               # ---- In widgets/ipythonconsole/shell.py ----
               'ipython_console/new tab': "Ctrl+T",
+              'ipython_console/new pylab tab': "Ctrl+U",
```
CAM-Gerlach: Is there a particular motivation for the choice of these shortcuts? Is it the best idea to occupy top-level (Ctrl + Letter) shortcuts with these commands?

bcolsen: The only motivation is `T` `Y` `U` are next to each other on the keyboard and `S` and `P` were already taken.  I don't really need any of those shortcuts that often even "Ctrl+T".

jitseniesen: I think the IPython console uses Emacs-style shortcuts, so Ctrl-U means "kill everything in front of the cursor" and Ctrl-Y means "yank previously killed text" (kill and yank are the Emacs terms for cut and paste). Maybe it's best not to introduce the shortcuts, at least unless somebody actually wants them?

bcolsen: > Maybe it's best not to introduce the shortcuts

I agree. I'll remove the shortcuts defaults.  I'll try leave them as unset in the shortcut preferences so users can set them to whatever they.

CAM-Gerlach: Sounds good to me!

bcolsen: I can't seem to get the shortcuts to work at all no matter what I set them to.  I must be missing a place where I need to register them. So these shortcuts don't currently work at all.

If I set them to `None` they appear blank in the preferences and I can set them. But the shortcuts conflict warning comes up saying they are set to `None` and they wouldn't work when I set them.

This is outside the scope of this PR so the shortcuts were removed entirely

bcolsen:
I added code to rename the console tabs to Sympy or Pylab when they are created.

I'm having trouble finding something that would be a meaningful test without testing pylab or sympy directly. Other than that I think this is ready

bcolsen:
I wrote tests, but i required sympy to test it.  I saw that it was a testing requirement but it seems unavailable. 

https://github.com/spyder-ide/spyder/blob/03c716133d9c9a4c8b424adece53436df799fb86/requirements/test_requirements.txt#L9

ccordoba12:
@bcolsen, you need to add it to `appveyor.yml` and `.travis.yml`.

spyder/plugins/ipythonconsole.py
```diff
@@ -816,13 +816,27 @@ def get_plugin_actions(self):
...
+                                   _("Open a &new console"),
                                    icon=ima.icon('ipython_console'),
                                    triggered=self.create_new_client,
                                    context=Qt.WidgetWithChildrenShortcut)
         self.register_shortcut(create_client_action, context="ipython_console",
                                name="New tab")
 
+        create_pylab_action = create_action(
+                                   self,
+                                   _("Open a new &pylab console"),
```
ccordoba12: Remove the ampersand and capitalize `pylab`.

spyder/plugins/ipythonconsole.py
```diff
@@ -816,13 +816,27 @@ def get_plugin_actions(self):
...
+                                   icon=ima.icon('ipython_console'),
+                                   triggered=self.create_pylab_client,
+                                   context=Qt.WidgetWithChildrenShortcut)
+
+        create_sympy_action = create_action(
+                                   self,
+                                   _("Open a new &sympy console"),
```
ccordoba12: Remove the ampersand and capitalize `sympy`.

spyder/plugins/ipythonconsole.py
```diff
@@ -1079,11 +1099,38 @@ def create_new_client(self, give_focus=True, filename='', is_cython=False):
...
 
-        self.connect_client_to_kernel(client, is_cython=is_cython)
+        self.connect_client_to_kernel(client, is_cython=is_cython,
+                                      is_pylab=is_pylab, is_sympy=is_sympy)
         if client.shellwidget.kernel_manager is None:
             return
         self.register_client(client)
 
+    def create_pylab_client(self, give_focus=True, filename='',
+                            is_cython=False, is_pylab=True, is_sympy=False):
```
ccordoba12: `is_pylab=True` is wrong. It should `is_pylab=False`, as the others.

ccordoba12: Sorry, silly me. It's right.

But we certainly don't need all these args here. This should be a method without args because their values can be passed directly to `self.create_new_client` inside it.

spyder/plugins/ipythonconsole.py
```diff
@@ -1079,11 +1099,38 @@ def create_new_client(self, give_focus=True, filename='', is_cython=False):
...
+        """Force creation of pylab client"""
+        self.create_new_client(give_focus=give_focus,
+                               filename=filename,
+                               is_cython=is_cython,
+                               is_pylab=is_pylab,
+                               is_sympy=is_sympy)
+        # Rename client tab with Pylab
+        client = self.get_current_client()
+        tab_text = _("Pylab") + u' ' + client.id_['int_id']
```
ccordoba12: This is not the right way to do this. Please add a new kwarg to `create_new_client` called `given_name=None` and pass its value to `ClientWidget`. Then here you simply need to do

```python
console_name = _('Pylab')
self.create_new_client(is_pylab=True,
                       given_name=console_name)
```

and `ClientWidget` methods will take care of correctly naming the new console.

spyder/plugins/ipythonconsole.py
```diff
@@ -1079,11 +1099,38 @@ def create_new_client(self, give_focus=True, filename='', is_cython=False):
...
+        self.rename_client_tab(client, tab_text)
+
+    def create_sympy_client(self, give_focus=True, filename='',
```
ccordoba12: Please apply here the same changes I mentioned above for `create_pylab_client`.

spyder/plugins/ipythonconsole.py
ccordoba12: Another correction, `console_name` needs to be `console_name = 'Pylab'` because `Pylab` can't be translated.

ccordoba12:
@bcolsen, this is a really great addition! It hadn't occurred to me, but it'll certainly help to use our Sympy and Pylab integrations quite more frequently.

I left some comments to simplify your code, but other than that this looks really good!

spyder/plugins/ipythonconsole.py
CAM-Gerlach: To be clear, proper capitalization is ``SymPy``

spyder/plugins/ipythonconsole.py
ccordoba12: I forgot to mention that `ClientWidget` has an arg called `given_name`.

bcolsen: Using `given_name` seems to give me `SymPy/A` every time I open a new console instead of `SymPy 2/A` and 3 and 4 as I create more.  Should I change the behavior of `ClientWidget.get_name` to always include the console id so I can have the original behavior? 

The only other place I can see that uses `get_name` is for remote connections. I have never used remote sessions but if I opened 2 of the same connections I would still want them to automatically get unique name.

Also what's the `/A` for? It seem to be static so all local console get it.

spyder/widgets/ipythonconsole/client.py
```diff
@@ -297,6 +297,9 @@ def get_name(self):
             # Adding id to name
             client_id = self.id_['int_id'] + u'/' + self.id_['str_id']
             name = name + u' ' + client_id
+        elif self.given_name in ["PyLab", "SymPy"]:
```
bcolsen: This is a little silly, but it works.  The down side is that you can't rename a tab to "PyLab" or "SymPy" but that's a corner case.

ccordoba12:
Sorry for not mentioning it before, but I don't like the capitalization proposed by @CAM-Gerlach. I'd like to revert to `Pylab` and `Sympy` instead.

Other than that, this PR looks ready to be merged.

CAM-Gerlach:
> Sorry for not mentioning it before, but I don't like the capitalization

@ccordoba12 Well haha whether or not you like it doesn't change the fact that it is the *correct* capitalization for the [SymPy](http://www.sympy.org/en/index.html) and [PyLab](https://scipy.github.io/old-wiki/pages/PyLab) names :) which I why I recommended it be corrected to such. 

Put another way, would you want another application expressing the Spyder name as SPyder or sPYder in their UI, when our correct name is Spyder?

ccordoba12:
What about `Open a new symbolic math (Sympy) console`? Is that too much?

The problem is I think not many newbies know what Sympy alone means (The same goes for Pylab, but I don't how we could name that one).

CAM-Gerlach:
> those SymPy guys do seem to be stuck on that strange capitalization.

Well, its not that strange considering its also what [NumPy](http://www.numpy.org/), [SciPy](https://www.scipy.org/about.html) ,  and of course [IPython](https://ipython.org/) itself (i.e. all the core PyData ecosystem packages that have Py in their names) all use.

> Pylab is really just a Matplotlib thing that they want to depreciate. 

Right, that's not so much of a big deal...

bcolsen:
> Well, its not that strange considering its also what NumPy, SciPy , and of course IPython itself (i.e. all the core PyData ecosystem packages that have Py in their names) all use.

That's true. It is certainly not without precedent.

> What about Open a new symbolic math (Sympy) console? Is that too much?

I like the Idea of being verbose so people can see what they might be getting.

We could try:
`Open a new SymPy (symbolic math) console`
`Open a new Pylab (scientific computing) console`

CAM-Gerlach:
> I like the Idea of being verbose so people can see what they might be getting.

> The problem is I think not many newbies know what Sympy alone means 

Yeah, definitely agreed and that fits with the Zen of Python—explicit is better than implicit.

> Open a new Pylab (scientific computing) console

I mean, is that fully accurate/most representative? I'm not certain, but I haven't really used PyLab since it was deprecated before my time, just asking to make sure.

bcolsen:
It looks like this:

new_console

bcolsen:
Pylab basically imports matplotlib.pyplot(plus some soon to be depreciated convenience functions that were common on matlab) and numpy.  

Scientific computing is rather broad and not specific.

Some options:
1. `Open a new Pylab (plot and array) console` (vary literal but array might be confusing)
1. `Open a new Pylab (data plotting) console` (That's what I use it for quick and dirty data browsing)
1. `Open a new Pylab (data analysis and plotting) console`

bcolsen:
pushed option 1

CAM-Gerlach:
Option 2 is probably the best, since it describes the intended task well without being too broad or specific. 3 (and your original option, even more so) is a little better verbiage, but is somewhat broad—it could imply that it was the primary console intended for data analysis in Spyder, when in fact the opposite is the case—we don't want to encourage people to use it if they don't already know they need it. Finally, 1 is very confusing as you say, unlikely to help a new person and isn't at all coherent from a grammar/syntax perspective, so I'd strongly recommend against it.

CAM-Gerlach:
> pushed option 1

I wasn't fast enough for you? :S

CAM-Gerlach:
> Sorry about that, but I was going to bed and I wanted the repo up to date in case I needed to access it from my work.computer. 

No worries!

bcolsen:
Not super important but what about:

`Open a new Pylab console (data plotting)`
`Open a new SymPy console (symbolic math)`

CAM-Gerlach:
Even better, IMO :+1:

ccordoba12:
Yes, it's much better for me too. Thanks @bcolsen!

ccordoba12:
I left a couple of tiny comments, then this should be ready.

I have a last question though (that could be solved here or in another PR): What about adding a Cython console too? I mean, at least there are options for SymPy and Pylab in our Prefs, but there's none about Cython. So if people don't know that Spyder has support for Cython beforehand, they'll probably never find out about it.

@bcolsen, @CAM-Gerlach, what do you think?

CAM-Gerlach:
@ccordoba12 Sure, that's a fine idea; I myself wasn't even fully aware that you could launch a dedicated Cython console in Spyder (rather than having to run a Cython file). If it wouldn't take too much to do here for @bcolsen , then we can just knock it out now. Also, we can add our project to the [Projects using Cython](https://github.com/cython/cython/wiki/projects) list for a little more visibility and SEO (as part of spyder-ide/spyder-docs#39 )

bcolsen:
> What about adding a Cython console too?

It would be a pretty small change the keyword is already there. I can do it.

Only one problem, what do we call it?  :-) 

Options:
1. `Open a new Cython console (static compiler)`
1. `Open a new Cython console (C extensions)`
1. `Open a new Cython console (Python and C)`

CAM-Gerlach:
Good question! Come to think of it, why not cut the unnecessary "Open a" verbiage and make all three just "New XXX Console"; its shorter, more succient, and more idiomatic while avoiding duplicate text in all four entries and allows more room to add another word to the description if needed, since I would recommend ``(Python with C extensions)``; thus, the menu would look like:

```
New IPython console (default)
New SymPy console (symbolic math)
New Pylab console (data plotting)
New Cython console (Python with C extensions)
```

bcolsen:
I like the idea of dropping the obvious "Open a".

I think the Cython console would be fine as `(Python + C)` if "extensions" is too long.  The tips are for new people so the mention a C should be good enough to let then know what going on and "extensions" wouldn't mean much to them anyway.

CAM-Gerlach:
Yup,  agreeed. :+1:

ccordoba12:
> New IPython console (default)

Should we really need `IPython` here? We could simply use `New console (default)`

> I like the idea of dropping the obvious "Open a".

Agreed. Thanks for the suggestion @CAM-Gerlach.

> I think the Cython console would be fine as (Python + C)

I'd prefer

    New Cython console (C/C++ extensions for Python)

CAM-Gerlach:
>  Should we really need `IPython` here?

Very good point. Better to drop it as you suggest as it makes it simpler and less confusing for new people about which console they should open, and if I understand correctly all the others are using the IPython backend anyway just with stuff on top. To this end, it might also help to add some sort of separator between ``New console (default)`` and the other three, to make it even clearer.

> New Cython console (C/C++ extensions for Python)

Is even longer than my original suggestion, and is less clear that it is *also* a full superset of Python, which could mislead beginners away. Also unlike the above, it describes what *Cython* is, not what the *Cython console* provides as it should (both syntactically and conceptually). 

I'm not totally convinced about adding "extensions" considering the audience (beginners who've never heard of Cython; otherwise they already have at least a rudimentary idea of what it is beyond whatever we can stuff into a few words), but if so, then we should go with ``Python + C extensions`` for the above reasons, and is substantially shorter and thus looking less awkward and making the whole menu a more reasonable size. 

With regard to adding the ``/++`` (the extra ``C`` is redundant),  its nominally three extra chars but Cython's own [short descriptor](http://cython.org/) doesn't see it necessary to include it, and more importantly ``Python + C/++ extensions`` might be confusing due to the two different usages of ``+``, in which case we'd have to go with ``with`` instead (``Python with C/++ extensions``) which makes it overall longer by 7 characters (equaling your suggestion above), not a trivial amount.

Comparison of the three alternatives, in context:

```
New console (default)
New SymPy console (symbolic math)
New Pylab console (data plotting)
New Cython console (Python with C/++ extensions)
```

vs.

```
New console (default)
New SymPy console (symbolic math)
New Pylab console (data plotting)
New Cython console (Python + C extensions)
```

vs.

```
New console (default)
New SymPy console (symbolic math)
New Pylab console (data plotting)
New Cython console (Python + C)
```

bcolsen:
I went with 
`Python with C-extensions`
for now

spyder/plugins/ipythonconsole.py
```diff
@@ -821,13 +821,35 @@ def get_plugin_actions(self):
         """Return a list of actions related to plugin."""
         create_client_action = create_action(
                                    self,
-                                   _("Open an &IPython console"),
+                                   _("Open a new console"),
```
ccordoba12: Let's drop the `Open` here.

spyder/plugins/tests/test_ipythonconsole.py
```diff
@@ -132,6 +147,70 @@ def test_auto_backend(ipyconsole, qtbot):
     assert 'Error' not in control.toPlainText()
 
 
+@pytest.mark.slow
+@flaky(max_runs=3)
+@pytest.mark.pylab_client
+def test_pylab_client(ipyconsole, qtbot):
+    """Test that the automatic backend is working correctly."""
```
ccordoba12: Please change this docstring.

spyder/plugins/tests/test_ipythonconsole.py
```diff
@@ -132,6 +147,70 @@ def test_auto_backend(ipyconsole, qtbot):
...
+    qtbot.waitUntil(lambda: shell._prompt_html is not None,
+                    timeout=SHELL_TIMEOUT)
+
+    # This is here to generate further errors
+    with qtbot.waitSignal(shell.executed):
+        shell.execute("e")
+
+    # Assert there are no errors in the console
+    control = ipyconsole.get_focus_widget()
+    assert 'NOTE' not in control.toPlainText()
```
ccordoba12: Please remove this line. I think it's not necessary.

spyder/plugins/tests/test_ipythonconsole.py
```diff
@@ -132,6 +147,70 @@ def test_auto_backend(ipyconsole, qtbot):
...
+    assert 'Error' not in control.toPlainText()
+
+
+@pytest.mark.slow
+@flaky(max_runs=3)
+@pytest.mark.sympy_client
+@pytest.mark.xfail('1.0' < sympy.__version__ < '1.2',
+                   reason="A bug with sympy 1.1.1 and IPython-Qtconsole")
+def test_sympy_client(ipyconsole, qtbot):
+    """Test that the automatic backend is working correctly."""
```
ccordoba12: Fix docstring.

spyder/plugins/tests/test_ipythonconsole.py
```diff
@@ -132,6 +147,70 @@ def test_auto_backend(ipyconsole, qtbot):
...
+    qtbot.waitUntil(lambda: shell._prompt_html is not None,
+                    timeout=SHELL_TIMEOUT)
+
+    # This is here to generate further errors
+    with qtbot.waitSignal(shell.executed):
+        shell.execute("x")
+
+    # Assert there are no errors in the console
+    control = ipyconsole.get_focus_widget()
+    assert 'NOTE' not in control.toPlainText()
```
ccordoba12: Remove line.

spyder/plugins/tests/test_ipythonconsole.py
```diff
@@ -132,6 +147,70 @@ def test_auto_backend(ipyconsole, qtbot):
...
+    assert 'Error' not in control.toPlainText()
+
+
+@pytest.mark.slow
+@flaky(max_runs=3)
+@pytest.mark.cython_client
+def test_cython_client(ipyconsole, qtbot):
+    """Test that the automatic backend is working correctly."""
```
ccordoba12: Fix docstring.

spyder/plugins/tests/test_ipythonconsole.py
```diff
@@ -132,6 +147,70 @@ def test_auto_backend(ipyconsole, qtbot):
...
+
+    # This is here to generate further errors
+    with qtbot.waitSignal(shell.executed):
+        shell.execute("%%cython\n"
+                      "cdef int ctest(int x, int y):\n"
+                      "    return x + y")
+
+    # Assert there are no errors in the console
+    control = ipyconsole.get_focus_widget()
+    assert 'NOTE' not in control.toPlainText()
```
ccordoba12: Remove line.

ccordoba12:
> Python with C-extensions

Ok, fair enough. Let's go with it.

There are a couple of things we should also consider, but they could be solved in other PRs:

1. If the kernel of a Pylab, SymPy, Cython console is restarted, then the new kernel should also be in Pylab, SymPy, Cython mode. I think that's not the case right now.
2. The kernel needs to inform Spyder if Matplotlib, SymPy and Cython are installed on it, so these options are enabled/disabled in the Spyder UI. I mean, it makes no sense to show the *New SymPy console* as enabled when there's no SymPy on the kernel.

spyder/plugins/ipythonconsole.py
CAM-Gerlach: And ``a``, of course, if you're going to do that.

bcolsen: how about:

`New console (default settings)`

CAM-Gerlach:
> If the kernel of a Pylab, SymPy, Cython console is restarted, then the new kernel should also be in Pylab, SymPy, Cython mode. I think that's not the case right now.

Ooh, good catch. That's serious enough that I'd consider it a bug; at the very least it breaks a strong user expectation of the console behavior. If its possible to quickly fix it in this PR without holding things up, then IMO it should be done as its in scope if its a bug with the functionality being added.

> The kernel needs to inform Spyder if Matplotlib, SymPy and Cython are installed on it, so these options are enabled/disabled in the Spyder UI. I mean, it makes no sense to show the New SymPy console as enabled when there's no SymPy on the kernel.

Yeah; that's good Ux, though as you imply its not necessarily needed for this PR right away.

spyder/plugins/ipythonconsole.py
CAM-Gerlach: How about just ``New console (default)`` as discussed above.

ccordoba12: I agree with @CAM-Gerlach on this one.

CAM-Gerlach:
> ``Python with C-extensions``

I know Cython uses it, but at least in this context, the meaning (that ``C`` is the language) is clearer without the hyphen, IMO.

So I tested it, and we got some problems. The Pylab (and default) console restarts just fine, and furthermore back into Pylab. However, restarting the SymPy console leads to the following error:

```python-traceback

  File "<ipython-input-1-f706315e3fb6>", line 3, in <module>
    init_session()

  File "C:\Anaconda3\lib\site-packages\sympy\interactive\session.py", line 433, in init_session
    auto_int_to_Integer=auto_int_to_Integer)

  File "C:\Anaconda3\lib\site-packages\sympy\interactive\session.py", line 264, in init_ipython_session
    app.initialize(argv)

  File "<decorator-gen-113>", line 2, in initialize

  File "C:\Anaconda3\lib\site-packages\traitlets\config\application.py", line 87, in catch_config_error
    return method(app, *args, **kwargs)

  File "C:\Anaconda3\lib\site-packages\IPython\terminal\ipapp.py", line 317, in initialize
    self.init_shell()

  File "C:\Anaconda3\lib\site-packages\IPython\terminal\ipapp.py", line 333, in init_shell
    ipython_dir=self.ipython_dir, user_ns=self.user_ns)

  File "C:\Anaconda3\lib\site-packages\traitlets\config\configurable.py", line 423, in instance
    '%s are being created.' % cls.__name__

MultipleInstanceError: Multiple incompatible subclass instances of TerminalInteractiveShell are being created.
```

and restarting the Cython console gives the following error:

```python-traceback

  File "<ipython-input-1-4aee485a5aa2>", line 1, in <module>
    get_ipython().run_line_magic('reload_ext', 'Cython')

  File "C:\Anaconda3\lib\site-packages\IPython\core\interactiveshell.py", line 2095, in run_line_magic
    result = fn(*args,**kwargs)

  File "<decorator-gen-68>", line 2, in reload_ext

  File "C:\Anaconda3\lib\site-packages\IPython\core\magic.py", line 187, in <lambda>
    call = lambda f, *a, **k: f(*a, **k)

  File "C:\Anaconda3\lib\site-packages\IPython\core\magics\extension.py", line 63, in reload_ext
    self.shell.extension_manager.reload_extension(module_str)

  File "C:\Anaconda3\lib\site-packages\IPython\core\extensions.py", line 135, in reload_extension
    self.load_extension(module_str)

  File "C:\Anaconda3\lib\site-packages\IPython\core\extensions.py", line 92, in load_extension
    if self._call_load_ipython_extension(mod):

  File "C:\Anaconda3\lib\site-packages\IPython\core\extensions.py", line 139, in _call_load_ipython_extension
    mod.load_ipython_extension(self.shell)

  File "C:\Anaconda3\lib\site-packages\Cython\__init__.py", line 11, in load_ipython_extension
    from .Build.IpythonMagic import CythonMagics  # pylint: disable=cyclic-import

  File "C:\Anaconda3\lib\site-packages\Cython\Build\IpythonMagic.py", line 86, in <module>
    from .Inline import cython_inline

  File "C:\Anaconda3\lib\site-packages\Cython\Build\Inline.py", line 17, in <module>
    from ..Compiler.ParseTreeTransforms import (CythonTransform,

  File "C:\Anaconda3\lib\site-packages\Cython\Compiler\ParseTreeTransforms.py", line 22, in <module>
    from .TreeFragment import TreeFragment

  File "C:\Anaconda3\lib\site-packages\Cython\Compiler\TreeFragment.py", line 21, in <module>
    from . import Parsing

  File "Cython\Compiler\Parsing.py", line 30, in init Cython.Compiler.Parsing
    from .ModuleNode import ModuleNode

  File "C:\Anaconda3\lib\site-packages\Cython\Compiler\ModuleNode.py", line 3119, in <module>
    capsule_utility_code = UtilityCode.load("Capsule")

  File "Cython\Compiler\Code.py", line 251, in Cython.Compiler.Code.UtilityCodeBase.load

ValueError: More than one filename match found for utility code Capsule
```

CAM-Gerlach:
Also, as a minor note, Pylab imports several key constants, and a few other objects of indeterminate purpose. When clearing variables, these also get cleared, which is probably not desired. However, this probably isn't exactly trivial to fix, if desired; it would probably just be easiest to autoreload Pylab itself to get them back, rather than maintaining a fragile blacklist of Pylab variable names to not clear.

Which brings up another point...what happened to the "Gear" icon menu in the Variable Explorer, that allowed the user to filter Variables and show array min/max? Both of these functions are now missing, both of which I use quite frequently.

ccordoba12:
> I'm not sure why, but it seems to work with kernel restarts.

Yep, I saw that. That's just great, less work for you!

> I'll make a separate issue for this. I personally don't like grayed out or missing menu items because I usually have to google to find out why I can't click it or why it's not there. In this case a popup to tell users to install the missing packages might be more instructive.

Ok, but popups are bad UI too because people discard them immediately. What we could do instead is displaying an error message inside our consoles say something like `Simpy is not available in the kernel, sorry`, after asking the kernel for the availability of Sympy. And since Spyder 4 is also going to have the ability to install packages, we could also say `Go to the <packages menu> to install it`.

CAM-Gerlach:
> What we could do instead is displaying an error message inside our consoles say something like `Simpy is not available in the kernel, sorry`,

I actually dislike that UX even more, since the user goes through the steps of creating a new console and waiting for it to load, only to find that it didn't actually work and knew that it wouldn't ahead of time (and also has to close the now useless console Spyder opened just to tell them that).

Instead, what about, say, changing the text of the menu item when grayed out instead to make it clear why, e.g. ``New SymPy Console (SymPy not installed!)``; then if/when we add package management features in Spyder 4, we can change the parenthetical to ``(Install SymPy from Packages menu)`` or even change the menu item to instead actually tell our package manager to start the SymPy installation process, and change the text to e.g. ``New SymPy Console (Will install SymPy first)`` 

This lets the user know both clearly and at the earliest possible time to avoid wasted time, effort and frustration, avoids both popups, extra console windows and unexplained grayed out menu items, and eventually gives them an easy, foolproof way to directly fix the problem.

PR #7107, opened 2018-05-11
PR: Show group cells in Outline Explorer

### LINKED ISSUE #7086: Grouping cells in Outline without code indentation
## Problem Description

Spyders Outline works well most of the time with cells and grouping by indentation to represent the code. Sometimes, manual setting would be advantageous (e.g. group functions desfined at top). See example..

Add the following to spyder. See Outline. 

```
#%%a simple cell with some code
def withAFunction(x):
    return x
y = withAFunction(x)

#%%another cell with a function which gets grouped, but produces indentation error
    def withAnotherFunction(x):
        return x
    y = withAFunction(x)
```

First one expected, second would be good sometimes I think. However, getting second forces a wrong ident in code.

No traceback

## Versions

Python 3.6.3 64bits, Qt 5.9.4, PyQt5 5.9.2 on Windows 
* Spyder version:  3.2.8
* Python version: 3.6.3
* Qt version: 5.9.4
* PyQt version: 5.9.2
* Operating System name/version: Win10 x64

bcolsen:
### Pull Request Checklist

* [x] Ensured your pull request hasn't eliminated unrelated blank lines/spaces, modified the ``spyder/defaults`` directory, or added new icons/assets

## Description of Changes

**Goal:** Being able to use code cells to organize and group the code blocks in the outline panel.

1. The first necessary change is to group all code blocks into code cells.  When present, code cells are now the top level objects in the outliner and all code block will be grouped under them.

2. Second the ability to use `# %%%` as a grouped cell that will be place under the previous top level `# %%` cell.

See it in action:
outline_cell_folding

I didn't actually change much code and seems to cover most cases.  If you put a code cell comment in a code block like a class or a function the out line misses the rest of the block.  I don't see a use case of a code cell in a function or class. Unfortunately this simple code doesn't know about function or classes in defined in a code block such as an if statement so I will have to do something more sophisticated with cell blocks in code blocks 

### Issue(s) Resolved

Fixes #7086 

### Things to do

* [x] Deal with cell blocks in code blocks
* [x] More group levels? (I think it's unnecessary/messy, but possible)
* [x] Fix updating the tree_cache
* [x] More manual testing
* [x] Fix bug in "previous"
* [x] Unit testing functions

### DIFF
--- a/spyder/utils/syntaxhighlighters.py
+++ b/spyder/utils/syntaxhighlighters.py
@@ -441,7 +441,7 @@ class PythonSH(BaseSH):
         else:
             offset = 0
             prev_state = self.NORMAL
-
+        
         oedata = None
         import_stmt = None
 
@@ -480,7 +480,8 @@ class PythonSH(BaseSH):
                                 oedata.text = to_text_string(text).strip()
                                 # cell_head: string contaning the first group
                                 # of '%'s in the cell header
-                                cell_head = re.findall(r"%+", text.lstrip())[0]
+                                cell_head = re.search(r"%+",
+                                                      text.lstrip()).group()
                                 oedata.cell_level = len(cell_head) - 2
                                 oedata.fold_level = start
                                 oedata.def_type = OutlineExplorerData.CELL
@@ -500,7 +501,7 @@ class PythonSH(BaseSH):
                                                    self.formats["definition"])
                                     oedata = OutlineExplorerData()
                                     oedata.text = to_text_string(text)
-                                    oedata.fold_level = (start)
+                                    oedata.fold_level = start
                                     oedata.def_type = self.DEF_TYPES[
                                                         to_text_string(value)]
                                     oedata.def_name = text[start1:end1]
@@ -511,7 +512,7 @@ class PythonSH(BaseSH):
                                 if text.lstrip().startswith(value):
                                     oedata = OutlineExplorerData()
                                     oedata.text = to_text_string(text).strip()
-                                    oedata.fold_level = (start)
+                                    oedata.fold_level = start
                                     oedata.def_type = \
                                         OutlineExplorerData.STATEMENT
                                     oedata.def_name = text.strip()
--- a/spyder/utils/syntaxhighlighters.py
+++ b/spyder/utils/syntaxhighlighters.py
@@ -420,7 +420,6 @@ class PythonSH(BaseSH):
         BaseSH.__init__(self, parent, font, color_scheme)
         self.import_statements = {}
         self.found_cell_separators = False
-        self.cell_offset = 0
         self.cell_separators = CELL_LANGUAGES['Python']
 
     def highlight_block(self, text):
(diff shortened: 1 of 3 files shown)


spyder/utils/syntaxhighlighters.py
```diff
@@ -478,7 +479,12 @@ def highlight_block(self, text):
                                 self.found_cell_separators = True
                                 oedata = OutlineExplorerData()
                                 oedata.text = to_text_string(text).strip()
-                                oedata.fold_level = start
+                                if text.lstrip().startswith(("# %%%", "#%%%")):
```
CAM-Gerlach: You should be able to allow for an unlimited number of levels and make your code cleaner to boot if you just count the number of ``"%"``s after the initial ``# %%`` or ``#%%`` and set your offset/fold level to ``base_value`` + ``count_%`` * 4.

bcolsen: I agree.  The code could be cleaner here.  I'll see what I need to do to fix "cell blocks in code blocks" and then clean it up.  Even the `4` could get messy with those "2-space" people ;-)

After looking at the "cell blocks in code blocks" issue, I think more than one subgroup level would be a testing nightmare with lots of corner cases.  What would happen if some one decorates the code block with `# %%%%%%%%%%% Cell Name %%%%%%%%%%%%`? This currently makes a code block, but you can see the problem.

bcolsen:
> could you make this behavior optional

I should be able to. I don't know where to start on that however. 

It might not be necessary to have an option. I think the old behavior is pretty similar to the expended view in the outline, and if there are no cells in the file there are no changes.

CAM-Gerlach:
You could use something like the commits adding the other options, like the one for "[Show special comments](https://github.com/spyder-ide/spyder/commit/0dd458a30383e557f690b0ac169c1ea203d6d655)" or the one for "[Show all files](https://github.com/spyder-ide/spyder/commit/f3967d3eff3b5e2b4e3f9ae320b35c5d9bbb5852)" as a template for the basic things you'd need to add (ignoring the bad code/git/etc. style, of course), and viewing current versions of the files changed since it was all added a while ago.

bcolsen:
I completely reworked the code to make the code cells there own entities outside the code with there own "indentation" levels based on the number of `%`'s in the cell header.  The tree cache is very broken at this point however.

bcolsen:
That's a lot of levels:

cell_blocks_levels

bcolsen:
I added a test that I think does everything the original test does and test the new things in this PR as well. My new code even passes the old tests, which I found surprising considering the changes.

@ccordoba12 I'd didn't want to remove a test without asking first, but I think the old test is unnecessary at his point.

CAM-Gerlach:
Thanks! How long does each test take to run?

bcolsen:
I less than a second for both.  I'll check at home.

I would only remove it for code cleaning reasons, since the new test is very derivative.

bcolsen:
I found a blocking bug in manual testing. I'll modify the test to find it

bcolsen:
It's good now

ccordoba12:
Thanks a lot @bcolsen for this! I'll try to take a look at it this weekend.

PR #7115, opened 2018-05-12
PR: Add --safe-mode option to start with a clean config directory

### LINKED ISSUE #7112: Running ./bootstap.py --test should open an empty history.py
### Issue Report Checklist

    * [x] Restarted Spyder

## Problem Description
I found that ./bootstap.py --test would also open a default Spyder instance for debugging.

But it opens your local history and opens new file with my template

It would be easier to debug issues with out resetting my Syder config all the time.

1. Start Spyder Master
2. See your 10+mb(I guess I like my code cells) history.py in the history
3. Start a new file and see your template

Like temporary a spyder reset 

## Versions

From anaconda in a new environ.
Spyder started from the command line with ```./bootstrap.py --test```

* Spyder version: current master
* Python version: 3.6.5 64bits
* Qt version: 5.9.4
* PyQt version: 5.9.2
* Operating System name/version: Ubuntu 16.04

### Dependencies

```
pycodestyle >=2.3:  2.4.0 (OK)
numpy >=1.7      :  None (NOK)
sphinx >=0.6.6   :  1.7.4 (OK)
jedi >=0.11.0    :  0.12.0 (OK)
psutil >=0.3     :  5.4.5 (OK)
IPython >=4.0    :  6.4.0 (OK)
pylint >=0.25    :  1.8.4 (OK)
```

CAM-Gerlach:
### Pull Request Checklist

* [x] ~~Wrote at least one-line docstrings for any new functions~~
* [x] ~~Added at least one unit test covering the changes, if at all possible~~
* [x] ~~Included a screenshot, if this PR makes any visible changes to the UI~~

## Description of Changes

Thanks to @bcolsen for his detailed report leading to this PR.

Currently, passing the ``--test`` flag to ``bootstrap.py`` just sets the ``SPYDER_TEST`` env variable which instead tells spyder to not load and save the user's individual settings file, e.g. ``spyder.ini``, while not affecting handling of anything else inside for ``spyder-py3`` directory, including the history logs for the normal and internal consoles, along with your pylint, profiler, working dir, help, and online help histories, your language and ``rope`` settings, any third party plugins you may have installed, and your default ``temp.py`` and ``template.py`` (new file template) files. Obviously, this is not particularly desirable for most testing scenarios—one would want "clean" to be truly "clean", without such cruft. 

Therefore, this PR just makes ``--test`` set the Spyder config dir to a clean temp one, just like ``SPYDER_PYTEST`` currently does. It also makes sure to clean the current temp directory if one exists, removes the old, unnecessary code to just load a clean config file, and does some minor cleanup immediately around changed lines.

It seems its not really possible to include a unit test for this as the Pytest suite already does this anyway, and has other checks for it, if its otherwise even really possible to test for.

### Issue(s) Resolved

Fixes #7112

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -12,6 +12,8 @@ See Issue 741
 """
 
 # pylint: disable=C0103
+# pylint: disable=C0412
+# pylint: disable=C0413
 
 import time
 time_start = time.time()
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -176,12 +176,12 @@ if options.test:
         shutil.rmtree(conf_dir)
 
 print("04. Running Spyder")
-from spyder.app import start
+from spyder.app import start  # analysis:ignore
 
-time_lapse = time.time()-time_start
-print("Bootstrap completed in " +
-    time.strftime("%H:%M:%S.", time.gmtime(time_lapse)) +  
-    # gmtime() converts float into tuple, but loses milliseconds
-    ("%.4f" % time_lapse).split('.')[1])
+time_lapse = time.time() - time_start
+print("Bootstrap completed in "
+      + time.strftime("%H:%M:%S.", time.gmtime(time_lapse))
+      # gmtime() converts float into tuple, but loses milliseconds
+      + ("%.4f" % time_lapse).split('.')[1])
 
 start.main()
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -42,8 +42,8 @@ parser.add_option('--show-console', action='store_true', default=False,
                   "is to show the console")
 parser.add_option('--hide-console', action='store_true',
                   default=False, help="Hide parent console window (Windows only)")
-parser.add_option('--test', dest="test", action='store_true', default=False,
-                  help="Test Spyder with a clean settings dir")
+parser.add_option('--clean', dest="clean", action='store_true', default=False,
+                  help="Start Spyder with a clean settings directory")
 parser.add_option('--no-apport', action='store_true',
                   default=False, help="Disable Apport exception hook (Ubuntu)")
 parser.add_option('--debug', action='store_true',
@@ -57,9 +57,9 @@ os.environ['SPYDER_BOOTSTRAP_ARGS'] = str(sys.argv[1:])
 assert options.gui in (None, 'pyqt5', 'pyqt', 'pyside'), \
        "Invalid GUI toolkit option '%s'" % options.gui
 
-# For testing purposes
-if options.test:
-    os.environ['SPYDER_TEST'] = 'True'
+# Start Spyder with a clean configuration directory for testing purposes
+if options.clean:
+    os.environ['SPYDER_CLEAN'] = 'True'
 
 # Prepare arguments for Spyder's main script
 sys.argv = [sys.argv[0]] + args
@@ -168,8 +168,8 @@ if options.hide_console and os.name == 'nt':
     print("0x. Hiding parent console (Windows only)")
     sys.argv.append("--hide-console")  # Windows only: show parent console
 
-# Reset temporary config directory if in --test mode
-if options.test:
+# Reset temporary config directory if in --clean mode
+if options.clean:
     from spyder.config.base import get_conf_path  # analysis:ignore
     conf_dir = get_conf_path()
     if osp.isdir(conf_dir):
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -20,6 +20,7 @@ import os
 import os.path as osp
 import sys
 import optparse
+import shutil
 
 
 # --- Parse command line
@@ -164,6 +165,13 @@ if options.show_console:
 if options.hide_console and os.name == 'nt':
     print("0x. Hiding parent console (Windows only)")
     sys.argv.append("--hide-console")  # Windows only: show parent console
+
+# Reset temporary config directory if in --test mode
+if options.test:
+    from spyder.config.base import get_conf_path  # analysis:ignore
+    conf_dir = get_conf_path()
+    if osp.isdir(conf_dir):
+        shutil.rmtree(conf_dir)
 
 print("04. Running Spyder")
 from spyder.app import start
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -42,8 +42,9 @@ parser.add_option('--show-console', action='store_true', default=False,
                   "is to show the console")
 parser.add_option('--hide-console', action='store_true',
                   default=False, help="Hide parent console window (Windows only)")
-parser.add_option('--clean', dest="clean", action='store_true', default=False,
-                  help="Start Spyder with a clean settings directory")
+parser.add_option('--safe-mode', dest="safe_mode",
+                  action='store_true', default=False,
+                  help="Start Spyder with a clean configuration directory")
 parser.add_option('--no-apport', action='store_true',
                   default=False, help="Disable Apport exception hook (Ubuntu)")
 parser.add_option('--debug', action='store_true',
@@ -58,8 +59,8 @@ assert options.gui in (None, 'pyqt5', 'pyqt', 'pyside'), \
        "Invalid GUI toolkit option '%s'" % options.gui
 
 # Start Spyder with a clean configuration directory for testing purposes
-if options.clean:
-    os.environ['SPYDER_CLEAN'] = 'True'
+if options.safe_mode:
+    os.environ['SPYDER_SAFE_MODE'] = 'True'
 
 # Prepare arguments for Spyder's main script
 sys.argv = [sys.argv[0]] + args
@@ -168,8 +169,8 @@ if options.hide_console and os.name == 'nt':
     print("0x. Hiding parent console (Windows only)")
     sys.argv.append("--hide-console")  # Windows only: show parent console
 
-# Reset temporary config directory if in --clean mode
-if options.clean:
+# Reset temporary config directory if starting in --safe-mode
+if options.safe_mode:
     from spyder.config.base import get_conf_path  # analysis:ignore
     conf_dir = get_conf_path()
     if osp.isdir(conf_dir):
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -170,7 +170,7 @@ if options.hide_console and os.name == 'nt':
     sys.argv.append("--hide-console")  # Windows only: show parent console
 
 # Reset temporary config directory if starting in --safe-mode
-if options.safe_mode:
+if options.safe_mode or os.environ.get('SPYDER_SAFE_MODE'):
     from spyder.config.base import get_conf_path  # analysis:ignore
     conf_dir = get_conf_path()
     if osp.isdir(conf_dir):
--- a/spyder/config/base.py
+++ b/spyder/config/base.py
@@ -124,7 +124,7 @@ def get_home_dir():
 def get_conf_path(filename=None):
     """Return absolute path for configuration file with specified filename"""
     # Define conf_dir
-    if running_under_pytest():
+    if running_under_pytest() or TEST:
         import py
         from _pytest.tmpdir import get_user
         conf_dir = osp.join(str(py.path.local.get_temproot()),
@@ -144,7 +144,7 @@ def get_conf_path(filename=None):
 
     # Create conf_dir
     if not osp.isdir(conf_dir):
-        if running_under_pytest():
+        if running_under_pytest() or TEST:
             os.makedirs(conf_dir)
         else:
             os.mkdir(conf_dir)
@@ -152,7 +152,7 @@ def get_conf_path(filename=None):
         return conf_dir
     else:
         return osp.join(conf_dir, filename)
-        
+
 
 def get_module_path(modname):
     """Return module *modname* base path"""
--- a/spyder/config/base.py
+++ b/spyder/config/base.py
@@ -34,9 +34,9 @@ from spyder.py3compat import (is_unicode, TEXT_TYPES, INT_TYPES, PY3,
 # SPYDER_DEV is (and *only* has to be) set in bootstrap.py
 DEV = os.environ.get('SPYDER_DEV')
 
-# For testing purposes
-# SPYDER_TEST can be set using the --test option of bootstrap.py
-TEST = os.environ.get('SPYDER_TEST')
+# Make Spyder use a temp clean configuration directory for testing purposes
+# SPYDER_CLEAN can be set using the --clean option of bootstrap.py
+CLEAN = os.environ.get('SPYDER_CLEAN')
 
 
 def running_under_pytest():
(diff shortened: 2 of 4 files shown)


CAM-Gerlach:
Wrt @pep8speaks , I was just following the convention already in the file for all other spyder imports, it is likely needed for safety, and I added # analysis:ignore for a good measure.

bcolsen:
I tested it on my home computer and it seem to work well.  

I'll test it a work tomorrow against the 10 mb history.py :-)

CAM-Gerlach:
> I'll test it a work tomorrow against the 10 mb history.py :-)

I'm not sure I understand how the size of any ``history.py`` file in the normal Spyder config dir would have any effect on ``--test`` mode whatsoever. Either it works (using the temp config dir), or it doesn't (it still uses the main one), which would be obvious already on your home machine assuming you had anything at all in the history before.

bcolsen:
Sorry bad joke! You are right the file size would clearly have no affect.

I just had another idea.  We could add this to the main Spyder executable as well to make trouble shooting easier.  This would be like having a "safe mode" like Firefox has. It's like testing a Spyder reset, but if it doesn't fix the problem you don't lose your settings. We could even call it --safe-mode.

CAM-Gerlach:
> Sorry bad joke! You are right the file size would clearly have no affect.

I've certainly had more than my share of those, so I'm not one to talk...

> I just had another idea. We could add this to the main Spyder executable as well to make trouble shooting easier.

Great idea! Shouldn't be hard to do; the only thing is that normal Spyder and ``bootstrap.py`` use two separate argument parsers, so we'd have to either stick it in both (very easy to do, but not exactly DRY) or refactor that, and move the small block of code that wipes the temp config to near the beginning of ``start.py``. @ccordoba12 , agree?

CAM-Gerlach:
The broader change I would suggest is *unifying the bootstrap parser into the main ``cli_options`` ``get_options()`` parser since the great majority if not all of their options are useful to either start mode*; in particular, this relates to your #7123 , since you can't actually set ``spyder --debug`` with normal Spyder even though that's what we've been telling a lot of users to help figure out problems (we've instead had to have them manually set the `SPYDER_DEBUG` environment variable to `3`, in a platform unportable way). 

*If we needed a few different options for each, we could either add an parameter to ``get_options()`` in ``cli_options`` to tell it which set of extra options to add, or we could just add them to the parser returned by ``get_options()`` in ``start.py`` and ``boostrap.py`` respectively.* Then, the only other thing we'd need to change is to have start.main() take one parameter, ``options`` or the like, with a default value of None that bootstrap.py could pass the rest of the parsed options that start.py would need, and then enclose the current parser block in ``start.py`` with ``if options is None:`` which would replicate the current behavior by default.

However, the one complexity here is that Spyder 4 was converted over to use ``argparse``, while Spyder 3 is still stuck on the deprecated ``argparse``, which will make more complicated, at least for @ccordoba12 when merging it to ``master``. We could do this just in ``master``, but since its not really a feature but rather an important change helping users find and solve bugs and problems, and would make our lives a good deal easier maintaining Spyder 3, I would think we would want to get it out to them as soon as possible. *So, we can either keep things using the deprecated optparse in Spyder 3 and Carlos can fix any conflicts merging to master, or as part of this we can backport the argparse change.*

@ccordoba12 , opinions? I italicized the two portions most in need of your input so you could find them easily.

spyder/config/base.py
```diff
@@ -144,15 +144,15 @@ def get_conf_path(filename=None):
...
     # Create conf_dir
     if not osp.isdir(conf_dir):
-        if running_under_pytest():
+        if running_under_pytest() or TEST:
             os.makedirs(conf_dir)
         else:
             os.mkdir(conf_dir)
     if filename is None:
         return conf_dir
-    else:
```
ccordoba12: This needs to stay, so please revert this change. It's how we retrieve files placed placed under our config directory, like this `get_conf_path('history.py')`

ccordoba12:
> This would be like having a "safe mode" like Firefox has. It's like testing a Spyder reset, but if it doesn't fix the problem you don't lose your settings. We could even call it --safe-mode.

We're using pytest for this, so this we'll need to improve our current implementation.

> The broader change I would suggest is unifying the bootstrap parser into the main cli_options get_options() parser since the great majority if not all of their options are useful to either start mode

I don't know about this. I'll have a PR with this change, but the initial idea behind different cli options for both is that boostrap options don't apply to the main Spyder script and vice versa.

ccordoba12:
> I didn't mean to have this setting for unit testing but as a convenient way for users with bugs to troubleshoot their local install problems without deleting all their settings

I understood that. So, if we want to add an option to use a clean prefs directory as part of a regular Spyder installation, we'd need to not use pytest for the implementation because then pytest would become a Spyder dependency, and we don't want that.

bcolsen:
Sorry for my confusion. I get it now, thanks.  It will require more work than I thought.

spyder/config/base.py
CAM-Gerlach: I can, but there is no functional difference to dropping the ``else,`` it works exactly the same (or I would have noticed immediately since this was exactly the functionality I was testing)—did you scroll down a few lines? if filename is not None, it just executes the next statement, which is ``return osp.join(conf_dir, filename)`` exactly as before. I just cleaned it up while I was at it since it was in the same function I made other more substantial changes to.

CAM-Gerlach:
> So, if we want to add an option to use a clean prefs directory as part of a regular Spyder installation, we'd need to not use pytest for the implementation because then pytest would become a Spyder dependency, and we don't want that.

Good point, I didn't think of that. In fact, since pytest (and py, which is also needed) aren't dependencies for a basic source install either, even in its current form its a problem. 

However, upon further investigation, it actually would be quite simple (and both in scope and necessary for just this PR even, given the above) as well as probably optimal overall to just reimplement the two relevant functions (one from ``pytest`` and one from ``py``), because both packages are imported locally just for those two functions and used nowhere else in the file, and both functions are just one-line thin wrappers around standard library functions;

* ``py.path.local.get_temproot()`` is a oneliner that just ``import``s ``tempfile``,  calls ``tempfile.gettempdir()`` and then casts it to a py.path.local convenience wrapper, which in our code is then immediately casted back to ``str`` making it pointless and wasteful to use the wrapper instead of just ``tempfile.gettempdir()`` directly.
* ``_pytest.tmpdir.get_user`` is a private (so should we even be relying on it?) wrapper for ``getpass.getuser()`` that just calls it in a try-except block, returning ``None`` if it doesn't work due to a narrow set of conditions (Windows + a ``tox`` based CI that strips env variables). It would be trivial to include just try-except in ``base.py``, if it is even really necessary in our use case, or even including the user name to begin with—on Windows, the one platform where the standard library get_user() could fail, each user has their own temp directory such that including the username is unnecessary anyway

Therefore, I'll implement those two minor changes here, and we can go from there.

>  I'll have to see a PR with this change, but the initial idea behind different cli options for both is that boostrap options don't necessarily apply to the main Spyder script and vice versa.

I'll submit a separate issue and PR. Looking through the existing CLI options for normal Spyder, it looks like pretty much every single one of them would be applicable to ``boostrap.py`` as far as I can tell; on the other hand, two of the six in bootstrap.py's are identical to those in ``cli_options``, (show and hide console), two would be very useful to have there (``test`` and ``debug``), and two are ``bootstrap`` specific (``no-apport`` and ``gui``). 

Therefore, would seem the best course of action might be to add the two commonly useful ones to ``cli_options``, do a small refactor to get_options so it returns the parser rather than the evaluated options (its only one extra line, and IIRC a best practice anyway), and then add the two extra options in ``boostrap.py``, passing the evaluated options on to main(). All in all, maybe a dozen lines worth of changes.

CAM-Gerlach:
Okay, I implemented what I discussed in the first part of previous post to remove the dependencies on ``py`` and ``pytest`` in the relevant portions; overall its actually a small net improvement to do so even aside from the proximate reason. I tested it with and without ``clean``, doing various activities in Spyder and restarting several times, and everything related to the change works well. Therefore, this PR looks ready on my end.

With regard to the second point, I figured out it already works to call ``bootstrap.py --clean -- --spyder-options-here`` so there seems little point to doing most of what I suggest above. However, this just means we can move ``debug`` and ``clean`` over to the main parser, handle them appropriately in ``start.py`` and/or ``mainwindow`` instead of ``boostrap.py``, and be done with it—I'll just need to make sure they "stick" before they're needed. I'll open another issue/PR to take a crack at that.

bcolsen:
I tested this and it works on linux :+1:

spyder/config/base.py
ccordoba12: This is a critical part of our settings infrastructure, so I don't want to play with it simply for correctness, so please revert this change.

CAM-Gerlach: Okay, done as requested. 

In general, I abstain from making these sorts of changes (that aren't purely code style) for the reasons you allude to, namely the small but non-negligible chance of breaking something that's rarely worth the lesser benefit from cleaner code. In this case I ended up doing so since I'd checked it with both my own knowledge and the static analyzer, and the fact that the unit tests and numerous real world-tests exercising that block of code all had no issues, but it obviously wasn't worth doing and I'll be even more conservative in the future.

CAM-Gerlach:
@ccordoba12 Addressed your comment; thanks.

bootstrap.py
```diff
@@ -39,8 +42,8 @@
                   "is to show the console")
 parser.add_option('--hide-console', action='store_true',
                   default=False, help="Hide parent console window (Windows only)")
-parser.add_option('--test', dest="test", action='store_true', default=False,
-                  help="Test Spyder with a clean settings dir")
+parser.add_option('--clean', dest="clean", action='store_true', default=False,
```
ccordoba12: Please change this to `--clean-config`. The problem is `--clean` is too close in meaning to `--reset`.

bootstrap.py
```diff
@@ -54,9 +57,9 @@
 assert options.gui in (None, 'pyqt5', 'pyqt', 'pyside'), \
        "Invalid GUI toolkit option '%s'" % options.gui
 
-# For testing purposes
-if options.test:
-    os.environ['SPYDER_TEST'] = 'True'
+# Start Spyder with a clean configuration directory for testing purposes
+if options.clean:
```
ccordoba12: This should be `options.clean_config`

bootstrap.py
```diff
@@ -54,9 +57,9 @@
...
+    os.environ['SPYDER_CLEAN'] = 'True'
```
ccordoba12: `SPYDER_CLEAN` -> `SPYDER_CLEAN_CONFIG`

spyder/config/base.py
```diff
@@ -34,9 +34,9 @@
 # SPYDER_DEV is (and *only* has to be) set in bootstrap.py
 DEV = os.environ.get('SPYDER_DEV')
 
-# For testing purposes
-# SPYDER_TEST can be set using the --test option of bootstrap.py
-TEST = os.environ.get('SPYDER_TEST')
+# Make Spyder use a temp clean configuration directory for testing purposes
+# SPYDER_CLEAN can be set using the --clean option of bootstrap.py
+CLEAN = os.environ.get('SPYDER_CLEAN')
```
ccordoba12: This should be

    CLEAN_CONFIG = os.environ.get('SPYDER_CLEAN_CONFIG')

spyder/config/base.py
```diff
@@ -122,13 +122,18 @@ def get_home_dir():
...
+    if running_under_pytest() or CLEAN:
+        # Use clean config dir if running tests or the user requests it.
+        import getpass  # analysis:ignore
+        if sys.platform.startswith("win"):
+            current_user = ''
+        else:
+            current_user = '-' + str(getpass.getuser())
+        import tempfile  # analysis:ignore
+        conf_dir = osp.join(str(tempfile.gettempdir()),
+                            'pytest-spyder{0!s}'.format(current_user),
```
ccordoba12: Let's move this whole block to a new function called `clean_conf_dir`

spyder/config/base.py
```diff
@@ -144,15 +149,15 @@ def get_conf_path(filename=None):
...
-        if running_under_pytest():
+        if running_under_pytest() or CLEAN:
             os.makedirs(conf_dir)
         else:
             os.mkdir(conf_dir)
     if filename is None:
         return conf_dir
     else:
         return osp.join(conf_dir, filename)
-
```
ccordoba12: Please avoid adding/removing blanks in the 3.x branch. It makes my life harder when merging with `master` and also with `split-plugins`. All pep8 stuff needs to be done in `master`, not here, because this branch is in maintenance now.

spyder/config/main.py
```diff
@@ -12,21 +12,21 @@
...
 from spyder.config.user import UserConfig
 from spyder.config.utils import IMPORT_EXT
 from spyder.utils import codeanalysis
 
 
-#==============================================================================
+# =============================================================================
 # Main constants
-#==============================================================================
+# =============================================================================
```
ccordoba12: The same goes for this. No worries in this PR, but please avoid it in the future.

bootstrap.py
CAM-Gerlach: Good point @ccordoba12 . On that note, ``--clean-config`` could also be misinterpreted in the same way, i.e. that the option would ``clean`` the ``config`` i.e. what ``reset`` does. What about ``--clean-mode``? That would make it much clearer that its temporary and distinct from ``reset``.

ccordoba12:
I left several new comments because I don't like `--clean` as an option.

spyder/config/base.py
CAM-Gerlach: Sorry; my mistake. It must have happened when I did the ``else:`` change; one more reason to avoid that sort of thing I suppose.

spyder/config/main.py
CAM-Gerlach: Okay, noted and will now do so from here on out.

bootstrap.py
ccordoba12: Firefox call it `--safe-mode`. What about that?

ccordoba12: You're right about `--clean-config`, by the way.

spyder/config/base.py
CAM-Gerlach: Will do, thanks.

bootstrap.py
CAM-Gerlach: Done.

bootstrap.py
CAM-Gerlach: Changed to ``options.safe_mode``.

bootstrap.py
CAM-Gerlach: Changed to ``SPYDER_SAFE_MODE``.

spyder/config/base.py
CAM-Gerlach: Changed to ``SAFE_MODE``/``SPYDER_SAFE_MODE`` respectively.

spyder/config/base.py
CAM-Gerlach: Done; moved to ``get_clean_config_dir``.

ccordoba12:
Ok, no problem. Good catch!

PR #7132, opened 2018-05-15
PR: Limit the number of lines in history according to the relevant setting

### LINKED ISSUE #7080: Spyder is slow to start due to big history file
When starting, spyder reads a 10Mo text file in ~/.spyder-py3 called history.py.
That takes ~15s, which is a painfully noticeable start-up time. 
Is there an option to limit how far back I want to load the history?

The easy work-around is for me to edit manually this file, but if an user is not aware of this, he might get annoyed by the start-up time.
I can go from 22s to 6s start-up time, which is not fast but way less annoying.

Also closing spyder takes almost 9s. I couldn't figure out why. If I have to restart spyder, usually I just kill it to avoid having to wait.

Python 3.6.5 64bits, Qt 5.10.1, PyQt5 5.10.1 on Darwin 
OSX 10.13.4

impact27:
### Pull Request Checklist

* [x] ~~Included a screenshot, if this PR makes any visible changes to the UI~~

## Description of Changes

A better thing would be to read the history from ipython, but I am not sure that this would work in every cases.

### Issue(s) Resolved

Fixes #7080

### DIFF
--- a/spyder/plugins/history.py
+++ b/spyder/plugins/history.py
@@ -235,7 +235,7 @@ class HistoryLog(SpyderPluginWidget):
         linebreaks = [m.start() for m in re.finditer('\n', text)]
         maxNline = CONF.get('historylog', 'max_entries')
         if len(linebreak) > maxNline:
-            text = text[linebreaks[-maxNline] +1:]
+            text = text[linebreaks[-maxNline] + 1:]
             encoding.write(text, filename)
         editor.set_text(text)
         editor.set_cursor_position('eof')
--- a/spyder/plugins/history.py
+++ b/spyder/plugins/history.py
@@ -236,7 +236,7 @@ class HistoryLog(SpyderPluginWidget):
         linebreaks = [m.start() for m in re.finditer('\n', text)]
         maxNline = self.get_option('max_entries')
         if len(linebreaks) > maxNline:
-            text = text[linebreaks[-maxNline] + 1:]
+            text = text[linebreaks[-maxNline - 1] + 1:]
             encoding.write(text, filename)
         editor.set_text(text)
         editor.set_cursor_position('eof')
--- a/spyder/plugins/history.py
+++ b/spyder/plugins/history.py
@@ -30,7 +30,6 @@ from spyder.utils.qthelpers import (add_actions, create_action,
 from spyder.widgets.tabs import Tabs
 from spyder.widgets.sourcecode import codeeditor
 from spyder.widgets.findreplace import FindReplace
-from spyder.config.main import CONF
 
 
 class HistoryConfigPage(PluginConfigPage):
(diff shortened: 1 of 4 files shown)


CAM-Gerlach:
Thanks for doing this! I need to head to work now, but a few quick things:

* Please briefly describe what your PR actually does in the title and the description, so we can follow it easily
* Fix the PEP8 issue
* The test failures are legitimate, and it looks like its due to a typo in your code (see comment). It's easy to make, but if you're using Spyder or had actually tested the change, it would have been easy to catch
* Can you add a test for this?
* Not sure why you deleted the PR template that specified several of these things in a checklist...

spyder/plugins/history.py
```diff
@@ -230,6 +232,11 @@ def add_history(self, filename):
         editor.toggle_wrap_mode( self.get_option('wrap') )
 
         text, _ = encoding.read(filename)
+        linebreaks = [m.start() for m in re.finditer('\n', text)]
+        maxNline = CONF.get('historylog', 'max_entries')
+        if len(linebreak) > maxNline:
```
CAM-Gerlach: This is where all the tests are failing--it should be ``linebreaks`` presumably. Easy mistake to make, but one a basic code checker or a simple test (unit test or just actually testing that the functionality worked without error) would have caught.

spyder/plugins/history.py
```diff
@@ -29,6 +30,7 @@
 from spyder.widgets.tabs import Tabs
 from spyder.widgets.sourcecode import codeeditor
 from spyder.widgets.findreplace import FindReplace
+from spyder.config.main import CONF
```
ccordoba12: Please remove this. It's not needed.

spyder/plugins/history.py
```diff
@@ -230,6 +232,11 @@ def add_history(self, filename):
         editor.toggle_wrap_mode( self.get_option('wrap') )
 
         text, _ = encoding.read(filename)
+        linebreaks = [m.start() for m in re.finditer('\n', text)]
+        maxNline = CONF.get('historylog', 'max_entries')
```
ccordoba12: Plugins can access config their own options like this

    maxNline = self.get_option('max_entries')

ccordoba12:
Things look good, I just left a minor comment.

CAM-Gerlach:
@ccordoba12 Do we want a regression test for this to ensure the change is and remains effective?

ccordoba12:
A test would be nice, yes, but it doesn't seem easy to implement. Besides, the code looks good.

spyder/plugins/history.py
```diff
@@ -230,6 +231,11 @@ def add_history(self, filename):
         editor.toggle_wrap_mode( self.get_option('wrap') )
 
         text, _ = encoding.read(filename)
+        linebreaks = [m.start() for m in re.finditer('\n', text)]
```
ccordoba12: We need to convert all EOL's to `\n` before doing this.

CAM-Gerlach: Can't we just always use ``\n`` for our own history files, independent of the platform?

PR #7141, opened 2018-05-17
PR: Load history from IPython directly

### LINKED ISSUE #7080: Spyder is slow to start due to big history file

### LINKED ISSUE #7142: The iPython console history is duplicated
### Issue Report Checklist

* [ x] Searched the [issues page](https://github.com/spyder-ide/spyder/issues?q=is%3Aissue) for similar reports
* [ x] Read the relevant sections of the [Spyder Troubleshooting Guide](https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ) and followed its advice
* [ x] Reproduced the issue after updating with ``conda update spyder`` (or ``pip``, if not using Anaconda)
* [x ] Could not reproduce inside ``jupyter qtconsole`` (if console-related)
    * [ ] Restarted Spyder
* [x ] Completed the **Problem Description**, **Steps to Reproduce** and **Version** sections below

## Problem Description
The history is saved both as a python file in "~/.spyder-py3/history.py" and as a database in "~/.ipython/profile_default/history.sqlite" (Or whatever the setting of ipython is).

## Versions
Python 3.6.5 64bits, Qt 5.9.1, PyQt5 5.9 on Darwin

impact27:
### Pull Request Checklist

## Description of Changes
Spyder now loads the history from iPython instead of duplicating the history in a local file.
That is less confusing for the user as the history in iPython and the history in historylog are now the same.

### Issue(s) Resolved
Fixes #7142

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -1331,7 +1331,9 @@ class MainWindow(QMainWindow):
 
         # Show history file if no console is visible
         if not self.ipyconsole.isvisible:
-            self.historylog.add_history(get_conf_path('history.py'))
+            self.historylog.add_history(
+                    self.ipyconsole.history_title,
+                    self.ipyconsole.get_current_shellwidget()._history)
 
         if self.open_project:
             self.projects.open_project(self.open_project)
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -1792,6 +1792,38 @@ def test_render_issue():
     assert test_description in test_issue_2
     assert test_traceback in test_issue_2
 
+@flaky(max_runs=3)
+@pytest.mark.slow
+def test_history(main_window, qtbot):
+    """Test History"""
+    # ---- Setup ----
+    # Create 3 clients and wait until the window is fully up
+    for i in range(3):
+        main_window.ipyconsole.create_new_client()
+
+    clients = main_window.ipyconsole.get_clients()
+    for client in clients:
+        shell = client.shellwidget
+        qtbot.waitUntil(lambda: shell._prompt_html is not None, timeout=SHELL_TIMEOUT)
+
+    #Execute some code in each widget
+    commands = []
+    for i, client in enumerate(clients):
+        shell = client.shellwidget
+        commands.append('i = {i}\n'.format(i=i))
+        shell.execute(commands[-1])
+        qtbot.waitUntil(lambda: shell._prompt_html is not None, timeout=SHELL_TIMEOUT)
+        print(main_window.historylog.editors[0].toPlainText()[-10:])
+
+    commands_text = ''.join(commands)
+
+    index = main_window.historylog.filenames.index(main_window.ipyconsole.history_title)
+    assert index == 0 
+    text = main_window.historylog.editors[index].toPlainText()[-len(commands_text):]
+    history = main_window.historylog.histories[index][-len(clients):]
+
+    assert history == commands
+    assert text == commands_text
 
 @pytest.mark.slow
 @flaky(max_runs=3)
--- a/spyder/config/base.py
+++ b/spyder/config/base.py
@@ -503,7 +503,7 @@ def running_in_mac_app():
 #==============================================================================
 SAVED_CONFIG_FILES = ('help', 'onlinehelp', 'path', 'pylint.results',
                       'spyder.ini', 'temp.py', 'temp.spydata', 'template.py',
-                      'history.py', 'history_internal.py', 'workingdir',
+                      'history_internal.py', 'workingdir',
                       '.projects', '.spyproject', '.ropeproject',
                       'monitor.log', 'monitor_debug.log', 'rope.log',
                       'langconfig', 'spyder.lock')
--- a/spyder/plugins/console/plugin.py
+++ b/spyder/plugins/console/plugin.py
@@ -103,12 +103,6 @@ class Console(SpyderPluginWidget):
         self.dismiss_error = False
 
     #------ Private API --------------------------------------------------------
-    def set_historylog(self, historylog):
-        """Bind historylog instance to this console
-        Not used anymore since v2.0"""
-        historylog.add_history(self.shell.history_filename)
-        self.shell.append_to_history.connect(historylog.append_to_history)
-
     def set_help(self, help_plugin):
         """Bind help instance to this console"""
         self.shell.help = help_plugin
--- a/spyder/plugins/history/plugin.py
+++ b/spyder/plugins/history/plugin.py
@@ -50,6 +50,7 @@ class HistoryLog(SpyderPluginWidget):
 
         self.editors = []
         self.filenames = []
+        self.histories = []
 
         # Initialize plugin actions, toolbutton and general signals
         self.initialize_plugin()
@@ -89,14 +90,14 @@ class HistoryLog(SpyderPluginWidget):
     def get_plugin_icon(self):
         """Return widget icon."""
         return ima.icon('history')
-    
+
     def get_focus_widget(self):
         """
         Return the widget to give focus to when
         this plugin's dockwidget is raised on top-level
         """
         return self.tabwidget.currentWidget()
-        
+
     def closing_plugin(self, cancelable=False):
         """Perform actions before parent main window is closed"""
         return True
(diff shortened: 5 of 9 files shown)


ccordoba12:
@impact27, looks good at first glance, but we really need a test for this because it's a major internal change. You need to add it to `test_mainwindow.py` (because it involves two plugins), and it could be very simple: evaluate a command in the console and register that the last entry in the History Log corresponds to that command.

Also, please create a new issue for this because #7080 is really solved by your other PR.

bcolsen:
I did some tests on this and it works.  I think this is the way forward.  Having the history stored per entered commands in this way could lead to great new functionality in the history explorer.

The only regression is the old history had date headers, but they weren't very useful.

ccordoba12:
@impact27, please also migrate this to the master branch.

impact27:
I opened #7142

impact27:
The dates are saved in the ipython database so It could be added back.

CAM-Gerlach:
Thanks @impact27 . FYI you should ``rebase`` your original work on ``master`, not ``merge`` it, so as to not dirty the history. You can get the original back with ``reflog`` in case you don't have it already in a branch, or an interactive ``rebase -i``.

Also, apparently AppVeyor on Py3 isn't able to find ``pytest-mock`` to install it and is thus failing; not sure why. @ccordoba12 ?

CAM-Gerlach:
@pep8speaks suggest diff for testing purposes

bcolsen:
I've been testing this and it's working great.

Summary of text below: I think it would be best to have one history tab per console. 

On a more broader view it seems like there is a lot of code here that seems to be used to get all the ipython consoles to talk to one history tab so it mimics the way the previous code worked. 

From a use case prospective I think this isn't a good idea to have the history mixed and it would be better and simpler to have IPython take care of the history entirely. We can just take what the console gives not worry about it...it works great. This would require us to have one history tab per console, but luckily the history widget already has tabs implemented.

Currently, If you change between tabs and run commands in each the history gives you a mess of code that is most likely unassociated and you can tell what commands were executed in each console. The up arrow history at the ipython prompt keeps the history separate and when both sessions are closed the two histories are appended into the main history one console after the other so the code is still in blocks that are still associated.

The hardest part would be to synchronize the tab switching between the consoles and the history.

I would open a new issue but this idea would greatly impact this work.

ccordoba12:
@impact27, as @CAM-Gerlach suggested, please rebase this PR in top of latest master. Your merge was really unnecessary.

ccordoba12:
> I've been testing this and it's working great.

Great!

> Currently, If you change between tabs and run commands in each the history gives you a mess of code that is most likely unassociated and you can tell what commands were executed in each console.

What if we use a QStackedWidget instead to only display the history of the current console, but change to another history when users select a different console? That's what we do in the Variable Explorer and I think it works quite well.

> I would open a new issue but this idea would greatly impact this work.

Please do. We can work on it after this PR is merged.

CAM-Gerlach:
> What if we use a QStackedWidget instead to only display the history of the current console, but change to another history when users select a different console? That's what we do in the Variable Explorer and I think it works quite well.

The only disadvantage there, that isn't the case for the variable explorer, is if a user wants to view or copy one console's history while working in another. It could be solved by a "lock" toggle locking the history shown to that of the currently selected console (like the one for Help). However, we'd then need to implement tabs in the history log like we have in the console to navigate between different console's logs independently, so not sure if the different use cases that require it would be worth the effort at this point.

ccordoba12:
@CAM-Gerlach, we can still use a QStackedWidget widget and an option that allows users to select the console's history they want to see at the moment (but that would be change to the current console as soon as they change consoles).

I really don't want to use tabs for this because I really think it makes no sense. In any case, please stop discussing that in this PR (because it's unrelated to @impact27's work) and open a new issue about it.

CAM-Gerlach:
Sorry :open_mouth: 

@bcolsen was the one with the detailed proposal above and said he could create the new issue, so I didn't (and don't) want to just steal it from him.

ccordoba12:
> Sorry

No worries, I just didn't us to keep polluting this PR ;-)

impact27:
I rebased the pull request

spyder/app/mainwindow.py
```diff
@@ -1261,7 +1261,9 @@ def post_visible_setup(self):
 
         # Show history file if no console is visible
         if not self.ipyconsole.isvisible:
-            self.historylog.add_history(get_conf_path('history.py'))
+            self.historylog.add_history(
+                    "iPython history",
```
ccordoba12: Let's call this simply `History`.

CAM-Gerlach: Also, its a moot point now but just for the record, its ``IPython``—they are actually pretty emphatic on that distinction.

spyder/plugins/console.py
```diff
@@ -101,12 +101,6 @@ def __init__(self, parent=None, namespace=None, commands=[], message=None,
         self.dismiss_error = False
 
     #------ Private API --------------------------------------------------------
-    def set_historylog(self, historylog):
-        """Bind historylog instance to this console
-        Not used anymore since v2.0"""
-        historylog.add_history(self.shell.history_filename)
-        self.shell.append_to_history.connect(historylog.append_to_history)
-
```
ccordoba12: This needs to stay. It's used to save and load history in the internal console. So please put it back it.

spyder/plugins/history.py
```diff
@@ -205,37 +205,34 @@ def move_tab(self, index_from, index_to):
...
         editor = self.editors.pop(index_from)
-        
+
         self.filenames.insert(index_to, filename)
         self.editors.insert(index_to, editor)
-        
+
     #------ Public API ---------------------------------------------------------
-    def add_history(self, filename):
+    def add_history(self, filename, history_list):
```
ccordoba12: `history_list` must be optional, so that history in the internal console can still work.

spyder/plugins/ipythonconsole.py
```diff
@@ -1310,9 +1309,19 @@ def register_client(self, client, give_focus=True):
 
         # Connect client to our history log
         if self.historylog is not None:
-            self.historylog.add_history(client.history_filename)
-            client.append_to_history.connect(self.historylog.append_to_history)
-        
+            self.historylog.add_history(
+                    "iPython history",
```
ccordoba12: Please change this tp `History`

ccordoba12:
@impact27, this looks good, but you need to preserve saving history in the internal console.

impact27:
I am not sure where set_historylog is actually called? The only reference I found in the code was already commented out. Are you sure the internal history is not saved? It is working as is in my code.

CAM-Gerlach:
@impact27 Looks like this got hit big-time by the split-plugins merge. Are you interested in finishing it? 

With regard to your question, I'm not really sure why there need be a tab in the History log for Internal Console history, considering it is only used for testing and introspecting Spyder itself rather than production work, and can in any case still be accessed inside the internal console's one pane anyway, but even so it looks like it still works just fine.

impact27:
Unfortunately I don't really have time these days. Does somebody else want to finish that?

ccordoba12:
I'll finish it. I'll implement the QStackedWidget solution I proposed some time ago.

CAM-Gerlach:
@ccordoba12 @impact27 Hey, is this ever going to be finished? A lot of conflicting files have built up over the time this has been sitting around...

goanpeca:
@ccordoba12 thoughts on this? Do you want some of the other devs to tackle this at some point?

ccordoba12:
We'll take another look after we release beta2.

ccordoba12:
@impact27, could you update and finish this PR? For now let's go with my proposed approach, i.e. showing the history of the current console when switching from one to another.

impact27:
I rebase it on the master and it still works, but I am not sure how to do the `QStackedWidget`, so I would be glad if someone else could finish it.

ccordoba12:
> I rebase it on the master and it still works

Thanks a lot for that!

> but I am not sure how to do the QStackedWidget

Ok, no problem. We'll help you with that part then.

impact27:
Closing because too many conflicts accumulated and this would need a rewrite

PR #7202, opened 2018-05-25
PR: Change behavior of Variable Explorer editor buttons

### LINKED ISSUE #7156: Change behavior of the Variable Explorer editors buttons
The problem is right now the Ok button sends a serialized copy of the viewed object to the kernel, even if the user has not made changes to it.

dalthviz:
### Pull Request Checklist

## Description of Changes

Change behavior of default `OK` and `Cancel` buttons. Now `OK` is `Save and close` (disabled by default) and `Cancel` is the `Close` button.

### Issue(s) Resolved

Fixes #7156

### DIFF
--- a/spyder/widgets/variableexplorer/arrayeditor.py
+++ b/spyder/widgets/variableexplorer/arrayeditor.py
@@ -606,6 +606,8 @@ class ArrayEditor(QDialog):
         self.arraywidget = None
         self.stack = None
         self.layout = None
+        self.btn_apply = None
+        self.btn_ok = None
         # Values for 3d array editor
         self.dim_indexes = [{}, {}, {}]
         self.last_dim = 0  # Adjust this for changing the startup dimension
@@ -727,12 +729,14 @@ class ArrayEditor(QDialog):
                                    "to masked array won't be reflected in "\
                                    "array's data (and vice-versa)."))
                 btn_layout.addWidget(label)
-            btn_layout.addStretch()
 
-        self.btn_apply = QPushButton(_('Apply'))
-        self.btn_apply.setDisabled(True)
-        self.btn_apply.clicked.connect(self.accept)
-        btn_layout.addWidget(self.btn_apply)
+        btn_layout.addStretch()
+
+        if not readonly:
+            self.btn_apply = QPushButton(_('Apply'))
+            self.btn_apply.setDisabled(True)
+            self.btn_apply.clicked.connect(self.accept)
+            btn_layout.addWidget(self.btn_apply)
 
         self.btn_ok = QPushButton(_('OK'))
         self.btn_ok.setAutoDefault(True)
@@ -750,9 +754,10 @@ class ArrayEditor(QDialog):
 
     def apply_enable(self):
         """Handle the data change event to enable the apply button."""
-        self.btn_apply.setEnabled(True)
-        self.btn_apply.setAutoDefault(True)
-        self.btn_apply.setDefault(True)
+        if self.btn_apply:
+            self.btn_apply.setEnabled(True)
+            self.btn_apply.setAutoDefault(True)
+            self.btn_apply.setDefault(True)
 
     def current_widget_changed(self, index):
         self.arraywidget = self.stack.widget(index)
--- a/spyder/widgets/variableexplorer/arrayeditor.py
+++ b/spyder/widgets/variableexplorer/arrayeditor.py
@@ -607,7 +607,7 @@ class ArrayEditor(QDialog):
         self.stack = None
         self.layout = None
         self.btn_save_and_close = None
-        self.btn_ok = None
+        self.btn_close = None
         # Values for 3d array editor
         self.dim_indexes = [{}, {}, {}]
         self.last_dim = 0  # Adjust this for changing the startup dimension
@@ -739,11 +739,11 @@ class ArrayEditor(QDialog):
             self.btn_save_and_close.clicked.connect(self.accept)
             btn_layout.addWidget(self.btn_save_and_close)
 
-        self.btn_ok = QPushButton(_('Close'))
-        self.btn_ok.setAutoDefault(True)
-        self.btn_ok.setDefault(True)
-        self.btn_ok.clicked.connect(self.reject)
-        btn_layout.addWidget(self.btn_ok)
+        self.btn_close = QPushButton(_('Close'))
+        self.btn_close.setAutoDefault(True)
+        self.btn_close.setDefault(True)
+        self.btn_close.clicked.connect(self.reject)
+        btn_layout.addWidget(self.btn_close)
         self.layout.addLayout(btn_layout, 2, 0)
 
         self.setMinimumSize(400, 300)
@@ -753,7 +753,8 @@ class ArrayEditor(QDialog):
         
         return True
 
-    def save_and_close_enable(self):
+    @Slot(QModelIndex, QModelIndex)
+    def save_and_close_enable(self, left_top, bottom_right):
         """Handle the data change event to enable the save and close button."""
         if self.btn_save_and_close:
             self.btn_save_and_close.setEnabled(True)
--- a/spyder/widgets/variableexplorer/arrayeditor.py
+++ b/spyder/widgets/variableexplorer/arrayeditor.py
@@ -524,6 +524,7 @@ class ArrayView(QTableView):
 
 
 class ArrayEditorWidget(QWidget):
+
     def __init__(self, parent, data, readonly=False,
                  xlabels=None, ylabels=None):
         QWidget.__init__(self, parent)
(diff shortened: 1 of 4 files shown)


dalthviz:
@ccordoba12 a preview:

imagen

jitseniesen:
I am not so sure about the names of the buttons. I believe that usually if you have an `OK` and an `Apply` button, the `Apply` button applies the changes without closing the window and the `OK` button applies the changes and closes the window, while here (if I understand it correctly), `OK` closes the window without applying the changes.

Maybe `Apply` and `Close` is better?

ccordoba12:
> Maybe Apply and Close is better?

Yes, I think you're right. @dalthviz, please apply @jitseniesen's suggestion.

CAM-Gerlach:
So, just to be clear, if I ``Apply`` the changes and then click ``Close``, the changes stay applied, correct? Also, the ``Close`` button has the exact same effect as the normal window close button (``X``), correct?

CAM-Gerlach:
As @jitseniesen also stated, the universally followed UI standard is that ``OK`` saves/commits the changes and closes the dialog, while ``Apply`` saves/commits but doesn't close the dialog. Therefore, if what you've stated is the behavior this PR currently implements, either one of the two possibilities should presumably be applied—changing the name of the buttons, or their exact function: 

* The ``Apply`` button should read ``OK`` and close the dialog, to conform with the universal convention, and, really, the ``Close`` button changed back to ``Cancel`` since it is both clearer about its own function (both closing the dialog and not saving the changes) and how it is different from ``Close``, as well as following the standard idiomatic pattern users expect (OK/Cancel, Apply/Close)
* Or, what I would think is better and more in the spirit of this PR, keeping the names but making the ``Apply`` button save the changes but *not* close the dialog, while the ``Close`` button closes the dialog without applying any unapplied changes, but keeping anything that has been applied (as it does currently). This makes more sense, I think, with what this PR is trying to do, and is clearer to users what each does vs. the previous buttons.

dalthviz:
@ccordoba12 `Apply` closes the dialog immediatly.

About the comments of @CAM-Gerlach maybe `OK` and `Cancel` are good or we could use something like `Save Changes` and `Cancel` to be explicit about the action of each button? The second suggestion is also valid, although it would need probably to override the closeEvent of the dialogs and some signal to trigger the save of the changes. What do you guys think @ccordoba12 @CAM-Gerlach @jitseniesen ?

jitseniesen:
I prefer CAM-Gerlach's second suggestion.

As an alternative which is probably less work and fixes the problem with that applying the changes can be expensive, and it is an expensive no-op if the user did not make any changes (which is quite common). Is that correct ... how about we leave everything as it is, except that if the user does not make any changes and clicks `OK`, the editor does not send a serialized copy of the unchanged object to the kernel?

CAM-Gerlach:
@dalthviz One option is keeping the current ``Apply`` button's function, renaming it ``OK``, and adding a another button named ``Apply`` that would save without closing, (like we have in our preferences and is pretty standard), but that last part might be outside the scope of this PR. 

> ``Save Changes`` 

Hmmm...well, in a way ``OK`` is actually clearer since it also makes clear that it will close the dialog; generally it is assumed that ``OK`` will save changes (as that is why it is distinct from ``Cancel``) and it is used everywhere for that purpose; e.g. in our very own preferences. Furthermore, ``OK`` and ``Cancel`` have better parallelism and clearer separation than ``Save Changes`` and ``Close``; while ``Apply`` and ``Close`` are also fairly clear.

> how about we leave everything as it is, except that if the user does not make any changes and clicks OK, the editor does not send a serialized copy of the unchanged object to the kernel?

Yeah; that's essentially what I was thinking of for my first option—simpler, though a little less clear/functional.

CAM-Gerlach:
> I don't like Cancel because the real operation is simply closing the dialog. So the Close button doesn't need to be changed.

Right, but the problem with the ``Close`` wording is that its closing without saving changes; users could assume their changes are automatically saved. More essentially, ``Close`` and ``OK`` are just rather confusing together.

To make it clearer with minimum effort and keeping the ``Close`` wording on that button, how about we just rename the ``Apply`` button ``Save and Close`` instead? Then it makes it very clear and explicit what it does, both by itself and relative to the other button.

ccordoba12:
> how about we just rename the Apply button Save and Close instead?

Ok, that's fine. @dalthviz, please implement this suggestion and also change the name of the variables that hold the editor buttons (e.g. `btn_apply` needs to be `btn_save_and_close`).

spyder/widgets/variableexplorer/arrayeditor.py
```diff
@@ -605,6 +606,8 @@ def __init__(self, parent=None):
         self.arraywidget = None
         self.stack = None
         self.layout = None
+        self.btn_save_and_close = None
+        self.btn_ok = None
```
ccordoba12: @dalthviz, this needs to be renamed to `btn_close`.

spyder/widgets/variableexplorer/collectionseditor.py
```diff
@@ -1321,6 +1325,8 @@ def __init__(self, parent=None):
 
         self.data_copy = None
         self.widget = None
+        self.btn_save_and_close = None
+        self.btn_ok = None
```
ccordoba12: Rename to `btn_close`

spyder/widgets/variableexplorer/collectionseditor.py
```diff
@@ -1374,6 +1389,13 @@ def setup(self, data, title='', readonly=False, width=650, remote=False,
         # Make the dialog act as a window
         self.setWindowFlags(Qt.Window)
 
+    def save_and_close_enable(self):
```
ccordoba12: Please add `@Slot()` above this line.

spyder/widgets/variableexplorer/dataframeeditor.py
```diff
@@ -722,15 +724,28 @@ def setup_and_check(self, data, title=''):
...
-        bbox.accepted.connect(self.accept)
-        bbox.rejected.connect(self.reject)
-        btn_layout.addWidget(bbox)
+
+        self.btn_save_and_close = QPushButton(_('Save and Close'))
+        self.btn_save_and_close.setDisabled(True)
+        self.btn_save_and_close.clicked.connect(self.accept)
+        btn_layout.addWidget(self.btn_save_and_close)
+
+        self.btn_ok = QPushButton(_('Close'))
```
ccordoba12: Rename to `btn_close`

spyder/widgets/variableexplorer/dataframeeditor.py
```diff
@@ -722,15 +724,28 @@ def setup_and_check(self, data, title=''):
...
+        self.btn_ok.setAutoDefault(True)
+        self.btn_ok.setDefault(True)
+        self.btn_ok.clicked.connect(self.reject)
+        btn_layout.addWidget(self.btn_ok)
 
         self.layout.addLayout(btn_layout, 2, 0)
 
         return True
 
+    def save_and_close_enable(self):
```
ccordoba12: This function needs to receive two parameters: `top_left` and `bottom_right` that account for the indexes passed by `dataChanged`.

ccordoba12: Also, above this method you need to add

`@Slot(QModelIndex, QModelIndex)`

spyder/widgets/variableexplorer/arrayeditor.py
```diff
@@ -737,9 +752,17 @@ def setup_and_check(self, data, title='', readonly=False,
         self.setWindowFlags(Qt.Window)
         
         return True
-            
+
+    def save_and_close_enable(self):
```
ccordoba12: Above this method you need to add

    @Slot(QModelIndex, QModelIndex)

ccordoba12: And this needs two parameters: `top_left` and `bottom_right`.

spyder/widgets/variableexplorer/texteditor.py
```diff
@@ -50,23 +52,31 @@ def __init__(self, text, title='', font=None, parent=None,
...
-        bbox.accepted.connect(self.accept)
-        bbox.rejected.connect(self.reject)
-        self.layout.addWidget(bbox)
-        
+            self.btn_save_and_close = QPushButton(_('Save and Close'))
+            self.btn_save_and_close.setDisabled(True)
+            self.btn_save_and_close.clicked.connect(self.accept)
+            btn_layout.addWidget(self.btn_save_and_close)
+
+        self.btn_ok = QPushButton(_('Close'))
```
ccordoba12: Rename to `btn_close`

spyder/widgets/variableexplorer/texteditor.py
```diff
@@ -50,23 +52,31 @@ def __init__(self, text, title='', font=None, parent=None,
 
         # Text edit
         self.edit = QTextEdit(parent)
-        self.edit.textChanged.connect(self.text_changed)
         self.edit.setReadOnly(readonly)
+        self.edit.textChanged.connect(self.text_changed)
```
ccordoba12: Please add `@Slot()` above the definition of the `text_changed` method.

PR #7234, opened 2018-05-31
PR: Call garbage collector from Main Window after closing Variable Explorer editors

### LINKED ISSUE #5830: Variable Explorer is leaking memory when editing variables

### LINKED ISSUE #7157: Ensure garbage collection is run in the main window when closing Variable Explorer editors
That way we could remove all memory used when passing objects from the kernel to Spyder.

dalthviz:
### Pull Request Checklist

## Description of Changes

Add signals to call the garbage collector from the Main Window after closing Variable Explorer editors.

### Issue(s) Resolved

Fixes #7157 
Fixes #5830

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -1955,7 +1955,6 @@ class MainWindow(QMainWindow):
         self._update_show_toolbars_action()
 
     # --- Other
-    @Slot()
     def free_memory(self):
         """Free memory after event."""
         gc.collect()
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -23,6 +23,7 @@ from __future__ import print_function
 
 import atexit
 import errno
+import gc
 import os
 import os.path as osp
 import re
@@ -1954,6 +1955,11 @@ class MainWindow(QMainWindow):
         self._update_show_toolbars_action()
 
     # --- Other
+    @Slot()
+    def free_memory(self):
+        """Free memory after event."""
+        gc.collect()
+
     def plugin_focus_changed(self):
         """Focus has changed from one plugin to another"""
         self.update_edit_menu()
--- a/spyder/plugins/variableexplorer.py
+++ b/spyder/plugins/variableexplorer.py
@@ -61,6 +61,7 @@ class VariableExplorer(QWidget, SpyderPluginMixin):
     CONF_SECTION = 'variable_explorer'
     CONFIGWIDGET_CLASS = VariableExplorerConfigPage
     sig_option_changed = Signal(str, object)
+    sig_free_memory = Signal()
 
     def __init__(self, parent):
         QWidget.__init__(self, parent)
@@ -114,6 +115,11 @@ class VariableExplorer(QWidget, SpyderPluginMixin):
             new_value = new_value[1:]
         self.sig_option_changed.emit(option_name, new_value)
 
+    @Slot()
+    def free_memory(self):
+        """Free memory signal."""
+        self.sig_free_memory.emit()
+
     # ----- Stack accesors ----------------------------------------------------
     def set_current_widget(self, nsb):
         self.stack.setCurrentWidget(nsb)
@@ -144,6 +150,7 @@ class VariableExplorer(QWidget, SpyderPluginMixin):
             nsb.set_shellwidget(shellwidget)
             nsb.setup(**self.get_settings())
             nsb.sig_option_changed.connect(self.change_option)
+            nsb.sig_free_memory.connect(self.free_memory)
             self.add_widget(nsb)
             self.shellwidgets[shellwidget_id] = nsb
             self.set_shellwidget_from_id(shellwidget_id)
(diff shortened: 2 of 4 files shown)


spyder/app/mainwindow.py
```diff
@@ -1954,6 +1955,11 @@ def show_toolbars(self):
         self._update_show_toolbars_action()
 
     # --- Other
+    @Slot()
```
ccordoba12: There's no need for a `Slot` here

spyder/plugins/variableexplorer.py
```diff
@@ -61,6 +61,7 @@ class VariableExplorer(QWidget, SpyderPluginMixin):
     CONF_SECTION = 'variable_explorer'
     CONFIGWIDGET_CLASS = VariableExplorerConfigPage
     sig_option_changed = Signal(str, object)
+    sig_free_memory = Signal()
```
ccordoba12: Let's remove this signal because it's not needed.

spyder/plugins/variableexplorer.py
```diff
@@ -114,6 +115,11 @@ def change_option(self, option_name, new_value):
             new_value = new_value[1:]
         self.sig_option_changed.emit(option_name, new_value)
 
+    @Slot()
+    def free_memory(self):
+        """Free memory signal."""
+        self.sig_free_memory.emit()
```
ccordoba12: Since all plugins have a reference to the wain window through `self.main`, there's no need for this. We can simply do

```python
self.main.free_memory()
```

spyder/plugins/variableexplorer.py
```diff
@@ -203,6 +210,7 @@ def get_plugin_actions(self):
     def register_plugin(self):
         """Register plugin in Spyder's main window"""
         self.main.add_dockwidget(self)
+        self.sig_free_memory.connect(self.main.free_memory)
```
ccordoba12: Let's remove this.

ccordoba12:
@dalthviz, it looks really good and it was way simpler than I thought. Thanks!

I left some minor comments, then it should be ready.

spyder/widgets/variableexplorer/collectionseditor.py
```diff
@@ -579,6 +580,7 @@ def editor_rejected(self, editor_id):
 
     def free_memory(self):
         """Free memory after closing an editor."""
+        self.sig_free_memory.emit()
         gc.collect()
```
ccordoba12: Let's remove this call to `gc.collect` because I think it's not needed anymore.

spyder/app/mainwindow.py
```diff
@@ -1954,6 +1955,10 @@ def show_toolbars(self):
         self._update_show_toolbars_action()
 
     # --- Other
+    def free_memory(self):
+        """Free memory after event."""
+        gc.collect()
```
ccordoba12: Let's also add a timer that calls this method and is executed every minute and every 5 minutes after this method is called.

spyder/plugins/variableexplorer.py
```diff
@@ -60,6 +60,8 @@ class VariableExplorer(QWidget, SpyderPluginMixin):
     """
     CONF_SECTION = 'variable_explorer'
     CONFIGWIDGET_CLASS = VariableExplorerConfigPage
+    INITIAL_FREE_MEMORY_TIME_TRIGGER = 60000
```
ccordoba12: Let's change this to `60 * 1000 # ms` and the same below.

spyder/plugins/variableexplorer.py
```diff
@@ -60,6 +60,8 @@ class VariableExplorer(QWidget, SpyderPluginMixin):
     """
     CONF_SECTION = 'variable_explorer'
     CONFIGWIDGET_CLASS = VariableExplorerConfigPage
+    INITIAL_FREE_MEMORY_TIME_TRIGGER = 60000
+    SECONDARY_FREE_MEMORY_TIME_TRIGGER = 300000
```
ccordoba12: Let's make this 3 minutes instead of 5

PR #7248, opened 2018-06-03
PR: Save credentials in the Github login dialog

ccordoba12:
Screenshots:

seleccion_002

seleccion_003

*Note*: This will add a new dependency to Spyder, called `keyring`.


ccordoba12:
@dalthviz, please add `keyring` to CircleCI too to fix the failure there.

ccordoba12:
And you need to also add it to Appveyor and Travis. Please add it with `pip` and not with `conda` because there are no conda packages for it yet.

dalthviz:
@ccordoba12 while implementing the support of the remember me function I installed `keyring` from conda. It seems that the package was included [recently](https://anaconda.org/anaconda/keyring)

ccordoba12:
Yep, just noticed that. That's really great news!

spyder/widgets/github/backend.py
```diff
@@ -14,9 +14,12 @@
 import logging
 import webbrowser
 
+import keyring
```
ccordoba12: Please enclose this in a try/except and make `keyring = None` if it fails.

spyder/widgets/github/backend.py
```diff
@@ -159,9 +164,44 @@ def send_report(self, title, body, application_log=None):
...
+        return username, remember
+
+    def _store_credentials(self, username, password, remember=False):
+        """Store credentials for future use."""
+        if username and password:
+            CONF.set('main', 'report_error/username', username)
+            try:
+                keyring.set_password('github', username, password)
+            except RuntimeError:  # pragma: no cover
+                _logger().warn(_('failed to save password in keyring, you '
```
ccordoba12: Please change this to use a `QMessageBox.warning`, with this message:

> It was not possible to securely save your password. You will be prompted for
your Github credentials next time you want to report an issue.

spyder/widgets/github/backend.py
```diff
@@ -159,9 +164,44 @@ def send_report(self, title, body, application_log=None):
...
+    def get_user_credentials(self):
         """Get user credentials with the login dialog."""
-        credentials = DlgGitHubLogin.login(self.parent_widget, None)
+        password = None
+        username, remember = self._get_credentials_from_settings()
+        if username and remember:
+            # get password from keyring
+            try:
+                password = keyring.get_password('github', username)
+            except RuntimeError:
```
ccordoba12: Change this `except Exception`

spyder/widgets/github/backend.py
```diff
@@ -159,9 +164,44 @@ def send_report(self, title, body, application_log=None):
...
+                # no safe keyring backend
+                _logger().warn('failed to retrieve password from keyring...')
```
ccordoba12: Please change this to use a QMessagebox, with this message

> It was not possible to retrieve your password. Please introduce it again.

spyder/widgets/github/gh_login.py
```diff
@@ -53,31 +55,58 @@ def __init__(self, parent, username):
...
         lbl_password = QLabel(_("Password: "))
-        basic_form_layout.setWidget(1, QFormLayout.LabelRole, lbl_password)
+        basic_form_layout.setWidget(2, QFormLayout.LabelRole, lbl_password)
         self.le_password = QLineEdit()
         self.le_password.setEchoMode(QLineEdit.Password)
         self.le_password.textChanged.connect(self.update_btn_state)
-        basic_form_layout.setWidget(1, QFormLayout.FieldRole, self.le_password)
+        basic_form_layout.setWidget(2, QFormLayout.FieldRole, self.le_password)
+
+        self.cb_remember = QCheckBox(_("Remember me"))
```
ccordoba12: Please disable this checkbox when Spyder is running with Python 2 and Linux.

spyder/widgets/github/backend.py
ccordoba12: On a second thought, we don't need to make `keyring = None`. We need to `pass` in the `except` part.

spyder/widgets/github/gh_login.py
```diff
@@ -53,31 +55,63 @@ def __init__(self, parent, username):
...
-        basic_form_layout.setWidget(1, QFormLayout.LabelRole, lbl_password)
+        basic_form_layout.setWidget(2, QFormLayout.LabelRole, lbl_password)
         self.le_password = QLineEdit()
         self.le_password.setEchoMode(QLineEdit.Password)
         self.le_password.textChanged.connect(self.update_btn_state)
-        basic_form_layout.setWidget(1, QFormLayout.FieldRole, self.le_password)
+        basic_form_layout.setWidget(2, QFormLayout.FieldRole, self.le_password)
+
+        self.cb_remember = None
+        # Check if we are in Python 2 and Linux where keyring needs a backend
```
ccordoba12: Please change this comment to

    # Check if we are not in Python 2 and Linux because there's no keyring backend there

ccordoba12:
@dalthviz, this is looking really great! Please solve the conflict in `requirements.txt`, address the last comment I added above and don't forget to also add `keyring` to our Readme (in the *Runtime dependencies* section).

Then, this one should be ready.

ccordoba12:
Please also add `spyder-kernels` to our Readme (I forgot to do it in #7306), like this:

> * **spyder-kernels** 0.2.3+: Jupyter kernels for the Spyder console.
* **keyring**: Save Github credentials to report errors securely.

ccordoba12:
@dalthviz, we also need to add a *Remember me* checkbox to the *Access Token* tab so we can save the token there.

ccordoba12:
Also, please notice that I just added two commits to this PR.

ccordoba12:
This is looking so nice!! Thanks @dalthviz for all your work on this one!

@CAM-Gerlach, what do you think? I think we fulfilled all your specs ;-)

ccordoba12:
@CAM-Gerlach, I updated the PR description with the latest screenshots.

CAM-Gerlach:
@ccordoba12 Looking great! Sorry for slacking off over the past week, and thanks @ccordoba12 and @dalthviz for all your hard work on this! I just have one minor review comment.

spyder/widgets/github/gh_login.py
```diff
@@ -47,51 +50,102 @@ def __init__(self, parent, username):
...
 
         # Tabs
-        tabs = QTabWidget()
+        self.tabs = QTabWidget()
 
         # Basic form layout
         basic_form_layout = QFormLayout()
         basic_form_layout.setContentsMargins(-1, 0, -1, -1)
 
+        basic_lbl_msg = QLabel(_("For regular users, i.e. users <b>without</b>"
```
CAM-Gerlach: One minor comment—I'd suggest changing "For regular users, i.e. users without two-factor..." to just read "For users without two-factor...", since ideally all users *should* be using two-factor as a security best practice, and therefore nominally a "regular" user should be one using two-factor (or at least we shouldn't normalize *not* using such). It also reads shorter and cleaner.

ccordoba12: I thought about that but people probably doesn't know what 2FA means. So I wanted to emphasize this is for regular users.

ccordoba12:
This is finally ready!!

spyder/widgets/github/gh_login.py
CAM-Gerlach: What's done is done, but for the record, this is a *very* bad practice from a security education standpoint. 

Considering that most major online services (email, social network, education, cloud, etc) offer it, and Spyder's audience is likely much more technologically literate than the average user, I would dearly hope so otherwise our security education efforts thus far have been a catastrophic failure. Furthermore, it is just a *really* bad practice to be explicitly declaring *not* using 2FA to be the norm for a critical service like Github in 2018; after all its been out for *five years*.

PR #7287, opened 2018-06-13
PR: Improve detection of users' home directory

ccordoba12:
The problem was we were not checking if what's saved in env vars `HOME`, `USERPROFILE` and `TMP` is really a file system path. 

Please see https://github.com/ContinuumIO/anaconda-issues/issues/9593 for the context.

### DIFF
--- a/spyder/config/base.py
+++ b/spyder/config/base.py
@@ -117,10 +117,11 @@ def get_home_dir():
         # decoded with the codec that the OS is using to represent environment
         # variables.
         path = encoding.to_unicode_from_fs(os.environ.get(env_var, ''))
-    if path:
+    if path and osp.isdir(path):
         return path
     else:
-        raise RuntimeError('Please define environment variable $HOME')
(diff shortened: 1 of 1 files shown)


spyder/config/base.py
```diff
@@ -117,10 +117,11 @@ def get_home_dir():
         # decoded with the codec that the OS is using to represent environment
         # variables.
         path = encoding.to_unicode_from_fs(os.environ.get(env_var, ''))
-    if path:
+    if path and osp.isdir(path):
         return path
     else:
-        raise RuntimeError('Please define environment variable $HOME')
+        raise RuntimeError('Please define the environment variable HOME for '
+                           'Spyder be able to work in your system.')
```
ccordoba12: @CAM-Gerlach, is this message well written in English? Or does it have syntactic or semantic errors? I have my doubts :-)

ccordoba12: Thanks, great suggestion! I'll add it right away.

PR #7546, opened 2018-07-24
PR: Do not automatically open Project Explorer when opening project if user closed it

### LINKED ISSUE #6292: Do not reopen Project explorer on startup
## Description

1. Close _Project explorer_ pane
2. Close Spyder
3. Open Spyder

**What is the expected output?** 
No _Project explorer_

**What do you see instead?**
_Project explorer_ is back, however the corresponding checkbox in _Projects_ drop-down menu is unchecked

I don't know if it's a bug or a feature. Anyway, Spyder should not decide for me what to see. Exists since 3.2.4 at least.

* Spyder Version: 3.2.6 
* Python Version: 3.6.4
* Qt Versions: 5.10.0, PyQt5 5.9.2 on Linux

## Dependencies
```
sphinx >=0.6.6   :  1.6.6 (OK)
rope >=0.9.4     :  0.10.6 (OK)
psutil >=0.3     :  5.4.3 (OK)
pylint >=0.25    :  1.8.1 (OK)

```

jitseniesen:
### Pull Request Checklist

* [ ] Included a screenshot, if this PR makes any visible changes to the UI **N/A**

## Description of Changes

Create a new config option for the projects plugin called `visible_if_project_open`, defaulting to `True` to mimic current behaviour. When closing a project or quitting Spyder with a project open, set or clear this according to whether the Project Explorer is open. When creating a new project or opening an existing project, open the Project Explorer only if the config option is set.

The effect is that if a user opens a project, closes the Project Explorer and then restarts Spyder, the Project Explorer is not opened again, as opposed to the current situation where it is opened.

### Issue(s) Resolved

Fixes #6292

### DIFF
--- a/spyder/plugins/tests/test_projects.py
+++ b/spyder/plugins/tests/test_projects.py
@@ -69,7 +69,9 @@ def test_close_project_sets_visible_config(projects_with_dockwindow, tmpdir,
     projects.set_option('visible_if_project_open', not value)
 
     projects.open_project(path=to_text_string(tmpdir))
-    if value is False:
+    if value:
+        projects.show_explorer()
+    else:
         projects.dockwidget.close()
     projects.close_project()
     assert projects.get_option('visible_if_project_open') == value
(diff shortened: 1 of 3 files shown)


spyder/plugins/tests/test_projects.py
```diff
@@ -14,6 +14,7 @@
 import pytest
 
 # Local imports
+import spyder.plugins.base
```
ccordoba12: This is only valid in `master`. For `3.x` you need to write:

    import spyder.plugins

jitseniesen: Yes, I tested against master, then rebased against 3.x, but forgot to run the tests again. Stupid.

ccordoba12:
This looks very good, but I have one remark: the project explorer is a good indicator that Spyder is in *project mode*.

Now, after this, if users create a new project but closed the project explorer while the previous one was active, it won't show up again. So the only indication left would be the addition of the projects' name to the window title. Is that ok for you @jitseniesen?

jitseniesen:
For me it's okay. I don't need an indication of project mode. I have half a dozen of projects I work on and I open them immediately when opening Spyder, or I work on single files in which case I don't use a project. The main difference of project mode for me is that the project directory is in the Python path, so my scripts have an ImportError if I run them outside project mode.

To be honest, I don't know what other users think of it, but (barring usability testing) the only way to find out is to try and the current behaviour is annoying several people. If we do need an indicator, surely we can do this with less screen area real estate.

ccordoba12:
Ok, in the spirit of "open source is about scratching your own itch", I'm going to merge this one.

Let's see what other users think about it.

PR #7566, opened 2018-07-27
PR: Improve selection of custom interpreter

### LINKED ISSUE #7529: Improve selection of custom Python interpreter in Preferences
There are three things I'd like to see improved over our current process in `Preferences > Python Interpreter > Use the following interpreter`:

1. We should save the list of all interpreters users have introduced there. This would make their life easier when they want to select previous interpreters for different conda/venvs.
2. Having that list, we could show it in a combox of options, along with the current text field.
3. The validation to see if a file is a real Python interpreter needs to be done when users press *Apply* or *Ok*, not when they're writing a path by hand. I've been bitten by this usability error a lot of times (for example when changing a conda env).

@dalthviz, please work on this one.

dalthviz:
### Pull Request Checklist

## Description of Changes

Improve custom interpreter selection.

imagen

### Issue(s) Resolved

Fixes #7529

### DIFF
--- a/spyder/plugins/maininterpreter.py
+++ b/spyder/plugins/maininterpreter.py
@@ -54,12 +54,14 @@ class MainInterpreterConfigPage(GeneralConfigPage):
             # the Python executable has already been set with pythonw.exe:
             self.set_option('executable',
                             executable.replace("pythonw.exe", "python.exe"))
-        self.set_custom_interpreters_list(executable, executable)
-        self.validate_custom_interpreters_list()
+        if not self.get_option('default'):
+            if not self.get_option('custom_executable'):
+                self.set_option('custom_executable', ' ')
+            self.set_custom_interpreters_list(executable, executable)
+            self.validate_custom_interpreters_list()
 
     def initialize(self):
         GeneralConfigPage.initialize(self)
-        self.cus_exec_radio.toggled.connect(self.python_executable_switched)
 
     def setup_page(self):
         newcb = self.create_checkbox
@@ -69,7 +71,7 @@ class MainInterpreterConfigPage(GeneralConfigPage):
         pyexec_bg = QButtonGroup(pyexec_group)
         pyexec_label = QLabel(_("Select the Python interpreter for all Spyder "
                                 "consoles"))
-        def_exec_radio = self.create_radiobutton(
+        self.def_exec_radio = self.create_radiobutton(
                                 _("Default (i.e. the same as Spyder's)"),
                                 'default', button_group=pyexec_bg)
         self.cus_exec_radio = self.create_radiobutton(
@@ -82,17 +84,18 @@ class MainInterpreterConfigPage(GeneralConfigPage):
 
         pyexec_layout = QVBoxLayout()
         pyexec_layout.addWidget(pyexec_label)
-        pyexec_layout.addWidget(def_exec_radio)
+        pyexec_layout.addWidget(self.def_exec_radio)
         pyexec_layout.addWidget(self.cus_exec_radio)
         self.validate_custom_interpreters_list()
         self.cus_exec_combo = self.create_file_combobox(
                                             _('Recent custom interpreters'),
                                             self.get_option('custom_list'),
-                                            'executable',
+                                            'custom_executable',
                                             filters=filters,
-                                            default_line_edit=True
+                                            default_line_edit=True,
+                                            adjust_to_contents=True
                                             )
-        def_exec_radio.toggled.connect(self.cus_exec_combo.setDisabled)
+        self.def_exec_radio.toggled.connect(self.cus_exec_combo.setDisabled)
         self.cus_exec_radio.toggled.connect(self.cus_exec_combo.setEnabled)
         pyexec_layout.addWidget(self.cus_exec_combo)
         pyexec_group.setLayout(pyexec_layout)
@@ -159,14 +162,9 @@ class MainInterpreterConfigPage(GeneralConfigPage):
                     _("You selected an invalid Python interpreter for the "
                       "console so the previous interpreter will stay. Please "
                       "make sure to select a valid one."), QMessageBox.Ok)
-            self.pyexec_edit.setText(def_pyexec)
+            self.def_exec_radio.setChecked(True)
             return False
         return True
-
-    def python_executable_switched(self, custom):
-        """Python executable default/custom radio button has been toggled"""
-        if custom:
-            self.python_executable_changed(self.pyexec_edit.text())
 
     def warn_python_compatibility(self, pyexec):
         if not osp.isfile(pyexec):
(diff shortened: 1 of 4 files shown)


ccordoba12:
Very nice @dalthviz! However, I'd like to have something similar to the working directory toolbar: a combobox to be able to enter by hand or with a file dialog a Python interpreter. The one in the working directory is called `PathComboBox`, so you'd need to create another class called `FileComboBox` for files.

Besides, you'd need to create another method in `SpyderConfigPage` that creates and deals with the option selected by users in a `FileComboBox`.

dalthviz:
A new oreview:

imagen

ccordoba12:
Much, much better! Thanks @dalthviz!

One last thing: Could you hide and/or remove the validation widget next to the file name (i.e. the one with the blue check mark)? Then I'd say this is ready.

dalthviz:
New preview:

imagen

spyder/plugins/maininterpreter.py
```diff
@@ -222,5 +230,20 @@ def set_umr_namelist(self):
                 fixed_namelist = []
             self.set_option('umr/namelist', fixed_namelist)
 
+    def set_custom_interpreters_list(self, executable):
+        """Update the list of interpreters used and the current one."""
+        custom_list = self.get_option('custom_list')
```
ccordoba12: We need to add a method that takes this list and verifies the entries that still exist (with a simple `osp.isfile`). The ones that don't need to be removed and a new list of interpreters needs to be saved with `set_option`.

spyder/plugins/maininterpreter.py
```diff
@@ -222,5 +230,20 @@ def set_umr_namelist(self):
                 fixed_namelist = []
             self.set_option('umr/namelist', fixed_namelist)
 
+    def set_custom_interpreters_list(self, executable):
+        """Update the list of interpreters used and the current one."""
+        custom_list = self.get_option('custom_list')
+        if (executable, executable) not in custom_list:
```
ccordoba12: This looks confusing, i.e. naming two variables with the same name (`execuatble`, `executable`). Please change that.

spyder/plugins/maininterpreter.py
ccordoba12: Sorry, this probably doesn't need to be done here, but above, where you create the filecombobox.

spyder/plugins/maininterpreter.py
```diff
@@ -77,19 +78,25 @@ def setup_page(self):
...
-        self.cus_exec_radio.toggled.connect(pyexec_file.setEnabled)
+
         pyexec_layout = QVBoxLayout()
         pyexec_layout.addWidget(pyexec_label)
         pyexec_layout.addWidget(def_exec_radio)
         pyexec_layout.addWidget(self.cus_exec_radio)
-        pyexec_layout.addWidget(pyexec_file)
+        self.cus_exec_combo = self.create_file_combobox(
+                                            _('Recent custom interpreters'),
+                                            self.get_option('custom_list'),
```
ccordoba12: So here, before setting the custom list of interpreters, we need to validate that they still exist and remove the ones that don't.

dalthviz: Maybe it would be better to do the validation in the `__init__` method?

ccordoba12: Sure, that's ok too.

spyder/plugins/maininterpreter.py
```diff
@@ -53,10 +54,11 @@ def __init__(self, parent, main):
             # the Python executable has already been set with pythonw.exe:
             self.set_option('executable',
                             executable.replace("pythonw.exe", "python.exe"))
+        self.set_custom_interpreters_list(executable, executable)
```
ccordoba12: I think this should be

```python
if not self.get_option('default'):
    self.set_custom_interpreters_list(executable, executable)
```

because we don't want the default interpreter to the list of custom ones.

spyder/plugins/maininterpreter.py
```diff
@@ -152,6 +161,7 @@ def python_executable_changed(self, pyexec):
                       "make sure to select a valid one."), QMessageBox.Ok)
             self.pyexec_edit.setText(def_pyexec)
             return
+        return True
```
ccordoba12: If you make this method to `return True` at the end, then you need to change the behavior of the previous `return`'s to be `return False`, and not simply `return` (as it right now). That's to be consistent.

spyder/plugins/maininterpreter.py
```diff
@@ -222,5 +232,29 @@ def set_umr_namelist(self):
                 fixed_namelist = []
             self.set_option('umr/namelist', fixed_namelist)
 
+    def set_custom_interpreters_list(self, display_value, value):
+        """Update the list of interpreters used and the current one."""
+        custom_list = self.get_option('custom_list')
+        if (display_value, value) not in custom_list:
+            custom_list.append((display_value, value))
+            self.set_option('custom_list', custom_list)
+        self.set_option('executable', value)
```
ccordoba12: I think this line shouldn't be here but applied after this method is called.

ccordoba12: This is true specially due to the comment I left in the `__init__` method.

spyder/plugins/maininterpreter.py
dalthviz: After trying this change I notice that we need at least one entry in the combobox. If not this triggers and error when loading the config (since there is no option to give to the current index in the combobox). The line where [this happens](https://github.com/spyder-ide/spyder/blob/aed5a5a190e92e8e653d92218582715039aa6f23/spyder/plugins/configdialog.py#L345)

ccordoba12: Mmmh, I don't like this because the default interpreter is going to be part of the list of custom ones too. Besides, users would find strange that there's one entry in the custom interpreters list if they haven't selected none.

Couldn't you find a way to make `configdialog` work with no interpreter?

dalthviz: I could add a validation in `configdialog` on the line that I linked above to not raise and error, and also change a statement in `python_executable_changed` since currently if an invalid interpreter is found it will put the [default one as an option](https://github.com/spyder-ide/spyder/blob/7d786f889b42695636d5e81ad5642f4e19e8b2ad/spyder/plugins/maininterpreter.py#L153), and fallback to select the default interpreter option when the settings are applied with and empty custom interpreter probably.

Should I do this changes?

CAM-Gerlach:
@ccordoba12 @dalthviz Just tested this on the latest ``3.x`` on Windows 8.1, and there appears to be a UI bug with this: when selecting a new interpreter from the file browser (i.e. rather than choosing it from the dropdown list), and then hitting "Apply" and "OK", just "OK", "Apply" then closing the dialog, etc., the new setting sticks. However, if I re-open preferences, the interperter path displayed in the combobox is that of the interpreter I previously had selected, rather than the actual one. However, so long as one does not actually interact with this field, making other changes in the preferences and then saving makes the interpreter for new consoles stay the one the user selected, rather than the one displayed, and selecting interperters from the list *does* make them display correctly in the box once preferences is re-opened.

Also, conversely, when selecting one or multiple new interpreters from the file browser dialog that aren't currently in the list, their path(s) is (are) not added to the list, even if the Apply button is clicked to set them, until the preferences dialog is closed and re-opened.

ccordoba12:
@CAM-Gerlach, please open a new issue about it.

CAM-Gerlach:
@ccordoba12 In work.

PR #7597, opened 2018-08-01
PR: Separate startup run_lines with semicolon instead of comma

bcolsen:
### Pull Request Checklist

* [ ] Wrote at least one-line docstrings for any new functions

## Description of Changes

Currently IPython startup lines like these won't work in Sypder since the line is parsed with comma separated:

```python
from modual import func1, func2
```
or 
```python
import os, sys
```

These common python import structures don't work in startup lines. I actually thought it was broken until I read the note in the ipython config and thought about the implications of using comma separation.

## Solution

If we use semicolon to separate the lines like you can do in python then any valid python should be possible.

Then you could use:
```python
import os, sys; from modal import func1, func2
```
## Problem
This needs the new config parser form python 3 because semicolons are inline comments in config parser in python 2

### Issue(s) Resolved

Associated With: 
https://github.com/spyder-ide/spyder-kernels/pull/15

### DIFF
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -460,9 +460,9 @@ class IPythonConsoleConfigPage(PluginConfigPage):
         run_lines_group = QGroupBox(_("Run code"))
         run_lines_label = QLabel(_("You can run several lines of code when "
                                    "a console is started. Please introduce "
-                                   "each one separated by commas, for "
-                                   "example:<br>"
-                                   "<i>import os, import sys</i>"))
+                                   "each one separated by semicolons and a "
+                                   "space, for example:<br>"
+                                   "<i>import os; import sys</i>"))
         run_lines_label.setWordWrap(True)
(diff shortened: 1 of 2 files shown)


ccordoba12:
I'll merge after releasing a new version of spyder-kernels.

spyder/plugins/ipythonconsole.py
```diff
@@ -460,9 +460,9 @@ def setup_page(self):
         run_lines_group = QGroupBox(_("Run code"))
         run_lines_label = QLabel(_("You can run several lines of code when "
                                    "a console is started. Please introduce "
-                                   "each one separated by commas, for "
-                                   "example:<br>"
-                                   "<i>import os, import sys</i>"))
+                                   "each one separated by semicolons and a "
```
bcolsen: The semicolon and a space are only needed for python 2.7. Python 3 works without the space

ccordoba12: Ok, I'll add that to our release noes for beta1 when I merge this one.

PR #7605, opened 2018-08-02
PR: Send reported issues to Spyder's main repo in DEV mode

ccordoba12:
Issues were sent to my fork before. We introduced that for testing but it affects people running Spyder with `bootstrap.py`.


ccordoba12:
@emanspeaks, this is the fix.

Issue #7447, opened 2018-07-13
Main window is black when starting Spyder over ssh/X11


bscipio:
### Issue Report Checklist

* [x] Searched the [issues page](https://github.com/spyder-ide/spyder/issues?q=is%3Aissue) for similar reports
* [x] Read the relevant sections of the [Spyder Troubleshooting Guide](https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ) and followed its advice
* [x] Reproduced the issue after updating with ``conda update spyder`` (or ``pip``, if not using Anaconda)
* [n/a] Could not reproduce inside ``jupyter qtconsole`` (if console-related)
* [X] Tried basic troubleshooting (if a bug/error)
    * [X] Restarted Spyder
    * [X] Reset preferences with ``spyder --reset``
    * [X] Reinstalled the latest version of [Anaconda](https://www.anaconda.com/download/)
    * [X] Tried the other applicable steps from the Troubleshooting Guide
* [X] Completed the **Problem Description**, **Steps to Reproduce** and **Version** sections below

## Problem Description

I'm attempting to use Spyder on a remote Linux server by SSH and X11 forwarding (Windows 10, PuTTY, VcXsrv).

Launch spyder (`% spyder`), splash screen appears, and spyder windows is black and unusable. 
Only windows title bar exists to close it.

and PuTTY terminal displays following:
```
% spyder
QQuickWidget: Failed to make context current
QQuickWidget::resizeEvent() no OpenGL context
QQuickWidget: Failed to make context current
QQuickWidget: Failed to make context current
QQuickWidget::resizeEvent() no OpenGL context
QQuickWidget: Failed to make context current
QQuickWidget: Attempted to render scene with no context
QQuickWidget: Attempted to render scene with no context
composeAndFlush: makeCurrent() failed
```

Believe issue is related to Qt/PyQt (5.9.4/5.9.2) versions.
Since same 3.3.0 version of Spyder,  but older Qt/PyQt (5.6.2/5.6) will work.
More details below.

May be somewhat related to #7152 and #6968

### What steps reproduce the problem?

1. conda create -n a52spy33 anaconda=5.2
2. conda activate a52spy33
3. conda install spyder=3.3.0
4. From Windows PC using PuTTY ssh to Linux host (VcXsrv for X11), and launch spyder (`% spyder`)
5. Splash screen appears, and spyder windows is black (only windows title bar exists)

[...]

PR #7700, opened 2018-08-15
PR: Add option to select OpenGL implementation used by Qt

ccordoba12:
To try to avoid the many issues people are finding with WebEngine, we will allow them to select the OpenGL implementation used by Qt, like this:

seleccion_008

They could also select it with a new `--opengl` command line option, that can have two options: `software` and `desktop`.

----

Note: We are following RStudio's lead in this camp. See:

https://github.com/spyder-ide/spyder/issues/7447#issuecomment-412700201

and

https://github.com/rstudio/rstudio/issues/3185#issuecomment-412575262

for the details.


bklyons:
I think the new command line option should be added [here](https://docs.spyder-ide.org/options.html)?

PR #7734, opened 2018-08-22
PR: Replace debug_print for the logging module and deprecate its use

### LINKED ISSUE #5326: Improve debug logging

ccordoba12:
Fixes #5326 

This is the continuation of PR #5502, started by @rlaverde.

TODO:

- [x] Change print and traceback calls in editor widget `if DEBUG_PRINT:    print( file=STDOUT)`
- [x] Add logging configuration
- [x] Remove DEBUG constant
- [x] Improve a bit some messages
- [x] Fix `bootstrap.py --debug`


spyder/config/base.py
```diff
@@ -66,6 +67,7 @@ def _get_debug_env():
 
 def debug_print(*message):
     """Output debug messages to stdout"""
+    warnings.warn("Debug_print is deprecated use logging module instead.")
```
CAM-Gerlach: Ungrammatical; needs a semicolon like so:

```
"Debug_print is deprecated; use the logging module instead."
```

spyder/tests/test_dont_use.py
```diff
@@ -21,11 +21,11 @@
     (r"^[\s\#]*\bprint\(((?!file=).)*\)", ['.*test.*', 'example.py',
                                            'example_latin1.py', 'binaryornot'],
      ("Don't use print() functions, ",
-      "for debuging you could use debug_print instead")),
+      "for debuging you could use logging module instead")),
```
CAM-Gerlach: Correct spelling is "debugging"; comma splice; grammar issues. Thus, recommend

```
"Don't use the print() function; "
"for debugging, use the logging module instead."
```

spyder/tests/test_dont_use.py
```diff
@@ -21,11 +21,11 @@
...
 
     (r"^[\s\#]*\bprint\s+(?!>>)((?!#).)*", ['.*test.*', 'example_latin1.py'],
      ("Don't use print __builtin__, ",
-      "for debuging you could use debug_print instead")),
+      "for debuging you could use logging module instead")),
```
CAM-Gerlach: As above:

```
"Don't use print statements; "
"for debugging, use the logging module instead."
```

PR #7738, opened 2018-08-22
Make Group cells setting persistent when restarting spyder

### LINKED ISSUE #7736: Add the outline explorer option "Group code cells" to config and make it False by default
## Problem Description

I think the `Group code cells` option added in PR#7107 in the outline explorer is nice, but...

1. The option needs to be added to our configs so that I won't have to turn it off every-time I open Spyder
2. I think the default for this option should be `False`

@ccordoba12 what do you think? I can do it if you agree with this.

image

bcolsen:
### Pull Request Checklist

* [ ] Wrote at least one-line docstrings for any new functions

## Description of Changes

This need to make the "Group code cells" setting for the outline explorer persistent 

### Issue(s) Resolved

Fixes #7736

### DIFF
--- a/spyder/plugins/editor/widgets/editor.py
+++ b/spyder/plugins/editor/widgets/editor.py
@@ -2803,7 +2803,7 @@ class EditorSplitter(QSplitter):
 
 class EditorWidget(QSplitter):
     def __init__(self, parent, plugin, menu_actions, show_fullpath,
-                 show_all_files, show_comments):
+                 show_all_files, group_cells, show_comments):
         QSplitter.__init__(self, parent)
         self.setAttribute(Qt.WA_DeleteOnClose)
 
@@ -2823,6 +2823,7 @@ class EditorWidget(QSplitter):
         self.outlineexplorer = OutlineExplorerWidget(self,
                                             show_fullpath=show_fullpath,
                                             show_all_files=show_all_files,
+                                            group_cells=group_cells,
                                             show_comments=show_comments)
         self.outlineexplorer.edit_goto.connect(
                      lambda filenames, goto, word:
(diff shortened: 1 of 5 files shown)


spyder/plugins/editor/widgets/editor.py
```diff
@@ -2820,11 +2820,12 @@ def __init__(self, parent, plugin, menu_actions, show_fullpath,
         self.find_widget = FindReplace(self, enable_replace=True)
         self.plugin.register_widget_shortcuts(self.find_widget)
         self.find_widget.hide()
-        self.outlineexplorer = OutlineExplorerWidget(self,
-                                            show_fullpath=show_fullpath,
-                                            show_all_files=show_all_files,
-                                            group_cells=group_cells,
-                                            show_comments=show_comments)
+        self.outlineexplorer = OutlineExplorerWidget(
+                self,
```
jnsebgosselin: I think a 4-space indentation should be used here instead of 8 (even though it is what Spyder is proposing by default...). Also, this could be written on less lines. Something like:

```python
self.outlineexplorer = OutlineExplorerWidget(
    self, show_fullpath=show_fullpath, show_all_files=show_all_files,
    group_cells=group_cells, show_comments=show_comments)
```

@ccordoba12 Am I ok with this so that I do not repeat wrong code style in my own PR.

ccordoba12: I think it's fine as it is. Besides, IMHO it's more readable.

jnsebgosselin: Ok thanks for the feedback about keeping one line per argument, but what about the indentation? 

I've been asked in another PR to use 4 spaces instead of 8 as per some pep rules. Are we following any convention in Spyder for this?

bcolsen: I think 8 spaces is nice to differentiate it as a line continuation and not a 'normal' code block, but it really doesn't matter.

jnsebgosselin: >but it really doesn't matter.

I know. I do not care much either, but I'm trying to follow the "official" codestyle that is used in the spyder codebase as much as possible. I'm asking so that I can save revision time in my future PR.

ccordoba12: > but I'm trying to follow the "official" codestyle

Wed don't really have one. In this case I think using 8 spaces is fine too.

jnsebgosselin: > Wed don't really have one. In this case I think using 8 spaces is fine too.
Alright, thank you Carlos. This is good to know.

PR #7766, opened 2018-08-24
PR: Fix Browse and Commit vcs actions in file and project explorers

### LINKED ISSUE #7733: Error when browsing git repository

I'm currently working on a Git repository, and when I clicked on the "browse repository" option from the right-click drop down menu, this error popped up.

### Traceback
```python-traceback
  File "C:\Users\manma\Anaconda3\lib\site-packages\spyder\widgets\explorer.py", line 395, in <lambda>
    self.vcs_command(fnames, 'browse')
  File "C:\Users\manma\Anaconda3\lib\site-packages\spyder\widgets\explorer.py", line 825, in vcs_command
    for path in sorted(fnames):
TypeError: 'bool' object is not iterable
```

## Versions

* Spyder version: 3.3.1 
* Python version: 3.6.6
* Qt version: 5.9.6
* PyQt5 version: 5.9.2

### Dependencies

```
sphinx >=0.6.6    :  1.7.6 (OK)
jedi >=0.9.0      :  0.12.1 (OK)
numpy >=1.7       :  1.15.0 (OK)
sympy >=0.7.3     :  1.2 (OK)
qtconsole >=4.2.0 :  4.4.0 (OK)
IPython >=4.0     :  6.5.0 (OK)
matplotlib >=2.0.0:  2.2.3 (OK)
```

ccordoba12:
Fixes #7733.

Also add some tests for `utils/vcs.py`

### DIFF
--- a/spyder/utils/vcs.py
+++ b/spyder/utils/vcs.py
@@ -8,9 +8,10 @@
 
 from __future__ import print_function
 
-import sys
+import os
 import os.path as osp
 import subprocess
+import sys
 
 # Local imports
 from spyder.utils import programs
@@ -31,7 +32,7 @@ SUPPORTED = [
     'name': 'Git',
     'rootdir': '.git',
     'actions': dict(
-        commit=( ('git', ['gui']), ),
+        commit=( ('git', ['gui' if os.name == 'nt' else 'cola']), ),
         browse=( ('gitk', []), ))
 }]
 
--- a/spyder/utils/vcs.py
+++ b/spyder/utils/vcs.py
@@ -14,6 +14,7 @@ import subprocess
 import sys
 
 # Local imports
+from spyder.config.base import running_under_pytest
 from spyder.utils import programs
 from spyder.utils.misc import abspardir
 from spyder.py3compat import PY3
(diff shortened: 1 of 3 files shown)


spyder/utils/vcs.py
```diff
@@ -31,7 +32,7 @@
     'name': 'Git',
     'rootdir': '.git',
     'actions': dict(
-        commit=( ('git', ['gui']), ),
+        commit=( ('git', ['gui' if os.name == 'nt' else 'cola']), ),
```
jitseniesen: What happens if git-cola is not installed?

ccordoba12: Nothing happens, i.e. no error is generated but also no message is shown. That was the previous behavior.

Would you like to improve it?

jitseniesen: > Would you like to improve it?

Not at the moment, perhaps at some time in the undetermined future as part of a bigger overhaul.

CAM-Gerlach: @ccordoba12 As @juanis2112 confirmed, this change appears to silently breaks the `Commit` action in the Files pane context menu on all OSes except Windows unless the third party `git cola` GUI is installed. Is there some reason Spyder can't simply use the built-in `git gui` for this on all platforms, as it did prior to this PR, instead of a third party one that likely only a small minority of users have installed? It should work on any mainstream platform that Git itself is supported on (at least for Git distributions that include it, which is most though not all of them), and I can confirm it works fine on Windows, and appears to work on macOS as well in limited testing, at least with standard `brew`-installed Git. Is there a Spyder-specific reason it cannot be used here on those platforms?

PR #7790, opened 2018-08-30
PR: Protecting calls to unregister_editorstack_cb and pin sip in our tests

jnsebgosselin:
## Description of Changes

From time to time, a test is failing on our CI with this error:

https://travis-ci.org/spyder-ide/spyder/jobs/422626738#L932
```python
==================================== ERRORS ====================================
 ERROR at setup of test_reorder_filenames[filenames2-expected_filenames2-None-None] 
SETUP ERROR: Qt exceptions in virtual methods:
________________________________________________________________________________
Traceback (most recent call last):
  File "/home/travis/build/spyder-ide/spyder/spyder/plugins/editor/widgets/editor.py", line 2629, in <lambda>
    self.editorstack.destroyed.connect(lambda: self.editorstack_closed())
  File "/home/travis/build/spyder-ide/spyder/spyder/plugins/editor/widgets/editor.py", line 2654, in editorstack_closed
    self.unregister_editorstack_cb(self.editorstack)
  File "/home/travis/build/spyder-ide/spyder/spyder/plugins/editor/plugin.py", line 1525, in unregister_editorstack
    self.remove_last_focus_editorstack(editorstack)
  File "/home/travis/build/spyder-ide/spyder/spyder/plugins/editor/plugin.py", line 1375, in remove_last_focus_editorstack
    for editorwindow, widget in list(self.last_focus_editorstack.items()):
RuntimeError: wrapped C/C++ object of type Editor has been deleted
________________________________________________________________________________
Traceback (most recent call last):
  File "/home/travis/build/spyder-ide/spyder/spyder/plugins/editor/widgets/editor.py", line 2629, in <lambda>
    self.editorstack.destroyed.connect(lambda: self.editorstack_closed())
  File "/home/travis/build/spyder-ide/spyder/spyder/plugins/editor/widgets/editor.py", line 2654, in editorstack_closed
    self.unregister_editorstack_cb(self.editorstack)
  File "/home/travis/build/spyder-ide/spyder/spyder/plugins/editor/plugin.py", line 1525, in unregister_editorstack
    self.remove_last_focus_editorstack(editorstack)
  File "/home/travis/build/spyder-ide/spyder/spyder/plugins/editor/plugin.py", line 1375, in remove_last_focus_editorstack
    for editorwindow, widget in list(self.last_focus_editorstack.items()):
RuntimeError: wrapped C/C++ object of type Editor has been deleted
________________________________________________________________________________
Traceback (most recent call last):
[...]

PR #7789, opened 2018-08-30
PR: Add "find_replace" widget to qtbot in fixtures of "test_editor.py"

jnsebgosselin:
## Description of Changes

This is a simple proposition to make the `find_replace` widget window close automatically after each test in `test_editor.py`.

Otherwise, it looks like this after running pytest on `.\spyder\plugins\editor\widgets\tests\test_editor.py`:

image

PR #7792, opened 2018-08-30
PR: Split "test_mainwindow.test_help" and skip tests on Windows CI because they are timing out.

jnsebgosselin:
## Description of Changes

It seems that tests related to the LSP are failing on AppVeyor with this error:

image

This PR proposes to skip the problematic tests on Windows CI. Note that the tests are passing locally on my Windows 10 laptop.

Also this PR proposes to split the `test_mainwindow.test_help` in two so that `Help` is tested independently when called from the Editor or from the IPython console. Otherwise, it is not possible to know if the test is timing out because of the action made in the Editor or in the IPython console.

Unfortunately, after the split, I found out that both tests were timing out on AppVeyor, so they had to be skipped. Still, I think splitting the test in two is an improvement, so I'm proposing it here anyway.

### Issue(s) Resolved

This PR will allow the tests on Appveyor to pass for the following three PR:

PR #7790
PR #7789
PR #7768

### DIFF
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -266,6 +266,8 @@ def test_filter_numpy_warning(main_window, qtbot):
 
 @pytest.mark.slow
 @flaky(max_runs=3)
+@pytest.mark.skipif(os.name == 'nt' and os.environ.get('CI') is not None,
+                    reason="Times out on AppVeyor")
 @pytest.mark.use_introspection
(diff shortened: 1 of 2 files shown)


ccordoba12:
Tests are failing in Travis in the exact same point, so it seems you have found another error due to this.

ccordoba12:
Ok, I see you're addressing that in PR #7790.

ccordoba12:
Failures in Travis were already fixed.

PR #7837, opened 2018-09-05
PR: Fix tests after ci-helpers updated conda version

ccordoba12:
This uses `conda activate` to activate the test env. This is the new way to activate envs since conda 4.5, updated by ci-helpers.


ccordoba12:
The coveralls failure was due to someone touching the previous settings.

PR #7864, opened 2018-09-09
PR: More improvements to Find in Files

### LINKED ISSUE #4474: Find in files exclude
This makes sense as an extra default for find in files (produced by git when rebasing merging)
`\.py\.orig$`

The issue with the exclude textbox is that it is very limited as more and more things are added. I think exclude should be a two column table, where each row had one entry and the column had a description, so that users can add comments to specific patterns of exclusion. Also this makes the list/table usable when we start adding more exclusion patterns.

### LINKED ISSUE #7809: Make visible by default the widget to exclude file types in Find in Files
In view of what @jnsebgosselin said on issue #7803, we really need to make that widget visible so that people know that that functionality exists ass part of the plugin.

We also need to change the default options present in that widget.

### LINKED ISSUE #7810: Find In Files: regexp icon
The button to toggle regular expression matching when excluding files has a gear icon, which is normally interpreted as "settings". I suggest using something else to represent regular expressions.

### LINKED ISSUE #7819: Make exclude patterns in Find in files only work for file names
Right now they are working for the entire file path.

ccordoba12:
Fixes #7810 
Fixes #7809 
Fixes #4474 

----

TODO

- [x] Add icon for regexp's (there's none in iconic fonts).
- [x] Change color of invalid regexp in Search and Exclude fields (right now it's orange, it should be red).
- [x] Skip VCS dirs completely (i.e. `.git` and `.hg`)


CAM-Gerlach:
> Make exclude patterns to only work for filenames and not entire paths.

What is the rationale for this change? This is a *huge* downgrade in functionality and reduces performance by at least 1-2 orders of magnitude on VCS repositories (i.e. 90+% of what I use it on), to the point of making the Find in Files feature virtually useless for me. I routinely use it with ``|\.git|`` nearly all the time, as otherwise it will search the ``.git`` directory which makes even simple no-regex searches with few hits take one or more *minutes* rather than a few seconds on large repos, like Spyder's own—and that on a high-end SSD; the hit would be 1-2 orders of magnitude worse on any hard drive, with 1-2 orders of magnitude fewer IOPS (which is what matters for the 10,000s of small files you get with VCS systems). It is not possible to use filename exclusions, since Git objects are just random hashes with no extension.

Furthermore, I quite often use it to exclude specific directories, e.g. with vendored/external code like ``mathjax``, build output or cache directories, or other undesired components from a search, that would otherwise swamp the results I'm actually trying to look for. Again, that would make it much less useful in many common scenarios.

If this is really still desired for some reason, perhaps you could make just the "simple" comma-separated mode work only for filenames, and the regex mode work on the full path, since this would reduce the chance of a newbie not using regex finding something on the full path, while a more experienced user would likely be in regex mode to do this sort of thing anyway, and this could be noted in the docs.

ccordoba12:
> What is the rationale for this change?

That it should be easier for users to understand how to use the Exclude field.

> I routinely use it with |\.git| nearly all the time

Ok, then we should exclude dot directories by default. I'll add it to the TODO for this PR.

> Furthermore, I quite often use it to exclude specific directories, e.g. with vendored/external code like mathjax

Ok, I haven't thought about that. Then we shouldn't need to make this change. Thanks for pointing it out.

CAM-Gerlach:
> Ok, then we should exclude dot directories by default. I'll add it to the TODO for this PR.

We could just exclude ``.git`` and ``.hg`` by default, since the user might want to search other dotfiles and .dirs sometimes (I sometimes do). It is only those two that are the problem, since they contain so many tiny files.

> Ok, I haven't thought about that. Then we shouldn't need to make this change. Thanks for pointing it out.

A good compromise might be to have the non-regex mode (i.e. what new users would be using) only work on file names (with ``.git`` and ``.hg`` excluded), while regex mode works on full paths for experienced users to retain the full functionality as before.

ccordoba12:
> A good compromise might be to have the non-regex mode (i.e. what new users would be using, while I never use it) only work on file names

No, I prefer to leave the non-regex mode as powerful as possible.

CAM-Gerlach:
I have no problem with that :+1:

ccordoba12:
@CAM-Gerlach, I added a new icon for regexp's. Please review that everything is fine about that.

spyder/utils/misc.py
```diff
@@ -278,3 +279,15 @@ def getcwd_or_home():
         debug_print("WARNING: Current working directory was deleted, "
                     "falling back to home dirertory")
         return get_home_dir()
+
+
+def regexp_error_msg(pattern):
+    """
+    Returns None if the pattern is a valid regular expression or
```
CAM-Gerlach: Imperative tense ("Return None"). Also, per PEP 257, docstrings must begin with a one-line summary, separated by a blank line from the rest of the docstring.

spyder/widgets/findinfiles.py
```diff
@@ -139,8 +139,10 @@ def find_files_in_path(self, path):
                         if self.stopped:
                             return False
                     dirname = os.path.join(path, d)
-                    if self.exclude \
-                       and re.search(self.exclude, dirname + os.sep):
+                    if (self.exclude and
```
CAM-Gerlach: Break before binary operator?

spyder/widgets/findinfiles.py
```diff
@@ -918,13 +932,15 @@ class FindInFilesWidget(QWidget):
     sig_finished = Signal()
 
     def __init__(self, parent,
-                 search_text=r"# ?TODO|# ?FIXME|# ?XXX",
-                 search_text_regexp=True,
-                 exclude=r"\.pyc$|\.orig$|\.hg|\.svn", exclude_idx=None,
-                 exclude_regexp=True,
+                 search_text="",
+                 search_text_regexp=False,
+                 exclude="*.json, *.csv, *.js",
```
CAM-Gerlach: Is there a motivation for js as one of the default excludes (but not, e.g., HTML or CSS)?

spyder/config/main.py
```diff
@@ -28,7 +28,7 @@
 # Main constants
 # =============================================================================
 # Find in files exclude patterns
-EXCLUDE_PATTERNS = [r'\.pyc$|\.pyo$|\.git']
+EXCLUDE_PATTERNS = ['*.json, *.csv, *.js']
```
CAM-Gerlach: See other comment.

NOTICE.txt
```diff
@@ -1494,6 +1494,50 @@ Files covered:
 spyder/images/winpython.svg
 
 
+-------------------------------------------------------------------------------
+
+
+
+OOjs UI MediaWiki lib
+---------------------
+
```
CAM-Gerlach: Add one blank line

NOTICE.txt
```diff
@@ -1494,6 +1494,50 @@ Files covered:
...
+Roan Kattouw <roan@wikimedia.org>
+Rob Moen <rmoen@wikimedia.org>
+Timo Tijhof <krinklemail@gmail.com>
+Trevor Parscal <trevor@wikimedia.org>
+Volker E. <volker.e@wikimedia.org>
+
+Source/Download: https://phabricator.wikimedia.org/diffusion/GVED/browse/master/lib/oojs-ui/themes/mediawiki/images/icons/
+License: MIT (Expat) License | https://opensource.org/licenses/MIT
+
+Modifications made: Removed parentheses, make it bigger.
```
CAM-Gerlach: The actual line is fine, perfect in fact. 

With regard to the parenthesis removal though, I'd personally prefer it with the parens. It helps it avoid being misread as "glob" (which the simple mode at least kinda sorta looks like, with a ``*.`` as part of the default strings) rather than "regex" (which I'd be uncertain about it if I didn't actually know the truth), as well as seems more standard, judging from the icons on Wikimedia commons (where you got this) and the first hits on Google images, and looks a little cleaner and more self-contained. But its not the end of the world.

NOTICE.txt
```diff
@@ -1494,6 +1494,50 @@ Files covered:
...
+Kunal Mehta <legoktm@gmail.com>
+Moriel Schottlender <moriel@gmail.com>
+Prateek Saxena <prtksxna@gmail.com>
+Roan Kattouw <roan@wikimedia.org>
+Rob Moen <rmoen@wikimedia.org>
+Timo Tijhof <krinklemail@gmail.com>
+Trevor Parscal <trevor@wikimedia.org>
+Volker E. <volker.e@wikimedia.org>
+
+Source/Download: https://phabricator.wikimedia.org/diffusion/GVED/browse/master/lib/oojs-ui/themes/mediawiki/images/icons/
```
CAM-Gerlach: This link's not current (as of the time you downloaded the image), as the files were moved in the repo. Keep this as the source, but you can add the Wikimedia commons link, https://commons.wikimedia.org/wiki/File:OOjs_UI_icon_regular-expression.svg , as the Download link since that's where you actually downloaded it from.

NOTICE.txt
```diff
@@ -1494,6 +1494,50 @@ Files covered:
...
+
+
+
+OOjs UI MediaWiki lib
+---------------------
+
+Copyright (c) 2011-2018 OOUI Team and other contributors.
+
+
+Principal Authors:
```
CAM-Gerlach: Instead of pasting all the authors here, to keep the format clean and consistent you can just paste this block (with the ``(major contributors, alphabetically)`` so as to be verbatim to the source) one blank line below the ``Licensed under the terms of the MIT license`` line, and use the simpler form for the authors:

```
Authors: OOUI Team and other contributors
https://phabricator.wikimedia.org/diffusion/GOJU/browse/master/AUTHORS.txt
Source: https://phabricator.wikimedia.org/foobar...
Download: https://commons.wikimedia.org/foobar...
```

That follows what I did in the other cases like this, so as to not overcomplicate things :)

spyder/widgets/findinfiles.py
ccordoba12: I don't understand what you mean here.

spyder/widgets/findinfiles.py
ccordoba12: No problem. I'll add html and css.

What do you think about excluding json and csv by default? Perhaps it would be better to skip `*.dat` and `*.log` instead?

spyder/widgets/findinfiles.py
CAM-Gerlach: https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

It seems our code generally follows the modern "break before the operator", including that bit before you modified it.

NOTICE.txt
ccordoba12: I removed the parenthesis because `.*` is also used to denote regexps in Sublime and VSCode, so it's kind of universal right now.

CAM-Gerlach: Okay; if its the commonly understood convention then its fine I guess.

spyder/widgets/findinfiles.py
CAM-Gerlach: > No problem. I'll add html and css.

I guess my comment was unclear; the intent of question was more "What's the motivation to add ``.js`` as default ignores in the first place" with HTML and CSS as examples of the resulting "slippery slope". *Personally,* I typically have ``,js`` in there myself as an inclusion, as I don't usually want to search the Mathjax files, jQuery, etc. in my searches of Spyder, Javascript stuff searching the docs or the website, etc., but that's all for particular cases and more oriented toward software or web development vs. data science that Spyder is targeted at so I'm not sure it really makes sense in the general case.

CSV certainly seems sensible to have, as it is one of the most comment, if not *the* most common plain-text data formats, which would often be very large and burdensome to search, while finding results in them are rarely desired. By the same token, ``.dat`` (which to my knowledge is used for other types of ASCII as well as binary data files) would probably similarly make sense. I'm not sure how much JSON is used in data science especially for large data storage, so while it probably doesn't hurt to include it, it might not be priority. Excluding temporary/ephemeral files is generally a good idea, so under that justification ``.log`` makes sense; I typically also include ``.bak`` and (Git) ``.orig`` for that purpose, and you could include ``.tmp`` as well. So, in summary, my suggestion would be something like ``*.bak, *.tmp, *.orig, *.log, *.csv, *.dat``

ccordoba12: Good suggestion. I'll change the default pattern for that one :+1:

spyder/widgets/findinfiles.py
CAM-Gerlach: PEP 8 recommends it, but doesn't require it. Just wanted to put that out there since that seems to be the convention a lot of our code follows, including the modified portion.

Also, just FYI, PEP8Speaks doesn't seem to be really actively maintained right now; the last commit is almost five months ago and my issues and PRs have been ignored unanswered for over three. However, the maintainer may simply be on hiatus since I did see him post one issue himself a week ago, so there's hope.

NOTICE.txt
```diff
@@ -1494,6 +1494,39 @@ Files covered:
...
+OOjs UI MediaWiki lib
+---------------------
+
+
+Copyright (c) 2011-2018 OOUI Team and other contributors.
+
+
+Authors: OOUI Team and other contributors
+https://phabricator.wikimedia.org/diffusion/GOJU/browse/master/AUTHORS.txt
+Source/Download: https://commons.wikimedia.org/wiki/File:OOjs_UI_icon_regular-expression.svg
```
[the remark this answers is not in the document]
ccordoba12: And that's why I said this is very boring stuff... Hopefully we'll get rid of this when we move our icon set of Spyder itself.

spyder/utils/misc.py
```diff
@@ -281,6 +282,18 @@ def getcwd_or_home():
         return get_home_dir()
 
 
+def regexp_error_msg(pattern):
+    """
+    Return None if the pattern is a valid regular expression or
```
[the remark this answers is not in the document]
ccordoba12: Sorry, this is a very simple change, so I'm not going to do anything about it.

spyder/widgets/tests/test_findinfiles.py
```diff
@@ -113,7 +107,8 @@ def test_find_in_files_search(qtbot):
 
 
 def test_exclude_extension_regex(qtbot):
-    find_in_files = setup_findinfiles(qtbot, exclude="\.py$")
+    find_in_files = setup_findinfiles(qtbot, exclude="\.py$",
```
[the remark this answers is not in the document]
ccordoba12: Ok, I'll change it.

ccordoba12:
> It matches a bunch of random strings in the binary pdn file

`binaryornot` reports it as a text file, that's the problem. We skip all files `binaryornot` report as binary.

CAM-Gerlach:
> `binaryornot` reports it as a text file, that's the problem. We skip all files `binaryornot` report as binary.

Unrelated, but IMO we can remove those ``pdn`` files (other than the spash screen one) since they were from the very beginning of Spyder, and carefully checking through the ``image`` directory, the icons associated with them by all appearances no longer exist anywhere; particularly since they are in a closed, proprietary format that can only be opened by a single esoteric application on a single platform.

ccordoba12:
> but IMO we can remove those pdn files

If they are not used anymore, then please go ahead and remove them. Thanks!

PR #7898, opened 2018-09-14
PR: Clarify and improve Connect to kernel dialog text

### LINKED ISSUE #7868: Connect to kernel dialog needs to be clearer that either the SSH keyfile *or* the password is required
## Problem Description

Several users have stated that the connect to kernel dialog is confusing when connecting to a remote kernel, because it isn't clear that either a SSH keyfile *or* a password is required, not both—and looking at it now, I can see why.

image

### User Reports

In particular, on our Gitter, a user said:

> Hi, I am trying to connect Spyder on my PC (running within Anaconda) to a remote ipython kernel on an AWS EC2 AMI and I'm struggling. I followed the instructions here: http://docs.spyder-ide.org/ipythonconsole.html#connect-to-an-external-kernel and the kernel is running fine on the EC2 AMI and I have the JSON file on my PC but I'm not sure what to enter in the 'Connect to an existing kernel' box. I add the path to the JSON file in the 'Connection info' box. I check the 'This is a remote kernel' box. I enter my EC2 host: 'ec2-user@blah-blah:8888' and then get stuck on the SSH key and password boxes. When I connect to the EC2 AMI from MobaXterm I use an imported Putty key (.ppk) file (and don't use a password) whereas the Spyder dialogue box is asking for a .pem file. Do I need the original AWS .pem key?

And on spyder-ide/spyder-docs#78 , @jsh9 said:

> It might also be worth noting (either in Spyder's UI, or in the documentation) that the "ssh key" field is optional. I spent quite some time googling what ssh key file I should provide to no avail. Other users may run into the same confusion.

### Proposed Solutions

I'll clarify it in the docs as part of spyder-ide/spyder-dcos#78 , but it needs to be clearer in the UI too.

One possibility would be including the text "Or" between the two fields and grouping them in some way, e.g. 

Host name: ``____________________________``

SSH Key: ``_______________________________``
Or
Password: ``______________________________``

Alternatively, we could just include a line of explanatory text above the two fields saying only one of the two are required, e.g. 

Host name: ``____________________________``
Only one of the following is required:
SSH Key: ``_______________________________``
Password: ``______________________________``

CAM-Gerlach:
### Pull Request Checklist

### Affirmation

## Description of Changes

Revises the Connect to kernel dialog text to be more clear, descriptive and stylistically correct, particularly with regard to specifying that only a SSH keyfile *or* user password is required, not both.

image

image

### Issue(s) Resolved

Fixes #7868

### DIFF
--- a/spyder/plugins/ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole.py
@@ -167,11 +167,11 @@ def openssh_tunnel(self, lport, rport, server, remoteip='127.0.0.1',
 
 class KernelConnectionDialog(QDialog):
     """Dialog to connect to existing kernels (either local or remote)."""
-    
+
     def __init__(self, parent=None):
         super(KernelConnectionDialog, self).__init__(parent)
         self.setWindowTitle(_('Connect to an existing kernel'))
-        
+
         main_label = QLabel(_(
             "Please select the JSON connection file (<i>e.g.</i> "
             "<tt>kernel-3764.json</tt>) or enter the 4-digit ID (<i>e.g.</i> "
@@ -182,8 +182,8 @@ class KernelConnectionDialog(QDialog):
             "unless the keyfile is protected with a passphrase.\n"))
         main_label.setWordWrap(True)
         main_label.setAlignment(Qt.AlignJustify)
-        
-        # connection file
+
+        # Connection file
         cf_label = QLabel(_('Kernel ID/Connection file:'))
         self.cf = QLineEdit()
         self.cf.setPlaceholderText(_('ID number or path to connection file'))
@@ -195,14 +195,14 @@ class KernelConnectionDialog(QDialog):
         cf_layout.addWidget(cf_label)
         cf_layout.addWidget(self.cf)
         cf_layout.addWidget(cf_open_btn)
-        
-        # remote kernel checkbox 
+
+        # Remote kernel checkbox
         self.rm_cb = QCheckBox(_('This is a remote kernel (via SSH)'))
-        
-        # ssh connection 
+
+        # SSH connection
         self.hn = QLineEdit()
         self.hn.setPlaceholderText(_('username@hostname:port'))
-        
+
         self.kf = QLineEdit()
         self.kf.setPlaceholderText(_('Path to SSH keyfile'))
         kf_open_btn = QPushButton(_('Browse'))
@@ -211,7 +211,7 @@ class KernelConnectionDialog(QDialog):
         kf_layout = QHBoxLayout()
         kf_layout.addWidget(self.kf)
         kf_layout.addWidget(kf_open_btn)
-        
+
         self.pw = QLineEdit()
         self.pw.setPlaceholderText(_('Remote user password or '
                                      'SSH keyfile passphrase'))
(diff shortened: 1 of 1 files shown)


bcolsen:
As a quick thought `Path to SSH keyfile (optional)` in the field might be nice as well.

CAM-Gerlach:
@bcolsen I would have, but the issue is that either field is only optional if the other field is filled out, which might be confusing while the text description makes it more clear. *Optional* is more for fields that are truly optional, independent of what other fields may or may not be filled.

ccordoba12:
> As a quick thought Path to SSH keyfile (optional)

I agree with @bcolsen here. You usually have your SSH password (which is the same one as your login password) but few seem to know what a passphrase file is.

ccordoba12:
My comments:

1. Please move the new `Note` text to be below the `This is a remote kernel` checkbox.
2. Please move password to be below `Host name`.
3. Please add @bcolsen suggestion in `SSH key file`.

CAM-Gerlach:
I updated the PR and the screenshot in the first post with all three changes as requested:

>  You usually have your SSH password (which is the same one as your login password) but few seem to know what a passphrase file is.

(I assume you mean "keyfile" here; "passphrase files" are what is sometimes used with GPG to encrypt/decrypt local files.)

That's quite surprising to me; I don't use SSH that regularly and never have within Spyder, nor (unfortunately) even *nix/POSIX as my primary OS recently, but I'm quite familiar with keyfiles and have generated and used ones both for both my main application of regular SSH, and for Git/Github. However, I'd take your word for it, given your much greater experience and exposure to the Spyder user community. 

> 1. Please move the new `Note` text to be below the `This is a remote kernel` checkbox.

Great point; I was going to do this initially but just didn't get around to fiddling with the layout.

> 2. Please move password to be below `Host name`.

I guess the justification for it being below "Keyfile" is that it may also be the keyfile's passphrase, but if many more users really just use their user password, then it makes more sense to put it under there even if it is somewhat less intuitive for a keyfile passphrase.

> 3. Please add @bcolsen suggestion in `SSH key file`.

I actually considered adding something very similar or identical when first making the changes, but couldn't find a short wording that would be clear, correct, and not redundant to the Note directly above. I'm still not a huge fan since the field *isn't* technically optional by itself as conventionally defined thus being potentially misleading and confusing, and it doesn't really add anything that isn't better explained by the Note immediately above. However, its enough of a borderline case that I figured it not worthwhile to engage in a futile argument over.

bcolsen:
I just mocked up this rough (needs layout and label name help) but functional alternative:

image
image



I would have to do a bit of Qt learning to clean it up myself but I think separating out the passwords fields and using radio buttons are a good way of only letting people do the right thing.

ccordoba12:
> I would have to do a bit of Qt learning to clean it up myself but I think separating out the passwords fields and using radio buttons are a good way of only letting people do the right thing.

I agree. It's also clear that the Passphrase is only need for the key file.

@CAM-Gerlach, what do you say?

CAM-Gerlach:
@bcolsen Thanks! I agree, once you've mocked it up it really looks like it would make for cleaner, clearer and more consistent UX. Of course, it really isn't an alternative but rather an improvement on top of this PR, since the dialog still needs most of the text changes anyway given the problems with and refinements to the existing the existing. So this could be merged if @ccordoba12 thinks its ready, then you base that on top of this.

ccordoba12:
I think @CAM-Gerlach is right. Let's merge this one, which is a considerable enhancement over the previous UI and then @bcolsen can submit the additional improvements he mocked up.

bcolsen:
Sounds good to me. This PR is a great improvement.

Should I submit against 3.x or master?

ccordoba12:
Let's leave it for master, to not change this dialog too much in 3.x

PR #7902, opened 2018-09-15
PR:  Improve detection of new Spyder versions to be offered for updates

### LINKED ISSUE #7862: 4.0beta1 is offered as an update when using a development version and a virtualenv
We need to avoid showing beta versions when checking for updates

ccordoba12:
Fixes #7862

### DIFF
--- a/spyder/utils/tests/test_programs.py
+++ b/spyder/utils/tests/test_programs.py
@@ -15,7 +15,8 @@ from spyder.utils.programs import (run_python_script_in_terminal,
                                    is_python_interpreter,
                                    is_python_interpreter_valid_name,
                                    find_program, shell_split, check_version,
-                                   is_module_installed, get_temp_dir)
+                                   is_module_installed, get_temp_dir,
+                                   is_stable_version)
 
 
 if os.name == 'nt':
@@ -130,6 +131,12 @@ def test_get_temp_dir_ensure_dir_exists():
 
     assert os.path.exists(another_call)
     assert another_call == temp_dir
+
+
+def test_is_stable_version():
+    """Tests for is_stable_version."""
+    assert is_stable_version('3.3.0')
+    assert not is_stable_version('4.0.0b1')
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


ccordoba12:
The code to detect a development version was wrong.

spyder/utils/tests/test_programs.py
```diff
@@ -132,5 +133,11 @@ def test_get_temp_dir_ensure_dir_exists():
     assert another_call == temp_dir
 
 
+def test_is_stable_version():
+    """Tests for is_stable_version."""
```
CAM-Gerlach: Imperative tense.

spyder/utils/tests/test_programs.py
```diff
@@ -132,5 +133,11 @@ def test_get_temp_dir_ensure_dir_exists():
...
+    assert is_stable_version('3.3.0')
+    assert not is_stable_version('4.0.0b1')
```
CAM-Gerlach: For completeness, would you want to check ``'3.3.2.dev0`` as well, since its of a slightly different form?

spyder/workers/tests/test_update.py
```diff
@@ -22,13 +22,20 @@ def test_update(qtbot):
...
-def test_not_update(qtbot):
+def test_no_update(qtbot):
     """Test the update checking for a version that don't needs an update."""
     worker = WorkerUpdates(None, False, version="1000.0.0")
     worker.start()
     assert not worker.update_available
     assert not check_version("1000.0.0", worker.latest_release, '<')
 
 
+def test_no_update_development(qtbot):
```
CAM-Gerlach: Should beta versions be considered development versions for the purposes of the updater? Either way, shouldn't we test that they're handled properly?

ccordoba12:
> A few minor comments/suggestions; feel free to ignore them if you don't agree with any of them.

All of them were good suggestions and led me to improve much more this work. Thanks a lot for your review @CAM-Gerlach!

PR #7914, opened 2018-09-18
PR: Have a choice between password or keyfile in "Connect to remote kernel" dialog

### LINKED ISSUE #7905: Connect to remote kernel: Have a choice between password or keyfile
Feature Request:

As discussed in #7898 

image
image

bcolsen:
### Pull Request Checklist

### Affirmation

## Description of Changes

Making the external connections more user friendly

image

remote kernel

### Issue(s) Resolved

Fixes #7905


bcolsen:
I forgot to leave some comments.

I had to but the link in the top Qlabel beacause the Qboxgroup title dosn't take html.

The link in the top doesn't do anything when I click it, but I can right click and copy the link

bcolsen:
I also need to do some manual testing on this still

ccordoba12:
> I had to but the link in the top Qlabel beacause the Qboxgroup title dosn't take html.

Then please add a QLabel below `This is a remote kernel` and add the link there. If we leave it in the top label, I'm afraid very few people will take a look at it.

> The link in the top doesn't do anything when I click it, but I can right click and copy the link

You need to activate an special option in QLabel for that to happen. Please take at our Github login dialog for that.

CAM-Gerlach:
Could you add a little more vertical space between ``Username`` and ``Authentication method``?

> You need to activate an special option in QLabel for that to happen. 

You can try ``self.setTextInteractionFlags(Qt.LinksAccessableByMouse)`` in ``KernelConnectionDialog``, which fixed the same issue in the About dialog. Or there might be something different @ccordoba12 's referring to.

> If we leave it in the top label, I'm afraid very few people will take a look at it.

I'm not so sure about that, since its bright blue and quite prominent there too. Additionally, the linked section describes the whole external kernel connection process (i.e. the whole dialog), not just connecting to a remote kernel over SSH, so I'd aver that the top of the dialog is therefore a more appropriate spot anyway. Finally, it avoids adding a whole extra line of text to the dialog, which itself generally makes people more likely to skim/skip things or miss important points.

ccordoba12:
Ok, but perhaps we could separate the top label into two paragraphs for clarity? The second paragraph could be something like

> If you are connecting to a remote kernel, please refer to [our documentation](http://docs.spyder-ide.org/ipythonconsole.html#connect-to-an-external-kernel) to understand how to achieve it.

(I'm sure @CAM-Gerlach can improve the wording)

CAM-Gerlach:
> Ok, but perhaps we could separate the top label into two paragraphs for clarity?

I'm not sure I see how it would make it any more clear, as it just adds extra words and lines of text without adding any more useful information. As I mentioned above, keeping such filler to a minimum is key to maximizing the chance users actually read and understand it, and don't miss what's actually important. Furthermore, in my experience, ([more details](http://docs.spyder-ide.org/ipythonconsole.html#connect-to-an-external-kernel)) is a more common UI text pattern for these situations, while including a whole extra paragraph just to link to a documentation section with more details is not, at least for a dialog like this.

With regard to the text above:

> If you are connecting to a remote kernel

Again, the referenced documentation section explains how to connect to any external kernel (i.e. the purpose of the dialog), not just a remote one, so this is both redundant and not fully accurate.

Cutting out that as well as the filler and verbosity, we are left with:

> See [the documentation](http://docs.spyder-ide.org/ipythonconsole.html#connect-to-an-external-kernel) for details.

But again, that doesn't say anything more than ([more details](http://docs.spyder-ide.org/ipythonconsole.html#connect-to-an-external-kernel)) on the line above, while taking up needless space on the dialog, time for a user to read, and making things more complex than they should be.

bcolsen:
> @bcolsen , what do you think?

Currently 4 of the 5 bullet points in the linked documentation are for finding the number or file of the kernel you want to link too and the 5th being SSH. So I think the documentation link with the top paragraph is fine. 

I don't know if most users would even know how to start a IPython kernel to connect to for the first time without using the link to the documentation. I didn't until a few days ago. So a link to the docs is generally handy.

I guess we could mention `python -m spyder_kernel.console` directly in the header. 

I could also separate the ID and file part for the kernel in a similar way to authentications methods sections.

CAM-Gerlach:
Your thought is indeed a good one; I'd just question the proposed implementation details:

> I guess we could mention `python -m spyder_kernel.console` directly in the header. 

Well, perhaps, but then its a slippy slope—for the new users who would benefit from seeing it, it wouldn't do much to include that in isolation without mentioning installing ``spyder-kernels`` first so it actually works, and then in turn getting the right version of Spyder-Kernels which has tripped up a lot of people, and making sure their in an Anaconda Prompt (on Win) and the right env, etc, and so on down the rabbit hole, plus potentially some other basic background. That's at least a few lines at the very minimum, plus we'd have to format the inline code, set the flag to make the text selectable (so people can copy it), etc. Again, as mentioned before, that adds more complexity, potential confusion, and people skipping/missing something important about the dialog itself (which the questions we got that prompted this were all about in the first place).

There's also a fair argument to be made that all of this is out of scope for the dialog's explanatory text to begin with, since its titular purpose is to connect to *existing* kernels that the user has *already* started, not create new ones (which can, would and typically should be done via changing the python interpreter in prefs and launching the kernel via the normal way in Spyder, at least on the local machine); the primary scenario I'm aware of where the user might want to create a new kernel right then specifically for Spyder and then connect to it would be a remote one, which is more complex and requires explaining more steps and caveats anyway, thus the need to reference the docs regardless if one is unfamiliar with the parts that don't directly relate to the dialog itself, which are already explained there.

What I might suggest, actually, if we agree this is a significant concern, is to bring back @ccordoba12 's previous suggestion of a separate one-sentence paragraph with the docs link, only now with an clearer purpose behind the extra text. E.g. something like:

> For more details on starting and connecting to an external kernel, see [our documentation].(http://docs.spyder-ide.org/ipythonconsole.html#connect-to-an-external-kernel).

> I could also separate the ID and file part for the kernel in a similar way to authentications methods sections.

I also thought about suggesting that for similar reasons, but have somewhat mixed feelings—it does make it somewhat more clear, explicit and unambiguous, but there is the cost in increased size and complexity of the dialog to consider, which is why I ended up not advocating it since it wasn't clear which was the better tradeoff. So, I could see either way, although a lot would hinge on how it ended up looking. @ccordoba12 , your opinion?

bcolsen:
> For more details on starting and connecting to an external kernel, see [our documentation]

I think I'll go with that.  This is an advanced feature so reading the docs is a good idea.

As for splitting the id and files. We could just get rid of the ID option entirely then we wouldn't need the header to explain it.

I consider this especially important since the ID don't work for remote kernels(correct me if I'm wrong) and that's not obvious going in. When I try ID number on a remote kernel get `Unable to connect to 22922`. I could make this error better but we would also need to add another line in the header to indicate that ID's don't work for remote kernels.  

For local kernels we don't even lose too much, since the browse button takes you to the directory with the files so you just pick the number from there. 

As far as I can see the ID numbers bring more trouble and confusion than they are worth.

If we get rid of the ID the only header we need would say something like this:
>This dialog connects to an existing local or remote IPython kernel using a JSON connection file. For more details on starting and connecting to an external kernel, see [our documentation]

Like so:
image


That being said I think most people would rather the ID's work for the remote kernels because coping the file from a remote computer is inconvenient especially given the non-user path. It should be possible for Spyder to read the file over ssh given the ID (we have the password) but if we're going that far we could also start a kernel for them. Obviously, this would be outside the scope of  this PR.

ccordoba12:
> I consider this especially important since the ID don't work for remote kernels(correct me if I'm wrong)

It works if you put your kernel json file inside the jupyter runtime dir

> As far as I can see the ID numbers bring more trouble and confusion than they are worth.

This is probably true. But it's a great way to quickly test things, so please don't remove the functionality that we already have in qtconsole.

> Like so:

I really like how the dialog looks now! Thanks a lot for this.

I also agree with leaving the link to our docs in the top label and first paragraph.

ccordoba12:
@bcolsen, what about remembering user introduced data?

We could make the dialog remember hostname/username/port and password. For the last one we can use keyring, which is already one of our dependencies.

CAM-Gerlach:
>> Like so:

> I really like how the dialog looks now! Thanks a lot for this.

@ccordoba12 Do note that is with the ID number removed, which you just asked @bcolsen not to do, jeje.

> I also agree with leaving the link to our docs in the top label and first paragraph.

Looks like we've sort of "swapped sides" then, heh, with me suggesting your previous suggestion! But that's fine with me I guess.

> We could make the dialog remember hostname/username/port and password. For the last one we can use keyring, which is already one of our dependencies.

Great idea; I presume you saw the user request and our discussion about it or else that's a very lucky coincidence that you're suggesting the exact same feature and implementation, ha! If you think that's in scope and @bcolsen 's on board to do it, I'm all for it.

We would also presumably want to store the path and passphrase to the SSH keyfile with ``keyring``.

CAM-Gerlach:
> I suggest we can leave the dialog without mentioning ID and leave the ID number implemented in the code and mention it in the documentation. This makes the dialog simple and keeps the functionality for those who currently use it.

I dunno how I feel about leaving un-documenting such functionality; sure the current UI is not the cleanest, but I'm not sure if it justifies doing that. But again, I dunno.

> I could add ID to the end of hint:

Something like that is just likely to make it more confusing then the current situation. Either we don't mention it at all, or we mention it properly.

bcolsen:
>Either we don't mention it at all, or we mention it properly.
I'm all for not mentioning it and not implementing it.  The browse button goes to the only path that matters anyway.

I also think that not mentioning it in the dialog, leaving it implement in code and having a line in the documentation like:
>As a convenience kernel ID numbers (e.g. 20345) entered in the connection file path field will be expanded to `jupter_runtime_dir\kernal-<ID>.json` on your local machine.

This explains exactly what the code does.

CAM-Gerlach:
> I'm all for not mentioning it and not implementing it. The browse button goes to the only path that matters anyway.

Did you mean that to be quoted?

I mean, I'm not dead-set against it. I just don't think its really much of an net improvement to the current situation. @ccordoba12 what do you think?

ccordoba12:
> I also think that not mentioning it in the dialog, leaving it implement in code and having a line in the documentation like:

>>    As a convenience kernel ID numbers (e.g. 20345) entered in the connection file path field will be expanded to jupter_runtime_dir\kernal-<ID>.json on your local machine.

> This explains exactly what the code does.

I totally agree with @bcolsen here. That's what I'd like to have too.

CAM-Gerlach:
Alright, sure. I added it to my existing docs PR clarifying how to find the kernel connection file/ID on the remote machine, since its small and pretty related.

ccordoba12:
> Yes this would be handy. I'll have a look at the github login implementation and see what I can do

Nice! Then I'd suggest to move KernelConnectionDialog out of plugin.py into its own module. It really doesn't need to live in the same file as the plugin and this work will certainly make that class bigger.

CAM-Gerlach:
Oh darn, I didn't see you already merged that one a few minutes before I added the change. My bad; I'll create a new PR when I get back but I'm now very late for class as it is.

CAM-Gerlach:
Done, as spyder-ide/spyder-docs#84 .

Also, @bcolsen , I'd suggest some minor revision to the UI text. For the top label:

> Please select the JSON connection file (*e.g.* ``kernel-1234.json``) of the existing kernel to connect to, and enter the SSH connection information if a remote kernel.

> To learn more about starting external kernels and connecting to them, see  [our documentation](http://docs.spyder-ide.org/ipythonconsole.html#connect-to-an-external-kernel).

I actually now prefer it with the linebreak that @ccordoba12 originally advocated for, since ~~I always gotta be a CAMtrarian~~ since it has a meaningfully distinct purpose from the first line. But I'm not going to make a huge fuss about it if not.

Also, for the placeholder text, a slight tweak: I suggest ``Kernel connection file path`` so as to make it more concise and to the point.

I really like how you have iterated this; thanks for bearing with us on this!

bcolsen:
> I really like how you have iterated this; thanks for bearing with us on this!

Thanks for all the ideas and encouragement. I think this is looking quite nice at this point.

Here is the latest:
image

I changed the top text to have the if statement before the ssh stuff, because people might not read to the end of that sentence.

I'm going to make a commit here because everything works right now. Then I'm going to break it with the keyring stuff.

CAM-Gerlach:
The e.g. is not italic and the ``kernel-1234.json`` is not code/monospaced, which are both regressions from the previous version; that needs to be fixed.

> I changed the top text to have the if statement before the ssh stuff, because people might not read to the end of that sentence.

I'm not a big fan of the change, since its a tad awkward, wordier than the original, and introduces some problematic repetition, and so I'm not entirely sure I follow how adding more verbosity makes people less likely to skim over it. (Of course, I'm guilty myself in that latter regard. using some variant of "connect" and "kernel" no fewer than three times each in that sentence...wow, how could I miss that?)

If you really want two sentences for some reason, the best I can do is:

> Please select the JSON connection file (*e.g.* ``kernel-1234.json``) of the existing kernel. Additionally, enter the SSH information if connecting to a remote machine.

The one sentence version of the same (that I prefer):

> Please select the JSON connection file (*e.g.* ``kernel-1234.json``) of the existing kernel, and enter the SSH information if connecting to a remote machine.

bcolsen:
> If you really want two sentences

I don't need two sentences. I liked having the if statement first before the SSH stuff. Maybe I've been programing too long.

I'm good with this one though:

>Please select the JSON connection file (e.g. kernel-1234.json) of the existing kernel, and enter the SSH information if connecting to a remote machine.

image

CAM-Gerlach:
Okay, thanks.

Ugh that hanging chad is aweful, but since it depends on the platform, UI theme, system font and dialog size, I'm not sure how much can really be done to micromanage it.

image

bcolsen:
>Ugh that hanging chad is aweful, but since it depends on the platform, UI theme, system font and dialog size, I'm not sure how much can really be done to micromanage it.

Yeah it's system dependent. By removing the paragraph it can only happen once instead of twice.

image

CAM-Gerlach:
In our public Gitter chat, @hofingermarkus brought up the point that if any of the details are incorrect or the connection otherwise fails, the dialog immediately closes, and then issues a warning to the effect. The user then has to reopen it, re-enter everything (since the dialog doesn't remember the last entries), and try again, hoping they don't get the same outcome. 

Of course, this is completely nonsensical UX (independent of remembering the entries), as in essentially all reasonable cases, the user is likely to want to try again with different settings, fix some extra-Spyder problem and retry, or at the very least review the settings that failed. Therefore, the warning dialog should simply open as a child of the connection dialog, and when closed the user can view or edit the still-existing  parent connection dialog and retry.

ccordoba12:
I think we can merge these improvements at this point. @bcolsen, do you think this is ready for review?

bcolsen:
@ccordoba12  I've been working on contract work for the last couple of weeks ( and the next few) so this is a good place to wrap this one up. It's ready for review.

I'll make an issue to track the form waiting until connected, and another one for remembering passwords ( thought of some ideas).

bcolsen:
I'll put in some pep8 fixes first.

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -28,10 +28,11 @@
...
 from qtpy.QtCore import Qt, Signal, Slot
-from qtpy.QtWidgets import (QApplication, QCheckBox, QDialog, QDialogButtonBox,
-                            QFormLayout, QGridLayout, QGroupBox, QHBoxLayout,
-                            QLabel, QLineEdit, QMessageBox, QPushButton,
-                            QTabWidget, QVBoxLayout, QWidget)
+from qtpy.QtWidgets import (QApplication, QButtonGroup, QCheckBox, QDialog,
+                            QDialogButtonBox, QFormLayout, QGridLayout,
+                            QGroupBox, QHBoxLayout, QLabel, QLineEdit,
+                            QMessageBox, QPushButton, QRadioButton,
+                            QSpacerItem, QTabWidget, QVBoxLayout, QWidget)
```
ccordoba12: I think most of these imports are not needed now that you moved the dialog to a different file. So please check if they are still used here or not.

ccordoba12: @bcolsen, I think these imports (in the plugin, I mean) also need some cleanup.

bcolsen: Oops. I need to read more closely. I'll get those too.

PR #7929, opened 2018-09-21
PR: Improve the Shortcut Editor and fix and extend emacs shortcut support

### LINKED ISSUE #7109: Allow keyboard shorcuts to be cleared in preferences
I looked but I could find a way to clear a keyboard shortcut in the preferences.  This option would disable the shortcut without needing to set it something else. This could be a button like `clear` on the keyboard setting dialog.

shortcuts
With a `clear` or `unset` button on there.

This would be handy sine then we could have entries in the shortcuts that are set to nothing by default but could be user enabled

spyder version: master

### LINKED ISSUE #7854: Problem with shortcuts using the "Shift" and another key
### Issue Report Checklist

## Problem Description

As mentioned in https://github.com/spyder-ide/spyder/pull/7768#issuecomment-418938416, it is not possible to assign a key sequence using Shift and another key in Spyder.

image

Unfortunately, 4 shortcuts use this by default in Spyder :

image

This means that if one binds `Ctrl+Y` to the `clear line` shortcut for example, they won't be able to use the `Shift+Esc` sequence elsewhere or even to revert to the `Shift+Esc` afterwards for the `clear line` shortcut, unless they reset all shortcuts to their default key sequences. 

So, I think we need to change a little bit this rule and add exceptions to it, so that one can use `Shift+Enter`, `Shift+Delete`, `Shift+Insert`, etc as valid shortcut key sequences.

### LINKED ISSUE #7883: GNU Emacs Style Key Sequences do not work in master AND shortcut conflicts
## Problem Description

Key sequences similar to those used in GNU Emacs , allowing up to four key codes, do not work in the Editor after the changes introduced in PR #7768. I haven't thought about this case when working on this PR, so this needs to be fix.

@ccordoba12 what do you think  is the severity of this?

Also, I'm not sure, but I think there is a problem with how we check for conflicts for these kind of shortcuts in our keyboard preference editor. For example, I do not understand how a shortcut with a keyseq defined as `Ctrl+C, then Ctrl+X` could work if `Ctrl+C` is already used for `copy` and `Ctrl+X` is already used for `cut`.

image

Ressources:
https://srinikom.github.io/pyside-docs/PySide/QtGui/QKeySequence.html#gnu-emacs-style-key-sequences
https://en.wikipedia.org/wiki/Table_of_keyboard_shortcuts
https://www.gnu.org/software/emacs/refcards/pdf/refcard.pdf

jnsebgosselin:
### Issue(s) Resolved

Fixes #7883 
Fixes #7109
Fixes #7854

## Description of Changes

- [x] Add a button to get the default key sequence.
- [x] Allow clearing the key sequence for a shortcut.
- [x] Allow `Shift+<another key>` as a valid shortcut in the Editor.
- [x] ~~Fix problems when the `Shift` modifier is used with international keyboards.~~
- [x] Extend and fix support for emacs shortcut in the Shortcut Editor
- [x] ~~Add support for emacs key sequence style in the Editor~~
- [x] Add tests for the Shortcuts Editor
- [x] ~~Create an Issue for moving all shortcut ultilities from `spyder/utils/qthelpers` into a new module `spyder/utils/shortcuts`~~

Edit: Some items were stiked because they were fixed or made irrelevent with the changes made in PR #7954

image

shortcut_editor

### Affirmation

<!--- TYPE YOUR GITHUB USERNAME OR FULL NAME AFTER THE BELOW STATEMENT ---!>
I affirm all of the above: Jean-Sébastien Gosselin


jnsebgosselin:
@ccordoba12 If you want to give me some feedback about the changes I'm proposing for the Shortcut Editor, I would appreciate it very much.

I still need to add emacs support in the Editor and add tests for the Shortcut Editor since this was not covered at all it seems.

spyder/preferences/shortcuts.py
```diff
@@ -16,62 +16,87 @@
...
+                       'are supported in the Editor only.')
+
+
+class ShortcutTranslater(QKeySequenceEdit):
+    """
+    A `QKeySequenceEdit` that is not meant to be shown and is used only
+    to convert `QKeyEvent`s into `QKeySequence`s. To our knowledge, this is
+    the only way to do this within the Qt framework because the code that does
+    this in Qt is protected. Porting the code to Python would be nearly
+    impossible because it relies on low level and OS dependent Qt librairies
```
ccordoba12: librairies -> libraries

spyder/preferences/shortcuts.py
```diff
@@ -16,62 +16,87 @@
...
 VALID_FINDER_CHARS = r"[A-Za-z\s{0}]".format(VALID_ACCENT_CHARS)
 
 BLACKLIST = {}
-BLACKLIST['Shift'] = _('Shortcuts that use Shift and another key'
-                       ' are unsupported')
+BLACKLIST['Shift'] = _('Shortcuts using Shift with no other modifier keys '
+                       'are supported in the Editor only.')
+
+
+class ShortcutTranslater(QKeySequenceEdit):
```
ccordoba12: Is this right? Why not `ShortcutTranslat*o*r`

CAM-Gerlach: > ShortcutTranslator

would be the correct spelling in English, for what its worth.

spyder/preferences/shortcuts.py
```diff
@@ -16,62 +16,87 @@
...
+    """
+    A `QKeySequenceEdit` that is not meant to be shown and is used only
+    to convert `QKeyEvent`s into `QKeySequence`s. To our knowledge, this is
```
CAM-Gerlach: Again, same as above with regard to the syntax for code/literal names.

spyder/preferences/shortcuts.py
```diff
@@ -16,62 +16,87 @@
...
+    the only way to do this within the Qt framework because the code that does
```
CAM-Gerlach: framework*,* because

spyder/preferences/shortcuts.py
CAM-Gerlach: OS-dependent

spyder/preferences/shortcuts.py
```diff
@@ -81,9 +106,18 @@ def keyReleaseEvent(self, e):
         """Qt Override"""
         self.parent().keyReleaseEvent(e)
 
+    def setText(self, sequence):
+        """Qt method extension."""
+        if sys.platform == 'darwin':
+            sequence = sequence.replace('Meta', 'Ctrl')
```
CAM-Gerlach: The official name of the key on all Mac keyboards as well as most if not all the Mac literature is ``Control``, not ``Ctrl``. Furthermore, this also makes it more likely to be confused with ``Cmd``, since the general convention is to translate ``Ctrl`` (the PC/Linux key) --> ``Cmd``. Therefore, that should be used, unless some technical restriction prevents it.

spyder/preferences/shortcuts.py
```diff
@@ -122,47 +156,77 @@ def keyPressEvent(self, event):
...
+    def setup(self):
+        """Setup the ShortcutEditor with the provided arguments."""
         # Widgets
         self.label_info = QLabel()
-        self.label_info.setText(_("Press the new shortcut and select 'Ok': \n"
-             "(Press 'Tab' once to switch focus between the shortcut entry \n"
-             "and the buttons below it)"))
+        self.label_info.setText(
+            _("Press the new shortcut and select 'Ok' "
+              "or press 'Clear' to unbind the shortcut."))
```
CAM-Gerlach: Press the new shortcut and select ``Ok`` to confirm,
click ``Cancel`` to revert to the previous state,
or use ``Clear`` to unbind the command from a shortcut.

spyder/preferences/shortcuts.py
```diff
@@ -122,47 +156,77 @@ def keyPressEvent(self, event):
...
+        button_box.addWidget(self.button_default)
+        button_box.addStretch(100)
+        button_box.addWidget(self.button_ok)
+        button_box.addWidget(self.button_clear)
+        button_box.addWidget(self.button_cancel)
+
+        # New Sequence button box
+        self.btn_clear_sequence = create_toolbutton(
+            self, icon=ima.icon('editclear'),
+            tip=_("Clear"),
```
CAM-Gerlach: Something like "Clear all entered shortcuts" is more helpful/informative here.

spyder/preferences/shortcuts.py
```diff
@@ -122,47 +156,77 @@ def keyPressEvent(self, event):
...
+        self.button_cancel = QPushButton(_('Cancel'))
+        button_box = QHBoxLayout()
+        button_box.addWidget(self.button_default)
+        button_box.addStretch(100)
+        button_box.addWidget(self.button_ok)
+        button_box.addWidget(self.button_clear)
+        button_box.addWidget(self.button_cancel)
+
+        # New Sequence button box
+        self.btn_clear_sequence = create_toolbutton(
```
CAM-Gerlach: Since it clears the entire dialog vs. just deleting the most recent shortcut, it stands to reason that this should go second, and the "delete last shortcut" button first (i.e. swap them).

spyder/preferences/shortcuts.py
```diff
@@ -122,47 +156,77 @@ def keyPressEvent(self, event):
...
+            self, icon=ima.icon('editclear'),
+            tip=_("Clear"),
+            triggered=self.clear_new_sequence)
+        self.button_back_sequence = create_toolbutton(
+            self, icon=ima.icon('ArrowBack'),
+            tip=_("Back"),
```
CAM-Gerlach: This could be made more accurate and useful with a tooltip like "Remove last shortcut entered``.

spyder/preferences/shortcuts.py
```diff
@@ -214,118 +292,82 @@ def accept(self):
...
-        self.npressed += 1
-        self.key_non_modifiers.append(key)
-        self.key_modifiers.add(key)
-        self.key_text.append(e.text())
-        self.invalid_key_flag = False
-
-        debug_print('key {0}, npressed: {1}'.format(key, self.npressed))
+    def event(self, e):
+        """Qt method override."""
+        if type(e) in [QEvent.Shortcut, QEvent.ShortcutOverride]:
```
CAM-Gerlach: Technically, best practices would suggest making this a set (or at least a tuple), since its much faster and more conceptually linked to the use case, if my understanding is correct. Also, is there a reason for using ``type(e) in [...]`` instead of ``isinstance(e, (QEvent.Shortcut, QEvent.ShortcutOverride))``?

spyder/preferences/shortcuts.py
```diff
@@ -214,118 +292,82 @@ def accept(self):
...
+                valid_single_keys = (EDITOR_SINGLE_KEYS if
+                                     self.context == 'editor' else SINGLE_KEYS)
+                if any((k == keystr for k in valid_single_keys)):
+                    return True
+                else:
+                    return False
+
+    def check_shift_keyseq(self):
+        """
+        Check if the first sub-sequence of the new key sequence is composed
```
CAM-Gerlach: See comment regarding PEP 257.

spyder/preferences/shortcuts.py
```diff
@@ -214,118 +292,82 @@ def accept(self):
...
+        with the 'Shift' modifier only and another key.
```
CAM-Gerlach: --> with 'Shift' and no other modifier keys.

spyder/preferences/shortcuts.py
```diff
@@ -381,99 +423,62 @@ def update_warning(self, warning_type=NO_WARNING, conflicts=[]):
...
-        while u'' in self.key_text:
-            self.key_text.remove(u'')
+            self.button_ok.setEnabled(warning == NO_WARNING)
+            self.update_warning(warning_type=warning, conflicts=conflicts)
 
-        self.key_text = [k.upper() for k in self.key_text]
+    def set_sequence_to_default(self):
+        """Set the new sequence to the default value defined in the config."""
+        default_keystr = QKeySequence(
+            CONF.get_default('shortcuts', '%s/%s' % (self.context, self.name))
```
CAM-Gerlach: Any reason to not use the more modern and clear ``format()`` syntax here? Particularly since its what you use elsewhere...

spyder/preferences/shortcuts.py
jnsebgosselin: I do not know Mac at all, so I'll take your words for it. I wish we had someone with a Mac on the team to test stuff....

spyder/preferences/shortcuts.py
jnsebgosselin: I ported this from C++ Qt codebase, that's why. I'm going to make it more Pythonic.

jnsebgosselin: mmmh, your suggestion results in this error:
```TypeError: isinstance() arg 2 must be a type or tuple of types```
So I'm going to keep my formulation.

File: spyder/preferences/shortcuts.py (hunk 11 above)
Comment (jnsebgosselin) [reply spyder/preferences/shortcuts.py]: You mean adding a one-line summary to a two-line docstring? I don't see the added value here. I mean, I get it for a public API or a Library where the docstrings are used to actually build some public documentation, but for documenting the code source of a GUI application? I only see noise and wasted space...

If @ccordoba12 starts enforcing this in all new PR, I'll happily conform to this rule though :P

**Comment by jnsebgosselin** [comment]
@ccordoba12 @CAM-Gerlach Thank you very much for your help with this, I really appreciate it.

**Comment by jnsebgosselin** [comment]
image

File: spyder/preferences/shortcuts.py (hunk 11 above)
Comment (CAM-Gerlach) [reply spyder/preferences/shortcuts.py]: Yeah, you have a point. I know they hasn't been really enforced in similar situations (two-line docstrings) for tests, since, well, they're only tests after all. In our contributing guide/other resources, we do specify following PEP 257, which states:

> Multi-line docstrings consist of a summary line just like a one-line docstring, followed by a blank line, followed by a more elaborate description. The summary line may be used by automatic indexing tools; it is important that it fits on one line and is separated from the rest of the docstring by a blank line.

Of course, following the ethos of Python, the PEP also reminds us at the top,

> The PEP contains conventions, not laws or syntax. If you violate these conventions, the worst you'll get is some dirty looks. But some software [...] will be aware of the conventions, so following them will get you the best results.

So yeah, up to @ccordoba12 what to do. Either way, it sets a bit of a precedent to follow going forward.

Comment (ccordoba12) [reply spyder/preferences/shortcuts.py]: I'm with Jean on this one. Most of the docstrings we add are informative for other core developers but not meant for public consumption. So we can relax PEP 257 in our codebase (I'm not following either).

File: spyder/preferences/shortcuts.py (hunk 5 above)
Comment (CAM-Gerlach) [reply spyder/preferences/shortcuts.py]: I grew up a big Mac user, but I switched to Windows (with some use of Linux) in high school since I couldn't afford to buy my own Mac laptop, and particularly with Apple's increasing "smartphonification" and the inability to build my own machine, there's little chance of me ever switching back.

I suppose we could consider using some of our $$$ and purchasing e.g. a cheap MacBook Pro (>= mid-2012, so its compatible with 10.14+) for one of the devs to test on, but that would be [at least $300-400](https://www.ebay.com/sch/i.html?_sop=12&Release%2520Year=2012%7C2013%7C2014&LH_ItemCondition=3000%7C2500%7C2000%7C1500&Processor%2520Type=Intel%2520Core%2520i5%25203rd%2520Gen%252E%7CIntel%2520Core%2520i5%25204th%2520Gen%252E%7CIntel%2520Core%2520i7%25203rd%2520Gen%252E&_fsrp=1&_nkw=MacBook+Pro&_sacat=111422&LH_Complete=1&rt=nc&_dcat=111422&LH_Sold=1).

File: spyder/preferences/shortcuts.py [hunk 14]
```diff
@@ -16,62 +16,87 @@
...
 
 # Valid finder chars. To be improved
 VALID_ACCENT_CHARS = "ÁÉÍOÚáéíúóàèìòùÀÈÌÒÙâêîôûÂÊÎÔÛäëïöüÄËÏÖÜñÑ"
 VALID_FINDER_CHARS = r"[A-Za-z\s{0}]".format(VALID_ACCENT_CHARS)
 
 BLACKLIST = {}
-BLACKLIST['Shift'] = _('Shortcuts that use Shift and another key'
-                       ' are unsupported')
+BLACKLIST['Shift'] = _('Shortcuts using Shift with no other modifier keys '
+                       'are supported in the Editor only.')
```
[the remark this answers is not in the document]
CAM-Gerlach: Is there a reason this was not implemented, or was this not meant to be marked resolved without fixing it?

spyder/preferences/shortcuts.py
CAM-Gerlach: Okay, sure, if you think so.

CAM-Gerlach:
Glad I could be helpful, @jnsebgosselin !

A couple followups—that's a lot more vertical space than I was thinking of between the top label and the Current shortcut; I meant more like 1.5-2x the previous space, not close to ~5x like the above.

Also, the revised text for the "Shortcuts using Shift..." message was not implemented, at least judging from the above. Not sure if that was unintentional, or you have some rationale for not doing so.

spyder/preferences/shortcuts.py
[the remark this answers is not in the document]
jnsebgosselin: This is quoted from you in Gitter :

> "Shortcuts using Shift with no other modifier keys are supported only in the Editor."
Or "Shortcuts using Shift with no other modifier keys are supported in the Editor only."
I think that's a little better for flow/readability putting the "only" at the end.

Last call, where do you want that only finally? :P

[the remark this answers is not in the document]
CAM-Gerlach: > This is quoted from you in Gitter

Ah, so that's why I couldn't find it :) I only looked on Github, but I was wondering if I was disagreeing with myself and looking silly in the process since I'm always tweaking and refining my work. My bad.

Considering all that, I can confirm I stand by what I stated above:

> Shortcuts using Shift with no other modifier keys are only supported in the Editor.

Is better than both my Gitter suggestions; putting "only" before "supported" is what really makes the difference, IMO. Not sure why I didn't try it before (or discounted it).

jnsebgosselin:
> A couple followups—that's a lot more vertical space than I was thinking of between the top label and the Current shortcut; I meant more like 1.5-2x the previous space, not close to ~5x like the above.

Thanks for noticing. This was not by design, but a problem with the way I structured the layout. I had to rework it a little and here is the result:

image

jnsebgosselin:
@ccordoba12 I need to use the new `ShortcutTranslator` class in the `CodeEditor`. The problem is, I cannot import it in `codeeditor.py` because it causes an error due to circular imports.

Can I move this class in a new module in `spyder/utils/shortcuts.py`, so that I can import it from `spyder/preferences/shortcuts.py` and `spyder/plugins/editor/widgets/codeeditor.py` ? Or is there another place that would be better suited to save this new class?

ccordoba12:
> Can I move this class in a new module in spyder/utils/shortcuts.py

Please put it in `utils/qthelpers.py` because that module already has other shortcuts functions. Or you could move all those functions and ShortcutTranslator to `utils/shortcuts.py`.

It's your call, but I think the second option is the best one.

jnsebgosselin:
Ok, thank you Carlos. I will go with the first option. The second is a good idea, but I would prefer to do it in a separate PR.

jnsebgosselin:
@ccordoba12 would it be ok to support Emacs keybindings only for the Editor? Otherwise, it gets complicated to manage key sequence conflicts.

For example, if a global shortcut is set to `Ctrl+C, X`, then a local shortcut set to `Ctrl+C, U` in the Editor won't work because the `Ctrl+C` event will be eaten by the global shortcut and will never be sent to the Editor.

jnsebgosselin:
Besides, we still have problems with builtin shortcuts outside of the Editor.

So key sequences like `Ctrl+X, Ctrl+F` or `Ctrl+X, Ctr+S` won't work when set to a global shortcut, because `Ctrl+X` is a key sequence associated to the builtin shortcut `cut`.

ccordoba12:
I understand your predicament but there are some Emacs shortcuts that should work globally. For example:

    Exit Emacs permanently: C-x C-c

Ideally, we should try to support as most shortcuts as we can from page 1 of this cheat-sheet:

[Emacs keybindings.pdf](https://github.com/spyder-ide/spyder/files/2412886/Emacs.keybindings.pdf)

jnsebgosselin:
> 
> 
> I understand your predicament but there are some Emacs shortcuts that should work globally. For example:
> 
> ```
> Exit Emacs permanently: C-x C-c
> ```
> 
> Ideally, we should try to support as most shortcuts as we can from page 1 of this cheat-sheet:
> 
> [Emacs keybindings.pdf](https://github.com/spyder-ide/spyder/files/2412886/Emacs.keybindings.pdf)

Ok I will see what I can do.

That being said, due to the situation with builtin shortcuts we covered in PR #7768, I do not think it will be possible to make Emacs shortcut starting  with `Ctrl+C`, `Ctrl+Z`, `Ctrl+X`, and the likes to work globally. To make this work, we would have to completely refactor how we handle shortcut in Spyder...

ccordoba12:
> To make this work, we would have to completely refactor how we handle shortcut in Spyder...

Ok, then please focus on the Editor for now because it took us years and several refactorings to achieve the current state.

A new refactoring could be left for Spyder 5.

jnsebgosselin:
> 
> 
> > To make this work, we would have to completely refactor how we handle shortcut in Spyder...
> 
> Ok, then please focus on the Editor for now because it took us years and several refactorings to achieve the current state.
> 
> A new refactoring could be left for Spyder 5.

Ok, thanks Carlos. Even with refactoring, I really don't know how this could be achieved... maybe I'll find a solution while working on the Editor.

jnsebgosselin:
https://stackoverflow.com/questions/13363332/hard-coded-shortcuts

jnsebgosselin:
@ccordoba12 good news. I have found a way to work around the limitation I was talking about in https://github.com/spyder-ide/spyder/pull/7929#issuecomment-424063897

I still have to find a solution for builtin shortcut though.

ccordoba12:
> I have found a way to work around the limitation I was talking about

Great!

> I still have to find a solution for builtin shortcut though.

What do you mean by this?

jnsebgosselin:
> > I still have to find a solution for builtin shortcut though.
> 
> What do you mean by this?

I mean, the fact that key sequences associated to builtin editing shortcut like `Ctrl+X` for `Cut` and `Ctrl+V` for `Paste` are eaten by any text editing widget when they have focus and prevent these key bindings to work properly and consistently when assigned to a global shortcut.

But I think I have found a solution for this problem also... So basically, I think I have found a solution for both problems that were preventing me to fully support Emacs keybindings application wise.

ccordoba12:
> So basically, I think I have found a solution for both problems that were preventing me to fully support Emacs keybindings application wise.

Nice! But don't you think that you should implement that in another PR?

jnsebgosselin:
> 
> 
> > So basically, I think I have found a solution for both problems that were preventing me to fully support Emacs keybindings application wise.
> 
> Nice! But don't you think that you should implement that in another PR?

This is a good idea. Would you like that I focus only on the `ShortcutEditor` in this PR and try to fix the aforementioned issues in another PR?

ccordoba12:
> Would you like that I focus only on the ShortcutEditor in this PR and try to fix the aforementioned issues in another PR?

Yep, that's my suggestion. That way it'd be easier for me to review and merge your work.

ccordoba12:
This needs a merge with master now.

jnsebgosselin:
Oh my gosh, it is working... I just need to write tests now!

I have another question though for you @ccordoba12. I think the situation you were referring to in #7854 is an old bug in Qt that has been resolved since, because I can now set shortcuts to `Shift + <any key>` and it is working.

ccordoba12:
> Oh my gosh, it is working... I just need to write tests now!

These are great news! Yay!

> I think the situation you were referring to in #7854 is an old bug in Qt that has been resolved since

Ok, that's good to know too!

jnsebgosselin:
Do I remove the restriction in the `ShortcutEditor` for `Shift+<any key>, since it appears to be working now?

spyder/preferences/tests/test_shorcuts.py
```diff
@@ -9,29 +9,207 @@
...
+    shorcut_table, qtbot = shorcut_table_bot
+    shortcuts = shorcut_table.source_model.shortcuts
+
+    def _create_bot(context, name):
+        sequence = CONF.get('shortcuts', "{}/{}".format(context, name))
+        shortcut_editor = ShortcutEditor(
+            shorcut_table, context, name, sequence, shortcuts)
+        qtbot.addWidget(shortcut_editor)
+        shortcut_editor.show()
+        return shortcut_editor, qtbot
```
ccordoba12: I don't like this style of fixtures: we should make our fixtures to return the single object we want to test and not the tuple of `widget, qtbot`. Instead `qtbot` needs to be passed as an additional fixture to every test, as I point out in a test below.

So please change this.

spyder/preferences/tests/test_shorcuts.py
```diff
@@ -9,29 +9,207 @@
...
+    return _create_bot
 
 
-@pytest.mark.skipif(not os.name == 'nt', reason="It segfaults too much on Linux")
-def test_shortcuts(qtbot):
+# ---- Tests ShortcutsTable
+@pytest.mark.skipif(
+    sys.platform.startswith('linux') and os.environ.get('CI') is not None,
+    reason="It fails on Linux due to the lack of a proper X server.")
+def test_shortcuts(shorcut_table_bot):
```
ccordoba12: This needs to be

```python
def test_shortcuts(shorcut_table_bot, qtbot)
```

and the same for all other tests present in this file.

ccordoba12: It makes tests easier to read because you can see all fixtures it depends on.

spyder/preferences/shortcuts.py
```diff
@@ -16,62 +16,94 @@
...
 
-
-MODIFIERS = {Qt.Key_Shift: Qt.SHIFT,
-             Qt.Key_Control: Qt.CTRL,
-             Qt.Key_Alt: Qt.ALT,
-             Qt.Key_Meta: Qt.META}
-
 # Valid shortcut keys
-SINGLE_KEYS = ["F{}".format(_i) for _i in range(1, 36)] + ["Delete", "Escape"]
-KEYSTRINGS = ["Tab", "Backtab", "Backspace", "Return", "Enter",
```
ccordoba12: Why did you remove all these KEYSTRINGS? Just curious, this was written by @goanpeca, not me.

spyder/preferences/tests/test_shorcuts.py
jnsebgosselin: ok, I understand. I will change that.

ccordoba12: Great, thanks!

spyder/preferences/shortcuts.py
jnsebgosselin: This is a good question.

Validating each `QKeyEvent.key()` in the `keyPressEvent` to a set of harcoded values was not working very well for international keyboards. For example,  doing `Ctrl+Alt+P` on my keyboard results in the `Qt.Key_Odiaeresis` key (Ö), which was obviously not in our list of predefined hardcoded keys. So, most of the `Ctrl+Alt` and `Ctrl+Alt+Shift` key sequences were not working with my keyboard with the previous Shortcut Editor.

So to solve that issue, I created the `ShortcutTranslator` that is used to convert `QKeyEvent` into `QKeySequence`. With this, pressing `Ctrl+Alt+Shift+P` results correctly in the key sequence `Ctrl+Alt+Shift+P` and not `Ctrl+Alt+Shift+Ö`.

In case an invalid key is still entered for some reason (due for example to the bug mentioned in https://github.com/spyder-ide/spyder/pull/7929#issuecomment-425202097), I check also that all characters of the key sequences are ASCII in `check_ascii`.

So basically, the harcoded KEYSTRINGS were not needed anymore, so that is why I removed them. That being said, it is quite possible that there is some corner cases that my new approach is missing. Worst scenario if this happens is that the new custom shortcut will simply do nothing.

ccordoba12: Ok, thanks a lot the detailed explanation!

PR #7954, opened 2018-09-25
PR: Improve how keyboard shortcuts are handled in the Editor (Take 2)

### LINKED ISSUE #7883: GNU Emacs Style Key Sequences do not work in master AND shortcut conflicts

jnsebgosselin:
### Issue(s) Resolved

Fixes #7883
Blocking #7929

## Description of Changes

In this PR, we propose to revert some of the changes made in PR #7768 and to go back to use `QShortcut` to manage local keyboard shortcuts in the Editor.

The reason is that I found a way to fix all the problems we had with builtin shortcuts in the editor using the Qt Framework and the same architecture that is used everywhere to manage shortcuts in Spyder.

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -44,7 +44,7 @@ from spyder_kernels.utils.dochelpers import getobj
 
 # Local imports
 from spyder.config.base import get_conf_path, _, DEBUG
-from spyder.config.gui import get_shortcut, iter_shortcuts
+from spyder.config.gui import get_shortcut, config_shortcut
 from spyder.config.main import (CONF, RUN_CELL_SHORTCUT,
                                 RUN_CELL_AND_ADVANCE_SHORTCUT)
 from spyder.py3compat import to_text_string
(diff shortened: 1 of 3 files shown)


jnsebgosselin:
@ccordoba12 All that was needed to fix our problems with builtin shortcuts in the Editor was to override the `event`  method of the `CodeEditor` like this :

```python
    def event(self, event):
        """Qt method override."""
        if event.type() == QEvent.ShortcutOverride:
            event.ignore()
            return False
        else:
            return super(CodeEditor, self).event(event)
```

So I did that and reverted back to using `QShortcut` to manage local shortcuts in the `Editor`. I can't believed I worked all these hours to end up reverting almost everything for such a simple fix...

The only downside is that the tests I've written won't work on Travis anymore since.

ccordoba12:
>  All that was needed to fix our problems with builtin shortcuts in the Editor was to override the event method of the CodeEditor like this

That was simple. Thanks!

PR #7962, opened 2018-09-28
PR: Improvement to navigation and file switching in the Outline Explorer

### LINKED ISSUE #7214: Outline pane, clicking on file name should move to start of file
## Feature Suggestion  Description

Clicking on a leaf node in the outline pane tree takes you to the start of the relevant bit of code, but clicking on the root node (the filename) does nothing. It would be better if it took you to the start of the file.

This would be a useful feature for when one realizes halfway through adding some code that some extra modules need to be imported at the start of the file - being able to quickly get there using the outline view would be handy. It would also make the root node behave in a more consistent fashion to the leaf nodes.

## Package Versions

* Spyder version: 3.2.8 
* Python version: 3.6.4
* Qt version: 5.6.2
* PyQt5 version: 5.6
* Operating system: Linux 4.13.0-43-generic

jnsebgosselin:
### Issue(s) Resolved

Fixes #7214

### Pull Request Checklist

## Description of Changes

First, when `Show all Files` is selected in the `Outline explorer`, the root item will not automatically expand and collapse when the current `Editor` change. IMO, this change brings 3 UX improvements:
- The Outline Explorer can now be used more easily as a "vertical tab bar".
- Double clicking a root file item will now expand/contract that item, like it does for the other tree items.
- This allows to keep more than one file expanded at the same time, which is very convenient to navigate between elements that are not saved in a same file.

outline_navigation

Second, if a root file item is clicked and that item corresponds to the current Editor, the cursor will be moved at the start of the file. Double-clicking a root file item, whether this item correspond to the current Editor or not, will expand/contract the item and move the cursor at the beginning of the file. This behavior is also coherent with what happens when double-clicking on a normal tree item.

Note that, the user can always go back to the previous position of the cursor with `Source > Previous cursor position` or by doing the `previous cursor position` keyboard shortcut (which is `Ctrl+Alt+Left`  by default).

outline_navigation_top_file

### DIFF
--- a/spyder/plugins/editor/widgets/tests/test_editor_and_outline.py
+++ b/spyder/plugins/editor/widgets/tests/test_editor_and_outline.py
@@ -97,7 +97,7 @@ def test_load_files(empty_editor_bot, python_files):
     editorstack.tabs.currentIndex() == 0
 
     results = [item.text(0) for item in treewidget.get_visible_items()]
-    assert results == ['foo1.py', 'foo']
+    assert results == ['foo1.py']
 
     # Load the second file and assert the content of the outline explorer tree.
     editorstack.load(python_files[1])
@@ -105,7 +105,7 @@ def test_load_files(empty_editor_bot, python_files):
     editorstack.tabs.currentIndex() == 1
 
     results = [item.text(0) for item in treewidget.get_visible_items()]
-    assert results == ['foo1.py', 'foo2.py', '---- a comment']
+    assert results == ['foo1.py', 'foo2.py']
 
 
 def test_close_editor(editor_bot):
@@ -138,7 +138,7 @@ def test_close_a_file(editor_bot):
     # tree has been updated.
     editorstack.close_file(index=1)
     results = [item.text(0) for item in treewidget.get_visible_items()]
-    assert results == ['foo1.py', 'foo']
+    assert results == ['foo1.py']
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


ccordoba12:
This is really nice! Thanks!

jnsebgosselin:
Thank you very much @CAM-Gerlach for taking the time to test this!

I think that now we need to add a way to better highlight the currently focused file in the Outline Explorer though... Maybe making the root file items of the currently selected file in bold text?

Also, I would like to have the order of the files in the Outline Explorer to be synced with the tabbar of the Editorstack instead of sorting the files alphabetically.

CAM-Gerlach:
Great followups @jnsebgosselin 

> I think that now we need to add a way to better highlight the currently focused file in the Outline Explorer though... Maybe making the root file items of the currently selected file in bold text?

Sure, unless Carlos has a better idea. I'd think only the file name would need to be bolded, but I'd have to see how it looks.

> Also, I would like to have the order of the files in the Outline Explorer to be synced with the tabbar of the Editorstack instead of sorting the files alphabetically.

Yes, absolutely.

ccordoba12:
> I think that now we need to add a way to better highlight the currently focused file in the Outline Explorer though... Maybe making the root file items of the currently selected file in bold text?

Let's leave that for another PR, given that this one is simple and directed to fix one thing and one thing only.

My recommendation is to highlight the current file and class/method/cell/function by making the selected element to have a different background color, like VSCode does:

seleccion_015

PR #7968, opened 2018-09-30
PR: Sync Outline Explorer at startup and preserve the file order in the tabbar

### LINKED ISSUE #7338: Outline explorer not synchronized at startup
A small GUI bug.

When starting spyder with several files open, and the "structure panel" switched on. The GUI ends up, just after booting, with first file tab and apparently last parsed structure (from last tab), hence leading to a non synchronization between the 2. Switching twice file tabs resynchronizes everything hopefully.

(current mainstream)

jnsebgosselin:
### Issue(s) Resolved

Fixes #7338

### Pull Request Checklist

### Before :

image

### After :

image

### DIFF
--- a/spyder/plugins/editor/plugin.py
+++ b/spyder/plugins/editor/plugin.py
@@ -2948,13 +2948,15 @@ class Editor(SpyderPluginWidget):
                 index = filenames.index(cfname)
                 # First we load the last focused file.
                 self.load(filenames[index], goto=clines[index], set_focus=True)
-                # Then we load the files located before the last focused file
-                # in the tabbar.
+                # Then we load the files located to the left of the last
+                # focused file in the tabbar, while keeping the focus on
+                # the last focused file.
                 if index > 0:
                     self.load(filenames[index::-1], goto=clines[index::-1],
                               set_focus=False, add_where='start')
-                # Finally we load the files located after the last focused file
-                # in the tabbar.
+                # Finally we load the files located to the right of the last
+                # focused file in the tabbar, while keeping the focus on
+                # the last focused file.
                 if index < (len(filenames) - 1):
                     self.load(filenames[index+1:], goto=clines[index:],
                               set_focus=False, add_where='end')
--- a/spyder/plugins/editor/plugin.py
+++ b/spyder/plugins/editor/plugin.py
@@ -2063,7 +2063,8 @@ class Editor(SpyderPluginWidget):
     @Slot(str, int, str)
     @Slot(str, int, str, object)
     def load(self, filenames=None, goto=None, word='',
-             editorwindow=None, processevents=True, start_column=None):
+             editorwindow=None, processevents=True, start_column=None,
+             set_focus=True, add_where='end'):
         """
         Load a text file
         editorwindow: load in this editorwindow (useful when clicking on
@@ -2155,10 +2156,7 @@ class Editor(SpyderPluginWidget):
 
         for index, filename in enumerate(filenames):
             # -- Do not open an already opened file
-            if index == 0:  # this is the last file focused in previous session
-                focus = True
-            else:
-                focus = False
+            focus = set_focus and index == 0
             current_editor = self.set_current_filename(filename,
                                                        editorwindow,
                                                        focus=focus)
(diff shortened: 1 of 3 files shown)


ccordoba12:
@jnsebgosselin, please fix the failure in Travis, it was introduced by a change done in this PR.

spyder/plugins/editor/plugin.py
```diff
@@ -2959,39 +2973,6 @@ def setup_open_files(self):
             self.__load_temp_file()
         self.set_create_new_file_if_empty(True)
 
-    def reorder_filenames(self, filenames):
```
ccordoba12: This method seems to cover more things than what your current code does (e.g. it seems to handle splitted editor stacks). Please explain why you're removing it.

spyder/plugins/editor/tests/test_plugin.py
```diff
@@ -63,36 +80,40 @@ def test_basic_initialization(setup_editor):
...
-        (['/file1.py', '/file2.py', '/file3.py', '/file4.py'],
-         ['/file2.py', '/file1.py', '/file3.py', '/file4.py'],
-         [(False, '/file2.py', [2, 4, 8, 12])],
-         [(False, '/file2.py', [4, 2, 8, 12])]),
-        (['/file1.py', '/file2.py', '/file3.py', '/file4.py'],
-         ['/file1.py', '/file2.py', '/file3.py', '/file4.py'],
-         None,
-         None, )
-    ])
-def test_reorder_filenames(setup_editor, expected_filenames, filenames,
```
ccordoba12: This test seems more comprehensive than the test you're replacing it with. Please explain why you're replacing it.

spyder/plugins/editor/plugin.py
jnsebgosselin: The **only** thing this method does, is to load the `splitsettings`, take the filename that had focus last in the first `EditorStack` (`index = 0`), move it to position 0, and save the modified `splitsettings`. 

The settings for the additional splitted editor stacks (`index > 0`) are not modified by this.

This PR proposes to preserve the file order that was saved when Spyder was last closed and to stop reordering files during the setup. Therefore, with the proposed changes, this method is not used anywhere else in the code. I can keep it however if you prefer in case it is needed for something else at some point.

ccordoba12: But does focus jumps from one file to another while the Editor is opening them? That's my main concern here.

spyder/plugins/editor/tests/test_plugin.py
jnsebgosselin: I'm not actually replacing it. It is more like I'm removing `test_reorder_filenames` that was checking only the output of the `reorder_filenames` method and I'm creating a new test that is checking the results of `setup_open_files` instead. So in fact, I think the new test is more comprehensive because it actually checks the resulting order of the files in the `Editorstack` and also check that the right file is focused.

The only thing I do not cover compared to the old test, is when `splitsettings` is `None`, because unless I'm mistaken and I can be, this case is not supposed to happen at this point in the setup of Spyder, so there is no point in testing it.

I simplified the `@pytest.mark.parametrize` because some of the arguments were not needed and there was also duplication of information. Maybe that is why this test looks less comprehensive?

spyder/plugins/editor/plugin.py
jnsebgosselin: No it does not. I'll post an animation to show you what it is doing, one sec.

jnsebgosselin:
setup_files

spyder/plugins/editor/plugin.py
ccordoba12: Ok, that's good. Preserving file order and loading the last file with focus first is ice, UX-wise.

jnsebgosselin: The Outline Explorer is jumping around though, but it was doing that before.At least with this PR, it ends up showing the tree of the right file at the end of the setup. I can fix this in another PR if you think it should.

CAM-Gerlach: Yeah, that's fine and does look a little nicer at startup as well, although since its only transient its not the biggest concern for me personally.

Of course, the outline explorer is still ordered alphabetically, rather than the order of the tabs in the Editor. Presumably out of scope for now, but if/when you make that change, if would be nice to have the option to either alphabetize it or make it follow the order of the editor, since both could potentially be useful depending on one's workflow.

Also, to note, the outline explorer only show Python files, so it isn't a true file switcher for more general cases. Perhaps, as a longer-term idea, we could have an option to show them (just the files without the outlines, of course).

jnsebgosselin: > Yeah, that's fine and does look a little nicer at startup as well, although since its only transient its not the biggest concern for me personally.

The most important part of this change is not the transient effect, but the fact that we do not need to reorder the tabs during setup.

spyder/plugins/editor/tests/test_plugin.py
ccordoba12: Ok, thanks for the explanation.

jnsebgosselin: No problem. Thanks for the review, it is always appreciated.

PR #8021, opened 2018-10-06
PR: Fix tests by pinning testpath and switching to Python 3.5 on Windows

ccordoba12:
Conda package for `testpath` **0.4.1** contains wrong METADATA and that's breaking pip.

Easy way to reproduce:

    conda create -n pip-break python=3.6 testpath
    conda activate pip-break
    pip install ipython

Then see the last command generate a long traceback.

Pinging @mingwandroid and @jjhelmus about this. This requires an urgent fix.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,8 +18,7 @@ environment:
 
   matrix:
     - PYTHON_VERSION: "2.7"
-    #- PYTHON_VERSION: "3.5"
-    - PYTHON_VERSION: "3.6"
+    - PYTHON_VERSION: "3.5"
 
 platform:
   -x64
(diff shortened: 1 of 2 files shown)


jjhelmus:
The `testpath` conda package was built with a version of flit which does not include takluyver/flit#212.  The result is that pkg_resources is not able to parse the Required-Dist line in the METADATA file.  I'll update the `flit` conda package and start new builds of `testpath`.

jjhelmus:
testpath 0.4.1 (build 1) packages with a correct METADATA file and 0.4.2 packages are now available.  I tested that pip works with both.

ccordoba12:
Thanks @jjhelmus for the quick response!

PR #8059, opened 2018-10-11
Variable Explorer: Use distinctive background for missings in DataFrames

### LINKED ISSUE #8037: DataFrames in Variable Explorer: Missings should have distinctive background color
### Issue Report Checklist

    * [ ] Restarted Spyder

## Problem Description
When opening a Pandas DataFrame with missing values in the Variable Explorer, the missings are displayed with red cell background, which is similar to the red background of the minimal values. So the missings are difficult to spot.

I suggest using a white background for the missings instead.

1. `df=pd.DataFrame([1,None,5,9])`
2. Open df in the Variable Explorer.

I see
screenshot-missings-in-variable-explorer

I expect the nan field to have a visually distinctive color, e.g. white.

## Versions

* Spyder version: 3.3.1
* Python version: 3.7.0 64bits
* Qt version: Qt 5.9.3
* PyQt version: 5.9.2
* Operating System name/version: Win7 64-bit

(everything from WinPython64-3.7.0.2)

### Dependencies

Available on request.

jondo:
### Pull Request Checklist

* [x] Ensured this pull request hasn't eliminated unrelated blank lines/spaces,
* [x] Wrote at least one-line docstrings following PEP 257for any new functions

## Description of Changes

When opening a Pandas DataFrame with missing values in the Variable Explorer, the missings are displayed with red cell background, which is similar to the red background of the minimal values. So the missings are difficult to spot, see #8037.

I suggest using a white background for the missings instead:
white-missings

### Issue(s) Resolved

Fixes #8037

### DIFF
--- a/spyder/plugins/variableexplorer/widgets/dataframeeditor.py
+++ b/spyder/plugins/variableexplorer/widgets/dataframeeditor.py
@@ -45,7 +45,7 @@ from qtpy.QtWidgets import (QApplication, QCheckBox, QDialog, QGridLayout,
                             QMessageBox, QPushButton, QTableView,
                             QScrollBar, QTableWidget, QFrame,
                             QItemDelegate)
-from pandas import DataFrame, Index, Series
+from pandas import DataFrame, Index, Series, isna
 try:
     from pandas._libs.tslib import OutOfBoundsDatetime
 except ImportError:  # For pandas version < 0.20
@@ -272,7 +272,7 @@ class DataFrameModel(QAbstractTableModel):
         if not self.bgcolor_enabled:
             return
         value = self.get_value(index.row(), column)
-        if self.max_min_col[column] is None:
+        if self.max_min_col[column] is None or isna(value):
             color = QColor(BACKGROUND_NONNUMBER_COLOR)
             if is_text_string(value):
                 color.setAlphaF(BACKGROUND_STRING_ALPHA)
--- a/spyder/plugins/variableexplorer/widgets/tests/test_dataframeeditor.py
+++ b/spyder/plugins/variableexplorer/widgets/tests/test_dataframeeditor.py
@@ -276,6 +276,27 @@ def test_dataframemodel_get_bgcolor_with_object():
     a = dataframeeditor.BACKGROUND_MISC_ALPHA
     assert colorclose(bgcolor(dfm, 0, 0), (h, s, v, a))
 
+
+def test_dataframemodel_get_bgcolor_with_missings():
+    """
+    Test that df bg colors are correct for missing values of various types.
+
+    The types `bool`, `object`, `datetime`, and `timedelta` are omitted,
+    because missings have no different background there yet.
+    """
+    df = DataFrame({'int': [1, None], 'float': [.1, None],
+                    'complex': [1j, None], 'string': ['a', None]})
+    df['category'] = df['string'].astype('category')
+    dfm = DataFrameModel(df)
+    h, s, v, __ = QColor(dataframeeditor.BACKGROUND_NONNUMBER_COLOR).getHsvF()
+    alpha = dataframeeditor.BACKGROUND_MISC_ALPHA
+    for idx, column in enumerate(df.columns):
+        assert not colorclose(bgcolor(dfm, 0, idx), (h, s, v, alpha)), \
+            'Wrong bg color for value of type ' + column
+        assert colorclose(bgcolor(dfm, 1, idx), (h, s, v, alpha)), \
+            'Wrong bg color for missing of type ' + column
+
+
 def test_dataframemodel_with_format_percent_d_and_nan():
     """
     Test DataFrameModel with format `%d` and dataframe containing NaN


ccordoba12:
@jondo, thanks a lot for this! Please take a look at our tests because they are failing due to your change.

spyder/plugins/variableexplorer/widgets/dataframeeditor.py
```diff
@@ -271,6 +271,8 @@ def get_bgcolor(self, index):
         if not self.bgcolor_enabled:
             return
         value = self.get_value(index.row(), column)
+        if np.isnan(value):
+            return Qt.white
```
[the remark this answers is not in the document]
CAM-Gerlach: However, this isn't really necessary at all, since you can just add the previous line as a condition to the if statement directly below, which does exactly what you want anyway.

jondo:
@CAM-Gerlach Thank you for your feedback!

ccordoba12:
> I can't create a unit test for now because I have a problem running the tests

Please only run the tests of the Dataframe Editor. You don't need to run all our tests for this.

jondo:
Sorry, I have not worked with pytest before, and I could not find out how to do this yet. Can you give me a hand?

jondo:
@CAM-Gerlach, the requested test is now ready for review. The types bool, object, and timedelta have the same gray background in normal cells, so the test is not useful there. Is the patch good enough for now, or do you want the missings in such columns to have a different background color?

CAM-Gerlach:
Your builds are failing due to an unrelated issue that we can fix momentarily (you'll need to ``rebase`` to grab the changes once we do). I'll have a review for you in a moment; sorry for the delay.

ccordoba12:
@CAM-Gerlach, is this ready?

CAM-Gerlach:
@ccordoba12 Not quite...

ccordoba12:
@CAM-Gerlach, please apply your suggested changes and don't bother contributors with pep8 stuff in our tests. We're not following pep8 in a lot of places in our tests.

ccordoba12:
> please apply your suggested changes

Sorry, I just saw only PR authors can accept or reject the changes.

jondo:
Thanks for the review. I will care for the changes by myself, if you don't mind.

Am 27. Oktober 2018 20:28:24 MESZ schrieb Carlos Cordoba <notifications@github.com>:

ccordoba12:
Sure, no problem. I'm still unsure how the *Suggested changes* feature works.

CAM-Gerlach:
> Thanks for the review. I will care for the changes by myself, if you don't mind.

@jondo Its your time/effort, and your call..

@ccordoba12 See elsewhere for my reply, to avoid getting OT.

jondo:
OK, here's my update.

Boy, 79 lines is narrow :-) Still good to think of pep8.
I did my edits locally, so no test of the _Suggested changes_ feature.

W.r.t. the (now removed) assertion message, I used it because otherwise I see no indication which column failed. Here is an example where the assertion message would have told me that the 'int' column was failing:

```
$ pytest spyder/plugins/variableexplorer/widgets/tests/test_dataframeeditor.py::test_dataframemodel_get_bgcolor_with_missings
==================================================================== test session starts =====================================================================
platform linux -- Python 3.6.6, pytest-3.8.2, py-1.7.0, pluggy-0.7.1
PyQt5 5.11.2 -- Qt runtime 5.11.1 -- Qt compiled 5.11.1
rootdir: /home/robert/bin/spyder, inifile:
plugins: timeout-1.3.2, qt-3.2.1, mock-1.10.0, cov-2.6.0, flaky-3.4.0
collected 1 item                                                                                                                                             

spyder/plugins/variableexplorer/widgets/tests/test_dataframeeditor.py F                                                                                [100%]

========================================================================== FAILURES ==========================================================================
_______________________________________________________ test_dataframemodel_get_bgcolor_with_missings ________________________________________________________

    def test_dataframemodel_get_bgcolor_with_missings():
        """
        Test that df bg colors are correct for missing values of various types.
    
        The types `bool`, `object`, and `timedelta` are omitted,
        because missings have no different background there yet.
        """
        df = DataFrame({'int': [1, 2], 'float': [.1, None], 'complex': [1j, None],
                        'string': ['a', None],
                        'datetime': [datetime(2000, 1, 1), None]})
        df['category'] = df['string'].astype('category')
        dfm = DataFrameModel(df)
        h, s, v, __ = QColor(dataframeeditor.BACKGROUND_NONNUMBER_COLOR).getHsvF()
        alpha = dataframeeditor.BACKGROUND_MISC_ALPHA
        for column_idx in range(len(df.columns)):
>           assert colorclose(bgcolor(dfm, 1, column_idx), (h, s, v, alpha))
E           assert False
E            +  where False = colorclose(<PyQt5.QtGui.QColor object at 0x7fb981d1c7b8>, (-1.0, 0.0, 0.7529411764705882, 0.3))
E            +    where <PyQt5.QtGui.QColor object at 0x7fb981d1c7b8> = bgcolor(<spyder.plugins.variableexplorer.widgets.dataframeeditor.DataFrameModel object at 0x7fb96e781af8>, 1, 0)

spyder/plugins/variableexplorer/widgets/tests/test_dataframeeditor.py:295: AssertionError
===Flaky Test Report===


===End Flaky Test Report===
================================================================== 1 failed in 1.77 seconds ==================================================================
```

By the way, `datetime(1000, 1, 1)` would have worked. The docs [say](https://docs.python.org/3/library/datetime.html) that the date is:
> An idealized naive date, assuming the current Gregorian calendar always was, and always will be, in effect.

CAM-Gerlach:
> By the way, `datetime(1000, 1, 1)` would have worked

I wasn't entirely clear in my comment, but it will either get coerced to a pandas-native ``datetime64[ns]`` which is restricted to only the modern era, or it will be treated as an ``object`` column.

Interestingly, in my testing in Spyder 3 and 4 (without this PR applied at least), the full timestamp column displays in the grey "object" color while datetimes outside the acceptable range for "timestamp" and ``None``s display in the BG color (white/black), while the rest display in the "object" color.

image

CAM-Gerlach:
> Boy, 79 lines is narrow :-) Still good to think of pep8.

Yeah, its not ideal in some ways but it avoids text getting cut off when using 2+ columns in Spyder and other editors or on portrait-format screens.

> W.r.t. the (now removed) assertion message, I used it because otherwise I see no indication which column failed. Here is an example where the assertion message would have told me that the 'int' column was failing:

Technically, it does, since it shows that ``column_idx`` is ``0``, which is the ``int`` column, but it does take a more careful look to find it. I prefer the full output, but its up to @ccordoba12 . If we do go with the message instead, it needs to say much more than just the column name so people know what's actually going on, e.g. "Column " + str(column) + " with value " str(df.iloc[1, column+idx]) + " does not have the specified color for None/NA values".

jondo:
Hmm ... when I re-introduce the assertion message
```python
    for idx, column in enumerate(df.columns):
        assert colorclose(bgcolor(dfm, 1, idx), (h, s, v, alpha)), \
               'Wrong bg color for type ' + column
```
, an error message would look like
```
E           AssertionError: Wrong bg color for type int
E           assert False
E            +  where False = colorclose(<PyQt5.QtGui.QColor object at 0x7ff5f527a7b8>, (-1.0, 0.0, 0.7529411764705882, 0.3))
E            +    where <PyQt5.QtGui.QColor object at 0x7ff5f527a7b8> = bgcolor(<spyder.plugins.variableexplorer.widgets.dataframeeditor.DataFrameModel object at 0x7ff5e1cb8af8>, 1, 0)
```
,  which still contains the introspection info.
Wouldn't that be a little more useful?

CAM-Gerlach:
> Having non-gray background colors for normal bool, object, timedelta, and datetime columns should be a separate issue, please.

Yup, you're right. I'll go about making one when I get the chance. Thanks!

> ``assert colorclose(bgcolor(dfm, 1, idx), (h, s, v, alpha)), \``

Modern convention is to use explicit continuation with parents, FYI:

```python
    assert (colorclose(bgcolor(dfm, 1, idx), (h, s, v, alpha)),
            'Wrong bg color for type ' + column)
```

> Wouldn't that be a little more useful?

It certainly sounds like it would be. @ccordoba12 opinion?

ccordoba12:
Sure, it seems more useful.

jondo:
Here's the re-added assertion message.
Thanks for the hint with the implicit continuation with parentheses, which I have even found described in PEP 8. It can't be used here, however, because otherwise the `assert` keyword will get a tuple as first argument. The tuple evaluates to `True`, so the assert never fails.

jondo:
I am now also checking that non-missing values don't have gray background, to make sure that the gray is distinctive (see the PR title).

CAM-Gerlach:
> It can't be used here, however, because otherwise the assert keyword will get a tuple as first argument.

I was wondering if that would be the case; makes sense.

ccordoba12:
@CAM-Gerlach, please approve when you consider @jondo's work is ready.

spyder/plugins/variableexplorer/widgets/tests/test_dataframeeditor.py
```diff
@@ -276,6 +276,27 @@ def test_dataframemodel_get_bgcolor_with_object():
...
+    """
+    df = DataFrame({'int': [1, None], 'float': [.1, None],
+                    'complex': [1j, None], 'string': ['a', None]})
+    df['category'] = df['string'].astype('category')
+    dfm = DataFrameModel(df)
+    h, s, v, __ = QColor(dataframeeditor.BACKGROUND_NONNUMBER_COLOR).getHsvF()
+    alpha = dataframeeditor.BACKGROUND_MISC_ALPHA
+    for idx, column in enumerate(df.columns):
+        assert not colorclose(bgcolor(dfm, 0, idx), (h, s, v, alpha)), \
+               'Wrong bg color for value of type ' + column
```
CAM-Gerlach: A hanging indent should be four spaces; Spyder makes it 8 spaces but that's a bug.

spyder/plugins/variableexplorer/widgets/tests/test_dataframeeditor.py
```diff
@@ -276,6 +276,27 @@ def test_dataframemodel_get_bgcolor_with_object():
...
+        assert colorclose(bgcolor(dfm, 1, idx), (h, s, v, alpha)), \
+               'Wrong bg color for missing of type ' + column
```
CAM-Gerlach: Same as above

spyder/plugins/variableexplorer/widgets/tests/test_dataframeeditor.py
jondo: Even worse, I had aligned the two assert arguments :-P  It's changed now.

PR #8123, opened 2018-10-22
PR: Improve usability of runtests locally and test output on the CIs

CAM-Gerlach:
### Pull Request Checklist

* [x] Ensured this pull request hasn't eliminated unrelated blank lines/spaces,
* [x] Wrote at least one-line docstrings following PEP 257 for any new functions

## Description of Changes

This PR does a few things to improve usability of both runtests locally (that have been bothering me and perhaps others for quite some time), as well as the CI reports for users developing both Spyder 3 and Spyder 4, and does a little opportunistic cleanup/refacotring of code that hasn't been touched between ``3.x`` and ``master``.

Namely, it:

* Adds the ability to pass any arbitrary additional pytest options of the users choosing through to pytest with ``argparse``, as well as some user-friendly help text and error handling
* Makes the ``-x`` option the default only on the CIs, so local users can see full test output without hacking ``runtests.py`` and can still pass  ``-x`` or Ctrl-C the test cycle manually if they so desire
* Removes the remaining % file coverage by default, as it can easily be re-enabled locally in the case the user really desires it for every file, and is already displayed in far more accessible form via two different dedicated services when run the CIs and otherwise just adds extra clutter to the logs
* Does some minor opportunistic cleanup/refactoring of non-conflicting ``3.x``/``master`` code

It was made against ``3.x``since both branches are still being actively developed, and thus the local and CI test output and runtests API would be inconsistent between them and have extra clutter. There should only be one small conflict—the Spyder profiler coverage stuff was removed, which was already removed in ``master``).

### DIFF
--- a/runtests.py
+++ b/runtests.py
@@ -10,7 +10,6 @@ Script for running Spyder tests programmatically.
 
 # Standard library imports
 import os
-import sys
 import argparse
 
 # Third party imports
@@ -19,10 +18,9 @@ import pytest
 
 # To run our slow tests only in our CIs
 RUN_CI = os.environ.get('CI', None) is not None
-run_slow = RUN_CI
 
 
-def main(run_slow=run_slow, extra_args=None):
+def main(extra_args=None):
     """
     Run pytest tests for Spyder.
     """
@@ -33,11 +31,9 @@ def main(run_slow=run_slow, extra_args=None):
                    '--durations=10']
 
     if RUN_CI:
-        pytest_args.append('-x')
+        pytest_args += ['-x', '--run-slow']
     elif extra_args:
         pytest_args += extra_args
-    if run_slow and '--run-slow' not in pytest_args:
-        pytest_args.append('--run-slow')
 
     print("Pytest Arguments: " + str(pytest_args))
     errno = pytest.main(pytest_args)
@@ -52,10 +48,7 @@ def main(run_slow=run_slow, extra_args=None):
 if __name__ == '__main__':
     test_parser = argparse.ArgumentParser(
         usage='python runtests.py [--run-slow] [-- pytest_args]')
-    test_parser.add_argument('--run-slow', action='store_true',
-                             default=run_slow,
-                             help='Run the slow (mainly GUI/functional) tests')
     test_parser.add_argument('pytest_args', nargs='*',
                              help="Args to pass to pytest")
     test_args = test_parser.parse_args()
-    main(run_slow=test_args.run_slow, extra_args=test_args.pytest_args)
+    main(extra_args=test_args.pytest_args)
--- a/runtests.py
+++ b/runtests.py
@@ -5,7 +5,7 @@
 #
 
 """
-File for running tests programmatically.
+Script for running Spyder tests programmatically.
 """
 
 # Standard library imports
@@ -13,7 +13,6 @@ import os
 import sys
 
 # Third party imports
-import qtpy  # to ensure that Qt4 uses API v2
 import pytest
 
 
@@ -22,9 +21,10 @@ run_slow = False
 if os.environ.get('CI', None) is not None or '--run-slow' in sys.argv:
     run_slow = True
 
+
 def main():
     """
-    Run pytest tests.
+    Run pytest tests for Spyder.
     """
     pytest_args = ['spyder',
                    'spyder_profiler',
@@ -38,10 +38,9 @@ def main():
 
     errno = pytest.main(pytest_args)
 
-    # sys.exit doesn't work here because some things could be running
-    # in the background (e.g. closing the main window) when this point
-    # is reached. And if that's the case, sys.exit does't stop the
-    # script (as you would expected).
+    # sys.exit doesn't work here because some things could be running in the
+    # background (e.g. closing the main window) when this point is reached.
+    # If that's the case, sys.exit doesn't stop the script as you would expect.
     if errno != 0:
         raise SystemExit(errno)
 
--- a/runtests.py
+++ b/runtests.py
@@ -11,6 +11,7 @@ Script for running Spyder tests programmatically.
 # Standard library imports
 import os
 import sys
+import argparse
 
 # Third party imports
 import pytest
@@ -18,10 +19,10 @@ import pytest
 
 # To run our slow tests only in our CIs
 RUN_CI = os.environ.get('CI', None) is not None
-run_slow = RUN_CI or '--run-slow' in sys.argv
+run_slow = RUN_CI
 
 
-def main():
+def main(run_slow=run_slow, extra_args=None):
     """
     Run pytest tests for Spyder.
     """
@@ -33,9 +34,12 @@ def main():
 
     if RUN_CI:
         pytest_args.append('-x')
-    if run_slow:
+    elif extra_args:
+        pytest_args += extra_args
+    if run_slow and '--run-slow' not in pytest_args:
         pytest_args.append('--run-slow')
 
+    print("Pytest Arguments: " + str(pytest_args))
     errno = pytest.main(pytest_args)
 
     # sys.exit doesn't work here because some things could be running in the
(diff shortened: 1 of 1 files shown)


ccordoba12:
> the Spyder profiler coverage stuff was removed, which was already removed in master

It was removed in master because all plugins are now inside `spyder.plugins`, but that's not the case for our 3.x branch because `spyder_profiler` is a package on its own and it lives out of the `spyder` one. So please restore coverage for it.

runtests.py
```diff
@@ -5,49 +5,57 @@
...
     """
     pytest_args = ['spyder',
                    'spyder_profiler',
-                   '-x',
                    '-vv',
                    '-rw',
-                   '--durations=10',
-                   '--cov=spyder',
-                   '--cov=spyder_profiler',
-                   '--cov-report=term-missing']
```
ccordoba12: Please restore everything of what you removed here, except for `'--cov-report=term-missing'`. That's still in master:

https://github.com/spyder-ide/spyder/blob/master/runtests.py#L41-L42

runtests.py
```diff
@@ -5,49 +5,57 @@
...
         raise SystemExit(errno)
 
 
 if __name__ == '__main__':
-    main()
+    test_parser = argparse.ArgumentParser(
+        usage='python runtests.py [--run-slow] [-- pytest_args]')
+    test_parser.add_argument('--run-slow', action='store_true',
+                             default=run_slow,
+                             help='Run the slow (mainly GUI/functional) tests')
```
ccordoba12: There's no need to have the `--run-slow` option defined here. It's handled in `conftest.py`, as the [Pytest docs](https://docs.pytest.org/en/latest/example/simple.html) recommend to do it:

https://github.com/spyder-ide/spyder/blob/3.x/conftest.py#L30

runtests.py
```diff
@@ -5,49 +5,57 @@
...
+    test_parser.add_argument('pytest_args', nargs='*',
+                             help="Args to pass to pytest")
+    test_args = test_parser.parse_args()
+    main(run_slow=test_args.run_slow, extra_args=test_args.pytest_args)
```
ccordoba12: Given the above comment, this only needs to be

    main(extra_args=test_args.pytest_args)

runtests.py
```diff
@@ -5,49 +5,57 @@
...
-                   '--cov=spyder_profiler',
-                   '--cov-report=term-missing']
+                   '--durations=10']
 
-    if run_slow:
+    if RUN_CI:
+        pytest_args.append('-x')
+    elif extra_args:
+        pytest_args += extra_args
+    if run_slow and '--run-slow' not in pytest_args:
```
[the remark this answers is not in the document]
CAM-Gerlach: > pytest_args is a fixed list, so there's no need to have the second condition.

Actually it isn't, as you can see from the line just before this one—the user may have added it as one of the ``extra_args``, a very likely possibility after the change you request below.

Also, ``run_slow`` != ``RUN_CI`` on ``master`` as discussed above, so unless you want to restructure how that is handled, this would still have to be there.

runtests.py
CAM-Gerlach: I'm aware; this PR when merged would remove it there as well, per one of its major goals:

> Removes the remaining % file coverage by default, as it can easily be re-enabled locally in the case the user really desires it for every file, and is already displayed in far more accessible form via two different dedicated services when run the CIs and otherwise just adds extra clutter to the logs

However, if you have some practical use case for still keeping them in every CI report that isn't equally or better served by Codecov, Coveralls and running it locally, then I'm all ears.

runtests.py
CAM-Gerlach: The reason I kept it was so it'd be exposed to the user e.g. through ``runtests.py --help`` and have clearer error handling with ``argparse``, etc. However, I can ahead and eliminate it since it does make the code cleaner and it isn't that big of a deal.

runtests.py
CAM-Gerlach: Yes

runtests.py
ccordoba12: Yes, please remove it, but leave it in the `usage` text above. I think that's enough.

runtests.py
ccordoba12: Ok, please go ahead with this.

runtests.py
```diff
@@ -5,49 +5,57 @@
...
-import qtpy  # to ensure that Qt4 uses API v2
 import pytest
 
 
 # To run our slow tests only in our CIs
-run_slow = False
-if os.environ.get('CI', None) is not None or '--run-slow' in sys.argv:
-    run_slow = True
+RUN_CI = os.environ.get('CI', None) is not None
+run_slow = RUN_CI
```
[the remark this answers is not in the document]
ccordoba12: This is very confusing and I don't like it. `run_slow` is not the same as `RUN_CI` (which is what you're equating here), so please find a better way to write this or just leave them separate:

    RUN_CI = os.environ.get('CI', None) is not None
    RUN_SLOW = False

runtests.py
ccordoba12: I don't have a problem with coverage being reported, but I'm almost never run the entire test suite locally, just the things I'm testing at the moment.

But sure, we can remove this.

runtests.py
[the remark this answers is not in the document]
ccordoba12: Ok, then this needs to be

    if RUN_CI and '--run-slow' not in pytest_args:

runtests.py
CAM-Gerlach: Okay, sure.

runtests.py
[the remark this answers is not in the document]
CAM-Gerlach: Okay, however you prefer. I just dropped ``run_slow`` entirely then since it doesn't actually do anything (different from ``RUN_CI``) in ``3.x``; I did it that way originally to make your life a little easier when merging it. When you merge, you'll need to do something about it so the ``run_slow`` tests skip on Darwin, but that's up to you ofc.

runtests.py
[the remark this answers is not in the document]
CAM-Gerlach: Okay sure, I did  that and ended up just refactoring the section entirely to make the logic simpler in ``3.x``. When you merge, you'll need to change it to check for the extra constant you add, or just directly, as to whether it is running on a macOS CI.

runtests.py
CAM-Gerlach: Okay, sure; done.

runtests.py
CAM-Gerlach: Yup, done.

CAM-Gerlach:
> The difference between RUN_CI and run_slow is still confusing to me (as proposed in this PR).

Right, its understandably confusing since there is no difference in ``3.x``, while there is on ``master`` since the ``run_slow`` tests are skipped when running on the CI under Darwin which ``run_slow`` currently checks for. I just left it as is to make your life easier when merging, but on request I removed it and you'll need to check for it somewhere when you merge with ``master``.

ccordoba12:
Thanks for the change @CAM-Gerlach! It looks much better now!

However, please notice that without the `--cov` options, Codecov and Coveralls don't report anything in the PR checks. Therefore, I think you need to put them back, but please do it only when `RUN_CI == True`

CAM-Gerlach:
> However, please notice that without the --cov options, Codecov and Coveralls don't report anything in the PR checks.

Thanks; good catch—I should have checked that. I added ``--cov=spyder`` and ``--cov=spyder-profiler`` to the ``if RUN_CI`` additional arguments list (without the ``--cov-report=term-missing`` that produced overly verbose text-mode reports, unless that's needed for Codecov/Coveralls—it was removed in master already so I assume not). I also added the ``--no-cov-on-fail`` option, which will avoid printing a full coverage report for failed runs (which can get particularly troublesome due to the 3x repeat runs on fail) and also wasting time and resources calculating coverage for a failed build.

ccordoba12:
> I added --cov=spyder and --cov=spyder-profiler to the if RUN_CI additional arguments list

Yeah, that's fine :+1: 

> I also added the --no-cov-on-fail option

Even better, I didn't know about that option.

PR #8187, opened 2018-11-01
PR: Don't install qtconsole from git on Appveyor because it's generating errors

ccordoba12:


ccordoba12:
No, it has to do with conda: after uninstalling qtconsole (to install its git version later), conda brought Qt 4 and when reinstalling qtconsole and Qt5, it occurred a Qt dependencies conflict that prevented Spyder to start.

PR #8192, opened 2018-11-02
PR: Remove dockwidget title bars by default and improve dock/undock behavior

### LINKED ISSUE #2854: Add custom title bar to dockwidgets/panes
# Description

Some PRs will depend on this functionality so this should be implemented in a separate PR where the new `TitleBarWidget` is made and tested

### LINKED ISSUE #2855: Hide the titlebar from panes/dockwidgets if locked
# Description

This depends on #2854

### LINKED ISSUE #5515: Enhancements proposal related to the undocking and docking of plugins in the Main Window
Some observations that I've noted regarding the undocking/docking of plugins in the mainwindow. This is a follow up of PR #5487 and #3824.

1. I think that new windows should  be created when undocking plugin from their cog menu when `Lock panes` is checked. Currently, undocking a plugin from its cog menu never create a new window, whether `Lock panes` is checked or not.

undocking_cog

2. The `Undock` action in the plugins cog menu is not disabled when undocking plugins through their undock button. This does not generate any error, so it's not a big deal, but it is confusing a little bit.

undock_not_deactivated

3. Instead of disabling the `Undock` action in the plugin cog menu when it is undocked, I would replaced it by a `Dock` action.

image

### LINKED ISSUE #8011: Restore option to open a new editor window
Hi guys,

First let me say kudos to the whole team and community for version 4! Even though it is still in beta I've been using it with no problems and it's amazing! I just love the new features in the editor e.g. the foldable code, the vertical guides and scrolling past the end! Great job!

I am missing one feature from Spyder3 though - the option to open a new window of the editor - it was extremely convenient when using a dual monitor display. Is it possible to put it back?

Thanks for all your hard work!

### LINKED ISSUE #8171: Plots plugin is not docked correctly after first start
I think this happens because it's not added to the layout methods of MainWindow

ccordoba12:
The purpose of this PR is to implement the following:

- [x] Hide DockWidgets title bar when locked and lock toolbars too.
- [x] Remove the undock button and make the undock action to create a new window
- [x] Show Editor's current file path on its title bar under all circumstances
- [x] Fix placement of File Explorer and Plots plugin
- [x] Add a close action to the Options menu of every plugin.
- [x] Change Options menu icon to be hamburger icon.
- [x] Restore "New window" action for the Editor
- [x] Change tour color widget to follow QDarkStyle Python 3 theme. Since we can't get the colors used by `qdarkstyle` programmatically, we have to hardcode them and we prefer the Python 3 theme instead of Python 2 one.
- [x] Make env and sys.path dialogs of the IPython console to follow the current theme.
- [x] Change "Browse tabs" icon
- [x] Fix icon color of "Restart kernel" action.
- [x] Close undocked windows when closing Spyder main window

----

Resources:

* https://woboq.com/blog/qdockwidget-changes-in-56.html
* https://stackoverflow.com/a/21701604/438386
* https://stackoverflow.com/a/23751750/438386

----

Screenshots

* Custom dockwidget title bar / New Options menu icon (the same one used in web browsers)
seleccion_002

* New Options menu actions (the last there in this screenshot): All plugins have now an Undock and a Close action.
seleccion_013

* Lock panes action locks toolbars now too
seleccion_017

----

Fixes #8171.
Fixes #8011.
Fixes #5515.
Fixes #2855.
Fixes #2854.


ccordoba12:
@jnsebgosselin, I'm taking over these tasks from you, hope that's fine :-)

With the new dark theme, I really think we need this.

jnsebgosselin:
> @jnsebgosselin, I'm taking over these tasks from you, hope that's fine :-)
> 
> With the new dark theme, I really think we need this.

Sure go ahead!

I have another suggestion that could be considered for this PR. Maybe when we check the `View ->Lock panes` option, we could also lock the toolbars in addition to the panels.

Here is an example of what I mean in LibreOffice:

image

jnsebgosselin:
Hum... This is strange that test that have been passing for the past 4-5 weeks or more now need to be skipped...

Since I wrote the problematic tests, do you want that I take a look at this? I'm currently out of town. I'll be back tomorrow night.

ccordoba12:
Those tests are passing for me locally, but I don't understand why they're failing now on Circle, Travis (only Linux) and Appveyor.

The last commit was just a test to see if only that test is failing or the other ones in that file too.

If you can help me to understand what's happening, that'd be great!

ccordoba12:
@jnsebgosselin, please review my changes carefully. They seem to have introduced failures in your tests because other PRs are passing them just fine (e.g. PR #8202). However, I don't understand how that could have happened.

jnsebgosselin:
Nice! I looked at the changes yesterday, but could not figured why this was causing the tests to fail.

ccordoba12:
@jnsebgosselin, @CAM-Gerlach, I think this PR is ready now. Please test it and let me know what you think.

CAM-Gerlach:
Testing in progress, but have to deal with a dumb browser issue first preventing me from uploading screenshots of the issues I've found.

ccordoba12:
> The new close button in the project explorer dropdown menu doesn't work

Will fix, thanks for testing that case.

> After undocking a pane while panes are unlocked, locking panes makes the undocked pane revert to a "docked" appearance

I haven't considered this situation. The easiest thing to do is to forbid locking panes when a there's a floating pane around.

> When opening a project started in --safe-mode and with only one Editor pane and the default temp file open, the Editor pane just goes blank with nothing open and I get the following error

Should be an easy fix, thanks.

> When the tile bars are hidden, the tabs of the Console pane are missing their top lines when active

Yeah, I noticed that too. This could be caused by tabs of dockwidgets on top.

> I'd prefer to be able to drag and drop both types of undocked panes when "Lock panes" is unchecked, as it could otherwise be very confusing to the user (as it was to me).

I don't know if this is possible: we would need to close the current PluginWindow and make the plugin floating. The problem is to put it in the same place as the window were (and what if it was present in another monitor?).

To avoid this, we could simply dock all undocked plugins when "Lock panes" is unchecked. I think that's the safest (although unpleasant) way. What do you think?

> The "Undock" and "New window" options in the Editor menu have the exact same duplicate effect

Are you saying that the undock action is doing the same as the "new window" one? Because the first removes the editor from the main window and other leaves it there.

> However, none of these methods at resolves the core request in #8011

That's not true. "New window" allows you to open as many other Editor windows as you want because you can keep selecting "New Window" in the Options menu of the Editor.

Also, the Options menu of the new window has also a "New window" action (and not the undock and close ones) to create more windows.

So why does that not solve #8011? 

> Therefore, in the case of the Editor, the "New Window" option should preserve the Editor pane as-is in the original window, like it did previously

I don't understand what you mean by this. The splitting scheme present in the main window is not preserved in the new one, not even in 3.x (the new window is to the right here):

seleccion_018

> (or at least allow opening a new one)

This is easy, and I think it makes sense. Pierre decided to clone to current tabs disposition to the new window, but that's really not intuitive. This is an easy change.

> Meanwhil,e the undock menu option should do the same thing the undock button in the title bar does except for making the window "independant"

Sure, that's easy to do too.

> I'm personally a soft +-1+ for the tabs on top change

I can revert that for now. I don't feel comfortable with that change either but I wanted to see how things looked with it.

@jnsebgosselin, any objections to revert it?

jnsebgosselin:
I  will take a look at this tonight and will comment back my feedback.

CAM-Gerlach:
> The easiest thing to do is to forbid locking panes when a there's a floating pane around.

I guess that makes sense, if the non-independent-window panes (undock button in titlebar) are only intended to be a temporary convenience for the user to easily reorganize them, while the undock menu option is intended to be the the mechanism to undock a pane for more than just a transient period. None of this was really obvious to me as a user just from the UI, but I'm not 100% how to make it such other than writing tooltips. Maybe @jnsebgosselin has some ideas.

> I don't know if this is possible: we would need to close the current PluginWindow and make the plugin floating.

Okay, if its requires doing something like that (with all the many edge cases and potential problems, etc) then I guess its not worth it. Part of it was that even then I didn't fully understand the intended distinction between the undock button and the undock menu option.

> To avoid this, we could simply dock all undocked plugins when "Lock panes" is unchecked. I think that's the safest (although unpleasant) way. What do you think?

I'd say that's actually a regression and doesn't really resolve the problem, so I guess we can just leave it as is. The underlying issue is that its unclear, both in the UI to trigger each and in the look of the resulting windows, that the function of the undock button is substantially different from the menu option.

> Are you saying that the undock action is doing the same as the "new window" one? Because the first removes the editor from the main window and other leaves it there.

Huh, somehow I thought I'd confirmed that when choosing the "New Window" option, the old Editor was still closed, but evidently not and I messed up somehow, unless it was an edge case based on the specific dock and undock actions I'd already taken (in the ``--safe-mode``  environment) since I couldn't repro it now either in a clean or dirty config, (i,e. ``--safe-mode`` or not) and with lock panes checked or unchecked and with a few different pane/panel configurations. This is what I was saying it *should* do, so it looks like this part is fine.

> So why does that not solve #8011?

It does indeed solve it, you're correct, and sorry for wasting your time on that one. As mentioned above, I just evidently messed up somehow testing it such that I thought it closed the original one. However, it seems that this is in fact not the case; either it was an error on my part (e.g. clicking on Undock accidentally instead, or not seeing it still there) or there was some corner-case reason it didn't.

> > Therefore, in the case of the Editor, the "New Window" option should preserve the Editor pane as-is in the original window, like it did previously

> I don't understand what you mean by this.

Again, sorry for confusing you. This just describes exactly what *does* happen correctly with regard to the old Editor staying open that I somehow missed.

So, **TL;DR**, the only remaining change I was requesting with regard to "New Window" and "Undock" in the Editor is that "Undock" preserve the split panel layout and tab order in each panel. "New window" starting with the default-one panel layout and tab order, as it currently does, is fine.

However, based on your comments above, we should also consider allowing different files to be open in each window's Editor; i.e. opening/closing a file in one window's does not close it other Windows' editors, but it *does* still sync it between panels of the same Editor like before. We still should, however, clone the currently opened files of the Editor in which "New Window" is clicked to the new editor window initially for the user's convenience, since it is way easier for them to close existing files individually or all at once than navigate to and re-open them.

jnsebgosselin:
Sorry guys, I haven't been able to take a good look at this yet. I'll try to do it this weekend.

CAM-Gerlach:
@ccordoba12 FYI, in my testing when panels are undocked as this PR implements, they do not close when Spyder does and hang around indefinitely (this is also true of e.g. the prefs window outside of this PR). There appears to be two methods (as I discuss in #8474 ) to address this. I assume that's in scope here.

ccordoba12:
I think this is finally ready!! My comments:

> The new close button in the project explorer dropdown menu doesn't work

I couldn't reproduce this one, sorry.

> After undocking a pane while panes are unlocked, locking panes makes the undocked pane revert to a "docked" appearance

I removed the undock button and leave only the undock action to avoid confusion. I also made undocked windows to be closed when Spyder is closed.

> When the tile bars are hidden, the tabs of the Console pane are missing their top lines when active

I reverted to have tabs on top to avoid this.

> I'd prefer to be able to drag and drop both types of undocked panes when "Lock panes" is unchecked, as it could otherwise be very confusing to the user (as it was to me).

This no longer applies because I removed the undock button. Besides, I added a custom title bar to all plugins that is shown when "Lock panes" is unchecked and that only let's you drag and drop panes to a different location.

>  Pierre decided to clone to current tabs disposition to the new window, but that's really not intuitive. This is an easy change.

I decided to leave the current behavior because this PR was getting too big (it could be implemented in another PR, but I think it's low priority).

----

@CAM-Gerlach, do you have something to add?

CAM-Gerlach:
@ccordoba12 Sounds good, thanks for all your hard work and for addressing my comments! I'll take one last look at it and let you know momentarily if I find anything egregious. This indeed has gotten pretty big so I don't want to get off topic and make it any bigger, just if I see a significant issue with the current functionality here.

ccordoba12:
Ok, thanks! If you have more suggestions on how to improve this functionality, please open a new issue about it. But for now, please focus on reviewing the things implemented here.

CAM-Gerlach:
@ccordoba12 I like the new UX and everything works and looks good (enough) and it seems the project explorer close button not working issue was fixed somewhere down the line. However, I did see some significant issue: 

On opening Spyder with panes locked (The default), the toolbar of the project explorer:

image

and the file explorer:

image

look all messed up, with the close button embedded in them. Resizing the window doesn't fix it; the only thing that does is toggling "unlock panes" then "lock panes" again and it thereafter looks normal. I assume something is/isn't getting fired on startup but isn't/is triggering on toggling that option.

Correct look:

image

image

ccordoba12:
> On opening Spyder with panes locked (The default), the toolbar of the project explorer and the file explorer look all messed up

Thanks for noticing it. This was due to a bug in the initialization of our custom title bar for dock widgets and it should be fixed with my latest commits.

ccordoba12:
Please check if that's the case to merge this one.

ccordoba12:
I finally added a couple of tests to verify that the basic functionality is working as expected.

Thanks @CAM-Gerlach for your help with this one!

Issue #8068, opened 2018-10-13
Fully implement Spyder's new dark theme


CAM-Gerlach:
## Problem Description

This issue serves to track the sub-issues related to fully implementing the dark theme across Spyder's various panes, text and icons, as a number of steps, both internal to Spyder's codebase and external in ``QDarkStyle``, are required.

Blocked by #8020 , the initial implementation. Originally spawned from #2350 , the initial issue and discussion thread.

Places that need to be "darkified":

* [x] #8087: Dark css style for Help and Console panes
* [x] Dialogs:
    - Pylint/Profiler ``Output`` window
    - Dependencies dialog
    - Github authentication
* [x] Interactive tour and menus:
    - Interactive tour
    - Online help and interactive tours submenus (of Help menu)
* [x] Help minor things
    - Change color of math equations
    - Change style of notes
    - Theme scroll bar
    - Theme table headers
* [x] Undocked windows
* [x] ~~Non-inline plot dialog (not sure if possible)~~ @ccordoba12: This is not possible because that dialog is created by Matplotlib.
* [x] ~~"Information" icon in file switcher isn't themed~~ (opened as a separate issue to close this one out)

Places that still have black/dark colors:

* [x] Find in files results
* [x] Editor widgets:
    - File switcher file list 
    - Symbol finder results
* [x] Outline Explorer code cell names 
* [x] Preferences table columns:
    - Keyboard shortcuts pref pane ``names`` column 
    - LSP Manager ``Command to Execute`` column
* [x] DataFrameEditor Index column
* [x] Pylint and Profiler panes' date and rating text
* [x] Icons:
    - "Gear"/pane options icon (all panes; dark gray is completely invisible/should be white)
    - Regex icon (find in files, find/replace; black/should be white)
    - About Spyder "i" icon (Help menu; black/should be white)
    - Case sensitivity icon (find in files, find/replace; dark blue is low contrast/try light blue)

Other Spyder Issues/Bugs:

* [x] #8069 : Spyder dark theme overrides background color set in syntax highlighting theme, except behind text
* [x] #8070 : Explanatory text for Greedy Completion preferences option cut off only in dark theme
* [x] #8071 : Centralize remaining theme-related options (like Rstudio does) under a renamed "Themes" pref pane
* [x] #8072 : Consider making internal console respect user's syntax highlighting theme
* [x] #8075 : Toolbar icons are too large particularly in dark theme
[...]

PR #8251, opened 2018-11-15
[WIP] PR: dark icons for regex, case sensitive and message box information

### LINKED ISSUE #8072: Make internal console respect user's syntax highlighting theme
## Problem Description

A part of #8068 .

Currently, the Internal Console and the Issue Report dialog use the default Spyder syntax highlighting theme, which wasn't as big of a problem in the past. However, now that Spyder is offering an application-wide dark theme, it looks odd to have these panes in a different color as the rest of the UI. As a stopgap, we could have them default to the Spyder Dark scheme if using the dark UI theme, but if we're going that far and it isn't much more effort, it would seem prudent to just use the proper theme to begin with.

e.g.

dalthviz:
### Pull Request Checklist

## Description of Changes

- [x] Add icons for regex, case sensitive and message box information that support the dark theme 
- [ ] Internal console dark theme support

### Issue(s) Resolved

Fixes #8072
Part of #8068


ccordoba12:
@dalthviz, please remove `regexp.svg` and the note about it in https://github.com/spyder-ide/spyder/blob/master/NOTICE.txt

Also, please remove `upper_lower.png`.

goanpeca:
Screenshot :-p ?

dalthviz:
@ccordoba12 @goanpeca a preview:

## Find/Replace:
imagen

## Find in Files:
imagen

## MessageboxInformation:
info

## Internal Console:
imagen

README.md
```diff
@@ -244,7 +244,7 @@ a Python version greater than 2.7 or 3.4 (Python <=3.3 is no longer supported).
 * **Cloudpickle**: Serialize variables in the IPython kernel to send to Spyder.
 * **spyder-kernels** 1.0+: Jupyter kernels for the Spyder console.
 * **keyring**: Save Github credentials to report errors securely.
-* **QDarkStyle**: A dark stylesheet for Qt applications, used for Spyder's dark theme.
+* **QDarkStyle** 2.6.0+: A dark stylesheet for Qt applications, used for Spyder's dark theme.
```
CAM-Gerlach: We should require 2.6.3 IMO, since that's the first version with all the fixes for the major issues mentioned in #8068 . Otherwise, it won't look nearly as good as users expect and has several visible bugs and problems in our UI.

requirements/requirements.txt
```diff
@@ -15,4 +15,4 @@ pyqt5
 keyring
 spyder-kernels>=1.0
 python-language-server
-qdarkstyle
+qdarkstyle>=2.6.0
```
CAM-Gerlach: See above comment re: 2.6.3 .

spyder/plugins/console/plugin.py
```diff
@@ -219,7 +239,13 @@ def exception_occurred(self, text, is_traceback):
 
         if CONF.get('main', 'show_internal_errors'):
             if self.error_dlg is None:
-                self.error_dlg = SpyderErrorDialog(self)
+                self.error_dlg = SpyderErrorDialog(
+                        self,
```
CAM-Gerlach: Correct hanging indent should be 4 spaces here, not 8

spyder/plugins/console/widgets/internalshell.py
```diff
@@ -136,12 +136,20 @@ class InternalShell(PythonShellWidget):
...
                  max_line_count=300, font=None, exitfunc=None, profile=False,
-                 multithreaded=True, light_background=True):
-        PythonShellWidget.__init__(self, parent,
-                                   get_conf_path('history_internal.py'),
-                                   profile)
+                 multithreaded=True, default_foreground_color=None,
+                 error_foreground_color=None, traceback_foreground_color=None,
+                 prompt_foreground_color=None, background_color=None):
+        PythonShellWidget.__init__(
+                        self, parent,
```
CAM-Gerlach: What's up with this hanging indent? Its like 16 spaces jeje

spyder/plugins/console/widgets/shell.py
```diff
@@ -53,11 +53,19 @@ class ShellBaseWidget(ConsoleBaseWidget, SaveHistoryMixin,
...
-                 initial_message=None):
+                 initial_message=None, default_foreground_color=None,
+                 error_foreground_color=None, traceback_foreground_color=None,
+                 prompt_foreground_color=None, background_color=None):
         """
         parent : specifies the parent widget
         """
-        ConsoleBaseWidget.__init__(self, parent)
+        ConsoleBaseWidget.__init__(
+                        self, parent,
```
CAM-Gerlach: Again, strange hanging indent.

spyder/plugins/console/widgets/shell.py
```diff
@@ -631,9 +639,18 @@ class PythonShellWidget(TracebackLinksMixin, ShellBaseWidget,
...
     
-    def __init__(self, parent, history_filename, profile=False, initial_message=None):
-        ShellBaseWidget.__init__(self, parent, history_filename, profile,
-                                 initial_message)
+    def __init__(self, parent, history_filename, profile=False,
+                 initial_message=None, default_foreground_color=None,
+                 error_foreground_color=None, traceback_foreground_color=None,
+                 prompt_foreground_color=None, background_color=None):
+        ShellBaseWidget.__init__(
+                        self, parent, history_filename, profile=profile,
```
CAM-Gerlach: Hanging indent—is there a reason for the apparently non-standard positioning I'm missing?

spyder/widgets/findreplace.py
```diff
@@ -97,13 +97,14 @@ def __init__(self, parent, enable_replace=False):
...
-        self.re_button = create_toolbutton(self, icon=get_icon('regexp.svg'),
+        self.re_button = create_toolbutton(self, icon=ima.icon('regex'),
                                            tip=_("Regular expression"))
         self.re_button.setCheckable(True)
         self.re_button.toggled.connect(lambda state: self.find())
         
         self.case_button = create_toolbutton(self,
-                                             icon=get_icon("upper_lower.png"),
+                                             icon=ima.icon(
+                                                     "format_letter_case"),
```
CAM-Gerlach: 4 spaces not 8 (Spyder's automatic behavior is incorrect)

spyder/widgets/reporterror.py
```diff
@@ -114,8 +113,16 @@ class ShowErrorWidget(TracebackLinksMixin, ConsoleBaseWidget, BaseEditMixin):
     QT_CLASS = QPlainTextEdit
     go_to_error = Signal(str)
 
-    def __init__(self, parent=None):
-        ConsoleBaseWidget.__init__(self, parent)
+    def __init__(self, parent=None, default_foreground_color=None,
+                 error_foreground_color=None, traceback_foreground_color=None,
+                 prompt_foreground_color=None, background_color=None):
+        ConsoleBaseWidget.__init__(
+                self, parent,
```
CAM-Gerlach: 4 spaces not 8

spyder/widgets/reporterror.py
```diff
@@ -184,7 +194,13 @@ def __init__(self, parent=None, is_report=False):
         self.input_description.textChanged.connect(self._contents_changed)
 
         # Widget to show errors
-        self.details = ShowErrorWidget(self)
+        self.details = ShowErrorWidget(
+                self,
```
CAM-Gerlach: 4 spaces not 8 per PEP8

jnsebgosselin:
> 
> 
> > while PEP 8 specifies 4 [spaces for hanging indents] except in function `def`s
> 
> That is not my understanding from https://www.python.org/dev/peps/pep-0008/#indentation and in particular "The 4-space rule is optional for continuation lines." Nevertheless, I usually prefer 4 spaces.

That is my understanding also from. PEP 8 is suggesting, not specifying to use 4 spaces for hanging indents.

image

But, in our wiki (https://github.com/spyder-ide/spyder/wiki/Dev:-Coding-Style#python-style), we clearly state to "**Always** use four (4) spaces for indentation."

So I suggest that we update this section every time a coding style issue arise in one of our PR and stick to it.

CAM-Gerlach:
> > while PEP 8 specifies 4 [spaces for hanging indents] except in function defs
>
> That is my understanding also. PEP 8 is suggesting, not specifying to use 4 spaces for hanging indents.

I mis-remembered that it was at the recommendation, not suggestion level but you are clearly correct. My mistaken wording, sorry. Ultimately, the justification for the change was that 4 spaces was what we universally use everywhere else for continuation lines, and therefore we should be consistent.

> So I suggest that we update this section every time a coding style issue arise in one of our PR and stick to it.

Great suggestion. For starters I can add what @ccordoba12 decided about not following the "one-line summary" rule for docstrings in the tests.

We can also add a "Project Standards" section in the Contributing Guide with a bullet point summary and then links to that wiki page for more information.

spyder/plugins/console/widgets/shell.py
dalthviz: Just a personal preference. I normally left the indent in a way I find easier to read the begin and end of the parameters while using hanging indent (without using a specific number of spaces)

CAM-Gerlach:
Responding to @dalthviz 's comment above:

> > ```
> >         ShellBaseWidget.__init__(
> >                         self, parent, history_filename, profile=profile,
> > ```
> >
> >  Hanging indent—is there a reason for the apparently non-standard positioning I'm missing?
>
> Just a personal preference. I normally left the indent in a way I find easier to read the begin and end of the parameters while using hanging indent (without using a specific number of spaces)

PEP 8 is meant to be broken if its clear that doing something different is optimal for a specific case, *and* it is broken consistently.

The problem, however, arises when one's own personal preference is not the same as everybody else's personal preference in a large project like Spyder, leading to both inconsistency (both de-jure, with our coding guidelines, and de-facto, with most if not all of the other new code in Spyder), and being more difficult to understand, never mind apply to one's own code, for anyone else not used to it. Personally, I find it confusing and harder to grasp at a glance myself; that may be because I'm not used to it but that's exactly my point). In fact, even in this PR it isn't done consistently, as several others I pointed out have 4 or 8 spaces.

Furthermore, this isn't even really a hanging indent at all, since by the definitions I've seen, a hanging indent is *necessarily* some fixed number >0 of spaces right of the start of the previous line or the margin, whereas this indent, like that lining up with the parens, is indexed to a particular character in the above. There's no particular name for it, since it doesn't exist at all in PEP 8. Finally, it due to the choice of indent (lining up with the ``ClassName.`` directly implies that all of the parameters of __init__ are properties of ``ClassName``, which isn't necessarily true for any arbitrary parameter of __init__ (they may not be set as properties, or at least not by their argument names, etc).

IMO, unless we all agree we like this style, will follow this style in all new code, and will instruct new contributors on this style, we should stick with a consistent 4-space hanging indent (or at least a n-space hanging indent with some consistent number of spaces, not...whatever this is, since its not even really a hanging indent at all).

@jnsebgosselin @jitseniesen Your opinions, since we just discussed this below?

jnsebgosselin:
> @jnsebgosselin @jitseniesen Your opinions, since we just discussed this below?

It's not my call, but I can still give my opinion :)

I'm in favor of using 4-space hanging indents everywhere. It's clear, easy to follow and ensure consistency throughout the code.

CAM-Gerlach:
> I'm in favor of using 4-space hanging indents everywhere. It's clear, easy to follow and ensure consistency throughout the code.

That's my opinion as well, and it also seems to be what we've been doing for all new and modified code.

jitseniesen:
> I'm in favor of using 4-space hanging indents everywhere. It's clear, easy to follow and ensure consistency throughout the code.

Me too, with the exception of situations like

```python
if some_long_function_name(
        param1, param2, param3):
    do_something()
    do_something_else()
```

where a 4-space indent in the second line would be quite confusing IMO.

CAM-Gerlach:
> with the exception of situations like

Yes, agreed; PEP 8 proposed a few different ways to handle ``if``, etc. blocks like that, but I recalled you mentioning that before as your preferred strategy and its consistent with how ``def`` is handled, so I'd already included it on the coding style page since we'd discussed in previously when it came up on a PR a while back.

spyder/app/mainwindow.py
```diff
@@ -2568,8 +2568,17 @@ def render_issue(self, description='', traceback=''):
...
+                    MAIN_TB_FG_COLOR, MAIN_PROMPT_FG_COLOR)
             from spyder.widgets.reporterror import SpyderErrorDialog
-            report_dlg = SpyderErrorDialog(self, is_report=True)
+            report_dlg = SpyderErrorDialog(
+                    self, is_report=True,
+                    default_foreground_color=MAIN_DEFAULT_FG_COLOR,
+                    error_foreground_color=MAIN_ERROR_FG_COLOR,
+                    traceback_foreground_color=MAIN_TB_FG_COLOR,
+                    prompt_foreground_color=MAIN_PROMPT_FG_COLOR,
+                    background_color=MAIN_BG_COLOR)
```
ccordoba12: @dalthviz, I really don't like that you define and pass all these parameters to widgets that inherit or use `ConsoleBaseWidget`. I think there are two possible ways to simplify this:

1. Define these constants where `ConsoleBaseWidget` is defined and use them only there (perhaps by applying its `set_color_scheme` method on its `__init__`).
2. If that's not possible, define a single parameter called `is_dark_interface` and pass that parameter to widgets that inherit or use `ConsoleBaseWidget`

spyder/widgets/reporterror.py
```diff
@@ -28,7 +28,6 @@
 from spyder.plugins.editor.widgets.codeeditor import CodeEditor
 from spyder.widgets.mixins import BaseEditMixin, TracebackLinksMixin
 from spyder.plugins.editor.widgets.base import ConsoleBaseWidget
-from spyder.plugins.editor.widgets.codeeditor import CodeEditor
```
ccordoba12: Please restore this import. `DescriptionWidget` below inherits `CodeEditor`.

requirements/requirements.txt
```diff
@@ -15,4 +15,4 @@ pyqt5
 keyring
 spyder-kernels>=1.0
 python-language-server
-qdarkstyle
+qdarkstyle>=2.6.3
```
ccordoba12: Please apply these changes to `setup.py` too.

spyder/app/mainwindow.py
```diff
@@ -2568,8 +2568,17 @@ def render_issue(self, description='', traceback=''):
     def report_issue(self, body=None, title=None, open_webpage=False):
         """Report a Spyder issue to github, generating body text if needed."""
         if body is None:
+            from spyder.plugins.console import (
+                    MAIN_BG_COLOR, MAIN_DEFAULT_FG_COLOR, MAIN_ERROR_FG_COLOR,
+                    MAIN_TB_FG_COLOR, MAIN_PROMPT_FG_COLOR)
```
ccordoba12: I think after the last changes you made, these three lines are not needed.

spyder/app/mainwindow.py
ccordoba12: And passing these parameters to `SpyderErrorDialog` is not needed either. Please restore this too.

spyder/widgets/reporterror.py
ccordoba12: Sorry, I just that import was repeated, so it was fine to remove it.

PR #8273, opened 2018-11-20
PR: Include QDarkStyle as an explicitly installed dep for the dev builds in the Contributing Guide

leportella:
### Pull Request Checklist

* [ ] Followed [PEP 8](https://www.python.org/dev/peps/pep-0008/) code style
* [ ] Ensured this pull request hasn't eliminated unrelated blank lines/spaces,
* [ ] Wrote at least one-line docstrings following PEP 257for any new functions
* [ ] Noted what issue(s) this pull request resolves, creating one if needed

## Description of Changes

I was following the `CONTRIBUTING` file to contribute to Spyder and after doing all the installation using the `conda`s commands, I still ended up with dependencies problems. Specifically, I had trouble with the `QDarkStyle` package. After I run the command `pip install -r requirements/requirements.txt` it worked. So, I added a line saying that if the user had trouble with dependencies he/she should run this command, to fix it. 

### Issue(s) Resolved
No issues on this

Fixes #

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -83,6 +83,9 @@ Spyder's necessary dependencies. The easiest way to do so (with Anaconda) is
 This installs all of Spyder's dependencies into the environment along with
 the stable/packaged version of Spyder itself, and then removes the latter.
 
+If still having troubles with dependencies after the steps above, please also 
+install the libraries by using the `requirements` file as shown below. 
+
 If using `pip` and `virtualenv` (not recommended), you need to `cd` to
 the directory where your git clone is stored and run:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,15 +76,12 @@ Spyder's necessary dependencies. The easiest way to do so (with Anaconda) is
 
 ```bash
   $ conda install -c conda-forge/label/beta spyder=4.0.0b1
-  $ conda install -c conda-forge python-language-server
+  $ conda install -c conda-forge python-language-server qdarkstyle
   $ conda remove spyder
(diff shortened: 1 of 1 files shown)


CAM-Gerlach:
> Specifically, I had trouble with the QDarkStyle package. 

You just needed to install QDarkStyle (``pip install qdarkstyle``), presumably.

> pip install -r requirements/requirements.txt

The problem is this will install/update all the dependencies using ``pip``, which is a recipe for problems especially if the user runs it in existing environment that may have older versions than the ``pip`` ones, especially  `` base``.

The basic issue here is the dependencies for the development Spyder are different from those for the latest beta; right now that is the case since we have PyLS, QDarkStyle, etc. as deps for Spyder dev but not beta 1, which was released a long time ago. Instead of this change, we should instead just add the necessary dep(s) to the line that already tells the user to explicitly install PyLS. Of course, this won't be an issue anymore in a week or two when Beta 2 is released and we'll have to update this anyway, so with only a few reports of this I'm not sure its worth doing but if you want to go ahead.

leportella:
I understand the problem of adding the requirements. Sorry! I just added it on the dependencies. Hope it is right :)

CAM-Gerlach:
@leportella For a neater commit history, since the change is so trivial and the first commit is entirely redundant, can you make the change in one commit (squash them, ``rebase -i`` then edit or fixup, ``git reset HEAD~2`` then commit again, etc)? And while your at it, if you don't mind starting the commit with a capital letter per convention. Thanks!

ccordoba12:
Closing in favor or PR #8299.

PR #8299, opened 2018-11-23
PR: Improve contributing guide

ccordoba12:
@spyder-ide/core-developers, this PR not only simplifies our Contributing guide but also the way to install our dependencies to get master up and running.

I dropped the need to use conda-forge (which is the source of multiple headaches) and instead moved the few packages we require and not present in the defaults Anaconda channel to our own channel.

This will also help new contributors and junior devs (like @leportella and @juanis2112) to have an easier first time experience (I attested that with @juanis2112 this morning: with these changes she set up an Spyder development environment in about 20 min).

@leportella, this supersedes your PR #8273.


requirements/conda.txt
```diff
@@ -11,8 +11,8 @@ pickleshare
 pyzmq
 chardet>=2.0.0
 numpydoc
-pyqt5
+pyqt>=5.6
```
CAM-Gerlach: Do we also want to specify <5.10 here like we have in setup.py (IIRC) since a number of bugs were reported with that?

requirements/conda.txt
```diff
@@ -11,8 +11,8 @@ pickleshare
...
 keyring
 spyder-kernels>=1.0
-python-language-server
+python-language-server=0.19*
```
CAM-Gerlach: With conda, you can just do ``0.19``, although it might just ignore the asterisk (I know it does work when you do 0.19.*``.

requirements/pip.txt
```diff
@@ -0,0 +1,18 @@
...
+qtpy>=1.2.0
+pickleshare
+pyzmq
+chardet>=2.0.0
+numpydoc
+pyqt5>=5.6
+keyring
+spyder-kernels>=1.0
+python-language-server==0.19.0
+qdarkstyle>=2.6.3
```
CAM-Gerlach: According to @dpizetta we should require 2.6.4 since it fixed some important bugs that broke QDarkStyle for Spyder users running under Python 2 .

requirements/pip.txt
```diff
@@ -0,0 +1,18 @@
...
+sphinx
+pylint
+psutil
+qtawesome>=0.5.0
+qtpy>=1.2.0
+pickleshare
+pyzmq
+chardet>=2.0.0
+numpydoc
+pyqt5>=5.6
```
CAM-Gerlach: See above pyqt version comment.

requirements/conda.txt
ccordoba12: Right, I forgot about it.

requirements/conda.txt
ccordoba12: Right, it's simpler.

requirements/pip.txt
ccordoba12: Ok, will do in this PR.

ccordoba12:
Thanks!

stonebig:
is it really neede to have python-language-server=0.19  limited ? 0.21.4 is not ok ?

CAM-Gerlach:
Yep, unfortunately, for the moment. At least on Windows in both my and @jnsebgosselin 's testing we found 0.21 crashed on startup and 0.20 didn't work reliably.

PR #8342, opened 2018-12-01
Converts title arg to unicode with the function to_text_string

juanis2112:
## Description of Changes

I converted title arg to unicode with the function to_text_string to solve issue 7930 in python 3

screen shot 2018-11-30 at 5 05 00 pm

### Issue(s) Resolved

Fixes # 7930

### Affirmation

I certify the above statement is true and correct: juanis2112

### DIFF
--- a/spyder/plugins/variableexplorer/widgets/texteditor.py
+++ b/spyder/plugins/variableexplorer/widgets/texteditor.py
@@ -20,7 +20,7 @@ from qtpy.QtWidgets import (QDialog, QHBoxLayout, QPushButton, QTextEdit,
 from spyder.config.base import _
 from spyder.config.gui import get_font
 from spyder.py3compat import (is_binary_string, to_binary_string,
-                              to_text_string)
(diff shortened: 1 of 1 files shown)


spyder/plugins/variableexplorer/widgets/collectionseditor.py
```diff
@@ -332,7 +332,10 @@ def data(self, index, role=Qt.DisplayRole):
         if index.column() == 3:
             display = value_to_display(value, minmax=self.minmax)
         else:
-             display = to_text_string(value)
+            if PY2:
+                display = to_text_string(value, encoding="utf-8")
```
ccordoba12: Tests are failing because we're assuming here that `value` is always a string, but that's wrong, it can be any object.

So please change this line to be

```python
if is_type_text_string(value):
    display = to_text_string(value, encoding="utf-8")
else:
    display = to_text_string(value)
```

(`is_type_text_string` is a function present in `py3compat`).

That should be enough.

PR #8347, opened 2018-12-02
PR: Do atomic writes when saving files

### LINKED ISSUE #8237: Mitigate Spyder wiping users' files with more robust atomic saves/autosaves

1. Save file to disk with connection issues (confirmed)
2. Crashes during runfile or during saving immediatly prior (reported 2x)
3. Other problems during saving???

Given we've publicly advertised the autosave feature as the solution to the reports we've been getting of Spyder completely wiping users' files on occasion e.g. @ccordoba12 in #6571:

> I don't understand how Spyder can delete your script, but we're working to provide autosaving functionality for our next major release (Spyder 4).

and particularly now that our own developer @jnsebgosselin is regularly experiencing a such a problem and can reproduce it, we should ensure autosave is reasonably "fail-safer" in the event that saving the file normally fails (silently or otherwise) and wipes or corrupts the file on disk. As currently implemented by @jitseniesen following his hard work in #7660 , the feature is reasonably sophisticated with a nice UI. However, as an initial implementation, it does leave open several significant opportunities for catastrophic data loss in several major scenarios:

*Failure mode:*
Saving may fail silently, or in flushing the data to the storage medium, which would result in the autosave still being deleted and the save being corrupted or wiped (or at least out of date). This would possibly apply to the repeated instances of files being wiped that @jnsebgosselin is experiencing (depending on exactly how the failure occurs).

*Mitigations:*
* Hardening the e.g. FileInfo._write_to_file() function, e.g. with ``<file>.flush()`` then ``os.fsync()``; switching to the more modern ``io.open()`` on Python 2 and making sure files are explicitly closed might help a little too
* Writing to a tempfile when saving, then doing an atomic replace of the original by the appropriate platform-specific method
* ~~If there are significant cases where the OS cannot be trusted to throw an error if something goes wring, we might need a readback check for the tempfile vs. the in memory version (at minimum, that the file reads without error and is the correct length, and perhaps comparing the first line and last lines), though this could be user-optional if its not determined to be critical or it takes a non-trivial amount of time~~
* ~~Temporarily disable autosave and warn the user if a save failure is detected, and notify the user of this; only re-enable once a manual save is successful~~

I've included the full form of the discussion between various parties on this issue below, for reference in one place.

dalthviz:
## Description of Changes

Use of `atomicwrites` to prevent possibility of Spyder wiping user's files.

### Issue(s) Resolved

Fixes #8237 

### Affirmation

### DIFF
--- a/spyder/utils/encoding.py
+++ b/spyder/utils/encoding.py
@@ -235,7 +235,9 @@ def write(text, filename, encoding='utf-8', mode='wb', overwrite=True):
         with open(filename, mode) as textfile:
             textfile.write(text)
     else:
-        with atomic_write(filename, overwrite=overwrite, mode=mode) as textfile:
+        with atomic_write(filename,
+                          overwrite=overwrite,
+                          mode=mode) as textfile:
             textfile.write(text)
     return encoding
 
--- a/spyder/utils/encoding.py
+++ b/spyder/utils/encoding.py
@@ -231,8 +231,12 @@ def write(text, filename, encoding='utf-8', mode='wb', overwrite=True):
     Return (eventually new) encoding
     """
     text, encoding = encode(text, encoding)
-    with atomic_write(filename, overwrite=overwrite, mode=mode) as textfile:
-        textfile.write(text)
+    if 'a' in mode:
+        with open(filename, mode) as textfile:
+            textfile.write(text)
+    else:
+        with atomic_write(filename, overwrite=overwrite, mode=mode) as textfile:
+            textfile.write(text)
     return encoding
 
 
--- a/spyder/utils/encoding.py
+++ b/spyder/utils/encoding.py
@@ -20,6 +20,7 @@ import sys
 
 # Third-party imports
 from chardet.universaldetector import UniversalDetector
+from atomicwrites import atomic_write
 
 # Local imports
 from spyder.py3compat import (is_string, to_text_string, is_binary_string,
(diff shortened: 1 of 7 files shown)


spyder/utils/encoding.py
```diff
@@ -224,16 +225,23 @@ def to_unicode(string):
     return string
     
 
-def write(text, filename, encoding='utf-8', mode='wb'):
+def write(text, filename, encoding='utf-8', mode='wb', overwrite=True):
```
ccordoba12: I don't think we need to set `overwrite` to any other than `True`, so please remove this keyword arg from here.

spyder/utils/encoding.py
```diff
@@ -224,16 +225,23 @@ def to_unicode(string):
...
     """
     text, encoding = encode(text, encoding)
-    with open(filename, mode) as textfile:
-        textfile.write(text)
+    if 'a' in mode:
+        with open(filename, mode) as textfile:
+            textfile.write(text)
+    else:
+        with atomic_write(filename,
+                          overwrite=overwrite,
```
ccordoba12: Let's change this to `overwrite=True`.

CAM-Gerlach:
Don't we also need to add it to to the Contributing Guide?

ccordoba12:
No need because now the guide recommends to use `conda.txt` to install all our deps.

CAM-Gerlach:
Oh, that's right; sorry.

ccordoba12:
And I also improved the guide a little bit more in PR #8348.

PR #8348, opened 2018-12-02
PR: Stop using ci-helpers to simplify testing

ccordoba12:
This will also help us to check that the files in `requirements` are up to date and will reduce installation time on Appveyor (because it comes with Miniconda installed).


appveyor.yml
```diff
@@ -36,24 +23,25 @@ install:
       https://ci.appveyor.com/api/projects/$env:APPVEYOR_ACCOUNT_NAME/$env:APPVEYOR_PROJECT_SLUG/history?recordsNumber=50).builds | `
       Where-Object pullRequestId -eq $env:APPVEYOR_PULL_REQUEST_NUMBER)[0].buildNumber) { `
         throw "There are newer queued builds for this pull request, failing early." }
-  - "git clone git://github.com/astropy/ci-helpers.git"
-  - "powershell ci-helpers/appveyor/install-miniconda.ps1"
   - "SET PATH=%PYTHON%;%PYTHON%\\Scripts;%PATH%"
+  - "conda create -y -n test python=%PYTHON_VERSION%"
```
jnsebgosselin: It's not a review, it's a question. I hope you don't mind me asking here.

Instead of creating a new environment, would it have been ok to define the variable "PYTHON" in the matrix instead as PYTHON: "C:\\Miniconda27-x64" and `PYTHON: "C:\\Miniconda36-x64"`, respectively, for Python 2.7 and 3.6 and test directly in the base environment?

I think that may reduce the installation time, but I don't know if there is any downside to that.

goanpeca:
Yesss! I stopped using them a long time ago on my projects, they things more complicated at the end :-\

appveyor.yml
```diff
@@ -36,24 +23,25 @@ install:
...
   - "activate test"
-  # Install qtpy from Github
-  - "pip install git+https://github.com/spyder-ide/qtpy.git"
-  # Fix problems with latest pyqt
-  - "conda install pyqt=5.6*"
+  - "conda install -q -y -c spyder-ide --file requirements/conda.txt"
```
jnsebgosselin: I think the current version of conda install on AppVeyor is old, do you think that calling `conda update -q conda` is necessary? I'm doing it for my projects... but I don't know if I should or not.

jnsebgosselin: If you think this is not, I think I will stop doing that to reduce installation time.

ccordoba12: It seems newer versions of conda are very, very slow. So I'd prefer to leave it at its current version.

jnsebgosselin: Thank you very much for the insight.

PR #8360, opened 2018-12-04
PR: Change creationflags to prevent showing a cmd for pyls

### LINKED ISSUE #8066: cosmetic: the pyls process opens a black DOS windows under Windows10
trying under Windows Python-3.7.1rc1 under with spyder-master of 2018-10-13 (commit https://github.com/spyder-ide/spyder/commit/31f49d10ee407ea19b9bd78a14352f77869d2133)

The Python-langage-Server process opens a Dos window, that remains by default in front of Spyder screen.
Would it be possible to have an option to launch it hidden ?
image

dalthviz:
## Description of Changes

Change `creationflags` used for the start of the pyls server 

### Issue(s) Resolved

Fixes #8066 

### Affirmation


spyder/plugins/editor/lsp/client.py
```diff
@@ -124,7 +124,7 @@ def start(self):
                 ' '.join(self.server_args)))
             creation_flags = 0
             if WINDOWS:
-                creation_flags = subprocess.CREATE_NEW_PROCESS_GROUP
+                creation_flags = 0x08000000 # CREATE_NO_WINDOW
```
ccordoba12: Why not append to the previous flag instead of replacing it entirely? What does `subprocess.CREATE_NEW_PROCESS_GROUP` does?

dalthviz: Totally right, `CREATE_NEW_PROCESS_GROUP` makes possible to use os.kill() in the subprocess

dalthviz: This is what the docs say: 

>Windows: The signal.CTRL_C_EVENT and signal.CTRL_BREAK_EVENT signals are special signals which can only be sent to console processes which share a common console window, e.g., some subprocesses. Any other value for sig will cause the process to be unconditionally killed by the TerminateProcess API, and the exit code will be set to sig. The Windows version of kill() additionally takes process handles to be killed.

ccordoba12: Ok, thanks!

Issue #8375, opened 2018-12-06
Spyder editor does not open previously opened files in a project


jhoutword:
### Issue Report Checklist

* [x] Searched the [issues page](https://github.com/spyder-ide/spyder/issues?q=is%3Aissue) for similar reports
* [x] Read the relevant sections of the [Spyder Troubleshooting Guide](https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ) and followed its advice
* [x] Reproduced the issue after updating with ``conda update spyder`` (or ``pip``, if not using Anaconda)
* [ ] Could not reproduce inside ``jupyter qtconsole`` (if console-related)
* [x] Tried basic troubleshooting (if a bug/error)
    * [x] Restarted Spyder
    * [ ] Reset preferences with ``spyder --reset``
    * [ ] Reinstalled the latest version of [Anaconda](https://www.anaconda.com/download/)
    * [ ] Tried the other applicable steps from the Troubleshooting Guide
* [x] Completed the **Problem Description**, **Steps to Reproduce** and **Version** sections below

## Problem Description
Spyder editor does not open previously opened files in a project.

At any time, I may switch between projects. I would close project A, and open project B. But if I want to return to work on project A, I find that the editor does not automatically open all the files that was opened before I closed project A the last time.

Only when I close Spyder entirely, then will the project state be saved.

### What steps reproduce the problem?

1. Install anaconda
2. Open spyder
3. Create a project with files in them. 
4. Open the project, open some saved file Z in editor. Close all other files.
5. Close project.
6. Open project again.
7. Check if file Z is opened in editor.

### What is the expected output? What do you see instead?
File Z should be opened. It is not opened.

* Spyder version: 3.3.1
* Python version: 3.5.4 64 bits
* Qt version: 5.6.2
* PyQt version: 5.6
* Operating System name/version: Windows 8.1

jnsebgosselin:
Thanks for reporting @jhoutword. I think I can reproduce this in master also. I will investigate and see if I can fix this quickly.

jnsebgosselin:
@ccordoba12 Milestone?

jnsebgosselin:
I think this was introduced here: https://github.com/spyder-ide/spyder/commit/3d7a94278f7bb5d58d35b9ff0e33e00519dfc594#diff-271b198d60769127e85929e3fcd8f61a, so in version 3.2.7.

[...]

PR #8415, opened 2018-12-11
[WIP] PR: Version Control Color coding

alexschw:
## Description of Changes
### TODO (I would need a hint)
* [ ] Color code should not be hardcoded

I added a color code for the project explorer that is responding to the version control system(vcs) and colors files accordingly.

Added = green
Changed = blue
Untracked = red
Ignored = grey

Colored Spyder

### Affirmation

I certify the above statement is true and correct: alexschw


ccordoba12:
This looks really good, thanks!

A very quick review: please revert all blank space removal you performed because that could impact the work of others or mine too when merging between branches. Thanks!

spyder/plugins/explorer/widgets.py
```diff
@@ -372,24 +407,24 @@ def create_file_manage_actions(self, fnames):
...
         if sys.platform == 'darwin':
-            text=_("Show in Finder")
+            text = _("Show in Finder")
         else:
-            text=_("Show in Folder")
-        external_fileexp_action = create_action(self, text, 
-                                triggered=self.show_in_external_file_explorer)        
+            text = _("Show in Folder")
+        external_fileexp_action = create_action(self, text,
+                            triggered=self.show_in_external_file_explorer)
```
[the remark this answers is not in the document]
alexschw: I reverted all whitespace changes.

spyder/plugins/explorer/widgets.py
```diff
@@ -181,6 +181,38 @@ def icon(self, icontype_or_qfileinfo):
...
 
 
+class ColorModel(QFileSystemModel):
+    """FileSystemModel providing a color-code for different commit-status."""
+    def __init__(self, *args, **kwargs):
+        super(ColorModel, self).__init__(*args, **kwargs)
+        self.vcs_state = None
+        self.color_array = [QColor("#ff0000"), QColor("#555555"),
+                            QColor("#0099ff"), QColor("#00ff00"),
+                            QColor("#ffffff")]
```
alexschw: I need to replace those with values from the styles. Otherwise, this will cause problems in light themes.

alexschw:
Thank you for the thorough review, @CAM-Gerlach. 

I am very sorry, that it crashed your Projects. I could reproduce the same error on a freshly pulled master repo. I think this #8375 issue is related.

> And of course, nothing showed up in terms of added/changed/etc files in the Project Explorer

I could not yet reproduce that yet. Maybe it is due to MINGW instead of the linux git, otherwise it may be that the directory is not correctly split due to spaces. I will get a Windows and look into that as soon as the previous problem is solved.

CAM-Gerlach:
Thanks for your response @alexschw !

> I could reproduce the same error on a freshly pulled master repo.

The not being able to load the saved session or the actual hang/crash? Also, I did test on ``master`` and didn't reproduce either of the errors nor not loading the save session, but now that I try again I can repro the not loading saved state, although I can't confirm if it happens on projects that I haven't already loaded/closed with your PR branch active. However, I don't get either of the errors loading projects either during my previous testing or now (nor the crash, but I couldn't repro that consistently in either) so those are apparently indeed due to this PR.

> Maybe it is due to MINGW instead of the linux git

I don't think so, since I just tested on my Linux VM (Fedora 28) and I got the exact same error on your PR branch after creating and opening a fresh VCS-tracked project and I wasn't able to get files to show up as added, modified, etc just like on Windows.

> otherwise it may be that the directory is not correctly split due to spaces.

Some of the directories I tried are on an absolute path that has spaces, while others are not, and I reprod one error or the other consistently on both,

CAM-Gerlach:
@alexschw It crashes? Or just raises the above exception?

I can now repro it in master as well (I couldn't before); my guess is that its due to a side effect of #8375 , which results in ``layout`` being ``None``. If that's the case, it should only happen when opening an existing project with no files saved. Regardless, its a separate issue so I'll open one and try to submit a PR with a patch. Thanks.

alexschw:
@CAM-Gerlach I think this solves the problem. 
If it does work, I will work on unit-tests.
If it doesn't, my environment is a Windows 10
PyQt5 5.11.3
QtPy 1.5.2
pygments 2.3.0
qdarkstyle 2.6.4
sphinx 1.8.2
pyls 0.21.4
nbconvert 5.4.0
qtconsole 4.4.3
IPython 7.2.0
pylint 2.2.2

spyder/config/main.py
```diff
@@ -270,6 +270,14 @@
               'show_hscrollbar': True,
               'visible_if_project_open': True
               }),
+            ('vcs',
```
CAM-Gerlach: It would be useful to be able to modify these settings in the UI; is there anywhere this can be done?

spyder/config/main.py
```diff
@@ -270,6 +270,14 @@
               'show_hscrollbar': True,
               'visible_if_project_open': True
               }),
+            ('vcs',
+             {
+              'use_vcs_higlighting': True,
+              'color/untracked': "#ff0000",
+              'color/ignored': "#555555",
```
CAM-Gerlach: It is very difficult to read the filenames with this color in the dark theme. Can you use ``#999999`` from our Spyder Dark syntax highlighting theme palette? Thanks.

spyder/plugins/explorer/widgets.py
```diff
@@ -192,6 +193,41 @@ def icon(self, icontype_or_qfileinfo):
...
+        normalstyle = CONF.get('appearance', 'selected') + '/normal'
+        self.color_array = [QColor(CONF.get('vcs', 'color/untracked')),
+                            QColor(CONF.get('vcs', 'color/ignored')),
+                            QColor(CONF.get('vcs', 'color/modified')),
+                            QColor(CONF.get('vcs', 'color/added')),
+                            QColor(CONF.get('appearance', normalstyle)[0])]
+        self.root_path = ''
+        super(ColorModel, self).__init__(*args, **kwargs)
+
+    def setVCSState(self, root_path):
```
CAM-Gerlach: Please use standard Pythonic ``snake_case()`` for your function and method names unless you are overriding a Qt method.

spyder/plugins/explorer/widgets.py
```diff
@@ -865,7 +901,7 @@ def vcs_command(self, fnames, action):
             QMessageBox.critical(self, _("Error"),
                 _("""<b>Unable to find external program.</b><br><br>%s""")
                     % to_text_string(msg))
-
```
CAM-Gerlach: Please don't remove unrelated blank lines or trailing spaces (we'll do it globally at some point) since it makes your PR harder to review and leads to merge conflicts with other PRs. Thanks.

spyder/utils/vcs.py
```diff
@@ -118,7 +157,6 @@ def get_hg_revision(repopath):
 def get_git_revision(repopath):
     """
     Return Git revision for the repository located at repopath
-
```
CAM-Gerlach: Not sure why this was removed as it appears unrelated, but a space between the one line summary and the main docstring body is required per PEP257.

spyder/config/main.py
alexschw: good idea. I will look into that.

alexschw:
> Looking forward to your tests as well! Let us know if you need guidance on that.
I tried to work on the tests, but if I'm trying to run the current tests (python3, pytest, pytest-cov and pytest-qt installed) I get Errors saying fixture 'mocker' not found...

jnsebgosselin:
You need to install `pytest-mock`

spyder/config/main.py
CAM-Gerlach: Does this happen when you pass ``--safe-mode`` to ``bootstrap.py`` (i.e. a clean preferences file)? I cannot reproduce the issue opening and closing preferences and moving through every tab on both a dirty and a clean config.

alexschw: okay, it was just due to my setup

alexschw:
Okay. I hope the tests run through now.
Could you review it again please @CAM-Gerlach?

spyder/config/main.py
CAM-Gerlach: Thanks.

spyder/utils/tests/test_vcs.py
```diff
@@ -46,5 +47,19 @@ def test_git_revision():
     assert all([isinstance(x, str) for x in get_git_revision(root)])
 
 
+def test_vcs_state(tmpdir):
+    """Test if the vcs state of the directory and subdirectories is returned"""
```
CAM-Gerlach: Missing period.

alexschw:
>    * Ignored directories (as opposed to files) are shown in the normal color, and ignored files, even those with an ignored extension (that works elsewhere) and newly added and saved within Spyder in that directory are located in an ignored directory, also show up in the normal color.

I also got that feeling, yet in my case it is just that there is such a little difference between `#FFFFFF` and the proposed `#999999` color. So I suggest using something inbetween (like `#777777`)

>    * At least on my machine, staged files no longer show up in green properly. When doing `git add` on an existing modified file across many tests, the file continues showing its old color, even after triggering a save of another file.
> 
>    * Furthermore, if additional modifications are then made to the file and it is saved, it turns the standard (white) color, even if it is not tracked, ignored, modified, etc.

yes, this is not working by now. I will work on that.

> Is there anything you can do about these performance hits? Having to deal with this much lag, especially when saving, would make me hesitate to use Spyder projects in these situations. since I save very frequently.

I am not sure how to shorten the loading time when opening a project. I would either check if there are more than 5 subrepos and disable the checking or i could make the coloring a project option that you could disable by hand.
I have an idea, on how to speed up the saving. Will work on that. Good suggestion, thanks.

CAM-Gerlach:
> I also got that feeling, yet in my case it is just that there is such a little difference

Okay, how about we try ``#808080``, 50% gray?

> I am not sure how to shorten the loading time when opening a project.

This is probably acceptable if you can't reduce it, since it only seems to occur the first time a project is opened. It would be nice to have an option to disable the coloring in the Project Explorer options menu since some users may prefer to have the feature off, which would also disable the checking until it was turned back on.

It is the lag when saving and the clicking the directory dropdown arrows in the Project Explorer that is much more problematic, so hopefully you can greatly reduce both of those. Thanks!

CAM-Gerlach:
Hey @alexschw , any updates on this? Also, just curious, would it be possible (either in this PR or in the future) to add a specific color for files with merge conflicts as well? Thanks!

alexschw:
> Hey @alexschw , any updates on this? 

not jet. Not much time currently. I hope to get more time in the next days.
> Also, just curious, would it be possible (either in this PR or in the future) to add a specific color for files with merge conflicts as well? Thanks!

Yes of course. What color would be good for that? I suggest `#ff7700` but `#cc00cc` would be okay too (Any other suggestions?).

CAM-Gerlach:
@alexschw ``#ff7700`` should be fine, I guess, as long as it looks distinct from red. 

You'll need to ``merge`` to fix the conflicts as well as grab the latest test fixes.

Thanks!

CAM-Gerlach:
Hey @alexschw , any luck finding the time to finish this?

alexschw:
The confpage.py file of the project doesn't appear in the Preferences. Is this also happening in the current master? I will need some hints on how this should work. Otherwise I think I'll have to rebuild it from scratch to get it working.

CAM-Gerlach:
Your test is failing on Windows. You can see the detailed output [here](https://dev.azure.com/spyder-ide/spyder/_build/results?buildId=776).

> The confpage.py file of the project doesn't appear in the Preferences.

If I understand what you're saying correctly, we're still in the process of implementing project-specific settings for Spyder 4, so they don't work yet. You should make it a standard global setting instead, and we'll transition it to also be a project-specific setting as well once that's done.

CAM-Gerlach:
@alexschw Hey, did you delete your reply? I got a notification about it in my email, but I can't see it here.

> I mean there is already a file that should define such a global settings page (confpage.py) 

Sorry, I was confused and I thought you meant for per-project configuration. The top-level ``confpage.py`` isn't displayed directly; it just defines some basic classes for each "plugin"'s (as each Spyder module, like the Editor, Console, Variable Explorer, help viewer, etc is implemented under the hood as a modular plugin) to use to define its own configuration page.

The  ``plugins/projects/confpage.py`` would be the natural place to put your settings (presumably an enable/disable checkbox and the color configuration; the former you'd also want to include in the drop-down options menu on the Project Explorer itself). However, it contains configuration options that aren't currently functional to my knowledge, which is why it isn't displayed. In order to enable showing it in the GUI, you'd need to add ``CONFIGWIDGET_CLASS = ProjectConfigPage`` right under ``CONF_SECTION = 'project_explorer'`` in ``plugins/projects/plugin.py``. However, it would show the existing non-working settings as well, which we don't necessarily want. Alternatively, if you're just adding the on/off option and not the color settings, you could just add it to the options menu of the Project Explorer and not include it in the Preferences.

 @ccordoba12 , what do you suggest we do about this?

goanpeca:
@alexschw could you please add screenshots of the config dialogs you are implementing?

I think that making that change seems too much for this PR, so I would rather this PR added an option that is "on" by default, and a subsequent PR work on all the project configuration. 

I will start working of revamping of projects in spyder to enable workspace with multiple projects (located in different folders) so the changes you are making to the config dialog will most likely not work in the new scheme.

alexschw:
> @alexschw could you please add screenshots of the config dialogs you are implementing?
preference_menu

> I think that making that change seems too much for this PR, so I would rather this PR added an option that is "on" by default, and a subsequent PR work on all the project configuration.

It is nothing big. I moved the other confpage.py back to confdialog.py and changed the tests accordingly. For some reason it is not shown that way in github. 

> I will start working of revamping of projects in spyder to enable workspace with multiple projects (located in different folders) so the changes you are making to the config dialog will most likely not work in the new scheme.

okay. Say, if I could help you @goanpeca.

goanpeca:
>It is nothing big. I moved the other confpage.py back to confdialog.py and changed the tests accordingly. For some reason it is not shown that way in github.

Will probably change that dialog anyway as I don't think the colors should be exposed there. 

Did you manage to solve all the issues raised by @CAM-Gerlach ?

CAM-Gerlach:
@alexschw Let me know when you're ready for me to do another review.

Looking good on the prefs dialog overall. A few comments based on the screenshot:

* Capitalize the names of the file categories
* Position the color selector widgets much closer to the right of the file categories, like they are in the color scheme editor (``Appearance > Syntax Highlighting theme > Edit selected theme``). rather than way far to the right
* It is very unclear what ``Interface > Show all data`` and ``Version Control > Use version control`` do, even to me as a developer and reviewer on this PR. Could you explain the precise function of each, so I can suggest much more helpful names and tooltips? Thanks.

> I think that making that change seems too much for this PR, so I would rather this PR added an option that is "on" by default, and a subsequent PR work on all the project configuration.

@goanpeca Well, unless performance has increase by a factor of ~10x since the last time I tested it (@alexschw , let me know when you're ready for me to do another review), then at the very least an on/off option is mandatory (which is all I initially envisioned as being necessary) since otherwise it induces a large amount of lag in large project directories like mine.

Once the items above in my review are addressed and the preferences dialog is satisfactory (and any other feedback is addressed), this can be ready to merge and iterate upon as needed.

CAM-Gerlach:
> Will probably change that dialog anyway as I don't think the colors should be exposed there.

Where would you suggest, @goanpeca ? It isn't like that dialog is crowded as it is... The colors should presumably be only set on a global level, while enabling/disabling VCS should definitely we configurable on the project level as well.

goanpeca:
>Where would you suggest, @goanpeca ? What's the rationale?

They should not be configurable by the user. They should be inferred from global colors used in spyder (by theme or syntax highlight) so I would not bother to expose them.

>while enabling/disabling VCS should definitely we configurable on the project level as well.

Agreed (if the performance issue is still a real problem, which will need to be addressed soon anyway)

CAM-Gerlach:
> They should be inferred from global colors used in spyder (by theme or syntax highlight)

Is the rationale for this so they match the theme and aren't hard to see when switching? if so, they should correspond to wherever we pulled the dark and light theme color palette for the Variable Explorer, since AFAIK all the same base colors are used (with different variations for dark and light BG as appropriate). There's no sense matching them to the syntax highlighting theme, since that's 100% user configurable too and there are no conceptual linkage between them.

The advantage of being able to customize the colors, however, is that users can suppress certain color codes they don't care to see (i.e. by setting them to the same color as normal/unmodified files), or group them as they see fit. For example, one user may want to only see modified files in one specific color, while another might not want to see ignored files differently from normal. Whether that capability outweighs any downside of retaining the modest additional complexity is something that would need to be discussed and decided upon before unilaterally removing it.

On the other hand, the primary advantage I see of not letting users customize them is that the colors can adapt automatically to the theme, but perhaps adding a checkbox along the lines of ``Use automatic VCS colors from theme``, checked by default, would take care of that (while unchecking would allow manual modification as currently).

alexschw:
Thanks for the suggestions @CAM-Gerlach and @goanpeca. I hope that the test work now. You may review it, if you have the time. Sadly there is not yet a solution, if the files are changed from outside of spyder ([See here](https://stackoverflow.com/questions/47851152/qfilesystemmodel-not-updating-when-files-change/47956650#47956650)). It might be possible. But I think this is outside of the scope of this PR. ;) If this is done, I might add another (hopefully) short PR for the context menu of vcs add and remove options.

goanpeca:
>On the other hand, the primary advantage I see of not letting users customize them is that the colors can adapt automatically to the theme, but perhaps adding a checkbox along the lines of Use automatic VCS colors from theme, checked by default, would take care of that (while unchecking would allow manual modification as currently).

No @CAM-Gerlach. We really should not expose this. This is very specific and just adds more clutter to Spyder.

CAM-Gerlach:
@goanpeca I was saying it shouldn't be a unilateral decision, without a clearly stated rationale. But given the above one is likely to be shared by @ccordoba12 as well, I suppose there isn't much point in discussing it.

With that out of the way, I recommend sharing the same specific color shades as the object type colors in the Variable Explorer in the dark and light themes, preferably sharing some common source. Do you have another idea in mind?

goanpeca:
>With that out of the way, I recommend sharing the same specific color shades as the object type colors in the Variable Explorer in the dark and light themes, preferably sharing some common source

At some point we will need to decouple styles for general purpose use. But in the meantime those colors seem fine.

ccordoba12:
@CAM-Gerlach, I agree with @goanpeca here. Letting users to configure colors would add too much configurability to our Preferences, something that I'm always being opposed to.

Besides, it's really out of the scope of this PR.

CAM-Gerlach:
@alexschw I'll review your PR in more detail when I get back (and can test on my Linux box), but at least on Windows, I cannot get any of the colors to appear other than the blue "modified" one (I tested all of them except for "merge conflict"—creating, modifying and repeatedly saving a new untracked file with Spyder, viewing existing files that should be ignored, and adding files with git).

Your tests are [failing on Python 2.7](https://travis-ci.org/spyder-ide/spyder/jobs/510133482) on all tested platforms due to an apparent Unicode coercion issue (you can easily create a Py27 env with Anaconda to debug that, if needed). Meanwhile, on [Windows, they are failing](https://dev.azure.com/spyder-ide/spyder/_build/results?buildId=794) due to the colors being different from expected; my guess would be either a different ordering/offset than on the other platforms, or a platform-specific issue with VCS statuses being read differently (which, so far, isn't immediately apparent in my testing). If you don't have access to a Windows machine to test the latter on, I can run them on my Windows and Linux boxes and try to give you some insight into what's going on.

Given what has been discussed, you may as well just revert your changes to the ``Projects`` preferences pane entirely (since there's no point in having a whole pane for just one preference, and @goanpeca was going to need to redo it anyway), and instead add the option ``Show git/hg status`` in the Project Explorer options menu, where it should have been originally (I use ``git/hg`` instead of ``VCS`` since the average Spyder user, not being a full on software engineer, is much less likely to be familiar with the latter term). Given (in my initial testing thus far) the performance issues appear to be essentially resolved, we can enable it by default.

@ccordoba12 Right, I figured you'd agree with that rationale, so I didn't bother continuing the discussion further. To be fair, I explicitly stated well before they were added that I didn't think configurable colors needed to be in scope here; I only brought them up because they'd already been implemented.

CAM-Gerlach:
Hey @alexschw , any progress on this PR? This would be a really nice feature to have, as as far as I'm aware, the only major thing this needs (assuming the performance problems have been resolved) is fixing the tests and any remaining issues in my old reviews, and then I can do a final review. Thanks!

ccordoba12:
@goanpeca, please help @alexschw to finish this PR.

goanpeca:
@ccordoba12 I am delaying this work after the projects refactoring, otherwise I think it would be double work.

ccordoba12:
Wouldn't it be better to finish it now? I mean, I can ask another developer to finish it, but it's your call.

alexschw:
Could you review this again @CAM-Gerlach? Maybe it can still be included in spyder4.

ccordoba12:
@alexschw, thanks a lot for your work but unfortunately this won't be included in Spyder 4 because we're in bugfixing mode now.

However, it'll be one of the first PRs to be merged for Spyder 5 (to be released in the next six to eight months).

CAM-Gerlach:
@alexschw I'm no longer a Spyder core developer, so unfortunately I cannot, besides for the reasons @ccordoba12 explained.

@ccordoba12 Since I'd been wondering the same, do you have a rough idea at what point will you branch `4.x` and `master` will mean Spyder 5, so this PR can be merged? As of the Spyder 4 final release, before, or after? Somewhat tangential, but will that also be the point at which new code will no longer need to be backward compatible with Python 2, you can drop Py2 CIs on `master` and we can help strip out the Python 2 cruft from the codebase?

goanpeca:
>As of the Spyder 4 final release, before, or after?

After

> you can drop Py2 CIs on master and we can help strip out the Python 2 cruft from the codebase?

I will be handling the Python 3 migration. The plan is to update py3compat so that only python3 is used. The plan will not be to remove code as that takes time, effort and can lead to subtle bugs.

CAM-Gerlach:
@goanpeca Replied to you on Gitter to avoid going too far off topic.

alexschw:
@CAM-Gerlach Okay, thank you anyways.
@ccordoba12 Oh, damn I am a few days to late... ;) Could you assign another reviewer than?

CAM-Gerlach:
@alexschw I'm neither assigned to this PR nor do I have a pending review for it. As you can see @dalthviz was assigned to complete this for Spyder 4, and either he, @goanpeca or @ccordoba12 can review it, all of whom are already subscribed to this PR.

alexschw:
okay. I though a reviewer has to be requested somewhere. I just saw, that you are the only one listed at the top under reviewers. But that is just due to you having reviewed this PR in the past. Thank you for the clarification.

ccordoba12:
@dalthviz, please give this one a review when you have some free time.

alexschw:
I am not sure if this is needed anymore, as @trollodel has made a very good contribution. It fits and exceeds my vision of this PR. I would be fine with closing this.

trollodel:
@alexschw I'm not sure what are the intentions about #13562, but if it will be merged at some point, IMO this PR is still valid. It requires a bit of rework to use the VCS API I made in #13562.

trollodel:
@alexschw (and anyone interested) you may be interested in https://github.com/spyder-ide/spyder-vcs. Feel free to join us there.

PR #8419, opened 2018-12-12
PR: Increase width of the ThumbnailScrollBar and add render message

### LINKED ISSUE #8320: Some issues with the Plots plugin
While working on PR #8192, I discovered a couple of problems with the Plots plugin:

1. Miniatures are not shown on Python 2. I went back to the PR where the plugin was introduced and verified that the problem was present from the beginning.

2. The Save and Remove buttons on the miniatures scrollbar are not shown correctly in the dark theme:

    seleccion_003

    @jnsebgosselin, I think you crafted things very well for Fusion but did not considered other themes ;-)

3. By default plots are shown in the IPython console *and* the Plots plugin. I think we should change that and instead:

    1. Print a message in the console the first time, saying that plots are now rendered in the Plots plugin.
    2. This message should also mention how to get the previous behavior.

@jnsebgosselin, could you help us with this one? Thanks!

dalthviz:
## Description of Changes

Improvements discussed for the Plots plugin
- [x] Increase width if the thumbnail scrollbar
imagen

- [x] Make `Mute inline plotting` the default behavior and add message explain it.
imagen

- [x] Fix `test_varexp_magic_dbg` that checks for a plot in the console (and other tests if any with similar checks).
- [x] Fix thumbnail scrollbar in Python 2. 

### Issue(s) Resolved

Fixes #8320 

### Affirmation

### DIFF
--- a/spyder/plugins/plots/widgets/figurebrowser.py
+++ b/spyder/plugins/plots/widgets/figurebrowser.py
@@ -11,6 +11,7 @@ This is the main widget used in the Plots plugin
 """
 
 # ---- Standard library imports
+from __future__ import division
 import os.path as osp
 
 # ---- Third library imports
--- a/spyder/plugins/plots/widgets/figurebrowser.py
+++ b/spyder/plugins/plots/widgets/figurebrowser.py
@@ -456,7 +456,7 @@ class ThumbnailScrollBar(QFrame):
         scrollarea = self.setup_scrollarea()
         up_btn, down_btn = self.setup_arrow_buttons()
 
-        self.setFixedWidth(135)
+        self.setFixedWidth(150)
         layout = QVBoxLayout(self)
         layout.setContentsMargins(0, 0, 0, 0)
         layout.setSpacing(0)
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -150,6 +150,9 @@ def main_window(request):
     # Tests assume inline backend
     CONF.set('ipython_console', 'pylab/backend', 0)
 
+    # Test assume the plots are rendered in the console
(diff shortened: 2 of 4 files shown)


spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -456,7 +456,7 @@ def setup_gui(self):
         scrollarea = self.setup_scrollarea()
         up_btn, down_btn = self.setup_arrow_buttons()
 
-        self.setFixedWidth(135)
+        self.setFixedWidth(150)
```
jnsebgosselin: I would like not to hardcode a value here anymore. I will come back with a solution so that you don't waste time on this.

CAM-Gerlach:
@jnsebgosselin Right, that's not grammatical. Try this:

> Figures now render in the Plots pane by default. To make them also appear inline in the Console, uncheck "Mute Inline Plotting" under ??????

(I don't know an exact description of where the option can be found, since I don't see it in the screenshot. In the options menu of the plots pane? The context menu? The main Spyder preferences?)

jnsebgosselin:
> (I don't know an exact description of where the option can be found, since I don't see it in the screenshot. In the options menu of the plots pane? The context menu? The main Spyder preferences?)

It is in the cog menu of the Plots Plugin.

CAM-Gerlach:
> cog menu

Its apparently now going to become a "hamburger menu" with the new icon, so we'll need to change that everywhere...

spyder/plugins/plots/widgets/figurebrowser.py
jnsebgosselin: Off course, I have no problem with that, please go ahead.

ccordoba12: Ok, thanks!

PR #8426, opened 2018-12-12
[WiP] PR: Add a menu entry to select between available color schemes 

### LINKED ISSUE #8317: Feature Request: Keyboard Shortcut for Dark/Light background
### Issue Report Checklist

* **N/A** Could not reproduce inside ``jupyter qtconsole`` (if console-related)
* **N/A** Tried basic troubleshooting (if a bug/error)

## Problem Description

I generally prefer working with a dark background on the editor (I use the Spyder Dark colour scheme).  However, when working with others or showing work on a big screen / projector, it's useful to have a quick method of switching to a light background (e.g. Spyder colour scheme).  It would be really nice if a keyboard shortcut could be set up to switch between two specified colour schemes.

1. Go to keyboard shortcuts configuration page
2. Search in vain for anything colour scheme related
3. 

## Versions

* Spyder version: 3.3.1
* Python version: 3.7.1
* Qt version: 5.6.2
* PyQt version: 5.6

(P.S. If you're suggesting in the template that people go to Help...About Spyder for this information, it would be really nice if it was copy-able for pasting in here rather than having to type it and risk making a mistake)

### Dependencies

```
IPython >=4.0     :  7.1.1 (OK)
cython >=0.21     :  0.29 (OK)
jedi >=0.9.0      :  0.13.1 (OK)
matplotlib >=2.0.0:  2.2.2 (OK)
numpy >=1.7       :  1.15.3 (OK)
qtconsole >=4.2.0 :  4.4.2 (OK)
sphinx >=0.6.6    :  1.8.1 (OK)
sympy >=0.7.3     :  1.3 (OK)
```

chends888:
## Description of Changes
I added two options in the 'Source menu' for alternating between two schemes, Spyder and Spyder Dark. This is a bit hardcoded, I could improve (like making the two opting merge into one, making spyder check the current scheme and applying the oposite color) in case reviewers get interested  :D

spyderissue8317

### Issue(s) Resolved

Fixes #8317 

### Affirmation

I certify the above statement is true and correct: Chends888

### DIFF
--- a/spyder/plugins/editor/plugin.py
+++ b/spyder/plugins/editor/plugin.py
@@ -1082,6 +1082,17 @@ class Editor(SpyderPluginWidget):
                                       _("Remove trailing spaces"),
                                       triggered=self.remove_trailing_spaces)
 
+        darkscheme_action = create_action(self,
+                                      _("Change scheme to dark mode"),
+                                      tip=_("Set editor scheme to "
+                                      "'Spyder Dark'"),
+                                      triggered=self.change_scheme_todark)
+
+        whitescheme_action = create_action(self,
+                                      _("Change scheme to white mode"),
+                                      tip=_("Set editor scheme to 'Spyder'"),
+                                      triggered=self.change_scheme_towhite)
+
         # Checkable actions
         showblanks_action = self._create_checkable_action(
                 _("Show blank spaces"), 'blank_spaces', 'set_blanks_enabled')
@@ -1231,6 +1242,8 @@ class Editor(SpyderPluginWidget):
                                show_classfunc_dropdown_action,
                                showcode_analysis_pep8_action,
                                trailingspaces_action,
+                               darkscheme_action,
+                               whitescheme_action,
                                fixindentation_action,
                                MENU_SEPARATOR,
                                self.todo_list_action,
@@ -2416,6 +2429,18 @@ class Editor(SpyderPluginWidget):
         self.switch_to_plugin()
         editorstack = self.get_current_editorstack()
         editorstack.fix_indentation()
+
+    @Slot()
+    def change_scheme_todark(self):
+        "Changes current editors to 'Spyder Dark' scheme"
+        for editorstack in self.editorstacks:
+                editorstack.set_color_scheme('spyder/dark')
+
+    @Slot()
+    def change_scheme_towhite(self):
+        "Changes current editors to 'Spyder' scheme"
+        for editorstack in self.editorstacks:
+                editorstack.set_color_scheme('spyder')
 
     #------ Cursor position history management
     def update_cursorpos_actions(self):


jnsebgosselin:
Thank you very much for your PR!

However, I think @ccordoba12 idea's to fix this was to "add an entry to the Source menu with all color schemes available so users can easily select from them." Not 2 entries for switching between a dark or light scheme.

ccordoba12:
> However, I think @ccordoba12's idea was to "add an entry to the Source menu with all color schemes available so users can easily select from them", not to add an option for switching between a dark or light scheme.

Yes, that was my suggestion because there's no single dark or light color scheme we could go to.

chends888:
Nice, thanks for the feedback, I can work on that the next few days if you're interested on seeing a list of current schemes on the _Source Menu_. Just let me know if it'll really be useful for Spyder

CAM-Gerlach:
As others have noted, realistically the way to implement this is a submenu in the Source menu showing all available color schemes. However, I'm not sure if it'd really be worth doing, in terms of addressing the use case requested by @abudden , which was a quick way to switch between light and dark syntax themes. 

First, the user would still have to open the menu, navigate to the submenu, and scroll through the (soon to be much longer, thanks to #8381 ) list of schemes to find the one they want, and do the same thing to switch back; meanwhile, tapping ``Ctrl-Alt-Shift-P`` to summon preferences, selecting the scheme from a menu and hitting apply doesn't take all that much longer.

This is particularly true in light of the (at least on my machine) 20-30 second delay needed to apply the scheme, so the at most a few seconds it would save are inconsequential, to say nothing of the full restart required to actually switch from the light to the dark mode and back (which would presumably be the case if the user switched from a light to a dark theme as the original reporter wanted to with the UI Theme set to ``Automatic``, unless this has special logic to override that (which could even cause user confusion later). Therefore, unless the huge time lag can be resolved (even cutting it by 1/2 would save more time than a submenu vs. preferences), I'm not sure I understand how this really adds value for the stated use case.

jnsebgosselin:
> Nice, thanks for the feedback, I can work on that the next few days if you're interested on seeing a list of current schemes on the Source Menu. Just let me know if it'll really be useful for Spyder

Just to be sure, what we are talking about is adding a single subemenu named *Color schemes* to the *Source* menu, not adding the complete list directly to the *Source* menu.

I think this is worth it and would be a nice addition to Spyder if you have the time to do it.

CAM-Gerlach:
> adding a single subemenu named Color schemes to the Source menu

That was my impression as well; adding all the syntax highlighting schemes (especially after #8381 ) would might not even fit on the screen (at least for 720p). I guess my objection is I don't see the point of a feature that saves negligible time relative to the switching overhead and adds edge cases and therefore potential bugs (needs to handle Color Scheme Automatic and switching from light to dark, stay in sync with prefs changes, etc) , when e.g. optimizing the scheme switching routine to not take 20-30 seconds is likely going to save far more time for users than a few seconds opening preferences. However, there isn't *really* a harm in it if implemented correctly, other than adding one more item in an already very messy Source menu.

Also, if we do it, it should be named ``Syntax highlighting scheme`` for clarity, consistency and idiomatic correctness.

jnsebgosselin:
@CAM-Gerlach The relevance of this feature was kind of already established in the associated Issue. Your arguments are admissible though, so maybe they could be synthesized a little and added to the Issue, so that we can try to keep the PR focused more on the implementation details instead.

CAM-Gerlach:
> The relevance of this feature was kind of already established in the associated Issue. 

My argument is that this _implementation_ (submenu for all color schemes in the source menu, no shortcuts) as it stands doesn't solve the issue discussed in  #8381 , wanting to be able to more quickly toggle between dark and light syntax themes, unless it can reduce the massive delay (~10 seconds on a clean prefs file, 20-30s on a dirty one, if the results on my i7/4core/8thread/24GB/SSD/Windows machine are representative) needed to apply a theme. Furthermore, in my testing, opening the preferences with the existing shortcut, selecting the theme, and hitting enter/okay took essentially the same time as navigating to the Source menu, moving to the submenu, finding the theme in the (soon to be much longer) list (I simulated it with an appropriate length warnings/errors list), and clicking it both took between 5-6 seconds per attempt, averaged over 3 attempts each with one practice run (excluding the compute time to apply the syntax theme). Therefore, I assert that it seems this implementation as it stands now (without shortcuts, pruning the displayed theme list to the ~5 most recently used themes, or speeding up the time needed to apply a theme) does not seem to fulfill the quite valid use case outlined in the issue, and @ mentioned the requester of the feature so they could comment for themselves as to whether it does.

jnsebgosselin:
I quite got your point the first time you posted thank you. Repeating the same argument over and over again with that level of unnecessary details is a waste of my time and yours and is quite irritating.

The very idea you are objecting to was proposed by @ccordoba12 in Issue #8317 as a valid solution to solve the problem and **no one** has objected to that over there... So this PR should have been very straightforward, with maybe a suggestion or two to improve it, like your idea of list pruning. The only reason I posted here in the first place was because the implementation was not respecting what it was agreed and discussed in the issue.

ccordoba12:
> What will be very helpful is a convenient way to switch between dark mode and light mode by user base on the time or the environment. Maybe one can select one color scheme as the default light mode and one as the default dark mode.

The idea is kind of interesting but I don't know how to implement it graphically and how other people (who doesn't know the context or the purpose of that) would understand the meaning of *default* light and dark themes.

----

In all fairness (and with my sincere apologies to @chends888), I think we should stop here and not implement this feature (at least with my initial suggestion) given @CAM-Gerlach's and @gepcel's remarks.

Besides, @gepcel's initial remark was that this would be useful to do presentations. So, unless you give presentations like crazy, I think it's simpler and more intuitive to simply go to Preferences and select your scheme there.

jnsebgosselin:
> I think we should stop here and not implement this feature (at least with my initial suggestion) given @CAM-Gerlach's and @gepcel's remarks.

I agree with this. I think Issue #8317 is still a valid one though and we should all go back brainstorming over there. Maybe we can find a better solution.

jnsebgosselin:
Thank you @CAM-Gerlach, it is much appreciated!

ccordoba12:
Sorry for making you waste your time @chends888.

PR #8429, opened 2018-12-13
PR: Set projet filenames correctly when closing a project

### LINKED ISSUE #8375: Spyder editor does not open previously opened files in a project

jnsebgosselin:
## Description of Changes

When closing a project, call `set_project_filenames` before setting the current project to `None`.

### Issue(s) Resolved

Fixes #8375

### Affirmation


jnsebgosselin:
@ccordoba12 Is this worth it to add a test for such a small fix? Would you?

jnsebgosselin:
Forget my last comment, I found a way to cover this at the plugin level without too much effort.

spyder/plugins/projects/tests/test_plugin.py
```diff
@@ -120,5 +127,33 @@ def test_open_project_uses_visible_config(projects, tmpdir, value):
...
+
+    Regression test for Issue #8375
+    """
+    opened_files = ['file1', 'file2', 'file3']
+    mocker.patch.object(
+        projects.main.editor, 'get_open_filenames', return_value=opened_files)
+
+    # We mock os.path.isfile so that we do not have to
+    # actually create the files on the disk.
+    mocker.patch('spyder.plugins.projects.api.os.path.isfile',
```
CAM-Gerlach: Is this what we're using now instead of ``monkeypatch``?

spyder/plugins/projects/tests/test_plugin.py
```diff
@@ -25,16 +25,25 @@
 from spyder.py3compat import to_text_string
 
 
-# =============================================================================
-# Fixtures
-# =============================================================================
+# ---- Fixtures
```
CAM-Gerlach: ??? I thought this was our standard test file formatting for new and substantially revised files? I've never seen the ``# ---- NAME`` formatting anywhere else in our code (or anywhere else that I've seen) while it seems the previous is much more conventional (and besides, looks way better and clearer IMO).

spyder/plugins/projects/tests/test_plugin.py
```diff
@@ -57,9 +66,7 @@ def __getattr__(self, attr):
     return projects
 
 
-# =============================================================================
-# Tests
-# =============================================================================
+# ---- Tests
```
CAM-Gerlach: Same as above

spyder/plugins/projects/tests/test_plugin.py
```diff
@@ -120,5 +127,33 @@ def test_open_project_uses_visible_config(projects, tmpdir, value):
...
+    mocker.patch('spyder.plugins.projects.api.os.path.isfile',
+                 return_value=True)
+
+    # Create the project.
+    path = to_text_string(tmpdir.mkdir('project1'))
+    projects.open_project(path=path)
+    assert projects.get_project_filenames() == []
+
+    # Close and reopen the project.
+    projects.close_project()
```
CAM-Gerlach: Although the original bug only occurred on closing and re-opening projects, would it be too much work to test switching projects as well in case there's a regression or edge case in the future?

spyder/plugins/projects/tests/test_plugin.py
ccordoba12: I agree with @CAM-Gerlach here. Please revert.

spyder/plugins/projects/tests/test_plugin.py
ccordoba12: I agree with this too. It shouldn't be that hard.

spyder/plugins/projects/tests/test_plugin.py
jnsebgosselin: !!! The reason I did this was because special comments are added to the Outline Explorer, which I find very useful to separate at a glance the fixtures from the tests. But that's ok, I'm going to revert it.

image

ccordoba12: Sorry, then I agree with your change. Please remove your last commit.

spyder/plugins/projects/tests/test_plugin.py
jnsebgosselin: This is a good idea. I will add that.

ccordoba12:
@jnsebgosselin, you need to rebase/merge with master to fix the tests in Azure.

spyder/plugins/projects/tests/test_plugin.py
CAM-Gerlach: I don't know, that was just my understanding from the previous files but its up to @ccordoba12

spyder/plugins/projects/tests/test_plugin.py
CAM-Gerlach: What about

```python
# =============================================================================
# ---- Fixtures
# =============================================================================
```

That keeps most of the existing convention (and more visible/better looking styling) and only requires adding four characters on one line to existing files as they are modified, and is more visible inline without the outline explorer while still achieving the same affect with it.

jnsebgosselin: 👍 👍 👍 I think this is a good idea. If @ccordoba12 agrees with it I will do that.

ccordoba12: Sure, I agree with that too.

spyder/plugins/projects/tests/test_plugin.py
ccordoba12: I don't have a problem with using any approach.

jnsebgosselin:
@ccordoba12 I added the requested additional test. There is a chance you won't like my `create_projects` fixture that I've created to remove some repetitiveness of the code in both tests I've added.

There is no problem if you don't like it, I will just remove the 2 latest commits.

spyder/plugins/projects/tests/test_plugin.py
```diff
@@ -120,5 +156,49 @@ def test_open_project_uses_visible_config(projects, tmpdir, value):
...
+    then reopened.
+
+    Regression test for Issue #8375
+    """
+    opened_files = ['file1', 'file2', 'file3']
+    path = to_text_string(tmpdir.mkdir('project1'))
+
+    # Create the projects plugin.
+    projects = create_projects(path, opened_files)
+    assert projects.get_project_filenames() == []
```
ccordoba12: Why is this empty and not equal to `opened_files`?

jnsebgosselin: `get_project_filenames` returns what is saved in the project's config file and this gets updated only when closing the project, not as new files are opened or closed in the Editor.

jnsebgosselin: So `create_project` open a new project with an empty list of files at first in the config. Then when we close the project, the list of files we are mocking as open in the Editor are set as the project filenames in the config file.

ccordoba12: Ok, I see. Thanks for the explanation!

PR #8452, opened 2018-12-17
PR: Fix Recent Project submenu actions in Projects menu.

### LINKED ISSUE #8450: Project switching with recent projects menu fails to open correct project

Using the Recent Projects menu always reopens the last project opened.  This occurs even after a restart of Spyder.  It seems to occur regardless of the project selected from the menu, so doesn't seem immediately to be a corruption of the file where the recent projects list is stored.  

It is easily reproducible for me using the current state of the master branch, and has been for about a month.  I am not sure which PR introduced the problem for me at this time.

## Versions

* Spyder version: 4.0.0.dev0 
* Python version: 3.7.0
* Qt version: 5.9.6
* PyQt5 version: 5.9.2

### Dependencies

```
qdarkstyle >=2.6.4:  2.6.4 (OK)
sphinx >=0.6.6    :  2.0.0 (OK)
pyls >=0.19.0     :  0+unknown (OK)
numpy >=1.7       :  1.15.2 (OK)
sympy >=0.7.3     :  1.3 (OK)
qtconsole >=4.2.0 :  4.4.2 (OK)
IPython >=4.0     :  7.0.1 (OK)
matplotlib >=2.0.0:  3.0.1 (OK)
```

jnsebgosselin:
## Description of Changes

I followed the solution proposed https://stackoverflow.com/a/46300450/4481445 to solve this Issue. 

Basically, it is required to use the lambda expression with solid variables, or else the path to the oldest project saved in the most recent project list is used as the `path` value for the action of every project.

### Issue(s) Resolved

Fixes #8450

### Affirmation

### DIFF
--- a/spyder/plugins/projects/plugin.py
+++ b/spyder/plugins/projects/plugin.py
@@ -195,15 +195,13 @@ class Projects(SpyderPluginWidget):
             for project in self.recent_projects:
                 if self.is_valid_project(project):
                     name = project.replace(get_home_dir(), '~')
-
-                    def slot():
-                        self.switch_to_plugin()
-                        self.open_project(path=project)
-
-                    action = create_action(self,
+                    action = create_action(
+                        self,
                         name,
-                        icon = ima.icon('project'),
-                        triggered=slot)
+                        icon=ima.icon('project'),
+                        triggered=(
+                            lambda _, p=project: self.open_project(path=p))
+                        )
                     self.recent_projects_actions.append(action)
                 else:
                     self.recent_projects.remove(project)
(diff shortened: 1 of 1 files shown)


spyder/plugins/projects/plugin.py
```diff
@@ -195,15 +195,13 @@ def setup_menu_actions(self):
             for project in self.recent_projects:
                 if self.is_valid_project(project):
                     name = project.replace(get_home_dir(), '~')
-
-                    def slot():
-                        self.switch_to_plugin()
```
jnsebgosselin: We do not need to call this here, since it is the first thing that is called in `open_project`

jnsebgosselin:
@ccordoba12 for your information, this bug should be present also in 3.x, but I think we are done fixing minor bugs in this branch, right?

ccordoba12:
> for your information, this bug should be present also in 3.x, but I think we are done fixing minor bugs in this branch, right?

No, we aren't anymore.

PR #8460, opened 2018-12-19
PR: Handle non-existing saved editor layout when loading projects

### LINKED ISSUE #8458: Spyder won't load previously open files from any project with a clean prefs file
## Problem Description

As originally reported in #8415 and thought to be due to that ~~issue~~ PR (I couldn't repro it in ``master`` at the time), and experienced by the PR contributor as well. From a clean prefs file (``--safe-mode``), I get the below error when trying to open any project; the project opens but without its previous files open and a completely blank Editor. If I then close the project, the previously recorded open files are wiped. This occurs both on opening a project with, and without one open, and on both new and previous projects. Also, it even happens on closing a project if there were any Editor files open previously, since Spyder will run into the same error trying to restore the previously open files in the Editor.

1. Open Spyder in ``--safe-mode``
2. Open an existing project with files in the workspace, or create, save and re-open a new one

```python-traceback
  File "C:\Users\C. A. M. Gerlach\Documents\dev\SpyderDev\spyder\spyder\plugins\projects\plugin.py", line 147, in <lambda>
    lambda v: self.main.editor.setup_open_files())
  File "C:\Users\C. A. M. Gerlach\Documents\dev\SpyderDev\spyder\spyder\plugins\editor\plugin.py", line 2932, in setup_open_files
    is_vertical, cfname, clines = layout.get('splitsettings')[0]
AttributeError: 'NoneType' object has no attribute 'get'
```

## Proposed Solution

It seems to be that the ``layout_settings`` option for the Editor is either ``None`` or not present, which results in ``layout`` being ``None``. I already implemented and tested an if/else around the problem section that required ``layout``  to fix this, loading the files with the default focuswhich resulted in the projects and their saved workspaces being opened, switched, saved and closed successfully. I can go ahead with a PR, but is my understanding correct? Is there a better way to handle/implement this?

## Versions

* Spyder version: master
* Python version: 3.6.7
* Qt version: 5.9.6
* PyQt version: 5.9.2

CAM-Gerlach:
## Description of Changes

When opening a project with a clean preferences file, the project's previously opened files fail to load into the Editor, an exception is raised, and is the project is then closed or switched, the previously opened files are wiped; this even happens when closing a project and trying to load the prevously open files in the Editor. Therefore, here we check that the relevant preferences item exists before trying to get the saved layout (focused file, file position, etc) and if it no layout information is remembered, just load the files without restoring the focused line and file.

In functional testing both switching projects and opening and closing them, no errors were generated and all behavior was expected when examining opening, saving and restoring the correct files.

I used an if block ("LBYL") instead of a try/except block ("EAFP") since the former is what other blocks in the function used, and in this context an ``AttributeError`` due to something other than ``None`` being returned, if just caught and passed is likely indicative of changes in the config system rather than anything on the user end and would silently break the editor remembering its last focused file, position, split, etc. without any indication to us.

This also adds a test for this behavior, as well as refactors some of the existing tests into a fixture to avoid a significant amount of duplicated code.

### Issue(s) Resolved

Fixes #8458

### Affirmation

### DIFF
--- a/spyder/plugins/editor/plugin.py
+++ b/spyder/plugins/editor/plugin.py
@@ -2904,25 +2904,31 @@ class Editor(SpyderPluginWidget):
 
         if filenames and any([osp.isfile(f) for f in filenames]):
             layout = self.get_option('layout_settings', None)
-            is_vertical, cfname, clines = layout.get('splitsettings')[0]
-            if cfname in filenames:
-                index = filenames.index(cfname)
-                # First we load the last focused file.
-                self.load(filenames[index], goto=clines[index], set_focus=True)
-                # Then we load the files located to the left of the last
-                # focused file in the tabbar, while keeping the focus on
-                # the last focused file.
-                if index > 0:
-                    self.load(filenames[index::-1], goto=clines[index::-1],
-                              set_focus=False, add_where='start')
-                # Finally we load the files located to the right of the last
-                # focused file in the tabbar, while keeping the focus on
-                # the last focused file.
-                if index < (len(filenames) - 1):
-                    self.load(filenames[index+1:], goto=clines[index:],
-                              set_focus=False, add_where='end')
+            # Check if no saved layout settings exist, e.g. clean prefs file
+            # If not, load with default focus/layout, to fix issue #8458 .
+            if layout:
+                is_vertical, cfname, clines = layout.get('splitsettings')[0]
+                if cfname in filenames:
+                    index = filenames.index(cfname)
+                    # First we load the last focused file.
+                    self.load(filenames[index], goto=clines[index],
+                              set_focus=True)
+                    # Then we load the files located to the left of the last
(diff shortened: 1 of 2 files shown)


ccordoba12:
Thanks @CAM-Gerlach!

@jnsebgosselin, could you review this one? Thanks!

jnsebgosselin:
Sure, but probably not today. I'll try to do this tomorrow.

ccordoba12:
No problem, thanks!

ccordoba12:
@CAM-Gerlach, I rebased and fixed this one.

ccordoba12:
I'll merge after everything is green.

CAM-Gerlach:
Thanks for the rebase. Did @jnsebgosselin ever review it @ccordoba12 ?

ccordoba12:
No, but I think the fix and the test are fine.

PR #8469, opened 2018-12-20
PR: Make indentation guides to go up to last line with text

### LINKED ISSUE #7339: Indent guides going too far down
Another 2c bug.

In my opinion indent guides are going too far.

image

Even worst

image

Thanks.

Khalilsqu:
This resolves the issue Indent guides going too far down #7339

## Description of Changes

This PR solves the issue https://github.com/spyder-ide/spyder/issues/7339

Sample code before the fix
spyder_issue_indentation_before

Sample code after the fix
spyder_issue_indentation

### Issue(s) Resolved

Fixes #7339

### Affirmation

### DIFF
--- a/spyder/plugins/editor/panels/indentationguides.py
+++ b/spyder/plugins/editor/panels/indentationguides.py
@@ -57,7 +57,8 @@ class IndentationGuide(Panel):
             end_of_sub_fold = block
             if last_line:
                 block = block.document().findBlockByNumber(last_line)
-                while block.blockNumber() and block.text().strip() == '':
+                while block.blockNumber() and block.text().strip() == '' or
+                    block.text().strip().startswith('#')):
                     block = block.previous()
                     last_line = block.blockNumber()
 
--- a/spyder/plugins/editor/panels/indentationguides.py
+++ b/spyder/plugins/editor/panels/indentationguides.py
@@ -36,6 +36,8 @@ class IndentationGuide(Panel):
         color = QColor(self.color)
         color.setAlphaF(.5)
         painter.setPen(color)
+        offset = self.editor.document().documentMargin() + \
+            self.editor.contentOffset().x()
 
         for _, line_number, block in self.editor.visible_blocks:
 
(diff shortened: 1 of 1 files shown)


ccordoba12:
@Khalilsqu, thanks a lot for your contribution! Unfortunately you took code from Ninja-IDE, whose license is incompatible with ours.

So you need to re-implement the bi yout took from that project with code from your own.

Khalilsqu:
Hi @ccordoba12 Thanks for your feedback. I updated the code. The offset parameter is removed. It is purpose is to make the indentation guides lines perfectly below the first letter of the block scope, below words def, for, if, and etc. The code still fixes the issue #7339 . This line of code actually does not belong to Ninja IDE but is is used in many cases.  I think it appeared in pyQode.

1.
https://github.com/pyQode/pyqode.core/blob/53448cadc388dc42a05c408a97d6bedbc316ffb6/pyqode/core/modes/right_margin.py#L78
2.
https://github.com/pyzo/pyzo/blob/559d1720dcd19742dd503a50c9f8e3f2fcbe25c4/pyzo/codeeditor/extensions/appearance.py#L499

I included Ninja since I found it first there. However, I you would like to include it back the reference from pyQode might be used since their license is compatible with Spyder.

ccordoba12:
> However, If you would like to include it back the reference from pyQode might be used since their license is compatible with Spyder

We took a lot of code from PyQode and we already make reference to it, so don't worry about it.

spyder/plugins/editor/panels/indentationguides.py
```diff
@@ -36,14 +37,47 @@ def paintEvent(self, event):
...
+            top = int(self.editor.blockBoundingGeometry(block).translated(
+                self.editor.contentOffset()).top())
             bottom = top + int(self.editor.blockBoundingRect(block).height())
 
             indentation = TextBlockHelper.get_fold_lvl(block)
 
             for i in range(1, indentation):
-                x = self.editor.fontMetrics().width(i * self.i_width * '9')
-                painter.drawLine(x, top, x, bottom)
+
```
ccordoba12: Please remove this line.

spyder/plugins/editor/panels/indentationguides.py
```diff
@@ -36,14 +37,47 @@ def paintEvent(self, event):
...
+                if (line_number > last_line and
+                        TextBlockHelper.get_fold_lvl(end_of_sub_fold.next())
+                        <= i + 1):
+                    continue
+
```
ccordoba12: Please remove this blank line.

Khalilsqu:
The commit `correction to avoid line breaking` avoid line breaking in case the block level is continued after a nested inner block.

before correction 
nested complex loop

After correction
spyder_after_corr

Khalilsqu:
I have one suggestion. However, if it is up to you to accept it or find the current case as the best.

I would to include indentation guide line for the first block level (base level).
image

The current implementation skips indentation level zero.
image

I find it visually more appealing to include all levels. But, again it up to you to decide? 

Thanks you for your time and feedback improving this PR.

PR #8481, opened 2018-12-22
PR: Add cursor position bookmarks in Editor

### LINKED ISSUE #1759: sypder support bookmarks?
_From [my....@163.com](https://code.google.com/u/107821550222058971796/) on 2014-03-20T01:32:29Z_

I use pyscripter before, it can shift + ctrl + num set bookmarks and ctrl + num go to bookmark, very convenient, can sypyder also add such features?

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=1759_

thepycoder:
Functionality until now is as follows:
A user can either set or load a bookmark using a keyboard shortcut. By default it's ctrl+shift+number to save and ctrl-shift-number to load. 

By saving, the column number and slot to which the bookmark has been saved (the pressed number after the ctrl-shift) are saved in the blockUserData of the textblock the cursors is currently in. This way it can keep track of the bookmarks line number when it's position changed due to adding text earlier in the file.

By loading, a user can snap their cursor back to a certain location in a certain file, opening that file if necessary. The saved positions are persistent after a restart.

As for now all of the necessary groundwork for this to work should be in place. There is no GUI part yet tough, apart from the keyboard shortcuts, it's invisible.

## Description of Changes

* [X] Wrote at least one-line docstrings (for any new functions)
* [X] Added unit test(s) covering the changes (if testable)

### Issue(s) Resolved

Related to issue #910, but a UI still needs to be built to show bookmarks to the user.

### Affirmation

I certify the above statement is true and correct: ThePyCoder


thepycoder:
Reposting this PR because of switched account mishap on last PR. I'm sorry for the inconvenience. I'll push a new commit based on the code commentary of the last PR soon!

spyder/plugins/editor/widgets/editor.py
```diff
@@ -694,6 +696,126 @@ def create_shortcuts(self):
                                       context="Editor",
                                       name="close split panel",
                                       parent=self)
+        save_bookmark_1 = config_shortcut(lambda:
+                                          self.sig_save_bookmark.emit(0),
+                                          context="Editor",
+                                          name="bookmark current position "
+                                               "to slot 0",
+                                          parent=self)
```
CAM-Gerlach: This is very un-DRY code. Can you make these two into list comprehensions or for loops instead, and then just concatenate the resulting list to the one being returned at the end?

spyder/config/main.py
```diff
@@ -435,6 +435,26 @@
               'editor/split vertically': "Ctrl+{",
               'editor/split horizontally': "Ctrl+_",
               'editor/close split panel': "Alt+Shift+W",
+              "editor/bookmark current position to slot 0": "Ctrl+Shift+0",
```
CAM-Gerlach: Can you greatly shorten this, e.g. just ``Save bookmark N``

spyder/config/main.py
```diff
@@ -435,6 +435,26 @@
...
+              "editor/bookmark current position to slot 1": "Ctrl+Shift+1",
+              "editor/bookmark current position to slot 2": "Ctrl+Shift+2",
+              "editor/bookmark current position to slot 3": "Ctrl+Shift+3",
+              "editor/bookmark current position to slot 4": "Ctrl+Shift+4",
+              "editor/bookmark current position to slot 5": "Ctrl+Shift+5",
+              "editor/bookmark current position to slot 6": "Ctrl+Shift+6",
+              "editor/bookmark current position to slot 7": "Ctrl+Shift+7",
+              "editor/bookmark current position to slot 8": "Ctrl+Shift+8",
+              "editor/bookmark current position to slot 9": "Ctrl+Shift+9",
+              "editor/go to bookmarked position in slot 0": "Ctrl+Alt+0",
```
CAM-Gerlach: Similarly, how about ``Go to bookmark N``?

spyder/plugins/editor/widgets/editor.py
```diff
@@ -694,6 +696,126 @@ def create_shortcuts(self):
                                       context="Editor",
                                       name="close split panel",
                                       parent=self)
+        save_bookmark_1 = config_shortcut(lambda:
+                                          self.sig_save_bookmark.emit(0),
+                                          context="Editor",
+                                          name="bookmark current position "
```
CAM-Gerlach: Name as above.

spyder/plugins/editor/widgets/editor.py
```diff
@@ -694,6 +696,126 @@ def create_shortcuts(self):
...
+        save_bookmark_9 = config_shortcut(lambda:
+                                          self.sig_save_bookmark.emit(8),
+                                          context="Editor",
+                                          name="bookmark current position "
+                                               "to slot 8",
+                                          parent=self)
+        save_bookmark_10 = config_shortcut(lambda:
+                                           self.sig_save_bookmark.emit(9),
+                                           context="Editor",
+                                           name="bookmark current position "
```
CAM-Gerlach: Name as above.

spyder/plugins/editor/plugin.py
```diff
@@ -2701,6 +2706,35 @@ def re_run_last_cell(self):
...
+            self.last_edit_cursor_pos = None
+            return
+        else:
+            self.load(filename)
+            editor = self.get_current_editor()
+            if position < editor.document().characterCount():
+                editor.set_cursor_position(position)
+
+    def init_bookmarks(self):
+        """Load last saved cursor bookmarks for config file"""
```
CAM-Gerlach: from config file, presumably. Also, period at the end (per PEP 257, docstrings should always end with a period).

spyder/plugins/editor/plugin.py
```diff
@@ -2701,6 +2706,35 @@ def re_run_last_cell(self):
         editorstack = self.get_current_editorstack()
         editorstack.re_run_last_cell()
 
+    @Slot(int)
+    def save_bookmark(self, num):
+        """Save current cursor position as bookmark in slot based
+        on number of shortcut e.g Ctrl+Shift+1 for slot 1"""
```
CAM-Gerlach: Per PEP 257, docstrings should have a one-line summary, following by a blank line then additional details if necessary.

spyder/plugins/editor/plugin.py
```diff
@@ -2701,6 +2706,35 @@ def re_run_last_cell(self):
...
+        filename = self.get_current_filename()
+        position = self.get_current_editor().get_position('cursor')
+        self.bookmarks[num] = (filename, position)
+        CONF.set('editor', 'bookmarks', self.bookmarks)
+
+    @Slot(int)
+    def load_bookmark(self, num):
+        """Set cursor to saved file and position in selected slot
+        based on number of shortcut e.g Ctrl+Alt+1 for slot 1"""
```
CAM-Gerlach: Per PEP 257, docstrings should have a one-line summary, following by a blank line then additional details if necessary.

spyder/plugins/editor/widgets/editor.py
thepycoder: This should be fixed now!

spyder/config/main.py
thepycoder: I'm not exactly sure what you mean here. If I only use this line
`"editor/Save bookmark N": "Ctrl+Shift+N"`
this can't be linked to all shortcuts right?
Or do you mean to just make the string itself shorter, from
`bookmark current position to slot 0` to `Save bookmark N`?

CAM-Gerlach: I meant to make the string itself shorter, as it is far too long as is.

thepycoder: Agreed, I changed them to your suggestions.

jnsebgosselin:
Thanks for your contribution!

I only have one comment/question:

If changes are made before a bookmark in a file, should the position of the bookmark be updated? I also think that if the line which contains a bookmark is deleted/cut, the bookmark should be deleted. In fact, the behavior should be like it is actually for the `break points`.

CAM-Gerlach:
> In fact, the behavior should be like it is actually for the break points.

Exactly; you could even sharing the same code if possible for tracking the line number, with the only difference being that the bookmarks also remembers the position in the line (that's fixed, except that its truncated to the end of the line if the line is modified to contain fewer characters than the bookmark's location).

thepycoder:
I reused parts of the breakpoints as suggested. It now should have all the functionality described by @CAM-Gerlach and @jnsebgosselin above. I have been using it for some time now and it appears to work fine. However a code review is definitely welcome, I'm new to this and can't shake the feeling I could've done it more efficiently. Looking forward to your feedback!

spyder/plugins/editor/widgets/editor.py
```diff
@@ -217,6 +219,13 @@ def breakpoints_changed(self):
             self.editor.breakpoints = breakpoints
             self.save_breakpoints.emit(self.filename, repr(breakpoints))
 
+    def bookmarks_changed(self):
+        """Breakpoint list has changed"""
```
CAM-Gerlach: Probably should rename this :) Also, period.

spyder/plugins/editor/plugin.py
```diff
@@ -107,6 +107,32 @@ def clear_breakpoint(filename, lineno):
...
+def _load_all_bookmarks():
+    slots = CONF.get('editor', 'bookmarks', {})
+    for slot_num in list(slots.keys()):
+        if not osp.isfile(slots[slot_num][0]):
+            slots.pop(slot_num)
+    return slots
+
+
+def load_bookmarks(filename):
+    bookmarks = _load_all_bookmarks()
```
CAM-Gerlach: Missing docstring.

spyder/plugins/editor/plugin.py
```diff
@@ -107,6 +107,32 @@ def clear_breakpoint(filename, lineno):
...
+    return {k: v for k, v in bookmarks.items() if v[0] == filename}
+
+
+def load_bookmarks_without_file(filename):
+    bookmarks = _load_all_bookmarks()
```
CAM-Gerlach: Missing docstring.

spyder/plugins/editor/plugin.py
```diff
@@ -107,6 +107,32 @@ def clear_breakpoint(filename, lineno):
...
+    return {k: v for k, v in bookmarks.items() if v[0] != filename}
+
+
+def save_bookmarks(filename, bookmarks):
+    if not osp.isfile(filename):
```
CAM-Gerlach: Missing docstring.

spyder/plugins/editor/plugin.py
```diff
@@ -1860,6 +1889,14 @@ def save_breakpoints(self, filename, breakpoints):
         save_breakpoints(filename, breakpoints)
         self.breakpoints_saved.emit()
 
+    # ------ Bookmarks
+    def save_bookmarks(self, filename, bookmarks):
+        filename = to_text_string(filename)
```
CAM-Gerlach: Missing docstring.

spyder/plugins/editor/plugin.py
```diff
@@ -2701,6 +2739,46 @@ def re_run_last_cell(self):
         editorstack = self.get_current_editorstack()
         editorstack.re_run_last_cell()
 
+    # ------ Code bookmarks
+    @Slot(int)
+    def save_bookmark(self, slot_num):
+        """Save current line and position as bookmark"""
```
CAM-Gerlach: Missing period.

spyder/plugins/editor/plugin.py
```diff
@@ -2701,6 +2739,46 @@ def re_run_last_cell(self):
...
+        bookmarks = CONF.get('editor', 'bookmarks')
+        editorstack = self.get_current_editorstack()
+        if slot_num in bookmarks:
+            filename, line_num, column = bookmarks[slot_num]
+            if osp.isfile(filename):
+                index = editorstack.has_filename(filename)
+                if index is not None:
+                    block = editorstack.tabs.widget(index).document() \
+                        .findBlockByNumber(line_num)
```
CAM-Gerlach: Use parenthesis continuation instead.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -181,14 +181,16 @@ def __init__(self, editor, cursor=None, color=None):
...
         self.code_analysis = []
         self.todo = ''
         self.selection = cursor
         self.color = color
         self.editor.blockuserdata_list.append(self)
 
     def is_empty(self):
-        return not self.breakpoint and not self.code_analysis and not self.todo
+        return not self.breakpoint and not self.code_analysis \
+            and not self.todo and not self.bookmarks
```
CAM-Gerlach: Use parenthesis continuation instead.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -181,14 +181,16 @@ def __init__(self, editor, cursor=None, color=None):
...
+        self.bookmarks = []
         self.code_analysis = []
         self.todo = ''
         self.selection = cursor
         self.color = color
         self.editor.blockuserdata_list.append(self)
 
     def is_empty(self):
-        return not self.breakpoint and not self.code_analysis and not self.todo
+        return not self.breakpoint and not self.code_analysis \
```
CAM-Gerlach: Missing docstring.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1444,6 +1449,54 @@ def update_breakpoints(self):
         """Update breakpoints"""
         self.breakpoints_changed.emit()
 
+    # -----Code bookmarks
+    def add_bookmark(self, slot_num, line=None, column=None):
+        """Add bookmark"""
```
CAM-Gerlach: Missing period; be a little more descriptive.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1444,6 +1449,54 @@ def update_breakpoints(self):
...
+        data = block.userData()
+        if not data:
+            data = BlockUserData(self)
+        if slot_num not in data.bookmarks:
+            data.bookmarks.append((slot_num, column))
+        block.setUserData(data)
+        self.bookmarks_changed.emit()
+
+    def get_bookmarks(self):
+        """Get bookmarks"""
```
CAM-Gerlach: Missing period; be a little more descriptive.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1444,6 +1449,54 @@ def update_breakpoints(self):
...
+        for line_number in range(0, self.document().blockCount()):
+            data = block.userData()
+            if data and data.bookmarks:
+                for slot_num, column in data.bookmarks:
+                    bookmarks[slot_num] = [line_number, column]
+            block = block.next()
+        return bookmarks
+
+    def clear_bookmarks(self):
+        """Clear bookmarks"""
```
CAM-Gerlach: Missing period; be a little more descriptive.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1444,6 +1449,54 @@ def update_breakpoints(self):
...
+        for data in self.blockuserdata_list[:]:
+            data.bookmarks = []
+            if data.is_empty():
+                # This is not calling the __del__ in BlockUserData.  Not
+                # sure if it's supposed to or not, but that seems to be the
+                # intent.
+                del data
+
+    def set_bookmarks(self, bookmarks):
+        """Set bookmarks when opening file"""
```
CAM-Gerlach: Missing period.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1444,6 +1449,54 @@ def update_breakpoints(self):
...
+        self.clear_bookmarks()
+        for slot_num, bookmark in bookmarks.items():
+            self.add_bookmark(slot_num, bookmark[1], bookmark[2])
+
+    def update_bookmarks(self):
+        """Update bookmarks"""
```
CAM-Gerlach: Missing period; be a little more descriptive.

spyder/plugins/editor/widgets/editor.py
```diff
@@ -854,6 +880,12 @@ def set_or_edit_conditional_breakpoint(self):
             editor = self.get_current_editor()
             editor.add_remove_breakpoint(edit_condition=True)
 
+    def set_bookmark(self, slot_num):
+        """Set bookmark"""
```
CAM-Gerlach: Missing period; be a little more descriptive.

thepycoder:
Glad to hear it worked! I fixed the docstrings (sorry about that) and am going to write some tests probably this evening and will push everything soon.

As for the second point: I started from the structure of the breakpoints, but the end result is quite far from the breakpoints. For example:

- Breakpoints only save a boolean in a block's userData whether it has a breakpoint or not, bookmarks save both the bookmark number and the column position.
- One can have arbitrarily many breakpoints, bookmarks need a backcheck to remove a bookmark from one block and add it to another if it's reassigned. Even in another file.
- Breakpoint setting functions don't require any arguments passed through the signal, they can only be on or off. Bookmarks need to know which slot (number) is used.

Joining them would, in my opinion, make both needlessly complicated and hard to read. What do you think? :)

CAM-Gerlach:
Also, as you can see, you'll need to resolve some merge conflicts that have accumulated over the time this PR was dormant, and it'll probably be easier to do that sooner rather than later.

> I fixed the docstrings (sorry about that) and am going to write some tests probably this evening and will push everything soon.

Great; looking forward to it!

> Joining them would, in my opinion, make both needlessly complicated and hard to read.

Your reasons sound solid, and I trust your judgement, so that sounds okay to me.

thepycoder:
Hey @ccordoba12 thanks for reviewing! I completely agree with your points and it would be awesome if someone else could take it from here. 
It's not due to time constraint but more that I'm still learning gui programming. I'll keep working on spyder's smaller issues for now and gain some experience first :)

CAM-Gerlach:
> So before you proceed further, please see how bookmarks are handled in Matlab and try to create a similar interface to that.

Well, do note that this bookmark system is implemented quite a bit differently from Matlab at a fundamental/conceptual level, and thus the same UI won't really work. It appears that bookmarks in Matlab are:

* Arbitrary (number of) slots
* Per-file only
* Disappear when the file is closed
* Can only be accessed sequentially. 

Whereas in this system, they are 

* Saved in specific slots
* Can be accessed globally; will switch to or even open the file if needed
* Are persistent across opening/closeing files and Spyder
* Can be unambiguously accessed at any time via a dedicated shortcut.

Both designs have their advantages, though @thepycoder 's system is actually strictly more functional (aside from the slot count limitation, but that's unlikely to be a major issue in practice), and Spyder's existing previous/next code cell functionality already mostly covers the "sequential movement through significant points in file" workflow that Matlab's are more suited for.

This means, however, that there's no real meaningfulness to a "previous"/"next" bookmark command other than the rather arbitrary order of the slots, since they are just accessed directly rather than sequential in a given document. Furthermore, its much less necessary/helpful to have an indicator in the panels at least for a first PR, since bookmarks don't need to be navigated to in order to be cleared, most or all may not even be in the current file (most likely only a few would be) so they wouldn't show up, the direct-access method is less reliant on users being able to see the location of their bookmarks, and the icon would have to clearly indicate the bookmark's slot number to be useful at all.

However, I do agree with @ccordoba12 that a GUI should be implemented, at least to expose the functionality to the user so it is actually discoverable. The obvious choice is a ``Bookmarks`` submenu of the ``Source`` menu with a list of ``Go to bookmark N`` (grayed out if not set) and ``Save bookmark N``, separated by a divider. This should actually be straightforward for @thepycoder to do even as novice in just a few lines of code; see e.g. [how the EoL submenu is created](https://github.com/spyder-ide/spyder/blob/d579bb5de466f3543c27d4c55fad749c71ad8e7e/spyder/plugins/editor/plugin.py#L719).

bcolsen:
@thepycoder Could you edit your PR description (the top post here) with some details with what has currently been implemented here. I thought that the bookmarks would be more matlab style like what @ccordoba12 was describing, until I read @CAM-Gerlach's previous post.

I agree with @CAM-Gerlach. I like the slots idea that's done here, but there is a little overhead to remember which slots are currently used when I want to make a new one.

Ideas:
1. We could alleviate the affect of limited slots by defining bookmark slots on a per project as well.
2. The GUI can show the slot number of the bookmark 
3. A bookmark side panel like an outliner for bookmarks
3. It would be great to have a previous bookmark shortcut

CAM-Gerlach:
> but there is a little overhead to remember which slots are currently used when I want to make a new one.

This could be at least somewhat helped by greying out the ``Go to bookmark N`` text of the unused slots in the ``Bookmarks`` submenu I proposed (I can't recall if I already mentioned that above).

> 1. We could alleviate the affect of limited slots by defining bookmark slots on a per project as well.

100% agree, but not sure its practicable until we implement per-project preferences, particularly for a first implementation PR.

> 2. The GUI can show the slot number of the bookmark

I'm not sure what you mean by this. The GUI where? Do you mean a side panel showing the location? It would be nice to have at some point, but not vital for a first implementation since A. The only time it would be seen is when the bookmarked file and line is already open, active and scrolled to the right position, which makes triggering the bookmark based on that number pointless; and B, given one's bookmarks would typically be spread out among many lines of at least several files, not all of which might even be open, its essentially always quicker to just trigger the bookmark to see where it goes than manually scroll through looking for the right numbered icon.

> 3. A bookmark side panel like an outliner for bookmarks

This is a great idea and a much more powerful UI than a simple menu; it could show not only the slot number, but the file and line. Its a bit overkill for only 10 bookmarks, and certainly so for a first implementation like this PR, but once its implemented it could extend the system to allow unlimited bookmarks by allowing the user to easily create, edit and delete bookmarks and dynamically assign desired ones to one of the 10 keyboard slots.

> It would be great to have a previous bookmark shortcut

Great idea, and probably pretty straightforward to implement here. What would be even cooler is to check if the last bookmark location is still within the visible scroll area, and if so jump to the bookmark before the previous instead, so it could also be used to toggle between the two most recently used bookmarks (i.e. two well-separated code blocks of interest, a common application of bookmarks).

thepycoder:
> > 1. We could alleviate the affect of limited slots by defining bookmark slots on a per project as well.
> 
> 100% agree, but not sure its practicable until we implement per-project preferences, particularly for a first implementation PR.

I don't fully agree actually. I would propose to make it toggle-able. I have the personal habit of referring back to files from other projects when writing a new one. I believe I'm not alone in this. The fact that it opens the file for you could make it very useful when 'referencing' accross multiple projects.

I do agree with all you other points and would love to implement it myself if possible :)

CAM-Gerlach:
> I don't fully agree actually.

Fair point, I guess—it depends on the user's workflow and habits. Regardless, yeah, we can worry about it later vs. in this PR.

> I do agree with all you other points and would love to implement it myself if possible :)

Fantastic! <3 Really glad to hear! Let us know if you need help/guidance on anything.

spyder/plugins/editor/widgets/tests/test_bookmarks.py
```diff
@@ -0,0 +1,275 @@
...
+        editor.bookmarks_changed.emit.assert_called_with()
+    else:
+        editor.bookmarks_changed.emit.assert_not_called()
+
+
+# --- Fixtures
+# -----------------------------------------------------------------------------
+@pytest.fixture
+def code_editor_bot(qtbot):
+    """Create code editor with default Python code."""
```
CAM-Gerlach: Instead of duplicating this code from another test function (non-DRY and modular), you should make it a globally accessible fixture instead. See my [comment](https://github.com/spyder-ide/spyder/pull/8644#discussion_r251208258) on another PR for how to do that.

spyder/plugins/editor/widgets/tests/test_bookmarks.py
```diff
@@ -0,0 +1,275 @@
...
+            '    return c\n'
+            )
+    editor.set_text(text)
+    return editor, qtbot
+
+
+# ---- Qt Test Fixtures
+@pytest.fixture
+def setup_editor(qtbot, monkeypatch):
+    """Set up the Editor plugin."""
```
CAM-Gerlach: Instead of duplicating this code from another test function (non-DRY and modular), you should make it a globally accessible fixture instead. See my [comment](https://github.com/spyder-ide/spyder/pull/8644#discussion_r251208258) on another PR for how to do that.

spyder/plugins/editor/widgets/tests/test_bookmarks.py
```diff
@@ -0,0 +1,275 @@
...
+
+    window = MainMock()
+    editor = Editor(window)
+    window.setCentralWidget(editor)
+    window.resize(640, 480)
+    qtbot.addWidget(window)
+    window.show()
+
+    yield editor
+
```
CAM-Gerlach: Missing the ``editor.close()`` after the ``yield`` (as the fixture you copied had); this would be resolved doing the above.

spyder/plugins/editor/widgets/tests/test_bookmarks.py
```diff
@@ -0,0 +1,275 @@
...
+                "    print(Hello World!)\n")
+
+    return filename
+
+
+# --- Tests
+# -----------------------------------------------------------------------------
+def test_add_bookmark(code_editor_bot, mocker):
+    """Test CodeEditor.add_bookmark. Adds bookmark data to Textblock."""
+    editor, _ = code_editor_bot
```
CAM-Gerlach: Use ``__`` for a dummy variable, not ``_`` (since we use that for ``gettext``)

spyder/plugins/editor/widgets/tests/test_bookmarks.py
```diff
@@ -0,0 +1,275 @@
...
+    cursor = editor.textCursor()
+    cursor.movePosition(QTextCursor.Right, n=2)
+    editor.setTextCursor(cursor)
+    ab(2)
+    editor_assert_helper(editor, block, bm=[(1, 0), (2, 2)], emits=True)
+
+
+def test_get_bookmarks(code_editor_bot):
+    """Test CodeEditor.get_bookmarks. Returns data found in textblocks."""
+    editor, _ = code_editor_bot
```
CAM-Gerlach: Use ``__`` for a dummy variable, not ``_`` (since we use that for ``gettext``)

spyder/plugins/editor/widgets/tests/test_bookmarks.py
```diff
@@ -0,0 +1,275 @@
...
+
+    # Add bookmarks.
+    bm = {1: ('filename', 1, 0), 2: ('filename', 3, 5), 3: ('filename', 4, 3)}
+    editor.set_bookmarks(bm)
+    assert(gb() == {1: [1, 0], 2: [3, 5], 3: [4, 3]})
+
+
+def test_clear_bookmarks(code_editor_bot):
+    """Test CodeEditor.clear_bookmarks. Removes bookmarks from all blocks."""
+    editor, _ = code_editor_bot
```
CAM-Gerlach: Use ``__`` for a dummy variable, not ``_`` (since we use that for ``gettext``)

spyder/plugins/editor/widgets/tests/test_bookmarks.py
```diff
@@ -0,0 +1,275 @@
...
+    # Even though there is a 'del data' that would pop the item from the
+    # list, the __del__ funcion isn't called.
+    assert len(editor.blockuserdata_list) == 2
+    for data in editor.blockuserdata_list:
+        assert not data.bookmarks
+
+
+def test_update_bookmarks(code_editor_bot):
+    """Test CodeEditor.update_bookmarks. Check if signal is emitted."""
+    editor, _ = code_editor_bot
```
CAM-Gerlach: Use ``__`` for a dummy variable, not ``_`` (since we use that for ``gettext``)

spyder/plugins/editor/widgets/tests/test_bookmarks.py
```diff
@@ -0,0 +1,275 @@
...
+
+# --- Helper methods
+# -----------------------------------------------------------------------------
+def reset_emits(editor):
+    "Reset signal mocks."
+    editor.bookmarks_changed.reset_mock()
+
+
+def editor_assert_helper(editor, block=None, bm=None, emits=True):
+    """Run the tests for call to add_remove_breakpoint.
```
CAM-Gerlach: Instead of duplicating this code from another test function (non-DRY and modular), you should make it a globally accessible helper function instead. See my [comment](https://github.com/spyder-ide/spyder/pull/8644#discussion_r251208258) on another PR for how to do that.

spyder/plugins/editor/widgets/tests/test_bookmarks.py
```diff
@@ -0,0 +1,275 @@
...
+
+    return filename
+
+
+# --- Tests
+# -----------------------------------------------------------------------------
+def test_add_bookmark(code_editor_bot, mocker):
+    """Test CodeEditor.add_bookmark. Adds bookmark data to Textblock."""
+    editor, _ = code_editor_bot
+    ab = editor.add_bookmark
```
CAM-Gerlach: Is there a particular justification for remapping this name? It just seems to obfuscate the logic of the test merely to save a few characters, which are nearly inconsequential if using autocompletion or line duplication.

spyder/plugins/editor/widgets/tests/test_bookmarks.py
```diff
@@ -0,0 +1,275 @@
...
+    with open(filename, 'w') as f:
+        f.write("# -*- coding: utf-8 -*-\n"
+                "def foo:\n"
+                "    print(Hello World!)\n")
+
+    return filename
+
+
+# --- Tests
+# -----------------------------------------------------------------------------
```
CAM-Gerlach: Our current convention is ``# --------------- [etc]`` above and below

spyder/plugins/editor/widgets/tests/test_bookmarks.py
```diff
@@ -0,0 +1,275 @@
...
+    data = block.userData()
+    assert data.bookmarks == bm
+    if emits:
+        editor.bookmarks_changed.emit.assert_called_with()
+    else:
+        editor.bookmarks_changed.emit.assert_not_called()
+
+
+# --- Fixtures
+# -----------------------------------------------------------------------------
```
CAM-Gerlach: Our current convention is ``# --------------- [etc]`` above and below

spyder/plugins/editor/widgets/tests/test_bookmarks.py
```diff
@@ -0,0 +1,275 @@
...
+    assert(gb() == {})
+
+    # Add bookmarks.
+    bm = {1: ('filename', 1, 0), 2: ('filename', 3, 5), 3: ('filename', 4, 3)}
+    editor.set_bookmarks(bm)
+    assert(gb() == {1: [1, 0], 2: [3, 5], 3: [4, 3]})
+
+
+def test_clear_bookmarks(code_editor_bot):
+    """Test CodeEditor.clear_bookmarks. Removes bookmarks from all blocks."""
```
CAM-Gerlach: *Remove (imperative).

spyder/plugins/editor/widgets/tests/test_bookmarks.py
```diff
@@ -0,0 +1,275 @@
...
+    reset_emits(editor)
+    editor.bookmarks_changed.emit.assert_not_called()
+    # update_bookmarks is the slot for the blockCountChanged signal.
+    editor.textCursor().insertBlock()
+    editor.bookmarks_changed.emit.assert_called_with()
+
+
+def test_save_bookmark(setup_editor):
+    """Test Plugin.save_bookmark. 
+
```
CAM-Gerlach: Our agreed convention is to make the summary line on its own line for multi-line docstrings

spyder/plugins/editor/widgets/tests/test_bookmarks.py
```diff
@@ -0,0 +1,275 @@
...
+    assert bookmarks == [(1, 2)]
+
+    # Check if bookmark was removed from previous block
+    bookmarks = edtr.document().findBlockByNumber(0).userData().bookmarks
+    assert bookmarks == []
+
+
+def test_load_bookmark(setup_editor, test_file):
+    """Test that loading a bookmark works.
+
```
CAM-Gerlach: Our agreed convention is to make the summary line on its own line for multi-line docstrings

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1444,6 +1450,54 @@ def update_breakpoints(self):
         """Update breakpoints"""
         self.breakpoints_changed.emit()
 
+    # -----Code bookmarks
```
CAM-Gerlach: Space between ``-----`` and ``Code bookmarks``

CAM-Gerlach:
Sorry for the delay reviewing this, @thepycoder  . I ran into serious issues this past week with my Anaconda install that ended up taking several days to resolve, and I then had some other commitments and catch-up to do after finally getting that working again.

thepycoder:
Hey @CAM-Gerlach, No sorry, I have been really busy lately. I'm also having trouble with the rebasing because someone refactored the userblockdata code as part of the debugging code base. Now my changes are done in a file that doesn't exists anymore.

ccordoba12:
I don't like the addition of so many keyboard shortcuts to save/remove bookmarks. So I'd like to be removed of this PR before proceeding further. I don't plan to merge this without that gone because I agree what we really need to do is to add Bookmarks pane.

@thepycoder, you can still keep working in your branch without rebasing. When you address @CAM-Gerlach's remarks and remove the shortcuts you added, I'll ask @dalthviz to help you update your work and merge it.

CAM-Gerlach:
> what we really need to do is to add a Bookmarks pane.

My only concern with a Bookmarks pane is that it's one more thing to stuff into an already fairly crowded UI, and having to tab to it or even open it just to browse bookmarks diminishes its utility (in particular, this makes *some* sort of KB shortcut system mandatory). Could we consider including a "Show bookmarks" option in the Outline Explorer for more convenient navigation, and then making the Bookmark pane just an Outline Explorer showing just bookmarks and the extra bookmark management features (add, remove, display/assign shortcut, etc)? Then, the user would only need to open the latter for major management of their bookmarks, while the Outline Explorer w/Bookmarks enabled could suffice for most general browsing and recalling of them. Furthermore, the functionality would already exist to show only the bookmarks in the current file, or all bookmarks. Not sure if this is the best solution, but its an option to consider.

> I don't like the addition of so many keyboard shortcuts to save/remove bookmarks. 
> I don't plan to merge this without that gone

I agree that an Outline-Explorer-like pane is a fantastic GUI for this, which would free us conceptually from the idea of 10 keyboard-shortcut "slots" to store and retrieve the bookmarks. However, we still need *some* form of keyboard shortcut UI to access them, as the whole point is quick access and instantly tabbing back and forth between, which switching to the mouse, possibly opening the bookmarks pane, navigating to the right one, and then switching focus back to the Editor obviously does not provide.

As I explained previously, a "Previous" and "Next" shortcut like with Matlab doesn't really make sense with this system, at least not as the primary keyboard UI, since on our system shortcuts aren't per file, and there is no definitive order to them; it doesn't scale with an arbitrary number of shortcuts in an arbitrary number of files; and it defeats a lot of the advantages of this implementation (being able to switch, with one consistent keypress, directly between different points in the same/different file).

Therefore, other than the ability to assign on the order of 5-10 numbered shortcuts to user-selected bookmarks in the bookmarks pane, how would you propose implementing the keyboard UI for this?

spyder/plugins/editor/widgets/tests/test_bookmarks.py
thepycoder: I can't find this, where does this ```editor.close()``` come from?

spyder/plugins/editor/widgets/tests/test_bookmarks.py
thepycoder: Nope, no logic, removed.

thepycoder:
I have addressed @CAM-Gerlach's notes. They should be fixed now :)
For the UI: if I remove the shortcuts, do I replace them with the menu entry? Or something else? Your call!

CAM-Gerlach:
> For the UI: if I remove the shortcuts, do I replace them with the menu entry? Or something else? Your call!

It hasn't been decided yet; we're awaiting @ccordoba12 's response to my above comment, which I just cut down to be more concise.

However, it looks like its going to be some form of a bookmark explorer, likely either based off the outline explorer, breakpoint pane or perhaps both, which should make implementation much simpler than coming up with something from scratch (especially with the latter, you basically would just need to swap bookmarks for breakpoints—which already use the same tracking code—and replace "Condition" with "Shortcut"); combined with some form of keyboard shortcuts (I'm suggesting either a set of assignable shortcuts similar to the current system, or custom user-defined ones set per-bookmark).

ccordoba12:
@thepycoder, what I have in mind is mostly what this extension provides for VSCode:

https://marketplace.visualstudio.com/items?itemName=alefragnani.Bookmarks

So, for the UI part I'd like to see:

1. An icon next to the line that has a bookmark.
2. A panel to browse bookmarks.
3. A keyboard shortcut to move to the next/previous bookmark.

Points 1. and 2. are out of the scope of this PR. If you're willing to implement point 3. here then please go ahead. If not, then please remove all shortcuts you introduced (as I said before) and we'll help you to fix conflicts and merge to master.

ccordoba12:
> However, we still need some form of keyboard shortcut UI to access them, as the whole point is quick access and instantly tabbing back and forth between

To address this, we can add the list of bookmarks to the file switcher or add a shortcut to move to the bookmarks pane and filter file names with the keyboard there.

The point is that there's no need to keep discussing about it here because @thepycoder work provides mainly the initial implementation, so we have to help him finish his work and after discuss about what UI build for this.

> A "Previous" and "Next" shortcut like with Matlab doesn't really make sense here as the primary UI

It won't be the primary UI. We agreed on the panel too, so please stop arguing about this. It's useful to move among several bookmarks in the same file and among files (in alphabetical or temporal order of setting your bookmarks, for example).

thepycoder:
I removed the shortcuts, so only the main functions remain with no user interaction.

> Points 1. and 2. are out of the scope of this PR. If you're willing to implement point 3. here then please go ahead. If not, then please remove all shortcuts you introduced (as I said before) and we'll help you to fix conflicts and merge to master.

I fully agree on point 1. I use the VSCode extention and it is a must.
I'm willing to implement 3 but I think the result will be worse than just waiting for the full panel to be ready. 
I thought about how it would work and there can be no structure. What does Next/Previous mean in this context? If we do it alphabetically then we'd have to cycle several bookmarks to get to the one we need, switching editors and potentially opening files for every bookmark in between.
When using temporal order, is the last used one Next or Previous?
I think it would be quite clunky using these shortcuts, if I misinterpreted your idea of Prev/Next I'm still willing to implement it your way :)

CAM-Gerlach:
> Points 1. and 2. are out of the scope of this PR.

:+1: , and I also agree with both of them.

> If you're willing to implement point 3. here then please go ahead. If not, then please remove all shortcuts you introduced

Aside from allowing >10 bookmarks, the former is a strict regression from the currently implemented functionality as it does not allow random access or navigating to bookmarks in other files (or else, there is no defined order to them; this is all stuff that should be addressed in the UI discussion as you say, not here), while the latter makes it impossible to use or functionally test the bookmarks feature entirely. While this PR is an initial implementation, it should at least allow for basic usage of the implemented parts of the feature, even with a rough/non-ideal UI.

As to @ccordoba12 's points on the UI, I will open a separate issue to discuss that so as to not derail this PR.

> I'm willing to implement 3 but I think the result will be worse than just waiting for the full panel to be ready.

Agreed. These questions will need to be discussed, decided upon and implemented in order to add this UI, but as @ccordoba12 has stated the finer points of the UI is out of scope for this PR. Therefore, I suggest we just keep the existing UI for now so the feature is actually usable as implemented, and then discuss them in a separate issue.

ccordoba12:
> I'm willing to implement 3 but I think the result will be worse than just waiting for the full panel to be ready.
I thought about how it would work and there can be no structure. What does Next/Previous mean in this context? If we do it alphabetically then we'd have to cycle several bookmarks to get to the one we need, switching editors and potentially opening files for every bookmark in between.
When using temporal order, is the last used one Next or Previous?

You're right. I haven't considered the complexity and implications of such a shortcut. I think we could start with something very simple, i.e. the shortcut should move to the previous/next bookmark in the current file (where it makes sense of talking about prev/next and it'd be very useful).

However, let's leave that for a future PR too. I don't want you to make more changes here because the conflicts in this PR don't allow to run our full test suite.

So @thepycoder, please don't add more commits here. @dalthviz, will fix the conflicts, make our tests pass and merge it. Thanks a lot for your work and your patience with our review process. I want to emphasize that I think this is really great contribution and would be invaluable to work with large codebases (including our own!!).

@CAM-Gerlach, please open an issue to discuss about the UI for bookmarks.

CAM-Gerlach:
@ccordoba12 Agree with all. 

> @CAM-Gerlach please open an issue to discuss about the UI for bookmarks.

Yep, already done as #9021 .

ccordoba12:
> Therefore, I suggest we just keep the existing UI for now so the feature is actually usable as implemented, and then discuss them in a separate issue.

Sorry, I don't agree with this. Let's make this PR to add the low-level implementation and create the right UI for it afterwards, instead of leaving something that we know it's going to be replaced in the future.

CAM-Gerlach:
> Let's make this PR to add the low-level implementation

Okay, sure. Sounds like a good plan.

thepycoder:
@CAM-Gerlach @ccordoba12 Great! I'll leave the rest up to the pros. Thank you guys for your patience too, I haven't always been speedy with my work here. Especially @CAM-Gerlach thanks a lot for the code reviews, I learned a lot and will try to keep contributing to spyder in the future. Be it with less pep fix commits ;)

spyder/plugins/editor/widgets/editor.py
```diff
@@ -434,6 +443,9 @@ class EditorStack(QWidget):
     sig_go_to_definition = Signal(str, int, int)
     perform_lsp_request = Signal(str, str, dict)
     sig_option_changed = Signal(str, object)  # config option needs changing
+    save_bookmark = Signal(int)
+    load_bookmark = Signal(int)
+    save_bookmarks = Signal(str, str)
```
ccordoba12: @dalthviz, please change these signals to `sig_save_bookmark`, `sig_load_bookmark` and `sig_save_bookmarks`.

dalthviz:
@ccordoba12 should I merge this?

ccordoba12:
Yep, please merge it. Thanks a lot @thepycoder for your contribution and @dalthviz for your help to finish it!

agentmorris:
My understanding is that this is merged into the most recent 4.0b release, and that despite some discussion above, the shortcuts remains as ctrl-shift-N and ctrl-alt-N to set/unset bookmarks.  This is consistent with what I see in the code associated with this PR.  I'm *really* excited about this feature, but can't get it to work in 4.0b7.  Do I have the interaction right, or after discussion did the paradigm change?  Are others able to use this feature?

CAM-Gerlach:
@agentmorris The backend was merged a long time ago, but we were never able to come to consensus on a mutually agreeable UI for the feature so it remained stuck in development limbo. If you're presumably welcome to propose a UI implementation if you'd like to see it implemented, though I would discuss it with @ccordoba12 and the gang and make sure you're all in rough agreement before going ahead with it.

agentmorris:
Got it, now I feel like I'm not missing anything. :)

My strong preference is for exactly what both Matlab and Visual Studio do, which is very simple:

1) Ctrl-F2 sets or removes a bookmark on the current line.  This is indicated with a dot in the line numbering area, where breakpoints are indicated.

2) F2 cycles the cursor position through bookmarks in all open files. (Shift-F2 cycles backwards).

The keyboard shortcuts are different in VS, but the paradigm is identical.

There is no numbering to bookmarks.  IIRC bookmarks are not saved in either Matlab or VS.

I personally prefer this simple approach to vi-style named/numbered bookmarks; the way I use this (and apparently the way most Matlab users use this, else they presumably would have changed it) is pretty temporary, and the cognitive load of remembering multiple bookmark numbers would be too much.

The "bang for the buck" on this simple approach is very high IMO.  I could easily live without the visual indicator as well if that's the dealbreaking issue.

CAM-Gerlach:
Thanks for your input; I should have made clear that such discussion should happen on  #9021 , not here, sorry. To note, the backend would probably need to be heavily reworked and a number of its features thrown out to accommodate that, as it was designed around a very different UX paradigm (something closer to the Vi-style you mentioned), and there's also the ambiguous question of how to order the bookmarks, especially between files. However, the UI you propose is presumably more likely to get accepted then what we've been discussing, and better to have something working than nothing.

agentmorris:
Oops, will more or less re-post there, in what I believe is only a marginal violation of GitHub etiquette to avoid cross-posting.  Thanks!

PR #8569, opened 2019-01-13
PR: Prevent pasting non-text data from clipboard into the editor

Khalilsqu:
## Description of Changes

* [ ] Wrote at least one-line docstrings (for any new functions)
* [ ] Added unit test(s) covering the changes (if testable)
* [x] Included a screenshot (if affecting the UI)

Old behaviour:
copy_names_olddd

New behaviour
copy_names_new

This PR prevents pasting files/folders paths into Spyder editor when they are copied into the OS clipboard. Text data still can be pasted safely within spyder editor, from spyder editor to external file or from an external source into spyder editor.

Folders/files but can be pasted into Project/File explorers if #8494 is accepted and merged into the master.

### Issue(s) Resolved

Fixes #8566

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct:
@Khalilsqu

Issue #8566, opened 2019-01-13
File names pasted in editor when copied from OS file manager.


Khalilsqu:
So I noticed that when copying a file from OS file manager and pasting it into the editor, its name will be pasted. A similar behaviour is solved with the commit `correct tests and remove text from copy file` in PR #8494 

![name_issue_copy_issue](https://user-images.githubusercontent.com/22060413/51087973-134bc280-1752-11e9-9bea-a84bffbe2da0.gif)

It is not considered a big issue.  But might be preferred to not allow this behaviour to happen.

Spyder: Master (currently under development) 
OS: windows: 10
Python version: 3.6

ccordoba12:
So what should be the expected behavior in this case?

Khalilsqu:
> So what should be the expected behavior in this case?

I Think there should be no text in the clipboard when pasting clipboard data which contains none text contents (mimeData) e.g. files/folders. The text should be set to empty string if the data within the clipboard are mimeData.

Khalilsqu:
@ccordoba12 

By the way. It can be fixed by changing this [function](https://github.com/spyder-ide/spyder/blob/a5ca1f20f69ba8f9bc1b85e166ce865302c06a54/spyder/plugins/editor/widgets/codeeditor.py#L1578) as such.

    ```
@Slot()
    def paste(self):
        """
        Reimplement QPlainTextEdit's method to fix the following issue:
        on Windows, pasted text has only 'LF' EOL chars even if the original
        text has 'CRLF' EOL chars
        """
        clipboard = QApplication.clipboard()
        if clipboard.mimeData().hasUrls():
            clipboard.setText('')
        else:
            text = to_text_string(clipboard.text())
            if len(text.splitlines()) > 1:
                eol_chars = self.get_line_separator()
                text = eol_chars.join((text + eol_chars).splitlines())
                clipboard.setText(text)
            # Standard paste
            TextEditBaseWidget.paste(self)
            self.document_did_change(text)
```

I tested it, This changes does not affect pasting text data (within Spyder editor itself or from outside Spyder) but prevent pasting file/folders names if they are copied in the clipboard from outside Spyder.

ccordoba12:
Ok, I agree with this, except for one thing: instead of using `clipboard.setText('')` we should simply change that line for `return`. That will avoid resetting user's clipboard and pasting its contents in our editor.

What do you think?

Khalilsqu:
[...]

PR #8571, opened 2019-01-14
PR: Remove the test np_threshold from test_mainwindow.py

juanis2112:
## Description of Changes

* [ ] Wrote at least one-line docstrings (for any new functions)
* [ ] Added unit test(s) covering the changes (if testable)
* [ ] Included a screenshot (if affecting the UI)

I removed the test np_threshold from test_mainwindow because it was added to Spyder-kernels.

### Issue(s) Resolved

Fixes #

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct: Juanis2112

PR #8494, opened 2018-12-27
PR: Add the ability to copy/paste files and their paths in the File/Project Explorers

Khalilsqu:
## Description of Changes

* [x] Wrote at least one-line docstrings (for any new functions)
* [x] Read and followed this repo's [Contributing Guidelines](https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md)
* [x] Based your PR on the latest version of the correct branch (master)
* [x] Followed [PEP8 ](https://www.python.org/dev/peps/pep-0008/)for code style
* [x] Added unit test(s) covering the changes (if testable)
* [x] Ensured your pull request hasn't eliminated unrelated blank lines/spaces, modified the spyder/defaults directory, or added new icons/assets
* [x] Described your changes and the motivation for them below
* [x] Noted what issue(s) this pull request resolves
* [x] Included a screenshot (if affecting the UI)

Successor of PR #8485 .

Add ability to copy and save files or their paths in project or file explorer.

**Motivation**
This PR adds some desirable functionality to Spyder-IDE.
1. Copy and paste a single or multiple files absolute or relative paths from Project or File Explorer.
2. Copy a single or multiple file(s)/folder(s) within Project/File Explorer. The copied files/folders can be pasted into OS File Manager or vise versa.
3. Simplified these functionalities with keyboard shortcuts.

### Issue(s) Resolved

Fixes #2641 
Fixes #8321

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct:

PR #8576, opened 2019-01-14
PR: Add ability to paste auto-formatted file paths into the Editor from the system file manager

Khalilsqu:
## Description of Changes

The original behaviour before the PR #8569 being merged has some undesirable side effects.

1. When pasting clipboard files in Spyder main editor, these files will be lost from OS clipboard and cannot be pasted in other applications or OS file manager. This happens only when multiple files/folders are copied but not a single one.
a. multiple files case

multiple_files_before
b . Single file/folder copy works fine, Spyder does not have effect on it.

single_files_before

2. The standard in windows application is that when files/folders are copied into the clipboard, their none useful paths will not be pasted as a text in any text editor. But after pasting them to spyder editor, their names can be pasted in other editors as well. This behaviour happens only when multiple files are selected and pasted in Spyder editor.

paste_files_before

3. Finally, as you can see files names are not pasted properly as useful python path in win OS but are pasted as usable python file/folder paths in Linux.

Therefore, I proposed the change in PR #8569. But, after discussion in the issue #8566 . It is found the last merged PR is not an ideal solution.

1. No path at all pasted in all OS in Spyder when files/folders are copied into Spyder.
2. Sometime even actual text data are ignored as was noted by @ok97465 when he was tried to copy codes from Pycharm to Spyder.

Therefore, I am proposing this new PR: My tests shows that it resolves all above concerns.
1. Files/folders names are pasted as useful paths in win and linux. (single/Multiple)
2. Files/Folders are not lost when multiple files folders are copied into Spyder.
3. No text are pasted in other application when multiple files/folders are pasted in Spyder editor.
4. The original files contents are restored into the clipboard after pasting into Spyder as useful file paths for WIN and Linux.
 5. You can paste from Pycharm to Spyder, or any editor/source to Spyder and vise versa.

### Affirmation


Khalilsqu:
Sorry about that, I tried to resolve merge conflict. and got all these pep8 errors. I only change I made is this.

```
    @Slot()
    def paste(self):
        """
        Reimplement QPlainTextEdit's method to fix the following issue:
        on Windows, pasted text has only 'LF' EOL chars even if the original
        text has 'CRLF' EOL chars
        """
        clipboard = QApplication.clipboard()
        # This is here to prevent pasting mime-urls in the Editor in Windows.
        # See issue 8566 for the details.
        if clipboard.mimeData().hasFormat("text/plain"):
            text = to_text_string(clipboard.text())
            if len(text.splitlines()) > 1:
                eol_chars = self.get_line_separator()
                text = eol_chars.join((text + eol_chars).splitlines())
                clipboard.setText(text)
            # Standard paste
            TextEditBaseWidget.paste(self)
            self.document_did_change(text)
```

CAM-Gerlach:
> Sorry about that, I tried to resolve merge conflict. and got all these pep8 errors. I only change I made is this.

My guess is that you accidentally saved changed the file's line endings (``LF`` -> ``CRLF``, or ``CRLF`` -> ``LF``), such that the entire file is showing as modified in Github (and thus triggering PEP8Speaks). Just ``git reset --hard ccbea56742`` this branch and force-push. Also, especially if you only have a few or especially one commit, just ``rebase`` on ``master`` instead of ``merge``-ing to avoid needlessly polluting the commit history.

Khalilsqu:
I am very sorry about all this hassle. : not having much familiarity with git system!!

> On Windows using this branch, as is presumably intended, I can't copy files as paths into the Editor at all, which is a straight step backwards in functionality from at least being able to paste a path with `file:///` prepended like before. Instead, you should implement what @bcolsen requested,

But in windows the recommended way to copy paths is through `home >> copy path` in the file manager after selecting the files/folders you want to copy.

> > If the windows file paths are unusable in a python script with the file:\\ stuff in the front then the fix to the this issue is to make the url paste as a usable python path string like "D:\path\file.ext".
> 
> More specifically, just remove the `file:///` in front when pasting and everything should work fine.
> 
> Meanwhile, on my Fedora 28 LXQt VM using VBox, I get `file:///path/to/file.py` MIME URLs when pasting files from PCManFM-Qt into Spyder (as well as into Kate), not actual paths. Therefore, the solution appears to be to strip the `file:///` on all platforms. If there's not a more sophisticated/unambiguous method of detecting a MIME URL, you could simply do
> 
Okay, seems that different distributions of Linux works differently. I am having WSL with ubuntu and the file manager used is the method proposed just works fine.

> ```python
> if  text.startswith('file:///') and len(text) > 10:
>     if os.name() == 'nt':
>         text.replace('file:///', '')
>     else:
>         text.replace('file:///', '/')
> ```

Good point, but it will not work if multiple files/folders are copied. need to step through each line and remove 'file:///'.

CAM-Gerlach:
@Khalilsqu Make sure you ``pull`` the latest master before rebasing so as to fix the unrelated tests issue due to #8571 . Thanks!

CAM-Gerlach:
BTW, this may be out of scope, but the ``file:///`` MIME URL thing happens when pasting into the Console as well which this PR or the previous naturally doesn't effect.

> I am very sorry about all this hassle. : not having much familiarity with git system!!

No worries; ``git`` can be quite tough to learn for anyone.

> But in windows the recommended way to copy paths is through `home >> copy path` in the file manager after selecting the files/folders you want to copy.

I was unaware of this (I use Multi Commander, not Windows Explorer, as my primary file manager since the latter is such unadulterated garbage ;), but there appears to be no obvious keyboard shortcut for that command and many users are likely equally not familiar with it as I was, and simply expect copy and file and pasting a path when appropriate (in the Editor) to just work (as really, it should)—and most importantly, there seems to be no good reason why it shouldn't, since it should be as simple as using the above code block on ``text``, ``if not clipboard.mimeData().hasFormat("text/plain")``.

> Okay, seems that different distributions of Linux works differently. I am having WSL with ubuntu and the file manager used is the method proposed just works fine.

Right; it could also be something to do with either VBox or Qt vs. GTK. The above solution I propose should handle either case just fine since the only OS specific thing is whether to remove the leading ``/``.

> Good point, but it will not work if multiple files/folders are copied. need to step through each line and remove 'file:///'.

What do you mean? The above solution with ``replace`` works perfectly fine on any number of files (I've refined it slightly to be a touch more robust).

CAM-Gerlach:
Some refinements on the above:

```python
text = to_text_string(clipboard.text())
if not clipboard.mimeData().hasFormat("text/plain"):
      if text.startswith('file:///') and len(text) > 9:
          if os.name() == 'nt':
              replace_prefix = '\n'
          else:
              replace_prefix = '\n/'
          text = text[8:].replace('file:///', replace_prefix)
if len(text.splitlines()) > 1:
[REST OF FUNCTION]
```

Khalilsqu:
@CAM-Gerlach Amazing, manual tests shows that it is working in both system. Linux:  ubuntu WSL and windows-10.

Do we need to insert a new line?? the format I am getting is different in ubuntu than windows

windows:

```
C:/Users/kalho/Desktop/spyder/spyder/img_src

C:/Users/kalho/Desktop/spyder/spyder/requirements

C:/Users/kalho/Desktop/spyder/spyder/rope_profiling

C:/Users/kalho/Desktop/spyder/spyder/scripts
```
ubuntu:

```
/home/kalhooti/spyder_test/spyder/LICENSE.txt
/home/kalhooti/spyder_test/spyder/MANIFEST.in
/home/kalhooti/spyder_test/spyder/NOTICE.txt
/home/kalhooti/spyder_test/spyder/README.md
/home/kalhooti/spyder_test/spyder/RELEASE.md
```

Also, frankly I could not understand the purpose of and `and len(text) > 9` and `text[8:]`..
it works without them.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1582,14 +1582,17 @@ def paste(self):
...
         clipboard = QApplication.clipboard()
-        text = to_text_string(clipboard.text())
-        if len(text.splitlines()) > 1:
-            eol_chars = self.get_line_separator()
-            text = eol_chars.join((text + eol_chars).splitlines())
-            clipboard.setText(text)
-        # Standard paste
-        TextEditBaseWidget.paste(self)
-        self.document_did_change(text)
+        # This is here to prevent pasting mime-urls in the Editor in Windows.
```
CAM-Gerlach: This needs to be updated based on what changes you end up making.

CAM-Gerlach:
> Do we need to insert a new line?? the format I am getting is different in ubuntu than windows

Great catch, I made a typo (now corrected above). There should be a ``\n`` at the beginning of the string to replace, i.e, ``\nfile///``.

CAM-Gerlach:
> Also, frankly I could not understand the purpose of and len(text) > 9 and text[8:]..it works without them. Do all paths exceeds this character limit?

``len(text) > 9`` (really, should be ``len(text) > 8`` since I forgot ``/`` was counted in the initial ``file:///``) ensures an ``IndexError`` isn't raised when ``file:///`` is the entire content of the paste and that it isn't replaced when it is the only thing being pasted, i.e. without a path. We can do ``text[8:]`` to remove the first ``file:///`` from the string (which we know is the first 8 characters since we checked for it), so that we can check for ``/n`` in ``replace()`` so that ``file:///`` is only replaced at the beginning of paths like it should be, just in case.

Khalilsqu:
Okay I did it in slightly different way. I think it is less prone to errors and OS independent. It works for single or multiple files, the previous was not working correctly for a single file

```
    @Slot()
    def paste(self):
        """
        Reimplement QPlainTextEdit's method to fix the following issue:
        on Windows, pasted text has only 'LF' EOL chars even if the original
        text has 'CRLF' EOL chars
        """
        clipboard = QApplication.clipboard()
        text = to_text_string(clipboard.text())
        # This is here to make copied files/folders to be pasted as path.
        # See issue 8566 for the details.
        if clipboard.mimeData().hasUrls():
            urls = clipboard.mimeData().urls()
            text = "".join(url.toLocalFile() + '\n' for url in urls)[:-1]
        if len(text.splitlines()) > 1:
            eol_chars = self.get_line_separator()
            text = eol_chars.join((text + eol_chars).splitlines())
            clipboard.setText(text)
        # Standard paste
        TextEditBaseWidget.paste(self)
        self.document_did_change(text)
```

@CAM-Gerlach would you please test it in your system. It is working in my simple manual test both in windows and ubuntu (WSL)

CAM-Gerlach:
> Okay I did it in slightly a different way. I think it is less prone to errors and is OS independent. 

That's way better! I figured there was a better way like that, and you'd probably find it on your own since you have a knack for doing that before :)

> would you please test it in your system

I hacked together some code to test it manually inside the Spyder internal console and it looks like it works fine, but please push the actual changes so I can do a proper test. Thanks!

Khalilsqu:
@CAM-Gerlach 
```
text = "".join('"' + url.toLocalFile().replace(osp.os.sep, '/') +
                           '"' + '\n' for url in urls)[:-1]
```

Would you prefer it like above, so each path is enclosed inside double quotes. for both single and multiple files.

Khalilsqu:
> but please push the actual changes so I can do a proper test. Thanks!

Would you please tell what commands I should use, I really a noob with github!! and I am afraid my other branch for copying files and paths might be damaged.

I am simply doing like that

```
git add -A
git commit -m "some comment"
git push origin prevent-copy-paths-windows
```

I do not understated how to make rebase!!!

CAM-Gerlach:
> Would you prefer it like above?, so each path is enclosed inside double quotes. for both single or multiple files.

Good thinking; how about you do it just like in #8494 where a single file has no quotes, while a list of files does, and commas after each one.

> Would you please tell what commands I should use

``git push``

Also, ``commit -am "Your descriptive commit message here"`` will add and commit your changes in one step so there's no need for the redundant ``git add -A``.

> I do not understated how to make rebase!!!

To rebase your branch from the lastest ``master``

```
git checkout master
git fetch upstream
git pull upstream master
git checkout prevent-copy-path-files-windows
git rebase master prevent-copy-path-files-windows
```

Then fix the merge conflicts if any (manually in the files, or with ``git mergetool``), then finally do ``git add .`` ``git rebase --continue`` and force push.

However, it seems you fixed the conflicts so everything should be okay.

> Still some issue!! needs more investigations!!

Your tests passed; can you be more specific?

Khalilsqu:
> Your tests passed; can you be more specific?

It was not working and pasting the file with prefix file:/// when only one file folder is copied.

Now it is working for all cases, and when only one files folder is copied, it is not enclosed into quotes in contrary to multiple files/folders. 

I hope @bcolsen will agree on this new changes. This is I believe more professional than having to paste none useful path names into the editor.

spyder/plugins/editor/widgets/codeeditor.py
Khalilsqu: Please resolve if passed!!

CAM-Gerlach: Sure

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1582,19 +1582,22 @@ def paste(self):
...
-            self.document_did_change(text)
+            urls = clipboard.mimeData().urls()
+            if len(urls) > 1:
+                text = ''.join('"' + url.toLocalFile().replace(osp.os.sep, '/')
+                               + '",' + '\n' for url in urls)
+            else:
+                text = urls[0].toLocalFile() + '\n'
+        eol_chars = self.get_line_separator()
+        text = eol_chars.join((text + eol_chars).splitlines())
+        clipboard.setText(text)
```
CAM-Gerlach: The issue here is we're overwriting the content of the global system clipboard, which is not the user-expectation when pasting. Can you either save and restore the previous clipboard contents once done pasting here, or add a ``text`` parameter to ``TextEditBaseWidget`` (defaulting to ``None`` with the current behavior, of course) that allows passing text to paste directly to that function?

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1582,19 +1582,29 @@ def paste(self):
...
+        # See issue 8566 and PR: 8576 for the details.
+        urls = []
+        if (not clipboard.mimeData().hasFormat('text/plain') and
+                clipboard.mimeData().hasUrls()):
+            urls = clipboard.mimeData().urls()
+            if len(urls) > 1:
+                text = ''.join('"' + url.toLocalFile().replace(osp.os.sep, '/')
+                               + '",' + '\n' for url in urls)
+            else:
+                text = urls[0].toLocalFile() + '\n'
```
bcolsen: I don't think that most users would want a `'\n'` here. Usually the path will be pasted in line.

Khalilsqu: @bcolsen , I agree to your point, I was wrong, I will remove from multiple files too. Thanks.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1582,19 +1582,29 @@ def paste(self):
...
+        text = to_text_string(clipboard.text())
+        # This is here to make copied files/folders to be pasted as text path.
+        # See issue 8566 and PR: 8576 for the details.
+        urls = []
+        if (not clipboard.mimeData().hasFormat('text/plain') and
+                clipboard.mimeData().hasUrls()):
+            urls = clipboard.mimeData().urls()
+            if len(urls) > 1:
+                text = ''.join('"' + url.toLocalFile().replace(osp.os.sep, '/')
+                               + '",' + '\n' for url in urls)
```
bcolsen: This `\n` should be replaced `eol_chars` from `eol_chars = self.get_line_separator()` so they match the user's preferences.  

`eol_chars = self.get_line_separator()` should be define only once in the function.

spyder/plugins/editor/widgets/codeeditor.py
Khalilsqu: in #8494 your suggested standard is followed.

bcolsen:
@Khalilsqu It looks good so far! Left a few initial comments and I'll try to test it out tonight.

spyder/plugins/editor/widgets/codeeditor.py
CAM-Gerlach: @Khalilsqu You should do this [here](https://github.com/spyder-ide/spyder/pull/8494/files#diff-e9841c32fa6af428b302e4ba35544d0eR890) in #8494 as well, btw.

Khalilsqu: I need more investigation about this, should it go outside if statement or not! I will do it sometime tomorrow.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1582,19 +1582,29 @@ def paste(self):
...
-            text = to_text_string(clipboard.text())
-            if len(text.splitlines()) > 1:
-                eol_chars = self.get_line_separator()
-                text = eol_chars.join((text + eol_chars).splitlines())
-                clipboard.setText(text)
-            # Standard paste
-            TextEditBaseWidget.paste(self)
-            self.document_did_change(text)
+        text = to_text_string(clipboard.text())
+        # This is here to make copied files/folders to be pasted as text path.
```
bcolsen: I'll put on my @CAM-Gerlach hat here but this line would be more concise as.

```python
# Have copied file and folder urls pasted as text paths.`
# See issue: 8566 and PR: 8576 for details.
```

CAM-Gerlach: Heh :) I was itching to point that out but I didn't want to be too @CAM-Gerlach -y

bcolsen: > I was itching to point that out 

I thought you might be ;)

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1582,19 +1582,29 @@ def paste(self):
...
+        # See issue 8566 and PR: 8576 for the details.
+        urls = []
+        if (not clipboard.mimeData().hasFormat('text/plain') and
```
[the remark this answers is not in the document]
CAM-Gerlach: At least in my testing on Fedora 28 LXQt in VBox, files copied to the clipboard did *not* have the ``'text/plain'`` MIME type. However, that check is a holdover from before he implemented the check for ``clipboard.mimeData().hasUrls()``, which presumably should supersede it.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1582,19 +1582,29 @@ def paste(self):
...
+            else:
+                text = urls[0].toLocalFile() + '\n'
+            eol_chars = self.get_line_separator()
+            text = eol_chars.join((text + eol_chars).splitlines())
+            clipboard.setText(text)
+        # Standard paste
+        TextEditBaseWidget.paste(self)
+        self.document_did_change(text)
+        # Restore files/folders into original state instead as text. See #8576.
+        if urls:
```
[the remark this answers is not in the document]
CAM-Gerlach: That might be more [complicated](https://stackoverflow.com/questions/13762140/proper-way-to-copy-a-qmimedata-object) than it seems. As I suggested previously, all this can be avoided by not modifying the clipboard at all, and instead just pass the formatted text directly to ``TextEditBaseWidget.paste()`` (with appropriate modification to take passed text as a parameter, etc).

spyder/plugins/editor/widgets/codeeditor.py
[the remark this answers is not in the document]
bcolsen: It seem files copied from Nautilus have urls and plain/text. Which I guess is why pasting paths copied from nautilus works nicely.

I guess some one should test MacOS too.

spyder/plugins/editor/widgets/codeeditor.py
bcolsen: @Khalilsqu @CAM-Gerlach Should we put quotes around single lines?  It could be quite handy.

spyder/plugins/editor/widgets/codeeditor.py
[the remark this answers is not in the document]
bcolsen: > just pass the formatted text directly to `TextEditBaseWidget.paste()`

That is definitely the way to go. :+1:

spyder/plugins/editor/widgets/codeeditor.py
CAM-Gerlach: Good thought, but IMO no, for the same reasons articulated in #8494 . Users may prefer either single or double quotes in any given context; the single path could be part of a longer string, and the user may well be pasting the path within existing quotes. Meanwhile, if the user wants single or double quotes, the appropriate one can be added with one keystroke with Spyder's auto-quote insertion turned on.

bcolsen: Ok no quotes then. Auto-quote insertion works for me. :+1:

spyder/plugins/editor/widgets/codeeditor.py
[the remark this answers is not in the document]
Khalilsqu: > Why are you adding `not clipboard.mimeData().hasFormat('text/plain')`?
> 
> If it's removed then linux can get some of this formatting help too

In windows, some application even format text data as url, and my testing shows sometimes I cannot paste codes from pycharm to spyder if only `hasUrls()` is used therefore I am also including `not hasFormat('text/plain')`. The later is general but `urls = clipboard.mimeData().urls()` might return nothing if we do not include `hasUrls()`. My manual testing shows that when the clipboard include files/folders both statement `hasUrls()` and `not hasFormat('text/plain')` return True.

spyder/plugins/editor/widgets/codeeditor.py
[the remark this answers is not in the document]
CAM-Gerlach: No, the purpose of this function (``CodeEditor.paste()``) is to handle pasting data, including files and paths, into a ``CodeEditor``; the fact that ``TextEditBaseWidget.paste()`` currently only pulls data from the clipboard rather than allowing it to be passed directly as a string is merely an incidental implementation detail out of convenience, and a poorly chosen one if it requires data in the clipboard to be permanently modified or lost by the simple act of pasting it.

Furthermore, I'm not sure I understand what practical objection there is to this. Adding ``text=None`` to the latter's function signature and pasting ``text`` instead of first getting the clipboard contents is a strict superset of its current functionality while still maintaining full backwards compatibility with the present invocation, so there is no real downside to adding it that I'm aware of.

spyder/plugins/editor/widgets/codeeditor.py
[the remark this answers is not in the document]
CAM-Gerlach: @Khalilsqu Are you deleting your comments or is there a reason I can only see them in my email and not on the current PR thread?

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1582,19 +1582,29 @@ def paste(self):
...
-            TextEditBaseWidget.paste(self)
-            self.document_did_change(text)
+            urls = clipboard.mimeData().urls()
+            if len(urls) > 1:
+                text = ''.join('"' + url.toLocalFile().replace(osp.os.sep, '/')
+                               + '",' + '\n' for url in urls)[:-2]
+            else:
+                text = urls[0].toLocalFile()
+        if len(text.splitlines()) > 1:
+            eol_chars = self.get_line_separator()
```
CAM-Gerlach: Move this to the top of the function, right after the ``clipboard = QApplication.clipboard()`` call per @bcolsen 's previous review.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1582,19 +1582,29 @@ def paste(self):
         text has 'CRLF' EOL chars
```
CAM-Gerlach: Briefly discuss how the file path formatting works in the extended description, and add a one-line summary of the function separated by a blank line at the top per PEP 257.

CAM-Gerlach:
@Khalilsqu Why did you close this and delete your branch??? These were very valuable additions to Spyder that we very much appreciate, and they only needed a few more tweaks to be ready to merge. If we offended you in some way or I was too harsh with my critique, please let us know and we can work things out since you have been a very helpful contributor. Thanks.

PR #8584, opened 2019-01-16
PR: Enclose console warning about rendering of plots

### LINKED ISSUE #8520: Enclose console warning about plots being rendered in the Plots plugin in horizontal bars
The current warning looks like this:

seleccion_002

This can clearly go overlooked because it doesn't have an `\n` at the beginning and also when users print some text and then create a plot.

My idea to improve this is to enclose the warning in top and bottom horizontal bars, like `Remove all variables ...` (which only has a bottom bar):

seleccion_003

Besides, this warning only needs to be printed once per session (i.e. only in the first console a plot is generated) and not every time a plot is created for the first time in each console.

juanis2112:
## Description of Changes

I enclosed the warning about plots being rendered in the Plots plugin in top and bottom horizontal bars to solve issue 8520

screen shot 2019-01-15 at 11 37 44 pm

### Issue(s) Resolved

Fixes #8520

### Affirmation

I certify the above statement is true and correct: Juanis 2112

### DIFF
--- a/spyder/plugins/ipythonconsole/widgets/figurebrowser.py
+++ b/spyder/plugins/ipythonconsole/widgets/figurebrowser.py
@@ -61,11 +61,11 @@ class FigureBrowserWidget(RichJupyterWidget):
                     self.figurebrowser.mute_inline_plotting):
                 if not self.sended_render_message:
                     msg['content']['data']['text/plain'] = self._append_html(
-                      _('<hr>'
-                        '\nFigures now render in the Plots pane by default. '
-                        'To make them also appear inline in the Console, '
-                        'uncheck "Mute Inline Plotting" under the Plots '
-                        'pane options menu. \n<hr>'))
+                        _('<hr>'
+                          '\nFigures now render in the Plots pane by default. '
+                          'To make them also appear inline in the Console, '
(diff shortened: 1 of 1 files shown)


spyder/plugins/ipythonconsole/widgets/figurebrowser.py
```diff
@@ -60,11 +60,11 @@ def _handle_display_data(self, msg):
             if (self.figurebrowser is not None and
                     self.figurebrowser.mute_inline_plotting):
                 if not self.sended_render_message:
-                    msg['content']['data']['text/plain'] = _(
-                        'Figures now render in the Plots pane by default. '
+                    msg['content']['data']['text/plain'] = self._append_html("<hr>"
```
dalthviz: @juanis2112 please enclose all the message using the underscore function (`_`). This is the way we use to manage the translations of the strings (If you see the diff, before the changes, the string was enclosed in this function). Also, please use `'` to enclose the string (here the `<hr>` has `"` and the rest uses `'`)

ok97465:
It seems that the command line does not appear in ipython console after "render_message" is displayed.

error

So I changed the code as follows and the command line is shown.
code

PR #8590, opened 2019-01-17
PR: Adjust attribute icons size on Linux and Windows

ccordoba12:
We did the initial size selection on macOS with @juanis2112, but things need to be adjusted for other platforms.

This is how things look on Linux:

Before:

seleccion_006

After

seleccion_005


jnsebgosselin:
Can you explain a little bit why these changes are needed please? I'm interested by this.

jnsebgosselin:
Ok thanks, I understand. I wish a solution with no hard coded value was possible... maybe I can take a quick look to see if I can propose something, but that would be for another PR.

Also, I think you will need to consider the `%` and `#` icons, or else it will look weird in the Outline Explorer.

ccordoba12:
> they now look rather small and its hard to read the letters on Windows

I'm going to merge this as it is because a) the current icons look rather bad on Linux; and b) I don't have time to avoid the hard-coding I did in this PR.

As usual, we can revisit this later.

PR #8599, opened 2019-01-20
PR: Add ability to delete long nested directories on Windows from File/Project Explorers

### LINKED ISSUE #8567: Deleting a folder which contains a sub-folder(s) raises an error
When deleting a folder which contains subfolders(s), an error is raised saying that cannot delete the folder. But eventually folder's contents are deleted.

delete_folder_issue

Spyder: Master (currently under development) 
OS: windows: 10
Python version: 3.6

Khalilsqu:
## Description of Changes

As promised. Here is my solution of the issue: #8567.
The issue only happens in windows and the reason behind it is explained in [Here](https://github.com/pypa/pip/issues/4910) and [Here](https://github.com/ansible/ansible/issues/34335). 

Although the sub-folder(s) are deleted yet an error is raised stating that they are not empty which are in fact empty. therefore, we delete this sub-folder(s) silently and retry shutil.rmtree within a while loop. 
Another solution is to call the function `shutil.rmtree` recursively and introduce sleep before deleting the folder which raises the error as has been done [here](https://stackoverflow.com/a/25831267) and [here](https://github.com/hashdist/hashdist/blob/a322a66d4bcd4b989a6a163cd2569f3e71995f60/hashdist/core/fileutils.py#L69-L89).
I tested both method and they work fine, but found the implementation in this PR easier without the need to import new modules.

delete_folder
`untitled1.py` is not inside the deleted directory.

### Issue(s) Resolved

Fixes #8567

### Affirmation

@Khalilsqu

### DIFF
--- a/spyder/plugins/explorer/tests/test_explorer.py
+++ b/spyder/plugins/explorer/tests/test_explorer.py
@@ -14,10 +14,11 @@ import os.path as osp
 
 # Third party imports
 import pytest
+from qtpy.QtWidgets import QMessageBox
 
 # Local imports
 from spyder.plugins.explorer.widgets import (FileExplorerTest,
-                                             ProjectExplorerTest, QMessageBox)
+                                             ProjectExplorerTest)
 from spyder.py3compat import to_text_string
 from spyder.plugins.projects.widgets.explorer import ProjectExplorerTest as \
     ProjectExplorerTest2
--- a/spyder/plugins/explorer/tests/test_explorer.py
+++ b/spyder/plugins/explorer/tests/test_explorer.py
@@ -39,11 +39,11 @@ def project_explorer(qtbot):
     return widget
 
 
-@pytest.fixture(params=[['script.py', 'dir1/dir2/dir3/dir4/dir5/dir6/dir7'],
+@pytest.fixture(params=[['script.py', 'dir1/dir2/dir3/dir4'],
                         ['script.py', 'script1.py', 'testdir/dir1/script2.py'],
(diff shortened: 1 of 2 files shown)


spyder/plugins/explorer/tests/test_explorer.py
```diff
@@ -8,12 +8,19 @@
...
+import os
+import os.path as osp
+
+# Third party imports
 import pytest
 
 # Local imports
 from spyder.plugins.explorer.widgets import (FileExplorerTest,
-                                             ProjectExplorerTest)
+                                             ProjectExplorerTest, QMessageBox)
```
ccordoba12: Please import `QMessageBox` like this

`from qtpy.QtWidgets import QMessageBox`

spyder/plugins/explorer/widgets.py
```diff
@@ -603,7 +603,14 @@ def run(self, fnames=None):
     def remove_tree(self, dirname):
         """Remove whole directory tree
         Reimplemented in project explorer widget"""
-        shutil.rmtree(dirname, onerror=misc.onerror)
+        while osp.exists(dirname):
+            try:
+                shutil.rmtree(dirname, onerror=misc.onerror)
+            # Windows Problem with shutil.rmtree. See issue #8567.
```
ccordoba12: Please move this comment to be inside the except block below, like this

```python
except Exception as e:
    # This handles a Windows problem with shutil.rmtree.
    # See issue #8567.
    if type(e).__name__ == "OSError":
```

PR #8616, opened 2019-01-22
PR: Use single row for file name and path and use gray for paths in file switcher

### LINKED ISSUE #8580: Improvements to the File switcher
I've been using VSCode a lot these days and I find its file switcher very neat and well designed:

seleccion_001

in comparison to our own:

seleccion_004

So I'd like to improve it for Spyder 4 with @juanis2112's help. These are my suggestions:

- [ ] Use a single row to show a file name and its path (instead of two, which is what we have right now).
- [ ] Use the same font size for file names as the one used in the symbol switcher (@juanis2112, you can get that one by pressing `Cmd+Alt+P`). Also, use a font size one point smaller than that one for paths.
- [ ] Use a grey color for paths instead of italics.
- [ ] Make the switcher taller (look how many files are shown in VSCode in comparison to our own switcher). We should see at most 15 files on it.
- [ ] Use the newly introduced file type icons to show them to the left of each file name (instead of the generic text file icon).
- [ ] Remove the section header rows we have right now (Editor/Notebook). Instead, show those headers in blue to the right of the first file in the corresponding section. Also show an horizontal line when the section finishes.
- [ ] Add a new section called *Recent files* to the switcher that shows the last 5 or 10 opened files that are not present in the Editor.

@juanis2112, please address each of these improvements in a different pull request.

juanis2112:
## Description of Changes

Changed file switcher to use a single row for a file name and its path and changed paths color to gray.

screen shot 2019-01-25 at 8 10 09 am

screen shot 2019-01-25 at 3 57 25 pm

### Issue(s) Resolved

Part of #8580 (Items 1 and 3)

### Affirmation

Juanis2112

### DIFF
--- a/spyder/widgets/fileswitcher.py
+++ b/spyder/widgets/fileswitcher.py
@@ -229,6 +229,7 @@ class FileSwitcher(QDialog):
     # in a given file when using the '@' symbol.
     FILE_MODE, SYMBOL_MODE = [1, 2]
     MAX_WIDTH = 600
+    PATH_FG_COLOR = 'rgb(153, 153, 153)'
 
     def __init__(self, parent, plugin, tabs, data, icon):
         QDialog.__init__(self, parent)
@@ -663,13 +664,13 @@ class FileSwitcher(QDialog):
                     text_item += " [{0:} {1:}]".format(self.line_count[index],
                                                        _("lines"))
                 if max_width > self.list.width():
-                    text_item += (u" <i style='color:rgb(153,153,153)'>"
-                                  " {1:}</i>").format(
-                            ima.MAIN_FG_COLOR, short_paths[index])
+                    text_item += (u" <span style='color:{0:}'>{1:}"
+                                  "</span>").format(self.PATH_FG_COLOR,
+                                                    short_paths[index])
                 else:
-                    text_item += (u" <i style='color:rgb(153,153,153)'> "
-                                  " {1:}</i>").format(
-                            ima.MAIN_FG_COLOR, paths[index])
+                    text_item += (u" <span style='color:{0:}'>{1:}"
+                                  "</span>").format(self.PATH_FG_COLOR,
+                                                    paths[index])
                 if (trying_for_line_number and self.line_count[index] != 0 or
                         not trying_for_line_number):
                     results.append((score_value, index, text_item))
--- a/spyder/widgets/fileswitcher.py
+++ b/spyder/widgets/fileswitcher.py
@@ -664,11 +664,11 @@ class FileSwitcher(QDialog):
                     text_item += " [{0:} {1:}]".format(self.line_count[index],
                                                        _("lines"))
                 if max_width > self.list.width():
-                    text_item += (u" <span style='color:{0:}'>{1:}"
+                    text_item += (u" &nbsp; <span style='color:{0:}'>{1:}"
                                   "</span>").format(self.PATH_FG_COLOR,
                                                     short_paths[index])
                 else:
-                    text_item += (u" <span style='color:{0:}'>{1:}"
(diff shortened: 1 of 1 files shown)


spyder/widgets/fileswitcher.py
```diff
@@ -663,10 +663,12 @@ def setup_file_list(self, filter_text, current_path):
                     text_item += " [{0:} {1:}]".format(self.line_count[index],
                                                        _("lines"))
                 if max_width > self.list.width():
-                    text_item += u"<br><i style='color:{0:}'>{1:}</i>".format(
+                    text_item += (u" <i style='color:rgb(153,153,153)'>"
```
dalthviz: @juanis2112 please put the `rgb(153,153,153)` in a constant, something like `PATH_FG_COLOR` below [`MAX_WIDTH`](https://github.com/spyder-ide/spyder/blob/4db4465f09c8d48cc26dc7feed539db532ea4677/spyder/widgets/fileswitcher.py#L231) and use it as the first argument (`{0:}`) of the format (replacing the `ima.MAIN_FG_COLOR`). Also I think that we should remove the italics (maybe we can use `p` instead to wrap the path) You should get something similar to this:

```python
text_item += u"<br><p style='color:{0:}'>{1:}</p>".format(
    self.PATH_FG_COLOR, short_paths[index])
```

spyder/widgets/fileswitcher.py
```diff
@@ -663,10 +663,12 @@ def setup_file_list(self, filter_text, current_path):
...
+                                  " {1:}</i>").format(
                             ima.MAIN_FG_COLOR, short_paths[index])
                 else:
-                    text_item += u"<br><i style='color:{0:}'>{1:}</i>".format(
+                    text_item += (u" <i style='color:rgb(153,153,153)'> "
```
dalthviz: Please use here the new `PATH_FG_COLOR` constant in a similar way as explained above

dalthviz:
About @CAM-Gerlach comment, what do you think @ccordoba12 ? From what I see in VSCode the paths are align to the left. Although, they seem kind of squished as @CAM-Gerlach says, maybe we need to add more space between the path and the file name?)

spyder/widgets/fileswitcher.py
CAM-Gerlach: Is ``#999999`` a standard color already in use in our dark theme? If not, can we pick one that is, so our color scheme is more consistant and we don't have a dozen different shades of gray floating around our codebase?

CAM-Gerlach:
Can we at least have a screenshot to see how it looks with them aligned to the right? Keeping them left might be fine, but they definitely need either more space or a separator of some sort.

ccordoba12:
This is looking really good, thanks @juanis2112!

> About @CAM-Gerlach comment, what do you think @ccordoba12 ?

I don't like this idea too much, but we can add a bit more space between the file name and its path. @juanis2112, please do that in this PR.

ccordoba12:
@juanis2112, please update the screenshot in the PR description above to see your latest changes.

dalthviz:
About the color used for the path, @CAM-Gerlach can you make some suggestions? Also @juanis2112 could you upload a screenshot with the light theme to see how it looks like, thanks :)

ccordoba12:
@juanis2112, please also upload a screenshot for the light theme.

CAM-Gerlach:
Actually, @dalthviz . ``#999999`` may actually be the best choice after all, since the closest QDarkstyle palette color is ``#787878`` which is even lower contrast than it is currently, when it already is at the lower limit of what I'd consider legible. I didn't realize it, but ``#999999`` is also what we use at the comment color in the Spyder Dark theme, so it has precedent.

PR #8636, opened 2019-01-24
PR: Improve warning for when only some namespace objs couldn't be saved

CAM-Gerlach:
## Description of Changes

We should presumably merge spyder-ide/spyder-kernels#81 first, just in case, so I marked it WiP (it should be done otherwise).

Once spyder-ide/spyder-kernels#81 is merged and released as part of Spyder-Kernels 0.4.0, Spyder-Kernels will include, as part of the error message, a list of the objects from the namespace that did not serialize successfully, although the operation as a whole did succeed. In that case, Spyder will currently state ``Unable to save current workspace`` followed by ``Some objects could not be saved: obj1, obj2, obj3...``. To be clearer for users while still fully backward and forward compatible in both directions, with the changes in this PR Spyder will automatically detect these strings and display a more appropriate warning message:

image

Also, it updates the copyright date in the readme and about dialog (and adds a link to the AUTHORS file in the latter).

### Affirmation

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 ![Spyder — The Scientific Python Development Environment](
 ./img_src/spyder_readme_banner.png)
 
-*Copyright © 2009–2018 [Spyder Project Contributors](
+*Copyright © 2009–2019 [Spyder Project Contributors](
 https://github.com/spyder-ide/spyder/graphs/contributors)*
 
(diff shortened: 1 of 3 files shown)


CAM-Gerlach:
@ccordoba12 This one is now ready to go on my end.

PR #8647, opened 2019-01-26
PR: Provide graphical options to configure the PyLS

### LINKED ISSUE #3064: Allow users to configure PEP8 options

I would like to have a GUI option in the  Tools->Preferences->Editor menu in Spyder so as to increase maximum line length in Spyder Editor and enforce remaining PEP8 standards
1. 
2. 
3. 

Pls. also refer to comment by Carlos in the below link: 
http://stackoverflow.com/questions/26426689/python-how-to-tell-spyders-style-analysis-pep8-to-read-from-a-setup-cfg-or-inc
- Spyder Version: 2.3.8
- Python Version:2.7
- Operating system:Windows
## Dependencies

and paste the contents below:

### LINKED ISSUE #8816: Deleting a server in the LSP preferences pane when only one is present raises exceptions
## Problem Description

With default preferences, deleting the only server listed in the LSP preference pane will cause an exception immediately, and further exceptions when clicking anywhere within the pane. Clearly, these routines are expecting at least one server to be present, and not handling the case if none are.

1. Select only (default) LSP server in the list in the LSP preference pane.
2. Press ``Delete currently selected server``

Immediately on clicking delete

```python-traceback

  File "C:\Users\C. A. M. Gerlach\Documents\dev\SpyderDev\spyder\spyder\preferences\languageserver.py", line 646, in delete_server
    self.table.delete_server(idx)
  File "C:\Users\C. A. M. Gerlach\Documents\dev\SpyderDev\spyder\spyder\preferences\languageserver.py", line 554, in delete_server
    self.adjust_cells()
  File "C:\Users\C. A. M. Gerlach\Documents\dev\SpyderDev\spyder\spyder\preferences\languageserver.py", line 524, in adjust_cells
    self.setColumnWidth(CMD, max(names))
ValueError: max() arg is an empty sequence

```

Clicking in the pane thereafter:

```python-traceback
  File "C:\Users\C. A. M. Gerlach\Documents\dev\SpyderDev\spyder\spyder\preferences\languageserver.py", line 605, in mouseDoubleClickEvent
    self.show_editor()
  File "C:\Users\C. A. M. Gerlach\Documents\dev\SpyderDev\spyder\spyder\preferences\languageserver.py", line 561, in show_editor
    server = self.source_model.row(idx)
  File "C:\Users\C. A. M. Gerlach\Documents\dev\SpyderDev\spyder\spyder\preferences\languageserver.py", line 478, in row
    return self.servers[row_num]
IndexError: list index out of range
```

## Versions

* Spyder version: 4.0.0.dev0 a411d61d7
* Python version: 3.7.2
* Qt version: 5.9.6
* PyQt5 version: 5.9.2

### Dependencies

```
sympy >=0.7.3     :  1.3 (OK)
```

ccordoba12:
This PR will provide a graphical interface to configure most of the options provided by the PyLS.

Fixes #8816.
Fixes #3064


spyder/preferences/languageserver.py
```diff
@@ -608,30 +610,70 @@ def mouseDoubleClickEvent(self, event):
...
+        # --- Other servers tab ---
+        # Section label
+        servers_label = QLabel(
+            _("Spyder uses the <a href=\"{lsp_url}\">Language Server "
+              "Protocol</a> technology to provide code completion and linting "
+              "for its Editor. Here you can define how to start a server for "
+              "other programming languages besides Python, along with its "
+              "respective configuration. After doing that, Spyder will be "
+              "able to give completions and linting for the files associated "
+              "with that language."
```
CAM-Gerlach: ``Here you can [...] associated with that language`` -> ``Here, you can setup and configure LSP servers for languages other than Python, so Spyder can provide such features for those languages as well.``

spyder/preferences/languageserver.py
```diff
@@ -608,30 +610,70 @@ def mouseDoubleClickEvent(self, event):
 class LSPManagerConfigPage(GeneralConfigPage):
     """Language Server Protocol manager preferences."""
     CONF_SECTION = 'lsp-server'
-    NAME = _('Language Server Protocol')
+    NAME = _('Code completion and linting')
```
CAM-Gerlach: We should make this shorter, e.g. just ``Code Analysis`` or come up with our own umbrella name for completion, linting, introspection, etc. like Microsoft did (I vote for ``Spyder Sense`` :) ) Otherwise, it makes the sidebar way too wide and leaves less space for the rest of the preferences. (I also plan to shorten ``Current working directory`` to ``Working directory`` in my UI text pass when Spyder 4 is near-final).

ccordoba12: I understand what you mean but I think it's better to be clear and explicit with the section name so that people understands what it stands for. `Code Analysis`, while short, is extremely vague and hard to grasp in my opinion.

CAM-Gerlach: Right; it is important to be clear and explicit with the name. What about ``Completion and linting? Is that clear enough?

Aside from that, @jnsebgosselin @jitseniesen any suggestions on a clear umbrella term for completion/introspection/linting (real-time-analysis) functionality? Or are we stuck with enumerating them every time (that has been an issue in other contexts too).

ccordoba12: > Completion and linting

I like this suggestion :+1:

goanpeca:
@ccordoba12 you want me to continue this work?

ccordoba12:
@goanpeca, don't worry, @dalthviz is going to finish this one.

CAM-Gerlach:
@ccordoba12 @dalthviz Let me know when its ready for a final review pass, paying particular attention to UI, UX and text.

goanpeca:
Could we paste a screenshot @dalthviz

ccordoba12:
When things are ready, @dalthviz will add the necessary screenshots.

goanpeca:
Question: Even if at the moment and for the foreseeable future we are going to support python only, should we namespace all these options to `python` (filename, config options, etc...)?

ccordoba12:
> Question: Even if at the moment and for the foreseeable future we are going to support python only, should we namespace all these options to python (filename, config options, etc...)?

Yes, for now that's be the case. I don't know what kind of UI we could create to expose similar options for other languages and not making things confusing at the same time.

However, other servers can be fully configured (using json though) thanks to @andfoy's previous work.

goanpeca:
>Yes, for now that's the case. I don't know what kind of UI we could create to expose similar options for other languages and not making things confusing at the same time.

I do not mean at the UI level, I meant at how the info is stored, accessed and the name of the files. Seems like all the config options could be nested under a `python` key or something.

ccordoba12:
> I do not mean at the UI level, I meant at how the info is stored, accessed and the name of the files. Seems like all the config options could be nested under a python key or something.

Oh, I see. We give special treatment to Python by not namespacing its options, but for all other languages we create a `lang: {options}` dict in the `lsp-server` section of `spyder.ini`.

dalthviz:
A preview:

## Code completion tab
imagen

## Linting tab

imagen

## Code style tab

imagen

## Docstring style tab

imagen

## Advanced tab

imagen

## Other languages tab

imagen

ccordoba12:
@CAM-Gerlach, @goanpeca, I think this is ready. We added a graphical counterpart to configure most of the options provided by the PyLS (not all of them are worth to provide a graphical option, though).

@CAM-Gerlach, please help us to review the texts we added so we can merge it. Thanks!

spyder/preferences/languageserver.py
```diff
@@ -608,30 +607,356 @@ def mouseDoubleClickEvent(self, event):
...
+        code_style_widget = QWidget()
+        code_style_widget.setLayout(code_style_layout)
+
+        # --- Docstring tab ---
+        # Docstring style label
+        numpy_url = (
+            "<a href='https://numpydoc.readthedocs.io/en/"
+            "latest/format.html'>Numpy</a>")
+        pep257_url = (
+            "<a href='https://www.python.org/dev/peps/pep-0257/'>Pep 257</a>")
```
CAM-Gerlach: ``PEP 257``. Be consistent.

spyder/preferences/languageserver.py
```diff
@@ -608,30 +607,356 @@ def mouseDoubleClickEvent(self, event):
...
+        code_style_layout.addWidget(code_style_label)
+        code_style_layout.addWidget(code_style_check)
+        code_style_layout.addWidget(code_style_g_widget)
+
+        code_style_widget = QWidget()
+        code_style_widget.setLayout(code_style_layout)
+
+        # --- Docstring tab ---
+        # Docstring style label
+        numpy_url = (
```
CAM-Gerlach: ``numpydoc_url``

spyder/preferences/languageserver.py
```diff
@@ -608,30 +607,356 @@ def mouseDoubleClickEvent(self, event):
...
+            "<a href='https://numpydoc.readthedocs.io/en/"
+            "latest/format.html'>Numpy</a>")
```
CAM-Gerlach: ``Numpydoc``

spyder/preferences/languageserver.py
```diff
@@ -608,30 +607,356 @@ def mouseDoubleClickEvent(self, event):
...
+        pep257_url = (
+            "<a href='https://www.python.org/dev/peps/pep-0257/'>Pep 257</a>")
+        docstring_style_codes = _(
+            "<a href='http://www.pydocstyle.org/en/stable"
+            "/error_codes.html'>page</a>")
+        docstring_style_label = QLabel(
+            _("Here you can decide if you want to perform style analysis on "
+              "your docstrings according to the {} or {} conventions. You can "
+              "also decide if you want to show or ignore specific errors, "
+              "according to the codes found on this {}.").format(
```
CAM-Gerlach: ``Spyder can analyze your docstrings with pydocstyle for conformance to the basic {pep257_url} standard or the specific {numpydoc_url} convention. You can manually show or hide specific warnings by their <a href={pydocstyle_link}>pydocstyle error codes</a>.``

CAM-Gerlach: TODO: (Later PR, if not simple) Add Googledoc, since we have Googledoc generation already?

spyder/preferences/languageserver.py
```diff
@@ -608,30 +607,356 @@ def mouseDoubleClickEvent(self, event):
...
+        # Docstring style checkbox
+        docstring_style_check = self.create_checkbox(
+            _("Enable docstring style linting"),
+            'pydocstyle')
+
+        # Docstring style options
+        docstring_style_convention = self.create_combobox(
+            _("Choose the convention used to lint docstrings: "),
+            (("Numpy", 'numpy'),
+             ("Pep 257", 'pep257')),
```
CAM-Gerlach: Assuming I understand what the various fields currently do correctly (and if not, that's a problem), to simplify the options and reduce duplication and confusion on this tab, I suggest including an option "Custom" in this list that makes the ``Show/ignore the following errors`` fields behave as exclusive overrides (i.e. *not* in addition to the specified convention), while otherwise they act in addition to such, and eliminate the extra set of duplicate fields.

CAM-Gerlach: Also, ``PEP 257`` (compare ``Pep 257`` viz. ``PEP8``...) and ``Numpydoc``.

spyder/preferences/languageserver.py
```diff
@@ -608,30 +607,356 @@ def mouseDoubleClickEvent(self, event):
...
+            'pydocstyle/convention')
+        docstring_style_add_select = self.create_lineedit(
+            _("Display the following errors in addition to the "
```
CAM-Gerlach: Consistency. ``Show the following errors:``

spyder/preferences/languageserver.py
```diff
@@ -608,30 +607,356 @@ def mouseDoubleClickEvent(self, event):
...
+              "specified convention:"),
+            'pydocstyle/add_select', alignment=Qt.Horizontal, word_wrap=False,
+            placeholder=_("Example codes: D100, D200"))
+        docstring_style_add_ignore = self.create_lineedit(
+            _("Ignore the following errors in addition to "
```
CAM-Gerlach: ``Ignore the following errors:``

spyder/preferences/languageserver.py
```diff
@@ -608,30 +607,356 @@ def mouseDoubleClickEvent(self, event):
...
+              "the specified convention:").format(code_style_codes),
+            'pydocstyle/add_ignore', alignment=Qt.Horizontal, word_wrap=False,
+            placeholder=_("Example codes: D400, D405"))
+        docstring_style_select = self.create_lineedit(
+            _("Select the following errors to show:"),
```
CAM-Gerlach: See above.

spyder/preferences/languageserver.py
```diff
@@ -608,30 +607,356 @@ def mouseDoubleClickEvent(self, event):
...
+            'pydocstyle/select', alignment=Qt.Horizontal, word_wrap=False,
+            placeholder=_("Example codes: D413, D414"))
+        docstring_style_ignore = self.create_lineedit(
+            _("Ignore the following errors:"),
```
CAM-Gerlach: See above.

spyder/preferences/languageserver.py
```diff
@@ -608,30 +607,356 @@ def mouseDoubleClickEvent(self, event):
...
+        advanced_layout.addWidget(advanced_label)
+        advanced_layout.addLayout(advanced_g_layout)
+        advanced_widget.setLayout(advanced_layout)
+
+        # --- Other servers tab ---
+        # Section label
+        servers_label = QLabel(
+            _("Spyder uses the <a href=\"{lsp_url}\">Language Server "
+              "Protocol</a> technology to provide code completion and linting "
+              "for its Editor. Here you can setup and configure LSP servers "
```
CAM-Gerlach: ``Here, ``

spyder/preferences/languageserver.py
```diff
@@ -608,30 +607,356 @@ def mouseDoubleClickEvent(self, event):
...
+        advanced_layout = QVBoxLayout()
+        advanced_layout.addWidget(advanced_label)
+        advanced_layout.addLayout(advanced_g_layout)
+        advanced_widget.setLayout(advanced_layout)
+
+        # --- Other servers tab ---
+        # Section label
+        servers_label = QLabel(
+            _("Spyder uses the <a href=\"{lsp_url}\">Language Server "
+              "Protocol</a> technology to provide code completion and linting "
```
CAM-Gerlach: I think you can eliminate ``technology``...

spyder/preferences/languageserver.py
```diff
@@ -608,30 +607,356 @@ def mouseDoubleClickEvent(self, event):
...
+        # Servers table
+        table_group = QGroupBox(_('Available servers:'))
         self.table = LSPServerTable(self, text_color=ima.MAIN_FG_COLOR)
+        table_layout = QVBoxLayout()
+        table_layout.addWidget(self.table)
+        table_group.setLayout(table_layout)
+
+        # Buttons
         self.reset_btn = QPushButton(_("Reset to default values"))
         self.new_btn = QPushButton(_("Setup a new server"))
```
CAM-Gerlach: ``Set up`` is the verb (action), ``Setup`` is the noun/adjective

spyder/preferences/languageserver.py
ccordoba12: > TODO: (Later PR, if not simple) Add Googledoc, since we have Googledoc generation already?

Pydocstyle doesn't not provide such an option, just Numpy and Pep 257

goanpeca:
(Actually is Carlos and @dalthviz who did all this work @CAM-Gerlach  :-) )

spyder/preferences/languageserver.py
CAM-Gerlach: Thanks.

CAM-Gerlach:
Oh, sorry @goanpeca and @dalthviz ! Updated my review accordingly.

spyder/preferences/languageserver.py
dalthviz: But then how we can differenciate between this option and the one below? (add_ignore vs ignore or add_select vs select)?

CAM-Gerlach: See my [comment](https://github.com/spyder-ide/spyder/pull/8647#discussion_r275167990) above, which replaces the two sets of fields with one, controlled by the dropdown.

dalthviz: Oh didn't see that, I will check :+1:

spyder/utils/introspection/module_completion.py
```diff
@@ -17,11 +17,18 @@
...
+# List of preferred modules
+PREFERRED_MODULES = ['numpy', 'scipy', 'sympy', 'pandas', 'networkx',
+                     'statsmodels', 'matplotlib', 'sklearn', 'skimage',
+                     'mpmath', 'os', 'PIL', 'OpenGL', 'array', 'audioop',
+                     'binascii', 'cPickle', 'cStringIO', 'cmath',
+                     'collections', 'datetime', 'errno', 'exceptions', 'gc',
+                     'imageop', 'imp', 'itertools', 'marshal', 'math', 'mmap',
+                     'msvcrt', 'nt', 'operator', 'parser', 'rgbimg', 'signal',
+                     'strop', 'sys', 'thread', 'time', 'wx', 'xxsubtype',
+                     'zipimport', 'zlib', 'nose', 'PyQt4', 'PyQt5', 'PySide',
```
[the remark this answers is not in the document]
ccordoba12: Ok. @dalthviz, please change `nose` for `pytest`. I think it's fine to leave `PyQt4` (I don't see any harm on doing it).

spyder/preferences/languageserver.py
```diff
@@ -608,30 +607,356 @@ def mouseDoubleClickEvent(self, event):
...
+
+        # --- Code completion ---
+        # Introspection group
+        introspection_group = QGroupBox(_("Introspection"))
+        completion_box = newcb(_("Enable code completion"), 'code_completion')
+        goto_definition_box = newcb(
+            _("Go to definitions"),
+            'jedi_definition',
+            tip=_("If this option is enabled, doing a left mouse click on\n"
+                  "an object name while pressing the Ctrl key will go to\n"
```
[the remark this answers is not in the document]
CAM-Gerlach: We don't handle this more centrally (e.g. the same as in the Shortcuts Summary)?

spyder/preferences/languageserver.py
```diff
@@ -608,30 +607,356 @@ def mouseDoubleClickEvent(self, event):
...
+                  "that object's definition (if resolved)."))
+        follow_imports_box = newcb(_("Follow imports when going to a "
+                                     "definition"),
+                                   'jedi_definition/follow_imports')
```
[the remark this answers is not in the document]
CAM-Gerlach: Oh, okay.

spyder/preferences/languageserver.py
```diff
@@ -608,30 +607,356 @@ def mouseDoubleClickEvent(self, event):
...
+        introspection_group = QGroupBox(_("Introspection"))
+        completion_box = newcb(_("Enable code completion"), 'code_completion')
+        goto_definition_box = newcb(
+            _("Go to definitions"),
+            'jedi_definition',
+            tip=_("If this option is enabled, doing a left mouse click on\n"
+                  "an object name while pressing the Ctrl key will go to\n"
+                  "that object's definition (if resolved)."))
+        follow_imports_box = newcb(_("Follow imports when going to a "
+                                     "definition"),
```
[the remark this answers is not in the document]
CAM-Gerlach: I was referring to the same thing as the above, sorry, which you agreed it was a suboption. But its a moot point since like you said we don't have the UI yet.

spyder/utils/introspection/module_completion.py
```diff
@@ -17,11 +17,18 @@
...
 
-# Modules database
-modules_db = PickleShareDB(MODULES_PATH)
+# List of preferred modules
+PREFERRED_MODULES = ['numpy', 'scipy', 'sympy', 'pandas', 'networkx',
+                     'statsmodels', 'matplotlib', 'sklearn', 'skimage',
+                     'mpmath', 'os', 'PIL', 'OpenGL', 'array', 'audioop',
+                     'binascii', 'cPickle', 'cStringIO', 'cmath',
+                     'collections', 'datetime', 'errno', 'exceptions', 'gc',
+                     'imageop', 'imp', 'itertools', 'marshal', 'math', 'mmap',
```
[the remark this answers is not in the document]
CAM-Gerlach: Also, to be clear, ``pillow`` should be added.

spyder/utils/introspection/module_completion.py
```diff
@@ -17,11 +17,18 @@
...
+                     'msvcrt', 'nt', 'operator', 'parser', 'rgbimg', 'signal',
+                     'strop', 'sys', 'thread', 'time', 'wx', 'xxsubtype',
```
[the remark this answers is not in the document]
CAM-Gerlach: You might want to add ``threading`` if the reason to include ``thread`` originally is valid.

spyder/utils/introspection/module_completion.py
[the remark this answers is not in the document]
CAM-Gerlach: :+1:

spyder/preferences/languageserver.py
```diff
@@ -608,30 +607,356 @@ def mouseDoubleClickEvent(self, event):
...
+
+        # --- Linting ---
+        # Linting options
+        linting_label = QLabel(_("Here you can decide if you want to get "
+                                 "syntax errors and informative warnings "
+                                 "about the code you write in the editor."))
+        linting_label.setOpenExternalLinks(True)
+        linting_label.setWordWrap(True)
+        linting_check = self.create_checkbox(
+            _("Enable linting"),
```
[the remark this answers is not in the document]
CAM-Gerlach: Okay, sure.

spyder/preferences/languageserver.py
```diff
@@ -608,30 +607,356 @@ def mouseDoubleClickEvent(self, event):
...
+            '<a href="https://www.python.org/dev/peps/pep-0008">PEP8</a>')
+        code_style_codes = _(
+            "<a href='http://pycodestyle.pycqa.org/en/latest"
+            "/intro.html#error-codes'>page</a>")
+        code_style_label = QLabel(
+            _("Here you can decide if you want to perform style analysis on "
+              "your code according to the {} convention. You can also "
+              "decide if you want to show or ignore specific errors or "
+              "warnings of that convention, according to the codes "
+              "found on this {}.").format(pep_url, code_style_codes))
```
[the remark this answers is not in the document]
CAM-Gerlach: > Pep 8 is a convention, not a standard

I agree, which is why I in fact didn't call it the "PEP 8 standard" and noted and retained your use of "convention" instead. Used before the word as I did, it has the same meaning as "standard" in "standard practice", i.e. something widely used and accepted by the community (as PEP 8 is), but not *itself* a standard. However, if you think that's confusion, I'm fine with just cutting it.

> ... You can also manually ...

:+1:

spyder/preferences/languageserver.py
```diff
@@ -608,30 +607,356 @@ def mouseDoubleClickEvent(self, event):
...
+            'pycodestyle/select', alignment=Qt.Horizontal, word_wrap=False,
+            placeholder=_("Example codes: E113, W391"))
+        code_style_ignore = self.create_lineedit(
+            _("Ignore the following errors or warnings:"),
+            'pycodestyle/ignore', alignment=Qt.Horizontal, word_wrap=False,
+            placeholder=_("Example codes: E201, E303"))
+        code_style_max_line_length = self.create_spinbox(
+            _("Maximum allowed line length:"), None,
+            'pycodestyle/max_line_length', min_=10, max_=500, step=1,
+            tip=_("Default is 79"))
```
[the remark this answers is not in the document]
CAM-Gerlach: Okay, sure.

spyder/preferences/languageserver.py
CAM-Gerlach: Alright, I guess...its not that important.

spyder/preferences/languageserver.py
CAM-Gerlach: That does make it clearer for PEP 257 and Numpydoc. My concern was that it doesn't make 100% sense for custom since it isn't in addition to anything.

spyder/preferences/languageserver.py
CAM-Gerlach: Discussed above.

spyder/preferences/languageserver.py
CAM-Gerlach: Yes

spyder/preferences/languageserver.py
CAM-Gerlach: Yes, sorry.

spyder/preferences/languageserver.py
CAM-Gerlach: > the fields Show the following errors and Ignore the following errors need to be disabled.

What I'm suggesting is instead of including two essentially duplicate sets of fields, only one of which is ever useful at a time, is to just have one set of fields (``Show in addition [...]``/``Ignore in addition``) and have Custom make them be in addition to nothing, i.e. an absolute whitelist/blacklist (as the lower set of fields, without the "in addition", does now). Otherwise, there isn't much point, IMO, in having Custom other than adding an extra step to do the same thing, as it adds rather than reduces complexity in our preferences for only a modest gain in clarity.

spyder/preferences/languageserver.py
```diff
@@ -608,30 +607,356 @@ def mouseDoubleClickEvent(self, event):
...
+
+        advanced_layout = QVBoxLayout()
+        advanced_layout.addWidget(modules_textedit)
+        advanced_group.setLayout(advanced_layout)
+
+        # --- Linting ---
+        # Linting options
+        linting_label = QLabel(_("Here you can decide if you want to get "
+                                 "syntax errors and informative warnings "
+                                 "about the code you write in the editor."))
```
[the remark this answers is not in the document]
CAM-Gerlach: Whereas I find the above to be far too verbose and syntactically-tortured for the information it conveys. That said, I agree I don't like the imperative tone of my suggestion either.

Approaching it from a different angle, if one simply cuts all the *italic* portions from the current version: "*Here you can decide if you want to* get syntax errors and informative warnings about the code *you write* in the Editor" the meaningful information content is exactly the same (``Get syntax errors and informative warnings about the code in the Editor.``) while being much shorter and with much less complex syntax.

Alternatively, what about ``Spyder can optionally display syntax errors and informative warnings about the code in the Editor.`` or ``Spyder can optionally highlight syntax errors and possible problems with your code in the Editor.``?

spyder/preferences/languageserver.py
ccordoba12: Yes, that's why this field is going to be disabled when `Custom` is active.

spyder/preferences/languageserver.py
ccordoba12: Well, the problem is Pydocstyle and the PyLS make a difference between these options, so we need to configure them separately.

Ok, what about this:

1. Remove `Select/Ignore the following errors` options.
2. When Numpy or PEP 257 are selected, the labels of the options should be `Show/Ignore the following errors in addition to the specified convention`.
3. When `Custom` is selected, we change the text of those labels to `Show/ignore the following errors`.
4. We set the right options under the hood.

spyder/preferences/languageserver.py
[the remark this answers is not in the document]
ccordoba12: I like these suggestions better. I think we should go for the second one:

> Spyder can optionally highlight syntax errors and possible problems with your code in the editor

@dalthviz, please make this change.

spyder/preferences/languageserver.py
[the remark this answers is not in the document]
CAM-Gerlach: Gotcha, thanks.

spyder/preferences/languageserver.py
CAM-Gerlach: :+1: Agreed, that's exactly what I'm suggesting.

spyder/preferences/languageserver.py
```diff
@@ -608,30 +608,392 @@ def mouseDoubleClickEvent(self, event):
...
+            self.docstring_style_select.label.setVisible(True)
+            self.docstring_style_select.textbox.setVisible(True)
+            self.docstring_style_ignore.label.setVisible(True)
+            self.docstring_style_ignore.textbox.setVisible(True)
+
+            # Hide add_select and add_ignore options
+            self.docstring_style_add_select.label.setVisible(False)
+            self.docstring_style_add_select.textbox.setVisible(False)
+            self.docstring_style_add_ignore.label.setVisible(False)
+            self.docstring_style_add_ignore.textbox.setVisible(False)
```
ccordoba12: @dalthviz, what we agreed to do with @CAM-Gerlach is the following:

1. Remove `docstring_style_add_select` and `docstring_style_add_ignore` options, and only leave `docstring_style_select` and `docstring_style_ignore`.
2. When `Numpy` or `PEP 257` are selected as conventions, then:
    1. The labels on those options must display `Show the following errors in addition to the specified convention:` and `Ignore the following errors in addition to the specified convention:`.
    2. The values saved on those options need to be passed to `addSelect`/`addIgnore` in the `update_python_config` method of `LSPManager`.
    3, `select` and `ignore` in `update_python_config` need to be made empty.
3. When `Custom` is selected, then:
    1. The labels must display `Show the following errors:` and `Ignore the following errors:`.
    2. The values of those options need to be passed to `select` and `ignore` in  `update_python_config`.
    3. `addSelect`/`addIgnore` in `update_python_config` need to be made empty.

I know this is a bit confusing to explain. Please let me know if it's clear for you or you need me to clarify things a bit more.

ccordoba12:
@dalthviz, please address the other changes in labels required by @CAM-Gerlach, so we can merge this PR. Thanks!

ccordoba12:
> Will the options to control automatic "dot" completion and case-sensitive completion, previously present at Editor > Code Introspection/Analysis, be added here in the future? It would be nice to not have a regression over the old system.

No because the missing old options can't be mapped to the new architecture. Besides, there's little value on exposing some of those options (e.g. configuring `Enter` to *not* select an item from the code completion widget).

CAM-Gerlach:
> No because the missing old options can't be mapped to the new architecture.

Case-sensitive completion isn't a huge loss since I'm not 100% sure how many people want it, and I presume it would require LSP-level configuration if its even possible rather than something on the Spyder side. But isn't "dot" completion triggered by the Spyder side? I assume we can address this later when we add on the fly completion, which is still planned for Spyder 4, correct?

> Besides, there's little value on exposing some of those options (e.g. configuring Enter to not select an item from the code completion widget).

Right, that's why I only mentioned those two specific ones.

ccordoba12:
> But isn't "dot" completion triggered by the Spyder side?

No, it's controlled by the PyLS now. However, we can still decide if the want to follow the list of `trigger completion characters` sent from the server

> I assume we can address this later when we add on the fly completion, which is still planned for Spyder 4, correct?

Yes, that's for beta3.

CAM-Gerlach:
> However, we can still decide if the want to follow the list of trigger completion characters sent from the server

Thanks. Do you want to add that here or later (e.g. as part of the options for on the fly)?

ccordoba12:
Let's leave that for later. This PR is big enough already.

PR #8657, opened 2019-01-29
PR: Change font size for file name and path in the file switcher

### LINKED ISSUE #8580: Improvements to the File switcher

juanis2112:
## Description of Changes

I changed the font size for the file name and the path in the file switcher to solve issue 8580 item 2.

The following screenshot is using the default rich font size.
screen shot 2019-01-30 at 11 41 59 am

The following screenshot is setting the rich font size to 15pt.
screen shot 2019-01-30 at 11 42 37 am

### Issue(s) Resolved

Part of #8580 Item 2 

### Affirmation

Juanis2112

### DIFF
--- a/spyder/widgets/fileswitcher.py
+++ b/spyder/widgets/fileswitcher.py
@@ -656,14 +656,14 @@ class FileSwitcher(QDialog):
         self.fix_size(paths)
 
         # Build the text that will appear on the list widget
-        PATH_TEXT_FONT_SIZE = CONF.get('appearance', 'rich_font/size', 11)
-        FILENAME_TEXT_FONT_SIZE = PATH_TEXT_FONT_SIZE + 2
+        path_text_font_size = CONF.get('appearance', 'rich_font/size', 11)
+        filename_text_font_size = path_text_font_size + 2
         for index, score in enumerate(scores):
             text, rich_text, score_value = score
             if score_value != -1:
                 text_item = ("<span style='color:{0:}; font-size:{1:}pt'>{2:}"
                              "</span>").format(ima.MAIN_FG_COLOR,
-                                               FILENAME_TEXT_FONT_SIZE,
+                                               filename_text_font_size,
                                                rich_text.replace('&', ''))
                 if trying_for_line_number:
                     text_item += " [{0:} {1:}]".format(self.line_count[index],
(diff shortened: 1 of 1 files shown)


spyder/widgets/fileswitcher.py
```diff
@@ -658,17 +658,20 @@ def setup_file_list(self, filter_text, current_path):
         for index, score in enumerate(scores):
             text, rich_text, score_value = score
             if score_value != -1:
-                text_item = "<big style='color:{0:}'>{1:}</big>".format(
-                        ima.MAIN_FG_COLOR, rich_text.replace('&', ''))
+                text_item = ("<span style='font-size: 13pt; color:{0:}'>{1:}"
```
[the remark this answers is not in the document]
dalthviz: Yep you are right @ccordoba12 that's better. Please @juanis2112 instead of using the hard-coded value create variables that get the current rich font size (for the paths) and calculate the size for the filenames, something like:
```python
        path_text_font_size = CONF.get('main', 'rich_font/size', 11)
        filename_text_font_size = path_text_font_size + 2
```
And use these variables in the `formats` :)

dalthviz:
@CAM-Gerlach I think you are right about the paths issue, although for simplicity this PR is just going to add the use of the preferences to manage the size of the paths. I will add the paths length issue as an element of #8580 :+1:

ccordoba12:
I agree with @dalthviz's here. This PR is very well simple and focused, so let's not digress the discussions towards other topics.

spyder/widgets/fileswitcher.py
```diff
@@ -655,21 +656,29 @@ def setup_file_list(self, filter_text, current_path):
         self.fix_size(paths)
 
         # Build the text that will appear on the list widget
+        path_text_font_size = CONF.get('appearance', 'rich_font/size', 11)
+        filename_text_font_size = path_text_font_size + 2
```
ccordoba12: Why did you use `+ 2` here instead of `+ 1`? Is there too little difference with only one point between filename and path_text_font_size and filename_text_font_size?

ccordoba12: @juanis2112? or @dalthviz?

dalthviz: Yep, the difference seemed too little

dalthviz: imagen

or what do you think?

CAM-Gerlach:
> This PR is very well simple and focused, so let's not digress the discussion towards other topics.

@ccordoba12 Right; the only reason I brought it up is because without that, this PR is strictly speaking a net negative for Spyder's usability, since the primary motivation for making them smaller is to allow for a longer path to be displayed (like VSCode has).

PR #8665, opened 2019-01-29
PR: Move config pages of missing plugins to their own modules

ccordoba12:
This is just more refactoring to make it easier to find where the config pages of all modules are located.


goanpeca:
@ccordoba12 it might be better to use `git move` so that history is preserved

ccordoba12:
I'm moving part of the `plugin.py`'s files to a new file, not the entire file itself. That's why I didn't use `git mv`.

Rafa said he was using a trick for this but I couldn't find a simple and good enough solution myself.

CAM-Gerlach:
> it might be better to use git move so that history is preserved, since these are renames

I thought there was no functional difference between ``git mv`` and moving it with the OS and then adding both, since git tracks moves heuristically via contents as opposed to some bespoke system? Or are you referring to some script/plugin actually called ``git move``?

PR #8700, opened 2019-02-03
PR: Automatic docstring generation for functions

### LINKED ISSUE #4936: Feature request: automatic generation of docstring template
Hi,

One thing that I think will be very useful for Spyder users (like myself) is automatic generation of docstring templates for functions and classes. PyCharm actually has this useful functionality. To give a concrete example, let's say I have code as shown below:

screenshot from 2017-08-10 18-18-01

If now I just type three quotation marks `"""` right under the `add` function and hit Enter, a template docstring will be generated for me:

screenshot from 2017-08-10 18-18-29

In addition, if your function yields instead of returns, `Yields` instead of `Returns` will be generated. If you raise an exception in your code, a `Raises` block will be included in the docstring template. This may requirement some more code search/inference.

Pycharm now also allows you to choose docstring styles among numpydoc, Google, reStructuredText, and others. I think editors like Sublime and Atom also have plugins that achieve the exact same thing. In addition, I know there's something called `pyment` (https://github.com/dadadel/pyment) that works as a command-line program and can do this to your python code files.

Is it possible to build this same thing into the Spyder editor.

Thanks very much for your consideration.

ok97465:
## Description of Changes

### Issue(s) Resolved
- Use the context menu.
docstring_mouse

- Use shortcut (ctrl+alt+d) when the textcusor placed on 'def' line or below function definition.
docstring_shortcut

- Use """ trigger when the textcusor placed below function definition.
  - When the popup menu is shown, if a input of keyboard is not the "enter key", the menu is closed and the input is inserted to editor.
enter

Fixes #4936

### Affirmation

I certify the above statement is true and correct: ok97465


ok97465:
I do not know if my approach is correct.
If the problem with my code is serious, I will close the PR.

Thank you for reading my PR.

ccordoba12:
This looks really, really nice! Thanks @ok97465 for your work on this, it seems you work very hard on this! Please don't worry about the failure in Travis, that's unrelated to your work.

For now I have one comment to make: instead of using an option, could it be possible to introduce a docstring template by a keyboard shortcut? I think that would be simpler for most users because sometimes you'll have Spyder to give you a template and others you don't. And right now (by activating this behavior with an option) you have to select one or the other, but you can't have both at the same time.

ok97465:
Thank you very much for your replys @ccordoba12 @CAM-Gerlach .

> For now I have one comment to make: instead of using an option, could it be possible to introduce a docstring template by a keyboard shortcut?

Does "instead of using an option" mean the following things?
  - remove EditorExtension for triggering (""" +Enter).
  - remove checkbox for auto docstring in confpage.

And adding shortcuts is not likely to be difficult. I'll try it on the weekend.

ccordoba12:
> So this isn't redundant with any of the functionality that can be done through the LSP architecture?

No, the LSP can't provide this functionality.

ccordoba12:
> Does "instead of using an option" mean the following things?

It means:

1. "remove checkbox for auto docstring in confpage." (as you said).
2. And, instead of using `""" + Enter` (which is what you added), please add a shortcut (it could be `Ctrl+Alt+D`) that detects if there's a triple quote behind the current mouse position, if that triple quote is placed below of what it seems to be a callable, and adds the docstring template there.

ccordoba12:
However, Pycharm does this with a context menu:

https://www.jetbrains.com/help/pycharm/creating-documentation-comments.html

and I think that would be much easier to implement. @CAM-Gerlach, what do you think?

CAM-Gerlach:
@ccordoba12 If I understand it correctly, that's not really a context menu but rather a dedicated "Refactor/Snippits tools"-type menu that's triggered via a dedicated shortcut, like our Tab and dot completion menu. If that's what you're suggesting, I definitely agree that having a menu pop up when ``"""`` is typed with that option (similar to that with automatic "dot" completion) is a better approach than the keyboard shortcut or the automatic option since it both is easily triggered/ignored on demand like a shortcut but is also much more discoverable to most users like the automatic option, who might never know the keyboard shortcut existed.

impact27:
Something that could also be interesting is to add a warning if the doc string doesn't correspond to the function (If I rename a variable after generating the doc string), and eventually to be able to update the doc string? I don't know how hard that would be.

ok97465:
@impact27 Thank you for your opinion. I know the function you mentioned. And It was a useful feature when using Pycharm.

Probably the first suggestion seems to be related to pyls, so it seems hard to solve in this PR.
The second suggestion would be better to proceed with the other PR after completing this PR.

CAM-Gerlach:
@ok97465 Let us know when this is ready for review!

ok97465:
I implemented 3 UI.
- UI1 : using """ + Enter


- UI2 : using Popup after typing """
docstring_popup
  - After typing ''', if there was no input for 300ms, show the context menu.
  - Cons. : UX2 can not be used when Closequotes is disabled.

- UI3 : using shortcut
docstring_shortcut

docstring_mouse

ok97465:
@CAM-Gerlach Please let me know what to fix and I will make up for the weak points. Thank you very much.

CAM-Gerlach:
@ok97465 Sure; taking a look at it now.

spyder/plugins/editor/extensions/tests/test_docstring.py
```diff
@@ -0,0 +1,225 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+#
+"""Tests for close quotes."""
+# Third party imports
```
CAM-Gerlach: Add a blank line here to cleanly separate the ``import``s from the header and docstring (it looks strange to have one line between import sections, and two after, but none before, and doesn't match our convention elsewhere).

spyder/plugins/editor/extensions/tests/test_docstring.py
```diff
@@ -0,0 +1,225 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+#
```
CAM-Gerlach: Missing last line of header.

spyder/plugins/editor/extensions/docstring.py
```diff
@@ -0,0 +1,441 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+"""Generate Docstring."""
```
CAM-Gerlach: Add a blank line here to cleanly separate the ``import``s from the header and docstring (it looks strange to have one line between import sections, and two after, but none before, and doesn't match our convention elsewhere).

spyder/widgets/mixins.py
```diff
@@ -399,7 +399,7 @@ def get_current_line_to_cursor(self):
     def get_line_number_at(self, coordinates):
         """Return line number at *coordinates* (QPoint)"""
         cursor = self.cursorForPosition(coordinates)
-        return cursor.blockNumber()-1
+        return cursor.blockNumber()+1
```
CAM-Gerlach: Per PEP 8: Use spaces on both sides of a binary operator.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -2668,6 +2673,12 @@ def setup_context_menu(self):
             self, _("Zoom reset"), shortcut=QKeySequence("Ctrl+0"),
             triggered=self.zoom_reset.emit)
 
+        # Docstring
+        writer = self.writer_docstring
+        self.docstring_action = create_action(
+            self, _("Write Docstring"),
```
CAM-Gerlach: I suggest ``Generate function docstring`` as clearer UI text (also note our general convention is Sentance case, not Title Case, though we're not 100% consistent about it).

ok97465:
@CAM-Gerlach  Thank you for your great review. I learned a lot through your review.
UI2 can only be activated in previous commit. I'm sorry I did not let you know in advance.

> * The `Returns` doesn't list the return values my testing even on relatively trivial functions; it always says `None`.

I changed None to RETURN_TYPE.

> * The placeholder `[type]` and `[description]` should not include the brackets, since they aren't selected when the user doubleclicks the word, uses Ctrl-Shift-Left/Right or presses Ctrl-Del to clear the placeholder and insert their own text. Instead, you could e.g. make them ALL CAPS and/or add a character that will select, like `_` to each side.

I used the uppercase for the placeholder.

>* The user's cursor should be placed on the line below the `"""` to allow them to type the one-line summary and the extended description; this usage is both more common and looks much cleaner for a multi-line, numpy-style docstring IMO vs. putting it on the same line as the `"""`.

I changed it your way.

>* Furthermore, we should also list the shortcut next to the context menu item; otherwise there is no obvious way for the user to discover it.

I added shortcut name.

>* Similarly, there's no need to offer the option for single-line docstrings, and making it the default no less, especially if the user has to deliberately press the shortcut to get it—it would no faster or more convenient than triple-tapping `"""`. (However, if it would be easy to do, you could make close quotes insertion insert a closing `"""` automatically (here or in another PR), which would have the same effect but be much more useful.)

I removed single-line option.

>* If the trigger mechanism is going to be a shortcut, it doesn't make sense to require the user to type `"""` first, since they are explicitly requesting an automatic docstring be generated, so it should work just like the `Write docstring` menu item at least as long as the cursor is either on the `def` line or the line below it.

I have a question about ```the cursor is either on the `def` line or the line below it```. The trigger mechanism uses a TextCursor. So if the textcursor is below def line, the trigger mechanism is activated.
The shortcut of context menu uses mouse cursor. So if mouse cursor is on the 'def' line, shortcurt is operated.
If the textcusor is below def line and the mouse cursor is on the another def line, shortcurt is not working well. It is difficult to use one shortcut for context menu and trigger mechanism.

In my last commit, the context menu and shortcut is working only when mouse cursor is on the def line.

I hope my sentence is well communicated.

CAM-Gerlach:
Don't worry about the test failures for now; they are unrelated to this PR.

> I removed single-line option.

Very good, but the preferences option for picking docstring type is still there. This will be needed once we support multiple docstring types, but for now its not needed so you could comment it out with an additional comment explaining this.

> It is difficult to use one shortcut for context menu and trigger mechanism.

Okay; I guess not worrying about having it work after the opening ``"""`` or otherwise on the first line below the def statement (which it currently doesn't) and only considering the def statement it is immediately on, can you just have it use the position of the Textcursor when triggered via the shortcut (as users expect) and the position of the mousecursor when triggered from the context menu, but still display the shortcut key in the context menu?

>  If UX2 is used with close quotes, it is annoying for someone who use closequote but don't use docstring.

Can you make it so that the ``Generate docstring`` item isn't selected by default in the popup, so the user can simply press enter or keep typing and the popup will go away without doing anything (thus making for minimal annoyance other than a brief visual popup if the user pauses after typing ``"""``), and to actually select the ``Generate docstring`` they need to press the down arrow key once?

So there's no way to call the functionality of CloseQuotes without actually activating it for the user? @bcolsen any ideas? Thanks.

spyder/plugins/editor/extensions/__init__.py
```diff
@@ -13,4 +13,5 @@
 
 from .closebrackets import CloseBracketsExtension
 from .closequotes import CloseQuotesExtension
+from .docstring import WriterDocstring
```
bcolsen: `WriterDocstring` would be more recognizable as an extension if it was called `DocstringWriterExtension`

spyder/plugins/editor/extensions/docstring.py
```diff
@@ -0,0 +1,451 @@
...
+                if c == quote and text[i - 1] != '\\':
+                    pos[left_pos] = i
+                    is_found_left_quote = False
+
+        if is_found_left_quote:
+            raise IndexError("No matching close quote at: " + str(left_pos))
+
+        return pos
+
+    def find_bracket_position(self, text, bracket_left, bracket_right,
```
bcolsen: You might be able to use this code to match brackets instead:
https://github.com/spyder-ide/spyder/blob/76edb5222ad8196791f6fc7177cf5eded1b6f542/spyder/plugins/editor/widgets/base.py#L449

spyder/plugins/editor/confpage.py
```diff
@@ -179,6 +179,10 @@ def enable_tabwidth_spin(index):
         removetrail_box = newcb(_("Automatically remove trailing spaces "
                                   "when saving files"),
                                 'always_remove_trailing_spaces', default=False)
+        docstring_combo_choices = ((_("numpy"), 'numpy'),)
+        docstring_combo = self.create_combobox("Docstring type",
+                                               docstring_combo_choices,
+                                               'docstring_type')
```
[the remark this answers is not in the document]
ok97465: I added GoogleDoc.

spyder/plugins/editor/extensions/__init__.py
ok97465: I renamed the classname.

spyder/plugins/editor/extensions/docstring.py
ok97465: In order to use "find_brace_match", other parts of the code should also be modified.

ok97465:
> can you just have it use the position of the Textcursor when triggered via the shortcut (as users expect) and the position of the mousecursor when triggered from the context menu, but still display the shortcut key in the context menu?

Yes. I fix it.

CAM-Gerlach:
Thanks for the revision, @ok97465 ! I'll review it when I get home from work later.

CAM-Gerlach:
Sorry for the delay reviewing this, @ok97465 ; I'm working on it now. I ran into serious issues this past week with my Anaconda install that ended up taking several days to resolve, and I then had some other commitments and catch-up to do after finally getting that working again.

spyder/plugins/editor/extensions/tests/test_docstring.py
```diff
@@ -0,0 +1,355 @@
...
+        pos = editor.mapToGlobal(pos)
+        writer.line_number_cursor = editor.get_line_number_at(pos)
+        writer.write_docstring_at_first_line_of_function()
+
+    assert editor.toPlainText() == expected
+
+
+# =============================================================================
+# ---- Example for manual Testing of numpy docstring
+# =============================================================================
```
CAM-Gerlach: I appreciate the effort, but this really should not be in the test file, but rather in a separate one, if present at all.

spyder/plugins/editor/confpage.py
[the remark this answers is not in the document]
CAM-Gerlach: Nice!!! Fantastic job!

spyder/plugins/editor/extensions/docstring.py
CAM-Gerlach: How difficult would that be to implement? That seems the one last issue holding up this PR, as it would greatly improve the usability and discoverability of this feature to have the automatic popup along with the existing shortcut/context menu command that users are unlikely to find on their own.

ok97465: @CAM-Gerlach Do you want to add UX2? or change "find_bracket_position" to "find_brace_match"?

spyder/plugins/editor/extensions/tests/test_docstring.py
CAM-Gerlach: I appreciate the effort, but this really should not be in the test file, but rather in a separate one, if present at all.

spyder/plugins/editor/extensions/docstring.py
CAM-Gerlach: Oh sorry, maybe I was confused—I thought that was proposed as a way to work around UX2 relying on the CloseQuotes extension being enabled by the user.

ok97465:
> Sorry for the delay reviewing this, @ok97465 ; I'm working on it now. I ran into serious issues this past week with my Anaconda install that ended up taking several days to resolve, and I then had some other commitments and catch-up to do after finally getting that working again.

I'm alright. It is meaningless for me to merge my code to spyder-ide quickly. I'm ok that you review my pr slowly.

spyder/plugins/editor/extensions/docstring.py
ok97465: I will add the UX2 to this PR this weekend.

CAM-Gerlach:
> It is meaningless for me to merge my code to spyder-ide quickly.

I appreciate your patience, but the longer I delay, the more chance there is of a merge conflict which means more work for you, which I'd hate to see. Almost done with my review.

spyder/plugins/editor/extensions/docstring.py
```diff
@@ -0,0 +1,517 @@
...
+                              pos_quote):
+        r"""Return the start and end position of pairs of brackets.
+
+        https://stackoverflow.com/questions/29991917/
+        indices-of-matching-parentheses-in-python
+        """
+        pos = {}
+        pstack = []
+
+        for i, c in enumerate(text):
```
CAM-Gerlach: Please be more clear and readable with your variable names—``idx`` and ``character`` perhaps?

spyder/plugins/editor/extensions/docstring.py
```diff
@@ -0,0 +1,517 @@
...
+                pstack.append(i)
+            elif c == bracket_right and not self.is_char_in_pairs(i,
+                                                                  pos_quote):
+                if len(pstack) == 0:
+                    raise IndexError(
+                        "No matching closing parens at: " + str(i))
+                pos[pstack.pop()] = i
+
+        if len(pstack) > 0:
+            raise IndexError(
```
CAM-Gerlach: You catch these somewhere else rather than dumping them at the user, right?

spyder/plugins/editor/extensions/tests/test_docstring.py
```diff
@@ -0,0 +1,355 @@
...
+        ('def foo():', '', [], [], [], None),
+        (""" def foo(arg0, arg1=':', arg2: str='-> (float, str):') -> \
+             (float, int): """,
+         ' ', ['arg0', 'arg1', 'arg2'], [None, None, 'str'],
+         [None, "':'", "'-> (float, str):'"],
+         '(float, int)')
+    ])
+def test_information_of_function(text, indent, name_list, type_list,
+                                 value_list, rtype):
+    """Test FunctionInfo."""
```
CAM-Gerlach: Can you be a little more descriptive here?

spyder/plugins/editor/extensions/docstring.py
CAM-Gerlach: Thanks! Looking forward to it!

spyder/plugins/editor/extensions/docstring.py
```diff
@@ -0,0 +1,517 @@
...
+            number_of_lines_of_function += 1
+
+            if cur_text.endswith(':'):
+                return func_text, number_of_lines_of_function
+
+            cursor.movePosition(QTextCursor.NextBlock)
+
+        return None
+
+    def get_function_definition_from_below_last_line(self):
```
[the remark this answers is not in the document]
ok97465: I fixed to work the shortcut below the line.

spyder/plugins/editor/extensions/docstring.py
```diff
@@ -0,0 +1,517 @@
...
+                    is_found_left_quote = True
+                    quote = c
+                    left_pos = i
+            else:
+                if c == quote and text[i - 1] != '\\':
+                    pos[left_pos] = i
+                    is_found_left_quote = False
+
+        if is_found_left_quote:
+            raise IndexError("No matching close quote at: " + str(left_pos))
```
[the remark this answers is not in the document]
ok97465: I changed find_quote_position method to private method(_find_quote_position).
This method is used only in split_args_text_to_list method.

spyder/plugins/editor/extensions/docstring.py
ok97465: I fixed it

spyder/plugins/editor/extensions/docstring.py
ok97465: I changed find_bracket_position method to private method(_find_bracket_position).
This method is used only in split_args_text_to_list method.

spyder/plugins/editor/extensions/docstring.py
```diff
@@ -0,0 +1,517 @@
...
+
+        return pos
+
+    def find_bracket_position(self, text, bracket_left, bracket_right,
+                              pos_quote):
+        r"""Return the start and end position of pairs of brackets.
```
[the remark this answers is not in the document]
ok97465: I fixed it.

spyder/plugins/editor/extensions/docstring.py
```diff
@@ -0,0 +1,517 @@
...
+        docstring_triggers = ['"""', 'r"""', "'''", "r'''"]
+        if text.lstrip() in docstring_triggers:
+            return True
+
+        return False
+
+    def get_function_definition_from_first_line(self):
+        """
+        Get the definition of function when there is the cursor at first
+        line of function definition.
```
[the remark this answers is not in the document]
ok97465: I fixed it.

ok97465:
> Get any lines beginning with `raise` in the function body and, if any are found, inserting a `Raises` section as specified in each docstring format with the name of each exception and a `DESCRIPTION` placeholder.

I added a 'Raises' section.

> Check if `yield` is found in the function body, and retitle the `Returns` section accordingly

I added the functionality for checking yield.

> implementing "UX2"

I added UX2 but my ux2 is working with CloseQuotes. I have no idea for implementing UX2 without CloseQuotes.

ok97465:
> Properly fill in the `Returns` (`Yields`) section

In the following case, it is difficult to get return name.
```python
def foo(nx)
    if nx == 1:
        value = 1
        return value
    else:
        arr = np.ones(nx)
        return arr
```

The return statement in the above case should be written as below.
```python
"""
    Returns
    --------
    Union[float, ndarray]
"""
```
  
  
It is not easy to find the return variable name.
```python
def foo(nx)
    return np.ones((nx, ny, \
        nz)) 
```
  
  
So it is difficult for me to fill the return section properly.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3064,6 +3088,40 @@ def visible_blocks(self):
...
+
+        writer = self.writer_docstring
+        if writer.get_function_definition_from_below_last_line():
+            point = self.cursorRect().bottomRight()
+            point = self.calculate_real_position(point)
+            point = self.mapToGlobal(point)
+
+            self.menu_docstring = QMenuOnlyForEnter(self)
+            self.docstring_action = create_action(
+                self, _("Write docstring"), icon=ima.icon('TextFileIcon'),
```
CAM-Gerlach: Again, like in the context menu, this should be ``Generate docstring`` for clarity and consistency.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3064,6 +3088,40 @@ def visible_blocks(self):
     def is_editor(self):
         return True
 
+    def popup_docstring(self, prev_text, prev_pos):
+        """Show the menu for generating docstring"""
```
CAM-Gerlach: Missing period.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3064,6 +3088,40 @@ def visible_blocks(self):
...
+                self, _("Write docstring"), icon=ima.icon('TextFileIcon'),
+                triggered=writer.write_docstring)
+            self.menu_docstring.addAction(self.docstring_action)
+            self.menu_docstring.setActiveAction(self.docstring_action)
+            self.menu_docstring.popup(point)
+
+    def delayed_popup_docstring(self):
+        """Show context menu for docstring.
+        This method is called after typing '''. After typing ''', This function
+        wait 300ms. If there was no input for 300ms, show the context menu.
```
CAM-Gerlach: Put the summary line on its own line (below the ``"""``) and make sure to leave a blank line between it and the extended description, to follow PEP 257.

CAM-Gerlach: Also, lowercase ``T`` in ``this function`` since it doesn't begin a sentence, and should be ``waits 300 ms``.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -2671,6 +2678,13 @@ def setup_context_menu(self):
             self, _("Zoom reset"), shortcut=QKeySequence("Ctrl+0"),
             triggered=self.zoom_reset.emit)
 
+        # Docstring
+        writer = self.writer_docstring
+        self.docstring_action = create_action(
+            self, _("Generate function docstring"),
```
CAM-Gerlach: Sorry for changing my mind on this, but ``Generate docstring`` would be just as clear and more general (since it already generates method docstrings, and at some point could be extended to classes and modules.

spyder/plugins/editor/extensions/docstring.py
```diff
@@ -0,0 +1,615 @@
...
+            pos_round = self._find_bracket_position(args_text, '(', ')',
+                                                    pos_quote)
+            pos_curly = self._find_bracket_position(args_text, '{', '}',
+                                                    pos_quote)
+            pos_square = self._find_bracket_position(args_text, '[', ']',
+                                                     pos_quote)
+        except IndexError:
+            return None
+
+        while 1:
```
CAM-Gerlach: ``while True:`` is clearer, more explicit and idiomatic.

spyder/plugins/editor/extensions/docstring.py
CAM-Gerlach: I don't see any difference here, sorry.

spyder/plugins/editor/extensions/docstring.py
```diff
@@ -0,0 +1,615 @@
...
+
+        return False
+
+    @staticmethod
+    def _find_quote_position(text):
+        """Return the start and end position of pairs of quotes."""
+        pos = {}
+        is_found_left_quote = False
+
+        for i, c in enumerate(text):
```
CAM-Gerlach:
```suggestion
        for idx, character in enumerate(text):
```

CAM-Gerlach:
> So it is difficult for me to fill the return section properly.

The algorithm I outlined above and provide below should handle even complex cases like the one you mentioned on a "best-effort" basis, while being conservative to avoid filling in incorrect information and not requiring too much parsing or code complexity. I initially was just going to write up a much simpler version in pseudocode, but I ended up just writing a full, working example, minus the function to retrieve the return values from the function body (I provided a stub so its a fully working example), and you'll want to replace the marked constants at the top with the appropriate user-selected docstring format. Also, you might want to consider replacing my bespoke bracket-matching code with something from CloseBrackets, though I'm not sure it would work if that's disabled.

> In the following case, it is difficult to get return name.

Currently, the algorithm I've proposed would just pick the longest as a "best guess", but I've revised it below to handle this more conservatively by listing the placeholder ``RETURN_TYPE`` and ``DESCRIPTION`` (or multiple, if both have tuples of equal length).

> It is not easy to find the return variable name.

Although the above algorithm takes a conservative/easy approach and simply lists the placeholder in that case, the algorithm I've implemented below handles it more smartly, printing the name of the first variable and putting a placeholder for the second.

The return parsing algorithm I've whipped up:

```python
import re

# FIXME Indent chars from the Editor, set based on user prefs (as you do elsewhere)
EDITOR_INDENT = '    '

# FIXME These should be set based on chosen docstring format
header = 'Returns\n-------\n'
return_element_type = '{return_type}\n' + EDITOR_INDENT + 'DESCRIPTION.'
placeholder = return_element_type.format(return_type='RETURN_TYPE')
return_element_name = '{return_name} : ' + placeholder


def get_all_return_yield_values_in_function_body():
    """
    Returns a list of return values (the part after return) as strings.

    Concatenates continuation lines after stripping the surrounding whitespace,
    comments and newlines on each line.

    """
    return_vals = [
        '"f, b", v1, v2, 3.0, .7, (,), {}, [ab], f(a), None, a.b, a+b, True',
        '"f, b", v1, v3, 420, 5., (,), {}, [ab], f(a), None, a.b, a+b, False',
        ]
    return return_vals


# TODO: Replace this with something involving CloseBrackets???
def find_top_level_bracket_locations(string_toparse):
    """Get the locations of top-level brackets in a string."""
    bracket_stack = []
    replace_args_list = []
    bracket_type = None
    literal_type = ''
    brackets = {'(': ')', '[': ']', '{': '}'}
    for idx, character in enumerate(string_toparse):
        if (not bracket_stack and character in brackets.keys()
                or character == bracket_type):
            bracket_stack.append(idx)
            bracket_type = character
        elif bracket_type and character == brackets[bracket_type]:
            begin_idx = bracket_stack.pop()
            if not bracket_stack:
                if not literal_type:
                    if bracket_type == '(':
                        literal_type = '(None)'
                    elif bracket_type == '[':
                        literal_type = '[list]'
                    elif bracket_type == '{':
                        if idx - begin_idx <= 1:
                            literal_type = '{dict}'
                        else:
                            literal_type = '{set}'
                replace_args_list.append((string_toparse[begin_idx:idx + 1],
                                          literal_type, 1))
                bracket_type = None
                literal_type = ''
        elif len(bracket_stack) == 1:
            if bracket_type == '(' and character == ',':
                literal_type = '(tuple)'
            elif bracket_type == '{' and character == ':':
                literal_type = '{dict}'
            elif bracket_type == '(' and character == ':':
                literal_type = '[slice]'

    if bracket_stack:
        raise IndexError('Bracket mismatch')
    for replace_args in replace_args_list:
        string_toparse = string_toparse.replace(*replace_args)
    return string_toparse


def parse_return_elements(return_vals_group):
    """Return the appropriate text for a group of return elements."""
    all_eq = (return_vals_group.count(return_vals_group[0])
              == len(return_vals_group))
    if all([{'[list]', '(tuple)', '{dict}', '{set}'}.issuperset(
            return_vals_group)]) and all_eq:
        return return_element_type.format(
            return_type=return_vals_group[0][1:-1])
    # Output placeholder if special Python chars present in name
    py_chars = {' ', '+', '-', '*', '/', '%', '@', '<', '>', '&', '|', '^',
                '~', '=', ',', ':', ';', '#', '(', '[', '{', '}', ']', ')', }
    if any([any([py_char in return_val for py_char in py_chars])
            for return_val in return_vals_group]):
        return placeholder
    # Output str type and no name if only string literals
    if all(['"' in return_val or '\'' in return_val
            for return_val in return_vals_group]):
        return return_element_type.format(return_type='str')
    # Output bool type and no name if only bool literals
    if {'True', 'False'}.issuperset(return_vals_group):
        return return_element_type.format(return_type='bool')
    # Output numeric types and no name if only numeric literals
    try:
        [float(return_val) for return_val in return_vals_group]
        num_not_int = 0
        for return_val in return_vals_group:
            try:
                int(return_val)
            except ValueError:  # If not an integer (EAFP)
                num_not_int = num_not_int + 1
        if num_not_int == 0:
            return return_element_type.format(return_type='int')
        elif num_not_int == len(return_vals_group):
            return return_element_type.format(return_type='float')
        else:
            return return_element_type.format(return_type='numeric')
    except ValueError:  # Not a numeric if float conversion didn't work
        pass
    # If names are not equal, don't contain "." or are a builtin
    if ({'self', 'cls', 'None'}.isdisjoint(return_vals_group) and all_eq
            and all(['.' not in return_val
                     for return_val in return_vals_group])):
        return return_element_name.format(return_name=return_vals_group[0])
    return placeholder


def generate_docstring_return_section():
    """Generate the Returns section of a function/method docstring."""
    return_vals = get_all_return_yield_values_in_function_body()

    # If all return values are None, return none
    non_none_vals = [return_val for return_val in return_vals
                     if return_val and return_val != 'None']
    if not non_none_vals:
        return header + 'None.'

    # Get only values with matching brackets that can be cleaned up
    non_none_vals = [return_val.strip(' ()\t\n').rstrip(',')
                     for return_val in non_none_vals]
    non_none_vals = [re.sub('([\"\'])(?:(?=(\\\\?))\\2.)*?\\1',
                            '"string"', return_val)
                     for return_val in non_none_vals]
    unambiguous_vals = []
    for return_val in non_none_vals:
        try:
            cleaned_val = find_top_level_bracket_locations(return_val)
        except IndexError:
            continue
        unambiguous_vals.append(cleaned_val)
    if not unambiguous_vals:
        return header + placeholder

    # If remaining are a mix of tuples and not, return single placeholder
    single_vals, tuple_vals = [], []
    for return_val in unambiguous_vals:
        (tuple_vals.append(return_val) if ',' in return_val
         else single_vals.append())
    if single_vals and tuple_vals:
        return header + placeholder

    # If return values are tuples of different length, return one placeholder
    if tuple_vals:
        num_elements = [return_val.count(',') + 1
                        for return_val in tuple_vals]
        if num_elements.count(num_elements[0]) != len(num_elements):
            return header + placeholder
        num_elements = num_elements[0]
    else:
        num_elements = 1

    # If all have the same len but some ambiguous return that many placeholders
    if len(unambiguous_vals) != len(non_none_vals):
        return header + '\n'.join([placeholder for __ in range(num_elements)])

    # Handle tuple (or single) values position by position
    return_vals_grouped = zip(*[
        [return_element.strip() for return_element in return_val.split(',')]
        for return_val in unambiguous_vals])
    return_elements_out = []
    for return_vals_group in return_vals_grouped:
        return_elements_out.append(parse_return_elements(return_vals_group))

    return header + '\n'.join(return_elements_out)
```

CAM-Gerlach:
Also, @ok97465 your yield checking needs to be refined to check that, e.g., whitespace is present before and after ``yield`` and that it is the first statement on a given line, etc., since it incorrectly concludes that ``generate_docstring_return_section`` has a ``yield`` statement just because it is in the name of the `` get_all_return_yield_values_in_function_body()`` that function calls.

ok97465:
> I ended up just writing a full

Thank you very much (I do not know any other way to express gratitude in English.). 
I'll try it this weekend.

CAM-Gerlach:
> Thank you very much (I do not know any other way to express gratitude in English.).

That's exactly it! We're very grateful ourselves for you taking the time to work on this very comprehensive, substantial and useful PR and being very patient with all our suggestions and feedback.

I intended to only write some simple pseudocode to outline the algorithm, but one thing led to another which led to another (as often happens with me; I started out making a few straightforward changes to the Lektor-Icon theme, and it eventually turned into into spyder-ide/lektor-icon#1 ). and I ended up spending a good half a day developing a much more comprehensive, working version of the real thing.

Looking forward to seeing the final result in action! You might want to consider manually testing it on lots of different functions (should be fairly quick to do) to make sure it doesn't encounter anything that triggers an error/exception that stops the whole thing, and then perhaps enclosing the ``generate_docstring_return_section()`` function call in a try-except Exception that falls back to the current placeholder text in case of an exception, to avoid raising a user-visible error and not producing any docstring output at all if the function runs into trouble parsing a return statement for some reason (it would be nice to test and fix every possible case and it should catch the vast majority of them, but there are so many branches and potential inputs that it isn't really practical).

CAM-Gerlach:
A rather straightforward enhancement to the above (that I can try to add if I get some time) is if a string can be multiple literal values or a variable and one or more literals, to actually list the types rather than just putting in a generic placeholder (e.g. ``str or None``). Also, for return values that are only string or numeric literals, I could list them in a set, although this probably isn't very common.

CAM-Gerlach:
@ok97465 Sorry for the delay; I wasn't sure if you were done making changes since I didn't see a reply comment, and then I got really busy finishing a big grant proposal that I just submitted today.

spyder/plugins/editor/extensions/docstring.py
```diff
@@ -0,0 +1,851 @@
...
+
+                is_found_return = False
+                line_return_tmp = ''
+
+
+class QMenuOnlyForEnter(QMenu):
+    """
+    If a input of keyboard is the "enter key", the selected action is executed.
+    If a input of keyboard is not the "enter key", the menu is closed and
+    the input is inserted to code editor.
```
CAM-Gerlach: For a multi-line docstring, per PEP 257, first line should be a summary, followed by a blank line, followed by the full description.

ccordoba12:
@ok97465, please rebase only if it's strictly necessary (i.e. there are conflicts between your work and master).

ok97465:
@ccordoba12 Yes, I will. 
Is "merge with master" available  frequently during developing the branch?

ccordoba12:
There's no need of merging with master either. The reason we only ask contributors to rebase/merge with master when needed is because a) Rebases make the review history here on Github messy; and b) merges make the `git log` graph history messy.

So please do one or the other (preferably merges) only if Github reports that there is a merge conflict.

ok97465:
> There's no need of merging with master either. The reason we only ask contributors to rebase/merge with master when needed is because a) Rebases make the review history here on Github messy; and b) merges make the `git log` graph history messy.

Thank you for your kind explanation. :thumbsup:

ccordoba12:
@CAM-Gerlach, please add an entry about this in

https://github.com/spyder-ide/spyder/wiki/Beta-version-changelog

under the Editor section, since you understand better than me how this functionality works.

CAM-Gerlach:
@ccordoba12 Thanks, done. Have we been doing this comprehensively for other major PRs, or should I look through them and add other new features (this is why one reason taking a second to add labels and such to PRs really helps, BTW).

ok97465:
@CAM-Gerlach Thank you very much for your review, suggestion and code. Thanks to your help, this functionality have been upgraded much more than originally thought. So I have been able to use it really useful for a few days. :thumbsup:

PR #8734, opened 2019-02-08
PR: Remove blanks if no content is added in a line and Enter is pressed to create a new line

### LINKED ISSUE #8564: New line whitespaces conflict with PEP
### Issue Report Checklist

* [ x ] Searched the [issues page](https://github.com/spyder-ide/spyder/issues?q=is%3Aissue) for similar reports
* [ x ] Read the relevant sections of the [Spyder Troubleshooting Guide](https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ) and followed its advice
* [ x ] Reproduced the issue after updating with ``conda update spyder`` (or ``pip``, if not using Anaconda)
* [ x ] Completed the **Problem Description**, **Steps to Reproduce** and **Version** sections below

## Problem Description
When creating a new line by `return` in an indented environment, spyder makes whitespaces to keep the indentation. However, when subsequently keeping the line empty by pressing `return` again, those whitespaces make a PEP 8 warning .

1. Open a new file
2. Create an indentation, e.g. using an `if`
3 .Add some code
3. Have a blank line by double tapping return and then some more code

A PEP8 warning occurs with whitespaces in blank line. My suggestion would be to catch if a line is intentionally blank and then remove the whitespaces needed for indentation. However, when the user decide to add code, the whitespaces  have to be added manually to be on the right indentation which may be tedious.

impact27:
## Description of Changes

When a line is left, by pressing return, the mouse, or a direction arrow, if the line was empty except for spaces, remove the spaces.

### Issue(s) Resolved

Fixes #8564

### Affirmation

impact27

### DIFF
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -227,6 +227,7 @@ DEFAULTS = [
               'autosave_enabled': True,
               'autosave_interval': 60,
               'docstring_type': 'Numpydoc',
+              'strip_trailing_spaces_on_modify': True,
               }),
             ('historylog',
              {
--- a/spyder/plugins/editor/confpage.py
+++ b/spyder/plugins/editor/confpage.py
@@ -122,6 +122,12 @@ class EditorConfigPage(PluginConfigPage):
                   "of a line (when this option is enabled, code\n"
                   "completion may be triggered using the alternate\n"
                   "shortcut: Ctrl+Space)"))
+        strip_mode_box = newcb(
+            _("Automatically strip trailing spaces on changed lines"),
+            'strip_trailing_spaces_on_modify', default=True,
+            tip=_("If enabled, modified lines of code (excluding strings)\n"
+                  "will have their trailing whitespace stripped when leaving them.\n"
+                  "If disabled, only whitespace added by Spyder will be stripped."))
         ibackspace_box = newcb(
             _("Intelligent backspace"),
             'intelligent_backspace',
@@ -178,6 +184,7 @@ class EditorConfigPage(PluginConfigPage):
         sourcecode_layout.addWidget(tab_mode_box)
         sourcecode_layout.addWidget(ibackspace_box)
         sourcecode_layout.addWidget(removetrail_box)
+        sourcecode_layout.addWidget(strip_mode_box)
         sourcecode_layout.addLayout(indent_tab_layout)
 
         sourcecode_widget = QWidget()
--- a/spyder/plugins/editor/plugin.py
+++ b/spyder/plugins/editor/plugin.py
@@ -1205,6 +1205,7 @@ class Editor(SpyderPluginWidget):
             ('set_tab_stop_width_spaces',           'tab_stop_width_spaces'),
             ('set_wrap_enabled',                    'wrap'),
             ('set_tabmode_enabled',                 'tab_always_indent'),
+            ('set_stripmode_enabled',               'strip_trailing_spaces_on_modify'),
             ('set_intelligent_backspace_enabled',   'intelligent_backspace'),
             ('set_highlight_current_line_enabled',  'highlight_current_line'),
             ('set_highlight_current_cell_enabled',  'highlight_current_cell'),
@@ -2562,6 +2563,8 @@ class Editor(SpyderPluginWidget):
             indentguides_o = self.get_option(indentguides_n)
             tabindent_n = 'tab_always_indent'
             tabindent_o = self.get_option(tabindent_n)
+            stripindent_n = 'strip_trailing_spaces_on_modify'
+            stripindent_o = self.get_option(stripindent_n)
             ibackspace_n = 'intelligent_backspace'
             ibackspace_o = self.get_option(ibackspace_n)
             removetrail_n = 'always_remove_trailing_spaces'
@@ -2606,6 +2609,8 @@ class Editor(SpyderPluginWidget):
                     editorstack.set_wrap_enabled(wrap_o)
                 if tabindent_n in options:
                     editorstack.set_tabmode_enabled(tabindent_o)
+                if stripindent_n in options:
+                    editorstack.set_stripmode_enabled(stripindent_o)
                 if ibackspace_n in options:
                     editorstack.set_intelligent_backspace_enabled(ibackspace_o)
                 if removetrail_n in options:
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -483,6 +483,13 @@ class CodeEditor(TextEditBaseWidget):
 
         self.oe_proxy = None
 
+        # Line stripping
+        self.last_change_position = None
+        self.last_position = None
+        self.last_auto_indent = None
+        self.skip_rstrip = False
+        self.strip_automatic_spaces_only = False
+
         # Language Server
         self.lsp_requests = {}
         self.document_opened = False
@@ -1290,8 +1297,10 @@ class CodeEditor(TextEditBaseWidget):
             # We do the following rather than using self.setPlainText
             # to benefit from QTextEdit's undo/redo feature.
             self.selectAll()
+            self.skip_rstrip = True
             self.insertPlainText(text_after)
             self.document_did_change()
+            self.skip_rstrip = False
 
     def get_current_object(self):
         """Return current object (string) """
@@ -1346,6 +1355,9 @@ class CodeEditor(TextEditBaseWidget):
         if self.occurrence_highlighting:
             self.occurrence_timer.stop()
             self.occurrence_timer.start()
+
+        # Strip if needed
+        self.strip_trailing_spaces()
 
     def __clear_occurrences(self):
         """Clear occurrence markers"""
@@ -1452,6 +1464,7 @@ class CodeEditor(TextEditBaseWidget):
 
     def __text_has_changed(self):
         """Text has changed, eventually clear found results highlighting"""
+        self.last_change_position = self.textCursor().position()
         if self.found_results:
             self.clear_found_results()
 
@@ -1728,24 +1741,30 @@ class CodeEditor(TextEditBaseWidget):
         if len(text.splitlines()) > 1:
             eol_chars = self.get_line_separator()
             text = eol_chars.join((text + eol_chars).splitlines())
+        self.skip_rstrip = True
         TextEditBaseWidget.insertPlainText(self, text)
         self.document_did_change(text)
+        self.skip_rstrip = False
 
     @Slot()
     def undo(self):
         """Reimplement undo to decrease text version number."""
         if self.document().isUndoAvailable():
             self.text_version -= 1
+            self.skip_rstrip = True
             TextEditBaseWidget.undo(self)
             self.document_did_change('')
+            self.skip_rstrip = False
 
     @Slot()
     def redo(self):
         """Reimplement redo to increase text version number."""
         if self.document().isRedoAvailable():
             self.text_version += 1
+            self.skip_rstrip = True
             TextEditBaseWidget.redo(self)
             self.document_did_change('text')
+            self.skip_rstrip = False
 
     def get_block_data(self, block):
         """Return block data (from syntax highlighter)"""
@@ -2176,10 +2195,11 @@ class CodeEditor(TextEditBaseWidget):
     def fix_indent(self, *args, **kwargs):
         """Indent line according to the preferences"""
         if self.is_python_like():
-            return self.fix_indent_smart(*args, **kwargs)
+            ret = self.fix_indent_smart(*args, **kwargs)
         else:
-            return self.simple_indentation(*args, **kwargs)
+            ret = self.simple_indentation(*args, **kwargs)
         self.document_did_change()
+        return ret
 
     def simple_indentation(self, forward=True, **kwargs):
         """
@@ -2393,7 +2413,9 @@ class CodeEditor(TextEditBaseWidget):
             # We do the following rather than using self.setPlainText
             # to benefit from QTextEdit's undo/redo feature.
             self.selectAll()
+            self.skip_rstrip = True
             self.insertPlainText(nbformat.writes(nb))
+            self.skip_rstrip = False
         except Exception as e:
             QMessageBox.critical(self, _('Removal error'),
                            _("It was not possible to remove outputs from "
@@ -2852,10 +2874,21 @@ class CodeEditor(TextEditBaseWidget):
         next_char = to_text_string(cursor.selectedText())
         return next_char
 
-    def in_comment(self):
+    def in_comment(self, cursor=None):
         if self.highlighter:
-            current_color = self.__get_current_color()
+            current_color = self.__get_current_color(cursor)
             comment_color = self.highlighter.get_color_name('comment')
+            if current_color == comment_color:
+                return True
+            else:
+                return False
+        else:
+            return False
+
+    def in_string(self, cursor=None):
+        if self.highlighter:
+            current_color = self.__get_current_color(cursor)
+            comment_color = self.highlighter.get_color_name('string')
             if current_color == comment_color:
                 return True
             else:
@@ -3033,7 +3066,7 @@ class CodeEditor(TextEditBaseWidget):
                   self.autoinsert_colons():
                     self.textCursor().beginEditBlock()
                     self.insert_text(':' + self.get_line_separator())
-                    self.fix_indent()
+                    self.fix_and_strip_indent()
                     self.textCursor().endEditBlock()
                 elif self.is_completion_widget_visible():
                     self.select_completion_list()
@@ -3054,7 +3087,7 @@ class CodeEditor(TextEditBaseWidget):
 
                     self.textCursor().beginEditBlock()
                     TextEditBaseWidget.keyPressEvent(self, event)
-                    self.fix_indent(comment_or_string=cmt_or_str)
+                    self.fix_and_strip_indent(comment_or_string=cmt_or_str)
                     self.textCursor().endEditBlock()
         elif key == Qt.Key_Insert and not shift and not ctrl:
             self.setOverwriteMode(not self.overwriteMode())
@@ -3148,6 +3181,21 @@ class CodeEditor(TextEditBaseWidget):
             # Modifiers should be passed to the parent because they
             # could be shortcuts
             event.accept()
+
+    def fix_and_strip_indent(self, comment_or_string=False):
+        """Automatically fix indent and strip previous automatic indent."""
+        # Fix indent
+        cursor_before = self.textCursor().position()
+        # A change just occured on the last line (return was pressed)
+        if cursor_before > 0:
+            self.last_change_position = cursor_before - 1
+        self.fix_indent(comment_or_string=comment_or_string)
+        cursor_after = self.textCursor().position()
+        # Remove previous spaces and update last_auto_indent
+        nspaces_removed = self.strip_trailing_spaces()
+        self.last_auto_indent = (cursor_before - nspaces_removed,
+                                 cursor_after - nspaces_removed)
+        self.document_did_change()
 
     def run_pygments_highlighter(self):
         """Run pygments highlighter."""
@@ -3253,6 +3301,92 @@ class CodeEditor(TextEditBaseWidget):
         else:
             self._last_hover_uri = uri
             return False
+
+    def line_range(self, position):
+        """
+        Get line range from position
+        """
+        if position is None:
+            return None
+        if position >= self.document().characterCount():
+            return None
+        # Check if still on the line
+        cursor = self.textCursor()
+        cursor.setPosition(position)
+        line_range = (cursor.block().position(),
+                      cursor.block().position()
+                      + cursor.block().length() - 1)
+        return line_range
+
+    def strip_trailing_spaces(self):
+        """
+        Strip trailing spaces if needed.
+
+        Remove trailing whitespace on leaving a non-string line containing it.
+        returns the number of removed spaces
+        """
+        # Update current position
+        current_position = self.textCursor().position()
+        last_position = self.last_position
+        self.last_position = current_position
+
+        if self.skip_rstrip:
+            return 0
+
+        line_range = self.line_range(last_position)
+        if line_range is None:
+            # Doesn't apply
+            return 0
+
+        def pos_in_line(pos):
+            """Check if pos is in last line."""
+            if pos is None:
+                return False
+            return line_range[0] <= pos <= line_range[1]
+
+        if pos_in_line(current_position):
+            # Check if still on the line
+            return 0
+
+        if self.strip_automatic_spaces_only:
+            if self.last_auto_indent is None:
+                return 0
+            elif (self.last_auto_indent !=
+                  self.line_range(self.last_auto_indent[0])):
+                # line not empty
+                self.last_auto_indent = None
+                return 0
+            line_range = self.last_auto_indent
+            self.last_auto_indent = None
+        elif not pos_in_line(self.last_change_position):
+            # Should process if pressed return or made a change on the line:
+            return 0
+
+        # Check if end of line in string
+        cursor = self.textCursor()
+        cursor.setPosition(line_range[1])
+        if self.in_string(cursor=cursor):
+            return 0
+
+        cursor.setPosition(line_range[0])
+        cursor.setPosition(line_range[1],
+                           QTextCursor.KeepAnchor)
+        # remove spaces on the right
+        text = cursor.selectedText()
+        strip = text.rstrip()
+
+        if line_range[0] + len(strip) < line_range[1]:
+            # Select text to remove
+            cursor.setPosition(line_range[0] + len(strip))
+            cursor.setPosition(line_range[1],
+                               QTextCursor.KeepAnchor)
+            cursor.removeSelectedText()
+            self.document_did_change()
+            # Correct last change position
+            self.last_change_position = line_range[1]
+            return line_range[1] - (line_range[0] + len(strip))
+        return 0
+
 
     def mouseMoveEvent(self, event):
         """Underline words when pressing <CONTROL>"""
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -488,7 +488,7 @@ class CodeEditor(TextEditBaseWidget):
         self.last_position = None
         self.last_auto_indent = None
         self.skip_rstrip = False
-        self.strip_automatic_spaces_only = False
+        self.strip_trailing_spaces_on_modify = True
 
         # Language Server
         self.lsp_requests = {}
@@ -686,6 +686,7 @@ class CodeEditor(TextEditBaseWidget):
                      color_scheme=None,
                      wrap=False,
                      tab_mode=True,
+                     strip_mode=False,
                      intelligent_backspace=True,
                      highlight_current_line=True,
                      highlight_current_cell=True,
@@ -785,6 +786,8 @@ class CodeEditor(TextEditBaseWidget):
         # Class/Function dropdown will be disabled if we're not in a Python file.
         self.classfuncdropdown.setVisible(show_class_func_dropdown
                                           and self.is_python_like())
+
+        self.set_strip_mode(strip_mode)
 
     # --- Language Server Protocol methods -----------------------------------
     # ------------------------------------------------------------------------
@@ -1095,6 +1098,12 @@ class CodeEditor(TextEditBaseWidget):
         (otherwise tab indents only when cursor is at the beginning of a line)
         """
         self.tab_mode = enable
+
+    def set_strip_mode(self, enable):
+        """
+        Strip all trailing spaces if enabled, else only strip on auto-indents.
+        """
+        self.strip_trailing_spaces_on_modify = enable
 
     def toggle_intelligent_backspace(self, state):
         self.intelligent_backspace = state
@@ -3304,7 +3313,7 @@ class CodeEditor(TextEditBaseWidget):
 
     def line_range(self, position):
         """
-        Get line range from position
+        Get line range from position.
         """
         if position is None:
             return None
@@ -3323,7 +3332,7 @@ class CodeEditor(TextEditBaseWidget):
         Strip trailing spaces if needed.
 
         Remove trailing whitespace on leaving a non-string line containing it.
-        returns the number of removed spaces
+        Return the number of removed spaces.
         """
         # Update current position
         current_position = self.textCursor().position()
@@ -3348,7 +3357,7 @@ class CodeEditor(TextEditBaseWidget):
             # Check if still on the line
             return 0
 
-        if self.strip_automatic_spaces_only:
+        if not self.strip_trailing_spaces_on_modify:
             if self.last_auto_indent is None:
                 return 0
             elif (self.last_auto_indent !=
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -2897,8 +2897,8 @@ class CodeEditor(TextEditBaseWidget):
     def in_string(self, cursor=None):
         if self.highlighter:
             current_color = self.__get_current_color(cursor)
-            comment_color = self.highlighter.get_color_name('string')
-            if current_color == comment_color:
+            string_color = self.highlighter.get_color_name('string')
+            if current_color == string_color:
                 return True
             else:
                 return False
--- a/spyder/plugins/editor/widgets/editor.py
+++ b/spyder/plugins/editor/widgets/editor.py
@@ -539,6 +539,7 @@ class EditorStack(QWidget):
         self.default_font = None
         self.wrap_enabled = False
         self.tabmode_enabled = False
+        self.stripmode_enabled = False
         self.intelligent_backspace_enabled = True
         self.highlight_current_line_enabled = False
         self.highlight_current_cell_enabled = False
@@ -1099,6 +1100,13 @@ class EditorStack(QWidget):
         if self.data:
             for finfo in self.data:
                 finfo.editor.set_tab_mode(state)
+
+    def set_stripmode_enabled(self, state):
+        # CONF.get(self.CONF_SECTION, 'strip_trailing_spaces_on_modify')
+        self.stripmode_enabled = state
+        if self.data:
+            for finfo in self.data:
+                finfo.editor.set_strip_mode(state)
 
     def set_intelligent_backspace_enabled(self, state):
         # CONF.get(self.CONF_SECTION, 'intelligent_backspace')
@@ -2277,6 +2285,7 @@ class EditorStack(QWidget):
             markers=self.has_markers(), font=self.default_font,
             color_scheme=self.color_scheme,
             wrap=self.wrap_enabled, tab_mode=self.tabmode_enabled,
+            strip_mode=self.stripmode_enabled,
             intelligent_backspace=self.intelligent_backspace_enabled,
             highlight_current_line=self.highlight_current_line_enabled,
             highlight_current_cell=self.highlight_current_cell_enabled,
--- a/spyder/plugins/editor/widgets/tests/test_codeeditor.py
+++ b/spyder/plugins/editor/widgets/tests/test_codeeditor.py
@@ -157,5 +157,51 @@ def test_editor_rstrip_keypress(
     assert widget.toPlainText() == expected_text
 
 
+@pytest.mark.parametrize(
+    "input_text, expected_state", [
+        ("'string ", [True, False]),
+        ('"string ', [True, False]),
+        ("'string \\", [True, True]),
+        ('"string \\', [True, True]),
+        ("'string \\ ", [True, False]),
+        ('"string \\ ', [True, False]),
+        ("'string ' ", [False, False]),
+        ('"string " ', [False, False]),
+        ("'string \"", [True, False]),
+        ('"string \'', [True, False]),
+        ("'string \" ", [True, False]),
+        ('"string \' ', [True, False]),
+        ("'''string ", [True, True]),
+        ('"""string ', [True, True]),
+        ("'''string \\", [True, True]),
+        ('"""string \\', [True, True]),
+        ("'''string \\ ", [True, True]),
+        ('"""string \\ ', [True, True]),
+        ("'''string ''' ", [False, False]),
+        ('"""string """ ', [False, False]),
+        ("'''string \"\"\"", [True, True]),
+        ('"""string \'\'\'', [True, True]),
+        ("'''string \"\"\" ", [True, True]),
+        ('"""string \'\'\' ', [True, True]),
+    ])
+def test_in_string(editorbot, input_text, expected_state):
+    """
+    Test that in_string works correctly.
+    """
+    qtbot, widget = editorbot
+    widget.set_text(input_text + '\n  ')
+    cursor = widget.textCursor()
+
+    for blanks_enabled in [True, False]:
+        widget.set_blanks_enabled(blanks_enabled)
+
+        cursor.setPosition(len(input_text))
+        assert cursor.position() == len(input_text)
+        assert widget.in_string(cursor) == expected_state[0]
+
+        cursor.setPosition(len(input_text) + 3)
+        assert widget.in_string(cursor) == expected_state[1]
+
+
 if __name__ == '__main__':
     pytest.main(['test_codeeditor.py'])
--- a/spyder/plugins/editor/widgets/tests/test_codeeditor.py
+++ b/spyder/plugins/editor/widgets/tests/test_codeeditor.py
@@ -5,8 +5,8 @@
 #
 
 # Third party imports
-from qtpy.QtCore import Qt
-from qtpy.QtGui import QFont, QTextCursor
+from qtpy.QtCore import Qt, QEvent
+from qtpy.QtGui import QFont, QTextCursor, QMouseEvent
 from pytestqt import qtbot
 import pytest
 
@@ -44,6 +44,7 @@ def test_editor_upper_to_lower(editorbot):
     widget.transform_to_lowercase()
     new_text = widget.get_text('sof', 'eof')
     assert text != new_text
+
 
 def test_editor_lower_to_upper(editorbot):
     qtbot, widget = editorbot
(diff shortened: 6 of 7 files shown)


impact27:
Is there a case where you would not want to remove these white spaces?

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -2872,6 +2887,27 @@ def run_pygments_highlighter(self):
         if isinstance(self.highlighter, sh.PygmentsSH):
             self.highlighter.make_charlist()
 
+    def left_line(self, selection_position):
+        """If we left a line with only spaces, remove them (W293)"""
+        cursor = self.textCursor()
+        current_pos = cursor.position()
+        # Chech if still on the line
+        if min(selection_position) <= current_pos and \
+                max(selection_position) >= current_pos:
```
ccordoba12: Please enclose this condition in parenthesis instead of using `\`, like this:

```python
if (min(selection_position) <= current_pos and
        max(selection_position) >= current_pos):
```

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -56,3 +56,47 @@ def test_editor_lower_to_upper(editorbot):
     widget.transform_to_uppercase()
     new_text = widget.get_text('sof', 'eof')
     assert text != new_text
+
+def test_editor_remove_empty_line(editorbot):
```
ccordoba12: Please add two blank lines of separation between tests. That's why pep8speaks is still complaining.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -2872,6 +2887,27 @@ def run_pygments_highlighter(self):
         if isinstance(self.highlighter, sh.PygmentsSH):
             self.highlighter.make_charlist()
 
+    def left_line(self, selection_position):
+        """If we left a line with only spaces, remove them (W293)"""
```
CAM-Gerlach: ``Remove whitespace if we left a line composed only of it,``` (Imperative and period per PEP 257).

impact27:
somehow qtbot.mouseClick doesn't change the position of the cursor, and doesn't trigger codeeditor.mousePressEvent. Could you help with that?

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -2872,6 +2887,28 @@ def run_pygments_highlighter(self):
         if isinstance(self.highlighter, sh.PygmentsSH):
             self.highlighter.make_charlist()
 
+    def left_line(self, selection_position):
+        """Remove whitespace if we left a line with trailing whitespace,
+        and that we where not in a comment or string."""
```
CAM-Gerlach: Per PEP 257, the first line of a docstring should be a one-line summary, followed by a blank line and additional description if necessary; the trailing ``"""`` needs to be on its own line if the docstring is not a oneliner, and per our conventions, multi-line docstrings should have the summary line on its own line (not the same line as the opening ``"""``. Furthermore, the actual text is unclear and ungrammatical. Also, it *does* remove whitespace (as it should) in comments, though not strings.

Therefore, I suggest:

``"""Remove trailing whitespace on leaving a non-string line containing it."""``

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -56,3 +57,57 @@ def test_editor_lower_to_upper(editorbot):
...
+                           'somecode = 1\nmyvar = 2 ',
+                           Qt.Key_Left),
+                          ('"""This is a string with important spaces\n    ',
+                           '"""This is a string with important spaces\n    \n',
+                           Qt.Key_Enter)
+                          ])
+def test_editor_rstrip_keypress(
+        editorbot, input_text, expected_text, keypress):
+    """Remove whitespace if we left a line with trailing whitespace
+    by keypress,"""
```
CAM-Gerlach: This function tests that the whitespace is removed, not removes the whitespace itself; docstrings should end in a period (I made a small typo in my comment) and multi-line docstrings should have the ``"""``s on their own lines per our general convention, although this is *already* short enough to fit in one.

Therefore, I suggest:

``"""Test that whitespace is removed when leaving a line with it by keypress."""``

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -56,3 +57,57 @@ def test_editor_lower_to_upper(editorbot):
...
+                           'somecode = 1\nmyvar = 2',
+                           0),
+                          ('somecode = 1\nmyvar = 2 ',
+                           'somecode = 1\nmyvar = 2 ',
+                           23)
+                          ])
+def test_editor_rstrip_mousepress(
+        editorbot, input_text, expected_text, position):
+    """Remove whitespace if we left a line with trailing whitespace
+    by mouseclick,"""
```
CAM-Gerlach: This function tests that the whitespace is removed, not removes the whitespace itself; docstrings should end in a period (I made a small typo in my comment) and multi-line docstrings should have the ``"""``s on their own lines per our general convention, although this is *already* short enough to fit in one.

Therefore, I suggest:

``"""Test that whitespace is removed when leaving a line by mouseclick."""``

CAM-Gerlach:
> somehow qtbot.mouseClick doesn't change the position of the cursor, and doesn't trigger codeeditor.mousePressEvent

I'm not sure; I almost never use it myself in favor of keyboard navigation in tests. Could you try adding ``qtbot.wait(2000)`` so you can see exactly what its doing step by step and play around with it from there?

impact27:
Here is how the PR now works:
 - On textChanged: Save the position of the cursor. This is the last modified position.

There is several way of changing line:
- Moving with mouse or arrow keys: rstrip if the last modified position in on this line.
- Adding a `\n`: rstrip the last line
- Removing a `\n`: Still on the same line, don't do anything
- Mass adding (Paste): not covered by this PR
- Mass removing: as with backspace on a `\n`, don't do anything

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -2872,6 +2883,63 @@ def run_pygments_highlighter(self):
...
+
+        Remove trailing whitespace on leaving a non-string line containing it.
+        This could also do other processing in the future based on the button
+        pressed ('source') or settings."""
+
+        # Get current position
+        cursor = self.textCursor()
+        current_pos = cursor.position()
+
+        # Chech if still on the line
```
CAM-Gerlach: *Check

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -56,3 +57,56 @@ def test_editor_lower_to_upper(editorbot):
...
+               'somecode = 1\nmyvar = 2 ',
+               [' ', (Qt.LeftButton, 23)]),
+              ('a=1\na=2 \na=3',
+               'a=1\na=2 \na=3',
+               [(Qt.LeftButton, 6), Qt.Key_Up]),
+              ])
+def test_editor_rstrip_keypress(
+        editorbot, input_text, expected_text, keys):
+    """Test that whitespace is removed when leaving a line.
+    """
```
CAM-Gerlach: Either fit this on one line, or have both ``"""`` on their own line, with the summary line in the middle.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -2872,6 +2883,63 @@ def run_pygments_highlighter(self):
         if isinstance(self.highlighter, sh.PygmentsSH):
             self.highlighter.make_charlist()
 
+    def left_line(self, position, source_button):
+        """Process a line at position 'position' on leaving it.
+
+        Remove trailing whitespace on leaving a non-string line containing it.
+        This could also do other processing in the future based on the button
+        pressed ('source') or settings."""
+
```
CAM-Gerlach: ``"""`` must be on its own line

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -2872,6 +2883,63 @@ def run_pygments_highlighter(self):
         if isinstance(self.highlighter, sh.PygmentsSH):
             self.highlighter.make_charlist()
 
+    def left_line(self, position, source_button):
+        """Process a line at position 'position' on leaving it.
```
CAM-Gerlach: Make ``""""`` on its own line per our agreed convention.

CAM-Gerlach:
Merging/rebasing on the latest ``master`` should fix the failing test.

CAM-Gerlach:
Sorry for the delay doing a final review of this, @impact27 . I ran into serious issues this past week with my Anaconda install that ended up taking several days to resolve, and I then had some other commitments and catch-up to do after finally getting that working again.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -2751,8 +2753,8 @@ def keyPressEvent(self, event):
 
                     # Check if the line start with a comment or string
                     cursor = self.textCursor()
-                    cursor.setPosition(cursor.block().position(),
```
bcolsen: Why is this code removed? That could have lead to unexpected results in other places.

spyder/plugins/editor/widgets/tests/test_autoindent.py
```diff
@@ -38,7 +38,6 @@ def get_indent_fix(text, indent_chars=" " * 4, tab_stop_width_spaces=4,
     editor.fix_indent(forward=forward)
     return to_text_string(editor.toPlainText())
 
-
```
bcolsen: Undo this change

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -2872,6 +2883,63 @@ def run_pygments_highlighter(self):
         if isinstance(self.highlighter, sh.PygmentsSH):
             self.highlighter.make_charlist()
 
+    def left_line(self, position, source_button):
```
bcolsen: I would like to see this split into two functions. 

One called `cursor_changed _lines` that checks if the cursor has changed lines and returns true. This could be useful to other code.

The second would could be called `strip_trailing_spaces` and this would be called first and `cursor_changed _lines` would be called from there.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -2763,6 +2765,15 @@ def keyPressEvent(self, event):
                     TextEditBaseWidget.keyPressEvent(self, event)
                     self.fix_indent(comment_or_string=cmt_or_str)
                     self.textCursor().endEditBlock()
+
+                    self.left_line(position, source_button=key)
+
+        elif key in [Qt.Key_Left, Qt.Key_Up, Qt.Key_Right, Qt.Key_Down]:
```
bcolsen: This should include page up and page down as well.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -2872,6 +2883,63 @@ def run_pygments_highlighter(self):
...
+        cursor.setPosition(line_range[1])
+        if self.in_comment_or_string(cursor=cursor):
+            return
+
+        # We should process if:
+        shouldProcess = False
+        if source_button in (Qt.Key_Enter, Qt.Key_Return):
+            # Pressed enter (There is a change on that line)
+            shouldProcess = True
+        elif source_button in [Qt.Key_Left, Qt.Key_Up, Qt.Key_Right,
```
bcolsen: page up and down here too

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -2872,6 +2883,63 @@ def run_pygments_highlighter(self):
...
+
+        def pos_in_line(pos):
+            if pos is None:
+                return False
+            return pos >= line_range[0] and pos <= line_range[1]
+
+        if pos_in_line(current_pos):
+            return
+
+        # Check if end of line in comment or string
```
bcolsen: I don't remember why we arn't removing the trailing spaces in strings and comments but PEP8 still complains about them.

spyder/plugins/editor/widgets/codeeditor.py
CAM-Gerlach: Does it also work for ``Ctrl-Left/Right/Up/Down/Home/End/PageUp/Pagedown`` as well?

spyder/plugins/editor/widgets/codeeditor.py
CAM-Gerlach: It is removing the trailing spaces in comments, as it should (I don't see why they should be any different than anything but strings), so this should be corrected to indicate that correctly.

Strings are a different matter, as trailing whitespace can be meaningful in them if the string is not terminated on that line. Therefore, the safe course of action is not to remove it (which this feature currently does not, as it should) and have the user deal with it themselves (if they want all trailing whitespace removed no matter what, they can turn on "Remove trailing whitespace on save").

spyder/plugins/editor/widgets/codeeditor.py
CAM-Gerlach: ^

spyder/plugins/editor/widgets/codeeditor.py
bcolsen: > Strings are a different matter, as trailing whitespace can be meaningful in them if the string is not terminated on that line

It's a bad idea to have meaningful trailing whitespace even in a string literal but I can see that some people would want to do this. That code will however have a style error.

bcolsen: @impact27 The trailing spaces are being striped from `"""string   ` in my testing

spyder/plugins/editor/widgets/codeeditor.py
bcolsen: It doesn't it also won't work with moving the cursor with the outliner, runcell, taceback links or the upcoming bookmarks. This should really be tied into some cursor movement event instead of key based.

CAM-Gerlach: @bcolsen Good point. We already track when the textcursor changes line in order to move the "Current line" syntax highlighting background color around, and that works correctly in all those scenarios. Therefore, I imagine you could just use whatever event that uses?

spyder/plugins/editor/widgets/codeeditor.py
CAM-Gerlach: > It's a bad idea to have meaningful trailing whitespace even in a string literal

I agree, but because it can have a functional effect, its better to not silently strip it and allow the user to see the PEP 8 warning and decide what to do.

> The trailing spaces are being striped from `"""string ` in my testing

Strange; I am not able to reproduce this behavior from that ``"""string`` or several other triple and single-quoted strings with ``Enter``, clicks or arrow keys, and as far as I can recall its worked pretty consistently throughout.

bcolsen: Interesting. It seems to be consistently removed on my system 
rstrip strings

definite-purple:
This might be beyond the scope of this PR, but if the introduced functionality could electively be applied to lines in docstrings but not other triple-quoted strings, that would be a pretty useful addition.

CAM-Gerlach:
> if the introduced functionality could electively be applied to lines in docstrings but not other triple-quoted strings

@impact27 You could check if ``cmt_or_str_line_begin `` has a ``"""`` as its first three non-blank characters, and that the previous non-blank line either ends with ``):`` (therefor it must be either a class or function/method definition, is a comment (to catch module docstrings and those with a comment between the object and the docstring; not technically illegal to put a comment in the middle of an assignment statement, but its never done in practice), or is the beginning of the file (must be module docstring). I don't think that should be too hard to do, I wouldn't think.

Also, you need to merge with master to fix the tests. Thanks!

CAM-Gerlach:
Hey @impact27 , looks like this great enhancement is almost ready to go, aside from a few relatively modest changes. Do you think you'll have time to finish it up soon? Thanks!

impact27:
I will work on that this week-end.

spyder/plugins/editor/widgets/codeeditor.py
impact27: I am a bit confused. What I do is: 
1) set the position to the end of the line
2) call `self.in_comment_or_string`
3) If it is true, then stop
So it should not strip comments or strings
If it does then there probably is an error in `self.in_comment_or_string`

impact27:
> > if the introduced functionality could electively be applied to lines in docstrings but not other triple-quoted strings
> 
> @impact27 You could check if `cmt_or_str_line_begin ` has a `"""` as its first three non-blank characters, and that the previous non-blank line either ends with `):` (therefor it must be either a class or function/method definition, is a comment (to catch module docstrings and those with a comment between the object and the docstring; not technically illegal to put a comment in the middle of an assignment statement, but its never done in practice), or is the beginning of the file (must be module docstring). I don't think that should be too hard to do, I wouldn't think.
> 
> Also, you need to merge with master to fix the tests. Thanks!

I would say this is beyond the scope of this PR. To do that reliably I would need to use a proper python code parser. Just knowing where the string begins already requires me to do code parsing.

spyder/plugins/editor/widgets/codeeditor.py
CAM-Gerlach: > So it should not strip comments or strings

It most certainly does strip trailing whitespace from comments, as it well should—there's no rationale I'm aware of that would motivate not doing so. For strings, I can repro the behavior shown in @bcolsen 's screenshot—space is stripped in a string if on the same line as the opening ``"""`` (when it presumably shouldn't be), but is not stripped otherwise.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -2899,6 +2925,82 @@ def run_pygments_highlighter(self):
...
+        cursor = self.textCursor()
+        cursor.setPosition(self.last_position)
+        line_range = (cursor.block().position(),
+                      cursor.block().position()
+                      + cursor.block().length() - 1)
+        return line_range
+
+    def strip_trailing_spaces(self):
+        """
+        Strip trailing spaces if needed
```
CAM-Gerlach: Add period, per PEP 257.

CAM-Gerlach:
> To do that reliably I would need to use a proper python code parser. Just knowing where the string begins already requires me to do code parsing.

You should be able to get that already with ``cursor.block().position()``, if I understand correctly. If so, then a ~>99% reliable implementation can be done with the above simple algorithm, implemented below in quasi-pseudocode. While it won't handle every possible corner case, remember that the ultimate goal is detecting whether or not whitespace is significant, and in that regard ~99% reliable docstring detection with no or negligible false positives is a major improvement in reliability over not checking for them at all.

```python
is_docstring = False
if GET_TEXT_LINE(STRING_START).strip()[:3] == '"""':
    for line_idx in range(BLOCK_START, -1, -1):
        if line_idx == 0:
            is_docstring = True
            break
        line_text = GET_TEXT_LINE(line_idx - 1).strip()
        if not line_text or line_text[0] == '#':
            continue
        if line_text[-2:] == '):':
            is_docstring = True
        break
```

impact27:
> You should be able to get that already with `cursor.block().position()`, if I understand correctly.
`cursor.block().position()` gives you the beginning of the line, not the beginning of the string, unless I am mistaken.

impact27:
One way to make it would be to follow `spyder/utils/syntaxhighlighters.py` solution and to find a regexp that would match a string that is not a docstring. 

The problem I see now is that functions like `in_comment_or_string` are using the colour in the editor to decide what the object is. It would probably be more reliable to have a set of functions that do the identification, that both the syntax highlighter and the editor would use. These functions could then differentiate between types like strings and docstrings, and the syntax highlighter could decide to treat them both similarly. 

This is all way beyond the scope of this PR. What I am imagining is a class that would create a tree representation of the code in the editor (Is the outline explorer doing that?). You could find a node by using the cursor position. Then I can do something like:
```
node = self.node_for_position(cursor.position())
if node.type() == 'docstring' and node.parent().type() == 'method':
```
I am sure this could be useful for many other editor features.

CAM-Gerlach:
> It would probably be more reliable to have a set of functions that do the identification, that both the syntax highlighter and the editor would use.

Currently, we do this with ``pygments``'; have a look e.g. [here](https://github.com/spyder-ide/spyder/blob/master/spyder/utils/syntaxhighlighters.py#L1115). But it would be great to see that improve.

> What I am imagining is a class that would create a tree representation of the code in the editor (Is the outline explorer doing that?)

That's what it displays, in a greatly simplified form, but its ultimately just getting its data from pygments, using the above.

> This is all way beyond the scope of this PR. 

I understand now, agreed.

CAM-Gerlach:
@impact27 Did you fix the problem @bcolsen reported with trailing whitespace being incorrectly stripped if on the first line of a ``"""`` string, and adding a preferences option as I detail above? If so, then this should be ready for a final review, unless I'm forgetting something. Thanks.

impact27:
> @impact27 <https://github.com/impact27> Did you fix the problem @bcolsen <https://github.com/bcolsen> reported with trailing whitespace being incorrectly stripped if on the first line of a """ string, and adding a preferences option as I detail above? If so, then this should be ready for a final review, unless I'm forgetting something. Thanks.
> 
I am not able to reproduce that problem. I would be glad to have some help on that.

About the option, the current behaviour is to process only if we made a change on the line. Otherwise the lines can be stripped by the "Strip on save" option. Do you really want an option to leave trailing spaces on edited lines? As I don't know exactly how to add an option I would be glad to have some help on that as well.

CAM-Gerlach:
> I am not able to reproduce that problem. I would be glad to have some help on that.

Sure, I had trouble repro'ing it until I replicated *exactly* what @bcolsen did.

> Space is stripped in a string if on the same line as the opening ``"""`` (when it presumably shouldn't be), but is not stripped otherwise.

rstrip strings

> Do you really want an option to leave trailing spaces on edited lines?

It would do so by default, but yes, because otherwise our editor is modifying the user's files without their consent and with no way to disable it. There could conceivably be circumstances where there is undesired, and if this features is implemented properly, there is no way to work around it. Furthermore, trailing white space can sometimes be meaningful in HTML, and if there is some edge case bug where it is being stripped from strings (as there currently is) this could theorectically forcibly change the meaning of a problem. It is just a general principle that forcible automated editing features must always have some way to be disabled.

In this case, disabling it should still remove just the spaces added by the automatic indent feature (as discussed previously), since *Spyder* adds those spaces, not the user.

> As I don't know exactly how to add an option I would be glad to have some help on that as well.

Sure, Its pretty simple; see, e.g. [this commit](https://github.com/spyder-ide/spyder/pull/9191/commits/3f578278c3d953d5daef9cf8c1a920f2f64cbb6c) for an example. You'd want to add it to the editor plugin and preference pane rather than the LSP one, of course, under the ``Source Code`` tab.

impact27:
I added the option. I can not reproduce the bug
cap

CAM-Gerlach:
FYI @impact27 , you can avoid having to do all these ``pep8`` commits (which really should be amended/fixup'ed into your previous commits as they will have to be eventually to clean up your commit history) if you turn on ``Editor > Introspection > Real-time code style analysis``.

In general, any commit that merely fixes something relatively small wrong with a previous commit, addresses feedback (other than adding something completely new), corrects text or style, makes tests pass, etc. should be combined with the previous commit, so that they stand on their own as an independently valid point in Spyder's history that does one self-contained thing and will not suddenly break anything other than what its adding if attempting to debug an issue. Of course, you can always do this right before a PR is merged with ``git rebase -i`` and ``fixup``, but its much easier to just amend your commits as you work. I have a ``git fixup`` command myself that aliases ``git commit --amend --no-edit`` to quickly allow me to amend previous commits with minimal changes.

Also, make sure your commit messages are understandable on their own (i.e. without the reader needing to have external knowledge of what feature they apply to), since folks browsing through Spyder's commit history (as I've often had to for various reasons) won't see them as part of a PR and thus will have to work and/or guess to figure out what feature they apply to. Fewer, atomic, focused commits should also make this easier.

ccordoba12:
@CAM-Gerlach, is this ready?

CAM-Gerlach:
@ccordoba12 This needs a last review to make sure it has fixed the bug @bcolsen reported and I repro'd, implemented the preferences setting correctly, and there are no other lingering issues. I'll do that now.

@impact27 , in the future, please leave a comment or request a re-review when a PR is ready to be reviewed again so we know when you want us to look it over. Thanks!

impact27:
> @impact27 , in the future, please leave a comment or request a re-review when a PR is ready to be reviewed again so we know when you want us to look it over. Thanks!

I will. I rebased it on the current master.

CAM-Gerlach:
> I will. I rebased it on the current master.

To fix the tests, or why?

CAM-Gerlach:
@impact27 Okay, so is it ready for review now, or are you still working on it?

impact27:
It is ready for review. I am just not sure about the bug @bcolsen reported. I am unable to reproduce it. It would be great if you could add a test for this it is still there, so I can fix it.

impact27:
Can you rerun the failing test?

CAM-Gerlach:
@impact27 Looks like everything's passing now. I'll re-review it momentarily; its number 1 in my review queue. I'll see if I can repro @bcolsen 's issue again,

bcolsen:
@impact27 I tried the latest code here on `safe-mode` and I can only reproduce the bug if I enable the `show blank spaces` option in the sources menu. There is something very strange happening there.

blank_space

CAM-Gerlach:
Its likely due to the how the ``is_comment_or_string`` function works, based on the syntax highlighting color, not the underlying code. The spaces do not have the string color, so ``is_comment_or_string`` returns False and the blanks are stripped.

CAM-Gerlach:
@impact27 To avoid this, if ``Show blank spaces`` is enabled, you could check that the last non-whitespace character of the previous line, including the shown spaces, is either not colored as a string (with the same function), or is ``"`` or ``'`` (implying further whitespace is outside the string), before stripping spaces.

impact27:
I have added a test to test if `'"""string   '` is replaced by `'"""string   \n'`. Do you mean you only have the error when `show blank spaces` is enabled?

impact27:
I am not able to turn `show blank spaces` on with the current master somehow.

impact27:
Fixed.

ccordoba12:
@bcolsen, please confirm and approve if ready.

impact27:
There was a similar problem for single quotes strings that I fixed as well

CAM-Gerlach:
@impact27 Reviewing it now.

spyder/plugins/editor/confpage.py
```diff
@@ -122,6 +122,12 @@ def setup_page(self):
                   "of a line (when this option is enabled, code\n"
                   "completion may be triggered using the alternate\n"
                   "shortcut: Ctrl+Space)"))
+        strip_mode_box = newcb(
+            _("Strip only automatic indent"),
```
CAM-Gerlach:
```suggestion
            _("Automatically strip trailing spaces on changed lines"),
```

spyder/config/main.py
```diff
@@ -227,6 +227,7 @@
               'autosave_enabled': True,
               'autosave_interval': 60,
               'docstring_type': 'Numpydoc',
+              'strip_automatic_spaces_only': False,
```
CAM-Gerlach:
```suggestion
              'strip_trailing_spaces_on_modify': True,
```

spyder/plugins/editor/confpage.py
```diff
@@ -122,6 +122,12 @@ def setup_page(self):
...
+            'strip_automatic_spaces_only', default=False,
```
CAM-Gerlach:
```suggestion
            'strip_trailing_spaces_on_modify', default=True,
```

spyder/plugins/editor/confpage.py
```diff
@@ -122,6 +122,12 @@ def setup_page(self):
...
+            tip=_("If enabled, pressing Return will not strip,\n"
```
CAM-Gerlach:
```suggestion
            tip=_("If enabled, modified lines of code (excluding strings)\n"
```

spyder/plugins/editor/confpage.py
```diff
@@ -122,6 +122,12 @@ def setup_page(self):
...
+                  "the previous line, unless it was automatically\n"
```
CAM-Gerlach:
```suggestion
                  "will have their trailing whitespace stripped when leaving them.\n"
```

spyder/plugins/editor/confpage.py
```diff
@@ -122,6 +122,12 @@ def setup_page(self):
...
+                  " added by spyder"))
```
CAM-Gerlach:
```suggestion
                  "If disabled, only whitespace added by Spyder will be stripped."))
```

spyder/plugins/editor/plugin.py
```diff
@@ -1205,6 +1205,7 @@ def register_editorstack(self, editorstack):
             ('set_tab_stop_width_spaces',           'tab_stop_width_spaces'),
             ('set_wrap_enabled',                    'wrap'),
             ('set_tabmode_enabled',                 'tab_always_indent'),
+            ('set_stripmode_enabled',           'strip_automatic_spaces_only'),
```
CAM-Gerlach:
```suggestion
            ('set_stripmode_enabled',               'strip_trailing_spaces_on_modify'),
```

spyder/plugins/editor/plugin.py
```diff
@@ -2562,6 +2563,8 @@ def apply_plugin_settings(self, options):
             indentguides_o = self.get_option(indentguides_n)
             tabindent_n = 'tab_always_indent'
             tabindent_o = self.get_option(tabindent_n)
+            stripindent_n = 'strip_automatic_spaces_only'
```
CAM-Gerlach:
```suggestion
            stripindent_n = 'strip_trailing_spaces_on_modify'
```

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -483,6 +483,13 @@ def __init__(self, parent=None):
 
         self.oe_proxy = None
 
+        # Line stripping
+        self.last_change_position = None
+        self.last_position = None
+        self.last_auto_indent = None
+        self.skip_rstrip = False
+        self.strip_automatic_spaces_only = False
```
CAM-Gerlach:
```suggestion
        self.strip_trailing_spaces_on_modify = False
```

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1089,6 +1099,13 @@ def set_tab_mode(self, enable):
         """
         self.tab_mode = enable
 
+    def set_strip_mode(self, enable):
+        """
+        enabled = strip automatic spaces only
```
CAM-Gerlach:
```suggestion
        Strip all trailing spaces if enabled, else only strip on auto-indents.
```

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1089,6 +1099,13 @@ def set_tab_mode(self, enable):
         """
         self.tab_mode = enable
 
+    def set_strip_mode(self, enable):
+        """
+        enabled = strip automatic spaces only
+        (otherwise rstrip)
```
CAM-Gerlach:
```suggestion
```

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1089,6 +1099,13 @@ def set_tab_mode(self, enable):
         """
         self.tab_mode = enable
 
+    def set_strip_mode(self, enable):
+        """
+        enabled = strip automatic spaces only
+        (otherwise rstrip)
+        """
+        self.strip_automatic_spaces_only = enable
```
CAM-Gerlach:
```suggestion
        self.strip_trailing_spaces_on_modify = enable
```

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3254,6 +3309,92 @@ def _handle_goto_uri_event(self, pos):
             self._last_hover_uri = uri
             return False
 
+    def line_range(self, position):
+        """
+        Get line range from position
```
CAM-Gerlach:
```suggestion
        Get line range from position.
```

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3254,6 +3309,92 @@ def _handle_goto_uri_event(self, pos):
...
+                      cursor.block().position()
+                      + cursor.block().length() - 1)
+        return line_range
+
+    def strip_trailing_spaces(self):
+        """
+        Strip trailing spaces if needed.
+
+        Remove trailing whitespace on leaving a non-string line containing it.
+        returns the number of removed spaces
```
CAM-Gerlach:
```suggestion
        Return the number of removed spaces.
```

spyder/plugins/editor/widgets/editor.py
```diff
@@ -1100,6 +1101,13 @@ def set_tabmode_enabled(self, state):
             for finfo in self.data:
                 finfo.editor.set_tab_mode(state)
 
+    def set_stripmode_enabled(self, state):
+        # CONF.get(self.CONF_SECTION, 'strip_automatic_spaces_only')
```
CAM-Gerlach: This should probably be either uncommented or removed?

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -83,3 +84,78 @@ def test_editor_log_lsp_handle_errors(editorbot, capsys):
     test_1 = "Error when processing signature" in captured.err
     test_2 = "codec can't decode byte 0x81" in captured.err
     assert test_1 or test_2
+
+
+@pytest.mark.parametrize(
+    "input_text, expected_text, keys, auto_only",
```
CAM-Gerlach:
```suggestion
    "input_text, expected_text, keys, strip_all",
```

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -83,3 +84,78 @@ def test_editor_log_lsp_handle_errors(editorbot, capsys):
...
+    [
+        ("for i in range(2): ",
+         "for i in range(2): \n\n     \n    ",
+         [Qt.Key_Enter, Qt.Key_Enter, ' ', Qt.Key_Enter],
+         True),
```
CAM-Gerlach:
```suggestion
         False),
```

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -83,3 +84,78 @@ def test_editor_log_lsp_handle_errors(editorbot, capsys):
...
+        ('for i in range(2): ',
+         'for i in range(2):\n\n    ',
+         [Qt.Key_Enter, Qt.Key_Enter],
+         False),
```
CAM-Gerlach:
```suggestion
         True),
```

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -83,3 +84,78 @@ def test_editor_log_lsp_handle_errors(editorbot, capsys):
...
+        ('myvar = 2 ',
+         'myvar = 2\n',
+         [Qt.Key_Enter],
+         False),
```
CAM-Gerlach:
```suggestion
         True),
```

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -83,3 +84,78 @@ def test_editor_log_lsp_handle_errors(editorbot, capsys):
...
+        ('somecode = 1\nmyvar = 2 \nmyvar = 3',
+         'somecode = 1\nmyvar = 2 \nmyvar = 3',
+         [' ', Qt.Key_Up, Qt.Key_Up],
+         False),
```
CAM-Gerlach:
```suggestion
         True),
```

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -83,3 +84,78 @@ def test_editor_log_lsp_handle_errors(editorbot, capsys):
...
+        ('somecode = 1\nmyvar = 2 ',
+         'somecode = 1\nmyvar = 2 ',
+         [Qt.Key_Left],
+         False),
```
CAM-Gerlach:
```suggestion
         True),
```

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -83,3 +84,78 @@ def test_editor_log_lsp_handle_errors(editorbot, capsys):
...
+        ('"""This is a string with important spaces\n    ',
+         '"""This is a string with important spaces\n    \n',
+         [Qt.Key_Enter],
+         False),
```
CAM-Gerlach:
```suggestion
         True),
```

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -83,3 +84,78 @@ def test_editor_log_lsp_handle_errors(editorbot, capsys):
...
+        ('"""string   ',
+         '"""string   \n',
+         [Qt.Key_Enter],
+         False),
```
CAM-Gerlach:
```suggestion
         True),
```

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -83,3 +84,78 @@ def test_editor_log_lsp_handle_errors(editorbot, capsys):
...
+        ('somecode = 1\nmyvar = 2',
+         'somecode = 1\nmyvar = 2',
+         [' ', (Qt.LeftButton, 0)],
+         False),
```
CAM-Gerlach:
```suggestion
         True),
```

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -83,3 +84,78 @@ def test_editor_log_lsp_handle_errors(editorbot, capsys):
...
+        ('somecode = 1\nmyvar = 2',
+         'somecode = 1\nmyvar = 2 ',
+         [' ', (Qt.LeftButton, 23)],
+         False),
```
CAM-Gerlach:
```suggestion
         True),
```

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -83,3 +84,78 @@ def test_editor_log_lsp_handle_errors(editorbot, capsys):
...
+        ('a=1\na=2 \na=3',
+         'a=1\na=2 \na=3',
+         [(Qt.LeftButton, 6), Qt.Key_Up],
+         False),
```
CAM-Gerlach:
```suggestion
         True),
```

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -83,3 +84,78 @@ def test_editor_log_lsp_handle_errors(editorbot, capsys):
...
+    ])
+def test_editor_rstrip_keypress(
+        editorbot, input_text, expected_text, keys, auto_only):
+    """
+    Test that whitespace is removed when leaving a line.
+    """
+    qtbot, widget = editorbot
+    widget.strip_automatic_spaces_only = auto_only
```
CAM-Gerlach:
```suggestion
    widget.strip_trailing_spaces_on_modify = strip_all
```

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -83,3 +84,78 @@ def test_editor_log_lsp_handle_errors(editorbot, capsys):
...
+         'somecode = 1\nmyvar = 2 ',
+         [' ', (Qt.LeftButton, 23)],
+         False),
+        ('a=1\na=2 \na=3',
+         'a=1\na=2 \na=3',
+         [(Qt.LeftButton, 6), Qt.Key_Up],
+         False),
+    ])
+def test_editor_rstrip_keypress(
+        editorbot, input_text, expected_text, keys, auto_only):
```
CAM-Gerlach:
```suggestion
        editorbot, input_text, expected_text, keys, strip_all):
```

impact27:
> Other than that, is there a reason several places it claims to test for "in comment or string" before performing the strip, when in reality its only the latter (as it should be)? It could be confusing to future readers trying to parse the code.

Mostly because I forgot to update the comments :p

CAM-Gerlach:
@impact27 Nooooooooooooooo batch the commits, otherwise you'll end up with 100 tiny commits!!!

impact27:
I don’t know how to do that. I wanted to then batch the commits on my computer by rebase. I didn’t find the “accept all” button.

impact27:
I’ll finish that next week

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3254,6 +3308,92 @@ def _handle_goto_uri_event(self, pos):
...
+                      cursor.block().position()
+                      + cursor.block().length() - 1)
+        return line_range
+
+    def strip_trailing_spaces(self):
+        """
+        Strip trailing spaces if needed.
+
+        Remove trailing whitespace on leaving a non-string line containing it.
+        Return the number of removed spaces.
```
CAM-Gerlach:
```suggestion
        Return the number of removed spaces as an integer.
```

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3254,6 +3308,92 @@ def _handle_goto_uri_event(self, pos):
...
+            """Check if pos is in last line."""
+            if pos is None:
+                return False
+            return line_range[0] <= pos <= line_range[1]
+
+        if pos_in_line(current_position):
+            # Check if still on the line
+            return 0
+
+        if self.strip_automatic_spaces_only:
```
CAM-Gerlach: I missed a spot somehow, that's why your tests are failing.

```suggestion
        if not self.strip_trailing_spaces_on_modify:
```

CAM-Gerlach:
> I don’t know how to do that. 

@impact27 Simple, just click the ``Add to batch`` button instead of the commit button on the files tab, then commit everything in one go (just like doing a review).

image

> I’ll finish that next week

No pressure, but the only thing you need to do now is press apply on more suggestion, move line 185 in ``confpage.py`` down by one, check one comment, and squash the commits, which shouldn't take more than a few minutes, and then we can finally merge this long-running PR.

bcolsen:
I tried to test and I get the traceback below whenever I make a new line in a string.

```python traceback
  File "/home/bcolsen/spyder-dev/spyder/plugins/editor/widgets/codeeditor.py", line 1369, in __cursor_position_changed
    self.strip_trailing_spaces()
  File "/home/bcolsen/spyder-dev/spyder/plugins/editor/widgets/codeeditor.py", line 3357, in strip_trailing_spaces
    if self.strip_automatic_spaces_only:
AttributeError: 'CodeEditor' object has no attribute 'strip_automatic_spaces_only'
```

CAM-Gerlach:
@impact27 As I mentioned in my comment above, that's because I missed one spot. Commit my aforementioned suggestion to fix it.

spyder/plugins/editor/widgets/editor.py
impact27: well every other function has that.

CAM-Gerlach: @ccordoba12 ?

impact27:
everything is squashed now

impact27:
Wait, my single string fix broke the highlighting slightly. Let me have a look.

impact27:
string

CAM-Gerlach:
@impact27 Sure, @bcolsen 's request changes review is still blocking merge and I was going to wait for @ccordoba12 to look at that one comment anyway.

impact27:
I added a difference between a continuing single quote string `"string \` and a non-continuing `"string `

bcolsen:
Look good to me. Nice work @impact27!

impact27:
I will just add a few tests for in_string

impact27:
Done

spyder/plugins/editor/widgets/editor.py
ccordoba12: Please leave it for now. That's old stuff that Pierre left and it should be fixed in another PR.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -2179,7 +2207,6 @@ def fix_indent(self, *args, **kwargs):
             return self.fix_indent_smart(*args, **kwargs)
         else:
             return self.simple_indentation(*args, **kwargs)
-        self.document_did_change()
```
ccordoba12: Why did you remove this? This prevents linting to be updated when `fix_indent` is called.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3149,6 +3189,20 @@ def keyPressEvent(self, event):
...
+        cursor_before = self.textCursor().position()
+        # A change just occured on the last line (return was pressed)
+        if cursor_before > 0:
+            self.last_change_position = cursor_before - 1
+        self.fix_indent(comment_or_string=comment_or_string)
+        cursor_after = self.textCursor().position()
+        # Remove previous spaces and update last_auto_indent
+        nspaces_removed = self.strip_trailing_spaces()
+        self.last_auto_indent = (cursor_before - nspaces_removed,
+                                 cursor_after - nspaces_removed)
```
ccordoba12: This should also call `self.document_did_change()` at this point for linting to be updated.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3254,6 +3308,92 @@ def _handle_goto_uri_event(self, pos):
...
+            # Select text to remove
+            cursor.setPosition(line_range[0] + len(strip))
+            cursor.setPosition(line_range[1],
+                               QTextCursor.KeepAnchor)
+            cursor.removeSelectedText()
+            self.document_did_change()
+            # Correct last change position
+            self.last_change_position = line_range[1]
+            return line_range[1] - (line_range[0] + len(strip))
+        return 0
```
ccordoba12: I think this also needs `self.document_did_change()` before this return because it modifies the text programmatically.

spyder/plugins/editor/widgets/codeeditor.py
impact27: Well it is after the return of the function, so it won't get executed either way. Should the function be changed?

spyder/plugins/editor/widgets/codeeditor.py
impact27: Does it? only the first return changes the text.

spyder/plugins/editor/widgets/codeeditor.py
ccordoba12: Right, I didn't take into account. Then I think it should be added to `fix_indent_smart` and `simple_indentation`.

impact27: I changed it so document_did_change is called

spyder/plugins/editor/widgets/codeeditor.py
impact27: done

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3254,6 +3311,92 @@ def _handle_goto_uri_event(self, pos):
...
+        cursor.setPosition(line_range[1])
+        if self.in_string(cursor=cursor):
+            return 0
+
+        cursor.setPosition(line_range[0])
+        cursor.setPosition(line_range[1],
+                           QTextCursor.KeepAnchor)
+        # remove spaces on the right
+        text = cursor.selectedText()
+        strip = text.rstrip()
```
ccordoba12: This also modifies the text, so could add a `document_did_change` before this line? Thanks!

spyder/plugins/editor/widgets/codeeditor.py
ccordoba12: You're right. This is not the right place.

spyder/plugins/editor/widgets/codeeditor.py
impact27: Well it modifies the local variable `text`, but not the document.

impact27: The only line that does anything to the editor is `cursor.removeSelectedText()`

impact27: Otherwise I am just changing local variables, am I not?

ccordoba12: Yes, you're right here too. Sorry for the hassle.

Issue #8519, opened 2019-01-04
Change IPython console icon and connect consoles to the File switcher


ccordoba12:
I'd like to use a different icon for Spyder 4 because the current icon (**`IP:`**) is barely visible in the dark theme. My proposal is to use `mdi-console`

![seleccion_001](https://user-images.githubusercontent.com/365293/50696664-0b7f7600-100e-11e9-9bf4-569f75f07187.png)

because it's easier to recognize and also because a similar icon is used in JupyterLab already.

Besides, since people have now the possibility to create different kinds of consoles (Sympy, Cython, etc) and those are named, I think it's really important that we connect them to the file switcher.

PR #8768, opened 2019-02-14
PR: Change IPython Console icon

juanis2112:
## Description of Changes

I changed IPython Console icon.

screen shot 2019-02-14 at 12 30 04 am

### Affirmation

Juanis2112


spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -388,6 +388,15 @@ def get_plugin_actions(self):
             return client.get_options_menu()
         return self.menu_actions
 
+    def get_current_tab_manager(self):
+        """Return the plugin as tab_manager"""
```
CAM-Gerlach: Missing period.

spyder/plugins/ipythonconsole/widgets/client.py
```diff
@@ -135,6 +135,8 @@ def __init__(self, plugin, id_,
         self.allow_rename = True
         self.stderr_dir = None
         self.is_error_shown = False
+        self.filename = self.given_name
```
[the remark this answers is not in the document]
CAM-Gerlach: More broadly, we should rename the ``File Switcher`` to just the ``Switcher``, or to something else more representative of its expanded functionality (``Focus Switcher``? ``Quick Nav``? What do other IDEs call this?)

ccordoba12:
@juanis2112, please remove the last two commits in this PR and merge it with master, so we can have it for beta2.

We really need to create a proper API for the fileswitcher to show the open IPython consoles there, but @goanpeca is going to take care of that because it's not an easy task.

spyder/app/mainwindow.py
```diff
@@ -3060,9 +3060,11 @@ def add_to_fileswitcher(self, plugin, tabs, data, icon):
             self.fileswitcher = FileSwitcher(self, plugin, tabs, data, icon)
         else:
             self.fileswitcher.add_plugin(plugin, tabs, data, icon)
-
-        self.fileswitcher.sig_goto_file.connect(
+        try:
+            self.fileswitcher.sig_goto_file.connect(
                 plugin.get_current_tab_manager().set_stack_index)
+        except:
+            pass
```
ccordoba12: @juanis2112, please revert this change.

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -405,6 +405,8 @@ def register_plugin(self):
                                      self.set_current_client_working_directory)
 
         self.tabwidget.currentChanged.connect(self.update_working_directory)
+        self.main.add_to_fileswitcher(self, self.tabwidget, self.clients,
+                                      ima.icon('ipython_console'))
```
ccordoba12: And also revert this change.

spyder/app/mainwindow.py
CAM-Gerlach: Also, as a general note, you should essentially *never* use a bare except (equivalent to ``except BaseException:``, vs, at the very least, ``except Exception:``), as the former incorrectly catch ``SystemExit``, ``KeyboardInterrupt`` and ``GeneratorExit`` and prevents Spyder from being correctly closed by these if happens to be in the block in question (or gets stuck in it, etc.). 

Furthermore, if you do ``except Exception:``, you should think very carefully about what purpose it serves and include a comment justifying it over something more specific and tailored, as it will also hide errors from our tests and error reporting system, silently retaining potentially fixable problems.

spyder/utils/icon_manager.py
```diff
@@ -162,8 +162,8 @@
     'error':                   [('fa.times-circle',), {'color': 'darkred'}],
     'warning':                 [('fa.warning',), {'color': 'orange'}],
     'todo':                    [('fa.exclamation',), {'color': '#3775a9'}],
-    'ipython_console':         [('spyder.ipython-logo-alt',), {'color': MAIN_FG_COLOR}],
-    'ipython_console_t':       [('spyder.ipython-logo-alt',), {'color':'gray'}],
+    'ipython_console':         [('mdi.console',), {'color': MAIN_FG_COLOR}],
+    'ipython_console_t':       [('mdi.console',), {'color':'gray'}],
```
CAM-Gerlach: Fix PEP 8 warning

```suggestion
    'ipython_console_t':       [('mdi.console',), {'color': 'gray'}],
```

ccordoba12:
@CAM-Gerlach, please see the new title of this PR. Now it's very simple, it just wants to change the IPython console icon, but @juanis2112 left some things that I asked her to remove.

Instead of wasting your time and ours testing this, please help me to decide if PR #8734 is ready or not.

CAM-Gerlach:
@ccordoba12 New title or not, it still says it fixes #8519 which includes connecting consoles to the file switcher, and your only comment that actually said it was being descoped was almost two months old and gave the reason as making it in for Beta 2, which has already been released, so it wasn't 100% clear.

Therefore, given the unclear situation, severity of the problem and the fact that the CIs weren't detecting it, it was not an unreasonable decision to post a short comment pointing it out, as if there was even a small chance it related to desired functionality in this PR, the potential consequences outweighed the mere few seconds it "wasted" of your time.

> Instead of wasting your time and ours testing this, please help me to decide if PR #8734 is ready or not.

That's *exactly* what I *was* doing, until you chose to waste your time (far more than the few seconds you spent reading my comment) writing up a whole reply criticizing me for it, and now requiring me to waste mine to respond to your needless accusations.

ccordoba12:
I modified the PR description. Look, when I'll require your help, I'll ask for it. Otherwise, please don't make me argue about small details.

CAM-Gerlach:
I will follow this up in a more appropriate venue, namely private Gitter chat.

CAM-Gerlach:
For what its worth, I agree with your requested icon change, though.

spyder/app/mainwindow.py
ccordoba12: @juanis2112, what I meant by "revert" is that you need to leave things as they were before. This means removing the try/except you added here, so this change doesn't appear as part of this PR.

spyder/plugins/ipythonconsole/plugin.py
ccordoba12: And please remove these lines because they are not needed anymore.

spyder/utils/icon_manager.py
ccordoba12: To address this, please press the button `Commit suggestion` above and after that `Commit changes`.

Issue #6984, opened 2018-04-18
Automatic error report


ashourm:
<!---   *** BEFORE SUBMITTING: PASTE CLIPBOARD HERE TO COMPLETE YOUR REPORT ***   ---!>
Sometimes when Spider it is unable to connect to kernel & you have to restart it to connect.

[spydernotconnectingtokernel](https://user-images.githubusercontent.com/35844281/38942996-d31901a2-431f-11e8-8767-b3b85800bf37.png)

CAM-Gerlach:
Thanks for reporting. Your image didn't load

> you have to restart it to connect. 

The kernel or Spyder? Assuming the former, I used to occasionally see this a while ago with older Spyder, IPython etc. version but haven't noticed it since.

Please try the steps listed in the [Spyder Troubleshooting Guide](https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ), specifically the [Basic First Aid](https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ#basic-first-aid-general-troubleshooting) and [Emergency CPR](https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ#emergency-cpr-spyder-wont-launch) sections, which resolve the great majority of Spyder installation issues.

If it still isn't working, please describe in detail the circumstances in which the error occurred, and post the information listed in [relevant subsection](https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ#must-have-and-helpful-items-to-include) of the Calling for Help section as a comment to this issue, and we'll try to help if we can. Otherwise, your issue will likely have to be closed after one week (7 days). Thanks, and best of luck!

ashourm:
Thank you for your quick response and directions.

CAM-Gerlach:
Closed due to lack of response. If you can provide more information, please re-open and do so. Thanks.

Issue #6894, opened 2018-04-04
pyside2 support


looooo:
I am trying to get spyder working with pyside2. I am using qt5.6 and pyside2 from conda-forge channel.

So far there are some problems in the initialization (qtpy) and some not available members of QtCore:
- QtCore.qround  which I replaced with pythons round
```
  File "~/conda/envs/pyside2/lib/python3.6/site-packages/qtawesome/iconic_font.py", line 24, in <module>
    from qtpy.QtCore import QObject, QPoint, QRect, qRound, Qt
ImportError: cannot import name 'qRound'

```
- QWebEngineSettings (which I simple disabled)
```
  File "~/projects/spyder/spyder/widgets/browser.py", line 18, in <module>
    from qtpy.QtWebEngineWidgets import (QWebEnginePage, QWebEngineSettings,
ImportError: cannot import name 'QWebEngineSettings'
Fatal Python error: deallocating None

```
- QRegularExpression (should be the same as QRegExpr)
```
  File "~/projects/spyder/spyder/widgets/mixins.py", line 40, in <module>
    from qtpy.QtCore import QRegularExpression
ImportError: cannot import name 'QRegularExpression'
Fatal Python error: deallocating None
```
- QAbstractItemModel.index (no idea)
```
Traceback (most recent call last):
  File "~/projects/spyder/spyder/widgets/explorer.py", line 969, in filterAcceptsRow
    index = self.sourceModel().index(row, 0, parent_index)
NotImplementedError: pure virtual method 'QAbstractItemModel.index()' not implemented.
```
depends on https://bugreports.qt.io/browse/PYSIDE-570

are there any plans to support pyside2. It seems like it is almost possible.

ccordoba12:
> are there any plans to support pyside2. It seems like it is almost possible.

No. It'd be a terrible waste of time for us to support PySide2 when it's still very buggy and it doesn't offer any gain to the project.

Closing this one for the time being.

stonebig:
@looooo You may check at the bottom of this page the advance of PySide2 towards PyQt5 parity https://wiki.qt.io/Qt_for_Python_Missing_Bindings

Today they are 75 points behind PyQt5, at PySide-5.11a1 (alpha1)

You may check their progress on this metric after their preview release, end of may 2018.

stonebig:
the patch for Qround is there https://github.com/spyder-ide/qtawesome/pull/81/files.  will re-propose in a more compatible way

stonebig:
[...]

PR #8774, opened 2019-02-15
avoid calling aliased object's methods twice

pijyoi:
On non-darwin platforms, _action_no_icon simply points to self.
The original code thus ends up calling methods twice on the same object.

This change removes the assumption that calling methods twice has no
side-effects.

## Description of Changes

### Issue(s) Resolved

Fixes #

### Affirmation

### DIFF
--- a/spyder/utils/qthelpers.py
+++ b/spyder/utils/qthelpers.py
@@ -450,6 +450,9 @@ class SpyderAction(QAction):
         """Intercept method calls and apply to both actions, except signals."""
         attr = super(SpyderAction, self).__getattribute__(name)
 
+        if super(SpyderAction, self).__getattribute__('_action_no_icon') is self:
+            return attr
+
         if hasattr(attr, '__call__') and name not in ['triggered', 'toggled',
                                                       'changed', 'hovered']:
             def newfunc(*args, **kwargs):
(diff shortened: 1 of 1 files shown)


ccordoba12:
@pijyoi, thanks for your contribution. What does this fix exactly? It's unclear from your description.

ccordoba12:
I mean, what exact behavior is wrong right now on Linux or Windows that this is going to fix?

pijyoi:
When using PySide2 as bindings (despite the wontfix status of https://github.com/spyder-ide/spyder/issues/6894,) menu actions get triggered twice on Windows and Linux. 
e.g. Clicking Help->About shows the About menu twice.

To my understanding, this happens on PySide2 because a hidden "connect" method gets called when connecting signals. This "connect" method does not get filtered by the list ['triggered', 'toggled', 'changed', 'hovered'] and is thus called twice on the same QAction object.

For PyQt5, this fix has no visible effect. But it does at least reduce unnecessary method calls to Qt.

ccordoba12:
> When using PySide2 as bindings (despite the wontfix status of #6894,)

Mmmh, I don't know what to think about that. I mean, you're trying to fix something that it's not currently broken.

ccordoba12:
If you want PySide2 support, wouldn't it be better for you to commit the necessary time to make our test suite pass and fix all other broken things with PySide2?

#6984 is not really a *won't fix*, but a call for action for the community.

PR #8798, opened 2019-02-18
[WIP] PR: Provide an API for external plugins to load/save/create new files

### LINKED ISSUE #7794: Allow plugins to hook into File > Open
This is inspired by spyder-ide/spyder-notebook#134. Users of the spyder-notebook plugin expect to open a notebook using `File` > `Open`. At the moment, this causes the notebook to be opened in the editor, which is not so useful. Somehow the notebook plugin (or any other plugin) should be able to indicate that it should be used to open files of a particular type.

Maybe the plugin initialization code should include a call like:
```
register_file_extension('.ipynb')
```
When the user opens a file with that extension, Spyder calls the plugin's `load()` function, similar to `Editor.load()`.

ccordoba12:
This PR provides a public API for external plugins to register their own functionality when the following buttons/actions are pressed in Spyder toolbar and its File menu:

- [ ] New file
- [x] Load file
- [x] Save file
- [ ] Save all
- [ ] Save as
- [ ] Recent files

Fixes #7794.

Temporal TODOs (just so I don't forget to address them):

- [ ] Save button is disabled when a file in the Editor is saved. It should be enabled/disabled in a per plugin basis.
- [ ] Rename PluginWidget to BasePluginWidget (update changelog about that).
- [ ] Add comments explaining what the signals in BasePluginWidget do.

### DIFF
--- a/spyder/api/plugins.py
+++ b/spyder/api/plugins.py
@@ -235,7 +235,6 @@ class SpyderPluginWidget(PluginWidget):
     """
 
     # ---------------------------- ATTRIBUTES ---------------------------------
-
     # Name of the configuration section that's going to be
     # used to record the plugin's permanent data in Spyder
     # config system (i.e. in spyder.ini)
@@ -275,7 +274,6 @@ class SpyderPluginWidget(PluginWidget):
     file_extensions = []
 
     # ------------------------------ METHODS ----------------------------------
-
     def get_plugin_title(self):
         """
         Return plugin title.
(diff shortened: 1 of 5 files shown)


spyder/api/plugins.py
```diff
@@ -268,6 +268,12 @@ class SpyderPluginWidget(PluginWidget):
     # Status: Optional
     shortcut = None
 
+    # Registered file extesions that the plugin can open
+    # inside it.
+    # Example: ['.ipynb'] for spyder-notebook
+    # Status: Optional
+    registered_file_extensions = []
```
goanpeca: Let's make it immutable? a tuple perhaps?

ccordoba12: The only way this could be modified at runtime is if one of the core Spyder plugins changes it. So I don't see the need to make immutable.

goanpeca: Good practices :-p

jitseniesen:
@ccordoba12 What's the plan with this? How about I take over?

My intention is to do some more work on spyder-notebook as it seems there is quite some interest in it.

ccordoba12:
Closing because this was superseded by #22564.

PR #8837, opened 2019-02-24
PR: Automatically use a separate but persistent configuration directory for non-stable releases

### LINKED ISSUE #7680: Use a separate configuration directory for beta and dev releases to avoid contaminating user settings
Can the *.beta releases be updated to use a different assortment of dot files, so `.spyder` could become `.spyder-beta`, this would preserve the beta alongside the current `3.3.1` release so that a production and testing environment can co-exist happily. I also wonder whether the same could be possible for the `.ipython` and `.matplotlib` subdirs?

CAM-Gerlach:
## Description of Changes

Spyder betas and development versions (i.e. anything with alphabetical characters in the last version element), or with the ``SPYDER_DEV`` set) will automatically use a separate, but persistent config directory, in order to not pollute, modify or wipe the user's primary production config. This behavior can be customized/overriden (necessary for the test) with a environment variable ``SPYDER_USE_DEV_CONFIG_DIR`` environment variable. The tests took far more effort and complexity than they should to get working, but I finally got everything to play nice in the end.

### Issue(s) Resolved

Fixes #7680

### Affirmation

### DIFF
--- a/spyder/config/tests/test_base.py
+++ b/spyder/config/tests/test_base.py
@@ -10,11 +10,18 @@
 Tests for the spyder.config.base module.
 """
 
+# Standard library imports
+import os.path as osp
+try:
+    from importlib import reload
+except ImportError:  # A builtin on Python 2
+    pass
+
 # Third party imports
 import pytest
 
 # Local imports
-from spyder.config.base import is_stable_version
+import spyder.config.base
 
 
 # ============================================================================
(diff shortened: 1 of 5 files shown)


spyder/config/tests/test_base.py
```diff
@@ -0,0 +1,51 @@
...
+# Local imports
+import spyder.config.base
+
+
+# ============================================================================
+# ---- Tests
+# ============================================================================
+@pytest.mark.parametrize('version_input, expected_result', [
+    ('3.3.0', True), ('2', True), (('0', '5'), True), ('4.0.0b1', False),
+    ('3.3.2.dev0', False), ('beta', False), (('2', '0', 'alpha'), False)])
```
ccordoba12: This is quite confusing. Please don't abuse of `paremetrize` and simply test `is_stable_version` as it was tested before.

CAM-Gerlach: I'm not sure how this can be fairly characterized as "abuse" considering it follows an essentially *identical* pattern to the canonical example of parameterization in the [Pytest docs](https://docs.pytest.org/en/latest/parametrize.html), and I find it clearer and much less code smelly than the pathological antipattern that is a long list of identical aside from input/expected assert statements. However, it is perhaps slightly less performant and adds extra lines to the logs for a relatively minor test, so I de-parameterized it as requested.

CAM-Gerlach:
@ccordoba12 De-paramaterized the test and fixed the Python 2 issue as requested, thanks.

spyder/config/tests/test_base.py
ccordoba12: > I'm not sure how this can be fairly characterized as "abuse"

Because there are lot of test cases packed up in three lines and it's hard to understand what's going there.

ccordoba12:
Great, thanks!

spyder/config/tests/test_base.py
CAM-Gerlach: "Abuse" is a pretty serious accusation, though though I know you didn't mean anything bad by it. Its perfectly valid for someone to find a certain pattern confusing and ask for it to be made clearer, which I hopefully did. However, I don't see how calling it "abuse" of a certain feature is really a fair characterization when its exactly how said feature is designed to be used (it was virtually *identical* to the very first example given in the docs, if you replace the input/expected values with these ones and ``eval()`` with ``is_stable_version``).

CAM-Gerlach:
@ccordoba12 Checks green :heavy_check_mark:

PR #8842, opened 2019-02-24
PR: Change default root_path for PyLS and refactor LSP related code

ccordoba12:
* Before we were using HOME by default for `root_path` and that was causing a huge delay when trying to get Rope completions. Now we use an empty dir inside our config one, just like we did in Spyder 3.
* This PR also refactors and renames several methods and constants related to our LSP integration.

@steff456, this should fix the Rope delays you found before. Please try it on your side (by enabling `rope_completion` and disabling `jedi_completion`) and let me know if it works for you or not.

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -903,7 +903,7 @@ class MainWindow(QMainWindow):
 
         # Start LSP client
         self.set_splash(_("Launching LSP Client for Python..."))
-        self.lspmanager.start_client('python')
+        self.lspmanager.start_client(language='python')
 
         # Populating file menu entries
         quit_action = create_action(self, _("&Quit"),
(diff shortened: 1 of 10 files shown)


andfoy:
@ccordoba12 Is everything LSP-related more stable now?

steff456:
@ccordoba12, with this change the rope_completion is working! 

I checked out the completion race and now it is working as expected with this change.

ccordoba12:
> Is everything LSP-related more stable now?

@andfoy, yes, I think it is! We're working now with @steff456 to get feature parity with Spyder 3 and (as you know) making Rope completions work again in the PyLS is a top priority for us.

ccordoba12:
> with this change the rope_completion is working!

Great to hear!! Thanks @steff456 for the confirmation!

PR #8852, opened 2019-02-27
PR: Add an Object Explorer to the Variable Explorer

### LINKED ISSUE #558: Add Object Explorer for exploring an object's properties
_From [airdrik@gmail.com](https://code.google.com/u/airdrik@gmail.com/) on 2011-02-11T18:15:26Z_

Please add a tree-view of the currently-available objects and their properties.
e.g. given a class A with properties b and c, and an instance of class A called a, there would be a collapsible entry:

variable  type  value
a         A     <whatever repr(a) gives>

which would expand to:

variable  type  value
a         A     <whatever repr(a) gives>
+-b       str   hello
+-c       str   world

(a good example is in wxPython's PyCrust shell - the Namespace explorer)

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=558_

dalthviz:
## Description of Changes

imagen

imagen

### Issue(s) Resolved

Fixes #558 

### Affirmation


ccordoba12:
@dalthviz, before you proceed further, please break `objecteditor2.py` into multiple files (just like the original author of `objbrowser` organized its code) and make `objecteditor2` a module with those files inside it.

ccordoba12:
@dalthviz, this is looking nice, thanks! Jus another note so you keep working on this: I noticed some of the new files you added are using the `six` library (which seems to be [the only](https://github.com/titusjan/objbrowser/blob/master/setup.py#L38) external dependency of `objbrowser`).

Please remove that dependency and instead replace calls to `six` with calls to our own `py3compat` module. That way we'll avoid to make Spyder depend on `six`.

NOTICE.txt
```diff
@@ -1006,6 +1006,45 @@ Files covered:
...
+The current gtabview license can be viewed at:
+https://github.com/titusjan/objbrowser/blob/master/LICENSE.txt
+
+The current version of the original files can be viewed at:
+https://github.com/titusjan/objbrowser/tree/master/objbrowser
+
+
+Files covered:
+
+spyder/plugins/variableexplorer/widgets/objecteditor2/
```
CAM-Gerlach: We should be more specific and list the files covered, since the tests are not (and we don't want to misrepresent their authorship), nor the __init__.py files, nor anything we might add in the future.

CAM-Gerlach:
Overall, good job handling incorporating the external project, aside from a few minor issues I point out in my review. Additionally, one more thing:

Three different headers types are used in three different sets of files:
* Some places, like in __init__,py files, you use the one with the ``(c)`` copyright, no copyright year and ``(see LICENSE.txt for details)``
* Elsewhere, e.g. the tests, you use the one with the ``©`` copyright, no copyright year and ``(see spyder/__init__.py for details)``
* Still elsewhere, in files taken from the external project, you use the one with the ``(c)`` copyright, beginning copyright year and ``see NOTICE.txt in the Spyder root directory for details`` (correctly different, but horribly spliced over two lines instead of putting the line break after the ``;`` where it belongs, and where it is in other such files.

Particularly for new files, we should use one consistent form and stick with it. The canonical form I've been using in new files going forward is:

```
# -*- coding: utf-8 -*-
# -----------------------------------------------------------------------------
# [Copyright (c) YEARS OTHER AUTHOR(S)]
# Copyright (c) YEAR- Spyder Project Contributors
#
# Distributed under the terms of the LICENSENAME License
# (see [LICENSE.txt [in this directory] [and] [NOTICE.txt in the root] for details)
# -----------------------------------------------------------------------------
```

(Picking the appropriate bracketed verbiage for the use case). Unless the original file has explicit license-related text in its headers that is substantially different from just a statement that the file is released under the LICENSENAME (which if so, should be preserved) there is no need to repeat the same information (distributed under the MIT License) twice. Any other information should be included below the "see LICENSE.txt..." line separated by one blank line, so the license statement is clean and readable and there is no possibility of missing or misreading it.

In this case since this is included as a self-contained directory, you should include the original ``LICENSE.txt`` (as required by the license) in this directory, adding ``Copyright (c) 2019- Spyder Project Contributors (see AUTHORS.txt for details)``` below the original copyright statement line in that file, and use the following form of the above line for files originally from the external project:

``# (see LICENSE.txt in this directory and NOTICE.txt in the root for details)``

dalthviz:
@CAM-Gerlach thank you for testing this and the detailed list of missing elements! I will take them as a TODO list :+1:

CAM-Gerlach:
Thanks @dalthviz ! And if any of them are determined not to be in scope here, I can make them a separate (meta)-issue instead.

ccordoba12:
@dalthviz, this is looking really good, thanks! Before proceeding further, please fix our tests. This is important to be sure that you're not breaking previous functionality.

ccordoba12:
@CAM-Gerlach, I edited your comment above to remove this task from the list

> Functionality: The object browser does not work if the package the object's class is from is not installed, which is common under our recommended workflow of keeping Spyder in a separate environment from the working environment packages (especially important if their deps are incompatible with Spyder, or they could contaminate Spyder's environment with pip or conda-forge)

I've been thinking about it and this is out of the scope of this PR. To solve it we need to create a pure text, json representation of an object in the kernel and send it to Spyder to display it in this new object editor. So, please open a new issue about it in `spyder-kernels`.

Besides, this is a complex project, so we're not going to address it for Spyder 4.

ccordoba12:
> Also, is there any way to avoid the smelly objecteditor/objecteditor2/objecteditor2.objecteditor situation? It just reeks of inelegance and the structure is confusing and unclear to someone not intimately familiar with the codebase.

@CAM-Gerlach, this could have been worded way better (and it's another example of the kind of comments that I really don't like from you, as I told you privately). Besides, instead of making a concrete suggestion, you just point to the problem.

@dalthviz, please rename `objecteditor2/objecteditor.py` to `objecteditor2/editor.py` and also `test_objecteditor.py` to `test_editor.py`

CAM-Gerlach:
> I've been thinking about it and this is out of the scope of this PR. 

Yes, I only put it here since you had mentioned to me that it might be able to be resolved with this implementation. Thanks for thinking about it. Therefore, per your instructions, I opened spyder-ide/spyder-kernels#99 .

> this could have been worded way better

Yes, you are absolutely right. I got way too "editorial"/opinionated there, as I sometimes do, and used much too over the top language to describe the problem. Therefore, I edited my comment to remove all of that as well as make a more specific suggestion of a potential change to fix the issue.

>  please rename objecteditor2/objecteditor.py to objecteditor2/editor.py and also test_objecteditor.py to test_editor.py

Actually, this doesn't really fix the problem I was mentioning, as it retains the ``objecteditor2`` vs. ``objecteditor`` confusion and furthermore just calling it the "Editor" is even less clear (currently, it can't even "edit" anything), and at odds with the parent being called ``objecteditor2``. This was ultimately my fault for not providing a concrete suggestion to make clear what I intended.

Instead, I'd suggest renaming ``objecteditor2`` and ``objecteditor2/objecteditor.py`` to ``objectexplorer`` and ``objectexplorer/objectexplorer.py``, to match the official name of the feature, and its functionality since it doesn't actually allow editing yet and is quite different from all the other "editors".

ccordoba12:
> Instead, I'd suggest renaming objecteditor2 and objecteditor2/objecteditor.py to objectexplorer and objectexplorer/objectexplorer.py

Ok, I agree with that, it's a good suggestion. @dalthviz, please follow @CAM-Gerlach's suggestion instead of mine.

dalthviz:
A preview of the menu action for the `Object Explorer`:

oexplorer

ccordoba12:
This is looking really good, thanks @dalthviz!

My comments:

* [ ] The object name that appears in the explorer needs to be the actual object name (not `Object`, as it is shown now, but the one displayed in the Variable Explorer).
* [ ] Let's change the title of the dialog to: <Name> - Object.
* [ ] Let's remove the Refresh and Auto-refresh buttons because they are not doing anything.
* [ ] Let's remove the following columns: summary, repr, str, unicode, inspect predicates, type, inspect.getmodule.
* [ ] Let's rename the following columns:
    - [ ] type name: Type
    - [ ] is attribute: Attribute
    - [ ] is callable: Callable
    - [ ] is routine: Routine
    - [ ] inspect.getfile: File
    - [ ] inspect.getsourcefile: Source file
    - [ ] Length: Size
    - [ ] path: Path
* [ ] Let's add a column called Value that displays values using the function called `value_to_display` from spyder-kernels.
* [ ] Let's make the order of columns like this: Name, Type, Size, Value, Callable, Path. The others need to be hidden by default.
* [ ] Let's make showing callables the default.
* [ ] Let's remove the blue color for callables and italics for text. All text must use the same, regular font.
* [ ] Let's hide the `dict` callables that appear at first level (i.e. at the same level of the object). 
* [ ] For the Details section:
    - [ ] Let's remove: pretty print, path, unicode, inspect.getcomments.
    - [ ] Let's rename:
        * [ ] inspect.getdoc: Documentation
        * [ ] inspect.getfile: File
        * [ ] inspect.getsource: Source code
    - [ ] Let's make Documentation the first and default entry.
    - [ ] Let's use two instances of CodeEditor for the widget to the right of Details. One should have Python as it's language (for Source Code) and the other should have its language in Rest (For the Documentation, File and repr).
    - [ ] Let's use the same font used in the editor for those CodeEditor instances.

These are a lot of tasks, but fortunately almost all of them are small.

dalthviz:
@ccordoba12 also spyder-ide/spyder-kernels#100 needs to be merge for this to be done

dalthviz:
A new preview:

oe

dalthviz:
A new preview with the `CodeEditor` as the details widget:

oe_details

spyder/plugins/variableexplorer/widgets/namespacebrowser.py
```diff
@@ -116,7 +122,10 @@ def setup(self, check_all=None, exclude_private=None,
                         data=None,
                         minmax=minmax,
                         shellwidget=self.shellwidget,
-                        dataframe_format=dataframe_format)
+                        dataframe_format=dataframe_format,
+                        show_callable_attributes=show_callable_attributes,
+                        show_special_attributes=show_special_attributes
+                        )
```
ccordoba12: This is poorly indented (and I was the one who did it, so sorry for that). It should be

```python
self.editor = RemoteCollectionsEditorTableView(
    self,
    data=None,
    minmax=minmax,
    shellwidget=self.shellwidget,
    dataframe_format=dataframe_format,
    show_callable_attributes=show_callable_attributes,
    show_special_attributes=show_special_attributes
)
```

spyder/plugins/variableexplorer/widgets/namespacebrowser.py
```diff
@@ -219,6 +228,7 @@ def setup_options_button(self):
         if not self.options_button:
             self.options_button = create_toolbutton(
                 self, text=_('Options'), icon=ima.icon('tooloptions'))
+            self.options_button.setPopupMode(QToolButton.InstantPopup)
```
ccordoba12: Why is this necessary?

spyder/plugins/variableexplorer/widgets/objectexplorer/utils.py
```diff
@@ -0,0 +1,48 @@
...
+# the MIT (Expat) license.
+# Licensed under the terms of the MIT License; see NOTICE.txt in the Spyder
+# root directory for details
+# -----------------------------------------------------------------------------
+
+# Standard library imports
+import logging
+
+
+def logging_basic_config(level='INFO'):
```
ccordoba12: This function is not necessary because we already setup our logging in a particular way. Besides, it doesn't seem to be used in the rest of the code in this PR, so please remove it.

spyder/plugins/variableexplorer/widgets/objectexplorer/utils.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    Setup basic config logging. Useful for debugging to
+    quickly setup a useful logger.
+    """
+    fmt = '%(filename)25s:%(lineno)-4d : %(levelname)-7s: %(message)s'
+    logging.basicConfig(level=level, format=fmt)
+
+
+def check_class(obj, target_class, allow_none=False):
```
ccordoba12: This function doesn't seem to be used either in this PR. If that's the case, please remove it.

NOTICE.txt
```diff
@@ -1006,6 +1006,50 @@ Files covered:
...
+objbrowser is distributed under the MIT license.
+
+
+Use of the core functionality of the project changing the non-patched
+version of qtpy used for the actual qtpy project (since it is already a
+dependency of Spyder)
+
+See below for the full text of the MIT (Expat) license.
+
+The current gtabview license can be viewed at:
```
ccordoba12: `gtabview` -> `objbrowser`.

NOTICE.txt
```diff
@@ -1006,6 +1006,50 @@ Files covered:
...
+
+
+Classes from objbrowser 1.2.1
+-----------------------------
+
+
+Copyright (c) 2016 Pepijn Kenter
+
+
+Author: Pepijn Kenter | <author-email> | <author-website>
```
ccordoba12: Please try to add `author-email` here (you can remove `author-website` if you can't find one quickly).

NOTICE.txt
```diff
@@ -1006,6 +1006,50 @@ Files covered:
...
+
+Modifications made to each class to adapt them to Spyder's needs.
+
+
+objbrowser is distributed under the MIT license.
+
+
+Use of the core functionality of the project changing the non-patched
+version of qtpy used for the actual qtpy project (since it is already a
+dependency of Spyder)
```
ccordoba12: Please change this to

> Use the core functionality of the project, mainly changing the code to use
the upstream version of qtpy project (since it is already a dependency of
Spyder).

ccordoba12:
Some last comments about the organization of the widget:

- [ ] Let's organize the entries in `Details` like this: Documentation, Source Code, File, repr.
- [ ] Let's add an asterisk next to `repr (*)`.
- [ ] Please add a QLabel at the bottom of `Details` and that expands both its entries and the CodeEditor widget next to it. This QLabel needs to have the following text: `(*) Some objects have very large repr's, which can freeze Spyder. Please use this with care`.

spyder/plugins/variableexplorer/widgets/namespacebrowser.py
dalthviz: I think is not necessary, this line probably got there when I was checking the `options_button` for the `Object Explorer`. I will remove it :+1:

dalthviz:
A new preview:

imagen

ccordoba12:
@dalthviz, please don't forget to open an issue about this:

> Major Regression: Cannot edit or open a viewer/editor for collections, values, arrays, dataframes etc. that are attributes of objects, like I could before

so you don't forget to address it in beta4.

PR #8907, opened 2019-03-08
PR: Fix force_refresh to only force with valid object entries (Help plugin)

### LINKED ISSUE #8762: Text written in help pane gets automatically selected and replaced with every keystroke
**Steps to reproduce:**
* Type something in the help pane, in the field named “Object” where one puts names like `np.arange`.

**Expected behavior:**
After every keystroke, the text that was written there should remain, unless deleted or selected by the user.

**Actual behavior:**
After every keystroke, the full text there becomes selected, which makes it be deleted/replaced with every keystroke (so it only displays the last character that one typed).

**Reproducible:** always

**Setup:** Spyder 3.3.3, Python 3.6.5, running on either Windows 10 or Debian Linux (KDE).

dalthviz:
## Description of Changes

Use value of the `ObjectComboBox.valid` signal to ensure a force refresh only with valid values

### Issue(s) Resolved

Fixes #8762 

### Affirmation

### DIFF
--- a/spyder/plugins/help.py
+++ b/spyder/plugins/help.py
@@ -397,7 +397,7 @@ class Help(SpyderPluginWidget):
         self.combo.setMaxCount(self.get_option('max_history_entries'))
         self.combo.addItems( self.load_history() )
         self.combo.setItemText(0, '')
-        self.combo.valid.connect(lambda valid: self.force_refresh())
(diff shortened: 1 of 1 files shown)


ccordoba12:
Merging because the failure in Azure is unrelated to this.

Issue #134, opened 2015-02-16
matplotlib plot does not display properly in spyder.


spyder-issues-migrator:
_From [ron...@gmail.com](https://code.google.com/u/101299234695429332746/) on 2010-01-07T12:31:45Z_

What steps will reproduce the problem?

 1. In spyder --pylab, In the interactive console, plot: t =
linspace(1,10,20); plot(t,t); (Or anything else..)
2. Close the plot-output window.
3. Plot again: plot(t,t)
4. An Error presented: 
 plot(t,t**3)
[<matplotlib.lines.Line2D object at 0xaeadbcc>]Traceback (most recent call
last):
  File "/usr/lib/pymodules/python2.6/matplotlib/backends/backend_qt4.py",
line 205, in idle_draw
    self.draw()
  File
"/usr/lib/pymodules/python2.6/matplotlib/backends/backend_qt4agg.py", line
131, in draw
    self.update()
RuntimeError: underlying C/C++ object has been deleted

Different symptom: If I run spyder without --pylab:
t = linspace(1,10,20)
from pylab import *
t = linspace(1,10,20)
plot(t,t) #Nothing happen
show() #A graph window opens, this time, I can't dock it to spyder
\#Closing the graph window
show()
Traceback (most recent call last):
  File "/usr/lib/pymodules/python2.6/matplotlib/backends/backend_qt4.py",
line 63, in show
    manager.window.show()
RuntimeError: underlying C/C++ object has been deleted What version of the product are you using? On what operating system? Python 2.6.4 ( r264 :75706, Dec  7 2009, 18:45:15) 
[GCC 4.4.1] on linux2
\>>> import spyderlib
\>>> spyderlib.__version__
'1.0.1'
cat /proc/version
Linux version 2.6.31-16-generic (buildd@rothera) (gcc version 4.4.1 (Ubuntu
4.4.1-4ubuntu8) ) `#53`-Ubuntu SMP Tue Dec 8 04:01:29 UTC 2009

installed using 
sudo easy_install spyder

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=134_

spyder-issues-migrator:
_From [pierre.raybaut](https://code.google.com/u/pierre.raybaut/) on 2010-01-07T09:53:41Z_

Already fixed in mercurial repository (actually this is a PyQt 4.6 bug)

**Status:** Duplicate  
  **Mergedinto:** 94

spyder-issues-migrator:
[...]

Issue #131, opened 2015-02-16
getargtxt runs into problems when there are tuples in the arguments


spyder-issues-migrator:
_From [schreib.mal.sevi@gmail.com](https://code.google.com/u/schreib.mal.sevi@gmail.com/) on 2009-12-19T08:32:22Z_

Example:
def remove(self, (sub, pred, obj)):
...

When I then type eg. ab.remove it throws an error
as soon as I hit the left parentheses key concluding
that:
TypeError: cannot concatenate 'str' and 'list' objects

bg.removeTraceback (most recent call last):
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.6/lib/
python2.6/site-packages/spyder-1.0.1-py2.6.egg/spyderlib/widgets/
interactiveshell.py", line 363, in keyPressEvent
    self.postprocess_keyevent(event)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.6/lib/
python2.6/site-packages/spyder-1.0.1-py2.6.egg/spyderlib/widgets/
shell.py", line 764, in postprocess_keyevent
    ShellBaseWidget.postprocess_keyevent(self, event)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.6/lib/
python2.6/site-packages/spyder-1.0.1-py2.6.egg/spyderlib/widgets/
shell.py", line 420, in postprocess_keyevent
    self._key_parenleft(text)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.6/lib/
python2.6/site-packages/spyder-1.0.1-py2.6.egg/spyderlib/widgets/
shell.py", line 853, in _key_parenleft
    self.show_docstring(self.get_last_obj(), call=True)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.6/lib/
python2.6/site-packages/spyder-1.0.1-py2.6.egg/spyderlib/widgets/
shell.py", line 944, in show_docstring
    arglist = self.get_arglist(text)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.6/lib/
python2.6/site-packages/spyder-1.0.1-py2.6.egg/spyderlib/widgets/
interactiveshell.py", line 520, in get_arglist
    return getargtxt(obj)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.6/lib/
python2.6/site-packages/spyder-1.0.1-py2.6.egg/spyderlib/utils/
[...]

Issue #112, opened 2015-02-16
Installer/Spyder 1.0.1 doesn't work


spyder-issues-migrator:
_From [techtonik@gmail.com](https://code.google.com/u/techtonik@gmail.com/) on 2009-11-09T04:44:45Z_

What steps will reproduce the problem?

 1.Download spyder-1.0.1.exe
2.Install
3.Execute shortcut What is the expected output? What do you see instead? Nothing happens. What version of the product are you using? On what operating system? Vista 

Please provide any additional information below

. I do not have PyQT4 installed, so either installer is incomplete and should 
be updated or Spyder should warn about missing library or at least leave 
some trace on the screen about error occurred. It would be also nice to see 
Spyder download PyQT4, install it into its own Spyder's PYTHONPATH and 
restart.

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=112_

spyder-issues-migrator:
_From [pierre.raybaut](https://code.google.com/u/pierre.raybaut/) on 2009-11-09T07:32:29Z_

Spyder is just a Python package: it has dependencies but like for any other Python
package ( http://pypi.python.org/pypi ) the Windows installer does not handle dependencies.
The installation notes are available here: https://code.google.com/p/spyderlib/wiki/Installation That's why distributions like Python(x,y) are made for: to solve this kind of issues
related to the fact that Windows does not have an integrated package manager system
(unlike Ubuntu for example).

Note that it does not install PyQt4 but it does not install NumPy, SciPy, PIL or
matplotlib whereas these modules are as important as PyQt4. So this kind of request
has no end and the conclusion is always the same: on Windows, the end user has to
either install all the requirements on his own, or use ready-to-use distributions
like Python(x,y) which includes PyQt4, NumPy, SciPy, PIL and matplotlib.

**Status:** WontFix

PR #9014, opened 2019-03-23
PR: Use cell name in runcell to run it from the console and add functionality to debug cells

### LINKED ISSUE #9361: Add a command to debug a single cell
### Issue Report Checklist

    * [ ] Restarted Spyder

## Problem Description
If I have cells in my code that take a long time to run, I want to avoid having to re run everything. Yet, if I want to debug just a part of my code, this is what I need to do.
It would be great to use a mechanism similar to `runcell` (called `debugcell`) to debug a single cell.

The logic is implemented in https://github.com/spyder-ide/spyder-kernels/pull/112, but I wonder if there should be a keyboard shortcut or something to trigger that from spyder.

If https://github.com/spyder-ide/spyder-kernels/pull/112 is merged, should there be a mechanism to trigger the cell debugging? For now I type `debugcell('mycell')` in the console.

### LINKED ISSUE #9963: Running an empty cell is broken

The error happens when running an empty cell
raise RuntimeError('Not a cell?')

### Traceback
```python-traceback
  File "C:\Users\Chan Chak Tong\Anaconda3\lib\site-packages\spyder\plugins\editor\widgets\editor.py", line 2551, in run_cell
    self._run_cell_text(text, block)
  File "C:\Users\Chan Chak Tong\Anaconda3\lib\site-packages\spyder\plugins\editor\widgets\editor.py", line 2614, in _run_cell_text
    cell_name = self._get_cell_name(block)
  File "C:\Users\Chan Chak Tong\Anaconda3\lib\site-packages\spyder\plugins\editor\widgets\editor.py", line 2595, in _get_cell_name
    raise RuntimeError('Not a cell?')
RuntimeError: Not a cell?
WARNING:spyder.widgets.github.backend:Failed to send bug report on Github. response={'code': 401, 'json': {'message': 'Bad credentials', 'documentation_url': 'https://developer.github.com/v3'}}
```

## Versions

* Spyder version: 4.0.0b4 
* Python version: 3.7.3
* Qt version: 5.9.6
* PyQt5 version: 5.9.2

### Dependencies

```
qdarkstyle >=2.6.4:  2.7 (OK)
qtconsole >=4.5.2 :  4.5.2 (OK)
IPython >=4.0     :  7.7.0 (OK)
```

impact27:
## Description of Changes

 - This PR needs an api to call spyder from the kernel, which is implemented in #9343 and https://github.com/spyder-ide/spyder-kernels/pull/111/.
 - This PR needs unique cell names, which is implemented in #9082 
 - Changes to the kernel are implemented in https://github.com/spyder-ide/spyder-kernels/pull/112/

This PR sends a request to spyder to receive the code from a cell when the `runcell()` function is called. This allows to repeat calls to the same cell from the console, and to run cells in a loop.

### Issue(s) Resolved

Fixes #9963.
Fixes #9361.

### Affirmation

impact27


ccordoba12:
@bcolsen, please take a look at this one.

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -522,10 +522,14 @@ def norm(text):
...
             # Internal kernels, use runcell
             if client.get_kernel() is not None and not run_cell_copy:
-                line = (to_text_string("{}('{}','{}')")
-                            .format(to_text_string('runcell'),
-                                (to_text_string(cell_name).replace("\\","\\\\")
-                                    .replace("'", r"\'")),
+                cell_name = (to_text_string(
+                        cell_name).replace("\\", "\\\\").replace("'", r"\'"))
+                if cell_name == '':
+                    cell_name = 'Unnamed cell (l.{})'.format(line_number)
```
CAM-Gerlach: ``(line {})`` Explicit is better than implicit.

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -522,10 +522,14 @@ def norm(text):
         if client is not None:
             # Internal kernels, use runcell
             if client.get_kernel() is not None and not run_cell_copy:
-                line = (to_text_string("{}('{}','{}')")
-                            .format(to_text_string('runcell'),
-                                (to_text_string(cell_name).replace("\\","\\\\")
-                                    .replace("'", r"\'")),
+                cell_name = (to_text_string(
+                        cell_name).replace("\\", "\\\\").replace("'", r"\'"))
+                if cell_name == '':
```
CAM-Gerlach: ``if not cell_name``?

spyder/plugins/editor/widgets/editor.py
```diff
@@ -2532,9 +2532,10 @@ def _run_cell_text(self, text, line):
         except AttributeError:
             cell_name = ''
         if finfo.editor.is_python() and text:
-            self.run_cell_in_ipyclient.emit(text, cell_name,
-                                            finfo.filename,
-                                            self.run_cell_copy)
+            self.run_cell_in_ipyclient.emit(
+                    text, cell_name, finfo.filename,
+                    self.run_cell_copy,
+                    line)
```
CAM-Gerlach: Why put this on its own line?

CAM-Gerlach:
Also, FYI @impact27 , Github offers a new Draft Pull Requests feature you can use instead of old-style WiP tagging to properly mark your PR as a draft (see the dropdown menu in the submit PR button when submitting your PR).

bcolsen:
@impact27 I like the idea of having a default code cell name so much I would like to see this default cell name change pushed into syntaxhighlighters.py so the outline explorer can have a default name for unnamed cells as well. Currently the cells are just empty in the outliner. This would mean reveting all the changes you have made to these file, but I will show you where they can go and also it should mean fewer changes to the code overall.

Also, I agree with @CAM-Gerlach about the format of the current default cell name being not explicit enough. I think the default cell name should be `"Unnamed cell, line {}"`, this would be similar to line numbers in tracebacks.

I think this can be done with a few lines of code placed around here:

https://github.com/spyder-ide/spyder/blob/dd631ac9bfeab94f74e8372e6703fb0aad08b17a/spyder/utils/syntaxhighlighters.py#L488

or in the `get_code_cell_name` function directly:

https://github.com/spyder-ide/spyder/blob/dd631ac9bfeab94f74e8372e6703fb0aad08b17a/spyder/utils/syntaxhighlighters.py#L388

You should be able to get the code line in the `PythonSH` class like this:
`block_nb = self.currentBlock().blockNumber()`
The line number would need to be passed to `get_code_cell_name` if you make the changes there.

As for a test, I think you can safely remove this code cell name to be an unnamed cell and then change the test so it works with the default name vs `MGroup4`

https://github.com/spyder-ide/spyder/blob/dd631ac9bfeab94f74e8372e6703fb0aad08b17a/spyder/plugins/outlineexplorer/tests/test_widgets.py#L127

bcolsen:
>I think I should split this PR into 2:
>
>    The default name for unnamed cell
>    The runcell code

Please split this PR as you said. It's getting a little too broad in its scope.

>For the runcell part, the logic is as follow:
>
>    If run the cell from the editor, save the code and then call runcell. (This could be changed to only use the second code pathway)
>    If run from the console, get the code from the editor if it is there, otherwise get from the saved file.

I'm :-1: for running code cells from files especially closed ones. This seems like a bad practice. I don't like how we would have the code cell parsing logic twice in code. 

I also don't think that saving the file should be necessary for code cells.

Line numbers in general are very fragile, they change easily I would rather not define code cells this way.

CAM-Gerlach:
> I'm -1 for running code cells from files especially closed ones. This seems like a bad practice. I don't like how we would have the code cell parsing logic twice in code.

The issue I see with this is given the identified use case—re-running the cell with ``Up, Enter``—is this would not pick up the changes to the code. I don't see major utility in of re-running the exact same code again, at least for a workflow with deterministic output (or even non-deterministic, except for certain cases). What is valuable, IMO, is being able to re-run the last cell, but with whatever code modifications have been made.

This is exactly what the "Run Last Cell" command already does, and it seems to track the same cell even when moved around, lines moved up and down, the cell contents modified and re-ordered since it was run, whether or not the cell is named or the file is saved. Could you just use the same logic to track the cell to run with ``Up, Enter``?

bcolsen:
> The issue I see with this is given the identified use case—re-running the cell with Up, Enter—is this would not pick up the changes to the code. 

This code actually does get the changes from the editor.

>This is exactly what the "Run Last Cell" command already does, and it seems to track the same cell even when moved around, lines moved up and down, the cell contents modified and re-ordered since it was run, whether or not the cell is named or the file is saved. Could you just use the same logic to track the cell to run with Up, Enter?

Yes, that's possible.

I think it is overly complicated to have runcell() be able execute arbitrary code cells. If runcell() should do anything when run from the console it should run the current cell that's highlighted in the editor. Even this could have some corner cases with split and extra panes, but then we would have to call runcell() without the file and cell names loosing the history part.

On the other hand this would allow multiple code cells to be executed at once or even in a loop.

impact27:
> I'm 👎 for running code cells from files especially closed ones. This seems like a bad practice. I don't like how we would have the code cell parsing logic twice in code.

I changed it to not load code from closed files. Only files open in the editor will be parsed. The logic is not duplicated as I am using `editor.get_cell_as_executable_code`

> I also don't think that saving the file should be necessary for code cells.

It is not

> Line numbers in general are very fragile, they change easily I would rather not define code cells this way.

I changed it to a combination of line number and cell name. Both together should be more robust.

> I think it is overly complicated to have runcell() be able execute arbitrary code cells. If runcell() should do anything when run from the console it should run the current cell that's highlighted in the editor. Even this could have some corner cases with split and extra panes, but then we would have to call runcell() without the file and cell names loosing the history part.

If the user knows the filename, the cell name and the cell number, why not let him run the code?
I am not sure I see the corner case. Is it possible to have 2 versions of the same file open in the editor? I thought it was not even with split.

impact27:
This PR would solve the `Up + Enter` problem in a way the user would expect. The problems I see with the current implementation is that the runcell function has a behaviour that is extremely surprising.
 - I can call, `debugfile` and `runfile` from the console but not `runcell`
 - It enters the history but can't actually be executed
 - If we implement the "Run cell under cursor", it would be the only function that depends on the cursor position! I would expect it to depend on the state of the code, but not on the cursor.

With this PR:
 - you can call `runcell` from the editor or from the console with the same effect. The cell is identified by name and line number.
 - Repeat a call to `runcell` from the history and have the same result. (Unless the code changed but this is the same as `runfile` and `debugfile`)

CAM-Gerlach:
@impact27 Overall, the design of your implementation sounds reasonable and consistent to me.

> I changed it to a combination of line number and cell name. Both together should be more robust.

So if a named code cell changes its start line, what happens? What about an unnamed cell? Currently, the "Run last cell" command already appears to still track the correct cell in these situations (somehow...); what about ``runcell``? Could the two implementations be unified (i.e. ``Run last cell`` just calls ``runcell()`` with the previous arguments, in turn ``runcell()`` always finds the cell to run via ``Run last cell``'s approach if not run directly from the Editor?

impact27:
> So if a named code cell changes its start line, what happens?

The errors are:
 - `"File {} not open in the editor"` : The file is not open
 - `"Cell first line is {}, not {}"`: The line doesn't correspond to a cell
 - `"Cell at line {} is called \\"{}\\", not \\"{}\\"."`: The name doesn't match



> What about an unnamed cell?

An unnamed cell is just a cell named '', or whatever the default name will be. This is not a special case.



> Currently, the "Run last cell" command already appears to still track the correct cell in these situations (somehow...);

The cursor is saved and it is used to perform the action in the same way. Maybe not the more robust.

> what about `runcell`? Could the two implementations be unified (i.e. `Run last cell` just calls `runcell()` with the previous arguments, in turn `runcell()` always finds the cell to run via `Run last cell`'s approach if not run directly from the Editor?

I could implement a new logic to replace and do that. What happens if `run_cell_copy` is true?

impact27:
Also Run last cell runs the last cell from the current editor?

bcolsen:
> The cursor is saved and it is used to perform the action in the same way. Maybe not the more robust.

I actually like the cursor position better than line numbers because that's how runcell() works in the editor. The current cell is defined by the current cursor position. The runcell function could be very robust by taking a file name, cursor position and name. All these values could be optional(see the pseudo code below). The runcell command from `ctrl+enter ` would populate runcell function values with the current file, current cursor position and current cell name. In this case the unnamed cells would only be called `Unnamed cell` with no line number.

Runcell pseudo code:
``` python
def runcell(file=None, line=None, name=None):
    if file is None:
        file = get_cell_file_in_editor()

    if line is not None and name is not None:
        cell_name = get_cell_name_by_position(name)
        Assert cell_name == name # error: "Cell at line {} is called \\"{}\\", not \\"{}\\"."
        cursor_pos = line
    elif line is not None and name is None:
        cursor_pos = line
    elif line is not None and name is None:
        # If only a name is supplied pick the last cell in the file with that name. 
        # It's the same as redefining a function in python.
        # or the kinder option raise an error that the cell isn't unique
        cursor_pos = get_cell_cursor_pos_by_name(name)
    else:
        cursor_pos = get_current_cursor_position()

    cell = get_cell(file, cursor_pos)
```

impact27:
The problem with a cursor is that you can't print it. It needs to be passed as a string to the console.

bcolsen:
> The problem with a cursor is that you can't print it. It needs to be passed as a string to the console.

Sorry I didn't mean the cursor object just the starting line number of the cursor. So the effectively any line number with in the cell would be fine.

impact27:
> The runcell function could be very robust by taking a file name, cursor position and name

I would tend to disagree. If I can just reference any line in the cell and then run, this could create some ambiguity as to which cell is actually running, especially if all the unnamed cells are called `Unnamed cell`. I think it would be more robust to have a unique identifier for a cell: `(filename, cell_name, line_number)`, which is what the code is doing now. The `Run last cell` logic now only works if this identifier is still valid. (Should I display a warning if it is not?)


> All these values could be optional

I think this would complicate things a lot, and create gaps in the history.

bcolsen:
> I think it would be more robust to have a unique identifier for a cell

I agree, a unique identifier would be best, but the line numbers aren't very stable. We don't have line numbers for breakpoints or for the new cursor bookmarks for this reason.

Here's another idea, we can define the cell by it's name if it's unique, or if name isn't unique(or unnamed) we can give it a unique name.

This name could be based on the name plus the order of the cells in the file like `Unnamed cell #1` or maybe just `[1]` like the notebook cells counting from the top. We can have #9082 implement a `cell_data` dictionary of cells with unique names and keys and line_numbers as values. Then we can call runcell like: `runcell("mycell")` and it will pull runcell from the current file, that's a nice simple interface or specify a file like `runcell("mycell", file="/path/to /file")` or `runcell("Unnamed cell #7", file="/path/to /file")`. If you use unique names, you could even write runcell functions in your code to run more than one cell at a time.

You can see all the names of the cells in the Outliner so you know what they are and if they are unique. The cell order in the file can change but it's much more stable than the line number. 

pseudo code for #9082:
``` python
cell_data={}
for syntax_parsing_loop:
    cell_name = get_code_cell_name(text)
    if cell_name == '':
        cell_name = 'Unnamed cell' # or just '[{}]'.format(len(cell_data))
    if cell_name in cell_data.keys():
        cell_name += ' #{}'.format(len(cell_data))
    cell_data[cell_name] = self.currentBlock().firstLineNumber() + 1
    oedata.def_name = cell_name
```
Now all the cell names in the file are unique and the line numbers of the cells are updated automatically when the file changes so we don't have to worry about then.

impact27:
The problem with the syntax highlighter is that it only parses the line that was changed.

So if I have
```
cell [1]
cell [2]
cell [3]
```
And add a cell in the middle, then the cell will be named:
```
cell [1]
cell [4]
cell [2]
cell [3]
```
Then, on the next full scan:
```
cell [1]
cell [2]
cell [3]
cell [4]
```
Which means that the cell number uniquely identifies a cell, but that the cell it identifies can change!

I would rather print an error than taking the risk of surprising the user. If the user calls runcell with a `file_name`, a `cell_name`, and a `line_number`, he will understand that this call fails if the line number or cell name is incorrect, which should catch most of the errors. And if the cell is somehow replaced by another cell with the same name and file number, the user will understand why the other cell was executed.

I see that tracking the cell when it changes would be great, but the complexity and number of edge cases it adds is huge! What if the cell is split in two? What if it is renamed? What if two cells are swapped? Having the code simply look for a cell called `cell_name` at line `line_number` is unsurprising to the user and fixes the "execute from history" problem.

bcolsen:
> The problem with the syntax highlighter is that it only parses the line that was changed.

That is unfortunate. It seems to change the whole file when triple quotes are added. Code cells could be implemented similarly. I think that whole oe_data system might need to be redone for #7976 eventually.

But that is certainly outside the scope here.

CAM-Gerlach:
> if name isn't unique(or unnamed) we can give it a unique name.

> This name could be based on the name plus the order of the cells in the file 

For non-named cells, this would be somewhat of a regression on the current ``Run Last Cell`` functionality I mentioned basing this on (which has never been a problem for users despite it being a prominent command in the UI, versus a secondary method like this). The latter correctly handles added and removed cells above and below the current one in my testing, still running the correct, previously run cell. However, as a small minority of changes would involve changing cell order, its not anywhere near as a huge a cost to pay as requiring the line number to be completely static between runs.

>  If the user calls runcell with a file_name, a cell_name, and a line_number, he will understand that this call fails if the line number or cell name is incorrect, which should catch most of the errors.

Normally I am all in favor of explicit over implicit and failing fast with a clear error rather than succumbing to the temptation to guess, but this is taking it to a bit of an extreme in my personal opinion, by artificially crippling this functionality to break completely whenever the start line of any cell changes, even when in the huge majority of cases it is easy and unambiguous to track and the functionality already exists to do so. Therefore, I don't see why we should artificially limit this functionality for the sake of the potential for a few minor corner cases, particularly for the more experienced users likely running cells via this secondary method anyway.

> he will

Or she, or they...

``<Puts copyeditor hat on>`` Not sure if you're a native, but in modern English convention "they" has become the standard gender-nuetral singular pronoun. I used to use "he or she" religiously, but its really quite awkward and outdated these days and widely accepted to just use "they"; if its good enough for Shakespeare its good enough for me. ``</hat>``

bcolsen:
> this would be somewhat of a regression on the current `Run Last Cell` functionality I mentioned basing this on

I agree. The `Run Last Cell` functionality is based on the cursor position from the last time a cell was run. Cursor positions are based on the text block not the line numbers so this is robust against line changes in the same way as the debug points and bookmarks. I think the `Run Last Cell` functionality is currently changed in this PR but it doesn't need to be.

> to break completely whenever the start line of any cell changes

I agree. As [Raymond Hettinger](https://www.youtube.com/watch?v=p33CVV29OG8) would say: "There must be a better way!"

impact27:
> > The problem with the syntax highlighter is that it only parses the line that was changed.
> 
> That is unfortunate. It seems to change the whole file when triple quotes are added. Code cells could be implemented similarly. I think that whole oe_data system might need to be redone for #7976 eventually.
> 
> But that is certainly outside the scope here.

If I am not mistaken, if a line is changed by the syntax highlighter, the next one will be checked. Anyway, the problem of a single name being reordered is still there. Why doesn't the oe_data system keep track of the block instead of the line number? 

> For non-named cells, this would be somewhat of a regression on the current `Run Last Cell` functionality I mentioned basing this on (which has never been a problem for users despite it being a prominent command in the UI, versus a secondary method like this). The latter correctly handles added and removed cells above and below the current one in my testing, still running the correct, previously run cell. However, as a small minority of changes would involve changing cell order, its not anywhere near as a huge a cost to pay as requiring the line number to be completely static between runs.

The current behaviour for `Run Last Cell` can be easily kept. I agree this would be a regression. I changed it to keep things consistent (`Run Last Cell` = `Up + Enter`), but we can as well change this.

> Normally I am all in favor of explicit over implicit and failing fast with a clear error rather than succumbing to the temptation to guess, but this is taking it to a bit of an extreme in my personal opinion, by artificially crippling this functionality to break completely whenever the start line of any cell changes, even when in the huge majority of cases it is easy and unambiguous to track and the functionality already exists to do so. Therefore, I don't see why we should artificially limit this functionality for the sake of the potential for a few minor corner cases, particularly for the more experienced users likely running cells via this secondary method anyway.

What if I just rename the cell? Should I rerun the cell? Is there a way of referencing a block rather than a line? maybe id(block)? 

what if I have:
```
9
10 #%%
11 a = 1
12 #%% 
```
then remove line 9:
```
9  #%%
10 a = 1
11 #%% 
```
Which cell would line 11 run? 



> `<Puts copyeditor hat on>` Not sure if you're a native, but in modern English convention "they" has become the standard gender-nuetral singular pronoun. I used to use "he or she" religiously, but its really quite awkward and outdated these days and widely accepted to just use "they"; if its good enough for Shakespeare its good enough for me. `</hat>`

Thanks for telling me :)


> > to break completely whenever the start line of any cell changes
> 
> I agree. As [Raymond Hettinger](https://www.youtube.com/watch?v=p33CVV29OG8) would say: "There must be a better way!"

Assuming the worst case: No cell is named / All the cells have the same name. 
The constraint is that I need to be able to type the `runcell` command in an explicit way. (I don't feel comfortable with looking up the current cursor position as there is no cursor in the editor while I am in the console)
 - If I number the cells, adding another cell would change all the numbers below.
 - I can reference by line number, which would also break if I add a cell above but at least it would fail fast with a clear error.
 - If I find a way of referencing the block with a number/string, I can follow the block around. This is what `Run Last Cell` and the breakpoint do. Maybe `id(block)`. The problem is that the id would only be valid while the document is open. If I restart spider, or reopen the file, the identifier is stale.
- Maybe a combination of the last two? 

The logic could be:
 - If I find the name at the given line number, run the cell
 - Else, look for a block whose id is `id_block`. If it exists, and if the cell it references has the correct name, run it.

The `block_id` would allow the line to be tracked if the line number changes. The `line_number` would allow the line to be tracked if the file is reopened.

impact27:
So maybe the `line_number` parameter could accept either an int or a tuple/dict, `(line_number, block_id)`

impact27:
I now use https://github.com/spyder-ide/spyder/pull/9082 to get an unique name, and then use this to identify the cell (eg. `Untitled Cell, #3`)

impact27:
Uses https://github.com/spyder-ide/spyder-kernels/pull/111
With this, `runcell` can be used in a script.

impact27:
I have added a function `cellcount(filename)` that returns the number of cells in `filename`. I also added the possibility to refer to the cells by index. This allows me to run a code like:
```python
for i in range(cellcount()-1): 
    runcell(i)
```
Which would run all the cells in the currently opened file (as `filename` is ommited), except for the last one.

impact27:
This is the next one. I think the failing tests are unrelated to my PR.

spyder/plugins/editor/widgets/editor.py
```diff
@@ -2577,25 +2597,39 @@ def advance_cell(self, reverse=False):
...
     def re_run_last_cell(self):
         """Run the previous cell again."""
-        last_cell = (self.get_current_editor()
-                     .get_last_cell_as_executable_code())
-        if not last_cell:
+        if self.last_cell_call is None:
             return
-        self._run_cell_text(*last_cell)
+        filename, cell_name = self.last_cell_call
+
```
ccordoba12: Remove this blank line.

spyder/plugins/ipythonconsole/comms/tests/test_console_kernel.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python3
```
ccordoba12: Please rename this file to be `test_runcell.py` instead of `test_console_kernel.py`.

spyder/plugins/ipythonconsole/comms/tests/test_console_kernel.py
```diff
@@ -0,0 +1,82 @@
...
+# -*- coding: utf-8 -*-
+"""
+Created on Mon Aug  5 16:43:33 2019
+
+@author: quentinpeter
+"""
```
ccordoba12: Please remove this header and use the one we use in all our other files instead.

spyder/plugins/ipythonconsole/comms/tests/test_console_kernel.py
```diff
@@ -0,0 +1,82 @@
...
+import pytest
+
+from spyder.plugins.ipythonconsole.comms.kernelcomm import KernelComm
+from spyder_kernels.py3compat import PY3, to_text_string
+
+
+TIMEOUT = 15
+
+
+def test_runcell(tmpdir):
```
ccordoba12: Is it possible for you add tests for `debugcell` here? I don't want to add such an important new functionality without being tested.

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -546,13 +546,6 @@ def norm(text):
                         before_prompt=True)
```
ccordoba12: Above this line there's a restriction that doesn't allow to call `runcell` when the console is `_reading` (i.e. our old way to detect if it's in debugging mode).

I think that no longer applies if `function='runcell'`, right? Of course, we can't call `debugcell` if the console is debugging, so the above message would also need to be updated.

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -454,18 +458,68 @@ def change_mpl_backend(self, command):
...
 
     # ---- Spyder-kernels methods -------------------------------------------
+    def get_editor(self, filename=None):
+        """Get editor for filename and set it as the current editor."""
+        editorstack = self.get_editorstack()
+        if editorstack is None:
+            return None
+
+        if not filename:
+            return editorstack.get_current_editor()
```
ccordoba12: Why this? I don't like to use the current editor as the default editor if not `filename` is passed. Instead, I think `filename` should be mandatory.

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -454,18 +458,68 @@ def change_mpl_backend(self, command):
...
+        editorstack.set_stack_index(index)
+        return editor
+
     def get_editorstack(self):
         """Get the current editorstack."""
         plugin = self.ipyclient.plugin
         if plugin.main.editor is not None:
             editor = plugin.main.editor
             return editor.get_current_editorstack()
+        raise RuntimeError('No editorstack found.')
```
ccordoba12: Why the need to raise this error here?

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -454,18 +458,68 @@ def change_mpl_backend(self, command):
...
+        """
         editorstack = self.get_editorstack()
-        if editorstack is not None:
-            editorstack.save()
+        editorstack.save()
+        editor = self.get_editor(filename)
+
+        if editor is None:
+            raise RuntimeError(
+                "File {} not open in the editor".format(filename))
```
ccordoba12: Do we really need to raise an error here? Why not simply show a QMessageBox?

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -454,18 +458,68 @@ def change_mpl_backend(self, command):
...
+
+    def handle_cell_count(self, filename=None):
+        """Get number of cells in file to loop."""
+        editorstack = self.get_editorstack()
+        editorstack.save()
+        editor = self.get_editor(filename)
+
+        if editor is None:
+            raise RuntimeError(
+                "File {} not open in the editor".format(filename))
```
ccordoba12: Again, do we need to raise an error here?

ccordoba12:
@impact27, please make sure that this PR also fixes #9963.

spyder/plugins/editor/widgets/base.py
```diff
@@ -757,31 +755,23 @@ def get_selection_as_executable_code(self):
...
+
         if text is not None:
             text = ls * line_from + text
         return text, block
 
-    def get_cell_as_executable_code(self):
-        """Return cell contents as executable code"""
-        return self.__exec_cell()
-
-    def get_last_cell_as_executable_code(self):
```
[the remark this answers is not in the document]
impact27: `re_run_last_cell` now uses a different approach, so it wan't needed anymore.

spyder/plugins/ipythonconsole/comms/tests/test_console_kernel.py
impact27: I'll do that soon

spyder/plugins/ipythonconsole/plugin.py
impact27: Well if the shell is _reading or _executing, we can't really start a new execution?

spyder/plugins/ipythonconsole/widgets/shell.py
impact27: Well if I call cell_count(), what I probably want to do is to count the number of cells in the current file.

impact27: So in general, this is probably what should happen.

spyder/plugins/ipythonconsole/widgets/shell.py
impact27: This error and the three errors below are caused when the kernel is calling functions, so possiblt it would want to do something else than show an error.

impact27: Like: save this file, and if it is not possible, I will execute it anyway.

spyder/plugins/ipythonconsole/widgets/shell.py
bcolsen: It is nice to not have to type whole filename in the console to run a cell. If I run `runcell('cellname')` in the console I would expect it to run the cell called 'cellname' from the current file.

However, I think this functionally might be better placed in `handle_run_cell()`. 

It would also be handy if `runcell()` with no aguments would run the current cell(where the cursor is/was. The cell that is highlighted) from the current editor.

spyder/plugins/ipythonconsole/plugin.py
ccordoba12: Let me put in another way: Are you able to run `runcell` by calling directly in `ipdb`? It'd be nice if users could do that as a way to execute multi-line statements while debugging.

spyder/plugins/ipythonconsole/widgets/shell.py
ccordoba12: > It is nice to not have to type whole filename in the console to run a cell

Ok, fair enough.

spyder/plugins/ipythonconsole/widgets/shell.py
ccordoba12: So are these errors handled in the kernel?

spyder/plugins/editor/widgets/editor.py
ccordoba12: This is still unaddressed.

spyder/plugins/ipythonconsole/comms/tests/test_runcell.py
```diff
@@ -0,0 +1,81 @@
+# -*- coding: utf-8 -*-
+# -----------------------------------------------------------------------------
```
ccordoba12: Please remove this line.

spyder/plugins/ipythonconsole/comms/tests/test_runcell.py
```diff
@@ -0,0 +1,81 @@
...
+# Copyright © 2018- Spyder Kernels Contributors
+# Licensed under the terms of the MIT License
+# (see spyder_kernels/__init__.py for details)
+# -----------------------------------------------------------------------------
```
ccordoba12: Also this line.

spyder/plugins/ipythonconsole/widgets/shell.py
impact27: They are displayed in the console

spyder/plugins/ipythonconsole/plugin.py
impact27: I enabled runfile and runcell while debugging

spyder/plugins/ipythonconsole/widgets/shell.py
impact27: The kernel will now request the filename from the frontend before calling this function.

spyder/plugins/ipythonconsole/widgets/shell.py
impact27: It makes more sense to display this error in the shell:
if I type `runcell(1, 'foo.py')`, I want an error in the shell if `'foo.py'` is not open.

spyder/plugins/ipythonconsole/widgets/shell.py
impact27: If the call fails, the other side must be notified. Raising an error is the standard way of notifying the other side of an error.

impact27: This error must be handled by the calling function. Here, if the calling function is `runcell`, it will simply print the error in the shell.

impact27:
Retriggering tests

spyder/plugins/ipythonconsole/plugin.py
ccordoba12: That's so cool!! Thanks a lot for doing that!

spyder/plugins/ipythonconsole/widgets/shell.py
ccordoba12: Ok, thanks. That makes more sense to me.

spyder/plugins/ipythonconsole/widgets/shell.py
ccordoba12: Ok, agreed.

spyder/plugins/ipythonconsole/widgets/shell.py
ccordoba12: Ok, no problem with this displaying this errors in the console.

ccordoba12:
@impact27, I'm fixing some failures in our test suite (unrelated to this) in PR #10044.

ccordoba12:
@impact27, please merge with master to fix the merge conflicts.

ccordoba12:
I have a general question about `debugcell`: what happens if a brekpoint is set inside the cell? Is that used by `debugcell` or not?

impact27:
> I have a general question about `debugcell`: what happens if a brekpoint is set inside the cell? Is that used by `debugcell` or not?

Yes it is

impact27:
spyder-kernels #131 deletes `__main__`, which causes problems. I fixed this in #134.

impact27:
Is there a way of including the console output when the tests fail in CI? that would make debugging console errors easier.

impact27:
Alternatively I can keep this test out, and open a new PR with just this test, if you would rather merge #112 before #134.

ccordoba12:
> Alternatively I can keep this test out

You mean the last test you added that tests unnamed cells? If so, then yes, please leave it out.

I want to merge spyder-ide/spyder-kernels#112 and this PR after that. Then I'd take a look at spyder-ide/spyder-kernels#134 along with the test for unnamed cells.

ccordoba12:
@impact27, this has now a small merge conflict with master. Please fix it so I can merge this one.

ccordoba12:
@impact27, please don't add more commits here. This is ready.

PR #9031, opened 2019-03-26
PR: Automatically unbind conflicting shortcuts when pressing okay in shortcut manager

### LINKED ISSUE #8293: More easily override conflicted shortcuts
During changing keyboard shortcuts of a function (let's say `func`), after pressing a new (conflicted) shortcut, the "OK" button is disabled, and a message showed that: "The new shortcut conflicts with - func1: short1". Then you have to cancel and find `func1`, and change it to `short2`. Great chances that `short2` will be conflicted to `func2`, then `func2` -> `func3` -> `func4` all the way in.  And go through `func4` -> `func3` -> `func2` -> `func1` all the way out. And finally set the shortcut you like in the first place if you still remember it.

Or you can just go to `func1` and clear the shortcut, and go back to set `func` in the first place.

Which leads to this feature request: an `override` button to override the shortcut. By override I mean if there's `func1: short1`, I want `short1` to `func`, a button to force `short1` to `func` and clear shortcut of `func1`

gepcel:
## Description of Changes

Adds the ability to automatically unbind a conflicting shortcut when clicking okay in the shortcut manager.

I'm not an expert but want to contribute. So could anyone help me with this?

### Issue(s) Resolved

Fixes #8293 

### Affirmation

I certify the above statement is true and correct: gepcel

### DIFF
--- a/spyder/preferences/shortcuts.py
+++ b/spyder/preferences/shortcuts.py
@@ -412,7 +412,7 @@ class ShortcutEditor(QDialog):
                 tip_body += ' - {0}: {1}<br>'.format(s.context, s.name)
             tip_body = tip_body[:-4]  # Removing last <br>
             tip_override = '<br>Press <b>OK</b> to unbind '
-            tip_override += 'it' if len(conflicts)==1 else 'them'
+            tip_override += 'it' if len(conflicts) == 1 else 'them'
             tip_override += ' and assign it to <b>{}</b>'.format(self.name)
             tip = template.format(tip_title, tip_body, tip_override)
             icon = get_std_icon('MessageBoxWarning')
--- a/spyder/preferences/shortcuts.py
+++ b/spyder/preferences/shortcuts.py
@@ -476,6 +476,7 @@ class ShortcutEditor(QDialog):
         self.accept()
 
     def accept_override(self):
+        """Unbind all conflicted shortcuts, and accept the new one"""
         conflicts = self.check_conflicts()
         if conflicts:
             for shortcut in conflicts:
--- a/spyder/preferences/shortcuts.py
+++ b/spyder/preferences/shortcuts.py
@@ -272,7 +272,7 @@ class ShortcutEditor(QDialog):
         layout_sequence.addLayout(newseq_btnbar, 3, 3)
         layout_sequence.addWidget(self.label_warning, 4, 2, 1, 2)
         layout_sequence.setColumnStretch(2, 100)
-        layout_sequence.setRowStretch(4, 100)
+        layout_sequence.setRowStretch(4, 120)
 
         layout = QVBoxLayout()
         layout.addLayout(layout_sequence)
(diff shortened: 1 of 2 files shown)


ccordoba12:
Thanks for your contribution @gepcel!

@goanpeca, please review this one.

gepcel:
I've noticed the test failure (before the pull actually). There should be four of them at least about:

```
assert shortcut_editor.warning == SEQUENCE_CONFLICT
assert not shortcut_editor.button_ok.isEnabled()
```

What should I do about them?

CAM-Gerlach:
> What should I do about them?

Fix them, to ensure that Ok is enabled and, as necessary, that the other shortcut is unbound. Let us know if you need more specific guidance on this.

gepcel:
@CAM-Gerlach ,  tests are fixed. Before it was just me being cautious, feels like need some permission when I changed something and needed to meddle the existed test to let pass.

I actually noticed the dialog height, and googled around a little bit, without a quick answer. I think to enlarge the default height by 3 lines more is not good enough. Since without any conflicts, we should not leave that much empty space, and with more than 1 conflicts, the warning still gets cut out (more than 3 lines). And also, with some shortcut having long context and name, the warning message may be too long to get wrapped into multiple lines. How to the height auto-expand? Could you please help me with this? These pulls recently are my first time dealing with qt.

Actually, while reviewing the code, please consider me as very new and lack of experience on the whole programming and git and pyqt thing, and this my first time digging into spyder source code. I'm not a programmer, only have experience writing scripts with python for data analysis. For example, after pressing a new shortcut, originally the `OK` event is dealted by `def accept(self)`, with a `@Slot` decorator that I don't understand, which is why I added a new func `def unbind_shortcut(self):`. I'm not confident it's appropriate.

ccordoba12:
> How to make the height auto-expand? Could you please help me with this?

@goanpeca, please guide @gepcel to solve this.

gepcel:
Now, the height of ShortcutEditor dialog will auto-grow when the warning message gets long. But it won't shrink back when the message gets short after.

gif

spyder/preferences/shortcuts.py
```diff
@@ -470,6 +475,13 @@ def unbind_shortcut(self):
         self._qsequences = []
         self.accept()
 
+    def accept_override(self):
```
ccordoba12: Please add a docstring to this method.

gepcel: Done.

goanpeca: Please add a period to the end

like `"""Unbind all conflicted shortcuts, and accept the new one."""`

https://www.cockos.com/licecap/

goanpeca:
@gepcel just one minor comment to fix. Could you make a small screencast of the functionality?

You can use https://www.cockos.com/licecap/

spyder/preferences/shortcuts.py
gepcel: I've added a period, but seems too late. Sorry.

gepcel:
> @gepcel just one minor comment to fix. Could you make a small screencast of the functionality?
> 
> You can use https://www.cockos.com/licecap/

Done

spyder/preferences/shortcuts.py
goanpeca: Opps xD

PR #9036, opened 2019-03-27
PR: Add Ctrl+Shift+G to switch to plots plugin

### LINKED ISSUE #8640: Assign a keyboard shortcut to switch to the Plots pane
I'm using spyder 4 from master. How can I assign a shortcut to switch to the plot widget?

gepcel:
As in issue #8640, now with this change, Ctrl+Shift+G can switch to plots.

The feature `adding shortcuts for "next figure" and "previous figure" of plots pane`  mentioned in this issue is still not done.  There are ''Ctrl+PgUp" and "Ctrl+PgDown" in the plots plugin, but they only work when the plugin gets focused. If someone can give me some guidance, maybe I can do something more. Or I may have to leave it to others.

Fixes #8640.

### DIFF
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -355,6 +355,7 @@ DEFAULTS = [
               '_/switch to variable_explorer': "Ctrl+Shift+V",
               '_/switch to find_in_files': "Ctrl+Shift+F",
               '_/switch to explorer': "Ctrl+Shift+X",
+              '_/switch to plots': "Ctrl+Shift+G",
               # -- In widgets/findreplace.py
               '_/find text': "Ctrl+F",
               '_/find next': "F3",


ok97465:
To focus on plots panes after pressing shortcut, I inserted the following method into ```class Plots(SpyderPluginWidget)``` of ```spyder/plugins/plots/plugin.py```.
```python
    def switch_to_plugin(self):
        """Switch to plots pane plugin by shortcut key.

        This method is called when pressing plugin's shortcut key
        """
        current_widget = self.current_widget()
        if current_widget and current_widget.figviewer.figcanvas.fig:
            current_widget.setFocus()
        super(SpyderPluginWidget, self).switch_to_plugin()
```
This workaround is just for me.
I do not know if the spyder team will agree on this approach.

ccordoba12:
@ok97465, why do you need that workaround?

ok97465:
> There are ''Ctrl+PgUp" and "Ctrl+PgDown" in the plots plugin, but they only work when the plugin gets focused.

The workaround is to use shortcuts of plots pane after switching to plots plugin by Ctrl+Shift+G.

ccordoba12:
@ok97465, I imagined that. @gepcel, do you understand how to include @ok97465 suggestion in this PR?

gepcel:
@ccordoba12 I understand and I can do that. But after "Ctrl+Shift+G", do people want the focus to be on the plots plugin or do we want the focus stays where ever it is?

We need to consider that some of the time (most of the time for me), people may just want to have a look at what's the plot looks like, and keep coding. So I need opinions from the developer team.

ccordoba12:
`Ctrl+Shift+G` gives focus to the Plots plugin, with or without @ok97465 suggestion. That suggestion only improves things so the user can be able to browse the list of generated plots with the keyboard (which I think it's really nice).

gepcel:
@ccordoba12 
It's not likely true from what I tested. Now with this pull only (without ok97465 's code), after `Ctrl+Shift+G`, the plot shows without the editor losing focus. 
screen

ccordoba12:
> It's not likely true from what I tested.

Ok, my mistake. That's probably the case because we don't define a default widget to give focus to in Plots when the plugin receives focus.

Then I'd say this is ready. @ok97465, you're suggestion is very good but you also need to create an option in Plots' options menu to allow users to decide if they want to give focus to the plugin or not after switching to it.

PR #9054, opened 2019-03-28
PR: Add an "Fit plots to window" option to the plots pane

gepcel:
## Description of Changes

While using the plotting plugin, I found it annoying to have to zoom in and out the image frequently when testing different sizes of image. Feels like an `Auto fit plotting` option is necessary.

So I tried adding one, like (default to False, aka, unchecked):
option

An example of the original behavior when dealing with different sizes of images (with the auto_fit_plotting option set to false):
false

Compared to the new behavior when set to true:
true

Example code:
```python
from matplotlib.pyplot import figure, text
def test_plot(w, h):
    figure(figsize=(w, h))
    text(.5, .5, f'figsize={w}, {h}', ha='center', fontsize=2.5*w)
    
test_plot(4, 3)
test_plot(15, 7)
test_plot(8, 12)
```

### Issue(s) Resolved

Fixes #

### Affirmation

I certify the above statement is true and correct: gepcel

### DIFF
--- a/spyder/plugins/plots/plugin.py
+++ b/spyder/plugins/plots/plugin.py
@@ -47,7 +47,8 @@ class Plots(SpyderPluginWidget):
     def get_settings(self):
         """Retrieve all Plots configuration settings."""
         return {name: self.get_option(name) for name in
-                ['mute_inline_plotting',  'show_plot_outline']}
+                ['mute_inline_plotting', 'show_plot_outline',
+                 'auto_fit_plotting']}
 
     # ---- Stack accesors
     def set_current_widget(self, fig_browser):
--- a/spyder/plugins/plots/widgets/figurebrowser.py
+++ b/spyder/plugins/plots/widgets/figurebrowser.py
@@ -238,8 +238,8 @@ class FigureBrowser(QWidget):
         self.show_plot_outline_action.setChecked(show_plot_outline)
 
         self.auto_fit_action = create_action(
-            self, _("Auto fit plotting"),
-            tip=_("Auto fit plotting in the widget."),
+            self, _("Fit plots to window"),
+            tip=_("Automatically fit plots to Plot pane size."),
             toggled=self.change_auto_fit_plotting
             )
         self.auto_fit_action.setChecked(auto_fit_plotting)
@@ -308,7 +308,7 @@ class FigureBrowser(QWidget):
         self.option_changed('show_plot_outline', state)
 
     def change_auto_fit_plotting(self, state):
-        """Change the auto_fit_plotting option and scale images"""
+        """Change the auto_fit_plotting option and scale images."""
         self.option_changed('auto_fit_plotting', state)
         self.figviewer.auto_fit_plotting = state
         self.figviewer.scale_image()
--- a/spyder/plugins/plots/widgets/figurebrowser.py
+++ b/spyder/plugins/plots/widgets/figurebrowser.py
@@ -490,11 +490,13 @@ class FigureViewer(QScrollArea):
             new_height = int(fheight * self._scalestep ** self._scalefactor)
 
         # Auto fit plotting
-        # Scale the image to fit figcanvas size while respect the ratio
+        # Scale the image to fit figviewer size while respect the ratio
         else:
             size = self.size()
-            width = size.width() - 15  # How to get the actual figcanvas size?
-            height = size.height() - 15
+            scrollbar_width = self.verticalScrollBar().sizeHint().width()
+            width = size.width() - scrollbar_width
+            scrollbar_height = self.horizontalScrollBar().sizeHint().height()
+            height = size.height() - scrollbar_height
             if (fwidth / fheight) > (width / height):
                 new_width = width
                 new_height = int(width / fwidth * fheight)
--- a/spyder/plugins/plots/widgets/figurebrowser.py
+++ b/spyder/plugins/plots/widgets/figurebrowser.py
@@ -481,8 +481,8 @@ class FigureViewer(QScrollArea):
 
     def scale_image(self):
         """Scale the image size."""
-        fwidth = self.figcanvas.fwidth
-        fheight = self.figcanvas.fheight
+        fwidth = self.figcanvas.fwidth * 1.0
+        fheight = self.figcanvas.fheight * 1.0
 
         # Don't auto fit plotting
         if not self.auto_fit_plotting:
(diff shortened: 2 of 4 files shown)


goanpeca:
@gepcel thanks for working on this. Looks pretty useful :-)

@jnsebgosselin could you chime in with your thoughts?

jnsebgosselin:
@goanpeca Sure, I'll take a look at it as soon as I can.

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -290,6 +307,12 @@ def show_fig_outline_in_viewer(self, state):
             self.figviewer.figcanvas.setStyleSheet("FigureCanvas{}")
         self.option_changed('show_plot_outline', state)
 
+    def change_auto_fit_plotting(self, state):
+        """Change the auto_fit_plotting option and scale images"""
```
CAM-Gerlach: Per PEP 257, all docstrings must end with periods.

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -456,10 +481,26 @@ def zoom_out(self):
...
+        # Don't auto fit plotting
+        if not self.auto_fit_plotting:
+            new_width = int(fwidth * self._scalestep ** self._scalefactor)
+            new_height = int(fheight * self._scalestep ** self._scalefactor)
+
+        # Auto fit plotting
+        # Scale the image to fit figcanvas size while respect the ratio
+        else:
+            size = self.size()
+            width = size.width() - 15  # How to get the actual figcanvas size?
```
CAM-Gerlach: @jnsebgosselin ?

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -229,7 +237,16 @@ def setup_option_actions(self, mute_inline_plotting, show_plot_outline):
             )
         self.show_plot_outline_action.setChecked(show_plot_outline)
 
-        self.actions = [self.mute_inline_action, self.show_plot_outline_action]
+        self.auto_fit_action = create_action(
+            self, _("Auto fit plotting"),
```
CAM-Gerlach: ``'Fit plots to window'``

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -229,7 +237,16 @@ def setup_option_actions(self, mute_inline_plotting, show_plot_outline):
...
+            tip=_("Auto fit plotting in the widget."),
```
CAM-Gerlach: ``'Automatically fit plots to Plot pane size'``

jnsebgosselin:
@goanpeca I'll let @CAM-Gerlach handle this, thanks.

gepcel:
@CAM-Gerlach I've updated some commits. Now:

- [x] Affirmed the DCO.
- [x] Added a test. I do need guidance on this, added some code for a start.
- [x] The option is default now.
- [ ]   Shouldn't we make it a persistent config setting so we don't have to set it every time
- [x] Changes based on code comments.

As to the "persistent config setting" suggestion, I'm not sure I understand you correctly, now:
1. After setting `Fit plots to windows`  Checked/Unchecked, the state will be remembered after closing and reopenning spyder, until the next change or until start spyder with `--safe-mode`. 
2. The `Fit plots to windows` state change is not `one plot specific`, it applies to all existing and all after comming plots in the

CAM-Gerlach:
> After setting Fit plots to windows Checked/Unchecked, the state will be remembered after closing and reopenning spyder, until the next change or until start spyder with --safe-mode.

This is what I meant, thanks. I thought it wasn't but it was probably because I was switching in and out of ``--safe-mode``, Windows and Linux and lost track.

CAM-Gerlach:
Other than your call-out about proper plot window width that perhaps @jnsebgosselin can provide insight on, looks like you addressed all my suggestions, thanks!

spyder/plugins/plots/widgets/figurebrowser.py
ccordoba12: @jnsebgosselin, I think we need your help to clarify this. Otherwise, @CAM-Gerlach has done a great job reviewing this PR.

CAM-Gerlach: There's a similar comment in the tests, too, that was added after I reviewed FYI.

gepcel: I think I have made this right, also with the test.

gepcel:
I think I have figured out the accurate plot size when scaling image.

jnsebgosselin:
I'll try to take a look at this PR a review it before the end of the week if it is possible to wait a little before merging this.

spyder/plugins/plots/plugin.py
```diff
@@ -47,7 +47,8 @@ def __init__(self, parent):
     def get_settings(self):
```
goanpeca: @ccordoba12 do you think it is worth having a method like this one as part of the plugin api?

ccordoba12:
@gepcel, tests are failing for Python 2.

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -481,8 +481,8 @@ def zoom_out(self):
 
     def scale_image(self):
         """Scale the image size."""
-        fwidth = self.figcanvas.fwidth
-        fheight = self.figcanvas.fheight
+        fwidth = self.figcanvas.fwidth * 1.0
```
ccordoba12: I think it should be more elegant to make use here `float` instead of multiplying by `1.0`. Also for all changes you made in this commit.

jnsebgosselin: @ccordoba12 Can we add `from __future__ import division` in this module instead to solve this for Python 2?

ccordoba12: Oh yeah, that's better! @gepcel, please do that instead.

gepcel: Good to learn about `from __future__ import division`

spyder/plugins/plots/plugin.py
ccordoba12: I don't think so.

goanpeca:
@ccordoba12 good to go?

ccordoba12:
We're waiting for @jnsebgosselin's review.

goanpeca:
@jnsebgosselin :-p ?

jnsebgosselin:
@goanpeca I've tested it locally and I agree with the way this feature is implementer as it is. I've also reviewed the code and everything LGTM. I apologize for taking so long to review this and thanks for waiting, this is much appreciated.

@gepcel Thanks for the awesome work on this, this is going to be very handy!

ccordoba12:
Thanks @jnsebgosselin for the review! Merging then.

PR #9057, opened 2019-03-28
PR: Fix completion popup dismissed after typing delimeters or operators

### LINKED ISSUE #8859: Completion popup menu isn't dismissed after typing a delimiter and causes unexpected behavior
## Problem Description

The completion popup dialog (list of possible completions) that appears with either automatic or manual completion isn't dismissed once the user types a Python delimiter (paren, bracket, space, etc).

Therefore, if the user manually types in the desired name, it stays up when they continue typing, and sometimes even continues to change the completion results depending on whatever they are typing next. Then, when they press ``Enter`` to move to the next line, it triggers completion instead, modifies the name they originally triggered completion on and moves their cursor to the end of that.

Some times, I didn't notice this happening, and as a result it broke my code by changing what I'd previously typed to something completely different, and was otherwise a quite serious nuisance to the point I would have considered turning Automatic Completions off if that were currently possible.

Note: ``|`` is the cursor position.

1. On a new line,  type ``np.ndarray.size == test|``
2. Press ``Enter``
3. Note the result is now ``np.ndarray.take| == test``

# Proposed Solution/Expected result

Cancel the current completion attempt if any Python [operator](https://docs.python.org/3/reference/lexical_analysis.html#operators) or [delimiter](https://docs.python.org/3/reference/lexical_analysis.html#delimiters) is typed.

## Versions

* Spyder version: 4.0.0.dev0 ``386403ed52``
* Python version: 3.7.2
* Qt version: 5.9.6
* PyQt5 version: 5.9.2

### Dependencies

```
sympy >=0.7.3     :  1.3 (OK)
```

steff456:
## Description of Changes

Dismiss the completion popup if a space is typed in the editor

Mar-28-2019 18-08-15

### Issue(s) Resolved

Fixes #8859 

### Affirmation

### DIFF
--- a/spyder/plugins/editor/tests/test_introspection.py
+++ b/spyder/plugins/editor/tests/test_introspection.py
@@ -100,6 +100,20 @@ def test_space_completion(setup_editor):
 
 
 @pytest.mark.slow
+def test_hide_widget_completion(setup_editor):
+    """Validate hiding completion widget after a delimeter or operator."""
+    editor, qtbot = setup_editor
+    code_editor = editor.get_focus_widget()
+    completion = code_editor.completion_widget
+
+    # Set cursor to start
+    code_editor.go_to_line(1)
+
+    qtbot.keyClicks(code_editor, 'from numpy [')
+    assert completion.isVisible() is False
+
+
+@pytest.mark.slow
 @pytest.mark.skipif(PY2, reason="Segfaults with other tests on Py2.")
 @pytest.mark.skipif(os.name == 'nt' and not PY2,
                     reason="Times out on AppVeyor and fails on PY3")
--- a/spyder/plugins/editor/tests/test_introspection.py
+++ b/spyder/plugins/editor/tests/test_introspection.py
@@ -12,6 +12,7 @@ import os.path as osp
 # Third party imports
 import pytest
 import pytestqt
+import random
 try:
     from unittest.mock import Mock
 except ImportError:
@@ -99,13 +100,16 @@ def test_space_completion(setup_editor):
     assert code_editor.toPlainText() == 'from numpy import\n'
 
 
-@pytest.mark.slow
+# @pytest.mark.slow
 def test_hide_widget_completion(setup_editor):
     """Validate hiding completion widget after a delimeter or operator."""
     editor, qtbot = setup_editor
     code_editor = editor.get_focus_widget()
     completion = code_editor.completion_widget
 
+    delimiters = ['(', ')', '[', ']', '{', '}', ',', ':', ';', '@', '=', '->',
+                  '+=', '-=', '*=', '/=', '//=', '%=', '@=', '&=', '|=', '^=',
+                  '>>=', '<<=', '**=']
     # Set cursor to start
     code_editor.go_to_line(1)
 
@@ -121,11 +125,14 @@ def test_hide_widget_completion(setup_editor):
     # Check the completion widget is visible
     assert completion.isHidden() is False
 
-    # Write a delimeter on the code editor
-    qtbot.keyClicks(code_editor, '(')
+    # Write a random delimeter on the code editor
+    delimeter = random.choice(delimiters)
+    qtbot.keyClicks(code_editor, delimeter)
 
     # Check the completion widget is not visible
     assert completion.isHidden() is True
+
+    qtbot.keyPress(code_editor, Qt.Key_Enter)
 
 
 @pytest.mark.slow
--- a/spyder/plugins/editor/tests/test_introspection.py
+++ b/spyder/plugins/editor/tests/test_introspection.py
@@ -109,8 +109,23 @@ def test_hide_widget_completion(setup_editor):
     # Set cursor to start
     code_editor.go_to_line(1)
 
-    qtbot.keyClicks(code_editor, 'from numpy [')
-    assert completion.isVisible() is False
+    # Complete from numpy --> from numpy import
+    qtbot.keyClicks(code_editor, 'from numpy ')
+    qtbot.wait(20000)
+
+    # press tab and get completions
+    with qtbot.waitSignal(completion.sig_show_completions,
+                          timeout=10000) as sig:
+        qtbot.keyPress(code_editor, Qt.Key_Tab)
+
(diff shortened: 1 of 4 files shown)


ccordoba12:
@steff456, does this also cover the other delimiters mentioned by @CAM-Gerlach on issue #8859? In other words, do we hide the completion widget after users press `(`, `)`, `{`, `}`, `[` or `]`?

CAM-Gerlach:
@ccordoba12 At least in my testing, it does not.

ccordoba12:
Ok, @steff456, we need to hide the completion widget in those cases too. Please address that in this PR.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -2600,6 +2600,8 @@ def __has_colon_not_in_brackets(self, text):
     def autoinsert_colons(self):
         """Decide if we want to autoinsert colons"""
         bracket_ext = self.editor_extensions.get(CloseBracketsExtension)
+        logger.debug('holowis')
```
ccordoba12: Is this line necessary?

ccordoba12:
@CAM-Gerlach, could check again this PR? It should be working now.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -2775,6 +2776,13 @@ def keyPressEvent(self, event):
         has_selection = self.has_selected_text()
         ctrl = event.modifiers() & Qt.ControlModifier
         shift = event.modifiers() & Qt.ShiftModifier
+        operators = {'+', '-', '*', '**', '/', '//', '%', '@', '<<', '>>',
+                     '&', '|', '^', '~', '<', '>', '<=', '>=', '==', '!='}
+        delimeters = {',', ':', ';', '@', '=', '->', '+=', '-=', '*=', '/=',
+                      '//=', '%=', '@=', '&=', '|=', '^=', '>>=', '<<=', '**='}
+        logger.debug(text)
```
ccordoba12: @steff456, please remove this line if it's not needed.

ccordoba12:
> Any chance we can get a test for this?

Tests for code completion take a lot of time to run so I think we can merge this one without one for now.

ccordoba12:
@steff456, on a second thought, please add a simple test for this. You can add it here:

https://github.com/spyder-ide/spyder/blob/master/spyder/plugins/editor/tests/test_introspection.py

and you can test that after writing a delimiter (e.g. `[`) the completion widget is hidden.

Thanks!

spyder/plugins/editor/tests/test_introspection.py
```diff
@@ -99,6 +99,20 @@ def test_space_completion(setup_editor):
...
+def test_hide_widget_completion(setup_editor):
+    """Validate hiding completion widget after a delimeter or operator."""
+    editor, qtbot = setup_editor
+    code_editor = editor.get_focus_widget()
+    completion = code_editor.completion_widget
+
+    # Set cursor to start
+    code_editor.go_to_line(1)
+
+    qtbot.keyClicks(code_editor, 'from numpy [')
```
ccordoba12: @steff456, this is not testing things correctly. First we need to make the completion widget appear, then write `[` and finally verify that the widget is not visible.

steff456: I'm going to check it!

spyder/plugins/editor/tests/test_introspection.py
```diff
@@ -109,8 +109,23 @@ def test_hide_widget_completion(setup_editor):
...
+        qtbot.keyPress(code_editor, Qt.Key_Tab)
+
+    # Check the completion widget is visible
+    assert completion.isHidden() is False
+
+    # Write a delimeter on the code editor
+    qtbot.keyClicks(code_editor, '(')
+
+    # Check the completion widget is not visible
+    assert completion.isHidden() is True
```
goanpeca: @steff456 could you rewrite this test so that the test is run for all the chars?

Perhaps using a `for` and iterating over `self.BRACKETS_CHAR` so that we are indeed checking for all delimiters?

ccordoba12: Introspection tests take a lot of time to run, so I don't think rerunning this test for all bracket chars would be necessary.

goanpeca: At least then it should randomly pick the char from the list so we are randomly testing all the options, otherwise this test is broken and incomplete.

goanpeca: Something like

```python
import random
delimeter = random.choice(BRACKETS_CHAR)
```

ccordoba12: Ok, that's better. @steff456, please implement Gonzalo's suggestion of picking up a random bracket char and running the test with it.

steff456: Ok, I will add all delimiters in the test

ccordoba12: There's no need to add all delimiters, just one selected at random.

spyder/plugins/editor/tests/test_introspection.py
```diff
@@ -99,13 +100,16 @@ def test_space_completion(setup_editor):
...
 
-@pytest.mark.slow
+# @pytest.mark.slow
 def test_hide_widget_completion(setup_editor):
     """Validate hiding completion widget after a delimeter or operator."""
     editor, qtbot = setup_editor
     code_editor = editor.get_focus_widget()
     completion = code_editor.completion_widget
 
+    delimiters = ['(', ')', '[', ']', '{', '}', ',', ':', ';', '@', '=', '->',
```
goanpeca: Question @steff456: Are this defined on some widget so that we can import the widget, instead of rewriting the delimiters here?

Otherwise it looks great :-p

steff456: No, they are defined as a set inside keyPressEvent function in the code editor.

goanpeca: Cool! Thanks @steff456 :-)

PR #9123, opened 2019-04-06
PR: Add preference to show/hide cmd windows generated by the subprocess module

### LINKED ISSUE #9113: Opening window in a subprocess from Spyder
I asked the question there:
https://stackoverflow.com/questions/55511541/subprocess-in-spyder-matplotlib-and-prints-do-not-work

Running this on a regular console works well, but from Spyder the window does not show up. Prints neither btw.

main.py:

```
import subprocess

subprocess.Popen("test.py", shell=True)

```
test.py:

```

plt.plot([1,3,2])
plt.show()

```
I am using Spyder 3.3.3, python 3.7.3 on windows 10

dalthviz:
## Description of Changes

imagen

### Issue(s) Resolved

Fixes #9113 

### Affirmation


spyder/plugins/ipythonconsole.py
```diff
@@ -621,6 +621,14 @@ def setup_page(self):
         prompts_layout.addLayout(prompts_g_layout)
         prompts_group.setLayout(prompts_layout)
 
+        # Windows adjustments
+        windows_group = QGroupBox(_("Windows adjustments"))
+        show_subprocess_window = newcb(_("Show subprocess windows"),
```
ccordoba12: I think it'b be better to phrase this like

> Show command line windows generated by the `subprocess` module.

@CAM-Gerlach, do you have another suggestion?

spyder/config/main.py
```diff
@@ -170,7 +170,8 @@
               'in_prompt': '',
               'out_prompt': '',
               'show_elapsed_time': False,
-              'ask_before_restart': True
+              'ask_before_restart': True,
+              'subprocess_windows': False
```
goanpeca: Should this be `True` by default? Users using `shell=True` expect this, no @ccordoba12 ?

ccordoba12: This is False because there are libraries like Pyomo that generate a lot of Command Prompts while running, and that's extremely annoying for Windows users.

goanpeca: That could be a great comment :-)

``` 
#This is False because there are libraries like Pyomo that generate a lot of Command Prompts while running, and that's extremely annoying for Windows users.
```

ccordoba12: Ok, that's true :+1:. @dalthviz, please add my comment above that option, and also change its name to `show_cmd_windows`.

spyder/plugins/ipythonconsole.py
CAM-Gerlach: @ccordoba12 I agree with your suggestion, with the slight tweak to "command line *output* windows" to make it clear what these windows would be showing (that would be of interest to the user).

Also, I would consider inverting the name to "Hide" rather than "Show" (and therefore making it checked by default, so the default still hides the windows), since this is what Spyder is actually doing relative to a standard Python interpreter particularly from the user's perspective.

ccordoba12: I agree, thanks @CAM-Gerlach!

spyder/config/main.py
ccordoba12: So @dalthviz, please change this option to `'hide_cmd_windows': True` instead.

spyder/utils/ipython/kernelspec.py
```diff
@@ -117,7 +117,8 @@ def env(self):
             'SPY_JEDI_O': CONF.get('ipython_console', 'jedi_completer'),
             'SPY_SYMPY_O': CONF.get('ipython_console', 'symbolic_math'),
             'SPY_RUN_CYTHON': self.is_cython,
-            'SPY_TESTING': running_under_pytest() or SAFE_MODE
+            'SPY_TESTING': running_under_pytest() or SAFE_MODE,
+            'SPY_SHOW_CMD': CONF.get('ipython_console', 'subprocess_windows')
```
ccordoba12: This needs to be now `SPY_HIDE_CMD` instead.

spyder/plugins/ipythonconsole.py
```diff
@@ -621,6 +621,16 @@ def setup_page(self):
         prompts_layout.addLayout(prompts_g_layout)
         prompts_group.setLayout(prompts_layout)
 
+        # Windows adjustments
+        windows_group = QGroupBox(_("Windows adjustments"))
+        show_subprocess_window = newcb(
```
ccordoba12: Please change `show_subprocess_window` to `hide_cmd_windows`

Issue #7724, opened 2018-08-20
Finish LSP integration


ccordoba12:
- [ ] Add new requirements to setup.py -> @ccordoba12
- [ ] Remove Jedi and Rope from our dependencies and add new ones -> @ccordoba12
- [ ] Remove old architecture -> @ccordoba12
- [ ] Migrate tests to the new architecture -> @ccordoba12
- [ ] Implement/add missing features from LSP -> @andfoy
- [ ] Make code completion work on the fly -> @andfoy 
- [ ] Reorganize or remove LSP preferences page.

@andfoy, please don't start to work on this. I'll merge `split-plugins` first and then we can continue with this.

Also, please make a separate PR for each new feature you add and add new things that need to be done here.

ccordoba12:
@andfoy, I merged split-plugins last night, so you can start working in your assigned tasks now.

ccordoba12:
We decided with @andfoy that some LSP features will be implemented for beta2 and the rest in beta3.

stonebig:
hi all. In current master, I fail to make code completion working on the left editor screen. Is it normal ?

ccordoba12:
Yes, it is. We need to fix some things for Windows yet.

stonebig:
Isn't it working now  on Wndows ? I just tried with laster master.

ccordoba12:
It should be fixed now after PR #8600.

PR #9140, opened 2019-04-10
PR: Use calltip widget to display calltips and tooltips

goanpeca:
## Description of Changes

### Issue(s) Resolved

Related to #7724

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -886,6 +886,7 @@ class CodeEditor(TextEditBaseWidget):
                     parameter,
                     parameter_documentation,
                     color='#999999',
+                    is_python=self.is_python()
                 )
         except Exception:
             self.log_lsp_handle_errors("Error when processing signature")
@@ -1734,6 +1735,10 @@ class CodeEditor(TextEditBaseWidget):
         self.setUpdatesEnabled(True)
         self.linenumberarea.update()
         self.classfuncdropdown.update()
+
+    def hide_tooltip(self):
+        """"""
+        self.tooltip_widget.hide()
 
     def show_code_analysis_results(self, line_number, block_data):
         """Show warning/error messages."""
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -863,8 +863,9 @@ class CodeEditor(TextEditBaseWidget):
                 line, _ = self.get_cursor_line_column()
                 active_parameter_idx = signature['activeParameter']
                 signature = signature['signatures']
-                func_doc = signature['documentation'].replace(u'\xa0', ' ')
-                print(func_doc)
+
+                func_doc = signature['documentation']
+                func_doc = func_doc.replace(u'\xa0', ' ')
                 func_signature = signature['label']
                 parameters = signature['parameters']
                 parameter = parameters[active_parameter_idx]
@@ -890,8 +891,7 @@ class CodeEditor(TextEditBaseWidget):
                 tooltip_text = "{0}{1}".format(parameter_str, func_doc)
                 self.show_calltip(
                     title, tooltip_text, color='#999999', signature=True)
-        except Exception as e:
-            print(e)
+        except Exception:
             self.log_lsp_handle_errors("Error when processing signature")
 
     # ------------- LSP: Hover ---------------------------------------
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -865,6 +865,8 @@ class CodeEditor(TextEditBaseWidget):
                 documentation = signature_data['documentation']
 
                 if PY2:
+                    # The language server is returning encoded text with
+                    # spaces defined as `\xa0`
                     documentation = documentation.replace(u'\xa0', ' ')
 
                 parameter_idx = signature_params['activeParameter']
@@ -1733,7 +1735,14 @@ class CodeEditor(TextEditBaseWidget):
         self.classfuncdropdown.update()
 
     def hide_tooltip(self):
-        """"""
+        """
+        Hide the tooltip widget.
+
+        The tooltip widget is a special QLabel that looks like a tooltip,
+        this method is here so it can be hidden as necessary. For example,
+        when the user leaves the Linenumber area when hovering over lint
+        warnings and errors.
+        """
         self.tooltip_widget.hide()
 
     def show_code_analysis_results(self, line_number, block_data):
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -766,7 +766,7 @@ class CodeEditor(TextEditBaseWidget):
         self.auto_completion_characters = (
             completion_options['triggerCharacters'])
         self.signature_completion_characters = (
-            signature_options['triggerCharacters'])
+            signature_options['triggerCharacters'] + ['='])  # FIXME:
         self.go_to_definition_enabled = config['definitionProvider']
         self.find_references_enabled = config['referencesProvider']
         self.highlight_enabled = config['documentHighlightProvider']
@@ -857,6 +857,9 @@ class CodeEditor(TextEditBaseWidget):
         """Handle signature response."""
         try:
             signature_params = params['params']
+            # import json
+            # print(json.dumps(signature_params, sort_keys=True,
+            #                  indent=4, separators=(',', ': ')))
 
             if (signature_params is not None and
                     'activeParameter' in signature_params):
@@ -2958,6 +2961,7 @@ class CodeEditor(TextEditBaseWidget):
                     self.do_completion(automatic=True)
         elif (text != '(' and text in self.signature_completion_characters and
                 not self.has_selected_text()):
+            print(text)
             self.insert_text(text)
             self.request_signature()
         elif key == Qt.Key_Colon and not has_selection \
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -1745,7 +1745,7 @@ class CodeEditor(TextEditBaseWidget):
         self.show_tooltip(
             _("Code analysis"),
             msglist,
-            color='#129625', 
+            color='#129625',
             at_line=line_number,
         )
         self.highlight_line_warning(block_data)
@@ -2961,7 +2961,6 @@ class CodeEditor(TextEditBaseWidget):
                     self.do_completion(automatic=True)
         elif (text != '(' and text in self.signature_completion_characters and
                 not self.has_selected_text()):
-            print(text)
             self.insert_text(text)
             self.request_signature()
         elif key == Qt.Key_Colon and not has_selection \
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -856,41 +856,34 @@ class CodeEditor(TextEditBaseWidget):
     def process_signatures(self, params):
         """Handle signature response."""
         try:
-            signature = params['params']
-            if (signature is not None and
-                    'activeParameter' in signature):
+            signature_params = params['params']
+
+            if (signature_params is not None and
+                    'activeParameter' in signature_params):
                 self.sig_signature_invoked.emit()
-                line, _ = self.get_cursor_line_column()
-                active_parameter_idx = signature['activeParameter']
-                signature = signature['signatures']
 
-                func_doc = signature['documentation']
-                func_doc = func_doc.replace(u'\xa0', ' ')
-                func_signature = signature['label']
-                parameters = signature['parameters']
-                parameter = parameters[active_parameter_idx]
+                signature_data = signature_params['signatures']
+                documentation = signature_data['documentation']
 
-                font = self.font()
-                size = font.pointSize()
-                family = font.family()
+                if PY2:
+                    documentation = documentation.replace(u'\xa0', ' ')
 
-                parameter_str = ''
-                color_change_str = ('<span style=\'font-family: "{0}"; '
-                                    'font-size: {1}pt; color: {2}\'>')
-                if (parameter['documentation'] is not None and
-                        len(parameter['documentation']) > 0):
-                    parameter_fmt = ('{0}<b>{1}</b></span>: {2}\n')
-                    parameter_str = parameter_fmt.format(
-                        color_change_str.format(family, size, '#daa520'),
-                        parameter['label'], parameter['documentation'])
+                parameter_idx = signature_params['activeParameter']
+                parameters = signature_data['parameters']
+                parameter_data = parameters[parameter_idx]
 
-                title = func_signature.replace(
-                    parameter['label'], '{0}{1}</span>'.format(
-                        color_change_str.format(family, size, '#daa520'),
-                        parameter['label']))
-                tooltip_text = "{0}{1}".format(parameter_str, func_doc)
+                signature = signature_data['label']
+                parameter = parameter_data['label']
+                parameter_documentation = parameter_data['documentation']
+
+                # This method is part of spyder/widgets/mixins
                 self.show_calltip(
-                    title, tooltip_text, color='#999999', signature=True)
+                    signature,
+                    documentation,
+                    parameter,
+                    parameter_documentation,
+                    color='#999999',
+                )
         except Exception:
             self.log_lsp_handle_errors("Error when processing signature")
 
@@ -911,8 +904,7 @@ class CodeEditor(TextEditBaseWidget):
         try:
             text = contents['params']
             self.sig_display_signature.emit(text)
-            self.show_calltip(_("Hint"), text)
-            # QTimer.singleShot(20000, lambda: QToolTip.hideText())
+            self.show_tooltip(_("Hint"), text)
         except Exception:
             self.log_lsp_handle_errors("Error when processing hover")
 
@@ -924,12 +916,16 @@ class CodeEditor(TextEditBaseWidget):
         if (not self.go_to_definition_enabled or
                 self.in_comment_or_string()):
             return
+
         if cursor is None:
             cursor = self.textCursor()
+
         text = to_text_string(cursor.selectedText())
+
         if len(text) == 0:
             cursor.select(QTextCursor.WordUnderCursor)
             text = to_text_string(cursor.selectedText())
+
         if text is not None:
             line, column = self.get_cursor_line_column()
             params = {
@@ -1737,12 +1733,18 @@ class CodeEditor(TextEditBaseWidget):
         self.classfuncdropdown.update()
 
     def show_code_analysis_results(self, line_number, block_data):
-        """Show warning/error messages"""
+        """Show warning/error messages."""
         code_analysis = block_data.code_analysis
-        msglist = ['{0} [{1}]: {2}'.format(
-            src, code, msg) for src, code, _sev, msg in code_analysis]
-        self.show_calltip(_("Code analysis"), msglist,
-                          color='#129625', at_line=line_number)
+        template = '{0} [{1}]: {2}'
+        msglist = [template.format(src, code, msg) for src, code, _sev, msg
+                   in code_analysis]
+
+        self.show_tooltip(
+            _("Code analysis"),
+            msglist,
+            color='#129625', 
+            at_line=line_number,
+        )
         self.highlight_line_warning(block_data)
 
     def highlight_line_warning(self, block_data):
@@ -1823,8 +1825,13 @@ class CodeEditor(TextEditBaseWidget):
                 break
         line_number = block.blockNumber()+1
         self.go_to_line(line_number)
-        self.show_calltip(_("To do"), data.todo,
-                          color='#3096FC', at_line=line_number)
+        self.show_tooltip(
+            _("To do"),
+            data.todo,
+            color='#3096FC',
+            at_line=line_number,
+        )
+
         return self.get_position('cursor')
 
     def process_todo(self, todo_results):
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -857,10 +857,6 @@ class CodeEditor(TextEditBaseWidget):
         """Handle signature response."""
         try:
             signature_params = params['params']
-            # import json
-            # print(json.dumps(signature_params, sort_keys=True,
-            #                  indent=4, separators=(',', ': ')))
-
             if (signature_params is not None and
                     'activeParameter' in signature_params):
                 self.sig_signature_invoked.emit()
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -888,7 +888,7 @@ class CodeEditor(TextEditBaseWidget):
                         parameter['label']))
                 tooltip_text = "{0}{1}".format(parameter_str, func_doc)
                 self.show_calltip(
-                    title, tooltip_text, color='#999999')
+                    title, tooltip_text, color='#999999', signature=True)
         except Exception:
             self.log_lsp_handle_errors("Error when processing signature")
 
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -863,7 +863,8 @@ class CodeEditor(TextEditBaseWidget):
                 line, _ = self.get_cursor_line_column()
                 active_parameter_idx = signature['activeParameter']
                 signature = signature['signatures']
-                func_doc = signature['documentation']
+                func_doc = signature['documentation'].replace(u'\xa0', ' ')
+                print(func_doc)
                 func_signature = signature['label']
                 parameters = signature['parameters']
                 parameter = parameters[active_parameter_idx]
@@ -889,7 +890,8 @@ class CodeEditor(TextEditBaseWidget):
                 tooltip_text = "{0}{1}".format(parameter_str, func_doc)
                 self.show_calltip(
                     title, tooltip_text, color='#999999', signature=True)
-        except Exception:
+        except Exception as e:
+            print(e)
             self.log_lsp_handle_errors("Error when processing signature")
 
     # ------------- LSP: Hover ---------------------------------------
--- a/spyder/plugins/ipythonconsole/widgets/help.py
+++ b/spyder/plugins/ipythonconsole/widgets/help.py
@@ -125,5 +125,4 @@ class HelpWidget(RichJupyterWidget):
             if content.get('status') == 'ok' and content.get('found', False):
                 signature = self.get_signature(content)
                 if signature:
-                    self._control.show_calltip(_("Arguments"), signature,
-                                               signature=True, color='#2D62FF')
+                    self._control.show_calltip(signature, color='#2D62FF')
--- a/spyder/plugins/ipythonconsole/widgets/help.py
+++ b/spyder/plugins/ipythonconsole/widgets/help.py
@@ -60,6 +60,10 @@ class HelpWidget(RichJupyterWidget):
                 signature = name + argspec
             else:
                 signature = getsignaturefromtext(text, name)
+
+            # Remove docstring for uniformity with editor
+            signature = signature.split('Docstring:')[0]
+
             return signature
         else:
             return ''
@@ -125,4 +129,6 @@ class HelpWidget(RichJupyterWidget):
             if content.get('status') == 'ok' and content.get('found', False):
                 signature = self.get_signature(content)
                 if signature:
-                    self._control.show_calltip(signature, color='#2D62FF')
+                    # TODO: Pass the language from the Console to the calltip
+                    self._control.show_calltip(signature, color='#999999',
+                                               is_python=True)
--- a/spyder/widgets/calltip.py
+++ b/spyder/widgets/calltip.py
@@ -46,7 +46,6 @@ class ToolTipWidget(QLabel):
         self.tip = None
 
         # Setup
-        # self.setFont(text_edit.document().defaultFont())
         self.setForegroundRole(QPalette.ToolTipText)
         self.setBackgroundRole(QPalette.ToolTipBase)
         self.setPalette(QToolTip.palette())
@@ -73,7 +72,8 @@ class ToolTipWidget(QLabel):
     # --- 'ToolTipWidget' interface
     # ------------------------------------------------------------------------
     def show_tip(self, point, tip):
-        """ Attempts to show the specified tip at the current cursor location.
+        """
+        Attempts to show the specified tip at the current cursor location.
         """
         # Don't attempt to show it if it's already visible and the text
         # to be displayed is the same as the one displayed before.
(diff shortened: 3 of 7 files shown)


goanpeca:
@ccordoba12 is the expected return of the LSP a plain text or HTML ?

It seems the signature format code expects HTML, but the LSP is returning plain text.

Could you confirm?

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -859,7 +859,8 @@ def process_signatures(self, params):
                 line, _ = self.get_cursor_line_column()
                 active_parameter_idx = signature['activeParameter']
                 signature = signature['signatures']
-                func_doc = signature['documentation']
+                func_doc = signature['documentation'].replace(u'\xa0', ' ')
```
goanpeca: This was generating an error on Python 2

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -884,8 +885,9 @@ def process_signatures(self, params):
                         parameter['label']))
                 tooltip_text = "{0}{1}".format(parameter_str, func_doc)
                 self.show_calltip(
-                    title, tooltip_text, color='#999999')
-        except Exception:
+                    title, tooltip_text, color='#999999', signature=True)
+        except Exception as e:
+            print(e)
```
goanpeca: Why we do not use the error as part of the logging?

spyder/plugins/editor/widgets/codeeditor.py
ccordoba12: What kind of error?

spyder/plugins/editor/widgets/codeeditor.py
ccordoba12: See the definition of `log_lsp_handle_errors` above.

spyder/plugins/editor/widgets/codeeditor.py
goanpeca: cannot decode `ascii` '\xa0'` (which are spaces)

spyder/plugins/editor/widgets/codeeditor.py
goanpeca: But it seems on Py2 we completely loose the error, cause I did not know what was happening until I printed the actual exception. All I was getting was an error on the logger and not the actual error.

ccordoba12: Yeah, that needs to be fixed (I didn't test things in Python 2). Please open a new PR about it.

ccordoba12:
> It seems the signature format code expects HTML, but the LSP is returning plain text.
Could you confirm?

Yes, CalltipWidget expects HTML, but you can format the plain text signature from the LSP (the old architecture also returned plain text), with the method `_format_signature`:

https://github.com/spyder-ide/spyder/blob/master/spyder/widgets/mixins.py

goanpeca:
There seems to be a bit of mixed logic in these calls a lot of string style formating that take place some in the code editor, and some in the mixin. These changes should probably only happen at the mixin level to decouple the logic a bit.

But before I modify them @ccordoba12 :

How are these call tips supposed to look? Like these:

Screen Shot 2019-04-14 at 09 46 42

Screen Shot 2019-04-14 at 09 46 53

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -888,7 +890,7 @@ def process_signatures(self, params):
                         parameter['label']))
```
goanpeca: All the code above is already performing styling, this should probably be moved to the mixin.

Thoughts?

spyder/widgets/mixins.py
```diff
@@ -68,9 +68,12 @@ def calculate_real_position(self, point):
...
-        name = text.split('(')[0]
-        rows = textwrap.wrap(text, width=50,
-                             subsequent_indent=' '*(len(name)+1))
+        # This code was assuming the text included the signature, which is
+        # no longer the case
+        # name = text.split('(')[0]
+        # rows = textwrap.wrap(text, width=50,
+        #                      subsequent_indent=' '*(len(name)+1))
+        rows = text.split('\n')
         for r in rows:
```
goanpeca: This was assuming the signature was included in the text which is no longer the case. But probably all the styling that is being made on the code editor, should be made here.

Thougths?

spyder/plugins/editor/widgets/codeeditor.py
ccordoba12: Agreed. That was added by @andfoy to (temporarily) make things work with tooltips, but there's no need to have it here.

spyder/widgets/mixins.py
ccordoba12: I think this part could be important to maintain it

> rows = textwrap.wrap(text, width=50, subsequent_indent=' '*(len(name)+1))

It wraps the signature to make it 50 chars long, which avoids displaying a huge calltip for objects with a lot of arguments. You can try with `subprocess.Popen` in Python 3 to see what I mean.

ccordoba12:
> How are these call tips supposed to look? Like these:

I think so. Do they look the same in the IPython console? The idea is to unify the look and feel of both editor and console calltips.

goanpeca:
@ccordoba12 this is mixing the logic that was before highlighting special characters, besides the current arg (although the current logic is brittle)

Screen Shot 2019-04-15 at 20 27 02

Now, how should we crop the docstring?

goanpeca:
Also, the calltip widget should be scrollable, so the full docs can be seen?

VSCode does this:

Screen Shot 2019-04-15 at 20 51 05

goanpeca:
Screen Shot 2019-04-15 at 22 12 05

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -766,7 +766,7 @@ def parse_lsp_config(self, config):
         self.auto_completion_characters = (
             completion_options['triggerCharacters'])
         self.signature_completion_characters = (
-            signature_options['triggerCharacters'])
+            signature_options['triggerCharacters'] + ['='])  # FIXME:
```
goanpeca: This is needed for this to work properly, but I dont know if this should be part of the dedault settings or something I should add here ?

spyder/plugins/ipythonconsole/widgets/help.py
```diff
@@ -125,5 +125,4 @@ def _handle_inspect_reply(self, rep):
             if content.get('status') == 'ok' and content.get('found', False):
                 signature = self.get_signature(content)
                 if signature:
-                    self._control.show_calltip(_("Arguments"), signature,
-                                               signature=True, color='#2D62FF')
+                    self._control.show_calltip(signature, color='#2D62FF')
```
goanpeca: This is what is being displayed on the console, but it seems to come from another source (not the LSP)

ccordoba12:
> this is mixing the logic that was before highlighting special characters, besides the current arg

I prefer this version because it makes easier to discern the list of kwargs with the red commas and equal signs.

What do you think? Perhaps we should only leave the red commas?

> (although the current logic is brittle)

What do you mean by this? Can we improve the situation?

ccordoba12:
> Also, the calltip widget should be scrollable, so the full docs can be seen?

> It is messy, but the whole docs can be explored. We could argue that the help pane is or this... but, it should be possible to see complete calltips with scrolling

There's no problem in making the calltip widget scrollable, but I think it'd be nicer if we added a link at the end of calltip with a text like `Press here for addtional help`. Then, by pressing on that line, the entire docstring would be displayed in the Help pane.

I got this idea from RStudio's calltips:

Selección_033

spyder/plugins/ipythonconsole/widgets/help.py
ccordoba12: Yes, it comes from the kernel. The important thing is that both calltips (Editor and Console ones) have the same look and feel.

spyder/plugins/editor/widgets/codeeditor.py
ccordoba12: I don't understand very well where we need to set this. @andfoy, could you give us a hand here? Thanks.

spyder/widgets/mixins.py
```diff
@@ -44,6 +44,11 @@
 
 class BaseEditMixin(object):
 
+    _PARAMETER_HIGHLIGHT_COLOR = '#DAA520'
+    _DEFAULT_TITLE_COLOR = '#2D62FF'
+    _CHAR_HIGHLIGHT_COLOR = 'red'
+    _DEFAULT_TEXT_COLOR = '#999999'
```
ccordoba12: I think some of these colors need to be different according to the current interface theme (dark/light).

For example, the current background color in the screenshots you uploaded is yellow but I think it should be the same (or perhaps a bit lighter) than the one used in the Editor.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1736,12 +1737,18 @@ def process_code_analysis(self, check_results):
...
+        template = '{0} [{1}]: {2}'
+        msglist = [template.format(src, code, msg) for src, code, _sev, msg
+                   in code_analysis]
+
+        self.show_tooltip(
+            _("Code analysis"),
+            msglist,
+            color='#129625',
+            at_line=line_number,
+        )
```
ccordoba12: When are these warnings hidden? The idea is that they should be hidden without user interaction. Or in other words, they should behave like a regular tooltip (i.e hidden after the mouse is out of the widget).

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1822,8 +1829,13 @@ def go_to_next_todo(self):
...
         line_number = block.blockNumber()+1
         self.go_to_line(line_number)
-        self.show_calltip(_("To do"), data.todo,
-                          color='#3096FC', at_line=line_number)
+        self.show_tooltip(
+            _("To do"),
+            data.todo,
+            color='#3096FC',
+            at_line=line_number,
+        )
```
ccordoba12: These TODO's should behave in the same way as code analysis warnings.

spyder/widgets/mixins.py
```diff
@@ -755,7 +905,7 @@ def show_object_info(self, text, call=False, force=False):
                             tiptext = name + argspec
                         else:
                             tiptext = signature
-                        self.show_calltip(_("Arguments"), tiptext,
+                        self.show_tooltip(_("Arguments"), tiptext,
```
ccordoba12: Why this change?

spyder/plugins/editor/widgets/codeeditor.py
ccordoba12: It took me a long time but I finally found where those options are set:

https://github.com/spyder-ide/spyder-pyls/blob/0.22.0/pyls/python_ls.py#L141-L143

So we should add `=` to them. But could you explain why is that needed?

goanpeca: >But could you explain why is that needed?

Yes, of course. Take for example this code (the `*` represents the cursor)

```python
def somefunc(param1=0, param2=1, param3=None):
    pass

somefunc(param2=*
```

Without the `=` sign, when we are using a `parameter` that is located somewhere in the call, if we don't add the `=` it won't retrrigger a signature request, and the `parameter` highlighted will be incorrect. In the example `param2` will not be in red. Adding the `=` to the list of "trigger" characters solves this :-)

spyder/plugins/ipythonconsole/widgets/help.py
goanpeca: Ok, (for now) the `console` one will not display highlighted parameter, nor the docstrings.

goanpeca:
>What do you think? Perhaps we should only leave the red commas?

Will paste some new screenshots and we can decide

>(although the current logic is brittle)
>What do you mean by this? Can we improve the situation?

I meant the logic that was highlighting the current parameter. I thin I fixed that by adding a more "complex" regex substitution, rather than a simple string replace (which could potentially replace something in the underlying html code)

spyder/widgets/mixins.py
goanpeca: Yes, you are right, I will fix this!

spyder/plugins/editor/widgets/codeeditor.py
goanpeca: Since the signature call is now much more complex, to keep things more organized I created 2 separate methods. 

`show_tooltip` and `show_calltip`. The names explain what they are using and this allows to move all the styling logic inside the mixin.

spyder/plugins/editor/widgets/codeeditor.py
goanpeca: See https://github.com/spyder-ide/spyder/pull/9140#discussion_r275781064

spyder/widgets/mixins.py
goanpeca: https://github.com/spyder-ide/spyder/pull/9140#discussion_r275781064

spyder/plugins/ipythonconsole/widgets/help.py
ccordoba12: Ok, the console works differently and changing things there to adjust its calltips to the look and feel of Editor is out of the scope of this PR.

spyder/plugins/editor/widgets/codeeditor.py
ccordoba12: Ok, thanks for digging that logic up! Could you submit a PR upstream:

https://github.com/palantir/python-language-server

explaining why that's needed? Thanks!

(Let's see if upstream accept it before adding it to our fork).

spyder/plugins/editor/widgets/codeeditor.py
ccordoba12: Ok, thanks for the explanation.

spyder/widgets/mixins.py
ccordoba12: I understand the reasoning you mentioned above, but a call to `show_object_info` (where this code is placed) shows the signature of a callable in our editor as well as its docstring rendered in our help pane.

So I think this should continue to be `show_calltip` and not `show_tooltip`.

ccordoba12:
> There's no problem in making the calltip widget scrollable, but I think it'd be nicer if we added a link at the end of calltip with a text like Press here for addtional help. Then, by pressing on that line, the entire docstring would be displayed in the Help pane.

About my own suggestion: I think for now we should only show signatures in the calltip and not the docstring nor the link I suggested.

We can decide later on what we want to do with the docstring (i.e. making the calltip widget scrollable or adding a link to the bottom of it). But this PR should stay focused on using again the calltip widget and fixing its style.

@goanpeca, what do you think?

goanpeca:
@ccordoba12 this is going to require some more changes. It seems that it is not possible to style a `QToolTip` if it is called using `QToolTip.showText()`, so in order to be consistent with the theming, we basically have to replace any call to `QToolTip.showText()` with something like `self.tooltip_widget.show_text`, and have 2 different CalltipWidgets, one for calltips, and the other for tooltips, with changes to the calltip widget, so that it hides as a QToolTip will do.

I am having some small issue, but it would look something like:

Screen Shot 2019-04-16 at 10 33 15

ccordoba12:
> so in order to be consistent with the theming, we basically have to replace any call to QToolTip.showText() with something like self.tooltip_widget.show_text

Ok, no problem with that.

> and have 2 different CalltipWidgets, one for calltips, and the other for tooltips, with changes to the calltip widget, so that it hides as a QToolTip will do.

Ok, but you mean two instances of CalltipWidget with something like `tooptip=False` on the `__init__` of CalltipWidget, right?

> I am having some small issue, but it would look something like:

It looks really polished and professional!! Thanks!!

Please also remember to test this with the light theme to see if things look ok with it too.

goanpeca:
>Ok, but you mean two instances of CalltipWidget with something like tooptip=False on the __init__ of CalltipWidget, right?

Yes, correct

>Please also remember to test this with the light theme to see if things look ok with it too.

Screen Shot 2019-04-16 at 10 51 21

ccordoba12:
@goanpeca, there's also another important thing: the process of modifying the signature to add those red parens, commas and equal signs must *only* be done when we're handling an LSP response that corresponds to *Python*.

Else that doesn't apply. I mean, those characters can have a different meaning in other programming languages.

goanpeca:
>goanpeca, there's also another important thing: the process of modifying the signature to add those red parens, commas and equal signs must only be done when we're handling an LSP response that corresponds to Python.

>Else that doesn't apply. I mean, those characters can have a different meaning in other programming languages.

Fixed!

Still need to fix the internal console and the QtConsole

goanpeca:
@ccordoba12 and @CAM-Gerlach I would appreciate if you could give it a test. I still need to fix some stuff in the QTconsole, but the Editor should be working as expected now

goanpeca:
@CAM-Gerlach thanks for testing. Hmmm what system are you using (and env) cause I don't have the trimmed down signatures. (as you can see in my screenshots above)

CAM-Gerlach:
Thanks @goanpeca 

For ``subprocess.Popen()``, I got the truncated signature unlike the above, and I can also repro it in ``master``, both on my Windows 8.1 Miniconda 2018.12+ LSP 0.21.5 env. However, I can also repro it on my Fedora 28 LXQt VM running much older deps and LSP 0.19. Full ``conda list`` for each is below. 

However, all of them, including yours, do not correctly show the summary line, but rather an arbitrary, unformatted line from deeper in the docstring.

image

vs.

image

On Linux:

image

``conda list``:

<details>

```
# packages in environment at C:\Miniconda3\envs\spyder-4:
#
# Name                    Version                   Build  Channel
alabaster                 0.7.12                   py37_0
asn1crypto                0.24.0                   py37_0
astroid                   2.1.0                    py37_0
atomicwrites              1.3.0                      py_0
attrs                     18.2.0           py37h28b3542_0
autopep8                  1.4.3                    py37_0
babel                     2.6.0                    py37_0
backcall                  0.1.0                    py37_0
blas                      1.0                         mkl
bleach                    3.1.0                    py37_0
ca-certificates           2019.1.23                     0
certifi                   2018.11.29               py37_0
cffi                      1.11.5           py37h74b6da3_1
chardet                   3.0.4                    py37_1
cloudpickle               0.7.0                      py_0
colorama                  0.4.1                    py37_0
coverage                  4.5.2            py37he774522_0
cryptography              2.5              py37h7a1dbc1_0
cycler                    0.10.0                   py37_0
cython                    0.29.4           py37ha925a31_0
decorator                 4.3.2                    py37_0
defusedxml                0.5.0                    py37_1
docutils                  0.14                     py37_0
entrypoints               0.3                      py37_0
fastcache                 1.0.2            py37hfa6e2cd_2
flaky                     3.5.3                    py37_0
freetype                  2.9.1                ha9979f8_1
future                    0.17.1                   py37_0
icc_rt                    2019.0.0             h0cc432a_1
icu                       58.2                 ha66f8fd_1
idna                      2.8                      py37_0
imagesize                 1.1.0                    py37_0
intel-openmp              2019.1                      144
ipykernel                 5.1.0            py37h39e3cac_0
ipython                   7.2.0            py37h39e3cac_0
ipython_genutils          0.2.0                    py37_0
isort                     4.3.4                    py37_0
jedi                      0.13.2                   py37_0
jinja2                    2.10                     py37_0
jpeg                      9b                   hb83a4c4_2
jsonschema                2.6.0                    py37_0
jupyter_client            5.2.4                    py37_0
jupyter_core              4.4.0                    py37_0
keyring                   17.1.1                   py37_0
kiwisolver                1.0.1            py37h6538335_0
lazy-object-proxy         1.3.1            py37hfa6e2cd_2
libpng                    1.6.36               h2a8f88b_0
libsodium                 1.0.16               h9d3ae62_0
libtiff                   4.0.10               hb898794_2
markupsafe                1.1.0            py37he774522_0
matplotlib                3.0.2            py37hc8f65d3_0
mccabe                    0.6.1                    py37_1
mistune                   0.8.4            py37he774522_0
mkl                       2019.1                      144
mkl_fft                   1.0.10           py37h14836fe_0
mkl_random                1.0.2            py37h343c172_0
mock                      2.0.0                    py37_0
more-itertools            5.0.0                    py37_0
mpmath                    1.1.0                    py37_0
nbconvert                 5.4.0                    py37_1
nbformat                  4.4.0                    py37_0
numpy                     1.15.4           py37h19fb1c0_0
numpy-base                1.15.4           py37hc3f5095_0
numpydoc                  0.8.0                    py37_0
olefile                   0.46                     py37_0
openssl                   1.1.1a               he774522_0
packaging                 19.0                     py37_0
pandas                    0.24.1           py37ha925a31_0
pandoc                    1.19.2.1             hb2460c7_1
pandocfilters             1.4.2                    py37_1
parso                     0.3.2                    py37_0
pbr                       5.1.1                    py37_0
pickleshare               0.7.5                    py37_0
pillow                    5.4.1            py37hdc69c19_0
pip                       19.0.1                   py37_0
pluggy                    0.8.1                    py37_0
prompt_toolkit            2.0.8                      py_0
psutil                    5.5.0            py37he774522_0
py                        1.7.0                    py37_0
pycodestyle               2.5.0                    py37_0
pycparser                 2.19                     py37_0
pydocstyle                3.0.0                    py37_0
pyflakes                  2.1.0                    py37_0
pygments                  2.3.1                    py37_0
pylint                    2.2.2                    py37_0
pyopenssl                 19.0.0                   py37_0
pyparsing                 2.3.1                    py37_0
pyqt                      5.9.2            py37h6538335_2
pysocks                   1.6.8                    py37_0
pytest                    4.0.2                    py37_0
pytest-cov                2.6.1                    py37_0
pytest-mock               1.10.0                   py37_0
pytest-qt                 3.2.1                      py_0    spyder-ide
python                    3.7.2                h8c8aaf0_2
python-dateutil           2.7.5                    py37_0
python-jsonrpc-server     0.1.2                      py_0    spyder-ide
python-language-server    0.21.5                py37_1000    spyder-ide
pytz                      2018.9                   py37_0
pywin32                   223              py37hfa6e2cd_1
pyzmq                     17.1.2           py37ha925a31_2
qdarkstyle                2.6.5                      py_0    spyder-ide
qt                        5.9.7            vc14h73c81de_0
qtawesome                 0.5.6                      py_0
qtconsole                 4.4.3                    py37_0
qtpy                      1.6.0                      py_0
requests                  2.21.0                   py37_0
rope                      0.11.0                   py37_0
scipy                     1.2.0            py37h29ff71c_0
setuptools                40.7.3                   py37_0
sip                       4.19.8           py37h6538335_0
six                       1.12.0                   py37_0
snowballstemmer           1.2.1                    py37_0
sphinx                    1.8.4                    py37_0
sphinxcontrib             1.0                      py37_1
sphinxcontrib-websupport  1.1.0                    py37_1
spyder-kernels            1.2.0                    py37_0    spyder-ide
sqlite                    3.26.0               he774522_0
sympy                     1.3                      py37_0
testpath                  0.4.2                    py37_0
tk                        8.6.8                hfa6e2cd_0
tornado                   5.1.1            py37hfa6e2cd_0
traitlets                 4.3.2                    py37_0
urllib3                   1.24.1                   py37_0
vc                        14.1                 h0510ff6_4
vs2015_runtime            14.15.26706          h3a45250_0
wcwidth                   0.1.7                    py37_0
webencodings              0.5.1                    py37_1
wheel                     0.32.3                   py37_0
win_inet_pton             1.0.1                    py37_1
wincertstore              0.2                      py37_0
wrapt                     1.11.1           py37he774522_0
xz                        5.2.4                h2fa13f4_4
yapf                      0.25.0                   py37_0
zeromq                    4.3.1                h33f27b4_3
zlib                      1.2.11               h62dcd97_3
zstd                      1.3.7                h508b16e_0
```

</details>

goanpeca:
@CAM-Gerlach it is not an arbitrary line. It is the documentation for the particular parameter. Sometimes it will just not be parsed correctly. But that is what is coming from the LSP server. I can try to add the main documentation plus the active parameter.

ccordoba12:
> I can try to add the main documentation plus the active parameter.

I think for now we should only show the signature and not any other additional info below it. I already mentioned it above.

Let's leave for another PR what to do with that info (i.e. displaying the entire docstring by making calitps scrollable or adding a link at the end of the signature, as RStudio does).

CAM-Gerlach:
> it is not an arbitrary line. It is the documentation for the particular parameter.

Ah, silly me. You're right, of course; the formatting had me confused somehow. Sorry about that.

> I can try to add the main documentation plus the active parameter.

Just the summary line and the active parameter would be fine, so as to not clutter it.

> I think for now we should only show the signature and not any other additional info below it. 

We can discuss it elsewhere, but for now I don't see why we should deliberately impose a regression on currently-implemented behavior by eliminating the param description without justification; that is more of an active change than just leaving what we have currently. Describing the current parameter is a major reason to have the calltip in the first place, after all, and we don't want that getting baked into beta 2. We may as well at least leave that IMO, possibly with the summary line if @goanpeca has that implemented, and iterate further later, unless there is a clear rationale presented for why its harmful to leave it.

ccordoba12:
> I don't see why we should deliberately impose a regression on currently-implemented behavior by eliminating the param description without justification

It's not a regression because:

1. I want feature parity with Spyder 3 to start with. Spyder 3 doesn't have that additional info, so (for now) Spyder 4 won't have it either.
2. @andfoy implemented the LSP stuff in a bit of a hurry, so I never had the chance to properly review what he did. In other words, he left that additional info to see what we should do about it afterwards, but only until now we've came to point to see what to do with it.

@CAM-Gerlach, I don't want to derail this PR discussing further about it, please don't insist. And just to clarify things (because it seems sometimes you think my decisions are definite): we will review what to do with that additional info in another PR (I said it clearly above). I hope the lack of that info for a couple of weeks won't be that serious for you or other people using `master`.

CAM-Gerlach:
Okay sure, if you "insist" jeje. Hopefully whatever gets decided will make it into beta 2, and I'll stay tuned for that next PR.

ccordoba12:
> Hopefully whatever gets decided will make it into beta 2

Yes, it will. The thing is we can also display additional info in the Console (sorry, I forgot to mention it), so it's better to leave for another PR how to display it for both the Editor and the Console at the same time.

CAM-Gerlach:
We presumably should also decide what to do with the ``Hint`` displayed when pressing the help shortcut as well, as it seems to be using similar machinery and could benefit from some of these enhancements (syntax highlighting, formatting, etc) as I mentioned above.

goanpeca:
>As part of that discussion, we presumably should also decide what to do with the Hint displayed when pressing the help shortcut as well, as it seems to be using similar machinery and could benefit from some of these enhancements (syntax highlighting, formatting, etc) as I mentioned above.

Indeed, I will open an issue with some VSCode hint screenshots for discussion.

spyder/plugins/editor/widgets/codeeditor.py
goanpeca: See: https://github.com/palantir/python-language-server/pull/542

spyder/widgets/mixins.py
goanpeca: For now the colors used work for both themes, but perhaps we should indeed update the logic to use "better" colors. I think that deserves a separate PR

spyder/widgets/mixins.py
goanpeca: Fixed!

goanpeca:
Current look:

## Editor
Screen Shot 2019-04-22 at 10 10 02


## Console
Screen Shot 2019-04-22 at 10 09 53

goanpeca:
@ccordoba12 ready for review again. I suggest that any additional style (theme) related issues, are solved in a separate PR

CAM-Gerlach:
@goanpeca I'm still only seeing the first few parameters (even after I type more into the function itself) shown in the calltip, on both my Windows and Linux machines with quite different environments as documented above. But since I can repro that outside this PR, it could be left for a followup. I'll pull those comments into a separate issue as requested by @goanpeca .

ccordoba12:
@dalthviz, please test things on Windows to see if you can reproduce @CAM-Gerlach's problems (I can't do it on Linux).

spyder/plugins/ipythonconsole/widgets/help.py
```diff
@@ -125,5 +129,6 @@ def _handle_inspect_reply(self, rep):
             if content.get('status') == 'ok' and content.get('found', False):
                 signature = self.get_signature(content)
                 if signature:
-                    self._control.show_calltip(_("Arguments"), signature,
-                                               signature=True, color='#2D62FF')
+                    # TODO: Pass the language from the Console to the calltip
+                    self._control.show_calltip(signature, color='#999999',
+                                               is_python=True)
```
goanpeca: @ccordoba12 how can I get the information of the actual language the (Ipython) console is running. To make sure this is generic.

CAM-Gerlach:
> @dalthviz, please test things on Windows to see if you can reproduce @CAM-Gerlach's problems (I can't do it on Linux)

Since it was not introduced here, I thought we were dealing with that in a separate issue, #9189 , so as to not derail this PR? ;)

ccordoba12:
Ok, I thought it was introduced by this PR. No problem then :+1:

CAM-Gerlach:
> Ok, I thought it was introduced by this PR. No problem then +1

I can sympathize with making that mistake ^_^

ccordoba12:
@goanpeca, please don't merge until I show you how things look on Linux.

goanpeca:
I installed a Kubuntu VM and tried to fix the issue.

Now it looks like this:

VirtualBox_Kubuntu_22_04_2019_22_13_09

I guess the border is no longer neeeded

spyder/plugins/editor/widgets/codeeditor.py
goanpeca: Maybe you can make some noise @ccordoba12 ?

spyder/widgets/mixins.py
```diff
@@ -64,72 +70,260 @@ def calculate_real_position(self, point):
...
+        if is_dark_interface():
+            css = qdarkstyle.load_stylesheet_from_environment()
+            widget.setStyleSheet(css)
+            palette = widget.palette()
+            background = palette.color(palette.Window).lighter().name()
+            name = widget.__class__.__name__
+            widget.setObjectName(name)
+            extra_css = '''
+                {0}#{0} {{background-color:{1};}}'''.format(name, background)
+            widget.setStyleSheet(css + extra_css)
```
goanpeca: So I could not get it to work by updating the palette, but the palette colors are updated by the stylesheet that is why I first apply the CSS then get the color to make it lighter and then reapply an extended CSS which includes the lighter background.

ccordoba12: Ok, no problem with that. Hopefully @dpizetta will release `qdarkstyle` 3 soon, making this far more easier.

spyder/widgets/mixins.py
```diff
@@ -64,72 +70,264 @@ def calculate_real_position(self, point):
...
+            widget.setObjectName(name)
+            extra_css = '''
+                {0}#{0} {{
+                    background-color:{1};
+                    border: 1px solid {2};
+                }}'''.format(name, background, border)
+            widget.setStyleSheet(css + extra_css)
+
+    def _format_text(self, title, text, color, ellide=False):
+        """"""
```
ccordoba12: Missing docstring.

spyder/widgets/mixins.py
```diff
@@ -64,72 +70,264 @@ def calculate_real_position(self, point):
...
+            title=title,
+            text=text,
+        )
+
+        return tiptext
+
+    def _format_signature(self, signature, doc='', parameter='',
+                          parameter_doc='', color=_DEFAULT_TITLE_COLOR,
+                          is_python=False):
+        """"""
```
ccordoba12: Missing docstring.

spyder/widgets/mixins.py
```diff
@@ -755,8 +953,8 @@ def show_object_info(self, text, call=False, force=False):
                             tiptext = name + argspec
                         else:
                             tiptext = signature
-                        self.show_calltip(_("Arguments"), tiptext,
-                                          signature=True, color='#2D62FF')
+                        # TODO: Select language and pass it to call
```
ccordoba12: Please mention this in the same issue about getting the kernel language from the console. I think both things can be solved in the same PR.

goanpeca: https://github.com/spyder-ide/spyder/issues/9194

spyder/widgets/mixins.py
goanpeca: Done

spyder/widgets/mixins.py
goanpeca: Done

spyder/widgets/calltip.py
```diff
@@ -29,6 +29,91 @@
...
 
 
+class ToolTipWidget(QLabel):
+    """
+    Shows tooltips that can be styled with the different themings.
+    """
+
+    def __init__(self, parent=None, as_tooltip=False):
+        """
+        Shows tooltips that can be styled with the different themings.
```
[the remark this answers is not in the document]
goanpeca: Done

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1734,13 +1732,23 @@ def process_code_analysis(self, check_results):
         self.linenumberarea.update()
         self.classfuncdropdown.update()
 
+    def hide_tooltip(self):
+        """"""
```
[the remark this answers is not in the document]
goanpeca: Done

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -856,39 +856,34 @@ def request_signature(self):
...
-                title = func_signature.replace(
-                    parameter['label'], '{0}{1}</span>'.format(
-                        color_change_str.format(family, size, '#daa520'),
-                        parameter['label']))
-                tooltip_text = "{0}{1}".format(parameter_str, func_doc)
+
+                signature_data = signature_params['signatures']
+                documentation = signature_data['documentation']
+
+                if PY2:
```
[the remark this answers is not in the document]
goanpeca: Done

spyder/plugins/ipythonconsole/widgets/help.py
goanpeca: https://github.com/spyder-ide/spyder/issues/9194

spyder/widgets/mixins.py
goanpeca: https://github.com/spyder-ide/spyder/issues/9194

PR #9138, opened 2019-04-10
PR: Catch errors generated when handling LSP responses in CodeEditor

ccordoba12:
@CAM-Gerlach, this should avoid the LSP services to stop from working when there's an error in our side handling its responses.

So you don't need to have debugging on from now on, just report the errors you encounter through our error report dialog.

Fixes #8213.

PR #9142, opened 2019-04-10
PR: Correctly log errors when handling LSP responses for Python 2

goanpeca:
## Description of Changes

- [x] Adds tests
- [x] Adds screenshots

On Python 2 the `log_lsp_handle_errors` method was not working properly as `stack_info` is not a valid kwarg in Python 2

The following code was used as example:

```python
# -*- coding: utf-8 -*-
"""Example."""


def hello(text='', number=1):
    """
    Returns
    -------
        Testing
    """
    print(text, number)


hello(
```

After typing the `(` the error dialog was triggered.

### Affirmation


spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -736,6 +736,10 @@ def log_lsp_handle_errors(self, message):
             # there. So we intentionally leave an error in this call
             # to get the entire stack info generated by it, which
             # gives the info we need from users.
+            if PY2 and error:
+                # For Python 2 we need the actual error so we can raise it
+                raise error
```
ccordoba12: The whole point of logging LSP handling errors, instead of raising them, is that after they are raised, then code completion, linting and all other services provided by LSP servers can't be handled by Spyder anymore.

I just tested this with your solution and it happens for Python 2 as it was happening for both Python 2 and 3 before PR #9138. So my proposed solution is this:

```python
if PY2:
    logger.error(message, exc_info=True)
    print(message, file=sys.stderr)
else:
    logger.error('%', 1, stack_info=True)
```

It doesn't show as much info as `stack_info`, but it's informative enough. Besides, given that Python 2 downloads have decreased a lot in the last year (just 18%, according to Anaconda statistics), I think we should not waste too much time on this.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -736,11 +736,12 @@ def log_lsp_handle_errors(self, message, error=None):
             # there. So we intentionally leave an error in this call
             # to get the entire stack info generated by it, which
             # gives the info we need from users.
-            if PY2 and error:
+            if PY2:
                 # For Python 2 we need the actual error so we can raise it
```
ccordoba12: This comment doesn't apply anymore. Please change it.

PR #9155, opened 2019-04-14
PR: Simplify custom layout definition with the new features in Qt 5.6

### LINKED ISSUE #8613: Use new functionality in Qt 5.6 to resize dockwidgets programatically
This way we could avoid the kind of hacky methods we have right now to create the Matlab, RStudio, etc, layouts.

This new change is mentioned here:

https://woboq.com/blog/qdockwidget-changes-in-56.html

goanpeca:
## Description of Changes

- [x] Code was organized
- [x] Additional comments added
- [x] Added tests

Not sure how this changes could be tested at the moment

The code for setting the default layouts has been simplified to make use of Qt 5.6 `resizeDocks`

Issues on linux ubuntu have been solved. Also a message handler for Qt was added to avoid printing messages that might be unnecessary warnings. On dev mode we will still print all warnings but on the internal console.

### Issue(s) Resolved

Fixes #8613

### Affirmation


goanpeca:
>I'm not sure if there's something specific you want me to test, but on Windows 8.1 x64 with Qt/PyQt 5.9 and Python 3.7, both with and without safe mode (clean prefs) I checked basic Spyder usage, changing to several different layouts, moving around panes, editing the layout preferences and resetting to the default layout, and didn't experience any visible issues or error messages. I did get the ``QMainWidget::resizeDocks: all sizes need to be larger than 0`` message as you did in the console without debug mode on quite a bit, but there was no apparent ill effect.

>On Fedora 28 LXQt in a VM, with Qt/PyQt 5.9 and Python 3.6, on starting up Spyder in clean and non-clean settings mode, the window as maximized was seemingly twice the height of the screen, with only the first set of panes visible. Switching to the Spyder default and Matlab layouts produced the same effect, with the latter also displaying a strange artifact where the window behind Spyder showed through for part of the screen, which went away when I minimized and re-maximized it. Resetting the layout to default didn't change this either; in fact, the confirmation dialog showed up near the bottom of the screen. Even exploring all the options in my window manager, I could not get the window's contents to resize properly, and if I ran the latest ``master`` in the same environment with or without clean prefs as well, everything looked normal. I also got the same Console message as on Windows. 

image

Again, let me know what I can do to help isolate this further.

goanpeca:
@CAM-Gerlach I saw the issue you mentioned on Ubuntu and I think I have fixed it. Could you check again when you have some time?

Thanks :-)

spyder/app/mainwindow.py
```diff
@@ -246,6 +247,28 @@ def setup_logging(cli_options):
...
+    blacklist for messages that are being printed for no apparent
+    reason. Anything else will get printed in the internal console.
+
+    On DEV mode, all messages are printed.
+    """
+    BLACKLIST = [
+        'QMainWidget::resizeDocks: all sizes need to be larger than 0',
+    ]
+    if DEV or msg_string not in BLACKLIST:
+        print(msg_string)  # spyder: test-skip
```
goanpeca: @ccordoba12 do we want this printed to the internal console or we want this on the terminal?

If so how is that accomplished?

Thanks

goanpeca:
CI is ok, it is just pointing to an old Build 🤷‍♂️ 
 Screen Shot 2019-04-14 at 12 22 53

spyder/app/mainwindow.py
```diff
@@ -246,6 +247,28 @@ def setup_logging(cli_options):
...
+def qt_message_handler(msg_type, msg_log_context, msg_string):
+    """
+    Qt warning messages are intercepted by this handler.
+
+    On some oprating systems, warning messages might be displayed
+    even if the actual message does not apply. This filter adds a
+    blacklist for messages that are being printed for no apparent
+    reason. Anything else will get printed in the internal console.
+
+    On DEV mode, all messages are printed.
```
CAM-Gerlach: ``In DEV mode`` [...]`

spyder/app/mainwindow.py
ccordoba12: If you want to print to the terminal, you need to use the `logging` module. It can be `logger.info` to not miss the message among the huge amount produced by `logger.debug`.

I mean, I say it because you can start Spyder like this

    python bootstrap.py -- --debug-info minimal

to only see info messages.

ccordoba12:
I restarted the build.

spyder/app/mainwindow.py
```diff
@@ -246,6 +247,28 @@ def setup_logging(cli_options):
                             filename=log_file,
                             filemode='w+')
 
+
+def qt_message_handler(msg_type, msg_log_context, msg_string):
+    """
+    Qt warning messages are intercepted by this handler.
+
+    On some oprating systems, warning messages might be displayed
```
ccordoba12: Please change `oprating` to `operating`

spyder/app/mainwindow.py
```diff
@@ -1556,115 +1579,142 @@ def setup_default_layouts(self, index, settings):
         global_hidden_toolbars = [self.source_toolbar, self.edit_toolbar,
                                   self.search_toolbar]
         # Layout definition
-        # layouts are organized by columns, each colum is organized by rows
-        # widths have to add 1.0, height per column have to add 1.0
+        # --------------------------------------------------------------------
+        # Layouts are organized by columns, each column is organized by rows
```
ccordoba12: Is this line missing a period? And the next one should start with `Width`?

If not, I don't understand the meaning of this comment.

goanpeca:
>Could we add tests for custom layouts? Some people complained that those are not working in Spyder 3.

You mea custom in general or the defaults ones (matlab, rstudio, horizontal/verical spit?)

spyder/app/mainwindow.py
goanpeca: Thanks!

spyder/app/mainwindow.py
goanpeca: Yes it was.

spyder/app/mainwindow.py
goanpeca: Thanks

goanpeca:
>Could we add tests for custom layouts? Some people complained that those are not working in Spyder 3.

I am no sure what we could test. The only thing I can think of is having the method return the location of widgets and compare if after changing the layout (and firing a `sig_layout_set` or something) compare the dockwidget positions.

goanpeca:
@CAM-Gerlach what do you mean by `the window cannot be moved and various graphical glitches result in the "unocuppied" area,` ? and 

```
. The apparent size of the window contents appears to be the same regardless of what the previous size was.
```

CAM-Gerlach:
> the window cannot be moved and various graphical glitches result in the "unocuppied" area,

On Fedora 28 LXQt, setting the window layout to a different layout maximizes the window frame and it behaves accordingly (cannot be moved, etc), but the window contents are set to a much smaller fixed size and a (sometimes artifact-filled) version of whatever is behind it is shown throughout the rest of the window. The above screenshot demonstrates how this looks, without artifacts for clarity. I haven't observed this problem previously without this PR.

> The apparent size of the window contents appears to be the same regardless
of what the previous size was.

The size of the window contents (the visible portion of the Spyder UI inside the outer window frame) is set to the same fixed size (much smaller than the full maximized size) when changing the layout, regardless of whether the window was previously maximized or not, or what the dimensions of the unmaximized window were. Upon unmaximizing the resulting window, the window frame is set to this size, same as the contents.

goanpeca:
Ok, will look more into it. I think it is safe to assume that changing the layout requires going to maximized mode, right?

CAM-Gerlach:
> I think it is safe to assume that changing the layout requires going to maximized mode, right?

Not sure if that was directly at me as opposed to @ccordoba12 ; (I don't see any real UX reason why it should; the panel layout should presumably be defined relative to the window size, not the screen size either way, but either of you have way more insight on the technical reasons for this ). Just to be clear, the key issue for me is that regardless the window contents should match the size of the window frame, whatever dimensions it needs to be.

goanpeca:
Yes i still need to address the issues

ccordoba12:
> I don't see any real UX reason why it should; the panel layout should presumably be defined relative to the window size, not the screen size either way, but either of you have way more insight on the technical reasons for this

I really don't see a problem with maximizing the screen after a layout switch, if that fixes the bug @CAM-Gerlach is experiencing. It's a minor annoyance, but users would have to accept it if that means a reliable switch.

ccordoba12:
> I am no sure what we could test.

You could save a custom layout in `spyder.ini` (with `CONF.set`), then load it through the graphical interface (i.e. by calling the appropiate method in `app/mainwindow.py`) and finally verifying that some of the panes of the loaded layout are in the right place according to the configuration you saved in the first place.

CAM-Gerlach:
> I really don't see a problem with maximizing the screen after a layout switch, if that fixes the bug @CAM-Gerlach is experiencing. 

I don't either; the window *already* is being maximized; the actual drawn content just isn't following suit. To be clear, what I was responding to whether there was a reason *to* maximize the window from a UX perspective, not whether there was a reason *not* to.

To be more specific, with the overhead with the VM, I can't record at a high enough frame rate to reliably capture it but what I *believe* I see when I trigger a layout change on Linux is first the window contents and frame are both maximized (correctly), then the window contents jumps back to the smaller size but not the window frame. On Windows, by contrast, the window and contents first resizes to a smaller size, then both correctly maximize. This happens over the course of a few drawn frames.

goanpeca:
Understood. The behavior I want to achieve is that I don't like the jumping back and fort, so I am trying to disable that programmatically, and that is why we need to maximize (to have a reliable widget size that can be fixed during the layout definition.

I have reproduced the error and I am looking at options now. Thanks for the feedback @CAM-Gerlach

goanpeca:
@CAM-Gerlach could you please test tis again when you have some time?

Thanks :-)

goanpeca:
@ccordoba12 added a simple test.

CAM-Gerlach:
@goanpeca Everything is as before on Windows, while on Linux, changing layouts now works without graphical glitches, although it very annoyingly unmaximizes itself and sets itself to an arbitrary fixed size and location, ignoring any previous non-maximized window size. If we're going to do that, we may as well just have it maximize if we can't just preserve the previous window size, rather than just putting it at an arbitrary position and size that is almost certain to be less desirable than just maximizing it.

goanpeca:
@CAM-Gerlach thanks for testing this. 

>although it very annoyingly unmaximizes itself and sets itself to an arbitrary fixed size and location, ignoring any previous non-maximized window size.

That is strange and not the behavior I see (or expect, based on what the code "does")

>If we're going to do that, we may as well just have it maximize if we can't just preserve the previous window size, rather than just putting it at an arbitrary position and size that is almost certain to be less desirable than just maximizing it.

This is exactly what the code is supposed to do. Maximize, freeze the size, change layouts, unfreeze the size.

Just to try something.

If you start from a fresh spyder (reset), then close it. Then start it again, and try the layouts, do you see the same thing?

spyder/app/tests/test_mainwindow.py
```diff
@@ -1784,5 +1792,33 @@ def test_render_issue():
     assert test_traceback in test_issue_2
 
 
+@flaky(max_runs=3)
+def test_custom_layouts(main_window, qtbot):
```
ccordoba12: Please mark this test as slow with the following decorator:

`@pytest.mark.slow`

spyder/app/tests/test_mainwindow.py
```diff
@@ -156,7 +156,11 @@ def main_window(request):
 
     # Check if we need to use introspection in a given test
     # (it's faster and less memory consuming not to use it!)
-    use_introspection = request.node.get_marker('use_introspection')
+    try:
+        use_introspection = request.node.get_marker('use_introspection')
+    except AttributeError:
+        use_introspection = False
```
ccordoba12: What's your pytest version? I mean, this is not necessary with pytest 4.0

spyder/app/tests/test_mainwindow.py
ccordoba12: That will make this test to only run in our CIs and not locally. That's better because most main window tests are quite consuming.

spyder/app/tests/test_mainwindow.py
goanpeca: I had 4.3 and now 4.4, will check again

spyder/app/tests/test_mainwindow.py
goanpeca: Yes, I forgot to re-add it. Thanks

spyder/app/mainwindow.py
goanpeca: I think `print` does what I had more in mind for `dev` mode

spyder/app/tests/test_mainwindow.py
goanpeca: This is a Python2 vs 3 issues apparently. I have 4.4.0  no both envs, but it only fails on my py2 env. So I will leave it for now.

goanpeca:
@ccordoba12 made the changes.

spyder/app/tests/test_mainwindow.py
ccordoba12: Ok, thanks for the clarification.

spyder/app/mainwindow.py
ccordoba12: Ok, good to know :+1:

ccordoba12:
I'll test this on my end before approving. @CAM-Gerlach, if I don't see any issues I'll merge because the problem you're experiencing is probably related to the window manager used by LXQt (a very minor desktop environment compared to Gnome and KDE).

ccordoba12:
@goanpeca, thing are working well for me, expect for one thing: if you activate `Horizontal split` and then activate `Spyder default layout`, the only plugin shown is the Editor.

To fix that, you have to activate any another layout (e.g. `Matlab layout`) and then `Spyder default layout` to get the correct layout.

goanpeca:
>if you activate Horizontal split and then activate Spyder default layout, the only plugin shown is the Editor.

Hmm that is odd. Will take a look. Thanks for testing!

CAM-Gerlach:
I can also confirm I see the same effect as @ccordoba12 described, on both Fedora and Windows and in and not in ``--safe-mode``.

goanpeca:
@ccordoba12 I think this last commit fixes the issue. Not sure what caused it, but it seems to work as expected now.

goanpeca:
🎉

ccordoba12:
@CAM-Gerlach, please give it a last try so we can merge it.

CAM-Gerlach:
@ccordoba12 In work now.

goanpeca:
Thanks for the thorough testing @CAM-Gerlach :-)

PR #9166, opened 2019-04-17
PR: Fix Help plugin string validation for the console

### LINKED ISSUE #9143: Help pane doesn't accept aliased modules
**Steps to reproduce:**
* Import some module in Python with an alias - e.g.
```python
import numpy as np
```
(not `import numpy`)
* Type the name of one of the imported module's function/classes in the help pane along with its prefix - e.g. `np.arange`
* Press enter.

**Expected behavior:** should show the help for the function/class that was requested. Should also ideally not need an enter keypress, like it was in 3.3.1.

**Actual behavior:** deletes what was just written and goes back to what was previously there, or to blank if nothing was written before. As such, cannot get help for a function inside an aliased package unless importing it like
```python
from numpy import arange
```
Will also not take it with the package's full name - e.g. `numpy.arange` would also not work.

**Setup:** Spyder 3.3.4, running on windows 10, installed from Anaconda, Python 3.6.8.

dalthviz:
## Description of Changes

Fix validation to correctly check if an object is defined in the shell attached to the Help or current Console.

### Issue(s) Resolved

Fixes #9143 

### Affirmation

### DIFF
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -266,6 +266,57 @@ def test_filter_numpy_warning(main_window, qtbot):
     CONF.set('variable_explorer', 'minmax', False)
 
 
+@pytest.mark.use_introspection
+def test_get_help_combo(main_window, qtbot):
+    """
+    Test that Help works when called from the Editor and the IPython console.
+    """
+    shell = main_window.ipyconsole.get_current_shellwidget()
+    control = shell._control
+    qtbot.waitUntil(lambda: shell._prompt_html is not None,
+                    timeout=SHELL_TIMEOUT)
+
+    help_plugin = main_window.help
+    webview = help_plugin.rich_text.webview._webview
+    if WEBENGINE:
+        webpage = webview.page()
+    else:
+        webpage = webview.page().mainFrame()
+
+    # --- From the console ---
+    # Write some object in the console
+    with qtbot.waitSignal(shell.executed):
+        shell.execute('import numpy as np')
+
+    # Get help - numpy
+    help_plugin.combo.selected()
+
+    qtbot.keyClicks(help_plugin.combo, 'numpy', delay=100)
+
+    # Check that a expected text is part of the page
+    qtbot.waitUntil(lambda: check_text(webpage, "NumPy"), timeout=6000)
+
+    # Get help - numpy.arange
+    qtbot.keyClick(help_plugin.combo, Qt.Key_Right)
+    qtbot.keyClicks(help_plugin.combo, '.arange', delay=100)
+
+    # Check that a expected text is part of the page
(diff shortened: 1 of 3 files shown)


CAM-Gerlach:
> Fix validation to correctly check if and object if denifed in the shell attached to the Help or current Console

I think I parsed this correctly and fixed up the various problems, but just letting you know in case I didn't; it was a tad garbled...

spyder/app/tests/test_mainwindow.py
```diff
@@ -266,6 +266,60 @@ def test_filter_numpy_warning(main_window, qtbot):
     CONF.set('variable_explorer', 'minmax', False)
 
 
+@pytest.mark.slow
+@flaky(max_runs=3)
+@pytest.mark.use_introspection
+@pytest.mark.skipif(PY2,
+                    reason="Times out in PY2")
```
ccordoba12: Please rewrite these two lines like this:

`@pytest.mark.skipif(PY2, reason="Times out in PY2")`

spyder/app/tests/test_mainwindow.py
```diff
@@ -266,6 +266,60 @@ def test_filter_numpy_warning(main_window, qtbot):
     CONF.set('variable_explorer', 'minmax', False)
 
 
+@pytest.mark.slow
+@flaky(max_runs=3)
+@pytest.mark.use_introspection
```
ccordoba12: Please remove this marker. It's not needed because in Spyder 3 we only extract object info from the Console, not from the Editor (that could change in Spyder 4).

spyder/app/tests/test_mainwindow.py
```diff
@@ -266,6 +266,60 @@ def test_filter_numpy_warning(main_window, qtbot):
...
+@pytest.mark.skipif(PY2,
+                    reason="Times out in PY2")
+def test_get_help_combo(main_window, qtbot):
+    """
+    Test that Help works when called from the Editor and the IPython console.
```
ccordoba12: Please change this to

    Test that Help works when called from the IPython console.

ccordoba12: Sorry, that's not correct. This is:

    Test that Help combo box is working as expected.

CAM-Gerlach: Since we have some extra characters, we should probably be clearer and more explicit about what ``as expected`` specifically means in this context.

ccordoba12: Ok, let's change this to

    Test that Help can get display docstrings for objects introduced on its combobox.

CAM-Gerlach: Fixed some issues:

```
Test that Help can display docstrings for names typed in its combobox.
```

ccordoba12:
@dalthviz, this is ready, please don't add more commits here. I'll fix the Azure error and then merge.

PR #9182, opened 2019-04-20
PR: Fix Source menu entry to show code style warnings

### LINKED ISSUE #8121: Option to show/hide code style analysis warnings does not work
## Problem Description

Unchecking.checking the menu option `Source > Show code style warning (pep8)` does not do anything.

image

1. Uncheck the menu option  `Source > Show code style warning (pep8)`

Warning should disappears from the left and right panels in the Editor, but they don't. 

## Versions

* Spyder version: Master https://github.com/spyder-ide/spyder/commit/53b3619422d8d916baa7aced9afdc58833955b20
* Python version: Python 3.6.6 64-bit
* Qt version: Qt 5.9.6
* PyQt version:  PyQt5 5.9.2

### Dependencies

```
sphinx >=0.6.6    :  1.8.1 (OK)
pyls >=0.19.0     :  0.19.0 (OK)
numpy >=1.7       :  1.15.2 (OK)
qtconsole >=4.2.0 :  4.4.1 (OK)
IPython >=4.0     :  7.0.1 (OK)
```

steff456:
## Description of Changes

Correct the action for the option under `Source > Show code style warnings`, add a new option `Source > Show docstring style warnings` and fix TODO menu.

### Issue(s) Resolved

Fixes #8121

### Affirmation

steff456


spyder/plugins/editor/plugin.py
```diff
@@ -1007,18 +1013,12 @@ def _toggle_checkable_action(self, checked, editorstack_method, conf_name):
         """
         if self.editorstacks:
             for editorstack in self.editorstacks:
-                try:
-                    editorstack.__getattribute__(editorstack_method)(checked)
-                except AttributeError as e:
-                    logger.error(e, exc_info=True)
```
ccordoba12: Please restore these three lines above. They are needed for the other options in the Source menu to work (e.g. to activate/deactivate indent guides, etc).

spyder/plugins/editor/plugin.py
```diff
@@ -761,12 +761,17 @@ def get_plugin_actions(self):
                 _("Show code style warnings (pep8)"),
```
ccordoba12: Please remove `(pep8)` from this text.

spyder/plugins/editor/plugin.py
```diff
@@ -761,12 +761,17 @@ def get_plugin_actions(self):
...
+            'set_docstring_warnings')
+
         self.checkable_actions = {
                 'blank_spaces': showblanks_action,
                 'scroll_past_end': scrollpastend_action,
                 'indent_guides': showindentguides_action,
                 'show_class_func_dropdown': show_classfunc_dropdown_action,
-                'code_analysis/pep8': showcode_analysis_pep8_action}
+                'code_analysis/pep8': showcode_analysis_pep8_action,
+                'show_docstring_warnings': show_docstring_warnings_action}
```
CAM-Gerlach: Is there a reason code style analysis is called ``code_analysis/pep8`` while docstring style analysis is called ``show_docstring_warnings``? Seems rather inconsistent to me, but it may be too late/out of scope to change or there's other reasoning behind it.

spyder/plugins/editor/plugin.py
```diff
@@ -1473,7 +1473,7 @@ def update_code_analysis_actions(self):
         results = editor.get_current_warnings()
         # Update code analysis buttons
         state = CONF.get('lsp-server', 'code_completion') \
-                and results is not None and len(results)
+            and results is not None and len(results)
```
goanpeca: We could maybe not use `\` (It is a bit unreadable)

Perhaps we could use?

```python
        state = (CONF.get('lsp-server', 'code_completion')
                 and results is not None and len(results)
                 and results is not None and len(results))
```

Thoughts @spyder-ide/core-developers ?

steff456: I will change it!

CAM-Gerlach: > We could maybe not use \ (It is a bit unreadable)

Yes, I think we (including @ccordoba12 ) all agree on this and have been instructing contributors accordingly. Its also strongly discouraged in PEP 8 (and was even proposed to be removed from the language at some point)

> Perhaps we could use?

PEP 8 recommends prefixing rather than suffixing the operator here (i.e. the first option, which is likely why linters prefer it), and I find it much more readable myself (its what I use); I'm not sure why you find it less readable at least in this case since the ``and`` operators line up rather than being awkwardly stuck on the end, while the first statement doesn't parallel with the rest anyway. However, IIRC @ccordoba12 prefers suffixing it at least for ``+``, etc himself. As far as I'm aware; it has been left up to the discretion of the contributor.

ccordoba12:
@steff456, I helped you to finish this one because it turned out to be trickier than I thought.

@goanpeca, please give it another review.

steff456:
> @steff456, I helped you to finish this one because it turned out to be trickier than I thought.
> 
> @goanpeca, please give it another review.

Thanks!

spyder/plugins/editor/plugin.py
```diff
@@ -898,7 +900,11 @@ def get_plugin_actions(self):
                                MENU_SEPARATOR,
                                self.previous_edit_cursor_action,
                                self.previous_cursor_action,
-                               self.next_cursor_action]
+                               self.next_cursor_action,
+                               MENU_SEPARATOR,
+                               eol_menu,
+                               trailingspaces_action,
+                               fixindentation_action]
```
goanpeca: For this log code snippets I suggest something like:


```python
        source_menu_actions = [
            showblanks_action,
            MENU_SEPARATOR,
            self.previous_edit_cursor_action,
            self.previous_cursor_action,
            self.next_cursor_action,
            ]
```

spyder/plugins/editor/plugin.py
```diff
@@ -972,50 +978,56 @@ def update_font(self):
...
+            action.setChecked(self.get_option(conf_name))
+        else:
+            action.setChecked(CONF.get('lsp-server', conf_name))
+
         return action
 
     @Slot(bool, str, str)
-    def _toggle_checkable_action(self, checked, editorstack_method, conf_name):
-        """Handle the toogle of a checkable action.
+    def _toggle_checkable_action(self, checked, method, conf_name):
```
goanpeca: perhaps it would be better to make more clear that `method` is a `method_name`

The same way `conf_name` is hinting its `str` nature.

spyder/plugins/editor/plugin.py
```diff
@@ -972,50 +978,56 @@ def update_font(self):
...
-                    #    finfo.run_code_analysis(
-                    #            self.get_option('code_analysis/pyflakes'),
-                    #            checked)
-                    pass
-        CONF.set('editor', conf_name, checked)
+        if method:
+            if self.editorstacks:
+                for editorstack in self.editorstacks:
+                    try:
+                        editorstack.__getattribute__(method)(checked)
```
goanpeca: Perhaps using getatrr is "better" ?  (See above)

```python
method = getattr(editorstack, method_name)
method(checked)
```

spyder/plugins/editor/widgets/editor.py
```diff
@@ -889,17 +889,17 @@ def set_bookmark(self, slot_num):
     def inspect_current_object(self):
         """Inspect current object in the Help plugin"""
         editor = self.get_current_editor()
-        editor.sig_display_signature.connect(self.display_signature_help)
+        editor.sig_display_object_info.connect(self.display_help)
```
goanpeca: This may affect the Hover Hints PR

ccordoba12: I agree but the signal name was misleading.

ccordoba12:
@goanpeca, please give it another review. It should be ready.

spyder/plugins/editor/plugin.py
```diff
@@ -866,85 +866,114 @@ def get_plugin_actions(self):
...
+                self.print_action,
+                self.save_all_action,
+                gotoline_action,
+                workdir_action,
+                self.close_action,
+                self.close_all_action,
+                self.toggle_comment_action,
+                self.revert_action,
+                self.indent_action,
+                self.unindent_action
```
goanpeca: I suggest we leave the trailing comma on the last element as well.

This allows for cleaner DIFFs when having to add new elements. Something like this:
```python
...
                self.indent_action,
                self.unindent_action'
            ]
```

goanpeca: This applies to all the other structures above @steff456

ccordoba12: I don't agree with this. We don't follow that convention and it seems odd to me too (I've never seen it before).

CAM-Gerlach: I strongly agree; in fact, I think we should adopt the consistent convention for new/sufficiently modified code to write non-inline collections of three or more elements that each fit on one line in the general form [recommended by PEP 8](https://www.python.org/dev/peps/pep-0008/#id17)

```python
my_list = [
    item_one,
    item_two,
    item_three,
    ]
```

(i.e. opening/closing braces not on the same line as the first/last item, and commas after every tiem) given we're currently all over the place in that regard, and this form produces the most readable diffs and (IMO) code as well. I personally prefer ``]`` aligned with the first non-whitespace line of the last list item, as it is more consistent with the other indentation rules and keeps the scope within the assignment as it should be, but PEP 8 also allows the more "traditional" style so I'm not going to raise a fuss about it (though it would be nice to pick one and be consistent).

> self.unindent_action'

I assume you mean ``,``, not ``'``, @goanpeca :)

ccordoba12: I'm strongly -1 on this, really. It's a purely stylistic stuff that we shouldn't enforce. If PRs authors like to use it, I'm fine with it, but it's up to each one.

So, if there's no more things to address, I'd like to merge this one. @goanpeca?

CAM-Gerlach: > We don't follow that convention

We don't, but we should. Its [recommended by PEP 8](https://www.python.org/dev/peps/pep-0008/#id29) and the [Google styleguide](https://google.github.io/styleguide/pyguide.html#341-trailing-commas-in-sequences-of-items) among others, and it both plays better with VCSes and avoids silly errors when re-ordering, adding or removing elements from lists.

> (I've never seen it before).

You've never read PEP 8? :) Its used throughout all of the examples there.

CAM-Gerlach: > It's a purely stylistic stuff that we shouldn't enforce.

So is the rest of PEP 8. This in particular has *objective* benefits with both cleaner diffs and avoiding errors.

ccordoba12: > You've never read PEP 8? :)

I meant I've never seen it used in the lots of projects I read code from.

ccordoba12: > So is the rest of PEP 8

We don't follow Pep 8 to the letter, and this is not even suggested by Pep8speaks.

CAM-Gerlach: > this is not even suggested by Pep8speaks.

Because PEP8Speaks only enforces the strict requirements of PEP 8, and not the recommended things (like 4-space indents, which we *do* enforce).

goanpeca: Many projects do use it and many styling tools enforce it as part of their rules. There are also blogpost on why this makes cleaner diffs and easier to read code/cleaner code. Also @CAM-Gerlach has pointed out official resources, that recommend that as well.

Maybe it is a discussion for another thread or separate issue.

---

I have advocated in the past (and I still do) to use tools like https://github.com/google/yapf so we do not have to worry too much about these discussions, commit to a style almost all feel comfortable with (as it is impossible to get a perfect style that fits all of the needs of every developer). It is much easier to run `yapf <file-name>.py` before every commit (or as part of CI) and use the time in coding. Also we can enforce this for every other project we maintain.

>The ultimate goal is that the code YAPF produces is as good as the code that a programmer would write if they were following the style guide. It takes away some of the drudgery of maintaining your code.

CAM-Gerlach: > We don't follow Pep 8 to the letter

Other than line length in ``icon_manager.py`` which is explicitly justified, I'm not sure where we don't follow a non-optional PEP 8 convention in new code, and our [Style Guide](https://github.com/spyder-ide/spyder/wiki/Dev%3A-Coding-Style), which you approved, says ``Unless otherwise specified, follow PEP 8 specifications and recommendations for coding style`` (and the only PEP 8 exception noted is retaining PyQt method names, which we have to do to override them).

Furthermore, this is orthogonal to my point, which was that we *do* enforce *lots* of other ``purely stylistic stuff``, i.e. anything we do enforce from PEP 8 (it is a *style* guide after all).

ccordoba12: > Because PEP8Speaks only enforces the strict requirements of PEP 8, and not the recommended things

Exactly, this is just a recommendation, not another thing we should start enforcing.

goanpeca: Sure we can merge @ccordoba12.

---

(Still auto formatting code is a discussion we should be having and hopefully implementing. See https://github.com/spyder-ide/spyder/issues/9272)

CAM-Gerlach: We already do enforce other "recommendations"/"suggestions"/"optional" elements in PEP 8, including using ``()`` for continuations, 4-space indents, 8 spaces for hanging indents of ``if`` statements, etc, and again our style guide explicitly says ``follow PEP 8 specifications and recommendations for coding style``, and this one in particular has objective advantages outside of just style.

goanpeca:
Thans @steff456 for working on this :-) !

PR #9175, opened 2019-04-18
PR: Remove sig_lsp_notification and improve introspection tests

ccordoba12:
- That signal was passed from the Editor to LSP clients, which is not how signals are used in Qt.
- I also simplified and improved several tests due to that.
- Now all introspection tests are running on Windows!
- I added a new test dependency called `pytest-ordering` to be able to run tests that use a new fixture called `lsp_codeeditor` at the beginning of our test suite because those tests generate focus errors on Linux if allowed to run at any point.

Fixes #9173.

PR #9186, opened 2019-04-22
PR: Use a single LSP manager instance per module in our tests

ccordoba12:
This PR adds a new fixture called `lsp_manager` to avoid creating an LSP client per test (which is very costly). Now we create a client per module and run all tests with it.

### DIFF
--- a/spyder/plugins/editor/lsp/manager.py
+++ b/spyder/plugins/editor/lsp/manager.py
@@ -15,7 +15,7 @@ import os
 import os.path as osp
 
 # Third-party imports
-from qtpy.QtCore import QObject, Signal, Slot
+from qtpy.QtCore import QObject, Slot
 
 # Local imports
 from spyder.config.base import get_conf_path, running_under_pytest
@@ -35,9 +35,6 @@ class LSPManager(QObject):
     RUNNING = 'running'
     CONF_SECTION = 'lsp-server'
     LOCALHOST = ['127.0.0.1', 'localhost']
-
-    # For testing
-    sig_initialize = Signal(dict, str)
 
     def __init__(self, parent):
         QObject.__init__(self)
@@ -167,9 +164,6 @@ class LSPManager(QObject):
                 if self.main and self.main.editor:
                     language_client['instance'].sig_initialize.connect(
                         self.main.editor.register_lsp_server_settings)
-                elif running_under_pytest():
-                    language_client['instance'].sig_initialize.connect(
-                        self.sig_initialize.emit)
 
                 logger.info("Starting LSP client for {}...".format(language))
                 language_client['instance'].start()
--- a/spyder/plugins/editor/lsp/manager.py
+++ b/spyder/plugins/editor/lsp/manager.py
@@ -15,7 +15,7 @@ import os
 import os.path as osp
 
 # Third-party imports
-from qtpy.QtCore import QObject, Slot
+from qtpy.QtCore import QObject, Signal, Slot
 
 # Local imports
 from spyder.config.base import get_conf_path, running_under_pytest
@@ -35,6 +35,9 @@ class LSPManager(QObject):
     RUNNING = 'running'
     CONF_SECTION = 'lsp-server'
     LOCALHOST = ['127.0.0.1', 'localhost']
+
+    # For testing
+    sig_initialize = Signal(dict, str)
 
     def __init__(self, parent):
         QObject.__init__(self)
@@ -164,6 +167,9 @@ class LSPManager(QObject):
                 if self.main and self.main.editor:
                     language_client['instance'].sig_initialize.connect(
                         self.main.editor.register_lsp_server_settings)
+                elif running_under_pytest():
+                    language_client['instance'].sig_initialize.connect(
+                        self.sig_initialize.emit)
 
                 logger.info("Starting LSP client for {}...".format(language))
                 language_client['instance'].start()
--- a/spyder/plugins/editor/lsp/tests/fixtures.py
+++ b/spyder/plugins/editor/lsp/tests/fixtures.py
@@ -1,0 +1,48 @@
+# -*- coding: utf-8 -*-
+
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+
+import os
+
+import pytest
+from pytestqt.plugin import QtBot
+
+from spyder.config.main import CONF
+from spyder.plugins.editor.lsp import SERVER_CAPABILITES
+from spyder.plugins.editor.lsp.manager import LSPManager
+
+
+@pytest.fixture(scope="module")
+def qtbot_module(qapp, request):
+    """Module fixture for qtbot."""
+    result = QtBot(request)
+    return result
+
+
+@pytest.fixture(scope="module")
+def lsp_manager(qtbot_module):
+    """Create an LSP manager instance."""
+    # Activate pycodestyle and pydocstyle
+    CONF.set('lsp-server', 'pycodestyle', True)
+    CONF.set('lsp-server', 'pydocstyle', True)
+
+    # Create the manager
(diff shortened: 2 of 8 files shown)


spyder/plugins/editor/widgets/tests/fixtures.py
```diff
@@ -68,19 +47,42 @@ def setup_editor(qtbot):
...
-    lsp_manager = LSPManager(parent=None)
+    with qtbot_module.waitSignal(manager.sig_initialize, timeout=30000) as blocker:
+        manager.start_client('python')
 
+    settings, language = blocker.args
+    manager.clients[language]['server_settings'] = settings
+
+    yield manager
+
+    # Tear down operations
```
goanpeca: I suggest we use the `request.addfinalizer(fin)` syntax as it makes the fixture code more clear and as they state in the docs 

https://docs.pytest.org/en/latest/fixture.html

>Both yield and addfinalizer methods work similarly by calling their code after the test ends, but addfinalizer has two key differences over yield:

>It is possible to register multiple finalizer functions.


>Finalizers will always be called regardless if the fixture setup code raises an exception. This is handy to properly close all resources created by a fixture even if one of them fails to be created/acquired:



Example:
```python
# content of conftest.py
import smtplib
import pytest


@pytest.fixture(scope="module")
def smtp_connection(request):
    smtp_connection = smtplib.SMTP("smtp.gmail.com", 587, timeout=5)

    def fin():
        print("teardown smtp_connection")
        smtp_connection.close()

    request.addfinalizer(fin)
    return smtp_connection  # provide the fixture value
```

spyder/plugins/editor/widgets/tests/fixtures.py
```diff
@@ -68,19 +47,42 @@ def setup_editor(qtbot):
...
+    manager.shutdown()
+    os.environ['SPY_TEST_USE_INTROSPECTION'] = 'False'
+    CONF.set('lsp-server', 'pycodestyle', False)
+    CONF.set('lsp-server', 'pydocstyle', False)
```
goanpeca: Probably work for another PR, but I suggest we run the tests in a separate config that should be reset to defaults after each test (if using such fixture) or one that does not reset for module level tests.

spyder/plugins/editor/widgets/tests/fixtures.py
```diff
@@ -89,31 +91,22 @@ def lsp_codeeditor(qtbot):
...
         editor.document_did_open()
 
     yield editor
 
-    # Tear down operations
-    os.environ['SPY_TEST_USE_INTROSPECTION'] = 'False'
-    CONF.set('lsp-server', 'pycodestyle', False)
-    CONF.set('lsp-server', 'pydocstyle', False)
-    lsp_manager.shutdown()
+    # Teardown operations
```
goanpeca: See above

spyder/plugins/editor/widgets/tests/fixtures.py
ccordoba12: Thanks for the suggestion. I added it to fixture.

spyder/plugins/editor/widgets/tests/fixtures.py
ccordoba12: I agree, but I don't know how we could do it.

spyder/plugins/editor/widgets/tests/fixtures.py
ccordoba12: Done.

ccordoba12:
@goanpeca, this should be ready. Please take another look at it.

spyder/plugins/editor/lsp/manager.py
```diff
@@ -36,6 +36,9 @@ class LSPManager(QObject):
     CONF_SECTION = 'lsp-server'
     LOCALHOST = ['127.0.0.1', 'localhost']
 
+    # For testing
+    sig_initialize = Signal(dict, str)
```
goanpeca: On Python 2 are there any issues on using `str` vs `unicode` ? Should we just use `object` ?

spyder/plugins/editor/lsp/manager.py
```diff
@@ -164,6 +167,9 @@ def start_client(self, language):
                 if self.main and self.main.editor:
                     language_client['instance'].sig_initialize.connect(
                         self.main.editor.register_lsp_server_settings)
+                elif running_under_pytest():
```
goanpeca: This was created for the tests, but I see no "harm" in leaving it always connected?

spyder/plugins/editor/widgets/tests/test_warnings.py
```diff
@@ -89,10 +123,10 @@ def test_move_warnings(qtbot, lsp_codeeditor):
 
 
 @pytest.mark.slow
-@pytest.mark.first
+@pytest.mark.second
```
goanpeca: @ccordoba12 what does this decorator refer to?

spyder/plugins/editor/lsp/manager.py
ccordoba12: No problem here, because the `str` is the language name we created an LSP client for. And programming languages don't have unicode names.

spyder/plugins/editor/lsp/manager.py
ccordoba12: I think it's more elegant to use a Mock for the editor plugin and main window in `lsp_manager`. That would avoid this.

spyder/plugins/editor/widgets/tests/test_warnings.py
ccordoba12: It comes from `pytest-ordering`, a new dependency I added in PR #9175. It tells `pytest` to run these tests after the ones in `test_introspection`, which are marked with `@pytest.mark.first`. The thing is I noticed LSP tests tend to introduce odd problems among each other and with other tests if they are not run in this particular order (especially on Linux).

spyder/plugins/editor/lsp/manager.py
ccordoba12: Let me address this and I'll let you know when it's ready for another review.

spyder/plugins/editor/widgets/tests/test_warnings.py
goanpeca: Interesting! Thanks for the explanation

spyder/plugins/editor/lsp/manager.py
ccordoba12: I address this in my last commit. Please review again.

ccordoba12:
@goanpeca, do you want to take a last look at this or should I merge?

goanpeca:
Looking now

spyder/plugins/editor/lsp/tests/fixtures.py
```diff
@@ -0,0 +1,87 @@
...
+        if attr == 'editor':
+            return self.editor
+        elif attr == 'projects':
+            # TODO: Add tests for project switching
+            return None
+        else:
+            return Mock()
+
+
+@pytest.fixture(scope="module")
```
goanpeca: Is there any reason we need this? seems to only be used on the fixture below?

goanpeca: I mean why not add all on `lsp_manager` instead of having `qtbot_module` and `lsp_manager`

ccordoba12: Because `qtbot` is a `scope=function` fixture, so you can't create a `scope=module` fixture with it (`pytest` rules forbid it). So we have to redefine it as a `module` fixture to be able to catch the Qt signal that tells us that an LSP client is up and ready.

Besides, this fixture is also used by `lsp_codeeditor` now.

goanpeca: Ok, thanks for the explanation

goanpeca:
LGTM!

Issue #9187, opened 2019-04-22
Define Hint behavior


goanpeca:
Currently the hints on spyder display the full information of the function etc.

We need to define what teh behavior of this should be to align the calltips on editors an consoles.

Also we need to define if the behavior will work on hover or with shortcut?

Some screenshots of what VSCode displays.

Hint on imports

<img width="754" alt="Screen Shot 2019-04-22 at 08 51 45" src="https://user-images.githubusercontent.com/3627835/56503704-e68b1200-64db-11e9-87f2-c7e237c52fa3.png">

Hint on funcs
<img width="712" alt="Screen Shot 2019-04-22 at 08 51 53" src="https://user-images.githubusercontent.com/3627835/56503709-ebe85c80-64db-11e9-8a13-69c42830ce22.png">

goanpeca:
![Selección_033](https://user-images.githubusercontent.com/365293/56197843-fd43eb80-6039-11e9-919f-468df435439e.png)

PR #9133, opened 2019-04-08
PR: Create generic switcher

goanpeca:
## Description of Changes

* [x] Wrote at least one-line docstrings (for any new functions)
* [x] Added unit test(s) covering the changes (if testable)
* [x] Included a screenshot (if affecting the UI)

switcher

s

### Issue(s) Resolved

Fixes #

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct: @goanpeca

PR #9191, opened 2019-04-23
PR: Enable hover hints

goanpeca:
## Description of Changes

Tasks:
- [x] Enable got to help for hover hints
- [x]  Add a delay (500 ms) to hint popup
- [x]  Make hints disappear on clicking anywhere in the UI (not just on the hint), clicking any key, or moving the mouse at least a few px away from the boundary of the hovered-over word
- [x]  Syntax-highlight signature like in calltip
- [x]  Only show ~8-10 lines max, and only of description, not Parameters/:params: section, to avoid taking up too much space.
* [x] Hints appear above all other windows, even other applications, as well as on top of code completion. They should only appear on top of Spyder's window, and code completion popups should be layered above them since if the user triggers it, they basically always want to see that with a higher priority.
* [x] The cursor doesn't change to the pointer finger as is expected when hovering over a link, though that may be due to it not being implemented yet. 
* [x] To also help reduce those concerns, I'd favor bumping the delay a tad to 500 ms, to help reduce the chance of the hint popping up when undesired, such as briefly hovering over a function name when moving the cursor around 
* [x] Add config option

### Issue(s) Resolved

Fixes https://github.com/spyder-ide/spyder/issues/9187

### Affirmation


goanpeca:
This needs https://github.com/spyder-ide/spyder/pull/9133 merged

goanpeca:
Screen Shot 2019-04-23 at 00 05 12

ccordoba12:
That looks really nice, thanks!

goanpeca:
@ccordoba12 how do I request the info so that the help pane is opened?

ccordoba12:
> how do I request the info so that the help pane is opened?

Do you want to automatically show it after the link is pressed?

ccordoba12:
I have a question already:

Is the link also showed on signatures? Or is it working only for hints?

goanpeca:
>Do you want to automatically show it after the link is pressed?

Yes

I suggest that we add the text that says

`Click here or Press Ctrl+I for more info` (if Ctrl+I is pressed we check if the hint is visible and if it is we move the cursos to the word and execute the info) If the user clicks we hide the hint, do not move cursor and display info

>Is the link also showed on signatures? Or is it working only for hints?

For now hints, but could be added to calltips as well

ccordoba12:
> > Do you want to automatically show it after the link is pressed?
>
> Yes

The `show_object_info` method of CodeEditor and ShellWidget should take care of that (although it seems Edgar removed the CodeEditor connection, so should rework that).

> Click here or Press Ctrl+I for more info

Ok, good suggestion.

> For now hints, but could be added to calltips as well

I'd like to see it in both. @CAM-Gerlach was also asking for something similar, I think (i.e. displaying more info in the signature).

CAM-Gerlach:
I like the popup hints overall, thanks @goanpeca . However, the delay before popping up the hint needs to be much longer IMO, and it needs to be able to be configurable or at least able to be disabled (i.e. an option in prefs). Just dragging my cursor over a file in the Editor makes them pop up like crazy, and its extremely annoying as a user having them constantly block my code and have to be clicked off of. The problem with the link is since it means we can't just move the cursor off the current object to dismiss it, making it  much less fluid to move from one hint to another, or from a hint to no hint, and requires a significantly longer popup delay for a decent UX (since it requires more active work to make the hint disappear) as well as requires the user to discover how to close them rather than just doing so naturally. 

Furthermore, it is layered above the completion popup, and thus blocks being able to view it (as it also doesn't dismiss on clicking anywhere not on the Hint, or starting to type, even on a different line).

> instead of adding a button, we could add a link like this.

I assume it's not supposed to work yet. In any case, I just don't see the link as valuable in the real world and would personally never use it, because pressing the shortcut is just way faster and more foolproof (as it requires a relatively precise mouse movement and click. This is especially true I have ``Inspect current object`` bound to a far more sensible shortcut, ``F1``, so its a single tap away with my non-mouse hand rather than an awkward 2-handed keypress (especially if using the mouse). Coupled with the downsides above, I would favor going back to the ``Press {shortcut} for more help`` approach we used previously.

With regard to content, we should just include up to ~8 lines of the prose description section (up to the ``Parameters``/``:params:`` section), and syntax-highlight the signature like we do in the calltips.

goanpeca:
>I like the popup hints overall, thanks @goanpeca . However, the delay before popping up the hint needs to be much longer IMO, and it needs to be able to be configurable or at least able to be disabled (i.e. an option in prefs). Just dragging my cursor over a file in the Editor makes them pop up like crazy, and its extremely annoying as a user having them constantly block my code and have to be clicked off of. The problem with the link is since it means we can't just move the cursor off the current object to dismiss it, making it much less fluid to move from one hint to another, or from a hint to no hint, and requires a significantly longer popup delay for a decent UX (since it requires more active work to make the hint disappear) as well as requires the user to discover how to close them rather than just doing so naturally.

>I assume it's not supposed to work yet. In any case, I just don't see the link as valuable in the real world and would personally never use it, because pressing the shortcut is just way faster and more foolproof (as it requires a relatively precise mouse movement and click. This is especially true I have Inspect current object bound to a far more sensible shortcut, F1, so its a single tap away with my non-mouse hand rather than an awkward 2-handed keypress (especially if using the mouse). Coupled with the downsides above, I would favor going back to the Press {shortcut} for more help approach we used previously.

Thanks for the feedback. Yes this is not yet ready and indeed we need to add an artificial delay. Working on that.

I agree on the `F1` idea, except that is not good for mac, as mac users do not have functions keys exposed by default. So there a `Cmd+I` is probably better (as it is now)?

@ccordoba12 thoughts on @CAM-Gerlach comments?

CAM-Gerlach:
> I agree on the F1 idea,

I didn't really mean that as a formal suggestion; I was just mentioning that was what I always use personally (and others may as well), making opening help way faster and more efficient. However, I wouldn't mind it, of course, its also what many of the other scientificIDEs use (e.g. Mathematica and Rstudio use it; Matlab uses ``Ctrl-F1``). I typically switch Spyder Documentation to ``Ctrl-F1``; I've also customized a lot of other shortcuts for speed and effic

> except that is not good for mac, as mac users do not have functions keys exposed by default

Its still a two-hand, two-button shortcut like ``Ctrl-I``. And don't a ton of other extremely commonly-used default shortcuts, like the Run and Debug ones, use F-keys by default on Mac as well (I change all of them to <Various-modifiers>+Enter myself, for speed and cohesion; the defaults are pretty crazy IMO)?

Distilling my other suggestions into bullets for readability:

* Add a delay (500-1000ish ms?) to hint popup
* Remove not-useful link from hint, only the button (and don't hardcode it)
* Make hints disappear on clicking anywhere in the UI (not just on the hint), clicking any key, or moving the mouse at least a few px away from the boundary of the hovered-over word
* Make hints always appear *below* completion popups (shouldn't be an issue with the above)
* Syntax-highlight signature like in calltip
* Only show ~8-10 lines max, and only of description, not ``Parameters``/``:params:`` section, to avoid taking up too much space.

goanpeca:
Thanks again @CAM-Gerlach. Regarding `Only show ~8-10 lines max, and only of description, not Parameters/:params: section, to avoid taking up too much space.` not sure as this comes from the LSP client.

CAM-Gerlach:
We can't just post-process it on the Spyder side, before (or as) we sent the text to the widget for display?

goanpeca:
I guess :-p

ccordoba12:
> Distilling my other suggestions into bullets for readability

These are good suggestions, thanks!

> Add a delay (500-1000ish ms?) to hint popup

This should be 300ms, 500 tops. More than that and users would assume to hints are not working.

> Remove not-useful link from hint, only the button (and don't hardcode it)

I don't agree with this. I think the link is very useful and I'd like it to stay.

CAM-Gerlach:
> This should be 300ms, 500 tops. More than that and users would assume to hints are not working.

Okay, we can try that.

> I don't agree with this. I think the link is very useful and I'd like it to stay.

I'm not sure I understand the specific use case where having to precisely mouse to and click some text would be faster than tapping a single key. At the very least, we should move the link to be directly to the right of the ``Hint`` title, so it is as close as possible to the user's cursor when its opened, and which would also allow us to only make the moving the cursor over that top line keep the Hint onscreen, and moving it more than a short distance lower (or anywhere else not on either that bar or the original object name) dismiss it.

In the future, we could add other links there, e.g. to the Numpy/Scipy/Pandas/Matplotlib docs for that function, for a Google search (both very common for me, to find out more than just what Spyder's help shows or if it doesn't work), other services, etc.

goanpeca:
Styling issues apart, we could do something like this. This way the link is close to the user as @CAM-Gerlach suggested:

Screen Shot 2019-04-24 at 09 15 00

goanpeca:
This is not ready @CAM-Gerlach , but I added the delay logic and some nice changes.

Could you give it a try?

spyder/widgets/mixins.py
```diff
@@ -130,18 +131,28 @@ def _update_stylesheet(self, widget):
...
         This will display title and text as separate sections and add `...`
         if `ellide` is True and the text is too long.
         """
+        shortcut = 'Ctrl+I'  # FIXME:
+        help_link = ''
+        if help_button:
+            help_link = (
+            '<a href="{0}"><small>'
+            '<span style="float:right;color:white;">'
+            '(For more information, click here or press {1})'
```
CAM-Gerlach: Before, the shortcut was formatted as a <kbd>keyboard button</kbd> (IIRC), which looked quite nice. Could you add ``<kbd>`` tags (or whatever else was used before)?

ccordoba12:
> It would be a little more succinct if we said Click here or press {shortcut} for more

I agree with this

> Also, I suggest offsetting the position of the dialog by perhaps ~80 px to the left from the function start

@goanpeca, please take a look at where VSCode puts the hint to follow its lead on that.

> It appears on both the Hint and on the calltip, making the latter extremely long

For calltips I suggest to readd the title `Signature` to them and put the link there.

----

My only suggestion so far is to align the link to the right of hints/calltips to be more visually distinct from their titles.

ccordoba12:
@goanpeca, please also fix the tests before proceeding further, to see if things are working correctly so far.

goanpeca:
>I agree with this

Yep


>@goanpeca, please take a look at where VSCode puts the hint to follow its lead on that.

Currently it is like vscode, no matter where the word hovers, the dialog is displayed aligned to the left of the word.

>For calltips I suggest to readd the title Signature to them and put the link there.

Ok

>My only suggestion so far is to align the link to the right of hints/calltips to be more visually distinct from their titles.

Yes, that is why I said Styling issues appart ;-), will work on that

>please also fix the tests before proceeding further, to see if things are working correctly so far.

Sure, that will be that last step

goanpeca:
@ccordoba12 @CAM-Gerlach how about this

Screen Shot 2019-04-24 at 11 04 05
?

ccordoba12:
> I actually like the extra line space the hint gives, feels a bit like some "breath space"/less crowded

I don't understand what you mean by this.

> If we really need it styled like that, we will need to reimplement the ToolTipWidget cause a QLabel does not support complex styling within its text content. We could do this, but beyond the scope of this PR.

Yeah, and also too much work for a such a simple thing. -1 on that.

goanpeca:
>I don't understand what you mean by this.

I used two lines instead of having Hint and the click here on the same line. I like it like that cause it feels more clean (the way ythe screnshot looks)

Now the issue I need some help. How can I get the shortcut since this is configurable?

CAM-Gerlach:
> The offset to the left looks odd, so I would just leave it like it is

Okay, that's alright then.

> For calltips I suggest to readd the title Signature to them and put the link there.

As I've already mentioned above, IMO, that doesn't add anything useful, just extra words and clutter to the calltip that detract from its core functionality—allowing users to quickly see the function's signature and purpose *as they are typing it*. Unlike the ``Hints``, where a link makes sense if positioned directly below the cursor position, since the user is already using their mouse to call up the hint and can merely move it down a few mm to click the link, in order to summon a calltip the user *must* have their hands on the keyboard (to type the ``(`` or ``,`` that triggers it), so in order for them to use the link (as opposed to the shortcut), they need to:

1. Move their hand to their mouse
2. move their mouse to the link
3. click the link,
4. move their hands back to their keyboard,
5. resummon the calltip
6. proceed with typing.

I don't understand in what logical scenario that would make more sense than tapping a single keystroke to summon help, particularly relative to the verbosity and clutter it adds to an otherwise concise, quick to read calltip.

> My only suggestion so far is to align the link to the right of hints/calltips to be more visually distinct from their titles.

While visually this would look better, UX should supersede aesthetics, and (as can be seen in @goanpeca ;s screenshots) the user would have to move the cursor many times further to click the link. 

How about instead we just drop the rather redundant ``Hint`` title entirely (VSCode doesn't have it, after all) and just put the link (without the parens) in its place? Then it will always be in the same spot under the cursor, and be on its own visual distinct section like @ccordoba12 mentions, as well as leaves somewhat more whitespace like @goanpeca likes (which we could later use for more links, if desired)? 

Alternatively, we could go with ``Hint  |  Click here...`` but I like the first idea better; I think it satisfies each of our concerns more fully and looks cleaner.

CAM-Gerlach:
> If we really need it styled like that

Alright, sure; that's okay I guess,

> I used two lines instead of having Hint and the click here on the same line. I like it like that cause it feels more clean (the way ythe screnshot looks)

It does look a little less crowded, but IMO I don't think the tradeoffs are worth it—it makes the link further away and harder to click, and the hint takes up even more space and obstructs 1-2 more lines of user code. See my suggestion above for a way to avoid these tradeoffs but still reduce clutter/increase whitespace.

goanpeca:
I agree with @CAM-Gerlach concerns @ccordoba12, the link on call signatures makes little, only the shortcut. Also this line should probably go on the bottom for Calltips and on top for hover hints to address @CAM-Gerlach minimal moving suggestion. I would argue that:

- For hover hints only clicking makes sense 
- For calltips only shortcut makes sense.
- Titles in any of the tooltip seem unnecessary at this point

ccordoba12:
> Also this line should probably go on the bottom for Calltips and on top for hover hints

I don't agree with that. It should go on the same place for both (I'd suggest on the bottom).

> For hover hints only clicking makes sense

Agreed.

> For calltips only shortcut makes sense.

Agreed too.

> Titles in any of the tooltip seem unnecessary at this point

Sure, but we should leave the `Code analysis` title available. I think that one is important.

CAM-Gerlach:
> I don't agree with that. It should go on the same place for both (I'd suggest on the bottom).

If we are going to have the link, it needs to be on the top for at least the hover hints as we've already discussed, otherwise the user has to waste time moving their mouse all the way down to the bottom and being careful not to overshoot or they have to repeat the whole process. It makes more sense on the bottom of calltips, IMO, since it allows the first thing the user reads to be just the information they actually want (the function signature), then followed by how to get more info if needed.

While consistency would be nice, I agree, ``a foolish consistency is the hobgoblin of little minds`` to quote PEP 8. In this case, the two items aren't really directly parallel, since one is a link to click, while the other is a shortcut reminder, and will already be styled/formatted differently and require different user actions (and thus have different motivating reasons to be where they are). If they were both links, then it would 100% make sense to put them in the same spot since users have to navigate to them.

> Sure, but we should leave the Code analysis title available. I think that one is important.

I don't mind having that one either; it makes the popup clearer and more polished IMO.

goanpeca:
I will post some screenshots of the options. Thanks for the feedback guys :-)

goanpeca:
Ok, I had to rework a bit of logic to be able to handle this.

So far it looks like this:

# Calls, only the text after the signature is cropped
Screen Shot 2019-04-24 at 15 50 52

# Rest, text is cropped
Screen Shot 2019-04-24 at 15 51 02

ccordoba12:
@goanpeca, I like your last iteration above, so I think we should go with it.

> If we are going to have the link, it needs to be on the top for at least the hover hints as we've already discussed, otherwise the user has to waste time moving their mouse all the way down to the bottom

Both calltip and hint show almost the same info, so it'd be *really* odd to show the link on top for for one and bottom for the other. So I'm -1 on this. If users complain then we can revise our decision, but for now let's go with bottom for both.

CAM-Gerlach:
> Both calltip and hint show almost the same info so it'd be really odd to show the link on top for one and bottom for the other.

One is a clickable link, and one is a reminder of the keyboard shortcut, and serve quite different practical UI purposes. Therefore, while it is still not ideal to have in different places, the difference is explicitly motivated by practical usability trumping "a foolish consistency". Putting both at the top makes the calltip harder to read at a glance, while putting both at the bottom requires users to move their mouse a large distance very precisely, or risk the frustration of dismissing the hint entirely.

Thinking orthogonally again for a solution that will better satisfy both of our concerns, how about we keep both on the bottom as you request, but for the Hints, instead of making only the bottom bar clickable, we make clicking anywhere on the calltip summon the help pane and tweak the text to read ``Click this popup for additional help``? This is even simpler for the user to trigger than clicking a specific line, ensures the calltip is still easy to read, and keeps both in a consistent, logical location.

goanpeca:
> Click this popup for additional help? This is even simpler for the user to trigger than clicking a specific line, ensures the calltip is still easy to read, and keeps both in a consistent, logical location.

@ccordoba12 thoughts?

ccordoba12:
> but for the Hints, instead of making only the bottom bar clickable, we make clicking anywhere on the calltip summon the help pane

That breaks the expectation that clicking on the hint should dismiss it (as it also happens to the signature).

@CAM-Gerlach, sorry but we need to keep moving with this. I really appreciate your input, but let's go with the last design and see how other people think of it. If they also find things cumbersome, then we'll explore some of your proposed alternatives.

goanpeca:
@ccordoba12 I am leaving the actual `click handling` for the next PR as well as getting the info to show correctly when called inside a calltip

CAM-Gerlach:
> That breaks the expectation that clicking on the hint should dismiss it

I'm not sure where that is established as an expectation; in my UI experience, clicking *outside* the popup is what is expected to dismiss it. But if that *is* the case, clicking a box at the very end of the popup (your current proposal) would *also* break the same expectation. Furthermore, if that's what we're implementing, clicking *anywhere* but the exact target area for the link will close the popup instead of getting more info, requiring users to go through the whole process again: a sure way of frustrating our users even further.

goanpeca:
As extra input,  VSCode does not dismiss the dialog after click. In fact it lets you copy content if needed. In the case of hover y will be dismissed if moving outside the word of interest

CAM-Gerlach:
> As extra input, VSCode does not dismiss the dialog after click. In fact it lets you copy content if needed. 

In my testing just now, this is also precisely how your implementation works. Quite handy.

Therefore, I amend my above proposal to either trigger more help on single-click, and otherwise behave as current on double+ click and click and drag (for selection, etc). Or, we could trigger more help on right-click instead; there is currently a context menu but it only contains Copy and Select-All, which are also access able via the commonly-known shortcuts. There may also be a better idea than all of those we haen't thought of.

However, if we're delaying the actual implementation to another PR, at this point I'm okay with keeping the status quo for now so we can get this merged and then we can discuss this further in that PR.

ccordoba12:
@goanpeca, I'm going to fix the failures in Travis here because I know how to do it and they are unrelated to your PR.

goanpeca:
>@goanpeca, I'm going to fix the failures in Travis here because I know how to do it and they are unrelated to your PR.

Thanks @ccordoba12 , I had no idea what the issue was!

- I don't have many tests in mind for this, but please let me know.
- Do you think I should add a configuration option for this?

CAM-Gerlach:
> Do you think I should add a configuration option to enable/disable hover hints?

I would be strongly against merging this without one, given the [page after page after page](https://www.google.com/search?q=disable+hover+hints+IDE&ie=utf-8&oe=utf-8) of SO questions, Github issues and other Google results asking about how to disable them in other IDEs, and expressing strong annoyance and frustration if its not possible.

goanpeca:
Screen Shot 2019-04-28 at 18 47 15

@CAM-Gerlach @ccordoba12 added the config option

ccordoba12:
@goanpeca, there's one little problem with passing that option to the PyLS: it disables not only hints but also `Ctrl+I` functionality. That's not probably what users expect, so we'd need a more fine grained than this.

Thoughts @CAM-Gerlach?

goanpeca:
Ok will look into it.

Looking at this change I noticed that we coupled the other options directly to the pyls configuration, so I followed suit. However I think these options should not be tied to those options and should be handled in the code editor and not the pyls as they apply to all languages.

ccordoba12:
> Looking at this change I noticed that we coupled the other options directly to the pyls configuration, so I followed suit. However I think these options should not be tied to those options and should be handled in the code editor and not the pyls as they apply to all languages.

Each language is configured separately, so it depends on the server implementation if it allows to enable/disable hovers, linting, etc.

The options we provided only apply to the PyLS because a) we're focused on Python and b) the PyLS allows being configured like that.

goanpeca:
So this option is now on the editor settings.

Screen Shot 2019-04-29 at 12 44 14

Might be more intuitive on the Completion and linting page, but cross config setup requires another PR if is something we want to do eventually.

goanpeca:
@ccordoba12 I could not fin a straightforward way of dealing with the config in a plugin different from where the configuration is found. I have some ideas but it is out of the scope of this PR, I think we can merge and proceed with that in another configuration cleanup one

spyder/plugins/editor/confpage.py
```diff
@@ -39,6 +39,12 @@ def setup_page(self):
                 'show_class_func_dropdown')
         showindentguides_box = newcb(_("Show Indent Guides"),
                                      'indent_guides')
+        show_hover_hints_box = newcb(
+            _("Show hover hints"),
+            'show_hover_hints',
+            tip=_("If this option is enabled, hovering the mouse \n"
+                  "pointer over an object name will display a hint of\n"
+                  "that object's definition (if resolved)."))
```
ccordoba12: I think we should move this option to `lsp-server` and use it through CONF in CodeEditor.

It's just odd to have almost all options in "Completion and linting", but not this one.

spyder/plugins/editor/lsp/manager.py
```diff
@@ -300,6 +300,12 @@ def generate_python_config(self):
             'enabled': self.get_option('jedi_definition'),
             'follow_imports': self.get_option('jedi_definition/follow_imports')
         }
+        jedi_hover = {
+            # This option needs to be True always so that we can request
+            # information for the Object Inspection Help Pane
+            # TODO: Something similar might be needed for other languages 
+            'enabled': True,
+        }
```
ccordoba12: There's no need to have this here because this value is already `True` in `PYTHON_CONFIG`. But you can move the comment above there.

spyder/plugins/editor/lsp/manager.py
```diff
@@ -322,5 +328,6 @@ def generate_python_config(self):
         plugins['jedi_signature_help'] = jedi_signature_help
         plugins['preload']['modules'] = self.get_option('preload_modules')
         plugins['jedi_definition'] = jedi_definition
+        plugins['jedi_hover'] = jedi_hover
```
ccordoba12: This can be removed too.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -762,7 +818,6 @@ def parse_lsp_config(self, config):
         self.will_save_notify = sync_options['willSave']
         self.will_save_until_notify = sync_options['willSaveWaitUntil']
         self.save_include_text = sync_options['save']['includeText']
-        self.enable_hover = config['hoverProvider']
```
ccordoba12: This needs to be restored, I think. @andfoy?

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -875,38 +930,43 @@ def process_signatures(self, params):
 
                 signature = signature_data['label']
                 parameter = parameter_data['label']
-                parameter_documentation = parameter_data['documentation']
+                # parameter_documentation = parameter_data['documentation']
```
ccordoba12: Can this be removed?

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -875,38 +930,43 @@ def process_signatures(self, params):
...
     def handle_hover_response(self, contents):
         """Handle hover response."""
         try:
-            text = contents['params']
-            self.sig_display_signature.emit(text)
-            self.show_tooltip(_("Hint"), text)
+            content = contents['params']
+            self.sig_display_signature.emit(content)
+
+            if self._enable_hover_hints and content:
```
ccordoba12: According to my comment above, we should change this to

    if CONF.get('lsp-server', 'show_hints') and content:

spyder/widgets/mixins.py
```diff
@@ -130,66 +129,144 @@ def _update_stylesheet(self, widget):
...
+        shortcut_console = getattr(app, 'shortcut_inspect_console', None)
+        shortcut = shortcut_editor or shortcut_console
+
+        if shortcut:
+            key_sequence = shortcut.data[0].key()
+            if key_sequence:
+                value = key_sequence.toString()
+            if sys.platform == "darwin":
+                value = value.replace('Ctrl', 'Cmd')
+        return value
```
ccordoba12: This is very very convoluted. I think we should use `CONF` for now and later decide how to better handle this kind of use cases.

goanpeca: Fixed

goanpeca:
@ccordoba12 made the suggested fixes.

CAM-Gerlach:
Sorry I've been AWOL; I've been finishing up final projects and exams. I'll do a final test of everything now. Thanks @goanpeca and @ccordoba12 !

goanpeca:
No worries @CAM-Gerlach, thanks for testing this out and all the useful suggestions!

goanpeca:
@CAM-Gerlach I corrected the windows issues.

CAM-Gerlach:
Thanks @goanpeca , review in work now.

> except for the completion problems as they seem unrelated. 

Yes, I've strongly suspected the same. I just wanted to mention it as part of the continuing investigation into that issue.

spyder/widgets/mixins.py
```diff
@@ -601,6 +735,10 @@ def get_current_word(self, completion=False):
         if ret is not None:
             return ret[0]
 
+    def get_hover_word(self):
+        """"""
```
CAM-Gerlach: Either add a docstring, or remove this?

spyder/widgets/mixins.py
```diff
@@ -130,66 +130,134 @@ def _update_stylesheet(self, widget):
...
 
-        return tiptext
-
-    def _format_signature(self, signature, doc='', parameter='',
-                          parameter_doc='', color=_DEFAULT_TITLE_COLOR,
-                          is_python=False):
+        help_text = ''
+        if inspect_word:
+            if display_link:
+                help_text = 'Click for additional help'
```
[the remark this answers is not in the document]
CAM-Gerlach: And it shouldn't be formatted as a link anymore, just *italic* text.

spyder/preferences/languageserver.py
```diff
@@ -619,6 +619,12 @@ def setup_page(self):
         # Basic features group
         basic_features_group = QGroupBox(_("Basic features"))
         completion_box = newcb(_("Enable code completion"), 'code_completion')
+        enable_hover_hints_box = newcb(
+            _("Enable hover hints"),
+            'enable_hover_hints',
+            tip=_("If this option is enabled, hovering the mouse \n"
```
CAM-Gerlach: *If enabled, hovering...

spyder/preferences/languageserver.py
```diff
@@ -619,6 +619,12 @@ def setup_page(self):
...
+                  "pointer over an object name will display a hint of\n"
+                  "that object's definition (if resolved)."))
```
CAM-Gerlach: *will display that object's signature and/or docstring (if present).

spyder/widgets/calltip.py
```diff
@@ -67,6 +82,25 @@ def paintEvent(self, event):
 
         super(ToolTipWidget, self).paintEvent(event)
 
+    def _update_hover_html_link_style(self, url):
+        """"""
```
CAM-Gerlach: Missing docstring.

spyder/widgets/mixins.py
```diff
@@ -130,66 +130,134 @@ def _update_stylesheet(self, widget):
                 }}'''.format(name, background, border)
             widget.setStyleSheet(css + extra_css)
 
-    def _format_text(self, title, text, color, ellide=False):
+    def _get_inspect_shortcut(self):
+        """
+        Queries the configuration for the inspect shortcut defined by the
+        editor.
```
CAM-Gerlach: PEP 257 summary must be one line. We've been consistently enforcing this (even with Carlos)

``Queries the editor's config to get the current "Inspect" shortcut.``

spyder/widgets/mixins.py
```diff
@@ -130,66 +130,134 @@ def _update_stylesheet(self, widget):
...
+        if signature:
+            template += BASE_TEMPLATE.format(
+                font_family=font_family,
+                size=text_size,
+                color=text_color,
+                main_text=signature,
+            )
+
+        if not text:
+            text = '\n<i>No further documentation available</i>\n'
```
CAM-Gerlach: Remove ``further``

goanpeca:
I believe this one is good to go. Merging!

ArrowRaider:
These are really poor.  How do I disable them?

PR #9203, opened 2019-04-25
PR: Add option to connect to external PyLS servers

### LINKED ISSUE #9195: Cannot connect to an external PyLS server
## Problem Description

The new menu does not allow to connect to an external PyLS server. 

1. Run spyder-pyls server on one terminal with the command `python -m pyls -vvv --port 2089 --tcp`
2. Run spyder on another terminal
3. Try to connect with the running server by changing the port in Preferences > Completion and linting > Advanced 
4. After giving apply, there's no new connection received by the local pyls server.
* The same happens after trying to setup a new server for python.

Connect to an external PyLS server running locally.

## Versions

* Spyder version: Spyder 4.0.0.dev0 (Commit: ba26876de) 
* Python version: 3.6.8
* Qt version: 5.12.2
* PyQt version: 5.12.1 
* Operating System name/version: Darwin 18.2.0 

### Dependencies

```
sphinx >=0.6.6    :  1.8.5 (OK)
psutil >=0.3      :  5.6.1 (OK)
nbconvert >=4.0   :  5.4.1 (OK)
numpy >=1.7       :  1.16.2 (OK)
IPython >=4.0     :  7.4.0 (OK)
```

dalthviz:
## Description of Changes

imagen

Properly set the `external` property for the lsp-server

### Issue(s) Resolved

Fixes #9195 

### Affirmation

### DIFF
--- a/spyder/plugins/editor/lsp/manager.py
+++ b/spyder/plugins/editor/lsp/manager.py
@@ -301,14 +301,16 @@ class LSPManager(QObject):
             'follow_imports': self.get_option('jedi_definition/follow_imports')
         }
 
+        # Advanced
+        advanced_external = self.get_option('advanced/external')
+
         # Setup options in json
         self.python_config['cmd'] = cmd
         if host in self.LOCALHOST:
             self.python_config['args'] = '--host {host} --port {port} --tcp'
-            self.python_config['external'] = False
         else:
             self.python_config['args'] = ''
-            self.python_config['external'] = True
+        self.python_config['external'] = advanced_external
         self.python_config['host'] = host
         self.python_config['port'] = port
 
--- a/spyder/plugins/editor/lsp/manager.py
+++ b/spyder/plugins/editor/lsp/manager.py
@@ -82,7 +82,7 @@ class LSPManager(QObject):
         """Get language configuration options from our config system."""
         if language == 'python':
             self.update_python_config()
-            return self.python_config
+            return self.python_config.copy()
         else:
             return self.get_option(language)
 
--- a/spyder/plugins/editor/lsp/manager.py
+++ b/spyder/plugins/editor/lsp/manager.py
@@ -43,7 +43,6 @@ class LSPManager(QObject):
         self.clients = {}
         self.requests = {}
         self.register_queue = {}
-        self.python_config = PYTHON_CONFIG.copy()
 
         # Register languages to create clients for
         for language in self.get_languages():
@@ -81,8 +80,7 @@ class LSPManager(QObject):
     def get_language_config(self, language):
         """Get language configuration options from our config system."""
         if language == 'python':
-            self.update_python_config()
-            return self.python_config.copy()
+            return self.generate_python_config()
         else:
             return self.get_option(language)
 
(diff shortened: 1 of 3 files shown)


spyder/plugins/editor/lsp/manager.py
```diff
@@ -82,7 +82,7 @@ def get_language_config(self, language):
         """Get language configuration options from our config system."""
         if language == 'python':
             self.update_python_config()
-            return self.python_config
+            return self.python_config.copy()
```
ccordoba12: Why is this needed? I mean, it shouldn't be necessary.

spyder/plugins/editor/lsp/manager.py
```diff
@@ -301,14 +301,16 @@ def update_python_config(self):
             'follow_imports': self.get_option('jedi_definition/follow_imports')
         }
 
+        # Advanced
+        advanced_external = self.get_option('advanced/external')
```
ccordoba12: Let's change this to

    external_server = self.get_option('advanced/external')

spyder/plugins/editor/lsp/manager.py
```diff
@@ -301,14 +301,16 @@ def update_python_config(self):
...
+
         # Setup options in json
         self.python_config['cmd'] = cmd
         if host in self.LOCALHOST:
             self.python_config['args'] = '--host {host} --port {port} --tcp'
-            self.python_config['external'] = False
         else:
             self.python_config['args'] = ''
-            self.python_config['external'] = True
+        self.python_config['external'] = advanced_external
```
goanpeca: This line alone is the actual fix?

spyder/plugins/editor/lsp/manager.py
dalthviz: While testing the changes, the restart was not being triggered (the `current_config` and `new_config` in `update_server_list` were the same). This was caused because the client for `python` was being affected for the change in `self.python_config` (adding the `copy` prevents this to happen)

spyder/plugins/editor/lsp/manager.py
dalthviz: Yep, in this way we can handle an pyls instant launched outside Spyder (even if it's launched in the same machine where Spyder is running and besides the error in the restart of the client)

spyder/plugins/editor/lsp/manager.py
ccordoba12: Ok, I understand what you mean now and that's a serious problem because by using a shared `self.python_config` attribute we are modifying the configuration of the existing Python client and of future clients too.

To avoid that, please do the following:

1. Remove `self.python_config` from `LSPManager`
2. Rename `update_python_config` to `generate_python_config`
3. Use a local variable inside that method called `python_config`, like this:

    ```python
    def generate_python_config(self):
    """
    Generate Python server configuration with the options saved in our
    config system
    """
    python_config = PYTHON_CONFIG.copy()
    ...
    ...
    return python_config
    ```

    and use `python_config` instead of `self.python_config` in that method.

spyder/plugins/editor/lsp/manager.py
ccordoba12: You changed the names of other variables, but forgot to change precisely this name ;-)

ccordoba12:
@dalthviz, did you test that my last suggestion works as expected?

dalthviz:
@ccordoba12 yep, it works for me :+1:

PR #9082, opened 2019-04-02
PR: Use blocks to identify lines and add unique cell names

impact27:
## Description of Changes

* [x] Wrote at least one-line docstrings (for any new functions)
* [x] Added unit test(s) covering the changes (if testable)
* [ ] Included a screenshot (if affecting the UI)

- Use blocks instead of lines number to identify a line, so spyder can find the line of the number changes.
- Change default cell name from '' to 'Unnamed cell'
- Ensure cell names are unique by appending a #N at the end if the cell name is not
- Update the outline explorer automatically

### Issue(s) Resolved

Fixes #9085
Fixes #7976

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct:

PR #9219, opened 2019-04-27
PR: Highlight current entry in the Outline Explorer and update it on-the-fly

### LINKED ISSUE #885: Outline Explorer: Highlight current class/method automatically
_From [techtonik@gmail.com](https://code.google.com/u/techtonik@gmail.com/) on 2011-12-15T02:49:19Z_

It would be nice if OutlineExplorer could highlight class/function that corresponds to cursor location in editor.

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=885_

impact27:
## Description of Changes

follow

Add the possibility to follow the cursor in the outline explorer, so the user knows where (s)he is.

Updates automatically the cell name when the syntax highlighter sees a change

### Issue(s) Resolved

Fixes #885.

### Affirmation

impact27

### DIFF
--- a/spyder/plugins/editor/widgets/editor.py
+++ b/spyder/plugins/editor/widgets/editor.py
@@ -2800,9 +2800,7 @@ class EditorSplitter(QSplitter):
 
 
 class EditorWidget(QSplitter):
-    def __init__(self, parent, plugin, menu_actions, show_fullpath,
-                 show_all_files, group_cells, show_comments,
-                 sort_files_alphabetically):
+    def __init__(self, parent, plugin, menu_actions, outline_explorer_options):
         QSplitter.__init__(self, parent)
         self.setAttribute(Qt.WA_DeleteOnClose)
 
@@ -2821,12 +2819,14 @@ class EditorWidget(QSplitter):
         self.plugin.register_widget_shortcuts(self.find_widget)
         self.find_widget.hide()
         self.outlineexplorer = OutlineExplorerWidget(
-                self,
-                show_fullpath=show_fullpath,
-                show_all_files=show_all_files,
-                group_cells=group_cells,
-                show_comments=show_comments,
-                sort_files_alphabetically=sort_files_alphabetically)
+            self,
+            show_fullpath=outline_explorer_options['show_fullpath'],
+            show_all_files=outline_explorer_options['show_all_files'],
+            group_cells=outline_explorer_options['group_cells'],
+            show_comments=outline_explorer_options['show_comments'],
+            sort_files_alphabetically=outline_explorer_options[
+                'sort_files_alphabetically'],
+            )
         self.outlineexplorer.edit_goto.connect(
                      lambda filenames, goto, word:
                      plugin.load(filenames=filenames, goto=goto, word=word,
@@ -2891,8 +2891,7 @@ class EditorWidget(QSplitter):
 
 class EditorMainWindow(QMainWindow):
     def __init__(self, plugin, menu_actions, toolbar_list, menu_list,
-                 show_fullpath, show_all_files, group_cells, show_comments,
-                 sort_files_alphabetically):
+                 outline_explorer_options):
         QMainWindow.__init__(self)
         self.setAttribute(Qt.WA_DeleteOnClose)
 
@@ -2900,9 +2899,7 @@ class EditorMainWindow(QMainWindow):
         self.window_size = None
 
         self.editorwidget = EditorWidget(self, plugin, menu_actions,
-                                         show_fullpath, show_all_files,
-                                         group_cells, show_comments,
-                                         sort_files_alphabetically)
+                                         outline_explorer_options)
         self.setCentralWidget(self.editorwidget)
 
         # Setting interface theme
--- a/spyder/plugins/outlineexplorer/api.py
+++ b/spyder/plugins/outlineexplorer/api.py
@@ -22,14 +22,19 @@ outlineexplorer.set_current_editor(oe_proxy, update=True, clear=False)
 
 outlineexplorer.edit_goto.connect(handle_go_to)
 """
+from qtpy.QtCore import Signal, QObject
 
 
-class OutlineExplorerProxy(object):
+class OutlineExplorerProxy(QObject):
     """
     Proxy class between editors and OutlineExplorerWidget.
     """
 
+    sig_cursor_position_changed = Signal(int, int)
+    sig_outline_explorer_data_changed = Signal()
+
     def __init__(self):
+        super(OutlineExplorerProxy, self).__init__()
         self.fname = None
 
     def is_python(self):
@@ -58,7 +63,7 @@ class OutlineExplorerProxy(object):
         """Return the number of lines of the editor (int)."""
         raise NotImplementedError
 
-    def parent():
+    def parent(self):
         """This is used for diferenciate editors in multi-window mode."""
         return None
 
--- a/spyder/plugins/outlineexplorer/api.py
+++ b/spyder/plugins/outlineexplorer/api.py
@@ -67,6 +67,10 @@ class OutlineExplorerProxy(QObject):
         """This is used for diferenciate editors in multi-window mode."""
         return None
 
+    def get_cursor_line_number(self):
+        """Return the cursor line number."""
+        raise NotImplementedError
+
 
 class OutlineExplorerData(object):
     CLASS, FUNCTION, STATEMENT, COMMENT, CELL = list(range(5))
--- a/spyder/plugins/outlineexplorer/tests/test_widgets.py
+++ b/spyder/plugins/outlineexplorer/tests/test_widgets.py
@@ -207,6 +207,29 @@ def test_go_to_cursor_position(create_outlineexplorer, qtbot):
     assert outlineexplorer.treewidget.currentItem().text(0) == 'method1'
 
 
+def test_outlineexplorer_updates(create_outlineexplorer, qtbot):
+    """
+    Test that the cursor is followed
+    """
+    outlineexplorer = create_outlineexplorer(TEXT, 'test.py',
+                                             follow_cursor=True)
+    # Move the mouse cursor in the editor to line 5 :
+    editor = outlineexplorer.treewidget.current_editor
+    editor._editor.go_to_line(4)
+    assert editor._editor.get_text_line(4) == ""
+    # Go to cursor to open the cursor
+    qtbot.mouseClick(outlineexplorer.fromcursor_btn, Qt.LeftButton)
+
+    newcell_txt = "# %% newcell"
+    with qtbot.waitSignal(editor.sig_outline_explorer_data_changed):
+        qtbot.keyClicks(editor._editor, newcell_txt)
+
+    # editor._editor.go_to_line(3)
+    assert editor._editor.get_text_line(3) == newcell_txt
+    # check the outline explorer is up to date
+    assert outlineexplorer.treewidget.currentItem().text(0) == 'newcell'
+
+
 def test_go_to_cursor_position_with_new_file(create_outlineexplorer, qtbot):
     """
     Test that clicking on the 'Go to cursor position' button located in the
--- a/spyder/plugins/outlineexplorer/tests/test_widgets.py
+++ b/spyder/plugins/outlineexplorer/tests/test_widgets.py
@@ -132,14 +132,14 @@ CODE = """# -*- coding: utf-8 -*-
 # ---- Qt Test Fixtures
 @pytest.fixture
 def create_outlineexplorer(qtbot):
-    def _create_outlineexplorer(code, filename):
+    def _create_outlineexplorer(code, filename, follow_cursor=False):
         code_editor = CodeEditor(None)
         code_editor.set_language('py', filename)
         code_editor.set_text(code)
 
         editor = OutlineExplorerProxyEditor(code_editor, filename)
 
-        outlineexplorer = OutlineExplorerWidget()
+        outlineexplorer = OutlineExplorerWidget(follow_cursor=follow_cursor)
         outlineexplorer.set_current_editor(editor, False, False)
         outlineexplorer.show()
         outlineexplorer.setFixedSize(400, 350)
@@ -194,7 +194,6 @@ def test_go_to_cursor_position(create_outlineexplorer, qtbot):
     Regression test for issue #7729.
     """
     outlineexplorer = create_outlineexplorer(TEXT, 'test.py')
-
     # Move the mouse cursor in the editor to line 31 :
     editor = outlineexplorer.treewidget.current_editor
     editor._editor.go_to_line(31)
@@ -204,6 +203,29 @@ def test_go_to_cursor_position(create_outlineexplorer, qtbot):
     # toolbar :
     assert outlineexplorer.treewidget.currentItem() is None
     qtbot.mouseClick(outlineexplorer.fromcursor_btn, Qt.LeftButton)
+    assert outlineexplorer.treewidget.currentItem().text(0) == 'method1'
+
+
+def test_follow_cursor(create_outlineexplorer, qtbot):
+    """
+    Test that the cursor is followed.
+    """
+    outlineexplorer = create_outlineexplorer(TEXT, 'test.py',
+                                             follow_cursor=True)
+    # Move the mouse cursor in the editor to line 31 :
+    editor = outlineexplorer.treewidget.current_editor
+    editor._editor.go_to_line(31)
+    assert editor._editor.get_text_line(31) == "        if False:"
+    # method1 is collapsed
+    assert outlineexplorer.treewidget.currentItem().text(0) == 'test.py'
+
+    # Go to cursor to open the cursor
+    qtbot.mouseClick(outlineexplorer.fromcursor_btn, Qt.LeftButton)
+
+    # Check if follows
+    editor._editor.go_to_line(1)
+    assert outlineexplorer.treewidget.currentItem().text(0) == 'test.py'
+    editor._editor.go_to_line(31)
     assert outlineexplorer.treewidget.currentItem().text(0) == 'method1'
 
 
--- a/spyder/plugins/outlineexplorer/widgets.py
+++ b/spyder/plugins/outlineexplorer/widgets.py
@@ -174,11 +174,27 @@ class OutlineExplorerTreeWidget(OneColumnTree):
         self.editor_tree_cache = {}
         self.editor_ids = {}
         self.ordered_editor_ids = []
-        self.current_editor = None
+        self._current_editor = None
         title = _("Outline")
         self.set_title(title)
         self.setWindowTitle(title)
         self.setUniformRowHeights(True)
+
+    @property
+    def current_editor(self):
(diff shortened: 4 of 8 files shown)


goanpeca:
@impact27 thanks a lot for working on this!

---

For this kind of PRs it really helps to be able to see a quick animation of the change as it allows the developers team to spot issues without having to checkout the branch (sometimes we are on mobile, but can still provide useful insight!)

You could use something like https://www.cockos.com/licecap/, or similar. Please attach one! cheers!

(I will update the PR template to include this info!)

bcolsen:
@impact27 This is great! This really improves the outline explorer as a navigation tool.

Is there a particular reason for the collapse all call? I don't think users will like it collapsing things.

You need to add this option to config/main.py so the option state is saved when Spyder is closed. Yu can use #7738 as a guide for what needs to be done here. 

I think that the outline explorer should do this by default, but expanding the tree every time you click somewhere in a file would be a bit much for some files. It would be nice if it would just highlight where you are in the current tree state without expanding.

ccordoba12:
> I think that the outline explorer should do this by default

Agreed.

> It would be nice if it would just highlight where you are in the current tree state without expanding.

Agreed too.

impact27:
updated

impact27:
update

bcolsen:
@impact27 I've been testing this out and it's just what the outliner needs.

One bug I found that you probably are already be aware of is nothing is removed from the outline when blocks are deleted.

Another bug is when I change the name of a parent in the tree it collapses the branch if it's open

I'm pretty sure we don't need an option to turn this off anymore. It's just part of the base functionality.

impact27:
> One bug I found that you probably are already be aware of is nothing is removed from the outline when blocks are deleted.
> 


Yes this is slightly complicated as the syntax highlighter doesn't let me know when a block is deleted. An alternative would be to put a signal on the `userData` when it is deleted, but I had to explicitly prevent that from happening in https://github.com/spyder-ide/spyder/pull/9014 because of some strange bug (I suspect a Qt error in Qt5.9, My guess is that Qt somehow references the `userData` after deleting the associated block?).

Otherwise https://github.com/spyder-ide/spyder/pull/9014 should fix a lot of the problems in the outline explorer, as I replaced the line numbers with blocks, so the outline explorer should be able to actually know where the items are if the line changes. There is probably a bit more work needed on the outline explorer but this is out of scope for this PR. (And very out of scope for the cell name PR! I should probably rename it or something. - Use blocks instead of lines? -)

> Another bug is when I change the name of a parent in the tree it collapses the branch if it's open
> 
I would say this is a separate issue? My guess is new name => new item, so you can't keep the same settings. Maybe someone could open a new PR after this one and https://github.com/spyder-ide/spyder/pull/9014 have been merged?

> I'm pretty sure we don't need an option to turn this off anymore. It's just part of the base functionality.
> 

So should I remove the code to turn it into an option?

ccordoba12:
> So should I remove the code to turn it into an option?

Yes, I agree with @bcolsen that there's no need to make this behavior configurable.

@goanpeca, what do you think?

impact27:
>  One bug I found that you probably are already be aware of is nothing is removed from the outline when blocks are deleted.

fixed with #9014 



> So should I remove the code to turn it into an option?

Done

goanpeca:
>So should I remove the code to turn it into an option?
>Yes, I agree with @bcolsen that there's no need to make this behavior configurable.

I agree as well, I am 👍 on les configuration! The default seems pretty sensible to me :-)

Great work @impact27

ccordoba12:
@goanpeca, please help me to review this one.

spyder/plugins/editor/widgets/editor.py
```diff
@@ -2781,9 +2781,7 @@ def set_layout_settings(self, settings, dont_goto=None):
 
 
 class EditorWidget(QSplitter):
-    def __init__(self, parent, plugin, menu_actions, show_fullpath,
-                 show_all_files, group_cells, show_comments,
-                 sort_files_alphabetically):
+    def __init__(self, parent, plugin, menu_actions, oe_options):
```
goanpeca: I suggest we rename this to something more self explanatory

For example: (I am open to suggestions) 
`oe_options`  -> `outline_options`

spyder/plugins/editor/widgets/editor.py
```diff
@@ -2872,18 +2871,15 @@ def unregister_editorstack(self, editorstack):
...
         self.setAttribute(Qt.WA_DeleteOnClose)
 
         self.plugin = plugin
         self.window_size = None
 
         self.editorwidget = EditorWidget(self, plugin, menu_actions,
-                                         show_fullpath, show_all_files,
-                                         group_cells, show_comments,
-                                         sort_files_alphabetically)
+                                         oe_options)
```
goanpeca: See above

spyder/plugins/editor/widgets/editor.py
```diff
@@ -2872,18 +2871,15 @@ def unregister_editorstack(self, editorstack):
 
 class EditorMainWindow(QMainWindow):
     def __init__(self, plugin, menu_actions, toolbar_list, menu_list,
-                 show_fullpath, show_all_files, group_cells, show_comments,
-                 sort_files_alphabetically):
+                 oe_options):
```
goanpeca: See above

spyder/plugins/outlineexplorer/api.py
```diff
@@ -58,10 +63,14 @@ def get_line_count(self):
         """Return the number of lines of the editor (int)."""
         raise NotImplementedError
 
-    def parent():
+    def parent(self):
         """This is used for diferenciate editors in multi-window mode."""
         return None
 
+    def get_cursor_line_number(self):
+        """Return the cursor line number"""
```
goanpeca: Please add a period here:

```"""Return the cursor line number."""```

File: spyder/plugins/outlineexplorer/widgets.py [hunk 5]
```diff
@@ -250,15 +263,50 @@ def toggle_sort_files_alphabetically(self, state):
...
+                    if not tree_iter.isExpanded():
+                        item = tree_iter
+                    tree_iter = tree_iter.parent()
             self.setCurrentItem(item)
             self.scrollToItem(item)
-                
+
+    @Slot()
+    def go_to_follow_cursor(self):
+        """Collapse all and go to cursor position"""
```
goanpeca: Add a period: ```"""Collapse all and go to cursor position."""```

spyder/plugins/outlineexplorer/widgets.py
```diff
@@ -250,15 +263,50 @@ def toggle_sort_files_alphabetically(self, state):
...
+        if self.follow_cursor:
+            self.go_to_cursor_position(expand=False)
+
+    @Slot(bool)
+    def toggle_follow_cursor(self, state):
+        """Follow the cursor"""
```
goanpeca: Same here

spyder/plugins/outlineexplorer/widgets.py
```diff
@@ -250,15 +263,50 @@ def toggle_sort_files_alphabetically(self, state):
...
+        self.follow_cursor = state
+
+    def connect_current_editor(self, state):
+        """Connect or disconnect the editor from signals"""
```
goanpeca: Same here

spyder/plugins/outlineexplorer/widgets.py
goanpeca: Also this is a confusing name

`go_to_follow_cursor`, suggestions?

spyder/plugins/outlineexplorer/widgets.py
```diff
@@ -605,20 +655,22 @@ class OutlineExplorerWidget(QWidget):
...
-                group_cells=group_cells,
-                show_comments=show_comments,
-                sort_files_alphabetically=sort_files_alphabetically)
+            self,
+            show_fullpath=show_fullpath,
+            show_all_files=show_all_files,
+            group_cells=group_cells,
+            show_comments=show_comments,
+            sort_files_alphabetically=sort_files_alphabetically,
+            follow_cursor=follow_cursor)
```
goanpeca: Let's split this into:

```python
        self.treewidget = OutlineExplorerTreeWidget(
                self,	            self,
                show_fullpath=show_fullpath,
                show_all_files=show_all_files,
                group_cells=group_cells,
                show_comments=show_comments,
                sort_files_alphabetically=sort_files_alphabetically,
               follow_cursor=follow_cursor,
        )
```

spyder/utils/syntaxhighlighters.py
```diff
@@ -17,7 +17,7 @@
 import weakref
 
 # Third party imports
-from qtpy.QtCore import Qt
+from qtpy.QtCore import Qt, Signal, Slot, QTimer
```
goanpeca: `from qtpy.QtCore import Qt,  QTimer, Signal, Slot`

spyder/utils/syntaxhighlighters.py
```diff
@@ -110,6 +110,8 @@ class BaseSH(QSyntaxHighlighter):
     # Syntax highlighting parameters.
     BLANK_ALPHA_FACTOR = 0.31
 
+    sig_outline_explorer_changed = Signal()
```
goanpeca: Maybe adding a comment here would help better understand how this works

spyder/utils/syntaxhighlighters.py
```diff
@@ -422,6 +424,12 @@ def __init__(self, parent, font=None, color_scheme='Spyder'):
         self.import_statements = {}
         self.found_cell_separators = False
         self.cell_separators = CELL_LANGUAGES['Python']
+        self.oetimer = QTimer()
+        self.oetimer.timeout.connect(self.emit_oe_changed)
+
+    @Slot()
+    def emit_oe_changed(self):
+        self.sig_outline_explorer_changed.emit()
```
goanpeca: I would really prefer if we used longer descriptive names.

`emit_outline_changed` or even `emit_outline_explorer_changed` 

Same goes for the `self.oetimer` variable

spyder/plugins/outlineexplorer/api.py
```diff
@@ -22,14 +22,19 @@ def handle_go_to(name, line, text):
...
 
 
-class OutlineExplorerProxy(object):
+class OutlineExplorerProxy(QObject):
     """
     Proxy class between editors and OutlineExplorerWidget.
     """
 
+    sig_cursor_position_changed = Signal(int, int)
+    sig_outline_explorer_changed = Signal()
```
goanpeca: What is `outline_explorer_changed` ? the user selected something on the Outline?

impact27:
> What is outline_explorer_changed ? the user selected something on the Outline?
> 
The syntax highlighter created a new `oedata` or deleted one. So the user typed something in the editor.

impact27:
> The syntax highlighter created a new `oedata` or deleted one. So the user typed something in the editor.


Maybe it should be called `sig_outline_explorer_data_changed`?

spyder/plugins/outlineexplorer/api.py
goanpeca: > The syntax highlighter created a new `oedata` or deleted one. So the user typed something in the editor.
>Maybe it should be called `sig_outline_explorer_data_changed`?

Definitely :-) `sig_outline_explorer_data_changed` 👍

spyder/plugins/outlineexplorer/widgets.py
goanpeca: Period at the end ;-)

impact27:
> > One bug I found that you probably are already be aware of is nothing is removed from the outline when blocks are deleted.
> 
> fixed with #9014

Correction: with #9014, an update_all would pick it up. So the solution could be to have a timer that checks from time to time if a line was deleted.

impact27:
or `update_all` on `QPlainTextEdit::blockCountChanged`

goanpeca:
Which one would you rather implement? Or do you see any pros cons in one over the other?

impact27:
`QPlainTextEdit::blockCountChanged` is probably the better choice, otherwise we would be updating the outline explorer without reason.

goanpeca:
Thanks @impact27 , I will now test to see how it feels. Codewise looks great :-).

@bcolsen would you like to give it another try?

bcolsen:
I'll test it out and look over the code again.

goanpeca:
@impact27 there still seems to be a [pep8speaks issue](https://github.com/spyder-ide/spyder/pull/9219#issuecomment-487275346)

bcolsen:
I found a couple more bugs but I think they would be cleared up with #9082. Should that be merged first? 

1. The outline duplicates an item when the line changes.
outline new line

2. Deleted items reappear when the their lines are repopulated.
outline delete

impact27:
> I found a couple more bugs but I think they would be cleared up with #9082 <https://github.com/spyder-ide/spyder/pull/9082>. Should that be merged first?

I think that it would solve these bugs but it is not necessary. The current behaviour is to not update until you save. If you merge #9219 then there will be update but some bug as you noticed, but these errors go away when you save the file. Then when you merge #9082 these bugs will go away.

spyder/utils/syntaxhighlighters.py
```diff
@@ -422,6 +424,11 @@ def __init__(self, parent, font=None, color_scheme='Spyder'):
         self.import_statements = {}
         self.found_cell_separators = False
         self.cell_separators = CELL_LANGUAGES['Python']
+        # Avoid updating the outline explorer with every single letter typed
+        self.outline_explorer_data_update_timer = QTimer()
+        self.outline_explorer_data_update_timer.setSingleShot(True)
+        self.outline_explorer_data_update_timer.timeout.connect(
+                self.sig_outline_explorer_data_changed)
```
CAM-Gerlach: 4 spaces for hanging indents (Spyder is wrong in that regard, I *still* need to fix it...)

spyder/utils/syntaxhighlighters.py
```diff
@@ -552,6 +559,7 @@ def highlight_block(self, text):
             block_nb = self.currentBlock().blockNumber()
             self.outlineexplorer_data[block_nb] = oedata
             self.outlineexplorer_data['found_cell_separators'] = self.found_cell_separators
+            self.outline_explorer_data_update_timer.start(500)
```
CAM-Gerlach: Maybe ``1000`` is enough to cut the overhead in half? Every second should be fine, I'd think...

spyder/plugins/outlineexplorer/tests/test_widgets.py
```diff
@@ -207,6 +206,52 @@ def test_go_to_cursor_position(create_outlineexplorer, qtbot):
...
+    editor = outlineexplorer.treewidget.current_editor
+    editor._editor.go_to_line(4)
+    assert editor._editor.get_text_line(4) == ""
+    # Go to cursor to open the cursor
+    qtbot.mouseClick(outlineexplorer.fromcursor_btn, Qt.LeftButton)
+
+    newcell_txt = "# %% newcell"
+    with qtbot.waitSignal(editor.sig_outline_explorer_data_changed):
+        qtbot.keyClicks(editor._editor, newcell_txt)
+
```
CAM-Gerlach: This is what ``PEP8Speaks`` is complaining about, FYI. If you turn on real time code style analysis in Spyder, these sorts of errors get highlighted for you automatically.

spyder/plugins/editor/widgets/editor.py
```diff
@@ -2802,12 +2800,14 @@ def __init__(self, parent, plugin, menu_actions, show_fullpath,
...
-                group_cells=group_cells,
-                show_comments=show_comments,
-                sort_files_alphabetically=sort_files_alphabetically)
+            self,
+            show_fullpath=outline_explorer_options['show_fullpath'],
+            show_all_files=outline_explorer_options['show_all_files'],
+            group_cells=outline_explorer_options['group_cells'],
+            show_comments=outline_explorer_options['show_comments'],
+            sort_files_alphabetically=outline_explorer_options[
+                    'sort_files_alphabetically'],
```
CAM-Gerlach: 4 spaces for a hanging indent

spyder/plugins/outlineexplorer/widgets.py
```diff
@@ -174,12 +175,26 @@ def __init__(self, parent, show_fullpath=False, show_all_files=True,
...
         self.setWindowTitle(title)
         self.setUniformRowHeights(True)
 
+    @property
+    def current_editor(self):
+        return self._current_editor
+
+    @current_editor.setter
+    def current_editor(self, value):
+        if self._current_editor == value:
```
CAM-Gerlach: At least a oneliner docstring would be nice.

spyder/utils/syntaxhighlighters.py
impact27: It is a one shot timer, so it doesn't really matter. The goal is to avoid calling it multiple times while typing a name (slightly slower ) saving a file (massively slower). I took 500ms as being larger than the time between two keys, but still small so it seems fast. I could even decrease it in my opinion as the file saving scenario is the one I really want to avoid.

ccordoba12:
@impact27, this needs a merge with master to fix the merge conflict.

@goanpeca, please take a look to this PR again to see if we can merge it.

spyder/utils/syntaxhighlighters.py
CAM-Gerlach: Okay, sure. 500 ms makes sense, then. Thanks!

CAM-Gerlach:
> Comments

@impact27 Could you please try to be a little more properly descriptive with your commits and messages, so its clear what meaningful changes are being summarized when browsing through Spyder's commit history? Alternatively, if they are very minor fixes (PEP 8, PEP 257, fix tests, minor feedback, etc) to a previous commit in the same PR branch, I usually just ``--amend``, ``fixup``, etc. them in to the appropriate commit to avoid cluttering the final history with broken, pointless or trivial changes, but that's my personal approach. I'm not sure what @ccordoba12 prefers, if he has any one.

goanpeca:
@CAM-Gerlach I think it would be great to document this. Could you give us a hand with this?

Maybe the wiki should be ok for now? Somewhere over here https://github.com/spyder-ide/spyder/wiki/Dev%3A-Index

spyder/plugins/outlineexplorer/tests/test_widgets.py
```diff
@@ -207,6 +206,52 @@ def test_go_to_cursor_position(create_outlineexplorer, qtbot):
     assert outlineexplorer.treewidget.currentItem().text(0) == 'method1'
 
 
+def test_follow_cursor(create_outlineexplorer, qtbot):
+    """
+    Test that the cursor is followed
+    """
```
goanpeca: This is not really a fix needed, but just a reminder that docstrings should be handle as complete sentences, hence we should try to add a period at the end ;-)

CAM-Gerlach: Correct, per PEP 257.

CAM-Gerlach:
@goanpeca Yup, we've been doing so on [this page](https://github.com/spyder-ide/spyder/wiki/Dev%3A-Coding-Style). I'll add it there, thanks.

PR #9230, opened 2019-04-29
PR: Display extension type icons and create entry as files are opened in "Open recent" menu

### LINKED ISSUE #9150: Improve the "Open recent" menu
@juanis2112, I refer to this menu:

Selección_024

I would like to see the following improvements:

- [x] Create entries on it after a file is opened in the Editor. Right now, we only create an entry after the file is closed, which is odd and counter-intuitive.
- [x] Display the icon type corresponding to the file on its left. Right now we are displaying the same icon for all files, which is not so good. If a file doesn't have an icon, then we should show the same generic icon that we're displaying now.

juanis2112:
## Description of Changes

Display extension type icons in Recent Files menu and create entries as soon as file is opened in editor.

IMG_1349

### Issue(s) Resolved

Fixes #9150 

### Affirmation

Juanis2112

### DIFF
--- a/spyder/plugins/editor/plugin.py
+++ b/spyder/plugins/editor/plugin.py
@@ -1658,9 +1658,10 @@ class Editor(SpyderPluginWidget):
         self.recent_file_menu.clear()
         if recent_files:
             for fname in recent_files:
-                action = create_action(self, fname,
-                                       icon=ima.get_icon_by_extension(fname, scale_factor=1.0),
-                                       triggered=self.load)
+                action = create_action(
+                    self, fname,
+                    icon=ima.get_icon_by_extension(fname, scale_factor=1.0),
+                    triggered=self.load)
                 action.setData(to_qvariant(fname))
                 self.recent_file_menu.addAction(action)
         self.clear_recent_action.setEnabled(len(recent_files) > 0)
--- a/spyder/plugins/editor/plugin.py
+++ b/spyder/plugins/editor/plugin.py
@@ -1653,13 +1653,13 @@ class Editor(SpyderPluginWidget):
         """Update recent file menu"""
         recent_files = []
         for fname in self.recent_files:
-            if self.is_file_opened(fname) is None and osp.isfile(fname):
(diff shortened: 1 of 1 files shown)


spyder/plugins/editor/plugin.py
```diff
@@ -1653,13 +1653,13 @@ def update_recent_file_menu(self):
         """Update recent file menu"""
         recent_files = []
         for fname in self.recent_files:
-            if self.is_file_opened(fname) is None and osp.isfile(fname):
+            if osp.isfile(fname):
```
goanpeca: @juanis2112 question, why did we remove `self.is_file_opened(fname)`?

juanis2112: @goanpeca I understand this line is the one that doesn't let the entry to be created while the file is open. Please correct me if I'm wrong.

goanpeca: Yes you are right, missed that part, I thought this was only about the extension!, y bad :-)

spyder/plugins/editor/plugin.py
```diff
@@ -1653,13 +1653,13 @@ def update_recent_file_menu(self):
...
-            if self.is_file_opened(fname) is None and osp.isfile(fname):
+            if osp.isfile(fname):
                 recent_files.append(fname)
         self.recent_file_menu.clear()
         if recent_files:
             for fname in recent_files:
                 action = create_action(self, fname,
-                                       icon=ima.icon('FileIcon'),
+                                       icon=ima.get_icon_by_extension(fname, scale_factor=1.0),
                                        triggered=self.load)
```
ccordoba12: To fix the Pep8speaks warning, please change this code to be

```python
action = create_action(
    self, fname,
    icon=ima.get_icon_by_extension(fname, scale_factor=1.0),
    triggered=self.load)
```

juanis2112: @ccordoba12 done!

PR #9231, opened 2019-04-29
PR: Handle empty options in pycodestyle and pydocstyle preferences

### LINKED ISSUE #9209: Setting ignore rules for Pycodestyle is not working
While working on PR #9186 I noticed that when you set something to ignore in the "Code style" tab of "Completion and linting", that's not getting send to server.

@dalthviz, I think the problem is here

https://github.com/spyder-ide/spyder/blob/5a39ace4a24d95b0b4ecaf0fada0317988d37dc3/spyder/plugins/editor/lsp/manager.py#L245

because:

1. In case 'pycodestyle/select' is an empty string, then this code generates `['']` to pass to the server and it should be `[]` (i.e. an empty list, not a list with an empty string).
2. I think Pycodestyle doesn't accept that both `select` and `ignore` are passed at the same time (but please verify that).

Also, please remove the `xfail` of this test after fixing this issue:

https://github.com/spyder-ide/spyder/blob/d3ca22ab0fbe99ed4d64efcb3ee78b13e893452e/spyder/plugins/editor/widgets/tests/test_warnings.py#L27-L32

dalthviz:
## Description of Changes

### Issue(s) Resolved

Fixes #9209 

### Affirmation

### DIFF
--- a/spyder/plugins/editor/lsp/manager.py
+++ b/spyder/plugins/editor/lsp/manager.py
@@ -248,10 +248,11 @@ class LSPManager(QObject):
 
         pycodestyle = {
             'enabled': self.get_option('pycodestyle'),
-            'exclude': [exclude.strip() for exclude in cs_exclude],
-            'filename': [filename.strip() for filename in cs_filename],
-            'select': [select.strip() for select in cs_select],
-            'ignore': [ignore.strip() for ignore in cs_ignore],
+            'exclude': [exclude.strip() for exclude in cs_exclude if exclude],
+            'filename': [filename.strip()
+                         for filename in cs_filename if filename],
+            'select': [select.strip() for select in cs_select if select],
+            'ignore': [ignore.strip() for ignore in cs_ignore if ignore],
             'hangClosing': False,
             'maxLineLength': cs_max_line_length
         }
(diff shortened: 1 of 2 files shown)


dalthviz:
@ccordoba12 testing the behavior of the two options (`select` and `ignore`), what I see is that when you use the `select` option, only the warnings selected are shown (the other ones are ignore). Also, the `select` option takes precedence to the `ignore` option, so for example if you put for the `select` and `ignore` options the `E202` code, only the warnings associted with an `E202` warning will be shown

spyder/plugins/editor/lsp/manager.py
```diff
@@ -248,10 +248,11 @@ def generate_python_config(self):
...
             'enabled': self.get_option('pycodestyle'),
-            'exclude': [exclude.strip() for exclude in cs_exclude],
-            'filename': [filename.strip() for filename in cs_filename],
-            'select': [select.strip() for select in cs_select],
-            'ignore': [ignore.strip() for ignore in cs_ignore],
+            'exclude': [exclude.strip() for exclude in cs_exclude if exclude],
+            'filename': [filename.strip()
+                         for filename in cs_filename if filename],
+            'select': [select.strip() for select in cs_select if select],
+            'ignore': [ignore.strip() for ignore in cs_ignore if ignore],
```
ccordoba12: Pretty nice usage of list comprehension! :+1:

ccordoba12:
Merging because `pycodestyle` accepts both `select` and `ignore`, as @dalthviz pointed out.

The question about mentioning this in our docs is a separate issue.

CAM-Gerlach:
Thanks, @ccordoba12 . I created spyder-ide/spyder-docs#96 to cover this.

PR #9249, opened 2019-05-01
PR: Add stdio LSP transport client

andfoy:
## Description of Changes
Add a LSP transport client that allows to communicate with servers that only allow to use standard input/output pipes as communication streams.

By introducing a new STDIO transport client, we extend Spyder to communicate with the remaining LSP servers that only use STDIO pipes as communication channels.

### Issue(s) Resolved

See #7724

### Affirmation

@andfoy

### TODOs:
- [x] Generalize the main consumer/producer functions to ensure code reuse
- [x] Implement the stdio transport client
- [x] Add graphical options to the Completion and Linting panel to enable LSP stdio communication
- [x] Add tests?


andfoy:
Note to self: It seems that redirecting pyls process stdout to transport stdin and viceversa, does not work, due to pipe buffer starvation. Instead, we should let the transport process spawn the lsp_server and read/write directly/to from that process stdout/stdin.

To do so, we should use pexpect Popen in Unix and pywinpty in Windows to spawn and watch the processes. 

PD: To uniformize the expected code base, pexpect interface should be complete on that platform https://github.com/spyder-ide/pywinpty/pull/79

andfoy:
@ccordoba12 This is working now on POSIX systems, we're only missing tests on windows using pywinpty

andfoy:
We need to add support to standard error streams on pywinpty, so we can debug stdio servers on Windows

ccordoba12:
Please fix our tests in Python 2 and Windows. The introduction of the stdio transport layer should not affect them given that it's not activated for the PyLS.

spyder/plugins/editor/lsp/client.py
```diff
@@ -89,6 +85,13 @@ def __init__(self, parent,
         self.transport_args = [sys.executable, '-u',
                                osp.join(LOCATION, 'transport', 'main.py')]
         self.external_server = server_settings.get('external', False)
+        self.stdio = server_settings.get('stdio', False)
+        # Setting stdio on implies that external_server is off
+        if self.stdio and self.external_server:
+            error = ('If server is set to use stdio communication, '
+                     'then it cannot be an external server')
```
ccordoba12: Why is this restriction necessary?

spyder/plugins/editor/lsp/client.py
```diff
@@ -98,15 +101,23 @@ def __init__(self, parent,
...
         server_args = server_args_fmt.format(**server_settings)
-        # transport_args = self.local_server_fmt % (server_settings)
-        # if self.external_server:
         transport_args = self.external_server_fmt % (server_settings)
 
-        self.server_args = [sys.executable, '-m', server_settings['cmd']]
-        self.server_args += server_args.split(' ')
+        self.server_args = []
+        if language == 'python':
+            self.server_args += [sys.executable, '-m']
```
ccordoba12: Good catch!

spyder/plugins/editor/lsp/client.py
```diff
@@ -126,13 +137,25 @@ def start(self):
...
                 self.server_args.append('-vv')
 
+        server_stdin = subprocess.PIPE
+        server_stdout = server_log
+        server_stderr = subprocess.STDOUT
+
+        # server_stderr = server_log
+        #     server_stdin, server_stdout = os.pipe()
+        #     # server_stdin = os.fdopen(server_stdin, 'wr')
+        #     # server_stdout = os.fdopen(server_stdout, 'wr')
```
ccordoba12: Please remove this if it's not necessary.

spyder/plugins/editor/lsp/client.py
```diff
@@ -177,9 +201,18 @@ def patched_cleanup():
...
         logger.info('Starting transport: {0}'
                     .format(' '.join(self.transport_args)))
+        transport_stdout = client_log
+        transport_stdin = subprocess.PIPE
+        transport_stderr = subprocess.STDOUT
+        if self.stdio:
+            transport_stdin = subprocess.PIPE
+            transport_stdout = subprocess.PIPE
+            transport_stderr = client_log
+            self.transport_args += self.server_args
```
ccordoba12: Please rewrite this like this

```python
if self.stdio:
    transport_stdin = subprocess.PIPE
    transport_stdout = subprocess.PIPE
    transport_stderr = client_log
    self.transport_args += self.server_args
else:
    transport_stdout = client_log
    transport_stdin = subprocess.PIPE
    transport_stderr = subprocess.STDOUT
```

spyder/plugins/editor/lsp/transport/common/__init__.py
```diff
@@ -0,0 +1,6 @@
+
+from .consumer import IncomingMessageThread
+from .producer import LanguageServerClient
+
+IncomingMessageThread
+LanguageServerClient
```
ccordoba12: I don't think these two lines are necessary, right?

spyder/plugins/editor/lsp/transport/common/producer.py
```diff
@@ -0,0 +1,114 @@
...
+
+# Third party imports
+import zmq
+
+TIMEOUT = 5000
+
+logger = logging.getLogger(__name__)
+
+
+class LanguageServerClient:
```
ccordoba12: Please make this class inherit from `object` to make it a new class in Python 2. That should fix the errors on it.

spyder/plugins/editor/lsp/transport/main.py
```diff
@@ -109,16 +113,19 @@ def restore(self):
...
-                                  zmq_out_port=args.zmq_out_port,
-                                  use_external_server=args.external_server,
-                                  server=args.server,
-                                  server_args=unknownargs)
+    LanguageServerClient = partial(TCPLanguageServerClient,
+                                   host=args.server_host,
+                                   port=args.server_port)
+    if args.stdio_server:
+        LanguageServerClient = partial(StdioLanguageServerClient,
+                                       server_args=unknownargs)
```
ccordoba12: Is it possible to write this like:

```python
if args.stdio_server:
    LanguageServerClient = partial(StdioLanguageServerClient,
                                   server_args=unknownargs)
else:
    LanguageServerClient = partial(TCPLanguageServerClient,
                                   host=args.server_host,
                                   port=args.server_port)
```

?

I think it'd be clearer this way.

spyder/preferences/languageserver.py
```diff
@@ -203,6 +208,12 @@ def __init__(self, parent, language=None, cmd='', host='127.0.0.1',
         self.external_cb.setChecked(external)
         self.external_cb.stateChanged.connect(self.set_local_options)
 
+        self.stdio_cb = QCheckBox(_('Use stdio'), self)
```
ccordoba12: Please make the label here more informative.

spyder/plugins/editor/lsp/client.py
andfoy: Unless an user gives a process ID to redirect its stdin and stdout pipes, I think is impossible to connect to an external server using stdio

continuous_integration/azure/install.bat
```diff
@@ -34,3 +34,7 @@ if errorlevel 1 exit 1
 :: Install codecov
 pip install -q codecov
 if errorlevel 1 exit 1
+
+:: Install pywinpty
+pip install -q pywinpty
```
ccordoba12: This is not correct. Please remove it from here and add it instead to `requirements/conda.txt`

spyder/plugins/editor/lsp/transport/common/consumer.py
```diff
@@ -126,3 +129,7 @@ def parse_headers(self, headers):
     def stop(self):
         with self.mutex:
             self.stopped = True
+
+    def read_num_bytes(self, n):
+        """Subclasses should override this method"""
+        return b''
```
ccordoba12: Perhaps this should be

    raise NotImplementedError

?

I mean, that should force subclassess to really implement this.

spyder/plugins/editor/lsp/transport/main.py
```diff
@@ -109,16 +116,21 @@ def restore(self):
 
 if __name__ == '__main__':
     logger_init(args.transport_debug)
+    unknownargs = [x for x in unknownargs if len(x) > 0]
+    unknownargs = ' '.join(unknownargs)
```
ccordoba12: What does it mean `unknownargs`? Could we use a better variable name instead?

spyder/preferences/languageserver.py
```diff
@@ -203,6 +208,12 @@ def __init__(self, parent, language=None, cmd='', host='127.0.0.1',
         self.external_cb.setChecked(external)
         self.external_cb.stateChanged.connect(self.set_local_options)
 
+        self.stdio_cb = QCheckBox(_('Use stdio pipes for communication'), self)
```
ccordoba12: This is not correct. Since using an external server and stdio are mutually exclusive options, you need to use radio buttons for both here with the same button group.

Please see how we do that for the screen resolution options here:

https://github.com/spyder-ide/spyder/blob/ffb1f9361c0f6357b13f2f45b6dfe4f9e17cea70/spyder/preferences/general.py#L192-L234

spyder/preferences/languageserver.py
```diff
@@ -354,12 +369,47 @@ def set_defaults(self):
...
+            self.stdio_cb.setChecked(False)
+            self.stdio_cb.stateChanged.connect(self.set_stdio_options)
+        try:
+            self.validate()
+        except Exception:
+            pass
+
+    @Slot(bool)
+    @Slot(int)
+    def set_stdio_options(self, enabled):
```
ccordoba12: i think this method is not needed by using radio buttons, as I suggested above.

spyder/preferences/languageserver.py
```diff
@@ -951,6 +1008,40 @@ def setup_page(self):
         vlayout.addWidget(tabs)
         self.setLayout(vlayout)
 
+    def disable_tcp(self, state):
```
ccordoba12: This is probably not needed either?

spyder/preferences/languageserver.py
andfoy: But we may have TCP "internal" servers that are launched by Spyder

andfoy: I think I expressed myself in a wrong way. stdio implies not tcp

ccordoba12: Sorry, where is this option being added? In the "Other languages" dialog you designed? Or in the "Advanced settings" we added with Daniel?

andfoy: On both panels, to be consistent

ccordoba12: Ok, then I think for "Advanced settings" we should really use ratio buttons for "External server" and "Stdio" because both can't be selected at the same time, right?

andfoy: But, both can be unselected at the same time, i.e., An internal TCP server

continuous_integration/azure/install.bat
andfoy: Is it possible to mark ``pywinpty`` as a OS-dependent dependency?

spyder/preferences/languageserver.py
ccordoba12: You're right. But happens when users try to select both at the same time?

spyder/plugins/editor/lsp/transport/main.py
andfoy: Such args are passed to the stdio server, when the transport layer is on charge of spawning it. Could it be ``extra_args``?

continuous_integration/azure/install.bat
ccordoba12: Sorry, you're right. Then please add a new file called `requirements/windows.txt` and use it only on Windows.

Besides, you need to update our Contributing guide to mention that. And also our Readme to mention the new dependency.

spyder/plugins/editor/lsp/transport/main.py
ccordoba12: Yep, much better :+1:

spyder/preferences/languageserver.py
andfoy: They simply can't. Once an option is selected, the other is unchecked and disabled

spyder/plugins/editor/lsp/transport/common/consumer.py
andfoy: You're right

andfoy:
@ccordoba12 I'll test the pywinpty version on an actual Windows box later on today (the VM is too slow and pywinpty hangs), I'll get back to you once it works, such that this PR can be merged

andfoy:
@ccordoba12  I have bad news regarding Windows support using pywinpty. It seems that due to this [reason](https://support.microsoft.com/en-us/help/830473/command-prompt-cmd-exe-command-line-string-limitation), it is not possible to give more than 1024 characters to any console process (including PowerShell).

andfoy:
Now it works on Windows!

spyder/preferences/languageserver.py
ccordoba12: Ok, great!

ccordoba12:
> Now it works on Windows!

I don't understand. How did you manage to make things work on Windows?

andfoy:
> Changing the options This is an external server and Use stdio pipes should trigger an Spyder restart. This is obvious to us and people that knows about the LSP, but not to everyone. 

Is there any example to restart Spyder?

ccordoba12:
Great!! Things are working fine on my testing with the PyLS in stdio mode.

I have some more comments:

1. Changing the options `This is an external server` and `Use stdio pipes` should trigger an Spyder restart. This is obvious to us and people that knows about the LSP, but not to everyone.
2. We also need to generate a restart when those options are changed in the dialog that setups servers for other languages.
3. About the tests, please parametrize the `lsp_manager` fixture to be started with tcp and stdio, and then use that fixture with both parameters in `plugins/editor/lsp/tests/test_client.py`.

goanpeca:
See:
https://github.com/spyder-ide/spyder/blob/master/spyder/preferences/configdialog.py#L463
https://github.com/spyder-ide/spyder/blob/master/spyder/preferences/configdialog.py#L490
https://github.com/spyder-ide/spyder/blob/master/spyder/preferences/configdialog.py#L834

goanpeca:
@ccordoba12 why is a restart required?

ccordoba12:
> why is a restart required?

Mmhh, yeah, on a second thought, I think it's not strictly necessary.

@andfoy, for both options (external server and stdio) can't we kill the current server and start a new transport layer?

andfoy:
>  can't we kill the current server and start a new transport layer?

I think that's happening right now

ccordoba12:
> I think that's happening right now

Are you sure? Please point to the code that does that (I don't remember seeing it).

andfoy:
https://github.com/spyder-ide/spyder/blob/01720a8563ad3ac44c6b887f73564c767204dab8/spyder/plugins/editor/lsp/manager.py#L195

ccordoba12:
@andfoy, this has a conflict now, please merge with master to solve it.

continuous_integration/azure/install.sh
```diff
@@ -8,6 +8,9 @@ if [ "$USE_CONDA" = "yes" ]; then
     # Install test ones
     conda install -q -y -c spyder-ide --file requirements/tests.txt
 
+    # Install pytest-lazy-fixture
+    pip install -q pytest-lazy-fixture
```
ccordoba12: Please remove this too.

continuous_integration/travis/install.sh
```diff
@@ -26,6 +26,9 @@ if [ "$USE_CONDA" = "yes" ]; then
     # Install coveralls
     pip install -q coveralls
 
+    # Install pytest-lazy-fixture
+    pip install -q pytest-lazy-fixture
```
ccordoba12: And also this.

spyder/plugins/editor/lsp/tests/fixtures.py
```diff
@@ -58,30 +58,37 @@ def qtbot_module(qapp, request):
...
-
-    def teardown():
-        manager.shutdown()
-        os.environ['SPY_TEST_USE_INTROSPECTION'] = 'False'
-        CONF.set('lsp-server', 'pycodestyle', False)
-        CONF.set('lsp-server', 'pydocstyle', False)
-
-    request.addfinalizer(teardown)
-    return manager
+def lsp_context(stdio):
```
ccordoba12: Let's change this arg to `is_stdio` for clarity.

spyder/plugins/editor/lsp/tests/fixtures.py
```diff
@@ -58,30 +58,37 @@ def qtbot_module(qapp, request):
...
+            CONF.set('lsp-server', 'pycodestyle', False)
+            CONF.set('lsp-server', 'pydocstyle', False)
+
+        request.addfinalizer(teardown)
+        return manager
+    return wrapper
+
+
+lsp_manager = lsp_context(False)
+lsp_stdio_manager = lsp_context(True)
```
ccordoba12: Also change this to

    lsp_manager = lsp_context(is_stdio=False)
    lsp_stdio_manager = lsp_context(is_stdio=True)

goanpeca:
Awesome work!

Now let's get fortran LSP Working :-p !

andfoy:
> Now let's get fortran LSP Working :-p !

And x86 by the way? Haha

goanpeca:
It is made in python so it should make things easier :-)

ccordoba12:
I agree with @goanpeca. We should definitely add tests for other languages, and since the Fortran server is made in Python, it's the easiest one.

andfoy:
> I agree with @goanpeca. We should definitely add tests for other languages, and since the Fortran server is made in Python, it's the easiest one.

[Elixir-ls](https://github.com/JakeBecker/elixir-ls) works exclusively on stdio also

ccordoba12:
Sure, but how could we add it to our test suite?

PR #9266, opened 2019-05-05
PR: Show PyLS server errors in Spyder's error report dialog

### LINKED ISSUE #9211: Show PyLS server errors in Spyder's error report dialog
Instead of simply logging the errors received from the LSP servers, as it's currently done here:

https://github.com/spyder-ide/spyder/blob/c5f95e341f64189a06930dc6750d8429d97a7a4b/spyder/plugins/editor/lsp/client.py#L239-L241

I'd like those errors to be routed to our error repor dialog, so people can send them to us, but only if the language is Python.

For that @steff456, please emit a Qt signal with the error, then capture it from with the manager, like this

https://github.com/spyder-ide/spyder/blob/5a39ace4a24d95b0b4ecaf0fada0317988d37dc3/spyder/plugins/editor/lsp/manager.py#L164-L166

and send it to the main window, to a method similar to `report_issue`.

steff456:
## Description of Changes
Show PyLS server errors on the main window using a Qt signal 

Improve the logging for LSP PyLS errors

### Issue(s) Resolved

Fixes #9211

### Affirmation

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -2678,7 +2678,7 @@ class MainWindow(QMainWindow):
 
     def report_pyls_error(self, message):
         """Report a PyLS internal error in our report dialog."""
-        from spyder.widget.reporterror import SpyderErrorDialog
+        from spyder.widgets.reporterror import SpyderErrorDialog
         error_dialog = SpyderErrorDialog(self, is_report=True)
         title = "Internal Python Language Server error"
         description = ("An error has occurred with the code completion and "
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -2676,10 +2676,18 @@ class MainWindow(QMainWindow):
                 url.setQuery(query)
                 QDesktopServices.openUrl(url)
 
-    def report_lsp_error(self, message):
-        """Report a LSP error in the main window"""
-        # TODO: Display error message in a Qt widget
-        pass
+    def report_pyls_error(self, message):
+        """Report a PyLS internal error in our report dialog."""
+        from spyder.widget.reporterror import SpyderErrorDialog
+        error_dialog = SpyderErrorDialog(self, is_report=True)
+        title = _("Internal Python Language Server error")
+        description = _("An error has occurred with the code completion and "
+                        "linting services provided by the Python Language "
+                        "Server. Please submit this report to Github if you "
+                        "want to see this fixed.")
+        error_dialog.set_title_description(title, description)
+        error_dialog.append_traceback(message)
+        error_dialog.show()
 
     @Slot()
     def trouble_guide(self):
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -2676,6 +2676,11 @@ class MainWindow(QMainWindow):
                 url.setQuery(query)
                 QDesktopServices.openUrl(url)
 
+    def report_lsp_error(self, message):
+        """Report a LSP error in the main window"""
+        # TODO: Display error message in a Qt widget
+        pass
+
     @Slot()
     def trouble_guide(self):
         """Open Spyder troubleshooting guide in a web browser."""
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -2679,12 +2679,14 @@ class MainWindow(QMainWindow):
     def report_pyls_error(self, message):
         """Report a PyLS internal error in our report dialog."""
         from spyder.widgets.reporterror import SpyderErrorDialog
-        error_dialog = SpyderErrorDialog(self, is_report=True)
+        error_dialog = SpyderErrorDialog(self)
         title = "Internal Python Language Server error"
         description = ("An error has occurred with the code completion and "
                        "linting services provided by the Python Language "
-                       "Server. Please submit this report to Github if you "
-                       "want to see this fixed.")
+                       "Server.\n\n"
+                       "Please submit this report to Github if you want "
+                       "to see it fixed.")
+        error_dialog.main_label.hide()
         error_dialog.set_title(title)
         error_dialog.set_description(description)
         error_dialog.append_traceback(message)
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -2676,22 +2676,6 @@ class MainWindow(QMainWindow):
                 url.setQuery(query)
                 QDesktopServices.openUrl(url)
 
-    def report_pyls_error(self, message):
-        """Report a PyLS internal error in our report dialog."""
-        from spyder.widgets.reporterror import SpyderErrorDialog
-        error_dialog = SpyderErrorDialog(self)
-        title = "Internal Python Language Server error"
-        description = ("An error has occurred with the code completion and "
-                       "linting services provided by the Python Language "
-                       "Server.\n\n"
-                       "Please submit this report to Github if you want "
-                       "to see it fixed.")
-        error_dialog.main_label.hide()
-        error_dialog.set_title(title)
-        error_dialog.set_description(description)
-        error_dialog.append_traceback(message)
-        error_dialog.show()
-
     @Slot()
     def trouble_guide(self):
         """Open Spyder troubleshooting guide in a web browser."""
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -2680,12 +2680,13 @@ class MainWindow(QMainWindow):
         """Report a PyLS internal error in our report dialog."""
         from spyder.widget.reporterror import SpyderErrorDialog
         error_dialog = SpyderErrorDialog(self, is_report=True)
-        title = _("Internal Python Language Server error")
-        description = _("An error has occurred with the code completion and "
-                        "linting services provided by the Python Language "
-                        "Server. Please submit this report to Github if you "
-                        "want to see this fixed.")
-        error_dialog.set_title_description(title, description)
+        title = "Internal Python Language Server error"
+        description = ("An error has occurred with the code completion and "
+                       "linting services provided by the Python Language "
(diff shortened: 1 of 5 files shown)


spyder/app/mainwindow.py
```diff
@@ -2676,6 +2676,11 @@ def report_issue(self, body=None, title=None, open_webpage=False):
                 url.setQuery(query)
                 QDesktopServices.openUrl(url)
 
+    def report_lsp_error(self, message):
+        """Report a LSP error in the main window"""
+        # TODO: Display error message in a Qt widget
+        pass
```
ccordoba12: To display the error, please do the following:

1. Create an new instance of `SpyrderErrorDialog`.
2. Set its title to `Internal Python Language Server error`. For this please add a new `set_title` method to `SpyrderErrorDialog` and use it here.
3. Set its description to "An error has occurred with the code completion and linting services provided by the Python Language Server. Please submit this report to Github if you want to see this fixed". For this please add a new method called `set_description` to `SpyrderErrorDialog` and use it here.
4. Set its traceback to the error coming from the server with `SpyrderErrorDialog`'s method called `add_traceback`.

See the `report_issue` method to see how we create the instance of `SpyrderErrorDialog` and display it.

spyder/app/mainwindow.py
```diff
@@ -2676,6 +2676,11 @@ def report_issue(self, body=None, title=None, open_webpage=False):
                 url.setQuery(query)
                 QDesktopServices.openUrl(url)
 
+    def report_lsp_error(self, message):
```
ccordoba12: Let's rename this to `report_pyls_error` (because it only applies to the Python server).

spyder/app/mainwindow.py
```diff
@@ -2676,6 +2676,11 @@ def report_issue(self, body=None, title=None, open_webpage=False):
                 url.setQuery(query)
                 QDesktopServices.openUrl(url)
 
+    def report_lsp_error(self, message):
+        """Report a LSP error in the main window"""
```
ccordoba12: Please change this to

    Report a PyLS internal error in our report dialog.

spyder/plugins/editor/lsp/client.py
```diff
@@ -57,6 +57,8 @@ class LSPClient(QObject, LSPMethodProviderMixIn):
     # Signals
     sig_initialize = Signal(dict, str)
 
+    sig_lsp_error = Signal(str)
```
ccordoba12: Please change this to `sig_server_error` and add the following comment above:

    # This signal is used to report internal server errors through
    # Spyder's facilities.

spyder/plugins/editor/lsp/client.py
```diff
@@ -239,6 +241,8 @@ def on_msg_received(self):
                 if 'error' in resp:
                     logger.debug('{} Response error: {}'
                                  .format(self.language, repr(resp['error'])))
+                    if self.language == 'python':
+                        self.sig_lsp_error.emit(repr(resp['error']))
 
                 if 'method' in resp:
```
ccordoba12: I think we should change this to

    elif 'method' in resp:

because that's another possible response besides `error`.

@andfoy, is this correct?

spyder/plugins/editor/lsp/manager.py
```diff
@@ -128,6 +128,11 @@ def reinitialize_all_clients(self):
                 instance.folder = folder
                 instance.initialize()
 
+    @Slot(str)
+    def receive_error(self, value):
```
ccordoba12: Please change this to `report_server_error`

spyder/plugins/editor/lsp/manager.py
```diff
@@ -128,6 +128,11 @@ def reinitialize_all_clients(self):
...
+        """Receive the LSP message error and redirect to the main window."""
```
ccordoba12: Please change this to

    Report server errors in our error report dialog.

spyder/app/mainwindow.py
```diff
@@ -2676,10 +2676,18 @@ def report_issue(self, body=None, title=None, open_webpage=False):
...
 
-    def report_lsp_error(self, message):
-        """Report a LSP error in the main window"""
-        # TODO: Display error message in a Qt widget
-        pass
+    def report_pyls_error(self, message):
+        """Report a PyLS internal error in our report dialog."""
+        from spyder.widget.reporterror import SpyderErrorDialog
+        error_dialog = SpyderErrorDialog(self, is_report=True)
+        title = _("Internal Python Language Server error")
```
ccordoba12: Please remove the usage of `_(...)` here. It's not needed.

spyder/app/mainwindow.py
```diff
@@ -2676,10 +2676,18 @@ def report_issue(self, body=None, title=None, open_webpage=False):
...
+        description = _("An error has occurred with the code completion and "
```
ccordoba12: Please remove `_` and just leave the text with the parenthesis.

spyder/app/mainwindow.py
```diff
@@ -2676,10 +2676,18 @@ def report_issue(self, body=None, title=None, open_webpage=False):
...
+                        "linting services provided by the Python Language "
+                        "Server. Please submit this report to Github if you "
+                        "want to see this fixed.")
+        error_dialog.set_title_description(title, description)
```
ccordoba12: I think you misunderstood me. This needs to be

    error_dialog.set_title(title)
    error_dialog.set_description(description)

I'll explain below what these methods do.

spyder/plugins/editor/lsp/client.py
ccordoba12: Pinging @andfoy again about this.

spyder/widgets/reporterror.py
```diff
@@ -138,7 +138,7 @@ def __init__(self, parent=None, is_report=False):
             title = _("Please fill the following information")
         else:
             title = _("Spyder has encountered an internal problem!")
-        main_label = QLabel(
+        self.main_label = QLabel(
```
ccordoba12: Please revert this change. This is not needed.

spyder/widgets/reporterror.py
```diff
@@ -149,10 +149,10 @@ def __init__(self, parent=None, is_report=False):
...
               ).format(title=title, trouble_url=__trouble_url__,
                           project_url=__project_url__))
-        main_label.setOpenExternalLinks(True)
-        main_label.setWordWrap(True)
-        main_label.setAlignment(Qt.AlignJustify)
-        main_label.setStyleSheet('font-size: 12px;')
+        self.main_label.setOpenExternalLinks(True)
+        self.main_label.setWordWrap(True)
+        self.main_label.setAlignment(Qt.AlignJustify)
+        self.main_label.setStyleSheet('font-size: 12px;')
```
ccordoba12: Please revert these changes too.

spyder/widgets/reporterror.py
```diff
@@ -345,6 +345,12 @@ def _contents_changed(self):
                               title_chars >= TITLE_MIN_CHARS)
         self.submit_btn.setEnabled(submission_enabled)
 
+    def set_title_description(self, title, description):
+        """Set a title and description to the window."""
+        self.main_label.setText(
+            _("<h3>{title}</h3>"
+              "{description}").format(title, description))
```
ccordoba12: Ok, this is not what I have in mind. Please change this to

```python
def set_title(self, title):
    self.title.setText(title)

def set_description(self, description):
    self.input_description.setPlainText(description)
```

The idea is to automatically fill the title and description of the report because we already know what this error is about.

spyder/plugins/editor/lsp/client.py
steff456: Then we should send the signal inside the next if block (`if 'method' in resp:`) ?

spyder/widgets/reporterror.py
steff456: I understand now what you meant :o

spyder/plugins/editor/lsp/client.py
ccordoba12: No, the signal is fine where it is. The thing is I really don't need to have:

```python
if 'error' in resp:
   ...
if 'method' in resp:
   ...
```

Instead, I think it should be

```python
if 'error' in resp:
   ...
elif 'method' in resp:
   ...
```

spyder/widgets/reporterror.py
ccordoba12: Great!!

spyder/widgets/reporterror.py
```diff
@@ -222,7 +222,7 @@ def __init__(self, parent=None, is_report=False):
 
         # Main layout
         layout = QVBoxLayout()
-        layout.addWidget(main_label)
+        layout.addWidget(self.main_label)
```
ccordoba12: And this is not needed either.

steff456: Reverting all that stuff !

spyder/plugins/editor/lsp/client.py
steff456: I think that we are able to make that change

andfoy: @steff456 is completely right

ccordoba12: Great, thanks!

spyder/plugins/editor/lsp/client.py
```diff
@@ -239,8 +243,10 @@ def on_msg_received(self):
                 if 'error' in resp:
                     logger.debug('{} Response error: {}'
                                  .format(self.language, repr(resp['error'])))
+                    if self.language == 'python':
+                        self.sig_server_error.emit(repr(resp['error']))
```
ccordoba12: Please remove this blank line.

ccordoba12:
This is not ready yet. I'll finish it and let you know when you can review it.

PR #9274, opened 2019-05-05
PR: Shutdown running kernels more forcefully so they close faster

impact27:
## Description of Changes

Everytime I close a console that is running, I have to wait 5s for anything to happen. Not anymore.

### Issue(s) Resolved

Fixes #

### Affirmation

Impact27

### DIFF
--- a/spyder/plugins/ipythonconsole/widgets/client.py
+++ b/spyder/plugins/ipythonconsole/widgets/client.py
@@ -495,7 +495,7 @@ class ClientWidget(QWidget, SaveHistoryMixin):
     def shutdown(self):
         """Shutdown kernel"""
         if self.get_kernel() is not None and not self.slave:
-            self.shellwidget.kernel_manager.shutdown_kernel()
+            self.shellwidget.kernel_manager.shutdown_kernel(now=True)
         if self.shellwidget.kernel_client is not None:
             background(self.shellwidget.kernel_client.stop_channels)
 


goanpeca:
@impact27 thanks for this!, now:

- What is the difference in terms of what really goes on?
- Could you point to docs where this is explained a bit on the jupyter side?

impact27:
from https://jupyter-client.readthedocs.io/en/stable/api/manager.html

> now (bool) – Should the kernel be forcible killed now. This skips the first, nice shutdown attempt.

impact27:
In `/spyder/plugins/ipythonconsole/plugin.py` `close_client`, the following line is called:
`client.stop_button_click_handler()`
So this is the expected behaviour

impact27:
I don't see any usecase where it makes sense to wait 5s when calling the shutdown function.

goanpeca:
>Should the kernel be forcible killed now. This skips the first, nice shutdown attempt.

@impact27 I agree, I am just a bit concerned what the side effects of `not clean` are

ccordoba12:
> I am just a bit concerned what the side effects of not clean are

Probably the kernel connection file is not removed. If that's the case, I'm a bit -1 on this PR.

impact27:
I mean the only thing this does it instead of waiting 5 seconds and then kill the kernel, it just goes straight to killing the kernel. If there is any side effect, it already affects Spyder anytime a user closes a running kernel. But now the user has to loose 5s in addition, which is a pretty bad side effect.

goanpeca:
@ccordoba12 

Screen Shot 2019-05-05 at 15 34 02

ccordoba12:
This is breaking our Cython integration. @impact27, please see what happens in that case, or skip passing `now` when the client is a Cython one.

ccordoba12:
Ok, it seems our CIs had some glitches before because our tests are passing now. Thanks @impact27 for forcing a rerun.

PR #9280, opened 2019-05-06
PR: Avoid the possibility of duplicate preferences dialogs

### LINKED ISSUE #9281: Avoid the possibility of duplicate preferences dialogs
I noticed on OSX that pressing the shortcut several times in a row would display several Preferences dialogs.

Screen Shot 2019-05-06 at 18 14 49

goanpeca:
Fixes #9281

## Description of Changes

I noticed on OSX that pressing the shortcut several times in a row would display several Preferences dialogs.

This checks if one is already open to prevent duplicate instances.

## Tasks

* [x] Reorganized the code a bit
* [ ] Wrote tests? @ccordoba12 hmm not sure what could be tested...

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -2910,6 +2910,8 @@ class MainWindow(QMainWindow):
                         if widget is not None:
                             dlg.add_page(widget)
                     except Exception:
+                        # Avoid a crash at startup if a plugin's config
+                        # page fails to load.
                         traceback.print_exc(file=sys.stderr)
 
             if self.prefs_index is not None:
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -421,6 +421,7 @@ class MainWindow(QMainWindow):
                               RunConfigPage, LSPManagerConfigPage]
         self.prefs_index = None
         self.prefs_dialog_size = None
+        self.prefs_dialog_instance = None
 
         # Quick Layouts and Dialogs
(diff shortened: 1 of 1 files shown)


spyder/app/mainwindow.py
```diff
@@ -2874,32 +2875,50 @@ def apply_statusbar_settings(self):
...
+                           self.projects, self.ipyconsole,
+                           self.historylog, self.help, self.variableexplorer,
+                           self.onlinehelp, self.explorer, self.findinfiles
+                           ] + self.thirdparty_plugins:
+                if plugin is not None:
+                    try:
+                        widget = plugin.create_configwidget(dlg)
+                        if widget is not None:
+                            dlg.add_page(widget)
+                    except Exception:
```
CAM-Gerlach: Can we catch a more specific exception, and/or include a comment explaining why ``Exception`` is caught here?

goanpeca: I don't really know what was the use for it so I cannot explain

ccordoba12: The exception is fine because it avoids a crash at startup when the config page for a plugin fails to be created.

ccordoba12:
```suggestion
                    except Exception:
                        # Avoid a crash at startup if a plugin's config page
                        # fails to load.
```

goanpeca:
Thanks @CAM-Gerlach and @ccordoba12 for the review

PR #9290, opened 2019-05-08
PR: Correctly handle programming language in calltips

### LINKED ISSUE #9194: Pass language from console and editor on calltip call
On spyder/widgets/mixins.py

```python
    _PARAMETER_HIGHLIGHT_COLOR = '#DAA520'
    _DEFAULT_TITLE_COLOR = '#2D62FF'
    _CHAR_HIGHLIGHT_COLOR = 'red'
    _DEFAULT_TEXT_COLOR = '#999999'
```

and 

```python
# TODO: Select language and pass it to call
```
----

@dalthviz we need to be able to ask the kernel what language it is running (right now python, but it could be anything)

spyder/plugins/ipythonconsole/widgets/help.py
```python
# TODO: Pass the language from the Console to the calltip
```

dalthviz:
## Description of Changes

### Issue(s) Resolved

Fixes #9194

### Affirmation


ccordoba12:
@dalthviz, I released qtconsole 4.5.0 a couple fo weeks ago, which your addition of `language_name`. So please add that requirement to our `setup.py` and mention it in our Readme.

After doing that, do you think this will be ready? Or do you plan to add more things here?

dalthviz:
@ccordoba12 I left the PR as WIP since I was unsure if there was any other place to make changes. I will do the `setup.py` and `README` changes

spyder/widgets/mixins.py
```diff
@@ -327,7 +327,7 @@ def handle_sub(matchobj):
 
         return title
 
-    def show_calltip(self, signature, parameter=None):
+    def show_calltip(self, signature, parameter=None, language="python"):
```
ccordoba12: You probably need to pass `language` to the call to `_format_signature` inside this method.

But @goanpeca knows more about it. @goanpeca?

ccordoba12:
@dalthviz, your changes broke calltips in the IPython console. Please take a look at that.

spyder/widgets/mixins.py
```diff
@@ -53,7 +53,7 @@ class BaseEditMixin(object):
     _DEFAULT_TITLE_COLOR = '#2D62FF'
     _CHAR_HIGHLIGHT_COLOR = 'red'
     _DEFAULT_TEXT_COLOR = '#999999'
-    _DEFAULT_LANGUAGE = 'python'
+    _DEFAULT_LANGUAGE = ''
```
ccordoba12: I think we should use `'python'` here. Why did you decide to leave it empty?

dalthviz: I was checking the use of `getattr` for trying to get the `language` and the default values used are the empty `string` or `None`, should I revert the change and use `python` instead?

ccordoba12: Yeah, please do. I think it'd be easier to understand the code that way.

PR #9310, opened 2019-05-09
PR: Restrict PyLS version to be less than 0.25

ccordoba12:
That's because 0.25 added support for Pylint as a linter but now it it's not possible to turn it off.

Pinging @stonebig about it.


stonebig:
ok

PR #9320, opened 2019-05-10
PR: Reimplement fallback plugin for code completions

andfoy:
## Description of Changes
Add the fallback plugin to complete words, even if an LSP server is not available to do the corresponding completion. This applies to all source/plain text files.

ezgif-4-ffacc45d3ca8

### Issue(s) Resolved

See #7724

### Affirmation

@andfoy


goanpeca:
@andfoy you could use this instead

https://github.com/spyder-ide/spyder/blob/master/spyder/utils/workers.py

(there are some examples in there)

spyder/plugins/editor/fallback/fallback.py
```diff
@@ -0,0 +1,40 @@
...
+import re
+import logging
+from queue import Queue
+from threading import Thread, Lock
+
+logger = logging.getLogger(__name__)
+
+# Get all valid tokens that start by a letter and are
+# followed by a sequence of letters, numbers or underscores of length > 0
+regex = re.compile(r'[a-zA-Z]\w+')
```
goanpeca: Hmmm this might be extended, since for instance on CSS names usually use `-`

andfoy:
> https://github.com/spyder-ide/spyder/blob/master/spyder/utils/workers.py

I might better reimplement this as a ``QObject``, in the same fashion of ``LSPClient``

spyder/plugins/editor/fallback/fallback.py
andfoy: We could add ``-`` to the pattern, but I'm wondering what would happen in this scenario: ``var1 = var2-var3``, a subtraction without whitespace around the operator

andfoy: Another solution would be to define another regex for languages/plain text that allow ``kebab-case`` (CSS, LISP, Perl) and other for the rest of languages that use ``snake_case`` or ``camelCase``

goanpeca: Hmmm well for now I would just check if it is CSS os SCSS or HTML or XML ?

andfoy:
@ccordoba12 This one is only missing tests

ccordoba12:
@andfoy, although I clearly instructed you to use our previous code, you implemented this from scratch. So I have some questions and observations regarding that:

1. Why didn't you use the Worker architecture to handle non-blocking operations that we already have in place? Unless there's a very strong reason not to do it, I'd really prefer fallback completions to be based on that.
2. Your code is not providing the keywords Pygments gives for different programming languages. That was done in the `get_completions` method of `FallbackPlugin`, so please use that code here too.
3. Words or tokens for a file are given by the function `get_words` present in `utils/introspection/utils.py`. And that function is very well tested too. So please merge your work to get tokens with that and use our previous tests too.
5. You need to restore the functionality of `FallbackPlugin` to get definitions and info of Python-like files. That was working for Enaml (which we would like to keep having) and we could also make it work for Cython (but that's for another PR). If that proves to be to much for this PR, then we'll leave it for later.

andfoy:
> Why didn't you use the Worker architecture to handle non-blocking operations that we already have in place? Unless there's a very strong reason not to do it, I'd really prefer fallback completions to be based on that.

The idea here is to implement the fallback completion engine such that it looks like an LSP server. By doing so, we uniformize our completion code (Fallback updates documents on updates, and returns completions on request).

The implementation is based on the [actor](https://en.m.wikipedia.org/wiki/Actor_model) used in concurrent programming. Actors act as processes that do not have a shared state and that can communicate via message passing. This is the standard model used in concurrent programming languages, such as Erlang/Elixir and Clojure. Actors are designed to prevent the pitfalls of plain thread, shared-state programming such as deadlocks, race conditions and process starvation.

Under this setup, we maintain a fallback actor that has a copy of the different editor files. The editor comunicates the changes to the actor and requests completions when necessary. 

> Your code is not providing the keywords Pygments gives for different programming languages. That was done in the get_completions method of FallbackPlugin, so please use that code here too.

We could match Pygments lexer here with our tokens.

ccordoba12:
> The idea here is to implement the fallback completion engine such that it looks like an LSP server

Ok, thanks for the explanation. Things are clearer now.

> We could match Pygments lexer here with our tokens.

I think that's done already. What I meant is that Pygments gives some keywords for the programming languages it can tokenize (e.g. `else`, `elif`, etc in Bash), and we were showing those keywords with our fallback plugin before, even if they are not present in the file.

Since that functionality is already in place, please add that to the completions FallbackActor returns.

andfoy:
It now looks like this, after adding pygments completions:
Peek 14-05-2019 16-54

spyder/app/mainwindow.py
```diff
@@ -884,6 +884,11 @@ def create_edit_action(text, tr_text, icon):
         from spyder.plugins.editor.lsp.manager import LSPManager
         self.lspmanager = LSPManager(self)
 
+        # Fallback completion thread
+        self.set_splash(_("Creating fallback completion engine"))
+        from spyder.plugins.editor.utils.fallback import FallbackActor
+        self.fallback = FallbackActor(self)
```
ccordoba12: Please call this attribute `fallback_completions` instead of `fallback` so it's easier to understand.

spyder/plugins/editor/utils/fallback.py
```diff
@@ -0,0 +1,116 @@
+
```
ccordoba12: Please create a module called ` spyder/plugins/editor/fallback` and move this module inside it with the name `actor`.

spyder/plugins/editor/utils/lsp.py
```diff
@@ -17,10 +17,10 @@ def request(req=None, method=None, requires_response=True):
 
     @functools.wraps(req)
     def wrapper(self, *args, **kwargs):
-        if self.lsp_ready:
-            params = req(self, *args, **kwargs)
-            if params is not None:
-                self.emit_request(method, params, requires_response)
+        # if self.lsp_ready:
```
ccordoba12: Please remove this comment if it's not needed anymore.

spyder/utils/introspection/utils.py
```diff
@@ -184,14 +184,20 @@ def find_lexer_for_filename(filename):
 def get_keywords(lexer):
```
ccordoba12: Please move this file to ` spyder/plugins/editor/fallback/utils.py`

ccordoba12:
Please update the `default_info_response` in `spyder/plugins/help/tests/test_plugin.py` to fix the tests.

ccordoba12:
Now please remove `spyder/utils/introspection/tests/test_utils.py` because it's not needed anymore.

continuous_integration/azure/install.bat
```diff
@@ -27,6 +27,9 @@ if %USE_CONDA% == yes (
 ::    conda install -q -y python=3.6.8=h9f7ef89_0
 ::)
 
+:: Install diff-match-patch
+pip install -q diff-match-patch
```
ccordoba12: Please remove this from here and instead add `diff-match-patch` to `requirements/conda.txt`. Also add a mention to this package in our Readme.

continuous_integration/azure/install.sh
```diff
@@ -8,6 +8,9 @@ if [ "$USE_CONDA" = "yes" ]; then
     # Install test ones
     conda install -q -y -c spyder-ide --file requirements/tests.txt
 
+    # Install diff-match-patch
+    pip install -q diff-match-patch
```
ccordoba12: Remove this too.

continuous_integration/travis/install.sh
```diff
@@ -26,6 +26,9 @@ if [ "$USE_CONDA" = "yes" ]; then
     # Install coveralls
     pip install -q coveralls
 
+    # Install diff-match-patch
+    pip install -q diff-match-patch
```
ccordoba12: And also this.

andfoy:
@ccordoba12 This one is ready

ccordoba12:
@andfoy, I thought we agreed on adding a `fallback_codeeditor` fixture to test the fallback completions in an actual CodeEditor. But I don't see that as part of this PR.

andfoy:
@ccordoba12, tests are passing locally on my machine, but not in CI, do you have any idea?

ccordoba12:
@andfoy, please continue with issue #9007 while i give you a hand here.

ccordoba12:
@andfoy, I helped you to solve the problems on Linux, but the hanging on Windows when running `test_fallback_completions` is real (I tested it locally and tests don't pass from that point). That should be the same problem that's happening on macOS.

So please run our tests locally in a VM and see how to solve that issue.

ccordoba12:
@andfoy, forget about my last comment. I spent several hours trying to make `test_fallback_completions` work on Windows but I couldn't. However, I can confirm things are working well on a Spyder instance.

PR #9324, opened 2019-05-11
PR: Have LSP update after closequotes and closebrackets

### LINKED ISSUE #9323: Close brackets or close quotes doesn't update the LSP
### Issue Report Checklist

* [X] Searched the [issues page](https://github.com/spyder-ide/spyder/issues?q=is%3Aissue) for similar reports
* [X] Read the relevant sections of the [Spyder Troubleshooting Guide](https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ) and followed its advice
* [X] Reproduced the issue after updating with ``conda update spyder`` (or ``pip``, if not using Anaconda)
    * [ ] Restarted Spyder

## Problem Description

Adding brackets or quotes with close_brackets or close_quotes enabled doesn't update the code error checking
`print("sweet)`

1. `print("sweet)` gives error EOL sting literal error
2. type a quote to make `print("sweet")`
3. The error stays

No errors!

```python-traceback

PASTE TRACEBACK HERE

```

## Versions

* Spyder version: master
* Python version: 3.6
* Qt version: 
* PyQt version: 
* Operating System name/version: Linux

### Dependencies

```
PASTE DEPENDENCIES HERE
```

bcolsen:
## Description of Changes

Add document_did_change to closequotes and closebrackets so the LSP will update.

The LSP would remove errors after typing brackets or quotes to correct them. Now it does.

### Issue(s) Resolved

Fixes #9323

### Affirmation

I certify the above statement is true and correct: bcolsen

### DIFF
--- a/spyder/plugins/editor/widgets/tests/test_warnings.py
+++ b/spyder/plugins/editor/widgets/tests/test_warnings.py
@@ -8,7 +8,6 @@
 
 # Stdlib imports
 import os
-from time import sleep
 
 # Third party imports
 import pytest
--- a/spyder/plugins/editor/widgets/tests/test_warnings.py
+++ b/spyder/plugins/editor/widgets/tests/test_warnings.py
@@ -187,19 +187,14 @@ def test_update_warnings_after_closequotes(qtbot, lsp_codeeditor):
     Regression test for #9323.
     """
     editor, _ = lsp_codeeditor
-    
-    kwargs = {}
-    kwargs['language'] = 'Python'
-    kwargs['close_quotes'] = True
-    editor.setup_editor(**kwargs)
-    
+
     # Test closing when following character is a right parentheses
     editor.textCursor().insertText('foo()')
     editor.move_cursor(-1)
     qtbot.keyClicks(editor, '"')
     assert editor.toPlainText() == 'foo("")'
     assert editor.textCursor().columnNumber() == 5
-    
+
     editor.textCursor().insertText("\nprint('test)\n")
 
     expected = [['EOL while scanning string literal', 2]]
--- a/spyder/plugins/editor/widgets/tests/test_warnings.py
+++ b/spyder/plugins/editor/widgets/tests/test_warnings.py
@@ -209,7 +209,7 @@ def test_update_warnings_after_closequotes(qtbot, lsp_codeeditor):
         sleep(1)
 
     # Assert that the error is gone.
-    expected = []
+    expected = [['D100: Missing docstring in public module', 1]]
     assert editor.get_current_warnings() == expected
 
 
@@ -243,5 +243,5 @@ def test_update_warnings_after_closebrackets(qtbot, lsp_codeeditor):
         sleep(1)
 
     # Assert that the error is gone.
-    expected = []
+    expected = [['D100: Missing docstring in public module', 1]]
     assert editor.get_current_warnings() == expected
--- a/spyder/plugins/editor/widgets/tests/test_warnings.py
+++ b/spyder/plugins/editor/widgets/tests/test_warnings.py
@@ -200,6 +200,7 @@ def test_update_warnings_after_closequotes(qtbot, lsp_codeeditor):
     # Add a single quote to fix the error
     editor.move_cursor(-2)
     qtbot.keyClicks(editor, "'")
+    assert editor.toPlainText() == "print('test')\n"
 
     # Wait for the lsp_response_signal.
     qtbot.waitSignal(editor.lsp_response_signal, timeout=30000)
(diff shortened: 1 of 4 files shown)


bcolsen:
I got thinking about this when I saw #9300. The tests for that PR were put in test_warnings is there a similar test_error file? Quote and brackets tend to make errors not warnings.

We might want this in beta2

goanpeca:
@bcolsen could we add a test for this?

Maybe add something here https://github.com/spyder-ide/spyder/blob/master/spyder/plugins/editor/widgets/tests/test_warnings.py ?

bcolsen:
@goanpeca That would work but I can't think of a warning that would involve quotes or brackets. You usually get errors in that case.

goanpeca:
I don't follow... those tests (the file I attached) include warnigns and erros etc...

bcolsen:
> I don't follow... those tests (the file I attached) include warnings and errors etc... (not only "warnings")

Sorry for the confusion. I thought there was some differentiation between warnings and errors in the code, but `editor.get_current_warnings()` gets both so that works.

ccordoba12:
Thanks a lot @bcolsen! Please fix the tests you added so we can include this one in beta2.

bcolsen:
I can't get the test to run locally on my computer. It seem like the LSP service doesn't start because it hits the time out error. Can you think of anything I might be missing?

ccordoba12:
I restricted the lastet version of the PyLS we support to be less than 0.25 in the last couple of days. So that could be what's not allowing you to run our tests locally.

spyder/plugins/editor/widgets/tests/test_warnings.py
```diff
@@ -175,3 +177,71 @@ def test_update_warnings_after_delete_line(qtbot, lsp_codeeditor):
...
+    assert editor.get_current_warnings() == expected
+
+    # Add a single quote to fix the error
+    editor.move_cursor(-2)
+    qtbot.keyPress(editor, Qt.Key_Apostrophe)
+    assert editor.toPlainText() == "print('test')\n"
+
+    # Wait for the lsp_response_signal.
+    with qtbot.waitSignal(editor.lsp_response_signal, timeout=30000):
+        sleep(1)
```
ccordoba12: Why is this necessary? I think instead of this you need to run `editor.document_did_change()`.

bcolsen: I'm not entirely  sure honestly but in my mind I'm  waiting for the server to be done processing.

> I think instead of this you need to run editor.document_did_change()

The only change I made to the the spyder code other than tests was adding editor.document_did_change() to close quotes and brackets. I don't  know how editor.document_did_change() works but if I just call editor.document_did_change() here wouldn't that defeat the purpose of the test?

goanpeca: I agree with @bcolsen. I think if we are using `editor.document_did_change()` on the tests, we are probably doing something wrong as the editor should be doing this "on its own". 

Now the way to use the context blockers should be:

```python
with qtbot.waitSignal(editor.lsp_response_signal, timeout=30000):
    # do something (add a quote.... remove a quote) no need to add waits
```

I suggest you put the code that triggers something inside the `with` instead of document change or a wait. The code inside the with should be triggering some signal.

bcolsen: I tried that first, but I couldn't get it to fail and I needed to add the sleep in side as well to make it pass.

ccordoba12: @bcolsen, ok, you are the one who fought with the test, so I defer to you on this one. Please add a comment above this line then, explaining why it's needed.

Also, why didn't you use `qtbot.wait` instead?

bcolsen: I didn't try qtbot.wait. That should work.

ccordoba12: Yeah, I think it's better because it's non-blocking (while `sleep` is).

goanpeca: Yep! correct. 

@bcolsen I will check later what might be the problem that requires manually calling `document_did_change` Thanks for the work here!

bcolsen:
I tried the tests on my work computer today and it works with the code in the `with` block now. That makes the test cleaner.  However, when I remove the `document_did_change` from the extensions the test just stalls and times out at the `with` statement instead of returning the error.

I think this might might indicate something deeper and that manually calling `document_did_change` in the extensions might just be a workaround, but does work :-)

spyder/plugins/editor/widgets/tests/test_warnings.py
```diff
@@ -8,9 +8,11 @@
 
 # Stdlib imports
 import os
+from time import sleep
```
ccordoba12: Please remove this line because it's not needed anymore. The rest looks good.

goanpeca:
Great work! :-)

bcolsen:
Thank you. That was a rough "one line change"  ;-)

Issue #9265, opened 2019-05-04
Autosaves are created for unmodified files and not removed on close with split Editor panes


CAM-Gerlach:
## Problem Description

I can confirm that with the latest ``master``, I'm still getting the ``Recover from autosave`` for files that were either saved or never unmodified when I closed Spyder (via the normal close button), with a clean ``autosave`` directory  and not other Spyders open before or currently. All the files I have open have autosaves created 1 minute after startup even if I've never touched Spyder, and they don't get removed when I close Spyder again. This happens consistently, every time with every file I either have open on startup or open subsequently, regardless of whether I view, modify, etc. it. 

I *finally* reproduced it starting from a clean ``--safe-mode`` Spyder instance by opening a two-pane layout ("horizontal split") and a few files and waiting 1 minute; it seems to be the two pane layout that triggers it, as it does not happen with only one pane open. I've tested enabling/disabling code style, docstring style analysis, projects,  and more, and none of them have an effect on whether it happens or not, just the two pane layout.

I also thought they should be removed by spyder-ide/spyder#8733 anyway, but apparently that handles it by merely removing the autosave when the "Don't save" button on the "Unsaved changes" dialog when closing the file is clicked, rather than when the file is actually closed (e.g. by the user, closing Spyder, switching projects, etc), and thus was not robust against any bugs (like that one) that might result in an autosave file being created for a reason other than an actual modification. 

### What steps reproduce the problem?

1. Open a clean copy of Spyder ``master``
2. Select "Split panes horizontally"
3. Open a file
4. Wait 60 seconds

## Versions

* Spyder version: ``master`` as of post time (``c90e2aea``)
* Python version: 3.7.2
* Qt version: 5.9.7
* PyQt version: 5.9.2
* Operating System name/version: Windows 8.1 x64

### Dependencies

```
pygments >=2.0    :  2.3.1 (OK)
qdarkstyle >=2.6.4:  2.6.5 (OK)
sphinx >=0.6.6    :  1.8.4 (OK)
pyls >=0.19.0     :  0.21.5 (OK)
nbconvert >=4.0   :  5.4.0 (OK)
pandas >=0.13.1   :  0.24.2 (OK)
numpy >=1.7       :  1.15.4 (OK)
sympy >=0.7.3     :  1.3 (OK)
cython >=0.21     :  0.29.4 (OK)
qtconsole >=4.2.0 :  4.4.3 (OK)
IPython >=4.0     :  7.2.0 (OK)
[...]

PR #9327, opened 2019-05-11
PR: Remove autosaves on successful file close to partially mitigate spurious creation bugs

CAM-Gerlach:
## Description of Changes

Partially mitigates the impact of #9265 .

It was much more complex than initially anticipated, but with this PR Spyder now removes autosave files after successfully closing them (either choosing not to save, or with no changes present) if an autosave exists, either when closing individual files or when closing the editor entirely. In order to do the latter, since the action is not already synchronized between panes like closing files and the autosaves are not in the ``name_mapping`` dict of the first split pane, I just iterate through all the EditorStacks, ignoring errors (which will occur if there are more than two, or if another Spyder instance removes them first), to make sure all autosaves are removed.

I also added both unit (autosave) and functional (editor) tests of the ``remove_all_autosave_files()`` function I added, as well as improved the existing tests of the ``remove_autosave_file()`` function, and refactored the latter to take a filename directly rather than an finfo object to greatly simplify the former and because only the filename was ever actually used.

Note: This should *not* close the linked issue, because it doesn't resolve the root of the problem and there are other, less serious side effects this mitigation is inherently not able to address directly.

### Affirmation

I certify the above statement is true and correct: CAM_Gerlach

### DIFF
--- a/spyder/plugins/editor/plugin.py
+++ b/spyder/plugins/editor/plugin.py
@@ -354,7 +354,7 @@ class Editor(SpyderPluginWidget):
 
         active_project_path = None
         if self.projects is not None:
-             active_project_path = self.projects.get_active_project_path()
+            active_project_path = self.projects.get_active_project_path()
         if not active_project_path:
             self.set_open_filenames()
         else:
@@ -377,6 +377,14 @@ class Editor(SpyderPluginWidget):
             else:
                 for win in self.editorwindows[:]:
                     win.close()
+                # Remove autosave on successful close to avoid issue #9265.
+                # Probably a good idea anyway to mitigate autosave issues.
+                # Ignore errors resulting from file being open in > 2
+                # editorstacks or another Spyder being closed first and
+                # removing the spurious files.
+                for editorstack_split in self.editorstacks:
+                    editorstack_split.autosave.remove_all_autosave_files(
+                        errors="ignore")
                 return True
         except IndexError:
             return True
--- a/spyder/plugins/editor/utils/autosave.py
+++ b/spyder/plugins/editor/utils/autosave.py
@@ -151,28 +151,42 @@ class AutosaveForStack(object):
                 autosave_filename = osp.join(autosave_dir, autosave_basename)
         return autosave_filename
 
-    def remove_autosave_file(self, fileinfo):
+    def remove_autosave_file(self, filename, errors='raise'):
         """
         Remove autosave file for specified file.
 
         This function also updates `self.autosave_mapping` and clears the
         `changed_since_autosave` flag.
         """
-        filename = fileinfo.filename
         if filename not in self.name_mapping:
             return
         autosave_filename = self.name_mapping[filename]
         try:
             os.remove(autosave_filename)
         except EnvironmentError as error:
-            action = (_('Error while removing autosave file {}')
-                      .format(autosave_filename))
-            msgbox = AutosaveErrorDialog(action, error)
-            msgbox.exec_if_enabled()
+            error_message = (_('{} while removing autosave file {}')
+                             .format(type(error), autosave_filename))
+            if errors == 'ignore':
+                logger.debug('%s : %s', error_message, str(error))
+            else:
+                msgbox = AutosaveErrorDialog(error_message, error)
+                msgbox.exec_if_enabled()
         del self.name_mapping[filename]
         self.stack.sig_option_changed.emit(
-                'autosave_mapping', self.name_mapping)
-        logger.debug('Removing autosave file %s', autosave_filename)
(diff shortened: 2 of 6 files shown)


spyder/plugins/editor/utils/autosave.py
```diff
@@ -151,28 +151,42 @@ def create_unique_autosave_filename(self, filename, autosave_dir):
...
         del self.name_mapping[filename]
         self.stack.sig_option_changed.emit(
-                'autosave_mapping', self.name_mapping)
-        logger.debug('Removing autosave file %s', autosave_filename)
+            'autosave_mapping', self.name_mapping)
+        logger.debug('Removed autosave file %s', autosave_filename)
+
+    def remove_all_autosave_files(self, errors='raise'):
+        """
+        Remove all autosave files stored in this componenet's mapping.
```
ccordoba12: componenet's -> component's

spyder/plugins/editor/plugin.py
```diff
@@ -377,6 +377,14 @@ def closing_plugin(self, cancelable=False):
             else:
                 for win in self.editorwindows[:]:
                     win.close()
+                # Remove autosave file on successful close to avoid #9265
```
ccordoba12: ... to avoid *issue* 9265 ...

ccordoba12: This is our convention to reference issues.

spyder/plugins/editor/widgets/editor.py
```diff
@@ -1505,6 +1505,10 @@ def close_file(self, index=None, force=False):
 
             self.add_last_closed_file(finfo.filename)
 
+            # Remove autosave file on successful close to work around #9265
```
ccordoba12: ... around *issue* 9265 ...

spyder/plugins/editor/plugin.py
CAM-Gerlach: Done

spyder/plugins/editor/widgets/editor.py
CAM-Gerlach: Done

spyder/plugins/editor/utils/autosave.py
CAM-Gerlach: I can't spell, lol...new feature request: spellchecker for docstrings.

CAM-Gerlach:
@ccordoba12 Updated with the requested changes, thanks.

@jitseniesen I'd feel much better about my hack job if you could give this a review before it goes in, since autosave is your "baby"...

spyder/plugins/editor/widgets/editor.py
goanpeca: @cam-gerlach could you add this comment on hiw we reference issues somewhere on the wiki or even on the PR template ? 

Thanks! Great work

CAM-Gerlach: @goanpeca I reorganized the [Code style](https://github.com/spyder-ide/spyder/wiki/Dev:-Coding-Style) page and added this along with a number of other standards and conventions

jitseniesen:
This is fine for beta but I want to give it more thought for Spyder 4. I am afraid that this PR removes autosave files in some circumstances where we don't want to remove them. Specifically, what if `close_file()` is called with `force=True`, and what if a file is saved unsuccessfully?

CAM-Gerlach:
> Specifically, what if `close_file()` is called with `force=True`, and what if a file is saved unsuccessfully?

@jitseniesen The only place this is currently done in the codebase that I could find is ``close_file_in_all_editorstacks``, which is called to close the file in split panes other than the one the user clicked after saving/closing the file (without force) and either successively saved it, chose not to save or the file was unmodified. If the user choses to cancel, or saving doesn't complete, this function shouldn't ever fire at all, whereas if they click no or the save completes (or the file is unmodified) the autosave should be removed as per the existing logic you implemented.

If we simply didn't remove the autosave if ``force=True``,the problem is the autosave doesn't exist in the name mapping of the first editorstack (only the other ones) so it presumably wouldn't get removed at all if the user clicked "close" in the left/top pane, whereas it seems it shouldn't be possible to delete it if saving were to not complete.

However, if further assurance is desired, I can just perform a pre-check in ``close_file`` before ``force``/``save_if_changed`` that the file is unmodified and only remove the autosave if that's the case, which should be mostly true if this bug is triggered (the only edge case I can think of right now would be modifying an unmodified file within the 60 second window, then closing it from the left pane). Of course, once #9265 is fixed, that won't be an issue anymore.

jitseniesen:
> I can just perform a pre-check in `close_file` before `force`/`save_if_changed` that the file is unmodified

I don't think that is necessary.

PR #9301, opened 2019-05-08
PR: Fix hover/calltip for Python objects without signature

goanpeca:
## Description of Changes

* [x] Wrote at least one-line docstrings (for any new functions)
* [x] Added unit test(s) covering the changes (if testable)

* [x] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

hints

### Issue(s) Resolved

Fixes #9294

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct: @goanpeca

PR #9334, opened 2019-05-12
PR: Fix showing signatures in the IPython console after an open paren

ccordoba12:
This is a regression of PR #9301.

### DIFF
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -428,6 +428,24 @@ def test_get_help_ipython_console(main_window, qtbot):
 @flaky(max_runs=3)
 @pytest.mark.skipif(os.name == 'nt' and os.environ.get('CI') is not None,
                     reason="Times out on AppVeyor")
+def test_get_calltip_ipython_console(main_window, qtbot):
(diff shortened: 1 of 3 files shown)


ccordoba12:
Can you help me with that (tomorrow, I mean)? I don't have time to devise one right now, but I consider it very important to avoid this in the future.

You can push it to this branch instead of opening a new PR.

goanpeca:
Ok will do

goanpeca:
Added a test to check the regression @ccordoba12

goanpeca:
I think this test needs the help plugin since that is where the actual call to get the signature is. No?

ccordoba12:
Nop, calltips are handled by the console alone. And we already have a test for displaying help from the IPython console. It's called `test_get_help_ipython_console` and it's present in `test_mainwindow.py`.

ccordoba12:
I think this ready. Thanks a lot @goanpeca for your help with this one!!

goanpeca:
Thanks, I now understand why I thought the help plugin was involved. `help.py` name tripped me off 😆

ccordoba12:
Yeah, that's not a good name for things, but it's the one we decided to use.

PR #9343, opened 2019-05-14
PR: Use Jupyter comms to communicate between frontend and kernel

### LINKED ISSUE #9342: File not saved when calling runfile from console

When pressing the green "Run file" arrow, the file is automatically saved. It would be cool if the same appened when I type `runfile('file.py')` in the console.

## Versions

* Spyder version: 4.0.0.dev0 f1daab7a4
* Python version: 3.7.3
* Qt version: 5.9.1
* PyQt5 version: 5.9
* Operating System: Darwin 18.5.0

### Dependencies

```
qdarkstyle >=2.6.4 :  2.6.5 (OK)
psutil >=0.3       :  5.6.1 (OK)
pyls >=0.19.0;<0.25:  0.25.0 (NOK)
qtconsole >=4.2.0  :  4.5.0.dev0 (OK)
IPython >=4.0      :  7.3.0 (OK)
```

### LINKED ISSUE #9403: Add breakpoints while debugging
### Issue Report Checklist

    * [ ] Restarted Spyder

## Problem Description
When debugging, it is only possible to add a breakpoint if the prompt is present. 

1. debug
```
import time
for i in range(100):
    time.sleep(.1)
```
2. Press continue
3. Realise there is no breakpoint,  try to add a breakpoint on the ` time.sleep(.1)` line
4. See the debugger finish without stopping

I would like the debugger to stop.

## Versions

* Spyder version: 
* Python version: 
* Qt version: 
* PyQt version: 

### Dependencies

```
PASTE DEPENDENCIES HERE
```

impact27:
## Description of Changes

* [na] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

This PR allows the console to send informations to spyder IDE. In this case, it is used to save the editor before running `runfile()` or `debugfile()`. This goes together with https://github.com/spyder-ide/spyder-kernels/pull/111.

It also interrupts the kernel to add breakpoints while debugging.

### Issue(s) Resolved

Fixes #9403

### Affirmation

impact27

### DIFF
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -1864,7 +1864,8 @@ def test_report_comms_error(qtbot, main_window):
     """Test if a comms error is correctly displayed."""
     CONF.set('main', 'show_internal_errors', True)
     shell = main_window.ipyconsole.get_current_shellwidget()
-    qtbot.waitUntil(lambda: not shell._executing)
+    qtbot.waitUntil(lambda: shell._prompt_html is not None,
+                    timeout=SHELL_TIMEOUT)
     # Create a bogus get_cwd
     with qtbot.waitSignal(shell.executed):
         shell.execute('def get_cwd(): import foo')
@@ -1878,6 +1879,7 @@ def test_report_comms_error(qtbot, main_window):
     assert error_dlg is not None
     assert 'Exception in comms call get_cwd' in error_dlg.error_traceback
     assert 'ImportError: No module named' in error_dlg.error_traceback
+    CONF.set('main', 'show_internal_errors', False)
 
 
 if __name__ == "__main__":
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -1821,6 +1821,27 @@ def test_custom_layouts(main_window, qtbot):
                                 print(widget)  # spyder: test-skip
                                 assert widget.isVisible()
 
+# @pytest.mark.slow
+@flaky(max_runs=3)
+def test_save_on_runfile(main_window, qtbot):
+    """Test that layout are showing the expected widgets visible."""
+    # Load test file
+    test_file = osp.join(LOCATION, 'script.py')
+    test_file_copy = test_file[:-3] + '_copy.py'
+    shutil.copyfile(test_file, test_file_copy)
+    main_window.editor.load(test_file_copy)
+    code_editor = main_window.editor.get_focus_widget()
+    # Verify result
+    shell = main_window.ipyconsole.get_current_shellwidget()
+    qtbot.keyClicks(code_editor, 'test_var = 123', delay=100)
+    filename = code_editor.filename
+    with qtbot.waitSignal(shell.sig_prompt_ready):
+        shell.execute('runfile("{}")'.format(filename))
+
+    assert shell.get_value('test_var') == 123
+    main_window.editor.close_file()
+    os.remove(test_file_copy)
+
 
 # @pytest.mark.slow
 def test_pylint_follows_file(qtbot, tmpdir, main_window):
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -1879,6 +1879,7 @@ def test_report_comms_error(qtbot, main_window):
     assert error_dlg is not None
     assert 'Exception in comms call get_cwd' in error_dlg.error_traceback
     assert 'No module named' in error_dlg.error_traceback
+    main_window.console.close_error_dlg()
     CONF.set('main', 'show_internal_errors', False)
 
 
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -1860,5 +1860,25 @@ def test_pylint_follows_file(qtbot, tmpdir, main_window):
         assert fname == pylint_plugin.get_filename()
 
 
+def test_report_comms_error(qtbot, main_window):
+    """Test if a comms error is correctly displayed."""
+    CONF.set('main', 'show_internal_errors', True)
+    shell = main_window.ipyconsole.get_current_shellwidget()
+    qtbot.waitUntil(lambda: not shell._executing)
+    # Create a bogus get_cwd
+    with qtbot.waitSignal(shell.executed):
+        shell.execute('def get_cwd(): import foo')
+    with qtbot.waitSignal(shell.executed):
+        shell.execute("get_ipython().kernel.frontend_comm."
+                      "register_call_handler('get_cwd', get_cwd)")
+    with qtbot.waitSignal(shell.executed):
+        shell.execute('ls')
+
+    error_dlg = main_window.console.error_dlg
+    assert error_dlg is not None
+    assert 'Exception in comms call get_cwd' in error_dlg.error_traceback
+    assert 'ImportError: No module named' in error_dlg.error_traceback
+
+
 if __name__ == "__main__":
     pytest.main()
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -1821,27 +1821,6 @@ def test_custom_layouts(main_window, qtbot):
                                 print(widget)  # spyder: test-skip
                                 assert widget.isVisible()
 
-# @pytest.mark.slow
-@flaky(max_runs=3)
-def test_save_on_runfile(main_window, qtbot):
-    """Test that layout are showing the expected widgets visible."""
-    # Load test file
-    test_file = osp.join(LOCATION, 'script.py')
-    test_file_copy = test_file[:-3] + '_copy.py'
-    shutil.copyfile(test_file, test_file_copy)
-    main_window.editor.load(test_file_copy)
-    code_editor = main_window.editor.get_focus_widget()
-    # Verify result
-    shell = main_window.ipyconsole.get_current_shellwidget()
-    qtbot.keyClicks(code_editor, 'test_var = 123', delay=100)
-    filename = code_editor.filename
-    with qtbot.waitSignal(shell.sig_prompt_ready):
-        shell.execute('runfile("{}")'.format(filename))
-
-    assert shell.get_value('test_var') == 123
-    main_window.editor.close_file()
-    os.remove(test_file_copy)
-
 
 # @pytest.mark.slow
 def test_pylint_follows_file(qtbot, tmpdir, main_window):
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -1878,7 +1878,7 @@ def test_report_comms_error(qtbot, main_window):
     error_dlg = main_window.console.error_dlg
     assert error_dlg is not None
     assert 'Exception in comms call get_cwd' in error_dlg.error_traceback
-    assert 'ImportError: No module named' in error_dlg.error_traceback
+    assert 'No module named' in error_dlg.error_traceback
     CONF.set('main', 'show_internal_errors', False)
 
 
--- a/spyder/plugins/editor/widgets/tests/test_introspection.py
+++ b/spyder/plugins/editor/widgets/tests/test_introspection.py
@@ -60,6 +60,7 @@ def test_space_completion(lsp_codeeditor, qtbot):
 
 @pytest.mark.slow
 @pytest.mark.first
+@flaky(max_runs=5)
 def test_hide_widget_completion(lsp_codeeditor, qtbot):
     """Validate hiding completion widget after a delimeter or operator."""
     code_editor, _ = lsp_codeeditor
--- a/spyder/plugins/ipythonconsole/comms/__init__.py
+++ b/spyder/plugins/ipythonconsole/comms/__init__.py
@@ -1,0 +1,56 @@
+# -*- coding: utf-8 -*-
+# -----------------------------------------------------------------------------
+# Copyright (c) 2009- Spyder Kernels Contributors
+#
+# Licensed under the terms of the MIT License
+# (see spyder_kernels/__init__.py for details)
+# -----------------------------------------------------------------------------
+
+"""
+API to communicate between the Spyder IDE and the Spyder kernel.
+It uses Jupyter Comms for messaging. The messages are sent by calling an
+arbitrary function, with the limitation that the arguments have to be
+picklable. If the function must return, the call must be blocking.
+
+In addition, the frontend can interrupt the kernel to process the message sent.
+This allows, for example, to set a breakpoint in pdb while the debugger is
+running. The message will only be delivered when the kernel is checking the
+event loop, or if pdb is waiting for an input.
+
+Example:
+
+On one side:
+
+    ```
+    def hello_str(msg):
+        print('Hello ' + msg + '!')
+
+    def add(a, d):
+        return a + b
+
+    left_comm.register_call_handler('add_numbers', add)
+    left_comm.register_call_handler('print_hello', hello_str)
+    ```
+
+On the other:
+
+    ```
+    right_comm.remote_call().print_hello('world')
+    res =  right_comm.remote_call(blocking=True).add_numbers(1, 2)
+    print('1 + 2 = ' + str(res))
+    ```
+
+Which prints on the right side (The one with the `left_comm`):
+
+    ```
+    Hello world!
+    ```
+
+And on the left side:
+
+    ```
+    1 + 2 = 3
+    ```
+"""
+
+from spyder_kernels.comms.commbase import CommError
--- a/spyder/plugins/ipythonconsole/comms/__init__.py
+++ b/spyder/plugins/ipythonconsole/comms/__init__.py
@@ -7,50 +7,7 @@
 # -----------------------------------------------------------------------------
 
 """
-API to communicate between the Spyder IDE and the Spyder kernel.
-It uses Jupyter Comms for messaging. The messages are sent by calling an
-arbitrary function, with the limitation that the arguments have to be
-picklable. If the function must return, the call must be blocking.
-
-In addition, the frontend can interrupt the kernel to process the message sent.
-This allows, for example, to set a breakpoint in pdb while the debugger is
-running. The message will only be delivered when the kernel is checking the
-event loop, or if pdb is waiting for an input.
-
-Example:
-
-On one side:
-
-    ```
-    def hello_str(msg):
-        print('Hello ' + msg + '!')
-
-    def add(a, d):
-        return a + b
-
-    left_comm.register_call_handler('add_numbers', add)
-    left_comm.register_call_handler('print_hello', hello_str)
-    ```
-
-On the other:
-
-    ```
-    right_comm.remote_call().print_hello('world')
-    res =  right_comm.remote_call(blocking=True).add_numbers(1, 2)
-    print('1 + 2 = ' + str(res))
-    ```
-
-Which prints on the right side (The one with the `left_comm`):
-
-    ```
-    Hello world!
-    ```
-
-And on the left side:
-
-    ```
-    1 + 2 = 3
-    ```
+See spyder_kernels.comms for description.
 """
 
 from spyder_kernels.comms.commbase import CommError
--- a/spyder/plugins/ipythonconsole/comms/kernelcomm.py
+++ b/spyder/plugins/ipythonconsole/comms/kernelcomm.py
@@ -1,0 +1,105 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+
+"""
+In addition to the remote_call mechanism implemented in CommBase:
+ - Send a message to a debugging kernel
+"""
+import logging
+
+from qtpy.QtCore import QEventLoop, QObject, QTimer, Signal
+
+from spyder_kernels.comms.commbase import CommBase
+from spyder_kernels.py3compat import TimeoutError
+
+logger = logging.getLogger(__name__)
+
+
+class KernelComm(CommBase, QObject):
+    """
+    Class with the necessary attributes and methods to handle
+    communications with a console.
+    """
+
+    sig_got_reply = Signal(str)
+
+    def __init__(self):
+        super(KernelComm, self).__init__()
+
+        self.kernel_client = None
+        self._debugging = False
+        self._wait_list = {}
+
+        self.register_call_handler(
+            'set_debug_state', self._handle_debug_state)
+
+    def set_kernel_client(self, kernel_client):
+        """Register new kernel client and open comm."""
+        self._register_comm(
+            kernel_client.comm_manager.new_comm(self._comm_name))
+        self.kernel_client = kernel_client
+
+    def remote_call(self, interrupt=False, blocking=False):
+        """Get a handler for remote calls."""
+        return super(KernelComm, self).remote_call(
+            interrupt=interrupt, blocking=blocking)
+
+    # ---- Private -----
+    def _get_call_return_value(self, call_dict):
+        """
+        Interupt the kernel if needed.
+        """
+        settings = call_dict['settings']
+        if 'interrupt' in settings and settings['interrupt']:
+            self._signal_update_kernel()
+
+        return super(KernelComm, self)._get_call_return_value(call_dict)
+
+    def _signal_update_kernel(self):
+        """Interput the kernel to give a chance to read the messages."""
+        self._pdb_update()
+
+    def _pdb_update(self):
+        """
+        Update by sending an input to pdb.
+        """
+        if self._debugging and self.kernel_client:
+            cmd = (u"!get_ipython().kernel.frontend_comm" +
+                   ".remote_call(blocking=True).pong()")
+            self.kernel_client.input(cmd)
+
+    def _handle_debug_state(self, is_debugging):
+        """Update the debug state."""
+        self._debugging = is_debugging
+
+    def _wait_reply(self, call_id, call_name, timeout):
+        """Wait for the other side reply."""
+        if call_id in self._call_reply_dict:
+            return
+        self._wait_list[call_id] = call_name
+
+        # Create event loop to wait with
+        wait_loop = QEventLoop()
+        self.sig_got_reply.connect(wait_loop.quit)
+        wait_timeout = QTimer()
+        wait_timeout.setSingleShot(True)
+        wait_timeout.timeout.connect(wait_loop.quit)
+
+        # Wait until the kernel returns the value
+        wait_timeout.start(timeout * 1000)
+        while len(self._wait_list) > 0:
+            if not wait_timeout.isActive():
+                raise TimeoutError(
+                    "Timeout while waiting for {}".format(
+                        self._wait_list))
+            wait_loop.exec_()
+
+        wait_timeout.stop()
+
+    def _reply_recieved(self, call_id):
+        """A call got a reply."""
+        self._wait_list.pop(call_id, None)
+        self.sig_got_reply.emit(call_id)
--- a/spyder/plugins/ipythonconsole/comms/kernelcomm.py
+++ b/spyder/plugins/ipythonconsole/comms/kernelcomm.py
@@ -106,8 +106,8 @@ class KernelComm(CommBase, QObject):
         wait_timeout.stop()
         self.sig_got_reply.disconnect(wait_loop.quit)
 
-    def _reply_recieved(self, call_id):
+    def _reply_received(self, call_id):
         """A call got a reply."""
-        super(KernelComm, self)._reply_recieved(call_id)
+        super(KernelComm, self)._reply_received(call_id)
         self._wait_list.pop(call_id, None)
         self.sig_got_reply.emit(call_id)
--- a/spyder/plugins/ipythonconsole/comms/kernelcomm.py
+++ b/spyder/plugins/ipythonconsole/comms/kernelcomm.py
@@ -83,7 +83,7 @@ class KernelComm(CommBase, QObject):
 
     def _wait_reply(self, call_id, call_name, timeout):
         """Wait for the other side reply."""
-        if call_id in self._call_reply_dict:
+        if call_id in self._reply_inbox:
             return
 
         # Create event loop to wait with
--- a/spyder/plugins/ipythonconsole/comms/kernelcomm.py
+++ b/spyder/plugins/ipythonconsole/comms/kernelcomm.py
@@ -13,7 +13,7 @@ import logging
 from qtpy.QtCore import QEventLoop, QObject, QTimer, Signal
 
 from spyder_kernels.comms.commbase import CommBase
-from spyder_kernels.py3compat import TimeoutError
+from spyder.py3compat import TimeoutError
 
 logger = logging.getLogger(__name__)
 
--- a/spyder/plugins/ipythonconsole/comms/kernelcomm.py
+++ b/spyder/plugins/ipythonconsole/comms/kernelcomm.py
@@ -43,13 +43,15 @@ class KernelComm(CommBase, QObject):
             kernel_client.comm_manager.new_comm(self._comm_name))
         self.kernel_client = kernel_client
 
-    def remote_call(self, interrupt=False, blocking=False, comm_id=None):
+    def remote_call(self, interrupt=False, blocking=False, callback=None,
+                    comm_id=None):
         """Get a handler for remote calls."""
         return super(KernelComm, self).remote_call(
-            interrupt=interrupt, blocking=blocking, comm_id=comm_id)
+            interrupt=interrupt, blocking=blocking, callback=callback,
+            comm_id=comm_id)
 
     # ---- Private -----
-    def _get_call_return_value(self, call_dict):
+    def _get_call_return_value(self, call_dict, callback):
         """
         Interupt the kernel if needed.
         """
@@ -57,7 +59,8 @@ class KernelComm(CommBase, QObject):
         if 'interrupt' in settings and settings['interrupt']:
             self._signal_update_kernel()
 
-        return super(KernelComm, self)._get_call_return_value(call_dict)
+        return super(KernelComm, self)._get_call_return_value(
+            call_dict, callback)
 
     def _signal_update_kernel(self):
         """Interrupt the kernel to give a chance to read other messages."""
@@ -94,14 +97,17 @@ class KernelComm(CommBase, QObject):
         wait_timeout.start(timeout * 1000)
         while len(self._wait_list) > 0:
             if not wait_timeout.isActive():
+                self.sig_got_reply.disconnect(wait_loop.quit)
                 raise TimeoutError(
                     "Timeout while waiting for {}".format(
                         self._wait_list))
             wait_loop.exec_()
 
         wait_timeout.stop()
+        self.sig_got_reply.disconnect(wait_loop.quit)
 
     def _reply_recieved(self, call_id):
         """A call got a reply."""
+        super(KernelComm, self)._reply_recieved(call_id)
         self._wait_list.pop(call_id, None)
         self.sig_got_reply.emit(call_id)
--- a/spyder/plugins/ipythonconsole/comms/kernelcomm.py
+++ b/spyder/plugins/ipythonconsole/comms/kernelcomm.py
@@ -9,6 +9,7 @@ In addition to the remote_call mechanism implemented in CommBase:
  - Send a message to a debugging kernel
 """
 import logging
+import traceback
 
 from qtpy.QtCore import QEventLoop, QObject, QTimer, Signal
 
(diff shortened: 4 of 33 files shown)


ccordoba12:
I'm -1 on this. The right way to proceed is to add Comms to qtconsole and then use those to do things like this.

impact27:
While Comms could be a nicer way of doing the same thing, I wouldn't agree that this is more brittle than say using `silent_execute` to send messages to the console. `raw_input` is the standard way for a console to ask for an input from a frontend. 

The problem I see with Comms is, from https://ipython.org/ipython-doc/stable/development/messaging.html:
> These messages are fully symmetrical - both the Kernel and the Frontend can send each message, and no messages expect a reply.

This means that in the case of `runfile` for example, you can send a message saying that you are executing the file, but not wait for the file to be saved before executing. Of course you could reimplement a waiting system but you would do a lot of work to basically clone the `raw_input` behaviour.

About `raw_input`:
> This is a socket where the request/reply pattern goes in the opposite direction: from the kernel to a single frontend, and its purpose is to allow raw_input and similar operations that read from sys.stdin on the kernel to be fulfilled by the client. The request should be made to the frontend that made the execution request that prompted raw_input to be called. For now we will keep these messages as simple as possible, since they only mean to convey the raw_input(prompt) call.

`raw_input` has several advantages:
- Already implemented and well tested
- If several frontend are connected to the same kernel, only the frontend that made the execution request will be notified. 

In my opinion, reimplementing a custom messaging api would be time consuming. Using existing communications channels is probably more robust as this communication method is extensively tested. 

You might call this method a bit hacky, but I would argue that the problem I am trying to solve is to get an answer from the frontend, which is exactly what `raw_input` is designed to do. The `raw_input` message is well documented, and `msg['content']['prompt']` is the standard way of requesting informations.

bcolsen:
This is definitely a bit hacky, but I think that's fine as the previous solution to to runcells was a bit hacky as well. At least this is able to open doors to new capabilities in spyder 

I'm ok with this as long as the API is separate enough from raw_input implementation that it could be implemented using Comms at a later date, and that it doesn't interfere with the user interaction with console of course.

impact27:
First, I want to clarify my intent. I want to use the jupyter stdin channel, which is used for a kernel to request informations from a frontend. This is used by Python 3 in the `input()` function; by Python 2 in the `raw_input()` function; and in R by `readline()`. I will refer to these as `python::input()` in what follows, and use `ipykernel::raw_input()` to refer to the ipykernel implementation of the messaging protocol. This solution is therefore not specific to Python 2 `raw_input()` function. REF: https://jupyter-client.readthedocs.io/en/latest/messaging.html#messages-on-the-stdin-router-dealer-channel

As this messaging protocol is doing exactly what I need, I am not sure creating a new messaging protocol is needed, but if can be done using `Comms`. In `ipykernel`, this protocol is implemented in a method named `raw_input`. The name is from the Python 2 function. This function is just a convenient way of exchanging `input_request` and `input_reply` messages:

Message type: `input_request`:
```
content = {
    # the text to show at the prompt
    'prompt' : str,
    # Is the request for a password?
    # If so, the frontend shouldn't echo input.
    'password' : bool
}
```
Message type: `input_reply`:
```
content = { 'value' : str }
```






> This is definitely a bit hacky, but I think that's fine as the previous solution to to runcells was a bit hacky as well. At least this is able to open doors to new capabilities in spyder

The previous solution was really hack and brittle, as it would only work if the function was called in just the right way. With this implementation, the methods can be called from anywhere. I can for example run a script like:
```python
print(0)
runcell('mycell')
# %% mycell
print(1)
```
Where running the first cell would print:
```
0
1
```

> I'm ok with this as long as the API is separate enough from raw_input implementation that it could be implemented using Comms at a later date, and that it doesn't interfere with the user interaction with console of course.

The communication method and the api syntax are separated, making it easy to reimplement using a different method. I am not sure I see what in my code is dependent on `ipykernel::raw_input()` implementation, as I am just trying to use the standard messaging protocol. The only interference is that users can't use the `python::input()` function with a prompt that starts with `%spyder_api `.

The interactions with `raw_input` are concentrated in two functions:
 - in `spyder-kernels`, the requests are made in `def _spyder_api_request(request)`.
 - in `spyder`,  the requests are intercepted by `def _handle_input_request(self, msg)`. 

The assumptions I make are the following:

In `spyder-kernels`:

 - I can send a request to the frontend using `ipykernel::raw_input(prompt)`. This is a blocking call. I can send a string in `prompt`, and `ipykernel::raw_input` will return a string. This is unlikely to change as this is the standard way of handling `input()`.
 - To differentiate the api requests from a regular `python::input()` call, I add a `%spyder_api ` prefix to the prompt. This is the most 'hacky' part. Do you think that it could be problematic? Would any user use `%spyder_api` as a prefix in their calls to `input()`? 

In `spyder`:

 - `RichJupyterWidget` will receive `ipykernel::raw_input` requests with the method `_handle_input_request(self, msg)`. When this method is called, the frontend is expected to reply by passing a string to `self.kernel_client.input(code)`. This is already an assumption spyder makes in  https://github.com/spyder-ide/spyder/blob/5affb8dd30b7b6725e76bcb951ec172f3ac1acc8/spyder/plugins/ipythonconsole/widgets/debugging.py#L77. 

 - The prompt can be accessed with `msg['content']['prompt']`.

impact27:
In short, it should be easy to change the messaging method, and it is not interfering except if the user uses `input('%spyder_api' + a_string)` in the code or in the console.

I just think that the `input_request` and `input_reply` messaging protocol already solve the problem of requesting something from the kernel, so I wonder if it is necessary to reimplement a custom messaging protocol for that.

impact27:
I have added Comms to qtconsole (https://github.com/jupyter/qtconsole/pull/338). 
When this is merged, this should work.

spyder/plugins/ipythonconsole/widgets/spyderapi.py
```diff
@@ -0,0 +1,193 @@
...
+from qtconsole.rich_jupyter_widget import RichJupyterWidget
+
+from spyder.py3compat import PY2, to_text_string
+
+logger = logging.getLogger(__name__)
+
+# To be able to get and set variables between Python 2 and 3
+PICKLE_PROTOCOL = 2
+
+class ShellApiWidget(RichJupyterWidget):
```
ccordoba12: What if a third-party plugin wants to access this communication layer? In that case I think this is the wrong abstraction because it shouldn't be a widget.

Instead it could be a normal class or a mixin that can be inherited by the `SpyderPlugin` class we're defining in PR #9226 to expose (at least) `send_kernel_msg`, `handle_kernel_message`, `handle_kernel_request` and `kernel_reply`.

What do you think?

impact27: The only thing I need in this class that are external are: `kernel_client.comm_manager` and `kernel_manager.signal_kernel`, so changing that should be easy.

`send_kernel_msg` and `kernel_reply` can be exposed, but `handle_kernel_message` and `handle_kernel_request` need to be reimplemented by `NamepaceBrowserWidget` and `ShellWidget`, which are nor plugins. What would be the better way of handling these?

impact27:
It is now a stand alone class. You start the comm by calling `connect(kernel_client, kernel_manager)`, and you just need to register the methods you want for the different handlers.

impact27:
re-running tests

impact27:
What is the proper way of running the tests with a given PR of spyder-kernels?

impact27:
I have been trying it up for the last few weeks and it works quite well. I can't get the test to run properly as I don't know how to test with a spyder-kernel PR.
I get the error:
`/Users/vsts/agent/2.154.1/work/_temp/58375231-9d81-4a8b-80f1-964798b1587e.sh: line 1: ./continuous_integration/azure/install.sh: Permission denied`
Otherwise I would be glad to have a review.

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -32,6 +32,7 @@
         ControlWidget, DebuggingWidget, FigureBrowserWidget,
         HelpWidget, NamepaceBrowserWidget, PageControlWidget)
 
+from spyder_kernels.comms.kernelcomm import KernelComm
```
ccordoba12: Please move this file to `spyder/plugins/ipythonconsole/utils/kernelcomm.py`

spyder/plugins/ipythonconsole/comms/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+running. The message will only be delivered when the kernel is checking the
+event loop, or if pdb is waiting for an input.
+
+Example:
+
+On one side:
+
+    ```
+    def hello_str(msg):
+        print('Hello ' + msg + '!')
```
ccordoba12: Please change this to

      print('Hello ' + msg + '!')  # spyder: test-skip

to avoid an error with the test that verifies that no new print's are introduced in a PR.

spyder/plugins/ipythonconsole/comms/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+    left_comm.register_call_handler('add_numbers', add)
+    left_comm.register_call_handler('print_hello', hello_str)
+    ```
+
+On the other:
+
+    ```
+    right_comm.remote_call().print_hello('world')
+    res =  right_comm.remote_call(blocking=True).add_numbers(1, 2)
+    print('1 + 2 = ' + str(res))
```
ccordoba12: Please also add `  # spyder: test-skip` here.

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -910,7 +909,7 @@ def register_client(self, client, give_focus=True):
             shellwidget.set_cwd(cwd_path)
             if give_focus:
                 # Syncronice cwd with explorer and cwd widget
-                shellwidget.get_cwd()
+                shellwidget.update_cwd()
```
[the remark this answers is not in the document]
impact27: on the kernel side, `get_cwd` just returns the cwd, while `update_cwd` sends the cwd to the frontend, who then uses it to change the state. So `update_cwd` is more than a getter as it has side effects.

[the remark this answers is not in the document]
ccordoba12: Ok, fair enough.

impact27:
closing to retrigger the tests

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -28,27 +28,57 @@ class DebuggingWidget(RichJupyterWidget):
     Spyder
     """
 
+    def __init__(self, *args, **kwargs):
+        super(DebuggingWidget, self).__init__(*args, **kwargs)
+        self._previous_prompt = None
+        self._input_queue = []
+        self._input_ready = False
+
+    def monkeypatch_kernel_client(self, client):
```
ccordoba12: I don't like this name too much (monkey patching is not how things are done in Python, but sub-classing).

What this is essentially doing is replacing the kernel client's `input`, so what if we avoid the inner function definition (`monkeypatched_input`), call it `set_queued_input` and use it like this in `set_kernel_client_and_manager`

```python
self.kernel_client = kernel_client
if self.kernel_client is not None:
    self.set_queued_input(self.kernel_client)
```

?

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -28,27 +28,57 @@ class DebuggingWidget(RichJupyterWidget):
...
+
+    def register_message_handler(self, spyder_kernel_comm):
+        """Register messages to be handled by comm."""
+        spyder_kernel_comm.register_call_handler(
+            'get_breakpoints', self.get_spyder_breakpoints)
+        spyder_kernel_comm.sig_debugging.connect(self._debugging_hook)
+
+    def _debugging_hook(self, debugging):
+        """Catches the debugging state."""
+        # If the debugging starts or stops, clear the input queue.
```
ccordoba12: If ~~the~~ debugging ...

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -82,6 +110,21 @@ def _handle_input_request(self, msg):
...
+        self._input_ready = True
+        if not self.is_debugging():
+            return super(DebuggingWidget, self)._handle_input_request(msg)
+
+        # While the widget thinks only one input is going on,
+        # Other functions are sending messages to the kernel.
+        # This must be properly processed to avoid dropping messages.
+        # If the kernel was not ready, the messages are queued.
+        # Send the next one
+
```
ccordoba12: Please remove this blank line.

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -82,6 +110,21 @@ def _handle_input_request(self, msg):
         # Make sure that all output from the SUB channel has been processed
         # before entering readline mode.
         self.kernel_client.iopub_channel.flush()
+        self._input_ready = True
+        if not self.is_debugging():
+            return super(DebuggingWidget, self)._handle_input_request(msg)
+
+        # While the widget thinks only one input is going on,
+        # Other functions are sending messages to the kernel.
```
ccordoba12: Other -> other

ccordoba12: Also is right to says

    other functions are sending messages to the kernel

or would it be better to say

    other functions can be sending messages to the kernel

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -82,6 +110,21 @@ def _handle_input_request(self, msg):
...
+        # This must be properly processed to avoid dropping messages.
+        # If the kernel was not ready, the messages are queued.
+        # Send the next one
```
ccordoba12: This last line doesn't make sense to me. What do you mean by

    Send to next one

?

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -92,6 +135,9 @@ def callback(line):
                 if "do_" + cmd not in dir(pdb.Pdb):
                     self._control.history.append(line)
 
+            # must match ConsoleWidget.do_execute
+            self._executing = True
```
ccordoba12: This is already an attribute present in qtconsole:

https://github.com/jupyter/qtconsole/blob/f9ee7285f5ea6bd4a4aa411dbaae279dac43ac5d/qtconsole/console_widget.py#L266

so I think you don't need to declare it here again.

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -28,27 +28,57 @@ class DebuggingWidget(RichJupyterWidget):
...
+                if self._input_ready:
+                    self._input_ready = False
+                    return old_input(string)
+                elif self.is_debugging():
+                    self._input_queue.append(string)
+
+            client.input = monkeypatched_input
+        return client
+
+    def register_message_handler(self, spyder_kernel_comm):
```
ccordoba12: It's very confusing to have several `register_message_handler` methods in the different widgets that receive `spyder_kernel_comm` as an argument, when a) `spyder_kernel_comm` is an attribute of `ShellWidget` and b) these widgets are merely a code trick to not have a gigantic `ShellWidget`.

Therefore, I think we should unify all these `register_message_handler` methods in `ShellWidget` and call it only there (on it's `__init__` method, as you're doing it already).

ccordoba12: However, the handler definitions can keep living in their respective widgets (e.g. `get_spyder_breakpoints` here and `set_pdb_state` in `NamepaceBrowserWidget`).

spyder/plugins/variableexplorer/widgets/collectionseditor.py
```diff
@@ -1435,13 +1434,9 @@ def get_value(self, name):
...
         """Create new value in data"""
         try:
-            # We need to enclose values in a list to be able to send
-            # them to the kernel in Python 2
-            svalue = [cloudpickle.dumps(value, protocol=PICKLE_PROTOCOL)]
-
             # Needed to prevent memory leaks. See spyder-ide/spyder#7158.
-            if len(svalue) < MAX_SERIALIZED_LENGHT:
-                self.shellwidget.set_value(name, svalue)
+            if sys.getsizeof(value) < MAX_SERIALIZED_LENGHT:
```
ccordoba12: Are you sure that `getsizeof` is equivalent to the size of the serialized object? It's very important to have this right to avoid the serious memory leaks we had in the past.

spyder/plugins/ipythonconsole/widgets/debugging.py
impact27: I'm not declaring it, I'm setting it to True to have correct printing behaviour. Otherwise the messages printed during execution might not be displayed.

impact27: Which breaks some tests sometimes.

spyder/plugins/variableexplorer/widgets/collectionseditor.py
impact27: So what are these memory leaks caused by? Is `MAX_SERIALIZED_LENGHT` an estimation of the maximum size or a hard limit?
`sys.getsizeof(value)` is probably similar to the size of the pickled object.
If anything that transit should be below this limit, I should add that to the comms.

impact27:
```python
a = [0] * 10000
len(cloudpickle.dumps(a)) = 20034
sys.getsizeof(a) = 80064
a = np.zeros(10000)
len(cloudpickle.dumps(a)) = 80162
sys.getsizeof(a) = 80096
```

spyder/plugins/ipythonconsole/widgets/debugging.py
ccordoba12: Ok, thanks for the clarification.

spyder/plugins/ipythonconsole/widgets/debugging.py
ccordoba12: You miss this comment.

spyder/plugins/variableexplorer/widgets/collectionseditor.py
ccordoba12: On a second thought, this probably doesn't apply to comms but only to our previous architecture. We needed this check because we were executing commands in the kernel with strings, and we noticed that trying to do that for large strings (in this case, large serialized values), was freezing Spyder.

However, I think this should stay on this side and not on the kernel one because we should still prevent users trying to change values on big objects (e.g. big dataframes). Setting values consumes a lot of memory because values have to be serialized here and deserialized in the kernel.

So, could you revert your last commit here and also the one you added in spyder-ide/spyder-kernels#111? Thanks!

ccordoba12: And giving your comparisons, I think using `getsizeof` is fine.

ccordoba12:
I fixed the problems in Python 2 (we were using an old version of wurlitzer in our tests).

spyder/plugins/ipythonconsole/comms/kernelcomm.py
```diff
@@ -0,0 +1,107 @@
...
+"""
+In addition to the remote_call mechanism implemented in CommBase:
+ - Send a message to a debugging kernel
+"""
+import logging
+
+from qtpy.QtCore import QEventLoop, QObject, QTimer, Signal
+
+from spyder_kernels.comms.commbase import CommBase
+from spyder_kernels.py3compat import TimeoutError
```
ccordoba12: Please move this definition of TimeoutError to the `py3compat` module in Spyder. We don't mix py3compat modules of our packages.

spyder/plugins/ipythonconsole/comms/kernelcomm.py
```diff
@@ -0,0 +1,107 @@
...
+        wait_timeout.setSingleShot(True)
+        wait_timeout.timeout.connect(wait_loop.quit)
+
+        # Wait until the kernel returns the value
+        wait_timeout.start(timeout * 1000)
+        while len(self._wait_list) > 0:
+            if not wait_timeout.isActive():
+                raise TimeoutError(
+                    "Timeout while waiting for {}".format(
+                        self._wait_list))
```
ccordoba12: I think this error should be shown in a QMessageBox instead of raising it. That's because a raise would cause our error report dialog to be displayed, which would make users to post this to Github. And (it seems to me) this is informative and not a cause of concern.

spyder/plugins/variableexplorer/widgets/collectionseditor.py
impact27: Well actually I realized:
``` python
a = [[0] * 10000]
len(cloudpickle.dumps(a)) = 20034
sys.getsizeof(a) = 72
```
But I can also do a recursive check.

spyder/plugins/ipythonconsole/comms/kernelcomm.py
impact27: Maybe it would be better to handle that where the blocking call is made? Or alternaltively to disable blocking calls from the frontend and only use callbacks?

impact27: Actually it makes a lot of sense to add to option to have a callback on the remote calls. The blocking option should only be used when we need to block the thread.

spyder/plugins/variableexplorer/widgets/collectionseditor.py
impact27: And am I missing something or was `len(svalue)` always 1? svalue is an array with one element, that is the buffer. So was this condition ever triggered?

impact27: Can I add a new dependency? https://pympler.readthedocs.io/en/latest/asizeof.html#asizeof

impact27:
```python
from pympler.asizeof import asizeof
import sys
import cloudpickle
a = [[0] * 1000]
asizeof(a)
8160
sys.getsizeof(a)
72
len(cloudpickle.dumps(a))
2019
```

impact27:
```python
a = [[0] * 1000000]
%timeit len(cloudpickle.dumps(a))
1.65 s
%timeit asizeof(a)
728 ms
```

ccordoba12: > And am I missing something or was len(svalue) always 1?

Right, but that was a mistake that went unnoticed. We should have checked for `svalue[0]` instead.

> Can I add a new dependency?

Yes, but you need to add it to conda-forge first. [Here](https://github.com/conda-forge/staged-recipes) you can find the instructions on how to do that.

impact27: Alternatively I can just use `len(cloudpickle.dumps(a))`

spyder/plugins/ipythonconsole/comms/kernelcomm.py
ccordoba12: > Actually it makes a lot of sense to add to option to have a callback on the remote calls

Ok, but that still doesn't solve my comment here. Please show here a QMessageBox instead of raising an error, so users don't post this error on Github.

spyder/plugins/variableexplorer/widgets/collectionseditor.py
ccordoba12: I don't like that because it we would be serializing `value` twice (first to check its size and then to send it to the kernel). Besides, that check is very slow compared to `asizeof` (as you showed above).

Before it was fine for us to do that because we had to serialize `value` at this point before sending it to the kernel. But now it doesn't make sense to me.

spyder/plugins/ipythonconsole/comms/kernelcomm.py
impact27: So the problem here is that the calling function expects a reply, so the QMessageBox should be done there.

impact27: If I just show a QMessageBox at this point and return None, any function that calls the kernel in a blocking manner might get a None instead of the expected value? Or we expect any functions that makes a blocking call to handle a timeout?

impact27: Is it ok for the following functions to return None in case of timeout?
```
is_defined
get_doc
get_source
get_value
load_data
save_namespace
```

ccordoba12: > Is it ok for the following functions to return None in case of timeout?

It was like that before (you can check that in the code you removed in this PR). And we handled things that way because most of the methods you mentioned above were not supported while the kernel was on debugging mode (are they working while debugging now?)

The only exception was `get_value`. In that case, if we were not able to retrieve a value, we sent the traceback generated in the kernel when trying to do that and showed it in another QMessageBox.

spyder/plugins/variableexplorer/widgets/collectionseditor.py
impact27: ok I'll figure out how to send pympler to  conda-forge

spyder/plugins/ipythonconsole/comms/kernelcomm.py
impact27: ok I'll add a try: to these functions to catch the timeout and return None

spyder/plugins/variableexplorer/widgets/collectionseditor.py
```diff
@@ -26,7 +26,7 @@
 import warnings
 
 # Third party imports
-import cloudpickle
+from pympler.asizeof import asizeof
```
ccordoba12: I thought the name of the package was `asizeof`, but it's really pympler! That package is already present in conda-forge, so I just added it to our Anaconda.org channel so we can use it in our tests.

However, you also need to add it here to our setup.py, `requirements/conda.txt` and our Readme.

spyder/plugins/ipythonconsole/widgets/namespacebrowser.py
```diff
@@ -50,160 +48,80 @@ class NamepaceBrowserWidget(RichJupyterWidget):
...
-                msg = self._kernel_reply[:]
-                self._kernel_reply = None
-                raise ValueError(msg)
-
-        return self._kernel_value
+        try:
+            return self.call_kernel(
+                interrupt=True, blocking=True).get_value(name)
+        except TimeoutError:
+            raise ValueError('Could not read value.')
```
ccordoba12: How do you plan to handle serialization errors returned by the kernel?

spyder/plugins/variableexplorer/widgets/collectionseditor.py
impact27: It is already there https://anaconda.org/conda-forge/pympler

ccordoba12:
Tests are failing now, most probably due to the move to callbacks (i.e. because things are non-bliocking now).

Please take a look at that.

impact27:
There was a problem with how I was getting the python version for cloudpickle, but htis is fixed now. Now there is the option to have either blocking or non-blocking calls.

spyder/plugins/ipythonconsole/widgets/namespacebrowser.py
impact27: I changed it to except Exception so anything is transformed to a `return None` / `raise ValueError`

impact27: Wait should I raise a `ValueError` or just display the error in a QMessageBox?

impact27:
Could someone have a look at the Windows fails? I don't have windows but I don't really see what is plateform specific in this code.

spyder/app/tests/test_mainwindow.py
```diff
@@ -1860,5 +1860,25 @@ def test_pylint_follows_file(qtbot, tmpdir, main_window):
         assert fname == pylint_plugin.get_filename()
 
 
+def test_report_comms_error(qtbot, main_window):
+    """Test if a comms error is correctly displayed."""
+    CONF.set('main', 'show_internal_errors', True)
+    shell = main_window.ipyconsole.get_current_shellwidget()
+    qtbot.waitUntil(lambda: not shell._executing)
```
ccordoba12: I think the problem on Windows is related to this line. Please use the same line we use in all other tests to wait for the console to be fully ready to execute code:

```python
qtbot.waitUntil(lambda: shell._prompt_html is not None, timeout=SHELL_TIMEOUT)
```

spyder/app/tests/test_mainwindow.py
```diff
@@ -1860,5 +1860,25 @@ def test_pylint_follows_file(qtbot, tmpdir, main_window):
...
+                      "register_call_handler('get_cwd', get_cwd)")
+    with qtbot.waitSignal(shell.executed):
+        shell.execute('ls')
+
+    error_dlg = main_window.console.error_dlg
+    assert error_dlg is not None
+    assert 'Exception in comms call get_cwd' in error_dlg.error_traceback
+    assert 'ImportError: No module named' in error_dlg.error_traceback
+
+
```
ccordoba12: Also add here

```python
CONF.set('main', 'show_internal_errors', False)
```

impact27:
Also the last problem I have with error handling is this:

 - If I get an asyncronous error, I report it in the error window.
 - If I have a syncronous error, I raise it to be handled by the calling function

But in the latter case, I can't attach the traceback to the error. This is because:
 - `traceback` objects are not really python objects and can't be pickled
 - Therefore I need to use `traceback.extract_tb` to get a pre-processed traceback that I can send.
 - On the frontend side, I need a traceback object, which I can't create (python < 3.7) and can't arrach to the exception (python 2)
 - So I just print the traceback (On python2 : with `print`, on python3: with `traceback.print_list`) and raise the exception.

Do you see a better way of handling the traceback for syncronous errors?

ccordoba12:
> If I get an asyncronous error, I report it in the error window.

This is now working as expected. Thanks a lot for that!

> Do you see a better way of handling the traceback for syncronous errors?

Yes, instead of trying to send the traceback as a live object (which is not feasible, as you said), you can simply transform it to a list of strings and send it serialized like that. That's how the Json RPC server that supports the Python language server handles sending errors to clients connected to it:

https://github.com/palantir/python-jsonrpc-server/blob/7d53ae167fbc2d53c91efec3c0445fe820bd0bf1/pyls_jsonrpc/exceptions.py#L77-L78

(Here 'message' contains only the exception) 

In our side we take this dictionary and process it as shown below:

https://github.com/spyder-ide/spyder/blob/42832bbed3355bbd52ec50fede6225cd034da099/spyder/plugins/completion/languageserver/client.py#L303-L310

Finally, we reroute this content to the our error dialog here:

https://github.com/spyder-ide/spyder/blob/42832bbed3355bbd52ec50fede6225cd034da099/spyder/plugins/completion/languageserver/plugin.py#L135-L138

ccordoba12:
Also, please skip `test_range_indicator_visible_on_hover_only` on Windows to see if tests pass there (I don't know why that test started to fail).

spyder/plugins/ipythonconsole/widgets/namespacebrowser.py
```diff
@@ -50,160 +50,80 @@ class NamepaceBrowserWidget(RichJupyterWidget):
...
-                msg = self._kernel_reply[:]
-                self._kernel_reply = None
-                raise ValueError(msg)
-
-        return self._kernel_value
+        try:
+            return self.call_kernel(
+                interrupt=True, blocking=True).get_value(name)
+        except (TimeoutError, UnpicklingError):
+            raise ValueError('Kernel did not send the value.')
```
ccordoba12: This is still problematic because a failure in `get_value` is showing both our error dialog and a QMessageBox that displays the above message. And I only want to show the later here because most of these errors are caused by the inability to pickle the corresponding value in the kernel.

So please make the errors corresponding to `get_value` to not be reported. If you could do that in a generic way (i.e. if we can decide what handlers are allowed to report errors), that'd be even better.

ccordoba12: After a bit of thought, I think I'm seeing this issue the wrong way by relying in the mental model I have for our previous architecture.

Now I think this is something that needs to be solved in CommBase by addressing failures in pickling the requested data. I'll left you a comment there instead.

impact27:
> )



> > If I get an asyncronous error, I report it in the error window.
> 
> This is now working as expected. Thanks a lot for that!
> 
> > Do you see a better way of handling the traceback for syncronous errors?
> 
> Yes, instead of trying to send the traceback as a live object (which is not feasible, as you said), you can simply transform it to a list of strings and send it serialized like that. That's how the Json RPC server that supports the Python language server handles sending errors to clients connected to it:
> 
> https://github.com/palantir/python-jsonrpc-server/blob/7d53ae167fbc2d53c91efec3c0445fe820bd0bf1/pyls_jsonrpc/exceptions.py#L77-L78
> 
> (Here 'message' contains only the exception)
> 
> In our side we take this dictionary and process it as shown below:
> 
> https://github.com/spyder-ide/spyder/blob/42832bbed3355bbd52ec50fede6225cd034da099/spyder/plugins/completion/languageserver/client.py#L303-L310
> 
> Finally, we reroute this content to the our error dialog here:
> 
> https://github.com/spyder-ide/spyder/blob/42832bbed3355bbd52ec50fede6225cd034da099/spyder/plugins/completion/languageserver/plugin.py#L135-L138

This is what I am doing for asyncronous errors but this might not be the best solution for syncronous error. The solution I have now is:
If we have an error, raise an error with the same error type containing:
 - the original error
 - the name and id of the comm
 - the traceback

Then, I replace `sys.excepthook` to correctly handle this _if the developper using the comm did not catch it_, so the error can be properly displayed.

ccordoba12:
> might not be the best solution for syncronous error

Why not?

> The solution I have now is

But that's for async errors, right?

impact27:
> But that's for async errors, right?

Yes, if I recieve an error fron an async call, I can just send it to the error display.

If I recieve an error for a sync call, there is a function on top of me that expects one of two things:

 - The call succeeds and a valid value is returned
 - The call fails and an exception is raised

Clearly I an in the second case, so I must raise an exception. And I am in luck, I have a meaningful exception that I can raise. But I have more information that I also want to give the exception (mainly the original traceback). So Instead of raising the error without the traceback, I can create a new error that contains both the original error and the traceback.

If the error is caught, this information can be read by the calling programmer, so no problem.

If the error ends up in `sys.excepthook`, It will be displayed in the error display. I just need to adjust the way it is printed.

ccordoba12:
> I just need to adjust the way it is printed.

Ok, I was the only suggestion I had to make.

impact27:
Ok I think I found a bug that I solved in commit f340a89affbecf7174a680c7631c3966faf8e6f6. 
(basically it caused traceback to be non completely transmitted to the error display). Should I create a new pull request for this?

ccordoba12:
No, please solve that here.

impact27:
retriger the tests for python2

impact27:
I now skip `test_range_indicator_visible_on_hover_only` and `test_tab_moves_focus_from_search_to_replace`. The error looks very strange:
```
finder.search_text.setFocus()
qtbot.wait(100)
>       assert finder.search_text.hasFocus()
E       assert False
```
Could someone check on another windows machine if this really fails?

impact27:
I revert the test skips because:
 - I tried on a windows computer and the failing test pass.
 - Other tests are failing on these machines when I skip these.

Is is possible to restart this machines maybe?

spyder/app/tests/test_mainwindow.py
```diff
@@ -1860,5 +1860,27 @@ def test_pylint_follows_file(qtbot, tmpdir, main_window):
...
+    with qtbot.waitSignal(shell.executed):
+        shell.execute("get_ipython().kernel.frontend_comm."
+                      "register_call_handler('get_cwd', get_cwd)")
+    with qtbot.waitSignal(shell.executed):
+        shell.execute('ls')
+
+    error_dlg = main_window.console.error_dlg
+    assert error_dlg is not None
+    assert 'Exception in comms call get_cwd' in error_dlg.error_traceback
+    assert 'No module named' in error_dlg.error_traceback
```
ccordoba12: The errors on Windows are probably caused because you're closing `error_dlg` after it's displayed by this test. So below this line please add another one that closes it or hides it and then the other tests should pass without problems.

impact27:
retriggering the tests

ccordoba12:
@impact27, please don't add more commits here. I'll merge your spyder-kernels PR, then make the corresponding changes to our CI scripts and finally merge this.

spyder/plugins/editor/widgets/tests/test_introspection.py
```diff
@@ -60,6 +60,7 @@ def test_space_completion(lsp_codeeditor, qtbot):
 
 @pytest.mark.slow
 @pytest.mark.first
+@flaky(max_runs=5)
```
goanpeca: We should simply disable this test on CI. It checks a very simple thing but creates a lot of problems on the CI an is too flaky...

ccordoba12: It should be passing now without giving so much problems.

impact27:
> the `NUMPY_REQVER` doesn't look right here - shouldn't that be `PYMPLER_REQVER`?
Well spotted.

Issue #9354, opened 2019-05-17
Trailing \ (backslash) in -w confuses Spyder


nerdfever:
## Problem Description

If you start Spyder from Windows with the `-w` switch (set working directory), the switch works only if the path that follows the `-w` _does not_ contain a trailing backslash.

This is especially unfortunate as `%~p1` in the Windows command processor produces paths with trailing backslashes.

### What steps reproduce the problem?

1.  Start Spyder with command line `spyder.exe "file" -w "path"`, where path ends in `\`
2.  Observe the working directory is not "path"
3.  Start Spyder with command line `spyder.exe "file" -w "path"`, omitting the final `\` from the path.
4. Observe the working directory is now "path"

Example - works:

`Spyder.exe "C:\Endoscope parts\LED_adapter_dl18.py" -w "\Endoscope parts"`

Example - fails:

`Spyder.exe "C:\Endoscope parts\LED_adapter_dl18.py" -w "\Endoscope parts\"`

### What is the expected output? What do you see instead?

I expect it should work the same regardless of the presence or absence of the trailing backslash.

* Spyder version: 3.3.1
* Python version: 3.7
* Operating System name/version: Windows 10 Pro x64

stonebig:
hi @nerdfever , try the suggested patch and tell us , via a beautifull animated gif ?

nerdfever:
It is very strange. I've been trying the patch. When I put it in, Spyder simply doesn't start. When I take it out, all is fine.

To debug, I tried this code (three commented-out lines):

```
    def __init__(self, options=None):
        QMainWindow.__init__(self)

        qapp = QApplication.instance()
        if PYQT5:
            # Enabling scaling for high dpi
            qapp.setAttribute(Qt.AA_UseHighDpiPixmaps)
        self.default_style = str(qapp.style().objectName())

        self.dialog_manager = DialogManager()
        
#        with open(r'C:\Users\Dave\Desktop\out.txt') as o:
#            o.print(options)
#            o.print(options.working_directory)

        self.init_workdir = options.working_directory
        self.profile = options.profile
```

When those 3 lines are uncommented, I get the Spyder splash screen for a fraction of a second, then nothing. On the 2nd attempt I get:

![image](https://user-images.githubusercontent.com/9833299/58029996-c30fb300-7aeb-11e9-8641-2729f609f93f.png)

Commenting-out those 3 lines makes everything work normally again.

I'm mystified.

nerdfever:
[...]

PR #9360, opened 2019-05-18
help spyder registery integration on windows

stonebig:
see https://github.com/spyder-ide/spyder/issues/9354

## Description of Changes

removes the un-needed trailing r"\" present in a windows path returned per "~ps1", when calling spyder via a registery shortcut


stonebig:
hum, it' doesn't work, spyder can't succeed to start if the directory is recognized and  changed ... that's not the problem.

PR #9371, opened 2019-05-19
PR: Fix test_get_git_refs when run by Travis on tags

ccordoba12:
I found this problem after releasing beta2 (it was not possible to notice it before).


goanpeca:
And thanks for fixing the typo :-p

PR #9384, opened 2019-05-21
PR: Add fuzzy search to the Variable Explorer

### LINKED ISSUE #4067: Add fuzzy search functionality to the Variable Explorer
It is hard to explain so let me present a scenario:
You are researching and have many dataframes with various names in your current session. you want to look at some dataframe you just created. printing it is rarely helpful.
Right now, I search for my dataframe in the variable explorer and open it there to have a nicer view of it.
But if I created many variables searching for the one I want becomes tedious.
What I want is the ability to have faster access to the variable by providing the variable's name.
maybe writing:
view(mydataframe)

Or having a shortcut to search all variables in variable explorer (hopefully in a fuzzy manner) and hitting Enter to view it.

Any suggestions to the tired researcher?

dalthviz:
## Description of Changes

fuzzy

### Issue(s) Resolved

Fixes #4067 

### Affirmation


spyder/plugins/variableexplorer/widgets/collectionseditor.py
```diff
@@ -139,7 +144,12 @@ def __init__(self, parent, data, title="", names=False,
         self.sizes = []
         self.types = []
         self.set_data(data)
-        
+
+    def current_index(self):
+        """Get the currently selected index in the parent table view."""
+        i = self._parent.proxy_model.mapToSource(self._parent.currentIndex())
```
goanpeca:
```suggestion
        idx = self._parent.proxy_model.mapToSource(self._parent.currentIndex())
```

spyder/plugins/variableexplorer/widgets/collectionseditor.py
```diff
@@ -139,7 +144,12 @@ def __init__(self, parent, data, title="", names=False,
         self.sizes = []
         self.types = []
         self.set_data(data)
-        
+
+    def current_index(self):
+        """Get the currently selected index in the parent table view."""
+        i = self._parent.proxy_model.mapToSource(self._parent.currentIndex())
+        return i
```
goanpeca:
```suggestion
        return idx
```

spyder/plugins/variableexplorer/widgets/collectionseditor.py
```diff
@@ -253,7 +263,7 @@ def sort(self, column, order=Qt.AscendingOrder):
                 self.sizes.sort(reverse=reverse)
             except:
                 pass
-        elif column == 3:
+        elif column == 3 or column == 4:
```
goanpeca:
```suggestion
        elif column in [3, 4]:
```

spyder/plugins/variableexplorer/widgets/collectionseditor.py
```diff
@@ -322,11 +332,30 @@ def get_bgcolor(self, index):
             color.setAlphaF(.3)
         return color
 
+    def update_search_letters(self, text):
+        """Update search letters with text input in search box."""
+        self.letters = text
+        names = self.keys
+        results = get_search_scores(text, names, template='<b>{0}</b>')
+        self.normal_text, self.rich_text, self.scores = zip(*results)
```
goanpeca: Are we using the `rich_text` ? (could not see that on the gif)

spyder/plugins/variableexplorer/widgets/collectionseditor.py
```diff
@@ -1547,7 +1577,17 @@ def __init__(self, parent, data, minmax=False, shellwidget=None,
...
+        self.proxy_model = CollectionsCustomSortFilterProxy(self)
+        self.last_regex = ''
+
+        self.proxy_model.setSourceModel(self.model)
+        self.proxy_model.setDynamicSortFilter(True)
+        self.proxy_model.setFilterKeyColumn(0)  # Col 0 for Name
+        self.proxy_model.setFilterCaseSensitivity(Qt.CaseInsensitive)
+        self.setModel(self.proxy_model)
+
+        self.hideColumn(4)  # COl 4 for Score
```
goanpeca:
```suggestion
        self.hideColumn(4)  # Column 4 for Score
```

spyder/plugins/variableexplorer/widgets/namespacebrowser.py
```diff
@@ -383,3 +396,20 @@ def save_data(self, filename=None):
                     '<br><br>Error message:<br>') + error_message
             QMessageBox.critical(self, _("Save data"), save_data_message)
         self.save_button.setEnabled(self.filename is not None)
+
+
+class NamespacesBrowserFinder(ShortcutFinder):
+    """Textbox for filtering listed variables in the table."""
+    # TODO: Refactor to use base class for shortcuts and namespacebrowser
```
goanpeca: 👍

spyder/plugins/variableexplorer/widgets/collectionseditor.py
```diff
@@ -24,6 +24,7 @@
 import gc
 import sys
 import warnings
+import re
```
goanpeca: Add sorted perhaps ?
```python
import gc
import re
import sys
import warnings
```

spyder/plugins/variableexplorer/widgets/collectionseditor.py
```diff
@@ -54,8 +55,10 @@
 from spyder.utils.misc import getcwd_or_home
 from spyder.utils.qthelpers import (add_actions, create_action,
                                     mimedata2url)
+from spyder.utils.stringmatching import get_search_scores
 from spyder.plugins.variableexplorer.widgets.importwizard import ImportWizard
 from spyder.plugins.variableexplorer.widgets.texteditor import TextEditor
+from spyder.preferences.shortcuts import CustomSortFilterProxy
```
goanpeca: Same here (as above)?

spyder/plugins/variableexplorer/widgets/collectionseditor.py
dalthviz: I will check :+1:

spyder/plugins/variableexplorer/widgets/collectionseditor.py
dalthviz: I think here the import order is fine, `plugins` takes precedence of `preferences`, right?

spyder/plugins/variableexplorer/widgets/collectionseditor.py
dalthviz: @goanpeca  a preview with using the `rich_text`:
imagen

spyder/plugins/variableexplorer/widgets/collectionseditor.py
goanpeca: Opps yes!

spyder/plugins/variableexplorer/widgets/collectionseditor.py
goanpeca: It looks nice, but leave the code somewhere while we can fix the HTMLDelegate height as we discussed on the meeting

spyder/widgets/helperwidgets.py
```diff
@@ -283,6 +288,84 @@ def paintEvent(self, event):
             self._refresh()
 
 
+# Valid finder chars. To be improved
+VALID_ACCENT_CHARS = "ÁÉÍOÚáéíúóàèìòùÀÈÌÒÙâêîôûÂÊÎÔÛäëïöüÄËÏÖÜñÑ"
+VALID_FINDER_CHARS = r"[A-Za-z\s{0}]".format(VALID_ACCENT_CHARS)
+
+
+class FinderLineEdit(QLineEdit):
```
goanpeca: Nice! :-)

ccordoba12:
@dalthviz, since tests are passing with the last commit, could you change its commit message?

Also, please let me know when this is ready to review it.

spyder/plugins/variableexplorer/widgets/namespacebrowser.py
```diff
@@ -176,8 +194,13 @@ def setup_toolbar(self):
                 self, text=_("Remove all variables"),
                 icon=ima.icon('editdelete'), triggered=self.reset_namespace)
 
+        fuzzy_search_button = create_toolbutton(
+            self, text=_("Fuzzy search"),
```
ccordoba12: Let's call this `Search` instead of `Fuzzy search`.

spyder/plugins/variableexplorer/widgets/namespacebrowser.py
```diff
@@ -176,8 +194,13 @@ def setup_toolbar(self):
                 self, text=_("Remove all variables"),
                 icon=ima.icon('editdelete'), triggered=self.reset_namespace)
 
+        fuzzy_search_button = create_toolbutton(
```
ccordoba12: Let's call this button `search_button`

spyder/widgets/helperwidgets.py
```diff
@@ -283,6 +288,84 @@ def paintEvent(self, event):
             self._refresh()
 
 
+# Valid finder chars. To be improved
+VALID_ACCENT_CHARS = "ÁÉÍOÚáéíúóàèìòùÀÈÌÒÙâêîôûÂÊÎÔÛäëïöüÄËÏÖÜñÑ"
+VALID_FINDER_CHARS = r"[A-Za-z\s{0}]".format(VALID_ACCENT_CHARS)
```
ccordoba12: Please move these constants to the beginning of this file.

Also, how do these constants relate to filtering variables in the Variable Explorer? I say it because in Python 3 any utf-8 character can be a variable, so users are not limited to only using non-ascii variables, as in Python 2.

spyder/plugins/variableexplorer/widgets/namespacebrowser.py
```diff
@@ -392,3 +415,20 @@ def save_data(self, filename=None):
                     '<br><br>Error message:<br>') + error_message
             QMessageBox.critical(self, _("Save data"), save_data_message)
         self.save_button.setEnabled(self.filename is not None)
+
+
+class NamespacesBrowserFinder(FinderLineEdit):
+    """Textbox for filtering listed variables in the table."""
+
+    def keyPressEvent(self, event):
```
ccordoba12: Please add here the possibility to hide this widget when users press the `Esc` key.

ccordoba12:
This also needs a merge with master to fix the merge conflict.

spyder/widgets/helperwidgets.py
dalthviz: This constants are used as a regex to check the input of the search field, so you can't write something like `_` but you can write something like `á` as a valid search text:

image

ccordoba12: Ok, that works correctly for the keyboard shortcuts filtering but it doesn't work here for Python 3.

So we need a VALID_VARIABLES constant here and use it to filter variables, but with different values for Python 2 and 3:

* For Python 2 it only needs to ascii characters, digits and the `_` character.
* For Python 3 it needs to contain *any* utf-8 character.

For reference, please see the section called *Variable Names* in

https://realpython.com/python-variables/

You can also see

https://stackoverflow.com/questions/36330860/pythonically-check-if-a-variable-name-is-valid

spyder/plugins/variableexplorer/widgets/namespacebrowser.py
```diff
@@ -39,10 +40,16 @@
...
 from spyder.plugins.variableexplorer.widgets.importwizard import ImportWizard
+from spyder.widgets.helperwidgets import FinderLineEdit
 
 
 SUPPORTED_TYPES = get_supported_types()
 
+if PY2:
+    VALID_VARIABLE_CHARS = r"^[^\W]\w*\Z"
+else:
+    VALID_VARIABLE_CHARS = r"^[^\W]\w*\Z"
```
ccordoba12: Why do these regexp's are exactly the same for Python 2 and 3? Shouldn't they be different?

dalthviz: Yep, I just pushed the commit to check the testing. I haven't checked the regex for Python 3

ccordoba12: Ok, thanks for the clarification @dalthviz!

spyder/plugins/variableexplorer/widgets/namespacebrowser.py
```diff
@@ -46,9 +46,9 @@
 SUPPORTED_TYPES = get_supported_types()
 
 if PY2:
-    VALID_VARIABLE_CHARS = r"^[^\W]\w*\Z"
+    VALID_VARIABLE_CHARS = r"[^\W]\w*\Z"
 else:
-    VALID_VARIABLE_CHARS = r"^[^\W]\w*\Z"
+    VALID_VARIABLE_CHARS = r"[^\w+*=¡!¿?'\"#$%&()/<>\-\[\]{}^`´;,|¬]*\w"
```
goanpeca: @dalthviz are you checking is a variable has a valid name?

If that is the case, something like https://stackoverflow.com/a/29586366 might be useful?

dalthviz: I was thinking the validation per character, so in Python 2 something like numbers, underscore and letters (ASCII lower or uppercase) should be the only valid ones. In the case of Python 3, the set of valid characters increases to anything besides the ones for operations (and in general the characters in the regex), allowing to use the filter with something like this:

image

goanpeca: Understood :-)

ccordoba12: Yep, @dalthviz is right: we need to allow users to filter any kind of utf-8 character.

PR #9385, opened 2019-05-21
PR: Improve LSP preferences UI

ccordoba12:
After PR #9249, I considered necessary to improve the LSP server editor by using QGroupBoxes to group its options. I also improved its main label text to make it more straightforward.

Besides, I improved a bit the "Advanced" options tab.

This is part of issue #7724.

----

*Before*:

Selección_020

Selección_019

----

After

Selección_022

Selección_018


spyder/preferences/languageserver.py
```diff
@@ -120,33 +120,33 @@ def __init__(self, parent, language=None, cmd='', host='127.0.0.1',
...
+            "programming language, set the command to start its associated "
+            "server and any arguments that should be passed to it on "
+            "startup. Additionally, you can set the server's hostname and "
+            "port, if using a tcp connection to it or you want to "
+            "connect to an external server, or decide to use pipes for "
+            "communication instead."
+            "<br><br>"
+            "<i>Note</i>: You can use the placeholders <tt>{host}</tt> and "
+            "<tt>{port}</tt> in the server arguments field to automatically "
+            "fill in the respective values.<br>"
```
ccordoba12: @CAM-Gerlach, please help me to review this text.

ccordoba12:
@goanpeca, how can I decrease a little bit the width of the "Server configuration" layout?

spyder/preferences/languageserver.py
```diff
@@ -917,7 +929,7 @@ def setup_page(self):
         # --- Advanced tab ---
         # Advanced label
         advanced_label = QLabel(
-            _("Please don't modify these values unless "
+            _("<b>Warning</b>: Please don't modify these values unless "
```
CAM-Gerlach: Its usually better to tell people what *to* do, rather than what *not* to do.

```suggestion
            _("<b>Warning</b>: Only modify these values if "
```

spyder/preferences/languageserver.py
```diff
@@ -120,33 +120,33 @@ def __init__(self, parent, language=None, cmd='', host='127.0.0.1',
...
-            'programming language and the server hostname/IP and port to '
-            'use. If starting a local server, also provide the command to '
-            'execute it and any arguments that should be passed on startup.'
-            '<br><br>'
-            'You can use the placeholders <tt>{host}</tt> and '
-            '<tt>{port}</tt> in the server arguments field to '
-            'automatically fill in the respective values.<br>'
+            "To create a new server configuration, you need to select a "
+            "programming language, set the command to start its associated "
+            "server and any arguments that should be passed to it on "
```
CAM-Gerlach: Parallelism

```suggestion
            "server and enter any arguments that should be passed to it on "
```

spyder/preferences/languageserver.py
```diff
@@ -120,33 +120,33 @@ def __init__(self, parent, language=None, cmd='', host='127.0.0.1',
...
+            "startup. Additionally, you can set the server's hostname and "
+            "port, if using a tcp connection to it or you want to "
```
CAM-Gerlach: Correct caps and a little clearer.

```suggestion
            "port, if not connecting to a local server using pipes instead of TCP."
```

spyder/preferences/languageserver.py
```diff
@@ -120,33 +120,33 @@ def __init__(self, parent, language=None, cmd='', host='127.0.0.1',
...
+            "connect to an external server, or decide to use pipes for "
```
CAM-Gerlach: See above

```suggestion
```

spyder/preferences/languageserver.py
```diff
@@ -120,33 +120,33 @@ def __init__(self, parent, language=None, cmd='', host='127.0.0.1',
...
+            "communication instead."
```
CAM-Gerlach: See above

```suggestion
```

goanpeca:
@ccordoba12 here

```

        vlayout.addStretch()
        hlayout.addLayout(vlayout, 1)
        hlayout.addLayout(conf_layout, 3)
        general_vlayout.addLayout(hlayout)
```

The `1` and the `3` are the stretches, so the total layout is `4 = 1 + 3`. You could change those numbers to be `2` and `3` for example

goanpeca:
Thanks for this @ccordoba12 I really like how it looks for the server address and the advanced part, however for the other options (being a single option for each) looks a bit crowded

CAM-Gerlach:
Yeah, it does look a bit crowded. Perhaps put ``Command to execute`` and ``Server arguments`` into one overall box instead of two, since they're directly related?

spyder/preferences/languageserver.py
ccordoba12: I don't like this suggestion. I think it's better to be explicit here: you need to pass a host and port if you want to connect to an external server *or* use tcp instead of stdio.

CAM-Gerlach: Okay, then how about

```suggestion
            "port if connecting to an external server, "
            "or to a local one using TCP instead of stdio pipes."
```

ccordoba12: Much better, thanks!

ccordoba12:
> Perhaps put Command to execute and Server arguments into one overall box instead of two, since they're directly related?

Good suggestion. I grouped them under the "Server configuration" group (see the update screenshot above).

PR #9387, opened 2019-05-21
PR: Underline URIs in the Editor and enable opening them by Ctrl+click

goanpeca:
## Description of Changes

urlawareness

## What this PR does
This is including several enhancements

- [x] Add URL, mailto and FILE (paths without spaces for now!) awareness for any file
- [x] Add hover hint to these texts
- [x] Underlines any matching URI
- [x] Now goto definition and goto uri work for keyPress alone (not only mouse move), so that having the mouse on top of a word or uri and pressing cmd/ctrl should now change the highlight
- [x] For files that do not exist the color is red
- [x] Relative file paths
- [x] Paths with spaces

All `file://      /` must include a trailing `/` to work. 

### Issue(s) Resolved

Fixes #

### Affirmation


goanpeca:
@ccordoba12 I am not sure how we can handle spaces for the `file://` unless we have to assume that that will be the only entry on he line

goanpeca:
(travis tests passed, but the PR is pointing to an earlier build) 

https://travis-ci.org/spyder-ide/spyder/builds/535595769

jitseniesen:
> I am not sure how we can handle spaces for the `file://`

I thought spaces in URI are supposed to be percent encoded.

goanpeca:
Yes but we could be flexible 🤷🏻‍♂️?

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -230,6 +230,8 @@ class CodeEditor(TextEditBaseWidget):
     sig_bookmarks_changed = Signal()
     get_completions = Signal(bool)
     go_to_definition = Signal(str, int, int)
+    sig_go_to_uri = Signal(str)  # Used for testing
+    sig_uri_found = Signal(str)  # Used for testing
```
ccordoba12: Please add these signals below `sig_process_code_analysis` and explain more clearly when they are emitted and how they are used.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3108,6 +3147,54 @@ def run_pygments_highlighter(self):
         if isinstance(self.highlighter, sh.PygmentsSH):
             self.highlighter.make_charlist()
 
+    def _handle_goto_definition_event(self, pos):
+        """Check if goto definition can be applied and apply highlight."""
+        text = self.get_word_at(pos)
+        if (text and not sourcecode.is_keyword(to_text_string(text))):
```
ccordoba12: The parenthesis after the `if` is not needed anymore.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3108,6 +3147,54 @@ def run_pygments_highlighter(self):
         if isinstance(self.highlighter, sh.PygmentsSH):
             self.highlighter.make_charlist()
 
+    def _handle_goto_definition_event(self, pos):
+        """Check if goto definition can be applied and apply highlight."""
+        text = self.get_word_at(pos)
+        if (text and not sourcecode.is_keyword(to_text_string(text))):
+            if not self.__cursor_changed:
+                QApplication.setOverrideCursor(
+                                            QCursor(Qt.PointingHandCursor))
```
ccordoba12: Please move this to be only four spaces indented from QApplication above.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3108,6 +3147,54 @@ def run_pygments_highlighter(self):
...
+    def _handle_goto_uri_event(self, pos):
+        """Check if go to uri can be applied and apply highlight."""
+        uri, cursor = self.get_uri_at(pos)
+        if uri and cursor:
+            color = self.ctrl_click_color
+
+            if uri.startswith('file://'):
+                fname = uri.replace('file://', '')
+                if not osp.isfile(fname):
+                    color = QColor('red')
```
ccordoba12: This is a very strong color. What is you use instead `rgb(255, 80, 80)`? That's what we use in other places to signal failure.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3179,17 +3260,35 @@ def leaveEvent(self, event):
...
         alt = event.modifiers() & Qt.AltModifier
         pos = event.pos()
         if event.button() == Qt.LeftButton and ctrl:
             TextEditBaseWidget.mousePressEvent(self, event)
             cursor = self.cursorForPosition(pos)
-            self.go_to_definition_from_cursor(cursor)
+
+            if self._last_hover_uri:
+                uri = self._last_hover_uri
+                if uri.startswith('file://'):
```
ccordoba12: The logic to open files is already handled in the `open_file` method of `MainWindow`. I think we should use that instead.

spyder/plugins/editor/widgets/tests/test_goto.py
```diff
@@ -0,0 +1,66 @@
...
+            ('" https://google.com"\n', 'https://google.com'),  # String
+            ('# https://google.com"\n', 'https://google.com'),  # Comment
+            # Files that exist
+            (TEST_FILE_TEXT, TEMPFILE_PATH),  # File without spaces
+            # Mail to
+            ('" mailto:goanpeca@gmail.com"\n', 'mailto:goanpeca@gmail.com'),
+            ('# mailto:goanpeca@gmail.com\n', 'mailto:goanpeca@gmail.com'),
+        ]
+    )
+def test_goto_uri(qtbot, lsp_codeeditor, params):
```
ccordoba12: There's no need to use `lsp_codeeditor` because this doesn't depend on the LSP facilities at all. Since that's costly and prone to generate conflicts with other tests, we should only use it when it's really necessary.

So please use a fixture that creates a CodeEditor without LSP activated instead.

spyder/plugins/editor/widgets/tests/test_goto.py
```diff
@@ -0,0 +1,66 @@
...
+import pytest
+
+# Constants
+TEST_FOLDER = os.path.abspath(os.path.dirname(__file__))
+_, TEMPFILE_PATH = tempfile.mkstemp()
+TEST_FILE_TEXT = '"file://{}"\n'.format(TEMPFILE_PATH)
+
+
+@pytest.mark.slow
+@pytest.mark.second
```
ccordoba12: Please remove this markers because they are not necessary.

spyder/utils/syntaxhighlighters.py
```diff
@@ -688,7 +716,7 @@ def make_opencl_patterns():
...
...XP, CL_DBL_MIN_EXP, CL_DBL_RADIX, CL_DBL_MAX, CL_DBL_MIN, CL_DBL_EPSILON, CL_SUCCESS, CL_DEVICE_NOT_FOUND, CL_DEVICE_NOT_AVAILABLE, CL_COMPILER_NOT_AVAILABLE, CL_MEM_OBJECT_ALLOCATION_FAILURE, CL_OUT_OF_RESOURCES, CL_OUT_OF_HOST_MEMORY, CL_PROFILING_INFO_NOT_AVAILABLE, CL_MEM_COPY_OVERLAP, CL_IMAGE_FORMAT_MISMATCH, CL_IMAGE_FORMAT_NOT_SUPPORTED, CL_BUILD_PROGRAM_FAILURE, CL_MAP_FAILURE, CL_INVALID_VALUE, CL_INVALID_DEVICE_TYPE, CL_INVALID_PLATFORM, CL_INVALID_DEVICE, CL_INVALID_CONTEXT, CL_INVALID_QUEUE_PROPERTIES, CL_INVALID_COMMAND_QUEUE, CL_INVALID_HOST_PTR, CL_INVALID_MEM_OBJECT, CL_INVALID_IMAGE_FORMAT_DESCRIPTOR, CL_INVALID_IMAGE_SIZE, CL_INVALID_SAMPLER, CL_INVALID_BINARY, CL_INVALID_BUILD_OPTIONS, CL_INVALID_PROGRAM, CL_INVALID_PROGRAM_EXECUTABLE, CL_INVALID_KERNEL_NAME, CL_INVALID_KERNEL_DEFINITION, CL_INVALID_KERNEL, CL_INVALID_ARG_INDEX, CL_INVALID_ARG_VALUE, CL_INVALID_ARG_SIZE, CL_INVALID_KERNEL_ARGS, CL_INVALID_WORK_DIMENSION, CL_INVALID_WORK_GROUP_SIZE, CL_INVALID_WORK_ITEM_SIZE, CL_INVALID_GLOBAL_OFFSET, CL_INVALID_EVENT_WAIT_LIST, CL_INVALID_EVENT, CL_INVALID_OPERATION, CL_INVALID_GL_OBJECT, CL_INVALID_BUFFER_SIZE, CL_INVALID_MIP_LEVEL, CL_INVALID_GLOBAL_WORK_SIZE'
```
ccordoba12: What did you change in this string?

spyder/widgets/mixins.py
```diff
@@ -848,6 +848,47 @@ def get_word_at(self, coordinates):
 
         return word
 
+    def get_uri_at(self, coordinates):
```
ccordoba12: Why did you add these methods to the mixins? You have to be aware that these mixins are shared by the Internal console and the IPython one. And since you're not adding URI detection to those, I think these methods should be moved to a better place (perhaps to CodeEditor itself, or at least TextEditBaseWidget).

ccordoba12: By these methods I mean both `get_uri_at` and `get_pattern_cursor_at`.

spyder/plugins/editor/widgets/tests/test_goto.py
ccordoba12: There's a `code_editor_bot` or a similar fixture used in a lot of tests in `spyder/plugins/editor/widgets/tests`. Please move it to `spyder/plugins/editor/widgets/tests/fixtures.py` and use it for all tests that need it (including this one).

spyder/plugins/editor/widgets/codeeditor.py
goanpeca: I do not understand

spyder/utils/syntaxhighlighters.py
goanpeca: It was a typo. Fixed

goanpeca:
made the suggested fixed @ccordoba12. I do not understand the suggestion on `open_file`

Maybe the logic there should be moved to a util, but it feels weird to depend on the main window to do these operations since we are handling files and mailto and uris.

spyder/plugins/editor/widgets/tests/fixtures.py
```diff
@@ -19,11 +19,31 @@
...
 
 # Local imports
+from spyder import version_info
 from spyder.plugins.editor.widgets.codeeditor import CodeEditor
 from spyder.plugins.editor.widgets.editor import EditorStack
 from spyder.widgets.findreplace import FindReplace
 
 
+@pytest.fixture
+def code_editor_bot(qtbot):
```
ccordoba12: Tests are failing because there is another `code_editor_bot` fixture in `conftest.py` but with different content. Let me do a quick PR to try to improve the situation and then you can rebase in top of master.

goanpeca:
Thanks

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3179,17 +3318,36 @@ def leaveEvent(self, event):
...
+                    fname = self._preprocess_file_uri(uri)
+
+                    if osp.isfile(fname) and encoding.is_text_file(fname):
+                        # Open in editor
+                        self.go_to_definition.emit(fname, 0, 0)
+                    else:
+                        # Use external program
+                        quri = QUrl.fromLocalFile(fname)
+                        QDesktopServices.openUrl(quri)
```
ccordoba12: For local files I'd prefer to use the following (which is what `open_file` in `MainWindow` does too):

```python
fname = file_uri(fname)
programs.start_file(fname)
```

`file_uri` is present in `qthelpers` and `programs` refers to `utils/programs.py`

spyder/plugins/editor/widgets/codeeditor.py
ccordoba12: Yeah, this is too convoluted. I left a comment below.

goanpeca:
Done @ccordoba12

PR #9402, opened 2019-05-23
PR: Connect project explorer to file explorer when single click to open is updated

### LINKED ISSUE #9303: Single click to open files doesn't work in project explorer
### Issue Report Checklist

    * [x] Restarted Spyder

When enabling single click to open files I find it only works on file explorer but not in project explorer. Is this the default or the functionality should work similarly in both explorers?

1. 
2. 
3. 

```python-traceback

PASTE TRACEBACK HERE

```

## Versions

* Spyder version: 
* Python version: 
* Qt version: 
* PyQt version: 

### Dependencies

```
PASTE DEPENDENCIES HERE
```

goanpeca:
## Description of Changes

### Issue(s) Resolved

Fixes #9303

### Affirmation

### DIFF
--- a/spyder/plugins/projects/plugin.py
+++ b/spyder/plugins/projects/plugin.py
@@ -22,6 +22,7 @@ from qtpy.QtWidgets import QMenu, QMessageBox, QVBoxLayout
 
 # Local imports
 from spyder.config.base import _, get_home_dir
+from spyder.config.main import CONF
 from spyder.api.plugins import SpyderPluginWidget
 from spyder.py3compat import is_text_string, to_text_string
 from spyder.utils import encoding
@@ -47,11 +48,13 @@ class Projects(SpyderPluginWidget):
         SpyderPluginWidget.__init__(self, parent)
 
         self.explorer = ProjectExplorerWidget(
-                            self,
-                            name_filters=self.get_option('name_filters'),
-                            show_all=self.get_option('show_all'),
-                            show_hscrollbar=self.get_option('show_hscrollbar'),
-                            options_button=self.options_button)
+            self,
+            name_filters=self.get_option('name_filters'),
+            show_all=self.get_option('show_all'),
+            show_hscrollbar=self.get_option('show_hscrollbar'),
+            options_button=self.options_button,
+            single_click_to_open=CONF.get('explorer', 'single_click_to_open'),
+        )
 
         layout = QVBoxLayout()
         layout.addWidget(self.explorer)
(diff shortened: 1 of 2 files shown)


spyder/plugins/projects/plugin.py
```diff
@@ -47,11 +48,13 @@ def __init__(self, parent=None):
...
-                            name_filters=self.get_option('name_filters'),
-                            show_all=self.get_option('show_all'),
-                            show_hscrollbar=self.get_option('show_hscrollbar'),
-                            options_button=self.options_button)
+            self,
+            name_filters=self.get_option('name_filters'),
+            show_all=self.get_option('show_all'),
+            show_hscrollbar=self.get_option('show_hscrollbar'),
+            options_button=self.options_button,
+            single_click_to_open=CONF.get('explorer', 'single_click_to_open'),
```
ccordoba12: I think we should a separate option for this, instead of using the same of the File explorer. What do you think?

goanpeca: Well, I think the user that reported the issue was expecting the behavior on the file explorer to be the same on the project explorer, so it would make sense to use a single configuration value.

That is what I would expect as well. I think it would be weird to have single click on one vs double click on the other.

ccordoba12: Ok, fair enough. But in that case you need to make the File Explorer emit `sig_option_changed` when the `single_click_to_open` changes and apply the corresponding changes in the project explorer.

Else switching from single to double click wouldn't take effect immediately in the project explorer, but only until the next restart.

spyder/plugins/projects/plugin.py
```diff
@@ -165,6 +168,16 @@ def register_plugin(self):
...
+        # Connect to file explorer to keep single click to open files in sync
+        self.main.explorer.fileexplorer.sig_option_changed.connect(
+            self.set_single_click_to_open
+        )
+
+    def set_single_click_to_open(self, option, value):
+        """Set single click to open files and directories."""
+        if option == 'single_click_to_open':
+            self.explorer.treewidget.set_single_click_to_open(value)
+
```
goanpeca: >Ok, fair enough. But in that case you need to make the File Explorer emit sig_option_changed when the single_click_to_open changes and apply the corresponding changes in the project explorer.

>Else switching from single to double click wouldn't take effect immediately in the project explorer, but only until the next restart.

That is already handled here

spyder/plugins/projects/widgets/explorer.py
```diff
@@ -173,16 +173,20 @@ class ProjectExplorerWidget(QWidget):
     sig_open_file = Signal(str)
```
goanpeca: >sig_option_changed = Signal(str, object)	    sig_option_changed = Signal(str, object)

This is already being emitted

spyder/plugins/projects/plugin.py
ccordoba12: Ups, sorry, I didn't notice this.

Llaves:
Single-click open in the project explorer is not working in 5.1.5
I don't know if this is a reversion or if this is as-designed behavior.
Not sure if it's a separate issue, but the hidden files setting for the file explorer also does not work in the project explorer.

ccordoba12:
@Llaves, thanks for mentioning that. The thing is that now (i.e. in Spyder 5) the file and project explorers are disconnected from each other and have their own set of options.

In our 5.2.0 version (released a few days ago), we added an option to do single click for Projects and also hidden directories are shown by default on that pane (there's no option to hide them).

Llaves:
I"ll download and enjoy the new feature
Thx

PR #9404, opened 2019-05-23
PR: Add a code of conduct

ccordoba12:
In view of the recent and sometimes ugly discussions between @CAM-Gerlach and me, I decided it is time to adopt a code of conduct to try to avoid them in the future.

Both @CAM-Gerlach, me and all other members of our community will be accountable to this code from now on. So if I fall out of line with my comments according to this code, please let me or @goanpeca know about it.

I also want to apologize to @CAM-Gerlach and the rest of the @spyder-ide/core-developers and @spyder-ide/junior-developers for my behavior. I'll try to handle my disagreements better in the future.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,75 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to make participation in our project and
+our community a harassment-free experience for everyone, regardless of age, body
+size, disability, ethnicity, gender identity and expression, level of experience,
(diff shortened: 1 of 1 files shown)


CAM-Gerlach:
@ccordoba12 Thanks for doing this, I'd meant to do it a long time ago (it is a requirement of being a Numfocus affliated proect) but I never got to it. We should also reference this in our Readme and Contributing Guide, so its clear we have one. Otherwise, people have to go trawling through the repo root to find it. Do you want me to submit a PR for that or do you want to handle it?

ccordoba12:
> Do you want me to submit a PR for that or do you want to handle it?

Yes, please do.

CAM-Gerlach:
@ccordoba12 , @goanpeca , ping so that the above task can be assigned to someone still part of the project and doesn't get lost.

PR #9406, opened 2019-05-23
PR: Ignore FileNotFoundError when removing autosave file

impact27:
## Description of Changes

* [] Added unit test(s) covering the changes (if testable)

* [] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

If the autosave file was removed, a FileNotFoundError is raised. 
As the filed was going to be removed either way, we can ignore this error.

### Issue(s) Resolved

Fixes #

### Affirmation

### DIFF
--- a/spyder/plugins/editor/utils/autosave.py
+++ b/spyder/plugins/editor/utils/autosave.py
@@ -19,10 +19,12 @@ from spyder.config.base import _, get_conf_path
 from spyder.config.main import CONF
 from spyder.plugins.editor.widgets.autosaveerror import AutosaveErrorDialog
 from spyder.plugins.editor.widgets.recover import RecoveryDialog
-
+from spyder.py3compat import PY2
 
 logger = logging.getLogger(__name__)
 
+if PY2:
+    FileNotFoundError = OSError
 
 class AutosaveForPlugin(object):
     """Component of editor plugin implementing autosave functionality."""
@@ -163,6 +165,8 @@ class AutosaveForStack(object):
         autosave_filename = self.name_mapping[filename]
         try:
             os.remove(autosave_filename)
+        except FileNotFoundError:
+            pass
         except EnvironmentError as error:
             error_message = (_('{} while removing autosave file {}')
                              .format(type(error), autosave_filename))


jitseniesen:
Did you test this? I don't think it will work.

impact27:
```python-traceback
ERROR:spyder.plugins.editor.widgets.autosaveerror:<class 'FileNotFoundError'> while removing autosave file /Users/quentinpeter/.spyder-py3-dev/autosave/spydercommbase.py
  File "/Users/quentinpeter/pyscripts/spyder/spyder/plugins/editor/utils/autosave.py", line 165, in remove_autosave_file
    os.remove(autosave_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/Users/quentinpeter/.spyder-py3-dev/autosave/spydercommbase.py'
```
Why wouldn't this work?

ccordoba12:
I think `FileNotFoundError` is not present in Python 2, so please change that.

jitseniesen:
I thought the idea was to ignore any errors when the autosave file is not found (e.g. because it has been deleted outside Spyder). @impact27 Are you saying that when you test your change, you still get a file not found error?

impact27:
What is the correct error in python2?

impact27:
> I thought the idea was to ignore any errors when the autosave file is not found (e.g. because it has been deleted outside Spyder). @impact27 Are you saying that when you test your change, you still get a file not found error?

No, I just pasted the error I got to illustrarte.

jitseniesen:
> No, I just pasted the error I got to illustrate.

I see. So did you test whether your change works? I am asking partially because there is no unit test. Why did you not include a unit test?

goanpeca:
@impact27 https://stackoverflow.com/posts/21368457/revisions

goanpeca:
(And yes, we could add a simple test for this I guess?)

CAM-Gerlach:
Can you clearly articulate a motivation for removing this; i.e. when was the dialog popping up when you believe it shouldn't? If its when closing individual files that you've opened and closed in another copy of Spyder (thus removing the autosave) due to PR #9327 , then that error should be caught and passed in the file close function only.

This defeats the point of the error dialog entirely, which is to warn if the autosave file does not exist when it should (i.e. something is external deleting it, Spyder has a bug, etc), so they are aware that their autosave "safety net" may be in jeopardy (either due to it not working itself, or a problem with their disk) and report the error to us if it occurs consistently, so we can fix it.  Conversely, so probability of an error due to anything but the file not being found is virtually nil, so no one would likely ever see this dialog if so. If it annoys users (and it shouldn't, if everything is done on our end correctly) they can always click the ignore checkbox.

impact27:
> Can you clearly articulate a motivation for removing this; i.e. when was the dialog popping up when you believe it shouldn't? If its when closing individual files that you've opened and closed in another copy of Spyder (thus removing the autosave) due to PR #9327 , then that error should be caught and passed in the file close function only.
> 
> This defeats the point of the error dialog entirely, which is to warn if the autosave file does not exist when it should (i.e. something is external deleting it, Spyder has a bug, etc), so they are aware that their autosave "safety net" may be in jeopardy (either due to it not working itself, or a problem with their disk) and report the error to us if it occurs consistently, so we can fix it. Conversely, so probability of an error due to anything but the file not being found is virtually nil, so no one would likely ever see this dialog if so. If it annoys users (and it shouldn't, if everything is done on our end correctly) they can always click the ignore checkbox.

You are correct

Issue #9396, opened 2019-05-23
Unify code editor fixtures for editor plugin/widget/tests


goanpeca:

ccordoba12:
I'm working on part of this right now. It's kind of tricky, so let me push my PR first, then you can proceed with the unification of fixtures.

goanpeca:
I am not sure if you finished this or if this needs extra work?

ccordoba12:
This is on me. It's not critical, so we can leave it for later.

PR #9417, opened 2019-05-24
PR: Reorganize global fixtures in conftest files

ccordoba12:
This also removes some repeated fixtures.

TODO:

- [x] Move remaining global fixtures to module conftest's 
- [x] Fix autosave failing tests
- [x] Add some wait time in the teardown of `lsp_manager`.

This is part of issue #9396.


ccordoba12:
Ok, this is ready. A warning to @spyder-ide/core-developers and @spyder-ide/junior-developers: please don't add *any* fixture from now on to our main `conftest.py` (i.e. the one in the root of our repo). Instead, to reuse fixtures please add them to a `conftest.py` in the same directory where test files are present (that's what I did here too).

The problem with fixtures in the main `conftest.py` is that some plugin fixtures require to create a QApplication due to QtAwesome not being able to work without one. So adding them to the main `conftest.py` either introduces crashes for other plugin tests or prevents `pytest-xvfb` from working, which is really important for me on Linux.

PR #9422, opened 2019-05-24
PR: Enhance the display of warnings and errors 

### LINKED ISSUE #9247: Improve messages of code analysis tooltip
See:

Screen Shot 2019-04-30 at 16 39 41
Screen Shot 2019-04-30 at 16 39 51

goanpeca:
## Description of Changes

Screen Shot 2019-05-24 at 15 40 54

### Issue(s) Resolved

Fixes #9247

### Affirmation

### DIFF
--- a/spyder/utils/icon_manager.py
+++ b/spyder/utils/icon_manager.py
@@ -5,13 +5,15 @@
 # (see spyder/__init__.py for details)
 
 # Standard library imports
+import base64
 import os
 import os.path as osp
 import mimetypes as mime
 import sys
 
 # Third party imports
-from qtpy.QtGui import QIcon
+from qtpy.QtCore import QBuffer, QByteArray
+from qtpy.QtGui import QIcon, QImage
 from qtpy.QtWidgets import QStyle, QWidget
 
 # Local imports
@@ -161,6 +163,8 @@ _qtaargs = {
     'gotoline':                [('fa.sort-numeric-asc',), {'color': MAIN_FG_COLOR}],
     'error':                   [('fa.times-circle',), {'color': 'darkred'}],
     'warning':                 [('fa.warning',), {'color': 'orange'}],
+    'information':             [('fa.info-circle',), {'color': '#3775a9'}],
+    'hint':                    [('fa.lightbulb-o',), {'color': 'yellow'}],
     'todo':                    [('fa.exclamation',), {'color': '#3775a9'}],
     'ipython_console':         [('spyder.ipython-logo-alt',), {'color': MAIN_FG_COLOR}],
     'ipython_console_t':       [('spyder.ipython-logo-alt',), {'color':'gray'}],
@@ -464,3 +468,13 @@ def get_icon_by_extension(fname, scale_factor):
                         icon_by_extension = icon(
                             application_icons[bin_name], scale_factor)
     return icon_by_extension
+
+
+def base64_from_icon(icon_name, width, height):
+    """Convert icon to base64 encoding."""
+    icon_obj = icon(icon_name)
+    image = QImage(icon_obj.pixmap(width, height).toImage())
+    byte_array = QByteArray()
+    buffer = QBuffer(byte_array)
+    image.save(buffer, "PNG")
+    return byte_array.toBase64().data().decode()
--- a/spyder/plugins/editor/panels/linenumber.py
+++ b/spyder/plugins/editor/panels/linenumber.py
@@ -97,6 +97,9 @@ class LineNumberArea(Panel):
                                  Qt.AlignRight | Qt.AlignBottom,
                                  to_text_string(line_number))
 
+            icon_size = QSize(self._markers_margin_width - 1,
+                              self._markers_margin_width - 1)
+
             data = block.userData()
             if self._markers_margin and data:
                 if data.code_analysis:
@@ -109,9 +112,6 @@ class LineNumberArea(Panel):
                         warnings += sev == DiagnosticSeverity.WARNING
                         infos += sev == DiagnosticSeverity.INFORMATION
                         hints += sev == DiagnosticSeverity.HINT
-
-                    icon_size = QSize(self._markers_margin_width - 1,
-                                      self._markers_margin_width - 1)
 
                     if errors:
                         draw_pixmap(top, self.error_icon.pixmap(icon_size))
--- a/spyder/plugins/editor/panels/linenumber.py
+++ b/spyder/plugins/editor/panels/linenumber.py
@@ -43,7 +43,7 @@ class LineNumberArea(Panel):
         # Icons
         self.error_icon = ima.icon('error')
         self.warning_icon = ima.icon('warning')
-        self.information_icon = ima.icon('information')
+        self.info_icon = ima.icon('information')
         self.hint_icon = ima.icon('hint')
         self.todo_icon = ima.icon('todo')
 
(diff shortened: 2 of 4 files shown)


goanpeca:
@ccordoba12 currently the line number area is not changing size when zooming in/out. I think it perhaps should?

CAM-Gerlach:
@goanpeca I cannot properly test this due to #9425 , which I've repro'd on multiple branches consistently. However, some initial comments:

* Could you capitalize the first letter in the Pyflakes message, so it matches the others and doesn't look out of place?
* If Pyflakes doesn't have error codes, can you just make the parenthetical be ``(pyflakes)`` s (or ``(pyflakes error)``/``(pyflakes warning)`` if there are multiple types) since there's no reason to show the whole thing when there's no error code?
* ``Code`` is redundant and appears in every message, taking up space; can you just remove it?

Also, how do you enable those pylint errors? They don't show up, even after running static code analysis.

goanpeca:
Thanks for the review @CAM-Gerlach 

>Could you capitalize the first letter in the Pyflakes message, so it matches the others and doesn't look out of place?

Yep

>If Pyflakes doesn't have error codes, can you just make the parenthetical be (pyflakes) s (or (pyflakes error)/(pyflakes warning) if there are multiple types) since there's no reason to show the whole thing when there's no error code?

Agreed

>Code is redundant and appears in every message, taking up space; can you just remove it?

Not sure, @ccordoba12 wanted it to be obvious so I will defer to his judgement.


>Also, how do you enable those pylint errors? They don't show up, even after running static code analysis.

You have to use the latest pyls (but spyder will complain you need a lower version)

goanpeca:
Screen Shot 2019-05-24 at 21 40 23

CAM-Gerlach:
I assume you're aware, but just in case, ``pyflakes`` messages still have the spurious ``E`` in your screenshot.

goanpeca:
>I assume you're aware, but just in case, pyflakes messages still have the spurious E in your screenshot.

They do have error codes http://flake8.pycqa.org/en/2.5.5/warnings.html

CAM-Gerlach:
@goanpeca Okay, but I'm a little confused that's ``flake8``, not vanilla ``pyflakes``, which it explicitly says on that page does not provide error code (and none are provided in your screenshot).

goanpeca:
@CAM-Gerlach you are right, my bad!

goanpeca:
Screen Shot 2019-05-24 at 23 34 00

ccordoba12:
Thanks @goanpeca for this! Things look really nice!!

I have a couple of quick comments:

1. Instead of using the information sign for hints, please use `mdi-lightbulb` to follow VSCode lead regarding that symbol.
2. Please use a nice shade of green instead of blue for hints, because blue is already used by TODO's.

ccordoba12:
> > Code is redundant and appears in every message, taking up space; can you just remove it?

I disagree. In ten years of maintaining Spyder nobody has complained about it. Besides, it's much better to be explicit than a bit vague about those messages too. So in my role as BDFL of Spyder I say that stays.

spyder/utils/icon_manager.py
```diff
@@ -161,6 +163,8 @@
     'gotoline':                [('fa.sort-numeric-asc',), {'color': MAIN_FG_COLOR}],
     'error':                   [('fa.times-circle',), {'color': 'darkred'}],
     'warning':                 [('fa.warning',), {'color': 'orange'}],
+    'information':             [('fa.info-circle',), {'color': '#3775a9'}],
+    'hint':                    [('fa.lightbulb-o',), {'color': 'yellow'}],
```
goanpeca: >Instead of using the information sign for hints, please use mdi-lightbulb to follow VSCode lead regarding that symbol.

@ccordoba12 this is already the case, but none of those messages are hints, they are information. I am not sure how to even get a hint

goanpeca:
>Please use a nice shade of green instead of blue for hints, because blue is already used by TODO's.

Ok, I will post a screenshot to see how it looks

goanpeca:
>I disagree. In six years of maintaining Spyder nobody has complained about it. Besides, it's much better to be explicit than a bit vague about those messages too. So in my role as BDFL of Spyder I say that stays.

I was referring to the word `code` in messages I added (see first screenshot see the word `code` in parentheses), not about the title.

ccordoba12:
> I was referring to the word code in messages I added

Ok, sorry. Perhaps they should stay? I mean it's hard to understand what those messages mean without the word `code` on them.

spyder/utils/icon_manager.py
ccordoba12: Ok, don't worry, I thought those were hints. Let's go for using a shade of green for now.

goanpeca:
Screen Shot 2019-05-25 at 16 11 31

CAM-Gerlach:
If you want to keep "code" repeated in every message for the sake of explicitness, its not a big deal, though I'd prefer it without it since the user can reasonably infer what it means (at worst, they have to google it their first time, which is exactly what they likely want to do anyway).

However, I do strongly disagree with the green color for information. Blue has long been the standard color for this across all the UIs I'm familiar with, while Green means "Good"/"Ok", which is not the case here. For example, the [first](https://mdbootstrap.com/docs/jquery/css/colors/) [four](https://blog.prototypr.io/basic-ui-color-guide-7612075cc71a) [relevant](https://blog.prototypr.io/basic-ui-color-guide-7612075cc71a) [results](https://isabelcastillo.com/error-info-messages-css) for ``errors warnings information colors`` state that the universal convention is red-error, orange/yellow-warning, blue-info, and green-ok/success/good. If you feel it conflicts unduly with the annotation color, then that color should be changed to something else, like grey.

bcolsen:
I really like this change!

> However, I do strongly disagree with the green color for information.

I would go back to the blue for the info.

> Ok, sorry. Perhaps they should stay? I mean it's hard to understand what those messages mean without the word code on them.

I think `pyflake: E0602` would look good and still googles well. I think the word "code" in there is a little distracting.

I also think that the error codes are in general a little distracting and not too important on first glance or for new users. Can we make the error stand out from the error code with a different brightness or maybe color. For example have the `(pyflake: E0602)` part a darker grey or make the other part lighter.

ccordoba12:
> I would go back to the blue for the info.

We have a lot of things that are blue in the editor panels: the TODO's symbol (as I said before) and the debugger arrow. So if you want blue for info, we really need to change to green those other symbols. @goanpeca, please do that as part of this PR, but you also need to change those colors in `ScrollFlagArea`.

> I think pyflake: E0602 would look good and still googles well. I think the word "code" in there is a little distracting

Ok, fair enough. I agree with this suggestion.

> For example have the (pyflake: E0602) part a darker grey or make the other part lighter

This is a good suggestion too. I think the error code should be lighter to make it less relevant than the rest of the text.

CAM-Gerlach:
> We have a lot of things that are blue in the editor panels: the TODO's symbol (as I said before) and the debugger arrow. 

TODOs/etc can be grey or (if not visible enough) purple. Debugger error can be green.

>I think pyflake: E0602 would look good

Agreed as well (though to be really nitpicky, its ``pyflakes`` and its the one that doesn't have error codes :)

> I think the error code should be lighter to make it less relevant than the rest of the text.

I agree with having it a different shade, but in our default dark theme, having it lighter would make it much *more* visible/prominent, not less. If we wanted something that would work across our themes, we could just use *italic* instead (though that might not look quite as clean)?

goanpeca:
Screen Shot 2019-05-29 at 18 55 46

goanpeca:
We will still need to go over all the colors in Spyder at some point before the official release of 4.0 so I suggest we do not take more time than it is needed to get this functionality going. We should be moving faster and leave all the color and style discussions for another issue/PR.

goanpeca:
On another note, do we plan to add additional highlights to the code editor words to indicate the problems (errors, warning, info, hint) on the text itself, using curly underline and colors?

Screen Shot 2019-05-29 at 19 15 33

ccordoba12:
> Thoughts @bcolsen @ccordoba12 ?

Beautiful!! I really like it! I was about to suggest using magenta for TODOs, so that's really cool.

> We should be moving faster and leave all the color and style discussions for another issue/PR and focus on the actual tooltip (what this PR is about) and not on the icons and colors on the LineNumber/ScrollBar area.

Ok, agreed. Please open a new issue about it so we don't forget to address that, either in beta3 or beta4.

We should do two things about that: define a central place to import different colors and (in my opinion) change the debugger arrow to green.

> On another note, do we plan to add additional highlights to the code editor words to indicate the problems (errors, warning, info, hint) on the text itself, using curly underline and colors?

Yes, I was also about to suggest that. Please open another issue about it.

PR #9452, opened 2019-05-28
PR: Add option to increase/decrease icon size in toolbars

### LINKED ISSUE #9413: Add options to increase/decrease icon size in toolbars
## Problem Description

This is part of https://github.com/spyder-ide/spyder/issues/9004 Epic.

As suggested in https://github.com/spyder-ide/spyder/issues/8075, the idea is to add an `Icon size` option so that it is possible to increase/decrease icon size in toolbars.

I think there should be 3 options to choose from: `small`, `normal`, and `large`.

Obviously, this option should affect the icon size in the toolbar of the main interface, as well as the icons of the working directory plugin.

But what about the icon size in the toolbars of the other plugins like the outline explorer and plots plugins? @ccordoba12 @goanpeca any thoughts? Once this question is settled, I will submit a PR to implement this.

image

image

### LINKED ISSUE #9612: Simulate toolbar separator for fake toolbars
### Issue Report Checklist

* [X] Searched the [issues page](https://github.com/spyder-ide/spyder/issues?q=is%3Aissue) for similar reports
* [X] Read the relevant sections of the [Spyder Troubleshooting Guide](https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ) and followed its advice

## Problem Description

They appears very different. The size in toolbars are 16px and they are using icons. May it is a good idea create images for them or adjust size and colors for VLines.

Screenshot Linux 2019-06-19 14 49 44

Identical separators as toolbars.

## Versions

- Spyder 4.0.0.dev0 (Commit: 1a079fc93) 
- Python 3.6.8 64-bit
- Qt 5.12.3
- PyQt5 5.12.2
- Linux 4.15.0-52-generic
- QdarkStyle Develop (Commit: 18ef745e2674) 

### Dependencies

pygments >=2.0     :  2.4.2 (OK)
qdarkstyle >=2.6.4 :  2.8 (OK)
sphinx >=0.6.6     :  2.0.1 (OK)
psutil >=0.3       :  5.6.2 (OK)
nbconvert >=4.0    :  5.5.0 (OK)
pandas >=0.13.1    :  None (NOK)
numpy >=1.7        :  1.16.4 (OK)
sympy >=0.7.3      :  None (NOK)
cython >=0.21      :  None (NOK)
qtconsole >=4.5.0  :  4.5.1 (OK)
IPython >=4.0      :  7.5.0 (OK)
matplotlib >=2.0.0 :  None (NOK)

jnsebgosselin:
## Description of Changes

* [x] Implement the iconsize option for the toolbar of the mainwindow
* [x] Implement the iconsize option for the toolbars of all the plugins
* [x] Implement iconsize option for find and replace widgets.
* [x] Improve sizeHint calculation in `BaseTabBar`.

image

icon_size_option_spyder

### Issue(s) Resolved

Fixes #9413
Fixes #9612

### Affirmation


jnsebgosselin:
@goanpeca or @ccordoba12 Can one of you take a quick look at this before I implement the option to each plugin. I've done the `Plots`, `OutlineExplorer` and `IPython` for now as a POC.

ccordoba12:
I skimmed through this and it looks good. We can help you to sort out the details of how to better expose `SpyderPluginToolbar` in our API, but that's kind of easy.

So please keep working on this!

jnsebgosselin:
> 
> 
> I skimmed through this and it looks good. We can help you to sort out the details of how to better expose `SpyderPluginToolbar` in our API, but that's kind of easy.
> 
> So please keep working on this!

Excellent, thank you Carlos.

goanpeca:
Looks great @jnsebgosselin ! thanks

jnsebgosselin:
Ok, I think I am mostly done here guys. This was a little bit more work than what I anticipated. I just need to add a test, but I would like to do it after a first review is done of the PR. 

Two comments :

- The horizontal spacing of the items in the toolbar is set from the pixel metric `PM_ToolBarItemSpacing` which is taken from the style. The value is both 0 for the WindowsVista and for the dark style too it seems. So that is why there is no (nice) spacing anymore between the button for any pane in the dark style.

- I don't know if the height of non-QToolButton element (e.g. QLineEdit, QComboBox) should be expanded as well to match the height of the QToolButton. This should be possible to do this fairly easy application-wise through the new `SpyderPluginToolbar` class by changing the vertical `QSizePolicy` on those elements.

This is how it looks with an icon size value of 32 for the main toolbar and 24 for the panes:

image

image

goanpeca:
This is uber cool @jnsebgosselin 

Thanks a lot!

now regarding your comments

>The horizontal spacing of the items in the toolbar is set from the pixel metric PM_ToolBarItemSpacing which is taken from the style. The value is both 0 for the WindowsVista and for the dark style too it seems. So that is why there is no (nice) spacing anymore between the button for any pane in the dark style.

We should probably add a spacing or a spaceritem or update that style value you point out? so that buttons have a small space, otherwise it looks weird

>I don't know if the height of non-QToolButton element (e.g. QLineEdit, QComboBox) should be expanded as well to match the height of the QToolButton. This should be possible to do this fairly easy application-wise through the new SpyderPluginToolbar class by changing the vertical QSizePolicy on those elements.


I agree, please do this so the heights are consistent :-).

I will do a first pass review

jnsebgosselin:
> We should probably add a spacing or a spaceritem or update that style value you point out? so that buttons have a small space, otherwise it looks weird

I think this would be a bad idea to hard code a value for the spacing and that it is better to use the values that are defined in the style whenever possible.

I think the solution would be to change the value for `PM_ToolBarItemSpacing`, but for the dark style only. In the WindowsVista theme, I think having no space between tool buttons looks great and consistent with the overall Windows 10 desktop experience.

If possible, maybe the new value for `PM_ToolBarItemSpacing` could be set in the dark style itself instead of doing this in Spyder?

>> I don't know if the height of non-QToolButton element (e.g. QLineEdit, QComboBox) should be expanded as well to match the height of the QToolButton. This should be possible to do this fairly easy application-wise through the new SpyderPluginToolbar class by changing the vertical QSizePolicy on those elements.

> I agree, please do this so the heights are consistent :-).

Ok, that is what I thought too.

goanpeca:
>I think the solution would be to change the value for PM_ToolBarItemSpacing, but for the dark style only. In the WindowsVista theme, I think having no space between tool buttons looks great and consistent with the overall Windows 10 desktop experience.

@dpizetta thoughts?

jnsebgosselin:
@goanpeca I think this can also be handled this directly in the `SpyderPluginToolbar` by using the `from spyder.config.gui import is_dark_interface`. It should be very simple to do.

goanpeca:
Yes, it is what I hd in mind but you raised a valid point and maybe qe should update thing directly on qdarkstyle

jnsebgosselin:
Another idea also that looks good I think would be to remove the frame around the toolbuttons when there is no mouse hover or when the button are not pressed. I think there is too much line in the dark style, but that is only me.

toolbutton_noframe

dpizetta:
> Yes, it is what I hd in mind but you raised a valid point and maybe qe should update thing directly on qdarkstyle
@goanpeca  we can apply those changes, it will benefit all.

@jnsebgosselin  I agree with you about the lines/frames. We can add the changes for the buttons for the next version 2.7. I've already implemented the way you suggest. I'm testing remove other non-used lines or frames but, depending on the interface some of them seems strange. So more lines will be cut with time but not all of them (for Spyder) could be implemented, so some patch should be needed. [And this week I will push the changes for the buttons](https://github.com/ColinDuquesnoy/QDarkStyleSheet/issues/176). If possible, share this gif there for documenting also.

I've added other thing to [qdarkstyle about the background color for menus ](https://github.com/ColinDuquesnoy/QDarkStyleSheet/issues/179) 

I'm running the last version of your repo, may we can add some padding 2px inside docks. The things are to close here for example. Some docks have space, others not. (Pyqt5=5.12.2, Qt=5.12.3)

Screenshot Linux 2019-05-30 15 32 12
Screenshot Linux 2019-05-30 15 31 54
Screenshot Linux 2019-05-30 15 32 53


Screenshot of the unreleased version of qdarkstyle (commit 785ef7d0f6091297f1ff72a70f39e1dcf31fc7d3 (HEAD -> enhance-standard-svg, origin/enhance-standard-svg)).
Spyder from (commit 7a17e4c2b64b32016fa8b6b0a5202e72fd3caf1c (HEAD -> master, origin/master, origin/HEAD))
Screenshot Linux 2019-05-30 15 32 53

jnsebgosselin:
Thanks you very much @dpizetta 

> I'm running the last version of your repo, may we can add some padding 2px inside docks. The things are to close here for example. Some docks have space, others not. (Pyqt5=5.12.2, Qt=5.12.3)

The space around the docks are handled by a Spyder preference, so this is not really needed I think. I've fixed many layout glitch in this PR that should ensure that all panes now have the same padding, set from the config value.

However, as you have noticed, there is an extra padding around the `Variable Explorer` and `Plots` panes in the the dark theme that is not present in the WindowsVista theme. I'll open a new issue to discuss this on the QDarkStyleSheet repo once I'm done here.

image

spyder/api/plugins.py
```diff
@@ -16,14 +16,14 @@
...
-from qtpy.QtCore import Qt, Signal
+from qtpy.QtCore import QSize, Qt, Signal
 from qtpy.QtGui import QCursor
 from qtpy.QtWidgets import (QApplication, QMenu, QMessageBox, QToolButton,
                             QWidget)
 
 # Local imports
 from spyder.config.base import _
-from spyder.config.gui import get_color_scheme, is_dark_interface
+from spyder.config.gui import get_color_scheme, is_dark_interface, get_iconsize
```
goanpeca:
```suggestion
from spyder.config.gui import get_color_scheme, get_iconsize, is_dark_interface
```

spyder/api/plugins.py
```diff
@@ -282,6 +284,18 @@ def get_plugin_icon(self):
         """
         return ima.icon('outline_explorer')
 
+    def set_plugin_icon_size(self, iconsize):
+        """
+        Set the icon size of plugin.
+
+        This must be reimplemented by plugins that need to adjust the icon
+        size of their toolbar.
```
goanpeca: @jnsebgosselin, question

Since we are making things uniform here

>This must be reimplemented by plugins that need to adjust the icon size of their toolbar.

Is this docstring still valid? I mean we are handling this for other developers, no?

spyder/api/plugins.py
```diff
@@ -282,6 +284,18 @@ def get_plugin_icon(self):
...
+        """
+        try:
+            self.options_button.setIconSize(QSize(iconsize, iconsize))
+        except AttributeError:
```
goanpeca: Why do we need the try/except? An example in a comment could add clarity to this section of code

spyder/api/toolbar.py
```diff
@@ -0,0 +1,146 @@
...
+        super(SpyderPluginToolbar, self).__init__(parent)
+        self.layout = QGridLayout(self)
+        self.layout.setHorizontalSpacing(0)
+
+        bottom_margin = QApplication.instance().style().pixelMetric(
+            QStyle.PM_ToolBarItemMargin)
+        self.layout.setContentsMargins(0, 0, 0, bottom_margin)
+
+    def _add_hboxlayout_at_row(self, row):
+        """Add a new QHBoxLayout at row row."""
```
goanpeca:
```suggestion
        """Add a new QHBoxLayout at row."""
```

jnsebgosselin:
Ok, I've found a workaround to remove the extra padding that appears in the `Variable Explorer` and `Plots` plugin. We need to do this to remove the padding and the frame of the `QStackedWidget`:
```
self.stack = QStackedWidget(self)
self.stack.setStyleSheet("QStackedWidget{padding: 0px; border: 0px}")
```

spyder/api/toolbar.py
```diff
@@ -0,0 +1,146 @@
...
+        row_layout = QHBoxLayout()
+        row_layout.setContentsMargins(0, 0, 0, 0)
+        row_layout.setSpacing(QApplication.instance().style().pixelMetric(
+            QStyle.PM_ToolBarItemSpacing))
+        colspan = 1 if row == 0 else 3
+        self.layout.addLayout(row_layout, row, self.CONTENT_COL, 1, colspan)
+
+    def _get_hboxlayout_at_row(self, row):
+        """
+        Return the QHBoxLayout at row row. Add one if it doesn't already
```
goanpeca: row row?

Another way of writing these docstrings if you want to include the parameter is to write.


```python
"""
Return the QHBoxLayout at `row`. Add one if it doesn't already
"""
```

spyder/api/toolbar.py
```diff
@@ -0,0 +1,146 @@
...
+        exist.
+        """
+        if self.layout.itemAtPosition(row, self.CONTENT_COL) is None:
+            self._add_hboxlayout_at_row(row)
+        return self.layout.itemAtPosition(row, self.CONTENT_COL)
+
+    def set_row_visible(self, row, state):
+        """
+        Set the visibility of all widgets at row row to state.
```
goanpeca: same here

spyder/api/toolbar.py
```diff
@@ -0,0 +1,146 @@
...
+        for index in range(hboxlayout.count()):
+            try:
+                hboxlayout.itemAt(index).widget().setVisible(state)
+            except AttributeError:
+                pass
+
+    def add_item(self, item, stretch=None, row=0):
+        """
+        Add a widget or a layout with an horizontal stretch factor stretch
+        to the end of row row of this toolbar.
```
goanpeca: see above

spyder/api/toolbar.py
```diff
@@ -0,0 +1,146 @@
...
+        except TypeError:
+            row_layout.insertLayout(column, item)
+
+        if stretch is not None:
+            row_layout.setStretchFactor(item, stretch)
+
+    def add_widget(self, widget, stretch=None, row=0):
+        """
+        Add a widget with an horizontal stretch factor stretch to the end
+        of row row of this toolbar.
```
goanpeca: see above

spyder/api/toolbar.py
```diff
@@ -0,0 +1,146 @@
...
+        row_layout = self._get_hboxlayout_at_row(row)
+        if spacing is None:
+            spacing = QApplication.instance().style().pixelMetric(
+                QStyle.PM_LayoutHorizontalSpacing)
+        elif spacing == 'label':
+            spacing = QApplication.instance().style().pixelMetric(
+                QStyle.PM_CheckBoxLabelSpacing)
+        row_layout.addSpacing(spacing)
+
+    def add_options_btn(self, options_btn, spacing=None, stretch=1):
```
goanpeca: I am personally not a fan of contractions on names :-p 

I suggest we use `add_options_button`

spyder/app/mainwindow.py
```diff
@@ -171,7 +171,7 @@
                                     create_module_bookmark_actions,
                                     create_program_action, DialogManager,
                                     create_python_script_action, file_uri)
-from spyder.config.gui import get_shortcut
+from spyder.config.gui import get_shortcut, get_iconsize
```
goanpeca:
```suggestion
from spyder.config.gui import get_iconsize, get_shortcut
```

spyder/plugins/console/plugin.py
```diff
@@ -39,6 +40,8 @@
 from spyder.widgets.reporterror import SpyderErrorDialog
 from spyder.api.plugins import SpyderPluginWidget
 from spyder.py3compat import to_text_string
+from spyder.api.toolbar import SpyderPluginToolbar
```
goanpeca:
```
from spyder.api.plugins import SpyderPluginWidget
from spyder.api.toolbar import SpyderPluginToolbar

```

spyder/plugins/findinfiles/widgets.py
```diff
@@ -44,6 +45,7 @@
 from spyder.utils.qthelpers import create_toolbutton
 from spyder.config.gui import get_font
 from spyder.widgets.waitingspinner import QWaitingSpinner
+from spyder.api.toolbar import SpyderPluginToolbar
```
goanpeca: let's try keep em sorted (by alphabet)

spyder/utils/qthelpers.py
```diff
@@ -562,13 +582,14 @@ def create_plugin_layout(tools_layout, main_widget=None):
     """
     layout = QVBoxLayout()
     layout.setContentsMargins(0, 0, 0, 0)
-    spacing = calc_tools_spacing(tools_layout)
-    if spacing is not None:
-        layout.setSpacing(spacing)
-
-    layout.addLayout(tools_layout)
+    layout.setSpacing(0)
+    try:
```
goanpeca: Why is this needed? perhaps a comment would be nice :-)

spyder/widgets/tabs.py
```diff
@@ -132,14 +134,50 @@ def edit_finished(self):
...
 
-class TabBar(QTabBar):
+class BaseTabBar(QTabBar):
+    """Tabs base class with variable tab height support."""
+
+    def __init__(self, parent=None):
+        QTabBar.__init__(self, parent)
+        self._tabheight = None
+
+    def set_tabheight(self, height):
```
goanpeca: `set_tab_height`?

spyder/widgets/tabs.py
```diff
@@ -132,14 +134,50 @@ def edit_finished(self):
...
+            QStyle.PM_TabBarTabVSpace)
+
+        self._tabheight = max(iconsize.height(), fm.height(), height) + vframe
+        if self.parent():
+            self.parent().repaint()
+        else:
+            self.repaint()
+
+    def tabSizeHint(self, index):
+        """Qt method override."""
```
goanpeca: `Override Qt method.`

spyder/widgets/tabs.py
```diff
@@ -132,14 +134,50 @@ def edit_finished(self):
...
+        if self._tabheight is None:
+            return QTabBar.tabSizeHint(self, index)
+        else:
+            return QSize(QTabBar.tabSizeHint(self, index).width(),
+                         self._tabheight)
+
+    def sizeHint(self):
+        """Qt method override."""
```
goanpeca: Same as above

goanpeca:
@jnsebgosselin the variable explorer and plots pane show some space that the other plugins are not showing.

Also the (Ipyhton) Console icon on the left does not look positioned the same way as the icon on the right.

toolbar

jnsebgosselin:
Thanks for your review @goanpeca. I'll adress your comment/suggestion soon.

> @jnsebgosselin the variable explorer and plots pane show some space that the other plugins are not showing.

This was fixed by my latest commit.

> Also the (Ipyhton) Console icon on the left does not look positioned the same way as the icon on the right.

Good point. I'll investigate and try to fix this.

goanpeca:
>Ok, I've found a workaround to remove the extra padding that appears in the Variable Explorer and Plots plugin. We need to do this to remove the padding and the frame of the QStackedWidget:

Awesome, this is really looking slick! :-)

jnsebgosselin:
> Also the (Ipyhton) Console icon on the left does not look positioned the same way as the icon on the right.

This is supposed to be fixed now by https://github.com/spyder-ide/spyder/pull/9452/commits/486534cb34cf759f9a6571f8f5c2012ed8611910. It looks ok for me on Windows 10. 

@goanpeca can you check if this is fix for you too? Thank you!

spyder/api/plugins.py
jnsebgosselin: That was me being lazy.

I did it to avoid refactoring the Project and Breakpoints plugin layout, since their toolbar only add the options button. Anyway, since we are making things uniform, I did it in the end and so I removed this confusing implementation from the API.

goanpeca:
Screen Shot 2019-05-31 at 09 49 31

@jnsebgosselin it does

My other concern (but it might be a thing we need to fix on QDarkStyle). Do you see that the tool buttons top border is cropped?

spyder/api/plugins.py
jnsebgosselin: In this PR, I tried to make the creation of the toolbars and main layout uniform.

However, the layout of the plugins encompass a wide range of complexities and more often than not, the plugin themselves are not aware of the toolbar, because it is added deeper in the layout.

Moreover, some plugins have tab bar, other have find widget(s), other have a stack widget. So I think it is better to let the developers some flexibility and let them handle this? Otherwise, we would have to cover a lot of possibilities.

I've implemented the `set_plugin_icon_size` for all the default plugins in this PR. You can take a look and if you think we should handle this all here instead, I can try to find a way.

dpizetta:
Something minor, but I've noticed the colors from bug mark are not matching. Also, the X appears to be a bit darker, may using the color from bug-bars marks are better.

Screenshot Linux 2019-05-31 11 43 43

dpizetta:
> My other concern (but it might be a thing we need to fix on QDarkStyle). Do you see that the tool buttons top border is cropped?

I can't see this (either using latest version of qdarkstyle nor pypi version)

Screenshot Linux 2019-05-31 12 00 08

goanpeca:
Ok @dpizetta it is a mac issue then

jnsebgosselin:
> My other concern (but it might be a thing we need to fix on QDarkStyle). Do you see that the tool buttons top border is cropped?

This doesn't happen for the buttons in the tabbar because the corner widgets are aligned to the bottom in Windows 10 (while they seems align to the top for you). However, it does happens in the pane toolbars, but only when `custom margins for pane` is set to 0px. Setting this setting to a value of 1px or more fix the issue.

image

jnsebgosselin:
> Something minor, but I've noticed the colors from bug mark are not matching. Also, the X appears to be a bit darker, may using the color from bug-bars marks are better.

@dpizetta nice catch. Could you please open a new issue about this because this is a little bit out of scope for this PR.

spyder/utils/qthelpers.py
jnsebgosselin: Just to be sure, are you talking about the `layout.setSpacing(0)` or the `try` block? Thanks.

ccordoba12:
> Could you please open a new issue about this because this is a little bit out of scope for this PR.

@goanpeca is already tracking that in a previous issue.

spyder/utils/qthelpers.py
goanpeca: (the try!)

jnsebgosselin: I added a comment to explain this.

Maybe we can also raise a deprecated warning to warn maintainers of external plugins of this change?

jnsebgosselin:
image

jnsebgosselin:
> I don't know if the height of non-QToolButton element (e.g. QLineEdit, QComboBox) should be expanded as well to match the height of the QToolButton. This should be possible to do this fairly easy application-wise through the new SpyderPluginToolbar class by changing the vertical QSizePolicy on those elements.

>> I agree, please do this so the heights are consistent :-).

I have tested this and I'm not sure I like the results. Do you think this can be discussed and/or implemented in a following Issue/PR?

jnsebgosselin:
I've added some horizontal space around the button for the dark style just to see how this look.

image

dpizetta:
About toolbar separator, try v2.7 recently released. There are many improvements also. TKS, good job!

jnsebgosselin:
After much thinking and testing, and after the changes introduced to QDarkStyle for the ToolButton (see https://github.com/ColinDuquesnoy/QDarkStyleSheet/commit/7d1eab947d758ef6a773c48e36feabd3245b8160), I think we should add a little bit of horizontal spacing between the toolbar items independently of the theme used. Even for the WindowsVista theme, I think this would look better.

See for example the example for the toolbar of the `Find` panel:

image

@goanpeca what do you think?

dpizetta:
> After much thinking and testing, and after the changes introduced to QDarkStyle for the ToolButton (see ColinDuquesnoy/QDarkStyleSheet@7d1eab9), I think we should add a little bit of horizontal spacing between the toolbar items independently of the theme used. Even for the WindowsVista theme, I think this would look better.

Yeh! And we can use it directly on the qdark, it seems better for everyone. I would like to change the 3px to 2px, because toolbar already has a 2px padding, so I've tried 2px margin in tool buttons, then we get 4px of spacing. As each button has 2px margin, it will reflect 4px spacing between them. What do you think about it @jnsebgosselin ?

Take a look here: 

Screenshot Linux 2019-06-04 09 48 42

About this one:

Screenshot Linux 2019-06-04 09 48 04

It is strange that it seems that they are not separators from toolbars but the vertical (or horizontal in other cases) lines. Could you confirm this to me? So, I can fix it also. 

**Separators in toolbars and different toolbars**

Another thing, more or less related to this PR, take a look in the spyder toolbar, there are no separators for the buttons nor different toolbars (you can't move just one toolbar). It is common that when you create the toolbars, the same separators from the menu are applied for users to see the groups. Example in New (Novo) Menu, Open (Abrir) and Save (Salvar) are separated by a line. At least different operations should be separated. The menus seem to be not separated in different toolbars (for me this is more critical), like debug toolbar inside the main toolbar. So, the user cannot rearrange it. 
**Now I see the v3, there exist those different toolbars. This was changed in v4? Or it is a bug?**

**Version 4**

Screenshot Linux 2019-06-04 09 55 19

Screenshot Linux 2019-06-04 09 56 08

**Version 3**

Screenshot Linux 2019-06-04 10 07 22

jnsebgosselin:
> It is strange that it seems that they are not separators from toolbars but the vertical (or horizontal in other cases) lines. Could you confirm this to me? So, I can fix it also.

We do not use a `QToolbar` for the panes but a simple `QWidget`, that is why. So the separators are simple `QFrame` with a `Vline` style applied to them. I don't know if there is anything to fix here, at least in the scope of this PR I mean. Besides, only the `Plots` pane use separator in its toolbar.

jnsebgosselin:
> **Separators in toolbars and different toolbars**
> 
> Another thing, more or less related to this PR, take a look in the spyder toolbar, there are no separators for the buttons nor different toolbars (you can't move just one toolbar). It is common that when you create the toolbars, the same separators from the menu are applied for users to see the groups. Example in New (Novo) Menu, Open (Abrir) and Save (Salvar) are separated by a line. At least different operations should be separated. The menus seem to be not separated in different toolbars (for me this is more critical), like debug toolbar inside the main toolbar. So, the user cannot rearrange it.
> **Now I see the v3, there exist those different toolbars. This was changed in v4? Or it is a bug?**
> 
> **Version 4**

In Spyder 4, look for the option `Lock panes and toolbars`

image

jnsebgosselin:
> Yeh! And we can use it directly on the qdark, it seems better for everyone. I would like to change the 3px to 2px, because toolbar already has a 2px padding, so I've tried 2px margin in tool buttons, then we get 4px of spacing. As each button has 2px margin, it will reflect 4px spacing between them. What do you think about it @jnsebgosselin ?

This looks great, but does this means that anyone using a `QToolButton` in their app with qdark will now have a 2 px margin even if they set the spacing of their layout to 0? If they set the spacing of their layout to 3 px, is this going to result in a 5px margin? I can see this as an undesired side effect in many cases...

I may be wrong but I'm under the impression it is more flexible to let the developers manage the spacing between the various elements of their UI through their layout instead of forcing this on them through the style of the widgets.

What I'm trying to say here is, I'm not sure it is a good idea to tailor qdark specifically for Spyder because it makes it less flexible and I wonder if this spacing of the toolbuttons  shouldn't be handled on the Spyder side through the tool that are available with Qt layouts.

@goanpeca any thoughts?

goanpeca:
>What I'm trying to say here is, I'm not sure it is a good idea to tailor qdark specifically for Spyder because it makes it less flexible and I wonder if this spacing of the toolbuttons shouldn't be handled on the Spyder side through the tool that are available with Qt layouts.

My believe is QDarkStyle should either:
- Adapt to the specifics of the OS but in Dark or
- Provide a complete QDarkStyle that will not exactly follow the system OS

I think right now it is a mix of those.

Now to @jnsebgosselin question, agreed, QDarkStyle should not be changed so Spyder looks good, cause we of course use different things to make stuff work. But all "normal" usage and widgets do need to look good no matter what out of the box. In this respect it feels weird that the out of the box setting for a toolbar are buttons touching each other.

dpizetta:
> > It is strange that it seems that they are not separators from toolbars but the vertical (or horizontal in other cases) lines. Could you confirm this to me? So, I can fix it also.
> 
> We do not use a `QToolbar` for the panes but a simple `QWidget`, that is why. So the separators are simple `QFrame` with a `Vline` style applied to them. I don't know if there is anything to fix here, at least in the scope of this PR I mean. Besides, only the `Plots` pane use separator in its toolbar.

I can create the vlines/hlines with no double lines, as it is expected to be.

jnsebgosselin:
> Now to @jnsebgosselin question, agreed, QDarkStyle should not be changed so Spyder looks good, cause we of course use different things to make stuff work. But all "normal" usage and widgets do need to look good no matter what out of the box. In this respect it feels weird that the out of the box setting for a toolbar are buttons touching each other.

Good point. I don't understand well how qdark works, but what I think should be done and would be the ideal case is that qdark sets a default spacing value for the `QToolbar`, lets say to 2 or 3 px.

I still believe `QToolButtons` shouldn't have an embedded margins added by the style, because I think this should be handled by the layout to ensure consistency with all the other Qt widgets.

In the case of the pane toolbars, which are custom toolbars built from a standard `QWidget` that mimic a `QToolbar` (but with more flexibility), we should set the spacing between the elements ourselves, using the default value set for the `QToolbar` by the style.

dpizetta:
> Now to @jnsebgosselin question, agreed, QDarkStyle should not be changed so Spyder looks good, 

Agreed too. I'm just using the ideas to create a better UI for all other projects. If there is something that only Spyder (or another project) uses, then we don't need to change in qdark.

> But all "normal" usage and widgets do need to look good no matter what out of the box. In this respect it feels weird that the out of the box setting for a toolbar are buttons touching each other.

Yes, this is I meant.

> My believe is QDarkStyle should either: Adapt to the specifics of the OS but in Dark or; Provide a complete QDarkStyle that will not exactly follow the system OS;

I don't really get your point here. To me, the second option makes more sense.

General thoughts: qdark should make some choices that provide a better UI for most of the projects without developer intervention, so it is easy to just use it. Otherwise, if the project does not agree with the style characteristic it should be easy to change, and possible. I think we cannot escape from creating issues, by leaving that way or changing. However, offering concise app (and more stable as possible) is more important, I think so.

dpizetta:
> Good point. I don't understand well how qdark works, but what I think should be done and would be the ideal case is that qdark sets a default spacing value for the QToolbar, lets say to 2 or 3 px.
> I still believe QToolButtons shouldn't have an embedded margins added by the style, because I think this should be handled by the layout to ensure consistency with all the other Qt widgets.
> In the case of the pane toolbars, which are custom toolbars built from a standard QWidget that mimic a QToolbar (but with more flexibility), we should set the spacing between the elements ourselves, using the default value set for the QToolbar by the style.

We can add those only in tool buttons inside toolbars.  I'll open a discussion in qdark for this, so we not make noise here :), which is out of this PR. Tks for the discussion @jnsebgosselin and @goanpeca .

jnsebgosselin:
> We can add those only in tool buttons inside toolbars.

That would be great, but why not just set the spacing of `QToolbar` with something like `'QToolBar{spacing:2px;}'`, so that we get a uniform and homogeneous spacing between all items added to the toolbar, not just toolbuttons

But more importantly in regard to this PR, how can I get properly the value use for toolbars spacing from qdark, so that I can use it in our own custom pane toolbar and tabbar?

Currently, I am using a pixel metrics from QStyle with :
`QApplication.instance().style().pixelMetric(QStyle.PM_ToolBarItemSpacing))`

Is qdark going to affect the value returned by the pixel metric above?

> I'll open a discussion in qdark for this, so we not make noise here :), which is out of this PR. Tks for the discussion @jnsebgosselin and @goanpeca .

Good idea!

jnsebgosselin:
I'm going to try using a `QMainWindow` as the base class for the `SpyderPluginToolbar` , I've played with the idea a little and I think this looks very promising. So this it not ready for review yet.

jnsebgosselin:
Ok I have tried to make things work using a QMainWindow to build the plugins toolbar and I must say this is not easy due to the high variability and complexity of our plugin layout. Moreover :

- This QT bug https://bugreports.qt.io/browse/QTBUG-597 causes problem, especially for the `Find` widget.
- This also makes the plugin toolbars look bad, especially under the dark style. So we need to add more workaround to make things look good.
- Adding spaces, stretching and even hiding widgets become more complicated and require more workarounds.

So I propose we keep using a more versatile and standard layout to build our plugin toolbars and I will propose a solution for the separators that will hopefully address #9612.

jnsebgosselin:
Here is a screenshot of how the plugin toolbar separators look now:

image

dpizetta:
Amazing!

ccordoba12:
@jnsebgosselin, please merge with latest master to get the fixes for the currently failing tests.

ccordoba12:
@jnsebgosselin, please give another review to @jnsebgosselin's work. Thanks!

jnsebgosselin:
I think I messed up with one of the merge conflicts. I'm going to investigate and fix it tonight.

goanpeca:
>@jnsebgosselin, please give another review to @jnsebgosselin's work. Thanks!

I think you meant me :-p @ccordoba12 

@jnsebgosselin sorry for all the merge conflicts, let me know when I can take a final review at this one and thanks again for the amazing work :-)

jnsebgosselin:
Ok @goanpeca I fixed the bad merge. It is ready for your review. Thank you very much!

spyder/plugins/findinfiles/widgets.py
```diff
@@ -991,16 +983,15 @@ def __init__(self, parent,
...
 
-        layout = QVBoxLayout()
-        left, _x, right, bottom = layout.getContentsMargins()
-        layout.setContentsMargins(left, 0, right, bottom)
-        layout.addWidget(self.find_options)
-        layout.addLayout(hlayout)
-        layout.addWidget(self.status_bar)
-        self.setLayout(layout)
+        # Create plugin layout.
+        self.setLayout(create_plugin_layout(self.find_options, main_widget))
```
ccordoba12: Perhaps we should rename `create_plugin_layout` because now it's not only used for plugins?

spyder/api/plugins.py
```diff
@@ -381,6 +381,15 @@ def get_plugin_icon(self):
         """
         return ima.icon('outline_explorer')
 
+    def set_plugin_icon_size(self, iconsize):
```
ccordoba12: Please describe in the docstring what values `iconsize` receives.

spyder/utils/qthelpers.py
```diff
@@ -575,26 +595,55 @@ def calc_tools_spacing(tools_layout):
         return max(spacing, 0)
 
 
-def create_plugin_layout(tools_layout, main_widget=None):
+def calcul_toolbutton_size(iconsize, style=None):
```
ccordoba12: `calcul` -> `compute`?

jnsebgosselin:
Hey @ccordoba12, sorry for the delay on that one. To be honest, I'm not really happy with the work I have done here. I had to implement a feature similar to this in one of my current project at work and I think there is a way to do this a lot better using the functionality of `QMainWindow` and `QToolbar` like @dpizetta suggested earlier. This would make the code a lot cleaner and also make the layout of the toolbars adaptable, which is a nice thing to have. 

Once I'm finished with my own project, and if this feature is still needed, I think I could try to implement this in a better way from a fresh start.

By adaptable layout, I mean:

> When a toolbar is resized in such a way that it is too small to show all the items it contains, an extension button will appear as the last item in the toolbar. Pressing the extension button will pop up a menu containing the items that do not currently fit in the toolbar.

ccordoba12:
Thanks for the update @jnsebgosselin!

> I think I could try to implement this in a better way from a fresh start.

Ok, then I'm going to close this one for now.

PR #9453, opened 2019-05-28
PR: Enable line stripping by default

impact27:
## Description of Changes

I missed this before the merging of https://github.com/spyder-ide/spyder/pull/8734

### Issue(s) Resolved

Fixes #

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -686,7 +686,7 @@ class CodeEditor(TextEditBaseWidget):
                      color_scheme=None,
                      wrap=False,
                      tab_mode=True,
-                     strip_mode=False,
+                     strip_mode=True,
                      intelligent_backspace=True,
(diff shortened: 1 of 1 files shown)


ccordoba12:
I'm not sure if this is the right thing to do because it could be confusing to newbies.

@bcolsen, @goanpeca, what do you think about this?

bcolsen:
> I'm not sure if this is the right thing to do because it could be confusing to newbies.

I agree that it might be confusing having white space removed without an explicit command. Even the the removal of the automatic indentation on blank line makes it more difficult to to insert new code into those spaces later on.

In general I would categorize this as a code style feature, and I personally I don't think that code style features should be on by default. Programming is already too picky for the average person so we shouldn't make it even more picky.

goanpeca:
Yes, I agree that this might be too unexpected for newbies :-|

ccordoba12:
Thanks for the input guys. Sorry @impact27 but this could really be a lot for newbies.

PR #9485, opened 2019-06-01
PR: Use file contents to decide when to autosave

### LINKED ISSUE #9265: Autosaves are created for unmodified files and not removed on close with split Editor panes

jitseniesen:
## Description of Changes

This PR changes the autosave system so that it uses the file contents to decide whether to autosave or not. Currently, we use Qt's `textChanged` to see whether the user made any changes since the last autosave, and if yes, we autosave. It turns out that this leads to a lot of false positives. Running the highlighter emits `textChanged`, as does splitting the editor window. This leads to files being autosaved needlessly and possibly to user confusion when the user is asked whether to recover from these unnecessary autosave files. We can guard against this, but I think the design is quite fragile. In particular, if a user makes an edit and then undoes the edit, it will count as a change for the autosave system and I don't see a good way to detect this.

Here, I try a different approach: Spyder compares the contents of the editor against the contents of the file on disk and autosaves if they are different. The logic is in commit a079e6b. Spyder stores the hash of the file contents on disk so that they don't need to be read in again. A disadvantage of this approach is that we have to compute the hash of the contents of the editor whenever the autosave system fires (by default, once a minute). This may take some time when editing large files (I did not test this, but I guess it may be noticeable for files of 100 MB or more). However, I guess that other parts of Spyder are also problematic for such large files.

Additionally, this PR moves ownership of the autosave data from EditorStack level to plugin level (commit 
9c31c23). This ensures that when the user splits the editor pane, the autosave data in the EditorStacks are synchronized.

* [x] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/)) **N/A**

### Issue(s) Resolved

Fixes #9265

### Affirmation

I certify the above statement is true and correct: Jitse Niesen

### DIFF
--- a/spyder/plugins/editor/widgets/editor.py
+++ b/spyder/plugins/editor/widgets/editor.py
@@ -199,7 +199,6 @@ class FileInfo(QObject):
     def text_changed(self):
         """Editor's text has changed"""
         self.default = False
-        self.editor.document().changed_since_autosave = True
         self.text_changed_at.emit(self.filename,
                                   self.editor.get_position('cursor'))
 
@@ -1753,7 +1752,6 @@ class EditorStack(QWidget):
                                  finfo.filename, finfo.filename)
 
             finfo.editor.document().setModified(False)
-            finfo.editor.document().changed_since_autosave = False
             self.modification_changed(index=index)
             self.analyze_script(index)
 
@@ -2238,7 +2236,6 @@ class EditorStack(QWidget):
         position = finfo.editor.get_position('cursor')
         finfo.editor.set_text(txt)
         finfo.editor.document().setModified(False)
-        finfo.editor.document().changed_since_autosave = False
         self.autosave.file_hashes[finfo.filename] = hash(txt)
         finfo.editor.set_cursor_position(position)
 
@@ -2330,7 +2327,6 @@ class EditorStack(QWidget):
         if cloned_from is None:
             editor.set_text(txt)
             editor.document().setModified(False)
-            editor.document().changed_since_autosave = False
         finfo.text_changed_at.connect(
             lambda fname, position:
             self.text_changed_at.emit(fname, position))
--- a/spyder/plugins/editor/widgets/editor.py
+++ b/spyder/plugins/editor/widgets/editor.py
@@ -1514,7 +1514,7 @@ class EditorStack(QWidget):
                 self.set_stack_index(new_index)
 
             self.add_last_closed_file(finfo.filename)
- 
+
             if finfo.filename in self.autosave.file_hashes:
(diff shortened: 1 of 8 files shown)


jitseniesen:
Don't merge just yet, I realized something I need to check when writing the above.

jitseniesen:
This is now ready for review.

spyder/plugins/editor/widgets/editor.py
```diff
@@ -1668,6 +1666,19 @@ def save_if_changed(self, cancelable=False, index=None):
...
+        """Compute hash of contents of editor.
+
+        Args:
+            fileinfo: FileInfo object associated to editor whose hash needs
+                to be computed.
+
+        Returns:
+            int: computed hash.
+        """
+        txt = to_text_string(fileinfo.editor.get_text_with_eol())
```
ccordoba12: There's no need to use `to_text_string` here because `get_text_with_eol` returns text in unicode.

PR #9502, opened 2019-06-04
PR: Restore cursor properly in the Editor after a key release or a focus out event.

### LINKED ISSUE #9501: Cursor is not restore properly in the Editor
## Problem Description

The cursor in the Editor is not restored properly after it was changed to a `PointingHandCursor`.

cursor

1. Select a word or a url in the Editor
2.  Press `Ctrl + F` or `Ctrl + C` or `Ctrl + R`  combination
3.  Move the mouse around

Also

1. Go over a word with the mouse cursor
2.  Press `Ctrl`
3.  Release `Ctrl`

The cursor remains to `Qt.PointingHandCursor` while it should be restored with `restoreOverrideCursor`

## Versions

* Spyder version: Spyder 4.0.0.dev0 (Commit: 64a9d6ad9) 
* Python version: Python 3.7.1 64-bit
* Qt version: Qt 5.9.6
* PyQt version: PyQt5 5.9.2

jnsebgosselin:
## Description of Changes

In the animation below, first I'm pressing and releasing the `Ctrl` key two times, than I do the following shortcuts: `Ctrl+C`, `Ctrl+F` and `Ctrl+R`.

cursor_correct

### Issue(s) Resolved

Fixes #9501

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -3565,6 +3565,10 @@ class CodeEditor(TextEditBaseWidget):
     def focusOutEvent(self, event):
         """Reimplement Qt method"""
         self.sig_focus_changed.emit()
+        self.clear_extra_selections('ctrl_click')
+        QApplication.restoreOverrideCursor()
+        self.__cursor_changed = False
+        self._last_hover_uri = None
         super(CodeEditor, self).focusOutEvent(event)
 
     def leaveEvent(self, event):
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -3119,10 +3119,7 @@ class CodeEditor(TextEditBaseWidget):
         """Override Qt method."""
         self.sig_key_released.emit(event)
         self.timer_syntax_highlight.start()
-        self.clear_extra_selections('ctrl_click')
-        QApplication.restoreOverrideCursor()
-        self.__cursor_changed = False
-        self._last_hover_uri = None
+        self._restore_editor_cursor_and_selections()
         super(CodeEditor, self).keyReleaseEvent(event)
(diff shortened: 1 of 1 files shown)


goanpeca:
@ccordoba12 test in and works as expected. Thanks @jnsebgosselin

ccordoba12:
Great!! I'll merge as soon as everything is green.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3563,6 +3565,10 @@ def setPlainText(self, txt):
     def focusOutEvent(self, event):
         """Reimplement Qt method"""
         self.sig_focus_changed.emit()
+        self.clear_extra_selections('ctrl_click')
+        QApplication.restoreOverrideCursor()
+        self.__cursor_changed = False
+        self._last_hover_uri = None
```
goanpeca: Now that I think about it, could we refactor these actions? we are basically doing the same on both methods and what we are doing is kinda like a `reset editor stuff`

Thoughts @jnsebgosselin ?

jnsebgosselin: Yes, I agree with this. Do you have a name in mind for the method that I could use?

jnsebgosselin: Also @goanpeca, change your review to `request changes` so that Carlos don't ninja merge this before I make the changes :P

goanpeca: hahaha will do 

Well, it should begin with an underscore and we could call it

```python
def _restore_editor_cursor_and_selections(self):
```

Or somethinng similar ;-), up to you

jnsebgosselin: The only thing I am not sure is if it is correct to add `self._last_hover_uri = None` in there.

PR #9482, opened 2019-05-31
PR: Add support for LSP workspace calls

andfoy:
## Description of Changes
We've enabled support for LSP workspace capabilities, including changing workspace folders on projects and single files, as well as adding support for simultaneous workspaces. This PR also enables symbol searching.

* [x] Wrote at least one-line docstrings (for any new functions)
* [ ] Added unit test(s) covering the changes (if testable)

* [ ] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

### Issue(s) Resolved

Fixes #9007

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct:
@andfoy 

---
**TODO tasks**
- [x] Handle ``workspace/workspaceFolders`` call
- [x] Send ``workspace/didChangeWorkspaceFolders``
- [x] Handle ``workspace/configuration`` request
- [x] Send ``workspace/didChangeWatchedFiles`` request
- [x] Send ``workspace/symbol`` request
- [x] Connect ``workspace/didChangeWorkspaceFolders`` with project file watcher
~- [ ] Define a workspace client per registered project.~
~- [ ] Connect symbol search in project with find in files widget~

PR #9541, opened 2019-06-10
PR: Watch a project file tree for changes

andfoy:
## Description of Changes
Added a watchdog observer to receive notifications about files and folders changes. i.e., Additions, deletions, move to other folders.

### Issue(s) Resolved

Required by #9482 

### Affirmation

@andfoy


spyder/plugins/projects/workspace/watcher.py
```diff
@@ -0,0 +1,72 @@
...
+        logger.info("Created {0}: {1}".format(
+            self.fmt_is_dir(is_dir), src_path))
+        self.sig_file_created.emit(src_path, is_dir)
+
+    def on_deleted(self, event):
+        src_path = event.src_path
+        is_dir = event.is_directory
+        logger.info("Deleted {0}: {1}".format(
+            self.fmt_is_dir(is_dir), src_path))
+        self.sig_file_created.emit(src_path, is_dir)
```
ccordoba12: I think this should be `sig_file_deleted`.

spyder/plugins/projects/workspace/watcher.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def on_modified(self, event):
+        src_path = event.src_path
+        is_dir = event.is_directory
+        logger.info("Modified {0}: {1}".format(
+            self.fmt_is_dir(is_dir), src_path))
+        self.sig_file_created.emit(src_path, is_dir)
```
ccordoba12: And this `sig_file_modified`.

spyder/plugins/projects/workspace/watcher.py
```diff
@@ -0,0 +1,72 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+
```
ccordoba12: Please add a docstring to this module.

spyder/plugins/projects/workspace/watcher.py
```diff
@@ -0,0 +1,72 @@
...
+# Third-party imports
+from qtpy.QtCore import QObject, Signal, Slot
+
+from watchdog.observers import Observer
+from watchdog.events import FileSystemEventHandler
+
+logger = logging.getLogger(__name__)
+
+
+class WorkspaceEventHandler(QObject, FileSystemEventHandler):
```
ccordoba12: Please add a docstring for this class.

spyder/plugins/projects/workspace/watcher.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def on_modified(self, event):
+        src_path = event.src_path
+        is_dir = event.is_directory
+        logger.info("Modified {0}: {1}".format(
+            self.fmt_is_dir(is_dir), src_path))
+        self.sig_file_created.emit(src_path, is_dir)
+
+
+class WorkspaceWatcher(QObject):
```
ccordoba12: And also a docstring for this class.

andfoy:
> That way this watcher will be started both in the File and Project Explorers.

Maybe we should run some tests on a large directory to see if watchdog does not crash, at least that would be the case when Spyder is started on the home directory.

ccordoba12:
> Maybe we should run some tests on a large directory to see if watchdog does not crash, at least that would be the case when Spyder is started on the home directory.

The idea is to have activate the watcher only a project is active. Otherwise, it shouldn't be necessary.

andfoy:
>  I think you should move this to spyder/workers and add it to the File explorer directly

The idea here is to update the file explorer when watchdog notifies a change?

goanpeca:
@andfoy currently the file explorer is getting all these updates, no?? how is it working at the moment?

ccordoba12:
> The idea here is to update the file explorer when watchdog notifies a change?

Yeah, but only to the root folder (or folders) of a project (or projects).

> currently the file explorer is getting all these updates, no?? how is it working at the moment?

Nothing is connected yet.

goanpeca:
>Nothing is connected yet.

I meant that if I delete a file on the OS the file explorer is updated. So there must be some detection going on ?

andfoy:
> I meant that if I delete a file on the OS the file explorer is updated. So there must be some detection going on ?

I think the tree updates when the file is deleted/renamed directly on Spyder, what happens if the file is removed outside Spyder?

goanpeca:
file

andfoy:
> It works... (it has been like this for a loooong time)

https://github.com/spyder-ide/spyder/blob/f0e625450c6fb74347eb474989f2cab1c4099977/spyder/plugins/explorer/widgets.py#L1086

However, this won't tell us if a file/directory changes its name

ccordoba12:
That only changes the collapsed state, so it has no effect on the file system, I'd say.

andfoy:
@ccordoba12 So we stick with developing the file watcher?

goanpeca:
I am not saying we should not develop. I was just wondering how the file explorer does it (if it is using a Qt thing or what?) and also the watcher if developed to include missing functionality should allow a way to hook it to the file explorer (and the project explorer)

ccordoba12:
Yeah, I think we should keep developing it because `QFileSystemModel` doesn't seem to have the functionality that we need:

https://doc.qt.io/qt-5/qfilesystemmodel.html#signals

ccordoba12:
However, this could be used instead:

https://doc.qt.io/qt-5/qfilesystemwatcher.html#details

(although, again, it doesn't seem to have everything that we need).

andfoy:
@ccordoba12 This one is blocked by https://github.com/conda-forge/watchdog-feedstock/issues/9

ccordoba12:
> This one is blocked by conda-forge/watchdog-feedstock#9

Solved. I moved the new watchdog 0.9.0 package to our Anaconda channel, so everything should be fine now.

andfoy:
5

andfoy:
@ccordoba12 We're still having issues with watchdog installation on some images

ccordoba12:
Please try again, it should work now.

requirements/conda.txt
```diff
@@ -19,4 +19,4 @@ qdarkstyle >=2.6.4
 atomicwrites
 diff-match-patch
 pexpect
-
+watchdog
```
ccordoba12: Please also add `watchdog` to our `setup.py` and our Readme.

goanpeca: Should we add this also to the `requirements.txt?`

andfoy:
@ccordoba12 Tests run locally, however they are hanging on CIs

spyder/plugins/projects/workspace/watcher.py
ccordoba12: This module still has no docstring.

spyder/plugins/projects/workspace/watcher.py
```diff
@@ -0,0 +1,90 @@
...
+from watchdog.observers import Observer
+from watchdog.events import FileSystemEventHandler
+
+logger = logging.getLogger(__name__)
+
+
+class WorkspaceEventHandler(QObject, FileSystemEventHandler):
+    """Event handler for watchdog notifications. This class recieves
+       notifications about file/folder moving, modification,
+       creation and deletion and emits its corresponding signal."""
```
ccordoba12: This is a very odd way to write docstrings. Please change it too:

```python
"""
Event handler for watchdog notifications.

This class receives notifications about file/folder moving, modification,
creation and deletion and emits a corresponding signal about it.
"""
```

spyder/plugins/projects/workspace/watcher.py
```diff
@@ -0,0 +1,90 @@
...
+        src_path = event.src_path
+        is_dir = event.is_directory
+        logger.info("Modified {0}: {1}".format(
+            self.fmt_is_dir(is_dir), src_path))
+        self.sig_file_modified.emit(src_path, is_dir)
+
+
+class WorkspaceWatcher(QObject):
+    """Wrapper class around watchdog observer and notifier. Provides methods
+       to start and stop watching folders."""
```
ccordoba12: This docstring needs to be

```python
"""
Wrapper class around watchdog observer and notifier.

It provides methods to start and stop an observer to watch folders.
"""
```

spyder/plugins/projects/tests/test_plugin.py
```diff
@@ -289,5 +292,95 @@ def test_project_explorer_tree_root(projects, tmpdir, qtbot):
...
+    # Test folder creation
+    with qtbot.waitSignal(fs_handler.sig_file_created,
+                          timeout=3000) as blocker:
+        folder2 = project_root.mkdir('folder2')
+
+    folder_created, is_dir = blocker.args
+    assert folder_created == osp.join(to_text_string(project_root), 'folder2')
+
+    # Test file move/renaming
+    new_file = osp.join(to_text_string(folder0), 'new_file')
```
ccordoba12: I could only make this work if `new_file` is in the same directory as the original file. Is this perhaps a limitation of watchdog?

andfoy: At least on my local setup, it is working as expected, watchdog should be monitoring all the files and folders under the same root folder. I.e., recursive option is set to True

ccordoba12: Ok, that's great to know!

andfoy:
> Besides, I really don't know why the new test fails in macOS

Maybe is due to https://github.com/conda-forge/watchdog-feedstock/pull/8

ccordoba12:
You're most probably right. That package can't be `noarch` (the guy of that PR confused me because he said that the previous `noarch` package was working, but there was no such package).

spyder/plugins/projects/tests/test_plugin.py
```diff
@@ -289,5 +292,95 @@ def test_project_explorer_tree_root(projects, tmpdir, qtbot):
         assert topleft_index.data() == osp.basename(ppath)
 
 
+@pytest.mark.skipif(sys.platform == 'darwin', reason='Fails in macOS')
```
ccordoba12: I uploaded new packages for `watchdog` that include the missing library in macOS, so this shouldn't be necessary.

Please remove it to see what happens. If tests pass then this should be ready to go.

spyder/plugins/projects/workspace/watcher.py
ccordoba12: And please address this one!

andfoy:
@ccordoba12 It seems that we're still having issues with watchdog on Mac. @goanpeca does the test run in Mac locally?

ccordoba12:
> It seems that we're still having issues with watchdog on Mac

I skipped the failing part of the new test on macOS. I think it's really a minor, so we can skip it for now.

goanpeca:
I will check in a couple of hours

ccordoba12:
Thanks @andfoy, this looks really good!

PR #9554, opened 2019-06-11
PR: Remove stylesheet workarounds for the options and browse button when under the dark theme

jnsebgosselin:
## Description of Changes

As reported in https://github.com/ColinDuquesnoy/QDarkStyleSheet/issues/187, `QToolButton` were not styled correctly in `QDarkStyleSheet` when the `PopupMode` of the button was set to `InstantPopup`. Because of this, workarounds were added in Spyder to make the options and browse buttons look ok.

Since this is going to be fixed in version 2.8 of `QDarkStyleSheet`, these workarounds are not longer necessary, so this PR propose to remove them. Besides, there were some problems with these workarounds :

- First, the padding of the button in `QDarkStyleSheet` is not `3 px`  but `2 px`.
- Second, the workaround was applied to the browse button of the `BaseTabs` class even when not in dark mode, causing a distortion of the aspect ratio of this button. This also caused this button to be not positioned the same way as the other tab corner widgets as noticed in PR https://github.com/spyder-ide/spyder/pull/9452#issuecomment-497460070.

image

- Third, the hard coded padding added to the stop and reset button for the IPython plugin were not needed and ignored by Qt anyway.

Here is a screenshot of Spyder after the changes running the latest version of `QDarkStyleSheet` on the `develop` branch :

image

### Issue(s) Resolved

Fixes what was reported in PR https://github.com/spyder-ide/spyder/pull/9452#issuecomment-497460070

Note: these changes were initially part of PR https://github.com/spyder-ide/spyder/pull/9452 (see commit https://github.com/spyder-ide/spyder/commit/486534cb34cf759f9a6571f8f5c2012ed8611910) but I think it is better that they are considered and reviewed in a separate PR.

### Affirmation


goanpeca:
Awesome @jnsebgosselin, looks good to me!

goanpeca:
@jnsebgosselin do you think is worth moving the actual style to a centralized place and import it as needed?

Would this make the code more/less readable?

jnsebgosselin:
> 
> 
> @jnsebgosselin do you think is worth moving the actual style to a centralized place and import it as needed?
> 
> Would this make the code more/less readable?

I do not understand what you mean by that, sorry.

goanpeca:
I meant:

```python
STYLE_BUTTON_CSS = "QToolButton::menu-indicator{image: none;}"

...

from somewhere import STYLE_BUTTON_CSS
```

jnsebgosselin:
> 
> 
> I meant:
> 
> ```python
> STYLE_BUTTON_CSS = "QToolButton::menu-indicator{image: none;}"
> 
> ...
> 
> from somewhere import STYLE_BUTTON_CSS
> ```

Ah yes, this is a very good idea! I will do that in that PR.

jnsebgosselin:
Also, I don't know, but maybe you will want to wait for the fix in `QDarkStyleSheet` to be merged in master before merging this PR.

goanpeca:
Sure @jnsebgosselin no worries :-)

ccordoba12:
Thanks for this @jnsebgosselin! You're always fixing my half-baked hacks!

dpizetta:
Some suggestions below:

- Separator in 'fake' toolbars: As you are creating the toolbars as widgets and using VLine as separators, the toolbars separators are 16px height in qdark. To create a similarity you can use the VLine with 60-80% height for all themes or keep them fixed at 16px.

- I notice some problems in headers from the different dock. See the comparison between "files" and "variable profile". Padding or margins seems to be different, left and right, and top and bottom. Maybe some hard-coded values. Also, the "Name" column could be expanded at the start.

diff-space-header-profile-table

- Icons out of center: I don't know if it is intended, but it seems they are not in the center.
icons-image-out-off-center

If you would like I open new issues, please, let me know. Tks

jnsebgosselin:
> If you would like I open new issues, please, let me know. Tks

I think this should go in another issue because this doesn't seem to be related to the topic of this PR...

ccordoba12:
I think we should increase our minimal supported version of QDarkStyle for this, right @jnsebgosselin?

jnsebgosselin:
> 
> 
> I think we should increase our minimal supported version of QDarkStyle for this, right @jnsebgosselin?

Yes, you are right.  So it needs to be updated in :

- setup.py
- environment.yml
- conda.txt

Anywhere else?

ccordoba12:
> Anywhere else?

https://github.com/spyder-ide/spyder/blob/master/spyder/dependencies.py

Thanks!

ccordoba12:
Is this one ready @jnsebgosselin?

jnsebgosselin:
Yes it is ready.

PR #9571, opened 2019-06-14
PR: Fix error when launching Preferences in Chinese

### LINKED ISSUE #9390: Error when launching Preferences in Chinese

1. Download Spyder 4.0.0b2 from conda
2. Start Spyder
3. Launch the preference window and then an error report window shows. However, the preference window seems working correctly.

### Traceback
```python-traceback
    widget = plugin.create_configwidget(dlg)
    configwidget.initialize()
    self.setup_page()
    "separate window.") % (inline, automatic))
TypeError: not all arguments converted during string formatting
```

## Versions

* Spyder version: 4.0.0b2 
* Python version: 3.6.7
* Qt version: 5.9.7
* PyQt5 version: 5.9.2

### Dependencies

```
qdarkstyle >=2.6.4 :  2.6.8 (OK)
sphinx >=0.6.6     :  2.0.1 (OK)
nbconvert >=4.0    :  5.5.0 (OK)
numpy >=1.7        :  1.16.3 (OK)
sympy >=0.7.3      :  1.4 (OK)
cython >=0.21      :  0.29.7 (OK)
qtconsole >=4.2.0  :  4.4.4 (OK)
IPython >=4.0      :  7.5.0 (OK)
```

steff456:
## Description of Changes

Fix the error than appear when trying to run preferences when spyder is in chinese.

### Issue(s) Resolved

Fixes #9390

### Affirmation


spyder/plugins/ipythonconsole/confpage.py
```diff
@@ -113,10 +113,10 @@ def setup_page(self):
         backend_group = QGroupBox(_("Graphics backend"))
```
ccordoba12: @steff456, please revert the changes you made in this file because they are not necessary to solve this issue.

Besides, it would render the texts you changed to fallback to English because there would be a mismatch between the strings in our code and the translated ones.

steff456: In fact, by changing only the other file the error persists.

When I debug the error, the texts existed but for some reason by the string formatting the error was present. So I tried a different formatting and the bug was fixed.

ccordoba12: This is no fix at all because (as I told you above) by changing the string (in this case, by using `format` instead of classical string replacement), you're making the current Chinese translation of that string outdated, which makes it fallback to English:

Selección_007

So, please revert these changes and fix the other errors that appear.

spyder/locale/zh_CN/LC_MESSAGES/spyder.po
```diff
@@ -2422,8 +2422,8 @@ msgid ""
 "please select <b>%s</b> to put graphics inside the console or <b>%s</b> to "
 "interact with them (through zooming and panning) in a separate window."
 msgstr ""
-"确定如何在控制台中显示图形。 如果不确定，请选择<b>%s</b>将图形放入控制台或<b>"
-"％s</b>以在单独的窗口中与它们进行交互（如缩放和平移）。"
+"确定如何在控制台中显示图形。 如果不确定，请选择<b>%s</b>将图形放入控制台或<b>％s</b>"
+"以在单独的窗口中与它们进行交互（如缩放和平移）。"
```
ccordoba12: @steff456, I found the error. The real reason for this is that the symbol used by the translator on the second replacement (`％`) is not a regular percentage symbol (`%`) (I don't know why).

Therefore, please leave this string as it was and simply change that symbol. After doing that, you'll see more errors like this one to appear too, so please fix them too.

Something I forgot to tell you is that you need to use a translation program (like `poedit`) to update the translation (i.e. to generate a new `.mo` file) after making changes to source translation file (i.e. the `.po` file).

So in this PR you need to submit both the new `po` and mo files.

ccordoba12: In other words, you need to open this file in Poedit, make your changes and save it. That will automatically generate a new `mo` file for you.

ccordoba12:
@steff456, did you make sure that now our Preferences open correctly in Chinese?

steff456:
> @steff456, did you make sure that now our Preferences open correctly in Chinese?

Yes! Here is a screenshot of one of the troubled sentences

image

ccordoba12:
Great! Thanks a lot @steff456!

Issue #123, opened 2015-02-16
Enhancement: VI keyboard personality


spyder-issues-migrator:
_From [jfburkh...@gmail.com](https://code.google.com/u/114996126900533963350/) on 2009-11-19T16:21:00Z_

Hello,

I'm trying to understand how to enable a 'vi keyboard' personality. Is it
possible in Spyder?

I've searched google and the 'issues' here, but I came up with nothing.

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=123_

spyder-issues-migrator:
_From [pierre.raybaut](https://code.google.com/u/pierre.raybaut/) on 2010-01-03T11:34:34Z_

**Status:** Duplicate  
  **Mergedinto:** 73

PR #9572, opened 2019-06-14
PR: URL awareness improvements

### LINKED ISSUE #9513: Improvements to URL awareness
* Make URLs work with `%20`.
* Remove the requirement of `mailto:`
* Make `{gh/bb/gl}-123` work.

goanpeca:
## Description of Changes

- [x] Make URLs work with %20.
- [x] Remove the requirement of mailto:
- [x] Make `{gh/bb/gl}-123` work.

### Issue(s) Resolved

Fixes #9513

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -20,13 +20,13 @@ Editor widget based on QtGui.QPlainTextEdit
 # Standard library imports
 from __future__ import division, print_function
 
+from unicodedata import category
 import logging
 import os.path as osp
 import re
 import sre_constants
 import sys
 import time
-from unicodedata import category
 
 # Third party imports
 from diff_match_patch import diff_match_patch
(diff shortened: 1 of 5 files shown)


goanpeca:
@rgommers for the shorthand notation

`gh-123` this will only work if we are cloning from the actual repo and not from the fork (like many projects follow) 

So I can ask git for this
`$ git config --get remote.origin.url`

And use the output to get the repo, but if the repo is a fork we may or may not have the `upstream` remote set. Thoughts no this?

rgommers:
Oh yeah, having to check for the parent repo via the GitHub API would work but that's a weird design. The standard convention is that the parent is named `upstream`, but it's not always the case. I think the choice is between using `upstream` if it exists, or just forgetting about this.

ccordoba12:
> The standard convention is that the parent is named upstream, but it's not always the case. I think the choice is between using upstream if it exists, or just forgetting about this.

I agree with this. Let's try to detect `upstream` using git locally and then use it for as the main URL for all `{gh/bb/gl}-123` kind of references.

goanpeca:
The following dialog is displayed if `gh-123` is clicked on a file that is not being tracked by git or if there is no upstream, or no origin

ccordoba12:
> The following dialog is displayed if gh-123 is clicked on a file that is not being tracked by git or if there is no upstream, or no origin

I agree with ti, thanks for adding it.

PR #9585, opened 2019-06-15
PR: Wrap text in tooltips, hints and calltips

### LINKED ISSUE #9543: Wrap text in warnings and hover tips
### Issue Report Checklist

## Problem Description
Hover tips and warnings extend beyond Spyder screen

dalthviz:
## Description of Changes

Wrap text of the warnings and hover tips
### Elements
* [x] Calltips

#### Editor
editor

#### Console
console

* [x] Tooltips

k
kk

* [x] Hints
Width 120
hint

* [x] Improve handling of signature and documentation in the IPython Console
 
console_call

### Issue(s) Resolved

Fixes #9543 

### Affirmation


spyder/widgets/mixins.py
```diff
@@ -142,7 +144,7 @@ def _get_inspect_shortcut(self):
 
     def _format_text(self, title=None, signature=None, text=None,
                      inspect_word=None, title_color=None, max_lines=None,
-                     display_link=False):
+                     width=None, display_link=False):
```
ccordoba12: It makes no sense to make `width=None` because `textwrap.wrap` won't accept it as a parameter. So let's give it a default value of `60`.

spyder/widgets/mixins.py
```diff
@@ -197,12 +199,15 @@ def _format_text(self, title=None, signature=None, text=None,
         if not text:
             text = '\n<i>No documentation available</i>\n'
```
ccordoba12: I think it makes no sense to add the last `\n` here because it adds a blank line for no reason, right @goanpeca?

So I'd say we should remove it.

spyder/widgets/mixins.py
```diff
@@ -496,14 +510,16 @@ def show_tooltip(self, title=None, signature=None, text=None,
...
         )
 
         self._update_stylesheet(self.tooltip_widget)
 
         # Display tooltip
         self.tooltip_widget.show_tip(point, tiptext, cursor=cursor)
 
-    def show_hint(self, text, inspect_word, at_point):
+    def show_hint(self, text, inspect_word, at_point,
+                  max_lines=_DEFAULT_MAX_LINES, width=_DEFAULT_WIDTH):
```
ccordoba12: Please put `width=_DEFAULT_WIDTH` in a new line below this one.

ccordoba12:
@dalthviz, did you try this with the Java LPS server to see how things look there?

spyder/widgets/mixins.py
```diff
@@ -517,7 +533,8 @@ def show_hint(self, text, inspect_word, at_point):
 
         self.show_tooltip(signature=html_signature, text=extra_text,
                           at_point=point, inspect_word=inspect_word,
-                          display_link=True, max_lines=10, cursor=cursor)
+                          display_link=True, max_lines=max_lines, width=width,
```
ccordoba12: Perhaps in the case of tooltips we should show more lines (30?) and make the wrapping wider (120?) to not lose the important information they give to the user.

But please test things with other LSPs to see how to adjust that.

dalthviz:
@ccordoba12 I have been asking @andfoy about what server to launch and how to do it, although I have not test it yet with any server besides the Python one

spyder/widgets/mixins.py
```diff
@@ -318,7 +327,8 @@ def handle_sub(matchobj):
             formatted_lines = []
             name = signature.split('(')[0]
             indent = ' ' * (len(name) + 1)
-            rows = textwrap.wrap(signature, width=60, subsequent_indent=indent)
+            rows = textwrap.wrap(signature, width=width,
+                                 subsequent_indent=indent)
```
goanpeca: We might want to have a different width for the signatures, it might look better to have the wrapping for the signature smaller than the wrapping for the docstring.

Thoughts @ccordoba12 and @dalthviz ?

spyder/widgets/mixins.py
```diff
@@ -142,7 +144,8 @@ def _get_inspect_shortcut(self):
 
     def _format_text(self, title=None, signature=None, text=None,
                      inspect_word=None, title_color=None, max_lines=None,
-                     display_link=False):
+                     width=_DEFAULT_WIDTH, display_link=False,
```
goanpeca: Maybe  `max_width` is a better name as it is more clear what it implies (same as `max_lines`), it could be less than that but never larger. What do you think @dalthviz ?

spyder/widgets/mixins.py
```diff
@@ -263,6 +271,7 @@ def _format_text(self, title=None, signature=None, text=None,
         return template
 
     def _format_signature(self, signatures, parameter=None,
+                          width=_DEFAULT_WIDTH,
```
goanpeca: Maybe  `max_width` is a better name as it is more clear what it implies (same as `max_lines`), it could be less than that but never larger. What do you think @dalthviz ?

spyder/widgets/mixins.py
```diff
@@ -356,6 +366,7 @@ def handle_sub(matchobj):
         return '<br><br>'.join(new_signatures)
 
     def _check_signature_and_format(self, signature_or_text, parameter=None,
+                                    width=_DEFAULT_WIDTH,
```
goanpeca: Maybe  `max_width` is a better name as it is more clear what it implies (same as `max_lines`), it could be less than that but never larger. What do you think @dalthviz ?

spyder/widgets/mixins.py
```diff
@@ -496,14 +514,17 @@ def show_tooltip(self, title=None, signature=None, text=None,
...
 
         self._update_stylesheet(self.tooltip_widget)
 
         # Display tooltip
         self.tooltip_widget.show_tip(point, tiptext, cursor=cursor)
 
-    def show_hint(self, text, inspect_word, at_point):
+    def show_hint(self, text, inspect_word, at_point,
+                  max_lines=30,
+                  width=120):
```
goanpeca: Any particular reason for doing

```python
    def show_hint(self, text, inspect_word, at_point,
                  max_lines=30,
                  width=120)
```

as opposed to 

```
    def show_hint(self, text, inspect_word, at_point,
                  max_lines=30, width=120)
```
?

spyder/widgets/mixins.py
dalthviz: Yep :+1:

spyder/widgets/mixins.py
dalthviz: I was testing what @ccordoba12  commented [here](https://github.com/spyder-ide/spyder/pull/9585#discussion_r294055120) and before that I probably have there the constants `_DEFAULT_MAX_LINES` and `_DEFAULT_WIDTH` that take more space than the numbers (and I only replaced the values without reformating the params into only one line).

spyder/widgets/mixins.py
dalthviz: Having the width of the signatures smaller than the one for the documentation looks something like this:

diff

Having it the same looks something like this:

same

dalthviz: I kind like it better with the signature and documentation having the same width but it's not much of a difference for me

spyder/plugins/ipythonconsole/widgets/help.py
```diff
@@ -44,21 +44,39 @@ def get_documentation(self, content):
         if text:
```
ccordoba12: Please put all the code in this `if` inside another `if`, like this:

```python
if text:
    if self.language_name is not None and self.language_name == 'python':
        text = re.compile(ANSI_PATTERN).sub('', text)
        ...
    else:
        return text
else:
    return ''
```

spyder/plugins/ipythonconsole/widgets/help.py
```diff
@@ -44,21 +44,39 @@ def get_documentation(self, content):
...
+            signature = name + argspec
+        else:
+            signature = getsignaturefromtext(text, name)
+        return signature
+
     def get_signature(self, content):
         """Get signature from inspect reply content"""
         data = content.get('data', {})
         text = data.get('text/plain', '')
         if text:
```
ccordoba12: Also put all the code in this `if` inside another `if`, like this:

```python
if text:
    if self.language_name is not None and self.language_name == 'python':
        self._control.current_prompt_pos = self._prompt_pos
        ...
    else:
        return ''
else:
    return ''
```

spyder/plugins/ipythonconsole/widgets/help.py
```diff
@@ -69,18 +87,25 @@ def get_signature(self, content):
             except:
```
ccordoba12: Let's change this to `except Exception:`

dalthviz:
@ccordoba12 a preview with `Julia`:

imagen

ccordoba12:
Things look pretty good with Julia, thanks @dalthviz!

One small detail though: in your screenshots there's a blank line at the beginning of the calltip that shouldn't be there. Please see how to remove it.

dalthviz:
A new preview with `Julia`:
imagen

spyder/plugins/ipythonconsole/widgets/help.py
```diff
@@ -183,9 +183,12 @@ def _handle_inspect_reply(self, rep):
             if content.get('status') == 'ok' and content.get('found', False):
                 signature = self.get_signature(content)
                 documentation = self.get_documentation(content)
+                documentation_new_line = (self.language_name is not None
```
ccordoba12: Let's call this variable `new_line` instead of `documentation_new_line`.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -2002,6 +2003,11 @@ def show_code_analysis_results(self, line_number, block_data):
             msg = msg.strip()
             # Avoid messing TODO, FIXME
             msg = msg[0].upper() + msg[1:]
+            msg = textwrap.wrap(msg, width=60)
```
ccordoba12: CodeEditor inherits from TextEditBaseWidget, which in turn inherits from BaseEditMixin. Therefore you can use here `_DEFAULT_MAX_WIDTH` instead of a fixed width of 60.

PR #9595, opened 2019-06-17
PR: Send fallback completions to the end of the completion widget

andfoy:
## Description of Changes
A character 'z' is prepended to the sortText response of the ``FallbackActor``, such that LSP completions are presented first, and then the fallback ones. If LSP is not available, then the fallback ones are displayed sorted.

ezgif com-gif-maker

### Issue(s) Resolved

Fixes #

### Affirmation

@andfoy


ccordoba12:
Is it `z` enough? Or should we use `zz` to be totally sure?

andfoy:
> Is it z enough? Or should we use zz to be totally sure?

We could use zz

ccordoba12:
> We could use zz

Ok, then I'd prefer that to not mix fallback completions with possible LSP completions that start by `z`.

PR #9632, opened 2019-06-21
PR: Save project recent files as relative paths in workspace.ini

### LINKED ISSUE #5491: Relative paths in workspace.ini to make projects moveable

1. Add a new project `Projects > New project...`
2. Open some files.
3. Close Spyder.
4. Copy the whole project's directory to another computer to a different path (or have it synced with other means to different directories, e.g. `/home/mm/Software/Project` and `/home/MickeyMouse/Work/Software/Project` for your home and work computers).
5. Open the project on the other computer.

I would like to see the same files opened automatically on both computers when opening the project, but the `.spyproject/workspace.ini` file stores the full paths in `[main]recent_files`. If the paths were stored relative to `.spyproject/..` (the project's base directory), it would be possible to move the project to a different location.

There seems to be two related bugs: #1133 and #1343. Maybe more?

* Spyder Version: 3.2.3
* Python Version: 3.5.2 64bits
* Qt Version: 5.5.1
* PyQt Version: 5.5.1
* Operating system: Linux Mint 18.2 Cinnamon 64-bit

## Dependencies

and paste the contents below:

psutil >=0.3     :  5.3.1 (OK)
sympy >=0.7.3    :  1.0 (OK)

Keou0007:
## Description of Changes

* When closing a project, recent_files are written to workspace.ini relative to project root_path
* When opening a project, check if recent_files paths are absolute and if not, join with project root_path to make absolute before opening.

This allows better portability of projects. Project directories can be moved, or stored in different locations on different systems, and opening the project will still successfully open all the recent files into tabs.

In my testing, the second change listed above is not necessary, presumably because spyder changes the active working directory to the project root_path before opening recent files, so both absolute and relative paths will always work. Adding this line is probably safer though, with the abspath checking necessary for backward compatibility.

### Issue(s) Resolved

Fixes #5491

### Affirmation

I certify the above statement is true and correct: Shannon Keough


ccordoba12:
It seems our tests assume files are saved with absolute paths, so you need to fix that to make them pass.

Keou0007:
You might need to help me with that. I can't figure out how the test works.

Keou0007:
Ok, I'm pretty sure I figured out how the tests in the projects plugin work. I've added a small change to the relevant tests so they now pass on my system how they should. But...

From my understanding of how it works. The test was broken anyway because it uses a mocker function that was returning relative paths, while the function being mocked (main.editor.get_open_filenames) returns absolute paths. I have fixed this and submitted a seperate PR #9648

It seems to me that these tests really shouldn't be using mockers at all because it can cause the tests to produce false passes when they should probably be failing. In this case the tests aren't actually testing that spyder works as intended, but are instead only testing that the mocker function returns the values that it is hardcoded to return. In this case, this caused test failures for me (and confused me alot) because I was editing the paths in between the mocker function return and the assertion statement. My change to the plugin worked as intended when incorporated into a running version of spyder, but failed the test because the test was testing the wrong thing.

I think a much much better option is to actually create a few files, open them, use the real main.editor.get_open_filenames function to get those file paths, and then use the real os.path.isfile to check that the files exist. Then the test is actually testing the program, not just testing mocker functions.

It seems the whole point of using the mocker functions in the first place is to avoid having to create files on disk, the code says exactly that in the comments, but this is pointless because the tests create a temporary project directory on the disk anyway. This temporary project contains a .spyproject folder with all the expected files inside, as is necessary for the test. Since the test is already creating all of these, why not just create a few empty test files inside the temporary project as well?

jitseniesen:
> It seems to me that these tests really shouldn't be using mockers at all because it can cause the tests to produce false passes when they should probably be failing.

These tests are (meant to be) unit tests, which test a unit (here, the project plugin) in isolation. This means that other units (here, the editor plugin, and also the file system) are mocked. The idea is that these tests should run fast so that we can write a lot of them, and that when they fail it is clear where the fault is.

The disadvantage of unit tests is what you say, there is a danger that the mock does not behave like the editor plugin does, so ideally there are also integration tests that use both the editor and project plugins. These integration tests usually test things a user can actually see, e.g. that if a project is opened, some files are opened, and the project is closed, the test checks that the opened files are closed, and also that they are opened again when user opens the project again.

At least, that is my understanding of how tests ideally should work. In practice, the separation between unit and integration tests is not that clear.

Keou0007:
Thanks for the explanation @jitseniesen. That makes more sense to me now. I have very little experience with unit testing and I've never come across mocker functions before!

PR #9648, opened 2019-06-23
PR: Fix test_set_get_project_filenames* functions so the test matches actual spyder behaviour

Keou0007:
## Description of Changes

* Changed the arbitrarily defined lists of test files to be absolute paths rather than relative

These tests use a mocker function to return files open in the editor without actually having to create and open any files. The problem is that the mocker function was returning a list of relative paths, while the actual function (main.editor.get_open_filenames) returns a list of absolute paths.

This change means the mocker function in the test is correctly replicating the actual function now, while still achieving the goal of not having to create and open files during the test.

### Affirmation

I certify the above statement is true and correct: Shannon Keough

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -1006,6 +1006,50 @@ Files covered:
 spyder/plugins/variableexplorer/widgets/dataframeeditor.py
 
 
+-------------------------------------------------------------------------------
+
+
+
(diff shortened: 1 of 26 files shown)


ccordoba12:
@Keou0007, please rebase of merge with master to fix the errors in our tests.

Issue #9420, opened 2019-05-24
FileNotFoundError when closing a file that's opened in several Spyder instances


impact27:
### Issue Report Checklist

* [ ] Searched the [issues page](https://github.com/spyder-ide/spyder/issues?q=is%3Aissue) for similar reports
* [ ] Read the relevant sections of the [Spyder Troubleshooting Guide](https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ) and followed its advice
* [ ] Reproduced the issue after updating with ``conda update spyder`` (or ``pip``, if not using Anaconda)
* [ ] Could not reproduce inside ``jupyter qtconsole`` (if console-related)
* [ ] Tried basic troubleshooting (if a bug/error)
    * [ ] Restarted Spyder
    * [ ] Reset preferences with ``spyder --reset``
    * [ ] Reinstalled the latest version of [Anaconda](https://www.anaconda.com/download/)
    * [ ] Tried the other applicable steps from the Troubleshooting Guide
* [ ] Completed the **Problem Description**, **Steps to Reproduce** and **Version** sections below

## Problem Description
Error when closing a tab whose autosave file was destroyed.

### What steps reproduce the problem?

1. 
2. 
3. 

### Paste Traceback/Error Below (if applicable)

```python-traceback
ERROR:spyder.plugins.editor.widgets.autosaveerror:<class 'FileNotFoundError'> while removing autosave file /Users/quentinpeter/.spyder-py3-dev/autosave/spydercommbase.py
Traceback (most recent call last):
  File "/Users/quentinpeter/pyscripts/spyder/spyder/plugins/editor/utils/autosave.py", line 165, in remove_autosave_file
    os.remove(autosave_filename)
FileNotFoundError: [Errno 2] No such file or directory: '/Users/quentinpeter/.spyder-py3-dev/autosave/spydercommbase.py'
```

## Versions

* Spyder version: 
* Python version: 
* Qt version: 
* PyQt version: 
* Operating System name/version: 

### Dependencies

```
PASTE DEPENDENCIES HERE
```

ccordoba12:
Could you describe when or how this problem happens?

impact27:
This happens when the autosave file is deleted by an other program. (I am debugging spyder with spyder). The problem is that I have the window telling me the file was deleted, as I expect, but also the internal exception window.

ccordoba12:
[...]

PR #9660, opened 2019-06-23
PR: Disable autosave if not running in single instance mode

jitseniesen:
This is a temporary stop gap to work around bugs in the autosave system like #9420 when multiple Spyder instances are running simultaneously. Once a more permanent fix is developed, this can be reverted.

<s>I did not include a test, because the code is already run by existing tests and it seems a bit too much effort to write a test for what is meant to be a temporary change, </s>

### Issue(s) Resolved

This does not truly resolve the issue, so no "Fixes #..."

### Affirmation

I certify the above statement is true and correct: Jitse Niesen

### DIFF
--- a/spyder/plugins/editor/tests/conftest.py
+++ b/spyder/plugins/editor/tests/conftest.py
@@ -48,6 +48,8 @@ def editor_plugin(qtbot, monkeypatch):
                 projects = Mock()
                 projects.get_active_project.return_value = None
                 return projects
+            elif attr == 'new_instance':
+                return False  # otherwise autosave is disabled
             else:
                 return Mock()
 
--- a/spyder/plugins/editor/tests/conftest.py
+++ b/spyder/plugins/editor/tests/conftest.py
@@ -120,3 +120,19 @@ def editor_plugin_open_files(request, editor_plugin, python_files):
         return editor, expected_filenames, expected_current_filename
 
     return _get_editor_open_files
+
+
+@pytest.fixture
+def conf(request):
+    """
+    Fixture yielding the Spyder config and resetting it after the test.
+
+    You can use conf.set() to change the Spyder config in your test function.
+    If you want to change the config at setup time, then use
+    @pytest.mark.parametrize('conf', [(section, label, value)], indirect=True)
+    in front of your test function
+    """
+    if hasattr(request, 'param'):
+        CONF.set(*request.param)
+    yield CONF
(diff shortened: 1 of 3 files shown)


jitseniesen:
To clarify, I am submitting a temporary fix so that the issue does not block the release of v4.0 beta 3.

jitseniesen:
Don't be lazy, Jitse, just add a test!

jitseniesen:
As I understand it, the test failures are due to on-going work on the Object Explorer and not related to this PR.

ccordoba12:
Yes, that's right. I'm working right now to finish PR #8852 so our tests pass again. I'll let you know when that's done.

ccordoba12:
Please merge with master to get a new set of fixes for our tests.

jitseniesen:
The test `test_get_hints[params4]` in `spyder/plugins/editor/widgets/tests/test_hints_and_calltips.py` failed consistently, but only on Python 3.6 in Windows on Azure. I think it is unlikely to be related to this PR, so I don't intend to investigate it further.

ccordoba12:
That's fine @jitseniesen, don't worry about it. Unfortunately those tests are very flaky.

jitseniesen:
It is not only `remove_autosave_files` that causes issues. PR #10077 is a permanent fix for the issue.

ccordoba12:
> This might be excusable if it was only in the Github builds for a short time, but this has made it into not one but two public betas so far

Betas are betas, i.e. they are not expected to be stable nor the final version. So I don't agree with making the extra effort of informing users that autosaving doesn't work with multiple instances when that's going to be fixed before the final release.

ccordoba12:
> However, the 27 629 total downloads of Spyder from the spyder-ide beta channel would seem to suggest otherwise.

Compared to the ~200k from Anaconda + ~40K from PyPi downloads per month, that's certainly negligible. And I'm sorry if those users feel frustrated but that's the chance they take by using a beta.

And again, betas are really for experimentation without disrupting our entire user base. So please don't insist on us working to make our betas production ready.

> Furthermore, unless doing so would take more than a few hours to implement, it would already would have been worthwhile in terms of avoiding wasted time, effort and frustration this cost one single user (myself) out of at least thousands

If you felt so concerned about this, I don't understand why you didn't take the time to add this yourself instead of asking us to do it.

Final note: I'm locking this PR because it makes no sense to keep discussing about this.

PR #9673, opened 2019-06-25
PR: Add support to run files in external system terminal for macOS

### LINKED ISSUE #5062: Run in external system terminal doesn't work in macOS
## Description

1. open python file
2. "run"->"configuration per file.."
3. set "execute in an external system terminal"
4. "run"
5. nothing happens

I expected Terminal app appearance, but terminal doesn't load

* Spyder Version: 3.2.1 
* Python Version: 3.6.1 and 2.7
* Qt Versions: 5.6.2, PyQt5 5.6 on Darwin

## Dependencies
```
sympy >=0.7.3    :  1.0 (OK)

```

Best regards,
Andrew

steff456:
## Description of Changes

Add the option for running programs in an external system terminal in macOS

image

### Issue(s) Resolved

Fixes #5062

### Affirmation


ccordoba12:
@steff456, please rebase in top of master to get the fix for our tests. Also, instead of using `appscript`, I think it should be possible to use `open Terminal.app` instead.

@goanpeca, could you take a look at @steff456's work here? Thanks!

spyder/utils/programs.py
```diff
@@ -340,7 +340,13 @@ def run_python_script_in_terminal(fname, wdir, args, interact,
...
                     run_program(program['cmd'], arglist)
                 return
-        # TODO: Add a fallback to OSX
+    elif sys.platform == 'darwin':
+        import appscript
+        command = ""
+        if wdir:
+            command += "cd {0}\n".format(wdir)
+        command += "{0} {1}".format(p_args[0], p_args[1])
+        appscript.app('Terminal').do_script(command)
```
goanpeca: As discussed in our call earlier @steff456 

```
python
        fpath = os.path.join(TEMP_PATH, 'a.tool')
        with open(fpath, 'w') as f:
            f.write(command)
        os.chmod(path, 0o777)
        subprocess.call(['open ' + fpath], shell=True)
```

spyder/utils/programs.py
```diff
@@ -710,7 +712,27 @@ def run_python_script_in_terminal(fname, wdir, args, interact,
...
+    elif sys.platform == 'darwin':
+        f = tempfile.NamedTemporaryFile('wt', suffix='.sh',
+                                        dir=get_temp_dir(), delete=False)
+        f.write(' '.join(p_args))
+        f.close()
+        os.chmod(f.name, 0o777)
+
+        def run_terminal_thread():
+            if wdir:
+                proc = subprocess.Popen(
```
ccordoba12: I think we should use here the function `run_shell_command` (present in this same module), instead of `subprocess` directly. Please try that to see if it works.

spyder/utils/programs.py
```diff
@@ -710,7 +712,27 @@ def run_python_script_in_terminal(fname, wdir, args, interact,
...
+                    'open -a Terminal.app ' + f.name, cwd=wdir, shell=True)
+            else:
+                proc = subprocess.Popen(
+                    'open -a Terminal.app ' + f.name, shell=True)
+            # Prevent race condition
+            time.sleep(3)
```
ccordoba12: Why do we have to wait for so long here? Could we reduce this time?

Problem is if an operation takes more than 300ms users will think something is working incorrectly or not working at all (that's something backed up by research).

spyder/utils/programs.py
```diff
@@ -710,7 +712,27 @@ def run_python_script_in_terminal(fname, wdir, args, interact,
...
+            proc.wait()
+            os.remove(f.name)
+
+        thread = threading.Thread(target=run_terminal_thread)
```
ccordoba12: Why not use QThread instead of threading here? Just asking, I don't know if it makes any difference or not.

spyder/utils/programs.py
steff456: This time is necessary to prevent a race condition from the call of the subprocess and the removal of the temporal file.

spyder/utils/programs.py
steff456: There's no difference between them in this case

goanpeca: I guess it is better to use a QThread to keep things more in sync with the rest of the application?

spyder/utils/programs.py
```diff
@@ -710,7 +712,27 @@ def run_python_script_in_terminal(fname, wdir, args, interact,
...
+        def run_terminal_thread():
+            if wdir:
+                proc = run_shell_command('open -a Terminal.app ' + f.name,
+                                         cwd=wdir)
+            else:
+                proc = run_shell_command('open -a Terminal.app ' + f.name)
+            # Prevent race condition
+            time.sleep(2)
+            proc.wait()
+            os.remove(f.name)
```
goanpeca: Would it be better to then not remove these files?

They will be stored in a temp folder so the user can delete this folder any time it is needed (which for these small files will take a very very long time anyway)

I agree with @ccordoba12 that 2 seconds is just too much.

steff456: The problem is that the folder for these temporal files is stored in a place not that easy to track for an user, and every time they run the program from the external console a file will be created. At the end, a lot of files will remain on the system. 

If you try it, that time will delay the deletion of the file rather than waiting 2 seconds to open the terminal.

goanpeca: Is this blocking the UI? or the 2 second wait is on a thread?

steff456: Is on the thread, the UI isn't blocked

goanpeca: Ok then all good :-) !

goanpeca: But I would like to have some tests added to this :-)

The "script" could be the creation of a file with some content so that we can check that happens after some time.

steff456: I'm not sure of how to create the test, because there's not a way to know if the terminal is open or not.

goanpeca: @steff456 I suggest we run a script that does some operation that ca be checked not directly check if a terminal is open or not, so:

```python
def run_python_script_in_terminal(fname, wdir, args, interact,
                                  debug, python_args, executable=None):
```

where:

```python
# fname
with open('/Users/gpena-castellanos/Desktop/somefile.txt', 'w') as f:
    f.write('This is a test')
```

And we check that the file is created and that the contents match.

goanpeca:
@steff456 also when testing this, the title of the terminal window is
Screen Shot 2019-07-03 at 17 56 28

Maybe we could use the name of the file instead of `temp` name? like:

`run_spyder_<name-of-the-file.py>.sh`

Thoughts @ccordoba12 ?

I think it would make it clearer for users

goanpeca:
>Please increase the timeout in that line to 3000 because we're waiting for 2 seconds in macOS before running a file.

Actually we are waiting for 2 seconds before erasing the sh file, the command is run right away.

ccordoba12:
Right, sorry. Then I don't why those tests are failing.

goanpeca:
Yeiiii its working :-)

spyder/utils/programs.py
```diff
@@ -710,7 +712,25 @@ def run_python_script_in_terminal(fname, wdir, args, interact,
                 else:
                     run_program(program['cmd'], arglist)
                 return
-        # TODO: Add a fallback to OSX
+    elif sys.platform == 'darwin':
+        f = tempfile.NamedTemporaryFile('wt', prefix='run_spyder_',
+                                        suffix='.sh', dir=get_temp_dir(),
+                                        delete=False)
+        if wdir:
+            f.write('cd {}\n'.format(wdir))
```
goanpeca: Great work @steff456!  

Lets add a test to check for spaces in paths

```
wdir = '/users/tefa/Dekstop/carpeta con espacio/'
wdir = '/users/tefa/Dekstop/carpeta con espacio/archivo con espacio'
# It may fail here
'cd {}'.format(wdir)
```

steff456: I should create a new one or modify the existing one?

goanpeca: Lets try to create a parametrized test?

like done here https://github.com/spyder-ide/spyder/blob/master/spyder/plugins/editor/widgets/tests/test_goto.py#L31 ?

PR #9682, opened 2019-06-25
PR: Fix QComboBox with newer Qt versions and qdarkstyle

goanpeca:
## Description of Changes

## Before
Screen Shot 2019-06-25 at 22 28 24

## After
Screen Shot 2019-06-25 at 22 27 42

### Issue(s) Resolved

Fixes ColinDuquesnoy/QDarkStyleSheet#191

### Affirmation


goanpeca:
@dpizetta, this fix took longer than expected but it will be something extra to add to the known issues page of QDarkStyle. I will add the code snippet as well.

ccordoba12:
This is causing some serious crashes and errors in our tests. Perhaps we should deactivate the style proxy when `running_under_pytest`?

dpizetta:
Very nice! Tks for the support @goanpeca ! Nice, you could manage to keep the list below the combo box current value.  I've noticed that when it is in the editable mode those problems (the margin and the list position) problem does not occur.  See above. Please, check your code against your example to see if it is ok too. Sorry for the big pictures here.

goanpeca:
>I've noticed that when it is in the editable mode those problems (the margin and the list position) problem does not occur. 

Indeed as it is an editable thing, the widgets underneath are different. (no problem about the large pics :-p)

I added the "fix" to the example and the result is:

Screen Shot 2019-06-26 at 09 36 59
Screen Shot 2019-06-26 at 09 36 49

Opening the PR now. This should go to develop right?

---

Also, we need to intermix some items with icons and some without icons on the same combobox, on the `example.py` so we can see if any problems arrive there

Since that is done with QTDesigner, could you add that? (I can open an issue over QDark)

goanpeca:
This fix might be making Spyder unstable under some specific circumstances. For instance changing theme from light to dark was resulting in a `Seg Fault 11`.

ccordoba12:
> For instance changing theme from light to dark was resulting in a Seg Fault 11.

This is really bad! What do you think we should do about it?

goanpeca:
>This is really bad! What do you think we should do about it?

I am not sure at this moment, this was the only "solution" that actually solved the issue. The only other way of handling this (what I did with Navigator) is to have a custom popup widget, which is basically a `QListWidget` that replaces what the ComboBox uses. This would solve all issues that the ComboBox is currently having.

I could add this "widget" to QDarkStyle or add it on spyder so we can replace the QComboBox, by a SpyderComboBox.

Thoughts @dpizetta ?

dpizetta:
> Thoughts @dpizetta , @ccordoba12 ?

Not sure either.  I think adding a widget in substitution on for qcombobox in qdark is interesting if people are having problems with the solution with proxy also. So, in my opinion, this new widget goes into spyder.  code, as it is supposed to be temporarily. I'll try to find out other types of solution, has it appears only in dark theme, may changing in the cascade definition will take some effect. 

@goanpeca if you could share other trials you've made on qdark issues, I can try other things.

Tks

goanpeca:
This is the same idea of replacing the popup but using a QMenu (it was easier to make a POC with it)

Screen Shot 2019-07-04 at 10 33 50

goanpeca:
@ccordoba12 I would say lets go with this fix and try it a bit to see how it runs for users and devs. I really ran out of ideas here besides the reimplementation of a custom QComboBox, but that would still require a lot of testing for corner cases and functionality to get right. Would be the safest but is extra work :-|

spyder/app/mainwindow.py
```diff
@@ -269,6 +270,19 @@ def qt_message_handler(msg_type, msg_log_context, msg_string):
 qInstallMessageHandler(qt_message_handler)
 
 
+class SpyderProxyStyle(QProxyStyle):
```
ccordoba12: Let's move this one to `spyder/utils/qthelpers.py` because I think it fits better there.

goanpeca: Ok will do!

PR #9687, opened 2019-06-26
PR: Restored selection after replaced in selection

### LINKED ISSUE #9685: Selected text unselected after "Replace selection"
### Issue Report Checklist

## Problem Description
When selecting a text and doing "Replace selection", the selection is gone. This is a bit inconvenient when wanting to make several replacements within the same selection as you need to select the text again.

After checking the code, the "Replace selection", removes the old text, hence, the selections are gone missing, and then inserts the new text. Hence, to solve this, one would need to perform a new selection in the code (or perform the replacement in a different way).
https://github.com/spyder-ide/spyder/blob/1c795a749b73ab39184fc7f54188502c967526ab/spyder/widgets/findreplace.py#L573-L579
This holds for all versions I've tested 3.3.2, 3.3.4, and the latest master. (And since the behaviour is expected from the code, I'm pretty certain it is valid.)

(I've also noted some issues with not being able to do "Replace all" when there is some text selected, but that I just noticed so I cannot really say how consistent that issue is.)

1. Select a block of text
2. Perform "Replace selection"

The resulting text block is still selected.

oscargus:
## Description of Changes
After performing "Replace selection", the selection used to go away. With this PR it is restored after replacement.

afterreplaceselection

As far as I can tell there are not tests for the findreplace widget, so it is not straightforward to add a test for it.

### Issue(s) Resolved

Fixes #9685

### Affirmation

I certify the above statement is true and correct: Oscar Gustafsson/oscargus

### DIFF
--- a/spyder/widgets/findreplace.py
+++ b/spyder/widgets/findreplace.py
@@ -571,12 +571,20 @@ class FindReplace(QWidget):
             replacement = re_pattern.sub(replace_text, selected_text)
             if replacement != selected_text:
                 cursor = self.editor.textCursor()
+                start_pos = cursor.selectionStart()
                 cursor.beginEditBlock()
                 cursor.removeSelectedText()
                 if not self.re_button.isChecked():
                     replacement = re.sub(r'\\(?![nrtf])(.)', r'\1', replacement)
                 cursor.insertText(replacement)
+                # Restore selection
+                self.editor.set_cursor_position(start_pos)
+                newl_cnt = replacement.count(self.editor.get_line_separator())
+                sel_len = len(replacement) - newl_cnt
+                for c in range(sel_len):
+                    self.editor.extend_selection_to_next('character', 'right')
                 cursor.endEditBlock()
+
             if focus_replace_text:
                 self.replace_text.setFocus()
             else:
--- a/spyder/plugins/editor/widgets/tests/test_editor.py
+++ b/spyder/plugins/editor/widgets/tests/test_editor.py
@@ -419,8 +419,10 @@ def test_selection_escape_characters(editor_find_replace_bot, qtbot):
     cursor.select(QTextCursor.LineUnderCursor)
     assert cursor.selection().toPlainText() == "\\n \\t escape characters"
 
-    #replace
+    # Replace
     finder.replace_find_selection()
+    # Test that selection is correct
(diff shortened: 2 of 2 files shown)


spyder/widgets/findreplace.py
```diff
@@ -571,12 +571,18 @@ def replace_find_selection(self, focus_replace_text=False):
...
                 cursor = self.editor.textCursor()
+                start_pos = cursor.selectionStart()
                 cursor.beginEditBlock()
                 cursor.removeSelectedText()
                 if not self.re_button.isChecked():
                     replacement = re.sub(r'\\(?![nrtf])(.)', r'\1', replacement)
                 cursor.insertText(replacement)
+                # Restore selection
+                self.editor.set_cursor_position(start_pos)
+                for c in range(len(replacement) - replacement.count(self.editor.get_line_separator())):
```
ccordoba12: This line is too long. Please define a new variable and assign it the value of `len(replacement) - replacement.count(self.editor.get_line_separator())` to avoid that.

goanpeca:
Thanks for working on this @oscargus!

spyder/widgets/findreplace.py
```diff
@@ -571,12 +571,20 @@ def replace_find_selection(self, focus_replace_text=False):
...
+                start_pos = cursor.selectionStart()
                 cursor.beginEditBlock()
                 cursor.removeSelectedText()
                 if not self.re_button.isChecked():
                     replacement = re.sub(r'\\(?![nrtf])(.)', r'\1', replacement)
                 cursor.insertText(replacement)
+                # Restore selection
+                self.editor.set_cursor_position(start_pos)
+                sel_len = len(replacement) - \
+                            replacement.count(self.editor.get_line_separator())
```
goanpeca: @oscargus thanks for the change!

We generally discourage the use of the `\` for continuation lines, so maybe if it is still too long we could do something like

```python
                # Restore selection
                self.editor.set_cursor_position(start_pos)
                separator = self.editor.get_line_separator()
                selection_len = len(replacement) - replacement.count(separator)
```

oscargus:
I think I've figured out how to add a test for this so I will add that soon.

ccordoba12:
Ok, thanks a lot!

PR #9710, opened 2019-06-28
PR: Replace in correct file after startup

### LINKED ISSUE #9688: Replace selection in wrong tab on startup
### Issue Report Checklist

    * [x] Restarted Spyder

## Problem Description
After startup, if the first thing you do is: select a text, pressing ctrl+r, and press "Replace selection", the selection is performed in the last open tab, which is not necessarily the one on top. See movie (note that it is the rightmost tab that changes as seen by the *, I can confirm that the replacement happens there):

replacewrongtab

It is also possible to change the search/replace patterns.

Probably the solution is to make sure that the tab on top is also the active editor. However, I have no idea where that should be done in the code. 

Now it is kept on top, but (probably) not activated after all tabs are opened. As soon as something other than selection happens in the top window, it will replace correctly. (I think it happens for the other replace buttons as well, but mainly tried the "Replace selection".)

1. Close Spyder with multiple tabs open, with the focus not on the right-most one.
2. Open Spyder
3. Mark some text in the open tab
4. Press ctrl+r to open the find-replace widget
5. Press "Replace selection"

The replacement should be done in the active window. 

## Versions

* Spyder version: 4.0.0.dev0
* Python version: 3.7.3 64-bit
* Qt version: 5.9.6
* PyQt version: 5.9.2

### Dependencies

```
qdarkstyle >=2.6.4:  2.7 (OK)
sphinx >=0.6.6    :  2.1.0 (OK)
pyls >=0.27.0     :  0.27.0 (OK)
sympy >=0.7.3     :  1.4 (OK)
cython >=0.21     :  0.29.10 (OK)
qtconsole >=4.5.0 :  4.5.1 (OK)
IPython >=4.0     :  7.5.0 (OK)

```

oscargus:
## Description of Changes
Make sure that the active tab is also the one set for the find/replace widget. This was a problem on startup as the find/replace widget had the last opened tab even though another tab may be on top.

### Issue(s) Resolved

Fixes #9688

### Affirmation

I certify the above statement is true and correct: Oscar Gustafsson/oscargus

### DIFF
--- a/spyder/plugins/editor/widgets/editor.py
+++ b/spyder/plugins/editor/widgets/editor.py
@@ -780,7 +780,7 @@ class EditorStack(QWidget):
            and not sys.platform == 'darwin':
             # Don't set document mode to true on OSX because it generates
             # a crash when the editor is detached from the main window
-            # Fixes Issue 561
+            # Fixes spyder-ide/spyder#561
             self.tabs.setDocumentMode(True)
         self.tabs.currentChanged.connect(self.current_changed)
 
@@ -1282,7 +1282,7 @@ class EditorStack(QWidget):
         if set_new_index:
             self.set_stack_index(new_index)
         else:
-            # Fixes Issue 1287
+            # Fixes spyder-ide/spyder#1287
             self.set_current_filename(current_fname)
         if self.outlineexplorer is not None:
             self.outlineexplorer.file_renamed(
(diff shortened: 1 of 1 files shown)


spyder/plugins/editor/widgets/editor.py
```diff
@@ -2028,6 +2028,10 @@ def current_changed(self, index):
             pass
 
         self.update_plugin_title.emit()
+        # Make sure that any replace happens in the editor on top
+        # See issue 9688
```
goanpeca:
```suggestion
        # See spyder-ide/spyder#9688
```

goanpeca: @oscargus thanks for working on this!

We now have `url awareness` for issues so we now use this syntax for issues on comments so we can `Ctrl/Cmd` click on the text and go to the issue :-)

oscargus:
Good to know! I fixed the other issues mentioned in the file as well.

No, I think that it is hard to add a test for it. It may be that the test fixtures activates the top tab during the test, so one will not know the actual reason...

Issue #9367, opened 2019-05-19
Some issues with the Plot viewer


goanpeca:
## Problem Description

I will group a couple of issues I see currently with the `Plots` pane:

1. Images on retina displays look distorted. (maybe this thread helps https://gist.github.com/minrk/3301035)
2. There seems to be some horizontal scrolling with the list of images to the right:
![problem1](https://user-images.githubusercontent.com/3627835/57983399-fdf3e700-7a16-11e9-820e-c83ebd9bdcc3.gif)
3. The down arrow should allow moving between images, instead it only move the list downwards
![problem2](https://user-images.githubusercontent.com/3627835/57983424-457a7300-7a17-11e9-8598-a5afcd20e188.gif)

goanpeca:
Thoughts @jnsebgosselin :-) ?

goanpeca:
@jnsebgosselin any thoughts here?

jnsebgosselin:
I @goanpeca. Thanks for taking the time to report these issues.

> Images on retina displays look distorted.

I will try to take a look at this. However I don't have access to a Mac. Is this doing this also with any high resolution monitors?

> There seems to be some horizontal scrolling with the list of images to the right

I will take care of this, I know what needs to be done.

> The down arrow should allow moving between images, instead it only move the list downwards

I don't know about this. The down/up arrows act like scrollbar buttons, so it makes sense to me that clicking on them doesn't change the current selected image.

This is similar to the caret not changing line when we use the arrows on the Editor's vertical scrollbar. 

Besides, There is already the `Previous Figure` and `Next Figure`  buttons and their associated shortcut to navigate/cycle through the images.

But that being said, if you still think it is a change that should be implemented, I can do it.

jnsebgosselin:
> Images on retina displays look distorted.

@goanpeca is this related to this https://github.com/spyder-ide/spyder/issues/9148?

goanpeca:
Yes, I think they are related

RIM-LLC:
I'm also having what appears to be an issue with the quality of the images plotted on the Plots window. See on the image below that the letters on the charts are blurry. If I save the image using the "Save" icon on the navigation bar, the quality is OK. This image was a snap from a 4K monitor (Windows 10). Note that I have "Enable auto high DPI scaling" checked (which indeed helped with the size of menus, icons, etc).
[...]

PR #9720, opened 2019-06-29
PR: Make the width of thumbnails scrollbar resizable

### LINKED ISSUE #9747: Separator on plots plugin acts more like a button than a resize area
The separator has the appearance of allowing the user to change the width gradually, but currently the behavior is more of something being toggled (a button instead of a draggable area)

plots

jnsebgosselin:
### Issue(s) Resolved

Fixes item 2 of #9367
Fixes #9747

thumbnail_scrollbar

### Affirmation


jnsebgosselin:
This is not ready for review yet @goanpeca, there is something I still need to figure out with the dark style.

jnsebgosselin:
Ok @goanpeca, I've added some workaround to handle some padding and margin values that are hard coded in the qdarkstyle.

I haven't been able to get rid or figure out exactly where the extra 2px * 2  comes from in the dark style... if you can find a better solution, I'm all hears.

goanpeca:
Will take a loon

goanpeca:
@jnsebgosselin I think it looks good for now. Maybe adding a `TODO:` where you left the darkstyle hardcoding.

Also it would be nice to have a centralized place for all these hardcoded nitpicks we will probably still need with QDark to keep things in order. Any thoughts there?

---

I have some comments regarding the widget but this can be addressed on a separate PR if you wish so. I write them here cause I noticed this while reviewing the PR

1. The arrows look rather slim and are hard to get when using the mouse.
Screen Shot 2019-07-02 at 20 22 13

2. The separator has the appearance of allowing the user to change the width gradually, but currently the behavior is more of something being toggled (a button instead of a draggable area)

plots

I think we can improve that UX @jnsebgosselin, thoughts?

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -636,6 +640,30 @@ def setup_arrow_buttons(self):
...
+
+        # Calculate and set the width of the thumbnail scrollbar using
+        # the size of the fake thumbnail.
+        thumbnail_scrollbar_width = (
+            thumbnail.width() +
+            self.scrollarea.viewportMargins().left() +
+            self.scrollarea.viewportMargins().right() +
+            2 * self.lineWidth())
+        if is_dark_interface():
+            # This is required to take into account some hard-coded padding
```
goanpeca:
```suggestion
            # TODO: This is required to take into account some hard-coded padding
```

Maybe to point that this could be improved in the future?

jnsebgosselin: > I have some comments regarding the widget but this can be addressed on a separate PR if you wish so. I write them here cause I noticed this while reviewing the PR
> 
>     1. The arrows look rather slim and are hard to get when using the mouse.
> 
> 
> Screen Shot 2019-07-02 at 20 22 13
> 
>     1. The separator has the appearance of allowing the user to change the width gradually, but currently the behavior is more of something being toggled (a button instead of a draggable area)
> 
> 
> plots
> 
> I think we can improve that UX @jnsebgosselin, thoughts?

These are good point. Please open a new issue for these and I will address them in a subsequent PR.

jnsebgosselin:
> I have some comments regarding the widget but this can be addressed on a separate PR if you wish so. I write them here cause I noticed this while reviewing the PR
> 
>     1. The arrows look rather slim and are hard to get when using the mouse.
> 
> 
> Screen Shot 2019-07-02 at 20 22 13
> 
>     1. The separator has the appearance of allowing the user to change the width gradually, but currently the behavior is more of something being toggled (a button instead of a draggable area)
> 
> 
> plots
> 
> I think we can improve that UX @jnsebgosselin, thoughts?

These are good point. Please open a new issue for these and I will address them in a subsequent PR.

jnsebgosselin:
> Also it would be nice to have a centralized place for all these hardcoded nitpicks we will probably still need with QDark to keep things in order. Any thoughts there?

It is a good point, but in that case, since I do not really know where this padding comes from, I don't know how useful that would be.

It is just a thought, but I don't understand why there is a need to have hardcoded margin and padding in qdarkstyle, in addition to the default one. I mean, if an application is tailored to use qdarkstyle and qdarkstyle only, there is no problem, but in a scenario where we have to support multiple styles, this makes things a lot more complicated...

spyder/plugins/plots/widgets/figurebrowser.py
jnsebgosselin: I agree with this, but imo the comment then needs to be reworded so that it explains what needs to be done, not why this exception was added here, otherwise adding a TODO is kind of futile.

jnsebgosselin:
And @goanpeca thanks for the review btw!

spyder/plugins/plots/widgets/figurebrowser.py
jnsebgosselin: Gimme a day or too just to look at this again with a fresh eye.

goanpeca:
@jnsebgosselin I am fie with this PR how it is :-), anything extra you wanted to do?

jnsebgosselin:
> 
> 
> @jnsebgosselin I am fie with this PR how it is :-), anything extra you wanted to do?

Yes, if the third option listed in https://github.com/spyder-ide/spyder/issues/9747 is selected , I would like to implement it here.

ccordoba12:
@goanpeca, I dismissed your review to not merge this one by accident.

ccordoba12:
@dalthviz, please help @jnsebgosselin to fix the failing tests and finish this one.

jnsebgosselin:
@ccordoba12 Sorry for the delay in completing this one. I can work on this next Sunday if you this can wait a little bit more...

Same for the other PR about icon size.

ccordoba12:
> I can work on this next Sunday if you this can wait a little bit more...
Same for the other PR about icon size.

Thanks a lot @jnsebgosselin! It'd be great if we could include those two PRs in beta5!! (to be released at the end of the month).

I'll wait for your updates then :+1:

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -634,6 +637,15 @@ def set_figureviewer(self, figure_viewer):
...
         self.figure_viewer = figure_viewer
 
+    def eventFilter(self, widget, event):
+        """
+        An event filter to trigger an update of the thumbnails size so that
+        their width fit that of the scrollarea.
+        """
+        if event.type() == QEvent.Resize:
+            self._update_thumbnail_size()
+        return super().eventFilter(widget, event)
```
ccordoba12: This syntax is incorrect in Python 2. That's what's making our tests fail in that version.

ccordoba12:
@jnsebgosselin, one of Plots tests is failing in Python 2 and `test_custom_layouts` is causing a segfault on Linux (please skip it in that platform, as I suggested above).

jnsebgosselin:
@ccordoba12 I fixed the test and made some improvement to the code to make it clearer.

That would be cool though if Spyder could remember the state of the splitter between the main figure viewer and the thumbnails scrollbar. This is not very hard to implement. I can do this here if you wish or this can go in another PR.

goanpeca:
>That would be cool though if Spyder could remember the state of the splitter between the main figure viewer and the thumbnails scrollbar. This is not very hard to implement. I can do this here if you wish or this can go in another PR.

@jnsebgosselin I think this PR is "simple" enough, so I would say implement it here :-)

jnsebgosselin:
> 
> 
> > That would be cool though if Spyder could remember the state of the splitter between the main figure viewer and the thumbnails scrollbar. This is not very hard to implement. I can do this here if you wish or this can go in another PR.
> 
> @jnsebgosselin I think this PR is "simple" enough, so I would say implement it here :-)

Thanks @goanpeca.

But, now that I'm thinking about this a little bit harder, I'm not sure it was a good idea to propose to store and restore the state of the splitter in this PR. The Plot plugin has a stackwidget that can hold multiple plot viewers, each with their own splitter state. So this makes things a little bit more complicated UX wise. Beside, the state of the plot plugin is reset anyway after each restart (i.e. plots are not reloaded in the viewer), so I'm not sure it is a big deal if the state of the splitter is reset too.

Bottom line, I think we should discard this idea for the moment.

goanpeca:
>Bottom line, I think we should discard this idea for the moment.

Sound good too 

Is this ready for a final review then @jnsebgosselin ?

jnsebgosselin:
> Is this ready for a final review then @jnsebgosselin ?

Yes it is, thanks!

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -685,34 +696,74 @@ def save_figure_as(self, fig, fmt):
...
+            2 * self.lineWidth() -
+            self.scrollarea.viewportMargins().left() -
+            self.scrollarea.viewportMargins().right() -
+            thumbnail.savefig_btn.width() -
+            thumbnail.layout().spacing()
+            )
+        if is_dark_interface():
+            # This is required to take into account some hard-coded padding
+            # and margin in qdarkstyle.
+            figure_canvas_width = figure_canvas_width - 6
```
goanpeca: Is this looking ok on Win/Linux/MAc ( I can check mac :-p )

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -685,34 +696,74 @@ def save_figure_as(self, fig, fmt):
...
             canvas_width = canvas_height / fheight * fwidth
         thumbnail.canvas.setFixedSize(int(canvas_width), int(canvas_height))
+        thumbnail.layout().setColumnMinimumWidth(0, max_canvas_size)
+
+    def _update_thumbnail_size(self):
+        """
+        Update the thumbnails size so that their width fit that of
+        the scrollarea.
+        """
+        # NOTE: we hide the view widget while we update the thumbnails size
```
goanpeca: When is this happening  @jnsebgosselin, I do not follow

goanpeca:
Couple of questions :-p, otherwise looks good!

jnsebgosselin:
> 
> 
> Couple of questions :-p, otherwise looks good!

Awesome, I will answer them later tonight and thank you.

goanpeca:
Thanks for working on this 😄

spyder/plugins/plots/widgets/figurebrowser.py
jnsebgosselin: I've only test on Windows unfortunately.

spyder/plugins/plots/widgets/figurebrowser.py
jnsebgosselin: This happens every time a `Resize` event occur on the scrollarea of the thumbnail scrollbar widget.

spyder/plugins/plots/widgets/figurebrowser.py
goanpeca: Ok, I will check on linux and mac ;-)

jnsebgosselin: Thanks!

spyder/plugins/plots/widgets/figurebrowser.py
jnsebgosselin: @goanpeca Is this answering your question or you would like to have more explanation about the hide/show calls enclosing the updating of the thumbnails size?

goanpeca: What I mean is that I do not see what is hidden in the animation you posted. Or maybe I am still not understanding 🙃 ?

jnsebgosselin: Ok I understand. In reality, nothing is really hidden, it is just a trick to make the resizing animation of the thumbnails look smoother.

The goal here is to prevent a repaint of each thumbnail as soon as their size is updated in the loop, which causes some flickering when the scrollbar is resized. So what I do to prevent that is to hide temporarily all the thumbnails while their size is updated. Once the size of all the thumbnails has been updated, I then show them back so that they are repainted all at once instead of one after the other.

goanpeca: Ahhh ok got it. Maybe adding that as a comment makes life easy :-) for next devs

goanpeca:
@jnsebgosselin on OSX there are the following issues:

- For HDPI screens (like my mac) the images looks pixelated
- The buttons on the thumbnails to the right are cropped

plots

jnsebgosselin:
Thanks you very much for the tests.

>     * For HDPI screens (like my mac) the images looks pixelated

The plots plugin generates the image from the data it receives from the console. The images look pixelated because the pngs sent by the console have a low resolution. There is not much that can be done here, I mean from the plot viewer perspective. 

Does changing the resolution of the inline backend helps in the console preferences?

>     * The buttons on the thumbnails to the right are cropped

Ok, I do not see another solution than to increase the hard-coded padding in `_calculate_figure_canvas_width`. How much more padding do you think would be required?

goanpeca:
>Ok, I do not see another solution than to increase the hard-coded padding in _calculate_figure_canvas_width. How much more padding do you think would be required?

Let me push a PR so we gte the right value for OSX :-)

>The plots plugin generates the image from the data it receives from the console. The images look pixelated because the pngs sent by the console have a low resolution. There is not much that can be done here, I mean from the plot viewer perspective.

>Does changing the resolution of the inline backend helps in the console preferences?


Yes, this is for another PR (I think I tried this before and I got bigger images... but still pixelated)

jnsebgosselin:
> Let me push a PR so we gte the right value for OSX :-)

Ok thanks. Is this cropped in the light theme also or only in the dark one?

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -685,34 +696,70 @@ def save_figure_as(self, fig, fmt):
...
     # ---- Thumbails Handlers
-    def add_thumbnail(self, fig, fmt):
-        thumbnail = FigureThumbnail(background_color=self.background_color)
-        thumbnail.canvas.load_figure(fig, fmt)
+    def _calculate_figure_canvas_width(self, thumbnail):
+        """
+        Calculate the witdh the thumbnail's figure canvas need to have for the
+        thumbnail to fit the scrollarea.
+        """
+        figure_canvas_width = (
```
goanpeca:
```suggestion
        extra_padding = 10 if sys.platform == 'darwin' else 0
        figure_canvas_width = (
```

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -685,34 +696,70 @@ def save_figure_as(self, fig, fmt):
...
+            self.scrollarea.width() -
+            2 * self.lineWidth() -
+            self.scrollarea.viewportMargins().left() -
+            self.scrollarea.viewportMargins().right() -
+            thumbnail.savefig_btn.width() -
+            thumbnail.layout().spacing()
```
goanpeca:
```suggestion
            thumbnail.layout().spacing() - extra_padding
```

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -21,8 +21,8 @@
 from qtpy.QtGui import QPixmap, QPainter, QKeySequence
 from qtpy.QtWidgets import (QApplication, QHBoxLayout, QMenu,
                             QVBoxLayout, QWidget, QGridLayout, QFrame,
-                            QScrollArea, QPushButton, QScrollBar, QSizePolicy,
-                            QSpinBox, QSplitter, QStyle)
+                            QScrollArea, QPushButton, QScrollBar, QSpinBox,
+                            QSplitter, QStyle)
```
goanpeca: @jnsebgosselin we need to add an `import sys`

goanpeca:
@jnsebgosselin there must be something missing on the space calculation for OSX, the fixed 10 value, seems to work as expected and for now I think it is good enough 🤷‍♂ ?

Let me know what you think!

jnsebgosselin:
> 
> 
> @jnsebgosselin there must be something missing on the space calculation for OSX, the fixed 10 value, seems to work as expected and for now I think it is good enough 🤷‍♂ ?
> 
> > Ok thanks. Is this cropped in the light theme also or only in the dark one?
> 
> It was the same on both themes, dark and light!
> 
> Let me know what you think!

Ok thank you @goanpeca for your help with this. I think it is good enough for now. I really don't know where these extra padding come from... 

The question now, before we can finish this, is to know whether the same problem also appears in Linux...

ccordoba12:
> The question now, before we can finish this, is to know whether the same problem also appears in Linux...

I tested on Linux and buttons are not cropped there, so this is only a macOS problem.

@jnsebgosselin, please address @goanpeca's last comments so we can merge this one because I think it's almost ready. Thanks!

jnsebgosselin:
> > The plots plugin generates the image from the data it receives from the console. The images look pixelated because the pngs sent by the console have a low resolution. There is not much that can be done here, I mean from the plot viewer perspective.
> 
> > Does changing the resolution of the inline backend helps in the console preferences?
> 
> Yes, this is for another PR (I think I tried this before and I got bigger images... but still pixelated)

@goanpeca  I think this should be true for the inline plots in the IPython console, but is it also the case for the plots plugin?

ccordoba12:
> I think this should be true for the inline plots in the IPython console, but is it also the case for the plots plugin?

@jnsebgosselin, we need to tell the inline backend to generate retina display plots (there's an option for that in IPython). However, that's a tricky business now because users can move Spyder or the IPython console window (if detached) to a HiDPI monitor from a low resolution one and vice-versa.

So we have to detect when the window is moved to another monitor and its screen resolution to tell all kernels when to turn on/off retina display plots automatically.

ccordoba12:
@jnsebgosselin, for next time please add all code change suggestions to the batch and apply them as a single commit. Thanks!

jnsebgosselin:
> 
> 
> @jnsebgosselin, for next time please add all code change suggestions to the batch and apply them as a single commit. Thanks!

This is done from the GitHub interface? OK I will do that next time, thanks for the tip.

ccordoba12:
> This is done from the GitHub interface? OK I will do that next time, thanks for the tip.

Yeah, there's a button in each suggestion called `Add suggestion to batch`.

jnsebgosselin:
> 
> 
> > I think this should be true for the inline plots in the IPython console, but is it also the case for the plots plugin?
> 
> @jnsebgosselin, we need to tell the inline backend to generate retina display plots (there's an option for that in IPython). However, that's a tricky business now because users can move Spyder or the IPython console window (if detached) to a HiDPI monitor from a low resolution one and vice-versa.
> 
> So we have to detect when the window is moved to another monitor and its screen resolution to tell all kernels when to turn on/off retina display plots automatically.

Ok, but the plots plugin doesn't work exactly like the IPython console because it allow for the figures to be resized. So I thought that increasing the dpi of the inline figures would increase the quality of the images and thumbnails shown in the plots plugin and offer a temporary solution to this problem.

goanpeca:
Awesome! thanks @jnsebgosselin :-)

jnsebgosselin:
Yeah, thanks a lot for your generous patience and help with this. This has proven to be more tricky than I anticipated...

PR #9726, opened 2019-07-01
Right justify keyboard shortcuts in menus

Khalilsqu:
## Description of Changes

shortcuts_right_justify

This PR right justify all keyboard shortcuts found in all menus and context menus in Spyder. Unfortunately it does not work when dark theme is used. The reason is probably related to the fact that Qt style sheets are currently not supported for custom QStyle subclasses as described [here](https://doc.qt.io/archives/qt-4.8/stylesheet.html). I could not find another simple way to make the same changes except using QStyle subclasses.

This is a very minor change and probably might not be necessary one. However, I am just submitting it for your review. 

### Issue(s) Resolved

Fixes #

### Affirmation

I certify the above statement is true and correct: @Khalilsqu

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -105,7 +105,7 @@ if hasattr(Qt, 'AA_EnableHighDpiScaling'):
 # Create our QApplication instance here because it's needed to render the
 # splash screen created below
 #==============================================================================
-from spyder.utils.qthelpers import qapplication, MENU_SEPARATOR
+from spyder.utils.qthelpers import qapplication, MENU_SEPARATOR, MenuProxyStyle
 from spyder.config.base import get_image_path
 MAIN_APP = qapplication()
 
@@ -1267,11 +1267,19 @@ class MainWindow(QMainWindow):
         # Menu about to show
         for child in self.menuBar().children():
             if isinstance(child, QMenu):
+                # Next 2 lines used to right justify keyboard shortcuts.
+                self._proxy = MenuProxyStyle(child.style())
+                child.setStyle(self._proxy)
                 try:
                     child.aboutToShow.connect(self.update_edit_menu)
                     child.aboutToShow.connect(self.update_search_menu)
                 except TypeError:
                     pass
+
+        # Right justify shortcuts.
+        for menu_keyb_shortcut in [self.plugins_menu, self.quick_layout_menu]:
+            self._proxy = MenuProxyStyle(menu_keyb_shortcut.style())
+            menu_keyb_shortcut.setStyle(self._proxy)
 
         logger.info("*** End of MainWindow setup ***")
         self.is_starting_up = False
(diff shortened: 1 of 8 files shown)


ccordoba12:
To avoid the failures in our tests perhaps you should avoid applying MenuProxyStyle when `running_under_pytest` returns True.

spyder/app/mainwindow.py
```diff
@@ -1273,6 +1273,18 @@ def add_ipm_action(text, path):
                 except TypeError:
                     pass
 
+        # Right justify keyboard shortcuts.
+        # Apply only for none dark themes, otherwise tests fail.
+        if not is_dark_interface():
```
ccordoba12: This means that your fix won't apply to the dark theme. Since that's going to be the default theme in Spyder 4, that sort of defeats the purpose of your work (which I find very nice).

So instead of using `is_dark_interface` in this PR, please use instead `running_under_pytest` (as I instructed you to do it before).

spyder/utils/qthelpers.py
```diff
@@ -598,6 +598,31 @@ def create_plugin_layout(tools_layout, main_widget=None):
...
+            vals = option.text.split("\t")
+
+            if len(vals) == 2:
+                text, shortcut = vals
+                option.text = text
+        super(MenuProxyStyle, self).drawControl(element, option, painter,
+                                                widget)
+        if shortcut:
+            margin = 10  # Margin from the right of menu column to shortcuts.
+            self.proxy().drawItemText(painter,
```
goanpeca: This indentation though valid, looks a bit confusing. I suggest something like:

```python
            self.proxy().drawItemText(
                painter,
                option.rect.adjusted(margin, 0, -margin, 0),
                Qt.AlignRight | Qt.AlignVCenter,
                option.palette,
                option.state & QStyle.State_Enabled,
                shortcut, QPalette.Text)
```

spyder/utils/qthelpers.py
```diff
@@ -598,6 +598,31 @@ def create_plugin_layout(tools_layout, main_widget=None):
...
+        if element == QStyle.CE_MenuItem:
+            vals = option.text.split("\t")
+
+            if len(vals) == 2:
+                text, shortcut = vals
+                option.text = text
+        super(MenuProxyStyle, self).drawControl(element, option, painter,
+                                                widget)
+        if shortcut:
+            margin = 10  # Margin from the right of menu column to shortcuts.
```
goanpeca: This works as a constant, so I would probably moved this out of the method and make it a class variable

```
class MenuProxyStyle(QProxyStyle):
    RIGHT_MARGIN = 10  # Margin from the right of menu column to shortcuts.
```
It might help to add the units? I am guessing pixels/px?

goanpeca:
@ccordoba12 these proxy styles seem to be causing issues in general, as mentioned in the other PR about the combobox issue. Do you think this could be a bug with pyqt5?

Since this change is applied to menus only I am ok with merging as it will not likely generate crashes on the application.

Khalilsqu:
@ccordoba12 @goanpeca . Unfortunately this will not work for dark themes which are using QSS wither tests are activated or not. This is a limitation with QSS. I think I probably will go with reimplementation of QMenu. Proxy styles cause serious artifacts when used for widgets getting their styles from QSS. Hence I am closing this PR for the time being until I find a better solution.

ccordoba12:
Ok, thanks for your effort!

PR #9741, opened 2019-07-04
PR: Add basic config dlg tests for plugins

goanpeca:
## Description of Changes

In preparing for the changes that will be made on the config system, the plugins now have a basic fixture to test the config dialog. These tests also allow to show what are the methods in the main window that should be exposed by the api (see the mainwindowmocks)

### Affirmation


goanpeca:
To explain a little better what is going on. This tests is using parametrizable fixtures which can be achieved by using the indirect option. Unfortunately pytest explanation for this usage is a bit cryptic. 

Beside the pytest "weirdness" the tests in here allow the following:
- Use a single fixture to test any type of config plugin. This will be also usable by any plugin developers
- Test the plugins config pages in standalone mode (that is why we use windows mocks on some places)
- Test the save_to_conf method on all the plugin config pages. This is to check that something like what happened on this issue (https://github.com/spyder-ide/spyder/pull/9706), does not happen with the plugins.

spyder/plugins/onlinehelp/tests/test_onlinehelp_config_dialog.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+@pytest.mark.parametrize(
+    'config_dialog',
+    # [[MainWindowMock, [ConfigPlugins], [Plugins]]]
+    [[MainWindowMock, [], [OnlineHelp]]],
+    indirect=True)
+def test_config_dialog(config_dialog):
+    configpage = config_dialog.get_page()
+    assert configpage is None
```
ccordoba12: Do we really need to add tests for a non-existing config pages? I'd say this is really not necessary and it could be confusing to new contributors.

goanpeca: @ccordoba12 the idea is that if at some point we do add a config page for those plugins test would fail, which require the user to pay attention that tests need to be added to this plugin.

ccordoba12: Ok, I understand. However, I'm a bit -1 on this. Tests should be added when the confpage is added too.

So I'd say we should remove these tests for now.

goanpeca: Ok will remove them!

Also I have a problem with the `help` plugin. I will post in a comment

ccordoba12: Ok, thanks!

goanpeca:
Remove test for plugins with no config page @ccordoba12

PR #9745, opened 2019-07-04
PR: Fix height of thumbnail scrollbar arrow buttons in the Plots plugin

### LINKED ISSUE #9746: Plots viewer arrows a bit too small
The arrows look rather slim and are hard to get when using the mouse.
Screen Shot 2019-07-02 at 20 22 13

jnsebgosselin:
Fixes #9746

## Description of Changes

As pointed out in https://github.com/spyder-ide/spyder/pull/9720#issuecomment-507902110, the height of the arrow buttons in the thumbnail scrollbar of the plot plugins was too small.

I think this is a macOS issue caused by the fact that scrollbars do not seem to have arrow buttons in this OS.

So instead of setting the height of the arrow button by getting the value from the style, we will set it from the size hint height of the horizontal scrollbar.

image

### DIFF
--- a/spyder/plugins/plots/widgets/figurebrowser.py
+++ b/spyder/plugins/plots/widgets/figurebrowser.py
@@ -614,24 +614,18 @@ class ThumbnailScrollBar(QFrame):
         Setup the up and down arrow buttons that are placed at the top and
         bottom of the scrollarea.
         """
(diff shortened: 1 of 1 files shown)


goanpeca:
This looks great

Thanks for working on this @jnsebgosselin!

PR #9774, opened 2019-07-07
[WIP] PR: Fixed problems with mixed line ending and find in files regex search

### LINKED ISSUE #9765: Find in files with mixed line endings
### Issue Report Checklist

    * [x] Restarted Spyder

## Problem Description
When searching a regex using "Find in files" (now "Find") with line ending (in my case `#[0-9]+$`), it seems like Spyder does not find all instances. My best guess is that this is caused by mixed line endings and the regex only finding one of them. Not sure which one though.

Probably relates to https://github.com/spyder-ide/spyder/issues/2958 and https://github.com/spyder-ide/spyder/issues/2424 when it comes to the Spyder code, but in general it is a problem working on any project/code/file that may have mixed line-endings (or not the same as the system or whatever it is that causes it).

This is the most likely way I have figured out:

On a Windows(?) machine search for `#[0-9]+$` which gives 64 hits. Note that there are no hits in CHANGELOG.md.

Now, instead search for `#[0-9]+` (which does not involve line endings). Note that there are several instances of a valid pattern in CHANGELOG.md ending a line. For example, lines 2286 and 2305.

Probably it works differently on a different platform.

All hits, independent of line ending should be found.

## Versions

* Spyder version: 4.0.0.dev0 (also 3.3.4)
* Python version: 3.7.3 64-bit
* Qt version: 5.9.6
* PyQt version: 5.9.2

### Dependencies

```
qdarkstyle >=2.6.4:  2.7 (OK)
sphinx >=0.6.6    :  2.1.0 (OK)
pyls >=0.27.0     :  0.27.0 (OK)
sympy >=0.7.3     :  1.4 (OK)
cython >=0.21     :  0.29.10 (OK)
qtconsole >=4.5.0 :  4.5.1 (OK)
IPython >=4.0     :  7.5.0 (OK)

```

oscargus:
## Description of Changes

Normalized the line endings so that the find in files functionality works with the `$` regex classifier (end of line).

As the file was opened in binary mode, the built-in Python normalization (from Python 3) did not work, neither did the Python 2 Universal Newlines support.

Still looking to add a test, but have some problems creating a test file with mixed line endings...

Is support for a single `\r` really required? This is only used on pre-OSX-Mac and I can change it to a single replace if there is no need to support it. (Also slightly easier to create a test file for it without the \r case...)

### Issue(s) Resolved

Fixes #9765

### Affirmation


ccordoba12:
> Is support for a single \r really required?

Not really, so please don't consider it either in code and the test you're planning to add.

oscargus:
I've added a test now. Also, the approach is slightly improved although there are some potential problems. Right now, any `\r\n` line endings are changed to `\n` (as `$` only seems to match `\n`) if there is a regex search and a `$` in the search pattern. 

The potential problem is that it is not possible to explicitly find lines ending with `\r\n` in them if one also searches for a `$`, e.g. `\$.*\r\n`. Still, I think that this case is so rare that it is not worth adding code to detect (as this test has to be done for each line searched and one need to consider the possibility of `\$.*$` as well).

Maybe it is better to just check if the last character is `$` as the current find approach cannot handle multiple lines anyway?

ccordoba12:
> Still, I think that this case is so rare that it is not worth adding code to detect

Ok, no problem with that.

> Maybe it is better to just check if the last character is $ as the current find approach cannot handle multiple lines anyway?

Sure, that's fine for me.

oscargus:
Given that the tests pass, I think this is the best solution. I cannot even see that there is a problem with it as the only time it will "fail" is if the pattern ends with `\$`. In this case it will replace in vain, but it will not miss any results as there will be no matching on `\r\n` ending the line anyway...

ccordoba12:
> Given that the tests pass

Tests are definitely failing for Python 2, so please take a look at that.

> I think this is the best solution

Great! Thanks a lot for your help!

goanpeca:
Is this still valid @ccordoba12 ?

ccordoba12:
It seems relevant, yes. We need to update it to fix the merge conflicts.

PR #9778, opened 2019-07-07
PR: Add conda environment status widget

### LINKED ISSUE #9008: Add current conda environment as status widget
This is an initial read-only change so that the status bar widget (to the right) displays the current name of a conda environment. On hover we will display the full path of the environment.

This is probably a functionality that should be implemented fully on `spyder-conda` package.

Also this ties up with interpreter selection which seems to be subset of this functionality so this initial implementation could live within spyder and check if the selected interpreter lives in a conda environment and display that.

Display the current environment in use as a StatusBar Widget. Allow to select the environments found from there.

This work will tie with the projects refactoring

See #5947 for reference

goanpeca:
## Description of Changes

### Issue(s) Resolved

coda-status

Fixes #9008

### Affirmation

### DIFF
--- a/spyder/utils/icon_manager.py
+++ b/spyder/utils/icon_manager.py
@@ -104,7 +104,7 @@ _resource = {
 }
 
 _qtaargs = {
-    'environment':             [('mdi.package-variant',), {'color': MAIN_FG_COLOR}],
+    'environment':             [('mdi.cube-outline',), {'color': MAIN_FG_COLOR}],
     'drag-horizontal':         [('mdi.drag-horizontal',), {'color': MAIN_FG_COLOR}],
     'format_letter_case':      [('mdi.format-letter-case',), {'color': MAIN_FG_COLOR}],
     'regex':                   [('mdi.regex',), {'color': MAIN_FG_COLOR}],
--- a/spyder/utils/icon_manager.py
+++ b/spyder/utils/icon_manager.py
@@ -104,6 +104,7 @@ _resource = {
 }
 
 _qtaargs = {
+    'environment':             [('fa.archive',), {'color': MAIN_FG_COLOR}],
     'drag-horizontal':         [('mdi.drag-horizontal',), {'color': MAIN_FG_COLOR}],
     'format_letter_case':      [('mdi.format-letter-case',), {'color': MAIN_FG_COLOR}],
     'regex':                   [('mdi.regex',), {'color': MAIN_FG_COLOR}],
--- a/spyder/utils/icon_manager.py
+++ b/spyder/utils/icon_manager.py
@@ -104,7 +104,7 @@ _resource = {
 }
 
 _qtaargs = {
-    'environment':             [('fa.archive',), {'color': MAIN_FG_COLOR}],
+    'environment':             [('mdi.package-variant',), {'color': MAIN_FG_COLOR}],
     'drag-horizontal':         [('mdi.drag-horizontal',), {'color': MAIN_FG_COLOR}],
     'format_letter_case':      [('mdi.format-letter-case',), {'color': MAIN_FG_COLOR}],
     'regex':                   [('mdi.regex',), {'color': MAIN_FG_COLOR}],
--- a/spyder/widgets/tests/test_status.py
+++ b/spyder/widgets/tests/test_status.py
@@ -63,19 +63,31 @@ def test_status_bar_widget_signal(status_bar, qtbot):
 
 def test_status_bar_conda_status(status_bar, qtbot, mocker):
     mocker.patch.object(utils, 'is_anaconda', return_value=True)
+    mock_py_ver = 'Python 6.6.6'
 
     win, statusbar = status_bar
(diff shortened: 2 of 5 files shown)


goanpeca:
@spyder-ide/core-developers ideas to discuss.

I added a click signal to the status widgets so we can do something interesting with them without having to add the possibility of adding any type of widget (QDropdow.. etc...)

For the moment this signal is not used and here are some ideas of how the information could be displayed here:

- If it is a python file display `conda: <env name> (Python interpreter)` and the hover hint should display the path to the python executable. (this is updated with the python interpreter option)
- If it is any other type of file display only the env name `conda: <env name>`
- Eventually the python interpreter preferences pane will become a interpreters pane I guess so that if it is an R file, we can display the R interpreter etc...
- I added the package icon from font awesome but I am of course open to ideas (thoughts @jnsebgosselin ?)

I want to add the ability of changing the interpreter directly from the status bar, so users do not need to go to the preferences. Thoughts here @ccordoba12 ?

jitseniesen:
With Emacs you can change the line endings of a file by clicking on the corresponding indicator (`LF` in your image). I find that quite useful and intuitive, so this is a nice idea.

spyder/utils/icon_manager.py
```diff
@@ -104,6 +104,7 @@
 }
 
 _qtaargs = {
+    'environment':             [('fa.archive',), {'color': MAIN_FG_COLOR}],
```
ccordoba12: I don't like this icon too much. What about `mdi.package-variant`?

goanpeca:
My reply to the comments/questions @ccordoba12 👍 

>What happens if the environment name is too long?

Not at the moment. What should be a reasonable **maximum character** count for this widget?

>Are you eliding it somehow?

I could, see previous question.

>Could we only show the Python version next to environment name? I say it because in your screenshot one of your environments is shown as Python 2.7.5 :: Anaconda, Inc, which I think is too long.

I am getting this info by running `python_exec -V`, in the Anaconda (base environment) case it includes that extra string. Assuming all values will be in the form `Python N.N.N` I can strip it and make it smaller, so it does not include the Anaconda part.

goanpeca:
Screen Shot 2019-07-09 at 10 26 09

jnsebgosselin:
Other options worth considering for the icon :

image

spyder/utils/icon_manager.py
```diff
@@ -104,6 +104,7 @@
 }
 
 _qtaargs = {
+    'environment':             [('mdi.package-variant',), {'color': MAIN_FG_COLOR}],
```
ccordoba12: I agree with @jnsebgosselin here: let's change this to `mdi.cube-outline`.

goanpeca:
Made the fixes @ccordoba12 

Looks like: 

Screen Shot 2019-07-17 at 00 57 24

spyder/widgets/tests/test_status.py
```diff
@@ -40,5 +42,51 @@ def test_status_bar(status_bar, qtbot):
...
+    w = StatusBarWidgetTest(win, statusbar)
+
+    with qtbot.waitSignal(w.sig_clicked, timeout=1000):
+        qtbot.mouseRelease(w, Qt.LeftButton)
+
+    assert w.get_tooltip() == 'tooltip'
+    assert w.get_icon() == 'icon'
+
+
+def test_status_bar_conda_status(status_bar, qtbot, mocker):
```
ccordoba12: This is failing on Windows. Since the test seem to be adapted to Posix, you should skip it or rewrite to work on both Windows and Posix.

spyder/widgets/tests/test_status.py
```diff
@@ -40,5 +42,51 @@ def test_status_bar(status_bar, qtbot):
...
+    assert w.get_tooltip() == interpreter
+    assert 'base' in w._process_conda_env_info()
+
+    interpreter = '/miniconda/envs/foo/bin/python'
+    w.update_interpreter(interpreter)
+    assert w.get_tooltip() == interpreter
+    assert 'foo' in w._process_conda_env_info()
+
+
+def test_status_bar_no_conda_status(status_bar, qtbot, mocker):
```
ccordoba12: I guess this test will also fail on Windows. Please take a look at that or skip it (I really have no problem with skipping it, but it's your choice).

spyder/widgets/tests/test_status.py
goanpeca: I generalized the code to make it work on all platforms.

spyder/widgets/tests/test_status.py
goanpeca: This one works fine since we are checking the widget is basically "empty" but I added a code in the test to make it more clear.

goanpeca:
Made the fixes @ccordoba12

spyder/widgets/tests/test_status.py
ccordoba12: Nice, thanks!

PR #9782, opened 2019-07-08
PR: Allow for theme options to be updated without the need for a restart

goanpeca:
## Description of Changes

- By changing how the icon manager works from a bunch of funcs to a class that can update icons on demand, we can avoid having to restart spyder for changing the theming options.
- A new method was created on plugins `update_style` that needs to take care of updating the actions icons as needed.
- This way of handling icons also will allow for the registration of new icons so plugin developers can register icons before they use them (like we do with `'advanced': [('fa.gear',), {'color': MAIN_FG_COLOR}],`)
- This will make also the process of loading different configurations per project much easier as well as providing a much better user experience.

## TODO
- [x] Update toolbar icons
- [ ] Complete updating all actions for all plugins
- [x] Update icons for status bar widgets
- [ ] Update style for widgets like help plugin 
- [ ] Add icon registration method to the api

### Issue(s) Resolved

Fixes #

### Affirmation


goanpeca:
This is already making a lot of progress!

theme

goanpeca:
Also by having this now in place we can further simplify the appearance preferences

Screen Shot 2019-07-07 at 22 41 05

Instead of having the light/dark option, I suggest we define that as part of the syntax highlight. So solarized dark implies dark interface, etc... I think this makes it more in line with what other IDEs offer and avoids having to change 2 settings when switching from a dark to light theme and viceversa

ccordoba12:
> Instead of having the light/dark option, I suggest we define that as part of the syntax highlight. So solarized dark implies dark interface, etc... I think this makes it more in line with what other IDEs offer and avoids having to change 2 settings when switching from a dark to light theme and viceversa

That was CAM's suggestion and I think it's a good one. It allows users to select a dark color scheme with the light appearance theme and vice-versa.

Since that's already in place, I'd prefer to leave it.

goanpeca:
>Since that's already in place, I'd prefer to leave it.

Ok

spyder/api/plugins.py
```diff
@@ -472,6 +472,21 @@ def update_font(self):
         """
         pass
 
+    def update_style(self):
```
goanpeca: @ccordoba12 you agree with this name for updating icons and styles?

ccordoba12: Sure, that's fine.

ccordoba12:
Why was this closed @goanpeca?

goanpeca:
Not working on this anymore.

PR #9789, opened 2019-07-08
PR: Add package name to dependencies

### LINKED ISSUE #9714: Dependencies dialog should provide correct package name
For issues when users have an incorrect dependency in the case of the `pyls` package:

Screen Shot 2019-06-28 at 11 15 48

My immediate action is to do something like

```bash
$ conda install pyls -c spyder-ide
```

But this will not work as the actual package name is `python-language-server`. I see this happening to users trying to do `pip install pyls`

Could we either change the name of the package or add an extra column for package name? (which is not always the same as module name?)

---

Also we should be able to provide more help on the command users need to run to fix this

goanpeca:
## Description of Changes

Screen Shot 2019-07-08 at 17 13 56

### Issue(s) Resolved

Fixes #9714

### Affirmation

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -273,7 +273,8 @@ qInstallMessageHandler(qt_message_handler)
 # Dependencies
 # =============================================================================
 QDARKSTYLE_REQVER = '>=2.6.4'
-dependencies.add("qdarkstyle", _("Dark style for the entire interface"),
+dependencies.add("qdarkstyle", "qdarkstyle",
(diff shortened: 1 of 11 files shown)


goanpeca:
@jnsebgosselin I liked your suggestion for adding a URL but that would require adding an HTML delegate, so for now I just added the extra column which should provide (hopefully) enough information to devs using `master`

goanpeca:
The pep8 speaks warning seemed stuck. (I fixed the issue it reported)

PR #9791, opened 2019-07-08
PR: Add different icon color for style warninngs

### LINKED ISSUE #5489: Different Icon for code style warning
## Issue

Code Warnings and Style Warnings look the same.

## Solution

Different Icon for Syle Warnings. Maybe a blue square with an 8 or an S.

goanpeca:
## Description of Changes

Style warnings are now treated as another type of warning (with lesser severity) and a different icon.

warnings

### Issue(s) Resolved

Fixes #5489

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -2000,6 +2000,16 @@ class CodeEditor(TextEditBaseWidget):
         self.tooltip_widget.hide()
         self.clear_extra_selections('code_analysis_highlight')
 
+    def get_style_warning_sources(self):
+        """
+        Return sources to take into account for checking style warninigs.
+        """
+        if self.language.lower() == 'python':
+            style_warning_sources = ['pycodestyle', 'pydocstyle']
+        else:
+            style_warning_sources = []
+        return style_warning_sources
+
     def show_code_analysis_results(self, line_number, block_data):
         """Show warning/error messages."""
         from spyder.config.base import get_image_path
@@ -2010,7 +2020,8 @@ class CodeEditor(TextEditBaseWidget):
             DiagnosticSeverity.INFORMATION: 'information',
             DiagnosticSeverity.HINT: 'hint',
         }
-
+        style_icon = 'style-warning'
+        style_sources = self.get_style_warning_sources()
         code_analysis = block_data.code_analysis
 
         # Size must be adapted from font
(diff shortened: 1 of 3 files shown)


goanpeca:
@bcolsen for now I used the same icon but with a different color.

@spyder-ide/core-developers open for suggestions on which icon should be. Thoughts @jnsebgosselin?

bcolsen:
The same icon but a different color will work. Currently I think that special comments (#TODO and such) are blue `!`. So maybe magenta?

Are you going to have the scroll bar flags match the color as well?

goanpeca:
>Are you going to have the scroll bar flags match the color as well?

Es bu on a separate PR

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -2000,6 +2000,16 @@ def hide_tooltip(self):
         self.tooltip_widget.hide()
         self.clear_extra_selections('code_analysis_highlight')
 
+    def get_style_warning_sources(self):
```
oscargus: Could this possibly be moved to a separate file/static method? (Maybe that is what you are referring to in the comments.) It would be useful to be able to obtain this for #9836. In fact, one may even include
https://github.com/spyder-ide/spyder/blob/c188748789c8b069b5fc383061219f875fe36959/spyder/plugins/editor/widgets/codeeditor.py#L1971-L1977
there (see end of #9924), and possibly a convenience function for getting the correct icon providing DiagnosticSeverity and source (not sure if it is enough to instantiate them once when it comes to style changes etc)?

oscargus: Also, the menu can use the same function, which I do not think it does at the moment.

goanpeca: Yes @oscargus there will be a single place to store this info, not related to this widget. But that requires finishing another PR, which will not happen for at least 3 weeks. 

I need to work on something else and then I will tackle that and come back to this ;-).

goanpeca: >(not sure if it is enough to instantiate them once when it comes to style changes etc)

#9782 will allow to have style changes without a restart, so some methods will need to be called again if style changes. Since the theme can have a "dirty" flag if style change is requested, then we can have a cache of the style that is refreshed on style change to avoid recreating objects like icons, although for what I have seen Icons are relatively lightweight.

goanpeca:
@ccordoba12 this one is a pretty "simple" change. that could readily go for 4.0rc

goanpeca:
@jnsebgosselin what do you think we should use as different colors?

PR #9792, opened 2019-07-08
PR: Preserve selected working directory in Preferences after a restart

### LINKED ISSUE #9561: Selected working directory in Preferences is not preserved after a restart

under preferance i chose a working directory as a start up directrory when starting spyder again reverts back to c:\users\user directory

## Versions

* Spyder version: 4.0.0b2 
* Python version: 3.7.3
* Qt version: 5.9.6
* PyQt5 version: 5.9.2

### Dependencies

```
qdarkstyle >=2.6.4 :  2.6.8 (OK)
qtconsole >=4.2.0  :  4.4.3 (OK)
IPython >=4.0      :  7.4.0 (OK)
```

steff456:
## Description of Changes

Add options to configurations file and check the working directory preferences.

### Issue(s) Resolved

Fixes #9561 

### Affirmation


spyder/config/main.py
```diff
@@ -307,6 +307,9 @@
               'console/use_project_or_home_directory': False,
               'console/use_cwd': True,
               'console/use_fixed_directory': False,
+              'console/fixed_directory': '',
+              'startup/fixed_directory': '',
```
goanpeca: @steff456 since these values are empty we probably do not need to add them here
```
              'console/fixed_directory': '',
              'startup/fixed_directory': '',

```

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -896,6 +896,8 @@ def register_client(self, client, give_focus=True):
             if (self.main.projects is not None and
                     self.main.projects.get_active_project() is not None):
                 cwd_path = self.main.projects.get_active_project_path()
+        elif CONF.get('workingdir', 'startup/use_fixed_directory'):
+            cwd_path = CONF.get('workingdir', 'startup/fixed_directory')
```
goanpeca: Since this value may not exist this call should probably include a default

`CONF.get('workingdir', 'startup/fixed_directory', default=somethig?)`

ccordoba12:
@steff456, please merge with master so you can use the work done by @goanpeca in PR #9741 for your test (that'd be the right thing to do, right @goanpeca?)

goanpeca:
Yes, that can be used to check this. @steff456 ping me if you need some help.

spyder/plugins/workingdirectory/tests/test_workingdirectory.py
```diff
@@ -7,14 +7,32 @@
...
+    workingdirectory.show()
+
+    return workingdirectory, qtbot
+
+@pytest.fixture
+def setup_workingdirectory_startup(qtbot):
+    new_wdir = 'new_workingdir/'
+    if not osp.exists(new_wdir):
+        os.mkdir(new_wdir)
+    """Setup working directory plugin."""
```
goanpeca: @steff456 this docstring might need to be move upwards?

spyder/plugins/workingdirectory/tests/test_workingdirectory.py
```diff
@@ -7,14 +7,32 @@
...
+    workingdirectory = WorkingDirectory(None)
+    qtbot.addWidget(workingdirectory)
+    workingdirectory.show()
+
+    return workingdirectory, qtbot
+
+@pytest.fixture
+def setup_workingdirectory_startup(qtbot):
+    new_wdir = 'new_workingdir/'
```
goanpeca: Maybe it is better to define this as a CONSTANT outside so that it can be used later at the 


```
CONSTANT = 'new_workingdir/'
...

assert act_wdir == CONSTANT
```

spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py
```diff
@@ -92,6 +92,13 @@ def __getattr__(self, attr):
     # Tests assume inline backend
     CONF.set('ipython_console', 'pylab/backend', 0)
 
+    # Start in a new working directory the console
+    new_wdir = 'new_workingdir/'
```
goanpeca: Maybe it is better to define this as a CONSTANT outside so that it can be used later at the
```
CONSTANT = 'new_workingdir/'
...

assert act_wdir == CONSTANT
```

ccordoba12: This also needs to be handled with a pytest marker because it's modifying the fixture. Please see how we handle the `use_introspection` marker in the `main_window` fixture, present in `spyder/app/tests/test_mainwindow.py`.

spyder/plugins/workingdirectory/tests/test_workingdirectory.py
ccordoba12: I don't think you need a new fixture here, just adding a pytest marker to modify the current one, as I described it above for the IPython console fixture.

steff456:
The tests that are failing in azure are due to one test of the LSP, not the ones that were created in this PR.

ccordoba12:
@steff456, please merge with the latest master to fix those problems. Things should be better after PR #9798.

spyder/plugins/workingdirectory/tests/test_workingdirectory.py
```diff
@@ -16,7 +16,7 @@
 from spyder.config.main import CONF
 
 
-NEW_DIR = 'new_workingdir'
+NEW_DIR = 'new_workingdir´'
```
goanpeca: is this a typo @steff456 ?

steff456: Yes! Thanks!!!

spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py
```diff
@@ -92,6 +93,19 @@ def __getattr__(self, attr):
     # Tests assume inline backend
     CONF.set('ipython_console', 'pylab/backend', 0)
 
+    # Start in a new working directory the console
+    try:
+        use_startup_wdir = request.node.get_marker('use_startup_wdir')
```
goanpeca: I have question @ccordoba12 is this standard way for marker use? I was trying to look for documentation on pytest but found this instead

https://docs.pytest.org/en/latest/historical-notes.html#update-marker-code

It might not affect us or maybe it does. it seems the way to parametrize a fixture is with the use of [indirect](https://docs.pytest.org/en/latest/example/parametrize.html)

ccordoba12: > is this standard way for marker use?

Probably things changed between Pytest 3 (where we started to use markers) and 4+ (where there were big changes in pytest). So we should use the latest way of using markers, I'd say.

> it seems the way to parametrize a fixture is with the use of indirect

But that's to use two versions of the same fixture for every test that calls it, which I think is not the case here, right?

goanpeca: The documentation in pytest is horrible, let me write an example to explain how parametrize with indirect works

steff456: should I try to modify the fixture with pytest.mark.parametrize?

goanpeca: No, no need. I just wanted to ask about this. Ignore these comments Tefa, I will open and issue to keep track of this.

spyder/config/main.py
```diff
@@ -307,6 +307,7 @@
               'console/use_project_or_home_directory': False,
               'console/use_cwd': True,
               'console/use_fixed_directory': False,
+              'startup/use_fixed_directory': False,
```
ccordoba12: Why is this is necessary? If it's only required for the new tests, this should be removed from here.

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -901,6 +901,9 @@ def register_client(self, client, give_focus=True):
             if (self.main.projects is not None and
                     self.main.projects.get_active_project() is not None):
                 cwd_path = self.main.projects.get_active_project_path()
+        elif CONF.get('workingdir', 'startup/use_fixed_directory'):
+            cwd_path = CONF.get('workingdir', 'startup/fixed_directory',
+                                default=get_home_dir())
```
ccordoba12: Again, is this really necessary? I wouldn't like to add code here that's only needed by our tests.

spyder/plugins/workingdirectory/plugin.py
```diff
@@ -113,7 +108,21 @@ def __init__(self, parent, workdir=None, **kwds):
                                       _('Change to parent directory'),
                                       triggered=self.parent_directory)
         self.toolbar.addAction(parent_action)
-                
+
+    def get_workdir(self):
+        """Get current workdir from the CONF file."""
+        if self.get_option('startup/use_fixed_directory'):
+            workdir = self.get_option('startup/fixed_directory',
+                                      default='')
```
ccordoba12: If you use `default` here, there's no need to add `'startup/fixed_directory'` to `config/main.py`

spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py
```diff
@@ -92,6 +93,19 @@ def __getattr__(self, attr):
...
+    except AttributeError:
+        use_startup_wdir = False
+
+    if use_startup_wdir:
+        new_wdir = osp.join(os.getcwd, NEW_DIR)
```
ccordoba12: Tests are failing because you need to call `os.getcwd()` here.

spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py
ccordoba12: @steff456, please merge with master (after squashing some of your last commits because it seems you were tryong to fix tests with them). That will allow you to avoid using the try/except here (see PR #9818).

spyder/plugins/workingdirectory/tests/test_workingdirectory.py
```diff
@@ -7,14 +7,38 @@
...
+NEW_DIR = 'new_workingdir'
 
 
 @pytest.fixture
-def setup_workingdirectory(qtbot):
-    """Setupp working directory plugin."""
+def setup_workingdirectory(qtbot, request):
+    """Setup working directory plugin."""
+    try:
+        use_startup_wdir = request.node.get_closest_marker('use_startup_wdir')
```
ccordoba12: With `get_closest_marker` you don't need to enclose this in a try/except (see PR #9818).

Please also add `use_startup_wdir` tp the [pytest.ini](https://github.com/spyder-ide/spyder/blob/master/pytest.ini) markers, with a description.

spyder/plugins/ipythonconsole/plugin.py
steff456: That is the change that makes the console start in the `'startup/use_fixed_directory'`, before it wasn't connected and that's why the bug was present

spyder/config/main.py
steff456: It is required for correcting the bug

spyder/plugins/workingdirectory/plugin.py
steff456: this option is not included in `config/main.py`

spyder/config/main.py
ccordoba12: Ok, thanks for the clarification @steff456!

spyder/plugins/ipythonconsole/plugin.py
ccordoba12: Ok, no problem.

steff456:
This PR is working now!

PR #9818, opened 2019-07-12
PR: Increase maximum Pytest supported version

ccordoba12:
Now that Pytest is at version 5, I think it's time to support more recent versions of it.


ccordoba12:
@goanpeca, now we can use Pytest versions less than 5.0

PR #9836, opened 2019-07-16
PR: Added Todos and Warnings/Errors panes

### LINKED ISSUE #9564: Feature request: To Do and Error/Warning Panes
Currently, Spyder produces a list of lines marked with #TODO, which can be found under Source>Show todo list. Something similar exists for the errors/warnings. 

However, accessing this is a pain if you want to go through several, or don't know which one you need, since the menue closes after each selection (it also doesn't seem to behave properly if you select multiple Todos/Warnings after each other, but I haven't done systematic testing there). It's also not possible to keep them in view while working, which would be quite useful for a Todo-list.

I suggest creating new panes for the Todos and Errors/Warnings. The basic functionality would just be to show the already existing list, and allow jumping to the line on left click. The outline and Breakpoints panes implement this already. 

Nice additional features for the Todo pane would be deleting and editing from the pane, but that's secondary.

oscargus:
## Description of Changes

Added panes for Todos and (eventually) Warnings/Errors.

todo

~~Question: my impression is that the current trend is that all plugins are moving to the plugin directory (from widgets). These two panes can share the basic table and related code. Where should I put it? (Basically, most of the code now in todogui.py.)~~

To do:
 - [x]  Add warnings/error pane (name? "Warnings"?)
 - [x] Decrease padding between lines
 - [x] Scale icons properly (may improve with decreased padding)
 - [x] Add tests
 - [x] Make sure that the panes are updated when opening a new file (not only when changing tab...)
 - [x] Make warning type column sortable

warnings

### Issue(s) Resolved

Fixes #9564 

### Affirmation

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -1020,7 +1020,8 @@ class MainWindow(QMainWindow):
         # Load other plugins (former external plugins)
         # TODO: Use this bucle to load all internall plugins and remove
         # duplicated code
-        other_plugins = ['breakpoints', 'profiler', 'pylint']
+        other_plugins = ['breakpoints', 'profiler', 'todos', 'warnings',
+                         'pylint']
         for plugin_name in other_plugins:
             if CONF.get(plugin_name, 'enable'):
                 module = importlib.import_module(
@@ -2278,8 +2279,8 @@ class MainWindow(QMainWindow):
         order = ['editor', 'ipython_console', 'variable_explorer',
                  'help', 'plots', None, 'explorer', 'outline_explorer',
                  'project_explorer', 'find_in_files', None, 'historylog',
-                 'profiler', 'breakpoints', 'pylint', None,
-                 'onlinehelp', 'internal_console', None]
+                 'profiler', 'breakpoints', 'todos', 'warnings', 'pylint',
+                 None, 'onlinehelp', 'internal_console', None]
         for plugin in self.widgetlist:
             action = plugin._toggle_view_action
             action.setChecked(plugin.dockwidget.isVisible())
--- a/spyder/plugins/editor/utils/findtasks.py
+++ b/spyder/plugins/editor/utils/findtasks.py
@@ -29,5 +29,6 @@ def find_tasks(source_code):
         for todo in re.findall(TASKS_PATTERN, text):
             todo_text = (todo[-1].strip(' :').capitalize() if todo[-1]
                          else todo[-2])
-            results.append((todo_text, line + 1))
+            tag = todo[-2]
+            results.append((todo_text, line + 1, tag))
     return results
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -294,6 +294,14 @@ DEFAULTS = [
              {
               'enable': True,
               }),
+            ('todos',
+             {
+              'enable': True,
+              }),
+            ('warnings',
+             {
+              'enable': True,
+              }),
             ('pylint',
              {
               'enable': True,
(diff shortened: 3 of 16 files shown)


spyder/app/mainwindow.py
```diff
@@ -1012,9 +1012,9 @@ def create_edit_action(text, tr_text, icon):
         # Load other plugins (former external plugins)
         # TODO: Use this bucle to load all internall plugins and remove
         # duplicated code
-        other_plugins = ['breakpoints', 'profiler', 'pylint']
+        other_plugins = ['breakpoints', 'profiler', 'todos', 'pylint']
         for plugin_name in other_plugins:
-            if CONF.get(plugin_name, 'enable'):
+            if CONF.get(plugin_name, 'enable', True):
```
oscargus: This will be reverted. Forgot to change it back after testing...

spyder/plugins/editor/utils/findtasks.py
```diff
@@ -29,5 +29,7 @@ def find_tasks(source_code):
         for todo in re.findall(TASKS_PATTERN, text):
             todo_text = (todo[-1].strip(' :').capitalize() if todo[-1]
                          else todo[-2])
-            results.append((todo_text, line + 1))
+            tag = todo[-2]
+            results.append((todo_text, line + 1, tag))
+    print(results)
```
oscargus: This will be removed.

oscargus:
Spacing is improved:
image

However, it does not work consistently. Every now and then, the icons are larger... Not sure why nor how to fix it...

oscargus:
I've added tests and sorted out the icon scaling issues (I think), so this is ready for review. No test for the actual plugin. I browsed other plugins and I couldn't really find anything to test. Apart from the goto-functionality, but it appears that it doesn't work testing for.

Screenshot from 2019-07-31 11-57-03

goanpeca:
Thanks for working on this @oscargus 

----

@ccordoba12 my only concern here is that we are adding to many panels (2 more panels) and I feel that is making the UI to cluttered. 

I feel like bookmarks, TODOs, Warnings all are doing something similar and could be bundled into a single panel (with perhaps tabs for each kind? or a combobox selector?)

oscargus:
Pane with tabs or tabs inside pane? Any major difference? ;-)

I think the main concern is maybe how to distribute the plugins in the different layouts. Right now all `other_plugins = ['breakpoints', 'profiler', 'todos', 'warnings', 'pylint']`plus third-party goes in the same tabwidget by design.

For the default Spyder layout together with 
`[help_plugin, explorer_variable, plots, helper, explorer_file, finder] + plugins`
so from 8 to 10 (soon 11 hopefully) in the top one. Compared to 2, maximum 3, in the bottom one.

As they can be disabled, I do not really see it as a major issue. Just show the ones you fancy. Personally, I think it often makes sense to have both Todo and Warnings open at the same time, so I would prefer that single use is not enforced. (Now, I have a desktop of 4960 x 1920 over three screens at my office, but also on a laptop it works well...)

However, ideally it should be possible to move these between tabwidgets or at least one should try to add them in a sensible way to start with. Maybe beta 5 is a good time to look this through and also clean up the plugin loading so that all built-in plugins can be assigned freely in the layout? Once these new plugins are introduced people will get used to the location and so on.

oscargus:
For the latter, I think creating a dict `plugins` in MainWindow, where all loaded (default) plugins are stored by name is a rather good way to go. Instead of having member variables for some of them and a list with the rest.

This would also allow getting rid of the duplicated initialization code for most of the plugins (some still require special treatment a quick look gives).

ccordoba12:
I agree with @goanpeca here: bookmarks, TODOs and warnings should be combined in a single plugin, with a combobox and a QStackedWidget to select and show one of them at a time, respectively.

oscargus:
OK. I'll close this then.

PR #9838, opened 2019-07-16
[WIP] PR: Cleanup of TextHelper

oscargus:
## Description of Changes
https://github.com/spyder-ide/spyder/blob/54f8303cfd936d5821f1cfded5ede2184e66d596/spyder/plugins/editor/utils/editor.py#L142-L143

Refactored code and moved all used methods from TextHelper to CodeEditor. Added a method `ensure_single_empty_ending_line` which can be used to add a method for automatic single empty line upon saving (the other part of the previous method, removing trailing spaces, is implemented elsewhere).

All remaining methods in TextHelper are currently unused. Some of them can make sense to include (I think I may initially have moved a few methods to CodeEditor that are not used, but seems obviously useful). 

 - [ ] Decide which more functions to include in CodeEditor
 - [ ] Remove TextHelper completely
 - [ ] Merge `go_to_line` and `goto_line` into a single function

(Also removed some unused code, which was the original purpose of the PR...)

### Affirmation

### DIFF
--- a/spyder/plugins/editor/extensions/tests/test_closebrackets.py
+++ b/spyder/plugins/editor/extensions/tests/test_closebrackets.py
@@ -12,7 +12,6 @@ from qtpy.QtGui import QTextCursor
 # Local imports
 from spyder.utils.qthelpers import qapplication
 from spyder.plugins.editor.widgets.codeeditor import CodeEditor
-from spyder.plugins.editor.utils.editor import TextHelper
 from spyder.plugins.editor.extensions.closebrackets import (
         CloseBracketsExtension)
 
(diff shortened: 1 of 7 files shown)


goanpeca:
Thanks for working on this @oscargus 

@ccordoba12 should this have been moved to the mixin instead of the code editor?

ccordoba12:
@oscargus, the idea is **not** to remove the duplicated code from TextHelper and move it back to CodeEditor, but the other way around.

Therefore I'm -1 on this PR and I won't merge it was it is, sorry.

oscargus:
> @oscargus, the idea is **not** to remove the duplicated code from TextHelper and move it back to CodeEditor, but the other way around.

Ahh, my bad. I'll update the comment to make it more clear. :-)


> Therefore I'm -1 on this PR and I won't merge it as it is, sorry.

Fully understandable! Let us see if I can do some work in the right direction. I have some better understanding now if nothing else.

PR #9852, opened 2019-07-17
PR: Rewrite test to make it pass (on mac)

goanpeca:
## Description of Changes

### Issue(s) Resolved

The way this test was written was failing on my mac all the time. I rewrote it in a slightly different way and it passes what I believe is the actual thing to be tested. Is this the case? See gif:

warnings

### Affirmation


spyder/plugins/editor/widgets/tests/test_warnings.py
```diff
@@ -36,8 +37,20 @@ def test_ignore_warnings(qtbot, lsp_codeeditor):
 
     CONF.set('lsp-server', 'pydocstyle/ignore', 'D100')
     CONF.set('lsp-server', 'pycodestyle/ignore', 'E261')
+
+    # After this call the manager needs to be reinitialized
     manager.update_server_list()
-    qtbot.wait(2000)
+
+    test_ci = os.environ.get('CI_PYTEST', os.environ.get('CI', None))
```
goanpeca: @ccordoba12 @andfoy I really do not know why I need this locally :-| 🤷‍♂

spyder/plugins/completion/languageserver/client.py
```diff
@@ -207,6 +208,12 @@ def patched_cleanup():
         new_env = dict(os.environ)
         python_path = os.pathsep.join(sys.path)[1:]
         new_env['PYTHONPATH'] = python_path
+
+        # This allows running the tests directly without having to install
+        # spyder
```
ccordoba12: spyder -> Spyder

spyder/plugins/completion/languageserver/client.py
```diff
@@ -207,6 +208,12 @@ def patched_cleanup():
         new_env = dict(os.environ)
         python_path = os.pathsep.join(sys.path)[1:]
         new_env['PYTHONPATH'] = python_path
+
+        # This allows running the tests directly without having to install
```
ccordoba12: ... running ~~the~~ *LSP* tests ...

spyder/plugins/editor/widgets/tests/test_warnings.py
ccordoba12: There's not a single instance of `CI_PYTEST` in our codebase, so I don't understand where you got that name. So I think you should remove this and simply test for the `CI` variable below.

spyder/plugins/editor/widgets/tests/test_warnings.py
```diff
@@ -36,8 +37,20 @@ def test_ignore_warnings(qtbot, lsp_codeeditor):
...
+    if not test_ci and sys.platform == 'darwin':
```
ccordoba12: Please change this line to

    if os.environ.get('CI', None) is None and sys.platform == 'darwin':

spyder/plugins/editor/widgets/tests/test_warnings.py
goanpeca: I was looking for CI variables and found this -> https://github.com/spyder-ide/spyder/blob/master/.travis.yml#L9

spyder/plugins/editor/widgets/tests/test_warnings.py
goanpeca: https://github.com/spyder-ide/spyder/blob/master/.travis.yml#L9  Is there a Global environment variable that is defined on all CI tests? I looked at the config on travis yaml and azure.yaml adn got it from there.

goanpeca:
Made the suggested changes @ccordoba12.

>@goanpeca, great job in making all the LSP related tests run independently!! It was very frustrating to not be able to do that!

Yes this was very annoying 😅!

spyder/plugins/editor/widgets/tests/test_warnings.py
ccordoba12: Ok, I see. I introduced that variable to share the same bash scripts between Travis and Circle but now it's not needed, so you can remove it.

ccordoba12:
@goanpeca, do you prefer to remove `CI_PYTEST` as part of this PR? Or do you prefer to leave it for later?

goanpeca:
Will do it now :-) gimme a sec I push the change.

PR #9857, opened 2019-07-18
PR: Fix control-click URL in tutorial ValueError

### LINKED ISSUE #9849: Control-click URL in tutorial causes ValueError

In the Spyder tutorial, the 'Recommended first steps for Python beginners' section, control-left click on the 'IPython interpreter' link.

I did this because I thought it might open the URL in a new window.  It opens the URL successfully (in my web browser, same as a normal left-click), but it issues the "Spyder has encountered an internal problem!" bug report page.

I'm using windows 10, 64-bit.  Default browser is Firefox.

### Traceback
```python-traceback
  File "C:\Users\mckay\Anaconda3\lib\site-packages\spyder\widgets\browser.py", line 168, in createWindow
    webbrowser.open(to_text_string(self.url().toString()))
  File "C:\Users\mckay\Anaconda3\lib\webbrowser.py", line 78, in open
    if browser.open(url, new, autoraise):
  File "C:\Users\mckay\Anaconda3\lib\webbrowser.py", line 582, in open
    os.startfile(url)
ValueError: startfile: filepath too long for Windows
```

## Versions

* Spyder version: 3.3.1 
* Python version: 3.7.0
* Qt version: 5.8.0
* PyQt5 version: 5.8.2

### Dependencies

```
sphinx >=0.6.6    :  1.7.9 (OK)
jedi >=0.9.0      :  0.12.1 (OK)
numpy >=1.7       :  1.15.1 (OK)
sympy >=0.7.3     :  1.1.1 (OK)
qtconsole >=4.2.0 :  4.4.1 (OK)
IPython >=4.0     :  6.5.0 (OK)
matplotlib >=2.0.0:  3.0.3 (OK)
```

steff456:
## Description of Changes

Add a try catch to prevent the ValueError in tutorial

### Issue(s) Resolved

Fixes #9849

### Affirmation

### DIFF
--- a/spyder/widgets/browser.py
+++ b/spyder/widgets/browser.py
@@ -168,6 +168,7 @@ class WebView(QWebEngineView):
     #------ QWebEngineView API -------------------------------------------------------
     def createWindow(self, webwindowtype):
         import webbrowser
+        # See: spyder-ide/spyder#9849
         try:
             webbrowser.open(to_text_string(self.url().toString()))
         except ValueError:
--- a/spyder/widgets/browser.py
+++ b/spyder/widgets/browser.py
@@ -167,8 +167,8 @@ class WebView(QWebEngineView):
 
     #------ QWebEngineView API -------------------------------------------------------
     def createWindow(self, webwindowtype):
+        import webbrowser
         try:
-            import webbrowser
             webbrowser.open(to_text_string(self.url().toString()))
         except ValueError:
             pass
--- a/spyder/widgets/browser.py
+++ b/spyder/widgets/browser.py
@@ -167,8 +167,11 @@ class WebView(QWebEngineView):
 
     #------ QWebEngineView API -------------------------------------------------------
     def createWindow(self, webwindowtype):
-        import webbrowser
-        webbrowser.open(to_text_string(self.url().toString()))
+        try:
+            import webbrowser
+            webbrowser.open(to_text_string(self.url().toString()))
+        except ValueError:
+            pass
 
     def contextMenuEvent(self, event):
         menu = QMenu(self)
--- a/spyder/widgets/tests/test_browser.py
+++ b/spyder/widgets/tests/test_browser.py
@@ -10,9 +10,9 @@ Tests for browser.py
 
 # Test library imports
 import pytest
-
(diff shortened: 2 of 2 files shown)


ccordoba12:
> We could probably leave the import outside the try/except as the thing causing problems is the call to open

I agree with this.

> We should probably add a regression test for this.

I don't think this is necessary. Making a test for could be very time consuming and there's very little to gain with it.

goanpeca:
>I don't think this is necessary. Making a test for this could be very time consuming and there's very little to gain with it.

The test can be as simple as this:

```python
def test_webview_open_regression(qtbot):
    # See: spyder-ide/spyder#....
    wb = WebView()
    qtbot.addWidget(wb)
    wb.setUrl('problem_url')
    wb.createWindow(webwindowtype)
```

Without the fix it should fail (because of the uncaught exception)

ccordoba12:
@goanpeca, your proposed test checks at the end the functionality of the `webbrowser` module, which I'm sure is tested elsewhere.

So I'd say we should skip testing this and move to more important things. I think we really don't need to invest time to test every minor change we do.

goanpeca:
>So I'd say we should skip testing this and move to more important things. I think we really don't need to invest time to test every minor change we do.

Sure that is fine, @ccordoba12 .

@steff456 please undo the commit that include the testing and add a comment to point to the issue (added a comment).

---

One suggestion though, we should merge these type of PRs right away, since "waiting for tests to pass" does not really check anything.

spyder/widgets/browser.py
```diff
@@ -168,7 +168,10 @@ def zoom_in(self):
     #------ QWebEngineView API -------------------------------------------------------
     def createWindow(self, webwindowtype):
         import webbrowser
-        webbrowser.open(to_text_string(self.url().toString()))
+        try:
```
goanpeca:
```suggestion
        # See: spyder-ide/spyder#9849
        try:
```

ccordoba12:
> One suggestion though, we should merge these type of PRs right away, since "waiting for tests to pass" does not really check anything.

I think if we merge right away PRs will be flagged as failing, and I wouldn't like to give that impression to people following the project.

I'll test it with this one to see if that's the case or not.

PR #9869, opened 2019-07-20
PR: Improve unindent behavior

### LINKED ISSUE #5734: Shift-Tab does not remove final indent level for non 4-multiple indents (e.g. continuations)
## Description

If I have the cursor on an indented line or (particularly troublesome) a block of code selected, that doesn't have indent perfectly a multiple of 4, e.g. continuation lines/blocks, Shift-Tab will not remove the last indent level of >4 spaces and instead do nothing. Instead, I must go to every line and delete the last 1-3 spaces manually, which is rather tedious and error prone and takes far more time than tapping shift-tab one more time. A relatively common case is when a block picks up an undesired indent from the block above, or I otherwise want to shift it all the way to the left quickly, and I cannot. I don't see an immediate reason why this wouldn't be the desired behavior, as it is in most (most all?) other editors and applications, but perhaps there is some use case I'm missing.

1. Copy a block of code with a non mod 4 = 0 indent. E.g.
```
       (1.2e-5, 2.0e8),
       (1.6e-5, 2.3e8),
       (1.8e-5, 1.5e7)]
```
2. Select the block (or place the cursor at the beginning of a line)
3. Hit shift-tab until it stops before the left margin

For the above example
Expected:
```
(1.2e-5, 2.0e8),
(1.6e-5, 2.3e8),
(1.8e-5, 1.5e7)]
```
Actual:
```
   (1.2e-5, 2.0e8),
   (1.6e-5, 2.3e8),
   (1.8e-5, 1.5e7)]
```

As usual, occurred on multiple systems, installs and OS versions so clearly not specific to that.

* Spyder Version: 3.2.4 
* Python Version: 3.6.3
* Qt Versions: 5.6.2, PyQt5 5.6 on Windows 8.1 Pro x64 and Win 10 Ent x64

## Dependencies
```

```

oscargus:
## Description of Changes

Improved the unindent behavior as suggested in https://github.com/spyder-ide/spyder/issues/5734#issuecomment-346209380

For a multiline selection, the lowest number of spaces is used

So assuming four spaces indentation level:
```
5 spaces
9 spaces
```

becomes:
```
4 spaces
8 spaces
```

Also,
```
5 spaces 
6 spaces
```

becomes:
```
4 spaces 
5 spaces
```

which next time becomes:
```
0 space
1 space
```

and stays as
```
0 space
1 space
```

Not sure what the actually expected behavior is here. One may consider first time aligning all rows to the next lower spacing multiple, so 
```
5 spaces
6 spaces
```

becomes
```
4 spaces
4 spaces
```

That require a bit more work though and I do not know if there are any obvious drawbacks with it.

Also added tests for unindent in general, which I do not think was there before. (Which lead to the question if tab and shift+tab are always reciprocal, I think not.) And rewrote the indent test code for easier copy-and-paste...

### Issue(s) Resolved

Fixes #5734

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -2643,7 +2643,21 @@ class CodeEditor(TextEditBaseWidget):
         force=True: unindent even if cursor is not a the beginning of the line
         """
         if self.has_selected_text():
-            self.remove_prefix(self.indent_chars)
+            if self.indent_chars == "\t":
+                # Tabs, remove one tab
+                self.remove_prefix(self.indent_chars)
+            else:
+                # Spaces
+                space_count = len(self.indent_chars)
+                leading_spaces = self.__spaces_for_prefix()
+                remainder = leading_spaces % space_count
+                if remainder:
+                    # Get block on "space multiple grid".
+                    # See spyder-ide/spyder#5734.
+                    self.remove_prefix(" "*remainder)
+                else:
+                    # Unindent one space multiple
+                    self.remove_prefix(self.indent_chars)
         else:
             leading_text = self.get_text('sol', 'cursor')
             if force or not leading_text.strip() \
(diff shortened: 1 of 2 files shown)


ccordoba12:
> That require a bit more work though and I do not know if there are any obvious drawbacks with it.

I think what you did for now is enough, thanks!

> Also added tests for unindent in general, which I do not think was there before

Even better!! Thanks a lot for that too!

PR #9872, opened 2019-07-20
[TEST] Python 3.8 build on Travis

oscargus:
## Description of Changes
Trying to see what happens when running the tests using 3.8-dev on Travis. Built on #9870 to increase the chance of success.

Not really meant for merging, but if #9870 is merged and this passes, so, well...

### Affirmation

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,7 @@ jobs:
             sudo apt-get -qq update
             sudo apt-get install libegl1-mesa
             sudo apt-get install libzmq3-dev
+            sudo apt-get -y --reinstall install libxcb*
       - run:
           command: sudo ./continuous_integration/travis/install.sh
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,6 +12,7 @@ jobs:
             sudo apt-get -qq update
             sudo apt-get install libegl1-mesa
             sudo apt-get install libzmq3-dev
+            sudo apt-get install libxcb
       - run:
           command: sudo ./continuous_integration/travis/install.sh
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,51 +1,18 @@
 version: 2
 
-main: &main
-    machine: true
+jobs:
+  build:
+    docker:
+      - image: circleci/python:3.8.0b2
     steps:
       - checkout
-      - run:
-          command: docker pull dorowu/ubuntu-desktop-lxde-vnc:xenial
       - run:
           name: Install system packages
           command: |
             sudo apt-get -qq update
             sudo apt-get install libegl1-mesa
+            sudo apt-get install libzmq3-dev
       - run:
-          command: ./continuous_integration/travis/install.sh
+          command: sudo ./continuous_integration/travis/install.sh
       - run:
-          command: ./continuous_integration/circle/modules_test.sh || ./continuous_integration/circle/modules_test.sh
-
-jobs:
-  python2.7:
-    <<: *main
-    environment:
-      - PYTHON_VERSION: "2.7"
-      - USE_CONDA: "yes"
-
-  python3.5:
-    <<: *main
-    environment:
-      - PYTHON_VERSION: "3.5"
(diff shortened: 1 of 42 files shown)


ccordoba12:
This is not the right way to test Python 3.8. Please undo your changes to Circle and Azure, and simply add a new entry to Travis for Python 3.8 to be tested with pip.

I don't see the need of commenting tests for the rest of the Python versions.

oscargus:
This "PR" has got a bit out of hand... :-) So now, it is more an experiment of what happens with 3.8 than something that can be merged.

I expect two outcomes: 

1) A proper PR for 3.8 testing.
It is currently not possible to do a conda install using 3.8. There is a single version of 3.8a3 in conda-forge, but there are many packages that are not set up properly, so I cannot really see that as a way forward using conda (hence, it may not be a good idea to try, let us see how far I can get).

For Travis, there is a 3.8-dev installed, but then pip must be used. For some reason, zmq does not build properly using the standard pip install with dependencies, so I had to do it like this just to get it up and running. It may be that I learnt some things so that I can get it back to a somewhat more standard install variant. Still, it is not possible to introduce conda at the moment (as far as I can tell at least).

For CircleCI, there are lots of different 3.8-versions available. There, the main problem at the moment seems to be that there are fewer packages installed by default. Again, there is a problem with zmq, so cython must be installed explicitly. There are also some issues with the graphical interface and as I cannot use the xenial-vnc-docker (as that doesn't have 3.8 and conda doesn't have 3.8), I'm trying out some other things there.

Still, using this rather brute-force approach, I still manage to run it to some extent, which leads to:

2) A PR with 3.8-related fixes
Some things changes with 3.8. Primarily, implicit conversions to int (from float or different Qt flags) are deprecated, so quite a lot of deprecation warnings show up (hundreds of them). Another thing that confused me first was that the completion was tested with math.degree. In 3.8 there is another method starting with d (dist), so I changed to hypot (which is the only one starting with h). However, it also tuned out that the signature changed from `hypot(x, y)` to `hypot(*coordinates)` (which I think is not valid 2.7 Python...). Anyway, I have these changes in separate commits so that I (hopefully) can cherry pick them eventually.

The reason for skipping the other versions is simply to save server time. As the scripts are so messed up at the moment, there is no point in even trying to run them every time.

oscargus:
There are about 40.000 deprecation warnings in the last run (where at least some were already removed). Will update after

>     1. (hundreds of them)

There are about 40.000 deprecation warnings it turns out (not that many positions to change, but the number of rows in the warnings section). And I had already removed some of them at that stage. Hopefully they have reduced significantly in the current build. Although some of them come from other packages.

oscargus:
Got rid of about 17.000 warnings and most of the remaining comes from other packages. The ones I cannot really figure out that are Spyder related are:

  /home/travis/build/spyder-ide/spyder/spyder/plugins/plots/widgets/figurebrowser.py:534: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
    self.sig_zoom_changed.emit(self.get_scaling())

sig_zoom_changed is defined as `Signal(float)`, so not clear.

  /home/travis/build/spyder-ide/spyder/spyder/app/mainwindow.py:1808: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
    self.resizeDocks(_widgets, width_fractions, Qt.Horizontal)
  /home/travis/build/spyder-ide/spyder/spyder/app/mainwindow.py:1815: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
    self.resizeDocks(_widgets, height_fractions[idx], Qt.Vertical)

resizeDocks should have a (list of) integer as second argument, while the current ones are fractions of the total window size. However, since that works, I cannot really understand how this is supposed to work.

Will try out a bit more and eventually provide at least the warnings removal PR.

oscargus:
The code modifications are in #9879 I doubt that I can get a reasonable build environment up and running at the moment, so I guess it is better to wait until 3.8 is available in conda.

PR #9879, opened 2019-07-21
PR: Modifications for easier compatibility with Python 3.8

oscargus:
## Description of Changes

Some preparations for Python 3.8. The primary changes are caused by two reasons:

1) New methods in `math` lead to that some completion tests are changed. For example, earlier there was only one method starting with d, now there are two, making it less suitable for what it was used for.

2) Implicit casting to int is deprecated. This hold primarily for floats, but also for some Qt-constants/flags. In the code, I used a mix of int, round, and ceil depending on what I thought made sense for the use case. int is clearly the obvious replacement, but when it comes to sizing things, the others may makes more sense. I have not seen any drastic changes (if any).

This takes care of all spyder-related casting warnings, except for the ones in #9878 

There are still some test failures remaining, but they are hard to debug since it is not clear if they depend on the build environment or not (some for sure do). It seems like the object explorer has a different number of elements in 3.8 compared to e.g. 3.7 (there is a comment in the code that 2.7 also has that). Not clear why and how, and will be easier to figure out when being able to run it locally. The major obstacle seems to be warnings widget, one test crashed, but possibly this has to do with the setup of the test machine.

### Affirmation


ccordoba12:
@oscargus, please merge with master to fix the merge conflict.

Also, is this ready to be merged or not?

ccordoba12:
@oscargus, I updated this with master and I'll merge if our CIs pass. Hope that's ok for you.

PR #9895, opened 2019-07-24
PR: Disable automatic completions when spacebar or backspace are pressed

### LINKED ISSUE #9871: New automatic completion makes it hard to work...
## Problem Description
While I in many ways appreciate the new automatic completion stuff, it also introduced a few challenges. Two which I have isolated and lists here.

It is crucial that Ctrl+s is passed through to the editor. It has happened quite a few times that I have edited a file and pressed Ctrl+s, just to later find out that the file is still marked as modified. (One reason for the many test runs for #9870...)

The behavior below is a bit annoying. What happens is that I press backspace one time to many (in this case it is a Mickey Mouse example) and press return as I would like to insert a line break here again. Not sure what the proper solution is here though (except that I am surprised that the completions start with the first letter of the, earlier, next line, which doesn't really make sense).

autocompletion

andfoy:
## Description of Changes
Disable automatic completions when whitespace or backspace characters are introduced, as they cause an unpleaseant experience and a one-size-fits-all solution is not clear.

### Issue(s) Resolved

Fixes #9871 

### Affirmation

@andfoy

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -3356,19 +3356,8 @@ class CodeEditor(TextEditBaseWidget):
             word_text = to_text_string(cursor.selectedText())
             # Perform completion on the fly
             if self.automatic_completions:
-                if text == '\b' or text == ' ':
-                    offset = 1 + (text == ' ')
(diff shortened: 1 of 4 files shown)


oscargus:
Thanks @andfoy!

PR #9913, opened 2019-07-29
PR: Add unambiguous file names to Breakpoints pane

oscargus:
## Description of Changes

When having breakpoints in files with the same name, it was hard to tell the difference. (Although I realized at a rather later stage that there was a tooltip hover on the file name with the full path.) This PR adds unambiguous file names to the Breakpoints pane.

breakpoints

It should also be slightly more efficient as the name is only computed once per data update.

I also added/corrected an icon for the Breakpoints pane. After some deliberation, I thought that mdi-dots-vertical probably was the best I could find. Three stacked dots.

Earlier, it was the profiler icon, although it wasn't shown in the menu.

In addition, I right-aligned the line numbers (not shown in the screen shot), and refactored the code a bit, using constants for the different columns.

### Affirmation

### DIFF
--- a/spyder/plugins/breakpoints/plugin.py
+++ b/spyder/plugins/breakpoints/plugin.py
@@ -49,12 +49,12 @@ class Breakpoints(SpyderPluginWidget):
 
         path = osp.join(self.PLUGIN_PATH, self.IMG_PATH)
         self.icon = ima.icon('breakpoints', icon_path=path)
-    
+
     #------ SpyderPluginWidget API --------------------------------------------
     def get_plugin_title(self):
         """Return widget title"""
         return _("Breakpoints")
-    
+
     def get_plugin_icon(self):
         """Return widget icon"""
         return self.icon
@@ -85,7 +85,7 @@ class Breakpoints(SpyderPluginWidget):
         self.add_dockwidget()
 
         list_action = create_action(self, _("List breakpoints"),
-                                   triggered=self.show, icon=self.icon)
+                                    triggered=self.show, icon=self.icon)
         list_action.setEnabled(True)
         pos = self.main.debug_menu_actions.index('list_breakpoints')
         self.main.debug_menu_actions.insert(pos, list_action)
--- a/spyder/plugins/breakpoints/plugin.py
+++ b/spyder/plugins/breakpoints/plugin.py
@@ -46,6 +46,9 @@ class Breakpoints(SpyderPluginWidget):
         self.setLayout(layout)
 
         self.breakpoints.set_data()
+
+        path = osp.join(self.PLUGIN_PATH, self.IMG_PATH)
+        self.icon = ima.icon('breakpoints', icon_path=path)
     
     #------ SpyderPluginWidget API --------------------------------------------
     def get_plugin_title(self):
@@ -54,9 +57,8 @@ class Breakpoints(SpyderPluginWidget):
     
     def get_plugin_icon(self):
         """Return widget icon"""
-        path = osp.join(self.PLUGIN_PATH, self.IMG_PATH)
-        return ima.icon('profiler', icon_path=path)
-    
+        return self.icon
+
     def get_focus_widget(self):
         """
         Return the widget to give focus to when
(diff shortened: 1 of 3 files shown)


oscargus:
Maybe mdi-dots-vertical is an option. Three stacked dots.

oscargus:
Oh, I realized that I want to right-align the line numbers. I'll add it to this PR as well.

ccordoba12:

oscargus:
Tanks @dalthviz ! I changed the indentation (I thought that it pointed to breakpointsgui.py where I had a more obvious issue on the same line number in an earlier commit and that it hadn't updated...).

ccordoba12:
I have a question: What's shown to the right of the "Condition" column? I mean, if nothing is shown there I think we should show the entire path of the file containing the breakpoint.

oscargus:
That is the "default" empty column that show up in most of these "table style" widgets. I can for sure add the full path there and post an image of it (it may be a bit cluttered on small displays or if one undock and makes it narrow, but apart from that I think it makes sense)

oscargus:
Here are the results.

With unambiguous file names kept:
breakpointsfullpath1

Without unambiguous file names:
breakpointsfullpath2

I think that the last one is quite confusing as one will have to look at the right to figure out which file is which.

For the first, I'm personally not sure that it actually adds something.

(But I found a bug, so will update. Turns out that all filenames that ever had a breakpoint is stored in the breakpoint data structure, so removing breakpoints for a file still takes it into account when determining the name.)

I was considering that maybe adding the line content may help in navigating. The problem there is that the file may not be open, so probably not a good idea.

ccordoba12:
After seeing your first screenshot above, I think there's too much info in the Breakpoints pane if we add file paths to it. So we should go with your current improvement.

Thoughts @goanpeca?

goanpeca:
>add file paths to it

Agreed lets keep it simple!

ccordoba12:
@dalthviz, could you post a screenshot with the new icon introduced by @oscargus? Thanks!

dalthviz:
@ccordoba12 a preview of the icon:

bp

ccordoba12:
@dalthviz, I don't like that, so please remove it. We don't have the custom to assign icons to plugins, so I don't want this to be the exception.

oscargus:
I can remove it.

And, well, the plugins do have icons. Should I remove all of them?

Den tors 8 aug. 2019 12:25Carlos Cordoba <notifications@github.com> skrev:

spyder/plugins/breakpoints/plugin.py
```diff
@@ -46,17 +46,19 @@ def __init__(self, parent=None):
...
         """Return widget title"""
         return _("Breakpoints")
-    
+
     def get_plugin_icon(self):
         """Return widget icon"""
-        path = osp.join(self.PLUGIN_PATH, self.IMG_PATH)
-        return ima.icon('profiler', icon_path=path)
-    
+        return self.icon
```
oscargus: @ccordoba12 Probably should remove the `get_plugun_icon` method altogether then? Will remove quite a bit of icons from `icon_manager` as well.

goanpeca: What Carlos meant to say is that qe do not use icons for toggling plugin panes? The panes menu has no icons.  This one is located on a different menu for convenience so it is being duplicated? 

We DO use icons for plugins in the preferences, so there is. O plan to remove any method.

oscargus:
All back as before with the profiler icon for the breakpoints plugin (I had to use the light theme as the icons have the same color as the program bar in the dark theme).

image

And, oh, plugins even have icons in menus:

image

image

But not breakpoints and pylint.

spyder/plugins/breakpoints/plugin.py
oscargus: That haven't changed here. I only added an icon for the debug menu (it does not show up for the toggling anyway). And as shown in my recent comment, two out of four plugins accessible from the menus do have icons already.

I was trying to make it more consistent (and correct, by not using the profiler icon...). But if there are other opinions, I'm not the one to decide.

ccordoba12:
> All back as before with the profiler icon for the breakpoints plugin (I had to use the light theme as the icons have the same color as the program bar in the dark theme).

@oscargus, I'm very sorry, I forgot that we do have icons for some plugins (like the Profiler and Find). Then I think it's find to use the three vertical dots as an icon for the Breakpoints plugin (I can't think of a better one right now).

Could you remove your last two commits and force-push to this branch again? Thanks!

oscargus:
Done! (I think, the earlier two commits with the same name only showed up once in the rebase, but it looks OK.)

PR #9921, opened 2019-07-30
PR: Improve shortcuts entry in Preferences

### LINKED ISSUE #9891: Improve shortcuts table UI
Right now there's very hard for people to know that they can customize our keyboard shortcuts in our Preferences. That's because there's no message mentioning that and no UI element signaling the possibility.

To fix this I think that we could do the following:

1. Add this message on top of the shortcuts table:
    > Here you can browse the list of all available shortcuts in Spyder. You can also customize them by clicking on any entry in this table.

2. Change the cursor to be a hand when hovering the shortcuts table and make a single click open the dialog to change shortcuts.

@goanpeca, @jnsebgosselin, what do you think about my suggestions?

steff456:
## Description of Changes

Add a text label and placeholder for users to know they can customise their shortcuts. 

Captura de pantalla 2019-07-30 a la(s) 11 17 08 a  m

### Issue(s) Resolved

Fixes #9891 

### Affirmation

### DIFF
--- a/spyder/preferences/shortcuts.py
+++ b/spyder/preferences/shortcuts.py
@@ -14,7 +14,7 @@ import re
 from qtpy import PYQT5
 from qtpy.compat import from_qvariant, to_qvariant
 from qtpy.QtCore import QAbstractTableModel, QModelIndex, Qt, Slot, QEvent
-from qtpy.QtGui import QKeySequence, QIcon
+from qtpy.QtGui import QKeySequence, QIcon, QCursor, QMouseEvent
 from qtpy.QtWidgets import (QAbstractItemView, QApplication, QDialog,
                             QGridLayout, QHBoxLayout, QLabel,
                             QLineEdit, QMessageBox, QPushButton, QSpacerItem,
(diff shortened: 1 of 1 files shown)


spyder/preferences/shortcuts.py
```diff
@@ -808,12 +817,22 @@ def setup_page(self):
...
+        self.table.finder.setPlaceholderText(
+            _("Search a shortcut in the table"))
+        self.table.setMouseTracking(True)
         self.label_finder = QLabel(_('Search: '))
         self.reset_btn = QPushButton(_("Reset to default values"))
+        self.label = QLabel()
+        self.label.setText(
+            _("Here you can browse the list of all available shortcuts in "
+              "Spyder. You can also customize them by double-clicking on any "
+              "entry in this table."))
```
ccordoba12: There's no need to create the label first and then set its text. You can do it in a single step, like this:

```python
self.top_label = QLabel(
    _("Here you can browse the list of all available shortcuts in "
    ...
)
```

And yes, please rename this variable to `top_label`, instead of simply `label`.

PR #9930, opened 2019-07-31
[WIP] PR: Bookmarks GUI

### LINKED ISSUE #910: Bookmarks for Spyder Editor
_From [w.hortsc...@gmail.com](https://code.google.com/u/104312223359878511004/) on 2012-01-15T11:12:30Z_

Hallo,

The spyder editor should support bookmarks like the VIM-editor or the editor in MATLAB.
Both bookmark-systems differ in the way you set and access the bookmarks and have their own charm.

MALAB:
In Matlab all bookmarks were set with the same bookmark-keyboard-shortcut (Ctrl-F2). This bookmarks can then be cycled trough by pressing a different shortcut like F2.
A short description (including pictures) about the behaviour of the bookmark-system can be found here http://blogs.mathworks.com/desktop/2007/06/15/scroll-less-with-editor-bookmarks/ This variant is better if you have only two or three bookmarks because you do not have to worry about remembering the specific shortcuts for different locations.

VIM
In VIM it is possible to set several bookmarks which can be accessed (moved to) separately.
See http://www.thegeekstuff.com/2009/02/how-to-add-bookmarks-inside-vi-and-vim-editor/ Especially if you use bookmarks a lot it is better to have specific jump-locations so you do not have to cycle trough all of them.

Thanks in advance,
HoWil

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=910_

### LINKED ISSUE #9021: Implement a full UI for the Bookmarks feature
PR #8481 implements a bookmarks feature for Spyder's Editor. Here, we will discuss and decide on an appropriate UI for the same.

@ccordoba12 's relevant comments:

> For the UI part I'd like to see:
> 
> 1. An icon next to the line that has a bookmark.
> 2. A panel to browse bookmarks.
> 3. A keyboard shortcut to move to the next/previous bookmark.

> We can add the list of bookmarks to the file switcher or add a shortcut to move to the bookmarks pane and filter file names with the keyboard there.

@CAM-Gerlach 's proposed UI:

* Highlighted line number (to save space) of lines with bookmarks
* Bookmarks Manager, based on either Breakpoint panel (has all needed management features) or Outline Explorer (file-based)
* Previous/next shortcut based on most recently accessed order (or maybe even integrate it with Ctrl-Tab?)
* Ability to assign (either preset, or user-defined) shortcuts to specific bookmarks via the Manager for quick access
* Maybe a ``Show bookmarks`` option in the Outline Explorer?

oscargus:
## Description of Changes

This is the current status of the bookmarks GUI integration.
bookmarks

Still to be done (short term):
 - [ ] Refactor the signalling system for bookmarks
       - [ ] Write a bookmark manager
       - [ ] Remove bookmarks using the pane
       - [ ] Remove unused code
 - [ ] Enable/disable panel from preferences
 - [ ] Add keyboard shortcuts

Still to be done (maybe now, maybe later):
  - [ ] Add Bookmarks sub menu to Search(?)-menu with bookmarks in file(?)
  - [x] Reintroduce column-based bookmarks (adding from panel will give beginning of line or first non-space character or cursor position(?)) but limit to one per line.

Still to be done (most likely not now):
 - [ ] Fuzzy filtering in Bookmarks pane
 - [ ] Remove all bookmarks in file (right-click menu on line number/tab?)
 - [ ] Reassign slots/line numbers from pane

### Issue(s) Resolved

Fixes #910 
Fixes #9021

### Affirmation

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -1020,7 +1020,7 @@ class MainWindow(QMainWindow):
         # Load other plugins (former external plugins)
         # TODO: Use this bucle to load all internall plugins and remove
         # duplicated code
-        other_plugins = ['breakpoints', 'profiler', 'pylint']
+        other_plugins = ['breakpoints', 'profiler', 'pylint', 'bookmarks']
         for plugin_name in other_plugins:
             if CONF.get(plugin_name, 'enable'):
                 module = importlib.import_module(
@@ -2278,7 +2278,7 @@ class MainWindow(QMainWindow):
         order = ['editor', 'ipython_console', 'variable_explorer',
                  'help', 'plots', None, 'explorer', 'outline_explorer',
                  'project_explorer', 'find_in_files', None, 'historylog',
-                 'profiler', 'breakpoints', 'pylint', None,
+                 'profiler', 'breakpoints', 'pylint', 'bookmarks', None,
                  'onlinehelp', 'internal_console', None]
         for plugin in self.widgetlist:
             action = plugin._toggle_view_action
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -298,6 +298,10 @@ DEFAULTS = [
              {
               'enable': True,
               }),
+            ('bookmarks',
+             {
+              'enable': True,
+              }),
             ('workingdir',
              {
               'working_dir_adjusttocontents': False,
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -188,6 +188,7 @@ DEFAULTS = [
               'realtime_analysis/timeout': 2500,
               'outline_explorer': True,
               'line_numbers': True,
+              'bookmarks_panel_enabled': True,
               'blank_spaces': False,
               'edge_line': True,
               'edge_line_columns': '79',
--- a/spyder/plugins/bookmarks/__init__.py
+++ b/spyder/plugins/bookmarks/__init__.py
@@ -1,0 +1,14 @@
+# -*- coding: utf-8 -*-
+# -----------------------------------------------------------------------------
+# Copyright (c) 2009- Spyder Project Contributors
+#
+# Distributed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+# -----------------------------------------------------------------------------
+
+
+# =============================================================================
+# The following statement is required to register this 3rd party plugin:
+# =============================================================================
+
+from .plugin import Bookmarks as PLUGIN_CLASS
--- a/spyder/plugins/bookmarks/plugin.py
+++ b/spyder/plugins/bookmarks/plugin.py
@@ -69,7 +69,7 @@ class Bookmarks(SpyderPluginWidget):
 
     def register_plugin(self):
         """Register plugin in Spyder's main window"""
-        self.bookmarks.edit_goto.connect(self.main.editor.load)
+        self.bookmarks.load_bookmark.connect(self.main.editor.load_bookmark)
 
         # Follow bookmark changes
         self.main.editor.bookmarks_changed.connect(self.set_data)
--- a/spyder/plugins/bookmarks/plugin.py
+++ b/spyder/plugins/bookmarks/plugin.py
@@ -1,0 +1,90 @@
+# -*- coding: utf-8 -*-
+# -----------------------------------------------------------------------------
+# Copyright (c) 2009- Spyder Project Contributors
+#
+# Distributed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+# -----------------------------------------------------------------------------
+
+
+"""Bookmarks/errors Plugin."""
+
+
+# pylint: disable=C0103
+# pylint: disable=R0903
+# pylint: disable=R0911
+# pylint: disable=R0201
+
+# Standard library imports
+import os.path as osp
+
+# Third party imports
+from qtpy.QtCore import Slot
+from qtpy.QtWidgets import QVBoxLayout
+
+# Local imports
+from spyder.config.base import _
+from spyder.utils import icon_manager as ima
+from spyder.utils.qthelpers import create_action
+from spyder.api.plugins import SpyderPluginWidget
+
+from .widgets.bookmarksgui import BookmarkWidget
+
+
+class Bookmarks(SpyderPluginWidget):
+    """Bookmarkslist"""
+    CONF_SECTION = 'bookmarks'
+
+    def __init__(self, parent=None):
+        """Initialization."""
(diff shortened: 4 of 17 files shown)


bcolsen:
I think that the side bar in the editor is already too large, I don't think we should be making it even larger. 

Could the bookmarks be combined with the debug section with a control-click to set the bookmark(shift-click currently does conditional break points)? Your current bookmark icon would overlap nicely behind the debug circle

Currently the debug sections doesn't have a separate context menu. We could add one offering breakpoint insertion and removal and bookmark insertion and removal. In the future we could add conditional breakpoints there.

ccordoba12:
> I think that the side bar in the editor is already too large, I don't think we should be making it even larger.

I agree with @bcolsen. We really need to avoid adding a new panel just for this.

oscargus:
It can probably be merged, yes. Is it possible to stack icons while drawing or should one create combined ones?

goanpeca:
> Is it possible to stack icons while drawing

Yes it is possible

bcolsen:
We stack the `# TODO`'s over the errors in the info panel so it should work.

oscargus:
Should I also try to add markers to the right-hand side "band"? Personally, I do not see the need. Although it may be good to see where the bookmarks are, I doubt that one will use that to jump to them. Being able to add them graphically is primarily a convenience thing.

bcolsen:
> Should I also try to add markers to the right-hand side "band"? 

I think would be handy to see at a glace if you have created a bookmark in a file. The only reason not to would be whether we have any colors left that could be used for bookmarks(we might not).

oscargus:
Closing this as the enforced design with showing at most one of bookmarks/todos/warnings doesn't work for me and do not want to spend more time on it then.

ccordoba12:
> Closing this as the enforced design with showing at most one of bookmarks/todos/warnings doesn't work for me and do not want to spend more time on it then.

I said *we should*, not *we must* implement things using a single panel. So I don't see the need for you to close this and PR #9836 because we can still keep discussing the user interface for this.

Why do you think it's better to add multiple panes for these plugins?

goanpeca:
As I wrote my concern in the other thread, I believe we cannot really keep adding base plugins and panels to Spyder just "for the sake of it" we really need to think things more thoroughly so that the UI we offer is not too crowded or too confusing for users. 


>Closing this as the enforced design with showing at most one of bookmarks/todos/warnings doesn't work for me and do not want to spend more time on it then.

Sorry to hear this @oscargus.

We have other options, like mixing not with an exclusive combobox but using checkboxes, so that the user can display several things at the same time, or just some.

I think the bigger concept here is that we have some type of annotation (be it a bookmark, a warning are error, a code tip etc) and we want to provide a single interface to access to these annotations/messages etc. This is where a single plugin idea makes sense, as conceptually they are one and the same, messages for the user with a give position and meaning per file.

ccordoba12:
> As I wrote my concern in the other thread, I believe we cannot really keep adding base plugins and panels to Spyder just "for the sake of it" we really need to think things more thoroughly so that the UI we offer is not too crowded or too confusing for users.

I agree. We've added a lot of complexity in terms of new features to Spyder 4 but I think as a user you don't feel that reflected in the user interface, and that's a good thing.

Therefore adding 3 new panels is something we need to consider very carefully.

> We have other options, like mixing not with an exclusive combobox but using checkboxes, so that the user can display several things at the same time, or just some.

Or another option could be using tabs.

> I think the bigger concept here is that we have some type of annotation (be it a bookmark, a warning are error, a code tip etc) and we want to provide a single interface to access to these annotations/messages etc.

This is a good point, I haven't seen things like that.

goanpeca:
>Or another option could be using tabs.

I think @oscargus mentioned he wanted to have a look simultaneously at different things (todos, and warnings for example) so that is why I suggest the having a single view that could mix everything.

Khalilsqu:
The main reason I started using spyder over pycharm although I am at a university and can obtain the professional version for free, is that it has simple user interface, ideal for use in scientific computing which is the main purpose spyder built for. Other students love it as well.

ccordoba12:
Thanks @Khalilsqu for your input! Glad to know our user interface and user experience design choices is what drives people to Spyder!

As you probably have seen, we put a lot of effort on them.

oscargus:
The thing is that there is no need to show all the plugins, only the activated ones will allocate a tab. However, with the suggestion you put forward (I didn't really see much room for discussion at that time, maybe a language thing, as I'm a non-native English speaker, but for me should is the same as must, I would have used could) **limits** the possibility to show information concurrently. Information that I would like to see concurrently. By having them as separate plugins, I can undock them, independent of where they are positioned.

Hence, for me it is better to develop the todo and warning panes as external plugins. In that way, anyone who wants to see the information concurrently can do so. 

As I wrote in another thread: one can discuss where to put the various plugins. In the code now, all "additional" plugins always ends up in the same position. This is not ideal as it in the Spyder default layout gives up to 11 (plus any external plugins) in the same "screen area"(? not sure what to call it). This may indeed be a problem, if all are shown at the same time. Ideally, the user should be able to position plugins in any screen area, and create its own layouts, but that is clearly a rather big effort.

For me, when I write code, I would prefer to see the bookmarks, TODOs, warnings, (ideally some GitHub information, like loading an issue), the code of course, and possibly the find in files and help.

While running/debugging: console, breakpoints, variables, and editor.

I understand that keeping things simple is important, but also to make things flexible. These are not contradictory. One can have a simple default setting and the ability to add additional features. (Not sure that there should only be one way to do one thing either.)

There is also a fundamental difference between todos/warnings and bookmarks which in my opinion makes them unsuitable to merge into a single interface. Todos/warnings are file-based, while bookmarks are for all files. One could imagine putting todos/warnings in the same interface for sure, it is simple to align the columns. One main reason for not merging todos/warnings was that I couldn't really figure out a single name, even worse an icon, for it (I also started with one of them just to see how it could be done). Merging those into one should be rather straightforward, and one can imagine using checkboxes/menu options to select what to show (one can discuss if these should effect the corresponding settings in preferences or just be session-based).

Also, it can be worth noting that I didn't delete the branch, so anyone wanting to take it up can do so. I just didn't want to put the effort in to get an interface that did not bring me any benefit.

(On another note, ideally, bookmarks should be session/project-based somehow.)

ccordoba12:
> By having them as separate plugins, I can undock them, independent of where they are positioned.

What if this "Extra info" plugin has the possibility to show each tab in a separate window? That's doable and it would cover your use case as well as our design decision.

> As I wrote in another thread: one can discuss where to put the various plugins. In the code now, all "additional" plugins always ends up in the same position

I don't understand what you mean by this. Could you provide a screenshot?

PR #9935, opened 2019-07-31
PR: Fix some issues with debugger panel and history pane

oscargus:
## Description of Changes

* [x] Added tests

Some minor fixes related to the debugger panel and history pane:

 * The transparent marker shows the type of breakpoint (conditional when holding shift). As preparation for changes in #9930 
 * The condition dialog doesn't show up if a breakpoint cannot be added to the line. (Instead of realizing that afterwards.)
 * Removed debug panel from history pane 
 * Deleted unused history widget (`show_class_func_dropdown=False` from that). Seems useful if not more people spend hours trying to understand why their changes in that file doesn't affect the history pane...
 * Made code folding optional in history pane (earlier on, it was on by default and not optional).

Pressing shift with the mouse static over the debugger panel doesn't change the icon. I can add it, but seemed a bit overkill.

I noted that changing this an history option in the preferences does not directly affect the history pane. However, this is a behaviour not affected by this PR. Not sure if I should open an issue for it (or if there already is one) or if this will be solved automatically with the preferences update.

### Affirmation


oscargus:
Will have to update the tests as sig_breakpoints_changed.emit no longer emits unless a breakpoint actually was changed. Not today though.

goanpeca:
Is this still valid @ccordoba12 ?

ccordoba12:
I don't understand very well what it does. Let's close it.

PR #9966, opened 2019-08-05
PR: Save files when calling runfile directly in the console

### LINKED ISSUE #9342: File not saved when calling runfile from console

impact27:
## Description of Changes
Save files before running when running files from the kernel

### Issue(s) Resolved

Fixes #9342

### Affirmation

### DIFF
--- a/spyder/plugins/ipythonconsole/widgets/shell.py
+++ b/spyder/plugins/ipythonconsole/widgets/shell.py
@@ -98,7 +98,7 @@ class ShellWidget(NamepaceBrowserWidget, HelpWidget, DebuggingWidget,
             'pdb_state': self.set_pdb_state,
             'pdb_continue': self.pdb_continue,
             'get_breakpoints': self.get_spyder_breakpoints,
-            'savefiles': self.handle_kernel_api_savefiles,
+            'save_files': self.handle_save_files,
         }
 
         for request_id in handlers:
@@ -469,7 +469,7 @@ the sympy module (e.g. plot)
         editorstack.set_stack_index(index)
         return editor
 
-    def handle_kernel_api_savefiles(self):
+    def handle_save_files(self):
         """Save the open files."""
         self.editorstack().save()
 
--- a/spyder/plugins/ipythonconsole/widgets/shell.py
+++ b/spyder/plugins/ipythonconsole/widgets/shell.py
@@ -98,6 +98,7 @@ class ShellWidget(NamepaceBrowserWidget, HelpWidget, DebuggingWidget,
             'pdb_state': self.set_pdb_state,
             'pdb_continue': self.pdb_continue,
             'get_breakpoints': self.get_spyder_breakpoints,
+            'savefiles': self.handle_kernel_api_savefiles,
         }
 
         for request_id in handlers:
@@ -451,6 +452,26 @@ the sympy module (e.g. plot)
           len(command.splitlines()) == 1:
             if not 'inline' in command:
                 self.silent_execute(command)
+
+    # ---- Spyder-kernels methods -------------------------------------------
+    def get_editor(self, filename=None):
+        """Get editor for filename and set it as the current editor."""
+        editorstack = self.editorstack()
+
+        if not filename:
+            return editorstack.get_current_editor()
+
+        index = editorstack.has_filename(filename)
+        if index is None:
+            return None
+
+        editor = editorstack.data[index].editor
+        editorstack.set_stack_index(index)
+        return editor
+
+    def handle_kernel_api_savefiles(self):
+        """Save the open files."""
+        self.editorstack().save()
 
     # ---- Private methods (overrode by us) ---------------------------------
 
--- a/spyder/plugins/ipythonconsole/widgets/shell.py
+++ b/spyder/plugins/ipythonconsole/widgets/shell.py
@@ -454,23 +454,6 @@ the sympy module (e.g. plot)
                 self.silent_execute(command)
 
     # ---- Spyder-kernels methods -------------------------------------------
-    def get_editor(self, filename=None):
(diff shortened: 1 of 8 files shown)


impact27:
Could someone have a look at the windows test fails?

ccordoba12:
@impact27, please remove all merges in this branch and leave a single one with master.

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -452,6 +453,26 @@ def change_mpl_backend(self, command):
...
+        if index is None:
+            return None
+
+        editor = editorstack.data[index].editor
+        editorstack.set_stack_index(index)
+        return editor
+
+    def handle_kernel_api_savefiles(self):
+        """Save the open files."""
+        self.editorstack().save()
```
ccordoba12: This is not the way we decided to handle interactions between plugins: if a widget from one plugin needs to call another plugin, it needs to emit a Qt signal, then that signal must be handled by its parent plugin, which asks the other plugin to perform the desired action.

So in this case you need to emit a signal at this point (say `sig_save_file`) with the filename you want to save. Then you need to connect that signal to a slot in `register_client` of `plugins/ipythonconsole/plugin.py` (say `save_runfile`). Since that slot is at the plugin level, it can call directly `self.main.editor` to detect if `filename` is open there, then call the current editorstack to save it.

impact27: Are the Qt signals syncronous? The goal here is to save the file before resuming execution.

ccordoba12: Right, then you need to use a QEventLoop to make it synchronous.

impact27: Does using a `QEventLoop` to make a blocking call not kind of defeat the point of using Qt signals? I mean I can either:
 - Call a method of the plugin and use the return value.
 - Emit a signal to the plugin that will then emit a signal with the reply, while I have a `QEventLoop` running, inplement a mechanism to save the reply (Another method and a class attribute?), implement a mechanism to stop the `QEventLoop` when the reply is recieved (Probably just connecting the reply signal to the Slot of `QEventLoop`), and implement a mechanism to get the reply after the `QEventLoop` finishes.

The second one seems to add quite a lot of complexity to the code. What is the adventage of this approach? I mean in general I would agree that for async calls Qt signals are superior, but I am not so sure about sync calls. Is there anywhere else in the code where a sync call is made with a `QEventLoop`? The only place I can think of is the communication with the kernel, but there a direct call was not possible.

ccordoba12: > The second one seems to add quite a lot of complexity to the code

Yeah, you're right. As soon as I wrote you about using a QEventLoop, I realized it makes things way too complex.

Then let's change this to the following:

```python
plugin = self.ipyclient.plugin
if plugin.main.editor is not None:
    editor = plugin.main.editor
    editorstack = editor.get_current_editorstack()
    editorstack.save()
```

For this to work you also need to create an attribute called `plugin` in `ClientWidget` that holds the arg `plugin` passed to it.

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -452,6 +453,26 @@ def change_mpl_backend(self, command):
             if not 'inline' in command:
                 self.silent_execute(command)
 
+    # ---- Spyder-kernels methods -------------------------------------------
+    def get_editor(self, filename=None):
```
ccordoba12: This method is unnecessary, right? If so, please remove it.

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -452,6 +453,26 @@ def change_mpl_backend(self, command):
...
+
+        index = editorstack.has_filename(filename)
+        if index is None:
+            return None
+
+        editor = editorstack.data[index].editor
+        editorstack.set_stack_index(index)
+        return editor
+
+    def handle_kernel_api_savefiles(self):
```
ccordoba12: Let's rename this to `handle_save_files`.

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -98,6 +98,7 @@ def __init__(self, ipyclient, additional_options, interpreter_versions,
             'pdb_state': self.set_pdb_state,
             'pdb_continue': self.pdb_continue,
             'get_breakpoints': self.get_spyder_breakpoints,
+            'savefiles': self.handle_kernel_api_savefiles,
```
ccordoba12: Let's call the handler `save_files` instead of `savefiles`.

spyder/plugins/ipythonconsole/widgets/shell.py
impact27: Yes but not for long. It is used for run_cell

ccordoba12:
@impact27, this is ready. Please don't more commits here.

PR #10010, opened 2019-08-13
PR: Add permanent history to the debugger

impact27:
## Description of Changes
Small changes to pdb to save the history, so I can access the history of a previous debugging session.

This is not a really nice way of doing thing, but it works. Two (non easy) way of improving this:

 - The history of pdb should be saved at the spyder-kernels level, as is the case with ipython.
    (Well only kind of because the history is duplicated, see #7141 )
 - The history should be saved in a sqlite file, in the same way ipython does. I tried to reuse ipython core functions but I could't get it to work quickly. This allows to save the time the command was executed, and to separate between different sessions.

### Issue(s) Resolved

Fixes #

### Affirmation


spyder/plugins/ipythonconsole/widgets/control.py
```diff
@@ -26,13 +30,25 @@ class ControlWidget(TracebackLinksMixin, GetHelpMixin, BrowseHistoryMixin,
     visibility_changed = Signal(bool)
     go_to_error = Signal(str)
     focus_changed = Signal()
+    SEPARATOR = '{0}## ---({1})---'.format(os.linesep*2, time.ctime())
+    INITHISTORY = ['# -*- coding: utf-8 -*-',
+                   '# *** Spyder Python Console History Log ***']
```
ccordoba12: This second line is not necessary, please remove it.

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -117,10 +117,12 @@ def callback(line):
             # Save history to browse it later
             if not (len(self._control.history) > 0
                     and self._control.history[-1] == line):
-                # do not save pdb commands
+                # do not save pdb commands unless they have arguments
```
ccordoba12: do -> Do

ccordoba12:
> This is not a really nice way of doing thing, but it works

I know this is not the best solution, but it's good enough until we have time to seriously improve our debugger (which is not going to happen for Spyder 4, but probably for version 5).

impact27:
The only think that could be

> > This is not a really nice way of doing thing, but it works
> 
> I know this is not the best solution, but it's good enough until we have time to seriously improve our debugger (which is not going to happen for Spyder 4, but probably for version 5).

I have been playing with that a bit and there is problems with multilines inputs. I will try to fix that with sqlite.

ccordoba12:
> I have been playing with that a bit and there is a problem with multilines inputs. I will try to fix that with sqlite.

You shouldn't waste your time with that because the solution we're thinking to implement doesn't involve enhancing IPdb but creating a debugging kernel.

Besides, I think multiline is really not an option in Pdb because you have to prefix Python code with `!` so that Pdb parses it correctly.

impact27:
I used `from IPython.core.history import HistoryManager` so the history will be correctly saved.

ccordoba12:
> I used from IPython.core.history import HistoryManager so the history will be correctly saved.

I'm -1 on that and I'd like you to revert that change. That not only makes things more complex than they need to be, but also there's no need of it because you can write multi-line statements in Pdb.

impact27:
Well the complexity is all in `IPython.core.history.HistoryManager`, so the spyder part is actually simpler. The multiline problem comes when I have:
```python
import numpy as np
np.abs([1,
        2])
```
If I want to copy-paste the `abs` in pdb, it will work as is:

>Screenshot 2019-08-14 at 15 08 43

 But in the old implementation, as a newline separate the statement, I ended up with:
```python
history = [
'import numpy as np', 
'np.abs([1,',
'        2])']
```
With the new implementation, the newlines are not specific characters, so there is no problem. 

Using `IPython.core.history.HistoryManager` adds the benefits that sessions are correctly handled. It is possible to separate the history of two different session, event if they took place at the same place.

Futhermore, all the tools that are used to manage ipython history can be easily adapted to manage this new pdb history. 

I therefore believe that this implementation is more future proof and has less bugs.

impact27:
I mean with a database I can store executed cells, not just lines, and I can attach more informations, such as the session this line was from, the line number, and more. So in the future if we want to add functionalities to the history, the database already exists and we don't need to migrate from a text file.

ccordoba12:
> The multiline problem comes when I have

I haven't considered this use case (I don't know how it could arise, beside copy/paste though).

> I mean with a database I can store executed cells, not just lines

Cells executed with `runcell`, right?

> So in the future if we want to add functionalities to the history, the database already exists and we don't need to migrate from a text file.

Ok, that's a good point too.

spyder/plugins/ipythonconsole/widgets/control.py
```diff
@@ -5,17 +5,33 @@
...
+class PdbHistory(HistoryManager):
+
+    def _get_hist_file_name(self, profile=None):
+        """
+        Get default pdb history file name.
+
+        The profile parameter is ignored, but must exist for compatibility with
+        the parent class.
+        """
+
```
ccordoba12: Please remove this blank line.

spyder/plugins/ipythonconsole/widgets/control.py
```diff
@@ -73,6 +95,24 @@ def focusOutEvent(self, event):
...
         return super(ControlWidget, self).focusOutEvent(event)
 
+    def add_to_pdb_history(self, line_num, line):
+        """Add command to history"""
+        self.histidx = None
+        if not line:
+            return
+        line = line.strip()
+
+        # If repeated line
```
ccordoba12: Let's change this comment to

    # Ignore repeated lines.

impact27:
> Cells executed with `runcell`, right?

Multiline statment pasted by the user in pdb, but it still needs to be a single statment. otherwise:
`*** SyntaxError: multiple statements found while compiling a single statement`

spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py
```diff
@@ -683,6 +684,31 @@ def test_browse_history_dbg(ipyconsole, qtbot):
     qtbot.keyClick(control, Qt.Key_Up)
     assert '!aa = 10' in control.toPlainText()
 
+    # Open new widget
+    client_name = ipyconsole.get_current_client().get_name()
```
ccordoba12: This line is not needed.

spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py
```diff
@@ -683,6 +684,31 @@ def test_browse_history_dbg(ipyconsole, qtbot):
...
+    ipyconsole.create_new_client()
+    assert client_name != ipyconsole.get_current_client().get_name()
```
ccordoba12: This line is not needed.

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -10,19 +10,36 @@
...
+class PdbHistory(HistoryManager):
+
+    def _get_hist_file_name(self, profile=None):
+        """
+        Get default pdb history file name.
+
+        The profile parameter is ignored, but must exist for compatibility with
+        the parent class.
+        """
+
```
ccordoba12: Please remove this blank.

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -36,6 +53,10 @@ def __init__(self, *args, **kwargs):
         self._input_ready = False
         self._last_pdb_cmd = ''
         self._pdb_line_num = 0
+        self.pdb_history_file = PdbHistory()
+        self._control.history = [
```
ccordoba12: Why does `history` have to be an attribute of `_control`? Why not using an attribute here like `pdb_history`?

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -196,3 +217,21 @@ def _event_filter_console_keypress(self, event):
     def is_debugging(self):
         """Check if we are debugging."""
         return self.spyder_kernel_comm._debugging
+
+    def add_to_pdb_history(self, line_num, line):
+        """Add command to history"""
+        self._control.histidx = None
```
ccordoba12: The same for this: Why not make `histidx` an attribute of `DebuggingWidget`?

spyder/plugins/ipythonconsole/widgets/debugging.py
impact27: _control still needs to control the history because it must be able to modify the text. For example, when I press up, the cirrent text is removed, and a new text is replaced.

spyder/plugins/ipythonconsole/widgets/debugging.py
impact27: Same reason, BrowseHistory is a parent of the control.

spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py
impact27: I want to make sure I am in a new tab. How should I do that?

ccordoba12: > ipyconsole.create_new_client()

That always create a new tab, so there's no need of that extra check.

spyder/plugins/ipythonconsole/widgets/debugging.py
ccordoba12: Ok, got it.

ccordoba12:
@impact27, this now needs a merge with master.

impact27:
There is a small issue with backward search in pdb that I will have a look at now.

PR #10016, opened 2019-08-13
PR: Escape regex characters from parameters while processing a signature

### LINKED ISSUE #10011: re.error when getting signature in the editor

I was just typing the statement new_shape = np.concatenate(()) and the issue reporter popped up.

### Traceback
```python-traceback
    documentation=documentation,
    max_width=max_width)
    max_width=max_width
    row = re.sub(pattern, handle_sub, row)
    return _compile(pattern, flags).sub(repl, string, count)
    p = sre_compile.compile(pattern, flags)
    p = sre_parse.parse(p, flags)
    p = _parse_sub(source, pattern, flags & SRE_FLAG_VERBOSE, 0)
    not nested and not items))
    source.tell() - start)
re.error: missing ), unterminated subpattern at position 9

    msg = self.format(record)
    return fmt.format(record)
    record.message = record.getMessage()
    msg = msg % self.args
ValueError: incomplete format
Call stack:
    sys.exit(main())
    mainwindow.main()
    app.exec_()
    handler(resp['result'], req_id)
    {'params': response})
    self.COMPLETION_CLIENT_NAME, req_id, response)
    completion_source, response_instance, req_type, req_id)
    response_instance.handle_response(req_type, responses)
    handler(params)
    self.log_lsp_handle_errors("Error when processing signature")
    logger.error('%', 1, stack_info=True)
Arguments: (1,)
```

## Versions

* Spyder version: 4.0.0b4 
* Python version: 3.7.4
* Qt version: 5.12.4
* PyQt5 version: 5.12.3
* Operating System: Linux 4.15.0-54-generic

### Dependencies

```
qdarkstyle >=2.6.4:  2.7 (OK)
psutil >=0.3      :  5.6.3 (OK)
cython >=0.21     :  0.29.12 (OK)
qtconsole >=4.5.2 :  4.5.2 (OK)
IPython >=4.0     :  7.7.0 (OK)
```

dalthviz:
## Description of Changes

### Issue(s) Resolved

Fixes #10011 

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/tests/test_codeeditor_1.py
+++ b/spyder/plugins/editor/widgets/tests/test_codeeditor_1.py
@@ -61,27 +61,29 @@ def copy_files_clipboard(create_folders_files):
 def test_format_signature(code_editor_bot):
     """Test signature format method."""
     signature = """
-    concatenate((a1, a2, a...), [b1, b2, b...], axis={}, index=[], **args)"""
+    concatenate((a1, a2, a...), [b1, b2, b...], axis={}, index=[],
+                *args, **kargs)"""
     editor, qtbot = code_editor_bot
 
-    format_signature = editor._format_signature(signature, parameter="a1")
+    format_signature = editor._format_signature(signature, parameter="(a1")
 
     assert "color:#DAA520'><b>a1</b></span>" in format_signature
 
     format_signature = editor._format_signature(signature, parameter="a2")
     assert "color:#DAA520'><b>a2</b></span>" in format_signature
 
-#    format_signature = editor._format_signature(signature, parameter="a...")
-#    assert "color:#DAA520'><b>a...</b></span>" in format_signature
+    format_signature = editor._format_signature(signature, parameter="a...")
+    print(format_signature)
+    assert "color:#DAA520'><b>a...</b></span>" in format_signature
 
-#    format_signature = editor._format_signature(signature, parameter="b1")
-#    assert "color:#DAA520'><b>b1</b></span>" in format_signature
-#
-#    format_signature = editor._format_signature(signature, parameter="b2")
-#    assert "color:#DAA520'><b>b2</b></span>" in format_signature
-#
-#    format_signature = editor._format_signature(signature, parameter="b...")
-#    assert "color:#DAA520'><b>b...</b></span>" in format_signature
+    format_signature = editor._format_signature(signature, parameter="[b1")
+    assert "color:#DAA520'><b>b1</b></span>" in format_signature
+
+    format_signature = editor._format_signature(signature, parameter="b2")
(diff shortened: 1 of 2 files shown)


ccordoba12:
@dalthviz, could you explain why this change is necessary?

ccordoba12:
> please add a test for this one

@dalthviz, instead of testing this directly with the PyLS (which is not  easy), I think it's simpler to create a fixture that contains an instance of CodeEditor with no LSP services (by using `codeeditor_factory` in `spyder/plugins/editor/widgets/tests/conftest.py`), and then pass different parameters to `_format_signature` to test what it returns.

dalthviz:
@ccordoba12 when processing the signature and more precisely the parameters if any, we create a regex pattern with the parameter's text at [this line](https://github.com/spyder-ide/spyder/blob/37952393120d49231ef1b63c90cab97b632f1c3a/spyder/widgets/mixins.py#L361) which is used with `re.sub` at [this line](https://github.com/spyder-ide/spyder/blob/37952393120d49231ef1b63c90cab97b632f1c3a/spyder/widgets/mixins.py#L371). The error was caused because the signature of `numpy.concatenate` includes a tuple (array like parameter) as a parameter:

```python
numpy.concatenate((a1, a2, ...), axis=0, out=None)
```

So when processing such signature and parameters, the regex pattern had an extra `(` (from the initial parameter taken `(a1`), making an invalid pattern.

ccordoba12:
Ok, thanks for the explanation @dalthviz! Then let's try to exercise `_format_signature` as much as possible in the new test, to try to avoid similar failures in the future.

dalthviz:
@ccordoba12 while testing this I realized that part of the problem is that the `activeParameter` retrieved doesn't respect that the first parameter is a tuple or array like element (in the case of `numpy.concatenate`, for example). So the signature's info has, as list of parameters, something like:

```python
['(a1', 'a2', '...)',  'axis', 'out']
```
instead of

```python
 ['(a1, a2, ...)', 'axis', 'out']
```
From what I saw this, is the way Jedi parses the parameters ([since the `pyls` at the end just calls `jedi.Script(...).call_signatures()`](https://github.com/palantir/python-language-server/blob/1b5b69b0a23c26b6767511cb7c6f3e0a5df2b2ac/pyls/plugins/signature.py#L16))

ccordoba12:
> So the signature's info has, as list of parameters, something like:
['(a1', 'a2', '...)',  'axis', 'out']

That's really awful indeed!!

However, after talking with @dalthviz about the way Jedi returns parameters for collections, we decided that the best thing we can do for now is to merge this PR to avoid the error reported on issue #10011 and similar errors in the future.

Then we need to report this in the Jedi repo and try to fix it there.

PR #10060, opened 2019-08-22
[WIP] PR: Migrate FileSwitcher functionality to Switcher

dalthviz:
## Description of Changes

### Issue(s) Resolved

Fixes #

### Affirmation

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -411,7 +411,8 @@ class MainWindow(QMainWindow):
         self.tours_available = None
 
         # File switcher
-        self.fileswitcher = None
+        # TODO: Change to Switcher
+        self.switcher = None
 
         # Check for updates Thread and Worker, refereces needed to prevent
         # segfaulting
@@ -666,12 +667,13 @@ class MainWindow(QMainWindow):
         self.register_shortcut(self.toggle_previous_layout_action, "_",
                                "Use previous layout")
         # File switcher shortcuts
+        # TODO: Change to Switcher
         self.file_switcher_action = create_action(
                                     self,
                                     _('File switcher...'),
                                     icon=ima.icon('filelist'),
                                     tip=_('Fast switch between files'),
-                                    triggered=self.open_fileswitcher,
+                                    triggered=self.open_switcher,
                                     context=Qt.ApplicationShortcut)
         self.register_shortcut(self.file_switcher_action, context="_",
                                name="File switcher")
@@ -3205,35 +3207,46 @@ class MainWindow(QMainWindow):
         self.tour.set_tour(index, frames, self)
         self.tour.start_tour()
 
-    # ---- Global File Switcher
-    def open_fileswitcher(self, symbol=False):
-        """Open file list management dialog box."""
-        if self.fileswitcher is not None and \
-          self.fileswitcher.is_visible:
-            self.fileswitcher.hide()
-            self.fileswitcher.is_visible = False
+    # ---- Global Switcher
+    def open_switcher(self, symbol=False):
+        """Open switcher dialog box."""
+        # TODO: Change to Switcher
+        if self.switcher is not None and self.switcher.isVisible():
+            self.switcher.hide()
             return
+        self.switcher.clear()
         if symbol:
-            self.fileswitcher.plugin = self.editor
-            self.fileswitcher.set_search_text('@')
+            self.switcher.set_search_text('@')
         else:
-            self.fileswitcher.set_search_text('')
-        self.fileswitcher.show()
-        self.fileswitcher.is_visible = True
+            self.switcher.set_search_text('')
+        self.switcher.show()
 
     def open_symbolfinder(self):
         """Open symbol list management dialog box."""
-        self.open_fileswitcher(symbol=True)
+        # TODO: Change to Switcher
+        self.open_switcher(symbol=True)
 
-    def add_to_fileswitcher(self, plugin, tabs, data, icon):
-        """Add a plugin to the File Switcher."""
-        if self.fileswitcher is None:
-            from spyder.widgets.fileswitcher import FileSwitcher
-            self.fileswitcher = FileSwitcher(self, plugin, tabs, data, icon)
-        else:
-            self.fileswitcher.add_plugin(plugin, tabs, data, icon)
-        self.fileswitcher.sig_goto_file.connect(
-            plugin.get_current_tab_manager().set_stack_index)
+    def create_switcher(self, mode_symbol, mode_name, mode_handler):
+        """Create switcher dialog instance."""
+        # TODO: Change to Switcher
+        if self.switcher is None:
+            from spyder.widgets.switcher import Switcher
+            self.switcher = Switcher(self)
+            self.switcher.add_mode('>', _('Commands'))
+            self.switcher.add_mode('?', _('Help'))
+            self.switcher.add_mode(':', _('Go to Line'))
+            self.switcher.add_mode('@', _('Go to Symbol in File'))
+            self.switcher.sig_mode_selected.connect(self.handle_switcher_modes)
+
+    def handle_switcher_modes(self, mode):
+        """Handle switcher mode change."""
+        if mode == ':':
+            self.editor.create_line_switcher(self.switcher)
+        elif mode == '@':
+            self.editor.create_symbol_switcher(self.switcher)
+        elif mode == '':
+            # Get list with all the plugins with tabs?
+            self.create_plugins_switcher()
 
     # ---- Check for Spyder Updates
     def _check_updates_ready(self):
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -885,6 +885,10 @@ class MainWindow(QMainWindow):
         self.main_toolbar = self.create_toolbar(_("Main toolbar"),
                                                 "main_toolbar")
 
+        # Switcher instance
+        self.create_switcher()
+        logger.info("Loading switcher...")
+
         # Internal console plugin
         logger.info("Loading internal console...")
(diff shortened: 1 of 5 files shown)


ccordoba12:
This is looking pretty good, thanks @dalthviz!!

I have a quick comment: could you decrease the amount of vertical space between rows in the new switcher to the minimal possible? I say it because in the screenshot you posted above there's too much space now, and I'd like to see it reduced as much as possible.

goanpeca:
This is looking great @dalthviz :-)

goanpeca:
Moving forward @dalthviz do you see the need to make some convenience method on the switcher, or perhaps encapsulate all the methods needed in a separtae clasee for the editor?

dalthviz:
@goanpeca yep I was thinking that probably I should create a mixin with all the logic for the Editor to not duplicate code between the plugin and the widget of the Editor (which is basically the same)

goanpeca:
Yes a switcher manager could work ;-)

spyder/widgets/switcher.py
```diff
@@ -269,6 +269,10 @@ def __init__(self, parent=None, icon=None, title=None, description=None,
 
         if icon:
             self.setIcon(icon)
+            # TODO: Change fixed icon size value
+            self._icon_width = 20
```
goanpeca: So this should be updated based on the editor zoom maybe?

dalthviz: I was thinking to get the size from the icon but I'm unsure how to get it

spyder/plugins/editor/utils/switcher.py
```diff
@@ -0,0 +1,219 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+
+"""
+Editor Switcher manager
```
goanpeca:
```suggestion
Editor Switcher manager.
```

spyder/plugins/editor/utils/switcher.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+
+    SYMBOL_MODE = '@'
+    LINE_MODE = ':'
+    FILES_MODE = ''
+
+    def __init__(self, switcher_instance, editor, editorstack,
+                 section=_("Editor")):
+        """
+        Params editor and editorstack should be callables i.e functions
```
goanpeca:
```suggestion
        Params editor and editorstack should be callables i.e functions.
```

spyder/plugins/editor/utils/switcher.py
```diff
@@ -0,0 +1,219 @@
...
+        if mode == self.SYMBOL_MODE:
+            self.create_symbol_switcher()
+        elif mode == self.LINE_MODE:
+            self.create_line_switcher()
+        elif mode == self.FILES_MODE:
+            # Each plugin that wants to attach to the switcher should do this?
+            self.create_editor_switcher()
+
+    def create_editor_switcher(self):
+        """Populate switcher with ."""
```
goanpeca: with ?

spyder/widgets/switcher.py
```diff
@@ -654,6 +682,28 @@ def setup_sections(self):
                     item.set_section_visible(True)
 
         self.proxy.sortBy('_score')
+        self.sig_item_changed.emit(self.current_item())
+
+    def set_position(self, top):
+        """Positions the dialog."""
```
goanpeca:
```suggestion
        """Set the position of the dialog."""
```

goanpeca:
@dalthviz left a few comments. Found a couple of issues.

1.) I see the `editor` text in blue for all the items. I think the idea was to have a single "header" per block. (on your animation looks different)

Screen Shot 2019-09-04 at 23 34 17

2.) When in LINE_MODE, I can use up and down arrows which is a bit weird

image

3.) Also when in line mode if I type lets say

`:12` and press `Escape` the document the cursor should return to the original position it had before changig (since escape effectively cancelled that move)

dalthviz:
> @dalthviz left a few comments. Found a couple of issues.
> 
 @goanpeca from the images you show it seems you did the tests with the master version of Spyder or a different branch from the one on this PR (there is a info icon that is not present in the new Switcher) could be that the case and the reason you experienced such behaviors?

goanpeca:
@dalthviz yes you are absolutely right! I was on master 🤦‍♂ 

Just checked again!

All seems in order ignore the last comments!

spyder/plugins/editor/widgets/editor.py
```diff
@@ -857,36 +861,42 @@ def clone_from(self, other):
         self.set_stack_index(other.get_stack_index())
 
     @Slot()
-    def open_fileswitcher_dlg(self):
+    @Slot(str)
+    def open_switcher_dlg(self, initial_text=''):
```
ccordoba12: When is this method called? I did some tests and it seems that now we're always calling the MainWindow method to open the switcher, so perhaps this is not needed anymore?

spyder/plugins/editor/widgets/editor.py
```diff
@@ -857,36 +861,42 @@ def clone_from(self, other):
...
+        self.switcher_dlg.set_search_text(initial_text)
+        self.switcher_dlg.setup()
+        self.switcher_dlg.show()
+        # Note: the +1 pixel on the top makes it look better
+        delta_top = (self.tabs.tabBar().geometry().height() +
+                     self.fname_label.geometry().height() + 1)
+        self.switcher_dlg.set_position(delta_top)
 
     @Slot()
     def open_symbolfinder_dlg(self):
```
ccordoba12: The same goes for this method: we still need it? Or can we remove it?

spyder/utils/sourcecode.py
```diff
@@ -212,3 +221,188 @@ def shortest_path(files_path_list):
                 shortest_path_length = len(path_elmts)
                 shortest_path = path_elmts
         return os.path.join(*shortest_path)
+
+
+def shorten_paths(path_list, is_unsaved):
```
ccordoba12: Please move all new functions in this file to `spyder/utils/switcher.py`. They really don't belong here.

spyder/app/mainwindow.py
```diff
@@ -883,6 +883,10 @@ def create_edit_action(text, tr_text, icon):
         self.main_toolbar = self.create_toolbar(_("Main toolbar"),
                                                 "main_toolbar")
 
+        # Switcher instance
+        self.create_switcher()
+        logger.info("Loading switcher...")
```
ccordoba12: Let's move this line above `self.create_switcher()`.

ccordoba12:
@dalthviz, I forgot to mention: if the old fileswitcher is not used anymore, please remove it as part of this PR too.

spyder/plugins/editor/widgets/editor.py
dalthviz: This method is called from the actions available on the Options Menu. When using them on a Editor window the Switcher is shown as part of the Editor window:

image

image

spyder/plugins/editor/widgets/editor.py
dalthviz: Same as above

spyder/plugins/editor/widgets/editor.py
ccordoba12: Ok, thanks for the clarification.

spyder/plugins/editor/utils/switcher.py
```diff
@@ -9,11 +9,205 @@
...
+        result_paths = ["...{}".format(
+                        path.rstrip(os.sep).split(common_prefix)[-1])
+                        for path in new_path_list]
+    else:
+        result_paths = [path.rstrip(os.sep) for path in new_path_list]
+
+    return result_paths
+
+
+def get_file_icon(path):
```
ccordoba12: I think this function also needs to go to `spyder/utils/switcher.py`.

spyder/plugins/editor/utils/switcher.py
```diff
@@ -9,11 +9,205 @@
...
+    """Get icon for file by extension."""
+
+    if sys.platform == 'darwin':
+        scale_factor = 0.9
+    elif os.name == 'nt':
+        scale_factor = 0.8
+    else:
+        scale_factor = 0.7
```
ccordoba12: Let's change this to 0.6 for Linux.

dalthviz:
@ccordoba12 @goanpeca a preview with the latest changes:
s

ccordoba12:
It looks amazing!! Thanks a lot @dalthviz for your hard work on this one!

I think this is ready. @goanpeca, do you have some final words to say?

spyder/app/mainwindow.py
```diff
@@ -3223,9 +3223,9 @@ def open_switcher(self, symbol=False):
             self.switcher.setup()
         self.switcher.show()
 
-        # Note: The +1 pixel on the top makes it look better
+        # Note: The +6 pixel on the top makes it look better
```
ccordoba12: Probably +1 pixel is needed in macOS and +6 in Linux/Windows.

@goanpeca, could you check this?

goanpeca: Looks good!

Screen Shot 2019-09-11 at 17 39 54

goanpeca:
>I think this is ready. @goanpeca, do you have some final words to say?

It looks really good @dalthviz ! :-)

My only comment is regarding the size of the Switcher.

Should it update based on which configuration?

For instance, let's assume I have a lot of zoom on the editor because...
Or if I set the appearance of the fonts to be larger

Screen Shot 2019-09-11 at 17 43 11

Then we get this:
image

So I think we should use perhaps the settings from the appearance? Thoughts @ccordoba12, @dalthviz ?

PR #10072, opened 2019-08-24
PR: Hide size/type columns and add show/hide actions to header (right click) and plugin opts.

goanpeca:
## Description of Changes

Hide size/type (or kind on mac) columns on file explorer and a add show/hide actions to header with right click (context menu) and plugin options.

filexplorer

- Explorer is now on the left and tabbed with the project explorer
- Size/Modified/Kind Columns have been removed
- Sorting by these columns is still possible via Context menu

### Issue(s) Resolved

Fixes #

### Affirmation

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -1655,14 +1655,14 @@ class MainWindow(QMainWindow):
         s_layout = {
             'widgets': [
                 # Column 0
-                [[explorer_file, explorer_project]],
+                [[explorer_project]],
                 # Column 1
                 [[editor]],
                 # Column 2
                 [[outline]],
                 # Column 3
                 [[help_plugin, explorer_variable, plots,     # Row 0
-                  helper, finder] + plugins,
+                  helper, explorer_file, finder] + plugins,
                  [console_int, console_ipy, history]]        # Row 1
                 ],
             'width fraction': [15,            # Column 0 width
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -1655,18 +1655,18 @@ class MainWindow(QMainWindow):
         s_layout = {
             'widgets': [
                 # Column 0
-                [[explorer_project]],
+                [[explorer_file, explorer_project]],
                 # Column 1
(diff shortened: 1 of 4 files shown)


ccordoba12:
My comments:

* Let's leave the "Last modified" column so that it's easier for users to visually distinguish between the Files and Project panes (else they are two similar and can be very easily confused).

* That change will make us closer to the JupyerLab interface:

    Selección_024

* I don't know if this is worth to be added for Spyder 4 or if we should leave it for Spyder 5. The thing is I'd also like to have a left sidebar associated with this (as JupyterLab and VSCode) so that users can easily collapse or show those panes (i.e. Files and Project). That would prevent taking valuable horizontal space from users' screens without an easy way to revert that.

goanpeca:
>I don't know if this is worth to be added for Spyder 4 or if we should leave it for Spyder 5. The thing is I'd also like to have a left sidebar associated with this (as JupyterLab and VSCode) so that users can easily collapse or show those panes (i.e. Files and Project). That would prevent taking valuable horizontal space from users' screens without an easy way to revert that.

Ok we can move to 5 then

goanpeca:
I created the milestone

ccordoba12:
I think for Spyder 4 we could leave the "Name" and "Last modified" columns and add options to show/hide "Size" and "Type". What do you think?

goanpeca:
>I think for Spyder 4 we could leave the "Name" and "Last modified" columns and add options to show/hide "Size" and "Type". What do you think?

Ahh I did not understand. Sure, let me fix it then.

goanpeca:
>show/hide "Size" and "Type". What do you think?

Should this go on the context menu of the header? or the pane menu?

ccordoba12:
> Ahh I did not understand. Sure, let me fix it then.

Sorry, I think I was not clear enough either. For Spyder 4 I agree with simplifying the default view of Files by leaving "Name" and "Last modified" by default, but I don't agree with moving Files next to Project.

> or the pane menu?

I think it should go in the Options menu, yes.

goanpeca:
image

PR #10082, opened 2019-08-26
PR: Change the current cell color of the Spyder Dark theme

ccordoba12:
I was never satisfied with the current color because it doesn't have enough contrast with respect to the background one. However, I think this new color has.

**Before**

Selección_026

**After**

Selección_027

@goanpeca, what do you think?

### DIFF
--- a/spyder/config/appearance.py
+++ b/spyder/config/appearance.py
@@ -154,7 +154,7 @@ APPEARANCE = {
     #           Name             Color     Bold  Italic
     'spyder/dark/background':  "#19232D",
     'spyder/dark/currentline': "#3a424a",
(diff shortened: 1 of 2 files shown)


goanpeca:
Looks good @ccordoba12. Question, there color are taken from some sort of palette generator online or are more tweaked by hand. The palette generator might help keeping them in sync with the overall them

ccordoba12:
> Question, there color are taken from some sort of palette generator online or are more tweaked by hand

Spyder dark is very similar to "One Dark Pro" (the Atom default theme), but adapted to look good with QDarkstyle.

I took this color in particular from

https://github.com/material-theme/vsc-material-theme

which has some variants similar to "One Dark Pro".

PR #10085, opened 2019-08-26
PR: Stop IPython console channels faster

impact27:
## Description of Changes

This avoids messages of the type `WARNING:traitlets:kernel died: 18.0` that are created by #10066. I forgot that I already solved this problem in https://github.com/jupyter/jupyter_client/pull/420, so I added a dependency on jupyter_client 5.2.5+. 

### Issue(s) Resolved

Fixes #

### Affirmation

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,7 +242,6 @@ a Python version equal or greater than 3.5 or 2.7.
 * **Numpydoc**: Used by Jedi to get function return types from Numpydocstrings.
 * **Cloudpickle**: Serialize variables in the IPython kernel to send to Spyder.
 * **spyder-kernels** 1.4.0+: Jupyter kernels for the Spyder console.
-* **jupyter_client** 5.2.5+: Client for Jupyter kernels.
 * **keyring**: Save Github credentials to report errors securely.
 * **QDarkStyle** 2.6.4+: A dark stylesheet for Qt applications, used for Spyder's dark theme.
 * **atomicwrites**: Atomic file writes.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -242,6 +242,7 @@ a Python version equal or greater than 3.5 or 2.7.
 * **Numpydoc**: Used by Jedi to get function return types from Numpydocstrings.
 * **Cloudpickle**: Serialize variables in the IPython kernel to send to Spyder.
 * **spyder-kernels** 1.4.0+: Jupyter kernels for the Spyder console.
+* **jupyter_client** 5.2.5+: Client for Jupyter kernels.
```
ccordoba12: Please move this dependency to spyder-kernels (which already has a minimal required version of jupyter_client).

spyder/plugins/ipythonconsole/widgets/client.py
```diff
@@ -504,7 +504,7 @@ def shutdown(self):
             self.shellwidget.spyder_kernel_comm.close()
             self.shellwidget.kernel_manager.shutdown_kernel(now=now)
         if self.shellwidget.kernel_client is not None:
-            self.shellwidget.kernel_client.shutdown()
```
ccordoba12: What's the difference between `shutdown` and `stop_channels`?

impact27: `shutdown` stops the kernel, but is already called in `self.shellwidget.kernel_manager.shutdown_kernel(now=now)`. `stop_channels` stops the channels, including the heartbeat channel. If the heartbeat channel is not stopped but the kernel is down, the channel will notice and send `WARNING:traitlets:kernel died: 18.0`

impact27:
There was still a problem where spyder would hang if closing while closing a console. It is solved here: https://github.com/jupyter/jupyter_client/pull/470.

PR #10087, opened 2019-08-26
PR: Separate pdb input from regular input

impact27:
## Description of Changes
A few part of the code now use the pdb command line to send messages to pdb while debugging. (For example, to ask pdb to check for incoming comm messages.) These change allow the frontend to separate between `input` calls that are done during debugging, and that should not recieve these messages, and pdb input. The detection is done by comparing the prompt to 'ipdb> '. 

All the direct calls to `input` or `write_to_stdin` are replaced by a new function `pdb_execute`, as this is what these calls are used for.

### Issue(s) Resolved

Fixes #

### Affirmation

### DIFF
--- a/requirements/conda.txt
+++ b/requirements/conda.txt
@@ -16,7 +16,7 @@ pyxdg
 pyzmq
 qdarkstyle >=2.7
 qtawesome >=0.5.7
-qtconsole >=4.5.2
+qtconsole >=4.5.5
 qtpy >=1.5.0
 sphinx
 spyder-kernels >=1.4.0
--- a/requirements/conda.txt
+++ b/requirements/conda.txt
@@ -16,7 +16,7 @@ pyxdg
 pyzmq
 qdarkstyle >=2.7
 qtawesome >=0.5.7
-qtconsole >=4.5.5
+qtconsole >=4.5.2
 qtpy >=1.5.0
 sphinx
 spyder-kernels >=1.4.0
--- a/spyder/plugins/ipythonconsole/widgets/debugging.py
+++ b/spyder/plugins/ipythonconsole/widgets/debugging.py
@@ -195,13 +195,13 @@ class DebuggingWidget(RichJupyterWidget):
         self._previous_prompt = (prompt, password)
 
         if print_prompt:
-            self._highlighter.highlighting_on = True
             # Reset reading in case it was interrupted
             self._reading = False
             self._readline(prompt=prompt, callback=self._readline_callback,
                            password=password)
 
         if self.is_waiting_pdb_input():
+            self._highlighter.highlighting_on = True
             self._pdb_input_ready = True
             self._executing = False
 
--- a/spyder/plugins/ipythonconsole/widgets/debugging.py
+++ b/spyder/plugins/ipythonconsole/widgets/debugging.py
@@ -86,6 +86,7 @@ class DebuggingWidget(RichJupyterWidget):
                     self._get_end_cursor().position())
             self._input_ready = False
             # Match ConsoleWidget.do_execute
+            self._input_buffer_executing = self.input_buffer
             self._executing = True
             return self.kernel_client.input(line)
 
--- a/spyder/plugins/ipythonconsole/widgets/debugging.py
+++ b/spyder/plugins/ipythonconsole/widgets/debugging.py
@@ -184,8 +184,7 @@ class DebuggingWidget(RichJupyterWidget):
         # Check if this is a duplicate that we shouldn't reprint.
         # This can happen when sending commands to pdb from the frontend.
         print_prompt = (not self._reading
-                        or not (prompt, password) == self._previous_prompt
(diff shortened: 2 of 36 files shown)


spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -122,7 +126,48 @@ def refresh_from_pdb(self, pdb_state):
...
+        self.add_to_pdb_history(self._pdb_line_num, line)
+
+        if line:
+            self._last_pdb_cmd = line
+
+        # This is the Spyder addition: add a %plot magic to display
+        # plots while debugging
+        if line.startswith('%plot '):
+            line = line.split()[-1]
+            line = "__spy_code__ = get_ipython().run_cell('%s')" % line
```
impact27: I am not sure what this code does. Should we keep it? What is it supposed to do?

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -57,35 +57,39 @@ def __init__(self, *args, **kwargs):
...
+            # Print the string to the console
+            if not hidden:
+                # This emulates an user interaction
+                self._control.insert_text(line + '\n')
+                self._reading = False
+                self._append_before_prompt_cursor.setPosition(
+                    self._get_end_cursor().position())
+            self._input_ready = False
+            # Match ConsoleWidget.do_execute
+            self._executing = True
```
impact27: Should `self._executing = True` only be called when `not hidden`?

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -133,59 +178,36 @@ def _handle_input_request(self, msg):
...
 
-            # Save history to browse it later
-            self._pdb_line_num += 1
-            self.add_to_pdb_history(self._pdb_line_num, line)
+        self._previous_prompt = (prompt, password)
 
-            if line:
-                self._last_pdb_cmd = line
+        if print_prompt:
+            self._highlighter.highlighting_on = True
```
impact27: Should the highlighter only be turned on for a pdb request? Here it is turned on for any request.

spyder/plugins/ipythonconsole/widgets/debugging.py
impact27: I updated the PR to handle this more cleanly.

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -122,7 +149,47 @@ def refresh_from_pdb(self, pdb_state):
...
+        # Fixes 2034
+        self.pdb_execute('continue')
+
     # ---- Private API (overrode by us) ----------------------------
+    def _readline_callback(self, line):
+        """Callback used when the user inputs text in stdin."""
+        if not self.is_waiting_pdb_input():
+            if self._input_ready:
+                # This is a regular input call
+                self._finalize_input_request()
```
impact27: This is maybe something that should be changed in qtconsole as well. I will open an issue there.

impact27: https://github.com/jupyter/qtconsole/issues/370

impact27:
This will pass the tests when https://github.com/jupyter/qtconsole/pull/371 is merged.

impact27:
Summary of work done:
 - Created a `pdb_execute` function to execute statments in pdb kernel while debugging. This replace the `stdin` and `input` calls.
 - Properly detect if a pdb command loop is running. Separate between pdb requests and general `input` requests. 
 - Remove debugging handling from `KernelComm` and place it in `debugging.py`.

requirements/conda.txt
```diff
@@ -16,7 +16,7 @@ pyxdg
 pyzmq
 qdarkstyle >=2.7
 qtawesome >=0.5.7
-qtconsole >=4.5.2
+qtconsole >=4.5.5
```
ccordoba12: Please revert this change because it's not necessary given that we're pulling qtconsole from Github and we also need to wait until Anaconda releases packages for this version.

spyder/plugins/ipythonconsole/widgets/debugging.py
ccordoba12: It's used to generate inline plots while debugging by calling this `%plot` pseudo-magic I added to our debugger some time ago. Without this trick, plots can't be seen while debugging.

If you think this can implemented better, please leave it for a another PR.

spyder/plugins/ipythonconsole/widgets/debugging.py
ccordoba12: Yes, please only turn it on for Pdb requests.

impact27:
I have no idea why `spyder/app/tests/test_mainwindow.py::test_lock_action` would fail. It doesn't fail on my computer.

impact27:
re triggering the tests

Issue #10029, opened 2019-08-16
ignore `ConnectionError` on Kite completions request


metalogical:
## Problem Description

If Kite is running when Spyder is started, but then exits (e.g. crashes, or manually), Spyder reports `ConnectionError`s whenever making completions requests.

### What steps reproduce the problem?

1. Start Kite
2. Start Spyder
3. Wait until Kite completions are working
4. Quit Kite
5. Type in Spyder

### What is the expected output? What do you see instead?

No Kite completions should show up, and no error should be reported, since I quit Kite. Instead, I see an error:

```python-traceback

raise ConnectionError(e, request=request)
requests.exceptions.ConnectionError: HTTPConnectionPool(host='localhost', port=46624): Max retries exceeded with url: /clientapi/editor/event (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fe1e8a996d0>: Failed to establish a new connection: [Errno 61] Connection refused'))

```

## Versions

* Spyder version: master
* Python version: 3.7
* Qt version: 5.9.7
* PyQt version: 5.9.2
* Operating System name/version: Darwin 10.12.6

ccordoba12:
Thanks @metalogical for the report!

ccordoba12:
Thanks @metalogical for the confirmation!

PR #10088, opened 2019-08-26
PR: Always enable Kite completions

metalogical:
## Description of Changes

Previously, we were using logic to ping Kite up to 40 times before considering Kite unreachable, and permanently disabling the Kite completions plugin.

It is much more robust to always make Kite completions requests, and let the connection error out. The `ConnectionError` happens very quickly, so this has no noticeable performance impact. There are many other types of I/O happening per-keystroke that are much slower.

In order for the user experience to be pleasant, this depends on completely ignoring `ConnectionError`s: #10029. This PR does not address that issue.

### Affirmation

@metalogical

### DIFF
--- a/spyder/plugins/completion/kite/__init__.py
+++ b/spyder/plugins/completion/kite/__init__.py
@@ -14,7 +14,6 @@ class _KiteEndpoints(type):
     KITE_PORT = 46624
     KITE_URL = 'http://localhost:{0}'.format(KITE_PORT)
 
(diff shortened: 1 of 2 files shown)


ccordoba12:
@metalogical, quick question: Does this fix issue #10029 or do you plan to fix that in another PR?

ccordoba12:
Thanks @metalogical for your help and @andfoy for the review!

PR #10115, opened 2019-08-31
PR: Run and debug files without saving them

### LINKED ISSUE #1643: Spyder does not fully support IPython scripts (*.ipy)
_From [DavidAnt...@gmail.com](https://code.google.com/u/108698404730356683634/) on 2013-11-13T17:36:56Z_

IPython scripts ending in .ipy are similar to regular python files, with the critical difference that they support IPython magics such as &#37;timeit.

Under Spyder's IPython console, I can &#37;run my_script.ipy with no problems. Under Spyder, I can load .ipy files, which are a recognised extensions. However, when clicking on "run", a syntax error is generated, since the script is run in such a way that the IPython magics are not recognised.

Clearly, Spyder's runfile() is doing something different from IPython's &#37;run.

Additionally the Spyder editor does not have an icon for .ipy files, and pyflakes marks IPython magics as errors. (I appreciate that solving the latter could easily be far more hassle than it's worth)

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=1643_

impact27:
## Description of Changes
Add `save_new_files` option to the `save()` function to allow running files that are not saved. 
A known limitation is that the code doesn't allow to set breakpoints in non-existing files, but this can be address in a later PR.  Works with https://github.com/spyder-ide/spyder-kernels/pull/167

### Affirmation

video


spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -483,16 +483,35 @@ def get_editorstack(self):
...
+        editorstack.save_all(save_new_files=False)
+        editor = self.get_editor(filename)
+
+        if editor is None:
+            # load from file
+            mode = 'rb'
+            if PY2:
+                mode = 'r'
+            with open(filename, mode) as f:
+                return f.read()
```
impact27: Instead of reading the file from memory here, we could raise an error and let the kernel load the file (Which it already does if this function raises an error).

ccordoba12: I don't understand. When can this case happen?

impact27: Instead of reading the file from memory, the kernel asks the frontend for the file content, and it this doesn't work, loads it from memory. In the case where the file is not open in the frontend, the frontend can either:
 - Raise an error to notify the kernel that the file is not open, so the kernel can load from memory
 - load the file from memory and send this content.
I like the first option better to avoid code duplication and having to send large files over the comm without reason.

ccordoba12: What if the kernel lives in a remote server and tries to open a file there with the same name and location but that it doesn't correspond to the one the user wants to really run?

ccordoba12:
By the way, the test failure on Linux and Python 2 look legit (the same test fails every time). So please take a look at that.

spyder/plugins/ipythonconsole/widgets/shell.py
impact27: > By the way, the test failure on Linux and Python 2 look legit (the same test fails every time). So please take a look at that.

I think this is fixed in qtconsole#375

impact27:
> By the way, the test failure on Linux and Python 2 look legit (the same test fails every time). So please take a look at that.

I think this is fixed in qtconsole#375

spyder/plugins/ipythonconsole/widgets/shell.py
impact27: Should the file be with the frontend or with the kernel? IN the current master the file lives with the kernel. Should that change?

bcolsen: > Should the file be with the frontend or with the kernel?

I think that a basic user would expect the 'F5' shortcut to run the file they are looking at in Spyder regardless of what is on the remote file system. Advanced users would be surprised that it works at all.

Would it simplify anything if the `runfile` could only run file that are open in spyder and if user want to run a file from the file system they could use ipython's %run magic.

We could even change the name of `runfile`

impact27: > Would it simplify anything if the runfile could only run file that are open in spyder and if user want to run a file from the file system they could use ipython's %run magic.

> We could even change the name of runfile

Or that runfile would open the file in the editor. (As could runcell). This would make running things from history easier. Also it could be a bit surprising if calling runfile gives you:
`File script.py not open in the editor`. It is already a bit surprising that this happens with runcell.

I would say that `runfile` should be used to run things from the frontend side, and `%run` from the kernel side. The error message could say: "the editor can not open `/path/on/kernel/machine.py`" so the problem would be obvious to the advanced user.

impact27:
Maybe not. The problem seems to be similar though: `In [3]: !import os; os.getcwd()\n   ...: ` somehow `qtbot.keyClick(control, Qt.Key_Enter)` doesn't execute the line but adds a continuation prompt. (this should have been fixed in jupyter/qtconsole#375). Could someone with linux have a look at this? An easy fix would be to add a shift modifier, but this wouldn't solve the underlying issue.

impact27:
TODO: add tests

impact27:
The tests are added in #10428

runtests.py
```diff
@@ -32,7 +32,7 @@ def main(run_slow=False, extra_args=None):
     """
     Run pytest tests for Spyder.
     """
-    pytest_args = ['spyder',
+    pytest_args = ['spyder/app/tests/test_mainwindow.py::test_debug_unsaved_file',
```
goanpeca: Why is this change needed @impact27 ?

impact27: I removed this commit. The test was failing and I was trying to understand what was happening. The tests should be passing now.

goanpeca: 👍

ccordoba12:
@impact27, please update this PR and also post a screenshot in the description of how things looks while debugging an unsaved file.

impact27:
> @impact27, please update this PR and also post a screenshot in the description of how things looks while debugging an unsaved file.

I am not sure what you want me to update. Also this PR doesn't add anything to the GUI. It just avoids a pop-up when you run a file.

PR #10150, opened 2019-09-06
PR: Avoid RuntimeError when handling completion responses

ccordoba12:
These errors were triggered when a CodeEditor instance was removed before responses can be processed by it.


ccordoba12:
@andfoy, do we need something similar for Kite? I mean, I don't know if Kite holds CodeEditor instances as our LSP plugin does.

andfoy:
@ccordoba12, this PR handles all CodeEditor references, including both Kite and LSP

ccordoba12:
Nice! Thanks for the confirmation!

Issue #10144, opened 2019-09-05
Options for the LSP server are not being updated.


goanpeca:
## Problem Description

Show code style / Show doc style options are not working from source or preferences menu.

### What steps reproduce the problem?

1. Click `Source` menu and the unselect `Show code style warnings`
2. Warnings do not disappear

### What is the expected output? What do you see instead?

Warnings do not disappear

![warnings](https://user-images.githubusercontent.com/3627835/64303976-29f89100-cf50-11e9-95b2-a4c93bc9542e.gif)

Latest master

pygments >=2.0    :  2.3.1 (OK)
qdarkstyle >=2.6.4:  2.7 (OK)
sphinx >=0.6.6    :  1.8.5 (OK)
psutil >=0.3      :  5.6.1 (OK)
pyls >=0.27.0     :  0.27.0 (OK)
nbconvert >=4.0   :  5.4.1 (OK)
pandas >=0.13.1   :  0.24.2 (OK)
numpy >=1.7       :  1.16.2 (OK)
pympler >=0.7     :  None (NOK)
sympy >=0.7.3     :  1.4 (OK)
cython >=0.21     :  0.29.12 (OK)
qtconsole >=4.5.2 :  4.5.4 (OK)
IPython >=4.0;<6.0:  5.8.0 (OK)
matplotlib >=2.0.0:  2.2.3 (OK)
pylint >=0.25     :  1.9.2 (OK)

ccordoba12:
Please look for all PyLS and transport scripts running in your system and kill them. Then restart Spyder to see if that helps.

ccordoba12:
I found a very simple way to reproduce this: using two Spyder instances. The problem is serious because every time users change an LSP option in the second instance, its related services (completion and linting) stop working completely.

ccordoba12:
This is a duplicate of issue #9933.

PR #10164, opened 2019-09-09
PR: Fix PyLS options not being updated correctly when a previous server was running

### LINKED ISSUE #9933: PyLS settings are not being applied live
@andfoy, after PR #9704 the PyLS preferences are not being applied live (I just verified that, although I haven't noticed it before).

For example, you can go to the menu

    Source > Show code style warnings

and see that activating/deactivating that option doesn't do anything after PR #9704, whereas it shows/hides pycodestyle errors before that.

ccordoba12:
This PR avoids touching LSP server configurations when selecting a free port for the server. That was triggering a server restart, which (due to another bug, fixed here too) was stopping LSP services for the current open files.

This PR also renamed the PyLS option about the "command" we use to start it. Since we really use `sys.executable -m <command>`, that option doesn't correspond to a command (or executable) in your system, but to a Python module that implements the server.

Fixes #9933

### DIFF
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -490,7 +490,7 @@ DEFAULTS = [
               'pydocstyle/ignore': '',
               'pydocstyle/match': '(?!test_).*\\.py',
               'pydocstyle/match_dir': '[^\\.].*',
-              'advanced/command_launch': 'pyls',
+              'advanced/module': 'pyls',
               'advanced/host': '127.0.0.1',
               'advanced/port': 2087,
               'advanced/external': False,
@@ -590,4 +590,4 @@ NAME_MAP = {
 #    or if you want to *rename* options, then you need to do a MAJOR update in
 #    version, e.g. from 3.0.0 to 4.0.0
 # 3. You don't need to touch this value if you're just adding a new option
-CONF_VERSION = '52.1.0'
+CONF_VERSION = '53.0.0'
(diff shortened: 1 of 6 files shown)


spyder/config/main.py
```diff
@@ -490,7 +490,7 @@
               'pydocstyle/ignore': '',
               'pydocstyle/match': '(?!test_).*\\.py',
               'pydocstyle/match_dir': '[^\\.].*',
-              'advanced/command_launch': 'pyls',
+              'advanced/module': 'pyls',
```
andfoy: I think this shouldn't be called module, as it may contain other commands beyond Python

ccordoba12: You mean that we could test things with Microsoft's LSP for Python?

andfoy: For instance, or other executables such as Elixir-LS

ccordoba12: This change only applies to the Python language server (I didn't touch the dialog for other languages).

The thing is we're running the PyLS with this command:

    sys.executable -m pyls --tcp --host {host} --port {port}

As you can see, users need to provide a Python module (not a Python script or another command) to run the server. That's why our current text in Preferences is misleading and I think it's better to change it.

Of course, this rules out using Microsoft's Python server for now. But we're totally integrated with the PyLS, so trying to support it would require a big refactoring (which could be work for Spyder 5 or even 6, but certainly not for version 4).

goanpeca:
@ccordoba12 

>@goanpeca, could you test this one to see if it fixes what you reported on issue #10144? Thanks!

I confirm this fixes the problems I was seeing 🎉 !

ccordoba12:
> I confirm this fixes the problems I was seeing

Great! I think this is a huge step forward!!

> (added the fix# to the description)

I closed #10144 as a duplicate of #9933 because both are basically the same.

PR #9850, opened 2019-07-17
[WIP] PR: Implement code snippets for our completion clients

andfoy:
## Description of Changes

* [ ] Wrote at least one-line docstrings (for any new functions)
* [ ] Added unit test(s) covering the changes (if testable)

* [ ] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

This PR introduces code snippet processing on Spyder, these code snippets can be produced by LSP servers and also by Kite.

### Issue(s) Resolved

Fixes #

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct:
@andfoy

PR #10166, opened 2019-09-09
PR: Add status bar widget for Kite

dalthviz:
## Description of Changes

is

### Issue(s) Resolved

Fixes #

### Affirmation

I certify the above statement is true and correct: @dalthviz


spyder/plugins/completion/kite/status.py
```diff
@@ -0,0 +1,31 @@
...
+    """Status bar widget for Kite completions status."""
+
+    def __init__(self, parent, statusbar):
+        super(KiteStatus, self).__init__(parent, statusbar,
+                                         icon=ima.get_kite_icon())
+
+    def update(self, kite_enabled):
+        """Update kite completions status."""
+        self.set_value(_("Kite: enabled"))
+        self.setVisible(bool(kite_enabled))
```
goanpeca: On VSCode they display if kite is not running (and kite is installed) the following:

Screen Shot 2019-09-10 at 09 31 54

Maybe we should check if Kite is installed and running (display the current behavior) and if it is installed and NOT running something like the screenshot. And if kite is not installed then we hide the status widget (or not even create it?)

Also this needs to be a polling Status Widget, since Kite might be opened and the status will (and should change)

I guess Kite guys will like something similar @ccordoba12 ?

spyder/plugins/editor/plugin.py
```diff
@@ -150,6 +150,8 @@ def __init__(self, parent, ignore_last_opened_files=False):
         self.dialog_size = None
 
         statusbar = parent.statusBar()  # Create a status bar
+        # Remove separator line
+        statusbar.setStyleSheet('QStatusBar::item {border: None;}')
```
goanpeca: @dalthviz why was this necessary?

spyder/widgets/status.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, parent, statusbar, icon=None):
         self.set_icon(icon)
 
         # See spyder-ide/spyder#9044.
-        self.text_font = QFont(get_font(option='font'))
+        self.text_font = QFont(QFont().defaultFamily())
```
goanpeca: @dalthviz why was this necessary?

spyder/plugins/editor/plugin.py
ccordoba12: This is needed to remove some white vertical lines between widgets that appear on Windows.

spyder/widgets/status.py
ccordoba12: I asked @dalthviz to use the default system for our status bar instead of the monospaced one used in the Editor.

spyder/plugins/completion/kite/status.py
ccordoba12: > I guess Kite guys will like something similar

Ok, I agree. @dalthviz, please follow @goanpeca's suggestions.

spyder/plugins/editor/plugin.py
goanpeca: Is this a QDarkstyle issue, or something from Spyder alone?

spyder/widgets/status.py
goanpeca: Ahhh, ok :-) Got it!

spyder/plugins/editor/plugin.py
dalthviz: Basically when using the dark theme the lines are more visible (since they are white). With the light theme the lines are also visible but are light gray:

image


image

goanpeca: Hmmmm ok 

@dalthviz could you submit this on Qdarkstyle repo, as this should not be the case by default (i think)

spyder/plugins/completion/kite/status.py
```diff
@@ -0,0 +1,45 @@
...
+    def __init__(self, parent, statusbar):
+        super(KiteStatus, self).__init__(parent, statusbar,
+                                         icon=ima.get_kite_icon())
+
+    def import_test(self):
+        """Raise ImportError if feature is not supported."""
+        try:
+            from spyder.plugins.completion.kite.plugin import (
+                KiteCompletionPlugin)
+            KiteCompletionPlugin.kite_status()
```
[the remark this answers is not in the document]
ccordoba12: Ok, I noticed that we're not saving `parent` in `StatusBarWidget` (the base class of `BaseTimerStatus`), but I think we could do that to avoid adding and calling class methods.

@goanpeca, what do you think?

spyder/plugins/completion/kite/plugin.py
```diff
@@ -21,12 +21,17 @@
...
+from spyder.plugins.completion.kite.status import KiteStatus
 
 # Third-party imports
 import psutil
 
 logger = logging.getLogger(__name__)
 
+NOT_INSTALLED = 'Not installed'
+RUNNING = 'Running'
+NOT_RUNNING = 'Not running'
```
ccordoba12: Let's move these constants to be part of `KiteCompletionPlugin` so we don't have to import them separately. You can place them below `COMPLETION_CLIENT_NAME`.

spyder/plugins/completion/kite/status.py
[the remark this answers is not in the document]
dalthviz: I think we would need to use a new argument then since `parent` needs to be a `QWidget` (`KiteCompletionPlugin` isn't a QWidget), and I can add that to the `KiteStatus` subclass

ccordoba12:
@dalthviz, tests are failing due to an error unrelated to this. I'm fixing them in PR #9850.

spyder/plugins/completion/kite/status.py
metalogical: I agree with this, with the extra edge case:

if Kite is not detected as installed, but is enabled by the user (as per the option in #10169), we should probably continue to display this widget.

spyder/plugins/completion/kite/status.py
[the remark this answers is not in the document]
goanpeca: Maybe we can have `main` and `parent` as the parameters of the constructor and have the parent stored as `self._parent` and main as `self._main` (or calling it as the actual parent in the super method

spyder/plugins/completion/kite/status.py
```diff
@@ -0,0 +1,69 @@
...
+                if (client_status['status'] == 'ready'
+                        or client_status['status'] == 'unsupported'):
+                    self.open_file_updated = False
+            else:
+                kite_status = 'not reacheable'
+                self.tooltip = self.BASE_TOOLTIP
+                self.open_file_updated = False
+        else:
+            self.tooltip = self.BASE_TOOLTIP
+        text = '𝕜𝕚𝕥𝕖: {}'.format(kite_status)
```
its-dhung: If we are showing the Kite icon, we don't need to show `𝕜𝕚𝕥𝕖` logo since it's redundant. This saves a little bit of horizontal space on the status bar as well.

dalthviz:
A new preview:

s

spyder/plugins/completion/kite/utils/status.py
```diff
@@ -16,6 +16,10 @@
 # Third-party imports
 import psutil
 
+NOT_INSTALLED = 'not installed'
+RUNNING = 'running'
```
its-dhung: Change this to
```python
RUNNING = 'ready'
```
This will prevent the status indicator from flickering between `running` and `ready` as the user clicks around the editor UI.

spyder/plugins/completion/kite/client.py
```diff
@@ -68,6 +68,16 @@ def get_languages(self):
             response = ['python']
         return response
 
+    def get_status(self, filename, saved=True):
+        """Get Kite status taking into account a filename."""
+        if saved:
+            verb, url = KITE_ENDPOINTS.SAVED_STATUS_ENDPOINT
+            url = url.format(filename=filename)
+        else:
+            verb, url = KITE_ENDPOINTS.UNSAVED_STATUS_ENDPOINT
```
metalogical: If we have a filename, we should always use the `filename=` URL.

The `filetype=` URL is for cases where the user is editing a python buffer but there is no filename (and there is no corresponding file on disk).

The side effect of getting this wrong is that there is a lot of status bar noise claiming that Kite has no index for the file, once the file goes into an unsaved state, even though that is not true.

Sorry if this wasn't clear in the spec.

spyder/plugins/completion/kite/utils/status.py
metalogical: (@its-dhung we should take note of this on our side so that we know changing the short text to something other than `"ready"` in the Kite status response may end up causing flickering on Spyder)

its-dhung: Will make a note of this in the API doc.

dalthviz:
A new preview:

status

metalogical:
awesome! this looks good on my end

adamsmith:
A quick design comment: the Kite icon is too big and too high contrast. There are a couple of possible solutions for this. I think the first/main solution to try is making it smaller — closer to the size of the box for the “conda” text.

Here is a mockup with the logo at **85% of its current size**:

image

dalthviz:
@ccordoba12 I think this is ready for a review

spyder/plugins/completion/kite/plugin.py
```diff
@@ -64,18 +81,34 @@ def http_client_ready(self, languages):
...
         Called when the setup of the main window finished
-        to let us do onboarding if necessary
+        to let us setup elements like installation and
+        onboarding if necessary
         """
-        self._show_installation_dialog()
+        self.show_installation_dialog()
+
+    @Slot(str)
+    @Slot(object)
```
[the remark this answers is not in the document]
dalthviz: It's a dict

spyder/plugins/completion/kite/widgets/status.py
```diff
@@ -0,0 +1,65 @@
...
+        elif value is None:
+            value = self.DEFAULT_STATUS
+            self.tooltip = self.BASE_TOOLTIP
+
+        self.update_tooltip()
+        self.setVisible(value != NOT_INSTALLED or kite_enabled)
+
+        super(KiteStatusWidget, self).set_value(value)
+
+    def get_tooltip(self):
```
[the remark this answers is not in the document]
dalthviz: This is a reimplementation of the method `StatusBarWidget` class to handle the update of the tooltip. Should I add that this is a reimplementation on the docstring?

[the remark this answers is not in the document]
ccordoba12: Yes, please do that.

spyder/plugins/completion/kite/widgets/status.py
```diff
@@ -16,50 +16,52 @@
...
-        self.set_value(None)
+        self.tooltip_widget = ToolTipWidget(self, as_tooltip=True)
+        is_installed, _ = check_if_kite_installed()
+        self.setVisible(is_installed)
 
     def set_value(self, value):
         """Return Kite completions state."""
-        kite_enabled = self.plugin.get_option('enable')
+        is_installing = self.plugin.is_installing()
+        installation_status = self.plugin.installation_cancelled_or_errored()
```
ccordoba12: Are you sure we should show the widget if the installation erred? That doesn't make much sense to me.

Could you explain your reasoning?

spyder/plugins/completion/kite/widgets/status.py
```diff
@@ -16,50 +16,52 @@
...
             if value == NOT_INSTALLED:
                 return
         elif value is None:
             value = self.DEFAULT_STATUS
             self.tooltip = self.BASE_TOOLTIP
-
         self.update_tooltip()
-        self.setVisible(value != NOT_INSTALLED or kite_enabled)
-
+        self.setVisible(value != NOT_INSTALLED)
```
ccordoba12: I agree with hiding the widget if kite is not enabled (as it was before this change).

spyder/plugins/editor/plugin.py
```diff
@@ -320,9 +320,9 @@ def send_completion_request(self, language, request, params):
         self.main.completions.send_request(language, request, params)
 
     def kite_completions_file_status(self):
-        """Connect open file update to Kite completion."""
+        """Connect open_file_update to Kite update the status."""
```
ccordoba12: Let's change this to be

> Connect open_file_update to Kite's status

spyder/plugins/completion/kite/client.py
```diff
@@ -29,12 +31,15 @@ class KiteClient(QObject, KiteMethodProviderMixIn):
     sig_client_started = Signal(list)
     sig_client_not_responding = Signal()
     sig_perform_request = Signal(int, str, object)
+    sig_perform_status_request = Signal(str)
+    sig_status_response_ready = Signal(object)
```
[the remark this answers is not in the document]
dalthviz: @ccordoba12 the thing here is that the method that emits the signal (`get_status`) can send a `str` or a `dict`. Should I made the method emit only a `dict`?

PR #10167, opened 2019-09-09
PR: Enable Pylint linting in the editor

ccordoba12:
This PR allows users to select between Pyflakes and Pylint for linting.

Current implementation:

Selección_001

TODO:

- [ ] Add a way to pass a confg file to Pylint (it throws too much info without it).
- [ ] Allow users to select the threshold in McCabe.

This PR is part of issue #7724.


goanpeca:
Note to @ccordoba12. So we can look for pylint configuration on the home directory and if none is found create one with some default options that make sense. We can also inform the user that that file is being used with a label, and even allow to open it within the editor with a hyperlink o the QLabel?

Form [pylint docs](https://docs.pylint.org/en/1.6.0/run.html)

```
Specifying all the options suitable for your setup and coding standards can be tedious, so it is possible to use a configuration file to specify the default values. You can specify a configuration file on the command line using the --rcfile option. Otherwise, Pylint searches for a configuration file in the following order and uses the first one it finds:

1. pylintrc in the current working directory
2. .pylintrc in the current working directory
3. If the current working directory is in a Python module, Pylint searches up the hierarchy of Python modules until it finds a pylintrc file. This allows you to specify coding standards on a module-by-module basis. Of course, a directory is judged to be a Python module if it contains an __init__.py file.
4. The file named by environment variable PYLINTRC
if you have a home directory which isn’t /root:
5. .pylintrc in your home directory
6. .config/pylintrc in your home directory
7. /etc/pylintrc
```

We might need an extra check for the `PYLINTRC` env variable case.

ccordoba12:
Thanks @goanpeca! I'll try to finish this one before 4.0 final.

goanpeca:
Merging with master @ccordoba12 

Thoughts on this now?
Screen Shot 2019-11-25 at 19 21 24

ccordoba12:
Looks good to me. The only thing missing here is handling the presence of a `pylintrc` file and adding one of our own in case there's none present in the user system.

goanpeca:
>The only thing missing here is handling the presence of a pylintrc file and adding one of our own in case there's none present in the user system.

Yep, working on this.

goanpeca:
@ccordoba12 POC

Screen Shot 2019-11-25 at 21 32 24

spyder/plugins/completion/languageserver/confpage.py
```diff
@@ -10,34 +10,70 @@
...
-from spyder.utils.programs import find_program
+from spyder.utils.programs import find_program, run_program
 from spyder.widgets.helperwidgets import ItemDelegate
 
 LSP_LANGUAGE_NAME = {x.lower(): x for x in LSP_LANGUAGES}
 LSP_URL = "https://microsoft.github.io/language-server-protocol"
 
 
+
+def get_pylint_default_path():
```
goanpeca: Maybe I should move all these to a separate `.../utils/pylint.py` file?

ccordoba12:
We decided not to implement this at the end because the Code Analysis pane provides the same functionality.

PR #10169, opened 2019-09-10
PR: Allow selecting the priority of the completion clients to use

### LINKED ISSUE #10071: Kite is always started
I am not sure if this is a bug or a feature. I have not been using `kite` that much since it consumes too much resources, so I usually shut down the client on OSX, however spyder is always starting the client and I do not see any option to prevent this.

Thoughts @andfoy, @ccordoba12 ?

goanpeca:
## Description of Changes

image

### Issue(s) Resolved

Fixes #10071.

### Affirmation


spyder/config/main.py
```diff
@@ -496,6 +497,14 @@
               'advanced/external': False,
               'advanced/stdio': False
              }),
+             ('completion-fallback',
```
ccordoba12: Let's change this to

    fallback-completions

spyder/config/main.py
```diff
@@ -496,6 +497,14 @@
               'advanced/external': False,
               'advanced/stdio': False
              }),
+             ('completion-fallback',
+             {
+              'enable': True,
+              }),
+             ('completion-kite',
```
ccordoba12: Let's change this to

    kite-completions

spyder/plugins/completion/languageserver/confpage.py
```diff
@@ -1029,6 +1044,7 @@ def setup_page(self):
 
         # --- Tabs organization ---
         self.tabs = QTabWidget()
+        self.tabs.addTab(self.create_tab(clients_widget), _('Clients'))
```
ccordoba12: Instead of adding a new tab just with these three options, perhaps we should add the new checkboxes to the "Advanced" one? I mean, we could create two QGroupBoxes: one for the advanced options of the PyLS and the other to select among the available completion clients.

If you think that's not a good suggestion, then we should put this tab to the left ot the "Advanced" tab.

@andfoy, any input on your side about this?

ccordoba12:
@dalthviz, please take note of this PR so you can mention it in this Friday's meeting with the Kite people. This will allow users to easily turn on/off Kite completions through our interface.

spyder/plugins/completion/languageserver/confpage.py
goanpeca: >Instead of adding a new tab just with these three options, perhaps we should add the new checkboxes to the "Advanced" one? 

Sure

goanpeca:
@ccordoba12 

image

Thoughts?

goanpeca:
@andfoy if these options are changed how should we handle so the servers can restart (fallback etc...?)?

goanpeca:
@ccordoba12  should I hide the option for kite completions if kite is not installed?

andfoy:
> if these options are changed how should we handle so the servers can restart (fallback etc...?)?

We should handle configuration changes by using the update_configuration method of the main completion plugin, this will be merged today as part of the snippets

ccordoba12:
> should I hide the option for kite completions if kite is not installed?

I think the combobox should be disabled but not hidden because Kite can be installed while Spyder is running.

goanpeca:
>I think the combobox should be disabled but not hidden because Kite can be installed while Spyder is running.

Agreed, but if the user closes and opens preferences, it would then be enabled (or appear) unless we want this to be reflected life (with some timer) while the preferences are open.

goanpeca:
>We should handle configuration changes by using the update_configuration method of the main completion plugin, this will be merged today as part of the snippets

Ok @andfoy thanks! @ccordoba12 should we wait for that PR so I can update the configuration to enable/disable clients while spyder runs (instead of on restart ?)

ccordoba12:
> We should handle configuration changes by using the update_configuration method of the main completion plugin, this will be merged today as part of the snippets

This is a good point. @goanpeca, your work enables/disables completion plugins at startup but not while Spyder is running.

So I think we should wait until @andfoy's PR is merged to be able to use that method here.

ccordoba12:
> Agreed, but if the user closes and opens preferences, it would then be enabled (or appear) unless we want this to be reflected life (with some timer) while the preferences are open.

If Kite is installed while Spyder is running, we need to show a dialog saying to the user that Kite is available now, that they can start using it if they want to and if not, they can go to Preferences to activate it.

goanpeca:
>If Kite is installed while Spyder is running, we need to show a dialog saying to the user that Kite is available now, that they can start using it if they want to and if not, they can go to Preferences to activate it.

Got it! Should I add this as part of this PR ?

ccordoba12:
By the way, I moved this PR to the release candidate because it's not a priority for beta5.

ccordoba12:
@goanpeca, fix our tests here so we can merge this one.

goanpeca:
@ccordoba12 ?

goanpeca:
@ccordoba12 let me know if this needs to be changed

ccordoba12:
We discussed about this PR with @andfoy and agreed that before merging it we need to provide a way for Kite to take over the LSP when Kite is selected, instead of simply turning the LSP off completely. 

That's because the LSP provides things that Kite doesn't, like linting, workspace and soon folding too.

ccordoba12:
@andfoy is going to work on that PR first and you'll need to update this one afterwards.

metalogical:
Subsumed and expanded on in #10354. This can be closed.

PR #10170, opened 2019-09-10
PR: Add class variables to set configuration options for plugins

goanpeca:
## Description of Changes

The plugin API now includes CONF_FILE, CONF_DEFAULTS, CONF_NAME_MAP

```python
class Explorer(SpyderPluginWidget):
    """File and Directories Explorer DockWidget."""

    CONF_SECTION = 'explorer'
    CONFIGWIDGET_CLASS = ExplorerConfigPage
    CONF_FILE = True
    CONF_DEFAULTS = [            ('explorer',
             {
              'enable': True,
              'wrap': True,
              'name_filters': [],
              'show_hidden': True,
              'show_all': True,
              'show_icontext': False,
              'single_click_to_open': False,
              }),]
```
This enables the use of a dedicated configuration file per plugin.

This is by default `True` so external plugins default to use their own configuration file.

The location of storage is (for the example):
image

### Issue(s) Resolved

Fixes #

### Affirmation

### DIFF
--- a/spyder/api/plugins.py
+++ b/spyder/api/plugins.py
@@ -31,6 +31,10 @@ class BasePlugin(BasePluginMixin):
 
     WARNING: Don't override any methods or attributes present here!
     """
+    # Define configuration name map for plugin to split configuration
+    # among several files. See spyder/config/main.py
+    # Status: Hidden
+    _CONF_NAME_MAP = None
 
     # Use this signal to display a message in the status bar.
     # str: The message you want to display
@@ -157,17 +161,29 @@ class SpyderPlugin(BasePlugin):
     # Status: Optional
     CONF_FILE = True
 
-    # Define configuration name map for plugin to split configuration
-    # among several files
+    # Define configuration defaults if using a separate file.
+    # List of tuples, with the first item in the tuple being the section
+    # name and the second item being the default options dictionary.
     # Status: Optional
-    CONF_NAME_MAP = None
-
-    # Define configuration defaults if using a separate file
-    # Status: Optional
+    #
+    # CONF_DEFAULTS_EXAMPLE = [
+    #     ('section-name', {'option-1': 'some-value',
+    #                       'option-2': True,}),
+    #     ('another-section-name', {'option-3': 'some-other-value',
+    #                               'option-4': [1, 2, 3],}),
+    # ]
     CONF_DEFAULTS = None
 
     # Define configuration version if using a separate file
     # Status: Optional
+    #
+    # IMPORTANT NOTES:
+    # 1. If you want to *change* the default value of a current option, you
+    #    need to do a MINOR update in config version, e.g. from 3.0.0 to 3.1.0
+    # 2. If you want to *remove* options that are no longer needed or if you
+    #    want to *rename* options, then you need to do a MAJOR update in
+    #    version, e.g. from 3.0.0 to 4.0.0
+    # 3. You don't need to touch this value if you're just adding a new option
     CONF_VERSION = None
 
     # ------------------------------ METHODS ----------------------------------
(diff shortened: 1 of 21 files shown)


spyder/api/plugins.py
```diff
@@ -149,11 +149,27 @@ class SpyderPlugin(BasePlugin):
...
+    # Widget to be used as entry in Spyder Preferences dialog
     # Status: Optional
     CONFIGWIDGET_CLASS = None
 
+    # Use separate configuration file for plugin
+    # Status: Optional
+    CONF_FILE = True
+
+    # Define configuration name map for plugin to split configuration
+    # among several files
```
ccordoba12: Please expand this comment. I don't understand how to use this constant.

spyder/api/plugins.py
```diff
@@ -149,11 +149,27 @@ class SpyderPlugin(BasePlugin):
...
+    # Status: Optional
+    CONF_NAME_MAP = None
+
+    # Define configuration defaults if using a separate file
+    # Status: Optional
```
ccordoba12: This also needs to be expanded with an example.

spyder/api/plugins.py
```diff
@@ -149,11 +149,27 @@ class SpyderPlugin(BasePlugin):
...
+    CONF_DEFAULTS = None
+
+    # Define configuration version if using a separate file
+    # Status: Optional
```
ccordoba12: I think we should explain here how CONF_VERSION works, with a similar comment to the one present in `config/main.py`.

spyder/config/manager.py
```diff
@@ -52,6 +56,33 @@ def __init__(self, parent=None, active_project_callback=None):
...
+        if plugin_class.CONF_FILE and conf_section:
+            path = self.get_plugin_config_path(conf_section)
+            version = plugin_class.CONF_VERSION
+            version = version if version else '0.0.0'
+            name_map = plugin_class.CONF_NAME_MAP
+            name_map = name_map if name_map else {'spyder': []}
+            defaults = plugin_class.CONF_DEFAULTS
+
+            if conf_section in self._plugin_configs:
+                raise Exception('A plugin with section "{}" already exsists!'
```
ccordoba12: exsists -> exists

ccordoba12: Also, perhaps it'd be better to use `RuntimeError` here instead of `Exception`, which is too broad.

spyder/config/manager.py
```diff
@@ -78,6 +114,18 @@ def get_user_config_path(self):
 
         return path
 
+    def get_plugin_config_path(self, plugin_folder):
+        """Return the plugin configuration path."""
+        base_path = get_conf_path()
+        path = osp.join(base_path, 'plugins')
+        if plugin_folder is None:
+            raise Exception('Plugin configuration needs to define a `CONF_SECTION`!')
```
ccordoba12: Here we should also use `RuntimeError`.

spyder/api/plugins.py
```diff
@@ -31,6 +31,10 @@ class BasePlugin(BasePluginMixin):
 
     WARNING: Don't override any methods or attributes present here!
     """
+    # Define configuration name map for plugin to split configuration
+    # among several files. See spyder/config/main.py
+    # Status: Hidden
+    _CONF_NAME_MAP = None
```
goanpeca: @ccordoba12 since the CONF_MAP option is advanced, I will hide it for now from the API.

spyder/api/plugins.py
goanpeca: added example

spyder/api/plugins.py
goanpeca: added notes

spyder/api/plugins.py
ccordoba12: Ok, no problem. Then you should move it to `BasePluginMixin` until we decide expose it here.

goanpeca: 👍

PR #10220, opened 2019-09-16
PR: Catch errors when sending messages to tcp sockets that reject a connection from our LSP client

ccordoba12:
This avoids a total freeze at startup when there's a TCP socket listening at `host:port` (as defined in our preferences), which rejects our connection.

I also tested things with a simple ping/pong server and (fortunately) in that case Spyder doesn't freeze, but this fix is not really needed.

### DIFF
--- a/spyder/plugins/completion/languageserver/transport/tcp/producer.py
+++ b/spyder/plugins/completion/languageserver/transport/tcp/producer.py
@@ -72,8 +72,7 @@ class TCPLanguageServerClient(LanguageServerClient):
         except (BrokenPipeError, ConnectionError) as e:
             # This avoids a total freeze at startup
             # when we're trying to connect to a TCP
-            # (but not LSP) server that rejects our
-            # connection
+            # socket that rejects our connection
             logger.error(e)
 
     def is_server_alive(self):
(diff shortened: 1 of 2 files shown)


spyder/plugins/completion/languageserver/transport/tcp/producer.py
```diff
@@ -70,6 +70,10 @@ def transport_send(self, content_length, body):
             self.socket.send(content_length)
             self.socket.send(body)
         except (BrokenPipeError, ConnectionError) as e:
+            # This avoids a total freeze at startup
+            # when we're trying to connect to a TCP
+            # (but not LSP) server that rejects our
```
goanpeca: `(but not LSP). I don't understand that phrase @ccordoba12. What do you mean?

ccordoba12: It means any TCP server (for example an http server). The problem is we don't have a way to tell if a server is LSP because the LSP protocol doesn't provide the means for that.

andfoy: We could paraphrase it to refer to a TCP socket instead of a server

ccordoba12: Ok, will do.

PR #10249, opened 2019-09-18
PR: Avoid stripping if the event comes from another editor

impact27:
## Description of Changes
When the editor is split, a change in one editor should not trigger a strip in the other editor. Here, the focus is checked before stripping.

### Issue(s) Resolved

Fixes #

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -3811,6 +3811,9 @@ class CodeEditor(TextEditBaseWidget):
         Remove trailing whitespace on leaving a non-string line containing it.
         Return the number of removed spaces.
         """
+        if not self.hasFocus():
+            # Avoid problem when using split editor
+            return 0
         # Update current position
         current_position = self.textCursor().position()
         last_position = self.last_position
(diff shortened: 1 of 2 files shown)


ccordoba12:
The failure in our tests look legit although I don't understand why they are failing only for conda and not for pip.

impact27:
This doesn't make sense to me:
```
1741        widget.setFocus()
1742>       assert widget.hasFocus()
1743E       assert False
```
How can `widget.hasFocus()` not work properly?

ccordoba12:
Focus doesn't work properly on Travis because it uses a frame buffer instead of a proper Xorg server.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3811,6 +3811,9 @@ def strip_trailing_spaces(self):
         Remove trailing whitespace on leaving a non-string line containing it.
         Return the number of removed spaces.
         """
+        if not self.hasFocus():
```
ccordoba12: Please enclose this in another `if`, like this:

to avoid the failure in our tests.

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -154,6 +154,8 @@ def test_editor_rstrip_keypress(
     cursor = widget.textCursor()
     cursor.movePosition(QTextCursor.End)
     widget.setTextCursor(cursor)
+    widget.setFocus()
+    assert widget.hasFocus()
```
ccordoba12: These two lines are not necessary anymore.

PR #10250, opened 2019-09-18
PR: Use only tab to select completions

### LINKED ISSUE #10134: On the fly code completion interferes with normal typing

This may be intended behavior, however, it didn't happen with 3.X. 

With code completion and on the fly code completion turned on (note these are on different panes in preferences, which makes discovery hard), try to type the following function definition in a new editor window.

    def f(x):
        return x

The code completion pops up a window with xrange as the completion of the `return x`. If you are typing quickly (as I do) it will auto-complete x as xrange, yielding

    def f(x):
        return xrange

I think 3.x had a way to control whether ENTER auto completes or not, I don't think I see that in 4 (this issue reporter window is modal, so I can't check while I am typing this)

Anyway, this may be the intended/desired behavior, but I find it hard to imagine most will find it useful. One may counter that x is a bad variable name, but I only really do math in Python, so x is very common usage. 

## Versions

* Spyder version: 4.0.0b4 
* Python version: 3.7.3
* Qt version: 5.9.6
* PyQt5 version: 5.9.2
* Operating System: Windows 7

### Dependencies

```
qdarkstyle >=2.6.4:  2.7 (OK)
sympy >=0.7.3     :  1.4 (OK)
```

goanpeca:
## Description of Changes

### Issue(s) Resolved

Fixes #10134

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/completion.py
+++ b/spyder/plugins/editor/widgets/completion.py
@@ -277,9 +277,10 @@ class CompletionWidget(QListWidget):
         QListWidget.hide(self)
         QToolTip.hideText()
 
-
     def _select_completion(self, event):
         """
+        Use selected completion from list.
+
         Check that what was selected can be selected otherwise timing issues.
         """
         if self.is_up_to_date():
--- a/spyder/plugins/editor/widgets/completion.py
+++ b/spyder/plugins/editor/widgets/completion.py
@@ -10,7 +10,7 @@
 import sys
 
 # Third psrty imports
-from qtpy.QtCore import QPoint, Qt, Signal, Slot
+from qtpy.QtCore import QEvent, QPoint, Qt, Signal, Slot
 from qtpy.QtGui import QFontMetrics
 from qtpy.QtWidgets import (QAbstractItemView, QApplication, QListWidget,
                             QListWidgetItem, QToolTip)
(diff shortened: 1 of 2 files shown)


ccordoba12:
@goanpeca, I think you need to fix our completion tests to use Tab instead of Enter.

spyder/plugins/editor/widgets/completion.py
```diff
@@ -22,7 +22,7 @@
 from spyder.widgets.helperwidgets import HTMLDelegate
 
 
-COMPLETION_ITEM_TEMPLATE = """
+COMPLETION_ITEM_TEMPLATE = u"""
```
goanpeca: Either we add this `u` or `unicode_literals` import on top.

Any preference @ccordoba12 ? (this was failing for me on python 2)

spyder/plugins/editor/widgets/completion.py
```diff
@@ -277,14 +277,30 @@ def hide(self):
         QListWidget.hide(self)
         QToolTip.hideText()
 
+    def event(self, event):
+        """Override Qt method to process tab keypress."""
```
goanpeca: The `Tab` key is not actually propagated to he keypress so I had to filter it here

goanpeca:
Not sure what the failure on Mac is @ccordoba12

CAM-Gerlach:
@goanpeca Could you considering making this a configurable option, like it was in Spyder 3 (which had return as default) and a number of other IDEs? Otherwise, it will be a regression to hardcode it, whereas (particularly if the other suggestions in #10134 are implemented) return for completion offers substantial ergonomic, speed and consistency benefits over tab for many users and is a strong per-user personal preference. Thanks.

spyder/plugins/editor/widgets/completion.py
ccordoba12: Using `u` is fine. `unicode_literals` can introduce undesired side effects, that's why we don't use it.

ccordoba12:
Let's follow the approach of PR #10262 instead.

PR #10260, opened 2019-09-19
PR: Show Kite onboarding file once after Kite was installed

jansorg:
## Description of Changes
Kite onboarding means that a Python-file is opened at startup to show what Kite is able to provide.
The file is only shown once (controlled by a setting).

This PR updates the completion extension to open up the file after the main window was displayed and when the kite client is ready.

To do / open questions:

* [x] I couldn't find a way to run code after all startup actions were run. Currently the last opened files are restored after `main.restore_scrollbar_position` was emitted. This restore takes the focus away from the current file. Therefore I'm using a delay via `QTimer.singleShot` here, but this is more of a hack. <br>*Is there a hook to run code at the end of the startup?* 
* [x] Is accessing `self.main` and opening a new editor in a completion plugin okay or should this be avoided? An editor plugin could do the same, but I didn't know how to handle the dependency to the Kite client in the completion plugin in this case.
* [x] ~~Kite's spec states that a notification should be displayed if the retrieval of the onboarding file from Kite failed. Is there existing functionality in the Spyder SDK to display a notification with links/actions?~~
* [ ] Disable this if the `Enable Kite` setting is off. The setting is not yet on master, afaik.

### Issue(s) Resolved

Fixes #

### Affirmation

I certify the above statement is true and correct: jansorg

kite-onboarding


dalthviz:
Hi @jansorg , some feedback to your questions:

>I couldn't find a way to run code after all startup actions were run. Currently the last opened files are restored after main.restore_scrollbar_position was emitted. 

Maybe we can add a signal for this propose. I think we can add it at the end of the method `post_visible_setup` (the same method that emits the `restore_scrollbar_position`) of the MainWindow (`mainwindow.py`).

> Is accessing self.main and opening a new editor in a completion plugin okay or should this be avoided?

You can call the MainWindow (`mainwindow.py`) from the Kite plugin, although I think there is a method on the MainWindow called `open_file` which you can use instead of calling directly the editor through the MainWindow

> Kite's spec states that a notification should be displayed if the retrieval of the onboarding file from Kite failed. Is there existing functionality in the Spyder SDK to display a notification with links/actions?

In case of error you could display a `QMessageBox` dialog. You can find some examples of this dialog at the `mainwindow.py` and some more info about this widget from the docs of Qt [here](https://doc.qt.io/qt-5/qmessagebox.html).

spyder/plugins/completion/kite/client.py
```diff
@@ -68,6 +68,13 @@ def get_languages(self):
             response = ['python']
         return response
 
+    def get_onboarding_file(self):
+        verb, url = KITE_ENDPOINTS.ONBOARDING_ENDPOINT
+        success, response = self.perform_http_request(verb, url)
+        if response is None:
```
dalthviz: I think that maybe this `if` isn't needed

spyder/plugins/completion/kite/plugin.py
```diff
@@ -34,16 +34,32 @@ def __init__(self, parent):
...
         self.client = KiteClient(None, enable_code_snippets)
         self.kite_process = None
+        self.main_window_visible = False
+        self.kite_initialized = False
+        self.onboarding_shown = False
         self.client.sig_client_started.connect(self.http_client_ready)
         self.client.sig_response_ready.connect(
             functools.partial(self.sig_response_ready.emit,
                               self.COMPLETION_CLIENT_NAME))
+        self.main.sig_setup_finished.connect(self.mainwindow_setup_finished)
```
dalthviz: Maybe we could connect the new `sig_setup_finished` signal directly with the `_kite_onboarding` method (and remove then the `main_window_visible` attribute and the `main_window_setup_finished` method)

spyder/plugins/completion/kite/plugin.py
```diff
@@ -34,16 +34,32 @@ def __init__(self, parent):
         enable_code_snippets = CONF.get('lsp-server', 'code_snippets')
         self.client = KiteClient(None, enable_code_snippets)
         self.kite_process = None
+        self.main_window_visible = False
+        self.kite_initialized = False
+        self.onboarding_shown = False
```
dalthviz: Is this attribute needed? or maybe it was a placeholder for the value that we should get from the preferences?

spyder/plugins/completion/kite/plugin.py
```diff
@@ -34,16 +34,32 @@ def __init__(self, parent):
...
                               self.COMPLETION_CLIENT_NAME))
+        self.main.sig_setup_finished.connect(self.mainwindow_setup_finished)
 
     @Slot(list)
     def http_client_ready(self, languages):
         logger.debug('Kite client is available for {0}'.format(languages))
         self.available_languages = languages
         self.sig_plugin_ready.emit(self.COMPLETION_CLIENT_NAME)
+        self.kite_initialized = True
+        self._kite_onboarding()
```
dalthviz: Is not enough with the signal from the `MainWindow` or maybe sometimes the client isn't ready even when the MainWindow already ended up its setup?

spyder/plugins/completion/kite/client.py
jansorg: Definitely not needed!

spyder/plugins/completion/kite/plugin.py
jansorg: sorry, I should have removed this. 
The setting is called `kite_show_onboarding`, is that okay or should I move it to another section or config file?

spyder/plugins/completion/kite/plugin.py
jansorg: My thinking is that `self.client.start()` is starting the client async and will trigger `http_client_ready` at some point. `sig_setup_finished` is triggered independently from this. So both may happen in any order and this shouldn't trigger onboarding just once (not 0 or 2 times).
This might be too cautious if there's some synchronization happening which I didn't notice.
What do you think?

jansorg:
@dalthviz Thanks for the feedback! I've responded and fixed the obvious issues.

spyder/plugins/completion/kite/plugin.py
dalthviz: Ok, makes sences, thanks for the explanation :+1:

spyder/plugins/completion/kite/plugin.py
dalthviz: I think is ok :+1:

jansorg:
I fixed a bug which I accidentally added with my last fix to this PR.
This is ready for Spyder. It's probably best to wait until kite's public release is ready too. It needs to bundle the Spyder onboarding file data, this will most likely happen with today's releas.

jansorg:
The latest public release of Kite now includes the onboarding file data which is needed for this PR.

ccordoba12:
> Disable this if the Enable Kite setting is off. The setting is not yet on master, afaik.

That will be added in PR #10301, so let's wait until that one is merged so that you can do this afterwards.

metalogical:
@ccordoba12 @dalthviz 

Please take a look at the updated logic here. I now attempt to request & show the Kite onboarding file in two cases:
1. When Spyder starts, after the editor is fully initialized.
2. When we receive a valid response from the Kite daemon (connected to the `sig_response_ready`, etc signals).
    - this is necessary if the user installs & starts Kite in the middle of a Spyder session: either via the in-Spyder install flow, or manually by downloading it from Kite.com
    - also, this cleanly handles the race condition where Spyder finishes initializing before the Kite daemon is available.

After the onboarding file is shown once, we never show it again. Let me know what you think.

dalthviz:
@naman @jansorg @ccordoba12 seems like Spyder could freeze if Kite isn't installed or is unreachable with this PR. I will push some changes to tackle such behavior (basically by using signals instead of directly calling the client for the onboarding file). After that this should be ready :+1:

metalogical:
@dalthviz thanks!

I think it wouldn't freeze if Kite isn't installed/running (since the `ConnectionError` happens fast), but there may be other edge cases in which it might (if the Kite daemon had a bug and didn't return the onboarding file quickly), so it makes sense to be extra safe.

Thanks for handling the fix -- let me know if you need anything from my end.

ccordoba12:
> I will push some changes to tackle such behavior

Thanks a lot @dalthviz for helping @metalogical to finish this one!

dalthviz:
Failing tests are unrelated to the changes on this PR

PR #10253, opened 2019-09-18
PR: Move some options from editor to LSP Preferences page

goanpeca:
## Description of Changes

* [x] Wrote at least one-line docstrings (for any new functions)
* [x] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))
comple

### Issue(s) Resolved

Fixes #

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct: @goanpeca

PR #10262, opened 2019-09-19
PR: Add delay and minimum chars to on the fly completions

### LINKED ISSUE #10134: On the fly code completion interferes with normal typing

goanpeca:
## Description of Changes

- Add configurable time delay to on the fly completion
- Add configurable characters after which on the fly completion is triggered
- Remove on the fly completion for strings and/or comments (normal `tab` completion will still work)
- Split configuration options and add a new tab 

image

image

image

### Issue(s) Resolved

Fixes #10134

### Affirmation


goanpeca:
Screen Shot 2019-09-19 at 23 39 31
Screen Shot 2019-09-19 at 23 39 38
Screen Shot 2019-09-19 at 23 39 46

@ccordoba12

goanpeca:
@CAM-Gerlach could you try this one?

CAM-Gerlach:
@goanpeca I tested it out and the character, time delay, string and comment no-completion, and settings all worked very well, and the preferences looked great, and I didn't get any errors during limited testing. There may be a few corner cases (I don't have much time at the moment to test with my usual level of exhaustiveness) but overall its a huge improvement from the previous state of affairs.

There's one thing, though—since you already have the option implemented, can you allow a minimum delay of 0 ms rather than 200 ms, for users who prefer maximum responsiveness in their OTF autocompletion? As far as I'm aware there's no technical reason to restrict it (given it was 0 before, and its already configurable so it doesn't increase complexity any further) and some users may not want any delay or very little (I had it set to 100 ms in Rstudio) though 300 ms is more than fine as the default.

In any case, I'll certainly be making great use of OTF autocompletion thanks to this patch as well as all the previous work done on it. Aside from the handful of remaining errors I regularly run into, as a user I can say I'm very pleased with how things have turned out in Spyder 4 in this department, and you all should be very proud of your work. Thank you!

spyder/plugins/completion/languageserver/confpage.py
```diff
@@ -692,17 +692,59 @@ class LanguageServerConfigPage(GeneralConfigPage):
     def setup_page(self):
         newcb = self.create_checkbox
 
-        # --- Introspection ---
-        # Basic features group
-        basic_features_group = QGroupBox(_("Basic features"))
-        completion_box = newcb(_("Enable code completion"), 'code_completion')
+        # --- Completion ---
+        # Completion group
+        completion_label = QLabel(_('Some text for this section!'))
```
ccordoba12: There's no need for a label in this section, so please remove this one.

spyder/plugins/completion/languageserver/confpage.py
```diff
@@ -692,17 +692,59 @@ class LanguageServerConfigPage(GeneralConfigPage):
...
+        self.completions_after_characters = self.create_spinbox(
+            _("Show automatic completions after characters:"), None,
+            'automatic_completions_after_chars', min_=1, step=1,
+            tip=_("Default is 3"), section='editor')
+        self.completions_after_ms = self.create_spinbox(
+            _("Show automatic completions after keyboard idle (ms):"), None,
+            'automatic_completions_after_ms', min_=200, max_=5000, step=10,
+            tip=_("Default is 300"), section='editor')
         code_snippets_box = newcb(_("Enable code snippets"), 'code_snippets')
         enable_hover_hints_box = newcb(
```
ccordoba12: Please move this checkbox to the `Introspection` tab because it's not related to code completion.

spyder/plugins/completion/languageserver/confpage.py
```diff
@@ -692,17 +692,59 @@ class LanguageServerConfigPage(GeneralConfigPage):
...
+        self.completion_hint_box = newcb(
+            _("Show completion details"),
+            'completions_hint',
+            section='editor')
+        self.automatic_completion_box = newcb(
+            _("Show completions on the fly"),
+            'automatic_completions',
+            section='editor')
+        self.completions_after_characters = self.create_spinbox(
+            _("Show automatic completions after characters:"), None,
```
ccordoba12: Let's follow RStudio here and change this text to be

    Show automatic completions after characters entered:

spyder/plugins/completion/languageserver/confpage.py
```diff
@@ -1053,20 +1099,34 @@ def setup_page(self):
 
         # --- Tabs organization ---
         self.tabs = QTabWidget()
-        self.tabs.addTab(self.create_tab(basic_features_group, advanced_group),
-                    _('Introspection'))
+        self.tabs.addTab(self.create_tab(completion_widget),
+                         _('Completion'))
+        self.tabs.addTab(self.create_tab(introspection_group, advanced_group),
+                         _('Introspection'))
```
ccordoba12: Please move this tab to be after the `Linting` one (as I mentioned before) because it's very strange that the entry in Preferences is called "Completion and linting" but not seeing those two tabs one after the other here.

CAM-Gerlach: The rationale here is likely that which I outlined on #10134 : given code analysis ("linting"), code style analysis and docstring style analysis are closely connected at least from a user's perspective, it makes sense to keep them together.

If you insist on linting second after completion (which from the perspective of a user I really don't see as mattering all that much, at least compared to grouping the three real-time code analysis features together), Introspection could be moved after docstring style, or you could consider renaming the preferences section to a broader term more inclusive of its contents (though I'm not sure what, exactly).

goanpeca:
@CAM-Gerlach thanks for testing this out!

>There's one thing, though—since you already have the option implemented, can you allow a minimum delay of 0 ms rather than 200 ms, for users who prefer maximum responsiveness in their OTF autocompletion? As far as I'm aware there's no technical reason to restrict it (given it was 0 before, and its already configurable so it doesn't increase complexity any further) and some users may not want any delay or very little (I had it set to 100 ms in Rstudio) though 300 ms is more than fine as the default.

Sure sounds fine.

CAM-Gerlach:
Thanks @goanpeca !

spyder/plugins/completion/languageserver/confpage.py
goanpeca: I think both are valid points!

goanpeca:
>This looks pretty good, thanks @goanpeca! I left you some minor comments and I guess you also need to do a rebase on top of master because I already merged PR #10253.

Fixed @ccordoba12!

PR #10300, opened 2019-09-24
PR: Add branding for Kite completion items

dalthviz:
## Description of Changes

b

### Issue(s) Resolved

Fixes #

### Affirmation

### DIFF
--- a/spyder/plugins/completion/kite/providers/document.py
+++ b/spyder/plugins/completion/kite/providers/document.py
@@ -228,6 +228,7 @@ class DocumentProvider:
                 params = {
                     'signatures': base_signature,
                     'activeSignature': 0,
-                    'activeParameter': arg_idx
+                    'activeParameter': arg_idx,
+                    'provider': KITE_COMPLETION
                 }
         return {'params': params}
--- a/spyder/plugins/completion/kite/providers/document.py
+++ b/spyder/plugins/completion/kite/providers/document.py
@@ -30,6 +30,8 @@ KITE_DOCUMENT_TYPES = {
     'unknown': CompletionItemKind.TEXT
 }
 
+KITE_COMPLETION = 'Kite'
+
 logger = logging.getLogger(__name__)
 
 
@@ -122,7 +124,8 @@ class DocumentProvider:
                     'insertText': completion['snippet']['text'],
                     'filterText': completion['display'],
                     'sortText': completion['display'],
-                    'documentation': completion['documentation']['text']
+                    'documentation': completion['documentation']['text'],
+                    'provider': KITE_COMPLETION
                 }
                 spyder_completions.append(entry)
                 if 'children' in completion:
--- a/spyder/plugins/editor/widgets/completion.py
+++ b/spyder/plugins/editor/widgets/completion.py
@@ -33,6 +33,10 @@ COMPLETION_ITEM_TEMPLATE = u"""
         <td valign="middle" align="right" float="right" style="color:{color}">
             {type}
         </td>
+        <td valign="top" align="right" float="right" width="{img_width}">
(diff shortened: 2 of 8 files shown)


spyder/plugins/completion/kite/providers/document.py
```diff
@@ -30,6 +30,8 @@
     'unknown': CompletionItemKind.TEXT
 }
 
+KITE_COMPLETION = 'Kite'
```
its-dhung: I would change this to:
```python
KITE_COMPLETION = '[kite]'
```

ccordoba12:
@dalthviz, I don't like this approach too much. I'd prefer to use Kite's icon instead of our current icons to signal that a completion is coming from Kite.

The problem is this approach is a bit misleading because a regular class (like `ImportError`, in your screenshot above) seems to be become a special class (called `Kite class`) when the user select the Kite entry for it.

Thoughts @its-dhung, @goanpeca?

goanpeca:
is there a reason why the message of the type of completion kite or not appears only when selecting it, instead o seeing, for example, the icon all the time?

Also we could considering showing this icon to the right of the box (instead of before the `class` text)

dalthviz:
@goanpeca I think the initial idea was to show the provider of the completion only for the current selected item, but maybe is better to show the provider (probably the icon) always?

goanpeca:
>but maybe is better to show the provider (probably the icon) always?

It would make it clear, besides we could explain on the initial part, that the completion widgets obtined by kite will display the icon (so users have a clear expectation)

metalogical:
I'm generally in agreement with all the comments here. The approximate behavior we use for other editors is:
- if we're showing the icon, we show it for all completions (we overlay it on the left, but the right is fine as well).
- if we're showing the `[kite]` text, then we show it only for the selected completion to reduce visual noise.

its-dhung:
To reitreate what @metalogical said, if we use Kite's icon, I would place it on the left. However, if we have to place it on the right, I would put it _after_ the type of completion e.g. `class [icon]`

dalthviz:
@its-dhung is the new preview what you spect (I updated the preview gif in the initial comment of the PR)?

its-dhung:
@dalthviz yes that looks great!

spyder/plugins/editor/widgets/completion.py
```diff
@@ -253,14 +257,42 @@ def update_list(self, filter_text):
...
+        item_provider = item_info['provider']
+        item_type = self.ICONS_MAP.get(item_info['kind'], 'no_match')
+        item_label = item_info['label']
+        icon_provider = ("iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0l"
+                         "EQVR42mNkYAAAAAYAAjCB0C8AAAAASUVORK5CYII=")
+        img_height = height - 2
+        img_width = img_height * 0.8
+
+        if item_provider == KITE_COMPLETION:
+            icon_provider = ima.get_kite_icon()
```
its-dhung: The icon appears to be blurry due to the resizing being done by the GUI.

image

You can fix this by resizing the image here:

```python
kite_height = img_height
kite_width = (416.14/526.8) * kite_height
icon_provider = ima.get_kite_icon()
icon_provider = ima.base64_from_icon_obj(
    icon_provider, kite_width, kite_height)
```

This is what it looks like when manually resized:

image

goanpeca:
@dalthviz is this ready?

dalthviz:
Yep @goanpeca I think so 👍

goanpeca:
The windows error on azure is not related to the PR. Merging

sergeyFlakon:
I see branding displayed correctly on Commit: (13c208033)

PR #10301, opened 2019-09-25
PR: Fix completions prioritization and Kite completions behavior

### LINKED ISSUE #10226: Kite (and LSP) completions frequently not displayed
## Problem Description

Kite (and LSP) completions frequently don't show up:

1. start Spyder in debug mode with Kite installed & running
2. type the following
```python
import numpy as np

np.ar‸
```
3. backspace and type `r` repeatedly

- sometimes no completions show up
- sometimes `ArithmeticError` shows up (presumably from LSP?)
- sometimes Kite completions show up
- in all cases, you can see the Kite completions request returns very quickly, and the response is shown in the debugging output

I expect Kite completions (and LSP?) to always show up. As a separate issue, the completions ordering returned by Kite is not respected: I will create a separate issue to track this.

## Versions

* Spyder version: Spyder 4.0.0.dev0 (Commit: 11611776d) 
* Python version: 3.7
* Operating System name/version: macOS

### LINKED ISSUE #10227: Kite completions ordering is not respected
## Problem Description

Kite completions ordering is not respected

1. start Spyder in debug mode with Kite installed & running
2. type the following
```python
import numpy as np

np.ar‸
```
3. backspace and type `r` repeatedly until Kite completions show up (see #10226 )

Screen Shot 2019-09-16 at 4 36 08 PM

These completions appear to be sorted alphabetically, but the Kite response shown in the debug logs has `array(...)` as the first completion, which is much better, since `array` is much more likely to be called based on usage frequency.

metalogical:
## Description of Changes 

1. add "enable Kite", "enable fallback" booleans to the config file
2. if Kite is enabled, prioritize Kite requests over LSP for completions/signatures/hover (and if the Kite request fails, *then* trigger LSP)
3. if Kite is enabled, make Kite requests regardless of whether Kite is detected as installed.
    * this allows developer builds, test builds, etc.
4. fix completions ordering to show Kite -> LSP -> fallback, and for Kite completions, use the order returned by Kite instead of alphabetical.
5. fix small Kite completions display text bug that I discovered while making the original fix.
6. fix completions "hints" mapping
7. make the completions chars inserted and ms time-to-completions behavior actually respected, and update to 1 char/300ms, since the time is the only thing that really affects perceived editor lag.

Performance does *not* appear to be impacted:
- If Kite is enabled but unavailable, LSP completions still show as fast as before.
- Thus we enable Kite by default, under the assumption that if the user has Kite running, they want Kite completions. If not, they'll get LSP completions anyways.

Please let me know if you see stylistic issues, or have any suggestions on the appropriate way to test these changes.

Screen Shot 2019-09-24 at 5 49 08 PM

### Issue(s) Resolved

Fixes #10226 
Fixes #10227 

### Affirmation


spyder/plugins/completion/plugin.py
```diff
@@ -170,19 +189,36 @@ def gather_and_send(self, principal_source, response_instance, req_type,
...
-    def send_request(self, language, req_type, req, req_id=None):
+    def _is_client_running(self, name):
+        return self.clients.get(name, {}).get('status', self.STOPPED) == self.RUNNING
+
+    def send_request(self, language, req_type, req):
         req_id = self.req_id
-        for client_name in self.clients:
+
+        client_names = []
+        if req_type not in self.MUTEX_REQUEST_TYPES:
```
andfoy: Don't we need any language-related check here?

metalogical: We didn't have any such checks before, so I assume they weren't necessary before.

Kite returns an empty response for unsupported languages (updated in this PR), and LSP throws away requests for unsupported languages.

Since we try Kite first, and it will always return a response for all languages, the LSP fallback will always get triggered, even if Kite is first called on an unsupported language, so this should all work.

That being said, I'm open to other suggestions for how to structure this.

spyder/plugins/completion/plugin.py
```diff
@@ -170,19 +193,43 @@ def gather_and_send(self, principal_source, response_instance, req_type,
...
+            client_names.append(kite)
+        else:
+            client_names.append(lsp)
+
+        if (self.get_option('enable_fallback', True) and
+                self._is_client_running(fallback)):
+            client_names.append(fallback)
+
+        for client_name in client_names:
             client_info = self.clients[client_name]
```
ccordoba12: `test_lsp_splitter_close` is failing at this point because it seems the Kite client is not added to the request that test is checking.

metalogical:
@ccordoba12 after a bunch of trial and error, I've fixed the build and addressed your comments

spyder/config/main.py
```diff
@@ -500,6 +500,11 @@
               'advanced/external': False,
               'advanced/stdio': False
              }),
+            ('kite',
+             {
+              'enable': True,
+              'enable_snippets': True,
```
andfoy: @metalogical, how does an user can disable code snippets for Kite?

andfoy:
Does Kite provide code snippets for local functions?

spyder/config/main.py
```diff
@@ -205,7 +205,7 @@
               'tab_always_indent': False,
               'intelligent_backspace': True,
               'automatic_completions': True,
-              'automatic_completions_after_chars': 3,
+              'automatic_completions_after_chars': 1,
```
andfoy: @ccordoba12 What do you think about this one, leaving it in 1 by default may cause regressions, such as #10134

metalogical: based on my testing, the more important value for performance is the 300ms, but I'm willing to change this if you disagree.

spyder/config/main.py
metalogical: currently they cannot (apart from editing the config file), but as a follow-up, we should expose this setting in a UI

metalogical:
@andfoy yes, Kite can provide code snippets for local functions.

spyder/config/main.py
andfoy: Sounds reasonable

andfoy:
ezgif com-video-to-gif(13)

spyder/config/main.py
goanpeca: I will update the UI after this PR is merged!

goanpeca: @andfoy @ccordoba12 since this adds even more granularity. Since we will use one or the other, we do not need to expose a setting for Kite AND a setting for the PYLS right?

spyder/plugins/completion/kite/client.py
```diff
@@ -75,7 +75,7 @@ def perform_http_request(self, verb, url, params=None):
         try:
             http_response = http_method(url, json=params)
         except Exception as error:
-            logger.debug('Kite request error: {0}'.format(error))
+            # this is expected, since Kite may not be running, so we don't log
```
goanpeca:
```suggestion
            # This is expected, since Kite may not be running, so we don't log
```

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -502,13 +500,18 @@ def __init__(self, parent=None):
...
         self.intelligent_backspace = True
 
         # Automatic (on the fly) completions
-        self.automatic_completions = True
+        self.automatic_completions = \
+            CONF.get('editor', 'automatic_completions')
+        self.automatic_completions_after_chars = \
+            CONF.get('editor', 'automatic_completions_after_chars')
+        self.automatic_completions_after_ms = \
+            CONF.get('editor', 'automatic_completions_after_ms')
```
ccordoba12: This changes shouldn't be necessary because:

1. Options from our config system are applied to each CodeEditor at startup and also when an option is changed in our Preferences. If you that that's not the case, then it's a bug and we need to fix it.
2. There's no need for widgets to access our config system directly. Plugins (i.e. the ones that contain widgets) are the ones in charge of that.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -771,9 +774,9 @@ def setup_editor(self,
                      tab_mode=True,
                      strip_mode=False,
                      intelligent_backspace=True,
-                     automatic_completions=True,
-                     automatic_completions_after_chars=3,
-                     automatic_completions_after_ms=300,
+                     automatic_completions=None,
+                     automatic_completions_after_chars=None,
+                     automatic_completions_after_ms=None,
```
ccordoba12: These changes should be reverted. They are only used when testing CodeEditor instances independently from the editor, but they can be changed if needed by modifying CodeEditor's attributes associated with them directly.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1306,7 +1304,8 @@ def toggle_intelligent_backspace(self, state):
         self.intelligent_backspace = state
 
     def toggle_automatic_completions(self, state):
-        self.automatic_completions = state
+        if state is not None:
+            self.automatic_completions = state
```
ccordoba12: This should be reverted because `state` can only be True or False.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1322,13 +1321,15 @@ def set_automatic_completions_after_chars(self, number):
         """
         Set the number of characters after which auto completion is fired.
         """
-        self.automatic_completions_after_chars = number
+        if number is not None:
+            self.automatic_completions_after_chars = number
```
ccordoba12: Please revert this because there's no need for it.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1322,13 +1321,15 @@ def set_automatic_completions_after_chars(self, number):
...
+        if number is not None:
+            self.automatic_completions_after_chars = number
 
     def set_automatic_completions_after_ms(self, ms):
         """
         Set the amount of time in ms after which auto completion is fired.
         """
-        self.automatic_completions_after_ms = ms
+        if ms is not None:
+            self.automatic_completions_after_ms = ms
```
ccordoba12: Please revert.

ccordoba12:
@metalogical, this also needs a merge with master now.

spyder/config/main.py
andfoy: Personally, I would prefer to keep a single option to set all the completion configuration options. However, this will fixed as part of a future completion refectoring that we plan to do for Spyder 5. Maybe we should stick to a single configuration value, or set this one using the current checkbox, otherwise users would get completions even if they explicitly set that option disabled.

metalogical: ok, I'm fine with keeping a single config option. I just thought it was strange that we had to check the LSP config option inside of the Kite plugin. Will revert to the previous way.

spyder/config/main.py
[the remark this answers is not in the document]
metalogical: @ccordoba12 with 1 char (but leaving the 300ms wait) the UX seems to be significantly improved for me (completions show up more frequently), without impacting perceived performance.

That being said, it's not entirely related to this PR. Let me know if you want me to remove this change.

spyder/config/main.py
andfoy: Yes, it's a little bit confusing to use LSP references everywhere, but we will do the corresponding refactoring later on

spyder/config/main.py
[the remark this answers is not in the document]
goanpeca: @metalogical the idea of this setting is to not display "annoying" completions if one is typing single length variables like `i` or `j` or `x`. So on one side we use the ms after stopping as one way of preventing popups, but we also use this setting for disabling any completion for variables of length 1 and 2 (by default, which we took from RStudio and we agree is a valid value).

So please revert this setting as we would like to have it as a default of `3`.

Thanks!

spyder/plugins/editor/widgets/codeeditor.py
metalogical: I can confirm that without this change, the config options are not being propagated here at startup, but *are* when the options are changed.

It seems like checking the `CONF` value here would solve the startup case, but I'm not sure if there's another way of fixing that. I can create an issue to track that and revert it here.

spyder/config/main.py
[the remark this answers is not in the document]
metalogical: I reverted this change for now, but still do think that this default value results in a better UX.

spyder/plugins/editor/widgets/codeeditor.py
metalogical: https://github.com/spyder-ide/spyder/issues/10331

metalogical:
@ccordoba12 I made the changes you requested. Two caveats:

- propagating the options at startup isn't working, which I made a separate issue for: https://github.com/spyder-ide/spyder/issues/10331
- I believe that `automatic_completions_after_chars = 1` is a better default value (but removed the change from this PR)

goanpeca:
>propagating the options at startup isn't working, for which I made a separate issue: #10331

I can take a look at this @metalogical 

>I believe that automatic_completions_after_chars = 1 is a better default value (but removed the change from this PR)

That is the same as not having an option at all so having a default of 1 does not fix the default case for users, which is having an intrusive popup for every single character typed in the editor.

spyder/plugins/completion/kite/providers/document.py
```diff
@@ -46,11 +47,14 @@ def convert_text_snippet(snippet_info):
...
-        snippet_text = text[next_pos:placeholder_end][1:-1]
+        snippet_text = text[next_pos:placeholder_end]
         prev_pos = placeholder['end']
         text_builder.append(standard_text)
         placeholder_number = (i + 1) % total_placeholders
-        snippet = '${%d:%s}' % (placeholder_number, snippet_text)
+        if snippet_text:
+            snippet = '${%d:%s}' % (placeholder_number, snippet_text)
+        else:
+            snippet = '$%d' % (placeholder_number)
```
andfoy: I think this is redundant by the change introduced on line 56

metalogical: We were not previously calling `convert_text_snippet` on non-children completions below (see my change below on new line 129). Now that we are, it bubbled up a bug, where placeholder text returned by Kite may be empty.

If we use `${1:}` for empty placeholder text, the snippet parser doesn't fail outright, but an exception is thrown when trying to insert the snippet. The correct syntax for placeholders without text is `$1`.

As far as I can tell, putting `$0` at the end, as you do on line 56 only fixes the final tabstop. Let me know if I'm missing something here.

its-dhung:
> That is the same as not having an option at all so having a default of 1 does not fix the default case for users, which is having an intrusive popup for every single character typed in the editor.

@goanpeca I'm not quite sure I understand this statement?

Personally, and based on our extensive experience about talking to users about the role of autocompletions in their workflows, I believe most users expect autocompletions to trigger after 1 character. Setting this value to 3 chars is quite arbitrary, and though it may make sense to the people who implemented the logic, it will appear broken or inconsistent to many other users.

If you're worried about the noise of completions, it's worth realizing that the noise is due to the poor performance of the completions engine, *not* because of the triggering after N characters. Making this value of N arbitrarily high is a bandaid solution, not a real one.

@ccordoba12, @goanpeca I would really reconsidering the default value of this. No other IDE that provides autocompletions has an arbitrary setting like this and I don't believe implementing this setting would provide some sort of UX breakthrough that other IDE authors aren't aware of.

spyder/plugins/completion/kite/providers/document.py
andfoy: So, this should cover cases when the cursor lands inside a function call brackets?

goanpeca:
RStudio does, which is where we drew inspiration from

Screen Shot 2019-09-30 at 13 38 06

its-dhung:
Good to know, though our experience is in Python and no other popular Python IDE does. In either case, I don't think this should hold up the merging of this PR. I just think we should take a more principled and user focused approach when it comes to settings like this that can drastically alter the user experience.

spyder/plugins/completion/kite/providers/document.py
metalogical: that's right (there are also some other cases where Kite currently returns empty placeholders / tabstops without text).

andfoy: Then, we're good with this one!

spyder/plugins/completion/kite/plugin.py
```diff
@@ -46,27 +49,32 @@ def http_client_ready(self, languages):
...
     def shutdown(self):
         self.client.stop()
         if self.kite_process is not None:
             self.kite_process.kill()
 
     def update_configuration(self):
-        enable_code_snippets = CONF.get('lsp-server', 'code_snippets')
-        self.client.enable_code_snippets = enable_code_snippets
+        self.client.enable_code_snippets = \
+            CONF.get('lsp-server', 'code_snippets')
```
ccordoba12: We prefer to avoid the usage of line continuations in new work. So please change the two lines above to be

```python
self.client.enable_code_snippets = CONF.get('lsp-server',
    'code_snippets')
```

metalogical: fixed!

PR #10321, opened 2019-09-27
PR: Show exact match completions

metalogical:
## Description of Changes

This is a proposal to change the completions filtering behavior to show completions that exactly match what the user has typed under the cursor.

As far as I can tell, the main reason to have the filtering was that the fallback plugin would generate "identity" completions via tokenization, so I filter out tokens that are under the cursor there.

The advantage to showing exact matches is primarily if there are many other completions, and exact matches are not shown, then if I hit enter, the wrong completion will be inserted:

Screen Shot 2019-09-27 at 11 47 39 AM

Please let me know if you disagree with this change. Also, I'm sure tests will fail, so I'll fix those.

### Affirmation

### DIFF
--- a/spyder/plugins/completion/fallback/actor.py
+++ b/spyder/plugins/completion/fallback/actor.py
@@ -53,7 +53,7 @@ class FallbackActor(QObject):
         self.thread.started.connect(self.started)
         self.sig_mailbox.connect(self.handle_msg)
 
-    def tokenize(self, text, language):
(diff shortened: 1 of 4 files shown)


andfoy:
Sounds good to me, what do you think @ccordoba12

ccordoba12:
I also agree with this change too. Thanks a lot @metalogical for helping us with to improve this!

sergeyFlakon:
fix is working fine for me

PR #10192, opened 2019-09-13
PR: Add missing completions call to action

edzkite:
## Description of Changes
Screen Shot 2019-09-05 at 11 43 48 PM

This PR adds a call to action when users aren’t getting completions, indicating they can get more completions if they add Kite.
The main intention is to educate the user about what Kite adds at a moment when they could benefit from it. It is easy to dismiss forever.

This PR is not ready to merge until the Kite enabled/disabled toggle and install flow is implemented.

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct:

PR #10251, opened 2019-09-18
PR: Add an installation UI for Kite

dalthviz:
## Description of Changes

* [x] Wrote at least one-line docstrings (for any new functions)
* [ ] Added unit test(s) covering the changes (if testable)

* [x] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

if

### Issue(s) Resolved

Fixes #

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct: dalthviz

PR #10332, opened 2019-09-30
PR: Add a call-to-action for Kite

metalogical:
## Description of Changes

This PR adds a call-to-action to educate users about Kite at a moment the user would benefit from Kite completions (only if no other completions are shown to the user). It is an improved version of #10192.
Screen Shot 2019-10-07 at 6 06 18 PM

We err on the side of minimizing false positives, so the dialog is not displayed in every case where Kite may provide completions, but only in cases where where we are highly confident Kite would produce sensible results.

The detection is implemented only for the following case, using regular expressions to identify the function that must be understood by Kite in order to provide completions:
```python
foo = full.path.to.function.that.kite.understands(...)
...
x = foo.bar.b‸
```
We then match this function against a pre-computed bloom filter containing common function paths that Kite understands.

### Affirmation


metalogical:
Now that the install UI is in, I'm updating this to trigger that. I'm also cleaning up some design elements (e.g. buttons instead of links)

spyder/config/main.py
```diff
@@ -92,6 +92,7 @@
               'completion/size': (300, 180),
               'report_error/remember_me': False,
               'report_error/remember_token': False,
+              'show_kite_cta': True,
```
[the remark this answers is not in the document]
metalogical: We need a way of having "Dismiss Forever" state persist. Those options seem separate (e.g. Kite can be enabled by default, without Kite actually being installed, in which case the CTA should still show, until the user clicks "dismiss").

This is analogous to the `show_installation_dialog` option added for the purposes of dismissing the install dialog: https://github.com/spyder-ide/spyder/pull/10251/files#diff-33b5b1425c385f61c36db4a4f1350bf9R73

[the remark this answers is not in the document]
metalogical: moving to the `kite` section of the config though.

spyder/plugins/completion/kite/ui/cta.py
```diff
@@ -0,0 +1,181 @@
...
+        labels = QFrame(self)
+        labels_layout = QHBoxLayout()
+        labels_layout.setContentsMargins(5, 5, 5, 5)
+        labels_layout.setSpacing(15)
+        labels_layout.addStretch()
+        labels.setLayout(labels_layout)
+        # TODO Learn More & Install Kite should trigger the install flow UI
+        # and disable the CTA for the session
+        labels_layout.addWidget(self._create_link_label(
+            translate("Install Kite"), "https://kite.com/download/"))
```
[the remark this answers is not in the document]
metalogical: Did you mean replace with the `_` way you suggested above (that's what I've done)? Or remove entirely? If the latter, I can knock that out whenever you say so.

metalogical:
@ccordoba12 I've addressed all your comments, excepting one question about translation, which I left inline.

Regarding your other concern: we autogenerated a list of around 13K function paths (fully qualified, and also a few common aliases) for which we understand the return types / semantics. We did some filtering, etc as well to make sure the list is not excessively large (no private functions, only functions exported in top-level packages, etc).

[paths.txt](https://github.com/spyder-ide/spyder/files/3692825/paths.txt)

The list still contains a decent amount of noise, which we could clean up a bit, but didn't expend the energy, since we're happy with the resulting size of the bloom filter.

spyder/config/main.py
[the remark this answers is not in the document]
metalogical: oh nvm, I see what you're saying now 🤦‍♂ I added the setting there and here.

I think the right approach here is to address the merge in whichever PR ends up getting merged second.

spyder/plugins/completion/kite/widgets/calltoaction.py
```diff
@@ -0,0 +1,151 @@
...
+            self._dismiss_forever()
+
+    def set_color_scheme(self, color_scheme):
+        if not isinstance(color_scheme, dict):
+            color_scheme = get_color_scheme(color_scheme)
+        bg_color = color_scheme['background']
+        border_color = color_scheme['sideareas']
+        text_color, __, __ = color_scheme['normal']
+        button_color = color_scheme['currentline']
+        hover_color = color_scheme['currentcell']
```
ccordoba12: I think using our color scheme for the CTA colors is the wrong thing to do. I mean `color_scheme['normal']` is fine, but the other colors vary wildly aming color schemes. For instance, using `color_scheme['background']` in light themes makes no sense because that color is `white`. And `color_scheme['currentline']` is a shade purple in other schemes.

@metalogical, I think you should use the same colors as the completion widget for background and foreground, and also preserve the same colors and hover behavior for buttons as any other widget.

ccordoba12:
> Regarding your other concern: we autogenerated a list of around 13K function paths (fully qualified, and also a few common aliases) for which we understand the return types / semantics

Nice, thanks for sharing those paths! That's really useful to understand for us to understand when the CTA is going to be displayed.

spyder/plugins/completion/kite/widgets/calltoaction.py
metalogical: done: I had to use `setFrameStyle` to force the frame style to match `QListWidget`, use `setAutoFillBackground` to prevent the frame from being transparent in the Light UI mode, and copied the window flags from the completion widget.

It still doesn't look 100% identical, but extremely close: the completions widget has rounded corners, and in light mode, has a blue highlight around the border. Not sure how that happens.

spyder/plugins/completion/kite/widgets/calltoaction.py
```diff
@@ -0,0 +1,134 @@
...
+      " Get completions for this case and more by installing Kite.")
+)
+
+
+class KiteCallToAction(QFrame):
+    def __init__(self, textedit, ancestor):
+        super(KiteCallToAction, self).__init__(ancestor)
+        self.textedit = textedit
+
+        self.setFrameStyle(54)
```
dalthviz: Please change this to `self.setFrameStyle(QFrame.StyledPanel | QFrame.Plain)` (the equivalent using constants to the value `54`)

spyder/plugins/completion/kite/widgets/calltoaction.py
```diff
@@ -0,0 +1,134 @@
...
+        self.setAutoFillBackground(True)
+        self.setWindowFlags(Qt.SubWindow | Qt.FramelessWindowHint)
+        self.setFocusPolicy(Qt.NoFocus)
```
dalthviz: Please, below this line, add the following to have the correct behavior of the border when using the dark theme:
```python
    if is_dark_interface():
            self.setObjectName("kite-call-to-action")
            self.setStyleSheet(
                self.styleSheet() + ("#kite-call-to-action "
                                     "{ border: 1px solid; "
                                     "  border-color: #32414B; "
                                     "  border-radius: 4px;} "
                                     "#kite-call-to-action:hover "
                                     "{ border:1px solid #148CD2; }"))
```

You will need to add the import for `is_dark_interface` at the imports section of the file. The import is 

```python
from spyder.utils.icon_manager import is_dark_interface
```

metalogical: lovely, this fixes it, thanks!

spyder/plugins/completion/kite/widgets/calltoaction.py
```diff
@@ -0,0 +1,144 @@
+# -*- coding: utf-8 -*-
+
+# Copyright © Spyder Project Contributors
+# Licensed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+
+from qtpy.QtCore import Qt, QTimer, QUrl
```
dalthviz: Also @naman I think there are some imports that are not needed anymore (like QTimer, QUrl, QDesktopServices, etc). Please remove them

PR #10354, opened 2019-10-04
enable better configuration of completions clients

### LINKED ISSUE #10071: Kite is always started

metalogical:
## Description of Changes

1. Add ability to enable/disable each of Kite, LSP, and fallback providers independently
Screen Shot 2019-10-03 at 5 04 29 PM

2. Make configuration changes propagate correctly. In particular, enabling/disabling Kite & LSP in the middle of a session was not working.
    * For Kite, we just needed to make `DOCUMENT_DID_CHANGE` handler add/update the file text regardless of whether or not it already exists in the cache.
    * For LSP we need to start the LSP client/server for the relevant language on request (if it's not already running). This will cause a bit of lag for the first few characters typed after enabling LSP, but that should be fine, since the option shouldn't be changed frequently.
3. Make sure each of the 6 combinations of Kite/LSP/fallback work correctly.
    * There was a bug where when fallback is the only provider enabled, it wasn't working properly. The fixes for this are in the top-level completions plugin.

I'm not sure if/how to test this, so please let me know if there are obvious unit tests I should add.

### Issue(s) Resolved

Fixes #10071 

### Affirmation

### DIFF
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -470,6 +470,7 @@ DEFAULTS = [
             ('appearance', APPEARANCE),
             ('lsp-server',
              {
+              'enable': True,
               # This option is not used with the LSP server config
               # It is used to disable hover hints in the editor
               'enable_hover_hints': True,
@@ -500,9 +501,15 @@ DEFAULTS = [
               'advanced/external': False,
               'advanced/stdio': False
              }),
+            ('fallback-completions',
+             {
+              'enable': True,
+             }),
             ('kite',
              {
               'enable': True,
+              'autostart': True,
+              'cta': True,
              }),
             ]
 
--- a/spyder/plugins/completion/kite/plugin.py
+++ b/spyder/plugins/completion/kite/plugin.py
@@ -61,13 +61,19 @@ class KiteCompletionPlugin(SpyderCompletionPlugin):
     def start(self):
         # Always start client to support possibly undetected Kite builds
         self.client.start()
+
+        # Autostart engine
+        if not self.autostart:
+            return
         installed, path = check_if_kite_installed()
-        if installed:
-            logger.debug('Kite was found on the system: {0}'.format(path))
-            running = check_if_kite_running()
-            if not running:
-                logger.debug('Starting Kite service...')
-                self.kite_process = run_program(path)
+        if not installed:
+            return
+        logger.debug('Kite was found on the system: {0}'.format(path))
+        running = check_if_kite_running()
+        if running:
+            return
+        logger.debug('Starting Kite service...')
+        self.kite_process = run_program(path)
 
     def shutdown(self):
         self.client.stop()
@@ -77,4 +83,5 @@ class KiteCompletionPlugin(SpyderCompletionPlugin):
     def update_configuration(self):
         self.client.enable_code_snippets = CONF.get('lsp-server',
                                                     'code_snippets')
+        self.autostart = self.get_option('autostart')
         self.enabled = self.get_option('enable')
--- a/spyder/plugins/completion/languageserver/plugin.py
+++ b/spyder/plugins/completion/languageserver/plugin.py
@@ -52,14 +52,7 @@ class LanguageServerPlugin(SpyderCompletionPlugin):
         self.requests = set({})
         self.register_queue = {}
 
-        # Register languages to create clients for
-        for language in self.get_languages():
-            self.clients[language] = {
-                'status': self.STOPPED,
-                'config': self.get_language_config(language),
-                'instance': None
-            }
-            self.register_queue[language] = []
+        self.update_configuration()
 
     def register_file(self, language, filename, codeeditor):
         if language in self.clients:
@@ -68,10 +61,6 @@ class LanguageServerPlugin(SpyderCompletionPlugin):
                 self.register_queue[language].append((filename, codeeditor))
             else:
                 language_client.register_file(filename, codeeditor)
-
-    def get_option(self, option):
-        """Get an option from our config system."""
-        return CONF.get(self.CONF_SECTION, option)
 
     def get_languages(self):
         """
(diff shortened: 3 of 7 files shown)


andfoy:
@metalogical I think we cannot disable LSP completely, as we have other requests that depend on it

metalogical:
@andfoy thanks for the tip, removed that checkbox.

spyder/plugins/completion/languageserver/plugin.py
```diff
@@ -52,14 +52,7 @@ def __init__(self, parent):
...
         self.register_queue = {}
 
-        # Register languages to create clients for
-        for language in self.get_languages():
-            self.clients[language] = {
-                'status': self.STOPPED,
-                'config': self.get_language_config(language),
-                'instance': None
-            }
-            self.register_queue[language] = []
```
ccordoba12: I don't think replacing these lines by `update_configuration` is the right thing to do here. I mean, here we're only creating the list of client to start clients for, which is done later. However, `udpate_configuration` tries to restart clients too.

@andfoy, what do you say?

spyder/plugins/completion/plugin.py
```diff
@@ -110,21 +110,31 @@ def receive_response(self, completion_source, req_id, resp):
...
-            # since we don't have "good" completions from the current response
-            if any(self._is_client_running(source) and
-                   source not in request_responses['sources']
-                   for source in wait_for):
-                return
-        else:
-            del self.requests[req_id]
 
         with QMutexLocker(self.collection_mutex):
+            request_responses = self.requests[req_id]
```
ccordoba12: Why the changes inside this QMutexLocker are necessary as part of this PR?

spyder/config/main.py
```diff
@@ -470,6 +470,7 @@
             ('appearance', APPEARANCE),
             ('lsp-server',
              {
+              'enable': True,
```
ccordoba12: I think we should remove this option (and its usage in this PR) because we can't really turn off the LSP completely.

spyder/plugins/completion/languageserver/plugin.py
andfoy: I think this is valid because we're performing the same initialization on ``update_configuration``

spyder/plugins/completion/languageserver/plugin.py
```diff
@@ -290,17 +280,19 @@ def receive_response(self, response_type, response, language, req_id):
                 self.COMPLETION_CLIENT_NAME, req_id, response)
 
     def send_request(self, language, request, params, req_id):
-        if language in self.clients:
+        if self.enabled and language in self.clients:
+            # Ensure client is started
+            self.start_client(language)
```
andfoy: Some comments here:
1. We should remove all references to self.enabled, as it is True by default
2. On which cases should the client be started from scratch here? We currently enable them if a file is opened, or during startup, maybe there's some corner case we're not taking into account?

spyder/plugins/completion/plugin.py
andfoy: I think they are to prevent concurrency problems. Is this right @metalogical?

spyder/plugins/completion/languageserver/plugin.py
metalogical: ah yes this was added when LSP could be enabled in the middle of a session, but that's no longer possible. If you prefer, I can revert these changes as well.

spyder/plugins/completion/plugin.py
metalogical: Yes, I noticed some issues when both Kite & LSP were disabled, since fallback completions would not end up being shown. So I just moved more code into the locked segment.

spyder/plugins/completion/languageserver/plugin.py
metalogical: I removed all this stuff, since we can always add it back in the future if we need it.

metalogical:
@ccordoba12 @andfoy ready for another pass. Thanks for the reviews!

andfoy:
@ccordoba12 LGTM

spyder/plugins/completion/plugin.py
ccordoba12: Ok, thanks @metalogical for the explanation.

ccordoba12:
@metalogical, last question from my side: Do we really need the `autostart` option? I mean, why is not enough with the option to enable/disable Kite completions?

Or in other words: what's the benefit (from the user point of view) of activating/deactivating this `autostart` option?

metalogical:
@ccordoba12 we have had experience with other editor integrations where the user did not appreciate Kite automatically starting, and wanted to manually control when Kite should run.

I'm ok with removing the option, but that was the motivation for adding it for our other editor plugins.

cc @its-dhung

ccordoba12:
> we have had experience with other editor integrations where the user did not appreciate Kite automatically starting, and wanted to manually control when Kite should run.

I understand that, but isn't the `Enable Kite` option enough? I mean, as a user I'd expect that option to let me turn on/off Kite on the fly during the current session.

ccordoba12:
In other words, I think "Enable Kite" and "Autostart" seem to have the same meaning, so I'd prefer to remove the second option.

its-dhung:
> In other words, I think "Enable Kite" and "Autostart" seem to have the same meaning, so I'd prefer to remove the second option.

Yeah that's fine with me.

metalogical:
@ccordoba12 done, and updating the screenshot

spyder/plugins/completion/kite/plugin.py
```diff
@@ -90,13 +90,19 @@ def start_client(self, language):
     def start(self):
         # Always start client to support possibly undetected Kite builds
         self.client.start()
+
+        # Autostart engine
```
ccordoba12: This comment is not necessary anymore.

metalogical: removed

PR #10419, opened 2019-10-15
PR: Handle paths for recent files on a different mount than the project root path

### LINKED ISSUE #10045: ValueError when closing project on USB stick

1. Open project from USB stick with Spyder 3
2. Update to Spyder 4 beta
3. Try closing project. Results in an error.

### Traceback
```python-traceback
  File "C:\Users\jmgro\Anaconda3\lib\site-packages\spyder\plugins\projects\plugin.py", line 342, in close_project
    self.main.editor.get_open_filenames())
  File "C:\Users\jmgro\Anaconda3\lib\site-packages\spyder\plugins\projects\plugin.py", line 427, in set_project_filenames
    self.current_active_project.set_recent_files(recent_files)
  File "C:\Users\jmgro\Anaconda3\lib\site-packages\spyder\plugins\projects\api.py", line 76, in set_recent_files
    for recent_file in recent_files]
  File "C:\Users\jmgro\Anaconda3\lib\site-packages\spyder\plugins\projects\api.py", line 76, in <listcomp>
    for recent_file in recent_files]
  File "C:\Users\jmgro\Anaconda3\lib\ntpath.py", line 562, in relpath
    path_drive, start_drive))
ValueError: path is on mount 'C:', start on mount 'E:'
```

## Versions

* Spyder version: 4.0.0b4 
* Python version: 3.7.3
* Qt version: 5.9.6
* PyQt5 version: 5.9.2

### Dependencies

```
qdarkstyle >=2.6.4:  2.7 (OK)
sympy >=0.7.3     :  1.4 (OK)
cython >=0.21     :  0.29.12 (OK)
qtconsole >=4.5.2 :  4.5.3 (OK)
IPython >=4.0     :  7.6.1 (OK)
```

dalthviz:
## Description of Changes

Handle recent files saving in cases where paths doesn't have a relative path to the project root path

### Issue(s) Resolved

Fixes #10045

### Affirmation

### DIFF
--- a/spyder/plugins/projects/projecttypes/__init__.py
+++ b/spyder/plugins/projects/projecttypes/__init__.py
@@ -63,13 +63,12 @@ class BaseProject(object):
         """Set a list of files opened by the project."""
         processed_recent_files = []
         for recent_file in recent_files:
-            if os.path.isfile(recent_file):
-                try:
-                    relative_recent_file = os.path.relpath(
-                        recent_file, self.root_path)
-                    processed_recent_files.append(relative_recent_file)
-                except ValueError:
-                    processed_recent_files.append(recent_file)
+            try:
+                relative_recent_file = os.path.relpath(
+                    recent_file, self.root_path)
+                processed_recent_files.append(relative_recent_file)
+            except ValueError:
+                processed_recent_files.append(recent_file)
         files = list(OrderedDict.fromkeys(processed_recent_files))
         self.config.set('main', 'recent_files', files)
 
@@ -79,11 +78,6 @@ class BaseProject(object):
         recent_files = [recent_file if os.path.isabs(recent_file)
                         else os.path.join(self.root_path, recent_file)
                         for recent_file in recent_files]
-
-        for recent_file in recent_files[:]:
-            if not os.path.isfile(recent_file):
-                recent_files.remove(recent_file)
-
         return list(OrderedDict.fromkeys(recent_files))
 
 
--- a/spyder/plugins/projects/projecttypes/__init__.py
+++ b/spyder/plugins/projects/projecttypes/__init__.py
@@ -63,12 +63,13 @@ class BaseProject(object):
         """Set a list of files opened by the project."""
         processed_recent_files = []
         for recent_file in recent_files:
-            try:
-                relative_recent_file = os.path.relpath(
-                    recent_file, self.root_path)
-                processed_recent_files.append(relative_recent_file)
-            except ValueError:
-                processed_recent_files.append(recent_file)
+            if os.path.isfile(recent_file):
+                try:
+                    relative_recent_file = os.path.relpath(
+                        recent_file, self.root_path)
+                    processed_recent_files.append(relative_recent_file)
+                except ValueError:
+                    processed_recent_files.append(recent_file)
         files = list(OrderedDict.fromkeys(processed_recent_files))
         self.config.set('main', 'recent_files', files)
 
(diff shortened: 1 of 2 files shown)


spyder/plugins/projects/projecttypes/__init__.py
```diff
@@ -75,11 +78,6 @@ def get_recent_files(self):
         recent_files = [recent_file if os.path.isabs(recent_file)
                         else os.path.join(self.root_path, recent_file)
                         for recent_file in recent_files]
-
-        for recent_file in recent_files[:]:
-            if not os.path.isfile(recent_file):
-                recent_files.remove(recent_file)
-
```
ccordoba12: Why did you remove these lines? Without these lines `get_recent_files` does nothing.

dalthviz: I made this change (removing the validation of the path being a file) to check the tests (`test_set_get_project_filenames_when_closing` and `test_set_get_project_filenames_when_switching`). Seems like the existing tests were passing because of the mutability property of a list (the `set_recent_files` method was changing directly the `opened_files` list that at the end is the object passed by parameter, so the [assert](https://github.com/spyder-ide/spyder/blob/aafb1c21f6e8c58997e43ccfc8cf8d06e2feb9c6/spyder/plugins/projects/tests/test_plugin.py#L197) was checking the return with a list changed by the `set_recent_files` method (the recent files in the tests always were an empty list, since the paths we are testing here aren't actually created).

However, probably we need here to change the tests to create the files and add again the check for the path being a file. So this is not ready yet :)

ccordoba12: > So this is not ready yet :)

@dalthviz, then please don't forget to add WIP to the PR title.

dalthviz: Yep :+1:

goanpeca:
@dalthviz just to be sure, could you run the tests reverting the fix on code and making sure it fails? 

This is to ensure that we test the regression.

spyder/plugins/projects/tests/test_plugin.py
```diff
@@ -175,18 +175,52 @@ def test_open_project_uses_visible_config(projects, tmpdir, value):
     assert projects.dockwidget.isVisible() == value
 
 
-def test_set_get_project_filenames_when_closing(create_projects, tmpdir):
+def test_set_get_project_filenames_when_closing_no_files(create_projects,
+                                                         tmpdir):
```
ccordoba12: I don't understand how this new test checks a regression for issue #10045

dalthviz: This is to check the old unwanted behavior of the `set_recent_files` where the method changed the `opened_files` list (because the old method modified the parameter as explained above) making the old tests pass. If you try to run this test with the old implementation of the `set_recent_files` method it fails.

ccordoba12: But is not the issue related to using relative paths? Or what's the real problem here?

dalthviz: Yes, the main issue is related to the root path of a project being in a different mount path/drive than the files currently open. So, for example working on Windows, the project is on `C` but the files are on `D`, so no relative path exist between the project root path and the file. (that way I added the `try-except`).

However, while doing the fix, I discover that the tests which were supposed to check that the recent files elements worked on projects were kind of faulty, partially because of the way `set_recent_files` was implemented (the variable `opened_files` use to do the assert in the test, was modified by the `set_recent_files` method itself).

ccordoba12: Ok, thanks @dalthviz for the thorough explanation!

PR #10428, opened 2019-10-16
PR: Test debug thread

### LINKED ISSUE #8801: Cannot debug a function running in a thread
## Problem Description
Cannot debug a function that runs inside a thread

1. 
2. 
3. 

### Example code

from threading import Thread 

class projectLoader (Thread):
    def __init__(self):
        Thread.__init__(self)
        self.runError='' #error message        
        self.start()    # start
        
    def run(self): #I cannot debug this function
        count=0
        for idx in range(1000000):
            count+=1
        print(str(count))

projectLoader()

## Versions

* Spyder version: 3.3.2
* Operating System name/version: Windows

impact27:
## Description of Changes

Test if breakpoints work in an unsaved file and a thread

### Issue(s) Resolved

Fixes #8801

### Affirmation


impact27:
retrigger tests.

impact27:
This works but I am not sure we want to support it: 
While debugging, if several threads reach a breakpoint at the same time, what whould be the behaviour? only one can block, so the others have to skip the breakpoint, which is strange from a user perspective.

impact27:
closing for now because I am not happy with how it handles simultaneous stopping

PR #10438, opened 2019-10-17
PR: Use a new comm socket for comms

impact27:
## Description of Changes

Calling `do_one_iteration` is not a good idea as it might process messages out of order.

Here, a new socket is created. It is a one way socket designed to skip the message queue in the kernel, and runs in a thread. This means that:

 - Only messages that are explicitely sent to this socket are processed out of order, instead of all messages as it is the case now.
 - Any thread can make a blocking call to the frontend without needing to start an event loop.
 - We can set breakpoints while executing!

### Issue(s) Resolved

Fixes #

### Affirmation


impact27:
retrigger tests

impact27:
This is ready for review

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -115,7 +114,7 @@ def call_kernel(self, interrupt=False, blocking=False, callback=None):
     def set_kernel_client_and_manager(self, kernel_client, kernel_manager):
         """Set the kernel client and manager"""
         self.kernel_manager = kernel_manager
-        self.kernel_client = kernel_client
+        self.kernel_client = self.patch_kernel_client(kernel_client)
```
ccordoba12: Let's leave this as before, and call `redefine_complete_for_dbg` below, like this

```python
self.spyder_kernel_comm.open_comm(kernel_client)

# Redefine the complete method to work while debugging.
self.redefine_complete_for_dbg(self.kernel_client)
```

PR #10450, opened 2019-10-18
PR: Increase timeout for several calls to the kernel

### LINKED ISSUE #10449: Kernel times out when sending big variable (it worked in Spyder 3)
Hi, all. there is problem when I edited the dict with 40000keys and values whose shape is (100,1) .
360截图18180711365759

ccordoba12:
This PR also fixes and improves the warning we show when a value is too large before trying to retrieve it from the kernel.

Fixes #10449.


spyder/plugins/ipythonconsole/widgets/namespacebrowser.py
```diff
@@ -85,7 +85,11 @@ def get_value(self, name):
             return self.call_kernel(
                 interrupt=True, blocking=True).get_value(name)
         except (TimeoutError, UnpicklingError):
-            raise ValueError('Kernel did not send the value.')
+            msg = _("The variable is either too big or not picklable."
+                    "<br><br>"
```
goanpeca: Is there a way to know the difference between errors? (too big or not pickable?)

spyder/plugins/ipythonconsole/widgets/namespacebrowser.py
```diff
@@ -85,7 +85,11 @@ def get_value(self, name):
...
+                    "<i>Note</i>: Please don't report this problem on "
+                    "Github, there's nothing to do about it.")
```
goanpeca: I understand the idea, but this is a weird message to give to the user :-|

spyder/plugins/ipythonconsole/widgets/namespacebrowser.py
ccordoba12: Yes. TimeoutError is generated when the variable is too big and UnpicklingError when it's not picklable. I'll generate different messages for each one.

spyder/plugins/ipythonconsole/widgets/namespacebrowser.py
ccordoba12: There's really nothing we can do about it.

PR #10464, opened 2019-10-19
PR: Print shell content upon failure and try to correct segfaults in IPython console tests

impact27:
## Description of Changes

### Issue(s) Resolved

Fixes #

### Affirmation

### DIFF
--- a/spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py
+++ b/spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py
@@ -164,16 +164,24 @@ def ipyconsole(qtbot, request):
                               is_sympy=is_sympy,
                               is_cython=is_cython)
     window.setCentralWidget(console)
-
-    # Close callback
-    def close_console():
-        console.closing_plugin()
-        console.close()
-    request.addfinalizer(close_console)
-
-    qtbot.addWidget(window)
     window.show()
-    return console
+
+    yield console
+
+    # Print shell content if failed
+    if request.node.rep_setup.passed:
+        if request.node.rep_call.failed:
+            # Print content of shellwidget and close window
+            print(console.get_current_shellwidget(
+                )._control.toPlainText())
+            client = console.get_current_client()
+            if client.info_page != client.blank_page:
+                print('info_page')
+                print(client.info_page)
+
+    # Close
+    console.closing_plugin()
+    console.close()
 
 
 # =============================================================================
@@ -237,14 +245,14 @@ def test_get_calltips(ipyconsole, qtbot, function, signature, documentation):
         qtbot.keyClicks(control, 'np.' + function + '(')
 
     # Wait a little bit for the calltip to appear
-    qtbot.wait(500)
+    qtbot.waitUntil(lambda: control.calltip_widget.isVisible())
 
     # Assert we displayed a calltip
     assert control.calltip_widget.isVisible()
 
     # Hide the calltip to avoid focus problems on Linux
     control.calltip_widget.hide()
-    
+
     # Check spected elements for signature and documentation
     for element in signature:
         assert element in control.calltip_widget.text()
@@ -384,7 +392,7 @@ def test_tab_rename_for_slaves(ipyconsole, qtbot):
 
     cf = ipyconsole.get_current_client().connection_file
     ipyconsole._create_client_for_kernel(cf, None, None, None)
-    qtbot.wait(1000)
+    qtbot.waitUntil(lambda: len(ipyconsole.get_clients()) == 2)
 
     # Rename slave
     ipyconsole.rename_tabs_after_change('foo')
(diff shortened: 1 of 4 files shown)


impact27:
I think removing `qtbot.addWidget(window)` corrects segfaults. Indeed, `qtbot` calls `.deleteLater()`, which deletes the underlying C++ item while leaving the python wrapper alive.

goanpeca:
We should probably add a fixed and known timeout to all the waitUntil conditions.

impact27:
> We should probably add a fixed and known timeout to all the waitUntil conditions.

The timeout is 1000ms. I don't know how to change it

goanpeca:
It is probably an argument to that method

goanpeca:
`waitUntil(callback, timeout=1000)`

impact27:
Thanks! it was not clear from the doc (https://pytest-qt.readthedocs.io/en/latest/wait_until.html#)

impact27:
This is ready for review

spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py
```diff
@@ -237,14 +251,11 @@ def test_get_calltips(ipyconsole, qtbot, function, signature, documentation):
         qtbot.keyClicks(control, 'np.' + function + '(')
 
     # Wait a little bit for the calltip to appear
-    qtbot.wait(500)
-
-    # Assert we displayed a calltip
-    assert control.calltip_widget.isVisible()
```
ccordoba12: I don't like too much changing `assert` statements by `qtbot.waitUntil` ones because the latter ones are not really tests.

@goanpeca, what do you think?

goanpeca: I agree with @ccordoba12. We should tests with assert in generals and use `waitUntil` to replace `qtbot.wait` that are generic and might take longer that waitUntil

impact27: I see your point, but here I am not replacing just an `assert`, I am replacing a `wait` + `assert`.
I mean the trade-off is: Faster tests vs slightly more explicit tests?

impact27: My point of view is:
 - If something shouldn't change now => use `assert`
 - If something should change now => use `assert`
 - If something shouldn't change in the future => use `wait` + `assert`
 - If something should change in the future => use `waitUntil` to continue the test as quickly as possible

jitseniesen: You can also use `waitUntil` and `assert`, this is verbose but fast and explicit.

impact27: If I write:

```
qtbot.waitUntil(lambda: condition)
assert condition
```
Then `assert condition` can not fail (Because it has to be `True` to get out of the `waitUntil`)

goanpeca: @impact27 the main point is that explicit tests should use assert and not waituntil. If we want to speed up tests to wait for a specific thing we can use waitUntil but we have to use assert to make it explicit.

If the condition is to wait for some text (any text) then we should check if the context is text and then assert the actual exact text we are waiting.

>Then assert condition can not fail

Maybe, but it makes things explicit. Again waitUntil should be used to control flow of the runnning test, and not to make the actual assertions.

ccordoba12:
@impact27, these changes now make our tests on Linux very flaky (they're segfaulting every now and then), so this is not good.

Besides, one of your previous PRs made our tests in macOS / Python 3.6 to start segfaulting too. Could you let me know which one was it so I can revert it?

I mean, I really appreciate your efforts on this area but having a failure in our test suite delays or even blocks the work of the rest of the team, so that's something I need to fix ASAP.

impact27:
> @impact27, these changes now make our tests on Linux very flaky (they're segfaulting every now and then), so this is not good.
> 
> Besides, one of your previous PRs made our tests in macOS / Python 3.6 to start segfaulting too. Could you let me know which one was it so I can revert it?
> 
> I mean, I really appreciate your efforts in this area but having a failure in our test suite delays or even blocks the work of the rest of the team, so that's something I need to fix ASAP.

I am trying to understand what is happening. 

 - The segfaults in OSX seems to come from qtbot calling `deleteLater`. This is why I am removing `qtbot.addWidget(window)` in this PR.

 > these changes now make our tests on Linux very flaky (they're segfaulting every now and then)

Are they? They are failing for some reason but I don't see segfaults? or am I missing something?

ccordoba12:
> Are they? They are failing for some reason but I don't see segfaults? or am I missing something?

Sorry, my bad. They're hanging, not segfaulting.

impact27:
So I would say that the segfault in OSX would be fixed by this PR. The errors in linux seems to be caused by `kernel_manager.start_kernel(stderr=stderr_handle, **kwargs)` failing for some reason.

ccordoba12:
> The errors in linux seems to be caused by kernel_manager.start_kernel(stderr=stderr_handle, **kwargs) failing for some reason

Ok, but they were working fine before.

impact27:
I know, they where working fine yesterday as well. These failure are appearing only today for some reason.

impact27:
OSError: [Errno 12] Cannot allocate memory

This doesn’t look linked to my changes?

impact27:
looks like it works now. The failure in miniconda.sh is definitively not related to my code. This solves the segfault in OSX I think

impact27:
Retriggering the tests to make sure it works correctly. All the tests pass in < 20 min.

impact27:
Do you know of a way of flagging `miniconda.sh` as flaky? (in the travis install scripts)

PR #10493, opened 2019-10-22
Cooperative completions

### LINKED ISSUE #10447: Kite completions not being shown
## Problem Description
Sometimes Kite completions aren't shown even if they are available and if Kite has returned faster than LSP.

1. Enable Kite
2. Type the following in a file `import matplotlib.py`
3. Request completions manually if necessary

Based on the logs, I would expect to see completions from Kite since Kite returned faster than LSP:

```
2019-10-17 15:27:25,235 [DEBUG] [spyder.plugins.completion.languageserver.client] -> python request: textDocument/completion
2019-10-17 15:27:25,237 [DEBUG] [spyder.plugins.completion.fallback.actor] -> Got request id 74: textDocument/completion for file /Users/dhung/src/tests/spyder.py
2019-10-17 15:27:25,240 [DEBUG] [spyder.plugins.completion.plugin] -> Completion plugin: Request 74 Got response from fallback
2019-10-17 15:27:25,242 [DEBUG] [urllib3.connectionpool] -> http://localhost:46624 "POST /clientapi/editor/event HTTP/1.1" 200 0
2019-10-17 15:27:25,243 [DEBUG] [spyder.plugins.completion.plugin] -> Completion plugin: Request 73 Got response from kite
2019-10-17 15:27:25,243 [DEBUG] [spyder.plugins.completion.kite.client] -> Perform textDocument/completion request with id 74
2019-10-17 15:27:25,245 [DEBUG] [spyder.plugins.completion.plugin] -> Completion plugin: Request 74 Got response from lsp
2019-10-17 15:27:25,245 [DEBUG] [spyder.plugins.completion.plugin] -> Gather responses for textDocument/completion
2019-10-17 15:27:25,246 [DEBUG] [spyder.plugins.completion.plugin] -> Responses statistics: {'fallback': 281, 'lsp': 0}
2019-10-17 15:27:25,265 [DEBUG] [urllib3.connectionpool] -> http://localhost:46624 "POST /clientapi/editor/complete HTTP/1.1" 200 1358
2019-10-17 15:27:25,266 [DEBUG] [spyder.plugins.completion.kite.providers.document] -> {'completions': [{'snippet': {'text': 'pyplot as plt', 'placeholders': []}, 'replace': {'begin': 37, 'end': 39}, 'display': 'pyplot as plt', 'documentation': {'text': 'Provides a MATLAB-like plotting framework.\n\npylab ( #matplotlib.pylab ) combines pyplot with numpy into a single namespace. This is convenient for interactive work, but for programming it is recommended that the namespaces be kept separate, e.g.:\n\nimport numpy as np import matplotlib.pyplot as plt x = np.arange(0, 5, 0.1); y = np.sin(x) plt.plot(x, y)'}, 'web_id': 'matplotlib.pyplot', 'local_id': 'python;;;;matplotlib.pyplot', 'hint': 'module', 'smart': False}, {'snippet': {'text': 'pylab as plt', 'placeholders': []}, 'replace': {'begin': 37, 'end': 39}, 'display': 'pylab as plt', 'documentation': {'text': 'This is a procedural interface to the matplotlib object-oriented plotting library.\n\nThe following plotting commands are provided; the majority have MATLAB ® [*] ( ##id2 ) analogs and similar arguments.\n\n_Plotting commands\n\nacorr - plot the autocorrelation function annotate - annotate something in the figure arrow - add an arrow to the axes axes - Create a new axes axhline - draw a horizontal line across axes axvline - draw a vertical line across axes axhspan - draw a horizontal bar across axes '}, 'web_id': 'matplotlib.pylab', 'local_id': 'python;;;;matplotlib.pylab', 'hint': 'module', 'smart': False}]}
2019-10-17 15:27:25,267 [DEBUG] [spyder.plugins.completion.plugin] -> Completion plugin: Request 74 Got response from kite
2019-10-17 15:27:25,745 [DEBUG] [spyder.plugins.completion.languageserver.client] -> python response: textDocument/publishDiagnostics
```

However, no completions are shown.

## Versions

* Spyder version: 4.0.0.dev0 (Commit: c2b7673b1)
* Python version: 3.7.4 64-bit
* Qt version: Qt 5.9.6
* PyQt version: PyQt5 5.9.2
* Operating System name/version: Darwin 18.7.0

### Dependencies

```
pylint >=0.25                :  2.4.2 (OK)
spyder_kernels >=1.5.0;<2.0.0:  1.6.0 (OK)
matplotlib >=2.0.0           :  None (NOK)
pandas >=0.13.1              :  None (NOK)
numpy >=1.7                  :  None (NOK)
scipy >=0.17.0               :  None (NOK)
```

its-dhung:
## Description of Changes

This PR changes the completions logic to allow multiple completions providers to return completions for any given request.

Previously, completions providers would "race" against one another such that only one was allowed to return completions to the user. There are a couple of downsides to this methodology.

1. In some cases, if the "wrong" completions provider returns first, the user is left with completions that are semantically incorrect. In the video below, notice how the LSP provider returns first but the completions it provides are semantically incorrect and would result in a bug.
lsp
2. The design does not allow for a future where there could be multiple completions providers that are used for different reasons. This inflexibility therefore greatly reduces the value users can get from their completions experience.

With this PR, the `CompletionManager` class now waits for all completions providers to return before returning them back to the user. In the video below, notice how both Kite and LSP completions are returned (though the LSP completions are semantically wrong).

kite+lsp

To prevent slow completions providers from keeping the user waiting, the `CompletionManager` institutes a timeout per completions request at which point, all available completions are gathered and sent back. This is controlled through a setting:

image

### Issue(s) Resolved

Fixes #10447 

### Affirmation

I certify the above statement is true and correct: @its-dhung

### DIFF
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -207,6 +207,7 @@ DEFAULTS = [
               'automatic_completions': True,
               'automatic_completions_after_chars': 3,
               'automatic_completions_after_ms': 300,
+              'completions_max_request_ms': 200,
               'completions_hint': True,
               'underline_errors': False,
               'highlight_current_line': True,
--- a/spyder/plugins/completion/languageserver/confpage.py
+++ b/spyder/plugins/completion/languageserver/confpage.py
@@ -723,6 +723,10 @@ class LanguageServerConfigPage(GeneralConfigPage):
             _("Show automatic completions after keyboard idle (ms):"), None,
             'automatic_completions_after_ms', min_=0, max_=5000, step=10,
             tip=_("Default is 300"), section='editor')
+        self.completions_max_request_ms = self.create_spinbox(
+            _("Time to wait for completions to return (ms):"), None,
+            'completions_max_request_ms', min_=10, max_=5000, step=10,
+            tip=_("Default is 200"), section='editor')
         code_snippets_box = newcb(_("Enable code snippets"), 'code_snippets')
 
         completion_layout = QGridLayout()
@@ -735,7 +739,11 @@ class LanguageServerConfigPage(GeneralConfigPage):
                                     3, 1)
         completion_layout.addWidget(self.completions_after_ms.plabel, 4, 0)
         completion_layout.addWidget(self.completions_after_ms.spinbox, 4, 1)
-        completion_layout.addWidget(code_snippets_box, 5, 0)
+        completion_layout.addWidget(self.completions_max_request_ms.plabel,
+                                    5, 0)
+        completion_layout.addWidget(self.completions_max_request_ms.spinbox,
+                                    5, 1)
+        completion_layout.addWidget(code_snippets_box, 6, 0)
         completion_layout.setColumnStretch(2, 5)
         completion_widget = QWidget()
         completion_widget.setLayout(completion_layout)
@@ -1164,6 +1172,8 @@ class LanguageServerConfigPage(GeneralConfigPage):
         self.completions_after_characters.plabel.setEnabled(state)
         self.completions_after_ms.spinbox.setEnabled(state)
         self.completions_after_ms.plabel.setEnabled(state)
+        self.completions_max_request_ms.spinbox.setEnabled(state)
+        self.completions_max_request_ms.plabel.setEnabled(state)
 
     def disable_tcp(self, state):
         if state == Qt.Checked:
@@ -1388,6 +1398,8 @@ class LanguageServerConfigPage(GeneralConfigPage):
                 'editor', 'automatic_completions_after_chars'),
             'set_automatic_completions_after_ms': (
                 'editor', 'automatic_completions_after_ms'),
+            'set_completions_max_request_ms': (
+                'editor', 'completions_max_request_ms'),
         }
         for editorstack in editor.editorstacks:
             for method_name, (sec, opt) in editor_method_sec_opts.items():
--- a/spyder/plugins/completion/languageserver/confpage.py
+++ b/spyder/plugins/completion/languageserver/confpage.py
@@ -725,8 +725,8 @@ class LanguageServerConfigPage(GeneralConfigPage):
             tip=_("Default is 300"), section='editor')
         self.completions_max_request_ms = self.create_spinbox(
             _("Time to wait for completions to return (ms):"), None,
-            'completions_max_request_ms', min_=10, max_=5000, step=10,
-            tip=_("Default is 200"), section='editor')
+            'completions_max_request_ms', min_=0, max_=5000, step=10,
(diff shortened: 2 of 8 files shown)


ccordoba12:
@andfoy, please review @its-dhung's work.

spyder/config/main.py
```diff
@@ -207,6 +207,7 @@
               'automatic_completions': True,
               'automatic_completions_after_chars': 3,
               'automatic_completions_after_ms': 300,
+              'completions_max_request_ms': 200,
```
andfoy: Sometimes Jedi takes long to complete some compiled packages such as numpy, which Kite returns faster. What happens if just LSP is working on such cases? I think that this value is small when LSP is the only provider available, besides, it could depend on the actual speed and load of the machine running Spyder.

its-dhung: > What happens if just LSP is working on such cases?

This setting acts as a ceiling on the amount of time completions requests are allowed to take. So if LSP is the only provider here, the request will time out and the user will see no completions.

> I think that this value is small when LSP is the only provider available

If we think that this default value isn't giving LSP enough time to return, then we can increase the default value to something like 500ms. In any case, the point of a setting is to let the user change it according to their personal tastes or based on the specs of their machine. The default value is just a best guess at what should be a reasonable number.

ccordoba12: @its-dhung, this is a solution we considered before with @andfoy but we found problematic because Jedi could take up to 3 or 4 seconds to return its first completion, if it involves a compiled library (e.g. Numpy or Pandas). Besides, it also takes a long time (2 secs) for big classes (e.g. our own MainWindow one).

Therefore, we think this solution is unfeasible because it'd rule out Jedi in several important cases.

its-dhung: I've included a change to allow the user to disable this functionality by setting the value to 0.

metalogical: I think the right solution based on the feedback thus far is:

1. Before the timeout, continue waiting if not all `WAIT_FOR` sources have returned.
2. After the timeout, return once any `WAIT_FOR` source has returned a non-empty response (the same as the current behavior).

So, if only LSP is enabled, and LSP takes longer than 200ms, then we'd continue waiting for LSP to return due to criterion 2.

Do you agree with this? @ccordoba12 @andfoy

metalogical: I'm going to open a separate PR, because this branch is on @its-dhung's personal GH, so I can't modify it.

metalogical:
@ccordoba12 @andfoy please close this PR in favor of #10509 (and review that one). Your concerns about the default value making the LSP-only experience poor should be mitigated there.

ccordoba12:
Thanks @metalogical!

PR #10502, opened 2019-10-23
PR: Completions performance improvements

### LINKED ISSUE #10457: Decreasing automatic_completions_after_ms setting causes lag when typing
## Problem Description
Changing the setting `automatic_completions_after_ms` to a low but reasonable (from user's point of view) number causes noticeable lag when editing code.

1. Change the `automatic_completions_after_ms` setting to 50ms:
image
2. Try editing a Python file e.g. type out:
```python
import numpy as np
```

__Expected__
As a user, I would expect minimal lag.

__Actual__
There is noticeable lag in between typing a character and it being inserted into the buffer. This happens regardless of what completions engines are enabled and the value of the setting `automatic_completions_after_chars`. It appears as though something is blocking the UI thread in between keystroke → completion request → buffer update in UI.

## Versions

* Spyder version: 4.0.0.dev0 (Commit: c2b7673b1)
* Python version: 3.7.4 64-bit
* Qt version: Qt 5.9.6
* PyQt version: PyQt5 5.9.2
* Operating System name/version: Darwin 18.7.0

### Dependencies

```
pylint >=0.25                :  2.4.2 (OK)
spyder_kernels >=1.5.0;<2.0.0:  1.6.0 (OK)
matplotlib >=2.0.0           :  None (NOK)
pandas >=0.13.1              :  None (NOK)
numpy >=1.7                  :  None (NOK)
scipy >=0.17.0               :  None (NOK)
```

metalogical:
## Description of Changes

I used cProfile & the [profilehooks](https://mg.pov.lt/profilehooks/) library to profile completions-related logic. The reason for the lag was all in the rendering logic, so I focused on reducing wasted work there. In particular some Qt functions are slow, so we should avoid calling them when necessary.

1. reduce calls to `hide_tooltip()`
2. reduce calls to `icon_manager.icon()`
3. use `''.join()` instead of string formatting with `str.format`
4. reduce calls to `setStylesheet`
5. use a fixed size hint for completions list items, because creating a `QTextDocument` and calling `idealWidth` every time is very slow.
    - I tested it on various screen sizes, and it seems to scale very well.
    - I used the size that I was seeing the old behavior typically return (it always returned the same size in my testing).
6. Add a separate delay on showing completion detail so that the user can configure them separately. This is important since rendering the completion hint widget is slow.

As a result, the lag is much less perceptible, but not entirely eliminated. I may have more ideas on how to improve this in the future, but this is definitely an incremental effort, so we can start with these low-hanging fruit.

Screen Shot 2019-10-25 at 2 40 06 PM

### Issue(s) Resolved

fixes #10457 

### Affirmation

### DIFF
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -208,7 +208,7 @@ DEFAULTS = [
               'automatic_completions_after_chars': 3,
               'automatic_completions_after_ms': 300,
               'completions_hint': True,
-              'completions_hint_after_ms': 1000,
+              'completions_hint_after_ms': 500,
               'underline_errors': False,
               'highlight_current_line': True,
               'highlight_current_cell': True,
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -208,6 +208,7 @@ DEFAULTS = [
               'automatic_completions_after_chars': 3,
               'automatic_completions_after_ms': 300,
               'completions_hint': True,
+              'completions_hint_after_ms': 1000,
               'underline_errors': False,
               'highlight_current_line': True,
               'highlight_current_cell': True,
--- a/spyder/plugins/editor/widgets/completion.py
+++ b/spyder/plugins/editor/widgets/completion.py
@@ -49,9 +49,7 @@ class CompletionWidget(QListWidget):
                      CompletionItemKind.FILE: 'filenew',
                      CompletionItemKind.REFERENCE: 'reference',
                      }
-
-    ICON_MAP = {typ: ima.icon(typ) for typ in set(ITEM_TYPE_MAP.values())}
-    ICON_MAP['no_match'] = ima.icon('no_match')
+    ICON_MAP = {}
 
     sig_show_completions = Signal(object)
 
@@ -198,6 +196,11 @@ class CompletionWidget(QListWidget):
         if self.count() == 0:
             self.hide()
 
+    def _get_cached_icon(self, name):
+        if name not in self.ICON_MAP:
+            self.ICON_MAP[name] = ima.icon(name)
+        return self.ICON_MAP[name]
+
     def set_item_display(self, item_widget, item_info, height, width):
         """Set item text & icons using the info available."""
         item_provider = item_info['provider']
@@ -221,8 +224,7 @@ class CompletionWidget(QListWidget):
             width=width)
 
         item_widget.setText(item_text)
-        if item_type in self.ICON_MAP:
-            item_widget.setIcon(self.ICON_MAP[item_type])
+        item_widget.setIcon(self._get_cached_icon(item_type))
 
     def get_html_item_representation(self, item_completion, item_type,
                                      icon_provider=None,
--- a/spyder/plugins/editor/widgets/completion.py
+++ b/spyder/plugins/editor/widgets/completion.py
@@ -24,48 +24,35 @@ from spyder.py3compat import to_text_string
 from spyder.widgets.helperwidgets import HTMLDelegate
 
 
-COMPLETION_ITEM_TEMPLATE = u"""
-<table width="{width}" height="{height}">
-    <tr>
-        <td valign="middle" style="color:{color}">
-            {completion}
-        </td>
-
-        <td valign="middle" align="right" float="right" style="color:{color}">
-            {type}
-        </td>
-        <td valign="top" align="right" float="right" width="{img_width}">
-            <img src="data:image/png;base64, {icon_provider}"
-                 height="{img_height}" width={img_width}/>
-        </td>
-    </tr>
-</table>
-"""
 DEFAULT_COMPLETION_ITEM_HEIGHT = 15
 DEFAULT_COMPLETION_ITEM_WIDTH = 250
 
 
 class CompletionWidget(QListWidget):
     """Completion list widget."""
-    ICONS_MAP = {CompletionItemKind.TEXT: 'text',
-                 CompletionItemKind.METHOD: 'method',
-                 CompletionItemKind.FUNCTION: 'function',
-                 CompletionItemKind.CONSTRUCTOR: 'constructor',
-                 CompletionItemKind.FIELD: 'field',
-                 CompletionItemKind.VARIABLE: 'variable',
-                 CompletionItemKind.CLASS: 'class',
-                 CompletionItemKind.INTERFACE: 'interface',
-                 CompletionItemKind.MODULE: 'module',
-                 CompletionItemKind.PROPERTY: 'property',
-                 CompletionItemKind.UNIT: 'unit',
-                 CompletionItemKind.VALUE: 'value',
-                 CompletionItemKind.ENUM: 'enum',
-                 CompletionItemKind.KEYWORD: 'keyword',
-                 CompletionItemKind.SNIPPET: 'snippet',
-                 CompletionItemKind.COLOR: 'color',
-                 CompletionItemKind.FILE: 'filenew',
-                 CompletionItemKind.REFERENCE: 'reference',
-                 }
+    ITEM_TYPE_MAP = {CompletionItemKind.TEXT: 'text',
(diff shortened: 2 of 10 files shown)


spyder/widgets/helperwidgets.py
```diff
@@ -152,13 +152,9 @@ def paint(self, painter, option, index):
...
-        options = QStyleOptionViewItem(option)
-        self.initStyleOption(options, index)
-
-        doc = QTextDocument()
-        doc.setHtml(options.text)
-
-        return QSize(round(doc.idealWidth()), round(doc.size().height() - 2))
+        # use a fixed size instead of computing an "ideal width"
+        # because it's more performant, and this is a reasonable size
+        return QSize(621, 22)
```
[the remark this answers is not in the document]
metalogical: ah yeah. This is the change I was least comfortable with making. I'll perhaps look into alternatives on this.

spyder/widgets/mixins.py
```diff
@@ -122,7 +122,13 @@ def _calculate_position(self, at_line=None, at_point=None):
 
         return point
 
-    def _update_stylesheet(self, widget):
+    def _update_stylesheet(self, widget, _updated=set()):
+        # Update the stylesheet for a given widget at most once
```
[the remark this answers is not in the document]
metalogical: This relies on a subtlety of Python semantics, which is that default values are evaluated when the function is defined (not when the function is called). So there is one value of `_updated`, which is modified below.

Since it confused you, exploiting this property is probably a bad idea. I'll change this to define the cache outside the function definition.

metalogical:
Thanks for the review. I like the suggestion to delay the completion detail widget. I'll take a stab at that and fix the other issues here.

spyder/widgets/mixins.py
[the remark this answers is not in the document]
dalthviz: Oh I didn't know about that, awesome, but yes please, or maybe you can add an explanation of the behavior used just to be sure that is clear for anyone what we are doing here

spyder/plugins/editor/widgets/completion.py
```diff
@@ -24,48 +24,35 @@
...
+                     CompletionItemKind.ENUM: 'enum',
+                     CompletionItemKind.KEYWORD: 'keyword',
+                     CompletionItemKind.SNIPPET: 'snippet',
+                     CompletionItemKind.COLOR: 'color',
+                     CompletionItemKind.FILE: 'filenew',
+                     CompletionItemKind.REFERENCE: 'reference',
+                     }
+
+    ICON_MAP = {typ: ima.icon(typ) for typ in set(ITEM_TYPE_MAP.values())}
+    ICON_MAP['no_match'] = ima.icon('no_match')
```
dalthviz: @metalogical @ccordoba12 @goanpeca checking through SSH why the tests on CircleCI are failing, seems like the above lines are some how triggering the error. I tested by changing the call from `ima.icon()` to an empty `QIcon` and then I'm able to run `reporterror.py` (I had to change both lines):

```python
    ICON_MAP = {typ: QIcon() for typ in set(ITEM_TYPE_MAP.values())}
    ICON_MAP['no_match'] = QIcon()
```

metalogical: instead of eagerly precaching these, I can just use a lazy cache.

ccordoba12: We can simply skip this file from CircleCI because there are no tests in the file itself (CircleCI runs very simple tests that we have at the end of each file).

I'll push a commit here to take care of that.

spyder/widgets/helperwidgets.py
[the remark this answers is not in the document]
metalogical: I just reverted this, since I can't figure out any better approach, and it's not as critical as the completions hint fix.

metalogical:
@dalthviz @ccordoba12 updated this PR (other than the CircleCI issue) based on comments.

spyder/plugins/editor/widgets/completion.py
ccordoba12: I understood this incorrectly, sorry. It's making to fail all files that use QtAwesome.

So @metalogical, please implement your suggestion above:

> instead of eagerly precaching these, I can just use a lazy cache.

to see if that fixes the problems seen in CircleCI.

spyder/plugins/editor/widgets/completion.py
```diff
@@ -248,16 +231,37 @@ def get_html_item_representation(self, item_completion, item_type,
...
-
-        display = html.escape(item_completion).replace(' ', '&nbsp;')
-        return COMPLETION_ITEM_TEMPLATE.format(completion=display,
-                                               type=item_type,
-                                               color=ima.MAIN_FG_COLOR,
-                                               icon_provider=icon_provider,
-                                               img_height=img_height,
-                                               img_width=img_width,
-                                               height=height,
-                                               width=width)
```
ccordoba12: Why not using a template here? It is far more readable and maintainable than using a list and joining its elements.

ccordoba12:
@goanpeca, please help us to review this one (as I already requested you before).

spyder/plugins/editor/widgets/completion.py
metalogical: `.format` is actually quite slow. If we were restricting ourselves to 3.7, then the new f-strings get precompiled into efficient code, but those aren't fast / available in older versions of Python.

spyder/config/main.py
```diff
@@ -208,6 +208,7 @@
               'automatic_completions_after_chars': 3,
               'automatic_completions_after_ms': 300,
               'completions_hint': True,
+              'completions_hint_after_ms': 1000,
```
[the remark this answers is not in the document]
metalogical: We can do 500, I just chose a number that felt comfortable to me.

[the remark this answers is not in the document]
ccordoba12: Yeah, I think 500 is better.

spyder/plugins/editor/widgets/completion.py
ccordoba12: @goanpeca, what do you think?

ccordoba12: @metalogical, what about the old string replacement?

ccordoba12:
@metalogical, please reset the last commit ("Kite Completion: Skip reporterror.py running on CircleCI") before proceeding because now it's an spurious one.

ccordoba12:
> @metalogical, please reset the last commit ("Kite Completion: Skip reporterror.py running on CircleCI") before proceeding because now it's an spurious one.

Ok, don't worry about this. @dalthviz just did it.

spyder/plugins/editor/widgets/completion.py
metalogical: I think old-style and new-style string formatting are comparable in performance -- they both need to make function calls and parse the format string every time you do the format operation.

f-strings are fast because Python compiles the syntax into efficient string processing logic. My approach is similarly faster than `.format` because it's similar to what the code would compile into, and doesn't require the parsing operation & function calls.

metalogical:
@ccordoba12 I fixed the build, and updated the default timeout value. The only outstanding piece is the string formatting discussion. Let me know what the final verdict on that is.

spyder/plugins/editor/widgets/completion.py
ccordoba12: Ok, I understand your reasoning now and see no other alternative for now.

I'd just like to ask you to add a comment above `parts = [` describing why we use here this approach instead of `.format` or string replacement.

spyder/widgets/helperwidgets.py
```diff
@@ -156,6 +156,7 @@ def sizeHint(self, option, index):
         self.initStyleOption(options, index)
 
         doc = QTextDocument()
+        doc.setDocumentMargin(self._margin)
```
goanpeca: @metalogical could you add a comment for what this does or why it was needed?

PR #10504, opened 2019-10-24
PR: Correctly handle Kite replacement range, fixing dict completions

metalogical:
## Description of Changes

Kite returns a replacement range with each completion indicating where to insert the completion within the buffer. This is particularly useful for more non-traditional completions, such as Kite's dict completions, where simply matching the preceding identifier is not enough. Kite's dict completions were previously broken in Spyder, but this change fixes them.

1. Support the `textEdit` functionality of the [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#textDocument_completion)
2. Properly respect the `filterText` when deciding whether completions are still valid.
3. Update the Kite provider to specify a `textEdit` instead of `insertText`, and also not use `filterText`, since Kite filters internally (and will always do so -- it is part of the API spec).

Should I add any unit tests for this? @andfoy @ccordoba12 

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/base.py
+++ b/spyder/plugins/editor/widgets/base.py
@@ -854,11 +854,20 @@ class TextEditBaseWidget(QPlainTextEdit, BaseEditMixin):
         """Completion list is active, Enter was just pressed"""
         self.completion_widget.item_selected()
 
-    def insert_completion(self, text, completion_position):
-        if text:
+    def insert_completion(self, completion, completion_position):
+        if not completion:
+            return
+        cursor = self.textCursor()
+        if isinstance(completion, dict):
+            cursor.setPosition(completion['textEdit']['range']['start'])
+            cursor.setPosition(completion['textEdit']['range']['end'],
+                               QTextCursor.KeepAnchor)
+            text = to_text_string(completion['textEdit']['newText'])
+        else:
+            text = to_text_string(completion)
+
             # Get word on the left of the cursor.
             result = self.get_current_word_and_position(completion=True)
-            cursor = self.textCursor()
             if result is not None:
                 current_text, start_position = result
                 end_position = start_position + len(current_text)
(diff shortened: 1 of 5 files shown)


metalogical:
@andfoy I have not tested with other LSP servers. I'm not sure which server uses the `textEdit` functionality instead of `insertText`, so it's a little difficult to test. Any suggestions?

metalogical:
I fixed a bug that was caught by tests where when on-the-fly completions are turned off, the completion filtering was broken:

1. use the old insert logic for the old-style `insertText` completions
2. remove `textEdit` completions from the completion list when doing in-editor filtering, since it's difficult to correctly update the replacement range.
    - so if on-the-fly completions are turned off, you'll see Kite completions on tab or ctrl-space, but they'll disappear when you continue typing.

metalogical:
@andfoy I looked into the unit tests, and it seems like we already have fairly good coverage on tests that check that the PyLS (`insertText`) completions continue to work. So this should at least be fairly safe on that front.

We don't have much coverage on Kite completions (with `textEdit`), but I'm not sure if there's any good way to test that right now with your infrastructure.

andfoy:
> We don't have much coverage on Kite completions (with textEdit), but I'm not sure if there's any good way to test that right now with your infrastructure.

Actually, right now we don't have any Kite tests, we can add them later on, so don't worry about that

goanpeca:
Hi @metalogical thanks for working on this! Looks awesome :-)

I have one question. I was expecting to get completions without the need of typing the string quotes

```
dic = {}
dic['one'] = 1
dic['two'] = 2

dic[*]
```

So if I have the cursor (*) and press `tab` I just get a tab space and not a completion. Is this expected behavior?

spyder/plugins/editor/widgets/completion.py
```diff
@@ -170,7 +170,7 @@ def show_list(self, completion_list, position, automatic):
         # signal used for testing
         self.sig_show_completions.emit(completion_list)
 
-    def update_list(self, filter_text):
+    def update_list(self, current_word, new=True):
```
goanpeca: @metalogical could you add more information in this docstring about `current_word` and `new` parameters?

spyder/plugins/editor/widgets/base.py
```diff
@@ -854,11 +854,23 @@ def select_completion_list(self):
         """Completion list is active, Enter was just pressed"""
         self.completion_widget.item_selected()
 
-    def insert_completion(self, text, completion_position):
-        if text:
+    def insert_completion(self, completion, completion_position):
```
goanpeca: Please add a docstring here

metalogical:
@goanpeca thanks for the review.

There is currently a deficiency in Kite where we cannot parse incomplete subscript expressions, which is why you don't see completions in the `dic[*]` case.

It's a bit tricky for us to fix this right now, so no word on when there will be a fix, but it's in our queue.

I addressed your other comments.

ccordoba12:
@metalogical, this has a conflict with master now (probably due to PR #10502). Please solve it so I can merge it.

metalogical:
resolved the conflict, just waiting on CI

goanpeca:
The following error is appearing @metalogical 

```python-traceback
completion = self.currentItem().data(Qt.UserRole)
AttributeError: 'NoneType' object has no attribute 'data'
```

metalogical:
@ccordoba12 @goanpeca this should be good to go

PR #10542, opened 2019-10-30
PR: Add startup lines to the debugger

### LINKED ISSUE #12795: Startup script for debugger
Is it possible to set one? I have functions useful for debugging that I placed in `Preferences > IPython console > Startup > Run a file`, but they're are undefined in `ipdb>`. Further, running debugger appears to _undo_ the startup script; I can access functions before running debugger, but not after (unless I restart kernel).

I figure this isn't as simple due to debugger namespace scoping, so I suppose the question's rather whether there's a way to define _global_ functions - but that IPython "forgets" the startup script after debugging remains unclear to me.

impact27:
## Description of Changes
Execute some code before every prompt in pdb. This is useful eg. to import matplotlib.
(I am tierd of `import matplotlib.pyplot as plt` on every pdb prompt.)

### Issue(s) Resolved

Fixes #12795

### Affirmation

impact27

video


impact27:
The macOS fail looks unrelated to this PR.

impact27:
I have tested this on my machine for a while and I must say it is quite great to be able to automatically import matplotlib while debugging. Otherwise I have to reimport it every time I change frames. I think this would be a great addition to Spyder 4 if you think there is still time to add this.

impact27:
Restarting the tests as the slow linux tests are failing for some reason...

ccordoba12:
@impact27, could you open a new PR and skip the failing test on Linux? That test is very simple, so I don't see a big problem on doing that.

PR #10549, opened 2019-10-30
PR: Add scrollbar to autosave dialog

### LINKED ISSUE #9125: Autosave recovery dialog expands beyond screen instead of using scrollbars with too many items
If I have more autosave files than will fit on my monitor (I have a 1200p one and just had it happen), then the dialog just expands outside of the visible screen (making it impossible to interact with the rest) rather than limiting its size and adding a scroll bar to the file list. Therefore, we should limit the size of the dialog, or the number of list items shown on screen accordingly to comfortably fit on a 720p monitor (or the current size, if practicable) and add a scroll bar widget to show the rest if that is the case. @jitseniesen take note, thanks!

juanis2112:
## Description of Changes

autosave

### Issue(s) Resolved

Fixes #9125

### Affirmation

Juanis2112
dalthviz

### DIFF
--- a/spyder/plugins/editor/utils/autosave.py
+++ b/spyder/plugins/editor/utils/autosave.py
@@ -18,7 +18,7 @@ import re
 from qtpy.QtCore import QTimer
 
 # Local imports
-from spyder.config.base import _, get_conf_path
+from spyder.config.base import _, get_conf_path, running_under_pytest
 from spyder.plugins.editor.widgets.autosaveerror import AutosaveErrorDialog
 from spyder.plugins.editor.widgets.recover import RecoveryDialog
 from spyder.utils.programs import is_spyder_process
@@ -174,7 +174,8 @@ class AutosaveForPlugin(object):
         the pid files.
         """
         files_to_recover, pidfiles = self.get_files_to_recover()
-        dialog = RecoveryDialog(files_to_recover, parent=self.editor.main)
+        parent = self.editor if running_under_pytest() else self.editor.main
+        dialog = RecoveryDialog(files_to_recover, parent=parent)
         dialog.exec_if_nonempty()
         self.recover_files_to_open = dialog.files_to_open[:]
         for pidfile in pidfiles:
--- a/spyder/plugins/editor/utils/autosave.py
+++ b/spyder/plugins/editor/utils/autosave.py
@@ -174,7 +174,7 @@ class AutosaveForPlugin(object):
         the pid files.
         """
         files_to_recover, pidfiles = self.get_files_to_recover()
-        dialog = RecoveryDialog(files_to_recover, parent=self.editor)
+        dialog = RecoveryDialog(files_to_recover, parent=self.editor.main)
         dialog.exec_if_nonempty()
         self.recover_files_to_open = dialog.files_to_open[:]
         for pidfile in pidfiles:
--- a/spyder/plugins/editor/widgets/recover.py
+++ b/spyder/plugins/editor/widgets/recover.py
@@ -16,10 +16,8 @@ import time
 from qtpy.compat import getsavefilename
 from qtpy.QtCore import Qt
 from qtpy.QtWidgets import (QApplication, QDialog, QDialogButtonBox,
-                            QGridLayout, QGroupBox, QHBoxLayout, QLabel,
-                            QMessageBox, QPushButton, QRadioButton, QScrollArea,
-                            QTableWidget, QSizePolicy,
-                            QVBoxLayout, QWidget)
+                            QHBoxLayout, QLabel, QMessageBox, QPushButton,
+                            QTableWidget, QVBoxLayout, QWidget)
 # Local imports
 from spyder.config.base import _, running_under_pytest
 
--- a/spyder/plugins/editor/widgets/recover.py
+++ b/spyder/plugins/editor/widgets/recover.py
@@ -15,8 +15,10 @@ import time
 # Third party imports
 from qtpy.compat import getsavefilename
 from qtpy.QtCore import Qt
-from qtpy.QtWidgets import (QDialog, QDialogButtonBox, QHBoxLayout, QLabel,
-                            QMessageBox, QPushButton, QTableWidget,
+from qtpy.QtWidgets import (QApplication, QDialog, QDialogButtonBox,
+                            QGridLayout, QGroupBox, QHBoxLayout, QLabel,
+                            QMessageBox, QPushButton, QRadioButton, QScrollArea,
+                            QTableWidget, QSizePolicy,
                             QVBoxLayout, QWidget)
 # Local imports
 from spyder.config.base import _, running_under_pytest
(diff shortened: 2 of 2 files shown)


goanpeca:
@juanis2112 remember to add a screenshot 📷

spyder/plugins/editor/utils/autosave.py
```diff
@@ -174,7 +174,8 @@ def try_recover_from_autosave(self):
         the pid files.
         """
         files_to_recover, pidfiles = self.get_files_to_recover()
-        dialog = RecoveryDialog(files_to_recover, parent=self.editor)
+        parent = self.editor if running_under_pytest() else self.editor.main
+        dialog = RecoveryDialog(files_to_recover, parent=parent)
```
jitseniesen: Out of curiosity, why this change?

dalthviz: The initial change was to use the `MainWindow` instead of the `Editor` to properly set the `styleSheet` for the Widget (before this the dialog always was shown using the light theme). However, when running the test suit we use a mock for the `MainWindow` that inherits from a 
`QObject`, and since the dialog needs a QWidget as parent, the tests suit was failing here. Any suggestions to handle this are greatly appreciated :)

spyder/plugins/editor/widgets/recover.py
```diff
@@ -99,6 +99,8 @@ def __init__(self, autosave_mapping, parent=None):
         self.add_table()
         self.add_cancel_button()
         self.setWindowTitle(_('Recover from autosave'))
+        self.setFixedSize(670, 400)
```
goanpeca: Any particular reason for this fixed size? should we make it resizable?

spyder/plugins/editor/widgets/recover.py
```diff
@@ -147,8 +149,8 @@ def add_table(self):
         table.setHorizontalHeaderLabels(labels)
         table.verticalHeader().hide()
 
-        table.setHorizontalScrollBarPolicy(Qt.ScrollBarAlwaysOff)
-        table.setVerticalScrollBarPolicy(Qt.ScrollBarAlwaysOff)
+        table.setHorizontalScrollBarPolicy(Qt.ScrollBarAsNeeded)
```
goanpeca: Should not this be off? by default?

spyder/plugins/editor/widgets/recover.py
dalthviz: Basically I use this fixed size to show all the table horizontally and to prevent the dialog to get too tall, but maybe I can set the width to the width of the table instead of a fixed number. About the resize I think @ccordoba12 told me to prevent it

spyder/plugins/editor/widgets/recover.py
dalthviz: Maybe if I set the width using the table width I can disable this, should I do it?

spyder/plugins/editor/widgets/recover.py
dalthviz: Basically I use this fixed size to show all the table horizontally and to prevent the dialog to get too tall, but maybe I can set the width to the width of the table instead of a fixed number. About the resize I think @ccordoba12 told me to prevent it

spyder/plugins/editor/widgets/recover.py
goanpeca: It is ok, we can merge as is!

spyder/plugins/editor/widgets/recover.py
goanpeca: It is ok, we can merge as is!

spyder/plugins/editor/widgets/recover.py
jitseniesen: > Maybe if I set the width using the table width I can disable this, should I do it?

The dialog can get arbitrarily wide if you have really long directory or file names, so I think it is dangerous to use the table width.

spyder/plugins/editor/utils/autosave.py
jitseniesen: I see, I had this problem that the parent should be a `QWidget` before; unfortunately, I don't know of a way around this, so I'm afraid I can't help you.

spyder/plugins/editor/widgets/recover.py
```diff
@@ -206,6 +202,13 @@ def add_cancel_button(self):
         button_box.rejected.connect(self.reject)
         self.layout.addWidget(button_box)
 
+    def center(self):
```
ccordoba12: @dalthviz, what happens when we try to center this and you have two monitors in your system?

dalthviz: The dialog is shown centered in the first screen

dalthviz: Maybe instead of `QApplication.desktop().screenGeometry(0)` we should use `QApplication.primaryScreen()` here @ccordoba12 ?

spyder/plugins/editor/utils/autosave.py
dalthviz: No problem, thanks for looking into this :+1:

spyder/plugins/editor/widgets/recover.py
ccordoba12: `primaryScreen` is the screen where Spyder was launched?

goanpeca: Probably not, it is what the system defines as "primary" so to know which screen the widget lives (The QMainWindow for example) we could look into:

https://stackoverflow.com/questions/49167204/find-screen-a-qwidget-is-located-on

goanpeca: Probably not, it is what the system defines as "primary" so to know which screen the widget lives (The QMainWindow for example) we could look into:

https://stackoverflow.com/questions/49167204/find-screen-a-qwidget-is-located-on

dalthviz: At the end this resulted in using the window settings from the config files. This is a preview:

auto

spyder/app/mainwindow.py
```diff
@@ -1522,6 +1522,19 @@ def load_window_settings(self, prefix, default=False, section='main'):
         return hexstate, window_size, prefs_dialog_size, pos, is_maximized, \
                is_fullscreen
 
+    def get_window_center(self, load=False):
```
ccordoba12: @dalthviz, I think you're assuming here that the window was maximized the previous time it was closed, which is normally the case. But what would happen if it was closed with half the size of the screen?

dalthviz: It will appear centered relative to the last position of the MainWindow:

autod

ccordoba12: Mmmh, I don't like this too much. How is the splash screen correctly centered?

dalthviz: The splash screen is always centered relative to the first screen (it doesn't depend on where the MainWindow will appear)

ccordoba12: Then let's do the same for the Autosave dialog. I think it's the best we can do.

PR #10553, opened 2019-10-31
PR: Allow to report segmentation faults to Github

### LINKED ISSUE #10536: Send segmentation fault reports to Github
When an internal python error happens in spyder, a report can be sent automatically by the user to github. I think it would be a good idea to extend this behaviour to C errors such as segmentation faults.

Steps:
 [ ] Define a file where the crash data is saved
 [ ] Have spyder check if the file is empty when starting. If it is not, show a message like "It appears Spyder crashed last time you openned it. Do you want to send a report on Github?"
 [ ] When launching spyder, open a file handler to the log file, and pass it to `faulthandler.enable`

I am currently using this:
```
from spyder.app import start
import faulthandler
with open('/Users/quentinpeter/spyder_faulthandler.log', 'w') as f:
    faulthandler.enable(file=f)
    start.main()
```

I think this code should be moved to `main()`

impact27:
Selección_003

## Description of Changes

### Issue(s) Resolved

Fixes #10536

### Affirmation


goanpeca:
This is really great @impact27 !

ccordoba12:
I also think this is really nice, thanks @impact27! However, I think this needs more UI/UX work to be ready.

In my opinion, asking for a "Title" and "Steps to reproduce" of a problem when the app crashed is really different than when it just generated an internal traceback. In the second case the app is still up, so you can inspect what you were doing and try to understand what generated the error. Whereas in the first case you don't have that possibility, which makes reporting errors in the way our dialog is designed right now really hard for users.

So I'm afraid users would start to simply fill garbage when encountering crashes in order to get rid of the dialog. That has happened for internal errors too, but I observed it in a very small amount of cases. In other words, most users try to give us some feedback, even if minimal or vague. But this could change for crashes, given that in this case we're asking too much from our users. 

Therefore, I think we need to do the following in order to finish this:

* Remove the first dialog (i.e. the one that starts with "Spyder crashed during last session"). That dialog is too verbose and it's not clear enough what it is about. Instead, we should show directly the familiar error report dialog to our users.
* But we need to make some modifications to it, according to what I said above:
  - The text on the top needs to talk about a crash (not an internal error) and it doesn't need to mention that users need to check the Troubleshooting guide because that's not so helpful in this case.
  - "Title" and "Steps to reproduce" need to be hidden, but internally we'll set the title of the issue to be submitted to Github to something like "Spyder crash" because requires that info.
  - The checkbox "Hide all future errors during this session" needs to be hidden too because it doesn't make sense to show it in this case either.
  - "Submit to Github" needs to be active as soon as the dialog is shown.

@goanpeca, what do you think?

goanpeca:
I agree with your suggestions @ccordoba12 !

impact27:
Screenshot 2019-11-02 at 09 30 25

impact27:
> The checkbox "Hide all future errors during this session" needs to be hidden too because it doesn't make sense to show it in this case either.

I disagree. This means: don't bother me with other errors during this session.

PR #10567, opened 2019-11-01
PR: Add button and keyboard shortcut to refresh the variable explorer during execution

### LINKED ISSUE #4398: Variable explorer is not updated during code execution

All this apply for the qtconsole only (not the standard python).
The variable explorer completely freezes during code execution since spyder 3.
As highlighted in #3523 it's no more possible to see updates, but even worse it's not possible to explore variables at all.

#### Half-working
1. type `l = [[1, 2, 3], [4, 5, 6]]`
2. open 'l' in the variable explorer.
3. start : 
```python
import time
for _ in range(1000):
    time.sleep(0.5)
```
4. look into the nested list 'l' in the variable explorer.

Note that any change you make from the variable explorer is discarded if you interrupt the console execution.

#### Not-working
1. type `l = [[1, 2, 3], [4, 5, 6]]`
2. start : 
```python
import time
for _ in range(1000):
    time.sleep(0.5)
```
3. double click on 'l' in the variable explorer : nothing happens.

I would expect to be able to interact with the variable explorer even during a computation (and better, with up-to-date values), as in spyder 2 or the python console.

* Spyder Version: 3.1.4
* Python Version: 3.4
* Qt Version: 5.2.1
* PyQt Version: 5.2.1
* Operating system: Linux

## Dependencies

and paste the contents below:

sphinx >=0.6.6   :  1.5.5 (OK)
sympy >=0.7.3    :  1.0 (OK)

## Possible solutions
I've tried to find a solution but I don't have the necessary knowledges about Qt and IPython kernels.
Here is a review of what I've found (perhaps more on the possibility to have auto-refresh back).

Starting from [this nice enhancement proposal for Ipython](https://github.com/ipython/ipython/issues/9354) (that would actually be great to have as option in spyder), I've [found](https://gist.github.com/takluyver/8c25dcf1f175159b0c741207a6dda94a) this way  to interact with running python program (linux specific) : 

1. run %connect_info and find the kernel in htop
2. launch in the qtconsole
```python
import signal
import time

def handle_sigint(signo, frame):
    print(globals())

signal.signal(signal.SIGUSR1, handle_sigint)

for a in range(10000):
    time.sleep(0.5)
    print(a)
```
3. from htop send SIGUSR1 to the kernel
4. look at what happened in the console.

Now I wonder how to take advantage of it to be able to refresh/interact with the variable explorer during a computation. 
Maybe for the refresh it would be sufficient to give the possibility to the console to call refresh_namespacebrowser() via get_ipython().*, make the signal connection at the initialisation of the console and send periodically SIGUSR1 to it. Also we should find a way to make the same thing only with IPython if we want to have this cross-platform.

Another idea is to make a queue of spyder instruction to the kernel and make some handler called when it receive SIGUSR1.

Maybe using get_ipython().events could help but I don't understand everything with that. Here seems to be the [doc](http://ipython.readthedocs.io/en/stable/config/callbacks.html).

For scientific computations we often don't set up tests, have dirty code and long execution time. Hence it was really nice in spyder 2 to be able without any effort (or just a 'global') to see early anomalies during execution. Maybe you should keep the python console until we have a solution for the qtconsole.
I'm willing to help but really need help to go further.

impact27:
refresh

## Description of Changes

### Issue(s) Resolved

Fixes #4398

### Affirmation

impact27

### DIFF
--- a/spyder/plugins/variableexplorer/widgets/namespacebrowser.py
+++ b/spyder/plugins/variableexplorer/widgets/namespacebrowser.py
@@ -215,7 +215,7 @@ class NamespaceBrowser(QWidget):
 
         self.refresh_button = create_toolbutton(
             self, text=_("Refresh variables"),
-            icon=ima.icon('redo'),
+            icon=ima.icon('refresh'),
             triggered=self.refresh_table)
         config_shortcut(self.refresh_table,
                         context='variable_explorer',
--- a/spyder/plugins/variableexplorer/widgets/namespacebrowser.py
+++ b/spyder/plugins/variableexplorer/widgets/namespacebrowser.py
@@ -216,7 +216,7 @@ class NamespaceBrowser(QWidget):
         self.refresh_button = create_toolbutton(
             self, text=_("Refresh variables"),
             icon=ima.icon('redo'),
-            toggled=self.refresh_table)
+            triggered=self.refresh_table)
         config_shortcut(self.refresh_table,
                         context='variable_explorer',
                         name='refresh', parent=self)
--- a/spyder/plugins/variableexplorer/widgets/namespacebrowser.py
+++ b/spyder/plugins/variableexplorer/widgets/namespacebrowser.py
@@ -213,7 +213,6 @@ class NamespaceBrowser(QWidget):
                         context='variable_explorer',
                         name='search', parent=self)
 
-
         self.refresh_button = create_toolbutton(
             self, text=_("Refresh variables"),
             icon=ima.icon('redo'),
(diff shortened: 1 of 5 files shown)


spyder/plugins/variableexplorer/widgets/namespacebrowser.py
```diff
@@ -213,8 +213,17 @@ def setup_toolbar(self):
                         context='variable_explorer',
                         name='search', parent=self)
 
+        self.refresh_button = create_toolbutton(
+            self, text=_("Refresh variables"),
+            icon=ima.icon('redo'),
```
ccordoba12: Please add a new entry for this in the `ima` module and use `mdi.refresh` for it.

ccordoba12: And please name the entry `refresh`.

PR #10582, opened 2019-11-02
PR: Add results in batches and limit results on Find pane to a maximum

### LINKED ISSUE #10264: Find in files freezes spyder
### Issue Report Checklist

    * [x] Restarted Spyder

## Problem Description
Find in files will freeze spyder (alsmost completely not resonsive) if there is too many results.

The profiles tells me that most of the time is spent in `spyder.plugins.findinfiles.ItemDelegate.sizeHint`

1. Change cwd to `/usr/local/lib/python3.7/site-packages` (or whatever the site-packages folder is)
2. Search for `in`
3. Try to use spyder while the search takes place.

Spyder becomes almost completely unresponsive.

```python-traceback

PASTE TRACEBACK HERE

```

## Versions

* Spyder version: 
* Python version: 
* Qt version: 
* PyQt version: 

### Dependencies

```
PASTE DEPENDENCIES HERE
```

goanpeca:
## Description of Changes

finder

- The find in files now adds the files in batches that increase over time in power of 2
2^1, 2^2 etc.... up to a max of 9 (512 items per load batch)
- The maximum is now configurable, default is 1000
- The stop button and search button now act as "one single button"
- Enter now toggles between searching and stoping (instead of always restart a search.
- The regex section ins hidden by default

### Issue(s) Resolved

Fixes #10264

### Affirmation

### DIFF
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -288,8 +288,9 @@ DEFAULTS = [
               'search_text_regexp': False,
               'search_text': [''],
               'search_text_samples': [TASKS_PATTERN],
-              'more_options': True,
+              'more_options': False,
               'case_sensitive': False,
+              'max_results': 1000,
               }),
             ('breakpoints',
              {
@@ -516,10 +517,6 @@ DEFAULTS = [
               'show_installation_dialog': True,
               'show_onboarding': True,
              }),
-            ('find_in_files',
-             {
-              'max_results': 1000,
-              }),
             ]
 
 
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -611,4 +611,4 @@ NAME_MAP = {
 #    or if you want to *rename* options, then you need to do a MAJOR update in
 #    version, e.g. from 3.0.0 to 4.0.0
 # 3. You don't need to touch this value if you're just adding a new option
-CONF_VERSION = '53.1.0'
+CONF_VERSION = '53.2.0'
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -516,6 +516,10 @@ DEFAULTS = [
               'show_installation_dialog': True,
               'show_onboarding': True,
              }),
+            ('find_in_files',
+             {
+              'max_results': 1000,
+              }),
             ]
 
 
--- a/spyder/plugins/findinfiles/plugin.py
+++ b/spyder/plugins/findinfiles/plugin.py
@@ -53,7 +53,7 @@ class FindInFiles(SpyderPluginWidget):
         case_sensitive = self.get_option('case_sensitive')
         path_history = self.get_option('path_history', [])
         search_in_index = self.get_option('search_in_index', default=0)
-        max_results = self.get_option('max_results', default=1000)
+        max_results = self.get_option('max_results')
 
         self.findinfiles = FindInFilesWidget(
             self,
--- a/spyder/plugins/findinfiles/plugin.py
+++ b/spyder/plugins/findinfiles/plugin.py
@@ -4,7 +4,7 @@
 # Licensed under the terms of the MIT License
 # (see spyder/__init__.py for details)
 
-"""Find in Files Plugin"""
+"""Find in Files Plugin."""
 
 # pylint: disable=C0103
 # pylint: disable=R0903
@@ -17,7 +17,7 @@ import sys
 # Third party imports
 from qtpy.QtCore import Signal, Qt
 from qtpy.QtGui import QKeySequence
-from qtpy.QtWidgets import QApplication, QVBoxLayout
+from qtpy.QtWidgets import QApplication, QInputDialog, QVBoxLayout
 
 # Local imports
 from spyder.api.plugins import SpyderPluginWidget
@@ -42,7 +42,7 @@ class FindInFiles(SpyderPluginWidget):
         supported_encodings = self.get_option('supported_encodings')
         self.search_text_samples = self.get_option('search_text_samples')
         search_text = self.get_option('search_text')
-        search_text = [txt for txt in search_text \
+        search_text = [txt for txt in search_text
                        if txt not in self.search_text_samples]
         search_text += self.search_text_samples
         search_text_regexp = self.get_option('search_text_regexp')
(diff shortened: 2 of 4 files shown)


goanpeca:
@impact27 what do you think a reasonable limit is?

@ccordoba12 thoughts on this?

ccordoba12:
> a maximum limit to results after which a message is displayed and the search is stopped

If the changes you've made here prevent Spyder to freeze, why do we need to stop searching at some point and show a message?

goanpeca:
> If the changes you've made here prevent Spyder to freeze,

They do not prevent it, it just makes it less terrible. But as the list keeps growing and growing it will still become unusable so we really need to set a hard limit.

impact27:
> > If the changes you've made here prevent Spyder to freeze,
> 
> It prevents the deadlock freeze, but as the list keeps growing Spyder is not very responsive since every update of the list (even in batch) blocks the GUI (as you can see on the animation, the spinner stops for a moment every time the batch is loaded into the tree).
> 
> Thoughts @impact27 could you test this?

Is it possible to do part of the processing in a thread? I guess only GUI operations need to be done in the main thread. What part of the processing actually takes time?

impact27:
> @impact27 what do you think a reasonable limit is?

I am not sure I would want to go through more than 1000 results. This could be a setting that is initially at 1000.

impact27:
For reference, searchin for `in` took 15 min:
 - 10.5 min in `spyder/plugins/findinfiles/widgets.py::ItemDelegate::sizeHint`:
    - 6 min in `idealWidth`
    - 3.5 min in `setHtml`
 - 3 min in `spyder/plugins/findinfiles/widgets.py::FindInFilesWidget::search_complete`:
    - 3 min in `expandAll`
 - 1.5 min in `spyder/plugins/findinfiles/widgets.py:: ResultsBrowser::append_result`

goanpeca:
@impact27 

>For reference, searching for in took 15 min:

Are these values with this PR? or with master?

impact27:
Oups you are right somehow I didn’t checkout the correct branch. So these values are for the master.

goanpeca:
>Is it possible to do part of the processing in a thread? I guess only GUI operations need to be done in the main thread. What part of the processing actually takes time?

The search is donde in a thread, I will try to move everything except the actual loading of the tree and see how it looks.

goanpeca:
Could you provide some numbers for this branch :-) ?

impact27:
I’ll try to do this tomorrow

impact27:
The interface is unresposnsive for ~10s (which is much better) I ran 3 searches:
 - 18 sec in `spyder/plugins/findinfiles/widgets.py:: ResultsBrowser::append_result`:
    - 10 sec in `__init__::__repr__::get_font` (in spyder/config/gui.py)
 - 5 sec in `paintEvent` (spyder/plugins/editor/widgets/codeeditor.py)

impact27:
Maybe `get_font` could be cached?

goanpeca:
@impact27 I incorporated your suggestion and added some extra stuff. Please see the Animation on the description.

If you could test this, it would be great!

ccordoba12:
> Please see the Animation on the description.

Things look pretty slick now @goanpeca! Thanks a lot for working on this!

goanpeca:
I still need to add 1 test , working on it now.

goanpeca:
This one is ready @ccordoba12 and @impact27 for review and final testing!

impact27:
This totally solves the issue for me.

spyder/plugins/findinfiles/plugin.py
```diff
@@ -53,24 +53,30 @@ def __init__(self, parent=None):
         case_sensitive = self.get_option('case_sensitive')
         path_history = self.get_option('path_history', [])
         search_in_index = self.get_option('search_in_index', default=0)
+        max_results = self.get_option('max_results', default=1000)
```
ccordoba12: Please add this option to our DEFAULTS dict in `config/main.py` and remove the `default=1000` from here.

spyder/plugins/findinfiles/widgets.py
```diff
@@ -181,40 +196,157 @@ def find_string_in_file(self, fname):
...
                             if found > -1:
                                 break
+
         except IOError as xxx_todo_changeme:
             (_errno, _strerror) = xxx_todo_changeme.args
             self.error_flag = _("permission denied errors were encountered")
+
         self.completed = True
 
+    def process_results(self):
```
ccordoba12: Please add a docstring to this method to understand what it does.

spyder/plugins/findinfiles/widgets.py
```diff
@@ -181,40 +196,157 @@ def find_string_in_file(self, fname):
...
+            if nb_files > 1:
+                text_files += 's'
+            text = "%d %s %d %s" % (num_matches, text_matches,
+                                    nb_files, text_files)
+        title = title + text
+
+        self.partial_results = []
+        self.sig_line_match.emit(items, title)
+
+    def truncate_result(self, line, start, end):
```
ccordoba12: Please add a docstring to this too.

spyder/plugins/findinfiles/plugin.py
goanpeca: Ok

spyder/plugins/findinfiles/widgets.py
goanpeca: Done

spyder/plugins/findinfiles/widgets.py
goanpeca: Done

spyder/config/main.py
```diff
@@ -516,6 +516,10 @@
               'show_installation_dialog': True,
               'show_onboarding': True,
              }),
+            ('find_in_files',
+             {
+              'max_results': 1000,
+              }),
```
ccordoba12: `find_in_files` already has its own section, so please add this entry there.

Also, please change the `more_options` of `find_in_files` to False, to hide the exclusion options by default from our config.

goanpeca: Fixed (I actually had added the option in the wrong place)

spyder/config/main.py
```diff
@@ -288,8 +288,9 @@
               'search_text_regexp': False,
               'search_text': [''],
               'search_text_samples': [TASKS_PATTERN],
-              'more_options': True,
+              'more_options': False,
```
ccordoba12: You forgot to bump CONF_VERSION to account for this change.

goanpeca: 🤦‍♂

goanpeca: Fixed

jnsebgosselin:
Thanks a lot @goanpeca for working on this! Find in files is one of my favorite feature in Spyder actually, so I'm glad it got an improvement.

Would you guys mind if I open a PR to reduce a little bit the vertical space between the items, because I think they are a little bit too spaced now after this PR. 

image

goanpeca:
>Would you guys mind if I open a PR to reduce a little bit the vertical space between the items, because I think they are a little bit too spaced now after this PR.

@jnsebgosselin that had to be done like that for performance reasons. We can set the same height for all items which avoids recalculating heights everytime. For big searches the time spend  on sizeHint is too much.

jnsebgosselin:
Cool! I hope you find a workaround for this, thanks.

goanpeca:
>Cool! I hope you find a workaround for this, thanks.

@jnsebgosselin so I tried and this method is a bottleneck, so Spyder becomes completely unresponsive if the amount of results is big (this needs a better definition).

So the workaround is still calling for every item even if I cache the heigh values with the first entries.

The only workaround would be to us my "hack" for maximum results bellow... 1000, but for anything above that value, go back to fixed heights.

Thoughts @ccordoba12 ?

jnsebgosselin:
Ok thanks for looking into it. I will try to take a look at it in case I can find a solution.

ccordoba12:
> The only workaround would be to us my "hack" for maximum results bellow... 1000, but for anything above that value, go back to fixed heights.

Sounds good to me.

jnsebgosselin:
> > The only workaround would be to us my "hack" for maximum results bellow... 1000, but for anything above that value, go back to fixed heights.
> 
> Sounds good to me.

Ok, I've looked into this a little bit and yeah... considering the amount of data involved sometimes, I think the only solution to fix this would be to use a QTreeView instead of a QTreeWidget, but that may require a lot of work...

So like @ccordoba12, I agree with the workaround you are proposing too.

jnsebgosselin:
What if we go with a GUI design similar to the file switcher? That would solve the variable lines height issue while not altering the performances. Also, this doesn't look bad at all...

image

goanpeca:
Fine by me. Thoughts @ccordoba12 ?

ccordoba12:
I like it too. Please go ahead with that idea @jnsebgosselin!

PR #10605, opened 2019-11-04
PR: Filter completions with an empty value for the 'insertText' key

dalthviz:
## Description of Changes

* [ ] Wrote at least one-line docstrings (for any new functions)
* [ ] Added unit test(s) covering the changes (if testable)

* [ ] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

Filter completion list to prevent completions with an empty 'insertText' value.

Seems like the LSP is returning this as a completion:

```python
{'label': '',
'kind': 9,
'detail': 'cv2.cv2',
'documentation': 'cv2.cv2()\n\nPython wrapper for OpenCV.',
'sortText': (1, 'a'),
'insertText': '',
'filterText': '',
'insertTextFormat': 1,
'provider': 'LSP'}
```

### Issue(s) Resolved

Fixes #10593 

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct: dalthviz

PR #10630, opened 2019-11-06
PR: Fix completions bug in Kite introduced by PR 10605

metalogical:
## Description of Changes

Fixes exception raised due to #10605 (described [here](https://github.com/spyder-ide/spyder/pull/10605#issuecomment-550529046))

How can we make sure we don't regress again? Maybe we need fixturized test cases with sample completions dicts that we return to the completions logic and ensure it works properly with `textEdit`s. Thoughts?

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/tests/conftest.py
+++ b/spyder/plugins/editor/widgets/tests/conftest.py
@@ -139,7 +139,6 @@ def mock_completions_codeeditor(qtbot_module, request):
     editor.show()
 
     mock_response = Mock()
-    mock_response.side_effect = lambda *args: None
 
     def perform_request(lang, method, params):
         resp = mock_response(lang, method, params)
--- a/spyder/plugins/editor/widgets/tests/conftest.py
+++ b/spyder/plugins/editor/widgets/tests/conftest.py
@@ -127,18 +127,22 @@ def lsp_plugin(qtbot_module, request):
 
 
 @pytest.fixture
-def mock_codeeditor(qtbot_module, request):
-    """CodeEditor instance with ability to mock the completions response"""
+def mock_completions_codeeditor(qtbot_module, request):
+    """CodeEditor instance with ability to mock the completions response.
+
+    Returns a tuple of (editor, mock_response). Tests using this fixture should
+    set `mock_response.side_effect = lambda lang, method, params: {}`.
+    """
     # Create a CodeEditor instance
     editor = codeeditor_factory()
     qtbot_module.addWidget(editor)
     editor.show()
 
-    mock = Mock()
-    mock.side_effect = lambda *args: None
+    mock_response = Mock()
+    mock_response.side_effect = lambda *args: None
 
     def perform_request(lang, method, params):
-        resp = mock(lang, method, params)
+        resp = mock_response(lang, method, params)
         print("DEBUG {}".format(resp))
         if resp is not None:
             editor.handle_response(method, resp)
@@ -154,7 +158,7 @@ def mock_codeeditor(qtbot_module, request):
         editor.completion_widget.hide()
     request.addfinalizer(teardown)
 
-    return editor, mock
+    return editor, mock_response
 
 
 @pytest.fixture
--- a/spyder/plugins/editor/widgets/tests/conftest.py
+++ b/spyder/plugins/editor/widgets/tests/conftest.py
@@ -127,6 +127,37 @@ def lsp_plugin(qtbot_module, request):
 
 
 @pytest.fixture
+def mock_codeeditor(qtbot_module, request):
+    """CodeEditor instance with ability to mock the completions response"""
+    # Create a CodeEditor instance
+    editor = codeeditor_factory()
+    qtbot_module.addWidget(editor)
+    editor.show()
+
+    mock = Mock()
+    mock.side_effect = lambda *args: None
+
+    def perform_request(lang, method, params):
+        resp = mock(lang, method, params)
+        print("DEBUG {}".format(resp))
+        if resp is not None:
+            editor.handle_response(method, resp)
+    editor.sig_perform_completion_request.connect(perform_request)
+
+    editor.filename = 'test.py'
+    editor.language = 'Python'
+    editor.completions_available = True
+    qtbot_module.wait(2000)
+
+    def teardown():
+        editor.hide()
+        editor.completion_widget.hide()
+    request.addfinalizer(teardown)
+
+    return editor, mock
+
+
+@pytest.fixture
 def lsp_codeeditor(lsp_plugin, qtbot_module, request):
     """CodeEditor instance with LSP services activated."""
     # Create a CodeEditor instance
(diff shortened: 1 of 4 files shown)


metalogical:
I'm currently trying to add a test that mocks the completions response in various ways.

ccordoba12:
> I'm currently trying to add a test that mocks the completions response in various ways.

That sounds very nice, thanks! However, we'd prefer integration tests, so if you could add tests that call Kite directly, that would be much better.

metalogical:
@ccordoba12 I added a unit test (and fixtures) that allow mocking the completions responses, and verify that a simple dict key completion example with `textEdit` works correctly.

I agree that integration tests would be nice, but I'm not sure what the best approach to get that working is. We have a `kited_test` binary that we use in CI for other editor plugins, which disables some excess logic (such as authentication, etc), and also exposes some additional endpoints useful for testing.

However, we'd need to set up a way to retrieve the this binary from the web or somewhere else. I think it's worth investing in (and we'd be willing to spend resources here whenever we have them), but would prefer to tackle this as a separate issue.

spyder/plugins/editor/widgets/tests/conftest.py
```diff
@@ -126,6 +126,37 @@ def teardown():
     return completions
 
 
+@pytest.fixture
+def mock_codeeditor(qtbot_module, request):
```
ccordoba12: Let's rename this to `mock_completions_codeeditor` because `mock_codeeditor` is too broad.

spyder/plugins/editor/widgets/tests/test_introspection.py
```diff
@@ -764,6 +767,51 @@ def test_fallback_completions(fallback_codeeditor, qtbot):
     code_editor.toggle_code_snippets(True)
 
 
+@pytest.mark.slow
+@pytest.mark.first
+@flaky(max_runs=5)
+def test_kite_textEdit_completions(mock_codeeditor, qtbot):
+    """Test on-the-fly completions."""
```
ccordoba12: Please fix this docstring.

spyder/plugins/editor/widgets/tests/test_introspection.py
```diff
@@ -764,6 +767,51 @@ def test_fallback_completions(fallback_codeeditor, qtbot):
...
+
+    code_editor.toggle_automatic_completions(False)
+    code_editor.toggle_code_snippets(False)
+
+    # Set cursor to start
+    code_editor.go_to_line(1)
+
+    qtbot.keyClicks(code_editor, 'my_dict.')
+
+    # Complete f -> from
```
ccordoba12: Is this comment right?

spyder/plugins/editor/widgets/tests/test_introspection.py
```diff
@@ -764,6 +767,51 @@ def test_fallback_completions(fallback_codeeditor, qtbot):
...
+    code_editor, mock = mock_codeeditor
+    completion = code_editor.completion_widget
+
+    code_editor.toggle_automatic_completions(False)
+    code_editor.toggle_code_snippets(False)
+
+    # Set cursor to start
+    code_editor.go_to_line(1)
+
+    qtbot.keyClicks(code_editor, 'my_dict.')
```
ccordoba12: Why the `.` at end of `my_dict` is necessary here?

spyder/plugins/editor/widgets/tests/conftest.py
```diff
@@ -126,6 +126,37 @@ def teardown():
...
+    """CodeEditor instance with ability to mock the completions response"""
+    # Create a CodeEditor instance
+    editor = codeeditor_factory()
+    qtbot_module.addWidget(editor)
+    editor.show()
+
+    mock = Mock()
+    mock.side_effect = lambda *args: None
```
ccordoba12: Please add a comment above this line saying that `mock.side_effect` needs to be redefined in tests that depend on this fixture

spyder/plugins/editor/widgets/tests/conftest.py
```diff
@@ -126,6 +126,37 @@ def teardown():
...
 
+@pytest.fixture
+def mock_codeeditor(qtbot_module, request):
+    """CodeEditor instance with ability to mock the completions response"""
+    # Create a CodeEditor instance
+    editor = codeeditor_factory()
+    qtbot_module.addWidget(editor)
+    editor.show()
+
+    mock = Mock()
```
ccordoba12: Let's rename this to `mock_response`.

spyder/plugins/editor/widgets/tests/test_introspection.py
metalogical: The reason I use this test case is because this is a type of completion that Kite returns (it allows users to quickly access dict keys with a `.`), and it's a case where the completion `["dict-key"]` replacing the `.` can't be represented with just `insertText`.

metalogical:
Thanks for the look-over @ccordoba12 

I addressed your comments.

spyder/plugins/editor/widgets/tests/test_introspection.py
```diff
@@ -764,6 +767,55 @@ def test_fallback_completions(fallback_codeeditor, qtbot):
     code_editor.toggle_code_snippets(True)
 
 
+@pytest.mark.slow
+@pytest.mark.first
+@flaky(max_runs=5)
+def test_kite_textEdit_completions(mock_codeeditor, qtbot):
```
ccordoba12: You forgot to change this to `mock_completions_codeeditor`.

spyder/plugins/editor/widgets/tests/test_introspection.py
```diff
@@ -764,6 +767,55 @@ def test_fallback_completions(fallback_codeeditor, qtbot):
...
+    """Test textEdit completions such as those returned by the Kite provider.
+
+    This mocks out the completions response, and does not test the Kite
+    provider directly.
+    """
+    code_editor, mock = mock_codeeditor
```
ccordoba12: Let's rename here `mock` to `mock_responsoe` to go with the changes done in the fixture.

spyder/plugins/editor/widgets/tests/test_introspection.py
ccordoba12: Ok, I didn't know that and it's pretty cool!

spyder/plugins/editor/widgets/tests/test_introspection.py
metalogical: 🤦‍♂

andfoy:
Just for the record, a contributor is migrating pyls completions to use textEdit: https://github.com/palantir/python-language-server/pull/690

metalogical:
@andfoy interesting

Note that currently Spyder's support for `textEdit` completions is very limited: if the user keeps typing, it filters out all such completions, due to the fact that the replacement range may be out-of-date. I didn't want to add a lot of complexity, so I just depend on completions being re-requested.

If `pyls` starts returning those, it may be wise to invest further in expanding support for those completions. Especially since re-requesting completions with `pyls` may be slow.

PR #10631, opened 2019-11-06
PR: Don't repeatedly request Kite onboarding file

metalogical:
## Description of Changes

Currently, we send a request many times, due to many events triggering that code path.

This isn't a huge deal, but every time the file is requested, a new file is written (to a temp directory). So it's better to avoid sending a new request whenever possible.

Not tested, but not that critical, so it's not worth the time.

### Affirmation

### DIFF
--- a/spyder/plugins/completion/kite/plugin.py
+++ b/spyder/plugins/completion/kite/plugin.py
@@ -175,8 +175,6 @@ class KiteCompletionPlugin(SpyderCompletionPlugin):
             return
         if not self.available_languages:
             return
-        if not self._show_onboarding:
-            return
         # Don't send another request until this request fails.
         self._show_onboarding = False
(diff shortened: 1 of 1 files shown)


spyder/plugins/completion/kite/plugin.py
```diff
@@ -167,8 +167,19 @@ def _kite_onboarding(self):
...
+        if not self.enabled:
+            return
+        if not self._show_onboarding:
+            return
+        if self.main.is_setting_up:
+            return
+        if not self.available_languages:
+            return
+        if not self._show_onboarding:
+            return
```
ccordoba12: This condition is repeated, i.e. `if not self._show_onboarding` is also above.

metalogical: good catch! fixed

PR #10483, opened 2019-10-22
PF: Fix kite logo for OSX

goanpeca:
#### Before
Screen Shot 2019-10-21 at 21 58 03

#### After
Screen Shot 2019-10-21 at 21 55 57

* [x] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

The kite logo was being scaled as a pixmap and then passed to the icon which would distrot the icon for scale factor that did not adjust to specific ratios. I edited the Kite logo images (in SVG) so that the image includes a padding, like most of the Font Awesome icons. I based the size to 16 and left a padding of 1, which results in a kite logo of 14px (effective) height. This removes the need to pass a custom scale factor, and since the icon is used as is, it does not get distorted.

As an extra for this PR I removed the padding o the layout between icons and text which results in a better UI and reduced status widgets total width.

### Issue(s) Resolved

Fixes #10442 

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct: @goanpeca

PR #10644, opened 2019-11-08
PR: Prevent status bar widget blinking during startup

jnsebgosselin:
This is to fix some widget blinking that occurs during the startup of Spyder after PR #10483

blinking

### Affirmation

### DIFF
--- a/spyder/widgets/status.py
+++ b/spyder/widgets/status.py
@@ -44,6 +44,14 @@ class StatusBarWidget(QWidget):
         self.label_icon = QLabel()
         self.label_value = QLabel()
 
+        # Layout setup
+        layout = QHBoxLayout(self)
+        layout.setSpacing(0)  # Reduce space between icon and label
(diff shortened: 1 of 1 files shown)


spyder/widgets/status.py
```diff
@@ -52,20 +60,6 @@ def __init__(self, parent, statusbar, icon=None):
         self.label_value.setAlignment(Qt.AlignRight)
         self.label_value.setFont(self.text_font)
 
-        # Layout
-        layout = QHBoxLayout()
-        layout.setSpacing(0)  # Reduce space between icon and label
-        layout.addWidget(self.label_icon)
-        self.label_icon.setVisible(icon is not None)
```
ccordoba12: Why are you removing this condition?

jnsebgosselin: Because this condition is already applied in `set_icon`

PR #10715, opened 2019-11-15
PR: Add restart option to PluginConfigPage

steff456:
## Description of Changes

This change allows to restart spyder from any configuration page that inherits from PluginConfigPage. 

### Issue(s) Resolved

This change is necessary for PR https://github.com/spyder-ide/spyder-terminal/pull/156 to work

### Affirmation

### DIFF
--- a/spyder/api/preferences.py
+++ b/spyder/api/preferences.py
@@ -8,8 +8,10 @@
 API to create an entry in Spyder Preferences dialog associated to a
 given plugin.
 """
-
+# Third-party imports
+from qtpy.QtWidgets import QMessageBox
 # Local imports
+from spyder.config.base import _
 from spyder.preferences.configdialog import SpyderConfigPage
 
 class PluginConfigPage(SpyderConfigPage):
@@ -17,6 +19,7 @@ class PluginConfigPage(SpyderConfigPage):
 
     def __init__(self, plugin, parent):
         self.plugin = plugin
+        self.main = parent.main
         self.get_option = plugin.get_option
         self.set_option = plugin.set_option
         self.get_font = plugin.get_font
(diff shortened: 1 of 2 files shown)


spyder/api/preferences.py
```diff
@@ -8,15 +8,18 @@
 API to create an entry in Spyder Preferences dialog associated to a
 given plugin.
 """
-
+# Third-party imports
```
ccordoba12: Please add a blank line above this line.

spyder/api/preferences.py
```diff
@@ -8,15 +8,18 @@
...
+from qtpy.QtWidgets import QMessageBox
```
ccordoba12: Please add a blank line below this line.

spyder/api/preferences.py
```diff
@@ -56,3 +59,32 @@ def setup_page(self):
         self.setLayout(layout)
         """
         raise NotImplementedError
+
+    def prompt_restart_required(self):
```
ccordoba12: This is not the right approach @steff456. We're repeating here the same method present in `GeneralConfigPage`. Since that class and `PluginConfigPage` both inherit from `SpyderConfigPage`, let's move this method to that class instead, so the child classes have access to it.

spyder/api/preferences.py
```diff
@@ -56,3 +59,32 @@ def setup_page(self):
...
+            msg_options += u"<li>{0}</li>".format(option)
+
+        msg_title = _("Information")
+        msg = u"{0}<ul>{1}</ul><br>{2}".format(msg_start, msg_options, msg_end)
+        answer = QMessageBox.information(self, msg_title, msg,
+                                         QMessageBox.Yes | QMessageBox.No)
+        if answer == QMessageBox.Yes:
+            self.restart()
+
+    def restart(self):
```
ccordoba12: You also need to move this method to `SpyderConfigPage`.

PR #10725, opened 2019-11-15
PR: Add a clock to the status bar

### LINKED ISSUE #10712: Add a clock to the statusbar
## Description
I usually work in a fullscreen mode, and I find having a clock on the statusbar help the workflow.

Screenshot from 2019-11-14 13-35-41_edited

I am willing to open a PR for that if you find it a useful widget to add.

rotemgb:
## Description of Changes

Added a clock to the statusbar, and an option to show or hide clock in the status bar preferences.

### Issue(s) Resolved

Fixes #10712 

### Affirmation

I certify the above statement is true and correct: rotemgb


spyder/config/main.py
```diff
@@ -83,6 +83,8 @@
               'memory_usage/timeout': 2000,
               'cpu_usage/enable': False,
               'cpu_usage/timeout': 2000,
+              'clock/enable': True,
```
ccordoba12: I think this should be `False` by default because it's something useful only when working in fullscreen mode, as you pointed out.

spyder/config/main.py
```diff
@@ -83,6 +83,8 @@
               'memory_usage/timeout': 2000,
               'cpu_usage/enable': False,
               'cpu_usage/timeout': 2000,
+              'clock/enable': True,
+              'clock/timeout': 2000,
```
ccordoba12: Perhaps this should be updated every minute? In other words, this value should be `60000` instead of `2000`?

spyder/config/main.py
goanpeca: Agreed!

spyder/config/main.py
goanpeca: This depends on when in time the timer starts so the only "issue" here is that if we put a larger value (>2000) then there might be situations in which the clock on Spyder and the clock of the operating system (taskbar etc..) will be "off" by 1 minute. 

So I think we should probably have 1000 to avoid this issue.

Thoughts @ccordoba12 ?

spyder/widgets/status.py
```diff
@@ -284,6 +284,28 @@ def update_interpreter(self, interpreter):
         self.update_tooltip()
 
 
+class ClockStatus(BaseTimerStatus):
+    """"Add clock to statusbar in a fullscreen mode"""
```
goanpeca: Please add a period. Docstrings are sentences:

`""""Add clock to statusbar in a fullscreen mode."""`

spyder/widgets/status.py
```diff
@@ -284,6 +284,28 @@ def update_interpreter(self, interpreter):
...
+
+    def import_test(self):
+        pass
+
+    def get_value(self):
+        """Return the time"""
```
goanpeca: Same here.

spyder/config/main.py
ccordoba12: > So I think we should probably have 1000 to avoid this issue.

Ok, agreed.

spyder/config/main.py
rotemgb: I agree!

PR #10732, opened 2019-11-16
PR: Add delay to dot completion logic

### LINKED ISSUE #10634: 'Show completions on the fly' cannot be turned off
## Problem Description

Completions on the fly kicks in even when the option is unchecked.

bug_auto_completion

1. Uncheck the option `Show completions on the fly`
2. Type some code that includes a '.'
3. Wait for completions to kicks in.

The problem seems to come from the code here :

https://github.com/spyder-ide/spyder/blob/90df471cff7d2713d20ebce900d27ede7b11e6f5/spyder/plugins/editor/widgets/codeeditor.py#L3602-L3610

It results in a call to `self.do_completion` directly even when the option `Show completions on the fly` is unchecked. Adding `and self.automatic_completions` to the condition `elif text in self.auto_completion_characters:` seems to fix the problem.

## Versions

* Spyder 4.0.0.dev0 (Commit: 3fd0553b4) 
* Python 3.7.3 64-bit | Qt 5.9.6 | PyQt5 5.9.2 | Windows 10

goanpeca:
## Description of Changes

### Issue(s) Resolved

Fixes #10634

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -3476,10 +3476,17 @@ class CodeEditor(TextEditBaseWidget):
         else:
             return super(CodeEditor, self).event(event)
 
+    def _start_completion_timer(self):
+        """Helper to start timer or complete."""
+        if self.automatic_completions_after_ms > 0:
+            self._timer_autocomplete.start(
+                self.automatic_completions_after_ms)
+        else:
+            self._handle_completions()
+
     def keyPressEvent(self, event):
         """Reimplement Qt method."""
-        if self.automatic_completions_after_ms > 0:
-            self._timer_autocomplete.start(self.automatic_completions_after_ms)
+        self._start_completion_timer()
 
         if self.completions_hint_after_ms > 0:
             self._completions_hint_idle = False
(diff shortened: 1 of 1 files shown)


goanpeca:
@jnsebgosselin could you test this?

@ccordoba12 if '.' completion now is treated as a on the fly completion. Should `(` for calltip display be treated as that as well?

ccordoba12:
> Should ( for calltip display be treated as that as well?

What do you mean?

goanpeca:
```python

import numpy as np

np.array(*   # * is the cursor
```

After typing the `(` a calltip hint will appear. Is this considered on the fly completion as well?

ccordoba12:
> Is this considered on the fly completion as well?

No, because it doesn't correspond to completing certain text in the editor. So I think you shouldn't consider that case in this PR.

goanpeca:
>No, because it doesn't correspond to completing certain text in the editor. So I think you shouldn't consider that case in this PR.

In that case, is the `.` dot (the subject of this pr) considered as text for completion on the fly?

ccordoba12:
> In that case, is the . dot (the subject of this pr) considered as text for completion on the fly?

Well, since the dot calls our completion widget, I'd say it breaks users expectations about the options they have set for that.

Issue #10627, opened 2019-11-06
Variable explorer shortcuts not updated after configuration change


goanpeca:
The variable explorer copy, search shortcuts are not being updated on configuration change, but only on restart

<img width="781" alt="Screen Shot 2019-11-06 at 12 26 38" src="https://user-images.githubusercontent.com/3627835/68322043-b62e4e00-0090-11ea-94ff-6bbef8942b33.png">

PR #10734, opened 2019-11-16
PR: Add option to exclude callables and modules in the Variable Explorer

### LINKED ISSUE #8715: Add filters to exclude modules and functions in Variable Explorer
## Problem Description

In the 'Variable Explorer', the objects of classes I've created only appear if I uncheck the filter 'Exclude unsupported data types'. But, at the same time, if I uncheck it, all the imported modules and functions unnecessarily appear (at least for me).

For example, in the image below, no variables were shown before unchecking this option. But it's a great tool for my specific application to see the variable mesh (type Mesh) in the Explorer.

image

## Desired Solution
I'd really appreciate if it's added in the following versions of Spyder more options in the Variable Explorer Filters, like:

- Exclude modules
- Exclude functions
- Exclude builtin methods

## Versions
* Spyder version: 3.3.2
* Python version: 3.7.1
* Operating System name/version: Windows10
* Dependencies: All OK!

ccordoba12:
This will allow us to only exclude these kind of objects by default in Spyder 4, so users can have access to the Object Explorer integrated by @dalthviz.

Depends on spyder-ide/spyder-kernels#183
Fixes #8715.

TODO:

- [x] Change default value of "Exclude unsupported types" to `False`.
- [x] Check why options in the Options menu of the Variable Explorer are not changing their values correctly.

### DIFF
--- a/continuous_integration/azure/install.bat
+++ b/continuous_integration/azure/install.bat
@@ -34,7 +34,7 @@ if %USE_CONDA% == yes (
 )
 
 :: Install spyder-kernels from master
-pip install -q --no-deps git+https://github.com/spyder-ide/spyder-kernels.git@refs/pull/183/head
+pip install -q --no-deps git+https://github.com/spyder-ide/spyder-kernels
 if errorlevel 1 exit 1
 
 :: Install python-language-server from master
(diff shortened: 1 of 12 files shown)


ccordoba12:
@goanpeca, I think you reported an issue about the second TODO above, right?

goanpeca:
@ccordoba12 this https://github.com/spyder-ide/spyder/issues/10627 ?

PR #10738, opened 2019-11-16
PR: Automatic completion fix

### LINKED ISSUE #10634: 'Show completions on the fly' cannot be turned off

impact27:
## Description of Changes

### Issue(s) Resolved

Fixes #10634
(I guess it solves it in a slightly different way than https://github.com/spyder-ide/spyder/pull/10732)

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -3608,7 +3608,8 @@ class CodeEditor(TextEditBaseWidget):
             # redefine this basic action which should have been implemented
             # natively
             self.stdkey_end(shift, ctrl)
-        elif text in self.auto_completion_characters:
+        elif (CONF.get('editor', 'automatic_completions')
+                  and text in self.auto_completion_characters):
             self.insert_text(text)
             if text == ".":
                 if not self.in_comment_or_string():


ccordoba12:
What do you think @goanpeca about this one?

goanpeca:
Hi @impact27 this solution is not taking into account that there should be a delay in the timer when the user presses `.` so this solution only solves the issue of not showing the compltion when disabled, but when enabled there should be a delay.

https://github.com/spyder-ide/spyder/pull/10732 does solve both cases

ccordoba12:
Closing then.

PR #10751, opened 2019-11-18
PR: Make Pylint plugin to use cwd as working dir

### LINKED ISSUE #5345: Static Code Analysis Not Reading .pylintrc file
I have a .pylintrc file in the Spyder project directory and in the file directory (it is located in a subdirectory). It has the following line in the [MASTER] block:

extension-pkg-whitelist=numpy

It also has the following lines in the [TYPECHECK] block:
ignored-modules=numpy
ignored-classes=optparse.Values,thread._local,_thread._local,numpy

However, the code analysis still gives me an error on the second line of these two:
time1 = df.index.to_datetime()
time2 = np.array([time1.year, time1.month, time1.day,
                          time1.hour, time1.minute,
                          time1.second, time1.microsecond/1000]).transpose()

The error is [E1101] Instance of 'list' has no 'transpose' member

Is there another directory that Spyder's call of Pylint must be reading from? How do I check this? Also, my project is on a shared network drive (temporarily). 

* Spyder Version: 3.2.3
* Python Version: Anaconda 3.6
* Qt Version: 5.6.2
* PyQt Version: 5.6.0
* Operating system: Windows 7

tovogt:
## Description of Changes

I changed the behavior of the static code analysis to set the working directory to the current project's root directory before executing pylint. This makes sure that project-specific pylint settings apply recursively to all files in a project.

### Issue(s) Resolved

Fixes #5345 

### Affirmation

I certify the above statement is true and correct: tovogt


CAM-Gerlach:
@ccordoba12 @goanpeca Is there a reason this hasn't been reviewed?

spyder/plugins/pylint/widgets/pylintgui.py
```diff
@@ -317,7 +317,7 @@ def start(self):
 
         self.process = QProcess(self)
         self.process.setProcessChannelMode(QProcess.SeparateChannels)
-        self.process.setWorkingDirectory(osp.dirname(filename))
+        self.process.setWorkingDirectory(getcwd_or_home())
```
ccordoba12: @tovogt, you said:

> I changed the behavior of the static code analysis to set the working directory to the current project's root directory before executing pylint

but this change doesn't implement that.

@dalthviz, please add here the project's directory (if there's one active) or `getcwd_or_home` in case there isn't.

CAM-Gerlach: Pylint should run in the working directory, not the project directory, for several reasons. 

First, having it use the working directory is consistent with the options in the run configuration for running the code normally, and if and when the pylint plugin is taught to pick up this config (as we've planned in the past but never got time to implement) this would be the most seamless transition here.

Second, the user is most likely to expect pylint will use the same working directory, and `.pylintrc` thereof, that they actually have set in Spyder, not revert back to their project directory. Of course, if they haven't manually changed the working directory to something different, then it will simply fall back to their project directory anyway. 

Finally, if the `.pylintrc` is in the working directory, not the project directory, Spyder won't find it, and the user won't have any easy way to fix this without adding an extra option and UI or keeping duplicate `.pylintrc` files in both (which is the current situation now). For example, I'll often use one Spyder project for multiple closely related subprojects, or my Spyder project directory will be one level above the actual git repo directory where the `.pylintrc` lives. If Spyder simply set pylint to use the working directory, this would just work, or at most I'd have to set my working directory appropriately. If pylint was locked to the project directory instead of the working dir, I either have to make an extra copy of the `.pylintrc` in the project directory, create a whole new project just for running pylint, or not use Spyder's projects feature at all.

tovogt: > @tovogt, you said (...) but this change doesn't implement that.
Yes, that's true. Sorry for that, I don't remember what made me think that this would return the project's directory. Instead, we should use `self.parentWidget().main.projects.get_active_project_path()` or something similar.

ccordoba12: I understand your reasons @CAM-Gerlach. But what if you want to use a single `.pylintrc`, the one in your home directory, for all files? Or one `.pylintrc` per project?

To solve this conundrum and provide a more sensible user experience, I propose the following:

* First we should check if there's a `.pylintrc` in the current working directory. If that's the case, then we should set the working directory of the Pylint process to that one.
* Then we should check if there's a `.pylintrc` in the project directory and do the same.
* Finally we should use the home directory for the process.

@tovogt, what do you think?

CAM-Gerlach: I would actually prefer that and was going to suggest that approach, but I thought you'd think that was too complicated :)

Pylint actually is supposed to do something like that itself; per [the pylint docs](https://docs.pylint.org/en/stable/user_guide/run.html#command-line-options) its search order is as follows (which gets even more complicated with the latest dev version, adding `pyproject.toml` and `setup.cfg` right below the working directory):

> 1. pylintrc in the current working directory
> 2. .pylintrc in the current working directory
> 3. If the current working directory is in a Python module, Pylint searches up the hierarchy of Python modules until it finds a pylintrc file. This allows you to specify coding standards on a module-by-module basis. Of course, a directory is judged to be a Python module if it contains an __init__.py file.
> 4. The file named by environment variable PYLINTRC
> 5. If you have a home directory which isn't /root:
>      1.  pylintrc or .pylintrc in your home directory
>      2.  .config/pylintrc in your home directory
> 6. /etc/pylintrc

I confirmed all this works normally when running pylint outside of Spyder (or running `pylint.lin.Run(FILENAME)` in the Internal Console), and I understand why inside of Spyder it doesn't find 1. and 2. since Spyder sets the working dir to be the of the file, and why it does find it in 3. (which is what I use now as a workaround). However, what I haven't figured out is why neither 4 or 5 work when running pylint inside Spyder, while they work just fine running it externally, I couldn't find anything obvious in Spyder's invocation of pylint that made it ignore 4 and 5. but not 3. while calling [`pylint.config.find_pylintrc()`](https://github.com/PyCQA/pylint/blob/e169e83e52fedc6624235d45e8e8737294a0fedf/pylint/config.py#L90) from the internal console did find everything, so I'd have to look more carefully at how Spyder's calling it.  

In order to inject the project directory into the existing pylint search path without ignoring the whole thing or re-implementing it ourselves, ensure that the most local `.pylintrc` to the analyzed file is used (as intended in the pylint docs), and get around whatever is causing it not to work correctly, I suggest doing something like the following, which I've tested in the internal console and seems to work:

```python
SEARCH_PATHS = [FILE_DIR, working_dir, PROJECT_DIR, USER_HOME_DIR]
def get_pylintrc_path(search_paths=SEARCH_PATHS):
    current_cwd = os.getcwd()
    try:
        def _find_pylintrc_path(path):
            os.chdir(path)
            return pylint.config.find_pylintrc()
        pylintrc_paths = [_find_pylintrc_path(path) for path in search_paths]
        for pylintrc_path in pylintrc_paths:
            if pylintrc_path != pylintrc_paths[-1]:
                break
        return pylintrc_path
    finally:
        os.chdir(current_cwd)
````

Then just add `'--rcfile="{}"'.format(get_pylintrc_path())` to the arguments to the pylint invocation.

This ensures that a `.pylintrc` with module-specific settings is preferred first, as is the intent of 3. (and as the user would get if running pylint as `pylint PATH_TO_FILE`, which is what the Spyder pane effectively emulates), then the Spyder working directory, then the project directory, and finally 4-6 of the above hierarchy, and if and when pylint changes/adds to their search path logic (as they will in the next version), Spyder will pick up those changes automatically so its behavior most closely matches user expectations.

What would be really nice is calling pylint with the interpreter of the working environment, not Spyder's, so it accurately reflects the environment the code is running in (Python version, packages installed, etc) but that's out of scope and and would require that pylint be installed in the working environment, e.g. making it a dep of spyder-kernels.

Comment (CAM-Gerlach) [reply spyder/plugins/pylint/widgets/pylintgui.py]: As for the working directory, that presumably should be set as Spyder's current working directory (i.e. as it is in this PR now) for consistency with anything else Pylint uses it for, though it won't be used directly to pick up the correct config file since we're passing that explicitly. I also updated the code above into a fully working implementation, including adding the pylintrc path to the invocation, getting the necessary paths, and accounting for a `.pylintrc` not existing anywhere.

Comment (ccordoba12) [reply spyder/plugins/pylint/widgets/pylintgui.py]: @tovogt, please try to implement CAM's suggested changes as soon as you can so we can include your work in our next release (4.1.0). Else you'll have to wait until 4.2.0, to be released in a couple of months.

Comment (CAM-Gerlach) [reply spyder/plugins/pylint/widgets/pylintgui.py]: @tovogt Also, you'll want to add a few tests for this, which you can do in a new file `test_pylint.py` under `spyder/plugins/pylint/tests`.

For unit test, you can test that `get_pylintrc_path` returns the expected path to the `.pylintrc file` when it is placed in a few different locations (you'll want to use Pytest's parameterize decorator to a fixture that adds and removes the .pylintrc files, and mocks `getcwd_or_home()` and `self.parentWidget().main.projects.get_active_project_path()` to return appropriate paths to automate this).

I also suggest modifying the [integration test in mainwindow](https://github.com/spyder-ide/spyder/blob/524428b494c5d4595142c7390ec88b9731e8f5ce/spyder/app/tests/test_mainwindow.py#L1933) to take the same parameterized fixture that tests with and without `.pylintrc` files present in one or more of the locations to ensure the settings are actually honored. For your `.pylintrc` files make sure to disable one or more of the messages found in the current example code file used for the tests, and make sure they each disable different messages so you can check that the right one is being applied.

Ideally, you could add both, but if its too much for you one should suffice (preferably the integration test I'd think). You also don't technically _have_ to parameterize the fixture, you could just have it straight-up just add `.pylintrc` config files to a couple of the locations and ensure the number and type of messages in the reported results is correct which would basically only require adding the fixture and tweaking the existing test/checks.

Comment (CAM-Gerlach) [reply spyder/plugins/pylint/widgets/pylintgui.py]: And if you don't feel comfortable doing that in the next couple days, @dalthviz or I could step in and take care of it instead so the change you helped prompt makes it into Spyder 4.1.0, just let us know.

Comment (tovogt) [reply spyder/plugins/pylint/widgets/pylintgui.py]: Okay, thanks for your instructions. I'm planing to look into this and taking care of the implementation in a couple of hours.

Comment (tovogt) [reply spyder/plugins/pylint/widgets/pylintgui.py]: The implementation is ready. Still struggling to setup a working test environment though ...

Comment (tovogt) [reply spyder/plugins/pylint/widgets/pylintgui.py]: I implemented a fixture that would be useful in the implementation of the tests, but I was only able to implement the test for `get_pylintrc_path`. For the integration test in mainwindow, I don't know how to setup my own environment in which I could run that test.

@CAM-Gerlach Can you implement the integration tests, please? The test cases are already included in the fixture `pylint_test_setup`. It defines output messages that should and shouldn't appear.

Comment (CAM-Gerlach) [reply spyder/plugins/pylint/widgets/pylintgui.py]: @tovogt Sorry for the delay in replying. I can give it a crack shortly and submit a PR to your branch, but with regard to getting a test environment set up, did you just follow the instructions in the [contributing guide](https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md)? In a nutshell:

```bash
conda create -n spyder-test-env python=3.7
conda activate spyder-test-env
<cd to the root of your local spyder repo>
conda install -c spyder-ide --yes --file requirements/conda.txt
conda install -c spyder-ide --yes --file requirements/test.txt
python runtests.py
```

I ran the test suite just now and though my deps are somewhat out of date since I haven't been an active Spyder developer for almost a year, everything ran (minus the slow tests which are skipped by default) with only 3 individual test failures (on Windows) that are probably either random or have to do with dependency issues. Is there something in particular that's giving you trouble? Can you give us more information so we can address it?

In compiling this advice, I did find some issues passing options to the `runtests.py` script that manifested itself when specifically trying to run the slow tests in the `test_mainwindow.py` file, which I've resolved in PR #11704, but other than that everything should work

tovogt: Ah okay, now it works. I misinterpreted the error messages. Moreover, I was confused that the command line `python runtests.py --run-slow -- -k test_run_static_code_analysis` didn't work as expected. Now I was able to add the integration test, as well. Thanks for taking the time!

CAM-Gerlach: > Ah okay, now it works. I misinterpreted the error messages.

As a general note, that is why you should always, always, always state the _precise_, _complete_ error message(s) you received when reporting an error rather than simply a generic statement like "I couldn't set up a working test environment", as it will save both you and us a huge amount of unnecessary time, effort and heartache.

> Now I was able to add the integration test, as well.

Welp, that's a thing. I've been working on it myself since around 10 PM, and now that its 4:30 am and now I'm finally done you post this. Unfortunately, the basic ideas were sound (you're clearly more knowledgeable, motivated and able to tackle tough problems than the average contributor) there were a large range of issues with the implementation at least as of the version as of when I pull it to start work on mine (which I would otherwise leave as review comments, but simply implemented myself since you requested that I complete the integration tests), such that I've basically reimplemented the tests from scratch with numerous enhancements over the implementation here, and I also made a substantial number of improvements to my initial design for the code itself. However, since you then force-pushed your branch and made a number of further changes, it is essentially impossible for me to merge those unless you retrieve and roll back to the right overwritten commit from your reflog, I push my branch and submit a PR to yours, you merge that, and then @ccordoba12 review and merge this.

As such, while I hate to do this as I advocated with @ccordoba12 for allowing you time to finish it yourself given the impending release deadline, our best option at this point is for me to simply submit a PR of the branch containing your original comments and my follow-ons, and close this one in favor of that. You'll still get commit credit and everything since your commits are still in there, it'll just make sure the changes get merged in as soon as possible to make the next release without a lot more complications. That okay with you?

tovogt: Sorry I caused you trouble. But if you like your own implementation better anyways, it was worth it, right? I only decided to do this because I was interested in learning from it. I don't care at all about credit being given or not. Just do whatever is best for the progress of the spyder project. If that means, discarding my PR for yours, that's fine with me.

tovogt: Note that `main.projects.get_active_project_path()` might return `None` in practice if there is no active project. This case is not covered in your tests and actually your code will break if `_find_pylintrc_path(path)` is called with `path is None`.

CAM-Gerlach:
@ccordoba12 @tovogt Per discussion above, opened a successor PR #11708 . Once that is successfully merged, this one can be presumably be closed.

spyder/plugins/pylint/widgets/pylintgui.py
CAM-Gerlach: > Sorry I caused you trouble.

You didn't cause me trouble, I caused me trouble by being way too through and perfectionist about my tests.

> But if you like your own implementation better anyways, it was worth it, right?

Considering the relatively minimal gain in the bigger picture from the more extensive tests relative to the many other higher-priority things on my plate right now, I wouldn't say so. But I did refresh my memory of a lot of pytest stuff that I haven't used for a year since I was a Spyder developer, which hopefully will be useful in the near future.

> If that means discarding my PR in favor of yours, that's fine with me.

It still has all your changes up to your last one, it just adds my revisions and integration test instead of yours. I was going to just submit it as a PR to your branch but after what happened it ended up being easier this way.

> Note that main.projects.get_active_project_path() might return None in practice if there is no active project

Thanks. As you noticed, it is both fixed in the code and covered by the tests now. I was indeed aware this was likely to happen and was already working on a fix and a test by that time, I just wanted to ensure I could repro the expected error interactively, write an appropriate test case separate from the others that xfailed, and then fix it.

ccordoba12:
> Per discussion above, opened a successor PR #11708 . Once that is successfully merged, this one can be presumably be closed.

Closing now. Let's continue the discussion there.

Issue #10786, opened 2019-11-21
Alt+Space shortcut cannot be set in Spyder 4


gerazov:
Moved to Spyder 4 RC2 - awesome work guys! The dark theme is super slick!

I miss something from previouos Spyders  though - I cannot set the `Alt+Space` shortcut anymore. I had this for delete line (and have it in all my other editors). 

So if it's fixable I would appreciate if you can bring it back : )

ccordoba12:
Ok, we'll take a look at it in the future.

jnsebgosselin:
Please see https://github.com/spyder-ide/spyder/issues/7981, it is closely related to the issue reported here.

jnsebgosselin:
Also, @gerazov, what is your operating system? Thanks!

gerazov:
Yeah, I saw #7981 but most of it was about Emacs support, I just wanted my shortcut back :smile: 
I'm on Manjaro 18.1.2 with KDE running on kernel 5.3.8-3.

jnsebgosselin:
> Yeah, I saw #7981 but most of it was about Emacs support, I just wanted my shortcut back 😄
> I'm on Manjaro 18.1.2 with KDE running on kernel 5.3.8-3.

What happens when you try to set the shortcut with `Alt + Space`?

gerazov:
Nothing happens - like I didn't press anything, but `Alt+x` gets registered fine.

FabianHofmann:
Hey, I'm having the same issue using Ubuntu 18.04

ccordoba12:
@FabianHofmann, those kind of comments are not helpful. For next time please leave an emoji reaction on the issue description.

jnsebgosselin:
> 
> Nothing happens - like I didn't press anything, but `Alt+x` gets registered fine.

Ok thanks. I don't understand why it is not registered because we are not blocking it. So if the OS is not eating the event, it should be passed to the shortcut editor dialog. 

When you press `Alt + Space` in the Editor, is this registered as a normal space? Sorry, but I don't have access to your OS, so I'm trying to guess where the problem is from your inputs.

goanpeca:
>Manjaro 18.1.2 with KDE running on kernel 5.3.8-3.

I could create a VM later and try to see what the issue is. On Mac `alt+space` is registered fine.

goanpeca:
This is close enough, right?
[...]

PR #10836, opened 2019-11-25
PR: Fix the shortcut override in the ShortcutEditor.

jnsebgosselin:
### Issue(s) Resolved

With this change, I can now register correctly the `Ctrl + Space` keysequence in the `ShortcutEditor`.

This may also solve #10786, but this will need to be tested, because on Windows, `Alt + Space` is still not usable after these changes.

### Affirmation

### DIFF
--- a/spyder/preferences/shortcuts.py
+++ b/spyder/preferences/shortcuts.py
@@ -291,7 +291,16 @@ class ShortcutEditor(QDialog):
 
     def event(self, event):
         """Qt method override."""
-        if event.type() in (QEvent.Shortcut, QEvent.ShortcutOverride):
+        # We reroute all ShortcutOverride events to our keyPressEvent and block
+        # any KeyPress event. This allows to register some key sequences for
+        # which no key press event are created by Qt.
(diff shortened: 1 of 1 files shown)


spyder/preferences/shortcuts.py
```diff
@@ -291,7 +291,16 @@ def accept(self):
 
     def event(self, event):
         """Qt method override."""
-        if event.type() in (QEvent.Shortcut, QEvent.ShortcutOverride):
+        # We reroute all ShortcutOverride events to our keyPressEvent and block
```
goanpeca: Why do we remove the `QEvent.Shortcut` @jnsebgosselin  ?

jnsebgosselin: It's never called... probably since we are blocking them in the override. But I'm not sure. I can add this back in the condition.

goanpeca: @ccordoba12 ?

ccordoba12: > It's never called

If it's never called, then we should leave it out.

PR #10843, opened 2019-11-26
PR: Fix sorting and loading data in the Variable Explorer

### LINKED ISSUE #10702: Variables disappear from Spyder 4.0 Variable Explorer
## Problem Description
Variables intermittently disappear from Variable Explorer
When running various code, a variable that was visible in the Variable Explorer disappears
_Sometimes_, scrolling away from the area of list where the variable _should_ be, then clicking the column heading once or twice, and scrolling back makes the variable appear

In this image the variable experiments, a Pandas dataframe, is missing:
image
Here, after scrolling and sorting as noted, experiments now appears:
image

1.  In some cases, running a code line that replaces the variable with a new value causes it to disappear, if it is not being displayed at present (i.e. if scrolled away).
e.g.
```plot_name = study_name + ts.timestamp() + '.png'``` might cause plot_name to disappear
2.  In other cases, if the code is run as a script, then Variable Explorer is checked, the variable may be missing

Note: Variable Explorer Options were tested.  Following is the current configuration, which did not resolve the issue:
image

Expected outcome is that the variable appears in Variable Explorer as soon as it is created, and remains there until it is removed from the environment.

```python-traceback

PASTE TRACEBACK HERE

```

## Versions
Spyder 4.0.0b5 
Python 3.6.9 64-bit
Qt 5.13.1
PyQt5 5.13.1
Windows 10 

### Dependencies
pylint >=0.25                :  2.3.1 (OK)
spyder_kernels >=1.5.0;<2.0.0:  1.5.0 (OK)
matplotlib >=2.0.0           :  3.1.2 (OK)
pandas >=0.13.1              :  0.25.2 (OK)
numpy >=1.7                  :  1.16.5 (OK)
scipy >=0.17.0               :  1.3.1 (OK)

```
PASTE DEPENDENCIES HERE
```

impact27:
## Description of Changes

This PR does the following:

* Sort dictionary keys to avoid jumping in the Variable Explorer (this was the problem reported on issue #10702).
* Load all data in the Variable Explorer when sorting because otherwise that operation is not performed correctly.
* Load all data when filtering variables because it's really odd to not find the variable you're looking for in case it's not shown among the first 50 ones.
* Leave the proxy model just in `RemoteCollectionsEditorTableView` because it's only necessary there. This was causing that sorting was not performed correctly for any collection opened through the Variable Explorer.

### Issue(s) Resolved

Fixes #10702 

### Affirmation


ccordoba12:
@impact27, thanks a lot for your help with this. Could you explain how this fixes issue #10702?

impact27:
The way the variable explorer works is that if there is more than 100 variables, only 50 are loaded. Then when the user scrolls to the end, 50 more are loaded.
Assuming the user sorts the variables alphabetically, the first 50 variable displayed needs to be the first 50 variables alphabetically, not the first 50 in the dictionnary, as is the case now.

Of course if the user choses to sort the variables in another way, the behaviour is still broken. This PR solves the most commun case.

impact27:
So now I am loading everything if the user tries to sort in another way. This is the only way to be sure.

spyder/plugins/variableexplorer/widgets/collectionseditor.py
```diff
@@ -1115,6 +1123,8 @@ def __init__(self, parent, data, readonly=False, title="",
         self.source_model = CollectionsModelClass(self, data, title,
                                                   names=names,
                                                   minmax=minmax)
+        self.horizontalHeader().sortIndicatorChanged.connect(
+            self.source_model.load_all)
```
ccordoba12: I don't think this is a good strategy and we should remove it. The thing is CollectionsEditor is used to view the contents of any dictionary, list, tuple and set. We use it as the main widget of the Variable Explorer because Python's namespace is a dictionary, but this change would affect how we display the contents of any other object viewed through it.

So if you have a list with 100,000 elements and you reverse its index, we'd try to show the full contents of it. And that would certainly block Spyder.

spyder/plugins/variableexplorer/widgets/collectionseditor.py
```diff
@@ -1416,6 +1426,9 @@ def __init__(self, parent, data, minmax=False, shellwidget=None,
             show_special_attributes=show_special_attributes,
             remote=True)
 
+        self.horizontalHeader().sortIndicatorChanged.connect(
+            self.source_model.load_all)
```
ccordoba12: We should also remove this to.

spyder/plugins/variableexplorer/widgets/collectionseditor.py
```diff
@@ -290,9 +295,12 @@ def canFetchMore(self, index=QModelIndex()):
         else:
             return False
 
-    def fetchMore(self, index=QModelIndex()):
+    def fetchMore(self, index=QModelIndex(), number_to_fetch=None):
```
ccordoba12: `fetchMore` is a Qt method, so I don't we should change its signature. So I don't agree with this change.

spyder/plugins/variableexplorer/widgets/collectionseditor.py
impact27: Ok but we need to know what the new first element is? And what if it is sorted by type or size? how do I know what to show without loading everything?

PR #10868, opened 2019-11-27
PR: Add traditional Chinese translation for Spyder 4

ccordoba12:

PR #10870, opened 2019-11-28
PR: Update Simplified Chinese translation against rc3

suokunlong:
and fixed some translation errors.


suokunlong:
related to PR #10868.

ccordoba12:
@suokunlong, your PR has conflicts with master, which shouldn't be the case if you had used our latest HEAD.

Although I could those conflicts for you, I'd prefer for you to create a new PR so we're sure your translation is not altered during that process. I hope that's ok for you.

Issue #10745, opened 2019-11-17
Variable explorer viewers too small on high DPI


jsh9:
On both Spyder 3.3.6 and Spyder 4.0.0rc1, when I tried to view some variable by double-clicking a variable from the variable explorer, I found the popup window is too small (in terms of the proportion to the whole screen).

I am using Ubuntu 18.04 on a 4k monitor, with a system-wide 200% zoom.

I tried to play with Spyder's high DPI scaling options (setting it to auto, or to manual 1.5, or to manual 1), and did not see any significant changes.

ccordoba12:
I think our viewers have a fixed size, and that was adjusted for non-high resolution monitors.

This is a very low priority issue for us, so maybe you or someone else in the community would like to tackle it.

jsh9:
Could you point me to the place in the code base? I may try to fix it, but I need some pointers first :)

Thanks!

ccordoba12:
Our viewers are placed here

https://github.com/spyder-ide/spyder/tree/master/spyder/plugins/variableexplorer/widgets

jsh9:
Hi @ccordoba12 , I found the hard-coded window sizes in this folder. Those numbers were written down 10 years ago, when the mainstream screen resolution was around 1024 x 768. That's why the variable explorer windows look so small on today's monitors.

I can manually increase the window sizes and submit a pull request.

One question: I also found that some dialog box (such as "Open file...") to be also very small. Shall I also fix it?  If so, please point me to the directory.

ccordoba12:
> I can manually increase the window sizes and submit a pull request.

Well, perhaps we should detect if the window is high resolution or not and adjust the sizes accordingly. I say it because the current sizes are still fine for me in regular screens.

> One question: I also found that some dialog box (such as "Open file...") to be also very small. Shall I also fix it? If so, please point me to the directory.

Is not that defined by the operating system? How does that dialog look in other applications?

jsh9:
Could you tell me how we can detect screen resolution? Are there any Python libraries?
  A few things we need to consider:
- Cross platform
- Real time detection? (If I have two monitors with different resolutions, and I could potentially more Spyder from one monitor to another any time.)

If we don't want to dynamically adjust the variable explorer window size, we can use some hard-coded window sizes that are more suitable for today's standard.

[...]

PR #10880, opened 2019-11-29
Increase variable explorer window size

jsh9:
## Description of Changes

Related issue: https://github.com/spyder-ide/spyder/issues/10745

The current variable explorer window sizes are small in today's higher resolution monitors. So I'm increasing their sizes.

Current size of a data frame viewer:

image

New size (just an example):

image

I'm waiting for @dalthviz 's help on implementing a dynamic window sizing, and I'll push up another commit to this PR.

### Issue(s) Resolved

Fixes #

### Affirmation


dalthviz:
Hi @jsh9 about a dynamic window sizing maybe we could:

1. Create a new method in the dialog class. The method will get as params the info about the geometry of the screen and the proportions for width and height (it will do the resize of the dialog too). Something like this:

```python
def set_dynamic_width_and_height(self, screen_geometry, width_proportion=0.5,
    height_proportion=0.5):
    """
    Update width and height using an updated screen geometry.
    Use a proportion for the width and height of the dialog.
    """
    screen_width = screen_geometry.width()
    screen_height = screen_geometry.height()
    self.resize(screen_width * width_proportion, screen_height * height_proportion)
```
2. Get the `QScreen` instance of the window of the dialog in the constructor (or maybe in a setup method since some elements of the call chain could be `None` inside of the constructor, I'm not sure). Something like this:
```python
screen = self.window().windowHandle().screen().geometry()
```
3. Connect the new method created at `1.` to the `geometryChanged` signal of the QScreen instance that we get at `2.`. Something like this:
```python
screen.geometryChanged.connect(self.set_dynamic_width_and_height)
```
Hope the elements above give you some hints around the things that can be done for the dynamic resizing. Let me know if you have questions or need more help :)

jsh9:
Hi @dalthviz , I add the following lines to the end of the constructor (for example, the constructor of `TextEditor(QDialog)`:
```python
    window = self.window()
    windowHandle = window.windowHandle()
    screen = windowHandle.screen()
    geometry = screen.geometry()
    screen.geometryChanged.connect(self.set_dynamic_width_and_height)
```

However, when I create a `str` variable and tried to view it, I got a null pointer exception:
```
    screen = windowHandle.screen()
AttributeError: 'NoneType' object has no attribute 'screen'
```
This indicates that `self.window().windowHandle()` is `None`.

So how can write this "setup method" that you mentioned? Thanks.

dalthviz:
@jsh9 rethinking this, maybe a better approach could be the creation of a dialog base class. The base class should look something like this:

```python
class BaseDialog(QDialog):
    def set_dynamic_width_and_height(self, screen_geometry, width_proportion=0.5,
        height_proportion=0.5):
        """
        Update width and height using an updated screen geometry.
        Use a proportion for the width and height of the dialog.
        """
        screen_width = screen_geometry.width()
        screen_height = screen_geometry.height()
        self.resize(screen_width * width_proportion, screen_height * height_proportion)

    def show(self):
        super(BaseDialog, self).show()
        window = self.window()
        windowHandle = window.windowHandle()
        screen = windowHandle.screen()
        geometry = screen.geometry()
        self.set_dynamic_width_and_height(geometry)
        screen.geometryChanged.connect(self.set_dynamic_width_and_height)
```
And then the dialogs need to inherit from this base class. For example, something like:

```Python
class TextEditor(BaseDialog):
    ...
```

jsh9:
Hi @dalthviz , I implemented your suggestion.  Please take a look at my changes.

I've tested the result on 3840x2160, 2560x1440, and 1366x768 resolutions, and they all have the expected behavior (i.e., the variable explorer window shows 50% of the screen height and 50% of the screen width).

There is one small remaining issue though: on higher resolution (for example 3840x2160 and 2560x1440), the initial position of the explorer window is always on the lower-right of the screen. Since we've hard-coded the relative window size to be 50%, it isn't an issue. But it would be better if we can center these windows. How do you think we can achieve this?

Thanks again!

dalthviz:
@jsh9 to center the dialogs we could add the end of the `set_dynamic_width_and_height` method something like this:

```python
        x = screen_geometry.center().x() - self.width() / 2
        y = screen_geometry.center().y() - self.height() / 2
        self.move(x, y)
```

spyder/plugins/variableexplorer/widgets/arrayeditor.py
```diff
@@ -40,7 +40,7 @@
                               to_text_string)
 from spyder.utils import icon_manager as ima
 from spyder.utils.qthelpers import add_actions, create_action, keybinding
-
+from .basedialog import BaseDialog
```
dalthviz: @jsh9 I think this import should be something like `spyder.plugins.variableexplorer.widgets.basedialog import BaseDialog` (and should be changed also for the other dialogs)

jsh9: Fixed. Thank you.

jsh9:
Hi @dalthviz , I made some changes according to your suggestions. Thank you!

Now I have three questions remaining:

1. Is there a way to apply an offset to the explorer windows, if more than one is open? This is a nice-to-have, and maybe out of the scope of this PR. But if there is a simple way to solve it, maybe I can just put it in as well.
2. I'd also like to change the size of the object editor, but it inherits from `QObject`. `QOobject`'s documentation (https://doc.qt.io/qtforpython/PySide2/QtCore/QObject.html) is quite complicated  to me. Do you have a good suggestion?
3. Some of the CI checks failed for no apparent reasons. I looked at the logs of those that failed, and everything looked fine.

dalthviz:
Hi @jsh9 about your questions:

1. I would need to think about that but, at least for now, I'm unsure of a proper way to achieve such behavior (maybe we could instead change the proportion per dialog reimplementing in each one the `set_dynamic_..` method, probably calling `super` but providing different default values for the height and width proportions for each dialog). What do you think @ccordoba12 ?

2. If you refer to the new `Object Explorer`, the dialog is at `spyder\spyder\plugins\variableexplorer\widgets\objectexplorer\objectexplorer.py`. Maybe you checked the object editor module (at `spyder\spyder\plugins\variableexplorer\widgets\objecteditor.py`) but that is used to create dialogs (is not a dialog itself).

3. Could you try to merge with master?

Thanks for all the work in this one @jsh9 👍. Also, could you post here a preview of how things are looking (you can create a `.gif` with something like [LICEcap](https://www.cockos.com/licecap/)) Thanks!

jsh9:
Hi @dalthviz , I've made `ObjectExplorer` and `ImportWizart` to inherit from `BaseDialog`. I've also pulled in the latest upstream master branch.

As for the "automatic window offser" feature, I'd say let's put it in the future wish list?

Below is the screen shot of the new window size. I manually dragged different variable explorer windows so that we can see all of the in a single screenshot:
image

ccordoba12:
@jsh9, this is looking pretty good. Thanks for your help with it!

We'd like to include your contribution in our next release in the Spyder 4 series. For that, please remove your last comment (i.e. your merge with master) and make your branch to derive from our `4.x` branch (instead of master), with the following command:

    git rebase --onto 4.x master variable_explorer_window

If you decline to do that, your work will be part of Spyder 5, to be released much later in 2020.

jsh9:
Hi @ccordoba12 , I reverted my merge commit, and did what you asked.  I'm not sure I did the entirely correct thing, because I'm not a git expert at all. Could you please take a look?

If you could make changes directly, it will be easier.  Thanks!

dalthviz:
Closing in favor of #10976

PR #10892, opened 2019-11-30
PR: Only assign icons to files and directories in Files

ccordoba12:
- This avoids a total freeze when moving to a directory with a pipe in it.
- I also did a refactoring of the way we get icons for SVG files according to the interface Spyder is running at the moment.


ccordoba12:
@dalthviz, please test that Kite completions don't break with this PR.

dalthviz:
@ccordoba12 Kite completions keep working for me with this PR :+1:

ccordoba12:
Thanks for verification @dalthviz!

PR #10903, opened 2019-12-02
Update conda bug

A01023437:
Else it throws an error

## Description of Changes

* Screenshot of error with the bug
Screenshot from 2019-12-01 21-50-45

I certify the above statement is true and correct: A01023437

### DIFF
--- a/requirements/conda.txt
+++ b/requirements/conda.txt
@@ -4,7 +4,7 @@ chardet >=2.0.0
 cloudpickle
 diff-match-patch
 intervaltree
-jedi =0.14.1
+jedi>=0.14.1
 keyring
 nbconvert
 numpydoc


ccordoba12:
Thanks for your interest in improving Spyder, but the file is called `conda.txt` for a reason, i.e. because it only works with conda.

If you want to use pip to run Spyder from master, you need to use

    pip install -e .

as it's clearly mentioned in our [Contributing guide](https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md#installing-dependencies).

ccordoba12:
And Spyder 4 doesn't work with Jedi `>=0.14.1`, but with that exact version.

A01023437:
Thanks @ccordoba12 for clearing it out!

PR #10910, opened 2019-12-02
PR: Show indent guides for splited editor panels

### LINKED ISSUE #10900: Regression: Indent guides broken in any Editor split panel except the first
As requested by @ccordoba12 . This is a serious regression in a major new feature that's been present since Beta 1, so it should presumably be fixed before the final release.

PR #10782 adds a new system for indent guides, but they are broken (don't appear at all) in any split Editor panel that isn't the first, leading me to initially think that they weren't working at all until I could get them to work in the first panel. Ergo, this should be fixed.

dalthviz:
## Description of Changes

ig

### Issue(s) Resolved

Fixes #10900 

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -854,9 +854,12 @@ class CodeEditor(TextEditBaseWidget):
         # Scrolling past the end
         self.set_scrollpastend_enabled(scroll_past_end)
 
-        # Line number area
+        # Line number area and indent guides
         if cloned_from:
             self.setFont(font) # this is required for line numbers area
+            # Needed to show indent guides for splited editor panels
+            # See spyder-ide/spyder#10900
+            self.patch = cloned_from.patch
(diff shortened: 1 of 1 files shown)


spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -854,9 +854,10 @@ def setup_editor(self,
         # Scrolling past the end
         self.set_scrollpastend_enabled(scroll_past_end)
 
-        # Line number area
+        # Line number area and indent guides
         if cloned_from:
             self.setFont(font) # this is required for line numbers area
+            self.patch = cloned_from.patch  # this is required for indentguides
```
ccordoba12: Since our tests fail due to this change, and we're not able to test this, let's enclose this line in an `if running_under_pytest():` so we can merge this PR.

ccordoba12: Also add a comment above this line with the following text:

```python
# NOTE: Please **do not** remove the line below because it'll break
folding and indent guides in splitted editors.
```

PR #10948, opened 2019-12-06
PR: Correctly show several missing deps in our dialog

ccordoba12:
Before

Selección_055

After

Selección_054


ccordoba12:
I'm leaving this for 4.1 because it touches a translation string.

PR #10963, opened 2019-12-07
PR: Remove Python 2 and 3.5 from our CIs

ccordoba12:
This is the first (and pretty simple) step to support only Python 3 in Spyder 5. Additionally, our plan is to support only Python 3.6+ in that version.

This will also be helpful to reduce the amount of available slots required to run our tests in each CI service.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,18 +17,6 @@ main: &main
           command: bash ./continuous_integration/circle/modules_test.sh || bash ./continuous_integration/circle/modules_test.sh
 
 jobs:
-  python2.7:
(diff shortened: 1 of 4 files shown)


CAM-Gerlach:
@ccordoba12 In order for users still running Spyder on Python 2 to not have their machines get an incompatible Spyder version, break things, and more than likely complain here, you should add `python_requires='>=3.6',` to your setup arguments alongside these changes for Spyder 5. I can submit a PR if you want.

ccordoba12:
> I can submit a PR if you want.

Ok, please do. Thanks!

PR #10973, opened 2019-12-08
Make shortcutkey list and dependency category localizable

suokunlong:
This commit makes the following strings translatable:
1) Help > Shortcut Keys Summary dialog. I added a dictionary holding all
the possible shortcut key names to do this.
2) The dependencies category strings (Help > Dependencies:
Mandatory, Optional.

I don't know how to update the POT file and thoese PO files, so someone
please help to update them in a separate commit.

### Affirmation

Kevin Suo <suokunlong at 126.com>


spyder/widgets/shortcutssummary.py
```diff
@@ -56,6 +56,144 @@ def __init__(self, parent=None):
             }"""
         self.setStyleSheet(style)
 
+
+        # Make the name of shortcut keys localizable
+        shortcut_names_map = {
```
ccordoba12: I don't think this is the right solution because we'd need to remember to update this dict every time a new shortcut is introduced anywhere in our source code.

Instead, we should generate this dict programmatically from our config options (present in `config/main.py`, see the `shortcuts` section) because every new shortcut needs to be registered there.

That way the shortcuts in `Preferences > Keyboard shortcuts` will be localized too.

goanpeca: I have a more general approach for this as you suggest. I think it is better to close this PR as the solution needs to take into account the Spyder plugin API

ccordoba12: Agreed.

ccordoba12:
@suokunlong, thanks for your contribution but @goanpeca already has a different kind of solution for this.

Unfortunately the solution will be available for Spyder 5, to be released next year.

PR #10990, opened 2019-12-09
PR: Add scroll line up/down keyboard shortcuts (Editor)

### LINKED ISSUE #10968: Feature request: scroll line up/down using keyboard shortcut
Hi!

I've found that vscode and pycharm have a quite neat feature, which is to scroll down/up one line per key press, using ctrl + up/down.

Here is vscode's keyboard shortcut: https://code.visualstudio.com/docs/getstarted/keybindings (seacrh for "scrollLineDown"), and PyCharm has the same keyboard shortcut.

Currently, Spyder's ctrl + up/down is reserved for navigating between different code cells, so I wonder if we can have a line-by-line scroll assigned to another keyboard shortcut?

jnsebgosselin:
## Description of Changes

scroll_with_keyboard

image

### Issue(s) Resolved

Fixes #10968

### Affirmation


spyder/config/main.py
```diff
@@ -438,6 +438,8 @@
               'editor/debug cell': 'Alt+Shift+Return',
               'editor/go to next cell': 'Ctrl+Down',
               'editor/go to previous cell': 'Ctrl+Up',
+              'editor/scroll line down': 'Ctrl+Shift+Down',
+              'editor/scroll line up': 'Ctrl+Shift+Up',
```
goanpeca: I do not think these shortcuts make a lot of sense, on mac.

I would expect that to select the whole document on one direction or the other I think these shortcuts should be empty and let the user use them as they see fit

ccordoba12: That's something to be solved with a new set of shortcuts specific for macOS, as you mentioned in another issue, not here, right?

jnsebgosselin: I think you are both right, but for the moment, the solution proposed by @goanpeca to just let these shortcuts empty and let the user use them as they see fit is the best.

goanpeca: >That's something to be solved with a new set of shortcuts specific for macOS, as you mentioned in another issue, not here, right?

Yes but something to consider is that there will always be less option for shortcuts on Mac since a lot of keyboards do not have a PGUp/PgDown and that is usually what Cmd+Up/Cmd+Down  translates to.

There is also User Interface Guidelines which do not really mention this at all. I am not sure what the current behavior is if a shortcut is `Empty` but I do not think we have to always define shortcut values.

jnsebgosselin: > I am not sure what the current behavior is if a shortcut is Empty but I do not think we have to always define shortcut values.

The shortcut is simply not bind to any key sequence, so that is not a problem.

jnsebgosselin: @ccordoba12 @goanpeca Just to be sure, do I go with the option where I leave this new shortcut blank?

ccordoba12: Yes, please do.

jnsebgosselin:
So as I've said in gitter, I'm currently unable to work on master after PR spyder-ide/spyder#11064 because of the following error ```refusing to allow an OAuth App to create or update workflow `.github/workflows/duplicates.yml` without `workflow` scope```. So that is why this branch is not up-to-date yet. I'll update it as soon as I find a working solution to this problem.

dalthviz:
Hi @jnsebgosselin sorry for the sudden ping, but just wondering, is okay if I go ahead and finish this? Let me know!

spyder/plugins/editor/widgets/editorstack/tests/test_editorstack.py
```diff
@@ -219,22 +267,17 @@ def test_move_multiple_lines_up(editor_bot):
     assert editor.toPlainText() == expected_new_text
 
 
-@pytest.mark.skipif(not sys.platform.startswith('linux'),
-                    reason="Works only on Linux")
-def test_copy_lines_down_up(editor_bot, mocker, qtbot):
+@pytest.mark.skipif(os.environ.get('CI', None) is not None and
+                    os.name == 'nt', reason="It fails on Windows CI")
```
ccordoba12, line 271: Please use here the `running_in_ci` function from `spyder/config/base.py`.

ccordoba12 (commented):
One comment for you @dalthviz, the rest looks good to me.

jnsebgosselin:
@dalthviz awesome, thanks for finishing this!

ccordoba12 (approved):
Thanks @jnsebgosselin for your here and @dalthviz for finishing it.

PR #11011, opened 2019-12-11
PR: Optimize editor scrollflag panel painting

### LINKED ISSUE #11000: Spyder very slow when working on large python files

A few things in spyder are slowing down spyder considerably, mostly whrn working with big files. The functions that take most of the time are:
 - the `paintEvent` from `spyder/plugins/editor/panels/scrollflag.py`
     - Mostly computing the flags and adding all the painters
 - `__mark_occurences` in `spyder/plugins/editor/widgets/codeeditor.py`

## Versions

* Spyder version: 5.0.0.dev0 acb3dbff4
* Python version: 3.7.5
* Qt version: 5.12.6
* PyQt5 version: 5.12.3
* Operating System: Darwin 19.0.0

### Dependencies

```
qtconsole >=4.6.0            :  4.7.0.dev0 (OK)
nbconvert >=4.0              :  5.6.1 (OK)
sphinx >=0.6.6               :  2.2.1 (OK)
pylint >=0.25                :  2.4.4 (OK)
psutil >=0.3                 :  5.6.5 (OK)
zmq >=17                     :  18.1.1 (OK)
spyder_kernels >=1.8.1;<2.0.0:  1.9.0.dev0 (OK)
sympy >=0.7.3                :  1.4 (OK)
IPython >=4.0                :  7.11.0.dev (OK)
matplotlib >=2.0.0           :  3.1.2 (OK)
pandas >=0.13.1              :  0.25.3 (OK)
numpy >=1.7                  :  1.17.4 (OK)
scipy >=0.17.0               :  1.3.3 (OK)
pyls >=0.31.2;<0.32.0        :  0.31.1 (NOK)
applaunchservices >=0.1.7    :  0.2.1 (OK)
```

jnsebgosselin:
## Description of Changes

This makes the painting of the flags a lot more efficient.

However, I get the feeling that the scrollflag is getting updated way to often due to signals sent by the editor. So I think there is still room for improvements on the editor side.

Here is what made the most impacts on improving the performance of the scrollflag panel painting:
- Caching the `QColor` instead of generating them for every flag.
- Passing the `x, y, width, height` values directly to `QPainter.drawRect` instead of creating a `QRect`
- Caching the style pixel metrics instead of calculating them for every flag.

scrollflag_painting

### Issue(s) Resolved

I think this Fixes #11000, but as I said, there is still room for improvement on the editor side.

### Affirmation

### DIFF
--- a/spyder/plugins/editor/panels/scrollflag.py
+++ b/spyder/plugins/editor/panels/scrollflag.py
@@ -12,8 +12,8 @@ from __future__ import division
 from math import ceil
 
 # Third party imports
-from qtpy.QtCore import QSize, Qt, QRect
-from qtpy.QtGui import QPainter, QBrush, QColor, QCursor
+from qtpy.QtCore import QSize, Qt
+from qtpy.QtGui import QPainter, QColor, QCursor
 from qtpy.QtWidgets import (QStyle, QStyleOptionSlider, QApplication)
 
 # Local imports
--- a/spyder/plugins/editor/panels/scrollflag.py
+++ b/spyder/plugins/editor/panels/scrollflag.py
@@ -45,8 +45,8 @@ class ScrollFlagArea(Panel):
             'error': QColor(editor.error_color),
             'todo': QColor(editor.todo_color),
             'breakpoint': QColor(editor.breakpoint_color),
-            'occurrence': editor.occurrence_color,
-            'found_results': editor.found_results_color
+            'occurrence': QColor(editor.occurrence_color),
+            'found_results': QColor(editor.found_results_color)
             }
         self._edgecolors = {key: color.darker(120) for
                             key, color in self._facecolors.items()}
@@ -62,6 +62,7 @@ class ScrollFlagArea(Panel):
         editor.sig_alt_mouse_moved.connect(self.mouseMoveEvent)
         editor.sig_leave_out.connect(self.update)
         editor.sig_flags_changed.connect(self.update)
+        editor.sig_flag_colors_changed.connect(self.update_flag_colors)
 
     @property
     def slider(self):
@@ -71,6 +72,15 @@ class ScrollFlagArea(Panel):
     def sizeHint(self):
         """Override Qt method"""
         return QSize(self.WIDTH, 0)
+
+    def update_flag_colors(self, color_dict):
+        """
+        Update the permanent Qt colors that are used for painting the flags
+        and the slider range with the new colors defined in the given dict.
+        """
+        for name, color in color_dict.items():
+            self._facecolors[name] = QColor(color)
+            self._edgecolors[name] = self._facecolors[name].darker(120)
 
     def paintEvent(self, event):
         """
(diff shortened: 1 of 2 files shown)


ccordoba12:
> So I think there is still room for improvements on the editor side.

It looks pretty smooth already!

spyder/plugins/editor/panels/scrollflag.py
```diff
@@ -34,6 +38,23 @@ def __init__(self, editor):
         self._range_indicator_is_visible = False
         self._alt_key_is_down = False
 
+        # Define permanent Qt colors that are needed for painting the flags
+        # and the slider range.
+        self._facecolors = {
```
goanpeca: When are these values updated in case of syntax color theme change?

jnsebgosselin: Good point. Unless I'm mistaken, only the `occurence_color` is affected by this, so I guess that value could be simply updated in the `_apply_highlighter_color_scheme` method of the codeeditor?

jnsebgosselin: hmmm, I think this should be done through a signal instead because the editor seems to never act directly on panels.

jnsebgosselin: ...or I can put this back in the paintEvent, but this will degrade a little bit the painting performances. @goanpeca What do you think?

goanpeca: I woulf move the creation of those helper variables to a separate method and see what signal method could be attached to panels 🤷🏻‍♂️

jnsebgosselin: > 
> 
> I woulf move the creation of those helper variables to a separate method and see what signal method could be attached to panels 🤷🏻‍♂️

Yes, I agree with this. We will need to add a new signal for this I think.

jnsebgosselin:
With the proposed changes, the scrollflagarea now take on average 0.043 sec, down from 0.293 sec with the sample file provided in #11000.

spyder/plugins/editor/panels/scrollflag.py
goanpeca: The signal should probably be related to the theme color which is more “general” and could be used by other panels.

jnsebgosselin: What do you mean concretely by that? Are you asking to just change the name of the signal I added or does this imply anything else?

goanpeca: Yes the signal name :-)

jnsebgosselin: Ok thank you! :)

jnsebgosselin:
> 
> 
> Looks great @jnsebgosselin. Thanks for working on this!
> 
> Not something to change, but do you have some ideas of tests we could write to keeps these benchmarks somewhere?

Thanks for the review @goanpeca .

I really don't know how we could write a test to assert these benchmarks. I'm not really knowledgeable regarding this kind of stuff. Maybe @impact27 can give us some insights in that regard.

ccordoba12:
> I really don't know how we could write a test to assert these benchmarks

No problem. I don't think we could to create a graphical benchmark for this easily.

Issue #5085, opened 2017-08-29
There are two scroll bars in the Editor


ma-sadeghi:
## Description of your problem
There are two scroll bars instead of one in the text editor. Here's how my text editor looks like:
![image](https://user-images.githubusercontent.com/14086031/29847266-bd031684-8ce8-11e7-851d-3d23dbba8973.png)

**What is the expected output? What do you see instead?**
I expect to only see one scroll bar!

## Versions and main components

* Spyder Version: 3.2.1
* Python Version: 3.6
* Qt Version: 5.9.1
* PyQt Version: 5.9
* Operating system: Windows

## Dependencies
IPython >=4.0    :  6.1.0 (OK)
cython >=0.21    :  None (NOK)
jedi >=0.9.0     :  0.10.2 (OK)
nbconvert >=4.0  :  5.2.1 (OK)
numpy >=1.7      :  1.13.1 (OK)
pandas >=0.13.1  :  0.20.3 (OK)
pycodestyle >=2.3:  2.3.1 (OK)
pyflakes >=0.6.0 :  1.6.0 (OK)
pygments >=2.0   :  2.2.0 (OK)
pylint >=0.25    :  1.7.2 (OK)
qtconsole >=4.2.0:  4.3.1 (OK)
rope >=0.9.4     :  0.10.7 (OK)
sphinx >=0.6.6   :  1.6.3 (OK)
sympy >=0.7.3    :  None (NOK)

jnsebgosselin:
The "scrollbar" on the left is not actually a scrollbar. This is an indicator that shows the portion of the file that is visible in the Editor, with the actual position of code errors (red lines), code analysis (yellow lines), positions of break points (red lines), etc... This is quite handy to navigate quickly to a given flag in a file.

ma-sadeghi:
@jnsebgosselin Maybe, but it's annoying to me. I'd like the editor to be minimal like in Sublime or Atom. Right now Spyder GUI is cluttered IMO. Is there any way to get rid of this so-called "scroll bar". Thanks for your quick reply btw 👍

ma-sadeghi:
@jnsebgosselin Also, I understand that the errors/warnings show up in this area, but I'm talking about the grayish box that looks like a scroll bar, not the whole area itself.

jnsebgosselin:
Unless I'm mistaken, I do not think there is such an option to do that in the options. @ccordoba12 what do you think?

[...]

PR #11057, opened 2019-12-12
PR: Set a minimum value for the scrollflag's slider height

### LINKED ISSUE #11050: Feature request: add a minimum height for the slider range of the scrollflag panel
## Problem Description

The slider range of the scrollflag panel is really useful for navigating quickly from one place to another in a file.

By design, the height of the slider range is scaled to the total height of the file. However, unlike the scrollbar's slider, there is no minimum height set on the scrollflag panel's slider range. So for files with a lot of lines, it becomes very small and can be hard to see on the screen.

I was thinking that maybe we should set a minimum height for the scrollflag panel's slider range. We could use the same minimum value as that used for the scrollbar?

image

jnsebgosselin:
## Description of Changes

Set the minimum height of the srollflag slider to the pixel metric set in the style for the scrollbar's slider.

image

### Issue(s) Resolved

Fixes #11050

### Affirmation

### DIFF
--- a/spyder/plugins/editor/panels/scrollflag.py
+++ b/spyder/plugins/editor/panels/scrollflag.py
@@ -226,6 +226,14 @@ class ScrollFlagArea(Panel):
         groove_rect = self.get_scrollbar_groove_rect()
         return groove_rect.y()
 
+    def get_slider_min_height(self):
+        """
+        Return the minimum height of the slider range based on that set for
+        the scroll bar's slider.
+        """
+        return QApplication.instance().style().pixelMetric(
+            QStyle.PM_ScrollBarSliderMin)
+
     def get_scrollbar_groove_rect(self):
         """Return the area in which the slider handle may move."""
         vsb = self.editor.verticalScrollBar()
@@ -298,6 +306,7 @@ class ScrollFlagArea(Panel):
         vsb = self.editor.verticalScrollBar()
         slider_height = self.value_to_position(
             vsb.pageStep(), scale_factor, offset) - offset
+        slider_height = max(slider_height, self.get_slider_min_height())
 
         # Calcul the minimum and maximum y-value to constraint the slider
         # range indicator position to the height span of the scrollbar area


goanpeca:
@jnsebgosselin now that you are working on this, should not both scrollbars look mor eor less the same?

Having two rectangles slightly out os sync looks a bit, weird... thoughts?

jnsebgosselin:
> @jnsebgosselin now that you are working on this, should not both scrollbars look mor eor less the same?
> 
> Having two rectangles slightly out os sync looks a bit, weird... thoughts?

By "out of sync",  are you talking about the style or the position?

goanpeca:
I see no reason why the slider should not have **the exact same height** and exact position (one next to the other) on both the flag area and the scroller

Screen Shot 2019-12-12 at 17 18 52

jnsebgosselin:
>Other IDEs solve this by only having a single thing instead of two bars...

This was discussed before in https://github.com/spyder-ide/spyder/issues/5085 and the single scrollbar option was rejected.

goanpeca:
>This was discussed before in #5085 and the single scrollbar option was rejected.

I know, but still it is very weird to have two scrollbars that offer the same function and look very different

ccordoba12:
One is the OS scrollbar and the other is equivalent to our minimap. So I don't think that's confusing. Besides, the scrollflag is only shown when the mouse is on top of it.

jnsebgosselin:
> > This was discussed before in #5085 and the single scrollbar option was rejected.
> 
> I know, but still it is very weird to have two scroll bars that offer the same function and look very different

IMO they don't offer the same function at all... besides, the slider on the scrollflag is shown only on mouse hover or when the Alt key is pressed.

goanpeca:
>One is the OS scrollbar and the other is equivalent to our minimap. So I don't think that's confusing.

It is redundant and it wastes spaces

>Besides, the scrollflag is only shown when the mouse is on top of it.

Still, it should not look different to the other one, it really makes o sense.

ccordoba12:
> IMO they don't offer the same function at all...

I agree with @jnsebgosselin here and it's pretty clear: the OS scrollbar allows users to hold it and move up and down quickly, whereas the scrollflag area let's you go to a specific region easily (usually one with warnings, errors or TODOs).

I use both objects quite comfortably and I don't think we should conflate them in a single scrollbar.

jnsebgosselin:
I use the functionalities offered by the scrollflag area all the time. Since we do not have bookmarks yet, I use breakpoints to mimic bookmarks. 

Pressing the alt key, I can then navigate really quickly from one breakpoint to the other with the scrollflag area by moving the mouse vertically and single left-clicking with the mouse. This is very useful.

I can also navigate really quickly going over the scrollflag area and single clicking on it. Since the slider follows the cursor, I can easily select the place in the file that contains the flags that I want to show on screen.

This is not something that the default OS scrollbar provides.

scrollflag_ux

PR #11064, opened 2019-12-13
PR: Add Github action to close duplicate issues

goanpeca:
This adds the action to handle duplicates and add a label.

Any thoughts @ccordoba12 ?

### DIFF
--- a/.github/workflows/duplicates.yml
+++ b/.github/workflows/duplicates.yml
@@ -13,7 +13,7 @@ jobs:
             [
              {
                "pattern": "ValueError: cannot set WRITEABLE flag to True of this array",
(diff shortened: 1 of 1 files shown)


ccordoba12:
Do we need a different repo per action? I mean, can't we have multiple actions in a single repo?

ccordoba12:
Also, can we add more patterns to close duplicates?

goanpeca:
>Do we need a different repo per action? I mean, can't we have multiple actions in a single repo?

If we want to share them with "the world" or other repos we need to have it like that.

>Also, can we add more patterns to close duplicates?

Sure

ccordoba12:
The error reported on issue https://github.com/spyder-ide/spyder/issues/11086 is still reported a lot. Please add it to the list.

goanpeca:
Will do!

PR #11089, opened 2019-12-14
WIP: Fix copying or duplicating multiple selected lines

### LINKED ISSUE #11070: Copying or duplicating multiple selected lines deselects last line each time
Currently on Spyder 4.0.0 but this one was there in Spyder 3 too, yet it seems no one was bothered enough to report it :smile: 

There's a small but irritating bug in the way `copy line` and `duplicate line` work for a selection of lines in the Editor.

If you select the lines up to the next new line, which is the normal way to select multiple lines using the keyboard, that each time you do `copy line` (`Alt+Shift+Down`) or `duplicate line` (`Alt+Shift+Up`) it forgets about the last copied line, i.e. deselects it. Thus, with each new copy, a smaller number number of the originally selected lines is copied, like so:

On the other hand if you select the lines up to the end of the last line, which takes more keystrokes to do, the copying works as expected:

jnsebgosselin:
## Description of Changes

@ccordoba12 Ok, so the fix for this is pretty straightforward. I also fixed an issue where the position of the cursor was not preserved after duplicating or copying current line or selected text. I also would like to add a test since this feature doesn't seemed to be covered by one.

However, as you can see in the animation below, there is a glitch that is clearly noticeable when duplicating due to the fact that text selection gets automatically expanded when inserting text at the current position of the cursor. That, is not easy to fix...

After some efforts, I found a workaround that allow to avoid the glitch that I have implemented in my third commit. I don't like workarounds, but I haven't been able to find a better way. Maybe I've missed something... Anyway, feel free to reject this workaround if you don't like it. I will simply revert the commit.

Without the workaround: 

duplicate_line_glitch

With the workaround: 

duplicate_line_glitch_2

### Issue(s) Resolved

Fixes #11070

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/base.py
+++ b/spyder/plugins/editor/widgets/base.py
@@ -617,6 +617,7 @@ class TextEditBaseWidget(QPlainTextEdit, BaseEditMixin):
         """Duplicate current line or selected text"""
         cursor = self.textCursor()
         cursor.beginEditBlock()
+        cur_pos = cursor.position()
         start_pos, end_pos = self.__save_selection()
         end_pos_orig = end_pos
         if to_text_string(cursor.selectedText()):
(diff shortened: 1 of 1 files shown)


ccordoba12:
I'm really sorry butI didn't notice that you changed the title of the PR before merging. Please open a new PR for the workaronud you mentioned in the description.

jnsebgosselin:
> I'm really sorry butI didn't notice that you changed the title of the PR before merging. Please open a new PR for the workaronud you mentioned in the description.

Ok good. I will do that, no problem.

PR #11114, opened 2019-12-17
PR: Handle CompletionWidget position when undocking editor

### LINKED ISSUE #11076: Kite pop-up doesn't show up if Editor pane undocked
Kite works fine if Editor pane docked (see Picture 1). If I undock the Editor pane and move to another monitor (let's say the "Vertical Right" monitor you see on Picture 1), Kite's pop-up window will not show up. You can see on Picture 2 an example of the issue (i.e. the absence of Kite's pop-up window).

I also observed that some shortcuts don't work if the Editor pane is undocked and on another monitor. For instance, if the cursor is on the Editor pane when undocked, and I press F5, it will not run the file. I have to click on the Console pane first. Maybe both issues are interconnected?

PS: tagging @metalogical as requested (as I contacted Kite first about the issue)

### Picture 1
Full screen

### Picture 2
Kite not working

dalthviz:
## Description of Changes

undock

### Issue(s) Resolved

Fixes #11076 

### Affirmation

### DIFF
--- a/spyder/api/plugins.py
+++ b/spyder/api/plugins.py
@@ -523,3 +523,9 @@ class SpyderPluginWidget(SpyderPlugin, BasePluginWidget):
             self.dockwidget.raise_()
         else:
             self.dockwidget.hide()
+
+    def set_ancestor(self, ancestor):
+        """
+        Needed to update the ancestor/parent of child widgets when undocking.
+        """
+        pass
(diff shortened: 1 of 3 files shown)


ccordoba12:
Great job @dalthviz! I only have one question before merging: what happens to the completion widget of the IPython console? Does it have the same problem as the one in the editor?

goanpeca:
> I only have one question before merging: what happens to the completion widget of the IPython console? Does it have the same problem as the one in the editor?

Good question, it probably has the same issue. We should maybe have a default behavior for the set_ancestor and only override for special plugins? so that not all plugins need to write custom code?

ccordoba12:
Note: The failure on Azure is unrelated to this.

PR #11180, opened 2019-12-25
PR: Account for find_replace shortcut context when checking for conflicts

### LINKED ISSUE #11169: shortcut would not correctly check valid

as the title

I change one shortcut but it fail to check whether the new shortcut is unique.

## Versions

* Spyder version: 4.0.0 
* Python version: 3.7.3
* Qt version: 5.9.6
* PyQt5 version: 5.9.2

### Dependencies

```
cloudpickle >=0.5.0          :  1.2.1 (OK)
qtconsole >=4.6.0            :  4.6.0 (OK)
nbconvert >=4.0              :  5.5.0 (OK)
sphinx >=0.6.6               :  2.1.2 (OK)
pylint >=0.25                :  2.3.1 (OK)
qtawesome >=0.5.7            :  0.5.7 (OK)
qtpy >=1.5.0                 :  1.8.0 (OK)
zmq >=17                     :  18.0.0 (OK)
spyder_kernels >=1.8.1;<2.0.0:  1.8.1 (OK)
sympy >=0.7.3                :  1.4 (OK)
cython >=0.21                :  0.29.12 (OK)
IPython >=4.0                :  7.6.1 (OK)
matplotlib >=2.0.0           :  3.1.0 (OK)
pandas >=0.13.1              :  0.24.2 (OK)
numpy >=1.7                  :  1.16.4 (OK)
scipy >=0.17.0               :  1.2.1 (OK)
pyls >=0.31.2;<0.32.0        :  0.31.2 (OK)
```

jnsebgosselin:
## Description of Changes

image

### Issue(s) Resolved

Fixes #11169

### Affirmation

### DIFF
--- a/spyder/preferences/shortcuts.py
+++ b/spyder/preferences/shortcuts.py
@@ -335,7 +335,8 @@ class ShortcutEditor(QDialog):
                 continue
             if (shortcut.context, shortcut.name) == (self.context, self.name):
                 continue
-            if shortcut.context in [self.context, '_'] or self.context == '_':
+            if (shortcut.context in [self.context, '_', 'find_replace'] or
+                    self.context in ('_', 'find_replace')):
                 if (shortcut_qsequence.matches(new_qsequence) or
                         new_qsequence.matches(shortcut_qsequence)):
                     conflicts.append(shortcut)


jnsebgosselin:
Failure is legit. I will fix this later.

jnsebgosselin:
@ccordoba12 @goanpeca I've looked into this issue further and I'm not sure how to handle this correctly, so I need your inputs.

The shortcuts of the `find_replace` context can conflicts only with those of the `console`, `editor`, and `ipython_console`, since these are used in plugins equipped with a `FindReplace` widget.

So my question is, should I check for conflicts between the `find_replace` context and only these three contexts or should I threat `find_replace` as the `_` (global) context and check for conflicts for all contexts?

goanpeca:
>So my question is, should I check for conflicts between the find_replace context and only these three contexts or should I threat find_replace as the _ (global) context and check for conflicts for all contexts?

I would say we should check for the specific contexts however we should be able to define this in a more stable and predictable way so we do not end up hardcoding this logic.

jnsebgosselin:
> I would say we should check for the specific contexts however we should be able to define this in a more stable and predictable way so we do not end up hard-coding this logic.

I agree, but how? I'm under the impression that this would require some major refactoring.

Maybe we can temporarily hardcode the logic for now, so that we can at least fix https://github.com/spyder-ide/spyder/issues/11169, and implement a better solution in another PR?

goanpeca:
>I agree, but how? I'm under the impression that this would require some major refactoring.

Yes for what I have in mind :-| please hardcode for now and create an issue with reference to this assigned to me, I think I will work on the shortcuts api refactor.

PR #11270, opened 2020-01-08
PR: Add css to style parameters in the Help pane when using Sphinx >=2

### LINKED ISSUE #9888: Missing space when rendering NumPy docstrings
In NumPy-style docs such as this snippet below:
```
Parameters
----------
some_param : type of this parameter
    Explanation about the parameter
```

Spyder's help renders the type of the parameter without any spacing from the name. Attached is a screenshot highlighting this:
spyder_format

Sphinx itself adds that extra separating space, as can be seen in e.g. the official numpy docs:
https://docs.scipy.org/doc/numpy-1.15.0/reference/generated/numpy.arange.html

dalthviz:
## Description of Changes

### With Sphinx 2.x
image

image

### With Sphinx 1.x

image

image

### Issue(s) Resolved

Fixes #9888 

### Affirmation


spyder/plugins/help/utils/static/base_css/base.css
```diff
@@ -0,0 +1,12 @@
+/*
+ * --- Parameters ---
+ * Needed to provide a separator between a
+ * parameters name and its classifier with sphinx>=2
+ * See spyder-ide/spyder#9888
+ * Based on https://github.com/numpy/numpydoc/issues/215#issuecomment-529037079
```
ccordoba12: @dalthviz, according to the last comment on this issue, i.e. https://github.com/numpy/numpydoc/issues/215#issuecomment-568261611, it seems this is going to be needed only for Numpydoc `< 1.0`?

dalthviz: @ccordoba12 I think that more than a numpydoc issue it is related to the change in Sphinx default writer (which in 1.8.x defaults to html4 and in 2.x to html5). After reviewing, one of the options was to add this new CSS, the other one was to go back to the html4 rendering with the [`html4_writer` option available for Sphinx >=2](https://github.com/sphinx-doc/sphinx/issues/6472) (although this last option didn't work for me after setting it in the `config.py` file we create to get the rich text from Sphinx). Should I give another try to the change of html rendering?

ccordoba12: > Should I give another try to the change of html rendering?

No, don't worry. I think the solution you found is doing the job that we need.

ccordoba12: > Should I give another try to the change of html rendering?

No, don't worry. I think the solution you found is doing the job that we need.

PR #11361, opened 2020-01-20
PR: Fix tooltip position after the editor is undocked

### LINKED ISSUE #11355: Editor: wrong tooltip position when undocking
The tooltips are being placed incorrectly after a docking or undocking of the Editor:

tooltipe

As discussed with @ccordoba12

dalthviz:
## Description of Changes

tooltip

### Issue(s) Resolved

Fixes #11355 

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/editor.py
+++ b/spyder/plugins/editor/widgets/editor.py
@@ -2395,10 +2395,10 @@ class EditorStack(QWidget):
             state = finfo.editor.document().isModified()
             self.set_stack_title(index, state)
 
-    def __reset_tooltip_tip(self):
-        """Reset last tooltip tip info."""
+    def __reset_tooltip(self):
+        """Reset tooltip tip info for all the editors."""
         for index, finfo in enumerate(self.data):
-            finfo.editor.reset_tooltip_tip()
+            finfo.editor.reset_tooltip()
 
     def refresh(self, index=None):
         """Refresh tabwidget"""
@@ -2416,7 +2416,7 @@ class EditorStack(QWidget):
             self.__refresh_readonly(index)
             self.__check_file_status(index)
             self.__modify_stack_title()
-            self.__reset_tooltip_tip()
+            self.__reset_tooltip()
             self.update_plugin_title.emit()
         else:
             editor = None
--- a/spyder/plugins/editor/widgets/editor.py
+++ b/spyder/plugins/editor/widgets/editor.py
@@ -2395,6 +2395,11 @@ class EditorStack(QWidget):
             state = finfo.editor.document().isModified()
             self.set_stack_title(index, state)
 
+    def __reset_tooltip_tip(self):
+        """Reset last tooltip tip info."""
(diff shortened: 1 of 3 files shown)


spyder/widgets/calltip.py
```diff
@@ -45,7 +45,7 @@ def __init__(self, parent=None, as_tooltip=False):
         """
         Shows tooltips that can be styled with the different themes.
         """
-        super(ToolTipWidget, self).__init__(parent, Qt.ToolTip)
+        super(ToolTipWidget, self).__init__(None, Qt.ToolTip)
```
goanpeca: Why is this needed?

spyder/widgets/mixins.py
```diff
@@ -661,6 +661,10 @@ def hide_tooltip(self):
         self._last_point = None
         self.tooltip_widget.hide()
 
+    def reset_tooltip_tip(self):
```
goanpeca: `reset_tooltip_tip`?

Is there any other kind of reset, could we just use `reset_tooltip`?

spyder/plugins/editor/widgets/editor.py
```diff
@@ -2395,6 +2395,11 @@ def __modify_stack_title(self):
             state = finfo.editor.document().isModified()
             self.set_stack_title(index, state)
 
+    def __reset_tooltip_tip(self):
```
ccordoba12: Let's change this to name to `__reset_tooltip`

spyder/widgets/calltip.py
ccordoba12: Why does the parent of this widget need to be None?

@goanpeca, could this have other undesired side effects?

spyder/widgets/mixins.py
ccordoba12: Let's change this name to `reset_tooltip`

spyder/widgets/calltip.py
dalthviz: Without this the tooltips position sometimes is incorrectly set. While testing, this was the way I found to make the position calculation to work correctly. I think that maybe the `Qt.ToolTip` window flag has something to do with this behavior too but I'm unsure.

Also, I choose to try this because the `CallTipWidget` does the same thing ([sets the parent to `None` when calling the super `__init__`](https://github.com/spyder-ide/spyder/blob/5b9d1e98e54c4a44f18f454d6f1169e7a57e49e3/spyder/widgets/calltip.py#L257))
```

spyder/widgets/mixins.py
dalthviz: Yep `reset_tooltip` everywhere sounds better. I will change it :+1:

spyder/widgets/calltip.py
dalthviz: @ccordoba12 please see the comment above

ccordoba12: Ok, thanks @dalthviz for the explanation. If we're already doing that for `CallTipWidget`, then we're on the safe side doing this here too.

PR #11415, opened 2020-01-28
PR: Fix some spelling errors

hjung4:


ccordoba12:
Hey @hjung4, thanks a lot for your contribution! We'd like to include it in our next release in the Spyder 4 series. For that, please make your branch to derive from our `4.x` branch (instead of master), with the following command:

    git checkout spell
    git rebase --onto 4.x master spell
    git push -f origin spell

If you decline to do that, your work will be part of Spyder 5, to be released much later in 2020 or 2021.

hjung4:
The command:
```
git rebase --onto 4.x master spell
```
gives the following error:
```
fatal: Does not point to a valid commit '4.x'
```

ccordoba12:
Then I think you need to run this command

    git checkout 4.x

before running those commands above.

hjung4:
This branch should now derive from the 4.x branch (instead of master).

spyder/locale/spyder.pot
```diff
@@ -3524,7 +3524,7 @@ msgid "Unlock"
 msgstr ""
 
 #: spyder/plugins/help/plugin.py:645
-msgid "The following error occured when calling <b>Sphinx %s</b>. <br>Incompatible Sphinx version or doc string decoding failed.<br><br>Error message:<br>%s"
+msgid "The following error occurred when calling <b>Sphinx %s</b>. <br>Incompatible Sphinx version or doc string decoding failed.<br><br>Error message:<br>%s"
```
ccordoba12: Please revert this change. There's no need to touch pot files because they are updated internally after translations strings are updated (as you already did for this string in `spyder/plugins/help/plugin.py`).

hjung4:
Changes have been made.

PR #11416, opened 2020-01-28
PR: Interpret -p project argument relative to the directory of invocation

### LINKED ISSUE #11363: Feature request: implicitly create project when launched with a folder path as argument
It would be really nice if I could load Spyder from the CLI when I am in a directory, with the contents of that directory, by running `spyder .` (or `spyder /home/akdor1154/src/project`, etc). AFAICT wiith Spyder's project model, this would require Spyder to
 - check if a project exists for that directory (i guess this already happens)
 - if not, create a new one
 - open said project.

I guess an implementation of this would be quite simple, would you accept a PR to that effect?

akdor1154:
## Description of Changes

When Spyder is invoked with `-p project`, this resolves `project` relative to the directory of invocation. IMO this is generally expected behaviour for running CLI tools. Currently it is interpreted relative to Spyder's working directory, which is generally my home directory.

E.g. if I am in `/home/akdor1154/src/python`, if I type `spyder -p folder1` or `spyder -p .`, I expect these to be interpreted as `/home/akdor1154/src/python/folder1` and `/home/akdor1154/src/python`, respectively. Currently they are interpreted as `/home/akdor1154/folder1` (surprising) and `/home/akdor1154` (really surprising).

N.B. existing behvaiour of `-p`, to create a project if it does not exist, is unchanged.

### Issue(s) Resolved

Fixes #11363 

### Affirmation

I certify the above statement is true and correct: akdor1154

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -2972,10 +2972,22 @@ class MainWindow(QMainWindow):
         variable explorer inside Spyder.
         """
         fname = encoding.to_unicode_from_fs(fname)
-        if osp.isfile(fname):
-            self.open_file(fname, external=True)
-        elif osp.isfile(osp.join(CWD, fname)):
-            self.open_file(osp.join(CWD, fname), external=True)
+        if osp.exists(osp.join(CWD, fname)):
+            fpath = osp.join(CWD, fname)
+        elif osp.exists(fname):
+            fpath = fname
+        else:
+            return
+
+        if osp.isfile(fpath):
+            self.open_file(fpath, external=True)
+        elif osp.isdir(fpath):
+            QMessageBox.warning(None, "Spyder",
+                _('Not opening folder <code>{fpath}</code> . To open that '
+                  'folder as a project from the CLI, use '
+                  '<code>spyder -p "{fname}"</code> .')
+                .format(fpath=osp.normpath(fpath), fname=fname)
+            )
 
     # --- Path Manager
     # ------------------------------------------------------------------------
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -348,7 +348,7 @@ class MainWindow(QMainWindow):
         self.profile = options.profile
         self.multithreaded = options.multithreaded
         self.new_instance = options.new_instance
-        self.open_project = options.project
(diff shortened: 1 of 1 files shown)


ccordoba12:
Hey @akdor1154, thanks a lot for your contribution! We'd like to include it in our next release in the Spyder 4 series. For that, please make your branch to derive from our `4.x` branch (instead of master), with the following command:

    git checkout project-relative
    git rebase --onto 4.x master project-relative
    git push -f origin project-relative

If you decline to do that, your work will be part of Spyder 5, to be released much later in 2020 or 2021.

akdor1154:
Rebased and snuck another commit in to show a warning if a directory is passed without -p, thanks.

ccordoba12:

spyder/app/mainwindow.py
```diff
@@ -2969,10 +2972,23 @@ def open_external_file(self, fname):
...
+        elif osp.exists(fname):
+            fpath = fname
+        else:
+            return
+
+        if osp.isfile(fpath):
+            self.open_file(fpath, external=True)
+        elif osp.isdir(fpath):
+            QMessageBox.warning(
+                None, "Spyder",
```
dalthviz: Please @akdor1154 lets change `None` for `self` and set as title `Warning` instead of `Spyder`.

spyder/app/mainwindow.py
```diff
@@ -2969,10 +2972,23 @@ def open_external_file(self, fname):
...
+                _('Not opening folder <code>{fpath}</code> . To open that '
```
dalthviz: Maybe we should phrase this message like:
```
Spyder didn't open the folder <code>{fpath}</code>. To open a folder 
as a project, please use <code>spyder -p "{fpath}"</code>.
```
what do you think @ccordoba12 ?

ccordoba12: Nice improvement @dalthviz! My suggestion:

```
To open <code>{fpath}</code> as a project with Spyder, 
please use <code>spyder -p "{fpath}"</code>.
```

spyder/app/mainwindow.py
ccordoba12: Let's use `_("Error")` instead of `Warning` or `Spyder` for this dialog's title, because this is really an error.

akdor1154:
thanks for the review guys :)

PR #11441, opened 2020-01-30
PR: Properly register shortcuts for some actions of the Variable Explorer

### LINKED ISSUE #10627: Variable explorer shortcuts not updated after configuration change

dalthviz:
## Description of Changes

Properly register shortcuts for the `NamespaceBrowser` and `CollectionsEditor` widget of the Variable Explorer to update shortcuts without a restart:
- [x] Search
- [x] Refresh
~~- [ ] Copy (seems like it's being done using a `keyPressEvent`?)~~ The copy shortcut preference is only for the `ArrayEditor` and `DataframeEditor`

### Issue(s) Resolved

Fixes #10627 

### Affirmation

### DIFF
--- a/spyder/plugins/variableexplorer/plugin.py
+++ b/spyder/plugins/variableexplorer/plugin.py
@@ -106,6 +106,7 @@ class VariableExplorer(SpyderPluginWidget):
         self.stack.removeWidget(nsb)
 
     def add_widget(self, nsb):
+        self.register_widget_shortcuts(nsb)
         self.stack.addWidget(nsb)
 
     # ----- Public API --------------------------------------------------------
--- a/spyder/plugins/variableexplorer/widgets/namespacebrowser.py
+++ b/spyder/plugins/variableexplorer/widgets/namespacebrowser.py
@@ -187,7 +187,34 @@ class NamespaceBrowser(QWidget):
 
         self.setLayout(layout)
 
+        # Local shortcuts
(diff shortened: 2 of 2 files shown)


goanpeca:
@dalthviz do you think we could add a regression test, or would it be just too much work for this use case?

dalthviz:
I think it could be too much work (mainly because I'm unsure on how to implement such a test, probably I will need to do a test using the whole main window?)

ccordoba12:
> The copy shortcut preference is only for the ArrayEditor

Could we add it to the DataframeEditor?

dalthviz:
> Could we add it to the DataframeEditor?

Checking, actually it is also used by the DataframeEditor

goanpeca:
I think we would need to use the whole window correct

ccordoba12:
> I think we would need to use the whole window correct

What do you mean?

goanpeca:
I mean that we qould need to run the main windows tests as daniel was asking. I thin it is doable. Not sure if it worth the troubel?

ccordoba12:
> I mean that we qould need to run the main windows tests as daniel was asking. I thin it is doable. Not sure if it worth the troubel?

I think it'll require too much for a very little gain. Let's merge this one as is for now.

PR #11530, opened 2020-02-11
PR: Prevent editor stealing focus of other widgets while scrolling

### LINKED ISSUE #11502: Unintuitive behavior: Mouse wheel scrolling in editor steals keyboard focus
## Problem Description

With keyboard focus in the internal console, hovering the mouse cursor over the editor window and scrolling through the code with the mouse wheel places keyboard focus back into the editor window instead of leaving it in the console. This feels very counterintuitive to me compared to the general behavior in Windows, where the mouse wheel can be used to scroll through background windows without the active one losing its focus. 

It seems like a minor issue but it's very annoying to me because now I keep accidentally messing up my code instead of typing in the console. I also think Spyder 3 did not show this behavior; not quite sure, but I don't remember anything like this before upgrading to 4.

## Versions

* Spyder version: 4.0.1
* Python version: Python 3.7.6 64-bit
* Qt version: 5.9.6 
* PyQt version: 5.9.2 

dalthviz:
## Description of Changes

Doing some updates while handling the `wheelEvent` (scroll related event) in the Editor, stealed focus of widgets like the Find widget:

A preview:

find

### Issue(s) Resolved

Fixes #11502 

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/base.py
+++ b/spyder/plugins/editor/widgets/base.py
@@ -1046,8 +1046,13 @@ class TextEditBaseWidget(QPlainTextEdit, BaseEditMixin):
                         self.zoom_in.emit()
                 return
         QPlainTextEdit.wheelEvent(self, event)
+        # Needed to prevent stealing focus to the find widget when scrolling
+        # See spyder-ide/spyder#11502
+        current_widget = QApplication.focusWidget()
         self.hide_completion_widget()
         self.highlight_current_cell()
+        if current_widget is not None:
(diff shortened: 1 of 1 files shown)


ccordoba12:
@dalthviz, does this also fix the issue reported in #11502? So the OP there says that if you have focus on the internal console and scroll in the editor, the focus is moved to the editor.

dalthviz:
@ccordoba12 yep it also solves that 👍 :

internal

PR #11576, opened 2020-02-17
PR: UX and UI Improvements to the Plots pane

jnsebgosselin:
## Description of Changes

- Removed the `save` and `delete` buttons from the thumbnails and moved these actions to the context menu.
- Removed the up/down buttons and added a vertical standard scrollbar instead.
- The keyboard arrow up/down keys are now moving the thumbnail selection instead of scrolling the list
- Added customizable shortcuts to every action in the toolbar that were missing one.
- Improved and homogenized tool-tips. The tooltips are now also updated if the shortcuts associated with a given action is changed.
- Improved the style of the figure viewer (see https://github.com/spyder-ide/spyder/pull/11576#issuecomment-587589621).
- Improved the style of the plot outline in the figure viewer (see https://github.com/spyder-ide/spyder/pull/11576#issuecomment-589696874).

The only thing I haven't done is to allow multiple thumbnails selection, but I think this should get implemented in another PR.

plot_changes

### Issue(s) Resolved

Fixes https://github.com/spyder-ide/spyder/issues/9367#issuecomment-586320067

### Affirmation


jnsebgosselin:
I want to try to make the `ThumbnailScrollBar` inherit from the standard `QListWidget`. I'm looking into it.

jnsebgosselin:
> 
> 
> I want to try to make the `ThumbnailScrollBar` inherit from the standard `QListWidget`. I'm looking into it.

Ok well, I think it is easier said than done... To be able to implement all the functionalities and behaviour that we have at the moment, this would require some work. It is not working out of the box, especially with the auto-resizing feature.

I think I will stick with the current structure of the thumbnail scrollbar and investigate if we could use a QListWidget later in another PR.

jnsebgosselin:
I have updated the desciption of changes section.

I just need to update the tests and this will be ready for review.

jnsebgosselin:
I decided to remove the custom stylesheet that was applied on the figure viewer. Even though it was looking good on Windows with the Vista theme, it was not looking so good with the dark style.

image

goanpeca:
@jnsebgosselin could you add an image of the context menu? (or perhaps a small animation?)

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -601,30 +646,24 @@ def __init__(self, figure_viewer, parent=None, background_color=None):
...
+        layout.addWidget(self.setup_scrollarea())
 
     def setup_scrollarea(self):
         """Setup the scrollarea that will contain the FigureThumbnails."""
         self.view = QWidget()
 
         self.scene = QGridLayout(self.view)
         self.scene.setContentsMargins(0, 0, 0, 0)
-        self.scene.setSpacing(3)
+        self.scene.setSpacing(5)
```
goanpeca: @jnsebgosselin pleas add a comment here to have a clearer idea of what this does.

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -638,36 +677,24 @@ def setup_scrollarea(self):
...
+        events to mimick navigational behaviour of a Qt widget list.
         """
+        if event.type() == QEvent.KeyPress:
+            key = event.key()
+            if key == Qt.Key_Up:
+                self.go_previous_thumbnail()
+                return True
+            elif key == Qt.Key_Down:
+                self.go_next_thumbnail()
+                return True
```
goanpeca: Just to check. When this is moved the selection changes and the actual view scrolls?

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -955,10 +961,10 @@ def highlight_canvas(self, highlight):
             # See spyder-ide/spyder#10255.
             if is_dark_interface():
                 self.canvas.setStyleSheet(
-                        "FigureCanvas{border: 2px solid %s;}" % "#148CD2")
+                    "FigureCanvas{border: 2px solid %s;}" % "#148CD2")
```
goanpeca: Where does this value come from? Maybe adding a reference to QDarkStyle could help, so that eventually we can remove this and use a centralized color handler.

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -1005,6 +1011,8 @@ class FigureCanvas(QFrame):
     """
     A basic widget on which can be painted a custom png, jpg, or svg image.
     """
+    sig_clear_fig_request = Signal()
```
goanpeca: I think signals are supposed to be named more like

`sig_clear_fig_requested` ?

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -1005,6 +1011,8 @@ class FigureCanvas(QFrame):
...
+    sig_save_fig_request = Signal()
```
goanpeca: same comment

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -1027,11 +1035,21 @@ def context_menu_requested(self, event):
         if self.fig:
             pos = QPoint(event.x(), event.y())
             context_menu = QMenu(self)
+            context_menu.addAction(
+                ima.icon('filesave'),
+                "Save plot as...",
```
goanpeca: Please add localization 
`_("Save plot as..."),`

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -1027,11 +1035,21 @@ def context_menu_requested(self, event):
...
+                lambda: self.sig_save_fig_request.emit(),
+                QKeySequence(CONF.get_shortcut('plots', 'save')))
             context_menu.addAction(
                 ima.icon('editcopy'),
                 "Copy Image",
                 self.copy_figure,
                 QKeySequence(CONF.get_shortcut('plots', 'copy')))
+            context_menu.addAction(
+                ima.icon('editclear'),
+                "Remove plot",
```
goanpeca: Pleas add localization

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -145,59 +155,70 @@ def setup(self, mute_inline_plotting=None, show_plot_outline=None,
 
     def setup_toolbar(self):
         """Setup the toolbar"""
```
goanpeca: Please add a . at the end

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -145,59 +155,70 @@ def setup(self, mute_inline_plotting=None, show_plot_outline=None,
...
-                tip=_("Next figure ({})").format(
-                      CONF.get_shortcut('plots', 'next figure')),
-                triggered=self.go_next_thumbnail)
+            self, icon=ima.icon('ArrowForward'),
+            tip=_("Next plot"),
+            triggered=self.go_next_thumbnail)
+        gonext_btn.shortcut_data = ('plots', 'next figure')
+        gonext_btn.installEventFilter(self)
 
         vsep2 = QFrame()
```
goanpeca: I know this is not related to this PR, but we shoul probably use longer descriptive names, we do not get points for shorter variables :-p

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -314,7 +359,7 @@ def show_fig_outline_in_viewer(self, state):
         """Draw a frame around the figure viewer if state is True."""
         if state is True:
             self.figviewer.figcanvas.setStyleSheet(
-                    "FigureCanvas{border: 1px solid lightgrey;}")
+                "FigureCanvas{border: 1px solid lightgrey;}")
```
goanpeca: this color is coming from?

spyder/plugins/plots/widgets/figurebrowser.py
jnsebgosselin: yes

spyder/plugins/plots/widgets/figurebrowser.py
jnsebgosselin: I removed `self.scene.setSpacing(5)` since 5 is already the default value, so there is no need to set it explicitly.

jnsebgosselin:
Improvements to the plot outline style.

```python
for i in range(5):
    plt.plot(np.random.rand(100), '.')
    plt.gcf().set_facecolor('none')
    plt.gca().set_facecolor('none')
    plt.tight_layout()
    plt.show()
```

image

goanpeca:
@jnsebgosselin is this ready for another review round, or you want to add something else?

ccordoba12:
@ok97465, do you agree with these changes? I mean, we've seen you're a heavy user of the Plots pane, so we'd like to know your opinion about these changes.

jnsebgosselin:
> 
> 
> @jnsebgosselin is this ready for another review round, or you want to add something else?

I have one last commit to do before it is ready. I'm waiting for the tests to finish.

goanpeca:
@jnsebgosselin ok, ping me when you do :-) Thanks!

jnsebgosselin:
> 
> 
> @jnsebgosselin ok, ping me when you do :-) Thanks!

Ok it is ready for another review.

ok97465:
@ccordoba12  I've been waiting for PR to be merged. I'm using it now, and it's more convenient than before. Thank you for the great work of the Spyder team. :+1:

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -144,60 +155,71 @@ def setup(self, mute_inline_plotting=None, show_plot_outline=None,
...
+            self, icon=ima.icon('ArrowForward'),
+            tip=_("Next plot"),
+            triggered=self.go_next_thumbnail)
+        gonext_btn.shortcut_data = ('plots', 'next figure')
+        gonext_btn.installEventFilter(self)
 
-        vsep2 = QFrame()
-        vsep2.setFrameStyle(53)
+        vertical_separator2 = QFrame()
+        vertical_separator2.setFrameStyle(53)
```
goanpeca: What is this 53 mean? is it a constant we can get from another place?

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -144,60 +155,71 @@ def setup(self, mute_inline_plotting=None, show_plot_outline=None,
...
+            tip=_("Copy plot to clipboard as image"),
             triggered=self.copy_figure)
+        copyfig_btn.shortcut_data = ('plots', 'copy')
+        copyfig_btn.installEventFilter(self)
 
-        closefig_btn = create_toolbutton(
-                self, icon=ima.icon('editclear'),
-                tip=_("Remove image"),
-                triggered=self.close_figure)
+        self.closefig_btn = create_toolbutton(
```
goanpeca: Not work for this PR, but should not we provide this information to the create tool button func?

So that we do not have to do this 

`self.closefig_btn.shortcut_data = ('plots', 'close')`  ?

Thoughts @jnsebgosselin if you think this would be useful?

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -144,60 +155,71 @@ def setup(self, mute_inline_plotting=None, show_plot_outline=None,
...
-        vsep2.setFrameStyle(53)
+        vertical_separator2 = QFrame()
+        vertical_separator2.setFrameStyle(53)
 
         self.zoom_out_btn = create_toolbutton(
-                self, icon=ima.icon('zoom_out'),
-                tip=_("Zoom out (Ctrl + mouse-wheel-down)"),
-                triggered=self.zoom_out)
+            self, icon=ima.icon('zoom_out'),
+            tip=_("Zoom out (Ctrl + mouse-wheel-down)"),
```
goanpeca: I think we need to add here something to take OSX users into account where this would be a `Cmd`instead of a `Ctrl.

Is this the case @impact27 (I do not have a mouse so I am not sure)

spyder/plugins/plots/widgets/figurebrowser.py
```diff
@@ -144,60 +155,71 @@ def setup(self, mute_inline_plotting=None, show_plot_outline=None,
...
+            triggered=self.zoom_out)
 
         self.zoom_in_btn = create_toolbutton(
-                self, icon=ima.icon('zoom_in'),
-                tip=_("Zoom in (Ctrl + mouse-wheel-up)"),
-                triggered=self.zoom_in)
+            self, icon=ima.icon('zoom_in'),
+            tip=_("Zoom in (Ctrl + mouse-wheel-up)"),
```
goanpeca: Same question here

jnsebgosselin:
> 
> 
> LGTM, just some final questions @jnsebgosselin (which might require a small change)
> 
> Thanks! great work :-)

Thanks @goanpeca.

Would it be OK if the style of the separators and the shortcuts to zoom in / out are addressed in their own PR?  I think this PR is already all over the place, so I hesitate to add even more changes here.

goanpeca:
Sure, but please open a issue to keep track of them.

spyder/plugins/plots/widgets/figurebrowser.py
jnsebgosselin: I don't know... That is the trick I used to update the tooltips of the toolbuttons for the plots plugin, but I think we need to think of how we could achieve the same results, but for all plugins that have toolbuttons with customizable shortcuts.

spyder/plugins/plots/widgets/figurebrowser.py
jnsebgosselin: I think this is out of scope for this PR and should be reported in another issue and addressed in another PR.

spyder/plugins/plots/widgets/figurebrowser.py
jnsebgosselin: Very valid point here, but same comment as before.

spyder/plugins/plots/widgets/figurebrowser.py
jnsebgosselin: Ok I used variable names instead of codes, so that it is more explicit.

spyder/plugins/plots/widgets/figurebrowser.py
impact27: Yes usually Ctrl and Cmd are swapped on OSX.

jnsebgosselin: I got an idea, I'm going to fix this here afterall.

jnsebgosselin:
I think the CI failure is not related to this work.

jnsebgosselin:
@goanpeca just letting you know that I adressed/answered all of your points after all, so there is no need to open other issues.

PR #11585, opened 2020-02-18
PR: Set OpenGL backend before the QApplication is created

### LINKED ISSUE #11227: set_options_opengl is called too late
https://github.com/spyder-ide/spyder/blob/26504d5ed48fdaa28b76cf72962859d4fdbce0cf/spyder/app/mainwindow.py#L3651

It needs to get called before the application is created. This means, before the splash screen.

The whole block that gets the options needs to be imported sooner, but importing
```
spyder.apps.cli_options
```
might be expensive to load

In either case, without moving it forward, Qt warns the user

Attribute Qt::AA_UseSoftwareOpenGL must be set before QCoreApplication is created.

ccordoba12:
Fixes #11227 

@dalthviz, please help me to finish this PR. The idea is to set the OpenGL implementation we use before we create `MAIN_APP`, i.e. below this point:

https://github.com/spyder-ide/spyder/blob/26504d5ed48fdaa28b76cf72962859d4fdbce0cf/spyder/app/mainwindow.py#L99-L104

The only problem is that I don't know how to capture our command line options at that point. That's the tricky part of this work.


ccordoba12:
@dalthviz, did you test your solution to see if it avoids the error reported on issue #11227 when using PyQt 5.12?

dalthviz:
Nop @ccordoba12 , I will setup a test env to check and I will let you know if works with that PyQt version

dalthviz:
@ccordoba12 I don't see the warning message when running Spyder using PyQt 5.12, so I think this is ready  :+1:

ccordoba12:
Great, thanks for extra verification @dalthviz!

PR #11593, opened 2020-02-18
PR: Ensure Help pane actions are enabled after being undocked

### LINKED ISSUE #11591: Options for different modes of documentation are not available when the Help Pane is undocked.
## Problem Description
 
Options for the pane's menu are deactivated when the pane is undocked

Screen Shot 2020-02-18 at 2 52 19 PM

1. 
2. 
3. 

```python-traceback

PASTE TRACEBACK HERE

```

## Versions

* Spyder version: 
* Python version: 
* Qt version: 
* PyQt version: 

### Dependencies

```
PASTE DEPENDENCIES HERE
```

dalthviz:
## Description of Changes

Enable all the actions when refreshing them for example when undocking:

help

### Issue(s) Resolved

Fixes #11591 

### Affirmation

### DIFF
--- a/spyder/plugins/help/plugin.py
+++ b/spyder/plugins/help/plugin.py
@@ -45,6 +45,7 @@ class Help(SpyderPluginWidget):
     CONF_FILE = False
     LOG_PATH = get_conf_path(CONF_SECTION)
     FONT_SIZE_DELTA = DEFAULT_SMALL_DELTA
+    DISABLE_ACTIONS_WHEN_HIDDEN = False  # SpyderPluginWidget class attribute
 
     # Signals
     focus_changed = Signal()


spyder/plugins/base.py
```diff
@@ -406,6 +406,11 @@ def _refresh_actions(self):
 
         # Create actions list
         self._plugin_actions = self.get_plugin_actions() + additional_actions
+        # Ensure all actions are enabled when adding them to the menu.
+        # See spyder-ide/spyder#11591
+        for action in self._plugin_actions:
+            if action is not None:
```
ccordoba12: Instead of `None`, perhaps we should use `MENU_SEPARATOR` here?

dalthviz:
Note: I'm quite unsure about the PEP8 error that popped-up (seems like the line has an indentation of 12 spaces / 3 tabs).

spyder/plugins/base.py
```diff
@@ -406,6 +406,11 @@ def _refresh_actions(self):
 
         # Create actions list
         self._plugin_actions = self.get_plugin_actions() + additional_actions
+        # Ensure all actions are enabled when adding them to the menu.
+        # See spyder-ide/spyder#11591
+        for action in self._plugin_actions:
+            if action is not MENU_SEPARATOR:
+                action.setEnabled(True)
```
ccordoba12: Not all actions are enabled or should be enabled when refreshing the Options menu. For example, the Options menu of the editor has one disabled action by default:

Selección_114

How does your proposed solution work in that case?

dalthviz: I need to rethink the approach here 🤔 (this is more a workaround than an actual fix of the root cause of the problem to be fair)

dalthviz: Although testing a little bit seems like this change doesn't affect the current disabled actions. For example, in the Editor the action `Close this panel` remains disable (I think the plugins and widgets change the state of the actions after the `_refresh_actions` is called):

help2

ccordoba12: > this is more a workaround than an actual fix of the root cause of the problem to be fair

So what could be the cause of this problem?

dalthviz: Maybe that an ancestor is being disable or hide when undocking and the actions get disabled. Anyhow, I need to check more deeply what its actually happening here

PR #11610, opened 2020-02-20
PR: The profiler is able to save data in a readable format

### LINKED ISSUE #11383: Make Profiler save data in a readable format
This is a feature request connected to an old closed issue #1544.

The `Profiler` is really useful, but it would be great to be able to visualise its output. The data can be saved but there is no documentation on its formatting. 

It would be nice to know how to read it, e.g. `pkl` or `csv`, and it would be great if it's in something that allows easy plotting, e.g. a `pandas` dataframe.

steff456:
## Description of Changes

Extend the profiler `save_data` to save both the object (for loading purposes) and a csv readable format.

image

image

### Issue(s) Resolved

Fixes #11383 

### Affirmation

### DIFF
--- a/spyder/plugins/profiler/widgets/profilergui.py
+++ b/spyder/plugins/profiler/widgets/profilergui.py
@@ -17,6 +17,7 @@ https://docs.python.org/3/library/profile.html
 from __future__ import with_statement
 import os
 import os.path as osp
+import io
 from itertools import islice
 import sys
 import time
@@ -516,18 +517,22 @@ class ProfilerDataTree(QTreeWidget):
 
     def save_data(self, filename):
         """Save profiler data."""
-        with open(filename, 'w') as f:
-            self.stats1[0].print_stats()
-            self.format_data(self.stats1[0])
-        # self.stats1[0].dump_stats(filename)
+        # Save the profiler data using a stream in a csv file
+        stream = io.StringIO()
+        self.stats1[0].stream = stream
+        self.stats1[0].print_stats()
+        data = stream.getvalue()
+        data = 'ncalls' + data.split('ncalls')[-1]
+        data = '\n'.join([','.join(
+            line.rstrip().split(None, 6)) for line in data.split('\n')])
 
-    def format_data(self, stats):
-        """Format the profiler data into a readable format."""
-        stats_list = stats.print_stats()
-        for stat in stats_list:
-            logger.debug('&&&&&')
-            logger.debug(stats_list)
+        # # save it to disk
+        f = open(filename.rsplit('.')[0] + '.csv', 'w')
+        f.write(data)
+        f.close()
 
+        # Save the profiler as object for loading purposes
+        self.stats1[0].dump_stats(filename)
 
     def find_root(self):
         """Find a function without a caller"""
--- a/spyder/plugins/profiler/widgets/profilergui.py
+++ b/spyder/plugins/profiler/widgets/profilergui.py
@@ -516,7 +516,18 @@ class ProfilerDataTree(QTreeWidget):
 
     def save_data(self, filename):
         """Save profiler data."""
-        self.stats1[0].dump_stats(filename)
+        with open(filename, 'w') as f:
+            self.stats1[0].print_stats()
+            self.format_data(self.stats1[0])
+        # self.stats1[0].dump_stats(filename)
+
+    def format_data(self, stats):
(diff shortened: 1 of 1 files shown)


goanpeca:
@steff456 is this preserving parent/child relationship?

Also we need to add some tests.

spyder/plugins/profiler/widgets/profilergui.py
```diff
@@ -516,6 +517,21 @@ def hide_diff_cols(self, hide):
...
+        self.stats1[0].print_stats()
+        data = stream.getvalue()
+        data = 'ncalls' + data.split('ncalls')[-1]
+        data = '\n'.join([','.join(
+            line.rstrip().split(None, 6)) for line in data.split('\n')])
+
+        # # save it to disk
+        f = open(filename.rsplit('.')[0] + '.csv', 'w')
+        f.write(data)
+        f.close()
```
goanpeca: Let's change this to

```python
    with open(filename.rsplit('.')[0] + '.csv', 'w') as fh:
        fh.write(data)
```

(it is more pythonic)

spyder/plugins/profiler/widgets/profilergui.py
```diff
@@ -516,6 +517,21 @@ def hide_diff_cols(self, hide):
...
+        # Save the profiler data using a stream in a csv file
+        stream = io.StringIO()
+        self.stats1[0].stream = stream
+        self.stats1[0].print_stats()
+        data = stream.getvalue()
+        data = 'ncalls' + data.split('ncalls')[-1]
+        data = '\n'.join([','.join(
+            line.rstrip().split(None, 6)) for line in data.split('\n')])
+
+        # # save it to disk
```
goanpeca: double `#`, also please start comments with uppercase

goanpeca:
@steff456 what do you mean by

>Extend the profiler save_data to save both the object (for loading purposes) and a csv readable format.

`object (for loading purposes)` ?

goanpeca:
Please ping the user that requested to check if this is ok.

ccordoba12:
@gerazov, is this enough for you?

goanpeca:
How about using xml as the user suggested? Would that avoid saving the ovject?

steff456:
> How about using xml as the user suggested? Would that avoid saving the ovject?

Whe need to save the object in order to load it after.

goanpeca:
>How about using xml as the user suggested? Would that avoid saving the object?

I still do not understand why this is needed. First what object are you refering to?

gerazov:
Yeah - looks great! Thanks @steff456! Sorry for not taking part in the discussion - I wasn't aware of the thread, thanks for the ping!

Looking at the printscreens it's not immediately clear to me how to transition from the CSV data to what you guys have in the output, i.e. the total time for each call and it's children. And this is what I would like to plot. Can the processed output be saved instead? 

If not then maybe a pointer to a way to process the CSV would be great.

spyder/plugins/profiler/widgets/profilergui.py
gerazov: is the uppercase comments something general for Python, or it's Spyder specific?

impact27:
> The tree is built with some information, some may be visible some other may be hidden, but we should be able to store this information as text (xml... etc) and load it to the Tree in the same way. This should be possible and if it is not clear right now how then we might need to rework the plugin, but storing and object via pickling etc seems unnecessary and points to a bad pattern.

The data saved by the python profiler is saved as a Stats object. https://docs.python.org/3/library/profile.html#pstats.Stats

From the docs:

> The file selected by the above constructor must have been created by the corresponding version of profile or cProfile. To be specific, there is no file compatibility guaranteed with future versions of this profiler, and there is no compatibility with files produced by other profilers, or the same profiler run on a different operating system. 

So I am not sure saving the dat in a text format and loading it would be easy without rewriting the Stats object.

ccordoba12:
> load the stats object into a custom tree structure

Do you have a specific structure in mind @impact27?

impact27:
> > load the stats object into a custom tree structure
> 
> Do you have a specific structure in mind @impact27?

I think what would make sense is a tree structure. We could adapt `show_tree` to load the tree in a tree structure instead of showing it. Then this structure can be saved and loaded from a file, as well as shown.

ccordoba12:
Closing this because it's outdated now with the refactoring done in PR #24794.

PR #11684, opened 2020-02-29
PR: Give focus to Plots pane when switching to it by shortcut if it's visible

ok97465:
## Description of Changes
I added funtionality for focus on plots pane when pressed the shortcut for switching to plot pane (ctrl + shift + G) if plots pane is visible.
If users pressed the switching shortcut and plots pane is visible, the thumbnail list of plots pane is controlled by arrow key. 

- visible is False
not_visible

- visible is True
visible

### Issue(s) Resolved

Fixes #

### Affirmation

I certify the above statement is true and correct: ok97465

### DIFF
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -175,8 +175,7 @@ DEFAULTS = [
              {
               'mute_inline_plotting': True,
               'show_plot_outline': False,
-              'auto_fit_plotting': True,
-              'focus_after_switching': False
+              'auto_fit_plotting': True
              }),
             ('editor',
              {
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -175,7 +175,8 @@ DEFAULTS = [
              {
               'mute_inline_plotting': True,
               'show_plot_outline': False,
-              'auto_fit_plotting': True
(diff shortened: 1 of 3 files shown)


ccordoba12:
Thanks a lot for your contribution @ok97465! But do we really need an option for this? I mean, in what situations do you need to go to Plots but not have focus on it?

ok97465:
[#9036 comments](https://github.com/spyder-ide/spyder/pull/9036#issuecomment-477373882)
I think that some users don't like sudden change in UX that lasted for about a year.

jnsebgosselin:
> 
> But do we really need an option for this? I mean, in what situations do you need to go to Plots but not have focus on it?

Maybe instead of adding an option, we could give focus to the Plots pane only when it is already visible, else we only switch to it without giving it focus. One would have to press the shortcut twice to switch to the Plots pane and give focus to it.

ccordoba12:
@jnsebgosselin, please approve if you consider this one is ready.

jnsebgosselin:
I've reviewed the code and everything LGTM and I've tested this locally and this is working as expected. Thanks @ok97465, this is a great improvement!

PR #11708, opened 2020-03-03
PR: Make pylint plugin find most local .pylintrc to given file and set cwd correctly

### LINKED ISSUE #5345: Static Code Analysis Not Reading .pylintrc file

CAM-Gerlach:
## Description of Changes

* [x] ~~Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))~~

Supersedes PR #10751 by @tovogt . This should be ready to merge as-is, presuming tests pass on the CIs (as they all did locally).

Previously, the Pylint plugin (Static Code Analysis pane) would not find the correct .pylintrc configuration file (or any config file at all) when run, either in the working dir, the project dir or the home directory. Furthermore, the working directory was not set to match that of Spyder. With these improvements, which were originally implemented on my suggestion by @tovogt , the cwd is now set correctly, and more importantly the most proximate `.pylintrc` file is retrieved and used, smoothly integrated with Pylint's built-in search path: 

1. `.pylintrc` (or `pylintrc`, `pyproject.toml`, or `setup.cfg`) or in the file's directory*
2. Any `.pylintrc` (etc.) in the Python module hierarchy at and above the file, at the closest level to the file*
3. `.pylintrc` (etc.) in Spyder's working directory*
4. `.pylintrc` (etc) in Spyder's active project directory (if any)*
5. `.pylintrc` (etc) named in the `PYLINTRC` environment variable
6. `.pylintrc` (etc) or `.config/pylintrc` in your home directory
7. `/etc/pylintrc` (on POSIX, obviously)

* Denotes items added/modified on the search path relative to pylint's naive config

In addition, I've implemented a battery of comprehensive unit and integration tests for this functionality.

### Issue(s) Resolved

Fixes #5345

### Affirmation

### DIFF
--- a/spyder/plugins/pylint/widgets/pylintgui.py
+++ b/spyder/plugins/pylint/widgets/pylintgui.py
@@ -317,7 +317,7 @@ class PylintWidget(QWidget):
 
         self.process = QProcess(self)
         self.process.setProcessChannelMode(QProcess.SeparateChannels)
-        self.process.setWorkingDirectory(osp.dirname(filename))
+        self.process.setWorkingDirectory(getcwd_or_home())
         self.process.readyReadStandardOutput.connect(self.read_output)
         self.process.readyReadStandardError.connect(
                                           lambda: self.read_output(error=True))
@@ -338,9 +338,9 @@ class PylintWidget(QWidget):
                 # 1.0
                 p_args += ["--msg-template='{msg_id}:{line:3d},"\
                            "{column}: {obj}: {msg}"]
-            p_args += [osp.basename(filename)]
+            p_args += [filename]
         else:
-            p_args = [osp.basename(filename)]
+            p_args = [filename]
         processEnvironment = QProcessEnvironment()
         processEnvironment.insert("PYTHONIOENCODING", "utf8")
         self.process.setProcessEnvironment(processEnvironment)
--- a/spyder/plugins/pylint/widgets/pylintgui.py
+++ b/spyder/plugins/pylint/widgets/pylintgui.py
@@ -58,7 +58,8 @@ def get_pylintrc_path(search_paths=None, home_path=None):
     if home_path is None:
         home_path = osp.expanduser("~")
     try:
-        pylintrc_paths = [_find_pylintrc_path(path) for path in search_paths]
+        pylintrc_paths = [
+            _find_pylintrc_path(path) for path in search_paths if path]
         pylintrc_path_home = _find_pylintrc_path(home_path)
         for pylintrc_path in pylintrc_paths:
             if (pylintrc_path is not None
--- a/spyder/plugins/pylint/widgets/pylintgui.py
+++ b/spyder/plugins/pylint/widgets/pylintgui.py
@@ -21,6 +21,7 @@ import time
 
 # Third party imports
 import pylint
+import pylint.config
 from qtpy.compat import getopenfilename
 from qtpy.QtCore import QByteArray, QProcess, Signal, Slot, QProcessEnvironment
 from qtpy.QtWidgets import (QHBoxLayout, QLabel, QMessageBox, QTreeWidgetItem,
--- a/spyder/plugins/pylint/widgets/pylintgui.py
+++ b/spyder/plugins/pylint/widgets/pylintgui.py
@@ -51,7 +51,8 @@ def _find_pylintrc_path(path):
     return pylint.config.find_pylintrc()
 
 
-def get_pylintrc_path(search_paths, home_path=None):
+def get_pylintrc_path(search_paths=None, home_path=None):
+    """Get the path to the highest pylintrc file on a set of search paths."""
     current_cwd = os.getcwd()
     pylintrc_path = None
     if home_path is None:
@@ -334,6 +335,17 @@ class PylintWidget(QWidget):
         """Try to start code analysis when Analyze button pressed."""
         self.start_analysis.emit()
 
+    def get_pylintrc_path(self, filename):
+        """Get the path to the most proximate pylintrc config to the file."""
+        proj_dir = self.parentWidget().main.projects.get_active_project_path()
+        search_paths = [
+            osp.dirname(filename),  # File's directory
+            getcwd_or_home(),  # Working directory
+            proj_dir,  # Project directory
+            osp.expanduser("~"),  # Home directory
+        ]
+        return get_pylintrc_path(search_paths=search_paths)
+
     @Slot()
     def start(self):
         """Start the code analysis."""
@@ -360,17 +372,10 @@ class PylintWidget(QWidget):
             else:
                 # Option '-i' (alias for '--include-ids') was removed in pylint
                 # 1.0
-                p_args += ["--msg-template='{msg_id}:{line:3d},"\
+                p_args += ["--msg-template='{msg_id}:{line:3d},"
                            "{column}: {obj}: {msg}"]
 
-        proj_dir = self.parentWidget().main.projects.get_active_project_path()
-        search_paths = [
-            osp.dirname(filename),  # File's directory
-            getcwd_or_home(),  # Working directory
-            proj_dir,  # Project directory
-            osp.expanduser("~"),  # Home directory
-        ]
-        pylintrc_path = get_pylintrc_path(search_paths)
+        pylintrc_path = self.get_pylintrc_path(filename=filename)
         if pylintrc_path is not None:
             p_args += ['--rcfile={}'.format(pylintrc_path)]
         p_args += [filename]
--- a/spyder/plugins/pylint/widgets/pylintgui.py
+++ b/spyder/plugins/pylint/widgets/pylintgui.py
@@ -13,7 +13,6 @@
 
 # Standard library imports
 from __future__ import print_function, with_statement
-import os
 import os.path as osp
 import re
 import sys
@@ -21,7 +20,6 @@ import time
 
 # Third party imports
 import pylint
-import pylint.config
 from qtpy.compat import getopenfilename
 from qtpy.QtCore import QByteArray, QProcess, Signal, Slot, QProcessEnvironment
 from qtpy.QtWidgets import (QHBoxLayout, QLabel, QMessageBox, QTreeWidgetItem,
@@ -36,6 +34,7 @@ from spyder.utils.misc import getcwd_or_home
 from spyder.widgets.comboboxes import (is_module_or_package,
                                        PythonModulesComboBox)
 from spyder.widgets.onecolumntree import OneColumnTree
+from spyder.plugins.pylint.utils import get_pylintrc_path
 from spyder.plugins.variableexplorer.widgets.texteditor import TextEditor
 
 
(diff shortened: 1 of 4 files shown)


CAM-Gerlach:
@ccordoba12 @tovogt After a three hour battle with the Py2 and legacy CIs, this is finally ready for final review and merge.

ccordoba12:
@CAM-Gerlach, thanks a lot for your hard work on this one!

@dalthviz, please review it.

spyder/plugins/pylint/widgets/pylintgui.py
```diff
@@ -44,6 +46,31 @@
...
+    os.chdir(path)
+    return pylint.config.find_pylintrc()
+
+
+def get_pylintrc_path(search_paths=None, home_path=None):
+    """Get the path to the highest pylintrc file on a set of search paths."""
+    current_cwd = os.getcwd()
+    pylintrc_path = None
+    if home_path is None:
+        home_path = osp.expanduser("~")
```
goanpeca: I think we have a more robust func to do this on https://github.com/spyder-ide/spyder/blob/master/spyder/config/base.py#L153

spyder/plugins/pylint/widgets/pylintgui.py
```diff
@@ -44,6 +46,31 @@
...
+    try:
+        pylintrc_paths = [
+            _find_pylintrc_path(path) for path in search_paths if path]
+        pylintrc_path_home = _find_pylintrc_path(home_path)
+        for pylintrc_path in pylintrc_paths:
+            if (pylintrc_path is not None
+                    and pylintrc_path != pylintrc_path_home):
+                break
```
goanpeca: What is the part in this code that we are trying to catch?

Ideally try/catch should be as small as possible to be very granular about what we are catching.

spyder/plugins/pylint/widgets/pylintgui.py
```diff
@@ -44,6 +46,31 @@
...
+    finally:
+        os.chdir(current_cwd)
+    return pylintrc_path
```
goanpeca: Add space above return for better readability

spyder/plugins/pylint/widgets/pylintgui.py
```diff
@@ -310,14 +337,29 @@ def analyze_button_handler(self):
...
+        parent = self.parentWidget()
+        if parent is not None:
+            project_dir = parent.main.projects.get_active_project_path()
+        else:
+            project_dir = None
+        search_paths = [
+            osp.dirname(filename),  # File's directory
+            getcwd_or_home(),  # Working directory
+            project_dir,  # Project directory
+            osp.expanduser("~"),  # Home directory
```
goanpeca: I think we have a more robust func to do this on https://github.com/spyder-ide/spyder/blob/master/spyder/config/base.py#L153

spyder/plugins/pylint/widgets/pylintgui.py
```diff
@@ -336,11 +378,13 @@ def start(self):
...
+                p_args += ["--msg-template='{msg_id}:{line:3d},"
                            "{column}: {obj}: {msg}"]
-            p_args += [osp.basename(filename)]
-        else:
-            p_args = [osp.basename(filename)]
+
+        pylintrc_path = self.get_pylintrc_path(filename=filename)
+        if pylintrc_path is not None:
+            p_args += ['--rcfile={}'.format(pylintrc_path)]
+        p_args += [filename]
```
goanpeca: Add space above p_args for better readability

spyder/plugins/pylint/widgets/pylintgui.py
```diff
@@ -44,6 +46,31 @@
...
+        home_path = osp.expanduser("~")
+    try:
+        pylintrc_paths = [
+            _find_pylintrc_path(path) for path in search_paths if path]
+        pylintrc_path_home = _find_pylintrc_path(home_path)
+        for pylintrc_path in pylintrc_paths:
+            if (pylintrc_path is not None
+                    and pylintrc_path != pylintrc_path_home):
+                break
+    finally:
```
dalthviz: Maybe should we add here an except statement to log the exception that could raise and properly show the kind of exception that could raise here?

spyder/plugins/pylint/widgets/pylintgui.py
dalthviz: I think the home path isn't needed here since the `get_pylintrc_path` function ignores it, right? or maybe should be passed as the `home_path` kwarg?

spyder/plugins/pylint/widgets/pylintgui.py
CAM-Gerlach: There are two reasons why we don't use this here:

* The path here needs to match the home path used by pylint, which uses `os.path.expanduser("~")`, otherwise the behavior will not be correct, and it raises a RuntimeError in case of home not being found rather than returning None, which is the desired behavior.

* At least under modern Python on Spyder's supported platforms (Windows and *nix), it appears doubtful that `get_home_dir()` is more "robust" in any case. It uses `os.path.expandvars("~")` under the hood, the only differences being that it performs an unnecessary decode from bytes when it is in fact already a string (i.e. unicode) on modern Python, and as a fallback haphazardly tries to get and decode the path from a few env vars, a more carefully-tailored, platform-specific superset of which are already queried by any supported version of [os.path.expanduser](https://docs.python.org/3/library/os.path.html#os.path.expanduser), plus a few strange and useless assignments to the `path` variable that is never used or returned.

This function was originally written by Pierre sometime over [11 years ago before](https://github.com/spyder-ide/spyder/blame/218473f89d9fcd0a48aace045cde4b5e99ff8080/spyderlib/userconfig.py#L55) by Pierre and added in the first commit to this repo from the even earlier Pydee, with only minor modifications as recently 8 years ago, so it is no surprise that it is no longer necessary or a best practice, if it ever was.

spyder/plugins/pylint/widgets/pylintgui.py
CAM-Gerlach: Its not excepting ("catching") anything, as if you look at the function this isn't a try-except at all, its a try-finally, in order to ensure the working directory gets set back to what it was prior to the function's execution. I normally include an explanatory comment documenting the purpose of any such similar construct, but I must not have added when I originally wrote the code since I didn't intend to implement it myself. I've added it now for the next commit.

spyder/plugins/pylint/widgets/pylintgui.py
CAM-Gerlach: Done, I normally do so but again, I originally offered this code as a proof of concept, not a 100% final version I would implement myself.

spyder/plugins/pylint/widgets/pylintgui.py
CAM-Gerlach: See previous comment for why we can't and shouldn't use that.

spyder/plugins/pylint/widgets/pylintgui.py
CAM-Gerlach: Ibid, done.

spyder/plugins/pylint/widgets/pylintgui.py
```diff
@@ -44,6 +46,31 @@
     import gettext
     _ = gettext.gettext
 
+
+def _find_pylintrc_path(path):
```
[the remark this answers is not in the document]
CAM-Gerlach: I suppose it could be, but IMO it makes much more sense to keep it somewhere inside the pylint plugin (maybe a different file if you suggest such), since its intrinsically part of the pylint plugin's implementation, not an external/general use utility. Furthermore, it would introduce complications, confusion and extra work for testing since it would either bifurcate the test code testing essentially the same thing at two different levels of abstraction, or require testing an external utility inside a plugin's tests.

spyder/plugins/pylint/widgets/pylintgui.py
CAM-Gerlach: A fair point. We don't expect any particular exception here, this is only for safety so that the working directory is reset properly in case of an unexpected error, following the general principle of not modifying de-facto global state unless necessary. Two possible scenarios do come to mind: First, pylint not being installed, which would raise (and presumably handle, e.g. by displaying a user warning or disabling the plugin) many errors long before this in the code; the error would in fact the error would presumably occur on import of the pylint module, which would be the appropriate place to handle it if we want to if it is not handled with a warning message on startup, not here.

Second if some new major version of pylint were to move, rename or remove the `pylint.config.find_pylintrc()` function, which we _want_ to fail with an actual error, not just a subtle log message, so that we see it immediately on our CIs before users do as soon as such an update goes out and gets pulled in by Spyder, and if for whatever reason users _do_ see it first, it is handled by Spyder's normal error report dialog so it gets reported so we can see and fix it, not buried in a local log somewhere.

spyder/plugins/pylint/widgets/pylintgui.py
CAM-Gerlach: Initially I removed it here when I added the `home_path` argument, but having it here makes it more clear that its on the search path and makes the operation of `find_pylintrc_path()` more clear and explicit to the reader, falling back to the base `.pylintrc` in the global user dirs, env vars and etc instead of the user having to infer this behavior if the a .pylintrc isn't found on the project path, as well as making the resulting tests simpler and more consistent.

spyder/plugins/pylint/widgets/pylintgui.py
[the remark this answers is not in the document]
dalthviz: Yep, when I said a `utils` module I was thinking in a new file inside the `pylint` directory. Maybe following a directory structure like:
```
spyder/plugins/pylint/utils/path.py
```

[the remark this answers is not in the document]
CAM-Gerlach: Oh, sorry, that makes sense then. I would suggest `spyder/plugins/pylint/utils/pylintrc.py` if design, spec or policy prohibits making it just a `utils` module in the base `pylint` dir. Given the relatively short length and unlikelihood of much else being added about path/pylintrc specifically as opposed to the higher possibility of more general pylint utils, for what its worth I would think it best to avoid adding yet another level of nesting to an already very deeply nested python package (given PEP 8 and `import this` (the Zen of Python) both strongly urge for keeping the namespace as flat as practical). However, this may be an consequence of the Spyder plugin spec or similar.

@ccordoba12 What do you suggest we go with?

CAM-Gerlach:
BTW, I realized I neglected to do so before, but for the record thanks @dalthviz and @goanpeca for your speedy reviews.

spyder/plugins/pylint/widgets/pylintgui.py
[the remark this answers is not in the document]
CAM-Gerlach: @goanpeca You may want to chime in too.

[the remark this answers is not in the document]
goanpeca: I would do something like: 

`spyder/plugins/pylint/utils.py`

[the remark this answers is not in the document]
CAM-Gerlach: Yeah, that was my preference too as I mentioned, I just wasn't sure if the Spyder plugin spec or policy/convention prohibited it based on @dalthviz not suggesting it himself. I'll implement it now.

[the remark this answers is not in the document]
CAM-Gerlach: Implemented and pushed now, plus some more minor formatting cleanup and a few more code comments.

CAM-Gerlach:
Thanks @goanpeca and @dalthviz for reviewing this and your suggestions!

@ccordoba12 ?

ccordoba12:
Thank you @CAM-Gerlach for your hard work on this one and @tovogt for the initial push to solve this problem. Great work guys!

PR #11714, opened 2020-03-04
PR: Fix issues with autocompletion characters and prevent whitespace from triggering completions

andfoy:
## Description of Changes
This PR fixes some regressions of autocompletion triggering behaviour, such as ignoring "dot" completions and preventing whitespace from triggering completions.

Peek 03-03-2020 20-42

### Issue(s) Resolved

Fixes #

### Affirmation

@andfoy

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -3824,7 +3824,9 @@ class CodeEditor(TextEditBaseWidget):
                     last_obj = getobj(text)
                     prev_char = text[-2] if len(text) > 1 else ''
                     if prev_char in {')', ']', '}'} or (last_obj and not last_obj.isdigit()):
-                        self._start_completion_timer()
+                        # Completions should be triggered immediately when an
+                        # autocompletion character is introduced.
+                        self.do_completion(automatic=True)
             else:
                 self.do_completion(automatic=True)
         elif (text in self.signature_completion_characters and
@@ -3895,7 +3897,7 @@ class CodeEditor(TextEditBaseWidget):
         key = self._last_pressed_key
         if key is not None:
             if key in [Qt.Key_Backspace, Qt.Key_Return, Qt.Key_Escape,
-                       Qt.Key_Tab, Qt.Key_Backtab]:
+                       Qt.Key_Tab, Qt.Key_Backtab, Qt.Key_Space]:
                 self._last_pressed_key = None
                 return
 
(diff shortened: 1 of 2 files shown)


spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3824,7 +3824,9 @@ def insert_text(event):
                     last_obj = getobj(text)
                     prev_char = text[-2] if len(text) > 1 else ''
                     if prev_char in {')', ']', '}'} or (last_obj and not last_obj.isdigit()):
-                        self._start_completion_timer()
+                        # Completions should be triggered immediately when an
+                        # autocompletion character is introduced.
+                        self.do_completion(automatic=True)
```
ccordoba12: So do you agree with this change @goanpeca? What about the problem you tried to solve with the timer before?

goanpeca: Yes I agree, I do not really mind. The only question is:

is using `.` considered automatic completion? in which case, should disabling the option on the preferences disable also this?

andfoy: I think we shouldn't handle it as automatic

ccordoba12: @andfoy, so you say we should change `automatic=False` here?

goanpeca: No @ccordoba12 that is a different thing. I mean that if you go to preferences and disable automatic completions then.

is `.` (dot completion) considered an automatic completion? and if not? is there a way to disable it if for some reason a user didi not want to have completions with a dot.

ccordoba12: > is . (dot completion) considered an automatic completion?

I'd say it is and this should be disabled accordingly.

goanpeca: In that case please add that fix @andfoy since currently this is not disabled if automatic completions are disabled.

andfoy: Now that you mention it, dot and others are defined as autocompletion trigger characters per the LSP specs. So I don't know if we should disable those as part of the automatic completions?

goanpeca: Maybe we should have aa different option for "token completions" @ccordoba12 ?

ccordoba12: Yes, I agree.

PR #11722, opened 2020-03-04
PR: Use resolved path (without symlinks) to save files

### LINKED ISSUE #11718: Symbolic and hard links "broken" when saving files
### Issue Report Checklist

    * [x] Restarted Spyder

## Problem Description

I have a few files which are either symbolic or hard links pointing to other files. Before I updated spyder from ~v.3 to the latest v.4.0.1 (from my anaconda distribution), if I modified and saved the symbolic/hard link, it also modified the other links pointing to the same file accordingly.
However, with v.4.0.1 if I modify a link (symbolic or hard, actually it does not matter), the link is "broken". 

As an example, if I have a file test.py and a symbolic link in another directory also called test.py, if I open the sym link, update its content and then save it, it will no longer be a sym link, but instead a new python file called test.py.
The same applies for hard links, where the link no longer exists after the update (typically stat test.py would show 2 links before saving, and 1 after).

From what I have tried, just opening the file without modifying it does not break the link, only when saving the link no longer exists.

Not sure if it's a bug or the normal behaviour, but it definitely changed between ~v.3 and v.4.1.0.

1. Create a python file test.py and fill it with some lines of code
2. Make a symbolic/hard link pointing to test.py, (e.g. `ln -s test.py test2.py`)
3. You can check if it worked with `ls -l test2.py` (you should see at the end of the line something like test2.py -> test.py)
4. Open test2.py with spyder, mofidy it and save it
5. Check the sym link again with `ls -l test2.py` (you should  only see test2.py at the end of the line)

Expected output would be that symbolic or hard links would not be broken, and that modifying one link would modify the content of all the other links pointing to the same file.

## Versions

* Spyder version: 4.0.1
* Python version: 3.7.6
* Qt version: 5.14.1
* PyQt version: 5.14.1
* Operating System name/version: Linux 5.3.0-40-generic (Ubuntu 19.10)

### Dependencies

```
cloudpickle >=0.5.0          :  1.3.0 (OK)
IPython >=4.0                :  7.12.0 (OK)
jedi =0.14.1                 :  0.14.1 (OK)
nbconvert >=4.0              :  5.6.1 (OK)
numpydoc >=0.6.0             :  0.9.2 (OK)
pexpect >=4.4.0              :  4.8.0 (OK)
psutil >=0.3                 :  5.6.7 (OK)
pygments >=2.0               :  2.5.2 (OK)
pylint >=0.25                :  2.4.4 (OK)
pyls >=0.31.2;<0.32.0        :  0.31.7 (OK)
xdg >=0.26                   :  0.26 (OK)
zmq >=17                     :  18.1.1 (OK)
qdarkstyle >=2.7             :  2.8 (OK)
qtawesome >=0.5.7            :  0.6.1 (OK)
qtconsole >=4.6.0            :  4.6.0 (OK)
rtree >=0.8.3                :  0.9.3 (OK)
sphinx >=0.6.6               :  2.4.0 (OK)
spyder_kernels >=1.8.1;<2.0.0:  1.8.1 (OK)
cython >=0.21                :  0.29.15 (OK)
matplotlib >=2.0.0           :  3.1.3 (OK)
numpy >=1.7                  :  1.18.1 (OK)
pandas >=0.13.1              :  1.0.1 (OK)
scipy >=0.17.0               :  1.4.1 (OK)
sympy >=0.7.3                :  1.5.1 (OK)
```

dalthviz:
## Description of Changes

To properly handle saves of files using symlinks the path used for atomicwrites needs to be the real path of the file

### Issue(s) Resolved

Fixes #11718 

NOTE: Seems like using `realpath` only works on Windows with Python 3.8+: https://bugs.python.org/issue9949 so maybe `pathlib` is a better alternative (although for Python 2 `pathlib2` needs to be used)

### Affirmation

### DIFF
--- a/spyder/utils/encoding.py
+++ b/spyder/utils/encoding.py
@@ -233,15 +233,16 @@ def write(text, filename, encoding='utf-8', mode='wb'):
     Return (eventually new) encoding
     """
     text, encoding = encode(text, encoding)
+    absolute_filename = os.path.realpath(filename)
     if 'a' in mode:
-        with open(filename, mode) as textfile:
+        with open(absolute_filename, mode) as textfile:
             textfile.write(text)
     else:
         # Based in the solution at untitaker/python-atomicwrites#42.
         # Needed to fix file permissions overwritting.
         # See spyder-ide/spyder#9381.
         try:
-            file_stat = os.stat(filename)
+            file_stat = os.stat(absolute_filename)
             original_mode = file_stat.st_mode
             creation = file_stat.st_atime
         except OSError:  # Change to FileNotFoundError for PY3
@@ -251,20 +252,20 @@ def write(text, filename, encoding='utf-8', mode='wb'):
             original_mode = 0o777 & ~umask
             creation = time.time()
         try:
-            with atomic_write(filename, overwrite=True,
+            with atomic_write(absolute_filename, overwrite=True,
                               mode=mode) as textfile:
                 textfile.write(text)
         except OSError as error:
             # Some filesystems don't support the option to sync directories
             # See untitaker/python-atomicwrites#17
             if error.errno != errno.EINVAL:
-                with open(filename, mode) as textfile:
+                with open(absolute_filename, mode) as textfile:
                     textfile.write(text)
         try:
-            os.chmod(filename, original_mode)
-            file_stat = os.stat(filename)
+            os.chmod(absolute_filename, original_mode)
+            file_stat = os.stat(absolute_filename)
             # Preserve creation timestamps
-            os.utime(filename, (creation, file_stat.st_mtime))
+            os.utime(absolute_filename, (creation, file_stat.st_mtime))
         except OSError:
             # Prevent error when chmod/utime is not allowed
             # See spyder-ide/spyder#11308
--- a/spyder/utils/encoding.py
+++ b/spyder/utils/encoding.py
@@ -26,8 +26,13 @@ from atomicwrites import atomic_write
 
 # Local imports
 from spyder.py3compat import (is_string, to_text_string, is_binary_string,
-                              is_unicode)
+                              is_unicode, PY2)
 from spyder.utils.external.binaryornot.check import is_binary
+
+if PY2:
+    import pathlib2 as pathlib
+else:
+    import pathlib
 
 
 PREFERRED_ENCODING = locale.getpreferredencoding()
@@ -233,7 +238,7 @@ def write(text, filename, encoding='utf-8', mode='wb'):
     Return (eventually new) encoding
     """
     text, encoding = encode(text, encoding)
-    absolute_filename = os.path.realpath(filename)
+    absolute_filename = to_text_string(pathlib.Path(filename).resolve())
     if 'a' in mode:
         with open(absolute_filename, mode) as textfile:
             textfile.write(text)
--- a/spyder/utils/encoding.py
+++ b/spyder/utils/encoding.py
@@ -240,8 +240,12 @@ def write(text, filename, encoding='utf-8', mode='wb'):
     """
     text, encoding = encode(text, encoding)
 
-    if os.name == "nt":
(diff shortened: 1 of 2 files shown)


ccordoba12:
> so maybe pathlib is a better alternative (although for Python 2 pathlib2 needs to be used)

It's already used in our LSP client. For example, in this file:

https://github.com/spyder-ide/spyder/blob/master/spyder/plugins/completion/languageserver/client.py

So please use it for this case too.

goanpeca:
Looks good @dalthviz :-), just missing a test :-p

dalthviz:
Note: I'm unsure how the changes made here affect the test that is failing: `spyder/app/tests/test_mainwindow.py::test_run_code`

ccordoba12:
Is this ready for review @dalthviz?

dalthviz:
Yep @ccordoba12 👍

goanpeca:
@dalthviz do you think we should add a similar test for the whole main windows by opening a file changing its contents and saving and checking the symlink remains and the original file was changed.?

dalthviz:
@goanpeca I think isn't need since at the end, when saving a file, what the Editor calls is the `encoding.write` function (passing the text and a filename), but if you guys want I could implement such a test using the whole main window :+1:

goanpeca:
> I think isn't need since at the end, when saving a file, what the Editor calls is the encoding.write function (passing the text and a filename), but if you guys want I could implement such a test using the whole main window 👍

Ok, no need to  do this ;-) just wanted to double check!

spyder/utils/encoding.py
```diff
@@ -233,15 +239,21 @@ def write(text, filename, encoding='utf-8', mode='wb'):
     Return (eventually new) encoding
     """
     text, encoding = encode(text, encoding)
+
+    if os.name == "nt":
+        absolute_filename = to_text_string(pathlib.Path(filename).resolve())
```
goanpeca: Does this work independently if the file exists or not?

If the file for some reason was "deleted"by the time this is called, is there an error we should try to handle @dalthviz ?

dalthviz: Checking it seems like it can throw some errors if the path doesn't exists https://docs.python.org/3.5/library/pathlib.html#pathlib.Path.resolve

goanpeca: In that case, could we add a try:except there and write a test for it?

spyder/utils/tests/test_encoding.py
```diff
@@ -10,12 +10,46 @@
...
+    # Assert the symlink was created
+    assert os.path.islink(symlink_file_path)
+
+    # Write using the symlink
+    enconding = write("New text for symlink", symlink_file_path)
+
+    # Assert symlink is valid and contents of the file
+    assert os.path.islink(symlink_file_path)
+    assert base_file.read_text(enconding) == symlink_file.read_text(enconding)
+    assert symlink_file.read_text(enconding) == 'New text for symlink'
```
ccordoba12:
```suggestion
    text = write("New text for symlink", symlink_file_path)

    # Assert symlink is valid and contents of the file
    assert os.path.islink(symlink_file_path)
    assert base_file.read_text(text) == symlink_file.read_text(text)
    assert symlink_file.read_text(text) == 'New text for symlink'
```

dalthviz: @ccordoba12 I called the variable `encoding` since that's what the `write` function returns and also that's the first argument of [`read_text`](https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text). Maybe I should change the call to read text from `read_text(encoding)` to `read_text(encoding=encoding)` for clarity?

ccordoba12: Ok, no problem. Still, there was a spelling error because you wrote `enconding` instead of `encoding` as the variable defined above. I updated my suggestion to fix that.

PR #11732, opened 2020-03-05
PR: Do not include complete words as part of completion suggestions

### LINKED ISSUE #11600: Auto-completion should not complete already complete words
## Problem Description
Auto-completion (without kite, maybe also with, but I didn't test) propose to complete already complete words.

Use LSP auto-completion (not tested with kite).

Write
```
something = 1
something
```
and wait. Auto-completion propose `something`

I would expect auto-completion to disappear when I write entirely what I wanted to.
It's not a major bug, but quite annoying.

## Versions

* Spyder version: 4.0.1
* Python version: 3.7.6
* Jedi version: 0.14.1
*  python-language-server version: 0.31.8

steff456:
## Description of Changes

Add a comparison to check if the prefix of a completion request is the response so we don't show it.

completion

### Issue(s) Resolved

Fixes #11600 

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/tests/test_introspection.py
+++ b/spyder/plugins/editor/widgets/tests/test_introspection.py
@@ -405,9 +405,6 @@ def test_completions(lsp_codeeditor, qtbot):
 
     qtbot.keyPress(code_editor, Qt.Key_Escape)
 
-    with qtbot.waitSignal(completion.sig_show_completions,
-                          timeout=10000) as sig:
-        qtbot.keyPress(code_editor, Qt.Key_Tab)
     if PY2:
         assert "hypot(x, y)" in [x['label'] for x in sig.args[0]]
     else:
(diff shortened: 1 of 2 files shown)


spyder/plugins/editor/widgets/tests/test_introspection.py
```diff
@@ -416,6 +411,19 @@ def test_completions(lsp_codeeditor, qtbot):
...
 
+    qtbot.keyPress(code_editor, Qt.Key_Escape)
+
+    try:
+        with qtbot.waitSignal(completion.sig_show_completions,
+                              timeout=10000) as sig:
+            qtbot.keyPress(code_editor, Qt.Key_Tab)
+    except pytestqt.exceptions.TimeoutError:
+        # This one should generate a timeout error because the prefix is the
+        # same that the returned completions of jedi
```
ccordoba12:
```suggestion
        # This should generate a timeout error because the completion
        # prefix is the same that the completions returned by Jedi.
        # This is a regression test for spyder-ide/spyder#11600
```

PR #11755, opened 2020-03-09
PR: Some improvements to fallback completions

ccordoba12:
This PR does the following things:

* It sends the current word to be completed to the fallback plugin, so it can filter among all tokens those that start with that word.
* It avoids the waiting logic between Kite and the LSP when fallback is the only client available.

With these changes, now we can have nice fallback completions for any file, as shown below:

Selección_040

### DIFF
--- a/spyder/plugins/completion/plugin.py
+++ b/spyder/plugins/completion/plugin.py
@@ -50,6 +50,7 @@ class CompletionManager(SpyderCompletionPlugin):
             LSPRequestTypes.DOCUMENT_COMPLETION: {
                 KiteCompletionPlugin.COMPLETION_CLIENT_NAME,
                 LanguageServerPlugin.COMPLETION_CLIENT_NAME,
+                FallbackPlugin.COMPLETION_CLIENT_NAME
             },
             LSPRequestTypes.DOCUMENT_SIGNATURE: {
                 KiteCompletionPlugin.COMPLETION_CLIENT_NAME,
@@ -143,7 +144,7 @@ class CompletionManager(SpyderCompletionPlugin):
         request_responses = self.requests[req_id]
 
         def send():
-            # Needed to prevent the send of completions for old requests
+            # Needed to prevent send of completions for old requests
             # See spyder-ide/spyder#10798
             req_type = self.requests[req_id]['req_type']
             response_instance = id(self.requests[req_id]['response_instance'])
--- a/spyder/plugins/completion/plugin.py
+++ b/spyder/plugins/completion/plugin.py
@@ -120,8 +120,7 @@ class CompletionManager(SpyderCompletionPlugin):
         with QMutexLocker(self.collection_mutex):
             request_responses = self.requests[req_id]
             request_responses['sources'][completion_source] = resp
-
-            self._maybe_send_locked(req_id)
+            self.howto_send_to_codeeditor(req_id)
 
     @Slot(int)
     def receive_timeout(self, req_id):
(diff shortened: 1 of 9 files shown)


spyder/plugins/completion/plugin.py
```diff
@@ -50,6 +50,7 @@ class CompletionManager(SpyderCompletionPlugin):
             LSPRequestTypes.DOCUMENT_COMPLETION: {
                 KiteCompletionPlugin.COMPLETION_CLIENT_NAME,
                 LanguageServerPlugin.COMPLETION_CLIENT_NAME,
+                FallbackPlugin.COMPLETION_CLIENT_NAME
```
ccordoba12: @metalogical, I think you decided to not include fallback here. Do you remember why?

metalogical: After the `wait_for` timeout has elapsed, we return immediately once *any one* of the `WAIT_FOR_SOURCE`s has returned a completion.

If we add fallback here, then basically, we'll return immediately once `wait_for` timeout elapses regardless of whether or not either of LSP/Kite has returned completions.

This may be ok, but we decided to wait for one of those and ignore fallback when we first implemented this.

goanpeca:
Does it make sense to add some fuzzy matching to this? So that not only words that start with but are part of the words in the completions list?

spyder/plugins/completion/plugin.py
ccordoba12: Well, our tests are failing precisely for Kite, so this is not working correctly. Then how is it that fallback is working for Python files and not for non-Python files with no LSP nor Kite support? Adding this line here was the only way I found to make fallback work for those files.

Should we make LSP and Kite return an empty list (as suggested by @andfoy) to fix that?

ccordoba12:
> Does it make sense to add some fuzzy matching to this?

Sure, that would be really nice to have. But we can leave it for another PR because that's an enhancement not directly related to this.

spyder/plugins/completion/plugin.py
metalogical: yes, the assumption is that Kite/LSP always return at least an empty list. I thought I fixed that when I implemented this, but I may have missed something. I also didn't consider the non-Python file case carefully.

ccordoba12:
@metalogical, I think I found two problems with files that don't provide LSP nor Kite:

1. This line in the `_maybe_send_locked` method

    https://github.com/spyder-ide/spyder/blob/1f0dfd9dc16459af8a0a24f909d053c0d5cbc048/spyder/plugins/completion/plugin.py#L159

    removes the `req_id` from `self.requests` before fallback completions can be processed because LSP and Kite return immediately and their results are empty.

2. Even if that line were handled appropriately, there is still this last part in `_maybe_send_locked`

    https://github.com/spyder-ide/spyder/blob/1f0dfd9dc16459af8a0a24f909d053c0d5cbc048/spyder/plugins/completion/plugin.py#L165-L181

    which doesn't include the fallback plugin among the possible sources for completions. I think that should be changed in `any_nonempty` to include it.

What do you think?

spyder/plugins/completion/plugin.py
ccordoba12: I found a way to avoid the waiting logic you added when fallback is the only client available.

spyder/plugins/completion/plugin.py
```diff
@@ -156,19 +157,35 @@ def send():
...
+                         "removed".format(req_id))
             del self.requests[req_id]
 
             # Response only to recent requests.
             if send:
                 self.gather_and_send(request_responses)
 
+        # Skip the waiting logic below when fallback is the only
+        # client that works for language
+        if self._is_fallback_only(language):
```
metalogical: this seems like a good approach

ccordoba12: Thanks! It took me several hours of hard thinking to come up with it.

PR #11764, opened 2020-03-10
PR: Fix saving protected files on Windows by checking if they are read-only

### LINKED ISSUE #10657: Saving a protected file hangs Spyder
Here protected means "read only".
Actually I've tried to fix a python package (.py) directly in spyder, it hanged upon saving, instead than rejected the save without administrator rights.
[Window, last master, Python 3.6]

steff456:
## Description of Changes

Open the protected files in Windows by verifying if they have only read access, so they cannot be modified inside spyder and saved.

### Issue(s) Resolved

Fixes #10657 

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/editor.py
+++ b/spyder/plugins/editor/widgets/editor.py
@@ -2329,14 +2329,15 @@ class EditorStack(QWidget):
             if not osp.isfile(finfo.filename):
                 # This is an 'untitledX.py' file (newly created)
                 read_only = False
-            try:
-                # Try to open the file to see if the permissions allow writing
-                # in Windows. For further info, see issue
-                # https://github.com/spyder-ide/spyder/issues/10657
-                fd = open(finfo.filename, os.O_RDWR)
-                fd.close()
-            except PermissionError:
-                read_only = True
+            elif os.name == 'nt':
+                try:
+                    # Try to open the file to see if its permissions allow
+                    # to write on it
+                    # Fixes spyder-ide/spyder#10657
+                    fd = os.open(finfo.filename, os.O_RDWR)
+                    os.close(fd)
+                except (IOError, OSError):
+                    read_only = True
             finfo.editor.setReadOnly(read_only)
             self.readonly_changed.emit(read_only)
 
--- a/spyder/plugins/editor/widgets/editor.py
+++ b/spyder/plugins/editor/widgets/editor.py
@@ -2329,6 +2329,14 @@ class EditorStack(QWidget):
             if not osp.isfile(finfo.filename):
                 # This is an 'untitledX.py' file (newly created)
                 read_only = False
+            try:
+                # Try to open the file to see if the permissions allow writing
+                # in Windows. For further info, see issue
+                # https://github.com/spyder-ide/spyder/issues/10657
+                fd = open(finfo.filename, os.O_RDWR)
+                fd.close()
+            except PermissionError:
+                read_only = True
             finfo.editor.setReadOnly(read_only)
             self.readonly_changed.emit(read_only)
 


spyder/plugins/editor/widgets/editor.py
```diff
@@ -2329,6 +2329,14 @@ def __refresh_readonly(self, index):
             if not osp.isfile(finfo.filename):
                 # This is an 'untitledX.py' file (newly created)
                 read_only = False
+            try:
+                # Try to open the file to see if the permissions allow writing
+                # in Windows. For further info, see issue
+                # https://github.com/spyder-ide/spyder/issues/10657
+                fd = open(finfo.filename, os.O_RDWR)
+                fd.close()
+            except PermissionError:
```
ccordoba12: `PermissionError` doesn't exist in Python 2. I think in Python 3 it inherits from `OSError`, so that's the error we need to capture here.

spyder/plugins/editor/widgets/editor.py
```diff
@@ -2329,6 +2329,14 @@ def __refresh_readonly(self, index):
             if not osp.isfile(finfo.filename):
                 # This is an 'untitledX.py' file (newly created)
                 read_only = False
+            try:
+                # Try to open the file to see if the permissions allow writing
+                # in Windows. For further info, see issue
+                # https://github.com/spyder-ide/spyder/issues/10657
```
ccordoba12: Please change this to

    Fixes spyder-ide/spyder#10657

spyder/plugins/editor/widgets/editor.py
```diff
@@ -2329,6 +2329,14 @@ def __refresh_readonly(self, index):
             if not osp.isfile(finfo.filename):
                 # This is an 'untitledX.py' file (newly created)
                 read_only = False
+            try:
+                # Try to open the file to see if the permissions allow writing
+                # in Windows. For further info, see issue
```
ccordoba12: Please remove this line completely.

spyder/plugins/editor/widgets/editor.py
```diff
@@ -2329,6 +2329,14 @@ def __refresh_readonly(self, index):
             if not osp.isfile(finfo.filename):
                 # This is an 'untitledX.py' file (newly created)
                 read_only = False
+            try:
```
ccordoba12: This try/except should only apply to Windows, right @dalthviz?

spyder/plugins/editor/widgets/editor.py
```diff
@@ -2329,6 +2329,14 @@ def __refresh_readonly(self, index):
             if not osp.isfile(finfo.filename):
                 # This is an 'untitledX.py' file (newly created)
                 read_only = False
+            try:
+                # Try to open the file to see if the permissions allow writing
```
ccordoba12: Please change this to

     # Try to open the file to see if its permissions allow to write on it

spyder/plugins/editor/widgets/editor.py
dalthviz: Yep @ccordoba12 also I think this validation needs to be done only when the file isn't an untitled one

spyder/plugins/editor/widgets/editor.py
```diff
@@ -2329,6 +2329,14 @@ def __refresh_readonly(self, index):
             if not osp.isfile(finfo.filename):
                 # This is an 'untitledX.py' file (newly created)
                 read_only = False
+            try:
+                # Try to open the file to see if the permissions allow writing
+                # in Windows. For further info, see issue
+                # https://github.com/spyder-ide/spyder/issues/10657
+                fd = open(finfo.filename, os.O_RDWR)
+                fd.close()
```
dalthviz: @steff456 here the `open`/`close` that we want to call is from `os` not the built-in one:
```suggestion
                fd = os.open(finfo.filename, os.O_RDWR)
                os.close(fd)
```

dalthviz: This why the tests are failing with a message of `open` not accepting `int` type as second parameter  (`TypeError: open() argument 2 must be str, not int`)

ccordoba12:
@dalthviz, please test this PR manually to see if it fixes the problem, and approve it if that's the case.

PR #11769, opened 2020-03-11
PR: Update macOS image used in Azure

ccordoba12:
It seems Azure was not being triggered due to that.


ccordoba12:
Closing because this is not the solution.

PR #11786, opened 2020-03-12
PR: Use Miniconda 4.7 in Travis because 4.8 consumes too much memory

ccordoba12:
There are some tests that are failing due to lack of memory in Travis.


ccordoba12:
This was not necessary at the end.

PR #11824, opened 2020-03-17
PR: Correctly hide already complete word in completion widget

ElieGouzien:
##  Description of Changes

This pull request contains two commits:
* e1b39295a **corrects** a bug in #11732: now the completion widget is actually hidden when no completion is available.
* 5a51edc2b **modifies** the behavior to match these described in  https://github.com/spyder-ide/spyder/issues/11600#issuecomment-591374070. This is obviously a personal choice and I would understand if you prefer previous behavior.
* 0813477d6 **adds** a regression test for this PR.

Result can be seen (with only letter keys stroke):
Peek 17-03-2020 14-32

I did not include new test since the one introduced by #11732 should be sufficient (even if it doesn't test everything).

### Issue(s) Resolved

Should fix better #11600
Correct #11732

### Affirmation

ElieGouzien

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -1185,9 +1185,8 @@ class CodeEditor(TextEditBaseWidget):
                         completion['insertText'] = insert_text
                         del completion['textEdit']
 
-            if len(completion_list) > 0:
-                self.completion_widget.show_list(
-                        completion_list, position, automatic)
+            self.completion_widget.show_list(
+                completion_list, position, automatic)
 
             self.kite_call_to_action.handle_processed_completions(completions)
         except RuntimeError:
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -1141,8 +1141,12 @@ class CodeEditor(TextEditBaseWidget):
             prefix = self.get_current_word()
             completions = ([] if completions is None else
                            [completion for completion in completions
-                            if completion.get('insertText') != prefix
+                            if completion.get('insertText')
                             or completion.get('textEdit', {}).get('newText')])
+            if (len(completions) == 1
+                    and completions[0].get('insertText') == prefix
+                    and not completions[0].get('textEdit', {}).get('newText')):
+                completions.pop()
 
             replace_end = self.textCursor().position()
             under_cursor = self.get_current_word_and_position(completion=True)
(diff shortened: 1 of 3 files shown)


ElieGouzien:
For comparison, as of spyder==4.1.0 (with linux), behavior was:
Peek 17-03-2020 14-40

goanpeca:
@ElieGouzien Thanks for working on this. I like this change. However it seems some tests are now failing.

ElieGouzien:
> However it seems some tests are now failing.

I've rebased on more recent 4.x version (not sure it changes anything).

Problem seems to come from `test_kite_textEdit_completions` in `spyder/plugins/editor/widgets/tests/test_introspection.py`. Since I don't have Kite installed and don't know anything of qutest it's very hard to find out what is the problem. Could anyone help me on that ?

ElieGouzien:
@goanpeca: I think I managed to fix the problem.

There are still builds failing, but they fail at the CondaEnvironment stage, so I don't think it's liked with my pull request.

ElieGouzien:
I finally added a new test to check that in the example above:
  * when typing `some`, `some` *and* `something` are proposed
  * when typing `something` the completion widget is hidden.

goanpeca:
Great news! Thanks will test this later today!

ElieGouzien:
I've only re-based and tests now pass, except one that seems to randomly fail also on 4.x and don't seems to be linked with my changes.

ccordoba12:
@ElieGouzien, you have two options: either skip your test on Windows (that's fine because it'll be tested in the other platforms) or try to move it to a different localtion in `test_introspection.py`, to see if it doesn't introduce the failure in the other test.

ElieGouzien:
@ccordoba12: the test I added is passing fine, even under windows. The problem comes from another test, which seems to randomly fail (I encourage you to follow the links I posted).

It might be linked with a timeout chosen too tight. If you want I can try to change it, but I don't think it should be done within this pull request since the problem already exist in branch `4.x`.

ElieGouzien:
I made #11916 to see if it work better with more time where it fails.

If you accept this new pull request, I will rebase this one, hopping the problem will be solved.

ccordoba12:
> the test I added is passing fine, even under windows. The problem comes from another test, which seems to randomly fail (I encourage you to follow the links I posted).

I know what test is failing, but it only fails the first time and then pass the second one (we run our test suite several times due to segfaults and problems like this one).

I think the solution is simply to put `test_automatic_completions_hide_complete` at the end of the file, or simply after `test_completions` so it doesn't interfere with it.

ElieGouzien:
@ccordoba12 thanks for the advice !

I finally had to move the test to a new file to make the other ones pass (even ignoring was not enough).

PR #11826, opened 2020-03-17
PR: Fix getting help from the console for dot objects and remove redundant info from completion hints

### LINKED ISSUE #11821: Completion doc widget and editor/console to help improvements
- [x]  `Ctrl + I` not working for elements with `.` (for example `np.sin`)
- [x] ~~Doc tooltip from current completion not being updated (for example with `pd.Categorical`)~~ I was unable to reproduce this
- [x] Remove `ufunc` signature from completion doc tooltip (for example with `numpy.sign`, `numpy.sin`, `numpy.signbit`, etc).

dalthviz:
## Description of Changes

### Issue(s) Resolved

Fixes #11821 

### Affirmation

### DIFF
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -459,14 +459,14 @@ def test_get_help_ipython_console_dot_notation(main_window, qtbot, tmpdir):
     webpage = webview.page() if WEBENGINE else webview.page().mainFrame()
 
     # Write function name
-    qtbot.keyClicks(control, u'np.sin')
+    qtbot.keyClicks(control, u'np.linalg.norm')
 
     # Get help
     control.inspect_current_object()
 
     # Check that a expected text is part of the page
     qtbot.waitUntil(
-        lambda: check_text(webpage, "Trigonometric sine, element-wise."),
+        lambda: check_text(webpage, "Matrix or vector norm."),
         timeout=6000)
 
 
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -433,6 +433,45 @@ def test_get_help_combo(main_window, qtbot):
 
 @pytest.mark.slow
 @pytest.mark.skipif(PY2, reason="Invalid definition of function in Python 2.")
+def test_get_help_ipython_console_dot_notation(main_window, qtbot, tmpdir):
+    """
+    Test that Help works when called from the IPython console
+    with dot calls i.e np.sin
+
+    See spyder-ide/spyder#11821
+    """
+    shell = main_window.ipyconsole.get_current_shellwidget()
+    control = shell._control
+    qtbot.waitUntil(lambda: shell._prompt_html is not None,
+                    timeout=SHELL_TIMEOUT)
+
+    # Open test file
+    test_file = osp.join(LOCATION, 'script_unicode.py')
+    main_window.editor.load(test_file)
+    code_editor = main_window.editor.get_focus_widget()
+
+    # Run test file
+    qtbot.keyClick(code_editor, Qt.Key_F5)
+    qtbot.wait(500)
+
+    help_plugin = main_window.help
+    webview = help_plugin.rich_text.webview._webview
+    webpage = webview.page() if WEBENGINE else webview.page().mainFrame()
+
+    # Write function name
+    qtbot.keyClicks(control, u'np.sin')
+
+    # Get help
+    control.inspect_current_object()
+
+    # Check that a expected text is part of the page
+    qtbot.waitUntil(
+        lambda: check_text(webpage, "Trigonometric sine, element-wise."),
+        timeout=6000)
+
+
+@pytest.mark.slow
+@pytest.mark.skipif(PY2, reason="Invalid definition of function in Python 2.")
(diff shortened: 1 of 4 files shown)


spyder/widgets/mixins.py
```diff
@@ -1359,9 +1363,17 @@ def set_help_enabled(self, state):
         self.help_enabled = state
 
     def inspect_current_object(self):
-        text = self.get_current_word(help_req=True)
```
ccordoba12: Is `req=True` still necessary in `get_current_word`? If not, please remove it.

spyder/widgets/mixins.py
```diff
@@ -1359,9 +1363,17 @@ def set_help_enabled(self, state):
         self.help_enabled = state
 
     def inspect_current_object(self):
-        text = self.get_current_word(help_req=True)
-        if text:
-            self.show_object_info(text, force=True)
+        regex_word = ''
+        start_text = self.get_text('sol', 'cursor')
+        tl1 = re.findall(r'([a-zA-Z_]+[^\s^\(]*)', start_text)
```
ccordoba12: Does this capture unicode variables? Please remember that unicode identifiers is valid in Python 3.

spyder/widgets/mixins.py
```diff
@@ -1359,9 +1363,17 @@ def set_help_enabled(self, state):
...
-        text = self.get_current_word(help_req=True)
-        if text:
-            self.show_object_info(text, force=True)
+        regex_word = ''
+        start_text = self.get_text('sol', 'cursor')
+        tl1 = re.findall(r'([a-zA-Z_]+[^\s^\(]*)', start_text)
+        if tl1 and start_text.endswith(tl1[-1]):
+            regex_word += tl1[-1]
+        end_text = self.get_text('cursor', 'eol')
+        tl2 = re.findall(r'([^\s^\(]+[0-9a-zA-Z_\.]*)', end_text)
```
ccordoba12: How many words with dots this regexp can capture? Please remember that we can have things like `np.linalg.norm` too.

Also, perhaps this should only apply to Python and not to other languages (e.g. the equivalent of `.` in R is `$`).

ccordoba12: Also, why can't we use here the current text from the cursor to the left, until an space is found? That seems simpler to me than using regexp's.

spyder/widgets/mixins.py
dalthviz: Yep the `help_req` is needed for the help request made from the `Editor`

spyder/widgets/mixins.py
dalthviz: I tried that approach at first, but then I realized that I needed also to get sometimes characters to the right of the cursor, and I wasn't sure of the logic to follow to get does characters and then mix both (left and right selected texts).

I will recheck that approach and see what is better

spyder/widgets/mixins.py
dalthviz: Will check

spyder/widgets/mixins.py
ccordoba12: Thanks @dalthviz!

spyder/app/tests/test_mainwindow.py
```diff
@@ -431,6 +431,45 @@ def test_get_help_combo(main_window, qtbot):
...
+    # Write function name
+    qtbot.keyClicks(control, u'np.sin')
+
+    # Get help
+    control.inspect_current_object()
+
+    # Check that a expected text is part of the page
+    qtbot.waitUntil(
+        lambda: check_text(webpage, "Trigonometric sine, element-wise."),
+        timeout=6000)
```
ccordoba12: Also, below this, please add a test for the aaʹbb function present in `script_unicode.py`, to test if we can show help for objects that contain unicode chars.

spyder/widgets/mixins.py
```diff
@@ -971,14 +1017,24 @@ def is_space(move):
                 return not to_text_string(curs.selectedText()).strip()
 
             def is_special_character(move):
```
ccordoba12: Please add a docstring to this function. What characters is it supposed to catch?

spyder/widgets/mixins.py
```diff
@@ -971,14 +1017,24 @@ def is_space(move):
...
+                text_cursor = to_text_string(curs.selectedText()).strip()
+                return text_cursor == '.'
+
             if help_req:
-                if is_special_character(QTextCursor.NoMove):
+                if (is_special_character(QTextCursor.NoMove) and
+                        not is_dot(QTextCursor.PreviousCharacter)):
                     cursor.movePosition(QTextCursor.WordLeft)
+                elif is_dot(QTextCursor.PreviousCharacter):
+                    cursor.movePosition(QTextCursor.NextCharacter)
```
ccordoba12: Checking for `dot` here makes things specialized for Python. What's the plan for other languages?

spyder/app/tests/test_mainwindow.py
dalthviz: See `test_mainwindow.test_get_help_ipython_console_special_characters`

ccordoba12: Ok, I haven't noticed that one :+1:

Issue #11819, opened 2020-03-17
Activation of kernel env fails when using anaconda shortcut (FileNotFound error kernel start 4.1.0)


leavor:
## Problem Description

Spyder4.1.0 IPython kernel startup error
Close spyder and restart, the problem persists.
### What steps reproduce the problem?

After upgrading spyder4.0.1 to 4.1.0, IPython kernel startup error occurs when starting spyder.
The following command is used to upgrade in a virtual environment:
conda install spyder=4.1.0
```python-traceback
启动 IPython 内核时发生错误（An error occurred while starting the IPython kernel）
错误：（Error）
Traceback (most recent call last):
File "C:\Users\Admin\Anaconda3\envs\python37\lib\site‑packages\spyder\plugins\ipythonconsole\plugin.py", line 1209, in create_kernel_manager_and_kernel_client
kernel_manager.start_kernel(stderr=stderr_handle, **kwargs)
File "C:\Users\Admin\Anaconda3\envs\python37\lib\site‑packages\jupyter_client\manager.py", line 259, in start_kernel
**kw)
File "C:\Users\Admin\Anaconda3\envs\python37\lib\site‑packages\jupyter_client\manager.py", line 204, in _launch_kernel
return launch_kernel(kernel_cmd, **kw)
File "C:\Users\Admin\Anaconda3\envs\python37\lib\site‑packages\jupyter_client\launcher.py", line 138, in launch_kernel
proc = Popen(cmd, **kwargs)
File "C:\Users\Admin\Anaconda3\envs\python37\lib\subprocess.py", line 800, in __init__
restore_signals, start_new_session)
File "C:\Users\Admin\Anaconda3\envs\python37\lib\subprocess.py", line 1207, in _execute_child
startupinfo)
FileNotFoundError: [WinError 2] 系统找不到指定的文件（The system can not find the file specified）。
```

## Versions

* Spyder version: 4.1.0
* Python version: Python3.7.5
* Qt version: 5.9.6
* PyQt version: 5.9.2
* Operating System name/version: win10

leavor:
Roll back to the 4.0.1 version, there is no error when starting the spyder.

ccordoba12:
@dalthviz, please try to reproduce this one by installing Spyder in a conda env and trying to start from it.

leavor:
Create the new Python virtual environment with the following command, and install spyder4.1.0 at the same time, the same error occurs

[...]

PR #11836, opened 2020-03-17
PR: Improve Python executable validation in kernelspec

dalthviz:
## Description of Changes

Change comparison between selected python executable and `sys.executable` since the executable could be of the form `python/python.exe` or `pythonw/pythonw.exe` but chairing the same path

### Issue(s) Resolved

Part of #11819 

### Affirmation

### DIFF
--- a/spyder/plugins/ipythonconsole/utils/kernelspec.py
+++ b/spyder/plugins/ipythonconsole/utils/kernelspec.py
@@ -35,7 +35,7 @@ logger = logging.getLogger(__name__)
 
 def is_different_interpreter(pyexec):
     """Check that pyexec is a different interpreter from sys.executable."""
-    executable_validation = 'python' in osp.basename(pyexec)
+    executable_validation = osp.basename(pyexec).startswith('python')
     directory_validation = osp.dirname(pyexec) != osp.dirname(sys.executable)
     return directory_validation and executable_validation
 
--- a/spyder/plugins/ipythonconsole/utils/kernelspec.py
+++ b/spyder/plugins/ipythonconsole/utils/kernelspec.py
@@ -95,7 +95,7 @@ class SpyderKernelSpec(KernelSpec):
                 CONF.set('main_interpreter', 'custom', False)
 
         # Part of spyder-ide/spyder#11819
-        is_different_interpreter = is_different_interpreter(pyexec)
+        is_different = is_different_interpreter(pyexec)
 
         # Fixes spyder-ide/spyder#3427.
         if os.name == 'nt':
@@ -105,7 +105,7 @@ class SpyderKernelSpec(KernelSpec):
                 pyexec = pyexec_w
 
         # Command used to start kernels
-        if is_different_interpreter and is_conda_env(pyexec=pyexec):
+        if is_different and is_conda_env(pyexec=pyexec):
             # If this is a conda environment we need to call an intermediate
             # activation script to correctly activate the spyder-kernel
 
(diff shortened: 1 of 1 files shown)


spyder/plugins/ipythonconsole/utils/kernelspec.py
```diff
@@ -33,6 +33,13 @@
 logger = logging.getLogger(__name__)
 
 
+def validate_interpreter(pyexec):
+    """Check python excutable path and name."""
+    executable_validation = 'python' in osp.basename(pyexec)
```
goanpeca: Maybe it is better to use startswith( to ensure it starts with python

spyder/plugins/ipythonconsole/utils/kernelspec.py
```diff
@@ -87,7 +94,8 @@ def argv(self):
                 CONF.set('main_interpreter', 'default', True)
                 CONF.set('main_interpreter', 'custom', False)
 
-        is_different_interpreter = pyexec != sys.executable
+        # Part of spyder-ide/spyder#11819
+        is_different_interpreter = validate_interpreter(pyexec)
```
goanpeca: Is this correct? Should it ve Not ?

dalthviz: This is like doing:
```
is_different_interpreter = (
    osp.dirname(pyexec) != osp.dirname(sys.executable) and 'python' in osp.basename(pyexec))
```
I think this gives the correct value but maybe I'm wrong @ccordoba12 @goanpeca ?

ccordoba12:
@dalthviz, please rebase on top of the latest 4.x to get the fixes to our tests.

spyder/plugins/ipythonconsole/utils/kernelspec.py
```diff
@@ -33,6 +33,13 @@
 logger = logging.getLogger(__name__)
 
 
+def validate_interpreter(pyexec):
```
ccordoba12: Let's call this: `is_different_interpreter`.

spyder/plugins/ipythonconsole/utils/kernelspec.py
```diff
@@ -33,6 +33,13 @@
...
+    """Check python excutable path and name."""
```
ccordoba12: Please change this to:

    Check that pyexec is a different interpreter from sys.executable

spyder/plugins/ipythonconsole/utils/kernelspec.py
ccordoba12: It's fine as it is (that's how it was done before).

spyder/plugins/ipythonconsole/utils/kernelspec.py
goanpeca: Why did you mark this as solved @ccordoba12 ?

`executable_validation = 'python' in osp.basename(pyexec)`

Should be

`executable_validation = osp.basename(pyexec).stratswith('python')`

dalthviz: Since I rebased I think the initial commit where I did the change was removed. I re-did the change in the last commit

spyder/plugins/ipythonconsole/utils/kernelspec.py
```diff
@@ -87,7 +94,8 @@ def argv(self):
                 CONF.set('main_interpreter', 'default', True)
                 CONF.set('main_interpreter', 'custom', False)
 
-        is_different_interpreter = pyexec != sys.executable
+        # Part of spyder-ide/spyder#11819
+        is_different_interpreter = is_different_interpreter(pyexec)
```
ccordoba12: Perhaps the variable on the left shouldn't have the same name as the new function to avoid the odd errors generated in Travis?

So let's change this to

    is_different = is_different_interpreter(pyexec)

And also change the usage of `is_different_interpreter` below.

PR #11977, opened 2020-03-25
PR: Fix German translation of "preferences"

bnavigator:
## Description of Changes

* [ ] Wrote at least one-line docstrings (for any new functions)
* [ ] Added unit test(s) covering the changes (if testable)
* [ ] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

Change the German translation of "preferences" from "Voreinstellungen" (="defaults") to "Einstellungen". This is the correct translation commonly used in other software.

If this is not the correct way of changing translations, please advice to sign up and approve a translator role in crowdin or just change it there on your own.

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct: bnavigator

PR #11982, opened 2020-03-25
PR: Make a string translatable

bnavigator:
The configuration page for `Advanced settings` in the Editor plugin had an untranslated label.

I did not regenerate the strings yet, because I am not familiar with the workflow in and out of Crowdin. (see #11977)

German translation would be "Typ:"


ccordoba12:
> I did not regenerate the strings yet

That will be done by @goanpeca later this week. Please don't do it yourself in case you submit more PRs like this in the future.

PR #11991, opened 2020-03-26
PR: Check if block is valid before painting it in the scrollflag panel

impact27:
## Description of Changes
Check if block is valid before painting it and reduce code repetition.

### Issue(s) Resolved

Fixes #

### Affirmation

### DIFF
--- a/spyder/plugins/editor/panels/scrollflag.py
+++ b/spyder/plugins/editor/panels/scrollflag.py
@@ -205,11 +205,19 @@ class ScrollFlagArea(Panel):
 
                 return ceil(middle-self.FLAGS_DY/2)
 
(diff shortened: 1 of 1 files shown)


ccordoba12:
@impact27, could you explain why this change is necessary?

impact27:
> @impact27, could you explain why this change is necessary?

I got a crash because spyder tried to paint a non-existing block. The blocks are saved and used at a later point, so it is conceivable that they are deleted in the meantime. 

I couldn't reproduce after pulling the latest master but I think it is good to make sure the block is still valid at that point. This should be true at any point in the code where we store blocks and access them at a later point.

ccordoba12:
Ok, thanks for the explanation.

PR #12055, opened 2020-03-31
PR: Correctly unindent code for valid Python

### LINKED ISSUE #10148: Editor unindents when typing :

Enter the following in an editor (without pressing enter after typing 'else'):

    def foo(x):
        if x == 1:
        else

To do that you would have had to press backspace before typing 'else' because the editor automatically indents the cursor to col 8 after hitting return on the 'if' line. This is all correct behavior.

Now, with the cursor at the end of else, type a colon. At this point the editor unindents the else, and you have

    def foo(x):
        if x == 1:
    else:

In other words, the else is unindented to column 1, when it should be lined up with the if, like so:

    def foo(x):
        if x == 1:
        else:

3.x correctly (IMO) doesn't unindent the else; it should be okay to enter your code in a not entirely linear fashion (in this case I was thinking more about how to implement the logic I needed for the else part, and so I just wanted to ignore coding the case for x==1 for the moment)

## Versions

* Spyder version: 4.0.0b4 
* Python version: 3.7.3
* Qt version: 5.9.6
* PyQt5 version: 5.9.2
* Operating System: Windows 7

### Dependencies

```
qdarkstyle >=2.6.4:  2.7 (OK)
sympy >=0.7.3     :  1.4 (OK)
```

remisalmon:
## Description of Changes

Does not fix the original issue where for ex. a `if` statement is immediately followed by a `else` statement (not valid Python)

To fix the original issue, `if ind(leading_text) == ind(prevtxt.rstrip()) and not prevtxt[-1] == ':':` works but is unnecessary in my opinion

### Issue(s) Resolved

Fixes #10148

### Affirmation

I certify the above statement is true and correct: remisalmon

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -3982,7 +3982,7 @@ class CodeEditor(TextEditBaseWidget):
                 ind = lambda txt: len(txt)-len(txt.lstrip())
                 prevtxt = to_text_string(self.textCursor(
                                                 ).block().previous().text())
-                if ind(leading_text) == ind(prevtxt):
+                if ind(leading_text) == ind(prevtxt.rstrip()):
                     self.unindent(force=True)
             insert_text(event)
         elif key == Qt.Key_Space and not shift and not ctrl \
@@ -3993,7 +3993,7 @@ class CodeEditor(TextEditBaseWidget):
                 ind = lambda txt: len(txt)-len(txt.lstrip())
                 prevtxt = to_text_string(self.textCursor(
                                                 ).block().previous().text())
-                if ind(leading_text) == ind(prevtxt):
+                if ind(leading_text) == ind(prevtxt.rstrip()):
                     self.unindent(force=True)
             insert_text(event)
         elif key == Qt.Key_Tab and not ctrl:
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -3982,7 +3982,9 @@ class CodeEditor(TextEditBaseWidget):
                 ind = lambda txt: len(txt)-len(txt.lstrip())
                 prevtxt = to_text_string(self.textCursor(
                                                 ).block().previous().text())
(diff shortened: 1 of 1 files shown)


ccordoba12:
Thanks a lot for your contribution! We'd like to include it in our next release in the Spyder 4 series. For that, please make your branch to derive from our `4.x` branch (instead of master), with the following commands:

    git checkout 4.x
    git checkout fix_unindents
    git rebase --onto 4.x master fix_unindents
    git push -f origin fix_unindents

If you decline to do that, your work will be part of Spyder 5, to be released at the end of 2020 or in 2021.

remisalmon:
Just rebased on `4.x`

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3903,7 +3903,7 @@ def insert_text(event):
                 ind = lambda txt: len(txt)-len(txt.lstrip())
```
steff456, line 3903: We should be checking that this only occurs in python with `self.language`

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3914,7 +3914,7 @@ def insert_text(event):
...
```
steff456, line 3914: Same in here, this should only happen in python

steff456:
Hi @remisalmon, do you think you can make the small changes today?

remisalmon:
> Hi @remisalmon, do you think you can make the small changes today?

What changes? Sorry I don't get your comments about ind() and how they relate to my PR.

ccordoba12:
@remisalmon, what @steff456 asked you to do is to apply your changes only when the language of the file you're editing is Python.

If you don't understand very well how to do that, @steff456 could give you a hand with it.

remisalmon:
@ccordoba12 thanks, I got confused by Github highlighting that lambda function in the review above
@steff456 I have pushed the changes you requested, I hope that works!

ccordoba12:
Please also rebase this PR so you can get the latest fixes to our tests.

ccordoba12:
Thanks @remisalmon for your help with this one and @steff456 for the review!

PR #12134, opened 2020-04-04
PR: Use '!' for Pdb commands and add other options to control the debugger (IPython console)

### LINKED ISSUE #620: Unable to see plots made with Matplotlib while debugging
_From [al.dan...@gmail.com](https://code.google.com/u/100487118575071222856/) on 2011-04-12T17:33:50Z_

This script, when run within spyder, works fine and produces a contour plot:

 #!/usr/bin/env python
 from   matplotlib.pyplot import imshow, show
 import numpy as np
 x  = np.random.rand(4,5)
 imshow(x)
 show()

However, if I enter debug mode and put a breakpoint at the imshow(x) line then manually type the imshow(x) and show() commands at the (Pdb) prompt, no plot appears.  I only get a reference to the plot object:

 (Pdb) imshow(x)
 <matplotlib.image.AxesImage object at 0x183f5150>
 (Pdb) show()
 (Pdb) 

Would be a nice enhancement to be able to plot variables in debug mode.

Python 2.7.1
NumPy  1.5.1
matplotlib 1.0.1  (WXAgg backend)
RHEL 4.8 x86_64

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=620_

impact27:
## Description of Changes

Use '!' before Pdb commands and nothing before python code, as is the case in IPython

### Issue(s) Resolved

Depends on spyder-ide/spyder-kernels#214

### Affirmation


impact27:
The circleci need to be rerun

impact27:
I clean up the code a bit so sorry if it is a bit difficult to review.

impact27:
I switched commands and text because I am tired to have a variable named `step` that I can't use while debugging

so:
```
ipdb> step = 10

ipdb> step
10

ipdb> !step
> /usr/local/Cellar/python/3...
```

spyder/plugins/editor/plugin.py
```diff
@@ -537,6 +537,12 @@ def get_plugin_actions(self):
         pdb_execute_events_action.setChecked(
             CONF.get('run', 'pdb_execute_events'))
 
+        pdb_use_exclamation_mark_action = create_action(
+            self, _("Use exclamation mark for pdb commands"),
```
ccordoba12, line 541: pdb -> Pdb

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -13,18 +13,37 @@
...
+        0, (r'(!)(\w+)(.*\n)', bygroups(Operator, Keyword, Text)))
+
+
+class SpyderIPy2Lexer(IPythonLexer):
+    tokens = IPython3Lexer.tokens
+    tokens['root'].insert(
+        0, (r'(!)(\w+)(.*\n)', bygroups(Operator, Keyword, Text)))
+
+
+
```
ccordoba12, line 46: Please remove one blank line here (pep8 only requires two blank lines between classes).

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -301,6 +343,14 @@ def set_pdb_execute_events(self):
         self.call_kernel(interrupt=True).set_pdb_execute_events(
             CONF.get('run', 'pdb_execute_events', True))
 
+    def set_pdb_use_exclamation_mark(self):
+        """Set pdb_use_exclamation_mark into a debugging session"""
+        self.call_kernel(interrupt=True).set_pdb_use_exclamation_mark(
+            self.is_pdb_using_exclamantion_mark())
+
+    def is_pdb_using_exclamantion_mark(self):
+        return CONF.get('run', 'pdb_use_exclamation_mark', False)
```
ccordoba12, line 352: There's no need to pass `False` here as a default value because `pdb_use_exclamation_mark` is declared in our config system.

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -368,6 +418,39 @@ def reset(self, clear=False):
...
+            prompt = self._current_prompt()
+        else:
+            self._pdb_single_letter_mode = True
+            prompt = self._pdb_single_letter_prompt
+        if self._reading:
+            self._update_pdb_prompt(prompt)
+            self._show_prompt(prompt)
+
+    def _current_prompt(self):
+        return "pdb [{}]: ".format(self._pdb_history_input_number + 1)
```
ccordoba12, line 435: My suggestions for the Pdb prompt are `IPdb [n]:` or `In (dbg) [n]:` (I think I'm more in favor of the first one).

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -511,6 +602,21 @@ def _show_prompt(self, prompt=None, html=False, newline=True):
...
         if self.is_waiting_pdb_input():
+            if (self._reading and self._pdb_single_letter_mode and
+                    event.modifiers() in (Qt.NoModifier, Qt.ShiftModifier)):
+                key = event.text()
+                if len(key) > 0 and key.isprintable():
+                    if key not in (
+                            'a', 'b', 'c', 'd', 'h', 'l',
+                            'n', 'q', 'r', 's', 'u', 'w'):
+                        self._append_plain_text(
+                            "** Error: '" + key + "' is not a pdb command.\n",
```
ccordoba12, line 613: pdb -> Pdb

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -511,6 +602,21 @@ def _show_prompt(self, prompt=None, html=False, newline=True):
...
+                            before_prompt=True)
+                    else:
+                        self.pdb_execute(self._pdb_cmd_prefix() + key)
+                elif event.key() in (Qt.Key_Escape, Qt.Key_Return, Qt.Key_Enter):
+                    self._toggle_pdb_single_letter_mode()
```
ccordoba12, line 618: This seems handy bur it could cause confusions for our users, i.e. that by pressing these keys (especially Enter) you can switch between single and non-single letter modes.

What if we leave only `Esc` to switch between these modes? That's like a switching between command and edit mode in the notebook, so perhaps people wouldn't be too surprised about it.

ccordoba12 (commented):
This looks quite promising @impact27! I'm really looking forward to see it finished!

I left an initial, simple review for you to take a look at.

spyder/plugins/ipythonconsole/widgets/debugging.py
impact27: The problem is that you have a prompt where you only can enter a subset of commands, so it is important to be able to leave it quickly to test something. The use case I have in mind is to use `(n)ext`, `(s)tep` and other "moving" commands until I reach a frame I want to analyse more, at which point I can quickly leave the single letter mode. "Enter" means I want to start typing something, so it is feels quite natural.

impact27:
I am not sure that the failure is related to this PR (`spyder/plugins/variableexplorer/widgets/tests/test_collectioneditor.py::test_sort_numpy_numeric_collectionsmodel`)

spyder/plugins/ipythonconsole/widgets/debugging.py
ccordoba12: > The problem is that you have a prompt where you only can enter a subset of commands

Why is that the case? I mean, why is it not possible to enter all Pdb command prefixed by `!`?

impact27:
I fix the tests,  #13892, and #13891 here

spyder/plugins/ipythonconsole/widgets/debugging.py
impact27: The idea is that you have the regular mode where you can do whatever you need, and the single letter mode that lets you very quickly explore the code. So instead of typing "!" + "s" + "<Enter>", sou can just type "s" to step through the code. When you reach a frame you want to analyse, you want to exit quickly. Because you r finger is likely already above the "Enter" key because you enter a lot of short commands, it makes sense to be able ot use it to enter the regular mode whee it is useful.

ccordoba12: > Because you r finger is likely already above the "Enter" key because you enter a lot of short commands, it makes sense to be able ot use it to enter the regular mode whee it is useful.

I understand your point, but since `Enter` is primarily used for execution, associating this additional functionality to it could be very confusing. That's why I think we should only use `Esc` to switch between these modes.

impact27: I moved that to a setting that is disabled by default if that is ok?

impact27:
I wonder if the prompt for the debugger is too similar to the regular prompt. Maybe they should be more distinct?

impact27:
I have added parenthesis to separate:
IPython: `In [1]:`
pdb : `(IPdb [1]):`
recursive debugger: `((IPdb [1])):`

ccordoba12:
Hey @impact27, could you rebase this PR on 4.x once I merge PR #13902? The thing is I cherry-picked it your commit fd36686 (`Fix test collectioneditor`) there to fix our tests (thanks for fixing them, by the way)!

ccordoba12:
> I have added parenthesis to separate:
IPython: In [1]:
pdb : (IPdb [1]):

I don't agree with this. `IPdb [n]` is sufficiently different from `In [n]` to guarantee that no confusion would arise among our users between those two prompts.

I agree with modifying the prompt for the recursive debugger though, so the first one of those should show `(IPdb [n]):`.

spyder/plugins/ipythonconsole/widgets/debugging.py
ccordoba12: Yes, that's ok for me. Thanks for doing it!

spyder/plugins/ipythonconsole/confpage.py
```diff
@@ -255,6 +238,69 @@ def setup_page(self):
...
+            _("Ignore Python libraries while debugging"), 'pdb_ignore_lib',
+            tip=_("This option lets you decide if the debugger should "
+                  "ignore the system libraries while debugging."))
+        debug_layout.addWidget(libraries_box)
+
+        execute_events_box = newcb(
+            _("Process execute events while debugging"), 'pdb_execute_events',
+            tip=_("This option lets you decide if the debugger should "
+                  "process the 'execute events' after each prompt, such as "
+                  "matplotlib 'show'."))
```
ccordoba12, line 287:
```suggestion
                  "matplotlib 'show' command."))
```

spyder/plugins/ipythonconsole/confpage.py
```diff
@@ -255,6 +238,69 @@ def setup_page(self):
...
+
+        prevent_closing_box = newcb(
+            _("Prevent editor from closing files while debugging"),
+            'pdb_prevent_closing',
+            tip=_("This option prevents the user from closing a file while"
+                  " it is debugged."))
+        debug_layout.addWidget(prevent_closing_box)
+
+        continue_box = newcb(
+            _("Stop debugging on first line of files without breakpoints"), 'pdb_stop_first_line',
```
ccordoba12, line 271: This option is new, right? Could you explain a bit more why you added it and how it works?

spyder/plugins/ipythonconsole/confpage.py
```diff
@@ -255,6 +238,69 @@ def setup_page(self):
...
+            tip=_("This option lets you decide if the debugger should "
+                  "process the 'execute events' after each prompt, such as "
+                  "matplotlib 'show'."))
+        debug_layout.addWidget(execute_events_box)
+
+        exclamation_mark_box = newcb(
+            _("Use exclamation mark prefix for Pdb commands"), 'pdb_use_exclamation_mark',
+            tip=_("This option lets you decide if the Pdb commands should "
+                  "be prefixed by an excalamtion mark. This helps in separating Pdb "
+                  "commands from python code."))
```
ccordoba12, line 294:
```suggestion
                  "be prefixed by an exclamation mark. This helps in separating Pdb "
                  "commands from Python code."))
```

spyder/plugins/ipythonconsole/confpage.py
```diff
@@ -255,6 +238,69 @@ def setup_page(self):
...
+        debug_layout.addWidget(exclamation_mark_box)
+
+        single_letter_enter_box = newcb(
+            _("Use enter to leave explore mode."), 'pdb_single_letter_enter',
+            tip=_("This option lets you stop the explore mode by pressing enter."))
```
ccordoba12, line 299: I really don't understand the idea behind this new "Explore mode". Is "Explore mode" not the same as Pdb without exclamations?

spyder/plugins/ipythonconsole/confpage.py
```diff
@@ -391,9 +437,10 @@ def setup_page(self):
                                     source_code_group), _("Display"))
         tabs.addTab(self.create_tab(pylab_group, backend_group, inline_group),
                     _("Graphics"))
-        tabs.addTab(self.create_tab(run_lines_group, pdb_run_lines_group,
-                                    run_file_group),
+        tabs.addTab(self.create_tab(run_lines_group, run_file_group),
                     _("Startup"))
+        tabs.addTab(self.create_tab(debug_group, pdb_run_lines_group),
+                    _("Debug"))
```
ccordoba12, line 443:
```suggestion
                    _("Debugger"))
```

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -202,6 +202,9 @@ def apply_plugin_settings(self, options):
                 client.reset_warning = reset_namespace_o
             if ask_before_restart_n in options:
                 client.ask_before_restart = ask_before_restart_o
+            client.shellwidget.set_pdb_ignore_lib()
+            client.shellwidget.set_pdb_execute_events()
+            client.shellwidget.set_pdb_use_exclamation_mark()
```
ccordoba12, line 207: Since you moved the settings for these options to the IPython console, you need to apply them like this now:

```python
...
pdb_ignore_lib_n = 'pdb_ignore_lib'
pdb_ignore_lib_o = self.get_option(pdb_ignore_lib_n)
for client in self.clients:
   ...
   if pdb_ignore_lib_n in options:
      client.shellwidget.set_pdb_ignore_lib(pdb_ignore_lib_o)
```

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -13,18 +13,37 @@
 import pdb
 import re
 
+from qtpy.QtCore import Qt
```
ccordoba12, line 16: Please move this line to be below this one:

    from qtconsole.rich_jupyter_widget import RichJupyterWidget

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -169,6 +193,8 @@ def __init__(self, *args, **kwargs):
         self._pdb_input_queue = []  # List of (code, hidden, echo_stack_entry)
         # Temporary flags
         self._tmp_reading = False
+        self._pdb_single_letter_mode = False
+        self._pdb_single_letter_prompt = "explore mode: "
```
ccordoba12, line 197: I don't like showing a different prompt for this. What about showing `IPdb [n]*:` to signal that something changed in the way we're evaluating things?

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -294,12 +338,20 @@ def set_spyder_breakpoints(self):
     def set_pdb_ignore_lib(self):
         """Set pdb_ignore_lib into a debugging session"""
         self.call_kernel(interrupt=True).set_pdb_ignore_lib(
-            CONF.get('run', 'pdb_ignore_lib', False))
+            CONF.get('ipython_console', 'pdb_ignore_lib', False))
```
ccordoba12, line 341: You need to rewrite this method as it appears below, so you can apply our Preferences in the way I mentioned above:

```python
def set_pdb_ignore_lib(self, state):
    """Set pdb_ignore_lib into a debugging session"""
    self.call_kernel(interrupt=True).set_pdb_ignore_lib(state)
```

ccordoba12: The same goes for `set_pdb_execute_events` and `set_pdb_use_exclamation_mark` below.

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -494,8 +494,17 @@ def create_shortcuts(self):
...
+            self._toggle_pdb_single_letter_mode,
+            context='ipython_console',
+            name='pdb single letter',
+            parent=self)
+
+        return [
+            inspect, clear_console, restart_kernel, new_tab,
+            reset_namespace, array_inline, array_table, clear_line,
+            pdb_single_letter
+            ]
```
ccordoba12, line 507: Please unindent this four spaces.

ccordoba12 (commented):
Thanks @impact27 for your continued work on this one! I think it's close to be merged.

ccordoba12:
> I don't agree with this. `IPdb [n]` is sufficiently different from `In [n]` to guarantee that no confusion would arise among our users between those two prompts.
>
> I agree with modifying the prompt for the recursive debugger though, so the first one of those should show `(IPdb [n]):`.

According to the code I just reviewed, it seems you implemented things as I described them above. Sorry for the confusion if that's the case.

impact27:
> > I don't agree with this. `IPdb [n]` is sufficiently different from `In [n]` to guarantee that no confusion would arise among our users between those two prompts.
> > I agree with modifying the prompt for the recursive debugger though, so the first one of those should show `(IPdb [n]):`.
> 
> According to the code I just reviewed, it seems you implemented things as I described them above. Sorry for the confusion if that's the case.

I changed that in commit d4f953a

spyder/plugins/ipythonconsole/confpage.py
impact27: Ha yes - I wanted to write a comment about that. Because I am adding a Debugging settings tab, I added two small tweaks I am using locally but didn't have time to make into proper PR:
 - `pdb_stop_first_line`: The problem I have is that the debugger will stop on the first line if there is no breakpoints. Because it is annoying to always have to press continue, spyder will automatically do it if a breakpoint is detected. The problem I have is that it doesn't do that if no breakpoints are detected, event if the code will call other files where I have breakpoints. So I added an option to always do continue, so the user is responsible to add breakpoins before debugging.
 - `pdb_prevent_closing`: This allows me to close a file in the editor even if I am debugging it. I am quite frustrated every time spyder doen't let me close a file I asked it to close, so I like this option a lot.

spyder/plugins/ipythonconsole/confpage.py
impact27: So the idea is that when you follow the code execution, you want to be as fast as possible because you might need to step through a lot of code to understand what is happening. Everything I can do to reduce the amount of typing is good in this situation. This is why Pdb has single letter commands. The "Explore mode" removes the need to use "Enter", which divides by two the number of keystrokes needed compared to regular pdb (by 3 if you have to add "!")

impact27: > Is "Explore mode" not the same as Pdb without exclamations?

No because the only thing you can do is the pdb single letter commands. The ones I use most are:
 - (n)ext
 - (s)tep
 - (r)eturn
 - (c)ontinue
 - (u)p
 - (d)own
 - (w)here

The first 4 have a shortcut but I find it more intuitive to use the keyboard key associated with the command rather than a F11-based shortcut

spyder/plugins/ipythonconsole/widgets/debugging.py
impact27: Well it is a pretty fundamental change, We are only accepting a single letter that is executed without the need of pressing enter. I don't think this should look like a prompt really. I was going for something like a vim command mode

impact27:
I removed the explore mode and replaced by https://github.com/spyder-ide/spyder-kernels/pull/214/commits/82617ad15a8e4dc4fd72e1409777db524ffea7db

This means typing "u" in the terminal will go up unless u is defined. So the code takes precedence on the commands rather than the other way around. This way pdb can be used in essentially the current way, but the conflicts resolve in favor of the code rather than the commands, which is more intuitive.

spyder/config/main.py
```diff
@@ -636,4 +636,4 @@
 #    or if you want to *rename* options, then you need to do a MAJOR update in
 #    version, e.g. from 3.0.0 to 4.0.0
 # 3. You don't need to touch this value if you're just adding a new option
-CONF_VERSION = '58.1.0'
+CONF_VERSION = '59.0.0'
```
ccordoba12, line 639: This needs to be bumped to `60.0.0` because I already bumped it to 59 in another PR.

ccordoba12:
> This means typing "u" in the terminal will go up unless u is defined. So the code takes precedence on the commands rather than the other way around. This way pdb can be used in essentially the current way, but the conflicts resolve in favor of the code rather than the commands, which is more intuitive.

Great! That's a very nice and intuitive solution solution!

Thanks @impact27 for implementing it.

ccordoba12 (commented):
Apart from the minor comment I left and fixing the merge conflict with our subrepo, I think this one is ready.

impact27:
Should this behaviour be the default? I think it makes sense as new users will be less surprised but could be a bit surprising for users that are used to pdb.

ccordoba12:
> Should this behaviour be the default?

Yes, I totally agree.

> I think it makes sense as new users will be less surprised but could be a bit surprising for users that are used to pdb.

Agreed too. And what do you think if we give seasoned Pdb users a hint about what's happening, by showing a message like this?

Selección_032

ccordoba12:
@impact27, I think this is good to go now. Please resync our subrepo because I just merged PR spyder-ide/spyder-kernels#214.

ccordoba12:
Please merge with 4.x to fix the merge conflict.

ccordoba12:
I did it myself because I need to base another PR on top of this one. Sorry for the hassle.

ccordoba12 (approved):
Thanks a lot @impact27 for this! Terrific addition!

PR #11741, opened 2020-03-06
PR: Create new API for plugins and widgets

goanpeca:
## Description of Changes

* [x] Wrote at least one-line docstrings (for any new functions)
* [ ] Added unit test(s) covering the changes (if testable)

* [ ] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

### Issue(s) Resolved

Fixes #12180

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct: @goanpeca

PR #12160, opened 2020-04-06
PR: Move Pylint plugin to new API

### LINKED ISSUE #12183: Move Pylint plugin to use new API
- [ ] Move to new API
- [ ] Make sure test pass

goanpeca:
## Description of Changes

Took the opportunity to simplify the buttons so that it is similar to find in files (start/stop are the same button)

pylint

### Issue(s) Resolved

Depends on #11741
Fixes #12183

### Affirmation


juanis2112:
Some comments about this PR:
1. The shortcut F8 is not working for running the code analysis 
2. Clicking on an element in the pane (warning, error, message) doesn't go to the line corresponding to this element. However if I close the file and click the error message in the pane it opens the file (but again it doesn't go to the corresponding line) This was a bug that we found and it was fixed by @ccordoba12  in 4.1.4 so not sure why it is not working here. I tested it in 4x and it works fine.
3. The hamburger menu doesn't have the expand/collapse items that it had before. 
4. When selecting a file in the pane with the "Select python file" button, the analysis doesn't run automatically which it did before. It does display the path of the file selected in the combobox but when pressing analyze, it runs the analysis on the file that is selected in the editor and displays the path of this last file in the combobox. 
5. When reducing the width of the pane it is noticeable that there is some sort of "invisible" item between the analyze button and the hamburger menu button which shouldn't be there. In contrast, the combobox should take the full space when the width is less (as it did before in 4.x)
Screen Shot 2020-07-30 at 1 28 01 AM

goanpeca:
Hi @juanis2112, thanks for the review:

>1. The shortcut F8 is not working for running the code analysis

I cannot reproduce this. On my Mac pressing `Fn+F8` run the analysis as expected.

>2. Clicking on an element in the pane (warning, error, message) doesn't go to the line corresponding to this element.

Fixed

>3. The hamburger menu doesn't have the expand/collapse items that it had before.

Fixed

>4. When selecting a file in the pane with the "Select python file" button, the analysis doesn't run automatically which it did before

Fixed

>5. When reducing the width of the pane it is noticeable that there is some sort of "invisible" item between the analyze button

Fixed

goanpeca:
>more differences between 4.x and this PR:

Fixed.

ccordoba12:
> The shortcut F8 is not working for running the code analysis

This is still a problem on Linux and Windows. Please check it on a virtual machine.

All the other problems reported by @juanis2112 and I are solved now. Great job @goanpeca!

goanpeca:
>This is still a problem on Linux and Windows. Please check it on a virtual machine.

Could not reproduce but introduced a change that might help. Could you check @juanis2112, @ccordoba12 ?

ccordoba12:
> Could not reproduce but introduced a change that might help.

It's working now. Thanks @goanpeca.

@dalthviz will give this one a final manual review and if he finds everything in order, he'll merge it.

goanpeca:
>While testing this I found some things but the only related with this PR is that seems like a min width is being set so is not possible to resize the window (it only allows you to resize up to certain point):

This one is related to this PR, I will fix!

Thanks @dalthviz :-)

ccordoba12:
> This error is also present in 4.1.4 so I will create an issue or maybe already one is open @ccordoba12 ?

Please open a new issue about this one. I don't remember seeing this reported yet.

goanpeca:
>While testing this I found some things but the only related with this PR is that seems like a min width is being set so is not possible to resize the window (it only allows you to resize up to certain point):

@dalthviz I could not reproduce on Mac, will check on a Win VM and see if it is a windows thing only.

ccordoba12:
@goanpeca, please address this comment from @dalthviz:

> While testing this I found some things but the only related with this PR is that seems like a min width is being set so is not possible to resize the window (it only allows you to resize up to certain point)

so we can merge it.

goanpeca:
>@dalthviz I could not reproduce on Mac, will check on a Win VM and see if it is a windows thing only.

I cannot reproduce this on the windows VM, any particular thing that might be triggering it?

dalthviz:
@goanpeca after checking a little bit more seems like the strange behavior is related with the way the toolbars are being contracted (sometimes the resize gets stuck, although I don't now why):

resize

Looking at the code seems like a custom handling is being implemented for the `resizeEvent` to emit a custom signal: https://github.com/spyder-ide/spyder/blob/a24398264e5cb5dbbb72fda6a35555fd08fb40a4/spyder/api/widgets/__init__.py#L540-L545. Maybe that could be causing the resize to get stuck?

goanpeca:
Thanks for checking @dalthviz. Just to check is it associated to the speed in which you move?

>Looking at the code seems like a custom handling is being implemented for the resizeEvent to emit a custom signal:

Correct, that change was needed to be able to have the toolbar items use as much area as possible.

Question: If you do it slowly it does not get stuck, but if you do it faster it does?

goanpeca:
@dalthviz made some small changes to see if the situation improves.

Does increasing this number helps?

https://github.com/spyder-ide/spyder/pull/12160/files#diff-f34c043d56467c819418192611e89a6eR421

dalthviz:
@goanpeca checked again and seems the problem is still there :/

> Question: If you do it slowly it does not get stuck, but if you do it faster it does?

The behavior is the same

> Does increasing this number helps?

I'm not sure what number is that, is an option?

From what I'm seeing seems like the resize signal is applying only for the first toolbar that isn't contracted from right to left. So, to contract the rest, you need to again start the resize action. Maybe this GIF illustrates that (first the working directory toolbar needs two resize events to contract - probably do the combobox - and then the rest of the toolbars need one resize event each):

resize

goanpeca:
>From what I'm seeing seems like the resize signal is applying only for the first toolbar that isn't contracted from right to left. So, to contract the rest, you need to again start the resize action. Maybe this GIF illustrates that (first the working directory toolbar needs two resize events to contract - probably do the combobox - and then the rest of the toolbars need one resize event each)

Thanks @dalthviz, I understand what you mean. However the changes  I made refer to the toolbar in the Docks (each pane has now the options as a toolbar)

Screen Shot 2020-08-20 at 17 33 48

But it should now have any effect on the application toolbars. It seems like it is a separate issue. On master you see the same behavior?

dalthviz:
Checking with master (3bf92535c) I don't see that behavior:

resize2

goanpeca:
@dalthviz I cannot reproduce on any of the systems (binder linux) local mac and Win 10 VM. It is going to be really hard to try to fix something I cannot reproduce.

Could we add an item to https://github.com/spyder-ide/spyder/issues/13527 and move forward. Seems like a very minor issue to block this PR from merging.

ccordoba12:
> Could we add an item to #13527 and move forward

I don't agree with this. Please set up a Zoom call with @dalthviz so they two of you can fix this issue.

goanpeca:
@ccordoba12 the issue stems from trying to address this comment:

Screen Shot 2020-08-20 at 18 48 45

If I revert the "fix", then the problem is gone, but the "invisible" item is back.

That is a problem that affects all plugins that provide a widget inside the Pane (like the combobox in this case) and it deserves a fix that covers all widgets. It is not a problem specific to this Plugin, even if it was a comment that came out of reviewing it.

--- 

@dalthviz please create an event whenever it suits you. Available hours are visible on the company calendar.

>I don't agree with this. Please set up a Zoom call with @dalthviz so the two of you can fix this issue.

Fair enough, but even if you do not agree with this, @dalthviz should be able to have also an opinion and a vote, after all the question was addressed to him.

dalthviz:
> @dalthviz please create an event whenever it suits you. My available hours are visible on my company calendar.

Ok :+1: done

dalthviz:
@goanpeca please rebase this one to get the changes done in #13600

goanpeca:
>@goanpeca please rebase this one to get the changes done in #13600

Done @dalthviz.

spyder/api/widgets/__init__.py
```diff
@@ -766,7 +766,11 @@ def get_options_menu(self):
 
     def get_options_menu_button(self):
         """
+<<<<<<< HEAD
         Return the main options button of the widget.
+=======
+        Return the main options menu button of the widget.
+>>>>>>> Migrate Pylint plugin to new API
```
dalthviz, line 773: @goapeca seems like the rebase had some issues, please fix this

goanpeca: Oops. Fixing

goanpeca: Fixed @dalthviz

spyder/plugins/pylint/main_widget.py
```diff
@@ -0,0 +1,899 @@
...
+        self.code_analysis_action = None
+        self.browse_action = None
+
+        # Widgets
+        self.filecombo = PythonModulesComboBox(self)
+        self.ratelabel = QLabel(self)
+        self.datelabel = QLabel(self)
+        self.treewidget = ResultsTree(self)
+
+        # FIXME: Move to plugin?
```
dalthviz, line 262: So, should we move the things below this line to the plugin? (or was this comment before the move to the API maybe?)

goanpeca: Removed the FIXME notice.

goanpeca:
>@goanpeca left a question and also I think this plugin should use the spinner. Besides that LGTM 👍

Thanks for the feedback @dalthviz, made the requested fix.

dalthviz:
@andfoy @steff456 could you give to this PR a manual test on Linux and Mac? Thanks!

steff456:
Hi @goanpeca and @dalthviz,

With @andfoy we manually review this PR and everything is working as expected.

dalthviz (commented):
@goanpeca one last comment, please unify the use of `"` to enclose strings, there is a mix (as an example the pylint `main_widget.py` in the setup of the actions) and using just `"` could improve future readability of the code. Thanks!

goanpeca:
>@goanpeca one last comment, please unify the use of " to enclose strings, there is a mix (as an example the pylint main_widget.py in the setup of the actions) and using just " could improve future readability of the code. Thanks!

Fixed @dalthviz.

spyder/plugins/pylint/main_widget.py
```diff
@@ -0,0 +1,899 @@
...
+
+    def get_focus_widget(self):
+        return self.treewidget
+
+    def setup(self, options):
+        change_history_depth_action = self.create_action(
+            PylintWidgetActions.ChangeHistory,
+            text=_("History..."),
+            tip=_("Set history maximum entries"),
+            icon=self.create_icon('history'),
```
dalthviz, line 423: @goapeca seems like the mix of `'` and `"` is still there, maybe the rebase didn't pick the changes?

dalthviz (commented):
@goanpeca seems like no change regarding the mix of `'` and `"` was added with the rebase, could you check what happen?

spyder/plugins/pylint/main_widget.py
goanpeca: You re right @dalthviz I pushed the rebase but not the commit for fixing the quotes. My bad.

Fixed noew.

spyder/plugins/pylint/main_widget.py
```diff
@@ -0,0 +1,899 @@
...
+        Return command to use to run code analysis on given filename
+        """
+        command_args = []
+        if PYLINT_VER is not None:
+            command_args = [
+                "-m",
+                "pylint",
+                "--output-format=text",
+                "--msg-template="
+                ""{msg_id}:{symbol}:{line:3d},{column}: {msg}"",
```
dalthviz, line 804: Seems like the replace needs some changes here (this is causing the error in the CIs)

dalthviz (commented):
@goanpeca a comment pointing were the CIs are failing

spyder/plugins/pylint/main_widget.py
dalthviz: Also, some PEP8 errors raised. I know these kind of things are kind of a pain to change so thanks for the effort :+1:

goanpeca: No worries @dalthviz, it is how it goes.

Thanks for all the reviewing. You are all doing a great job, and practice makes perfect :-) !

dalthviz (commented):
@goanpeca checking seems like the selected path in the combobox is not being taken in to account unless you use the `Select python file` button to update it:

combopylint

Edit: Seems like the `Analyze` button always triggers the action using the current filename in the Editor

goanpeca:
>Edit: Seems like the Analyze button always triggers the action using the current filename in the Editor

Bug!

Will fix.

ccordoba12:
Hey @dalthviz, is this one ready for merge?

dalthviz:
@ccordoba12 I would say that some testing could be nice, otherwise the missing thing is a squash

ccordoba12 (approved):
Thanks @goanpeca for your work on this PR and @dalthviz for your help to finish it.

I also gave this work some manual testing too (per @dalthviz's request) and it's working as expected.

PR #12177, opened 2020-04-06
PR: Register actions without shortcuts and only display registered actions

### LINKED ISSUE #12179: Shortcuts table loads whatever is saved on the shortcuts section
We currently load shortcuts from the configuration. This means anything there will be loaded, even if it is a leftover from a shortcut name rename or similar.

image

goanpeca:
## Description of Changes

We currently load shortcuts from the configuration. This means anything there will be loaded, even if it is a leftover from a shortcut name rename or similar.

This PR adds a check on the shortcut preferences so that we displayed shortcuts that ara registered on the main window. This way we display only valid active shortcuts. Additional to this we update the `apply_shortcuts` method so that actions without a shortcut in CONF can be successfully registered. With the new work on API migration we will expose many actions without shortcuts and leave it to the user to define them if they want to use them.

Also by using the `shortcut_data` info  from the main window, we can offer translations for shortcuts via those QActions and QShortcuts. (Work for a separate PR)

### Issue(s) Resolved

Fixes #12179

### Current behavior (adding some random shortcut on the shortcut section)

image

### Affirmation

### DIFF
--- a/spyder/preferences/shortcuts.py
+++ b/spyder/preferences/shortcuts.py
@@ -631,7 +631,7 @@ class ShortcutsTable(QTableView):
         QTableView.__init__(self, parent)
         self._parent = parent
         self.finder = None
-
+        self.shortcut_data = None
         self.source_model = ShortcutsModel(
                                     self,
                                     text_color=text_color,
@@ -654,7 +654,14 @@ class ShortcutsTable(QTableView):
         self.selectionModel().selectionChanged.connect(self.selection)
 
         self.verticalHeader().hide()
-        self.load_shortcuts()
+
+    def set_shortcut_data(self, shortcut_data):
+        """
+        Shortcut data comes from the registration of actions on the main
+        window. This allows to only display the right actions on the
+        shortcut table. This also allows to display the localize text.
+        """
+        self.shortcut_data = shortcut_data
 
     def focusOutEvent(self, e):
         """Qt Override."""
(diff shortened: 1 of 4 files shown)


goanpeca:
>spyder-ide.spyder (macOS Python27-slow)

Failure seems like a azure issue.

ccordoba12:
> Additional to this we update the apply_shortcuts method so that actions without a shortcut in CONF can be successfully registered.

This could cause shortcut conflicts if we're not register an action with the proper context. So I think this addition should be left for Spyder 5 and only fix the problem with loading anything in the shortcuts section for Spyder 4 here.

goanpeca:
>This could cause shortcut conflicts if we're not register an action with the proper context.

I do not understand

ccordoba12:
Actions can be placed on the main window menus, toolbars and Options menus. That could make Qt to report a shortcut conflict for several actions that we're not exposing right now.

goanpeca:
>Actions can be placed on the main window menus, toolbars and Options menus. That could make Qt to report a shortcut conflict for several actions that we're not exposing right now.

But this PR is not exposing any new actions, it is only making sure that only registered actions appear on the Shortcut  Preferences.

So even if there are duplicate actions right now (same name, same context but different QAction or QShortcut) we only check `(name, context)` pair so there is no possibility of clashes with this PR.

goanpeca:
>Actions can be placed on the main window menus, toolbars and Options menus. That could make Qt to report a shortcut conflict for several actions that we're not exposing right now.

But this PR is not exposing any new actions, it is only making sure that only registered actions appear on the Shortcut  Preferences.

So even if there are duplicate actions right now (same name, same context but different QAction or QShortcut) we only check `(name, context)` pair so there is no possibility of clashes with this PR.

spyder/preferences/shortcuts.py
```diff
@@ -681,14 +688,22 @@ def adjust_cells(self):
 
     def load_shortcuts(self):
         """Load shortcuts and assign to table model."""
+        # item[1] -> context, item[2] -> name
+        shortcut_data = [(item[1], item[2]) for item in self.shortcut_data]
```
goanpeca, line 692: @ccordoba12 there is no possibility of clash since we only check if the (context, name) of the `iter_shortcuts` appear on the registered actions in the `shortcut_data`

spyder/preferences/shortcuts.py
```diff
@@ -681,14 +688,22 @@ def adjust_cells(self):
 
     def load_shortcuts(self):
         """Load shortcuts and assign to table model."""
+        # item[1] -> context, item[2] -> name
+        shortcut_data = [(item[1], item[2]) for item in self.shortcut_data]
         shortcuts = []
         for context, name, keystr in CONF.iter_shortcuts():
-            shortcut = Shortcut(context, name, keystr)
-            shortcuts.append(shortcut)
+            if (context, name) in shortcut_data:
```
goanpeca, line 695: Here is the actual check.

Issue #9987, opened 2019-08-09
SVG renderer does not properly render matplotlib plots


akdor1154:
### Issue Report Checklist

* [x] Searched the [issues page](https://github.com/spyder-ide/spyder/issues?q=is%3Aissue) for similar reports
* [x] Read the relevant sections of the [Spyder Troubleshooting Guide](https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ) and followed its advice
* [x] Reproduced the issue after updating with ``conda update spyder`` (or ``pip``, if not using Anaconda)
* [x] Completed the **Problem Description**, **Steps to Reproduce** and **Version** sections below

## Problem Description
Spyder's SVG renderer does not support SVG features that are used by matplotlib.

### What steps reproduce the problem?
0. run `%config InlineBackend.figure_formats = ['svg']`

1. Render the following plot with matplotlib: 
![Screenshot_2019-08-09_12-19-32](https://user-images.githubusercontent.com/6732831/62749505-069d0c00-baa0-11e9-9026-b7554b0a043a.png) (as rendered by Ristretto on my machine)
[example.svg.txt](https://github.com/spyder-ide/spyder/files/3484418/example.svg.txt) (matplotlib's svg output, sorry github doesn't accept direct svg uploads)

(sorry I don't have a MRE, but it uses hatching, .e.g.
```
    ax.fill_between(
        summarySNS['Date'],
        summarySNS['lowBoundary'],
        summarySNS['highBoundary'],
        facecolor="none",
        edgecolor=getFillColour(color, dl=+0.2, ds=-0.1),
        hatch='//////'
    )
```
)

2.  Observe that spyder renders it as
![Screenshot_2019-08-09_11-54-17](https://user-images.githubusercontent.com/6732831/62749304-3c8dc080-ba9f-11e9-9968-d8ef1a3d6a68.png)

### What is the expected output? What do you see instead?
The expected output would be the SVG properly rendered, at it would be by a browser.

To be honest I'd love to see you use QWebEngine to display it - that way interactive SVG with Javascript could work as well, and it would be a gateway to displaying plot libraries that output as ipython html widgets... 

## Versions

* Spyder version: 4.0b4
* Python version: 3.7.3 x64
* Qt version:  5.12.2
* PyQt version: 5.12.1
* Operating System name/version:  Ubuntu 19.04

### Dependencies

```
pygments >=2.0    :  2.3.1 (OK)
[...]

PR #12219, opened 2020-04-08
Improve plot pane

impact27:
## Description of Changes
Add saving figures in different formats and add interactive figures (only during first draw)

### Issue(s) Resolved

Fixes #
plot
12211

### Affirmation

### DIFF
--- a/external-deps/spyder-kernels/.gitrepo
+++ b/external-deps/spyder-kernels/.gitrepo
@@ -6,7 +6,7 @@
 [subrepo]
 	remote = https://github.com/spyder-ide/spyder-kernels.git
 	branch = improve_plot_pane
-	commit = df35784943a5316729b3808231d8edcd0ef76674
-	parent = d0ce0ef86487a92e09351c53351faebc136fa54d
+	commit = 9b2aaccd4b9420fd42a8f09e4cced7a386085f70
+	parent = a91afb7d6216beabdeae230e0b8023dc283dc31e
 	method = merge
 	cmdver = 0.4.1
--- a/external-deps/spyder-kernels/.gitrepo
+++ b/external-deps/spyder-kernels/.gitrepo
@@ -5,8 +5,8 @@
 ;
 [subrepo]
 	remote = https://github.com/spyder-ide/spyder-kernels.git
-	branch = master
-	commit = ae3957c5a01fac95515aff964fb64af30d188716
-	parent = 652cfa2c786d02c92021c117deb03548df2b2207
+	branch = improve_plot_pane
+	commit = df35784943a5316729b3808231d8edcd0ef76674
+	parent = d0ce0ef86487a92e09351c53351faebc136fa54d
 	method = merge
 	cmdver = 0.4.1
--- a/external-deps/spyder-kernels/spyder_kernels/customize/spydercustomize.py
+++ b/external-deps/spyder-kernels/spyder_kernels/customize/spydercustomize.py
@@ -37,6 +37,32 @@ else:
 
 logger = logging.getLogger(__name__)
 
+# =============================================================================
+# Figure plots additions
+# =============================================================================
+# Catch the first draw call
+# Subsequent draw calls not called but could modify to update the figure
+try:
+    from IPython.display import set_matplotlib_formats
+    from matplotlib.figure import Figure
+
+    super_draw = Figure.draw
+
+    def draw(self, *args, **kwargs):
+        sent = getattr(self, '_was_sent', False)
+        if not sent:
+            # Send all file types so we can choose the extension
+            # Should we replace retina by png? large png can always be resized.
+            set_matplotlib_formats('svg', 'pdf', 'retina', 'jpeg')
+            self._was_sent = True
+            format = get_ipython().kernel.shell.display_formatter.format
+            format_dict, md_dict = format(self)
+            frontend_request(blocking=False).display_data(format_dict)
+        return super_draw(self, *args, **kwargs)
+
+    Figure.draw = draw
+except Exception:
+    pass
 
 # =============================================================================
 # sys.argv can be missing when Python is embedded, taking care of it.
--- a/external-deps/spyder-kernels/spyder_kernels/customize/spydercustomize.py
+++ b/external-deps/spyder-kernels/spyder_kernels/customize/spydercustomize.py
@@ -44,23 +44,23 @@ logger = logging.getLogger(__name__)
 # Subsequent draw calls not called but could modify to update the figure
 try:
     from IPython.display import set_matplotlib_formats
-    from matplotlib.figure import Figure
+    import matplotlib.figure
 
-    super_draw = Figure.draw
+    class MyFigure(matplotlib.figure.Figure):
+        def draw(self, *args, **kwargs):
+            sent = getattr(self, '_was_sent', False)
+            if not sent:
+                # Send all file types so we can choose the extension
+                # Should we replace retina by png? large png can be resized.
+                # TODO: add options for 'pdf', 'retina', 'jpeg'
+                set_matplotlib_formats('svg', 'png')
+                self._was_sent = True
+                format = get_ipython().kernel.shell.display_formatter.format
+                format_dict, md_dict = format(self)
+                frontend_request(blocking=False).display_data(format_dict)
+            return super().draw(*args, **kwargs)
 
-    def draw(self, *args, **kwargs):
-        sent = getattr(self, '_was_sent', False)
-        if not sent:
-            # Send all file types so we can choose the extension
-            # Should we replace retina by png? large png can always be resized.
-            set_matplotlib_formats('svg', 'pdf', 'retina', 'jpeg')
-            self._was_sent = True
-            format = get_ipython().kernel.shell.display_formatter.format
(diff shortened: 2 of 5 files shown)


impact27:
One problem with this approach is that the plots take more time to render as there is 4x more plots to render (svg + pdf + jpeg + png). What do you think? Should we limit ourselves to 1 that the user can coose, or is the slower plotting time acceptable?

jnsebgosselin:
> 
> 
> One problem with this approach is that the plots take more time to render as there is 4x more plots to render (svg + pdf + jpeg + png). What do you think? Should we limit ourselves to 1 that the user can choose, or is the slower plotting time acceptable?

That and this may also increase the memory footprint of the plugin substantially.

Also, does this PR completely remove the ability to render svg images directly in the plot plugin? If I understand correctly, if there is no png in the data received by the console then nothing is plotted? If that is the case, I'm not sure I like that.

bcolsen:
We could just use the svg file by default and convert to other formats on save from the svg. 

On the flip side we could use (potentially) even more ram and place QtWebEngine element in the tabs and use the webagg (notebook) backend. (This could also give us all the other web based plotting libraries)

https://matplotlib.org/3.1.3/gallery/user_interfaces/embedding_webagg_sgskip.html

impact27:
> Also, does this PR completely remove the ability to render svg images directly in the plot plugin? If I understand correctly, if there is no png in the data received by the console then nothing is plotted? If that is the case, I'm not sure I like that.

The idea is that they would _always_ be png in the data. The kernel would send png + something else. The png can be used as a thumbnail and the something else can be saved to file. This means that we don't need to render pdf or svg in the plot pane. svg are rasterised anyway now. This adds 100ms of processing time to create both instead of just one and increases the memory footprint a bit (~factor two if we select just one other extension)

jnsebgosselin:
> We could just use the svg file by default and convert to other formats on save from the svg.

The problem is that the `QSvgRenderer ` doesn't support all svg features, so we would have to use another svg rendered to do this properly. See See https://github.com/spyder-ide/spyder/issues/9987.

Personnally, I think we should only send the format that is specified by the user in the `inline backed format` preference and limit the saving format option that the ones that are supported by the renderer we are using.

jnsebgosselin:
> 
> 
> > Also, does this PR completely remove the ability to render svg images directly in the plot plugin? If I understand correctly, if there is no png in the data received by the console then nothing is plotted? If that is the case, I'm not sure I like that.
> 
> The idea is that they would _always_ be png in the data. The kernel would send png + something else. The png can be used as a thumbnail and the something else can be saved to file. This means that we don't need to render pdf or svg in the plot pane. svg are rasterised anyway now. This adds 100ms of processing time to create both instead of just one and increases the memory footprint a bit (~factor two if we select just one other extension)

Even when plots are not produced with matplotlib?

What if we want to load images in the plot plugin directly from a source other than the kernel?

impact27:
> Personnally, I think we should only send the format that is specified by the user in the `inline backed format` preference and limit the saving format option that the ones that are supported by the renderer we are using.

What about sending `png` + the `inline backed format` so there is no rendering required and we can accept any format?

impact27:
> Even when plots are not produced with matplotlib?

You are right, I think the svg processing should stay then and be used if no png is avilable.

> What if we want to load images in the plot plugin directly from a source other than the kernel?
You would need to load a version of the image we can render + any version you want to save

jnsebgosselin:
> 
> 
> > Personnally, I think we should only send the format that is specified by the user in the `inline backed format` preference and limit the saving format option that the ones that are supported by the renderer we are using.
> 
> What about sending `png` + the `inline backed format` so there is no rendering required and we can accept any format?

Rendering from a svg results in a better onscreen image quality than from a png because of the scaling.

I agree that svg rendering is not ideal currently due to what was discussed in #9987, but I hope that one day either `QSvgRenderer ` will be improved or that we will use a better svg renderer in Spyder, which should solve those issues.

ccordoba12:
> I agree that svg rendering is not ideal currently due to what was discussed in #9987, but I hope that one day either QSvgRenderer will be improved or that we will use a better svg renderer in Spyder, which should solve those issues.

We can use a QWebView to render svg's, as I suggested above. That should be really simple to do.

jnsebgosselin:
> 
> 
> > I agree that svg rendering is not ideal currently due to what was discussed in #9987, but I hope that one day either QSvgRenderer will be improved or that we will use a better svg renderer in Spyder, which should solve those issues.
> 
> We can use a QWebView to render svg's, as I suggested above. That should be really simple to do.

This is worth a try.

In my opinion, I think the problem here is that this PR is addressing two topics that should be adressed/discussed separately:

1 - Display plot in the plots plugin even when the backend isn't set to inline.
2 - Improvement/refactoring of the rendering and saving format in the plot plugin.

jnsebgosselin:
> 
> 
> > > I agree that svg rendering is not ideal currently due to what was discussed in #9987, but I hope that one day either QSvgRenderer will be improved or that we will use a better svg renderer in Spyder, which should solve those issues.
> > 
> > 
> > We can use a QWebView to render svg's, as I suggested above. That should be really simple to do.
> 
> This is worth a try.
> 
> In my opinion, I think the problem here is that this PR is addressing two topics that should be adressed/discussed separately:
> 
> 1 - Display plots produced with matplotlib in the plots plugin even when the backend isn't set to inline.
> 2 - Improvement/refactoring of the rendering and saving format in the plot plugin.

bcolsen:
> We can use a QWebView to render svg's, as I suggested above. That should be really simple to do.

Would this allow webagg and all the other web visualizations or would we need  QWebEngine for that?

I looks like QWebView can save rendered bitmaps.

ccordoba12:
> Would this allow webagg and all the other web visualizations or would we need QWebEngine for that?

No, that requires a different plugin. Plots will be in charge of handling plots and not anything else.

jnsebgosselin:
> 
> 
> > We can use a QWebView to render svg's, as I suggested above. That should be really simple to do.
> 
> Would this allow webagg and all the other web visualizations or would we need QWebEngine for that?
> 
> I looks like QWebView can save rendered bitmaps.

Are you suggesting to use `WebAgg` to embed a fully interactive matplotlib plot? If that is the case, I'm not sure I agree with this idea. Embedding interactive matplotlib canvas into Spyder GUI looks nice on paper, but because of scaling, it's going to be horrible to use. I much much prefer to have my interactive matplotlib plots to pop up in independent, floating free windows that respect faithfully the size and scaling of the plots I have created.

goanpeca:
>but because of scaling, it's going to be horrible to use. I

@jnsebgosselin could you elaborate on this a bit?

jnsebgosselin:
> 
> 
> > but because of scaling, it's going to be horrible to use. I
> 
> @jnsebgosselin could you elaborate on this a bit?

You can't scale an interactive matplotlib figure like you do for a svg or a png. Let's say you produce a nice 8 x 4i nches matplotlib figure with a complex layout. This same interactive figure is going to look horrible in Spyder after we resized it so that it fits into, lets say, a 500 x 400 pixels panel. The fontsize, line thickness, marker size, margins, etc. won't scale when we resize the figure. The layout of the whole figure is going to look very bad.

We could make it work, maybe, by rigging some gui events with methods like `tight_layout`, etc, but still, I think it would only work well for very simple figure layouts.

To illustrate my point, in the image below we can see a figure I've produced with matplotlib. You can see how it looks in the plots plugin and how it looks in a Qt5 matplotlib figure manager in its original size. I have then resized (manually) the manager so that the size of the figure fits that of the plots plugin. You can see how it looks... very bad. Embedding an interactive matplotlib figure in a GUI works well if the layout of the figure is specifically engineered to fit in that specific GUI, otherwise, chances are it's going to look bad.

Does it make any sense? In my opinion, embedding interactive matplotlib figures in a Spyder panel is not as simple as it may look and not as cool as we may think...

image

goanpeca:
Understood! Thanks for the thorough explanation.

bcolsen:
@goanpeca @jnsebgosselin Sorry for the late reply and off topic here but scaling could be solved if we use page rendering zoom the webview.

The issue @jnsebgosselin describes is true if we resize the figure, which would result in a mess as he shows. Instead however we can leave the figure size unchanged and create a webview large enough for the figure and then use the webview zoom feature to scale the whole page. The matplotlib webagg backed is entirely pixel based so the plots only scale as well as a png image but that's currently what we are using.

Gif of changing the figure size and then zooming a plot in a web browser instead:

PLot_zoom

jnsebgosselin:
It sounds pretty interesting. As soone as I have time, I think I'll give it a try unless someone beat me to it :P.

impact27:
I will close this PR as the code is specific to matplotlib.

ccordoba12:
> I will close this PR as the code is specific to matplotlib.

I don't understand your reasoning @impact27. The Plots pane only works with Matplotlib at the moment (and I think that won't change in the future).

impact27:
> > I will close this PR as the code is specific to matplotlib.
> 
> I don't understand your reasoning @impact27. The Plots pane only works with Matplotlib at the moment (and I think that won't change in the future).

Not really, it works with any png or svg displayed in the console. This PR would break this behaviour.

PR #12231, opened 2020-04-09
WIP: MacApp: Launch jedi with Spyder's python executable

mrclary:
## Description of Changes
Ensured that jedi is launched with Spyder's python executable, regardless of what iPython console executable is selected.

Previously, the `jedi` environment was set to the configured `main_interpreter`, which could be Spyder's python executable _or_ an external one. If Spyder is launched from a conda environment, this is inconsequential; but if Spyder is launched from elsewhere (such as in the Mac standalone application) `jedi` will fail with error:

```
jedi.api.exceptions.InternalError: The subprocess <environment path>/bin/python3.7 has crashed (EOFError('Ran out of input'), stderr=)
```

### Issue(s) Resolved

Fixes spyder-ide/mac-application#5 and spyder-ide/mac-application#4

### Affirmation

@mrclary

### DIFF
--- a/spyder/plugins/completion/languageserver/plugin.py
+++ b/spyder/plugins/completion/languageserver/plugin.py
@@ -574,13 +574,8 @@ class LanguageServerPlugin(SpyderCompletionPlugin):
         }
 
         # Jedi configuration
-        if self.get_option('default', section='main_interpreter'):
-            environment = None
-        else:
-            environment = self.get_option('custom_interpreter',
-                                          section='main_interpreter')
         jedi = {
-            'environment': environment,
+            'environment': sys.executable,
             'extra_paths': self.get_option('spyder_pythonpath',
                                            section='main', default=[]),
         }


ccordoba12:
We can't accept as it is this because it'll break functionality already in place.

mrclary:
> We can't accept this as it is this because it'll break functionality already in place.

@ccordoba12, can you be more specific? What functionality will break? Perhaps I'm missing something, but shouldn't jedi be using the same python executable as Spyder, and not the external iPython console's? I didn't notice any effect for bootstrap or starting Spyder from conda environments, but jedi's functionality _is_ broken for a stand-alone application if it uses the external environment interpreter.

ccordoba12:
> but shouldn't jedi be using the same python executable as Spyder, and not the external iPython console's?

Sorry for being so short (I thought you were going to take a look at the tests to see why they are failing). This is the problem exactly: Jedi can provide code completions for other Python interpreters, different from the one it's installed in.

So, if you have `xarray` in the conda env you want to work with, but not in the one Spyder is installed in, then for sure you'd like to have code completions for `xarray`. And that's what we are already providing and what this PR removes.

So you need to find another way to fix what's happening in the Mac app because we're not going to remove this functionality, sorry.

mrclary:
> Sorry for being so short (I thought you were going to take a look at the tests to see why they are failing). This is the problem exactly: Jedi can provide code completions for other Python interpreters, different from the one it's installed in.
> 
> So, if you have `xarray` in the conda env you want to work with, but not in the one Spyder is installed in, then for sure you'd like to have code completions for `xarray`. And that's what we are already providing and what this PR removes.

let me see if I understand you correctly...
* I have environment A from which I will launch Spyder; it does not have package b installed.
* I have environment B which has package b installed.
* Code completion will not work in this situation if jedi's environment is set to A.

Is this correct? Does this apply only to the Editor? or the iPython console as well? Are there any backup code completers that I should disable in order to test this functionality as I work on a solution?

mrclary:
@ccordoba12, okay I can now confirm what you were saying. It appears only to affect the Editor; I guess iPython handles its own completion...

I'll work on a different solution for the Mac app.

ccordoba12:
> okay I can now confirm what you were saying. It appears only to affect the Editor; I guess iPython handles its own completion...

Yep, that's the idea. We can provide code completions in the editor for package b, installed in environment B, in your example above, if users select the interpreter for that env in `Preferences > Python interperter`.

mrclary:
Okay, I want to thank everyone for being so patient with me. There is good news and bad news.

### Good news:
I don't know how I came to think the changes I made for this PR were required for the stand-alone Mac app, but they are not. I must have confused them with other changes I made during my experimentation process. So, we can close this PR.

### Bad news:
Jedi's command completion does not seem to work _completely_ for the stand-alone app. The silver-lining here is that the behavior can be reproduced in bootstrap, so it is something to do with the difference between pyenv+pip and anaconda, _not_ with the stand-alone build.

I will start a new issue for this to discuss the details.

ccordoba12:
> Okay, I want to thank everyone for being so patient with me.

No worries. Thanks for all the effort you're putting into this!

PR #12235, opened 2020-04-09
PR: Use clean environment for Python version and module checks (Mac app)

mrclary:
## Description of Changes
Use a clean shell environment when obtaining python interpreter version or installed module version.

Currently, when checking for python version or installed module versions in subprocesses in preparation for launching a `spyder_kernel`, the subprocess inherits Spyder's environment variables, which can lead to incorrect results if Spyder's environment is not a conda environment (such as in a standalone Mac application). By using a clean subprocess environment, the python version and installed module versions are correctly checked.

Currently, installed module versions are checked by writing a new script to a temporary script file and running it in a python subprocess, possibly several times, depending on the version requirements. This is replaced with a much simpler, single python subprocess call to obtain the installed module version, which can be compared to the requirement(s).

### Issue(s) Resolved

This allows the stand-alone Mac application to function properly.

### Affirmation

@mrclary

### DIFF
--- a/spyder/plugins/ipythonconsole/plugin.py
+++ b/spyder/plugins/ipythonconsole/plugin.py
@@ -884,7 +884,8 @@ class IPythonConsole(SpyderPluginWidget):
                     cmd = cmd.encode('utf-8')
 
                 try:
-                    proc = programs.run_shell_command(cmd)
+                    # Use clean environment
+                    proc = programs.run_shell_command(cmd, env={})
                     output, _err = proc.communicate()
                 except subprocess.CalledProcessError:
                     output = ''
--- a/spyder/utils/programs.py
+++ b/spyder/utils/programs.py
@@ -12,9 +12,9 @@ from __future__ import print_function
 from ast import literal_eval
 from distutils.version import LooseVersion
 from getpass import getuser
+from textwrap import dedent
 import glob
 import imp
-import inspect
 import itertools
 import os
 import os.path as osp
@@ -824,38 +824,49 @@ def get_module_version(module_name):
 
 def is_module_installed(module_name, version=None, interpreter=None):
     """
-    Return True if module *module_name* is installed
+    Return True if module ``module_name`` is installed
 
-    If version is not None, checking module version
-    (module must have an attribute named '__version__')
+    If ``version`` is not None, checks that the module's installed version is
+    consistent with ``version``. The module must have an attribute named
+    '__version__' or 'VERSION'.
 
     version may start with =, >=, > or < to specify the exact requirement ;
     multiple conditions may be separated by ';' (e.g. '>=0.13;<1.0')
 
-    interpreter: check if a module is installed with a given version
-    in a determined interpreter
+    If ``interpreter`` is not None, checks if a module is installed with a
+    given ``version`` in the ``interpreter``'s environment. Otherwise checks
+    in Spyder's environment.
     """
-    module_version = None
     if interpreter is not None:
         if is_python_interpreter(interpreter):
-            cmd = 'import %s; print(%s.__version__)' % ((module_name,) * 2)
+            cmd = dedent('''
+                         try:
+                             import %s as mod
+                         except Exception:
+                             print('No Module')
+                         print(getattr(mod, '__version__',
+                                       getattr(mod, 'VERSION', None)))
+                         ''' % module_name)
             try:
                 # use clean environment
                 proc = run_program(interpreter, ['-c', cmd], env={})
-                output, _err = proc.communicate()
+                stdout, stderr = proc.communicate()
+                stdout = stdout.decode().strip()
             except Exception:
                 return False
 
-            if _err.decode() == '':
-                module_version = output.decode().strip()
-            else:
+            if 'No Module' in stdout:
                 return False
+            elif stdout != 'None':
+                # the module is installed and it has a version attribute
+                module_version = stdout
+            else:
+                module_version = None
         else:
-            # Try to not take a wrong decision if interpreter check
-            # fails
+            # Try to not take a wrong decision if interpreter check fails
             return True
-
-    if module_version is None:
+    else:
+        # interpreter is None, just get module version in Spyder environment
         try:
             module_version = get_module_version(module_name)
         except Exception:
--- a/spyder/utils/programs.py
+++ b/spyder/utils/programs.py
@@ -139,17 +139,18 @@ def alter_subprocess_kwargs_by_platform(**kwargs):
         CONSOLE_CREATION_FLAGS |= CREATE_NO_WINDOW
         kwargs.setdefault('creationflags', CONSOLE_CREATION_FLAGS)
 
-        # ensure Windows subprocess environment has SYSTEMROOT, but variable
-        # is case insensitive
-        if ((kwargs.get('env') is not None) and
-            ('SYSTEMROOT' not in map(str.upper, kwargs['env'].keys()))):
-            sys_root_key = None
-            for k, v in os.environ.items():
-                if 'SYSTEMROOT' == k.upper():
-                    sys_root_key = k
-                    break  # don't risk multiple values
-            if sys_root_key is not None:
-                kwargs['env'].update({sys_root_key: os.environ[sys_root_key]})
+        # ensure Windows subprocess environment has SYSTEMROOT
+        if (kwargs.get('env') is not None):
+            # Is SYSTEMROOT in env? case insensitive
+            if ('SYSTEMROOT' not in map(str.upper, kwargs['env'].keys())):
+                # Add SYSTEMROOT from os.environ
+                sys_root_key = None
+                for k, v in os.environ.items():
+                    if 'SYSTEMROOT' == k.upper():
+                        sys_root_key = k
+                        break  # don't risk multiple values
+                if sys_root_key is not None:
+                    kwargs['env'].update({sys_root_key: os.environ[sys_root_key]})
 
     return kwargs
 
--- a/spyder/utils/programs.py
+++ b/spyder/utils/programs.py
@@ -822,8 +822,7 @@ def get_module_version(module_name):
     return getattr(mod, '__version__', getattr(mod, 'VERSION', None))
 
 
-def is_module_installed(module_name, version=None, installed_version=None,
-                        interpreter=None):
+def is_module_installed(module_name, version=None, interpreter=None):
     """
     Return True if module *module_name* is installed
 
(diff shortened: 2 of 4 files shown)


goanpeca:
is `env`something that works on windows?

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -844,11 +844,9 @@ def interpreter_versions(self):
             import subprocess
             versions = {}
             pyexec = CONF.get('main_interpreter', 'executable')
-            py_cmd = u'%s -c "import sys; print(sys.version)"' % pyexec
-            ipy_cmd = (
-                u'%s -c "import IPython.core.release as r; print(r.version)"'
-                % pyexec
-            )
+            py_cmd = u'env -i %s -c "import sys; print(sys.version)"' % pyexec
```
goanpeca, line 847: Unless this works on windows, this is a no go

mrclary: Ahh, yes, of course. I don't know if there is an equivalent on windows and a quick google search didn't reveal anything. Since this is only relevant for the Mac app, and should work for *nix systems, perhaps just a control statement here?

goanpeca: Sure, anywhere you are using the `env` command there should be a `if os.name != 'nt'` or similar.

I would argue that if some things are really specific to the MacApp and how it works (or does not work) then we should probably consider adding a `IS_MAC_APP` somewhere and only add those changes conditionally as well.

mrclary: Thanks to @ccordoba12 for bringing `running_in_mac_app` to my attention (#12271); I'll keep that in mind in the future. As I mentioned there, I think the benefits extend beyond the Mac app so I think the flow control option is best here.

mrclary: I'm backtracking from my last statement. I'll use `running_in_mac_app` for the cases in this PR. I double-checked and they only apply when launched from the mac app. Launching from non-conda environment (at least in bootstrap) does not cause the issue.

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -859,7 +859,8 @@ def interpreter_versions(self):
                     cmd = cmd.encode('utf-8')
 
                 try:
-                    proc = programs.run_shell_command(cmd)
+                    # use clean environment
```
goanpeca, line 862: use clean environment -> Use clean environment

spyder/utils/programs.py
```diff
@@ -823,78 +836,59 @@ def is_module_installed(module_name, version=None, installed_version=None,
...
-            try:
-                actver = get_module_version(module_name)
-            except:
-                # Module is not installed
-                return False
-        else:
-            actver = installed_version
-        if actver is None and version is not None:
+
+    if actver is None:
```
goanpeca, line 866: actver -> expand to full readable name :-p ?

spyder/utils/programs.py
```diff
@@ -823,78 +836,59 @@ def is_module_installed(module_name, version=None, installed_version=None,
...
+            versions = version.split(';')
         else:
-            if ';' in version:
-                output = True
-                for ver in version.split(';'):
-                    output = output and is_module_installed(module_name, ver)
-                return output
-            match = re.search(r'[0-9]', version)
+            versions = [version]
+        output = True
```
goanpeca, line 880: Add a space on top of output for readability

spyder/utils/programs.py
```diff
@@ -823,78 +836,59 @@ def is_module_installed(module_name, version=None, installed_version=None,
...
-                # Module is not installed
-                return False
-        else:
-            actver = installed_version
-        if actver is None and version is not None:
+
+    if actver is None:
+        try:
+            actver = get_module_version(module_name)
+        except:
```
goanpeca, line 869: We need to define an exception here

`except Exception:`  (at least)

spyder/utils/programs.py
```diff
@@ -823,78 +836,59 @@ def is_module_installed(module_name, version=None, installed_version=None,
...
-            finally:
-                if not f.closed:
-                    f.close()
-                os.remove(script)
+                # use clean environment
+                proc = run_program(interpreter, ['-c', cmd], env={})
+                output, _err = proc.communicate()
+            except Exception:
+                return False
+            if _err.decode() == '':
```
goanpeca, line 857: At a space on top of the if please

mrclary:
I think this is ready

spyder/utils/programs.py
```diff
@@ -817,84 +830,67 @@ def is_module_installed(module_name, version=None, installed_version=None,
...
-            get_modver = inspect.getsource(get_module_version)
-            stable_ver = inspect.getsource(is_stable_version)
-            ismod_inst = inspect.getsource(is_module_installed)
+    if installed_version is not None:
+        '''
+        installed_version only passed in from dependencies.Dependency.check
+        which retrieved it from get_module_version, so we know the module is
+        already installed
+        '''
+        return True
```
ccordoba12, line 845: It seems now `installed_version` is unnecessary. Please remove it if that's the case.

spyder/utils/programs.py
```diff
@@ -817,84 +830,67 @@ def is_module_installed(module_name, version=None, installed_version=None,
...
 
-            f = tempfile.NamedTemporaryFile('wt', suffix='.py',
-                                            dir=get_temp_dir(), delete=False)
+    module_version = None
+    if interpreter is not None:
+        if is_python_interpreter(interpreter):
+            cmd = 'import %s; print(%s.__version__)' % ((module_name,) * 2)
```
ccordoba12, line 850: This is incorrect because not all modules have a `__version__` variable, for instance `keyring`. Therefore, this will report `keyring` as missing when in fact it's installed.

The previous version reported it as installed, by the way.

spyder/utils/programs.py
```diff
@@ -817,84 +830,67 @@ def is_module_installed(module_name, version=None, installed_version=None,
...
-            try:
-                actver = get_module_version(module_name)
-            except:
-                # Module is not installed
-                return False
-        else:
-            actver = installed_version
-        if actver is None and version is not None:
+
+    if module_version is None:
```
ccordoba12, line 867: How can `module_version` be other than `None` if nothing has modified it between your declaration above and this point?

This looks ugly to me, so could you rewrite it a bit?

spyder/utils/programs.py
ccordoba12: So I think you need to run two `run_program` calls: one to detect if the module is present (simply by trying to import it); and the second trying to get its version (which should be assigned as `None` if that call returns an error).

spyder/utils/programs.py
```diff
@@ -138,6 +138,19 @@ def alter_subprocess_kwargs_by_platform(**kwargs):
         # We "or" them together
         CONSOLE_CREATION_FLAGS |= CREATE_NO_WINDOW
         kwargs.setdefault('creationflags', CONSOLE_CREATION_FLAGS)
+
+        # ensure Windows subprocess environment has SYSTEMROOT, but variable
+        # is case insensitive
+        if ((kwargs.get('env') is not None) and
+            ('SYSTEMROOT' not in map(str.upper, kwargs['env'].keys()))):
```
ccordoba12, line 145: This is very convoluted. Please move the detection of `SYSTEMROOT`among the env vars to a local variable above this `if`.

ccordoba12 (commented):
Hey @mrclary, thanks for your patience with this PR! I left an initial review for you.

spyder/utils/programs.py
mrclary: Okay, I'll look into it.

spyder/utils/programs.py
mrclary: Okay, I'll look at this.

spyder/utils/programs.py
mrclary: Okay, I'll look at this.

spyder/utils/programs.py
mrclary: Okay, will do.

mrclary: I must check that ``env`` is in ``kwargs`` first, so I cannot move the check for ``SYSTEMROOT`` above the ``if`` statement. But I tried to clean up the code so that it was more readable.

spyder/utils/programs.py
mrclary: Done.

spyder/utils/programs.py
mrclary: Done

spyder/utils/programs.py
mrclary: Done

spyder/utils/programs.py
mrclary: I did not need to ``run_program`` twice, but just improved ``cmd``

ccordoba12 (commented):
Thanks @mrclary, I left a couple of suggestions to improve the style of your code and fix our tests.

spyder/utils/programs.py
```diff
@@ -138,6 +138,20 @@ def alter_subprocess_kwargs_by_platform(**kwargs):
         # We "or" them together
         CONSOLE_CREATION_FLAGS |= CREATE_NO_WINDOW
         kwargs.setdefault('creationflags', CONSOLE_CREATION_FLAGS)
+
+        # ensure Windows subprocess environment has SYSTEMROOT
+        if (kwargs.get('env') is not None):
+            # Is SYSTEMROOT in env? case insensitive
+            if ('SYSTEMROOT' not in map(str.upper, kwargs['env'].keys())):
```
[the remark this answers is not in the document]
mrclary: Sorry about that. Fixed.

spyder/utils/programs.py
```diff
@@ -809,92 +823,77 @@ def get_module_version(module_name):
...
-            f = tempfile.NamedTemporaryFile('wt', suffix='.py',
-                                            dir=get_temp_dir(), delete=False)
+            cmd = dedent('''
+                         try:
+                             import %s as mod
+                         except Exception:
+                             print('No Module')
+                         print(getattr(mod, '__version__',
+                                       getattr(mod, 'VERSION', None)))
+                         ''' % module_name)
```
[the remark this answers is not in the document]
mrclary: Done. Thanks for diagnosing that test failure; I never would have gotten that.

ccordoba12 (approved):
Thanks @mrclary, looks good to me now.

PR #12232, opened 2020-04-09
PR: Display interpreter status if Spyder not launched from conda envrionment (Mac app)

mrclary:
* display interpreter status in statusbar, even if it isn't a conda interpreter or spyder isn't started from conda environment.

## Description of Changes
The conda status bar no longer tests if Spyder's python interpreter is from a conda environment.

* [x] Wrote at least one-line docstrings (for any new functions)
* [x] Added unit test(s) covering the changes (if testable)

* [x] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

Currently, when updating the conda status in the status bar, Spyder's python environment is tested for being a conda environment rather than the iPython console's interpreter environment.
When Spyder is launched from a non-conda environment (such as in a standalone Mac application), the iPython console interpreter status is not displayed.
This PR fixes that.

### Issue(s) Resolved

Fixes spyder-ide/mac-application#6

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct:
@mrclary

PR #12011, opened 2020-03-27
PR: Initial changes to create again a macOS standalone app

mrclary:
* [x] Wrote at least one-line docstrings (for any new functions)
* [x] Added unit test(s) covering the changes (if testable)

* [x] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

Move `create_app.py` to spyder-ide/mac-application

### Issue(s) Resolved

Addresses spyder-ide/mac-application#2

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct:
@mrclary

PR #12236, opened 2020-04-09
PR: Start kernels and Jedi envs with explicit environment (Mac app)

mrclary:
## Description of Changes

Currently, `spyder_kernel` shell environment variables are inherited from Spyder's environment _implicitly_. This prevents `spyder_kernel` from launching if Spyder is not launched from a conda environment (such as in a standalone Mac application). This PR solves this issue by _explicitly_ passing the desired environment variables to the `spyder_kernel` via `kernel_spec`.

### Issue(s) Resolved

This allows a stand-alone Mac application to work properly.

### Affirmation

@mrclary

### DIFF
--- a/spyder/config/lsp.py
+++ b/spyder/config/lsp.py
@@ -67,6 +67,7 @@ PYTHON_CONFIG = {
                 'jedi': {
                     'environment': None,
                     'extra_paths': None,
+                    'env_vars': None,
                 },
                 'jedi_completion': {
                     'enabled': True,
--- a/spyder/plugins/completion/languageserver/plugin.py
+++ b/spyder/plugins/completion/languageserver/plugin.py
@@ -21,7 +21,8 @@ from qtpy.QtCore import Slot, QTimer
 from qtpy.QtWidgets import QMessageBox
 
 # Local imports
-from spyder.config.base import _, get_conf_path, running_under_pytest
+from spyder.config.base import (_, get_conf_path, running_under_pytest,
+                                running_in_mac_app)
 from spyder.config.lsp import PYTHON_CONFIG
 from spyder.config.manager import CONF
 from spyder.api.completion import SpyderCompletionPlugin
@@ -654,13 +655,21 @@ class LanguageServerPlugin(SpyderCompletionPlugin):
         # Jedi configuration
         if self.get_option('default', section='main_interpreter'):
             environment = None
+            env_vars = None
         else:
             environment = self.get_option('custom_interpreter',
                                           section='main_interpreter')
+            env_vars = os.environ.copy()
+            # external interpreter should not use internal PYTHONPATH
+            env_vars.pop('PYTHONPATH', None)
+            if running_in_mac_app():
+                env_vars.pop('PYTHONHOME', None)
+
         jedi = {
             'environment': environment,
             'extra_paths': self.get_option('spyder_pythonpath',
                                            section='main', default=[]),
+            'env_vars': env_vars,
         }
         jedi_completion = {
             'enabled': self.get_option('code_completion'),
--- a/spyder/plugins/completion/languageserver/plugin.py
+++ b/spyder/plugins/completion/languageserver/plugin.py
@@ -14,7 +14,6 @@ import functools
 import logging
 import os
 import os.path as osp
-import sys
 
 # Third-party imports
 from qtpy.QtCore import Slot, QTimer
@@ -26,7 +25,7 @@ from spyder.config.base import (_, get_conf_path, running_under_pytest,
 from spyder.config.lsp import PYTHON_CONFIG
 from spyder.config.manager import CONF
 from spyder.api.completion import SpyderCompletionPlugin
-from spyder.utils.misc import check_connection_port, getcwd_or_home
+from spyder.utils.misc import check_connection_port
 from spyder.plugins.completion.languageserver import LSP_LANGUAGES
 from spyder.plugins.completion.languageserver.client import LSPClient
 from spyder.plugins.completion.languageserver.confpage import (
--- a/spyder/plugins/ipythonconsole/utils/kernelspec.py
+++ b/spyder/plugins/ipythonconsole/utils/kernelspec.py
@@ -18,7 +18,8 @@ import sys
 from jupyter_client.kernelspec import KernelSpec
 
 # Local imports
-from spyder.config.base import DEV, running_under_pytest, SAFE_MODE
+from spyder.config.base import (DEV, running_under_pytest, SAFE_MODE,
+                                running_in_mac_app)
 from spyder.config.manager import CONF
 from spyder.py3compat import PY2, iteritems, to_binary_string, to_text_string
 from spyder.utils.conda import (add_quotes, get_conda_activation_script,
@@ -84,9 +85,6 @@ class SpyderKernelSpec(KernelSpec):
         if CONF.get('main_interpreter', 'default'):
             pyexec = get_python_executable()
         else:
-            # Avoid IPython adding the virtualenv on which Spyder is running
-            # to the kernel sys.path
-            os.environ.pop('VIRTUAL_ENV', None)
             pyexec = CONF.get('main_interpreter', 'executable')
             if not is_python_interpreter(pyexec):
                 pyexec = get_python_executable()
@@ -106,7 +104,7 @@ class SpyderKernelSpec(KernelSpec):
             # the activation scripts at spyder/plugins/ipythonconsole/scripts/
             kernel_cmd = [
                 get_activation_script(),  # This is bundled with Spyder
-                get_conda_activation_script(),
+                get_conda_activation_script(pyexec),
                 get_conda_env_path(pyexec),  # Might be external
                 pyexec,
                 '{connection_file}',
@@ -127,6 +125,12 @@ class SpyderKernelSpec(KernelSpec):
     def env(self):
         """Env vars for kernels"""
         default_interpreter = CONF.get('main_interpreter', 'default')
+        env_vars = os.environ.copy()
+
+        # Avoid IPython adding the virtualenv on which Spyder is running
+        # to the kernel sys.path
+        env_vars.pop('VIRTUAL_ENV', None)
+
         pathlist = CONF.get('main', 'spyder_pythonpath', default=[])
 
         # Add spyder-kernels subrepo path to pathlist
@@ -146,6 +150,9 @@ class SpyderKernelSpec(KernelSpec):
         pypath = add_pathlist_to_PYTHONPATH([], pathlist, ipyconsole=True,
                                             drop_env=False)
 
+        # Add our PYTHONPATH to env_vars
+        env_vars.update(pypath)
+
         # Environment variables that we need to pass to our sitecustomize
         umr_namelist = CONF.get('main_interpreter', 'umr/namelist')
 
@@ -159,7 +166,7 @@ class SpyderKernelSpec(KernelSpec):
             if original_list != umr_namelist:
                 CONF.set('main_interpreter', 'umr/namelist', umr_namelist)
 
-        env_vars = {
+        env_vars.update({
             'SPY_EXTERNAL_INTERPRETER': not default_interpreter,
             'SPY_UMR_ENABLED': CONF.get('main_interpreter', 'umr/enabled'),
             'SPY_UMR_VERBOSE': CONF.get('main_interpreter', 'umr/verbose'),
@@ -186,7 +193,7 @@ class SpyderKernelSpec(KernelSpec):
             'SPY_SYMPY_O': CONF.get('ipython_console', 'symbolic_math'),
             'SPY_TESTING': running_under_pytest() or SAFE_MODE,
             'SPY_HIDE_CMD': CONF.get('ipython_console', 'hide_cmd_windows')
-        }
+        })
 
         if self.is_pylab is True:
             env_vars['SPY_AUTOLOAD_PYLAB_O'] = True
@@ -201,8 +208,9 @@ class SpyderKernelSpec(KernelSpec):
             env_vars['SPY_SYMPY_O'] = False
             env_vars['SPY_RUN_CYTHON'] = True
 
-        # Add our PYTHONPATH to env_vars
-        env_vars.update(pypath)
+        # macOS app considerations
+        if running_in_mac_app() and not default_interpreter:
+            env_vars.pop('PYTHONHOME', None)
 
         # Making all env_vars strings
         clean_env_vars = clean_env(env_vars)
--- a/spyder/plugins/ipythonconsole/utils/kernelspec.py
+++ b/spyder/plugins/ipythonconsole/utils/kernelspec.py
@@ -21,10 +21,9 @@ from jupyter_client.kernelspec import KernelSpec
 from spyder.config.base import (DEV, running_under_pytest, SAFE_MODE,
                                 running_in_mac_app)
 from spyder.config.manager import CONF
-from spyder.py3compat import PY2, iteritems, to_binary_string, to_text_string
+from spyder.py3compat import PY2
 from spyder.utils.conda import (add_quotes, get_conda_activation_script,
                                 get_conda_env_path, is_conda_env)
-from spyder.utils.encoding import to_unicode_from_fs
 from spyder.utils.environ import clean_env
 from spyder.utils.misc import add_pathlist_to_PYTHONPATH, get_python_executable
 from spyder.utils.programs import is_python_interpreter
--- a/spyder/plugins/ipythonconsole/utils/kernelspec.py
+++ b/spyder/plugins/ipythonconsole/utils/kernelspec.py
@@ -148,7 +148,7 @@ class SpyderKernelSpec(KernelSpec):
 
         # Create PYTHONPATH env entry to add it to the kernel
         pypath = add_pathlist_to_PYTHONPATH([], pathlist, ipyconsole=True,
-                                            drop_env=False)
+                                            drop_env=True)
 
         # Add our PYTHONPATH to env_vars
         env_vars.update(pypath)
--- a/spyder/plugins/ipythonconsole/utils/kernelspec.py
+++ b/spyder/plugins/ipythonconsole/utils/kernelspec.py
@@ -21,7 +21,6 @@ from jupyter_client.kernelspec import KernelSpec
 from spyder.config.base import (DEV, running_under_pytest, SAFE_MODE,
                                 running_in_mac_app)
 from spyder.config.manager import CONF
-from spyder.py3compat import PY2
 from spyder.utils.conda import (add_quotes, get_conda_activation_script,
                                 get_conda_env_path, is_conda_env)
 from spyder.utils.environ import clean_env
@@ -73,8 +72,8 @@ class SpyderKernelSpec(KernelSpec):
         self.is_pylab = is_pylab
         self.is_sympy = is_sympy
 
-        self.display_name = 'Python 2 (Spyder)' if PY2 else 'Python 3 (Spyder)'
-        self.language = 'python2' if PY2 else 'python3'
+        self.display_name = 'Python 3 (Spyder)'
+        self.language = 'python3'
         self.resource_dir = ''
 
     @property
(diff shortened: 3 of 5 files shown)


spyder/plugins/ipythonconsole/utils/kernelspec.py
```diff
@@ -106,7 +106,7 @@ def argv(self):
             # the activation scripts at spyder/plugins/ipythonconsole/scripts/
             kernel_cmd = [
                 get_activation_script(),  # This is bundled with Spyder
-                get_conda_activation_script(),
+                get_conda_activation_script(pyexec),
```
goanpeca, line 109: why do we need this change 🙃 ?

spyder/plugins/ipythonconsole/utils/kernelspec.py
```diff
@@ -201,7 +197,19 @@ def env(self):
...
+        # settings
+        if default_interpreter and os.environ.get('PYTHONHOME'):
+            # default interpreter should keep PYTHONHOME
+            env_vars.update({'PYTHONHOME': os.environ.get('PYTHONHOME')})
+        elif not default_interpreter and os.environ.get('HOME'):
+            # Add HOME to env_vars
+            env_vars.update({'HOME': os.environ.get('HOME')})
+
         # Add our PYTHONPATH to env_vars
+        # spyder env python should be omitted in kernels
```
goanpeca, line 210: why is this the case?

spyder/plugins/ipythonconsole/utils/kernelspec.py
mrclary: If `pyexec` is not provided, then the conda activation script associated with Spyder's launch environment is searched for (specifically `sys.prefx`). This is a problem if Spyder is not started from a conda environment (Mac app); but also, more generally, this will ensure that the conda activation script associated with the desired python executable is returned. For example, if Spyder is started from a conda environment, but the provided interpreter is from an environment in a miniconda installation, the miniconda activation script should be returned, not the anaconda one.

spyder/plugins/ipythonconsole/utils/kernelspec.py
mrclary: This refers to the PYTHONPATH variable in Spyder's launch environment. When starting a new iPython kernel, we don't want to inherit the PYTHONPATH from Spyder's launch environment, which could give access to python packages the user does not expect and pollutes the PYTHONPATH. Or worse: accessing the _same_ packages of a different version unwittingly. I think the PYTHONPATH should be comprised _only_ of those explicitly provided by the user in the Python Path Manager and possibly a user's login shell (~/.bash_profile, etc.).

goanpeca: So in this case it sounds like this is a bug in spyder and not a MacApp thing...

spyder/plugins/ipythonconsole/utils/kernelspec.py
goanpeca: Understood

spyder/plugins/ipythonconsole/utils/kernelspec.py
mrclary: In this specific case, I believe so. I think the Mac app effort has just revealed few "cross-environment" issues.

goanpeca: in that case this needs a test ;-)

mrclary: Okay, I can work on a test...

mrclary: Oh, I forgot that I had already modified `test_preserve_pypath` in this PR. Is that what you're looking for?

mrclary:
All, thank you so much for your patience and feedback over the last several weeks. I know you are busy, so I understand that you may not be able to help out right away, but when you get a chance to take a look at these I'd appreciate it.

**Linux tests / Py3.7, conda, fast & Linux tests / Py3.7, pip, fast**
@ccordoba12: The failure here is
`spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py::test_auto_backend`
`ImportError: Cannot load backend 'Qt5Agg' which requires the 'qt5' interactive framework, as 'headless' is currently running`.
Test passes locally on Mac, don't know what the issue is for Linux.

**Win tests / Py3.7, conda, fast**
@goanpeca : The failure here is
`spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py::test_conda_env_activation`
`AssertionError: waitUntil timed out in 20000 miliseconds`
Test passes locally on Mac, don't know what the issue is for Windows.

**Linux tests / Py3.7, conda, slow & Mac tests / Py3.7, conda, slow & Linux tests / Py3.7, pip, slow & Win tests / Py3.7, conda, slow**
@goanpeca @impact27: The failure here is
`spyder/app/tests/test_mainwindow.py::test_single_instance_and_edit_magic`
`AttributeError: 'NoneType' object has no attribute 'close'`
This is run in CI only, so I can't test locally.

mrclary:
@goanpeca, the same failure above for `test_conda_env_activation` occurs in #12237, but for Linux only. I wonder if in both cases an environment variable is causing the problem...
Here, perhaps windows is missing a required environment variable, since we're passing environments explicitly. For #12237, perhaps Linux has an incorrect environment variable from the `-l` bash option; I don't know why Mac would pass and Linux fail, though.

goanpeca:
Will take a look at the failing tests @mrclary, thanks!

goanpeca:
>@goanpeca, the same failure above for test_conda_env_activation occurs in #12237, but for Linux only. I wonder if in both cases an environment variable is causing the problem...
Here, perhaps windows is missing a required environment variable, since we're passing environments explicitly. For #12237, perhaps Linux has an incorrect environment variable from the -l bash option; I don't know why Mac would pass and Linux fail, though.

Could you add some prints )before and after) of the environment on the tests that way we can see on the fail, what the env is before and after?

mrclary:
@goanpeca, sure, I'll do that.

mrclary:
@goanpeca , I'm having some trouble with this. I can't get pytest to capture any of the stdout from print statements, so I have no idea whether what I pushed will be helpful at all. I tried printing the `os.environ` and the `kernel_spec.env` just prior to the `start_kernel` call in `start_new_kernel` in `spyder/app/tests/test_mainwindow.py` as well as in `create_kernel_manager_and_kernel_client` in ipythonconsole plugins.

The test that is failing is `test_single_instance_and_edit_magic`, which is CI only, so I can't test it locally, but should be using the ipythonconsole when creating the main window.

mrclary:
It looks like this passes with all changes in this PR except when passing the explicit `kernel_spec.env` to `kernel_spec.start_kernel`.
I'm in the process of installing Linux and Windows VMs on my Mac in order to test these systems locally.

goanpeca:
Thanks a lot for this work .

It will pay off big time!

mrclary:
@goanpeca, wondering if you can provide some feedback on the following.
I've got my Linux VM setup and was able to fix the failure for
`spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py::test_auto_backend`.

Then I moved on to the next failed test
`spyder/app/tests/test_mainwindow.py::test_run_cython_code`
which failed because `unable to execute 'gcc': No such file or directory`.
This error occurred for me on both this branch _and_ 4.x branch. I had to `sudo apt-get install gcc` for it to work, but then it works on both branches locally, yet it still fails on the server. I don't see that gcc is installed explicitly in the server environment, and I'm surprised that my local Ubuntu (18.04.4) didn't already have it installed...

Additionally, the following 6 tests fail for me locally, but _pass_ on the server (?). What am I missing?
* `spyder/plugins/editor/panels/tests/test_scrollflag::test_range_indicator_visible_on_hover_only`
* `spyder/plugins/editor/panels/tests/test_scrollflag::test_range_indicator_alt_modifier_response`
* `spyder/plugins/editor/widgets/tests/test_shortcuts::test_delete_line_shortcut`
* `spyder/plugins/editor/widgets/tests/test_shortcuts::test_go_to_line_shortcut`
* `spyder/plugins/editor/widgets/tests/test_shortcuts::test_transform_to_lowercase_shortcut`
* `spyder/plugins/editor/widgets/tests/test_shortcuts::test_transform_to_uppercase_shortcut`

mrclary:
@goanpeca, looks like all tests now pass!
Unfortunately, I don't know exactly what changed :-/
I only cleaned up the kernelspec a bit...

goanpeca:
>looks like all tests now pass!

Awesome news :-)

ccordoba12 (commented):
Thanks @mrclary for your work on this. I can help you to fix the failing test once everything is ready to be merged.

ccordoba12:
> I can help you to fix the failing test once everything is ready to be merged.

Sorry, after looking at the failing test, this looks quite serious: the kernel seems unable to restart automatically after it dies. I don't know what's causing that.

spyder/config/lsp.py
```diff
@@ -67,6 +67,9 @@
                 'jedi': {
                     'environment': None,
                     'extra_paths': None,
+                    # jedi should have clean environment to work properly with
+                    # external interpreters
+                    'env_vars': {},
```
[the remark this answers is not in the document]
mrclary: Yes. Before Jedi 0.17.2 it always used ``os.environ`` from the launching environment (Spyder's runtime environment in our case) and did not permit sending environment variables. Jedi 0.17.2 added support for passing environment variables to Jedi's analysis environment, but the default behavior is to revert to ``os.environ``; we want to use a clean environment, so we must pass an empty dictionary. This is the exact behavior of subprocess.Popen (which Jedi uses).

Note that ``env_vars`` should be updated with the user's PYTHONPATH as specified in Spyder's Python Path Manager, i.e. we want Jedi completions to look there as well.

spyder/plugins/ipythonconsole/utils/tests/test_spyder_kernel.py
```diff
@@ -20,7 +20,7 @@
 @pytest.mark.parametrize('default_interpreter', [True, False])
 def test_preserve_pypath(tmpdir, default_interpreter):
```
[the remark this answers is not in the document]
mrclary: Done.

spyder/plugins/completion/languageserver/plugin.py
```diff
@@ -649,6 +649,9 @@ def generate_python_config(self):
             'environment': environment,
             'extra_paths': self.get_option('spyder_pythonpath',
                                            section='main', default=[]),
+            # jedi should have clean environment to work properly with external
+            # interpreters
+            'env_vars': {},
```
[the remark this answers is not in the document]
mrclary: Same as above.

spyder/app/tests/test_mainwindow.py
```diff
@@ -261,7 +261,7 @@ def test_single_instance_and_edit_magic(main_window, qtbot, tmpdir):
     # Test single instance
     with qtbot.waitSignal(shell.executed, timeout=2000):
         shell.execute(lock_code)
-    assert not shell.get_value('lock_created')
+    assert shell.get_value('lock_created')
```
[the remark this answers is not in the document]
mrclary: Actually, I don't remember.
My commit message for this change says "* why was 'lock_created' value supposed to be False?"
It must have been failing tests for some reason?
I'll change it back and we can see what happens.

spyder/plugins/ipythonconsole/utils/kernelspec.py
```diff
@@ -84,41 +55,30 @@ def argv(self):
...
                 pyexec = get_python_executable()
                 CONF.set('main_interpreter', 'executable', '')
                 CONF.set('main_interpreter', 'default', True)
                 CONF.set('main_interpreter', 'custom', False)
 
-        # Part of spyder-ide/spyder#11819
-        is_different = is_different_interpreter(pyexec)
-
         # Command used to start kernels
-        if is_different and is_conda_env(pyexec=pyexec):
```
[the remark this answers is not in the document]
mrclary: All iPython consoles should be treated with parity, i.e. it shouldn't matter whether they share the same interpreter location on disk as Spyder's runtime interpreter. In all cases, they should be launched with clean environments and independent of Spyder's runtime environment. Since ``is_different`` only tells me whether the new iPython console interpreter is the same as Spyder's runtime interpreter, it serves no purpose; the ``spyder_kernel.console`` must still be launched with the appropriate script and environment variables.

ccordoba12:
@mrclary, please rebase this one to remove the commits related to PR #12235.

mrclary:
Okay, upon investigation I've discovered that in ``test_kernel_kill``, the kernel fails to restart. This also seems to be an issue when killing the python process of any console started via shell script, including the preexisting ``conda_activate.sh``. This does not seem to affect kernel restarts initiated via Spyder's menu "Consoles"->"Restart kernel". This only occurs if the process is killed externally such as in the unit test (``os.kill(os.getpid(), signal.SIGTERM)``) or via a terminal.

What, exactly, are we trying to test? Restarting the console by Spyder's menu and/or shortcut? Or restarting the console when the process is killed by external means?

mrclary:
Steps to reproduce manually:
1. checkout latest 4.x Spyder commit prior to #12011, #12232, #12235 (c1da08b7d2826fe4b1a04ab70961ecbd65926bb5)
1. create a conda bootstrap environment from which to launch Spyder
1. create another conda environment with spyder-kernels=1.9.4. Use this environment for your iPython Console
1. enter ``import os, signal; os.kill(os.getpid(), signal.SIGTERM)`` in the console, or get the PID (``os.getpid()``) and kill in a terminal (``$ kill <PID>``). The console will fail to restart
1. close the console window or use Spyder's menu "Consoles"->"Restart kernel". The console will successfully restart.

It seems that in order to properly restart the kernel from a kill process, the parent (bash) process must be killed first.
``import os, signal; os.kill(os.getppid(), signal.SIGTERM); os.kill(os.getpid(), signal.SIGTERM)``
We'll see how the update to this test affects the other operating systems...

spyder/app/tests/test_mainwindow.py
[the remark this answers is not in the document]
mrclary: Okay, it looks like ``spyder/app/tests/test_mainwindow.py::test_single_instance_and_edit_magic`` fails due to ``main_window.window`` is ``NoneType``. Don't know what could be causing this.

spyder/plugins/ipythonconsole/utils/kernelspec.py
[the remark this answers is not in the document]
ccordoba12: > All iPython consoles should be treated with parity, i.e. it shouldn't matter whether they share the same interpreter location on disk as Spyder's runtime interpreter

This sounds good in principle but it introduces some headaches in practice that I would like to avoid. For instance, with this approach same-interpreter kernels won't restart automatically if they die while running your code.

So I'd prefer to maintain this difference, just to avoid the risk of a very disruptive change in the way we handle kernels in 4.2 (we had that in 4.1.0 and it was a nightmare of bug reports).

ccordoba12:
> Okay, upon investigation I've discovered that in test_kernel_kill, the kernel fails to restart. This also seems to be an issue when killing the python process of any console started via shell script, including the preexisting conda_activate.sh

Yes, that's correct and I was unaware of that problem.

> What, exactly, are we trying to test? Restarting the console by Spyder's menu and/or shortcut? Or restarting the console when the process is killed by external means?

The second. If monitored properly, the kernel can restart automatically if it dies while running code.

> It seems that in order to properly restart the kernel from a kill process, the parent (bash) process must be killed first.

This is not the right approach. I think it's very hard for the parent (i.e. the bash/batch script that activates the kernel) to die, unless you look for it and kill it intentionally. But it's sort of common for the kernel (i.e. its children) to die, when users execute code that causes that (like `import os, signal; os.kill(os.getpid(), signal.SIGTERM`).

So we need to properly monitor both parent and children for sudden deaths. If that's detected for any of them, the kernel will be restarted it automatically by `jupyter_client` (one of the packages the kernel depends on).

This monitoring can be done using `psutil`, but I'm not sure how reliable it is in all OSes (especially on Windows). That's why I'd like to maintain the difference between same-interpreter and external-interpreter kernels for 4.2

ccordoba12:
In my last comment, I forgot to mention that right now we're only monitoring the parent.

mrclary:
@ccordoba12,
Okay, so it seems to me that the best approach, then, is to move away from external scripts altogether. We only had an external script for conda before, but I don't think we have to do it that way. Any interpreter can use the correct environment by supplying the appropriate environment variables and we should not *have* to use the activation commands (e.g. ``source path/to/conda/activate env``).

In this way we could still obtain parity treatment of different interpreters and therefore avoid the kernel restart issues for all interpreters.

I'll push up some commits that implement this and you can tell me what you think.

ccordoba12:
Hold on a bit. First, conda does a lot at activation time (not just setting env vars), so we still need those scripts to start kernels in conda envs.

And second, I think I found a reliable way to monitor the kernel with psutil. So I'll push a commit here with that later today.

ccordoba12:
I managed to solve the bug you found and a similar one. However, I've been thinking hard about this and what you propose in this PR is more than what we'd planned to have for 4.2. Therefore, I'd like to limit its scope.

My rationale is the following:

* Some libraries require environment variables to work (I think CUDA is one example, but I'm sure there are more). I understand there's no way for the new self-contained apps to pick them up. But people could still use Anaconda or pip to install Spyder, and start it from a system terminal to get them. However, by merging this PR we'd prevent that and I don't think it's fair.
* Therefore, we'd need to provide a graphical way to pass env vars to the kernel. That would be nice to have, but it's out of the scope for 4.2.
* Another option could be to source `~/.profile`, `/etc/profile` and/or `~/.bashrc`. However, that could introduce bugs or unexpected results. There's also the possibility that people don't want that to be done automatically, so giving them a graphical way to handle env vars would be the best option, in my opinion.
* I feel the most dangerous env var (the one the installers declare and we don't want to pass to kernels) is PYTHONPATH. So we could simply drop it in our kernel spec, instead of dropping all of them.
* That would also allow us to drop the activation step for same-interpreter kernels. As I mentioned before, that could introduce unsuspected bugs so I'd prefer to avoid it.

I'm open to revisit this decision for Spyder 5, but for now to I'd like to:

* Only drop PYTHONPATH.
* Don't activate same-interpreter kernels.

@mrclary, would that be enough for the macOS app? I know it is for the Windows app because @dalthviz already tried it.

mrclary:
> I'm open to revisit this decision for Spyder 5, but for now to I'd like to:
> - Only drop PYTHONPATH.
> - Don't activate same-interpreter kernels.
> 
> 
> @mrclary, would that be enough for the macOS app? I know it is for the Windows app because @dalthviz already tried it.

Okay, I'll rework this PR to leave Spyder's existing paradigm, but use running_in_mac_app if necessary to get the macOS app to work. Give me a few days to make the changes.

When we're ready to revisit this for Spyder 5, we can look at the user's ability to better control kernel environment variables and perhaps harmonize and unify interpreter startup.

ccordoba12:
> Okay, I'll rework this PR to leave Spyder's existing paradigm, but use running_in_mac_app if necessary to get the macOS app to work. Give me a few days to make the changes.

No problem and thanks for understanding.

> When we're ready to revisit this for Spyder 5, we can look at the user's ability to better control kernel environment variables and perhaps harmonize and unify interpreter startup.

Agreed.

mrclary:
@ccordoba12,
Okay, I've been able to revise this PR per our discussion and the macOS app appears to be working properly.
Now, let's see if the tests pass...

mrclary:
It looks like only conda slow on Windows fails now, but it fails early.
The failure is at ``spyder/plugins/editor/widgets/tests/test_introspection.py::test_completions_environment`` with ``pytestqt.exceptions.TimeoutError: Signal sig_show_completions(PyQt_PyObject) not emitted after 10000 ms``.

I suspect an issue with Windows environment variables. If anyone can take a look, I'd appreciate it. My Windows VM is not currently available, so I don't know how soon I can look at it myself.

ccordoba12:
> Okay, I've been able to revise this PR per our discussion and the macOS app appears to be working properly.

Great! I took a quick look and everything seems fine now.

> I suspect an issue with Windows environment variables

Yep, me too. I think we need to pass at least `USERPROFILE`. Also, probably it's a good idea to pass `HOME` on Linux and macOS. My guess is that that's necessary for Jedi to use Parso's cache on different interpreters.

mrclary:
Okay

> Yep, me too. I think we need to pass at least `USERPROFILE`. Also, probably it's a good idea to pass `HOME` on Linux and macOS. My guess is that that's necessary for Jedi to use Parso's cache on different interpreters.

That's a good idea. I'll push up some changes and see what you think.

mrclary:
Are we dropping support for Python 2?

ccordoba12:
> Are we dropping support for Python 2?

Yes, 4.2.0 is no longer going to be compatible with Python 2.

ccordoba12 (approved):

PR #12237, opened 2020-04-09
WIP: MacApp: Use login-shell option on spyder_kernel launch

mrclary:
## Description of Changes
Use login shell option in shell script to start `spyder_kernel` for both internal and external iPython console interpreters.

By adding the login shell option, a user's local environment variables will be passed to the iPython console interpreter environment. This is required when using external applications such as LaTeX or MDSPlus.

This is an issue when Spyder is not launched from a login-shell environment, such as with a stand-alone Mac application.

### Affirmation

@mrclary

### DIFF
--- a/spyder/config/base.py
+++ b/spyder/config/base.py
@@ -533,17 +533,21 @@ else:
     MAC_APP_NAME = 'Spyder-Py2.app'
 
 
-def running_in_mac_app():
+def running_in_mac_app(pyexec=None):
     """
-    Check if Spyder is running inside an app on macOS.
+    Check if Python executable is located inside a standalone Mac app.
 
-    Check if the app is a stand-alone app.
-    This means this file is located inside 'Spyder.app' and not in the
-    python path.
-    This is important for example for the single_instance option.
+    If no executable is provided, the default will check `sys.executable`, i.e.
+    whether Spyder is running from a standalone Mac app.
+
+    This is important for example for the single_instance option and the
+    interpreter status in the statusbar.
     """
+    if pyexec is None:
+        pyexec = sys.executable
+
     if sys.platform == "darwin":
-        if MAC_APP_NAME not in __file__:
+        if MAC_APP_NAME not in pyexec:
             return False
         return True
     else:
--- a/spyder/utils/programs.py
+++ b/spyder/utils/programs.py
@@ -138,6 +138,19 @@ def alter_subprocess_kwargs_by_platform(**kwargs):
         # We "or" them together
         CONSOLE_CREATION_FLAGS |= CREATE_NO_WINDOW
         kwargs.setdefault('creationflags', CONSOLE_CREATION_FLAGS)
+
+        # ensure Windows subprocess environment has SYSTEMROOT, but variable
+        # is case insensitive
+        if ((kwargs.get('env') is not None) and
+            ('SYSTEMROOT' not in map(str.upper, kwargs['env'].keys()))):
+            sys_root_key = None
+            for k, v in os.environ.items():
+                if 'SYSTEMROOT' == k.upper():
+                    sys_root_key = k
+                    break  # don't risk multiple values
+            if sys_root_key is not None:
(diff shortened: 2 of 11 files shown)


ccordoba12:
@mrclary, I'm going to cancel your builds in Azure for now because I'm trying to fix failures in our tests on Windows.

I'll restart them once I managed to do that. Also, it'd be great if you could wait before pushing more commits until I let you know things are good to go again, so I don't have to keep doing this manually.

Thanks!

spyder/utils/programs.py
```diff
@@ -138,6 +138,19 @@ def alter_subprocess_kwargs_by_platform(**kwargs):
         # We "or" them together
         CONSOLE_CREATION_FLAGS |= CREATE_NO_WINDOW
         kwargs.setdefault('creationflags', CONSOLE_CREATION_FLAGS)
+
+        # ensure Windows subprocess environment has SYSTEMROOT, but variable
+        # is case insensitive
+        if ((kwargs.get('env') is not None) and
+            ('SYSTEMROOT' not in map(str.upper, kwargs['env'].keys()))):
```
goanpeca, line 145: What is this doing 🙃 ?

spyder/config/base.py
```diff
@@ -536,7 +536,7 @@ def translate_gettext(x):
     MAC_APP_NAME = 'Spyder-Py2.app'
 
 
-def running_in_mac_app():
+def running_in_mac_app(pyexec=None):
```
goanpeca, line 539: Could we add an explanation in the docstring why this is needed?

goanpeca:
Will look at why the tests are failing

spyder/utils/programs.py
mrclary: When running a subprocess, an 'env' keyword argument can be given which defines the environment variables in the subprocess shell environment. If the 'env' keyword argument is not supplied, the current shell environment is used. For situations where one wants a clean environment, simply providing `env={}` is all that is required. Except for windows. For Windows systems, `systemroot` must be in the environment variables; unfortunately, it's not case sensitive and multiple variations may exist, e.g. `SystemRoot` or `SYSTEMROOT`.

spyder/config/base.py
mrclary: You mean: why the keyword argument input is needed?
Most uses of this function will only need to know if `sys.executable` (Spyder's) is inside a standalone Mac application. However, the interpreter status in the statusbar needs to know if `self._interpreter` is running in a Mac app ("Same as Spyder"), i.e. 'internal', which is not conda!

goanpeca: Ok let;'s addd that nicely to the docstring :-) 

I meant the pyexec. Your explanation is perfect, just need to add it there for context :-)

spyder/utils/programs.py
goanpeca: Understood! Was not clear at first :-)

goanpeca:
@mrclary which PR is next in line and has passing tests (to review and merge) ?

mrclary:
#12011 and #12271 pass and should be next. Then #12236 and #12237, but these still don't pass

mrclary:
@goanpeca, the only failure here is `test_conda_env_activation` which passes on my local Ubuntu, so I don't know what the issue could be.

mrclary:
This PR should follow #12236.

mrclary:
This still passes on my local VM, but now `test_auto_backend` is failing on the server !?
I fixed `test_auto_backend` in #12236; why is it showing up again here?

This PR is the lowest priority of group for the Mac app, so we can defer.

goanpeca:
>This still passes on my local VM, but now test_auto_backend is failing on the server !?
>I fixed test_auto_backend in #12236; why is it showing up again here?

Sometimes the changes make subtle changes in the pytest runs, so I would not worry too much about this. As soon as the other PRs are in we can look again into this one, maybe changing the order in which the failing test is ran.

Thanks again.

ccordoba12:
@mrclary, I think we can close this one, right?

mrclary:
Yep.

PR #12281, opened 2020-04-11
PR: Add skip to builds plus check for skipping if PRs do not change Python files

goanpeca:
## Description of Changes

Add script to skip a build if a commit message contains either the words `[ci skip]` or `[skip ci]` 

Also added skip if no py files where changed on a PR

Right now there is no way to avoid builds from starting but once a skip is set, it will take about 20-30 seconds to finish all.

### Affirmation

### DIFF
--- a/.github/scripts/check_skip.sh
+++ b/.github/scripts/check_skip.sh
@@ -1,0 +1,17 @@
+# Get text for last commit
+LAST_COMMIT_TEXT="$(git log -1 --pretty=format:%s)"
+echo "Last commit text: $LAST_COMMIT_TEXT"
+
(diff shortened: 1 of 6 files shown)


goanpeca:
@ccordoba12 took a different approach and now although a bit annoying with the ifs on every step, this now works as expected.

There are two checks now:

we can use `[ci skip]` or `[skip ci]` to skip all build, but also now we check what files changed against the base branch so that if no `*.py` files where changed, then we also skip the python builds.

We might need to include some other extensions just to be sure.

Thoughts @spyder-ide/core-developers ?

PR #12436, opened 2020-04-21
PR: Use runfile for external spyder-kernels

### LINKED ISSUE #11923: Run in console's namespace with existing kernel not working
I cannot access variables from an existing kernel hooked to spyder from a spyder script.
I am trying to use spyder with an embedded kernel from a C# application and there are things in the kernels namespace I need to access. They are accessible in the console but not from scripts. I'd be happy for any way to accomplish this.

**Repro**

Start a spyder kernel:
      python -m spyder_kernels.console

In spyder connect a console to this existing kernel
Set configuration per file to use the current console and "Run in console's namespace"

In the console set a variable:
In []:   x = 5

Run a script in spyder to print that variable:
print(x}

You get:
   NameError: name 'x' is not defined

This works if I use a console made in spyder but not with an existing kernel.

dalthviz:
## Description of Changes

### Issue(s) Resolved

Fixes #11923 

### Affirmation

### DIFF
--- a/spyder/plugins/ipythonconsole/plugin.py
+++ b/spyder/plugins/ipythonconsole/plugin.py
@@ -454,7 +454,7 @@ class IPythonConsole(SpyderPluginWidget):
                 if console_namespace:
                     line += ", current_namespace=True"
                 line += ")"
-            else: # External kernels: use %run for non spyder-kernels
+            else:  # External kernels, use %run for non spyder-kernels
                 line = "%run "
                 if debug:
                     line += "-d "
--- a/spyder/plugins/ipythonconsole/plugin.py
+++ b/spyder/plugins/ipythonconsole/plugin.py
@@ -442,7 +442,8 @@ class IPythonConsole(SpyderPluginWidget):
 
         if client is not None:
             # Internal kernels, use runfile
-            if client.get_kernel() is not None or client.has_spyder_kernel():
+            if (client.get_kernel() is not None or
+                    client.shellwidget.is_spyder_kernel()):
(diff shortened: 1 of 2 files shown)


spyder/plugins/ipythonconsole/widgets/client.py
```diff
@@ -711,6 +711,10 @@ def set_info_page(self):
                 QUrl.fromLocalFile(self.css_path)
             )
 
+    def has_spyder_kernel(self):
```
ccordoba12, line 714: I think this new method is not necessary.

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -442,7 +442,7 @@ def run_script(self, filename, wdir, args, debug, post_mortem,
 
         if client is not None:
             # Internal kernels, use runfile
-            if client.get_kernel() is not None:
+            if client.get_kernel() is not None or client.has_spyder_kernel():
```
ccordoba12, line 445: Let's change this to

```python
if (client.get_kernel() is not None or
        client.shellwidget.is_spyder_kernel()):
```

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -454,7 +454,7 @@ def run_script(self, filename, wdir, args, debug, post_mortem,
                 if console_namespace:
                     line += ", current_namespace=True"
                 line += ")"
-            else: # External kernels, use %run
+            else:  # External kernels, use %run for non spyder-kernels
```
ccordoba12, line 457: Let's change this to

```python
else:  # External, non spyder-kernels, use %run
```

PR #12453, opened 2020-04-23
PR: Fix unstable completions when using backspace

### LINKED ISSUE #12417: Inconsistencies with the completion widget
I found the following inconsistencies with that widget:

* If you write `os.pat`, then press `Tab` to complete, then one `Backspace`, we hide the widget (but it should remain visible)

    close-widget

* If you write `os.p`, then press `Tab`, then write the letter `a`, then press `Backspace`, the widget remains visible (as it should)

    no-close-widget

* In the same example as before, if you press `Backspace` twice, then we hide the widget but it should remain visible.

* If you write `os.`, then press `Tab`, then write `pa`, then press `Backspace` twice, the widget remains visible too (which is the right behavior)

    no-close-widget-2

----

These small inconsistencies are very, very annoying and we need to fix them

@steff456, please take a look at this one to see how to fix it.

steff456:
## Description of Changes

Remove check of `_last_pressed_text` and length of current text to trigger `document_did_change` and do completions on the fly.

completion

### Issue(s) Resolved

Fixes #12417

### Affirmation

I certify the above statement is true and correct:Steff456

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -4064,14 +4064,16 @@ class CodeEditor(TextEditBaseWidget):
 
         self.document_did_change(text)
 
-        # Perform completion on the fly
-        if self.automatic_completions and not self.in_comment_or_string():
-            # Variables can include numbers and underscores
-            if (text.isalpha() or text.isalnum() or '_' in text
-                    or '.' in text):
-                self.do_completion(automatic=True)
-                self._last_key_pressed_text = ''
-                self._last_pressed_key = None
+        if (len(text) >= self.automatic_completions_after_chars
+                and self._last_key_pressed_text) or key == Qt.Key_Backspace:
+            # Perform completion on the fly
+            if self.automatic_completions and not self.in_comment_or_string():
+                # Variables can include numbers and underscores
+                if (text.isalpha() or text.isalnum() or '_' in text
+                        or '.' in text):
+                    self.do_completion(automatic=True)
+                    self._last_key_pressed_text = ''
+                    self._last_pressed_key = None
 
     def fix_and_strip_indent(self, *args, **kwargs):
         """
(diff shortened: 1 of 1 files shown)


spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -4063,18 +4063,16 @@ def _handle_completions(self):
             cursor.select(QTextCursor.WordUnderCursor)
             text = to_text_string(cursor.selectedText())
 
-        if (len(text) >= self.automatic_completions_after_chars
-                and self._last_key_pressed_text):
-            self.document_did_change(text)
```
ccordoba12, line 4068: Did you read the `git blame` for these lines? I mean, I'd like to know who did them and what was the rationale behind them to understand why they were added in the first place.

steff456: @goanpeca added this lines to the function

goanpeca: >        if (len(text) >= self.automatic_completions_after_chars
>                and self._last_key_pressed_text):

That logic is/was there to ensure the minimum chars to trigger a completionit worked properly . It seems that now that is broken with these changes?

ccordoba12: I think the `document_did_change` statement needs to go outside this if, but perhaps we need to preserve it, as @goanpeca mentioned.

@steff456, what do you think?

steff456: `document_did_change` should be called always, so it should be outside this if

ccordoba12: I agree. Then I think the solution is to leave the current `if` and move `document_did_change` before it.

steff456: If I put the `if`, then there's some cases of the backspace that won't be called because the last instruction is to generate the completions

goanpeca: Ok, that is a corner case that needs to be addressed @steff456 without deleting code that is actually accomplishing a feature.

ccordoba12: Ok, but if we don't leave the if we'll break showing completions after a certain number of characters.

So please try to fix the case when the number of chars is 0. If that's not possible, we'd have to drop that case from this PR.

ccordoba12:
@steff456, I made a last commit to add a couple of places where `did_document_change` was missing and fix the code style of several statements.

I also rebased this PR to avoid the error in our tests I fixed this morning.

PR #12480, opened 2020-04-24
PR: Add an option to filter logger messages when debugging

andfoy:
## Description of Changes
This PR adds the option ``--filter-log`` and the ``SPYDER_FILTER_LOG`` environment variable that receives a set of full-qualified module names separated by space to filter and display the log messages incoming only from those modules. 

This PR also introduces and vendors the [Deterministic Acyclic Finite State Automata](https://github.com/tresoldi/dafsa) data structure used to compress strings by prefix.

### Issue(s) Resolved

Fixes #

### Affirmation


bootstrap.py
```diff
@@ -115,6 +119,10 @@
         sys.exit("ERROR: Can't enable debug mode - Spyder is already imported")
     print("0x. Switching debug mode on")
     os.environ["SPYDER_DEBUG"] = "3"
+    if len(args.filter_log) > 0:
+        print("0x-1. Displaying log messages only from the "
```
ccordoba12, line 123: Please change this to

```python
print("*. Displaying log messages only from the ")
```

spyder/app/utils.py
```diff
@@ -77,15 +82,32 @@ def setup_logging(cli_options):
...
-                            format=log_format,
-                            filename=log_file,
-                            filemode='w+')
+        match_func = lambda x: True
+        if FILTER_NAMES != [''] and len(FILTER_NAMES) > 0:
+            dafsa = DAFSA(FILTER_NAMES)
+            match_func = lambda x: (dafsa.lookup(x, stop_on_prefix=True)
+                                    is not None)
+
+        f = logging.Formatter(log_format)
```
ccordoba12, line 98: Please use a better name for this variable instead of simply `f`.

spyder/app/utils.py
```diff
@@ -77,15 +82,32 @@ def setup_logging(cli_options):
...
+
+        class LSPFilter(logging.Filter):
```
ccordoba12, line 100: Why is this class necessary? Please add a docstring to understand its purpose.

spyder/app/utils.py
```diff
@@ -77,15 +82,32 @@ def setup_logging(cli_options):
...
+            def filter(self, record):
+                return match_func(record.name)
+
+        filter = LSPFilter()
+        root_logger.setLevel(log_level)
+        for h in handlers:
```
ccordoba12, line 106: Please change this to

```python
for handler in handlers:
```

and adjust the code below accordingly.

spyder/app/utils.py
ccordoba12: Also, it's name is not clear to me. Is this related to the LSP? Or does it work as the general filter for our logging messages?

andfoy: Sorry, I left that name when I used it to filter LSP messages, that class is on charge of deciding which messages should be displayed and which not

ccordoba12: Ok, then it could be called `ModuleFilter`.

andfoy:
> I left some small suggestions for you, but I'd also like to see this functionality working not just from bootstrap.py, but also from a regularly installed Spyder. So please add --filter-log to spyder/app/cli_options.py .

We may have some problems adding this options to ``cli_options.py``, as it conflicts with the ``files`` option, which also recieves variable length arguments

ccordoba12:
> We may have some problems adding this option to cli_options.py, as it conflicts with the files option, which also recieves variable length arguments

What do you mean by this?

andfoy:
> What do you mean by this?

The expected argument input for spyder is: ``spyder [--options] files``, where files is a variable list of arguments, that is not declared by a flag in specific, i.e., ``--files``. If we introduce the ``--filter-log`` flag here (which also is variable length), we could have this case ``spyder --filter-log patterns files``  where we cannot distinguish files from logging patterns.

goanpeca:
@andfoy could you add an example fo filter log patterns?

andfoy:
> @andfoy could you add an example fo filter log patterns?

Here are some examples:
* ```python bootstrap.py --filter-log spyder``` would display all the messages produced in Spyder
* ```python bootstrap.py --filter-log spyder.plugins.completions spyder.plugins.editor``` would display all the messages produced by all the modules under spyder.plugins.editor and spyder.plugins.completions

jitseniesen:
> We may have some problems adding this option to `cli_options.py`, as it conflicts with the `files` option, which also recieves variable length arguments

One possibility is to use a different separator for the filter-log option, like a comma. For example, the following command would set up two log filters and opens three files:

```
spyder --filter-log spyder.plugins.completions,spyder.plugins.editor ham.py spam.py eggs.py
```

Another possibility is to require the user to specify the option multiple times (`action='append'`), as in:

```
spyder --filter-log spyder.plugins.completions --filter-log spyder.plugins.editor ham.py spam.py eggs.py
```

andfoy:
> One possibility is to use a different separator for the filter-log option, like a comma. For example, the following command would set up two log filters and opens three files:

I think this is a great suggestion @jitseniesen, what do you think about this @ccordoba12?

ccordoba12:
> I think this is a great suggestion @jitseniesen, what do you think about this @ccordoba12?

I totally agree! Please implement it as part of this PR.

PR #12710, opened 2020-05-13
PR: Fix showing/hiding completions when backspace is pressed

### LINKED ISSUE #12704: Deleting space before a word trigger meaningless completion
Hi spyder devs,
I've find a new way of improving the completion:

## Problem Description
Deleting space before a word trigger meaningless completion.

1. In a new document type `None`;
2. add a space before it;
3. remove this space.

I would expect nothing, I see the completion with all possibilities.

## Versions

* Spyder version: 4.1.3
* Python version: 3.8

I'm sorry, but this time I don't have time to investigate more.

steff456:
## Description of Changes

Add counter to backspace and check if the text is not empty before showing the completions widget.

completions

### Issue(s) Resolved

Fixes #12704 

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -4072,6 +4072,8 @@ class CodeEditor(TextEditBaseWidget):
                 self._last_pressed_key = None
                 return
 
+        # Handle the correct completion when Backspace key is pressed
+        # We should not show the widget if deleting a space before a word.
         if key == Qt.Key_Backspace:
             cursor.setPosition(pos - 1, QTextCursor.MoveAnchor)
             cursor.select(QTextCursor.WordUnderCursor)
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -4072,7 +4072,7 @@ class CodeEditor(TextEditBaseWidget):
                 self._last_pressed_key = None
                 return
 
-        # Handle the correct completion when Backspace key is pressed
+        # Correctly handle completions when Backspace key is pressed.
         # We should not show the widget if deleting a space before a word.
         if key == Qt.Key_Backspace:
             cursor.setPosition(pos - 1, QTextCursor.MoveAnchor)
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -4050,6 +4050,14 @@ class CodeEditor(TextEditBaseWidget):
                 self._last_pressed_key = None
                 return
 
+        cursor.setPosition(pos - 1, QTextCursor.MoveAnchor)
+        cursor.select(QTextCursor.WordUnderCursor)
+        prev_text = to_text_string(cursor.selectedText())
+        cursor.setPosition(pos + 1, QTextCursor.MoveAnchor)
+
+        if prev_text == '' and key == Qt.Key_Backspace:
+            return
+
         # Text might be after a dot '.'
         if text == '':
             cursor.setPosition(pos - 1, QTextCursor.MoveAnchor)
@@ -4069,6 +4077,7 @@ class CodeEditor(TextEditBaseWidget):
 
         is_backspace = (
             self.is_completion_widget_visible() and key == Qt.Key_Backspace)
+
         if (len(text) >= self.automatic_completions_after_chars
                 and self._last_key_pressed_text or is_backspace):
             # Perform completion on the fly
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -364,7 +364,6 @@ class CodeEditor(TextEditBaseWidget):
 
         # Typing keys / handling on the fly completions
         # See: keyPressEvent
-        self._count_backspace = 0
         self._last_key_pressed_text = ''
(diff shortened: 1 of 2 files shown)


ElieGouzien:
Thank for taking car of this issue so fast!

Unfortunately, for now it's not fixing what I described in  #12704.

I might suggest the following behavior for the backspace key, in addition with what already in this pull request:
  * if cursor is at the end of a word, open the completion widget;
  * if cursor in the middle of a word, keep it as it is (open if it is opened (just refresh the propositions), and close if closed) (even if I don't see the point of having completion in the middle of a word, it should happen only when it was previously explicitly requested);
  * if cursor is at the beginning of a word, close it (not sure it worth making the difference with previous point).

ccordoba12:
Sorry @ElieGouzien but we don't agree with your proposal. The thing is showing completions after pressing backspace (even in the middle of a word) is really useful in case you want to fix some incorrectly written method or function, but want to see the completion widget again to understand where your error was or to pick up something different.

Besides, that's how other editors and IDEs work (e.g. VSCode), so we're just following their lead.

ElieGouzien:
@ccordoba12: No problem, that was just a suggestion.

Maybe it would make sense when completing from the middle of a word to have the previous end removed when applying the completion.
I also still think that removing white-space in front of a word should not trigger the completion.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -4067,8 +4074,10 @@ def _handle_completions(self):
 
         self.document_did_change(text)
 
+        is_backspace = (
+            self._count_backspace >= self.automatic_completions_after_chars)
```
ccordoba12, line 4078: Does this mean that you have to press Backspace at least `automatic_completions_after_chars` to get completions for it?

What happens if you press Backspace only one time?

ccordoba12:
> I also still think that removing white-space in front of a word should not trigger the completion.

I totally agree with this. @steff456, please also deal with that case in this PR.

spyder/plugins/editor/widgets/codeeditor.py
steff456: This is true! It should be fixed by the next commit

steff456:
@ElieGouzien can you please try it with the last modification to see if it resolves your issue?

ElieGouzien:
Unfortunately, no. What I see is as before:
Peek 14-05-2020 21-20

steff456:
Yes, can you please test it again with the latest change?

ccordoba12:
@steff456, please merge with the latest `4.x` to get the fixes to the tests that are failing here.

ElieGouzien:
@steff456: Yes, it does the job.
Thanks a lot.

ElieGouzien:
One suggestion: you could put the computation of `prev_text` inside a `key == Qt.Key_Backspace:` condition. Since backspaces are "rare events", it would improve performance, and make code clearer that `prev_text` is only useful in that case.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -4050,6 +4050,14 @@ def _handle_completions(self):
                 self._last_pressed_key = None
                 return
 
+        cursor.setPosition(pos - 1, QTextCursor.MoveAnchor)
+        cursor.select(QTextCursor.WordUnderCursor)
+        prev_text = to_text_string(cursor.selectedText())
+        cursor.setPosition(pos + 1, QTextCursor.MoveAnchor)
+
+        if prev_text == '' and key == Qt.Key_Backspace:
+            return
```
ElieGouzien, line 4059: To be more precise, here is my suggestion:
```suggestion
        if key == Qt.Key_Backspace:
            cursor.setPosition(pos - 1, QTextCursor.MoveAnchor)
            cursor.select(QTextCursor.WordUnderCursor)
            prev_text = to_text_string(cursor.selectedText())
            cursor.setPosition(pos + 1, QTextCursor.MoveAnchor)
            if prev_text == '':
                return
```

steff456:
Thanks @ElieGouzien for your suggestion!

ccordoba12:
Thanks for your help @ElieGouzien! This PR is only missing some tests that @steff456 is going to add next week.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -4072,6 +4072,14 @@ def _handle_completions(self):
                 self._last_pressed_key = None
                 return
 
+        if key == Qt.Key_Backspace:
```
ccordoba12, line 4075: @steff456, please add a comment above this line to explain why this `if` is necessary here.

spyder/plugins/editor/widgets/tests/test_completions_hide.py
```diff
@@ -54,8 +55,48 @@ def test_automatic_completions_hide_complete(lsp_codeeditor, qtbot):
...
+
     code_editor.toggle_code_snippets(True)
 
 
+@pytest.mark.slow
+@pytest.mark.first
+@flaky(max_runs=5)
+def test_automatic_completions_widget_visible(lsp_codeeditor, qtbot):
+    """Test on-the-fly completion when widget is visible and the backspace key
+    is pressed.
```
ccordoba12, line 75:
```suggestion
    """Test on-the-fly completions when the widget is visible and the
    Backspace key is pressed.
```

ElieGouzien:
Thanks @steff456 : great and fast job !

PR #12771, opened 2020-05-17
PR: Set LSP server stdout to None on Windows

### LINKED ISSUE #11506: Linter stays half-dead on Windows
[demo.mp4](https://www.dropbox.com/s/2qkp7zd97fu49a0/deadlint_demo.mp4?dl=0) No clue how to reproduce, it's a first; other `spyder` instance is running a TensorFlow session. Linter does detect `#TODO` and unused imports, but not syntax errors, and code folding fails. Kernel restart, Spyder restart, Anaconda Powershell Prompt restart didn't help - didn't try restarting computer, but think it'll work.

<hr>

**Problems:** (reproduced in 4.1.2)

 1. Syntax error red flag persists after fixing the error<sup>1</sup>
 2. Syntax errors no longer detected
 3. Unused variables no longer detected
 4. Code folding markers misplaced<sup>4</sup>
 5. Code folding folds only partially<sup>5</sup>

1: This is usually the first 'symptom'; everything goes downhill from there. The red flags can be dismissed by cut-repasting the entire code in the file.
4: 
image
5:
image

ccordoba12:
This finally fixes the "linter is dead" problem reported a lot during the last couple of months!

By changing the server's stdout to None on Windows, our client is now able to process errors correctly (e.g. errors that come from Jedi, which are the most frequent).

Before this, the client was unable to process any message coming from the server after the error. So not only the linter, but also folding and symbols totally died.

Fixes #11506.

### DIFF
--- a/spyder/plugins/completion/languageserver/client.py
+++ b/spyder/plugins/completion/languageserver/client.py
@@ -166,7 +166,11 @@ class LSPClient(QObject, LSPMethodProviderMixIn):
         self.transport_args += ['--zmq-in-port', self.zmq_out_port,
                                 '--zmq-out-port', self.zmq_in_port]
 
-        server_log = subprocess.PIPE
+        # The server stdout needs to be set to 'None' by default on Windows
+        # to correctly handle errors that come from the server.
+        # Fixes spyder-ide/spyder#11506
+        server_log = None if os.name == 'nt' else subprocess.PIPE
+
         pid = os.getpid()
         if get_debug_level() > 0:
             # Create server log file


OverLordGoldDragon:
At last, I won't be questioning the linter each time it plants a red flag.

BUG, BEGONE!

OverLordGoldDragon:
@ccordoba12 This could do for 'scripted' replacing:

```python
# if using Anaconda virtual environment, be sure to run `conda activate env_name` first
import os
import spyder

root = os.path.dirname(spyder.__file__)
path = os.path.join(root, "plugins", "completion", "languageserver", "client.py")

with open(path, 'r') as f:
    txt = f.readlines()

replace = "server_log = None if os.name == 'nt' else subprocess.PIPE"
target  = "server_log = subprocess.PIPE"

for i, line in enumerate(txt):    
    if line.lstrip(" ").rstrip("\n") == target:
        before = txt[i]
        left_spaces = len(line) - len(line.lstrip(" "))
        txt[i] = " " * left_spaces + replace + '\n'
        print("Line {} replaced in {}\nBEFORE:\n{}\nAFTER:\n{}".format(
            i, path, before, after))

with open(path, 'w') as f:
    f.write(''.join(txt))
```

```
Line 171 replaced in D:\Anaconda\envs\tf2_env\lib\site-packages\spyder\plugins\completion\languageserver\client.py
BEFORE:
        server_log = subprocess.PIPE

AFTER:
        server_log = None if os.name == 'nt' else subprocess.PIPE
```

Unsure how it could go wrong, but it's an option.

PR #12777, opened 2020-05-17
PR: Don't load files twice at startup and remove multiple calls to document_did_open

ccordoba12:
This PR does the following:

* Avoid opening files twice at startup (this problem has been present since the release of Spyder 3!).
    * That fixes code completion for languages other than Python for the first opened file of that language (linting was working fine, but code completion was not).
    * It also fixes the very strange order seen in the "Recent files" menu after Spyder is started.
* Don't call `document_did_open` twice when opening a file. That was also messing with other LSP servers.
* Perform a bit of refactoring to make clearer what several code completion methods do and unify names between codeeditor, editorstack and editor plugin.
* Correctly register that an LSP client was started the first time.

TODO:

- [x] Check if `document_did_open` is called after reloading a file that was changed outside Spyder (it is not, which is really good!).


ccordoba12:
@dalthviz, please help me to review this one.

spyder/plugins/completion/languageserver/plugin.py
```diff
@@ -401,6 +401,7 @@ def start_client(self, language):
                 logger.info("Starting LSP client for {}...".format(language))
                 language_client['instance'].start()
                 language_client['status'] = self.RUNNING
+                started = True
```
dalthviz, line 404: @ccordoba12 is this really needed? If so, maybe adding a comment could be worthy

spyder/plugins/completion/languageserver/tests/conftest.py
```diff
@@ -39,7 +39,7 @@ def __init__(self):
         self.lsp_editor_settings = {}
 
     @Slot(dict, str)
-    def register_completion_server_settings(self, settings, language):
+    def register_completion_settings(self, settings, language):
         self.lsp_editor_settings[language] = settings
```
dalthviz, line 43: Maybe since the signature name was changed we could add a docstring describing the arguments? (probably `settings` is the dict with the config that follows some specific structure, right?)

spyder/plugins/completion/languageserver/plugin.py
ccordoba12: Good eye! It's not directly related to this PR but I found it while testing completion and linting for languages other than Python.

I updated the PR description to include it and will add a couple of comments to clarify why it's necessary.

spyder/plugins/completion/languageserver/tests/conftest.py
ccordoba12: You're right. I'll add docstrings for this and the other methods I changed in my refactoring.

ccordoba12:
@dalthviz, I implemented your suggestions, so this is ready for another review.

PR #12812, opened 2020-05-20
PR: Implement workspace/didChangeWorkspaceFolders correctly and fix major issues with workspace-only servers

### LINKED ISSUE #11654: Fix request_params in send_workspace_folders_change
### Issue Report Checklist

    * [x] Restarted Spyder

## Problem Description

Ticket creation suggest by @goanpeca [here](https://github.com/palantir/python-language-server/pull/749#issuecomment-591103381).

The [python-language-server](https://github.com/palantir/python-language-server/) call for `m_workspace__did_change_workspace_folders` changed in [pyls#749](https://github.com/palantir/python-language-server/pull/749).

The parameters changed according to microsofts language-server-protocol spec for [`DidChangeWorkspaceFolders`](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_didChangeWorkspaceFolders).

In fact the `request_params` must be changed in [spyder/spyder-ide/workspace.py](https://github.com/spyder-ide/spyder/blob/47726b4d3d297b4bfbe6351905b6c44ef2b71e93/spyder/plugins/completion/languageserver/providers/workspace.py#L67) to

```
        request_params = {
            'event': {
                'added': added_folders,
                'removed': removed_folders
             }
        }
```

1.  Use python-language-server 0.32 if released
2.  Try `send_workspace_folders_change`

```python-traceback

PASTE TRACEBACK HERE

```

## Versions

* Spyder version: 
* Python version: 
* Qt version: 
* PyQt version: 

### Dependencies

```
PASTE DEPENDENCIES HERE
```

andfoy:
## Description of Changes
This PR does the following:

* Fix the implementation of `workspace/didChangeWorkspaceFolders`, which was incorrect according to the LSP spec.
* Implement the response to the call `client/registerCapability`, essential for some workspace-oriented servers.
* Reduce the total number of calls to `workspace/didChangeConfiguration` for non-Python servers.
* Enable stdio server logs.
* Refactor the way server and transport log filenames are generated. We were using almost the same code for both, so we moved that to a single method.
* Refactor how the arguments for the server and transport processes are created. Instead of defining them in several methods, now they are unified in a single one per process.
* Pass all environment variables to the server and transport processes for LSPs other than Python.
* Fix `id` of `server_ready` request.

### Issue(s) Resolved

Fixes #11654

TODO

- [x] Check that files are closed and opened correctly when switching projects.

### Affirmation


spyder/plugins/completion/languageserver/tests/test_client.py
```diff
@@ -217,3 +218,30 @@ def test(a, b):
     # Assert the response has what we expect
     definition = response['params']
     assert 'Test docstring' in definition
+
+
+@pytest.mark.slow
+@pytest.mark.third
+def test_didChangeWorkspaceFolders(lsp_client_and_completion, qtbot):
```
andfoy, line 225: I think this test help us to test our side of the call, rather than the pyls one

ccordoba12: Yeah, that's right, it only tests things in our side because the PyLS doesn't return a response after sending `didChangeWorkspaceFolders` (as you told me).

However, I found an error in our side (we were not deleting workspaces after switching projects), so that's why I decided to add it. Do you want me to move it somewhere else to make things clearer?

andfoy: I think renaming the test would help to clarify that the test is over `send_workspace_folders_change`, rather than on the LSP call

ccordoba12: Ok, will do.

PR #12834, opened 2020-05-22
PR: Improve handling of IPython Console preferences

### LINKED ISSUE #11357: IPython Console: Improvements for preferences application
* [x] Add validations for custom consoles dependencies (SymPy, Matplotlib, Cython)
* [x] Improve preferences handling for the IPython Console (for example, if the preference requires a restart of the console, show a message).

As discussed with @ccordoba12

dalthviz:
## Description of Changes

When at least an option needs a restart:
prefs

All the options that don't need a restart are applied always

### TODO

- [ ] Support `Completion type` hot change (needs change over `QtConsole`)
- [x] Check preferences application when a console is running something (currently execution is interrupted, for example when debugging)

### Issue(s) Resolved

Depends on spyder-ide/spyder-kernels#223
Depends on jupyter/qtconsole#451
Fixes #11357 

### Affirmation


spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -170,35 +170,181 @@ def update_font(self):
...
-                client.set_elapsed_time_visible(show_time_o)
-            if reset_namespace_n in options:
-                client.reset_warning = reset_namespace_o
-            if ask_before_restart_n in options:
-                client.ask_before_restart = ask_before_restart_o
+        ask_before_closing_n = 'ask_before_closing'
+        use_pager_n = 'use_pager'
+        show_calltips_n = 'show_calltips'
+        buffer_size_n = 'buffer_size'
+        # Advanced GUI options
```
ccordoba12, line 186: Please add a blank line above this one.

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -170,35 +170,181 @@ def update_font(self):
...
+        in_prompt_n = 'in_prompt'
+        out_prompt_n = 'out_prompt'
+        # Control widget
```
ccordoba12, line 189: Add a blank line above this one.

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -170,35 +170,181 @@ def update_font(self):
...
+        if show_calltips_n in options:
+            show_calltips_o = self.get_option(show_calltips_n)
+            client.set_show_calltips(show_calltips_o)
+        if use_pager_n in options:
+            use_pager_o = self.get_option(use_pager_n)
+            client.set_use_pager(use_pager_o)
+        if buffer_size_n in options:
+            buffer_size_o = self.get_option(buffer_size_n)
+            client.set_buffer_size(buffer_size_o)
+        # Advanced GUI options
```
ccordoba12, line 222: Add a blank line above this one.

spyder/plugins/ipythonconsole/widgets/client.py
```diff
@@ -504,6 +506,65 @@ def set_color_scheme(self, color_scheme, reset=True):
         except AttributeError:
             pass
 
+    def set_use_pager(self, use_pager):
```
ccordoba12, line 509: Let's rename this to `set_paging`.

ccordoba12: Also, since all these methods apply at the `shellwidget` level, I think we should remove them here, leave them only in `shellwidget` and call them like this

    sw = client.shellwidget
    sw.set_paging

(and the same for all other methods) from the plugin.

ccordoba12:
This looks pretty good @dalthviz and I don't have more comments about it. The only thing that worries me a bit is the dialog we'll show to our users to inform them about a change in Preferences, because it's very long and hard to read.

I think we should talk to Isabella in our next meeting about that. What do you say?

dalthviz:
Sure :+1:

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -63,6 +65,96 @@
...
+            self.RESTART_ALL: _("Restart All Kernels")
+            }
+        # Dialog widgets
+        # Text
+        self._text_label = QLabel(
+            _("By default, some IPython console preferences will be "
+              "applied to new consoles only. To apply preferences to "
+              "existing consoles, select from the options below.<br><br>"
+              "Please note: applying changes to running consoles will force"
+              " a kernel restart and all current work will be lost.<br>"),
```
ccordoba12, line 95:
```suggestion
              " a kernel restart and all current work will be lost."),
```

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -63,6 +65,96 @@
...
+            self)
+        self._text_label.setWordWrap(True)
```
ccordoba12, line 97:
```suggestion
        self._text_label.setWordWrap(True)
        self._text_label.setFixedWidth(350)
```

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -63,6 +65,96 @@
...
+            self._action_string[self.NO_RESTART], parent=self)
+
+        # Dialog Layout
+        layout = QVBoxLayout()
+        layout.addWidget(self._text_label)
+        layout.addWidget(self._restart_current)
+        layout.addWidget(self._restart_all)
+        layout.addWidget(self._action_button, 0, Qt.AlignRight)
+        self.setLayout(layout)
+        self.setFixedWidth(600)
```
ccordoba12, line 121:
```suggestion
        layout = QVBoxLayout(self)
        layout.addWidget(self._text_label)
        layout.addSpacing(5)
        layout.addWidget(self._restart_current)
        layout.addWidget(self._restart_all)
        layout.addSpacing(10)
        layout.addWidget(self._action_button, 0, Qt.AlignRight)
        layout.setContentsMargins(20, 20, 20, 20)
        self.setLayout(layout)
```

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -311,6 +311,74 @@ def request_env(self):
...
+    def get_matplotlib_backend(self):
+        """Call kernel to get current backend."""
+        return self.call_kernel(
+            interrupt=True,
+            blocking=True).get_matplotlib_backend()
+
+    def set_matplotlib_backend(self, backend_option, pylab=False):
+        """Set matplotlib backend given a backend name."""
+        self.call_kernel(
+            interrupt=True).set_matplotlib_backend(
```
impact27, line 347: I don't think that will work: When you call the kernel with `interrupt=True`, it will run the code in a thread. This function need to be run in the main thread. In https://github.com/spyder-ide/spyder/pull/14199, I solved it by:
```
backend = "inline" if self._gui != "inline" else "auto"
self.current_shellwidget().execute("%matplotlib " + backend)
```
Execute will be executed whenever a prompt becomes available.

impact27: Also avilable with hidden=True

impact27:
It looks quite good! A word of advice with comms: using `call_kernel(interrupt=True)` executes the code immediately and in a thread. You must be careful about thread safety. Matplotlib for example is not thread safe. I marked onr spot in the code where it might be problematic. You have two options if the code is not thread safe:
 - Call it with `(interrupt=False)` (the default). It will be executed when coming back to the main prompt (So it won't work while debugging)
 - use `shellwidget.execute(..., hidden=True)`. This will execute the code at the next prompt, including pdb prompt.

dalthviz:
Thanks for the feedback @impact27 ! :+1:

ccordoba12:
@dalthviz, I think this one is ready, except for a couple of improvements I'm working on. Please don't push more commits to this PR.

ccordoba12:
Given that users can be working in a long debugging session, I added a message about preferences not being applied in that case:

Selección_051

This is a limitation that we can address in a future PR (to not make this one longer). Perhaps @impact27 could help us with that?

ccordoba12 (approved):
Thanks a lot @dalthviz for your hard work and your patience. This is a terrific addition!

spyder/plugins/ipythonconsole/widgets/shell.py
ccordoba12: Sorry, I missed this one @impact27. I'll follow your suggestion in another PR.

PR #12835, opened 2020-05-23
PR: Test that a file with the same name of a standard library module doesn't break the console

ccordoba12:
This is due to a change in Python 3.7, which added the current working directory to `sys.path` when running modules with `python -m`. And that's is precisely the way we start our kernels.

Depends on spyder-ide/spyder-kernels#222


goanpeca:
Just to double check, so the test you just wrote fails without the updates on spyder-kernel :-p ?

ccordoba12:
Yes, I checked that without spyder-ide/spyder-kernels#222 this test fails, i.e. the kernel crashes because it uses the faulty `random.py` instead of the real, std library one.

Issue #12637, opened 2020-05-06
The right scroll bar is missing in Spyder 4 on MacOS 10.15


aosavas:
### Issue Report Checklist

* [x] Searched the [issues page](https://github.com/spyder-ide/spyder/issues?q=is%3Aissue) for similar reports
* [x] Read the relevant sections of the [Spyder Troubleshooting Guide](https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ) and followed its advice
* [x] Reproduced the issue after updating with ``conda update spyder`` (or ``pip``, if not using Anaconda)
* [ ] Could not reproduce inside ``jupyter qtconsole`` (if console-related)
* [x] Tried basic troubleshooting (if a bug/error)
    * [x] Restarted Spyder
    * [x] Reset preferences with ``spyder --reset``
    * [x] Reinstalled the latest version of [Anaconda](https://www.anaconda.com/download/)
    * [x] Tried the other applicable steps from the Troubleshooting Guide
* [x] Completed the **Problem Description**, **Steps to Reproduce** and **Version** sections below

## Problem Description
The scroll bar on the right hand side that shows overview of issues, highlighted items, etc... is missing in Spyder 4.1.2 and 4.x.x releases on MacOS 10.15.4

### What steps reproduce the problem?

1. I have tried to use "spyder --reset" no luck 
2. I have downgraded to 3.3.6 and it works

![image](https://user-images.githubusercontent.com/46607096/81224565-ea3de980-8fb5-11ea-8d82-d3353366d577.png)

3. I have upgraded to 4.0.0, missing.

![image](https://user-images.githubusercontent.com/46607096/81224763-3852ed00-8fb6-11ea-873e-a65587fbc2de.png)

### What is the expected output? What do you see instead?
I am not seeing the scroll bar that I am used to on 3.x versions. Talked to @bcolsen and he suggested that I should open a bug at github.

### Paste Traceback/Error Below (if applicable)

```python-traceback

PASTE TRACEBACK HERE

```

## Versions

* Spyder version: 
spyder                    4.1.2                    py37_0  
spyder-kernels            1.9.0                    py37_0  
* Python version: 
python                    3.7.3                h359304d_0  
python-chromedriver-binary 83.0.4103.14.0   py37hc8dfbb8_0    conda-forge
python-dateutil           2.8.1                      py_0  
python-jsonrpc-server     0.3.4                      py_0  
python-language-server    0.31.10                  py37_0  
[...]

PR #12852, opened 2020-05-25
PR: Fix hiding completion on backspace when nothing before cursor

ElieGouzien:
Dear spyder devs,

Using the fix brought by #12710, I realized it was not working as expected when not at the beginning of a line (or when end of previous line is a word).

Here is a proposition to fix those issues with #12637. 
Main idea is to also check for previous character, since "previous word" from qt behaves strangely.

### Affirmation

@ElieGouzien

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -4078,8 +4078,10 @@ class CodeEditor(TextEditBaseWidget):
             cursor.setPosition(pos - 1, QTextCursor.MoveAnchor)
             cursor.select(QTextCursor.WordUnderCursor)
             prev_text = to_text_string(cursor.selectedText())
-            cursor.setPosition(pos + 1, QTextCursor.MoveAnchor)
-            if prev_text == '':
+            cursor.setPosition(pos - 1, QTextCursor.MoveAnchor)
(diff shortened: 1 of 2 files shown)


spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -4078,8 +4078,10 @@ def _handle_completions(self):
             cursor.setPosition(pos - 1, QTextCursor.MoveAnchor)
             cursor.select(QTextCursor.WordUnderCursor)
             prev_text = to_text_string(cursor.selectedText())
-            cursor.setPosition(pos + 1, QTextCursor.MoveAnchor)
-            if prev_text == '':
+            cursor.setPosition(pos - 1, QTextCursor.MoveAnchor)
+            cursor.setPosition(pos, QTextCursor.KeepAnchor)
+            prev_char = cursor.selectedText()
+            if prev_text == '' or prev_char in ('\u2029', ' ', '\t'):
```
ccordoba12, line 4084: I think you need to use `u'\u2029'` here, for this to work in Python 2. We still support Python 2 an since our tests are failing only in that version, this could be the problem.

ElieGouzien: Thank for the trick. I didn't even think to write something compatible with python2.

PR #12912, opened 2020-06-01
PR: Use QProcess instead of subprocess for the LPS transport layer and server

### LINKED ISSUE #12225: LSP cycling : restarting ... ready ... restarting ...
### Issue Report Checklist

* [x ] Searched the [issues page](https://github.com/spyder-ide/spyder/issues?q=is%3Aissue) for similar reports
    * [x] Restarted Spyder

## Problem Description
I use latest version (last commit e3326bd531d16ff3551855a56882f8df129f5dfc) with conda and for some reason the LSP is not working and I have a notification on the bottom bar LSP python restarting and then ready and it cycles for ever

1. Just launch spyder via python bootstrap.py

LSP working normally. Linting and completion and everything related to LSP is not working but everything else seems OK !!

see the spyder-debug.log

[spyder-debug.log](https://github.com/spyder-ide/spyder/files/4450984/spyder-debug.log)

## Versions

* Spyder version: 5.0.0.dev0 - Revision e3326bd53
* Python version: Python 3.7.6 64bits
* Qt version: 5.12.5
* PyQt version: 5.12.3 on [Linux]
* Operating System name/version: Archlinux with conda

### Dependencies

[conda_list.log](https://github.com/spyder-ide/spyder/files/4451002/conda_list.log)

ccordoba12:
* This will allow us to have better control over those external processes. Specifically, Qt emits signals when they fail to start and when they crash, so we only need to connect to those signals and run our restart mechanism.
* In particular, this fixes the infinite restarting-ready-starting sequence we have right now when the PyLS fails to start. That happens because `subprocess` doesn't have an option to check if the process started correctly (calling `poll` works if we wait a little bit for the process to start, but that little bit depends on the OS and user hardware).
* Finally, this has some refactoring of the `LSPClient` code to make it more readable and easier to follow.

Fixes #12225.

TODO:

- [x] Connect to QProcess errorOccurred and finished signals.
- [x] Simplify hearbeat and restarting logic.
- [x] Check if `new_env` needs to have the entire environment or just PYTHONPATH.

### DIFF
--- a/spyder/plugins/completion/languageserver/client.py
+++ b/spyder/plugins/completion/languageserver/client.py
@@ -83,7 +83,7 @@ class LSPClient(QObject, LSPMethodProviderMixIn):
         self.zmq_out_socket = None
         self.zmq_in_port = None
         self.zmq_out_port = None
-        self.transport_client = None
+        self.transport = None
         self.lsp_server = None
         self.stdio_pid = None
         self.notifier = None
@@ -218,7 +218,7 @@ class LSPClient(QObject, LSPMethodProviderMixIn):
         self.transport_args = list(map(str, self.transport_args))
         logger.info('Starting transport: {0}'
                     .format(' '.join(self.transport_args)))
-        self.transport_client = QProcess(self)
+        self.transport = QProcess(self)
 
         # This allows running LSP tests directly without having to install
         # Spyder
@@ -234,34 +234,31 @@ class LSPClient(QObject, LSPMethodProviderMixIn):
         #if PY2:
         #    new_env = clean_env(new_env)
 
-        self.transport_client.setProcessEnvironment(new_env)
+        self.transport.setProcessEnvironment(new_env)
 
         # Set transport logging file
-        client_log_file = None
+        transport_log_file = None
         if get_debug_level() > 0:
-            # Client log file
-            client_log_fname = 'client_{0}_{1}.log'.format(self.language, pid)
-            client_log_file = get_conf_path(osp.join('lsp_logs',
-                                                     client_log_fname))
-            if not osp.exists(osp.dirname(client_log_file)):
-                os.makedirs(osp.dirname(client_log_file))
+            transport_log_fname = 'transport_{0}_{1}.log'.format(
+                self.language, pid)
+            transport_log_file = get_conf_path(
+                osp.join('lsp_logs', transport_log_fname))
+            if not osp.exists(osp.dirname(transport_log_file)):
+                os.makedirs(osp.dirname(transport_log_file))
 
         # Set channel properties
         if self.stdio:
             self.transport_args += self.server_args
-            self.transport_client.setProcessChannelMode(
-                QProcess.SeparateChannels)
-            if client_log_file is not None:
-                self.transport_client.setStandardErrorFile(client_log_file)
+            self.transport.setProcessChannelMode(QProcess.SeparateChannels)
+            if transport_log_file is not None:
+                self.transport.setStandardErrorFile(transport_log_file)
         else:
-            self.transport_client.setProcessChannelMode(
-                QProcess.MergedChannels)
-            if client_log_file is not None:
-                self.transport_client.setStandardOutputFile(client_log_file)
+            self.transport.setProcessChannelMode(QProcess.MergedChannels)
+            if transport_log_file is not None:
+                self.transport.setStandardOutputFile(transport_log_file)
 
         # Start transport
-        self.transport_client.start(self.transport_args[0],
-                                    self.transport_args[1:])
+        self.transport.start(self.transport_args[0], self.transport_args[1:])
 
         # Create notifier
         fid = self.zmq_in_socket.getsockopt(zmq.FD)
@@ -272,14 +269,14 @@ class LSPClient(QObject, LSPMethodProviderMixIn):
         logger.debug('LSP {} client started!'.format(self.language))
 
     def stop(self):
-        """Stop transport client and server."""
+        """Stop transport and server."""
         logger.info('Stopping {} client...'.format(self.language))
         if self.notifier is not None:
             self.notifier.activated.disconnect(self.on_msg_received)
             self.notifier.setEnabled(False)
             self.notifier = None
-        if self.transport_client is not None:
-            self.transport_client.kill()
+        if self.transport is not None:
+            self.transport.kill()
         self.context.destroy()
         if self.lsp_server is not None:
             self.lsp_server.kill()
@@ -287,8 +284,8 @@ class LSPClient(QObject, LSPMethodProviderMixIn):
     def is_transport_alive(self):
         """Detect if transport layer is alive."""
         alive = True
-        if self.transport_client is not None:
-            if self.transport_client.state() == QProcess.NotRunning:
+        if self.transport is not None:
+            if self.transport.state() == QProcess.NotRunning:
                 alive = False
 
         return alive
@@ -461,8 +458,7 @@ class LSPClient(QObject, LSPMethodProviderMixIn):
     @send_request(method=LSPRequestTypes.INITIALIZE)
     def initialize(self, params, *args, **kwargs):
         self.stdio_pid = params['pid']
-        pid = (self.transport_client.pid()
-               if not self.external_server else None)
+        pid = self.transport.pid() if not self.external_server else None
         params = {
             'processId': pid,
             'rootUri': pathlib.Path(osp.abspath(self.folder)).as_uri(),
--- a/spyder/plugins/completion/languageserver/client.py
+++ b/spyder/plugins/completion/languageserver/client.py
@@ -166,6 +166,11 @@ class LSPClient(QObject, LSPMethodProviderMixIn):
         self.transport_args += ['--zmq-in-port', self.zmq_out_port,
                                 '--zmq-out-port', self.zmq_in_port]
 
+    @Slot(QProcess.ProcessError)
+    def handle_process_errors(self, error):
+        """Handle errors with the transport layer or server processes."""
+        self.sig_went_down.emit(self.language)
+
     def start_server(self):
         """Start server."""
         # This is not necessary if we're trying to connect to an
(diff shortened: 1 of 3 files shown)


goanpeca:
Really good idea @ccordoba12 !

spyder/plugins/completion/languageserver/client.py
```diff
@@ -71,7 +71,7 @@ class LSPClient(QObject, LSPMethodProviderMixIn):
 
     #: Signal to warn the user when either the transport layer or the
     #  server are down
-    sig_lsp_down = Signal(str)
+    sig_down = Signal(str)
```
goanpeca, line 74: Hmm this is a weird name, what do you want to convey @ccordoba12 ?

goanpeca: `sig_stopped` 
`sig_errored`

ccordoba12: Yeah, I'm not happy with it either. It's used to inform the LSP plugin that either the transport layer or the server are down. Could you come up with a better name for it?

ccordoba12:
> Really good idea @ccordoba12 !

Thanks! It took several hours to figure it that this was the best move though.

spyder/plugins/completion/languageserver/client.py
ccordoba12: What about `sig_is_down`? Although `sig_errored` is good for me too.

ccordoba12: Another option could be `sig_error_occurred` (as the Qt signal).

jitseniesen:
The notebook plugin also uses `subprocess.Popen()` to start new servers. Would you recommend to use QProcess there too? I think I need to touch that code anyway.

spyder/plugins/completion/languageserver/client.py
ccordoba12: I decided to go for `sig_went_down`.

ccordoba12:
@goanpeca, this is ready fo review.

@jitseniesen, yes, I think you should use QProcess because it allows you to better monitor if server started correctly in the background.

spyder/plugins/completion/languageserver/client.py
goanpeca: hehe ok, that one works!

goanpeca:
Will check tonight!

spyder/plugins/completion/languageserver/client.py
```diff
@@ -193,93 +202,82 @@ def start(self):
...
-                stderr=server_stderr,
-                creationflags=creation_flags,
-                cwd=cwd)
+        # Setup server
+        self.server = QProcess(self)
+        self.server.errorOccurred.connect(self.handle_process_errors)
+        self.server.setWorkingDirectory(cwd)
+        self.server.setProcessChannelMode(QProcess.MergedChannels)
+        if server_log_file is not None:
+            self.server.setStandardOutputFile(server_log_file)
```
andfoy, line 224: I think most servers write on stderr

ccordoba12: I set the channels mode a couple of lines above to `QProcess.MergedChannels`, which means stderr and stdout are merged into a single channel, as described [here](https://doc.qt.io/qt-5/qprocess.html#ProcessChannelMode-enum).

spyder/plugins/completion/languageserver/client.py
```diff
@@ -70,8 +69,8 @@ class LSPClient(QObject, LSPMethodProviderMixIn):
     sig_server_error = Signal(str)
 
     #: Signal to warn the user when either the transport layer or the
-    #  server are down
-    sig_lsp_down = Signal(str)
+    #  server went down
+    sig_went_down = Signal(str)
```
goanpeca, line 73: Not related to this PR, but we might consider in the future documenting signals like
```python
sig_went_down = Signal(str)
"""Documentation of the signal.""""
```

This is what sphinx will parse

spyder/plugins/completion/languageserver/client.py
```diff
@@ -288,22 +286,23 @@ def patched_cleanup():
...
+            self.server.kill()
 
     def is_transport_alive(self):
         """Detect if transport layer is alive."""
         alive = True
-        if self.transport_client is not None:
-            if self.transport_client.poll() is not None:
+        if self.transport is not None:
+            if self.transport.state() == QProcess.NotRunning:
                 alive = False
```
goanpeca, line 306: How about:

```python
        alive = False
        if self.transport_client is not None:
            alive = self.transport.state() != QProcess.NotRunning:
```

Why is the default value True for alive. If there is no transport, it will return True which feels weird.

spyder/plugins/completion/languageserver/client.py
```diff
@@ -327,35 +326,45 @@ def is_stdio_alive(self):
     def is_tcp_alive(self):
         """Detect if a tcp server is alive."""
         alive = True
-        if self.lsp_server is not None:
-            if self.lsp_server.poll() is not None:
+        if self.server is not None:
+            if self.server.state() == QProcess.NotRunning:
                 alive = False
```
goanpeca, line 331: How about:

```python
        alive = False
        if self.server is not None:
            alive = self.server.state() != QProcess.NotRunning:
```

Why is the default value True for alive. If there is no transport, it will return True which feels weird.

spyder/plugins/completion/languageserver/client.py
```diff
@@ -393,7 +402,7 @@ def send(self, method, params, kind):
                 return int(_id)
             except zmq.error.Again:
                 if time.time() > timeout_time:
-                    self.sig_lsp_down.emit(self.language)
+                    self.sig_went_down.emit(self.language)
```
goanpeca, line 405: Should we actually kill the process at this stage?

spyder/plugins/completion/languageserver/client.py
ccordoba12: I refactored the code a little bit to improve the readability of this method.

spyder/plugins/completion/languageserver/client.py
ccordoba12: I also refactored this.

spyder/plugins/completion/languageserver/client.py
ccordoba12: I also refactored this.

spyder/plugins/completion/languageserver/client.py
ccordoba12: That's handled by the restart mechanism.

spyder/plugins/completion/languageserver/plugin.py
```diff
@@ -107,28 +107,24 @@ def restart_lsp(self, language, force=False):
                 try:
                     self.clients_hearbeat[language].stop()
                     client['instance'].disconnect()
+                    client['instance'].stop()
```
ccordoba12, line 110: @goanpeca, here we kill the LSP client if it definitely fails to run.

ccordoba12:
@goanpeca, @andfoy, this is ready for a final review.

spyder/plugins/completion/languageserver/client.py
```diff
@@ -288,74 +289,81 @@ def patched_cleanup():
...
-        if self.transport_client is not None:
-            if self.transport_client.poll() is not None:
-                alive = False
-
-        return alive
+        state = self.transport.state()
+        if state == QProcess.Starting:
+            return True
+        else:
+            return state != QProcess.NotRunning
```
goanpeca, line 310: Why not make this just

`return  state != QProcess.NotRunning`

If it is Starting or Running it will be True, if it is Not Running it will be False

spyder/plugins/completion/languageserver/client.py
```diff
@@ -288,74 +289,81 @@ def patched_cleanup():
...
-
-    def send(self, method, params, kind):
-        # Detect when the transport layer is down to show a message
-        # to our users about it.
-        if not self.is_transport_alive():
+        state = self.server.state()
+        if state == QProcess.Starting:
+            return True
+        else:
+            return state != QProcess.NotRunning
```
goanpeca, line 332: Why not make this just

`return  state != QProcess.NotRunning`

If it is Starting or Running it will be True, if it is Not Running it will be False

spyder/plugins/completion/languageserver/plugin.py
goanpeca: Ok

spyder/plugins/completion/languageserver/client.py
ccordoba12: Yoiu're right. Done.

spyder/plugins/completion/languageserver/client.py
ccordoba12: Done.

ccordoba12:
@OverLordGoldDragon, could you help us to test Spyder with these changes? The thing is this PR supersedes PR #12771, but it's a better solution to start and monitor the Python language server than before.

I tested that the problem on issue #11506 doesn't appear on Windows with these changes. But since you're a heavy Spyder user, we'd really appreciate your help to be really sure things keep working as they were after #12771.

OverLordGoldDragon:
@ccordoba12 Sure, I'm down. Though, that's a fair bit of changes to insert manually - is there a quicker approach? Unsure if cloning would work, or whether Spyder has a nightly.

ccordoba12:
Right. In that case, I could create a conda package for you. Just let me know your Python version and I'll upload it here or in Dropbox.

OverLordGoldDragon:
@ccordoba12 Sounds good; Python 3.7.7, IPython 7.13.0, conda 4.8.3. Btw, I'm using two separate envs at once, one of them has Python 3.7.5; if it makes a difference, I could limit the new install to one of them.

OverLordGoldDragon:
@ccordoba12 Apparently I forgot about the "View file" feature in "Files changed"; would the test work as intended if just copy-pasting these 3 files over? I also figure I should undo the other PR (12771) to test if this one works instead.

PR #12937, opened 2020-06-02
PR: Enable/disable folding guides when cursor enters/exits the left panel

### LINKED ISSUE #12315: Restore code folding for large files
In order to do that we need to do the following:

* Only draw and update the folding panel when the user moves the cursor over it.
* Only draw the visible region in the editor (although that's probably not so necessary if we draw the panel on demand).

@andfoy, please take a look at this one.

andfoy:
## Description of Changes
This PR disables folding guides visibility by default, instead, they will be shown only if the cursor enters the left panel and disabled when the cursor leaves it. We also re-enable the folding support for large files that was disabled previously. 

Peek 02-06-2020 14-39

### Issue(s) Resolved

Fixes #12315.

### Affirmation

### DIFF
--- a/spyder/plugins/editor/panels/codefolding.py
+++ b/spyder/plugins/editor/panels/codefolding.py
@@ -249,9 +249,27 @@ class FoldingPanel(Panel):
         # Paints the fold indicators and the possible fold region background
         # on the folding panel.
         super(FoldingPanel, self).paintEvent(event)
-        if not self._display_folding and not self._key_pressed:
-            return
         painter = QPainter(self)
+        if not self._display_folding and not self._key_pressed:
+            if any(self.folding_status.values()):
+                for top_position, line_number, block in self.editor.visible_blocks:
+                    if line_number in self.folding_regions:
+                        collapsed = self.folding_status[line_number]
+                        line_end = self.folding_regions[line_number]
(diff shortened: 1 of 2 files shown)


goanpeca:
Nice work @andfoy !

I suggest we always display the folded indicators, (much like VSCode doeS)

image

andfoy:
@goanpeca Here it is:

Peek 02-06-2020 15-21

goanpeca:
Awesome! 🥳

spyder/plugins/editor/panels/codefolding.py
```diff
@@ -248,6 +250,28 @@ def paintEvent(self, event):
...
+                            # it might have been folded by the parent
+                            # editor/document in the case of cloned editor
+                            for deco in self._block_decos:
+                                if deco.block == block:
+                                    # no need to add a deco, just go to the
+                                    # next block
+                                    break
+                            else:
+                                self._add_fold_decoration(block, line_end)
+            return
```
ccordoba12, line 274: The code you added in these lines is very, very similar to the code to draw fold triggers from like 10 lines below. Do you see a way to refactor both and have a single method here and there?

andfoy:
@ccordoba12, this is ready for a final review

PR #12972, opened 2020-06-07
PR: Fix several problems about kernel restarts

### LINKED ISSUE #13018: Kernel is not restarted correctly when RecursiveError is thrown on Windows
```python
class C():
    @property
    def x(self):
        return self.x
C().x
```
See [here](https://stackoverflow.com/q/62356253/10133797); same if running as a file (main) or from within IPython.

Restarting kernel later rather than sooner, it's non-responsive, so I closed the terminal - yielding this:

image

Closing again worked properly; opened Console 3/A.

ccordoba12:
This PR fixes the following problems:

* If the user restarted the kernel manually once and then it crashed, the console froze while trying to restart it.
* The loading page was not hidden if the kernel crashed several times while running.
* The variable explorer was not cleared after a kernel restart.
* The stop button was not disabled after a crash.
* ~~We were not giving focus to the console after a kernel restart~~ (I reverted this because it was changing the focus logic established before).
* Override `request_restart_kernel` from Qtconsole to use our own restart mechanism when users press `Ctrl+.`. This avoids a total freeze when doing that.

TODO:

- [x] Add a test to account for a manual kernel restart and then a crash.
- [x] Add a test for clearing the Variable Explorer after a kernel restart.

Fixes #13018.


ccordoba12:
@impact27, I think this one is ready. Could you help me to review it?

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -1228,7 +1228,7 @@ def create_kernel_manager_and_kernel_client(self, connection_file,
 
         # Increase time to detect if a kernel is alive.
         # See spyder-ide/spyder#3444.
-        kernel_client.hb_channel.time_to_dead = 45.0
+        kernel_client.hb_channel.time_to_dead = 25.0
```
goanpeca, line 1231: Please add the units of this value.

spyder/plugins/ipythonconsole/widgets/client.py
```diff
@@ -548,10 +556,20 @@ def restart_kernel(self):
             if sw.kernel_manager:
                 if self.infowidget.isVisible():
                     self.infowidget.hide()
+
+                if self._abort_kernel_restart():
+                    return
```
impact27, line 561: is the comm closed here?

spyder/plugins/ipythonconsole/plugin.py
ccordoba12: Fixed in the comment above.

spyder/plugins/ipythonconsole/widgets/client.py
ccordoba12: Fixed.

spyder/app/tests/test_mainwindow.py
```diff
@@ -3221,5 +3221,33 @@ def test_run_unsaved_file_multiprocessing(main_window, qtbot):
             lambda: 'None' in shell._control.toPlainText())
 
 
+@pytest.mark.slow
+@flaky(max_runs=3)
```
goanpeca, line 3225: Why is the flaky needed?

Did it fail intermittently when testing this PR?

spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py
```diff
@@ -1659,5 +1659,45 @@ def test_wrong_std_module(ipyconsole, qtbot):
     os.remove(wrong_random_mod)
 
 
+@flaky(max_runs=3)
```
goanpeca, line 1662: Same question

spyder/app/tests/test_mainwindow.py
ccordoba12: Sometimes the kernel fails to start or restart during the prescribed timeout, so we decided to mark these tests as flaky to avoid a full run of our test suite.

spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py
ccordoba12: See my explanation above. The same one applies here.

PR #13032, opened 2020-06-15
PR: Show the context menu with disabled options in the file explorer when it is empty

### LINKED ISSUE #13004: Right-clicking an empty folder doesn't show the context menu of the Files pane
Entering inside a folder without elements and then right clicking inside it doesn't show the context menu of the Files pane:
explorer1

_Originally posted by @dalthviz in https://github.com/spyder-ide/spyder/pull/12401#issuecomment-642978296_

steff456:
## Description of Changes

Add an empty element to the filenames in case there's not any filename selected in the files explorer. In this way, the context menu will be visible in Windows.

prueba

### Issue(s) Resolved

Fixes #13004 

### Affirmation

### DIFF
--- a/spyder/plugins/explorer/widgets/explorer.py
+++ b/spyder/plugins/explorer/widgets/explorer.py
@@ -287,8 +287,8 @@ class DirView(QTreeView):
         if self.selectionMode() == self.ExtendedSelection:
             if self.selectionModel() is None:
                 return []
-            return [self.get_filename(idx) for idx in
-                    self.selectionModel().selectedRows()]
+            rows = self.selectionModel().selectedRows()
+            return [self.get_filename(idx) for idx in rows]
         else:
             return [self.get_filename(self.currentIndex())]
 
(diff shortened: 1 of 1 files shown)


steff456:
@ccordoba12 checking the errors that @dalthviz found, they are also happening on 4.x branch in macOS and Linux.

It happens because there's not any filenames to `move`. Also the absolute and relative path is given from the filenames in the directory. So we wanted to ask you if we will prefer to have only the create option menu available,

image

Or to put the complete context menu with those options not activated?

ccordoba12:
> It happens because there's not any filenames to move. Also the absolute and relative path is given from the filenames in the directory. So we wanted to ask you if we will prefer to have only the create option menu available,

That makes a lot of sense. Please implement that.

steff456:
Now the context menu is disabled and the user can only create new things in all OS

image

dalthviz:
Seems like any action of the ones enabled works @steff456 :/ :

explorer

Maybe all of the actions need at least one element to work inside the directory?

PR #13047, opened 2020-06-17
PR: Clear old set shortcuts if they are now empty (Shortcuts)

### LINKED ISSUE #12992: Shift+return not working for run selection -  no solutions work
## Problem Description
Updated to most recent Spyder version (4.1.3). Wanted to change the "run selection" back to what I was used to in previous version, shift+return. So: preferences > "run selection" > pressed Shift+Return in new shortcut. Apply, Ok. 
Tried to run a selection with Shift+Return, nothing happens. The "run selection" button does work normally. Also tried to run it with the "enter" button on my numpad, doesn't work either. Restarted Spyder, used --reset, no changes. NOTE: if I change the "run selection" to something else (e.g. Shift+Backspace) it does work as expected.

1. Update Spyder to 4.1.3
2. Set "run selection" to Shift+Return
3. Try to run selection with Shift+Return

Expected: to run the selection
Instead: nothing happens

## Versions
* Spyder version: 4.1.3
* Python version: 3.7
* Operating System name/version: windows

Thanks for this otherwise wonderful software!

dalthviz:
## Description of Changes

If a shortcut conflict is detected, and the user decides to do an overwrite of a shortcut, the old action needs to unset the shortcut to prevent having two actions with the same shortcut.

### Issue(s) Resolved

Fixes #12992 

### Affirmation

### DIFF
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -2453,12 +2453,12 @@ def test_preferences_empty_shortcut_regression(main_window, qtbot):
     # Create new file
     main_window.editor.new()
     code_editor = main_window.editor.get_focus_widget()
-    code_editor.set_text('print(0)\nprint(1)\nprint(2)')
+    code_editor.set_text(u'print(0)\nprint(ññ)')
 
     with qtbot.waitSignal(shell.executed):
         qtbot.keyClick(code_editor, Qt.Key_Return, modifier=Qt.ShiftModifier)
-    qtbot.waitUntil(lambda: 'print(0)' in shell._control.toPlainText())
-    assert 'print(1)' not in shell._control.toPlainText()
+    qtbot.waitUntil(lambda: u'print(0)' in shell._control.toPlainText())
+    assert u'ññ' not in shell._control.toPlainText()
 
     # Reset shortcuts
     CONF.set_shortcut(
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -2422,6 +2422,57 @@ def test_preferences_change_font_regression(main_window, qtbot):
 
 
 @pytest.mark.slow
+def test_preferences_empty_shortcut_regression(main_window, qtbot):
+    """
+    Test for spyder-ide/spyder/#12992 regression.
+
+    Overwritting shortcuts results in a shortcuts conflict.
+    """
+    # Wait until the window is fully up
+    shell = main_window.ipyconsole.get_current_shellwidget()
+    qtbot.waitUntil(lambda: shell._prompt_html is not None,
+                    timeout=SHELL_TIMEOUT)
+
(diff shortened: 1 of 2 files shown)


spyder/config/main.py
```diff
@@ -437,7 +437,8 @@
               'editor/close file 1': "Ctrl+W",
               'editor/close file 2': "Ctrl+F4",
               'editor/run cell': CTRL + '+Return',
-              'editor/run cell and advance': 'Shift+Return',
+              'editor/run cell and advance': (
+                  CTRL + '+Shift+Return' if MAC else 'Shift+Return'),
```
ccordoba12, line 441: @dalthviz, we can't do this because it would change the way we run cells in Spyder, which (I think) many people would find unpleasant.

So, if the test you introduced is failing only in macOS, please skip it there.

dalthviz: Ok :+1: but the reason I change its because seems like that shortcut collides with something else (don't know with what other shortcut) and even when you try to reapply it over the same action on MacOS.

Seems like if you try to change any action shortcut to `Shift+Return` it doesn't work although other shortcut changes work as expected, is that valid?

ccordoba12: Ok, that seems really serious. Thanks @dalthviz for finding it!

@steff456, could you confirm if `Shift+Enter` can't be remapped to perform other actions in macOS?

dalthviz: Note: Checking a little bit seems that the ambiguous warning only appears when remapping `Shift+Return` in the editor

spyder/app/tests/test_mainwindow.py
```diff
@@ -2421,6 +2421,60 @@ def test_preferences_change_font_regression(main_window, qtbot):
...
+    # Check execution of shortcut
+    # Create new file
+    main_window.editor.new()
+    code_editor = main_window.editor.get_focus_widget()
+    code_editor.set_text('print(0)\nprint(1)\nprint(2)')
+
+    with qtbot.waitSignal(shell.executed):
+        qtbot.keyClick(code_editor, Qt.Key_Return, modifier=Qt.ShiftModifier)
+    qtbot.waitUntil(lambda: 'print(0)' in shell._control.toPlainText())
+    assert 'print(1)' not in shell._control.toPlainText()
```
ccordoba12, line 2461: I think this is incorrect because this is what you get when you have the following code:

```python
print(0)
print(1)
print(2)
```

and press `Shift+Enter` (i.e. in case the shortcut was not changed correctly):

Selección_013

As you can see, there's no `print(1)` there.

ccordoba12: So I'd suggest to change the text to be executed to

    code_editor.set_text("print(0)\nprint('ññ')")

and assert that `ññ` is not present in `shell._control.toPlainText()`. That's because there are lot of other numbers in the console (for instance in the banner on the top) that could match other print's with numbers.

PR #13100, opened 2020-06-24
PR: Migrate RunConfig to Plugin with new API

### LINKED ISSUE #13099: Migrate RunConfig to Plugin with new API

goanpeca:
## Description of Changes

### Issue(s) Resolved

Fixes #13099

### Affirmation

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -669,11 +669,9 @@ class MainWindow(QMainWindow):
         from spyder.preferences.appearance import AppearanceConfigPage
         from spyder.preferences.general import MainConfigPage
         from spyder.preferences.shortcuts import ShortcutsConfigPage
-        from spyder.preferences.runconfig import RunConfigPage
         from spyder.preferences.maininterpreter import MainInterpreterConfigPage
         self.general_prefs = [MainConfigPage, AppearanceConfigPage,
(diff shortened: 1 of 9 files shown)


goanpeca:
Hi @dalthviz, this one is ready for QA testing. Thanks!

goanpeca:
This is ready for review @ccordoba12

goanpeca:
@dalthviz and @ccordoba12 this one is ready for review.

goanpeca:
>one quick question so I understand things better: why do you think it's important to create a plugin for this config page? Since the plugin API for Run is not used anywhere in this case, it seems unnecessary to me.

Again the idea of generalizing all these things is to be able to have a single way to handle all the things. All plugins are loaded in the same way no matter what. This also applies to config pages.

This "run" plugin is a plugin that depends on the Editor AND the Console so that it can do things, so I think having it as a separate plugin (besides simplifying code in preferences/ and in main window) opens the door to handle in a cleaner way how the console interacts with the editor to run stuff. (like the current implicit dependency the ipyconsole shellwidget has on the editor). This is the plugin that should be in charge of all the Run actions (not the editor). Also the core plugin would override the run action (that would be handled by this plugin). All these things will become apparent with the editor migration.

That is why all the "global preferences" should become a plugin.

ccordoba12:
Ok, thanks for the clarification.

PR #13111, opened 2020-06-26
PR: Emit sig_project_closed when switching projects 

ccordoba12:
- This can happen either when you have an open project and (i) Change to another, existing project; or (ii) Create a new project.
- It's important to emit that signal to let language servers know that the previous open project needs to be removed from the list of workspaces.
- This won't be necessary once we support multiple workspaces in the interface, but for now I think it is.


ccordoba12:
@andfoy, please help me to review this one.

ccordoba12:
I talked to @andfoy about this one and he agreed it's what we need to do right now.

PR #13070, opened 2020-06-20
WIP: Add ability to hot start some plugins.

goanpeca:
## Description of Changes

* [ ] Wrote at least one-line docstrings (for any new functions)
* [ ] Added unit test(s) covering the changes (if testable)

* [ ] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

### Issue(s) Resolved

Fixes #13067

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct: @goanpeca

PR #13180, opened 2020-07-03
PR: Correctly handle visibility of plugins on close/start

### LINKED ISSUE #13127: All plugins in the new API are shown after startup
I'm seeing the following in master:

* I hid some of the plugins already migrated to the new API (e.g. Online Help), then close Spyder.
* After starting Spyder, I can see the plugin as visible again (when it should be hidden).

However, that behavior is not observed after running Spyder in safe mode (probably because the layout is applied for the first time and hides the plugins we don't want to show).

goanpeca:
## Description of Changes

### Issue(s) Resolved

Fixes #13127

### Affirmation

### DIFF
--- a/spyder/api/plugins.py
+++ b/spyder/api/plugins.py
@@ -896,9 +896,9 @@ class SpyderPluginV2(QObject, SpyderActionMixin, SpyderOptionMixin):
                     # TODO: Remove when migration is done, move to logger.
                     # Needed to check how the options API needs to cover
                     # options from all plugins
-                    print('\nspyder.api.plugins.options_from_conf\n'
-                          'Warning: option "{}" not found in section "{}" '
-                          'of configuration!'.format(option, self.NAME))
+                    # print('\nspyder.api.plugins.options_from_conf\n'
+                    #       'Warning: option "{}" not found in section "{}" '
+                    #       'of configuration!'.format(option, self.NAME))
 
                     # Currently when the preferences dialog is used, a set of
                     # changed options is passed.
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -1975,7 +1975,6 @@ class MainWindow(QMainWindow):
     # --- Layouts
     def setup_layout(self, default=False):
         """Setup window layout"""
-        print("setup_layout")
         prefix = 'window' + '/'
         settings = self.load_window_settings(prefix, default)
         hexstate = settings[0]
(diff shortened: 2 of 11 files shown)


goanpeca:
I still need to add some things to other plugins but could you check if this solves the problem you saw @ccordoba12 ?

spyder/api/plugins.py
```diff
@@ -896,9 +896,9 @@ def options_from_conf(self, options):
                     # TODO: Remove when migration is done, move to logger.
                     # Needed to check how the options API needs to cover
                     # options from all plugins
-                    print('\nspyder.api.plugins.options_from_conf\n'
-                          'Warning: option "{}" not found in section "{}" '
-                          'of configuration!'.format(option, self.NAME))
+                    # print('\nspyder.api.plugins.options_from_conf\n'
+                    #       'Warning: option "{}" not found in section "{}" '
+                    #       'of configuration!'.format(option, self.NAME))
```
ccordoba12, line 901: Why did you comment this?

spyder/app/mainwindow.py
```diff
@@ -1617,6 +1619,14 @@ def update_lsp_logs(self):
     def post_visible_setup(self):
         """Actions to be performed only after the main window's `show` method
         was triggered"""
+        # Show plugins that were visible
+        for plugin_name, plugin in self._DOCKABLE_WIDGETS.items():
+            plugin.change_visibility(plugin.get_conf_option("visible"))
+            if plugin.get_conf_option("visible"):
+                plugin.dockwidget.show()
+            else:
+                plugin.dockwidget.hide()
```
ccordoba12, line 1628: Could you explain what things changed in the new API that make this necessary now? I just want to understand why this was not necessary in Spyder 4.

ccordoba12:
> I still need to add some things to other plugins but could you check if this solves the problem you saw @ccordoba12 ?

Yes, it does. Thanks for your work on this!

spyder/app/mainwindow.py
goanpeca: I am not sure if the WindowState that gets saved actually saves this, but I was not able to get the original state to work unless I did this. So I am not sure. This also relates to work on https://github.com/spyder-ide/spyder/pull/13070. When that is in place this will not be needed.

ccordoba12: Ok, that's really good too know. Thanks!

Then please let me know when this is ready to be merged.

goanpeca:
Superseded by #13070

PR #13184, opened 2020-07-03
PR: Start online help server only after post visible setup and only when visible

goanpeca:
## Description of Changes

### Issue(s) Resolved

Fixes #

### Affirmation

### DIFF
--- a/spyder/plugins/onlinehelp/plugin.py
+++ b/spyder/plugins/onlinehelp/plugin.py
@@ -66,6 +66,13 @@ class OnlineHelp(SpyderDockablePlugin):
         widget.load_history(self.load_history())
         widget.sig_load_finished.connect(self.sig_load_finished)
 
+    def on_mainwindow_visible(self):
+        # Only initialize the server if the dockwidget is activated
+        widget = self.get_widget()
+        widget.server = None
+        widget.server_enabled = True
+        widget.sig_toggle_view_changed.connect(lambda v: widget.initialize())
+
     def update_font(self):
         self.get_widget().reload()
 
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -330,6 +330,7 @@ class MainWindow(QMainWindow):
         plugin.sig_status_message_requested.connect(self.show_status_message)
 
         if isinstance(plugin, SpyderDockablePlugin):
+            self._DOCKABLE_PLUGINS[plugin.NAME] = plugin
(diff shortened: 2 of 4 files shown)


spyder/plugins/onlinehelp/plugin.py
```diff
@@ -66,6 +66,13 @@ def register(self):
         widget.load_history(self.load_history())
         widget.sig_load_finished.connect(self.sig_load_finished)
 
+    def on_mainwindow_visible(self):
```
ccordoba12, line 69: I don't like this solution too much because starting the Pydoc server doesn't depend on the main window being visible. It depends on the plugin being visible. So, although this works, I find it a workaround that makes things confusing (I'm just thinking on the maintainability of this code in the future).

So I'd prefer to use the previous solution, which in the new API would translate (I think) to making `widget.server_enabled = True` when `widget.sig_toggle_view_changed` is emitted and then running `widget.initialize()`.

The problem with my proposal seems to be (if I'm not mistaken) that now `widget.sig_toggle_view_changed` is emitted for all plugins at startup, whether they are visible or not. If that's the case, could you explain why?

goanpeca:
>I don't like this solution too much because starting the Pydoc server doesn't depend on the main window being visible. It depends on the plugin being visible. So, although this works, I find it a workaround that makes things confusing (I'm just thinking on the maintainability of this code in the future).

This is an intermediate solution while https://github.com/spyder-ide/spyder/pull/13070 is finished. Since all plugins currently start and appear (hence triggering the signal) the only way was to have some intermediate state flag, and defer this flag to the latest state possible (on_window_visible..)

>So I'd prefer to use the previous solution, which in the new API would translate (I think) to making widget.server_enabled = True when widget.sig_toggle_view_changed is emitted and then running widget.initialize().
>The problem with my proposal seems to be (if I'm not mistaken) that now widget.sig_toggle_view_changed is emitted for all plugins at startup, whether they are visible or not. If that's the case, could you explain why?

Yes, this is correct. In the previous implementation the toggle actions from the menu and the plugin were different (and duplicated). Now they are one an the same, so there is no way to avoid having the toggle view not get triggered at startup, unless a plugins is effectively NOT loaded to begin with as it is the plan with https://github.com/spyder-ide/spyder/pull/13070.

ccordoba12:
> This is an intermediate solution while #13070 is finished.

Ok, that's really good to know. I'll leave comments with TODO's for them to be removed after PR #13070 then.

goanpeca:
Superseded by #13070

PR #13208, opened 2020-07-07
PR: Fix visible plugins on restart and online server start

### LINKED ISSUE #13127: All plugins in the new API are shown after startup

### LINKED ISSUE #13142: Spyder prints "code True" in internal console
When starting spyder with the master, I see:
```
Spyder Internal Console

This console is used to report application
internal errors and to inspect Spyder
internals with the following commands:
  spy.app, spy.window, dir(spy)

Please don't use it to run your code

>>> code True
```
`code True` is printed here:
https://github.com/spyder-ide/spyder/blob/fa39a2f4fd460bc6448ffc4845c3af95e539877b/spyder/plugins/onlinehelp/widgets.py#L263
Is that expected?

goanpeca:
## Description of Changes

This PR correctly addresses some issues with plugins on startup and on close. With this PR online server is never started unless the plugin was open on shutdown. This includes the first startup as well.

### Issue(s) Resolved

Fixes #13142
Fixes #13127

### Affirmation

### DIFF
--- a/spyder/api/widgets/auxiliary_widgets.py
+++ b/spyder/api/widgets/auxiliary_widgets.py
@@ -9,6 +9,7 @@ Spyder API auxiliary widgets.
 """
 
 # Third party imports
+from qtpy.QtCore import Signal
 from qtpy.QtWidgets import QHBoxLayout, QMainWindow, QWidget
 import qdarkstyle
 
@@ -21,6 +22,10 @@ class SpyderWindowWidget(QMainWindow):
     """
     MainWindow subclass that contains a Spyder Plugin.
     """
+    sig_closed = Signal()
+    """
+    This signal is emitted when this widget is closed.
+    """
 
     def __init__(self, widget):
         super().__init__()
@@ -29,6 +34,11 @@ class SpyderWindowWidget(QMainWindow):
         # Setting interface theme
         if is_dark_interface():
             self.setStyleSheet(qdarkstyle.load_stylesheet())
+
+    def closeEvent(self, event):
+        """Override Qt method to emit a custom `sig_close` signal."""
+        super().closeEvent(event)
+        self.sig_closed.emit()
 
 
 class MainCornerWidget(QWidget):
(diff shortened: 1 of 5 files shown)


spyder/app/mainwindow.py
```diff
@@ -2861,7 +2853,9 @@ def closing(self, cancelable=False):
         for plugin in (self.widgetlist + self.thirdparty_plugins):
             # New API
             try:
-                plugin.close_window()
+                if plugin.is_undocked():
```
goanpeca, line 2856: This was the problem of why all plugins were appearing. The new API needs to check that the plugin was undocked.

spyder/api/widgets/auxiliary_widgets.py
```diff
@@ -21,6 +22,7 @@ class SpyderWindowWidget(QMainWindow):
     """
     MainWindow subclass that contains a Spyder Plugin.
     """
+    sig_closed = Signal()
```
ccordoba12, line 25: Please add a docstring for this signal.

spyder/plugins/onlinehelp/widgets.py
```diff
@@ -314,11 +313,13 @@ def load_history(self, history):
         """
         self.url_combo.addItems(history)
 
-    def initialize(self):
+    def initialize(self, val=True):
```
ccordoba12, line 316: Since this is a slot now to `sig_toggle_view_changed`, please decorate it above with `Slot(bool)`. Also, please change `val` to `checked`, which makes more sense.

spyder/api/widgets/auxiliary_widgets.py
goanpeca: Fixed

spyder/plugins/onlinehelp/widgets.py
goanpeca: Fixed

PR #13209, opened 2020-07-07
PR: Improve Files plugin UI

### LINKED ISSUE #13179: Some improvements for the Files plugin
While taking a look at this pane with @juanis2112 to update its documentation, we noticed several things that are worth improving on it:

* Remove the entry called "Show icons and text"

    Selección_005

    because it only changes if we show text next to the icons on top of the pane.

    Selección_006

    Instead we should show only the icons.

* Add an entry to its options menu and Preferences page to show/hide hidden files and have it off by default. The problem is at the moment we are showing way too many directories in Posix systems that you really don't want to see all the time.

* On its context menu

    Selección_007

    we think it's better to rename `Module...` for `Python script...` and change its icon to the one used for the PYTHONPATH manager (i.e. the colored, two snakes one); and `Package` to `Python package`, and remove its icon.

* The `Show all files` entry should be the first one in its Options menu. `Edit filename filters` needs to be renamed to `Display files with these extensions...`, deactivated when `Show all files` is active and placed as the second entry.

* When pressing `Display files with these extensions...` we need to show a QTextEdit to edit those extensions, instead of QLineEdit:

    Selección_008

    That's because it's quite cumbersome to use a QLineEdit for that, specially if there are too many extensions.

* In the Preferences entry for that pane, we need to change  `Edit filename filters...` for `Display files with the following extensions:`

    Selección_009

dalthviz:
## Description of Changes

### Issue(s) Resolved

Fixes #13179 

### Affirmation

### DIFF
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -279,7 +279,7 @@ DEFAULTS = [
               'enable': True,
               'wrap': True,
               'name_filters': NAME_FILTERS,
-              'show_hidden': True,
+              'show_hidden': False,
               'show_all': True,
               'single_click_to_open': False,
               }),
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -281,7 +281,6 @@ DEFAULTS = [
               'name_filters': NAME_FILTERS,
               'show_hidden': True,
               'show_all': True,
-              'show_icontext': False,
               'single_click_to_open': False,
               }),
             ('find_in_files',
@@ -631,4 +630,4 @@ NAME_MAP = {
 #    or if you want to *rename* options, then you need to do a MAJOR update in
 #    version, e.g. from 3.0.0 to 4.0.0
 # 3. You don't need to touch this value if you're just adding a new option
-CONF_VERSION = '57.4.0'
+CONF_VERSION = '58.4.0'
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -630,4 +630,4 @@ NAME_MAP = {
 #    or if you want to *rename* options, then you need to do a MAJOR update in
 #    version, e.g. from 3.0.0 to 4.0.0
 # 3. You don't need to touch this value if you're just adding a new option
-CONF_VERSION = '58.4.0'
+CONF_VERSION = '58.0.0'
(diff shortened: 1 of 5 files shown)


spyder/config/main.py
```diff
@@ -630,4 +629,4 @@
 #    or if you want to *rename* options, then you need to do a MAJOR update in
 #    version, e.g. from 3.0.0 to 4.0.0
 # 3. You don't need to touch this value if you're just adding a new option
-CONF_VERSION = '57.4.0'
+CONF_VERSION = '58.4.0'
```
ccordoba12, line 632: This needs to be `58.0.0`

spyder/plugins/explorer/widgets/explorer.py
```diff
@@ -323,31 +334,40 @@ def reset_icon_provider(self):
 
     #---- Context menu
     def setup_common_actions(self):
-        """Setup context menu common actions"""
+        """Set-up context menu common actions."""
+        # Show all files
+        self.all_action = create_action(self, _("Show all files"),
+                                        toggled=self.toggle_all)
```
ccordoba12, line 340: Please format this code in the same way as you did with `self.hidden_action` below.

dalthviz:
No idea why pep8speaks put the comment

spyder/plugins/explorer/confpage.py
```diff
@@ -26,15 +26,17 @@ def setup_page(self):
         # Widgets
         general_widget = QWidget()
         check_show_all = newcb(_("Show all files"), 'show_all')
-        check_icon = newcb(_("Show icons and text"), 'show_icontext')
+        check_show_hidden_files = newcb(_("Show hidden files"), 'show_hidden')
         check_single_click = newcb(_("Single click to open files"),
                                    'single_click_to_open')
         edit_filename_filters = self.create_textedit(
-            _("Edit filename filters..."),
+            _("Display files with these extensions:"),
```
ccordoba12, line 33: Diplay -> Show

spyder/plugins/explorer/widgets/explorer.py
```diff
@@ -323,31 +334,42 @@ def reset_icon_provider(self):
...
+        # Show hidden files
+        self.hidden_action = create_action(
+            self,
+            _("Show hidden files"),
+            toggled=self.toggle_hidden,
+        )
         # Filters
         self.filters_action = create_action(
-            self, _("Edit filename filters..."), None, ima.icon('filter'),
+            self, _("Display files with these extensions..."), None,
```
ccordoba12, line 351: Display -> Show

PR #13216, opened 2020-07-07
PR: Fix segfault when right-clicking any entry in the project explorer

trollodel:
## Description of Changes

I just flip the condition to prevent `fsmodel.type` to make a segfault. I did not discover why this happen, but I suspect that is related to the model itself.

### Issue(s) Resolved

None

### Affirmation

I certify the above statement is true and correct: trollodel

P.S. This is my first contribute on Spyder. I hope that I respect the CONTRIBUTING correctly.

### DIFF
--- a/spyder/plugins/explorer/widgets/explorer.py
+++ b/spyder/plugins/explorer/widgets/explorer.py
@@ -577,8 +577,8 @@ class DirView(QTreeView):
         # empty folder. See spyder-ide/spyder#13004
         if len(fnames) == 1 and self.selectionModel():
             rows = self.selectionModel().selectedRows()
-            if (self.fsmodel.type(rows[0]) == 'Folder' and
-                    self.fsmodel.rowCount(rows[0]) == 0):
+            if (self.fsmodel.rowCount(rows[0]) == 0 and
+                    self.fsmodel.type(rows[0]) == 'Folder'):
                 fnames = []
         if not fnames:
             for action in actions:


jnsebgosselin:
Thanks a lot for your contribution @trollodel 

@ccordoba12 I'm experiencing the segfault crash on every right-click on any item in the project explorer, so I'm going to test and review this one.

jnsebgosselin:
Unfortunately, this doesn't solve the problem. As soon as the condition `self.fsmodel.rowCount(rows[0]) == 0` is met and `self.fsmodel.type` is called, Spyder crash.

Calling `self.fsmodel.isDir` also makes Spyder crash... I don't understand why this is happening.

jnsebgosselin:
Ok, I think I know why this is happening. I'm trying to find a way to solve this and when I do, I'll propose a solution here.

ccordoba12:
@trollodel, thanks so much for alerting us about this critical problem! Although your solution was not precisely the change necessary to fix it, we really appreciate your effort!

For next time, please don't forget to take a look at our [contributing guidelines](https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md) to learn how to do a successful pull request. For instance, in this case you created your pull request against our `master` branch (instead of our `4.x` one) and you forgot to create a branch for your pull request (i.e. you committed your changes directly against `master`).

Please don't feel bad about it, every project has its own rules and also git/github have a very steep learning curve. I just wanted to mention these couple of things so you can do better next time.

ccordoba12:
By the way, this problem was fixed by @jnsebgosselin on #13226.

trollodel:
> For instance, in this case you created your pull request against our `master` branch (instead of our `4.x` one).
To be honest, I discover this issue when I try to use the master branch version. I didn't see that when I use the stable version.
So I target the fix for master.

ccordoba12:
Ok, no problem. Thanks for your help anyway.

PR #13317, opened 2020-07-16
PR: Add name validation and status of project methods

goanpeca:
## Description of Changes
- Now the project type methods (create/open/close) return a bool and a message in case of issues and a MessageBox is displayed
- The project type now only has access to the parent plugin via the `self.plugin` attribute
- The project type now provides a (optional) static method for validating project names. The result of this validation is displayed on the bottom of the create project dialog.

Screen Shot 2020-07-16 at 10 04 23

### Affirmation

### DIFF
--- a/spyder/plugins/projects/api.py
+++ b/spyder/plugins/projects/api.py
@@ -38,8 +38,8 @@ class BaseProjectType:
     """
     ID = None
 
-    def __init__(self, root_path, projects_plugin):
-        self.projects_plugin = projects_plugin
+    def __init__(self, root_path, parent_plugin=None):
+        self.plugin = parent_plugin
         self.root_path = root_path
         self.open_project_files = []
         self.open_non_project_files = []
@@ -94,38 +94,67 @@ class BaseProjectType:
 
         return list(OrderedDict.fromkeys(recent_files))
 
-    def get_plugin(self, plugin_name):
-        """
-        Return a plugin by unique `plugin_name`.
-        """
-        return self.projects_plugin.get_plugin(plugin_name)
-
     # --- API
     # ------------------------------------------------------------------------
     @staticmethod
-    def get_name(self):
+    def get_name():
         """
         Provide a human readable version of NAME.
         """
         raise NotImplementedError("Must implement a `get_name` method!")
 
+    @staticmethod
+    def validate_name(path, name):
+        """
+        Validate the project's name.
+
+        Returns
+        -------
+        tuple
+            The first item (bool) indicates if the name was validated
+            successfully, and the second item (str) indicates the error
+            message, if any.
+        """
+        return True, ""
+
     def create_project(self):
         """
         Create a project and do any additional setup for this project type.
+
+        Returns
+        -------
+        tuple
+            The first item (bool) indicates if the project was created
+            successfully, and the second item (str) indicates the error
+            message, if any.
         """
-        raise NotImplementedError("Must implement a `create_project` method!")
+        return False, "A ProjectType must define a `create_project` method!"
 
     def open_project(self):
         """
         Open a project and do any additional setup for this project type.
+
+        Returns
+        -------
+        tuple
+            The first item (bool) indicates if the project was opened
+            successfully, and the second item (str) indicates the error
+            message, if any.
         """
-        raise NotImplementedError("Must implement a `open_project` method!")
+        return False, "A ProjectType must define an `open_project` method!"
 
     def close_project(self):
         """
         Close a project and do any additional setup for this project type.
+
+        Returns
+        -------
+        tuple
+            The first item (bool) indicates if the project was closed
+            successfully, and the second item (str) indicates the error
+            message, if any.
         """
-        raise NotImplementedError("Must implement a `close_project` method!")
+        return False, "A ProjectType must define a `close_project` method!"
 
 
 class EmptyProject(BaseProjectType):
@@ -136,10 +165,10 @@ class EmptyProject(BaseProjectType):
         return _("Empty project")
 
     def create_project(self):
-        pass
+        return True, ""
 
     def open_project(self):
-        pass
+        return True, ""
 
     def close_project(self):
-        pass
+        return True, ""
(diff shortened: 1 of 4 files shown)


goanpeca:
Hi @nerohmot, any suggestions/ideas?

spyder/plugins/projects/api.py
```diff
@@ -94,38 +94,67 @@ def get_recent_files(self):
...
     @staticmethod
-    def get_name(self):
+    def get_name():
         """
         Provide a human readable version of NAME.
         """
         raise NotImplementedError("Must implement a `get_name` method!")
 
+    @staticmethod
+    def validate_name(path, name):
```
ccordoba12, line 107: Why is it necessary to validate the project's name?

spyder/plugins/projects/api.py
```diff
@@ -94,38 +94,67 @@ def get_recent_files(self):
...
+        """
+        Validate the name of the project.
```
ccordoba12, line 109: Please change this to

    Validate the project's name.

spyder/plugins/projects/api.py
```diff
@@ -94,38 +94,67 @@ def get_recent_files(self):
...
+
+        Returns
+        -------
+        tuple
+            The first item (bool) indicates if the project was created
+            successfully, and the second item (str) indicates the error
+            message, if any.
         """
-        raise NotImplementedError("Must implement a `create_project` method!")
+        return None, "A ProjectType must define a `create_project` method!"
```
ccordoba12, line 131: This doesn't correspond to the docstring you wrote above because you're not returning a bool but `None`.

spyder/plugins/projects/api.py
```diff
@@ -94,38 +94,67 @@ def get_recent_files(self):
...
+
+        Returns
+        -------
+        tuple
+            The first item (bool) indicates if the project was opened
+            successfully, and the second item (str) indicates the error
+            message, if any.
         """
-        raise NotImplementedError("Must implement a `open_project` method!")
+        return None, "A ProjectType must define an `open_project` method!"
```
ccordoba12, line 144: Same problem here.

spyder/plugins/projects/api.py
```diff
@@ -94,38 +94,67 @@ def get_recent_files(self):
...
+
+        Returns
+        -------
+        tuple
+            The first item (bool) indicates if the project was closed
+            successfully, and the second item (str) indicates the error
+            message, if any.
         """
-        raise NotImplementedError("Must implement a `close_project` method!")
+        return None, "A ProjectType must define a `close_project` method!"
```
ccordoba12, line 157: And same here.

spyder/plugins/projects/plugin.py
```diff
@@ -321,8 +318,14 @@ def _create_project(self, root_path, project_type=EmptyProject.ID,
...
         if project_type in project_types:
-            project = project_types[project_type](root_path, self)
-            project.create_project()
+            project_type_class = project_types[project_type]
+            project = project_type_class(
+                root_path, project_type_class._PARENT_PLUGIN)
+
+            created_succesfully, message = project.create_project()
+            if not created_succesfully:
+                QMessageBox.warning(self, "Project creation", message)
```
ccordoba12, line 327: Instead of using a `warning` here, I think we should show an error.

spyder/plugins/projects/plugin.py
```diff
@@ -321,8 +318,14 @@ def _create_project(self, root_path, project_type=EmptyProject.ID,
...
+                # FIXME: What do do in failed creation?
```
ccordoba12, line 328: I think for now you should put a `return` here, so the project creation is aborted at this point.

spyder/plugins/projects/api.py
goanpeca: @nerohmot needs this.

nerohmot: a valid name for python files does include the underscore `_`, however we don't want the **project name** to include an underscore in the project name, as the resulting test-programs (in the project) all start with the project name (it is always something relatively small like `CTCA`, `HATB`, `MN99254`) and we use underscores to further differentiate in a structured way.

When the user has the ability to use underscores in the project name, he would be encouraged to start some kind of differentiation of his own which might confuse the system itself (or attached tools) and if not, will make confusing test program names for other users of the programs.

hope it makes a bit of sense :stuck_out_tongue_winking_eye: 

We did have this in the mock-up, but as the **project name** now is given in spyder's `create new project` dialog we are in a 'post-factum' position, with no means to warn the user that the name is not permitted. (for our plugin) :frowning_face:

We discussed how we could handle such case, and the resolution was that the plugin (can) provide a regex (that outlines VALID names) and a 'fail message'. This way the project type plugins don't have to jump to awkward hoops (remember, the project directory and the .spyproject directory are already created whenever the plugin receives control) to get the message to the user.

spyder/plugins/projects/plugin.py
nerohmot: I think that if the plugin reports an un-successful project creation (by pressing for example the cancel button), spyder can go ahead and remove the whole project directory, regardless what is inside.

spyder/plugins/projects/api.py
ccordoba12: > hope it makes a bit of sense

Yes, it does now. Thanks for the explanation.

spyder/plugins/projects/plugin.py
ccordoba12: I agree. Let's go with option 1.

spyder/plugins/projects/api.py
goanpeca: Fixed

spyder/plugins/projects/api.py
goanpeca: Fixed

spyder/plugins/projects/api.py
goanpeca: Fixed

spyder/plugins/projects/api.py
goanpeca: Fixed

spyder/plugins/projects/plugin.py
goanpeca: Fixed

PR #13359, opened 2020-07-22
PR: Add a "Show in external file browser" shortcut (Editor)

### LINKED ISSUE #13158: Feature request: keyboard shortcut for "Show in external file explorer"
**Feature request: keyboard shortcut for "Show in external file explorer"**

I use this feature quite often, and as a fan of keyboard shortcuts it would be very useful. I am using the "win + E" to open explorer all the time on windows, and I believe using something like "Ctrl + E" makes some amount of sense to open an explorer window to that file. I think it would also be totally fine to leave it with no pre-bound shortcut, and leave it to the user to decide to bind one.

I tried adding the shortcut myself, but I have not been successful in figuring out how to add new shortcuts, so I have no pull request to submit. I got so far as to get a listing in the settings menu, but nothing happened when the key combination was pressed. I did this by adding an entry to `config.main.DEFAULTS`:

```
DEFAULTS = [
    ...
    ("shortcuts", 
        {
         ...
         'editor/show in external file explorer': "Ctrl+E",
```

 and `plugins.editor.widgets.editor.EditorStack.create_shortcuts` 

```
 external_fileexp = CONF.config_shortcut(
     self.show_in_external_file_explorer,
     context="Editor",
     name="Show in external file explorer",
     parent=self)
```

* Spyder version: 4.1.3
* Python version: 3.8.2
* Qt version: 5.12.8
* PyQt version: 5.12.3

athompson673:
I finally spent the time to (somewhat) figure out how to use git... (a certain john smith accurately [sums up](https://stackoverflow.com/questions/25101983/can-i-revert-commits-directly-on-github#comment73500278_25102008) my feelings on using git as a full time engineer, and occasional programmer)

I also realized I somehow have two accounts on github, so please ignore the other one -_-

### Issue(s) Resolved

Fixes  #13158

### Affirmation

I certify the above statement is true and correct: athompson673

### DIFF
--- a/spyder/plugins/editor/widgets/editor.py
+++ b/spyder/plugins/editor/widgets/editor.py
@@ -516,7 +516,15 @@ class EditorStack(QWidget):
         else:
            text= _("Show in external file explorer")
         external_fileexp_action = create_action(self, text,
-                                triggered=self.show_in_external_file_explorer)
+                                triggered=self.show_in_external_file_explorer,
+                                shortcut=CONF.get_shortcut(context="Editor",
+                                        name="show in external file explorer"),
+                                context=Qt.WidgetShortcut)
+        
+        parent.plugin.register_shortcut(external_fileexp_action, 
+                                        context="Editor",
+                                        name="show in external file explorer", 
+                                        add_shortcut_to_tip=True)
 
         self.menu_actions = actions + [external_fileexp_action,
                                        None, switcher_action,
@@ -836,6 +844,12 @@ class EditorStack(QWidget):
             name="close split panel",
             parent=self)
 
+        external_fileexp = CONF.config_shortcut( #XXX here
+            self.show_in_external_file_explorer,
+            context="Editor",
+            name="show in external file explorer",
+            parent=self)
+            
         # Return configurable ones
         return [inspect, set_breakpoint, set_cond_breakpoint, gotoline, tab,
                 tabshift, run_selection, new_file, open_file, save_file,
@@ -846,7 +860,7 @@ class EditorStack(QWidget):
                 go_to_next_cell, go_to_previous_cell, re_run_last_cell,
                 prev_warning, next_warning, split_vertically,
                 split_horizontally, close_split,
-                prevtab, nexttab]
+                prevtab, nexttab, external_fileexp]
 
     def get_shortcut_data(self):
         """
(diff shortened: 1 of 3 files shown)


athompson673:
It would seem the call to `parent.plugin.register_shortcut` in "spyder/plugins/editor/widgets/editor.py" in `EditorStack.__init__` is problematic, as the type of `parent` is not the same all the time (only for unit test cases?). Some of the errors show `self.parent` or `parent.plugin` as `None` where others use `EditorPluginExample` for both. Is it not intended to be able to access these values from the widget itself, or should I simply add more type checking to protect the call to `register_shortcut`?

CAM-Gerlach:
> It would seem the call to `parent.plugin.register_shortcut` in "spyder/plugins/editor/widgets/editor.py" in `EditorStack.__init__` is problematic, as the type of `parent` is not the same all the time (only for unit test cases?). Some of the errors show `self.parent` or `parent.plugin` as `None` where others use `EditorPluginExample` for both. Is it not intended to be able to access these values from the widget itself, or should I simply add more type checking to protect the call to `register_shortcut`?

@ccordoba12 @dalthviz ?

dalthviz:
Thanks @athompson673 for your contribution!  

About the question, I will need to check since I'm unsure right away but surely I will try to check next week :+1:

spyder/plugins/editor/widgets/editor.py
```diff
@@ -515,8 +515,19 @@ def __init__(self, parent, actions):
...
+            shortcut=CONF.get_shortcut(context="Editor",
+                                       name="show in external file explorer"),
+            context=Qt.WidgetShortcut)
+
+        if self.parent() is not None:
+            plugin = self.parent().plugin
+            plugin.register_shortcut(external_fileexp_action,
+                                     context="Editor",
+                                     name="show in external file explorer",
+                                     add_shortcut_to_tip=True)
```
ccordoba12, line 530: This is not the right way to add a new shortcut. You should register it instead in the `create_shortcuts` method below.

The the problems you mentioned will simply go away.

athompson673: I must have left that part in assuming it was still needed from my testing. I already actually had an entry in `create_shortcuts` in fact.

ccordoba12: Great!

PR #13398, opened 2020-07-27
PR: Add option to disable tooltip on toolbars when creating actions

### LINKED ISSUE #13378: Some problems with the Find plugin in master
After merging PR #13376 I found the following problems with the Find plugin in master:

* The `Find in Files` action in the `Search` menu does nothing when Spyder is started with clean settings. That's because now it's associated to the `Search` button in its toolbar:

    Selección_005

   So it's clear why it does nothing: because there's no text to search by default. But in Spyder 4 that action was associated to the toggle action of the plugin, so it showed the plugin in case it was hidden (which is the default with clean settings).

* All buttons in the Find toolbar are now QAction's, whereas in Spyder 4 they were QToolButton's. That doesn't allow us to remove their tooltips because QActions always have an associated tooltip. For instance, in PR #13376 I removed the tooltip of the `Search` button  because it was really unnecessary.

----

The last problem points to a really significant change (switching QToolButton's for QAction's). I don't understand its purpose because it makes things less flexible for us in master (like the above issue clearly shows).

@goanpeca, could you explain your rationale for this? Also, is it possible for us to revert to QToolButton's everywhere where they were used in Spyder 4?

goanpeca:
## Description of Changes

### Issue(s) Resolved

Fixes #13378

### Affirmation

### DIFF
--- a/spyder/api/widgets/mixins.py
+++ b/spyder/api/widgets/mixins.py
@@ -443,7 +443,8 @@ class SpyderActionMixin:
         icon: QIcon,
             Icon for the action when applied to menu or toolbutton.
         icon_text: str
-            Icon for text in toolbars.
+            Icon for text in toolbars. If True, this will also disable
+            the tooltip on this toolbutton if part of a toolbar.
         tip: str
             Tooltip to define for action on menu or toolbar.
         toggled: callable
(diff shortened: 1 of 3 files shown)


goanpeca:
>Thanks @goanpeca for your fast response to solve the issues I reported on #13378. One quick question: does the ToolTipFilter class you added disable tooltips for all actions in toolbars?

@ccordoba12 the filter is only applied to actions that don't have `icon_text` (that is what the `icon_text` option for creating an action is) In this example.

image

ccordoba12:
> the filter is only applied to actions that don't have icon_text (that is what the icon_text option for creating an action is)

Ok, noted. That's what I needed to know. Thanks for the clarification!

PR #13440, opened 2020-07-31
PR: Create a simple code editor and use that for other plugins

goanpeca:
## Description of Changes

This is a preliminary step in the migration of the Editor Plugin to the new API. 

The CodeEditor is a very complex widget and has been used by many other plugins for providing a code editor, where the only basic need seems to have syntax highlighting capabilities at most.

Currently, the following widgets use the CodeEditor

### Variable Explorer

Object Editor (For python highlight)

### Before

Screen Shot 2020-08-03 at 8 30 17

### Now
Screen Shot 2020-08-03 at 8 27 26

## History Plugin
For python highlight

### Before

Screen Shot 2020-08-03 at 8 30 43

### Now
Screen Shot 2020-08-03 at 8 25 46

## Help Plugin

For the plain text view (For python highlight)

### Before

Screen Shot 2020-08-03 at 8 31 10

### Now

Screen Shot 2020-08-03 at 8 26 06

## Appearance plugin

For the Syntax Highlighter preview (For python highlight)

### Before

Screen Shot 2020-08-03 at 8 31 24

### Now

Screen Shot 2020-08-03 at 8 27 09

## Error report

For markdown highlight. This one uses more of the editor cursor and selection capabilities, but only a handful.

### Before

Screen Shot 2020-08-03 at 8 31 37

### Now

Screen Shot 2020-08-03 at 8 26 25

As an initial attempt to really decouple the different plugins the simplest way (by far) is to provide a simple code editor with the needed capabilities in the `spyder.widgets` module. This boils down to syntax highlighting and a few options. No Panels, no Extensions etc...

This has the following benefits.
- **Performance**: Spyder now can live without 5 extra instances (in the worst case) of a complex (and heavier) widget like the fully blown code editor. This could be more if the History plugin eventually tracks consoles separately.
- **Plugin dependencies**: hardcoded required dependencies are now optional so Spyder can start without the Code Editor and have all the other plugins work normally (History, VarExp, Help, Appearance).
- **Cleaner code** that avoids circular (or impossible) dependencies between plugins. (The appearance plugin needs to be loaded before the editor, but it uses the editor...)

- The new API dependencies (except for the Editor) can now really be tested.
- With the new work on syntax highlighting, having a simple code editor to test things might help in the process.

### (minor) Drawbacks:

   * Appearance: some things are not visible now on this panel. But, this panel is just a preview and the most important thing is the actual colors, not having line numbers in this context seems like a really minor issue.

**Update**

The simple code editor now implements line numbers, so there is no longer any limitation with replacing CodeEditor by this SimpleCodeEditor on the before mentioned places.

Some small cleanup is needed still but so far

### Affirmation

### DIFF
--- a/spyder/plugins/appearance/confpage.py
+++ b/spyder/plugins/appearance/confpage.py
@@ -18,10 +18,8 @@ from spyder.config.gui import (get_font, is_dark_font_color, is_dark_interface,
 from spyder.config.manager import CONF
 from spyder.config.utils import is_gtk_desktop
 from spyder.plugins.appearance.widgets import SchemeEditor
-# TODO: This should load befor the editor, but depends on the editor
-# the code editor, needs to be moved to core.
-from spyder.plugins.editor.widgets.codeeditor import CodeEditor
 from spyder.utils import syntaxhighlighters
+from spyder.widgets.simplecodeeditor import SimpleCodeEditor
 
 # Localization
 _ = get_translation('spyder')
@@ -88,7 +86,7 @@ class AppearanceConfigPage(PluginConfigPage):
         self.delete_button = QPushButton(_("Delete scheme"))
         self.reset_button = QPushButton(_("Reset to defaults"))
 
-        self.preview_editor = CodeEditor(self)
+        self.preview_editor = SimpleCodeEditor(self)
         self.stacked_widget = QStackedWidget(self)
         self.scheme_editor_dialog = SchemeEditor(parent=self,
                                                  stack=self.stacked_widget)
@@ -321,27 +319,23 @@ class AppearanceConfigPage(PluginConfigPage):
         """
         text = ('"""A string"""\n\n'
                 '# A comment\n\n'
-                '# %% A cell\n\n'
                 'class Foo(object):\n'
                 '    def __init__(self):\n'
                 '        bar = 42\n'
                 '        print(bar)\n'
                 )
-        show_blanks = CONF.get('editor', 'blank_spaces')
-        update_scrollbar = CONF.get('editor', 'scroll_past_end')
-        underline_errors = CONF.get('editor', 'underline_errors')
+
         if scheme_name is None:
             scheme_name = self.current_scheme
-        self.preview_editor.setup_editor(linenumbers=True,
-                                         markers=True,
-                                         tab_mode=False,
-                                         font=get_font(),
-                                         show_blanks=show_blanks,
-                                         underline_errors=underline_errors,
-                                         color_scheme=scheme_name,
-                                         scroll_past_end=update_scrollbar)
-        self.preview_editor.set_text(text)
+
+        self.preview_editor.setup_editor(
+            font=get_font(),
+            color_scheme=scheme_name,
+            show_blanks=False,
+            scroll_past_end=False,
+        )
         self.preview_editor.set_language('Python')
+        self.preview_editor.set_text(text)
 
     # Actions
     # -------------------------------------------------------------------------
--- a/spyder/plugins/help/widgets.py
+++ b/spyder/plugins/help/widgets.py
@@ -58,7 +58,7 @@ class HelpWidgetActions:
     ToggleRichMode = 'toggle_rich_mode_action'
     ToggleShowSource = 'toggle_show_source_action'
     ToggleWrap = 'toggle_wrap_action'
-    CopyAction = "copy_action"
+    CopyAction = "help_widget_copy_action"
     SelectAll = "select_all_action",
 
 
@@ -205,7 +205,11 @@ class PlainText(QWidget):
 
         # Read-only simple code editor
         self.editor = SimpleCodeEditor(self)
-        self.editor.setup_editor(language='py')
+        self.editor.setup_editor(
+            language='py',
+            highlight_current_line=False,
+            linenumbers=False,
+        )
         self.editor.sig_focus_changed.connect(self.focus_changed)
         self.editor.setReadOnly(True)
         self.editor.setContextMenuPolicy(Qt.CustomContextMenu)
--- a/spyder/plugins/help/widgets.py
+++ b/spyder/plugins/help/widgets.py
@@ -14,7 +14,7 @@ import socket
 import sys
 
 # Third party imports
-from qtpy.QtCore import QUrl, Signal, Slot
+from qtpy.QtCore import Qt, QUrl, Signal, Slot, QPoint
 from qtpy.QtGui import QColor
 from qtpy.QtWebEngineWidgets import WEBENGINE, QWebEnginePage
 from qtpy.QtWidgets import (QActionGroup, QComboBox, QLabel, QLineEdit,
@@ -26,7 +26,6 @@ from spyder.api.translations import get_translation
 from spyder.api.widgets import PluginMainWidget
 from spyder.config.base import get_image_path, get_module_source_path
 from spyder.config.gui import is_dark_interface
-from spyder.plugins.editor.widgets import codeeditor
 from spyder.plugins.help.utils.sphinxify import (CSS_PATH, DARK_CSS_PATH,
                                                  generate_context, loading,
                                                  usage, warning)
@@ -36,6 +35,7 @@ from spyder.utils import programs
 from spyder.widgets.browser import FrameWebView
 from spyder.widgets.comboboxes import EditableComboBox
 from spyder.widgets.findreplace import FindReplace
+from spyder.widgets.simplecodeeditor import SimpleCodeEditor
 
 
 # Localization
@@ -58,6 +58,8 @@ class HelpWidgetActions:
     ToggleRichMode = 'toggle_rich_mode_action'
     ToggleShowSource = 'toggle_show_source_action'
     ToggleWrap = 'toggle_wrap_action'
+    CopyAction = "copy_action"
+    SelectAll = "select_all_action",
 
 
 class HelpWidgetOptionsMenuSections:
@@ -195,16 +197,18 @@ class PlainText(QWidget):
     # Signals
     focus_changed = Signal()
 
+    sig_custom_context_menu_requested = Signal(QPoint)
+
     def __init__(self, parent):
         QWidget.__init__(self, parent)
         self.editor = None
 
-        # Read-only editor
-        self.editor = codeeditor.CodeEditor(self)
-        self.editor.setup_editor(linenumbers=False, language='py',
-                                 scrollflagarea=False, edge_line=False)
-        self.editor.focus_changed.connect(lambda: self.focus_changed.emit())
+        # Read-only simple code editor
+        self.editor = SimpleCodeEditor(self)
+        self.editor.setup_editor(language='py')
+        self.editor.sig_focus_changed.connect(self.focus_changed)
         self.editor.setReadOnly(True)
+        self.editor.setContextMenuPolicy(Qt.CustomContextMenu)
 
         # Find/replace widget
         self.find_widget = FindReplace(self)
@@ -217,21 +221,24 @@ class PlainText(QWidget):
         layout.addWidget(self.find_widget)
         self.setLayout(layout)
 
+        self.editor.customContextMenuRequested.connect(
+            self.sig_custom_context_menu_requested)
+
     def set_font(self, font, color_scheme=None):
         """Set font"""
-        self.editor.set_font(font, color_scheme=color_scheme)
+        self.editor.set_color_scheme(color_scheme)
+        self.editor.set_font(font)
 
     def set_color_scheme(self, color_scheme):
         """Set color scheme"""
         self.editor.set_color_scheme(color_scheme)
 
     def set_text(self, text, is_code):
-        self.editor.set_highlight_current_line(is_code)
-        self.editor.set_occurrence_highlighting(is_code)
         if is_code:
             self.editor.set_language('py')
         else:
             self.editor.set_language(None)
+
         self.editor.set_text(text)
         self.editor.set_cursor_position('sof')
 
@@ -241,12 +248,11 @@ class PlainText(QWidget):
     def set_wrap_mode(self, value):
         self.editor.toggle_wrap_mode(value)
 
-    def set_wrap_action(self, action):
-        """
-        Add wrap action on editor.
-        """
-        self.editor.readonly_menu.addSeparator()
-        self.editor.readonly_menu.addAction(action)
+    def copy(self):
+        self.editor.copy()
+
+    def select_all(self):
+        self.editor.selectAll()
 
 
 class HelpWidget(PluginMainWidget):
@@ -351,6 +357,18 @@ class HelpWidget(PluginMainWidget):
             toggled=lambda value: self.set_option('wrap', value),
             initial=self.get_option('wrap'),
         )
+        self.copy_action = self.create_action(
+            name=HelpWidgetActions.CopyAction,
+            text=_("Copy"),
+            triggered=lambda value: self.plain_text.copy(),
+            register_shortcut=False,
+        )
+        self.select_all_action = self.create_action(
+            name=HelpWidgetActions.SelectAll,
+            text=_("Select All"),
+            triggered=lambda value: self.plain_text.select_all(),
+            register_shortcut=False,
+        )
         self.auto_import_action = self.create_action(
             name=HelpWidgetActions.ToggleAutomaticImport,
             text=_("Automatic import"),
@@ -382,9 +400,6 @@ class HelpWidget(PluginMainWidget):
             initial=self.get_option('locked'),
         )
 
-        # TODO: Temporal while code editor migrates to SpyderMixin
-        self.plain_text.set_wrap_action(self.wrap_action)
-
         # Add the help actions to an exclusive QActionGroup
         help_actions = QActionGroup(self)
         help_actions.setExclusive(True)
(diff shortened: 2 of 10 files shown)


ccordoba12:
> As an initial attempt to really decouple the different plugins the simplest way (by far) is to provide a simple code editor with the needed capabilities in the spyder.widgets module

Ok, I agree with this.

> Since this simple editor does not implement panels, no line numbers are available.

Can't you add line numbers to this simplified version of CodeEditor? We had a line number area in the old days of Spyder 2 without a panel, so that should be possible in this case too.

goanpeca:
>Can't you add line numbers to this simplified version of CodeEditor? We had a line number area in the old days of Spyder 2 without a panel, so that should be possible in this case too.

Yes it is possible. I will add one.

ccordoba12:
Thanks @goanpeca!

goanpeca:
This is ready for review. Failure seems unrelated to the work of this PR.

ccordoba12:
@goanpeca, please rebase on top of the latest master to fix the error with our tests.

spyder/plugins/appearance/confpage.py
```diff
@@ -321,27 +319,32 @@ def update_preview(self, index=None, scheme_name=None):
         """
         text = ('"""A string"""\n\n'
                 '# A comment\n\n'
-                '# %% A cell\n\n'
```
ccordoba12, line 324: Why are we not displaying cells in the preview anymore? That's a very important functionality, so I don't agree with removing them.

spyder/plugins/appearance/confpage.py
```diff
@@ -321,27 +319,32 @@ def update_preview(self, index=None, scheme_name=None):
...
-        self.preview_editor.set_text(text)
+
+        # FIXME: The appearance plugin is not tied to the code editor
+        # so these calls here probably should not be used, since it
+        # assumes there is an editor, which might not be the case at
+        # all. We can probably make a decission for the user and
+        # that should be good enough, after all this is just a quick
+        # preview.
+        show_blanks = CONF.get('editor', 'blank_spaces')
+        update_scrollbar = CONF.get('editor', 'scroll_past_end')
```
ccordoba12, line 338: Let's take those decisions now: please make `show_blanks` and `update_scrollbar` `False` in the preview.

spyder/plugins/help/widgets.py
```diff
@@ -349,6 +353,18 @@ def setup(self, options):
             toggled=lambda value: self.set_option('wrap', value),
             initial=self.get_option('wrap'),
         )
+        self.copy_action = self.create_action(
+            name="help_plain_text_copy_action",
```
ccordoba12, line 357: Why is this action not listed in an enum class, like in other plugins moved to the new API?

spyder/plugins/help/widgets.py
```diff
@@ -349,6 +353,18 @@ def setup(self, options):
...
+            text=_("Copy"),
+            triggered=lambda value: self.plain_text.copy(),
+            register_shortcut=False,
+        )
+        self.select_all_action = self.create_action(
+            name="help_plain_text_select_all_action",
```
ccordoba12, line 363: And this one?

spyder/plugins/variableexplorer/widgets/objectexplorer/objectexplorer.py
```diff
@@ -420,20 +420,28 @@ def _update_details_for_item(self, tree_item):
...
             self.editor.setPlainText(data)
             self.editor.setWordWrapMode(attr_details.line_wrap)
+            # FIXME: The variable explorer is not tied to the code editor
+            # so these calls here probably should not be used, since it
+            # assumes there is an editor, which might not be the case at
+            # all. We can probably make a decission for the user and
+            # that should be good. Having all code editors in spyder in
+            # sync does not seem very useful in this case.
             show_blanks = CONF.get('editor', 'blank_spaces')
             update_scrollbar = CONF.get('editor', 'scroll_past_end')
```
ccordoba12, line 430: Let's make these options False in this PR.

spyder/widgets/editor.py
```diff
@@ -0,0 +1,549 @@
...
+from qtpy.QtGui import QColor, QPainter, QTextCursor, QTextFormat, QTextOption
+from qtpy.QtWidgets import QApplication, QPlainTextEdit, QTextEdit, QWidget
+
+# Local imports
+import spyder.utils.syntaxhighlighters as sh
+from spyder.widgets.mixins import BaseEditMixin
+
+
+# Constants
+LANGUAGE_EXTENSIONS = {
```
ccordoba12, line 30: You're repeating here almost the same values present in the `EDIT_FILETYPES` dictionary, placed in `spyder/config/utils.py`. So please do a bit of refactoring in that file, to be able to use the same file extensions here.

spyder/widgets/editor.py
```diff
@@ -0,0 +1,549 @@
...
+    Adapted from:
+    https://doc.qt.io/qt-5/qtwidgets-widgets-codeeditor-example.html
+    """
+
+    def __init__(self, code_editor=None):
+        super().__init__(code_editor)
+
+        self._editor = code_editor
+        self._left_padding = 6  # Pixels
+        self._right_padding = 3  # Pixels
```
ccordoba12, line 61: `_left_padding` and `_right_padding` are used directly below to set them, so I don't think there's a need to make them start with and underscore `_`.

Also, is there a need to make `_editor` a private attribute?

spyder/widgets/editor.py
```diff
@@ -0,0 +1,549 @@
...
+        'Markdown': (sh.MarkdownSH, '#'),
+        # Every other language
+        'None': (sh.TextSH, ''),
+    }
+
+    # --- Signals
+    # ------------------------------------------------------------------------
+    sig_focus_changed = Signal()
+    """
+    """
```
ccordoba12, line 98: Missing docstring.

spyder/widgets/editor.py
```diff
@@ -0,0 +1,549 @@
...
+
+    def _set_palette(self, background, foreground):
+        style = ("QPlainTextEdit#%s {background: %s; color: %s;}" %
+                 (self.objectName(), background.name(), foreground.name()))
+        self.setStyleSheet(style)
+        self.rehighlight()
+
+    def _highlight_current_line(self):
+        if self._highlighter:
+            extraSelections = []
```
ccordoba12, line 142: Please don't use camel case completions. We don't use them anywhere else in our code.

spyder/widgets/editor.py
```diff
@@ -0,0 +1,549 @@
...
+            selection = QTextEdit.ExtraSelection()
+            lineColor = self._highlighter.get_currentline_color()
```
ccordoba12, line 144: Also remove camel case here.

Besides, this simple editor needs to support cell highlighting so we can display that in the Appearance entry in Preferences.

goanpeca:
>Looks good @goanpeca, thanks! I left some minor comments, but the most important thing is that this new SimpleCodeEditor needs to support cell highlighting so we can display it in Appearance. Also, perhaps in the future we could use cells in History too.

That is too much work for little gain. I do not agree.

spyder/plugins/appearance/confpage.py
goanpeca: It is not an important functionality for a simple code editor. Having cells appear in the appearance settings is completely secondary. I do not agree with implementing that.

spyder/plugins/appearance/confpage.py
goanpeca: 👍

spyder/plugins/variableexplorer/widgets/objectexplorer/objectexplorer.py
goanpeca: 👍

spyder/widgets/editor.py
goanpeca: >so I don't think there's a need to make them start with and underscore

If it is not meant to be exposed to the user it should have an underscore.

>Also, is there a need to make _editor a private attribute?

Cleanliness. The line number area accesses the editor internally so there is no reason why this should be exposed.

ccordoba12:
> That is too much work for little gain. I do not agree.

Cells is a very important functionality in Spyder (@isabela-pf remarked that in our last meeting with her). So we need to make it as visible as possible. Besides, as I told you, my plan is to use them in History too.

So please implement a simple version of them, or refactor and reuse what's necessary from `spyder/plugin/editor/widgets/base.py`.

goanpeca:
>Cells is a very important functionality in Spyder (@isabela-pf remarked that in our last meeting with her). So we need to make it as visible as possible. Besides, as I told you, my plan is to use them in History too.

I am not saying cells are not important. They are and there is a complex editor for that.

Using the complex editor in appearance is not an option as it results in a circular dependency.

Having a simple editor that is used in a lot of places for just syntax highlighting (I already agreed to add line numbers, cause I agree they are important) But cells are not for any of the current cases. Implementing cells (which is not important at the moment) is holding this PR and work that depends on it.

spyder/widgets/editor.py
ccordoba12: Ok

ccordoba12:
Ok, let's go without cells for now, then we'll what to do about them later.

> If Spyder was configured to not start with the Editor on (which is something the new API allows for), what does it mean to have cells appear on the Theme selection?

I disagree on not having cells in Appearance, but let's not delay this work for that. Also, this argument makes no sense to me: what is Spyder if it only displays the theme selection plugin?

goanpeca:
>Also, this argument makes no sense to me: what is Spyder if it only displays the theme selection plugin?

A glorified Github issue reporter :-), where the appearance actually changed how the Issue Reporter will look :-)

goanpeca:
>Ok, let's go without cells for now, we'll see what to do about them later.

Thanks.

Please let's try to iterate faster on these PRs.

It is a big refactor, it is dealing with a lot of technical debt and is aiming to reduce the growth of the technical debt moving forward.

Yes, some things might break in the process (this is an alpha after all).

Yes, it is annoying that things break, but things need to break to move forward and I am ready to fix any issues that I may have introduced, but the faster things get merged, the faster we can iterate, the faster issues can be found and solved and the faster things will fall into place.

spyder/widgets/editor.py
goanpeca: >So please do a bit of refactoring in that file, to be able to use the same file extensions here.

Well, it is not really the same and it would require doing some changes that really feel out of scope since currently the SimpleCodeEditor just depends on syntax highlighting and base. Now it will depend on some values on a utils file in a config module. 

This can be addressed in the Editor migration since all these values should probably live there anyway. I think these "config.utils" should be elsewhere so I prefer to address the changes in a holistic fashion he the work arrises.

spyder/plugins/appearance/confpage.py
goanpeca: Fixed.

spyder/plugins/help/widgets.py
goanpeca: Fixed.

spyder/plugins/help/widgets.py
goanpeca: Fixed.

spyder/plugins/variableexplorer/widgets/objectexplorer/objectexplorer.py
goanpeca: Fixed.

spyder/widgets/editor.py
goanpeca: Fixed.

spyder/widgets/editor.py
goanpeca: Fixed.

spyder/widgets/editor.py
goanpeca: Fixed.

isabela-pf:
It looks like this has been resolved before I got a chance to comment, but I wanted to respond to being mentioned. I do think cells are important, but I think it is fine that a simple code editor has limited functionality, especially when it seems to exist mainly to solve some technical debt issues. Cells are still functioning elsewhere, so I think it's good to focus on solving the problems and working up to add things back in as they are a priority.

goanpeca:
Fixing issues.

goanpeca:
Ready for review @ccordoba12.

spyder/config/utils.py
```diff
@@ -50,6 +50,28 @@
     (_("Markdown files"), ('.md', )),
 ]
 
+
+LANGUAGE_EXTENSIONS = {
```
ccordoba12, line 54: If you're not planning to use this dictionary in the SimpleCodeEditor, please remove it.

Else, please add a TODO comment above it to understand why you're putting it here and not using it.

spyder/widgets/editor.py
```diff
@@ -0,0 +1,551 @@
+# -*- coding: utf-8 -*-
```
ccordoba12, line 1: Please rename this file to `simplecodeeditor.py`. `editor.py` is very vague and could lead to confusions later.

ccordoba12:
> Cells are still functioning elsewhere, so I think it's good to focus on solving the problems and working up to add things back in as they are a priority.

Thanks @isabela-pf for your feedback!

spyder/widgets/editor.py
goanpeca: Fixed.

spyder/config/utils.py
goanpeca: Forgot to remove it. Fixed.  (tracking this TODO on a separate issue already)

ccordoba12:
This PR needs more work. There are still several issues that can be easily seen at plain sight (e.g. the font not being the same as the one used in the Editor).

@steff456 and @andfoy will manually review it and give you more concrete feedback.

goanpeca:
Thanks for the feedback @ccordoba12.

Looking forward to fixing any issues that @steff456 and @andfoy find!

steff456:
Hi @goanpeca,
This are the comments we found in this PR

# History plugin
The first one is a screenshot of the editor and the second is a screenshot of the history. As you can see the font doesn't match.
image

image

As well, we found that the debug bar is missing but we are not sure of this is intended or not, the first screenshot is from 4.x and the second one from this PR

image

image

# Help plugin
In here we noted that now the numbering of the editor is present, while in 4.x is not.
image

image

Also, before we didn't have any highlight on the help and now we have. 

image

image

# Appearance Dialog
At first sight we can see that you removed the `Run` config page, and I think that needs to be back. Also the order was changed.

image

image

In the editor with the preview of the theme, we can see that there's no longer a cell which we believe is important to show that it has a different color than the rest. Also we found weird that even by writing a cell, the option is not activated. Additionally, the right vertical pane with the minimap is not shown. And, in this pane is happening exactly the same that in the history pane where the vertical pane of debug is not shown. All this things can be seen in the following screenshots.

image

image

Also, we noted that by changing the theme the preview in editor is not shown, for example in this screenshot we should be seeing the preview of Pydev but we are still seeing Spyder Dark. Exactly the same behavior should be happening with the `edit selected scheme`.

image

# Error report
In this pane, in 4.x we had a vertical pane on the left side of the editor and now it is gone, we are not sure if that was intentional or not. As well, the highlight is present now, but it wasn't present before.

image

image

This were the things that we found on the appearance side that can be improved for this PR

goanpeca:
Thanks for the feedback @steff456 

# History plugin
>The first one is a screenshot of the editor and the second is a screenshot of the history. As you can see the font doesn't match.

Fixed.

>As well, we found that the debug panel is missing but we are not sure of this is intended or not, the first screenshot is from 4.x and the second one from this PR

That is expected. The SimpleCodeEditor only has line numbers, so not having a debug panel is normal. Also, there is no "debug"  functionality on the History plugin.

# Help plugin

>In here we noted that now the numbering of the editor is present, while in 4.x is not.

Fixed.

>Also, before we didn't have any highlight on the help and now we have. 

Fixed.

# Appearance Dialog

>At first sight we can see that you removed the `Run` config page, and I think that needs to be back. Also the order was changed.

That has been fixed already in master.

>In the editor with the preview of the theme, we can see that there's no longer a cell which we believe is important to show that it has a different color than the rest. Also we found weird that even by writing a cell, the option is not activated. 

This has been already addressed in the thread and will not be changed. The appearance plugin would need to depend on the editor plugin which creates a circular dependency. 

>Additionally, the right vertical pane with the minimap is not shown. And, in this pane is happening exactly the same that in the history pane where the vertical pane of debug is not shown. All this things can be seen in the following screenshots.

That is also expected as the Appearance configuration page is not using the CodeEditor from the Editor plugin, but a SimpleCodeEditor (what this or created) which only provides, syntax highlighting and the line numbers panel.

>Also, we noted that by changing the theme the preview in editor is not shown, for example in this screenshot we should be seeing the preview of Pydev but we are still seeing Spyder Dark. Exactly the same behavior should be happening with the `edit selected scheme`.

Fixed.

# Error report

>In this pane, in 4.x we had a vertical pane on the left side of the editor and now it is gone, we are not sure if that was intentional or not.

Yes, it is expected as this editor only provides a Line number area.

> As well, the highlight is present now, but it wasn't present before.

Fixed.

goanpeca:
I already provided the fixes requested on this PR review.

dalthviz:
@andfoy @steff456 could you help us reviewing again this one to validate the fixes? Thanks!

PR #13542, opened 2020-08-15
PR: Create new MainMenu plugin

goanpeca:
## Description of Changes

## This PR implements

- A new `MainMenu` plugin in charge of managing the application menu bar, menus, and menu sections.
- Plugins that need to add menus or add items (Actions or submenus) to the main application menu bar will require this plugin.
- The API of menus now lives in this plugin instead of the global `spyder.api` since the plugin is in charge of those specifics.

### Affirmation

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -99,12 +99,9 @@ from spyder.utils.misc import (select_port, getcwd_or_home,
                                get_python_executable)
 from spyder.utils.programs import is_module_installed
 from spyder.utils.qthelpers import (create_action, add_actions, get_icon,
-                                    add_shortcut_to_tooltip,
-                                    create_module_bookmark_actions,
                                     create_program_action, DialogManager,
                                     create_python_script_action, file_uri,
-                                    MENU_SEPARATOR, qapplication,
-                                    set_menu_icons)
+                                    MENU_SEPARATOR, qapplication, start_file)
 from spyder.otherplugins import get_spyderplugins_mods
 from spyder.app import tour
 from spyder.app.solver import find_external_plugins, solve_plugin_dependencies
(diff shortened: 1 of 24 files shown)


goanpeca:
This one is ready for review @ccordoba12.

goanpeca:
This one is ready for review and has not received feedback since August 15 @ccordoba12

spyder/plugins/help/plugin.py
```diff
@@ -21,18 +21,23 @@
...
+class HelpActions:
+    ShowSpyderTutorial = "spyder_tutorial_action"
+
+
 class Help(SpyderDockablePlugin):
     """
     Docstrings viewer widget.
     """
     NAME = 'help'
     REQUIRES = [Plugins.Console, Plugins.Editor]
```
goanpeca, line 39: Since it requires one or the other or both maybe they could also be OPTIONAL, and check that at least onf them exists (on the check_compatibilty func?)

spyder/plugins/mainmenu/plugin.py
```diff
@@ -0,0 +1,251 @@
...
+from spyder.plugins.mainmenu.api import ApplicationMenus, HelpMenuSections
+from spyder.utils.qthelpers import set_menu_icons
+
+# Localization
+_ = get_translation('spyder')
+
+
+class MainMenu(SpyderPluginV2):
+    NAME = 'mainmenu'
+    CONF_SECTION = 'mainmenu'
```
goanpeca, line 33: Could use

```
    NAME = 'mainmenu'
    CONF_SECTION = NAME
``` 

To keep them in sync. Ideally the `CONF_SECTION
` class attribute should go away.

spyder/plugins/pylint/plugin.py
```diff
@@ -34,7 +34,7 @@ class Pylint(SpyderDockablePlugin):
     WIDGET_CLASS = PylintWidget
     CONF_SECTION = NAME
     CONF_WIDGET_CLASS = PylintConfigPage
-    REQUIRES = [Plugins.Editor]
+    REQUIRES = [Plugins.Editor, Plugins.MainMenu]
```
goanpeca, line 37: Techincally the Pylint plugin could run a file without an editor (with the select file option of the combobox) so this should probably become also `OPTIONAL`

spyder/app/tests/test_mainwindow.py
```diff
@@ -2117,9 +2117,11 @@ def test_help_opens_when_show_tutorial_full(main_window, qtbot):
     'Show tutorial' opens the help plugin if closed.
     """
     HELP_STR = "Help"
-    from spyder.plugins.help.plugin import HelpActions
-    help_pane_menuitem = main_window.help.get_action(
-        HelpActions.ShowSpyderTutorialAction)
+    help_pane_menuitem = None
```
goanpeca, line 2120: How does this work with `localization` `_("")`?

dalthviz: I think this test was thought to work using the default language to run (I guess English?)

goanpeca: 🤦🏻‍♂️missed that this was test file.

dalthviz:
Pinging @ccordoba12 just in case

spyder/app/mainwindow.py
```diff
@@ -1466,7 +1466,7 @@ def update_lsp_logs(self):
     def post_visible_setup(self):
         """Actions to be performed only after the main window's `show` method
         was triggered"""
-        for _plugin_id, plugin in self._PLUGINS.items():
+        for __, plugin in self._PLUGINS.items():
```
goanpeca, line 1469: `_plugin_id` is more readable than `__`

ccordoba12: Thanks @goanpeca for the suggestion, but we're adopting the convention of using `__` for unnecessary variables in loops. Using a name for those variables, even if it starts with an underscore, gives the impression that it's required and somehow missing in the code below.

Actually, that was a small source of confusion for @dalthviz and I when we were reviewing this PR, so I suggested him to do this change.

ccordoba12 (approved):
Thanks @goanpeca and @dalthviz for your work!

PR #13546, opened 2020-08-15
WIP: Enh/api console cleanup

goanpeca:
## Description of Changes

### Issue(s) Resolved

Fixes #

### Affirmation


ccordoba12:
Although I understand this is a WIP, I want to give you my early feedback to see if it's justified to keep working on it:

* The Console plugin has two simple purposes:
    * Capture internal errors.
    * Have an Python interpreter that allow us to introspect Spyder within Spyder.

* With that in mind, I don't see much sense on moving the actions and functionality you're thinking to move to it on this PR. In other words, I don't see a relation between them and the console, so I think they should stay where they are.

* Since this is not critical, I'd suggest you to focus on other pending PRs and leave this one for later.

* You didn't consult with me or others in the core team about your decision to do this. Otherwise, I would have told you that I don't see a refactoring like this relevant nor important.

goanpeca:
Thanks for the feedback @ccordoba12.

In the new API model the concept of adding standalone menus and actions from the `mainwindow` does not apply. The idea is that menu items, actions, status bar widgets, and any action available in Spyder, must be provided by a plugin. This ensures that there are no "special cases"  in how things are done and unifies the codebase in Spyder. More information [here](https://spyder-ide.github.io/spyder-api-docs/new_api.html#the-guiding-principles)

There are several actions that are "generic" and are being added in the mainwindow.
- Show dependencies
- Show about
- Report errors

To be able to make all the API uniform and consistent, these actions need to be moved to some plugin.

Error reporting is handled by the Console and the `errorreport` widget (currently living on `spyder.widgets`) is only used by the console and globally by the MainWindow. The idea is to assign the error reporting to the plugin in charge of that. 

Since the Console plugin already handles the internal errors, it is natural to move the report dialog to the console plugin and provide the `Report issue` action from this plugin.

Leaving the actions on the mainwindow is not an option as it will create a "special case" of handling things, not following the principles outlined in the [API docs](https://spyder-ide.github.io/spyder-api-docs/new_api.html#the-guiding-principles).

If adding these actions (mostly populating the help menu) do not fit with the console then the other option is to create a different plugin to provide this. Some options include:

- Console plugin: Current approach
- MainMenu plugin:
- SomeNew Plugin: 

Please let me know @spyder-ide/core-developers which of the options you consider the best option.

goanpeca:
@ccordoba12 I am still waiting for feedback on the options.

I reiterate, not moving actions, menu creation from the mainwindow to a specific plugin (whatever of the options is decided) defeats the purpose of the new API, and would leave a mixture of things.

ccordoba12:
> MainMenu plugin: The plugin in charge of creating and updating the main application menubar, menus and menu sections.

Please move them here then.

goanpeca:
Do you agree that the error report dialog widget should live within the console?

The console already handles internals errors and the main window is the only other user.

ccordoba12:
> Do you agree that the error report dialog widget should live within the console?

No, please leave it where it is.

goanpeca:
Most of the API work has also aimed to reduce the global widgets found in `spyder.widgets` as that makes them non-extendable.

Some other things that will require moving to become a plugin include.
- Tours. For organization reasons
- Find/Replace. A plugin that can/wants provide Find and File would `REQUIRE` it and then register with it. 
- Array builder. A plugin that can/wants to provide an array builder  would `REQUIRE` it and then register with it. 

>No, please leave it where it is.

Could you please elaborate on the rationale?

goanpeca:
>MainMenu plugin: The plugin in charge of creating and updating the main application menubar, menus and menu sections.

Ok I will move them there but on a separate PR. https://github.com/spyder-ide/spyder/pull/13542 is already ready for review.

ccordoba12:
> Tours. For organization reasons

I don't agree with refactoring for the sake of refactoring. If there's no strong reason to do that, please leave where it is.

> Could you please elaborate on the rationale?

I see no need to move it to the Console because it's also used to report issues with no associated tracebacks.

goanpeca:
## 4.x dependency graph

image


- **Red lines**: represent hard dependencies, if the plugin is not available Spyder fails to start.
- **Purple lines**: represent circular dependencies, plugins that depend on each other.
- **Green nodes**: represent pseudo plugins, using special cases handling.
- **Blue nodes**: represent a diferent type of plugin, not following the API.
- **Black nodes**: Current plugins.

## 5.x dependency graph
image

This includes work already merged plus suggested ideas (not including the `AppPlugin`)

- **Red lines**: represent hard dependencies, if the plugin is not available the plugin will not load (but spyder can start).
- **Purple lines**: There are no circular dependencies.
- **Green nodes**: represent NEW plugins, which come from the pseudo plugins or splitting functionality living in the mainwindow.
- **Blue dashed lines**: represent optional dependencies. Some functionality of the loaded plugin will not be available, but Spyder still starts.

PR #13547, opened 2020-08-16
PR: Add new toolbar plugin

goanpeca:
# Description of changes

- A new ToolBar plugin in charge of managing the application toolbars, and toolbar sections sections.
- Plugins that need to add a toolbar or add items (Actions/Widgets) to the main application toolbar area bar will require this plugin.
- The API of toolbars now lives in this plugin instead of the global `spyder.api` since the plugin is in charge of those specifics.

### Affirmation

### DIFF
--- a/spyder/api/widgets/__init__.py
+++ b/spyder/api/widgets/__init__.py
@@ -39,7 +39,7 @@ from spyder.api.widgets.auxiliary_widgets import (MainCornerWidget,
                                                   SpyderWindowWidget)
 from spyder.api.widgets.menus import (MainWidgetMenu, OptionsMenuSections,
                                       PluginMainWidgetMenus, SpyderMenu)
-from spyder.api.widgets.mixins import SpyderToolBarMixin, SpyderWidgetMixin
+from spyder.api.widgets.mixins import SpyderToolbarMixin, SpyderWidgetMixin
 from spyder.api.widgets.toolbars import MainWidgetToolbar
 from spyder.config.gui import is_dark_interface
 from spyder.utils.qthelpers import (add_actions, create_waitspinner,
@@ -68,7 +68,7 @@ class PluginMainWidgetActions:
 
 # --- Spyder Widgets
 # ----------------------------------------------------------------------------
-class PluginMainContainer(QWidget, SpyderWidgetMixin, SpyderToolBarMixin):
+class PluginMainContainer(QWidget, SpyderWidgetMixin, SpyderToolbarMixin):
     """
     Spyder plugin main container class.
 
@@ -163,6 +163,16 @@ class PluginMainContainer(QWidget, SpyderWidgetMixin, SpyderToolBarMixin):
         self._plugin = plugin
         self._parent = parent
 
+        # Widget setup
+        # A PluginMainContainer inherits from QWidget so it can be a parent
+        # for the widgets it contains. Since it is a QWidget it will occupy a
+        # physical space on the screen and may cast "shadow" on the top left
+        # of the main window. To prevent this we ensure the widget has zero
+        # width and zero height.
+        # See: spyder-ide/spyder#13547
+        self.setMaximumWidth(0)
+        self.setMaximumHeight(0)
+
     # --- API: methods to define or override
     # ------------------------------------------------------------------------
     def setup(self, options=DEFAULT_OPTIONS):
@@ -192,7 +202,7 @@ class PluginMainContainer(QWidget, SpyderWidgetMixin, SpyderToolBarMixin):
             'method!')
 
 
-class PluginMainWidget(QWidget, SpyderWidgetMixin, SpyderToolBarMixin):
+class PluginMainWidget(QWidget, SpyderWidgetMixin, SpyderToolbarMixin):
     """
     Spyder plugin main widget class.
 
@@ -434,7 +444,7 @@ class PluginMainWidget(QWidget, SpyderWidgetMixin, SpyderToolBarMixin):
                 child.setCornerWidget(self._corner_widget)
 
                 # This is needed to ensure the corner ToolButton (hamburguer
-                # menu) is aligned with plugins that use ToolBars vs
+                # menu) is aligned with plugins that use Toolbars vs
                 # CornerWidgets
                 # See: spyder-ide/spyder#13600
                 # left, top, right, bottom
--- a/spyder/api/widgets/toolbars.py
+++ b/spyder/api/widgets/toolbars.py
@@ -19,12 +19,13 @@ from qtpy.QtWidgets import (QAction, QSizePolicy, QToolBar, QToolButton,
                             QWidget)
 
 # Local imports
+from spyder.api.exceptions import SpyderAPIError
 from spyder.config.gui import is_dark_interface
 
 
 # --- Constants
 # ----------------------------------------------------------------------------
-class ApplicationToolBars:
+class ApplicationToolbars:
     File = 'file_toolbar'
     Run = 'run_toolbar'
     Debug = 'defbug_toolbar'
@@ -35,7 +36,7 @@ class ApplicationToolBars:
     WorkingDirectory = 'working_directory_toolbar'
 
 
-class ToolBarLocation:
+class ToolbarLocation:
     Top = Qt.TopToolBarArea
     Bottom = Qt.BottomToolBarArea
 
@@ -59,9 +60,9 @@ class ToolTipFilter(QObject):
 
 # --- Widgets
 # ----------------------------------------------------------------------------
-class SpyderToolBar(QToolBar):
+class SpyderToolbar(QToolBar):
     """
-    Spyder ToolBar.
+    Spyder Toolbar.
 
     This class provides toolbars with some predefined functionality.
     """
@@ -70,16 +71,18 @@ class SpyderToolBar(QToolBar):
         super().__init__(parent=parent)
         self._section_items = OrderedDict()
         self._title = title
+        self._default_section = "default_section"
 
         self.setWindowTitle(title)
 
     def add_item(self, action_or_widget, section=None, before=None):
         """
-        Add action or widget item to given toolbar `section`. 
+        Add action or widget item to given toolbar `section`.
         """
-        if section is not None:
-            action_or_widget._section = section
+        if section is None:
+            section = self._default_section
 
+        action_or_widget._section = section
         if section is None and before is not None:
             action_or_widget._section = before._section
             section = before._section
@@ -126,13 +129,24 @@ class SpyderToolBar(QToolBar):
                     widget.setCheckable(True)
 
 
-class ApplicationToolBar(SpyderToolBar):
+class ApplicationToolbar(SpyderToolbar):
     """
-    Spyder Main application ToolBar.
+    Spyder Main application Toolbar.
     """
 
+    ID = None
+    """
+    Unique string toolbar identifier.
 
-class MainWidgetToolbar(SpyderToolBar):
+    This is used by Qt to be able to save and restore the state of widgets.
+    """
+
+    def _check_interface(self):
+        if self.ID is None:
+            raise SpyderAPIError("Toolbar must define an ID attribute.")
+
+
+class MainWidgetToolbar(SpyderToolbar):
     """
     Spyder Widget toolbar class.
 
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -288,12 +288,6 @@ class MainWindow(QMainWindow):
             self.register_shortcut(sc, context, name)
             self.register_shortcut(plugin.toggle_view_action, context, name)
 
-        toolbars = plugin.get_registered_application_toolbars()
-        for __, toolbar in toolbars.items():
-            # TODO: To update this render call
-            toolbar._render()
-            self.toolbarslist.append(toolbar)
-
     def unregister_plugin(self, plugin):
         """
         Unregister a plugin from the Spyder Main Window.
@@ -584,7 +578,6 @@ class MainWindow(QMainWindow):
 
         # Actions
         self.lock_interface_action = None
-        self.show_toolbars_action = None
         self.close_dockwidget_action = None
         self.undo_action = None
         self.redo_action = None
@@ -617,7 +610,6 @@ class MainWindow(QMainWindow):
 
         self.plugins_menu = None
         self.plugins_menu_actions = []
-        self.toolbars_menu = None
 
         # Status bar widgets
         self.interpreter_status = None
@@ -625,9 +617,7 @@ class MainWindow(QMainWindow):
         self.cpu_status = None
         self.clock_status = None
 
-        # Toolbars
-        self.visible_toolbars = []
-        self.toolbarslist = []
+        # TODO: Move to corresponding Plugins
         self.main_toolbar = None
         self.main_toolbar_actions = []
         self.file_toolbar = None
@@ -642,8 +632,6 @@ class MainWindow(QMainWindow):
         self.run_toolbar_actions = []
         self.debug_toolbar = None
         self.debug_toolbar_actions = []
-        self.layout_toolbar = None
-        self.layout_toolbar_actions = []
 
         self.menus = []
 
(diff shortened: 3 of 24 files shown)


ccordoba12:
@juanis2112, @steff456, please give a manual review to this PR.

juanis2112:
Thanks @goanpeca!

On my manual review I found out that everything looks/works fine for me except for one thing:
There is something weird going on with the hovering over the first three buttons (from left to right). It seems like hovering state is only activated when the cursor is not on top of the icon. If I put my cursor on the center of the button, it doesn't seem to change to the hover state and it doesn't work when pressing the button, however putting it closer to the edges works. Here is a gif that hopefully will make this clearer:

bug_1

(This only happens locally, I tested it on Binder and it looks fine)

steff456:
HI @goanpeca, I'm having the same behavior than Juanita and also as you can see in her gif, there's no tooltip on those buttons.

goanpeca:
>It seems like hovering state is only activated when the cursor is not on top of the icon. If I put my cursor on the center of the button, it doesn't seem to change to the hover state and it doesn't work when pressing the button, however putting it closer to the edges works.

Thanks for the feedback @juanis2112 and @steff456 , will take a look.

juanis2112:
@steff456 I do see a tooltip but I think it takes a while to load so that's why it is not shown in the gif. But that might be something to check too @goanpeca cause it might be taking longer than usual.

goanpeca:
Hi @juanis2112, there was indeed an issue. I believe is now solved, could you and @steff456 test again?

Thanks!

hover

In case the gif is not displaying the tooltips.
image

goanpeca:
A suggestion for future PR reviews is to take a look at the Screen Shot 2020-08-30 at 14 09 41 section so that you can provide the review using the `Review changes` buttons

Screen Shot 2020-08-30 at 14 08 39

In here you can either:
- `Comment` if you need to have more information have questions on something on the Pull Request to either Approve or Request further changes.
- `Approve`  if all looks fine and you agree with the current state of the Pull Request
- `Request changes` if this specific Pull Request introduced an issue or regression. If the problem is reproducible in 4.x or/and master then opening a separate new issue for the problems found may be a better choice, since the changes on the Pull request did not introduce the problem.

By doing that the Reviewer state 
Screen Shot 2020-08-30 at 14 14 23
will go from pending (yellow dot) to reviewed and waiting for answers or changes (double arrow icon).

And the cycle can then be repeated as requested changes are made from the PR author.

Also by doing this, when a PR is finally approved by 1 or more developers, the PR state can go from 
Screen Shot 2020-08-30 at 14 15 45

to ready for review.

Cheers!

spyder/app/tests/test_mainwindow.py
```diff
@@ -862,6 +862,7 @@ def test_connection_to_external_kernel(main_window, qtbot):
     kc.stop_channels()
 
 
+@pytest.mark.first
```
ccordoba12, line 865: Please remove this if it's not really necessary (why your changes affect this test?). We should avoid moving test to be run first whenever possible.

spyder/config/main.py
```diff
@@ -94,6 +94,7 @@
               'cursor/width': 2,
               'completion/size': (300, 180),
               'report_error/remember_token': False,
+              'last_visible_toolbars': [],
```
ccordoba12, line 97: If ToolBar is a plugin now, it deserves it's own configuration section. So why are you adding this option to `main`?

spyder/plugins/toolbar/container.py
```diff
@@ -0,0 +1,322 @@
...
+
+        self._APPLICATION_TOOLBARS = OrderedDict()
+        self._ADDED_TOOLBARS = OrderedDict()
+        self._toolbarslist = []
+        self._visible_toolbars = []
+
+    # --- Private Methods
+    # ------------------------------------------------------------------------
+    def _save_visible_toolbars(self):
+        """Saves the name of the visible toolbars in the options."""
```
ccordoba12, line 62: Docstring should start in infinitive: Saves -> Save

spyder/plugins/toolbar/container.py
```diff
@@ -0,0 +1,322 @@
...
+        toolbars = []
+        for toolbar in self._visible_toolbars:
+            toolbars.append(toolbar.objectName())
+
+        # TODO: Move this to a different section since there is no concept
+        # of a MainPlugin (Unless it is created)
+        self.set_option('last_visible_toolbars', toolbars)
+
+    def _get_visible_toolbars(self):
+        """Collects the visible toolbars."""
```
ccordoba12, line 72: Collects -> Collect

spyder/plugins/toolbar/container.py
```diff
@@ -0,0 +1,322 @@
...
+        toolbars = []
+        for toolbar in self._toolbarslist:
+            if toolbar.toggleViewAction().isChecked():
+                toolbars.append(toolbar)
+
+        self._visible_toolbars = toolbars
+
+    @Slot()
+    def _show_toolbars(self):
+        """Show/Hides toolbars."""
```
ccordoba12, line 82: Hides -> Hide

spyder/plugins/toolbar/container.py
```diff
@@ -0,0 +1,322 @@
...
+    # --- PluginMainContainer API
+    # ------------------------------------------------------------------------
+    def setup(self, options=DEFAULT_OPTIONS):
+        self.show_toolbars_action = self.create_action(
+            ToolBarActions.ShowToolBars,
+            text=_("Show toolbars"),
+            triggered=self._show_toolbars,
+            context=Qt.ApplicationShortcut,
+            shortcut_context="_",
+            register_shortcut=True
```
ccordoba12, line 107: Did this action have an associated shortcut before? If not, please remember what we agreed on, i.e. not adding new shortcuts if they are not present in `4.x`.

spyder/plugins/toolbar/container.py
```diff
@@ -0,0 +1,322 @@
...
+            tip = _("Show toolbars")
+
+        self.show_toolbars_action.setText(text)
+        self.show_toolbars_action.setToolTip(tip)
+
+    # --- Public API
+    # ------------------------------------------------------------------------
+    def create_application_toolbar(self, toolbar_id, title):
+        """
+        Create a Spyder application toolbar and add it to the main window.
```
ccordoba12, line 133: Please remove `Spyder` from here:

> Create an application toolbar ...

spyder/api/widgets/toolbars.py
```diff
@@ -129,6 +132,11 @@ class ApplicationToolBar(SpyderToolBar):
     """
     Spyder Main application ToolBar.
     """
+    ID = None
```
ccordoba12, line 135: Could we change this to `name` instead? It seems to me it corresponds to the toolbar name, right?

That way it'd be easier to remember this parameter, instead of `ID`.

spyder/plugins/toolbar/container.py
```diff
@@ -0,0 +1,322 @@
...
+        self._ADDED_TOOLBARS[toolbar_id] = toolbar
+        self._toolbarslist.append(toolbar)
+
+        if mainwindow:
+            mainwindow.addToolBar(toolbar)
+
+    def add_item_to_application_toolbar(self, item, toolbar, toolbar_id=None,
+                                        section=None, before=None):
+        """
+        Add action or widget `item` to given application menu `section`.
```
ccordoba12, line 191: Is `menu` correct in this docstring?

spyder/plugins/toolbar/container.py
```diff
@@ -0,0 +1,322 @@
...
+        toolbar_id: str or None
+            The application toolbar unique string identifier.
+        section: str or None
+            The section id in which to insert the `item` on the `toolbar`.
+        before: str or None
+            Make the item appear before another given item.
+
+        Notes
+        -----
+        Must provide a `toolbar` or a `toolbar_id`.
```
ccordoba12, line 208: This is incorrect. `toolbar` is a mandatory kwarg whereas `toolbar_id` is optional.

spyder/plugins/toolbar/container.py
```diff
@@ -0,0 +1,322 @@
...
+        """
+        if toolbar and toolbar_id:
+            raise SpyderAPIError('Must provide only toolbar or toolbar_id!')
```
ccordoba12, line 211: To make this work, `toolbar` should be optional too in the signature of this method.

spyder/plugins/toolbar/container.py
```diff
@@ -0,0 +1,322 @@
...
+
+        Returns
+        -------
+        list
+            The list of all the added application toolbars.
+        """
+        return self._toolbarslist
+
+    def load_last_visible_toolbars(self):
+        """Loads the last visible toolbars from the options."""
```
ccordoba12, line 267: Please change this to

    Load the last visible toolbars from our preferences.

spyder/plugins/toolbar/container.py
```diff
@@ -0,0 +1,322 @@
...
+        # TODO: Move this to a different section since there is no concept
+        # of a MainPlugin (Unless it is created)
```
ccordoba12, line 269: Please solve this in this PR (I already left a comment about it above).

spyder/plugins/toolbar/container.py
```diff
@@ -0,0 +1,322 @@
...
+        toolbars_names = self.get_option('last_visible_toolbars')
+
+        if toolbars_names:
+            dic = {}
```
ccordoba12, line 273: Ugly variable name. Please use a better name (`toolbars_dict` would work).

spyder/plugins/toolbar/plugin.py
```diff
@@ -0,0 +1,189 @@
...
+    Docstrings viewer widget.
+    """
+    NAME = 'toolbar'
+    CONF_SECTION = NAME
+    CONF_FILE = False
+    CONTAINER_CLASS = ToolBarContainer
+    CONF_FROM_OPTIONS = {
+        'last_visible_toolbars': ('main', 'last_visible_toolbars'),
+        'toolbars_visible': ('main', 'toolbars_visible'),
+    }
```
ccordoba12, line 33: Please don't use options from the `main` section here (as I already mentioned a couple of times). It's better to add a new `toolbar` section to our preferences instead.

I understand that would break the user selected toolbar when moving from Spyder 4 to 5, but in this case I think it's a small price to pay to have an easier to follow code.

spyder/plugins/toolbar/plugin.py
```diff
@@ -0,0 +1,189 @@
...
+
+    # --- SpyderDocakblePlugin API
+    #  -----------------------------------------------------------------------
+    def get_name(self):
+        return _('Toolbar')
+
+    def get_description(self):
+        return _('Application Toolbars management.')
```
ccordoba12, line 41: Toolbars -> toolbars

steff456:
\show binder

steff456:
Hi @goanpeca, that issue was fixed in the last commit!

spyder/plugins/toolbar/plugin.py
goanpeca: Fixed.

spyder/plugins/toolbar/plugin.py
goanpeca: Fixed.

spyder/plugins/toolbar/container.py
goanpeca: Fixed.

spyder/plugins/toolbar/container.py
goanpeca: Fixed.

spyder/plugins/toolbar/container.py
goanpeca: Fixed.

spyder/plugins/toolbar/container.py
goanpeca: Fixed.

spyder/plugins/toolbar/container.py
goanpeca: Fixed.

spyder/plugins/toolbar/container.py
goanpeca: Fixed.

spyder/api/widgets/toolbars.py
goanpeca: >Could we change this to name instead? It seems to me it corresponds to the toolbar name, right?

It corresponds to a unique identifier that is not exposed to the user and Qt uses to save and restore the state of the main window. I used the name ID to be in life with the `id()`  method in python. Even if Qt uses `setObjectName` as the method, the documentation explains that it needs to be unique, so uniqueness is the key property this attribute needs to provide.

Screen Shot 2020-08-31 at 20 17 34


>That way it'd be easier to remember this parameter, instead of ID.

Added a docstring for the attribute to further explain what it does.

There is a check for the interface right below, and the error is explaining what a developer would need to do. If it was not set, or there was a typo an error would be raised explaining to the developer what needs to be changed.

This is going to be documented as part of the API.
```python
    def _check_interface(self):
        if self.ID is None:
            raise SpyderAPIError("Toolbar must define an ID attribute.")
```

If `ID` is not clear then `IDENTIFIER` is another option.

The projects API already uses this (and performs similar checks.
https://github.com/spyder-ide/spyder/blob/master/spyder/plugins/projects/api.py#L39

I also suggest the `NAME` attribute of plugin API be changed to `ID`, or `IDENTIFIER` since they provide a unique string. It is not really a name as that is provided by the `get_name` method.

spyder/plugins/toolbar/container.py
goanpeca: Fixed.

spyder/plugins/toolbar/container.py
goanpeca: >Did this action have an associated shortcut before? 

Yes.

Screen Shot 2020-08-31 at 20 05 56

spyder/plugins/toolbar/container.py
goanpeca: Fixed.

spyder/plugins/toolbar/container.py
goanpeca: Fixed.

spyder/plugins/toolbar/container.py
goanpeca: Fixed.

spyder/config/main.py
goanpeca: Fixed.

spyder/app/tests/test_mainwindow.py
goanpeca: >Please remove this if it's not really necessary (why your changes affect this test?).

The changes do not affect the tests directly, but somehow the order in which pytest runs them causes segfaults, that I cannot reproduce locally.

>We should avoid moving test to be run first whenever possible.

This PR will not pass unless that change is made.

spyder/app/mainwindow.py
```diff
@@ -1066,6 +1014,49 @@ def create_edit_action(text, tr_text, icon):
...
+            "internals with the following commands:\n"
+            "  spy.app, spy.window, dir(spy)\n\n"
+            "Please don't use it to run your code\n\n"
+        )
+        CONF.set('internal_console', 'message', message)
+        CONF.set('internal_console', 'multithreaded', self.multithreaded)
+        CONF.set('internal_console', 'profile', self.profile)
+        CONF.set('internal_console', 'commands', [])
+        CONF.set('internal_console', 'namespace', {})
+        CONF.set('internal_console', 'show_internal_errors', True)
```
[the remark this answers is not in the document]
goanpeca: Fixed.

spyder/api/widgets/__init__.py
```diff
@@ -162,6 +162,12 @@ def __init__(self, name, plugin, parent=None, options=DEFAULT_OPTIONS):
         self._plugin = plugin
         self._parent = parent
 
+        # Widget setup
+        # Ensure the containers if visible, do not occupy any visible space
+        # See: spyder-ide/spyder#13547
```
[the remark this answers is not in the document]
goanpeca: >I don't understand this comment.

Expanded the comment.

>Does this solve the problem reported by @juanis2112?

Yes, it does.

ccordoba12 (approved):
Thanks @goanpeca and @steff456 for your help with this one.

Issue #13481, opened 2020-08-05
Can't open Dict in variable explorer with mixed data types as keys.


catcallofcthulhu:
### What steps will reproduce the problem?

1: Create a dictionary with a mixture of strings and numeric values as keys.
eg.
colors = {1: 'red','Y': 'yellow'}

2: Attempt to open the dictionary in the variable explorer.

3: no profit.

EDIT:
Using this function:
```
def mixs(self, num):
    try:
        ele = int(num)
        return (0, ele, '')
    except ValueError:
        return (1, num, '')
```
 as a key for the sorted() here:
```
File "C:\Users\ifeed\anaconda3\envs\spyder\lib\site-packages\spyder\plugins\variableexplorer\widgets\collectionseditor.py", line 173, in set_data
    self.keys = sorted(list(data.keys()))
```
fixes the problem, but probably not in the best way.

### Traceback
```python-traceback
  File "C:\Users\ifeed\anaconda3\envs\spyder\lib\site-packages\spyder\plugins\variableexplorer\widgets\collectionsdelegate.py", line 160, in createEditor
    readonly=readonly)
  File "C:\Users\ifeed\anaconda3\envs\spyder\lib\site-packages\spyder\plugins\variableexplorer\widgets\collectionseditor.py", line 1310, in setup
    remote=remote)
  File "C:\Users\ifeed\anaconda3\envs\spyder\lib\site-packages\spyder\plugins\variableexplorer\widgets\collectionseditor.py", line 1250, in __init__
    title)
  File "C:\Users\ifeed\anaconda3\envs\spyder\lib\site-packages\spyder\plugins\variableexplorer\widgets\collectionseditor.py", line 1117, in __init__
    minmax=minmax)
  File "C:\Users\ifeed\anaconda3\envs\spyder\lib\site-packages\spyder\plugins\variableexplorer\widgets\collectionseditor.py", line 143, in __init__
    self.set_data(data)
  File "C:\Users\ifeed\anaconda3\envs\spyder\lib\site-packages\spyder\plugins\variableexplorer\widgets\collectionseditor.py", line 173, in set_data
    self.keys = sorted(list(data.keys()))
TypeError: '<' not supported between instances of 'str' and 'int'
Traceback (most recent call last):
  File "C:\Users\ifeed\anaconda3\envs\spyder\lib\site-packages\spyder\plugins\variableexplorer\widgets\collectionsdelegate.py", line 160, in createEditor
    readonly=readonly)
[...]

PR #13548, opened 2020-08-16
Natural sorting for dicts in variable explorer

skjerns:
* [ N/A] Wrote at least one-line docstrings (for any new functions) / 
* [X ] Added unit test(s) covering the changes (if testable)

As suggested in  https://github.com/spyder-ide/spyder/issues/13481#event-3658990833 I've implemented natural/human sorting for the keys. This means that numbers within strings are sorted correctly as well, e.g. test3 comes before test10, which with standard sorting would not be the case (as it just compares per character).

This could also be easily added to the main sorting mechanism of the variable explorer at https://github.com/spyder-ide/spyder/blob/9991196c311074236bb7cd0e56230919453560ab/spyder/widgets/collectionseditor.py#L250, should that be done? Might be quite useful.

### Affirmation

I certify the above statement is true and correct: skjerns

### DIFF
--- a/spyder/widgets/collectionseditor.py
+++ b/spyder/widgets/collectionseditor.py
@@ -67,6 +67,13 @@ LARGE_NROWS = 100
 ROWS_TO_LOAD = 50
 
 
+def natsort(s):
+    """
+    natural sorting, e.g. test3 comes before test100
+    taken from https://stackoverflow.com/a/16090640/3110740
+    """
+    return [int(t) if t.isdigit() else t.lower() for t in re.split('(\d+)', s)]
+
 class ProxyObject(object):
     """Dictionary proxy to an unknown object."""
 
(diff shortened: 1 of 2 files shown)


skjerns:
can anyone enlighten me why the slow tests work but the fast tests fail?

ccordoba12:
@skjerns, thanks a lot for your contribution! We'd like to include it in our next release in the Spyder 4 series. For that, please make your branch to derive from our `4.x` branch (instead of master), with the following commands:

    git checkout 4.x
    git checkout 
    git rebase --onto 4.x master natural_sort_vareditor
    git push -f origin natural_sort_vareditor

If you decline to do that, your work will be part of Spyder 5, to be released in 2021.

skjerns:
@ccordoba12 I'm not getting the rebase to work and have a ton of merge conflicts. As I don't exactly know what belongs where I decided to create a new PR at #13558 . Sorry! But let's continue there for 4.x.

PR #13600, opened 2020-08-21
PR:  Refactor the toolbar handling on main widgets

goanpeca:
## Description of Changes

### Issue(s) Resolved

- The code was simplified so instead of using a QMainWindow and QToolBars added to the ToolBar Areas, we use a QWidget and QToolBars added to QBoxLayouts. This way we have full control of the location, and stretches which fix the issue.
- The API itself remained unchanged, the changes were internal to the `PluginMainWidget`.
- This is how Help and Find in Files look (which also had the similar issue found in Pylint)
- Included an extra fix on the find in files button, to make it have the same size, regardless of Search/Stop text.
- The spinner widget is "permanently visible"  so that it takes width and there are no jumps when it starts spinning (and is actually displayed).
- Added a fix for find in files to keep the search combobox and exclude combobox the same size.
Screen Shot 2020-08-21 at 20 10 14

## Help
Screen Shot 2020-08-21 at 20 03 16

# Find
Screen Shot 2020-08-21 at 20 03 23
Screen Shot 2020-08-21 at 20 03 29

### Affirmation

### DIFF
--- a/spyder/api/plugins.py
+++ b/spyder/api/plugins.py
@@ -817,10 +817,8 @@ class SpyderPluginV2(QObject, SpyderActionMixin, SpyderOptionMixin):
             # Widget setup
             # ----------------------------------------------------------------
             try:
-                _setup = getattr(container, "_setup", None)
-                if _setup:
-                    _setup(options=options)
-            except AttributeError as error:
+                container._setup(options=options)
+            except Exception as error:
                 logger.debug(
                     "Running `_setup` on {0}: {1}".format(self, error))
 
@@ -1607,14 +1605,10 @@ class SpyderDockablePlugin(SpyderPluginV2):
         widget.set_icon(self.get_icon())
         widget.set_name(self.NAME)
 
-        # TODO: Streamline this by moving to postvisible setup
         # Render all toolbars as a final separate step on the main window
         # in case some plugins want to extend a toolbar. Since the rendering
         # can only be done once!
-        widget._main_toolbar._render()
-        widget._corner_toolbar._render()
-        for __, toolbar in widget._auxiliary_toolbars.items():
-            toolbar._render()
+        widget.render_toolbars()
 
         # Default Signals
         # --------------------------------------------------------------------
--- a/spyder/api/plugins.py
+++ b/spyder/api/plugins.py
@@ -24,6 +24,7 @@ There are two types of plugins available:
 # Standard library imports
 from collections import OrderedDict
 import inspect
+import logging
 import os
 
 # Third party imports
@@ -51,6 +52,7 @@ from spyder.utils.qthelpers import create_action
 
 # Localization
 _ = get_translation('spyder')
+logger = logging.getLogger(__name__)
 
 
 # =============================================================================
@@ -815,9 +817,12 @@ class SpyderPluginV2(QObject, SpyderActionMixin, SpyderOptionMixin):
             # Widget setup
             # ----------------------------------------------------------------
             try:
-                container._setup(options=options)
-            except AttributeError:
-                pass
+                _setup = getattr(container, "_setup", None)
+                if _setup:
+                    _setup(options=options)
+            except AttributeError as error:
+                logger.debug(
+                    "Running `_setup` on {0}: {1}".format(self, error))
 
             if isinstance(container, SpyderWidgetMixin):
                 container.setup(options=options)
@@ -1606,10 +1611,10 @@ class SpyderDockablePlugin(SpyderPluginV2):
         # Render all toolbars as a final separate step on the main window
         # in case some plugins want to extend a toolbar. Since the rendering
         # can only be done once!
-        widget.get_main_toolbar()._render()
-        for __, toolbars in widget._aux_toolbars.items():
-            for toolbar in toolbars:
-                toolbar._render()
+        widget._main_toolbar._render()
+        widget._corner_toolbar._render()
+        for __, toolbar in widget._auxiliary_toolbars.items():
+            toolbar._render()
 
         # Default Signals
         # --------------------------------------------------------------------
--- a/spyder/api/widgets/__init__.py
+++ b/spyder/api/widgets/__init__.py
@@ -20,6 +20,7 @@ example it is just providing a new completion client.
 
 # Standard library imports
 from collections import OrderedDict
+import os
 import sys
 import textwrap
 import uuid
@@ -212,6 +213,27 @@ class PluginMainWidget(QWidget, SpyderWidgetMixin, SpyderToolBarMixin):
     """
     DEFAULT_OPTIONS = {}
 
+    # --- Attributes
+    # ------------------------------------------------------------------------
+    ENABLE_SPINNER = False
+    """
+    This attribute enables/disables showing a spinner on the top right to the
+    left of the corner menu widget (Hamburguer menu).
+
+    Plugins that provide actions that take time should make this `True` and
+    use accoringly with the `start_spinner`/`stop_spinner` methods.
+
+    The Find in files plugin is an example of a core plugin that uses it.
+
+    Parameters
+    ----------
+    ENABLE_SPINNER: bool
+        If `True` an extra space will be added to the toolbar (even if the
+        spinner is not moving) to avoid items jumping to the left/right when
+        the spinner appears. If `False` no extra space will be added. Default
+        is False.
+    """
+
     # --- Signals
     # ------------------------------------------------------------------------
     sig_free_memory_requested = Signal()
@@ -318,6 +340,7 @@ class PluginMainWidget(QWidget, SpyderWidgetMixin, SpyderToolBarMixin):
         self.dock_action = None
         self.undock_action = None
         self.close_action = None
+        self._toolbars_already_rendered = False
 
         # We create our toggle action instead of using the one that comes with
         # dockwidget because it was not possible to raise and focus the plugin
@@ -330,7 +353,11 @@ class PluginMainWidget(QWidget, SpyderWidgetMixin, SpyderToolBarMixin):
         self.windowwidget = None
         self.dockwidget = None
         self._icon = QIcon()
-        self._spinner = create_waitspinner(size=16, parent=self)
+        self._spinner = None
+
+        if self.ENABLE_SPINNER:
+            self._spinner = create_waitspinner(size=16, parent=self)
+
         self._corner_widget = MainCornerWidget(
             parent=self,
             name=PluginMainWidgetWidgets.CornerWidget,
@@ -405,6 +432,17 @@ class PluginMainWidget(QWidget, SpyderWidgetMixin, SpyderToolBarMixin):
                 # For widgets that use tabs, we add the corner widget using
                 # the setCornerWidget method.
                 child.setCornerWidget(self._corner_widget)
+
+                # This is needed to ensure the corner ToolButton (hamburguer
+                # menu) is aligned with plugins that use ToolBars vs
+                # CornerWidgets
+                # See: spyder-ide/spyder#13600
+                # left, top, right, bottom
+                if os.name == "nt":
+                    self._corner_widget.setContentsMargins(0, 0, 2, 0)
+                else:
+                    self._corner_widget.setContentsMargins(0, 2, 2, 0)
+
                 break
 
         self._options_button = self.create_toolbutton(
@@ -417,10 +455,11 @@ class PluginMainWidget(QWidget, SpyderWidgetMixin, SpyderToolBarMixin):
             PluginMainWidgetWidgets.OptionsToolButton,
             self._options_button,
         )
-        self.add_corner_widget(
-            PluginMainWidgetWidgets.Spinner,
-            self._spinner,
-        )
+        if self.ENABLE_SPINNER:
+            self.add_corner_widget(
+                PluginMainWidgetWidgets.Spinner,
+                self._spinner,
+            )
 
         # Widget setup
         # --------------------------------------------------------------------
@@ -639,13 +678,15 @@ class PluginMainWidget(QWidget, SpyderWidgetMixin, SpyderToolBarMixin):
         """
         Start default status spinner.
         """
-        self._spinner.start()
+        if self.ENABLE_SPINNER:
+            self._spinner.start()
 
     def stop_spinner(self):
         """
         Stop default status spinner.
         """
-        self._spinner.stop()
+        if self.ENABLE_SPINNER:
+            self._spinner.stop()
 
     def create_toolbar(self, toolbar_id):
         """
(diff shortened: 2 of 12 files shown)


dalthviz:
@ccordoba12 could you create a list of the panes we should be checking please? Thanks :)

spyder/api/plugins.py
```diff
@@ -815,9 +817,12 @@ def __init__(self, parent, configuration=None):
             # Widget setup
             # ----------------------------------------------------------------
             try:
-                container._setup(options=options)
-            except AttributeError:
-                pass
+                _setup = getattr(container, "_setup", None)
+                if _setup:
+                    _setup(options=options)
+            except AttributeError as error:
```
dalthviz, line 823: Since above we are using `getattr` and doing a validation for `_setup` is this except reachable? Also, is there any specific case for `_setup` to not be defined? Maybe documenting that here could be useful

spyder/api/plugins.py
```diff
@@ -1606,10 +1611,10 @@ def __init__(self, parent, configuration):
         # Render all toolbars as a final separate step on the main window
         # in case some plugins want to extend a toolbar. Since the rendering
         # can only be done once!
-        widget.get_main_toolbar()._render()
-        for __, toolbars in widget._aux_toolbars.items():
-            for toolbar in toolbars:
-                toolbar._render()
+        widget._main_toolbar._render()
+        widget._corner_toolbar._render()
+        for __, toolbar in widget._auxiliary_toolbars.items():
```
dalthviz, line 1616: Should the widget attributes above (`_main_toolbar`, `_corner_toolbar` and `_auxiliary_toolbars`) be called using a `get_` method (since seems like they are private)?

spyder/widgets/waitingspinner.py
```diff
@@ -121,9 +124,10 @@ def start(self):
             self._timer.start()
             self._currentCounter = 0
 
+        self.show()
+
     def stop(self):
         self._isSpinning = False
-        self.hide()
```
dalthviz, line 126: For the spinner now, instead of using `hide`, a variable to interrupt the paint of the widget is done? There was some issue that prevents to use hide maybe?

spyder/api/plugins.py
goanpeca: Good catch. Will fix.

spyder/api/plugins.py
goanpeca: Yes, much better like that. Will fix.

spyder/widgets/waitingspinner.py
goanpeca: @dalthviz the issue is that now the spinner is shown all the time either empty or spinning. The idea is to have it occupy spaced even if not spinning to avoid the toolbar widgets suddenly jumping to the left and to the right when the spinner starts or stops.

I can add this comment to explain why, to I could add a setting on the spinner to enable disable this with a setting on the spinner.

ccordoba12:
> could you create a list of the panes we should be checking please? Thanks :)

Plugins that were already moved to the new API and could be affected due to this are:

* History
* Internal Console
* Find
* Profiler
* Help
* Online help
* Breakpoints
* Plots

spyder/widgets/waitingspinner.py
dalthviz: It would be nice for the plugins to have an option to actually set the use of the spinner (where the behavior of leaving the space should be preserved) or not having it (where it should not be taking space at all). So maybe going for the spinner setting could be worthy :+1:

goanpeca: Ok @dalthviz, sounds good.

dalthviz:
Checking on Windows seems like, except for the History pane, everything looks good:

toolbar

I'm unsure but seems like the History toolbar looks a little bit small than the rest?

## History
image

## Internal Console
image

goanpeca:
Yeah, the history one looks funny. Will see what is going on.

dalthviz:
@andfoy @steff456 @juanis2112 @ccordoba12 could you guys help here testing this? Basically the idea is to check the resize of the window, the style of the toolbars of each pane, and the overall style of the actions in the toolbars in linux and mac

steff456:
Hi! I tested the following plugins in macOS
1. History
2. Internal Console
3. Find
4. Profiler
And they all resize well and the style of the toolbars also match with 4.x branch

goanpeca:
I have to make the history button fix and add the option to enable a spinner.

goanpeca:
Hi @dalthviz,

Added the fixes we talked about. Making some comments on the `files` section.

test

spyder/api/plugins.py
```diff
@@ -1607,14 +1607,10 @@ def __init__(self, parent, configuration):
...
 
-        # TODO: Streamline this by moving to postvisible setup
         # Render all toolbars as a final separate step on the main window
         # in case some plugins want to extend a toolbar. Since the rendering
         # can only be done once!
-        widget._main_toolbar._render()
-        widget._corner_toolbar._render()
-        for __, toolbar in widget._auxiliary_toolbars.items():
-            toolbar._render()
+        widget.render_toolbars()
```
goanpeca, line 1613: I added get methods as suggested but added this extra method to encapsulate all.

spyder/api/widgets/__init__.py
```diff
@@ -212,6 +212,27 @@ class PluginMainWidget(QWidget, SpyderWidgetMixin, SpyderToolBarMixin):
...
+
+    Parameters
+    ----------
+    ENABLE_SPINNER: bool
+        If `True` an extra space will be added to the toolbar (even if the
+        spinner is not moving) to avoid items jumping to the left/right when
+        the spinner appears. If `False` no extra space will be added. Default
+        is False.
+    """
+
```
goanpeca, line 235: This attribute now enables disables the spinner

spyder/api/widgets/__init__.py
```diff
@@ -405,6 +431,13 @@ def _setup(self, options=DEFAULT_OPTIONS):
                 # For widgets that use tabs, we add the corner widget using
                 # the setCornerWidget method.
                 child.setCornerWidget(self._corner_widget)
+                
+                # This is needed to ensure the corner ToolButton (hamburguer
+                # menu) is aligned with plugins that use ToolBars vs
+                # CornerWidgets
+                # See: spyder-ide/spyder#13600
+                # left, top, right, bottom
+                self._corner_widget.setContentsMargins(0, 2, 2, 0)
```
goanpeca, line 440: This is needed to ensure things look ok on Mac. This may need to be a different value on a different OS.

spyder/plugins/findinfiles/widgets.py
```diff
@@ -1046,8 +1047,9 @@ def update_actions(self):
             w1 = widget.fontMetrics().width(stop_text)
             w2 = widget.fontMetrics().width(search_text)
             # Ensure the search/stop button has the same size independent on
-            # the length of the words.
-            width = self.get_options_menu_button().width() + max([w1, w2]) + 2
+            # the length of the words. The 10 additional pixels account for 
+            # operating system spacing differences
+            width = self.get_options_menu_button().width() + max([w1, w2]) + 10
```
goanpeca, line 1052: This ensure the start/stop button on Find in Files keeps the same size, avoiding some jumps of the items (left/right)

spyder/widgets/tabs.py
```diff
@@ -256,16 +256,27 @@ def __init__(self, parent, actions=None, menu=None,
...
+        # See: spyder-ide/spyder#13600
+        self.setStyleSheet("""
+            QTabBar::tab {
+                margin-top: 5px;
+            }
+            QTabWidget::tab-bar {
+                alignment: left;
+            }
+            """
+        )
```
goanpeca, line 278: The comments explain why this is needed, but this may need different values on different OS.

spyder/api/widgets/__init__.py
dalthviz: I was checking this and seems like there is a TODO in the history pane that could be related with the different visual between OS?: 

https://github.com/spyder-ide/spyder/blob/289fc426337ae8d8348119514a9c96af92714c84/spyder/plugins/history/widgets.py#L94-L103

Also, on WIndows, seems like the values should be `0, 0, 0, 0`. With does values the icon size is correct but there is a little misalignment (maybe since the tab is not higher because no `x` button is present in the history tab?):

toolbars

spyder/api/plugins.py
dalthviz: Thanks, and about the `_setup` is there any specific situation where this happens?

spyder/plugins/findinfiles/widgets.py
dalthviz: Maybe the `10` should go to a constant?

spyder/widgets/tabs.py
dalthviz: Checking this need to be 8px on Windows :+1: (and also the changes in the values described above need to be done: https://github.com/spyder-ide/spyder/pull/13600/files/289fc426337ae8d8348119514a9c96af92714c84#r480266785):

toolbars

dalthviz:
Thanks @goanpeca , I left some comments and questions regarding the changes and some specific values for the history layout that worked for me on Windows (some tweaks to the values needed to handle a tabwidget as part of the toolbar and since for the different OS the values could change if I'm understanding correctly, right?)

spyder/api/plugins.py
goanpeca: Cannot recall now why I left this. Will remove it.

spyder/api/widgets/__init__.py
goanpeca: >TODO in the history pane that could be related to the different visual between OS?

No, that has to do with plugins using tabs. They should not worry about doing that manually and instead, we should handle that in the `tabs.py`  file directly. So it is a different issue.

Updated to 0,0,0,0 on windows

spyder/plugins/findinfiles/widgets.py
goanpeca: It is only used there. Could you elaborate on how this improves ... something  @dalthviz ?

spyder/widgets/tabs.py
goanpeca: Updated to 8.

goanpeca:
Thanks for the feedback @dalthviz. Added the suggested fixes and a question to one of the suggestions.

spyder/plugins/findinfiles/widgets.py
dalthviz: Just for readability and if needed to calculate something else in the future but mainly just for style proposes at this moment

spyder/widgets/tabs.py
```diff
@@ -256,16 +257,28 @@ def __init__(self, parent, actions=None, menu=None,
...
             if actions:
                 add_actions(self.menu, actions)
         else:
             self.menu = menu
 
+        # QTabBar forces the corner widgets to be smaller than they should on
+        # some plugins, like History. The top margin added allows the
+        # toolbuttons to expand to their normal size.
+        # See: spyder-ide/spyder#13600
+        top_margin = 8 if os.name == "nt" else 5
```
dalthviz, line 271: Checking again seems like in a resolution of 1366 x 768 a value of 9 looks better, but testing in a screen with 1920 x 1080 seems like 5 is better (if 9 is set, the misalignment appears again with the higher resolution):

1366 x 768
tlow


1920 x 1080
thigh


What do you think we can do to prevent this? Maybe qss style sheets allow media queries?

spyder/api/widgets/__init__.py
```diff
@@ -367,28 +429,22 @@ def _setup(self, options=DEFAULT_OPTIONS):
...
-        else:
-            corner_widget = self._corner_widget
+                
+                # This is needed to ensure the corner ToolButton (hamburguer
+                # menu) is aligned with plugins that use ToolBars vs
+                # CornerWidgets
+                # See: spyder-ide/spyder#13600
+                # left, top, right, bottom
+                if os.name == "nt":
+                    self._corner_widget.setContentsMargins(0, 0, 0, 0)
```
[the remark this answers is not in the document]
goanpeca: Thanks

spyder/api/plugins.py
```diff
@@ -816,8 +818,9 @@ def __init__(self, parent, configuration=None):
             # ----------------------------------------------------------------
             try:
                 container._setup(options=options)
```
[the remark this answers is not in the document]
goanpeca: I removed the attribute error, now it just logs if any exception occurs.

spyder/widgets/tabs.py
goanpeca: >Maybe qss style sheets allow media queries?

No, QSS does not include media queries as far as I can tell. (https://doc.qt.io/Qt-5/stylesheet-syntax.html)

The hamburger button being misaligned is an issue since 1.x, I think the problem at hand was actually having the tool button be smaller.

I tried to improve the misalignment as it is something that I think would make the UI consistent, but that fix should be addressed separately.

spyder/plugins/findinfiles/widgets.py
goanpeca: Fixed.

goanpeca:
>@goanpeca left some more comments

Left some more answers.

>and also I was thinking that maybe we should be using the spinner for other plugins like the Online Help pane and the Help pane

Made the fixes @dalthviz.

spyder/widgets/tabs.py
goanpeca: >Ok 👍 I thought that one of the goals of the new API was to solve such UI inconsistency

It does not fix it, but it will make fixing it much easier, since now the API is in charge of the hamburger button and providing a unified way of adding buttons and widgets to the toolbar of a panel. We can tweak a value in a single place, instead of having to tweak every single plugin to try to "manually (and unsuccessfully) recreate" the layout, as is the case of 4.x and before. 

>if you have any idea of how to prevent such an issue let us know

Besides manual tweaking and having it look best at a given resolution, I do not see any other way around it as the QSS to style the TabBar of a TabWidget is not as flexible as other widgets.

goanpeca: >Also please change this to be 9 instead of 8 on Windows then

Fixed.

goanpeca:
>@goanpeca left a couple of comments and seems like some PEP8 issues need a fix. Besides that this LGTM 👍

Thanks for the review @dalthviz.

I made the requested fixes.

dalthviz:
Last comment @goanpeca , please enable the spinner for the Profiler (I almost missed that pane), thanks!

goanpeca:
>Last comment @goanpeca , please enable the spinner for the Profiler (I almost missed that pane), thanks!

Fixed.

steff456:
@dalthviz we already checked this PR and it is ready for merge!

PR #13621, opened 2020-08-24
PR: Install PyLS locally without modifying site-packages

andfoy:
## Description of Changes
This PR installs pyls on a external directory different to site-packages, by doing so, Spyder won't remove any user-installed pyls

### Issue(s) Resolved

Fixes #

### Affirmation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ result.xml
 
 # Ignore setuptools development files in the PyLS subrepo
 external-deps/python-language-server/python_language_server.egg-info/
+external-deps/python-language-server/.installation-dir/
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -20,6 +20,7 @@ See spyder-ide/spyder#741.
 # pylint: disable=C0413
 
 # Standard library imports
+import atexit
 import argparse
 import os
 import os.path as osp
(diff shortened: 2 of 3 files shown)


jitseniesen:
@andfoy Can you please explain what this means for us, developers (or point to an existing description)? I never really understood all the magic that bootstrap.py is doing with pyls and spyder-kernels; I just know that it usually works, but sometimes it does not ...

andfoy:
@jitseniesen, basically Spyder starts with its own development version of pyls, due to the difficulty and complexity of producing releases each time a critical feature/bug is introduced there (pyls is controlled by Palantir and we have limited permissions). So basically we have a local version of pyls independent of any other version that would exist on site-packages, which point to the latest develop/branch commit that we require on Spyder

jitseniesen:
@andfoy Sounds good, thanks for the explanation!

ccordoba12:
That's the big picture @jitseniesen. Regarding this PR, it changes the way we use the subrepo present in `external-deps/python-language-server` from `bootstrap.py` to **not** remove the current version present in site-packages and reinstall it again in development mode (which is the solution I came up with in PR #12389).

Instead, @andfoy found a brilliant solution: install that subrepo in development mode in a temporary directory and patch `sys.path` to use it (this and removing that temp dir after closing Spyder is handled by `bootstrap.py` now). That way we don't need to touch the PyLS version present in `site-packages` anymore, which won't break people's environments from now on.

PR #13627, opened 2020-08-25
PR: Fix some issues with the PyLS subrepo

ccordoba12:
* This uses the PyLS subrepo to run our tests, which was missing from PR #13621.
* This also fixes starting Spyder from `bootstrap.py` on Windows after PR #13621.


dalthviz (approved):
Thanks @ccordoba12 LGTM :+1:

PR #13673, opened 2020-08-31
PR: Add a minimum width to the help pane

### LINKED ISSUE #13664: Help pane layout overflows
## Problem Description

If I reduce the size of the help pane, then the text will overflow.

image

In 4.x, we have a minimum width for all the panes in the right side so this issue doesn't happen,

image

1. Open the help pane and request some help via `Ctrl+I`
2. Reduce the pane width

I will expect to have a minimum width so the layout doesn't break

goanpeca:
## Description of Changes

This PR adds a minimum width to the `Help` pane of `400px`.

This fixes the reported issue.
image

However:
- Zooming in will get to a point where this fix will not work. Setting a minimum width cannot overcome this issue.
image

- There may be modules that have very long imports (79 or higher columns) which will continue to display the issue. Setting a minimum width cannot overcome this issue.

### Issue(s) Resolved

Fixes #13664

### Affirmation

### DIFF
--- a/spyder/plugins/help/widgets.py
+++ b/spyder/plugins/help/widgets.py
@@ -321,6 +321,7 @@ class HelpWidget(PluginMainWidget):
         self.object_edit = QLineEdit(self)
 
         # Setup
+        self.setMinimumWidth(400)
         self.object_edit.setReadOnly(True)
         self.object_combo.setMaxCount(self.get_option('max_history_entries'))
         self.object_combo.setItemText(0, '')


steff456:
@goanpeca there's a way to wrap the text so even if the user zooms in then the layout doesn't break? If that's possible I think it is a cleaner fix than the minimum width of the pane

goanpeca:
Thanks for the feedback @steff456. 

>@goanpeca there's a way to wrap the text so even if the user zooms in then the layout doesn't break? 

Not that I am aware of since that is how Sphinx is rendering the docstrings into HTML.

>If that's possible I think it is a cleaner fix than the minimum width of the pane

I don't know how it can be fixed. Also, that fix is unrelated to the new API work.

dalthviz:
Thanks @steff456 and @goanpeca for working on this. After some thought seems like the work needed goes beyond the scope of the work of the API and the fixes of migrations issues. I'm closing this one and rephrasing the open issue to be more clear about the __responsive__ behavior @steff456 was talking about.

PR #13675, opened 2020-08-31
PR: Fix some issues with the Find in files plugin

### LINKED ISSUE #13657: Find pane issues and details
Some issues and details that needed to be worked out in the Find pane

- [ ] The hamburger menu is not the same as in 4.x (some actions are not being listed)
- [ ] The label of the directory was changed (from `Search in:` to `Location`).
- [ ] A label that says `Search` was added to the combobox used to type search terms but the button already says `Search` on the right (so it is kind of redundant).
- [ ] In the Editor having a word selected and hitting `Cmd + Shift + F`/`Ctrl + Shift  + F` doesn’t automatically start a search.

goanpeca:
## Description of Changes

### Issue(s) Resolved

Fixes #13657

### Affirmation

### DIFF
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -2452,7 +2452,7 @@ def test_preferences_change_font_regression(main_window, qtbot):
 
 @pytest.mark.slow
 @pytest.mark.skipif(
-    sys.platform == 'darwin',
+    not sys.platform.startswith('linux'),
     reason="Changes of Shitf+Return shortcut cause an ambiguos shortcut")
 def test_preferences_empty_shortcut_regression(main_window, qtbot):
     """
--- a/spyder/config/main.py
+++ b/spyder/config/main.py
@@ -495,6 +495,8 @@ DEFAULTS = [
               'explorer/paste file': 'Ctrl+V',
               'explorer/copy absolute path': 'Ctrl+Alt+C',
               'explorer/copy relative path': 'Ctrl+Alt+Shift+C',
+              # ---- In plugins/findinfiles/plugin ----
+              'find_in_files/find in files': 'Ctrl+Alt+F',
               }),
             ('appearance', APPEARANCE),
             ('lsp-server',
--- a/spyder/plugins/findinfiles/plugin.py
+++ b/spyder/plugins/findinfiles/plugin.py
@@ -27,7 +27,7 @@ _ = get_translation('spyder')
 # --- Constants
 # ----------------------------------------------------------------------------
 class FindInFilesActions:
-    FindInFiles = 'find_in_files_action'
+    FindInFiles = 'find in files'
 
 
 # --- Plugin
@@ -79,13 +79,13 @@ class FindInFiles(SpyderDockablePlugin):
             tip=_("Search text in multiple files"),
             triggered=self.find,
             register_shortcut=True,
+            context=Qt.WindowShortcut
         )
         menu = self.get_application_menu(ApplicationMenus.Search)
         self.add_item_to_application_menu(
             findinfiles_action,
             menu=menu,
         )
-        findinfiles_action.triggered.connect(lambda: self.switch_to_plugin())
 
         search_toolbar = self.get_application_toolbar(
             ApplicationToolBars.Search)
--- a/spyder/plugins/findinfiles/plugin.py
+++ b/spyder/plugins/findinfiles/plugin.py
@@ -9,6 +9,7 @@ Find in Files Plugin.
 
 # Third party imports
 from qtpy.QtCore import Qt
+from qtpy.QtWidgets import QApplication
 
 # Local imports
 from spyder.api.menus import ApplicationMenus
@@ -72,7 +73,13 @@ class FindInFiles(SpyderDockablePlugin):
             working_directory.sig_current_directory_changed.connect(
                 self.refresh_search_directory)
 
-        findinfiles_action = self.get_action(FindInFilesWidgetActions.Find)
+        findinfiles_action = self.create_action(
+            FindInFilesActions.FindInFiles,
+            text=_("Find in files"),
+            tip=_("Search text in multiple files"),
+            triggered=self.find,
+            register_shortcut=True,
+        )
         menu = self.get_application_menu(ApplicationMenus.Search)
         self.add_item_to_application_menu(
             findinfiles_action,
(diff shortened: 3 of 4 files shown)


goanpeca:
@dalthviz this PR fixes the issues, except:

>The hamburger menu is not the same as in 4.x (some actions are not being listed)

This fix requires changes on the `onecolumntree.py`  widget already implemented on 
https://github.com/spyder-ide/spyder/pull/12160/files#diff-e48228a75dfcacb97995edd08cf6290cR69

To be able to apply the fix on this PR, that PR needs to be merged so those actions using the new API are exposed.

test

steff456 (changes requested):
Hi @goanpeca,

When I select something in the editor and try to use the shortcut `Ctrl + Shift + F` in Linux, the search starts but it is empty
image

In macOS, it opens the pane but it doesn't do anything,
image

If I use the option in the menu, it works.

goanpeca:
Thanks for the feedback @steff456, copying what I already posted on #13657

>In the Editor having a word selected and hitting Cmd + Shift + F/Ctrl + Shift + F doesn’t automatically start a search.

It cannot be fixed to behave as 4.x, since on 4.x it is merging 2 different actions under the same shortcut and the new API does not allow for these hacks to work. It is not a matter of the new API not being able to do something and more of a case that things in 4.x (and before since this probably goes back to 2.x ...) being done in a peculiar way, to save ... something. 

### On 4.x:
- There is the `Switch to Find in Files`  and that is set `Cmd + Shift + F` by default.
image

image

- Another completely different action that is added to the `Search Menu` is taking the selected text from the Editor, pasting on the Find in Files plugin, showing it if it hidden and then running the search.
image

- This completely different action is also using the same `Cmd + Shift + F` shortcut.

### On master:

- There is the `Switch to Find in Files`  and that is set `Cmd + Shift + F` by default. Triggering this action will show the pane, as expected
- There is the `Find in Files` action that will show the plugin if not visible and start running the `find in files plugin`if there is text in it. It is not copying the text and pasting it on the plugin (This I will fix)
- This separate action is available now on the shortcut preferences and can be configured, but cannot use the same shortcut. 

Action points:
1. I can have the menu do the expected action, but not use the same shortcut as another action (Switch to Find in files).
2. I can add a default shortcut (whatever is decided?) or this action will not have a shortcut by default. Note that currently this action is not exposed in the shortcut preferences (it is a matter of changing  `register_shortcut=True` when creating the action).
3. I can fix the `selected text` behavior so it does the same thing on master

ccordoba12:
> It is not a matter of the new API not being able to do something and more of a case that things in 4.x (and before since this probably goes back to 2.x ...) being done in a peculiar way, to save ... something.

This is clearly a regression in functionality with respect to 4.x, something we agreed we should avoid at all costs.

Since you're stating there's no problem with the new API not being able to do this (hence there are no technical reasons preventing it), please implement the necessary changes to be able to support how things were working in 4.x.

I think this requires creating a new action that uses the `Ctrl+Shift+F` shortcut to show the plugin and start a search with the currently select word in the editor (which was probably how things were working in 4.x); and use that action instead of the `toggle_view_action` for this particular plugin.

goanpeca:
>Since you're stating there's no problem with the new API not being able to do this (hence there are no technical reasons preventing it), please implement the necessary changes to support how things were working in 4.x.

I said the opposite. 

It cannot be fixed to behave as 4.x, since on 4.x it is merging 2 different actions under the same shortcut and the new API does not allow for these hacks to work. It is not a matter of the new API not being able to do something and more of a case that things in 4.x (and before since this probably goes back to 2.x ...) being done in a peculiar way, to save ... something.

>I think this requires creating a new action that uses the Ctrl+Shift+F shortcut to show the plugin and start a search with the currently select word in the editor (which was probably how things were working in 4.x); and use that action instead of the toggle_view_action for this particular plugin.

There are two actions (the same as 4.x) but they cannot use the same shortcut because there are completely different actions and the new API does not allow for that.

ccordoba12:
> I do not understand how that could have happened since Ctrl + Shift + F should only Switch to the plugin, not trigger anything.

This is not correct in 4.x. In that branch it's possible to redefine how the toggle_view_action works for each plugin to have more functionality associated to it. For instance, if the IPython console is hidden at startup and you click on that action, that plugin is not only shown but also a new console is created for the user. That saves resources (by not starting a kernel at startup) if users don't want to use the IPython console for some reason, but displays a new console as soon as they want to use one.

If this is not possible in the new API, you need to fix it because it's a clear regression in functionality. You can do that in a new PR and then come back to this one.

ccordoba12:
@goanpeca, I showed you a way in which this regression can be fixed (not with two actions, but by redefining `toggle_view_action` for `Find`). So please proceed with that plan.

ccordoba12:
> There are two actions (the same as 4.x) but they cannot use the same shortcut because there are completely different actions and the new API does not allow for that.

Taking a look at the code in `4.x`, there are no two actions (as you said) but a single action that calls the API method of the plugin plus another method to start the search:

https://github.com/spyder-ide/spyder/blob/c710b0259e4143df962cc61bae6170db43f7861b/spyder/plugins/findinfiles/plugin.py#L185-L192

https://github.com/spyder-ide/spyder/blob/c710b0259e4143df962cc61bae6170db43f7861b/spyder/plugins/findinfiles/plugin.py#L142-L150

And this exactly the same idea I mentioned we use in the IPython console.

goanpeca:
> I showed you a way in which this regression can be fixed (not with two actions, but by redefining toggle_view_action for Find). So please proceed with that plan.

@ccordoba12 I don't know how to have the same action show with a different name on two different menus. 

- Search menu says `Find in files`
- toggle view action uses the `get_name` of the plugin, which is `Find`

goanpeca:
I could reuse the toggle view action (switch to is not an action) but then the names would end up being both

`Find in files` or `Find` in both the Search Menu and the View/Pane Menu.

ccordoba12:
> I could reuse the toggle view action (switch to is not an action)

Great! Nice to know that's possible.

> Find in files or Find in both the Search Menu and the View/Pane Menu.

I think we could call it `Find in files`, but add a hack that changes its text to `Find` when users display the `View > Panes` menu. What do you think?

ccordoba12:
After trying to apply my own suggestions above, I noticed the following: although it's possible to redefine `toggle_view` to start a search besides switching to the plugin, we wouldn't be able to reuse its associate switch-to-plugin action in the `Search` menu because it's a toggle one.

Also, we could hack the `Find in Files` action added by @goanpeca to make it show `Ctrl+Shift+F` when the menu is displayed. But that would also require to synchronize it with the switch-to-plugin action, so both display the same shortcut all the time. That's too much hacky work just to restore what we have in Spyder 4.

So I decided to assign a new shortcut to that action: `Ctrl+Alt+F`. It's a small regression but I agree with @goanpeca that is the best thing we can do with the new API.

ccordoba12 (approved):
Thanks @goanpeca!

PR #13693, opened 2020-09-02
PR: Fix 2 issues related to PyLS on Windows

trollodel:
## Description of Changes

Probably caused by #13440.

#### Spyder unable to start

I don't know what is the real problem, but I fix it by passing the boostrap.py
environment variables to the PyLS setup.py.

Here is the error printed by the interpreter, with the traceback.
```
Fatal Python error: _Py_HashRandomization_Init: failed to get random numbers to initialize Python
Python runtime state: preinitialized

  File "bootstrap.py", line 166, in <module>
    subprocess.check_output(
  File "C:\Program Files\Python38\lib\subprocess.py", line 411, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "C:\Program Files\Python38\lib\subprocess.py", line 512, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['C:\\Program Files\\Python38\\python.exe', '-W', 'ignore', 'setup.py', 'develop', '--no-deps', '--install-dir', 'C:\\Users\\User\\spyder\\external-deps\\python-language-server\\.installation-dir']' returned non-zero exit status 1.
```

#### PyLS unable to start

After Spyder loading, PyLS raises an exception that make it unavailable.
Analyzing the issue, I discover that a  `RuntimeError: Can't determine home directory` is raised by [os.path.expanduser](https://docs.python.org/3/library/os.path.html#os.path.expanduser) because the USERPROFILE env variable is not set (the call to that function is done by matplotlib). So I pass that variable to the PyLS process.

### Issue(s) Resolved

None.

### Affirmation

I certify the above statement is true and correct: trollodel

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -163,6 +163,8 @@ if osp.exists(pyls_installation_dir) or osp.exists(pyls_installation_egg):
 
 # Install PyLS locally
 print("*. Installing PyLS locally")
+pyls_env = os.environ.copy()
+pyls_env['PYTHONPATH'] = pyls_installation_dir
 subprocess.check_output(
     [sys.executable,
      '-W',
@@ -172,7 +174,7 @@ subprocess.check_output(
      '--no-deps',
      '--install-dir',
      pyls_installation_dir],
-    env={'PYTHONPATH': pyls_installation_dir},
+    env=pyls_env,
     cwd=pyls_submodule
 )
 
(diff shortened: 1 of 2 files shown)


ccordoba12:
Thanks @trollodel for your help with this. I started working on this last week in PR #13627. I also added the `USERPROFILE` fix you found there so we can include it in our next release in the `4.x` series.

PR #13717, opened 2020-09-04
PR: Fix technical bugs on tour

### LINKED ISSUE #13240: Fix Introduction tour
## Problem Description

- [x] Add final message to point people to more help and letting them know the tour is over

- [x] Fix dialog bugs. (Some dialogs of the tour are appearing outside of Spyde'rs window)

- [x] Fix dropdown bug. (Clicking the dropdown reloads the dialog and doesn't show the dropdown)

- [x] Add panes that are missing (The tour only has the first 6 plugins so we should add the other ones that are not there)

1. 
2. 
3. 

```python-traceback

PASTE TRACEBACK HERE

```

## Versions

* Spyder version: 
* Python version: 
* Qt version: 
* PyQt version: 

### Dependencies

```
PASTE DEPENDENCIES HERE
```

juanis2112:
## Description of Changes

- [x] Add final message to point people to more help and letting them know the tour is over
Screenshot 2020-09-04 at 1 23 55 AM

- [x] Fix dialog bugs. (Some dialogs of the tour are appearing outside of Spyde'rs window)
(I changed all the tips to appear to the side and not on the bottom/top of the pane because most of them didn't fit ~~but there is one that even in the side doesn't fit, still trying to fix that one.~~ (Already fixed)
- [x] Fix dropdown bug. (Clicking the dropdown reloads the dialog and doesn't show the dropdown)
- [x] Add panes that are missing (The tour only has the first 6 plugins so we should add the other ones that are not there)
- [x] Maximize Spyder when starting the tour
- [x] Reset Spyder's layout when starting tour
- [x] Return to previous window size when closing the tour
- [x] Return to previous layout when closing the tour
- [x] Added a dialog asking to exit fullscreen for Mac users when they try to start tour in fullscreen as returning from fullscreen causes a lot of bugs.

Screenshot 2020-10-01 at 3 54 46 PM

- [x] Remove submenu from help pane

Screenshot 2020-09-30 at 4 44 13 PM

- [x] In MacOS close tour when switching to another application

### Issue(s) Resolved

Fixes #13240

### Affirmation

### DIFF
--- a/spyder/app/tour.py
+++ b/spyder/app/tour.py
@@ -1213,27 +1213,30 @@ class AnimatedTour(QWidget):
 
         self.is_tour_set = True
 
+    def _handle_fullscreen(self):
+        if self.spy_window.isFullScreen() or self.spy_window.fullscreen_flag:
+              if sys.platform == 'darwin':
+                  self.spy_window.setUpdatesEnabled(True)
+                  msg_title = _("Request")
+                  msg = _("To run the tour, please press the green button on "
+                          "the left of the Spyder window's title bar to take "
+                          "it out of fullscreen mode.")
+                  QMessageBox.information(self, msg_title, msg,
+                                          QMessageBox.Ok)
+                  return True
+              if self.spy_window.fullscreen_flag:
+                  self.spy_window.toggle_fullscreen()
+              else:
+                  self.spy_window.setWindowState(
+                      self.spy_window.windowState()
+                      & (~ Qt.WindowFullScreen))
+        return False
+  
     def start_tour(self):
         """ """
         self.spy_window.setUpdatesEnabled(False)
-        if self.spy_window.isFullScreen() or self.spy_window.fullscreen_flag:
-            if sys.platform == 'darwin':
-                self.spy_window.setUpdatesEnabled(True)
-                msg_title = _("Request")
-                msg = _("To run the tour, please press the green button on "
-                        "the left of the Spyder window's title bar to take "
-                        "it out of fullscreen mode.")
-                QMessageBox.information(self, msg_title, msg,
-                                        QMessageBox.Ok)
-                return
-            else:
-                if self.spy_window.fullscreen_flag:
-                    self.spy_window.toggle_fullscreen()
-                else:
-                    self.spy_window.setWindowState(
-                        self.spy_window.windowState()
-                        & (~ Qt.WindowFullScreen))
-
+        if self._handle_fullscreen():
+            return
         self.spy_window.save_current_window_settings(
             'layout_current_temp/',
             section="quick_layouts"
--- a/spyder/app/tour.py
+++ b/spyder/app/tour.py
@@ -958,8 +958,8 @@ class AnimatedTour(QWidget):
     def _resized(self, event):
         """ """
         if self.is_running:
-            size = event.size()
-            self.canvas.setFixedSize(size)
+            geom = self.parent.geometry()
+            self.canvas.setFixedSize(geom.width(), geom.height())
             self.canvas.update_canvas()
 
             if self.is_tour_set:
@@ -968,8 +968,8 @@ class AnimatedTour(QWidget):
     def _moved(self, event):
         """ """
         if self.is_running:
-            pos = event.pos()
-            self.canvas.move(QPoint(pos.x(), pos.y()))
+            geom = self.parent.geometry()
+            self.canvas.move(geom.x(), geom.y())
 
             if self.is_tour_set:
                 self._set_data()
--- a/spyder/app/tour.py
+++ b/spyder/app/tour.py
@@ -34,6 +34,7 @@ from spyder.config.gui import is_dark_interface
 from spyder.py3compat import to_binary_string
 from spyder.utils.qthelpers import add_actions, create_action
 from spyder.utils import icon_manager as ima
+from spyder import __docs_url__
 
 
 if is_dark_interface():
@@ -99,7 +100,8 @@ def get_tour(index):
     """
     sw = SpyderWidgets
     qtconsole_link = "https://qtconsole.readthedocs.io/en/stable/index.html"
-    ending_link = "https://docs.spyder-ide.org/current/index.html"
+    #docs_link = __docs_url__
+    docs_link = "https://docs.spyder-ide.org/current/index.html"
 
     # This test should serve as example of keys to use in the tour frame dics
     test = [{'title': "Welcome to Spyder introduction tour",
@@ -253,7 +255,7 @@ def get_tour(index):
              {'title': _("The end"),
               'content': _('You have reached the end of our tour and are ready to '
                            'start using Spyder! For more information, check out '
-                           f'our <a href="{ending_link}">documentation</a>.<br><br>'),
+                           f'our <a href="{docs_link}">documentation</a>.<br><br>'),
               'image': 'tour-spyder-logo.png'
               },
 
@@ -1146,7 +1148,11 @@ class AnimatedTour(QWidget):
                     y = y_glob + height/2 - self.tips.height()/2
 
                 if (y + self.tips.height()) > (self.y_main + self.height_main):
-                    y = y - ((y + self.tips.height())-(self.y_main + self.height_main))-offset
+                    y = (
+                        y
+                        - (y + self.tips.height() - (self.y_main + self.height_main))
+                        - offset
+                    )
         else:
             # Center on parent
             x = self.x_main + self.width_main/2 - self.tips.width()/2
--- a/spyder/app/tour.py
+++ b/spyder/app/tour.py
@@ -908,7 +908,7 @@ class AnimatedTour(QWidget):
         self.tips = None
         self.frames = None
         self.spy_window = None
-        self.saved_normal_geometry = None
+        self.initial_fullscreen_state = None
 
         self.widgets = None
         self.dockwidgets = None
@@ -1208,10 +1208,14 @@ class AnimatedTour(QWidget):
 
     def start_tour(self):
         """ """
-        if not self.spy_window.fullscreen_flag:
-            self.saved_normal_geometry = self.spy_window.saveGeometry()
-            self.spy_window.showMaximized()
 
+        if self.spy_window.fullscreen_flag:
+            print("STUPID .l.")
+            self.initial_fullscreen_state = True
+            self.spy_window.toggle_fullscreen()
+            #self.setWindowState(Qt.WindowNoState)
+        self.spy_window.save_current_window_settings('layout_current_temp/', section="quick_layouts")
+        self.spy_window.quick_layout_switch('default')
         geo = self.parent.geometry()
         x, y, width, height = geo.x(), geo.y(), geo.width(), geo.height()
 #        self.parent_x = x
@@ -1247,10 +1251,11 @@ class AnimatedTour(QWidget):
             pass
 
         self.is_running = False
-
-        if self.saved_normal_geometry: 
-            self.spy_window.restoreGeometry(self.saved_normal_geometry)
-            self.saved_normal_geometry = None
+        self.spy_window.quick_layout_switch('current_temp')
+        if self.initial_fullscreen_state: 
+            self.spy_window.toggle_fullscreen()
+            #self.setWindowState(Qt.WindowFullScreen)
+            self.initial_fullscreen_state = None
 
     def hide_tips(self):
         """Hide tips dialog when the main window loses focus."""
--- a/spyder/app/tour.py
+++ b/spyder/app/tour.py
@@ -908,6 +908,7 @@ class AnimatedTour(QWidget):
         self.tips = None
         self.frames = None
         self.spy_window = None
+        self.saved_normal_geometry = None
 
         self.widgets = None
         self.dockwidgets = None
@@ -1207,6 +1208,10 @@ class AnimatedTour(QWidget):
 
     def start_tour(self):
         """ """
+        if not self.spy_window.fullscreen_flag:
+            self.saved_normal_geometry = self.spy_window.saveGeometry()
+            self.spy_window.showMaximized()
+
         geo = self.parent.geometry()
         x, y, width, height = geo.x(), geo.y(), geo.width(), geo.height()
 #        self.parent_x = x
@@ -1242,6 +1247,10 @@ class AnimatedTour(QWidget):
             pass
 
         self.is_running = False
+
+        if self.saved_normal_geometry: 
+            self.spy_window.restoreGeometry(self.saved_normal_geometry)
+            self.saved_normal_geometry = None
 
     def hide_tips(self):
         """Hide tips dialog when the main window loses focus."""
@@ -1295,7 +1304,8 @@ class AnimatedTour(QWidget):
     def any_has_focus(self):
         """Returns if tour or any of its components has focus."""
         f = (self.hasFocus() or self.parent.hasFocus() or
-             self.tips.hasFocus() or self.canvas.hasFocus())
+             self.tips.hasFocus() or self.canvas.hasFocus() or
+             self.tips.isActiveWindow())
         return f
 
 # ----------------------------------------------------------------------------
--- a/spyder/app/tour.py
+++ b/spyder/app/tour.py
@@ -1209,11 +1209,9 @@ class AnimatedTour(QWidget):
     def start_tour(self):
         """ """
 
-        if self.spy_window.fullscreen_flag:
-            print("STUPID .l.")
+        if self.spy_window.isFullScreen():
             self.initial_fullscreen_state = True
-            self.spy_window.toggle_fullscreen()
-            #self.setWindowState(Qt.WindowNoState)
+            self.spy_window.showMaximized()
         self.spy_window.save_current_window_settings('layout_current_temp/', section="quick_layouts")
         self.spy_window.quick_layout_switch('default')
         geo = self.parent.geometry()
@@ -1253,8 +1251,7 @@ class AnimatedTour(QWidget):
         self.is_running = False
         self.spy_window.quick_layout_switch('current_temp')
         if self.initial_fullscreen_state: 
-            self.spy_window.toggle_fullscreen()
-            #self.setWindowState(Qt.WindowFullScreen)
+            self.spy_window.showFullScreen()
             self.initial_fullscreen_state = None
 
     def hide_tips(self):
--- a/spyder/app/tour.py
+++ b/spyder/app/tour.py
@@ -52,7 +52,8 @@ class SpyderWidgets(object):
     # Panes
     ipython_console = 'ipyconsole'
     editor = 'editor'
-    editor_line_number_area = 'editor.get_current_editor().linenumberarea'
+    editor_line_number_area = (
+        'editor.get_current_editor().panels._panels.get(1).values()')
     editor_scroll_flag_area = 'editor.get_current_editor().scrollflagarea'
     file_explorer = 'explorer'
     help_plugin = 'help'
@@ -510,14 +511,23 @@ class FadingCanvas(FadingDialog):
             self.path_current = self.path_full
 
         if self.decoration is not None:
-            for widget in self.decoration:
+            for widgets in self.decoration:
+                if isinstance(widgets, QWidget):
+                    widgets = [widgets]
+                geoms = []
+                for widget in widgets:
+                    widget.raise_()
+                    widget.show()
+                    geo = widget.frameGeometry()
+                    width, height = geo.width(), geo.height()
+                    point = widget.mapTo(self.parent, QPoint(0, 0))
+                    x, y = point.x(), point.y()
+                    geoms.append((x, y, width, height))
+                x = min([geom[0] for geom in geoms])
+                y = min([geom[1] for geom in geoms])
+                width = max([geom[0]+geom[2] for geom in geoms])-x
+                height = max([geom[1]+geom[3] for geom in geoms])-y
                 temp_path = QPainterPath()
-                widget.raise_()
-                widget.show()
-                geo = widget.frameGeometry()
-                width, height = geo.width(), geo.height()
-                point = widget.mapTo(self.parent, QPoint(0, 0))
-                x, y = point.x(), point.y()
                 temp_path.addRect(QRectF(x, y, width, height))
 
                 temp_region_1 = QRegion(x-1, y-1, width+2, height+2)
@@ -1029,19 +1039,22 @@ class AnimatedTour(QWidget):
         dockwidgets = []
 
         for name in names:
-            base = name.split('.')[0]
             try:
-                temp = getattr(spy_window, base)
+                base = name.split('.')[0]
+                try:
+                    temp = getattr(spy_window, base)
+                except AttributeError:
+                    for item in spy_window.thirdparty_plugins:
+                        if type(item).__name__ == name:
+                            temp = item
+                            break
+                else:
+                    # Check if it is the current editor
+                    if 'get_current_editor()' in name:
+                        temp = temp.get_current_editor()
+                        temp = getattr(temp, name.split('.')[-1])
             except AttributeError:
-                for item in spy_window.thirdparty_plugins:
-                    if type(item).__name__ == name:
-                        temp = item
-                        break
-            else:
-                # Check if it is the current editor
-                if 'get_current_editor()' in name:
-                    temp = temp.get_current_editor()
-                    temp = getattr(temp, name.split('.')[-1])
+                temp = eval(f"spy_window.{name}")
 
             widgets.append(temp)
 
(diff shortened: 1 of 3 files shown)


juanis2112:
The lines I added for the new panes in the tour match the length of the ones before which is causing the pep8 warnings.

spyder/app/tour.py
```diff
@@ -94,6 +99,7 @@ def get_tour(index):
     """
     sw = SpyderWidgets
     qtconsole_link = "https://qtconsole.readthedocs.io/en/stable/index.html"
+    ending_link = "https://docs.spyder-ide.org/current/index.html"
```
CAM-Gerlach, line 102: We should add `__doc_url__ = https://docs.spyder-ide.org/` to `spyder/__init__.py` and reference it here and in the relevant menu item in MainWindow (and use it as the base URL for other links to the docs), instead of scattering multiple references to it throughout the code, to avoid bugs, keep things clean and easy to follow, and make it much simpler to change in the future.

spyder/app/tour.py
```diff
@@ -1164,6 +1209,26 @@ def set_tour(self, index, frames, spy_window):
...
+                QMessageBox.information(self, msg_title, msg,
+                                        QMessageBox.Ok)
+                return
+            else:
+                if self.spy_window.fullscreen_flag:
+                    self.spy_window.toggle_fullscreen()
+                else:
+                    self.spy_window.setWindowState(
+                        self.spy_window.windowState()
+                        & (~ Qt.WindowFullScreen))
```
CAM-Gerlach, line 1228: On second thought, this got so long that I think we should put it into another method, e.g. `self._handle_fullscreen()`, which returns `False` if `start_tour` should re-enable window updates and exit (i.e. on macOS) and `True` otherwise.

spyder/app/tour.py
```diff
@@ -213,6 +226,37 @@ def get_tour(index):
...
+                           "your programs by detecting style issues, bad practices "
+                           "and potential bugs."),
+              'widgets': [sw.code_analysis],
+              'interact': True},
+
+             {'title': _("The end"),
+              'content': _("You have reached the end of our tour, you are ready to "
+                           "start using Spyder! If you want more information go to "
+                           "our  <a href=\"{0}\">documentation</a>.<br><br>"
+                           ).format(ending_link),
```
CAM-Gerlach, line 256:
```suggestion
              'content': _('You have reached the end of our tour and are ready to '
                           'start using Spyder! For more information, check out '
                           f'our <a href="{ending_link}">documentation</a>.<br><br>'),
```

Fix grammar and punctuation errors; avoid unnecessary escaping of quotes per PEP 8, simplify and clarify string formatting.

CAM-Gerlach (changes requested):
Great job @juanis2112 , this is a huge improvement to the tour! I just had a couple comments and suggestions and this should be ready to go. Thanks!

ccordoba12 (commented):
This is a terrific improvement! Thanks a lot @juanis2112 for your hard work on this one!

I left some minor style comments, but I found an important bug while testing in Binder: if you press the `Run code` button in the `IPython console` card, that card is hidden and you don't know how to continue the tour.

Selección_012

We could consider to remove that button if that bug is too hard to fix (given that all other panes have no content).

Besides, even if this is not a bug you introduced in this PR, I think it'd be nice to fix the red square that highlights the panel area in the editor:

Selección_013

spyder/app/tour.py
```diff
@@ -213,6 +226,37 @@ def get_tour(index):
                            "the Python and IPython consoles."),
```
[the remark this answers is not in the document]
CAM-Gerlach: To be a little more clear what the plural (previously a relic of the fact there were two different types of consoles) is referring to, I suggest simply:

```suggestion
                           "any IPython console."),

```

spyder/app/tour.py
```diff
@@ -1164,6 +1209,26 @@ def set_tour(self, index, frames, spy_window):
 
     def start_tour(self):
         """ """
+        self.spy_window.setUpdatesEnabled(False)
+        if self.spy_window.isFullScreen() or self.spy_window.fullscreen_flag:
+            if sys.platform == 'darwin':
+                self.spy_window.setUpdatesEnabled(True)
+                msg_title = _("Request")
+                msg = _("To run the tour, please press the green button in the top left corner of "
+                        "the screen to take Spyder out of fullscreen mode.")
```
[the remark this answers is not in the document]
CAM-Gerlach: In conventional English usage, it would be "round green button".

However, I'm not sure how "round" makes it any clearer that it is not part of the Spyder interface, and Spyder doesn't have any green buttons, round or otherwise, in that area of the default interface.

If you do want to be more explicit on that point. we could say something like "please press the green button on the left of the Spyder window's title bar, visible when moving your mouse to the top of the screen." but I'm not sure its necessary or really makes it clearer what to do as opposed to just complicating things further. I don't think don't need to really need to know whether its part of the Spyder interface, just where to find it.

CAM-Gerlach:
@ccordoba12 Is reviewing the existing tour content in scope for this PR? I found a handful of grammar, punctuation and phrasing errors but since Juanita didn't touch them, I didn't (and couldn't, under Github's comment system) suggest them here.

> I found an important bug while testing in Binder: if you press the `Run code` button in the `IPython console` card, that card is hidden and you don't know how to continue the tour.

This issue happens on Windows under `4.x` and is in fact fixed on this PR due to the additional focus check that was added, but presumably due to different focus semantics it still occurs on Linux. However, this can be resolved by adding `self.isActiveWindow()` to the focus check, either in `any_has_focus()` if the problem doesn't occur on Mac, or to `tour_has_focus()` if it does (where it would actually be more serious if it does, since it would not just temporarily hide the tip box, but in fact end the tour completely).

> Besides, even if this is not a bug you introduced in this PR, I think it'd be nice to fix the red square that highlights the panel area in the editor:

My testing confirms this was not introduced in this PR, it is apparently due to the changes made to the left panels in Spyder 4 to add the debug panel, whereas before it was just the line number panel (which is what is being used to determine the region to decorate). Is there a widget or similar meta-object that could be targeted instead that incorporates the area of both panels?

If not, fixing this shouldn't be too terribly difficult; you could pass a nested list of multiple widgets to decorate and and `FadingCanvas.update_canvas()` could do some relatively straightforward math to calculate the x/y origin and width/height of their combined area (`x =min(x_1, x_2, ... x_n)`, `width = max(x_1 + width_1 ... x_n + width_n) - x` and the same for y/height).

spyder/app/tour.py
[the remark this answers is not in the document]
ccordoba12: > Spyder doesn't have any green buttons, round or otherwise, in that area of the default interface.

Sure there is: the Run button. That's why I wanted to clarify what green button we're talking about.

> If you do want to be more explicit on that point. we could say something like "please press the green button on the left of the Spyder window's title bar, visible when moving your mouse to the top of the screen."

I like your suggestion until "title bar", so I changed mine to use yours instead.

ccordoba12:
> @ccordoba12 Is reviewing the existing tour content in scope for this PR? I found a handful of grammar, punctuation and phrasing errors but since Juanita didn't touch them, I didn't (and couldn't, under Github's comment system) suggest them here.

Let's leave that for another PR because this one is focused on fixing the many technical bugs the tour had (I left my suggestion above about not mentioning the Python console because I could do it as a suggestion).

> Is there a widget or similar meta-object that could be targeted instead that incorporates the area of both panels?

I think `PanelsManager` (present in `spyder/plugins/editor/panels/manager.py`) should easily give you the size of the left panel in the editor.

spyder/app/tour.py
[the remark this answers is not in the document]
CAM-Gerlach: > Sure there is: the Run button. That's why I wanted to clarify what green button we're talking about.

That's in the top center for me, but its conceivable it could be different at different screen widths or if people re-arrange their toolbars.

> I like your suggestion until "title bar", so I changed mine for yours instead.

Sounds good to me :+1:

CAM-Gerlach:
> Let's leave that for another PR because this one is focused on fixing the many technical bugs the tour had (I left my suggestion above about not mentioning the Python console because I could do it as a suggestion).

Fine with me :+1: @juanis2112 do you want to do the PR as well, or would you like me to?

> I think `PanelsManager` (present in `spyder/plugins/editor/panels/manager.py`) should easily give you the size of the left panel in the editor.

Okay, thanks. There doesn't appear to be an object that implements the methods of the `QWidget` API that `update_canvas()` expects, but `editor.get_current_editor().panels._panels[spyder.api.panel.Panel.Position.LEFT]` is a list of the panel widgets (exactly what my above suggested approach requires), so setting `decoration` to that instead of just the line number panel and adding a few lines in `FadingCanvas.update_canvas()` calculate the bounding box via the above math if a list of widgets is passed should do the trick using the existing functionality, without any bespoke code or special cases.

CAM-Gerlach:
There's also another bug, which again was not introduced by this PR: clicking the URLs in the tour, on both Linux (Binder) and Windows, appears to have no effect, including both the Docs link added here and the IPython console link untouched from the previous, which has the same behavior in `4.x`. My cursor changes to the pointing hand on mouseover, and I can right click, copy the link location and paste it in my web browser, but left-clicking it does nothing at all. Not sure what's going on; If this is not an easy fix we can leave it for later. @ccordoba12 any ideas?

ccordoba12:
> @ccordoba12 any ideas? Maybe the click event isn't getting send to the appropriate handler?

Yep, you need to call the following method in the appropriate label:

    label.setOpenExternalLinks(True)

That should be enough.

juanis2112:
@ccordoba12 and @CAM-Gerlach this is ready for another review

spyder/app/tour.py
```diff
@@ -103,6 +103,11 @@ def get_tour(index):
     qtconsole_link = "https://qtconsole.readthedocs.io/en/stable/index.html"
     #docs_link = __docs_url__
     docs_link = "https://docs.spyder-ide.org/current/index.html"
+    button_text = ""
+    if sys.platform != "darwin":
+        button_text = ("Please click on the button below to run some simple "
+                       "code in this console. This will be useful to show "
+                       "you other important features.")
```
ccordoba12, line 110: In order to have a common experience in all platforms, I think we should remove this button in all of them (instead of hiding it in macOS).

spyder/app/tour.py
```diff
@@ -782,8 +783,11 @@ def set_data(self, title, content, current, image, run, frames=None,
         if run is None:
             self.button_run.setVisible(False)
         else:
-            self.button_run.setDisabled(False)
-            self.button_run.setVisible(True)
+            if sys.platform == "darwin":
+                self.button_run.setVisible(False)
+            else:
+                self.button_run.setDisabled(False)
+                self.button_run.setVisible(True)
```
ccordoba12, line 790: Instead of hiding `button_run`, let's simply remove it. I think it won't be possible for us to show it in the future due to the problems it introduces in macOS.

spyder/app/tour.py
```diff
@@ -510,14 +511,23 @@ def update_canvas(self):
...
+                    widget.show()
+                    geo = widget.frameGeometry()
+                    width, height = geo.width(), geo.height()
+                    point = widget.mapTo(self.parent, QPoint(0, 0))
+                    x, y = point.x(), point.y()
+                    geoms.append((x, y, width, height))
+                x = min([geom[0] for geom in geoms])
+                y = min([geom[1] for geom in geoms])
+                width = max([geom[0]+geom[2] for geom in geoms])-x
+                height = max([geom[1]+geom[3] for geom in geoms])-y
```
ccordoba12, line 530:
```suggestion
                width = max([geom[0] + geom[2] for geom in geoms]) - x
                height = max([geom[1] + geom[3] for geom in geoms]) - y
```

Code is easier to read with spaces around math signs (even if it takes more than 79 columns).

spyder/app/tour.py
```diff
@@ -94,6 +101,13 @@ def get_tour(index):
     """
     sw = SpyderWidgets
     qtconsole_link = "https://qtconsole.readthedocs.io/en/stable/index.html"
+    #docs_link = __docs_url__
+    docs_link = "https://docs.spyder-ide.org/current/index.html"
```
ccordoba12, line 105:
```suggestion
    docs_link = __docs_url__
```

I think you forgot to make this change.

ccordoba12 (commented):
Thanks a lot @juanis2112 for your continued work on this one! I think it's almost ready!

spyder/app/tour.py
CAM-Gerlach: Disabling the button also means the Console, Variable Explorer, History Log, etc. won't have anything in them to show users if the user hasn't already manually run code that session, which the tour currently refers to (since the bug appears to be triggered by the actual code execution, not the button press, so we can't just run the code automatically). Therefore, this is a significant downgrade in functionality and experience for the 80-90% of users not on Mac for the sake of a bug (possibly fixable in the future) that affects 10-20% that are.

There are a number of features we disable or enable on specific platforms due to bugs or otherwise (e.g. the fullscreen mode, which affects this very PR), and I'm not sure I really see the value in making the presence of the button consistent across platforms that would outweigh this, since the tour is self-contained and self-documenting with the text adapting according to whether the button is there, and perhaps unlike the regular UI, a user is very unlikely to run the same tour on multiple platforms to even notice it is different.

spyder/app/tour.py
CAM-Gerlach: I don't agree with removing it in all platforms for the sake of one (see above), but if we do, we should remove the `run` key from the get_tour function (which prevents the button from being shown or code being run from this tour) and revert this change, rather than removing an otherwise working feature completely (and breaking the other tours too). We were almost able to fix/work around this bug but not quite, so it might well be possible to resolve in the future either with action ourselves or with dependency (Qt, etc) updates, and given the macOS transition to ARM, removal of software development features and shift away from the professional space, we may not support it in the future so I don't think it makes sense to throw out this otherwise useful feature completely.

spyder/app/tour.py
CAM-Gerlach: The line can be broken before the `for` to avoid PEP 8 issues

spyder/app/tour.py
```diff
@@ -50,12 +52,17 @@ class SpyderWidgets(object):
     # Panes
     ipython_console = 'ipyconsole'
     editor = 'editor'
-    editor_line_number_area = 'editor.get_current_editor().linenumberarea'
+    editor_line_number_area = (
+        'editor.get_current_editor().panels._panels.get(1).values()')
```
CAM-Gerlach, line 56:
```suggestion
    panel = Panel.Position.LEFT
    editor_line_number_area = (
        f'editor.get_current_editor().panels._panels[{panel}].values()')
```

Get actual panel ID value instead of hardcoding it; use shorter and more conventional dict access (which fails early)

spyder/app/tour.py
CAM-Gerlach:
```suggestion
```

I suggest simply removing it and using `__docs_url__` directly, it just adds extra complexity to redefine it here since its only used one place in this file.

spyder/app/tour.py
```diff
@@ -210,9 +229,40 @@ def get_tour(index):
...
+              'content': _("The Code Analysis helps you improve the quality of "
+                           "your programs by detecting style issues, bad practices "
+                           "and potential bugs."),
+              'widgets': [sw.code_analysis],
+              'interact': True},
+
+             {'title': _("The end"),
+              'content': _('You have reached the end of our tour and are ready to '
+                           'start using Spyder! For more information, check out '
+                           f'our <a href="{docs_link}">documentation</a>.<br><br>'),
```
CAM-Gerlach, line 262:
```suggestion
                           f'our <a href="{__docs_url__}">documentation</a>.<br><br>'),
```

Per above

spyder/app/tour.py
```diff
@@ -982,13 +1044,22 @@ def _process_widgets(self, names, spy_window):
...
-                temp = getattr(temp, name.split('.')[-1])
+            try:
+                base = name.split('.')[0]
+                try:
+                    temp = getattr(spy_window, base)
+                except AttributeError:
+                    for item in spy_window.thirdparty_plugins:
+                        if type(item).__name__ == name:
+                            temp = item
+                            break
```
CAM-Gerlach, line 1055:
```suggestion
                            break
                    else:
                        temp = None
```

We need this so that the `AttributeError` passes through to the outer except block if the item isn't found in this list.

spyder/app/tour.py
```diff
@@ -982,13 +1044,22 @@ def _process_widgets(self, names, spy_window):
...
-            temp = getattr(spy_window, base)
-
-            # Check if it is the current editor
-            if 'get_current_editor()' in name:
-                temp = temp.get_current_editor()
-                temp = getattr(temp, name.split('.')[-1])
+            try:
+                base = name.split('.')[0]
+                try:
+                    temp = getattr(spy_window, base)
```
CAM-Gerlach, line 1050:
```suggestion
                    temp = getattr(spy_window, name)
```

So that execution passes properly to the outer `except` block if `name` isn't a simple attribute (it only happened to work as is because `editor` happened to be part of the particular string we passed).

spyder/app/tour.py
```diff
@@ -982,13 +1044,22 @@ def _process_widgets(self, names, spy_window):
...
+                except AttributeError:
+                    for item in spy_window.thirdparty_plugins:
+                        if type(item).__name__ == name:
+                            temp = item
+                            break
+                else:
```
CAM-Gerlach, line 1056:
```suggestion

```

To correspond to the changes above, this needs to be part of the except block

spyder/app/tour.py
```diff
@@ -982,13 +1044,22 @@ def _process_widgets(self, names, spy_window):
...
+                    # Check if it is the current editor
+                    if 'get_current_editor()' in name:
+                        temp = temp.get_current_editor()
+                        temp = getattr(temp, name.split('.')[-1])
```
CAM-Gerlach, line 1060:
```suggestion
                        temp = getattr(temp, name.split('.')[-1])
                    if temp is None:
                        raise
```

Re-raise the error so we get to the outer except block as we should

spyder/app/tour.py
```diff
@@ -1162,8 +1231,35 @@ def set_tour(self, index, frames, spy_window):
...
+                          "it out of fullscreen mode.")
+                  QMessageBox.information(self, msg_title, msg,
+                                          QMessageBox.Ok)
+                  return True
+              if self.spy_window.fullscreen_flag:
+                  self.spy_window.toggle_fullscreen()
+              else:
+                  self.spy_window.setWindowState(
+                      self.spy_window.windowState()
+                      & (~ Qt.WindowFullScreen))
```
CAM-Gerlach, line 1250:
```suggestion
            if sys.platform == 'darwin':
                self.spy_window.setUpdatesEnabled(True)
                msg_title = _("Request")
                msg = _("To run the tour, please press the green button on "
                        "the left of the Spyder window's title bar to take "
                        "it out of fullscreen mode.")
                QMessageBox.information(self, msg_title, msg,
                                        QMessageBox.Ok)
                return True
            if self.spy_window.fullscreen_flag:
                self.spy_window.toggle_fullscreen()
            else:
                self.spy_window.setWindowState(
                    self.spy_window.windowState()
                    & (~ Qt.WindowFullScreen))
```

There's an extra indent here. I suggest setting your editor to 4 space indents for Python files to avoid this problem and the extra work and bugs of manual indents.

spyder/app/tour.py
```diff
@@ -1162,8 +1231,35 @@ def set_tour(self, index, frames, spy_window):
...
+        return False
+  
     def start_tour(self):
         """ """
+        self.spy_window.setUpdatesEnabled(False)
+        if self._handle_fullscreen():
+            return
+        self.spy_window.save_current_window_settings(
+            'layout_current_temp/',
+            section="quick_layouts"
```
CAM-Gerlach, line 1260:
```suggestion
            section="quick_layouts",
```

spyder/app/tour.py
```diff
@@ -1162,8 +1231,35 @@ def set_tour(self, index, frames, spy_window):
...
+                                          QMessageBox.Ok)
+                  return True
+              if self.spy_window.fullscreen_flag:
+                  self.spy_window.toggle_fullscreen()
+              else:
+                  self.spy_window.setWindowState(
+                      self.spy_window.windowState()
+                      & (~ Qt.WindowFullScreen))
+        return False
+
```
CAM-Gerlach, line 1252:
```suggestion
```

CAM-Gerlach (changes requested):
This looks good; I'll do a final functional test once all suggestions are applied. Revised a few of Carlos' suggestions, resolved some execution flow issues with the try-except block for the attribute lookup, and fixed the PEP 8 issues PEP8Speaks is complaining about (one of them something that Python itself will warn about, indentation)

spyder/app/tour.py
ccordoba12: Let's see what @juanis2112 thinks about this one. If she considers we should leave this button in Windows and Linux, then I'll defer to your decision.

My main concern is this could be confusing for students in classrooms if teachers tell them: "and now press the Run code button" because that button won't be available in macOS.

spyder/app/tour.py
ccordoba12: Ok, fair enough.

spyder/app/tour.py
CAM-Gerlach: > My main concern is this could be confusing for students in classrooms if teachers tell them: "and now press the Run code button" because that button won't be available in macOS.

I see, that's a good point. I'm not sure how likely it is that a teacher would have students simply step through the tour in class as opposed to showing them things themselves, having them complete tasks or do it on their own (since it doesn't require the user to actually do anything but click a few buttons; at least the professors I've known that teach with Spyder haven't). And in any case, students using Macs in classes where professors do that is a very small percentage of Spyder's userbase.

But if we want to avoid the chance of this confusion but not strip this feature from the vast majority of the Spyder userbase, instead of simply hiding the button we could disable it on macOS with a tooltip saying "Not available on macOS" or similar?

ccordoba12: > But if we want to avoid the chance of this confusion but not strip this feature from the vast majority of the Spyder userbase, instead of simply hiding the button we could disable it on macOS with a tooltip saying "Not available on macOS" or similar?

That's a pretty good compromise, thanks for the suggestion @CAM-Gerlach! I totally agree with it.

CAM-Gerlach: Okay, sure. I'll make the appropriate suggestion below.

spyder/app/tour.py
CAM-Gerlach: Per our consensus above. @juanis2112 if you have a better idea for the tooltip text, I'll defer to whatever you come up with.

```suggestion
            self.button_run.setVisible(True)
            self.button_run.setDisabled(sys.platform == "darwin")
            if sys.platform == "darwin":
                self.button_run.setToolTip("Not available on macOS")
```

CAM-Gerlach:
Once the final review comments are addressed, I'll do one last functional test on Windows to ensure we didn't break anything and approve so we can merge this.

CAM-Gerlach (approved):
Everything works great now in my testing on Windows in fullscreen, normal and maximized, checking all the buttons, steps and functionality, and running the tour multiple times. Once we confirm everything works similarly on Linux and macOS, this is good to merge from my end. Thanks again for all your hard work @juanis2112 !

ccordoba12 (approved):
Looks great to me too! Thanks @juanis2112!

I also confirm everything is working smoothly on Linux.

juanis2112:
To explain my last commit: For some reason @CAM-Gerlach suggestion for disabling the button didn't work so I had to move the method for disabling it to _enable_widgets method where the other buttons get disabled too. 

Here is a screenshot with the last commit. Please @ccordoba12 and @CAM-Gerlach test that this change doesn't break anything on windows. or linux. 

Screenshot 2020-10-09 at 1 49 05 PM

juanis2112:
Apart from that minor thing, I tested this again in Mac and it works as expected.

CAM-Gerlach:
> For some reason @CAM-Gerlach suggestion for disabling the button didn't work so I had to move the method for disabling it to _enable_widgets method where the other buttons get disabled too.

Probably because something else either got called later in the setup sequance or on the tour step with a run attribute that reset/re-enabled it.

ccordoba12 (approved):
Thanks @juanis2112, I retested things again on Linux and everything is working fine after your last change.

CAM-Gerlach (approved):
Re-tested on Windows and Binder (Linux) and everything still works as expected, thanks. (I couldn't actually test the external link on Binder since the Binder web browser didn't work for some reason, but it works on Windows).

ccordoba12:
> I couldn't actually test the external link on Binder since the Binder web browser didn't work for some reason, but it works on Windows

That's because there's no default browser on Binder and no easy way to do declare one.

PR #13828, opened 2020-09-21
PR: Don't send requests to the PyLS until it's been properly initialized

### LINKED ISSUE #13351: Some issues with Spyder projects and the workspace functionality in the PyLS
## Problem Description

When launching Spyder with a previously open project, jedi settings are not passed to the python language server workspace. This may result in incomplete or failed command completions.

1. launch Spyder (bootstrap) with a previously open project from an environment without a desired package (e.g. matplotlib) but have a different iPython Console environment that does have the desired package as the default python environment.
2. A minimal code file in the editor will illustrate the issue, e.g.
```python
plt.
```
3. tab completions do not work
4. closing the project restores completions. Also, reopening the project restores completions.

Upon investigation I found that `pyls.workspace.Workspace._config.plugin_settings('jedi',...)` is empty until the project is closed or reopened.
I suspect that the pyls workspace is created before the jedi settings are read/configured upon launching Spyder with an open project.

This is not an issue when launching Spyder without an open project.
Restarting the LSP server has no effect.

I'm not sure where to look for the offending code...
Any help would be appreciated.

ccordoba12:
- In previous versions we were sending some requests before the server responded to our `initialize` request, which is not expected according to [the protocol](https://microsoft.github.io/language-server-protocol/specification#initialize). That was causing problems when reloading the previous active project at startup.
- This also fixes the ordering of requests we send to the LSP server at startup.
- The main visible change of this PR is the addition of a spinner in the LSP status bar widget, to improve how we tell users that the LSP is still starting even after the window is fully loaded.

    Peek 07-10-2020 16-40

- I also decided to hide the splash screen in `post_visible_setup` so it remains visible until the very last second before we show the main window.

Fixes #13351.

TODO:

- [x] Add some tests to prevent regressions.
- [x] Fix scrolling in the internal console.


jnsebgosselin:
Hey @ccordoba12, I've tested this locally and this doesn't solve https://github.com/spyder-ide/spyder/issues/13897.

When starting Spyder, the outline explorer only populates for the file that is focused. When switching file, the tree for that file is not populated automatically. A change needs to be made to the file to trigger the update.

outline_explorer_dont_populate_multiplefiles

jnsebgosselin:
> 
> 
> > When starting Spyder, the outline explorer only populates for the file that is focused. When switching file, the tree for that file is not populated automatically. A change needs to be made to the file to trigger the update.
> 
> As I told @jnsebgosselin on issue #13928, that bug is caused by another reason. I'm preparing another PR with a fix for it.

Ok, I understand. Sorry for posting that here. I am a little bit confused since what I reported in issue #13928 isn't related to the problem I showed in https://github.com/spyder-ide/spyder/pull/13828#issuecomment-704253703, but I'm happy if this is taken care of.

spyder/plugins/completion/languageserver/widgets/status.py
```diff
@@ -25,25 +25,44 @@
 logger = logging.getLogger(__name__)
 
 
+# Main constants
+class ClientStatus:
```
dalthviz, line 29: @ccordoba12 maybe handling the dict with status translation in this class could make things easier to follow. Maybe the `ClientStatus` class could look something like this:

```python
class ClientStatus:
    STATUS = 'LSP {}: {}'
    STARTING = 'starting'
    READY = 'ready'
    RESTARTING = 'restarting'
    DOWN = 'down'
    STRINGS_FOR_TRANSLATION = {
        STARTING: _("starting"),
        READY: _("ready"),
        RESTARTING: _("restarting"),
        DOWN: _("down")
    }
```

dalthviz (commented):
Thanks @ccordoba12 left a comment otherwise this LGTM :+1:

spyder/plugins/completion/languageserver/widgets/status.py
ccordoba12: Good suggestion @dalthviz!

dalthviz (approved):
Thanks @ccordoba12 LGTM :+1:

PR #13833, opened 2020-09-22
PR: Fix some UX/UI issues in the Files pane

### LINKED ISSUE #13465: Fix some further issues with the Files pane UX
## Problem Description

After testing the new changes proposed in #13179 and executed in #13209 as part of reviewing spyder-ide/spyder-docs#146 , I found a few remaining issues that should be resolved. The changes should be simple and the implementation very straightforward, closely following that already reviewed #13209 , so I can have a PR ready as soon as we agree on them, which can hopefully be discussed, reviewed and merged in short order.

They are:

* [ ] ~~Rename the option `Show all files`, enabled by default, to be `Show only files matching patterns`, disabled by default, making it much clearer what this option does and what files _wouldn't_ be shown. This was very confusing to me as a user, and that was _before_ I tested the latest changes with the `Show hidden files` option, which makes what `Show all files` does even more befuddling. I would also move this option to be directly above the option to edit filename filters, so its more clearly connected with that.~~ Simply remove the option, have filters always on if any are present, and clear the filter list by default.
* [ ] Rename the option `Show filenames with these extensions...` to `Edit filename filters...` to avoid misleading users that only the extensions are matched, since the patterns can match anywhere in the name, and in fact a third of the default patterns don't involve extensions at all. It also links it more clearly with the option that enables/disables the filter.
* [ ] Make the window title of the filename filter dialog match the options menu text (it currently is stuck with the old name) and add a clearer description, e.g. `Enter the glob patterns, separated by commas, at least one of which a filename must match to be shown if any have been entered.`
* [ ] Modify the preferences accordingly and add a section around it
* [ ] Change `Python script` to `Python file` in the `New` submenu, since the file can be either a script, a module or both.

image

@ccordoba12 , am I good to go ahead with a PR with these changes? Thanks!

steff456:
## Description of Changes

Create the changes discussed over the file explorer pane,

- [x] Add filter icon as button on toolbar (needs the blue outline pressed state)

- [x] Add tooltip for new button (maybe “Activate and deactivate filter”)

- [x] Change hamburger menu description to “Edit filter settings” or something similar (this menu item is now the way to open the dialog)

- [x] Remove “Show all files” option in the hamburger menu

- [x] Add “Return to default” button in the filter settings dialog 

- [x] Change Python script to Python file in the New submenu

- [x] Make the window title of the filename filter dialog match the options menu text 

- [x] Add a description in the filter dialog

- [x] Change the filter dialog to a non blocking dialog

Pane layout:
image

Filter dialog:
image

Hamburger menu:
image

Context menu:
image

Preferences config page:
image

### Issue(s) Resolved

Fixes #13465 

### Affirmation

### DIFF
--- a/spyder/plugins/explorer/confpage.py
+++ b/spyder/plugins/explorer/confpage.py
@@ -7,11 +7,13 @@
 """File explorer configuration page."""
 
 # Third party imports
-from qtpy.QtWidgets import QTabWidget, QVBoxLayout, QWidget
+from qtpy.QtWidgets import (QTabWidget, QVBoxLayout, QWidget, QGroupBox,
+                            QLabel, QPushButton)
 
 # Local imports
 from spyder.api.preferences import PluginConfigPage
 from spyder.config.base import _
+from spyder.config.main import NAME_FILTERS
 from spyder.py3compat import to_text_string
 from spyder.plugins.explorer.widgets.fileassociations import (
     FileAssociationsWidget)
@@ -25,16 +27,40 @@ class ExplorerConfigPage(PluginConfigPage):
 
         # Widgets
         general_widget = QWidget()
+
+        # General options group
+        basic_group = QGroupBox(_("General options"))
         check_show_hidden_files = newcb(_("Show hidden files"), 'show_hidden')
         check_single_click = newcb(_("Single click to open files"),
                                    'single_click_to_open')
-        edit_filename_filters = self.create_textedit(
-            _("Show files with these extensions:"),
+        basic_layout = QVBoxLayout()
+        basic_layout.addWidget(check_show_hidden_files)
+        basic_layout.addWidget(check_single_click)
+        basic_group.setLayout(basic_layout)
+
+        # Filter options group
+        filter_group = QGroupBox(_("Filter settings"))
+        description_label = QLabel(
+            _('Filter files by name, extension, or more using '
+              '<a href="https://docs.python.org/3/library/glob.html">glob '
+              'patterns.</a> Please enter the glob patterns of the files you '
+              'want to show, separated by commas.'))
+        description_label.setOpenExternalLinks(True)
+        description_label.setWordWrap(True)
+        self.edit_filename_filters = self.create_textedit(
+            '',
             'name_filters',
             tip=("Enter values separated by commas"),
             content_type=list,
         )
-        edit_filename_filters.setEnabled(True)
+        self.edit_filename_filters.setEnabled(True)
+        reset_btn = QPushButton(_("Reset to default values"))
+        reset_btn.clicked.connect(self.reset_to_default)
+        filter_layout = QVBoxLayout()
+        filter_layout.addWidget(description_label)
+        filter_layout.addWidget(self.edit_filename_filters)
+        filter_layout.addWidget(reset_btn)
+        filter_group.setLayout(filter_layout)
 
         associations_widget = QWidget()
         self.edit_file_associations = self.create_textedit(
@@ -49,11 +75,9 @@ class ExplorerConfigPage(PluginConfigPage):
         # The actual config data is stored on this text edit set to invisible
         self.edit_file_associations.setVisible(False)
 
-        # Layout
         layout = QVBoxLayout()
-        layout.addWidget(check_show_hidden_files)
-        layout.addWidget(check_single_click)
-        layout.addWidget(edit_filename_filters)
+        layout.addWidget(basic_group)
+        layout.addWidget(filter_group)
         general_widget.setLayout(layout)
 
         layout_file = QVBoxLayout()
@@ -80,3 +104,9 @@ class ExplorerConfigPage(PluginConfigPage):
         """
         textedit = self.edit_file_associations.textbox
         textedit.setPlainText(to_text_string(data))
+
+    def reset_to_default(self):
+        """Reset the filter settings to default."""
+        self.set_option('name_filters', NAME_FILTERS)
+        textedit = self.edit_filename_filters.textbox
+        textedit.setPlainText(", ".join(NAME_FILTERS))
--- a/spyder/plugins/explorer/confpage.py
+++ b/spyder/plugins/explorer/confpage.py
@@ -42,7 +42,7 @@ class ExplorerConfigPage(PluginConfigPage):
         filter_group = QGroupBox(_("Filter settings"))
         description_label = QLabel(
             _('Filter files by name, extension, or more using '
-              '<a href="https://docs.python.org/3/library/glob.html">glob '
+              '<a href="https://en.wikipedia.org/wiki/Glob_(programming)">glob '
               'patterns.</a> Please enter the glob patterns of the files you '
               'want to show, separated by commas.'))
         description_label.setOpenExternalLinks(True)
@@ -53,7 +53,9 @@ class ExplorerConfigPage(PluginConfigPage):
             tip=("Enter values separated by commas"),
             content_type=list,
         )
+
         self.edit_filename_filters.setEnabled(True)
+
         reset_btn = QPushButton(_("Reset to default values"))
         reset_btn.clicked.connect(self.reset_to_default)
         filter_layout = QVBoxLayout()
--- a/spyder/plugins/explorer/confpage.py
+++ b/spyder/plugins/explorer/confpage.py
@@ -25,7 +25,6 @@ class ExplorerConfigPage(PluginConfigPage):
 
         # Widgets
         general_widget = QWidget()
-        check_show_all = newcb(_("Show all files"), 'show_all')
         check_show_hidden_files = newcb(_("Show hidden files"), 'show_hidden')
         check_single_click = newcb(_("Single click to open files"),
                                    'single_click_to_open')
@@ -35,7 +34,7 @@ class ExplorerConfigPage(PluginConfigPage):
             tip=("Enter values separated by commas"),
             content_type=list,
         )
-        edit_filename_filters.setEnabled(not check_show_all.isChecked())
+        edit_filename_filters.setEnabled(True)
 
         associations_widget = QWidget()
         self.edit_file_associations = self.create_textedit(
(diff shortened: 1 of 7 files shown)


steff456:
@ccordoba12, @juanis2112 and @CAM-Gerlach this PR is ready for review :)

juanis2112:
This looks great @steff456. Thanks for taking care of this!

Some minor comments:
1. In the default filters "*.txt" appears twice. 

Screenshot 2020-09-28 at 6 43 02 PM

2. I think the tooltip for the filters should be "Activate filter" when the filter is deactivated and "Deactivate filter" when it is activated. (Similar to the lock/unlock tooltips of the lock icon in the help pane)

3.  I think we said we could add a link to point people to what a "glob pattern" means, so please add this link 
https://docs.python.org/3/library/glob.html to the "glob patterns" text.

4. I like the new message in the filter dialog but maybe we should add something that clarifies the fact that the filters show files instead of hiding them. I would suggest: "Filer files by name, extension, or more using glob patterns. Please enter the glob patterns of the files you want to show, separated by commas". What do you think @isabela-pf?

juanis2112 (changes requested):
This looks great @steff456. Thanks for taking care of this!

Some minor comments:

- [x]  In the default filters "*.txt" appears twice. 

Screenshot 2020-09-28 at 6 43 02 PM

- [x]  I think the tooltip for the filters should be "Activate filter" when the filter is deactivated and "Deactivate filter" when it is activated. (Similar to the lock/unlock tooltips of the lock icon in the help pane)

- [x]  I think we said we could add a link to point people to what a "glob pattern" means, so please add this link 

https://docs.python.org/3/library/glob.html to the "glob patterns" text.

- [x] I like the new message in the filter dialog but maybe we should add something that clarifies the fact that the filters show files instead of hiding them. I would suggest: "Filer files by name, extension, or more using glob patterns. Please enter the glob patterns of the files you want to show, separated by commas". What do you think @isabela-pf?

spyder/plugins/explorer/widgets/explorer.py
```diff
@@ -348,7 +336,7 @@ def setup_common_actions(self):
         )
         # Filters
         self.filters_action = create_action(
-            self, _("Show files with these extensions..."), None,
+            self, _("Edit filter settings"), None,
```
CAM-Gerlach, line 339:
```suggestion
            self, _("Edit filter settings..."), None,
```

Per standard UI convention and as used elsewhere in Spyder, the `Edit filter settings` item should have an ellipses (`...`) at the end as it did previously, indicating that it opens a dialog that requires further interaction.

spyder/plugins/explorer/widgets/explorer.py
```diff
@@ -1681,9 +1694,17 @@ def __init__(self, parent=None, name_filters=['*.py', '*.pyw'],
         button_previous = QToolButton(self)
         button_next = QToolButton(self)
         button_parent = QToolButton(self)
+        self.filter_on = False
+        filter_button = create_toolbutton(
+            self, icon=ima.icon("filter"),
+            tip=_("Activate/deactivate filter"),
```
CAM-Gerlach, line 1700:
```suggestion
            tip=_("Activate/deactivate filename filters"),
```

Since we have plenty of space, I suggest the tooltip read `Activate/deactivate filename filters` to be a little more descriptive about what it does.

spyder/plugins/explorer/widgets/explorer.py
```diff
@@ -405,24 +391,43 @@ def get_file_associations(self, fname):
...
-        """Toggle all files mode"""
-        self.parent_widget.sig_option_changed.emit('show_all', checked)
-        self.show_all = checked
-        self.set_show_all(checked)
+        dialog.setWindowTitle(_('Edit filter settings'))
+
+        # Create dialog contents
+        description_label = QLabel(
+            _('Filter files by name, extension, or more using glob patterns. '
+              'Please enter the glob patterns, separated by commas.'))
```
CAM-Gerlach, line 402:
```suggestion
            _('Filter files by name, extension, or more using glob patterns. '
              'Please enter the glob patterns, separated by commas, '
              at least one of which a file must match to be shown when '
              filters are enabled.'))
```

To address what @juanis2112 brought up and be a little more clear, I propose the above as a suggestion for the dialog text.

CAM-Gerlach (changes requested):
Thanks for your work, @steff456 , this is major improvement in the UX of the files pane and looks really solid. Just a few things (in addition to @juanis2112 's suggestions):

* [ ] The corresponding section in the Files preferences pane still has the old, incorrect description, rather than the new one you added to the dialog, and should be updated accordingly. Its also missing the really nice <kbd>Reset</kbd> button you added to the dialog.
* [ ] Per standard UI convention and as used elsewhere in Spyder, the `Edit filter settings` item should have an ellipses (`...`) at the end, indicating that it opens a dialog that requires further interaction.
* [ ] Since we have plenty of space, I suggest the tooltip read `Activate/deactivate filename filters` to be a little more descriptive about what it does
* [ ] I implemented @juanis2112 's idea as a Github suggestion (I somehow missed that she had a specific proposal for the text the first time around, its way better than what I previously wrote).

Thanks, and looking forward to seeing this in Spyder.

CAM-Gerlach:
Didn't realize I never submitted this, sorry. I agree with @juanis2112 's suggestions as well, with a couple additional followup comments:

> 2. I think the tooltip for the filters should be "Activate filter" when the filter is deactivated and "Deactivate filter" when it is activated. (Similar to the lock/unlock tooltips of the lock icon in the help pane)

Agreed, this would be better. I suggest "filename filters" to be more clear, since we have plenty of space.

> 3. I think we said we could add a link to point people to what a "glob pattern" means, so please add this link: https://docs.python.org/3/library/glob.html to the "glob patterns" text.

We should definitely add a link, but I suggest https://en.wikipedia.org/wiki/Glob_(programming) instead, since its a much friendlier introduction to the subject that explains what globs are and what the characters mean (which is what users need to know to use them); the Python docs just discuss the specific functions and methods of the `glob` module without actually explaining how they work. (I think it might have been my fault for suggesting this one in the first place, sorry)

steff456:
@CAM-Gerlach and @juanis2112 I made all the requested changes, please let me know if there's something else that can be improved :)

CAM-Gerlach (commented):
Great job, thanks @steff456 ! Everything looks good in testing; just one remaining thing that got buried in my non-review comments, assuming @juanis2112 doesn't object to it:

* [ ] I suggest https://en.wikipedia.org/wiki/Glob_(programming) instead for the glob link, since its a much friendlier introduction to the subject that explains what globs are and what the characters mean (which is what users need to know to use them); the Python docs just discuss the specific functions and methods of the glob module without actually explaining how they work. (I think it might have been my fault for suggesting this one in the first place, sorry)

spyder/plugins/explorer/confpage.py
```diff
@@ -25,17 +27,40 @@ def setup_page(self):
...
+        basic_layout = QVBoxLayout()
+        basic_layout.addWidget(check_show_hidden_files)
+        basic_layout.addWidget(check_single_click)
+        basic_group.setLayout(basic_layout)
+
+        # Filter options group
+        filter_group = QGroupBox(_("Filter settings"))
+        description_label = QLabel(
+            _('Filter files by name, extension, or more using '
+              '<a href="https://docs.python.org/3/library/glob.html">glob '
```
CAM-Gerlach, line 45:
```suggestion
              '<a href="https://en.wikipedia.org/wiki/Glob_(programming)">glob '
```

juanis2112: I'm not sure how I feel about pointing to wikipedia instead of python docs. I would prefer leaving the python docs, what do you think @ccordoba12 and @steff456?

CAM-Gerlach: I naturally would prefer the Python docs if practicable (I believe I originally suggested this link), but it doesn't actually explain what a glob is or how to write one, nor does it even include a link to another resource that does so. Almost everything there is about the glob module API (which isn't relevant here), and even the first paragraph is packed with highly technical jargon that puts what I've written to shame. This would be very confusing and intimidating for a newer user who just wants to know what the heck a glob is and how they write one (which is our purpose for including the link in the first place), and so I favor a much friendlier introduction instead (doesn't have to be a Wikipedia article, but something other than the highly technical API docs).

Perhaps @isabela-pf might have valuable input as well?

steff456: I guess that the link that best describes to the user what a glob pattern is and how to create one should be the one we reference to, so I'll say Wikipedia is this case is better.

ccordoba12: I agree with @CAM-Gerlach and @steff456. We should use the Wikipedia link because it really describes how to use glob patterns.

spyder/plugins/explorer/widgets/explorer.py
```diff
@@ -1764,6 +1788,15 @@ def __init__(self, parent=None, name_filters=['*.py', '*.pyw'],
         self.treewidget.set_next_enabled.connect(next_action.setEnabled)
         self.sig_option_changed.connect(self.refresh_actions)
 
+    def change_filter_state(self):
```
ccordoba12, line 1791: Please add a docstring to this method.

ccordoba12 (commented):
This looks pretty good, thanks a lot @steff456 for your work on it!

spyder/plugins/explorer/confpage.py
juanis2112: Okay if you all agree I’m fine with Wikipedia then 👍🏻

juanis2112 (dismissed):
Apart from changing the link to wikipedia and Carlos minor suggestions I think this one is ready! Great improvement to our files pane. Thanks @steff456

ccordoba12 (approved):
Thanks @steff456, it's really great to have these improvements for 4.2!

CAM-Gerlach (approved):
LGTM, thanks @steff456 !

PR #13864, opened 2020-09-26
PR: Synchronize symbols and folding after a timeout (Editor)

ccordoba12:
This avoids requesting symbols and folding after linting results arrive from the PyLS. Instead, it does that after a timeout.

This should avoid the freezes reported on issue #14384.

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -1294,6 +1294,117 @@ class CodeEditor(TextEditBaseWidget):
         self.request_folding()
         self.request_symbols()
 
+    def process_code_analysis(self, diagnostics):
+        """Process code analysis results in a thread."""
+        self.cleanup_code_analysis()
+        self._diagnostics = diagnostics
+
+        # Process diagnostics in a thread to improve performance.
+        self.update_diagnostics = QThread()
+        self.update_diagnostics.run = self.set_errors
+        self.update_diagnostics.finished.connect(self.finish_code_analysis)
+        self.update_diagnostics.start()
+
+    def cleanup_code_analysis(self):
+        """Remove all code analysis markers"""
+        self.setUpdatesEnabled(False)
+        self.clear_extra_selections('code_analysis_highlight')
+        self.clear_extra_selections('code_analysis_underline')
+        for data in self.blockuserdata_list():
+            data.code_analysis = []
+
+        self.setUpdatesEnabled(True)
+        # When the new code analysis results are empty, it is necessary
+        # to update manually the scrollflag and linenumber areas (otherwise,
+        # the old flags will still be displayed):
+        self.sig_flags_changed.emit()
+        self.linenumberarea.update()
+
+    def set_errors(self):
+        """Set errors and warnings in the line number area."""
+        try:
+            self._process_code_analysis(underline=False)
+        except RuntimeError:
+            # This is triggered when a codeeditor instance was removed
+            # before the response can be processed.
+            return
+        except Exception:
+            self.log_lsp_handle_errors("Error when processing linting")
+
+    def underline_errors(self):
+        """Underline errors and warnings."""
+        try:
+            self._process_code_analysis(underline=True)
+        except RuntimeError:
+            # This is triggered when a codeeditor instance was removed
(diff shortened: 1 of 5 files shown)


ccordoba12:
Once an optimal timout is selected, I'll our tests.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -382,6 +382,14 @@ def __init__(self, parent=None):
         self.completion_widget.sig_completion_hint.connect(
             self.show_hint_for_completion)
 
+        # Handle calls to document_did_change when typing text.
+        # See: _handle_keypress_event.
+        self._timer_call_document_did_change = QTimer(self)
+        self._timer_call_document_did_change.setSingleShot(True)
+        self._timer_call_document_did_change.setInterval(250)
```
CAM-Gerlach, line 389: Instead of a magic number embedded in this call, I suggest making it either a class or a module constant and calling it that way.

CAM-Gerlach (commented):
Per my testing on a very short file (temp.py) and a very long file (mainwindow.py), same as before, it does appear to fix the problem:

<details>

4.x typing speed (short): 27.5 characters/s
PR typing speed (short): 27.5 characters/s

4.x typing speed (long): 17.5 characters/s
PR typing speed (long, 50 ms): 24.5 characters/s
PR typing speed (long, 100 ms): 23 characters/s
PR typing speed (long, 150 ms): 23 characters/s
PR typing speed (long, 200 ms): 22 characters/s
PR typing speed (long, 250 ms): 22.5 characters/s
PR typing speed (long, 500 ms): 23 characters/s


----

4.x line move speed (short): 25 lines/s
PR line move speed (short): 23 lines/s

4.x line move speed (long): 13.5 lines/s
PR line move speed (long, 50 ms): 19 lines/s
PR line move speed (long, 100 ms): 17 lines/s
PR line move speed (long, 250 ms): 19 lines/s
PR line move speed (long, 500 ms): 19 lines/s

</details>

I tested it at different settings for the timeout, but trying 50-500 ms for the delay value did not seem to produce meaningful differences in in my limited quantitative tests, at least. Therefore. I have no specific guidance on that point; I would just pick whichever one was the most responsive in your own tests.

I also somewhat informally tested enabling all of the decorations, with and without indent guides; the latter seemed to not render at all while the combination of the former (not shown) did worsen responsiveness, but only modestly.

Other than that, just one comment on the code and this should be ready. Thanks!

CAM-Gerlach:
Some additional results with indent guides, for reference (particularly with PR #13867 ):

4.x typing speed (short, indent guides): 19 characters/s
PR typing speed (short, indent guides): 19 characters/s

4.x typing speed (long, indent guides): characters/s
PR typing speed (long, indent guides): 6.7 characters/s

----

4.x line move speed (short, indent guides): 14 lines/s
PR line move speed (short, indent guides): 16 lines/s

4.x line move speed (long, indent guides): 7.5 lines/s
PR line move speed (long, indent guides): 6.0 lines/s

So as one would probably expect,, there is little difference between this PR and indent guides, BTW, when it comes to this particular issue.

CAM-Gerlach:
Testing the effect of this PR on typing speed across files of different sizes, I found interesting results that actually showed lower performance than even on 4.x for large files vs. previously, even when tested on the same file (mainwindow), and on larger files 4.x was better. 

EDIT: However, this might have to due with having multiple files open for the PR test but not 4.x, which I'll restest when I get back in a few hours.

4.x (~5 lines): 26 char/s
PR (~5 lines): 29 char/s

4.x (~20 lines): 27 char/s
PR (~20 lines): 25 char/s

4.x (~75 lines): 18.5 char/s
PR (~75 lines): 20 char/s

4.x (~250 lines): 18 char/s
PR (~250 lines): 19 char/s

4.x (~500 lines): 17.5 char/s
PR (~500 lines): 18.5 char/s

4.x (~1000 lines): 17.5 char/s
PR (~1000 lines): 17 char/s

4.x (~1600 lines): 16 char/s
PR (~1600 lines): 16 char/s

4.x (~3000 lines): 16 char/s
PR (~3000 lines): 14 char/s

4.x (~4000 lines): 15.5 char/s
PR (~4000 lines. mainwindow): 15 char/s

PR #13884, opened 2020-09-29
PR: Display an error message when Kite sends a non-dict response

### LINKED ISSUE #13121: TypeError: KiteClient.sig_response_ready in Kite's client

I the procedures on the uninstall of kite through the terminal. Then I found the kite folders on my macbook and i send them to trash, but now my spider give me this issue report constantly. 

i tryed to open a new enviroment in anaconda but the problem persist.

Should I unistall and reinstall all spider?

Thx

### Traceback
```python-traceback
  File "/Users/xaf/opt/anaconda3/lib/python3.7/site-packages/spyder/plugins/completion/kite/client.py", line 173, in perform_request
    self.sig_response_ready.emit(req_id, response or {})
TypeError: KiteClient.sig_response_ready[int, dict].emit(): argument 2 has unexpected type 'str'
  File "/Users/xaf/opt/anaconda3/lib/python3.7/site-packages/spyder/plugins/completion/kite/client.py", line 173, in perform_request
    self.sig_response_ready.emit(req_id, response or {})
TypeError: KiteClient.sig_response_ready[int, dict].emit(): argument 2 has unexpected type 'str'
```

## Versions

* Spyder version: 4.1.3 
* Python version: 3.7.6
* Qt version: 5.9.6
* PyQt5 version: 5.9.2
* Operating System: Darwin 19.3.0

### Dependencies

```

# Mandatory:
applaunchservices >=0.1.7      :  0.2.1 (OK)
atomicwrites >=1.2.0           :  1.4.0 (OK)
chardet >=2.0.0                :  3.0.4 (OK)
cloudpickle >=0.5.0            :  1.4.1 (OK)
diff_match_patch >=20181111    :  20181111 (OK)
intervaltree                   :  None (OK)
IPython >=4.0                  :  7.13.0 (OK)
jedi =0.15.2                   :  0.15.2 (OK)
nbconvert >=4.0                :  5.6.1 (OK)
numpydoc >=0.6.0               :  1.0.0 (OK)
parso =0.5.2                   :  0.5.2 (OK)
pexpect >=4.4.0                :  4.8.0 (OK)
pickleshare >=0.4              :  0.7.5 (OK)
psutil >=5.3                   :  5.7.0 (OK)
pygments >=2.0                 :  2.6.1 (OK)
pylint >=0.25                  :  2.4.4 (OK)
pyls >=0.31.9;<0.32.0          :  0.31.10 (OK)
qdarkstyle >=2.8               :  2.8.1 (OK)
qtawesome >=0.5.7              :  0.7.2 (OK)
qtconsole >=4.6.0              :  4.7.4 (OK)
qtpy >=1.5.0                   :  1.9.0 (OK)
rtree >=0.8.3                  :  0.9.4 (OK)
sphinx >=0.6.6                 :  3.0.4 (OK)
spyder_kernels >=1.9.1;<1.10.0 :  1.9.1 (OK)
watchdog                       :  None (OK)
zmq >=17                       :  18.1.1 (OK)

# Optional:
cython >=0.21                  :  0.29.20 (OK)
matplotlib >=2.0.0             :  3.1.3 (OK)
numpy >=1.7                    :  1.18.1 (OK)
pandas >=0.13.1                :  1.0.4 (OK)
scipy >=0.17.0                 :  1.4.1 (OK)
sympy >=0.7.3                  :  1.5.1 (OK)
```

andfoy:
## Description of Changes
This PR handles wrong Kite responses by displaying an error message and disabling Kite on Spyder 

imagen

### Issue(s) Resolved

Fixes #13121

### Affirmation

### DIFF
--- a/spyder/plugins/completion/kite/client.py
+++ b/spyder/plugins/completion/kite/client.py
@@ -19,7 +19,7 @@ from qtpy.QtWidgets import QMessageBox
 import requests
 
 # Local imports
-from spyder.config.base import _
+from spyder.config.base import _, running_under_pytest
 from spyder.config.manager import CONF
 from spyder.plugins.completion.kite import KITE_ENDPOINTS, KITE_REQUEST_MAPPING
 from spyder.plugins.completion.kite.decorators import class_register
@@ -173,15 +173,16 @@ class KiteClient(QObject, KiteMethodProviderMixIn):
                 if response is not None:
                     response = converter(response)
         if not isinstance(response, dict):
-            QMessageBox.critical(
-                self, _('Kite error'),
-                _("The Kite completion engine returned an unexpected result "
-                  "for the request <tt>{0}</tt>: <br><br><tt>{1}</tt><br><br>"
-                  "Please make sure that your Kite installation is correct. "
-                  "In the meantime, Spyder will disable the Kite client to "
-                  "prevent further errors. For more information, please "
-                  "visit the <a href='https://help.kite.com/'>Kite help "
-                  "center</a>").format(method, response))
-            CONF.set('kite', 'enable', False)
+            if not running_under_pytest():
+                QMessageBox.critical(
+                    self, _('Kite error'),
+                    _("The Kite completion engine returned an unexpected result "
+                    "for the request <tt>{0}</tt>: <br><br><tt>{1}</tt><br><br>"
+                    "Please make sure that your Kite installation is correct. "
+                    "In the meantime, Spyder will disable the Kite client to "
+                    "prevent further errors. For more information, please "
+                    "visit the <a href='https://help.kite.com/'>Kite help "
+                    "center</a>").format(method, response))
+                CONF.set('kite', 'enable', False)
         else:
             self.sig_response_ready.emit(req_id, response or {})
--- a/spyder/plugins/completion/kite/client.py
+++ b/spyder/plugins/completion/kite/client.py
@@ -15,10 +15,12 @@ except ImportError:
 
 # Third party imports
 from qtpy.QtCore import QObject, QThread, Signal, QMutex
+from qtpy.QtWidgets import QMessageBox
 import requests
 
 # Local imports
 from spyder.config.base import _
+from spyder.config.manager import CONF
 from spyder.plugins.completion.kite import KITE_ENDPOINTS, KITE_REQUEST_MAPPING
 from spyder.plugins.completion.kite.decorators import class_register
 from spyder.plugins.completion.kite.providers import KiteMethodProviderMixIn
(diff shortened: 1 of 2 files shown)


spyder/plugins/completion/kite/client.py
```diff
@@ -170,4 +172,16 @@ def perform_request(self, req_id, method, params):
                 converter = getattr(self, converter_name)
                 if response is not None:
                     response = converter(response)
-        self.sig_response_ready.emit(req_id, response or {})
+        if not isinstance(response, dict):
+            QMessageBox.critical(
+                self, _('Kite error'),
```
dalthviz, line 177: I think that the `QMessageBox.critical` needs a `QWidget` as parent. Since the `KiteClient` is only a QObject maybe that's the reason why some test are failing?

dalthviz (commented):
Thanks @andfoy seems like some tests are failing. Left a comment with a possible reason but not totally sure if that's the cause. The message LGTM :+1: (maybe @ccordoba12 can check it too just in case)

spyder/plugins/completion/kite/client.py
ccordoba12: I think to avoid this problem we should show the message only if `running_under_pytest()`  is False.

spyder/plugins/completion/kite/client.py
```diff
@@ -170,4 +172,17 @@ def perform_request(self, req_id, method, params):
                 converter = getattr(self, converter_name)
                 if response is not None:
                     response = converter(response)
-        self.sig_response_ready.emit(req_id, response or {})
+        if not isinstance(response, dict):
+            if not running_under_pytest():
```
dalthviz, line 176: @andfoy running on Windows I get the following traceback:

```
  File "C:\Users\Daniel\Documents\Spyder\Spyder otros\edgar\spyder\spyder\plugins\completion\kite\client.py", line 185, in perform_request
    "center</a>").format(method, response))
TypeError: critical(QWidget, str, str, buttons: Union[QMessageBox.StandardButtons, QMessageBox.StandardButton] = QMessageBox.Ok, defaultButton: QMessageBox.StandardButton = QMessageBox.NoButton): argument 1 has unexpected type 'KiteClient'
```

Which I think is related with my previous comment regarding test failing. Also seems like some PEP8 issues remain

dalthviz (commented):
Checking locally an error raised. Could you check @andfoy ? (I think is related with the initial comment regarding the `KiteClient` being a `QObject` instead of a `QWidget`) Thanks!

andfoy:
@dalthviz, this one is ready for review

spyder/plugins/completion/kite/plugin.py
```diff
@@ -222,3 +224,16 @@ def _show_onboarding_file(self, onboarding_file):
...
+    def _wrong_response_error(self, method, resp):
+        QMessageBox.critical(
+            self.main, _('Kite error'),
+            _("The Kite completion engine returned an unexpected result "
+              "for the request <tt>{0}</tt>: <br><br><tt>{1}</tt><br><br>"
+              "Please make sure that your Kite installation is correct. "
+              "In the meantime, Spyder will disable the Kite client to "
+              "prevent further errors. For more information, please "
+              "visit the <a href='https://help.kite.com/'>Kite help "
+              "center</a>").format(method, response))
```
dalthviz, line 238: @andfoy I think instead of `response` it should be `resp` (or maybe we should change the param to be `response`?)

dalthviz (commented):
Thanks @anfoy, I left a comment. Also, checking the changes manually, seems like the dialog is being trigger when starting Spyder:

image

Maybe we need to add some sort of validation for `None`?

andfoy:
@dalthviz, does the last commit fixes the last error?

dalthviz:
@andfoy the fix solves the syntax but the message keeps showing and reviewing again seems like multiple message are being created (my guess is that for each petition a message is being shown):

image

As I mentioned maybe a validation for a `None` value on the response is needed?

dalthviz (approved):
Thanks @andfoy LGTM :+1:

PR #13953, opened 2020-10-09
PR: Change Kite dialog to tour dialog when starting Spyder for first time

### LINKED ISSUE #13832: Replace kite dialog when opening spyder the first time for the tour dialog.
## Problem Description

Our plan is to do this change In order to give users the opportunity to know more about Spyder features.

juanis2112:
## Description of Changes

Changed Kite dialog to tour dialog when starting Spyder for first time.

Screenshot 2020-10-09 at 12 55 36 AM

### Issue(s) Resolved

Fixes #13832 

### Affirmation

Juanis2112

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -3432,12 +3432,15 @@ class MainWindow(QMainWindow):
 
     @Slot()
     def show_tour_message(self):
-        """Show message about starting the tour the first time Spyder starts."""
-        if  not running_under_pytest():
+        """
+        Show message about starting the tour the first time Spyder starts.
+        """
+        if not running_under_pytest():
             show_tour = CONF.get('main', 'show_tour_message')
             if show_tour:
                 CONF.set('main', 'show_tour_message', False)
-                self.tour_dialog = tour.OpenTourDialog(self, lambda: self.show_tour(0))
+                self.tour_dialog = tour.OpenTourDialog(
+                    self, lambda: self.show_tour(0))
                 self.tour_dialog.show()
 
     @Slot()
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -222,6 +222,11 @@ CWD = getcwd_or_home()
 qInstallMessageHandler(qt_message_handler)
 
 #==============================================================================
+# Set the index for the default tour
+#==============================================================================
+DEFAULT_TOUR = 0
+
+#==============================================================================
 # Main Window
 #==============================================================================
 class MainWindow(QMainWindow):
@@ -1075,8 +1080,7 @@ class MainWindow(QMainWindow):
         # self.tour_menu_actions = []
         # # TODO: Only show intro tour for now. When we are close to finish
         # # 3.0, we will finish and show the other tour
-        tour_index = 0
-        self.tours_available = tour.get_tours(tour_index)
+        self.tours_available = tour.get_tours(DEFAULT_TOUR)
 
         for i, tour_available in enumerate(self.tours_available):
             self.tours_available[i]['last'] = 0
@@ -1091,9 +1095,9 @@ class MainWindow(QMainWindow):
 
         # self.tours_menu.addActions(self.tour_menu_actions)
         self.tour_action = create_action(
-            self, self.tours_available[tour_index]['name'],
+            self, self.tours_available[DEFAULT_TOUR]['name'],
             tip=_("Interactive tour introducing Spyder's panes and features"),
-            triggered=lambda: self.show_tour(tour_index))
+            triggered=lambda: self.show_tour(DEFAULT_TOUR))
 
         self.help_menu_actions = [doc_action, vid_action, shortcuts_action,
                                   self.tour_action,
@@ -3440,7 +3444,7 @@ class MainWindow(QMainWindow):
             if show_tour:
                 CONF.set('main', 'show_tour_message', False)
                 self.tour_dialog = tour.OpenTourDialog(
-                    self, lambda: self.show_tour(0))
+                    self, lambda: self.show_tour(DEFAULT_TOUR))
                 self.tour_dialog.show()
 
     @Slot()
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -145,7 +145,7 @@ if hasattr(MAIN_APP, 'setDesktopFileName'):
 #==============================================================================
 # Create splash screen out of MainWindow to reduce perceived startup time.
 #==============================================================================
-from spyder.config.base import _, get_image_path, DEV
+from spyder.config.base import _, get_image_path, DEV, SAFE_MODE
 
 if not running_under_pytest():
     SPLASH = QSplashScreen(QPixmap(get_image_path('splash.svg')))
@@ -3435,7 +3435,7 @@ class MainWindow(QMainWindow):
         """
         Show message about starting the tour the first time Spyder starts.
         """
-        if not running_under_pytest():
+        if not running_under_pytest() and not SAFE_MODE:
             show_tour = CONF.get('main', 'show_tour_message')
             if show_tour:
                 CONF.set('main', 'show_tour_message', False)
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -528,6 +528,9 @@ class MainWindow(QMainWindow):
                                                    socket.SOCK_STREAM,
                                                    socket.IPPROTO_TCP)
 
+        # To show the message about starting the tour
+        self.sig_setup_finished.connect(self.show_tour_message)
+
         # Apply preferences
         self.apply_settings()
 
@@ -1400,10 +1403,6 @@ class MainWindow(QMainWindow):
         # Show dialog with missing dependencies
         if not running_under_pytest():
             self.report_missing_dependencies()
-
-        # Show tour dialog
-        self.tour_dialog = tour.OpenTourDialog(self, lambda: self.show_tour(0))
-        self.tour_dialog.show()
 
         # Raise the menuBar to the top of the main window widget's stack
         # Fixes spyder-ide/spyder#3887.
@@ -3430,6 +3429,16 @@ class MainWindow(QMainWindow):
 
         # Provide feeback when clicking menu if check on startup is on
         self.give_updates_feedback = True
+
+    @Slot()
+    def show_tour_message(self):
+        """Show message about starting the tour the first time Spyder starts."""
+        if  not running_under_pytest():
+            show_tour = CONF.get('main', 'show_tour_message')
+            if show_tour:
+                CONF.set('main', 'show_tour_message', False)
+                self.tour_dialog = tour.OpenTourDialog(self, lambda: self.show_tour(0))
+                self.tour_dialog.show()
 
     @Slot()
     def check_updates(self, startup=False):
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -3435,17 +3435,17 @@ class MainWindow(QMainWindow):
         self.give_updates_feedback = True
 
     @Slot()
-    def show_tour_message(self):
+    def show_tour_message(self, force=False):
         """
         Show message about starting the tour the first time Spyder starts.
         """
-        if not running_under_pytest() and not SAFE_MODE:
-            show_tour = CONF.get('main', 'show_tour_message')
-            if show_tour:
-                CONF.set('main', 'show_tour_message', False)
-                self.tour_dialog = tour.OpenTourDialog(
-                    self, lambda: self.show_tour(DEFAULT_TOUR))
-                self.tour_dialog.show()
+        should_show_tour = CONF.get('main', 'show_tour_message')
+        if force or (should_show_tour and not running_under_pytest()
+                     and not SAFE_MODE):
+            CONF.set('main', 'show_tour_message', False)
+            self.tour_dialog = tour.OpenTourDialog(
+                self, lambda: self.show_tour(DEFAULT_TOUR))
+            self.tour_dialog.show()
 
     @Slot()
     def check_updates(self, startup=False):
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -346,6 +346,7 @@ class MainWindow(QMainWindow):
         # Tour  # TODO: Should I consider it a plugin?? or?
         self.tour = None
         self.tours_available = None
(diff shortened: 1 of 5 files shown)


spyder/app/mainwindow.py
```diff
@@ -1396,6 +1397,10 @@ def post_visible_setup(self):
         if not running_under_pytest():
             self.report_missing_dependencies()
 
+        # Show tour dialog
+        self.tour_dialog = tour.OpenTourDialog(self, lambda: self.show_tour(0))
+        self.tour_dialog.show()
```
ccordoba12, line 1402: I think this makes the dialog to be displayed every time Spyder starts. To avoid that, you could connect a method called `show_tour_message` to the signal called `sig_setup_finished` (emitted at the end of `post_visibile_setup`), at the end of the `__init__` method of `MainWindow`:

```python
...

# To show the message about starting the tour
sig_setup_finished.connect(self.show_tour_message)

# Apply preferences
self.apply_settings()

# To set all dockwidgets tabs to be on top (in case we want to do it
...
```

Then you'd need to declare that method like this

```python
@Slot()
def show_tour_message(self):
    """Show message about starting the tour the first time Spyder starts."""
    show_tour = CONF.get('main', 'show_tour_message')
    if show_tour:
        CONF.set('main', 'show_tour_message', False)
        self.tour_dialog = tour.OpenTourDialog(self, lambda: self.show_tour(0))
        self.tour_dialog.show()
```

And finally you'd need to set the `show_tour_message` option as `True` in the `main` section of our defaults preferences, which are saved in `spyder/config/main.py`

ccordoba12 (commented):
Thanks @juanis2112 for this! It looks pretty good!

I left you a suggestion to display this new message only the first time Spyder starts, because with your current implementation is being shown every time.

CAM-Gerlach:
@ccordoba12 Are the tests failing because the dialog is interfering with qtbot manipulating the windows as it expects? Or is there another reason? Otherwise, I don't see a clear connection to this PR.

spyder/app/mainwindow.py
```diff
@@ -1396,6 +1397,10 @@ def post_visible_setup(self):
         if not running_under_pytest():
             self.report_missing_dependencies()
 
+        # Show tour dialog
+        self.tour_dialog = tour.OpenTourDialog(self, lambda: self.show_tour(0))
```
CAM-Gerlach, line 1401: What might be best here is to declare a module-level constant `DEFAULT_TOUR_INDEX` and use that both here and in PR #13717 but we'd need to merge that first, rebase and then make the change.

spyder/app/tour.py
```diff
@@ -1255,6 +1255,76 @@ def any_has_focus(self):
              self.tips.hasFocus() or self.canvas.hasFocus())
         return f
 
+
+class OpenTourDialog(QDialog):
+    """Initial Widget with tour"""
+    sig_launch_tour_button_clicked = Signal()
+    # Signal triggered for the 'Dismiss' button
+    sig_dismiss_button_clicked = Signal()
+
```
CAM-Gerlach, line 1264:
```suggestion
```

Currently, these signals aren't actually used. Unless they are needed for what @ccordoba12 suggested above, they should be removed.

spyder/app/tour.py
```diff
@@ -1255,6 +1255,76 @@ def any_has_focus(self):
...
+        image_label.setPixmap(image)
+
+        images_layout.addStretch()
+        images_layout.addWidget(image_label)
+        images_layout.addStretch()
+
+        # Label
+        tour_label = QLabel(
+            _("Welcome to Spyder! <br><br> Check our interactive tour to "
+              "explore some of Spyder's panes and features. "))
```
CAM-Gerlach, line 1298:
```suggestion
            _("Welcome to Spyder!<br><br>Check out our interactive tour to "
              "explore some of Spyder's panes and features."))
```

Fix phrasing and trim whitespace

spyder/app/tour.py
```diff
@@ -1255,6 +1255,76 @@ def any_has_focus(self):
...
+        self.setFixedHeight(170)
+        self.setFixedWidth(350)
+        self.tour_function = tour_function
+
+        # Image
+        images_layout = QHBoxLayout()
+        if is_dark_interface():
+            icon_filename = 'tour-spyder-logo.png'
+        else:
+            icon_filename = 'tour-spyder-logo.png'
```
CAM-Gerlach, line 1276:
```suggestion
            icon_filename = 'tour-spyder-logo.png'
```

Since we're not doing anything different for dark vs light mode, this can be simplified.

spyder/app/mainwindow.py
CAM-Gerlach: IMO, we should also not show the tour dialog when Spyder is launched in dev mode (to save us from having to click through it every time we launch Spyder in safe mode to test bugs and PRs, as we do dozens of times per day), i.e. when `DEV` is truthy, and also when `PYTEST` is truthy unless we want to write a test for it (which we really should). @ccordoba12 thoughts?

CAM-Gerlach (changes requested):
Thanks @juanis2112 ! Worked great in testing on Windows; just a few minor code comments.

Also, it would be really nice to write a simple test for this, since its the first UI element that new users will likely interact with so it looks really bad if its broken. Basically, we would want to just check that `Dismiss` hides the dialog and the other button starts the tour and also closes it, which should be pretty straightforward to test with Qtbot.

CAM-Gerlach:
Also, @ccordoba12 I'm no fan of Kite but I don't want Spyder to get in any hot water—is there anything contractual obligating you to include this dialog? Is Kite okay with us removing their dialog that they essentially paid you to put in?

ccordoba12:
> @ccordoba12 Are the tests failing because the dialog is interfering with qtbot manipulating the windows as it expects?

Yes, that's the problem. And now that you mention it, this dialog needs to be shown only if `not running_under_pytest()`

> Also, @ccordoba12 I'm no fan of Kite but I don't want Spyder to get in any hot water—is there anything contractual obligating you to include this dialog?

No, there isn't.

> Is Kite okay with us removing their dialog that they essentially paid you to put in?

Probably not, but it's up to us to decide what's best for our users.

CAM-Gerlach:
> Yes, that's the problem. And now that you mention it, this dialog needs to be shown only if not running_under_pytest()

I was going to suggest that, but if we write a simple test for it as I suggest in my review (since its the first UI element users are likely to interact with when opening Spyder for the first time, it would be really bad if it didn't work correctly) then we should do it with Pytest mark options or however we configure the editor to enable/disable completion and other features for specific tests, since we'll need it to open for that test.

> Probably not, but it's up to us to decide what's best for our users.

Well then I have no problem with that :+1:

ccordoba12 (commented):
@juanis2112, I think this is almost ready! I only left a minor suggestion for you.

ccordoba12:
> I was going to suggest that, but if we write a simple test for it as I suggest in my review

I don't see the need to add a test for this because the functionality is quite simple and clear. But you're welcome to push one to @juanis2112's branch (if she agrees with that, of course).

ccordoba12:
Screenshot 2020-10-09 at 12 55 36 AM

I was also thinking that we could show a nice icon to the left of text that implies "exploration", "getting familiar with a new territory", or something like that. Perhaps @isabela-pf could give us a hand with that?

spyder/app/mainwindow.py
ccordoba12: > IMO, we should also not show the tour dialog when Spyder is launched in dev mode (to save us from having to click through it every time we launch Spyder in safe mode to test bugs and PRs

Right, I agree with that. @juanis2112, please also avoid showing the dialog when `DEV` is `True`. That constant is present in `spyder/config/base.py`and it's `True` when we're running Spyder from `bootstrap.py`.

CAM-Gerlach: Actually, this will prevent us from actually testing it, manually or otherise. I was in the process of submitting a revised suggestion that instead checks for the safe mode env variable, so that if we're intentionally running Spyder under safe mode then it will not nag us (or the user) with the popup, while simply running Spyder under the normal dev environment with settings reset will, which we need to do anyway to properly test the only-run-once behavior. Standby.

spyder/app/mainwindow.py
```diff
@@ -3431,7 +3431,8 @@ def show_tour_message(self):
         """
         Show message about starting the tour the first time Spyder starts.
         """
-        if not running_under_pytest():
+        dev_mode = os.environ.get('SPYDER_DEV')
+        if not running_under_pytest() and not dev_mode:
```
ccordoba12, line 3435:
```suggestion
        if not running_under_pytest() and not DEV:
```

The line you introduced is not necessary. As I told you in my last comment, we declare a global variable that does exactly that in `spyder/config/base.py`:

https://github.com/spyder-ide/spyder/blob/9c422b81f7b839b79f88d50b0fc60e849be8cc4c/spyder/config/base.py#L39

And `DEV` is imported at the top of this file:

https://github.com/spyder-ide/spyder/blob/3ea0bdbbef55de203aabe800384e4fa811fb7434/spyder/app/mainwindow.py#L148

So we can use it here directly.

CAM-Gerlach: Sorry, I was in the process of writing this when the other suggestions got applied.

Actually, this will prevent us from actually testing it, manually or otherwise. I was in the process of submitting a revised suggestion that instead checks for the safe mode env variable, so that if we're intentionally running Spyder under safe mode then it will not nag us (or the user) with the spurious popup, while simply running Spyder under the normal dev environment with settings reset will show it as expected, which we need to do anyway to properly test the only-run-once behavior.

```suggestion
        if (not running_under_pytest()
                and not os.environ.get('SPYDER_SAFE_MODE')):
```

juanis2112: lol sorry. Fixing that now

spyder/app/mainwindow.py
```diff
@@ -3422,6 +3426,20 @@ def _check_updates_ready(self):
...
+        """
+        Show message about starting the tour the first time Spyder starts.
+        """
+        dev_mode = os.environ.get('SPYDER_DEV')
+        if not running_under_pytest() and not dev_mode:
+            show_tour = CONF.get('main', 'show_tour_message')
+            if show_tour:
+                CONF.set('main', 'show_tour_message', False)
+                self.tour_dialog = tour.OpenTourDialog(
+                    self, lambda: self.show_tour(0))
```
CAM-Gerlach, line 3440: It seems my suggestion before got lost—instead of hardcoding this magic number here, I suggest adding a module-level constant `DEFAULT_TOUR = 0` at the top of mainwindow.py or tour.py (@ccordoba12 opinion which?) and using that here and in the function that builds the tour menu that was added in #13717 (this will need a rebase).

CAM-Gerlach (commented):
Testing this produced the correct behavior when running multiple times, just two last code comments.

CAM-Gerlach:
> I was also thinking that we could show a nice icon to the left of text that implies "exploration", "getting familiar with a new territory", or something like that. Perhaps @isabela-pf could give us a hand with that?

Perhaps, but there is a connection to the Spyder logo since we say "Welcome to Spyder" on the first line, and I'm not sure this is worth blocking the whole PR over. However, I don't really have a strong opinion and don't want to cause more trouble.

spyder/app/mainwindow.py
ccordoba12: > Actually, this will make it impossible for us to actually test the popup, manually or automatically

Ok, fair enough. However, the same I said about `SPYDER_DEV` applies for `SPYDER_SAFE_MODE`: we already declare a global variable for that in `base/config.py`.

CAM-Gerlach:
> Ok, fair enough. However, the same I said about SPYDER_DEV applies for SPYDER_SAFE_MODE: we already declare a global variable for that in base/config.py.

@ccordoba12 Not sure where your comment went (I can see it on my email but can't find it here) but sorry about that, I forget that we had a constant for `SAFE_MODE` too and not just `DEV`.

juanis2112:
> > I was also thinking that we could show a nice icon to the left of text that implies "exploration", "getting familiar with a new territory", or something like that. Perhaps @isabela-pf could give us a hand with that?
> 
> Perhaps, but there is a connection to the Spyder logo since we say "Welcome to Spyder" on the first line, and I'm not sure this is worth blocking the whole PR over. However, I don't really have a strong opinion and don't want to cause more trouble.

I agree with CAM that is not worth it blocking the PR but we could consider another icon so I suggest merging this one and opening a new one for the icon if we do use another one. We will talk with @isabela-pf about that on Tuesday.

ccordoba12:
> I agree with CAM that is not worth it blocking the PR but we could consider another icon so I suggest merging this one and opening a new one for the icon if we do use another one. We will talk with @isabela-pf about that on Tuesday.

Ok, sounds good to me.

juanis2112:
> > I was going to suggest that, but if we write a simple test for it as I suggest in my review
> 
> I don't see the need to add a test for this because the functionality is quite simple and clear. But you're welcome to push one to @juanis2112's branch (if she agrees with that, of course).

Yeah, I don't have a problem with that. Feel free to push a test to this branch. Let us know if you will so we hold merging.

CAM-Gerlach:
> Yeah, I don't have a problem with that. Feel free to push a test to this branch. Let us know if you will so we hold merging @CAM-Gerlach

Okay, thanks. I'll take a look at that now.

CAM-Gerlach (dismissed):
Tested and looked over the code one last time, and everything's LGTM (aside from the test which I am working on). Thanks @juanis2112 .

ccordoba12:
> Tested and looked over the code one last time, and everything's LGTM (aside from the test which I am working on)

Ok, but do you plan to add that test here or in a different PR?

just asking to know if this PR is ready to go.

CAM-Gerlach:
I can add it here per what @juanis2112 mentioned, I confirmed I have access to the branch. I'm working on it right now, should be pretty straightforward.

ccordoba12 (approved):
Thanks @juanis2112 and @CAM-Gerlach!

isabela-pf:
> I was also thinking that we could show a nice icon to the left of text that implies "exploration", "getting familiar with a new territory", or something like that. Perhaps @isabela-pf could give us a hand with that?

@ccordoba12 I'm happy to work on that! Just know it will be a lower priority on my list and won't happen right away unless you want me to pause another priority and work on this first.

ccordoba12:
Thanks @isabela-pf! We need this for 4.2 and our plan is to release it in three weeks at most. So it's kind of urgent.

PR #13958, opened 2020-10-10
PR: Update, revise and copyedit existing tour steps

CAM-Gerlach:
## Description of Changes

Per @ccordoba12 's request, I've revised the existing tour steps to update some out of date references, fix grammar issues, improve phrasing, and add some additional content. However, I was fairly conservative and didn't add too much new stuff from our docs; @juanis2112 feel free to suggest/add additional details or further revise the text as desired.

### Affirmation

### DIFF
--- a/spyder/app/tour.py
+++ b/spyder/app/tour.py
@@ -144,7 +144,7 @@ def get_tour(index):
              'interact': False},
             ]
 
-    intro = [{'title': _("Welcome to Spyder!"),
+    intro = [{'title': _("Welcome to the introduction tour!"),
               'content': _("<b>Spyder</b> is a powerful Interactive "
                            "Development Environment (or IDE) for the Python "
                            "programming language.<br><br>"
@@ -154,19 +154,20 @@ def get_tour(index):
                            "below to move along the tour."),
               'image': 'tour-spyder-logo.png'},
 
-             {'title': _("The Editor"),
+             {'title': _("Editor"),
               'content': _("This is where you write Python code before "
                            "evaluating it. You can get automatic "
-                           "completions while typing by pressing the "
-                           "<b>Tab</b> key, or automatically by enabling "
-                           "on-the-fly completions in Spyder's Preferences. "
+                           "completions while typing, along with calltips "
+                           "when calling a function and help when hovering "
+                           "over an object."
                            "<br><br>The Editor comes "
                            "with a line number area (highlighted here in red) "
                            "where Spyder shows warnings and syntax errors. "
                            "They can help you to detect potential problems "
                            "before running your code.<br><br>"
                            "You can also set debug breakpoints in the line "
-                           "number area by clicking next to a "
+                           "number area by clicking next to "
+
                            "any non-empty line."),
               'widgets': [sw.editor],
               'decoration': [sw.editor_line_number_area]},
@@ -175,10 +176,9 @@ def get_tour(index):
               'content': _("This is where you can run Python code, either "
                            "from the Editor or interactively. To run the "
                            "current file, press <b>F5</b> by default, "
-                           "or press <b>Ctrl+Enter</b> "
-                           "(<b>Cmd-Return</b> on Mac) to execute the current "
+                           "or press <b>F9</b> to execute the current "
                            "line or selection.<br><br>"
-                           "This console comes with many  "
+                           "The IPython Console comes with many "
                            "useful features that greatly improve your "
                            "programming workflow, like syntax highlighting, "
                            "autocompletion, plotting and 'magic' commands. "
@@ -187,7 +187,7 @@ def get_tour(index):
                            f"<br><br>{button_text}"),
               'widgets': [sw.ipython_console],
               'run': [
-                  "test_list_tour = list(range(100))",
+                  "test_list_tour = [1, 2, 3, 4, 5]",
                   "test_dict_tour = {'a': 1, 'b': 2}",
                   ]
               },
@@ -195,7 +195,8 @@ def get_tour(index):
              {'title': _("Variable Explorer"),
               'content': _("In this pane you can view and edit the variables "
                            "generated during the execution of a program, or "
-                           "those entered directly in the Console.<br><br>"
+                           "those entered directly in the "
+                           "IPython Console.<br><br>"
                            "If you ran the code in the previous step, "
                            "the Variable Explorer will show "
                            "the list and dictionary objects it generated. "
@@ -208,9 +209,10 @@ def get_tour(index):
              {'title': _("Help"),
               'content': _("This pane displays documentation of the "
                            "functions, classes, methods or modules you are "
-                           "currently using in the Editor or the Console."
+                           "currently using in the Editor or the "
+                           "IPython Console."
                            "<br><br>To use it, press <b>Ctrl+I</b> "
-                           "(<b>Cmd-I</b> on Mac) with the text cursor "
+                           "(<b>Cmd-I</b> on macOS) with the text cursor "
                            "in or next to the object you want help on."),
               'widgets': [sw.help_plugin],
               'interact': True},
@@ -231,7 +233,7 @@ def get_tour(index):
                            "inside of Spyder.<br><br>"
                            "The Files pane also allows you to copy one or "
                            "many absolute or relative paths, automatically "
-                           "formatted as Python string or list, and perform "
+                           "formatted as Python strings or lists, and perform "
                            "a variety of other file operations."),
               'widgets': [sw.file_explorer],
               'interact': True},
--- a/spyder/app/tour.py
+++ b/spyder/app/tour.py
@@ -144,56 +144,63 @@ def get_tour(index):
              'interact': False},
             ]
 
-    intro = [{'title': _("Welcome to the Introduction tour"),
+    intro = [{'title': _("Welcome to Spyder!"),
               'content': _("<b>Spyder</b> is a powerful Interactive "
                            "Development Environment (or IDE) for the Python "
                            "programming language.<br><br>"
-                           "Here we are going to guide you through its most "
+                           "Here, we are going to guide you through its most "
                            "important features.<br><br>"
                            "Please use the arrow keys or click on the buttons "
                            "below to move along the tour."),
               'image': 'tour-spyder-logo.png'},
 
              {'title': _("The Editor"),
-              'content': _("This is the pane where you write Python code before "
-                           "evaluating it. You can get automatic suggestions "
-                           "and completions while writing, by pressing the "
-                           "<b>Tab</b> key next to a given text.<br><br>"
-                           "The Editor comes "
-                           "with a line number area (highlighted here in red), "
-                           "where Spyder shows warnings and syntax errors. They "
-                           "can help you to detect potential problems before "
-                           "running the code.<br><br>"
+              'content': _("This is where you write Python code before "
+                           "evaluating it. You can get automatic "
+                           "completions while typing by pressing the "
+                           "<b>Tab</b> key, or automatically by enabling "
+                           "on-the-fly completions in Spyder's Preferences. "
+                           "<br><br>The Editor comes "
+                           "with a line number area (highlighted here in red) "
+                           "where Spyder shows warnings and syntax errors. "
+                           "They can help you to detect potential problems "
+                           "before running your code.<br><br>"
                            "You can also set debug breakpoints in the line "
-                           "number area, by doing a double click next to "
-                           "a non-empty line."),
+                           "number area by clicking next to a "
+                           "any non-empty line."),
               'widgets': [sw.editor],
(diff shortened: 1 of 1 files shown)


spyder/app/tour.py
```diff
@@ -144,68 +144,74 @@ def get_tour(index):
...
-                           "with a line number area (highlighted here in red), "
-                           "where Spyder shows warnings and syntax errors. They "
-                           "can help you to detect potential problems before "
-                           "running the code.<br><br>"
+              'content': _("This is where you write Python code before "
+                           "evaluating it. You can get automatic "
+                           "completions while typing by pressing the "
+                           "<b>Tab</b> key, or automatically by enabling "
+                           "on-the-fly completions in Spyder's Preferences. "
+                           "<br><br>The Editor comes "
```
ccordoba12, line 163: This is the default now, i.e. you'll get automatic completions while typing. So, perhaps we should rephrase this?

spyder/app/tour.py
```diff
@@ -144,68 +144,74 @@ def get_tour(index):
...
-                           "execute the code you wrote on the Editor. To do it "
-                           "you need to press the <b>F5</b> key.<br><br>"
-                           "This console comes with several "
+             {'title': _("IPython Console"),
+              'content': _("This is where you can run Python code, either "
+                           "from the Editor or interactively. To run the "
+                           "current file, press <b>F5</b> by default, "
+                           "or press <b>Ctrl+Enter</b> "
+                           "(<b>Cmd-Return</b> on Mac) to execute the current "
+                           "line or selection.<br><br>"
```
ccordoba12, line 180: This doesn't work in Spyder 4 (it worked in version 3 though): the way to run the current line or selection is by pressing `F9`.

`Ctrl/Cmd + Enter` is used exclusively to run cells now (unless you remap that shortcut to run the current line/selection).

spyder/app/tour.py
```diff
@@ -144,68 +144,74 @@ def get_tour(index):
...
-                           qtconsole_link=qtconsole_link, button_text=button_text),
+                           "programming workflow, like syntax highlighting, "
+                           "autocompletion, plotting and 'magic' commands. "
+                           "To learn more, check out the "
+                           f"<a href=\"{qtconsole_link}\">documentation</a>."
+                           f"<br><br>{button_text}"),
               'widgets': [sw.ipython_console],
-              'run': ["li = list(range(100))", "d = {'a': 1, 'b': 2}"]
+              'run': [
+                  "test_list_tour = list(range(100))",
```
ccordoba12, line 190:
```suggestion
                  "test_list_tour = [1, 2, 3, 4, 5]",
```

Just to make this code less scary to newbies (but I like the new names for these variables).

spyder/app/tour.py
```diff
@@ -144,68 +144,74 @@ def get_tour(index):
...
+             {'title': _("Variable Explorer"),
               'content': _("In this pane you can view and edit the variables "
                            "generated during the execution of a program, or "
-                           "those entered directly in one of Spyder "
-                           "consoles.<br><br>"
-                           "As you can see, the Variable Explorer is showing "
-                           "the variables generated during the last step of "
-                           "this tour. By doing a double-click on any "
-                           "of them, a new window will be opened, where you "
+                           "those entered directly in the Console.<br><br>"
```
ccordoba12, line 198: I think we should we change `Console` to `IPython console` here because the previous pane was clearly called `IPython console`.

spyder/app/tour.py
```diff
@@ -144,68 +144,74 @@ def get_tour(index):
...
+                           "If you ran the code in the previous step, "
+                           "the Variable Explorer will show "
+                           "the list and dictionary objects it generated. "
+                           "By double-clicking any variable, "
+                           "a new window will be opened where you "
```
ccordoba12, line 203: Does this not interfere with the tour by creating strange focus problems? Just checking with you because I haven't tested that.

spyder/app/tour.py
```diff
@@ -144,68 +144,74 @@ def get_tour(index):
...
 
              {'title': _("Help"),
               'content': _("This pane displays documentation of the "
                            "functions, classes, methods or modules you are "
-                           "currently using in the Editor or the Consoles.<br><br>"
-                           "To use it, you need to press <b>Ctrl+I</b> in "
-                           "front of an object. If that object has some "
-                           "documentation associated with it, it will be "
-                           "displayed here."),
+                           "currently using in the Editor or the Console."
```
ccordoba12, line 211: Again: should we change `Console` to `IPython console`?

spyder/app/tour.py
```diff
@@ -144,68 +144,74 @@ def get_tour(index):
...
+                           "<br><br>To use it, press <b>Ctrl+I</b> "
+                           "(<b>Cmd-I</b> on Mac) with the text cursor "
```
ccordoba12, line 213:
```suggestion
                           "(<b>Cmd-I</b> on macOS) with the text cursor "
```

That's the right name for that OS.

spyder/app/tour.py
```diff
@@ -217,19 +223,23 @@ def get_tour(index):
...
-                           "files will always be opened in the Spyder Editor."),
+              'content': _("This pane lets you browse the files and "
+                           "directories on your computer.<br><br>"
+                           "You can open any file in its "
+                           "corresponding application by double-clicking it, "
+                           "and supported file types will be opened right "
+                           "inside of Spyder.<br><br>"
+                           "The Files pane also allows you to copy one or "
+                           "many absolute or relative paths, automatically "
+                           "formatted as Python string or list, and perform "
```
ccordoba12, line 234:
```suggestion
                           "formatted as Python strings or list, and perform "
```

Not sure if that's a small grammatical error.

spyder/app/tour.py
```diff
@@ -217,19 +223,23 @@ def get_tour(index):
...
               'widgets': [sw.file_explorer],
               'interact': True},
 
-             {'title': _("The History Log"),
-              'content': _("This pane records all commands introduced in "
-                           "any IPython console."),
+             {'title': _("History Log"),
+              'content': _("This pane records all the commands and code run "
+                           "in any IPython console, allowing you to easily "
+                           "retrace your steps for reproducible research."),
```
ccordoba12, line 242: Pretty nice description!

ccordoba12 (commented):
Thanks a lot @CAM-Gerlach for this! It looks pretty good overall.

spyder/app/tour.py
CAM-Gerlach: Sorry, I didn't remember it was on by default; despite finding it really useful in Rstudio, I've had it disabled due to the performance impacts and UX issues I and others have reported with it. I'll revise it accordingly.

spyder/app/tour.py
CAM-Gerlach: Sorry, on my stable copy I've had the various run shortcuts remapped to be Enter + modifier keys, rather than the default of some being F-keys and some being Enter + Modifiers that doesn't make any sense to me and is harder to press. I thought I'd referenced Spyder running in safe mode to check the default, as I usually do, but evidently I accidentally checked my stable copy instead, my mistake. I'll fix that.

spyder/app/tour.py
CAM-Gerlach: Sure, makes more sense.

Glad you like the names. I changed them primarily to avoid overwriting any variables in the user's current namespace, to minimize the tour's side effects.

spyder/app/tour.py
CAM-Gerlach: Okay, will do.

spyder/app/tour.py
CAM-Gerlach: The tour completely disables user interaction with Spyder on all the panes and menus, as it always has, so the user attempting to do this during the tour itself won't cause problems (it just won't work). Since users won't be able to actually do anything we tell users on any of the panes during the tour itself as opposed to after, and this was here before, I suggest we just leave it, otherwise there isn't much we can really say on any of them.

spyder/app/tour.py
CAM-Gerlach: Same rationale as above—assuming we do change it, we should do so both places.

spyder/app/tour.py
ccordoba12: I don't think there could arise a confusion with the Internal Console. I mentioned it just to be consistent.

@juanis2112, what do you think?

spyder/app/tour.py
CAM-Gerlach: macOS is the name of the OS, but Mac (short for Macintosh) is the name of the computer platform. I.e. I own a Mac that runs macOS 10.15. I would have used macOS here, but I wanted to be consistent with the other place I used it above, which would have required an extra line break to change from Mac which I judged wasn't worth the trouble. However, since that latter usage will be eliminated per your comment, there's no harm in applying this here.

spyder/app/tour.py
CAM-Gerlach: It is, but list also needs to be plural to match.

```suggestion
                           "formatted as Python strings or lists, and perform "

```

spyder/app/tour.py
CAM-Gerlach: Thanks, I figured you'd like that.

spyder/app/tour.py
CAM-Gerlach: I removed the unnecessary text, and added brief mention of calltips and hover hints as other helpful introspection features (particularly for newer users) that are on by default.

ccordoba12:
This looks good to me now, apart from the Console/IPython console detail.

spyder/app/tour.py
juanis2112: I think we should use IPython console.

spyder/app/tour.py
juanis2112: Yeap

juanis2112 (approved):
Found a minor typo but apart from that, it looks good to me. Thanks @CAM-Gerlach

spyder/app/tour.py
```diff
@@ -144,68 +144,73 @@ def get_tour(index):
...
+                           "over an object."
+                           "<br><br>The Editor comes "
+                           "with a line number area (highlighted here in red) "
+                           "where Spyder shows warnings and syntax errors. "
+                           "They can help you to detect potential problems "
+                           "before running your code.<br><br>"
                            "You can also set debug breakpoints in the line "
-                           "number area, by doing a double click next to "
-                           "a non-empty line."),
+                           "number area by clicking next to a "
```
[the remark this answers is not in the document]
CAM-Gerlach: Sorry, fhanks for the catch.

spyder/app/tour.py
CAM-Gerlach: Done here and two other places.

spyder/app/tour.py
CAM-Gerlach: Done.

ccordoba12 (approved):
Thanks @CAM-Gerlach! Great work here!

juanis2112 (approved):
Looks good to me

spyder/app/tour.py
```diff
@@ -144,68 +144,76 @@ def get_tour(index):
              'interact': False},
             ]
 
-    intro = [{'title': _("Welcome to the Introduction tour"),
+    intro = [{'title': _("Welcome to Spyder!"),
```
juanis2112, line 147: I just realised that if users go to the tour from the dialog that we just added in PR#13953 they will see "Welcome to Spyder" two times. One on that dialog and the other one in the introduction tour first dialog so maybe we should leave this one as "Welcome to the Introduction tour". What do you think @ccordoba12 and @CAM-Gerlach?

ccordoba12: That's a good point. What about if instead of repeating `Welcome` twice, we say here instead `Explore what Spyder has to offer` or `Learn about Spyder's cool features`.

Those are just suggestions, but my idea is to mention something more appealing and fun, instead of restating the obvious.

CAM-Gerlach: Excellent point @juanis2112 . While the current title may not be the most eye-catching, we have to remember that this is the name of the first step displayed in the popup menu, not body text or the title of the whole tour, so it needs to be kept short, appropriate and descriptive of that step.

Therefore, I going with @juanis2112 's suggestion of reverting to the original text (or perhaps a less drastic variation, if either of you have further suggestions:

```suggestion
    intro = [{'title': _("Welcome to the introduction tour!"),
```

ccordoba12: Ok, no problem if the two of you agree.

juanis2112: Yeah I agree with @CAM-Gerlach

CAM-Gerlach: Okay, thanks. Pushed.

juanis2112 (approved):
Thanks @CAM-Gerlach

ccordoba12 (approved):

PR #13269, opened 2020-07-12
PR: Windows Installer script

dalthviz:
## Description of Changes
Add a config file to use `pynsist` to create Windows installer

### Steps to build:

* Create an env with python 3.6.8 and `pynsist` (you should use only pip to install packages in this env):
* Install the packages listed in the `installer.cfg` file in the env:
     - `autopep8`
     - `diff_match_patch`
     - `intervaltree`
     - `numpydoc`
     - `pandocfilters`
     - `pathtools`
     - `pyrsistent`
     - `rope`
     - `watchdog`
     - `wrapt`

* The rest of packages listed are accessible via the `pywin32` wheel but needed here to prevent a `NoModuleFoundError` while starting the console kernel using the default interpreter (the one embedded in the installer):
     * `win32api`
     * `win32security`
     * `ntsecuritycon`

* Over the root of the repo run `pynsist installer.cfg`

### pynsist env packages and versions

<details>

```
alabaster==0.7.12
autopep8==1.5.3
Babel==2.8.0
certifi==2020.6.20
chardet==3.0.4
colorama==0.4.3
diff-match-patch==20181111
distlib==0.3.1
docutils==0.16
idna==2.10
imagesize==1.2.0
intervaltree==3.0.2
Jinja2==2.11.2
MarkupSafe==1.1.1
numpydoc==1.1.0
packaging==20.4
pandocfilters==1.4.2
pathtools==0.1.2
pip==20.1.1
pycodestyle==2.6.0
Pygments==2.6.1
pynsist==2.5.1
pyparsing==2.4.7
pyrsistent==0.16.0
pytz==2020.1
pywin32==228
pywin32-ctypes==0.2.0
requests==2.24.0
requests-download==0.1.2
rope==0.17.0
setuptools==47.3.1.post20200622
six==1.15.0
snowballstemmer==2.0.0
sortedcontainers==2.2.2
Sphinx==3.1.2
sphinxcontrib-applehelp==1.0.2
sphinxcontrib-devhelp==1.0.2
sphinxcontrib-htmlhelp==1.0.3
sphinxcontrib-jsmath==1.0.1
sphinxcontrib-qthelp==1.0.3
sphinxcontrib-serializinghtml==1.1.4
toml==0.10.1
urllib3==1.25.9
watchdog==0.10.3
wheel==0.34.2
wincertstore==0.2
wrapt==1.12.1
yarg==0.1.9
```

</details>

### Install preview

image

image

image

image

image

image

### Uninstall preview

image

[...]

PR #13992, opened 2020-10-16
PR: Add scripts and files necessary to create a macOS installer

### LINKED ISSUE #13978: Migrate mac-application repo to spyder repo
Per conversation https://github.com/spyder-ide/spyder/pull/12011#discussion_r503636816 : move changes from PR spyder-ide/mac-application#3 to this repository.

To be addressed after
1. #12232 
1. #12235 
1. #12236

mrclary:
## Description of Changes
Added files for creating Spyder.app macOS standalone application.

Included the necessary resources to build a standalone macOS application.
See spyder-ide/mac-application#3

### Issue(s) Resolved

Fixes #13978 

### Affirmation

mrclary

### DIFF
--- a/.github/workflows/installer-macos.yml
+++ b/.github/workflows/installer-macos.yml
@@ -14,6 +14,10 @@ jobs:
       run:
         shell: bash -l {0}
         working-directory: ./installers/macOS
+    env:
+      spy_repo: /Users/runner/work/spyder/spyder
+      spk_repo: /Users/runner/work/spyder/spyder/external-deps/spyder-kernels
+      pls_repo: /Users/runner/work/spyder/spyder/external-deps/python-language-server
     steps:
       - name: Checkout Code
         uses: actions/checkout@v2
@@ -23,10 +27,16 @@ jobs:
 #        uses: actions/setup-python@v2
 #        with:
 #          python-version: 3.7.9
-      - name: Install Dependencies
+      - name: Install Dependencies -- Release
+        if: ${{github.event_name == 'release'}}
         run: |
           # is there a better way to install only spyder dependents?
-          pip3 install -r req-build.txt -r req-extras.txt -c req-const.txt -e ../../
+          pip3 install -r req-build.txt -r req-extras.txt -c req-const.txt -e ${spy_repo}
+          pip3 uninstall -q -y spyder
+      - name: Install Dependencies -- PR
+        if: ${{github.event_name == 'pull_request'}}
+        run: |
+          pip3 install -r req-build.txt -r req-extras.txt -c req-const.txt -e ${spk_repo} -e ${pls_repo} -e ${spy_repo}
           pip3 uninstall -q -y spyder
       - name: Build Application Bundle -- Lite
         run: python3 setup.py --lite --dmg py2app
--- a/.github/workflows/installer-macos.yml
+++ b/.github/workflows/installer-macos.yml
@@ -24,18 +24,18 @@ jobs:
         shell: bash -l {0}
         run: |
           # is there a better way to install only spyder dependents?
-          pip install -r installers/macOS/req-build.txt -r installers/macOS/req-extras.txt -c installers/macOS/req-const.txt -e .
-          pip uninstall spyder
+           pip3 install -r installers/macOS/req-build.txt -r installers/macOS/req-extras.txt -c installers/macOS/req-const.txt -e .
+           pip3 uninstall spyder
       - name: Build Lite
         shell: bash -l {0}
         run: |
           cd installers/macOS
-          python setup.py --lite --dmg py2app
+          python3 setup.py --lite --dmg py2app
       - name: Build Full
         shell: bash -l {0}
         run: |
           cd installers/macOS
-          python setup.py --dmg py2app
+          python3 setup.py --dmg py2app
       - name: Upload Artifact Lite
         uses: actions/upload-artifact@v2
         with:
--- a/.github/workflows/installer-macos.yml
+++ b/.github/workflows/installer-macos.yml
@@ -10,14 +10,17 @@ jobs:
   build:
     name: macOS App Bundle
     runs-on: macos-latest
+    strategy:
+      matrix:
+        build_type: ['Lite', 'Full']
     defaults:
       run:
         shell: bash -l {0}
         working-directory: ./installers/macOS
     env:
-      spy_repo: /Users/runner/work/spyder/spyder
-      spk_repo: /Users/runner/work/spyder/spyder/external-deps/spyder-kernels
-      pls_repo: /Users/runner/work/spyder/spyder/external-deps/python-language-server
+      BUILD_TYPE: ${{ matrix.build_type }}
+      DMGNAME: ${{ matrix.build_type == 'Lite' && 'Spyder-Lite.dmg' || 'Spyder.dmg' }}
+      DMGPATH: installers/macOS/dist/${{ matrix.build_type == 'Lite' && 'Spyder-Lite.dmg' || 'Spyder.dmg' }}
     steps:
       - name: Checkout Code
         uses: actions/checkout@v2
@@ -27,54 +30,41 @@ jobs:
 #        uses: actions/setup-python@v2
 #        with:
 #          python-version: 3.7.9
-      - name: Install Dependencies -- Release
-        if: ${{github.event_name == 'release'}}
+      - name: Install Dependencies
         run: |
-          # is there a better way to install only spyder dependents?
-          pip3 install -r req-build.txt -r req-extras.txt -c req-const.txt -e ${spy_repo}
+          SPK_REPO=${GITHUB_WORKSPACE}/external-deps/spyder-kernels
+          PLS_REPO=${GITHUB_WORKSPACE}/external-deps/python-language-server
+          INSTALL_FLAGS=()
+          if [[ ${BUILD_TYPE} == 'Full' ]]; then
+              INSTALL_FLAGS+=('-r' 'req-extras.txt')
+          fi
+          if [[ ${GITHUB_EVENT_NAME} == 'pull_request' ]]; then
+              INSTALL_FLAGS+=('-e' ${SPK_REPO} '-e' ${PLS_REPO})
+          fi
+          pip3 install -c req-const.txt -r req-build.txt "${INSTALL_FLAGS[@]}" -e ${GITHUB_WORKSPACE}
           pip3 uninstall -q -y spyder
-      - name: Install Dependencies -- PR
-        if: ${{github.event_name == 'pull_request'}}
-        run: |
-          pip3 install -r req-build.txt -r req-extras.txt -c req-const.txt -e ${spk_repo} -e ${pls_repo} -e ${spy_repo}
-          pip3 uninstall -q -y spyder
-      - name: Build Application Bundle -- Lite
-        run: python3 setup.py --lite --dmg py2app
-      - name: Build Application Bundle -- Full
-        run: python3 setup.py --dmg py2app
-      - name: Upload Artifact -- Lite
+      - name: Build Application Bundle
+        run: python3 setup.py ${BUILD_FLAG} py2app
+      - name: Build Disk Image
+        run: python3 setup.py ${BUILD_FLAG} --dmg
+      - name: Upload Artifact
         uses: actions/upload-artifact@v2
         with:
-          path: installers/macOS/dist/Spyder-Lite.dmg
-          name: Spyder-Lite.dmg
-      - name: Upload Artifact -- Full
-        uses: actions/upload-artifact@v2
-        with:
-          path: installers/macOS/dist/Spyder.dmg
-          name: Spyder.dmg
+          path: ${{ env.DMGPATH }}
+          name: ${{ env.DMGNAME }}
       - name: Get Release
         if: ${{github.event_name == 'release'}}
         id: get_release
         env:
           GITHUB_TOKEN: ${{ github.token }}
         uses: bruceadams/get-release@v1.2.0
-      - name: Upload Release Asset -- Lite
+      - name: Upload Release Asset
         if: ${{github.event_name == 'release'}}
         uses: actions/upload-release-asset@v1
         env:
           GITHUB_TOKEN: ${{ github.token }}
         with:
           upload_url: ${{ steps.get_release.outputs.upload_url }}
-          asset_path: installers/macOS/dist/Spyder-Lite.dmg
-          asset_name: Spyder-Lite.dmg
-          asset_content_type: application/x-apple-diskimage
-      - name: Upload Release Asset -- Full
-        if: ${{github.event_name == 'release'}}
-        uses: actions/upload-release-asset@v1
-        env:
-          GITHUB_TOKEN: ${{ github.token }}
-        with:
-          upload_url: ${{ steps.get_release.outputs.upload_url }}
-          asset_path: installers/macOS/dist/Spyder.dmg
-          asset_name: Spyder.dmg
+          asset_path: ${{ env.DMGPATH }}
+          asset_name: ${{ env.DMGNAME }}
           asset_content_type: application/x-apple-diskimage
--- a/.github/workflows/installer-macos.yml
+++ b/.github/workflows/installer-macos.yml
@@ -25,7 +25,7 @@ jobs:
         run: |
           # is there a better way to install only spyder dependents?
            pip3 install -r installers/macOS/req-build.txt -r installers/macOS/req-extras.txt -c installers/macOS/req-const.txt -e .
-           pip3 uninstall spyder
+           pip3 uninstall -q -y spyder
       - name: Build Lite
         shell: bash -l {0}
         run: |
--- a/.github/workflows/installer-macos.yml
+++ b/.github/workflows/installer-macos.yml
@@ -16,11 +16,11 @@ jobs:
     defaults:
       run:
         shell: bash -l {0}
-        working-directory: ./installers/macOS
+        working-directory: ${{ github.workspace }}/installers/macOS
     env:
-      BUILD_TYPE: ${{ matrix.build_type }}
+      LITE_FLAG: ${{ matrix.build_type == 'Lite' && '--lite' || '' }}
       DMGNAME: ${{ matrix.build_type == 'Lite' && 'Spyder-Lite.dmg' || 'Spyder.dmg' }}
-      DMGPATH: installers/macOS/dist/${{ matrix.build_type == 'Lite' && 'Spyder-Lite.dmg' || 'Spyder.dmg' }}
+      DISTDIR: ${{ github.workspace }}/dist
     steps:
       - name: Checkout Code
         uses: actions/checkout@v2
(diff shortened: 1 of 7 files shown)


mrclary:
Requires #12236 to be merged.

mrclary:
@ccordoba12, we should be good to incorporate into github actions, though I don't have any experience there...

ccordoba12:
> @ccordoba12, we should be good to incorporate into github actions

Great! Please rebase to get the fix to our tests that was added by @dalthviz for the Windows installer in PR #13269.

> though I don't have any experience there...

@dalthviz and I can give you a hand with that. Could you describe the steps necessary to build the dmg?

dalthviz:
@mrclary for the Github Action to build the installer maybe checking the workflow definition for the windows installer could useful. You can check it here: https://github.com/spyder-ide/spyder/blob/4.x/.github/workflows/installer-win.yml

Basically the workflow defines a series of the events that trigger the workflow (for example a pull request, a release, etc), then the environment where the workflow will run (OS, env variables, etc) and finally a series of steps (to install python, install dependencies, run a custom script, upload artifacts, etc). 

Probably you can use the windows installer workflow to create a similar workflow for macOS but let us know if you have any question and we will check :+1:

mrclary:
@ccordoba12 @dalthviz , 
Okay, here is a first crack at it. Is there a way to test it?

mrclary:
It seems to have trouble installing dependencies...

ccordoba12:
There seems to be syntax error in one of your scripts:

```python-traceback
File "setup.py", line 161
    _py_ver = f'python{py_ver[0]}.{py_ver[1]}'
                                             ^
SyntaxError: invalid syntax
```

mrclary:
> There seems to be syntax error in one of your scripts:
> 
> ```
> File "setup.py", line 161
>     _py_ver = f'python{py_ver[0]}.{py_ver[1]}'
>                                              ^
> SyntaxError: invalid syntax
> ```

I saw that, but it did not make sense because the script runs fine on my local machine. It looks suspiciously like Python 2.7 is being referenced. I think it may be related to the errors in the previous "Install dependencies" step.
```
ERROR: Could not find a version that satisfies the requirement pyqt5<5.13 (from spyder==4.2.0.dev0) (from versions: none)
ERROR: No matching distribution found for pyqt5<5.13 (from spyder==4.2.0.dev0)
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
WARNING: Skipping spyder as it is not installed.
```
Why can't it resolve pyqt5 requirement? Why is here a Python 2.7 deprecation?

mrclary:
Okay, found the issue. As I suspected, `python` and `pip` are referencing `2.7` versions. I'll have to use `pip3` and `python3` commands.

mrclary:
Success! Sort of...
It seems that the only issue left is that Python 3.8.5 was used for the build but Python 3.7.9 was installed for the Github action. How does that work? Maybe it was just a cache issue with cancelling/restarting in the middle of the job? Hmm...

.github/workflows/installer-macos.yml
```diff
@@ -0,0 +1,74 @@
...
+    runs-on: macos-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Install Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7.9
```
ccordoba12, line 22: Please change this to to have the Python version you'd like to see for the installer.

installers/macOS/Readme.md
```diff
@@ -0,0 +1,77 @@
...
+I do not know exactly why an Anaconda installation does not work except that it has something to do with hardlinks.
+
+I recommend using [Homebrew](http://brew.sh/) to install pyenv and pyenv-virtualenv (if you plan to use these for virtual environment management).
+If you plan to use pyenv, you don't even need to install Python from Homebrew, since pyenv will install whatever python version you request.
+If you don't plan to use pyenv, then you will need to install Python from Homebrew or elsewhere.
+
+After installing Homebrew, run:
+```
+$ brew install pyenv, pyenv-virtualenv, xz
+```
```
ccordoba12, line 24: Please add blank lines above and below this block of code.

installers/macOS/Readme.md
```diff
@@ -0,0 +1,77 @@
...
+`xz` is a package that provides compression algorithms that Python should be built with to satisfy some packages, namely `pandas`.
+
+The Python frameworks must be copied to the stand-alone application, so if you use pyenv you must enable frameworks in any Python installation that you plan to use to build the Spyder app.
+```
```
ccordoba12, line 28: Please add a blank line above this one.

installers/macOS/Readme.md
```diff
@@ -0,0 +1,77 @@
...
+### Create Virtual Environment
+
+If you currently have any conda environment(s) activated, then deactivate them completely, i.e. you should not be in any conda environment, not even base.
+
+Create the virtual environment and populate it with the necessary package requirements.
+If you are using pyenv with pyenv-virtualenv, it will look like this:
+```
+$ pyenv virtualenv <python version> spy-build
+$ pyenv activate spy-build
+```
```
ccordoba12, line 41: Blank lines above and below.

installers/macOS/Readme.md
```diff
@@ -0,0 +1,77 @@
...
+If you are using venv, creating the environment will look like this:
+```
```
ccordoba12, line 43: Please add a blank line above this one.

installers/macOS/Readme.md
```diff
@@ -0,0 +1,77 @@
...
+$ python -m venv --clear --copies spy-build
+$ source spy-build/bin/activate
+```
+
+Now change your working directory to the macOS installation directory of this repo and install the necessary packages.
```
ccordoba12, line 48: I think this needs a bit of rephrasing now this directory is part of our repo.

installers/macOS/Readme.md
```diff
@@ -0,0 +1,77 @@
...
+```
+(spy-build) $ cd <path>/<to>/spyder/installers/macOS
+(spy-build) $ pip install -r req-build.txt -r req-extras.txt -c req-const.txt -e ../../
+(spy-build) $ pip uninstall -q -y spyder
+```
```
ccordoba12, line 53: Blanks above and below.

installers/macOS/Readme.md
```diff
@@ -0,0 +1,77 @@
...
+This will install the packages required for building Spyder and extra packages like matplotlib.
+Spyder is installed from this repository for the sole purpose of getting all of its dependencies.
+It must be uninstalled since py2app will use the repository when creating the application bundle.
```
ccordoba12, line 56: This sentence can be removed, right?

installers/macOS/Readme.md
```diff
@@ -0,0 +1,77 @@
...
+
+`req-build.txt` contains only those packages required to build the stand-alone application.
+`req-extras.txt` contains optional packages to include, if desired, for use in iPython consoles launched from the "Same as Spyder" environment.
+If you use external environments, such as conda, for your iPython consoles, you don't need `req-extras.txt`.
```
ccordoba12, line 60:
```suggestion
`req-extras.txt` contains optional packages to include, if desired, for use in IPython consoles launched from the "Same as Spyder" environment.
If you use external environments, such as conda, for your IPython consoles, you don't need `req-extras.txt`.
```

installers/macOS/Readme.md
```diff
@@ -0,0 +1,77 @@
...
+The build command also provides an option to exclude these packages, so you may install them and still build the application without them.
+`req-const.txt` contains package constraints that are known to cause problems with the build process if not satisfied.
+
+### Create the Standalone Application
+
+To create the standalone application and package it in a dmg disk image run:
+```
+(spy-build) $ python setup.py --dmg py2app
+```
```
ccordoba12, line 69: Blanks above and below.

installers/macOS/Readme.md
```diff
@@ -0,0 +1,77 @@
...
+Further documentation can be accessed via
+```
```
ccordoba12, line 71: Blank above.

installers/macOS/Readme.md
```diff
@@ -0,0 +1,77 @@
...
+(spy-build) $ python setup.py -h
+```
+
+After a whole lot of screen dump, and if everything went well, you should now have two files in the `installers/macOS/dist` folder of this repository:
+* Spyder.app
+* Spyder-\<Spyder version\> Py-\<Python version\>.dmg
```
ccordoba12, line 77: This is not correct, according to the generated artifacts. Please update it.

installers/macOS/dmg_settings.py
```diff
@@ -0,0 +1,22 @@
+# -*- coding: utf-8 -*-
+"""
+Settings file for building the dmg disk image.
+Adapted from dmgbuild example:
+    https://github.com/al45tair/dmgbuild/blob/master/examples/settings.py
+
```
ccordoba12, line 6: Please remove this line.

installers/macOS/setup.py
```diff
@@ -0,0 +1,201 @@
...
+    NotADirectoryError: [Errno 20] Not a directory: '<path>/Resourses/lib/
+    python38.zip/qtawesome/fonts/fontawesome4.7-webfont.ttf'
+spyder :
+    NotADirectoryError: [Errno 20] Not a directory: '<path>/Resources/lib/
+    python38.zip/spyder/app/mac_stylesheet.qss'
+spyder_kernels :
+    No module named spyder_kernels.console.__main__
+sphinx :
+    No module named 'sphinx.builders.changes'
+
```
ccordoba12, line 60: Remove this blank.

installers/macOS/setup.py
```diff
@@ -0,0 +1,201 @@
...
+            'keyring', 'parso', 'pygments', 'pyls', 'qtawesome', 'spyder',
+            'spyder_kernels', 'sphinx',
+            ]
+
+if args.make_lite:
+    INCLUDES = []
+    EXCLUDES = ['numpy', 'scipy', 'pandas', 'matplotlib', 'cython', 'sympy']
+else:
+    INCLUDES = ['numpy', 'scipy', 'pandas', 'matplotlib', 'cython', 'sympy']
+    EXCLUDES = []
```
ccordoba12, line 129: I think pip should be excluded in both cases (in case it's included) to forbid people messing with the installer.

installers/macOS/setup.py
```diff
@@ -0,0 +1,201 @@
...
+    'excludes': EXCLUDES,
+    'iconfile': iconfile,
+    'plist': {'CFBundleDocumentTypes': [{'CFBundleTypeExtensions': EDIT_EXT,
+                                         'CFBundleTypeName': 'Text File',
+                                         'CFBundleTypeRole': 'Editor'}],
+              'CFBundleIdentifier': 'org.spyder-ide',
+              'CFBundleShortVersionString': spy_version}
+}
+
+# copy main application script
```
ccordoba12, line 146:
```suggestion
# Copy main application script
```

ccordoba12 (commented):
Thanks a lot for this @mrclary! This looks pretty good!

.github/workflows/installer-macos.yml
mrclary: I don't understand your request. I set this to 3.7.9, which is what I expect to see for the installer. I believe that is the version that Spyder is officially supporting, correct?

installers/macOS/Readme.md
mrclary: I'm explaining that spyder is installed with the previous command for the purpose of installing its dependencies. But then it must be uninstalled so that `py2app` uses the local repository for app bundle creation and *not* one found in site-packages.

.github/workflows/installer-macos.yml
mrclary: I resolved the issue. It was related to the "shell: bash -l {0}" callout

mrclary:
@ccordoba12, I think this is ready now.

mrclary:
@ccordoba12, nevermind... still needs work

mrclary:
Okay, here's what I've learned.
* setting `shell: bash -l {0}` causes the default installed python to be used (brew Python 3.8.5) instead of the version set by `actions/setup-python@v2`. Why?
* Unfortunately, `actions/setup-python@v2` installs a python built without `--enable-framework` flag and therefore fails to create the application bundle. So, for now, I can only use the default installed Python.
* Without `-l` flag for the shell, `hdiutil` is unable to detach the DMG image resulting in `dmgbuild.core.DMGError: Unable to detach device cleanly`. Why?

So for now I will just leave the shell and login specification in the GitHub workflow since Spyder application bundles built on Python 3.8.5 work just fine and I see no issues. However, for future development I will need to look into the following.
* How to install a python version with `--enable-framework`. Perhaps installing `pyenv` and setting up a virtual environment will work. Though this may increase build time significantly if GitHub doesn't have that stuff in the cache
* Determine how the `-l` bash flag avoids the DMG detach error. This may useful to know if I have to eliminate the login flag in the future. Is there an environment variable responsible for this behavior?
* Is `Core Image is now using the software OpenGL renderer. This will be slow.` benign? Just a result of no GPU resource available? Or is this a harbinger of an underlying issue?

mrclary:
@ccordoba12, This is ready.

mrclary:
never mind.
What the heck is going on that `hdiutil` cannot detach the disk image?!

mrclary:
@ccordoba12 @dalthviz,
I've observed the following error when selecting "Show environment variables" from the console in both macOS and Windows standalone applications.

```python
  File "/Users/rclary/Applications/Spyder.app/Contents/Resources/lib/python3.8/spyder/plugins/variableexplorer/widgets/collectionseditor.py", line 265, in sort
    self.sizes = sort_against(self.sizes, self.keys,
TypeError: sort_against() got an unexpected keyword argument 'sort_key'
  File "/Users/rclary/Applications/Spyder.app/Contents/Resources/lib/python3.8/spyder/plugins/variableexplorer/widgets/collectionseditor.py", line 265, in sort
    self.sizes = sort_against(self.sizes, self.keys,
TypeError: sort_against() got an unexpected keyword argument 'sort_key'
  File "/Users/rclary/Applications/Spyder.app/Contents/Resources/lib/python3.8/spyder/plugins/variableexplorer/widgets/collectionseditor.py", line 265, in sort
    self.sizes = sort_against(self.sizes, self.keys,
TypeError: sort_against() got an unexpected keyword argument 'sort_key'
```

This is a result of a mismatch in the required version of spyder-kernels and the current release that it used for both standalone builds. This raises two points.
* spyder-kernels will have to be advanced and released immediately prior to Spyder 4.2 (although, I'm sure you are already planning this)
* we need to consider how to determine which spyder-kernels (and python-language-server) to use for standalone builds.
  * For release events, this is not an issue, since I think we will all agree that release versions of spyder, spyder-kernels, and python-language-server will be synchronized.
  * However, pull-request builds are more complicated: for a given PR, should the release versions of spyder-kernels and python-language-server be used or should they be installed from the subrepos? The latter makes the most sense to me, since any PR that depends on those changes should include an update to the subrepos.

What are your thoughts?

mrclary:
@dalthviz , windows installer is failing and I don't know why. I haven't touched your stuff, I swear! ;-)

dalthviz:
@mrclary seems like the artifact failed to be uploaded. I trigger a re-run and seems like now all is green :+1: Thanks for working on this one!

mrclary:
I worked around the `dmgbuild.core.DMGError: Unable to detach device cleanly` by increasing the number of attempts that `build_dmg` tries to detach the disk image. I also try to capturing the error and supply a warning instead. So far, a warning has not been issued, so perhaps increasing the number of tries has worked.

Nevertheless, the two warnings persist:
* `Core Image is now using the software OpenGL renderer. This will be slow.`
* `hdiutil: detach: timeout for DiskArbitration expired`

I have not been able to diagnose these, but the uploaded artifacts seem to be okay.

mrclary:
@ccordoba12, I've made quite a few changes to the app creation script, so you may want to review it again.
But everything seems to be working and I think we're good to go.

.github/workflows/installer-macos.yml
```diff
@@ -0,0 +1,70 @@
...
+      DISTDIR: ${{ github.workspace }}/dist
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+#      - name: Install Python
+#        uses: actions/setup-python@v2
+#        with:
+#          python-version: 3.7.9
```
ccordoba12, line 32: If this is no longer necessary, please remove it.

ccordoba12 (commented):
Thanks @mrclary for all your hard work on this one! I'm really glad that you were able to include our subrepos when creating the installers on pull requests. That was the right thing to do, but we totally missed it with @dalthviz.

I left a minor comment, otherwise looks good to me.

.github/workflows/installer-macos.yml
mrclary: Okay, no problem.
Is there a reason for using 3.7.9? Should I pursue using specified Python versions?

ccordoba12: There's one, but only apply to Windows. The latest version Python version supported by [Pynsist](https://github.com/takluyver/pynsist/), the package we use to create our Windows installer, is 3.7.

ccordoba12:
@mrclary, please rebase this one so the installer can get the fixes in PR #14033.

steff456:
Hi @mrclary!

I just tried the full installer and found a couple of things that are not working as expected, 
1. The autocompletion engine is not working
2. Spyder opens in a project I didn't create, and the temp file is outside that project. I'm not sure if we want to start spyder on a "new" project
image
3. I have this traceback in the ipython console:
image

ccordoba12:
> I have this traceback in the ipython console:

This should be solved by adding Pillow to the app.

steff456:
The Spyder-lite installer is working perfectly 🎉

mrclary:
@steff456, thanks for the feedback.
Yeah, I noticed that Spyder starts up with a bogus project. I just deleted it from the Projects menu; it seems to happen only the first time Spyder starts up and doesn't return. I don't know why that is or how to reproduce it. I suspect that if I reset spyder preferences that wouldn't happen, but I'm not sure.

Can you tell me the following?
1. Are you using the "Lite" version or the "Full", i.e. from Spyder-Lite.dmg or Spyder.dmg?
1. Are you using the "Same as Spyder" interpreter or an external one? If external, can you provide a list of installed packages?

@ccordoba12, I'll check on including Pillow with the full version.

steff456:
The errors I found appeared when using the "full" version dmg, the `spyder-lite.dmg` is working as a charm 🤩. I reset my Spyder configuration before running the installed spyder, so the changes I've been working on doesn't affect the behavior of the installer. So, everything is running with the default interpreter.

ccordoba12:
> The autocompletion engine is not working

I think this is also failing because Pillow is not part of the installer. That's because the PyLS imports several modules during initialization to speed up completions. But if some of them fail to be imported, the server dies completely.

That error was fixed by @dalthviz in https://github.com/palantir/python-language-server/pull/873. I'll update our subrepo with it.

> Spyder opens in a project I didn't create

This is worrisome. I'll try to give you a hand with that problem during the weekend.

mrclary:
@steff456 @ccordoba12, the latest commit should fix the issue with `PIL`. I did not see any completion issues, though. Screen shot of a local build shows `from PIL import Image` and command completion of `math.`. Please check that the artifact from my last commit gives you the same result.

FYI, on this local build, when first launching, I did *not* get a bogus project. I think it may be related to related to launching a newly downloaded Spyder application for the first time (you know how you get the warning message about an unknown developer). It does not seem to be related to whether Spyder had an open project on the last session.

Screen Shot 2020-10-22 at 9 29 03 PM

ccordoba12 (approved):
Thanks a lot @mrclary for your contribution!

@juanis2112 tested the installer generated by your last commit and found the issues reported by @steff456 are fixed now.

mrclary:
Thanks to all the core developers for their feedback! I'm glad I could contribute.

PR #14013, opened 2020-10-19
PR: Don't interrupt while copying (IPython console)

impact27:
## Description of Changes
One annoyance I have with the console is that when the code is not executing, ctrl + c means copy, and when executing, ctrl + c means interrupt. This changes the behaviour so that if text is selected, it will copy rather than interrupt.

Should that be an option?

### Issue(s) Resolved

Fixes #

### Affirmation

### DIFF
--- a/spyder/plugins/ipythonconsole/widgets/shell.py
+++ b/spyder/plugins/ipythonconsole/widgets/shell.py
@@ -677,7 +677,9 @@ the sympy module (e.g. plot)
 
     # ---- Private methods (overrode by us) -----------------------------------
     def _event_filter_console_keypress(self, event):
-        """ Reimplemented for execution interruption.
+        """
+        Reimplemented to avoid interrupting execution when trying to
+        copy text.
         """
         key = event.key()
         if self._control_key_down(event.modifiers(), include_command=False):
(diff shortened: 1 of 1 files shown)


spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -676,6 +676,21 @@ def request_restart_kernel(self):
         self.ipyclient.restart_kernel()
 
     # ---- Private methods (overrode by us) -----------------------------------
+    def _event_filter_console_keypress(self, event):
+        """ Reimplemented for execution interruption.
+        """
```
ccordoba12, line 681:
```suggestion
        """
        Reimplemented to avoid interrupting execution when trying to
        copy text.
        """
```

ccordoba12 (commented):
Great idea @impact27! I left a minor suggestion, otherwise looks good to me.

impact27:
> Great idea @impact27! I left a minor suggestion, otherwise looks good to me.

I think it should be changed to work on mac as copy is cmd + c not ctrl + c

ccordoba12:
Agreed.

impact27:
Closed in favour of https://github.com/jupyter/qtconsole/pull/447

PR #14014, opened 2020-10-19
PR: Add get_package_version for more complete version list

oscargus:
## Description of Changes

I was a bit confused why not all dependency versions were listed.

Added a method `get_package_version` to get the version based on the package name, see third answer here:
https://stackoverflow.com/questions/20180543/how-to-check-version-of-python-modules

One could have added the code in the `get_module_version` method, but I thought that in the long run one may want to change the version detection to check the package rather than the module (as that in general should give a more consistent result), but I couldn't dig into the code that much right now.

image

I assume that maybe one would like to add version requirements for the dependencies that do not have it at the moment now that it can be checked, but no idea which versions are required.

No idea how to test it in a general way.

### Affirmation

I certify the above statement is true and correct: oscargus


ccordoba12 (commented):
Thanks a lot for this @oscargus! It's a great idea!

Since `pkg_resources` is not a standard library module, you need to add it to

* setup.py
* requirements/conda.txt
* binder/environment.yml
* spyder/dependencies.py

Unfortunately you can't get `pkg_resources` version using `pkg_resources`, but I think that's fine. So please don't set any version for it in those files.

Also in those files, please set the required versions of the following packages we couldn't get their versions before:

* intervaltree: >=3.0.2
* keyring: >=17.0.0
* pyls_black: >=0.4.6
* watchdog: >=0.10.3

oscargus:
Thanks for the detailed feedback! I have changed accordingly and also added a small test, primarily for the dependency.

oscargus:
Btw, I found references to pyyaml in setup.py and binder/environment.yml (and requirements/tests.txt), but not in requirements/conda.txt or in the code. Is this a package that is still used?

oscargus:
HTTP-error. I assume that you can restart that job (and only that job). Seems to make more sense than restarting everything. (Unless more things should be changed.)

setup.py
```diff
@@ -206,17 +206,18 @@ def run(self):
...
     'python-language-server[all]>=0.35.0,<1.0.0',
-    'pyls-black',
+    'pyls-black>=0.4.6',
     'pyxdg>=0.26;platform_system=="Linux"',
     'pyzmq>=17',
     'qdarkstyle>=2.8',
     'qtawesome>=0.5.7',
     'qtconsole>=4.7.7',
     'qtpy>=1.5.0',
+    'setuptools>=39.0.0',
```
ccordoba12, line 216: So i guess `pkg_resources` is a dependency of `setuptools` since this version, right? Just checking with you.

spyder/utils/tests/test_programs.py
```diff
@@ -282,5 +283,10 @@ def test_open_files_with_application(tmp_path):
         return_codes = open_files_with_application('not-valid.ext', fnames)
 
 
+def test_get_package_version():
+    # Primarily a test of pkg_resources being installed properly
+    assert get_package_version('IPython')
```
ccordoba12, line 288: Please check instead with `pyls_black`, which doesn't have a `__version__` attribute and probably won't have one in the future.

ccordoba12 (commented):
Thanks @oscargus for addressing my comments! I left two more, then this should be ready.

setup.py
oscargus: Correct. Not sure which version is actually required, but saw some Stack exchange answer where 39 was suggested, but more or a guess.

spyder/utils/tests/test_programs.py
oscargus: OK! This tests get_package_version and not get_module_version, so it do test the expected thing. But it can indeed make sense to test get_module_version like that as well.

setup.py
ccordoba12: Ok, good to know. Thanks for digging into that.

spyder/utils/tests/test_programs.py
ccordoba12: Oh, I see. Please please leave this test as it is and add another for `get_module_version('pyls_black')`.

ccordoba12 (approved):
Thanks @oscargus! Great addition!

PR #14033, opened 2020-10-22
PR: Fix issue where external IPython consoles do not launch for macOS application

mrclary:
## Description of Changes
Remove `PYTHONPATH` from `env_vars` passed to external IPython consoles on the macOS application.

Commit cb13febc87d0b365d01ee184aaaf536076f96d48 introduced an error in which Spyder's runtime PYTHONPATH was passed to the kernel preventing external IPython consoles from launching.

Screen Shot 2020-10-21 at 5 59 57 PM

Now that the Python Path Manager values no longer overwrite `PYTHONPATH` in `env_vars`, we can just drop it for macOS app external consoles.

### Affirmation

mrclary

### DIFF
--- a/spyder/plugins/ipythonconsole/utils/kernelspec.py
+++ b/spyder/plugins/ipythonconsole/utils/kernelspec.py
@@ -192,6 +192,7 @@ class SpyderKernelSpec(KernelSpec):
         # macOS app considerations
         if running_in_mac_app() and not default_interpreter:
             env_vars.pop('PYTHONHOME', None)
+            env_vars.pop('PYTHONPATH', None)
 
         # Making all env_vars strings
         clean_env_vars = clean_env(env_vars)


ccordoba12:
This probably affects the Windows installer too. @dalthviz, could you check that?

mrclary:
@ccordoba12, actually, the paths in Python Path Manager are not added to `sys.path` for an external IPython console in the macOS application; I'm not sure why. This does not seem to be an issue when bootstrapping. 

It seems to work if I add `pypath` to `PYTHONPATH` instead of removing it.
```python
# macOS app considerations
if running_in_mac_app() and not default_interpreter:
    env_vars.pop('PYTHONHOME', None)
    env_vars.update({'PYTHONPATH': pypath})
```
But given your edits in the aforementioned commit, I think you intended to have the language server or spyder-kernels add the value of `SPY_PYTHONPATH` to `sys.path`, so I'm wondering why this doesn't happen for the macOS case. `os.environ` is what it should be.

ccordoba12:
> It seems to work if I add pypath to PYTHONPATH instead of removing it.

Unfortunately that will make the kernel crash if you have a file named `random.py` or `string.py` in one of the directories present in `pypath`.

> But given your edits in the aforementioned commit, I think you intended to have ~~the language server or~~ spyder-kernels add the value of SPY_PYTHONPATH to sys.path, so I'm wondering why this doesn't happen for the macOS case

That was the idea. I also don't understand why that doesn't work for the Mac app.

ccordoba12 (approved):
Thanks @mrclary! Merging so we can have a working console before merging PR #13992.

PR #14051, opened 2020-10-23
PR: Improvements to the About Spyder dialog

jnsebgosselin:
## Description of Changes
@ccordoba12 Feel free to revert any changes you don't agree with. I should have opened an issue first, but since I'm a little bit short on time these days, it is more efficient to open a PR directly, I apologize for that.

Also, I don't know how this will look in macOS or Linux.

1. Make text in the dialog selectable.
2. Change dialog logo to fit that used in the `Help` menu
3. Scale Spyder logo with a Smooth transform instead of Fast to improve how it look on-screen.
4. Don't hardcode the dialog size and let Qt determine the best size to fit the content.

image

### Affirmation

### DIFF
--- a/spyder/widgets/about.py
+++ b/spyder/widgets/about.py
@@ -143,7 +143,8 @@ class AboutDialog(QDialog):
 
         pixmap = QPixmap(get_image_path(icon_filename))
         self.label_pic = QLabel(self)
-        self.label_pic.setPixmap(pixmap.scaled(64, 64))
+        self.label_pic.setPixmap(
+            pixmap.scaledToWidth(64, Qt.SmoothTransformation))
         self.label_pic.setAlignment(Qt.AlignTop)
 
         btn = QPushButton(_("Copy to clipboard"), )
--- a/spyder/widgets/about.py
+++ b/spyder/widgets/about.py
@@ -151,7 +151,7 @@ class AboutDialog(QDialog):
         bbox = QDialogButtonBox(QDialogButtonBox.Ok)
 
         # Widget setup
-        self.setWindowIcon(ima.icon('tooloptions'))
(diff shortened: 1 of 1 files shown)


spyder/widgets/about.py
```diff
@@ -161,12 +164,11 @@ def __init__(self, parent):
...
         btmhlayout.addWidget(bbox)
 
-        vlayout = QVBoxLayout()
+        vlayout = QVBoxLayout(self)
         vlayout.addLayout(tophlayout)
+        vlayout.addSpacing(25)
         vlayout.addLayout(btmhlayout)
-
-        self.setLayout(vlayout)
-        self.setFixedSize(410, 560)
```
ccordoba12, line 169: I like that our current About dialog has more height than width. Any chance to get that also in this PR?

ccordoba12 (commented):
I really like your changes, thanks @jnsebgosselin!

I only disagree with one of them, mentioned below.

spyder/widgets/about.py
jnsebgosselin: Maybe we can hardcode a value for the width but let the height free, so that there is not so much empty space between the text and the buttons?

jnsebgosselin: For example, if we add `self.label.setFixedWidth(350)`, then it looks like this on my system:

image

ccordoba12: > Maybe we can hardcode a value for the width but let the height free, so that there is not so much empty space between the text and the buttons?

I like that suggestion. Please go ahead with it.

ccordoba12 (approved):
Thanks @jnsebgosselin, looks good to me now.

Issue #13882, opened 2020-09-29
Saving layout doesn't store the right sizes.


juanis2112:
## Problem Description
When saving a custom layout in spyder and trying to restore it, it shows different sizes for the panes.

### What steps reproduce the problem?

1. Save a layout

<img width="1440" alt="Screenshot 2020-09-29 at 2 58 54 PM" src="https://user-images.githubusercontent.com/18587879/94609379-4efb2880-0264-11eb-80aa-2c45b33e6408.png">

2. Close spyder
3. Load saved layout from the Window layouts list.

### What is the expected output? What do you see instead?

The layout that it loads is different from the one I saved.

Also:

I saved a layout called "Layout" without changing anything after loading the "Default Spyder Layout". Taking a look at the transient.ini file where the layouts are stored. I see that the parameters of these two are different when they should be the same. 

<img width="908" alt="Screenshot 2020-09-29 at 3 03 15 PM" src="https://user-images.githubusercontent.com/18587879/94609869-055f0d80-0265-11eb-97a8-38da9452b750.png">

### Paste Traceback/Error Below (if applicable)

```python-traceback

PASTE TRACEBACK HERE

```

## Versions

* Spyder version: 
* Python version: 
* Qt version: 
* PyQt version: 
* Operating System name/version: 

### Dependencies

```
PASTE DEPENDENCIES HERE
```

wuhp06:
I find this bug too, and the newest version did not fix it.

If you maximize the Spyder window and then close it, and reopen it.  Spyder cannot restore the window position and size of the former window( normal state ).

I checked the source code, the bug is located in "plugins\layout.py" save_current_window_settings method of Layout class, and "app\mainwindow.py", the resizeEvent and  moveEvent method of MainWindow class seem useless, because self.isMaximized() return False when double clicking the window to maximize it.

PR #14078, opened 2020-10-27
PR: Use current size and position when saving window settings (Main Window)

ccordoba12:
This PR also does the following:

* Improve the style and docstrings of several layout related methods. Those were really bad, so I took the opportunity to improve them while I was reviewing how their code works.
* Improve how git displays diffs for Python files. That's something I just found on [this page](https://tekin.co.uk/2020/10/better-git-diff-output-for-ruby-python-elixir-and-more) and it was useful while reviewing the diffs generated by my changes.

@juanis2112, could you review if this PR fixes the problem you reported? It does on my manual tests.

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -1976,11 +1976,9 @@ class MainWindow(QMainWindow):
         self.toggle_layout('next')
 
     def toggle_layout(self, direction='next'):
-        """ """
-        get = CONF.get
-        names = get('quick_layouts', 'names')
-        order = get('quick_layouts', 'order')
-        active = get('quick_layouts', 'active')
+        names = CONF.get('quick_layouts', 'names')
+        order = CONF.get('quick_layouts', 'order')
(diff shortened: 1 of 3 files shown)


spyder/app/mainwindow.py
```diff
@@ -1632,26 +1644,31 @@ def set_window_settings(self, hexstate, window_size, prefs_dialog_size,
...
-        the userconfig-based configuration, under *section*"""
-        win_size = self.window_size
+        """
+        Save current window settings with `prefix` in
+        the userconfig-based configuration, under `section`.
+        """
+        # Use current size and position when saving window settings.
+        # Fixes spyder-ide/spyder#13882
+        win_size = self.size()
+        pos = self.pos()
```
ccordoba12, line 1654: This is the actual fix of PR #13882.

juanis2112 (approved):
I tested this and now layouts seem to be restoring fine! Thanks @ccordoba12

PR #14096, opened 2020-10-30
PR: Set column width automatically after changing autoformatter

### LINKED ISSUE #14082: Set column width automatically after changing autoformatter
To provide a better UX for autoformatting, I think we should apply some options related to the max allowed column width in the editor automatically, either when changing formatters or other optiions in Spyder.

Concretely, i think we should do the following:

* If people select Black, we should set **88** as the value of `Completion and linting > Code style > Maximum allowed length`. And this value should be **79** if people select Autopep8 or Yapf.
* When using Black, we should set **88** as the value of `Editor > Display > Show vertical lines at`. And this value should be **79** if people select Autopep8 or Yapf.
* If people manually set the value of `Completion and linting > Code style > Maximum allowed length`, we should tell the current formatter to use that value instead of the one they used by default. And if people select another formatter, that option needs to be passed to the new formatter too.
* If people manually set the value of `Completion and linting > Code style > Maximum allowed length`, we should also use that value to set `Editor > Display > Show vertical lines at`.
* If people manually set `Editor > Display > Show vertical lines at`, we shouldn't do anything because that's not related to autoformatting.

steff456:
## Description of Changes

Enhance the UI of the autoformatter options in the completion engine.

- [x] If people select Black, we should set 88 as the value of Completion and linting > Code style > Maximum allowed length. And this value should be 79 if people select Autopep8 or Yapf.
- [x] When using Black, we should set 88 as the value of Editor > Display > Show vertical lines at. And this value should be 79 if people select Autopep8 or Yapf.
- [x] If people manually set the value of Completion and linting > Code style > Maximum allowed length, we should tell the current formatter to use that value instead of the one they used by default. And if people select another formatter, that option needs to be passed to the new formatter too.
- [x] Remove the option from `Editor > Display > Show vertical lines at` and create a new one in the codestyle preferences that is called `Show vertical lines`

### Issue(s) Resolved

Fixes #14082

### Affirmation


steff456:
I'm closing this PR because I did a bad merge :(

PR #14104, opened 2020-10-30
PR: Customize tour dialog when starting Spyder

juanis2112:
## Description of Changes

Customized tour dialog according to https://github.com/spyder-ide/ux-improvements/issues/22
Screenshot 2020-10-30 at 3 47 00 PM

We decided to let the buttons not aligned with the text, but with the default alignment as that is how the rest of the dialogs are but we will probably fix all of them in the future.

### Issue(s) Resolved

Fixes https://github.com/spyder-ide/ux-improvements/issues/22

### Affirmation

Juanis2112


juanis2112:
@CAM-Gerlach can you take a look at this please and post a screenshot on how it looks like on Windows?

CAM-Gerlach:
Yes

CAM-Gerlach:
Quite different, and not in a good way unfortunately.

image

juanis2112:
I think we might need to use the fixed height and width for the dialog @ccordoba12. What do you think?

ccordoba12:
I think we need to adjust the dialog's style per operating system. This is how it looks on Linux (not so good either):

Selección_001

CAM-Gerlach:
@ccordoba12 To my examination, it looks about the same on Linux as on Windows, allowing for the lower screen resolution/higher scaling, but as on the latter, the dialog/margins is too small while the font is too large (vs on Mac it looks much better).

ccordoba12:
I think you're right. I'll do the proper adjustments on Linux then. I'll ping you when it's ready, so you can take a look again on Windows.

CAM-Gerlach:
As requested by @ccordoba12 and @juanis2112 , I fixed a number of layout issues to make the tour dialog consistent across platforms, matching the mockup by @isabela-pf in spyder-ide/ux-improvements#22 and the current macOS appearance (preferring the former, in the cases where there was a discrepancy):

* Adjust spacing between title and body, body and buttons, between the buttons, between the image and the text, and top & bottom of the dialog
* Fix the size, padding and alignment of the buttons to match those in spyder-ide/ux-improvements#22
* Set the dialog to an appropriate, fixed size and resolve the layout issues that resulted due to inappropriate stretching
* Reduce the font size of most elements to match spyder-ide/ux-improvements#22 

macOS and/or HiDPI-specific adjustments may be necessary, but aside from minor differences due to screen size and OS font, the two dialogs render essentially identically on Linux and Windows, as expected.

Windows (Win 8):

image

Linux (Binder):

image

CAM-Gerlach:
Per @juanis2112 's request, re-aligned the tour dialog buttons right to match the other dialogs.

image

ccordoba12:
> macOS and/or HiDPI-specific adjustments may be necessary, but aside from minor differences due to screen size and OS font, the two dialogs render essentially identically on Linux and Windows, as expected.

Thanks for your help with this @CAM-Gerlach! It looks great now on Windows and Linux!

But I'm kind of worried about Mac. Does it look good there too @juanis2112?

juanis2112:
This is how it looks now on Mac. I think it doesn't look good. Maybe we should customise it separately for Mac?

Screenshot 2020-11-02 at 11 35 08 PM

ccordoba12:
> Maybe we should customise it separately for Mac?

Yep, I think so too. We should leave the previous style for Mac and add a new one for Windows/Linux.

CAM-Gerlach:
I suspect it may have something to do with either a HiDPI screen or the current DPI scaling setting on macOS, rather than a static offset on the OS itself (therefore applying a static offset by OS would be inappropriate and be as likely to make things worse as better). To confirm this, you should test on different monitors (low vs high DPI) and with different DPI scaling settings after closing and re-opening Spyder.

ccordoba12:
Ok, thanks for the feedback. We'll ask @steff456 for help tomorrow (she has a high dpi monitor, if I'm not wrong).

juanis2112:
This is 1920 X 1200
Screenshot 2020-11-02 at 11 35 08 PM

ccordoba12:
Ok, thanks @juanis2112 (I didn't know you also have a high resolution monitor). Then we definitely need two styles.

CAM-Gerlach:
@ccordoba12 1200p 23 inch monitors are not HiDPI (my monitors are also that), but her computer's internal monitor (previous screenshots) are. Given the font size changed as expected with text scaling, and did not change between the hidpi and non hpi monitors, it does not appear to be due to that directly.

On my Windows machine, when I change `pt` units to `px` for font size, a ~1.33 scaling factor results in equivalent sizes (which likely comes from either 96dpi/72dpi, or 100/72dpi), and converting the `pt` numbers to `px` without conversion results in a similar result on Windows to that on macOS converting between the two at equivalent sizes on Windows. Therefore, it seems plausible that this is not a mere coincidence, and replacing the `pt` sizes with those in `px` multiplied by roughly 1.33 may produce comparable results on macOS. Not sure, but worth a shot.

Other thing I can think of is something weird due to using QSS instead of e.g. `setFont()` or global defaults, as used by other dialogs which scale correctly. @juanis2112 could also try setting the font size via that method, rather than QSS. @ccordoba12 any idea about that?

If neither of those resolve it, as a fixed offset of font size appears to be the only major difference between macOS and the other platforms, a crude but simple hack would be to simply apply a ~1.33x multiplier to the font sizes on macOS, inserting them into the QSS via e.g. format strings. However, given this is a rather crude hack that may not work reliably across OS versions, hardware, system configuration and displays, and we may run into this again on other dialogs in the future, we should at least try the the two potential quick fixes above, along with anything else obvious @ccordoba12 can think of, before giving up.

juanis2112:
I applied CAM's suggestions along with changing the width of the text so we keep it under two lines.

This is how it looks now on Mac. 
Screenshot 2020-11-03 at 8 37 52 PM

Please @CAM-Gerlach post a screenshot of how it looks on Windows with these changes.

CAM-Gerlach:
Windows:

image

Linux (Binder):

image

CAM-Gerlach (dismissed):
LGTM

ccordoba12 (changes requested):
@juanis2112, I think you're original design on Mac looks much better (i.e. it has better proportions and it's closer to @isabela-pf's design) than the last changes proposed by @CAM-Gerlach.

So I think we really need to go with different styles for macOS and Windows/Linux (I also don't like too much how it looks on Linux and think it can be improved).

CAM-Gerlach:
@ccordoba12 The primary intent of the changes I made were to make the layout match on all platforms and make it easily customizable, without the previous layout issues that arose when changing anything in the original layout. Its easy to tweak it to closely match the original on macOS (I was instead trying to match @isabela-pf 's mockup, which has somewhat different proportions) and I'm confused why you are suggesting the need for platform specific code since after these changes, the layout is now essentially identical across the three platforms (aside from the different fonts, which is out of our control, and modestly greater padding between the text and the buttons on macOS), so we can simply change them all at one to look however you prefer.

I'll go ahead with making them match the original look on macOS more closely, and we can work from there.

CAM-Gerlach:
Also @ccordoba12 note that the original on macOS one has a larger OS-level scaling setting applied than the more recent screenshot, which makes everything look a little bigger.

CAM-Gerlach:
I have made the necessary adjustments so that all three platforms now look essentially identical to the original macOS screenshot. The only delta with the original I have not yet fixed is that the icon is modestly smaller on the new version, and the only platform deltas are A. Differences in the font size and font scaling, which vary by OS, version and user preference outside our control, and B. macOS has a little extra padding between the text and the buttons, which a platform specific conditional can subtract 5-10 from the spacing value to resolve. 

Original macOS

image

Windows:

image

juanis2112:
This is how it looks now in Mac. 
Screenshot 2020-11-03 at 11 42 58 PM

CAM-Gerlach:
@juanis2112 @ccordoba12 I updated my comment to show screenshots of the dialog on each of the three platforms plus the desired target (original macOS) at the same size, to remove the impact of screen resolution differences. You can see they are minimal, except for a slight difference in the vertical space between the text and the buttons (which can be addressed with a conditional, if desired), and different fonts used which depends on the OS, version, and user preferences. Relative to the original macOS, the only remaining non-platform-caused difference is the image is a little smaller, which can probably be fixed with a little more tweaking.

We still didn't quite get the Linux text on two lines, but that will really depend on the font face, size and rendering used by the particular Linux distro and DE in question (would be helpful to see how it looks locally on your machine), and we can increase the fixed width of the text area further, to 360px-380 px to achieve this if desired, without really harming the rest of the layout.

ccordoba12:
At the end I decided to create different styles for Mac and Windows/Linux because that gives us more freedom to adjust things per platform.

Final styles:

Selección_023

Selección_022

spyder/app/tour.py
```diff
@@ -44,6 +44,8 @@
     MAIN_TOP_COLOR = QColor.fromRgb(230, 230, 230)
     MAIN_BG_COLOR = QColor.fromRgb(255, 255, 255)
 
+MAC = sys.platform == 'darwin'
```
CAM-Gerlach, line 47: Maybe a little out of scope, but do we not have a global constant for this? Should we add one? We check for Mac countless different places, so it might make sense to centralize it (as we do for a lot of other such constants) at least for new code going forward. Then its cleaner and if we ever need to adjust the check, we can do it one consistent place instead of 1000.

ccordoba12 (approved):
Thanks @juanis2112 and @CAM-Gerlach for the implementation, and @isabela-pf for the design. It looks great!

ccordoba12 (approved):

CAM-Gerlach:
@ccordoba12 I was in the process of reviewing it to verify it works on my machine and point out a few minor fixes in the code; given all the work we've put into this PR I don't understand why you'd merge substantial code changes without anyone else testing them or even looking them over.

ccordoba12:
I tested it in all platforms myself.

spyder/app/tour.py
```diff
@@ -1388,59 +1390,114 @@ def tour_has_focus(self):
...
         layout = QHBoxLayout()
         layout.addLayout(images_layout)
-        layout.addWidget(tour_label)
-        general_layout = QVBoxLayout()
-        general_layout.addLayout(layout)
-        general_layout.addLayout(buttons_layout)
+
+        label_layout = QVBoxLayout()
+        label_layout.addWidget(tour_label_title)
+        if not MAC:
```
CAM-Gerlach, line 1464: This block would be a easier to understand if it were `if MAC` and then `else`, instead of the other way around; since the brain usually works better with positives than negatives and its really unintiutive that `else` corresponds to `MAC`.

spyder/app/tour.py
```diff
@@ -1388,59 +1390,114 @@ def tour_has_focus(self):
...
+            label_layout.addSpacing(3)
+            label_layout.addWidget(tour_label)
+        else:
+            label_layout.addWidget(tour_label)
+            label_layout.addSpacing(10)
+
+        vertical_layout = QVBoxLayout()
+        if not MAC:
```
CAM-Gerlach, line 1472: Same here

spyder/app/tour.py
```diff
@@ -1388,59 +1390,114 @@ def tour_has_focus(self):
...
+            vertical_layout.addStretch()
+        else:
+            vertical_layout.addLayout(label_layout)
+            vertical_layout.addLayout(buttons_layout)
+
+        general_layout = QHBoxLayout()
+        if not MAC:
+            general_layout.addStretch()
+            general_layout.addLayout(layout)
+            general_layout.addSpacing(1)
```
CAM-Gerlach, line 1486:
```suggestion
```

I forgot to eliminate this line, it doesn't actually do anything

spyder/app/tour.py
```diff
@@ -1388,59 +1390,114 @@ def tour_has_focus(self):
...
+            vertical_layout.addLayout(label_layout)
+            vertical_layout.addSpacing(20)
+            vertical_layout.addLayout(buttons_layout)
+            vertical_layout.addStretch()
+        else:
+            vertical_layout.addLayout(label_layout)
+            vertical_layout.addLayout(buttons_layout)
+
+        general_layout = QHBoxLayout()
+        if not MAC:
```
CAM-Gerlach, line 1483: And same here

spyder/app/tour.py
```diff
@@ -1388,59 +1390,114 @@ def tour_has_focus(self):
...
+            general_layout.addLayout(vertical_layout)
+
         self.setLayout(general_layout)
 
         self.launch_tour_button.clicked.connect(self._start_tour)
         self.dismiss_button.clicked.connect(self.close)
+        self.setStyleSheet("background-color: #262E38")
+        self.setContentsMargins(18, 40, 18, 40)
+        if not MAC:
+            self.setFixedSize(640, 280)
```
CAM-Gerlach, line 1500: Are you sure its a good idea to rely on whatever size the OS happens to pick, which varies depending on the version, screen resolution, DPI and scaling settings? Shouldn't we set _something_ we control? Or does that mess up the layout?

CAM-Gerlach (commented):
Looks great to me, thanks @ccordoba12 . Just a few minor code comments.

I didn't notice it as much in the light theme Binder is using, but I assume there isn't anything we  can do about the ugly dialog border on Linux?

PR #14146, opened 2020-11-04
PR: Change default auto-formatter to be Black (Editor)

ccordoba12:
Black is the one that people have been requesting the most during the last months, so I think it's better to make it the default.


ccordoba12:
We decided to revert this one in PR #14147 because Autopep8 by default preserves our current column width length.

Issue #14136, opened 2020-11-03
Conda Environment Status Shows Full Path


mrclary:
## Problem Description
@steff456 
Environment status shows full path to conda environment, rather than just environment name.
Pyenv environment status shows correctly.

<img width="528" alt="Screen Shot 2020-11-03 at 8 40 33 AM" src="https://user-images.githubusercontent.com/9618975/98029880-126f9f00-1dc5-11eb-9e03-16ab80b02792.png">
<img width="262" alt="Screen Shot 2020-11-03 at 8 40 51 AM" src="https://user-images.githubusercontent.com/9618975/98029893-16032600-1dc5-11eb-9645-644cf1814580.png">

## Versions

* Spyder version: 4.2.0.dev0 0dfbe8967
* Python version: 3.7.9 64-bit
* Qt version: 5.9.6
* PyQt5 version: 5.9.2
* Operating System: Darwin 19.6.0

### Dependencies

```

# Mandatory:
applaunchservices >=0.1.7      :  0.2.1 (OK)
atomicwrites >=1.2.0           :  1.4.0 (OK)
chardet >=2.0.0                :  3.0.4 (OK)
cloudpickle >=0.5.0            :  1.6.0 (OK)
diff_match_patch >=20181111    :  20200713 (OK)
intervaltree >=3.0.2           :  3.1.0 (OK)
IPython >=4.0                  :  7.18.1 (OK)
jedi =0.17.2                   :  0.17.2 (OK)
jsonschema >=3.2.0             :  3.2.0 (OK)
nbconvert >=4.0                :  6.0.7 (OK)
numpydoc >=0.6.0               :  1.1.0 (OK)
parso =0.7.0                   :  0.7.0 (OK)
pexpect >=4.4.0                :  4.8.0 (OK)
pickleshare >=0.4              :  0.7.5 (OK)
psutil >=5.3                   :  5.7.2 (OK)
pygments >=2.0                 :  2.7.2 (OK)
pylint >=1.0                   :  2.6.0 (OK)
pyls >=0.35.0;<1.0.0           :  test-selector-01+92.g0dfbe8967 (OK)
pyls_black >=0.4.6             :  0.4.6 (OK)
pyls_spyder >=0.1.1            :  0.1.1 (OK)
qdarkstyle >=2.8               :  2.8.1 (OK)
qtawesome >=0.5.7              :  1.0.1 (OK)
qtconsole >=4.7.7              :  4.7.7 (OK)
[...]

Issue #199, opened 2015-02-16
Add  option "Copy to clipboard" for matplotlib figures


spyder-issues-migrator:
_From [dimitri....@gmail.com](https://code.google.com/u/112741511649297883635/) on 2010-04-13T06:36:42Z_

What is the expected output? 

In the interatic console , plot(arange(4)), select the figure 1 widget as 
the active one, hit CNTRL + C to copy the figure to the clipboard for 
further processing, e.g in word,...

A similar function was implemented in Matlab What version of the product are you using? On what operating system? 1.1.0beta1 

Please provide any additional information below

. A suggestion for coding it I found on the web somewhere: 

first some help function first :

def env(name): 
    return os.environ.get( name, '' ) 

def getHomeDir(): 
    if sys.platform != 'win32': 
        return os.path.expanduser( '~' ) 
 
    homeDir = env( 'USERPROFILE' ) 
    if not valid(homeDir): 
        homeDir = env( 'HOME' ) 
        if not valid(homeDir) : 
            homeDir = '&#37;s&#37;s' &#37; (env('HOMEDRIVE'),env('HOMEPATH')) 
            if not valid(homeDir) : 
                homeDir = env( 'SYSTEMDRIVE' ) 
                if homeDir and (not homeDir.endswith('\\')) : 
                    homeDir += '\\' 
                if not valid(homeDir) : 
                    homeDir = 'C:\\' 
    return homeDir 

Add the following code to the widget where your figure canvas is embedded , 
class MatplotlibFigure(PluginWidget) in fugure.py

        self.cb = None #will be used for the clipboard 
        self.tempPath = getHomeDir() 
        self.tempPath = os.path.join(self.tempPath,'tempMPL.png')
        self.mpl2ClipAction = QtGui.QAction("Save to Clipboard",  self) 
        self.mpl2ClipAction.setShortcut("Ctrl+C") 
        self.addAction(self.mpl2ClipAction) 
                  
QtCore.QObject.connect(self.mpl2ClipAction,QtCore.SIGNAL("triggered()"), 
self.mpl2Clip) 
 
    def mpl2Clip(self): 
        try: 
            self.canvas.fig.savefig(self.tempPath) 
            tempImg = QtGui.QImage(self.tempPath) 
            self.cb = QtGui.QApplication.clipboard() 
            self.cb.setImage(tempImg) 
        except: 
            print 'Error copying figure to clipboard' 
[...]

PR #14160, opened 2020-11-06
PR: Fix detection of conda and pyenv environments on Windows

ccordoba12:
@steff456, the detection of environments was failing on Windows due to some small issues. Please take a look at this PR to see if it looks ok for you.

@CAM-Gerlach, could you test if this works locally for you? Thanks!

### DIFF
--- a/spyder/utils/conda.py
+++ b/spyder/utils/conda.py
@@ -107,7 +107,8 @@ def get_conda_env_path(pyexec, quote=False):
 def get_list_conda_envs():
     """Return the list of all conda envs found in the system."""
     env_list = {}
-    conda = find_program('conda')
+    conda_exec = 'conda.bat' if WINDOWS else 'conda'
+    conda = find_program(conda_exec)
     if conda is None:
         return env_list
 
--- a/spyder/utils/conda.py
+++ b/spyder/utils/conda.py
@@ -14,7 +14,8 @@ import subprocess
 import sys
 
 from spyder.config.base import running_in_mac_app, get_home_dir
-from spyder.utils.programs import find_program, run_shell_command
+from spyder.utils.programs import (find_program, run_program,
+                                   run_shell_command)
 
 
 WINDOWS = os.name == 'nt'
@@ -120,15 +121,11 @@ def get_list_conda_envs():
         out = {'envs': []}
         err = ''
     for env in out['envs']:
-        name = env.split('/')[-1]
+        name = env.split(osp.sep)[-1]
         try:
-            path = osp.join(env, 'python') if WINDOWS else osp.join(
+            path = osp.join(env, 'python.exe') if WINDOWS else osp.join(
                 env, 'bin', 'python')
-            version, err = subprocess.Popen(
-                [path, '--version'],
-                stdout=subprocess.PIPE,
-                stderr=subprocess.PIPE
-            ).communicate()
+            version, err = run_program(path, ['--version']).communicate()
             version = version.decode()
             err = err.decode()
         except Exception:
--- a/spyder/utils/programs.py
+++ b/spyder/utils/programs.py
@@ -1094,11 +1094,7 @@ def get_list_pyenv_envs():
 def get_interpreter_info(path):
     """Return version information of the selected Python interpreter."""
     try:
-        out, err = subprocess.Popen(
-                [path, '-V'],
-                stdout=subprocess.PIPE,
-                stderr=subprocess.PIPE
-        ).communicate()
+        out, err = run_program(path, ['-V']).communicate()
         out = out.decode()
         err = err.decode()
     except Exception:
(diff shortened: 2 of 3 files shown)


steff456 (approved):
Thanks for this PR @ccordoba12, I totally forgot to add this support in the original PR. I'll test it tomorrow on my VM, but all the code looks really good!

CAM-Gerlach (changes requested):
I'm still not 100% sure of the precise exact behaviors I'm supposed to be validating (nor how it really didn't occur to anyone to test highly platform-dependent functionality on Spyder's biggest and most problematic platform), but as far as I can tell, it still doesn't work. I don't see any difference from spyder-ide/spyder#199 on any of the [things](https://github.com/spyder-ide/spyder-docs/pull/199#pullrequestreview-524773922) I mentioned in my review there. I tested with and without safe mode, and it appeared to make no difference either.

Let me know what you would like me to test further.

ccordoba12:
@CAM-Gerlach, if you don't have conda in your PATH and you installed Anaconda in a non-standard location, the it won't be possible for Spyder to list your envs.

Is that the case for you?

CAM-Gerlach:
Some additional info now that I've tried to sleep a little...

I thought it was intentional, but I'm also still experiencing #14136 where the status bar widget shows the full path, instead of just the env name, though it may be a side effect of whatever is resulting in these changes not working on my system. Other than that, and the ability to click the env name to get a popup to go to preferences, nothing else appears to be different (good or bad) with this PR, as I can still browse for envs, switch between them, start consoles in them, etc. but they are not autodetected).

image

Taking a quick look at the code, I don't see anything immediately obvious as to why; there are no spaces in my Miniconda path, it uses the default directory name, and contains `conda` which it looks like the code checks for. I also have all the new deps installed and no warnings on that.

Some potentially useful output in my default Spyder env, from within Spyder:

image

And outside of it:

image

ccordoba12:
Ok, I think we haven't added `C:\Miniconda3` to the list of locations we search for the conda executable.

I'll do that in a bit so you can test again.

CAM-Gerlach:
@ccordoba12 Sorry, your message didn't how up when I posted mine.

On Windows, conda is not added to the `PATH` by default; there is a non-default option for that in the installer but it warns strongly against it, so we shouldn't rely on this on that platform. I did install to `C:/Miniconda3` via the conda installer instead of the default suggested path (as I did with `C:/Anaconda3` before it) due to the conda installer warning me about the space in my username (chosen many, many years ago, before I knew better), but I've seen a significant fraction of other users on our issue tracker do the same, either to avoid spaces/special characters on their path that conda warned about, or for other reasons.

Instead of hardcoding `C:/(Ana|Mini)conda` or the default Anaconda/Miniconda install paths (which would only work for users who happen to choose that), the officially supported and most reliable way to get the path to the system's default Python install on Windows (Anaconda or otherwise) is with the `py` launcher that comes with Windows by default, which Anaconda/Miniconda automatically registers itself with as the default Python. Therefore, all you need to do is run the following command (works regardless of what env is activated), which I verified works on my system, to get the path to the Python in the base conda env; its parent directory on Windows is the Anaconda/Minconda directory

```bash
py -c "import sys;print(sys.executable)"
```

image

If for some reason you need a fallback to that, you could also simply check the `sys.executable` returned by Spyder's own copy of Python to get the base path to its Python install, either Anaconda/Miniconda, or its standalone Python (not sure what you do already in that case).

Speaking of which, I probably should test this using the Windows installer too? Where can I find that?

mrclary:
> Ok, I think we haven't added `C:\Miniconda3` to the list of locations we search for the conda executable.
> 
> I'll do that in a bit so you can test again.

That should already be in the search path...
`spyder.utils.programs.py:107`
```python
elif WINDOWS:
    pyenv = [osp.join(home, '.pyenv', 'pyenv-win', 'bin')]
    
    a = [home, 'C:', osp.join('C:', 'ProgramData')]
    b = ['Anaconda3', 'Miniconda3']
    conda = [osp.join(*p, 'condabin') for p in itertools.product(a, b)]

    req_paths.extend(pyenv + conda)
```

ccordoba12:
Thanks @mrclary for your help! I ran that code locally and I think I found the error: `C:` needs to be `C:\\` so it's joined correctly by `osp.join`.

CAM-Gerlach (approved):

mrclary:
> Thanks @mrclary for your help! I ran that code locally and I think I found the error: `C:` needs to be `C:\\` so it's joined correctly by `osp.join`.

Thanks @ccordoba12 for catching that! I'm sorry I missed it in the first place.

ccordoba12:
No worries. @CAM-Gerlach also helped me to found other thorny issues that we hadn't considered before. So the credit to make this one finally work goes to him too.

PR #14185, opened 2020-11-10
PR: Fix TypeError in text snippets

### LINKED ISSUE #14183: TypeError when getting text snippets
## Problem Description
The problem encountered by chance cannot be reproduced. It can be determined that the "sequence" is None under certain special circumstances.

image

## Versions

* Spyder version: 4.2.0
* Python version: 3.7.6
* Qt version: 5.12.9
* PyQt version: 5.12.3
* Operating System name/version: win10

ElieGouzien:
Might not be perfect, but helps with #14183.

As this bug makes Spyder completely unusable, I suggest this patch to be released as soon as possible.

## Description of Changes
I replaced
```python
if len(sequences) == 0:
    [1]
else:
    [2]
```

by
```python
if sequences:
    [2]
else:
    [1]
```

### Affirmation

@ElieGouzien

### DIFF
--- a/spyder/plugins/completion/snippets/trie.py
+++ b/spyder/plugins/completion/snippets/trie.py
@@ -21,26 +21,26 @@ class TrieNode:
         if len(self.key) > 0:
             sequence = sequence[1:]
 
-        if len(sequence) == 0:
-            self.value = value
-        else:
+        if sequence:
             elem = sequence[0]
             node = self.children.get(elem, None)
             if node is None:
                 node = TrieNode()
                 self.children[elem] = node
             node[sequence] = value
+        else:
+            self.value = value
 
     def __getitem__(self, sequence):
         node = None
         if sequence[0] == self.key:
             sequence = sequence[1:]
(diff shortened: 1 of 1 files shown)


ElieGouzien:
One additional remark: I didn't try to understand why `None` is used as the `sequence` argument  in `__getitem__()`. They might be a hidden bug that is not fixed here.

ccordoba12:
Thanks for your help @ElieGouzien! You said:

> As this bug makes Spyder completely unusable, I suggest this patch to be released as soon as possible.

In what circumstances is completely unusable? We ran a lot of tests and we didn't find this error.

ElieGouzien:
> In what circumstances is completely unusable? We ran a lot of tests and we didn't find this error.

I'm not anymore able to reproduce the bug. It happened just after upgrading from 4.1.6, on debian/testing using pip.

ccordoba12 (approved):
Thanks @ElieGouzien! I think your changes in this PR make a lot of sense and are straightforward too.

PR #12926, opened 2020-06-02
PR: Add safe-mode option to Spyder in non dev mode

juanis2112:
## Description of Changes

* [ ] Wrote at least one-line docstrings (for any new functions)
* [ ] Added unit test(s) covering the changes (if testable)

* [ ] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

Added safe-mode option to Spyder in non dev mode to solve issue #12631 

### Issue(s) Resolved

Fixes #12631 

TODO:

- [ ] Make `--reset` work without resorting to `CONF`. See [here](https://stackoverflow.com/questions/64198008/unicodedecodeerror-on-spyder-startup) for an example issue that will be solved by that.

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct:
Juanis2112

PR #14229, opened 2020-11-14
PR: Simplify creation of main application and splash screen

ccordoba12:
After PR #12926, I noticed we could simplify those two things a lot.

Benefits:
* It helps to make the code way more readable
* It avoids the hacks of using constants for the app and splash screen that need to be passed around in several places.


dalthviz (approved):
Thanks @ccordoba12 LGTM :+1: but just in case, this can be merged, right?

ccordoba12:
Thanks for the approval @dalthviz! Let me do the merge, because this has some nasty conflicts with master.

PR #14231, opened 2020-11-15
PR: Improve startup time a bit

ccordoba12:
This improves startup time a bit by doing the following:

* Delaying computation of missing dependencies 2 seconds after the window is up.
* Not running the check for optional dependencies when computing missing deps.

It also fixes a bug in `is_module_installed` that was making Spyder crash at startup if a module was not uninstalled correctly.


dalthviz (approved):
Thanks @ccordoba12 LGTM :+1:

Issue #14222, opened 2020-11-13
Cannot launch Spyder after updating to macOS 11 Big Sur, please help.


jzzhou715:
After updating my OS to macOS 11 today, I cannot seem to launch Spyder. I tried reinstall everything even reconstruct the python framework. Still does not work. When I run the following command in my terminal, the Spyder icon pops up but it stops there, nothing is being loaded and the Spyder window do not show up.

```
Last login: Fri Nov 13 16:07:21 on ttys000
Zhou@Jizhous-iMac ~ % Spyder/bin/spyder
```

I am not using Anaconda, I installed Spyder using pip.

How can I fix this? Thanks.

gvizkelethy:
Same here. On all three of my Macs the same thing happened. After upgrading to Big Sur Spyder starts but the IDE never shows up. When I click on the Spyder icon in the dock it  does not switch to Python in the title bar as before but stays in the terminal. I reinstalled my virtual environment, 3.9 then back to 3.8 it is the same. Any idea?

adamgarbo:
I'm also experiencing this issue when attempting to launch Spyder using Anaconda Navigator. After upgrading to Big Sur, the Spyder icon will appear in the Dock, but never load.

When attempting to load Spyder from the Terminal, it appears to never make it past line 3 of: **/Users/adam/opt/anaconda3/envs/icebergdrift/bin/pythonw**

```
#!/bin/bash
export PYTHONEXECUTABLE=/Users/adam/opt/anaconda3/envs/icebergdrift/bin/python
/Users/adam/opt/anaconda3/envs/icebergdrift/python.app/Contents/MacOS/python "$@"
```

Additionally, `spyder --debug-info verbose --debug-output file` does not produce any debugging information to the Terminal.

Keen to learn what the cause and/or solution is!

gvizkelethy:
Hi Adam,

I am not using anaconda, so it is a more general issue. I just launch Spyder from the command line and the same thing happen. I hope it gets fixed soon because I use Spyder every day.

George

jzzhou715:
> Same here. On all three of my Macs the same thing happened. After upgrading to Big Sur Spyder starts but the IDE never shows up. When I click on the Spyder icon in the dock it does not switch to Python in the title bar as before but stays in the terminal. I reinstalled my virtual environment, 3.9 then back to 3.8 it is the same. Any idea?

wenlzhang:
The same issue happened to me after updating to Big Sur.

[...]

Issue #14218, opened 2020-11-13
High typing latency in MacOS Big Sur


doublezz10:
### What steps will reproduce the problem?

I updated to MacOS Big Sur yesterday, and every since, typing in the editor has come with a great deal of latency - keystrokes occur nearly 0.5 seconds before characters appear in the editor. Interestingly, this issue is restricted to the editor - typing is fine in the console and in this issue reporter window.

Typing is slow whether in a brand new file or an existing one.

I tried updating spyder via conda, updating all of my other packages via conda update --all, and restarting my Mac multiple times.

## Versions

* Spyder version: 4.1.5 
* Python version: 3.8.1
* Qt version: 5.9.7
* PyQt5 version: 5.9.2
* Operating System: Darwin 20.1.0

### Dependencies

```

# Mandatory:
applaunchservices >=0.1.7      :  0.2.1 (OK)
atomicwrites >=1.2.0           :  1.4.0 (OK)
chardet >=2.0.0                :  3.0.4 (OK)
cloudpickle >=0.5.0            :  1.6.0 (OK)
diff_match_patch >=20181111    :  20200713 (OK)
intervaltree                   :  None (OK)
IPython >=4.0                  :  7.19.0 (OK)
jedi =0.17.1                   :  0.17.1 (OK)
nbconvert >=4.0                :  6.0.7 (OK)
numpydoc >=0.6.0               :  1.1.0 (OK)
parso =0.7.0                   :  0.7.0 (OK)
pexpect >=4.4.0                :  4.8.0 (OK)
pickleshare >=0.4              :  0.7.5 (OK)
psutil >=5.3                   :  5.7.2 (OK)
pygments >=2.0                 :  2.7.2 (OK)
pylint >=1.0                   :  2.6.0 (OK)
pyls >=0.34.0;<1.0.0           :  0.35.1 (OK)
qdarkstyle >=2.8               :  2.8.1 (OK)
qtawesome >=0.5.7              :  1.0.1 (OK)
qtconsole >=4.6.0              :  4.7.7 (OK)
qtpy >=1.5.0                   :  1.9.0 (OK)
rtree >=0.8.3                  :  0.9.4 (OK)
sphinx >=0.6.6                 :  3.2.1 (OK)
spyder_kernels >=1.9.4;<1.10.0 :  1.9.4 (OK)
watchdog                       :  None (OK)
zmq >=17                       :  19.0.2 (OK)

# Optional:
[...]

PR #14256, opened 2020-11-18
PR: Set QT_MAC_WANTS_LAYER env var to solve problems in Big Sur

impact27:
## Description of Changes
Sets QT_MAC_WANTS_LAYER to '1' when launching spyder. This will hopefully solve the launch issue on Big Sur. I am not 100% on what that does exactly, as I couldn't find doc about it.

### Issue(s) Resolved

Fixes https://github.com/spyder-ide/spyder/issues/14222
Fixes https://github.com/spyder-ide/spyder/issues/14218

### Affirmation

### DIFF
--- a/spyder/app/start.py
+++ b/spyder/app/start.py
@@ -114,6 +114,8 @@ def main():
         os.environ['QT_SCREEN_SCALE_FACTORS'] = ''
 
     if sys.platform == 'darwin':
+        # Fixes launching issues with Big Sur (spyder-ide/spyder#14222)
+        os.environ['QT_MAC_WANTS_LAYER'] = '1'
         # Prevent Spyder from crashing in macOS if locale is not defined
         LANG = os.environ.get('LANG')
         LC_ALL = os.environ.get('LC_ALL')


spyder/__init__.py
```diff
@@ -48,7 +48,12 @@
 DATAPATH = LOCALEPATH = DOCPATH = MATHJAXPATH = JQUERYPATH = ''
 
 
-import os
+import os, sys
+
+# Fixes launching issues with Big Sur (GH14222)
```
ccordoba12, line 53: Please change `GH14222` for `spyder-ide/spyder#14222` (that's our convention to cite issues in our codebase).

spyder/__init__.py
```diff
@@ -48,7 +48,12 @@
...
+if sys.platform == "darwin":
+    os.environ['QT_MAC_WANTS_LAYER'] = '1'
```
ccordoba12, line 55: Could you move this to `spyder/app/start.py`? We're already setting other environment variables there to fix other issues.

ccordoba12 (commented):
Thanks @impact27 for your help with this!

spyder/__init__.py
impact27: Ok, I am just not sure if this needs to be set before Qt is imported, but I'll try

impact27: It works

impact27:
This prints the following warning on brew python (which was working without this):
`Layer-backing can not be explicitly controlled on 10.14 when built against the 10.14 SDK`

ccordoba12:
> This prints the following warning on brew python

Thanks @impact27 for the check. Is that only a single warning or one that is repeated constantly?

ccordoba12:
> This prints the following warning on brew python

Thanks @impact27 for the check. Is that only a single warning or one that is repeated constantly?

impact27:
> > This prints the following warning on brew python
> 
> Thanks @impact27 for the check. Is that only a single warning or one that is repeated constantly?

This is a single warning

ccordoba12 (approved):
Thanks @impact27 for your help with this!

PR #14320, opened 2020-11-26
PR: Create status bar plugin (New API)

ccordoba12:
This PR does the following:

* Adds a new `statusbar` plugin to be in charge of adding/removing widgets to the main window status bar.
* Improve the way status bar widgets are created.
* Move the `InterpreterStatus` widget to the Main Interpreter plugin.


ccordoba12:
This is finally ready!

PR #14332, opened 2020-11-28
PR: Improve detection of conda and pyenv environments

### LINKED ISSUE #14330: Spyder freezes for a long period every time preferences is open due to checking conda envs
## Problem Description

After #13950 was implemented, Spyder freezes for nearly 15 s on my machine every time I open Preferences, with no pyenv environments and a moderate number (half dozen to a dozen) conda envs. This is a pretty big cost to pay for this feature, so per @ccordoba12 it should be moved to a separate QThread to avoid blocking the main thread.

1. Install Spyder on a machine with a modest number of conda envs
2. Attempt to open Preferences

Expected: GUI remains responsive with minimal lag (≈> 1 s). Actual: Large (15 s) freeze.

## Versions

* Spyder version: 4.2.0
* Python version: 3.7.9
* Qt version: 5.9.7
* PyQt version: 5.9.2
* Operating System name/version: Windows 8.1 x64

ccordoba12:
## Description of Changes

This PR does the following:

* Compute the list of envs every 60 seconds in a thread in the interpreter status widget. That will keep them updated.
* Cache that list to retrieve it quickly when opening our Preferences
* Add tests for the utility functions that detect environments.
* Update interpreter in status bar when it's removed outside Spyder.

TODO

- [x] Test manually on Windows to see if nothing broke there.

### Issue(s) Resolved

Fixes #14330

### Affirmation

### DIFF
--- a/.github/scripts/install.sh
+++ b/.github/scripts/install.sh
@@ -70,5 +70,11 @@ conda list -n jedi-test-env
 conda create -n spytest-ž -q -y python=3.6 spyder-kernels
 conda list -n spytest-ž
 
+# Install pyenv
+if [ "$OS" != "win" ]; then
+    curl https://pyenv.run | bash
+    $HOME/.pyenv/bin/pyenv install 3.8.1
+fi
+
 # Coverage
 pip install codecov
--- a/pytest.ini
+++ b/pytest.ini
@@ -23,3 +23,10 @@ markers =
     cython_client
     change_directory
     use_startup_wdir: Test startup workingdir CONF
+    preload_project: Preload a project on the main window
+    first
+    second
+    third
+    no_xvfb
+    external_interpreter
+    test_environment_interpreter
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -3404,7 +3404,15 @@ class MainWindow(QMainWindow):
         if CONF.get('main_interpreter', 'default'):
             return sys.executable
         else:
-            return CONF.get('main_interpreter', 'custom_interpreter')
+            custom = CONF.get('main_interpreter', 'custom_interpreter')
+
+            # Check if custom interpreter is stil present
+            if osp.isfile(custom):
+                return custom
+            else:
+                CONF.set('main_interpreter', 'custom', False)
+                CONF.set('main_interpreter', 'default', True)
+                return sys.executable
 
     # --- For OpenGL
     def _test_setting_opengl(self, option):
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -831,7 +831,6 @@ class MainWindow(QMainWindow):
             status,
             icon=ima.icon('environment'),
         )
-        self.interpreter_status.update_interpreter(self.get_main_interpreter())
 
         # Editor plugin
         self.set_splash(_("Loading editor..."))
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -831,6 +831,7 @@ class MainWindow(QMainWindow):
             status,
             icon=ima.icon('environment'),
         )
+        self.interpreter_status.update_interpreter(self.get_main_interpreter())
 
         # Editor plugin
         self.set_splash(_("Loading editor..."))
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -830,8 +830,8 @@ class MainWindow(QMainWindow):
             self,
             status,
             icon=ima.icon('environment'),
+            interpreter=self.get_main_interpreter()
         )
-        self.interpreter_status.update_interpreter(self.get_main_interpreter())
 
         # Editor plugin
         self.set_splash(_("Loading editor..."))
--- a/spyder/preferences/maininterpreter.py
+++ b/spyder/preferences/maininterpreter.py
@@ -25,7 +25,7 @@ from spyder.utils import icon_manager as ima
 from spyder.utils import programs
 from spyder.utils.conda import get_list_conda_envs_cache
 from spyder.utils.misc import get_python_executable
-from spyder.utils.pyenv import get_list_pyenv_envs
+from spyder.utils.pyenv import get_list_pyenv_envs_cache
 
 
 class MainInterpreterConfigPage(GeneralConfigPage):
@@ -40,7 +40,7 @@ class MainInterpreterConfigPage(GeneralConfigPage):
         self.cus_exec_combo = None
 
         conda_env = get_list_conda_envs_cache()
-        pyenv_env = get_list_pyenv_envs()
+        pyenv_env = get_list_pyenv_envs_cache()
         envs = {**conda_env, **pyenv_env}
         valid_custom_list = []
         for env in envs.keys():
--- a/spyder/preferences/maininterpreter.py
+++ b/spyder/preferences/maininterpreter.py
@@ -25,7 +25,7 @@ from spyder.utils import icon_manager as ima
 from spyder.utils import programs
 from spyder.utils.conda import get_list_conda_envs_cache
 from spyder.utils.misc import get_python_executable
-from spyder.utils.programs import get_list_pyenv_envs
+from spyder.utils.pyenv import get_list_pyenv_envs
 
 
 class MainInterpreterConfigPage(GeneralConfigPage):
--- a/spyder/preferences/maininterpreter.py
+++ b/spyder/preferences/maininterpreter.py
@@ -23,7 +23,7 @@ from spyder.preferences.configdialog import GeneralConfigPage
 from spyder.py3compat import PY2, is_text_string, to_text_string
 from spyder.utils import icon_manager as ima
 from spyder.utils import programs
-from spyder.utils.conda import get_list_conda_envs
+from spyder.utils.conda import get_list_conda_envs_cache
 from spyder.utils.misc import get_python_executable
 from spyder.utils.programs import get_list_pyenv_envs
 
@@ -39,7 +39,7 @@ class MainInterpreterConfigPage(GeneralConfigPage):
         self.pyexec_edit = None
         self.cus_exec_combo = None
 
-        conda_env = get_list_conda_envs()
+        conda_env = get_list_conda_envs_cache()
         pyenv_env = get_list_pyenv_envs()
         envs = {**conda_env, **pyenv_env}
         valid_custom_list = []
--- a/spyder/preferences/maininterpreter.py
+++ b/spyder/preferences/maininterpreter.py
@@ -236,7 +236,7 @@ class MainInterpreterConfigPage(GeneralConfigPage):
         custom_list = self.get_option('custom_interpreters_list')
         valid_custom_list = []
         for value in custom_list:
-            if osp.isfile(value) and programs.is_python_interpreter(value):
+            if osp.isfile(value):
                 valid_custom_list.append(value)
         self.set_option('custom_interpreters_list', valid_custom_list)
 
--- a/spyder/preferences/tests/test_config_dialog.py
+++ b/spyder/preferences/tests/test_config_dialog.py
@@ -9,8 +9,28 @@
 Tests for configdialog.py
 """
 
+# Standard library imports
+import os
+import time
 
-def test_config_dialog_save_to_conf(qtbot, global_config_dialog):
+# Local imports
+from spyder.preferences.configdialog import ConfigDialog
+from spyder.preferences.tests.conftest import MainWindowMock
+from spyder.utils.conda import get_list_conda_envs
+from spyder.utils.pyenv import get_list_pyenv_envs
+
+
+# Get envs to show them in the Main interpreter page. This is actually
+# done in a thread in the InterpreterStatus widget.
+# We also recording the time needed to get them to compare it with the
+# loading time of that config page.
+t0 = time.time()
+get_list_conda_envs()
+get_list_pyenv_envs()
+GET_ENVS_TIME = time.time() - t0
+
+
+def test_config_dialog_save_to_conf(global_config_dialog):
     for index in range(global_config_dialog.pages_widget.count()):
         configpage = global_config_dialog.get_page(index)
         configpage.save_to_conf()
@@ -20,3 +40,32 @@ def test_config_dialog_save_to_conf(qtbot, global_config_dialog):
             _save_lang()
 
         assert configpage.is_valid()
+
+
+def test_maininterpreter_page(qtbot):
+    from spyder.preferences.maininterpreter import MainInterpreterConfigPage
+
+    # Create Preferences dialog
+    dlg = ConfigDialog()
+    dlg.show()
+    qtbot.addWidget(dlg)
+
+    # Create page and measure time to do it
+    t0 = time.time()
+    widget = MainInterpreterConfigPage(dlg, main=MainWindowMock())
+    widget.initialize()
+    load_time = time.time() - t0
+
+    # Add page to Preferences
+    dlg.add_page(widget)
+
+    # Assert the combobox is populated with the found envs
+    assert widget.cus_exec_combo.combobox.count() > 0
+
+    # Assert load time is smaller than the one required to get envs
+    # directly. This means we're using the cached envs instead
+    assert load_time < GET_ENVS_TIME
+
+    # Load time should be small in our CIs because we have few envs.
+    if bool(os.getenv('CI')):
+        assert load_time < 1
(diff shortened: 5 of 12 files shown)


CAM-Gerlach (commented):
Testing results:

* [x] I am pleased to report that env detection now appears to work perfectly under Git bash on Windows.
* [x] When adding or removing an env, the list updates as expected; I wasn't able to find a timing where it didn't update
* [x] Testing did not reveal a significant period performance hit, either globally or across Spyder's that corresponded with the list being updated, with the possible exception of when things actually changed
* [x] The currently selected env displayed in the status bar and preferences gets reset to the Spyder built-in env essentially instantly if the current env is removed, which is as intended and _generally_ makes sense, though we might want to notify the user in some way.
* [ ] However, when simply upgrading the version of Python in an env, even just to a new patch version (e.g. 3.7.6 -> 3.7.9), the current env gets silently reset to the Spyder built-in one; presumably, there is a very brief interstitial window where the python binary is getting replaced where `osp.isfile(custom)` returns `False`. Therefore, unless the user carefully checks the status bar, they are likely to be unaware they are opening consoles in a different env than they expect, until their code mysteriously fails or they don't have the packages they expect and they finally figure it out (in fact, I was explictly testing the behavior of the status bar widget when doing this and watching it carefully, but since I was expecting the Python version to change, not the env, even I didn't notice that the env name had also changed). To address this, you could only reset to default if the interpreter remains gone after e.g. 60 s worth of checks.


### Performance

I tested the "hang time" opening preferences, measured from the time I clicked the preferences menu item to the time the UI became responsive again. I repeated each test three times and report both the mean and range of the tests. Tests were conducted on a 4c/8t i7-3630QM/24GB RAM machine with fast 850 Evo SSD.

``--safe-mode``/freshly opened, 4.2.0: 4.25 s (+/- 0.25 s)
``--safe-mode``/freshly opened, ``4.x``: 4.5 s (+/- 0.5 s)
``--safe-mode``/freshly opened, this PR: 6 s (+/- 0.25 s)
Non-safe mode, Spyder used for a while, 4.2.0: 11.5 s (+/- 0.5 s)

In conclusion, at least under these test conditions, this PR appears to *increase* the lag time opening preferences by an average of 1/3rd (33%, 15%-56%). If the 2.7x ratio between the preferences hang time of a freshly opened Spyder in ``--safe-mode` and a more typical use case is roughly consistent between the two branches (while the scaling is likely to vary, a ratio of 2-3x between these scenarios is consistent with past testing of preferences opening and apply times), extrapolation suggests under these conditions opening preferences may take ≈16 seconds with this PR, compared to only ≈12 on `4.x`.

Looking at system telemetry, I didn't see much of any I/O spike while preferences was loading and only a modest increase in CPU on this PR, while I did see a small jump in I/O and near max CPU load on one core running 4.2.0 in a dirty env, so it appears this problem is due to being blocked on something other than system resources.

ccordoba12:
Thanks @CAM-Gerlach for the review. My comments:

> The currently selected env displayed in the status bar and preferences gets reset to the Spyder built-in env essentially instantly if the current env is removed, which is as intended and generally makes sense, though we might want to notify the user in some way.

I think we should leave the notification part for Spyder 5 because things are going to change significantly once we implement https://github.com/spyder-ide/ux-improvements/issues/10 in a couple of months.

> However, when simply upgrading the version of Python in an env, even just to a new patch version (e.g. 3.7.6 -> 3.7.9), the current env gets silently reset

That's a situation I hadn't considered, but the fix is relatively simple. Thanks for pointing it out.

> In conclusion, at least under these test conditions, this PR appears to increase the lag time opening preferences

Two things about this:

* I don't understand how that's possible because we're using now the cached list of environments (so we are not computing them when opening Preferences). 
* I also measured the time to open preferences is this PR decreases it by half on my development machine (from 4.5 seconds to ~2 seconds). I'll post a couple of gifs to show the actual decrease in place.

CAM-Gerlach:
> I think we should leave the notification part for Spyder 5 because things are going to change significantly once we implement spyder-ide/ux-improvements#10 in a couple of months.

Yeah, agreed.

> I don't understand how that's possible because we're using now the cached list of environments (so we are not computing them when opening Preferences).

Yeah, I'm mystified by this. I tested on the latest `4.x` with all deps up to date, and the tests were both done in the same environment, from the same terminal shell, in a freshly opened copy of Spyder after waiting for ~60 seconds to ensure everything else settled down, with enough free RAM and nothing else major running consuming significant CPU resources, and the difference was very consistent (within a quarter to a half second) and persisted across numerous runs, both within the same Spyder session and after closing and re-opening. Since all else is equal, I can only conclude that for some reason getting the data from the cache is taking longer than doing so directly...somehow.

FWIW I tested from Anaconda prompt as well as Git bash and saw exactly the same results.

> I also measured the time to open preferences is this PR decreases it by half on my development machine (from 4.5 seconds to ~2 seconds). I'll post a couple of gifs to show the actual decrease in place.

Your time for `4.x` matches mine essentially exactly, so the fact that we're seeing opposite swings is very strange indeed. My first guess is something OS-related, assuming you're running on Linux (I tested on Windows) since certain operations, particularly involving command line calls, can take way longer on Windows, and there are major differences in how processes, threads amd IPC are handled (Windows doesn't have `fork` so starting new processes is expensive, and not all the Linux mechanisms for fast IPC, like named pipes and Unix sockets are implemented so it could be falling back to something slower). If you could test on Windows that would help rule that out (and I can dig up my old Linux VM if needed).

One other potential thing is are you running on the conda-forge stack with Qt 5.12? If so maybe something different with that? If maybe we can both create a new env with the latest deps and the same channel choose (CF or AD) we can rule out anything with that and narrow it down to something OS (or machine) specific. And let me know anything else I can test.

ccordoba12:
I don't know how you are running your tests, but the comparison below clearly shows that this PR improves things a lot:

| OS | This PR | 4.x |
| - | - | - |
| Windows | new windows | old windows |
| Linux | new linux | old linux |

Specs:

* Windows: 6 conda envs
* Linux: 13 conda envs + 1 pyenv

ccordoba12:
> One other potential thing is are you running on the conda-forge stack with Qt 5.12?

I don't think that's relevant in this case, but I'm Qt 5.9 in both cases.

CAM-Gerlach:
> I don't know how you are running your tests, but the comparison below clearly shows that this PR improves things a lot

I explain my test test methodology in some detail above.

Here are GIFs, with start times synced between GIFs, that again demonstrate the time difference, which going frame by frame in the video I measured to be 6.9 s for this PR, and 4.6 s for 4.x (measured from the first frame the `Tools` menu disappears after I clicked it, to the first frame the contents of the preferences window are drawn; it is at least another few hundred ms until the window is responsive but harder to measure precisely and objectively).

`4.x`, freshly-started Spyder, ``--safe-mode``

prefs_without_pr

This PR, freshly started Spyder, ``--safe-mode``

prefs_with_pr

> I don't think that's relevant in this case, but I'm using Qt 5.9 in both cases.

As am I, `defaults` channel.

ccordoba12:
@CAM-Gerlach, in my last commit I added a test to verify that the load time of the Main interpreter preferences page actually decreased with this PR.

Please copy `spyder/preferences/tests/test_config_dialog.py` to 4.x, change the import

    from spyder.utils.pyenv import get_list_pyenv_envs

to

    from spyder.utils.programs import get_list_pyenv_envs

and run it. In that case `test_maininterpreter_page` should fail due to the way @steff456 implemented this functionality in 4.2.0.

So, I can't account for your reported differences, but at least this shows the problem is on your side and not part of this PR.

CAM-Gerlach:
As requested:

Load time (4.x): 4.578 s
Get envs time (4.x): 2.920 s

Load time (this PR): 1.691 s
Get envs time (this PR): 3.234 s

So indeed, this time is being reduced. Also, the increase in lag I reported in #14330 is not that dramatic in ``--safe-mode`` with a freshly-opened Spyder; preferences load time there isn't that much less, at 3.75 s (+/- 0.25 s).

However, this helped lead me to the culprit: the reason it was faster on commits prior to this PR is it didn't work in Git bash at all, which cut out the lag loading envs completely. Testing it on the commit immediately prior to the start of the feature branch in this PR (to avoid any dependencies due to any different commits on `4.x` vs. this branch), clearly illustrates both the improvement this PR brings, and the residual lag that environment loading still introduces:

Last `4.x` commit on this branch, Git bash: 4.0 s (+/- 0.25 s)
Last `4.x` commit on this branch, Anaconda prompt: 8.5 s (+/- 0.25 s)

This PR, Git bash: 5,75 s (+/- 0.25 s)
This PR, Anaconda prompt: 5.5 (+/- 0.25 s)

These deltas match up very well with the numbers above from your test; the env load time is equal to essentially all the difference between 4.1.4 and 4.2.0, and all of it between Git bash (which didn't load envs at all) and Anaconda prompt (which did) pre-this PR. Meanwhile, the load time figure from the test for this PR is almost precisely the delta between the measured prefs lag on this PR and immediately prior. They also explain most of the large gap between the production copy of Spyder 4.2.0 and the safe mode tests of the same tag in Git bash.

CAM-Gerlach (approved):
While it would still be nice to see closer to parity with pre-4.2.0, this is still a substantial improvement in all respects, so LGTM.

CAM-Gerlach:
@ccordoba12 Let me know when you're ready for me to do a final re-test.

ccordoba12:
> While it would still be nice to see closer to parity with pre-4.2.0

We were also performing an extra validation for all interpreters in the Main interpreter page to be sure they really correspond to Python executables. But in my last commit I removed that (and moved it to `get_list_conda_envs` and `get_list_pyenv_envs` instead) given that it's impossible to set an invalid interpreter in our Preferences.

That should reduce the time to load Preferences even more.

> However, when simply upgrading the version of Python in an env, even just to a new patch version (e.g. 3.7.6 -> 3.7.9), the current env gets silently reset to the Spyder built-in one

This should be fixed too with the changes I pushed some hours ago:

* We fallback to the default interpreter only if the environment is deleted (not its interpreter).
* The interpreter's version should be updated if it's changed in the environment.

Please test that and let me know if things work as expected now.

CAM-Gerlach (commented):
I did some basic tests (opening prefs, switching envs, starting consoles, etc) and everything appears to work fine, with a couple exceptions.

> That should reduce the time to load Preferences even more.

Indeed it does, thanks; in fact it reduces load time all the way down to 4.25 s (+/- 0.25 s) which is faster than Git bash with 4.2.0 and within a small margin of the pre-4.2.0 value. Excellent stuff, better than I could have hoped for.

> * We fall back to the default interpreter only if the environment is deleted (not its interpreter).

This works to keep the same env if the interpreter is upgraded. However, in cases where `conda` leaves some files in the env when running `conda -n env-name remove --all` (as seems to happen 80+% of the time I remove one on Windows, and the thing you mentioned a lot of users face), the status bar item still displays the old env, while preferences and create new console both do fall back correctly to the default env.

> * The interpreter's version should be updated if it's changed in the environment.

This worked the first time I tested it in a session; I set Spyder to use a specific (non-builtin/default) env and downgraded then upgraded the version, and it updated both times. However, after switching envs and trying it again, both times I tested it (with different Python versions each time), it continues to show the old version even after waiting several minutes, opening a new console, and switching to another env and back again. I can't even confirm restarting Spyder fixes it, since I was testing in safe mode (though restarting a new safe mode session and reselecting that environment did display the right version, and subsequently updated).

image

ccordoba12:
> This works to keep the same env if the interpreter is upgraded. However, in cases where conda leaves some files in the env ... the status bar item still displays the old env

Ok, I didn't consider that case, I'll fix it.

> However, after switching envs and trying it again, both times I tested it, it continues to show the old version

Please describe the exact steps you followed to get this failure.

Also, if you start Spyder and don't switch envs, things work or not?

CAM-Gerlach:
Sorry for the lack of detail on that. Here is the procedure that reproduced it all three times I tried, with different envs and Python versions (3.7.x, 3.8.x) for each step each time (the ones shown are those for the most recent attempt)

0. Reset Spyder preferences
1. Open Spyder
2. Set environment in preferences to a non-default one with Python 3.7.9. Status bar updates accordingly
3. Use conda to downgrade Python version in env to 3.7.6. Python version shown updates
4. Use conda to upgrade Python version back to 3.7.9. Python version shown updates
5. Switch to a different non-default env in Preferences, with Python 3.8.5. Status bar updates accordingly
6. Use conda to downgrade Python version in env to 38.3. Python version shown in status bar does _not_ update

In addition, after all the following steps, the version shown for that env was still the old one, 3.8.5:
7. Wait 2 minutes.
8. Open a new console and verify that `Python 3.8.3` is printed in the prompt, `sys.executable` is the correct path to that env's Python, and `sys.version` shows `3.8.3`.
9. Switch to the 3.7 environment. Status bar updates accordingly.
10. Switch back to the 3.8 environment. Status bar now again shows 3.8.5 as the version (and the correct env name)

However, after closing and re-opening Spyder, the correct Python version for the 3.8 environment is now shown.

> Also, if you start Spyder and don't switch envs, things work or not?

I couldn't test that before since I was in safe mode, but I tested it now without, and yes, everything worked; the Python version in the environment was displayed correctly in the status bar after a restart and everything else works normally (as it did before).

CAM-Gerlach:
Also, if I after step 10 above then upgrade the env back to 3.8.5, it is now stuck showing 3.8.3 until I restart again.

ccordoba12:
Hey @CAM-Gerlach, could you test again? The commit I just pushed shoud fix the last problem you reported.

CAM-Gerlach (commented):
Retested following the procedure above, and unfortunately it seems they made things much worse, sorry. Listing environments is now completely broken when running Spyder from Git bash with none showing up (like I reported before this PR), and run from Anaconda prompt, from the very first time I tried changing the Python version (3.7.9 -> 3.7.6) Spyder thought the environment was removed and reverted back to the default.

If its too tricky to fix the last issue reported before, of the version in the status bar not updating, without introducing compromises its not the end of the world—it only shows the incorrect Python version for the duration of the user's session, and only if the user changes it while Spyder is open, and everything still works to open new consoles and change the interpreter in preferences, so its not a big deal.

ccordoba12:
@CAM-Gerlach, please test again. I improved a bit the way we're detecting an env was removed on Windows and removed the changes that were causing issues in Pyenv.

Detecting that the Python version was changed on Windows is probably not working, but we can work on that for 4.2.2.

CAM-Gerlach (changes requested):
Thanks so much, @ccordoba12 . I tested in Anaconda prompt and Git bash, and everything appears to work great now...with one very big exception (see below).

> Detecting that the Python version of the current env was changed on Windows is probably not working, but we can work on that for 4.2.2.

It actually _does_ work now in my testing of switching multiple envs and multiple versions; it doesn't happen instantly like it sometimes did before but it does update whenever the 60 second interval worker runs, which is fine.

_However_, this all comes with a major caveat: Around 50% of the times I start Spyder, in Anaconda prompt or Git bash, no environments are listed (not even the current one) and this feature doesn't work, just like the previous behavior in Git bash. Opening preferences, starting consoles, attempting to switch envs and waiting several minutes do not resolve the issue; the environments never show up in either the status bar or preferences.

Other than consistantly in Git bash prior to this PR, this is the first time I've encountered this issue after dozens of previous runs on this PR, as well as a few launches of Spyder 4.2.0 and Spyder 4.x under Anaconda prompt, mostly on `6dbf8786595a`/`cd74982690561b3` (which differ only by a test). To remove all other variables, I closed any other copies of Spyder, verified no other copies of Python or related binaries were running, confirmed none of the files anywhere in my Miniconda install were in use by any other program), and didn't do anything else while testing, and it still appeared seemingly at random.

I did a manual git bisect to find the problem commit, and here were the results (testing 7x each, 4x under Git bash and 3x Anaconda prompt):

`cd74982`: OK (7/7)
`b4b3ba0`: OK (7/7)
`fdb1885`: FAIL (1 attempt) 
`0234269`: FAIL (3 attempts) 
`4102154`: FAIL (1 attempt) 

Therefore, it seems `fdb1885` introduces the problem.

I pored through the ``--debug-info`` output, but didn't find anything useful. Let me know if there's something I can test further for you, thanks.

ccordoba12:
Thanks @CAM-Gerlach for running `git bisect` and finding the problematic commit. I removed from it the changes that introduced the bug.

So please test again to see if things are working as expected now.

CAM-Gerlach (approved):
LGTM, thanks for all your work on this @ccordoba12 . Everything works now in my further repeated testing on Anaconda prompt and Git bash, and I didn't see anything too egregious in a brief skim of the code, so it looks like this one is about ready to finally go in.

ccordoba12:
Great! Thanks a lot for all your help testing this! The initial implementation I had in mind improved significantly thanks to that.

PR #14376, opened 2020-12-06
PR: Bracket matching fixes (affects highlighting and autocompletion)

### LINKED ISSUE #1354: The editor adds ":" when type in multiply lines list comprehension
_From [ruoyu0...@gmail.com](https://code.google.com/u/112768690977086123209/) on 2013-04-14T21:39:40Z_

Spyder Version:  2.2.0rc
Python Version:  2.7.3
Qt Version:      4.8.3, PyQt4 (API v2) 4.9.5 on Windows

Type in multiply lines list comprehension such as:

```
[some_very_long_expression(x)
    for x in range(n)
    if n &#37; 2 == 0]
```

":" will be appended to the "for" & "if" lines.

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=1354_

### LINKED ISSUE #2965: Parenthesis highlight in the Editor is inconsistent
## Description

1. Write in a new file: `a=(")")`
2. Put the cursor at the end of the line, observe that the last parenthesis is highlighted in red
3. Put the cursor in the middle of `)"`, observe that the first parenthesis and the one in the string are highlighted in green.

The parenthesis highlighter picks up the parenthesis inside the string.

- Spyder Version: 3.0.0dev 
- Python Version: 3.5.1
- Qt Versions: 5.5.1, PyQt5 5.5.1 on Linux
## Dependencies

```
pyflakes >=0.6.0   :  1.0.0 (OK)
pygments >=1.6     :  2.1 (OK)
qtconsole >=4.0    :  4.1.1 (OK)
nbconvert >=4.0    :  4.1.0 (OK)
pandas >=0.13.1    :  0.17.1 (OK)
numpy >=1.7        :  1.10.4 (OK)
sphinx >=0.6.6     :  1.3.4 (OK)
psutil >=0.3       :  3.3.0 (OK)
rope >=0.9.4       :  0.9.4-1 (OK)
sympy >=0.7.3      :  None (NOK)
pylint >=0.25      :  1.5.4 (OK)

```

### LINKED ISSUE #9179: brace matching confused by strings with braces

The following code:
print('string with parenthesis)')

confuses the brace matching logic.

It thinks the parenthesis in the string is part of the parenthesis in the print function.

## Versions

* Spyder version: 3.3.4 
* Python version: 3.7.3
* Qt version: 5.12.2
* PyQt5 version: 5.12.1

### Dependencies

```
pyflakes >=0.6.0  :  2.1.1 (OK)
pycodestyle >=2.3 :  2.5.0 (OK)
sphinx >=0.6.6    :  2.0.1 (OK)
rope >=0.9.4      :  0.14.0 (OK)
jedi >=0.9.0      :  0.13.3 (OK)
nbconvert >=4.0   :  5.4.1 (OK)
numpy >=1.7       :  None (NOK)
IPython >=4.0     :  7.4.0 (OK)
```

### LINKED ISSUE #14374: Matching bracket highlighting index error
## Problem Description
When the final closing bracket (`)`, `]` or `}`) of a bracket pair is the first character in its line, any empty lines between the bracket pair are discarded during the search for the opening bracket leading to an indexing error of what character is being highlighted.

The error and a method of reproduction should be apparent from this image:
spyder-highlight-bug

## Versions
This happens in 4.2.0. See permalinks to code below for more details.

## Root cause
The bracket finder searches text that it gets from `BaseEditMixin.get_text` (lines 286 and 291)
https://github.com/spyder-ide/spyder/blob/436290ba71bb16c5928676467a358b70f4f38c10/spyder/plugins/editor/widgets/base.py#L282-L293

However, `BaseEditMixin.get_text` removes newlines under some conditions, resulting in the aberrant behaviour (lines 907-912)
https://github.com/spyder-ide/spyder/blob/436290ba71bb16c5928676467a358b70f4f38c10/spyder/widgets/mixins.py#L900-L913

## Possible fixes
I encountered this bug while working on #1354 and I plan to include a fix for this issue in the same pull request. But there are several alternatives for fixing it and I'd like some input on which is best.

1. Remove lines 907-912 from `BaseEditMixin.get_text`. I've tested it and it solves the problem, but it will cause other errors if the line removal behaviour is depended on elsewhere.
2. Make `all_text` an optional parameter with default value `None`. If `None`, it will be computed as previously.
3. Write a new function identical to `get_text` except without the line removal.

Ideally I would like to go with option 1, but that requires figuring out why the newlines are removed in the first place. If this behaviour is relied on elsewhere I think the newline removal should be moved there instead of kept in `get_text`. I don't think it's suitable to have a function called `get_text` that filters the text before returning it without having any parameters to control that behaviour.

Option 2 is safer and requires no investigation into side effects. So that's my second choice.

And while I'm here: How would I go about writing a regression test for a GUI feature like this? Where should I look for examples?

hengin:
## Description of Changes

I have made several changes to the editor plugin to fix some issues related to detection of brace matching. The main change being that unmatched brackets in strings and comments are now ignored:

spyder-highlight-fix

This was implemented by adding optional parameters to `TextEditBaseWidget.find_brace_match` that evaluate if the current brace should be ignored (for example, because it's in a string) or if the search should be stopped (for example, because the search left the starting string/comment).

In the final implementation, the optional parameters are not used by any callers, however. Instead `TextEditBaseWidget.find_brace_match` uses the class functions `TextEditBaseWidget.in_comment` and `TextEditBaseWidget.in_string` to define a reasonable default behaviour. These functions are overloaded by `CodeEditor.in_comment` and `CodeEditor.in_string`.

For additional details, see linked issues below.

### Caveats
While the cursor is in a string (or comment), it should not match braces outside strings (or comments). But should it only match braces inside the current string (or comment), or should it match braces inside any neighbouring strings (or comments)? I opted for the former because I wanted to keep brace matching as local as possible.

Edge cases:
```
a = "(" ")" <- No match because space in between

b = "("")"  <- Match because all characters in between the braces are considered to be in strings

# {

# } <- no match because newline in between

# {
# abcdef
# } <- Match because all characters in between braces are considered to be in comments
```

### Possible new issues
Evaluate if there's any performance hit to the current way of detecting comments/strings. UI lag must be minimized for a pleasing editing experience. I didn't see anything in my small examples, but for a large file with deep nesting of brackets, it could potentially be noticeable. Particularly in if the feature is used with languages other than Python.

An optional parameter was added to
https://github.com/spyder-ide/spyder/blob/436290ba71bb16c5928676467a358b70f4f38c10/spyder/widgets/mixins.py#L900
in order to be able to prevent newlines at the end of text blocks from being removed. Otherwise indices and editor positions did not match causing subtle bugs.

### Issue(s) Resolved

Fixes #1354
Fixes #2965 
Fixes #9179
Fixes #14374 

### Regression tests
For colon autocompletion, there was already one (added 5 years ago):
https://github.com/spyder-ide/spyder/blob/436290ba71bb16c5928676467a358b70f4f38c10/spyder/plugins/editor/widgets/tests/test_autocolon.py#L86-L93
So I just removed `@pytest.mark.xfail`.

A parenthesis/brace/bracket matching test was added at the end of `.../editor/widgets/tests/test_codeeditor.py`.

### Affirmation

By submitting this Pull Request and typing my username below,

I certify the above statement is true and correct: hengin


hengin:
All bugs fixed (I hope) and regression tests added. Automatic merge seems possible.

However, it might make sense to refactor this feature. As suggested in #2965.

ccordoba12:
Thanks a lot for this @hengin! It's awesome!

Please rebase or merge with `4.x` to get the fixes to our tests.

> However, it might make sense to refactor this feature. As suggested in #2965.

If you think it's important, please go ahead with it. You have time because I'm leaving this for 4.2.2 given that 4.2.1 is almost ready.

hengin:
Thanks for the encouragement @ccordoba12!

I don't think I will do any refactoring any time soon (not for a few months at least). I'd need to know the codebase a bit better to even have an idea if it's important or not. 

I did have some problems getting the tests to run. Mainly warnings, but one or two errors as well. In the end I used
```> python runtests.py spyder/plugins/editor/widgets/tests```
to run only editor tests. Seems like f8a7c31001 fixed much of it though.

File: spyder/plugins/editor/widgets/base.py [hunk 1]
```diff
@@ -278,46 +278,111 @@ def unhighlight_current_cell(self):
...
+            [ <-> ], ( <-> ), { <-> }
+        forward : boolean
+            Whether to search forwards or backwards for a match.
+        ignore_brace : callable taking int returning boolean, optional
+            Whether to ignore a brace (as function of position).
+        stop : callable taking int returning boolean, optional
+            Whether to stop the search early (as function of position).
+
+        If both ignore_brace and brace are None, the default behaviour
+        then brace matching is handled differently depending on whether
```
ccordoba12, line 318: It seems there's a word missing in the sentence that starts after the comma, because I don't get its meaning.

spyder/plugins/editor/widgets/base.py
```diff
@@ -278,46 +278,111 @@ def unhighlight_current_cell(self):
...
+        the starting position cursor is inside a string, comment or
+        regular code. The member functions self.in_comment and
+        self.in_string are used.
+
+        If exactly one of ignore_brace and brace is None, it is
+        replaced by function returning False for every position. I.e.:
```
ccordoba12, line 324: > replace by `a` function returning False for every position, i.e.

Just some small corrections to the text.

spyder/plugins/editor/widgets/base.py
```diff
@@ -278,46 +278,111 @@ def unhighlight_current_cell(self):
...
+        Returns
+        -------
+        The position of the matching brace. If no matching brace
+        exists, then None is returned.
+        """
+
+        if ignore_brace is None and stop is None:
+            if self.in_string(position=position):
+                # Only search inside the current string
+                def stop(pos): return not self.in_string(position=pos)
```
ccordoba12, line 336: Please reformat this function to be in two lines:

```python
def stop(pos):
    return not self.in_string(position=pos)
```

spyder/plugins/editor/widgets/base.py
```diff
@@ -278,46 +278,111 @@ def unhighlight_current_cell(self):
...
+            elif self.in_comment(position=position):
+                # Only search inside the current comment
+                def stop(pos): return not self.in_comment(position=pos)
```
ccordoba12, line 339: The same for this one.

spyder/plugins/editor/widgets/base.py
```diff
@@ -278,46 +278,111 @@ def unhighlight_current_cell(self):
...
+            else:
+                # Ignore braces inside strings and comments
+                def ignore_brace(pos): return self.in_string(position=pos) \
+                    or self.in_comment(position=pos)
```
ccordoba12, line 343: Please reformat this function as follows:

```python
def ignore_brace(pos):
    return (self.in_string(position=pos) or
            or self.in_comment(position=pos))
```

We don't accustom to use continuation lines anymore.

spyder/plugins/editor/widgets/base.py
```diff
@@ -278,46 +278,111 @@ def unhighlight_current_cell(self):
...
+                        # found matching brace, but should we have
+                        # stopped before this point?
+                        if stop is not None:
+                            # There's room for optimization here...
+                            for i in range(1, i_close+1):
+                                if stop(ind2pos(i)):
+                                    return
+                        return ind2pos(i_close)
+                    elif ignore_brace is None \
+                            or not ignore_brace(ind2pos(i_open)):
```
ccordoba12, line 384: Please reformat this as follows:

```python
elif (ignore_brace is None or
      not ignore_brace(ind2pos(i_open))):
```

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3998,8 +4017,21 @@ def next_char(self):
         next_char = to_text_string(cursor.selectedText())
         return next_char
 
-    def in_comment(self, cursor=None):
+    def in_comment(self, cursor=None, position=None):
+        """Returns true if the given position is inside a comment
```
ccordoba12, line 4021: Please change true for True and also add a period at the end.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3998,8 +4017,21 @@ def next_char(self):
         next_char = to_text_string(cursor.selectedText())
         return next_char
 
-    def in_comment(self, cursor=None):
+    def in_comment(self, cursor=None, position=None):
+        """Returns true if the given position is inside a comment
+
+        Overrides TextEditBaseWidget.in_comment
```
ccordoba12, line 4023: Please remove this line. It's not really necessary.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -3998,8 +4017,21 @@ def next_char(self):
...
-    def in_comment(self, cursor=None):
+    def in_comment(self, cursor=None, position=None):
+        """Returns true if the given position is inside a comment
+
+        Overrides TextEditBaseWidget.in_comment
+
+        Argument priority:
+            1. cursor
+            2. position
+            3. self.textCursor()
```
ccordoba12, line 4028: I don't understand this part of the docstring. Could you expand it so it's clearer what you mean by it?

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -4009,8 +4041,21 @@ def in_comment(self, cursor=None):
...
-    def in_string(self, cursor=None):
+    def in_string(self, cursor=None, position=None):
+        """Returns true if the given position is inside a string
+
+        Overrides TextEditBaseWidget.in_string
+
+        Argument priority:
+            1. cursor
+            2. position
+            3. self.textCursor()
```
ccordoba12, line 4052: Please make the same changes here as the ones I suggested for the docstring of `in_comment` method above.

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -264,5 +270,96 @@ def test_undo_return(editorbot):
     assert editor.toPlainText() == text
 
 
+def test_brace_match():
+    """Tests for the highlighting of matching parenthesis, braces and brackets
```
ccordoba12, line 274: Please add a period at the end.

spyder/plugins/editor/widgets/tests/test_codeeditor.py
```diff
@@ -264,5 +270,96 @@ def test_undo_return(editorbot):
...
+
+    The functions being tested are essentially:
+     * TextEditBaseWidget.find_brace_match
+     * CodeEditor.in_comment
+     * CodeEditor.in_string
+    """
+    # Create editor with contents from assets/brackets.py
+    app = qapplication()
+    editor = codeeditor.CodeEditor(parent=None)
+    editor.setup_editor(language='Python')
```
ccordoba12, line 296: This is not necessary. Instead, you need to pass as an argument to `test_brace_match` the fixture called `editorbot`. That's in charge is creating and setting up a CodeEditor instance for you.

With that, at this point you only need to write:

    __, editor = editorbot

to get a reference to the instance created by `editorbot` and use it below.

spyder/widgets/mixins.py
```diff
@@ -897,15 +897,25 @@ def get_text_region(self, start_line, end_line):
         start_cursor.setPosition(end_position, mode=QTextCursor.KeepAnchor)
         return self.get_selected_text(start_cursor)
 
-    def get_text(self, position_from, position_to):
+    def get_text(self, position_from, position_to, remove_newlines=None):
```
ccordoba12, line 900: I think we change this to `remove_newlines=True` because it's easier to understand the code below that way.

spyder/widgets/mixins.py
```diff
@@ -897,15 +897,25 @@ def get_text_region(self, start_line, end_line):
...
         """
         Return text between *position_from* and *position_to*
         Positions may be positions or 'sol', 'eol', 'sof', 'eof' or 'cursor'
+
+        Unless position_from='sof' and position_to='eof' any trailing newlines
+        in the string are removed. This was added as a workaround for
+        spyder-ide/spyder#1546 and later caused spyder-ide/spyder#14374
+        The behaviour can be overridden by setting the optional parameter
+        *remove_newlines* to False or True.
```
ccordoba12, line 909: Please remove `or True` here because it no longer applies.

spyder/widgets/mixins.py
```diff
@@ -897,15 +897,25 @@ def get_text_region(self, start_line, end_line):
...
+
+        TODO: Evaluate if this is still a problem and if the workaround can
+              be moved closer to where the problem occurs.
         """
         cursor = self.__select_text(position_from, position_to)
         text = to_text_string(cursor.selectedText())
-        all_text = position_from == 'sof' and position_to == 'eof'
-        if text and not all_text:
+        if remove_newlines is None:
```
ccordoba12, line 916: This can become now simply

```python
if remove_newlines:
```

ccordoba12 (commented):
Thanks @hengin for your help with this! It looks pretty good!

For the most part my comments and suggestions are related to formatting stuff and improving docstrings.

hengin:
Thanks for reviewing! It looks like I forgot to finish writing that first docstring. It was a mess.

I have made the suggested changes. I'll read through them again tomorrow night before pushing though.

ccordoba12 (approved):
Great work @hengin! These bugs have been in Spyder for years, so your help has been invaluable to finally fix them!

PR #14418, opened 2020-12-14
PR: Show a warning telling people how to get out of the pager (IPython console)

### LINKED ISSUE #14192: Pager when getting help blocks IPython console
## Problem Description
Use "?" in the console to view the function help, 
continue to operate in spyder, 
there is no output in the console, 
nor can it be input in the console, 
even if the kernel is restarted, the problem still exists.

The screen recording is as follows

spyder控制台输入输出bug

## Versions

* Spyder version: 4.2.0
* Python version: 3.6.7
* Qt version: 5.12.9
* PyQt version: 5.12.3
* Operating System name/version: win10

ccordoba12:
## Description of Changes

The IPython console has a pager to show long texts in a separate widget to not overflow it with them. It was deactivated by default, but I removed that option with PR #14056 because I thought the pager was broken.

However, our users quickly found out that was not the case and @OverLordGoldDragon suggested we should add a warning to tell people how to get out of the pager. And that's what I've done in this PR.

pager

### Issue(s) Resolved

Fixes #14192.

### Affirmation

### DIFF
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -43,10 +43,9 @@ from qtpy.QtWebEngineWidgets import WEBENGINE
 # Local imports
 from spyder import __trouble_url__, __project_url__
 from spyder.app import start
-from spyder.app.mainwindow import MainWindow  # Tests fail without this import
+from spyder.app.mainwindow import MainWindow
 from spyder.config.base import get_home_dir, get_conf_path, get_module_path
 from spyder.config.manager import CONF
-from spyder.preferences.runconfig import RunConfiguration
 from spyder.plugins.base import PluginWindow
 from spyder.plugins.help.widgets import ObjectComboBox
 from spyder.plugins.help.tests.test_plugin import check_text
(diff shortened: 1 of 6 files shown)


ccordoba12:
@isabela-pf, @juanis2112, I'd like to know your opinion about this feature. Is the idea of a pager grasped by you only by seeing the gif I posted above?

Unfortunately, I couldn't find a reference in the web we could link to better describe this.

OverLordGoldDragon:
Unsure the bar occupying the entire width is the best we can do. Few alternatives:

  0. (A) `Q return to Console`  -- (B) `Press Q to exit pager`; shorter text that helps the below, and ought to suffice as it should be pretty clear to user that "I'm not in console" (and thus the "pager" thing in B is where I'm at).
  1. Make the bar `<div>` half in width, and pin it to the _right_, away from text, maybe to top-right as user scrolls for new text from bottom.
  2. Move info to console row:

image

  3. Mark the Console with a star (like an edited file), and when hovered over, display the tooltip
  4. Maybe best of all: autohide the bar after 1-2 secs, then bring it up again only if the user presses any key (as 99% of people will trying to restore the console).

juanis2112:
Thanks @OverLordGoldDragon, we like your first suggestion. We will make the text be "Press Q to exit pager" but autohiding it can be confusing if people miss the warning. Also the space on top is used when there are several tabs open so using that place for the warning might not be the best idea.

ccordoba12:
@steff456 suggested that we should only add a blue border around the pager message. This is how things look with that change:

Selección_070

And this how they look with the old style and new text suggested by @OverLordGoldDragon:

Selección_071

I think I prefer the second option because it seems easy to miss the warning with the first (as it looks part of the paged text).

ccordoba12:
Here it's another suggestion: using a red border around the label

Selección_072

steff456:
By seeing all the options I also prefer the blue background

PR #14471, opened 2020-12-21
PR: Limit call rate document_did_change

impact27:
## Description of Changes
document_did_change is called on each keystroke, producing a small lag. This is visible if a lot of text is typed quickly in a large file. This limits the calls to document_did_change to 1 per second, as having sub-second accuracy on linting is not necessary (And the server takes a while to answer anyway).

### Issue(s) Resolved

Fixes #

### Affirmation


ccordoba12:
Why did you close this one @impact27?

impact27:
It didn't work as well as I thought, I need to fix it first

ccordoba12:
I think the problem is not with `document_did_change`, but with us requesting folding and symbols when linting results return from the server.

I'll try to propose a different solution this week.

ccordoba12:
@impact27, the lags introduced by `textdistance` should be fixed now with PRs #14574 and #14600. Could you test again and let us know if that's the case for you?

PR #14496, opened 2020-12-24
PR: Add Rtree to setup.py and update handling it in macOS installer

### LINKED ISSUE #14748: Add rtree to our setup.py dependencies
- Now that `rtree` has wheels for Python 2.7 - 3.9, we can add a dependency for it in our `setup.py`.
- We also need to increase the minimal required version to `>=0.9.7`.

mrclary:
## Description of Changes

Rtree 0.9.6 now includes the `libspatialindex.*` libraries in new wheels.
This means that we no longer need to install `spatialindex` in the build process.

Recommend adding necessary changes to the Windows installer to accommodate the update to Rtree.

@ccordoba12 @dalthviz 

Fixes #14748

### Affirmation

@mrclary


mrclary:
Looks like Rtree 0.9.6 is not yet available on conda. When it is (perhaps a few days?), I'll rerun tests.

mrclary:
Update: 0.9.7 is now available on conda-forge, but pkgs/main is still only at 0.9.4
I don't know how long it will take for pkgs/main to catch up...

binder/environment.yml
```diff
@@ -38,7 +38,7 @@ dependencies:
 - qtawesome >=0.5.7
 - qtconsole >=5.0.1
 - qtpy >=1.5.0
-- rtree >=0.8.3
+- rtree >=0.9.6
```
ccordoba12, line 41: Why `0.9.6` and not `0.9.7`? I say it because those versions were released only a few days apart:

https://pypi.org/project/Rtree/#history

So probably `0.9.7` includes fixes that the Rtree maintainers missed in `0.9.6`

ccordoba12 (commented):
@mrclary, thanks for your work on this. I left a minor comment but I really don't know how to make this work, unless we switch to Conda-forge to run our tests.

The problem is I don't have time to do that at the moment, so it'll have to wait until we release Spyder 5 next month.

binder/environment.yml
mrclary: Correct. This should be changed to 0.9.7.

mrclary:
> @mrclary, thanks for your work on this. I left a minor comment but I really don't know how to make this work, unless we switch to Conda-forge to run our tests.
> 
> The problem is I don't have time to do that at the moment, so it'll have to wait until we release Spyder 5 next month.

@ccordoba12, I recommend the following.
1. I'll add the rtree constraint to this PR
2. I'll temporarily change the testing pipelines to install `rtree` from the conda-forge channel
3. When tests pass, perhaps we release an alpha 5.0 to conda in order to force the conda adoption of rtree to 0.9.7
4. After conda updates, I can open another PR that reverts the conda-channel installation of rtree in the testing pipelines

Does that sound like it will work? If so, I can take care of everything except the release to conda.

ccordoba12:
Hey @mrclary, I think this part of your plan

> I'll temporarily change the testing pipelines to install rtree from the conda-forge channel

could not work. Unfortunately, that would lead to mix compiled packages from defaults and conda-forge (i.e. `libspatialindex` from conda-forge and the rest from defaults), which can cause segfautls or odd errors. You can try it to see how it goes though.

> When tests pass, perhaps we release an alpha 5.0 to conda in order to force the conda adoption of rtree to 0.9.7

The Anaconda team will take the new constraint into account when the final Spyder 5 version is released. They don't package alpha/beta versions.

So every PR from now on would use Rtree from conda-forge.

mrclary:
> Hey @mrclary, I think this part of your plan
> 
> > I'll temporarily change the testing pipelines to install rtree from the conda-forge channel
> 
> could not work. Unfortunately, that would lead to mix compiled packages from defaults and conda-forge (i.e. `libspatialindex` from conda-forge and the rest from defaults), which can cause segfautls or odd errors. You can try it to see how it goes though.

Okay, I'll give it a try in this PR. I don't think there should be any issues since `libspatialindex` is only used by `rtree` and `rtree` doesn't use any other libraries, as far as I know.

> > When tests pass, perhaps we release an alpha 5.0 to conda in order to force the conda adoption of rtree to 0.9.7
> 
> The Anaconda team will take the new constraint into account when the final Spyder 5 version is released. They don't package alpha/beta versions.
> 
> So every PR from now on would use Rtree from conda-forge.

I think my main concern is that users don't have issues installing Spyder; if there is a period where our 5.0 is out in the wild _before_ Anaconda adds `rtree` to their pkgs/main, then there could be a lot of confusion for those who don't have the conda-forge channel configured as part of their default channels. If Anaconda adopts the new version of `rtree` simultaneously with our 5.0 release, then users won't have to worry about it, and we can immediately revert our test pipelines after the release.

mrclary:
@ccordoba12, I've updated the rtree requirements to >=0.9.7 and included conda-forge channel. It looks like all tests have passed. Removing WIP.

ccordoba12:
> I think my main concern is that users don't have issues installing Spyder; if there is a period where our 5.0 is out in the wild before Anaconda adds rtree to their pkgs/main, then there could be a lot of confusion for those who don't have the conda-forge channel configured as part of their default channels

Ok, I understand your concern now, but that's what our installers are basically for, i.e. to save people the pain of dealing with this kind of issues.

However, I don't know if people could hit this problem because Spyder 5 will have this requirement, so the Anaconda team will have to package Rtree 0.9.7 before releasing 5.0.

ccordoba12:
I see another problem with this: this will change the procedure we defined in our Contributing guide to create a development environment because it won't give new contributors an environment with the right dependencies.

Perhaps we should switch change them in the meantime to use conda-forge? What do you think @mrclary?

mrclary:
> I see another problem with this: this will change the procedure we defined in our Contributing guide to create a development environment because it won't give new contributors an environment with the right dependencies.
> 
> Perhaps we should switch change them in the meantime to use conda-forge? What do you think @mrclary?

Do you mean update the Contributing guide to instruct contributors to add conda-forge channel when creating development environments? I think this is a good idea; I just pushed up that addition. However, we should alert existing contributors that they will need to install rtree from conda-forge channel.

mrclary:
@ccordoba12 I propose the following.
Add this PR to the 5.0.0 milestone, but wait until all other issues and PRs are merged in, leaving this to be last merge just before release. Then we can test and merge this in, it will be included in 5.0.0, and conda will be updated. Immediately following the release we can apply #14891, so there will not be any disruption for developers and no need for anybody to worry about using conda-forge. What do you think?

mrclary:
@ccordoba12, We're getting issues building the conda environment. Did I misunderstand something, or are the changes you made for conda today not ready?

ccordoba12:
I think the problem is there's no way to install `rtree` from our channel without the other deps conda-forge pulls from its own channel.

So the right solution is the one you implemented before (sorry for going in circles with this, but I thought my changes were fine). So please revert to that so I can merge this PR.

mrclary:
@ccordoba12, it looks like all the tests are passing, but one of the Mac tests was cancelled. Is there a way to resume or rerun a single test? Or do we have to rerun all pipelines?

ccordoba12:
Unfortunately, tests are hanging on macOS (no idea why). So I'll leave this for 5.0.1, to be released in a couple of weeks.

mrclary:
I just fixed something. I had forced using the conda-forge channel by mistake. I meant to just add it to the channel list. Let's see if that fixes the issue...

mrclary:
Hmm...it still seems to be pulling everything from conda-forge. I wonder if it wasn't the changes you made to the spyder-ide/label/dev channel yesterday?

ccordoba12:
You're probably right. Let me remove them and restart our CIs.

mrclary:
@ccordoba12 shall we try this again for 5.0.1?

ccordoba12:
I think the best we can do is to move our entire testing infrastructure to use Conda-forge. The quality of defaults has decreased significantly during the last months, and it'll also help us to avoid problems with our compiled dependencies, like this one.

So, do you want to help us to do that?

mrclary:
> I think the best we can do is to move our entire testing infrastructure to use Conda-forge. The quality of defaults has decreased significantly during the last months, and it'll also help us to avoid problems with our compiled dependencies, like this one.
> 
> So, do you want to help us to do that?

I can certainly help with that. What is needed? Can we just add the conda-forge channel to our build/test environments? I don't see anything else...

ccordoba12:
> I don't see anything else...

Yep and add it with higher priority than our own channel.

mrclary:
@ccordoba12, I don't know why the conda fast tests for Win and Mac are taking so long. There does not seem to be any output (yet?) either. I wonder if we should try setting the conda-forge channel to be lowest priority instead of highest priority...what do you think?

mrclary:
@ccordoba12, rtree 0.9.7 is now available on conda's main channel. I think this is ready to move forward and shouldn't present any issues for developers. I still have conda-forge channel set as the lowest priority channel; don't know if this channel is necessary anymore...

ccordoba12:
> I think this is ready to move forward and shouldn't present any issues for developers.

Yes, I think so too.

> I still have conda-forge channel set as the lowest priority channel; don't know if this channel is necessary anymore...

No, I don't it's necessary anymore. Please remove it.

CONTRIBUTING.md
```diff
@@ -60,7 +60,7 @@ $ workon spyder-dev
 After you have created your development environment, you need to install Spyder's necessary dependencies. The easiest way to do so (with Anaconda) is
 
 ```bash
-$ conda install -c spyder-ide/label/dev --file requirements/conda.txt
+$ conda install -c spyder-ide/label/dev --file requirements/conda.txt python.app
```
ccordoba12, line 63: Please remove `python.app` from this line because is not available in other OSes and add below

    If you are on Mac, you also need to run

    ```bash
    $ conda install python.app
    ```

ccordoba12 (commented):
@mrclary, I left a minor comment for you, otherwise it's ready.

ccordoba12 (approved):
Thanks @mrclary for your help with this!

PR #14596, opened 2021-01-15
PR: Migrate Files to the new API

### LINKED ISSUE #12195: Move Explorer/Projects plugin to use new API
- [ ] Move to new API
- [ ] Ensure tests pass

steff456:
## Description of Changes

Migration of the explorer plugin to the new API

### Issue(s) Resolved

Fixes #12195 

### Affirmation

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -1139,12 +1139,9 @@ class MainWindow(QMainWindow):
 
         # Explorer
         if CONF.get('explorer', 'enable'):
-            self.set_splash(_("Loading file explorer..."))
             from spyder.plugins.explorer.plugin import Explorer
-            self.explorer = Explorer(self)
-            self.explorer.register_plugin()
-            self.add_plugin(self.explorer)
-            self.preferences.register_plugin_preferences(self.explorer)
+            self.explorer = Explorer(self, configuration=CONF)
+            self.register_plugin(self.explorer)
 
         # Online help widget
         if CONF.get('onlinehelp', 'enable'):
@@ -3375,7 +3372,7 @@ def create_window(app, splash, options, args):
         QCoreApplication.setAttribute(Qt.AA_DontShowIconsInMenus, True)
 
     # Open external files with our Mac app
-    if sys.platform == "darwin":
+    if running_in_mac_app():
         app.sig_open_external_file.connect(main.open_external_file)
         app._has_started = True
         if hasattr(app, '_pending_file_open'):
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -42,6 +42,7 @@ from qtpy.QtWebEngineWidgets import WEBENGINE
 
 # Local imports
 from spyder import __trouble_url__, __project_url__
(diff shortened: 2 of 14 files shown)


juanis2112 (dismissed):
Great work on this @steff456!
I tested the files pane and I found some issues with this PR:
1. The color of the "activated state" in the filter button is different to what it is in master. 
This PR:
Screenshot 2021-01-26 at 6 00 30 PM
Master:
Screenshot 2021-01-26 at 8 10 39 PM

2. The state of the options "Show hidden files" and "Single click to open files" in the hamburger menu is not sync with the ones in the preferences dialog. For example, If I activate the  "Show hidden files" option in the preferences dialog, it effectively shows the hidden files, but the option doesn't appear "active" in the hamburger menu. (See screenshot)
Screenshot 2021-01-26 at 9 59 31 AM

3. (This one is actually not a bug from this PR cause I can reproduce this in master as well but is worth mentioning). After reseting spyder, by default the files opens in my user directory and when I click the "back" button it goes to the spyder directory. I'm not sure if this is desired since this was not really the previous directory that I opened. I think it might make sense to have it as it was in 4.x where the "back" arrow is deactivated by default when there has being no previous navigation. 
4. (Also not a bug from this PR cause I can reproduce this in master). When the filename of some file is too long, the "date modified" column (Which is the one activated by default), moves a lot to the right and is not visible anymore without scrolling to the right. However when navigating to some other folder where the filenames are shorter, it doesn't restore the position of the "date modified" column.

See gif for items 3 and 4:
files

dalthviz (commented):
@steff456 @ccordoba12 after a quick check of the code this LGTM :+1: the only question I have is the change in the `mainwindow` regarding opening external files with the Mac application (the change from `sys.platform == "darwin"` to `running_in_mac_app()`), is that logic not needed for any instance of Spyder in Mac?

ccordoba12:
Thanks @juanis2112 for your review. About your comments:

> The state of the options "Show hidden files" and "Single click to open files" in the hamburger menu is not sync with the ones in the preferences dialog.

I fixed this last weekend.

> After reseting spyder, by default the files opens in my user directory and when I click the "back" button it goes to the spyder directory. I'm not sure if this is desired since this was not really the previous directory that I opened. I think it might make sense to have it as it was in 4.x where the "back" arrow is deactivated by default when there has being no previous navigation.

Totally agree. Could you open am issue about that so we don't forget to fix it for Spyder 5? Thanks!

> When the filename of some file is too long, the "date modified" column (Which is the one activated by default), moves a lot to the right and is not visible anymore without scrolling to the right

This is a very ugly behavior and I also agree we should fix it. Please open an issue for that too.

ccordoba12:
> @steff456 @ccordoba12 after a quick check of the code this LGTM

Great! Thanks a lot for the review @dalthviz!

> the only question I have is the change in the mainwindow regarding opening external files with the Mac application (the change from sys.platform == "darwin" to running_in_mac_app()), is that logic not needed for any instance of Spyder in Mac?

Nop, that is only necessary for our Mac app.

ccordoba12 (approved):
Thanks @steff456 for your work!

PR #14607, opened 2021-01-17
PR: Don't terminate folding thread before running new update (Editor)

ccordoba12:
## Description of Changes

- That was freezing when typing fast in large files.
- I verified folding updates as expected without that code.

### Affirmation


ccordoba12:
This is a temporal solution until we find a better way to update folding.

PR #14626, opened 2021-01-20
[WIP] PR: Build macOS app on Big Sur (11.0)

mrclary:
## Description of Changes

Update GitHub action to build on macOS 11.0

### Affirmation

@mrclary


mrclary:
This must wait for resolution of actions/virtual-environments#2486.

.github/workflows/installer-macos.yml
```diff
@@ -9,7 +9,7 @@ name: Create macOS App Bundle and DMG
 jobs:
   build:
     name: macOS App Bundle
-    runs-on: macos-latest
+    runs-on: macos-11.0
```
ccordoba12, line 12: I think we should add a new workflow for Big Sur. Otherwise, we'll affect people still running Catalina.

mrclary: Agreed. This is just for testing right now. When this becomes available, and we can build on Apple silicon, then we can consider adding separate workflows or expand the matrix workflow; in any case we should always have a Catalina build

mrclary:
Closing since our application seems to be working on Big Sur

PR #14650, opened 2021-01-26
PR: Update metainfo file and install it

ximion:
## Description of Changes

This updates the AppStream metainfo file and uses an actual rDNS ID string for Spyders project ID (should not be changed after this one-time change!) and registers its .desktop file as `launchable`, thereby permitting software centers on Linux to directly launch Spyder.
With this change, the file also gets actually installed on Linux-based platforms, and we explicitly mark this application as being mouse/keyboard-enabled but not supporting touch input controls.

You can validate the XML metainfo file with `appstreamcli` on Linux:
`appstreamcli validate org.spyder_ide.spyder.appdata.xml`
Except for one pedantic hint about lacking a screenshot caption, the file has no warnings or errors anymore.

Thanks for considering this change! (which should also be backport-safe to the stable series, as just metadata is changed)

### Affirmation

I certify the above statement is true and correct: ximion


dalthviz:
Pinging @ccordoba12 just in case and thanks @ximion !

ximion:
I should have mentioned that I changed none of the description texts - they are copied from the previous version.
Also, shameless plug: I made a web tool to make it easier for projects to create some of the Freedesktop metadata (but most notable AppStream metaInfo files): https://www.freedesktop.org/software/appstream/metainfocreator/#/ - this will be expanded in future and maybe you'll find it useful.
(AppStream is backwards-compatible though, so the file as it is now will never become invalid)

ccordoba12:
Hey @ximion, thanks a lot for your contribution!

> Thanks for considering this change! (which should also be backport-safe to the stable series, as just metadata is changed)

We don't do backports, but we can certainly include it in our next release in the Spyder 4 series. For that, you need to derive the branch for your pull request from our `4.x` one and open a new one because you didn't create a feature branch but instead committed directly to master here.

You can do it with the following commands:

    git checkout 4.x
    git checkout update-metainfo

Then please do your changes, commit as normal, push your branch and create a new pull request. Thanks!

ximion:
> [...] because you didn't create a feature branch but instead 

Yeah, I was lazy...

The PR now also exists as https://github.com/spyder-ide/spyder/pull/14655 for `4.x`.
Thanks! :-)

ccordoba12:
No problem, and thanks for the quick response. Closing this one in favor of #14655.

PR #14756, opened 2021-02-15
PR: Use Mojave Python Build for macOS Application

### LINKED ISSUE #15506: Spyder 5.0.1 crashes on MacOS Mojave
## Problem Description
I get a "Spyder Error" when I attempt to run Spyder 5.0.1 on my Macbook Pro (MacOS Mojave 10.14.6).  I installed using the dmg installer.  I had the same issue when trying to install and run Spyder 4.2.2.

When I examine the console I see a message that may be relevant:
```Failed to copy signing info for 3430, responsible for file:///Applications/Spyder%205.0.1.app/Contents/MacOS/Spyder: #-67062: Error Domain=NSOSStatusErrorDomain Code=-67062 "(null)"```

1. Install Spyder 5.0.1 (or 4.2.2) via the MacOS DMG installer, dragging the Spyder Icon into the Applications folder.
2. Open Applications and double click on the Spyder icon.

### PR DESCRIPTION
**Pull request by mrclary**
## Description of Changes

This just tests the ability to change Python's deployment target. Will this resolve spyder-ide/website-spyder#194?

### Issue(s) Resolved

Fixes spyder-ide/website-spyder#194
Fixes #15506 

### Affirmation

@mrclary

### CODE REVIEW

File: .github/workflows/installer-macos.yml [hunk 1]
```diff
@@ -21,37 +21,42 @@ jobs:
...
     steps:
       - name: Checkout Code
         uses: actions/checkout@v2
         with:
           fetch-depth: 0
       - name: Update System
         run: |
+          python3.9 -c "import sysconfig; print(f'''MACOSX_DEPLOYMENT_TARGET: {sysconfig.get_config_var('MACOSX_DEPLOYMENT_TARGET')}''')"
+          brew uninstall --force --ignore-dependencies python
           brew install spatialindex
```
ccordoba12, line 34: You need to make sure this library is compiled for the new `MACOSX_DEPLOYMENT_TARGET` and not installed from Homebrew Catalina's bottle (that's the name for binary packages in Homebrew). Otherwise things will fail too (I think).

Unfortunately, this is necessary because the new `rtree` wheels are only compatible with Catalina.

ccordoba12: From the logs that doesn't seem to be the case.

mrclary: I don't know if you saw my latest commit...I deliberately installed Python from python.org for macOS 10.9. At the moment, we're still using the wheel-less rtree (they haven't updated on the conda default channel yet). But that is something to consider in the future. We may end up having to do multiple platform builds...
Nevertheless, I was hoping to see if this at least launched correctly in Mojave, before tackling any subsequent issues.

ccordoba12: > I don't know if you saw my latest commit...I deliberately installed Python from python.org for macOS 10.9

@mrclary, yes, I saw that you're using a build from python.org now. My comment was meant to let you know that we can't use packages from `spatialindex` in the installer if we want to make it compatible with Mojave, because Homebrew is installing a Catalina bottle. And that will (probably) make the installer crash too.

> Nevertheless, I was hoping to see if this at least launched correctly in Mojave, before tackling any subsequent issues.

Ok, I have a Mojave VM so I can help you with that,

mrclary: I've just pushed up changes to use mojave bottles for both python and spatialindex, and builds the full version. Let's see if that works...

mrclary:
@ccordoba12, with the latest commit I was able to explicitly use Mojave bottles for both Python and spatialindex. Furthermore, I've been able to: 1) successfully launch the full Spyder.app on both Catalina and Big Sur; 2) confirm that code snippets work on both platforms; 3) confirm that the internal interpreter has `sysconfig` variable `MACOSX_DEPLOYMENT_TARGET = '10.14'` on both platforms.

If you can confirm that the latest artifact works on Mojave, I'd say we're golden.

mrclary:
@ccordoba12, Tangentially, I wonder if it would be better practice to build only the full version on merge requests (for testing, etc.) and build both full and lite versions only on releases. I think that the build process can be quite lengthy (in particularly the dmg, and I don't know why); while the builds nominally run in parallel, it may still be prudent to reduce unnecessary resource load. Do we have any resource limitations with GitHub? What are your thoughts?

mrclary:
I'm getting the following error with this artifact when switching projects.
```python
File "watchdog/observers/fsevents.pyc", line 283, in run
SystemError: <built-in function add_watch> returned NULL without setting an error
```
Don't know if this was introduced by this changeset or not. Don't know if it is unique to Catalina yet.

mrclary:
@ccordoba12, I don't know if you got a chance to test this on Mojave, but it is ready for merge if you want to include this before releasing 4.2.2

ccordoba12:
> I don't know if you got a chance to test this on Mojave

I'll test it later today.

> it is ready for merge if you want to include this before releasing 4.2.2

Do you feel confident enough to include this in 4.2.2?

mrclary:
> Do you feel confident enough to include this in 4.2.2?
I've been using the artifact on Catalina for my normal work and have not found any issues. I did not do any extensive testing on Big Sur, but it appears to be fine there as well. I think it should be fine to include in 4.2.2 as far as they are concerned. However, the point was to fix the app for Mojave users, so if this did not work for them, then there is no point; if it does work for Mojave, then it would be a good inclusion for 4.2.2.

ccordoba12:
Ok, I just tested the new installer and it's failing with the following traceback:

imagen

After taking a look at PyPI, I noticed that the latest available Cython version doesn't provide wheels for Python 3.9, so the one that ended up in the installer was compiled from its tarball in 10.15, hence the failure.

I would also advice to take a look at Tcl/Tk because I guess we need to install its Mojave bottle too.

mrclary:
@ccordoba12, it actually looks like `zmq` is giving the error. Looking at the build log I see that `pyzmq` is installed from a 10.15 universal wheel, which would include the libraries built on 10.15. I changed the build to install `pyzmq` from a 10.9 wheel, so maybe that will work. For good measure, I also installed all of Python's dependencies from Mojave bottles too, including Tcl/Tk.

If this doesn't work, then I don't think there is anything we can do about it in our Pipline. Perhaps some charitable soul is willing to build it on a Mojave system whenever we have a release?

ccordoba12:
I don't have more time to test this, sorry. So I'm leaving it for Spyder 5.

mrclary:
@ccordoba12, closing this PR since we are agreed that we will not support Mojave builds in our pipeline. See [comment](https://github.com/spyder-ide/spyder/issues/15506#issuecomment-845571213)

PR #14759, opened 2021-02-15
PR: Prevent double saving when running a file

### LINKED ISSUE #14653: Possible Black Autoformat Bug
## Problem Description
When I hit F5 to run a script, Spyder will weirdly copy the last few lines of my file, and paste them multiple times at the end of the file. So my file goes from:
```
foo()
bar()
baz()
```
to
```
foo()
bar()
baz()
baz()
baz()
```
I just upgraded to 4.2. I switched autoformatter to black, and turned on autoformat-on-save. When I edit my file, do not manually run autoformat, and do not manually save, but press F5 to run the script, spyder will trigger the autoformat (fine) but at the same time it also changes the content of the end of the file (not fine). If I manually save the file, or manually run autoformatter, I do not get this problem. Also, this problem does not hit me 100% of the time. It seems most likely to happen when black encounters a long list that it splits across multiple lines.

When I run a script, I do expect it to save the file and therefore trigger black, but adding new lines to the end of the file is not expected.

andfoy:
## Description of Changes
This PR fixes some autoformatting overlaps when a file is run in the iPython console due to duplicate calls to save when autoformatting on save is enabled

### Issue(s) Resolved

Fixes #14653

### Affirmation

I certify the above statement is true and correct:  @andfoy

### DIFF
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -467,6 +467,8 @@ def test_get_help_ipython_console_dot_notation(main_window, qtbot, tmpdir):
 
     See spyder-ide/spyder#11821
     """
+    CONF.set('editor', 'save_all_before_run', False)
+
     shell = main_window.ipyconsole.get_current_shellwidget()
     control = shell._control
     qtbot.waitUntil(lambda: shell._prompt_html is not None,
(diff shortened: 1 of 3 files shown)


ccordoba12 (approved):
Thanks @andfoy for your help!

PR #14467, opened 2020-12-20
PR: Fix indentation on paste

impact27:
## Description of Changes

depends on https://github.com/jupyter/qtconsole/pull/462

Keep relative indentation between first and following lines. Automatically unindent code if python like.

* [ ] Wrote at least one-line docstrings (for any new functions)
* [ ] Added unit test(s) covering the changes (if testable)

* [ ] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

### Issue(s) Resolved

Fixes #11701

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct:

PR #14768, opened 2021-02-17
PR: Fix commenting/uncommenting to not change leading whitespaces

### LINKED ISSUE #14326: Uncommenting adds an extra whitespace in some cases
### Issue Report Checklist

    * [x] Restarted Spyder

## Problem Description

Commenting and uncommenting a line with an odd number of leading whitespace adds an extra space.

1. Type
```
x = [1,
     2,
     3]
```
2. Select all 3 lines, comment and uncomment
3. The code is now
```
x = [1,
      2,
      3]
```

The code after step 3 should be
```
x = [1,
     2,
     3]
```

If the number of leading whitespace is even, for ex:
```
xx = [1,
      2,
      3]

```
then there is no issue. 

## Versions

* Spyder version: 4.2.0
* Python version: 3.8.5
* Qt version: 5.9.7
* PyQt version: 5.9.2
* Operating System name/version: Linux 5.9.10

### Dependencies

```
# Mandatory:
chardet >=2.0.0                 :  3.0.4 (OK)
intervaltree >=3.0.2            :  3.0.2 (OK)
IPython >=4.0                   :  7.19.0 (OK)
keyring >=17.0.0                :  21.5.0 (OK)
psutil >=5.3                    :  5.7.3 (OK)
pygments >=2.0                  :  2.7.2 (OK)
pylint >=1.0                    :  2.6.0 (OK)
pyls >=0.36.1;<1.0.0            :  0.36.1 (OK)
pyls_spyder >=0.1.1             :  0.2.1 (OK)
qtawesome >=0.5.7               :  1.0.1 (OK)
qtconsole >=4.7.7               :  5.0.1 (OK)
rtree >=0.8.3                   :  0.9.4 (OK)
setuptools >=39.0.0             :  50.3.1.post20201107 (OK)
sphinx >=0.6.6                  :  3.3.1 (OK)
spyder_kernels >=1.10.0;<1.11.0 :  1.10.0 (OK)
watchdog >=0.10.3               :  0.10.4 (OK)
xdg >=0.26                      :  0.26 (OK)

# Optional:
cython >=0.21                   :  None (NOK)
matplotlib >=2.0.0              :  None (NOK)
numpy >=1.7                     :  None (NOK)
pandas >=1.1.1                  :  None (NOK)
scipy >=0.17.0                  :  None (NOK)
```

remisalmon:
## Description of Changes

This PR changes the way spyder uncomments code and fixes an issue where toggling comments on and off over lines with an uneven number of leading whitespaces would add an extra whitespace (regardless of indentation level).
Example:
```python
from pprint import pprint
x = [1,
     2]
y=[1,
   2]
def f(x, y):
    pprint(x,
           y)
```

Commenting (Ctrl+1) then uncommenting (Ctrl+1), before:
```python
from pprint import pprint
x = [1,
      2]
y=[1,
    2]
def f(x, y):
    pprint(x,
            y)
```

Commenting (Ctrl+1) then uncommenting (Ctrl+1), after:
```python
from pprint import pprint
x = [1,
     2]
y=[1,
   2]
def f(x, y):
    pprint(x,
           y)
```

This PR breaks compatibility with spyder 3: commenting with spyder 3 and uncommenting with spyder 4 with break indentation but that should not be happening much.

This PR also break toggling comments on manually commented lines (prefixing `'#'` instead of `'# '`) but that should be expected and other IDE behave the same (at least VSCode and Atom do).

### Issue(s) Resolved

Fixes #14326

### Affirmation

I certify the above statement is true and correct: remisalmon

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor/codeeditor.py
@@ -2374,43 +2374,20 @@ class CodeEditor(LSPMixin, TextEditBaseWidget):
 
     def __remove_prefix(self, prefix, cursor, line_text):
         """Handle the removal of the prefix for a single line."""
-        start_with_space = line_text.startswith(' ')
-        if start_with_space:
-            left_spaces = self.__even_number_of_spaces(line_text)
-        else:
-            left_spaces = False
-        if start_with_space:
-            right_number_spaces = self.__number_of_spaces(line_text, group=1)
-        else:
-            right_number_spaces = self.__number_of_spaces(line_text)
+        cursor.movePosition(QTextCursor.Right,
+                            QTextCursor.MoveAnchor,
+                            line_text.find(prefix))
         # Handle prefix remove for comments with spaces
         if (prefix.strip() and line_text.lstrip().startswith(prefix + ' ')
                 or line_text.startswith(prefix + ' ') and '#' in prefix):
             cursor.movePosition(QTextCursor.Right,
-                                QTextCursor.MoveAnchor,
-                                line_text.find(prefix))
-            if (right_number_spaces == 1
-                    and (left_spaces or not start_with_space)
-                    or (not start_with_space and right_number_spaces % 2 != 0)
-                    or (left_spaces and right_number_spaces % 2 != 0)):
-                # Handle inserted '# ' with the count of the number of spaces
-                # at the right and left of the prefix.
-                cursor.movePosition(QTextCursor.Right,
-                                    QTextCursor.KeepAnchor, len(prefix + ' '))
-            else:
-                # Handle manual insertion of '#'
-                cursor.movePosition(QTextCursor.Right,
-                                    QTextCursor.KeepAnchor, len(prefix))
-            cursor.removeSelectedText()
+                                QTextCursor.KeepAnchor, len(prefix + ' '))
         # Check for prefix without space
         elif (prefix.strip() and line_text.lstrip().startswith(prefix)
                 or line_text.startswith(prefix)):
             cursor.movePosition(QTextCursor.Right,
-                                QTextCursor.MoveAnchor,
-                                line_text.find(prefix))
-            cursor.movePosition(QTextCursor.Right,
                                 QTextCursor.KeepAnchor, len(prefix))
-            cursor.removeSelectedText()
+        cursor.removeSelectedText()
 
     def __even_number_of_spaces(self, line_text, group=0):
         """
--- a/spyder/plugins/editor/widgets/codeeditor/tests/test_comments.py
+++ b/spyder/plugins/editor/widgets/codeeditor/tests/test_comments.py
@@ -62,21 +62,21 @@ def test_single_line_comment(codeeditor):
     # Toggle comment with space at the right of prefix but manually inserted
     text = toggle_comment(editor, start_line=2)
     assert text == ("class a():\n"
-                    "    self.b = 1\n"
+                    "   self.b = 1\n"
                     " #   print(self.b)\n"
                     "#    \n"
                     )
     # Toggle comment with space insertion
     text = toggle_comment(editor, start_line=2)
     assert text == ("class a():\n"
-                    "    # self.b = 1\n"
+                    "   # self.b = 1\n"
                     " #   print(self.b)\n"
                     "#    \n"
                     )
     # Toggle comment deleting inserted space
     text = toggle_comment(editor, start_line=2)
     assert text == ("class a():\n"
-                    "    self.b = 1\n"
+                    "   self.b = 1\n"
                     " #   print(self.b)\n"
                     "#    \n"
                     )
@@ -84,8 +84,8 @@ def test_single_line_comment(codeeditor):
     # but manually inserted
     text = toggle_comment(editor, start_line=3)
     assert text == ("class a():\n"
-                    "    self.b = 1\n"
-                    "    print(self.b)\n"
+                    "   self.b = 1\n"
+                    "   print(self.b)\n"
                     "#    \n"
                     )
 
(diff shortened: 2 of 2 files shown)


dalthviz:
Hi @remisalmon thanks for the help! However after a quick review seems like you changed the previous test available. So the new version of the tests seems like are not covering all the cases that the old test did?

Pinging @ccordoba12 just in case

remisalmon:
@dalthviz yes I did changes the tests because the changes to `__remove_prefix()` to fix the issue I described do not pass the current tests - which are design to keep compatibility with spyder 3-style comments (where the prefix is `'#'` instead of `'# '`).

So for ex. toggling the comments off (Ctrl+1) on spyder 3-style comments like:
```python
#from pprint import pprint
#x = [1,
#     2]
#y=[1,
#   2]
#def f(x, y):
#    pprint(x,
#           y)
```

Results in:
```python
from pprint import pprint
x = [1,
    2]
y=[1,
  2]
def f(x, y):
   pprint(x,
          y)
```

Which breaks the indentation but should be expected and is the default behavior in other editors (tested with VSCode and Atom). Hope that makes sense!

dalthviz:
Thanks for the explanation @remisalmon :+1:

What do you think @ccordoba12, @spyder-ide/core-developers ?

impact27:
I think it makes sense that commenting / uncommenting should not change the file. I think the change as described here looks reasonable.

bcolsen:
While I agree that it is very unfortunate that the change in code formatting, as describes #14326, occurs, it seems unfortunate to throw out all the compatibility with manual and Spyder 3 comments that we currently have.

This change goes from breaking formatting, to breaking code (indents). The code above may now get a difficult to find indentation error with the def block now at three spaces,

It seems that Spyder is alone in caring about this however as VSCode, Atom and even Pycharm and Jupyter-lab prefer this simpler approach. It's does have an advantage from a maintainability perspective but we are losing some user friendliness by making code errors.

impact27:
Well now if I uncomment this code:
```
#def a():
# def a():
#     return
#    return
```
The `return` statement is correctly uncommented to 4 spaces, even if one line has 4 spaces and one line has 5. 
So what about: remove one space unless it breaks indentation?
An alternative would be: if uncommmenting a block, the context can be extracted from the previous line:
```
#x = [1,
#     2]
```
vs
```
# x = [1,
#      2]
```
The number of spaces before the `x` may be used? Otherwise we can use the current method.

bcolsen:
I believe currently the code will remove one space only if the number of spaces between the `#` and the code is odd. This works pretty well for 2 space and 4 space indents.

> So what about: remove one space unless it breaks indentation?

Yes that would be the goal, but how can we tell if it would break the indentation?

> An alternative would be: if uncommmenting a block, the context can be extracted from the previous line:

I like this. If a block is continuing to multiple lines remove the same number of spaces from all the line of the block as the block starting line.

The code would also have to detect function call and definition headings continuations. Currently this is a problem as well:

```python
# def ab(a=1,
#        b=2):
#     return
```

impact27:
> Yes that would be the goal, but how can we tell if it would break the indentation?

`fix_indent_smart` kind of does that. I am actually changing this function a bin in https://github.com/spyder-ide/spyder/pull/14467 . I separate this code in two parts: `find_indentation` That gives you the indentation for the current line, extracted from the previous lines, and `fix_indent_smart` that uses this information to fix the indentation. Using these, we could decide how many spaces to remove based on the indentation from the previous uncommented lines, uncommmenting one line at a time. 

> The code would also have to detect function call and definition headings continuations. Currently this is a problem as well:

Alternatively, we keep the current behaviour, unless multiple lines are deindented, one of them has only a single space after the `#`, and none of them have zero spaces.

bcolsen:
`find_indentation` is cool. It has all the things that need to happen here. Does it only find one level for the whole block? So it wouldn't work with mixed comments(space and no space...Who would do such a thing?...I would ;-) ).

I looks like it has a bracket matching element in there with `self.__get_brackets` that might be all we really need to keep the code that same and just make sure we apply the right spacing in the bracket blocks.

remisalmon:
I rebased this on 5.x, this is still breaking compatibility with Spyder 3 which is still better than breaking formatting after uncommenting commented code, in my opinion.

dalthviz:
Hi @remisalmon thank you for checking this again! I think due to this breaking compatibility with Spyder 3 (and also since we are not planning to do further Spyder 5 release for the moment) it needs to be rebased on top of the `master` branch. But what do you think @spyder-ide/core-developers ?

spyder/plugins/editor/widgets/tests/test_comments.py
```diff
@@ -62,30 +62,30 @@ def test_single_line_comment(codeeditor):
     # Toggle comment with space at the right of prefix but manually inserted
     text = toggle_comment(editor, start_line=2)
     assert text == ("class a():\n"
-                    "    self.b = 1\n"
```
ccordoba12, line 65: Why is this space not present now? I think your changes shouldn't alter this, so I don't understand this change.

The same goes for all modifications you did for all other tests here.

ccordoba12 (commented):
Hey @remisalmon, thanks for your work and sorry for not reviewing it before.

I like that your changes made our code much simpler but I'm concerned about the changes you did to our tests (see below).

ccordoba12:
To move your work to the master branch, please run the following commands:

    git checkout master
    git checkout fix_uncomments
    git rebase --onto master 5.x fix_uncomments
    git push -f origin fix_uncomments

spyder/plugins/editor/widgets/tests/test_comments.py
remisalmon: The text line 35 is defined with Spyder 3 style comments (leading `#` with no space) but the tests are expecting Spyder 4/5/6 style comments (`# ` with a space).

This PR breaks compatibility with Spyder 3 style comments which I have argued elsewhere in the comments is prefered (in my opinion) to breaking the formatting after commenting/uncommenting. In this case we could rewrite the tests to use Spyder 4/5/6 style comments.

remisalmon:
Thanks @ccordoba12 for taking a look, I rebased on the master branch and replied to your comment.

(I also tried to look at the `find_indentation`, `get_line_indentation`, `fix_indent_smart`, `fix_indent`, `simple_indentation` etc. helper functions but my head started to hurt)

ccordoba12:
> Thanks @ccordoba12 for taking a look, I rebased on the master branch and replied to your comment.

Thanks @remisalmon! I tested your changes locally and they are working as expected, so thanks a lot for your work and your patience with us.

I think the new way of toggling comments feels natural and intuitive, so I really like it. I only have one question about a sentence in your OP:

> This PR also break toggling comments on manually commented lines (prefixing '#' instead of '# ') but that should be expected and other IDE behave the same (at least VSCode and Atom do).

Could you post an example about this new behavior to understand better what you mean?

remisalmon:
> > Thanks @ccordoba12 for taking a look, I rebased on the master branch and replied to your comment.
> 
> Thanks @remisalmon! I tested your changes locally and they are working as expected, so thanks a lot for your work and your patience with us.

:tada: 

> 
> I think the new way of toggling comments feels natural and intuitive, so I really like it. I only have one question about a sentence in your OP:
> 
> > This PR also break toggling comments on manually commented lines (prefixing '#' instead of '# ') but that should be expected and other IDE behave the same (at least VSCode and Atom do).
> 
> Could you post an example about this new behavior to understand better what you mean?

Take this code and uncomment the commented lines (first two use the old Spyder 3 comment style, last two use the new Spyder 4/5/6 comment syle):
```python
if True:
#    pass
    #pass
#     pass
    # pass
```

Spyder + this PR:
image

VSCode:
image

Atom/Pulsar:
image

Spyder currently:
image

ccordoba12:
Ok, thanks for the clarification. Since this makes Spyder behave as other IDEs in that particular case, I agree with you that it's the right thing to do.

ccordoba12:
> (I also tried to look at the find_indentation, get_line_indentation, fix_indent_smart, fix_indent, simple_indentation etc. helper functions but my head started to hurt)

Missed this one, lol! Yeah, that code is not easy to follow but we'd really appreciate your help to improve it later if you have the energy for it.

PR #14853, opened 2021-03-03
WIP: PR: Include user's local PYTHONPATH from .bash_profile

### LINKED ISSUE #14843: Ability to use PYTHONPATH from user bash_profile in macOS application

mrclary:
## Description of Changes

Retrieved PYTHONPATH from user's local .bash_profile at startup of the mainwindow and when PYTHONPATH Manager is launched.

### Issue(s) Resolved

Fixes #14843

### Affirmation

@mrclary

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -97,7 +97,7 @@ from spyder.utils import encoding, programs
 from spyder.utils import icon_manager as ima
 from spyder.utils.misc import (select_port, getcwd_or_home,
                                get_python_executable)
-from spyder.utils.programs import is_module_installed
+from spyder.utils.programs import is_module_installed, add_user_pythonpath
 from spyder.utils.qthelpers import (create_action, add_actions, get_icon,
                                     create_program_action, DialogManager,
                                     create_python_script_action, file_uri,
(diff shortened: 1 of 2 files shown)


mrclary:
@steff456, okay, here is the basic working functionality.
At the mainwindow creation (at application launch) when the python path is loaded from `SPYDER_PATH`, the user's local `PYTHONPATH` is also merged in. Similarly when the PYTHONPATH Manager is opened.

However, there are few things we need to consider.
* Are there considerations for Linux and Windows platforms? I've left a placeholder in the `spyder.utils.programs.add_user_pythonpath` function.
* There should be a preference that allows the user to toggle the functionality. I think this should be in the PYTHONPATH Manager, but that will require changing the user interface and adding another Spyder preference.
* How should this behave with regard to adding/deleting paths? For example, if the local `PYTHONPATH` is merged in, and the user later removes paths, local `PYTHONPATH` paths will be added back upon restarting the application or launching the PYTHONPATH Manager. We may not want this behavior.

Another option might be to have a simple button in the PYTHONPATH Manager that merges in the user's local `PYTHONPATH`, and not have it merged in automatically. But this would mean they have to do that any time their `.bash_profile` changed. Or maybe this option in conjunction with second bullet above?

What do you think?
Ping @ccordoba12

ccordoba12:
Hey @mrclary, please move your PR to master because we're not going to have more releases for Spyder 4.

mrclary:
@ccordoba12, I've rebased onto master and changed the PR target to master.
Should I rebase all outstanding PRs onto latest master? Will there be a 5.x branch? Please advise.

ccordoba12:
> @ccordoba12, I've rebased onto master and changed the PR target to master.

Thanks!

> Should I rebase all outstanding PRs onto latest master?

Yes, please do.

> Will there be a 5.x branch?

Yes, it will be. I'll create it after I release the first Spyder 5 version. For now, all PRs that we merge directly to master will be part of Spyder 5.0.0.

PR #14918, opened 2021-03-12
PR: Add options to include user's local PYTHONPATH and environment variables

### LINKED ISSUE #14843: Ability to use PYTHONPATH from user bash_profile in macOS application

mrclary:
## Description of Changes

Added preference to the Python Interpreter section allowing all platforms and launch methods to select whether a user's system `PYTHONPATH` and environment variables are used for Consoles.

Spyder's PYTHONPATH Manager list behavior is unchanged. 

Fixes #14843

Screen Shot 2021-05-31 at 4 02 16 PM

### New Behavior
* If "Use system PYTHONPATH" is deselected: paths in PYTHONPATH Manager are appended to the Console's `sys.path`
* If "Use system PYTHONPATH" is selected: in addition to the above, the paths in the user's local `PYTHONPATH` environment variable are appended to the Console's `sys.path`. On macOS and Linux, the user's local `PYTHONPATH` is determined by `/etc/profile`, `~/.bash_profile`, `~/.bash_login`, and `~/.profile`. On Windows, the user's local `PYTHONPATH` is extracted from the registry query `reg query HKEY_CURRENT_USER\Environment`.
* If "Use system environment variables" is deselected: only the following variables are passed to the Console's `os.environ`.
  * `PATH` and `HOME` on macOS
  * `PATH`, `HOME`, `DISPLAY` and `XAUTHORITY` on Linux
  * 'PATH', 'SYSTEMROOT', 'SYSTEMDRIVE', and 'USERPROFILE' on Windows
* If "Use system environment variables" is selected, all of the user's local environment variables, except `PYTHONPATH`, are passed to the Console's `os.environ`. On macOS and Linux, these are determined by `/etc/profile`, `~/.bash_profile`, `~/.bash_login`, and `~/.profile`. On Windows these are determined by the registry query `reg query HKEY_CURRENT_USER\Environment`

### Affirmation

@mrclary

### DIFF
--- a/spyder/plugins/maininterpreter/confpage.py
+++ b/spyder/plugins/maininterpreter/confpage.py
@@ -125,10 +125,9 @@ class MainInterpreterConfigPage(PluginConfigPage):
 
         # ENV Group
         pyenv_group = QGroupBox(_("Environment variables"))
-        pyenv_label = QLabel(_("For standalone Spyder applications "
-                               "(macOS, Windows), select whether to add user "
-                               "environment variables to the Python "
-                               "interpreter environment."))
+        pyenv_label = QLabel(_("Select whether to include system environment "
+                               "variables to the Python interpreter "
+                               "environment."))
         pyenv_label.setWordWrap(True)
         system_pythonpath = newcb(
             _("Use system PYTHONPATH"),
@@ -146,10 +145,6 @@ class MainInterpreterConfigPage(PluginConfigPage):
             tip=_("For standalone applications, use your "
                   "system environment variables")
         )
-        if not running_in_mac_app() and not is_pynsist():
-            # Disable preference if not a standalone application
-            system_pythonpath.setEnabled(False)
-            system_env_variables.setEnabled(False)
 
         pyenv_layout = QVBoxLayout()
         pyenv_layout.addWidget(pyenv_label)
--- a/spyder/plugins/maininterpreter/confpage.py
+++ b/spyder/plugins/maininterpreter/confpage.py
@@ -23,6 +23,7 @@ from spyder.utils import programs
 from spyder.utils.conda import get_list_conda_envs_cache
 from spyder.utils.misc import get_python_executable
 from spyder.utils.pyenv import get_list_pyenv_envs_cache
+from spyder.config.base import is_pynsist, running_in_mac_app
 
 # Localization
 _ = get_translation('spyder')
@@ -123,6 +124,7 @@ class MainInterpreterConfigPage(PluginConfigPage):
             'system_pythonpath',
             msg_info=_("Please note that these changes will "
                        "be applied only to new consoles"),
+            tip=_("For standalone applications, use your system PYTHONPATH")
         )
         pyexec_layout.addWidget(system_pythonpath)
         system_env_variables = newcb(
@@ -130,8 +132,14 @@ class MainInterpreterConfigPage(PluginConfigPage):
             'system_env_variables',
             msg_info=_("Please note that these changes will "
                        "be applied only to new consoles"),
+            tip=_("For standalone applications, use your "
+                  "system environment variables")
         )
         pyexec_layout.addWidget(system_env_variables)
+        if not running_in_mac_app() and not is_pynsist():
+            # Disable preference if not a standalone application
+            system_pythonpath.setEnabled(False)
+            system_env_variables.setEnabled(False)
 
         pyexec_group.setLayout(pyexec_layout)
 
--- a/spyder/plugins/maininterpreter/confpage.py
+++ b/spyder/plugins/maininterpreter/confpage.py
@@ -119,6 +119,17 @@ class MainInterpreterConfigPage(PluginConfigPage):
         self.cus_exec_radio.toggled.connect(self.cus_exec_combo.setEnabled)
         pyexec_layout.addWidget(self.cus_exec_combo)
 
+        pyexec_group.setLayout(pyexec_layout)
+
+        self.pyexec_edit = self.cus_exec_combo.combobox.lineEdit()
+
+        # ENV Group
+        pyenv_group = QGroupBox(_("Environment variables"))
+        pyenv_label = QLabel(_("For standalone Spyder applications "
+                               "(macOS, Windows), select whether to add user "
+                               "environment variables to the Python "
+                               "interpreter environment."))
+        pyenv_label.setWordWrap(True)
         system_pythonpath = newcb(
             _("Use system PYTHONPATH"),
             'system_pythonpath',
@@ -127,7 +138,6 @@ class MainInterpreterConfigPage(PluginConfigPage):
                        "and immediately to completions"),
             tip=_("For standalone applications, use your system PYTHONPATH")
         )
-        pyexec_layout.addWidget(system_pythonpath)
         system_env_variables = newcb(
             _("Use system environment variables"),
             'system_env_variables',
@@ -136,15 +146,16 @@ class MainInterpreterConfigPage(PluginConfigPage):
             tip=_("For standalone applications, use your "
                   "system environment variables")
         )
-        pyexec_layout.addWidget(system_env_variables)
         if not running_in_mac_app() and not is_pynsist():
             # Disable preference if not a standalone application
             system_pythonpath.setEnabled(False)
             system_env_variables.setEnabled(False)
 
-        pyexec_group.setLayout(pyexec_layout)
-
-        self.pyexec_edit = self.cus_exec_combo.combobox.lineEdit()
+        pyenv_layout = QVBoxLayout()
+        pyenv_layout.addWidget(pyenv_label)
+        pyenv_layout.addWidget(system_pythonpath)
+        pyenv_layout.addWidget(system_env_variables)
+        pyenv_group.setLayout(pyenv_layout)
 
         # UMR Group
         umr_group = QGroupBox(_("User Module Reloader (UMR)"))
@@ -193,6 +204,7 @@ class MainInterpreterConfigPage(PluginConfigPage):
 
         vlayout = QVBoxLayout()
         vlayout.addWidget(pyexec_group)
+        vlayout.addWidget(pyenv_group)
         vlayout.addWidget(umr_group)
         vlayout.addStretch(1)
         self.setLayout(vlayout)
--- a/spyder/plugins/maininterpreter/confpage.py
+++ b/spyder/plugins/maininterpreter/confpage.py
@@ -117,6 +117,22 @@ class MainInterpreterConfigPage(PluginConfigPage):
         self.def_exec_radio.toggled.connect(self.cus_exec_combo.setDisabled)
         self.cus_exec_radio.toggled.connect(self.cus_exec_combo.setEnabled)
         pyexec_layout.addWidget(self.cus_exec_combo)
+
+        system_pythonpath = newcb(
+            _("Use system PYTHONPATH"),
+            'system_pythonpath',
+            msg_info=_("Please note that these changes will "
+                       "be applied only to new consoles"),
+        )
+        pyexec_layout.addWidget(system_pythonpath)
+        system_env_variables = newcb(
+            _("Use system environment variables"),
+            'system_env_variables',
+            msg_info=_("Please note that these changes will "
+                       "be applied only to new consoles"),
+        )
+        pyexec_layout.addWidget(system_env_variables)
+
         pyexec_group.setLayout(pyexec_layout)
 
         self.pyexec_edit = self.cus_exec_combo.combobox.lineEdit()
--- a/spyder/plugins/maininterpreter/confpage.py
+++ b/spyder/plugins/maininterpreter/confpage.py
@@ -123,7 +123,8 @@ class MainInterpreterConfigPage(PluginConfigPage):
             _("Use system PYTHONPATH"),
             'system_pythonpath',
             msg_info=_("Please note that these changes will "
-                       "be applied only to new consoles"),
+                       "be applied only to new consoles but "
+                       "and immediately to completions"),
             tip=_("For standalone applications, use your system PYTHONPATH")
         )
         pyexec_layout.addWidget(system_pythonpath)
--- a/spyder/utils/programs.py
+++ b/spyder/utils/programs.py
@@ -40,6 +40,13 @@ from spyder.utils.misc import get_python_executable
 
 HERE = osp.abspath(osp.dirname(__file__))
 
+# Minimum environment variables required for subprocesses
+REQ_ENV_VARS = []
+if os.name == 'nt':
+    REQ_ENV_VARS.extend(['PATH', 'SYSTEMROOT', 'SYSTEMDRIVE', 'USERPROFILE'])
+else:
+    REQ_ENV_VARS.extend(['HOME'])
+
 
 class ProgramError(Exception):
     pass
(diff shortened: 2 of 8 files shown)


mrclary:
This PR and #14853 are mutually exclusive.

spyder/utils/programs.py
```diff
@@ -1062,3 +1062,24 @@ def find_git():
...
+
+def get_user_env_variables():
+    """Return local environment variables"""
+    cmdstr = ''
+    if sys.platform == 'darwin':
+        cmdstr = ('[[ -e /etc/profile ]] && source /etc/profile; '
+                  '[[ -e ~/.bash_profile ]] && source ~/.bash_profile; '
+                  'printenv')
+    elif sys.platform.startswith('linux'):
+        pass
```
ccordoba12, line 1075: Linux should work very similar to Mac, right? Could we make this case work too?

spyder/utils/programs.py
```diff
@@ -1062,3 +1062,24 @@ def find_git():
...
+    elif os.name == 'nt':
+        pass
```
ccordoba12, line 1077: I think it's necessary to make this work too and @dalthviz doesn't have time to help you with that. See for instance:

https://stackoverflow.com/questions/5327495/list-all-environment-variables-from-the-command-line

spyder/plugins/maininterpreter/confpage.py
```diff
@@ -117,6 +118,30 @@ def setup_page(self):
         self.def_exec_radio.toggled.connect(self.cus_exec_combo.setDisabled)
         self.cus_exec_radio.toggled.connect(self.cus_exec_combo.setEnabled)
         pyexec_layout.addWidget(self.cus_exec_combo)
+
+        system_pythonpath = newcb(
```
ccordoba12, line 122: Please add these new options under a new QGroupBox called `Environment variables`.

spyder/plugins/maininterpreter/confpage.py
```diff
@@ -117,6 +118,30 @@ def setup_page(self):
...
+        system_env_variables = newcb(
+            _("Use system environment variables"),
+            'system_env_variables',
+            msg_info=_("Please note that these changes will "
+                       "be applied only to new consoles"),
+            tip=_("For standalone applications, use your "
+                  "system environment variables")
+        )
+        pyexec_layout.addWidget(system_env_variables)
+        if not running_in_mac_app() and not is_pynsist():
```
ccordoba12, line 140: I think it'd be nice to have this on Linux because the usage of env vars is quite common there.

ccordoba12 (commented):
Thanks @mrclary for your work on this! I left an initial review for you.

spyder/utils/programs.py
mrclary: Absolutely! That is why I left a placeholder. Can you provide the profile files appropriate for linux?

spyder/utils/programs.py
mrclary: Okay, I'll take a look.

spyder/plugins/maininterpreter/confpage.py
mrclary: What purpose will the widget serve when launching Spyder from a non-app installation? When launching from commandline, the user environment variables will already be included, yes? If you want to keep the addition to the preferences for non-app installations, then deselecting `system_env_variables` in preferences would require _removing_ user environment variables which would be very tricky.

mrclary: In fact, it looks like even the Windows standalone application already includes the user's environment variables. So this is only relevant to the macOS app.

ccordoba12: > What purpose will the widget serve when launching Spyder from a non-app installation?

You can launch Spyder graphically from the Anaconda Navigator or the equivalent to the Windows menu. So, in those cases it's very useful to pass your env vars to the kernel.

> So this is only relevant to the macOS app.

Great! Then, I suggest making this work for Unix systems (because Spyder can also be launched graphically on Mac when not using our installer).

mrclary: Okay, I'm confused.
* Launching Spyder either from the commandline or graphically (Anaconda Navigator) on all platforms will already inherit the user's environment variables, including PYTHONPATH.
* Additionally, launching Spyder's standalone application on Windows platform will already inherit the user's environment variables, including PYTHONPATH.
* Only launching Spyder's standalone application on macOS platform will NOT inherit the user's environment variables, nor the user's system PYTHONPATH.

I'm not sure what you mean by "making this work for Unix systems".

In light of the above, we have the following possibilities:
1. Eliminate the preference altogether and always use the user's environment variables (including PYTHONPATH) when using the standalone application on macOS.
2. Provide the preference ONLY for the standalone application on macOS.
3. Provide the preference for all launching scenarios.

Options 1 and 2 are easy. Option 3 can be done but involves _removing_ environment variables when the option is deselected; this will be true for unix systems as well as Windows. Did you mean to support option 3?

mrclary:
I can't for the life of me figure out how to fix this problem...
I've gotten all tests to pass except for slow tests on Linux.
I've got the environment setup on my Ubuntu VM as close to identical to the Github test environment as possible, compared all the conda package versions, and cannot reproduce the error locally. The tests pass locally.
<details><summary>Failed test_run_cython_code</summary>

```python
=================================== FAILURES ===================================
_____________________________ test_run_cython_code _____________________________

main_window = <spyder.app.mainwindow.MainWindow object at 0x7f5ddcb26f50>
qtbot = <pytestqt.qtbot.QtBot object at 0x7f5ddc049c50>

    @pytest.mark.slow
    @flaky(max_runs=3)
    @pytest.mark.skipif(
        (os.name == 'nt' or sys.platform == 'darwin' or
         parse_version(ipy_release.version) == parse_version('7.11.0')),
        reason="Hard to test on Windows and macOS and fails for IPython 7.11.0")
    def test_run_cython_code(main_window, qtbot):
        """Test all the different ways we have to run Cython code"""
        # ---- Setup ----
        # Get a reference to the code editor widget
        code_editor = main_window.editor.get_focus_widget()
    
        # ---- Run pyx file ----
        # Load test file
        main_window.editor.load(osp.join(LOCATION, 'pyx_script.pyx'))
    
        # Run file
        qtbot.keyClick(code_editor, Qt.Key_F5)
    
        # Get a reference to the namespace browser widget
        nsb = main_window.variableexplorer.current_widget()
    
        # Wait until an object appears
        qtbot.waitUntil(lambda: nsb.editor.source_model.rowCount() == 1,
>                       timeout=COMPILE_AND_EVAL_TIMEOUT)
E       AssertionError: waitUntil timed out in 30000 miliseconds
E       assert not True
E        +  where True = <function QtBot.waitUntil.<locals>.timed_out at 0x7f5dd3f1e4d0>()

/home/runner/work/spyder/spyder/spyder/app/tests/test_mainwindow.py:1104: AssertionError
----------------------------- Captured Qt messages -----------------------------
    QtWarningMsg: Scenegraph already initialized, setBackend() request ignored
--------------------------- Captured stdout teardown ---------------------------

In [1]: runfile('/home/runner/work/spyder/spyder/spyder/app/tests/pyx_script.pyx', wdir='/home/runner/work/spyder/spyder/spyder/app/tests')
gcc: fatal error: cannot execute ‘cc1’: execvp: No such file or directory
compilation terminated.

In [2]: 
---------------------------- Captured stderr setup -----------------------------
<class 'spyder.app.tests.spyder_boilerplate.spyder.plugin.SpyderBoilerplate'>: A plugin with section "spyder_boilerplate" already exists!
  File "/home/runner/work/spyder/spyder/spyder/app/mainwindow.py", line 917, in setup
    self.register_plugin(plugin_instance, external=True)
  File "/home/runner/work/spyder/spyder/spyder/app/mainwindow.py", line 237, in register_plugin
    plugin._register()
  File "/home/runner/work/spyder/spyder/spyder/api/plugins.py", line 884, in _register
    self._conf.register_plugin(self)
  File "/home/runner/work/spyder/spyder/spyder/config/manager.py", line 119, in register_plugin
    'exists!'.format(conf_section))
RuntimeError: A plugin with section "spyder_boilerplate" already exists!
```
</details>

`test_move_to_first_breakpoint` for pip slow linux has the same stderr setup error: `RuntimeError: A plugin with section "spyder_boilerplate" already exists!`

I just can't get these tests to fail locally, so I can't figure out what is going on.
@spyder-ide/core-developers , can you provide any insight or recommendations?

dalthviz:
@mrclary I think this is related with the fake plugin I added and the reuse of the mainwindow instance in the tests. I will check :+1: (probably we need to add some sort of condition/validation so the addition of the fake plugin entrypoint is only done once)

mrclary:
@dalthviz, thanks for looking into the fake plugin, perhaps that is the issue. However, I am curious why that doesn't manifest in the mac and windows tests also.

mrclary:
@dalthviz, I was able to fix the issues with a combination of two changes.
* I made `register_fake_entrypoints` its own fixture with teardown. I think this is better, and seems to have moved things along some.
  * `test_run_cython_code` no longer has `RuntimeError: A plugin with section "spyder_boilerplate" already exists!`.
  * `test_move_to_first_breakpoint` now succeeds on pip, but pip slow now has the same `test_run_cython_code` error.

```python
In [1]: runfile('/home/runner/work/spyder/spyder/spyder/app/tests/pyx_script.pyx', wdir='/home/runner/work/spyder/spyder/spyder/app/tests')
gcc: fatal error: cannot execute ‘cc1’: execvp: No such file or directory
compilation terminated.
```
* Now, ensuring 'PATH' is in the environment variables solves the remaining issues.

mrclary:
@ccordoba12 @dalthviz,
I removed my edits to the test fixtures for `register_fake_entrypoints`. My changes had resulted in success for Linux, but caused Windows and Mac to fail; I'm not sure why that was the case.
Nevertheless, #15749 seems to have fixed everything all around.

I think this is ready for review. Please note the updated description of the PR. It should explain all the new behavior.

mrclary:
@dalthviz, I've rebased and the errors have returned.

<details><summary>Linux pip fast: test_closing_plugin</summary>

```python
==================================== ERRORS ====================================
____________________ ERROR at setup of test_closing_plugin _____________________

qtbot = <pytestqt.qtbot.QtBot object at 0x7fd730bf7f10>

    @pytest.fixture
    def findinfiles(qtbot):
        """Set up SearchInComboBox combobox."""
        findinfiles_plugin = FindInFiles(None, configuration=CONF)
    
        # qtbot wants to close the widget
        findinfiles_plugin.close = lambda: True
>       qtbot.addWidget(findinfiles_plugin)

/home/runner/work/spyder/spyder/spyder/plugins/findinfiles/tests/test_plugin.py:35: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pytestqt.qtbot.QtBot object at 0x7fd730bf7f10>
widget = <spyder.plugins.findinfiles.plugin.FindInFiles object at 0x7fd73069b230>

    def addWidget(self, widget, *, before_close_func=None):
        """
        Adds a widget to be tracked by this bot. This is not required, but will ensure that the
        widget gets closed by the end of the test, so it is highly recommended.
    
        :param QWidget widget:
            Widget to keep track of.
    
        :kwparam before_close_func:
            A function that receives the widget as single parameter, which is called just before
            the ``.close()`` method gets called.
    
        .. note:: This method is also available as ``add_widget`` (pep-8 alias)
        """
        if not isinstance(widget, qt_api.QtWidgets.QWidget):
>           raise TypeError("Need to pass a QWidget to addWidget!")
E           TypeError: Need to pass a QWidget to addWidget!
```
</details>
<details><summary>Linux pip slow: test_programmatic_custom_layouts</summary>

```python
=================================== FAILURES ===================================
_______________________ test_programmatic_custom_layouts _______________________

main_window = <spyder.app.mainwindow.MainWindow object at 0x7f04e07a3730>
qtbot = <pytestqt.qtbot.QtBot object at 0x7f04f2cac450>

    @pytest.mark.slow
    @flaky(max_runs=3)
    def test_programmatic_custom_layouts(main_window, qtbot):
        """
        Test that a custom layout gets registered and it is recognized."""
        mw = main_window
        mw.first_spyder_run = False
    
        # Test layout registration
        layout_id = 'testing layout'
        # Test the testing plugin is being loaded
>       mw.get_plugin('spyder_boilerplate')

/home/runner/work/spyder/spyder/spyder/app/tests/test_mainwindow.py:2309: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <spyder.app.mainwindow.MainWindow object at 0x7f04e07a3730>
plugin_name = 'spyder_boilerplate', error = True

    def get_plugin(self, plugin_name, error=True):
        """
        Return a plugin instance by providing the plugin class.
        """
        for name, plugin in self._PLUGINS.items():
            if plugin_name == name:
                return plugin
        else:
            if error:
                raise SpyderAPIError(
>                   'Plugin "{}" not found!'.format(plugin_name))
E               spyder.api.exceptions.SpyderAPIError: Plugin "spyder_boilerplate" not found!
```
</details>

dalthviz:
@mrclary will check :+1:  seems like other PRs are also affected in the linux tests

mrclary:
> @mrclary will check 👍 seems like other PRs are also affected in the linux tests

@dalthviz , would it better to make `register_fake_entrypoints` a function-scoped fixture using the code in `remove_fake_entrypoints` as the teardown? It looks to me like the `main_window` fixture invokes creating the fake entrypoints on a function scope, but the removing them does not occur until `cleanup`, which will only be called once since it is session scoped. Wouldn't this potentially cause issues?

dalthviz:
Maybe you are right @mrclary but I'm not totally sure, @andfoy what do you think?

dalthviz:
@mrclary checking seems like `pytest-qt` did a release yesterday and seems like that is making the tests on linux with pip fail. Will do a PR to fix the `pytest-qt` version to be < than `4.0.0`

mrclary:
> @mrclary checking seems like `pytest-qt` did a release yesterday and seems like that is making the tests on linux with pip fail. Will do a PR to fix the `pytest-qt` version to be < than `4.0.0`

Do you think that is a bug in `pytest-qt` or an updated feature that we'll need to eventually accommodate?

dalthviz:
Probably changes they added to support a new version of `pytest`. When we move to a new mayor release of `pytest` we will need to check and probably do a proper migration of the test suite

dalthviz:
@mrclary you can rebase now to get the change done at #15782 for the tests

mrclary:
@ccordoba12, another idea to consider here...
The current PR adds the contents of the user's system PYTHONPATH to sys.path in the Console and extra_paths for Jedi completion. I wonder if it would be better to simply append (uniquely) the user's PYTHONPATH to the contents of PYTHONPATH Manager. This would expose the user's system PYTHONPATH to the user and provide additional flexibility to activate/deactivate certain paths. When the preference is enabled, the paths would be added (and enabled). When the preference is disabled, paths would not be removed but no longer actively added to the PYTHONPATH Manager; any existing paths would remain unless explicitly removed by the user.

What do you think?

mrclary:
From conversation in #16200, remember to consider how `sys.path` is applied when using "Execute in an external systerm terminal" run option.

mrclary:
Closing in favor of #16429

Issue #33, opened 2015-02-16
Ctrl+F9, question mark in comments


spyder-issues-migrator:
_From [y.bonne...@gmail.com](https://code.google.com/u/111370336049651795761/) on 2009-09-12T04:52:08Z_

The first script works, the second doesn't

if 1:
    \# Comment
    pass
    if 1:
        pass

if 1:
    \# Comment ?
    pass
    if 1:
        pass

Seems the help is called for the question mark in the comments.

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=33_

spyder-issues-migrator:
_From [pierre.raybaut](https://code.google.com/u/pierre.raybaut/) on 2009-09-12T08:11:32Z_

fixed in v1.0.0beta7

**Status:** Fixed

Issue #13, opened 2015-02-16
ArrayEditor: add support for record arrays


spyder-issues-migrator:
_From [pierre.raybaut](https://code.google.com/u/pierre.raybaut/) on 2009-08-29T07:16:26Z_

xxx

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=13_

spyder-issues-migrator:
_From [pierre.raybaut](https://code.google.com/u/pierre.raybaut/) on 2009-08-30T04:14:52Z_

implemented in v1.0.0beta4

**Status:** Fixed

PR #14983, opened 2021-03-23
 Change all icons to Material Design Icons

isabela-pf:
## Description of Changes

---
- [x] Replace all non-custom Material Design icons and update with appropriate color constants
- [ ] Replace all custom Material Design-style icons where they are called (outside icon_manager.py)
- [x] Remove unused icons
- [ ] Update splash screen to say `Spyder 5`

Based on discussions to improve the UI with Spyder 5, we decided to move to a single icon set. All icons will be replaced with [Material Design Icons](https://pictogrammers.github.io/@mdi/font/5.4.55/) or custom icons based on that set. Custom icons are still being decided and you can follow the discussion at [ux-improvements #33](https://github.com/spyder-ide/ux-improvements/issues/33).

### Issue(s) Resolved
Fixes [ux-improvements #33](https://github.com/spyder-ide/ux-improvements/issues/33)

### Affirmation

isabela-pf


isabela-pf:
Here are some notes I took while working on this. I don't think all these questions need to be addressed before Spyder 5; I can always make a separate issue for the ones that we don't think need immediate attention. All of these relate to `icon_manager.py`.

Urgent notes

1. I changed the way icon colors are defined to match the constant names as mentioned in [ux-improvements #13](https://github.com/spyder-ide/ux-improvements/issues/13). I replaced the definition of `MAIN_FG_COLOR` (the old constant) in line 29, but I don't know if that is all that needs to be done and if I can do the same for the other icon constants.
2. Are `scale_factor` properties anything I need to worry about? They are one of the few I kept just in case.
3. We have a mix of icons referenced for `debug`. The actual `debug` icon is changed to `mdi.step-forward` here as it is in the chart, but we also approved the `debug_cell` icon (not yet added to this PR) with the icon that looks like a blue combined play and pause. Here is `mdi.step-forward` for reference. Screen Shot 2021-03-22 at 5 38 37 PM
4. What do we want to replace the default `gray` with? It looks like it's often used for inactive states, but we don't currently have an equivalent for it in the icon constants. Since the icon manager is already use opacity changes in a few other places for inactive states, we could continue that in these cases. For light mode though, they wouldn't look gray with lowered opacity dark icons on a light background.
5. Since we are changing the `maximize pane` icons from two icons for each state to one icon for a button that now has a pressed state, I might need some help making sure that works. I will do my best to make that change without help, but I wanted to give you a heads up in case.
6. `MessageBoxWarning` has no mdi replacement icon listed. I changed it from `fa.warning` to `mdi.alert` since we made that same replacement for other icons. If there's a different choice that should be made, let me know.
7. Some icons have the same form in different colors with the description `transparent` and seem to be called for inactive versions of buttons. Others have the `color_disabled` property. Most of our icons don't have any kind of inactive state described in the icon manager. Is one of these the style we want them all to be? I'm bringing this up so that I know what changes I need to make.

Non-urgent notes (These can definitely wait until after Spyder 5)

8. We have two icons for environments used in two different places (see line 112 `environment`and line 235 `environ`. Is there a reason we want two different icons?
9. With this replacement we are still using many different arrow icons. I don't see a real reason they can't match. 
10. The icon for `DialogHelpButton` is the red life buoy. I don't think it needs to be red since it's not an error.

ccordoba12:
@isabela-pf, do you want us to respond to your comment here? Or do you prefer to copy it to your new PR, so we can answer you there?

isabela-pf:
I was going to copy it to the new PR. Sorry, I went to eat dinner after our last meeting.

ccordoba12:
> I was going to copy it to the new PR. Sorry, I went to eat dinner after our last meeting.

No worries, I'll post my answers in your new then.

PR #15012, opened 2021-03-24
PR: Fix Kite issues for Spyder 5

### LINKED ISSUE #14888: Kite support migration issues
- [x] Enabling/disabling action not affecting the status bar (if you had already Kite installed seems like)

After unchecking the enable kite config (and after restarting Spyder) I still see the Kite logo and base text in the status bar:
image

Enabling/disabling Kite:

kite

_Originally posted by @dalthviz in https://github.com/spyder-ide/spyder/issues/14314#issuecomment-791629790_

- [x] Seems like the status bar is always showing the `Kite: Initializing message` and checking the copilot Kite says that was unable to detect Spyder installed (if you had already Kite installed seems like):
image

Also, probably the screenshot available in the Kite help page should be updated to add the new preferences layout in Spyder 5

- [x] After installation (uninstalling Kite and resetting Spyder preferences first) I didn't see the on bording message appearing
- [x] I was unable to trigger the call-to-action after checking the option in the preferences (I had to restart Spyder first)
- [x] Clicking the `Learn more` button in the call-to-action ends up throwing an exception:
```
  File "C:\Users\Daniel\Documents\Spyder\spyder\spyder\plugins\completion\providers\kite\widgets\calltoaction.py", line 135, in _learn_more
    kite = self.parent().completions.get_client('kite')
AttributeError: 'CompletionPlugin' object has no attribute 'get_client'
```

Just in case, I did all of this testing on Windows and to trigger the call-to-action I used this code:

```python
fig = plt.figure()
fig.add
```

dalthviz:
Fixes #14888 

~It's missing the item regarding the on boarding file (that seems like is being caused since the `sig_file_open` gets emitted before being actually connected to the Editor)~

### DIFF
--- a/spyder/plugins/completion/providers/kite/provider.py
+++ b/spyder/plugins/completion/providers/kite/provider.py
@@ -155,6 +155,12 @@ class KiteProvider(SpyderCompletionProvider):
         """Request status for the given file."""
         self.client.sig_perform_status_request.emit(filename)
 
+    @on_conf_change(
+        section='completions', option=('enabled_providers', 'kite'))
+    def on_kite_enable_changed(self, value):
+        self.sig_call_statusbar.emit(
+            KiteStatusWidget.ID, 'set_value', (None,), {})
+
     @on_conf_change(section='completions', option='enable_code_snippets')
     def on_code_snippets_changed(self, value):
         if running_under_pytest():
--- a/spyder/plugins/completion/providers/kite/provider.py
+++ b/spyder/plugins/completion/providers/kite/provider.py
@@ -201,7 +201,6 @@ class KiteProvider(SpyderCompletionProvider):
             # retry
             self._show_onboarding = True
             return
-        print(onboarding_file)
         self.sig_open_file.emit(onboarding_file)
         self.set_conf('show_onboarding', False)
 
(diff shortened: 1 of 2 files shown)


spyder/plugins/completion/providers/kite/provider.py
```diff
@@ -195,8 +201,9 @@ def _show_onboarding_file(self, onboarding_file):
             # retry
             self._show_onboarding = True
             return
-        self.set_conf('show_onboarding', False)
+        print(onboarding_file)
```
ccordoba12, line 204: Please remove this print.

spyder/plugins/completion/providers/kite/widgets/calltoaction.py
```diff
@@ -91,7 +88,7 @@ def handle_mouse_press(self, event):
         self.hide()
 
     def handle_processed_completions(self, completions):
-        if not self._enabled:
+        if not CONF.get('completions', 'kite_call_to_action'):
```
ccordoba12, line 91: Please make this class inherit from `SpyderConfigurationAccessor` and use `self.get_conf/self.set_conf` instead. That's the new way to access config options instead of calling `CONF` directly.

ccordoba12 (commented):
Thanks @dalthviz for your work on this!

spyder/plugins/completion/providers/kite/widgets/calltoaction.py
ccordoba12: You also need to set `CONF_SECTION = 'completions'` as a class constant for that to work.

ccordoba12 (approved):
Thanks a lot @dalthviz! Great work here!

PR #15100, opened 2021-04-03
PR: Fix error when detecting spyder-kernels version

### LINKED ISSUE #15091: spyder-kernels problem with Spyder 5
on MACOS (macos big sur 11.2.3) previous versions worked smoothly but after updating to spyder 5.0.0  spyder‑kernels not working and cannot be installed using ( conda install spyder‑kernels OR pip install spyder‑kernels )

#ERROR Message :
You have missing dependencies!

Mandatory:
spyder_kernels >=2.0.1,<2.1.0 : 2.0.1 (NOK)

Please install them to avoid this message.

Note: Spyder could work without some of these dependencies, however to have a smooth experience when using Spyder we strongly recommend you to install all the listed missing dependencies.

Failing to install these dependencies might result in bugs. Please be sure that any found bugs are not the direct result of missing dependencies, prior to reporting a new issue.

ccordoba12:
## Description of Changes

This fixes the detection of spyder-kernels at startup in our dependencies mechanism.

### Issue(s) Resolved

Fixes #15091.

### Affirmation


Kashish1000 (commented):
change SPYDER_KERNELS_REQVER = '>=2.0.1;<2.1.0' to 
SPYDER_KERNELS_REQVER = '>=2.0.1'  
This will fix this error

real-yfprojects:
> change SPYDER_KERNELS_REQVER = '>=2.0.1;<2.1.0' to
> SPYDER_KERNELS_REQVER = '>=2.0.1'
> This will fix this error

The `<2.1.0` is probably there, so that changes in new versions which aren't backwards compatible are noticed and told the user.
But if they comply to Semantic Versioning the minor version number should not increase in case of backwards incompatible changes. Instead the major _(first)_ version number should increase. The line should changed to
```python
 SPYDER_KERNELS_REQVER = '>=2.0.1;<3.0.0'
```

ccordoba12:
@Kashish1000, that's not the right solution.

@real-yfprojects, we know what we're doing.

Issue #15074, opened 2021-04-01
Screen flickering the first time I open Spyder (MacOS)


juanis2112:
After doing reset, the first time I open Spyder the application flickers. It seems like it loads the layout twice. (It might be an OS problem)

![screen](https://user-images.githubusercontent.com/18587879/113231078-0c8cc580-9260-11eb-8142-7f9b30890a41.gif)

dalthviz:
Hi @juanis2112 this happens to you systematically after a restart? Maybe is related with the application of the default layout after the mainwindow is visible which resizes the mainwindow and then in MacOS causes the 'flickering' (I'm guessing you mean the resize when the mainwindow is loading but I'm unsure)

PR #15131, opened 2021-04-06
PR: Fix a crash at startup when switching screens

ccordoba12:
## Description of Changes

- Fix crash at startup when Spyder is closed in a secondary screen, that screen is unplugged and then Spyder is opened in the primary screen.
- Add `get_conf_default` method to `SpyderConfigurationAccessor` to make that fix easier.
- Start to version our API to let developers know about additions or removals.

### Affirmation

### DIFF
--- a/.github/scripts/modules_test.sh
+++ b/.github/scripts/modules_test.sh
@@ -47,7 +47,10 @@ for f in spyder/*/*.py; do
         continue
     fi
     if [[ $f == spyder/widgets/about.py ]]; then
-    continue
+        continue
+    fi
+    if [[ $f == spyder/api/__init__.py ]]; then
+        continue
     fi
     python "$f"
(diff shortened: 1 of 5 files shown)


ccordoba12:
@dalthviz, I think the idea of the methods that were moved from the main window to Layouts was to leave them exactly as they were working before, right?

If that's the case, then I found and fixed some things that were changed or removed during the migration. One of them was causing the switching screens bug I mentioned above.

ccordoba12:
@juanis2112, I think this PR could also fix issue #15074. Please test it and let us know if it works for you.

dalthviz (approved):
LGTM :+1: @ccordoba12

ccordoba12:
According to @juanis2112, this doesn't fix issue #15074.

PR #15288, opened 2021-04-14
PR: Programmatic addition of new layouts and layouts config update

dalthviz:
## Description of Changes

Define how other plugins can declare custom layouts and add them into the layouts available

#### TODO

- [x] Change the handling of the config to support custom programmatic layouts, custom layouts snapshots (custom layouts through hexstate) order and activation
- [x] Toggle next/previous layout
- [x] Dialog actions to update config (move, delete, activate, etc)
- [x] Provide a way for plugins to declare the custom layout classes (could be a constant like `LAYOUTS` which is a list of custom layouts classes to be added)
- [x] Tests, code style issues

### Affirmation


dalthviz:
A preview: 

layouts

spyder/api/widgets/menus.py
```diff
@@ -62,6 +62,20 @@ def __init__(self, parent=None, title=None, dynamic=True):
...
         self.aboutToShow.connect(self._render)
 
+    def clear_actions(self):
+        """
+        Remove actions from the menu (including custom references)
+
+        Returns
+        -------
+        None.
+
```
ccordoba12, line 72:
```suggestion
```

Removes this empty line because it's not necessary.

spyder/app/mainwindow.py
```diff
@@ -1221,6 +1221,15 @@ def pre_visible_setup(self):
                 except TypeError:
                     pass
 
+        # Register custom layouts
+        for plugin, plugin_instance in self._PLUGINS.items():
+            if hasattr(plugin_instance, 'CUSTOM_LAYOUTS'):
+                if isinstance(plugin_instance.CUSTOM_LAYOUTS, list):
```
ccordoba12, line 1227: If `CUSTOM_LAYOUTS` is not a a list, perhaps we should print a warning saying that CUSTOM_LAYOUTS was detected for plugin `foo` but it can't be loaded because it's not a list.

spyder/plugins/layout/container.py
```diff
@@ -196,11 +193,11 @@ def trigger(i=index, self=self):
...
-                        name,
-                        text=name,
-                        triggered=trigger(),
-                        register_shortcut=False,
-                    )
+                            name,
+                            text=name,
+                            triggered=trigger(),
+                            register_shortcut=False,
+                        )
```
ccordoba12, line 200:
```suggestion
                        name,
                        text=name,
                        triggered=trigger(),
                        register_shortcut=False,
                    )
```

The extra indentation doesn't seem necessary anymore.

spyder/plugins/layout/container.py
```diff
@@ -271,13 +281,24 @@ def get_layout(self, layout_id):
...
 
         if dlg.exec_():
             name = dlg.combo_box.currentText()
+            if name in self._spyder_layouts:
+                QMessageBox.critical(
+                    self,
+                    _("Error"),
+                    _("Layout <b>{0}</b> was defined programatically. "
+                      "It is not possible to overwrite programatically "
+                      "registered layouts").format(name)
```
ccordoba12, line 299:
```suggestion
                      "registered layouts.").format(name)
```

spyder/plugins/layout/plugin.py
```diff
@@ -130,6 +132,9 @@ def register(self):
         self._update_fullscreen_action()
 
     def before_mainwindow_visible(self):
+        # Update layout menu
+        self.get_container().update_layout_menu_actions()
```
ccordoba12, line 136:
```suggestion
        self.update_layout_menu_actions()
```

Give that you added this method below, I think we should use it here instead.

ccordoba12 (commented):
Thanks @dalthviz for this! Great work here!

I only left some small suggestions for you.

ccordoba12 (commented):
@dalthviz, I left some small suggestions for you, otherwise it's ready.

ccordoba12:
Hi @dalthviz, one thing I forgot to mention is that, given that your PR added new things to the API, we need to change its version.

For that, please bump this number

https://github.com/spyder-ide/spyder/blob/b84f2221b253567dfc1f011dee0eda89ae3c36d6/spyder/api/_version.py#L9

to `(0, 2, 0)`.

ccordoba12 (approved):
Thanks a lot @dalthviz for all the time and effort you put into this. It's a terrific contribution!

PR #15366, opened 2021-04-20
PR: Minor layout improvements to the cornet widget of the panes toolbar

jnsebgosselin:
## Description of Changes

In this PR, I propose that we make `MainCornerWidget` inherit from `QToolBar` instead of `QWidget` so that we can set more easily a style that is consistent with that of the `MainWidgetToolbar` .

I also suggest to add a strut to the corner widget so that there is a space between the last item of the `MainWidgetToolbar` and the first item of the corner widget.

## Before
- there is no space to the left of the spinner
- the space to the right of the spinner is too large (though it is very subtle)

spinner_spacing_nok

## After
spinner_spacing_ok

### Affirmation

### DIFF
--- a/spyder/api/widgets/auxiliary_widgets.py
+++ b/spyder/api/widgets/auxiliary_widgets.py
@@ -9,8 +9,8 @@ Spyder API auxiliary widgets.
 """
 
 # Third party imports
-from qtpy.QtCore import Signal
-from qtpy.QtWidgets import QHBoxLayout, QMainWindow, QWidget
+from qtpy.QtCore import Signal, QSize
+from qtpy.QtWidgets import QMainWindow, QSizePolicy, QToolBar, QWidget
 
 # Local imports
 from spyder.api.exceptions import SpyderAPIError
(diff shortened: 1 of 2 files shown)


ccordoba12:
Thanks a lot for this @jnsebgosselin! There's one detail I'd like to talk about: the shape of the are that contains the spinner is not a square (like for all other widgets in pane toolbars)

imagen

However, I think we could leave it like that because the spinner is not a clickable button.

@isabela-pf, what do you think?

jnsebgosselin:
> Thanks a lot for this @jnsebgosselin! There's one detail I'd like to talk about: the shape of the area that contains the spinner is not a square (like for all other widgets in pane toolbars):

You are right. For consistency, it is probably better if the spinner has the same shape as the other buttons.

Then I have concerns that if we threat all toolbar items like that, we are going to waste even more screen real-estate, but that is another topic entirely.

We will see what @isabela-pf think and I will come up with a solution to set properly the size of the spinner if needed.

isabela-pf:
I've been summoned! I agree in an ideal world I would like to have it be the same shape as all other elements in the menu, but I also think we can afford to save the little bit of space here since it is not an interactive area where we need to be considering clickable area. I'm happy with the rectangular space in this case. Thank you for asking!

ccordoba12 (approved):
Thanks @jnsebgosselin!

PR #15372, opened 2021-04-21
PR: Profile script

impact27:
## Description of Changes
Add `profile_file` and `profile_cell` functions to the shell with the same signature as `runcell`/`debugcell` and `runfile`/`debugfile`

This solves a few issues I have with profiling, including:
 - Profiling when the user want to use the current namespace
 - Profiling when the user want to stop the profiled script in the middle
 - Profiling a Cell

### Issue(s) Resolved

Fixes #

### Affirmation

### DIFF
--- a/external-deps/spyder-kernels/.gitrepo
+++ b/external-deps/spyder-kernels/.gitrepo
@@ -5,8 +5,8 @@
 ;
 [subrepo]
 	remote = https://github.com/spyder-ide/spyder-kernels.git
-	branch = 2.x
-	commit = 126c46c79223b62d675099c595594a54b8a0cac2
-	parent = 5244ef9021757ed74e500d3f2f1c02b3b8386922
+	branch = profile_script
+	commit = 10ec83f4ee5d9d220bc4f9a1c4cdb4f6a7a3a236
+	parent = a06748e20eafb735a329ce51f3d2815bc4286673
 	method = merge
-	cmdver = 0.4.3
+	cmdver = 0.4.1
--- a/external-deps/spyder-kernels/spyder_kernels/customize/spydercustomize.py
+++ b/external-deps/spyder-kernels/spyder_kernels/customize/spydercustomize.py
@@ -13,10 +13,12 @@
 
 import bdb
 import cmd
+import cProfile
 import io
 import logging
 import os
 import pdb
+import tempfile
 import shlex
 import sys
 import time
@@ -27,7 +29,7 @@ from IPython.core.getipython import get_ipython
 
 from spyder_kernels.comms.frontendcomm import CommError, frontend_request
 from spyder_kernels.customize.namespace_manager import NamespaceManager
-from spyder_kernels.customize.spyderpdb import SpyderPdb, enter_debugger
+from spyder_kernels.customize.spyderpdb import SpyderPdb, get_new_debugger
 from spyder_kernels.customize.umr import UserModuleReloader
 from spyder_kernels.py3compat import TimeoutError, PY2, _print, encode
 
@@ -410,7 +412,8 @@ def transform_cell(code, indent_only=False):
     return '\n' * number_empty_lines + code
 
 
-def exec_code(code, filename, ns_globals, ns_locals=None, post_mortem=False):
+def exec_code(code, filename, ns_globals, ns_locals=None, post_mortem=False,
+              profile_filename=None, debugger=None):
     """Execute code and display any exception."""
     # Tell IPython to hide this frame (>7.16)
     __tracebackhide__ = True
@@ -450,7 +453,23 @@ def exec_code(code, filename, ns_globals, ns_locals=None, post_mortem=False):
                         SHOW_INVALID_SYNTAX_MSG = False
         else:
             compiled = compile(transform_cell(code), filename, 'exec')
-        exec(compiled, ns_globals, ns_locals)
+
+        if debugger:
+            debugger.run(compiled, 
+                globals=ns_globals,
+                locals=ns_locals
+            )
+        elif profile_filename:
+            # Run with profiler
+            cProfile.runctx(
+                compiled, 
+                globals=ns_globals,
+                locals=ns_locals,
+                filename=profile_filename
+            )
+        else:
+            exec(compiled, ns_globals, ns_locals)
+
     except SystemExit as status:
         # ignore exit(0)
         if status.code:
@@ -484,7 +503,7 @@ def get_file_code(filename, save_all=True):
 
 
 def runfile(filename=None, args=None, wdir=None, namespace=None,
-            post_mortem=False, current_namespace=False):
+            post_mortem=False, current_namespace=False, **kwargs):
     """
     Run filename
     args: command line arguments (string)
@@ -563,7 +582,8 @@ def runfile(filename=None, args=None, wdir=None, namespace=None,
                 ipython_shell.run_cell_magic('cython', '', f.read())
         else:
             exec_code(file_code, filename, ns_globals, ns_locals,
-                      post_mortem=post_mortem)
+                      post_mortem=post_mortem, 
+                      **kwargs)
 
         sys.argv = ['']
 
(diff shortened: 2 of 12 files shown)


ccordoba12:
Hey @impact27, please post an animated gif/video of the new behavior.

impact27:
plot

List of improvements:
 - Access to local variables in the console
 - Can compare several runs by profiling in different consoles
 - Can profile in the same environment as the code is run
 - Can get a result even if the run is interrupted (For example if the code reaches an infinite loop)

mrclary:
@impact27, so this replaces the existing "Run profiler" menu item with "Profile file" and "Profile cell", correct? And the heavy lifting is moved to `spyder-kernels` so that the profiler is always run in an existing IPython Console environment (inheriting its `os.environ`, etc.), correct?

If the above is correct, then I agree that this PR should resolve #18010 and obviate PR #18031.

impact27:
> @impact27, so this replaces the existing "Run profiler" menu item with "Profile file" and "Profile cell", correct? And the heavy lifting is moved to `spyder-kernels` so that the profiler is always run in an existing IPython Console environment (inheriting its `os.environ`, etc.), correct?
> 
> If the above is correct, then I agree that this PR should resolve #18010 and obviate PR #18031.

Yes exactly, the profiling would happen in the spyder-kernel, so there would be no difference with runfile and debugfile

mrclary:
@impact27 and @ccordoba12, when will this PR be ready for review? Should this be included in 5.3.2 milestone?

impact27:
This is ready for review, assuming the tests pass when they are fixed

andfoy:
If the IPythonConsole will profile the files, we would prefer to use the viewer run API instead.

impact27:
The toolobar looks like:
Dark:
Screenshot 2022-08-13 at 11 30 27
Light:
Screenshot 2022-08-13 at 11 31 02

ccordoba12:
@impact27, I think you need to rewrite this work on top of the new Run architecture defined in PR #17467, which I just merged.

impact27:
> @impact27, I think you need to rewrite this work on top of the new Run architecture defined in PR #17467, which I just merged.
> 
> Sorry for the inconvenience, but that's something I finally have time to review and merge until now

I am not sure what I should do, what I do here is basically the same as in the debug plugin, but the changes there include the following warning:

```
# TODO: This is a temporary measure to debug selections whilst the
# debug API is defined.
```

Should I replicate this here?

impact27:
I used the new Run architecture but I find it a bit confusing. If I want to switch from regular execute to profiling I have to go to the setting every time, and I need to use the green arrow to launch the profiling. I guess this would be the same for debugging

impact27:
@ccordoba12 I will adapt the debugger as well

impact27:
@ccordoba12 the profiler and debugger have been ported to the run plug-in. Furthermore I added a toolbar in both plugins so you can run the file/cell/selection from there. There also is a toolbar in the main toolbar and it can be disabled.

impact27:
@ccordoba12 I moved the "Run code" section of the editor plugin because it doesn't make sense there

impact27:
@ccordoba12 I split the debugger and editor cleanup in https://github.com/spyder-ide/spyder/pull/20557

impact27:
@ccordoba12 I moved the runfile to magics code to https://github.com/spyder-ide/spyder/pull/20633 so you can take a look at this one independently

dalthviz:
@ccordoba12 @impact27 I was checking and, if understand correctly, this PR changes quite a bit the way the Profiler pane looks and works, right? Should the changes here supersede the ones defined at https://github.com/spyder-ide/ux-improvements/issues/20 ? or how should we align the changes here with the ones proposed there in some other way?

impact27:
@dalthviz There are several things in spyder-ide/ux-improvements#20. I would say:

 - I am doing console integration rather than editor integration, because you could want to profile cells or selections or snippets and these do not match neatly to a file.
 - I am not against a history function but for now this PR improves the save / load features so part of the functionality is there

In general I think there are some ideas we can keep from spyder-ide/ux-improvements#20. But these can be included in future PRs?

dalthviz:
Thanks for the response @impact27 ! Seems like we should rediscuss https://github.com/spyder-ide/ux-improvements/issues/20 reevaluating how those old ideas apply not only for the interaction with the Editor but also with the integration you are doing here with the IPython console and in general the new functionality to not only profile files but cells/code selections, right?

> In general I think there are some ideas we can keep from https://github.com/spyder-ide/ux-improvements/issues/20. But these can be included in future PRs?

Definitely more things can be done in follow up PRs! I just wanted to have a more clear idea on how the profiler is expected to work and also bring up some UX/UI ideas that were discussed in the past about the Profiler plugin.

> Do you have a specific modification you want to propose to this PR?

I don't have any specific modification but I will try to check this more thoroughly to give some feedback :+1:

impact27:
Hi @dalthviz , I am trying to add the actions in the run menu, but I can't find a way to add them after the run actions. There is no "after_section" and the profiler is called before the run module.

impact27:
> Hi @dalthviz , I am trying to add the actions in the run menu, but I can't find a way to add them after the run actions. There is no "after_section" and the profiler is called before the run module.

@dalthviz I worked around this issue by checking that the main menu is setup before adding secondary run buttons, as you can see in my last commit

dalthviz:
>> Hi @dalthviz , I am trying to add the actions in the run menu, but I can't find a way to add them after the run actions. There is no "after_section" and the profiler is called before the run module.
> 
> @dalthviz I worked around this issue by checking that the main menu is setup before adding secondary run buttons, as you can see in my last commit

Hi @impact27 ! I think another option is to use the `RunMenuSections.RunInExecutors` section. I think the order for the Run menu section is `RunMenuSections.Run` (main run actions, i.e run file and config), then comes `RunMenuSections.RunExtras` (extra run options like run cell/selection) and then the final section which is `RunMenuSections.RunInExecutors` (I think this last section was though for the actions of executors like in this case the profiler).

Your approach seems fine too so not sure what could be the best 🤔 What do you think with this new info?

> Well this is not exactly the same as you have a list of plot and are browsing them vs the previous / next of a web browser. What do you think?

I was checking and seems like the same previous/next icon for Plots is being used in the Online Help pane:

image

Not sure if there are other panes where back/forward - previous/next icons are set. I think it would be nice to have the same icon for all panes for similar actions for consistency around the UI but its true that maybe we could have one set for more page like navigation (as a browser, like here and in the Online Help pane) and other set for more previous/next like actions (like the Plots pane) so actually not sure.

Pinging @ccordoba12 @conradolandia and @CAM-Gerlach maybe they could have some ideas regarding the icons

CAM-Gerlach:
Hmm, not sure but the Plots pane button match much more closely the typical icons for forward/back in a web browser and other contexts, vs. the buttons icons used in the Profiler pane.

Given the plots are actually listed vertical rather than horizontally, maybe vertical arrows would make more sense there instead?

impact27:
The name of the icons are undo / redo, which is what they are used for. I feel like left / right arrows would be more suited for browsing different profiling results (which this plugin doesn't do)

CAM-Gerlach:
Oh sorry, I guess I'm confused—I thought the buttons were for browsing previous profiling runs on the file, but I guess not? Could you explain what action/change the buttons undo and redo?

impact27:
So the idea is that you can change the way the profiling results are shown:
 - by default, only the top-level elements are shown
 - You can focus on elements with large local time (Which might be called from different code path)
 - You can switch the direction of the tree from callers to callees
 - You can focus on a specific function

When you change the view, some elements appear / disappear. The undo / redo button allow you to go back to the previous view

CAM-Gerlach:
Ah, gotcha, thanks. In that case, it seems to me that the undo/redo buttons make more sense, then, given I'd assume more typical "Back" and "Forward" buttons would be navigating the history of different profiles instead.

ccordoba12:
@impact27, could you fix the merge conflicts in this PR so I can start reviewing it? Thanks!

external-deps/spyder-kernels/.gitrepo
```diff
@@ -5,8 +5,8 @@
 ;
 [subrepo]
 	remote = https://github.com/spyder-ide/spyder-kernels.git
-	branch = remove_locals_inspection
-	commit = adb31866c64079bc023c4626f1e2cce3c51ab5f6
-	parent = d4fa5fd95e2ebe89cf49453ecd13cda9c72f73c2
+	branch = profile_script
```
ccordoba12, line 8: The branch in Spyder-kernels which has the new profile magics is called `improve_namespace`. So, please sync that branch here to test this PR manually.

ccordoba12 (commented):
Thanks @impact27 for this great improvement!

ccordoba12:
@impact27, please update this PR to see if we can merge it next.

impact27:
@ccordoba12 I updated this PR to merge on 6.x but there are no tests running?

ccordoba12:
@impact27, now 6.x and master are independent branches, so you can't perform merges between them (that's why your history now shows all backport PRs we created against 6.x, which can't be merged back to master because the code for them is already there).

Do you want me to give you hand to fix your branch?

impact27:
> Do you want me to give you a hand to fix your branch?
Yes that would be great thanks

ccordoba12:
Closing in favor of #24794, in which I applied this PR's patch to the latest master.

PR #15447, opened 2021-04-27
PR: Fix args/kwargs handling in the call_all_editorstacks signature to update completion related options for the Editor

### LINKED ISSUE #15324: TypeError when using kite_tutorial

follow instructions in kite_tutorial.py

just type a in line 84 to try autocompletion fwhen  importing matplotlib

boom

### Traceback
```python-traceback
  File "/home/explore/miniconda3/envs/minecraft/lib/python3.6/site-packages/spyder/plugins/editor/widgets/codeeditor.py", line 4490, in _handle_completions
    if (len(text) >= self.automatic_completions_after_chars
TypeError: '>=' not supported between instances of 'int' and 'tuple'
```

## Versions

* Spyder version: 5.0.0 
* Python version: 3.6.13
* Qt version: 5.9.6
* PyQt5 version: 5.9.2
* Operating System: Linux 5.4.0-72-generic

### Dependencies

```

# Mandatory:
cookiecutter >=1.6.0          :  1.7.2 (OK)
IPython >=7.6.0               :  7.16.1 (OK)
jedi =0.17.2                  :  0.17.2 (OK)
keyring >=17.0.0              :  22.3.0 (OK)
nbconvert >=4.0               :  6.0.7 (OK)
parso =0.7.0                  :  0.7.0 (OK)
pygments >=2.0                :  2.8.1 (OK)
pylint >=1.0                  :  2.7.4 (OK)
pyls >=0.36.2;<1.0.0          :  0.36.2 (OK)
pyls_black >=0.4.6            :  0.4.6 (OK)
pyls_spyder >=0.3.2           :  0.3.2 (OK)
qstylizer >=0.1.10            :  0.1.10 (OK)
qtawesome >=0.5.7             :  1.0.2 (OK)
qtconsole >=5.0.3             :  5.0.3 (OK)
qtpy >=1.5.0                  :  1.9.0 (OK)
rtree >=0.8.3                 :  0.9.7 (OK)
setuptools >=39.0.0           :  52.0.0.post20210125 (OK)
sphinx >=0.6.6                :  3.5.3 (OK)
spyder_kernels >=2.0.1;<2.1.0 :  2.0.1 (OK)
watchdog >=0.10.3;<2.0.0      :  1.0.2 (OK)
xdg >=0.26                    :  0.27 (OK)
zmq >=17                      :  20.0.0 (OK)

# Optional:
cython >=0.21                 :  None (NOK)
matplotlib >=2.0.0            :  None (NOK)
numpy >=1.7                   :  None (NOK)
pandas >=1.1.1                :  None (NOK)
scipy >=0.17.0                :  None (NOK)
sympy >=0.7.3                 :  None (NOK)
```

dalthviz:
## Description of Changes

Config updates in the completion manager for the providers were being wrapped inside a tuple

### Issue(s) Resolved

Fixes #15324

### Affirmation

### DIFF
--- a/spyder/plugins/completion/plugin.py
+++ b/spyder/plugins/completion/plugin.py
@@ -367,7 +367,7 @@ class CompletionPlugin(SpyderPluginV2):
             if opt in options:
                 opt_value = self.get_conf(opt, section=sec)
                 self.sig_editor_rpc.emit('call_all_editorstacks',
-                                         (method_name, opt_value,),
+                                         (method_name, (opt_value,),),
                                          {})
 
         # Update entries in the source menu
--- a/spyder/plugins/completion/plugin.py
+++ b/spyder/plugins/completion/plugin.py
@@ -367,7 +367,7 @@ class CompletionPlugin(SpyderPluginV2):
             if opt in options:
                 opt_value = self.get_conf(opt, section=sec)
                 self.sig_editor_rpc.emit('call_all_editorstacks',
-                                         (method_name, (opt_value,)),
+                                         (method_name, opt_value,),
                                          {})
 
         # Update entries in the source menu
--- a/spyder/plugins/editor/plugin.py
+++ b/spyder/plugins/editor/plugin.py
@@ -1603,7 +1603,7 @@ class Editor(SpyderPluginWidget):
             if str(id(editorstack)) != editorstack_id_str:
                 editorstack.rename_in_data(original_filename, filename)
 
-    def call_all_editorstacks(self, method, *args, **kwargs):
+    def call_all_editorstacks(self, method, args, **kwargs):
         """Call a method with arguments on all editorstacks."""
         for editorstack in self.editorstacks:
             method = getattr(editorstack, method)


spyder/plugins/completion/plugin.py
```diff
@@ -367,7 +367,7 @@ def after_configuration_update(self, options: List[Union[tuple, str]]):
             if opt in options:
                 opt_value = self.get_conf(opt, section=sec)
                 self.sig_editor_rpc.emit('call_all_editorstacks',
-                                         (method_name, (opt_value,)),
+                                         (method_name, opt_value,),
```
ccordoba12, line 370: @dalthviz, did you test that this change works for all other `editor_method_sec_opts` items?

ccordoba12: I mean, I say it because perhaps it breaks applying other options listed there.

dalthviz: @ccordoba12 no, I was hoping @andfoy could tell me if some setter there needs the tuple but then I will check :+1:

andfoy: This should be a tuple, due to the signature of this method: https://github.com/spyder-ide/spyder/blob/5c8cca33fe2cc012efdffe2c57dcf8a1143f3386/spyder/plugins/editor/plugin.py#L408

andfoy: Since the signal is only used to set options, maybe we could simplify it to consider a single argument?

dalthviz: From what I see and tested seems to work. Actually, seems like other options were using the truthy value of the tuple (so even if you uncheck them they didn't disable)

dalthviz: Just in case, checking a little bit more, seems like the tuple could be for the `edge_line_columns` however the panel handles non-tuple values too. The type of the values for the setters options are:

* `enable_code_snippets`: bool
* `enable_hover_hints`: bool
* `format_on_save`: bool
* `automatic_completions`: bool
* `completions_hint`: bool
* `completions_hint_after_ms`: int 
* `underline_errors`: bool
* `automatic_completions_after_chars`: int
* `automatic_completions_after_ms`: int
* `pycodestyle/max_line_length`: tuple
* `edge_line`: bool

dalthviz: @andfoy I didn't see your message above but I think I now know were the actual issue is :+1:

dalthviz: I think there is something strange passing the `args` and `kwargs` currently that is causing that they arrive to the setter methods still as a tuple (instead of being expanded when calling the corresponding setter)

dalthviz:
@andfoy could you check if the new change I added makes sense? Is working for me but just to be sure, thanks!

ccordoba12 (approved):
Thanks @dalthviz! I tried with all `editor_method_sec_opts` we can set through preferences and everything is working as expected.

Issue #15356, opened 2021-04-20
KeyError when restarting kernel on Windows


Patrick-Cole:
### What steps will reproduce the problem?

When some code is in debug (I use breakpoint()), the error occurs when closing the console using the 'x'. If I stop the debug mode or press the black stop button first, it does not occur.

### Traceback
```python-traceback
--- Logging error ---
Traceback (most recent call last):
  File "C:\Python\WPy64-3880\python-3.8.8.amd64\lib\site-packages\qtconsole\comms.py", line 113, in get_comm
    return self.comms[comm_id]
KeyError: '9ba3286fa1a511eba4f5fcaa14552b99'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Python\WPy64-3880\python-3.8.8.amd64\lib\logging\__init__.py", line 1084, in emit
    stream.write(msg + self.terminator)
AttributeError: 'NoneType' object has no attribute 'write'
Call stack:
  File "C:\Python\WPy64-3880\python-3.8.8.amd64\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Python\WPy64-3880\python-3.8.8.amd64\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Python\WPy64-3880\python-3.8.8.amd64\scripts\spyder.exe\__main__.py", line 7, in <module>
    sys.exit(main())
  File "C:\Python\WPy64-3880\python-3.8.8.amd64\lib\site-packages\spyder\app\start.py", line 213, in main
    mainwindow.main(options, args)
  File "C:\Python\WPy64-3880\python-3.8.8.amd64\lib\site-packages\spyder\app\mainwindow.py", line 2311, in main
    mainwindow = create_window(app, splash, options, args)
  File "C:\Python\WPy64-3880\python-3.8.8.amd64\lib\site-packages\spyder\app\mainwindow.py", line 2207, in create_window
    app.exec_()
  File "C:\Python\WPy64-3880\python-3.8.8.amd64\lib\site-packages\qtconsole\comms.py", line 48, in _dispatch
    getattr(self, msg_type)(msg)
  File "C:\Python\WPy64-3880\python-3.8.8.amd64\lib\site-packages\qtconsole\comms.py", line 172, in comm_msg
    comm = self.get_comm(comm_id)
[...]

PR #15462, opened 2021-04-29
PR: Fix close comm (IPython console)

impact27:
## Description of Changes
Close comm properly when the shutdown message is sent to the comm channel

### Issue(s) Resolved

Fixes https://github.com/spyder-ide/spyder/issues/15356
Removes "no such comm" messages
Depend on https://github.com/spyder-ide/spyder-kernels/pull/287

### Affirmation


ccordoba12:
Thanks a lot for your help with this @impact27!

> Depend on spyder-ide/spyder-kernels#287

Please sync our subrepo with that PR then. That's necessary to check that our tests here don't fail due to that PR in spyder-kernels.

> Actually there is still the possibility of sending a message before getting the shutwdown message

Is it possible to handle that case in the future?

impact27:
> Is it possible to handle that case in the future?

The problem is basically https://en.wikipedia.org/wiki/Two_Generals%27_Problem

But we could send a shutdown to the kernel, and expect an acknowledgment before removing the comm. I will think about that.

impact27:
the failure doesn't appear to be related to this PR?

ccordoba12:
> the failure doesn't appear to be related to this PR?

Right, I fixed it in PR #15483. I merged with `5.x` to see if our tests pass. After that, I'll test this on Windows.

ccordoba12 (approved):
Thanks a lot for your help with this bug @impact27!

I tested it on Windows and verified it fixes issue #15356.

PR #15523, opened 2021-05-06
PR: Fix loading complex third party plugins

steff456:
## Description of Changes

This changes enable Spyder to load a third party plugin that has it's own configuration. As well, some minor bugs with appearance configuration broadcasting are fixed. 

### Affirmation

### DIFF
--- a/spyder/api/config/mixins.py
+++ b/spyder/api/config/mixins.py
@@ -72,6 +72,35 @@ class SpyderConfigurationAccessor:
 
         return CONF.get(section, option, default)
 
+    def get_conf_options(self, section: Optional[str] = None):
+        """
+        Get all options from the given section.
+
+        Parameters
+        ----------
+        section: Optional[str]
+            Section in the configuration system, e.g. `shortcuts`. If None,
+            then the value of `CONF_SECTION` is used.
+
+        Returns
+        -------
+        values: BasicTypes
+            Values of the option in the configuration section.
+
+        Raises
+        ------
+        spyder.py3compat.configparser.NoOptionError
+            If the option does not exist in the configuration under the given
+            section and the default value is NoDefault.
+        """
+        section = self.CONF_SECTION if section is None else section
+        if section is None:
+            raise AttributeError(
+                'A SpyderConfigurationAccessor must define a `CONF_SECTION` '
+                'class attribute!'
+            )
+        return CONF.options(section)
+
     def set_conf(self,
                  option: ConfigurationKey,
                  value: BasicTypes,
--- a/spyder/api/config/mixins.py
+++ b/spyder/api/config/mixins.py
@@ -60,8 +60,7 @@ class SpyderConfigurationAccessor:
         Raises
         ------
         spyder.py3compat.configparser.NoOptionError
-            If the option does not exist in the configuration under the given
-            section and the default value is NoDefault.
+            If the section does not exist in the configuration.
         """
         section = self.CONF_SECTION if section is None else section
         if section is None:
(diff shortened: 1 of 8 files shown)


spyder/api/config/mixins.py
```diff
@@ -72,6 +72,35 @@ def get_conf(self,
...
+        Returns
+        -------
+        values: BasicTypes
+            Values of the option in the configuration section.
+
+        Raises
+        ------
+        spyder.py3compat.configparser.NoOptionError
+            If the option does not exist in the configuration under the given
+            section and the default value is NoDefault.
```
ccordoba12, line 94:
```suggestion
            If the section does not exist in the configuration.
```

spyder/api/config/mixins.py
```diff
@@ -150,6 +179,31 @@ def get_conf_default(self,
...
+        Returns
+        -------
+        shortcut: str
+            Key sequence of the shortcut.
+
+        Raises
+        ------
+        spyder.py3compat.configparser.NoOptionError
+            If the option does not exist in the configuration under the given
+            section and the default value is NoDefault.
```
ccordoba12, line 202:
```suggestion
            If the shortcut does not exist in the configuration.
```

spyder/app/mainwindow.py
```diff
@@ -1203,6 +1206,13 @@ def pre_visible_setup(self):
         The actions here are related with setting up the main window.
         """
         logger.info("Setting up window...")
+        # Create external plugins before loading the layout to include them in
+        # the restore state of the window after the restart.
```
ccordoba12, line 1210:
```suggestion
        # the window restore state after restarts.
```

spyder/config/manager.py
```diff
@@ -605,8 +605,9 @@ def config_shortcut(self, action, context, name, parent):
     def iter_shortcuts(self):
         """Iterate over keyboard shortcuts."""
         for context_name, keystr in self._user_config.items('shortcuts'):
-            context, name = context_name.split('/', 1)
-            yield context, name, keystr
+            if 'additional_configuration' not in context_name:
```
ccordoba12, line 608: Is this something that it's required now for all third-party plugins? I mean, the `additional_configuration` part in `context_name`?

spyder/plugins/appearance/confpage.py
```diff
@@ -26,7 +26,7 @@
 
 
 class AppearanceConfigPage(PluginConfigPage):
-    APPLY_CONF_PAGE_SETTINGS = True
+    APPLY_CONF_PAGE_SETTINGS = False
```
ccordoba12, line 29: The default is `False`, so you can remove this line instead.

ccordoba12 (dismissed):
Thanks a lot for this @steff456!! I left some small comments for you.

spyder/config/manager.py
steff456: It is because the additional conf options add that string to the list, and I didn't found a way to avoid this so the safest is assume that is not a shortcut

ccordoba12: What are the additional conf options? Could you give an example?

andfoy: If a plugin were to include new configuration options to `shortcuts`, it will introduce the `additional_configuration` key on the configuration. This is added when a plugin is registered on the preferences one: 

See https://github.com/spyder-ide/spyder/blob/3f7aba19d90c6e873b639ff5420afd642961494c/spyder/api/plugins.py#L703 and https://github.com/spyder-ide/spyder/blob/3f7aba19d90c6e873b639ff5420afd642961494c/spyder/plugins/preferences/plugin.py#L134

ccordoba12: Ok, thanks for the explanation. It's clearer for me now.

ccordoba12 (approved):
Looks good to me now, thanks @steff456!

Note: The failures in our tests are unrelated to this. I fixed them in PR #15739 (for now).

PR #15582, opened 2021-05-11
PR: Use a notification-based manager to load and manage plugins during startup

andfoy:
## Description of Changes
This PR replaces the plugin loading machinery, moving it from a fixed-order toposort logic to use asynchronous loading based on notifications, this way plugins can have bidirectional dependencies amongst them by decoupling the loading steps into two, one that initializes the plugin and other that connects a plugin with others once a required/optional plugin is available.

This change would eventually enable Spyder to load plugins in parallel, thus reducing the loading time (potentially).

* [x] Wrote at least one-line docstrings (for any new functions)
* [x] Added unit test(s) covering the changes (if testable)

* [ ] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

### Issue(s) Resolved

Fixes #

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct:

PR #15629, opened 2021-05-15
PR: Improve how to start Spyder in debug mode and show log in Tools menu

mrclary:
## Description of Changes

Currently, when Spyder is run in debug mode, the Tools menu provides access to the LSP logs, but Spyder's own logging output still requires a deliberate commandline option (`--debug-output`) and no convenient way to access it.

This PR brings parity to the log files by putting the `spyder-debug.log` file in the user's configuration directory (similar to the LSP logs), and provides access in the Tools menu. The commandline option `--debug-output=file` will still place `spyder-debug.log` in the current working directory of the commandline environment.

When Spyder is run in debug mode, logging is output to both terminal and `spyder-debug.log`.

Additionally, a menubar item "Restart in DEBUG mode" is added to the File menu which allows the user to restart Spyder in debug mode (verbose). This should make it easier to create debug logs, especially especially for standalone application users. The "Restart" menubar item will always restart Spyder in standard mode. Programmatic requests to restart Spyder (via the signal `sig_restart_requested`) will restart Spyder in the current mode.

This PR moved the restart logic from `mainwindow.py` to the `Application` plugin and the preferences reset logic from the `Application` plugin to the `Preferences` plugin.

### Affirmation

@mrclary


mrclary:
@ccordoba12, what do you think?

spyder/app/mainwindow.py
```diff
@@ -2253,6 +2252,7 @@ def main(options, args):
                                       CONF.get('main', 'high_dpi_scaling'))
 
     # **** Set debugging info ****
+    delete_debug_log_files()
```
ccordoba12, line 2255: Why did you move this call here? Also, you didn't add the logger call that was present before. Please restore that.

spyder/app/cli_options.py
```diff
@@ -111,16 +111,6 @@ def get_options(argv=None):
...
-    parser.add_argument(
-        '--debug-output',
-        default='terminal',
-        dest="debug_output",
-        choices=['terminal', 'file'],
-        help=("Print internal debugging info either to the "
-              "terminal or to a file called spyder-debug.log "
-              "in your current working directory. Default is "
-              "'terminal'.")
-    )
```
ccordoba12, line 123: Please revert this.

spyder/app/start.py
```diff
@@ -151,6 +151,8 @@ def main():
     if options.debug_info:
         levels = {'minimal': '2', 'verbose': '3'}
         os.environ['SPYDER_DEBUG'] = levels[options.debug_info]
+    elif CONF.get('main', 'startup_in_debug_mode'):
+        os.environ['SPYDER_DEBUG'] = '3'
```
ccordoba12, line 155: I don't like this one bit. The debugging state shouldn't be set in our Preferences but instead using environment variables in our `app/restart.py` script. Else Spyder could be left in debug mode forever if users forget to change the value of that option.

spyder/app/utils.py
```diff
@@ -112,13 +115,9 @@ def setup_logging(cli_options):
...
-        if cli_options.debug_output == 'file':
-            log_file = 'spyder-debug.log'
-            handlers.append(
-                logging.FileHandler(filename=log_file, mode='w+')
-            )
-        else:
-            log_file = None
+        handlers.append(
+            logging.FileHandler(filename=SPYDER_DEBUG_LOG_FILE, mode='w+')
+        )
```
ccordoba12, line 120: Please make this work for our command line interface and using a `Restart Spyder in debug mode` action in the Tools menu.

spyder/config/main.py
```diff
@@ -81,6 +81,7 @@
               'use_custom_cursor_blinking': False,
               'show_internal_errors': True,
               'check_updates_on_startup': True,
+              'startup_in_debug_mode': False,
```
ccordoba12, line 84: There's no need of this. Please remove it.

spyder/plugins/application/confpage.py
```diff
@@ -74,6 +74,8 @@ def setup_page(self):
                                     "them to Github"), 'show_internal_errors')
         check_updates = newcb(_("Check for updates on startup"),
                               'check_updates_on_startup')
+        start_in_debug = newcb(_("Startup in DEBUG mode"),
```
ccordoba12, line 77: I think there's no need to add an option for this, as I mentioned above.

ccordoba12 (commented):
Thanks a lot @mrclary for your contribution! Although I understand this is necessary for our Windows and macOS apps, it makes uploading our debug file to Github when starting Spyder from a terminal harder than before. That's something I'd like to preserve for Anaconda users, our main distribution channel by far. Besides, outputting to a debug file from the command line is also valuable when Spyder can't be started at all.

So I suggest to keep the current command line interface and instead do the following:

* Add an entry in the `Tools` menu that restarts Spyder in debug mode and outputs to a file in our config directory.
* Leave the entry you added to easily access that file from the Spyder interface, which is a really nice addition.

spyder/app/mainwindow.py
mrclary: The spyder-debug log file from a previous session must be deleted _before_ `setup_logging` is called, otherwise it will be deleted after the logger opens the file resulting in no log file.

spyder/app/start.py
mrclary: Okay. I will rework this.

spyder/app/utils.py
mrclary: will do

spyder/config/main.py
mrclary: yep, will revert cli option, add restart action to Tools, and use environment variable in restart.py

spyder/app/cli_options.py
mrclary: I am unclear about what purpose this serves. The LSP logs are written to files if Spyder is in debug mode. Shouldn't Spyder debug files also be written if in debug mode? Is there an issue if debug files are always written if debug is enabled?

ccordoba12: I mentioned my reasons in my main review https://github.com/spyder-ide/spyder/pull/15629#pullrequestreview-666225912 above:

> it makes uploading our debug file to Github when starting Spyder from a terminal harder than before. That's something I'd like to preserve for Anaconda users, our main distribution channel by far. Besides, outputting to a debug file from the command line is also valuable when Spyder can't be started at all.

mrclary: I'm sorry, I remember what you said before, I just don't completely understand it. I'm happy to revert it; I just want to understand better the points you made.
How does it make uploading the debug file to Github harder? Previously one had to use both `--debug-info=[minimal|verbose]` and `--debug-output=file`. With this they only need `--debug-info=[minimal|verbose]` (the file will be created if in debug mode).
How does this affect the situation when Spyder can't be started at all? Again, the file is create by virtue of the debug mode; it doesn't require changing anything while the app is running.

To be clear, this doesn't preclude creating a debug file from the commandline; it just no longer requires `--debug-output=file` to create it.

ccordoba12: > How does it make uploading the debug file to Github harder?

Because you have to find the debug file in our config directory, instead of getting it from the current working directory. I find that very convenient, especially on Linux where our config directory is hidden.

> To be clear, this doesn't preclude creating a debug file from the commandline; it just no longer requires --debug-output=file to create it.

Ohhh, ok, I understand your point now. So is it possible to have both cases? I mean, when running from the command line to generate the file in the current working directory and when using the Tools menu entry to save it in our config dir?

mrclary: Okay, I see what you mean now. This currently only sets one location for the debug file. I'll change it to what you describe.
* I'll revert the commandline option
* When the commandline option is used, then the debug file will go to the current directory.

spyder/app/mainwindow.py
```diff
@@ -1072,13 +1073,27 @@ def create_edit_action(text, tr_text, icon):
...
+        restart_in_debug_mode_action = create_action(
+            self,
+            _("Restart in DEBUG mode"),
+            None, icon=ima.icon('restart'),
+            triggered=self.restart_in_debug_mode,
+            tip=_("Restart in DEBUG mode"))
+        mainmenu.add_item_to_application_menu(
+            restart_in_debug_mode_action,
+            menu_id=ApplicationMenus.Tools
+        )
```
ccordoba12, line 1095: Please move this action and the code to add it to the Tools menu to the Application plugin (i.e. to `plugins/application/container.py`).

spyder/app/mainwindow.py
```diff
@@ -1151,16 +1166,17 @@ def __getattr__(self, attr):
...
-    def update_lsp_logs(self):
-        """Create an action for each lsp log file."""
-        self.menu_lsp_logs.clear()
-        lsp_logs = []
+    def update_debug_logs(self):
+        """Create an action for each lsp and debug log file."""
+        self.menu_debug_logs.clear()
+        debug_logs = []
         files = glob.glob(osp.join(get_conf_path('lsp_logs'), '*.log'))
+        files += glob.glob(os.environ['SPYDER_DEBUG_FILE'])
```
ccordoba12, line 1174: Is there a need to use `glob.glob` on this line if `SPYDER_DEBUG_FILE` is a file path and not a directory?

spyder/app/mainwindow.py
```diff
@@ -1983,9 +1996,21 @@ def reset_spyder(self):
...
+    def restart_in_debug_mode(self):
+        """
+        Quit and restart Spyder in debug mode.
+        """
+        answer = QMessageBox.warning(self, _("Warning"),
+             _("Spyder will restart in debug mode: <br><br>"
+               "Do you want to continue?"),
+             QMessageBox.Yes | QMessageBox.No)
+        if answer == QMessageBox.Yes:
+            self.restart(debug=True)
```
ccordoba12, line 2008: This can be moved to the Application plugin.

ccordoba12 (commented):
This looks pretty nice @mrclary, thanks! I left a second review for you below.

spyder/app/mainwindow.py
mrclary: No, there is not a need. I'll remove glob

spyder/app/mainwindow.py
mrclary: Perhaps this should be in the File menu, beneath "Restart", instead of the Tools menu?

spyder/app/mainwindow.py
mrclary: Should the reset_spyder just above it also be moved to the Application plugin?

spyder/app/mainwindow.py
ccordoba12: I was thinking that too and it makes a lot of sense.

@dalthviz, @isabela-pf, what do you think?

ccordoba12: So the idea is to add a "Restart in debug mode" entry either to the File or Tools menu to allow Spyder to print a lot of debugging diagnostics.

spyder/app/mainwindow.py
ccordoba12: Yeah, it makes sense too.

spyder/app/mainwindow.py
mrclary: That's right. In particular, for non-expert users of the macOS application, this will make it much easier to produce debug output to share with developers.

mrclary: I may need some tutelage on this. I'm confused about how this is supposed to work. I'm looking at the Restart stuff as an example but can't quite make sense of it.
So, the menu item for Restart is created in application/plugin.py and has a callable action assigned, `Application.restart_action`. `Application.restart_action` is just a wrapper to `ApplicationContainer.restart_action` in application/container.py. `ApplicationContainer.restart_action` has a callable, `PluginMainContainer.sig_restart_requested`, which is just a `Signal`. But I can't find that signal emitted anywhere. 
To further confuse things, the `Application` plugin in application/plugin.py also has a signal inherited from `SpyderPluginV2.sig_restart_requested`.

It seems to me, in my naive understanding, that if we want the menubar items created in the `Application` plugin, then the menubar item should just be attached to the `Applicationi.restart` method directly, or it should emit the `sig_restart_requested` (`PluginMainContainer.sig_restart_requested` or `SpyderPluginV2.sig_restart_requested`?) signal and the signal should be connected to `Application.restart`.  Am I misunderstanding something?

dalthviz: @mrclary yep can be quite confusing but basically all the plugins have the signal to requested a restart (event the Application plugin) but the actual plugin that implements the functionality is the Application plugin. Then what we are doing is using the mainwindow as the central receptor of this signals (for plugins and plugin containers I think) but at the end what the mainwindow does to handle the request is to ask the Application plugin to do the restart.

Also, inside the plugin (the plugin vs container relation), I think we ended up doing it like that to keep all the actions definitions in the same place (in this case the container). If we want to still keep all the actions in the  container then the new action should also trigger a `sig_restart_requested` but passing the debug flag (then we need also to add a new signature for the `sig_restart_requested` signal for SpyderV2 and PluginMainContainer ) or probably create a new signal that triggers a restart on debug mode (and add the logic to handle the new signal in a similar way as what is done for the `sig_restart_requested`).

The other option is to move the creation of the action to the plugin and directly use as trigger the restart method (one action will use the flag to call the restart with the debug flag an the other just the default restart).

mrclary: > @mrclary yep can be quite confusing but basically all the plugins have the signal to requested a restart (event the Application plugin) but the actual plugin that implements the functionality is the Application plugin. Then what we are doing is using the mainwindow as the central receptor of this signals (for plugins and plugin containers I think) but at the end what the mainwindow does to handle the request is to ask the Application plugin to do the restart.

Yes, I see that. It was confusing since the mainwindow actually handled the menu item and the menu item did not reference the signal, just called `self.restart`->`Application.restart`. While it connects the signal to `self.restart`, the signal is never emitted anywhere, so is never used. By moving the menu handling to the `Application` plugin, we would would also want to remove any reference to `sig_restart_requested` from mainwindow, since the signal would be handled in the plugin.

> Also, inside the plugin (the plugin vs container relation), I think we ended up doing it like that to keep all the actions definitions in the same place (in this case the container). If we want to still keep all the actions in the container then the new action should also trigger a `sig_restart_requested` but passing the debug flag (then we need also to add a new signature for the `sig_restart_requested` signal for SpyderV2 and PluginMainContainer ) or probably create a new signal that triggers a restart on debug mode (and add the logic to handle the new signal in a similar way as what is done for the `sig_restart_requested`).
> 
> The other option is to move the creation of the action to the plugin and directly use as trigger the restart method (one action will use the flag to call the restart with the debug flag an the other just the default restart).

We are already using the container paradigm, so I'll stick with that and add a signature for `sig_restart_requested`; but I'm still confused as to why we need `sig_restart_requested` defined in two places. Shouldn't we be trying to have a single source for the definition? Also, doesn't this make two separate `Signal` instances, so that it matters which one the slot is connected to? In other words, slots connected to `SpyderPluginV2.sig_restart_requested` don't get executed if `PluginMainContainer.sig_restart_requested` is emitted?

mrclary:
@dalthviz, I've got everything working, but `test_config_dialog` is failing because the mock dialogue for preferences doesn't have a `self.main.application` plugin defined. I don't know how to fix that. If you could help me out, I'd appreciate it.

dalthviz:
@mrclary will try to give it a check :+1:

spyder/api/plugins.py
```diff
@@ -748,6 +748,18 @@ class SpyderPluginV2(QObject, SpyderActionMixin, SpyderConfigurationObserver):
...
+    This signal can be emitted to request the main application to restart with
+    reset preferences.
+    """
+
+    sig_restart_debug_requested = Signal()
+    """
+    This signal can be emitted to request the main application to restart in
+    DEBUG mode.
+    """
+
```
ccordoba12, line 762: If we want to move all the restart logic to the Application plugin, I think having these signals (and `sig_restart_requested`) makes little sense. Instead, plugins that need to perform a restart should require Application and call their restart methods directly.

However, if we want to keep using signals, we should preserve the current mechanism, i.e. connecting slots in the main window to those signals as we do now:

https://github.com/spyder-ide/spyder/blob/77d11c006dd197a05bb029c9310b33c4c222c4b3/spyder/app/mainwindow.py#L224

Personally, I find that logic very convoluted and hard to follow: plugins need to emit signals, those signals are connected in the main window to slots that end up calling Application methods. So why not call those methods directly?

The current problem with that is that the Preferences plugin would require Application to perform a restart,  but Application requires Preferences to register its config page. So that would introduce a circular dependency between those plugins.

The good thing is that problem will be avoided when we merge PR https://github.com/spyder-ide/spyder/pull/15582 because it will remove our current linear dependency plugin mechanism for a notification based one.

@mrclary, as someone who has tried to understand how this code works, what would you prefer? I mean, using signals or calling Application methods directly?

mrclary: @ccordoba12, perhaps we can schedule a zoom discussion?

mrclary: @ccordoba12, I don't feel strongly one way or the other. From your previous suggestion to move actions to the plugin, I (apparently incorrectly) thought that included the signal handling. I can move things back as before, keeping the signal handling in the main window, but putting actions in the plugin.

mrclary: > @mrclary will try to give it a check 👍

@dalthviz, thanks, but hold off since I'll be making changes again.

ccordoba12: @mrclary, we'll discuss this topic with the rest of the team and have an answer for you soon. Please don't make more changes until then.

mrclary: @ccordoba12, I'd like to be part of that discussion; will we schedule a zoom meeting or something?

ccordoba12: Good thinking! Sorry, I didn't want to make feel excluded, I just didn't know you wanted to be part of that discussion.

Could you make it tomorrow at 12 pm central time? Or perhaps this Monday at 13:00, central time too?

mrclary: I can do both, thanks!

ccordoba12: Great! I'll send you the invite for Monday then because it's easier for me.

ccordoba12:
Hey @mrclary, PR #15582 was merged today. So you can start reworking this one with the new plugin registration mechanism.

mrclary:
@ccordoba12 , @dalthviz I think this is ready for review.

dalthviz (commented):
Thanks @mrclary LGTM :+1: although seems like now the reset is done by setting the env variable to reset before calling the restart instead of a flag?

mrclary:
> Thanks @mrclary LGTM 👍 although seems like now the reset is done by setting the env variable to reset before calling the restart instead of a flag?

Previously, when activating reset from either the menubar or from the Preferences widget, it would call the restart slot located in the mainwindow with the `reset=True` keyword argument. The only purpose of the reset keyword argument was to set the environment variable for reset prior to actually restarting. Since we appropriately moved this slot to the application plugin, I also thought it would be more transparent to have the preferences plugin handle the reset environment variable and the restart slot in the application plugin just handles restart, nothing else. Then, other plugins can just access the `sig_reset_preferences_requested` signal to reset preferences.

spyder/plugins/preferences/widgets/configdialog.py
```diff
@@ -71,8 +72,9 @@ def __init__(self, parent=None):
         self.setLayout(vlayout)
 
         # Signals and slots
-        if self.main:
-            self.button_reset.clicked.connect(self.main.reset_spyder)
+        container = self.main.get_plugin(Plugins.Preferences).get_container()
+        self.button_reset.clicked.connect(
+            container.sig_reset_preferences_requested)
```
ccordoba12, line 77: This is not correct. The idea is for widgets to not have access to plugins. Instead, please add a signal to this dialog and emit it at this point like this:

```python
self.button_reset.clicked.connect(self.sig_reset_preferences_requested)
```

Then connect that signal to the container one of the same name when the dialog is created in this method:

https://github.com/spyder-ide/spyder/blob/552017c858cacc10a27285707c6e60651d0f49e7/spyder/plugins/preferences/widgets/container.py#L41-L42

like this:

```python
dlg.sig_reset_preferences_requested.connect(
    self.sig_reset_preferences_requested)
```

spyder/plugins/preferences/plugin.py
```diff
@@ -279,6 +282,23 @@ def on_toolbar_available(self):
             section=MainToolbarSections.ApplicationSection
         )
 
+    @on_plugin_available(plugin=Plugins.Application)
+    def on_application_available(self):
+        container = self.get_container()
+        container.sig_reset_preferences_requested.connect(self.reset)
+
+    @Slot()
+    def reset(self):
```
ccordoba12, line 291: I think we should move the code of this method to Application because it doesn't have to do with Preferences, but it's another kind of restart.

Then we could write this one (which is required due to the signal that comes from the container), like this

```python
@Slot()
def reset(self):
    application = self.get_plugin(Plugins.Application)
    application.reset()
```

spyder/app/utils.py
```diff
@@ -143,8 +140,8 @@ def filter(self, record):
             root_logger.addHandler(handler)
 
 
-def delete_lsp_log_files():
-    """Delete previous dead Spyder instances LSP log files."""
+def delete_debug_log_files():
```
ccordoba12, line 143: This could be moved to `plugins/application/utils.py`, since it's not necessary in the main window anymore.

ccordoba12 (commented):
@mrclary, this is a new review for you.

spyder/app/mainwindow.py
```diff
@@ -237,7 +237,6 @@ def register_plugin(self, plugin_name, external=False, omit_conf=False):
         plugin.sig_exception_occurred.connect(self.handle_exception)
         plugin.sig_free_memory_requested.connect(self.free_memory)
         plugin.sig_quit_requested.connect(self.close)
-        plugin.sig_restart_requested.connect(self.restart)
```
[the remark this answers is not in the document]
mrclary: I guess I'm confused. The first option effectively makes the mainwindow a central dispatcher, so other plugins don't access the application plugin directly. I thought we were moving away from that.
Also, the second option doesn't require calling `application.restart` or connecting another slot to `sig_restart_requested`. If the plugin wants to request a restart, it just needs to emit the signal.

Perhaps I'm still confused about the architecture. My understanding is as follows:
1. The application plugin owns the `restart` slot and the `sig_restart_requested` signal. They are connected.
2. Any plugin that wants to _listen_ for `sig_restart_requested`, must require the application plugin and connect the signal to whatever slot it wants executed, e.g. shut something down. `application.restart` doesn't need to be connected here, since that is already connected to the signal in the application plugin and will be executed when the signal is emitted (no matter which plugin emits it).
3. Any plugin that wants to _request_ a restart must require the application plugin and simply emit `application.sig_restart_requested`. It should not call `application.restart`.

spyder/app/mainwindow.py
```diff
@@ -1064,11 +1063,13 @@ def create_edit_action(text, tr_text, icon):
...
-            self.menu_lsp_logs.aboutToShow.connect(self.update_lsp_logs)
+
+        # Debug logs
+        if get_debug_level() >= 2:
+            self.menu_debug_logs = QMenu(_("Debug logs"))
+            self.menu_debug_logs.aboutToShow.connect(self.update_debug_logs)
             mainmenu.add_item_to_application_menu(
-                self.menu_lsp_logs,
+                self.menu_debug_logs,
                 menu_id=ApplicationMenus.Tools)
```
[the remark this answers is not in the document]
mrclary: okay.

spyder/app/mainwindow.py
```diff
@@ -1301,7 +1300,7 @@ def handle_dpi_change_response(self, result, dpi):
             CONF.set('main', 'normal_screen_resolution', False)
             CONF.set('main', 'high_dpi_scaling', True)
             CONF.set('main', 'high_dpi_custom_scale_factor', False)
-            self.restart()
+            self.application.sig_restart_requested.emit()
```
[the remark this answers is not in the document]
mrclary: ...But then any other plugin that might be listening for that signal won't get it. See my previous comment.

spyder/plugins/application/plugin.py
```diff
@@ -58,6 +58,8 @@ def on_initialize(self):
         self.main_menu_available = False
         self.shortcuts_available = False
 
+        self.sig_restart_requested.connect(self.restart)
+
```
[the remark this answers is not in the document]
mrclary: the application plugin doesn't emit the signal here, it connects the slot. If another plugin emits this signal, then the application plugin knows to execute `self.restart`. Perhaps I don't fully understand what you mean. See my previous comment.

spyder/plugins/preferences/api.py
```diff
@@ -809,11 +809,7 @@ def prompt_restart_required(self):
         answer = QMessageBox.information(self, msg_title, msg,
                                          QMessageBox.Yes | QMessageBox.No)
         if answer == QMessageBox.Yes:
-            self.restart()
-
-    def restart(self):
-        """Restart Spyder."""
-        self.main.restart()
+            self.main.application.sig_restart_requested.emit()
```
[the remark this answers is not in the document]
mrclary: See previous comments

spyder/plugins/preferences/plugin.py
mrclary: Respectfully, I disagree. Resetting the preferences is not really another restart, it is really just setting the environment variable. The current behavior includes restarting, so I didn't change that, but fundamentally this is a preferences plugin action. For example, in the abstract, hypothetically, if we removed the preferences plugin, there would be an orphan `reset` slot in the application plugin that would not mean anything.

spyder/plugins/preferences/plugin.py
```diff
@@ -279,6 +282,23 @@ def on_toolbar_available(self):
...
+        answer = QMessageBox.warning(self.main, _("Warning"),
+             _("Spyder will restart and reset to default settings: <br><br>"
+               "Do you want to continue?"),
+             QMessageBox.Yes | QMessageBox.No)
+        if answer == QMessageBox.Yes:
+            os.environ['SPYDER_RESET'] = 'True'
+            application = self.get_plugin(Plugins.Application)
+            application.sig_restart_requested.emit()
```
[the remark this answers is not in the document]
mrclary: see previous comment

spyder/plugins/preferences/widgets/configdialog.py
mrclary: something has to emit the `sig_reset_preferences_requested` instance from the preferences plugin container. What you're suggesting is another signal connection between the widget and the widget container; whatever the merits of this are, it will not result in resetting the preferences. Unless you meant that the widget's container `sig_reset_preferences_request` then be connected to the preferences plugin container?

spyder/app/mainwindow.py
```diff
@@ -1112,16 +1113,17 @@ def __getattr__(self, attr):
             pass
         return super().__getattr__(attr)
 
-    def update_lsp_logs(self):
-        """Create an action for each lsp log file."""
-        self.menu_lsp_logs.clear()
-        lsp_logs = []
+    def update_debug_logs(self):
```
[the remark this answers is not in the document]
mrclary: I have to put this in the plugin rather than the container in order to access the Editor plugin for menu item triggered actions. We should talk about this.

spyder/app/mainwindow.py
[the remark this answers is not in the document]
ccordoba12: Ok, sorry for introducing more confusion myself. This comment helped me to understand your rationale behind this PR and I think it's pretty good.

Please ignore all my comments about this and proceed with your initial plan.

spyder/app/utils.py
```diff
@@ -145,8 +142,8 @@ def filter(self, record):
...
+    """Delete previous dead Spyder instances LSP and debug log files."""
```
ccordoba12, line 146:
```suggestion
    """Delete previous debug log files."""
```

I think the previous docstring was a bit hard to understand.

spyder/plugins/application/container.py
```diff
@@ -331,3 +349,19 @@ def report_missing_dependencies(self):
             message_box.setWindowTitle(_('Error'))
             message_box.setText(message)
             message_box.show()
+
+    @Slot()
+    def restart_normal(self):
+        # always restart in standard mode
```
ccordoba12, line 355:
```suggestion
        """Restart in standard mode."""
```

We can make this a docstring.

spyder/plugins/application/plugin.py
```diff
@@ -121,6 +126,25 @@ def on_mainwindow_visible(self):
             container.give_updates_feedback = False
             container.check_updates(startup=True)
 
+    def update_debug_logs(self):
```
ccordoba12, line 129: Please mode this method to the container. The standard we decided to settle in for all plugins is to declare actions, menus and other graphical things in the container and only leave in the plugin a small interface to be called from others.

spyder/plugins/application/plugin.py
```diff
@@ -121,6 +126,25 @@ def on_mainwindow_visible(self):
...
+
+        debug_logs = []
+        files = glob.glob(os.path.join(get_conf_path('lsp_logs'), '*.log'))
+        files.append(os.environ['SPYDER_DEBUG_FILE'])
+        for f in files:
+            action = self.create_action(
+                f, f, triggered=load)  # TODO: add triggered
+            action.setData(f)
+            debug_logs.append(action)
+        add_actions(self.debug_logs_menu, debug_logs)
```
ccordoba12, line 146: `add_actions` is a low level function that in general should not be used. The way to add actions to menus is by using the PluginMainContainer/PluginMainWidget API, like this:

```python
for action in debug_logs:
    self.add_item_to_menu(
        action,
        menu=self.debug_logs_menu,
    )
```

spyder/plugins/application/plugin.py
```diff
@@ -139,6 +167,13 @@ def _populate_tools_menu(self):
                 menu_id=ApplicationMenus.Tools,
                 section=ToolsMenuSections.Tools)
 
+        if get_debug_level() >= 2:
+            mainmenu.add_item_to_application_menu(
+                self.debug_logs_menu,
+                menu_id=ApplicationMenus.Tools,
+                section=ToolsMenuSections.Extras)
+            self.debug_logs_menu.aboutToShow.connect(self.update_debug_logs)
```
ccordoba12, line 175: I like this way of using `aboutToShow`, pretty cool!

spyder/plugins/preferences/plugin.py
ccordoba12: Ok, I think you're right. Let's leave it here then.

spyder/plugins/preferences/widgets/configdialog.py
ccordoba12: > Unless you meant that the widget's container sig_reset_preferences_request then be connected to the preferences plugin container's sig_reset_preferences_request?

Well, the container `sig_reset_preferences_request` is already connected to `self.reset`, and that's all that we need.

So, the thing is that Qt can chain signals without calling `emit` everywhere. So this code

    self.button_reset.clicked.connect(self.sig_reset_preferences_requested)

is equivalent to

    self.button_reset.clicked.connect(self.sig_reset_preferences_requested.emit)

and the same is true for this one:

    dlg.sig_reset_preferences_requested.connect(self.sig_reset_preferences_requested)

So when `button_reset` is clicked, it would emit `sig_reset_preferences_requested` at the dialog level, which then would emit `sig_reset_preferences_requested` at the container level, which finally would call `self.reset`.

I know this is convoluted, but (again) it's the standard for all other plugins. `ConfigDialog` is special because it has a reference to the main window, through which you can call `get_plugin`. But that's not the case for any other widget in the new API.

ccordoba12 (commented):
Thanks @mrclary for your continued work on this! If you manage to address my comments soon, I think I could include your PR in 5.1.2, to be released on Tuesday or Wednesday of next week.

spyder/plugins/preferences/widgets/configdialog.py
mrclary: Okay, I think I understand. Let me take another stab at it to see If I understand correctly.

spyder/plugins/application/container.py
```diff
@@ -151,11 +162,41 @@ def setup(self):
...
+        files.append(os.environ['SPYDER_DEBUG_FILE'])
+        for f in files:
+            action = self.create_action(
+                f, os.path.basename(f), tip=f,
+                triggered=lambda: self.load_log_file(f), overwrite=True)
+            action.setData(f)
+            debug_logs.append(action)
+        add_actions(self.menu_debug_logs, debug_logs)
+        # for action in debug_logs:
+        #     self.add_item_to_menu(action, self.menu_debug_logs)
```
mrclary, line 198: @ccordoba12, `add_item_to_menu` does not populate the menu for some reason...but `add_actions` does.

ccordoba12:
Hey @mrclary, I have some fixes to finally merge this one. So please don't push or rebase here until I push my changes. Thanks!

ccordoba12:
My fixes are the following:

* Use normal QPushButton's for Verbose/Minimal kind of restarts:

    imagen

    Before you were associating `Yes` to Verbose and `No` to Minimal, which is not correct from a UX point of view:

    imagen

* Fix triggered state of debug log actions (before all actions were opening the same file).

* Use `add_item_to_menu` (from the container API) to add debug actions to menu (it required calling the low level `_render` method because the menu is built dynamically). This allowed me to create two sections for those actions: one for `spyder-debug.log` and other for the LSP log files:

    imagen

ccordoba12 (approved):
Thanks a lot for your help (and patience) with this @mrclary!

mrclary:
Looks good, thanks @ccordoba12. I suppose my only critique is that the window title is "question" instead of something more informative, like "Restart in debug mode" or something.

mrclary:
@ccordoba12, thanks for helping me out with this and for your patience; I'm learning a lot!

ccordoba12:
> I suppose my only critique is that the window title is "question" instead of something more informative, like "Restart in debug mode" or something.

I left "Question" because you had that before. But I have no problem in changing it in another PR. I'd suggest "Debug mode" though because the text in the dialog already talks about restarting in debug mode.

> thanks for helping me out with this and for your patience; I'm learning a lot!

Great!

mrclary:
@ccordoba12, I did not know that I had the previous window title as "Question"; perhaps it is the default for Question type QMessageBox. It appears that QMessageBox does not display the window title on macOS. QDialog does, though. Just FYI.

ccordoba12:
> perhaps it is the default for Question type QMessageBox

Nope, you can set the title you want.

> It appears that QMessageBox does not display the window title on macOS. QDialog does, though. Just FYI.

Up to you then if you want to change that to use a QDialog.

PR #15702, opened 2021-05-26
PR: Force reading of path stored in Spyder configuration folder as utf-8

### LINKED ISSUE #15692: PYTHONPATH manager does not work with unicode characters in path
## Problem Description
Unable to add a path with a unicode character, for example "ö", using the path manager.
Adding the path actually works, and the string is written to the `path` file under spyder configuration directory, but it is not read back from there. If later adding an additional path the unicode path is also removed from the `path` file.

Tested on 4.2.5 and 5.0.3. If it turns out an easy fix would be great to have this on both branches.

1. Add a set of paths, some with unicode characters.
Example:
image

2. Click OK, see that all the path strings are written to the `path` file under spyder configuration directory
3. Reopen the PYTHONPATH manager, unicode paths are not populated in the list. Try to import from the unicode paths (not working). 
image

It is possible to add this type of path using sys.path.append(), and the expectation is that the path manager can also handle this.

## Versions
* Spyder version: 5.0.3 None
* Python version: 3.7.9 64-bit
* Qt version: 5.12.10
* PyQt5 version: 5.12.3
### Dependencies

<details>
```

# Mandatory:
intervaltree                  :  None (OK)
IPython >=7.6.0               :  7.23.1 (OK)
jedi =0.17.2                  :  0.17.2 (OK)
keyring >=17.0.0              :  23.0.1 (OK)
nbconvert >=4.0               :  6.0.7 (OK)
paramiko >=2.4.0              :  2.7.2 (OK)
parso =0.7.0                  :  0.7.0 (OK)
pygments >=2.0                :  2.9.0 (OK)
pylint >=1.0                  :  2.8.2 (OK)
pyls >=0.36.2;<1.0.0          :  0.36.2 (OK)
pyls_black >=0.4.6            :  0.4.6 (OK)
pyls_spyder >=0.3.2;<0.4.0    :  0.3.2 (OK)
qstylizer >=0.1.10            :  0.2.0 (OK)
qtawesome >=1.0.2             :  1.0.2 (OK)
qtconsole >=5.1.0             :  5.1.0 (OK)
qtpy >=1.5.0                  :  1.9.0 (OK)
setuptools >=39.0.0           :  56.2.0 (OK)
sphinx >=0.6.6                :  3.5.4 (OK)
spyder_kernels >=2.0.3;<2.1.0 :  2.0.3 (OK)
watchdog                      :  1.0.2 (OK)
zmq >=17                      :  22.0.3 (OK)

# Optional:
cython >=0.21                 :  0.29.23 (OK)
matplotlib >=2.0.0            :  3.4.2 (OK)
numpy >=1.7                   :  1.19.3 (OK)
pandas >=1.1.1                :  1.2.4 (OK)
scipy >=0.17.0                :  1.6.3 (OK)
sympy >=0.7.3                 :  1.8 (OK)

```
</details>

rhkarls:
## Description of Changes

See issue #15692 for description.
The `path` file in the spyder configuration directory, which holds the user provided paths added with the PYTHONPATH manager is not read with correct encoding, and thus not allowing for unicode characters in the paths (e.g. 'ö'). The current read functionality is handled by `encoding.readlines()`, with subsequent calls to `encoding.read()`,` encoding.decode()`, `encoding.get_coding()` which finally calls `chardet.UniversalDetector()`. `chardet` returns (at least sometimes) the wrong encoding as per issue #15692. 

Other fixes were considered
- write BOM to the path file to make sure `chardet` detects utf-8. Using BOM with utf-8 appears, however, to be generally discourages (https://stackoverflow.com/questions/2223882/whats-the-difference-between-utf-8-and-utf-8-without-bom/2223926#2223926)
- rewrite functions in `encoding.py` to allow for defining the encoding when reading a file (note that `read`/`readlines `has a kw encoding, but this is ignored and not used). However, would require more changes, and considered unnecessary complicated and more likely to break other parts of the code due to the extensive use of the `read()` function in `utils/encoding.py`

The proposed and simple solution in this PR simply reads the file with encoding set to utf-8. The `path` file is written by `encoding.write()` and `encoding.encode()` should encode the text using utf-8. 

### Issue(s) Resolved

Fixes #15692

### Affirmation

I certify the above statement is true and correct: Reinert Huseby Karlsen

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -1858,12 +1858,14 @@ class MainWindow(QMainWindow):
     def load_python_path(self):
         """Load path stored in Spyder configuration folder."""
         if osp.isfile(self.SPYDER_PATH):
-            path, _x = encoding.readlines(self.SPYDER_PATH)
+            with open(self.SPYDER_PATH, 'r', encoding='utf-8') as f:
+                path = f.read().splitlines()
             self.path = tuple(name for name in path if osp.isdir(name))
 
         if osp.isfile(self.SPYDER_NOT_ACTIVE_PATH):
-            not_active_path, _x = encoding.readlines(
-                self.SPYDER_NOT_ACTIVE_PATH)
+            with open(self.SPYDER_NOT_ACTIVE_PATH, 'r',
+                      encoding='utf-8') as f:
+                not_active_path = f.read().splitlines()
             self.not_active_path = tuple(name for name in not_active_path
                                          if osp.isdir(name))
 


ccordoba12 (approved):
@rhkarls, thanks a lot for your contribution! Very clean and simple solution to the problem you reported.

PR #15778, opened 2021-06-03
PR: Disable Kite call-to-action and dialog if installers are not available

andfoy:
## Description of Changes
Since the future of Kite is unclear, the Kite team asked us to disable all Kite call-to-action and dialog inside Spyder if the installers were to be unavailable at any point.

### Issue(s) Resolved

Fixes #

### Affirmation

### DIFF
--- a/spyder/plugins/completion/providers/kite/utils/status.py
+++ b/spyder/plugins/completion/providers/kite/utils/status.py
@@ -109,7 +109,7 @@ def status(extra_status=''):
 def check_kite_installers_availability():
     """Check if Kite installers are available."""
     url = LINUX_URL
-    if os.name == 'os':
+    if os.name == 'nt':
         url = WINDOWS_URL
     elif sys.platform == 'darwin':
         url = MAC_URL
--- a/spyder/plugins/completion/providers/kite/utils/status.py
+++ b/spyder/plugins/completion/providers/kite/utils/status.py
@@ -15,10 +15,16 @@ import sys
 
 # Third-party imports
 import psutil
+import requests
 
 NOT_INSTALLED = 'not installed'
 RUNNING = 'ready'
 NOT_RUNNING = 'not running'
+
+# Installer URLs
+WINDOWS_URL = "https://release.kite.com/dls/windows/current"
+LINUX_URL = "https://release.kite.com/dls/linux/current"
+MAC_URL = "https://release.kite.com/dls/mac/current"
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


spyder/plugins/completion/providers/kite/utils/status.py
```diff
@@ -98,3 +104,21 @@ def status(extra_status=''):
         return RUNNING + extra_status
     else:
         return NOT_RUNNING + extra_status
+
+
+def check_kite_installers_availability():
+    """Check if Kite installers are available."""
+    url = LINUX_URL
+    if os.name == 'os':
```
dalthviz, line 112:
```suggestion
    if os.name == 'nt':
```

spyder/plugins/completion/providers/kite/widgets/calltoaction.py
```diff
@@ -79,7 +79,9 @@ def __init__(self, textedit, ancestor):
         self.hide()
 
         is_kite_installed, __ = check_if_kite_installed()
-        if is_kite_installed:
+        installers_available = check_kite_installers_availability()
+
+        if is_kite_installed or not installers_available:
```
dalthviz, line 84: @andfoy I think this will leave the `call_to_action` disable after the first time the installer gets unavailable but I'm unsure how to preserver the config in this case. It is ok to leave it disabled even if the installers get available later @ccordoba12 ?

dalthviz: Also, I can manually go to the settings enable the `call_to_action` and get it again. Maybe a better place to add the validation is at `handle_processed_completions`: https://github.com/spyder-ide/spyder/blob/521839a573db348acf8bc61e5e00dd732db91573/spyder/plugins/completion/providers/kite/widgets/calltoaction.py#L96-L98

Something like:

```python
    installers_available = check_kite_installers_availability()
    if not installers_available:
        return
```

dalthviz (commented):
Left some comments @andfoy besides that, and a rebase to take the fix for the test done at #15782 , this LGTM :+1:

spyder/plugins/completion/providers/kite/widgets/calltoaction.py
```diff
@@ -94,6 +95,11 @@ def handle_mouse_press(self, event):
     def handle_processed_completions(self, completions):
         if not self.get_conf('kite_call_to_action'):
             return
+
+        installers_available = check_kite_installers_availability()
```
dalthviz, line 99: @andfoy I was checking and maybe here is worthy to just use the config value (getting the response from the actual check and doing the http gave me a little bit of lag)?

dalthviz (commented):
@andfoy another little comment after checking again manually. This time it is working for me :+1:

dalthviz (approved):
@andfoy @ccordoba12 LGTM :+1:

PR #15857, opened 2021-06-13
PR: Improve startup time in several ways

ccordoba12:
## Description of Changes

- Use new lazy modules defined in Spyder-kernels for Numpy, Pandas and PIL to avoid loading them at startup. That also helps to simplify the code.
- Call `QApplication.processEvents` before and during calling `on_mainwindow_visible`, so that users don't have to wait until those methods finish before seeing the main window.
-  Compute dependencies in a thread and after a 10 seconds delay to not block the interface at startup.
- Avoid importing Matplotlib and Numpy at startup.
- Don't compute stylesheets when importing their module.

Depends on spyder-ide/spyder-kernels#259

### Affirmation


dalthviz (dismissed):
LGTM @ccordoba12 :+1:

PR #15887, opened 2021-06-16
PR: Make IPython Console widgets and other elements to use the SpyderConfigurationAccessor

dalthviz:
## Description of Changes

Move IPython Console widgets and other components to use the `SpyderConfigurationAccessor` instead of using `CONF` directly

### Affirmation

### DIFF
--- a/spyder/api/config/mixins.py
+++ b/spyder/api/config/mixins.py
@@ -13,7 +13,11 @@ import logging
 from typing import Any, Union, Optional
 import warnings
 
+# Third-party imports
+from qtpy.QtWidgets import QAction, QWidget
+
 # Local imports
+from spyder.config.gui import Shortcut
 from spyder.config.manager import CONF
 from spyder.config.types import ConfigurationKey
 from spyder.config.user import NoDefault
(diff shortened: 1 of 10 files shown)


spyder/plugins/ipythonconsole/widgets/client.py
```diff
@@ -74,20 +74,21 @@
...
     """
     Client widget for the IPython Console
 
     This widget is necessary to handle the interaction between the
     plugin and each shell widget.
     """
 
+    sig_append_to_history_requested = Signal(str, str)
+
+    CONF_SECTION = 'inspect current object'
```
ccordoba12, line 87: I think this section is incorrect. Also, I think we should the shortcuts in this widget to the IPython console (right now we're using the Internal console ones, but I think that's kind of weird).

spyder/api/config/mixins.py
```diff
@@ -201,6 +205,36 @@ def get_shortcut(self, name: str, context: Optional[str] = None) -> str:
...
+            Parent widget for the shortcut.
+        context: Optional[str]
+            Name of the context (plugin) where the shortcut was defined.
+
+        Returns
+        -------
+        shortcut: Shortcut
+            Namedtuple with the information of the shortcut as used for the
+            shortcuts preferences page.
+
```
ccordoba12, line 233:
```suggestion
```

No need to leave this empty space here.

ccordoba12 (commented):
This looks pretty good, thanks @dalthviz! I only left a tiny suggestion for you.

ccordoba12 (approved):
Thanks @dalthviz!

PR #15927, opened 2021-06-23
PR: Restore ability to ignore linting messages with inline comments (Editor)

### LINKED ISSUE #11033: How to suppress errors found by pyflakes in Spyder 4
Problem Description

This is from a post I made at the "Spyder" Google Groups. Carlos recommended I open an issue here. The post is as follows:

I just updated to Spyder 4. Nice job developers-- many of the new features (code folding,...) are very useful.

On much of my existing code files that work fine, pyflakes now identifies errors. They appear as red circles in the left margin of the editor. The error message is "Undefined name 'funcname" (pyflakes E)". The errors are because the imports of functions "funcname" are done in an exec statement; I know, I know, using exec like this isn't good form, but I do have my reasons, and this post is about pyflakes more generally than just this specific type of error.

Here are my questions:

* In the past, other linters or code analysis tools have similarly flagged these same lines for errors, for the same reason. I have been suppressing those error notifications/messages successfully by adding  " # analysis:ignore " to the end of the line. Is there a similar way to suppress the pyflakes errors ("  # analysis:ignore " does not work)? This sort of solution is really what I need, otherwise I will want to turn pyflakes off. I tried "  # NOQA" but it apparently doesn't work. I did some searching around for info on this but came up empty.

* Is there a pyflakes config file somewhere, in which I can suppress this particular error wherever it occurs?
 
* Is there a way to just turn pyflakes off? I could not find it in Preferences.

Secondary question:
* When I hover my mouse over the red circle, a pop-up box appears, not adjacent to the red circle but some distance away to the right, showing the error message. The pop-up also has a red X in the corner as if I could click it to suppress the error. As soon as I move my mouse off the red circle, towards pop-up in order to attempt to click the red X, the pop-up disappears (because it is not directly adjacent to the red circle in the left margin). So... I can never actually get my mouse over the red X. Is this the intended behavior? Is the red X meant to enable clicking to suppress the error, or not? Only curious about this; these are such common errors that I doubt I would ever click each pop-up.

```

## Versions
Python 3.6.7 64-bit | Qt 5.9.7 | PyQt5 5.9.2 | Windows 10 

* Spyder version:      4
* Python version:      3.6.7
* Operating System name/version:   Windows 10

About my Qt, PyQt, and dependencies, I already rolled back to Spyder 3 so in order to get them for Spyder 4 I will need to reestablish a new conda env; if this is truly necessary I can do it.

ccordoba12:
## Description of Changes

`# noqa`, `# NOQA` and `# analysis:ignore` inline comments are supported now.

**Before**

imagen

**After**

imagen

### Issue(s) Resolved

Fixes #11033.

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -95,9 +95,10 @@ except Exception:
 logger = logging.getLogger(__name__)
 
 
+# Regexp to detect noqa inline comments.
+NOQA_INLINE_REGEXP = re.compile(r"#?noqa", re.IGNORECASE)
 
 
-# %% This line is for cell execution testing
 @class_register
 class CodeEditor(TextEditBaseWidget):
     """Source Code Editor Widget based exclusively on Qt"""
@@ -1260,6 +1261,17 @@ class CodeEditor(TextEditBaseWidget):
 
             block = document.findBlockByNumber(start['line'])
             data = block.userData()
+
+            # Skip messages according to certain criteria.
+            # This one works for any programming language
+            if 'analysis:ignore' in block.text():
+                continue
+
+            # This only works for Python.
+            if self.language == 'Python':
+                if NOQA_INLINE_REGEXP.search(block.text()) is not None:
+                    continue
+
             if not data:
                 data = BlockUserData(self)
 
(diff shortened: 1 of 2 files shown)


dxter-zz:
Is it still not possible to suppress a specific warning, like for long lines. e.g.  # noqa: E501

ccordoba12:
No, sorry. For that we'd need to switch to flake8 as our linting provider. We hope to do that for Spyder 6.

Issue #42, opened 2015-02-16
External Console doesn't work


spyder-issues-migrator:
_From [y.bonne...@gmail.com](https://code.google.com/u/111370336049651795761/) on 2009-09-13T11:53:59Z_

What steps will reproduce the problem?

 1. Click 'Console Externe'
2. Click 'Ouvrir un interpréteur'

A Window error (Do you want to send an error report etc) pops up. Version 
beta7 (also on beta6 I think) on XP.

Treat that report witrh caution : I _think_ I remember seeing it work. It 
may be cause of some uninstall/reinstall on my side.

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=42_

spyder-issues-migrator:
_From [y.bonne...@gmail.com](https://code.google.com/u/111370336049651795761/) on 2009-09-14T01:28:03Z_

Tested on another machine - console works there. It works for the machine with a
Pythonxy install. It doesn't work on the machine with an Enthought install.

spyder-issues-migrator:
_From [pierre.raybaut](https://code.google.com/u/pierre.raybaut/) on 2009-09-19T01:39:12Z_

Let's say that Python(x,y) rocks ;)

**Status:** Done

spyder-issues-migrator:
_From [y.bonne...@gmail.com](https://code.google.com/u/111370336049651795761/) on 2009-09-19T02:46:40Z_

The distribution _and_ the editor. I think it's already my main tool. Thanks 
Pierre :)

PR #15964, opened 2021-07-01
PR: Adjust icon colors for increased contrast

isabela-pf:
## Description of Changes

Screen Shot 2021-07-01 at 3 46 05 PM
Screen Shot 2021-07-01 at 3 47 47 PM

Spyder 5's light and dark mode icons in the toolbars don't all meet [WCAG 2.1 contrast standards for non-text elements](https://www.w3.org/TR/WCAG21/#non-text-contrast). While WCAG is a web-focused set of guidelines, contrast is not a web-specific problem so this makes sense in desktop applications as well. This PR is another tiny step for a more accessible Spyder.

Changes made:
- [x] Replace definition of `ICON_2` and `ICON_3` in both light and dark modes.
- [x] Replace non-font icons that use `ICON_2` and `ICON_3` in both light and dark modes. These are all SVGs in the `images` directory.
- [x] ~~Change the disabled button opacity for toolbars to 65% opacity. (I'm still looking for this.)~~ This is no longer a part of this PR.

### Issue(s) Resolved

Fixes [ux-improvements #42](https://github.com/spyder-ide/ux-improvements/issues/42). There may be work in this area in the future to see if we can come up with a better solution.

### Affirmation

isabela-pf


isabela-pf:
I've gotten feedback that the disabled button opacity is inherited from [QDarkStyleSheet](https://github.com/ColinDuquesnoy/QDarkStyleSheet/) right now, and that we can either override it within Spyder or put in a PR and request a release of QDarkStyleSheet.

Right now I'm leaning towards overriding it simply because it seems like a small change that is potentially only needed for Spyder. I'm open to feedback on this, though.

isabela-pf:
Meeting feedback
- [ ] Update branch
- [ ] Double check that changes still work
- [ ] Make 65% opacity for disabled buttons a QDarkStyleSheet PR

isabela-pf:
I talked with @steff456 because I couldn't find anything that looked to me like a disabled button state for the toolbar buttons in QDarkStyleSheet. This does need to happen in Spyder so I'll be adding it to this PR. Here's why:
- She figured out that this is because [QDarkStyleSheet's disabled button styling](https://github.com/ColinDuquesnoy/QDarkStyleSheet/blob/26c3878fff742c535058bf5d08643d0135cde2af/qdarkstyle/qss/_styles.scss#L1157) isn't using the opacity we use in Spyder. 
- Notice how the above link shows styling for QToolButton, a widget Spyder does import. The thing is that [Spyder uses QToolButton for many button types](https://github.com/spyder-ide/spyder/blob/08ca46b766556dd7b0b4357341e96ace99f1a742/spyder/utils/stylesheet.py#L157_) and changes here might change more buttons that we expect.
- I wanted to try these changes out, and [this](https://github.com/spyder-ide/spyder/blob/d829e6db68965bfa5b28d2c36d9d4c0bbb5f566d/external-deps/qdarkstyle/docs/_static/css/theme.css#L2701) seems like the place where opacity is applied. I also thinks this is an automatically generated file and changes here will be overridden. I'm now looking for where this is generated from.

spyder/utils/palette.py
```diff
@@ -39,8 +39,8 @@ class SpyderPaletteDark:
 
     # Icon colors
     ICON_1 = Gray.B140
-    ICON_2 = Blue.B70
-    ICON_3 = Green.B60
+    ICON_2 = Blue.B80
+    ICON_3 = Green.B80
```
ccordoba12, line 43: Is this right? I mean `ICON_2` and `ICON_3` point to the same color now.

spyder/utils/palette.py
```diff
@@ -78,6 +78,9 @@ class SpyderPaletteDark:
     SPYDER_LOGO_WEB = Logos.B40
     SPYDER_LOGO_SNAKE = Logos.B50
 
+    # Colors for disabled state toolbar buttons
+    COLOR_DISABLED = ICON_1, OPACITY_DISABLED
```
ccordoba12, line 82: You missed to prefix the `OPACITY_DISABLED` by the class that contains it. So please change this to

```python
COLOR_DISABLED = ICON_1, Opacity.OPACITY_DISABLED
```

spyder/utils/palette.py
```diff
@@ -136,6 +139,9 @@ class SpyderPaletteLight:
...
```
ccordoba12, line 143: The same applies here.

ccordoba12 (commented):
Thanks @isabela-pf for helping us to improve this! I left an initial review for you.

spyder/utils/palette.py
isabela-pf: Thanks for checking! I thought I made a typo, but it's correct because they are the same value of different colors. One is Blue and the other is Green.

spyder/utils/palette.py
isabela-pf: I double checked and these also match the values discussed in [ux-improvements #42](https://github.com/spyder-ide/ux-improvements/issues/42).

spyder/utils/palette.py
ccordoba12: My bad! I read these lines too quickly and thought they were poitining to the same color.

spyder/utils/palette.py
ccordoba12: Sorry @isabela-pf, I should have been clearer. I meant this line has the same problem of `OPACITY_DISABLED` not being prefixed by `Opacity`.

isabela-pf: No no no, I'm sorry I didn't catch it! Fixed now.

ccordoba12:
Hey @isabela-pf, this requires a new release of QtAwesome with support for setting alpha in icon colors. For now I opened PR https://github.com/spyder-ide/qtawesome/pull/166 to add that functionality.

spyder/utils/palette.py
```diff
@@ -78,6 +78,12 @@ class SpyderPaletteDark:
     SPYDER_LOGO_WEB = Logos.B40
     SPYDER_LOGO_SNAKE = Logos.B50
 
+<<<<<<< HEAD
+    # Colors for disabled state toolbar buttons
+    COLOR_DISABLED = ICON_1, Opacity.OPACITY_DISABLED
+
+=======
+>>>>>>> parent of 1373e8594... Add constant and argument for disabled toolbar buttons
```
ccordoba12, line 86: Hey @isabela-pf, you left this unnecessary code after you reverted your changes. Please remove it and after that merge your branch with `5.x` to make our tests pass.

Thanks!

ccordoba12 (commented):
@isabela-pf, I left one last comment for you then this should be ready.

ccordoba12 (commented):
@isabela-pf, I'm sorry but I missed another couple of details that were missing from your revert. I left suggestions for them below.

ccordoba12 (approved):
Thanks a lot @isabela-pf for your help with this!

PR #16011, opened 2021-07-13
PR: Fix compatibility with pytest-qt >= 4

bnavigator:
This reverts #15782 and fixes the reason why pytest-qt was pinned: qtbot.addWidget() expects a widget as argument, but a few plugins were not widgets.

### Affirmation

I certify the above statement is true and correct: @bnavigator

### DIFF
--- a/spyder/plugins/completion/tests/conftest.py
+++ b/spyder/plugins/completion/tests/conftest.py
@@ -27,7 +27,7 @@ app = qapplication()
 
 # PyTest imports
 import pytest
-from pytestqt.plugin import QtBot
+from pytestqt.qtbot import QtBot
 
 
 class MainWindowMock(QMainWindow):
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -66,7 +66,7 @@ dependencies:
 - pytest-lazy-fixture
 - pytest-mock
 - pytest-order
-- pytest-qt <4.0
+- pytest-qt
 - pytest-xvfb
 - pyyaml
 - scipy
--- a/requirements/tests.txt
+++ b/requirements/tests.txt
@@ -13,7 +13,7 @@ pytest-cov
 pytest-lazy-fixture
 pytest-mock
 pytest-order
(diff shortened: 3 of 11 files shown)


ccordoba12 (commented):
Thanks for your help @bnavigator! The only problem I see with this is that pytest-qt 4 requires PyQt 5.11, according to its Changelog:

https://github.com/pytest-dev/pytest-qt/blob/master/CHANGELOG.rst#400-2021-06-03

but we're still testing with 5.9.

bnavigator:
But you can still use pytest-qt < 4 with this PR. It's just that the upper restriction is lifted.

ccordoba12:
You're totally right. However, It seems a slow test is failing. I restarted the build a couple of times and it's still failing, so please take a look at that.

bnavigator:
> so please take a look at that.

Done.

spyder/plugins/completion/tests/conftest.py
```diff
@@ -13,6 +13,10 @@
 # Third party imports
 from qtpy.QtWidgets import QMainWindow
 
+# PyTest imports
+import pytest
+from pytestqt.qtbot import QtBot
+
```
ccordoba12, line 19: Please revert this change. Without it we can't run our tests individually due to the issue mentioned in the comment below with QtAwesome.

ccordoba12 (commented):
Thanks for your help @bnavigator! One last comment, then this should be ready.

spyder/plugins/completion/tests/conftest.py
bnavigator: Are you sure about this? Neither pytest nor pytestqt.qtbot are initializing anything which you haven't imported above the comment already.

I removed the last commit from HEAD.

ccordoba12 (approved):
@bnavigator, thanks for your help with this!

*Note*: The failures in our tests are unrelated to this and are fixed in PR #16026.

PR #16012, opened 2021-07-13
PR: Add plugin teardown operations to the plugin registry

andfoy:
## Description of Changes
This PR adds plugin teardown capabilities to the already existing plugin registration mechanism. By using the teardown functions, Spyder will be able to disable and enable plugins without restarting (theoretically).  

### Issue(s) Resolved

Fixes #

### Affirmation


spyder/api/plugin_registration/registry.py
```diff
@@ -317,6 +344,147 @@ def notify_plugin_availability(self, plugin_name: str,
...
+            # Disconnect depending plugins from the plugin to delete
+            self._notify_plugin_teardown(plugin_name)
+            if isinstance(plugin_instance, SpyderPluginWidget):
+                plugin_instance._close_window()
+
+        # Delete plugin from the registry and auxiliary structures
+        self.plugin_dependents.pop(plugin_name, None)
+        self.plugin_dependencies.pop(plugin_name, None)
+
+        for plugin_ in self.plugin_dependents:
```
ccordoba12, line 397: I find the naming of `plugin_` a bit odd. Why not call that variable simply `plugin`?

spyder/api/plugin_registration/registry.py
```diff
@@ -317,6 +344,147 @@ def notify_plugin_availability(self, plugin_name: str,
...
+            all_plugin_dependents = self.plugin_dependents[plugin_]
+            for key in {'requires', 'optional'}:
+                plugin_dependents = all_plugin_dependents.get(key, [])
+                if plugin_name in plugin_dependents:
+                    plugin_dependents.remove(plugin_name)
+
+        for plugin_ in self.plugin_dependencies:
```
ccordoba12, line 404: The same here.

spyder/app/mainwindow.py
```diff
@@ -34,6 +34,8 @@
 import threading
 import traceback
 
+from attr import has
```
ccordoba12, line 37: - Please move this import to be before `from collections import OrderedDict`-
- Also remove `from __future__ import print_function` because it's no longer necessary.

spyder/app/mainwindow.py
```diff
@@ -1465,12 +1468,13 @@ def resizeEvent(self, event):
 
     def moveEvent(self, event):
         """Reimplement Qt method"""
-        if not self.isMaximized() and not self.layouts.get_fullscreen_flag():
-            self.window_position = self.pos()
-        QMainWindow.moveEvent(self, event)
+        if hasattr(self, 'layouts'):
```
ccordoba12, line 1471: I think we decided `layouts` is a core plugin, so it can't be disabled. In that case, is this check necessary?

spyder/plugins/application/plugin.py
```diff
@@ -97,7 +100,10 @@ def on_editor_available(self):
         editor = self.get_plugin(Plugins.Editor)
         self.get_container().sig_load_log_file.connect(editor.load)
 
-    def on_close(self):
+    # -------------------------- PLUGIN TEARDOWN ------------------------------
+    # @on_plugin_teardown(plugin=Plugins.)
```
ccordoba12, line 104: Please remove this comment if it's not necessary.

spyder/plugins/application/plugin.py
```diff
@@ -97,7 +100,10 @@ def on_editor_available(self):
...
+
+    def on_close(self, _unused=True):
```
ccordoba12, line 106: Why do we need to set this parameter `_unused` to `True`? Can it simply be `_unused`?

ccordoba12 (commented):
Thanks a lot @andfoy for this! Really interesting stuff.

I left some minor comments for you, otherwise looks good to me.

spyder/app/mainwindow.py
andfoy: This is necessary in order to prevent some spurious errors in some of the tests

spyder/plugins/application/plugin.py
andfoy: Both work, actually, that `_unused=True` is a leftover of the previous `cancellable` parameter.

ccordoba12: Ok, then I think we can leave it.

ccordoba12:
Sorry @andfoy but this has now a conflict due to the changes I did in PR #16416, which (unfortunately) was necessary to solve some critical issues with external plugins.

PR #16014, opened 2021-07-14
PR: Fix resetting variables after clicking on the reset button (IPython console)

ccordoba12:
## Description of Changes

- The problem was I made non-modal the warning dialog we show to inform users that variables are going to be removed. And that changed the way we need to handle the reset, which is done in this PR.
- This was broken after the changes I made in PR #15988.

### Affirmation


steff456 (approved):
This is working now :)

PR #16177, opened 2021-08-08
[WIP] PR: Break test suite into groups

ccordoba12:
## Description of Changes

Hopefully whit will allow us to run them much faster than before because they'll run in parallel.

### Affirmation


ccordoba12:
This is an interesting idea to reduce test time, but we'd need much more slots to implement it.

ccordoba12:
Closing because I have a better idea on how to implement this.

PR #16199, opened 2021-08-11
Add PYTHONHOME to shell environment when executing in external terminal from macOS app

mrclary:
## Description of Changes

Added `PYTHONHOME` environment variable to the shell environment when executing a Python script with "Execute in an external system terminal" is selected and using the macOS application.

Environment variables cannot be inherited by the shell environment in the Terminal application, so `PYTHONHOME` must be explicitly set in the executed shell script.

### Issue(s) Resolved

Fixes #16185

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct:
@mrclary

PR #16200, opened 2021-08-11
PR: Add PYTHONHOME to shell environment when executing in external terminal from macOS app

### LINKED ISSUE #16185: Mac Big Sur - Executed in external system terminal not working
### Issue Report Checklist

    * [ ] Restarted Spyder

## Problem Description
Execution in an external system terminal not working for macOS Big Sur & Spyder 5.1.1 (standalone Mac Installer).

I'm working with miniconda environments and have changed python interpreter to the correct environment. Running the code in a dedicated or open console works just fine. 

1.  Go to Preferences
2. Under Run -> Console -> Select Execute in an external system terminal
3. Run code
4. External system terminal opens but get a fatal Python error

Running code in external system terminals works just like running in console within Spyder

```python-traceback

PASTE TRACEBACK HERE

```
image

## Versions

* Spyder version: 5.1.1
* Python version: Python 3.9.5 64-bit
* Qt version: 5.12.10
* PyQt version: 5.12.3
* Operating System name/version: Darwin 20.5.0 

### Dependencies

```
PASTE DEPENDENCIES HERE
```

mrclary:
## Description of Changes

Added `PYTHONHOME` environment variable to the shell environment when executing a Python script with "Execute in an external system terminal" is selected and using the macOS application.

Environment variables cannot be inherited by the shell environment in the Terminal application, so `PYTHONHOME` must be explicitly set in the executed shell script.

Replaces PR #16199

### Issue(s) Resolved

Fixes #16185

### Affirmation

@mrclary

### DIFF
--- a/spyder/utils/programs.py
+++ b/spyder/utils/programs.py
@@ -31,12 +31,11 @@ import psutil
 
 # Local imports
 from spyder.config.base import (is_stable_version, running_under_pytest,
-                                get_home_dir)
+                                get_home_dir, running_in_mac_app)
 from spyder.config.utils import is_anaconda
 from spyder.py3compat import PY2, is_text_string, to_text_string
 from spyder.utils import encoding
 from spyder.utils.misc import get_python_executable
-
 
 HERE = osp.abspath(osp.dirname(__file__))
 
@@ -794,6 +793,8 @@ def run_python_script_in_terminal(fname, wdir, args, interact,
                                         delete=False)
         if wdir:
             f.write('cd {}\n'.format(wdir))
+        if running_in_mac_app() and executable == get_python_executable():
+            f.write(f'export PYTHONHOME={os.environ["PYTHONPATH"]}\n')
         f.write(' '.join(p_args))
         f.close()
         os.chmod(f.name, 0o777)
--- a/spyder/utils/programs.py
+++ b/spyder/utils/programs.py
@@ -800,7 +800,7 @@ def run_python_script_in_terminal(fname, wdir, args, interact,
         os.chmod(f.name, 0o777)
 
         def run_terminal_thread():
-            proc = run_shell_command('open -a Terminal.app ' + f.name)
+            proc = run_shell_command('open -a Terminal.app ' + f.name, env={})
             # Prevent race condition
             time.sleep(3)
             proc.wait()
--- a/spyder/utils/programs.py
+++ b/spyder/utils/programs.py
@@ -793,7 +793,7 @@ def run_python_script_in_terminal(fname, wdir, args, interact,
                                         delete=False)
         if wdir:
             f.write('cd {}\n'.format(wdir))
-        if running_in_mac_app() and executable == get_python_executable():
+        if running_in_mac_app(executable):
             f.write(f'export PYTHONHOME={os.environ["PYTHONPATH"]}\n')
         f.write(' '.join(p_args))
         f.close()


mrclary:
@vancromy, when this artifact is ready, please test it.
I cannot reproduce the issue that you last reported in which Terminal.app is not launched by Spyder.

mrclary:
@vancromy, can you confirm that the external terminal setting is not overriden by the per file setting?
Run->Configure per file...
Screen Shot 2021-08-11 at 1 51 18 AM

vancromy:
@mrclary The fix seems to work in terms of launching Terminal and overcoming the fatal Python error! :) It was indeed a specific file run configuration that was overwriting the default setting. Speaking of which, do you know how I can delete/remove specific file run configurations in Spyder and instead make them fallback to the option I select in the Preferences?

On the other hand, I am getting a ModuleNotFoundError for my local package (see traceback below) when running in the external system terminal which I don't get when running in the Spyder console. My suspicion is this is due to certain paths missing somewhere or it being run in a slightly different context? Is it an intended feature that running a script in a Spyder Console vs an External System terminal has different running contexts?

```
Last login: Wed Aug 11 11:07:00 on ttys000
/var/folders/fb/qd4nqwh51plf244w72s3_9880000gn/T/spyder-yvc/run_spyder_oyoo_vck.sh ; exit;
(base) yvc@166-49-235-193 ~ % /var/folders/fb/qd4nqwh51plf244w72s3_9880000gn/T/spyder-yvc/run_spyder_oyoo_vck.sh ; exit;
  File "/Users/yvc/Documents/repos/ps-granulation-window/unittest/reconciler_test.py", line 14, in <module>
    from src.utilities import run_model, save_obj
ModuleNotFoundError: No module named 'src'
Saving session...
...copying shared history...
...saving history...truncating history files...
...completed.

[Process completed]
```

mrclary:
> Speaking of which, do you know how I can delete/remove specific file run configurations in Spyder and instead make them fallback to the option I select in the Preferences?

The only way that I know to do this is to remove the file entry from the run configurations in the configuration file. Close Spyder and open `~/spyder-py3/config/transient.ini` in a text editor. Find the section `[run]` and the entry `configurations =`. `configurations` should be a list of tuples with the pattern `[(<filename 1>, <config dict 1>), (<filename 2>, <config dict 2>), ...]`. Remove the tuple associated with the file you wish to use system preference behavior.

@ccordoba12, perhaps we should add an option to the per-file configuration dialog window that would reset the file to use system preference behavior?

> On the other hand, I am getting a ModuleNotFoundError for my local package (see traceback below) when running in the external system terminal which I don't get when running in the Spyder console. My suspicion is this is due to certain paths missing

@vancromy, your suspicions are correct. When a shell is launched by Terminal.app (an interactive login shell), the environment will inherit the environment variables from the parent task (Terminal.app) _and_ source the standard bash configuration files: first `/etc/profile`, then only the first available of `~/.bash_profile`, `~/.bash_login`, or `~/.profile`. If you define your `PYTHONPATH` in your `~/.bash_profile`, it should work for you.

mrclary:
> @mrclary, I think to solve the problem reported above by @vancromy we need to pass the entire additions to sys.path that Spyder handles to the external terminal process.
> 
> Do you want to address that as part of this PR or prefer to leave it for later?

@ccordoba12, I think we should address this in #14918 as it will require a little more thought.
The issue is that when Terminal.app launches an interactive login shell, the bash configuration files will _always_ be sourced. There is no way to pass `sys.path` contents except to assign `PYTHONPATH`, which would be overwritten by a source file if it is defined there. 

This is in contrast to our IPython consoles, which _do not_ source bash configuration files. The motivation behind #14918 was to give the user the option to harmonize the behavior (environment variables) between our IPython console and external system terminals (bash source files).

#14918 should be expanded to an Environment Variables plugin, as we've discussed, which should provide a widget for user configuration. I think that we need to think about what the intended behavior should be when running in an external system terminal and whether we should make it optional. Therefore, I think it should be addressed with #14918 in a holistic manner.

vancromy:
> @ccordoba12, perhaps we should add an option to the per-file configuration dialog window that would reset the file to use system preference behavior?

+1 for this. Could even be a separate dialog box e.g. "File Run Conffigurations" that would provide users with an overview of all know file run configurations along with a "+" and "-" button to add/remove file run configurations. Could even have a cogwheel button to edit a selected file run configuration.

vancromy:
> @vancromy, your suspicions are correct. When a shell is launched by Terminal.app (an interactive login shell), the environment will inherit the environment variables from the parent task (Terminal.app) and source the standard bash configuration files: first /etc/profile, then only the first available of ~/.bash_profile, ~/.bash_login, or ~/.profile. If you define your PYTHONPATH in your ~/.bash_profile, it should work for you.

Okay. Is this something that that would need to be done each time I'm changing my conda envs?

ccordoba12:
> @ccordoba12, perhaps we should add an option to the per-file configuration dialog window that would reset the file to use system preference behavior?

Totally agree, this is a serious UX limitation. Could you open an issue about it? And, if possible, take care of it?

> @ccordoba12, I think we should address this in #14918 as it will require a little more thought.

Ok, no problem.

ccordoba12 (approved):

mrclary:
> > @vancromy, your suspicions are correct. When a shell is launched by Terminal.app (an interactive login shell), the environment will inherit the environment variables from the parent task (Terminal.app) and source the standard bash configuration files: first /etc/profile, then only the first available of ~/.bash_profile, ~/.bash_login, or ~/.profile. If you define your PYTHONPATH in your ~/.bash_profile, it should work for you.
> 
> Okay. Is this something that that would need to be done each time I'm changing my conda envs?

No. If you define PYTHONPATH in ~/.bash_profile, that environment variable will be used for all bash shell sessions. Anywhere in that file you can put 
```bash
export PYTHONPATH=/one/path:/a/second/path:/a/third/path
```

PR #16209, opened 2021-08-12
PR: Add option to per-file run configuration to use default run configuration

### LINKED ISSUE #16207: Add option to reset per-file run configurations
### The Problem
Once a per-file run configuration is established for a file, there is no convenient way to revert it to follow the global run preference. Currently the only way is to remove the file's entry in the `[run]` section of the `transient.ini` configuration file.
This was first reported in discussion regarding #16200.

### Proposed Solution
Add "Use global run settings" checkbox to the "Run configuration per file" dialog window.
Note that the configuration for any configured file can be accessed from the same dialog using the file dropdown list.
Screen Shot 2021-08-12 at 8 16 04 AM

mrclary:
## Description of Changes

* Added option to use default run configuration in the per-file run configuration
* Moved external terminal elements moved to the console group
* Behavior is unchanged except that when this option is selected the file-specific run configuration is ignored and the default run configuration takes priority.
* Note: all other file-specific run configuration options are not modified. This will allow toggling the run configuration without requiring the user to re-establish file run configuration.
* Fixed an issue with running in an external terminal if path and/or script name had spaces.

Screen Shot 2022-03-15 at 6 28 37 PMScreen Shot 2022-03-15 at 6 29 09 PM

### Issue(s) Resolved

Fixes #16207

### Affirmation

@mrclary

### DIFF
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -966,6 +966,7 @@ def test_runconfig_workdir(main_window, qtbot, tmpdir):
 
     # --- Use cwd for this file ---
     rc = RunConfiguration().get()
+    rc['default'] = False
     rc['file_dir'] = False
     rc['cw_dir'] = True
     config_entry = (test_file, rc)
@@ -1022,8 +1023,8 @@ def test_dedicated_consoles(main_window, qtbot):
 
     # --- Set run options for this file ---
     rc = RunConfiguration().get()
-    # A dedicated console is used when these two options are False
-    rc['current'] = rc['systerm'] = False
+    # A dedicated console is used when these three options are False
+    rc['default'] = rc['current'] = rc['systerm'] = False
     config_entry = (test_file, rc)
     CONF.set('run', 'configurations', [config_entry])
 
--- a/spyder/plugins/editor/plugin.py
+++ b/spyder/plugins/editor/plugin.py
@@ -58,8 +58,8 @@ from spyder.plugins.editor.widgets.status import (CursorPositionStatus,
                                                   EncodingStatus, EOLStatus,
                                                   ReadWriteStatus, VCSStatus)
 from spyder.plugins.run.widgets import (ALWAYS_OPEN_FIRST_RUN_OPTION,
-                                        get_run_configuration, RunConfigDialog,
-                                        RunConfiguration, RunConfigOneDialog)
+                                        get_run_configuration,
+                                        RunConfigDialog, RunConfigOneDialog)
 from spyder.plugins.mainmenu.api import ApplicationMenus
 
 
@@ -2928,10 +2928,6 @@ class Editor(SpyderPluginWidget, SpyderConfigurationObserver):
             if show_dlg and not dialog.exec_():
                 return
             runconf = dialog.get_configuration()
-
-        if runconf.default:
-            # use global run preferences settings
-            runconf = RunConfiguration()
 
         args = runconf.get_arguments()
         python_args = runconf.get_python_arguments()
--- a/spyder/plugins/editor/plugin.py
+++ b/spyder/plugins/editor/plugin.py
@@ -58,8 +58,8 @@ from spyder.plugins.editor.widgets.status import (CursorPositionStatus,
                                                   EncodingStatus, EOLStatus,
                                                   ReadWriteStatus, VCSStatus)
 from spyder.plugins.run.widgets import (ALWAYS_OPEN_FIRST_RUN_OPTION,
-                                        get_run_configuration,
-                                        RunConfigDialog, RunConfigOneDialog)
+                                        get_run_configuration, RunConfigDialog,
+                                        RunConfiguration, RunConfigOneDialog)
 from spyder.plugins.mainmenu.api import ApplicationMenus
 
 
@@ -2928,6 +2928,10 @@ class Editor(SpyderPluginWidget, SpyderConfigurationObserver):
             if show_dlg and not dialog.exec_():
                 return
             runconf = dialog.get_configuration()
+
+        if runconf.default:
+            # use global run preferences settings
+            runconf = RunConfiguration()
 
         args = runconf.get_arguments()
         python_args = runconf.get_python_arguments()
--- a/spyder/plugins/run/widgets.py
+++ b/spyder/plugins/run/widgets.py
@@ -28,6 +28,8 @@ from spyder.utils.qthelpers import create_toolbutton
 # Localization
 _ = get_translation("spyder")
 
+RUN_DEFAULT_CONFIG = _("Run file with default configuration")
+RUN_CUSTOM_CONFIG = _("Run file with custom configuration")
 CURRENT_INTERPRETER = _("Execute in current console")
 DEDICATED_INTERPRETER = _("Execute in a dedicated console")
 SYSTERM_INTERPRETER = _("Execute in an external system terminal")
@@ -58,6 +60,7 @@ class RunConfiguration(object):
     """Run configuration"""
 
     def __init__(self, fname=None):
+        self.default = None
         self.args = None
         self.args_enabled = None
         self.wdir = None
@@ -78,6 +81,7 @@ class RunConfiguration(object):
         self.set(CONF.get('run', 'defaultconfiguration', default={}))
 
     def set(self, options):
+        self.default = options.get('default', True)
         self.args = options.get('args', '')
         self.args_enabled = options.get('args/enabled', False)
         self.current = options.get('current',
@@ -104,6 +108,7 @@ class RunConfiguration(object):
 
     def get(self):
         return {
+                'default': self.default,
                 'args/enabled': self.args_enabled,
                 'args': self.args,
                 'workdir/enabled': self.wdir_enabled,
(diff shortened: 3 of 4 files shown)


ccordoba12:
The problem with this dialog is that it's not quite clear that current settings apply to the file selected in the dropdown at the top. Perhaps we should emphasize that better?

I say it because (I think) it won't be clear to users what the "Remove" or "Reset" buttons are supposed to do.

mrclary:
> The problem with this dialog is that it's not quite clear that current settings apply to the file selected in the dropdown at the top. Perhaps we should emphasize that better?
> 
> I say it because (I think) it won't be clear to users what the "Remove" or "Reset" buttons are supposed to do.

What do you think could improve it?
* Maybe we could put the remove button next to combo box?
* Maybe we could move the combo box outside and above the "configuration" box?

@spyder-ide/core-developers, does anyone have a suggestion?

dalthviz:
Maybe the initial idea of using the checkbox could be better if the idea is to keep the possibility to change again to the custom settings for the given file. However maybe to make more clear that the default settings are being used instead of some custom specified all the config (in case the checkbox exists) should be disabled but preserving the custom values for the preferences (although I'm unsure how complex could that be). If the idea is just to not preserve the custom config and revert everything to the defaults a reset button is better.

Also, maybe @isabela-pf could give us some ideas regarding what could be more user friendly and clear

mrclary:
Maybe the button should read "Default" or "Reset to Defaults" instead of "Remove" or "Reset"? The desired behavior is that the file run settings should follow that which is specified in Preferences.

I think users already know that the "Run" button applies to the selected file in the combo box; so perhaps users will know that "Reset to Defaults" will also apply to the selected file?

CAM-Gerlach:
Sorry I'm late to the party, but I agree with @mrclary and @dalthviz that a checkbox is greatly preferable to just a reset button (if only one must be chosen, given they actually serve somewhat orthogonal functions). It makes this functionality much more useful, by being able to quickly switch between the default and a custom run config for testing, debugging or other purposes, rather than simply wiping all the settings, while not adding any more complexity to the UI. Adding a `Reset` button could be considered as a separate feature, but given in practical use the toggle checkbox covers close to a superset of a the actual use cases a reset button would, the latter doesn't appear to be necessary as a separate feature.

The conventional/expected UX for this would be exactly as described by @dalthviz , 

> However maybe to make more clear that the default settings are being used instead of some custom specified all the config (in case the checkbox exists) should be disabled but preserving the custom values for the preferences

I don't think pathological fear of configurability should, at the very least, limit genuinely useful new functionality and and a more convenient user worklow/UX.

However, if instead the originally proposed functionality, we instead want to merely offer a button resetting the dialog contents to the current default, it should say `Reset` or `Reset to default` to match the other similar dialogs (`Remove` is potentially quite confusing and ambiguous in this context), and it should not be irreversably applied until the user clicks OK, being undoable with `Cancel` (as otherwise a user may accidentally click it and wipe out all their custom config, with no way to get it back—something else the toggle functionality would better avoid).

mrclary:
Okay, to summarize where I think we're at...
### Original Issue
The original issue is that once a file specific run setting is created, there is currently no mechanism to revert to the default behavior specified in Preferences. This was first reported in the discussion for #16200.

### Proposed Solutions
#### Checkbox
1. Provide a checkbox in the per-file run settings that allows the user to toggle between using custom run settings and falling back to the default behavior in Preferences.
2. When the checkbox is selected to use default settings, custom settings are not removed, but the items in the dialog box are disabled and greyed.
3. This solution does not provide a mechanism for culling entries in `[run]` section of the transient.ini configuration file. This may or may not be a concern.
4. No permanent actions are taken unless the "OK" button is pressed.
#### Push Button
1. Provide a push button in the per-file run settings that removes the selected file from the drop-down list and its run settings widget from the dialog's widget stack. The label for this push button is to be determined ("Reset", "Use Default", or whatever).
2. Upon "OK" button press, the dialog's current settings widget stack is saved to the `[run]` section of the transient.ini file. Note this is already the behavior of the "OK" button; there is no change here.
3. This solution does provide culling of the `[run]` section of the transient.ini file, if this is any concern.
4. This solution does not provide a mechanism for toggling between the default behavior and custom settings.
#### Other Solutions
I'm open to any other suggestions.

ccordoba12:
@mrclary, sorry for not giving you feedback about this one before. We'll discuss it @isabela-pf and @steff456 soon to see if we can find a nice UX solution to this conundrum.

mrclary:
No problem. Whatever we decide, I'll make it happen. 👍🏼

isabela-pf:
Just wanted to let you all know I'm caught up on the issue and working on a solution. Thanks!

isabela-pf:
Okay, here's an idea!

For me, the idea of the checkbox is the proposal I prefer most. It has room for the most clarifying description and allows for users to go between options without losing their custom configuration.

The main thoughts I had were
-  Users cannot simultaneously have default settings and custom settings enabled. Check boxes typically allow multiple selections; for things like this a radio button may be better suited.
- Choosing what kind of run (default or custom) is the first thing users need to answer, not the last. Because if they want default, then they have potentially wasted time filing out everything else first not knowing the option is at the bottom.
- If a user wants the default settings, none of the other options seem to matter. They just make for a large, potentially overwhelming dialog.

Instead I think it would be beneficial to try making this option a radio button, moving it to the top, and only showing the custom settings when custom is requested. Here's an idea of what that could look like in-interface. 

Spyder Run settings interface with the file path and default settings.

Spyder Run settings interface with the file path and all custom setting options listed.

Feedback is welcome! Let me know if you have any questions, too.

mrclary:
> Okay, here's an idea!
> 
> For me, the idea of the checkbox is the proposal I prefer most. It has room for the most clarifying description and allows for users to go between options without losing their custom configuration.
> 
> The main thoughts I had were
> 
>     * Users cannot simultaneously have default settings and custom settings enabled. Check boxes typically allow multiple selections; for things like this a radio button may be better suited.
> 
>     * Choosing what kind of run (default or custom) is the first thing users need to answer, not the last. Because if they want default, then they have potentially wasted time filing out everything else first not knowing the option is at the bottom.
> 
>     * If a user wants the default settings, none of the other options seem to matter. They just make for a large, potentially overwhelming dialog.

This makes perfect sense to me and I agree completely. Thank you for articulating it so well.

> Instead I think it would be beneficial to try making this option a radio button, moving it to the top, and only showing the custom settings when custom is requested. Here's an idea of what that could look like in-interface.

I like where you're going with this. Would it be better to disable and gray-out the options instead of hiding them? Some arguments for this would be:

- If the default is selected, the user can still see what _potential_ options are available.
- If the default is selected but there are previously set custom settings, this information is still visible to the user.
- The affected file is displayed in a dropdown list, so changing the selected file, or changing the custom/default radio button, will change the window size. In my opinion, as a user, this can be obnoxious and I'd prefer that window sizes remain fixed unless I deliberately resize them.

Does anyone have an opinion regarding pruning the entries in the `[run]` section of the transient.ini configuration file? I don't necessarily think that the user should be bothered with this issue; I'm just wondering if there are any concerns that it could get bloated. I get a bit concerned when there are mechanisms to _add_ but never to _remove_ items from configuration lists. Could this potentially be a source for bugs in the future? As a user, I find it inconvenient to have to "nuke" my configuration files in applications as a way to solve some bug, especially when I've got a lot of custom settings that I have to manually restore. Maybe one solution is just to check for the existence of the file on startup and remove it from the list if it does not exist.

isabela-pf:
> I like where you're going with this. Would it be better to disable and gray-out the options instead of hiding them?

I think this will work fine. My main thought was that this can still be an overwhelming window even with the content disabled, but it's a loosely held opinion for sure. I think the most compelling argument to me was your comment about users still being able to access the custom settings so they better understand their two options.

Thanks for the feedback!

mrclary:
@ccordoba12 @andfoy @isabela-pf any chance of moving forward on this soon?

isabela-pf (approved):
From what I can see, this looks good. Binder isn't working quite right at the moment, but from what I understand in the code this is what I expect. 

I'm happy to approve this for merging; we can always iterate if we find something that needs changing later.

Thank you so much for working on this and taking feedback @mrclary!

mrclary:
@isabela-pf, it looks like I'm able to merge this, but I didn't do so before because I didn't know if I needed to wait for any other reviewers. Should I merge this or wait?

isabela-pf:
I'm personally fine if you merge this. @andfoy and @ccordoba12 any objections since you also reviewed this?

andfoy:
I also agree with the proposed changes

mrclary:
@isabela-pf @andfoy @ccordoba12, sorry to make you re-approve, but I thought this should have a rebase to double check the tests, etc.

mrclary:
@ccordoba12 @isabela-pf @andfoy , just updated the PR description with the current Run configuration window.

isabela-pf (dismissed):
Thanks for working on this @mrclary!

CAM-Gerlach (commented):
This is a great change to finally see happen! One request—could you add the radio buttons at the top-level of the dialog like the file path rather than within a subsidiary `Run Settings` `QGroupBox`, as it controls the behavior of the rest of the dialog as a whole?

mrclary:
> This is a great change to finally see happen! One request—could you add the radio buttons at the top-level of the dialog like the file path rather than within a subsidiary `Run Settings` `QGroupBox`, as it controls the behavior of the rest of the dialog as a whole?

No problem.

I see something else we might want to consider. The "External system terminal" `QGroupBox` is controlled by the "Execute in external system terminal" radio button. Would it make better sense to take "External system terminal" elements and put them with "Execute in external system terminal" (underneath, indented)?

CAM-Gerlach:
> I see something else we might want to consider. The "External system terminal" QGroupBox is controlled by the "Execute in external system terminal" radio button. Would it make better sense to take "External system terminal" elements and put them with "Execute in external system terminal" (underneath, indented)?

Yes, indeed—good catch.

mrclary:
@CAM-Gerlach @isabela-pf @ccordoba12 @andfoy, I updated the PR description with the external terminal elements moved to the console group and screenshot.

CAM-Gerlach:
@mrclary Thanks! Would you mind updating the screenshot as well?

mrclary:
> @mrclary Thanks! Would you mind updating the screenshot as well?

I did.

CAM-Gerlach:
Hmm, maybe its a caching issue but I'm not seeing it above after shift-reloading, and it looks like your most recent edit was over 24 hours ago?

mrclary:
Sorry, must have forgot to hit the green button 😬

isabela-pf (approved):

CAM-Gerlach (approved):
LGTM @mrclary , thanks! Great improvement; looking forward to it!

andfoy (approved):
Thanks for this great improvement @mrclary!

andfoy:
@ccordoba12, do you have any comments about this PR? We need to merge this before continuing with the Run refactoring PR

ccordoba12 (approved):
Thanks @mrclary for your work on this!

PR #16229, opened 2021-08-17
PR: Several improvements to Find

ccordoba12:
## Description of Changes

While using the Find plugin regularly to search for code, I found several things we could change on it to improve its user experience.

The changes in this PR are the following:

* Add placeholder text in the search field. Currently there's no clear feedback on where to start with this plugin.
* Highlight searched text in matches with the same color as the one used in the editor.
* Show directory where a file is placed relative to the path we're searching in.
* Show file icons in results.
* Reorganize icons in the main toolbar to match the layout of the find/replace widget.
* Increase font size of matched text and paths.
* Speed up search by assuming certain file extensions are plain text.
* Only search in regular files and directories (i.e. not pipes).

Visually the changes look like this:

**This PR**

imagen

**Before**

imagen

### Affirmation


isabela-pf:
Hey Carlos! Thanks for this PR! I think it's matching most of what we discussed in a meeting last week. The main thing @steff456 and I noticed was that you are using two different colors for the de-emphasized text with and `COLOR_TEXT_2`and `COLOR_TEXT_3`. I totally agree that `COLOR_TEXT_3` is dim, but if it's a problem everywhere is there a reason to not choose one and stick with it?

ccordoba12:
> Maybe make the file path COLOR_TEXT_1 just like the file name. It is already styled with italics that help separate it, and COLOR_TEXT_2is not very far from COLOR_TEXT_1 so contrast doesn't even change too much.

I agree with this suggestion because it's simpler and won't introduce an extensive revision to see how we're using text against a background color in other parts of Spyder.

isabela-pf (approved):

PR #16276, opened 2021-08-25
PR: Prevent slowdowns in the Variable Explorer when Numpy and Pandas are not installed

### LINKED ISSUE #16247: Variable explorer is very slow
The variable explorer in Spyder 5.1.1 performs very slowly. Consider the following example:

```
my_list = []

for i in range(300):
    # already slow
    # my_list.append(i)

    # more complex list --> even slower variable explorer
    a = (i, (2*i, 3*i, 4*i))
    my_list.append(a)  
```

The list `my_list` appears correctly in the variable explorer. But it is almost impossible to inspect the list in the variable explorer, the respective dialog freezes constantly. Problems can be observed already for the simple test case and are more severe for the complex list. I observed this behavior on two different computers with fresh installations. Inspecting the same list in Spyder 3.3.6 is no problem.

ccordoba12:
## Description of Changes

- This was a regression introduced PR #15857.
- The solution was memoizing the results of `is_module_installed` in spyder-kernels, so it's only called once to determine if a module is installed.
- Depends on spyder-ide/spyder-kernels#313

### Issue(s) Resolved

Fixes #16247.

### Affirmation

### DIFF
--- a/external-deps/spyder-kernels/.github/workflows/linux-tests.yml
+++ b/external-deps/spyder-kernels/.github/workflows/linux-tests.yml
@@ -38,13 +38,14 @@ jobs:
            python-version: ${{ matrix.PYTHON_VERSION }} 
       - name: Install package dependencies
         shell: bash -l {0}
-        run: conda install --file requirements/posix.txt -y -q
+        run: |
+          conda install --file requirements/posix.txt -y -q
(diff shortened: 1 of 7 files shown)


dalthviz (dismissed):
@ccordoba12 LGTM. Checking locally the collections viewer is way more responsive with the proposed changes :+1:

PR #16205, opened 2021-08-11
PR: Use menu and item identifiers to add items to the main menu plugin

andfoy:
## Description of Changes
This PR updates the Main Menu plugin methods to exclusively use string identifiers as opposed to using object references when adding items to a specific menu and/or before a given menu item. This is a breaking API change.

* [ ] Wrote at least one-line docstrings (for any new functions)
* [ ] Added unit test(s) covering the changes (if testable)

* [ ] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

### Issue(s) Resolved

Fixes #

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct: @andfoy

PR #16322, opened 2021-08-31
Make Spyder compatible (for the most part) with PySide2

### LINKED ISSUE #6894: pyside2 support

rear1019:
## Description of Changes
This pull request is a series of patches which makes Spyder compatible (for the most part) with PySide2. The patches have been developed against Spyder v5.0.5 and apply cleanly to the 5.x branch as of today. There are still some issues which we are aware of but haven’t looked into yet:
* Some Exceptions which occur during editing. According to the reporter of the issue they occur rarely and there seems to be no way to reproduce them.
* QThread related issues when Spyder is closed. (I have observed this issue with PyQt5 as well.)
We will create a pull request to fix these issues once we have a fix. Aside from these two issues, Spyder is usable with PySide2.

### Issue(s) Resolved

Fixes #6894 (for the most part)

### Affirmation

rear1019


ccordoba12:
This is a great contribution, thanks a lot @rear1019!

Perhaps it's too much to ask, but would you be willing to make our test suite run and pass with Pyside2? That would be the best way to be sure that Spyder works flawlessly with PySide2 (otherwise we could introduce regressions in the future given that our tests run only with PyQt).

spyder/plugins/preferences/api.py
```diff
@@ -211,7 +211,7 @@ def load_from_conf(self):
         """Load settings from configuration file."""
         for checkbox, (sec, option, default) in list(self.checkboxes.items()):
             checkbox.setChecked(self.get_option(option, default, section=sec))
-            checkbox.clicked.connect(lambda _, opt=option:
+            checkbox.clicked[bool].connect(lambda _, opt=option:
                                      self.has_been_modified(opt))
```
ccordoba12, line 215:
```suggestion
            checkbox.clicked[bool].connect(
                lambda _, opt=option: self.has_been_modified(opt))
```

Small style issue

spyder/plugins/toolbar/container.py
```diff
@@ -48,17 +48,6 @@ class ToolbarActions:
     ShowToolbars = "show toolbars"
 
 
-class QActionID(QAction):
```
ccordoba12, line 51: Why did you remove this? We need it for some changes part of our `5.x` branch and that are going to land in 5.1.2

If possible, please restore it.

ccordoba12 (commented):
This looks pretty good! I only left one important comment for you @rear1019.

spyder/plugins/toolbar/container.py
rear1019: I have restored the code for `QActionID`. But I am afraid that the other change in the file is required. See the commit `pyside: Do not modify QAction.__class__ in toolbar plugin…`.

rear1019:
> This is a great contribution, thanks a lot @rear1019!
> 
> Perhaps it's too much to ask, but would you be willing to make our test suite run and pass with Pyside2? That would be the best way to be sure that Spyder works flawlessly with PySide2 (otherwise we could introduce regressions in the future given that our tests run only with PyQt).

This is something we actually want to do, for exactly the reason you mention. However, to be honest, this has low priority as of now. But we will definitely provide patches once we have some.

spyder/plugins/toolbar/container.py
```diff
@@ -317,7 +317,6 @@ def create_toolbars_menu(self):
         for toolbar_id, toolbar in self._ADDED_TOOLBARS.items():
             if toolbar:
                 action = toolbar.toggleViewAction()
-                action.__class__ = QActionID
```
ccordoba12, line 320: If this is not possible in PySide2, please skip it just for it so we can have it for PyQt.

I know it's a hack, but we really need it until we're able to remove it in some months.

ccordoba12 (commented):
Besides addressing the comment I left for you, please merge with 5.x to make our tests pass.

Then I'd say this one is ready.

spyder/plugins/toolbar/container.py
rear1019: I removed this change (the pull request now doesn’t modify `container.py`).

ccordoba12: Ok, but how did you manage to solve the problem with PySide2?

rear1019: I did not. Changing `__class__` of a `QObject` created by C++ (e.g. `QMainWindow::addToolbar(QString)`, `QToolBar::addAction(QString)`, `QToolBar::toggleViewAction()`) seems to invalidate the corresponding Python (PySide) object. I have no idea how to work around this, if possible at all. (Changing `__class__` of `QObject` created in Python (e.g. `a = PySide2.QtWidgets.QAction()`) seems to work.)

Can you point to a PR or issue where `QActionID` is used? My question being: Why exactly is `QActionID` and changing of `__class__` required? Do you want to change the type of `QAction` to `QActionID` (because you you want `isinstance(ins, QActionID)` to be true)? Or do you use `QActionID` just to add the `action_id` attribute?

rear1019:
Just pushed two more fixes. We are currently evaluating one more patch.

RahulARanger (changes requested):
with this PR one can use _general [widgets](https://github.com/spyder-ide/spyder/wiki/Dev:-Spyder-Internals#widgets)_ using PySide2.

I believe this line need not be [changed](https://github.com/spyder-ide/spyder/blob/75ebd6abc953e80fa70f02fb775d95c081911ad9/spyder/plugins/outlineexplorer/widgets.py#L725) as `is` matches for a memory location and `!=` looks for it's values

spyder/plugins/outlineexplorer/widgets.py
```diff
@@ -722,7 +722,7 @@ def __sort_toplevel_items(self):
                 self.ordered_editor_ids if
                 self.editor_items.get(e_id) is not None]
 
-        if current_ordered_items != new_ordered_items:
+        if current_ordered_items is not new_ordered_items:
```
RahulARanger, line 725: We can have this previous one, as `is not` matches for the memory location whereas `!=` looks for values. Sorry if I am wrong.
```suggestion
        if current_ordered_items != new_ordered_items:
```

spyder/plugins/toolbar/container.py
ccordoba12: This is the PR that added the change: https://github.com/spyder-ide/spyder/pull/16205

As I said, if this is not possible for PySide2 but things are working fine without it, you can add an `if not PYSIDE2` (qtpy has that constant) so it's not used in that case.

rear1019: > As I said, if this is not possible for PySide2 but things are working fine without it, you can add an if not PYSIDE2 (qtpy has that constant) so it's not used in that case.

Sorry, I misunderstood that this is what you want to do. I will shortly push this change.

spyder/plugins/outlineexplorer/widgets.py
rear1019: You are right, strictly speaking the patch is not correct. I haven’t noticed this because it should be safe to remove the `if` clause, i.e. it is safe to always execute the code inside the `if` block (`is not` always returns `True` here because two distinct `list` object are compared). The `if` clause looks like some kind of optimization, which only applies in the following case (as far as I can tell):
* The "Show all files" option is enabled in the Outline plugin
* The tabs in the editor are alphabetically ordered and the "Sort files alphabetically" option is _off_
* The "Sort files alphabetically" option is toggled.

There are two options:
* Remove the `if` clause
* Add compatibility code for PySide2 (which could differentiate if the version is  `<= 5.15.0`. The bug which fixes `==`/`!=` has been fixed in `5.15.1` [1])

[1] https://bugreports.qt.io/browse/PYSIDE-74
[2] https://codereview.qt-project.org/c/pyside/pyside-setup/+/312945

spyder/plugins/toolbar/container.py
rear1019: Done, see commit "pyside: Do not modify __class__ in toolbar plugin (prevents startup)"

ccordoba12: Thanks! That's exactly what I had in mind.

spyder/plugins/outlineexplorer/widgets.py
ccordoba12: > The bug which fixes ==/!= has been fixed in 5.15.1

Well, since the support for PySide2 is brand new, you could require `>=5.15.1` and leave the `if` in place.

rear1019: Do you mean the global requirement (the one checked in `spyder/requirements.py` at startup) or should the following code be used in `outlineexplorer/widgets.py`:
```py
        # Always update if PySide <= 5.15.0 is used because == and != comparisons are not supported
        # for the data types inside the lists
        update = (
            (PYSIDE2 and parse_version(PYSIDE_VERSION) <= parse_version("5.15.0"))
            or (current_ordered_items != new_ordered_items)
        )
        if update:
            …
```

I prefer the latter or completely removing the `if` clause, but the decision is up to you.

ccordoba12: Ok, let's go with the code you posted above then. I'm fine with it.

rear1019:
Force push 2021-09-20:
* Rebased against current `5.x` branch
* Added code suggested in https://github.com/spyder-ide/spyder/pull/16322#discussion_r710778412

spyder/plugins/outlineexplorer/widgets.py
rear1019: Done in force push on 2021-09-20

ccordoba12:
@rear1019, please let me know if there's more planned work for this or I can leave my final review.

rear1019:
> @rear1019, please let me know if there's more planned work for this or I can leave my final review.

We are still evaluating patches for rarely occurring issues during editing. I will add the patch or by mid of next week.

rear1019:
> > @rear1019, please let me know if there's more planned work for this or I can leave my final review.
> 
> We are still evaluating patches for rarely occurring issues during editing. I will add the patch or by mid of next week.

Done

ccordoba12:
Hey @rear1019, this PR has conflicts now. Please solve them so I can leave my final review.

rear1019:
> Hey @rear1019, this PR has conflicts now. Please solve them so I can leave my final review.

Rebased against current `5.x` branch. No further changes.

ccordoba12:
@rear1019, please use an editor or IDE that preserves line endings in files, or adjust your current one to do so. That's because I found out this PR introduced mixed line endings in several files touched by it.

PR #16339, opened 2021-09-02
PR: Test installer

### LINKED ISSUE #14138: Run Tests on macOS Application
It would be very useful to be able to run tests on the macOS application artifact.
This would greatly help identify any issues that may impact the macOS application but can't be identified by running tests on source code.

Is it possible to run pytest on the spyder module _inside_ the artifact?

mrclary:
### Issue(s) Resolved

Fixes #14138

### Affirmation


mrclary:
Okay, proof of principle:
* I can successfully launch the application in the pipeline, capture the output, and set a timeout.
* Can I ensure that the job fails if errors occur within the application?
* Can I raise errors in the application conditional on running the app in ci environment?

mrclary:
@ccordoba12, does Qt generate this dialog box?
Screen Shot 2021-09-03 at 9 53 21 AM

I need a way to exit Spyder when an error occurs, but it must circumvent this pop-up. Any ideas?

mrclary:
My idea is not to run a full test suite necessarily, but to have a few strategic test points in the code base. 
Conditional on:
* Running macOS app
* Running in CI environment

The first things to consider are:
* Does the application launch?
* Does the IPython kernel launch with the internal interpreter?
If either of these occur, I'd like the application to exit (with an error code) but not trigger the aforementioned dialog.

ccordoba12:
> @ccordoba12, does Qt generate this dialog box?

No, it doesn't. As far as I know it's a macOS feature.

ccordoba12:
> My idea is not to run a full test suite necessarily, but to have a few strategic test points in the code base.

I agree with your plan.

mrclary:
`sys.exit(1)` will terminate the application without invoking the dialog.

mrclary:
Raising an Exception does seem to invoke the dialog since the timeout is reached. However, the process still exited with code 143, the exception is captured in the stdout, and the job failed. So perhaps I don't need to worry about the dialog.

Now, will the job succeed if the timeout is reached and an error has not been raised, or `sys.exit(0)` invoked?

mrclary:
Okay, without an exception raised, the job succeeds.

mrclary:
Okay, timeout is not reached and job succeeds with `sys.exit(0)`

mrclary:
Conclusion
-----------
* Job succeeds if:
  * The application does not raise any errors. In this case it will be timed out and terminated.
  * The application exits with `sys.exit(0)`. In this case it will not be timed out.
* Job fails if:
  * The application raises an error. In this case it will be timed out and terminated.
  * The application exits with `sys.exit(<nonzero>)`. In this case it will not be timed out.

So it looks like a viable solution!
All that remains is to place some conditional `Exceptions` in the source code for missing dependencies and IPython kernel failures.

This should catch the lion share of issues that arise with the macOS application. However, it should be easy to incorporate additional conditional `Exceptions` for particular cases that may arise. The key here is that these should only need to detect internal errors or warnings that don't already result in aborting the application but indicate an issue with the macOS application.

mrclary:
Ha, ha, ha! I didn't realize that we'd patched sys.exit. I need to use ORIGINAL_SYS_EXIT.

ccordoba12:
> Ha, ha, ha! I didn't realize that we'd patched sys.exit. I need to use ORIGINAL_SYS_EXIT.

Good point! I forgot we do that.

ccordoba12:
I've been thinking about this for some time and I'd like to suggest the following to test our installers in PRs:

1. Add pytest and pytest-qt to the installer.
2. Move the `mainwindow` fixture in `app/tests/test_mainwindow.py` to its own `conftest.py` file in the same directory.
3. Copy that file to some folder (not necessarily inside the installer).
4. Create a file with a series of simple tests using that fixture, e.g. loading a file, running code, checking variables are displayed in the Variable explorer and performing a bit of code completion.
5. Run that file using the interpreter's file (or perhaps through a simplified version of `runtests.py`).

I think pytest-qt is the simplest, more robust and the only multi-platform way of automatically testing Spyder. So I think it could help us in this case too.

mrclary:
> I've been thinking about this for some time and I'd like to suggest the following to test our installers in PRs:
> 
>     1. Add pytest and pytest-qt to the installer.
> 
>     2. Move the `mainwindow` fixture in `app/tests/test_mainwindow.py` to its own `conftest.py` file in the same directory.
> 
>     3. Copy that file to some folder (not necessarily inside the installer).
> 
>     4. Create a file with a series of simple tests using that fixture, e.g. loading a file, running code, checking variables are displayed in the Variable explorer and performing a bit of code completion.
> 
>     5. Run that file using the interpreter's file (or perhaps through a simplified version of `runtests.py`).
> 
> 
> I think pytest-qt is the simplest, more robust and the only multi-platform way of automatically testing Spyder. So I think it could help us in this case too.

Is it okay if we keep the scope small for the current PR and later expand the testing in a separate PR?
I like the idea of doing something relatively simple right now to catch the majority of the issues that have historically afflicted the macOS app (missing packages, packages that aren't zip safe, and bad environment variables) and later expanding the project to catch more corner cases or more nuanced behaviors.

ccordoba12:
> Is it okay if we keep the scope small for the current PR and later expand the testing in a separate PR?

Sure, no problem with that.

mrclary:
@ccordoba12, there seems to be some issues with the version of spyder-kernels. I'm not sure what's going on, but the subrepo version is 2.2.0dev0 but it shows up as 3.0.0dev0 in the environment list. This is causing the app test to fail. I suspect it is causing other tests to fail as well since they pass on my local machine...

mrclary:
@ccordoba12, I've noticed on my local machine I have Clang 12.0.5, while Github has Clang 11.0.0. I don't know if this is causing the issue.

ccordoba12:
> I'm not sure what's going on, but the subrepo version is 2.2.0dev0 but it shows up as 3.0.0dev0 in the environment list.

I think the problem is that the base of your PR was `master`, but you really did it against `5.x`. I changed that and retriggered a build of the Mac installers to see of that fixes the problem.

mrclary:
> > I'm not sure what's going on, but the subrepo version is 2.2.0dev0 but it shows up as 3.0.0dev0 in the environment list.
> 
> I think the problem is that the base of your PR was `master`, but you really did it against `5.x`. I changed that and retriggered a build of the Mac installers to see if that fixes the problem.

Doh! sorry about that.

spyder/app/utils.py
```diff
@@ -325,3 +334,68 @@ def create_window(WindowClass, app, splash, options, args):
...
+
+        super().__init__(msg, *args, **kwargs)
+
+        # Print the stack trace and provided message,
+        # then exit Spyder with code 1.
+        # Note: raising exceptions will not print the stack trace
+        stack = ['Traceback (most recent call last):\n']
+        stack.extend(traceback.format_stack()[:-2])
+        stack.extend(traceback.format_exception_only(type(self), self))
+        print(''.join(stack))
```
ccordoba12, line 359:
```suggestion
        print(''.join(stack))  # spyder: test-skip
```

This is why our fast tests are failing.

mrclary: Actually, I suppose the logger should be preferred over print.

spyder/app/utils.py
```diff
@@ -325,3 +334,59 @@ def create_window(WindowClass, app, splash, options, args):
     if not running_under_pytest():
         app.exec_()
     return main
+
+
+class SpyderInstallerError(object):
```
ccordoba12, line 339: Let's move this and the other classes you added here to `spyder/utils/installers.py`. That's because they are not used in `app/mainwindow.py` but all over the place instead.

spyder/config/base.py
```diff
@@ -68,6 +68,11 @@ def running_in_ci():
     return bool(os.environ.get('CI'))
 
 
+def running_installer_test():
```
ccordoba12, line 71: I think this function could also go in `utils/installers.py` because it's only used there.

ccordoba12 (commented):
This is a really improvement, thanks @mrclary! I left a couple of small comments for you, then it should be ready.

ccordoba12 (approved):

ccordoba12:
I'll merge after we merge PR #16012 to not introduce more conflicts on it.

PR #16375, opened 2021-09-07
PR: Improve appearance of toolbar extension button

ccordoba12:
## Description of Changes

- The size and icon of that button were almost imperceptible. This PR aims to improve that.
- Thanks to the new API, this fix will apply to both the main and pane toolbars as well.

**Before**

imagen

**After**

imagen

### Affirmation

### DIFF
--- a/spyder/api/widgets/toolbars.py
+++ b/spyder/api/widgets/toolbars.py
@@ -20,11 +20,15 @@ from qtpy.QtWidgets import (
 
 # Local imports
 from spyder.api.exceptions import SpyderAPIError
+from spyder.api.translations import get_translation
 from spyder.utils.icon_manager import ima
 from spyder.utils.qthelpers import SpyderAction
 from spyder.utils.stylesheet import (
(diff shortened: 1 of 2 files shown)


isabela-pf:
Hooray! I'm very glad to see this happening, especially since I think it's a really big experience change. Especially if people make extensions that use the toolbars, which are already tight on space. 

A few points addressing questions we discussed when meeting:

- Buttons need to have at least the minimum accessible clickable area. At that point, it's probably best to make it match the other buttons, too.
- Since we need the minimum clickable area, the vertical dots (`mdi-dots-vertical`) won't really save space. I think `mdi-dots-horizontal` is the more familiar version of this icon and may be preferable to use.

Thanks for doing this!

ccordoba12:
Thanks @isabela-pf for the review! I made the changes you requested and updated the screenshot above.

However, I still have to verify how it looks on Windows and macOS. I'll let you know when that's done.

steff456:
In mac the size of the widgets of the main toolbar 54x46 and in the plugin's toolbars the size is 42x42. You need to fix this.

PR #16389, opened 2021-09-09
PR: Add constraint for jupyter_client version on conda based test

dalthviz:
## Description of Changes

CI tests with conda failing due to `jupyter_client` 7.x being available in the conda main channel

### Affirmation

### DIFF
--- a/.github/scripts/install.sh
+++ b/.github/scripts/install.sh
@@ -24,6 +24,10 @@ if [ "$USE_CONDA" = "true" ]; then
         conda install pyzmq=19
     fi
 
+    # Constrain jupyter_client version on conda based tests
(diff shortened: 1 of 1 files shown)


ccordoba12 (commented):
@dalthviz, now that I see our tests failing I recalled the problem with the approach you're trying so far. Since we have a test that checks our dependencies among `setup.py`, Binder, `conda.txt` and `dependencies.py` are in sync, you'd need to add this constraint to all of them.

But in this case that's not really necessary because we depend indirectly on `jupyter-client` (through Spyder-kernels). So the right thing to do is to add it to `.github/scripts/install.sh`.

PR #16457, opened 2021-09-20
PR: Change cursor to pointing hand when single click option is on in Files and Projects

ccordoba12:
## Description of Changes

I think this makes more intuitive to use that option because, thanks to browsers, you know what to expect when clicking on something and the cursor has that shape.

Single-click

### Affirmation

### DIFF
--- a/spyder/plugins/explorer/widgets/explorer.py
+++ b/spyder/plugins/explorer/widgets/explorer.py
@@ -562,7 +562,8 @@ class DirView(QTreeView, SpyderWidgetMixin):
         self.context_menu.aboutToShow.connect(self.update_actions)
 
     @on_conf_change(option=['size_column', 'type_column', 'date_column',
-                            'name_filters', 'show_hidden'])
+                            'name_filters', 'show_hidden',
+                            'single_click_to_open'])
     def on_conf_update(self, option, value):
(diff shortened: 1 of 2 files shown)


isabela-pf (commented):
This seems fine to me, especially since this is so easy to toggle on and off as needed.

My main question is around whether or not single-click interferes with existing actions that users might want to take when only selecting a directory or file. For example, selecting a file and right-clicking to rename it. Playing with this more showed that making a selection and opening a context menu is still possible with single-click on, so I think it works.

It would be best if this experience was the same across all panes with trees/file navigation like this. To my knowledge, that includes Files, Projects, and Outlines.

Thanks for asking for feedback!

isabela-pf (approved):

ccordoba12:
Thanks for the review @isabela-pf! About your comments:

> My main question is around whether or not single-click interferes with existing actions that users might want to take when only selecting a directory or file

I tested this and verified single-click doesn't interfere with other operations you can do on files and directories.

> It would be best if this experience was the same across all panes with trees/file navigation like this. To my knowledge, that includes Files, Projects, and Outlines.

Files and Projects are covered by this PR. I'll work on making the same change for the Outline (which I think is important to be consistent) in another one.

PR #16488, opened 2021-09-24
PR: Clean up sys.path and entry-points in bootstrap

mrclary:
## Description of Changes

* Remove extraneous external-deps paths from sys.path
* Do not remove python-lsp-server egg-info at exit
* Create Spyder egg-info and use real plugin entry-points rather than patching provider entry-points to fake Spyder distribution.

### Comments
Adding the external-deps (python-lsp-server, qdarkstyle, and spyder-kernels) to the sys.path is not necessary. Whether they are installed in site-packages (as normal or on the CI) or whether they are installed in develop mode (under DEV) they will be included in Python's search path. Explicitly adding them to the sys.path here creates duplicate and redundant entries.

Removing the egg-info for python-lsp-server breaks its installation in other environments that have it installed in develop mode. There is no need to remove the egg-info anyway, regardless of the installation method.

Regarding the plugin entry-points, I've proposed this before and was unable to convince others to adopt it, but I will try again. Using Spyder's egg-info for plugin entry-point ensures that the actual entry-point mechanism is used (which is what should be tested) and `python setup.py egg_info` _only creates the egg-info and does not change existing environments_ .

If I'm mistaken on any of these points, please enlighten me :-)

### Affirmation

@mrclary

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -130,21 +130,8 @@ if args.debug:
     # if operation is successful
 
 
-# Add this path to the front of sys.path
-sys.path.insert(0, DEVPATH)
-print("*. Patched sys.path with %s" % DEVPATH)
-
-# Add external dependencies subrepo paths to be the next entries
-# (1, 2, etc) of sys.path
-DEPS_PATH = osp.join(DEVPATH, 'external-deps')
-i = 1
-for path in os.listdir(DEPS_PATH):
-    external_dep_path = osp.join(DEPS_PATH, path)
-    sys.path.insert(i, external_dep_path)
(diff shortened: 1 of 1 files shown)


mrclary:
@ccordoba12, this is not a critical PR, so I changed it to a draft.
I don't want to make the same communication error that I've made recently, so I will ask:
* It seems to me that adding the external-deps to the sys.path are not necessary, since installing them in developer mode automatically includes them in the sys.path. Doing so here adds duplicate entries. Am I missing something? Is there a corner case that I'm not aware of?
* Why do we need to manually register the entry-points instead of using Spyder's egg-info?

PR #16490, opened 2021-09-24
PR: Code sign macOS app

mrclary:
## Description of Changes

### Issue(s) Resolved

Fixes #

### Affirmation


mrclary:
I've been unable to get this to work. I've stripped down the workflow to just test the keychain operations. I can't seem to import the certificate into the keychain...

ccordoba12:
Do the commands you're using work locally?

mrclary:
I don't have a certificate to test with locally. But all the preceding commands work locally.

mrclary:
I was able to "successfully" sign Spyder's macOS application locally on my machine using a self-signed certificate that I created locally. By "successful" I mean that the keychain and code signing commands were completed without error. It will remain to be seen whether the process results in an _actual_ signed app bundle after trading out the self-signed cert for our spyder-ide one.

The next step is to get the self-signed cert to work in the CI.

mrclary:
Okay, it looks like I was able to get everything to run without error. I downloaded the artifact and it looks like the code signature is in the app (self-signed).

mrclary:
Okay, still having an issue importing the certificate into the keychain.
@juanis2112, can you try to setup the certificate and GitHub secrets again?
1. Ensure that you have the right certificate and private key. I think you can create several Developer ID Applications, so it should be okay to make a new one if you think that will help. I know that you were having trouble with your 2FA, so if you need to resolve that first, that's okay.
2. Delete the previous .p12 file that you created and re-export the certificate and private key to `Certificate.p12`. Save the password you create for this file as a Github repository secret (not environment secret) labeled `MACOS_CERTIFICATE_PWD`, replacing the existing secret already stored.
3. run `base64 Certificate.p12 | pbcopy`; the certificate and private key should now be in your clipboard (it should be about 4000 characters), and save that as a Github repository secret (not environment secret) labeled `MACOS_CERTIFICATE`, replacing the existing secret already stored. I noticed when trying this on my forked repository that pasting the certificate into the value field for the secret showed a newline after all the characters; I don't know if this will cause a problem or not.

I'm sorry to request that you do this again, but I think it may be possible that the certificate got messed up somehow or it got stored as Github secret incorrectly. Unfortunately, we cannot view the Github secrets in order to verify that they are correct.

mrclary:
Okay, I'm pretty sure that the issue must be with how the Github secrets were stored. I created a merge request on my forked repository of Spyder. Since I have ownership privileges there, I was able to set the Github secrets for my self-signed certificate and its password. Below, are the results of the workflows in spyder-ide and mrclary workspaces, respectively. spyder-ide fails, while mrclary succeeds. Notice the difference in the specifications for `MACOS_CERTIFICATE` and `MACOS_CERTIFICATE_PWD`: spyder-ide shows them to be empty, while mrclary obscures the values with `***`.

I had previously believed that Github obscured the secrets by not showing any value. Comparing this to the case for mrclary shows that this is not the case: Github obscures the secrets with `***` and the secrets for spyder-ide are not correct.

spyder-ide:
```bash
  ./test_sign.sh
  shell: /bin/bash -l {0}
  env:
    LITE_FLAG: --lite
    DMGNAME: Spyder-Lite.dmg
    DISTDIR: /Users/runner/work/spyder/spyder/dist
    pythonLocation: /Users/runner/hostedtoolcache/Python/3.9.5/x64
    MACOS_CERTIFICATE: 
    MACOS_CERTIFICATE_PWD: 
security: SecKeychainItemImport: One or more parameters passed to a function were not valid.
Error: Process completed with exit code 1.
```

mrclary:
```bash
   ./test_sign.sh
  shell: /bin/bash -l {0}
  env:
    LITE_FLAG: --lite
    DMGNAME: Spyder-Lite.dmg
    DISTDIR: /Users/runner/work/spyder/spyder/dist
    pythonLocation: /Users/runner/hostedtoolcache/Python/3.9.5/x64
    MACOS_CERTIFICATE: ***
    MACOS_CERTIFICATE_PWD: ***
```

@juanis2112, before trying to create any new certificates, can you just try recreating the Github secrets as I outlined in my previous comment?

mrclary:
I'm not going to worry about notarizing the dmg file right now. That appears to be a little more involved. If needed, we can open another issue and PR to do that later.

ccordoba12:
Agreed.

juanis2112:
@mrclary I think it is best if I give you the account credentials for both Spyder's GitHub account and the Mac developers account so that we don't  risk me doing it wrong again lol. @ccordoba12 say he is fine with it so if you agree I can send you the information.

mrclary:
That's fine with me. Just let me know if there is anything I need to do.

mrclary:
> I'm not going to worry about notarizing the dmg file right now. That appears to be a little more involved. If needed, we can open another issue and PR to do that later.

We may actually need to notarize for Catalina and above, or Gatekeeper may still give users a problem. This may be straightforward if we can save our Apple Developer credentials as Github secrets as well.

mrclary:
@ccordoba12 any chance of moving forward on this soon?

ccordoba12:
I'll try to set up a meeting with you in a couple of weeks to resume this work.

mrclary:
@juanis2112, will you contact me via email or private message regarding this PR?

ccordoba12:
> All I need is to be able to set the GitHub secrets in the spyder-ide/spyder GitHub repo (developer certificate, certificate password, and developer application password).

I just granted you Admin permissions, so you should be able to set the secrets that you need.

mrclary:
I think I've identified the issue for this PR: repository secrets are not released for PRs triggered from forked repositories.
https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow

ccordoba12:
> I'm curious how we go about solving code signing and/or notarization issues when they only run on releases?

I think you can create in-repo branches for that (as PR #18508).

> Does this workflow have the potential to fail?

Perhaps, but it hasn't failed so far (i.e. in more than a year and a half of producing installers).

> Can we submit new fix PRs, if needed? How does that work?

Sure. Although we try to only submit PRs from forks, if there's a need to submit them from in-repo branches, I see no problem with that.

mrclary:
So I'm concerned about three potential issues:
- Code signing breaks the application
- Code signing fails
- Notarization fails

I'll move the existing application test to occur after the code signing step. For standard PRs, this won't change anything, since the code signing step won't run. On releases, however, the test will be performed on the code-signed application. This should address the first issue above.

For all the issues, though, the release PR could fail. Is that okay? Meaning that we could discover the failure, fix the failure, and just re-attempt the release? Or does this pose a workflow problem?

ccordoba12:
> Meaning that we could discover the failure, fix the failure, and just re-attempt the release? Or does this pose a workflow problem?

We can't re-attempt a release, i.e. a new release would need to be made to fix the problem. That's why I'd suggest to do beta releases to test that things are working as expected.

mrclary:
From [your comment](https://github.com/spyder-ide/spyder/pull/18508#issuecomment-1171828924)
> That's great to hear! We could also make one or more beta releases to test that signing the app works on releases. I think that would be better than just hoping for the best.

I hope that my comment above addresses this. In short, I don't want to have to "release" something just to test that the code signing and notarization worked and didn't break the app. As long as the release PR is allowed to fail and resume after fixes, then I think it should be fine.

ccordoba12 (approved):
Thanks @mrclary for all the hard work you put into making this is a reality!

I'm sure our Mac app users will certainly appreciate.

PR #16505, opened 2021-09-29
Fix autoformat issue 16280 spyder

### LINKED ISSUE #16280: Spyder editor adds lines to code when pressing F-5
# Editor replicates lines of code at end of script
I have already written about this on stackoverflow (https://stackoverflow.com/questions/68472061/spyder-suddenly-copies-code-to-end-of-file), but at the time I could not reproduce the bug. Here is what I know today with an example to reproduce everything.

With certain script constellations, the last couple of lines of a script are replicated at the end of the script without user intervention when pressing F-5 (Run script). This can lead to a lot of time lost, e.g. when the main function is replicated at the end of a script. 

### Steps to Reproduce the Problem

1. Copy the following script to the spyder editor. Note that the code is not yet formatted and there are a couple of long lines towards then end.
```python
# -*- coding: utf-8 -*-
"""
# -*- coding: utf-8 -*-
"""
Created on Wed Aug 25 15:19:45 2021

@author: tobia
"""

import statistics
from edatest import Edatest


def get_quartiles(l, method):

    l = sorted(l)
    q0, q4 = min(l), max(l)
    q2 = statistics.median(l)
    if method == "MS":
        p1 = int((len(l) + 1) / 4 + 0.5)
        q1 = l[p1 - 1]
        p1 = int(3 * (len(l) + 1) / 4) - 1
        q3 = l[p1]
    else:
        len_odd = len(l) % 2 == 1
        mid_point = len(l) // 2

        left = l[0:mid_point]
        if len_odd:
            right = l[mid_point + 1 :]
        else:
            right = l[mid_point:]

        if (method == "T") and (len(l) % 2 == 1):  # odd number of elements
            left.append(q2)
            right = [q2] + right

        # print(f"left: {left}, right: {right}")
        q1 = statistics.median(left)
        q3 = statistics.median(right)

    return [q0, q1, q2, q3, q4]


if __name__ == "__main__":
    Test = Edatest()
    Test.assert_equals(get_quartiles([4, 1, 7, 8, 3, 6, 5, 2], "T"), [1, 2.5, 4.5, 6.5, 8])
    Test.assert_equals(get_quartiles([4, 1, 7, 8, 3, 6, 5, 2], "MS"), [1, 2, 4.5, 7, 8])
    Test.assert_equals(get_quartiles([2, 6, 1, 8, 5, 9, 7, 4, 3], "MM"), [1, 2.5, 5, 7.5, 9])
    Test.assert_equals(
    get_quartiles([2, 6, 1, 8, 5, 9, 7, 4, 3], "MS"), [1, 3, 5, 7, 9]
    )
    Test.assert_equals(get_quartiles([10, 3, 1, 8, 6, 4, 7, 5, 2, 9], "T"), [1, 3, 5.5, 8, 10])
    Test.assert_equals(get_quartiles([10, 3, 1, 8, 6, 4, 7, 5, 2, 9], "MM"), [1, 3, 5.5, 8, 10])
    Test.assert_equals(get_quartiles([10, 3, 1, 8, 6, 4, 7, 5, 2, 9], "MS"), [1, 3, 5.5, 8, 10])
    Test.assert_equals(get_quartiles([3, 9, 11, 2, 4, 1, 8, 6, 10, 5, 7], "T"), [1, 3.5, 6, 8.5, 11])
    Test.assert_equals(get_quartiles([3, 9, 11, 2, 4, 1, 8, 6, 10, 5, 7], "MM"), [1, 3, 6, 9, 11])

``` 
2. Press F-5 -> the script is saved, formatted and executed
3. Press F-5 again: the script is again executed and terminates with an error message
```
File "C:\Users\tobia\OneDrive\Documents\Python Scripts\edabit\quartiles\quartiles.py", line 63
    )
    ^
SyntaxError: unmatched ')'
```
because someone added a couple of lines to the end of the script. In this case it is the following lines:
```python
        get_quartiles([10, 3, 1, 8, 6, 4, 7, 5, 2, 9], "T"), [1, 3, 5.5, 8, 10]
    )
    Test.assert_equals(
        get_quartiles([10, 3, 1, 8, 6, 4, 7, 5, 2, 9], "MM"), [1, 3, 5.5, 8, 10]
    )
    Test.assert_equals(
        get_quartiles([10, 3, 1, 8, 6, 4, 7, 5, 2, 9], "MS"), [1, 3, 5.5, 8, 10]
    )
    Test.assert_equals(
        get_quartiles([3, 9, 11, 2, 4, 1, 8, 6, 10, 5, 7], "T"), [1, 3.5, 6, 8.5, 11]
    )
    Test.assert_equals(
        get_quartiles([3, 9, 11, 2, 4, 1, 8, 6, 10, 5, 7], "MM"), [1, 3, 6, 9, 11]
    )
```
which are the ~15 last lines of the script, which were just copied to the end.
4. You can go back by pressing Ctrl-Z twice, and try again from there.

It is interesting to note that one can observe this behaviour in real-time when watching the scroll bar at the right side of the window. Its size changes twice, presumably the first time when the code is reformatted, and then again, when the last couple of lines are replicated.
## Versions

Spyder version: 5.0.5
Python 3.8.10 64-bit | Qt 5.9.7 | PyQt5 5.9.2 | Windows 10 

I have observed this behaviour also in previous versions of spyder, but could only reproduce it now.

### Dependencies

For the script above to run edatest needs to be installed (https://pypi.org/project/edatest/), however this is not central to the bug, only required to run the script.

rear1019:
## Description of Changes
Fix autoformat on save when running a script: Autoformat may replicate lines of code when autosave before running a script is triggered. See commit message for details.

### Issue(s) Resolved

Fixes #16280.

### Affirmation

rear1019

### DIFF
--- a/external-deps/spyder-kernels/.gitrepo
+++ b/external-deps/spyder-kernels/.gitrepo
@@ -5,8 +5,8 @@
 ;
 [subrepo]
 	remote = https://github.com/spyder-ide/spyder-kernels.git
-	branch = 2.x
-	commit = d151e3b4f01a1284487b6183def17161307f8bb5
-	parent = eb1cd5b7fe9ddb77a1869512c129d89c93484a46
+	branch = fix_autoformat_issue_16280_kernel
+	commit = 15b66780db9f26d8db0df558d6c7fbbd7ba621fe
+	parent = fba6e064f630cb42db3b519509a984d87e3dbb6c
 	method = merge
 	cmdver = 0.4.3
--- a/external-deps/spyder-kernels/CHANGELOG.md
+++ b/external-deps/spyder-kernels/CHANGELOG.md
@@ -1,7 +1,19 @@
 # History of changes
 
-## Version 2.1.1 (2021-09-01)
+## Version 2.1.2 (2021-09-28)
 
+### Pull Requests Merged
+
+* [PR 323](https://github.com/spyder-ide/spyder-kernels/pull/323) - PR: Add `ipython_genutils` dependency for testing, by [@ccordoba12](https://github.com/ccordoba12)
+* [PR 322](https://github.com/spyder-ide/spyder-kernels/pull/322) - PR: Prevent other libraries to change the breakpoint builtin, by [@ccordoba12](https://github.com/ccordoba12)
+
+In this release 2 pull requests were closed.
+
+
+----
+
+
+## Version 2.1.1 (2021-09-01)
 
 ### Pull Requests Merged
 
--- a/external-deps/spyder-kernels/requirements/tests.txt
+++ b/external-deps/spyder-kernels/requirements/tests.txt
@@ -11,3 +11,5 @@ pytest-cov
 scipy
 xarray
 pillow
+# Remove when Anaconda updates to a newer ipykernel
+ipython_genutils
--- a/external-deps/spyder-kernels/spyder_kernels/customize/spydercustomize.py
+++ b/external-deps/spyder-kernels/spyder_kernels/customize/spydercustomize.py
@@ -481,12 +481,11 @@ def exec_code(code, filename, ns_globals, ns_locals=None, post_mortem=False):
     __tracebackhide__ = "__pdb_exit__"
 
 
-def get_file_code(filename, save_all=True):
+def get_file_code(filename):
     """Retrive the content of a file."""
     # Get code from spyder
     try:
-        file_code = frontend_request().get_file_code(
-            filename, save_all=save_all)
+        file_code = frontend_request().get_file_code(filename)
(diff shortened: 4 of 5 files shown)


ccordoba12 (commented):
I think this is incorrect because it removes functionality, i.e. being able to save all files in the editor from the kernel). Instead, I think if that option is enabled on the Spyder side, autoformat should be run but we shouldn't try to save files (even if the option is activated), which is what is generating the problem.

@impact27, what do you think?

impact27:
The problem here is that two formatting requests sent in quick succession should not mess up the document. The problem is not that we want to save twice. I am not sure what the best fix should be but maybe having a check that any changes requested by LSP should not come from a stale version of the document.

PR #16518, opened 2021-09-30
PR: Allow users to disable web widgets with a command line option

dmageeLANL:
## Description of Changes

Added an option to spyder.ini to disable the info_widget: ipython_console.info_widget = True (Default).  Added a command line option "--offline" to disable help, offline help, and the info_widget. This is an effective configuration for the HPC machines at Los Alamos National Laboratory (LANL) and, I suspect, other facilities with limited internet access and browser functionality.  

The interface launched with --offline will look like this (fewer bottom tabs on the explorer and console):

spyder_infolessGUI

The current spyder version crashes with the following error message on our systems.  These changes allow the user to avoid activating these features.  

``` bash
[18235:18235:0930/153240.489072:FATAL:zygote_host_impl_linux.cc(165)] Check failed: process.IsValid(). Failed to launch zygote process
#0 0x2b8a0df1f0be <unknown>
#1 0x2b8a0df316c2 <unknown>
#2 0x2b8a0dc9b8f8 <unknown>
#3 0x2b8a0dc9ab07 <unknown>
#4 0x2b8a0dc9ae96 <unknown>
#5 0x2b8a0da459a1 <unknown>
#6 0x2b8a0da49727 <unknown>
#7 0x2b8a0d8db542 <unknown>
#8 0x2b8a0d8db718 <unknown>
#9 0x2b8a0d8d2681 <unknown>
#10 0x2b8a0d8d2732 QtWebEngineCore::WebContentsAdapter::WebContentsAdapter()
#11 0x2b8a0d142376 <unknown>
#12 0x2b8a0d1400c4 <unknown>
#13 0x2b8a0d140252 QWebEnginePage::QWebEnginePage()
#14 0x2b8a0d0fa689 sipQWebEnginePage::sipQWebEnginePage()
#15 0x2b8a0d0fa812 init_type_QWebEnginePage
#16 0x2b8a0be1a874 <unknown>
#17 0x55d7233a51d4 _PyObject_MakeTpCall
#18 0x55d7234295e5 _PyEval_EvalFrameDefault
#19 0x55d7233f22a2 _PyEval_EvalCodeWithName
#20 0x55d7233f3d3f _PyObject_FastCallDict
#21 0x55d7233f4033 _PyObject_Call_Prepend
#22 0x55d7233f41ca slot_tp_init
#23 0x55d7233a51d4 _PyObject_MakeTpCall
#24 0x55d7234295e5 _PyEval_EvalFrameDefault
#25 0x55d7233f22a2 _PyEval_EvalCodeWithName
#26 0x55d7233f3243 _PyFunction_Vectorcall.localalias.353
#27 0x55d7233f3dcb _PyObject_FastCallDict
#28 0x55d7233f4033 _PyObject_Call_Prepend
#29 0x55d7233f41ca slot_tp_init
#30 0x55d7233a51d4 _PyObject_MakeTpCall
#31 0x55d72342a341 _PyEval_EvalFrameDefault
#32 0x55d7233f27af _PyEval_EvalCodeWithName
#33 0x55d7233f3620 method_vectorcall
#34 0x55d72336758e _PyEval_EvalFrameDefault.cold.2793
#35 0x55d7233f316b _PyFunction_Vectorcall.localalias.353
#36 0x55d723367819 _PyEval_EvalFrameDefault.cold.2793
#37 0x55d7233f316b _PyFunction_Vectorcall.localalias.353
#38 0x55d72336758e _PyEval_EvalFrameDefault.cold.2793
#39 0x55d7233f316b _PyFunction_Vectorcall.localalias.353
#40 0x55d723367819 _PyEval_EvalFrameDefault.cold.2793
#41 0x55d7233f22a2 _PyEval_EvalCodeWithName
#42 0x55d7233f3054 PyEval_EvalCodeEx
#43 0x55d7234815bc PyEval_EvalCode
#44 0x55d723481664 run_eval_code_obj
#45 0x55d7234b3874 run_mod
#46 0x55d72337c316 PyRun_FileExFlags
#47 0x55d72337c6fb PyRun_SimpleFileExFlags
#48 0x55d72337d5c0 Py_RunMain.cold.2912
#49 0x55d7234b6389 Py_BytesMain
#50 0x2b89f1905555 __libc_start_main
#51 0x55d723446553 <unknown>
```

Minor note: I changed "don't" to "do not" in the internal_console message because the apostrophe messes up coloration for spyder.ini in vscode.

### Issue(s) Resolved

Fixes #

### Affirmation


spyder/app/cli_options.py
```diff
@@ -135,6 +135,13 @@ def get_options(argv=None):
         default=False,
         help="Start Spyder with a clean configuration directory"
     )
+    parser.add_argument(
+        '--offline',
+        dest="offline",
```
ccordoba12, line 140:
```suggestion
        '--no-web-widgets',
        dest="no-web-widgets",
```

Spyder can work offline, so this is misleading. Instead, I suggest this change to make things more explicit.

spyder/app/cli_options.py
```diff
@@ -135,6 +135,13 @@ def get_options(argv=None):
...
+        action='store_true',
+        default=False,
+        help="Disable all web engine and download functionality."
```
ccordoba12, line 143:
```suggestion
        help="Disable the usage of web widgets in Spyder (e.g. the Help and "
             "Online help panes)."
```

spyder/app/mainwindow.py
```diff
@@ -1975,6 +1977,11 @@ def main(options, args):
             pass
     CONF.set('main', 'previous_crash', previous_crash)
 
+    if options.offline:
+        CONF.set('ipython_console', 'info_widget', False)
+        CONF.set('help', 'enable', False)
+        CONF.set('onlinehelp', 'enable', False)
```
ccordoba12, line 1983: There's a missing `else` here to set these options to `True`. Else, these elements would be hidden until users reset their preferences.

spyder/config/manager.py
```diff
@@ -44,8 +44,6 @@ def __init__(self, parent=None, active_project_callback=None):
         Configuration manager to provide access to user/site/project config.
         """
         path = self.get_user_config_path()
-        if not osp.isdir(path):
-            os.makedirs(path)
```
ccordoba12, line 48: Why this change? This has nothing to do with this PR, so it should be reverted and discussed in another PR.

spyder/plugins/ipythonconsole/widgets/client.py
```diff
@@ -785,7 +788,7 @@ def set_elapsed_time_visible(self, state):
 
     def set_info_page(self):
         """Set current info_page."""
-        if self.info_page is not None:
+        if self.infowidget is not None:
```
ccordoba12, line 791:
```suggestion
        if self.infowidget is not None and self.info_page is not None:
```

I think the logic set here requires that.

ccordoba12 (commented):
Hey @dmageeLANL, thanks a lot for your contribution! I think this is a really good idea.

Please see my comments below.

spyder/app/cli_options.py
dmageeLANL: I agree, this is a better flag.

spyder/app/cli_options.py
dmageeLANL: 👍

spyder/app/mainwindow.py
dmageeLANL: Yes this was something I was unsure about.   Because of how the configuration files are parsed, if you set these configuration options,  the spyder.ini values are ignored and overwritten.  

So this could cause confusion if a user sets options in spyder.ini and sees them ineffective and overwritten at runtime.  For example if you set help.enable = False in spyder.ini but then run without '--no-web-widgets', if we add an `else: #set to true` phrase your spyder.ini settings won't be honored, without the else it honors the command line, then the spyder.ini.  But in both cases it'll overwrite your spyder.ini values with true or false depending on the flag setting.  

It seems to me that the best thing would be to have the flag overwrite these settings but not write to spyder.ini.  So that to disable these features more permanently you'd need to edit spyder.ini, but disable them for a single run you'd just need to use the flag.  I couldn't find if it was possible to make that happen.

In any case, I was hoping you guys had a canonical way of dealing with this type of issue so whatever you think is best is what it should do.

spyder/config/manager.py
dmageeLANL: Oh, yeah I just noticed that was repeated because the method called in the line before it does the same operation.

spyder/plugins/ipythonconsole/widgets/client.py
dmageeLANL: Makes sense.

spyder/config/manager.py
dmageeLANL: I must have forgotten to submit the comment last time.  I noticed that the method in the previous line performs this exact operation, so it's repeated.

ccordoba12:
Hey @dmageeLANL, I'll give you a hand to finish this PR so we can include it in our next version (5.3.1), to be released soon.

ccordoba12 (approved):
Thanks a lot for your contribution @dmageeLANL!

PR #16586, opened 2021-10-14
PR: Added support for sphinx_math_dollar in help text generation

oscargus:
## Description of Changes

Added support for the `sphinx_math_dollar` plugin that e.g. SymPy uses.

https://www.sympy.org/sphinx-math-dollar/

The idea is that math equations can be marked using the LaTeX-style `$...$` or `$$...$$` instead of the more tedious Sphinz marking.

With:
image

Without:
image

This checks if `sphinx_math_dollar` exists and if so adds it to the sphinx configuration (assuming that MathJax is used).

Not clear how to add a test for it. It seems like there are no tests for math rendering in the help plugin at all.

If it makes sense, I can add an additional check box for it in the preferences. But I think that if someone has installed sphinx_math_dollar (it is not a mandatory dependency of SymPy), they probably would like to use it by default.

### Affirmation

oscargus, Oscar Gustafsson

### DIFF
--- a/installers/Windows/installer.py
+++ b/installers/Windows/installer.py
@@ -41,7 +41,7 @@ import yarg
 ASSETS_URL = os.environ.get(
     'ASSETS_URL',
     'https://github.com/spyder-ide/windows-installer-assets/'
-    'releases/download/0.0.1/assets.zip')
+    'releases/download/0.0.4/assets.zip')
 
 # Packages to remove from the requirements for example pip or
 # external direct dependencies (python-lsp-server spyder-kernels)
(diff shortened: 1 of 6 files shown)


oscargus:
Finally got around to updating this. Took a bit of time as `sphinx-math-dollar` does not have a wheel and although the solution wasn't that tricky in the end it wasn't straightforward for me...

I'd say that the logic is a bit more correct now since earlier on packages in SKIP_PACKAGES was not correctly handled as modules (as the set subtraction was on strings with version numbers). Although it doesn't really makes any difference, it should be better now.

Also, I added a feature so that all packages with `-` in them are replaced with `_` when the modules are asked for. This is probably the best guess one can make.

The MacOS-app failure seems unrelated and maybe I should have merged against `5.x` instead?

PR #16711, opened 2021-11-02
PR: UX improvements to panes that inherit from OneColumnTree widget

ccordoba12:
## Description of Changes

* The panes that use that widget are Outline, Find and Code Analysis.
* Change mouse shape to pointing hand when hovering over tree elements

    pointing

* Collapse tree roots with a single click

    collapse

* Show horizontal scrollbar if content is too wide

    scrollbar

* Showing a background color when hovering entries in Find (before we were not showing any color).

    hover

### Affirmation

### DIFF
--- a/.github/scripts/modules_test.sh
+++ b/.github/scripts/modules_test.sh
@@ -88,6 +88,9 @@ for f in spyder/*/*/*.py; do
     if [[ $f == spyder/api/plugins/*.py ]]; then
         continue
     fi
+    if [[ $f == spyder/plugins/pylint/main_widget.py ]]; then
+        continue
+    fi
     python "$f"
     if [ $? -ne 0 ]; then
         exit 1
--- a/spyder/plugins/findinfiles/widgets/main_widget.py
+++ b/spyder/plugins/findinfiles/widgets/main_widget.py
@@ -583,7 +583,8 @@ class FindInFilesWidget(PluginMainWidget):
 
         # Setup result_browser
         self.result_browser.set_path(options[0])
-        self.result_browser.longest_item = ''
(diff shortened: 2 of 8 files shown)


isabela-pf:
Thanks for bringing this up, @ccordoba12. Here's a summary of what we talked about in person below:

- Most of this makes sense! Adding more consistency to these panes that have similar behaviors is a good thing.
- The multi-colored selection in the Find pane looks like a mistake and could be confusing since the two different colors have two different meanings elsewhere in the interface (gray for the hover, blue for the selection). Since it will be difficult to fix and there are other things we'd like to merge in this PR @steff456 and @CAM-Gerlach suggest reverting commits for this change so we can revisit it without blocking these changes.
- We'd also like the Outline pane to match the behavior of the other panes and have a single click to open it content.

ccordoba12:
Thanks everyone for the review! My comments:

> The multi-colored selection in the Find pane looks like a mistake and could be confusing since the two different colors have two different meanings elsewhere in the interface

I reverted this change.

> We'd also like the Outline pane to match the behavior of the other panes and have a single click to open it content.

I took a look at it and there are several UX issues I think we need to be careful about. So I'd like to leave it for another PR in order to merge this one for 5.2.0. Is that ok?

isabela-pf:
Sounds reasonable. Thanks for the update and giving it a try!

isabela-pf (approved):

CAM-Gerlach (approved):
I did some functional testing of each pane and the discused changes, and everything LGTM. Thanks @ccordoba12 !

ccordoba12:
Thanks @isabela-pf and @CAM-Gerlach for the review! Merging now.

PR #16409, opened 2021-09-12
PR: Update PyQt requirement to 5.15

ccordoba12:
## Description of Changes

- This is finally the time to do it because we're going to have PyQt 5.15 packages available in conda-forge, thanks to @andfoy's work.
- I expect this to take some time because our test suite is big.

### Issue(s) Resolved

Fixes #17085
Fixes #17083
Fixes #17060 
Fixes #12829

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct: @ccordoba12

PR #16848, opened 2021-11-17
PR: Prevent some plugins from being disabled

steff456:
## Description of Changes

This PR prevents the following plugins from being disabled: 

- Layout
- Main interpreter
- Main menu
- Preferences
- Shortcuts
- Status bar
- Working directory
- Toolbar

The decision was mainly because these set of plugins shouldn't be disabled since they form part of the core of the application.

### Affirmation

### DIFF
--- a/spyder/plugins/completion/confpage.py
+++ b/spyder/plugins/completion/confpage.py
@@ -22,18 +22,6 @@ class CompletionConfigPage(PluginConfigPage):
     def setup_page(self):
         newcb = self.create_checkbox
 
-        # -------------------- Plugin status group -----------------------------
-        plugin_state_group = QGroupBox(_('Plugin status'))
-        plugin_state_layout = QVBoxLayout()
-        self.completion_box = newcb(
-            _("Enable code completion and linting in the editor"),
-            'enable'
-        )
-        plugin_state_layout.addWidget(self.completion_box)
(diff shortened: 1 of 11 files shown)


ccordoba12 (commented):
Thanks @steff456 for this! My comments:

* Why Shortcuts, the Working directory, Status bar and Toolbar plugins can't be disabled? I disable them successfully on my tests on #16409.
* Please don't allow the Internal console to be disabled. That's required because it's the default plugin against which any dockable plugin is docked in case it doesn't declare where it wants to be placed.
* As part of this PR, please remove the option to disable `Completion and linting` from its preferences page:

    imagen

* If we agree that the Status bar can be disabled, this option needs to be removed too

    imagen

steff456:
Given our meeting, we decided to avoid disabling Shortcuts, Working directory and Toolbar plugins because it may not have sense to the Spyder interface. 

The last two commits:
- Added the internal console and plugin that cannot be disabled
- Preferences options were removed from the completion and linting page. 
- Statusbar plugin can be disabled in order to remove the `show status bar` option from the Statusbar's preferences page.

ccordoba12 (commented):
Thanks @steff456! Looks good to me now.

Please merge with `5.x` to get the latest fixes to our tests, then this should be ready.

ccordoba12 (approved):
Thanks @steff456!

PR #16855, opened 2021-11-18
PR: Update missing translation strings for next release

dalthviz:
## Description of Changes

Update missing translation strings after https://github.com/spyder-ide/spyder/pull/16847

## NOTE

Running `gettext-helpers` the following message appears:

```
...spyder/spyder/plugins/projects/widgets/projectdialog.py:74: Seen unexpected token "f"Select a new or existing directory to create a new Spyder ""
```

~Not sure if there is something I should do~ Changed to use `format` instead of `f-strings`

### Affirmation

### DIFF
--- a/spyder/locale/de/LC_MESSAGES/spyder.po
+++ b/spyder/locale/de/LC_MESSAGES/spyder.po
@@ -2,7 +2,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: spyder\n"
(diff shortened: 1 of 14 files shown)


ccordoba12:
> Not sure if there is something I should do

Yes, please change that string to use `format` instead of an f-string because otherwise the `po` files won't be generated correctly.

ccordoba12 (approved):
Thanks @dalthviz!

Note: The failure in our tests is unrelated to this and I fixed it in PR #16856.

PR #16974, opened 2021-12-07
PR: Limit the number of flags in the editor

impact27:
## Description of Changes
If there are too many flags in the document, the flags will slow down the printing. This disables the flags of a type if there are more than 10000.

To reproduce, create a file with:
```
with open("test.py", "w") as f:
    for i in range(10000):
        f.write("aaaaaa\n")
```
This will create 10000 error flags, which will slow down spyder significantly

### Issue(s) Resolved

Fixes #

### Affirmation

### DIFF
--- a/spyder/plugins/editor/panels/scrollflag.py
+++ b/spyder/plugins/editor/panels/scrollflag.py
@@ -23,6 +23,7 @@ from spyder.plugins.completion.api import DiagnosticSeverity
 
 
 REFRESH_RATE = 1000
+
 # Maximum number of flags to print in a file
 MAX_FLAGS = 1000
 
@@ -220,10 +221,11 @@ class ScrollFlagArea(Panel):
                         continue
                     geometry = editor.blockBoundingGeometry(block)
                     rect_y = ceil(
-                        geometry.y()
-                        + geometry.height() / 2 + rect_h / 2)
+                        geometry.y() +
+                        geometry.height() / 2 +
+                        rect_h / 2
+                    )
                     painter.drawRect(rect_x, rect_y, rect_w, rect_h)
-
             elif last_line == 0:
                 # Only one line
                 for block in dict_flag_lists[flag_type]:
--- a/spyder/plugins/editor/panels/scrollflag.py
+++ b/spyder/plugins/editor/panels/scrollflag.py
@@ -23,6 +23,8 @@ from spyder.plugins.completion.api import DiagnosticSeverity
 
 
 REFRESH_RATE = 1000
+# Maximum number of flags to print in a file
+MAX_FLAGS = 1000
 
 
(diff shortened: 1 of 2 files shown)


spyder/plugins/editor/panels/scrollflag.py
```diff
@@ -23,6 +23,8 @@
 
 
 REFRESH_RATE = 1000
+# Maximum number of flags to print in a file
```
ccordoba12, line 26:
```suggestion

# Maximum number of flags to print in a file
```

spyder/plugins/editor/panels/scrollflag.py
```diff
@@ -216,14 +213,41 @@ def compute_flag_ypos(block):
...
-                painter.drawRect(rect_x, rect_y, rect_w, rect_h)
+            if editor.verticalScrollBar().maximum() == 0:
+                # No scroll
+                for block in dict_flag_lists[flag_type]:
+                    if not block.isValid():
+                        continue
+                    geometry = editor.blockBoundingGeometry(block)
+                    rect_y = ceil(
+                        geometry.y()
+                        + geometry.height() / 2 + rect_h / 2)
```
ccordoba12, line 224:
```suggestion
                    rect_y = ceil(
                        geometry.y() +
                        geometry.height() / 2 +
                        rect_h / 2
                    )
```

spyder/plugins/editor/panels/scrollflag.py
```diff
@@ -216,14 +213,41 @@ def compute_flag_ypos(block):
...
+                    painter.drawRect(rect_x, rect_y, rect_w, rect_h)
+
```
ccordoba12, line 226:
```suggestion
```

ccordoba12 (commented):
Thanks @impact27 for your help with this! I left some minor style suggestions, otherwise looks good to me.

ccordoba12:
@impact27, one thing I forgot to mention: could you add a test for this? Just to check that we're not trying to display more than `MAX_FLAGS` in the scrollflag area.

impact27:
I am not sure how to do that. the paint code does a bunch of "painter.drawRect" but I think the result is just an image. I wouldn't know how to check how many rectangles are drawn.

ccordoba12:
What if you check the length of `self._dict_flag_list[flag_type]` for a certain flag type that goes over `MAX_FLAGS` in a file?

impact27:
The filtering is applied at the print stage, so the dictionary would contain more than the limit of elements

ccordoba12:
Ok, no problem then. Creating a test for this is way harder than I thought.

ccordoba12 (approved):
Thanks @impact27!

PR #16991, opened 2021-12-09
PR: Fix changing color scheme and UI theme (Appearance)

ccordoba12:
## Description of Changes

- Show a restart message whenever there is a mismatch between the selected UI theme and color scheme in preferences. That functionality broke in #16012.
- Fix observers that need to react to changes in color scheme. This was not working as expected for all plugins that need it (i.e. editor, help, history and IPython console).
- Add tests for the Appearance config page so its UX to change color schemes and UI themes doesn't break yet again (that UX is tricky to get right).
- In Preferences, don't add `section` to changed and restart options when it is `None`. This allow us to send the set of changed options to plugins in the same way as before #16012 was merged (not only Appearance but several other things were broken due to that).
- Cleanup unused imports.

### Affirmation

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -16,7 +16,6 @@ import os.path as osp
 import shutil
 import subprocess
 import sys
-import warnings
 
 
 # ---- To activate/deactivate certain things for pytest's only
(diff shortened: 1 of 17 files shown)


dalthviz (approved):
Thanks @ccordoba12 ! LGTM :+1:

PR #17022, opened 2021-12-14
one liner - Make Variable Explorer column headers movable #16423

battaglia01:
## Description of Changes
This implements the one-line change from #16423, which makes Variable Explorer column headers movable. The one line is at collectionseditor.py:617:

```
    def setup_table(self):
        """Setup table"""
        self.horizontalHeader().setStretchLastSection(True)
        self.horizontalHeader().setSectionsMovable(True) # <-- NEW LINE
        self.adjust_columns()
        # Sorting columns
        self.setSortingEnabled(True)
        self.sortByColumn(0, Qt.AscendingOrder)
```

* [ ] Wrote at least one-line docstrings (for any new functions)
* [ ] Added unit test(s) covering the changes (if testable)

* [ ] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

### Issue(s) Resolved

Fixes #16423

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct:
Mike Battaglia

PR #17034, opened 2021-12-16
PR: Make Variable Explorer column headers movable

### LINKED ISSUE #16423: Make Variable Explorer column headers movable
The variable explorer currently has the columns as Name -> Type -> Size -> Value, which is kind of annoying as the main thing I usually care about is Value which is way off at the end. There is no way to change this order but a simple one-liner in Qt could make the column headers draggable. In `collectionseditor.py`, in class `BaseTableView`'s `def setup_table(self)`, simply add the one line

```
    def setup_table(self):
        """Setup table"""
        self.horizontalHeader().setStretchLastSection(True)
        self.horizontalHeader().setSectionsMovable(True) # <--- NEW LINE HERE
        self.adjust_columns()
        # Sorting columns
        self.setSortingEnabled(True)
        self.sortByColumn(0, Qt.AscendingOrder)
```

and you're done. Note this is for BaseTableView; you could also do this instead in RemoteCollectionsEditorTableView and CollectionsEditorTableView (not sure which subclass is the best to do this in, or what other things subclass BaseTableView).

FWIW, I'd much prefer it if the default were changed to Name -> Value -> Type -> Size, but that seems more involved and I'd be happy enough with this.

## Versions

Spyder: 5.0.5 | Python 3.9.5 64-bit | Qt 5.9.7 | PyQt5 5.9.2 | Darwin 17.7.0

battaglia01:
### Issue(s) Resolved

Fixes #16423. See also #17022, for which this is an identical PR but on a different branch as requested.

### Affirmation

I certify the above statement is true and correct: Mike Battaglia

### DIFF
--- a/spyder/widgets/collectionseditor.py
+++ b/spyder/widgets/collectionseditor.py
@@ -614,6 +614,7 @@ class BaseTableView(QTableView, SpyderConfigurationAccessor):
     def setup_table(self):
         """Setup table"""
         self.horizontalHeader().setStretchLastSection(True)
+        self.horizontalHeader().setSectionsMovable(True)
         self.adjust_columns()
         # Sorting columns
         self.setSortingEnabled(True)


ccordoba12:
@battaglia01, one of our tests is failing due to this change. Do you want me to help you with that?

battaglia01:
I'm not sure what the problem is; I literally only added that one line. What is failing?

battaglia01:
I looked at this again; it looks like something isn't sorting correctly. I haven't seen any such sorting issue on the patched version of Spyder I am running. Do we have any insight into what is causing this and what, in particular, isn't being sorted correctly?

ccordoba12:
I'll take a look at the test to see what's causing the failure.

ccordoba12:
Hey @battaglia01, I discovered the issue: we had to use a pytest fixture for of our Variable Explorer tests. Then the failing test pass without problems.

Since you didn't create a branch for your pull request, I had to create a new one to push the necessary changes (see #17127).

Thanks again for your contribution!

ccordoba12:
Closing in favor of #17127.

PR #17035, opened 2021-12-16
PR: Fix shutdown kernels associated to conda envs (IPython console)

### LINKED ISSUE #14739: ZMQError: Address already in use when restarting the kernel
I have installed Anaconda (python 3.7) in my ubuntu 20.04 system and I am using Spyder.  I am running CNN and while runs for some part of data it's fine but when I run it for entire data (more that 100000) I am getting the following error:

```an error occurred while starting the kernel```

Steps that reproduce the problem:

1.  run the code
2. running for about 20 minutes 
3. Spyder tries to restart the kernel
4. error raised

```python-traceback

File "/home/exx/anaconda3/lib/python3.7/runpy.py", line 193, in _run_module_as_main
"__main__", mod_spec)
File "/home/exx/anaconda3/lib/python3.7/runpy.py", line 85, in _run_code
exec(code, run_globals)
File "/home/exx/anaconda3/lib/python3.7/site‑packages/spyder_kernels/console/__main__.py", line 23, in 
start.main()
File "/home/exx/anaconda3/lib/python3.7/site‑packages/spyder_kernels/console/start.py", line 284, in main
kernel.initialize()
File "/home/exx/anaconda3/lib/python3.7/site‑packages/traitlets/config/application.py", line 87, in inner
return method(app, *args, **kwargs)
File "/home/exx/anaconda3/lib/python3.7/site‑packages/ipykernel/kernelapp.py", line 567, in initialize
self.init_sockets()
File "/home/exx/anaconda3/lib/python3.7/site‑packages/ipykernel/kernelapp.py", line 271, in init_sockets
self.shell_port = self._bind_socket(self.shell_socket, self.shell_port)
File "/home/exx/anaconda3/lib/python3.7/site‑packages/ipykernel/kernelapp.py", line 218, in _bind_socket
return self._try_bind_socket(s, port)
File "/home/exx/anaconda3/lib/python3.7/site‑packages/ipykernel/kernelapp.py", line 194, in _try_bind_socket
s.bind("tcp://%s:%i" % (self.ip, port))
File "/home/exx/anaconda3/lib/python3.7/site‑packages/zmq/sugar/socket.py", line 173, in bind
super().bind(addr)
File "zmq/backend/cython/socket.pyx", line 542, in zmq.backend.cython.socket.Socket.bind
File "zmq/backend/cython/checkrc.pxd", line 28, in zmq.backend.cython.checkrc._check_rc
zmq.error.ZMQError: Address already in use

```
 Versions:

* Spyder 4.2.1: 
* Python 3.7.9: 
* Qt 5.15.2: 
* PyQt 5.15.2: 
* Linux/5.4.0.65-generic: 
* Ubuntu 20.04

Dependencies:
```
# Mandatory:
pygments >=2.0                  :  2.7.4 (OK)
pylint >=1.0                    :  2.6.0 (OK)
pyls_spyder >=0.3.0             :  0.3.0 (OK)
qtawesome >=0.5.7               :  1.0.1 (OK)
rtree >=0.8.3                   :  0.9.4 (OK)
setuptools >=39.0.0             :  52.0.0.post20210125 (OK)
sphinx >=0.6.6                  :  3.4.3 (OK)
spyder_kernels >=1.10.1;<1.11.0 :  1.10.1 (OK)
watchdog >=0.10.3               :  1.0.2 (OK)
xdg >=0.26                      :  0.27 (OK)

# Optional:
cython >=0.21                   :  0.29.21 (OK)
matplotlib >=2.0.0              :  3.3.0 (OK)
numpy >=1.7                     :  1.19.2 (OK)
pandas >=1.1.1                  :  1.1.0 (NOK)
scipy >=0.17.0                  :  1.5.2 (OK)
sympy >=0.7.3                   :  1.6.1 (OK)
```

### LINKED ISSUE #17011: Kernel process lingers after closing tab when running different interpreter and Qt5 backend
### Issue Report Checklist

    * [x] Restarted Spyder

## Problem Description
When closing the console using the cross on the tab (ctrl-dot often does not work), the python process is not killed, but lingers around, until Spyder gets restarted. This issue only occurs when you use the Qt5 graphics backend **and** choose a different interpreter from the default Spyder one. 

Both by setting the graphics interpreter to *inline* or installing Spyder in the same environment will alleviate the problem (but is not a solution).

This could (for instance) result into issues when hardware connections are made, as the addresses are not freed up, and reconnection is not possible until Spyder is restarted.

1. Set Qt5 graphics backend in Spyder.
2. Start a kernel in a non-default interpreter (different conda env).
3. Restart kernel by clicking the cross.

Expected: underlying python thread is killed. 
Instead: it stays alive

A way to visualise this, would be to initiate a Tk window. You'll see it stays alive after restarting the kernel. NOTE: this is NOT a prerequisite to have the issue occur, just a way to illustrate it.

None

## Versions

* Spyder version: 5.2.0 (confirmed issue is also there in 5.1.5)
    * Tried both the standalone version and the conda version. No difference.
* Python version: 3.9.7

ccordoba12:
## Description of Changes

- That was broken since jupyter_client **6.1.12**.
- That made the kernel process to linger around after creating plots with the Automatic backend and then killing it (seems to happen only on Windows).
- Constraint IPython to be less than 8 to avoid headaches when that version is released.
- Depends on spyder-ide/spyder-kernels#356
- Depends on spyder-ide/spyder-kernels#357

### Issue(s) Resolved

Fixes #17011 
Fixes #14739 

### Affirmation


ccordoba12:
@impact27, I think this change makes unnecessary the shutdown thread you added to avoid freezes (since the kernel is killed immediately, Spyder is not blocked by having to wait for it). Could you test that?

Of course, that's for another PR, but I'm curious about it.

impact27:
The problem with that is that we don’t give any time to the kernel to try to shut down. The goal of my previous change was to give a little time (10s) to the kernel to shut down before killing it.

impact27:
That is what ipykernel is doing. check e.g. `_async_finish_shutdown`. I suspect something is wrong with `SpyderKernelManager`

ccordoba12:
As far as I can tell, there are two ways to kill a kernel:

1. Call `_kill_kernel` (when `now` is passed to the `shutdown_kernel` method)
2. Call `finish_shutdown`, as you said.

`finish_shutdown` in turn calls `_send_kernel_sigterm` here:

imagen

which ends up calling the `terminate` method of the `subprocess` associated to the kernel:

imagen

imagen

The problem with that is that, on Windows and after creating a plot with Automatic backend, `terminate` kills the [batch script](https://github.com/spyder-ide/spyder/blob/5.x/spyder/plugins/ipythonconsole/scripts/conda-activate.bat) we use to activate the conda environment associated to the kernel, but not the kernel itself due to an AccessDenied error.

So, in the `SpyderKernelManager` we use `psutil` to kill all child processes associated to the activation script:

https://github.com/spyder-ide/spyder/blob/d49b92a1c719ed8509baa8f87cac754e084cc045/spyder/plugins/ipythonconsole/utils/manager.py#L58-L75

But for that, we need to use the `now` parameter in `shutdown_kernel`, so that we end up calling the `_kill_kernel` method we override in that class and not `finish_shutdown`.

Of course, another option would be to override `finish_shutdown` in our kernel manager and use `psutil` there instead of calling `terminate`. But I still don't understand why it's important to give the kernel some time to die, instead of killing it directly.

impact27:
If the kernel is killed directly, it doesn't have the opportunity to run any cleanup code (Any `__del__` methods for example). This is not a massive issue but could be problematic in some cases.

impact27:
Is there a reason why `psutil` is not implemented upstream in `jupyter_client`?

impact27:
Maybe we should override `_send_kernel_sigterm`?

ccordoba12:
> Maybe we should override _send_kernel_sigterm?

@impact27, I followed your advice in my latest commits. Please take a look at them and let me know what you think.

impact27:
This PR now reimplements `_async_send_kernel_sigterm`. I think it should also reimplement `_async_kill_kernel`. This is because both are used to shut down the kernel. (One to terminate, the other to kill).

We call `_async_kill_kernel` in spyder/plugins/ipythonconsole/widgets/shell.py:
`_kill_kernel()` is called in `wait_all_shutdown`

Furthermore, when `shutdown_kernel()` is called: (assuming jupyter_clients 6.0.2) in  jupyter_client/manager.py
`async def _async_shutdown_kernel`
 - calls `self.interrupt_kernel()`
 - if now=True: calls `self._kill_kernel()` (not our case)
 - else: request shutdown and then `self.finish_shutdown()`

So the interesting code path is finish_shutdown. The shutdown time is controlled by `self.shutdown_wait_time` in `async def _async_finish_shutdown`. First, it tries to terminate with:
 - `await self._async_send_kernel_sigterm()`
If that doesn't work, it kills with:
 - `await ensure_async(self._kill_kernel())`

ccordoba12:
> This PR now reimplements _async_send_kernel_sigterm. I think it should also reimplement _async_kill_kernel. This is because both are used to shut down the kernel. (One to terminate, the other to kill).

Ok, I restored the previous `_async_kill_kernel`. I removed it because I thought it was not necessary anymore, but you're right, it is.

impact27:
It looks good to me now. I wonder if killing rather than sending sigterm could produce issues but I can’t really
Think of any

ccordoba12:
> It looks good to me now

Great! Thanks for your help reviewing this one.

>  I wonder if killing rather than sending sigterm could produce issues but I can’t really
Think of any

Well, we send sigterm to the parent process and all its children here:

https://github.com/spyder-ide/spyder/blob/dab29a261c48b625b3296c4c5f9a438507797b1b/spyder/plugins/ipythonconsole/utils/manager.py#L62-L69

(`sig` is sigterm).

PR #17076, opened 2021-12-22
PR: Fix traitlets deprecation warning

### LINKED ISSUE #14534: DeprecationWarning: ShellWidget._syntax_style_changed is deprecated in traitlets 4.1: use @observe and @unobserve instead
I received the following warning when running the test suite:

```
spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py::test_banners
  <<<PACKAGEDIR>>>/spyder/plugins/ipythonconsole/widgets/shell.py:789: DeprecationWarning: ShellWidget._syntax_style_changed is deprecated in traitlets 4.1: use @observe and @unobserve instead.
    def _syntax_style_changed(self):

```

juliangilbey:
## Description of Changes

Fixed the traitlets DeprecationWarning raised by `spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py::test_banners`.  The unit test should no longer raise a warning after this change.

### Issue(s) Resolved

Fixes #14534 with thanks to @dalthviz 

### Affirmation

I affirm the [Developer Certificate of Origin](https://developercertificate.org),

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -241,6 +241,7 @@ install_requires = [
     'spyder-kernels>=2.2.0,<2.3.0',
     'textdistance>=4.2.0',
     'three-merge>=0.1.1',
+    'traitlets>=4.1',
     'watchdog>=0.10.3'
 ]
 
--- a/setup.py
+++ b/setup.py
@@ -241,7 +241,6 @@ install_requires = [
     'spyder-kernels>=2.2.0,<2.3.0',
     'textdistance>=4.2.0',
     'three-merge>=0.1.1',
-    'traitlets>=4.1',
     'watchdog>=0.10.3'
 ]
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -241,6 +241,7 @@ def run(self):
     'spyder-kernels>=2.2.0,<2.3.0',
     'textdistance>=4.2.0',
     'three-merge>=0.1.1',
+    'traitlets>=4.1',
```
ccordoba12, line 244:
```suggestion
```

`trailets` is a transitive dependency, so we don't need to depend on it directly:

    spyder -> spyder-kernels -> ipykernel -> traitlets

ccordoba12 (commented):
Thanks for your help with this @juliangilbey! I left a minor comment, otherwise looks good to me.

setup.py
juliangilbey: This is true, though the remainder of the patch does explicitly import it.  I guess it's a matter of style to a great extent; this explicit listing just makes the version requirements explicit, though ipykernel does that anyway.  I'm not particularly bothered either way; happy to make that change or let you do it.

ccordoba12: Please make it. We already have a lot of dependencies to add one more for such a simple change.

juliangilbey: Done!

ccordoba12 (approved):
Thanks @juliangilbey for your help with this!

PR #17077, opened 2021-12-22
PR: Put user provided paths at the front of sys.path

### LINKED ISSUE #17066: Suggestion: insert paths from PYTHONPATH manager before system's PYTHONPATH
## Problem Description

It would be convenient if the paths added to the PYTHONPATH were added before (or could be made to be added before) the system PYTHONPATH.

Use case: I want to run a package from source (for development reasons) that I already have installed through Conda or pip. Right now, on Windows, I have to add it to the system PYTHONPATH, which, sort of, removes the benefit of the Spyder PYTHONPATH manager.

I noted #16429 and maybe that solves the problem (by importing  and then moving the source to top), but maybe there can be an alternative solution as well, like a check box selecting if the context of the PYTHONPATH manager should be prepended or appended? I ping @mrclary directly here, to confirm that it can be done when/if #16429 is merged and here opinions on a possible implementation as part of that.

mrclary:
## Description of Changes

Put user provided paths at the front of `sys.path` for both command completions in the Editor (language server) and the IPython console.

Requires:
python-lsp/python-lsp-server#139
spyder-ide/spyder-kernels#352

### Issue(s) Resolved

Fixes #17066

### Affirmation


mrclary:
Closing in favor of #18308

PR #17089, opened 2021-12-24
PR: Print last line of Pdb code

impact27:
## Description of Changes

- When entering code in IPython, if the last line is an expression, it is printed after an "Out [i]: " prompt. This PR reproduces this behaviour in the debugger.
- Depends on spyder-ide/spyder-kernels#355

### Affirmation


spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -431,6 +431,20 @@ def set_pdb_state(self, pdb_state):
         if pdb_state is not None and isinstance(pdb_state, dict):
             self.refresh_from_pdb(pdb_state)
 
+    def pdb_out(self, text):
+        """Show pdb out."""
```
ccordoba12, line 435:
```suggestion
    def show_pdb_output(self, text):
        """Show Pdb output."""
```

Let's rename this method to make clearer what it does.

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -131,6 +131,7 @@ def __init__(self, ipyclient, additional_options, interpreter_versions,
         handlers.update({
             'pdb_state': self.set_pdb_state,
             'pdb_execute': self.pdb_execute,
+            'pdb_out': self.pdb_out,
```
ccordoba12, line 134:
```suggestion
            'show_pdb_output': self.show_pdb_output,
```

ccordoba12 (commented):
Thanks @impact27! I left some minor suggestions for you, otherwise looks good to me.

impact27:
@ccordoba12 I added a return value to runcell and runfile following the same logic

ccordoba12:
> I added a return value to runcell

Nice! I think that's a great addition!

> and runfile following the same logic

I'm not sure about that. I mean, `runfile` is our equivalent of `python myfile.py`, so I think we should preserve that you need to print something if you want to observe an output in that case.

impact27:
@ccordoba12 removed for runfile

Issue #17067, opened 2021-12-22
test_load_time fails


juliangilbey:
### Issue Report Checklist

* [x] Searched the [issues page](https://github.com/spyder-ide/spyder/issues?q=is%3Aissue) for similar reports
* [x] Read the relevant sections of the [Spyder Troubleshooting Guide](https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ) and followed its advice
* [x] Reproduced the issue after updating with ``conda update spyder`` (or ``pip``, if not using Anaconda)
* [ ] Could not reproduce inside ``jupyter qtconsole`` (if console-related)
* [x] Tried basic troubleshooting (if a bug/error)
    * [ ] Restarted Spyder
    * [ ] Reset preferences with ``spyder --reset``
    * [ ] Reinstalled the latest version of [Anaconda](https://www.anaconda.com/download/)
    * [ ] Tried the other applicable steps from the Troubleshooting Guide
* [x] Completed the **Problem Description**, **Steps to Reproduce** and **Version** sections below

## Problem Description

OK, now I'm running the tests with `CI=true`, I've only got 20(!) failing tests and one segfault....  Several of those seem to be because I have a different setup from your CI one, so I'm not worried about those.  Anyway, this is the next one after https://github.com/spyder-ide/spyder/issues/17059 and https://github.com/spyder-ide/spyder/issues/17060.  It is `test_load_time` in `spyder/plugins/maininterpreter/tests/test_confpage.py`, and I don't know what might be causing it.

### What steps reproduce the problem?

```
(sid):~/debian/spyder-packages/spyder/build-area/spyder-5.2.0+dfsg1 $ CI=true xvfb-run -a -s "-screen 0 1024x768x24 +extension GLX" python3.9 -m pytest spyder/plugins/maininterpreter/tests/test_confpage.py
============================= test session starts ==============================
platform linux -- Python 3.9.9, pytest-6.2.5, py-1.10.0, pluggy-0.13.0
PyQt5 5.15.6 -- Qt runtime 5.15.2 -- Qt compiled 5.15.2
rootdir: /home/jdg/debian/spyder-packages/spyder/build-area/spyder-5.2.0+dfsg1, configfile: pytest.ini
plugins: typeguard-2.11.1, flaky-3.7.0, xvfb-2.0.0, mock-3.6.1, cov-3.0.0, order-1.0.0, qt-4.0.2, lazy-fixture-0.6.3
[...]

PR #17095, opened 2021-12-26
PR: Skip conda and pyenv tests if these tools are not present

juliangilbey:
## Description of Changes

This patch skips the remaining tests which rely on the presence of pyenv or conda, in the same way that #17067 was fixed in PR #17092.

### Issue(s) Resolved

(I never reported this as an issue; I just skipped them on the Debian side.)

### Affirmation

### DIFF
--- a/spyder/plugins/maininterpreter/widgets/tests/test_status.py
+++ b/spyder/plugins/maininterpreter/widgets/tests/test_status.py
@@ -17,15 +17,12 @@ import pytest
 
 # Local imports
 from spyder.config.base import running_in_ci
+from spyder.config.utils import is_anaconda
 from spyder.plugins.statusbar.widgets.tests.test_status import status_bar
 from spyder.plugins.maininterpreter.widgets.status import InterpreterStatus
-from spyder.utils.conda import get_list_conda_envs
 
 
-CONDA_MISSING = (len(get_list_conda_envs()) == 0)
-
-
-@pytest.mark.skipif(CONDA_MISSING, reason="Requires conda to be installed")
+@pytest.mark.skipif(not is_anaconda(), reason="Requires conda to be installed")
 @pytest.mark.skipif(not running_in_ci(), reason="Only meant for CIs")
 def test_status_bar_conda_interpreter_status(status_bar, qtbot):
     """Test status bar message with conda interpreter."""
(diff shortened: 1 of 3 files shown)


spyder/plugins/maininterpreter/widgets/tests/test_status.py
```diff
@@ -19,8 +19,13 @@
 from spyder.config.base import running_in_ci
 from spyder.plugins.statusbar.widgets.tests.test_status import status_bar
 from spyder.plugins.maininterpreter.widgets.status import InterpreterStatus
+from spyder.utils.conda import get_list_conda_envs
 
 
+CONDA_MISSING = (len(get_list_conda_envs()) == 0)
```
ccordoba12, line 25: We have a function called `is_anaconda` in `spyder.config.utils`. So please use that one instead.

Also, there's no need to declare an extra constant here for that.

spyder/utils/tests/test_conda.py
```diff
@@ -21,6 +21,8 @@
     get_conda_root_prefix, get_list_conda_envs, get_list_conda_envs_cache)
 
 
+CONDA_MISSING = (len(get_list_conda_envs()) == 0)
```
ccordoba12, line 24: Please change this line to

```python
if not is_anaconda():
    pytest.skip()
```

and remove the `skipif` you added in each test below. That will allow us to skip all tests in this file.

spyder/utils/tests/test_pyenv.py
```diff
@@ -15,6 +15,10 @@
 from spyder.utils.pyenv import get_list_pyenv_envs, get_list_pyenv_envs_cache
 
 
+PYENV_MISSING = (len(get_list_pyenv_envs()) == 0)
```
ccordoba12, line 18: Please change this line to

```python
if not find_program('pyenv'):
    pytest.skip()
```

and remove the `skipif` you added in each test below.

`find_program` is available in `spyder.utils.programs`.

ccordoba12 (commented):
Thanks for your help with this @juliangilbey!

spyder/plugins/maininterpreter/widgets/tests/test_status.py
juliangilbey: Ah, I didn't know about that function, thanks!
There's a newer comment on [the stackoverflow thread](https://stackoverflow.com/a/47610844/3889398), by the way, which suggests that `is_anaconda` should use `sys.base_prefix` instead of `sys.prefix` for greater robustness.

spyder/utils/tests/test_conda.py
juliangilbey: Ah, I didn't know the `pytest.skip()` function; that's nice!

spyder/utils/tests/test_pyenv.py
juliangilbey: All three requests fulfilled!

ccordoba12 (approved):
Looks good to me now, thanks @juliangilbey!

PR #17157, opened 2022-01-08
PR: Fix error with PyNaCl (Windows installer)

ccordoba12:
## Description of Changes

- The problem is its latest release (1.5.0) doesn't have wheels per Python version. So we have to do some workarounds to avoid an error caused by that.
- I'm not really sure if this is the right fix, but it seems to work.

### Affirmation


dalthviz (approved):
Thanks @ccordoba12 ! I checked locally the installers and they worked for me :+1:

PR #17176, opened 2022-01-11
PR: Show tooltip with full path name in Files and Projects

ccordoba12:
## Description of Changes

This is a small improvement to allow users quickly see the full path name in those panes, in case it's not clearly visible in the pane itself.

imagen

### Affirmation

### DIFF
--- a/spyder/plugins/explorer/widgets/explorer.py
+++ b/spyder/plugins/explorer/widgets/explorer.py
@@ -27,7 +27,8 @@ from qtpy.QtCore import QDir, QMimeData, Qt, QTimer, QUrl, Signal, Slot
 from qtpy.QtGui import QDrag
 from qtpy.QtWidgets import (QApplication, QDialog, QDialogButtonBox,
                             QFileSystemModel, QInputDialog, QLabel, QLineEdit,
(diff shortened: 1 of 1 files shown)


steff456 (approved):
This PR looks good! The ellipses on the directory seem a little bit hard to implement so maybe we can leave that for another PR.

isabela-pf:
Nice! Thanks for submitting this pull request. 

I like the addition of tool tips (I mistakenly thought we already had that) and, as usual, would prefer if that behavior was consistent. Since Projects and Files share a lot of their UI, I'd like this to be shared too. 

I know many panes in Spyder use horizontal scrollbars for overflowing content, and I think Projects and Files do the same. Long term, I don't think this is the best choice unless it is absolutely necessary for the pane. Projects and Files seem like prime candidates to have their directory and file names truncated so they fit the window. This may be outside the scope of the PR, but I wanted to bring it up for consideration in case it was easy to test out here.

(P.S. I'm glad the tool tip hover is a little delayed. Good attention to detail.)

ccordoba12:
Thanks @steff456 and @isabela-pf for the review!

PR #17181, opened 2022-01-12
PR: Create new file with %edit magic (IPython console)

### LINKED ISSUE #3394: %edit magic can't create a new file

From an ipython prompt in Spyder, editing a non-existent file gives an error.  It would be great if this just created a new, blank file.  If there's a way to do this, or different command, I couldn't find it.

1. Open ipython console in spyder
2. type "edit foo.py" in a directory with no file foo.py

This isn't an issue but a feature request maybe.  I'd like an editor window with a blank file to open.  Instead I get an error.

Using Anaconda distribution
- Spyder Version: 2.3.9
- Python Version: 3.5.2
- Operating system: Windows 7
## Dependencies

and paste the contents below:

impact27:
## Description of Changes
When using edit magic on a non-existing file, this creates a new file instead of doing nothing.

### Issue(s) Resolved

Fixes #3394

### Affirmation


spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -1314,6 +1314,10 @@ def tab_name_editor(self):
     @Slot(object, object)
     def edit_file(self, filename, line):
         """Handle %edit magic petitions."""
+        if not osp.isfile(filename):
+            # Create new file
+            with open(filename, "w") as f:
+                f.write("")
```
ccordoba12, line 1320: I think we should add our template for untitled files here, instead of creating the file with empty content. What do you think?

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -1314,6 +1314,10 @@ def tab_name_editor(self):
     @Slot(object, object)
     def edit_file(self, filename, line):
         """Handle %edit magic petitions."""
+        if not osp.isfile(filename):
```
ccordoba12, line 1317: Perhaps we should surround this with a try/except in case people try to run this in a remote kernel? I mean, since both local and remote filesystems are different, this makes little sense.

ccordoba12 (commented):
This is a great idea @impact27! Thanks for looking into it.

ccordoba12 (approved):
Thanks @impact27! This is a great UX enhancement!

PR #17185, opened 2022-01-13
throught -> through (misspelling)

samiam2013:
## Description of Changes

corrected an English misspelling I noticed when installing Kite.

### Issue(s) Resolved

Fixes # N/A

### Affirmation

Samiam2013

### DIFF
--- a/spyder/locale/de/LC_MESSAGES/spyder.po
+++ b/spyder/locale/de/LC_MESSAGES/spyder.po
@@ -1023,7 +1023,7 @@ msgid "Downloading"
 msgstr "Wird heruntergeladen"
 
 #: spyder/plugins/completion/providers/kite/widgets/install.py:213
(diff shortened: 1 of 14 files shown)


dalthviz:
Hi @samiam2013 thank you for checking into this :+1: However the change needs to be done in the string inside the code rater than in the translations files (since these ones are regenerated from the string found in the source code). Just in case, the line to be changed is: https://github.com/spyder-ide/spyder/blob/5.x/spyder/plugins/completion/providers/kite/widgets/install.py#L213-L218

Also, note the base branch you are using, I would say that this will need to use as base branch `5.x` so you should create a branch to push your changes with something like:

```
git checkout 5.x
git checkout -b fix_kite_text
```

And the when doing the PR on GitHub select as `base` `5.x`

For more info related to the branches management you can check https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md#spyder-branches

Issue #17232, opened 2022-01-22
Require IPython >= 7.31.1 due to CVE-2022-21699 ?


rhkarls:
In light of [CVE-2022-21699](https://github.com/advisories/GHSA-pq7m-3gw7-gq5x) and its severity, should Spyder require patched IPython versions? Considering that dropping 3.6 is planned (https://github.com/spyder-ide/spyder/issues/16971) and IPython >= 8 is not currently supported this should mean the requirement of `ipython >=7.31.1,<8.0.0`

ccordoba12:
Hey @rhkarls, thanks a lot for the suggestion! We'll request that IPython version in our next version (5.3.0), to be released in a couple of months.

rhkarls:
Thanks for the follow up @ccordoba12 .It appears that at least 5.2.2 Windows installer is already packaged with 7.31.1, so that's great!

PR #17282, opened 2022-02-01
Modifications to allow IPython 8.0.x

juliangilbey:
## Description of Changes

This patch allows the Spyder tests to pass with IPython 8.0.1, and in fact with IPython >= 7.10.0 (which was released just over two years ago).  Debian is planning on releasing this newer version of IPython in the not-too-distant future, so I've been exploring whether Spyder works with it; it seems to be fairly smooth, except for one significant IPython bug that showed up with a test failing in your excellent test suite, and which I've just reported at https://github.com/ipython/ipython/issues/13507  But once that has been addressed, this very simple patch will allow Spyder 5.x to work with the older or newer versions of IPython.

### Affirmation

I certify the above statement is true and correct: juliangilbey


ccordoba12:
Thanks for your help with this @juliangilbey! Since this also requires a synced pull request with Spyder kernels (which has the same requirement in IPython as the one here), I'll take care of it for Spyder 5.3.0, to be released in March.

Besides, the minimal version we plan to require is #17232 due to its severity.

juliangilbey:
Sounds good to me! The only non-version-number patch in this PR was a change to one of the tests. Thanks :)

PR #17288, opened 2022-02-02
PR: Skip more IPython console tests because they started to hang

ccordoba12:
## Description of Changes

- This should prevent the hangs we've been seeing lately.
- This also fixes a failure with the Mac app.

### Affirmation

### DIFF
--- a/installers/macOS/setup.py
+++ b/installers/macOS/setup.py
@@ -103,6 +103,10 @@ def make_app_bundle(dist_dir, make_lite=False):
     PIL :
         Library not loaded: @loader_path/.dylibs/libjpeg.9.dylib
         Note: only applicable to not-Lite build
+    pkg_resources:
+        ImportError: The 'more_itertools' package is required; normally this is
+        bundled with this package so if you get this warning, consult the
+        packager of your distribution.
     pygments :
         ModuleNotFoundError: No module named 'pygments.formatters.latex'
     pylint :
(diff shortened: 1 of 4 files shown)


ccordoba12:
@impact27, I don't know if this is right, but I think it'll fix the problems with sporadic hangs we've seen in our test suite lately.

impact27:
we can maybe send a keyboard interrupt after some time, so we would know where the tests are hanging. Something like a timer of 10 min that is starter before each test and cancelled after?

ccordoba12:
@impact27, after deactivating the checks you added for known leaks, I noticed the hangs keep happening. So they were not introduced by that change.

I don't why started to happen since last Friday, but I skipped other hanging tests here. We could use [pytest-timeout](https://github.com/pytest-dev/pytest-timeout) instead, but that will make our tests take much longer to pass, which at the end is a liability.

impact27:
I am not sure I understand why the tests would be slower? If the tests don’t hang they would take the same time and if they do hang there would fail anyway, which is better that waiting the timeout from CI?

ccordoba12:
According to its Readme:

> While by default on POSIX systems pytest will continue to execute the tests after a test has timed out this is not always possible. Often the only sure way to interrupt a hanging test is by terminating the entire process

So, a timeout could kill the script that runs the whole test suite, so it'd need to be restarted again.

impact27:
Well if a timeout occurs should we consider that the tests have failed and therefore not restart before the source of the timeout is addressed? This way the tests would not take longer to execute, and would fail quickly rather than hanging forever?

ccordoba12:
Good idea, we could do that instead. Unfortunately, I don't have more time to work on this.

PR #17502, opened 2022-03-16
PR: Update macOS app to import user environment variables

### LINKED ISSUE #14843: Ability to use PYTHONPATH from user bash_profile in macOS application

mrclary:
## Description of Changes

For the Windows installer and launching Spyder from the command line, user environment variables are inherited. This was not the case for the macOS application. This PR changes the macOS installer to inherit the user's environment variables, similar to the case for launching from the command line.

### Issue(s) Resolved

Fixes #14843 

### Affirmation


ccordoba12 (approved):
Thanks @mrclary for your continued work to improve our Mac app!

PR #17512, opened 2022-03-19
PR: Consistently handle PYTHONPATH and add Import feature to PYTHONPATH Manager

### LINKED ISSUE #17511: Inconsistent use of system PYTHONPATH in completions and subprocesses
# Statement of Problem

If a user has a system `PYTHONPATH` environment variable defined (e.g. in `~/.bash_profile` on unix, or Environment Variables on Windows), then Spyder inconsistently handles it in the context of completions in the editor, IPython Console's `os.environ['PYTHONPATH']`, and IPython Console's `sys.path`. The following table summarizes the different behavior based on platform and Spyder launch mechanism.
* "Completions" indicates whether the system `PYTHONPATH` paths are included in Jedi's completion search paths
* "os.environ" indicates whether the system `PYTHONPATH` paths are included in the IPython Console's `os.environ['PYTHONPATH']`
* "sys.path" indicates whether the system `PYTHONPATH` paths are included in the IPython Console's `sys.path`

| Platform | Launch | Interpreter | Completions | os.environ PYTHONPATH | sys.path |
|---|---|---|---|---|---|
| macOS | bootstrap | internal | Y | N* | N |
| macOS | bootstrap | external | Y | N* | N |
| macOS | conda | internal | Y | Y | Y |
| macOS | conda | external | N | Y | Y |
| macOS | app | internal | Y | Y | Y |
| macOS | app | external | N | N | N |
| Win | bootstrap | internal | Y | N* | N |
| Win | bootstrap | external | N | N* | N |
| Win | conda | internal | N | Y | Y |
| Win | conda | external | N | Y | Y |
| Win | app | internal | N | Y^ | N |
| Win | app | external | N | N | N |

`*` This is due to `os.environ['PYTHONPATH']` being entirely replaced by `spyder-kernels` under `DEV` and should be resolved by #17408
`^` In the case of Windows application with internal interpreter for the IPython Console, the internal Spyder `pkgs` directory is also in `os.environ['PYTHONPATH']`

# Discussion

@spyder-ide/core-developers , what is the desired treatment of the system `PYTHONPATH` environment variable with regard to completions in the editor and the IPython Console's `os.environ['PYTHONPATH']` and `sys.path`?

My opinion is that the system `PYTHONPATH` environment variable should be treated the same across platforms and launch mechanisms: either included in/excluded from each of completions paths, IPython Console's `os.environ['PYTHONPATH']`, and IPython Console's `sys.path`.

I propose the following.
1. Exclude the system `PYTHONPATH` paths from completions paths, IPython Console's `os.environ['PYTHONPATH']`, and IPython Console's `sys.path`.
2. Provide a mechanism whereby the user can import the system `PYTHONPATH` paths into the PYTHONPATH Manager.
3. (bonus points) Provide a mechanism whereby the user can export the PYTHONPATH Manager paths to the system `PYTHONPATH` environment variable.

I think that item 1 is important because the user should not have "hidden" paths affecting completions in the Editor or executing code in the IPython Console. All relevant paths should be defined by the PYTHONPATH Manager and the executable environment's `site.py` module. 

However, the system `PYTHONPATH` environment variable may be useful to users. By providing the mechanism in item 2, we provide access to those paths while maintaining transparency (no "hidden" paths). Additionally, the full power to arrange or enable/disable paths is available to the user.

Spyder already provides item 3 for the Windows platform, but it is lacking for the Unix platforms (macOS and Linux). This feature can be added to these platforms as well by cleverly handling shell source scripts (e.g. `.bash_profile`). This is certainly more complicated for Unix platforms than for Windows, but I think it may be possible to achieve in a safe and stable way.

mrclary:
## Description of Changes

Explicitly excluded system `PYTHONPATH` paths from Editor completion paths, IPython Console `sys.path`, and IPython Console `os.environ['PYTHONPATH']`.
Added feature to import system `PYTHONPATH` paths into PYTHONPATH Manager.

The "Syncronize..." feature (Windows only) is changed to "Export" to better complement "Import" and comport with what it actually does. This remains a Windows only feature for the time being.

Resolved an issue where a path was added after canceling "Select directory" dialog.

ppm

### Issue(s) Resolved

Fixes #17511 

### Affirmation

### DIFF
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -3092,7 +3092,7 @@ def test_preferences_change_interpreter(qtbot, main_window):
     lsp = main_window.completions.get_provider('lsp')
     config = lsp.generate_python_config()
     jedi = config['configurations']['pylsp']['plugins']['jedi']
-    assert jedi['environment'] is None
+    assert jedi['environment'] is sys.executable
     assert jedi['extra_paths'] == []
 
     # Change main interpreter on preferences
@@ -3589,8 +3589,6 @@ def test_runcell_cache(main_window, qtbot, debug):
 
 @pytest.mark.slow
 @flaky(max_runs=3)
-@pytest.mark.skipif(not sys.platform.startswith('linux'),
-                    reason="Works reliably on Linux")
 def test_path_manager_updates_clients(qtbot, main_window, tmpdir):
     """Check that on path manager updates, consoles correctly update."""
     # Wait until the window is fully up
@@ -3611,31 +3609,27 @@ def test_path_manager_updates_clients(qtbot, main_window, tmpdir):
 
     cmd = 'import sys;print(sys.path)'
 
-    # Check Spyder is updated
-    main_window.console.execute_lines(cmd)
-    syspath = main_window.console.get_sys_path()
-    assert folder in syspath
+    # Check that there is at least one shell
+    shells = [c.shellwidget for c in main_window.ipyconsole.get_clients()
+              if c is not None]
+    assert len(shells) >= 1
 
     # Check clients are updated
-    count = 0
-    for client in main_window.ipyconsole.get_clients():
-        shell = client.shellwidget
-        if shell is not None:
-            control = shell._control
-            control.setFocus()
+    for shell in shells:
+        control = shell._control
+        control.setFocus()
 
-            qtbot.waitUntil(lambda: shell._prompt_html is not None,
-                            timeout=SHELL_TIMEOUT)
+        qtbot.waitUntil(lambda: shell._prompt_html is not None,
+                        timeout=SHELL_TIMEOUT)
 
-            with qtbot.waitSignal(shell.executed):
-                syspath = shell.execute(cmd)
+        with qtbot.waitSignal(shell.executed, timeout=SHELL_TIMEOUT):
+            shell.execute(cmd)
 
-            # `shell.executed` signal was not working so we use waitUntil
-            qtbot.waitUntil(lambda: test_folder in control.toPlainText(),
-                            timeout=SHELL_TIMEOUT)
-            assert test_folder in control.toPlainText()
-            count += 1
-    assert count >= 1
+        # Shell sys.path should be updated
+        # Output to shell may be delayed, timeout stands in for assertion
+        # control.toPlainText may have extra file separators so use test_folder
+        qtbot.waitUntil(lambda: test_folder in control.toPlainText(),
+                        timeout=SHELL_TIMEOUT)
 
 
 @pytest.mark.slow
--- a/spyder/plugins/completion/providers/languageserver/provider.py
+++ b/spyder/plugins/completion/providers/languageserver/provider.py
@@ -801,6 +801,7 @@ class LanguageServerProvider(SpyderCompletionProvider):
 
         # Jedi configuration
         env_vars = os.environ.copy()  # Ensure env is indepependent of PyLSP's
+        env_vars.pop('PYTHONPATH', None)
         if self.get_conf('default', section='main_interpreter'):
             # If not explicitly set, jedi uses PyLSP's sys.path instead of
             # sys.executable's sys.path. This may be a bug in jedi.
--- a/spyder/plugins/completion/providers/languageserver/provider.py
+++ b/spyder/plugins/completion/providers/languageserver/provider.py
@@ -14,6 +14,7 @@ import functools
 import logging
 import os
 import os.path as osp
+import sys
 
 # Third-party imports
 from qtpy.QtCore import Signal, Slot, QTimer
@@ -799,15 +800,15 @@ class LanguageServerProvider(SpyderCompletionProvider):
         }
 
         # Jedi configuration
+        env_vars = os.environ.copy()  # Ensure env is indepependent of PyLSP's
         if self.get_conf('default', section='main_interpreter'):
-            environment = None
-            env_vars = None
+            # If not explicitly set, jedi uses PyLSP's sys.path instead of
+            # sys.executable's sys.path. This may be a bug in jedi.
+            environment = sys.executable
         else:
             environment = self.get_conf('executable',
                                         section='main_interpreter')
-            env_vars = os.environ.copy()
-            # external interpreter should not use internal PYTHONPATH
-            env_vars.pop('PYTHONPATH', None)
+            # External interpreter cannot have PYTHONHOME
             if running_in_mac_app():
                 env_vars.pop('PYTHONHOME', None)
 
--- a/spyder/widgets/pathmanager.py
+++ b/spyder/widgets/pathmanager.py
@@ -70,7 +70,7 @@ class PathManager(QDialog):
         self.setAttribute(Qt.WA_DeleteOnClose)
         self.setWindowTitle(_("PYTHONPATH manager"))
         self.setWindowIcon(ima.icon('pythonpath'))
-        self.resize(500, 300)
+        self.resize(500, 400)
         self.import_button.setVisible(sync)
         self.export_button.setVisible(os.name == 'nt' and sync)
 
@@ -221,17 +221,29 @@ class PathManager(QDialog):
 
         if env_pypath:
             env_pypath = env_pypath.split(os.pathsep)
-            env_pypath_msg = '<br>'.join(env_pypath)
-	        answer = QMessageBox.question(
-                self,
-                _("Import"),
-                _("Do you want to import the contents of your "
-                  "<tt>PYTHONPATH</tt> environment variable into Spyder? "
-                  "The following paths will be imported:"
-                  "<br><br>" + env_pypath_msg),
-                QMessageBox.No | QMessageBox.Yes, QMessageBox.Yes)
 
-            if answer == QMessageBox.Yes:
+            dlg = QDialog(self)
+            dlg.setWindowTitle(_("PYTHONPATH"))
+            dlg.setWindowIcon(ima.icon('pythonpath'))
+            dlg.setAttribute(Qt.WA_DeleteOnClose)
+            dlg.setMinimumWidth(400)
+
+            label = QLabel("The following paths will be imported.")
+            listw = QListWidget(dlg)
+            listw.addItems(env_pypath)
+
+            bbox = QDialogButtonBox(
+                QDialogButtonBox.Ok | QDialogButtonBox.Cancel)
+            bbox.accepted.connect(dlg.accept)
+            bbox.rejected.connect(dlg.reject)
+
+            layout = QVBoxLayout()
+            layout.addWidget(label)
+            layout.addWidget(listw)
+            layout.addWidget(bbox)
+            dlg.setLayout(layout)
+
+            if dlg.exec():
                 spy_pypath = self.get_path_dict()
                 n = len(spy_pypath)
 
@@ -476,7 +488,7 @@ def test():
     _ = qapplication()
     dlg = PathManager(
         None,
-        path=tuple(sys.path[:-2]),
+        path=tuple(sys.path[4:-2]),
         read_only_path=tuple(sys.path[-2:]),
     )
 
--- a/spyder/widgets/pathmanager.py
+++ b/spyder/widgets/pathmanager.py
@@ -226,8 +226,9 @@ class PathManager(QDialog):
         answer = QMessageBox.question(
             self,
             _("Import"),
-            _("Do you want to import the following paths into "
-              "PYTHONPATH Manager?"
+            _("Do you want to import the contents of your <tt>PYTHONPATH</tt> "
+              "environment variable into Spyder? The following paths will be "
+              "imported:"
               "<br><br>"
               + env_pypath_msg),
             QMessageBox.No | QMessageBox.Yes, QMessageBox.Yes)
--- a/spyder/widgets/pathmanager.py
+++ b/spyder/widgets/pathmanager.py
@@ -23,7 +23,6 @@ from qtpy.QtWidgets import (QDialog, QDialogButtonBox, QHBoxLayout,
 
 # Local imports
 from spyder.config.base import _
-from spyder.py3compat import PY2
 from spyder.utils.icon_manager import ima
 from spyder.utils.misc import getcwd_or_home
 from spyder.utils.qthelpers import create_toolbutton
@@ -53,6 +52,7 @@ class PathManager(QDialog):
         self.moveup_button = None
         self.movedown_button = None
         self.movebottom_button = None
+        self.import_button = None
         self.export_button = None
         self.selection_widgets = []
         self.top_toolbar_widgets = self._setup_top_toolbar()
@@ -168,8 +168,7 @@ class PathManager(QDialog):
             tip=_("Export to PYTHONPATH environment variable"),
             text_beside_icon=True)
 
-        self.selection_widgets.append(self.remove_button)
-        return [self.add_button, self.remove_button, None, self.import_button,
+        return [self.add_button, self.remove_button, self.import_button,
                 self.export_button]
 
     def _create_item(self, path):
@@ -193,10 +192,7 @@ class PathManager(QDialog):
     def editable_bottom_row(self):
         """Maximum bottom row count that is editable."""
         read_only_count = len(self.read_only_path)
-        if read_only_count == 0:
-            max_row = self.listwidget.count() - 1
-        else:
-            max_row = self.listwidget.count() - read_only_count - 1
+        max_row = self.listwidget.count() - read_only_count - 1
         return max_row
 
     def setup(self):
@@ -313,6 +309,8 @@ class PathManager(QDialog):
         for widget in disable_widgets:
             widget.setEnabled(False)
 
+        self.remove_button.setEnabled(self.listwidget.count()
+                                      - len(self.read_only_path))
         self.export_button.setEnabled(self.listwidget.count() > 0)
 
         # Ok button only enabled if actual changes occur
@@ -341,23 +339,7 @@ class PathManager(QDialog):
             directory = getexistingdirectory(self, _("Select directory"),
                                              self.last_path)
             self.redirect_stdio.emit(True)
-
-        if PY2:
-            is_unicode = False
-            try:
-                directory.decode('ascii')
-            except (UnicodeEncodeError, UnicodeDecodeError):
-                is_unicode = True
-
-            if is_unicode:
-                QMessageBox.warning(
-                    self,
-                    _("Add path"),
-                    _("You are using Python 2 and the selected path has "
-                      "Unicode characters."
-                      "<br> "
-                      "Therefore, this path will not be added."),
-                    QMessageBox.Ok)
+            if not directory:
                 return
 
         directory = osp.abspath(directory)
--- a/spyder/widgets/pathmanager.py
+++ b/spyder/widgets/pathmanager.py
@@ -37,7 +37,7 @@ class PathManager(QDialog):
                  not_active_path=None, sync=True):
         """Path manager dialog."""
         super(PathManager, self).__init__(parent)
-        assert isinstance(path, (tuple, None))
+        assert isinstance(path, (tuple, type(None)))
 
         self.path = path or ()
         self.read_only_path = read_only_path or ()
@@ -220,19 +220,31 @@ class PathManager(QDialog):
         env_pypath = os.environ.get('PYTHONPATH', '')
         if not env_pypath:
             return
+
         env_pypath = env_pypath.split(os.pathsep)
-        env_pypath.reverse()
+        env_pypath_msg = '<br>'.join(env_pypath)
+        answer = QMessageBox.question(
+            self,
+            _("Import"),
+            _("Do you want to import the following paths into "
+              "PYTHONPATH Manager?"
+              "<br><br>"
+              + env_pypath_msg),
+            QMessageBox.No | QMessageBox.Yes, QMessageBox.Yes)
 
-        spy_pypath = self.get_path_dict()
-        n = len(spy_pypath)
+        if answer == QMessageBox.Yes:
+            spy_pypath = self.get_path_dict()
+            n = len(spy_pypath)
 
-        for path in env_pypath:
-            if (path in spy_pypath) or not self.check_path(path):
-                continue
-            item = self._create_item(path)
-            self.listwidget.insertItem(n, item)
+            for path in reversed(env_pypath):
+                if (path in spy_pypath) or not self.check_path(path):
+                    continue
+                item = self._create_item(path)
+                self.listwidget.insertItem(n, item)
 
-        self.refresh()
+            self.refresh()
+        else:
+            return
 
     @Slot()
     def export_pythonpath(self):
--- a/spyder/widgets/pathmanager.py
+++ b/spyder/widgets/pathmanager.py
@@ -19,7 +19,7 @@ from qtpy.compat import getexistingdirectory
 from qtpy.QtCore import Qt, Signal, Slot
 from qtpy.QtWidgets import (QDialog, QDialogButtonBox, QHBoxLayout,
                             QListWidget, QListWidgetItem, QMessageBox,
-                            QVBoxLayout)
+                            QVBoxLayout, QLabel)
 
 # Local imports
 from spyder.config.base import _
@@ -75,6 +75,13 @@ class PathManager(QDialog):
         self.export_button.setVisible(os.name == 'nt' and sync)
 
         # Layouts
+        description = QLabel(
+            _("Only paths listed here will be passed to IPython consoles and "
+              "the language server<br>as additional Python search paths."
+              "<br>"
+              "Paths in your system <b>PYTHONPATH</b> environment variable "
+              "can be imported to this list.")
+        )
         top_layout = QHBoxLayout()
         self._add_widgets_to_layout(self.top_toolbar_widgets, top_layout)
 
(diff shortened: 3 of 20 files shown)


mrclary:
We had been removing `PYTHONPATH` from `env_vars` and using `SPY_PYTHONPATH` instead. This is only used to set the `sys.path` paths in the IPython Console at launch. Eliminating `SPY_PYTHONPATH` in favor of `PYTHONPATH` will produce the same result while also providing `os.environ['PYTHONPATH']` in the IPython Console and obviate `set_spyder_pythonpath` in `spyder-kernels`.

A companion PR spyder-ide/spyder-kernels#378 accommodates this change and updates IPython Console's `os.environ['PYTHONPATH']` along with the `sys.path` update.

ccordoba12:
Thanks for your work on this @mrclary! I'm leaving this for 5.3.1 because this is an important change and we need to carefully discuss about it.

spyder/app/tests/test_mainwindow.py
```diff
@@ -3470,31 +3468,24 @@ def test_path_manager_updates_clients(qtbot, main_window, tmpdir):
...
-            qtbot.waitUntil(lambda: shell._prompt_html is not None,
-                            timeout=SHELL_TIMEOUT)
-
-            with qtbot.waitSignal(shell.executed):
-                syspath = shell.execute(cmd)
-
-            # `shell.executed` signal was not working so we use waitUntil
-            qtbot.waitUntil(lambda: test_folder in control.toPlainText(),
-                            timeout=SHELL_TIMEOUT)
-            assert test_folder in control.toPlainText()
```
mrclary, line 3495: @goanpeca @dalthviz , why should `qtbot` wait for `test_folder in control.toPlainText()` to be True? This is the assertion condition, so if the shell failed to update, `qtbot` will unnecessarily wait for 20-30 seconds before moving forward to the assertion statement. If the intent is to ensure that the text is written to the console before checking the assertion, then isn't this satisfied by the previous `qtbot.waitSignal(shell.executed)`? Perhaps a `SHELL_TIMEOUT` added here as well?

spyder/app/tests/test_mainwindow.py
```diff
@@ -3453,8 +3453,6 @@ def test_runcell_cache(main_window, qtbot, debug):
 
 @pytest.mark.slow
 @flaky(max_runs=3)
-@pytest.mark.skipif(not sys.platform.startswith('linux'),
-                    reason="Works reliably on Linux")
```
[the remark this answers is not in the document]
ccordoba12: To be honest, I don't remember. Sometimes our test suite starts to fail at random and I have to add skip's here and there to avoid that.

ccordoba12:
@mrclary, we discussed your PR with @CAM-Gerlach and @isabela-pf this week and we consider it requires some UI/UX improvements, so that users better understand how the PYTHONPATH manager works with the improvements you propose.

@isabela-pf will be posting more details about it soon.

mrclary:
> @mrclary, we discussed your PR with @CAM-Gerlach and @isabela-pf this week and we consider it requires some UI/UX improvements, so that users better understand how the PYTHONPATH manager works with the improvements you propose.
> 
> @isabela-pf will be posting more details about it soon.

Thanks, @ccordoba12 @CAM-Gerlach and @isabela-pf .

Another thing to discuss is whether we should intercept `PYTHONPATH` in Spyder at launch. In this PR I have only ensured that `PYTHONPATH` is handled consistently between platforms, launch mechanisms, and completions/IPython Console. @ccordoba12 implemented `SPY_PYTHONPATH` to intercept PYTHONPATH Manager paths and re-inject them into `sys.path` in IPython Console without crashing the kernel if the user shadowed standard libraries. This was a good idea and I'm wondering if we shouldn't implement something similar for Spyder at launch. Spyder does not need the user's system `PYTHONPATH` for anything, but it does get injected into Spyder's runtime `sys.path`, so if it does shadow standard libraries it could cause problems for Spyder. Although I don't think we've seen any issues posted, so maybe it's nothing to be concerned about, it seems a relatively cheap thing to do to ensure Spyder is robust against user settings.

mrclary:
@isabela-pf, did you have some feedback on this PR?

mrclary:
@isabela-pf, thanks for responding. 

@spyder-ide/core-developers 
> * Since (from my understanding) they behave differently, I think there should be some kind of marker to label what is manually added versus what is added with the new feature. (This might be helpful to know after like for exporting or to remove things.)
> * Reorder the buttons to group them by use

I not sure what the above means; perhaps we can discuss at our meeting on Thursday.

> * Consider having more descriptive button names, like "Import from PYTHONPATH." Even though it's longer, the clarity might be better.

Is the hover text insufficient?
Screen Shot 2022-04-26 at 12 43 27 PM

> * @CAM-Gerlach recommended a multicolumn approach to paths by how they are brought in. this seems like a lot.

I don't see any purpose for designating the origin of paths in the PYTHONPATH Manager. Whether they are imported or manually entered, it is just a simple list that the user may do with what they wish (remove, reorder, disable).

> * Also, should this be a new section in Preferences since it is a global Spyder change?

I don't know what you mean by "global" Spyder change, nevertheless, I don't think there is a need for a section in Preferences related to anything in this PR.

mrclary:
Okay, per conversation with @CAM-Gerlach today, I will tidy up the UI
- Add explanation at the top of the PYTHONPATH Manager dialog explaining what `PYTHONPATH` does and what exactly Spyder does with it, so the user know what to expect.
- Ensure that the Add path, Remove path, Import, and Export buttons are grouped together on the left of the dialog.
- Ensure proper tool tips for the Import and Export buttons

If more clarifications are needed or anyone has further questions or concerns, please reach out.

isabela-pf:
I'm glad @CAM-Gerlach was able to clarify! I don't want to block this PR with my limited time, so please don't feel the need to wait for me.

mrclary:
@ccordoba12 @CAM-Gerlach, I think I've addressed the final issues. If there are any revisions, please let me know. I've updated the PR description as well.

mrclary:
@ccordoba12 @CAM-Gerlach @dalthviz Are we waiting on anything for this PR?

spyder/widgets/pathmanager.py
```diff
@@ -71,9 +71,17 @@ def __init__(self, parent, path=None, read_only_path=None,
...
+        self.import_button.setVisible(sync)
+        self.export_button.setVisible(os.name == 'nt' and sync)
 
         # Layouts
+        description = QLabel(
+            _("Only paths listed here will be passed to IPython consoles and "
+              "the language server<br>as additional Python search paths."
+              "<br>"
+              "Paths in your system <b>PYTHONPATH</b> environment variable "
+              "can be imported to this list.")
```
CAM-Gerlach, line 83: I tried to make the wording a little clearer and more understandable, though I'm still not 100% happy with it. I also elided the hard breaks for now, since they won't line up the same as before, and may not produce the desired results at different window widths, UI scaling, font sizes and translations; if you want the second break, it should generally be a full double paragraph `<br>`.

```suggestion
            _("The paths listed below will be passed to IPython consoles and "
              "the language server as additional locations to search for "
              "Python modules. "
              "Any paths in your system <b>PYTHONPATH</b> environment "
              "variable can be imported here if you'd like to use them.")
```

spyder/widgets/pathmanager.py
```diff
@@ -202,15 +214,35 @@ def setup(self):
...
-        Only apply to: current user, on Windows platforms.
+        Export to PYTHONPATH environment variable
+        Only apply to: current user.
         """
         answer = QMessageBox.question(
             self,
-            _("Synchronize"),
-            _("This will synchronize Spyder's path list with "
+            _("Export"),
+            _("This will export Spyder's path list to "
```
CAM-Gerlach, line 245:
```suggestion
            _("This will export Spyder's path list to the "
```

Missing word

CAM-Gerlach:
Sorry for the delay; I've been at PyCon US for the past >week, just got back yesterday and have been trying to catch up. This looks pretty good, aside from a few textual tweaks :+1:

ccordoba12 (commented):
Thanks @mrclary for your patience with this one! I think it's shaping very nicely.

spyder/widgets/pathmanager.py
```diff
@@ -71,9 +71,18 @@ def __init__(self, parent, path=None, read_only_path=None,
         self.setWindowTitle(_("PYTHONPATH manager"))
         self.setWindowIcon(ima.icon('pythonpath'))
         self.resize(500, 300)
-        self.sync_button.setVisible(os.name == 'nt' and sync)
+        self.import_button.setVisible(sync)
+        # self.export_button.setVisible(os.name == 'nt' and sync)
```
mrclary, line 75:
```suggestion
        self.export_button.setVisible(os.name == 'nt' and sync)
```

spyder/widgets/pathmanager.py
```diff
@@ -202,20 +215,40 @@ def setup(self):
...
+            item = self._create_item(path)
+            self.listwidget.insertItem(n, item)
+
+        self.refresh()
+
+    @Slot()
+    def export_pythonpath(self):
         """
-        Synchronize Spyder's path list with PYTHONPATH environment variable
-        Only apply to: current user, on Windows platforms.
```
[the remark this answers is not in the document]
mrclary: No, sorry; thanks for catching that. See my comment for line 75.

spyder/plugins/completion/providers/languageserver/provider.py
```diff
@@ -799,15 +800,15 @@ def generate_python_config(self):
...
-            env_vars = None
+            # If not explicitly set, jedi uses PyLSP's sys.path instead of
+            # sys.executable's sys.path. This may be a bug in jedi.
+            environment = sys.executable
         else:
             environment = self.get_conf('executable',
                                         section='main_interpreter')
-            env_vars = os.environ.copy()
-            # external interpreter should not use internal PYTHONPATH
-            env_vars.pop('PYTHONPATH', None)
```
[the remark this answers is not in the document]
mrclary: Oops, I missed that. You are correct, Jedi does not need `PYTHONPATH` and should not have it set. We pass all the extra paths via `extra_paths`

spyder/plugins/completion/providers/languageserver/provider.py
```diff
@@ -799,15 +800,15 @@ def generate_python_config(self):
         }
 
         # Jedi configuration
+        env_vars = os.environ.copy()  # Ensure env is indepependent of PyLSP's
```
mrclary, line 803:
```suggestion
        env_vars = os.environ.copy()  # Ensure env is indepependent of PyLSP's
        env_vars.pop('PYTHONPATH', None)
```

spyder/widgets/pathmanager.py
```diff
@@ -202,20 +215,40 @@ def setup(self):
         self.refresh()
 
     @Slot()
-    def synchronize(self):
+    def import_pythonpath(self):
```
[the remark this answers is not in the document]
mrclary: Okay, I'll implement.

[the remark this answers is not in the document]
CAM-Gerlach: > This action cannot be easily reversed. So, if users have a PYTHONPATH with say 20 paths and click on this but they really don't want to use them in Spyder, they will be forced to remove all those paths from our manager. Therefore, I think it's better to ask them if they really want to proceed before doing that.

Allowing multiple selection (Shift, Ctrl, etc) selection would greatly improve usability here, but presumably the button actions would need to be updated to handle that; I'm not sure how difficult that would be to do.

spyder/widgets/pathmanager.py
```diff
@@ -202,20 +215,52 @@ def setup(self):
...
+        if not env_pypath:
+            return
+
+        env_pypath = env_pypath.split(os.pathsep)
+        env_pypath_msg = '<br>'.join(env_pypath)
+        answer = QMessageBox.question(
+            self,
+            _("Import"),
+            _("Do you want to import the following paths into "
+              "PYTHONPATH Manager?"
```
ccordoba12, line 230:
```suggestion
            _("Do you want to import the contents of your <tt>PYTHONPATH</tt> "
              "environment variable into Spyder? The following paths will be "
              "imported:"
```

* I think we can be clearer and more explicit here.
* I like that users can check the paths that will be imported. Thanks for adding that.

spyder/widgets/pathmanager.py
```diff
@@ -202,20 +215,52 @@ def setup(self):
         self.refresh()
 
     @Slot()
-    def synchronize(self):
+    def import_pythonpath(self):
+        """Import from PYTHONPATH environment variable"""
+        env_pypath = os.environ.get('PYTHONPATH', '')
+        if not env_pypath:
+            return
```
ccordoba12, line 222: This is confusing because the button does nothing when clicked, which users could interpret as it being broken. So, in this case I suggest to add a `QMessageBox.information` saying that `PYTHONPATH` is empty, so there's nothing to import from it.

ccordoba12 (commented):
I think this is almost ready, thanks @mrclary for the updates.

spyder/widgets/pathmanager.py
mrclary: No problem, I'll add another modal window.

spyder/widgets/pathmanager.py
```diff
@@ -202,20 +215,58 @@ def setup(self):
...
+                        continue
+                    item = self._create_item(path)
+                    self.listwidget.insertItem(n, item)
+
+                self.refresh()
+        else:
+            QMessageBox.information(
+                self,
+                _("PYTHONPATH"),
+                _("<tt>PYTHONPATH</tt> is empty; nothing to import."),
```
ccordoba12, line 252:
```suggestion
                _("Your <tt>PYTHONPATH</tt> environment variable is empty, so "
                  "there is nothing to import."),
```

Sorry to be so picky about these messages, but I think the more explicit we are in them, the easier would be for users to understand what this is about in case they don't know it.

For example, by saying that `PYTHONPATH` is an env var, it'd be easier for them to google for it.

ccordoba12 (commented):
@mrclary, besides my small suggestion below, I think we should leave the feature of adding paths to the front of `sys.path` for another PR.

That's because if we want to make that configurable (as we agreed on https://github.com/spyder-ide/spyder-kernels/pull/378), I think we need to change the `update_syspath` handler on Spyder-kernels. And that requires a new minor Spyder-kernels version (i.e. 2.4.0) and a corresponding minor Spyder version too (i.e. 5.4.0).

spyder/widgets/pathmanager.py
mrclary: No problem.

mrclary:
> @mrclary, besides my small suggestion below, I think we should leave the feature of adding paths to the front of `sys.path` for another PR.
> 
> That's because if we want to make that configurable (as we agreed on [spyder-ide/spyder-kernels#378](https://github.com/spyder-ide/spyder-kernels/pull/378)), I think we need to change the `update_syspath` handler on Spyder-kernels. And that requires a new minor Spyder-kernels version (i.e. 2.4.0) and a corresponding minor Spyder version too (i.e. 5.4.0).

I'll submit a (draft) PR for adding the feature.

CAM-Gerlach:
> Additionally, it is not using the correct external-deps spyder-kernels.

Perhaps because it is running normally using the PyPI deps instead of installing the subrepos?

mrclary:
> > Additionally, it is not using the correct external-deps spyder-kernels.
> 
> Perhaps because it is running normally using the PyPI deps instead of installing the subrepos?

The behavior I see suggests that may be what is happening. But I don't know anything about the binder setup to say anything intelligent...

mrclary:
> > > Additionally, it is not using the correct external-deps spyder-kernels.
> > 
> > 
> > Perhaps because it is running normally using the PyPI deps instead of installing the subrepos?
> 
> The behavior I see suggests that may be what is happening. But I don't know anything about the binder setup to say anything intelligent...

Okay, so I forgot to add back the subrepo clone into this PR; that's why binder was mis behaving. I just tested it now and it is working as expected, as far as this PR is concerned. The jellyfish, pyls_spyder, and pylsp_black issue is still present.

CAM-Gerlach:
>  The jellyfish, pyls_spyder, and pylsp_black issue is still present.

Strange, since you added it to the binder environment.yaml at the same time as the conda, etc. requirements, and it indeed [appears to be present](https://github.com/mrclary/spyder/blob/pypath-manager-import/binder/environment.yml) on your branch. You could take a look with `conda list` on the Binder instance to see if it is present in the env

ccordoba12:
> There appears to be some issues with binder.

We haven't updated the Binder deps in a long time (perhaps eight months). They are declared here (per this repo branch):

https://github.com/spyder-ide/binder-environments

Could you give us a hand with that?

> Okay, so I forgot to add back the subrepo clone into this PR

I think your changes in Spyder-kernels should be dropped from this PR and left for PR #18308. That way we could merge this one (which I think is ready) and you could keep working on solving issue #17066 there.

What do you think?

mrclary:
> > There appears to be some issues with binder.
> 
> We haven't updated the Binder deps in a long time (perhaps eight months). They are declared here (per this repo branch):
> 
> https://github.com/spyder-ide/binder-environments
> 
> Could you give us a hand with that?

No problem

> > Okay, so I forgot to add back the subrepo clone into this PR
> 
> I think your changes in Spyder-kernels should be dropped from this PR and left for PR #18308. That way we could merge this one (which I think is ready) and you could keep working on solving issue #17066 there.
> 
> What do you think?

I'm happy to do whatever you think is best. However, if spyder-ide/spyder-kernels#378 is not included with this PR, then paths in IPython Console `sys.path` will continue to be positioned differently depending on whether the user enabled/disabled PPM paths. If spyder-ide/spyder-kernels#378 is rolled into #18308, then it may be some time before this behavior is fixed since we'd have to wait until a minor version bump in both repos.

ccordoba12:
> However, if https://github.com/spyder-ide/spyder-kernels/pull/378 is not included with this PR, then paths in IPython Console sys.path will continue to be positioned differently depending on whether the user enabled/disabled PPM paths

Ok, I see. What if you change https://github.com/spyder-ide/spyder-kernels/pull/378 so that paths are added to the bottom of `sys.path` for now? Then you could make that configurable on PR #18308.

CAM-Gerlach:
> We haven't updated the Binder deps in a long time (perhaps eight months). They are declared here (per this repo branch):

Is there a reason they are also declared under `binder/requirements.txt` in this repo, which has been kept up to date and is the location as documented in `requirements/conda.txt`? This was rather confusing to me, since I checked there as documented and they were all up to date.

mrclary:
> > However, if [spyder-ide/spyder-kernels#378](https://github.com/spyder-ide/spyder-kernels/pull/378) is not included with this PR, then paths in IPython Console sys.path will continue to be positioned differently depending on whether the user enabled/disabled PPM paths
> 
> Ok, I see. What if you change [spyder-ide/spyder-kernels#378](https://github.com/spyder-ide/spyder-kernels/pull/378) so that paths are added to the bottom of `sys.path` for now? Then you could make that configurable on PR #18308.

No problem.

ccordoba12:
@mrclary, I can't test this PR because for me the `spyder-kernels` subrepo is not added to `sys.path` with your changes here. So, we need to merge PR #17408 first and then continue with this one.

mrclary:
> Hi @mrclary, I already tested this PR in mac using the installer and it is working as expected. My only comment is why the text in the buttons seems so small? I'm not sure if we have any dialog with the text so small.

I assume that you are referring to the question message box. I chose to try to distinguish the question from the list of paths using the `setInformativeText` method in order to make everything more readable. However, the following illustrate the different options.
- `setInformativeText`
  Screen Shot 2022-06-20 at 2 28 38 PM
- Combined with the question text
  Screen Shot 2022-06-20 at 2 36 34 PM
- `setDetailedText`
  details

Perhaps others have a different preference than mine?

In any case, it would be nice if the message box width adapted to greater-than-or-equal-to the width of the largest path string since distinguishing paths can be difficult if their text wraps. If anyone has a clever suggestion there, I would much appreciate it. Note that while the pats do not wrap for the `setInformativeText` case shown above, this is only circumstantial and may wrap depending on the length of the paths listed.

mrclary:
> @mrclary, I can't test this PR because for me the `spyder-kernels` subrepo is not added to `sys.path` with your changes here. So, we need to merge PR #17408 first and then continue with this one.

That's fine with me. Sorry that is working; I thought I sufficiently separated those two PRs, such that this one should not have affected bootstrapping behavior (sys.path manipulations under DEV should not have been touched). Maybe I missed something.

steff456:
> Perhaps others have a different preference than mine?

I think of all the options you showed I prefer the third one, it will not overflow in case someone is importing a huge list. I know the wrap is inevitable, so that is not a huge blocker for me. Maybe adding something to recognize a new line may help (?) I have to check if we have something like that in the interface right now, but if not I think it is just fine as you have it.

mrclary:
I added a list marker for each path.

Screen Shot 2022-06-20 at 6 32 49 PM

mrclary:
> @mrclary Can you change the UI text of the Show details button to `Show paths` (unless that is hardcoded in Qt)? This would make it more obvious to users what it does and where to actually view the paths that will be imported.

Unfortunately it appears to be hardcoded. I can locate the button and change the text, but the text updates upon clicking from "Show details..." to "Hide details...". I can change the initial text to "Show paths...", but cycling through the button restores "Show details...". I'd have to locate the method changing the text and patch it. Perhaps it would be easier to just create our own QWidget... Let me see what I can do.

mrclary:
@spyder-ide/core-developers how about the following?

listwidget

steff456:
It looks better!

ccordoba12:
Agreed, it looks much nicer.

spyder/widgets/pathmanager.py
```diff
@@ -202,20 +215,68 @@ def setup(self):
...
+                    self.listwidget.insertItem(n, item)
+
+                self.refresh()
+        else:
+            QMessageBox.information(
+                self,
+                _("PYTHONPATH"),
+                _("Your <tt>PYTHONPATH</tt> environment variable is empty, so "
+                  "there is nothing to import."),
+                QMessageBox.Ok)
```
ccordoba12, line 263:
```suggestion
                QMessageBox.Ok
            )
```

Style improvement.

spyder/widgets/pathmanager.py
```diff
@@ -202,20 +215,68 @@ def setup(self):
...
+            _("This will export Spyder's path list to the "
               "<b>PYTHONPATH</b> environment variable for the current user, "
               "allowing you to run your Python modules outside Spyder "
               "without having to configure sys.path. "
-              "<br>"
-              "Do you want to clear contents of PYTHONPATH before "
+              "<br><br>"
+              "Do you want to clear the contents of PYTHONPATH before "
               "adding Spyder's path list?"),
             QMessageBox.Yes | QMessageBox.No | QMessageBox.Cancel)
```
ccordoba12, line 281:
```suggestion
            QMessageBox.Yes | QMessageBox.No | QMessageBox.Cancel
        )
```

Style improvement

ccordoba12 (commented):
Last minor details @mrclary, then this should be ready.

Also, since I just merged https://github.com/spyder-ide/spyder-kernels/pull/378, please resync the spyder-kernels subrepo with the instructions in our Contributing guide.

ccordoba12 (approved):

dalthviz (commented):
Thanks @mrclary ! Checking this locally on Windows seems like is working correctly although to import exported paths you need to restart Spyder and/or run it from a new console/cmd. Seems like the changes done to `PYTHONPATH`  are not being taken into account without that (I think that's something that was happening before this PR though, right?). So for example, if I use `Add a path`, use `Export`, remove the path from the manager with `Remove path` and then try to use `Import` I see the empty `PYTHONPATH` while in the consoles I can get the PYTHONPATH update:

imagen

If that is expected then this LGTM :+1:

mrclary:
> Thanks @mrclary ! Checking this locally on Windows seems like is working correctly although to import exported paths you need to restart Spyder and/or run it from a new console/cmd. Seems like the changes done to `PYTHONPATH` are not being taken into account without that (I think that's something that was happening before this PR though, right?). So for example, if I use `Add a path`, use `Export`, remove the path from the manager with `Remove path` and then try to use `Import` I see the empty `PYTHONPATH` while in the consoles I can get the PYTHONPATH update:

The "import" feature currently uses the existing `os.environ['PYTHONPATH']` in Spyder's runtime environment which is established at launch. We could change this with one of the following options:
- Upon exporting PPM to the system PYTHONPATH (Windows only), we could also appropriately update Spyder's runtime `os.environ['PYTHONPATH']`. Spyder does not use this internally for any reason, so this should be safe.
- Rather than relying on `os.environ['PYTHONPATH']`, we could retrieve the system environment variable directly. This should be straightforward on Windows by querying the registry (I think this is already done for the purpose of exporting anyway). On unix, a login shell subprocess would be needed to extract the environment variable; this would require a little more work to setup properly as it depends on the user's default shell etc., but is certainly feasible.

The second option may be more robust and future-proof. For example, @CAM-Gerlach has alluded to pending updates to Python which may use new `-P` flags by default which will preclude use of `PYTHON*` environment variables. If/when this is the case, `os.environ['PYTHONPATH']` may no longer be available by default, so a different mechanism may be prudent.

> If that is expected then this LGTM 👍

Technically, it is expected. Nevertheless, it may not be desirable. I think I'll include a fix for this in this PR. Unless @ccordoba12, @dalthviz, or @CAM-Gerlach have anything other recommendations, I'll proceed with the second option.

mrclary:
> Sounds good to me 👍 Although, since the current behavior is expected maybe a new issue could be open to track the missing improvements (probably using the comment above [#17512 (comment)](https://github.com/spyder-ide/spyder/pull/17512#issuecomment-1163398698) as an starting point). In that way we could check a little bit more ideas and other missing things regarding the PYTHONPATH manager without that being a blocker to merge this

Good idea. I'll proceed with the second option in another pull request in order to allow more in-depth feedback without blocking this PR.

dalthviz (approved):

steff456 (approved):
Thanks for working on this! I think it is ready to merge 😬

Issue #16927, opened 2021-12-01
Debugger message bug gets printed in the IPython Console


dalthviz:
## Problem Description

After starting an IPython Console with the standalon installer on Windows sometimes you get a message at the top regarding the debugger.

Checking seems like this was reported at https://github.com/spyder-ide/spyder/issues/16896#issuecomment-979837635 but I only notice it until now. Not totally sure if this is some sort of warning that needs to be handled/removed or something else @ccordoba12 @impact27 ?

### Paste Traceback/Error Below (if applicable)

```python-traceback

It seems the debugger cannot resolve D:\obj\windows-release\37amd64_Release\msi_python\zip_amd64\ntpath.py
This may make the debugger miss breakpoints in the standard library.
Related bug: https://bugs.python.org/issue1180193

```

## Versions
* Spyder version: 5.2.0 None
* Python version: 3.7.9 64-bit
* Qt version: 5.12.10
* PyQt5 version: 5.12.3
* Operating System: Windows 10

impact27:
In PR https://github.com/spyder-ide/spyder/pull/14025 there was an issue with: 


This version of python seems to be incorrectly compiled
  (internal generated filenames are not absolute).
  This may make the debugger miss breakpoints.
  Related bug: http://bugs.python.org/issue1666807


(Maybe related, we solved that one by skipping it)

impact27:
Do you see any adverse effect while debugging?

dalthviz:
I checked putting some breakpoints in a simple script and seems like the breakpoints work. 

Will check then the logic to skip it. Thanks for the info @impact27 :+1:

ccordoba12:
> Not totally sure if this is some sort of warning that needs to be handled/removed or something else @ccordoba12 @impact27 ?

This message (like the one referenced above by @impact27) is generated by `Pydevd`, a vendored dependency of `debugpy`, which in turn is a dependency of `IPykernel` 6+. We're not using it in our debugger, but it was introduced by the Jupyter team to implement the [Debugger Adapter Protocol](https://microsoft.github.io/debug-adapter-protocol/), which is like the LSP but for debuggers.

gaotxg:
This issue still can be reproduced in 5.2.1 Windows standalone installer.
Tested in Windows 10 21H2.

dalthviz:
Hi @gaotxg a new fix was done at https://github.com/spyder-ide/spyder/pull/17050 (will be available with the next release)

Issue #16828, opened 2021-11-15
Python seems to be incorrectly compiled for macOS installer


mrclary:
## Problem Description

The macOS standalone application reports the following warning in the IPython console when the Python interpreter is set to internal (same as Spyder):
```
This version of python seems to be incorrectly compiled
(internal generated filenames are not absolute).
This may make the debugger miss breakpoints.
Related bug: http://bugs.python.org/issue1666807
```

### What steps reproduce the problem?

1. Launch macOS standalone Spyder application with Python interpreter set to internal

## Versions

* Spyder version: 5.2.0.dev0 None
* Python version: 3.9.5 64-bit
* Qt version: 5.12.10
* PyQt5 version: 5.12.3
* Operating System: Darwin 20.6.0

### Dependencies

```
# Mandatory:
applaunchservices >=0.1.7     :  0.2.1 (OK)
atomicwrites >=1.2.0          :  1.4.0 (OK)
chardet >=2.0.0               :  4.0.0 (OK)
cloudpickle >=0.5.0           :  2.0.0 (OK)
cookiecutter >=1.6.0          :  1.7.3 (OK)
diff_match_patch >=20181111   :  20200713 (OK)
intervaltree >=3.0.2          :  3.1.0 (OK)
IPython >=7.6.0               :  7.29.0 (OK)
jedi >=0.17.2;<0.19.0         :  0.18.0 (OK)
jellyfish >=0.7               :  0.8.9 (OK)
jsonschema >=3.2.0            :  4.2.1 (OK)
keyring >=17.0.0              :  23.2.1 (OK)
nbconvert >=4.0               :  6.3.0 (OK)
numpydoc >=0.6.0              :  1.1.0 (OK)
parso >=0.7.0;<0.9.0          :  0.8.2 (OK)
pexpect >=4.4.0               :  4.8.0 (OK)
pickleshare >=0.4             :  0.7.5 (OK)
psutil >=5.3                  :  5.8.0 (OK)
pygments >=2.0                :  2.10.0 (OK)
pylint >=2.5.0                :  2.9.6 (OK)
pyls_spyder >=0.4.0           :  0.4.0 (OK)
pylsp >=1.2.2;<1.3.0          :  1.3.0.dev0 (OK)
pylsp_black >=1.0.0           :  None (OK)
qdarkstyle =3.0.2             :  3.0.2 (OK)
qstylizer >=0.1.10            :  0.2.1 (OK)
[...]

PR #17553, opened 2022-03-25
PR: Pass new environment variable to the kernel and remove some benign errors thanks to debugpy 1.6.0 (IPython console)

### LINKED ISSUE #17552: IPython kernel error with debugpy 1.6.0
`debugpy` was updated to 1.6.0 22 hours ago and is used by `ipykernel` and results in the following error in the IPython Console.
Screen Shot 2022-03-25 at 12 07 33 PM

This is an evolution of the problem in #16828. I'll provide a fix in a subsequent PR.

mrclary:
## Description of Changes

- Removed relevant errors from the benign error filter in `ipythonconsole` plugin
- Added the environment variable `PYDEVD_DISABLE_FILE_VALIDATION` to IPython Console 

The advent of `debugpy=1.6.0` provided an environment variable to bypass the errors encountered in #16927, #16828, and #17552.

Currently, only 1.5.2 is available on conda-forge; when 1.6.0 is available, I'll mark as ready for review.

### Issue(s) Resolved

Fixes #17552

### Affirmation

### DIFF
--- a/installers/macOS/req-build.txt
+++ b/installers/macOS/req-build.txt
@@ -2,4 +2,3 @@
 py2app>=0.27
 dmgbuild>=1.4.2
 setuptools<61.0.0  # remove when there is a better fix #17547
-debugpy<1.6.0  # remove when fix available for #17552
--- a/spyder/plugins/ipythonconsole/utils/kernelspec.py
+++ b/spyder/plugins/ipythonconsole/utils/kernelspec.py
@@ -198,9 +198,15 @@ class SpyderKernelSpec(KernelSpec, SpyderConfigurationAccessor):
             env_vars['SPY_RUN_CYTHON'] = True
 
         # App considerations
-        if (running_in_mac_app() or is_pynsist()) and not default_interpreter:
-            env_vars.pop('PYTHONHOME', None)
-            env_vars.pop('PYTHONPATH', None)
+        if (running_in_mac_app() or is_pynsist()):
+            if default_interpreter:
+                # See spyder-ide/spyder#16927
+                # See spyder-ide/spyder#16828
+                # See spyder-ide/spyder#17552
+                env_vars['PYDEVD_DISABLE_FILE_VALIDATION'] = 1
+            else:
+                env_vars.pop('PYTHONHOME', None)
+                env_vars.pop('PYTHONPATH', None)
 
         # Remove this variable because it prevents starting kernels for
         # external interpreters when present.
--- a/spyder/plugins/ipythonconsole/widgets/client.py
+++ b/spyder/plugins/ipythonconsole/widgets/client.py
@@ -521,14 +521,6 @@ class ClientWidget(QWidget, SaveHistoryMixin, SpyderWidgetMixin):
     def is_benign_error(self, error):
         """Decide if an error is benign in order to filter it."""
         benign_errors = [
-            # See spyder-ide/spyder#16828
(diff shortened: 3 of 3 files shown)


mrclary:
@ccordoba12 @dalthviz should we also require `debugpy>=1.6.0`? We don't have an explicit requirement on this right now, so I'm not sure what we want to do if users still have 1.5.1.

mrclary:
@ccordoba12, is there a reason why we are not testing Python 3.9 and conda for macOS? Just curious.

ccordoba12:
Because Github only offers 5 Mac Github Action slots per organization on the free plan, so we can spend them all in Spyder.

I think an option could be to run our tests only in Python 3.9 (instead of 3.7 and 3.9), so we can reduce the amount of slots we use per pull request. We'll take a look at that after 5.3.0 is released.

mrclary:
> Because Github only offers 5 Mac Github Action slots per organization on the free plan, so we can spend them all in Spyder.
> 
> I think an option could be to run our tests only in Python 3.9 (instead of 3.7 and 3.9), so we can reduce the amount of slots we use per pull request. We'll take a look at that after 5.3.0 is released.

I see, thanks for the explanation. So that means we use 4 right now, correct? 3.7 pip slow, 3.7 pip fast, lite build, and full build? I have been considering for a while whether we really needed to make Lite builds for every pull request. If we limited pull requests to just the full builds, that may help; on releases, we can have both full and lite. But as you say, we can discuss after 5.3.0.

ccordoba12:
> So that means we use 4 right now, correct? 3.7 pip slow, 3.7 pip fast, lite build, and full build?

Right.

> I have been considering for a while whether we really needed to make Lite builds for every pull request

Well, the nice thing about the Lite builds is that they are like half-size, so faster to download, install and test. I use the Windows Lite one all the time.

ccordoba12 (approved):
Thanks @mrclary for your help with this!

mrclary:
@ccordoba12, what about requiring `debugpy>=1.6.0`? I think that this PR may cause problems for users that are still using 1.5.1. 1.6.0 is not yet available on conda-forge, either.

dalthviz:
> @ccordoba12 @dalthviz should we also require debugpy>=1.6.0?

If the env var is available since debugpy 1.6.0 then I would say so. Probably the requirement should be done over spyder-kernels, right? and then we change the minimum version for spyder-kernels over Spyder before doing the release.

mrclary:
Upon further consideration, I don't think we should require `debugpy>=1.6.0`. First, 1.6.0 is not yet available on conda-forge, so we cannot set the requirement until it becomes available. Second, while it is true that the environment variable is not recognized by `debugpy<1.6.0`, it will do no harm there either. Third, the particular benign errors addressed are exclusive to the standalone applications for which 1.6.0 will be the default version anyway. In short, it doesn't matter which version users have installed because the only contexts that matter (standalone apps) will use 1.6.0, so I don't see a need to add another requirement to either `spyder` or `spyder-kernels`.

ccordoba12:
> Third, the particular benign errors addressed are exclusive to the standalone applications for which 1.6.0 will be the default version anyway. In short, it doesn't matter which version users have installed because the only contexts that matter (standalone apps) will use 1.6.0, so I don't see a need to add another requirement to either spyder or spyder-kernels.

Right, that's what I thought too (and sorry for letting that unanswered, I just missed the question).

PR #17646, opened 2022-04-08
PR: Ensure `QFont.setPointSize` receives an int argument (Shortcuts)

juliangilbey:
## Description of Changes

Thanks for your work on this version!  I'm close to getting it working for Debian....

The test `spyder/plugins/shortcuts/widgets/tests/test_summary.py::test_shortcutssummary` fails with Python 3.10 because `QFont().setPointSize()` requires an integer argument, and Python no longer performs implicit float -> integer conversions.  This patch fixes the cause.

### Affirmation

I certify the above statement is true and correct: juliangilbey

### DIFF
--- a/spyder/plugins/shortcuts/widgets/summary.py
+++ b/spyder/plugins/shortcuts/widgets/summary.py
@@ -45,7 +45,7 @@ class ShortcutsSummaryDialog(QDialog):
         # Calculate font and amount of elements in each column
         # according screen size
         width, height = self.get_screen_resolution()
-        font_size = height / 80
+        font_size = height // 80
         font_size = max(min(font_size, MAX_FONT_SIZE), MIN_FONT_SIZE)
         shortcuts_column = (height - 8 * font_size) / (font_size +16)
 
--- a/spyder/plugins/shortcuts/widgets/summary.py
+++ b/spyder/plugins/shortcuts/widgets/summary.py
@@ -45,7 +45,7 @@ class ShortcutsSummaryDialog(QDialog):
         # Calculate font and amount of elements in each column
         # according screen size
         width, height = self.get_screen_resolution()
-        font_size = height // 80
+        font_size = int(round(height / 80))
         font_size = max(min(font_size, MAX_FONT_SIZE), MIN_FONT_SIZE)
         shortcuts_column = (height - 8 * font_size) / (font_size +16)
 


dalthviz:
Note: This depends on https://github.com/spyder-ide/spyder/pull/17641 for the tests to pass

juliangilbey:
Ooh, I have done a rebase on a pull request before, but I don't remember exactly how; a quick reminder would be helpful, so that I don't mess this up!

dalthviz:
Sure! So you will need to have/set the original Spyder repo as a new remote like for example `upstream` and fetch it. So something like:

    git remote add upstream https://github.com/spyder-ide/spyder.git
    git fetch upstream

Then be sure to have your branch checkout and rebase using the `5.x` from the new `upstream` remote. So something like this:
    
    git checkout make_point_size_integer
    git rebase upstream/5.x

Finally force push the changes with something like:

    git push -f

juliangilbey:
OK, thanks!  I thought it was as simple as that, but didn't want to force-push without being sure :smile:   Have done it now - all being well, the tests will pass!

spyder/plugins/shortcuts/widgets/summary.py
```diff
@@ -45,7 +45,7 @@ def __init__(self, parent=None):
         # Calculate font and amount of elements in each column
         # according screen size
         width, height = self.get_screen_resolution()
-        font_size = height / 80
+        font_size = height // 80
```
ccordoba12, line 48:
```suggestion
        font_size = int(height / 80)
```

I think this is the right fix. Otherwise we could end up using the wrong font size below.

ccordoba12 (commented):
Thanks for your help with this @juliangilbey! I left a small comment for you, otherwise looks good for me.

spyder/plugins/shortcuts/widgets/summary.py
juliangilbey: I'm not sure I understand what you mean.  `height // 80` is equivalent to `math.floor(height / 80)`, but `int(height / 80)` might do rounding or truncation, depending on the behaviour of the C library (see note (3) to the table in https://docs.python.org/3/library/stdtypes.html#numeric-types-int-float-complex).  (Note (1) of that table says that `x // y` is an integer value but not necessarily of type `int`; in the case that both operands are of type `int`, as in the situation here, the result is of type `int`.)  It seems to make sense here to be explicit as to what is wanted: if truncation/floor is wanted, then `height // 80` does the job, as would `math.floor(height / 80); if rounding is preferred, then `int(round(height / 80))` would be needed.

ccordoba12: >  if rounding is preferred, then int(round(height / 80)) would be needed

Ok, I changed my suggestion accordingly.

ccordoba12: Let's go with `int(round(height / 80))`.

juliangilbey: OK, just pushed a commit making this change.

ccordoba12 (approved):
Thanks for you help with this @juliangilbey!

PR #17654, opened 2022-04-11
PR: Proof-of-principle environment/package manager

mrclary:
This is a proof-of-principle attempt to
* create virtual environments using Spyder's internal Python framework
* install external plugins _after_ bundle creation

This should provide the underlying mechanisms for a future environment/package manager plugin for Spyder.


ccordoba12:
@mrclary, this seems very interesting! My thoughts:

> create virtual environments using Spyder's internal Python framework

I think we should use [Micromamba](https://github.com/mamba-org/mamba/releases) for this. That's a single C++ binary, able to create conda envs and install packages from conda-forge quite quickly. As you know, Conda is way superior to pip for scientific development. And we could simply add the micromamba binaries to our installers, so they can provide env/package management out of the box.

> install external plugins after bundle creation

I think this requires extending the `PYTHONPATH` of our apps and probably provide a pip version that can only install Spyder plugins (so that people can't use it to install anything that they want).

mrclary:
> > create virtual environments using Spyder's internal Python framework
> 
> I think we should use [Micromamba](https://github.com/mamba-org/mamba/releases) for this. That's a single C++ binary, able to create conda envs and install packages from conda-forge quite quickly. As you know, Conda is way superior to pip for scientific development. And we could simply add the micromamba binaries to our installers, so they can provide env/package management out of the box.

I'll look into that. That would make it much easier to implement. What about packages that aren't available on conda? Can pip be used with this as well?

I don't know what your vision is, but I envision a future where there is no more "Same as Spyder" environment, rather a default (external) environment would be created (possibly in Spyder's config directory) which the user could customize through a dedicated Spyder interface.

> > install external plugins after bundle creation
> 
> I think this requires extending the `PYTHONPATH` of our apps and probably provide a pip version that can only install Spyder plugins (so that people can't use it to install anything that they want).

This would definitely be a "walled garden" scenario. I wanted to test the technical feasibility of including pip in the installer and being able to install packages after the app was created. So far, I think it will work just fine out-of-the-box. Ultimately, for our implementation, we'd have an interface that limited the actions that the user could perform, i.e. a fixed list of possible (external) plugins to install, so that it would not be possible for them to mess with the environment. Under the hood, a "frozen" package list, determined at build time on CI, would restrict what Spyder is allowed to install into its runtime environment.

ccordoba12:
> What about packages that aren't available on conda? Can pip be used with this as well?

Mmmh, good point, Micromamba can't handle that. But we could start with Micromamba first and then add support for pip afterwards.

> I envision a future where there is no more "Same as Spyder" environment, rather a default (external) environment would be created

That's a good idea but I think for people that only want to try Spyder and see how it works, or that only needs the libraries we already provide in our installers, it'd be overkill.

> I wanted to test the technical feasibility of including pip in the installer and being able to install packages after the app was created. So far, I think it will work just fine out-of-the-box.

Sure. The problem is `pip` can be accessed through the IPython console, so we'd need to prevent that.

> Ultimately, for our implementation, we'd have an interface that limited the actions that the user could perform, i.e. a fixed list of possible (external) plugins to install, so that it would not be possible for them to mess with the environment

Yes, but my idea is not install plugins in the same Python environment that the app lives because third-party plugins could break or users couldn't have permissions to do it. Instead, we should use something like `pip --user` to install in a special directory in `~/.spyder`.

mrclary:
> > What about packages that aren't available on conda? Can pip be used with this as well?
> 
> Mmmh, good point, Micromamba can't handle that. But we could start with Micromamba first and then add support for pip afterwards.

👍  

> > I envision a future where there is no more "Same as Spyder" environment, rather a default (external) environment would be created
> 
> That's a good idea but I think for people that only want to try Spyder and see how it works, or that only needs the libraries we already provide in our installers, it'd be overkill.

Currently, users download the application and launch Spyder; they don't initially know or care about an environment. I don't wish to change that. I'm proposing that the current "Same as Spyder" environment be _replaced_ with an environment managed by the new environment/package manager. This would be created either within the app at build time (but separate from Spyder's runtime environment) or automatically at first launch (if necessary, perhaps placed in Spyder's config dir?). Regardless of what happens under the hood, the user should not be able to tell the difference between current behavior and the new behavior, except that he should be able to install additional packages if desired.

> > I wanted to test the technical feasibility of including pip in the installer and being able to install packages after the app was created. So far, I think it will work just fine out-of-the-box.
> 
> Sure. The problem is `pip` can be accessed through the IPython console, so we'd need to prevent that.

With my proposal for the "default" environment, this becomes moot. Users cannot mess with Spyder's runtime environment. There would no longer be any exposure of Spyder's packages to the user in any way.

> 
> > Ultimately, for our implementation, we'd have an interface that limited the actions that the user could perform, i.e. a fixed list of possible (external) plugins to install, so that it would not be possible for them to mess with the environment
> 
> Yes, but my idea is not install plugins in the same Python environment that the app lives because third-party plugins could break or users couldn't have permissions to do it. Instead, we should use something like `pip --user` to install in a special directory in `~/.spyder`.

Hmm, yes. This may be a better idea. Regardless of the install location, the plugins and their dependencies would have to be compatible with Spyder's runtime packages, so we would want to maintain full control over the install process and just have users select whether they want to install it or not.

Thanks for your valuable input, @ccordoba12. I'm certainly not committed to any particular implementation. I was just exploring possibilities. We could further discuss development strategies at the developer meetings.

mrclary:
### Proposed User Experience
The following is a list of proposed features that Spyder should have, independent of _how_ each are implemented. Note that the first two items are already satisfied.

1. The user should have a default environment available upon first launch of Spyder.
2. The user should be able to use any externally created environments (pyenv, venv, conda, mamba, etc.).
3. The user should be able to install packages into the default environment using Spyder's environment manager.
5. The user should be able to reset/rebuild the default environment without interrupting Spyder.
6. The user should be able to create additional environments using Spyder's environment manager.
7. The user's customization to the default environment should persist after updating Spyder.
9. The user should be able to install/uninstall external plugins after installing Spyder.
10. The user's external plugin installation state should persist after updating Spyder.

mrclary:
### Proposed Implementation

- Spyder's runtime environment should be isolated from any IPython Console (user) environments, including the "default" environment.
- Spyder will use micromamba under the hood for creating/managing the default environment and environments created by the user using Spyder's environment manager.
- At first launch, Spyder will check for the existence of a `micromamba` installation and the default environment. If `micromamba` is not installed or no environment detected, Spyder will install micromamba and create the default environment.
- Spyder will maintain environment configuration files (yaml) in Spyder's config directory for the default environment and any user created environments.
- Spyder plugins will be installed by micromamba in a dedicated user-site directory in Spyder's config directory.

PR #17692, opened 2022-04-15
PR: Allow micromamba external environments.

### LINKED ISSUE #17665: IPython console does not work with Micromamba environments

Using a `micromamba` environment is currently not permitted.

```python
/Users/rclary/Applications/Spyder.app/Contents/Resources/lib/python3.9/spyder/plugins/ipythonconsole/scripts/conda‑activate.sh: line 11: /Users/rclary/opt/micromamba/bin/activate: No such file or directory
```

## Versions

* Spyder version: 5.3.0 
* Python version: 3.9.5
* Qt version: 5.15.2
* PyQt5 version: 5.15.6
* Operating System: Darwin 20.6.0

### Dependencies

```
# Mandatory:
applaunchservices >=0.1.7            :  0.2.1 (OK)
atomicwrites >=1.2.0                 :  1.4.0 (OK)
chardet >=2.0.0                      :  4.0.0 (OK)
cloudpickle >=0.5.0                  :  2.0.0 (OK)
cookiecutter >=1.6.0                 :  1.7.3 (OK)
diff_match_patch >=20181111          :  20200713 (OK)
intervaltree >=3.0.2                 :  3.1.0 (OK)
IPython >=7.31.1;<8.0.0              :  7.32.0 (OK)
jedi >=0.17.2;<0.19.0                :  0.18.1 (OK)
jellyfish >=0.7                      :  0.9.0 (OK)
jsonschema >=3.2.0                   :  4.4.0 (OK)
keyring >=17.0.0                     :  23.5.0 (OK)
nbconvert >=4.0                      :  6.4.5 (OK)
numpydoc >=0.6.0                     :  1.2.1 (OK)
parso >=0.7.0;<0.9.0                 :  0.8.3 (OK)
pexpect >=4.4.0                      :  4.8.0 (OK)
pickleshare >=0.4                    :  0.7.5 (OK)
psutil >=5.3                         :  5.9.0 (OK)
pygments >=2.0                       :  2.11.2 (OK)
pylint >=2.5.0                       :  2.13.3 (OK)
pyls_spyder >=0.4.0                  :  0.4.0 (OK)
pylsp >=1.4.1;<1.5.0                 :  1.4.1 (OK)
pylsp_black >=1.2.0                  :  1.2.0 (OK)
qdarkstyle >=3.0.2;<3.1.0            :  3.0.3 (OK)
qstylizer >=0.1.10                   :  0.2.1 (OK)
qtawesome >=1.0.2                    :  1.1.1 (OK)
qtconsole >=5.3.0;<5.4.0             :  5.3.0 (OK)
qtpy >=2.0.1                         :  2.0.1 (OK)
rtree >=0.9.7                        :  0.9.7 (OK)
setuptools >=49.6.0                  :  60.10.0 (OK)
sphinx >=0.6.6                       :  4.5.0 (OK)
spyder_kernels >=2.3.0;<2.4.0        :  2.3.0 (OK)
textdistance >=4.2.0                 :  4.2.2 (OK)
three_merge >=0.1.1                  :  0.1.1 (OK)
watchdog >=0.10.3                    :  2.1.7 (OK)
zmq >=17                             :  22.3.0 (OK)

# Optional:
cython >=0.21                        :  0.29.28 (OK)
matplotlib >=3.0.0                   :  3.5.1 (OK)
numpy >=1.7                          :  1.22.3 (OK)
pandas >=1.1.1                       :  1.4.1 (OK)
scipy >=0.17.0                       :  1.8.0 (OK)
sympy >=0.7.3                        :  1.10.1 (OK)

# Spyder plugins:
spyder_terminal.terminalplugin 1.2.2 :  1.2.2 (OK)
```

mrclary:
## Description of Changes

Updated `conda-activate.sh` to accommodate micromamba environments.

### Issue(s) Resolved

Fixes #17665

### Affirmation


mrclary:
@dalthviz, can you check that the `conda-activate.bat` file is appropriate?

mrclary:
I've checked functionality on my Windows VM, but I'd like another validation.

mrclary:
https://mamba.readthedocs.io/en/latest/installation.html#windows

Note that these instructions are not too verbose. I created a `C:\Users\<user>\Micromamba` directory, at the same level as my `Miniconda` installation location, and unzipped everything there. The instructions set variables for PowerShell profiles only, so I don't think it would work for Spyder out-of-the-box; so I also added `MAMBA_EXE` and `MAMBA_ROOT_PREFIX` to my environment variables.

mrclary:
I'm not experienced with Windows batch scripting, so if there's anything that needs improvement there, please let me know.

dalthviz:
Checking locally on Windows and I'm getting a `"starting"` output and importing from the env gets the correct package so seems like is working :+1:

imagen

However, as @mrclary said I had to manually set the `MAMBA_EXE` and `MAMBA_ROOT_PATH` env vars so maybe we should add some documentation regarding this support or some instructions to setup correctly micromamba along side Spyder in the docs?

spyder/plugins/ipythonconsole/scripts/conda-activate.bat
```diff
@@ -10,8 +10,23 @@ set SPYDER_KERNEL_SPEC=%4
...
+)
+
+IF DEFINED MAMBA_EXE (
+    :: Activate micromamba environment
+    for /f %%i in ('%MAMBA_EXE% shell activate %CONDA_ENV_PATH%') do set SCRIPT=%%i
+    call %SCRIPT%
+    goto start
+) ELSE (
+    exit 1
+)
```
dalthviz, line 26: Maybe we could leave this just as `exit 1` since there is a `goto` instructions inside both IF statements?

spyder/plugins/ipythonconsole/scripts/conda-activate.bat
```diff
@@ -10,8 +10,23 @@ set SPYDER_KERNEL_SPEC=%4
...
+
 
 :: Start kernel
+:start
+echo "starting"
```
dalthviz, line 31: I think we can remove this `echo` since it propbably was just for testing purposes, right?

mrclary:
> However, as @mrclary said I had to manually set the `MAMBA_EXE` and `MAMBA_ROOT_PATH` env vars so maybe we should add some documentation regarding this support or some instructions to setup correctly micromamba along side Spyder in the docs?

That may be a good idea...

ccordoba12:
> However, as @mrclary said I had to manually set the MAMBA_EXE and MAMBA_ROOT_PATH env vars so maybe we should add some documentation regarding this support or some instructions to setup correctly micromamba along side Spyder in the docs?

I think that's too much to ask from users. Instead, we should auto-detect the Mamba executable, the same way that we're doing it right now for Conda.

mrclary:
Yeah, that is kind of the issue. Right now, there's no way to detect the executable except to use MAMBA_EXE.

ccordoba12:
Here seems to be some default paths when you install Mambaforge:

https://github.com/conda-forge/miniforge/blob/db1fdcd16b497dea731170c8f2bc9d1fff675af9/scripts/build.sh#L40-L53

mrclary:
I don't think there is an issue with mamba installs, only with micromamba.

- For micromamba installation on all three platforms, downloading the `tar.bz` file is required.
- The instructions do not provide guidance on the root directory for extraction but do use the relative extraction location of `bin/microconda` for Unix and `Library\bin\micromamba.exe` for Windows. So this could be extracted anywhere on the user's system: either `<some path>/bin/micromamba` or `<some path>\Library\bin\micromamba.exe`. However, these are specified by the user at the extraction command, so even the relative location could be different.
- Indeed, the instructions for macOS, but not Linux, instruct moving the executable from `/bin/micromamba` to `./micromamba` and for Windows from `Library\bin\micromamba.exe` to `micromamba.exe`. So there seems to be some inconsistency between Unix platforms.
- Furthermore, the instructions for all platforms instruct setting `MAMBA_ROOT_PREFIX` to an arbitrary value: `/some/prefix` for Unix and `C:\Your\Root\Prefix` for Windows. This environment variable has no impact on the executable, but specifies the root directory in which to place created environments (if only the environment name is provided). Thus even a "standard" micromamba environment will not reside in a standard location relative to the micromamba executable, even if the user follows the installation instructions precisely, as there is no guidance whatsoever about the location of the root prefix.
- Finally, the installation instructions provide guidance on initializing the shell and shell hooks. If followed by the user, these will set `MAMBA_EXE` and `MAMBA_ROOT_PREFIX` in the appropriate shell startup file for Unix platforms _according to the values set by the user during the previous steps_, so there is no standard enforced there, but at least then the environment variables can be expected to be available. However, the shell init script on Windows only sets these values in the PowerShell profile, not the system environment variables. *Thus, on Windows, it is virtually impossible for Spyder to determine where the executable is located, even when the environment path is provided, without `MAMBA_EXE` set in the user's environment variables.*

mrclary:
@ccordoba12, I think I may have better solution than what is presented currently in this PR.

- Currently, `conda-activate.(sh|bat)` activates a conda environment then executes `python -m spyder_kernels.console` in that environment.
- The problem with this is that, for micromamba environments, the activation script is not in a known location relative to the environment.
- However, I don't think we really need to access the activation script for _any_ conda flavor environment. conda provides a `run` command which takes an environment path as an optional argument to the executable to run in that environment. I think, then, we can rely on the `CONDA_EXE` environment variable to launch our spyder-kernels in the appropriate environment, regardless of the conda flavor. This would also eliminate the need for `conda-activate.(sh|bat)` scripts, as we can do what we need in with a single `Popen` (or variant).

```python
Popen(['$CONDA_EXE', 'run', '-p', <env path>, 'python', '-m', 'spyder_kernels.console', '-f', <kernel file>], ...)
```

mrclary:
Okay, now I see the real issue.
`CONDA_EXE` is not available on Windows either, if launching from the installer. So, on Windows, installing any conda flavor _does not create environment variables_; they all create PowerShell profiles that have the necessary environment variables, e.g Anaconda PowerShell Prompt. Thus, there is not an issue when launching bootstrap or a conda installed Spyder: they are both started from a [Anaconda]PowerShell. The _only_ issue is with the Windows installer. This is also the root cause of #11819; the Windows standalone application has no reference for any conda executable and assuming an activation script location relative to the environment python executable skirts around this issue so long as there aren't any deviations from the assumption. This happens for micromamba, but could happen for other conda flavors too (though much less likely).

Conclusion: while micromamba does not seem to enforce a standard executable/activation location relative to the environment, I think that the OP is not actually an issue with micromamba, but the OP has just illuminated a limitation for the Windows standalone application.

Prescription: I think that eliminating the `conda-activate.(sh|bat)` scripts is still the prudent thing to do and rely on conda's own `run` command mechanism since it will handle the environment activation for us and will simplify the code. But we still need to address #11819. Two options are:
- Continue to assume an activation script located relative to the environment for Windows installer only, or
- Find a way to locate the root conda executable for Windows installer.

I think the latter is preferred, since it means we only have to find _one_ executable, independent of the environment, which makes it more robust and scalable, no matter where the user puts a conda environment.
One possibility here is to rely on the existence of `C:\Users\<users>\.conda\environments.txt`, the first line of which is the base conda environment; the conda executable, then should always be `Scripts\conda.exe` relative to the base environment.

mrclary:
Even better, Spyder already has `spyder.utils.conda.find_conda()`

mrclary:
It seems there is a problem with the interaction between `conda run` and `spyder_kernels`. For some reason, kernel interruption signals are captured by the `conda` process in addition to the kernel, and a kernel restart is triggered.

```python
import time; time.sleep(10)  # ctrl-c

  File "/var/folders/5v/28jqvwxs2cd5fj93gvwnykdrqc926z/T/ipykernel_2862/4263766136.py", line 1, in <cell line: 1>
    import time; time.sleep(10)

KeyboardInterrupt



CondaError: KeyboardInterrupt



Restarting kernel...
```
This is the reason that tests are failing. I can reproduce in bootstrap using any conda environment. However, I cannot reproduce using `qtconsole` from the command line
```
>> conda run --no-capture-output -p ~/opt/miniconda3/envs/spy-dev jupyter qtconsole
```
This has no issues with keyboard interrupt signals.
I wonder what would cause the difference for spyder-kernels...

ccordoba12:
@mrclary, in my view, this is mixing two things that don't belong in the same PR: adding support for Micromamba and changing the way we activate conda envs.

So I suggest just to stick to the Micromamba part here and leave the other one for `master` (i.e. the future Spyder 6). That would not only simplify what you're trying to do here, but avoid changing the activation process for now, which is probably is not really necessary if you follow my comment below.

Additionally, I suggest to add the Micromamba binary for Mac to `spyder/bin`, but only for the CI slot that builds the Mac installer (@dalthviz could take of the Windows installer later). Then you can instruct Spyder to use it from that location. That way users wouldn't need to set env vars to tell Spyder where it's located (*Note*: Since git doesn't allow to add empty dirs to the tree, you need to commit an empty file called `.keep` to `spyder/bin`).

ccordoba12:
> Including the Micromamba executable in the installers would also lay good ground work for the future environment manager.

Right, that's my idea too.

mrclary:
Well, a quick test on a local build and an abridged CI showed very good results.
Is there some reason why we should not or could not include the Micromamba executable as part of our repository? Not just the installers?

spyder/config/base.py
```diff
@@ -572,6 +572,22 @@ def running_in_mac_app(pyexec=None):
         return False
 
 
+# =============================================================================
+# Micromamba
+# =============================================================================
+def get_spy_umamba_path():
```
dalthviz, line 578: Maybe in this function we could validate directly with the `running_in_mac_app()` and `is_pynsist()` functions and in the `else` we could check if `MAMBA_EXE` existis and has a valid value. What do you think @mrclary ?

mrclary: That's fine. I would not check `MAMBA_EXE` here, since I think the purpose should be limited to returning Spyder's micromamba executable, if it exists. The `find_conda` function will check `MAMBA_EXE` for a micromamba executable on the system (not in Spyder).

mrclary:
@ccordoba12 @dalthviz Are we waiting for anything on this PR?

dalthviz:
From my side I'm missing to do a final test with the latest changes

dalthviz (commented):
Checking locally I'm getting an error/message when I have `MAMBA_EXE` but not `CONDA_EXE`:

`El sistema no puede encontrar la ruta especificada.` -> `The system cannot find the path specified`

imagen

Maybe is related to the `for /f` in the [conda-activate.bat](https://github.com/spyder-ide/spyder/pull/17692/files#diff-232f745444ede5e84394c2f650499bfe77a28ff55171a10a6b475fe992b8f662)

Other than that seems like the env is activated :+1:

mrclary:
> Checking locally I'm getting an error/message when I have `MAMBA_EXE` but not `CONDA_EXE`:
> 
> `El sistema no puede encontrar la ruta especificada.` -> `The system cannot find the path specified`

@dalthviz, I pushed up some changes that should fix this.

ccordoba12 (commented):
This looks pretty good to me, thanks @mrclary! I left a tiny suggestion for you, otherwise looks good to me.

ccordoba12:
> Is there some reason why we should not or could not include the Micromamba executable as part of our repository? Not just the installers?

There are several reasons I can think of for that:

* There's little gain in adding Micromamba for now to our conda/pip packages because there's no functionality tied to it. In other words, until we implement a frontend for Micromamba, this is still an experimental feature.
* Most probably, Linux package maintainers wouldn't take very well that we try to sneak an alien package manager as part of our package.
* Our mid term goal is to move as much people as possible to use our installers. And having a graphical, easy to use package manager only as part of them would be a quite powerful reason.

ccordoba12 (approved):
Thanks a lot @mrclary for your work on this!

PR #17788, opened 2022-04-28
Fix custom_interpreter option

maurerle:
until now, `custom_interpreter` option is never filled with the actual value.
This is fixed here.

Another possibility would be to remove the `custom_interpreter` variable at all.
If custom is set, the `custom_interpreter` is found in executable, else it is not of interest.
Old values are stored in the `custom_interpreters_list`, so they are unaffected.
The only actual usage of the `custom_interpreter` value exists in `get_main_interpreter`, which is never called too (but is a public API).

If would actually suggest to change `get_main_interpreter` code to

```
def get_main_interpreter(self):
    return self.get_conf('executable', section='main_interpreter')
```

and remove the `custom_interpreter`.

What do you think about that? The current situation seems to be wrong and it is not possible to see the custom_interpreter value in this variable.

## Description of Changes

### Issue(s) Resolved

Fixes #

### Affirmation


ccordoba12:
I don't think this is correct. I mean, `custom_interpreter` should be empty until the user selects one.

@dalthviz, what do you think?

dalthviz:
Following up https://github.com/spyder-ide/spyder/pull/17789#issuecomment-1113204709 I would say that we should create the `get_interpreter` method instead of changing the logic for the `custom_interpreter` option/config

maurerle:
As mentioned before:

- custom_interpreter is a variable to store the last custom_interpreter for the preferences dialog - it should not be used by other plugins to receive the current interpreter
- executable always contains the current executable and is already checked when set, so additional checks are not needed

The Variable flow is like this:

1. All listeners are executed, checking the availability of the last set executable
2. Opening Preferences - current value gets added to list of executables (`plugin._add_to_custom_interpreters(executable)`)
3. When applying a custom interpreter, the `on_conf_change` gets triggered for default/custom/custom_interpreter changes, updating the current executable if it needs an update:
```python
if executable != self.get_conf('executable'):
            self.set_conf('executable', executable)
```
4. the updated executable triggers the on_conf_change handler, which updates the status_widget and emits `sig_interpreter_changed` for other plugins

I think this is a clean solution and unifies the different mitigation approaches

spyder/plugins/profiler/widgets/main_widget.py
```diff
@@ -560,10 +560,7 @@ def start(self, wdir=None, args=None, pythonpath=None):
         if args:
             p_args.extend(shell_split(args))
 
-        executable = sys.executable
-        if executable.endswith("spyder.exe"):
-            # py2exe distribution
-            executable = "python.exe"
+        executable = self.get_conf('executable', section='main_interpreter')
```
ccordoba12, line 563: I see two options about this change: you could remove it and leave it for PR #17789, or leave it here and close that PR (since it's so small). It's up to you.

ccordoba12 (commented):
Hey @maurerle, thanks a lot for this! I thought about it a lot during the weekend and regard it now as a great improvement.

I left a small review for you. After you address it, I can give you a hand to fix our tests.

spyder/plugins/maininterpreter/confpage.py
```diff
@@ -48,30 +48,17 @@ def __init__(self, plugin, parent):
                 valid_custom_list.append(path)
         self.set_option('custom_interpreters_list', valid_custom_list)
 
-        # Python executable selection (initializing default values as well)
+        # add custom_interpreter to executable selection
         executable = self.get_option('executable', get_python_executable())
```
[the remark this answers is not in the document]
maurerle: If I see correctly, removing this line would remove `executable` variable from scope.
It is used to get the current active interpreter and add it to the dropdown menu in `_add_to_custom_interpreters` below?

spyder/plugins/profiler/widgets/main_widget.py
maurerle: I will leave it here and close the other PR

spyder/plugins/maininterpreter/container.py
```diff
@@ -53,31 +53,31 @@ def update_actions(self):
...
+            executable = get_python_executable()
+        else:
             executable = osp.normpath(self.get_conf('custom_interpreter'))
-            if (option in ['custom', 'custom_interpreter'] and
-                    osp.isfile(executable)):
+            if osp.isfile(executable):
                 self.sig_add_to_custom_interpreters_requested.emit(executable)
+            else:
+                executable = get_python_executable()
+                self.set_conf('custom_interpreter', ' ')
```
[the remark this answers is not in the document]
maurerle: I fully agree. Thanks!

I thought that it could have the use of also preventing programatically changes to the variable, but I hope this is not needed too.

spyder/plugins/maininterpreter/confpage.py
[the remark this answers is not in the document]
ccordoba12: You're right, my mistake. Then please change this to be

```python
executable = self.get_option('executable')
```

That's because the second parameter of `get_option` is a default value. But that's not necessary here since that value is set in `get_main_interpreter`

spyder/plugins/maininterpreter/container.py
```diff
@@ -53,31 +53,28 @@ def update_actions(self):
...
+        if ((option == 'default' and value) or
+                (option == 'custom' and not value)):
+            executable = get_python_executable()
+        else:
             executable = osp.normpath(self.get_conf('custom_interpreter'))
-            if (option in ['custom', 'custom_interpreter'] and
-                    osp.isfile(executable)):
-                self.sig_add_to_custom_interpreters_requested.emit(executable)
+            self.sig_add_to_custom_interpreters_requested.emit(executable)
+        # Setting executable option that will be used by other plugins in Spyder.
```
ccordoba12, line 62: Please add a blank line above this one, for clarity.

spyder/plugins/maininterpreter/container.py
```diff
@@ -53,31 +53,28 @@ def update_actions(self):
...
+        if executable != self.get_conf('executable'):
+                self.set_conf('executable', executable)
```
ccordoba12, line 64: Please indent correctly this line to 4 spaces (it seems it has 8 spaces of indentation).

ccordoba12 (commented):
@maurerle, thanks for addressing my previous suggestions. Second review now, much simpler than the first one.

ccordoba12:
> You can probably tell best how I can require it in the test?

So, we have two kind of tests: automatic tests based on Pytest and Pytest-Qt, and tests at the end of many files to quickly see if the widget in the file is displayed correctly.

The one that was failing with your changes was a file test and it happened because it doesn't assume that our config options are populated. And the fix was simple: trying to get the `executable` option with a default value before profiling a test script (as you can see in commit da69a18).

ccordoba12 (approved):
Thanks @maurerle for your contribution!! This is a great improvement!

PR #17789, opened 2022-04-28
PR: Set correct executable of profiler, when using non-default interpreter

### LINKED ISSUE #1884: Profiler plugin does not use non-default python interpreter
_From [breed...@gmail.com](https://code.google.com/u/105348566576685241380/) on 2014-07-08T17:28:48Z_

Spyder Version:  2.2.5
Python Version:  2.7.5
Qt Version    :  4.8.5, PyQt4 (API v2) 4.10.3 on Darwin
pyflakes >=0.5.0:  0.5.0 (OK)
pep8 >=0.6      :  1.4.6 (OK)
IPython >=0.13  :  1.1.0 (OK)
rope >=0.9.2    :  0.9.4 (OK)
sphinx >=0.6.6  :  1.1.3 (OK)
matplotlib >=1.0:  1.3.0 (OK)
sympy >=0.7.0   :  0.7.3 (OK)
pylint >=0.25   :  1.0.0 (OK) 

1. Specify the path to a non-default interpreter under Preferences -> Console -> Advanced Settings
   1. Run a script that has "print sys.executable"
   2. Profile that same script 

 When I run the script in the normal fashion, I get: /Users/breedlu/Library/Enthought/Canopy_64bit/User/bin/python
which is the interpreter I specified in step 1.

When I profile the script, and I click on the output button I get
/Applications/Spyder.app/Contents/MacOS/python
which is not interpreter I specified in step 1. 

.

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=1884_

maurerle:
## Description of Changes

I looked into this old community-issue and found that it is still reproducible.
The profiler always uses the sys.executable and not the configured one.
This is fixed by accessing the current executable from the config.

### Issue(s) Resolved

Fixes #1884

### Affirmation

### DIFF
--- a/spyder/plugins/completion/providers/languageserver/provider.py
+++ b/spyder/plugins/completion/providers/languageserver/provider.py
@@ -803,7 +803,7 @@ class LanguageServerProvider(SpyderCompletionProvider):
             environment = None
             env_vars = None
         else:
-            environment = self.get_conf('custom_interpreter',
+            environment = self.get_conf('executable',
                                         section='main_interpreter')
             env_vars = os.environ.copy()
             # external interpreter should not use internal PYTHONPATH
--- a/spyder/plugins/maininterpreter/confpage.py
+++ b/spyder/plugins/maininterpreter/confpage.py
@@ -48,30 +48,17 @@ class MainInterpreterConfigPage(PluginConfigPage):
                 valid_custom_list.append(path)
         self.set_option('custom_interpreters_list', valid_custom_list)
 
-        # Python executable selection (initializing default values as well)
+        # add custom_interpreter to executable selection
         executable = self.get_option('executable', get_python_executable())
-        if self.get_option('default'):
+
+        # the following checks are probably handled elsewhere, but they don't hurt either
+        if self.get_option('default') or not osp.isfile(executable):
             executable = get_python_executable()
+        elif not self.get_option('custom_interpreter'):
+            self.set_option('custom_interpreter', ' ')
 
-        if not osp.isfile(executable):
-            # This is absolutely necessary, in case the Python interpreter
-            # executable has been moved since last Spyder execution (following
-            # a Python distribution upgrade for example)
-            self.set_option('executable', get_python_executable())
-        elif executable.endswith('pythonw.exe'):
-            # That should not be necessary because this case is already taken
-            # care of by the `get_python_executable` function but, this was
-            # implemented too late, so we have to fix it here too, in case
-            # the Python executable has already been set with pythonw.exe:
-            self.set_option('executable',
-                            executable.replace("pythonw.exe", "python.exe"))
-
-        if not self.get_option('default'):
-            if not self.get_option('custom_interpreter'):
-                self.set_option('custom_interpreter', ' ')
-
-            plugin._add_to_custom_interpreters(executable)
-            self.validate_custom_interpreters_list()
+        plugin._add_to_custom_interpreters(executable)
+        self.validate_custom_interpreters_list()
 
     def initialize(self):
         super().initialize()
--- a/spyder/plugins/maininterpreter/container.py
+++ b/spyder/plugins/maininterpreter/container.py
@@ -53,31 +53,30 @@ class MainInterpreterContainer(PluginMainContainer):
 
     @on_conf_change(option=['default', 'custom_interpreter', 'custom'])
     def section_conf_update(self, option, value):
-        if option in ['default', 'custom_interpreter', 'custom'] and value:
-            self._update_status()
-            self.sig_interpreter_changed.emit()
-
-            # Set new interpreter
-            executable = osp.normpath(self.get_conf('custom_interpreter'))
-            if (option in ['custom', 'custom_interpreter'] and
-                    osp.isfile(executable)):
+        if (option == 'default' and value) or (option == 'custom' and not value):
+            executable = get_python_executable()
+        else:
+            executable = self.get_conf('custom_interpreter')
+            if osp.isfile(executable):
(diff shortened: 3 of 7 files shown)


jitseniesen:
As it happens, I looked into the same issue with the unittest plugin and came up with this code
```
if self.get_conf(section='main_interpreter', option='default'):
    executable = get_python_executable()
else:
    executable = self.get_conf(section='main_interpreter',
                               option='custom_interpreter')
```
My conclusion was that, contrary to what you say in PR #17788, the `custom_interpreter` config option does work for me and in fact it is always equal to the `executable` config option. However, if the user chooses the default Python interpreter in the options, then we should not use `custom_interpreter` or `executable` but use the function `get_python_executable()`.

I don't know why my experience is different from yours. One thing might be that I used the 5.x branch of Spyder, while you looked at master. Maybe something changed? If it did, it would be good to know.

I did not know that the widget also has `.get_conf()`, I will use that to simplify my code.

maurerle:
Hi, searching in the spyder code for 'custom_interpreter' only gives get_conf.
The only set_conf access of this config happens in
https://github.com/spyder-ide/spyder/blob/6d8472a75c9516393a94cb2579aa5c2eb155ef7b/spyder/plugins/maininterpreter/confpage.py#L71

Maybe I am missing something, but I just looked into 5.x and did not see something different than on master.
But you are right, that the custom_interpreter is somehow set correctly through the option param of create_file_combobox:
https://github.com/spyder-ide/spyder/blob/6d8472a75c9516393a94cb2579aa5c2eb155ef7b/spyder/plugins/maininterpreter/confpage.py#L108-L113

Somehow it still behaves weird.
Having an entry for a custom interpreter, and the interpreter set to default - I still get the custom_interpreter value when issuing `self.get_conf('executable', section='main_interpreter')` from a plugin..

maurerle:
Actually I think there should be an API which always returns the actually used interpreter for running code/profiler/unittest/... - I think this would be `Plugins.MainInterpreter.get_interpreter()`:

https://github.com/spyder-ide/spyder/blob/cdab0b3479b18178082233bb6abe8d8ba6724c51/spyder/plugins/maininterpreter/plugin.py#L103

It would be used like this:

```python
        maininterpreter = self.get_plugin(Plugins.MainInterpreter)
        executable = maininterpreter.get_interpreter()
```

Maybe thats the best way to go here?

ccordoba12:
> Actually I think there should be an API which always returns the actually used interpreter for running code/profiler/unittest/

This suggestion sounds very reasonable to me, so I agree with it.

@dalthviz, what do you think?

dalthviz:
>> Actually I think there should be an API which always returns the actually used interpreter for running code/profiler/unittest/

>This suggestion sounds very reasonable to me, so I agree with it.

That sounds good :+1: 

Just in case, I think the reason to have the `custom_interpreter` config value is to save and also be able to see in the preferences page the current selected custom interpreter. So, for example, even if I select the default interpreter option you can see in the custom interpreter dropdown which was the last custom selected value (I will have to check the code but I think that is the logic behind that):

imagen

jitseniesen:
> Actually I think there should be an API which always returns the actually used interpreter for running code/profiler/unittest/... - I think this would be `Plugins.MainInterpreter.get_interpreter()`:

It would be nice to have an API like that.

For my case, it would not be enough to have a function like `Plugins.MainInterpreter.get_interpreter()` because this function can't be called from a widget. There should also be a signal notifying changes in the interpreter, so that other plugins can be notified of any change and then use the `get_interpreter()` function to get the new interpreter and transmit that information to the plugin's widget.

maurerle:
What is speaking against using 
`executable = self.get_conf('executable', section='main_interpreter')`
as this can be used from "anywhere" - means, it is more flexible than calling the MainInterpreter plugin?

There are also many situations where 'executable' is assumed to always contain the currently used interpreter.
So it would be fitting.

Where should i implement the get_interpreter function otherwise?

dalthviz:
Checking, I think the `MainInterpreter` container has already a signal for that (`sig_interpreter_changed`):

https://github.com/spyder-ide/spyder/blob/ebc2587fad6d58e24f55e589173a921cbc08b1af/spyder/plugins/maininterpreter/container.py#L30-L33

Probably the `MainInterpreter` plugin should expose that?

Seems like the `Completion` plugin is using that signal already to handle the interpreter being updated:

https://github.com/spyder-ide/spyder/blob/ebc2587fad6d58e24f55e589173a921cbc08b1af/spyder/plugins/completion/plugin.py#L296-L306

maurerle:
As far as I understand the code:

1. maininterpreter executable always contains the current active executable used for code execution
2. many situations handle edge-cases multiple times and look into custom_executable which is not needed
3. status widget accesses the current interpreter to show it in the bottom bar (but it uses sys.executable directly, which is not always the used interpreter)
4. status widget checks updated interpreter and .conda_trash, which is not handled elsewhere?
5. maininterpreter executable is kept in config and 
6. pythonw and spyder.exe checks occur quite often, even when get_python_executable is used before as fallback
7. the current "public API" can only be called from a plugin. For now the "official way" is looking into the config, which isn't the best either

I updated my other pull request, as I think that this one only fixes the profiler executable

jitseniesen:
> 1. maininterpreter executable always contains the current active executable used for code execution

I think that when I tested, this is not true when the user uses the default Python; in that case, the `executable` config variable equals the custom interpreter, not the default one.

maurerle:
Yes exactly, i did fix this, as I think that this should be the behavior and allows to remove the other checks when acxessing the executable

maurerle:
Closing this in favor of #17788, which solves this issue and unifies access to the current used interpreter through `self.get_conf('executable', section='main_interpreter')`

PR #17854, opened 2022-05-10
PR: Wait for autoformat before returning on 'save' method (Editor)

### LINKED ISSUE #17836: Autoformat file on save prevents saving files
## Problem Description

Closing a modified file with `Autoformat file on save` active doesn't save the file

0. Activate `Autoformat file on save`
1. New untitled file
2. Modify file
3. Close file answering yes to save it

Saved file vs File not saved

dalthviz:
## Description of Changes

Since the autoformat option can take time we need to wait before returning on the editor `save` method to prevent lossing changes when closing an unsaved file.

Depends on https://github.com/spyder-ide/spyder/pull/17860

### Issue(s) Resolved

Fixes #17836 

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/editor.py
+++ b/spyder/plugins/editor/widgets/editor.py
@@ -22,8 +22,8 @@ import unicodedata
 # Third party imports
 import qstylizer.style
 from qtpy.compat import getsavefilename
-from qtpy.QtCore import (QByteArray, QFileInfo, QPoint, QSize, Qt, QTimer,
-                         Signal, Slot)
+from qtpy.QtCore import (QByteArray, QEventLoop, QFileInfo, QPoint, QSize, Qt,
+                         QTimer, Signal, Slot)
 from qtpy.QtGui import QFont, QTextCursor
 from qtpy.QtWidgets import (QAction, QApplication, QFileDialog, QHBoxLayout,
                             QLabel, QMainWindow, QMessageBox, QMenu,
@@ -1901,10 +1901,26 @@ class EditorStack(QWidget):
 
         try:
             if self.format_on_save and finfo.editor.formatting_enabled:
-                # Autoformat document and then save
+                # Wait for document autoformat and then save
+
+                # Waiting for the autoformat to complete is needed
+                # when the file is going to be closed after saving.
+                # See spyder-ide/spyder#17836
+                format_eventloop = QEventLoop(None)
+                format_timeout = QTimer(self)
+                format_timeout.setSingleShot(True)
+                format_timeout.timeout.connect(format_eventloop.quit)
+
                 finfo.editor.sig_stop_operation_in_progress.connect(
                     lambda: self._save_file(finfo))
+                finfo.editor.sig_stop_operation_in_progress.connect(
+                    format_timeout.stop)
+                finfo.editor.sig_stop_operation_in_progress.connect(
+                    format_eventloop.quit)
+
+                format_timeout.start(10000)
                 finfo.editor.format_document()
+                format_eventloop.exec_()
             else:
                 self._save_file(finfo)
             return True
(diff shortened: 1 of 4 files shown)


spyder/plugins/editor/widgets/editor.py
```diff
@@ -1901,10 +1901,26 @@ def save(self, index=None, force=False, save_new_files=True):
...
         try:
             if self.format_on_save and finfo.editor.formatting_enabled:
-                # Autoformat document and then save
+                # Wait for document autoformat and then save
+
+                # Waiting for the autoformat to complete is needed
+                # when the file is going to be closed after saving.
+                # See spyder-ide/spyder#17836
+                format_eventloop = QEventLoop(None)
+                format_timeout = QTimer(self)
```
ccordoba12, line 1910: What happens if the user tries to save two files in quick succession in the same editor stack? In that case, I think these two instances of QEventLoop and QTimer would be overwritten by the next save operation, which could cause a segfault.

To avoid that, perhaps they should be created at the CodeEditor level. Then, you could call them here like `finfo.editor.format_eventloop` and `finfo.editor.format_timeout`. Since it's not possible to save the same file twice or more times, then we would be safe.

ccordoba12 (commented):
Thanks @dalthviz for working on this!

Besides the initial review I left for you, please add a test to avoid a regression for this critical data loss when the editor's migration to the new API takes place.

dalthviz:
I think the test on conda are failing due to the PyQt 5.15.4 release on conda-forge. Maybe we are missing https://anaconda.org/conda-forge/pyqtwebengine as a dependency for the tests?

spyder/plugins/editor/widgets/tests/conftest.py
```diff
@@ -50,23 +50,82 @@ def codeeditor_factory():
...
+        editor.completion_widget.hide()
+        editor.tooltip_widget.hide()
+        editor.hide()
+
+        # Capture stderr and assert there are no errors
+        sys_stream = capsys.readouterr()
+        sys_err = sys_stream.err
+
+        if PY2:
+            sys_err = to_text_string(sys_err).encode('utf-8')
```
ccordoba12, line 121:
```suggestion
```

We don't support Python 2 anymore, so this is no longer necessary.

spyder/plugins/editor/widgets/tests/test_formatting.py
```diff
@@ -207,3 +208,45 @@ def test_max_line_length(formatter, completions_codeeditor, qtbot):
...
 
     assert code_editor.get_text_with_eol() == expected
+
+
+@pytest.mark.slow
+@pytest.mark.order(1)
+@pytest.mark.parametrize('formatter', [autopep8, black])
+def test_closing_document_formatting(
+        formatter, completions_editor, qtbot, monkeypatch):
+    """Validate text autoformatting via autopep8, yapf or black."""
```
ccordoba12, line 218:
```suggestion
    """Check that auto-formatting works when closing an usaved file."""
```

spyder/plugins/editor/widgets/tests/conftest.py
```diff
@@ -50,23 +50,82 @@ def codeeditor_factory():
...
+
+
+@pytest.fixture
+def completions_editor(
+        completion_plugin_all_started, qtbot_module, request, capsys,
+        tmp_path):
+    """Editorstack instance with LSP services activated."""
+    # Create a CodeEditor instance
+    editorstack, editor = editor_factory(new_file=False)
+    editorstack.set_format_on_save(True)
```
ccordoba12, line 87: Is it possible to move this line to the test that you created below? That way, other tests could reuse this fixture for other purposes.

ccordoba12 (commented):
Thanks @dalthviz for addressing my previous review! I left a much simpler one for you this time.

ccordoba12 (approved):
Thanks @dalthviz for solving this critical issue!

PR #17874, opened 2022-05-13
PR: Do not resolve symlink on Python interpreter file selection

### LINKED ISSUE #17872: Python interpreter file browser resolves symlinks
The Python interpreter file browser in the Preferences resolves symlinks and it should not.

Python executables in virtual environments frequently are symlinks to Python executables located elsewhere, and thus outside the virtual environment. When Spyder locates standard virtual environments (e.g. pyenv and conda) the path to the Python executable is correctly _unresolved_. However, when selecting the environment Python executable via the file browser, the symlink is resolved resulting in an incorrect environment when launched.

mrclary:
## Description of Changes

Send `QFileDialog.DontResolveSymlinks` option to `select_file` in `create_file_combobox` for Python interpreter preferences.
This ensures that symlinks are not resolved.

### Issue(s) Resolved

Fixes #17872 

### Affirmation

### DIFF
--- a/spyder/plugins/preferences/api.py
+++ b/spyder/plugins/preferences/api.py
@@ -19,10 +19,10 @@ from qtpy.compat import (getexistingdirectory, getopenfilename, from_qvariant,
 from qtpy.QtCore import Qt, Signal, Slot, QRegExp
 from qtpy.QtGui import QColor, QRegExpValidator, QTextOption
 from qtpy.QtWidgets import (QButtonGroup, QCheckBox, QComboBox, QDoubleSpinBox,
-                            QFontComboBox, QGridLayout, QGroupBox, QHBoxLayout,
-                            QLabel, QLineEdit, QMessageBox, QPushButton,
-                            QRadioButton, QSpinBox, QVBoxLayout, QWidget,
-                            QPlainTextEdit, QTabWidget)
+                            QFileDialog, QFontComboBox, QGridLayout, QGroupBox,
+                            QHBoxLayout, QLabel, QLineEdit, QMessageBox,
+                            QPlainTextEdit, QPushButton, QRadioButton,
+                            QSpinBox, QTabWidget, QVBoxLayout, QWidget)
 
 # Local imports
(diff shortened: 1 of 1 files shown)


ccordoba12 (commented):
This looks good to me. My only question is we should do this for all file dialogs present in Preferences (I'm not so sure about that).

@dalthviz, what do you think?

mrclary:
> This looks good to me. My only question is we should do this for all file dialogs present in Preferences (I'm not so sure about that).
> 
> @dalthviz, what do you think?

These are all the Select file or Select directory elements that I know of in Preferences.
- Completion and linting -> Snippits -> Import snippets from JSON.
- IPython console -> Startup -> Run a file -> Use the following file
- Python interpreter -> Python interpreter -> Use the following Python interpreter
- Run -> Working directory settings -> The following directory
- Working directory -> Startup -> The following directory
- Working directory -> New consoles -> The following directory

As far as I know, the only real issue with resolving symlinks occurs for the Python interpreter. So the only question is what is the best UX. Do users prefer or expect to see the file/directory path chosen or the fully resolved path? I suppose my preference would be to see the path chosen, not the fully resolved one; this is, after all, what I explicitly navigated to. But this is entirely a personal preference. Some may find it useful to see what the chosen path fully resolves to.

dalthviz:
I would say that if there is no issue with the other places we should keep them as they are. However as @mrclary says I think too that the issue to generalize the behavior implemented here is related with UX (is the user expecting their paths to be somehow resolved or not). Maybe we could open a follow up issue to discuss that and ask @isabela-pf or in general users their thoughs. An other option is to somehow make that choice a preference allowing users to select if they want or not their paths being resolved (and setting the default to 'do the resolve' since is the current behavior)

dalthviz (approved):
Thanks @mrclary ! LGTM :+1:

isabela-pf:
@dalthviz Follow up would be best because you shouldn't block on my input. I am not going to get back to this issue quickly enough. Thank you for the UX consideration!

PR #18106, opened 2022-06-01
PR: Fix tests due to Pylint 2.14.0 and some test stalling the CI or falling due to leak validations

dalthviz:
## Description of Changes

Test suite revision for failing tests due to:
* Pylint 2.14.0:  Fixes #18175 
* ~jupyter_client 7.3.2: #18174~ 
* ~`openpyxl` and `defusedxml` cause `nbconvert` NOK error on MacOS installers tests (probably `nconvert` is not importable after the addition of those dependencies into the installer build?): #18278~
* Some other failures (leak validations and tests stalling)

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct: dalthviz

PR #18278, opened 2022-06-13
PR: Fix issue where nbconvert was not importable in macOS application

mrclary:
`defusedxml` was already a dependency of `nbconvert`, so it was redundent to include in the scientific packages list for the full version, but this also caused it to be explicitly excluded for the lite version, hence the issue for `nbconvert`.

Related to #18106


ccordoba12 (approved):
Thanks @mrclary! I'm going to merge this one so that @dalthviz can pick it up on PR #18106.

PR #18304, opened 2022-06-15
PR: Fix syntax issues in macOS installer workflow file

mrclary:
#18120 had syntax issues in the workflow file that resulted in the build job failing to start without indicating failure.
This PR remedies that.

Additionally, pytests are not run if only installer files are changed.

### DIFF
--- a/.github/workflows/test-linux.yml
+++ b/.github/workflows/test-linux.yml
@@ -14,6 +14,8 @@ on:
       - '**.bat'
       - '**.py'
       - '**.sh'
+      - '!installers/**'
+      - '!.github/workflows/installer*.yml'
 
   pull_request:
     branches:
@@ -28,6 +30,8 @@ on:
       - '**.bat'
       - '**.py'
       - '**.sh'
+      - '!installers/**'
+      - '!.github/workflows/installer*.yml'
 
 jobs:
   build:
--- a/.github/workflows/installer-macos.yml
+++ b/.github/workflows/installer-macos.yml
@@ -8,7 +8,7 @@ name: Create macOS App Bundle and DMG
 
 jobs:
   matrix_prep:
-    name: Matrix Prep
+    name: Determine Build Matrix
     runs-on: macos-10.15
     outputs:
       build_type: ${{ steps.set-matrix.outputs.build_type }}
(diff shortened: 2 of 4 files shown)


.github/workflows/test-linux.yml
```diff
@@ -14,6 +14,8 @@ on:
       - '**.bat'
       - '**.py'
       - '**.sh'
+      - '!installers/**'
+      - '!.github/workflows/installer*.yml'
```
ccordoba12, line 18: I was thinking to do the same, i.e. not running our test suite when changes affect only the installer files. But it seems this is not working because the test workflows are still run.

Perhaps these exclusions should be at the top of `paths` to work?

mrclary: The pytest workflows were still run because _this_ PR also modified the pytest workflow files.

Per [GitHub documentation](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths),
>The order that you define patterns matters:
> - A matching negative pattern (prefixed with !) after a positive match will exclude the path.
> - A matching positive pattern after a negative match will include the path again.

So, they should remain at the bottom of the path list. Additionally, I successfully tested this on a PR in my forked repo by merging changes to the pytest workflow files first (i.e. ignore installer files), then on a subsequent PR modifying only the installer workflow files; then only the installer workflows ran.

Tangentially, I noticed that in the pytest workflow files for all platforms the `on: push:` and `on: pull_request:` directives are identical; if you don't anticipate these being different, then we can combine them into `on: [push, pull_request]`. See [this example](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#using-multiple-events).

ccordoba12: > The pytest workflows were still run because this PR also modified the pytest workflow files.

Ok, you're right.

> if you don't anticipate these being different, then we can combine them into on: [push, pull_request]

Sounds good to me. Please do that as part of this PR.

mrclary: Oops, nevermind. Looks like `on: [push, pull_request]` consolidation feature is not allowed if event type filters are used (i.e. `branch` and `path`). Each event must be [configured separately](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#using-activity-types-and-filters-with-multiple-events).

ccordoba12 (approved):
Thanks @mrclary!

PR #18308, opened 2022-06-16
PR: Add feature to append/prepend PPM to IPython Console sys.path

mrclary:
Add a checkbox to the PYTHONPATH Manager that allows the user to specify whether to append or prepend these paths to the IPython Console `sys.path`.

### Prerequisites
- [x] spyder-ide/spyder-kernels#378
- [x] #17512 
- [ ] #18778

Fixes: #17066


mrclary:
@ccordoba12, you mentioned using the `update_syspath` kernel handler. I wouldn't mind a quick overview via zoom when you have a moment. I see the basic functionality, as it is already used by the PPM on path updates. However, I think that we could completely eliminate `set_spyder_pythonpath` in `spyder_kernels/customize/spydercustomize.py` (and `SPY_PYTHONPATH`) by executing `update_syspath` upon client creation, but I'd want to know exactly the best signals to send and use for accomplishing this.

CAM-Gerlach:
>  However, I think that we could completely eliminate set_spyder_pythonpath in spyder_kernels/customize/spydercustomize.py (and SPY_PYTHONPATH) by executing update_syspath upon client creation,

Just FYI, I think that's what @ccordoba12 was saying as well over on the other issue:

> By the way, that would make useless the code you changed here in spydercustomize.py, so you could remove it.

mrclary:
@ccordoba12 @dalthviz, I think that @goanpeca began to use the modern configuration system for the PPM paths about two years ago, but there are still references to individual `path` and `not_active_path` files in the configuration directory. So I think some cleanup may be necessary here. However, I'm confused about the roles of `spyder.ini` vs. `transient.ini`. Could you explain their intended differentiated purpose and what configurations belong in which file?

dalthviz:
From my side I'm not sure to be honest 😅 However, I think the main difference between `spyder.ini` and `transient.ini` is that the config options at transient are usually loaded at startup or created at startup if not found and meant to keep the session state next time Spyder starts (so I think usually the options there are persisted just before Spyder closes). Probably @ccordoba12 knows better.

CAM-Gerlach:
As far as I understand, `transient.ini` is for ephemeral state that is not particularly critical and can be discarded without much loss, while the main `spyder.ini` is for persistent config options affirmatively set by the user, e.g. in the preferences dialog.

Issue #9051, opened 2019-03-28
Variable explorer viewers go to the background on macOS


lschneidpro:
### Issue Report Checklist

* [ ] Searched the [issues page](https://github.com/spyder-ide/spyder/issues?q=is%3Aissue) for similar reports
* [ ] Read the relevant sections of the [Spyder Troubleshooting Guide](https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ) and followed its advice
* [ ] Reproduced the issue after updating with ``conda update spyder`` (or ``pip``, if not using Anaconda)
* [ ] Could not reproduce inside ``jupyter qtconsole`` (if console-related)
* [ ] Tried basic troubleshooting (if a bug/error)
    * [ ] Restarted Spyder
    * [ ] Reset preferences with ``spyder --reset``
    * [ ] Reinstalled the latest version of [Anaconda](https://www.anaconda.com/download/)
    * [ ] Tried the other applicable steps from the Troubleshooting Guide
* [ ] Completed the **Problem Description**, **Steps to Reproduce** and **Version** sections below

## Problem Description
Hi there, using spyder on both windows and mac os, I noticed an odd behaviour compared on Mac OS comapred to the windows version.

On windows, I can easily open and explore different variables at the same time for comparison purpose. On Mac OS, when i open a second variable window, the first one would disapear behind spyder windows.

### What steps reproduce the problem?

1. Create at least two array, dataframe,list....
2. Inspect them with the variables explorer on Mac OS
3. 

### Paste Traceback/Error Below (if applicable)

```python-traceback

PASTE TRACEBACK HERE

```

## Versions

* Spyder version: last one
* Python version: 
* Qt version: 
* PyQt version: 
* Operating System name/version: 

### Dependencies

```
PASTE DEPENDENCIES HERE
```

ccordoba12:
@goanpeca, is there something we can do about this?

goanpeca:
Will look into it

goanpeca:
@lschneidpro like this?

![varexp](https://user-images.githubusercontent.com/3627835/55661510-c962fd80-57d1-11e9-8901-10615fb286c2.gif)

goanpeca:
@ccordoba12 should this go to 3.x?

lschneidpro:
@goanpeca

yes, looks good

ccordoba12:
> should this go to 3.x?

Sure, it's a very simple fix.

[...]

PR #18397, opened 2022-06-22
PR: Show environment variables for all operating systems and modernize `spyder.utils.environ` module

mrclary:
## What this PR does

- Provides a general tool for retrieving user environment variables: `spyder.utils.programs.get_user_environment_variables`. This returns a dictionary of string values.
- Updates `spyder.util.environ` classes and functions to be platform agnostic. Most notably, `WinUserEnvDialog` is renamed `UserEnvDialog`. The associated menubar item is now available on all platforms, though these are read-only, while Windows retains the ability to modify and export to the system environment.
- Removes Python 2.x dependence in `spyder.util.environ`.

## Motivation
#17512 Introduced a feature to import the user's `PYTHONPATH` environment variable into the PYTHONPATH Manager. This relies on Spyder inheriting this variable at runtime into `os.environ`. However, it may be more desirable to retrieve this environment variable on-demand rather than rely on its state at the time of Spyder launch.
See discussion comments: 
- https://github.com/spyder-ide/spyder/pull/17512#pullrequestreview-1015511299
- https://github.com/spyder-ide/spyder/pull/17408#issuecomment-1158428364

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -466,13 +466,8 @@ class MainWindow(QMainWindow, SpyderConfigurationAccessor):
             try:
                 context = '_'
                 name = 'switch to {}'.format(plugin.CONF_SECTION)
-                shortcut = self.get_shortcut(
-                    name,
-                    context,
-                    plugin_name=plugin.CONF_SECTION
-                )
             except (cp.NoSectionError, cp.NoOptionError):
-                shortcut = None
+                pass
 
             sc = QShortcut(QKeySequence(), self,
                            lambda: self.switch_to_plugin(plugin))
@@ -844,28 +839,30 @@ class MainWindow(QMainWindow, SpyderConfigurationAccessor):
 
         # Switcher shortcuts
         self.file_switcher_action = create_action(
-                                    self,
-                                    _('File switcher...'),
-                                    icon=ima.icon('filelist'),
-                                    tip=_('Fast switch between files'),
-                                    triggered=self.open_switcher,
-                                    context=Qt.ApplicationShortcut,
-                                    id_='file_switcher')
+            self,
+            _('File switcher...'),
+            icon=ima.icon('filelist'),
+            tip=_('Fast switch between files'),
+            triggered=self.open_switcher,
+            context=Qt.ApplicationShortcut,
+            id_='file_switcher'
+        )
         self.register_shortcut(self.file_switcher_action, context="_",
                                name="File switcher")
         self.symbol_finder_action = create_action(
-                                    self, _('Symbol finder...'),
-                                    icon=ima.icon('symbol_find'),
-                                    tip=_('Fast symbol search in file'),
-                                    triggered=self.open_symbolfinder,
-                                    context=Qt.ApplicationShortcut,
-                                    id_='symbol_finder')
+            self, _('Symbol finder...'),
+            icon=ima.icon('symbol_find'),
+            tip=_('Fast symbol search in file'),
+            triggered=self.open_symbolfinder,
+            context=Qt.ApplicationShortcut,
+            id_='symbol_finder'
+        )
         self.register_shortcut(self.symbol_finder_action, context="_",
                                name="symbol finder", add_shortcut_to_tip=True)
 
         def create_edit_action(text, tr_text, icon):
             textseq = text.split(' ')
-            method_name = textseq[0].lower()+"".join(textseq[1:])
+            method_name = textseq[0].lower() + "".join(textseq[1:])
             action = create_action(self, tr_text,
                                    icon=icon,
                                    triggered=self.global_callback,
@@ -901,10 +898,11 @@ class MainWindow(QMainWindow, SpyderConfigurationAccessor):
         ]
         for switcher_action in switcher_actions:
             mainmenu.add_item_to_application_menu(
-                    switcher_action,
-                    menu_id=ApplicationMenus.File,
-                    section=FileMenuSections.Switcher,
-                    before_section=FileMenuSections.Restart)
+                switcher_action,
+                menu_id=ApplicationMenus.File,
+                section=FileMenuSections.Switcher,
+                before_section=FileMenuSections.Restart
+            )
         self.set_splash("")
 
         # Toolbars
@@ -1324,7 +1322,7 @@ class MainWindow(QMainWindow, SpyderConfigurationAccessor):
     def update_edit_menu(self):
         """Update edit menu"""
         widget, textedit_properties = self.get_focus_widget_properties()
-        if textedit_properties is None: # widget is not an editor/console
+        if textedit_properties is None:  # widget is not an editor/console
             return
         # !!! Below this line, widget is expected to be a QPlainTextEdit
         #     instance
@@ -1343,10 +1341,10 @@ class MainWindow(QMainWindow, SpyderConfigurationAccessor):
         self.selectall_action.setEnabled(True)
 
         # Undo, redo
-        self.undo_action.setEnabled( readwrite_editor \
-                                     and widget.document().isUndoAvailable() )
-        self.redo_action.setEnabled( readwrite_editor \
-                                     and widget.document().isRedoAvailable() )
+        self.undo_action.setEnabled(readwrite_editor
+                                    and widget.document().isUndoAvailable())
+        self.redo_action.setEnabled(readwrite_editor
+                                    and widget.document().isRedoAvailable())
 
         # Copy, cut, paste, delete
         has_selection = widget.has_selected_text()
@@ -1369,7 +1367,7 @@ class MainWindow(QMainWindow, SpyderConfigurationAccessor):
             child.setEnabled(False)
 
         widget, textedit_properties = self.get_focus_widget_properties()
-        if textedit_properties is None: # widget is not an editor/console
+        if textedit_properties is None:  # widget is not an editor/console
             return
 
         # !!! Below this line, widget is expected to be a QPlainTextEdit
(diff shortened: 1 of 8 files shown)


CAM-Gerlach:
> So the primary motivation was the behavior noted by @dalthviz in which importing PYTHONPATH after exporting from PPM to the system PYTHONPATH indicated that there was nothing to import. 

Right, but as discussed in my reply over there and you also mention here, that is easily and better handled internally, since this change only happens externally due to internal action on Spyder's part, so there's no need to call out to the system. I was wondering if there are any other common cases that affect Spyder itself.

> This is essentially what is done in spyder.utils.environ, but for Windows only.

Well, not exactly, unless I'm misunderstanding—what it does is use registry access to retrieve them from the live registry, rather than doing anything related to opening a new Python interpreter and piping things back or accessing environment variables directly; in that sense the original approach you proposed is closer to what I suggested than this approach. To note, it _is_ a little more efficient since it doesn't require spawning a new Python subprocess, but for rarely-performed on-demand actions the cost is small (after all, we're doing more than that every minute checking for updated Anaconda envs) and the code is much less complex and more reliable, and of course supports any platform Python/Spyder does.

> Exactly. Given your clarification of the python flags, the only use case for obtaining the environment variables from the system rather than os.environ is for the user environment variable editor dialog spyder.utils.environ.WinUserEnvDialog.

Right, and in this case, for the reasons mentioned above, it might make more sense to replace the current complex logic with something closer to what I propose above (after suitable testing). Indeed, there seems to be no reason why it shouldn't be possible  to _view_ the current environment variables in an immutable variable explorer dialog on any platform.

mrclary:
@ccordoba12 @dalthviz, `spyder/plugins/editor/widgets/tests/test_introspection.py::test_completions_environment` is failing for this PR on pip on Windows and I'm not sure why. I reproduce the same exact error in my Windows VM _and_ macOS for _both_ this branch and 5.x using conda.

I see that this test has been updated recently, perhaps to address flakiness. Could this be a regression of some sort? Did this test exclude Linux and macOS for this same issue? Could Windows be having similar issues?

I verified that completions work when running Spyder locally from source and using the `jedi-test-env` for completions. 🤷🏼

dalthviz:
I think I also saw that test fail on the merge commits between 5.x and master recently. Not sure what could be causing the flakiness though :/

spyder/app/mainwindow.py
```diff
@@ -1465,7 +1467,8 @@ def resizeEvent(self, event):
     def moveEvent(self, event):
         """Reimplement Qt method"""
         if hasattr(self, 'layouts'):
-            if not self.isMaximized() and not self.layouts.get_fullscreen_flag():
+            if (not self.isMaximized()
+                    and not self.layouts.get_fullscreen_flag()):
```
ccordoba12, line 1471:
```suggestion
            if (
                not self.isMaximized()
                and not self.layouts.get_fullscreen_flag()
            ):
```

spyder/plugins/application/container.py
```diff
@@ -161,17 +157,18 @@ def setup(self):
...
-                      "variables in Windows registry "
-                      "(i.e. for all sessions)"),
-                triggered=self.show_windows_env_variables)
+        if os.name == 'nt':
+            tip = ("Show and edit current user environment variables in "
+                   "Windows registry (i.e. for all sessions)")
         else:
-            self.winenv_action = None
+            tip = ("Show current user environment "
+                   "variables (i.e. for all sessions)")
```
ccordoba12, line 165:
```suggestion
            tip = ("Show current user environment variables (i.e. for all "
                   "sessions)")
```

spyder/utils/environ.py
```diff
@@ -12,33 +12,74 @@
 import os
 
 # Third party imports
-from qtpy.QtWidgets import QDialog, QMessageBox
+from qtpy.QtWidgets import QMessageBox
+try:
+    import winreg
+except ImportError:
```
ccordoba12, line 18:
```suggestion
except Exception:
```

This could fail for reasons other than `ImportError` which would lead Spyder to crash at startup.

spyder/utils/environ.py
```diff
@@ -47,41 +88,23 @@ def clean_env(env_vars):
...
-    def __init__(self, environ, parent=None):
+    def __init__(self, environ, parent=None,
+                 title="Environment variables", readonly=True):
         super(RemoteEnvDialog, self).__init__(parent)
         try:
             self.setup(
                 envdict2listdict(environ),
-                title=_("Environment variables"),
-                readonly=True,
+                title=_(title),
```
ccordoba12, line 106: This doesn't work to translate strings, i.e. you can't use variables inside `_()` or strings for translations. You need to use a real strings here instead.

spyder/utils/environ.py
```diff
@@ -97,90 +120,57 @@ def __init__(self, environ, parent=None):
...
+                  "application from which you may have executed it, "
+                  "like <i>Python(x,y) Home</i> for example)")
+            )
+
+    def accept(self):
+        """Reimplement Qt method"""
+        if os.name == 'nt':
             set_user_env(listdict2envdict(self.get_value()), parent=self)
-            QDialog.accept(self)
+        super(UserEnvDialog, self).accept()
```
ccordoba12, line 162:
```suggestion
        super().accept()
```

This is the new, Python 3 way of using `super`.

spyder/utils/programs.py
```diff
@@ -347,6 +344,31 @@ def get_username():
     return username
 
 
+def get_user_environment_variables():
```
ccordoba12, line 347: Please add a test for this new function.

ccordoba12 (commented):
Thanks @mrclary for your work on this! It looks pretty good to me, except for the small comments and suggestions I left for you below.

Please also solve the conflict this PR has now with `5.x`.

spyder/widgets/collectionseditor.py
```diff
@@ -1512,12 +1516,7 @@ def setup(self, data, title='', readonly=False, remote=False,
         if icon is None:
             self.setWindowIcon(ima.icon('dictedit'))
 
-        if sys.platform == 'darwin':
-            # See spyder-ide/spyder#9051
-            self.setWindowFlags(Qt.Tool)
-        else:
-            # Make the dialog act as a window
-            self.setWindowFlags(Qt.Window)
```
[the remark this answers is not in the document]
mrclary: Sorry about that; I misunderstood spyder-ide/spyder#9051 and thought this fix was obsolete. I reverted.

spyder/utils/tests/test_environ.py
```diff
@@ -19,45 +18,46 @@
...
+        os_env.pop(k, None)
+    _os_env = sorted([k.lower() for k in os_env.keys()])
+
+    guev = get_user_environment_variables()
+    for k in ["QT_MAC_WANTS_LAYER", "prompt", "qt_d3dcreate_multithreaded"]:
+        # Remove extra keys
+        guev.pop(k, None)
+    _guev = sorted([k.lower() for k in guev.keys()])
+
+    assert _os_env == _guev
```
mrclary, line 49: @ccordoba12, I'm not entirely happy with this test. Essentially, I'm checking to see if the retrieved user environment variables are the same as `os.environ`, which should be the same; except that there are some exceptions. I fear that this is not very robust and will become a game of whack-a-mole. If you have a better recommendation, I'll be happy to consider it.

mrclary:
@ccordoba12 I've force-pushed but the tests are not running...

ccordoba12:
Closing and reopening usually re-runs the tests.

ccordoba12:
Or maybe something happens with Github, I don't know.

spyder/utils/environ.py
```diff
@@ -97,90 +144,56 @@ def __init__(self, environ, parent=None):
...
-        """Windows User Environment Variables Editor"""
-        def __init__(self, parent=None):
-            super(WinUserEnvDialog, self).__init__(parent)
-            self.setup(get_user_env(),
-                       title=r"HKEY_CURRENT_USER\Environment")
+class UserEnvDialog(RemoteEnvDialog):
+    """User Environment Variables Viewer/Editor"""
+
+    def __init__(self, parent=None):
+        title = _("User Environment variables")
```
ccordoba12, line 156:
```suggestion
        title = _("User environment variables")
```

spyder/utils/environ.py
```diff
@@ -97,90 +144,56 @@ def __init__(self, environ, parent=None):
...
+        readonly = True
+        if os.name == 'nt':
+            title = _(r"HKEY_CURRENT_USER\Environment")
```
ccordoba12, line 159:
```suggestion
            title = _(r"User environment variables in Windows registry")
```

I doubt many users understand what `HKEY_CURRENT_USER` is. However, I think my suggestion better says what we want to convey here.

spyder/plugins/application/container.py
```diff
@@ -170,17 +166,18 @@ def setup(self):
...
-                ApplicationActions.SpyderWindowsEnvVariables,
-                _("Current user environment variables..."),
-                icon=self.create_icon('win_env'),
-                tip=_("Show and edit current user environment "
-                      "variables in Windows registry "
-                      "(i.e. for all sessions)"),
-                triggered=self.show_windows_env_variables)
+        if os.name == 'nt':
+            tip = ("Show and edit current user environment variables in "
+                   "Windows registry (i.e. for all sessions)")
```
ccordoba12, line 171:
```suggestion
            tip = _("Show and edit current user environment variables in "
                    "Windows registry (i.e. for all sessions)")
```

spyder/plugins/application/container.py
```diff
@@ -170,17 +166,18 @@ def setup(self):
...
         else:
-            self.winenv_action = None
+            tip = ("Show current user environment variables (i.e. for all "
+                   "sessions)")
```
ccordoba12, line 174:
```suggestion
            tip = _("Show current user environment variables (i.e. for all "
                    "sessions)")
```

spyder/utils/environ.py
```diff
@@ -12,33 +12,98 @@
 import os
 
 # Third party imports
-from qtpy.QtWidgets import QDialog, QMessageBox
+from qtpy.QtWidgets import QMessageBox
+try:
+    import winreg
+except Exception:
+    pass
```
ccordoba12, line 19: This is a standard library import, so it needs to be in the section above.

ccordoba12 (commented):
@mrclary, last review for you then this should be ready.

ccordoba12 (approved):

Issue #18407, opened 2022-06-23
Numpy 1.23.0 breaks autocompletion and makes the tests fail


impact27:
With numpy 1.23 installed, autocompletion does not work in the editor for numpy (`np.` + `<tab>` doesn't do anything on my mac)

This causes the tests to fail, see:
https://github.com/spyder-ide/spyder/runs/7018273169?check_suite_focus=true
https://github.com/spyder-ide/spyder/runs/7018273253?check_suite_focus=true
https://github.com/spyder-ide/spyder/runs/7018273354?check_suite_focus=true

I can reproduce on my machine. The test pass with numpy 1.21 but fail with 1.23

dalthviz:
Just in case this seems like an issue with Jedi: https://github.com/davidhalter/jedi/issues/1864

PR #18412, opened 2022-06-23
PR: Add constraint on numpy <1.23 to solve tests

impact27:
## Description of Changes
solves https://github.com/spyder-ide/spyder/issues/18407

### Issue(s) Resolved

Fixes #

### Affirmation

### DIFF
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -20,6 +20,7 @@ dependencies:
 - keyring >=17.0.0
 - nbconvert >=4.0
 - numpydoc >=0.6.0
+- numpy >=1.7,<1.23
 - paramiko >=2.4.0
 - parso >=0.7.0,<0.9.0
 - pexpect >=4.4.0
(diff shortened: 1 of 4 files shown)


ccordoba12 (commented):
Thanks for your quick help to solve this problem @impact27! This looks good to me but what we usually do to solve problems like these is not to alter our dependencies because they will be eventually solved. In this case, I think Numpy developers will be informed about the lack of completions and they'll soon release 1.23.1.

Instead, we prefer to install older versions of the packages that are failing here:

https://github.com/spyder-ide/spyder/blob/master/.github/scripts/install.sh

That's the installation script we use for Github actions. So, by requiring Numpy 1.22 there our tests will run without problems too.

So, my advice is to revert the changes you did and only modify `install.sh`. Let us know if that's fine for you or you prefer for me to do it in another PR.

impact27:
You can do it in another PR

PR #18471, opened 2022-06-28
PR: Escape hyphen from update instructions for spyder-kernels (IPython Console)

### LINKED ISSUE #17425: Small bugs in kernel update error screen
Issue is raised in #17353 with updating spyder-kernel in custom environments.

Putting aside the fact that Anaconda doesn't have that version of spyder-kernels yet... there are still two small bugs here on the error screen. 

Firstly, if the required version is 2.2.1, then the error should say 2.2.1 needs installing (not 2.2). Installing 2.2 as suggested still results in the same error which coud be confusing for the user.

Secondly the dash in ``conda install spyder‑kernels=2.2`` doesn't copy in to a Windows terminal correctly (displays a question mark inside a box). A font issue?

dalthviz:
## Description of Changes

Use `&hyphen;` instead of plain `-` to prevent copy-pasting an incorrect command following kernel instructions to update

Pinging @jnsebgosselin just in case to check if another approach for this was found

### Issue(s) Resolved

Fixes #17425 

### Affirmation

### DIFF
--- a/spyder/plugins/ipythonconsole/widgets/main_widget.py
+++ b/spyder/plugins/ipythonconsole/widgets/main_widget.py
@@ -65,9 +65,9 @@ SPYDER_KERNELS_VERSION_MSG = _(
     '>= {0} and < {1}').format(
         SPYDER_KERNELS_MIN_VERSION, SPYDER_KERNELS_MAX_VERSION)
 SPYDER_KERNELS_CONDA = (
-    f'conda install spyder-kernels={SPYDER_KERNELS_MIN_VERSION[:-2]}')
+    f'conda install spyder&#45;kernels={SPYDER_KERNELS_MIN_VERSION[:-2]}')
 SPYDER_KERNELS_PIP = (
-    f'pip install spyder-kernels=={SPYDER_KERNELS_MIN_VERSION[:-1]}*')
+    f'pip install spyder&#45;kernels=={SPYDER_KERNELS_MIN_VERSION[:-1]}*')
 
 
 class IPythonConsoleWidgetActions:


jnsebgosselin:
This seems like a very good solution! Thank you for the fix, this is going to make my life easier.

jnsebgosselin:
I've tested locally and it does not solve the problem on my setup unfortunately. The hyphen still gets replaced by a non-breaking hypen when copied to the clipboard.

jnsebgosselin:
http://www.mauvecloud.net/charsets/CharCodeFinder.html
image

dalthviz:
Seems like now the code is the correct one, no? I think for hyphen the code is `&#x2010` while for non-breaking-hyphen is `&#x2011` However, even with that the command is not working for you then 🤔 ?

Something else I was thinking about is that probably we need to add the `-c conda-forge` part since the anaconda default channel goes just up to spyder-kernels 2.2.1

jnsebgosselin:
Oh yeah, you are right. `&#x2010`  is the code for the normal hyphen. It is still not working in the Windows cmd though. I still need to replace the `&#x2010` (0x2010) with a normal `-` (0x2d).

image

dalthviz:
Could you try using `&#45;`?

jnsebgosselin:
Yes, it works with `&#45;`.

I thought it was either the `pre` or `tt` environment that was causing the issue, but it seems it is something else because I tried without these and the hyphen was still replaced by a non-breaking hyphen. Anyway, it does not matter since you seem to have found a good working solution, thanks!

jnsebgosselin:
image

dalthviz:
Awesome! Thanks for the help checking this @jnsebgosselin !

ccordoba12 (approved):
Thanks @dalthviz and @jnsebgosselin!

PR #18481, opened 2022-06-29
PR: Copy all NSIS plugins into discoverable path

jsbautista:
## Description of Changes

Code to copy all NSIS plugins into discoverable path

### Affirmation

I certify the above statement is true and correct: @jsbautista

### DIFF
--- a/installers/Windows/installer.py
+++ b/installers/Windows/installer.py
@@ -372,7 +372,7 @@ def run(python_version, bitness, repo_root, entrypoint, package, icon_path,
             print("Copying NSIS plugins into discoverable path")
             contents = os.listdir(
                 "installers/Windows/assets/nsist/plugins/x86-unicode/")
-            for element in contents:             
+            for element in contents:
                 shutil.copy(
                     "installers/Windows/assets/nsist/plugins/x86-unicode/" +
                     element,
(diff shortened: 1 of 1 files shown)


jsbautista:
@dalthviz Hi!

installers/Windows/installer.py
```diff
@@ -370,10 +370,13 @@ def run(python_version, bitness, repo_root, entrypoint, package, icon_path,
...
-                "installers/Windows/assets/nsist/plugins/x86-unicode/"
-                "WinShell.dll",
-                "C:/Program Files (x86)/NSIS/Plugins/x86-unicode/WinShell.dll")
+            contents = os.listdir(
+                "installers/Windows/assets/nsist/plugins/x86-unicode/")
+            for element in contents:
+                shutil.copy(
+                    "installers/Windows/assets/nsist/plugins/x86-unicode/" +
+                    element,
+                    "C:/Program Files (x86)/NSIS/Plugins/x86-unicode/"+element)
```
dalthviz, line 379: Could you use `osp.join` instead of `+` to append the `element` variable to the path strings please?

dalthviz (commented):
Thanks @jsbautista ! I left a comment regarding the way to append the path strings. Also, could you squash your commits into a single one, please?

Other than that this looks good to me :+1:

installers/Windows/installer.py
dalthviz, line 379: @jsbautista , just in case, the changes as suggestions:
```suggestion
                    osp.join(
                        "installers/Windows/assets/nsist/plugins/x86-unicode/",
                        element),
                    osp.join(
                        "C:/Program Files (x86)/NSIS/Plugins/x86-unicode/",
                        element))
```

PR #18485, opened 2022-06-29
PR: Remove applaunchservices dependency for macOS app

mrclary:
## Description of Changes

Removed dependency on `applaunchservices` for the macOS standalone application. The features added by this package are not needed for the macOS app, but its dependency (`pyobjc-framework-CoreServices`) is not hardened runtime compatible, which prevents successful code signing. 

`applaunchservices` remains an installation dependency for the Spyder package, and thus is still included in the macOS application bundle. We should investigate whether `applaunchservices` installation dependence can be conditional (beyond platform) so that it can be excluded from the macOS app build.

### Affirmation


CAM-Gerlach:
Seems reasonable to me, though most of it relates to areas outside my direct expertise, i.e. the mac app bundling, and strongly within yours. Therefore, per our new strict review policy, I guess I should remove my review request and not review it, or I'll become a blocker on merging it.

mrclary:
> Seems reasonable to me, though most of it relates to areas outside my direct expertise, i.e. the mac app bundling, and strongly within yours. Therefore, per our new strict review policy, I guess I should remove my review request and not review it, or I'll become a blocker on merging it.

Actually, I was hoping to solicit comments from you regarding conditional installation requirements. I don't know if it is even possible to do, but it would be nice not to have `applaunchservices` as an installation requirement when building the macOS app. While the dependency checks internal to Spyder can be conditioned on the macOS app (as this PR does), I don't know of a way to do that in the `setup.py` file, similar to what we do for platform.

I can just exclude `applaunchservices` from the app build, but if it still shows up in the egg requirements, I don't know if that will end up causing issues with `pkg_resources` (or its future replacement).

CAM-Gerlach:
> While the dependency checks internal to Spyder can be conditioned on the macOS app (as this PR does), I don't know of a way to do that in the setup.py file, similar to what we do for platform.

Doesn't the Mac installer use its own `setup.py`, so you could just do the change there? Otherwise, while you could do some arbitrary hackery in the regular Spyder `setup.py` to detect whether it is being run by an installer, we're really trying to move away from that to a static, declarative `pyproject.toml`/`setup.cfg` config. However, if one of the [environment markers](https://peps.python.org/pep-0508/#environment-markers) will work to uniquely identify when the package is being built by the installer, then that would work statically. Also, you can definitely use the `sys_platform` environment marker to only install it on macOS to begin with.

To answer more completely, I'd have to understand more about how the installer is built and its dependencies are specified, versus the standard builds.

> I don't know if that will end up causing issues with pkg_resources (or its future replacement).

It *shouldn't* in the standalone installers at least using modern tooling, but I can't be 100% sure since I'm not super familiar with the standalone installers.

mrclary:
> Doesn't the Mac installer use its own `setup.py`, so you could just do the change there? 

Yes, but it does not have anything to do with Spyder's installation. Spyder needs to be already installed in the environment, and the `setup.py` for building the macOS app is the interface for `py2app`'s activities. So I could resort to hackery and reinstall Spyder upon build, but...

>Otherwise, while you could do some arbitrary hackery in the regular Spyder `setup.py` to detect whether it is being run by an installer, we're really trying to move away from that to a static, declarative `pyproject.toml`/`setup.cfg` config.

I agree. Hackery is not an acceptable solution to me.

> However, if one of the [environment markers](https://peps.python.org/pep-0508/#environment-markers) will work to uniquely identify when the package is being built by the installer, then that would work statically. Also, you can definitely use the `sys_platform` environment marker to only install it on macOS to begin with.

This is the right idea, but, alas, I don't think there is anything about a build environment that uniquely identifies it as such.

> > I don't know if that will end up causing issues with pkg_resources (or its future replacement).
> 
> It _shouldn't_ in the standalone installers at least using modern tooling, but I can't be 100% sure since I'm not super familiar with the standalone installers.

Well, I suppose I'll just have to test it. Fortunately, `applaunchservices` and its dependencies should be benign.

CAM-Gerlach:
If we had a default extra or subtracting extras, this would be easy. But alas, we don't, at least not yet. But I can think some more about this if you can't find a good solution.

Also, I assume there isn't any solution at the installer configuration level for `py2app`?

mrclary:
> If we had a default extra or subtracting extras, this would be easy. But alas, we don't, at least not yet. But I can think some more about this if you can't find a good solution.

Well, don't waste too much time worrying about it. It's not a show stopper.

> Also, I assume there isn't any solution at the installer configuration level for `py2app`?

Not that I've seen.

ccordoba12:
@impact27, thoughts about this one? @mrclary needs it to sign our Mac app.

ccordoba12:
Pinging @impact27 again about this one. We need this for 5.3.2, to be released at the end of the week.

mrclary:
@impact27, can you please provide your review of this PR soon?

impact27 (commented):
Looks good to me

ccordoba12:
Great! Merging then.

ccordoba12 (approved):

PR #18510, opened 2022-07-01
PR: Set permissions for GitHub actions

neilnaveen:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: neilnaveen <42328488+neilnaveen@users.noreply.github.com>

### DIFF
--- a/.github/workflows/installer-macos.yml
+++ b/.github/workflows/installer-macos.yml
@@ -8,6 +8,8 @@ name: Create macOS App Bundle and DMG
 
 jobs:
   matrix_prep:
+    permissions:
+      contents: none
     name: Determine Build Matrix
     runs-on: macos-10.15
     outputs:


ccordoba12:
@mrclary, thoughts about this one?

mrclary:
@ccordoba12, I don't know anything about the security issues referenced. I do not understand what the proposed change accomplishes. I also wonder, assuming that the proposed change is appropriate:
* Do we need to similarly "set permissions" on all of our workflows?
* What is the implication (if any) for code signing and notarization?

ccordoba12:
I have no idea either. @neilnaveen, could you try to address @mrclary's questions?

neilnaveen:
> @ccordoba12, I don't know anything about the security issues referenced. I do not understand what the proposed change accomplishes. I also wonder, assuming that the proposed change is appropriate:
> 
> * Do we need to similarly "set permissions" on all of our workflows?
yes, but for now I could only address these changes.
> * What is the implication (if any) for code signing and notarization?
TBH I don't know.

Thanks

ccordoba12:
Closing because we don't understand what this change does and no clarification was provided.

PR #17760, opened 2022-04-26
PR: Fix setting working directory at startup and for new IPython consoles

ccordoba12:
## Description of Changes

* This is for @isabela-pf and @CAM-Gerlach, so they can see the UI changes proposed to the Preferences page of the Working directory plugin:

    **Before**

    imagen

    **After**

    image

* And this is for @dalthviz, so he can better understand the changes proposed here when reviewing this code:

    - I made the consoles working directory to be set correctly at startup and for new consoles. The previous code was trying to set it before the kernel is up, which had no effect. Furthermore, if a project is active at startup, setting the working directory to the project's one was failing due to Projects not being listed as part of the optional requirements of the IPython console (and the problem mentioned above).
    - Now we raise an error if a method is decorated with `on_plugin_available` or `on_plugin_teardown`, but the plugin it makes a reference to is not listed in REQUIRES nor OPTIONAL.
    - The Working Directory plugin now only uses the `Startup` options listed in Preferences to get the initial directory (not the `New consoles` ones because they don't apply). Also, I fixed its `get_workdir` method, which was not giving the current directory but the initial one instead.
    - I added tests to the IPython console to check things are working as expected now.

### Issue(s) Resolved

Fixes #17737.

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct:

PR #18544, opened 2022-07-04
PR: Add new API methods to the IPython console

ccordoba12:
## Description of Changes

- These methods (i.e. `create_client_for_kernel` and `rename_client_tab`) are necessary to finish the migration of Spyder-notebook to the new API (see spyder-ide/spyder-notebook#369)
- This PR also corrects the way we handle the current working directory to use it to start new consoles. That was implemented as part of PR #17760 but it didn't follow the patterns of our new API.

### Affirmation

### DIFF
--- a/spyder/plugins/ipythonconsole/plugin.py
+++ b/spyder/plugins/ipythonconsole/plugin.py
@@ -13,7 +13,7 @@ import os
 import os.path as osp
 
 # Third party imports
-from qtpy.QtCore import Signal
+from qtpy.QtCore import Signal, Slot
 
 # Local imports
 from spyder.api.plugins import Plugins, SpyderDockablePlugin
@@ -329,6 +329,12 @@ class IPythonConsole(SpyderDockablePlugin):
         projects.sig_project_loaded.connect(self._on_project_loaded)
         projects.sig_project_closed.connect(self._on_project_closed)
 
+    @on_plugin_available(plugin=Plugins.WorkingDirectory)
+    def on_working_directory_available(self):
+        working_directory = self.get_plugin(Plugins.WorkingDirectory)
+        working_directory.sig_current_directory_changed.connect(
+            self._save_working_directory)
+
     @on_plugin_teardown(plugin=Plugins.Preferences)
     def on_preferences_teardown(self):
         # Register conf page
@@ -370,6 +376,12 @@ class IPythonConsole(SpyderDockablePlugin):
         projects = self.get_plugin(Plugins.Projects)
         projects.sig_project_loaded.disconnect(self._on_project_loaded)
         projects.sig_project_closed.disconnect(self._on_project_closed)
+
+    @on_plugin_teardown(plugin=Plugins.WorkingDirectory)
+    def on_working_directory_teardown(self):
+        working_directory = self.get_plugin(Plugins.WorkingDirectory)
+        working_directory.sig_current_directory_changed.disconnect(
+            self._save_working_directory)
 
     def update_font(self):
         """Update font from Preferences"""
@@ -425,11 +437,10 @@ class IPythonConsole(SpyderDockablePlugin):
                     except Exception:
                         pass
 
-    def _get_working_directory(self):
-        """Get current working directory from its plugin."""
-        workdir = self.get_plugin(Plugins.WorkingDirectory)
-        if workdir:
-            return workdir.get_workdir()
+    @Slot(str)
+    def _save_working_directory(self, new_dir):
+        """Save current working directory on the main widget."""
+        self.get_widget().current_working_directory = new_dir
 
     # ---- Public API
     # -------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -1075,155 +1080,16 @@ def _shellwidget_deleted(self, client):
         except RuntimeError:
             pass
 
-    def _create_client_for_kernel(self, connection_file, hostname, sshkey,
```
ccordoba12, line 1078: Here I simply moved this method to the `Public API` section and made the previous `create_client_for_kernel` private.

dalthviz: @ccordoba12 I think this method is being used on `spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py::test_tab_rename_for_slaves` so a change to the test needs to be done for the test to pass.

ccordoba12: Yep, it was used in a bunch of tests. I replaced that with the new API method.

dalthviz (dismissed):
Thanks @ccordoba12 ! LGTM 👍🏼

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -425,11 +437,10 @@ def _remove_old_std_files(self):
...
 
-    def _get_working_directory(self):
-        """Get current working directory from its plugin."""
-        workdir = self.get_plugin(Plugins.WorkingDirectory)
-        if workdir:
-            return workdir.get_workdir()
+    @Slot(str)
+    def _save_working_directory(self, new_dir):
+        """Save current working directory on the main widget."""
+        self.get_widget().current_working_directory = new_dir
```
steff456, line 443: I think it will be best to have a function inside the container that sets the current working directory rather than accessing attributes directly

spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py
```diff
@@ -2287,7 +2281,7 @@ def get_cwd_of_new_client():
 
     # Simulate a specific directory
     cwd_dir = str(tmpdir.mkdir('ipyconsole_cwd_test'))
-    ipyconsole._get_working_directory = lambda: cwd_dir
+    ipyconsole.get_widget().current_working_directory = cwd_dir
```
steff456, line 2284: You will be able to use the function here as well

spyder/plugins/ipythonconsole/widgets/client.py
```diff
@@ -386,7 +386,7 @@ def _set_initial_cwd(self):
                 if project_path is not None:
                     cwd_path = project_path
             elif self.get_conf('console/use_cwd', section='workingdir'):
-                cwd_path = self.container.get_working_directory()
+                cwd_path = self.container.current_working_directory
```
steff456, line 389: It is best to leave the setter/getter for the working directory

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -2366,6 +2372,18 @@ def run_script(self, filename, wdir, args, debug, post_mortem,
             )
 
     # ---- For working directory and path management
+    @property
+    def current_working_directory(self):
```
steff456, line 2376: This maybe should be `get_current_working_directory` and remove the property so it will remain as function

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -2366,6 +2372,18 @@ def run_script(self, filename, wdir, args, debug, post_mortem,
...
+        """
+        Current working directory, as reported by the Working Directory plugin.
+        """
+        return self._current_working_directory
+
+    @current_working_directory.setter
+    def current_working_directory(self, new_value):
```
steff456, line 2383: Maybe call this one `set_current_working_directory` and use it to set it instead of changing the attribute

steff456 (changes requested):
thanks @ccordoba12 for this work, I left a couple of comments of style

spyder/plugins/ipythonconsole/tests/test_ipythonconsole.py
```diff
@@ -554,7 +551,7 @@ def test_cython_client(ipyconsole, qtbot):
 def test_tab_rename_for_slaves(ipyconsole, qtbot):
     """Test slave clients are renamed correctly."""
     cf = ipyconsole.get_current_client().connection_file
-    ipyconsole.get_widget()._create_client_for_kernel(cf, None, None, None)
+    ipyconsole.create_client_for_kernel(cf)
```
dalthviz, line 554: Seems like there is an issue with the usage of `get_client_for_file` inside this method?: https://github.com/spyder-ide/spyder/runs/7199975972?check_suite_focus=true#step:13:866

spyder/plugins/ipythonconsole/plugin.py
ccordoba12: Ok, I'll do the necessary changes.

dalthviz (approved):

PR #18548, opened 2022-07-04
PR: Change default installers assets download URL (Windows)

dalthviz:
## Description of Changes

* Change default URL for Windows installer assets to be the latest release available on the `window-installer-assets` repository

### Affirmation

### DIFF
--- a/.github/workflows/installer-win.yml
+++ b/.github/workflows/installer-win.yml
@@ -19,7 +19,10 @@ jobs:
       matrix:
         build_type: ['Lite', 'Full']
     env:
-      ASSETS_URL: 'https://github.com/spyder-ide/windows-installer-assets/releases/download/0.0.7/assets.zip'
+      # The following line can be used to specify from where to download the installer assets.
+      # For example, if you want to download from an specific repo branch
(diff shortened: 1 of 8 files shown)


dalthviz:
Note: Not totally sure but most probably a similar change as the last commit here could be needed for the MacOS installer workflow. Seems like the installer workflows are not running at all like for example at https://github.com/spyder-ide/spyder/pull/18544

.github/workflows/installer-win.yml
```diff
@@ -2,6 +2,11 @@
 on:
   pull_request:
     paths:
+      - 'requirements/*.txt'
```
ccordoba12, line 5:
```suggestion
      - 'requirements/*.yml'
```

ccordoba12: Perhaps you should also change this in all other places that reference `'requirements/*.yml'`?

dalthviz: Should I also add the changes to the MacOS installer workflow @ccordoba12 ?

ccordoba12: Yes, please go ahead and do that too.

ccordoba12 (approved):
Looks good to me now, thanks @dalthviz!

PR #18619, opened 2022-07-11
PR: Improve update UX and automatically download latest installers

jsbautista:
## Description of Changes

* Show download status and status bar icon (Spyder version, if checking for updates, if an update is pending, downloading installer, installing, etc)
* Logic to download and run the new Windows/MacOS installer version if available

Part of #19112 

### Affirmation

I certify the above statement is true and correct: @jsbautista

### DIFF
--- a/spyder/plugins/application/container.py
+++ b/spyder/plugins/application/container.py
@@ -27,7 +27,8 @@ from spyder.api.translations import get_translation
 from spyder.api.widgets.main_container import PluginMainContainer
 from spyder.utils.installers import InstallerMissingDependencies
 from spyder.config.utils import is_anaconda
-from spyder.config.base import get_conf_path, get_debug_level, is_pynsist
+from spyder.config.base import (get_conf_path, get_debug_level, is_pynsist,
+                                running_in_mac_app)
 from spyder.plugins.application.widgets.status import ApplicationUpdateStatus
 from spyder.plugins.console.api import ConsoleActions
 from spyder.utils.qthelpers import start_file, DialogManager
@@ -314,7 +315,7 @@ class ApplicationContainer(PluginMainContainer):
                         "<code>conda install spyder={}</code><br><br>"
                     ).format(latest_release)
                 else:
-                    if is_pynsist():
+                    if is_pynsist() or running_in_mac_app():
                         box.setStandardButtons(QMessageBox.Yes |
                                                QMessageBox.No)
                         content = _(
--- a/spyder/plugins/application/container.py
+++ b/spyder/plugins/application/container.py
@@ -49,6 +49,16 @@ if os.name == 'nt':
 _ = get_translation('spyder')
 
 
+# Update Installation process statuses
+NO_STATUS = _("No status")
+DOWNLOADING_INSTALLER = _("Downloading installer")
+INSTALLING = _("Installing")
+FINISHED = _("Installation finished")
+PENDING = _("Update ready")
+CHECKING = _("Checking for updates")
+CANCELLED = _("Cancelled")
+
+
 class ApplicationPluginMenus:
     DebugLogsMenu = "debug_logs_menu"
 
@@ -80,16 +90,6 @@ class ApplicationActions:
     # so 'Restart' is used instead of something like 'restart_action'
     SpyderRestart = "Restart"
     SpyderRestartDebug = "Restart in debug mode"
-
-
-# Update Installation process statuses
-NO_STATUS = _("No status")
-DOWNLOADING_INSTALLER = _("Downloading installer")
-INSTALLING = _("Installing")
-FINISHED = _("Installation finished")
-PENDING = _("Update ready")
-CHECKING = _("Checking for updates")
-CANCELLED = _("Cancelled")
 
 
 class ApplicationContainer(PluginMainContainer):
@@ -364,7 +364,7 @@ class ApplicationContainer(PluginMainContainer):
             check_updates = box.is_checked()
         else:
 
-            if True:
+            if update_available:
 
                 header = _("<b>Spyder {} is available!</b><br><br>").format(
                     latest_release)
@@ -372,7 +372,7 @@ class ApplicationContainer(PluginMainContainer):
                     "For more information visit our "
                     "<a href=\"{}\">installation guide</a>."
                 ).format(url_i)
-                if not is_anaconda():
+                if is_anaconda():
                     content = _(
                         "<b>Important note:</b> Since you installed "
                         "Spyder with Anaconda, please <b>don't</b> use "
--- a/spyder/plugins/application/container.py
+++ b/spyder/plugins/application/container.py
@@ -294,7 +294,7 @@ class ApplicationContainer(PluginMainContainer):
 
                 header = _("<b>Spyder {} is available!</b> ",
                            "<i>(you have {})</i><br><br>").format(
-                    latest_release,)
+                    latest_release, __version__)
                 footer = _(
                     "For more information visit our "
                     "<a href=\"{}\">installation guide</a>."
--- a/spyder/plugins/application/container.py
+++ b/spyder/plugins/application/container.py
@@ -28,13 +28,13 @@ from spyder.api.widgets.main_container import PluginMainContainer
 from spyder.utils.installers import InstallerMissingDependencies
 from spyder.config.utils import is_anaconda
 from spyder.config.base import get_conf_path, get_debug_level, is_pynsist
+from spyder.plugins.application.widgets.status import ApplicationUpdateStatus
 from spyder.plugins.console.api import ConsoleActions
 from spyder.utils.qthelpers import start_file, DialogManager
 from spyder.widgets.about import AboutDialog
 from spyder.widgets.dependencies import DependenciesDialog
 from spyder.widgets.helperwidgets import MessageCheckBox
 from spyder.workers.updates import WorkerUpdates
-from spyder.plugins.application.widgets.status import ApplicationUpdateStatus
 
 
 WinUserEnvDialog = None
@@ -89,6 +89,7 @@ class ApplicationContainer(PluginMainContainer):
     """
     Signal to load a log file
     """
+
     sig_show_status_bar_widget = Signal(bool)
     """
     Signal to show the widget in status bar
@@ -107,6 +108,7 @@ class ApplicationContainer(PluginMainContainer):
 
         self.application_update_status = ApplicationUpdateStatus(parent=self)
         self.application_update_status.set_no_status()
+
         # Compute dependencies in a thread to not block the interface.
         self.dependencies_thread = QThread(None)
 
@@ -245,7 +247,6 @@ class ApplicationContainer(PluginMainContainer):
     # -------------------------------------------------------------------------
 
     def _check_updates_ready(self):
-
         """Show results of the Spyder update checking process."""
 
         # `feedback` = False is used on startup, so only positive feedback is
@@ -254,7 +255,6 @@ class ApplicationContainer(PluginMainContainer):
         feedback = self.give_updates_feedback
 
         # Get results from worker
-
         update_available = self.worker_updates.update_available
         latest_release = self.worker_updates.latest_release
         error_msg = self.worker_updates.error
@@ -316,17 +316,19 @@ class ApplicationContainer(PluginMainContainer):
                                                QMessageBox.No)
                         content = _(
                             "You want to download and install the latest "
-                            "version of spyder?<br><br>"
+                            "version of Spyder?<br><br>"
                         )
                     else:
                         content = _(
                             "Click <a href=\"{}\">this link</a> to "
                             "download it.<br><br>"
                         ).format(url_r)
+
                 msg = header + content + footer
                 box.setText(msg)
                 box.set_check_visible(True)
                 box.exec_()
+
                 if box.result() == QMessageBox.Yes:
                     self.application_update_status.start_installation()
                 elif(box.result() == QMessageBox.No):
--- a/spyder/plugins/application/container.py
+++ b/spyder/plugins/application/container.py
@@ -45,16 +45,6 @@ if os.name == 'nt':
 _ = get_translation('spyder')
 
 
-# Update Installation process statuses
-NO_STATUS = _("No status")
-DOWNLOADING_INSTALLER = _("Downloading installer")
-INSTALLING = _("Installing")
-FINISHED = _("Installation finished")
-PENDING = _("Pending update")
-CHECKING = _("Checking for updates")
-CANCELLED = _("Cancelled")
-
-
 class ApplicationPluginMenus:
     DebugLogsMenu = "debug_logs_menu"
 
@@ -116,7 +106,7 @@ class ApplicationContainer(PluginMainContainer):
     def setup(self):
 
         self.application_update_status = ApplicationUpdateStatus(parent=self)
-        self.application_update_status.set_value(NO_STATUS)
+        self.application_update_status.set_no_status()
         # Compute dependencies in a thread to not block the interface.
         self.dependencies_thread = QThread(None)
 
@@ -301,7 +291,7 @@ class ApplicationContainer(PluginMainContainer):
         else:
 
             if update_available:
-                self.application_update_status.set_value(PENDING)
+                self.application_update_status.set_status_pending()
 
                 header = _("<b>Spyder {} is available!</b><br><br>").format(
                     latest_release)
@@ -320,7 +310,7 @@ class ApplicationContainer(PluginMainContainer):
                         "<code>conda update anaconda</code><br>"
                         "<code>conda install spyder={}</code><br><br>"
                     ).format(latest_release)
-                    self.application_update_status.set_value(NO_STATUS)
+                    self.application_update_status.set_no_status()
                 else:
                     if os.name == 'nt' and is_pynsist():
                         box.setStandardButtons(QMessageBox.Yes |
@@ -346,6 +336,7 @@ class ApplicationContainer(PluginMainContainer):
                 box.set_check_visible(False)
                 box.exec_()
                 check_updates = box.is_checked()
+                self.application_update_status.set_no_status()
 
         # Update checkbox based on user interaction
         self.set_conf(option, check_updates)
@@ -362,9 +353,9 @@ class ApplicationContainer(PluginMainContainer):
         # Disable check_updates_action while the thread is working
         self.check_updates_action.setDisabled(True)
         if os.name == 'nt':
-            self.application_update_status.set_value(CHECKING)
+            self.application_update_status.set_status_checking()
         else:
-            self.application_update_status.set_value(NO_STATUS)
+            self.application_update_status.set_no_status()
 
         if self.thread_updates is not None:
             self.thread_updates.quit()
--- a/spyder/plugins/application/container.py
+++ b/spyder/plugins/application/container.py
@@ -311,12 +311,12 @@ class ApplicationContainer(PluginMainContainer):
                         "<code>conda install spyder={}</code><br><br>"
                     ).format(latest_release)
                 else:
-                    if os.name == 'nt' and is_pynsist():
+                    if is_pynsist():
                         box.setStandardButtons(QMessageBox.Yes |
                                                QMessageBox.No)
                         content = _(
-                            "You want to download and install the latest "
-                            "version of Spyder?<br><br>"
+                            "Would you like to automatically download and"
+                            " install the latest version of Spyder?<br><br>"
                         )
                     else:
                         content = _(
@@ -341,7 +341,8 @@ class ApplicationContainer(PluginMainContainer):
                 box.exec_()
                 check_updates = box.is_checked()
                 self.application_update_status.set_no_status()
-
+            else:
+                self.application_update_status.set_no_status()
         # Update checkbox based on user interaction
         self.set_conf(option, check_updates)
 
--- a/spyder/plugins/application/container.py
+++ b/spyder/plugins/application/container.py
@@ -21,7 +21,7 @@ from qtpy.QtGui import QGuiApplication
 from qtpy.QtWidgets import QAction, QMessageBox, QPushButton
 
 # Local imports
-from spyder import __docs_url__, __forum_url__, __trouble_url__
+from spyder import __docs_url__, __forum_url__, __trouble_url__, __version__
 from spyder import dependencies
 from spyder.api.translations import get_translation
 from spyder.api.widgets.main_container import PluginMainContainer
@@ -292,8 +292,9 @@ class ApplicationContainer(PluginMainContainer):
             if update_available:
                 self.application_update_status.set_status_pending()
 
-                header = _("<b>Spyder {} is available!</b><br><br>").format(
-                    latest_release)
+                header = _("<b>Spyder {} is available!</b> ",
+                           "<i>(you have {})</i><br><br>").format(
+                    latest_release,)
                 footer = _(
                     "For more information visit our "
                     "<a href=\"{}\">installation guide</a>."
--- a/spyder/plugins/application/container.py
+++ b/spyder/plugins/application/container.py
@@ -91,11 +91,6 @@ class ApplicationContainer(PluginMainContainer):
     Signal to load a log file
     """
 
-    sig_show_status_bar_widget = Signal(bool)
-    """
-    Signal to show the widget in status bar
-    """
-
     def __init__(self, name, plugin, parent=None):
         super().__init__(name, plugin, parent)
 
@@ -322,11 +317,6 @@ class ApplicationContainer(PluginMainContainer):
                             "Would you like to automatically download and"
                             " install the latest version of Spyder?<br><br>"
                         )
-                    else:
-                        content = _(
-                            "Click <a href=\"{}\">this link</a> to "
-                            "download it.<br><br>"
-                        ).format(url_r)
 
                 msg = header + content + footer
                 box.setText(msg)
--- a/spyder/plugins/application/container.py
+++ b/spyder/plugins/application/container.py
@@ -309,14 +309,13 @@ class ApplicationContainer(PluginMainContainer):
                         "<code>conda update anaconda</code><br>"
                         "<code>conda install spyder={}</code><br><br>"
                     ).format(latest_release)
-                else:
-                    if is_pynsist() or running_in_mac_app():
-                        box.setStandardButtons(QMessageBox.Yes |
-                                               QMessageBox.No)
-                        content = _(
-                            "Would you like to automatically download and"
-                            " install the latest version of Spyder?<br><br>"
-                        )
+                elif is_pynsist() or running_in_mac_app():
+                    box.setStandardButtons(QMessageBox.Yes |
+                                           QMessageBox.No)
+                    content = _(
+                        "Would you like to automatically download and"
+                        " install the latest version of Spyder?<br><br>"
+                    )
 
                 msg = header + content + footer
                 box.setText(msg)
--- a/spyder/plugins/application/container.py
+++ b/spyder/plugins/application/container.py
@@ -96,7 +96,7 @@ class ApplicationContainer(PluginMainContainer):
 
     def __init__(self, name, plugin, parent=None):
         super().__init__(name, plugin, parent)
-        
+
         # Keep track of dpi message
         self.current_dpi = None
         self.dpi_messagebox = None
@@ -318,7 +318,13 @@ class ApplicationContainer(PluginMainContainer):
                             "You want to download and install the latest "
                             "version of spyder?<br><br>"
                         )
-                        box.buttonClicked.connect(self.application_update_status.start_installation)
+                        box.buttonClicked.connect(
+                            lambda button:
+                            self.application_update_status.start_installation()
+                            if button.text() in ('&Yes')
+                            else
+                            self.application_update_status.set_status_pending()
+                            )
                     else:
                         content = _(
                             "Click <a href=\"{}\">this link</a> to "
@@ -621,4 +627,3 @@ class ApplicationContainer(PluginMainContainer):
             # Fixes spyder-ide/spyder#17677
             self.dpi_messagebox.move(int(x), int(y))
             self.dpi_messagebox.adjustSize()
-
--- a/spyder/plugins/application/container.py
+++ b/spyder/plugins/application/container.py
@@ -14,10 +14,6 @@ Holds references for base actions in the Application of Spyder.
 import os
 import sys
 import glob
-import subprocess
-from urllib.request import urlretrieve
-from tempfile import TemporaryDirectory
-import threading
 
 # Third party imports
 from qtpy.QtCore import Qt, QThread, QTimer, Signal, Slot
@@ -31,7 +27,7 @@ from spyder.api.translations import get_translation
 from spyder.api.widgets.main_container import PluginMainContainer
 from spyder.utils.installers import InstallerMissingDependencies
 from spyder.config.utils import is_anaconda
-from spyder.config.base import get_conf_path, get_debug_level
+from spyder.config.base import get_conf_path, get_debug_level, is_pynsist
 from spyder.plugins.console.api import ConsoleActions
 from spyder.utils.qthelpers import start_file, DialogManager
 from spyder.widgets.about import AboutDialog
@@ -49,7 +45,7 @@ if os.name == 'nt':
 _ = get_translation('spyder')
 
 
-# Update Installation process statusesf
+# Update Installation process statuses
 NO_STATUS = _("No status")
 DOWNLOADING_INSTALLER = _("Downloading installer")
 INSTALLING = _("Installing")
@@ -94,16 +90,6 @@ class ApplicationActions:
 
 class ApplicationContainer(PluginMainContainer):
 
-    # Signals
-    # Signal to get the current status of the update installation
-    # str: Status string
-    sig_installation_status = Signal(str)
-
-    # Signal to get the download progress
-    # int: Download progress
-    # int: Total download size
-    sig_download_progress = Signal(int, int)
-
     sig_report_issue_requested = Signal()
     """
     Signal to request reporting an issue to Github.
@@ -120,11 +106,7 @@ class ApplicationContainer(PluginMainContainer):
 
     def __init__(self, name, plugin, parent=None):
         super().__init__(name, plugin, parent)
-        # Windows(OS) update installation
-        self.cancelled = False
-        self.status = NO_STATUS
-        self.thread_install_update = None
-
+        
         # Keep track of dpi message
         self.current_dpi = None
         self.dpi_messagebox = None
@@ -134,7 +116,7 @@ class ApplicationContainer(PluginMainContainer):
     def setup(self):
 
         self.application_update_status = ApplicationUpdateStatus(parent=self)
-        self._change_update_installation_status(status=NO_STATUS)
+        self.application_update_status.set_value(NO_STATUS)
         # Compute dependencies in a thread to not block the interface.
         self.dependencies_thread = QThread(None)
 
@@ -271,52 +253,6 @@ class ApplicationContainer(PluginMainContainer):
 
     # ---- Updates
     # -------------------------------------------------------------------------
-    def _change_update_installation_status(self, status=NO_STATUS):
-        """Set the installation status."""
-        self.status = status
-        self.sig_installation_status.emit(self.status)
-
-    def _progress_reporter(self, block_number, read_size, total_size):
-        progress = 0
-        if total_size > 0:
-            progress = block_number * read_size
-        if self.cancelled:
-            raise UpdateInstallationCancelledException()
-        else:
-            self.sig_download_progress.emit(progress, total_size)
-
-    def cancell_thread_install_update(self):
-        self._change_update_installation_status(status=CANCELLED)
-        self.thread_install_update.join()
-
-    def _download_install(self):
-        try:
-            with TemporaryDirectory(prefix="Spyder-") as tmpdir:
-                destination = os.path.join(tmpdir, 'updateSpyder.exe')
-                download = urlretrieve(
-                    ('https://github.com/spyder-ide/spyder/releases/latest/'
-                        'download/Spyder_64bit_full.exe'),
-                    destination,
-                    reporthook=self._progress_reporter)
-                self._change_update_installation_status(status=INSTALLING)
-                install = subprocess.Popen(destination, shell=True)
-                install.communicate()
-        except UpdateInstallationCancelledException:
-            self._change_update_installation_status(status=CANCELLED)
-        finally:
-            self._change_update_installation_status(status=PENDING)
-
-    def installation_thread(self, option):
-        if option.text() in ('&Yes'):
-            self.cancelled = False
-            self._change_update_installation_status(
-                status=DOWNLOADING_INSTALLER)
-            """call a function in a simple thread, to prevent blocking"""
-            self.thread_install_update = threading.Thread(
-                target=self._download_install)
-            self.thread_install_update.start()
-        else:
-            self._change_update_installation_status(status=PENDING)
 
     def _check_updates_ready(self):
 
@@ -365,6 +301,7 @@ class ApplicationContainer(PluginMainContainer):
         else:
 
             if update_available:
+                self.application_update_status.set_value(PENDING)
 
                 header = _("<b>Spyder {} is available!</b><br><br>").format(
                     latest_release)
@@ -383,16 +320,16 @@ class ApplicationContainer(PluginMainContainer):
                         "<code>conda update anaconda</code><br>"
                         "<code>conda install spyder={}</code><br><br>"
                     ).format(latest_release)
-                    self._change_update_installation_status(status=NO_STATUS)
+                    self.application_update_status.set_value(NO_STATUS)
                 else:
-                    if os.name == 'nt':
+                    if os.name == 'nt' and is_pynsist():
                         box.setStandardButtons(QMessageBox.Yes |
                                                QMessageBox.No)
                         content = _(
                             "You want to download and install the latest "
                             "version of spyder?<br><br>"
                         )
-                        box.buttonClicked.connect(self.installation_thread)
+                        box.buttonClicked.connect(self.application_update_status.start_installation)
                     else:
                         content = _(
                             "Click <a href=\"{}\">this link</a> to "
@@ -409,9 +346,6 @@ class ApplicationContainer(PluginMainContainer):
                 box.set_check_visible(False)
                 box.exec_()
                 check_updates = box.is_checked()
-                self._change_update_installation_status(status=NO_STATUS)
-            else:
-                self._change_update_installation_status(status=NO_STATUS)
 
         # Update checkbox based on user interaction
         self.set_conf(option, check_updates)
@@ -428,9 +362,9 @@ class ApplicationContainer(PluginMainContainer):
         # Disable check_updates_action while the thread is working
         self.check_updates_action.setDisabled(True)
         if os.name == 'nt':
-            self._change_update_installation_status(status=CHECKING)
+            self.application_update_status.set_value(CHECKING)
         else:
-            self._change_update_installation_status(status=NO_STATUS)
+            self.application_update_status.set_value(NO_STATUS)
 
         if self.thread_updates is not None:
             self.thread_updates.quit()
--- a/spyder/plugins/application/container.py
+++ b/spyder/plugins/application/container.py
@@ -323,7 +323,8 @@ class ApplicationContainer(PluginMainContainer):
                 box.exec_()
 
                 if box.result() == QMessageBox.Yes:
-                    self.application_update_status.start_installation()
+                    self.application_update_status.start_installation(
+                        latest_release=latest_release)
                 elif(box.result() == QMessageBox.No):
                     self.application_update_status.set_status_pending()
                 check_updates = box.is_checked()
--- a/spyder/plugins/application/container.py
+++ b/spyder/plugins/application/container.py
@@ -310,7 +310,6 @@ class ApplicationContainer(PluginMainContainer):
                         "<code>conda update anaconda</code><br>"
                         "<code>conda install spyder={}</code><br><br>"
                     ).format(latest_release)
-                    self.application_update_status.set_no_status()
                 else:
                     if os.name == 'nt' and is_pynsist():
                         box.setStandardButtons(QMessageBox.Yes |
--- a/spyder/plugins/application/container.py
+++ b/spyder/plugins/application/container.py
@@ -107,6 +107,9 @@ class ApplicationContainer(PluginMainContainer):
     def setup(self):
 
         self.application_update_status = ApplicationUpdateStatus(parent=self)
+        self.application_update_status.sig_check_for_updates_requested.connect(
+            self.check_updates
+        )
         self.application_update_status.set_no_status()
 
         # Compute dependencies in a thread to not block the interface.
--- a/spyder/plugins/application/container.py
+++ b/spyder/plugins/application/container.py
@@ -276,7 +276,9 @@ class ApplicationContainer(PluginMainContainer):
                 raise UpdateInstallationCancelledException()
             else:
                 self.sig_download_progress.emit(progress, total_size)
+
     def _cancell_thread_install_update(self):
+        self._change_update_installation_status(status=CANCELLED)
         self.thread_install_update.join()
     
     def _check_updates_ready(self):
@@ -302,20 +304,22 @@ class ApplicationContainer(PluginMainContainer):
         url_i = 'https://docs.spyder-ide.org/installation.html'      
 
         def _download_install():            
-            with TemporaryDirectory(prefix = "Spyder-") as tmpdir:
+            with TemporaryDirectory(prefix = "Spyder-") as tmpdir:                
+                self._change_update_installation_status(status=DOWNLOADING_INSTALLER)
                 destination = os.path.join(tmpdir,'updateSpyder.exe')
                 download =urlretrieve(url_r,
                 destination,
                 reporthook=self._progress_reporter)
-                self._change_update_installation_status(status="Installing")
+                self._change_update_installation_status(status=INSTALLING)
                 install = subprocess.Popen(destination, shell=True)
                 install.communicate()
+            self._change_update_installation_status(status=FINISHED)
             self.application_update_status.setVisible(False)
         def _thread_launcher(option):
             if option.text() in ('&Yes'):
                 self.cancelled=False
-                """call a function in a simple thread, to prevent blocking"""
                 self.application_update_status.setVisible(True)
+                """call a function in a simple thread, to prevent blocking"""                
                 self.thread_install_update = threading.Thread(target = _download_install)                
                 self.thread_install_update.start()
         # Define the custom QMessageBox
--- a/spyder/plugins/application/container.py
+++ b/spyder/plugins/application/container.py
@@ -290,13 +290,14 @@ class ApplicationContainer(PluginMainContainer):
         else:
 
             if update_available:
-                self.application_update_status.set_status_pending()
+                self.application_update_status.set_status_pending(
+                    latest_release=latest_release)
 
                 header = _("<b>Spyder {} is available!</b> "
                            "<i>(you have {})</i><br><br>").format(
                     latest_release, __version__)
                 footer = _(
-                    "For more information visit our "
+                    "For more information, visit our "
                     "<a href=\"{}\">installation guide</a>."
                 ).format(url_i)
                 if is_anaconda():
--- a/spyder/plugins/application/container.py
+++ b/spyder/plugins/application/container.py
@@ -310,6 +310,7 @@ class ApplicationContainer(PluginMainContainer):
                         "<code>conda update anaconda</code><br>"
                         "<code>conda install spyder={}</code><br><br>"
                     ).format(latest_release)
+                    self.application_update_status.set_no_status()
                 else:
                     if os.name == 'nt' and is_pynsist():
                         box.setStandardButtons(QMessageBox.Yes |
@@ -318,13 +319,6 @@ class ApplicationContainer(PluginMainContainer):
                             "You want to download and install the latest "
                             "version of spyder?<br><br>"
                         )
-                        box.buttonClicked.connect(
-                            lambda button:
-                            self.application_update_status.start_installation()
-                            if button.text() in ('&Yes')
-                            else
-                            self.application_update_status.set_status_pending()
-                            )
                     else:
                         content = _(
                             "Click <a href=\"{}\">this link</a> to "
@@ -333,7 +327,11 @@ class ApplicationContainer(PluginMainContainer):
                 msg = header + content + footer
                 box.setText(msg)
                 box.set_check_visible(True)
-                box.show()
+                box.exec_()
+                if box.result() == QMessageBox.Yes:
+                    self.application_update_status.start_installation()
+                elif(box.result() == QMessageBox.No):
+                    self.application_update_status.set_status_pending()
                 check_updates = box.is_checked()
             elif feedback:
                 msg = _("Spyder is up to date.")
--- a/spyder/plugins/application/container.py
+++ b/spyder/plugins/application/container.py
@@ -78,13 +78,12 @@ class ApplicationActions:
     SpyderRestart = "Restart"
     SpyderRestartDebug = "Restart in debug mode"
 
-# Installation process statuses
+# Update Installation process statuses
 NO_STATUS = _("No status")
-DOWNLOADING_SCRIPT = _("Downloading script installer")
 DOWNLOADING_INSTALLER = _("Downloading installer")
 INSTALLING = _("Installing")
 FINISHED = _("Installation finished")
-ERRORED = _("Installation errored")
+CHECKING = _("Checking for updates")
 CANCELLED = _("Cancelled")
 
 class ApplicationContainer(PluginMainContainer):
@@ -282,6 +281,7 @@ class ApplicationContainer(PluginMainContainer):
         self.thread_install_update.join()
     
     def _check_updates_ready(self):
+        
         """Show results of the Spyder update checking process."""
 
         # `feedback` = False is used on startup, so only positive feedback is
@@ -290,6 +290,7 @@ class ApplicationContainer(PluginMainContainer):
         feedback = self.give_updates_feedback
         
         # Get results from worker
+        
         update_available = self.worker_updates.update_available
         latest_release = self.worker_updates.latest_release
         error_msg = self.worker_updates.error
@@ -304,8 +305,7 @@ class ApplicationContainer(PluginMainContainer):
         url_i = 'https://docs.spyder-ide.org/installation.html'      
 
         def _download_install():            
-            with TemporaryDirectory(prefix = "Spyder-") as tmpdir:                
-                self._change_update_installation_status(status=DOWNLOADING_INSTALLER)
+            with TemporaryDirectory(prefix = "Spyder-") as tmpdir:
                 destination = os.path.join(tmpdir,'updateSpyder.exe')
                 download =urlretrieve(url_r,
                 destination,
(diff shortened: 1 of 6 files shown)


jsbautista:
@dalthviz ; A preview of the process of checking for new updates:
Preview

dalthviz:
@jsbautista just in case, I think the tests are failling since for the new `application/widgets` dir you need to add a `__init__.py` file.

spyder/plugins/application/widgets/status.py
```diff
@@ -0,0 +1,82 @@
...
+    UpdateInstallerDialog)
+from spyder.utils.icon_manager import ima
+
+logger = logging.getLogger(__name__)
+
+
+class ApplicationUpdateStatus(StatusBarWidget):
+    """Status bar widget for Application update status."""
+    BASE_TOOLTIP = _("Application update status")
+    DEFAULT_STATUS = _('not reachable')
```
dalthviz, line 30: Maybe we can change this default status to something like the current Spyder version installed.

spyder/plugins/application/widgets/status.py
```diff
@@ -0,0 +1,82 @@
...
+        elif value == "Checking for updates" or value == "Update ready":
+            self.setVisible(True)
+            self.tooltip = value
+        else:
+            self.setVisible(False)
+            value = self.DEFAULT_STATUS
+            self.tooltip = self.BASE_TOOLTIP
+
+        self.update_tooltip()
+        value = "Update: {0}".format(value)
```
dalthviz, line 65: Maybe instead of `Update: ` here we can put `Spyder: `.

spyder/plugins/application/widgets/status.py
dalthviz, line 30: Maybe we can change this default status to something like the current Spyder version installed.

spyder/plugins/application/widgets/status.py
dalthviz, line 65: Maybe instead of `Update: ` here we can put `Spyder: `.

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,175 @@
...
+            self.cancel_button.hide()
+
+    def update_installation_progress(self, current_value, total):
+        """Update installation progress bar."""
+        self._progress_bar.setMaximum(total)
+        self._progress_bar.setValue(current_value)
+
+
+class UpdateInstallerDialog(QDialog):
+    """Kite installer."""
```
dalthviz, line 89: This probably should be Spyder installer

dalthviz:
Seems like there are some errors with the status constants @jsbautista. To check the tracebacks you can check the CI log: https://github.com/spyder-ide/spyder/runs/7567303687?check_suite_focus=true#step:15:1790

dalthviz:
@jsbautista I think that this needs a rebase on top of the latest 5.x to prevent errors in the CI. For that fetch in your local repo `upstream` and then rebase on top `upstream/5.x`. So running something like this should do it:

    git checkout 5.xUpdateSpyder
    git fetch upstream
    git rebase upstream/5.x

If that works without issue the you can force push the rebase with:

    git push -f

If you need help or some error message appears let me know :+1:

jsbautista:
Hi @dalthviz, could you give me a last review before removing the WIP?

dalthviz (commented):
Hi @jsbautista thanks for all the work on this! I left some suggestions and comments. I think the most important part to check is the location of the installation thread logic. Checking the code a little bit more seems to me that the status of the installation should be handle by the installation widget and we need to provide methods so the container can call some start method using the status widget (so we have a relation like container -> status widget -> installation widget).

If you have any questions or comments maybe we can do a pair programming session to test the ideas or check if there are more elements to take into account to handle better the interactions between container, status widget and installation widget.

Let me know!

jsbautista:
Hi @dalthviz

dalthviz (commented):
Thanks for all the work @jsbautista ! I think this is getting closer and closer to be finished :+1:

I left some suggestions regarding the calls to set the status and also some remaning logic position and status values suggestions (like using the Spyder version for the `NO_STATUS` constant). Besides that there are some PEP8 issue to be fixed.

Also, could you update the gif to preview the way this is working, please?

Thanks again! If you need anything to address the review, questions or comments, let me know!

jsbautista:
@dalthviz ; A preview of the process of checking for new updates:
UpdateSpyder

spyder/plugins/application/container.py
```diff
@@ -297,10 +311,25 @@ def _check_updates_ready(self):
...
+                            "You want to download and install the latest "
+                            "version of spyder?<br><br>"
+                        )
+                        box.buttonClicked.connect(
+                            lambda button:
+                            self.application_update_status.start_installation()
+                            if button.text() in ('&Yes')
+                            else
+                            self.application_update_status.set_status_pending()
+                            )
```
dalthviz, line 327: I think we could further simplify this by using the [`exec_` method](https://doc.qt.io/qt-5/qmessagebox.html#exec) instead of the `show` method on line 336. With that you can use the [`result` method](https://doc.qt.io/qt-5/qdialog.html#result) and compare the return value to the `QMessageBox.Yes` value. So this could be removed for something like the following at line 336:

```python
box.exec_()
if box.result() == QMessageBox.Yes:
    self.application_update_status.start_installation()

```

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,252 @@
...
+        reply.setWindowTitle("Spyder")
+        reply.setAttribute(Qt.WA_ShowWithoutActivating)
+        reply.setStandardButtons(QMessageBox.Yes | QMessageBox.No)
+        reply.buttonClicked.connect(
+                            lambda button:
+                            self.start_installation_update()
+                            if button.text() in ('&Yes')
+                            else self._change_update_installation_status(
+                                status=PENDING))
+        reply.show()
```
dalthviz, line 180: A similar approach as the one described at `container.py` could be used here (changing the usage of `show` for `exec_` and `result`)

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,250 @@
...
+        self._change_update_installation_status(status=CANCELLED)
+        self.thread_install_update.join()
+
+    def _download_install(self):
+        try:
+            with TemporaryDirectory(prefix="Spyder-") as tmpdir:
+                destination = os.path.join(tmpdir, 'updateSpyder.exe')
+                download = urlretrieve(
+                    ('https://github.com/spyder-ide/spyder/releases/latest/'
+                        'download/Spyder_64bit_full.exe'),
```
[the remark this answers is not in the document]
dalthviz: This is missing still I think? Probabably the validation at the end should use the `spyder.utils.programs.is_module_installed` function to check if `numpy` or `pandas` are installed (those modules come with the full type installer but are missing for the light type installer)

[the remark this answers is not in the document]
dalthviz: @jsbautista I think this is still missing or what do you think about this?

spyder/plugins/application/container.py
```diff
@@ -296,11 +310,21 @@ def _check_updates_ready(self):
                         "<code>conda update anaconda</code><br>"
                         "<code>conda install spyder={}</code><br><br>"
                     ).format(latest_release)
+                    self.application_update_status.set_no_status()
```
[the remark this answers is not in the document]
dalthviz: Is this set then needed @jsbautista ? (Seems like the conversation was resolved but the suggestion was not applied)

spyder/plugins/application/widgets/status.py
```diff
@@ -0,0 +1,95 @@
...
+        self.set_value(CHECKING)
+
+    def set_no_status(self):
+        self.set_value(NO_STATUS)
+
+    @Slot()
+    def show_installation_dialog(self):
+        """Show installation dialog."""
+        if (not self.tooltip == self.BASE_TOOLTIP and not
+                self.tooltip == PENDING):
```
dalthviz, line 92: Should we add here a validation to only call the `installer` if `is_pynsist` is `True` @jsbautista ?

jsbautista: ok, yes!!

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,258 @@
...
+
+# Local imports
+from spyder import __version__
+from spyder.config.base import _
+from spyder.utils.icon_manager import ima
+from spyder.utils.palette import QStylePalette
+from spyder.utils.programs import is_module_installed
+
+
+# Update installation process statuses
```
ccordoba12, line 31:
```suggestion
# Localization
_ = get_translation('spyder')

# Update installation process statuses
```

For translations.

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,258 @@
...
+            self.cancel_button.hide()
+
+    def update_installation_progress(self, current_value, total):
+        """Update installation progress bar."""
+        self._progress_bar.setMaximum(total)
+        self._progress_bar.setValue(current_value)
+
+
+class UpdateInstallerDialog(QDialog):
+    """Spyder installer."""
```
ccordoba12, line 99:
```suggestion
    """Update installer dialog."""
```

Improve this docstring a bit.

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,258 @@
...
+
+    # Signal to get the download progress
+    # int: Download progress
+    # int: Total download size
+    sig_download_progress = Signal(int, int)
+
+    # Signals
```
ccordoba12, line 106:
```suggestion
```

This is not necessary.

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,258 @@
...
+PENDING = _("Pending update")
+CHECKING = _("Checking for updates")
+CANCELLED = _("Cancelled")
+
+
+class UpdateInstallation(QWidget):
+    """Update progress installation widget."""
+
+    def __init__(self, parent):
+        super(UpdateInstallation, self).__init__(parent)
```
ccordoba12, line 45:
```suggestion
        super().__init__(parent)
```

That was the old way (i.e. Python 2 compatible) of using `__init__`, but it's no longer necessary.

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,258 @@
...
+    # Signal to get the current status of the update installation
+    # str: Status string
+    sig_installation_status = Signal(str)
+
+    def __init__(self, parent):
+
+        self.cancelled = False
+        self.status = NO_STATUS
+        self.thread_install_update = None
+        super(UpdateInstallerDialog, self).__init__(parent)
```
ccordoba12, line 116:
```suggestion
        super().__init__(parent)
```

That was the old way of using `__init__`.

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,258 @@
...
+        self.setStyleSheet(
+            f"background-color: {QStylePalette.COLOR_BACKGROUND_2}")
+        self.setWindowFlags(Qt.Dialog | Qt.MSWindowsFixedSizeDialogHint)
+        self._parent = parent
+        self._installation_widget = UpdateInstallation(self)
+
+        # Layout
+        installer_layout = QVBoxLayout()
+        installer_layout.addWidget(self._installation_widget)
+
```
ccordoba12, line 126:
```suggestion
```

This is not needed.

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,258 @@
...
+    def setup(self, installation=False):
+        """Setup visibility of widgets."""
+        self._installation_widget.setVisible(True)
+        self.adjustSize()
+
+    def cancel_install(self):
+        """Cancel the installation in progress."""
+        reply = QMessageBox.critical(
+            self._parent, 'Spyder',
+            _('Do you really want to cancel installing the Spyder update?'),
```
ccordoba12, line 154:
```suggestion
            _('Do you really want to cancel the Spyder update installation?'),
```

Make this a bit easier to read.

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,258 @@
...
+            self.cancelled = True
+            self.cancell_thread_install_update()
+            self.setup()
+            self.accept()
+            return True
+        return False
+
+    def continue_install(self):
+        """Continue the installation in progress
+        by downloading the installer."""
```
ccordoba12, line 166:
```suggestion
        """
        Continue the installation in progress by downloading the installer.
        """
```

Fix formatting.

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,258 @@
...
+        reply = QMessageBox(icon=QMessageBox.Question,
+                            text=_('Do you want to download and'
+                                   ' install the latest version of'
+                                   ' spyder?<br>'),
```
ccordoba12, line 170:
```suggestion
                                   ' Spyder?<br>'),
```

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,258 @@
...
+        """Close the installation dialog."""
+        if (self.status == FINISHED
+                or self.status == CANCELLED):
+            self.setup()
+            self.accept()
+        else:
+            self.hide()
+
+    def reject(self):
+        """Reimplement Qt method."""
```
ccordoba12, line 197:
```suggestion
        """Reimplemented Qt method."""
```

Fix grammar.

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,258 @@
...
+    def _progress_reporter(self, block_number, read_size, total_size):
+        progress = 0
+        if total_size > 0:
+            progress = block_number * read_size
+        if self.cancelled:
+            raise UpdateInstallationCancelledException()
+        else:
+            self.sig_download_progress.emit(progress, total_size)
+
+    def cancell_thread_install_update(self):
```
ccordoba12, line 218:
```suggestion
    def cancel_thread_install_update(self):
```

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,258 @@
...
+            status=DOWNLOADING_INSTALLER)
+        """call a function in a simple thread, to prevent blocking"""
+        self.thread_install_update = threading.Thread(
+            target=self._download_install)
+        self.thread_install_update.start()
+
+
+class UpdateInstallationCancelledException(Exception):
+    """Update installation was cancelled."""
+    pass
```
ccordoba12, line 258: Please move this class to be before `UpdateInstallation`. We tend define support classes before others and not after.

spyder/plugins/application/widgets/status.py
```diff
@@ -0,0 +1,95 @@
...
+"""
+
+# Standard library imports
+import logging
+
+# Third party imports
+from qtpy.QtCore import Slot
+
+# Local imports
+from spyder.api.widgets.status import StatusBarWidget
```
ccordoba12, line 18:
```suggestion
from spyder import get_versions
from spyder.api.widgets.status import StatusBarWidget
```

Move import below to be here so that it's sorted out alphabetically.

spyder/plugins/application/widgets/status.py
```diff
@@ -0,0 +1,95 @@
...
+from spyder.config.base import _
+from spyder.plugins.application.widgets.install import (
+    UpdateInstallerDialog, NO_STATUS, DOWNLOADING_INSTALLER, INSTALLING,
+    FINISHED, PENDING, CHECKING, CANCELLED)
+from spyder.utils.icon_manager import ima
+from spyder import get_versions
```
ccordoba12, line 24:
```suggestion
```

This was moved above.

spyder/plugins/application/widgets/status.py
```diff
@@ -0,0 +1,95 @@
...
+        self.cancelled = False
+        self.status = NO_STATUS
+        self.thread_install_update = None
+        self.tooltip = self.BASE_TOOLTIP
+        self._container = parent
+        super().__init__(parent)
+
+        # Installation dialog
+        self.installer = UpdateInstallerDialog(
+            self)
```
ccordoba12, line 45:
```suggestion
        self.installer = UpdateInstallerDialog(self)
```

This doesn't need to be in two lines.

spyder/plugins/application/widgets/status.py
```diff
@@ -0,0 +1,95 @@
...
+        self.update_tooltip()
+        value = "Spyder: {0}".format(value)
+        super().set_value(value)
+
+    def get_tooltip(self):
+        """Reimplementation to get a dynamic tooltip."""
+        return self.tooltip
+
+    def get_icon(self):
+        return ima.icon('spyder')
```
ccordoba12, line 74:
```suggestion
        return ima.icon('spyder_about')
```

This is our new icon. `spyder` is the old one.

ccordoba12 (commented):
Great work @jsbautista! This looks pretty good!! Most of my suggestions are style issues, otherwise looks good to me.

As a recommendation, please don't mark reviewer comments as solved. That's the work of reviewers, after checking that you correctly addressed their comments.

spyder/plugins/application/widgets/install.py
ccordoba12: This is still not addressed.

ccordoba12 (commented):
@jsbautista, last comments from my side.

spyder/plugins/application/widgets/status.py
```diff
@@ -0,0 +1,94 @@
...
+class ApplicationUpdateStatus(StatusBarWidget):
+    """Status bar widget for Application update status."""
+    BASE_TOOLTIP = _("Application update status")
+    ID = 'application_update_status'
+
+    def __init__(self, parent):
+
+        self.cancelled = False
+        self.status = NO_STATUS
+        self.thread_install_update = None
```
dalthviz, line 38: I think we could delete this attribute (is not used here anymore)

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,260 @@
...
+        """Reimplemented Qt method."""
+        on_installation_widget = self._installation_widget.isVisible()
+        if on_installation_widget:
+            self.close_installer()
+        else:
+            super(UpdateInstallerDialog, self).reject()
+
+    def _change_update_installation_status(self, status=NO_STATUS):
+        """Set the installation status."""
+        self.status = status
```
dalthviz, line 213: Lets add some debug logging here. For that, you will need to import the `logging` module and then at the begining of the file create a `logger` instance (similar setup at the one done in the `status.py` file). After that, we can add here something like:

```python
logger.debug(f"Installation status: {status}"))`
```

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,260 @@
...
+    def _download_install(self):
+        try:
+            with TemporaryDirectory(prefix="Spyder-") as tmpdir:
+                destination = os.path.join(tmpdir, 'updateSpyder.exe')
+                url = (
+                    'https://github.com/spyder-ide/spyder/releases/latest/'
+                    'download/Spyder_64bit_lite.exe')
+                val = (is_module_installed('numpy') or
+                       is_module_installed('pandas'))
+                if val:
```
dalthviz, line 238: Lets change this variable name to something like `is_full_installer`. So something like:

```suggestion
                is_full_installer = (is_module_installed('numpy') or
                                     is_module_installed('pandas'))
                if is_full_installer:
```

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,260 @@
...
+                    url = (
+                        'https://github.com/spyder-ide/spyder/releases/latest/'
+                        'download/Spyder_64bit_full.exe')
```
dalthviz, line 241: Lets add here another debbug entry checking on the url being used to download the installer. So something like:

```python
logger.debug(f"Downloading installer from: {url}")
```

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,260 @@
...
+        if self.cancelled:
+            raise UpdateInstallationCancelledException()
+        else:
+            self.sig_download_progress.emit(progress, total_size)
+
+    def cancel_thread_install_update(self):
+        self._change_update_installation_status(status=CANCELLED)
+        self.thread_install_update.join()
+
+    def _download_install(self):
```
dalthviz, line 229: Lets add here a debug log entry, something like:

```python
logger.debug("Downloading installer executable")
```

dalthviz (commented):
Thanks for all the work and addressing the reviews @jsbautista ! I left a review to add some debug log entries and also a couple of suggestions to change the variable name used to check the correct url from where download the installer and to delete an attribute that seems that is not used anymore.

I will try to do a final testing locally to check the overall functionality and if I don't find anything else then I would say this could be merged :+1:

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,265 @@
...
+                    url = (
+                        'https://github.com/spyder-ide/spyder/releases/latest/'
+                        'download/Spyder_64bit_full.exe')
+                logger.debug(f"Downloading installer from: {url}")
+                download = urlretrieve(url,
+                                       destination,
+                                       reporthook=self._progress_reporter)
+                self._change_update_installation_status(status=INSTALLING)
+                install = subprocess.Popen(destination, shell=True)
+                install.communicate()
```
dalthviz, line 252: Seems like the process ends when closing Spyder. Not totally sure if there is some `creationFlags` needed for the subprocess to continue after killing the Spyder process or if there is something we need to change over the installer assets .nsi template but probably this is the main thing to figure out @jsbautista

dalthviz (commented):
Hi @jsbautista , left another review adding some UI elements that I just rememered we can add (a spinner for the status bar) and also some comments regarding the text and the dialog style to fix the buttons appearing as the same color as the background.

However, most importantly, I checked locally and seems like when the installer closes Spyder is also closing itself :/ so we need to figure out what to do to prevent that while being sure that Spyder closes before proceeding to uninstall the previous version and then install the new one

spyder/plugins/application/container.py
```diff
@@ -27,7 +27,8 @@
 from spyder.api.widgets.main_container import PluginMainContainer
 from spyder.utils.installers import InstallerMissingDependencies
 from spyder.config.utils import is_anaconda
-from spyder.config.base import get_conf_path, get_debug_level
+from spyder.config.base import get_conf_path, get_debug_level, is_pynsist
```
mrclary, line 30:
```suggestion
from spyder.config.base import (get_conf_path, get_debug_level, is_pynsist,
                                running_in_mac_app)
```

spyder/plugins/application/container.py
```diff
@@ -297,21 +311,36 @@ def _check_updates_ready(self):
                         "<code>conda install spyder={}</code><br><br>"
                     ).format(latest_release)
                 else:
-                    content = _(
-                        "Click <a href=\"{}\">this link</a> to "
-                        "download it.<br><br>"
-                    ).format(url_r)
+                    if os.name == 'nt' and is_pynsist():
```
mrclary, line 314:
```suggestion
                    if is_pynsist() or running_in_mac_app():
```

mrclary: `os.name == 'nt'` is redundant with `is_pynsist()`.

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,265 @@
...
+# Licensed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+
+"""Update installation widgets."""
+
+# Standard library imports
+import logging
+import os
+import subprocess
+import sys
```
mrclary, line 13:
```suggestion
```

unused import

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,265 @@
...
+from tempfile import TemporaryDirectory
+import threading
+from urllib.request import urlretrieve
+
+# Third-party imports
+from qtpy.QtCore import QEvent, QObject, Qt, Signal
```
mrclary, line 19:
```suggestion
from qtpy.QtCore import Qt, Signal
```

unused imports

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,265 @@
...
+from qtpy.QtWidgets import (QDialog, QHBoxLayout, QMessageBox,
+                            QLabel, QProgressBar, QPushButton, QVBoxLayout,
+                            QWidget)
+
+# Local imports
+from spyder import __version__
+from spyder.api.translations import get_translation
+from spyder.utils.icon_manager import ima
+from spyder.utils.palette import QStylePalette
```
mrclary, line 28:
```suggestion
```

unused import

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,265 @@
...
+                if is_full_installer:
+                    url = (
+                        'https://github.com/spyder-ide/spyder/releases/latest/'
+                        'download/Spyder_64bit_full.exe')
+                logger.debug(f"Downloading installer from: {url}")
+                download = urlretrieve(url,
+                                       destination,
+                                       reporthook=self._progress_reporter)
+                self._change_update_installation_status(status=INSTALLING)
+                install = subprocess.Popen(destination, shell=True)
```
mrclary, line 249:
```suggestion
                is_full = (is_module_installed('numpy') or
                           is_module_installed('pandas'))
                if os.name == 'nt':
                    name = 'Spyder_64bit_{}.exe'.format('full' if is_full
                                                        else 'lite')
                else:
                    name = 'Spyder{}.dmg'.format('' if is_full else '-Lite')
                url = ('https://github.com/spyder-ide/spyder/releases/latest/'
                       f'download/{name}')
                destination = os.path.join(tmpdir, name)
                logger.debug(f"Downloading installer from: {url}")
                urlretrieve(url, destination,
                            reporthook=self._progress_reporter)
                self._change_update_installation_status(status=INSTALLING)

                cmd = ('' if os.name == 'nt' else 'open ') + destination
                install = subprocess.Popen(cmd, shell=True)
```

This should handle macOS as well.

spyder/plugins/application/widgets/status.py
```diff
@@ -0,0 +1,92 @@
...
+# Standard library imports
+import logging
+
+# Third party imports
+from qtpy.QtCore import Slot
+
+# Local imports
+from spyder import get_versions
+from spyder.api.widgets.status import StatusBarWidget
+from spyder.config.base import _, is_pynsist
```
mrclary, line 20:
```suggestion
from spyder.config.base import _, is_pynsist, running_in_mac_app
```

spyder/plugins/application/widgets/status.py
```diff
@@ -0,0 +1,92 @@
...
+    def set_no_status(self):
+        self.set_value(NO_STATUS)
+
+    @Slot()
+    def show_installation_dialog(self):
+        """Show installation dialog."""
+        if ((not self.tooltip == self.BASE_TOOLTIP and not
+                self.tooltip == PENDING) and is_pynsist):
+            self.installer.show()
+        elif ((self.tooltip == PENDING) and is_pynsist):
```
mrclary:
```suggestion
        if ((not self.tooltip == self.BASE_TOOLTIP
             and not self.tooltip == PENDING)
                and (is_pynsist() or running_in_mac_app())):
            self.installer.show()
        elif ((self.tooltip == PENDING)
              and (is_pynsist() or running_in_mac_app())):
```

mrclary (commented):
@jsbautista, I suggested some changes that should handle the macOS case. I believe they should work.

spyder/plugins/application/container.py
```diff
@@ -297,21 +311,36 @@ def _check_updates_ready(self):
...
+                elif(box.result() == QMessageBox.No):
+                    self.application_update_status.set_status_pending()
                 check_updates = box.is_checked()
             elif feedback:
                 msg = _("Spyder is up to date.")
                 box.setText(msg)
                 box.set_check_visible(False)
                 box.exec_()
                 check_updates = box.is_checked()
+                self.application_update_status.set_no_status()
```
dalthviz, line 343: Missing `else` in case `feedback` is `False` at startup @jsbautista :
```suggestion
                self.application_update_status.set_no_status()
            else:
                self.application_update_status.set_no_status()
```

spyder/plugins/application/container.py
```diff
@@ -297,21 +311,36 @@ def _check_updates_ready(self):
...
-                    content = _(
-                        "Click <a href=\"{}\">this link</a> to "
-                        "download it.<br><br>"
-                    ).format(url_r)
+                    if os.name == 'nt' and is_pynsist():
+                        box.setStandardButtons(QMessageBox.Yes |
+                                               QMessageBox.No)
+                        content = _(
+                            "You want to download and install the latest "
+                            "version of Spyder?<br><br>"
```
CAM-Gerlach, line 319:
```suggestion
                            "Would you like to automatically download and install"
                            "the latest version of Spyder?<br><br>"
```

Fix bad grammar and clarify that it will be done automatically for you

spyder/plugins/application/widgets/install.py
dalthviz: Just in case @jsbautista checking this locally changing the `.nsi` script repo to exclude the `updateSpyder.exe` from being killed over the Windows assets is what was missing. Adding the exclution the installation continues after the installer closes Spyder 🎉

mrclary:
@jsbautista @dalthviz, can you guys advise on how to test this PR? I'd like to check that the this downloads and launches the DMG for macOS, but I don't know how to artificially trigger the update notification.

dalthviz:
> can you guys advise on how to test this PR? I'd like to check that the this downloads and launches the DMG for macOS, but I don't know how to artificially trigger the update notification.

@mrclary on Windows I was testing this by changing the `__version__` variable at `spyder/__init__.py` to `5.3.1` from the installed version

spyder/plugins/application/widgets/status.py
```diff
@@ -0,0 +1,100 @@
...
+from spyder.utils.icon_manager import ima
+
+logger = logging.getLogger(__name__)
+
+
+class ApplicationUpdateStatus(StatusBarWidget):
+    """Status bar widget for Application update status."""
+    BASE_TOOLTIP = _("Application update status")
+    ID = 'application_update_status'
+
```
dalthviz, line 33: Following another suggestion from @CAM-Gerlach lets add the possibility to call the `Check from updates...` action  from the status bar. For that we could add here a signal that then we can connect from the container to the `check_update` function:

```suggestion

    sig_check_for_updates_requested = Signal()

```

spyder/plugins/application/widgets/status.py
```diff
@@ -0,0 +1,100 @@
...
+"""
+Status widget for Spyder updates.
+"""
+
+# Standard library imports
+import logging
+
+# Third party imports
+from qtpy.QtCore import Slot
```
dalthviz, line 15: To enable showing a menu from the status bar and adding a new `Signal`:
```suggestion
from qtpy.QtCore import QPoint, Signal, Slot
from qtpy.QtWidgets import QMenu
```

spyder/plugins/application/widgets/status.py
```diff
@@ -0,0 +1,100 @@
...
+
+# Local imports
+from spyder import get_versions
```
dalthviz, line 18: Let remove this unused import:
```suggestion

```

spyder/plugins/application/widgets/status.py
```diff
@@ -0,0 +1,100 @@
...
+from spyder.api.widgets.status import StatusBarWidget
+from spyder.config.base import _, is_pynsist
+from spyder.plugins.application.widgets.install import (
+    UpdateInstallerDialog, NO_STATUS, DOWNLOADING_INSTALLER, INSTALLING,
+    FINISHED, PENDING, CHECKING, CANCELLED)
+from spyder.utils.icon_manager import ima
```
dalthviz, line 24: New imports for the menu to show the `Check for updates...` action:
```suggestion
from spyder.utils.icon_manager import ima
from spyder.utils.qthelpers import add_actions, create_action
```

spyder/plugins/application/widgets/status.py
```diff
@@ -0,0 +1,100 @@
...
+    ID = 'application_update_status'
+
+    def __init__(self, parent):
+
+        self.tooltip = self.BASE_TOOLTIP
+        super().__init__(parent, show_spinner=True)
+
+        # Installation dialog
+        self.installer = UpdateInstallerDialog(self)
+
```
dalthviz, line 41: New attribute to hold a menu:
```suggestion
        # Check for updates action menu
        self.menu = QMenu(self)

```

spyder/plugins/application/widgets/status.py
```diff
@@ -0,0 +1,100 @@
...
+        self.sig_clicked.connect(self.show_installation_dialog)
```
dalthviz, line 42: Lets change the method name from `show_installation_dialog` to `show_installation_dialog_or_menu`:
```suggestion
        self.sig_clicked.connect(self.show_installation_dialog_or_menu)
```

spyder/plugins/application/widgets/status.py
```diff
@@ -0,0 +1,100 @@
...
+        self.spinner.stop()
+
+    @Slot()
+    def show_installation_dialog(self):
+        """Show installation dialog."""
+        if ((not self.tooltip == self.BASE_TOOLTIP and not
+                self.tooltip == PENDING) and is_pynsist()):
+            self.installer.show()
+        elif ((self.tooltip == PENDING) and is_pynsist()):
+            self.installer.continue_install()
```
dalthviz, line 100: To add the menu, and also use the actual value set for the validations then this method will become:
```suggestion
    def show_installation_dialog_or_menu(self):
        """Show installation dialog or menu."""
        value = self.value.split(":")[-1].strip()
        if (not self.tooltip == self.BASE_TOOLTIP and not
                value == PENDING and is_pynsist()):
            self.installer.show()
        elif value == PENDING and is_pynsist():
            self.installer.continue_install()
        elif value == NO_STATUS:
            self.menu.clear()
            check_for_updates_action = create_action(
                self,
                text=_("Check for updates..."),
                triggered=self.sig_check_for_updates_requested.emit
            )
            add_actions(self.menu, [check_for_updates_action])
            rect = self.contentsRect()
            os_height = 7 if os.name == 'nt' else 12
            pos = self.mapToGlobal(
                rect.topLeft() + QPoint(-10, -rect.height() - os_height))
            self.menu.popup(pos)
```

spyder/plugins/application/container.py
```diff
@@ -99,6 +105,10 @@ def __init__(self, name, plugin, parent=None):
     # ---- PluginMainContainer API
     # -------------------------------------------------------------------------
     def setup(self):
+
+        self.application_update_status = ApplicationUpdateStatus(parent=self)
```
dalthviz, line 109: To use the check_updates method from the status bar taking into account the other changes then here we just need to do the signal connection:
```suggestion
        self.application_update_status = ApplicationUpdateStatus(parent=self)
        self.application_update_status.sig_check_for_updates_requested.connect(
            self.check_updates
        )
```

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,271 @@
...
+
+# Localization
+_ = get_translation('spyder')
+
+# Update installation process statuses
+NO_STATUS = __version__
+DOWNLOADING_INSTALLER = _("Downloading installer")
+INSTALLING = _("Installing")
+FINISHED = _("Installation finished")
+PENDING = _("Pending update")
```
dalthviz, line 39: Lets change these values following the suggestions from @CAM-Gerlach:
```suggestion
DOWNLOADING_INSTALLER = _("Downloading update")
INSTALLING = _("Installing update")
FINISHED = _("Installation finished")
PENDING = _("Update available")
```

dalthviz (commented):
Hi @jsbautista another round of comments! Most of them are related with to adding to the status bar a menu to be able to launch the check for updates logic and also some changes to the state constants values following the feedback at #19112 

I think after applying these suggestions we can go ahead and check the changes suggested to support the MacOS installer download :+1:

If you need any help let me know!

spyder/plugins/application/widgets/status.py
```diff
@@ -0,0 +1,119 @@
...
+# Licensed under the terms of the MIT License
+# (see spyder/__init__.py for details)
+
+"""
+Status widget for Spyder updates.
+"""
+
+# Standard library imports
+import logging
+
```
dalthviz, line 13: I missed suggesting an import in my last review @jsbautista 😅 
```suggestion
import os

```

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,274 @@
...
+            url = ('https://github.com/spyder-ide/spyder/releases/latest/'
+                   f'download/{name}')
+            destination = os.path.join(tmpdir, name)
+            logger.debug(f"Downloading installer from: {url}")
+            download = urlretrieve(url,
+                                   destination,
+                                   reporthook=self._progress_reporter)
+            self._change_update_installation_status(status=INSTALLING)
+            cmd = ('start' if os.name == 'nt' else 'open')
+            subprocess.run([cmd, destination], shell=True)
```
mrclary, line 260: Seems that the command syntax was incorrect. `shell=True` requires a single string command; `shell=False` requires a list of command elements.
We can either use
```suggestion
            subprocess.run(' '.join([cmd, destination]), shell=True)
```
Or
```suggestion
            subprocess.run([cmd, destination])
```
Can you confirm that these also work for Windows?

mrclary (commented):
@jsbautista @dalthviz, I suggested a syntax fix. With the fix, I can confirm that the update and download work as expected on macOS.

spyder/plugins/application/widgets/install.py
dalthviz: Checked on Windows and I think we will need to go with the option that still uses `shell=True` (since on Windows an error - [WinError 2] - raises without it). Could you apply the suggestion that adds the `' '.join` call @jsbautista ? 

Also, thanks for checking @mrclary !

spyder/plugins/application/container.py
```diff
@@ -89,6 +91,11 @@ class ApplicationContainer(PluginMainContainer):
     Signal to load a log file
     """
 
+    sig_show_status_bar_widget = Signal(bool)
+    """
+    Signal to show the widget in status bar
+    """
+
```
dalthviz, line 98: I think this is not being used anymore so lets remove it:
```suggestion

```

spyder/plugins/application/container.py
```diff
@@ -297,22 +315,38 @@ def _check_updates_ready(self):
...
+                                               QMessageBox.No)
+                        content = _(
+                            "Would you like to automatically download and"
+                            " install the latest version of Spyder?<br><br>"
+                        )
+                    else:
+                        content = _(
+                            "Click <a href=\"{}\">this link</a> to "
+                            "download it.<br><br>"
+                        ).format(url_r)
```
dalthviz, line 329: I think this is not needed anymore as well as the `url_r` variable definition above, so:
```suggestion

```

And also lines 266-273 should be deleted

spyder/plugins/application/container.py
```diff
@@ -297,22 +315,38 @@ def _check_updates_ready(self):
                         "<code>conda install spyder={}</code><br><br>"
                     ).format(latest_release)
                 else:
-                    content = _(
-                        "Click <a href=\"{}\">this link</a> to "
-                        "download it.<br><br>"
-                    ).format(url_r)
+                    if is_pynsist() or running_in_mac_app():
```
dalthviz, line 318: I think we could change this from:

```python
else: 
      if is_pynsist() or running_in_mac_app():

```

to:

```python
elif is_pynsist() or running_in_mac_app():

```

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,274 @@
...
+
+        self.setLayout(general_layout)
+
+    def update_installation_status(self, status):
+        """Update installation status (downloading, installing, finished)."""
+        self._progress_label.setText(status)
+        self.install_info.setText(INSTALL_INFO_MESSAGES[status])
+        if status == INSTALLING:
+            self._progress_bar.setRange(0, 0)
+            self.cancel_button.hide()
```
dalthviz, line 108: Lets change this so instead of hidding it we just disable the button:
```suggestion
            self.cancel_button.setEnabled(False)
```

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,274 @@
...
+            self._installation_widget.update_installation_status)
+        self.sig_installation_status.connect(
+            self.finished_installation)
+
+        self._installation_widget.ok_button.clicked.connect(
+            self.close_installer)
+        self._installation_widget.cancel_button.clicked.connect(
+            self.cancel_install)
+
+        # Show integration widget
```
dalthviz, line 156: I think this should be something like:
```suggestion
        # Show installation widget
```

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,274 @@
...
+        self.setup()
+
+    def setup(self, installation=False):
```
dalthviz, line 159: I think the kwarg `installation` is not used anymore so:
```suggestion
    def setup(self):
```

dalthviz (commented):
Left some more comments @jsbautista ! Mostly regarding some missing things to remove (since now we handle here also the MacOS installer and other piece of code unused from what I can see)

dalthviz:
Just in case, a preview of the way this is working on Windows:

* Automatic check for updates enabled, but keep update pending and then launch it from the status bar:

    auto

* Cancel download:

    cancel

* Automatic check for updates disabled, access to the `Check for updates...` action from the status bar and start download:

    auto-disabled

* Installation process starting (which comes automatically after the download finishes):

    installing

@mrclary cloud it be possible for you to do some .gif or screen recording to check how things are looking on MacOS? Maybe just the part related to the download finishing and the .dmg running. I tried checking on my MacOS vm but I'm not being able to launch Spyder if I install the .dmg from PRs. I get a dialog saying that the process should be terminated:

imagen

Maybe I'm doing something wrong/missing some setup on my MacOS vm (I think I'm running Big Sur 11.0.1) to be able to run Spyder from a PR .dmg installation?

spyder/plugins/application/widgets/install.py
```diff
@@ -250,11 +250,15 @@ def _download_install(self):
...
-                                   reporthook=self._progress_reporter)
+            path_destination = os.path.join(tmpdir, 'spyder',
+                                            self.latest_release_version)
+            os.makedirs(path_destination, exist_ok=True)
+            destination = os.path.join(path_destination, name)
+            if (not os.path.isfile(destination)):
+                logger.debug(f"Downloading installer from: {url}")
+                download = urlretrieve(url,
+                                       destination,
+                                       reporthook=self._progress_reporter)
```
dalthviz, line 261: Lets use some more self explanatory variables names here, add another intermidiate level dir (`updates`) and also add more info about where the executable gets downloaded to the debug log entry. So this will end up something like:

```suggestion
            installer_dir_path = os.path.join(tmpdir, 'spyder', 'updates',
                                              self.latest_release_version)
            os.makedirs(installer_dir_path, exist_ok=True)
            installer_path = os.path.join(installer_dir_path, name)
            if (not os.path.isfile(installer_path)):
                logger.debug(f"Downloading installer from {url} to {installer_path}")
                download = urlretrieve(url,
                                       installer_path,
                                       reporthook=self._progress_reporter)
```

dalthviz: Also, I think we need to have a way to remove old executable versions downloaded. Probably, besides the validation to do or not the download, we also need to validate if there are more folders under the `spyder/updates` dir and remove them if they are not the latest release. What do you think @jsbautista ?

dalthviz (commented):
Hi @jsbautista ! Left a comment regarding the latest changes with a suggestion for the variables names and the addition of a new directory call `updates` in the path to be able to contain the directories for the different executables that could be downloaded. Also, left a comment regarding that we should add some logic to clean/remove the directories from previous dowloaded executables that could not be needed anymore (old versions).

If you have any comments or need help to apply the suggestions/comments let me know!

Other than that, I think a final manual test is missing and after that this could be merged :+1:

Also, something that is still pending from this whole new update flow is the division of the download and install process so users can decide if them want to proceed with the installation right now, when Spyder closes or not proceed at all, and how we should handle the case when the user says no, closes Spyder and launches it again.

Since that part has some initial ideas but probably will need more discussion and feedback I would say that we should leave it for a follow-up PR. Is that okay @mrclary @CAM-Gerlach @ccordoba12 ?

spyder/plugins/application/widgets/install.py
mrclary: > Also, something that is still pending from this whole new update flow is the division of the download and install process so users can decide if them want to proceed with the installation right now, when Spyder closes or not proceed at all, and how we should handle the case when the user says no, closes Spyder and launches it again.
> 
> Since that part has some initial ideas but probably will need more discussion and feedback I would say that we should leave it for a follow-up PR. Is that okay @mrclary @CAM-Gerlach @ccordoba12 ?

I think that is a good idea. This will require new/updated widgets as well as refactored logic. That will give us more time to discuss and finalize the UX that we want. A follow-up PR is my recommendation as well.

jsbautista: @dalthviz I think that this could work very well, besides that with this there will be a maximum of one folder with the latest version checked

spyder/plugins/application/widgets/install.py
CAM-Gerlach: As I can confirm in my testing, `start` is a shell-level command, so we do need the shell to execute it. However, we should really try to avoid `shell=True` if at all possible, as it is less performant, less reliable and there are [security issues](https://docs.python.org/3.12/library/subprocess.html#security-considerations) with (though this case is probably safe—though I can't figure out where `destination` comes from here. Is there any reason we need to use `start` from the shell here instead of just executing what we need directly with a fully qualified path?

spyder/plugins/application/widgets/install.py
CAM-Gerlach: > Since that part has some initial ideas but probably will need more discussion and feedback I would say that we should leave it for a follow-up PR. Is that okay @mrclary @CAM-Gerlach @ccordoba12 ?

Sure, though we really should get the UX in a good state before we actually release 5.4.0 to users, which as I understand is the plan here.

> Also, I think we need to have a way to remove old executable versions downloaded. Probably, besides the validation to do or not the download, we also need to validate if there are more folders under the spyder/updates dir and remove them if they are not the latest release. What do you think @jsbautista ?

Could be a good idea to at least cache the current/previous release in case something goes wrong and the user wants to roll back to the same version as before; that's what I see most other apps that do this do.

mrclary:
I see the progress bar, but the dmg is not opened. Let me take a look at what is going wrong there...

mrclary:
FYI, @dalthviz, I think the reason that your locally built macOS installer is failing is because the python-lsp-server was installed in editable mode during the build. That should not matter, but I noticed that for some reason this is causing a problem. The workaround is to install it in standard mode in the build environment. If you are in the macOS installer directory with the build environment activated:
```bash
$ python ../../install_dev_repos.py --not-editable --install python-lsp-server
```

mrclary:
On macOS, with my latest code suggestion...
updater

spyder/plugins/application/container.py
```diff
@@ -296,23 +309,34 @@ def _check_updates_ready(self):
...
-                else:
+                elif is_pynsist() or running_in_mac_app():
+                    box.setStandardButtons(QMessageBox.Yes |
+                                           QMessageBox.No)
                     content = _(
-                        "Click <a href=\"{}\">this link</a> to "
-                        "download it.<br><br>"
-                    ).format(url_r)
+                        "Would you like to automatically download and"
+                        " install the latest version of Spyder?<br><br>"
```
CAM-Gerlach, line 317:
```suggestion
                        "Would you like to automatically download and "
                        "install it?<br><br>"
```

We already say "Spyder X.Y.Z is available" in the previous sentence, so it sounds strange, confusing and repetitive to say "the latest version of Spyder" here. Also, fix whitespace to be more conventional.

spyder/plugins/application/container.py
```diff
@@ -278,7 +288,10 @@ def _check_updates_ready(self):
             box.exec_()
             check_updates = box.is_checked()
         else:
+
             if update_available:
+                self.application_update_status.set_status_pending()
+
                 header = _("<b>Spyder {} is available!</b><br><br>").format(
```
CAM-Gerlach, line 295: Can we say something like

```html
<b>Spyder {} is available!</b> <i>(you have {})</i>
```

so the user knows what version of Spyder they will be upgrading _from_? This is very common and often expected on update dialogs.

CAM-Gerlach (commented):
I had some final comments, suggestions and questions regarding the UI text...

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,279 @@
...
+
+INSTALL_INFO_MESSAGES = {
+    DOWNLOADING_INSTALLER: _("Downloading Spyder latest "
+                             "release installer executable"),
+    INSTALLING: _("Installing Spyder latest release"),
+    FINISHED: _("Spyder update installation finished"),
+    PENDING: _("An update is pending"),
+    CHECKING: _("Checking for updates"),
+    CANCELLED: _("Update cancelled")
+}
```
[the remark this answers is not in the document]
dalthviz: This `INSTALL_INFO_MESSAGES` should be a dict with the descriptions of the status available above (the contents should be improved for sure :+1:). Maybe changing the constant name for something like `STATUS_DESCRIPTIONS` could make things more clear? The idea is to have a different text than the status itself that can be used for the different widgets (like for example the one that shows the progress on downloading/installing) to show some description about what is happening in an specific status.

spyder/plugins/application/widgets/install.py
dalthviz: I think that was the way @jsbautista found to be able to launch the installer and keep it running even when it forces Spyder to exit (on the Windows case where the NSIS installer validates if Spyder is running and closes it if the user agrees)

dalthviz:
> On macOS, with my latest code suggestion...

Is this suggestion already applied @mrclary ? (I'm not totally sure where is was suggested). Is the one to use `' '.join...` for the command arguments?

mrclary:
> > On macOS, with my latest code suggestion...
> 
> Is this suggestion already applied @mrclary ? (I'm not totally sure where is was suggested). Is the one to use `' '.join...` for the command arguments?

Yes, it looks like it was fixed. I was not referring to `' '.join...` but to incorrect naming `destination` variable for the temporary downloaded file.

mrclary:
> > > On macOS, with my latest code suggestion...
> > 
> > 
> > Is this suggestion already applied @mrclary ? (I'm not totally sure where is was suggested). Is the one to use `' '.join...` for the command arguments?
> 
> Yes, it looks like it was fixed. I was not referring to `' '.join...` but to incorrect naming `destination` variable for the temporary downloaded file.

Sorry, that was my mistake. I forgot to "submit" my review that had the codes suggestion.

spyder/plugins/application/container.py
```diff
@@ -278,9 +288,13 @@ def _check_updates_ready(self):
...
         else:
+
             if update_available:
-                header = _("<b>Spyder {} is available!</b><br><br>").format(
-                    latest_release)
+                self.application_update_status.set_status_pending()
+
+                header = _("<b>Spyder {} is available!</b> ",
+                           "<i>(you have {})</i><br><br>").format(
+                    latest_release,)
```
dalthviz, line 297: I think here `__version__` is missing? So maybe @jsbautista this should be:
```suggestion
                    latest_release, __version__)
```

CAM-Gerlach: Yeah, I would have made it as a suggestion but it required importing `__version__` above, which I couldn't reach

spyder/plugins/application/widgets/install.py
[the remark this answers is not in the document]
CAM-Gerlach: Oh okay, thanks. So these would be used by the tooltip and such? I'll make a suggestion to fix these, thanks:

```suggestion
NO_STATUS = __version__
DOWNLOADING_INSTALLER = _("Downloading update")
INSTALLING = _("Installing update")
FINISHED = _("Installation finished")
PENDING = _("Update available")
CHECKING = _("Checking for updates")
CANCELLED = _("Cancelled update")

INSTALL_INFO_MESSAGES = {
    DOWNLOADING_INSTALLER: _("Downloading latest Spyder update"),
    INSTALLING: _("Installing Spyder update"),
    FINISHED: _("Spyder update installation finished"),
    PENDING: _("Spyder update available to download"),
    CHECKING: _("Checking for Spyder updates"),
    CANCELLED: _("Spyder update cancelled")
}
```

CAM-Gerlach:
I've made a [suggestion](https://github.com/spyder-ide/spyder/pull/18619#discussion_r961015932) to fix the description messages, FYI

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,279 @@
...
+CANCELLED = _("Cancelled")
+
+INSTALL_INFO_MESSAGES = {
+    DOWNLOADING_INSTALLER: _("Downloading Spyder latest "
+                             "release installer executable"),
+    INSTALLING: _("Installing Spyder latest release"),
+    FINISHED: _("Spyder update installation finished"),
+    PENDING: _("An update is pending"),
+    CHECKING: _("Checking for updates"),
+    CANCELLED: _("Update cancelled")
```
CAM-Gerlach, line 50: Making this as a seperate comment so its easier to find on this long PR

```suggestion
CANCELLED = _("Cancelled update")

INSTALL_INFO_MESSAGES = {
    DOWNLOADING_INSTALLER: _("Downloading latest Spyder update"),
    INSTALLING: _("Installing Spyder update"),
    FINISHED: _("Spyder update installation finished"),
    PENDING: _("Spyder update available to download"),
    CHECKING: _("Checking for Spyder updates"),
    CANCELLED: _("Spyder update cancelled")
```

CAM-Gerlach:
Also, just to confirm @dalthviz @ccordoba12 once this is ready, this should either have a major rebase to consolidate the many "WIP" commits into atomic, well-described units before merging, or just be squash merged.

dalthviz:
@CAM-Gerlach this will be squash merged :+1:

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,286 @@
...
+                            text=_("Would you like to automatically download "
+                                   "and install the latest version of Spyder?"
+                                   "<br><br>"),
+                            parent=self._parent)
+        reply.setWindowTitle("Spyder")
+        reply.setAttribute(Qt.WA_ShowWithoutActivating)
+        reply.setStandardButtons(QMessageBox.Yes | QMessageBox.No)
+        reply.exec_()
+        if reply.result() == QMessageBox.Yes:
+            self.start_installation_update()
```
dalthviz, line 191: This line is triggering an `TypeError` exception since the `latest_release_version` arg is not being passed:

```
  File "C:\Users\dalth\AppData\Local\Programs\Spyder\pkgs\spyder\plugins\application\widgets\status.py", line 111, in show_installation_dialog_or_menu
    self.installer.continue_install()
  File "C:\Users\dalth\AppData\Local\Programs\Spyder\pkgs\spyder\plugins\application\widgets\install.py", line 191, in continue_install
    self.start_installation_update()
TypeError: start_installation_update() missing 1 required positional argument: 'latest_release_version'
```

I think we should, from the container, pass the `latest_release_version` to the status widget/installer widget even if the user doesn't accept immediatly the first dialog to update (the one that informs that a new version is available)

dalthviz (commented):
Hi @jsbautista ! Left some comments regarding some issues raising when testing the latest changes.

Also, just in case, a preview for others to see how this will be working now to reuse a previously downloaded installer executable:

* The installer executable was downloaded in a previous session but not installed and the check update dialog appears:

initial

* In case the installer was downloaded in the current session or a previous one but the user decided to not proceed with the installation update (in case of the Windows installer the user choosed to not close Spyder for example or choosed `No` in the check update dialog):

continue

spyder/plugins/application/container.py
```diff
@@ -278,9 +288,13 @@ def _check_updates_ready(self):
             box.exec_()
             check_updates = box.is_checked()
         else:
+
             if update_available:
-                header = _("<b>Spyder {} is available!</b><br><br>").format(
-                    latest_release)
+                self.application_update_status.set_status_pending()
```
dalthviz, line 293: @jsbautista seems like passing `latest_release` is missing in this call, so:
```suggestion
                self.application_update_status.set_status_pending(
                    latest_release=latest_release)
```
Without that an error raises:
```
  File "C:\Program Files\Spyder\pkgs\spyder\plugins\application\container.py", line 293, in _check_updates_ready
    self.application_update_status.set_status_pending()
TypeError: set_status_pending() missing 1 required positional argument: 'latest_release'
```

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,289 @@
...
+CHECKING = _("Checking for updates")
+CANCELLED = _("Cancelled update")
+
+INSTALL_INFO_MESSAGES = {
+    DOWNLOADING_INSTALLER: _("Downloading latest Spyder update"),
+    INSTALLING: _("Installing Spyder update"),
+    FINISHED: _("Spyder update installation finished"),
+    PENDING: _("Spyder update available to download"),
+    CHECKING: _("Checking for Spyder updates"),
+    CANCELLED: _("Spyder update cancelled")
```
CAM-Gerlach, line 49:
```suggestion
    DOWNLOADING_INSTALLER: _("Downloading Spyder {version}"),
    INSTALLING: _("Installing Spyder {version}"),
    FINISHED: _("Finished installing Spyder {version}"),
    PENDING: _("Spyder {version} available to download"),
    CHECKING: _("Checking for new Spyder version"),
    CANCELLED: _("Spyder update cancelled")
```

Will need changes (something like `message.format(version=version)` where this is used) but as discussed in our meeting, this will avoid redundancy with the other message and provide useful information.

spyder/plugins/application/container.py
```diff
@@ -278,9 +288,13 @@ def _check_updates_ready(self):
...
+
+                header = _("<b>Spyder {} is available!</b> "
+                           "<i>(you have {})</i><br><br>").format(
+                    latest_release, __version__)
                 footer = _(
                     "For more information visit our "
```
CAM-Gerlach, line 299:
```suggestion
                    "For more information, visit our "
```

Missing comma

dalthviz:
Just in case @jsbautista seems like the error to generated the Windows installer is unrelated to the changes on the PR but to the PyPI json API to retrieve a release/info about a package for `certifi` which had a release some hours ago.

spyder/plugins/application/widgets/install.py
```diff
@@ -0,0 +1,290 @@
...
+        general_layout = QHBoxLayout()
+        general_layout.addLayout(action_layout)
+
+        self.setLayout(general_layout)
+
+    def update_installation_status(self, status, latestVersion):
+        """Update installation status (downloading, installing, finished)."""
+        self._progress_label.setText(status)
+        self.install_info.setText(INSTALL_INFO_MESSAGES[status].format(
+            version=latestVersion))
```
dalthviz, line 105: Lets change the variable name here to `latest_version` instead of `latestVersion`. So:
```suggestion
    def update_installation_status(self, status, latest_version):
        """Update installation status (downloading, installing, finished)."""
        self._progress_label.setText(status)
        self.install_info.setText(INSTALL_INFO_MESSAGES[status].format(
            version=latest_version))
```

dalthviz (commented):
Thanks for all the changes @jsbautista ! I left a simple comment for the latest changes regarding a variable name.

Also, as a suggestion we were discussing with the others, we should leave the update related status bar widget to the left of the rest of status bar widgets. Checking, for that we will need to add some entries in a couple of iterables in the status plugin. Basically, we need change the set and list at:

https://github.com/spyder-ide/spyder/blob/38e3adf82a9cae716a10f2773f860a8b6d897308/spyder/plugins/statusbar/plugin.py#L49-L52

And

https://github.com/spyder-ide/spyder/blob/38e3adf82a9cae716a10f2773f860a8b6d897308/spyder/plugins/statusbar/plugin.py#L216-L219

Adding to them the following entries: `'interpreter_status', 'application_update_status'`

With that the status widget for the updates will be shown at the left of the others:

imagen

If you have any question or comment let me know!

dalthviz (approved):
Thanks @jsbautista ! I think this is ready for a squash merge!

Do you have any other comments @mrclary @CAM-Gerlach @ccordoba12 ?

mrclary:
I'm good

CAM-Gerlach (approved):
Seems to look good from a UX and UI text perspective, with some improvements to follow in a followup PR. I haven't had the time to fully QA test it myself on a Windows machine, but it seems others have.

Issue #18661, opened 2022-07-14
Code signature breaks micromamba in macOS application


mrclary:
### What steps will reproduce the problem?

Code signing has broken the micromamba executable in the macOS application.
This means that the IPython Console will not work with external conda environments.

```
dyld: Library not loaded: @rpath/libc++.1.dylib
Referenced from: /Users/rclary/Applications/Spyder.app/Contents/Resources/lib/python3.9/spyder/bin/micromamba
Reason: image not found
```

## Versions

* Spyder version: 5.3.2 1172bb467 (standalone)
* Python version: 3.9.5 64-bit
* Qt version: 5.15.2
* PyQt5 version: 5.15.7
* Operating System: Darwin 20.6.0

### Dependencies

```
# Mandatory:
atomicwrites >=1.2.0                 :  1.4.1 (OK)
chardet >=2.0.0                      :  5.0.0 (OK)
cloudpickle >=0.5.0                  :  2.1.0 (OK)
cookiecutter >=1.6.0                 :  2.1.1 (OK)
diff_match_patch >=20181111          :  20200713 (OK)
intervaltree >=3.0.2                 :  3.1.0 (OK)
IPython >=7.31.1;<8.0.0              :  7.34.0 (OK)
jedi >=0.17.2;<0.19.0                :  0.18.1 (OK)
jellyfish >=0.7                      :  0.9.0 (OK)
jsonschema >=3.2.0                   :  4.7.2 (OK)
keyring >=17.0.0                     :  23.6.0 (OK)
nbconvert >=4.0                      :  6.5.0 (OK)
numpydoc >=0.6.0                     :  1.4.0 (OK)
parso >=0.7.0;<0.9.0                 :  0.8.3 (OK)
pexpect >=4.4.0                      :  4.8.0 (OK)
pickleshare >=0.4                    :  0.7.5 (OK)
psutil >=5.3                         :  5.9.1 (OK)
pygments >=2.0                       :  2.12.0 (OK)
pylint >=2.5.0;<3.0                  :  2.14.4 (OK)
pyls_spyder >=0.4.0                  :  0.4.0 (OK)
pylsp >=1.5.0;<1.6.0                 :  1.5.0 (OK)
pylsp_black >=1.2.0                  :  1.2.1 (OK)
qdarkstyle >=3.0.2;<3.1.0            :  3.0.3 (OK)
qstylizer >=0.1.10                   :  0.2.1 (OK)
[...]

PR #18679, opened 2022-07-15
PR: Fake release

mrclary:
DO NOT MERGE THIS PR!

- This is a retroactive fix for #18661 for [5.3.2](https://github.com/spyder-ide/spyder/releases/tag/v5.3.2) only.
- This is a retroactive fix for #18776 for [5.3.2](https://github.com/spyder-ide/spyder/releases/tag/v5.3.2) only.

Manually replace the 5.3.2 release macOS artifacts with the artifacts from this PR.

### DIFF
--- a/.github/workflows/installer-macos.yml
+++ b/.github/workflows/installer-macos.yml
@@ -27,7 +27,7 @@ jobs:
     steps:
     - id: set-matrix
       run: |
-        if [[ ${GITHUB_EVENT_NAME} == 'release' ]]; then
+        if [[ ${GITHUB_EVENT_NAME} == 'pull_request' ]]; then
             build_type=$(echo "['Full', 'Lite']")
         else
             build_type=$(echo "['Full']")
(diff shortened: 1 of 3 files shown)


mrclary:
@ccordoba12, so CI ran with the fix and I verified that the new artifact works (https://github.com/spyder-ide/spyder/actions/runs/2674779006).
However, I realized that the merge state had the wrong version, so I had to revert `__init__.py` with the correct version number. This will trigger all the test workflows. If any of them fail, I think we can safely ignore them, and just verify that the macOS installer builds are good.

ccordoba12:
> so CI ran with the fix and I verified that the new artifact works

Great! Thanks for your prompt help with that!

> However, I realized that the merge state had the wrong version, so I had to revert __init__.py with the correct version number.

I don't understand: so the artifacts in this PR have the right Spyder version or not?

mrclary:
> I don't understand: so the artifacts in this PR have the right Spyder version or not?

I'm checking right now. Stand by.

mrclary:
Confirmed. These artifacts have the correct 5.3.2 version number. I downloaded the full version and confirmed that everything is working.
https://github.com/spyder-ide/spyder/actions/runs/2677440607

ccordoba12:
Ok, good to know.

ccordoba12:
Thanks a lot @mrclary!

mrclary:
In response to #18776, I've reopened this PR.

I've found the issue: the QtWebEngineProcess.app within the Qt frameworks must have the `runtime` options flag for notarization, which breaks it.

I believe that the solution is to add the `com.apple.security.cs.disable-executable-page-protection` entitlement to its signature. See https://forum.qt.io/post/526704.

If the artifact is successful, I'll replace the existing release artifact as I did before.

Thanks all for your patience as I work out these code signing quirks 😊

mrclary:
As of this post, the 5.3.2 artifacts have been replaced with ones where QtWebEngine should be working properly.

ccordoba12:
> Thanks all for your patience as I work out these code signing quirks

Thanks for your lighting fast response to solve these issues!

PR #18768, opened 2022-07-21
PR: Prefer a run-slow skip reason over an already passed one (Testing)

juliangilbey:
## Description of Changes

Currently, on a subsequent test run, `conftest.py` will report the reason for skipping a "run-slow" test if `--run-slow` is not specified as "Test passed in previous runs", which is not true.  This patch changes the order of the conditionals so that the "Need/Don't need --run-slow option to run" reason will be given precedence.

### Affirmation

@juliangilbey


ccordoba12 (approved):
Looks good to me, thanks @juliangilbey!

PR #18808, opened 2022-07-24
PR: Use an instance of SimpleCodeEditor to print files (Editor)

### LINKED ISSUE #2521: Printing from the editor shows dark background or blank text
I've set the Editor syntax/color theme to Spyder/Dark which works fine.  The problem is that the dark background is now also being applied to the printed page (and print preview) which does not seem right and wastes ink.  I'm using Spyder 2.3.5.2 and do not recall having this problem before. Here is a screenshot of print-preview and that's how it came out on the printer.
image

ccordoba12:
## Description of Changes

- This allows us to always use a light syntax highlighting theme when printing.
- The `Print preview` looks like this now

    imagen

### Issue(s) Resolved

Fixes #2521.

### Affirmation

### DIFF
--- a/spyder/plugins/editor/plugin.py
+++ b/spyder/plugins/editor/plugin.py
@@ -2307,7 +2307,9 @@ class Editor(SpyderPluginWidget, SpyderConfigurationObserver):
     def _create_print_editor(self):
         """Create a SimpleCodeEditor instance to print file contents."""
         editor = SimpleCodeEditor(self)
-        editor.setup_editor(color_scheme="idle", highlight_current_line=False)
+        editor.setup_editor(
+            color_scheme="scintilla", highlight_current_line=False
+        )
         return editor
 
     @Slot()
--- a/spyder/plugins/editor/plugin.py
+++ b/spyder/plugins/editor/plugin.py
@@ -22,7 +22,9 @@ import time
 # Third party imports
 from qtpy.compat import from_qvariant, getopenfilenames, to_qvariant
 from qtpy.QtCore import QByteArray, Qt, Signal, Slot, QDir
-from qtpy.QtPrintSupport import QAbstractPrintDialog, QPrintDialog, QPrinter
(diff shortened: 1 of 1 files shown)


ccordoba12:
As you can see, I decided to use the IDLE theme for this because I think it's the crispest one among the light themes we have, but I'm not so sure about it.

@dalthviz, what do you think?

dalthviz:
Should there be an option/preference to enable choosing which light theme to use when printing? Besides that I'm not sure which could be the best light theme. Maybe we could check which light theme provides the most contrast or some other measure to make a decision.

ccordoba12:
> Should there be an option/preference to enable choosing which light theme to use when printing?

I thought about that, but I think we can leave it for another PR.

> Besides that I'm not sure which could be the best light theme. Maybe we could check which light theme provides the most contrast or some other measure to make a decision.

That's what I did and concluded that the best one is IDLE. Could you do the same and tell me what you think?

dalthviz (commented):
Thanks @ccordoba12 ! Checking locally on Windows seems like this is working as expected :+1:

Regarding the syntax theme, checked with this page (https://webaim.org/resources/contrastchecker/) the contrast between the themes colors (as foreground) and `#ffffff` (as background) and seems like `Scintilla` has better contrast mostly since `IDLE` has the `#ff7700` color which fails the check (even with bold letters).

Just in case this is how they look in comparison (`Scintilla` left, `IDLE` right):

imagen

So what do you think?

ccordoba12:
> Regarding the syntax theme, checked with this page (https://webaim.org/resources/contrastchecker/) the contrast between the themes colors (as foreground) and #ffffff (as background) and seems like Scintilla has better contrast mostly since IDLE has the #ff7700 color which fails the check (even with bold letters).

Great thinking! I'll change to use the `Scintilla` theme then.

dalthviz (approved):

PR #18831, opened 2022-07-26
PR: Don't apply color schemes that require a restart and fix other issues about them

ccordoba12:
## Description of Changes

- Don't apply new color schemes if they require a restart. Instead, simply show users the restart message.

    Right now users end up in this situation when changing to a light color scheme, which is confusing:

   imagen

    With this PR we only show the restart message, which is clearer according to this https://github.com/spyder-ide/spyder/issues/2521#issuecomment-1172457307

    imagen

- Simplify the restart message to not mention the options that require it (see the second screenshot above). I think that's an implementation detail that users don't need to know.
- Update occurrences, line and cell selections in the editor when applying a new color scheme. This allows color schemes to be applied correctly after switching them.

    **Before**

    imagen

    **After**

    imagen

- Add new API methods to disable/restore notifications for options.
- Make plugins react to UI theme option changes. This applies the color scheme correctly under some combinations of color scheme and UI theme option changes.
- Make strings related to UI theme options translatable.
- Fix a couple of small issues reported by our linter in the Pylint plugin (I noticed them when working on this PR).

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/editor.py
+++ b/spyder/plugins/editor/widgets/editor.py
@@ -1069,11 +1069,18 @@ class EditorStack(QWidget):
             for finfo in self.data:
                 finfo.editor.set_color_scheme(color_scheme)
 
-                # Update the omst important extra selections
+                # Update the most important extra selections so new color
+                # schemes appear to users as expected.
                 finfo.editor.unhighlight_current_line()
                 finfo.editor.unhighlight_current_cell()
-                finfo.editor.highlight_current_line()
-                finfo.editor.highlight_current_cell()
+                finfo.editor.clear_occurrences()
+
+                if self.highlight_current_line_enabled:
+                    finfo.editor.highlight_current_line()
+                if self.highlight_current_cell_enabled:
+                    finfo.editor.highlight_current_cell()
+                if self.occurrence_highlighting_enabled:
+                    finfo.editor.mark_occurrences()
 
     def set_wrap_enabled(self, state):
         # CONF.get(self.CONF_SECTION, 'wrap')
--- a/spyder/plugins/editor/widgets/editor.py
+++ b/spyder/plugins/editor/widgets/editor.py
@@ -1069,6 +1069,12 @@ class EditorStack(QWidget):
             for finfo in self.data:
                 finfo.editor.set_color_scheme(color_scheme)
 
+                # Update the omst important extra selections
+                finfo.editor.unhighlight_current_line()
+                finfo.editor.unhighlight_current_cell()
+                finfo.editor.highlight_current_line()
+                finfo.editor.highlight_current_cell()
+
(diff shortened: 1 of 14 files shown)


dalthviz (commented):
Thanks @ccordoba12 ! Checking locally seems like now you can't apply some of the possibilities mixing the light/dark themes with ligh/dark syntax themes. For example, selecting the Light theme when having previously Automatic and the Spyder Dark syntax theme causes no change or dialog to show up:

theme

Also, if you then change from Light to Automatic you get the restart dialog but the syntax theme gets applied too:

imagen

Other than that seems like the changes are working as expected :+1:

ccordoba12:
Thanks for the review @dalthviz! I think I fixed the error you reported above, so this should be ready for another review.

I also took the opportunity to fix other bugs about color schemes that were bothering me too. Please see the updated description above to learn about them.

spyder/plugins/editor/widgets/editor.py
```diff
@@ -1069,6 +1069,12 @@ def set_color_scheme(self, color_scheme):
             for finfo in self.data:
                 finfo.editor.set_color_scheme(color_scheme)
 
+                # Update the omst important extra selections
+                finfo.editor.unhighlight_current_line()
+                finfo.editor.unhighlight_current_cell()
+                finfo.editor.highlight_current_line()
+                finfo.editor.highlight_current_cell()
+
```
dalthviz, line 1077: Could here be also worty to update the highlighted occurences?

For example when going from a dark syntax to a ligh syntax you see this:

imagen

Instead of this:

imagen

dalthviz (commented):
Thanks @ccordoba12 ! I checked locally and seems like the previous behavior was fixed :+1:. I left some comments and a suggestion for a comment but other than that this LGTM :+1:

spyder/plugins/editor/widgets/editor.py
ccordoba12: Good catch! I'll implement it as part of this PR.

spyder/plugins/pylint/main_widget.py
```diff
@@ -343,9 +343,6 @@ def __init__(self, name=None, plugin=None, parent=None):
         self.treewidget.sig_edit_goto_requested.connect(
             self.sig_edit_goto_requested)
 
-    def on_close(self):
-        self.stop_code_analysis()
-
```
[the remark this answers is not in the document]
ccordoba12: There is another `on_close` method in this class that does exactly the same.

spyder/plugins/editor/widgets/editor.py
ccordoba12: This should be working as expected now @dalthviz.

dalthviz (approved):
:+1:

PR #18837, opened 2022-07-27
PR: Make the debugger work faster

impact27:
## Description of Changes

- Depends on spyder-ide/spyder-kernels#403

### Affirmation


impact27:
Looks like pdb is quite slow in Mac tests / Mac - Py3.9, conda, fast ?

impact27:
Maybe we should just skip `test_ipythonconsole.py` on mac? What do you think @ccordoba12?

ccordoba12:
> Maybe we should just skip test_ipythonconsole.py on mac? What do you think @ccordoba12?

All IPython console tests? That's seems a bit too much, but why do you think it's the best option?

dalthviz:
I think the approach of adding the `timeout=SHELL_TIMEOUT` worked for the `test_startup_code_pdb`, `test_recursive_pdb` and others. Maybe we should review the whole `test_ipythonconsole.py` and add the missing `timeout=SHELL_TIMEOUT` kwarg to all the `qtbot.waitSignal` calls at once?

If that is not enough then we could skip the tests that are still failing on MacOS

ccordoba12:
> I think the approach of adding the timeout=SHELL_TIMEOUT worked

I agree with @dalthviz. After reviewing the last run, it seems tests are failing because the debugger it's taking too much time to start on Mac 11. So adding `timeout=SHELL_TIMEOUT` should help to account for that.

impact27:
@ccordoba12 I am not sure why this was failing yesterday, but this should be delot with before 8/30/2022: 

> the image will be fully unsupported by 8/30/2022

impact27:
@ccordoba12 Should we use Mac 12 or Mac 11?

impact27:
@ccordoba12 the `test_completions_environment` seems to be very flaky (Only runs on windows), I think this is unrelated to this PR

impact27:
That seems to be addressed by the latest commit

spyder/plugins/ipythonconsole/widgets/debugging.py
```diff
@@ -440,11 +440,11 @@ def refresh_from_pdb(self, pdb_state):
...
                 self.sig_pdb_step.emit(fname, lineno)
-
-    def set_pdb_state(self, pdb_state):
-        """Set current pdb state."""
-        if pdb_state is not None and isinstance(pdb_state, dict):
-            self.refresh_from_pdb(pdb_state)
+        if 'stack' in pdb_state:
+            pdb_stack, pdb_index = pdb_state['stack']
+            self.sig_pdb_stack.emit(pdb_stack, pdb_index)
+        if 'request_pdb_input' in pdb_state:
```
ccordoba12, line 446:
```suggestion

        if 'stack' in pdb_state:
            pdb_stack, pdb_index = pdb_state['stack']
            self.sig_pdb_stack.emit(pdb_stack, pdb_index)

        if 'request_pdb_input' in pdb_state:
```

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -966,10 +974,20 @@ def _change_client_conf(self, client, client_conf_func, value):
...
-                ccf(value)
-                c.shellwidget.sig_pdb_prompt_ready.disconnect(change_conf)
-            sw.sig_pdb_prompt_ready.connect(change_conf)
+            if client_conf_func in [
+                    sw.set_spyder_breakpoints,
+                    sw.set_pdb_ignore_lib,
+                    sw.set_pdb_execute_events,
+                    sw.set_pdb_use_exclamation_mark,
+                    sw.set_pdb_stop_first_line,
+                    ]:
```
ccordoba12, line 983:
```suggestion
                    sw.set_pdb_stop_first_line]:
```

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -91,6 +92,9 @@ class ShellWidget(NamepaceBrowserWidget, HelpWidget, DebuggingWidget,
     sig_kernel_started = Signal()
     sig_kernel_reset = Signal()
 
+    # Request configuration from plugins
+    sig_configure_requested = Signal()
```
ccordoba12, line 96:
```suggestion
    # Request plugins to send additional configuration to the kernel
    sig_config_kernel_requested = Signal()
```

This follows my suggestion above.

ccordoba12 (commented):
Thanks @impact27 for this! It looks pretty good!

I only left so small suggestions for you, otherwise looks good to me.

ccordoba12 (approved):
Looks good to me now, thanks @impact27!

PR #18863, opened 2022-07-29
PR: Fix EOL for missed files

mrclary:
It seems that some files were missed in bdfe0b59 (#18691).

If I'm in error, please advise.


impact27:
I only addressed py files in the spyder folder. Somes files here should not be changed, for example: `external-deps/qdarkstyle/docs/make.bat`

ccordoba12:
I agree with @impact27: batch files should be left out of this. The rest looks good to me.

ccordoba12 (approved):
Thanks @mrclary!

impact27:
You should probably add the commit to [.git-blame-ignore-revs]. I wonder why the bat files are showing here, they should be excluded with .gitattributes file

ccordoba12:
@mrclary, please rebase to use the changes to our test suite done in PR #18802.

impact27:
Did you update the commit hash in .git-blame-ignore-revs after the rebase?

ccordoba12 (approved):
Thanks @mrclary!

ccordoba12:
@impact27, all good?

impact27:
Looks good to me!

PR #18871, opened 2022-07-30
PR: Make automatic completions more responsive (Editor)

### LINKED ISSUE #16342: Slow time to display completions
## Problem Description
The completion and snippets are much slower than in other IDEs. Something similar was described in #15042 which is now closed, but I don't feel comfortable with the actual completion in Spyder. This is not a new issue, but only now trying other IDEs I have noticed how slow is the completion in Spyder.

I know this GIFs are not a proof because I can type different anytime, but they meant to demonstrate the almost instantly completion and snippets options showing along with your typing in VSCode and how reliable they are. In Spyder they have a 0.5s to 1s delay to display. The fact is: **you have to stop your writing to wait for completion suggestions.** 

In both cases I was using Kite.
  
VSCode
vscode

Spyder. Sometimes I also need to navigate to second or third option, even if what I typed has no relation to the first option.
spyder

1. Write some code

I would expect to get a fast completion as in other IDEs like VSCode or PyCharm.

## Versions
* Spyder version: 5.1.2 None
* Python version: 3.9.5 64-bit
* Qt version: 5.12.10
* PyQt5 version: 5.12.3

# Dependencies

## Mandatory:
IPython >=7.6.0               :  7.27.0 (OK)
jedi >=0.17.2;<0.19.0         :  0.18.0 (OK)
keyring >=17.0.0              :  23.1.0 (OK)
nbconvert >=4.0               :  6.1.0 (OK)
paramiko >=2.4.0              :  2.7.2 (OK)
parso >=0.7.0;<0.9.0          :  0.8.2 (OK)
pygments >=2.0                :  2.10.0 (OK)
pylint >=2.5.0;<2.10.0        :  2.9.6 (OK)
pylsp >=1.2.2;<1.3.0          :  1.2.2 (OK)
pylsp_black >=1.0.0           :  None (OK)
qstylizer >=0.1.10            :  0.2.1 (OK)
qtawesome >=1.0.2             :  1.0.3 (OK)
qtconsole >=5.1.0             :  5.1.1 (OK)
qtpy >=1.5.0                  :  1.10.0 (OK)
setuptools >=49.6.0           :  52.0.0.post20210125 (OK)
sphinx >=0.6.6                :  4.1.2 (OK)
spyder_kernels >=2.1.1;<2.2.0 :  2.1.1 (OK)
watchdog >=0.10.3             :  2.1.5 (OK)
zmq >=17                      :  22.2.1 (OK)

### Optional:
cython >=0.21                 :  None (OK)
matplotlib >=2.0.0            :  3.4.3 (OK)
numpy >=1.7                   :  1.21.2 (OK)
pandas >=1.1.1                :  1.3.2 (OK)
scipy >=0.17.0                :  1.7.1 (OK)
sympy >=0.7.3                 :  None (OK)

ccordoba12:
## Description of Changes

- Send a completion request to the server before sending others,
- Reduce number of characters to get auto-completions to 1.
- Increase delay to send all scheduled requests to 300ms and then 50ms among them.

**Before**

slow

**After**

fast

### Issue(s) Resolved

Fixes #16342.

### Affirmation


spyder/config/main.py
```diff
@@ -222,8 +222,8 @@
               'tab_always_indent': False,
               'intelligent_backspace': True,
               'automatic_completions': True,
-              'automatic_completions_after_chars': 3,
-              'automatic_completions_after_ms': 300,
+              'automatic_completions_after_chars': 1,
+              'automatic_completions_after_ms': 50,
```
impact27, line 226: 50ms is very short and would likely be triggered at every keystroke. I think 300ms was chosen because that gives time to the user to type another key, avoiding too many requests?

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -611,14 +615,32 @@ def __init__(self, parent=None):
...
+        # Completion types
+        doc_completion = CompletionRequestTypes.DOCUMENT_COMPLETION
+        completion_resolve = CompletionRequestTypes.COMPLETION_RESOLVE
+
+        # Send completion requests first to the server to get faster
+        # completions
+        completion_requests = [
+            req for req in self._pending_server_requests
+            if (req[0] == doc_completion or req[0] == completion_resolve)
+        ]
```
impact27, line 634: Should there be two `_pending_server_requests` lists? One with higher priority? I am not sure which approach would be nicer.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -1204,9 +1226,24 @@ def process_symbols(self, params):
...
+
+        # If auto-completions are on, start the timer to request completions
+        # here. That way we can ensure an almost instant response when the user
+        # stops typing.
+        # Note: We only ask for auto-completions if there was some text
+        # generated as part of the keyPressEvent.
+        # Fixes spyder-ide/spyder#11021
+        if (
+            self.automatic_completions
+            and not self._server_requests_timer.isActive()
```
impact27, line 1241: Why are you checking `self._server_requests_timer.isActive()`?

impact27 (commented):
Looks good to me! I left a few points for clarification.

spyder/plugins/editor/widgets/codeeditor.py
ccordoba12: > Should there be two _pending_server_requests lists?

Perhaps, but I don't know how to handle that. And it'd probably kill the simplicity of your `schedule_request` decorator.

Since we can sort the higher/lower lists here, I think the gain would be small as well.

spyder/plugins/editor/widgets/codeeditor.py
ccordoba12: The idea is to use this timer to limit the amount of completion requests send to the server. So, we send a request when the starts to type and prevent to do it while it's still doing it.

I think it has the role you thought `automatic_completions_after_ms` had.

spyder/plugins/editor/widgets/codeeditor.py
impact27: Agreed

spyder/plugins/editor/widgets/codeeditor.py
impact27: I am not sure about that. If `_server_requests_timer` is active, the automatic completion would never trigger?
I would just drop this test. If `automatic_completions_after_ms` is long enough that the user has time to type a new key, `start()` will interrupt the timer and restart it, preventing the previous completion request from being sent.

impact27:
Ok so I had a look and think we should do the following:

There are two decorators for requests:
`@schedule_request(method=CompletionRequestTypes.DOCUMENT_COMPLETION)`
`@request(method=CompletionRequestTypes.DOCUMENT_COMPLETION)`

The first one schedules a request to be made later, and the second one sends it right away.

You do not want to send a request while typing in general as it slows down the editor. But there could be some exception. For example a complete request.

I therefore propose to replace `@schedule_request` by `@request` for `DOCUMENT_COMPLETION` (and also for `COMPLETION_RESOLVE`?)

This means that do_completion must update the document if needed by calling `document_did_change`

We should probably add `self._document_server_needs_update = False` inside `document_did_change` so if a completion is made, it cancels pending document update requests.

impact27:
So to summarize, there are two subjects:

1 - LSP requests: Most LSP requests are not time sensitive (It doesn't matter if the errors take a little while to show) so it is fine to put them on a short timer (`_server_requests_timer`). As you note in this PR, completion requests are an exception, and should be handled as soon as possible. 

2 - Autocompletion: A request should be sent when the user finished typing. My opinion is that 50ms is too short as it would trigger after every key, but I am open for debate. `_timer_autocomplete` is triggered with a reasonable inter-key time (300ms?) After that, the request should be sent immediately (and not after a further delay as was the case before this PR)

ccordoba12:
> I therefore propose to replace @schedule_request by @request for DOCUMENT_COMPLETION (and also for COMPLETION_RESOLVE?)

That's a good idea and would avoid the higher/lower priority lists approach. However, auto-completions would need to be triggered by a timer on typing, right? (I think both of us agree on that).

> This means that do_completion must update the document if needed by calling document_did_change

Correct. The problem is `document_did_change` is very slow right now because we're sending the whole document instead of the diff over its previous contents. So perhaps that should need to be solved first (I have an idea on how to do it).

Surprisingly, auto-completions are working just fine in this PR without doing a `document_did_change` first (no idea why).

> Autocompletion: A request should be sent when the user finished typing.

That's more or less how auto-completions are working right now. And that's what gives a huge delay to show completions (as shown in my gif above), which breaks the fluidity of writing code in our editor (see issue #16342 for more details). I thought it was a Jedi's problem but after a lot of debugging I realized it was ours (Jedi is really fast!).

Precisely to improve the current situation I proposed the changes in this PR, which consist in sending a preemptive completion request (delayed by `_timer_autocomplete`) as soon as the user *starts* typing (instead of waiting until they stop typing). And that makes the UX way better, in my opinion.

> My opinion is that 50ms is too short as it would trigger after every key, but I am open for debate. _timer_autocomplete is triggered with a reasonable inter-key time (300ms?) After that, the request should be sent immediately (and not after a further delay as was the case before this PR)

Ok, about this, I think you haven't understood my approach yet. I'm using `_timer_autocomplete` to delay sending a completion request a bit and nothing else. As I said before, I think we could even remove it because it makes little sense to allow users to control this. That way a completion request would really be sent as soon as the user starts typing (no delay at all). In addition, we could control that no other completion requests are sent while the user is still typing by checking the status of `_server_requests_timer` (as I'm doing now) because it's associated to the `textChanged` signal.

Another option would be to retain `_timer_autocomplete` and start it on every `textChanged` too, as we're doing with `_server_requests_timer`. But then we'd have two timers that do almost the same (one for completions and the other for scheduled requests), which I find a bit unnecessary. However, if you think that would make the code clearer, we could go with that approach instead.

dalthviz:
Could be worthy to do some benchmarking to check how much the completions improve with the changes proposed here?

Also, for a more in deep discussion about the points @impact27 raised and also for @ccordoba12 to better describe his ideas here maybe could be worthy to talk about this PR on Thursday?

impact27:
I think the main question is: would you rather send many unnecessary requests or wait an extra 300ms to send the requests.

> That's a good idea and would avoid the higher/lower priority lists approach. However, auto-completions would need to be triggered by a timer on typing, right? (I think both of us agree on that).

Correct, we agree, but the current behaviour on 5.x is to use a timer to start a timer to send the request, which is not ideal and probably the reason for the large delay. One timer level should be removed, which you do in this PR (not really but by calling `_process_server_requests` in `_handle_automatic_completions` the effect is the same). 

>Correct. The problem is document_did_change is very slow right now because we're sending the whole document instead of the diff over its previous contents. So perhaps that should need to be solved first (I have an idea on how to do it).

I agree that only the diff should be sent, but that could be tackled in another PR.

>Surprisingly, auto-completions are working just fine in this PR without doing a document_did_change first (no idea why).

My understanding of this PR is that in `_handle_automatic_completions` you call `_process_server_requests` which is indeed doing a `document_did_change` first.

> Precisely to improve the current situation I proposed the changes in this PR, which consist in sending a preemptive completion request (delayed by _timer_autocomplete) as soon as the user starts typing (instead of waiting until they stop typing). And that makes the UX way better, in my opinion.

Does the difference in delay of 250ms explains the long waiting time you see in the gif? I think the other improvements in this PR have a larger effect. The question is: 

Do you see, using all the other improvements in this PR, a big difference between `automatic_completions_after_ms=50` and `automatic_completions_after_ms=300`?

> Ok, about this, I think you haven't understood my approach yet. I'm using _timer_autocomplete to delay sending a completion request a bit and nothing else. As I said before, I think we could even remove it because it makes little sense to allow users to control this. That way a completion request would really be sent as soon as the user starts typing (no delay at all).

I would not call `_handle_automatic_completions` synchronously at every keystroke. It calls `document_did_change` which would slow down typing in spyder significantly. Also the function doesn't look simple (=fast?) The effect of this would need to be checked. 

> In addition, we could control that no other completion requests are sent while the user is still typing by checking the status of _server_requests_timer (as I'm doing now) because that timer is associated to the textChanged signal.

I am not sure about that, because `_server_requests_timer` could be running for another, unrelated, request, so you would be dropping autocomplete requests. It can also be started from `schedule_request`.

> Another option would be to retain _timer_autocomplete and start it on every textChanged too, as we're doing with _server_requests_timer. But then we'd have two timers that do almost the same (one for completions and the other for scheduled requests), which I find a bit unnecessary. However, if you think that would make the code clearer, we could go with that approach instead.

I agree that the two timers have a similar use, but not completely. Importantly, `_server_requests_timer` is started from `schedule_request` and not only from `_schedule_document_did_change`. But I agree that the `document_did_change` and autocomplete should be processed together. 

So my proposition is:

Do not change from 5.x:
------------------------

1 - `_server_requests_timer` should be started from `schedule_request` with a small delay(50ms) (LSP_REQUEST_DELAY). When processing, it calls `document_did_change` if needed `(self._document_server_needs_update == True`)
2 - At every `textChanged`, `self._document_server_needs_update` is set to `True`. 

Change from 5.x:
-----------------

1 - In `_process_server_requests`, if `self._document_server_needs_update == True`, send an autocomplete request after `document_did_change` (synchronous request).
2 - in `_schedule_document_did_change`, start `_server_requests_timer`, but with a longer timeout (300ms) than from `schedule_request` so this call could be prolonged.


The advantages are: 
 - Only one timer (get rid of `_timer_autocomplete`)
 - `_server_requests_timer` has two timeouts depending on the source
 - no separation of `_pending_server_requests` list (because the autocomplete requests are never put in the waiting list)
 - Might require two `do_completion`? But we might have completion requests be synchronous anyway.

> Could be worthy to do some benchmarking to check how much the completions improve with the changes proposed here?

Good idea

> Also, for a more in deep discussion about the points @impact27 raised and also for @ccordoba12 to better describe his ideas here maybe could be worthy to talk about this PR on Thursday?

I can't join on Thursday unfortunately

ccordoba12:
@impact27, thanks a lot for your suggestions in https://github.com/spyder-ide/spyder/pull/18871#issuecomment-1201759872. They helped me to simplify and make my initial implementation work much better. And they also allowed me to remove the `automatic_completions_after_ms` option, which is unnecessary now.

Could you take a look at this PR again and let me know what you think about it?

impact27:
Looks good to me now. Do you see that the delay issue has been resolved?

ccordoba12:
> Looks good to me now.

Great!

> Do you see that the delay issue has been resolved?

Yeah, it does. I've been testing this PR during the week and it works really well, i.e. completions feel much responsive now.

ccordoba12:
> Could be worthy to do some benchmarking to check how much the completions improve with the changes proposed here?

Good suggestion @dalthviz. I recorded the time since a completion request is sent and its results are about to be displayed in the editor, using the same example as in my initial post:

| Method | Time |
| -- | -- |
| This PR | ~0.3 s |
| 5.x with default options | ~1s |
| 5.x with the options shown below  | ~1.5s |

imagen

Interestingly, trying to get faster completions is actually giving users slower results.

dalthviz (approved):
Thanks @ccordoba12 ! LGTM :+1:

ccordoba12:
Ok, since @impact27 already said he's ok with this one, I'm going to merge it.

PR #18910, opened 2022-08-02
PR: Fix call to show compatibility message for plugins (Mainwindow)

### LINKED ISSUE #18901: Windows standalone version crashes on launch with an AttributeError
First of all, I can't try everything I would like to troubleshoot the issue because I don't have permissions to manage the virtual machine I'm using. I had to request the installation of spyder, which the responsible party did, but without adding shortcuts in the start menu or on the desktop. I found two ways of launching spyder: right click a file and open with spyder, or through cmd ```C:\Program Files\Spyder>Python\python Spyder.launch.pyw``` either way the logs show the exact same output.

## Problem Description

With a fresh install of the standalone version on windows, spyder crashes on launch after a brief splash screen. 
I tried restarting it and clearing AppData/Local/Spyder, which is about all I can try with the permissions I have.
I found someone with the same unanswered issue on [stackoverflow](https://stackoverflow.com/questions/73057717/spyder-dont-launch). I can't try the suggested solution, which didn't work for them anyway, because I don't have pip on this machine.

### Steps to reproduce

1. Install the latest windows standalone version of spyder
2. Right click any file
3. "Edit with Spyder"

### Expected and actual output
Expected a GUI window after the splash screen. Getting nothing instead, the splash screen just closes.

### Full logs

```logs/cmd output
fromIccProfile: failed minimal tag size sanity
  File "C:\Program Files\Spyder\pkgs\spyder\app\mainwindow.py", line 753, in <lambda>
    lambda plugin_name, omit_conf: self.register_plugin(
  File "C:\Program Files\Spyder\pkgs\spyder\app\mainwindow.py", line 213, in register_plugin
    self.show_compatibility_message(message)
  File "C:\Program Files\Spyder\pkgs\spyder\app\mainwindow.py", line 1069, in __getattr__
    return super().__getattr__(attr)
AttributeError: 'MainWindow' object has no attribute 'show_compatibility_message'
```

## Versions
Spyder won't launch and I'm not using the conda version, so here's what I know from the included package files.

* Spyder version: latest standalone windows version
* Python version: 3.8.10
* Qt version: ?
* PyQt version: 5.15.7
* Operating System name/version: Windows Server 2016 Datacenter 64 bits (10.0, version 14393)

dalthviz:
## Description of Changes

If a plugin implements a compatibility check and it fails we were calling `show_compatibility_message` instead of `show_plugin_compatibility_message` causing an `AttributeError`

### Issue(s) Resolved

Fixes #18901

### Affirmation

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -210,7 +210,7 @@ class MainWindow(QMainWindow):
         plugin.get_description()
 
         if not is_compatible:
-            self.show_compatibility_message(message)
+            self.show_plugin_compatibility_message(message)
             return
 
         # Connect Plugin Signals to main window methods


ccordoba12 (approved):
Thanks @dalthviz!

PR #18911, opened 2022-08-03
PR: Add methods to simplify access to our config system in `MainWindow`

ccordoba12:
## Description of Changes

- Small refactoring whose main purpose is to add `get_conf`/`set_conf` methods to MainWindow in order to simplify how we access config options on that class.
- Move `__init__` to be at the beginning of MainWindow (I think it's odd to have it in the middle of the class).
- Move the Qt methods we override on MainWindow to be on their own section.
- Add more section separators to MainWindow to make easier to navigate it.
- Remove old Mac stylesheet because it's no longer needed.

*Notes*:
1. I did these changes while working to fix another bug. I preferred to break that into two parts to facilitate reviewing it.
3. I think we should avoid using `CONF` because it's cleaner and easier to read to have methods that call it behind the scenes.

### Affirmation


ccordoba12:
Let's hold this PR until PR #18778 is merged because it touches the same code.

ccordoba12:
@dalthviz, scratch my last comment because Ryan agreed to leave PR #18778 for Spyder 6.

spyder/app/mainwindow.py
```diff
@@ -738,7 +743,7 @@ def _update_shortcuts_in_panes_menu(self, show=True):
                     try:
                         context = '_'
                         name = 'switch to {}'.format(section)
-                        shortcut = CONF.get_shortcut(
+                        shortcut = self._conf.get_shortcut(
```
dalthviz, line 746: Maybe a new method call `get_conf_shortcut` should be implemented which returns `self._conf.get_shorcut` in a similar way as done for `set_conf` and `get_conf`?

dalthviz (commented):
Thanks @ccordoba12 ! Left a comment regarding the creation of a method to access shortcuts in a similar manner as to get and set configurations. Other than that this LGTM :+1:

dalthviz (approved):
:+1:

PR #18928, opened 2022-08-03
PR: Improve `Maintain focus in the editor` option and unmaximize plugins when running/debugging code

### LINKED ISSUE #3221: Editor loses cursor focus after running cell when IPython console is detached

Normally, when all the Spyder panes are in a single window, running a cell (ctrl-enter) produces the desired output in the IPython console and leaves the cursor in the editor window.  This way, when you run a cell, you can immediately continue working in the editor window.

When I detach the editor window -- so that it floats in its own window -- and then run a cell, this behavior changes.  When detached, after you hit ctrl-enter in the editor pane, it dumps the cursor into the ipython console panel.  So after you hit ctrl-enter and start typing, the text appears directly in the console.  This is cumbersome because often after you run a cell, you'd like to modify the cell slightly, and you are stuck using the mouse to bring the cursor back to the detached editor window manually.

To make this clear graphically, these two setups produce different behavior for the cursor focus:

2016-06-11 14_39_55-spyder python 2 7
2016-06-11 14_40_18-spyder - editor
- Spyder Version:  2.3.8
- Python Version:  2.7
- Operating system: Windows 7
## Dependencies

I have not intentionally modified these after a fresh install

IPython >=1.0      :  4.1.2 (OK)
pandas >=0.13.1    :  0.18.0 (OK)
pyflakes >=0.5.0   :  1.1.0 (OK)
pygments >=1.6     :  2.1.1 (OK)
pylint >=0.25      :  None (NOK)
qtconsole >=4.0    :  4.2.0 (OK)
rope >=0.9.2       :  0.9.4 (OK)
sphinx >=0.6.6     :  1.3.5 (OK)
sympy >=0.7.3      :  1.0 (OK)
zmq >=2.1.11       :  15.2.0 (OK)

ccordoba12:
## Description of Changes

- Avoid giving focus to the IPython console when it's undocked, the user runs/debugs code and `Maintain focus to editor` is enabled.
- Keep focus in the editor in any run or debug operation and that option is enabled.
- Unmaximize any plugin before running or debugging code.
- Correctly set the state of the `Maximize` button when unmaximizing a plugin programmatically from other plugins (this was totally broken).
- Add `sig_unmaximize_plugin_requested` to `SpyderPluginV2` so that plugins can request to unmaximize the currently maximized plugin from others programmatically.
- Add `unmaximize_dockwidget` and `unmaximize_other_dockwidget` methods to the Layout plugin. Also move `switch_to_plugin` to it.
- Expand tests to catch possible regressions in the future.

### Issue(s) Resolved

Fixes #3221

### Affirmation

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -385,15 +385,28 @@ class MainWindow(QMainWindow):
         last_plugin = self.last_plugin
         try:
             # New API
-            if (last_plugin is not None
-                    and last_plugin.get_widget().is_maximized
-                    and last_plugin is not plugin):
-                self.layouts.maximize_dockwidget()
+            if (
+                last_plugin is not None
+                and last_plugin.get_widget().is_maximized
+                and last_plugin is not plugin
+            ):
+                maximize_action = self.layouts.maximize_action
+                if maximize_action.isChecked():
+                    maximize_action.setChecked(False)
+                else:
+                    maximize_action.setChecked(True)
         except AttributeError:
             # Old API
-            if (last_plugin is not None and self.last_plugin._ismaximized
-                    and last_plugin is not plugin):
-                self.layouts.maximize_dockwidget()
+            if (
+                last_plugin is not None
+                and self.last_plugin._ismaximized
+                and last_plugin is not plugin
+            ):
+                maximize_action = self.layouts.maximize_action
+                if maximize_action.isChecked():
+                    maximize_action.setChecked(False)
+                else:
+                    maximize_action.setChecked(True)
 
         try:
             # New API
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -213,13 +213,16 @@ class MainWindow(QMainWindow):
             self.show_plugin_compatibility_message(message)
             return
 
-        # Connect Plugin Signals to main window methods
+        # Connect plugin signals to main window methods
         plugin.sig_exception_occurred.connect(self.handle_exception)
         plugin.sig_free_memory_requested.connect(self.free_memory)
         plugin.sig_quit_requested.connect(self.close)
         plugin.sig_redirect_stdio_requested.connect(
             self.redirect_internalshell_stdio)
         plugin.sig_status_message_requested.connect(self.show_status_message)
+        plugin.sig_unmaximize_plugin_requested.connect(self.unmaximize_plugin)
+        plugin.sig_unmaximize_plugin_requested[object].connect(
+            self.unmaximize_plugin)
 
         if isinstance(plugin, SpyderDockablePlugin):
             plugin.sig_focus_changed.connect(self.plugin_focus_changed)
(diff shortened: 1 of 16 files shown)


spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -620,7 +628,8 @@ def close_client(self, index=None, client=None, ask_recursive=True):
 
     # ---- For execution and debugging
     def run_script(self, filename, wdir, args, debug, post_mortem,
-                   current_client, clear_variables, console_namespace):
+                   current_client, clear_variables, console_namespace,
+                   focus_to_editor):
```
dalthviz, line 632: Should this new `focus_to_editor` param be a kwarg with a default value? Seems like this is breaking the running file from the project explorer

ccordoba12: Good point. I think we should make most of these paramenters kwargs so that it's easier to call `run_script` from other plugins. I'll do that as part of this PR.

spyder/plugins/projects/plugin.py
```diff
@@ -392,10 +392,26 @@ def on_close(self, cancelable=False):
...
+                except AttributeError:
+                    # Old API
+                    is_maximized = last_plugin._ismaximized
+
+            if (
+                last_plugin is not None
+                and is_maximized
+                and last_plugin is not self
+            ):
+                layouts.unmaximize_dockwidget()
```
dalthviz, line 414: Should this logic be in the `Layout` plugin? Maybe a method with a signature like `unmaximize_other_dockwidget(self, plugin_instance)` or maybe could e worthy to modify the current `unmaximize_dockwidget` method to recieve a kwarg, something like `check_plugin_instance=None`. With that here the kwarg could be used to pass `self` and then this will become `layouts.unmaximize_dockwidget(check_plugin_instance=self)`

spyder/app/mainwindow.py
```diff
@@ -385,15 +370,28 @@ def switch_to_plugin(self, plugin, force_focus=None):
...
+            if (
+                last_plugin is not None
+                and self.last_plugin._ismaximized
+                and last_plugin is not plugin
+            ):
+                maximize_action = self.layouts.maximize_action
+                if maximize_action.isChecked():
+                    maximize_action.setChecked(False)
+                else:
+                    maximize_action.setChecked(True)
```
dalthviz, line 394: Should all this logic be moved to a new method in the `Layout` plugin? A method with a signature like `switch_to_plugin(self, last_plugin, plugin)` ? If that is done, maybe the method could be reused for the `_switch_to_plugin` method at `spyder.plugins.base`?

dalthviz (commented):
Thanks @ccordoba12 ! Checking locally seems like the changes are working as expected and I left some questions regarding code placement.

Otherwise this LGTM :+1:

spyder/plugins/projects/plugin.py
ccordoba12: Good idea. I'll go with your first suggestion because I find it easier to understand.

spyder/app/mainwindow.py
ccordoba12: I this it's a good idea to move this method to the Layout plugin because these checks use actions and methods present on it.

However, I don't agree with using it in `spyder.plugins.base`. I mean, I think there's little gain in improving that code given that it's going to be removed soon, once the editor is migrated to the new API.

dalthviz:
Checking locally this again seems like closing the Tour leaves the Maximize toolbar button in a incoherent state (it remains checked although the unmaximized action was done). Also, after closing the tour, should the plugins previously maximized be again maximized @ccordoba12 ?

ccordoba12:
> Checking locally this again seems like closing the Tour leaves the Maximize toolbar button in a incoherent state (it remains checked although the unmaximized action was done)

Thanks @dalthviz for manually checking that! I only connected the new `sig_unmaximize_plugin_requested` signal for dockable plugins, but the `Tours` plugin is not dockable. That's why it was not having effect.

I fixed that in my last commit and force pushed it again.

> Also, after closing the tour, should the plugins previously maximized be again maximized @ccordoba12 ? Maybe this was an issue before?

Yep, that was the case before. Since that's a minor thing, I think we should leave it for later.

dalthviz (approved):
👍🏼

PR #18962, opened 2022-08-06
PR: Restore plugins that were visible during the previous session

### LINKED ISSUE #12104: Save list of visible panes before closing Spyder and restore them on the next session
### Problem Description
* When you open Spyder, IPython console, Editor and the Help pane are shown by default.
Anmerkung 2020-04-02 103321

* **--> however, I'd like to have another pane to be selected at startup**, e.g. "_Variable Explorer_" (or whatever I had selected when I closed Spyder the last time)

* To my knowledge (info from @ccordoba12), at the moment there is no way to change this behavior e.g. in the settings. So this is basically a suggestion for a new feature.

* open Spyder

(not applicable)

(not applicable)

## Versions
* Spyder version: 4.1.1
* Python version: 3.8.2
* Qt version: 5.12.3
* PyQt version: 1.9.0
* Operating System name/version: Windows 10 (1909)

### Dependencies
(not applicable)

ccordoba12:
## Description of Changes

- This adds two new methods to the Layouts plugin to save and restore plugins that were visible during the previous session.
- They are run when the window is closed and on `post_visible_setup`, respectively.
- Use `SpyderConfigurationAccessor` to access config options on MainWindow. I realized that we could this after merging PR #18911.

### Issue(s) Resolved

Fixes #12104

### Affirmation


ccordoba12:
@CAM-Gerlach, this should fix the error you told me about Spyder-terminal being shown by default on startup.

CAM-Gerlach:
Thanks! I can test this with the installer built on this PR when I get back to the lab Monday.

dalthviz:
Just checked locally using `bootstrap.py` and this is what I got (the Terminal shows first until it loads and then the tab changes to the IPython Console):

test

Is that the desire behavior? Maybe something needs to be done over the spyder-terminal side to prevent it getting focus when starting?

I guess something similar is happening with the IPython Console and the Help pane. They are plugins that get focused when loading and then end up 'ovewritting' the previous session selected/focused pane in the respective layout region since the previous window state is loaded before all the plugins UI finishes setting up.

If what is wanted is to ensure that the last window state is applied and it does not get overwritten by plugins UI loading then a call to `layouts.setup_layout(default=False)`  on the `layouts.on_mainwindow_visible` could be enought?

Also, maybe there is a way for plugins like the IPython Console, Help, Terminal and probably QWebEngineWidget based plugin to not steal focus when loading?

ccordoba12:
> Is that the desire behavior?

Correct, that's how I implemented this because I'm calling `restore_visible_plugin` on `post_visible_setup` after all plugins have run their `on_mainwindow_visible` methods.

However, I could try to run it before that or in `pre_visible_setup` to see if that works as you expected. I thought it was not a big deal, but after seeing your gif I agree with you.

> Maybe something needs to be done over the spyder-terminal side to prevent it getting focus when starting?

I think we should try to solve it here because the behavior you showed can happen with other plugins as well, and we can't fix things in all of them.

> If what is wanted is to ensure that the last window state is applied and it does not get overwritten by plugins UI loading then a call to layouts.setup_layout(default=False) on the layouts.on_mainwindow_visible could be enought?

I think the problem with this approach is installing new plugins (e.g. Spyder-notebook) after Spyder was run for the first time. In that case, if we restore the layout from the previous session, those plugins wouldn't be shown in the interface. And that would be very confusing for users, perhaps leading them to conclude that the plugin was not installed correctly.

> Also, maybe there is a way for plugins like the IPython Console, Help, Terminal and probably QWebEngineWidget based plugin to not steal focus when loading?

I think this is more related to visibility than focus (because at the end only a single plugin can have focus at any time). And since Spyder-terminal is loaded at last (like all external plugins) then it's the last one to be visible. That's also why the Variable Explorer is the one that right now is visible after Spyder starts up (because we load plugins in alphabetical order).

ccordoba12:
@dalthviz, my last commit should fix the problem you reported above. So please test again.

CAM-Gerlach:
Sorry, I was too busy today with work, the Python 3.11 release, the Spyder governance repo and reviewing the new Editor PR and other things to test this on one of our machines.

> If what is wanted is to ensure that the last window state is applied and it does not get overwritten by plugins UI loading then a call to `layouts.setup_layout(default=False)` on the `layouts.on_mainwindow_visible` could be enought?

It has been a while, but I do remember in Spyder 4 a year or two ago we had a ton of problems getting the layout to reset correctly across platforms when we redid the whole introduction tour..,but hopefully things are better now, or at least this doesn't trigger one of the behaviors we found.

> I think the problem with this approach is installing new plugins (e.g. Spyder-notebook) after Spyder was run for the first time. In that case, if we restore the layout from the previous session, those plugins wouldn't be shown in the interface. And that would be very confusing for users, perhaps leading them to conclude that the plugin was not installed correctly.

We do talk about this and how to resolve it in the docs...somewhere, I'm not sure where. Maybe naive, but you could just store in the transient config what plugins were detected on the last run and if there any new ones, show them by default?

ccordoba12:
> We do talk about this and how to resolve it in the docs...somewhere, I'm not sure where. Maybe naive, but you could just store in the transient config what plugins were detected on the last run and if there any new ones, show them by default?

Right, that's another possibility to fix this. The problem is it wouldn't work for 5.3.3 (for instance) because we're not currently saving the list of plugins that were visible during the previous session. So, if before running 5.3.3 you install another third-party plugin, it wouldn't be shown in the interface once Spyder is started.

In other words, we'd need to add a way to save the list of visible plugins in 5.3.3 and leave this for 5.3.4. But I think this problem is significant enough that we should try to solve it asap.

dalthviz (approved):
Thanks @ccordoba12 ! This LGTM :+1:

spyder/app/mainwindow.py
```diff
@@ -1141,8 +1141,18 @@ def post_visible_setup(self):
...
+        # Populate `Panes > View` menu.
+        # This **MUST** be done before restoring the last visible plugins, so
+        # that works as expected.
+        self.layouts.create_plugins_menu()
+
+        # Restore last visible plugins.
+        # This **MUST** be done before running on_mainwindow_visible for
+        # plugins so that the user doesn't experience sudden jumps in the
+        # interface.
+        self.layouts.restore_visible_plugins()
```
ccordoba12, line 1153: Perhaps we could move this to `layouts.on_mainwindow_visible` in order to decrease the possibility of it being changed in the future. What do you think @dalthviz?

ccordoba12: And then here we'd only call `self.layouts.on_mainwindow_visible` before calling `on_mainwindow_visible` for the rest of the plugins.

dalthviz: Sure :+1:

spyder/app/mainwindow.py
```diff
@@ -1141,30 +1141,24 @@ def post_visible_setup(self):
...
-        # This **MUST** be done before restoring the last visible plugins, so
-        # that works as expected.
-        self.layouts.create_plugins_menu()
-
-        # Restore last visible plugins.
-        # This **MUST** be done before running on_mainwindow_visible for
-        # plugins so that the user doesn't experience sudden jumps in the
-        # interface.
-        self.layouts.restore_visible_plugins()
+        # This must be run before the window is shown.
```
dalthviz, line 1144: Maybe a reference to the original issue could be worthy here @ccordoba12 ?

CAM-Gerlach:
> In other words, we'd need to add a way to save the list of visible plugins in 5.3.3 and leave this for 5.3.4. But I think this problem is significant enough that we should try to solve it asap.

And not just that, many (most?) users do not update to each micro version in turn, so there would always be a long tail of users who that wouldn't address. On the other hand, its a one-time thing and requires a fairly specific set of circumstances to trigger, and if the user doesn't know how to open non-default panes (via `View` -> `Panes`, etc) then they are already missing out on a substantial amount of built-in Spyder functionality. This will hopefully improve if we implement the new window layout we talked about, though that would presumably only be for Spyder 6.

In any case, as you're far better informed on the implementation details and design constraints of this that I (I only had time to skim the discussion, to be frank), so take all that with a barrel of salt.

spyder/app/mainwindow.py
ccordoba12: Good idea. I amended my last commit with it.

ccordoba12:
> Currently the first_time config is used in the tabify_plugin method on the mainwindow and after detection is set to False but probably this set to False could be moved to the restore_visible_plugins method and then that config could be used to ensure that new plugins are visible regardless of what plugin was visible last session.

@dalthviz, I was thinking along similar lines to what you proposed here, but the problem with new plugins is a bit orthogonal to this. I mean, I've been thinking about it too after testing Spyder-notebook with 5.3.2 and it requires a couple of other things to be fixed.

So I'd like to get this PR merged first and then I could work on the new plugins problem. That's something I plan to do for 5.3.3 because we need it before releasing a new Spyder-notebook version.

dalthviz (commented):
Seems like a PEP8 issue can be fixed at https://github.com/spyder-ide/spyder/blob/9654e8554e5aa09365749b3c5b25238a4829c4d8/spyder/app/mainwindow.py#L125 @ccordoba12 

After that gets fix I will go ahead and merge this :+1:

ccordoba12:
Done.

PR #19008, opened 2022-08-10
PR: Add the ability to request unmaximizing plugins from widgets

ccordoba12:
## Description of Changes

- This was an oversight in PR #18928.
- I need it to fix our tests in master.
- Use this addition in the Tours plugin because it's more natural.
- Connect `PluginMainWidget` default signals to its plugin (we were only doing that for `PluginMainContainer`). Some things were broken due to that and others are now redundant.

### Affirmation

### DIFF
--- a/spyder/api/plugins/new_api.py
+++ b/spyder/api/plugins/new_api.py
@@ -321,21 +321,18 @@ class SpyderPluginV2(QObject, SpyderActionMixin, SpyderConfigurationObserver,
                 container.setup()
                 container.update_actions()
 
-            if isinstance(container, PluginMainContainer):
-                # TODO: This signals should also affect the main_widgets?
-                # Currently this is not working for
-                # instances of PluginMainWidget subclasses
-                # Default signals to connect in main container or main widget.
-                container.sig_exception_occurred.connect(
-                    self.sig_exception_occurred)
-                container.sig_free_memory_requested.connect(
-                    self.sig_free_memory_requested)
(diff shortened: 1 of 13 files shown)


spyder/api/plugins/new_api.py
```diff
@@ -321,21 +321,20 @@ def __init__(self, parent, configuration=None):
...
+            container.sig_redirect_stdio_requested.connect(
+                self.sig_redirect_stdio_requested)
+            container.sig_exception_occurred.connect(
+                self.sig_exception_occurred)
+            container.sig_unmaximize_plugin_requested.connect(
+                self.sig_unmaximize_plugin_requested)
+
+            # FIXME: This is semi-broken
+            # container.sig_restart_requested.connect(
+            #     self.sig_restart_requested)
```
dalthviz, line 337: Should an issue be created for this?

dalthviz (approved):
Thanks @ccordoba12 ! I left a question about the `FIXME` comment (to open a new issue). Regarding the changes to other signals (like `sig_free_memory_requested` or `sig_quit_requested`) took me a moment to understand what was happening so could you update the PR description to mention them explicitly (in case we need to look back in this PR in the future)?

Other than that this LGTM :+1:

spyder/api/plugins/new_api.py
ccordoba12: Don't worry, it's in todo list for 5.3.3. I'll work on that after fixing the tests in master.

PR #19068, opened 2022-08-15
PR: Fix syncing IPython console cwd with the Working directory toolbar

### LINKED ISSUE #19003: IPython ignore "Open IPython console here"
## Problem Description

When right-clicking a project folder and clicking "Open IPython console here", an IPython console is created.
But then when typing `pwd`, I always get the home/user directory instead of the current project directory.

1. Right-click a project folder and click "Open IPython console here"
2. Type `pwd` in the IPython console

I see the home/user directory instead of the current project directory.
Reproducible both on macOS and Windows.

```

## Versions

* Spyder version: 5.3.2
* Python version: 
* Qt version: 
* PyQt version: 
* Operating System name/version: macOS 12, Windows 10

### PR DESCRIPTION
**Pull request by ccordoba12**
## Description of Changes

- This was working before but I don't know when it broke. Now it works in three ways:
    1. When switching between consoles.
    2. When opening new consoles.
    3. When opening/closing projects.
- Add tests to check that the Working directory toolbar and Files are updated in those cases.
- Set the right working directory when opening a console from Files (this was reported on issue #19003).
- Restore functionality of IPython console's public API method `set_working_directory`, which I broke in PR #18544.
- Add a new method to it called `save_working_directory` to save the cwd reported by the Working directory plugin. This is necessary to start new consoles.
- Rename some methods related to the cwd in the IPython console for clarity.

### Issue(s) Resolved

Fixes #19003

### Affirmation

### DIFF
--- a/spyder/plugins/ipythonconsole/widgets/shell.py
+++ b/spyder/plugins/ipythonconsole/widgets/shell.py
@@ -296,8 +296,18 @@ class ShellWidget(NamepaceBrowserWidget, HelpWidget, DebuggingWidget,
         else:
             self.silent_exec_method(code)
 
-    def set_cwd(self, dirname, update_in_spyder=False):
-        """Set shell current working directory."""
+    def set_cwd(self, dirname, emit_cwd_change=False):
+        """
+        Set shell current working directory.
+
+        Parameters
+        ----------
+        dirname: str
+            Path to the new current working directory.
+        emit_cwd_change: bool
+            Whether to emit a Qt signal that informs other panes in Spyder that
+            the current working directory has changed.
+        """
         if os.name == 'nt':
             # Use normpath instead of replacing '\' with '\\'
             # See spyder-ide/spyder#10785
@@ -306,7 +316,7 @@ class ShellWidget(NamepaceBrowserWidget, HelpWidget, DebuggingWidget,
         if self.ipyclient.hostname is None:
             self.call_kernel(interrupt=self.is_debugging()).set_cwd(dirname)
             self._cwd = dirname
-            if update_in_spyder:
+            if emit_cwd_change:
                 self.sig_working_directory_changed.emit(self._cwd)
 
     def get_cwd(self):
--- a/spyder/plugins/ipythonconsole/widgets/shell.py
+++ b/spyder/plugins/ipythonconsole/widgets/shell.py
@@ -296,7 +296,7 @@ class ShellWidget(NamepaceBrowserWidget, HelpWidget, DebuggingWidget,
         else:
             self.silent_exec_method(code)
 
-    def set_cwd(self, dirname):
+    def set_cwd(self, dirname, update_in_spyder=False):
         """Set shell current working directory."""
         if os.name == 'nt':
             # Use normpath instead of replacing '\' with '\\'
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Review by dalthviz — COMMENTED** [review]
Thanks @ccordoba12 ! Checking locally seems like there is an odd behavior when opening a project. Seems like the working directory is not being updated and then any new consoles start with the previous working directory that was set before opening the project:

cwd




Also, when closing the project the working directory still references the project path but then the new console cwd is not set to be the last path but the default user directory?:

imagen

**Comment by ccordoba12** [comment]
Thanks checking those cases @dalthviz! I think I fixed them in my last commit, in which I also expanded the new tests I added them to cover them.

**Comment by dalthviz** [comment]
Thanks for the changes @ccordoba12 ! Seems like the first part is fixed However, the last part (when closing a project the old working directory set is changed to be again the user default/home directory) is still failing. Locally, I can see that when closing a project the path in the working directory toolbar is the correct one for an split second but then after the console finishes loading it changes to use the user default/home directory:

cwd

File: spyder/plugins/ipythonconsole/widgets/shell.py [hunk 1]
```diff
@@ -296,7 +296,7 @@ def check_spyder_kernel(self):
         else:
             self.silent_exec_method(code)
 
-    def set_cwd(self, dirname):
+    def set_cwd(self, dirname, update_in_spyder=False):
```
dalthviz, line 299: Maybe some docstring for this new `update_in_spyder` could be useful? Maybe chaging its name to `emit_cwd_change` could make it more clear?

ccordoba12:
> Locally, I can see that when closing a project the path in the working directory toolbar is the correct one for an split second but then after the console finishes loading it changes to use the user default/home directory

This is the default behavior @dalthviz, i.e. when closing a project the first console's cwd is set to the user's home directory. What would you expect it to be? The cwd before the project was open?

dalthviz:
> What would you expect it to be? The cwd before the project was open?

Yep exactly

ccordoba12:
Ok, that sounds like a reasonable expectation when there's a single console open (as in the gif you posted above). However, if users have many of them, we would have to use as the "previous cwd" the one of the console which has focus before the project is open.

And in that case, I don't know if users would be surprised by us setting the cwd at what for them could appear to be a random directory, specially if they forgot what was the cwd of the last console with focus.

dalthviz:
I see, also if the current behavior is the default behavior then I guess this is ok :+1: . Maybe creating a new issue to discuss that behavior in the ux-improvements repo could be worthy, but for sure not something to address here. So, besides the comment I left for the `update_in_spyder` new kwarg, this LGTM :+1:

spyder/plugins/ipythonconsole/widgets/shell.py
ccordoba12: Done.

ccordoba12:
> Maybe creating a new issue to discuss that behavior in the ux-improvements repo could be worthy

Good idea. I opened https://github.com/spyder-ide/ux-improvements/issues/74 to keep track of this discussion.

dalthviz (approved):
Thanks @ccordoba12 :+1:

PR #19074, opened 2022-08-16
PR: Improve validation for the right Spyder-kernels version

impact27:
## Description of Changes
If the version of spyder and spyder kernels doesn't match, the user will see many errors. This PR show an error if spyder-kernels has an incompatible version. 

- Depends on https://github.com/spyder-ide/spyder-kernels/pull/408
- Depends on https://github.com/spyder-ide/spyder-kernels/pull/409

### Affirmation

### DIFF
--- a/spyder/plugins/ipythonconsole/widgets/main_widget.py
+++ b/spyder/plugins/ipythonconsole/widgets/main_widget.py
@@ -33,6 +33,7 @@ from spyder.api.widgets.main_widget import PluginMainWidget
 from spyder.api.widgets.menus import MENU_SEPARATOR
 from spyder.config.base import (
     get_conf_path, get_home_dir, running_under_pytest)
+from spyder.plugins.ipythonconsole import SpyderKernelError
 from spyder.plugins.ipythonconsole.utils.kernelspec import SpyderKernelSpec
 from spyder.plugins.ipythonconsole.utils.manager import SpyderKernelManager
 from spyder.plugins.ipythonconsole.utils.client import SpyderKernelClient
@@ -2082,6 +2083,8 @@ class IPythonConsoleWidget(PluginMainWidget):
                 config=None,
                 autorestart=True,
             )
+        except SpyderKernelError as e:
+            return (e.args[0], None)
         except Exception:
             error_msg = _("The error is:<br><br>"
                           "<tt>{}</tt>").format(traceback.format_exc())
@@ -2094,6 +2097,8 @@ class IPythonConsoleWidget(PluginMainWidget):
             kernel_manager.start_kernel(stderr=stderr_handle,
                                         stdout=stdout_handle,
                                         env=kernel_spec.env)
+        except SpyderKernelError as e:
+            return (e.args[0], None)
         except Exception:
             error_msg = _("The error is:<br><br>"
                           "<tt>{}</tt>").format(traceback.format_exc())
--- a/spyder/plugins/ipythonconsole/widgets/main_widget.py
+++ b/spyder/plugins/ipythonconsole/widgets/main_widget.py
@@ -1642,6 +1642,7 @@ class IPythonConsoleWidget(PluginMainWidget):
         # Assign kernel manager and client to shellwidget
         kernel_client.start_channels()
         shellwidget = client.shellwidget
+
         if not known_spyder_kernel:
             shellwidget.sig_is_spykernel.connect(
                 self.connect_external_spyder_kernel)
--- a/spyder/plugins/ipythonconsole/widgets/main_widget.py
+++ b/spyder/plugins/ipythonconsole/widgets/main_widget.py
@@ -40,7 +40,9 @@ from spyder.plugins.ipythonconsole.utils.ssh import openssh_tunnel
 from spyder.plugins.ipythonconsole.utils.style import create_qss_style
 from spyder.plugins.ipythonconsole.widgets import (
     ClientWidget, ConsoleRestartDialog, COMPLETION_WIDGET_TYPE,
-    KernelConnectionDialog, PageControlWidget, ShellWidget)
+    KernelConnectionDialog, PageControlWidget, ShellWidget,
+    SPYDER_KERNELS_MIN_VERSION, SPYDER_KERNELS_MAX_VERSION,
+    SPYDER_KERNELS_VERSION)
 from spyder.py3compat import PY38_OR_MORE
 from spyder.utils import encoding, programs, sourcecode
 from spyder.utils.misc import get_error_match, remove_backslashes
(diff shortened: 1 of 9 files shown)


spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -318,8 +325,30 @@ def check_spyder_kernel_callback(self, reply):
...
+            if not version:
+                # The running_under_pytest() part can be removed when
+                # spyder-kernels 3 makes it into conda. This is needed for
+                # the test_conda_env_activation test
+                if running_under_pytest():
+                    return
+
+                if self.is_spyder_kernel:
+                    self.ipyclient.show_kernel_error(
+                        _("spyder_kernels version must be updated"))
```
ccordoba12, line 338:
```suggestion
                if self.is_spyder_kernel:
                    self.ipyclient.show_kernel_error(
                        _("Please update Spyder-kernels to a version greater "
                          "than 3.0")
                    )
```

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -318,8 +325,30 @@ def check_spyder_kernel_callback(self, reply):
...
+                return
+
+            if not check_version_range(version, SPYDER_KERNELS_VERSION):
+                if "dev0" not in version:
+                    # Development versions are acceptable
+                    self.ipyclient.show_kernel_error(
+                        ERROR_SPYDER_KERNEL_VERSION.format(
+                            version,
+                            SPYDER_KERNELS_VERSION
+                            ))
```
ccordoba12, line 348: Given that the kernel version is going to be validated like this now, I think we should remove the code we have for that in:

https://github.com/spyder-ide/spyder/blob/03fde9d4131b8c32874248900d46435c1f9eec03/spyder/plugins/ipythonconsole/widgets/main_widget.py#L1480-L1514

and replace the message shown here by the one in that code, adding to it also the actual Spyder-kernels version because we have a simple way to detect it.

spyder/utils/programs.py
```diff
@@ -798,6 +798,29 @@ def run_terminal_thread():
         raise NotImplementedError
 
 
+def check_version_range(module_version, version):
```
ccordoba12, line 801:
```suggestion
def check_version_range(module_version, version_range):
```

I think it's clearer if we rename here `version` by `version_range`.

spyder/utils/programs.py
```diff
@@ -798,6 +798,29 @@ def run_terminal_thread():
...
+    """
+    Check version string of a module against a required version.
```
ccordoba12, line 803:
```suggestion
    Check if a module's version lies in `version_range`.
```

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -1644,16 +1642,13 @@ def create_client_for_kernel(self, connection_file, hostname, sshkey,
         # Assign kernel manager and client to shellwidget
         kernel_client.start_channels()
         shellwidget = client.shellwidget
+        if not known_spyder_kernel:
```
ccordoba12, line 1645:
```suggestion

        if not known_spyder_kernel:
```

For clarity.

ccordoba12 (commented):
Thanks @impact27 for your work on this! I think it's a really good improvement because it'll allow us to easily check that users have the right Spyder-kernels version for remote kernels.

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -42,6 +44,41 @@
...
+    "create a console for you.<br><br>"
+    "You can install it by activating your environment (if "
+    "necessary) and then running in a system terminal:"
+    "<pre>"
+    "    <tt>{3}</tt>"
+    "</pre>"
+    "or"
+    "<pre>"
+    "    <tt>{4}</tt>"
+    "</pre>"
```
ccordoba12, line 78:
```suggestion
    "The Python environment or installation whose interpreter is located at"
    "<pre>"
    "    <tt>{0}</tt>"
    "</pre>"
    "doesn't have the right version of <tt>spyder-kernels</tt> installed ({1} "
    "instead of {2}). Without this module is not possible for Spyder to "
    "create a console for you.<br><br>"
    "You can install it by activating your environment (if necessary) and "
    "then running in a system terminal:"
    "<pre>"
    "    <tt>{3}</tt>"
    "</pre>"
    "or"
    "<pre>"
    "    <tt>{4}</tt>"
    "</pre>"
```

Improve formatting.

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -247,6 +284,8 @@ def set_kernel_client_and_manager(self, kernel_client, kernel_manager):
         """Set the kernel client and manager"""
         self.kernel_manager = kernel_manager
         self.kernel_client = kernel_client
+        # Send message to kernel to check status
+        self.check_spyder_kernel()
```
ccordoba12, line 288:
```suggestion

        # Send message to kernel to check status
        self.check_spyder_kernel()

```

For clarity.

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -318,8 +357,35 @@ def check_spyder_kernel_callback(self, reply):
         # Process kernel reply
```
ccordoba12, line 357: The docstring of this method is incorrect. I can directly leave a suggestion for it because it's not part of the diff, but I recommend to change it for:

```python
def check_spyder_kernel_callback(self, reply):
    """
    Check if the Spyder-kernels version is the right one after receiving it
    from the kernel.
    """
```

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -318,8 +357,35 @@ def check_spyder_kernel_callback(self, reply):
...
         data = reply.get('data')
         if data is not None and 'text/plain' in data:
-            is_spyder_kernel = data['text/plain']
-            if 'SpyderKernel' in is_spyder_kernel:
+            spyder_kernel_info = ast.literal_eval(data['text/plain'])
+            if not spyder_kernel_info:
+                # The running_under_pytest() part can be removed when
+                # spyder-kernels 3 makes it into conda. This is needed for
```
ccordoba12, line 363:
```suggestion
                # spyder-kernels 3 is released. This is needed for
```

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -318,8 +357,35 @@ def check_spyder_kernel_callback(self, reply):
...
+                if "dev0" not in version:
+                    # Development versions are acceptable
+                    self.ipyclient.show_kernel_error(
+                        ERROR_SPYDER_KERNEL_VERSION.format(
+                            pyexec,
+                            version,
+                            SPYDER_KERNELS_VERSION_MSG,
+                            SPYDER_KERNELS_CONDA,
+                            SPYDER_KERNELS_PIP
+                            ))
```
ccordoba12, line 385:
```suggestion
                        )
                    )
```

ccordoba12 (commented):
Thanks @impact27 for the update! It looks much better now.

ccordoba12:
@impact27, there's a small regression with respect to the current functionality. If Spyder-kernels is not installed in an environment, users would see this message with your PR:

image

But this is what they see with the code in master:

image

Is there a way to preserve this behavior? Perhaps we could catch the `ModuleNotFoundError` and display instead the current message.

ccordoba12 (commented):
Last stylistic suggestions then this should be ready @impact27 (unless, of course, you plan to make further changes).

ccordoba12 (approved):
Great work @impact27! This is a pretty nice improvement!

PR #19092, opened 2022-08-17
PR: Use cached kernel for faster kernel restart (IPython console)

impact27:
## Description of Changes
restart kernel instantaneously by using the cached kernel

depends on https://github.com/spyder-ide/spyder/pull/19411

### Issue(s) Resolved

Fixes #

### Affirmation


spyder/plugins/ipythonconsole/utils/kernel_handler.py
```diff
@@ -592,14 +592,11 @@ def emit_fault_text(self, fault):
             return
         self.sig_fault.emit(fault)
 
-    def restart_kernel(self):
-        """Restart kernel."""
-        if self.kernel_manager is None:
+    def fault_filename(self):
+        """get fault filename"""
```
ccordoba12, line 596:
```suggestion
        """Get fault filename."""
```

spyder/plugins/ipythonconsole/widgets/client.py
```diff
@@ -350,7 +328,20 @@ def connect_kernel(self, kernel_handler):
         self._show_loading_page()
 
         # Actually do the connection
-        self.shellwidget.connect_kernel(kernel_handler)
+        self.shellwidget.connect_kernel(kernel_handler, first_connect)
+
+    def disconnect_kernel(self, shutdown_kernel):
+        """Disconnect kernel."""
```
ccordoba12, line 334:
```suggestion
    def disconnect_kernel(self, shutdown_kernel):
        """Disconnect from current kernel."""
```

spyder/plugins/ipythonconsole/widgets/client.py
```diff
@@ -606,102 +590,16 @@ def interrupt_kernel(self):
...
-                return
-
-            # Start autorestart mechanism
-            sw.kernel_manager.autorestart = True
-            sw.kernel_manager.start_restarter()
-
-            if reset:
-                sw.reset(clear=True)
+    def replace_kernel(self, kernel, shutdown_kernel):
+        """Instead of restarting, close kernel and open a new one"""
```
ccordoba12, line 594:
```suggestion
    def replace_kernel(self, kernel_handler, shutdown_kernel):
        """
        Replace kernel by disconnecting from the current one and connecting to
        another kernel, which is equivalent to a restart.
        """
```

spyder/plugins/ipythonconsole/widgets/client.py
```diff
@@ -606,102 +590,16 @@ def interrupt_kernel(self):
...
+        # Connect kernel to client
+        self.disconnect_kernel(shutdown_kernel)
+        self.connect_kernel(kernel, first_connect=False)
```
ccordoba12, line 597:
```suggestion
        self.connect_kernel(kernel_handler, first_connect=False)
```

spyder/plugins/ipythonconsole/widgets/client.py
```diff
@@ -799,3 +697,4 @@ def set_info_page(self):
                 self.info_page,
                 QUrl.fromLocalFile(self.css_path)
             )
+            self.sig_execution_state_changed.emit()
```
ccordoba12, line 700: Why is this necessary now?

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -209,6 +212,48 @@ def connect_kernel(self, kernel_handler):
...
+        """
+        Disconnect kernel.
+
+        Parameters:
+        -----------
+        shutdown_kernel: bool
+            If True, the kernel is shut down
+        will_reconnect: bool
+            If False, emits `sig_shellwidget_deleted` so the plugins can close
+            related widgets
```
ccordoba12, line 225:
```suggestion
        Disconnect from current kernel.

        Parameters:
        -----------
        shutdown_kernel: bool
            If True, the kernel is shut down.
        will_reconnect: bool
            If False, emits `sig_shellwidget_deleted` so the plugins can close
            related widgets.
```

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -744,12 +786,6 @@ def create_shortcuts(self):
             name='Clear shell',
             parent=self)
 
-        restart_kernel = self.config_shortcut(
-            self.ipyclient.restart_kernel,
-            context='ipython_console',
-            name='Restart kernel',
-            parent=self)
```
ccordoba12, line 751: This shortcut is important, so please find a way to restore it. I know many users (and also I) use it all the time to quickly restart the kernel.

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -1046,11 +1078,31 @@ def _banner_default(self):
 
     def _kernel_restarted_message(self, died=True):
         msg = _("Kernel died, restarting") if died else _("Kernel restarting")
+
+        if died and self.kernel_manager is None:
+            # The kernel might never restart, show position of fault file
+            msg += (
+                "\nCrash file located at: "
+                + self.kernel_handler.fault_filename()
```
ccordoba12, line 1086:
```suggestion
            msg += (
                "\n" + _("Its crash file is located at:)" + " "
                + self.kernel_handler.fault_filename()
```

This makes the new message you added translatable.

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -1046,11 +1078,31 @@ def _banner_default(self):
...
+            )
+
         self.sig_kernel_restarted_message.emit(msg)
 
     def _handle_kernel_restarted(self, *args, **kwargs):
+        """The kernel restarted."""
         super(ShellWidget, self)._handle_kernel_restarted(*args, **kwargs)
```
ccordoba12, line 1093:
```suggestion
        super()._handle_kernel_restarted(*args, **kwargs)
```

Remove old `super` usage.

ccordoba12 (commented):
Some small suggestions for you @impact27, otherwise looks good to me.

ccordoba12:
@impact27, please also take care of the `pep8speaks` warnings above, which are very simple.

spyder/plugins/ipythonconsole/widgets/shell.py
impact27: but it still work, doesn't it? The same shortcut is created in main_widget.py:
```
self.restart_action = self.create_action(
            IPythonConsoleWidgetActions.Restart,
            text=_("Restart kernel"),
            icon=self.create_icon('restart'),
            triggered=self.restart_kernel,
            register_shortcut=True
        )
```

spyder/plugins/ipythonconsole/widgets/client.py
impact27: It was in `_finalise_restart` which was removed

impact27: Also it makes sense that the execution state is changed when an info page is shown.

impact27:
@ccordoba12 the `pep8speaks` are not applicable, the comment has not been updated since august and doesn't corresponds to the current state of the PR.

spyder/plugins/ipythonconsole/widgets/shell.py
ccordoba12: > but it still work, doesn't it?

Nop, it doesn't. I changed it to other shortcuts and they do nothing when pressed on the console. Probably it's necessary to have it here too so that it actually has effect.

ccordoba12:
I tested this manually and found a small regression: the `hr` painted after the `Restarting kernel` message doesn't adjust to the current interface theme, i.e. it's black when using a dark theme:

imagen

Perhaps you introduced that regression in commit 52532adc9782eacd3387093f26ff40d27a740d8a (not sure about that though).

spyder/plugins/ipythonconsole/widgets/shell.py
impact27: That is strange. When launching this PR, I can:

1 - click on the console
2 - "Cmd + ."
3 - A pop up appears asking if I want to restart the kernel
4 - Open settings and change the shortcut to "Cmd + o"
5 - click on the console
6 - "Cmd + o"
7 - A pop up appears asking if I want to restart the kernel

That is not what you are observing? I mean `create_action(register_shortcut=True)` should work and does work in my case, could i be a linux-only bug?

impact27:
> I tested this manually and found a small regression: the `hr` painted after the `Restarting kernel` message doesn't adjust to the current interface theme, i.e. it's black when using a dark theme:

This is strange as well, I see:

Screenshot 2022-11-11 at 06 35 28

Screenshot 2022-11-11 at 06 37 14

Which is what I expect from the line:
`ruler.setBackground(QColor(QStylePalette.COLOR_TEXT_1))`

spyder/plugins/ipythonconsole/widgets/shell.py
ccordoba12: > That is not what you are observing?

Nop. After I change the shortcut in Preferences, it doesn't have any effect in the console.

> could it be a linux-only bug?

Possibly. @dalthviz, could you check if you're able to change the "restart kernel" shortcut on Windows?

ccordoba12: A possible cause for this is that the shortcut is not registered anymore in `ShellWidget`. That's how I understand Qt triggers it when the focus widget is precisely that one and the reason why we had this piece of code you removed here (actually, I don't understand how a different shortcut is working for you on Mac).

ccordoba12:
> This is strange as well, I see

I understood what happens here: if you go to the Option menu and click on the `Restart kernel` action, then the line is painted according to the interface theme (i.e. as in your screenshots above).

However, if you press `Ctrl+.` in the console, then you get the black line (as in my screenshot). I think this problem happens because in this case the shortcut is calling the method to restart kernels available in Qtconsole, not the Spyder one. As evidence of that is the following screenshot, where I set my interface language to Spanish and press `Ctrl+.`:

imagen

You can see that the dialog is not translated, which means that comes from Qtconsole. We prevented that in Spyder 5 by overriding the `request_restart_kernel` method in `ShellWidget`

https://github.com/jupyter/qtconsole/blob/90ffe335cf9c82f7814c9a945553584d6665a62c/qtconsole/frontend_widget.py#L344-L346

but you removed it on this PR. So, that's another regression that you need to address here.

However, I think it should be relatively simple to resolve both this problem and the shortcut one by

1. Adding a new signal to `ShellWidget` called `sig_kernel_restart_requested`.
2. Connecting that signal to the new `restart_kernel` method in `IPythonConsoleWidget`.
3. Restoring `request_restart_kernel` and make it a emit that signal.
4. Restoring the shortcut declaration and make it emit that signal.

impact27:
I understand why it is different on mac: cmd + . works but ctrl + . has the behaviour you described!

spyder/plugins/ipythonconsole/widgets/shell.py
impact27: ctrl + . is hardcoded in qtconsole, so I skipped this. It should work now

ccordoba12: It still doesn't work for me @impact27.

impact27: This is perplexing. I will investigate but because on Mac ctrl + . And cmd + . Are distinct, it is not trivial for me to debug this.

impact27:
My understanding so far is:
in `qtconsole.frontend_widget.FrontendWidget._event_filter_console_keypress`: The console catches `ctrl + .` events and calls `request_restart_kernel`. We want to avoid that because the shortcut can not be changed. This was the original reason this PR didn't work in Linux and Window because on mac `ctrl` and `cmd` are distinct. So the code was messing with  `ctrl + .` but I didn't see it because I used `cmd + .`.

Now that `ctrl + .` is not caught by the console, it is free to propagate up the chain to `main_widget.IPythonConsoleWidget.restart_action`. On my mac, I can change my shortcut from `cmd + .` to `ctrl + .` and it still works. It did not work before my latest change.

This explain everything that you saw, but I am perplexed by the fact that it doesn't work for you @ccordoba12. Could you check that the shortcut was not changed in your settings?

ccordoba12:
> This explain everything that you saw, but I am perplexed by the fact that it doesn't work for you @ccordoba12. Could you check that the shortcut for restart kernel was not changed in your settings?

I'm running Spyder with clean settings and it still doesn't work for me.

But let's try this PR on Binder. A link for it will be generated below by calling the following command (we have an action that detects it and automatically creates the link).

ccordoba12:
No luck either, as you can check as well.

impact27:
I restored it, but I think you can't change the shortcut that way. However it appears to be already the case with master.

spyder/plugins/ipythonconsole/widgets/shell.py
```diff
@@ -873,9 +918,15 @@ def insert_horizontal_ruler(self):
         self._control.insert_horizontal_ruler()
 
     # ---- Public methods (overrode by us) ------------------------------------
-    def request_restart_kernel(self):
-        """Reimplemented to call our own restart mechanism."""
-        self.ipyclient.restart_kernel()
+    def _event_filter_console_keypress(self, event):
```
ccordoba12, line 921: I think this new method overrides the `_event_filter_console_keypress` one in DebuggingWidget (given that that class is sort of a mixin). If that's the case, this code should be moved there.

ccordoba12 (commented):
Thanks @impact27 for your last commit! I can confirm it fixes the restart issue on Linux, although you're right in the sense that changing the shortcut doesn't have any effect.

I left an additional comment for you, then this should be ready.

spyder/plugins/ipythonconsole/widgets/shell.py
impact27: Unless I am mistaken, `return super()._event_filter_console_keypress(event)` will call `DebuggingWidget._event_filter_console_keypress`.
I can confirm I see that when debugging Spyder on my mac.

ccordoba12: You're right, sorry for the noise. I just wasn't sure about this.

ccordoba12 (approved):
Looks good to me now. Thanks @impact27 for your hard work and patience with this one.

This is a really cool improvement!

PR #19163, opened 2022-08-24
PR: Several fixes related to the tabify plugins functionality

ccordoba12:
## Description of Changes

- Fix tabifying external and internal plugins (e.g. Spyder-notebook and Debugger) that are loaded after Spyder runs for the first time.
- Move methods related to tabifying plugins from MainWindow to the Layout plugin.
- Fix error when trying to restore visible plugins that are no longer available (this error was reported by @dalthviz).
- Show visible external plugins when switching layouts. For instance, if Spyder-notebook is displayed in the interface and the user switches to the Matlab layout, then that plugin is displayed again (it was not before).
- Fix `TABIFY` class attribute for the Pylint and Profiler plugins.
- Reorganize code a bit in the Layout plugin.

### Affirmation

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -1012,19 +1012,8 @@ class MainWindow(QMainWindow, SpyderConfigurationAccessor):
                     pass
 
         # Register custom layouts
-        for plugin_name in PLUGIN_REGISTRY.external_plugins:
-            plugin_instance = PLUGIN_REGISTRY.get_plugin(plugin_name)
-            if hasattr(plugin_instance, 'CUSTOM_LAYOUTS'):
-                if isinstance(plugin_instance.CUSTOM_LAYOUTS, list):
-                    for custom_layout in plugin_instance.CUSTOM_LAYOUTS:
-                        self.layouts.register_layout(
-                            self.layouts, custom_layout)
-                else:
-                    logger.info(
-                        'Unable to load custom layouts for {}. '
-                        'Expecting a list of layout classes but got {}'
-                        .format(plugin_name, plugin_instance.CUSTOM_LAYOUTS)
-                    )
+        if self.layouts is not None:
+            self.layouts.register_custom_layouts()
 
         # Needed to ensure dockwidgets/panes layout size distribution
         # when a layout state is already present.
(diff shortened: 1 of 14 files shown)


dalthviz:
Seems like an issue on with the `test_programmatic_custom_layouts` related with:

* `configparser.NoOptionError: No option 'layout_testing layout/size' in section: 'quick_layouts'`
* `AttributeError: 'MainWindow' object has no attribute 'tabify_plugin'`

Maybe with the changes made the `layout_testing layout` doesn't get loaded @ccordoba12 ?

ccordoba12:
I think I fixed that problem in my last commit.

spyder/plugins/layout/plugin.py
```diff
@@ -931,7 +947,7 @@ def restore_visible_plugins(self):
 
         # Restore visible plugins
         for plugin in visible_plugins:
-            plugin_class = self.get_plugin(plugin)
+            plugin_class = self.main.get_plugin(plugin, error=False)
```
dalthviz, line 950: Maybe the addition of the kwarg `error` to `get_plugin` method of the plugin class could be worthy instead of calling `self.main` directly here?

dalthviz, line 950: Maybe the addition of the kwarg `error` to `get_plugin` method of the plugin class could be worthy instead of calling `self.main` directly here?

spyder/app/mainwindow.py
```diff
@@ -1087,7 +1018,7 @@ def pre_visible_setup(self):
                 if isinstance(plugin_instance.CUSTOM_LAYOUTS, list):
                     for custom_layout in plugin_instance.CUSTOM_LAYOUTS:
                         self.layouts.register_layout(
-                            self, custom_layout)
+                            self.layouts, custom_layout)
```
dalthviz, line 1021: Shouldn't this be `plugin_instance` instead of `self.layouts`? Also, maybe moving this code to  register custom layouts inside the Layout plugin and instead here call something like `self.layouts.register_custom_layouts(PLUGIN_REGISTRY.external_plugins)` could be worthy?

dalthviz (commented):
Thanks @ccordoba12 ! I left some comments regarding the code to register custom layouts (maybe could be worthy to be moved to the Layouts plugins), and also the possibility to addof the `error` flag to the plugins `get_plugin` method instead of using the mainwindow version. Other than that this LGTM :+1:

spyder/app/mainwindow.py
ccordoba12: > Shouldn't this be plugin_instance instead of self.layouts?

Nop, it needs to be the Layout plugin instance.

> Also, maybe moving this code to register custom layouts inside the Layout plugin

Good idea. I'll implement it right away.

dalthviz:
@ccordoba12 I think that some tests fixtures that mock the MainWindow will need an update too (like for the Pytlint plugin tests)

ccordoba12:
Yep, I just did it.

dalthviz (approved):

PR #19194, opened 2022-08-26
PR: Pass `PYTHONPATH` to the kernel if not handled by Spyder (IPython console)

### LINKED ISSUE #18942: Starting a new console loses `PYTHONPATH` in Spyder 5.3.2
## Problem Description
In 5.3.2, if you open a new Ipython console, it will NOT carry the PYTHONPATH environment variable which we need for custom projects/packages.  The environment variables work in the first console that was opened when spyder starts, but it goes missing when you open more console (default setting), see screenshot
image

The same issue does NOT exist in 5.3.1 (see screenshot)
image

1. start spyder 5.3.2
2. import os; os.environ['PYTHONPATH']  # this works
3. open another console with default settings
4. import os; os.environ['PYTHONPATH'] throws KeyError 'PYTHONPATH'

PYTHONPATH variable 

```python-traceback

PASTE TRACEBACK HERE

```

## Versions

* Spyder version: 
* Python version: 
* Qt version: 
* PyQt version: 

### Dependencies

```
PASTE DEPENDENCIES HERE
```

ccordoba12:
## Description of Changes

- This was broken in version 5.3.2, but it's been the way we've handled `PYTHONPATH` since Spyder 3. So we need to restore the previous behavior.
- Add a regression test for this.

### Issue(s) Resolved

Fixes #18942.

### Affirmation


mrclary:
@ccordoba12, I don't understand the motivation for this PR. This was not broken in 5.3.2, as far as I know. The changes to handling PYTHONPATH were _designed_ to make the PPM the gateway for all PYTHONPATH paths used by IPython Console and completions, giving the user both control and transparency. This was agreed upon by @spyder-ide/core-developers in #17512.

With this PR:
- Search paths for completions in the Editor will not be the same as search paths in IPython Console
- PPM will not accurately reflect the search paths used by IPython Console
- The `ext_pypath_handled_by_spyder` is determined by the presence of a single common path. The addition/removal of a single path to PPM could add/remove _many_ paths from PYTHONPATH for IPython Console, resulting in unexpected behavior for the user.

ccordoba12:
> I don't understand the motivation for this PR. This was not broken in 5.3.2, as far as I know

The problem is users didn't have to import PYTHONPATH through our PPM before, and that's what they're complaining about. Actually, that has been the default behavior since Spyder 3 (i.e. six years ago), so I think it's not ok to break it in a bugfix release (i.e. 5.3.2).

> Search paths for completions in the Editor will not be the same as search paths in IPython Console

Right, I forgot about it. I can address that as part of this PR.

> The ext_pypath_handled_by_spyder is determined by the presence of a single common path

I'm aware of that (I mentioned it in a comment above that validation). But a more robust validation would require moving code  out of MainWindow to reuse it in other places. And I think we should leave that for Spyder 6 because we could handle that in a simpler way with the new PPM plugin.

mrclary:
> > I don't understand the motivation for this PR. This was not broken in 5.3.2, as far as I know
> 
> The problem is users didn't have to import PYTHONPATH through our PPM before, and that's what they're complaining about.

I think, then, a better solution would be to add a preference to automatically import the system PYTHONPATH. Nevertheless, users only have to do this once.

> Actually, that has been the default behavior since Spyder 3 (released six years ago), so I think it's not ok to break it in a bugfix release (i.e. 5.3.2).

I consider the inconsistent application of PYTHONPATH to be a bug. If you agree with the behavior of #17512 but think we should have waited for a minor release, do you expect to revert the current PR at a later release? If you disagree with the behavior of #17512, then I think we should reopen the discussion and decide on a strategy, which may include a mitigating PR such as this one.

> > The ext_pypath_handled_by_spyder is determined by the presence of a single common path
> 
> I'm aware of that (I mentioned it in a comment above that validation). But a more robust validation would require moving code out of MainWindow to reuse it in other places. And I think we should leave that for Spyder 6 because we could handle it in a simpler way with the new PPM plugin.

Yes, I saw that comment. But that suggests that the PPM plugin would continue the behavior proposed by this PR, which I disagree with and is contrary to the already agreed upon behavior of #17512. If you intend for the behavior of this PR to be permanent, then we need to reopen the discussion.

mrclary:
> Just to keep in mind here, the actual proximate bug with the new functionality that the user reported should be fixed regardless of what we decide

Agreed.

ccordoba12:
> I think, then, a better solution would be to add a preference to automatically import the system PYTHONPATH. Nevertheless, users only have to do this once.

For that we'd need a config page for PPM. Otherwise I think it'd be adding too much content to its dialog.

> I consider the inconsistent application of PYTHONPATH to be a bug

Sure, things were inconsistent before but they could be made consistent by adding PYTHONPATH implicitly to the LSP as well.

> Yes, I saw that comment. But that suggests that the PPM plugin would continue the behavior proposed by this PR

As we already noticed, users find quite counterintuitive that now Spyder doesn't make use of PYTHONPATH when it had done for a long time (as a matter of fact, I find it too). So, I think we should reconsider our decision.

The problem with not fixing this regression is that users that depend on this feature will remain on older versions because those are the ones that work for them. 

> I could see that arguably justifying a minor release (i.e. bump to 5.4.x), but I don't understand why that wasn't brought up before it you merged and released in a bugfix release, after a lengthy discussion you were heavily involved in, and with your approval at the end

I didn't understand all implications of @mrclary's PR.

ccordoba12:
One more thing: when you set PYTHONPATH, you expect that it'll be used to import code in the paths declared by it in order to run it, at least, and having code completions for it.

By forcing our users to go through the extra step of importing it in the PPM, we're breaking that expectation. I think that's why they are complaining about it.

mrclary:
> Sure, things were inconsistent before but they could be made consistent by adding PYTHONPATH implicitly to the LSP as well.

The inconsistent behavior was much broader and more complex than this statement suggests. See the table here: https://github.com/spyder-ide/spyder/issues/17511#issue-1173999486

> As we already noticed, users find quite counterintuitive that now Spyder doesn't make use of PYTHONPATH when it had done it for a long time (as a matter of fact, I find it too). So, I think we should reconsider our decision.
> 
> The problem with not fixing this regression is that users that depend on this feature will remain on older versions because those are the ones that work for them.

Why would they remain on older versions when they can just import their PYTHONPATH?

Nevertheless, I have no problem reopening a deeper discussion regarding the UX of PYTHONPATH and how we should handle it.

ccordoba12:
Closing in favor of #19937 which provides a better solution to this problem.

PR #19360, opened 2022-09-08
PR: Reload symbols when saving a file through the `Save as` menu entry (Outline)

### LINKED ISSUE #15517: Outline pane navigates to the wrong file

- open a file with some cells (outline pane should work now)
- go to "save as" and use a new name
- try to navigate with the outline pane in your renamed file
--> spyder re-opens the old file and jumps to the selected cell

## Versions

* Spyder version: 4.2.3 
* Python version: 3.6.8
* Qt version: 5.9.6
* PyQt5 version: 5.9.2

### Dependencies

```

# Mandatory:
IPython >=7.6.0                 :  7.16.1 (OK)
keyring >=17.0.0                :  22.3.0 (OK)
paramiko >=2.4.0                :  2.7.2 (OK)
pygments >=2.0                  :  2.8.1 (OK)
pylint >=1.0                    :  2.7.4 (OK)
qtconsole >=5.0.1               :  5.0.3 (OK)
rtree >=0.8.3                   :  0.9.7 (OK)
setuptools >=39.0.0             :  52.0.0.post20210125 (OK)
sphinx >=0.6.6                  :  3.5.3 (OK)

# Optional:
cython >=0.21                   :  0.29.23 (OK)
matplotlib >=2.0.0              :  3.1.3 (OK)
numpy >=1.7                     :  1.19.5 (OK)
pandas >=1.1.1                  :  1.1.5 (OK)
scipy >=0.17.0                  :  1.5.3 (OK)
```

ccordoba12:
## Description of Changes

- Symbols were pointing to the original file instead of the renamed one. This PR fixes that by deleting the current symbols and requesting them for the new file.

    **Before**

    rename-wrong

    **After**

    rename-right

- Add a regression test for this.
- Expand Outline tests to check that we can display special comments and cells.
- Remove some dead code in the Outline and its tests.

### Issue(s) Resolved

Fixes #15517

### Affirmation

### DIFF
--- a/spyder/plugins/outlineexplorer/tests/test_widgets.py
+++ b/spyder/plugins/outlineexplorer/tests/test_widgets.py
@@ -5,13 +5,13 @@
 #
 
 """
-Tests for editortool.py
+Tests for the Outline explorer widgets.
 """
+
 # Standard Libray Imports
 import json
 import os.path as osp
 import sys
-from textwrap import dedent
 from unittest.mock import MagicMock
 
 # Third party imports
@@ -23,10 +23,10 @@ from qtpy.QtCore import Qt
 from spyder.plugins.outlineexplorer.editor import OutlineExplorerProxyEditor
 from spyder.plugins.outlineexplorer.main_widget import (
     OutlineExplorerWidget, OutlineExplorerToolbuttons)
-from spyder.plugins.outlineexplorer.widgets import (
-    FileRootItem, SymbolStatus, TreeItem)
 from spyder.plugins.editor.widgets.codeeditor import CodeEditor
 
+
+# ---- Constants
 HERE = osp.abspath(osp.dirname(__file__))
 ASSETS = osp.join(HERE, 'assets')
 SUFFIX = 'test_widgets'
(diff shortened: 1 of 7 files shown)


spyder/plugins/outlineexplorer/tests/test_widgets.py
```diff
@@ -298,120 +208,6 @@ def test_go_to_last_item(create_outlineexplorer, qtbot):
     assert outlineexplorer.treewidget.currentItem().text(0) == 'method1'
 
 
-@pytest.mark.skip(reason='Cell support is disabled temporarily')
-def test_code_cell_grouping(create_outlineexplorer):
```
dalthviz, line 302: Is this test still failing? I think cell support is available, right? Seems like it could help checking the correct behavior of the Outline Explorer, maybe we should try to restore it instead of deleting it?

spyder/plugins/outlineexplorer/widgets.py
```diff
@@ -215,61 +220,6 @@ def update_info(self, name, kind, position, status, selected):
         self.setSelected(selected)
 
 
-class TreeItem(QTreeWidgetItem):
```
dalthviz, line 218: Is this class not used anymore anywhere? Just dead code, right?

dalthviz (commented):
Thanks @ccordoba12 ! Checked locally on WIndows and seems like the changes work there too :+1:

Left some comments/questions regarding some code (a class and a test) that were removed here. Otherwise this LGTM :+1:

spyder/plugins/outlineexplorer/tests/test_widgets.py
ccordoba12: This test worked for the previous implementation of the Outline (i.e. the one that didn't depend on the LSP), so it's not possible to restore it now.

However, I think there are some validations on it that are not performed in the new tests, so I'll try to port them.

spyder/plugins/outlineexplorer/widgets.py
ccordoba12: Yep, I checked and we're not using in any other place. Furthermore, I think it was the tree item used in the previous implementation, so we can safely remove it now.

dalthviz (approved):

Issue #16265, opened 2021-08-23
Outline pane is automatically shown when editor is maximized


newyorker123:
### Issue Report Checklist

* [ ] Searched the [issues page](https://github.com/spyder-ide/spyder/issues?q=is%3Aissue) for similar reports
* [ ] Read the relevant sections of the [Spyder Troubleshooting Guide](https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ) and followed its advice
* [ ] Reproduced the issue after updating with ``conda update spyder`` (or ``pip``, if not using Anaconda)
* [ ] Could not reproduce inside ``jupyter qtconsole`` (if console-related)
* [ ] Tried basic troubleshooting (if a bug/error)
    * [ ] Restarted Spyder
    * [ ] Reset preferences with ``spyder --reset``
    * [ ] Reinstalled the latest version of [Anaconda](https://www.anaconda.com/download/)
    * [ ] Tried the other applicable steps from the Troubleshooting Guide
* [ ] Completed the **Problem Description**, **Steps to Reproduce** and **Version** sections below

## Problem Description
When I click the maximize current pane on the top tool bar to maximize editor, the outline pane automatically shown on the right side. But in the view option, the outline is actually not checked. I have to first check the outline option(nothing happens), then check it again to close the outline pane.
 
<img width="1440" alt="截屏2021-08-23 下午5 03 20" src="https://user-images.githubusercontent.com/43869995/130421441-2ce52a98-8e8a-4fbd-a355-5ce8f5027e94.png">

 

### What steps reproduce the problem?

1. click the maximize current pane to maximize editor
2. the outline pane is shown on the right side
3. 

### What is the expected output? What do you see instead?
Outline pane not shown.
Or at least the outline option should be checked when shown. 

### Paste Traceback/Error Below (if applicable)

```python-traceback

PASTE TRACEBACK HERE

```

## Versions

* Spyder version: 5.1.1 (also happen in 5.0.5)
* Python version: 3.8.8
* Qt version: 
* PyQt version: 
* Operating System name/version: Mac BigSur

### Dependencies

```
PASTE DEPENDENCIES HERE
```

steff456:
Hi @newyorker123,

When you hit maximize the editor, it will open by default the outline on the right side. What is weird is that the pane is not checked on the menu, we will look into to add the fix in our future releases.

Thanks for reporting!

ccordoba12:
[...]

PR #19448, opened 2022-09-14
PR: Greatly improve Outline performance and update it correctly when it becomes visible

### LINKED ISSUE #16634: Undocked Outline doesn't update
When the outline is undocked, it doesn't update when a new file is open. 
Example: I have the outline undock on a second monitor and I just open `epochs.py`.

Screenshot 2021-10-25 at 11 07 18

Now if I dock it, it updates.

Screenshot 2021-10-25 at 11 08 09

If I undock it, it works as expected.

Screenshot 2021-10-25 at 11 08 51

Until I open a new file, e.g. `chpi.py`, which will not be displayed in the outline:

Screenshot 2021-10-25 at 11 09 27

Note that if I switch in the editor to the previous file, `epochs.py`, I do get the correct outline.

Screenshot 2021-10-25 at 11 10 09

----------------------

Conclusion, it looks like the outline doesn't update itself when a file is open if it is not dock. 
Note that I am on macOS and at some point on this installation I used Kite. It was messing up with the outline and with the editor. Since then, I managed to remove Kite from the system (a pain!!) and it's better. But it might have something to do with the outline behavior

----------------------

Running spyder 5.1.5 with Python 3.8.10 on macOS 11.6.

```
# Mandatory:
applaunchservices >=0.1.7     :  0.2.1 (OK)
cloudpickle >=0.5.0           :  2.0.0 (OK)
IPython >=7.6.0               :  7.28.0 (OK)
jedi >=0.17.2;<0.19.0         :  0.18.0 (OK)
jsonschema >=3.2.0            :  4.1.1 (OK)
keyring >=17.0.0              :  23.2.1 (OK)
nbconvert >=4.0               :  6.2.0 (OK)
parso >=0.7.0;<0.9.0          :  0.8.2 (OK)
pygments >=2.0                :  2.10.0 (OK)
pylint >=2.5.0;<2.10.0        :  2.9.6 (OK)
pylsp >=1.2.2;<1.3.0          :  1.2.4 (OK)
pylsp_black >=1.0.0           :  None (OK)
qstylizer >=0.1.10            :  0.2.1 (OK)
qtawesome >=1.0.2             :  1.0.3 (OK)
qtconsole >=5.1.0             :  5.1.1 (OK)
qtpy >=1.5.0                  :  1.11.2 (OK)
setuptools >=49.6.0           :  58.2.0 (OK)
sphinx >=0.6.6                :  4.2.0 (OK)
spyder_kernels >=2.1.1;<2.2.0 :  2.1.3 (OK)
watchdog >=0.10.3             :  2.1.6 (OK)
zmq >=17                      :  22.3.0 (OK)

# Optional:
cython >=0.21                 :  None (OK)
matplotlib >=2.0.0            :  3.4.3 (OK)
numpy >=1.7                   :  1.21.2 (OK)
pandas >=1.1.1                :  1.3.3 (OK)
scipy >=0.17.0                :  1.7.1 (OK)
sympy >=0.7.3                 :  None (OK)
```

ccordoba12:
## Description of Changes

* Simplify and greatly speed up updating trees when new symbols arrive from PyLSP. For instance, this decreases the time necessary to update the tree of our `codeeditor.py` file from 700 ms to 30 ms in my laptop.
* Update trees when the Outline is undocked, but not if its window is minimized.
* Avoid the Editor plugin to update current tree when given focus (this was very inefficient).
* Dock plugins if they are undocked before hiding them when going to the `Panes > View` menu. I discovered that UX bug while working on this PR.
* Add a test to check that we update (don't update) the Outline when it becomes visible (is not visible) under different circumstances.

### Issue(s) Resolved

Fixes #16634

### Affirmation

### DIFF
--- a/spyder/api/widgets/auxiliary_widgets.py
+++ b/spyder/api/widgets/auxiliary_widgets.py
@@ -9,7 +9,7 @@ Spyder API auxiliary widgets.
 """
 
 # Third party imports
-from qtpy.QtCore import Signal, QSize
+from qtpy.QtCore import QEvent, Qt, QSize, Signal
 from qtpy.QtWidgets import QMainWindow, QSizePolicy, QToolBar, QWidget
 
 # Local imports
@@ -18,12 +18,23 @@ from spyder.utils.stylesheet import APP_STYLESHEET
 
 
 class SpyderWindowWidget(QMainWindow):
-    """MainWindow subclass that contains a Spyder Plugin."""
+    """MainWindow subclass that contains a SpyderDockablePlugin."""
 
     # ---- Signals
     # ------------------------------------------------------------------------
     sig_closed = Signal()
     """This signal is emitted when the close event is fired."""
+
+    sig_window_state_changed = Signal(Qt.WindowStates)
+    """
+    This signal is emitted when the window state has changed (for instance,
+    between maximized and minimized states).
+
+    Parameters
+    ----------
+    window_state: Qt.WindowStates
+        The window state.
+    """
 
     def __init__(self, widget):
         super().__init__()
@@ -36,6 +47,15 @@ class SpyderWindowWidget(QMainWindow):
         """Override Qt method to emit a custom `sig_close` signal."""
         super().closeEvent(event)
         self.sig_closed.emit()
+
+    def changeEvent(self, event):
+        """
+        Override Qt method to emit a custom `sig_windowstate_changed` signal
+        when there's a change in the window state.
+        """
+        if event.type() == QEvent.WindowStateChange:
+            self.sig_window_state_changed.emit(self.windowState())
+        super().changeEvent(event)
 
 
 class MainCornerWidget(QToolBar):
--- a/spyder/api/widgets/main_widget.py
+++ b/spyder/api/widgets/main_widget.py
@@ -885,6 +885,10 @@ class PluginMainWidget(QWidget, SpyderWidgetMixin, SpyderToolbarMixin):
         if not self.dockwidget:
             return
 
+        # Dock plugin if it's undocked before hiding it.
+        if self.windowwidget is not None:
+            self.close_window(save_undocked=True)
+
         if checked:
             self.dockwidget.show()
             self.dockwidget.raise_()
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -4343,6 +4343,118 @@ def test_update_outline(main_window, qtbot, tmpdir):
 
 @pytest.mark.slow
 @flaky(max_runs=3)
+@pytest.mark.use_introspection
+@pytest.mark.order(after="test_debug_unsaved_function")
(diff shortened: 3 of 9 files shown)


dalthviz (commented):
Thanks @ccordoba12 ! Seems like things are working as described in the OP :+1:

However, regarding the issue of the undocked panes not respecting the selection on the `View  Panes` menu, we shouldn't instead of docking the pane, save the window pane geometry and restored it when checking the pane entry again from the menu (keeping the undocked panes state)?

Also, while checking this, I noticed another issue related with maximizing the Editor action and the Outline pane state on the `View > Panes` menu. Seems like although the Outline is being showed along side the Editor (being the Outline previously not visible), the checkbox state for the Outline is not showing is visible. I think this is not caused by the changes done here but maybe could be worthy to fix it here?

ccordoba12:
> However, regarding the issue of the undocked panes not respecting the selection on the View Panes menu, we shouldn't instead of docking the pane, save the window pane geometry and restored it when checking the pane entry again from the menu (keeping the undocked panes state)?

That's a very interesting idea! And something that will bring cool interactions in Spyder, like showing certain plugins on demand out of the main window. In other words, if users hide a plugin by going to `View > Panes` while it's undocked, and then press the shortcut to show it, with your suggestion it'd be displayed as undocked again.

I started to work on this idea but I think we'd need to differentiate between the `Dock` (bring back to the main window) and `Close` (hide) actions when the plugin is undocked for users to make sense of it. So I think it's work for another PR because my proposed fix is really simple but what you suggest would require things out of the current scope of this PR.

> Seems like although the Outline is being showed along side the Editor (being the Outline previously not visible), the checkbox state for the Outline is not showing is visible. I think this is not caused by the changes done here but maybe could be worthy to fix it here?

That's a good point too, but we could discuss how to solve (or even if we should show the Outline next to the Editor) when we're migrating the editor to the new API. What do you think?

dalthviz:
Then I think we need to create follow-up issues for both things for further discussion: the behavior when an undocked pane is requested to be hidden and the coherence between the Panes menu checkboxes and the maximize action (mostly for the Editor case where also the Outline is shown)

ccordoba12:
> the behavior when an undocked pane is requested to be hidden

See PR #19784. I'll rebase and finish it after this one is merged.

> the coherence between the Panes menu checkboxes and the maximize action (mostly for the Editor case where also the Outline is shown)

Instead of opening a new issue, I posted https://github.com/spyder-ide/spyder/issues/16265#issuecomment-1274109141 because precisely tracking if the Outline is visible when the Editor is maximized would help us to solve that issue in an intuitive way.

ccordoba12:
@dalthviz, given my last comment, I think this one is ready to be merged.

ccordoba12:
Note: I decided to rebase this PR in order to re-run our tests and check that everything is green, given that it's almost a month old.

dalthviz (approved):

PR #19461, opened 2022-09-15
PR: Create conda-based application installer for macOS and Linux

### LINKED ISSUE #18870: Create conda-based standalone application
Making conda-based standalone applications for macOS, Windows, and Linux is desirable to facilitate installing/uninstalling third-party plugins.

mrclary:
## Description of Changes

Creates installers for conda-based standalone applications for macOS, Windows, and Linux.

The installers created by this PR are experimental, at least until the release of Spyder 6.0. They will be produced in addition to the existing macOS and Windows installers.

Given the experimental nature of these installers, any changes required in Spyder's core source code will be made in a separate branch, and a patch supplied in this PR. In this way, _no changes will be made to Spyder's core source code_ by this PR; the patch is applied to Spyder's conda package built for this installer only. There will be no affect on Spyder's conda package uploaded to conda-forge, nor it existing macOS or Windows installers.

The objectives of this PR are:
- [x] Create a viable conda-based macOS installer
- [ ] Establish baseline GatHub workflow that does not block merging.

The Linux and Windows installers will come in follow-up PRs. A non-blocking workflow action will permit continued development on the experimental installers without blocking other pull requests.

At present, this package installer only works on x86 architecture. A subsequent PR will add arm64 to the matrix workflow.

### Issue(s) Resolved

Fixes #18870

### Affirmation


mrclary:
So, I'm having trouble building the `python-lsp-server` conda package from the local `external-deps` source. So, TL;DR, `setuptools-scm` is unable to detect the version since the source code directory is not a fully intact git repository.

<details><summary>Error</summary>

```python
    File "/Users/rclary/opt/mambaforge/envs/spy-inst/conda-bld/python-lsp-server_1663626470544/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib/python3.10/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 363, in <module>
      main()
    File "/Users/rclary/opt/mambaforge/envs/spy-inst/conda-bld/python-lsp-server_1663626470544/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib/python3.10/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 345, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/Users/rclary/opt/mambaforge/envs/spy-inst/conda-bld/python-lsp-server_1663626470544/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib/python3.10/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 164, in prepare_metadata_for_build_wheel
      return hook(metadata_directory, config_settings)
    File "/Users/rclary/opt/mambaforge/envs/spy-inst/conda-bld/python-lsp-server_1663626470544/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib/python3.10/site-packages/setuptools/build_meta.py", line 377, in prepare_metadata_for_build_wheel
      self.run_setup()
    File "/Users/rclary/opt/mambaforge/envs/spy-inst/conda-bld/python-lsp-server_1663626470544/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib/python3.10/site-packages/setuptools/build_meta.py", line 335, in run_setup
      exec(code, locals())
    File "<string>", line 9, in <module>
    File "/Users/rclary/opt/mambaforge/envs/spy-inst/conda-bld/python-lsp-server_1663626470544/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib/python3.10/site-packages/setuptools/__init__.py", line 87, in setup
      return distutils.core.setup(**attrs)
    File "/Users/rclary/opt/mambaforge/envs/spy-inst/conda-bld/python-lsp-server_1663626470544/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib/python3.10/site-packages/setuptools/_distutils/core.py", line 147, in setup
      _setup_distribution = dist = klass(attrs)
    File "/Users/rclary/opt/mambaforge/envs/spy-inst/conda-bld/python-lsp-server_1663626470544/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib/python3.10/site-packages/setuptools/dist.py", line 476, in __init__
      _Distribution.__init__(
    File "/Users/rclary/opt/mambaforge/envs/spy-inst/conda-bld/python-lsp-server_1663626470544/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib/python3.10/site-packages/setuptools/_distutils/dist.py", line 280, in __init__
      self.finalize_options()
    File "/Users/rclary/opt/mambaforge/envs/spy-inst/conda-bld/python-lsp-server_1663626470544/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib/python3.10/site-packages/setuptools/dist.py", line 900, in finalize_options
      ep(self)
    File "/Users/rclary/opt/mambaforge/envs/spy-inst/conda-bld/python-lsp-server_1663626470544/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib/python3.10/site-packages/setuptools_scm/integration.py", line 123, in infer_version
      _assign_version(dist, config)
    File "/Users/rclary/opt/mambaforge/envs/spy-inst/conda-bld/python-lsp-server_1663626470544/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib/python3.10/site-packages/setuptools_scm/integration.py", line 62, in _assign_version
      _version_missing(config)
    File "/Users/rclary/opt/mambaforge/envs/spy-inst/conda-bld/python-lsp-server_1663626470544/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib/python3.10/site-packages/setuptools_scm/__init__.py", line 109, in _version_missing
      raise LookupError(
  LookupError: setuptools-scm was unable to detect version for /Users/rclary/opt/mambaforge/envs/spy-inst/conda-bld/python-lsp-server_1663626470544/work.

  Make sure you're either building from a fully intact git repository or PyPI tarballs. Most other sources (such as GitHub's tarballs, a git checkout without the .git folder) don't contain the necessary metadata and will not work.

  For example, if you're using pip, instead of https://github.com/user/proj/archive/master.zip use git+https://github.com/user/proj.git#egg=proj
  error: subprocess-exited-with-error
  
  × Preparing metadata (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> See above for output.
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  full command: /Users/rclary/opt/mambaforge/envs/spy-inst/conda-bld/python-lsp-server_1663626470544/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/bin/python /Users/rclary/opt/mambaforge/envs/spy-inst/conda-bld/python-lsp-server_1663626470544/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib/python3.10/site-packages/pip/_vendor/pep517/in_process/_in_process.py prepare_metadata_for_build_wheel /tmp/tmpthdb46ws
  cwd: /Users/rclary/opt/mambaforge/envs/spy-inst/conda-bld/python-lsp-server_1663626470544/work
  Preparing metadata (pyproject.toml): finished with status 'error'
error: metadata-generation-failed

× Encountered error while generating package metadata.
╰─> See above for output.

note: This is an issue with the package mentioned above, not pip.
hint: See above for details.
Exception information:
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/operations/build/metadata.py", line 35, in generate_metadata
    distinfo_dir = backend.prepare_metadata_for_build_wheel(metadata_dir)
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/utils/misc.py", line 706, in prepare_metadata_for_build_wheel
    return super().prepare_metadata_for_build_wheel(
  File "$PREFIX/lib/python3.10/site-packages/pip/_vendor/pep517/wrappers.py", line 188, in prepare_metadata_for_build_wheel
    return self._call_hook('prepare_metadata_for_build_wheel', {
  File "$PREFIX/lib/python3.10/site-packages/pip/_vendor/pep517/wrappers.py", line 322, in _call_hook
    self._subprocess_runner(
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/utils/subprocess.py", line 252, in runner
    call_subprocess(
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/utils/subprocess.py", line 224, in call_subprocess
    raise error
pip._internal.exceptions.InstallationSubprocessError: Preparing metadata (pyproject.toml) exited with 1

The above exception was the direct cause of the following exception:

  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/cli/base_command.py", line 167, in exc_logging_wrapper
    status = run_func(*args)
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/cli/req_command.py", line 247, in wrapper
    return func(self, options, args)
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/commands/install.py", line 369, in run
    requirement_set = resolver.resolve(
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/resolution/resolvelib/resolver.py", line 73, in resolve
    collected = self.factory.collect_root_requirements(root_reqs)
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/resolution/resolvelib/factory.py", line 491, in collect_root_requirements
    req = self._make_requirement_from_install_req(
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/resolution/resolvelib/factory.py", line 453, in _make_requirement_from_install_req
    cand = self._make_candidate_from_link(
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/resolution/resolvelib/factory.py", line 206, in _make_candidate_from_link
    self._link_candidate_cache[link] = LinkCandidate(
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/resolution/resolvelib/candidates.py", line 297, in __init__
    super().__init__(
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/resolution/resolvelib/candidates.py", line 162, in __init__
    self.dist = self._prepare()
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/resolution/resolvelib/candidates.py", line 231, in _prepare
    dist = self._prepare_distribution()
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/resolution/resolvelib/candidates.py", line 308, in _prepare_distribution
    return preparer.prepare_linked_requirement(self._ireq, parallel_builds=True)
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/operations/prepare.py", line 438, in prepare_linked_requirement
    return self._prepare_linked_requirement(req, parallel_builds)
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/operations/prepare.py", line 524, in _prepare_linked_requirement
    dist = _get_prepared_distribution(
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/operations/prepare.py", line 68, in _get_prepared_distribution
    abstract_dist.prepare_distribution_metadata(
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/distributions/sdist.py", line 61, in prepare_distribution_metadata
    self.req.prepare_metadata()
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/req/req_install.py", line 533, in prepare_metadata
    self.metadata_directory = generate_metadata(
  File "$PREFIX/lib/python3.10/site-packages/pip/_internal/operations/build/metadata.py", line 37, in generate_metadata
    raise MetadataGenerationFailed(package_details=details) from error
pip._internal.exceptions.MetadataGenerationFailed: metadata generation failed
Removed file://$SRC_DIR from build tracker '/private/tmp/pip-build-tracker-4g4fih3v'
Removed build tracker: '/private/tmp/pip-build-tracker-4g4fih3v'
  File "/Users/rclary/opt/mambaforge/envs/spy-inst/bin/conda-mambabuild", line 10, in <module>
    sys.exit(main())
  File "/Users/rclary/opt/mambaforge/envs/spy-inst/lib/python3.9/site-packages/boa/cli/mambabuild.py", line 239, in main
    call_conda_build(action, config)
  File "/Users/rclary/opt/mambaforge/envs/spy-inst/lib/python3.9/site-packages/boa/cli/mambabuild.py", line 211, in call_conda_build
    result = api.build(
  File "/Users/rclary/opt/mambaforge/envs/spy-inst/lib/python3.9/site-packages/conda_build/api.py", line 186, in build
    return build_tree(
  File "/Users/rclary/opt/mambaforge/envs/spy-inst/lib/python3.9/site-packages/conda_build/build.py", line 3091, in build_tree
    packages_from_this = build(metadata, stats,
  File "/Users/rclary/opt/mambaforge/envs/spy-inst/lib/python3.9/site-packages/conda_build/build.py", line 2214, in build
    utils.check_call_env(cmd, env=env, rewrite_stdout_env=rewrite_env,
  File "/Users/rclary/opt/mambaforge/envs/spy-inst/lib/python3.9/site-packages/conda_build/utils.py", line 412, in check_call_env
    return _func_defaulting_env_to_os_environ('call', *popenargs, **kwargs)
  File "/Users/rclary/opt/mambaforge/envs/spy-inst/lib/python3.9/site-packages/conda_build/utils.py", line 392, in _func_defaulting_env_to_os_environ
    raise subprocess.CalledProcessError(proc.returncode, _args)
subprocess.CalledProcessError: Command '['/bin/bash', '-o', 'errexit', '/Users/rclary/opt/mambaforge/envs/spy-inst/conda-bld/python-lsp-server_1663626470544/work/conda_build.sh']' returned non-zero exit status 1.

```
</details>

Using the `meta.yaml` recipe file from `conda-forge/python-lsp-server-feedstock` as the template, and only replacing the `source`, I have the following success/failure:
- failure: source set to the path of `external-deps/python-lsp-server`
- success: source set to the path of my local clone of `python-lsp-server`
- success: source set to the path of `external-deps/python-lsp-server` and `.git` from local clone copied to `external-deps` subrepo
- success: source set to the path of `external-deps/python-lsp-server` and `PKG_INFO` (from PyPi distribution)  copied to `external-deps` subrepo

I do not have any issues building conda packages from any of the other `external-deps` subrepos, with similar adaptation to their respective feedstock recipe files. I guess they don't use the `setuptools-scm` mechanisms.

The question, then, is how best to proceed? Is there a way to make conda-build work with local subrepos in a robust way? I don't see a way to make this work and the only way forward that I can see is to provide a fully intact local git repository. Providing the Github repo url is convenient, since it doesn't change, but it won't have developer changes, so it's useless for the developer build. A local repository will have the developer's changes, but could be anywhere so it must be another parameter; similarly, a forked github repo will have developer changes, but must be another parameter. This seems particularly obnoxious to me since the all the changes and information is right there in the subrepo, but `setuptools-scm` doesn't like it.

@spyder-ide/core-developers if you have any suggestions or further insight, I'd appreciate it.

<details><summary>meta.yaml</summary>

```yaml
{% set name = "python-lsp-server" %}
{% set version = "1.5.0" %}

package:
  name: {{ name|lower }}
  version: {{ version }}

source:
  - path: /Users/rclary/Documents/Repos/Spyder-IDE/spyder/external-deps/python-lsp-server

build:
  number: 0
  noarch: python
  script: {{ PYTHON }} -m pip install . --no-deps -vv
  entry_points:
    - pylsp = pylsp.__main__:main

requirements:
  host:
    - python >=3.7
    - pip
    # Needed at build time to produce a package with the correct version for pip
    # See https://github.com/pypa/setuptools_scm/issues/636#issuecomment-1026435048
    - setuptools_scm
  run:
    - autopep8 >=1.6.0,<1.7.0
    - flake8 >=4.0.0,<4.1.0
    - jedi >=0.17.2,<0.19
    - mccabe >=0.6.0,<0.7.0
    - pluggy >=1.0.0
    - python-lsp-jsonrpc >=1.0.0
    - pycodestyle >=2.8.0,<2.9.0
    - pydocstyle >=2.0.0
    - pyflakes >=2.4.0,<2.5.0
    - pylint >=2.5.0
    - python >=3.7
    - rope >=0.10.5
    - setuptools >=39.0.0
    - ujson >=3.0
    - whatthepatch
    - yapf

about:
  home: https://github.com/python-lsp/python-lsp-server
  license: MIT
  license_family: MIT
  license_file: LICENSE
  summary: An implementation of the Language Server Protocol for Python
  description: |
    A Python 3.7+ implementation of the Language Server Protocol
    making use of Jedi, pycodestyle, Pyflakes and YAPF.
  dev_url: https://github.com/python-lsp/python-lsp-server

extra:
  recipe-maintainers:
    - ccordoba12
    - andfoy
    - steff456
```

dalthviz:
Not sure, but checking seems like there is a way to setup a version value via the environment variable `SETUPTOOLS_SCM_PRETEND_VERSION`: https://github.com/pypa/setuptools_scm/#environment-variables Maybe that could help? Also, an example of the usage here: https://github.com/deltachat/deltachat-core-rust/blob/8ab3415c5803c7826245d3eca085b873ece38895/python/tests/package_wheels.py

ccordoba12:
@mrclary, you can set env vars in conda recipes, so you probably need to modify the PyLSP one in conda-forge to add the env var mentioned by @dalthviz.

mrclary:
@dalthviz @ccordoba12, That was a good idea; unfortunately, I still get the same error.
I tried setting the environment variable in the shell environment as well as in meta.yaml without affect.

I think the best solution I've come with so far is to just `git init` the external deps directories and later remove the `.git` directory.

ccordoba12:
Ok, that's a good idea too.

CAM-Gerlach:
See, e.g., conda-forge/twine-feedstock#52 ; did you add the correct versions of `setuptools` and `setuptools_scm` as build-time deps in the recipe? At least in my testing, I was able to get everything working just doing that, without having to set `SETUPTOOLS_SCM_PRETEND_VERSION` (though at least on CF, that didn't work either until I added the build-time deps).

mrclary:
> See, e.g., [conda-forge/twine-feedstock#52](https://github.com/conda-forge/twine-feedstock/pull/52) ; did you add the correct versions of `setuptools` and `setuptools_scm` as `host` deps in the recipe?
The recipe only specifies `setuptools-scm` without constraint in the `host` directive. It looks like the following are installed during the build:
```
setuptools:         65.3.0-pyhd8ed1ab_1       conda-forge
setuptools-scm:     7.0.5-pyhd8ed1ab_0        conda-forge
setuptools_scm:     7.0.5-hd8ed1ab_0          conda-forge
```
Nevertheless, adding the constraints does not affect anything.

> At least in my testing, I was able to get everything working just doing that

You changed the `source:` from `url: <blah blah>` to `path: <path to python-lsp-server subrepo>` and it worked?

> without having to set `SETUPTOOLS_SCM_PRETEND_VERSION` (though at least on CF, that didn't work either until I added the build-time deps).

I've set this environment variable without affect, although I don't know what the parenthetical statement means.

installers-exp/spyder-app/meta.yaml
```diff
@@ -0,0 +1,119 @@
+{% set name = "spyder" %}
+{% set version = "5.4.0.dev0" %}
+{% set build = 0 %}
+
+package:
+  name: spyder
```
CAM-Gerlach, line 6:
```suggestion
  name: {{ name }}
```

Wouldn't it make sense to use `name` here (and elsewhere), since you've defined it above?

CAM-Gerlach:
Sorry for the confusion, I didn't read carefully, and rather just noticed that `setuptools_scm` wasn't in the `host` dep list of the `meta.yml`s in this PR, without actually seeing that you were using the `meta.yml` from CF for `python-lsp-server`, which does have it and works.

Basically, what seems to be happening is since unlike `git subodule`, `git subrepo` is not a standard Git feature and the subrepo is not itself a valid git repo with a `.git` directory inside that directory, `setuptools_scm` naturally can't find the version that way. Unfortunately, after a big of digging, I can't find an obvious way to extract this metadata in a standard form, aside from just cloning the commit specified in the `.gitrepo` config file and then pasting in the working dir on top of it.

As you discovered, it still works with sdists (like on Conda Forge) because evidently `setuptools_scm` also looks for the version in the (not actually standardized) `PKG-INFO` metadata at the root of the unpacked sdist, i.e. the current directory in that workflow, but that of course isn't isn't the case here—though we _could_ write one that just contains `Version: X.Y.Z.dev0` or something, which of course we'd itself have to get from somewhere.

There's also `SETUPTOOLS_SCM_PRETEND_VERSION` which again requires getting the version from somewhere else, but setting that env variable doesn't work because of PEP 517 build isolation, as it does not get propagated to the build environment. You might be able to add `--no-build-isolation` to the `pip` invocation and it should pass though, assuming there isn't another issue with it—though again, that's not ideal, though perhaps the least annoying and perhaps there's a better way to do that.

CAM-Gerlach:
To recap from our meeting, you can get the version that `setuptools_scm` would generate by installing `setuptools_scm` in your local Python environment, via `pipx` or another method, and running `python -m setuptools_scm`, which should print the version string to `stdout`. You can run `python -m setuptoosl_scm --help` to see other options, if needed.

mrclary:
@spyder-ide/core-developers, the first successful build artifact for macOS is now up.
https://github.com/spyder-ide/spyder/suites/8445398065/artifacts/374430312

Please test at your leisure and let me know what you think and what issues you find.
Note that the package installer is note notarized, so you will have to make two attempts to open it; the second attempt must be a right-click->open in order to circumvent gatekeeper. Release versions will be notarized (so look for 5.4.0 soon), but this installer will remain experimental at least until 6.0.

@CAM-Gerlach, thanks for pointing me to `setuptools_scm`; that works much better than my previous paradigm. Additionally, I was able to do a shallow clone using `--shallow-exclude='v1.4.1'`. The caveat here is that the 2nd-to-latest version must be specified since it excludes that commit as well as prior history and scm will require at least one tag to reference (in this case `v1.5.0`).

installers-exp/spyder-app/meta.yaml
mrclary: I've taken a lot from napri's workflow, with my primary goal to get something that "just works". My latest push should reflect a better rendition of how I want all the code organized, so please feel free to take another look at this and any other issues you find. Thanks for your help and feedback.

ccordoba12:
Please rebase to get the fixes to our tests in PR #19564.

dalthviz:
@mrclary just in case I did a quick check installing the `.pkg` and seems like I'm getting an error about `applaunchservices` when launching the application that gets created at the user `Applications` directory (I think that is the one that the installer creates righ?):

<details>

```
  File "/Users/daniel/opt/anaconda3/lib/python3.9/site-packages/pkg_resources/__init__.py", line 579, in _build_master
    ws.require(__requires__)
  File "/Users/daniel/opt/anaconda3/lib/python3.9/site-packages/pkg_resources/__init__.py", line 897, in require
    needed = self.resolve(parse_requirements(requirements))
  File "/Users/daniel/opt/anaconda3/lib/python3.9/site-packages/pkg_resources/__init__.py", line 788, in resolve
    raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (applaunchservices 0.2.1 (/Users/daniel/opt/anaconda3/lib/python3.9/site-packages), Requirement.parse('applaunchservices>=0.3.0'), {'spyder'})

During handling of the above exception, another exception occurred:

  File "/Users/daniel/opt/anaconda3/bin/spyder", line 33, in <module>
    sys.exit(load_entry_point('spyder', 'gui_scripts', 'spyder')())
  File "/Users/daniel/opt/anaconda3/bin/spyder", line 25, in importlib_load_entry_point
    return next(matches).load()
  File "/Users/daniel/opt/anaconda3/lib/python3.9/importlib/metadata.py", line 86, in load
    module = import_module(match.group('module'))
  File "/Users/daniel/opt/anaconda3/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 850, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/Users/daniel/Desktop/spyder_otros/carlos/spyder/spyder/app/start.py", line 55, in <module>
    from spyder.utils.external import lockfile
  File "/Users/daniel/Desktop/spyder_otros/carlos/spyder/spyder/utils/external/lockfile.py", line 31, in <module>
    from spyder.utils.programs import is_spyder_process
  File "/Users/daniel/Desktop/spyder_otros/carlos/spyder/spyder/utils/programs.py", line 28, in <module>
    import pkg_resources
  File "/Users/daniel/opt/anaconda3/lib/python3.9/site-packages/pkg_resources/__init__.py", line 3247, in <module>
    def _initialize_master_working_set():
  File "/Users/daniel/opt/anaconda3/lib/python3.9/site-packages/pkg_resources/__init__.py", line 3221, in _call_aside
    f(*args, **kwargs)
  File "/Users/daniel/opt/anaconda3/lib/python3.9/site-packages/pkg_resources/__init__.py", line 3259, in _initialize_master_working_set
    working_set = WorkingSet._build_master()
  File "/Users/daniel/opt/anaconda3/lib/python3.9/site-packages/pkg_resources/__init__.py", line 581, in _build_master
    return cls._build_from_requirements(__requires__)
  File "/Users/daniel/opt/anaconda3/lib/python3.9/site-packages/pkg_resources/__init__.py", line 594, in _build_from_requirements
    dists = ws.resolve(reqs, Environment())
  File "/Users/daniel/opt/anaconda3/lib/python3.9/site-packages/pkg_resources/__init__.py", line 783, in resolve
    raise DistributionNotFound(req, requirers)
pkg_resources.DistributionNotFound: The 'applaunchservices>=0.3.0' distribution was not found and is required by spyder

```

</details>

However, if I use Anaconda Navigator from my conda installation, I can access the underline env (that I think the installer creates) and I'm able to launch Spyder:

imagen

Also, I did the installation having already the other MacOS standalone installer, is okay if a have both version installed at the same time?

mrclary:
The traceback suggests that spyder was launched from your anaconda installation. The package installer did indeed create the environment `spyder-5.3.4.dev98`; this should be located at `~/Library/spyder-5.3.4.dev98/envs/spyder-5.3.4.dev98` or `/Library/spyder-5.3.4.dev98/envs/spyder-5.3.4.dev98`. I think it's interesting that Anaconda Navigator found it.

> Also, I did the installation having already the other MacOS standalone installer, is okay if a have both version installed at the same time?

If they are in different locations, yes. If not, the new one may have clobbered or mangled the old one.

Can you double-check a few things for me?
- Where did you have the original Spyder.app? `~/Applications` or `/Applications`?
- When you ran the new installer, did you select "Install for me only" or "Install on a specific disk..."? The former should place the root environment in `~/Library` and the application bundle in `~/Applications`. The latter in `/Library` and `/Applications`, respectively.
- If the installer put the application bundle in the same location as your original Spyder.app, did it replace it? The original Spyder.app should be ~1GB, the new Spyder.app should be ~70kB.
- How did you launch the new application? If the installer put Spyder.app in a different location than your original, or if it completely clobbered the original Spyder.app, you should be able to launch it by double-clicking in the Finder. If it unsuccessfully clobbered the original, it may have mangled it; I'm not sure what would happen.

mrclary:
Okay, I was able to reproduce @dalthviz error on my machine.
Note that to start Spyder.app from the command line, some environment variables specified in the Info.plist need to be specified on the command line as well.
```
$ PREFIX=<PREFIX> ROOT_PREFIX=<ROOT_PREFIX> SPYDER_APP=<SPYDER_APP> ~/Applications/Spyder.app/Contents/MacOS/Spyder
```
This results in the following error:
```
/Users/rclary/Applications/Spyder.app/Contents/MacOS/Spyder: ?&?F?
??"??)@?Ć'?A??bmlocation has changed!
```
Clearly this is an issue with the pre-compiled binary executable.

In order to resolve the `PermissionError` issue (accessing user's Documents, Desktop, and Download directories), the application bundle should be a compiled binary rather than a shell script. I'm working with @jaimergp to resolve this (see napari/napari#5097). My ad-hoc solution was to compile the shell script using [`shc`](https://github.com/neurobin/shc). This works when compiled on my local machine, but clearly did not work for redistribution.

mrclary:
@dalthviz, was able to fix the issue from the previous comment. At your convenience, you may test the latest artifact: https://github.com/spyder-ide/spyder/suites/8618423726/artifacts/386781999

mrclary:
It appears that the post-install script did not run properly. This explains the missing environment variables in the Info.plist and all the other behavior that you see. I'm not sure exactly why it closes after launching, but probably because `SPYDER_APP` is not set. Nevertheless, the missing dependencies issue is directly related to the PermissionError, which is directly related to the launch script not being a binary. The launch script does not require the environment variables in Info.plist, but the binary executable does.

- Can you send me the install.log? Open Console.app and locate the install.log and attach it to a comment here.
  Screen Shot 2022-10-05 at 9 47 37 AM
- It looks like the environment was installed in `/Library` instead of `/Users/daniel/Library`. Can you confirm where the environment was installed?
- Can you check that the binary executable exists? `/Users/daniel/Library/spyder-5.3.4.dev118/envs/spyder-5.3.4.dev118/Menu/Spyder`. If it exists, and if you would like to try it, you may:
  - Copy it to `~/Applications/Spyder.app/Contents/MacOS/Spyder`, replacing the existing `Spyder` shell script
  - Update `~/Applications/Spyder.app/Contents/Info.plist` to read (just adding variables):
    ```
        <?xml version="1.0" encoding="UTF-8"?>
    <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
    <plist version="1.0">
    <dict>
        <key>CFBundleDisplayName</key>
        <string>Spyder</string>
        <key>CFBundleExecutable</key>
        <string>Spyder</string>
        <key>CFBundleGetInfoString</key>
        <string>Spyder-5.3.4.dev118</string>
        <key>CFBundleIconFile</key>
        <string>spyder.icns</string>
        <key>CFBundleIdentifier</key>
        <string>org.spyder-ide.Spyder</string>
        <key>CFBundleName</key>
        <string>Spyder</string>
        <key>CFBundlePackageType</key>
        <string>APPL</string>
        <key>CFBundleShortVersionString</key>
        <string>5.3.4.dev118</string>
        <key>CFBundleVersion</key>
        <string>5.3.4.dev118</string>
        <key>LSEnvironment</key>
        <dict>
            <key>PREFIX</key>
            <string>/Users/daniel/Library/spyder-5.3.4.dev118/envs/spyder-5.3.4.dev118</string>
            <key>ROOT_PREFIX</key>
            <string>/Users/daniel/Library/spyder-5.3.4.dev118</string>
            <key>SPYDER_APP</key>
            <string>/Users/daniel/Applications/Spyder.app</string>
        </dict>
    </dict>
    </plist>
    ```

mrclary:
@dalthviz, out of curiosity, does your macOS VM have `/usr/libexec/PlistBuddy`? I hope yes. Does it have Xcode command line tools installed? I hope not; this will help test for generic systems.

dalthviz:
>does your macOS VM have /usr/libexec/PlistBuddy?
I think so I can do `/usr/libexec/PlistBuddy` --help and get information about commands to use with it :+1:

> Does it have Xcode command line tools installed? 

I think they are installed. Runned `xcode-select -p` and it returned `/Library/Developer/CommandLineTools`

> Can you send me the install.log?

[install.txt](https://github.com/spyder-ide/spyder/files/9718613/install.txt)

> Can you confirm where the environment was installed?

Seems like it was installed at `/Library/spyder-5.3.4.dev118`. Thinking about that maybe that is way the installer prompted me to put the user password

> Can you check that the binary executable exists? /Users/daniel/Library/spyder-5.3.4.dev118/envs/spyder-5.3.4.dev118/Menu/Spyder

Checking the env created at `/Library` seems like it exists at `/Library/spyder-5.3.4.dev118/envs/spyder-5.3.4.dev118/Menu/`

mrclary:
@dalthviz, thanks for that info; it is very useful. Here's what I've discovered:
1. The package installer fails if the root environment already exists
1. The package installer does _not_ fail if the application bundle already exists. In this case the package installer does _not_ over-write it but the post-install script should still create the `python` symbolic link, replace the `Spyder` shell script, and modify the Info.plist.
1. When installing for all users, the `ROOT_PREFIX` and `PREFIX` environment variables have and extra "/"

In a future PR, I will have to address (1). I will need to address (2) and (3) in this PR.

So it seems that your issues are due to (2) and possibly (3). Note that your original application bundle, `/Applications/Spyder.app` is likely broken and will have to be reinstalled from the DMG.

My recommendation is to:
```
$ rm -rf /Library/spyder-5.3.4.dev118
$ rm -rf /Users/daniel/Library/spyder-5.3.4*  # if there are any envs installed here
$ rm -rf /Users/daniel/Applications/Spyder.app
```
and try the package installer again, ensuring to "Install for me only".

dalthviz:
Did the test ensuring the selection of `Install for me only` and seems like the installation is working (cheked launching from Finder, Spotlight and the Dock)! Now I'm not seeing `PermissionsError` related issues (I was prompted with the dialog to give the application access to manage files, which I think was missing before) so :+1:

Also, checked launching the old dmg installation I had and yes, is now broken. I though the default was to `Install for me only` but maybe I was wrong (or maybe I changed it without noticing the first time I launched the installer 🤔).

Could that be setup thought? Probably is better to have as default the  `Install for me only` option.

dalthviz (approved):
Thanks @mrclary ! This LGTM :+1:

Just in case @ccordoba12 or @CAM-Gerlach do you have any other comments or things to address here?

mrclary:
@ccordoba12 and @dalthviz, I've got both Linux and Windows installers working now also. See https://github.com/mrclary/spyder/actions/runs/3211090787.

I propose expanding the scope of this PR to include all installers, not just for macOS. All three installers will remain experimental, at least until 6.0, as previously discussed. This will also provide further opportunity for feedback and to address issues with the other installers. What do you think?

Note that the current artifacts are not code-signed. I've successfully tested the notarization process for macOS, but not for Windows.

ccordoba12:
> I propose expanding the scope of this PR to include all installers, not just for macOS. All three installers will remain experimental, at least until 6.0, as previously discussed.

I agree with adding the Linux installer, given that we don't have one at the moment. But I'm -1 on adding the Windows one due to the following reasons:

1. All improvements related to the the Numfocus grant are planned for the current installer, so there wouldn't be a real gain for users with the new one.
3. It'd introduce confusion among our Windows users because, even if the installer would be labeled as experimental, I'm sure people would download and try it, and ask what's the deal with it.
4. It's also important to note that this installer is now the most popular way to install Spyder on Windows, so we need to make the way users interact with it as simple as possible.

mrclary:
> > I propose expanding the scope of this PR to include all installers, not just for macOS. All three installers will remain experimental, at least until 6.0, as previously discussed.
> 
> I agree with adding the Linux installer, given that we don't have one at the moment. But I'm -1 on adding the Windows one due to the following reasons:
> 
> 1. All improvements related to the the Numfocus grant are planned for the current installer, so there wouldn't be a real gain for users with the new one.
> 2. It'd introduce confusion among our Windows users because, even if the installer would be labeled as experimental, I'm sure people would download and try it, and ask what's the deal with it.
> 3. It's also important to note that this installer is now the most popular way to install Spyder on Windows, so we need to make the way users interact with it as simple as possible.

Delaying the conda-based Windows installer should not be a problem. Since it is nominally working now, whenever we are ready to build it in the future, we will only have to add it to the matrix build in the workflow.

dalthviz:
@mrclary could it be possible to build the conda-based installers for all three platforms on PRs but in the release just build the MacOS and Linux one following @ccordoba12 considerations? That will give us some place to start checking the generated Windows installer even if it is not part of an actual release like 5.4.0

mrclary:
> @mrclary could it be possible to build the conda-based installers for all three platforms on PRs but in the release just build the MacOS and Linux one following @ccordoba12 considerations? That will give us some place to start checking the generated Windows installer even if it is not part of an actual release like 5.4.0

I don't see why not. Since they are all nominally working, simply adding/removing Windows from the workflow matrix should suffice. We could include _only_ the macOS and Linux installer in this PR, intended for 5.4.0 release. After 5.4.0 is released, simply re-add the Windows installer. For any subsequent release, we can decide to either keep the Windows installer or exclude it.

dalthviz:
Just in case, checked the Linux installer script and seems like is working :+1:

dalthviz:
Is there something missing here @mrclary @ccordoba12 ? Or is okay if I merge this?

installers-conda/build_conda_pkgs.py
```diff
@@ -0,0 +1,324 @@
...
+    QdarkstyleCondaPkg.name: QdarkstyleCondaPkg,
+    QtconsoleCondaPkg.name: QtconsoleCondaPkg,
+    SpyderKernelsCondaPkg.name: SpyderKernelsCondaPkg
+}
+
+if __name__ == "__main__":
+    p = ArgumentParser(
+        description=dedent(
+            """
+            Build conda packages from local spyder and external-deps sources.
```
ccordoba12, line 284:
```suggestion
            Build conda packages from local Spyder and external-deps sources.
```

installers-conda/build_conda_pkgs.py
```diff
@@ -0,0 +1,324 @@
...
+            Alternative git repo for python-lsp-server may be provided by
+            setting the environment variable PYTHON_LSP_SERVER_SOURCE,
+            otherwise the upstream remote will be used. All other external-deps
+            use the subrepo source within the spyder repo.
```
ccordoba12, line 288:
```suggestion
            use the subrepo source within the Spyder repo.
```

installers-conda/build_conda_pkgs.py
```diff
@@ -0,0 +1,324 @@
+"""
```
ccordoba12, line 1:
```suggestion
# -*- coding: utf-8 -*-
#
# Copyright © Spyder Project Contributors
# Licensed under the terms of the MIT License
# (see spyder/__init__.py for details)

"""
```

installers-conda/build_installers.py
```diff
@@ -0,0 +1,402 @@
+"""
```
ccordoba12, line 1:
```suggestion
# -*- coding: utf-8 -*-
#
# Copyright © Spyder Project Contributors
# Licensed under the terms of the MIT License
# (see spyder/__init__.py for details)

"""
```

ccordoba12:
@dalthviz, I'm working on my review right now.

installers-conda/build_installers.py
```diff
@@ -0,0 +1,402 @@
...
+from argparse import ArgumentParser
+from datetime import timedelta
+from distutils.spawn import find_executable
+from functools import partial
+from logging import getLogger
+from pathlib import Path
+from ruamel.yaml import YAML
+from subprocess import check_call
+from textwrap import dedent, indent
+from time import time
```
ccordoba12, line 34: Please organize these imports in alphabetical order and break them in two sections: `Standard library imports` and `Third-party imports`

installers-conda/build_installers.py
```diff
@@ -0,0 +1,402 @@
...
+
+from build_conda_pkgs import HERE, DIST, RESOURCES, SPECS, h, get_version
```
ccordoba12, line 36:
```suggestion
# Local imports
from build_conda_pkgs import HERE, DIST, RESOURCES, SPECS, h, get_version
```

installers-conda/build_installers.py
```diff
@@ -0,0 +1,402 @@
...
+logger.addHandler(h)
+logger.setLevel('INFO')
+
+APP = "Spyder"
+SPYREPO = HERE.parent
+WINDOWS = os.name == "nt"
+MACOS = sys.platform == "darwin"
+LINUX = sys.platform.startswith("linux")
+TARGET_PLATFORM = os.environ.get("CONSTRUCTOR_TARGET_PLATFORM")
+if TARGET_PLATFORM == "osx-arm64":
```
ccordoba12, line 48:
```suggestion

if TARGET_PLATFORM == "osx-arm64":
```

installers-conda/build_installers.py
```diff
@@ -0,0 +1,402 @@
...
+logger = getLogger('BuildInstallers')
+logger.addHandler(h)
+logger.setLevel('INFO')
+
+APP = "Spyder"
+SPYREPO = HERE.parent
+WINDOWS = os.name == "nt"
+MACOS = sys.platform == "darwin"
+LINUX = sys.platform.startswith("linux")
+TARGET_PLATFORM = os.environ.get("CONSTRUCTOR_TARGET_PLATFORM")
```
ccordoba12, line 47:
```suggestion
TARGET_PLATFORM = os.environ.get("CONSTRUCTOR_TARGET_PLATFORM")
PY_VER = f"{sys.version_info.major}.{sys.version_info.minor}"
```

Moving to the section of constants and removing it below.

installers-conda/build_installers.py
```diff
@@ -0,0 +1,402 @@
...
+if TARGET_PLATFORM == "osx-arm64":
+    ARCH = "arm64"
+else:
+    ARCH = (platform.machine() or "generic").lower().replace("amd64", "x86_64")
+PY_VER = f"{sys.version_info.major}.{sys.version_info.minor}"
```
ccordoba12, line 52:
```suggestion

```

installers-conda/build_installers.py
```diff
@@ -0,0 +1,402 @@
...
+indent4 = partial(indent, prefix="    ")
+
+SPYVER = get_version(SPYREPO).strip().split("+")[0]
+
+specs = {
+    "spyder": "=" + SPYVER,
+    "paramiko": "",
+    "pyxdg": "",
+}
+if SPECS.exists():
```
ccordoba12, line 128:
```suggestion

if SPECS.exists():
```

installers-conda/build_installers.py
```diff
@@ -0,0 +1,402 @@
...
+    specs[k] = "".join(v).strip()
+    if k == "spyder":
+        SPYVER = v[-1]
+
+OUTPUT_FILE = DIST / f"EXPERIMENTAL-{APP}-{SPYVER}-{OS}-{ARCH}.{EXT}"
+INSTALLER_DEFAULT_PATH_STEM = f"{APP}-{SPYVER}"
+
+
+def _generate_background_images(installer_type):
+    """Requires pillow"""
```
ccordoba12, line 149:
```suggestion
    """This requires Pillow."""
```

installers-conda/build_installers.py
```diff
@@ -0,0 +1,402 @@
...
+
+
+def _get_condarc():
+    # we need defaults for tensorflow and others on windows only
+    defaults = "- defaults" if WINDOWS else ""
+    prompt = "[spyder]({default_env}) "
+    contents = dedent(
+        f"""
+        channels:  #!final
+          - spyder-ide
```
ccordoba12, line 186: Do we really need to have our channel here? I think there are no important packages there.

installers-conda/build_installers.py
```diff
@@ -0,0 +1,402 @@
...
+def _definitions():
+    condarc = _get_condarc()
+    definitions = {
+        "name": APP,
+        "company": "Spyder-IDE",
+        "reverse_domain_identifier": "org.spyder-ide.Spyder",
+        "version": SPYVER,
+        "channels": [
+            "napari/label/bundle_tools",
+            "spyder-ide",
```
ccordoba12, line 214: Again, I think we can remove our channel from here.

installers-conda/build_installers.py
```diff
@@ -0,0 +1,402 @@
...
+        },
+        "menu_packages": [
+            "spyder",
+        ],
+        "extra_files": {
+            str(RESOURCES / "bundle_readme.md"): "README.txt",
+            condarc: ".condarc",
+        },
+    }
+    if not args.no_local:
```
ccordoba12, line 240:
```suggestion

    if not args.no_local:
```

installers-conda/build_installers.py
```diff
@@ -0,0 +1,402 @@
...
+        definitions["channels"].insert(0, "local")
+
+    if LINUX:
+        definitions["default_prefix"] = os.path.join(
+            "$HOME", ".local", INSTALLER_DEFAULT_PATH_STEM
```
ccordoba12, line 245: I think we should validate that `~/.local` exists at this point and if it doesn't, then we should create it. We can't take that for granted given the immense variety of Linux distros.

installers-conda/build_installers.py
```diff
@@ -0,0 +1,402 @@
...
+                "default_location_pkg": "Library",
+                "installer_type": "pkg",
+                "welcome_image": str(DIST / "spyder_1227x600.png"),
+                "welcome_file": str(welcome_file),
+                "conclusion_text": "",
+                "readme_text": "",
+                "post_install": str(RESOURCES / "post-install.sh"),
+            }
+        )
+        if args.cert_id:
```
ccordoba12, line 271:
```suggestion

        if args.cert_id:
```

installers-conda/build_installers.py
```diff
@@ -0,0 +1,402 @@
...
+                    "%LOCALAPPDATA%", INSTALLER_DEFAULT_PATH_STEM
+                ),
+                "default_prefix_all_users": os.path.join(
+                    "%ALLUSERSPROFILE%", INSTALLER_DEFAULT_PATH_STEM
+                ),
+                "check_path_length": False,
+                "installer_type": "exe",
+            }
+        )
+        signing_certificate = os.environ.get("CONSTRUCTOR_SIGNING_CERTIFICATE")
```
ccordoba12, line 296:
```suggestion

        signing_certificate = os.environ.get("CONSTRUCTOR_SIGNING_CERTIFICATE")
```

installers-conda/build_installers.py
```diff
@@ -0,0 +1,402 @@
...
+
+    Parameters
+    ----------
+    version: str
+        Version of `spyder` to be built. Defaults to the
+        one detected by `importlib` from the source code.
+    extra_specs: list of str
+        Additional packages to be included in the installer.
+        A list of conda spec strings (`numpy`, `python=3`, etc)
+        is expected.
```
ccordoba12, line 319:
```suggestion
```

This function has no parameters.

mrclary:
@CAM-Gerlach, do you want to review this PR?

installers-conda/resources/bundle_readme.md
```diff
@@ -0,0 +1,56 @@
...
+This is the base installation of Spyder, the Scientific Python Development Environment.
+
+## How do I run Spyder?
+
+In most cases, you would run it through the platform-specific shortcut we created for your
+convenience. In other words, _not_ through this directory!
+
+* Linux: check your desktop launcher.
+* MacOS: check `~/Applications` or the Launchpad.
+* Windows: check the Start Menu or the Desktop.
```
ccordoba12, line 13:
```suggestion
* Linux: Check your desktop launcher.
* MacOS: Check `~/Applications` or the Launchpad.
* Windows: Check the Start Menu or the Desktop.
```

installers-conda/resources/bundle_readme.md
```diff
@@ -0,0 +1,56 @@
...
+
+* Linux and macOS: find it under `bin`, next to this file.
+* Windows: navigate to `Scripts`, next to this file.
+
+In unmodified installations, this _should_ be enough to launch `spyder`, but sometimes you will
+need to activate the `conda` environment to ensure all dependencies are importable.
+
+## What does `conda` have to do with `spyder`?
+
+The Spyder installer uses `conda` packages to bundle all its dependencies (Python, qt, etc).
```
ccordoba12, line 26:
```suggestion
The Spyder installer uses `conda` packages to bundle all its dependencies (Python, Qt, etc).
```

installers-conda/resources/bundle_readme.md
```diff
@@ -0,0 +1,56 @@
...
+This directory is actually a full `conda` installation! If you have used `conda` before, this
+is equivalent to what you usually call the `base` environment.
+
+## Can I modify the `spyder` installation?
+
+Yes. In practice, you can consider it a `conda` environment. You can even activate it as usual,
```
ccordoba12, line 32:
```suggestion
Yes, but it is not recommended (see below). In practice, you can consider it a `conda` environment. You can even activate it as usual,
```

installers-conda/build_installers.py
mrclary: I believe this is done by the installer at install time on the local machine. This `default_prefix` specifies where, on the local machine, the environment should be installed, but this script `build_installers.py` is not run on the local machine. So checking for `~.local` here will only check if it is present on the build machine.

mrclary:
> @mrclary, one additional question for you: in the description above you said that this _partially_ fixes issue #18870, I think it actually fixes it in its entirety, right?

Correct

installers-conda/build_installers.py
ccordoba12: Ok, I see. No problem with that then.

ccordoba12 (approved):
Thanks @mrclary for all your hard work to add these new, conda-based installers for Spyder! It's a great improvement!

jaimergp:
Yay! 🚀

dalthviz:
Thank you so much @jaimergp for all the info and help you provided us to move forward this :)

PR #19549, opened 2022-09-22
PR: Fix errors when Projects plugin is disabled

rear1019:
## Description of Changes
Fix two issues when Projects plugin is disabled:
* Fix re-opening of files from previous session when Projects plugin is disabled. Added test.
* Don’t add empty Projects menu when Projects plugin is disabled

### Affirmation

rear1019

### DIFF
--- a/spyder/plugins/editor/tests/conftest.py
+++ b/spyder/plugins/editor/tests/conftest.py
@@ -129,8 +129,10 @@ def editor_plugin_open_files(request, editor_plugin, python_files):
 
         def get_option(option, default=None):
             return options_dict.get(option)
+
         def set_option(option, value):
             options_dict[option] = value
+
         editor.get_option = get_option
         editor.set_option = set_option
 
--- a/spyder/plugins/editor/tests/conftest.py
+++ b/spyder/plugins/editor/tests/conftest.py
@@ -129,7 +129,10 @@ def editor_plugin_open_files(request, editor_plugin, python_files):
 
         def get_option(option, default=None):
             return options_dict.get(option)
+        def set_option(option, value):
+            options_dict[option] = value
         editor.get_option = get_option
+        editor.set_option = set_option
 
         editor.setup_open_files()
         return editor, expected_filenames, expected_current_filename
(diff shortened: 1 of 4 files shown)


spyder/plugins/editor/tests/conftest.py
```diff
@@ -129,7 +129,10 @@ def _get_editor_open_files(last_focused_filename,
 
         def get_option(option, default=None):
             return options_dict.get(option)
+        def set_option(option, value):
+            options_dict[option] = value
```
ccordoba12, line 133:
```suggestion

        def set_option(option, value):
            options_dict[option] = value

```

spyder/plugins/mainmenu/plugin.py
```diff
@@ -70,7 +70,8 @@ def on_initialize(self):
         create_app_menu(ApplicationMenus.Run, _("&Run"), dynamic=False)
         create_app_menu(ApplicationMenus.Debug, _("&Debug"), dynamic=False)
         create_app_menu(ApplicationMenus.Consoles, _("C&onsoles"))
-        create_app_menu(ApplicationMenus.Projects, _("&Projects"))
+        if self.get_conf("enable", section=Plugins.Projects):
```
ccordoba12, line 73:
```suggestion
        if self.get_plugin(Plugins.Projects, error=False):
```

This is how we detect if a plugin is active or not.

ccordoba12 (commented):
Thanks @rear1019 for your help with this! Two small suggestions for you, otherwise looks good to me.

Also, please rebase or merge with 5.x to get the fix to our tests in PR #19564.

spyder/plugins/mainmenu/plugin.py
rear1019: Using `get_plugin()` fails with `SpyderAPIError: Plugin "project_explorer" not part of REQUIRED or OPTIONAL requirements!`. I have opted for using `is_plugin_enabled()` which works, and fits even better.

rear1019:
Rebased and applied suggestions.

spyder/plugins/mainmenu/plugin.py
ccordoba12: > Using get_plugin() fails with SpyderAPIError: Plugin "project_explorer" not part of REQUIRED or OPTIONAL requirements!

You're right, I forgot that in order to use `get_plugin` you need to declare it as a dependency.

> I have opted for using is_plugin_enabled() which works, and fits even better

Ok, that's a great idea! Thanks a lot for looking into it.

ccordoba12 (approved):
Thanks @rear1019!

PR #19646, opened 2022-09-28
PR: Show Spyder-kernels message on restarts (IPython console)

ccordoba12:
## Description of Changes

- If users change their interpreter in `Preferences > Main Interpreter` to one without Spyder-kernels or the right version of it, we were showing the following after a restart:

     image

    Now we show this

    image

- Disable kernel related actions when it fails to start because it makes no sense to have them enabled.

    image

- Add a test that checks this.

### Affirmation


ccordoba12:
Note: I'll take care of merging this to `master` because it requires some different changes.

PR #19742, opened 2022-10-07
PR: Reinstall spyder in editable mode in bootstrap if branch changed to/from master

### LINKED ISSUE #19728: Traceback from Editor/Debugger plugin when using same environment to launch different Spyder branches (master vs 5.x)
## Problem Description

When launching from master (previously having launched from 5.x on the same env) I got a traceback on the Anaconda prompt regarding the Debugger actions from the Editor. Maybe a validation is missing there or something is missing on the `bootstrap.py` script when changing branches? 

1.  Launch Spyder from the master branch on an env that previously was used to launch the 5.x branch

No traceback

```python-traceback
  File "e:\acer\documentos\spyder\spyder\spyder\plugins\editor\plugin.py", line 1585, in setup_other_windows
    action = self.main.debugger.get_action(action_name)
AttributeError: 'NoneType' object has no attribute 'get_action'
```

## Versions

* Spyder version: 6.0.0.dev0 02aeb4d82 (conda)
* Python version: 3.10.5 64-bit
* Qt version: 5.15.3
* PyQt5 version: 5.15.4

mrclary:
## Description of Changes

When launching from bootstrap, detects whether the root branch has changed to/from master, in which case spyder is reinstalled in editable mode.

### Issue(s) Resolved

Fixes #19728

### Affirmation

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -28,8 +28,6 @@ import time
 from logging import Formatter, StreamHandler, getLogger
 from pathlib import Path
 
-from install_dev_repos import REPOS, install_repo
-
 # ---- Setup logger
 fmt = Formatter('%(asctime)s [%(levelname)s] [%(name)s] -> %(message)s')
 h = StreamHandler()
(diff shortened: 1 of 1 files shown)


bootstrap.py
```diff
@@ -74,14 +72,35 @@
...
+    boot_branch_file = Path(get_conf_path("boot_branch.txt"))
+    if boot_branch_file.exists():
+        prev_branch = boot_branch_file.read_text()
+
+    revision, branch = vcs.get_git_revision(Path(__file__).parent)
+    boot_branch_file.write_text(branch)
+
+    logger.info("Previous branch: %s; current branch: %s", prev_branch, branch)
+
+    if "master" in (branch, prev_branch) and branch != prev_branch:
```
dalthviz, line 92: Maybe the Spyder editable reinstall should be forced even if the branch is not `master`. Thinking about it is possible that a branch is based on `master` or `5.x` without being exactly `master` or `5.x` right?

mrclary: I suppose what matters is whether the parent commit has changed to/from the master branch. Pushed up a change.

bootstrap.py
```diff
@@ -74,14 +72,39 @@
...
+    ).communicate()
+    branch = "master" if res else "not master"
+    boot_branch_file.write_text(branch)
+
+    logger.info("Previous root branch: %s; current root branch: %s",
+                prev_branch, branch)
+
+    if branch != prev_branch:
+        logger.info("Detected root branch change to/from master. "
+                    "Will reinstall spyder in editable mode.")
```
ccordoba12, line 98:
```suggestion
                    "Will reinstall Spyder in editable mode.")
```

ccordoba12 (commented):
Thanks @mrclary for your prompt response to address this issue.

bootstrap.py
```diff
@@ -74,14 +72,39 @@
 
 # ---- Install sub repos
 
+from install_dev_repos import REPOS, install_repo
+from spyder import get_versions
+from spyder.config.base import get_conf_path
+from spyder.utils.programs import find_git, run_program
```
ccordoba12, line 78: Is it not possible to place these imports at the top of the file? If there's no technical reason that prevents it, I think they should be left there to follow our regular convention for imports.

ccordoba12 (commented):
@mrclary, one more small thing I just noticed, then this should be ready on my side.

ccordoba12 (approved):
Thanks @mrclary!

dalthviz (approved):
Thanks @mrclary !

PR #19750, opened 2022-10-09
PR: Increase minimal required version of Qstylizer to 0.2.2

ccordoba12:
## Description of Changes

That version has a new `alignment` value that allows us to use it to declare several styles instead of relying on strings, which are more difficult to read and change.

### Affirmation


ccordoba12:
> However, since it changes the version from qstylizer, maybe we should merge it just before doing the release (5.4.0) to prevent generating problems when running Spyder on dev mode for others?

Ok, I have no problem with that.

> Just in case, with qstylizer 0.2.1 I get the following traceback

Right, that's because I added the `alignment` property some weeks ago, which is now part of Qstylizer 0.2.2.

dalthviz (approved):
Thanks @ccordoba12 !

PR #19793, opened 2022-10-11
PR: Show installer download percentage progress in the status bar (Application)

### LINKED ISSUE #19112: New update status bar default message, visibility behavior and update process for the Windows standalone installer
As part of #18619 a new status bar to handle the Spyder update state was proposed (initially for the Windows standalone installer). Talking with @ccordoba12 some discussion regarding what the default message should be (currently the Spyder version installed: imagen) and if the status bar should only be shown for the update process (to show for example only if an update is available, is the update is being downloaded or installed, or if ti was cancelled) and hide it otherwise (instead of showing the status bar always) raised.

Currently at [#18619](https://github.com/spyder-ide/spyder/pull/18619/commits/98ec631729e94bdd3cc1f2f628eab4f9b94e0c38) the status bar always is visible and it shows these possible states:

* Spyder current version (for example 5.3.1): imagen
* Checking for updates: imagen

* An update is pending: imagen

* Downloading the new version installer: imagen

* Installing new version: imagen

Thinking about it probably the messages for the different states could be improved too.

So to discuss and make a decision without having to extend further the comments at the PR this issue is created.

Pinging @CAM-Gerlach since he could contribute to the discussion

dalthviz:
## Description of Changes

A preview:

installer_download

### Issue(s) Resolved

Fixes #19112

### Affirmation

### DIFF
--- a/spyder/plugins/application/widgets/status.py
+++ b/spyder/plugins/application/widgets/status.py
@@ -14,7 +14,7 @@ import os
 
 # Third party imports
 from qtpy.QtCore import QPoint, Signal, Slot
-from qtpy.QtWidgets import QMenu
+from qtpy.QtWidgets import QMenu, QLabel
 
 # Local imports
 from spyder.api.translations import get_translation
@@ -53,6 +53,8 @@ class ApplicationUpdateStatus(StatusBarWidget):
         Path to instal
     """
 
+    CUSTOM_WIDGET_CLASS = QLabel
+
     def __init__(self, parent):
 
         self.tooltip = self.BASE_TOOLTIP
@@ -65,6 +67,8 @@ class ApplicationUpdateStatus(StatusBarWidget):
 
         self.sig_clicked.connect(self.show_installation_dialog_or_menu)
 
+        self.installer.sig_download_progress.connect(
+            self.set_download_progress)
         self.installer.sig_installation_status.connect(
             self.set_value)
         self.installer.sig_install_on_close_requested.connect(
@@ -77,16 +81,24 @@ class ApplicationUpdateStatus(StatusBarWidget):
                              "background.\n"
                              "Click here to show the installation "
                              "dialog again.")
-            self.spinner.show()
-            self.spinner.start()
+            if value == DOWNLOADING_INSTALLER:
+                self.spinner.hide()
+                self.spinner.stop()
+                self.custom_widget.show()
+            else:
+                self.custom_widget.hide()
+                self.spinner.show()
+                self.spinner.start()
             self.installer.show()
-
         elif value == PENDING:
             self.tooltip = value
+            self.custom_widget.hide()
             self.spinner.hide()
             self.spinner.stop()
         else:
             self.tooltip = self.BASE_TOOLTIP
+            if self.custom_widget:
+                self.custom_widget.hide()
             if self.spinner:
                 self.spinner.hide()
                 self.spinner.stop()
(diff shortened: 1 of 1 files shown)


spyder/plugins/application/widgets/status.py
```diff
@@ -106,11 +118,15 @@ def get_icon(self):
     def start_installation(self, latest_release):
         self.installer.start_installation(latest_release)
 
+    def set_download_progress(self, current_value, total):
+        percentage_progress = 0
+        if total > 0:
+            percentage_progress = int(current_value/total*100)
```
ccordoba12, line 124:
```suggestion
            percentage_progress = int((current_value/total) * 100)
```

ccordoba12 (commented):
Small suggestion for you @dalthviz, otherwise looks good to me.

ccordoba12:
@dalthviz, could you post a screenshot in the description above to see how this looks? Thanks!

ccordoba12 (approved):
Thanks @dalthviz! This looks pretty cool!

spyder/plugins/application/widgets/status.py
```diff
@@ -106,11 +118,15 @@ def get_icon(self):
     def start_installation(self, latest_release):
         self.installer.start_installation(latest_release)
 
+    def set_download_progress(self, current_value, total):
+        percentage_progress = 0
+        if total > 0:
+            percentage_progress = int((current_value/total) * 100)
```
CAM-Gerlach, line 124:
```suggestion
            percentage_progress = round((current_value/total) * 100)
```

Perhaps slightly more idiomatic (and perhaps a tad more representative, since it rounds rather than truncates)

spyder/plugins/application/widgets/status.py
```diff
@@ -106,11 +118,15 @@ def get_icon(self):
...
+        self.custom_widget.setText(f"{percentage_progress} %")
```
CAM-Gerlach, line 125:
```suggestion
        self.custom_widget.setText(f"{percentage_progress}%")
```

No space between number and percent sign (at least in English)

CAM-Gerlach (commented):
Looking pretty good, thanks! A couple small tweaks, and two other visual issues:

* [ ] The horizontal spacing between the percentage number and the text in the status bar seems a little bit too much; could you reduce it by a 1/3 to 1/2?
* [ ] Maybe out of scope, but the overall text seems vertically misaligned with the other status bar items and its own icon, looking lower by at least a few pixels. Maybe out of scope here, but any way to align it properly?

dalthviz:
A preview with the latest changes: 

imagen

Checking the alignment more in detail I think that the further difference in alignment comes from the interpreter status not having an icon. If I put an icon to it I got this:

imagen

Should I add the icon to the interpreter status bar @CAM-Gerlach @ccordoba12  ?

CAM-Gerlach:
> A preview with the latest changes:

Thanks. The space between the percent number and the text to its left still seems a bit much; it could probably still be cut in half if that's possible

image

> Should I add the icon to the interpreter status bar @CAM-Gerlach @ccordoba12 ?

:+1: from me, it looks a lot better, though seems like best to do in a separate PR

ccordoba12:
> Thanks. The space between the percent number and the text to its left still seems a bit much; it could probably still be cut in half if that's possible

This is probably a Qt layout/css issue, which is usually not easy to fix. I'm fine to merge this one as it is, unless @dalthviz is willing to invest more time to find a solution for it.

> +1 from me, it looks a lot better, though seems like best to do in a separate PR

I agree. In addition, I think that shouldn't be a blocker for the 5.4.0 release.

dalthviz:
I have been checking and it could be related with the QLabel padding but it will need further research so sound goods to me to merge this as it is.

Regarding the icon for the interpreter status bar, should I create another PR for that, right?

ccordoba12:
> Regarding the icon for the interpreter status bar, should I create another PR for that, right?

I'd like to discuss first what we should do about the two status widgets that show almost the same interpreter information.

CAM-Gerlach (approved):
LGTM for now from a UX perspective, then, given those two follow ups

PR #19838, opened 2022-10-14
PR: Do not import spyder before install_repo in bootstrap.py

### LINKED ISSUE #19832: Spyder doesn't open on MyBinder
### Issue Report Checklist

* [x] Completed the **Problem Description** & **Steps to Reproduce**

## Problem Description

Spyder never opens in MyBinder sessions launched from this repo presently. 

When the sessions opens and you double click on the Spyder icon on the left side of the screen it doesn't open like it should. Splash screen never comes up and there is no indication though. You just wait forever for nothing, it seems.

1.  Try opening Spyder in sessions launched from https://mybinder.org/v2/gh/spyder-ide/spyder/master?urlpath=%2Fdesktop (which is the default link on the main page and uses the last commit ([Oct 11](https://github.com/spyder-ide/spyder/commits/master/binder)) by default.). Spyder FAILS to open in the session. It just hangs forever with the splash screen never coming up. No error is shown.

3. If you try launching with the previous commit, Spyder still opens by double-clicking on the icon in such sessions. The splash screen comes up after several seconds and Spyder opens. STILL WORKS using Aug 29th version at https://github.com/spyder-ide/spyder/commit/a3f49fa9ba51f40e6fe1e8d52768265b1da2f21f. The link to launch is:
https://mybinder.org/v2/gh/spyder-ide/spyder/a3f49fa9ba51f40e6fe1e8d52768265b1da2f21f?urlpath=%2Fdesktop

### Underlying cause?

It seems like the change was small in the binder directory, and so there is a chance it is something else **and not simply what was changed there**? I'm not quite sure. I only tried the October 11th commit (actually did it specifically, as well, and not just default) and the August 29th ones because that was only changes showing when I clicking into the binder directory.     

I did note someone else did make an issue post a few days ago saying they were seeing Spyder not opening, but that seemed to get sorted.

mrclary:
### Issue(s) Resolved

Fixes #19832

### Affirmation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,7 @@ spyder_crash.log
 .cache
 .coverage*
 MANIFEST
+boot_branch.txt
 
 # git .orig files
 *.orig
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -28,11 +28,7 @@ import time
 from logging import Formatter, StreamHandler, getLogger
 from pathlib import Path
 
-# Local imports
-from install_dev_repos import REPOS, install_repo
-from spyder import get_versions
-from spyder.config.base import get_conf_path
-from spyder.utils.programs import find_git, run_program
+from install_dev_repos import DEVPATH, REPOS, install_repo
 
 # ---- Setup logger
 fmt = Formatter('%(asctime)s [%(levelname)s] [%(name)s] -> %(message)s')
@@ -80,14 +76,14 @@ assert args.gui in (None, 'pyqt5', 'pyside2'), \
 installed_dev_repo = False
 if not args.no_install:
     prev_branch = None
-    boot_branch_file = Path(get_conf_path("boot_branch.txt"))
+    boot_branch_file = DEVPATH / "boot_branch.txt"
     if boot_branch_file.exists():
         prev_branch = boot_branch_file.read_text()
 
-    result, error = run_program(
-        find_git(), ['merge-base', '--fork-point', 'master']
-    ).communicate()
-    branch = "master" if result else "not master"
+    result = subprocess.run(
+        ["/usr/bin/env", "git", "merge-base", "--fork-point", "master"]
+    )
+    branch = "master" if result.stdout else "not master"
     boot_branch_file.write_text(branch)
 
     logger.info("Previous root branch: %s; current root branch: %s",
@@ -96,7 +92,7 @@ if not args.no_install:
     if branch != prev_branch:
         logger.info("Detected root branch change to/from master. "
                     "Will reinstall Spyder in editable mode.")
-        REPOS["spyder"]["editable"] = False
+        REPOS[DEVPATH.name]["editable"] = False
 
     for name in REPOS.keys():
         if not REPOS[name]['editable']:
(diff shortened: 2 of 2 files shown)


mrclary:
@ccordoba12 @dalthviz, Spyder now launches in Binder, but there are still a few problems. The following missing dependencies are, indeed, missing from the conda environment, but I'm not sure why.
Screen Shot 2022-10-14 at 5 28 31 PM

mrclary:
Okay, it seems that Binder behaves differently when built from the 5.x branch vs. the master branch. I'm not sure exactly why that is.

The binder link above, in this PR, on 5.x, results in the repository copied to a "spyder" directory in the home directory and missing dependencies in the environment. The shortcut on the desktop is labeled with 5.x.

I cherry-picked this PR's changes onto the master branch in my fork and built the binder from there and found that the repository is copied to the home directory and all dependencies are installed and Spyder launches and runs fine. The shortcut on the desktop is labeled with 4.x.

So I suppose everything is okay if run from the master branch. But it seems that some idiosyncrasies should be cleaned up at some point.

ccordoba12:
> Spyder now launches in Binder, but there are still a few problems. The following missing dependencies are not installed in the environment and I'm not sure why.

@mrclary, we've talked about this before but probably you forgot it. So, we create Binder links for PRs through an associated repo to avoid rebuilding the Binder instance on every PR commit, which is very inefficient:

https://github.com/spyder-ide/binder-environments

However, I haven't had time to update it in a long time, hence the missing dependencies.

I propose to do the following to improve the situation regarding Binder:

* Remove the `binder` folder in this repo to use the one in `binder-environments` instead (there's a branch on it per branch here).
* Update the `5.x`, `master` and `spyder-stable` dependencies in `binder-environments` with the latest changes here. Right now, those branches are separate, but you can push directly to them and then check that the Binder image is built as expected by clicking on the `launch | binder` button of the Readme associated to each branch.
* Create a subrepo for `binder-environments` in `external-deps` and sync it with the corresponding branch here.
* Change `test_dependencies_for_binder_in_sync` to do use `external-deps/binder-environments/binder/environment.yml` instead.
* Add a section in CONTRIBUTING.md explaining how to add/update a new dependency in `binder-environments`. Perhaps we should start merging from `5.x` to `master` on that repo so that contributors don't have to open two PRs there.
* Remove the `launch | binder` button in our Readme and add a new section instead called `Run Spyder in your browser`. That should contain a small explanation of what Binder is and three buttons (you can create custom Binder buttons [here](https://mybinder.readthedocs.io/en/latest/howto/badges.html)):
    * `launch | stable version`: That will use the link of the Binder button in https://github.com/spyder-ide/binder-environments/tree/spyder-stable.
    * `launch | 5.x`: It'll use the link in https://github.com/spyder-ide/binder-environments/tree/5.x
    * `launch | master`: It'll use the link in https://github.com/spyder-ide/binder-environments/tree/master
* Add a new instruction in RELEASE.md so that our release manager don't forget to update the dependencies of the `spyder-stable` branch in `binder-environments`.

This way we'd make sure that `binder-environments` is always up to date and it could be used reliably by users.

@mrclary, I think this would be a cool project for you to take on. Besides, it'd be great if you could work on it because I'm really busy with other things.

ccordoba12 (approved):
Thanks @mrclary! I'm going to merge this because it also fixes an error where the Internal console is hidden at startup on DEV mode, which I use lot when fixing things in Spyder.

PR #19867, opened 2022-10-17
PR: Fix `bootstrap.py` git subprocess call on Windows

### LINKED ISSUE #19865: Unable to launch Spyder from `bootstrap.py` on Windows
## Problem Description

A traceback appears when launching Spyder from `bootstrap.py`. This happened to me will try to check locally https://github.com/spyder-ide/spyder/pull/19855

Seems like is related with the subprocess that calls git to check the branch info. ~I guess on Windows instead of `"/usr/bin/env"` we need to use `"cmd.exe"` or something similar?~

Edit:

Just in case, seems like on Windows something like the following works:

```python
result = subprocess.run(
        ["git.cmd", "merge-base", "--fork-point", "master"], shell=True
    )
```

Try to run Spyder from the `bootstrap.py` script on Windows

Spyder launching, instead a traceback

```python-traceback

  File "E:\Acer\Documentos\Spyder\Spyder otros\carlos\spyder\bootstrap.py", line 83, in <module>
    result = subprocess.run(
  File "C:\Users\dalth\anaconda3\envs\spyder-dev\lib\subprocess.py", line 501, in run
    with Popen(*popenargs, **kwargs) as process:
  File "C:\Users\dalth\anaconda3\envs\spyder-dev\lib\subprocess.py", line 969, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "C:\Users\dalth\anaconda3\envs\spyder-dev\lib\subprocess.py", line 1438, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
FileNotFoundError: [WinError 2] El sistema no puede encontrar el archivo especificado


```

dalthviz:
## Description of Changes

Missing Windows case to call git from subprocess from `bootstrap.py`

### Issue(s) Resolved

Fixes #19865 

### Affirmation

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -80,14 +80,9 @@ if not args.no_install:
     if boot_branch_file.exists():
         prev_branch = boot_branch_file.read_text()
 
-    if os.name == "nt":
-        result = subprocess.run(
-            ["git.cmd", "merge-base", "--fork-point", "master"], shell=True
-        )
-    else:
-        result = subprocess.run(
-            ["/usr/bin/env", "git", "merge-base", "--fork-point", "master"]
-        )
+    result = subprocess.run(
+        ["git", "merge-base", "--fork-point", "master"]
+    )
     branch = "master" if result.stdout else "not master"
     boot_branch_file.write_text(branch)
 
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -80,9 +80,14 @@ if not args.no_install:
     if boot_branch_file.exists():
         prev_branch = boot_branch_file.read_text()
 
-    result = subprocess.run(
-        ["/usr/bin/env", "git", "merge-base", "--fork-point", "master"]
-    )
+    if os.name == "nt":
+        result = subprocess.run(
+            ["git.cmd", "merge-base", "--fork-point", "master"], shell=True
+        )
+    else:
+        result = subprocess.run(
+            ["/usr/bin/env", "git", "merge-base", "--fork-point", "master"]
+        )
     branch = "master" if result.stdout else "not master"
     boot_branch_file.write_text(branch)
 


ccordoba12 (approved):

bootstrap.py
```diff
@@ -80,9 +80,14 @@
...
-        ["/usr/bin/env", "git", "merge-base", "--fork-point", "master"]
-    )
+    if os.name == "nt":
+        result = subprocess.run(
+            ["git.cmd", "merge-base", "--fork-point", "master"], shell=True
+        )
+    else:
+        result = subprocess.run(
+            ["/usr/bin/env", "git", "merge-base", "--fork-point", "master"]
+        )
```
mrclary: I get the following error on my Windows VM:
```
'git.cmd' is not recognized as an internal or external command,
operable program or batch file.
```
However, the following does work for me on both macOS and my Windows VM:
```suggestion

    result = subprocess.run(
        ["git", "merge-base", "--fork-point", "master"]
    )
```
I think that the use of `shell=True` is discouraged, as well.

Previously, I used `spyder.utils.programs.find_git` and `spyder.utils.programs.run_program`, which did not use `shell=True`. However, this required that `spyder` be installed, which may not be the case at this point in `bootstrap.py`. 

I'd think the most elegant solution would be to import `spyder.utils.programs` directly using `importlib` tools, except that the module imports from other `spyder` modules as well.

So we may need to duplicate `find_git` here in order to get the correct `git` command on any given system. Furthermore, `find_git` also checked that `git` was properly installed for macOS.

Do you or @ccordoba12 have any better ideas?

mrclary:
@dalthviz @ccordoba12, will `subprocess.run(["git", ...])` find `git.cmd` as well as `git.exe` on Windows?

I know that `spyder.utils.programs.find_git` has a lot of robust checks, but maybe we don't need to worry about that for `bootstrap.py` since this is primarily for developers who should have `git` installed properly on their system. So maybe `"git"` will work well enough for all platforms in this situation?

ccordoba12:
> will subprocess.run(["git", ...]) find git.cmd as well as git.exe on Windows?

Good point. I think it should as long as `shell=True` on Windows.

mrclary:
`git.exe` was found for me with `shell=False`

dalthviz:
Thanks for the feedback @mrclary ! Checking I was able to change it to just `git` and without `shell=True`. Should I change this to just remove the `"/usr/bin/env"` so this will end up being for all the platforms:

```python
        result = subprocess.run(
            ["git", "merge-base", "--fork-point", "master"]
        )
```

ccordoba12:
+1 from me because it'd make the code simpler.

mrclary:
+1 from me too.
Just the caveats that I mentioned: there may be instances where `git` is not installed, in which case the subprocess will fail and `not master` will always be written to `boot_branch.txt` and `spyder` will always be installed. Since this is primarily for developers, I think this edge case will be very rare.

mrclary (approved):

PR #19700, opened 2022-10-04
PR: Add dialog to confirm downloaded installer installation and update related code organization (Application)

dalthviz:
## Description of Changes

* [ ] Wrote at least one-line docstrings (for any new functions)
* [ ] Added unit test(s) covering the changes (if testable)

* [x] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

See https://github.com/spyder-ide/spyder/issues/19112#issuecomment-1228678444

A preview showing the new dialog without having the installer downloaded:

installer_download

A preview showing the new dialog when the installer was already downloaded previously:

installer

### Issue(s) Resolved

Part of #19112

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct: dalthviz

PR #19871, opened 2022-10-17
PR: Only create the application update status bar for standalone installers, increase timer interval to start checking for updates and change cancel download button shape (Application)

### LINKED ISSUE #19868: Feedback about auto-update mechanism for installers
Before releasing 5.4.0, I'd like for us to address two issues with this new mechanism:

1. Spyder can't be closed quickly after startup because the thread that checks for updates takes too much time to complete and we wait for it to finish before closing. Is there a way to simply kill that thread? If that's not possible, I think we could start it with a timer that waits a minute or two after Spyder starts to check for updates. That would work similarly to the way we check for dependencies.
2. I think we should only show the status bar widget that handles updates for our Windows and Mac installers. We have multiple ways to install Spyder and I think in those cases that widget would be of little value to users.

dalthviz:
## Description of Changes

### Issue(s) Resolved

Fixes #19868

### Affirmation

### DIFF
--- a/spyder/plugins/application/container.py
+++ b/spyder/plugins/application/container.py
@@ -102,19 +102,20 @@ class ApplicationContainer(PluginMainContainer):
     # -------------------------------------------------------------------------
     def setup(self):
 
-        self.application_update_status = ApplicationUpdateStatus(parent=self)
-        self.application_update_status.sig_check_for_updates_requested.connect(
-            self.check_updates
-        )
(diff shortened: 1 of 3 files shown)


dalthviz:
As a side note, a preview of how things are working with the latest changes added with PR https://github.com/spyder-ide/spyder/pull/19700 and https://github.com/spyder-ide/spyder/pull/19793 on MacOS:

update_mac

ccordoba12 (approved):

PR #19969, opened 2022-11-02
PR: Some fixes to make Spyder work with the latest changes in PyLSP

ccordoba12:
## Description of Changes

- Request documentation in plain text to the server because it sends it now by default as Markdown, which we can't handle.
- Fix error when showing completion hints due to the previous change in the server as well.
- Fix not showing hovers for objects without docstrings. Signatures that come from the server are now separated from two LF characters, which was causing this problem.
- Update outline if a tree is empty but it has cached items from the server. I discovered this bug while working with Spyder these days and although it's not caused by a server update, I think it's important to fix it because it makes the Outline display files with no symbols when they actually have symbols for it.

### Affirmation

### DIFF
--- a/.github/scripts/install.sh
+++ b/.github/scripts/install.sh
@@ -38,6 +38,9 @@ if [ "$USE_CONDA" = "true" ]; then
     else
         mamba install 'numpy<1.23'
     fi
+
+    # Install docstring-to-markdown until we release PyLSP 1.6.0
+    mamba install docstring-to-markdown
 else
     # Update pip and setuptools
     python -m pip install -U pip setuptools wheel build
@@ -63,6 +66,8 @@ else
         pip install pyqt5==5.12.* pyqtwebengine==5.12.*
     fi
 
+    # Install docstring-to-markdown until we release PyLSP 1.6.0
+    pip install docstring-to-markdown
 fi
 
 # Install subrepos from source
--- a/installers/macOS/req-extras.txt
+++ b/installers/macOS/req-extras.txt
@@ -1,5 +1,6 @@
 # Spyder extra packages
 autopep8
+docstring-to-markdown
 flake8
 Paramiko
 pycodestyle
--- a/external-deps/python-lsp-server/.gitrepo
+++ b/external-deps/python-lsp-server/.gitrepo
@@ -6,7 +6,7 @@
 [subrepo]
 	remote = https://github.com/python-lsp/python-lsp-server.git
 	branch = develop
-	commit = 6501e9eb80c7c3b35cf93f634834558fc88dee68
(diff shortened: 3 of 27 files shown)


.github/scripts/install.sh
```diff
@@ -38,6 +38,9 @@ if [ "$USE_CONDA" = "true" ]; then
     else
         mamba install 'numpy<1.23'
     fi
+
+    # Install docstring-to-markdown until we release PyLSP 1.6.0
+    mamba install docstring-to-markdown
```
ccordoba12, line 43: @dalthviz, this needs to be removed when updating our dependencies before releasing 5.4.0.

.github/scripts/install.sh
```diff
@@ -63,6 +66,8 @@ else
         pip install pyqt5==5.12.* pyqtwebengine==5.12.*
     fi
 
+    # Install docstring-to-markdown until we release PyLSP 1.6.0
+    pip install docstring-to-markdown
```
ccordoba12, line 70: Also this.

installers/macOS/req-extras.txt
```diff
@@ -1,5 +1,6 @@
 # Spyder extra packages
 autopep8
+docstring-to-markdown
```
ccordoba12, line 3: And this.

ccordoba12:
@spyder-ide/core-developers, after this PR you need to install `docstring-to-markdown` in your development environments. Otherwise Spyder will crash.

mrclary:
> @spyder-ide/core-developers, after this PR you need to install `docstring-to-markdown` in your development environments. Otherwise Spyder will crash.

Under what circumstances with this cause Spyder to crash?
Is this not a Spyder dependency or python-lsp-server dependency?
How is this installed for end users? I only see it added to the macOS installer extra packages.

ccordoba12:
That package is a direct dependency of PyLSP 1.6.0, which I just released. The thing is Spyder imports a little function from a PyLSP module which happens to also import `docstring-to-markdown`, so it'll crash if it's not installed.

mrclary:
> That package is a direct dependency of PyLSP 1.6.0, which I just released. The thing is Spyder imports a little function from a PyLSP module which happens to also import `docstring-to-markdown`, so it'll crash if it's not installed.

Okay. Since it's not an optional dependency (plugin) for python-lsp-server, it does not need to be specified in the macOS extra packages (although there is no harm).

ccordoba12:
Correct. I had to add `docstring-to-markdown` as a requirement before releasing PyLSP 1.6.0 so that the Mac app workflow didn't fail. But @dalthviz will remove its entry tomorrow before releasing 5.4.0.

PR #20002, opened 2022-11-07
PR: Update remote when syncing subrepos

mrclary:
When updating subrepos in Spyder using the current instructions (`git subrepo clone`), the remote specification in `.gitrepo` is not updated to reflect the cloned remote, but remains as the original upstream specification. 

This can be an issue for any process that requires access to the remote repository at the commit and/or branch specified in `.gitrepo`, since that commit and/or branch may not exist on the (incorrectly) specified remote. The conda-based installers are an example of such a process.

This PR changes the instructions to use `git subrepo pull ... -r <remote> -u` to ensure that the remote specification is also updated.

### Affirmation

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -164,7 +164,7 @@ As an example, let's assume that (i) your Github user name is `myuser`; (ii) you
     ```
     $ cd ~/spyder
     $ git checkout fix_in_spyder
-    $ git subrepo clone https://github.com/myuser/spyder-kernels.git external-deps/spyder-kernels -b fix_in_kernel -f
+    $ git subrepo pull external-deps/spyder-kernels -r https://github.com/myuser/spyder-kernels.git -b fix_in_kernel -u -f
     ```
 
 * You can now open a PR on `https://github.com/spyder-ide/spyder` and on `https://github.com/spyder-ide/spyder-kernels` for each of your branches.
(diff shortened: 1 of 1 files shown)


ccordoba12:
A couple of comments:

- What does `-u` do?
- I tried to use `gitsubrepo pull` before instead of `clone`, but I think it doesn't work to update a subrepo here if your PR there was rebased (because the old parent commit is different after the rebase). Did you test that possibility?

mrclary:
> * What does `-u` do?

```
    -u, --update          Add the --branch and/or --remote overrides to .gitrepo
```

> * I tried to use `gitsubrepo pull` before instead of `clone`, but I think it doesn't work to update a subrepo here if your PR there was rebased (because the old parent commit is different after the rebase). Did you test that possibility?

Ahh, good point. I did not test that, but I'll be sure to do so.
We currently use the `-f` (force) flag, and it will persist with this PR. I'll check to see if that has any effect. Also, there is a `-M` (method) flag that can take values 'merge' or 'rebase'; 'merge' is the default. I will check to see whether this has any effect as well.

mrclary:
Okay, upon testing, I can confirm the following for `git subrepo pull` on a remote rebase:
- _Without_ `-f` (force), an error is raised because the previous commit no longer exists (no fast-forward merge is possible). This reproduces the condition that @ccordoba12 was concerned about.
- Using `-M rebase` (method) has no effect.
- _With_ `-f` (force), the subrepo _is_ successfully updated.

So, I think we should be good.

CAM-Gerlach:
As this PR seems focused on just updating the git-subrepo commands, and my experience with using git-subrepo pales next to @ccordoba12 and @dalthviz (whereas I do make much heavier use of git submodule), I'll leave the review in their much more capable hands

dalthviz (approved):
Thanks @mrclary ! LGTM :+1:

PR #20005, opened 2022-11-08
PR: Add left spacer to Working directory toolbar

ccordoba12:
## Description of Changes

- Prevent the working directory toolbar to take up a huge space on the main toolbar, especially on high resolution screens, which looks bad in my opinion.

    **Before**

    imagen

    **After**

    imagen

- Add ellipsis to the text displayed in the toolbar.
- Remove a class constant from the IPython console widget because it's not really necessary. The constant points to a specific section in the `Special consoles` menu, but that menu doesn't have sections (I realized that after #20000 was merged).
- Fix building the Windows installer, which started to fail due to the latest release of Psutil.

### Affirmation


ccordoba12:
@dalthviz. I also solved in this PR an error when building with the Windows installer, which was introduced by the latest release of Psutil. I basically fixed it in the same way as in PR #17157, but could you check locally if the installer runs as expected? Thanks!

dalthviz (commented):
Thanks @ccordoba12 ! Checked locally on WIndows with the windows installer and seems like things are working :+1:

I just have a question regarding the behavior when paths are longer than the now kind of fixed space. Should we do some ellipsis to the path? Checking seems like the ellipsis is done for the combobox options but no for the selected path:

pwd


Also, could it be possible to fix on this PR a typo on the README binder badge (the one that points to `spyder-stable` says `lastests` instead of `latest`)?

Beside that this LGTM :+1:

ccordoba12:
> I just have a question regarding the behavior when paths are longer than the now kind of fixed space. Should we do some ellipsis to the path?

Great idea @dalthviz! Although it was not trivial, I found a way to do it, which looks really nice

imagen

> Also, could it be possible to fix on this PR a typo on the README binder badge (the one that points to spyder-stable says lastests instead of latest)?

Good catch. Done as well.

spyder/widgets/helperwidgets.py
```diff
@@ -274,13 +275,33 @@ def update_status(self, value, value_set):
...
-        super(IconLineEdit, self).paintEvent(event)
-        painter = QPainter(self)
+        Include a validation icon to the left of the line edit and elide text
+        if requested.
+        """
+        # Elide text if requested.
+        # See PR spyder-ide/spyder#20005 for context.
+        if self.elide_text and not self._focus_in:
+            # This code is taken for the most part from the
+            # AmountEdit.paintEvent method, part of the Electrum project.
```
dalthviz, line 286: Should we add a link to the place on the Electrum project source code from where this was taken/is based?

dalthviz (commented):
Thanks for the changes @ccordoba12 , just left a comment regarding adding a link to the project source code which was used to implement the elide text logic. Other than that LGTM :+1:

spyder/widgets/helperwidgets.py
ccordoba12: I mentioned that explicitly in the `NOTICE.txt` entry I created for Electrum:

https://github.com/spyder-ide/spyder/pull/20005/files#diff-b810ebe4ffda4293c79d292059b45b59de7e649e86314ffd5a1ee4eb4b1e08c8R187

But I could add it here too if you that's important.

dalthviz: Yes please :+1:

ccordoba12: Ok, I added an extra line mentioning that for details people need to take a look at the Electrum entry in Preferences.

dalthviz (approved):
Thanks @ccordoba12 !

PR #10894, opened 2019-12-01
PR: Simplify find replace, add highlight by default and update colors

goanpeca:
## Description of Changes

* [x] Wrote at least one-line docstrings (for any new functions)
* [ ] Added unit test(s) covering the changes (if testable)
* [x] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

### Changes

- Added placeholder texts (`Find` `Replace`)
- Remove the unnecessary `Replace with` word. 
- Make both find/replace text boxes uniform (Same length)
- Update the color for match results (using unmatched braces color from theme)
- Update the highlight color of match results (no matches is highlighted with unmatched braces color from theme if no matches are found)
- Remove the textedit background coloring if no matches found in regex (it was really not providing much help). Instead added the unmatched_parens color from theme as the text color (not the background)
- Always highlight changes, remove "Highlight" toggle button. Why would anyone not want the matches to be highlighted ???
- Shorten texts for replace texts
- Add live update of matches if editor is updated and the widget is visible

#### Before

Screen Shot 2019-11-30 at 19 14 13

#### After

Screen Shot 2019-11-30 at 19 13 30

#### Workflow

**Dark**
find-dark

**Light**
find-light

**Live updates**
find-live

### Issue(s) Resolved

Fixes #7247
Fixes #10429
Fixes #1869

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct: @goanpeca

PR #20049, opened 2022-11-14
PR: Improve UX/UI of the `FindReplace` widget

### LINKED ISSUE #7247: Find and Replace stops highlighting any matches if any change is made in the Editor
## Problem Description

Often, I'll want to highlight all occurrences of a certain name, in order to modify, check something, or otherwise take action at or around instances of it. This would appear to be a prototypical use case of Find/Replace's highlighting feature; however, an apparent bug makes it almost useless—the highlighting won't stick if even one character is added, deleted or modified in the editor, whether or not in or near (even in a different line) any matched result,  In one stroke, this prevents most if not nearly all the use cases for highlighting I can think of, at least that aren't filled by double-clicking on a name, and is unlike other editors (e.g. NP++, Eclipse etc) where the highlighting sticks around until either the user turns it off or the Find/Replace dialog is closed.

1. Enter any text in the "Find" field with one or more matches
2. Enable the Highlighting option in that dialog
3. Enter, delete any text anywhere in the active editor

Poof, highlighting vanishes.

## Proposed Solutions

I haven't yet dug into the code, but any quick suggestions as to why this is happening, and if there are any major obstacles to fixing it?

## Package Versions

* Spyder version: 3.2.8 
* Python version: 3.6.4
* Qt version: 5.6.2
* PyQt5 version: 5.6
* Operating system: Windows 8.1

### Dependencies

```
sphinx >=0.6.6   :  1.7.2 (OK)
```

ccordoba12:
## Description of Changes

* Don't use background colors to signal that a search was unsuccessful or the used regular expression has errors. Instead, use an action and tooltips for that.

    **Warning**
    imagen

    **Error**
    imagen

* Remove text labels from replace buttons and make find and replace line edit widths to match.

    imagen

* Fix margins around Find/replace widget for all plugins that rely on it. Also fix button margins around them.

    **Before**

    imagen

    **After**

    imagen

* Add a clear button to all `PatternCombobox` and don't clear search results in the editor if text has changed on it and update total number of matches.

    dont-clear-found-results

* Add a Material Design icon for the "Whole words" button.

    imagen

* Improve tooltips for all buttons.
* Improve `FindReplace` tests and add several new ones.

### Issue(s) Resolved

Fixes #7247
Supersedes #10894

### Affirmation


ccordoba12:
@jitseniesen, I improved all the Find button tooltips, as discussed in our last developer meeting. Could you take a look at them and let me know what you think? Thanks!

dalthviz (commented):
Thanks @ccordoba12 ! Besides the tests that need an update, checking locally some questions/things I noticed:

* The replace button is now selected if using the replace shortcut (`Ctrl + R`) but not unselected if using the find shortcut (`Ctrl + F`):

findreplace

* The matches text `x of y` can change size so maybe that info should be also painted inside the combobox/lineedit as done with the warning/error icon?

* Before, you could click the combobox/lineedit to remove with a click all the text in it, but testing seems like that functionality was removed here?

ccordoba12:
Thanks @dalthviz for your review. My comments:

> The replace button is now selected if using the replace shortcut (Ctrl + R) but not unselected if using the find shortcut (Ctrl + F)

Fixed.

> The matches text x of y can change size so maybe that info should be also painted inside the combobox/lineedit as done with the warning/error icon?

I tried to do that but you have to do very ugly hacks in Qt to implement it and it doesn't look that good. To avoid that the size of that label affects the size of the find/replace line edits, I decided to set a max width for them and move the matches label to the far right:

find-replace

I took this idea from Firefox:

image

That way, in wide enough screens the line edits won't change in size.

> Before, you could click the combobox/lineedit to remove with a click all the text in it, but testing seems like that functionality was removed here?

I guess you refer to clicking on this button, right?

image

Well, I had to remove it to show the new messages label. Besides, we can't control the appearance of that button so it appears almost invisible in dark backgrounds:

image

dalthviz:
Thanks for the changes @ccordoba12 !

Checked locally and the fixes are working :+1: However seems like the fixed max-with for the inputs can leave a lot of space on the right side:

imagen

And it also can end with a kind of weird layout if you reduce the Editor width (probably the buttons should collapse into a `...` button or something instead of the input fields width being changed to be that small?):

imagen

Also, since this removes functionality (removes the button to erase the input content with one mouse click) maybe it could be better to leave it for Spyder 6? What do you think?

ccordoba12:
Thanks for the review @dalthviz! About your comments:

> However seems like the fixed max-with for the inputs can leave a lot of space on the right side

Ok, I changed the code so that the minimum width of the line edits is half of the available space:

imagen

Also, I increased their default width to 400px:

imagen

> And it also can end with a kind of weird layout if you reduce the Editor width (probably the buttons should collapse into a ... button or something instead of the input fields width being changed to be that small?)

That would be nice to have but it requires to add all search buttons to a QToolbar and then adjust its style to show the ellipsis. So, I think it's a bit out of the scope of this pull request.

> Also, since this removes functionality (removes the button to erase the input content with one mouse click) maybe it could be better to leave it for Spyder 6? What do you think?

I restored that button and made it use the right color for our interface theme.

imagen

ccordoba12:
@dalthviz, while reviewing old PRs, I found a couple of interesting ideas in #10894, which I decided to include here. 

So, now we don't clear the found results if the `FindReplace` widget is visible and users change text in the editor (which is really useful). And we update the total number of matches if that's the case:

dont-clear-found-results

dalthviz (approved):
Thanks @ccordoba12 for the changes!

Checking seems like the space management greatly improved :+1:

I think this LGTM as it is :+1:

However, just a couple of extra comments/questions/ideas:

* Could be possible to check if there is actual text to remove  in the field before showing the remove text action? Currently the action is always shown regardless of the field content.

* Is there a way to change the placeholder color for the `Replace` field to grey? I think you mentioned is was some sort of upstream bug but thinking about it is kind of weird having it in that color and could give the impression that it is something that you need to remove to use the field (in addition you have the action to clear the field, which could cause the impression that the clear field action doesn't work if you try to use it to remove the placeholder)

ccordoba12:
> Could be possible to check if there is actual text to remove in the field before showing the remove text action? Currently the action is always shown regardless of the field content.

Great idea @dalthviz! I implemented that in my second to last two commit, and in my last commit I added a test for it.

> Is there a way to change the placeholder color for the Replace field to grey?

This is actually a bug in the Qt package provided by Conda-forge. If you use the PyQt pip packages, you'll see that that placeholder text is show in grey, as expected:

imagen

dalthviz:
Merging this to update the translations strings. If there are any further comments about this let us know @jitseniesen @CAM-Gerlach to revisit them in a follow up PR :+1:

CAM-Gerlach:
I did have some feedback to @ccordoba12 regarding the icons, but it seems I never submitted my draft comment here, only the one I made over DM directly to Carlos, sorry.

ccordoba12:
@CAM-Gerlach, please open a new issue with your suggestions (which are really valuable) or post them here, and I'll address them in a follow-up PR, as @dalthviz suggested.

CAM-Gerlach:
Thanks, opened as issue #20255

PR #20056, opened 2022-11-15
PR: Update conda-based installers

mrclary:
## Description of Changes

* Fixed issue where installer job was run when external-deps package build jobs fail
* Fixed issue where certain directories were missing when package build jobs are skipped
* Fixed issue where scm version is improperly resolved due to cloned source creating an unclean repository state
* Minor refactor of conda package building module
* Use `--no-test` flag rather than removing test directives in `meta.yaml`
* Fix issue where shallow clones could result in unknown commit checkout
* Do not build for macOS arm64 architecture
* Update to accommodate new `spyder-kernels` feedstock
* Source code patch file is generated on demand rather than tracked


mrclary:
@ccordoba12, this should be ready for review now.

installers-conda/build_conda_pkgs.py
```diff
@@ -318,25 +319,48 @@ class SpyderKernelsCondaPkg(BuildCondaPkg):
...
             """
-            Build conda packages from local Spyder and external-deps sources.
-            Alternative git repo for python-lsp-server may be provided by
-            setting the environment variable PYTHON_LSP_SERVER_SOURCE,
-            otherwise the upstream remote will be used. All other external-deps
-            use the subrepo source within the Spyder repo.
+            Build conda packages to local channel.
+
+            This module builds conda packages for Spyder and external-deps for
+            inclusion in the conda-based installer. The Following classes are
```
ccordoba12, line 325:
```suggestion
            inclusion in the conda-based installer. The following classes are
```

installers-conda/resources/installers-conda.patch
```diff
@@ -1,313 +0,0 @@
-From 40db10418cf9b46a70b48026d202d4474d62b99a Mon Sep 17 00:00:00 2001
```
ccordoba12, line 1: Why is this patch no longer necessary? I thought you needed to do several adjustments to our source code for the installers to work as expected.

ccordoba12 (commented):
A couple of minor comments for you @mrclary, otherwise looks good to me.

installers-conda/resources/installers-conda.patch
mrclary: The patch is still necessary. However, I think it will be simpler to create the file on-demand rather than track and update it.

Currently, necessary changes are made on the `installers-conda-patch` branch, but then a separate PR must be made on the `5.x` branch to update the patch file when the `installers-conda-patch` gets new commits. With this PR we don't need to track and update a patch file on the `5.x` branch, rather the patch file is created when the spyder conda package is built.

ccordoba12: Ok, thanks for the clarification. I didn't understand where the patch went.

ccordoba12 (approved):
Thanks @mrclary! These are really nice improvements!

PR #13496, opened 2020-08-07
PR: Add Wrap lines checkbox to source menu

arteagac:
## Description of Changes

* [ ] Wrote at least one-line docstrings (for any new functions)
* [ ] Added unit test(s) covering the changes (if testable)

* [x] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))

Implemented checkbox for "Wrap lines" in the Source menu. I tested it and this properly synchronizes with the Editor Preferences.
image

### Issue(s) Resolved

Fixes #7997

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct:
arteagac

PR #20090, opened 2022-11-19
Feature/wrap lines menu

### LINKED ISSUE #7997: Include toggable "Wrap lines" checkbox in Source menu for easy access
## Problem Description

Currently, "Wrap lines" is buried deep within the Preferences and only available from there, whereas it is nearly always accessible from the toolbar (or at very least a top-level menu) in most other editors/IDEs, and I toggle it quite frequently in such editors, second only to the "Show blank spaces", which is currently menu-bound. Therefore, both certainly deserve a place on the "Source" (or "Edit") toolbar, as the latter are ideal for functions frequently toggled on or off and where observing the status is helpful.

maurerle:
## Description of Changes

* [x] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/)) - see #13496 

Added wrap lines Command in Source Menu
Added Alt + Z as a shortcut for it, as it is used by VSCode and this shortcut is currently unused

### Issue(s) Resolved

Fixes #13496
Fixes #7997

### Affirmation


ccordoba12 (commented):
Thanks @maurerle for reviving @arteagac's contribution! I tested this locally and it's working as expected.

However, I'd be in favor of wrapping lines per file instead of doing it for all files (as in this PR). That's why I'd suggest to change `wrap_lines_action` to be a regular action (i.e. non-checkable), such as `trailingspaces_action`, which only has effect for the current file.

By doing that, we'll also be able to remove this option in our Preferences:

image

which I think is of little value because you don't want to see all your files wrapped.

maurerle:
This is a display/visualization setting which has two states (wrap lines on or off) I don't think that something without a checkbox is sufficient here.
It does not change the actual lines like traliningspaces_action does.

Notepad++ has this feature implemented as it is currently in spyder (once for all views).
PyCharm and VSCode remember this setting per opened editor.
All three use checkboxes to indicate the current setting.

I actually like the behavior as it is best.

maurerle:
Wouldn't this be a nice addition to the 6.0 release too? :blush:

PR #20091, opened 2022-11-19
PR: Remove check_path to allow running in debugger

maurerle:
## Description of Changes

PYTHONPATH entries are removed on start, resulting in a the warning shown on `check_path` in requirements - program does not start.
Removing pythonpath was added to disable shadowed base libraries, but I think thats exactly how the PYTHONPATH could and would behave.
Still it makes sense to remove them so that it does not create problems.

Yet the resulting warning
"Spyder must be installed properly (e.g. from source: 'python setup.py install'), or directory 'dir' must be in PYTHONPATH environment variable."
is not helpful and is removed through this PR

~~I would therefore not remove entries on PYTHONPATH from PATH~~

### Affirmation

### DIFF
--- a/spyder/app/mainwindow.py
+++ b/spyder/app/mainwindow.py
@@ -37,7 +37,6 @@ import traceback
 # Check requirements before proceeding
 #==============================================================================
 from spyder import requirements
-requirements.check_path()
 requirements.check_qt()
 
 #==============================================================================
--- a/spyder/requirements.py
+++ b/spyder/requirements.py
@@ -32,16 +32,6 @@ def show_warning(message):
     raise RuntimeError(message)
 
 
-def check_path():
-    """Check sys.path: is Spyder properly installed?"""
-    dirname = osp.abspath(osp.join(osp.dirname(__file__), osp.pardir))
-    if dirname not in sys.path:
-        show_warning("Spyder must be installed properly "
-                     "(e.g. from source: 'python setup.py install'),\n"
-                     "or directory '%s' must be in PYTHONPATH "
-                     "environment variable." % dirname)
-
-
 def check_qt():
     """Check Qt binding requirements"""
     qt_infos = dict(pyqt5=("PyQt5", "5.9"), pyside2=("PySide2", "5.12"))


ccordoba12:
Hey @maurerle, I don't agree with this change. We remove PYTHONPATH in `start.py` to protect Spyder from crashing at startup when, for instance, a user adds to PYTHONPATH a module named `random.py` or `string.py`, which is not uncommon.

You also said:

> PYTHONPATH entries are removed on start, resulting in a the warning shown on check_path in requirements - program does not start.

The way you need to start Spyder when developing it is by running `python bootstrap.py` in terminal, as explained in our [Contributing guide](https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md#running-spyder). That makes a lot of adjustments necessary for development (like installing Spyder and our subrepos in `external-deps` in development mode). So if you're starting Spyder in a different way, please resort back to this mechanism, which is used by the entire development team.

maurerle:
So I sometimes debug Spyder (like with the actual debugger) using VSCode, that way:

```
conda create -n spyder python=3.10
cd spyder
pip install -e .[test]
```

Then select the interpreter spyder and hit F5 to debug the bootstrap.py command.
This runs ` /usr/bin/env /opt/conda/envs/spyder/bin/python /home/maurer/.vscode-o
ss/extensions/ms-python.python-2022.18.2/pythonFiles/lib/python/debugpy/adapter/../../debugpy/launcher 43327 -- /home/maurer/gitea/spyder/bootstrap.py `

And in that situation I am getting here:

image


So you might say that that way is not supported and i should rather run something like `python -m debugpy --listen 43327 bootstrap.py` and attach my debugger to that process but in this case the message is not very helpful, as my PYTHONPATH is correct?

ccordoba12:
Ok, I see. Then please check what happens if you remove the call to `check_path` in `mainwindow.py`. It shouldn't be necessary anymore.

maurerle:
Yes, that works too, great.
I updated my PR accordingly.

The check was introduced in 1f73f2f9ac798e09d724324f71a9fbae27de9d07 - I did not find which case it was meant to handle, but it is already 12 years ago.

ccordoba12:
> The check was introduced in https://github.com/spyder-ide/spyder/commit/1f73f2f9ac798e09d724324f71a9fbae27de9d07 - I did not find which case it was meant to handle, but it is already 12 years ago.

Until a new months ago we injected the git clone Spyder directory into `sys.path`, but now we install Spyder in development mode.

PR #20092, opened 2022-11-19
PR: Fix Pytest catching up tests using Spyder-unittest or VSCode

maurerle:
## Description of Changes

When Debugging in an editor with pytest support (for example spyder with spyder-unittest plugin, PyCharm or VS Code) the tests were not catched up.

This was due to the fact that there was
* an import error `E   ImportError: QtWebEngineWidgets must be imported or Qt.AA_ShareOpenGLContexts must be set before a QCoreApplication instance is created`
* pytest_plugins in external-deps `Defining 'pytest_plugins' in a non-top-level conftest is no longer supported:`

Both is fixed in this commit allowing to run a single Test file directly from the IDE or even debugging a single test directly.
Before that, one could only run a single test using `pytest testfile.py` instead of the whole suite

### Affirmation


pytest.ini
```diff
@@ -34,3 +34,5 @@ markers =
     no_new_console: Prevent creating a new IPython console when reusing a mainwindow instance
     close_main_window: Close main window instance after test
     no_web_widgets
+
+addopts = --ignore=./external-deps
```
ccordoba12, line 38:
```suggestion
addopts = --ignore=./external-deps

```

Add an empty line here, which we tend to do for all our files.

spyder/app/tests/conftest.py
```diff
@@ -17,6 +17,7 @@
 from qtpy.QtCore import Qt
 from qtpy.QtTest import QTest
 from qtpy.QtWidgets import QApplication, QFileDialog, QLineEdit, QTabBar
+from qtpy import QtWebEngineWidgets
```
ccordoba12, line 20:
```suggestion
# This is required to run our tests in VSCode or Spyder-unittest
from qtpy import QtWebEngineWidgets  # noqa
```

Add a clarifying comment and `# noqa` so we don't remove this line in the future if it's reported by linters as unnecessary.

ccordoba12 (commented):
Thanks @maurerle for your help with this! I left a couple of small suggestions for you, otherwise looks good to me.

ccordoba12 (approved):
Thanks @maurerle!

Note: In the future please add all suggestions to the batch, so they are pushed in a single commit. That's described here:

https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request

PR #20106, opened 2022-11-22
PR: Fix issue where user environment variables with line endings were not parsed correctly on Unix platforms

### LINKED ISSUE #20097: ValueError when trying to get environment variables on Linux

In Spyder open Tools menu, then click Current user environment variables...

Doesnt open

### Traceback
```python-traceback
  File "/home/mark/tutorial_env/lib64/python3.10/site-packages/spyder/plugins/application/container.py", line 252, in show_user_env_variables
    self.dialog_manager.show(UserEnvDialog(self))
  File "/home/mark/tutorial_env/lib64/python3.10/site-packages/spyder/utils/environ.py", line 162, in __init__
    super().__init__(get_user_env(), parent, title, readonly)
  File "/home/mark/tutorial_env/lib64/python3.10/site-packages/spyder/utils/environ.py", line 71, in get_user_env
    env_dict = get_user_environment_variables()
  File "/home/mark/tutorial_env/lib64/python3.10/site-packages/spyder/utils/environ.py", line 63, in get_user_environment_variables
    k, v = kv.split('=', 1)
ValueError: not enough values to unpack (expected 2, got 1)
```

## Versions

* Spyder version: 5.4.0  (pip)
* Python version: 3.10.8 64-bit
* Qt version: 5.15.2
* PyQt5 version: 5.15.7
* Operating System: Linux 6.0.5-200.fc36.x86_64

### Dependencies

```
# Mandatory:
atomicwrites >=1.2.0          :  1.4.1 (OK)
chardet >=2.0.0               :  5.0.0 (OK)
cloudpickle >=0.5.0           :  2.2.0 (OK)
cookiecutter >=1.6.0          :  2.1.1 (OK)
IPython >=7.31.1;<8.0.0       :  7.34.0 (OK)
jedi >=0.17.2;<0.19.0         :  0.18.1 (OK)
jellyfish >=0.7               :  0.9.0 (OK)
jsonschema >=3.2.0            :  4.17.0 (OK)
keyring >=17.0.0              :  23.11.0 (OK)
nbconvert >=4.0               :  7.2.5 (OK)
numpydoc >=0.6.0              :  1.5.0 (OK)
parso >=0.7.0;<0.9.0          :  0.8.3 (OK)
psutil >=5.3                  :  5.9.4 (OK)
pygments >=2.0                :  2.13.0 (OK)
pylint >=2.5.0;<3.0           :  2.15.6 (OK)
pylint_venv >=2.1.1           :  2.3.0 (OK)
pylsp >=1.6.0;<1.7.0          :  1.6.0 (OK)
pylsp_black >=1.2.0           :  1.2.1 (OK)
qdarkstyle >=3.0.2;<3.1.0     :  3.0.3 (OK)
qstylizer >=0.2.2             :  0.2.2 (OK)
qtawesome >=1.2.1             :  1.2.1 (OK)
qtconsole >=5.4.0;<5.5.0      :  5.4.0 (OK)
qtpy >=2.1.0                  :  2.3.0 (OK)
rtree >=0.9.7                 :  1.0.1 (OK)
setuptools >=49.6.0           :  59.6.0 (OK)
sphinx >=0.6.6                :  5.3.0 (OK)
spyder_kernels >=2.4.0;<2.5.0 :  2.4.0 (OK)
textdistance >=4.2.0          :  4.5.0 (OK)
watchdog >=0.10.3             :  2.1.9 (OK)
xdg >=0.26                    :  0.28 (OK)
zmq >=22.1.0                  :  24.0.1 (OK)

# Optional:
cython >=0.21                 :  None (NOK)
matplotlib >=3.0.0            :  None (NOK)
numpy >=1.7                   :  None (NOK)
pandas >=1.1.1                :  None (NOK)
scipy >=0.17.0                :  None (NOK)
sympy >=0.7.3                 :  None (NOK)
```

mrclary:
## Description of Changes

#20297 unsatisfactorily fixed #20097 by skipping environment variables that do not parse correctly. This PR remedies this.
* For unix, use a clean subprocess environment, invoke a login shell, extract `os.environ` from a python process. This circumvents any parsing issues and should reflect exactly what a user would expect from a python process in a standard terminal.
* For Windows, get environment variables directly from the system registry.
* Add regression test for #20097
  * Modify user environment variables in system registry or shell startup file
  * Restore user environment variables
  * Only run on CI to protect developers' local user environment variables

### Issue(s) Resolved

See #20097

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -11,7 +11,6 @@ NOTE: DO NOT add fixtures here. It could generate problems with
       QtAwesome being called before a QApplication is created.
 """
 
-# Standard libary imports
 import os
 import os.path as osp
 import re
@@ -23,11 +22,6 @@ os.environ['SPYDER_PYTEST'] = 'True'
 
 # ---- Pytest adjustments
 import pytest
-
-# Spyder imports
-from spyder.config.base import running_in_ci
-from spyder.utils.environ import (get_user_env, set_user_env,
-                                  amend_user_shell_init)
 
 
 def pytest_addoption(parser):
@@ -142,20 +136,3 @@ def reset_conf_before_test():
 
     CONF.set('completions', 'provider_configuration', provider_configurations,
              notification=False)
-
-
-@pytest.fixture
-def restore_user_env():
-    """Set user environment variables and restore upon test exit"""
-    if not running_in_ci():
-        pytest.skip("Skipped because not in CI.")
-
-    if os.name == "nt":
-        orig_env = get_user_env()
-
-    yield
-
-    if os.name == "nt":
-        set_user_env(orig_env)
-    else:
-        amend_user_shell_init(restore=True)
--- a/conftest.py
+++ b/conftest.py
@@ -11,6 +11,7 @@ NOTE: DO NOT add fixtures here. It could generate problems with
       QtAwesome being called before a QApplication is created.
 """
 
+# Standard libary imports
 import os
 import os.path as osp
 import re
@@ -22,6 +23,11 @@ os.environ['SPYDER_PYTEST'] = 'True'
 
 # ---- Pytest adjustments
 import pytest
+
+# Spyder imports
+from spyder.config.base import running_in_ci
+from spyder.utils.environ import (get_user_env, set_user_env,
+                                  amend_user_shell_init)
 
 
 def pytest_addoption(parser):
@@ -136,3 +142,20 @@ def reset_conf_before_test():
 
     CONF.set('completions', 'provider_configuration', provider_configurations,
              notification=False)
+
+
+@pytest.fixture
+def restore_user_env():
+    """Set user environment variables and restore upon test exit"""
+    if not running_in_ci():
+        pytest.skip("Skipped because not in CI.")
+
+    if os.name == "nt":
+        orig_env = get_user_env()
+
+    yield
+
+    if os.name == "nt":
+        set_user_env(orig_env)
+    else:
+        amend_user_shell_init(restore=True)
--- a/spyder/utils/environ.py
+++ b/spyder/utils/environ.py
@@ -10,6 +10,9 @@ Environment variable utilities.
 
 # Standard library imports
 import os
+from pathlib import Path
+import re
+import sys
 try:
     import winreg
 except Exception:
@@ -51,25 +54,22 @@ def get_user_environment_variables():
     env_var : dict
         Key-value pairs of environment variables.
     """
-    if os.name == 'nt':
-        cmd = "set"
-    else:
-        cmd = "printenv"
-    proc = run_shell_command(cmd)
-    stdout, stderr = proc.communicate()
-
     try:
-        res = stdout.decode().strip().split(os.linesep)
+        if os.name == 'nt':
+            key = winreg.OpenKey(winreg.HKEY_CURRENT_USER, "Environment")
+            N = winreg.QueryInfoKey(key)[1]
+            env_var = dict([winreg.EnumValue(key, k)[:2] for k in range(N)])
+        else:
+            shell = os.environ.get("SHELL", "/bin/bash")
+            cmd = (
+                f"{shell} -l -c"
+                f""" "{sys.executable} -c 'import os; print(dict(os.environ))'" """
+            )
+            proc = run_shell_command(cmd, env={}, text=True)
+            stdout, stderr = proc.communicate()
+            env_var = eval(stdout, None)
     except Exception:
         return {}
-
-    env_var = {}
-    for kv in res:
-        try:
-            k, v = kv.split('=', 1)
-            env_var[k] = v
-        except Exception:
-            pass
 
     return env_var
 
--- a/spyder/utils/environ.py
+++ b/spyder/utils/environ.py
@@ -81,35 +81,42 @@ def get_user_env():
 
 
 def set_user_env(env, parent=None):
-    """Set HKCU (current user) environment variables"""
-    if os.name != 'nt':
-        raise NotImplementedError("Not implemented for %s platforms", os.name)
+    """
+    Set user environment variables via HKCU (Windows) or
+    shell startup file (unix).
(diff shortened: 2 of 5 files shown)


ccordoba12 (commented):
Thanks @mrclary for your work on this! Looks good to me, but I think it'd be nice if you could add a test for issue #20097, so we don't introduce a regression for it in the future.

mrclary:
@ccordoba12, this might be a good time to review what the `get_user_environment_variables` function should actually do.

Perhaps we need to define, for the purposes of Spyder, what "user environment variables" means. I have a working definition which is:
* The exported environment variables present when a user starts a login shell.

For Windows users, this is very straight-forward: everything inherits from the registry, so these values can be collected directly from there.

For unix platforms, this is more complicated: it depends on the default shell and the profile startup scripts (e.g. /etc/bashrc, ~/.bash_profile, etc.). Ostensibly, `os.environ` has the "user environment variables", but it has additional variables used by Spyder, so cannot be relied on to be equivalent to "user environment variables". Additionally, the "user environment variables" in `os.environ` are set at Spyder launch, so Spyder would have to be restarted if these variables change.

For these reasons, I was motivated to use a subprocess to get the "user environment variables" rather than `os.environ`. However, there is an oversight here because the subprocess actually injects `os.environ` into the environment variables by default, so at present it does not achieve its intended goal. I was curious how the unit test for the pythonpath manager plugin worked when assigning `os.environ['PYTHONPATH']`; this is how I realized the oversight.

conftest.py
```diff
@@ -23,6 +24,11 @@
 # ---- Pytest adjustments
 import pytest
 
+# Spyder imports
+from spyder.config.base import running_in_ci
+from spyder.utils.environ import (get_user_env, set_user_env,
+                                  amend_user_shell_init)
```
ccordoba12, line 30: This file should be as free of Spyder imports as possible to avoid possible circular imports (`pytest` does a lot of black magic so it's hard to predict if this will have repercussions in the future). There's also a message on the docstring of this file that warns about adding fixtures here.

So, I suggest to move the `restore_user_env` fixture to `spyder/app/tests/conftest.py` and then import it in `spyder/utils/tests/test_environ.py` (no changes will be required in `test_mainwindow.py`).

spyder/utils/environ.py
```diff
@@ -51,26 +54,23 @@ def get_user_environment_variables():
...
-    else:
-        cmd = "printenv"
-    proc = run_shell_command(cmd)
-    stdout, stderr = proc.communicate()
-
     try:
-        res = stdout.decode().strip().split(os.linesep)
+        if os.name == 'nt':
+            key = winreg.OpenKey(winreg.HKEY_CURRENT_USER, "Environment")
+            N = winreg.QueryInfoKey(key)[1]
```
ccordoba12, line 60: What this `N` stands for? Could you come up with a more telling name please?

spyder/utils/environ.py
```diff
@@ -81,35 +81,80 @@ def get_user_env():
...
-    if os.name != 'nt':
-        raise NotImplementedError("Not implemented for %s platforms", os.name)
-
-    reg = listdict2envdict(env)
-    types = dict()
-    key = winreg.OpenKey(winreg.HKEY_CURRENT_USER, "Environment")
-    for name in reg:
+    """
+    Set user environment variables via HKCU (Windows) or
+    shell startup file (unix).
```
ccordoba12, line 86:
```suggestion
    Set user environment variables via HKCU (Windows) or shell startup file
    (Unix).
```

spyder/utils/environ.py
```diff
@@ -81,35 +81,80 @@ def get_user_env():
...
+            SendMessageTimeout(HWND_BROADCAST, WM_SETTINGCHANGE, 0,
+                               "Environment", SMTO_ABORTIFHUNG, 5000)
+        except Exception:
+            QMessageBox.warning(
+                parent, _("Warning"),
+                _("Module <b>pywin32 was not found</b>.<br>"
+                  "Please restart this Windows <i>session</i> "
+                  "(not the computer) for changes to take effect.")
+            )
+    elif os.name == 'posix':
```
ccordoba12, line 115:
```suggestion
    elif os.name == 'posix' and running_in_ci():
```

I think this is necessary because we don't to give the impression that Spyder can really add new variables to the environment in Posix systems.

spyder/utils/environ.py
```diff
@@ -81,35 +81,80 @@ def get_user_env():
...
+        text = "\n".join([f"export {k}={v}" for k, v in env_dict.items()])
+        amend_user_shell_init(text)
+    else:
+        raise NotImplementedError("Not implemented for platform %s", os.name)
+
+
+def amend_user_shell_init(text="", restore=False):
+    """Set user environment variable for pytests on unix platforms"""
```
ccordoba12, line 123:
```suggestion
    """Set user environment variable for testing on Unix platforms"""
```

spyder/utils/tests/test_environ.py
```diff
@@ -7,6 +7,8 @@
 """
 Tests for environ.py
 """
+# Standard imports
```
ccordoba12, line 10:
```suggestion
"""

# Standard library imports
```

conftest.py
```diff
@@ -11,6 +11,7 @@
       QtAwesome being called before a QApplication is created.
 """
 
+# Standard libary imports
```
ccordoba12, line 14:
```suggestion
# Standard library imports
```

ccordoba12 (commented):
Great work here @mrclary! I left some small suggestions for you, otherwise looks almost ready to me.

conftest.py
mrclary: Thank you for the clarification, @ccordoba12. I saw the warning message but also saw another fixture in here, so I didn't know how strict we were with that. But the concern for Spyder imports makes sense.

ccordoba12 (approved):

Issue #10108, opened 2019-08-29
Spyder Error on Startup


zdhughes5:
### What steps will reproduce the problem?

After updating spyder with:

conda install -c spyder-ide jedi=0.14.1

As a solution to this issue:

https://github.com/spyder-ide/spyder/issues/10057

Starting spyder with:

spyder --new-instance &

Creates to provided error message on the spyder splash screen when starting up.

UPDATE:
In my case Spyder is being loaded over a network filesystem. I noticed that the network bandwidth was being used to capacity; since returning to normal levels the issue seems to have been resolved.

### Traceback
```python-traceback
  File "/nfs/programs/zdhughes/anaconda3/lib/python3.7/site-packages/spyder/plugins/completion/languageserver/client.py", line 319, in on_msg_received
    handler(resp['params'])
  File "/nfs/programs/zdhughes/anaconda3/lib/python3.7/site-packages/spyder/plugins/completion/languageserver/client.py", line 378, in process_server_capabilities
    if isinstance(server_capabilites['textDocumentSync'], int):
KeyError: 'textDocumentSync'
```

## Versions

* Spyder version: 4.0.0b4 
* Python version: 3.7.4
* Qt version: 5.9.6
* PyQt5 version: 5.9.2
* Operating System: Linux 5.1.6-1.el7.elrepo.x86_64

### Dependencies

```
pygments >=2.0    :  2.4.2 (OK)
qdarkstyle >=2.6.4:  2.7 (OK)
sphinx >=0.6.6    :  2.1.2 (OK)
psutil >=0.3      :  5.6.3 (OK)
pyls >=0.27.0     :  0.28.0 (OK)
nbconvert >=4.0   :  5.5.0 (OK)
pandas >=0.13.1   :  0.25.0 (OK)
numpy >=1.7       :  1.16.4 (OK)
sympy >=0.7.3     :  1.3 (OK)
cython >=0.21     :  0.29.2 (OK)
qtconsole >=4.5.2 :  4.5.3 (OK)
IPython >=4.0     :  7.7.0 (OK)
matplotlib >=2.0.0:  3.1.0 (OK)
pylint >=0.25     :  2.3.1 (OK)
```

ccordoba12:
Sorry, you also need to update `python-language-server` with this command:

    conda install -c spyder-ide python-language-server=0.28.2

Issue #9755, opened 2019-07-05
KeyError textDocumentSync when opening Spyder with a project


Keou0007:
### What steps will reproduce the problem?

- Close Spyder with a project open by clicking the "X" in the top right corner
- Open Spyder. Previous project opens.
- Issue reporter appears before the main window.

### Traceback
```python-traceback
  File "/home/keoughs/anaconda3/lib/python3.7/site-packages/spyder/plugins/editor/lsp/client.py", line 306, in on_msg_received
    handler(resp['params'])
  File "/home/keoughs/anaconda3/lib/python3.7/site-packages/spyder/plugins/editor/lsp/client.py", line 365, in process_server_capabilities
    if isinstance(server_capabilites['textDocumentSync'], int):
KeyError: 'textDocumentSync'
```

## Versions

* Spyder version: 4.0.0b3 
* Python version: 3.7.3
* Qt version: 5.9.6
* PyQt5 version: 5.9.2
* Operating System: Linux 2.6.32-754.14.2.el6.x86_64

### Dependencies

```
pygments >=2.0    :  2.4.2 (OK)
qdarkstyle >=2.6.4:  2.7 (OK)
sphinx >=0.6.6    :  2.1.2 (OK)
psutil >=0.3      :  5.6.3 (OK)
pyls >=0.27.0     :  0.27.0 (OK)
nbconvert >=4.0   :  5.5.0 (OK)
pandas >=0.13.1   :  0.24.2 (OK)
numpy >=1.7       :  1.16.4 (OK)
sympy >=0.7.3     :  1.4 (OK)
cython >=0.21     :  0.29.11 (OK)
qtconsole >=4.5.0 :  4.5.1 (OK)
IPython >=4.0     :  7.6.0 (OK)
matplotlib >=2.0.0:  3.1.0 (OK)
pylint >=0.25     :  2.3.1 (OK)
```

Keou0007:
I don't really understand what this is doing here, it's probably better to fix why 'textDocumentSync' isn't being found in server_capabilities, but it seems to me you could patch this to avoid a traceback by using `get` with a default.
```
if isinstance(server_capabilites.get('textDocumentSync', None), int):
```

Keou0007:
I have made a few potential fixes for this on my fork, happy to submit a PR if you think what I've done is suitable, but all they do prevent the exception being raised, so it isn't actually fixing what might be broken.

[...]

Issue #9557, opened 2019-06-12
Traceback on Spyder launch when auto-opens previous project


Keou0007:
### What steps will reproduce the problem?

Close spyder, saving all currently open files in current project.
Open spyder
Traceback!

### Traceback
```python-traceback
  File "/home/keoughs/anaconda3/lib/python3.7/site-packages/spyder/plugins/editor/lsp/client.py", line 259, in on_msg_received
    handler(resp['params'])
  File "/home/keoughs/anaconda3/lib/python3.7/site-packages/spyder/plugins/editor/lsp/client.py", line 320, in process_server_capabilities
    if isinstance(server_capabilites['textDocumentSync'], int):
KeyError: 'textDocumentSync'
```

## Versions

* Spyder version: 4.0.0b2 
* Python version: 3.7.3
* Qt version: 5.9.6
* PyQt5 version: 5.9.2
* Operating System: Linux 2.6.32-754.14.2.el6.x86_64

### Dependencies

```
pygments >=2.0     :  2.4.2 (OK)
qdarkstyle >=2.6.4 :  2.6.8 (OK)
sphinx >=0.6.6     :  2.1.0 (OK)
psutil >=0.3       :  5.6.2 (OK)
pyls >=0.19.0;<0.25:  0.24.0 (OK)
nbconvert >=4.0    :  5.5.0 (OK)
pandas >=0.13.1    :  0.24.2 (OK)
numpy >=1.7        :  1.16.4 (OK)
sympy >=0.7.3      :  1.4 (OK)
cython >=0.21      :  0.29.10 (OK)
qtconsole >=4.2.0  :  4.5.1 (OK)
IPython >=4.0      :  7.5.0 (OK)
matplotlib >=2.0.0 :  3.1.0 (OK)
pylint >=0.25      :  2.3.1 (OK)
```

ccordoba12:
I can't reproduce this after PR #9482, so I'm going to close it for now.

@Keou0007, please test things after we release beta3 (on June 26 or 27) to see if that's the case for you too.

Keou0007:
Will do

Keou0007:
I just installed beta3 on my linux machine and got this traceback on launch again. It seems sporadic now though, I've opened spyder 3 or 4 times today and gotten the traceback once.

PR #20113, opened 2022-11-23
PR: Fix random rare error in completion plugin

rear1019:
See commit message for details.

Notes:
* We have observed the fixed error on Linux only.
* There is a (remote) possibility that the fixed error is the one reported in #9557, #9755 and #10108.
* The fix in this PR is strictly speaking a workaround. The “proper” fix is to _not_ use a port from the ephemeral port range for the language server. 

### Affirmation

rear1019


ccordoba12 (approved):
Looks good to me, thanks @rear1019!

Issue #250, opened 2015-02-16
IPython interpreter hangs when editing Mayavi pipeline


spyder-issues-migrator:
_From [jed.lud...@gmail.com](https://code.google.com/u/114333776555678317161/) on 2010-07-08T16:22:36Z_

What steps will reproduce the problem?

 1. Run the attached mlab_test.py in an IPython interpreter
2. Open the Mayavi pipeline editor (boxed in red in mlab_scene.png)
3. Try to edit bounds (any of the six text boxes shown in mlab_bounds.png) What is the expected output? What do you see instead? Expect the scene to update after making edits. Instead, scene locks up, and you have to kill the IPython interpreter to recover. What version of the product are you using? On what operating system? Happens in both Spyder 1.1.2 and Spyder 2 ( revision ab7ea54940b3 ). Python 2.6, Win XP. 

Please provide any additional information below

. Editing the bounds works fine if the test script is run in a regular interpreter inside Spyder. It also works fine from a regular IPython session run from a console. It only seems to hang when run from IPython inside of Spyder.

**Attachment:** [mlab_test.py mlab_scene.png mlab_bounds.png](http://code.google.com/p/spyderlib/issues/detail?id=250)

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=250_

spyder-issues-migrator:
_From [jed.lud...@gmail.com](https://code.google.com/u/114333776555678317161/) on 2010-07-08T14:27:18Z_

Here are the IPython interpreter options I'm using inside Spyder:

-pylab -colors LightBG -xmode Plain

spyder-issues-migrator:
_From [pierre.raybaut](https://code.google.com/u/pierre.raybaut/) on 2010-07-13T00:56:51Z_

I was unable to reproduce this bug: on my machine (XP SP3, Python 2.6, Spyder v2.0 rev. ab7ea54940), everything works as expected.

Are you sure that you've tried the latest development version of v2.0?

spyder-issues-migrator:
_From [jed.lud...@gmail.com](https://code.google.com/u/114333776555678317161/) on 2010-07-13T08:04:13Z_

As for Spyder v2.0, I'm positive I've got the latest. I even created a completely new repository clone to make sure.

As for Spyder v1.1.2, the problem occurs there, too.

I've tried both ETS 3.3 and 3.4, and the problem is present with both versions. I'll have to experiment a bit more....

spyder-issues-migrator:
[...]

Issue #262, opened 2015-02-16
Run inside interpreter should change working directory


spyder-issues-migrator:
_From [mproel...@googlemail.com](https://code.google.com/u/105494983286078857315/) on 2010-08-05T07:21:25Z_

Hello,

I face the following problem. I have a file let's say c:\test\test.py which uses a dll stored in the same path c:\test\test.dll

My code would be something like

import ctypes
ctypes.cdll.LoadLibrary("test.dll") #dll in same directory as test.py

when I "run and interact" everything works fine, but on "run inside interpreter" I get an error like the module (test.dll) was not found as the current working directory is not c:\test

Honestly I haven't got the difference between "run and interact" and "run inside interpreter" yet. What is the difference of these two commands and is the behaviour I have as expected?
(I keep searching the possibility to mark an issue as defect/enhancement)

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=262_

spyder-issues-migrator:
_From [pierre.raybaut](https://code.google.com/u/pierre.raybaut/) on 2010-08-05T05:39:03Z_

(a simple user can't mark an issue as an enhancement which is unfortunate as it would be easier for me as well)

Well, actually you've just pointed out the difference between these two 'run' commands:
[1] "run and interact": open a brand new Python interpreter, run the script and interact with it right after its execution (this is strictly equivalent to 'python -i script.py' in a terminal, so the current working directory is the script directory)
[2] "run inside interpreter": run the script inside an already existing interpreter which had its own life before executing this script (namespace, working directory, environment variables, ...) -- as a consequence, results are expected to be different from [1]. This is almost equivalent to the following actions performed in a terminal:  open a Python interpreter, do what you want inside of it (computings, ...) and run your script with the 'execfile' function.

In other words, [1] is to be used when you want to run your script as you would outside Spyder or any other IDE but you want to interact with it (for debugging purpose for example). [2] is to be used when you want to run one or more script(s) inside an existing interpreter in which you have already performed some computings and with data stored in the namespace that you don't want to loose between each script execution.

**Status:** WontFix

Issue #261, opened 2015-02-16
run selection or current block doesn't work


spyder-issues-migrator:
_From [mproel...@googlemail.com](https://code.google.com/u/105494983286078857315/) on 2010-08-05T04:48:48Z_

Hello,

in Spyer2.0.0alpha4 the command 'run selection or current block' doesn't work

1. For instance create a new file with:
    test = 13
2. Hit "run and interact"
3. >>> test
   13
4. change test = 13 to test = 15 in your file
5. select test = 15 and hit run selection...
6. goto console and type:
   \>>> test
   13

There should be test = 15 shouldn't?

_Original issue: http://code.google.com/p/spyderlib/issues/detail?id=261_

spyder-issues-migrator:
_From [pierre.raybaut](https://code.google.com/u/pierre.raybaut/) on 2010-08-05T03:25:30Z_

Fixed: https://code.google.com/p/spyderlib/source/detail?r=595c27559f1cc8e23755b0c4cb7d7c3947cc4a01

**Status:** Fixed

PR #20259, opened 2022-12-24
PR: Fix concurrency for installers-conda workflow

mrclary:
Recent concurrency specifications in Spyder GitHub workflows (spyder-ide/spyder#20254) do not prevent `build-installers` jobs in `installers-conda.yml` from running under a certain circumstance. Example [PR: Cancel runs in progress when pushing new commits (CI) #250](https://github.com/spyder-ide/spyder/actions/runs/3768853668)

Concurrence cancels jobs that are pending or currently running. However, `build-installers` jobs are neither pending nor running until _after_ `build-matrix` and `build-noarch-conda-pkgs` jobs are all complete. In the circumstance that one or more of these jobs is not complete, they will be cancelled but `build-installers` jobs will then commence.

This PR should remedy this by replacing the conditional for `build-installers` jobs with explicit dependence on the results of the needed jobs. `! failure()` returns true for `"success"`, `"skipped"`, and `"cancelled"`, and therefore did not achieve the desired effect.

Note that if `build-installers` jobs are already pending/in process they will not be cancelled immediately. This is because the new `build-installers` jobs will not be queued until completion of the `build-matrix` and `build-noarch-pkgs` jobs. Upon queueing the new `build-installers` jobs, the old ones will be cancelled.

### DIFF
--- a/.github/workflows/installers-conda.yml
+++ b/.github/workflows/installers-conda.yml
@@ -24,7 +24,7 @@ concurrency:
 name: Create conda-based installers for Windows, macOS, and Linux
 
 jobs:
-  build-noarch-conda-pkgs:
+  build-noarch-pkgs:
     name: Build ${{ matrix.pkg }}
     runs-on: ubuntu-latest
     if: github.event_name != 'release'
@@ -96,8 +96,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     needs:
       - build-matrix
-      - build-noarch-conda-pkgs
-    if: ${{ ! failure() }}
+      - build-noarch-pkgs
+    if: contains(fromJson('["success", "skipped"]'), needs.build-noarch-pkgs.result) && needs.build-matrix.result == 'success'
     strategy:
       matrix:
         target-platform: ${{fromJson(needs.build-matrix.outputs.target_platform)}}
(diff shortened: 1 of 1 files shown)


mrclary:
The appropriate effect was confirmed by the following workflows
- [PR: Fix concurrency for installers-conda workflow #259](https://github.com/spyder-ide/spyder/actions/runs/3773039583), while `build-installers` was in progress, was cancelled by...
- [PR: Fix concurrency for installers-conda workflow #260](https://github.com/spyder-ide/spyder/actions/runs/3773064003). While `build-noarch-pkgs` was in progress, was cancelled by...
- [PR: Fix concurrency for installers-conda workflow #261](https://github.com/spyder-ide/spyder/actions/runs/3773066049). While `build-noarch-pkgs` was in progress, was cancelled by...
- [PR: Fix concurrency for installers-conda workflow #262](https://github.com/spyder-ide/spyder/actions/runs/3773072154)

mrclary:
However, I do not understand why an entire workflow is not cancelled, rather than cancelling individual jobs...
It seems that the language used in the [documentation](https://docs.github.com/en/actions/using-jobs/using-concurrency) is a bit mis-leading.

ccordoba12 (approved):
Thanks a lot @mrclary for your prompt help with this!

PR #20319, opened 2023-01-07
PR: Improvements to Conda-based Linux installer

mrclary:
## Description of Changes

- Update shortcut icon for Linux
- Add post-install script for Linux: update shortcut name to allow pinning to Dash
- Updates to `napari/label/bundle_tools_2` also gave executable permissions to the installer shell script
- Alias added to `.bashrc` for command line Spyder launch in all Linux flavors
- Constructor 4.0b7 required changing post-install script substituted variables
- SPY_COMMIT SPY_BRANCH variables added to Spyder runtime environment (all platforms)
- Post-install script for macOS now unnecessary

Double-clicking the installer script should run the script in a new Terminal window in interactive mode.
To see additional installation options, execute in a Terminal:
```
$ EXPERIMENTAL-Spyder-<version>-Linux-x86_64.sh -h
```

### Issue(s) Resolved

Fixes #20176

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct: @mrclary

PR #20350, opened 2023-01-11
PR: Introduce completions correctly for autocompletion characters and improve file completions (Editor)

### LINKED ISSUE #20331: Fix on dot completion leads to dot replacement
on this 2023-01-08 early morning master of 5.X  https://github.com/spyder-ide/spyder/commit/da661ccfe2f0355459f4e18f87d22337e4cd36dc

the dot completion works again, but if we validate a choice, the dot is replaced also, if we don't type a first letter of the world choosen

GIF_dot_completion_20230108

ccordoba12:
## Description of Changes

- Completions were not introduced correctly when one was selected in the completion widget after they were triggered automatically next to a dot.
- Fix introducing completions for files and directories that start with a dot (they were not introduced correctly either).
- Fix introducing completions for files and directories in empty strings, i.e. you can hit `Tab` in a empty string (`"<Tab>"`) and Jedi will provide file completions in that case. The problem was the selected entry was not placed correctly in the editor either.
- Fix introducing file and directory completions next to a path separator (i.e. `os.sep`).
- Expand tests to cover all those cases.
- Depends on https://github.com/python-lsp/python-lsp-server/pull/337

### Issue(s) Resolved

Fixes #20331 

### Affirmation

### DIFF
--- a/.github/scripts/install.sh
+++ b/.github/scripts/install.sh
@@ -19,6 +19,9 @@ if [ "$USE_CONDA" = "true" ]; then
 
     # Install dependencies per operating system
     if [ "$OS" = "win" ]; then
+        # This is necessary for our tests related to conda envs to pass since
+        # the release of Mamba 1.1.0
+        mamba init
         mamba env update --file requirements/windows.yml
     elif [ "$OS" = "macos" ]; then
         mamba env update --file requirements/macos.yml
--- a/.github/workflows/test-win.yml
+++ b/.github/workflows/test-win.yml
@@ -110,6 +110,14 @@ jobs:
         run: |
           conda info
           conda list
+      # Notes:
+      # 1. This only works for conda, probably because it has the necessary
+      #    MSYS2 packages to create the connection.
+      # 2. Check https://github.com/marketplace/actions/debugging-with-tmate for
+      #    usage.
+      #- name: Setup remote ssh connection
+      #  if: env.USE_CONDA == 'true'
+      #  uses: mxschmitt/action-tmate@v3
       - name: Run manifest checks
         if: env.RUN_BUILD == 'true'
         shell: bash -l {0}
--- a/external-deps/python-lsp-server/.gitrepo
+++ b/external-deps/python-lsp-server/.gitrepo
@@ -4,9 +4,9 @@
 ; git-subrepo command. See https://github.com/git-commands/git-subrepo#readme
 ;
 [subrepo]
-	remote = https://github.com/python-lsp/python-lsp-server.git
-	branch = develop
-	commit = ce913f16359f03ffe4a60b6a2fe7ccae4b15bb4a
-	parent = 154269962fba47cd4d2360307e4a25d743e05c30
+	remote = https://github.com/ccordoba12/python-lsp-server.git
+	branch = improve-file-completions
+	commit = dbf75cf19084ab6dbe5e86aa2770b5ca23555c52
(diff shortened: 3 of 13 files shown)


ccordoba12:
@mrclary, the conda-based installer for Mac is failing (I don't understand why). Could you check at that? Thanks!

ccordoba12:
Great! Thanks a lot for your help with that 👍🏽

mrclary:
@ccordoba12, looks like work I already had in the pipeline (PR #20319) will fix the failing conda-based installers.

ccordoba12:
Ok, great news! I'll review it right away so we can include it in 5.4.2 then.

dalthviz (commented):
Thanks @ccordoba12 ! Checking this locally the only thing I manage to see that I'm not totally sure if it is correct is that the dot completion for filenames gives back the filename again instead of the extension for filename completions:

completion

Also, I left a suggestion in a comment for a typo and as we discussed, this needs a rebase to get the fixes done to the conda-based installers workflows

ccordoba12:
> Checking this locally the only thing I manage to see that I'm not totally sure if it is correct is that the dot completion for filenames gives back the filename again instead of the extension for filename completions

Good catch @dalthviz! I fixed that use case in my last commit and added a test for it.

dalthviz (commented):
Checked again but seems like there are still some cases not being correctly handle like a double dot for files:

completion2

ccordoba12:
@mrclary, I don't understand why the conda-based installers started to fail again. Could you take a look at that?

mrclary:
> @mrclary, I don't understand why the conda-based installers started to fail again. Could you take a look at that?

So, the real issue is that I should be using `git diff` instead of `git format-patch` to generate the patch. I'll submit a PR for the change. I don't know why this only plagued Linux and not macOS... 🤷🏼‍♂️

mrclary:
@ccordoba12 #20376 should resolve the failing conda-installers

mrclary:
Actually, that resolved the initial patch issue when building Spyder. But now I see a package conflict error when building the installer...

mrclary:
Nevermind, I was normalizing the version, which apparently is not good.
So, all should be good now.

ccordoba12:
@dalthviz, this should be ready for another review. I fixed the last problem you reported, fixed several issues with file completions next to path separators (i.e. `os.sep`), and added `os.sep` at the end of directory completions in the pylsp to ease introducing additional file completions.

dalthviz:
Checked again and the last issue is fixed :+1:, however, with further testing, there is still a strange behavior for the completion of files names:

completion3

ccordoba12:
> however, with further testing, there is still a strange behavior for the completion of files names

Yeah, I found that too while testing but there's no easy solution for it. The thing is we select the word under the cursor (using Qt) and send it to pylsp for completions. But in this case:

    part.0.parq<Tab>

the selected word is `parq` because Qt considers that a word goes up to a dot. And pylsp can find no completions for it.

To solve this, we should override Qt and extend the search for current words to go up to `part` (the file name). But that's not so simple and it could introduce more corner cases and regressions. So, I think for now this is the best we can do to provide good enough file completions.

dalthviz (approved):

luistrotta19:
Buenas tardes, qusiera que spyder proporcione detalles de los metodos al colocar el cursor encima del objeto, como lo hace VSC, antes se podía instalar kite, pero ahora no está disponible, hay algún sustituto?, gracias, saludos

PR #20364, opened 2023-01-12
Convert enums to Qt6 style

oscargus:
## Description of Changes

In Qt6 the enums are no longer located directly under QtCore.Qt. This is fully compatible with Qt5, so a small step towards easier Qt6 support.

This would convert most of them (probably some are still missing).

I went with separate commits to make it easier to review, but let me know if it is preferred to squash them.

(I do not understand what happened with `scripts/spyder.bat`. I did my best to revert the "change" locally, but for unknown reasons, my git client thought that it was modified...)

### Affirmation

oscargus/Oscar Gustafsson

### DIFF
--- a/external-deps/qtconsole/qtconsole/console_widget.py
+++ b/external-deps/qtconsole/qtconsole/console_widget.py
@@ -253,9 +253,9 @@ class ConsoleWidget(MetaQObjectHasTraits('NewBase', (LoggingConfigurable, superQ
         if self.paging in ('hsplit', 'vsplit'):
             self._splitter = QtWidgets.QSplitter()
             if self.paging == 'hsplit':
-                self._splitter.setOrientation(QtCore.Qt.Horizontal)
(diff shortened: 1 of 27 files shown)


oscargus:
I realize that I should have kept `external-deps` out of it... Problem with automatic search and replace...

dalthviz:
Checking I think QtPy should be taking care of this with the enums promotion to allow unscoped access even with PyQt6 at https://github.com/spyder-ide/qtpy/blob/a94438f50942c23031776d8698883661eeb8e954/qtpy/QtCore.py#L70-L75

Thinking, maybe this be related with https://github.com/spyder-ide/qtpy/issues/364 ? From what I have been able to check seems like PyQt6 is the only binding that dropped support for unscoped enum access. Maybe @oscargus do you have more info about the state of that or after which version the Qt6/PySide6 projects will be dropping support for unscoped enums access?

PR #20376, opened 2023-01-15
PR: Fix patch for conda-based installers

mrclary:
## Description of Changes

* Use `git diff` instead of `git format-patch`. This is more reliable on Linux and is the correct way to do it.
* ~~Do not normalize the tag name. This prevents `setuptools_scm` from replacing 'pre' with 'rc' in the version string.~~

### DIFF
--- a/installers-conda/build_conda_pkgs.py
+++ b/installers-conda/build_conda_pkgs.py
@@ -200,8 +200,8 @@ class SpyderCondaPkg(BuildCondaPkg):
     def _patch_source(self):
         self.logger.info("Creating Spyder source patch...")
 
-        patch = self.repo.git.format_patch(
-            "..origin/installers-conda-patch", "--stdout", "-U3"
+        patch = self.repo.git.diff(
+            "...origin/installers-conda-patch"
         )
         # newline keyword is not added to pathlib until Python>=3.10,
         # so we must use open to ensure LF on Windows


mrclary:
I noticed in separate testing that when pushing our release candidate tags, e.g. `v5.4.2pre1`, `setuptools_scm` designated the version as `v5.4.2rc1`. I thought that not "normalizing" the version number would fix this, which it did, but this caused version conflict errors for `constructor`. Apparently, `pre` may not be acceptable for semantic versioning and is the reason that `packaging.version.Version` replaces `'pre'` with `'rc'`. 

So, just note that while our tags say `'pre'`, the actual package versions will say `'rc'`.

ccordoba12 (approved):
Thanks @mrclary!

ccordoba12:
> Apparently, pre may not be acceptable for semantic versioning and is the reason that packaging.version.Version replaces 'pre' with 'rc'

Well, `pre` and `rc` are equivalent according to [PEP 440](https://peps.python.org/pep-0440/#pre-release-spelling), but this is a bit confusing. What if we revert back to using `rc` tags instead? For actual release candidate releases we'd have to use a different tag, I'd say.

mrclary:
I see; this is confusing then. If `packaging` normalizes (default) then specifying `pre` results in `rc` and conda has no problems solving with that spec. I suppose this is consistent with PEP 440 in that `packaging` is "translating" pre to rc. However, if `packaging` is directed to not normalize, then `pre` remains but conda fails to solve with that spec. So I suppose that conda does not recognize `pre` as equivalent to `rc`.

I suppose we can discuss exactly what we want to do at a developer meeting. However, I would point out that we don't need an actual release candidate release, just a tag. In other words, we don't need a release candidate to show up on the releases page; pushing the appropriately named tag will trigger the workflows to run _as if_ it were a release, but no release is created.

ccordoba12:
> However, I would point out that we don't need an actual release candidate release, just a tag

Ok, that's a good point. So you're saying that we'd need to set a version that is PEP 440-compatible in `spyder/__init__.py` for conda to recognize it, but we could use the tag we want to trigger the installer workflows for the release, right?

PR #20421, opened 2023-01-23
PR: Add menu to use specific environment interpreter for a new console instance

jsbautista:
## Description of Changes

Add menu to launch a console using the intepreter of a specific environments:

* [x] Adds a new submenu to launch consoles using a specific interpreter:

submenu

submenu_entry

* [x] initial env name handling in the console tab;
     * Adds the environment name instead of `Console` for consoles requested to be created using a specific env interpreter from the new submenu (following the current approach the special consoles use):

imagen

* [x] Updates some text of labels of the IPython Console preference page

### Issue(s) Resolved

Part of https://github.com/spyder-ide/ux-improvements/issues/10

### Affirmation

@jsbautista


jsbautista:
Preview of the changes made.
Menu Console tabs:
MenuTabsConsoles

Console Tabs:
TabsConsolesCaptura de pantalla 2023-02-01 194713
TabsConsoles2

Preferences:
image

jsbautista:
Preview of the changes made:
ConsolesSpyder

jsbautista:
Hi @dalthviz

mrclary:
@jsbautista, perhaps I don't remember a lot of the discussion in spyder-ide/ux-improvements#10, but would it be better to remove "Console X" from the tab text? It seems to me to be unnecessary.

Nevertheless, the work you've done on this looks real good to me, thanks!

ccordoba12:
> but would it be better to remove "Console X" from the tab text? It seems to me to be unnecessary.

That needs additional discussion because it doesn't seem so simple to me (I'm not going to describe my reasoning here to not derail @jsbautista's work).

So @mrclary, could you post your comment on https://github.com/spyder-ide/ux-improvements/issues/10 so we can keep talking about it there? Thanks!

spyder/plugins/ipythonconsole/utils/kernelspec.py
```diff
@@ -99,12 +99,14 @@ class SpyderKernelSpec(KernelSpec, SpyderConfigurationAccessor):
     CONF_SECTION = 'ipython_console'
 
     def __init__(self, is_cython=False, is_pylab=False,
-                 is_sympy=False, **kwargs):
+                 is_sympy=False, is_environment=False,
+                 path_to_environment='', **kwargs):
```
dalthviz, line 103: Maybe we could change these two `kwargs` to use a single one like `path_to_custom_interpreter=None`?

spyder/plugins/ipythonconsole/utils/kernelspec.py
```diff
@@ -113,10 +115,18 @@ def __init__(self, is_cython=False, is_pylab=False,
     def argv(self):
         """Command to start kernels"""
         # Python interpreter used to start kernels
+        if self.is_environment:
+            self.set_conf(
+                'executable', self.path_to_environment,
+                section='main_interpreter')
+            self.set_conf('default', False, section='main_interpreter')
+            self.set_conf('custom', True, section='main_interpreter')
```
dalthviz, line 123: I think here we need just to put `pyexec =  self.path_to_environment` (or the revised version of the variable name `self.path_to_custom_interpreter`)?

dalthviz: Or even just remove this validation since the new case is being handled at below line 128?

spyder/plugins/ipythonconsole/utils/kernelspec.py
```diff
@@ -187,7 +197,7 @@ def env(self):
 
         # Environment variables that we need to pass to the kernel
         env_vars.update({
-            'SPY_EXTERNAL_INTERPRETER': not default_interpreter,
+            'SPY_EXTERNAL_INTERPRETER': True,
```
dalthviz, line 200: Probably we should change this to `not default_interpreter or self.is_environment`

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -650,6 +675,90 @@ def update_actions(self):
...
+            environment_consoles.append(action)
+        for item in environment_consoles:
+            self.add_item_to_menu(
+                item,
+                menu=self.console_environment_menu
+            )
+        self.console_environment_menu._render()
+        
+        
+
```
dalthviz, line 761: Remove some extra empty lines:

```suggestion
```

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -1269,9 +1378,10 @@ def config_options(self):
         cfg._merge(spy_cfg)
         return cfg
 
-    def interpreter_versions(self):
+    def interpreter_versions(self, is_environment=False, env=''):
```
dalthviz, line 1381: I think we could use here also just one `kwarg` -> `path_to_custom_interpreter=None`

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -1363,24 +1477,26 @@ def create_new_client(self, give_focus=True, filename='', is_cython=False,
...
             give_focus=give_focus,
             handlers=self.registered_spyder_kernel_handlers,
             initial_cwd=initial_cwd,
         )
-
         # Add client to widget
         self.add_tab(
             client, name=client.get_name(), filename=filename,
             give_focus=give_focus)
-
```
dalthviz, line 1378: Please restore this blank line

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -1363,24 +1477,26 @@ def create_new_client(self, give_focus=True, filename='', is_cython=False,
...
         # Create new kernel
+
```
dalthviz, line 1493: Remove extra blank line:

```suggestion
```

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -1528,7 +1649,7 @@ def register_client(self, client):
 
         # For help requests
         control.sig_help_requested.connect(self.sig_help_requested)
-
+
```
dalthviz, line 1652: Remove trailing space:

```suggestion

```

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -650,6 +675,90 @@ def update_actions(self):
...
+        self.envs.update(**output)
+        for env in list(self.envs.keys()):
+            path, version = self.envs[env]
+            # Save paths in lowercase on Windows to avoid issues with
+            # capitalization.
+            path = path.lower() if os.name == 'nt' else path
+            self.path_to_env[path] = env
+
+    def update_environment_menu(self):
+        """Update context menu information."""
```
dalthviz, line 735: Probably this should mention that is the submenu related with consoles launched for a specific interpreter. So something like:
```suggestion
        """Update context menu entries to select specific interpreter to launch a console."""
```

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -650,6 +675,90 @@ def update_actions(self):
...
+
+    def update_envs(self, worker, output, error):
+        """Update the list of environments in the system."""
+        self.envs.update(**output)
+        for env in list(self.envs.keys()):
+            path, version = self.envs[env]
+            # Save paths in lowercase on Windows to avoid issues with
+            # capitalization.
+            path = path.lower() if os.name == 'nt' else path
+            self.path_to_env[path] = env
```
dalthviz, line 732: I think all this code above comes from the status widget of the mainterpreter plugin, right? I think we should try to refactor it so we don't duplicate it here and there. We could expose this code in the mainterpreter plugin API to be used in other plugins like the IPython console or it should go to some util module like `spyder.utils.envs` so both plugins can import it.

dalthviz (commented):
Thanks @jsbautista for the work here! Left an initial review regarding the kwargs passed to the methods (which maybe could be simplified to just use one) and some other questions regarding possible code refactoring. 

If you have any question let me know!

spyder/plugins/ipythonconsole/utils/kernelspec.py
```diff
@@ -187,7 +191,7 @@ def env(self):
 
         # Environment variables that we need to pass to the kernel
         env_vars.update({
-            'SPY_EXTERNAL_INTERPRETER': not default_interpreter,
+            'SPY_EXTERNAL_INTERPRETER': self.path_to_custom_interpreter,
```
dalthviz, line 194: I think this needs to be `not default_interpreter or self.path_to_custom_interpreter` (so when a custom interpreter is selected from the preferences the `SPY_EXTERNAL_INTERPRETER` variable is correctly set since the default interpreter is not being used):

```suggestion
            'SPY_EXTERNAL_INTERPRETER': not default_interpreter or self.path_to_custom_interpreter,
```

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -1346,15 +1422,17 @@ def get_current_shellwidget(self):
...
     def create_new_client(self, give_focus=True, filename='', is_cython=False,
                           is_pylab=False, is_sympy=False, given_name=None,
-                          cache=True, initial_cwd=None):
+                          cache=True, initial_cwd=None, environment='A',
+                          path_to_custom_interpreter=None):
         """Create a new client"""
         self.master_clients += 1
         client_id = dict(int_id=str(self.master_clients),
-                         str_id='A')
+                         str_id=environment)
```
dalthviz, line 1435: I was checking this and I think the tab name generation should keep the `A` id untouched (that is used to identify multiple consoles using the same kernel) and instead add a new element to the tab name and only with the env name. So for example an env that appears as `conda: my_env (Python 3.9.16)` in the console menu should generate as tab name `Console X (my_env)/A` where only `Console X` is editable text. We should discuss if we should remove or not the `Console` text but that is to be done in the issue (I think it make sense to remove the `Console` text but we need to check).

For the above probably we will need to add a new key to the `clien_id` dict like `str_env_name` which then will be used to create the tab name at: https://github.com/spyder-ide/spyder/blob/dd528385e9bbe589f9e033e59e5f2f33479ee811/spyder/plugins/ipythonconsole/widgets/client.py#L510-L526

There the client `self.id_` attribute will contain the `str_env_name` key

mrclary: @dalthviz, I think that `A` should be removed. I don't see a compelling reason to _distinguish_ between consoles using the same kernel. I see a more compelling case to _identify_ consoles that are using the _same_ kernel. However, currently in Spyder, opening another console will start another kernel, yet both consoles will indicate `/A` implying that both are using the same kernel when they aren't.

It seems to me to be less confusing to just get rid of `A` altogether and in the future determine how to identify consoles that are using the _same_ kernel.

dalthviz: Thanks for the comment @mrclary ! We need to discuss that for sure as well as the usage of `Console`. I will add a comment on the issue to track your new comment regarding the `str_id` and the option I described here on how the console name should be formatted :+1:

dalthviz:
@jsbautista checking the failing test I think we need to add also the `str_env_name` dict key around here:

https://github.com/spyder-ide/spyder/blob/ef84f78ed10a506e3d915c573bfcbf07c968d525/spyder/plugins/ipythonconsole/widgets/main_widget.py#L1501-L1511

So this code should become something like:

```python
            # Set full client name
            client_id = dict(int_id=master_client.id_['int_id'],
                             str_id=chr(slave_ord + 1),
                             str_env_name=master_client.id_['str_env_name'])
        else:
            # If we couldn't find a client with the same connection file,
            # it means this is a new master client
            self.master_clients += 1

            # Set full client name
            client_id = dict(int_id=str(self.master_clients), str_id='A',
                             str_env_name="Unknown")
```

dalthviz:
@ccordoba12 @mrclary could you give this a check? I think is ready for a review. Although the scope of this is not to do changes in the way the consoles are named, I added some changes to the implementation here to follow the way special consoles naming work (adding instead of `Pylab`, `SymPy` or `Cython` the name of the env when using the new menu to launch a console).

I think the changes regarding how in general the names of the consoles work should be done in a different PR since the main scope of this is to add a new menu from where you can launch a new console from a specific environment.

Let me know what you think!

spyder/plugins/ipythonconsole/utils/kernelspec.py
```diff
@@ -186,7 +190,7 @@ def env(self):
 
         # Environment variables that we need to pass to the kernel
         env_vars.update({
-            'SPY_EXTERNAL_INTERPRETER': not default_interpreter,
+            'SPY_EXTERNAL_INTERPRETER': not default_interpreter or self.path_to_custom_interpreter,
```
ccordoba12, line 193:
```suggestion
            'SPY_EXTERNAL_INTERPRETER': (not default_interpreter
                or self.path_to_custom_interpreter),
```

This makes this line fit in 79 columns.

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -46,6 +46,8 @@
 from spyder.widgets.browser import FrameWebView
 from spyder.widgets.findreplace import FindReplace
 from spyder.widgets.tabs import Tabs
+from spyder.utils.workers import WorkerManager
+from spyder.utils.envs import get_list_envs
```
ccordoba12, line 50: Please move these imports a bit above so that they are organized in alphabetical order.

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -488,12 +500,21 @@ def setup(self):
 
         # --- Setting options menu
         options_menu = self.get_options_menu()
+
+        self.console_environment_menu = self.create_menu(
+            IPythonConsoleWidgetOptionsMenus.EnvironmentConsoles,
+            _('New console in environment'))
```
ccordoba12, line 506:
```suggestion
            _('New console in environment')
        )
```

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -488,12 +500,21 @@ def setup(self):
...
+        self.console_environment_menu.setStyleSheet(
+            "QMenu { menu-scrollable: 1; }"
+        )
```
ccordoba12, line 509: Please use `qstylizer` for this (you can find examples in `utils/stylesheet.py`). We prefer not to use css as docstrings because they are hard to modify/maintain.

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -541,7 +562,8 @@ def setup(self):
             icon=self.create_icon('ipython_console'),
             triggered=self.create_cython_client,
         )
-
+        self.console_environment_menu.aboutToShow.connect(
+            self.update_environment_menu)
```
ccordoba12, line 566:
```suggestion

        self.console_environment_menu.aboutToShow.connect(
            self.update_environment_menu)

```

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -650,6 +673,48 @@ def update_actions(self):
...
+        Get the list of environments/interpreters in a worker.
+        """
+        self._worker_manager.terminate_all()
+        worker = self._worker_manager.create_python_worker(get_list_envs)
+        worker.sig_finished.connect(self.update_envs)
+        worker.start()
+
+    def update_envs(self, worker, output, error):
+        """Update the list of environments in the system."""
+        self.envs.update(**output)
```
ccordoba12, line 687:
```suggestion
        if output is not None:
            self.envs.update(**output)
```

See issue #20539.

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -1269,9 +1334,10 @@ def config_options(self):
         cfg._merge(spy_cfg)
         return cfg
 
-    def interpreter_versions(self):
+    def interpreter_versions(self, path_to_custom_interpreter=None):
         """Python and IPython versions used by clients"""
-        if self.get_conf('default', section='main_interpreter'):
+        if (self.get_conf('default', section='main_interpreter')
+           and not path_to_custom_interpreter):
```
ccordoba12, line 1340:
```suggestion
        if (
            self.get_conf('default', section='main_interpreter')
            and not path_to_custom_interpreter
        ):
```

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -1481,6 +1554,15 @@ def create_cython_client(self):
         """Force creation of Cython client"""
         self.create_new_client(is_cython=True, given_name="Cython")
 
+    def create_environment_client(
+            self, environment, path_to_custom_interpreter
+            ):
+        """Force creation of Environment client."""
```
ccordoba12, line 1560:
```suggestion
    def create_environment_client(
        self, environment, path_to_custom_interpreter
    ):
        """Create a client for a Python environment."""
```

ccordoba12 (commented):
Thanks @jsbautista and @dalthviz for your great work here! I left some small suggestions and comments for you, the rest looks good to me.

mrclary (approved):
I don't have any additional comments beyond @ccordoba12's comments. Looks good to me.

ccordoba12 (approved):
Looks good to me now, thanks @jsbautista and @dalthviz!

@dalthviz, I'll leave the merge to you in case you want to do a squash-merge.

PR #20425, opened 2023-01-23
PR: Better split test suite between slow and fast slots in CIs and other improvements for CIs

ccordoba12:
## Description of Changes

- Remove `slow` Pytest marker and instead mark all main window tests, and a percentage of the IPython console ones on Linux and Mac, as slow in CIs (locally there won't be any distinction between slow and fast tests).

    This will make our slow slots finish faster compared to the fast ones (the first ones are taking two times more than the later).

    This also allow us to adjust the amount of tests marked as slow/fast per operating system.

- Use micromamba to create our test environment on CIs. This significantly reduces the necessary time to do that.
- Remove script and code in our workflows that manually checked for `[ci skip]` in commit messages. That functionality is provided natively now by Github actions.

### Affirmation


ccordoba12:
Pinging @juliangilbey about this one because you could interested on this change.

PR #20466, opened 2023-02-03
PR: Fix installer triggers

mrclary:
Opening a PR and labeling it at the same time results in a race condition between the `opened` and `labeled` `pull_request` event triggers which can lead to cancelled installer workflows.

To fix this, we remove the `labeled` event trigger for `pull_requests`.
Since release candidates were the primary motivation for these triggers, but this must rely on the `workflow_dispatch` trigger anyway, the `workflow_dispatch` is now the recommended way to trigger workflows outside of `pull_request` and `release` events.

Additionally, `inputs` for the `workflow_dispatch` were added to the conda-based installer workflow for enhanced flexibility.

* `pull_request.labeled` trigger events are removed for installer workflows
* `push.tag` trigger events are removed for installer workflows
* `workflow_dispatch` trigger is added to unit test workflows. These can now be triggered to run on arbitrary branches, independent of pull_requests.
* `inputs` are added to the `workflow_dispatch` in conda-based installer workflow
  * "Build as release candidate": build the installers using released versions of `qtconsole`, `python-lsp-server`, and `spyder-kernels`; codesign and notarize the resulting installers (macOS only). If deselected, `qtconsole`, `python-lsp-server`, and `spyder-kernels` are built from the upstream remote and commit specified in `external-deps`
  * "Build macOS installer": build the macOS installer
  * "Build Linux installer": build the Linux installer
  * "Build Windows installer": build the Windows installer

### DIFF
--- a/.github/workflows/installer-macos.yml
+++ b/.github/workflows/installer-macos.yml
@@ -1,10 +1,5 @@
 on:
   pull_request:
-    types:
-      - opened
-      - reopened
-      - synchronize
-      - labeled
     paths:
       - 'installers/macOS/**'
       - '.github/workflows/installer-macos.yml'
@@ -23,10 +18,6 @@ on:
     types:
       - created
 
-  push:
-    tags:
-      - v*pre*
-
   workflow_dispatch:
 
 concurrency:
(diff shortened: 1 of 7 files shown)


ccordoba12 (approved):

PR #20471, opened 2023-02-03
PR: Update conda-based installers

mrclary:
## Description of Changes

* Added `StartupWMClass` to Linux desktop file. This obviates the need to change the name of the desktop file in the post-install script.
* Minor CI workflow updates

### DIFF
--- a/.github/workflows/installers-conda.yml
+++ b/.github/workflows/installers-conda.yml
@@ -184,7 +184,7 @@ jobs:
           [[ -d $ARTIFACTS_PATH ]] || mkdir $ARTIFACTS_PATH
           [[ -d $CONDA_BLD_PATH ]] || mkdir $CONDA_BLD_PATH
 
-          env
+          env | sort
 
       - name: Download Local Conda Packages
         if: needs.build-noarch-pkgs.result == 'success'
--- a/.github/workflows/installers-conda.yml
+++ b/.github/workflows/installers-conda.yml
@@ -172,6 +172,8 @@ jobs:
         with:
           environment-file: installers-conda/build-environment.yml
           extra-specs: python=${{ matrix.python-version }}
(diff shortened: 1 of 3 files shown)


mrclary:
@dalthviz @ccordoba12 I'm having trouble getting the Windows test to work properly; perhaps you guys have some insight.

The following should execute the `exe` installer from the `bash` command line in silent mode:
```bash
$ cmd.exe /C start /wait "<path to exe>" /S
```
I've successfully tested on my Windows VM (with bash shell), but can't get it working in CI. From the logs it appears that `cmd.exe` is executed but the executable path is not resolved properly: the `cmd.exe` banner is printed but the installer does not run.

ccordoba12:
> The following should execute the exe installer from the bash command line in silent mode

I can see that you're not using a Bash shell to run the installer tests on Windows (i.e. there's no `shell: bash` in the `Test Installer Application` step). And I think the default shell used by Github Actions on Windows is `cmd`.

mrclary:
> I can see that you're not using a Bash shell to run the installer tests on Windows (i.e. there's no `shell: bash` in the `Test Installer Application` step). And I think the default shell used by Github Actions on Windows is `cmd`.

Hmm. For the `build-installers` job I am using a default shell `defaults: run: shell: bash -l {0}`, and the `Test Application Bundle` step is a bash script. And in the same step I've confirmed `[[ -e "$PKG_FILE" ]] && echo "Installer exists" || echo "Installer does not exist"` successfully reports that the `$PKG_FILE` exists while running in Windows.

ccordoba12:
Ok, sorry for the noise.

mrclary:
> Ok, sorry for the noise.

No worries, thanks for looking at it.

ccordoba12 (approved):

PR #20498, opened 2023-02-07
PR: Update conda search paths to include mambaforge and miniforge

mrclary:
## Description of Changes

* Fix priority for `pyenv`
* Add `mambaforge` and `miniforge`
* Reprioritize `anaconda` and `miniconda`

### DIFF
--- a/spyder/utils/programs.py
+++ b/spyder/utils/programs.py
@@ -82,24 +82,11 @@ def is_program_installed(basename):
     """
     home = get_home_dir()
     req_paths = []
-    if sys.platform == 'darwin':
-        if basename.endswith('.app') and osp.exists(basename):
-            return basename
+    if (sys.platform == 'darwin' and basename.endswith('.app')
+            and osp.exists(basename)):
+        return basename
 
-        pyenv = [
-            osp.join(home, '.pyenv', 'bin'),
-            osp.join('/usr', 'local', 'bin'),
-        ]
-
-        # Prioritize Anaconda before Miniconda; local before global.
-        a = [osp.join(home, 'opt'), '/opt']
-        b = ['mambaforge', 'miniforge',
-             'miniconda3', 'anaconda3', 'miniconda', 'anaconda']
-        conda = [osp.join(*p, 'condabin') for p in itertools.product(a, b)]
-
-        req_paths.extend(pyenv + conda)
-
-    elif sys.platform.startswith('linux'):
+    if os.name == 'posix':
         pyenv = [
             osp.join(home, '.pyenv', 'bin'),
             osp.join('/usr', 'local', 'bin'),
@@ -108,19 +95,15 @@ def is_program_installed(basename):
         a = [home, '/opt']
         b = ['mambaforge', 'miniforge',
              'miniconda3', 'anaconda3', 'miniconda', 'anaconda']
-        conda = [osp.join(*p, 'condabin') for p in itertools.product(a, b)]
-
-        req_paths.extend(pyenv + conda)
-
-    elif os.name == 'nt':
+    else:
         pyenv = [osp.join(home, '.pyenv', 'pyenv-win', 'bin')]
 
         a = [home, 'C:\\', osp.join('C:\\', 'ProgramData')]
         b = ['Mambaforge', 'Miniforge',
              'Miniconda3', 'Anaconda3', 'Miniconda', 'Anaconda']
-        conda = [osp.join(*p, 'condabin') for p in itertools.product(a, b)]
 
-        req_paths.extend(pyenv + conda)
+    conda = [osp.join(*p, 'condabin') for p in itertools.product(a, b)]
+    req_paths.extend(pyenv + conda)
 
     for path in os.environ['PATH'].split(os.pathsep) + req_paths:
         abspath = osp.join(path, basename)
--- a/spyder/utils/programs.py
+++ b/spyder/utils/programs.py
@@ -93,7 +93,8 @@ def is_program_installed(basename):
 
         # Prioritize Anaconda before Miniconda; local before global.
         a = [osp.join(home, 'opt'), '/opt']
-        b = ['anaconda', 'miniconda', 'anaconda3', 'miniconda3']
+        b = ['mambaforge', 'miniforge',
+             'miniconda3', 'anaconda3', 'miniconda', 'anaconda']
         conda = [osp.join(*p, 'condabin') for p in itertools.product(a, b)]
 
         req_paths.extend(pyenv + conda)
@@ -105,7 +106,8 @@ def is_program_installed(basename):
         ]
 
         a = [home, '/opt']
-        b = ['anaconda', 'miniconda', 'anaconda3', 'miniconda3']
+        b = ['mambaforge', 'miniforge',
+             'miniconda3', 'anaconda3', 'miniconda', 'anaconda']
         conda = [osp.join(*p, 'condabin') for p in itertools.product(a, b)]
 
         req_paths.extend(pyenv + conda)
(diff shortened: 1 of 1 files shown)


spyder/utils/programs.py
```diff
@@ -82,42 +82,28 @@ def is_program_installed(basename):
...
 
         a = [home, '/opt']
-        b = ['anaconda', 'miniconda', 'anaconda3', 'miniconda3']
-        conda = [osp.join(*p, 'condabin') for p in itertools.product(a, b)]
-
-        req_paths.extend(pyenv + conda)
-
-    elif os.name == 'nt':
+        b = ['mambaforge', 'miniforge',
+             'miniconda3', 'anaconda3', 'miniconda', 'anaconda']
```
CAM-Gerlach, line 97:
```suggestion
        b = ['mambaforge', 'miniforge3', 'miniforge',
             'miniconda3', 'anaconda3', 'miniconda', 'anaconda']
```

AFAIK, Python 3 Miniforge (but not Mambaforge) is installed to `miniforge3/` rather than `miniforge/` on *nix.

spyder/utils/programs.py
```diff
@@ -82,42 +82,28 @@ def is_program_installed(basename):
...
+    else:
         pyenv = [osp.join(home, '.pyenv', 'pyenv-win', 'bin')]
 
         a = [home, 'C:\\', osp.join('C:\\', 'ProgramData')]
-        b = ['Anaconda', 'Miniconda', 'Anaconda3', 'Miniconda3']
-        conda = [osp.join(*p, 'condabin') for p in itertools.product(a, b)]
+        b = ['Mambaforge', 'Miniforge',
+             'Miniconda3', 'Anaconda3', 'Miniconda', 'Anaconda']
```
CAM-Gerlach, line 103:
```suggestion
        b = ['Mambaforge', 'Miniforge3', 'Miniforge',
             'Miniconda3', 'Anaconda3', 'Miniconda', 'Anaconda']
```

Same here

spyder/utils/programs.py
```diff
@@ -82,42 +82,28 @@ def is_program_installed(basename):
...
-
-        req_paths.extend(pyenv + conda)
-
-    elif os.name == 'nt':
+        b = ['mambaforge', 'miniforge',
+             'miniconda3', 'anaconda3', 'miniconda', 'anaconda']
+    else:
         pyenv = [osp.join(home, '.pyenv', 'pyenv-win', 'bin')]
 
         a = [home, 'C:\\', osp.join('C:\\', 'ProgramData')]
```
CAM-Gerlach, line 101:
```suggestion
        a = [home, osp.join(home, 'AppData/Local'),
             'C:/', osp.join('C:/', 'ProgramData')]
```

At least Miniforge and Miniconda default to being installed in `~/AppData/Local` instead of `~` on Windows.

CAM-Gerlach (commented):
Some initial comments. I was going to test on my colleague's machine with Miniforge installed, but she's out on a deployment right now and may not be back for a couple days, so I could test it on my machine if needed.

spyder/utils/programs.py
mrclary: Aah yes, you are correct.

spyder/utils/programs.py
mrclary: Mambaforge is in `~\AppData\Local` local as well. Do you think Anaconda is too?
Maybe `home` should be removed?
```suggestion
        a = [osp.join(home, 'AppData/Local'), osp.join('C:/', 'ProgramData')]
```

mrclary: Or can other programs be found in `home` or `C:/`?

CAM-Gerlach: Sorry, I meant Mambaforge and Miniforge. I've seen Miniconda, etc. in `~` quite commonly on Windows, so that should stay. And the drive root is a common location (I use it and have seen others do the same, for instance, since my username includes spaces which causes problems with some stuff, and IIRC some guides do too).

spyder/utils/programs.py
```diff
@@ -82,42 +82,28 @@ def is_program_installed(basename):
...
-            osp.join('/usr', 'local', 'bin'),
-            osp.join(home, '.pyenv', 'bin')
-        ]
-
-        # Prioritize Anaconda before Miniconda; local before global.
-        a = [osp.join(home, 'opt'), '/opt']
-        b = ['anaconda', 'miniconda', 'anaconda3', 'miniconda3']
-        conda = [osp.join(*p, 'condabin') for p in itertools.product(a, b)]
+    if (sys.platform == 'darwin' and basename.endswith('.app')
+            and osp.exists(basename)):
```
ccordoba12, line 86:
```suggestion
    if (
        sys.platform == 'darwin'
        and basename.endswith('.app')
        and osp.exists(basename)
    ):
```

Black style formatting of `if` statements makes them easier to read, in my opinion.

ccordoba12 (commented):
I left a small suggestion for you @mrclary, otherwise looks good to me.

spyder/utils/programs.py
```diff
@@ -82,42 +82,32 @@ def is_program_installed(basename):
...
-
-        req_paths.extend(pyenv + conda)
-
-    elif sys.platform.startswith('linux'):
-        pyenv = [
-            osp.join('/usr', 'local', 'bin'),
-            osp.join(home, '.pyenv', 'bin')
         ]
 
         a = [home, '/opt']
```
ccordoba12, line 98:
```suggestion
        a = [home, osp.join(home, 'opt'), '/opt']
```

This is missing from the removed code above and I think it's a possibility on Mac.

spyder/utils/programs.py
```diff
@@ -82,42 +82,32 @@ def is_program_installed(basename):
...
+        b = ['mambaforge', 'miniforge3', 'miniforge',
+             'miniconda3', 'anaconda3', 'miniconda', 'anaconda']
+    else:
         pyenv = [osp.join(home, '.pyenv', 'pyenv-win', 'bin')]
 
-        a = [home, 'C:\\', osp.join('C:\\', 'ProgramData')]
-        b = ['Anaconda', 'Miniconda', 'Anaconda3', 'Miniconda3']
-        conda = [osp.join(*p, 'condabin') for p in itertools.product(a, b)]
+        a = [home, osp.join(home, 'AppData/Local'),
+             'C:/', osp.join('C:/', 'ProgramData')]
```
ccordoba12, line 105: This would give a weird mix a `/` and `\\` for these paths (due to the usage of `osp.join`). Why not keep using `\\` as before?

spyder/utils/programs.py
mrclary: Okay, I can add that as well. Isn't that a really old standard?

spyder/utils/programs.py
mrclary: @CAM-Gerlach, do you want to comment on this?

spyder/utils/programs.py
CAM-Gerlach: I suppose it doesn't hurt, no? It could still be the case if people first installed a conda-based distribution a long time ago, perhaps (since I know many folks who don't reinstall it for years and years)?

spyder/utils/programs.py
ccordoba12:
```suggestion
        a = [home, osp.join(home, 'AppData', 'Local'),
             'C:\\', osp.join('C:\\', 'ProgramData')]
```

spyder/utils/programs.py
ccordoba12: > Okay, I can add that as well. Isn't that a really old standard?

I've seen people that still use it.

> I suppose it doesn't hurt, no?

Agreed

spyder/utils/programs.py
CAM-Gerlach: Or

```suggestion
        a = [home, osp.join(home, 'AppData', 'Local'),
             r'C:\', osp.join(r'C:\', 'ProgramData')]
```

mrclary: Perhaps this is a bug, but code analysis claims a syntax error with r-strings containing a single `\`

CAM-Gerlach: Oh, right—it's [this issue](https://docs.python.org/3/faq/design.html#why-can-t-raw-strings-r-strings-end-with-a-backslash). So, welp, that doesn't work either (without ugly workarounds).

One more reason to use pathlib :)

ccordoba12 (approved):
Thanks @mrclary!

spyder/utils/programs.py
mrclary: > One more reason to use pathlib :)

I support full conversion to `pathlib` in Spyder 6 😄

PR #20509, opened 2023-02-09
PR: Improve cells support in the Editor

jsbautista:
## Description of Changes

* Add a button in one of main toolbars to create a new cell.
* Add cell divider to also cross the line numbers.

### Issue(s) Resolved

Part of https://github.com/spyder-ide/ux-improvements/issues/30

### Affirmation

@jsbautista


jsbautista:
Hi @dalthviz

spyder/plugins/editor/plugin.py
```diff
@@ -774,6 +774,14 @@ def get_plugin_actions(self):
                                add_shortcut_to_tip=True)
 
         # --- Edit Toolbar ---
+        create_new_cell = create_action(self, _("Create new Cell at "
+                                                "current line"),
+                                        icon=ima.icon('run_selection'),
```
dalthviz, line 779: Lets change the icon for now to the `cell` icon (image):
```suggestion
                                        icon=ima.icon('cell'),
```

spyder/plugins/editor/plugin.py
```diff
@@ -1074,7 +1082,8 @@ def get_plugin_actions(self):
         self.main.run_menu_actions = (
             run_menu_actions + self.main.run_menu_actions)
         run_toolbar_actions = [run_action, run_cell_action,
-                               run_cell_advance_action, run_selected_action]
+                               run_cell_advance_action, run_selected_action,
+                               create_new_cell]
```
dalthviz, line 1086: Let move this new action to the begining of the toolbar actions list for the moment, so you would get something like this: 

image

dalthviz (commented):
Thanks @jsbautista for working on this! Checked locally and seems like is working correctly :+1: Left some initial comments to remove some imports that seem like unused, some code style issues, suggestions for the new action icon and placement.

However, seems like the tests are failing due to a missing shortcut definition. For that you will need to define a shortcut at https://github.com/spyder-ide/spyder/blob/450e50d08e12c96183769875ff164a515491fbad/spyder/config/main.py#L416-L417 (maybe `Ctrl + 2` could work for the moment).

Also could you add a GIF to preview the functionality here?

Thanks!

spyder/plugins/editor/panels/linenumber.py
```diff
@@ -98,6 +99,14 @@ def draw_pixmap(xleft, ytop, pixmap):
         for top, line_number, block in self.editor.visible_blocks:
             data = block.userData()
             if self._markers_margin and data:
+                if is_cell_header(block):
+                    cell_line_color = QColor(Qt.darkGray)
```
dalthviz, line 103: Let us here the color defined at https://github.com/spyder-ide/spyder/pull/20509/files#diff-04bc4dbe5c8af45aeb3dc9d7aef7317a8ae81284b422f34c5e6447c233d54324R37

spyder/plugins/editor/panels/codefolding.py
```diff
@@ -239,6 +240,14 @@ def paintEvent(self, event):
         # on the folding panel.
         super(FoldingPanel, self).paintEvent(event)
         painter = QPainter(self)
+        cell_line_color = QColor(Qt.darkGray)
```
dalthviz, line 243: Let set this as a class attribute like the one in the linenumber panel at: https://github.com/spyder-ide/spyder/pull/20509/files#diff-04bc4dbe5c8af45aeb3dc9d7aef7317a8ae81284b422f34c5e6447c233d54324R37

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -4233,6 +4234,28 @@ def __in_block_comment(cursor):
...
+        cursor.movePosition(QTextCursor.StartOfBlock)
+        cursor.setPosition(end_pos)
+        cursor.movePosition(QTextCursor.EndOfBlock)
+        cursor.insertText(endline)
+        cursor.setPosition(start_pos)
+        cursor.movePosition(QTextCursor.StartOfBlock)
+        cursor.insertText(firstline)
+        cursor.endEditBlock()
+        
+        return None
```
dalthviz, line 4258: I think we can remove this `return` (returning `None` is done even without an explicit `return` statement):
```suggestion
```

spyder/plugins/editor/plugin.py
```diff
@@ -2729,6 +2739,12 @@ def __move_cursor_position(self, index_move):
         self.__ignore_cursor_history = False
         self.update_cursorpos_actions()
 
+    @Slot()
+    def create_cell(self, prefix=None):
```
dalthviz, line 2743: Is this `prefix` kwarg used for something? If not lets remove it

dalthviz (commented):
Hi again @jsbautista , left a couple of new comments/suggestions but other than that this LGTM :+1:

I think the failing test job is unrelated to the changes here but lets check if it fails again after applying the changes for the new review

jsbautista:
Preview 
Cells

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -4233,6 +4234,27 @@ def __in_block_comment(cursor):
         cursor3.endEditBlock()
         return True
 
+    def create_new_cell(self):
+        firstline = '#%% Start Cell \r\n'
```
ccordoba12, line 4238:
```suggestion
        firstline = '#%% ' + self.get_line_separator()
```

* I like a lot the idea of enclosing selections between cell separators. But I don't think adding the extra `Star cell` is appropriate because the start of a cell is the end of the previous one. So that could introduce a bit of confusion for users because they could infer that the region before `Start cell` is not really a cell, when we know it is.
* This assumes that the line endings of all files are `\r\n`, but we can also handle `\n` (Linux) and `\r` (Mac). So, we need to use the method `get_line_separator` instead.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -4233,6 +4234,27 @@ def __in_block_comment(cursor):
...
+        endline = '\r\n'
```
ccordoba12, line 4239:
```suggestion
        endline = self.get_line_separator()
```

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -4233,6 +4234,27 @@ def __in_block_comment(cursor):
...
+        cursor = self.textCursor()
+        if self.has_selected_text():
+            self.extend_selection_to_complete_lines()
+            start_pos, end_pos = cursor.selectionStart(), cursor.selectionEnd()
+            endline = '\r\n#%% End Cell'
```
ccordoba12, line 4244:
```suggestion
            endline = self.get_line_separator() + '#%%'
```

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -4233,6 +4234,27 @@ def __in_block_comment(cursor):
...
+        else:
+            start_pos = end_pos = cursor.position()
+        cursor.beginEditBlock()
```
ccordoba12, line 4247:
```suggestion

        # Add cell comment or enclose current selection in cells
        cursor.beginEditBlock()
```

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -4233,6 +4234,27 @@ def __in_block_comment(cursor):
...
+        cursor.setPosition(start_pos)
+        cursor.movePosition(QTextCursor.StartOfBlock)
```
ccordoba12, line 4249:
```suggestion
```

These two instructions don't seem necessary because they are repeated below. However, I'm not really sure about it, so @jsbautista please check that before applying my suggestion.

ccordoba12 (commented):
Thanks for your work on this @jsbautista! It's really nice improvement!

ccordoba12:
@jsbautista, I think you need to merge with `master` to get make our Linux tests pass.

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -4233,6 +4234,27 @@ def __in_block_comment(cursor):
         cursor3.endEditBlock()
         return True
 
+    def create_new_cell(self):
+        firstline = '#%% ' + self.get_line_separator()
```
dalthviz, line 4238: Following comment at https://github.com/spyder-ide/ux-improvements/issues/30#issuecomment-914371091 and to prevent a trailing white space warning this could be

```suggestion
        firstline = '# %%' + self.get_line_separator()
```

What do you think @ccordoba12 ?

spyder/plugins/editor/widgets/codeeditor.py
```diff
@@ -4233,6 +4234,27 @@ def __in_block_comment(cursor):
...
+        endline = self.get_line_separator()
+        cursor = self.textCursor()
+        if self.has_selected_text():
+            self.extend_selection_to_complete_lines()
+            start_pos, end_pos = cursor.selectionStart(), cursor.selectionEnd()
+            endline = self.get_line_separator() + '#%%'
```
dalthviz, line 4244: if the suggestion above is correct then we should change this too:

```suggestion
            endline = self.get_line_separator() + '# %%'
```

spyder/plugins/editor/widgets/codeeditor.py
ccordoba12: Agreed, thanks for spotting that @dalthviz!

ccordoba12:
One last thing to consider, but perhaps for a follow-up PR: should we try to move the code that paints cell dividers on top of panels to the paintEvent method of the base Panel class?

@jsbautista, @dalthviz, I don't know if you considered that before, but it could be worth the try, don't you think?

dalthviz:
> One last thing to consider, but perhaps for a follow-up PR: should we try to move the code that paints cell dividers on top of panels to the paintEvent method of the base Panel class?

I think that could be a good idea. In that way if more panels are added then they will automatically add the cell line  :+1: Probably we will need to validate `self.position` to only apply the code to paint the cell line when the panel position is on the `LEFT` right @ccordoba12 ?

ccordoba12:
> In that way if more panels are added then they will automatically add the cell line +1 

Yep, that was my reasoning too.

> Probably we will need to validate self.position to only apply the code to paint the cell line when the panel position is on the LEFT right @ccordoba12 ?

I'd say `LEFT` and `RIGHT`, just to be safe (users could add panels to the right using our API).

ccordoba12:
@dalthviz, @jsbautista, I think this is ready to be merged, unless you prefer to make the changes to paintEvent we talked about here too.

dalthviz:
@jsbautista will working on the refactor mentioned above here :+1:

spyder/api/panel.py
```diff
@@ -132,9 +134,23 @@ def paintEvent(self, event):
                 self.palette().windowText().color()))
             painter = QPainter(self)
             painter.fillRect(event.rect(), self._background_brush)
+            # Paint cell dividers in the visible region
```
dalthviz, line 137: I think we can remove this comment:
```suggestion
```

spyder/api/panel.py
```diff
@@ -132,9 +134,23 @@ def paintEvent(self, event):
...
         else:
             logger.debug(f'paintEvent method must be defined in {self}')
 
+    def paint_cell(self, painter):
```
dalthviz, line 141: Lets add here a docstring for this method, maybe something like:
```suggestion
    def paint_cell(self, painter):
        """Paint cell dividers in the visible region if needed."""
```

spyder/plugins/debugger/panels/debuggerpanel.py
```diff
@@ -23,6 +23,7 @@ def __init__(self, breakpoints_manager):
         """Initialize panel."""
         Panel.__init__(self)
 
+        self.linecell_color = QColor(Qt.darkGray)
```
dalthviz, line 26: I think this attribute is not necessary since the base class defines it:
```suggestion
```

spyder/plugins/debugger/panels/debuggerpanel.py
```diff
@@ -87,11 +88,11 @@ def paintEvent(self, event):
 
         Paint breakpoints icons.
         """
-        super(DebuggerPanel, self).paintEvent(event)
         painter = QPainter(self)
         painter.fillRect(event.rect(), self.editor.sideareas_color)
 
         for top, line_number, block in self.editor.visible_blocks:
+            self.paint_cell(painter)
```
dalthviz, line 95: Should this be call outside the `for` statement? Or maybe the logic inside should handle the case of passing a single block variable as argument? Otherwise I think we will be painting the line cell for all the blocks multiple times since calling this method here is doing a `for` inside a `for`, right?

spyder/plugins/editor/panels/codefolding.py
```diff
@@ -143,6 +143,7 @@ def highlight_caret_scope(self, value):
     def __init__(self):
         Panel.__init__(self)
         self._native_icons = False
+        self.linecell_color = QColor(Qt.darkGray)
```
dalthviz, line 146: I think this attribute is not necessary since the base class defines it:
```suggestion
```

spyder/plugins/editor/panels/linenumber.py
```diff
@@ -97,6 +98,7 @@ def draw_pixmap(xleft, ytop, pixmap):
 
         for top, line_number, block in self.editor.visible_blocks:
             data = block.userData()
+            self.paint_cell(painter)
```
dalthviz, line 101: Same as the other call done in the debugger panel (probably we should move this outside the `for` statement or make the method work passing a specific `block` as argument)

dalthviz (commented):
Thanks for the refactoring @jsbautista ! Left some suggestions and a couple of comments about where the new `paint_cell` method is being called. Other possibility is to add to the method the option to just run the logic for a given block (otherwise seems to me that we will be calling a `for` inside a `for` with the current usage in some of the panels).

If you have any question about the feedback let me know!

dalthviz:
Note: this needs a merge with master for the test to pass

spyder/plugins/debugger/panels/debuggerpanel.py
```diff
@@ -8,7 +8,7 @@
 """
 # Third party imports
 from qtpy.QtCore import QSize, Qt, QRect, Slot
-from qtpy.QtGui import QPainter, QFontMetrics
+from qtpy.QtGui import QPainter, QFontMetrics, QColor
```
ccordoba12, line 11:
```suggestion
from qtpy.QtGui import QPainter, QFontMetrics
```

This import is unused now.

ccordoba12 (commented):
Thanks @jsbautista for your work on this! I left some additional comments for you.

spyder/plugins/debugger/panels/debuggerpanel.py
```diff
@@ -90,6 +90,7 @@ def paintEvent(self, event):
         super(DebuggerPanel, self).paintEvent(event)
         painter = QPainter(self)
         painter.fillRect(event.rect(), self.editor.sideareas_color)
+        self.paint_cell(painter)
```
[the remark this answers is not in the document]
dalthviz: We had to change the approach with @jsbautista since we realized that calling `super` in some of the panels while adding the `paint_cell` logic in the base class is not possible (the paint operations override between the base class and the subclasses)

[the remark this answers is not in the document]
ccordoba12: Ok, no problem then. By living in the `Panel` class, `paint_cell` can be easily called by third-party plugin developers.

ccordoba12 (approved):
Looks good to me now, thanks @jsbautista!

PR #20548, opened 2023-02-17
PR: Fix test_shell_execution for macOS

mrclary:
`zsh` does not have the `-u` shell option. macOS can use `bash` instead of `zsh`.

Correct execution (via `bash`) looks like this:
```bash
(spy-dev) >> /bin/bash spyder/app/tests/bash_example.sh $(pwd)
allexport      	off
braceexpand    	on
emacs          	off
errexit        	on
errtrace       	off
functrace      	off
hashall        	on
histexpand     	off
history        	off
ignoreeof      	off
interactive-comments	on
keyword        	off
monitor        	off
noclobber      	off
noexec         	off
noglob         	off
nolog          	off
notify         	off
nounset        	off
onecmd         	off
physical       	off
pipefail       	off
posix          	off
privileged     	off
verbose        	off
vi             	off
xtrace         	on
+ TEMP_DIR=/Users/rclary/Documents/Repos/Spyder-IDE/spyder
++ uname
+ echo 'This is a temporary file created by Darwin'
```

The error from `zsh` is:
```
(spy-dev) >> /bin/zsh spyder/app/tests/bash_example.sh $(pwd)
spyder/app/tests/bash_example.sh:set:1: no such option: u
```

Since `zsh` is the default macOS shell, `-u` is removed from `bash_example.sh`

### DIFF
--- a/spyder/app/tests/bash_example.sh
+++ b/spyder/app/tests/bash_example.sh
@@ -1,4 +1,4 @@
-set -exou
+set -exo
 
 TEMP_DIR=$1
 
--- a/spyder/app/tests/test_mainwindow.py
+++ b/spyder/app/tests/test_mainwindow.py
@@ -144,7 +144,7 @@ def test_leaks(main_window, qtbot):
                 and shell._prompt_html is not None
             ),
             timeout=SHELL_TIMEOUT)
-    
+
         # Count initial objects
         # Only one of each should be present, but because of many leaks,
         # this is most likely not the case. Here only closing is tested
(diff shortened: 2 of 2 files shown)


mrclary:
@ccordoba12,
Alternatively, the `-u` option can be removed from `bash_example.sh`.

https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html
>-u
Treat unset variables and parameters other than the special parameters ‘@’ or ‘*’, or array variables subscripted with ‘@’ or ‘*’, as an error when performing parameter expansion. An error message will be written to the standard error, and a non-interactive shell will exit.

I don't think it is necessary in our case since we do not perform any parameter expansion on unset variables.
Indeed, I don't know that the `set` command or any of the options are necessary for this test.

ccordoba12:
> Alternatively, the -u option can be removed from bash_example.sh

Ok, please do that instead because we should try to run the test with `zsh` since it's the default shell on Mac.

ccordoba12 (approved):
Thanks @mrclary!

PR #20558, opened 2023-02-19
Drop bytes encodings for translations (makes compatible with Python 3…

juliangilbey:
## Description of Changes

The `lgettext()` method of the object returned by `gettext.translation` was deprecated in Python 3.8 and dropped in Python 3.11, along with the `codeset` parameter of `gettext.translation`.  The objects returned by the `gettext()` method are `str`s.  This patch simplifies the translation handling by dropping all of the implicit Python 2 checking for the type of the returned object and bytes/string conversion.  Without this patch, the translation process failed silently, leaving the interface always in English regardless of the language setting.

I don't know how to write a unit test to check that this continues to work in the future.

### Affirmation


juliangilbey:
BTW, I would have caught this issue much earlier if `except Exception` had been more specific; the only expected exception is an `OSError` from `gettext.translation`; if that had been used instead, then the no-longer-recognised keyword argument to the call would have been flagged up earlier; fixing that would have resulted an exception saying that `_trans` has no `lgettext` attribute.

spyder/config/base.py
```diff
@@ -508,15 +505,10 @@ def translate_dumb(x):
...
 
         def translate_gettext(x):
-            y = lgettext(x)
-            if is_text_string(y):
-                return y
-            else:
-                return to_text_string(y, "utf-8")
+            return _trans.gettext(x)
         return translate_gettext
     except Exception:
```
ccordoba12, line 513:
```suggestion
    except OSError:
```

ccordoba12 (commented):
Thanks @juliangilbey for your help with this! I left a small suggestion according to what you said:

> BTW, I would have caught this issue much earlier (without having to wait for a Debian user to flag it up) if the except Exception in this code had been more specific; the only expected exception is an OSError from gettext.translation

We probably add `Exception` there because an error at that point makes Spyder crash at startup, which is really bad. So, we thought it's better to have a functioning Spyder in English than something that doesn't even starts.

But if you say that `OSError` is the only possible exception, then let's go with it.

spyder/config/base.py
juliangilbey: That's a good point; crashing at startup would be bad!
There is an alternative: have `OSError` fall back to the dumb non-translation function, and any other exception print a meaningful warning message to stderr and then continue with the dumb function. But this would have to use `print` directly (or something like `debug_print`), because logging is not yet enabled at that point (as I discovered while trying to debug!).

juliangilbey: Actually, even simpler: stick with `Exception` and print a debugging output in all cases.

ccordoba12: Ok, I like this suggestion better. Let's improve it a bit for readability:

```suggestion
    except Exception as exc:
        print(
            f"Could not load translation due to: "
            f"{exc.__class__.__name__} - {exc}",
            file=sys.stderr
        )

```

juliangilbey: Maybe "translations" rather than "translation", otherwise great.

juliangilbey: (I didn't use f-strings in my original version, as I'd forgotten they were already introduced in Python 3.6!)

juliangilbey:
Oh, interesting: now some tests are failing because they can't load the translation file. This is because there is no translation file when the language is `en`! So I've just pushed a change which bails out early if `language == "en"`; tested on my machine and it seems to work. Let's see if the automated tests are happy ;-)

juliangilbey:
Actually, there's something else slightly strange going on. With the error message displaying (when language is `en`), I got dozens of these error messages showing, which indicates that `get_translation` is being called repeatedly, presumably every time that the `spyder.config.base` module is loaded. And every call loads the translations file from scratch.  Perhaps something like the following would improve matters (not yet tested; I can do that tonight or tomorrow): replace

```python
# Translation callback
_ = get_translation("spyder")
```
with
```python
# Translation callback; only initialise once
if "_" not in locals():
    _ = get_translation("spyder")
```
There might, though, be a much more standard way to do this, I'm not sure.

juliangilbey:
Ah, no, that suggestion about `_ = get_translation("spyder")` is completely unnecessary. Instead, I notice that lots of places in the code import `get_translation` and then call `_ = get_translation("spyder")`, whereas other places import `_`. Some import from `spyder.config.base`, while others import from `spyder.api.translations`. A perhaps simpler alternative is to call `gettext.install` in `spyder/config/base.py`, so that `_` is installed into the `builtins` namespace and available everywhere. But if polluting (?) the `builtins` namespace is not desirable (and there are very good reasons for it not to be), then perhaps `spyder.api.translations` should import `_` from `spyder.config.api` in addition to (or instead of) `get_translations`, and then import `_` everywhere instead of importing `get_translations`.

But that's certainly unrelated to this PR, so let's leave this PR as is.

ccordoba12 (approved):
Looks good to me now, thanks @juliangilbey!

PR #20575, opened 2023-02-21
PR: Fix conda script selection for custom interpreter env activation on Windows

dalthviz:
## Description of Changes

Seems like the custom interpreters env activation was not working (at least on Windows). When selecting a custom interpreter running latest master (8dc6d182) on Windows I got the following message:

imagen

~With the changes here I was able to not show the error message, but seems like the console gets stuck loading the kernel 🤔 :~

~imagen~

~Not sure if there is something else that I'm missing to setup on my side to prevent this but creating this PR just in case.~

Also the validation for the spyder-kernels version was missing (as well as updating the min max version of spyder-kernels supported constants). Using an interpreter from an env with the incorrect `spyder-kernels` version shows now this:

image

### Affirmation


dalthviz:
> do you have master of Spyder-kernels installed in the custom interpreter you're trying to use?

I had spyder-kernels 2.4.2 installed in the env. By installing spyder-kernels 3.0.0.dev and the changes in this PR the console was able to load :+1: so I guess there is something happening with the logic to show the missing spyder-kernels installation/correct version message in the master branch?

Also, trying to debug this while I still had spyder-kernels 2.4.2 installed, seems like the `client.kernel_handler.connection_state` was in an `error` state but the `sig_stderr` was not being triggered (?). Maybe that is the cause of the console just getting stuck with the loading page?

spyder/plugins/ipythonconsole/utils/kernelspec.py
```diff
@@ -43,16 +44,16 @@
     "<pre>"
     "    <tt>{0}</tt>"
     "</pre>"
-    "doesn't have the <tt>spyder-kernels</tt> module installed. Without this "
+    "doesn't have the <tt>spyder-kernels{1}</tt> module installed. Without this "
     "module is not possible for Spyder to create a console for you.<br><br>"
     "You can install it by activating your environment (if necessary) and "
     "running in a system terminal:"
```
ccordoba12, line 50:
```suggestion
    "doesn't have <tt>spyder-kernels</tt> version <tt>{1}</tt> installed. "
    "Without this module is not possible for Spyder to create a console for "
    "you.<br><br>"
    "You can install it by activating your environment (if necessary) and "
    "then running in a system terminal:"
```

ccordoba12 (commented):
Thanks @dalthviz for your help to solve this problem! Small suggestion for you, the rest looks good to me.

ccordoba12 (approved):
Thanks @dalthviz!

PR #20582, opened 2023-02-22
Make translation object a singleton; don't reload translations for every module

juliangilbey:
## Description of Changes

As I noted in the discussion in PR #20558, the translations are reloaded repeatedly throughout Spyder, by calling `get_translation()`. Each call loads the translations into a separate object; there are about 90 such calls, and the translations file is about 180 kB in size (depending on the language), so this could lead to up to about 15 MB of memory being used for the multiple copies of the translations file.

This patch turns the translation object into a singleton: instead of importing `get_translation` and running `_ = get_translation("spyder")` at the start of lots of files, it instead imports `_` itself.

A less invasive patch would be for `get_translation` to cache its result and to return that rather than reloading the translations on each call, but in my opinion, this patch seems to be cleaner: it removes a few lines of boilerplate from about 90 files and saves as many function calls.

The unit tests all appear to still pass with this patch, and I have manually tested it with the interface language set to German - it all seems to run fine.

### Affirmation


spyder/api/translations.py
```diff
@@ -8,4 +8,4 @@
 This module contains the internationalization API.
 """
 
-from spyder.config.base import get_translation
+from spyder.config.base import _
```
ccordoba12, line 11:
```suggestion
from spyder.config.base import _


get_translation = _
```

* This is necessary to not break API compatibility for external plugins that import `get_translation` from here. For instance, Spyder-line-profiler:

    https://github.com/spyder-ide/spyder-line-profiler/blob/master/spyder_line_profiler/spyder/plugin.py#L19

* I think that still preserves the singleton character of `_`, right?

ccordoba12 (commented):
Thanks @juliangilbey for noticing this issue and fixing it!

Small suggestion for you, otherwise looks to me.

spyder/api/translations.py
juliangilbey: Ah, yes, that's a good point I hadn't thought of. I'll force-push an amended commit.

ccordoba12 (approved):
Thanks @juliangilbey!

PR #20587, opened 2023-02-24
PR: Add sha256sum to release assets

### LINKED ISSUE #20536: Attach sha256sum for installers
### Issue Report Checklist

## Problem Description

In the release page there is no sha256sum of the installers

It would be nice to have also the sha256sum -> example how it is in the release page from notepad++ https://github.com/notepad-plus-plus/notepad-plus-plus/releases/tag/v8.4.9

mrclary:
## Description of Changes

* Artifacts uploaded on pull request and "Build as release candidate" manual triggers, not on actual releases. Assets are uploaded on release.
* Added text files for sha256sum results for Linux and Windows artifacts. These are uploaded as artifacts on "Build as release candidate" manual triggers and release assets on release. This is not performed for pull requests.

### Issue(s) Resolved

Fixes #20536

### DIFF
--- a/.github/workflows/installers-conda.yml
+++ b/.github/workflows/installers-conda.yml
@@ -260,6 +260,7 @@ jobs:
         run: ./notarize.sh -p $APPLICATION_PWD $PKG_FILE
 
       - name: Upload Artifact
+        if: env.IS_RELEASE == 'false'
         uses: actions/upload-artifact@v3
         with:
           path: ${{ env.PKG_FILE }}
--- a/.github/workflows/installers-conda.yml
+++ b/.github/workflows/installers-conda.yml
@@ -235,15 +235,19 @@ jobs:
         run: |
           [[ -n $CNAME ]] && args=("--cert-id" "$CNAME") || args=()
           python build_installers.py ${args[@]}
-          PKG_FILE=$(python build_installers.py --artifact-name)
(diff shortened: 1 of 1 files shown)


.github/workflows/installers-conda.yml
```diff
@@ -279,4 +290,4 @@ jobs:
           GITHUB_TOKEN: ${{ github.token }}
         with:
           upload_url: ${{ steps.get_release.outputs.upload_url }}
-          asset_path: ${{ env.PKG_FILE }}
+          asset_path: ${{ env.PKG_GLOB }}*
```
ccordoba12, line 293: Does this step upload now both the installer and `sha256sum.txt` file?

mrclary: Yes, both the installer and hash file are uploaded. When uploading artifacts on "Build as release candidate", it will appear as one artifact but it is really a zip file containing both the installer and the hash file; see [this test workflow dispatch](https://github.com/mrclary/spyder/actions/runs/4259084419). When uploading release assets on release event, both the installer and text file will appear as separate assets; see [this test release](https://github.com/mrclary/spyder/releases/tag/v6.0.0a1).

ccordoba12 (approved):
Thanks @mrclary!

PR #20633, opened 2023-03-04
PR: Create magics for run|debug file|cell (IPython console)

### LINKED ISSUE #20630: Proposition: turn runfile / runcell to IPython magics
The functions:
 - `runfile`
 - `runcell`
 - `debugfile`
 - `debugcell`
 
Are now added to the builtins. This is a bit strange because you wouldn't want to be able to call them from deep inside a library.
I suggest replacing them with magic instead. The calling convention would be a bit different,  but it would be cleat that these are no ordinary functions (e.g. you can't really pass any object as an argument, only strings, which is what magics are for)

impact27:
## Description of Changes

- Depends on https://github.com/spyder-ide/spyder-kernels/pull/447

### Issue(s) Resolved

Fixes #20630

### Affirmation


impact27:
Before and after in the terminal
Screenshot 2023-03-10 at 08 11 38

spyder/app/tests/test_mainwindow.py
```diff
@@ -6285,5 +6283,54 @@ def test_recursive_debug_exception(main_window, qtbot):
     assert not shell.is_debugging()
 
 
+@flaky(max_runs=3)
+def test_runfile_namespace(main_window, qtbot, tmpdir):
+    """Test that the namespaces behave correctly when using runfile."""
```
ccordoba12, line 6288:
```suggestion
    """Test that namespaces behave correctly when using runfile."""
```

spyder/plugins/help/utils/tutorial.rst
```diff
@@ -53,7 +53,7 @@ If so, then you have just run your first Python program -- well done.
 
 .. note::
 
-   The particular paths shown inside ``runfile()`` will depend on where you have saved the file, but this is inserted by Spyder automatically.
+   The particular paths shown inside ``runfile`` will depend on where you have saved the file, but this is inserted by Spyder automatically.
```
ccordoba12, line 56:
```suggestion
   The particular paths shown next to ``%runfile`` will depend on where you have saved the file, but this is inserted by Spyder automatically.
```

spyder/plugins/ipythonconsole/plugin.py
```diff
@@ -645,10 +645,6 @@ def exec_files(
         console_namespace = params['console_namespace']
         run_method = params.get('run_method', 'runfile')
 
-        # Escape single and double quotes in fname and dirname.
-        # Fixes spyder-ide/spyder#2158.
-        filename = filename.replace("'", r"\'").replace('"', r'\"')
```
ccordoba12, line 650: Why is this not necessary anymore?

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -1899,22 +1904,21 @@ def run_script(self, filename, wdir, args, post_mortem, current_client,
...
-                    and os.path.samefile(wdir, os.path.dirname(filename))
-                ):
-                    # No working directory for external kernels
-                    wdir = ""
+                    line += ' --args {}'.format(repr(norm(args)))
                 if wdir:
-                    line += ", wdir='%s'" % norm(wdir)
+                    if wdir == os.path.dirname(filename):
+                        # No working directory for external kernels
+                        # If it has not been explicitely given
```
ccordoba12, line 1913:
```suggestion
                        # if it has not been explicitly given.
```

spyder/plugins/ipythonconsole/widgets/main_widget.py
```diff
@@ -1899,22 +1904,21 @@ def run_script(self, filename, wdir, args, post_mortem, current_client,
...
+                        if not client.shellwidget.is_external_kernel:
+                            line += " --wdir"
```
ccordoba12, line 1915: So, now we simply pass `--wdir` to the `runfile` magic when `wdir` is the same directory where the file we're trying to execute is placed?

ccordoba12 (commented):
Some small suggestions and questions for you @impact27.

spyder/plugins/ipythonconsole/widgets/main_widget.py
impact27: Yes so this decreases the clutter in the console

spyder/plugins/ipythonconsole/plugin.py
impact27: I replaced that with shlex.join

impact27: I added a test for it

spyder/plugins/run/container.py
```diff
@@ -207,7 +206,7 @@ def run_file(self, selected_uuid=None, selected_executor=None):
 
         first_execution = exec_params['first_execution']
         display_dialog = exec_params['display_dialog']
-        display_dialog = display_dialog or first_execution
+        display_dialog = display_dialog and first_execution
```
impact27, line 209: I am not 100% sure but this was causing the run configuration dialog to open every time when running a new file, which is annoying and makes the tests fail

spyder/plugins/ipythonconsole/widgets/main_widget.py
ccordoba12: That's what I thought and it's a great idea! Well done 👍🏽

ccordoba12 (approved):
Looks good to me now, thanks @impact27!

Issue #20658, opened 2023-03-08
IPython console stuck restarting kernel when using different interpreter


aberrington:
### Issue Report Checklist

* [X] Searched the [issues page](https://github.com/spyder-ide/spyder/issues?q=is%3Aissue) for similar reports
* [X] Read the relevant sections of the [Spyder Troubleshooting Guide](https://github.com/spyder-ide/spyder/wiki/Troubleshooting-Guide-and-FAQ) and followed its advice
* [X] Reproduced the issue after updating with ``conda update spyder`` (or ``pip``, if not using Anaconda)
* [ ] Could not reproduce inside ``jupyter qtconsole`` (if console-related)
* [X] Tried basic troubleshooting (if a bug/error)
    * [X] Restarted Spyder
    * [X] Reset preferences with ``spyder --reset``
    * [ ] Reinstalled the latest version of [Anaconda](https://www.anaconda.com/download/)
    * [X] Tried the other applicable steps from the Troubleshooting Guide
* [X] Completed the **Problem Description**, **Steps to Reproduce** and **Version** sections below

## Problem Description

IPython console is stuck 'Restarting kernel' when using interpreter of python environment separate to Spyder default installation. 

### What steps reproduce the problem?

1. Install Spyder from standalone Windows executable
2. Create new conda environment `spydertest` and install latest spyder-kernels in this environment
3. In Spyder, select 'Preferences -> Use following Python interpreter' and then point to this dummy environment, i.e. 
`...Miniconda3/envs/spydertest/python.exe`

### What is the expected output? What do you see instead?

Using default interpreter, the IPython Console starts up and can execute code as normal. 

Using the alternative python interpreter, in a separate conda environment, console doesn't initialize and is stuck at Restarting kernel ... 

![image](https://user-images.githubusercontent.com/17913587/223723479-cdf018e1-58b1-458d-8dfa-f25fe389518e.png)

## Versions

* Spyder version: 5.4.2
* Python version: (default in Spyder install = 3.8.10, in development env = 3.10.9)
* Qt version: 5.15.2
* PyQt version: 5.15.7
* Operating System name/version: Windows 10, 22H2

### Dependencies

```
# Mandatory:
atomicwrites >=1.2.0                 :  1.4.1 (OK)
chardet >=2.0.0                      :  5.1.0 (OK)
cloudpickle >=0.5.0                  :  2.2.0 (OK)
cookiecutter >=1.6.0                 :  2.1.1 (OK)
[...]

PR #20699, opened 2023-03-17
PR: Fix restart mechanism for the Windows standalone installer (Installers)

dalthviz:
## Description of Changes

Be sure to use the script the standalone installer uses in the shortcut to start and don't remove installer `pkgs` entry from the `PYTHONPATH` when starting Spyder.

### Issue(s) Resolved

See https://github.com/spyder-ide/spyder/issues/20658#issuecomment-1468338520

### Affirmation

### DIFF
--- a/spyder/app/start.py
+++ b/spyder/app/start.py
@@ -12,8 +12,10 @@ import os
 import sys
 if os.environ.get('PYTHONPATH'):
     for path in os.environ['PYTHONPATH'].split(os.pathsep):
-        if 'pkgs' not in path:
+        if os.name == 'nt' and 'pkgs' in path:
             # Don't remove pynsist installer entry for 'pkgs' directory
+            continue
+        else:
             try:
                 sys.path.remove(path.rstrip(os.sep))
             except ValueError:
--- a/spyder/app/start.py
+++ b/spyder/app/start.py
@@ -12,10 +12,12 @@ import os
 import sys
 if os.environ.get('PYTHONPATH'):
     for path in os.environ['PYTHONPATH'].split(os.pathsep):
-        try:
-            sys.path.remove(path.rstrip(os.sep))
-        except ValueError:
-            pass
+        if 'pkgs' not in path:
+            # Don't remove pynsist installer entry for 'pkgs' directory
+            try:
+                sys.path.remove(path.rstrip(os.sep))
+            except ValueError:
+                pass
 
 # Standard library imports
 import ctypes
--- a/spyder/app/restart.py
+++ b/spyder/app/restart.py
@@ -235,7 +235,6 @@ def main():
         pynsist_python = osp.join(pynsist_installation, "Python", "pythonw.exe")
         pynsist_script = osp.join(pynsist_installation, "Spyder.launch.pyw")
         command = [f'"{pynsist_python}"', f'"{pynsist_script}"']
-        print(command)
     else:
         if is_bootstrap:
             script = osp.join(spyder_dir, 'bootstrap.py')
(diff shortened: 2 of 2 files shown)


spyder/app/start.py
```diff
@@ -12,10 +12,12 @@
 import sys
 if os.environ.get('PYTHONPATH'):
     for path in os.environ['PYTHONPATH'].split(os.pathsep):
-        try:
-            sys.path.remove(path.rstrip(os.sep))
-        except ValueError:
-            pass
+        if 'pkgs' not in path:
+            # Don't remove pynsist installer entry for 'pkgs' directory
```
ccordoba12, line 16:
```suggestion
        if os.name == 'nt' and 'pkgs' in path:
            # Don't remove pynsist installer entry for 'pkgs' directory
            continue
        else:
```

To be more specific about when we avoid removing a directory that contains `pkgs` in its path.

ccordoba12 (approved):
Left a small suggestion for you @dalthviz, otherwise looks good to me.

ccordoba12:
@mrclary, it seems one of the patches for the Linux installer is not applying cleanly now. What can we can about that?

mrclary:
It is probably because there is a merge conflict with the patch. This should be resolved by rebasing the patch branch and resolving any conflicts. I'll take a look to be sure and fix any issues I find.

mrclary:
Regarding this PR, why does the Windows Spyder application require adding package paths to the user's PYTHONPATH? That seems unreliable as PYTHONPATH is intended for user customization and may be modified for any reason. Will the conda-based installer for Windows require use of PYTHONPATH as well?

mrclary:
Okay, there is a direct merge conflict between the patch and the changes in this PR.
My recommendation is to move forward with this PR and ignore the failing conda-based installer failure. After this PR is merged, I'll rebase the patch branch and resolve the conflict, verifying that the conda-based Linux installer successfully builds.

mrclary:
So `pynsist` creates the `Spyder-launch.pyw` script you listed?
It makes sense to me that `pkgdir` should be added to `sys.path`
```python
sys.path.insert(0, pkgdir)
# Ensure .pth files in pkgdir are handled properly
site.addsitedir(pkgdir)
```
But I don't know why it is added to `os.environ['PYTHONPATH']`
```python
os.environ['PYTHONPATH'] = pkgdir + os.pathsep + os.environ.get('PYTHONPATH', '')
```

Yes, the conda-based installers are exactly a conda installation, just in a non-standard location. You asked elsewhere whether we could use an nsist template for the conda-based installer, and I believe that we can. I don't know exactly what the template will do compared to the our existing Windows installer, e.g. will it generate the same `Spyder-launch.pyw`? I suppose we'll just have to cross that bridge when we get to it.

dalthviz:
@ccordoba12 @mrclary should we merge this one as it is, right?

mrclary:
> @ccordoba12 @mrclary should we merge this one as it is, right?

I think so. I'll resolve conflicts with the conda-based installer patch branch after this PR is merged.
I did not intend to hijack or delay this PR with my most recent questions/comments regarding PYTHONPATH.

dalthviz:
> I think so. I'll resolve conflicts with the conda-based installer patch branch after this PR is merged.
:+1:
> I did not intend to hijack or delay this PR with my most recent questions/comments regarding PYTHONPATH.
Not at all! And in fact, quite interesting questions to think about in order to understand better what we need to do to migrate from the old installers :+1:

ccordoba12 (approved):
Thanks @dalthviz!

PR #20725, opened 2023-03-23
PR: Fix clicking on file names in tracebacks for IPython 8 and minor fixes for messages shown in the console

### LINKED ISSUE #20407: IPython console does not link to the file and row that caused the error anymore
## Problem Description
IPython console does not link to the file and row that caused the error anymore in spyder 5.4.2. 
It worked fine in 5.4.0.

Create a script:
```python
import numpy as np


a = np.arrayy([1, 2, 3])
```
Run the script

```python
runfile('C:/Projects/Python/Scripts/Misc/spyder_no_error_linking.py', wdir='C:/Projects/Python/Scripts/Misc')

  File ~\AppData\Local\miniconda3\lib\site-packages\spyder_kernels\py3compat.py:356 in compat_exec
    exec(code, globals, locals)

  File c:\projects\python\scripts\misc\spyder_no_error_linking.py:4
    a = np.arrayy([1, 2, 3])

  File ~\AppData\Local\miniconda3\lib\site-packages\numpy\__init__.py:311 in __getattr__
    raise AttributeError("module {!r} has no attribute "

AttributeError: module 'numpy' has no attribute 'arrayy'
```
You should be able to click on 
```python
  File ~\AppData\Local\miniconda3\lib\site-packages\numpy\__init__.py:311 in __getattr__
```
And get to that file and row, this works in 5.4.0.

## Versions

* Spyder version: 5.4.2  (conda)
* Python version: 3.9.15 64-bit
* Qt version: 5.15.6
* PyQt5 version: 5.15.7

ccordoba12:
## Description of Changes

- The format for tracebacks changed in IPython 8. That's why our previous code to detect filenames inside them failed.
- Add a test for this functionality so we notice when it breaks in the future.
- Restore showing the pointing hand cursor when hovering over file names on tracebacks. This was broken in #15809.
- Change header for non-error messages from `Warning` to `Important` to better convey the meaning of those messages to users (not all of them are warnings). Also, use white color for it so that it's more readable in the light theme.

    **Before**

    image

    **After**

    image

- Change message shown the first time a plot is generated with the Inline backend. That message was written for Spyder 3 users so they understood where their plots went in Spyder 4. But it makes little sense now for new users.

    **Before**

    image

    **After**

    image

### Issue(s) Resolved

Fixes #20407.

### Affirmation

### DIFF
--- a/spyder/utils/misc.py
+++ b/spyder/utils/misc.py
@@ -154,18 +154,16 @@ def remove_backslashes(path):
 
 def get_error_match(text):
     """Check if text contains a Python error."""
-    # For IPython 8+ tracebacks. We check for this first because we assume most
-    # users need to click on file names present on tracebacks in the IPython
-    # console.
-    # Fixes spyder-ide/spyder#20407
-    ipython8_match = re.match(r'  File (.*):(\d*)', text)
-    if ipython8_match is not None:
-        return ipython8_match
-
     # For regular Python tracebacks and IPython 7 or less.
     match_python = re.match(r'  File "(.*)", line (\d*)', text)
     if match_python is not None:
         return match_python
+
+    # For IPython 8+ tracebacks.
+    # Fixes spyder-ide/spyder#20407
+    ipython8_match = re.match(r'  File (.*):(\d*)', text)
+    if ipython8_match is not None:
+        return ipython8_match
 
     return False
 
--- a/spyder/utils/misc.py
+++ b/spyder/utils/misc.py
@@ -153,9 +153,21 @@ def remove_backslashes(path):
 
 
 def get_error_match(text):
-    """Return error match"""
-    import re
-    return re.match(r'  File "(.*)", line (\d*)', text)
+    """Check if text contains a Python error."""
+    # For IPython 8+ tracebacks. We check for this first because we assume most
+    # users need to click on file names present on tracebacks in the IPython
+    # console.
+    # Fixes spyder-ide/spyder#20407
+    ipython8_match = re.match(r'  File (.*):(\d*)', text)
+    if ipython8_match is not None:
+        return ipython8_match
+
+    # For regular Python tracebacks and IPython 7 or less.
+    match_python = re.match(r'  File "(.*)", line (\d*)', text)
+    if match_python is not None:
+        return match_python
+
+    return False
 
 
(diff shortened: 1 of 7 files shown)


spyder/utils/misc.py
```diff
@@ -153,9 +153,21 @@ def remove_backslashes(path):
...
+    # console.
+    # Fixes spyder-ide/spyder#20407
+    ipython8_match = re.match(r'  File (.*):(\d*)', text)
+    if ipython8_match is not None:
+        return ipython8_match
+
+    # For regular Python tracebacks and IPython 7 or less.
+    match_python = re.match(r'  File "(.*)", line (\d*)', text)
+    if match_python is not None:
+        return match_python
```
dalthviz, line 168: @ccordoba12 checking locally, seems like the order here needs to be changed so Spyder can properly detect the error text with IPython 7.x (at least on Windows):

```suggestion
    # For regular Python tracebacks and IPython 7 or less.
    match_python = re.match(r'  File "(.*)", line (\d*)', text)
    if match_python is not None:
        return match_python

    # For IPython 8+ tracebacks. We check for this first because we assume most
    # users need to click on file names present on tracebacks in the IPython
    # console.
    # Fixes spyder-ide/spyder#20407
    ipython8_match = re.match(r'  File (.*):(\d*)', text)
    if ipython8_match is not None:
        return ipython8_match
```

I think this happens since the error text for IPython 7.x also can partially match the rule for IPython 8.x. There the path of the files starts with a driver letter plus `:` so then something like `File "c:\users\dalth\test\fooo.py"` matches the `r'  File (.*):(\d*)` expression used for IPython 8.x. Probably it would be good to check if the order change suggested can affect the functionality on Linux? 🤔

ccordoba12: > checking locally, seems like the order here needs to be changed so Spyder can properly detect the error text with IPython 7.x (at least on Windows)

Thanks for checking that @dalthviz!

> Probably it would be good to check if the order change suggested can affect the functionality on Linux?

Checked that and your suggestion works well with both IPython 7 and 8. I just made a little change to it to remove the part in my comment that explained why the IPython 8 regex was used first (which is not true anymore).

dalthviz (approved):
Thanks @ccordoba12 ! :+1:

PR #20758, opened 2023-03-31
PR: Remove quotes from `conda-activate.bat` params and fix `micromamba.exe` detection (Installers)

### LINKED ISSUE #16562: Cannot activate conda env when created in a path

1 Create a conda environment with the --prefix option, specifying a path rather than an environment name.

2 install spyder-kernel in the environment.

3 Inside the new environment, create an environment variable using: conda env config vars set MY_VAR='HELLO'

4 launch spyder from another conda environment (can be the base environment)

5 change the Python Interpreter to point to the python.exe file in the new environment.
The path is obtained from the command: python -c "import sys; print(sys.executable)"

6 Create a new console or restart the kernel. Spyder outputs the message '<path_to_env>/Scripts/activate' is not recognized as an internal or external command.

```
Python 3.9.7 | packaged by conda-forge | (default, Sep 29 2021, 19:20:16) [MSC v.1916 64 bit (AMD64)]
Type "copyright", "credits" or "license" for more information.

IPython  -- An enhanced Interactive Python.

Restarting kernel...
 

'C:/<path_to_dir>/.conda/dev_env/Scripts/activate' is not recognized as an internal or external command,
operable program or batch file.
```
7 Check that the environment variable MY_VAR does not exist.

If the environment is created with a name and placed in the default miniconda3/envs directory, there is no issue and the new environment appears directly in the Spyder drop-down menu for the Python interpreter.

### Environment

* Spyder version: 5.1.5 None
* Python version: 3.8.12 64-bit
* Qt version: 5.12.9
* PyQt5 version: 5.12.3

Installation from miniconda3

dalthviz:
## Description of Changes

* The installers use a bundled `micromamba` executable to activate environments but the logic to detect it in `conda-activate.bat` was not working (the activation was going through the conda option). This caused environments created with a prefix option to not be correctly activated showing the following message:

```
An error ocurred while starting the kernel
The following argument was not expected: <env path>
Run with ‑‑help for more information.
```
 
* When doing a all users install on Windows the default directory for the installation is inside `Program Files` which has a space in its name. The space was causing problems to run the micromamba command to activate environments

### Issue(s) Resolved

Fixes #16562

### Affirmation


dalthviz:
Note: This could require to also update the micromamba executable version that is bundled with the installer

ccordoba12 (approved):
Thanks @dalthviz!

PR #20812, opened 2023-04-12
PR: Ensure Pythonpath Manager widget is in front and has focus after adding path on macOS

### LINKED ISSUE #20808: PYTHONPATH manager just closes after upgrade to 5.4.3 in the Mac app
### Issue Report Checklist

    * [x] Restarted Spyder

## Problem Description

After upgrading to Spyder 5.4.3 (from 5.4.2), the PYTHONPATH manager window just closes without updating when I want to add a folder. The list remains empty when I reopen the PYTHONPATH manager.

My custom modules can no longer be imported. Instead, I get a `ModuleNotFoundError`. Some (hopefully useful hints):

- I get the correct PYTHONPATH with `echo $PYTHONPATH` on the macOS Terminal. 
- I set the Python interpreter to '/Users/<username>/opt/miniconda3/bin/python'
- On the Spyder IPython console, `os.environ['PYTHONPATH']` throws a `KeyError`

I don' know

The "system wide" PYTHONPATH should be used (as it was in 5.4.2?). The PYTHONPATH manager should add folders.

```python-traceback

Uncaught ReferenceError: spyder_terminal is not defined
Uncaught ReferenceError: spyder_terminal is not defined
Uncaught ReferenceError: spyder_terminal is not defined
Uncaught TypeError: Cannot read property 'fit' of undefined
_distutils_hack/__init__.pyc:18: UserWarning: Distutils was imported before Setuptools, but importing Setuptools also replaces the `distutils` module in `sys.modules`. This may lead to undesirable behaviors or errors. To avoid these issues, avoid using distutils directly, ensure that setuptools is installed in the traditional way (e.g. not an editable install), and/or make sure that setuptools is always imported before distutils.
_distutils_hack/__init__.pyc:33: UserWarning: Setuptools is replacing distutils.

```

## Versions

* Spyder version: 5.4.3 b6abc3d30 (standalone)
* Python version: 3.9.14 64-bit
* Qt version: 5.15.2
* PyQt5 version: 5.15.9
* Operating System: Darwin 22.3.0

### Dependencies

```
# Mandatory:
atomicwrites >=1.2.0                             :  1.4.1 (OK)
chardet >=2.0.0                                  :  5.1.0 (OK)
cloudpickle >=0.5.0                              :  2.2.1 (OK)
cookiecutter >=1.6.0                             :  2.1.1 (OK)
diff_match_patch >=20181111                      :  20200713 (OK)
intervaltree >=3.0.2                             :  3.1.0 (OK)
IPython >=7.31.1,<9.0.0,!=8.8.0,!=8.9.0,!=8.10.0 :  8.12.0 (OK)
jedi >=0.17.2,<0.19.0                            :  0.18.2 (OK)
jellyfish >=0.7                                  :  0.11.2 (OK)
jsonschema >=3.2.0                               :  4.17.3 (OK)
keyring >=17.0.0                                 :  23.13.1 (OK)
nbconvert >=4.0                                  :  7.3.0 (OK)
numpydoc >=0.6.0                                 :  1.5.0 (OK)
parso >=0.7.0,<0.9.0                             :  0.8.3 (OK)
pexpect >=4.4.0                                  :  4.8.0 (OK)
pickleshare >=0.4                                :  0.7.5 (OK)
psutil >=5.3                                     :  5.9.4 (OK)
pygments >=2.0                                   :  2.14.0 (OK)
pylint >=2.5.0,<3.0                              :  2.17.2 (OK)
pylint_venv >=2.1.1                              :  None (OK)
pyls_spyder >=0.4.0                              :  0.4.0 (OK)
pylsp >=1.7.2,<1.8.0                             :  1.7.2 (OK)
pylsp_black >=1.2.0                              :  1.2.1 (OK)
qdarkstyle >=3.0.2,<3.2.0                        :  3.1 (OK)
qstylizer >=0.2.2                                :  0.2.2 (OK)
qtawesome >=1.2.1                                :  1.2.3 (OK)
qtconsole >=5.4.2,<5.5.0                         :  5.4.2 (OK)
qtpy >=2.1.0                                     :  2.3.1 (OK)
rtree >=0.9.7                                    :  1.0.1 (OK)
setuptools >=49.6.0                              :  67.6.1 (OK)
sphinx >=0.6.6                                   :  5.1.1 (OK)
spyder_kernels >=2.4.3,<2.5.0                    :  2.4.3 (OK)
textdistance >=4.2.0                             :  4.5.0 (OK)
three_merge >=0.1.1                              :  0.1.1 (OK)
watchdog >=0.10.3                                :  3.0.0 (OK)
zmq >=22.1.0                                     :  25.0.2 (OK)

# Optional:
cython >=0.21                                    :  0.29.34 (OK)
matplotlib >=3.0.0                               :  3.7.1 (OK)
numpy >=1.7                                      :  1.24.2 (OK)
pandas >=1.1.1                                   :  2.0.0 (OK)
scipy >=0.17.0                                   :  1.10.1 (OK)
sympy >=0.7.3                                    :  1.11.1 (OK)

# Spyder plugins:
spyder_terminal.terminalplugin 1.2.2             :  1.2.2 (OK)```

### PR DESCRIPTION
**Pull request by mrclary**
## Description of Changes

* Ensure that PYTHONPATH Manager widget is the front window and has focus after adding a path.
* Do not update paths upon the `path_manager_action` _if_ the widget is already visible

These changes should be rebased _after_ merging #20811.
After merging into 5.x, these changes may be merged into master.

### Issue(s) Resolved

Fixes #20808

### Affirmation

### DIFF
--- a/spyder/plugins/pythonpath/container.py
+++ b/spyder/plugins/pythonpath/container.py
@@ -105,6 +105,8 @@ class PythonpathContainer(PluginMainContainer):
 
     def show_path_manager(self):
         """Show path manager dialog."""
+        # Do not update paths or run setup if widget is already open,
+        # see spyder-ide/spyder#20808
         if not self.path_manager_dialog.isVisible():
             # Set main attributes saved here
             self.path_manager_dialog.update_paths(
--- a/spyder/plugins/pythonpath/widgets/pathmanager.py
+++ b/spyder/plugins/pythonpath/widgets/pathmanager.py
@@ -482,9 +482,13 @@ class PathManager(QDialog, SpyderWidgetMixin):
                       "."),
                     QMessageBox.Ok)
 
-        self.activateWindow()
-        self.raise_()
-        self.setFocus()
+        # Widget moves to back and loses focus on macOS,
+        # see spyder-ide/spyder#20808
+        if sys.platform == 'darwin':
+            self.activateWindow()
+            self.raise_()
+            self.setFocus()
+
         self.refresh()
 
     @Slot()

### CODE REVIEW

File: spyder/plugins/pythonpath/container.py [hunk 1]
```diff
@@ -105,13 +105,14 @@ def update_active_project_path(self, path):
...
     def show_path_manager(self):
         """Show path manager dialog."""
-        # Set main attributes saved here
-        self.path_manager_dialog.update_paths(
-            self.path, self.not_active_path, get_system_pythonpath()
-        )
-
-        # Setup its contents again
-        self.path_manager_dialog.setup()
+        if not self.path_manager_dialog.isVisible():
```
ccordoba12, line 108: Why is this really necessary? And When could this happen?

I mean, since users can only have a single instance of the dialog, this should be necessary, right? So, to clarify why we need this to solve issue #20808, please add a comment about this line.

spyder/plugins/pythonpath/widgets/pathmanager.py
```diff
@@ -482,6 +482,9 @@ def add_path(self, directory=None):
                       "."),
                     QMessageBox.Ok)
 
+        self.activateWindow()
+        self.raise_()
+        self.setFocus()
```
ccordoba12, line 487: Is this necessary on all OSes? Or only on Mac?

ccordoba12 (commented):
Thanks for your help with this @mrclary!

spyder/plugins/pythonpath/widgets/pathmanager.py
mrclary: Good question, let me check.

spyder/plugins/pythonpath/container.py
mrclary: When the icon is clicked, the action triggers `PythonpathContainer.show_path_manager` which _always_ calls `PathManager.update_paths` and `PathManager.setup`. If the widget is not already "open" then this is correct. But if it is already "open", then those methods will reset any changes the user may have made since opening the widget.

From what I could tell, the `PathManager` widget is instantiated on line 60 when `PythonpathContainer.setup` is called (I presume when the plugin is started at Spyder launch) and is never destroyed. The widget instance persists for the entire Spyder session, so the only thing to test for is whether it is visible or not.

So, #20808 revealed two problems in my opinion. 1) focus on the widget was not maintained after a directory was selected in the system file browser. 2) When clicking the icon, PYTHONPATH Manager _reset_ as well as restored focus; I think it should only restore focus if the widget is already "open".

Perhaps I'm confused and there is a better way to do this?

spyder/plugins/pythonpath/widgets/pathmanager.py
mrclary: On Windows and Linxu, PPM window stays in front and retains focus after selecting a new directory from the system file browser. Additionally, PPM window remains in front when changing focus to the main window. Neither is the case on macOS.

Do you think we should try a different approach?

ccordoba12: > On Windows and Linxu, PPM window stays in front and retains focus after selecting a new directory from the system file browser. Additionally, PPM window remains in front when changing focus to the main window.

That's what I thought.

> Do you think we should try a different approach?

Nop, just enclose these lines in an `if sys.platform == 'darwin':`

spyder/plugins/pythonpath/container.py
ccordoba12: > When the icon is clicked, the action triggers PythonpathContainer.show_path_manager which always calls PathManager.update_paths and PathManager.setup. If the widget is not already "open" then this is correct. But if it is already "open", then those methods will reset any changes the user may have made since opening the widget.

Ok, I see. So that's the problem.

> Perhaps I'm confused and there is a better way to do this?

I don't think so. Simply add a clarifying comment on top of this line to understand why the `if not self.path_manager_dialog.isVisible():` is necessary.

ccordoba12:
@mrclary, I think you need to rebase this PR to get the fixes in #20817.

ccordoba12 (approved):

PR #20823, opened 2023-04-14
PR: Change reporting of macOS system from Darwin to macOS and include machine type

mrclary:
## Description of Changes

In About widget, changed system info reporting for macOS from "Darwin" to "macOS" with appropriately changed release designation.
Also added machine type ('arm64', 'x86_64', 'AMD64', etc) to the system info for all platforms.

In the system info for issue reporting (and "Copy version info" of About widget), operating system value is replaced with the value of `platform.platform`.

"macOS 13.3" is more meaningful than "Darwin 22.4.0" to users, issue reporters, and developers. Additionally, inclusion of machine type may eliminate the need for developers to ask issue reporters "Is your machine Intel or M1?".

# New "About" widget text
Screenshot 2023-04-17 at 8 40 32 AM

# New Issue text
* Spyder version: 6.0.0.dev0 47a906db (standalone)
* Python version: 3.9.16 64-bit
* Qt version: 5.15.6
* PyQt5 version: 5.15.7
* Operating System: macOS-13.3-x86_64-i386-64bit

### DIFF
--- a/spyder/__init__.py
+++ b/spyder/__init__.py
@@ -95,6 +95,7 @@ def get_versions(reporev=True):
         'revision': revision,  # '9fdf926eccce',
         'branch': branch,  # '4.x' or master,
         'machine': platform.machine(),  # 'arm64', 'x86_64', 'AMD64', ...
+        'platform': platform.platform(aliased=True),
     }
     if sys.platform == 'darwin':
         versions.update(system='macOS', release=platform.mac_ver()[0])
@@ -114,5 +115,5 @@ def get_versions_text(reporev=True):
 * Python version: {versions['python']} {versions['bitness']}-bit
 * Qt version: {versions['qt']}
 * {versions['qt_api']} version: {versions['qt_api_ver']}
-* Operating System: {versions['system']} {versions['release']} ({versions['machine']})
+* Operating System: {versions['platform']}
 """
--- a/spyder/__init__.py
+++ b/spyder/__init__.py
@@ -81,7 +81,7 @@ def get_versions(reporev=True):
     else:
         installer = 'pip'
 
-    return {
+    versions = {
         'spyder': __version__,
         'installer': installer,
         'python': platform.python_version(),  # "2.7.3"
@@ -93,8 +93,13 @@ def get_versions(reporev=True):
         'system': platform.system(),   # Linux, Windows, ...
         'release': platform.release(),  # XP, 10.6, 2.2.0, etc.
         'revision': revision,  # '9fdf926eccce',
-        'branch': branch,  # '4.x' or master
+        'branch': branch,  # '4.x' or master,
+        'machine': platform.machine(),  # 'arm64', 'x86_64', 'AMD64', ...
     }
+    if sys.platform == 'darwin':
+        versions.update(system='macOS', release=platform.mac_ver()[0])
+
+    return versions
 
 
 def get_versions_text(reporev=True):
@@ -109,5 +114,5 @@ def get_versions_text(reporev=True):
 * Python version: {versions['python']} {versions['bitness']}-bit
 * Qt version: {versions['qt']}
 * {versions['qt_api']} version: {versions['qt_api_ver']}
-* Operating System: {versions['system']} {versions['release']}
+* Operating System: {versions['system']} {versions['release']} ({versions['machine']})
 """
--- a/spyder/widgets/about.py
+++ b/spyder/widgets/about.py
@@ -75,7 +75,7 @@ class AboutDialog(QDialog):
             Python {versions['python']} {versions['bitness']}-bit |
             Qt {versions['qt']} |
             {versions['qt_api']} {versions['qt_api_ver']} |
-            {versions['system']} {versions['release']}
+            {versions['system']} {versions['release']} ({versions['machine']})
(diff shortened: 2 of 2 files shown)


mrclary:
@spyder-ide/core-developers, what do you think about this PR?

dalthviz:
Sounds like a good idea @mrclary ! Maybe is worthy to add this change in 5.x too?

mrclary:
> Sounds like a good idea @mrclary ! Maybe is worthy to add this change in 5.x too?

I have no problem with that, depending on what others have to say.

spyder/__init__.py
```diff
@@ -93,8 +93,13 @@ def get_versions(reporev=True):
         'system': platform.system(),   # Linux, Windows, ...
         'release': platform.release(),  # XP, 10.6, 2.2.0, etc.
         'revision': revision,  # '9fdf926eccce',
-        'branch': branch,  # '4.x' or master
+        'branch': branch,  # '4.x' or master,
+        'machine': platform.machine(),  # 'arm64', 'x86_64', 'AMD64', ...
     }
+    if sys.platform == 'darwin':
+        versions.update(system='macOS', release=platform.mac_ver()[0])
```
CAM-Gerlach, line 100: Instead of this manually and somewhat inelegant hardcoding, what about using the built in [`platform.system_alias`](https://docs.python.org/3.12/library/platform.html?highlight=platform#platform.system_alias) instead which is purpose-built for this (and can handle more than just macOS)?

mrclary: What will be gained by using `platform.system_alias`? Hardcoding is still required as I must provide `'macOS'`, `platform.mac_ver()[0]`, and `platform.version()` as positional arguments to `platform.system_alias`. As far as I can tell, `platform.system_alias` is utterly and completely useless. But if I'm missing something, please enlighten me.

mrclary: Investigating `platform.system_alias` source code reveals that macOS is explicitly not handled.

CAM-Gerlach: > Hardcoding is still required as I must provide `'macOS'`, `platform.mac_ver()[0]`, and `platform.version()` as positional arguments to `platform.system_alias`.

The idea was that you would pass `platform.system()`, `platform.release()` and `platform.version()` but it seems as you mention that doesn't actually work for macOS, and is also rather verbose.

A potentially much more sensible approach that should actually work, though and is much more concise is adding a new key `platform` with the output of [`platform.platform(aliased=True)`](https://docs.python.org/3.12/library/platform.html?highlight=platform#platform.platform), which contains the desired information in human-readible form (platform, release and version), and use that instead. As of Python 3.8, it is documented to return macOS and `mac_ver` rather than the Darwin version, thus avoiding the need for this replacement.

> Investigating `platform.system_alias` source code reveals that macOS is explicitly not handled.

Oh, oops—I assumed it would handle that, but apparently not per python/cpython#11207

mrclary: > `platform.platform(aliased=True)`

These are the results that I get...
* macOS: macOS-13.3-x86_64-i386-64bit
* Ubuntu: Linux-5.19.0-38-generic-x86_64-with-glibc2.36
* Windows: Windows-10-10.0.19045-SP0

It seems this may be suitable for issue reporting. Windows doesn't display machine, 'AMD64', although this is probably not as critical as it is for macOS and Linux.
For the About widget, parsing this correctly for Linux would be problematic, so it may be best to leave that as is, except for macOS.

So, I would summarize your proposal as:
* Set a new key `'platform'` with value from `platform.platform`
* Use `'platform'` key for issue reporting, since pretty formatting is not necessary
* For macOS only, the system info for the About widget can be parsed from `'platform'` value to avoid hard-coding

Is that a faithful summary? If so, I don't have any issues with it.

CAM-Gerlach: Yup, aside from a couple small details—

>  For the About widget, parsing this correctly for Linux would be problematic

Yeah, it is intended to be human-readable and output as-is and is not supposed to be parsed (that's what `platform.system()`, etc are for), so it shouldn't be used there.

> For macOS only, the system info for the About widget can be parsed from 'platform' value to avoid hard-coding

IMO, that isn't worth the complexity as we're parsing a human-readible string rather than something actually intended to be parsed, and thus would require more assumptions and hardcoding than replacing `darwin` for `macOS`. Therefore, I'd suggest just doing the latter if nessesary for the about dialog.

mrclary:
@spyder-ide/core-developers, do you think this should be moved to 5.x?

ccordoba12:
+1 from me. For that (if others thinks that's ok), please use following commands:

    git checkout mac-version
    git rebase --onto 5.x master mac-version
    git push -f origin mac-version

mrclary:
@dalthviz, did you want to say anything or are you good?

dalthviz (approved):
Thanks @mrclary ! Checked locally the changes on Windows and LGTM :+1:

As a suggestion it could be nice to have a test that runs on macOS, Linux and Windows checking the dialog since I think the only test that we have for the dialog is the one at the end of the file which only is run on Linux 🤔

Anyhow, not something to prevent this from being merged :+1:

ccordoba12:
Let's merge this one.

PR #20824, opened 2023-04-14
PR: Declare a clear API for Projects and move to its main widget the implementation code 

ccordoba12:
## Description of Changes

- Projects had no clear API because almost all its implementation code was part of its plugin.
- This PR moved that code to its main widget, so that we can have a discernible API in the plugin.
- Fix signatures of Project's public signals because they had some incorrect or outdated args.

### Affirmation


ccordoba12:
@dalthviz, this is ready for review.

spyder/plugins/projects/widgets/main_widget.py
```diff
@@ -91,18 +724,148 @@ def set_project_dir(self, directory):
...
+
+        for item in [
+                self.clear_recent_projects_action,
+                self.max_recent_action]:
+            self.add_item_to_menu(
+                item,
+                menu=self.recent_project_menu,
+                section=RecentProjectsMenuSections.Extras)
+
+        self._update_project_actions()
```
dalthviz, line 835: Not totally sure why but seems like something strange is happening when building the recent projects entry?:

Checked creating a project, closing it and reopening and I got this layout for the recent projects menu:

recent_projects


Besides the duplicated entry, should not the actions for maximum number and to clear the list be below the actual recent projects entries?

spyder/plugins/projects/plugin.py
```diff
@@ -148,251 +100,169 @@ def on_initialize(self):
...
 
-        self.sig_project_loaded.connect(self._setup_editor_files)
-        self.sig_project_closed[bool].connect(self._setup_editor_files)
+        treewidget.sig_open_file_requested.connect(editor.load)
+        treewidget.sig_removed.connect(editor.removed)
+        treewidget.sig_tree_removed.connect(editor.removed_tree)
+        treewidget.sig_renamed.connect(editor.renamed)
+        treewidget.sig_tree_renamed.connect(editor.renamed_tree)
+        treewidget.sig_module_created.connect(editor.new)
+        treewidget.sig_file_created.connect(self._new_editor)
```
dalthviz, line 141: Maybe the projects `main_widget` could expose these signals? So from the plugin is not necessary to know that a `treewidget` like widget is available? So for example something like:

```python
widget.sig_file_created.connect(self._new_editor)
```

If worthy to change, then something similar will be needed in other places where signals for the `treewidget` widget connection is done.

spyder/plugins/projects/widgets/main_widget.py
dalthviz: Closed and open again Spyder and after that the actions order was reverted:

 
recent_projects_2

dalthviz (commented):
Thanks @ccordoba12 ! Left a comment for some behavior I experienced when testing this locally on Windows and one other comment regarding the signals from the connections for the `treewidget`

Other than that this LGTM :+1:

spyder/plugins/projects/widgets/main_widget.py
ccordoba12: Good catches @dalthviz! I'm sure they also affect `5.x` (because I only moved code around) and I think I fixed them in my last commits, so please test again.

ccordoba12:
@dalthviz, in my last commit I also fixed some bugs about recent projects that I discovered while trying to fix the bugs you reported. Its list was not being updated at startup, shutdown and when deleting a project.

dalthviz:
Just in case, while checking again, seems like something else is still not working completly right with the recent projects functionality since you can change the maximum number of recent projects but that doesn't affect until you close Spyder:

projects

Seems like this also is happening on 5.x though

spyder/plugins/projects/widgets/main_widget.py
```diff
@@ -867,6 +881,19 @@ def _update_explorer(self, _unused):
         """Update explorer tree"""
         self._setup_project(self.get_active_project_path())
 
+    def _get_valid_recent_projects(self):
+        """Get the list of valid recent projects."""
+        if hasattr(self, 'recent_projects'):
+            recent_projects = self.recent_projects
+        else:
+            recent_projects = self.get_conf('recent_projects', [])
```
dalthviz, line 889: Could not be better for this method to have a `recent_projects` param? So it will be called to initialize `self.recent_projects` with something like:

```python
self.recent_projects = self._get_valid_recent_projects(self.get_conf('recent_projects', []))
```

And when saving the recent projects be called with something like:

```python
 self.set_conf('recent_projects', self._get_valid_recent_projects(self.recent_projects))
```

ccordoba12:
@dalthviz, I amended my last commit to fix the bug you found and implement the change you suggested.

harshsingh32 (commented):
LGTM

dalthviz (approved):
Thanks @ccordoba12 ! I checked locally and seems like things are working so this LGTM :+1:

However, I was thinking and, for the bugs found/fixes done related to the recent projects functionality, maybe it could be nice to add some tests? What do you think?

ccordoba12:
That's a good idea @dalthviz. Let's merge this one for now and I'll try to add tests for the bugs we found in a follow-up PR.

PR #20827, opened 2023-04-15
PR: Change logic to detect conda-based installers and micromamba on them

mrclary:
## Description of Changes

* Replaced `pcregrep` with `pcre2grep` in Github workflow since `pcre2grep` is already included in macOS runner image
* Distinction between conda-based installation and traditional installation of Spyder, for all platforms, is determined by the existence of `spyder-menu.json` in the environment.
* Renamed `running_in_mac_app` -> `is_conda_based_app`, which now determines whether the Spyder instance is a conda-based installation for any platform. `is_pyinst` is obsolete and removed.
* Spyder's conda executable is no longer micromamba since the conda-based installer includes a standalone conda executable as well as mamba; accommodations are made accordingly.
* For conda-based installations, Spyder's runtime environment and its root environment are excluded from the list of conda environments.
* spyder-kernels is now run using conda's `run` command rather than external shell scripts.


ccordoba12:
@mrclary, please rebase with master to get the fixes to our tests. Also, quick review from your message above:

> The standalone conda executable is included in the conda-based installer and is no longer micromamba and accommodations are made accordingly.

I think we should keep `micromamba` in the installers because we could use it for `Spyder-env-manager`, right @dalthviz?

jaimergp:
Not right now, since the installation code doesn't require a solver at all. Also conda-standalone has a really bad startup time (needs to self-extract), so not really useful for anything else than bootstrapping an install.

jaimergp:
You can add 'conda' to the base environment though, since Spyder needs a Python interpreter anyway.

mrclary:
> Not right now, since the installation code doesn't require a solver at all.

Just to be clear, `conda.exe` uses the standard conda solver, but does not use libmamba, correct? Because I haven't seen any issues creating environments with `conda.exe`.

> Also conda-standalone has a really bad startup time (needs to self-extract), so not really useful for anything else than bootstrapping an install.

Interesting...
Is there a reason, then, that `menuinst` uses `<base>/conda.exe` instead of `<base>/bin/activate` when specifying `"activate": True`? or is `<base>/bin/activate` only introduced by conda/mamba?

Also, I haven't noticed any performance issues using `conda.exe`. If there is any "unpacking" time, I don't notice it in the following:
<details><summary>`$ time command info -e`</summary>

```
>> time mamba info -e
# conda environments:
#
                         /Users/rclary/Library/spyder-5.4.4.dev1473
                         /Users/rclary/Library/spyder-5.4.4.dev1473/envs/spyder-5.4.4.dev1473
base                     /Users/rclary/opt/mambaforge
bokeh                    /Users/rclary/opt/mambaforge/envs/bokeh
bokeh_apps               /Users/rclary/opt/mambaforge/envs/bokeh_apps
bokeh_py310              /Users/rclary/opt/mambaforge/envs/bokeh_py310
c2w                      /Users/rclary/opt/mambaforge/envs/c2w
c2w_310                  /Users/rclary/opt/mambaforge/envs/c2w_310
spy-dev                  /Users/rclary/opt/mambaforge/envs/spy-dev
spy-inst                 /Users/rclary/opt/mambaforge/envs/spy-inst
test                     /Users/rclary/opt/mambaforge/envs/test


real	0m0.966s
user	0m0.434s
sys	0m0.209s
>> time ~/Library/spyder-5.4.4.dev1473/conda.exe info -e
# conda environments:
#
                         /Users/rclary/Library/spyder-5.4.4.dev1473
                         /Users/rclary/Library/spyder-5.4.4.dev1473/envs/spyder-5.4.4.dev1473
                         /Users/rclary/opt/mambaforge
                         /Users/rclary/opt/mambaforge/envs/bokeh
                         /Users/rclary/opt/mambaforge/envs/bokeh_apps
                         /Users/rclary/opt/mambaforge/envs/bokeh_py310
                         /Users/rclary/opt/mambaforge/envs/c2w
                         /Users/rclary/opt/mambaforge/envs/c2w_310
                         /Users/rclary/opt/mambaforge/envs/spy-dev
                         /Users/rclary/opt/mambaforge/envs/spy-inst
                         /Users/rclary/opt/mambaforge/envs/test
base                     /var/folders/5v/28jqvwxs2cd5fj93gvwnykdrqc926z/T/_MEIsjVFap


real	0m0.766s
user	0m0.465s
sys	0m0.192s
>> time ~/Library/spyder-5.4.4.dev1473/bin/mamba info -e
# conda environments:
#
base                     /Users/rclary/Library/spyder-5.4.4.dev1473
spyder-5.4.4.dev1473     /Users/rclary/Library/spyder-5.4.4.dev1473/envs/spyder-5.4.4.dev1473
                         /Users/rclary/opt/mambaforge
                         /Users/rclary/opt/mambaforge/envs/bokeh
                         /Users/rclary/opt/mambaforge/envs/bokeh_apps
                         /Users/rclary/opt/mambaforge/envs/bokeh_py310
                         /Users/rclary/opt/mambaforge/envs/c2w
                         /Users/rclary/opt/mambaforge/envs/c2w_310
                         /Users/rclary/opt/mambaforge/envs/spy-dev
                         /Users/rclary/opt/mambaforge/envs/spy-inst
                         /Users/rclary/opt/mambaforge/envs/test


real	0m0.772s
user	0m0.326s
sys	0m0.140s
```
</details>

> You can add 'conda' to the base environment though, since Spyder needs a Python interpreter anyway.

We currently add `python`, `conda`, `mamba`, and `pip` to the base environment of the installation. Perhaps this is why we have `<base>/bin/activate`...

Nevertheless, if there are any concerns about `conda.exe`, we can just change our `find_conda` command in this PR to grab `<base>/bin/mamba` instead of `<base>/conda.exe`.

dalthviz:
> I think we should keep micromamba in the installers because we could use it for Spyder-env-manager, right @dalthviz?

Well, to give some context, the plugin is using some of the spyder functions to define what 'conda-like' executable should be used. That is done at https://github.com/spyder-ide/spyder-env-manager/blob/37f2837a077bd5dabd2b123c015c5f071826e8b9/spyder_env_manager/spyder/config.py#L24-L36

After a quick check to the changes done here, seems like that will need an update, but we are using already there the `find_conda` function so I guess it should not be much of a problem. The plugin at the end can work with `micromamba` or `conda` without issue :+1: that's why we test the `envs-manager` with `conda` and `micromamba`: https://github.com/spyder-ide/envs-manager/actions/runs/4189392934

mrclary:
> Well, to give some context, the plugin is using some of the spyder functions to define what 'conda-like' executable should be used. That is done at https://github.com/spyder-ide/spyder-env-manager/blob/37f2837a077bd5dabd2b123c015c5f071826e8b9/spyder_env_manager/spyder/config.py#L24-L36
> 
> After a quick check to the changes done here, seems like that will need an update, but we are using already there the `find_conda` function so I guess it should not be much of a problem. The plugin at the end can work with `micromamba` or `conda` without issue 👍 that's why we test the `envs-manager` with `conda` and `micromamba`: https://github.com/spyder-ide/envs-manager/actions/runs/4189392934

Yes, `get_spyder_umamba_path` is removed and `find_conda` preferentially returns Spyder's runtime conda executable, falling back to `CONDA_EXE` or `MAMBA_EXE` environment variables, and finally resorting to `find_program`.

jaimergp:
> Interesting...
Is there a reason, then, that menuinst uses <base>/conda.exe instead of <base>/bin/activate when specifying "activate": True? or is <base>/bin/activate only introduced by conda/mamba?

`conda.exe` is guaranteed to exist, while `bin/activate` requires `conda` on your `base` environment, and that's an assumption we didn't want to make or force on users (like having `python`... this is being worked on). We can probably add some logic to use other (faster) activation providers and fallback to conda.exe if nothing else was found, but well... menuinst 2 is still an alpha project.

> Also, I haven't noticed any performance issues using conda.exe. If there is any "unpacking" time, I don't notice it in the following.

Usually it takes a few seconds to return, but maybe you have it cached already. I think `pyinstaller` extracts to a random, but predictable location? Not sure.

mrclary:
> `conda.exe` is guaranteed to exist, while `bin/activate` requires `conda` on your `base` environment, and that's an assumption we didn't want to make or force on users (like having `python`... this is being worked on). We can probably add some logic to use other (faster) activation providers and fallback to conda.exe if nothing else was found, but well... menuinst 2 is still an alpha project.

Ahh, okay, understood.

> Usually it takes a few seconds to return, but maybe you have it cached already. I think `pyinstaller` extracts to a random, but predictable location? Not sure.

I did notice that `conda.exe` required writing to a temp directory; now I know why.

spyder/plugins/variableexplorer/widgets/collectionsdelegate.py
```diff
@@ -119,7 +119,7 @@ def createEditor(self, parent, option, index, object_explorer=False):
                 message = _("Spyder is unable to show the {val_type} or object"
                             " you're trying to view because <tt>{module}</tt>"
                             " is not installed. ")
-                if running_in_mac_app():
+                if is_conda_based_app():
                     message += _("Please consider using the full version of "
                                  "the Spyder MacOS application.<br>")
```
ccordoba12, line 124: I think this could be removed because we're not planning to release light versions of our conda installers, right?

spyder/plugins/variableexplorer/widgets/collectionsdelegate.py
```diff
@@ -130,7 +130,7 @@ def createEditor(self, parent, option, index, object_explorer=False):
                     message.format(val_type=val_type, module=module))
                 return
             else:
-                if running_in_mac_app() or is_pynsist():
+                if is_conda_based_app():
```
ccordoba12, line 133: This `if` could also be removed, right?

spyder/utils/conda.py
```diff
@@ -114,12 +112,23 @@ def get_conda_env_path(pyexec, quote=False):
...
     """Find conda executable."""
-    # First try the environment variables
-    conda = os.environ.get('CONDA_EXE') or os.environ.get('MAMBA_EXE')
+    conda = None
+
+    # First try Spyder's conda executable
+    if is_conda_based_app():
+        # TODO: Change to CONDA_EXE when
+        # conda-forge/conda-standalone-feedstock#45 is resolved
+        conda = osp.dirname(os.environ.get('CONDA_PYTHON_EXE')) + '/bin/mamba'
```
ccordoba12, line 121: Two questions about this:

1. Are we going to package `mamba` or `micromamba` in our installers? This code implies that we're going to use `mamba`, but why not micromamba since it's a single binary?
2. Does this work on Windows? I think executables are put in `path\to\env\Library\bin` there.

spyder/utils/conda.py
```diff
@@ -132,9 +141,15 @@ def get_list_conda_envs():
...
+    if is_conda_based_app():
+        # Need root_prefix to remove Spyder's runtime environment
+        cmdstr = ' '.join([conda, 'info', '--json'])
+    else:
+        # micromamba info does not provide envs, so must keep env list
+        cmdstr = ' '.join([conda, 'env', 'list', '--json'])
+
     try:
-        out, __ = run_shell_command(cmdstr, env={}).communicate()
+        out, __ = run_shell_command(cmdstr).communicate()
```
ccordoba12, line 152: Why the `env` is not empty anymore here?

spyder/utils/conda.py
```diff
@@ -145,6 +160,14 @@ def get_list_conda_envs():
         path = osp.join(env, 'python.exe') if WINDOWS else osp.join(
             env, 'bin', 'python')
 
+        # Skip if executable does not exist
+        if not osp.exists(path):
+            continue
+
+        # Skip if conda-based Spyder runtime or base environment
```
ccordoba12, line 167: Why don't we list the base env among the available envs anymore?

spyder/plugins/application/confpage.py
```diff
@@ -69,7 +69,8 @@ def setup_page(self):
                               'check_updates_on_startup')
 
         # Decide if it's possible to activate or not single instance mode
-        if running_in_mac_app():
+        # ??? Should this apply to all conda-based installers?
```
ccordoba12, line 72:
```suggestion
```

I don't think it's necessary to require a single instance of our app on Windows and Linux. And perhaps we should remove the code below for our new Mac app because users have been requesting the ability to have multiple Spyder instances in that OS. But that's for another PR, I'd say.

spyder/config/base.py
```diff
@@ -541,44 +531,29 @@ def translate_gettext(x):
 
 
 #==============================================================================
-# Mac application utilities
+# Conda-based installer application utilities
 #==============================================================================
-def running_in_mac_app(pyexec=sys.executable):
+REAL_SYS_EXE = osp.realpath(sys.executable)  # sys.executable may be symlink
+
+
```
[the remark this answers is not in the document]
mrclary: PEP likes 2 lines before a def

spyder/plugins/application/confpage.py
```diff
@@ -69,7 +69,8 @@ def setup_page(self):
...
+        if is_conda_based_app():
```
mrclary, line 73: In that case, perhaps we should do this?
Multiple instances for macOS would have to be investigated later, as you suggest.
```suggestion
        # ??? Should we allow multiple instances for macOS?
        if sys.platform == 'darwin' and is_conda_based_app():
```

spyder/plugins/ipythonconsole/scripts/conda-activate.bat
```diff
@@ -11,10 +11,10 @@ set SPYDER_KERNEL_SPEC=%4
 chcp 65001>nul
 
 :: Activate kernel environment
-echo %CONDA_ACTIVATE_SCRIPT%| findstr /e "micromamba.exe">Nul && goto micromamba || goto conda
+echo %CONDA_ACTIVATE_SCRIPT%| findstr /e "conda.exe">Nul && goto standalone || goto conda
 
-:micromamba Activate using micromamba
-for /f %%i in ('%CONDA_ACTIVATE_SCRIPT% shell activate %CONDA_ENV_PATH%') do set SCRIPT=%%i
+:standalone Activate using standalone conda executable
+for /f %%i in ('%CONDA_ACTIVATE_SCRIPT% shell.cmd.exe activate %CONDA_ENV_PATH%') do set SCRIPT=%%i
```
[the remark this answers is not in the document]
mrclary: Yes. Although these changes may be moot after the discussion regarding `conda.exe`. In this PR, I've since changed Spyder's conda executable to the included `mamba`, so the `:standalone` block will never be executed.

spyder/plugins/onlinehelp/widgets.py
```diff
@@ -71,7 +71,8 @@ def spyder_safeimport(path, forceload=0, cache={}):
 
 # Needed to prevent showing a warning message regarding debugging
 # See spyder-ide/spyder#20390
-if is_pynsist():
+# ??? Do we need this?
```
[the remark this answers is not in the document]
mrclary: I'll check it.

spyder/plugins/variableexplorer/widgets/collectionsdelegate.py
mrclary: Yeah, I suppose we should discuss that. At the very least, `MacOS` should be removed from the message.

spyder/plugins/variableexplorer/widgets/collectionsdelegate.py
mrclary: I suppose we should review all the messaging in this file, yes?

mrclary: I'll remove the flow controls, and revise the messaging accordingly.

spyder/utils/conda.py
mrclary: Yes, we should have a deeper discussion about that.

Originally, I thought that `conda.exe` would be a direct replacement for `micromamba`. But per @jaimergp's comments, that is not the case (at least not yet). Currently, the conda-based installer includes `mamba` in the base environment, so it was already there and `micromamba` would be another element to deal with. However, we could possibly omit conda and mamba from the base environment and include micromamba instead.

spyder/utils/conda.py
mrclary: That is because `conda.exe` needed the `USERPROFILE` environment variable to write to temporary file. However, this can reverted since we aren't going to use `conda.exe`.

spyder/utils/conda.py
mrclary: Only the base environment of the conda-based installer is omitted. Base environments in general are not affected. Spyder's runtime environment for conda-based installers is still used for "Same as Spyder", but not listed in the "custom" Python interpreters. I don't think we should be advertising these environments to users, inviting them to manipulate them and break Spyder.

mrclary:
@ccordoba12, Thank you for all your feedback and taking the time to review the PR. I've marked it as draft until we investigate further some the review comments. Perhaps we can discuss some of the points at the next developer meeting.

spyder/utils/conda.py
jaimergp: Take into account that micromamba ships its own implementation of `menuinst 1.x`, in case that's relevant for your shortcuts.

mrclary: @jaimergp, thanks for clarifying. I don't think it will matter for us, since `conda.exe` is handling the shortcuts for us at install time. The other executable ('mamba' or 'micromamba') included with the base environment will be used by the Spyder application at runtime to do various things.

If `conda.exe` ever becomes an effective equivalent to `micromamba`, then we could just use that and not worry about including any other flavor.

ccordoba12: > the conda-based installer includes mamba in the base environment

Ok, I didn't know that. Then I think it's fine to use `mamba` here.

> Take into account that micromamba ships its own implementation of menuinst 1.x, in case that's relevant for your shortcuts.

Thanks for letting us know about that @jaimergp.

jaimergp: > So it seems that conda cannot be omitted in favor of micromamba. Is this still the case, or is the documentation outdated?

I guess we can make it work, but for now there are strict checks for the `conda` package itself. It's not needed for the solve but the code operates under the assumption that if you want a separate environment it's because you are going to use `conda` on it, otherwise you could just deploy to a conda-less `base`.

Happy to revisit that decision, though!

mrclary: > Happy to revisit that decision, though!

No need, I just wanted to clarify what could or could not be done. Thanks.
Given what I now understand, I think the best path forward is to just include mamba in the base environment, use this for Spyder's runtime conda executable, and then there is no need for micromamba. There are also some minor differences between micromamba and mamba with regard to their `info` command that makes mamba more preferable anyway.

jaimergp: If you already have a setup that supports `mamba`, very little I'd say. We have some tighter integration with the classic solver, but it's no big deal. The main advantage is not having to have code that checks for `mamba` or `conda` executables; it's just `conda` with some configuration on `.condarc`.

spyder/plugins/onlinehelp/widgets.py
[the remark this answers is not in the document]
mrclary: I'm getting the error `The baseline image directory does not exist. This is most likely because the test data is not installed. You may need to install matplotlib from source to get the test data.` when searching the online help from the master branch, prior to the changes in this PR.

mrclary:
@ccordoba12 @dalthviz, This should be ready for follow-up review. Aside from addressing the review comments, additional changes are:
* I went ahead and simplified the `spyder-kernel` startup as @ccordoba12 and I discussed. Changes to `mamba` from `micromamba` would have simplified the startup scripts a great deal, but it was just easier to make the change now instead of waiting for another PR.
* I changed how we distinguish conda-based app installations to look for `spyder-menu.json` in the environment instead of the environment variable `SPYDER_APP`. This should be more robust and properly distinguish even when launching from commandline.

mrclary:
I changed where conda-based environments were filtered: instead of in `spyder.utils.conda.get_list_conda_env`, now it is in `spyder.plugins.maininterpreter.confpage`.
This means that `get_list_conda_env` will still list all environments on the system without prejudice, but the custom interpreters still do not reveal conda-based runtime environment or its base.

spyder/utils/conda.py
```diff
@@ -114,12 +84,21 @@ def get_conda_env_path(pyexec, quote=False):
 
 def find_conda():
     """Find conda executable."""
-    # First try the environment variables
-    conda = os.environ.get('CONDA_EXE') or os.environ.get('MAMBA_EXE')
+    conda = None
+
+    # First try Spyder's conda executable
+    if is_conda_based_app():
+        conda = osp.join(osp.dirname(os.environ['CONDA_EXE']), 'mamba')
```
dalthviz, line 91: In the Windows case, I think this needs to be `mamba.exe` otherwise validations related to the path (like if it is a file or it exists) done elsewhere using the value returned by this function can fail. Noticed this while updating the Spyder-env-manager plugin which does a compatibility check and it was failing with the message about the conda executable not being available. The code validating this uses `pathlib.Path.exists` which returns False unless you add the extension of the file: https://github.com/spyder-ide/spyder-env-manager/blob/37f2837a077bd5dabd2b123c015c5f071826e8b9/spyder_env_manager/spyder/plugin.py#L111-L113

dalthviz (commented):
Thanks @mrclary ! Left a couple of comments, but other than that, this LGTM :+1:

spyder/plugins/variableexplorer/widgets/collectionsdelegate.py
```diff
@@ -119,22 +119,16 @@ def createEditor(self, parent, option, index, object_explorer=False):
                 message = _("Spyder is unable to show the {val_type} or object"
                             " you're trying to view because <tt>{module}</tt>"
                             " is not installed. ")
```
ccordoba12, line 121:
```suggestion
                message = _("Spyder is unable to show the {val_type} object "
                            "you're trying to view because the "
                            "<tt>{module}</tt> is missing. Please install that "
                            "package in your Spyder environment to fix this "
                            "problem.")
```

* This re-adds the second part of the message that you removed below (i.e. `Please install ...`) because it's the part that tells users how to solve this issue by themselves.
* I reworded the message a bit too because its meaning was not entirely clear.
* Also, please change `val_type = 'dataframe, series'` in line `118` to be `val_type = 'dataframe or series'` (I can't leave a suggestion for that, sorry).

ccordoba12 (commented):
Thanks @mrclary for addressing our reviews! One last suggestion for you, then this should be ready.

spyder/plugins/variableexplorer/widgets/collectionsdelegate.py
mrclary: I will remove `the` before `<tt>{module}</tt>` from your suggestion: `because numpy is missing` is better than `because the numpy is missing`.

ccordoba12 (approved):
Thanks @mrclary!

PR #20714, opened 2023-03-22
PR: Make the switcher a plugin

angelasofiaremolinagutierrez:
## Description of Changes

- I transformed the File Switcher widget into a SpyderPluginV2. However it is not working yet, because the old widget worked in the editor with a manager called `EditorSwitcherManager` and the methods that this manager handles are not migrated yet.
- Skipped switcher tests until migration is finished

## What is left to do 

- The switcher plugin needs a public API that contains the methods of the [`EditorSwitcherManager`](https://github.com/spyder-ide/spyder/blob/master/spyder/plugins/editor/utils/switcher.py)
- The order of the file menu got mixed up when the switcher action section was added. I [tried to fix it](923a9aadc414243da68f4af977ce7192259fb745) with the `before_section` parameter in each section, but it didn't work. There might be a bug in the way the menus are getting ordered:
https://github.com/spyder-ide/spyder/blob/ad14a7b50e9cee17dc35c60c30ff12e7ebb3ce23/spyder/api/widgets/menus.py#L220

* [X] Included a screenshot or animation (if affecting the UI, see [Licecap](https://www.cockos.com/licecap/))
mixed_order

### Issue(s) Resolved

This is necessary to solve #3860 

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct: @angelasofiaremolinagutierrez

Issue #3860, opened 2016-12-19
File Switcher should search all project files and not only the open ones


talkaminker:
Currently Ctrl+P fuzzy searches only the open files. in other editors (sublime, pycharm) it fuzzy searches the whole project so that you can easily find files regardless of whether they are open or not.

Since good implementation might be tricky, maybe you should somehow use [fzf project](https://github.com/junegunn/fzf)?

ccordoba12:
@mariacamilaremolinagutierrez, I'm assigning this one to you but for 3.2, not 3.1 ;-)

@talkaminker, unfortunately `fzf` is made in Go, but we already have our own fuzzy search library.

goanpeca:
@ccordoba12 is this going to v4.2 or v5 ?

ccordoba12:
This is for Spyder 5.

maurerle:
Just a finding:
Ctrl+Shift+P is used as the Print-Key on non OSX. Yet the Project Pane already has this Shortcut, so there already are shortcuts for:

* File Switcher -> Ctrl+P
* Project Pane -> Ctrl+Shift+P
* Printing -> Nothing

ccordoba12:
Note: This problem is really complex and difficult to solve in pure Python. So now I agree that we should use `fzf` to solve it.

degel123:
Guys any progress on that? It would save a lot of time

ccordoba12:
Perhaps we'll have time to work on this before the end of the year.

ccordoba12:
A few more details about how this could be implemented:

1. I think we need to create first a `SpyderPluginV2` for the switcher, so that it provides a public API that other plugins (in this case Projects) can rely on to populate it with the content they want to display there.
2. Then, we'd to implement a way to call [`fzf`](https://github.com/junegunn/fzf) with the query passed by the user in the switcher and gather the results it returns.
3. Finally we'd need to display those results in the Switcher.

Some important points to take into account are the following:

* This will only work for projects. In other words, if no project is active, then this functionality won't be available.
* We need to filter the results coming from `fzf` to only display plain text files because those are the only ones that can be handled by the editor.

PR #20837, opened 2023-04-17
PR: Define Switcher Plugin public API + fix switcher position bug

angelasofiaremolinagutierrez:
## Description of Changes

- Implemented the plublic API for the Switcher Plugin.
- Fixed the bug occurring with the position of switcher in an editor new window.
- Moved back the EditorSwitcherManager from `spyder/plugins/switcher/` to `spyder/plugins/editor/utils/` and rename it from `manager.py` to `switcher_manager.py`

* [x] Wrote at least one-line docstrings

* [x] Included a screenshot or animation

switcher_position

### Issue(s) Resolved

Continue PR #20714
Necessary to solve https://github.com/spyder-ide/spyder/issues/3860

### Affirmation

I certify the above statement is true and correct: @angelasofiaremolinagutierrez


spyder/plugins/editor/widgets/editor.py
```diff
@@ -287,10 +286,17 @@ def __init__(self, parent, actions):
         self.setLayout(layout)
 
         self.menu = None
-        self.switcher_dlg = None
         self.switcher_manager = None
         self.tabs = None
         self.tabs_switcher = None
+        self.switcher_plugin = None
```
ccordoba12, line 292: I think for simplicity we could rename this attribute to be `switcher`. What do you say @dalthviz?

spyder/plugins/switcher/container.py
```diff
@@ -84,18 +64,25 @@ def open_switcher(self, symbol=False):
...
-        # Note: The +8 pixel on the top makes it look better
-        # FIXME: Why is this using the toolbars menu? A: To not be on top of
-        # the toolbars.
-        # Probably toolbars should be taken into account for this 'delta' only
-        # when are visible
+        # Set position
         mainwindow = self._plugin.get_main()
-        delta_top = (mainwindow.toolbar.toolbars_menu.geometry().height() +
-                     mainwindow.menuBar().geometry().height() + 8)
+        # Note: The +8 pixel on the top makes it look better
```
ccordoba12, line 69:
```suggestion

        # Note: The +8 pixel on the top makes it look better
```

spyder/plugins/switcher/container.py
```diff
@@ -84,18 +64,25 @@ def open_switcher(self, symbol=False):
...
+        default_top = (mainwindow.toolbar.toolbars_menu.geometry().height() +
+                       mainwindow.menuBar().geometry().height() + 8)
+
+        current_window = QApplication.activeWindow()
+        if current_window == mainwindow:
+            option = 'toolbars_visible'
+            section = 'toolbar'
+            if self.get_conf(option, section):
```
ccordoba12, line 77:
```suggestion
            if self.get_conf('toolbars_visible', section='toolbar'):
```

I think there0s ni need to use those two intermediate variables here.

spyder/plugins/switcher/plugin.py
```diff
@@ -9,15 +9,20 @@
 Switcher Plugin.
 """
 
+# Third-party imports
+from qtpy.QtCore import Signal
+
 # Standard library imports
 import sys
```
ccordoba12, line 16:
```suggestion
# Standard library imports
import sys

# Third-party imports
from qtpy.QtCore import Signal
```

We always place standard library imports first.

spyder/plugins/switcher/plugin.py
```diff
@@ -41,6 +46,50 @@ class Switcher(SpyderPluginV2):
     CONF_FILE = False
 
     # --- Signals
+    sig_rejected = Signal()
+    """
+    This signal is emitted when the plugin is dismissed.
+    """
+
+    sig_text_changed = Signal(TEXT_TYPES[-1])
```
ccordoba12, line 54:
```suggestion
    sig_text_changed = Signal(str)
```

We can safely use `str` here because we don't support Python 2 anymore.

spyder/plugins/switcher/plugin.py
```diff
@@ -41,6 +46,50 @@ class Switcher(SpyderPluginV2):
...
+    search_text: str
+        The current search/filter text.
+    """
+
+    sig_item_changed = Signal(object)
+    """
+    This signal is emitted when the plugin current item changes.
+    """
+
+    sig_item_selected = Signal(object, TEXT_TYPES[-1], TEXT_TYPES[-1])
```
ccordoba12, line 69:
```suggestion
    sig_item_selected = Signal(object, str, str)
```

spyder/plugins/switcher/plugin.py
```diff
@@ -9,15 +9,20 @@
...
 
 # Local imports
 from spyder.api.translations import _
 from spyder.api.plugins import Plugins, SpyderPluginV2
-from spyder.api.plugin_registration.decorators import on_plugin_available, on_plugin_teardown
+from spyder.api.plugin_registration.decorators import (on_plugin_available,
+                                                       on_plugin_teardown)
 from spyder.plugins.switcher.container import SwitcherContainer
 from spyder.plugins.mainmenu.api import ApplicationMenus, FileMenuSections
+from spyder.py3compat import TEXT_TYPES
```
ccordoba12, line 25:
```suggestion
```

This is not really necessary (see below).

spyder/plugins/switcher/plugin.py
```diff
@@ -41,6 +46,50 @@ class Switcher(SpyderPluginV2):
...
+    ----------
+    item: object
+        The current selected item from the switcher list (QStandardItem).
+    mode: str
+        The current selected mode (open files "", symbol "@" or line ":").
+    search_text: str
+        Cleaned search/filter text.
+    """
+
+    sig_mode_selected = Signal(TEXT_TYPES[-1])
```
ccordoba12, line 84:
```suggestion
    sig_mode_selected = Signal(str)
```

spyder/plugins/switcher/plugin.py
```diff
@@ -89,3 +146,85 @@ def on_main_menu_teardown(self):
...
+        self._switcher.set_placeholder_text(text)
+
+    def setup(self):
+        """Set-up list widget content based on the filtering."""
+        self._switcher.setup()
+
+    def set_position(self, top, parent=None):
+        """Set the position of the dialog."""
+        self._switcher.set_position(top, parent)
+
```
ccordoba12, line 162:
```suggestion
```

This method doesn't seem to be necessary because it's not used in `EditorSwitcherManager`, only internally in the Switcher container.

spyder/plugins/switcher/plugin.py
```diff
@@ -89,3 +146,85 @@ def on_main_menu_teardown(self):
...
+    # QDialog methods
+    def show(self):
+        """Show switcher widget with QT hide method for QDialog."""
```
ccordoba12, line 165:
```suggestion
        """Show switcher."""
```

I don't think we need to be too explicit about what this method does.

spyder/plugins/switcher/plugin.py
```diff
@@ -89,3 +146,85 @@ def on_main_menu_teardown(self):
...
+        self._switcher.show()
+
+    def hide(self):
+        """Hide switcher widget with QT hide method for QDialog."""
```
ccordoba12, line 169:
```suggestion
        """Hide switcher."""
```

spyder/plugins/switcher/plugin.py
```diff
@@ -89,3 +146,85 @@ def on_main_menu_teardown(self):
...
+        self._switcher.hide()
+
+    def close(self):
```
ccordoba12, line 172: Do we really need this method as part of the public API? I think with `hide` should be enough, right @dalthviz?

spyder/plugins/switcher/plugin.py
```diff
@@ -89,3 +146,85 @@ def on_main_menu_teardown(self):
...
+        """Close switcher widget with QT close method for QDialog."""
+        self._switcher.close()
+
+    def setVisible(self, visible):
+        """Show or hide switcher widget with QT hide method for QDialog."""
```
ccordoba12, line 177:
```suggestion
    def set_visible(self, visible):
        """Show or hide switcher."""
```

We don't use camel-cased names for our own methods. And I think this requires a small change in `EditorSwitcherManager` for it to work.

spyder/plugins/switcher/plugin.py
```diff
@@ -89,3 +146,85 @@ def on_main_menu_teardown(self):
...
+        self._switcher.setVisible(visible)
+
+    def isVisible(self):
+        """Return if switcher widget is visible with QT hide method
+        for QDialog."""
```
ccordoba12, line 182:
```suggestion
    def is_visible(self):
        """Return if the switcher is visible."""
```

spyder/plugins/switcher/plugin.py
```diff
@@ -89,3 +146,85 @@ def on_main_menu_teardown(self):
...
+
+    # Item methods
+    def current_item(self):
+        """Return the current selected item in the list widget."""
+        return self._switcher.current_item()
+
+    def add_item(self, icon=None, title=None, description=None, shortcut=None,
+                 section=None, data=None, tool_tip=None, action_item=False,
+                 last_item=True):
+        """Add switcher list item."""
```
ccordoba12, line 193:
```suggestion
        """Add a switcher list item."""
```

spyder/plugins/switcher/plugin.py
```diff
@@ -89,3 +146,85 @@ def on_main_menu_teardown(self):
...
+        self._switcher.add_item(icon, title, description, shortcut,
+                                section, data, tool_tip, action_item,
+                                last_item)
+
+    def add_separator(self):
+        """Add separator item."""
```
ccordoba12, line 199:
```suggestion
        """Add a separator item."""
```

spyder/plugins/switcher/widgets/switcher.py
```diff
@@ -93,15 +93,51 @@ class Switcher(QDialog):
...
 
-    # Dismissed switcher
+    # --- Signals
     sig_rejected = Signal()
-    # Search/Filter text changes
+    """
+    This signal is emitted when the plugin is dismissed.
+    """
+
     sig_text_changed = Signal(TEXT_TYPES[-1])
```
ccordoba12, line 102:
```suggestion
    sig_text_changed = Signal(str)
```

And please don't forget to remove the `TEXT_TYPES` import above.

spyder/plugins/switcher/widgets/switcher.py
```diff
@@ -93,15 +93,51 @@ class Switcher(QDialog):
...
+    """
+
     sig_item_changed = Signal(object)
-    # List item selected, mode and cleaned search text
-    sig_item_selected = Signal(object, TEXT_TYPES[-1], TEXT_TYPES[-1], )
+    """
+    This signal is emitted when the plugin current item changes.
+    """
+
+    sig_item_selected = Signal(object, TEXT_TYPES[-1], TEXT_TYPES[-1])
```
ccordoba12, line 117:
```suggestion
    sig_item_selected = Signal(object, str, str)
```

spyder/plugins/switcher/widgets/switcher.py
```diff
@@ -93,15 +93,51 @@ class Switcher(QDialog):
...
+    ----------
+    item: object
+        The current selected item from the switcher list (QStandardItem).
+    mode: str
+        The current selected mode (open files "", symbol "@" or line ":").
+    search_text: str
+        Cleaned search/filter text.
+    """
+
     sig_mode_selected = Signal(TEXT_TYPES[-1])
```
ccordoba12, line 132:
```suggestion
    sig_mode_selected = Signal(str)
```

ccordoba12 (commented):
Great work @angelasofiaremolinagutierrez! I left some minor suggestions for you and a couple of questions, the rest looks pretty good to me.

spyder/plugins/editor/widgets/editor.py
dalthviz: I think leaving `switcher_plugin` is okay since that could be handy in the future (to know that the reference here is to the Switcher plugin by just checking the variable name)

spyder/plugins/switcher/plugin.py
angelasofiaremolinagutierrez: Oops you are right! sorry.

spyder/plugins/switcher/plugin.py
angelasofiaremolinagutierrez: Yes, I think one is enough, let me know if I should remove it.

ccordoba12: @dalthviz, what do you think about this?

dalthviz: Sure, maybe even instead of overriding `close` here the container `on_close` method should be implemented doing this?

spyder/app/tests/conftest.py
```diff
@@ -477,8 +477,8 @@ def main_window(request, tmpdir, qtbot):
                     for editorwindow in window.editor.editorwindows:
                         editorwindow.close()
                     editorstack = window.editor.get_current_editorstack()
-                    if editorstack.switcher_dlg:
-                        editorstack.switcher_dlg.close()
+                    if editorstack.switcher_plugin:
+                        editorstack.switcher_plugin.close()
```
dalthviz, line 481: Depending on what is done to this method this could end up being removed or changed to something like:

    editorstack.switcher_plugin.on_close()

Or

    editorstack.switcher_plugin.get_container().on_close()

dalthviz (commented):
Thanks for all the work here @angelasofiaremolinagutierrez ! Left some comments regarding methods that could be part of the plugin (`open_switcher` and `open_symbolfinder`), some info about the `on_close` method that the plugin or container can implement, and an attribute that seems like could be removed (`self._modes` at the plugin `on_initialize`).

If you have any question about the above or something else lets us know!

spyder/plugins/switcher/plugin.py
angelasofiaremolinagutierrez: okay, so should I remove the `hide()` method and leave only `on_close()` method? or should I keep both and just rename `close()` to `on_close()`?

dalthviz: I think leaving the `hide` and renaming the close to `on_close` works :+1:

However, just in case, what do you think @ccordoba12 ?

dalthviz (approved):
Thanks @angelasofiaremolinagutierrez !

ccordoba12 (approved):
Looks good to me too, thanks @angelasofiaremolinagutierrez!

Issue #20001, opened 2022-11-07
Conda-based Installer Parent Issue


mrclary:
This issue is intended to capture discussions related to the development of the conda-based standalone Spyder installers post release of Spyder 5.4.0.

This description will aim to capture the up-to-date status of the development strategy resulting from a consensus of comments. Bug fixes and explicit tasks should be reported in a separate issue and PR.

# Outstanding Questions, Ideas, & Proposals
1. Update the Spyder conda installation as follows:
   - Make the root conda environment the Spyder runtime environment
   - Name the root conda environment directory `spyder-X.X` with major and minor version only
   - The root environment will support all bug fix releases within the minor release, and can be updated via conda-forge
   - New minor and major releases will require downloading a new installer.
1. Update the Spyder conda installation as follows:
   - Make the root conda environment the Spyder runtime environment
   - Name the root conda environment directory `spyder-X` with major version only
   - The root environment will support all bug fix and feature releases, and can be updated via conda-forge
   - New major releases will require downloading a new installer.
1. All updates, including those requiring new installer, can be managed by Spyder's update manager
# To Implement
* [x] Linux installer suggestions, see [Suggestions from @ccordoba12](https://github.com/spyder-ide/spyder/issues/20176#issuecomment-1366006777) and #20319
   * Add command line alias to launch Spyder
   * Create uninstall script and alias for command line
   * Add message at the end of installation providing launch instructions
   * Update the application icon
* [x] Add installer tests for Linux and Windows #20601 
* [x] Remove options from Windows installer #20838 :
   * Add Spyder to my PATH environment variable
   * Register Spyder as my default Python 3.10
* [x] Group taskbar icons on Windows #20791 conda/menuinst#127
* [x] Prevent showing a cmd window when launching Spyder from the installer shorcut conda/menuinst#128
* [x] Installer should ensure that Spyder is closed. Cannot be done with `constructor`. Implemented with `UpdateManager` plugin (#21483)
* [x] Installer should provide option to uninstall existing conflicting installation. Cannot be done with `constructor`. Implemented with `UpdateManager` plugin (#21483)
* [x] #21640
* [x] #21647
* [x] #21782
* [x] #21785
* [x] #21836
* [x] #21871
* [x] #22059
* [x] #22072
[...]

PR #20838, opened 2023-04-18
PR: Remove installer option to initialize the conda environment and misc improvements

mrclary:
## Description of Changes

* Remove installer option to initialize the conda environment. See https://github.com/spyder-ide/spyder/issues/20001#issuecomment-1510448472
* Revise `uninstall-spyder` to prompt user for confirmation before uninstalling.

### DIFF
--- a/installers-conda/resources/post-install.sh
+++ b/installers-conda/resources/post-install.sh
@@ -38,6 +38,17 @@ fi
 echo "Creating uninstall script..."
 cat <<EOF > ${u_spy_exe}
 #!/bin/bash
+echo "You are about to uninstall Spyder."
+echo "If you proceed, aliases will be removed from ~/.bashrc (if present)"
+echo "and the following file and directory will be removed:"
+echo "  ${shortcut_path}"
+echo "  ${PREFIX}"
+echo "Do you wish to continue?"
+read -p " [yes|NO]: " confirm
+if [[ $confirm != [yY] && $confirm != [yY][eE][sS] ]]; then
+    echo "Uninstall aborted."
+    exit 1
+fi
 rm -rf ${shortcut_path}
 rm -rf ${PREFIX}
 EOF
--- a/installers-conda/resources/post-install.sh
+++ b/installers-conda/resources/post-install.sh
@@ -80,9 +80,9 @@ $ uninstall-spyder
 # !!! IMPORTANT !!! #
 #####################
 
-The spyder and uninstall-spyder command-line shortcuts will only be available
-in new shell sessions. To make them available in this session you must source
-your .bashrc file
+The spyder and uninstall-spyder commands will only be available in new shell
+sessions. To make them available in this session you must source your .bashrc
+file with:
 
 $ source ~/.bashrc
 
--- a/installers-conda/resources/post-install.sh
+++ b/installers-conda/resources/post-install.sh
@@ -76,6 +76,16 @@ To uninstall Spyder, you need to run from the following from the command line:
 
 $ uninstall-spyder
 
+#####################
+# !!! IMPORTANT !!! #
+#####################
+
+The spyder and uninstall-spyder command-line shortcuts will only be available
+in new shell sessions. To make them available in this session you must source
+your .bashrc file
+
+$ source ~/.bashrc
+
 ###############################################################################
 
 EOF
--- a/installers-conda/resources/post-install.sh
+++ b/installers-conda/resources/post-install.sh
@@ -47,7 +47,7 @@ echo "  ${PREFIX}"
 echo ""
 echo "Do you wish to continue?"
 read -p " [yes|NO]: " confirm
-if [[ $confirm != [yY] && $confirm != [yY][eE][sS] ]]; then
+if [[ \$confirm != [yY] && \$confirm != [yY][eE][sS] ]]; then
     echo "Uninstall aborted."
     exit 1
 fi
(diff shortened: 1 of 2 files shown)


mrclary:
@ccordoba12, I'm not sure if I reproduced the issue you reported regarding `spyder` and `uninstall-spyder` aliases.

If the aliases are not present in `~/.bashrc` and I then install spyder from an existing terminal, the aliases are not available in the existing terminal since the changes to `~/.bashrc` have not be sourced. The aliases are available in a new terminal window, however, because `~/.bashrc` is sourced again. However, I think you reported that you tried this and were not successful.

Are the aliases just not working or do they report an error?
The aliases should be written to `~/.bashrc`; are they there? If so, could there be a reason that it's not being sourced?

Another possibility is that the alias(es) are there from an older installation, pointing to another (older) executable that may or may not exist and the most recent installation did not correctly replace the aliases...

ccordoba12:
> If the aliases are not present in ~/.bashrc and I then install spyder from an existing terminal, the aliases are not available in the existing terminal since the changes to ~/.bashrc have not be sourced. The aliases are available in a new terminal window, however, because ~/.bashrc is sourced again. However, I think you reported that you tried this and were not successful.

I checked again and you're right, the aliases work when you open a new terminal. So perhaps I didn't open one while I was testing last time, sorry (I didn't expect that to be necessary).

> If so, could there be a reason that it's not being sourced?

Yeah, I think that's the main problem. Could you make that the installer automatically sources `.bashrc` before it finishes? That would solve the problem for me.

ccordoba12:
Another thing that I think it's important to address here is that the uninstall script is not asking the user if they really want to uninstall Spyder before proceeding. Since that's an operation that can't be reverted, I think we should be sure that that's what the user wants.

mrclary:
> I checked again and you're right, the aliases work when you open a new terminal. So perhaps I didn't open one while I was testing last time, sorry (I didn't expect that to be necessary).
>
> Could you make that the installer automatically source `.bashrc` before it finishes? That would solve the problem for me.

Sure thing.

> Another thing that I think it's important to address here is that the uninstall script is not asking the user if they really want to uninstall Spyder before proceeding. Since that's an operation that can't be reverted, I think we should be sure that that's what the user wants.

Good idea.

I'll go ahead and put these into this PR.

mrclary:
> Could you make that the installer automatically source .bashrc before it finishes? That would solve the problem for me.

Oops, I don't think this is possible. The installer is a shell script, so anything it does cannot change the existing shell that launched it, including the post-install script. I think the only thing you can do is manually source `~/.bashrc` in your shell session after the installer completes.
```
$ source ~/.bashrc
```

ccordoba12:
> I think the only thing you can do is manually source ~/.bashrc in your shell session after the installer completes.

Ok, in that case I suggest to add a message at the end of the installer telling users that they need to do that or start a new console to launch Spyder.

Since it's important that they don't miss that message, I also suggest to enclose it in a rectangle of `#` symbols, like this

```
#####################
# This is important #
#####################
```

installers-conda/resources/post-install.sh
```diff
@@ -65,6 +76,16 @@ To uninstall Spyder, you need to run from the following from the command line:
 
 $ uninstall-spyder
 
+#####################
+# !!! IMPORTANT !!! #
+#####################
+
+The spyder and uninstall-spyder command-line shortcuts will only be available
+in new shell sessions. To make them available in this session you must source
+your .bashrc file
```
ccordoba12, line 85:
```suggestion
The spyder and uninstall-spyder commands will only be available in new shell
sessions. To make them available in this session you must source your .bashrc
file with:
```

* I think `shortcuts` sounds a bit strange in a Linux context, but perhaps you used it because you wrote `command-line` before that term. So, I suggest to remove `command-line shortcuts` and just leave `commands` instead.
* Add `with` at the end of this paragraph to make it a bit easier to read.

mrclary: Unless you think `command-line aliases` is better. But I'm fine with your suggestion as well.

ccordoba12: I thought about that too, but I think the simpler this message the better (I guess several users don't know what an `alias` is).

ccordoba12 (commented):
I tested the installer locally and think that the uninstall message can be improved with my small suggestion below.

In addition, I found that the uninstall script is not working for me. It fails with the following message:

```
/home/test/.local/Spyder-5.4.4.dev29/uninstall-spyder.sh: line 9: conditional binary operator expected
/home/test/.local/Spyder-5.4.4.dev29/uninstall-spyder.sh: line 9: syntax error near `[yY]'
/home/test/.local/Spyder-5.4.4.dev29/uninstall-spyder.sh: line 9: `if [[  != [yY] &&  != [yY][eE][sS] ]]; then'
```

mrclary:
> In addition, I found that the uninstall script is not working for me. It fails with the following message:
> 
> ```
> /home/test/.local/Spyder-5.4.4.dev29/uninstall-spyder.sh: line 9: conditional binary operator expected
> /home/test/.local/Spyder-5.4.4.dev29/uninstall-spyder.sh: line 9: syntax error near `[yY]'
> /home/test/.local/Spyder-5.4.4.dev29/uninstall-spyder.sh: line 9: `if [[  != [yY] &&  != [yY][eE][sS] ]]; then'
> ```

@ccordoba12, I think I know what's happening here. I'll fix.

mrclary:
Sorry, should work now.

ccordoba12 (approved):

PR #20868, opened 2023-04-24
PR: Improve Variable Explorer UX

jsbautista:
## Description of Changes

Fixes variable explorer UX issues

Depends on https://github.com/spyder-ide/spyder-kernels/pull/460

### Issue(s) Resolved

Fixes https://github.com/spyder-ide/ux-improvements/issues/17

### Affirmation

@jsbautista


jsbautista:
Preview:
VariableExplorer

jsbautista:
Preview:
VariableExplorerPreview

dalthviz (commented):
Thanks @jsbautista for the work here! Left an initial review with a comment regarding the changes related with `spyder-kernels` and a comment suggestion.

If you have any question let me know!

spyder/config/main.py
```diff
@@ -180,7 +185,8 @@
               'truncate': True,
               'minmax': False,
               'show_callable_attributes': True,
-              'show_special_attributes': False
+              'show_special_attributes': False,
+              'filter_on': False
```
ccordoba12, line 189: The filter needs to be on by default (contrary to what was mentioned on issue https://github.com/spyder-ide/ux-improvements/issues/17). Sorry if that made you do more work than necessary, probably we got confused when we discussed about it some years ago.

That's because if the filter is not on, then the Variable Explorer would show too many unnecessary variables to users (like private IPython references that users don't need to touch, functions and modules).

spyder/plugins/variableexplorer/plugin.py
```diff
@@ -47,7 +47,19 @@ def get_icon(self):
         return self.create_icon('dictedit')
 
     def on_initialize(self):
-        pass
+        widget = self.get_widget()
+        widget.sig_open_preferences_requested.connect(
+            self._open_interpreter_preferences
```
ccordoba12, line 52:
```suggestion
            self._open_preferences
```

The name is not correct.

spyder/plugins/variableexplorer/plugin.py
```diff
@@ -47,7 +47,19 @@ def get_icon(self):
...
+        )
+
+    def _open_interpreter_preferences(self):
+        """Open the Preferences dialog in the variable explorer section."""
```
ccordoba12, line 56:
```suggestion
    def _open_preferences(self):
        """Open the Preferences dialog in the Variable Explorer section."""
```

ccordoba12: Please also add a new `Private API` block comment below and move this method to that section.

spyder/plugins/variableexplorer/widgets/main_widget.py
```diff
@@ -94,9 +100,16 @@ class VariableExplorerWidget(ShellConnectMainWidget):
...
 
+    sig_open_preferences_requested = Signal()
+    """
+    Signal to open the variable explorer preferences.
+    """
+
     def __init__(self, name=None, plugin=None, parent=None):
         super().__init__(name, plugin, parent)
 
+        self.filter_on = False
```
ccordoba12, line 111:
```suggestion
        self.filter_on = True
```

This should be True by default, as I mentioned above.

spyder/plugins/variableexplorer/widgets/main_widget.py
```diff
@@ -108,47 +121,52 @@ def get_title(self):
...
     def setup(self):
         # ---- Options menu actions
-        exclude_private_action = self.create_action(
+        self.exclude_private_action = self.create_action(
             VariableExplorerWidgetActions.ToggleExcludePrivate,
             text=_("Exclude private variables"),
             tip=_("Exclude variables that start with an underscore"),
             toggled=True,
             option='exclude_private',
+            initial=self.get_conf('exclude_private_preferences'),
```
ccordoba12, line 130: This shouldn't be necessary either, as I mentioned above.

spyder/plugins/variableexplorer/widgets/main_widget.py
```diff
@@ -203,17 +221,25 @@ def setup(self):
             triggered=self.refresh_table,
             register_shortcut=True,
         )
+        self.filter_button = self.create_action(
```
ccordoba12, line 224:
```suggestion

        self.filter_button = self.create_action(
```

spyder/plugins/variableexplorer/widgets/main_widget.py
```diff
@@ -309,20 +342,31 @@ def setup(self):
...
+                     self.exclude_capitalized_action,
+                     self.exclude_unsupported_action,
+                     self.exclude_callables_and_modules_action,
                      self.show_minmax_action]:
             self.add_item_to_menu(
                 item,
                 menu=options_menu,
                 section=VariableExplorerWidgetOptionsMenuSections.Display,
             )
+        self.filter_variables()
```
ccordoba12, line 356:
```suggestion

        self.filter_variables()

```

spyder/plugins/variableexplorer/widgets/main_widget.py
```diff
@@ -406,6 +450,20 @@ def switch_widget(self, nsb, old_nsb):
     # ---- Public API
     # ------------------------------------------------------------------------
 
+    def change_filter_state(self, value):
```
ccordoba12, line 453: Please make this method private and move it to the Private API section below.

spyder/plugins/variableexplorer/widgets/main_widget.py
```diff
@@ -406,6 +450,20 @@ def switch_widget(self, nsb, old_nsb):
...
+        """Handle the change of the filter state."""
+        self.filter_on = not self.filter_on
+        self.filter_button.setChecked(self.filter_on)
+        self.filter_button.setToolTip(_("Filter variables"))
+        self.filter_variables()
+
+    def filter_variables(self):
```
ccordoba12, line 460: Let's rename this method to `_enable_filter_actions` and move it to the Private API section.

spyder/widgets/collectionseditor.py
```diff
@@ -916,6 +924,10 @@ def mouseDoubleClickEvent(self, event):
         else:
             event.accept()
 
+    def mouseMoveEvent(self, event):
+        """Change cursor shape."""
+        self.setCursor(Qt.PointingHandCursor)
```
ccordoba12, line 929: I really like this new interaction!! Great work @jsbautista!

However, I forgot to mention it to @conradolandia (sorry about that). Will do in our next UX/UI meeting and let you know what he thinks.

ccordoba12 (commented):
Thanks @jsbautista for your work on this! It looks really nice!!

spyder/config/main.py
```diff
@@ -172,6 +172,11 @@
               'dataframe_format': '.6g',  # No percent sign to avoid problems
                                           # with ConfigParser's interpolation
               'excluded_names': EXCLUDED_NAMES,
+              'exclude_private_preferences': True,
+              'exclude_uppercase_preferences': False,
+              'exclude_capitalized_preferences': False,
+              'exclude_unsupported_preferences': False,
+              'exclude_callables_and_modules_preferences': True,
```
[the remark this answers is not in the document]
dalthviz: The idea around that was to prevent the context menu options changing the options displayed in the preference page following the point `In the edit filters option in preferences the options currently activated by defaullt should still be checked by default.` listed at the end of https://github.com/spyder-ide/ux-improvements/issues/17.  We interpreted that point as having a default filters definition in the config page which doesn't change with the changes done in the context menu.

Is then not necessary to have a default filter options state from the preference page which is not affected for the current context menu checked filters? Or maybe the idea is correct (the preferences page being the default options and only changeable from the preference page) but another approach should be followed @ccordoba12 ?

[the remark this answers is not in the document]
ccordoba12: > Is then not necessary to have a default filter options state from the preference page which is not affected for the current context menu checked filters?

That's correct, I don't think it's necessary. That would make this functionality work contrary to how things do in the rest of our plugins.

> Or maybe the idea is correct (the preferences page being the default options and only changeable from the preference page) but another approach should be followed @ccordoba12 ?

Well, I think this would make things inconsistent, which would be confusing for users. For instance, if you check `Exclude capitalized variable` in the Options menu, and then click on the new `Edit filters` context menu entry, users wouldn't see that option checked in Preferences. So, I think we should leave things as they are working right now.

What would be interesting is to apply our filters to each stacked widget in the Variable Explorer (instead to all of them). But since users haven't requested for that functionality, there's no need to implement it just yet.

spyder/widgets/collectionseditor.py
ccordoba12: I talked to @conradolandia and @CAM-Gerlach about this in our last UX meeting and they also it's a great way to let users know that entries in the Variable Explorer are clickable.

Our only suggestion is to change the cursor shape to `PointingHandCursor` only when the cursor is on top of a variable row. Right now, this changes the shape as soon as the cursor is on top of the Variable Explorer.

spyder/widgets/collectionseditor.py
```diff
@@ -926,6 +934,10 @@ def mouseDoubleClickEvent(self, event):
...
```
dalthviz, line 939: In order to change the cursor only when there is a row being hovered we need to change this for something like:

```suggestion
        if self.rowAt(event.y()) != -1:
            self.setCursor(Qt.PointingHandCursor)
        else:
            self.setCursor(Qt.ArrowCursor)
```

dalthviz (commented):
Thanks @jsbautista for the work here! I left a suggestion for the comment regarding the mouse cursor change which seems that was missing. Also, we need to wait for https://github.com/spyder-ide/spyder-kernels/pull/460 to be merged so we can update the `spyder-kernels` subrepo again to the main repo. But, besides that this LGTM :+1:

spyder/plugins/variableexplorer/plugin.py
```diff
@@ -47,7 +47,19 @@ def get_icon(self):
         return self.create_icon('dictedit')
 
     def on_initialize(self):
-        pass
+        widget = self.get_widget()
+        widget.sig_open_preferences_requested.connect(
+            self._open_preferences
+        )
+
+    def _open_preferences(self):
```
ccordoba12, line 55: This method shouldn't be in the middle of the `SpyderDockablePlugin API` section. Instead, please add a `Private API` section below and move it there.

spyder/plugins/variableexplorer/widgets/main_widget.py
```diff
@@ -204,16 +218,25 @@ def setup(self):
...
+        self.filter_button = self.create_action(
+            VariableExplorerWidgetActions.ToggleFilter,
+            text="",
+            icon=ima.icon('filter'),
+            toggled=self._enable_filter_actions,
+            option='filter_on',
+            tip=_("Filter variables")
+            )
+        self.filter_button.setCheckable(True)
         # ---- Context menu actions
```
ccordoba12, line 230:
```suggestion

        # ---- Context menu actions
```

spyder/plugins/variableexplorer/widgets/main_widget.py
```diff
@@ -571,3 +613,11 @@ def _set_actions_and_menus(self, nsb):
...
         editor.insert_action_above = QAction()
         editor.insert_action_below = QAction()
+
+    def _enable_filter_actions(self, value):
+        """Handle the change of the filter state."""
+        self.exclude_private_action.setEnabled(value)
+        self.exclude_uppercase_action.setEnabled(value)
+        self.exclude_capitalized_action.setEnabled(value)
+        self.exclude_unsupported_action.setEnabled(value)
+        self.exclude_callables_and_modules_action.setEnabled(value)
```
ccordoba12, line 623:
```suggestion
        self.exclude_callables_and_modules_action.setEnabled(value)

```

Add empty line at the end.

spyder/plugins/variableexplorer/plugin.py
```diff
@@ -47,7 +47,19 @@ def get_icon(self):
...
+        """Open the Preferences dialog in the Variable Explorer section."""
+        self._main.show_preferences()
+        preferences = self._main.preferences
+        container = preferences.get_container()
+        dlg = container.dialog
+        index = dlg.get_index_by_name(self.NAME)
+        dlg.set_current_index(index)
```
ccordoba12, line 62:
```suggestion
        self._main.show_preferences()
        preferences = self.get_plugin(Plugins.Preferences)
        if preferences:
            container = preferences.get_container()
            dlg = container.dialog
            index = dlg.get_index_by_name(self.NAME)
            dlg.set_current_index(index)
```

* Use the SpyderPluginV2 API to get Preferences, which is preferred to grab it directly from the main window.
* Check if it's available before using it.

dalthviz: Just in case, I think this was done following the logic at https://github.com/spyder-ide/spyder/blob/4d51bddbf225a148d5b2796cef938cf69f0cf527/spyder/plugins/maininterpreter/plugin.py#L112-L119

So maybe a change there to use `get_plugin` there too for consistency could be worthy

ccordoba12: Yep, I saw that. If you want to do these changes there, you're welcome too (I probably implemented that before the Preferences plugin was available).

ccordoba12 (approved):
Thanks @jsbautista and @dalthviz for your work on this!

PR #20884, opened 2023-05-01
PR: Cleanup Environments for Conda-based Installers

mrclary:
## Description of Changes

* Ensured that the Python version for the runtime environment is the same as the build environment.
* Decluttered the `dist` directory by putting intermediate build resources in `build`.
* Base environment is unified across platforms as lower-case and only using major version number. Runtime environment is designated as `spyder-runtime` without version number.
* Created uninstall script for macOS.
* Improved logging for CI.


installers-conda/build_installers.py
```diff
@@ -226,18 +227,16 @@ def _definitions():
...
-            "pip",
         ],
         "installer_filename": OUTPUT_FILE.name,
         "initialize_by_default": False,
         "initialize_conda": False,
         "register_python": False,
         "license_file": str(RESOURCES / "bundle_license.rtf"),
         "extra_envs": {
-            f"spyder-{SPYVER}": {
+            "spyder-rt": {
```
ccordoba12, line 239: A couple of questions about this change:

* This environment won't appear in the list of envs listed in `Preferences > Main interpreter`, right?
* What if we name it `spyder-runtime` instead to a bit more explicit about it?

ccordoba12 (commented):
I left a couple for questions for you @mrclary, the rest looks good to me.

installers-conda/build_installers.py
mrclary: Correct, `spyder-6` (base) and `spyder-6/envs/spyder-rt` (spyder-rt) will not show up in the custom interpreter list.
No problem, I'll rename to `spyder-runtime`.

dalthviz (approved):
Thanks @mrclary ! LGTM :+1:

As a side note, when installing the generated installer executable for Windows locally, seems like the old conda based installation I had got its shortcut replaced (the shortcut I had for the old installation was now pointing to the newer installation). Maybe is because the shortcut is called `Spyder` regardless of the version you are installing?

mrclary:
> As a side note, when installing the generated installer executable for Windows locally, seems like the old conda based installation I had got its shortcut replaced (the shortcut I had for the old installation was now pointing to the newer installation). Maybe is because the shortcut is called `Spyder` regardless of the version you are installing?

This behavior is still unclear to me. When I have a previously installed conda-based instance, the shortcut is not replaced on macOS or Linux; it seems that menuinst does not overwrite the existing shortcuts. Perhaps this is different for Windows. I'll have to perform a deliberate investigation. If there is a difference between Windows and macOS and Linux, then we will need to ensure consistent behavior.

ccordoba12:
I tested the new changes locally and found a minor inconvenience that perhaps we should address before releasing alpha1 (the rest is working as expected).

The update widget is shown in the status bar, at first trying to find an update, like this:

Asking for updates

and then showing the current Spyder version.

Since the update mechanism hasn't been ported to the conda-based installers yet, I'd suggest to hide that widget for now.

mrclary:
> Since the update mechanism hasn't been ported to the conda-based installers yet, I'd suggest to hide that widget for now.

For conda-based installations, I disabled the update action and disabled checking at startup.

mrclary:
> > Since the update mechanism hasn't been ported to the conda-based installers yet, I'd suggest to hide that widget for now.
> 
> For conda-based installations, I disabled the update action and disabled checking at startup.

That caused problems, so I just blocked the signals of the application update status widget for the conda-based instances.

ccordoba12 (approved):
Thanks @mrclary!

PR #21031, opened 2023-06-15
PR: Prevent setting cursor in a negative position (Editor)

### LINKED ISSUE #21028: Minor bug with empty files: `QTextCursor::setPosition: Position '-1' out of range`
If an empty file is opened with Spyder when spyder is launched, a message is printed:

QTextCursor::setPosition: Position '-1' out of range

This message can be suppressed by changing one line in plugins/editor/widgets/codeeditor.py:
(around line 4973)
        # Text might be after a dot '.'
        if text == '':
            #was: cursor.setPosition(pos - 1, QTextCursor.MoveAnchor)
            cursor.setPosition(**max(0, pos - 1)**, QTextCursor.MoveAnchor)

Hope that helps.
Thank you

dalthviz:
## Description of Changes

When doing autocompletions there is some logic related with setting the cursor position to -1 which causes a warning when the cursor is already at position 0 (so the cursor position was being set to -1).

Preventing setting the position to negative values while handling the autocompletions prevents the warning message.

### Issue(s) Resolved

Fixes #21028 

### Affirmation

### DIFF
--- a/spyder/plugins/editor/widgets/codeeditor.py
+++ b/spyder/plugins/editor/widgets/codeeditor.py
@@ -4960,10 +4960,10 @@ class CodeEditor(TextEditBaseWidget):
         # Correctly handle completions when Backspace key is pressed.
         # We should not show the widget if deleting a space before a word.
         if key == Qt.Key_Backspace:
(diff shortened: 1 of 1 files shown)


ccordoba12 (approved):

PR #21053, opened 2023-06-21
PR: Fix issues for all-user install in post-install script (Installers)

mrclary:
* Quote variable interpolation in conditional statements and check for file not just exists.
* Remove aliases before removing environment in uninstall.
* Set executable permission for user only for `uninstall-spyder.sh`.
* Only add aliases if `shell_init` is not null.
* Use relative path in shortcut command for macOS to allow Spyder.app to be moved. `menuinst` currently puts shortcut in `~/Applications` instead of `/Applications` for all-user install.


ccordoba12 (approved):

PR #21062, opened 2023-06-22
PR: Fix issue where cmd.exe window flashes on Spyder startup on Windows (installer)

mrclary:
`menuinst` still presents a temporary `cmd.exe` window on Spyder startup. See conda/menuinst#128.
This can be eliminated by restricting the Windows shortcut link target to use `pythonw.exe`.
However, this comes with the tradeoff that the environment cannot be activated and additional pre-activation commands cannot be provided.

The pre-commands are not required by Spyder. Since Spyder application does not require intrinsic conda environment manipulation, then activating the runtime environment in the runtime shell environment is also not strictly required.

### DIFF
--- a/installers-conda/build_installers.py
+++ b/installers-conda/build_installers.py
@@ -132,8 +132,6 @@ SPYVER = get_version(SPYREPO, normalize=False).lstrip('v').split("+")[0]
 specs = {
     "python": "=" + PY_VER,
     "spyder": "=" + SPYVER,
-    "paramiko": "",
-    "pyxdg": "",
 }
 specs.update(scientific_packages)
 
--- a/installers-conda/build-environment.yml
+++ b/installers-conda/build-environment.yml
@@ -7,5 +7,6 @@ dependencies:
   - conda-standalone
   - constructor
   - gitpython
+  - menuinst
   - ruamel.yaml.jinja2
   - setuptools_scm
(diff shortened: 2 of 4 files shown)


installers-conda/build_installers.py
```diff
@@ -132,8 +132,8 @@
 specs = {
     "python": "=" + PY_VER,
     "spyder": "=" + SPYVER,
-    "paramiko": "",
-    "pyxdg": "",
+    "paramiko": "",  # ??? Do we need this?
+    "pyxdg": "",  # ??? Do we need this?
```
ccordoba12, line 136: * Paramiko is required to establish SSH connections on remote servers from Windows hosts.
* Pyxdg is necessary to detect the list of installed applications on Windows. That's used to associate file types to be open in certain applications in Files.

    image

ccordoba12 (commented):
Just a small clarifying comment from me. I'll leave testing if effectively no cmd window is shown now to @dalthviz.

installers-conda/build_installers.py
mrclary: Thank you for looking at this. My question is related to whether they need to be explicitly listed here. I think that since these packages are already listed as Spyder dependencies elsewhere, I suspect that they do not need to be explicitly listed here for the runtime environment.

dalthviz (approved):
Thanks @mrclary ! Checked locally and no cmd was shown :tada: so this LGTM :+1:

PR #21065, opened 2023-06-23
PR: Fix issue getting user environment variables on Posix systems and pass them to the IPython console

### LINKED ISSUE #3891: Add a way to read, display and manage the environment variables passed to the IPython console
This is needed to make Spyder work correctly with Tensorflow without starting it from a terminal. For example see:

http://stackoverflow.com/questions/41344017/why-does-the-tensorflow-module-import-fail-in-spyder-and-not-in-jupyter-notebo

----

@andfoy, this is for 3.2, so please don't worry about it right now. I opened this issue so we don't forget to address it in the future ;-)

mrclary:
## Description of Changes

* Use a shell script in Spyder's config directory to retriev user environment variables on posix systems.
* The shell script is created on Spyder startup and only for posix systems.

To get user environment variables from login and interactive startup files on posix, both `-l` and `-i` flags must be used. Parsing the environment list is problematic if the variable has a newline character, but Python's `os.environ` can do this for us. However, executing Python in an interactive shell in a subprocess causes Spyder to hang on Linux. Using a shell script resolves the issue. Note that `-i` must be in the sha-bang line; if `-i` and `-l` are swapped, Spyder will hang.

Also note that for environment variables for `KernelSpec` objects are first set to the user's environment variables, then _updated_ with `os.environ`. Thus user login and interactive variables are ensured to be present, but values in `os.environ` may override user variable values.

### Issue(s) Resolved

Part of #3891

### Affirmation

### DIFF
--- a/spyder/utils/environ.py
+++ b/spyder/utils/environ.py
@@ -9,6 +9,7 @@ Environment variable utilities.
 """
 
 # Standard library imports
+from functools import lru_cache
 import logging
 import os
 from pathlib import Path
@@ -30,7 +31,9 @@ from spyder.utils.programs import run_shell_command
 
 logger = logging.getLogger(__name__)
 
-if os.name == 'posix':
+
+@lru_cache
+def _get_user_env_script():
     """
     To get user environment variables from login and interactive startup files
     on posix, both -l and -i flags must be used. Parsing the environment list
@@ -43,10 +46,13 @@ if os.name == 'posix':
     script_text = None
     shell = os.getenv('SHELL', '/bin/bash')
     user_env_script = Path(get_conf_path()) / 'user-env.sh'
+
     if Path(shell).name in ('bash', 'zsh'):
         script_text = (
             f"#!{shell} -i\n"
-            f"""{shell} -l -c "{sys.executable} -c 'import os; print(dict(os.environ))'"\n"""
+            f"{shell} -l -c"
+            f" \"{sys.executable} -c 'import os; print(dict(os.environ))'\" "
+            f"\n"
         )
     else:
         logger.info("Getting user environment variables is not supported "
@@ -54,7 +60,9 @@ if os.name == 'posix':
 
     if script_text is not None:
         user_env_script.write_text(script_text)
-        user_env_script.chmod(0o744)  # Make executable by user
+        user_env_script.chmod(0o744)  # Make it executable for the user
+
+    return str(user_env_script)
 
 
 def envdict2listdict(envdict):
@@ -90,9 +98,10 @@ def get_user_environment_variables():
         env_var = dict(
             [winreg.EnumValue(key, k)[:2] for k in range(num_values)]
         )
-    elif os.name == 'posix' and user_env_script.exists():
+    elif os.name == 'posix':
         try:
-            proc = run_shell_command(str(user_env_script), env={}, text=True)
+            user_env_script = _get_user_env_script()
+            proc = run_shell_command(user_env_script, env={}, text=True)
             stdout, stderr = proc.communicate()
             if stderr:
                 logger.info(stderr.strip())
--- a/spyder/utils/environ.py
+++ b/spyder/utils/environ.py
@@ -54,24 +54,30 @@ def get_user_environment_variables():
     env_var : dict
         Key-value pairs of environment variables.
     """
-    try:
-        if os.name == 'nt':
-            key = winreg.OpenKey(winreg.HKEY_CURRENT_USER, "Environment")
-            num_values = winreg.QueryInfoKey(key)[1]
-            env_var = dict(
-                [winreg.EnumValue(key, k)[:2] for k in range(num_values)]
-            )
-        else:
-            shell = os.environ.get("SHELL", "/bin/bash")
+    env_var = {}
+    if os.name == 'nt':
+        key = winreg.OpenKey(winreg.HKEY_CURRENT_USER, "Environment")
+        num_values = winreg.QueryInfoKey(key)[1]
+        env_var = dict(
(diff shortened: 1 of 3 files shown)


spyder/utils/environ.py
```diff
@@ -54,24 +83,23 @@ def get_user_environment_variables():
...
+    env_var = {}
+    if os.name == 'nt':
+        key = winreg.OpenKey(winreg.HKEY_CURRENT_USER, "Environment")
+        num_values = winreg.QueryInfoKey(key)[1]
+        env_var = dict(
+            [winreg.EnumValue(key, k)[:2] for k in range(num_values)]
+        )
+    elif os.name == 'posix' and user_env_script.exists():
+        try:
+            proc = run_shell_command(str(user_env_script), env={}, text=True)
```
ccordoba12, line 95:
```suggestion
            user_env_script = _get_user_env_script()
            proc = run_shell_command(user_env_script, env={}, text=True)
```

spyder/utils/environ.py
```diff
@@ -9,6 +9,7 @@
 """
 
 # Standard library imports
+import logging
```
ccordoba12, line 12:
```suggestion
from functools import cache
import logging
```

spyder/utils/environ.py
```diff
@@ -22,11 +23,39 @@
...
+    on posix, both -l and -i flags must be used. Parsing the environment list
+    is problematic if the variable has a newline character, but Python's
+    os.environ can do this for us. However, executing Python in an interactive
+    shell in a subprocess causes Spyder to hang on Linux. Using a shell script
+    resolves the issue. Note that -i must be in the sha-bang line; if -i and
+    -l are swapped, Spyder will hang.
+    """
+    script_text = None
+    shell = os.getenv('SHELL', '/bin/bash')
+    user_env_script = Path(get_conf_path()) / 'user-env.sh'
```
ccordoba12, line 45: We could also write this script in the user temp directory. Why do you think it's better to write it in `conf_path`?

ccordoba12 (commented):
Thanks @mrclary for your work on this!

ccordoba12:
Note: This doesn't completely solve issue #3891 because that would require adding a tab to the IPython console preferences so that users can add env vars without editing their `bashrc` or similar files.

spyder/utils/environ.py
mrclary: I don't feel strongly one way or the other. My reasoning was that we don't have control over when a temp file is removed. Presumably, this will only occur when the user logs out or the system is rebooted. Since Spyder will only create the file once per session, but requires the file throughout the session, is there a risk that the file will be purged during a Spyder session?

spyder/utils/environ.py
```diff
@@ -22,11 +23,39 @@
...
 # Local imports
-from spyder.config.base import _, running_in_ci
+from spyder.config.base import _, running_in_ci, get_conf_path
 from spyder.widgets.collectionseditor import CollectionsEditor
 from spyder.utils.icon_manager import ima
 from spyder.utils.programs import run_shell_command
 
+logger = logging.getLogger(__name__)
+
+if os.name == 'posix':
```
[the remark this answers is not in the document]
mrclary: Cool. I was not aware of this decorator.

mrclary:
> Note: This doesn't completely solve issue #3891 because that would require adding a tab to the IPython console preferences so that users can add env vars without editing their `bashrc` or similar files.

I suppose this is a UX issue and maybe we could discuss it more at one of our dev meetings. I see a few additional ideas:
1. We _could_ modify the user's `~/.bashrc`, similar to how we add the aliases for the Linux installer. We could just delimit a section like `conda` does, e.g. `>>> Spyder Injected Variables >>> ... <<< Spyder Injected Variables <<<`, or similar. This would then provide similar functionality to the user environment variables widget as we have for Windows.
2. Alternatively, we could treat the user environment variables similar to how we manage the PYTHONPATH. We could show what is provided by the system, allow the user to add and/disable variables, but these don't get applied upstream to the system, only to downstream processes. The current Windows functionality does not conform to this, so maybe we consider changing it in this scenario.

spyder/utils/environ.py
```diff
@@ -9,6 +9,8 @@
 """
 
 # Standard library imports
+from functools import cache
```
ccordoba12, line 12:
```suggestion
from functools import lru_cache
```

Sorry, the `cache` decorator was added in Python 3.9, so we need to use `lru_cache` instead.

spyder/utils/environ.py
```diff
@@ -22,11 +24,46 @@
...
-from spyder.config.base import _, running_in_ci
+from spyder.config.base import _, running_in_ci, get_conf_path
 from spyder.widgets.collectionseditor import CollectionsEditor
 from spyder.utils.icon_manager import ima
 from spyder.utils.programs import run_shell_command
 
+logger = logging.getLogger(__name__)
+
+
+@cache
```
ccordoba12, line 35:
```suggestion
@lru_cache
```

spyder/utils/environ.py
ccordoba12: > I don't feel strongly one way or the other. My reasoning was that we don't have control over when a temp file is removed

Ok, I think it's fine to leave it in `conf_path` then.

ccordoba12:
> I suppose this is a UX issue and maybe we could discuss it more at one of our dev meetings

Agreed. We need to solve that in another PR.

> Alternatively, we could treat the user environment variables similar to how we manage the PYTHONPATH.

I like this idea. That way the env vars managed by us would be propagated to the LSP, IPython console and all other things we run in external processes.

ccordoba12 (approved):
Great work here @mrclary! I tested this locally and it's working as expected.

PR #21083, opened 2023-06-29
PR: Fix issue where bootstrap incorrectly determines git branch

mrclary:
Calling `python .../spyder/bootstrap.py` when the current working directory is not the Spyder repo results in incorrectly determining the git branch. This PR fixes that.


ccordoba12 (approved):
Thanks @mrclary!

PR #20933, opened 2023-05-17
PR: Make the font used by the application configurable and other UI fixes

ccordoba12:
## Description of Changes

- Replace `Rich text` entry in Preferences by `Interface` and rename `Plain text` to `Monospace`.
- Add option to be able to change the app font. That was done so that Spyder follows the same font used by other programs in the system by default.
- Make this change to require a restart because it can't be applied while the app is running.
- Fix pixelated splash screen when using scale factors greater than 1 (this happens for certain screen aspect ratios).
- Remove setting font for labels of status bar widgets (this was different from the app font).
- Switch setting sizes from em's to pixels in most places. That's because I discovered that Qt adjusts em sizes according to the app font, but that's not what we want to have (for instance) in the main and pane toolbars.
- Remove hard-coded font sizes in the "Report error" dialog and improve its style a bit by correctly aligning its widgets to the left and right.
- Re-add the previous cell icon because it's used in the Outline pane and rename the one to create cells to `new_cell`.
- Add `SpyderFontType` enum to access the different fonts we have in Spyder.
- Add `SpyderFontsMixin` to get those fonts as QFont objects to be used in Qt widgets or other objects.
- Make `Online Help` use the font size of the rest of the app (this was not working).
- Remove vertical header in the Variable Explorer because it's not really necessary. Also, adjust its columns size when shown so that it looks good at startup.
- Fix setting the thumbnail scroll bar's size when `Plots` is shown (the previous hack was not working anymore).
- Add a new monospace interface font type, which is the same monospace font selected by users in Preferences, but with a size that matches the rest of the interface.
- Use monospace font for the `Index` column in our dataframe editor. This was reported by a user to be a UI inconsistency and I agree with them.

----

Visual changes:

* Replace the previous `Rich text` entry in Preferences with an `Interface` one:

    imagen

* Fonts and layout of the "Report error" dialog were not the best

    **Before**

    image

    **After**

    image

* Restore cell icon in the Outline pane

    **Before**

    image

    **After**

    image

* Resize Variable Explorer columns at startup

     **Before**

    imagen

    **After**

    imagen

* Hide vertical header in Variable Explorer because it's not necessary.

    **Before**

    imagen

    **After**

    imagen

* Use the monospace font for the `Index` column in our Dataframe editor

    **Before**

    imagen

    **After**

    imagen

* Scale size of monospace font to match the interface one when the latter is changed. That's applied to plugins that are not directly related to code, like the Variable Explorer.

    imagen

### Issue(s) Resolved

[...]

PR #21084, opened 2023-06-29
PR: Automatically launch Spyder after installation (Installers)

mrclary:
Automatically launch Spyder after install

### DIFF
--- a/installers-conda/resources/post-install.sh
+++ b/installers-conda/resources/post-install.sh
@@ -117,7 +117,7 @@ if [[ -n "$shell_init" ]]; then
 fi
 
 # ----
-if [[ $OSTYPE = "linux"* ]]; then
+if [[ "$OSTYPE" = "linux"* ]]; then
     cat <<EOF
 
 ###############################################################################
@@ -147,8 +147,22 @@ echo "*** Post install script for ${INSTALLER_NAME} complete"
 
 # ----
 [[ -n "$CI" ]] && exit 0  # Running in CI, don't launch Spyder
+
 if [[ "$OSTYPE" = "darwin"* ]]; then
-    open -a $shortcut_path
+    tmp_dir=${TMPDIR}spyder
+    launch_script=${tmp_dir}/post-install-launch.sh
+    echo "Creating post-install launch script ..."
+    mkdir -p $tmp_dir
+    cat <<EOF > $launch_script
+#!/bin/bash
+while [[ \$(pgrep -fq Installer.app) ]]; do
+    sleep 1
+done
+open -a $shortcut_path
+EOF
+    chmod +x $launch_script
+
+    nohup $launch_script &>/dev/null &
 elif [[ "$XDG_CURRENT_DESKTOP" =~ .*(Unity|GNOME|XFCE).* ]]; then
     gtk-launch spyder_spyder
 fi
--- a/installers-conda/resources/post-install.sh
+++ b/installers-conda/resources/post-install.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -i
 set -e
 
 echo "*** Running post install script for ${INSTALLER_NAME} ..."
--- a/installers-conda/resources/post-install.sh
+++ b/installers-conda/resources/post-install.sh
@@ -149,13 +149,13 @@ echo "*** Post install script for ${INSTALLER_NAME} complete"
 [[ -n "$CI" ]] && exit 0  # Running in CI, don't launch Spyder
 
 if [[ "$OSTYPE" = "darwin"* ]]; then
-    tmp_dir=${TMPDIR}spyder
+    tmp_dir=${SHARED_INSTALLER_TEMP}/spyder
     launch_script=${tmp_dir}/post-install-launch.sh
     echo "Creating post-install launch script ..."
     mkdir -p $tmp_dir
     cat <<EOF > $launch_script
 #!/bin/bash
-while [[ \$(pgrep -fq Installer.app) ]]; do
+while pgrep -fq Installer.app; do
     sleep 1
 done
 open -a $shortcut_path
--- a/installers-conda/resources/post-install.sh
+++ b/installers-conda/resources/post-install.sh
@@ -1,5 +1,6 @@
 #!/bin/bash -i
 set -e
+unset HISTFILE
 
 echo "*** Running post install script for ${INSTALLER_NAME} ..."
 
@@ -149,6 +150,7 @@ echo "*** Post install script for ${INSTALLER_NAME} complete"
 # ----
 [[ -n "$CI" ]] && exit 0  # Running in CI, don't launch Spyder
 
+echo "Launching Spyder now..."
 if [[ "$OSTYPE" = "darwin"* ]]; then
     tmp_dir=${SHARED_INSTALLER_TEMP}/spyder
     launch_script=${tmp_dir}/post-install-launch.sh
@@ -167,5 +169,5 @@ EOF
 elif [[ -n "$(which gtk-launch)" ]]; then
     gtk-launch $(basename $shortcut_path)
 else
-    $spy_exe
+    nohup $spy_exe &>/dev/null &
 fi
--- a/installers-conda/resources/post-install.sh
+++ b/installers-conda/resources/post-install.sh
@@ -63,20 +63,31 @@ add_alias() (
 
 # ----
 echo "Creating uninstall script..."
-cat <<EOF > ${u_spy_exe}
+cat <<END > ${u_spy_exe}
 #!/bin/bash
 
-echo "You are about to uninstall Spyder."
-echo "If you proceed, aliases will be removed from ~/.bashrc (if present)"
-echo "and the following will be removed:"
-echo "  ${shortcut_path}"
-echo "  ${PREFIX}"
-echo ""
-echo "Do you wish to continue?"
-read -p " [yes|NO]: " confirm
-if [[ \$confirm != [yY] && \$confirm != [yY][eE][sS] ]]; then
-    echo "Uninstall aborted."
-    exit 1
+while getopts "f" option; do
+    case "\$option" in
+        (f) force=true ;;
+    esac
+done
+shift \$((\$OPTIND - 1))
+
+if [[ -z \$force ]]; then
+    cat <<EOF
+You are about to uninstall Spyder.
+If you proceed, aliases will be removed from ${shell_init}
+(if present) and the following will be removed:
+  ${shortcut_path}
+  ${PREFIX}
+
+Do you wish to continue?
+EOF
+    read -p " [yes|NO]: " confirm
+    if [[ \$confirm != [yY] && \$confirm != [yY][eE][sS] ]]; then
+        echo "Uninstall aborted."
+        exit 1
+    fi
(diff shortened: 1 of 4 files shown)


ccordoba12:
@mrclary, I pushed directly to your branch a commit to improve the message we show on Linux when the installer finishes. This was bothering me a bit the last times I tested the installer and it was not easy to leave a suggestion for it here.

installers-conda/resources/post-install.sh
```diff
@@ -109,34 +120,35 @@ fi
...
 EOF
 fi
 
 echo "*** Post install script for ${INSTALLER_NAME} complete"
+
+# ----
+[[ -n "$CI" ]] && exit 0  # Running in CI, don't launch Spyder
+if [[ "$OSTYPE" = "darwin"* ]]; then
+    open -a $shortcut_path
+elif [[ "$XDG_CURRENT_DESKTOP" =~ .*(Unity|GNOME|XFCE).* ]]; then
```
ccordoba12, line 152: I think this simply should check if `gtk-launch` is available in PATH and then use it below to launch Spyder. That's because in my case I use KDE and I have `gtk-launch` installed, but due to this check Spyder is not launched automatically for me.

ccordoba12 (commented):
@mrclary, one small comment from me, the rest looks good.

I also forgot to mention that if you have comments or suggestions about the changes I pushed to your branch, please let me know about them.

dalthviz (approved):
Thanks @mrclary ! Checked locally on Windows and Spyder launches automatically! :+1:

However, if understand correctly. with this Spyder will always open after installation? There is no possibility to have a user confirmation before launch Spyder, right? Also, seems like the installer window is still present after Spyder opens and you have to click next to see the last installer page and close it:

image

Other than those pointers, which probably are caused do to the way `constructor` post install script works on Windows so there is not much we can do here about that, this LGTM :+1:

installers-conda/resources/post-install.sh
mrclary: Do you have `kde-open`? If so, can you test to see if that works?
```
$ kde-open spyder_spyder
```
Ideally, I'd like to start Spyder using the application shortcut in all circumstances, if possible, including for the `~/.bashrc` alias. Perhaps several fall-back options: `gtk-launch`, then `kde-open` (or desktop environment equivalent), then direct execution (`path/to/spyder`).

mrclary:
@dalthviz, thanks for reviewing this PR.

> However, if understand correctly. with this Spyder will always open after installation? There is no possibility to have a user confirmation before launch Spyder, right?

Correct. I think this could be fixed for Windows if we used an nsis template. I don't think `constructor` will ever adopt an application launcher mechanism so I don't think this will be a possibility for macOS and Linux, unless we could patch that into constructor ourselves.

> Also, seems like the installer window is still present after Spyder opens and you have to click next to see the last installer page and close it

That is correct. I may be able to devise a solution for all platforms to delay launching Spyder until the installer is closed. 👀

installers-conda/resources/post-install.sh
ccordoba12: I have `kde-open5` but that command doesn't launch Spyder (I tried several ways to use but none worked for me). However, running `gtk-launch spyder_spyder` did the trick.

mrclary:
> > Also, seems like the installer window is still present after Spyder opens and you have to click next to see the last installer page and close it
> 
> That is correct. I may be able to devise a solution for all platforms to delay launching Spyder until the installer is closed. 👀

@dalthviz, I just pushed up another commit that may resolve this issue. Another process is started that waits for the graphical installer to exit before starting Spyder.

installers-conda/resources/post-install.sh
mrclary: Okay, got it. So I will assume that `gtk-launch` is the only reliable desktop application launcher. The script will check for `gtk-launch`, and fall back to direct execution if it is not present.

mrclary: Pushed up changes.

ccordoba12: > Okay, got it. So I will assume that gtk-launch is the only reliable desktop application launcher.

Yep, it seems that way. I searched for KDE-only alternatives but I couldn't find one.

> The script will check for gtk-launch, and fall back to direct execution if it is not present.

Sounds good to me.

ccordoba12:
I found a bug while testing this (related to PR #20933) and I solve it on PR #21092.

Besides that, now Spyder launches after installation (which is great!), but the `spyder` bash alias is broken now with this error:

```bash
$ spyder
gtk-launch: missing application name
Try "gtk-launch --help" for more information.
```

`uninstall-spyder` is also broken. It simply reports `Uninstall aborted` at the end.

mrclary:
> I found a bug while testing this (related to PR #20933) and I solve it on PR #21092.

I just noticed that. Strange that it doesn't throw that error when bootstrapping...

> Besides that, now Spyder launches after installation (which is great!), but the `spyder` bash alias is broken with this error:
> 
> ```shell
> $ spyder
> gtk-launch: missing application name
> Try "gtk-launch --help" for more information.
> ```
> 
> I'd prefer to not use `gtk-launch` to launch Spyder in the command line, especially to pass cli arguments to it. So, I think the previous implementation of that command was fine.

Ohh, that's right. Will revert.

> `uninstall-spyder` is also broken. It simply reports `Uninstall aborted` at the end.

I'll look at these errors. Thanks for investigating this @ccordoba12 😄

dalthviz:
> I just pushed up another commit that may resolve this issue. Another process is started that waits for the graphical installer to exit before starting Spyder.

Awesome! However, checking again, seems like a cmd shows up in the task bar and you can click on it (I guess is related with the waiting process?). Is there something that could be done to prevent that cmd from showing/being shown in the task bar?

install

Anyhow, I think is nice that now Spyder launches after the installer window finishes/is closed :+1:

mrclary:
@ccordoba12 @dalthviz, I think all the issues have been worked out now. Additionally, Spyder should delay starting until the graphical installer exits (macOS and Windows).

mrclary:
> However, checking again, seems like a cmd shows up in the task bar and you can click on it (I guess is related with the waiting process?). Is there something that could be done to prevent that cmd from showing/being shown in the task bar?

That is correct; it is the waiting process. It is not possible to hide the cmd.exe window, only minimize it under these circumstances. However, I could try using a powershell process instead; maybe that will allow the hidden window... 👀

installers-conda/resources/post-install.sh
```diff
@@ -106,37 +118,54 @@ if [[ -n "$shell_init" ]]; then
...
+done
+open -a $shortcut_path
+EOF
+    chmod +x $launch_script
+
+    nohup $launch_script &>/dev/null &
+elif [[ -n "$(which gtk-launch)" ]]; then
+    gtk-launch $(basename $shortcut_path)
+else
+    $spy_exe
```
ccordoba12, line 170:
```suggestion
elif [[ -n "$(which gtk-launch)" ]]; then
    echo "Launching Spyder now..."
    gtk-launch $(basename $shortcut_path)
else
    echo "Launching Spyder now..."
    $spy_exe
```

ccordoba12 (commented):
Last, small suggestion from me, the rest is working as expected. Thanks @mrclary!

installers-conda/resources/post-install.sh
```diff
@@ -106,37 +118,54 @@ if [[ -n "$shell_init" ]]; then
...
 
 EOF
 fi
 
 echo "*** Post install script for ${INSTALLER_NAME} complete"
+
+# ----
+[[ -n "$CI" ]] && exit 0  # Running in CI, don't launch Spyder
+
+if [[ "$OSTYPE" = "darwin"* ]]; then
```
mrclary, line 152: @ccordoba12, how about this instead? macOS install log can have the output too.
```suggestion
echo "Launching Spyder now..."
if [[ "$OSTYPE" = "darwin"* ]]; then
```

installers-conda/resources/post-install.sh
mrclary, line 170:
```suggestion
    nohup $spy_exe $>/dev/null &
```
I think we may need this so that Spyder is not killed when the installer exits, if not launched with gtk-launch.

mrclary:
@dalthviz, so switching to powershell did allow me to eliminate the minimized window. However, there is still a temporary "flash" of the powershell window. This is the same issue @jaimergp and I ran into with the lnk target and `menuinst`. I think it may be either a bug in powershell or I just don't know the right way to work it 😬. Nevertheless, I think it may be an improvement over a minimized window. What do you think?

installers-conda/resources/post-install.sh
ccordoba12: Sure, that's fine for me.

installers-conda/resources/post-install.sh
ccordoba12: Let me check and I'll get back to you.

mrclary: I should clarify: when the shell exits, not just the installer.

The following only applies if `gtk-launch` is _not_ used. If the user launches the installer from the commandline, after the installer is done and Spyder is launched, the shell session remains open; when the user closes this shell session, Spyder will be killed. If the user runs the installer from the file explorer (right-click -> "Run as program"), then a new shell session opens, runs the installer, launches Spyder, and exits, killing Spyder before the user even knows it opens.

dalthviz:
> Nevertheless, I think it may be an improvement over a minimized window. What do you think?

Checked locally and for sure is an improvement! Even if it shows for a moment, I would say is better since users will not be able to interact with a cmd :+1:

installers-conda/resources/post-install.sh
ccordoba12: In my case, after running the installer in a shell, closing it didn't close Spyder but blocked the shell. So, I think using `nohup` is necessary in this case.

installers-conda/resources/post-install.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -i
```
ccordoba12, line 1: I noticed that this change adds all contents of this script to my bash history (which is really annoying). So you should disable doing that. [Here](https://stackoverflow.com/q/6475524/438386) are some options to achieve it.

By the way, the same is happening for the new script you added that gets the user env vars declared in `~/.bashrc`. So, you should do the same for it (either here or in another PR, up to you).

mrclary: Excellent observation! I would not have caught that. I'll fix the installers in this PR and make another one for the user env variables.

installers-conda/resources/post-install.sh
```diff
@@ -1,4 +1,4 @@
...
 set -e
```
mrclary, line 2: @ccordoba12, this should do the trick.
```suggestion
set -e
unset HISTFILE
```

ccordoba12: Yep, I think so too.

ccordoba12 (approved):
Works like a charm for me now, thanks @mrclary!

ccordoba12:
@dalthviz, are you good to merge now?

dalthviz (approved):
Thanks @mrclary !

PR #21102, opened 2023-07-06
PR: Fix error when running main window tests and failing tests

### LINKED ISSUE #21322: Tests seems to be passing but fail
Tests seems to be passing but fail. If you go to the detail of the tests, failing tests come back with a little green check somehow.

For example : https://github.com/spyder-ide/spyder/actions/runs/6132598053/job/16643637158?pr=21320

Under "run test" I get:
```
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
[90](https://github.com/spyder-ide/spyder/actions/runs/6132598053/job/16643637158?pr=21320#step:15:91)
======================== 8 warnings, 1 error in 19.75s =========================
```

But the test has a little green check???

ccordoba12:
## Description of Changes

- That error was preventing to reuse the window between tests, which significantly increased its running time on CIs.
- Decrease percentage of IPython tests run on slow slots a bit.
- Fix some tests related to the switcher, dependencies and getting environment variables on Mac.

### Issue(s) Resolved

Fixes #21322.

### Affirmation

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -84,11 +84,11 @@ def pytest_collection_modifyitems(config, items):
         ]
 
         if os.name == 'nt':
-            percentage = 0.6
-        elif sys.platform == 'darwin':
-            percentage = 0.6
-        else:
             percentage = 0.5
+        elif sys.platform == 'darwin':
+            percentage = 0.5
+        else:
+            percentage = 0.4
 
         for i, item in enumerate(ipyconsole_items):
             if i < len(ipyconsole_items) * percentage:
--- a/spyder/app/tests/conftest.py
+++ b/spyder/app/tests/conftest.py
@@ -462,12 +462,13 @@ def main_window(request, tmpdir, qtbot):
                 CONF.reset_to_defaults(notification=False)
             else:
                 try:
-                    # Close everything we can think of
-                    window.switcher.close()
+                    # Close or hide everything we can think of
+                    window.switcher.hide()
 
                     # Close editor related elements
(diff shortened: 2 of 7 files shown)


mrclary:
Okay, I found the issue. The subprocess communicate timeout is too short for CI.
I recommend [this line](https://github.com/ccordoba12/spyder/blob/938fa5a4751a04bd9d7ae651c6f4eeb6c5c3b5ae/spyder/utils/environ.py#L107) be changed to 
```
            stdout, stderr = proc.communicate(timeout=3 if os.getenv("CI") else 0.5)
```
or similar.

I was able to ssh into the workflow and tinker with it there. It seems on CI a timeout error results for 0.5 and 1, but not 2. I think 3 just gives us some extra room.

I don't know why this should be so much slower on CI than on my local machine; just performance? 🤷🏼 
If that is the case, we may want to increase the timeout to accommodate users with slow machines and not just for CI.

ccordoba12:
> Okay, I found the issue. The subprocess communicate timeout is too short for CI.

Great! Thanks @mrclary for your help with that.

> I don't know why this should be so much slower on CI than on my local machine; just performance? 🤷🏼
If that is the case, we may want to increase the timeout to accommodate users with slow machines and not just for CI.

The problem is that getting env vars is run at startup to populate our PPM and also when it's open. So, if we increase that timeout, Spyder would appear as frozen when running those operations.

PR #21125, opened 2023-07-12
PR: Update to napari/label/bundle_tools_3

mrclary:


.github/workflows/installers-conda.yml
```diff
@@ -272,14 +280,20 @@ jobs:
...
+          root_prefix=$(compgen -G $HOME/.local/spyder-*)
+
+          echo "Install info:"
+          echo "Contents of ${root_prefix}:"
+          ls -al $root_prefix
+          echo -e "\nContents of ${root_prefix}/uninstall-spyder.sh:"
+          cat ${root_prefix}/uninstall-spyder.sh
+
+          shortcut_path=/.local/share/applications/spyder_spyder.desktop
+          [[ -e ${root_prefix}/.nonadmin ]] && shortcut_path=${HOME}${shortcut_path} || shortcut_path=/usr${shortcut_path}
```
ccordoba12, line 292: I'm not sure this logic is correct for Linux. If I understand things correctly, with this change we'd install our desktop file to `/usr/.local/share/applications/spyder_spyder.desktop` for admin installations.

But the location any application I know use is `/usr/share/applications`. So, I think this code should be changed to use that path instead.

installers-conda/resources/post-install.sh
```diff
@@ -17,11 +17,12 @@ u_spy_exe=${PREFIX}/uninstall-spyder.sh
...
 if [[ $OSTYPE = "darwin"* ]]; then
-    shortcut_path="/Applications/Spyder.app"
-    [[ ${PREFIX} = "$HOME"* ]] && shortcut_path="${HOME}${shortcut_path}"
+    shortcut_path="/Applications/${INSTALLER_NAME}.app"
+    [[ -e ${PREFIX}/.nonadmin ]] && shortcut_path="${HOME}${shortcut_path}"
     sed_opts+=("", "-e")
 else
-    shortcut_path="$HOME/.local/share/applications/${name_lower}_${name_lower}.desktop"
+    shortcut_path="/.local/share/applications/${name_lower}_${name_lower}.desktop"
+    [[ -e ${PREFIX}/.nonadmin ]] && shortcut_path="${HOME}${shortcut_path}" || shortcut_path="/usr${shortcut_path}"
```
ccordoba12, line 25: This has the same problem I described above.

ccordoba12 (commented):
A couple of comments for you @mrclary, the rest looks good to me.

.github/workflows/installers-conda.yml
mrclary: That's correct. I made a transcription error.

mrclary:
Actually, I'm going to verify some other issues with all-user installation for Linux; converting back to draft.

mrclary:
@ccordoba12, thanks for your previous review; I've made the suggested correction. Additionally, when installing for all users, each user's shell startup will have the Spyder launch alias. The uninstall alias is only included for user-only install.

ccordoba12:
@mrclary, looks good to me now, but I'd like to ask you if you have tested the admin install in a VM. I'd like to help you with that but I don't want to mess with my system.

mrclary:
> @mrclary, looks good to me now, but I'd like to ask you if you have tested the admin install in a VM. I'd like to help you with that but I don't want to mess with my system.

Yes, I've tested the admin install on my Ubuntu VM. I added a test user as well and confirmed that the Spyder alias is properly created and Spyder launches for all accounts.

Note that `constructor` does not provide a default install location for all users on admin installs; `/usr/spyder-6` (or wherever) _must_ be specified either with `-p` flag or at the interactive prompt. Otherwise the environment prefix will fall back to `/root/.local/spyder-6` which will causes a permission denied error for any user other than root.

Ping @jaimergp.

mrclary:
@ccordoba12, I've removed the uninstall Spyder alias from shell startup for admin installs because it requires sudo privileges and when invoking sudo the alias isn't recognized anyway. Should the uninstall Spyder alias be added to `/root/.bashrc` in this case?

installers-conda/resources/spyder-menu.json
```diff
@@ -12,6 +12,7 @@
             "platforms": {
                 "win": {
                     "desktop": true,
+                    "app_user_model_id": "Spyder-IDE.Spyder",
```
jaimergp, line 15: According to my tests with `napari`, Qt seems to be setting this quite strictly. In our case it's `napari.napari.viewer.PKG_VERSION` (with PKG_VERSION replaced appropiately).

You can check which one you get in Spyder once running with this snippet:

```python
import ctypes
from ctypes import wintypes

lpBuffer = wintypes.LPWSTR()
AppUserModelID = ctypes.windll.shell32.GetCurrentProcessExplicitAppUserModelID
AppUserModelID(ctypes.cast(ctypes.byref(lpBuffer), wintypes.LPWSTR))
appid = lpBuffer.value
ctypes.windll.kernel32.LocalFree(lpBuffer)
print(appid)
```

mrclary: Thanks! I'll replace with the appropriate value.

ccordoba12:
> Yes, I've tested the admin install on my Ubuntu VM. I added a test user as well and confirmed that the Spyder alias is properly created and Spyder launches for all accounts.

Ok, great!

> Note that constructor does not provide a default install location for all users on admin installs; /usr/spyder-6 (or wherever) must be specified either with -p flag or at the interactive prompt. Otherwise the environment prefix will fall back to /root/.local/spyder-6 which will causes a permission denied error for any user other than root.

I think that the default directory should be `/opt/spyder-6`, but we can add it once constructor has support for it.

> I've removed the uninstall Spyder alias from shell startup for admin installs because it requires sudo privileges and when invoking sudo the alias isn't recognized anyway.

Ok, no problem. Then I recommend to show a message saying that system admins need to remove the directory where Spyder was installed plus the desktop file to uninstall it.

> Should the uninstall Spyder alias be added to /root/.bashrc in this case?

That probably won't work and [the alternatives](https://askubuntu.com/questions/22037/aliases-not-available-when-using-sudo) don't seem promising.

I think adding a proper script in `/path/to/install_dir/uninstall-spyder.sh` should be enough. The thing is in that case we don't want to provide a global command to uninstall Spyder because regular users won't be able to run it and it doesn't make sense for them.

mrclary:
> I think that the default directory should be `/opt/spyder-6`, but we can add it once constructor has support for it.

👍🏼 
 
> Ok, no problem. Then I recommend to show a message saying that system admins need to remove the directory where Spyder was installed plus the desktop file to uninstall it.

The uninstall script is always created, so system admins can still run
```
sudo /path/to/spyder-6/uninstall-spyder.sh
```
I'll add to the message for admin users how to uninstall Spyder.
 
> > Should the uninstall Spyder alias be added to /root/.bashrc in this case?
> 
> That probably won't work and [the alternatives](https://askubuntu.com/questions/22037/aliases-not-available-when-using-sudo) don't seem promising.

I see; I agree that will not work.

> I think adding a proper script in `/path/to/install_dir/uninstall-spyder.sh` should be enough. The thing is in that case we don't want to provide a global command to uninstall Spyder because regular users won't be able to run it and it doesn't make sense for them.

👍🏼

mrclary:
@ccordoba12, here is the proposed messages:

### Admin install (all-users)

```
###############################################################################
#                             !!! IMPORTANT !!!
###############################################################################
Spyder can be launched by standard methods in Gnome and KDE desktop
environments. It can also be launched from the command line on all Linux
distros with the command:

$ spyder

This command will only be available in new shell sessions. To make it
available in this session, you must source your /home/rclary/.bashrc file with:

$ source /home/rclary/.bashrc

To uninstall Spyder, run the following from the command line:

$ sudo /opt/spyder-6/uninstall-spyder.sh

###############################################################################
```

### User install

```
###############################################################################
#                             !!! IMPORTANT !!!
###############################################################################
Spyder can be launched by standard methods in Gnome and KDE desktop
environments. It can also be launched from the command line on all Linux
distros with the command:

$ spyder

To uninstall Spyder, run the following from the command line:

$ uninstall-spyder

These commands will only be available in new shell sessions. To make them
available in this session, you must source your /home/rclary/.bashrc file with:

$ source /home/rclary/.bashrc

###############################################################################
```

ccordoba12:
Looks good to me, but I have one question. The message for admin installs says:

> This command will only be available in new shell sessions. To make it
available in this session, you must source your /home/rclary/.bashrc file with:
>
> $ source /home/rclary/.bashrc

How are you writing to `~/.bashrc` from the installer, which is running under `sudo`? I'm worried that this will make users' `~/.bashrc` to be not writable by themselves after the installer modifies it.

I think a cleaner solution would be to add the `spyder` alias in `/etc/bash.bashrc`, which (according to its first line) is the "system-wide .bashrc file for interactive bash(1) shells". I guess something similar exists for `zsh`.

mrclary:
> How are you writing to `~/.bashrc` from the installer, which is running under `sudo`? I'm worried that this will make users' `~/.bashrc` to be not writable by themselves after the installer modifies it.
> 
> I think a cleaner solution would be to add the `spyder` alias in `/etc/bash.bashrc`, which (according to its first line) is the "system-wide .bashrc file for interactive bash(1) shells". I guess something similar exists for `zsh`.

Aahh. I didn't check whether permissions were messed up by writing using sudo. Nevertheless, your suggestion is a much better approach anyway. I will implement. Perhaps the instruction to source .bashrc in the current shell should be omitted in the case of admin installs?

ccordoba12:
> Perhaps the instruction to source .bashrc in the current shell should be omitted in the case of admin installs?

Agreed. That part should be replaced by telling users to open a new terminal and start Spyder there.

mrclary:
@ccordoba12 here is the new message for admin install on Linux.
```
###############################################################################
#                             !!! IMPORTANT !!!
###############################################################################
Spyder can be launched by standard methods in Gnome and KDE desktop
environments. It can also be launched from the command line on all Linux
distros with the command:

$ spyder

This command will only be available in new shell sessions.

To uninstall Spyder, run the following from the command line:

$ sudo /opt/spyder-6/uninstall-spyder.sh

###############################################################################
```

ccordoba12:
> @ccordoba12 here is the new message for admin install on Linux.

Looks good to me now, thanks!

installers-conda/resources/post-install.sh
```diff
@@ -48,19 +53,36 @@ add_alias() (
...
+if [[ "$all_user" = "true" ]]; then
+    shell_init_list=("/etc/zshrc")
+    [[ "$OSTYPE" = "darwin"* ]] && shell_init_list+=("/etc/bashrc") || shell_init_list+=("/etc/bash.bashrc")
+else
+    case $SHELL in
+        (*"zsh") shell_init_list=("$HOME/.zshrc") ;;
+        (*"bash") shell_init_list=("$HOME/.bashrc") ;;
+    esac
+fi
+for shell_init in ${shell_init_list[@]}; do
```
ccordoba12, line 80:
```suggestion

for shell_init in ${shell_init_list[@]}; do
```

ccordoba12 (commented):
A couple of additional comments/suggestions, then this should be ready.

installers-conda/resources/post-install.sh
```diff
@@ -48,19 +53,36 @@ add_alias() (
...
             H
         }
         x
     }" $shell_init
-)
+}
+
+# ----
+if [[ "$all_user" = "true" ]]; then
+    shell_init_list=("/etc/zshrc")
```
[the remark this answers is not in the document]
mrclary: On macOS it is `/etc/zshrc`. I'll make the accommodation.

installers-conda/resources/post-install.sh
```diff
@@ -48,19 +53,36 @@ add_alias() (
...
+    [[ "$OSTYPE" = "darwin"* ]] && shell_init_list+=("/etc/bashrc") || shell_init_list+=("/etc/bash.bashrc")
```
mrclary, line 73:
```suggestion
if [[ "$all_user" = "true" && "$OSTYPE" = "darwin"* ]]; then
    shell_init_list=("/etc/zshrc" "/etc/bashrc")
elif [[ "$all_user" = "true" ]]; then
    shell_init_list=("/etc/zsh/zshrc" "/etc/bash.bashrc")
```

.github/workflows/installers-conda.yml
```diff
@@ -272,14 +280,20 @@ jobs:
...
+          root_prefix=$(compgen -G $HOME/.local/spyder-*)
+
+          echo "Install info:"
+          echo "Contents of ${root_prefix}:"
+          ls -al $root_prefix
+          echo -e "\nContents of ${root_prefix}/uninstall-spyder.sh:"
+          cat ${root_prefix}/uninstall-spyder.sh
+
+          shortcut_path=/share/applications/spyder_spyder.desktop
+          [[ -e ${root_prefix}/.nonadmin ]] && shortcut_path=${HOME}/.local${shortcut_path} || shortcut_path=/usr${shortcut_path}
```
ccordoba12, line 292:
```suggestion
          [[ -e ${root_prefix}/.nonadmin ]] && shortcut_path=${HOME}/.local${shortcut_path} || shortcut_path=/usr/local${shortcut_path}
```

installers-conda/resources/post-install.sh
```diff
@@ -13,29 +13,34 @@ echo ""
...
+        shortcut_path="${HOME}${shortcut_path}"
+        unset alias_text  # Do not create uninstall alias
+    fi
     sed_opts+=("", "-e")
 else
-    shortcut_path="$HOME/.local/share/applications/${name_lower}_${name_lower}.desktop"
-    alias_text="alias spyder=${spy_exe}\n${alias_text}"
+    shortcut_path="/share/applications/${name_lower}_${name_lower}.desktop"
+    if [[ "$all_user" = "true" ]]; then
+        shortcut_path="/usr${shortcut_path}"
```
ccordoba12, line 29:
```suggestion
        shortcut_path="/usr/local${shortcut_path}"
```

ccordoba12 (commented):
Last suggestions to prevent overwriting the Spyder desktop file installed by the Linux distro package manager.

installers-conda/resources/post-install.sh
mrclary: What is the name of the desktop file created by the Linux distro package manager?
Couple of points:
* `menuinst` names the desktop file `spyder_spyder.desktop`. We have control over the app name, e.g. "spyder", but `menuinst` will always name the file `<app-name>_<app_name>.desktop`.
* `menuinst` will always place the shortcut in `/usr/share/applications` or `$HOME/.local/share/applications` and I don't think we have any control over that.

So, I don't think your suggestion is possible to do, but if desktop files don't have a name conflict, then perhaps there isn't a problem.

@jaimergp, am I mistaken about any of the above?

jitseniesen:
> here is the new message for admin install on Linux.

@mrclary Sorry to butt in, but I think it looks more professional to say "distributions" in full instead of "distros".

installers-conda/resources/post-install.sh
jaimergp: That's correct. It's the slugified names ([see here](https://github.com/conda/menuinst/blob/23c580b5d798db157ced01cc11278c1436572d3b/menuinst/platforms/linux.py#L184-L189)) and the XDG standard paths ([see here](https://github.com/conda/menuinst/blob/23c580b5d798db157ced01cc11278c1436572d3b/menuinst/platforms/linux.py#L28-L55)).

ccordoba12: > What is the name of the desktop file created by the Linux distro package manager?

I check in Ubuntu and it's `spyder.desktop`:

https://packages.ubuntu.com/kinetic/all/spyder/filelist

So, we should be fine.

> menuinst will always place the shortcut in /usr/share/applications

Ok, I didn't know that. Then please disregard my suggestion.

ccordoba12 (approved):

PR #21132, opened 2023-07-13
PR: Don't write to shell history file when getting user environment variables

mrclary:


ccordoba12 (approved):
Works as expected, thanks @mrclary!

PR #21071, opened 2023-06-26
PR: Change pane tabs design

jsbautista:
## Description of Changes

Change design of pane tabs for them to look different to the tabs of the console and editor

A preview:

* Tab selected:
image

* Hover over tab not selected:
image

* Scroll buttons when there is not enough space to show all the tabs
image

### Issue(s) Resolved

Fixes https://github.com/spyder-ide/ux-improvements/issues/4

### Affirmation

By submitting this Pull Request or typing my (user)name below,
I affirm the [Developer Certificate of Origin](https://developercertificate.org)
with respect to all commits and content included in this PR,
and understand I am releasing the same under Spyder's MIT (Expat) license.

I certify the above statement is true and correct:
@jsbautista

PR #21133, opened 2023-07-14
PR: Improve style of dockwidget tabbars (UI)

ccordoba12:
## Description of Changes

- This is another take on the design discussed on spyder-ide/ux-improvements#4.
- Supersedes PR #21071.

**Dark theme**

new-tabbar-on-dark-theme

**Light theme**

new-tabbar-on-light-theme

### Issue(s) Resolved

Fixes spyder-ide/ux-improvements#4.

### Affirmation

### DIFF
--- a/spyder/plugins/editor/plugin.py
+++ b/spyder/plugins/editor/plugin.py
@@ -70,6 +70,7 @@ from spyder.plugins.run.api import (
     RunContext, RunConfigurationMetadata, RunConfiguration,
     SupportedExtensionContexts, ExtendedContext)
 from spyder.plugins.toolbar.api import ApplicationToolbars
+from spyder.utils.stylesheet import MARGIN_SIZE
 from spyder.widgets.mixins import BaseEditMixin
 from spyder.widgets.simplecodeeditor import SimpleCodeEditor
 
@@ -367,6 +368,11 @@ class Editor(SpyderPluginWidget, SpyderConfigurationObserver):
         self.find_widget = FindReplace(self, enable_replace=True)
         self.find_widget.hide()
         self.register_widget_shortcuts(self.find_widget)
+
+        # TODO: This is a hack! Remove it after migrating to the new API
+        self.find_widget.layout().setContentsMargins(
+            2 * MARGIN_SIZE, MARGIN_SIZE, 2 * MARGIN_SIZE, MARGIN_SIZE
+        )
 
         # Start autosave component
         # (needs to be done before EditorSplitter)
--- a/spyder/utils/stylesheet.py
+++ b/spyder/utils/stylesheet.py
@@ -300,13 +300,14 @@ class BaseTabBarStyleSheet(SpyderStyleSheet):
         css = self.get_stylesheet()
         buttons_color = QStylePalette.COLOR_BACKGROUND_1
 
-        # Set style for scroller buttons
+        # Set style for scroll buttons
         css[f'QTabBar{self.OBJECT_NAME} QToolButton'].setValues(
             background=buttons_color,
             borderRadius='0px',
             borderRight=f'{self.BORDER_RIGHT_WIDTH} solid {buttons_color}'
         )
 
+        # Hover and pressed state for scroll buttons
         for state in ['hover', 'pressed', 'checked', 'checked:hover']:
             if state == 'hover':
                 color = QStylePalette.COLOR_BACKGROUND_2
@@ -316,6 +317,7 @@ class BaseTabBarStyleSheet(SpyderStyleSheet):
                 background=color
             )
 
+        # Set width for scroll buttons
         # This makes one button huge and the other very small in PyQt 5.9
         if not OLD_PYQT:
             css['QTabBar::scroller'].setValues(
@@ -412,6 +414,12 @@ class PanesTabBarStyleSheet(PanesToolbarStyleSheet, BaseTabBarStyleSheet):
             right='-3px' if WIN else '-1px'
         )
 
+        # Make scroll button icons smaller on Windows and Mac
+        if WIN or MAC:
+            css[f'QTabBar{self.OBJECT_NAME} QToolButton'].setValues(
+                padding='9px',
+            )
+
 
 class DockTabBarStyleSheet(BaseTabBarStyleSheet):
     """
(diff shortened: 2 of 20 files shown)


dalthviz (commented):
Thanks @ccordoba12 ! Checking on Windows and seems like is looking as previewed :+1:

The only thing that I noticed is that the scroll buttons for the tabs are now bigger. Is that the wanted look, right?:

Before:

image

After:

image

Also, just in case, for me the scroll buttons for the pane tabs look like:

image

ccordoba12:
@dalthviz, thanks for the review! I think I fixed the errors you found, so please test again.

I also made the additional changes:

* I revisited margins for all panes to make them work as expected with the new style.
* I asked for a restart to change the vertical tabs option because it requires now to adjust all PluginMainWidget margins.

dalthviz:
Checking this again seems like the tab arrows are now a little bit too small:

image
image

However, now the style adapts to the vertical layout :+1:

image

spyder/utils/stylesheet.py
```diff
@@ -396,8 +438,212 @@ def set_stylesheet(self):
             right='-3px' if WIN else '-1px'
         )
 
+        # Make scroll button icons smaller on Windows and Mac
+        if WIN or MAC:
+            css[f'QTabBar{self.OBJECT_NAME} QToolButton'].setValues(
+                padding='9px',
```
ccordoba12, line 444: This is for the Editor/IPython console tabs.

spyder/utils/stylesheet.py
```diff
@@ -396,8 +438,212 @@ def set_stylesheet(self):
...
+                f'{QStylePalette.COLOR_TEXT_1}' if is_dark_interface() else
+                f'{QStylePalette.COLOR_BACKGROUND_1}'
+            ),
+            backgroundColor=f'{self.color_selected_tab}',
+        )
+
+        # Make scroll button icons smaller on Windows and Mac
+        if WIN or MAC:
+            css['QTabBar QToolButton'].setValues(
+                padding='10px',
```
ccordoba12, line 483: This is for the new dockwidget tabs.

ccordoba12 (commented):
> Checking this again seems like the tab arrows are now a little bit too small

Ok, I adjusted that size to be similar to the one I'm seeing on Linux. But the nice thing is that it can be changed to be a bit bigger (which can't be done on Linux).

Below I left a comment in the places where I set those values, so please change them and let me know which ones work better for you.

spyder/utils/stylesheet.py
dalthviz: Using `5px` here I got:

image
image

spyder/utils/stylesheet.py
dalthviz: Using `4px` here I got:

image

spyder/utils/stylesheet.py
ccordoba12: Ok, thanks for the suggestion @dalthviz.

@conradolandia, what do you think about the size of these arrows? Do they look to you? I find them a bit too big, but I don't have a strong preference.

conradolandia: I thinkk they look too big, yes. They should not be taller than the hamburger icon.

ccordoba12: > They should not be taller than the hamburger icon.

Yep, I agree.

@dalthviz, I selected the arrows size to match the one of the other icons in the toolbar. Perhaps we could make them two pixels bigger so they don't appear too small, but I'd say no more than that.

dalthviz: Checking different values I got:

* `8px`:
image
image

* `7px`:

image
image

spyder/utils/stylesheet.py
dalthviz: To follow size with https://github.com/spyder-ide/spyder/pull/21133#discussion_r1272797950 values:

* `5px`:

image


* `6px`:

image

spyder/utils/stylesheet.py
ccordoba12: Thanks for the suggestions @dalthviz! I'll discuss them with @conradolandia in our UX meeting and I'll get back to you.

ccordoba12: @conradolandia mentioned he prefers `7px` here.

spyder/utils/stylesheet.py
ccordoba12: And `5px` here.

ccordoba12:
@dalthviz, this should be ready, unless you have more comments about the code.

dalthviz (approved):
Thanks @ccordoba12 for all the work here! I left a question regarding the line commented as a hack but otherwise this LGTM :+1:

spyder/plugins/editor/plugin.py
```diff
@@ -368,6 +369,11 @@ def __init__(self, parent, ignore_last_opened_files=False):
         self.find_widget.hide()
         self.register_widget_shortcuts(self.find_widget)
 
+        # TODO: This is a hack! Remove it after migrating to the new API
+        self.find_widget.layout().setContentsMargins(
+            2 * MARGIN_SIZE, MARGIN_SIZE, 2 * MARGIN_SIZE, MARGIN_SIZE
+        )
+
```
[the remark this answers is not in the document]
ccordoba12: > Why this was needed?

Because in this PR I added a bottom margin to all migrated plugins [here](https://github.com/spyder-ide/spyder/pull/21133/files#diff-64ac70728bd00daaa0c8a7638fe7633846b11d5c5260ee5ab5bf92e571a207d3R300), so that their borders don't touch the bottom. And by doing that, it was not necessary to add a bottom margin to the Find/Replace widget, so I removed it [here](https://github.com/spyder-ide/spyder/pull/21133/files#diff-18cad705e6ebe50fa941debcacfd9b695459a7f11ff32910b6b1bbb2b09aba35L74).

But since the editor is not migrated, we need to add that margin in the meantime (which is the change you highlighted here). Otherwise the Find/Replace widget would appear touching the bottom.

> Are there other changes that need to be done besides removing this line for the migration?

Not from this PR, but I've been documenting other things that need to be removed after the migration with TODOs in `spyder/plugins/editor/plugin.py`.

